PR #229, opened 2019-06-15
ref: Use C# 8 and nullable reference types

bruno-garcia:
Includes a work around for https://github.com/dotnet/roslyn/issues/34724 that should be undone once the tooling ships with a fix.

First commit was just to get `Sentry` project to compile with `<Nullable>` on.

- [x] Use stable 3.0.0 SDK
- [ ] Undo work around for Roslyn bug
- [ ] Opt in to `<NullableContextOptions>` (renamed to `<Nullable>` on the next SDK release) at the root `Directory.Build.props`
- [ ] Review the code to see where nullable reference can be dropped.
- [ ] Consider impact on consumers pre-C#8 (should we add/keep `throw new Arg` or other guard?)


bruno-garcia:
With regards to C# 8 on netstandard2.0:

https://github.com/dotnet/corefx/issues/40039#issuecomment-520133335

Issue #204, opened 2019-05-24
Add Roslyn analyzers


bruno-garcia:
Consider adding a public API analyzer to help avoid publishing packages with breaking changes without bumping a major version (semver).

Consider: [PublicApiAnalyzers](https://www.nuget.org/packages/Microsoft.CodeAnalysis.PublicApiAnalyzers/).
Other [Roslyn analyzers here](https://github.com/dotnet/roslyn-analyzers).
This would then also apply to other .NET repos in Sentry like:
* [sentry-dotnet-protocol](https://github.com/getsentry/sentry-dotnet-protocol)
* [sentry-dotnet-platform-abstraction](https://github.com/getsentry/sentry-dotnet-platform-abstraction)
* [sentry-dotnet-ef](https://github.com/getsentry/sentry-dotnet-ef)

PR #234, opened 2019-06-22
Adds PublicApiAnalyzers to public projects

josh-degraw:
RE: #204
Notes: 
- Moved `PackageReference` for `PublicApiAnalyzers` into `src` directory so it only shows for public projects.
- Turned off warning for `RS0017 (Remove deleted types and members from the declared API)`. 
  - Since some properties (specifically `SentryOptions.MaxRequestBodySize`) are only declared for certain platforms, the warning will show up if it's in the `.txt` file, and `RS0016 (Add public types and members to the declared API)` will show up if it's not, and since `TreatWarningsAsErrors` is enabled, it breaks the build either way.
  - I opted to suppress the first so that things will always be included, but that's a potential bug in the analyzer itself that should be noted.


bruno-garcia:
:shipit: ! Thanks Josh!

PR #238, opened 2019-07-01
tests: NLog default levels

josh-degraw:
Parity with Sentry.Serilog RE: #237 for testing default NLog log level options.


josh-degraw:
No problem!

PR #247, opened 2019-07-23
feat: Lock dependencies with NuGet

bruno-garcia:


bruno-garcia:
A new failed attempt exists (#731) so this can go away now.

PR #256, opened 2019-07-30
ref: Don't depend on System.Collections.Immutable

bruno-garcia:
Experiment to drop support to `System.Collections.Immutable` which didn't make it into .NET Standard 2.1 and will continue as an external dependency.


snakefoot:
Superseded by #405

bruno-garcia:
\o/

Issue #303, opened 2019-11-10
IApplicationLifetime and IHostingEnvironment are obsolete on ASP.NET Core 3


bruno-garcia:
Find usages such as [this](https://github.com/getsentry/sentry-dotnet/blob/50444e51b279d788ec89d5c45b9f9ebad1461b62/src/Sentry.AspNetCore/ApplicationBuilderExtensions.cs#L58) and [this](https://github.com/getsentry/sentry-dotnet/blob/6bd0f428e77d10b9a1518758b5f6564a5a2b8839/src/Sentry.AspNetCore/SentryAspNetCoreOptionsSetup.cs#L16) and replace accordingly.

See the [announcement for more](https://github.com/aspnet/AspNetCore/issues/7749)

bruno-garcia:
Solved in #306

PR #308, opened 2019-11-12
ref: multi-target netcore 3.0

bruno-garcia:
Addresses #303 but means targeting `netcoreapp3.0` which requires some thought and a major bump.

It was supposed to be a draft PR*


src/Sentry.AspNetCore/HostingEnvironmentExtensions.cs
```diff
@@ -0,0 +1,18 @@
...
+
+namespace Sentry.AspNetCore
+{
+    internal static class HostingEnvironmentExtensions
+    {
+        public static string RootPath(this IWebHostEnvironment env) => env.
+#if NETSTANDARD2_0
+            WebRootPath;
+#else
+            ContentRootPath;
```
khellang: These are not the same paths... 🙈

bruno-garcia: Lol, good catch, thanks @khellang
`WebRootPath` isn't there on the 3.0.0 version (it's the right namespace so I don't know what's wrong).

khellang: You could try to do something [like this] (https://github.com/khellang/Middleware/blob/fc75a0402df3a941c7fd464af4a8c116257e3e4e/src/ProblemDetails/ProblemDetailsMiddleware.cs#L14-L18) to reduce churn 😊

bruno-garcia: That looks better. Thanks for the tip.

PR #316, opened 2019-12-01
Add logger and category from Serilog SourceContext.

krisztiankocsis:
Logger and category (breadcrumbs) are lost when using Serilog extension. This PR adds support for them to store into the event and breadcrumbs.


krisztiankocsis:
Sure, I'll in the weekend.

src/Sentry.Serilog/SentrySink.cs
```diff
@@ -86,6 +96,7 @@ public void Emit(LogEvent logEvent)
                         Name = Constants.SdkName,
                         Version = NameAndVersion.Version
                     },
+                    Logger = logger,
```
bruno-garcia: Would this work instead of having to change the code above to assing `logger = context`?
```suggestion
                    Logger = context,
```

bruno-garcia:
Thanks for you PR and sorry for the delay.
Left a single comment before we can get this merged.

bruno-garcia:
Actually I just noticed you commited the suggestion but it involved getting rid of the `logger` variable altogether and replacing all occurrences with `context`.

Could you please fix that so it compiles?

bruno-garcia:
Thanks!

bruno-garcia:
Published: https://github.com/getsentry/sentry-dotnet/releases/tag/2.0.2

PR #323, opened 2019-12-17
null check contextProp.Value during IncludeEventDataOnBreadcrumbs

markballenger:
without this, our application blows up during logs when properties pass null values (which in some cases is intentional)


bruno-garcia:
Thanks for this @markballenger

bruno-garcia:
This fix was released in [2.0.0-beta7](https://github.com/getsentry/sentry-dotnet/releases/tag/2.0.0-beta7). Please give it a try before we release `2.0.0`.

markballenger:
Thanks will do

PR #324, opened 2019-12-17
fix: Reading request body throws on ASP.NET Core 3

### LINKED ISSUE #219: Async RequestPayloadExtractor?
The DefaultRequestPayloadExtractor has a comment stating that there would be no real value in making it asynchronous.
https://github.com/getsentry/sentry-dotnet/blob/e24115df3869601fb51a3d37392ae031a9ddb83b/src/Sentry/Extensibility/DefaultRequestPayloadExtractor.cs#L21

Whilst there might not be any gain in performance by making it async, I would like to suggest a scenario where there would be value in doing so.

There is nowhere in our APIs where we need to use synchronous IO, so to prevent developers from doing so accidentally, we configure Kestrel to disallow it:

```c#
builder.UseKestrel(options =>
{
    options.AllowSynchronousIO = false;
});
```

However setting those Kestrel options combined with Sentry reading the body synchronously means that when logging an exception to Sentry, we get an additional exception in the breadcrumbs in the event and the request body is also not part of the event because it could not be read.

```
Failed invoking event handler: System.InvalidOperationException: Synchronous operations are disallowed. Call ReadAsync or set AllowSynchronousIO to true instead.
   at Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http.HttpRequestStream.Read(Byte[] buffer, Int32 offset, Int32 count)
   at Microsoft.AspNetCore.WebUtilities.FileBufferingReadStream.Read(Byte[] buffer, Int32 offset, Int32 count)
   at System.IO.StreamReader.ReadBuffer()
   at System.IO.StreamReader.ReadToEnd()
   at Sentry.Extensibility.DefaultRequestPayloadExtractor.DoExtractPayLoad(IHttpRequest request) in C:\projects\sentry-dotnet\src\Sentry\Extensibility\DefaultRequestPayloadExtractor.cs:line 22
   at Sentry.Extensibility.BaseRequestPayloadExtractor.ExtractPayload(IHttpRequest request) in C:\projects\sentry-dotnet\src\Sentry\Extensibility\BaseRequestPayloadExtractor.cs:line 19
   at Sentry.Extensibility.RequestBodyExtractionDispatcher.ExtractPayload(IHttpRequest request) in C:\projects\sentry-dotnet\src\Sentry\Extensibility\RequestBodyExtractionDispatcher.cs:line 41
   at Sentry.AspNetCore.ScopeExtensions.SetBody(BaseScope scope, HttpContext context, SentryAspNetCoreOptions options) in C:\projects\sentry-dotnet\src\Sentry.AspNetCore\ScopeExtensions.cs:line 109
   at Sentry.AspNetCore.ScopeExtensions.Populate(Scope scope, HttpContext context, SentryAspNetCoreOptions options) in C:\projects\sentry-dotnet\src\Sentry.AspNetCore\ScopeExtensions.cs:line 43
   at Sentry.AspNetCore.SentryMiddleware.PopulateScope(HttpContext context, Scope scope) in C:\projects\sentry-dotnet\src\Sentry.AspNetCore\SentryMiddleware.cs:line 143
   at Sentry.Scope.Evaluate() in C:\projects\sentry-dotnet\src\Sentry\Scope.cs:line 135
```

bruno-garcia:
Resolves #219 which athough talks about changing the contract, only required to work around the breaking change done by ASP.NET Core 3.

Having this code `async` would mean changing the whole SDK to have everything `async/await`.

Meaning: `await SentrySdk.CaptureExceptionAsync(e);` which gives one the impression that's doing something async but in fact it won't unless the user **opts in** to capturing request bodies. In ASP.NET Core only (this SDK is used in many other scenarios).

This still wouldn't be useful to one of the most common ways of capturing events in Sentry. Using a logging integration.
Logger APIs are always sync: `_logger.LogError(..)`. Eventually it'll reach the request body via a sync API.
Also, the `HttpContext` object is not thread-safe so it can't just be given to a `Task.Run` of sorts.

Considering **request buffering** is only enabled if the **capture request body feature is opt in**, the SDK will only buffer request if you've chosen to. In this case, an error happens after the request is received and read (i.e: After model binding) in which case the `context.Request.Body` already has the data buffered anyway.

That means this is **not** a blocking call since the resolved Task will return synchronously with it's data.


bruno-garcia:
This fix was released in [2.0.0-beta7](https://github.com/getsentry/sentry-dotnet/releases/tag/2.0.0-beta7). Please give it a try before we release `2.0.0`.

ElBedeawi:
still not working on 2.0.0-beta7

michaelakin:
So, what is the fix?  I am getting this in asp.net core 3.1

Issue #322, opened 2019-12-17
Sentry.NLog and ${aspnet-user-identity} as user of event


oddeirik:
Please mark the type framework used:

- [X] ASP.NET MVC
- [ ] ASP.NET Web API (OWIN)
- [ ] ASP.NET Core
- [ ] WPF
- [ ] WinForms
- [ ] Xamarin
- [ ] Other:

Please mark the type of the runtime used:

- [X] .NET Framework
- [ ] Mono
- [ ] .NET Core

* Version:

Please mark the NuGet packages used:

- [X] Sentry
- [ ] Sentry.Serilog
- [X] Sentry.NLog
- [ ] Sentry.Log4Net
- [ ] Sentry.Extensions.Logging
- [ ] Sentry.AspNetCore

* Version: 1.2.0

The TLDR version is basically; I want to set the username in the Sentry NLog events to ${aspnet-user-identity}, any idea how this can be done?

The longer version:
SDK is initialized in Global.asax.cs, and on unhandled exceptions we're setting a scope to include the currently authenticated user (logged in via Azure AD). This works as expected for unhandled exceptions. (I've tried using `sendDefaultPii` here, but that only included the service user for the application, which isn't of interest to us.)

We also have a Sentry NLog target configured for log events containing exceptions where we don't necessarily want to throw an exception, but we want to send it to Sentry. Some of these exceptions are happening within the scope/request of an authenticated user, but for some reason this information is not included in the Sentry event.

I've tried setting `sendDefaultPii` in the NLog configuration, with no luck. I've also tried sending a custom tag with the user/username explicitly set:

`<tag name="user" layout="username:${aspnet-user-identity}" />`

However, this information is not included in the event being sent, so it's either ignored or filtered out. I've tried adding ${aspnet-user-identity} to log layout and the currently logged in user is being shown as expected there. Is there a way to set the scope of a log event with a specific username, similar to how we're doing it in Global.asax.cs?

josh-degraw:
So you're saying that the `aspnet-user-identity` variable works when you use it as the log message layout, but it gets dropped if it's included as a tag? That certainly is strange. I assume you're already referencing `NLog.Web` and including it in the NLog config file?

If you could post a minimal sample of your nlog config and some of the code of setting up the scope that could help nail down the problem.

Have you tried setting the username etc in code e.g.:
```c#
 SentrySdk.ConfigureScope(s => s.User.Username = theUserNameEtc);
```
?
[...]

PR #336, opened 2020-01-09
Adds ability to specify user values via NLog configuration

josh-degraw:
RE: #322 
I opted to have a user object to keep the user information logically coupled. If it would be more clear in another format I'm fine with changing it, but I also figured it would be more straightforward to do it this way especially if we want to add support for structured contexts or other things (i.e. fingerprint) for them to be logically structured like in sentry documentation.

### DIFF
--- a/src/Sentry.NLog/SentryTarget.cs
+++ b/src/Sentry.NLog/SentryTarget.cs
@@ -194,6 +194,15 @@ namespace Sentry.NLog
             set => Options.FlushTimeout = TimeSpan.FromSeconds(value);
         }
 
+        /// <summary>
+        /// Optionally configure one or more parts of the user information to be rendered dynamically from an NLog layout
+        /// </summary>
+        public SentryNLogUser User
+        {
+            get => Options.User;
+            set => Options.User = value;
+        }
+
         /// <inheritdoc />
         protected override void CloseTarget()
         {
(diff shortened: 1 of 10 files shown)


src/Sentry.NLog/SentryTarget.cs
```diff
@@ -345,6 +355,21 @@ protected override void Write(LogEventInfo logEvent)
             }
         }
 
+        private User GetUser(LogEventInfo logEvent)
+        {
+            if (User == null)
+                return null;
```
bruno-garcia: Please add `{}` to keep consitency with the code-base. 
If there's a flag for `editorconfig` it would be nice to get that on

test/Sentry.NLog.Tests/SentryTargetTests.cs
```diff
@@ -685,6 +715,25 @@ public void GetTagsFromLogEvent_NullPropertiesMapped()
...
+                Username = "${logger:shortName=true}"
+            };
+
+            var logger = factory.GetLogger("sentry");
+
+            logger.Fatal(DefaultMessage);
+
+            _fixture.Hub.Received(1)
+                .CaptureEvent(Arg.Is<SentryEvent>(e => e.User.Username == "sentry"));
+
```
bruno-garcia:
```suggestion
```

bruno-garcia:
❤️ !

PR #353, opened 2020-02-07
Set DateFormatHandling.IsoDateFormat for serializer. Fixes #351

### LINKED ISSUE #351: Input contains invalid characters error when MicrosoftDateFormat is used

- [ ] ASP.NET MVC
- [x] ASP.NET Web API (OWIN)
- [ ] ASP.NET Core
- [ ] WPF
- [ ] WinForms
- [ ] Xamarin
- [ ] Other:

- [x] .NET Framework
- [ ] Mono
- [ ] .NET Core

* Version:

- [x] Sentry
- [x] Sentry.Serilog
- [ ] Sentry.NLog
- [ ] Sentry.Log4Net
- [ ] Sentry.AspNetCore

* Version: 2.0.1

When we are trying to log an event, an HTTP body is not recorded. I can see HTTP headers, though.
There is an error in the Sentry event
image

I suppose that the issue was in the old SDK (Raven) too.

olsh:
Related #351 

According to [the documentation](https://www.newtonsoft.com/json/help/html/P_Newtonsoft_Json_JsonConvert_DefaultSettings.htm), the default settings are used by [JsonConvert](https://www.newtonsoft.com/json/help/html/T_Newtonsoft_Json_JsonConvert.htm) method.

So, we have two options:

- Override the default settings (this is done in the PR)
- Create a new serializer with [Create](https://www.newtonsoft.com/json/help/html/M_Newtonsoft_Json_JsonSerializer_Create.htm) method

I believe that the second approach is preferable since we have more control over the serializing settings (I can update the PR).

@bruno-garcia Please let me know what do you think.

### DIFF
--- a/src/Sentry/Internal/JsonSerializer.cs
+++ b/src/Sentry/Internal/JsonSerializer.cs
@@ -12,7 +12,8 @@ namespace Sentry.Internal
             NullValueHandling = NullValueHandling.Ignore,
             ReferenceLoopHandling = ReferenceLoopHandling.Ignore,
             Formatting = Formatting.None,
-            Converters = new[] { StringEnumConverter }
+            Converters = new[] { StringEnumConverter },
+            DateFormatHandling = DateFormatHandling.IsoDateFormat
         };
 
         public static string SerializeObject<T>(T @object) => JsonConvert.SerializeObject(@object, Settings);


bruno-garcia:
I believe the Create method might be a better option because it's possible users set other global settings to their app which break Sentry.

Thanks a lot for helping with this!

bruno-garcia:
I'll merge this meanwhile ;)

PR #358, opened 2020-02-11
Add possibility to override fingerprint using AdditionalGroupingKey property

Shtannikov:


Shtannikov:
@bruno-garcia thanks for replying so fast! ok, will do!

bruno-garcia:
Thanks!

Issue #366, opened 2020-02-24
AspNetCore Scope Extensions sets path incorrectly


christopher-taormina-zocdoc:
Please mark the type framework used:

- [X] ASP.NET MVC
- [ ] ASP.NET Web API (OWIN)
- [ ] ASP.NET Core
- [ ] WPF
- [ ] WinForms
- [ ] Xamarin
- [ ] Other:

Please mark the type of the runtime used:

- [ ] .NET Framework
- [ ] Mono
- [X] .NET Core

* Version:

Please mark the NuGet packages used:

- [ ] Sentry
- [ ] Sentry.Serilog
- [ ] Sentry.NLog
- [ ] Sentry.Log4Net
- [ ] Sentry.Extensions.Logging
- [X] Sentry.AspNetCore

* Version: 2.0.3

It appears that the AspNetCore package sets the scopes Request path to the http contexts request path, which is not the full url of the request. This leads to some functionality missing in the Sentry UI such as this ability to make the same request. 
![image](https://user-images.githubusercontent.com/32650271/75178400-098edf80-5706-11ea-8a5d-778c85701a0f.png)

This seems to happen at this line https://github.com/getsentry/sentry-dotnet/blob/6e6f3a6b52eae7be56c3a1dd11157b962c5ec222/src/Sentry.AspNetCore/ScopeExtensions.cs#L68.

Was this done for a particular reason? Using the web context we should be able to construct the full url as properties for scheme, host and path are all set.

bruno-garcia:
Resolved via #367

bruno-garcia:
Released as 2.1.0

https://www.nuget.org/packages/Sentry.AspNetCore/2.1.0

PR #367, opened 2020-02-24
fix: set score url to fully constructed url

christopher-taormina-zocdoc:
Previously the Url property on scope was only set as the path from the HttpContext. This is not the fully constructed url of the request and sending only the path here breaks some functionality in the Sentry UI. 

This addresses https://github.com/getsentry/sentry-dotnet/issues/366.


christopher-taormina-zocdoc:
No problem! When would we see a new release with this included?

bruno-garcia:
@christopher-taormina-zocdoc it's already indexed by nuget: version: 2.1.0

christopher-taormina-zocdoc:
@bruno-garcia Awesome thanks!

PR #374, opened 2020-03-19
Add environment option to Serilog sink, fixes #290

rmsy:
This adds `Environment` to the list of allowed configuration options for the Serilog sink. This fixes #290, making the following Serilog configuration in `appsettings.json` possible:

```json
      {
        "Name": "Sentry",
        "Args": {
          "Dsn": "<replaced>",
          "Environment": "QA",
          "MinimumBreadcrumbLevel": "Debug",
          "MinimumEventLevel": "Error"
        }
      }
```

This is my first PR to this repository, so please let me know if I've missed anything 😊 (I think I may need to update `PublicAPI.Unshipped.txt` - not sure on that.) Thanks.

Issue #290, opened 2019-10-24
Sentry.Serilog doesn't log Environment


DR9885:
Please mark the type framework used:

- [ ] ASP.NET MVC
- [ ] ASP.NET Web API (OWIN)
- [X] ASP.NET Core
- [ ] WPF
- [ ] WinForms
- [ ] Xamarin
- [ ] Other:

Please mark the type of the runtime used:

- [ ] .NET Framework
- [ ] Mono
- [X] .NET Core

* Version: 2.2.0

Please mark the NuGet packages used:

- [ ] Sentry
- [X] Sentry.Serilog
- [ ] Sentry.NLog
- [ ] Sentry.Log4Net
- [ ] Sentry.Extensions.Logging
- [ ] Sentry.AspNetCore

Please describe the the steps to reproduce the issue or link to a repository with a small reproducible code.

Currently we use Sentry.Serilog 1.2.0, and it does not correctly display Environment, it always displays "All Environments". We have "ASPNETCORE_ENVIRONMENT" set in the environment variables. We also use Serilog to distribute our logging, and have it configured with the config listed below.

Note: We dont use Sentry.AspNetCore for 2 reasons, 1 our console applications don't use AspNetCore it just uses NetCore, and 2 we use Serilog to enrich data meta data and to distribute our logging.

Also note, I changed the key below for security purposes.

```
  "Serilog": {
    "MinimumLevel": {
      "Default": "Information",
      "Override": {
        "Microsoft": "Warning",
        "System": "Warning"
      }
    },
    "WriteTo": [
      { "Name": "LiterateConsole" },
      {
        "Name": "Logentries",
        "Args": {
          "token": "99b37aaf-c472-40f7-a50f-5ab3d457f071-123",
          "outputTemplate": "{Timestamp:yyyy-MM-dd HH:mm:ss.fff zzz} [{Level:u3}] [{MachineName}] [{SourceContext}] {Message:lj}{NewLine}{Exception}"
        }
      },
      {
        "Name": "Sentry",
        "Args": {
          "MinimumBreadcrumbLevel": "Information",
          "MinimumEventLevel": "Error",
          "Dsn": "https://a16b8602ba8e46a49188c7de65af9072123@sentry.io/202852123",
          "AttachStacktrace": true,
          "SendDefaultPii": false
        }
      }
    ]
  }
```

bruno-garcia:
Thanks for raising this.
[...]

PR #380, opened 2020-03-28
Serilog Sentry sink does not load all options from IConfiguration (#290)

JimHume:
Note: this is effectively an extension of the pull request #374 from @rmsy

The json configuration loading for the Serilog sink was not taking into account all of the options that can be passed to the SentryOptions and SentrySerilogOptions objects.
I have added an additional extension that allows for all the configurable property items that can be loaded from the appsettings.json when NOT using asp.net (such as a windows service).

I added a new method instead of repurposing the existing one to ensure backwards compatibility. I also added an example in the xmldocs so that it is easier to understand how to properly configure the settings for Serilog in json.


JimHume:
https://github.com/getsentry/sentry-dotnet/issues/290

src/Sentry.Serilog/SentrySinkExtensions.cs
```diff
@@ -28,15 +32,149 @@ public static class SentrySinkExtensions
...
+        ///                 }
+        ///             }
+        ///         ]
+        ///     }
+        /// }
+        /// </code>
+        /// </example>
+        public static LoggerConfiguration Sentry(
+            this LoggerSinkConfiguration loggerConfiguration,
+            bool sendDefaultPii = false,
```
bruno-garcia: Please make all values nullable and verify `if (value.HasValue) .. before setting the value`:
Like:
```suggestion
            bool? sendDefaultPii = null,
```

src/Sentry.Serilog/SentrySinkExtensions.cs
```diff
@@ -28,15 +32,149 @@ public static class SentrySinkExtensions
...
+            string serverName = null,
+            bool attachStackTrace = false,
+            int maxBreadcrumbs = Constants.DefaultMaxBreadcrumbs,
+            float? sampleRate = null,
+            string release = null,
+            string environment = null,
+            string dsn = null,
+            int maxQueueItems = 30,
+            TimeSpan shutdownTimeout = default,
+            DecompressionMethods decompressionMethods = ~DecompressionMethods.None,
```
bruno-garcia: Same here, all value types to be nullable and only assign if not null

src/Sentry.Serilog/SentrySinkExtensions.cs
```diff
@@ -28,15 +32,149 @@ public static class SentrySinkExtensions
...
+            {
+                o.SendDefaultPii = sendDefaultPii;
+                o.IsEnvironmentUser = isEnvironmentUser;
+                o.ServerName = serverName;
+                o.AttachStacktrace = attachStackTrace;
+                o.MaxBreadcrumbs = maxBreadcrumbs;
+
+                if (sampleRate != null)
+                {
+                    o.SampleRate = sampleRate;
```
bruno-garcia: 👍

src/Sentry.Serilog/SentrySinkExtensions.cs
JimHume: Indeed--great idea / pattern.

JimHume:
Thank you for reviewing this @bruno-garcia -- I will make those changes and update the PR as soon as I can.

bruno-garcia:
@JimHume Thank you for contributing!

JimHume:
I've updated the PR with a few commits.
The last one is a bit of a stretch. I wasn't sure if the intent would be more obvious if it was portrayed through the json strings that Serilog uses & expects, so I went that route for now.

I have thick skin, so please let me know if this is overkill, ridiculous, or should otherwise be altered (I'm easy)--I'll use it as a learning experience at the very least.

Again, thank you for your time and review @bruno-garcia !

JimHume:
Update: the appveyor build failed due to an obscure error (transient, perhaps?)
`error:   Job update failed: Missing ZEUS_HOOK_BASE environment variable
+ check_branch
+ [[ ! master =~ ^release/ ]]
+ trap - EXIT`

src/Sentry.Serilog/SentrySinkExtensions.cs
```diff
@@ -28,15 +31,269 @@ public static class SentrySinkExtensions
...
+            {
+                sentrySerilogOptions.AttachStacktrace = attachStackTrace.Value;
+            }
+
+            if (maxBreadcrumbs.HasValue)
+            {
+                sentrySerilogOptions.MaxBreadcrumbs = maxBreadcrumbs.Value;
+            }
+
+            if (sampleRate != null) // .HasValue instead?
```
bruno-garcia: No strong opinion, [there's one occurance of `HasValue` in the code base](https://github.com/getsentry/sentry-dotnet/search?q=hasvalue&unscoped_q=hasvalue) so mostly `!= null` so far. Arguably `is null` and `HasValue` are "better" because `!=` is an operator.

But mostly since you used `HasValue` below, for consistency:

```suggestion
            if (sampleRate.HasValue)
```

test/Sentry.Serilog.Tests/Sentry.Serilog.Tests.csproj
```diff
@@ -4,8 +4,18 @@
     <TargetFrameworks>netcoreapp3.1;netcoreapp2.1;net461</TargetFrameworks>
   </PropertyGroup>
 
+  <ItemGroup>
+    <PackageReference Include="Microsoft.Extensions.Configuration" Version="2.1.1" />
+    <PackageReference Include="Microsoft.Extensions.Configuration.Abstractions" Version="2.1.1" />
+    <PackageReference Include="Microsoft.Extensions.Configuration.Json" Version="2.1.1" />
+    <PackageReference Include="Microsoft.Extensions.Primitives" Version="2.1.1" />
+    <PackageReference Include="Serilog" Version="2.7.1" />
```
bruno-garcia: Right now the `Serilog` dependency comes in through the project reference:

`<ProjectReference Include="../../src/Sentry.Serilog/Sentry.Serilog.csproj" />`

Is this explicit one here needed?

test/Sentry.Serilog.Tests/Sentry.Serilog.Tests.csproj
```diff
@@ -4,8 +4,18 @@
...
+    <PackageReference Include="Serilog.Settings.Configuration" Version="3.1.0" />
+  </ItemGroup>
+
   <ItemGroup>
     <ProjectReference Include="../../src/Sentry.Serilog/Sentry.Serilog.csproj" />
+    <ProjectReference Include="..\Sentry.Testing\Sentry.Testing.csproj" />
```
bruno-garcia: Just for consistency (see line above):
```suggestion
    <ProjectReference Include="../Sentry.Testing/Sentry.Testing.csproj" />
```

test/Sentry.Serilog.Tests/SentrySerilogConfigurationSettingsIntegrationTests.cs
```diff
@@ -0,0 +1,209 @@
...
+                }
+            }")]
+        public void SentrySerilogOptionsIsConfiguredFromJsonProperly(string when, string json)
+        {
+            Assert.False(string.IsNullOrWhiteSpace(when));
+
+            // This is used as the resulting object / output of the method being tested
+            var actualSentrySerilogOptions = new SentrySerilogOptions();
+
+            var configurationSections = GetConfigurationSectionsFromJson(json);
```
bruno-garcia: Would it be possible to avoid the reflection approach and instead simply "init" the SDK with the Serilog integration and then get a hold of the `SerilogSentryOptions` after the libraries binding values and asserting those?

It's quite a bit of supporting code and reflection (brittle) so I'm not sure it's really worth it.

If this is too hard, perhaps then just programatically calling your new method passing all parameters and making sure they all bind to the SentryOptions property you'd expect. And a second call passing null and verifying nothing was defined. What do you think?

bruno-garcia:
The build issue is a bit cryptic beause of the ZEUS error but that issue doesn't fail the build.

The actual error is:

```
SentrySinkExtensions.cs(161,28): error CS1591: Missing XML comment for publicly visible type or member 'SentrySinkExtensions.ConfigureSentrySerilogOptions(SentrySerilogOptions, bool?, bool?, string, bool?, int?, float?, string, string, string, int?, TimeSpan?, DecompressionMethods?, CompressionLevel?, bool?, bool?, SentryLevel?, bool?, DeduplicateMode?, bool?, LogEventLevel?, LogEventLevel?, IFormatProvider)' [C:\projects\sentry-dotnet\src\Sentry.Serilog\Sentry.Serilog.csproj]
```

I suggest running `./build.sh` or `powershell .\build.ps1` before pushing to remote just to make sure it builds. (it's the `Release` build that enforces some annoying checks like all public members must be documented).

test/Sentry.Serilog.Tests/Sentry.Serilog.Tests.csproj
JimHume: Good eye. This one was odd--visual studio added that reference and yet it was showing up with the yellow triangle/couldn't find it in the solution explorer.

test/Sentry.Serilog.Tests/SentrySerilogConfigurationSettingsIntegrationTests.cs
JimHume: I agree with the brittle nature of this. I struggled to find the best way to approach this--the only way to get at the underlying SentrySerilogOptions object is to perform reflection on the Log.Logger, and then dig through the 'RestrictedSinks' and try to convert said object back into something usable. This is even _more_ brittle than what I had as it would rely more on the Serilog project never changing.

I opted to do as you suggested and went the manually-call-the-method route. I performed a 'zero, one, or many' test including putting the parameters out of order, and also did an 'all parameters' type of test. 

I used a hack to do a deep equals--is there a library that should be used in its place? I couldn't find any similar item and the problem with Assert.Equals in XUnit is that it doesn't deeply compare objects--it seems to only compare references? Or perhaps I'm missing something?

JimHume:
Cheers, thanks for the help, approve & merge @bruno-garcia !

PR #382, opened 2020-03-30
Register when an exception was handled or not

ibm5155:
With this Pr, Sentry.io is going to register if an exception was handled by setting the handled flag in the Mechanism class.

It will not fix the "event handled" internal flag because I'm unsure which parameters I need to inform for that to work.

### DIFF
--- a/src/Sentry/Internal/MainExceptionProcessor.cs
+++ b/src/Sentry/Internal/MainExceptionProcessor.cs
@@ -110,24 +110,21 @@ namespace Sentry.Internal
         {
             Debug.Assert(exception != null);
 
-            Mechanism mechanism = new Mechanism()
-            {
-                Handled = true
-            };
+            var mechanism = new Mechanism();
 
             if (exception.HelpLink != null)
             {
                 mechanism.HelpLink = exception.HelpLink;
             }
-            if (exception.Data?.Contains("Sentry:Handled") == true)
+            if (exception.Data[Mechanism.HandledKey] is bool handled)
             {
-                mechanism.Handled = (exception.Data["Sentry:Handled"] as bool?).Value;
-                exception.Data.Remove("Sentry:Handled");
+                mechanism.Handled = handled;
+                exception.Data.Remove(Mechanism.HandledKey);
             }
-            if (exception.Data?.Contains("Sentry:Mechanism") == true)
+            if (exception.Data[Mechanism.MechanismKey] is string mechanismName )
             {
-                mechanism.Type = exception.Data["Sentry:Mechanism"] as string;
-                exception.Data.Remove("Sentry:Mechanism");
+                mechanism.Type = mechanismName;
+                exception.Data.Remove(Mechanism.MechanismKey);
             }
 
             return mechanism;
--- a/src/Sentry/Internal/MainExceptionProcessor.cs
+++ b/src/Sentry/Internal/MainExceptionProcessor.cs
@@ -110,14 +110,24 @@ namespace Sentry.Internal
         {
             Debug.Assert(exception != null);
 
-            Mechanism mechanism = null;
+            Mechanism mechanism = new Mechanism()
+            {
+                Handled = true
+            };
 
             if (exception.HelpLink != null)
             {
-                mechanism = new Mechanism
-                {
-                    HelpLink = exception.HelpLink
-                };
+                mechanism.HelpLink = exception.HelpLink;
+            }
+            if (exception.Data?.Contains("Sentry:Handled") == true)
+            {
+                mechanism.Handled = (exception.Data["Sentry:Handled"] as bool?).Value;
+                exception.Data.Remove("Sentry:Handled");
+            }
+            if (exception.Data?.Contains("Sentry:Mechanism") == true)
+            {
+                mechanism.Type = exception.Data["Sentry:Mechanism"] as string;
+                exception.Data.Remove("Sentry:Mechanism");
             }
 
             return mechanism;
--- a/src/Sentry/Internal/MainExceptionProcessor.cs
+++ b/src/Sentry/Internal/MainExceptionProcessor.cs
@@ -121,7 +121,7 @@ namespace Sentry.Internal
                 mechanism.Handled = handled;
                 exception.Data.Remove(Mechanism.HandledKey);
             }
-            if (exception.Data[Mechanism.MechanismKey] is string mechanismName )
+            if (exception.Data[Mechanism.MechanismKey] is string mechanismName)
             {
                 mechanism.Type = mechanismName;
                 exception.Data.Remove(Mechanism.MechanismKey);
--- a/test/Sentry.Tests/AppDomainUnhandledExceptionIntegrationTests.cs
+++ b/test/Sentry.Tests/AppDomainUnhandledExceptionIntegrationTests.cs
@@ -6,6 +6,7 @@ using Xunit;
 using System.Linq;
 using System.Threading.Tasks;
 using Sentry.Extensibility;
+using Sentry.Protocol;
 
 namespace Sentry.Tests
 {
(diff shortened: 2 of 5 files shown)


test/Sentry.Tests/AppDomainUnhandledExceptionIntegrationTests.cs
```diff
@@ -56,6 +59,27 @@ public void Handle_TerminatingTrue_DisposesHub()
...
+            Assert.True(exception.Data?.Contains("Sentry:Mechanism"));
+
+
+            var stackTraceFactory = Substitute.For<ISentryStackTraceFactory>();
+            var exceptionProcessor = new MainExceptionProcessor(SentryOptions, () => stackTraceFactory);
+            var @event = new SentryEvent(exception);
+
+            exceptionProcessor.Process(exception, @event);
+            Assert.NotNull(@event.SentryExceptions.ToList().FirstOrDefault(p => p.Mechanism?.Handled == false));
+
```
bruno-garcia:
```suggestion
```

src/Sentry/Internal/MainExceptionProcessor.cs
```diff
@@ -110,14 +110,24 @@ internal static Mechanism GetMechanism(Exception exception)
         {
             Debug.Assert(exception != null);
 
-            Mechanism mechanism = null;
+            Mechanism mechanism = new Mechanism()
```
bruno-garcia:
```suggestion
            var mechanism = new Mechanism()
```

src/Sentry/Internal/MainExceptionProcessor.cs
```diff
@@ -110,14 +110,24 @@ internal static Mechanism GetMechanism(Exception exception)
...
+            {
+                Handled = true
```
bruno-garcia: Can we leave this `null` by default instead of defaulting to `true`?
Not sure if Sentry expects it to be explicitly set to `true` though

src/Sentry/Internal/MainExceptionProcessor.cs
```diff
@@ -110,14 +110,24 @@ internal static Mechanism GetMechanism(Exception exception)
...
+            }
+            if (exception.Data?.Contains("Sentry:Handled") == true)
+            {
+                mechanism.Handled = (exception.Data["Sentry:Handled"] as bool?).Value;
+                exception.Data.Remove("Sentry:Handled");
+            }
+            if (exception.Data?.Contains("Sentry:Mechanism") == true)
+            {
+                mechanism.Type = exception.Data["Sentry:Mechanism"] as string;
+                exception.Data.Remove("Sentry:Mechanism");
```
bruno-garcia: Now we have two lookups in the hashtable:

```
            if (exception.Data?.Contains("Sentry:Handled") == true)
```

And then a second one:

```
                 mechanism.Handled = (exception.Data["Sentry:Handled"] as bool?).Value;
```

Ideally we'd use `TryGetValue` but since this is the old school `IDictionary`, it's not available.
We could at least use the indexers and pattern matching:

```suggestion
        // Throw these at the top of the Mechanism class so it can be used by AppDomainUnhandledExceptionIntegration.cs too
        // Can be used from the unit tests too
        public static readonly string HandledKey = "Sentry:Handled";
        public static readonly string MechanismKey = "Sentry:Mechanism";
        
            if (exception.Data[HandledKey] is bool handled)
            {
                mechanism.Handled = handled;
                exception.Data.Remove(HandledKey);
            }
            if (exception.Data[MechanismKey] is string mechanismName)
            {
                mechanism.Type = mechanismName;
                exception.Data.Remove(MechanismKey);
```

test/Sentry.Tests/AppDomainUnhandledExceptionIntegrationTests.cs
```diff
@@ -56,6 +59,27 @@ public void Handle_TerminatingTrue_DisposesHub()
...
 
+        [Fact]
+        public void Handle_TerminatingTrue_IsHandled()
+        {
+            var sut = _fixture.GetSut();
+            sut.Register(_fixture.Hub, SentryOptions);
+
+            var exception = new Exception();
+            sut.Handle(this, new UnhandledExceptionEventArgs(exception, true));
+            Assert.Equal(exception.Data["Sentry:Handled"] as bool?, (bool?)false);
```
bruno-garcia: The order here's inverted. Expected is on the left. But could just be:
```suggestion
            Assert.False((bool)exception.Data["Sentry:Handled"]);
```

Also now could use the `static readonly reference` from `Mechanism` instead of the string here

test/Sentry.Tests/AppDomainUnhandledExceptionIntegrationTests.cs
```diff
@@ -56,6 +59,27 @@ public void Handle_TerminatingTrue_DisposesHub()
...
+            Assert.True(exception.Data?.Contains("Sentry:Mechanism"));
```
bruno-garcia:
```suggestion
            Assert.True(exception.Data.Contains("Sentry:Mechanism"));
```

test/Sentry.Tests/AppDomainUnhandledExceptionIntegrationTests.cs
```diff
@@ -56,6 +59,27 @@ public void Handle_TerminatingTrue_DisposesHub()
...
+
```
bruno-garcia:
```suggestion
```

test/Sentry.Tests/AppDomainUnhandledExceptionIntegrationTests.cs
```diff
@@ -56,6 +59,27 @@ public void Handle_TerminatingTrue_DisposesHub()
...
+
+            var stackTraceFactory = Substitute.For<ISentryStackTraceFactory>();
+            var exceptionProcessor = new MainExceptionProcessor(SentryOptions, () => stackTraceFactory);
+            var @event = new SentryEvent(exception);
+
+            exceptionProcessor.Process(exception, @event);
+            Assert.NotNull(@event.SentryExceptions.ToList().FirstOrDefault(p => p.Mechanism?.Handled == false));
```
bruno-garcia:
```suggestion
            Assert.NotNull(@event.SentryExceptions.Single(p => p.Mechanism.Handled == false));
```

test/Sentry.Tests/AppDomainUnhandledExceptionIntegrationTests.cs
```diff
@@ -56,6 +59,27 @@ public void Handle_TerminatingTrue_DisposesHub()
             disposableHub.Received(1).Dispose();
         }
 
+        [Fact]
+        public void Handle_TerminatingTrue_IsHandled()
```
bruno-garcia: This tests assert the value Handled is false when app is terminating. Would be nice to have another test that tests it's set to `null` or `true` when this is not captured through the `UnhandledExceptionhandler`.

```suggestion
        public void Handle_TerminatingTrue_IsHandledFalse()
```

src/Sentry/Internal/MainExceptionProcessor.cs
ibm5155: image
The handled inside of Mechanism doesn't seems to affect that "handled vs unhandled", it only point that the exception was handled or not.
That's why I was thinking that by default, an exception is handled, unless it comes from the Unhandled Exception Integration.

test/Sentry.Tests/Internals/MainExceptionProcessorTests.cs
```diff
@@ -125,6 +126,33 @@ public void Process_TwoExceptionsWithData_DataOnEventExtra()
...
+        [Fact]
+        public void Process_ExceptionWithout_Handled()
+        {
+            var sut = _fixture.GetSut();
+            var evt = new SentryEvent();
+            var exp = new Exception();
+
+            sut.Process(exp, evt);
+
+            Assert.NotNull(evt.SentryExceptions.ToList().Single(p => p.Mechanism.Handled == null));
```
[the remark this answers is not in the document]
ibm5155: Assert.Single(evt.SentryExceptions, p => p.Mechanism.Handled == null);
doesn't work in this case, the predicate needs to be a SentryException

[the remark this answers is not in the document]
ibm5155: what about 
`Assert.Single(evt.SentryExceptions.Where(p => p.Mechanism.Handled == null));`
?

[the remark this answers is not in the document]
bruno-garcia: 👍

bruno-garcia:
Thanks @ibm5155 !

PR #393, opened 2020-04-27
Update README.md

pravincarvalho:
This change fixes the NLog error when using this configuration 
"Parameter value not supported on TargetPropertyWithContext"

The attribute should be "layout" and not "value"

Tested with NLog version : 4.7.0
Sentry NLog version : 2.1.1


bruno-garcia:
Thanks @pravincarvalho

PR #405, opened 2020-05-17
Removed dependency on System.Collections.Immutable

snakefoot:
Just for fun

### DIFF
--- a/src/Sentry/Internal/SentryScopeManager.cs
+++ b/src/Sentry/Internal/SentryScopeManager.cs
@@ -1,5 +1,5 @@
 using System;
-using System.Collections.Immutable;
+using System.Collections.Generic;
 using System.Diagnostics;
 using System.Threading;
 using System.Threading.Tasks;
@@ -10,15 +10,15 @@ namespace Sentry.Internal
     internal sealed class SentryScopeManager : IInternalScopeManager, IDisposable
     {
         private readonly SentryOptions _options;
-        private readonly AsyncLocal<ImmutableStack<ValueTuple<Scope, ISentryClient>>> _asyncLocalScope = new AsyncLocal<ImmutableStack<ValueTuple<Scope, ISentryClient>>>();
+        private readonly AsyncLocal<Stack<KeyValuePair<Scope, ISentryClient>>> _asyncLocalScope = new AsyncLocal<Stack<KeyValuePair<Scope, ISentryClient>>>();
 
-        internal ImmutableStack<ValueTuple<Scope, ISentryClient>> ScopeAndClientStack
+        internal Stack<KeyValuePair<Scope, ISentryClient>> ScopeAndClientStack
         {
             get => _asyncLocalScope.Value ?? (_asyncLocalScope.Value = NewStack());
             set => _asyncLocalScope.Value = value;
         }
 
-        private Func<ImmutableStack<ValueTuple<Scope, ISentryClient>>> NewStack { get; }
+        private Func<Stack<KeyValuePair<Scope, ISentryClient>>> NewStack { get; }
 
         public SentryScopeManager(
             SentryOptions options,
@@ -26,23 +26,23 @@ namespace Sentry.Internal
         {
             Debug.Assert(rootClient != null);
             _options = options;
-            NewStack = () => ImmutableStack.Create(new ValueTuple<Scope, ISentryClient>(new Scope(options), rootClient));
+            NewStack = () => new Stack<KeyValuePair<Scope, ISentryClient>>(new [] { new KeyValuePair<Scope, ISentryClient>(new Scope(options), rootClient) });
         }
 
-        public ValueTuple<Scope, ISentryClient> GetCurrent() => ScopeAndClientStack.Peek();
+        public KeyValuePair<Scope, ISentryClient> GetCurrent() => ScopeAndClientStack.Peek();
 
         public void ConfigureScope(Action<Scope> configureScope)
         {
             _options?.DiagnosticLogger?.LogDebug("Configuring the scope.");
             var scope = GetCurrent();
-            configureScope?.Invoke(scope.Item1);
+            configureScope?.Invoke(scope.Key);
         }
 
         public Task ConfigureScopeAsync(Func<Scope, Task> configureScope)
         {
             _options?.DiagnosticLogger?.LogDebug("Configuring the scope asynchronously.");
             var scope = GetCurrent();
-            return configureScope?.Invoke(scope.Item1) ?? Task.CompletedTask;
+            return configureScope?.Invoke(scope.Key) ?? Task.CompletedTask;
         }
 
         public IDisposable PushScope() => PushScope<object>(null);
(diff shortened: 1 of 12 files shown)


src/Sentry/Internal/SentryScopeManager.cs
```diff
@@ -10,65 +10,69 @@ namespace Sentry.Internal
...
                 clonedScope.Apply(state);
             }
+
             var scopeSnapshot = new ScopeSnapshot(_options, currentScopeAndClientStack, this);
             _options?.DiagnosticLogger?.LogDebug("New scope pushed.");
-            ScopeAndClientStack = currentScopeAndClientStack.Push(new ValueTuple<Scope, ISentryClient>(clonedScope, client));
-
+            var newScopeAndClientStack = new Stack<KeyValuePair<Scope, ISentryClient>>(currentScopeAndClientStack.Count + 1);
+            foreach (var item in currentScopeAndClientStack)
+                newScopeAndClientStack.Push(item);
```
bruno-garcia, line 73:
```suggestion
            foreach (var item in currentScopeAndClientStack)
            {
                newScopeAndClientStack.Push(item);
            }
```

src/Sentry/Internal/SentryScopeManager.cs
```diff
@@ -10,65 +10,69 @@ namespace Sentry.Internal
...
             {
                 clonedScope.Apply(state);
             }
+
             var scopeSnapshot = new ScopeSnapshot(_options, currentScopeAndClientStack, this);
             _options?.DiagnosticLogger?.LogDebug("New scope pushed.");
-            ScopeAndClientStack = currentScopeAndClientStack.Push(new ValueTuple<Scope, ISentryClient>(clonedScope, client));
-
+            var newScopeAndClientStack = new Stack<KeyValuePair<Scope, ISentryClient>>(currentScopeAndClientStack.Count + 1);
+            foreach (var item in currentScopeAndClientStack)
```
bruno-garcia, line 72: currentScopeAndClientStack is now a copy to a mutable reference so this would throw if we're iterating over it while it changes

src/Sentry/Internal/SentryScopeManager.cs
```diff
@@ -86,21 +90,21 @@ public void BindClient(ISentryClient client)
             _options?.DiagnosticLogger?.LogDebug("Binding a new client to the current scope.");
 
             var currentScopeAndClientStack = ScopeAndClientStack;
-            currentScopeAndClientStack = currentScopeAndClientStack.Pop(out var top);
-            currentScopeAndClientStack = currentScopeAndClientStack.Push(
-                new ValueTuple<Scope, ISentryClient>(top.Item1, client ?? DisabledHub.Instance));
-            ScopeAndClientStack = currentScopeAndClientStack;
+            var newScopeAndClientStack = new Stack<KeyValuePair<Scope, ISentryClient>>(currentScopeAndClientStack);
```
bruno-garcia, line 93: `currentScopeAndClientStack` is a copy to a mutable type, I'm not sure it's safe to pass it to `Stack`'s constructor. I didn't read the code but assume it'll iterate over to copy the references and that'll throw if we mutate it.

src/Sentry/Sentry.csproj
```diff
@@ -12,13 +12,13 @@
   <ItemGroup Condition="'$(TargetFramework)' == 'net461'">
     <PackageReference Include="Newtonsoft.Json" Version="6.0.8" />
     <Reference Include="System.Web" />
+    <Reference Include="System.Net.Http" />
```
bruno-garcia, line 15: Do we need to add this reference? Not sure I follow

src/Sentry/Sentry.csproj
```diff
@@ -12,13 +12,13 @@
...
   </ItemGroup>
   <ItemGroup Condition="'$(TargetFramework)' == 'netstandard2.0'">
     <PackageReference Include="Newtonsoft.Json" Version="11.0.2" />
   </ItemGroup>
   <ItemGroup>
     <PackageReference Include="Sentry.PlatformAbstractions" Version="1.1.0" />
-    <PackageReference Include="System.Collections.Immutable" Version="1.5.0" />
```
bruno-garcia, line 21: 🚀

src/Sentry/Internal/SentryScopeManager.cs
snakefoot: Not sure I understand. `currentScopeAndClientStack` is never modified. newScopeAndClientStack is a copy off `currentScopeAndClientStack`

src/Sentry/Internal/SentryScopeManager.cs
snakefoot: Not sure I understand. currentScopeAndClientStack is never modified. newScopeAndClientStack is a copy off currentScopeAndClientStack

src/Sentry/Sentry.csproj
snakefoot: No idea. It started to complain after I removed `System.Collections.Immutable`. Also curious if `<Reference Include="System.Web" />` should be removed.

src/Sentry/Internal/SentryScopeManager.cs
snakefoot: Added curly-braces

snakefoot:
I guess one could later replace the Stack with a linked-list. Like this:

https://github.com/aspnet/Logging/blob/master/src/Microsoft.Extensions.Logging.Abstractions/LoggerExternalScopeProvider.cs

But was thinking it would be easier on the eye to go from one stack to another in this PR (Using copy-on-write-semantics)

src/Sentry/Internal/SentryScopeManager.cs
bruno-garcia: It's just: `var currentScopeAndClientStack = ScopeAndClientStack;` so it's just a reference to `ScopeAndClientStack` which can be mutated by anyone calling `SentrySdk.PushScope` for example.

When it was immutable, we didn't have to worry. Now we need make it thread-safe (`ConcurrentStack<T>` would likely do) otherwise the collection can be mutated while iterating over the items in order to copy them.

src/Sentry/Internal/SentryScopeManager.cs
bruno-garcia: replied above

src/Sentry/Sentry.csproj
bruno-garcia: `System.Web` is used because we grab the request data from `HttpContext` when available.
It's odd that `System.Collections.Immutable` would drag `System.net.Http` but since it's just `Reference` it doesn't really matter

src/Sentry/Internal/SentryScopeManager.cs
snakefoot: Still don't understand. No one can modify `ScopeAndClientStack`-property because it is internal. And those calling `SentrySdk.PushScope` makes a clone of `currentScopeAndClientStack` and assigns a whole new object to `ScopeAndClientStack`-property. Think you have to show some code that can make your scenario happen.

bruno-garcia: On a second pass it doesn't look like we're mutating the list anywhere after it's creation so the older behavior is here.

It is racey though if two threads called `PushScope` concurrently. It was already actually while using Immutable collections. since we copy the reference to make a clone, then assign to the property to be made a side effect. So looks like it needs a refactor.

Unrelated to your change. So I'll resolve this issue and I'll do another pass later on the IDE itself. Sorry the delay

snakefoot: `ScopeAndClientStack` uses AsyncLocal (Advanced Thread-Static with async task support), so you have a references for each thread-execution-context. The modifications from one thread will not affect thread-execution-context of other concurrent threads.

bruno-garcia: Yes that's true. This is one thing the .NET SDK diverges with all the other Sentry SDKs. After having 2 other SDKs crashing because of concurrency issues on the Hub I even forgot we had AsyncLocal here. 2 years is a long time.

Issue #402, opened 2020-05-12
[2.1.2-beta4] Calling SerilogConfiguration.WriteTo.Sentry() with no argument is ambiguous


kanadaj:
Please mark the type framework used:

- [ ] ASP.NET MVC
- [ ] ASP.NET Web API (OWIN)
- [x] ASP.NET Core
- [ ] WPF
- [ ] WinForms
- [ ] Xamarin
- [ ] Other:

Please mark the type of the runtime used:

- [ ] .NET Framework
- [ ] Mono
- [x] .NET Core

* Version: 3.1.3

Please mark the NuGet packages used:

- [ ] Sentry
- [x] Sentry.Serilog
- [ ] Sentry.NLog
- [ ] Sentry.Log4Net
- [x] Sentry.Extensions.Logging
- [x] Sentry.AspNetCore

* Version: 2.1.2-beta4

This is an isue introduced in https://github.com/getsentry/sentry-dotnet/pull/380/files#diff-89769915c54d3c8360795be1d31974a6; the method used is ambiguous if no parameter is specified.

bruno-garcia:
Good catch. Breaking change. Needs to be fixed before dropping the `beta` from it.
PR would be highly appreciated!

JimHume:
What is the preferred behaviour when no args are provided?

I can add a new method at the top that only contains the `this LoggerSinkConfiguration loggerConfiguration` parameter and set the defaults on it, but I'd like to clarify what the preferred arguments are. Should they be similar to the first method in the class where it only sets the minimum breadcrumb level and minimum event level?

JimHume:
@bruno-garcia I collapsed the overload I had created to match the method that was already there in 2.1.1

This theoretically upholds backwards compatibility while still allowing for all parameters to be provided.

Please let me know if anything needs updating and I'll make it happen asap.

bruno-garcia:
Sounds good. Thanks @JimHume

bruno-garcia:
Resolved on: https://github.com/getsentry/sentry-dotnet/releases/tag/2.1.2-beta5

PR #406, opened 2020-05-18
Fix/402 sentry serilog extension method collapsing

JimHume:
Collapsing the overloaded methods down to a single extension method for the LoggerSinkConfiguration object from Serilog.

Related issue 402 and 290: 
https://github.com/getsentry/sentry-dotnet/issues/402
https://github.com/getsentry/sentry-dotnet/issues/290

Related to PR 380: https://github.com/getsentry/sentry-dotnet/pull/380/


JimHume:
I just updated the PR by changing the xml-docs a bit to reflect the param order change, and I also updated the example in the xml-docs.

PR #420, opened 2020-05-23
SentryScopeManager - Fixed clone of Stack so it does not reverse order

snakefoot:
Discovered I have introduced a subtle bug in #405

Where enumerating a Stack, then one actual get items in reversed order. So when cloning a stack without thinking then one gets reversed order.

The funny part is that it didn't matter much, since one didn't depend much on the stack-nature, except when validating parent on dispose. The most optimal order to scan is to check original parent first.

But when having reversed the order, then original-parent suddenly was at the bottom of the stack, because of the reverse-order on clone.

Alternative solution is to use `System.Linq.Enumerable.Reverse()`:

`var newScopeAndClientStack = new Stack<KeyValuePair<Scope, ISentryClient>>(currentScopeAndClientStack.Reverse());`

### DIFF
--- a/src/Sentry/Internal/SentryScopeManager.cs
+++ b/src/Sentry/Internal/SentryScopeManager.cs
@@ -10,15 +10,15 @@ namespace Sentry.Internal
     internal sealed class SentryScopeManager : IInternalScopeManager, IDisposable
     {
         private readonly SentryOptions _options;
-        private readonly AsyncLocal<Stack<KeyValuePair<Scope, ISentryClient>>> _asyncLocalScope = new AsyncLocal<Stack<KeyValuePair<Scope, ISentryClient>>>();
+        private readonly AsyncLocal<KeyValuePair<Scope, ISentryClient>[]> _asyncLocalScope = new AsyncLocal<KeyValuePair<Scope, ISentryClient>[]>();
 
-        internal Stack<KeyValuePair<Scope, ISentryClient>> ScopeAndClientStack
+        internal KeyValuePair<Scope, ISentryClient>[] ScopeAndClientStack
         {
             get => _asyncLocalScope.Value ?? (_asyncLocalScope.Value = NewStack());
             set => _asyncLocalScope.Value = value;
         }
 
-        private Func<Stack<KeyValuePair<Scope, ISentryClient>>> NewStack { get; }
+        private Func<KeyValuePair<Scope, ISentryClient>[]> NewStack { get; }
 
         public SentryScopeManager(
             SentryOptions options,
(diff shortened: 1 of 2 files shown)


snakefoot:
Discovered that `BindClient` would be affected by Stack-reverse-order, so added test that fails without this fix.

src/Sentry/Internal/SentryScopeManager.cs
```diff
@@ -94,21 +95,23 @@ public void BindClient(ISentryClient client)
             _options?.DiagnosticLogger?.LogDebug("Binding a new client to the current scope.");
 
             var currentScopeAndClientStack = ScopeAndClientStack;
-            var newScopeAndClientStack = new Stack<KeyValuePair<Scope, ISentryClient>>(currentScopeAndClientStack);
-            var top = newScopeAndClientStack.Pop();
-            newScopeAndClientStack.Push(new KeyValuePair<Scope, ISentryClient>(top.Key, client ?? DisabledHub.Instance));
+            var top = currentScopeAndClientStack[currentScopeAndClientStack.Length - 1];
+
+            var newScopeAndClientStack = new KeyValuePair<Scope, ISentryClient>[currentScopeAndClientStack.Length];
+            Array.Copy(currentScopeAndClientStack, newScopeAndClientStack, currentScopeAndClientStack.Length);
```
bruno-garcia, line 101: nit: Consider using `Buffer.BlockCopy`

test/Sentry.Tests/Internals/SentryScopeManagerTests.cs
```diff
@@ -79,6 +79,24 @@ public void BindClient_Scope_StaysTheSame()
...
+        public void BindClient_ScopeState_StaysTheSame()
+        {
+            var sut = _fixture.GetSut();
+            var currentScope = sut.GetCurrent();
+
+            var scope1 = sut.PushScope(1);
+            var scope2 = sut.PushScope(2);
+            Assert.Equal(2, sut.GetCurrent().Key.Extra["state"]);
+
+            sut.BindClient(Substitute.For<ISentryClient>());
```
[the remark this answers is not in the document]
snakefoot: `BindClient` performs a pop and then a push with the new client, but tries keeping the state. When "Stack" is wrongly reversed, then state becomes wrong.

src/Sentry/Internal/SentryScopeManager.cs
snakefoot: Any reason for this choice ? [StackOverflow seems to prefer Array.Copy](https://stackoverflow.com/questions/1389821/array-copy-vs-buffer-blockcopy)

test/Sentry.Tests/Internals/SentryScopeManagerTests.cs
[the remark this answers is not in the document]
snakefoot: It is `BindClient` that actually makes use of the "Stack"-nature. `PushScope` uses `ScopeSnapshot` to maintain the "Stack", and just uses the `_snapshot` for parent validation when disposing the scope.

src/Sentry/Internal/SentryScopeManager.cs
bruno-garcia: SO "prefers" because the Buffer.BlockCopy is easier to screw up with the parameters. But we'll test/review things correctly here. I prefer it in this case since they do the same and Buffer is slightly faster even though we're not benchmarking anything here so it might be premature optimization but since they do the same thing, I prefer one over the other.
But as I said: `nit` and _consider_. It's your call.

test/Sentry.Tests/Internals/SentryScopeManagerTests.cs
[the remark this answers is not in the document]
bruno-garcia: Makes sense

src/Sentry/Internal/SentryScopeManager.cs
snakefoot: Think `Array.Copy` is good enough for me. If something should be changed, then it is double-stack implementation. It seems a little organic to me. But there is probably a good reason for it.

bruno-garcia:
Thanks @snakefoot

PR #448, opened 2020-06-09
feat: Global Hub support

lucas-zimerman:
This "Feature" (more likely just adding the pattern from the new SDKs) allows you to choose how the SDK will handle Hubs.

By default, the GlobalHudMode is set to false so that it performs the same as before this PR (by that, one "Hub" per threads, Ideally used for applications where multiple users interact).

On the other hand, by setting GlobalHudMode true there'll be a unique Hub for the entire application, this flow is useful for applications where a single user interacts with the application (like an app).

### DIFF
--- a/samples/Sentry.Samples.Console.Customized/Program.cs
+++ b/samples/Sentry.Samples.Console.Customized/Program.cs
@@ -2,6 +2,7 @@ using System;
 using System.Net.Http;
 using System.Reflection;
 using System.Threading.Tasks;
+using System.Xml.Xsl;
 using Sentry;
 using Sentry.Extensibility;
 using Sentry.Protocol;
@@ -77,6 +78,9 @@ internal static class Program
                 return crumb;
             };
 
+            // Ignore exception by its type:
+            o.AddExceptionFilterForType<XsltCompileException>();
+
             // Configure the background worker which sends events to sentry:
             // Wait up to 5 seconds before shutdown while there are events to send.
             o.ShutdownTimeout = TimeSpan.FromSeconds(5);
@@ -105,6 +109,9 @@ internal static class Program
             };
         }))
         {
+            // Ignored by its type due to the setting above
+            SentrySdk.CaptureException(new XsltCompileException());
+
             SentrySdk.AddBreadcrumb(
                 "A 'bad breadcrumb' that will be rejected because of 'BeforeBreadcrumb callback above.'");
 
--- a/samples/Sentry.Samples.Console.Customized/Program.cs
+++ b/samples/Sentry.Samples.Console.Customized/Program.cs
@@ -9,7 +9,7 @@ using Sentry.Protocol;
 
 // One of the ways to set your DSN is via an attribute:
 // It could be set via AssemblyInfo.cs and patched via CI.
-// Other ways are via environment variable, configuration files and explictly via parameter to Init
+// Other ways are via environment variable, configuration files and explicitly via parameter to Init
 [assembly: Dsn(Program.DefaultDsn)]
 // Tracks the release which sent the event and enables more features: https://docs.sentry.io/learn/releases/
 // Much like the DSN above, this is only one of the ways to define the release.
@@ -42,7 +42,7 @@ internal static class Program
             o.AttachStacktrace = true;
 
             // Sentry won't consider code from namespace LibraryX.* as part of the app code and will hide it from the stacktrace by default
-            // To see the lines from non `AppCode`, select `Full`. That'll include non App code like System.*, Microsoft.* and LibraryX.*
+            // To see the lines from non `AppCode`, select `Full`. Will include non App code like System.*, Microsoft.* and LibraryX.*
             o.AddInAppExclude("LibraryX.");
 
             // Before excluding all prefixed 'LibraryX.', any stack trace from a type namespaced 'LibraryX.Core' will be considered InApp.
--- a/src/Sentry.Log4Net/SentryAppender.cs
+++ b/src/Sentry.Log4Net/SentryAppender.cs
@@ -118,7 +118,7 @@ namespace Sentry.Log4Net
                 evt.Environment = Environment;
             }
 
-            Hub.CaptureEvent(evt);
+            _ = Hub.CaptureEvent(evt);
         }
 
         private static IEnumerable<KeyValuePair<string, object>> GetLoggingEventProperties(LoggingEvent loggingEvent)
--- a/src/Sentry/Internal/Hub.cs
+++ b/src/Sentry/Internal/Hub.cs
@@ -56,6 +56,21 @@ namespace Sentry.Internal
             _rootScope = PushScope();
         }
 
+        public IHub Clone()
+        {
+            _options.DiagnosticLogger?.LogDebug("Cloning Hub.");
+
+            var clone = new Hub(_options);
+            foreach (var scopeClient in ScopeManager.ScopeAndClientStack)
+            {
+                var clonedScope = scopeClient.Key.Clone();
+                var cloneItem = new KeyValuePair<Scope, ISentryClient>(clonedScope, scopeClient.Value);
+
+                _ = clone.PushScope(cloneItem);
+            }
+            return clone;
+        }
+
         public void ConfigureScope(Action<Scope> configureScope)
         {
             try
--- a/src/Sentry/Internal/OptionalHub.cs
+++ b/src/Sentry/Internal/OptionalHub.cs
@@ -52,5 +52,7 @@ namespace Sentry.Internal
         public SentryId LastEventId => _hub.LastEventId;
 
         public void Dispose() => (_hub as IDisposable)?.Dispose();
+
+        public IHub Clone() => _hub.Clone();
     }
 }
--- a/src/Sentry/Internal/SentryScopeManager.cs
+++ b/src/Sentry/Internal/SentryScopeManager.cs
@@ -67,7 +67,7 @@ namespace Sentry.Internal
 
             if (state != null)
             {
-                if(state is KeyValuePair<Scope, ISentryClient> newScope)
+                if (state is KeyValuePair<Scope, ISentryClient> newScope)
                 {
                     clonedScope = newScope.Key;
                     client = newScope.Value;
--- a/src/Sentry/Internal/SentryScopeManager.cs
+++ b/src/Sentry/Internal/SentryScopeManager.cs
@@ -10,12 +10,12 @@ namespace Sentry.Internal
     internal sealed class SentryScopeManager : IInternalScopeManager, IDisposable
     {
         private readonly SentryOptions _options;
-        private readonly AsyncLocal<KeyValuePair<Scope, ISentryClient>[]> _asyncLocalScope = new AsyncLocal<KeyValuePair<Scope, ISentryClient>[]>();
+        private KeyValuePair<Scope, ISentryClient>[] _localScope;
 
         internal KeyValuePair<Scope, ISentryClient>[] ScopeAndClientStack
         {
-            get => _asyncLocalScope.Value ?? (_asyncLocalScope.Value = NewStack());
-            set => _asyncLocalScope.Value = value;
+            get => _localScope ?? (_localScope = NewStack());
+            set => _localScope = value;
         }
 
         private Func<KeyValuePair<Scope, ISentryClient>[]> NewStack { get; }
@@ -55,7 +55,7 @@ namespace Sentry.Internal
         {
             var currentScopeAndClientStack = ScopeAndClientStack;
             var scope = currentScopeAndClientStack[currentScopeAndClientStack.Length - 1];
-
+            var client = scope.Value;
             if (scope.Key.Locked)
             {
                 // TODO: keep state on current scope?
@@ -67,7 +67,15 @@ namespace Sentry.Internal
 
             if (state != null)
             {
-                clonedScope.Apply(state);
+                if(state is KeyValuePair<Scope, ISentryClient> newScope)
+                {
+                    clonedScope = newScope.Key;
+                    client = newScope.Value;
+                }
+                else
+                {
+                    clonedScope.Apply(state);
+                }
             }
 
             var scopeSnapshot = new ScopeSnapshot(_options, currentScopeAndClientStack, this);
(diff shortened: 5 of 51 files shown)


src/Sentry/Internal/Hub.cs
```diff
@@ -56,6 +56,11 @@ public Hub(SentryOptions options)
             _rootScope = PushScope();
         }
 
+        public IHub Clone()
+        {
+            _options.DiagnosticLogger?.LogDebug("Attempt to clone a Hub");
+            return new Hub(_options);
+        }
```
bruno-garcia, line 63:
```suggestion
        }

```

src/Sentry/Internal/OptionalHub.cs
```diff
@@ -52,5 +52,10 @@ public OptionalHub(SentryOptions options)
         public SentryId LastEventId => _hub.LastEventId;
 
         public void Dispose() => (_hub as IDisposable)?.Dispose();
+
+        public IHub Clone()
+        {
+            return _hub.Clone();
+        }
```
bruno-garcia, line 59:
```suggestion
        public IHub Clone() => _hub.Clone();
```

src/Sentry/Internal/SentryScopeManager.cs
```diff
@@ -143,7 +143,7 @@ public void Dispose()
         public void Dispose()
         {
             _options?.DiagnosticLogger?.LogDebug("Disposing SentryClient.");
```
bruno-garcia, line 145: Made me realize this: https://github.com/getsentry/sentry-dotnet/pull/449/files

src/Sentry/SentrySdk.cs
```diff
@@ -21,20 +21,68 @@ namespace Sentry
     /// </remarks>
     public static class SentrySdk
     {
-        private static IHub _hub = DisabledHub.Instance;
+
+        /// <summary>
+        ///  Holds Hubs per thread or only mainHub if globalHubMode is disabled.
```
bruno-garcia, line 26:
```suggestion
        /// The async-aware Hub if globalHubMode is disabled.
```

src/Sentry/SentrySdk.cs
```diff
@@ -21,20 +21,68 @@ namespace Sentry
...
+        /// </summary>
+        private static readonly AsyncLocal<IHub> _currentHub = new AsyncLocal<IHub>();
+        /// <summary>
```
bruno-garcia, line 29:
```suggestion

        /// <summary>
```

src/Sentry/SentrySdk.cs
```diff
@@ -21,20 +21,68 @@ namespace Sentry
...
+
+        /// <summary>
+        /// Default value for globalHubMode is false.
+        /// </summary>
+        private const bool _globalHudDefaultMode = false;
+
+        /// <summary>
+        /// whether to use a single (global) Hub as opposed to one per thread.
+        /// </summary>
+        private static bool _globalHudMode = _globalHudDefaultMode;
```
bruno-garcia, line 42: In case other threads are calling methods from this class before `Init` is called, and value is cached

```suggestion
        private static volatile bool _globalHubMode = _globalHubDefaultMode;
```

src/Sentry/SentrySdk.cs
```diff
@@ -21,20 +21,68 @@ namespace Sentry
...
 
         /// <summary>
         /// Last event id recorded in the current scope
         /// </summary>
-        public static SentryId LastEventId { [DebuggerStepThrough] get => _hub.LastEventId; }
+        public static SentryId LastEventId { [DebuggerStepThrough] get => GetCurrentHub().LastEventId; }
+
+        /// <summary>
+        /// Returns the current (threads) hub, if none, clones the mainHub and returns it.
+        /// </summary>
```
bruno-garcia, line 51:
```suggestion
        /// <summary>
        /// Returns the current hub.
        /// </summary>
        /// <remarks>
        /// This is an async-aware thread-local Hub, unless global Hub mode is enabled.
        /// </remarks>
```

src/Sentry/SentrySdk.cs
```diff
@@ -21,20 +21,68 @@ namespace Sentry
...
+        public static IDisposable Init() => Init(DsnLocator.FindDsnStringOrDisable(), _globalHudDefaultMode);
 
         /// <summary>
         /// Initializes the SDK while attempting to locate the DSN
         /// </summary>
         /// <remarks>
         /// If the DSN is not found, the SDK will not change state.
         /// </remarks>
-        public static IDisposable Init() => Init(DsnLocator.FindDsnStringOrDisable());
+        public static IDisposable Init(bool globalHudMode = _globalHudDefaultMode) => Init(DsnLocator.FindDsnStringOrDisable(), globalHudMode);
```
bruno-garcia, line 85: Please add an overload instead of default parameters here.

Also, annotate it with `[EditorBrowsable(EditorBrowsableState.Never)]` in an attempt to reduce adoption of direct dependency on this behavior without some wrapping package like Sentry.Xamarin

src/Sentry/SentrySdk.cs
```diff
@@ -47,36 +95,52 @@ public static class SentrySdk
         public static IDisposable Init(string dsn)
             => string.IsNullOrWhiteSpace(dsn)
                 ? DisabledHub.Instance
-                : Init(c => c.Dsn = new Dsn(dsn));
+                : Init(c => c.Dsn = new Dsn(dsn), _globalHudDefaultMode);
+
+        /// <summary>
+        /// Initializes the SDK with the specified DSN
```
bruno-garcia, line 101:
```suggestion
        /// Initializes the SDK with the specified DSN.
```

src/Sentry/SentrySdk.cs
```diff
@@ -47,36 +95,52 @@ public static class SentrySdk
...
+        /// </summary>
+        /// <remarks>
+        /// An empty string is interpreted as a disabled SDK
```
bruno-garcia, line 104:
```suggestion
        /// An empty string is interpreted as a disabled SDK.
```

src/Sentry/SentrySdk.cs
```diff
@@ -47,36 +95,52 @@ public static class SentrySdk
...
+        /// </remarks>
+        /// <seealso href="https://docs.sentry.io/clientdev/overview/#usage-for-end-users"/>
+        /// <param name="dsn">The dsn</param>
+        /// <param name="globalHudMode">whether to use a single (global) Hub as opposed to one per thread.</param>
+        public static IDisposable Init(string dsn, bool globalHudMode = _globalHudDefaultMode)
```
bruno-garcia, line 109: On a second thought I believe it's best if we add this to the `SentryOptions` class and don't add any overloads here to `Init`.

src/Sentry/SentrySdk.cs
```diff
@@ -115,7 +185,7 @@ internal static IDisposable UseHub(IHub hub)
         [EditorBrowsable(EditorBrowsableState.Never)]
         public static void Close()
         {
-            var oldHub = Interlocked.Exchange(ref _hub, DisabledHub.Instance);
+            var oldHub = Interlocked.Exchange(ref _mainHud, DisabledHub.Instance);
```
bruno-garcia, line 188: Now we're only disposing the main hub and not the other ones closed for each thread.

src/Sentry/SentrySdk.cs
```diff
@@ -147,21 +217,36 @@ public void Dispose()
...
         [DebuggerStepThrough]
-        public static IDisposable PushScope<TState>(TState state) => _hub.PushScope(state);
+        public static IDisposable PushScope<TState>(TState state)
+        {
+            // pushScope is no-op in global hub mode
+            if (!_globalHudMode)
+            {
+                return GetCurrentHub().PushScope(state);
+            }
+            return null;
```
bruno-garcia, line 227: Returning `null` for `IDisposable` is a bad idea. `using(SentrySdk.PushScope)` will throw.

src/Sentry/SentrySdk.cs
```diff
@@ -87,13 +151,19 @@ public static IDisposable Init(SentryOptions options)
                 }
                 options.Dsn = dsn;
             }
-
+            _globalHudMode = globalHudMode;
```
bruno-garcia, line 154: here we'd check the `Hub` mode from the options instead

src/Sentry/SentrySdk.cs
```diff
@@ -21,20 +21,68 @@ namespace Sentry
...
+        internal static IHub GetCurrentHub()
+        {
+            if (_globalHudMode)
+            {
+                return _mainHud;
+            }
+            IHub hub = _currentHub.Value;
+            if (hub == null)
+            {
+                lock (_mainHud)
```
bruno-garcia, line 62: We don't even need a lock here since each thread will have it's own `hub` here and `_mainHud` can be cloned in parallel

src/Sentry.Log4Net/SentryAppender.cs
```diff
@@ -44,18 +44,18 @@ public class SentryAppender : AppenderSkeleton
...
         { }
+
         internal SentryAppender(
             Func<string, IDisposable> initAction,
             IHub hubGetter)
         {
             Debug.Assert(initAction != null);
             Debug.Assert(hubGetter != null);
 
-            _initAction = initAction.Invoke;
```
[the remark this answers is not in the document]
lucas-zimerman: Indeed I noticed that and I reverted it

src/Sentry/SentrySdk.cs
lucas-zimerman: when a thread is closed it calls the Dispose from the Hubs I suppose.
I could set _currentHub.Value = null but I'm unsure about the effect of it,

src/Sentry/Internal/Hub.cs
```diff
@@ -56,6 +56,12 @@ public Hub(SentryOptions options)
             _rootScope = PushScope();
         }
 
+        public IHub Clone()
+        {
+            _options.DiagnosticLogger?.LogDebug("Attempt to clone a Hub");
+            return new Hub(_options);
```
bruno-garcia, line 62: Is this a `TODO`? The Hub isnt' being cloned here, it's just creating a new one with the same options

src/Sentry/Internal/Hub.cs
```diff
@@ -56,6 +56,12 @@ public Hub(SentryOptions options)
             _rootScope = PushScope();
         }
 
+        public IHub Clone()
+        {
+            _options.DiagnosticLogger?.LogDebug("Attempt to clone a Hub");
```
bruno-garcia, line 61:
```suggestion
            _options.DiagnosticLogger?.LogDebug("Cloning Hub.");
```

src/Sentry/Internal/Hub.cs
lucas-zimerman: https://github.com/getsentry/sentry-android/blob/master/sentry-core/src/main/java/io/sentry/core/Hub.java#L589

it's what the java code does, I didn't add the rest because there are no stacks from what I looked.

bruno-garcia: The stack is in the ScopeManager. That's where the state (tags, etc) are kept and Clone needs to copy them over to the new Hub

src/Sentry/SentrySdk.cs
lucas-zimerman: @bruno-garcia  The correct way would be to not return anything (changing from IDisposable to void) but afraid that will result in some breaking change.

lucas-zimerman:
```
    internal class FalseDisposable : IDisposable
    {

        public void Dispose() { }
    }

```
creating a static class from FalseDisposable inside of SentrySDK and only returning it on Push/Pop scope only if the global mode is true, seems to be an "OK" solution to avoid breaking changes (note that if not using global mode as most .Net application does, it'll work the same previously)

bruno-garcia: yeah we use this approach elsewhere with a `private class`, I can't think of a better way, sadly

lucas-zimerman:
@bruno-garcia The only remaining thing to change is the Global Hud Mode to be stored inside SentryOptions, should I change that?

bruno-garcia:
Sure, thanks @lucas-zimerman

lucas-zimerman:
@bruno-garcia  GlobalHubmode moved to SentryOptions

lucas-zimerman:
Synced with https://github.com/getsentry/sentry-dotnet/tree/lucas-zimerman-feature/HubTypeSelector with Bruno changes so this branch doesn't become too outdated.

Further analysis and tests are required to assure events are not going to be lost when GlobalHubMode is disabled and also the Close Method doesn't work properly in the same situation.

samples/Sentry.Samples.Console.Customized/Program.cs
```diff
@@ -105,6 +109,9 @@ private static async Task Main(string[] args)
             };
         }))
         {
+            // Ignored by its type due to the setting above
```
bruno-garcia, line 112: A bunch of changes that are already in `main` are showing up here somehow.

lucas-zimerman:
Moving to draft so it doesn't compile every time the main branch gets updated.

bruno-garcia:
Replaced by #1124

bruno-garcia:
We solved "Global Mode" on version **3.8.1**.

It makes it so that any call to `SentrySdk.SetTag` or `AddBreadcrumb` or anything else that you do through `SentrySdk.ConfigureScope` mutates a single, static `Scope` object that affects **all threads of the app**. This is useful for Desktop and Mobile apps when you have a single user session and you add context to Sentry anywhere in the app and want that to be included if it crashes in any other thread.

You can opt-in to this mode through:

`options.IsGlobalModeEnabled = true`

To clarify: This options doesn't make sense in any sort of web server where each individual request takes a separate scope (You want each `SetTag("url", ...)` to take a different request URL, headers etc). Nothing needs to be done for server apps, it continues to work as it did in the past.

We plan to ship `Sentry.Wpf` and `Sentry.WinForms` in the future that would already flip `IsGlobalModeEnabled = true` for you. Until then we'll add to the docs that you can do that if you prefer to have a single mutable state.

PR #450, opened 2020-06-13
NLog SentryTarget - NLogDiagnosticLogger for writing to NLog InternalLogger

snakefoot:


src/Sentry.NLog/NLogDiagnosticLogger.cs
```diff
@@ -0,0 +1,61 @@
...
+            if (!InternalLogger.LogToConsole || !(extraLogger is ConsoleDiagnosticLogger))
+            {
+                _extraLogger = extraLogger;
+            }
+        }
+
+        public bool IsEnabled(SentryLevel level)
+        {
+            if (_extraLogger?.IsEnabled(level)==true)
+                return true;
```
bruno-garcia, line 25:
```suggestion
            if (_extraLogger?.IsEnabled(level) == true)
            {
                return true;
            }
```

src/Sentry.NLog/NLogDiagnosticLogger.cs
```diff
@@ -0,0 +1,61 @@
...
+                case SentryLevel.Error: return InternalLogger.IsErrorEnabled;
+                case SentryLevel.Warning: return InternalLogger.IsWarnEnabled;
+                case SentryLevel.Info: return InternalLogger.IsInfoEnabled;
+                default: return InternalLogger.IsDebugEnabled;
+            }
+        }
+
+        public void Log(SentryLevel logLevel, string message, Exception exception = null, params object[] args)
+        {
+            switch (logLevel)
```
bruno-garcia, line 39: I believe you need to guard again for the log level here. Now one guard for each logger:

Before writing to `extraLogger` do the:

```
           if (_extraLogger?.IsEnabled(level) == true)
            {
                return true;
            }
```

Before writing to the `InternalLogger`, do the check:

```
            switch (level)
            {
                case SentryLevel.Fatal: return InternalLogger.IsFatalEnabled;
                case SentryLevel.Error: return InternalLogger.IsErrorEnabled;
                case SentryLevel.Warning: return InternalLogger.IsWarnEnabled;
                case SentryLevel.Info: return InternalLogger.IsInfoEnabled;
                default: return InternalLogger.IsDebugEnabled;
            }
```

That assumes they can have different log levels enabled and if either returns true on `IsEnabled` we'd get a call here to `Log`

src/Sentry.NLog/SentryTarget.cs
```diff
@@ -438,14 +448,28 @@ private User GetUser(LogEventInfo logEvent)
...
+            var user = new User
             {
                 Email = User.Email?.Render(logEvent),
                 Id = User.Id?.Render(logEvent),
                 IpAddress = User.IpAddress?.Render(logEvent),
                 Username = User.Username?.Render(logEvent),
-                Other = User.Other.ToDictionary(a => a.Name, b => b.Layout.Render(logEvent)),
             };
+
+            if (User.Other?.Count > 0)
```
bruno-garcia, line 459: This change is unrelated to the logger changes, right?
Ideally in the future it would land on a separate PR please

test/Sentry.NLog.Tests/SentryTargetTests.cs
```diff
@@ -488,17 +488,27 @@ public async Task LogManager_WhenFlushCalled_CallsSentryFlushAsync()
...
+                InternalLogger.LogLevel = LogLevel.Debug;
+
+                _ = _fixture.GetLoggerFactory();
+
+                var logoutput = logwriter.ToString();
+                Assert.Contains("Init was called but no DSN was provided nor located. Sentry SDK will be disabled.", logoutput);
+            }
+            finally
+            {
+                _fixture.Options.Debug = false;
```
bruno-garcia, line 508: There's a new instance of `_fixture` per test so there's no need to reset values.

test/Sentry.NLog.Tests/SentryTargetTests.cs
```diff
@@ -769,15 +779,15 @@ public void GetUserFromLayouts_PropertiesMapped()
             var sentryTarget = factory.Configuration.FindTargetByName<SentryTarget>("sentry");
             sentryTarget.User = new SentryNLogUser
             {
-                Username = "${logger:shortName=true}"
+                Username = "${logger:shortName=true}",
             };
+            sentryTarget.User.Other.Add(new TargetPropertyWithContext("Gender", "${event-properties:Gender}") { IncludeEmptyValue = false });
```
bruno-garcia, line 784: Do we have a test that verifies a value does end up on the Sentry User?

snakefoot: Nope. Just wanted to code-coverage to be complete.

test/Sentry.NLog.Tests/SentryTargetTests.cs
snakefoot: Ok will remove.

src/Sentry.NLog/SentryTarget.cs
snakefoot: I just reacted to warning about unnecessary allocations and delegate capture.

src/Sentry.NLog/NLogDiagnosticLogger.cs
snakefoot: The NLog  InternalLogger-methods performs implicit LogLevel-filtering-check, so no need to do it twice.

src/Sentry.NLog/NLogDiagnosticLogger.cs
snakefoot: Fixed

test/Sentry.NLog.Tests/SentryTargetTests.cs
snakefoot: Now changed the test, so it actually test output value is generated.

src/Sentry.NLog/SentryTarget.cs
snakefoot: But now you have also received a unit-test that verifies the Other-output actually works :)

PR #473, opened 2020-07-24
ref: dotnet-format

bruno-garcia:
running `dotnet format` on this repo.

It would need to made part of CI first.


bruno-garcia:
Done through #1166

PR #476, opened 2020-07-31
show an example of snapshot testing using Verify

SimonCropp:
https://github.com/VerifyTests/Verify

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,3 +21,4 @@ docs/api/*
 docs/_site/*
 docs/docfx/*
 docs/docfx.zip
+*.received.*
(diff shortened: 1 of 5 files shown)


bruno-garcia:
Nice snapshot library @SimonCropp.
I'll keep this open until I get some time to proper consider this.
I'd love to get this in and will probably do so when I get time to lift the dep to the test/Directory.Build.props and use it in other tests asserting all of the object's state.

SimonCropp:
I did a purge of forked repos. hence this was closed. let me know if u want it resubmitted. or u want to get on a call to discuss

bruno-garcia:
Thanks @SimonCropp.
I think the example in the PR is a great starting point for us.

PR #480, opened 2020-08-12
Update Program.cs

AndreasLangberg:
Tiny grammar fix in example. Might be overkill for a single PR, but it was haunting my dreams


bruno-garcia:
Good fix! Thanks @AndreasLangberg

PR #482, opened 2020-08-19
fix: Apply default tags in ASP.NET classic

bruno-garcia:


bruno-garcia:
This was just relevant for MEL

PR #500, opened 2020-09-05
net5.0

bruno-garcia:
Getting a head start with .NET 5 compatibility

### DIFF
--- a/src/Sentry.NLog/Sentry.NLog.csproj
+++ b/src/Sentry.NLog/Sentry.NLog.csproj
@@ -1,14 +1,14 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFrameworks>netstandard2.0;net461</TargetFrameworks>
+    <TargetFrameworks>net5.0;netstandard2.0;net461</TargetFrameworks>
     <PackageTags>$(PackageTags);Logging;NLog</PackageTags>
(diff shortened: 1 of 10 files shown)


src/Sentry.NLog/Sentry.NLog.csproj
```diff
@@ -1,14 +1,14 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFrameworks>netstandard2.0;net461</TargetFrameworks>
+    <TargetFrameworks>net5.0;netstandard2.0;net461</TargetFrameworks>
```
bruno-garcia, line 4: Target probably not needed

src/Sentry/Sentry.csproj
```diff
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFrameworks>netstandard2.0;net461</TargetFrameworks>
+    <TargetFrameworks>net5.0;netstandard2.0;net461</TargetFrameworks>
```
bruno-garcia, line 4: Probably don't need to target here too. Just let it resolve to ns2.0.
Unless we're changing stuff like using System.Json (while keeping the old targets as they are which would probably surprise some if they retarget their app and without changing this dependency get some different dependencies).

bruno-garcia:
Superseded by #583

PR #507, opened 2020-09-14
ref: new dsn for test org

bruno-garcia:
New test org for SDK development:

image

### DIFF
--- a/samples/Sentry.Samples.AspNetCore.Basic/Properties/launchSettings.json
+++ b/samples/Sentry.Samples.AspNetCore.Basic/Properties/launchSettings.json
@@ -13,7 +13,7 @@
       "launchBrowser": true,
       "launchUrl": "throw",
       "environmentVariables": {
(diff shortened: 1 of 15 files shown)


Tyrrrz (approved):
Looks good to me. Don't know if that's all the places that the DSN is used, but I'll trust you :)

Also, looking at `samples/Sentry.Samples.AspNetCore.Basic/Properties/launchSettings.json`, I would probably move environment variables to `appsettings.json`. I'm not sure if `launchSettings.json` is read by the IDE or by the framework, but `appsettings.json` seems like a better place for it. But it's my opinion and given that it's a sample, it probably doesn't matter.

bruno-garcia:
@Tyrrrz In that particular example I wanted to show that environment variables are also supported. 
`dotnet run` and the IDEs supporting loading env vars from `launchSettings.json` hence the sample there.

I just did a string replace on the old dsn. I hope I didn't miss any :)

bruno-garcia:
btw, There are other samples using `appsettings.json`

Tyrrrz:
Makes sense.

PR #508, opened 2020-09-15
Clean up solution file

Tyrrrz:
Cleaned up solution file by removing `global.json` which was deleted and also renaming `1 - Solution Items` to `.root`. I understand you put the `1` to make it appear first, so I replaced it with a dot which is a bit more conventional (i.e. in dot files). I think this looks a bit better. If you don't like it, I will revert it.


bruno-garcia (approved):
LGTM

Issue #504, opened 2020-09-11
Null annotate the API


bruno-garcia:
Blocked by #503
Follow up with: https://github.com/getsentry/sentry-dotnet/pull/229

Tyrrrz:
Done.

Issue #503, opened 2020-09-11
Migrate to C# 8


bruno-garcia:

bruno-garcia:
solution targets C# 8 (latest).
Thanks @Tyrrrz

PR #509, opened 2020-09-15
Add nullable reference types support (Sentry, Sentry.Protocol)

Tyrrrz:
**Work in progress**

### Some notes:

___

There are two sides to nullable reference types. One is the metadata we provide to users as part of the published library, which helps them correct nullability issues in their integration with Sentry, the other is for static analysis within Sentry itself. While the former can be achieved on all .NET frameworks with backporting (using NuGet package [Nullable](https://www.nuget.org/packages/Nullable)), the latter can only be done with .NET Core 3.0+ or .NET Std 2.1+. The reason for that is, because earlier versions of the framework are not annotated with NRT attributes, enabling warnings on them will result in huge amount of false positives. So as a solution, the `<Nullable>` project property is set to `enable` only when targeting `netstandard2.1`, while on other targets it's set to `annotations` (which allows us to annotate, but doesn't produce warnings for our code). Since `netstandard2.1` target wasn't there, I added it. This is the framework we should be using when developing in the IDE, so that we get proper nullability warnings/insights. As long as that target is also built on CI, we will get warnings there as well. The added bonus of this approach is that warnings are not duplicated per target either (otherwise we'd have one nullability warning per framework).

___

Note that NRTs works at compile time and not runtime and since it's not actually part of the type system but rather a mere annotation, it doesn't protect against `NullReferenceException`s. That said, I would suggest not to check for null and rely on static analysis to tell users when they're wrong. Thanks to backporting, NRTs should be available for users targeting older versions of the framework too. Ideally, we should also either hide nullable parameters behind overloads or default them to null so it's a bit more clear.

___

Some methods were easy enough to analyze in terms of nullability. Some were more difficult. Because there methods that rely a lot on mutability, it was hard to determine whether the value can be null or not at a specific point. In most cases I just marked such instances as nullable. It revealed many places where nulls were not handled. I added some TODOs for these.

___

I've noticed some types are designed as data contracts with public get/set properties. I see that some of these properties are meant to be required, but since it cannot be enforced at a type system level without a constructor (until C# 9 and records), I marked them as nullable. That's because you simply can't know whether they've been set to a valid value or not. Code that interacts with those types needs to be aware of that too.

___

There were some instances of defensive programming in the code base that checked method parameters for null even when they weren't really expected to be null. I marked those instances as nullable to be safe, but we should ideally avoid defensive programming and let type system and static analysis take care of such situations for us. In the future we should consider refactoring this.

___

I found that some method overloads can be simplified by using default parameter value of null where null was expected anyway. Unfortunately, removing overloads is a breaking change so I kept them and added a corresponding TODO to remember for the future.

___

As I was annotating stuff, I also cleaned up some XML docs by adding a period in the end where applicable (accessibility) and removing empty nodes (e.g. `<returns></returns>`).

___

With the advent of .NET 5 we should probably consider dropping runtime-specific target frameworks (i.e. `net461` etc) and only keep .NET Standard.

___

Closes #229
Related to #503 
Related to #504

### DIFF
--- a/src/Sentry.Protocol/BaseScopeExtensions.cs
+++ b/src/Sentry.Protocol/BaseScopeExtensions.cs
@@ -138,7 +138,7 @@ namespace Sentry
                 }
             }
 
-            var breadcrumbs = (ConcurrentQueue<Breadcrumb>) scope.Breadcrumbs;
+            var breadcrumbs = (ConcurrentQueue<Breadcrumb>)scope.Breadcrumbs;
 
             var overflow = breadcrumbs.Count - (scope.ScopeOptions?.MaxBreadcrumbs
                                                 ?? Constants.DefaultMaxBreadcrumbs) + 1;
@@ -165,7 +165,7 @@ namespace Sentry
         /// <param name="key">The key.</param>
         /// <param name="value">The value.</param>
         public static void SetExtra(this BaseScope scope, string key, object value)
-            => ((ConcurrentDictionary<string, object>) scope.Extra).AddOrUpdate(key, value, (s, o) => value);
+            => ((ConcurrentDictionary<string, object>)scope.Extra).AddOrUpdate(key, value, (s, o) => value);
 
         /// <summary>
         /// Sets the extra key-value pairs to the <see cref="BaseScope"/>.
@@ -174,7 +174,7 @@ namespace Sentry
         /// <param name="values">The values.</param>
         public static void SetExtras(this BaseScope scope, IEnumerable<KeyValuePair<string, object>> values)
         {
-            var extra = (ConcurrentDictionary<string, object>) scope.Extra;
+            var extra = (ConcurrentDictionary<string, object>)scope.Extra;
             foreach (var keyValuePair in values)
             {
                 _ = extra.AddOrUpdate(keyValuePair.Key, keyValuePair.Value, (_, o) => keyValuePair.Value);
@@ -197,7 +197,7 @@ namespace Sentry
         /// <param name="tags"></param>
         public static void SetTags(this BaseScope scope, IEnumerable<KeyValuePair<string, string>> tags)
         {
-            var internalTags = (ConcurrentDictionary<string, string>) scope.Tags;
+            var internalTags = (ConcurrentDictionary<string, string>)scope.Tags;
             foreach (var keyValuePair in tags)
             {
                 _ = internalTags.AddOrUpdate(keyValuePair.Key, keyValuePair.Value, (s, o) => keyValuePair.Value);
@@ -248,7 +248,7 @@ namespace Sentry
             {
                 foreach (var extra in from.Extra)
                 {
-                    _ = ((ConcurrentDictionary<string, object>) to.Extra).TryAdd(extra.Key, extra.Value);
+                    _ = ((ConcurrentDictionary<string, object>)to.Extra).TryAdd(extra.Key, extra.Value);
                 }
             }
 
@@ -256,7 +256,7 @@ namespace Sentry
             {
                 foreach (var tag in from.Tags)
                 {
-                    _ = ((ConcurrentDictionary<string, string>) to.Tags).TryAdd(tag.Key, tag.Value);
+                    _ = ((ConcurrentDictionary<string, string>)to.Tags).TryAdd(tag.Key, tag.Value);
                 }
             }
 
--- a/src/Sentry.Protocol/BaseScopeExtensions.cs
+++ b/src/Sentry.Protocol/BaseScopeExtensions.cs
@@ -39,7 +39,7 @@ namespace Sentry
         /// <param name="level">The level.</param>
         public static void AddBreadcrumb(
             this BaseScope scope,
-            string? message,
+            string message,
             string? category,
             string? type,
             in (string, string)? dataPair = null,
@@ -83,7 +83,7 @@ namespace Sentry
         /// <param name="level">The level.</param>
         public static void AddBreadcrumb(
             this BaseScope scope,
-            string? message,
+            string message,
             string? category = null,
             string? type = null,
             Dictionary<string, string>? data = null,
@@ -122,7 +122,7 @@ namespace Sentry
         public static void AddBreadcrumb(
             this BaseScope scope,
             DateTimeOffset? timestamp,
-            string? message,
+            string message,
             string? category = null,
             string? type = null,
             IReadOnlyDictionary<string, string>? data = null,
--- a/src/Sentry.Protocol/BaseScopeExtensions.cs
+++ b/src/Sentry.Protocol/BaseScopeExtensions.cs
@@ -29,23 +29,24 @@ namespace Sentry
 
 #if HAS_VALUE_TUPLE
         /// <summary>
-        /// Adds a breadcrumb to the scope
+        /// Adds a breadcrumb to the scope.
         /// </summary>
         /// <param name="scope">The scope.</param>
         /// <param name="message">The message.</param>
-        /// <param name="type">The type.</param>
         /// <param name="category">The category.</param>
+        /// <param name="type">The type.</param>
         /// <param name="dataPair">The data key-value pair.</param>
         /// <param name="level">The level.</param>
         public static void AddBreadcrumb(
-                    this BaseScope scope,
-                    string message,
-                    string category,
-                    string type,
-                    in (string, string)? dataPair = null,
-                    BreadcrumbLevel level = default)
+            this BaseScope scope,
+            string? message,
+            string? category,
+            string? type,
+            in (string, string)? dataPair = null,
+            BreadcrumbLevel level = default)
         {
-            Dictionary<string, string> data = null;
+            Dictionary<string, string>? data = null;
+
             if (dataPair != null)
             {
                 data = new Dictionary<string, string>
@@ -55,12 +56,12 @@ namespace Sentry
             }
 
             scope.AddBreadcrumb(
-                timestamp: null,
-                message: message,
-                category: category,
-                type: type,
-                data: data,
-                level: level);
+                null,
+                message,
+                category,
+                type,
+                data,
+                level);
         }
 #endif
 
@@ -75,23 +76,23 @@ namespace Sentry
         /// <param name="level">The level.</param>
         public static void AddBreadcrumb(
             this BaseScope scope,
-            string message,
-            string category = null,
-            string type = null,
-            Dictionary<string, string> data = null,
+            string? message,
+            string? category = null,
+            string? type = null,
+            Dictionary<string, string>? data = null,
             BreadcrumbLevel level = default)
         {
             scope.AddBreadcrumb(
-                timestamp: null,
-                message: message,
-                category: category,
-                type: type,
-                data: data,
-                level: level);
+                null,
+                message,
+                category,
+                type,
+                data,
+                level);
         }
 
         /// <summary>
-        /// Adds a breadcrumb to the scope
+        /// Adds a breadcrumb to the scope.
         /// </summary>
         /// <remarks>
         /// This overload is used for testing.
@@ -107,23 +108,23 @@ namespace Sentry
         public static void AddBreadcrumb(
             this BaseScope scope,
             DateTimeOffset? timestamp,
-            string message,
-            string category = null,
-            string type = null,
-            IReadOnlyDictionary<string, string> data = null,
+            string? message,
+            string? category = null,
+            string? type = null,
+            IReadOnlyDictionary<string, string>? data = null,
             BreadcrumbLevel level = default)
             => scope.AddBreadcrumb(new Breadcrumb(
-                timestamp: timestamp,
-                message: message,
-                type: type,
-                data: data,
-                category: category,
-                level: level));
+                timestamp,
+                message,
+                type,
+                data,
+                category,
+                level));
 
         /// <summary>
-        /// Adds a breadcrumb to the <see cref="BaseScope"/>
+        /// Adds a breadcrumb to the <see cref="BaseScope"/>.
         /// </summary>
-        /// <param name="scope">Scope</param>
+        /// <param name="scope">Scope.</param>
         /// <param name="breadcrumb">The breadcrumb.</param>
         internal static void AddBreadcrumb(this BaseScope scope, Breadcrumb breadcrumb)
         {
@@ -132,9 +133,9 @@ namespace Sentry
                 return;
             }
 
-            if (scope.ScopeOptions?.BeforeBreadcrumb is Func<Breadcrumb, Breadcrumb> callback)
+            if (scope.ScopeOptions?.BeforeBreadcrumb != null)
             {
-                breadcrumb = callback(breadcrumb);
+                breadcrumb = scope.ScopeOptions.BeforeBreadcrumb(breadcrumb);
 
                 if (breadcrumb == null)
                 {
@@ -142,7 +143,7 @@ namespace Sentry
                 }
             }
 
-            var breadcrumbs = (ConcurrentQueue<Breadcrumb>)scope.Breadcrumbs;
+            var breadcrumbs = (ConcurrentQueue<Breadcrumb>) scope.Breadcrumbs;
 
             var overflow = breadcrumbs.Count - (scope.ScopeOptions?.MaxBreadcrumbs
                                                 ?? Constants.DefaultMaxBreadcrumbs) + 1;
@@ -155,7 +156,7 @@ namespace Sentry
         }
 
         /// <summary>
-        /// Sets the fingerprint to the <see cref="BaseScope"/>
+        /// Sets the fingerprint to the <see cref="BaseScope"/>.
         /// </summary>
         /// <param name="scope">The scope.</param>
         /// <param name="fingerprint">The fingerprint.</param>
@@ -163,7 +164,7 @@ namespace Sentry
             => scope.InternalFingerprint = fingerprint;
 
         /// <summary>
-        /// Sets the extra key-value to the <see cref="BaseScope"/>
+        /// Sets the extra key-value to the <see cref="BaseScope"/>.
         /// </summary>
         /// <param name="scope">The scope.</param>
         /// <param name="key">The key.</param>
@@ -172,7 +173,7 @@ namespace Sentry
             => ((ConcurrentDictionary<string, object>)scope.Extra).AddOrUpdate(key, value, (s, o) => value);
 
         /// <summary>
-        /// Sets the extra key-value pairs to the <see cref="BaseScope"/>
+        /// Sets the extra key-value pairs to the <see cref="BaseScope"/>.
         /// </summary>
         /// <param name="scope">The scope.</param>
         /// <param name="values">The values.</param>
@@ -181,12 +182,12 @@ namespace Sentry
             var extra = (ConcurrentDictionary<string, object>)scope.Extra;
             foreach (var keyValuePair in values)
             {
-                _ = extra.AddOrUpdate(keyValuePair.Key, keyValuePair.Value, (s, o) => keyValuePair.Value);
+                _ = extra.AddOrUpdate(keyValuePair.Key, keyValuePair.Value, (_, o) => keyValuePair.Value);
             }
         }
 
         /// <summary>
-        /// Sets the tag to the <see cref="BaseScope"/>
+        /// Sets the tag to the <see cref="BaseScope"/>.
         /// </summary>
         /// <param name="scope">The scope.</param>
         /// <param name="key">The key.</param>
@@ -195,7 +196,7 @@ namespace Sentry
             => ((ConcurrentDictionary<string, string>)scope.Tags).AddOrUpdate(key, value, (s, o) => value);
 
         /// <summary>
-        /// Set all items as tags
+        /// Set all items as tags.
         /// </summary>
         /// <param name="scope">The scope.</param>
         /// <param name="tags"></param>
@@ -209,7 +210,7 @@ namespace Sentry
         }
 
         /// <summary>
-        /// Removes a tag from the <see cref="BaseScope"/>
+        /// Removes a tag from the <see cref="BaseScope"/>.
         /// </summary>
         /// <param name="scope">The scope.</param>
         /// <param name="key"></param>
@@ -217,7 +218,7 @@ namespace Sentry
             => scope.InternalTags?.TryRemove(key, out _);
 
         /// <summary>
-        /// Applies the data from one scope to the other while
+        /// Applies the data from one scope to the other.
         /// </summary>
         /// <param name="from">The scope to data copy from.</param>
         /// <param name="to">The scope to copy data to.</param>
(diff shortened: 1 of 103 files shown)


src/Sentry.Protocol/BaseScopeExtensions.cs
```diff
@@ -128,11 +128,6 @@ public static bool HasUser(this BaseScope scope)
         /// <param name="breadcrumb">The breadcrumb.</param>
         internal static void AddBreadcrumb(this BaseScope scope, Breadcrumb breadcrumb)
         {
-            if (scope == null)
```
bruno-garcia, line 130: We discussed this on discord and disagreed so I believe I should try to clarify my world view here so we can at least make sure we understand each side before we agree which direction to go.

In a pure C# 8 world, this change is correct. The compiler will make sure you can't:

```csharp
BaseScope scope = null;
scope.AddBreadcrumb(..);
```

But in the context of this code base, which includes a target `net461`, this could be problematic.

The potential problem I see is that an application targeting .NET 4.6.1 will compile just fine with the snippet above. The tooling at that version is unaware of the nullable annotations added by the C# 8 compiler. Which means, that in that case it could lead to a runtime error.

Options:

1.
```csharp

#if HAS_NULLABLE_REF_TYPE (or basically NET461)
if (scope == null) return;
#endif
```

2. 

```
#pragma CODE // net461 could pass null here
if (scope == null) return;
```

3. 

A better way?

src/Sentry.Protocol/BaseScopeExtensions.cs
```diff
@@ -170,7 +165,7 @@ public static void SetFingerprint(this BaseScope scope, IEnumerable<string> fing
         /// <param name="key">The key.</param>
         /// <param name="value">The value.</param>
         public static void SetExtra(this BaseScope scope, string key, object value)
-            => ((ConcurrentDictionary<string, object>)scope.Extra).AddOrUpdate(key, value, (s, o) => value);
+            => ((ConcurrentDictionary<string, object>) scope.Extra).AddOrUpdate(key, value, (s, o) => value);
```
bruno-garcia, line 168: Are these spaces and `.` on comments added automatically, on save?
These improvements are great but makes this PR a bit harder to read (1k+/1k-)

Maybe for the next PRs they can be split? If it's an automated thingy, feel free to run it on the whole codebase, we can merge without reviewing (more than skimming)

src/Sentry.Protocol/Context/Contexts.cs
```diff
@@ -59,13 +59,8 @@ internal Contexts Clone()
         /// <summary>
         /// Copies the items of the context while cloning the known types.
         /// </summary>
-        internal void CopyTo(Contexts? to)
+        internal void CopyTo(Contexts to)
         {
-            if (to == null)
```
bruno-garcia, line 62: Happy to kill these null checks on `internal` method btw, where we can guarantee we only call from C# 8+

src/Sentry/Extensibility/BaseRequestPayloadExtractor.cs
```diff
@@ -6,13 +6,12 @@ namespace Sentry.Extensibility
...
         /// <summary>
-        /// Extract the payload of the <see cref="IHttpRequest"/>
+        /// Extract the payload of the <see cref="IHttpRequest"/>.
         /// </summary>
-        /// <param name="request"></param>
-        /// <returns></returns>
-        public object ExtractPayload(IHttpRequest request)
+        public object? ExtractPayload(IHttpRequest request)
         {
-            if (!request.Body.CanSeek
```
bruno-garcia, line 15: You just fixed a bug :)

src/Sentry/Extensibility/SentryStackTraceFactory.cs
```diff
@@ -68,47 +69,41 @@ internal SentryStackTrace Create(StackTrace stackTrace, bool isCurrentStackTrace
                 stacktrace.Frames.Add(frame);
             }
 
-            return stacktrace.Frames.Count == 0
-                ? null
-                : stacktrace;
+            return stacktrace.Frames.Count != 0
```
bruno-garcia, line 72: Just curious: Why flip this?

src/Sentry/Extensibility/SentryStackTraceFactory.cs
```diff
@@ -68,47 +69,41 @@ internal SentryStackTrace Create(StackTrace stackTrace, bool isCurrentStackTrace
...
-        /// <param name="stackTrace"></param>
-        /// <param name="isCurrentStackTrace"></param>
-        /// <returns></returns>
         internal IEnumerable<SentryStackFrame> CreateFrames(StackTrace stackTrace, bool isCurrentStackTrace)
         {
-            var frames = stackTrace?.GetFrames();
+            var frames = stackTrace.GetFrames();
             if (frames == null)
             {
+                // TODO: this probably can never happen
```
bruno-garcia, line 85: Some environments don't bring stack traces like some Unity targets
Could be it returns 0 size array though, can't recall.

I'd say the TODO here is not helpful. We can discuss and remove the code if it's certain that it's never null.
```suggestion
```

bruno-garcia (commented):
First pass: Made it half way through.

src/Sentry/Internal/SentryScopeManager.cs
```diff
@@ -86,13 +86,13 @@ public void WithScope(Action<Scope> scopeCallback)
             using (PushScope())
             {
                 var scope = GetCurrent();
-                scopeCallback?.Invoke(scope.Key);
+                scopeCallback.Invoke(scope.Key);
```
bruno-garcia, line 89: Here I'm afraid `net461` apps might pass a null reference accidentally. That first point we discussed about null annotation vs defensive programming

bruno-garcia (commented):
90% done

src/Sentry.Protocol/BaseScopeExtensions.cs
Tyrrrz: Sorry, this one is automatic by Rider. I'll look through the changes and revert formatting changes to avoid noise.

src/Sentry.Protocol/Context/Contexts.cs
Tyrrrz: There's a bit of danger though as some internal methods are accessed from tests. I didn't want to accidentally break any of them so refrained from doing it.

src/Sentry/Extensibility/SentryStackTraceFactory.cs
Tyrrrz: I personally prefer when the positive/dominant branch appears first and the fallback appears last. I saw you had it like this in some other places anyway, so I flipped it to make it more consistent.

src/Sentry.Protocol/Dsn.cs
```diff
@@ -77,16 +81,16 @@ private Dsn(string dsn, string projectId, string path, string secretKey, string
...
 
         /// <summary>
-        /// Tries to parse the string into a <see cref="Dsn"/>
+        /// Tries to parse the string into a <see cref="Dsn"/>.
         /// </summary>
         /// <param name="dsn">The string to attempt parsing.</param>
         /// <param name="finalDsn">The <see cref="Dsn"/> when successfully parsed.</param>
         /// <returns><c>true</c> if the string is a valid <see cref="Dsn"/> as was successfully parsed. Otherwise, <c>false</c>.</returns>
-        public static bool TryParse(string dsn, out Dsn finalDsn)
+        public static bool TryParse(string dsn, out Dsn? finalDsn)
```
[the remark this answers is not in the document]
Tyrrrz: When I tried it last time, these attributes weren't well supported by IDEs and they would still show up a warning. I'll see if that has changed.

src/Sentry/Extensibility/SentryStackTraceFactory.cs
```diff
@@ -46,16 +45,18 @@ public SentryStackTrace Create(Exception exception = null)
         /// </summary>
         /// <param name="exception">The exception.</param>
         /// <returns>A StackTrace.</returns>
-        protected virtual StackTrace CreateStackTrace(Exception exception) =>
-            exception == null ? new StackTrace(true) : new StackTrace(exception, true);
+        protected virtual StackTrace CreateStackTrace(Exception? exception) =>
+            exception == null
```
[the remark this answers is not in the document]
Tyrrrz: Are we going to adopt `x is null` / `x is {}` convention?

src/Sentry/Extensibility/RequestBodyExtractionDispatcher.cs
```diff
@@ -31,22 +31,17 @@ public RequestBodyExtractionDispatcher(IEnumerable<IRequestPayloadExtractor> ext
         /// </summary>
         /// <param name="request">The request.</param>
         /// <returns>A serializable representation of the payload.</returns>
-        public object ExtractPayload(IHttpRequest request)
+        public object? ExtractPayload(IHttpRequest? request)
```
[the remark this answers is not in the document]
Tyrrrz: The return or the parameter? Both seem to be able to be null in some scenarios.

src/Sentry/HubExtensions.cs
```diff
@@ -1,96 +1,90 @@
...
         [EditorBrowsable(EditorBrowsableState.Never)]
         public static void AddBreadcrumb(
             this IHub hub,
-            ISystemClock clock,
-            string message,
-            string category = null,
-            string type = null,
-            IDictionary<string, string> data = null,
+            ISystemClock? clock,
+            string? message,
```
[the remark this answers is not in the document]
Tyrrrz: There are tests that set it to null. Should those be removed? I would prefer to remove the tests in a separate PR because then there's a risk I might break something.

src/Sentry/Internal/MainExceptionProcessor.cs
```diff
@@ -21,27 +21,27 @@ public MainExceptionProcessor(SentryOptions options, Func<ISentryStackTraceFacto
             SentryStackTraceFactoryAccessor = sentryStackTraceFactoryAccessor;
         }
 
-        public void Process(Exception exception, SentryEvent sentryEvent)
+        public void Process(Exception? exception, SentryEvent sentryEvent)
```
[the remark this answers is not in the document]
Tyrrrz: I believe the exception can be null as it's passed from downstream in tests.

src/Sentry/Internal/MainSentryEventProcessor.cs
```diff
@@ -148,9 +155,9 @@ public SentryEvent Process(SentryEvent @event)
                         Stacktrace = stackTrace
                     };
 
-                    @event.SentryThreads = @event.SentryThreads.Any()
+                    @event.SentryThreads = @event.SentryThreads?.Any() == true
                         ? new List<SentryThread>(@event.SentryThreads) { thread }
-                        : new[] { thread } as IEnumerable<SentryThread>;
+                        : new[] { thread }.AsEnumerable();
```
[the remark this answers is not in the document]
Tyrrrz: Shorter -> no need to specify generic argument -> fewer chances to make a mistake.
Also `as` can theoretically return `null`, although not in this situation. It may happen for example if we change the left side to something else and forget to update the right side. `AsEnumerable()` avoids all those problems.

src/Sentry/Internal/SdkComposer.cs
```diff
@@ -17,21 +16,23 @@ public SdkComposer(SentryOptions options)
...
-            if (_options.BackgroundWorker is IBackgroundWorker worker)
+            if (_options.BackgroundWorker is { } worker)
             {
                 _options.DiagnosticLogger?.LogDebug("Using IBackgroundWorker set through options: {0}.",
                     worker.GetType().Name);
 
                 return worker;
             }
 
+            // TODO: what to do here if Dsn is null?
```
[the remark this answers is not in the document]
Tyrrrz: Will do. I know the `Dsn` is expected to be set by this point, but since it has open getters/setter, there's no guarantee.

src/Sentry/Internal/MainExceptionProcessor.cs
```diff
@@ -21,27 +21,27 @@ public MainExceptionProcessor(SentryOptions options, Func<ISentryStackTraceFacto
...
         {
+            if (exception == null)
+                return;
```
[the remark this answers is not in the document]
Tyrrrz: Somehow didn't get any warnings. Not locally nor on CI. Changed.

src/Sentry/Extensibility/BaseRequestPayloadExtractor.cs
Tyrrrz: Great. Investment already paying off ;)

src/Sentry.Protocol/BaseScopeExtensions.cs
Tyrrrz: > The potential problem I see is that an application targeting .NET 4.6.1 will compile just fine with the snippet above. The tooling at that version is unaware of the nullable annotations added by the C# 8 compiler. Which means, that in that case it could lead to a runtime error.

That's actually not true. We use `Nullable` package for polyfilling, so the users on net461 will get the same NRT experience as on any other framework. Also, it's used for the same purpose for any target framework below netstandard2.1, which is basically everything Sentry has been targeting till now.

src/Sentry.Protocol/CollectionsExtensions.cs
```diff
@@ -9,7 +9,7 @@ internal static class CollectionsExtensions
             this ConcurrentDictionary<string, object> dictionary,
             string key)
             where TValue : class, new()
-            => dictionary.GetOrAdd(key, _ => new TValue()) as TValue;
+            => (TValue) dictionary.GetOrAdd(key, _ => new TValue());
```
Tyrrrz, line 12: This is an important change @bruno-garcia. Previously it was using `as` which could return `null` if the value has been set before, but to a different type. I glanced through code and it seems that _shouldn't_ happen so I replaced it with an explicit conversion which would throw instead. Can you validate my assumptions here?

src/Sentry.Protocol/CollectionsExtensions.cs
```diff
@@ -22,12 +22,8 @@ public static ConcurrentQueue<T> EnqueueAll<T>(this ConcurrentQueue<T> target, I
...
 
         public static void TryCopyTo<TKey, TValue>(this IDictionary<TKey, TValue> from, IDictionary<TKey, TValue> to)
+            where TKey : notnull
         {
-            if (from == null || to == null)
-            {
-                return;
-            }
-
             foreach (var kv in from)
```
Tyrrrz, line 27: This was not covered by tests so I removed it since it doesn't appear like the dictionaries are ever set to null. Is it okay @bruno-garcia?

src/Sentry.Protocol/Sentry.Protocol.csproj
```diff
@@ -1,15 +1,21 @@
...
-    <TargetFrameworks>netstandard2.0;netstandard1.3;net46;net45</TargetFrameworks>
+    <TargetFrameworks>netstandard2.1;netstandard2.0;netstandard1.3;net46;net45</TargetFrameworks>
     <Description>The Sentry Protocol used to communicate with Sentry - Open-source error tracking that helps developers monitor and fix crashes in real time.</Description>
+    <Nullable>enable</Nullable>
+  </PropertyGroup>
+
+  <!-- Disable nullability warnings on older frameworks because there is no nullability info for BCL -->
+  <PropertyGroup Condition="'$(TargetFramework)' != 'netstandard2.1'">
+    <Nullable>annotations</Nullable>
   </PropertyGroup>
```
Tyrrrz, line 12: Note this @bruno-garcia

src/Sentry/Extensibility/RequestBodyExtractionDispatcher.cs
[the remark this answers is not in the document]
Tyrrrz: In `BaseRequestPayloadExtractor` for example

src/Sentry/Extensibility/SentryStackTraceFactory.cs
```diff
@@ -117,9 +111,7 @@ internal IEnumerable<SentryStackFrame> CreateFrames(StackTrace stackTrace, bool
         /// <summary>
         /// Create a <see cref="SentryStackFrame"/> from a <see cref="StackFrame"/>.
         /// </summary>
-        /// <param name="stackFrame"></param>
-        /// <param name="isCurrentStackTrace"></param>
-        /// <returns></returns>
+        // TODO: should this really ignore isCurrentStackTrace?
```
Tyrrrz, line 114: @bruno-garcia

src/Sentry/SentryClient.cs
```diff
@@ -185,13 +187,13 @@ private SentryEvent BeforeSend(SentryEvent @event)
             _options.DiagnosticLogger?.LogDebug("Calling the BeforeSend callback");
             try
             {
-                @event = _options.BeforeSend?.Invoke(@event);
+                @event = @event != null ? _options.BeforeSend?.Invoke(@event) : null;
```
Tyrrrz, line 190: @bruno-garcia

src/Sentry.Protocol/BaseScopeExtensions.cs
bruno-garcia: > The only nuance is that C#8 is default on netcore3.0+ and netstd2.1+ but this is not the case on older targets, so the user will have to see the <LangVersion> themselves to benefit from NRTs and other features (but again, this is not just net461). This can be highlighted in docs.

Sadly this is an issue though. Most of users, for a long time, will simply update the SDK and not use C# 8 or newer targets.

In that case, it could throw at runtime.

src/Sentry/Extensibility/SentryStackTraceFactory.cs
bruno-garcia: I'm happy with that. I have no strong preferences here and do appreciate consistency so thanks for improcing this. Generally mixing different changes in the same PR make it harder to review. This is less important in small PRs but if they are huge, it's hard enough to go through them for what they are for

src/Sentry/Extensibility/RequestBodyExtractionDispatcher.cs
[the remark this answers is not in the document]
bruno-garcia: Sorry I wasn't clearn. I mean the IHttpRequest. I rather the caller not invoke this if no request object exists since there's no sense in running 'extract payload' in that case

src/Sentry/Extensibility/SentryStackTraceFactory.cs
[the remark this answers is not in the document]
bruno-garcia: What would you say? There's no operator overload on `==` in this code base but it is to my knowledge the only certain why to get the right behavior, aka: "the best way".

We could open a second PR later, to replace all the. == null with `is == null` instead though

src/Sentry/HubExtensions.cs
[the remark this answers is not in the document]
bruno-garcia: I'd say either we get the tests to run/compile with net461/netcore2 and pass null anyway (as discussed above we can't guarantee consumers will enable C# 8 and have polyfills), that'd be best.

src/Sentry/Internal/MainExceptionProcessor.cs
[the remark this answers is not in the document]
bruno-garcia: Once again probably tests were written to make sure a bad actor passing `null` wouldn't cause a whole thing to blow up or an event to be dropped/lost

src/Sentry/Internal/MainSentryEventProcessor.cs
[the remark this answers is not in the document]
bruno-garcia: I'm fine with changing this (best if we keep consistent with your approach going forward) even though it's more of a style thing. I personally can't move on from the fact it's a method call to simply make the compiler happy but one issue to me is that this is not really related to NRT so could be done on a separate PR, ideally.

Maybe a find and replace in the whole code base to make sure we stay consistent

bruno-garcia (commented):
I'm using the LOGAF scale here: https://blog.danlew.net/2020/04/15/the-logaf-scale/

src/Sentry/Extensibility/SentryStackTraceFactory.cs
bruno-garcia: Discussed on a call, public API used by consumers

bruno-garcia, line 114:
```suggestion
```

src/Sentry/SentryClient.cs
```diff
@@ -63,16 +63,15 @@ public SentryClient(SentryOptions options)
...
         /// </summary>
         /// <remarks>
         /// An optional scope, if provided, will be applied to the event.
         /// </remarks>
         /// <param name="event">The event to send to Sentry.</param>
         /// <param name="scope">The optional scope to augment the event with.</param>
-        /// <returns></returns>
         /// <inheritdoc />
-        public SentryId CaptureEvent(SentryEvent @event, Scope scope = null)
+        public SentryId CaptureEvent(SentryEvent? @event, Scope? scope = null)
```
bruno-garcia, line 74: Event shouldn't really be nullable (even though the code checks for it below to avoid NPE).
I'll address it in a follow up PR

src/Sentry/SentryClient.cs
```diff
@@ -102,27 +101,27 @@ public SentryId CaptureEvent(SentryEvent @event, Scope scope = null)
         /// <returns>A task to await for the flush operation.</returns>
         public Task FlushAsync(TimeSpan timeout) => Worker.FlushAsync(timeout);
 
-        private SentryId DoSendEvent(SentryEvent @event, Scope scope)
+        // TODO: this method needs to be refactored, it's really hard to analyze nullability
+        private SentryId DoSendEvent(SentryEvent @event, Scope? scope)
         {
-            if (_options.SampleRate is float sample)
+            if (_options.SampleRate != null)
```
bruno-garcia, line 107: Any specific reason to change this? Not sure why change from the pattern matching to null check and a call on the property again.

One of the reasons I prefer the pattern matching is that the value is available on the reference declared, as opposed to a second access to property, that could possibly have change in between the two reads if mutable.

I realize the amount of bikesheeding that C# has created by having so many ways to do the same thing

src/Sentry/SentryClient.cs
```diff
@@ -102,27 +101,27 @@ public SentryId CaptureEvent(SentryEvent @event, Scope scope = null)
...
-            if (@event.Exception is Exception ex
-                && _options.ExceptionFilters is IExceptionFilter[] filters && filters.Length > 0)
+            if (@event.Exception != null && _options.ExceptionFilters?.Length > 0)
             {
-                if (filters.Any(f => f.Filter(ex)))
+                if (_options.ExceptionFilters.Any(f => f.Filter(@event.Exception)))
                 {
                     _options.DiagnosticLogger?.LogInfo(
-                        "Event with exception of type '{0}' was dropped by an exception filter.", ex.GetType());
+                        "Event with exception of type '{0}' was dropped by an exception filter.", @event.Exception.GetType());
```
bruno-garcia, line 120: This doesn't warn that `@event.Exception` is nullable? The code before the change made sure it wasn't through pattern matching. Which IMHO is a better approach to deal with this

bruno-garcia (approved):
1 or 2 notes but LGTM, lets merge

src/Sentry/SentryClient.cs
bruno-garcia: I see the check above

PR #511, opened 2020-09-17
Add GitHubActionsTestLogger

Tyrrrz:
This will result in AppVeyor's output being a bit more noisy however. I don't know if that's an issue. If it is, I could improve the package to check if it actually runs on GHA or not.

### DIFF
--- a/build.ps1
+++ b/build.ps1
@@ -1,7 +1,8 @@
 $ErrorActionPreference = "Stop"
-$logger = If ($env:GITHUB_ACTIONS -eq "true") {"GitHubActions"} else {"console"}
 
-dotnet test -c Release -l $logger `
+$testLogger = if ($env:GITHUB_ACTIONS -eq "true") {"GitHubActions"} else {"console"}
+
+dotnet test -c Release -l $testLogger `
     /p:CollectCoverage=true `
     /p:CoverletOutputFormat=opencover `
     /p:Exclude='"""[Sentry.Protocol.Test*]*,[xunit.*]*,[Sentry.Test*]*\"""'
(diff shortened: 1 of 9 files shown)


build.ps1
```diff
@@ -1,6 +1,6 @@
 $ErrorActionPreference = "Stop"
 
-dotnet test -c Release `
+dotnet test -c Release -l GitHubActions `
```
bruno-garcia, line 3: Would be nice as you suggested for the package to auto detect, since folks (I for one) use these before pushing to remote

bruno-garcia (approved):

Tyrrrz:
This is how it looks:

image

Tyrrrz:
Should be good now. The build scripts will only use the logger when running on GHA and fallback to default console one locally and on travic/appveyor.

I had to bump target framework in Sentry.Protocols from net45 to net451. Didn't want to remove it outright as it seems like a bigger change to me, but we can do that in the future.

bruno-garcia:
image

src/Sentry.Protocol/BaseScope.cs
```diff
@@ -141,7 +141,7 @@ public User User
         /// An arbitrary mapping of additional metadata to store with the event.
         /// </summary>
         public
-#if NET45
+#if NET451
```
bruno-garcia, line 144: Next versio will be 3.0 so feel free to add breaking changes although we should keep track of them.

On a separate PR we can drop anything older than net461

bruno-garcia (approved):

src/Sentry.Protocol/BaseScope.cs
Tyrrrz: I'm stupid, we only need to change this in tests, not on the actual project. Will update.

PR #520, opened 2020-09-21
Add NRTs to Sentry.AspNetCore

Tyrrrz:


Tyrrrz:
There were a few guards here and there that were checking that injected objects were not null. I don't think you can inject a null in default AspNetCore DI container. Generally speaking, they still might make sense, but not sure if enough to keep them. Thoughts?

Tyrrrz:
@bruno-garcia looks like there are tests that pass `null`s into an internal class constructor. Should we remove those?

bruno-garcia:
> @bruno-garcia looks like there are tests that pass nulls into an internal class constructor. Should we remove those?

Yeah all the internal stuff we can guarantee will combile with C# 8 so feel free to delete all of those.

Guards in classes ctor I wouldn't touch if they are there. If they are instantiated via DI in ASP.NET Core they'd get an instance otherwise will fail to resolve but that assumes they'll only be resolved via DI and not used in a different way

bruno-garcia (approved):
✨ ✨ ✨

bruno-garcia (approved):
Makes sense, no need to optimzie for not having those. We just make sure they are there

PR #524, opened 2020-09-23
Add NRTS to Sentry.Extensions.Logging

Tyrrrz:


src/Sentry.Extensions.Logging/SentryLogger.cs
```diff
@@ -99,19 +99,20 @@ public bool IsEnabled(LogLevel logLevel)
...
                     // Avoid losing it by adding as data:
-                    data = data ?? new Dictionary<string, string>();
+                    data ??= new Dictionary<string, string>();
                     data.Add("exception_message", exception.Message);
                 }
 
                 _hub.AddBreadcrumb(
                     _clock,
-                    message ?? exception?.Message,
+                    message ?? exception?.Message!,
```
Tyrrrz, line 113: @bruno-garcia note the `!` here. This is needed because we annotated `message` as non-nullable even though it technically is, in fact nullable.

src/Sentry.Extensions.Logging/Sentry.Extensions.Logging.csproj
```diff
@@ -7,6 +7,12 @@
     <AssemblyName>Sentry.Extensions.Logging</AssemblyName>
     <RootNamespace>Sentry.Extensions.Logging</RootNamespace>
     <Description>Official Microsoft.Extensions.Logging integration for Sentry - Open-source error tracking that helps developers monitor and fix crashes in real time.</Description>
+    <Nullable>enable</Nullable>
```
bruno-garcia, line 10: If this is the last package, can we move this to the root Directory.Build.props?
Or actually just the `src`

src/Sentry.Extensions.Logging/SentryLogger.cs
bruno-garcia: OK, breadcrumb without a message looks crappy (useless?) but it's supported.
Let's leave the non-nullable annotation

src/Sentry.Extensions.Logging/SentryLoggerProvider.cs
```diff
@@ -57,9 +57,16 @@ public SentryLoggerProvider(IOptions<SentryLoggingOptions> options, IHub hub)
                 _scope = hub.PushScope();
                 hub.ConfigureScope(s =>
                 {
-                    s.Sdk.Name = Constants.SdkName;
-                    s.Sdk.Version = NameAndVersion.Version;
-                    s.Sdk.AddPackage(ProtocolPackageName, NameAndVersion.Version);
+                    if (s.Sdk is {} sdk)
```
bruno-garcia, line 60: Doesn't "CTRL + K + D" formats it like this?

```suggestion
                    if (s.Sdk is { } sdk)
```

I really would love for us to have an auto formatter that would apply as we save the file.
We'd run it on the whole solution first and then never have to pay attention at formatting again

bruno-garcia (approved):

src/Sentry.Extensions.Logging/SentryLoggerProvider.cs
Tyrrrz: Yes, resharper doesn't seem to care if it's `{}` or `{ }`...

PR #526, opened 2020-09-24
Framework and runtime info

bruno-garcia:


src/Sentry.NLog/SentryTarget.cs
```diff
@@ -404,14 +404,14 @@ protected override void Write(LogEventInfo logEvent)
             if (logEvent.Level >= Options.MinimumBreadcrumbLevel)
             {
                 var breadcrumbFormatted = RenderLogEvent(BreadcrumbLayout, logEvent);
-                var breadcrumbCategory = RenderLogEvent(BreadcrumbCategory, logEvent);
+                string? breadcrumbCategory = RenderLogEvent(BreadcrumbCategory, logEvent);
```
bruno-garcia, line 407: I dunno why this errors now

test/Sentry.Tests/Sentry.Tests.csproj
```diff
@@ -13,4 +13,17 @@
     <Reference Include="System.Web" />
   </ItemGroup>
 
+  <PropertyGroup Condition="$(TargetFramework.StartsWith('netcoreapp'))">
+    <DefineConstants>NETCOREAPP;$(AdditionalConstants)</DefineConstants>
```
bruno-garcia, line 17: There's probably a built-in

test/Sentry.Tests/Sentry.Tests.csproj
```diff
@@ -13,4 +13,17 @@
...
+  </PropertyGroup>
+
+  <PropertyGroup Condition="'$(TargetFramework)' == 'net461'">
+    <DefineConstants>NETFX;$(AdditionalConstants)</DefineConstants>
+  </PropertyGroup>
+
+  <ItemGroup Condition="$(TargetFramework.StartsWith('net4'))">
+    <PackageReference Include="Xunit.SkippableFact" Version="1.3.12" />
+    <Reference Include="System.Web" />
```
bruno-garcia, line 26: Do we need this?
```suggestion
```

src/Sentry.NLog/SentryTarget.cs
Tyrrrz: It only errors in Rider because the IDE is not smart enough to infer that a type can be nullable.

src/Sentry/PlatformAbstractions/FrameworkInfo.NetFx.cs
```diff
@@ -0,0 +1,195 @@
...
+                        && installation.Version >= latest.Version)
+                    {
+                        latest = installation;
+                    }
+                    else
+                    {
+                        break;
+                    }
+                }
+            }
```
Tyrrrz, line 79: Can this loop be replaced with `latest = installations.OrderBy(...).FirstOrDefault()` to avoid mutations and to make it simpler?

src/Sentry/PlatformAbstractions/FrameworkInfo.cs
```diff
@@ -0,0 +1,41 @@
...
+    /// The purpose of this partial class is to expose the API to all targets
+    /// For netstandard, the call to methods will be a simple no-op.
+    /// </summary>
+    public static partial class FrameworkInfo
+    {
+        /// <summary>
+        /// The map between release number and version number
+        /// </summary>
+        /// <see href="https://docs.microsoft.com/en-us/dotnet/framework/migration-guide/how-to-determine-which-versions-are-installed" />
+        public static IReadOnlyDictionary<int, string> NetFxReleaseVersionMap { get; set; }
```
Tyrrrz, line 16: Does this need a public setter?

src/Sentry/PlatformAbstractions/RegistryKeyExtensions.cs
```diff
@@ -0,0 +1,15 @@
+#if NETFX
+using Microsoft.Win32;
+
+namespace Sentry.PlatformAbstractions
+{
+    internal static class RegistryKeyExtensions
+    {
+        public static string GetString(this RegistryKey key, string value)
+            => key.GetValue(value) as string;
```
Tyrrrz, line 9: The return type should be nullable or the cast should be replaced with coercion.

src/Sentry/PlatformAbstractions/Runtime.cs
```diff
@@ -0,0 +1,138 @@
...
+        /// </example>
+        public string? Version { get; internal set; }
+#if NETFX
+        /// <summary>
+        /// The .NET Framework installation which is running the process
+        /// </summary>
+        /// <value>
+        /// The framework installation or null if not running .NET Framework
+        /// </value>
+        public FrameworkInstallation FrameworkInstallation { get; internal set; }
```
Tyrrrz, line 40: This class implements `IEquatable<T>` but has mutable properties. Could be dangerous if used in `Dictionary`.

src/Sentry/PlatformAbstractions/RuntimeExtensions.cs
```diff
@@ -0,0 +1,36 @@
...
+        /// <summary>
+        /// Is the runtime instance Mono.
+        /// </summary>
+        /// <param name="runtime">The runtime instance to check.</param>
+        /// <returns>True if it's Mono, otherwise false.</returns>
+        public static bool IsMono(this Runtime runtime) => runtime.IsRuntime("Mono");
+
+        private static bool IsRuntime(this Runtime? runtime, string runtimeName)
+        {
+            return runtime?.Name?.StartsWith(runtimeName) == true
```
Tyrrrz, line 32: Probably worth adding string comparison option here? I.e. ordinal or ordinal ignore case.

src/Sentry/PlatformAbstractions/RuntimeInfo.cs
```diff
@@ -0,0 +1,147 @@
...
+                        : match.Groups["name"].Value.Trim()),
+                    match.Groups["version"].Value,
+                    raw: rawRuntimeDescription);
+            }
+
+            return new Runtime(name, raw: rawRuntimeDescription);
+        }
+
+#if NETFX
+        internal static void SetReleaseAndVersionNetFx(Runtime runtime)
```
Tyrrrz, line 57: Naming is inconsistent with the other method, should be `SetNetFxReleaseAndVersion`

src/Sentry/PlatformAbstractions/RuntimeInfo.cs
```diff
@@ -0,0 +1,147 @@
...
+            {
+                return new Runtime(
+                    name ?? (match.Groups["name"].Value == string.Empty
+                        ? null
+                        : match.Groups["name"].Value.Trim()),
+                    match.Groups["version"].Value,
+                    raw: rawRuntimeDescription);
+            }
+
+            return new Runtime(name, raw: rawRuntimeDescription);
```
Tyrrrz, line 53: This block and conditionals can be simplified:

```suggestion
            var match = RuntimeParseRegex.Match(rawRuntimeDescription);
            
            return new Runtime(
                name ?? (match.Groups["name"].Value == string.Empty ? null : match.Groups["name"].Value.Trim()),
                match.Groups["version"].Value == string.Empty ? null : match.Groups["version"].Value.Trim(),
                rawRuntimeDescription
            );            
```

If `match.Success == false` then `match.Groups["foo"]` always evaluates to empty string. To simplify this further, I usually have a `string.NullIfEmptyOrWhiteSpace()` extension method.

src/Sentry/PlatformAbstractions/RuntimeInfo.cs
```diff
@@ -0,0 +1,147 @@
...
+                    friendlyVersion = "";
+                    break;
+                //case "4.0.30319.42000":
+                //    Debug.Fail("This is .NET Framework 4.6 or later which support RuntimeInformation");
+                //    break;
+                default:
+                    friendlyVersion = version.ToString();
+                    break;
+
+            }
```
Tyrrrz, line 143: Mutation can be avoided if you replace switch statement with switch expression.

Tyrrrz (commented):
Looks good, some small comments.

src/Sentry/PlatformAbstractions/RegistryKeyExtensions.cs
bruno-garcia: It compiles only for netfx so no warnings

bruno-garcia: Adding though

src/Sentry/PlatformAbstractions/Runtime.cs
bruno-garcia: True. Removed the interface

Issue #498, opened 2020-09-04
Reduce dependencies


bruno-garcia:
This will likely become multiple issues but for now to communicate the long term goal:

Vision: Have 0 dependencies.

This will improve the experience using Sentry with game engines, mobile. Avoid binding redirect issues with .NET Framework, and generally improve things.

- [x] Move code from Sentry.Protocol to namespace Sentry.Protocol in Sentry. Archive package on NuGet. Move SentryLevel to Sentry namespace
- [x] Move code from Sentry.PlatformAbstractions to Sentry (drop anything below ns2.0 and net461). Archive repo and project. First attempt is here: https://github.com/getsentry/sentry-dotnet/pull/264

~- [ ] Attempt to use System.Json (since it's not a dependency on netcore3.0+) See: https://github.com/getsentry/sentry-dotnet-protocol/pull/42 https://github.com/dotnet/runtime/issues/29975 (maybe .NET 5 [introduces a easy way into it](https://github.com/dotnet/runtime/issues/41313)) (Seems we'll be blocked by the lack of `JsonExtensionData`)~

- [x] Remove System.Web reference from the netfx target. To work on Mono builds on Android (i.e https://gatomalo.dev/blog/2020/03/21/error-monitoring-godot-sentry/ ) Alternative is creating `Sentry.AspNet` which will bring this dependency.

```
RROR: get_assembly_dependencies: Cannot load assembly (refonly): 'System.Web'.
   At: modules/mono/editor/godotsharp_export.cpp:93.
ERROR: get_assembly_dependencies: Cannot load one of the dependencies for the assembly: 'Sentry'.
   At: modules/mono/editor/godotsharp_export.cpp:98.
```

- [x] Docs for ASP.NET needs to be updated: https://docs.sentry.io/platforms/dotnet/guides/aspnet/

Tyrrrz:
@bruno-garcia seems like only the last point is left on the ticket.
Looking at https://docs.sentry.io/platforms/dotnet/guides/aspnet/, the only change seems to be the initialization (i.e. `SentrySdk.Init(o => o.AddAspNet())`), correct?
Also, how should we sync docs with the release cycle? Seeing as this change won't be in until the next major version.

bruno-garcia:
Good point. Maybe just leave a draft ready to be merged there? On the docs

bruno-garcia:
Or we could merge now the following:

```
Note: Version 3.0.0
[...]

PR #528, opened 2020-09-24
Move aspnet-classic integration to Sentry.AspNet

Tyrrrz:
Related to #498 

** To document breaking: **

ASP.NET classic users need to install a new package `Sentry.AspNet`.
`MaxRequestBodySize` is not longer on `SentryOptions` but instead is passed in when adding the ASP.NET integration.


Tyrrrz:
@bruno-garcia this is work in progress:

- `Sentry.AspNet` created
- `Sentry.AspNet.Tests` created
- Moved appropriate types to corresponding new projects

Problem: `SentryOptions` and `SentryOptionsExtensions` have dependency on `System.Web` currently. How do you advise to extract it?

I'm currently looking at this:

https://github.com/getsentry/sentry-dotnet-samples/blob/14945e07c73cba6c69ce0e404043ce0974d0a8cd/AspNetMvc5Ef6/Global.asax.cs#L21-L31

I suppose do something like this but replace that code with `SentrySdk.InitWeb()` or something?

Also where does `Sentry.EntityFramework` come from?

bruno-garcia:
>    Sentry.AspNet created
    Sentry.AspNet.Tests created
    Moved appropriate types to corresponding new projects

Awesome!

> Problem: SentryOptions and SentryOptionsExtensions have dependency on System.Web currently. How do you advise to extract it?


> I'm currently looking at this:
getsentry/sentry-dotnet-samples@14945e0/AspNetMvc5Ef6/Global.asax.cs#L21-L31
I suppose do something like this but replace that code with SentrySdk.InitWeb() or something?

I was thinking of simply adding an integration, like `AspNetIntegration` which will in turn upon `Register` being called, it would just add an event processor:

```csharp
Regsiter(IHub hub)
{
hub.ConfigureScope(s => s.AddEventProcessor<AspNetEventProcessor>());
}

```

`AspNetEventProcessor` we already have, it read the request data and adds stuff into the event.

To hook it up it could be an extension method to SentryOptions:

```
SentrySdk.Init(o => o.AddAspNet()); 
```

Or maybe some magic to (reflection) make `SentrySdk` in the core package discover integrations? I don't like that as I write it but could be an option to "auto discover" integration and avoid having to tell users: Write this extra snippet in this case.

** Also because if they upgrade now they'll simply lose the HTTP information.**


Maybe there's a better plan? We can discuss on a call.


> Also where does Sentry.EntityFramework come from?

https://github.com/getsentry/sentry-dotnet-ef/
We could bring in that to this repo but for the longest time I avoidad so we wouldn't need msbuild (outside `dotnet`). [It's not super popular](https://nugettrends.com/packages?months=12&ids=Sentry.EntityFramework), so I'm not sure it's worth it.

src/Sentry.AspNet/Sentry.AspNet.csproj
```diff
@@ -0,0 +1,24 @@
...
+
+  <PropertyGroup>
+    <TargetFramework>net461</TargetFramework>
+    <PackageId>Sentry.AspNet</PackageId>
+    <AssemblyName>Sentry.AspNet</AssemblyName>
+    <RootNamespace>Sentry.AspNet</RootNamespace>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <AssemblyAttribute Include="System.Runtime.CompilerServices.InternalsVisibleToAttribute">
```
bruno-garcia, line 11: Nice!
Doesn't need the signature stuff? even better, maybe a PR to replace it in the whole solution all of those `AssemblyInfo.cs`?

src/Sentry.Protocol/Properties/AssemblyInfo.cs
```diff
@@ -5,6 +5,7 @@
...
 [assembly: InternalsVisibleTo("Sentry.Testing,PublicKey=002400000480000094000000060200000024000052534131000400000100010059964a931488bcdbd14657f1ee0df32df61b57b3d14d7290c262c2cc9ddaad6ec984044f761f778e1823049d2cb996a4f58c8ea5b46c37891414cb34b4036b1c178d7b582289d2eef3c0f1e9b692c229a306831ee3d371d9e883f0eb0f74aeac6c6ab8c85fd1ec04b267e15a31532c4b4e2191f5980459db4dce0081f1050fb8")]
 [assembly: InternalsVisibleTo("Sentry.Log4Net.Tests,PublicKey=002400000480000094000000060200000024000052534131000400000100010059964a931488bcdbd14657f1ee0df32df61b57b3d14d7290c262c2cc9ddaad6ec984044f761f778e1823049d2cb996a4f58c8ea5b46c37891414cb34b4036b1c178d7b582289d2eef3c0f1e9b692c229a306831ee3d371d9e883f0eb0f74aeac6c6ab8c85fd1ec04b267e15a31532c4b4e2191f5980459db4dce0081f1050fb8")]
+[assembly: InternalsVisibleTo("Sentry.AspNet.Tests,PublicKey=002400000480000094000000060200000024000052534131000400000100010059964a931488bcdbd14657f1ee0df32df61b57b3d14d7290c262c2cc9ddaad6ec984044f761f778e1823049d2cb996a4f58c8ea5b46c37891414cb34b4036b1c178d7b582289d2eef3c0f1e9b692c229a306831ee3d371d9e883f0eb0f74aeac6c6ab8c85fd1ec04b267e15a31532c4b4e2191f5980459db4dce0081f1050fb8")]
```
bruno-garcia, line 8: Oh I thought the item in csproj replaced this

src/Sentry.AspNet/Sentry.AspNet.csproj
Tyrrrz: Nope, it only needs signature for strong named assemblies. I think it's the same for the attribute actually. In fact, this is the exact thing as applying the attribute, but using csproj instead.

src/Sentry.Protocol/Properties/AssemblyInfo.cs
Tyrrrz: Yeah, but it's a different project. I decided not to break consistency.

Tyrrrz:
> Or maybe some magic to (reflection) make SentrySdk in the core package discover integrations? I don't like that as I write it but could be an option to "auto discover" integration and avoid having to tell users: Write this extra snippet in this case.

> ** Also because if they upgrade now they'll simply lose the HTTP information.**

> Maybe there's a better plan? We can discuss on a call.

My opinion is that it's better to use explicit configuration for everything. That's the standard anyway, so you can expect the users to be familiar with it anyway.

Yes, by upgrading there are going to be breaking changes. That's the whole point of a breaking change right? ;)

src/Sentry.AspNet/Sentry.AspNet.csproj
Tyrrrz: Apparently the assemblies do get strong named on CI, so I have to add the public key anyway.

Tyrrrz:
@bruno-garcia this is my attempt at doing it. I couldn't use `SentryAspNetOptions` because then `SentrySdk.Init(() => ...)` wouldn't work. Instead I've added `MaxRequestBodySize` as parameter to `AddAspNet`. Not sure what to do about logging. Any ideas how to adapt that?

bruno-garcia:
That sounds good tbh. It's better than this inheritance hackage.

src/Sentry.AspNet/Sentry.AspNet.csproj
bruno-garcia: I think strong name in release builds, if not all builds.

Nice, can we have this done via csproj in all projs?

bruno-garcia (approved):
👍 

Is there a way to opt-out of those specific warnings annotations?
Maybe we need to ignore the warnings.

Those tests skip on some targets, but run on others, on all builds it'll be like that. Kind of noisy

Tyrrrz:
Yes there is. I'll add it after this.

PR #529, opened 2020-09-26
GCP Cloud Functions

bruno-garcia:
Hit this https://github.com/GoogleCloudPlatform/functions-framework-dotnet/issues/145 while trying this out.


src/Sentry.Google.Cloud.Functions/Class1.cs
```diff
@@ -0,0 +1,26 @@
...
+            base.ConfigureLogging(context, logging);
+        }
+
+
+        public override void ConfigureServices(WebHostBuilderContext context, IServiceCollection services)
+        {
+            base.ConfigureServices(context, services);
+        }
+    }
+}
```
Tyrrrz, line 26: File name is wrong

bruno-garcia:
Now missing some tests, and a refactor for the part I copied over from ASP.NET Core, to be reused there and avoid that duplication

Tyrrrz:
Do you have some analyzer that requires you to write `_ = expression`? I've seen it around the code. It's a bit weird, as very few methods that return something are actually pure in C#.

bruno-garcia:
@Tyrrrz .editorconfig
I adopted that recently even though Rider isn't nagging me, but I was hoping to be clear to when we discard returns. 

Strong feels against this?

Tyrrrz:
Yes, I don't think it adds value. It makes sense for pure functions (i.e. marked with `[Pure]`) but then you would never want to discard the return anyway. For impure functions, disgarding the return value often makes sense (which is evident by the fact that you very often do `_ =`). I think it just adds noise and nothing else.

bruno-garcia:
Discarding a `Task` is a valuable example. Do we have broken build or a warning if we don't discard it today?

bruno-garcia:
Work around on msbuild can be deleted: https://github.com/GoogleCloudPlatform/functions-framework-dotnet/issues/145

Tyrrrz:
> Discarding a `Task` is a valuable example. Do we have broken build or a warning if we don't discard it today?

Not sure about build, but Rider will definitely warn against it. If Task is the only such example, it doesn't make sense to add noise for all other types of expressions.

bruno-garcia:
`Task` is the biggest example I can think of. I don't have strong feelings about this, we can drop it.
I don't find it noise tbqh, I find that I can spot by readin the code outside an IDE (I do it more often than now these days) that the method indeed returns something but it's not used.

If that's just something that bothers your, feel free to open a PR deleting all cases and dropping the `.editorconfig` setting. Also feel free to merge without PR approval

bruno-garcia:
6th sense eh? 12 hours ago:

image

Tyrrrz:
> 6th sense eh? 12 hours ago:
> 
> image

Damn, you really foresaw it 9 months ago :joy:

Issue #235, opened 2019-06-26
Make Dsn internal


bruno-garcia:
Dsn parsing is not really used outside of the DSN and should be made internal on 2.0.0

Tyrrrz:
It would also be nice to refactor it so that the parsing happens outside of constructor, with a constructor only accepting the results from the parser. This would heavily reduce allocations and make the workflow more predictable.

The type can also be changed to `readonly struct` as well.

PR #532, opened 2020-09-28
Refactor Dsn type

Tyrrrz:
First pass: make parsing more linear.

Note I replaced the outdated `TryParse(out)` pattern with nullable `TryParse`. The original pattern was established way before value tuples or NRTs were added to the language, so I personally think there's no point in using it anymore. And `out` parameters are very ugly.

Second pass: make it internal?

I tried making Dsn internal (#235) but it caused a lot of issues. The majority of the pain comes from `SentryOptions` where `Dsn` is currently used in a bunch of delegate signatures. We could replace that with `string` but then it will be unclear what that string represents.

https://github.com/getsentry/sentry-dotnet/blob/709a5bb43078d0d130da95354ca7a98b0b847679/src/Sentry/SentryOptions.cs#L296-L310

Note: temporarily removed xmldocs to make it a bit easier to work with.

### DIFF
--- a/src/Sentry/Dsn.cs
+++ b/src/Sentry/Dsn.cs
@@ -1,7 +1,6 @@
 using System;
 using Sentry.Protocol;
 
-// ReSharper disable once CheckNamespace
 namespace Sentry
 {
     /// <summary>
@@ -10,42 +9,64 @@ namespace Sentry
     /// <remarks>
     /// <see href="https://docs.sentry.io/quickstart/#configure-the-dsn"/>
     /// </remarks>
-    public sealed class Dsn
+    internal sealed class Dsn
     {
-        public string OriginalString { get; }
+        /// <summary>
+        /// Source DSN string.
+        /// </summary>
+        public string Source { get; }
 
+        /// <summary>
+        /// The project ID which the authenticated user is bound to.
+        /// </summary>
         public string ProjectId { get; }
 
+        /// <summary>
+        /// An optional path of which Sentry is hosted.
+        /// </summary>
         public string? Path { get; }
 
+        /// <summary>
+        /// The optional secret key to authenticate the SDK.
+        /// </summary>
         public string? SecretKey { get; }
 
+        /// <summary>
+        /// The required public key to authenticate the SDK.
+        /// </summary>
         public string PublicKey { get; }
 
-        public Uri SentryUri { get; }
+        /// <summary>
+        /// Sentry API's base URI.
+        /// </summary>
+        private Uri ApiBaseUri { get; }
 
         private Dsn(
-            string originalString,
+            string source,
             string projectId,
             string? path,
             string? secretKey,
             string publicKey,
-            Uri sentryUri)
+            Uri apiBaseUri)
         {
-            OriginalString = originalString;
+            Source = source;
             ProjectId = projectId;
             Path = path;
             SecretKey = secretKey;
             PublicKey = publicKey;
-            SentryUri = sentryUri;
+            ApiBaseUri = apiBaseUri;
         }
 
-        public override string ToString() => OriginalString;
+        public Uri GetStoreEndpointUri() => new Uri(ApiBaseUri, "store/");
 
-        internal static bool IsDisabled(string dsn) =>
+        public Uri GetEnvelopeEndpointUri() => new Uri(ApiBaseUri, "envelope/");
+
+        public override string ToString() => Source;
+
+        public static bool IsDisabled(string? dsn) =>
             Constants.DisableSdkDsnValue.Equals(dsn, StringComparison.OrdinalIgnoreCase);
 
-        internal static Dsn Parse(string dsn)
+        public static Dsn Parse(string dsn)
         {
             var uri = new Uri(dsn);
 
@@ -56,32 +77,31 @@ namespace Sentry
             }
 
             var keys = uri.UserInfo.Split(':');
+
             var publicKey = keys[0];
             if (string.IsNullOrWhiteSpace(publicKey))
             {
                 throw new ArgumentException("Invalid DSN: No public key provided.");
             }
 
-            string? secretKey = null;
-            if (keys.Length > 1)
-            {
-                secretKey = keys[1];
-            }
+            var secretKey = keys.Length > 1
+                ? keys[1]
+                : null;
 
             var path = uri.AbsolutePath.Substring(0, uri.AbsolutePath.LastIndexOf('/'));
-            var projectId = uri.AbsoluteUri.Substring(uri.AbsoluteUri.LastIndexOf('/') + 1);
 
+            var projectId = uri.AbsoluteUri.Substring(uri.AbsoluteUri.LastIndexOf('/') + 1);
             if (string.IsNullOrWhiteSpace(projectId))
             {
                 throw new ArgumentException("Invalid DSN: A Project Id is required.");
             }
 
-            var sentryUri = new UriBuilder
+            var apiBaseUri = new UriBuilder
             {
                 Scheme = uri.Scheme,
                 Host = uri.DnsSafeHost,
                 Port = uri.Port,
-                Path = $"{path}/api/{projectId}/store/"
+                Path = $"{path}/api/{projectId}/"
             }.Uri;
 
             return new Dsn(
@@ -90,11 +110,11 @@ namespace Sentry
                 path,
                 secretKey,
                 publicKey,
-                sentryUri
+                apiBaseUri
             );
         }
 
-        internal static Dsn? TryParse(string? dsn)
+        public static Dsn? TryParse(string? dsn)
         {
             if (string.IsNullOrWhiteSpace(dsn))
             {
--- a/src/Sentry/Internal/Hub.cs
+++ b/src/Sentry/Internal/Hub.cs
@@ -24,7 +24,7 @@ namespace Sentry.Internal
             Debug.Assert(options != null);
             _options = options;
 
-            if (string.IsNullOrWhiteSpace(options.Dsn))
+            if (options.Dsn is null)
             {
                 var dsn = DsnLocator.FindDsnStringOrDisable();
 
--- a/src/Sentry/Internal/Hub.cs
+++ b/src/Sentry/Internal/Hub.cs
@@ -24,7 +24,7 @@ namespace Sentry.Internal
             Debug.Assert(options != null);
             _options = options;
 
-            if (options.Dsn == null)
+            if (string.IsNullOrWhiteSpace(options.Dsn))
             {
                 var dsn = DsnLocator.FindDsnStringOrDisable();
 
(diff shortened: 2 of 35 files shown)


bruno-garcia:
Please ValueTuple was used in the past but to avoid the dependency I ripped it off:

https://github.com/getsentry/sentry-dotnet/commit/f5fa8a7d72f2e08f088ab19541b868b898997473#diff-74a6cbd4a1979629a6cff4f9f9571585

bruno-garcia:
> Dsn is currently used in a bunch of delegate signatures. We could replace that with string but then it will be unclear what that string represents.

We could revisit those callbacks, maybe get rid of them or replace with something else.
Lets discuss what problem are we solving with them and how else could this be addressed?

Tyrrrz:
This approach actually invalidates the need for tuples completely @bruno-garcia. The constructor is private so there's no redundant passing of data around.

src/Sentry/Internal/Hub.cs
```diff
@@ -26,12 +26,15 @@ public Hub(SentryOptions options)
 
             if (options.Dsn == null)
             {
-                if (!Dsn.TryParse(DsnLocator.FindDsnStringOrDisable(), out var dsn))
+                var dsn = Dsn.TryParse(DsnLocator.FindDsnStringOrDisable());
```
bruno-garcia, line 29: What about we call it just `Parse` then? Since `TryParse` feels like the classic bool returning `out` parameter approach.

src/Sentry/Protocol/Dsn.cs
```diff
@@ -12,138 +10,56 @@ namespace Sentry
...
 
-        /// <summary>
-        /// An optional path of which Sentry is hosted.
-        /// </summary>
         public string? Path { get; }
 
-        /// <summary>
-        /// The optional secret key to authenticate the SDK.
-        /// </summary>
         public string? SecretKey { get; }
```
bruno-garcia, line 21: Why get rid of all the docs?

src/Sentry/Protocol/Dsn.cs
```diff
@@ -157,27 +73,43 @@ public static bool TryParse(string dsn, [NotNullWhen(true)] out Dsn? finalDsn)
...
+                throw new ArgumentException("Invalid DSN: A Project Id is required.");
             }
 
-            var builder = new UriBuilder
+            var sentryUri = new UriBuilder
             {
                 Scheme = uri.Scheme,
                 Host = uri.DnsSafeHost,
                 Port = uri.Port,
                 Path = $"{path}/api/{projectId}/store/"
```
bruno-garcia, line 84: Please note we'll have a new endpoint `/envelope` used soon.
In fact `/store` will no longer be supported.

Good chance for us to rethink this.

bruno-garcia (commented):
Good direction.
A couple notes so we can discuss further

src/Sentry/Internal/Hub.cs
Tyrrrz: Up to you. The current `Parse` works differently, i.e. it throws with a descriptive error. I think that could still be useful.

> Since TryParse feels like the classic bool returning out parameter approach.

I think this will pass :)
I had it at first and then got used to it, now I never use `out` for this.

src/Sentry/Protocol/Dsn.cs
Tyrrrz: I'll bring them back before "undrafting" the PR.

src/Sentry/Protocol/Dsn.cs
Tyrrrz: Is there anything to think about besides changing from `store` to `envelope`? Some potential pitfalls?

bruno-garcia: the problem is that it makes the SDK no longer work with old on-premise Sentry.
But since this is a major bump and other SDKs (mobile, native) have done this, it's just the way of things.

Folks can upgrade their on prem with Sentry, it's easier than ever these days.

For now just consider that `Dsn` not only has a "store" endpoint but will have "envelope" endpoint soon. And we'll delete the "store" one later.

Maybe a method `EnvelopeUrl` or something, didn't think this through.
In Java we just don't add neither segment:

https://github.com/getsentry/sentry-java/blob/main/sentry/src/main/java/io/sentry/Dsn.java

So the caller can figure it out

Tyrrrz: What we can do is store the base uri and expose methods to get the store/envelope uri. Wdyt?

src/Sentry/Internal/Hub.cs
bruno-garcia: It's a .NET convention that has existed for 20 years though, it's what people expect.

Since this is internal API it doesn't really matter. But I'd vote for a simple `Parse` or call it `ParseOrNull` or something descriptive.

Since it's internal the second method that throws isn't needed, we can have a single one, maybe one that throws and we handle it?

image
Reading this I assume `dsn` is bool.

src/Sentry/Protocol/Dsn.cs
bruno-garcia: sounds good!

src/Sentry/Internal/Hub.cs
Tyrrrz: Can you consider the difficulty of changing it to internal? I mentioned in the PR description.

bruno-garcia: I thought I did with this comment:

> We could revisit those callbacks, maybe get rid of them or replace with something else. 
Lets discuss what problem are we solving with them and how else could this be addressed?

bruno-garcia: Maybe we just pass SentryOptions to the callback. That give more control to the caller to decide how to create the the HttpClient.

Still up for discussion what hooks we should put there. I added those without a lot (any) of input from customers

Tyrrrz: > Maybe we just pass SentryOptions to the callback.

I like this way more, yes.

Tyrrrz: But then again, those callbacks are probably set from the context of initializing the SDK anyway. So something like this:

```csharp
SentrySdk.Init(o =>
{
    // SentryOptions (o) is already in scope here so no reason to pass it as a parameter
    o.CreateHttpClientHandler = () => new HttpClientHandler();
});
```

Tyrrrz: So maybe just simplify the signature by removing `Dsn` altogether?

Tyrrrz:
@bruno-garcia making Dsn internal might be a bad idea with the current architecture. Because otherwise we end up with a lot of `Dsn.TryParse()` that just check if the result is not null and then discard it because they can't assign it anywhere.
Of course, ideally we should [parse, not validate](https://lexi-lambda.github.io/blog/2019/11/05/parse-don-t-validate/), but it's even a bigger issue here because we're wasting resources on parsing URL every time.
This can be improved if the parts responsible for bootstrapping the client/hub/whatever used a more normalized version of `SentryOptions`. But ideally, the API should be reworked.

Tyrrrz:
@bruno-garcia ready for review.

src/Sentry/Dsn.cs
```diff
@@ -0,0 +1,135 @@
...
+                dsn,
+                projectId,
+                path,
+                secretKey,
+                publicKey,
+                apiBaseUri
+            );
+        }
+
+        public static Dsn? TryParse(string? dsn)
```
bruno-garcia, line 117: 🙄 

The other day a junior dev asked me the difference between TryParse and Parse (for int, double etc).

That dev would be really confused reading this code.

src/Sentry/Internal/Hub.cs
```diff
@@ -24,14 +24,17 @@ public Hub(SentryOptions options)
             Debug.Assert(options != null);
             _options = options;
 
-            if (options.Dsn == null)
+            if (string.IsNullOrWhiteSpace(options.Dsn))
```
bruno-garcia, line 27: In Sentry SDKs, `""` means: Disable SDK.

If dsn was null here, we try to lookup.
If dsn was `""`, we should just let the program run, the SDK will be disabled (don't try to lookup for a DSN in env var, don't throw exception, etc)

```suggestion
            if (options.Dsn is null)
```

src/Sentry/Internal/OptionalHub.cs
```diff
@@ -19,14 +19,17 @@ public OptionalHub(SentryOptions options)
         {
             options.SetupLogging();
 
-            if (options.Dsn == null)
+            if (string.IsNullOrWhiteSpace(options.Dsn))
```
bruno-garcia, line 22:
```suggestion
            if (options.Dsn is null)
```

src/Sentry/SentrySdk.cs
```diff
@@ -78,13 +71,16 @@ public static IDisposable Init(Action<SentryOptions>? configureOptions)
         [EditorBrowsable(EditorBrowsableState.Never)]
         public static IDisposable Init(SentryOptions options)
         {
-            if (options.Dsn == null)
+            if (string.IsNullOrWhiteSpace(options.Dsn))
```
bruno-garcia, line 74:
```suggestion
            if (options.Dsn is null)
```

src/Sentry/SentrySdk.cs
```diff
@@ -78,13 +71,16 @@ public static IDisposable Init(Action<SentryOptions>? configureOptions)
...
             {
-                if (!Dsn.TryParse(DsnLocator.FindDsnStringOrDisable(), out var dsn))
+                var dsn = DsnLocator.FindDsnStringOrDisable();
+
+                if (Dsn.TryParse(dsn) is null)
```
bruno-garcia, line 78: if it's not `null` (which means it could locate any DSN), and it's not `""` (which is intentional "disable SDK please", according to sentry SDK spec) and it failed parsing, it's OK to throw an exception here.

It means the user has a broken DSN and it's worth letting them know, since it'll fail consistently here during app bootstrap.

bruno-garcia (changes requested):
This is great! Looking forward to merge this.
A couple of small points first, forgot to mention that `""` has some meaning in Sentry

> The SDK should accept an empty DSN as valid configuration.

https://develop.sentry.dev/sdk/overview/#usage-for-end-users

src/Sentry/Internal/Hub.cs
Tyrrrz: Ah I see.  Will change.
Interestingly, no tests had failed ;)

src/Sentry/Dsn.cs
Tyrrrz: Says the person who thinks that namespaces shouldn't correspond to directory names ;)

Tyrrrz:
@bruno-garcia please re-review before I go absolutely insane with this.

bruno-garcia (approved):

bruno-garcia:
Thanks @Tyrrrz

PR #540, opened 2020-10-01
Rename LogEntry to SentryMessage

Tyrrrz:
https://github.com/getsentry/sentry-dotnet/projects/3#card-46040296

Open questions:

- What should `event.Message = "foo"` be replaced with? I replaced them with `event.Message = new SentryMessage { Raw = "foo" }` but not sure if it's correct.
- Should `LogEntryFilter` and related also be renamed?

### DIFF
--- a/samples/Sentry.Samples.Console.Customized/Program.cs
+++ b/samples/Sentry.Samples.Console.Customized/Program.cs
@@ -163,7 +163,7 @@ internal static class Program
                 {
                     Message = new SentryMessage
                     {
-                        Raw = msg,
+                        Message = msg,
                         Formatted = string.Format(msg, i, count)
                     },
                     Level = SentryLevel.Debug
@@ -178,7 +178,7 @@ internal static class Program
             // would get disposed by the container on app shutdown
 
             var evt = new SentryEvent();
-            evt.Message = new SentryMessage {Raw = "Starting new client"};
+            evt.Message = "Starting new client";
             evt.AddBreadcrumb("Breadcrumb directly to the event");
             evt.User.Username = "some@user";
             // Group all events with the following fingerprint:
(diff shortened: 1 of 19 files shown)


samples/Sentry.Samples.Console.Customized/Program.cs
```diff
@@ -161,9 +161,9 @@ private static async Task Main()
                 const string msg = "{0} of {1} items we'll wait to flush to Sentry!";
                 SentrySdk.CaptureEvent(new SentryEvent
                 {
-                    LogEntry = new LogEntry
+                    Message = new SentryMessage
                     {
-                        Message = msg,
+                        Raw = msg,
```
bruno-garcia, line 166: I agree the name isn't great, but I'd rather we stick to the C# name being closer to the protocol name.

src/Sentry.Extensions.Logging/SentryLogger.cs
```diff
@@ -59,7 +59,7 @@ public bool IsEnabled(LogLevel logLevel)
                 var @event = new SentryEvent(exception)
                 {
                     Logger = CategoryName,
-                    Message = message,
+                    Message = new SentryMessage{ Raw = message },
```
bruno-garcia, line 62: What about an `implicit` operator from `String` to `SentryMessage`?

bruno-garcia (approved):
Awesome! Thanks.

> Should LogEntryFilter and related also be renamed?

I guess, if they used to take just String, perhaps best is now the whole SentryMessage

Tyrrrz:
@bruno-garcia they don't appear to. In fact, it seems that the only thing they had in common is the name.

image

bruno-garcia:
Sonds good. yeha they only match on the values given to the MEL logger.

Tyrrrz:
So keep it as it is?

PR #542, opened 2020-10-01
Change type of Gpu.VendorId from int to string

### LINKED ISSUE #541: Bug: GPU vendor_id is string
https://github.com/getsentry/sentry-dotnet/blob/d44caebaa3514782daded255474d603c6f244b63/src/Sentry/Protocol/Context/Gpu.cs#L47

Server is complaining if `int` is sent, it expects `string`.

Tyrrrz:
Fixes #541


bruno-garcia (approved):
Dang! 🚀 
Thanks a lot!!

PR #546, opened 2020-10-12
Use ConfigureAwait.Fody

Tyrrrz:
Imagine having to write `ConfigureAwait(false)` in 2020.


Tyrrrz:
Hmm... CI is just stuck for some reason.

bruno-garcia (commented):
If you wanted to net some lines here, it backfire: 😄 
image

What about we add the `xml` files on the `src` folder and create a symlink?

Tyrrrz:
> If you wanted to net some lines here, it backfire:

True, but those added lines are configuration files, not in code ;)

> What about we add the xml files on the src folder and create a symlink?

Could try. I'm not sure how to do that.

bruno-garcia:
> Could try. I'm not sure how to do that.

Try adding the xml and xsd file to `src`, then on each project: `ln -s ../bla.xml bla.xml` to create `bla.xml` locally as a symlink to the parent folder's file. This way we have 1 file, and all projects just link to it.

Tyrrrz:
That's new

Tyrrrz:
Interesting, the tests pass in debug mode, but freeze up in release mode.

Tyrrrz:
Very interesting. Looks like in `Release` configuration the `.Do(p => ...)` delegate is not getting invoked at all, so as the result the MRE doesn't get triggered and the test hangs.

bruno-garcia:
Maybe a regression on the mocking lib? This is odd

src/Sentry.AspNetCore/FodyWeavers.xsd
```diff
@@ -0,0 +1,30 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
```
bruno-garcia, line 1: Anyv luck with the symlinks? These 300 lines of xml make me nervous

Tyrrrz:
No idea why it worked before (c), but now I changed it to trigger the signal on a separate thread, which avoids the deadlock.

src/Sentry.AspNetCore/FodyWeavers.xsd
Tyrrrz: Will look into this next

Tyrrrz:
Lol, now `BackgroundWorkerTests` fail on release but not on debug. Interesting.

bruno-garcia:
I'll blame it on IL magic breaking stuff :P

Tyrrrz:
Damn you're right:

image

My guess is that NSubstitute and Fody both muddle with IL and it ends up in this weird state. I don't know what to do here in the short term, apart from switching to Moq in hopes that it helps.

src/Sentry.AspNetCore/FodyWeavers.xsd
Tyrrrz: I seem to have created a symlink (using `mklink` on Windows tho), but not sure how git is supposed to handle that. Reading [here](https://stackoverflow.com/questions/954560/how-does-git-handle-symbolic-links), it seems that git still handles them as separate files (which seems to be confirmed by `git diff`):
image

bruno-garcia:
> My guess is that NSubstitute and Fody both muddle with IL and it ends up in this weird state. I don't know what to do here in the short term, apart from switching to Moq in hopes that it helps.

I propose we just write `ConfigureAwait(false)` then. It's just "ugly", there's no downside in affecting customers usage, performance. size of the binary or anything other than make programmer writing the SDK write boiler plate.

src/Sentry.AspNetCore/FodyWeavers.xsd
Tyrrrz: Nevermind, apparently I was creating a hardlink and not a symlink (no idea what the difference is, but it seems to work now).

Tyrrrz:
> ...anything other than make programmer writing the SDK write boiler plate.

Which results in noise that increases cognitive load. But yeah, I agree, it seems that we reached a point where it conflicts with other dependencies.

bruno-garcia:
> Which results in noise that increases cognitive load. But yeah, I agree, it seems that we reached a point where it conflicts with other dependencies.

Wanna contribute to the Objective-C SDK or Java 1.8 or some other goodie? :)

Tyrrrz:
> > Which results in noise that increases cognitive load. But yeah, I agree, it seems that we reached a point where it conflicts with other dependencies.
> 
> Wanna contribute to the Objective-C SDK or Java 1.8 or some other goodie? :)

I can contribute my disdain!

PR #547, opened 2020-10-14
Fix callback signature BeforeSend 

### LINKED ISSUE #545: Fix callback signature BeforeSend
The callbacks are expected to return null to indicate an event/crumb should be dropped:

https://github.com/getsentry/sentry-dotnet/blob/42b0f921f0503c561046b96c3ad486511487e7cb/src/Sentry/SentryOptions.cs#L214
`public Func<SentryEvent, SentryEvent>? BeforeSend { get; set; }`

Should be:

`public Func<SentryEvent, SentryEvent?>? BeforeSend { get; set; }`

Same for BeforeBreadcrumb:

`public Func<Breadcrumb, Breadcrumb>? BeforeBreadcrumb { get; set; }`

Should be:
`public Func<Breadcrumb, Breadcrumb?>? BeforeBreadcrumb { get; set; }`

chan18:
Fixes #545

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 * ASPNET.Core hosting environment is set to 'production' / 'development' (notice lower casing) if no custom options.Enviroment is set. (#554) @PureKrome
 * Add most popular libraries to InAppExclude #555 (@bruno-garcia)
 * Add support for individual rate limits.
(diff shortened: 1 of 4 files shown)


bruno-garcia:
Thanks for the PR. I've added the label to the issue: https://github.com/getsentry/sentry-dotnet/issues/545

Please make sure CI is passing properly

bruno-garcia:
And add a changelog item

PR #550, opened 2020-10-15
Default Environment setting if none provided. 

### LINKED ISSUE #549: Send a default Environment if nothing was set by the user
ASP.NET Core has the concept of Environment OOTB. That's not the case for the plain `Sentry`.

Sentry is more valuable when an environment is set. The SDK guidelines were updated to indicate an environment ([by default `production`](https://github.com/getsentry/develop/commit/676ab2bc2818d91ba8aa78dbf45768121f835420) should be set to all events, if nothing was set by the user).

This should happen right before calling user callbacks like `beforeSend`. MainEventProcessor is ideal. The logic could be:

```csharp
if (@event.Environment is null)
{
    @event.Environment = "production";
}
```

This needs tests:
* set a value to `null` and make sure `production is in the captured event
* set a value to `something` and make sure it's captures as `something`, to validate not overwritten.
* if aspnetcore doesn't have a test, make sure it does set the right environment, and doesn't break with this change (starts sending `production` regardless of `ASPNETCORE_ENVIRONMENT`)

PureKrome:
Fixes #549.

**Required**:

- ✔️ set a value to null and make sure `production is in the captured event
- ✔️ set a value to something and make sure it's captures as something, to validate not overwritten.
- ✔️ if aspnetcore doesn't have a test, make sure it does set the right environment, and doesn't break with this change (starts sending production regardless of ASPNETCORE_ENVIRONMENT)

**Done**:

- 👍🏻 Added check for `Environment` in `SentryEvent` and `SentryOptions` (see below, the reasoning)
- 👍🏻 Added `SentryClient` tests
- 👀 Checked existing ASP.NET Core tests and they seem to pass.

---

**Questions**:

A few questions with respect to this PR.

- I've included a `SentryOptions` check, also. My unskilled understanding is that the `SentryEvent` might have it set (which is very granular) but if not, then it's possible it was set via `SentryOptions` (higher up/earlier on) ... which we should then respect.

- I'm not sure I found the scenario's for testing the ASP.NET Core app with respect to testing+environments. I did end up finding [this test code](https://github.com/getsentry/sentry-dotnet/blob/f6aaeff0d010b272b1babc9d8c866f2af9ee305e/test/Sentry.AspNetCore.Tests/IntegrationMockedBackgroundWorker.cs#L209), which I tested again (and passed). This test code hinted at the use of setting (and testing) for options ([which I first read and referenced here](https://github.com/getsentry/sentry-dotnet/blob/f6aaeff0d010b272b1babc9d8c866f2af9ee305e/test/Sentry.AspNetCore.Tests/IntegrationMockedBackgroundWorker.cs#L213)). So I just added a note there, for future developers to get some hints to understanding the code base.


src/Sentry/SentryClient.cs
```diff
@@ -146,6 +146,15 @@ private SentryId DoSendEvent(SentryEvent @event, Scope? scope)
                 }
             }
 
+            // Each event should have an 'Environment' set. If none was set, then should default
+            // to 'production'.
+            if (string.IsNullOrWhiteSpace(@event.Environment))
+            {
+                @event.Environment = string.IsNullOrWhiteSpace(_options.Environment)
+                    ? "production"
+                    : _options.Environment;
```
bruno-garcia, line 155: Environment should be pulled already from option elsewhere. So right now I assume we apply the value from options here and there. Could we have this logic there? It's probably on MainEventProcessor so the whole block could go there.

bruno-garcia (changes requested):
That was quick @PureKrome ! Thanks a lot.
I left one note though, if you could please address that.

Regarding aspnet core there's a IntegrationTest.cs which could be the way to test it. It boots an app with options, so you could dupe a test and set an environment. Or assert th current one isn't default if that makes sense.

src/Sentry/SentryClient.cs
PureKrome: When I didn't pull the options out (above), the ASP.NET Core tests (linked in the OP) were failing, though.

I wasn't sure if this was a brittle test -or- a great test that picked up my mistake ?

PureKrome:
> there's a IntegrationTest.cs which could be the way to test it

Ah! i see it, i'll give that an 👀 and check it out.

src/Sentry/SentryClient.cs
PureKrome: Refactoring stuff.

Just found ` public SentryEvent Process(SentryEvent @event)` in `MainSentryEventProcessor`.

PureKrome:
@bruno-garcia Okay - found the more appropriate place for this code. I know you mentioned it in the Issue, but I just couldn't find it (global search foo also wasn't working for me 😊 )

What is really important here is that this code stops you from setting the `Environment` to 
- `null` (e.g. `options.Environment = null;`)
- `string.Empty` (e.g. `options.Environment = "";`)
- strings with only 1+ white spaces (e.g. `options.Environment = "        ";`)

So this means that there will (from now on) always be a value for the `options.Environment`). Not sure if this is what you want, etc.

src/Sentry/Internal/Constants.cs
```diff
@@ -20,6 +20,12 @@ internal static class Constants
         /// </summary>
         public const string EnvironmentEnvironmentVariable = "SENTRY_ENVIRONMENT";
 
+        /// <summary>
+        /// Default Sentry environment setting.
+        /// </summary>
+        /// <remarks>Best Sentry practice is to always try and have a value for this setting.</remarks>
+        public const string DefaultEnvironmentSetting = "Production";
```
bruno-garcia, line 27: In ASP.NET Core this value would be `Production` but in Sentry, all SDKs will default to `production` and this value IS case sensitive. So this is a bit of a pickle

```suggestion
        public const string DefaultEnvironmentSetting = "production";
```

I'll change ASP.NET Core to use the same for the default (IsDevelopment, IsProduction) so it matches. Would you be willing to file a PR for that too?

test/Sentry.AspNetCore.Tests/IntegrationMockedBackgroundWorker.cs
```diff
@@ -234,6 +234,7 @@ public void Environment_NotOnOptions_ValueFromEnvVar()
                 });
         }
 
+        // NOTE: Options take priority over EnvVar's.
```
bruno-garcia, line 237: This seem wrong. I believe env var should take priority.
Good we're going to release 3.0 so we can break this behavior

test/Sentry.Tests/Internals/MainSentryEventProcessorTests.cs
```diff
@@ -164,35 +164,45 @@ public void Process_NoReleaseOnOptions_SameAsCachedVersion()
...
         }
 
-        [Fact]
-        public void Process_NoEnvironmentOnOptions_SameAsEnvironmentVariable()
+        [Theory]
+        [InlineData(null, Constants.DefaultEnvironmentSetting)] // Missing: will get default value.
+        [InlineData("", Constants.DefaultEnvironmentSetting)] // Missing: will get default value.
+        [InlineData(" ", Constants.DefaultEnvironmentSetting)] // Missing: will get default value.
+        [InlineData("a", "a")] // Provided: nothing will change.
+        [InlineData("production", "production")] // Provided: nothing will change. (value has a lower case 'p', different to default value)
```
bruno-garcia, line 190: Given the privious suggestion change:

```suggestion
        [InlineData("Production", "Production")] // Provided: nothing will change. (value has a upper case 'p', different to default value)
```

src/Sentry/Internal/MainSentryEventProcessor.cs
```diff
@@ -136,9 +136,19 @@ public SentryEvent Process(SentryEvent @event)
                 @event.Release = _options.Release ?? Release;
             }
 
-            if (@event.Environment == null)
+            if (string.IsNullOrWhiteSpace(@event.Environment))
             {
-                @event.Environment = _options.Environment ?? EnvironmentLocator.Locate();
+                if (string.IsNullOrWhiteSpace(_options.Environment))
```
bruno-garcia, line 141: I understand you did this right thing here to stay compatibel.
But could youb please change to use env var instead of options first? Values set via env var should take priority over the options.

bruno-garcia (commented):
Nice and clean change. Thanks.
I did uncover some behavior which I believe isn't correct. If you don't mind, you could fix in this PR. Otherwise we cna merge and have a follow up

src/Sentry/Internal/Constants.cs
PureKrome: >  but in Sentry, all SDKs will default to production and this value IS case sensitive
So what happens if a user _manually_ sets the value to `Production` or `Development` (notice the case)? You're okay with that?

> I'll change ASP.NET Core to use the same for the default (IsDevelopment, IsProduction) so it matches

If I understand you correctly, we need to change the ASP.Net core code so that it _correctly_ sets the `.Environment` setting to be _case sensitive_ (instead of just taking the default value)?

> Would you be willing to file a PR for that too?

Yep.

test/Sentry.Tests/Internals/MainSentryEventProcessorTests.cs
PureKrome: I don't understand the suggestion here? I was trying to check that the word is "the same" but the casing different .. which means, we don't override and accept the user-defined value.

The comment I made there is wrong though. I'll need to fix it to: // Provided: nothing will change. (value has a lower case 'p', different to default value)

src/Sentry/Internal/MainSentryEventProcessor.cs
PureKrome: Totally. can do. 👍🏻

PureKrome:
@bruno-garcia 

> 
> What is really important here is that this code stops you from setting the `Environment` to
> 
> * `null` (e.g. `options.Environment = null;`)
> * `string.Empty` (e.g. `options.Environment = "";`)
> * strings with only 1+ white spaces (e.g. `options.Environment = "        ";`)
> 
> So this means that (from now on) there will _always_ be a value for the `options.Environment`. Not sure if this is what you want, etc.

Can you take a moment to think about this question? I just want to make sure it's been considered.

bruno-garcia:
@PureKrome you can opt out on BeforeSend since that runs after the event processors

test/Sentry.Tests/Internals/MainSentryEventProcessorTests.cs
bruno-garcia: there's nothing wrong here, I just suggested above we use `production` instead of `Production` so your test here would need to flip: `Production` would mean user-defined value, because `production` will be the new default

bruno-garcia:
> @PureKrome you can opt out on BeforeSend since that runs after the event processors


To clarify:

`SentrySdk.Init(o => o.BeforeSend e => { e.Environment = null; return e; });` this would make sure there's no env set. Right?

test/Sentry.Tests/Internals/MainSentryEventProcessorTests.cs
PureKrome: Yep - totally. I misread. yep. 💯

test/Sentry.AspNetCore.Tests/IntegrationMockedBackgroundWorker.cs
PureKrome: Will remove comment.

PureKrome: :wrench:

src/Sentry/Internal/MainSentryEventProcessor.cs
PureKrome: Now refactored. Env var gets checked (and used) first, then options second, finally default value.

PureKrome:
> `SentrySdk.Init(o => o.BeforeSend e => { e.Environment = null; return e; });` this would make sure there's no env set. Right?

Ah! gotcha! Yep, I think i noticed the `BeforeSend` code does get called after the MSEP method does it's stuff. Kewl.

Could this be documented? not sure if the docs need to get updated, at least with some guidence about the default value and order of setting, IMO.

Check the [Docs about Environments](https://docs.sentry.io/product/sentry-basics/environments/), I couldn't read anything there that suggests that the `Environment` setting is _required_. I'm not sure what the UX would look like, with an empty `Environment` setting, passed along?

bruno-garcia:
@PureKrome we've updated the SDK docs for SDK developers to add a default, but user docs wouldn't need to be changed in that way. We could/should document that a default is indeed sent in some SDKs, with the value `production`. @PeloWriter reached out to discuss this, I'm not sure though what will be the approach. Maybe we'll document one by one? The SDKs that have this default?

bruno-garcia (approved):
Excellent! Thanks a lot @PureKrome !

PureKrome:
(I'll leave the docs answer, to others :)   )

PR #552, opened 2020-10-16
📦 Include symbols (snupkg's) when creating nuget packages.

### LINKED ISSUE #530: Publish snupkg to nuget.org
Even though currently the pdb is inside the nupkg, it doesn't mean it gets published with the final app.
A debugger (i.e VS) by convention will hit a symbol server to look for those. Lets publish our symbols as snupkg) to nuget.org when we publish the packages.

Include sources in the debug package.

PureKrome:
Fixes #530 

While reviewing the code, I've noticed that the CI system is using something called [Zeus](https://zeus.ci/) for release management. So i'm not familiar with that project (sponsored by Sentry, btw 👏🏻 ). As such, I'm guessing certain steps and unable to do a full end-to-end test, so I'm not to sure how successful this PR will be/get.

Basic idea I've done:
- `dotnet pack` includes symbols
- push `nupkg` (currently the status quo) and now also the `snupkg` up to zeus 'artifacts'
- fix up the `build.props` to have the [default suggested source-link settings](https://github.com/dotnet/sourcelink#using-source-link-in-net-projects).

I can't see if the `snupkg`'s do end up getting copied to Zeus. Also, I can't see how to 'release' both if they are up in Zeus.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@
 * Rename `LogEntry` to `SentryMessage`. Change type of `SentryEvent.Message` from `string` to `SentryMessage`.
 * Change the type of `Gpu.VendorId` from `int` to `string`.
 * Add support for envelopes.
-* Souce-link integration.
+* Publishing symbols package (snupkg) to nuget.org with sourcelink
 
 # 3.0.0-alpha.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 * Rename `LogEntry` to `SentryMessage`. Change type of `SentryEvent.Message` from `string` to `SentryMessage`.
 * Change the type of `Gpu.VendorId` from `int` to `string`.
 * Add support for envelopes.
+* Souce-link integration.
 
 # 3.0.0-alpha.0
 
--- a/src/Directory.Build.props
+++ b/src/Directory.Build.props
@@ -22,12 +22,18 @@
     <PackageIcon>sentry-nuget.png</PackageIcon>
     <PackageProjectUrl>https://sentry.io</PackageProjectUrl>
     <PackageLicenseExpression>MIT</PackageLicenseExpression>
-    <AllowedOutputExtensionsInPackageBuildOutputFolder>$(AllowedOutputExtensionsInPackageBuildOutputFolder);.pdb</AllowedOutputExtensionsInPackageBuildOutputFolder>
 
     <PackageReleaseNotes>Can be found at: https://github.com/getsentry/sentry-dotnet/releases</PackageReleaseNotes>
 
     <!-- Prop required by Microsoft.SourceLink.GitHub -->
+    <!-- Optional: Publish the repository URL in the built .nupkg (in the NuSpec <Repository> element) -->
+    <PublishRepositoryUrl>true</PublishRepositoryUrl>
+    <!-- Optional: Embed source files that are not tracked by the source control manager in the PDB -->
+    <EmbedUntrackedSources>true</EmbedUntrackedSources>
+    <!-- Optional: Build symbol package (.snupkg) to distribute the PDB containing Source Link -->
+    <IncludeSymbols>true</IncludeSymbols>
     <IncludeSourceRevisionInInformationalVersion>false</IncludeSourceRevisionInInformationalVersion>
+
     <Nullable>annotations</Nullable>
   </PropertyGroup>
 
--- a/src/Directory.Build.props
+++ b/src/Directory.Build.props
@@ -32,6 +32,7 @@
     <EmbedUntrackedSources>true</EmbedUntrackedSources>
     <!-- Optional: Build symbol package (.snupkg) to distribute the PDB containing Source Link -->
     <IncludeSymbols>true</IncludeSymbols>
+    <SymbolPackageFormat>snupkg</SymbolPackageFormat>
     <IncludeSourceRevisionInInformationalVersion>false</IncludeSourceRevisionInInformationalVersion>
 
     <Nullable>annotations</Nullable>
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -30,6 +30,7 @@ dotnet_csproj:
   package_version: "{version}"
 artifacts:
   - path: '**\*.nupkg'
+  - path: '**\*.snupkg'
 on_success:
   - ps: |
       $env:PATH = 'C:\msys64\usr\bin;' + $env:PATH
--- a/Sentry.sln
+++ b/Sentry.sln
@@ -99,6 +99,19 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "scripts", "scripts", "{3454
 		scripts\zeus.sh = scripts\zeus.sh
 	EndProjectSection
 EndProject
+Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = ".github", ".github", "{F267BFB9-C5D5-4018-9D3F-B55443702AEC}"
+	ProjectSection(SolutionItems) = preProject
+		.github\CODEOWNERS = .github\CODEOWNERS
+		.github\ISSUE_TEMPLATE.md = .github\ISSUE_TEMPLATE.md
+		.github\release-drafter.yml = .github\release-drafter.yml
+	EndProjectSection
+EndProject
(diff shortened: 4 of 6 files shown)


CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 * Rename `LogEntry` to `SentryMessage`. Change type of `SentryEvent.Message` from `string` to `SentryMessage`.
 * Change the type of `Gpu.VendorId` from `int` to `string`.
 * Add support for envelopes.
+* Souce-link integration.
```
bruno-garcia, line 6: We already had sourcelink but pdbs were only shipped inside the `nupkg`.
Maybe:
```suggestion
* Publishing symbols package (snupkg) to nuget.org with sourcelink
```

src/Directory.Build.props
```diff
@@ -22,12 +22,18 @@
...
     <PackageLicenseExpression>MIT</PackageLicenseExpression>
-    <AllowedOutputExtensionsInPackageBuildOutputFolder>$(AllowedOutputExtensionsInPackageBuildOutputFolder);.pdb</AllowedOutputExtensionsInPackageBuildOutputFolder>
 
     <PackageReleaseNotes>Can be found at: https://github.com/getsentry/sentry-dotnet/releases</PackageReleaseNotes>
 
     <!-- Prop required by Microsoft.SourceLink.GitHub -->
+    <!-- Optional: Publish the repository URL in the built .nupkg (in the NuSpec <Repository> element) -->
+    <PublishRepositoryUrl>true</PublishRepositoryUrl>
+    <!-- Optional: Embed source files that are not tracked by the source control manager in the PDB -->
+    <EmbedUntrackedSources>true</EmbedUntrackedSources>
```
bruno-garcia, line 32: What would such files be? This was left out I believe because I didn't understand the value (or couldn't verify it would add value) when I originally added it

src/Directory.Build.props
```diff
@@ -22,12 +22,18 @@
...
+    <!-- Optional: Build symbol package (.snupkg) to distribute the PDB containing Source Link -->
+    <IncludeSymbols>true</IncludeSymbols>
```
bruno-garcia, line 34: Is this needed if the cli paramters are passed? Or do we need the CLI parameters then?
Maybe the csproj alone is enough?

src/Directory.Build.props
```diff
@@ -22,12 +22,18 @@
     <PackageIcon>sentry-nuget.png</PackageIcon>
     <PackageProjectUrl>https://sentry.io</PackageProjectUrl>
     <PackageLicenseExpression>MIT</PackageLicenseExpression>
-    <AllowedOutputExtensionsInPackageBuildOutputFolder>$(AllowedOutputExtensionsInPackageBuildOutputFolder);.pdb</AllowedOutputExtensionsInPackageBuildOutputFolder>
 
     <PackageReleaseNotes>Can be found at: https://github.com/getsentry/sentry-dotnet/releases</PackageReleaseNotes>
 
     <!-- Prop required by Microsoft.SourceLink.GitHub -->
+    <!-- Optional: Publish the repository URL in the built .nupkg (in the NuSpec <Repository> element) -->
+    <PublishRepositoryUrl>true</PublishRepositoryUrl>
```
bruno-garcia, line 30: Is this true by default? I wonder because sourcelink already worked on this repo

bruno-garcia (commented):
Thanks for tackling this one too @PureKrome 
Left a couple notes

PureKrome:
Ta for the feedback, I'll review it later on after i wake up. (nearly 2am here).

here's some results from the draft PR CI/CD..

Okay - so CI passes with some good results:

`snupkg`'s are getting built:
image

`nupkg`'s and `snupkg`'s are also getting collected as artifacts:
image

... and getting uploaded into the artifact storage:
image

and finally, uploaded to Zeus (if the zeus hook variable thingy was set ... which I have no idea why it was not .. i'm guessing because the branch was not `main` or `release/*` ?
image

So I hope this helps?

bruno-garcia:
All looking great @PureKrome 
My notes are just questions, could lift this from draft already eh?
Thanks again for your help and sleep tight 🛌 🇦🇺

src/Directory.Build.props
PureKrome: Tested with just having this (and the `SymbolPackageFormat` setting) in the `csproj` instead of the CLI. Worked. Cheers!

src/Directory.Build.props
PureKrome: I added a new comment in the original issue #530 with some screen shots about this. 

If I set that to `true` or `false`, it still says 'untracked'. Hmmm.

When i add the `ContinuousIntegrationBuild` value in .. i can get better results.

`ContinuousIntegrationBuild`added + `false`

image

`ContinuousIntegrationBuild` added + `true`

image


but if i don't add the `ContinuousIntegrationBuild` value... 

image


---

TL;DR;

To get NuGet package explorer to be 'happy', I need:
- ✔️ `<EmbedUntrackedSources>true</EmbedUntrackedSources>`
- ✔️ `<ContinuousIntegrationBuild>true</ContinuousIntegrationBuild>`

I'm honestly not too sure what these are doing and how they make things better/worse. Or in other words, what effect does it have when i DO NOT include them? Sure i know NuGet Package Explorer cries .. but what is happening behind the scenes?

PureKrome: Got some info on the [CIB setting](https://github.com/clairernovotny/DeterministicBuilds#deterministic-builds).

> This repo shows how to get a fully deterministic build using Source Link. Deterministic builds are important as they enable verification that the resulting binary was built from the specified source and provides traceability.
>
>Deterministic builds require a property to be set to true during CI: ContinuousIntegrationBuild. These should not be enabled during local dev or the debugger won't be able to find the local source files.

So i'll try and make sure the CIB setting is only set via the CLI (because we're using various CI environments right now)

PureKrome: Also, -FINALLY- stumbled across the [docs about the settings](https://github.com/dotnet/sourcelink/blob/master/docs/README.md#embeduntrackedsources).

> Set EmbedUntrackedSources to true to instruct the build system to embed project source files that are not tracked by the source control or imported from a source package to the generated PDB.
>
>If the project generates additional source files that are added to Compile item group in a custom target, this target must run before BeforeCompile target (specify BeforeTargets="BeforeCompile"). Otherwise, these additional source files will not be automatically embedded into the PDB.

Again, I don't really understand what this means with respect to Sentry / Sentry's project system :(

I'm going to leave it IN .. as Nuget Package Manager was having a cry, if it was `false`.

src/Directory.Build.props
PureKrome: Another great question. I found [the docs on this](https://github.com/dotnet/sourcelink/blob/master/docs/README.md#publishrepositoryurl): (Emphasis, mine)

> The URL of the repository supplied by the CI server or retrieved from source control manager is stored in PrivateRepositoryUrl variable.
>
>This value is not directly embedded in build outputs to avoid inadvertently publishing links to private repositories. Instead, PublishRepositoryUrl *needs to be set by the project* in order to publish the URL into RepositoryUrl property, which is used e.g. in the nuspec file generated for NuGet package produced by the project.

To me, this suggests that the default is `FALSE` .. but if that's the case ... how does this work?

I've removed it (locally) and generated a new package .. and  .. it lists the url in there!

I .. don't get it :(

I _think_ this value is _different_ to `<RepositoryUrl>https://github.com/getsentry/sentry-dotnet</RepositoryUrl>` which already exists.

So maybe:
- `<RepositoryUrl>` can be anything, which is meta data for someone to 'know' where the repo is if they want to _manually_ look at code
- `<PublishRepositoryUrl>` maybe related to how the symbol packages are known where to exist? 

I'm thinking, based on the docs, I should put it in there .. mainly because this is a PUBLIC repo.

PureKrome: Also, [asked a question](https://github.com/dotnet/sourcelink/issues/659) in the sourcelink repo.

bruno-garcia (approved):
Thanks for the thorough analysis. Let's add it in!

PureKrome:
Yay! Thanks @bruno-garcia.

One last request: could this get the `hacktoberfest-accepted` label added, please sir?

bruno-garcia:
Sorry I forgot @PureKrome

PR #554, opened 2020-10-18
ASPNET.Core default Environment settings are cleaned, if used.

### LINKED ISSUE #551: ASPNET.Core should use 'production' or 'development' for their default Environment settings
In #550  the [following suggestion was made ](https://github.com/getsentry/sentry-dotnet/pull/550#discussion_r505574686)by @bruno-garcia :

> I'll change ASP.NET Core to use the same for the default (IsDevelopment, IsProduction) so it matches. Would you be willing to file a PR for that too?

So If understand this comment correctly, the  SENTRY default value for prod/dev are:
- production environment: `production` (lowercase `p`) 
- development environment: `development` (lowercase `d`).

while the [ASPNET_CORE default Environment values are](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/environments?view=aspnetcore-3.1):
- `IsProduction()`: `Production` (uppercase `P`)
- `IsDevelopment()`: `Development` (uppercase `D`).

With the [Sentry ASPNETCORE docs](https://docs.sentry.io/platforms/dotnet/guides/aspnetcore/configuration/environments/) saying that the Environment setting is case _sensitive_, then we should also make sure we setting the correct casing for these two environment settings, _if they are used by the developer_.

**TODO**:
- If the environment setting is falling back to being set with the `ASPNET_ENVIRONMENT` then make sure the case is correctly set to
- `production`
- `development` 

(side note: ✔️ Yep, I can do a PR to this issue, too ❤️ )

PureKrome:
Resolves #551 

NOTES:

The ASPNET.Core project actually targets both NETCOREAPP30 and NS20 (I'm not sure why .....).

> `<TargetFrameworks>netcoreapp3.0;netstandard2.0</TargetFrameworks>`

As such, I've ended up NOT doing a single "if environment == "Production" string check (for both NS20 and NCA30) because NETCOREAPP30+ now [includes this string as a constant](https://docs.microsoft.com/en-us/dotnet/api/microsoft.extensions.hosting.environments.production?view=dotnet-plat-ext-3.1). Yay 🎉 

So I ended up using `#if NETSTANDARD2_0` etc, preprocessor directives. It makes the code a bit more messy, so i'm totally happy to strip that out and just have the a simple string check for both, instead of a string check and a const string check.

Finally, not sure if the test I added was into the correct file, etc. So happy to get told otherwise.


PureKrome:
@bruno-garcia Drats - the test failed, yet it works on my machine. Is there any way I can see _what_ the failure is?

(BTW, I can't run any NET461 or NC21 tests on my machine. I only have NetCore3+ installed)

image

Not sure if the failing test occurs there/then?

src/Sentry.AspNetCore/SentryAspNetCoreOptionsSetup.cs
```diff
@@ -25,10 +25,56 @@ public override void Configure(SentryAspNetCoreOptions options)
...
+                    //       .NET Core sets the ENV variable to 'Production' (upper case P) or
+                    //       'Development' (upper case D) which conflicts with the Sentry recommendation.
+                    //       As such, we'll be kind and override those values, here ... if applicable.
+                    // Assumption: The Hosting Environment is always set.
+                    //             If not set by a developer, then the framework will auto set it.
+                    //             Alternatively, developers might set this to a CUSTOM value, which we
+                    //             need to respect (especially the case-sensitivity).
+                    //             REF: https://docs.microsoft.com/en-us/aspnet/core/fundamentals/environments
+#if NETSTANDARD2_0
+                    if (_hostingEnvironment.Equals("Production"))
```
xt0rted, line 49: Shouldn't this be

```suggestion
                    if (_hostingEnvironment.EnvironmentName.Equals("Production"))
```

@PureKrome the failing tests create annotations in the [diff view](https://github.com/getsentry/sentry-dotnet/pull/554/files), if you scroll down you'll see which ones are failing along with their message.

test/Sentry.AspNetCore.Tests/SentryAspNetCoreOptionsSetupTests.cs
```diff
@@ -39,5 +39,49 @@ public void Filters_KestrelEventId1_WithException_NotFiltered()
...
+        public void Filters_Environment_SentryEnvironment_Set(string environment)
+        {
+            // Arrange.
+            Environment.SetEnvironmentVariable(Internal.Constants.EnvironmentEnvironmentVariable, environment);
+
+            // Act.
+            _sut.Configure(_target);
+
+            // Assert.
+            Assert.Equal(environment, _target.Environment);
```
xt0rted, line 84: Give this a try. Switching to this got everything working on my machine.

```suggestion
            // Arrange.
            EnvironmentVariableGuard.WithVariable(Internal.Constants.EnvironmentEnvironmentVariable, environment, () => {
                // Act.
                _sut.Configure(_target);

                // Assert.
                Assert.Equal(environment, _target.Environment);
            });
```

test/Sentry.AspNetCore.Tests/SentryAspNetCoreOptionsSetupTests.cs
```diff
@@ -39,5 +39,50 @@ public void Filters_KestrelEventId1_WithException_NotFiltered()
...
+            Assert.Equal(expectedEnvironment, _target.Environment);
+        }
+
+        [Theory]
+        [InlineData("foo")] // Random setting.
+        [InlineData("Production")] // Custom setting which is the same as ASPNET_ENVIRONMENT. But because this is manually set, don't change it.
+        public void Filters_Environment_SentryEnvironment_Set(string environment)
+        {
+            // Arrange.
+            EnvironmentVariableGuard.WithVariable(Internal.Constants.EnvironmentEnvironmentVariable, environment, () =>
```
xt0rted, line 78: @PureKrome I forgot to mention you also need to add `using Sentry.Testing;` to use this helper here. Sorry about that.

PureKrome: Yeah, i had to manually add that in, at the top. NP.

PureKrome:
@bruno-garcia as a side note to this PR, i don't understand how my PR dropped code coverage by 8% 😢 

I'd love to know how/why so I can make sure I'm covering the code I just created.

(note: I did click on the file in the codecoverage report above and not sure how to make it out :/ )

Tyrrrz:
> The ASPNET.Core project actually targets both NETCOREAPP30 and NS20 (I'm not sure why .....).

That's mainly because NS20 doesn't have nullable type annotations on BCL.

src/Sentry.AspNetCore/SentryAspNetCoreOptionsSetup.cs
```diff
@@ -25,10 +25,52 @@ public override void Configure(SentryAspNetCoreOptions options)
...
+            {
+                var locatedEnvironment = EnvironmentLocator.Locate();
+                if (!string.IsNullOrWhiteSpace(locatedEnvironment))
+                {
+                    // Sentry specific environment takes precedence #92.
+                    options.Environment = locatedEnvironment;
+                }
+                else
+                {
+                    // NOTE: Sentry prefers to have it's environment setting to be all lower case.
```
Tyrrrz, line 39: Nit: `it's` -> `its`.

src/Sentry.AspNetCore/SentryAspNetCoreOptionsSetup.cs
```diff
@@ -25,10 +25,52 @@ public override void Configure(SentryAspNetCoreOptions options)
...
+#else
+                    if (_hostingEnvironment.EnvironmentName.Equals(Microsoft.Extensions.Hosting.Environments.Production))
+                    {
+                        options.Environment = Internal.Constants.DefaultEnvironmentSetting;
+                    }
+                    else if (_hostingEnvironment.EnvironmentName.Equals(Microsoft.Extensions.Hosting.Environments.Development))
+                    {
+                        options.Environment = Microsoft.Extensions.Hosting.Environments.Development.ToLower();
+                    }
+#endif
```
Tyrrrz, line 66: To reduce essential nesting, I would do:

```csharp
public static class Constants
{
    public static string ProductionEnvironmentName =>
        #if NETSTANDARD2_0
             "Production";
        #else
             Microsoft.Extensions.Hosting.Environments.Production;
        #endif
}
```

and then...

```csharp
if (_hostingEnvironment.EnvironmentName.Equals(Constants.ProductionEnvironmentName)
{
    // ...
}
```

src/Sentry.AspNetCore/SentryAspNetCoreOptionsSetup.cs
```diff
@@ -25,10 +25,52 @@ public override void Configure(SentryAspNetCoreOptions options)
...
+                    //       .NET Core sets the ENV variable to 'Production' (upper case P) or
+                    //       'Development' (upper case D) which conflicts with the Sentry recommendation.
+                    //       As such, we'll be kind and override those values, here ... if applicable.
+                    // Assumption: The Hosting Environment is always set.
+                    //             If not set by a developer, then the framework will auto set it.
+                    //             Alternatively, developers might set this to a CUSTOM value, which we
+                    //             need to respect (especially the case-sensitivity).
+                    //             REF: https://docs.microsoft.com/en-us/aspnet/core/fundamentals/environments
+#if NETSTANDARD2_0
+                    if (_hostingEnvironment.EnvironmentName.Equals("Production"))
```
Tyrrrz, line 49: Should we consider case sensitivity here? Or maybe use `StringComparison.OrdinalIgnoreCase`?

src/Sentry.AspNetCore/SentryAspNetCoreOptionsSetup.cs
```diff
@@ -25,10 +25,52 @@ public override void Configure(SentryAspNetCoreOptions options)
...
+                        options.Environment = "development";
+                    }
+#else
+                    if (_hostingEnvironment.EnvironmentName.Equals(Microsoft.Extensions.Hosting.Environments.Production))
+                    {
+                        options.Environment = Internal.Constants.DefaultEnvironmentSetting;
+                    }
+                    else if (_hostingEnvironment.EnvironmentName.Equals(Microsoft.Extensions.Hosting.Environments.Development))
+                    {
+                        options.Environment = Microsoft.Extensions.Hosting.Environments.Development.ToLower();
```
Tyrrrz, line 64: `ToLowerInvariant()` should be more consistent as it's not culture-dependent.

src/Sentry.AspNetCore/SentryAspNetCoreOptionsSetup.cs
```diff
@@ -25,10 +25,52 @@ public override void Configure(SentryAspNetCoreOptions options)
...
+                    if (_hostingEnvironment.EnvironmentName.Equals("Production"))
+                    {
+                        options.Environment = Internal.Constants.DefaultEnvironmentSetting;
+                    }
+                    else if (_hostingEnvironment.EnvironmentName.Equals("Development"))
+                    {
+                        options.Environment = "development";
+                    }
+#else
+                    if (_hostingEnvironment.EnvironmentName.Equals(Microsoft.Extensions.Hosting.Environments.Production))
```
bruno-garcia, line 58: can we have a `using HostingEnvironments = Microsoft.Extensions.Hosting.Environments`?

src/Sentry.AspNetCore/SentryAspNetCoreOptionsSetup.cs
bruno-garcia, line 64:
```suggestion
                        options.Environment = Microsoft.Extensions.Hosting.Environments.Development.ToLowerInvariant();
```

bruno-garcia (commented):
great stuff!

src/Sentry.AspNetCore/SentryAspNetCoreOptionsSetup.cs
PureKrome: Great suggestion! i'll update!

src/Sentry.AspNetCore/SentryAspNetCoreOptionsSetup.cs
PureKrome: :wrench:

src/Sentry.AspNetCore/SentryAspNetCoreOptionsSetup.cs
PureKrome: 🔧

src/Sentry.AspNetCore/SentryAspNetCoreOptionsSetup.cs
PureKrome: :wrench:

Fixed by not requiring this namespace as i've moved the constant strings out to another class.

src/Sentry.AspNetCore/SentryAspNetCoreOptionsSetup.cs
PureKrome: I did case sensitivity to reduce cycles. Why set the value if it's already the exact same?

Tyrrrz (approved):

PureKrome:
sadness 😢 AppVeyor is dying cause of some transient error :(

```
C:\projects\sentry-dotnet\.tmp\dotnet\sdk\5.0.100-preview.8.20417.9\NuGet.targets(128,5): error : Unable to load the service index for source https://api.nuget.org/v3/index.json. [C:\projects\sentry-dotnet\Sentry.sln]
C:\projects\sentry-dotnet\.tmp\dotnet\sdk\5.0.100-preview.8.20417.9\NuGet.targets(128,5): error :   No connection could be made because the target machine actively refused it. (api.nuget.org:443) [C:\projects\sentry-dotnet\Sentry.sln]
C:\projects\sentry-dotnet\.tmp\dotnet\sdk\5.0.100-preview.8.20417.9\NuGet.targets(128,5): error :   No connection could be made because the target machine actively refused it. [C:\projects\sentry-dotnet\Sentry.sln]
Command exited with code 1
bash scripts\zeus.sh report_failed

```

bruno-garcia:
rerunning the build.
We'll hopefully soon move over to GHA completely

bruno-garcia (approved):
Good stuff!

bruno-garcia:
Thanks again @PureKrome !

PureKrome:
My pleasure 👍🏻

PR #555, opened 2020-10-19
Add most popular libraries to InAppExclude

bruno-garcia:
Frames marked `InApp=true` are considered when grouping issues in Sentry.
To help improve grouping, excluding a few more namespaces that are known to be external packages


src/Sentry/SentryOptions.cs
```diff
@@ -401,6 +401,30 @@ public SentryOptions()
...
+                    "Polly",
+                    "Swashbuckle",
+                    "Fluentassertions",
+                    "Autofac",
+                    "Stackexchange.Redis",
+                    "Dapper",
+                    "RestSharp",
+                    "SkiaSharp",
+                    "IdentityModel",
+                    "SqlitePclRaw",
```
bruno-garcia, line 418:
```suggestion
                    "SqlitePclRaw",
                    "Xamarin",
```

bruno-garcia:
@Tyrrrz did the tests flakey up now? Seem to break at random

Tyrrrz:
@bruno-garcia yes, AppVeyor did:

image

Not a test issue.

src/Sentry/SentryOptions.cs
```diff
@@ -401,6 +401,31 @@ public SentryOptions()
...
                     "Giraffe.",
                     "NLog",
+                    "Npgsql",
+                    "RabbitMQ",
+                    "Hangfire",
+                    "IdentityServer4",
+                    "AWSSDK",
+                    "Polly",
+                    "Swashbuckle",
+                    "Fluentassertions",
```
Tyrrrz, line 411: Why would someone use Fluentassertions in an application? Should it be FluentValidation?

Tyrrrz (approved):

Tyrrrz:
Is the purpose of this feature something like "show only my code"? Maybe in that case it's possible to implement some smarter heuristic in the future, instead of listing every single library in existence?

Tyrrrz:
Ok. one test did appear to fail, didn't notice that at first.

image

Judging from the stack trace, the error comes from the constructor of `Sentry.Internal.Hub`. Which shouldn't be called in this test because DSN is empty:

image

My guess is that the tests might be running in parallel and there's a race condition where some other test may set DSN environment variable, which gets read in this test in `var dsn = options.Dsn ?? DsnLocator.FindDsnStringOrDisable();`, resulting in the wrong behavior.

src/Sentry/SentryOptions.cs
bruno-garcia: yes

bruno-garcia:
@Tyrrrz yeah seems something got flakey

> FluentValidation

Will fix it

> yeah something with Assembly.GetEntrypoint().Namepsace of sorts but it comes with other down sides.

Would be worth a try if you're willing to give it a shot.

It's not *just* for the "just my code" feature, but also affects grouping. We only consider "InApp" frames when grouping issues, so those async await (MoveNext) frames in between shouldnt play a role, otherwise it gets badly grouped due to often differnt generated code frames in the middle

src/Sentry/SentryOptions.cs
bruno-garcia, line 411:
```suggestion
                    "FluentValidation",
```

PR #558, opened 2020-10-23
Add support for envelope deserialization

Tyrrrz:


src/Sentry/Internal/Extensions/DisposableExtensions.cs
```diff
@@ -0,0 +1,30 @@
...
+                {
+                    exceptions.Add(ex);
+                }
+            }
+
+            if (exceptions.Count > 0)
+            {
+                throw new AggregateException(exceptions);
+            }
+        }
```
bruno-garcia, line 28: Happy path == 0 allocations. Overkill?

```suggestion
            List<Exception>? exceptions = null;

            foreach (var i in disposables)
            {
                try
                {
                    i.Dispose();
                }
                catch (Exception ex)
                {
                	exception =?? new List<Exception>();
                    exceptions.Add(ex);
                }
            }

            if (exceptions?.Count > 0)
            {
                throw new AggregateException(exceptions);
            }
        }
```

src/Sentry/Protocol/Envelope.cs
```diff
@@ -75,5 +79,24 @@ public static Envelope FromEvent(SentryEvent @event)
...
+            Stream stream,
+            CancellationToken cancellationToken = default)
+        {
+            // Header
+            var header = await Json.DeserializeFromStreamAsync<Dictionary<string, object>>(stream, cancellationToken)
+                .ConfigureAwait(false);
+
+            // Items
+            var items = new List<EnvelopeItem>();
+            while (stream.ReadByte() == (byte)'\n')
```
bruno-garcia: Consider there might be a trailing `\n` and no more items and that should be OK (as in don't throw)

src/Sentry/Protocol/EnvelopeItem.cs
```diff
@@ -142,5 +146,24 @@ public static EnvelopeItem FromEvent(SentryEvent @event)
...
+
+        public static async Task<EnvelopeItem> DeserializeAsync(
+            Stream stream,
+            CancellationToken cancellationToken = default)
+        {
+            // Header
+            var header = await Json.DeserializeFromStreamAsync<Dictionary<string, object>>(stream, cancellationToken)
+                .ConfigureAwait(false);
+
+            var length = (long)header.GetValueOrDefault(LengthKey, long.MaxValue);
```
bruno-garcia: The `long.MaxValue` scares me. What's the use case/side effect for the default value?

bruno-garcia (commented):
First pass: Going great!

src/Sentry/Protocol/EnvelopeItem.cs
Tyrrrz: It's just a hack to get things working fast.

Tyrrrz:
@bruno-garcia very crude, but it works (at least on my tests).

Points of interest:
- I gave up using `StreamReader` as it's way too complicated with mixed content and just resorted to reading one byte at a time. This may result in actually worse performance than if we implemented all of this synchronously due to `Task` allocation overhead, especially if you consider that reading from file system (which is where envelopes gonna get stored) is very fast. Need to measure if this is a concern.
- As you can see, I read the header block by first buffering everything until `\n` in memory and then feeding it to Json.NET. This means that if, for some reason, we are fed invalid envelope without header (even empty header is fine) then we might accidentally put it all in memory.  You might wonder why I don't deserialize directly from stream. Well that's because, just like with `StreamReader`, when you give `JsonReader` a stream, it thinks it now fully belongs to it and starts messing with position. Also, in case with invalid envelope, Json.NET would also end up loading everything in memory due to the way it works (it can't deserialize from stream, only read from stream, but it needs to buffer the content to deserialize).
- Finally, like we discussed, this brings a few constraints on the usage of `Envelope`. Once you call `Envelope.DeserializeAsync(...)` on a stream, that stream must not be disposed. Additionally, that stream must be seekable too. Breaking any of those things will throw in current implementation. I think we need to re-think the design of `Envelope` to underline that the stream merges its lifetime with envelope, maybe `new Envelope(Stream ...)` should make that more clear, but it might complicate implementation.

Anyway, it's pretty clear that this most likely needs some critical redesign. Ultimately, I would prefer to also make `Envelope` and others internal and don't expose to end-users. Looking forward to your input.

Tyrrrz:
I also think I should add more unit tests (gasp) for `PartialStream` because the logic is pretty intense. But if we may end up redesigning, then it's likely we won't need it.

src/Sentry/Protocol/Envelope.cs
Tyrrrz: Reworked that entire part

src/Sentry/Internal/Extensions/StreamExtensions.cs
```diff
@@ -0,0 +1,34 @@
...
+
+namespace Sentry.Internal.Extensions
+{
+    internal static class StreamExtensions
+    {
+        public static async Task<int> ReadByteAsync(
+            this Stream stream,
+            CancellationToken cancellationToken = default)
+        {
+            using var buffer = new PooledBuffer<byte>(1);
```
bruno-garcia, line 13: What about stackalloc instead? Assign to Span and won't need unsafe.

Tyrrrz: Would it be beneficial? With pooling we will completely avoid allocations.

bruno-garcia:
> * I gave up using `StreamReader` as it's way too complicated with mixed content and just resorted to reading one byte at a time. This may result in actually worse performance than if we implemented all of this synchronously due to `Task` allocation overhead, especially if you consider that reading from file system (which is where envelopes gonna get stored) is very fast. Need to measure if this is a concern.

That's fine. Though I assume it would be a `ValueTask` and wouldn't allocate if the buffer had more data. The inner stream wouldn't be buffering a byte at a time.

> * As you can see, I read the header block by first buffering everything until `\n` in memory and then feeding it to Json.NET. This means that if, for some reason, we are fed invalid envelope without header (even empty header is fine) then we might accidentally put it all in memory.  

That'd be a bug and it's not really a big problem to read the whole stream into memory. We could cap the buffer and if we read more than X (500 KB?) header data we log and error and bail.

> You might wonder why I don't deserialize directly from stream. Well that's because, just like with `StreamReader`, when you give `JsonReader` a stream, it thinks it now fully belongs to it and starts messing with position. Also, in case with invalid envelope, Json.NET would also end up loading everything in memory due to the way it works (it can't deserialize from stream, only read from stream, but it needs to buffer the content to deserialize).

I'm replying this before I review. But I guess we could solve by creating a Stream and only letting it read up to `length`?

> * Finally, like we discussed, this brings a few constraints on the usage of `Envelope`. Once you call `Envelope.DeserializeAsync(...)` on a stream, that stream must not be disposed. Additionally, that stream must be seekable too. Breaking any of those things will throw in current implementation. I think we need to re-think the design of `Envelope` to underline that the stream merges its lifetime with envelope, maybe `new Envelope(Stream ...)` should make that more clear, but it might complicate implementation.

Stream has a `CanSeek`, it's part of the Stream abstraction that it can or not Read/Write/Seek

> Anyway, it's pretty clear that this most likely needs some critical redesign. Ultimately, I would prefer to also make `Envelope` and others internal and don't expose to end-users. Looking forward to your input.

Will review

src/Sentry/Protocol/Envelope.cs
```diff
@@ -75,5 +80,45 @@ public static Envelope FromEvent(SentryEvent @event)
...
+
+        /// <summary>
+        /// Deserializes envelope from stream.
+        /// </summary>
+        public static async Task<Envelope> DeserializeAsync(
+            Stream stream,
+            CancellationToken cancellationToken = default)
+        {
+            // Header
+            var header = await DeserializeHeaderAsync(stream, cancellationToken).ConfigureAwait(false);
```
bruno-garcia, line 111: Check if we got something back here? If we got null the thing is broken and we should bail.
Oh I guess this doesn't return null, right? it'll be an empty map.

bruno-garcia (approved):
Looking good! Dropped a couple notes

bruno-garcia (commented):
Looking good!

src/Sentry/Internal/Extensions/StreamExtensions.cs
Tyrrrz: Another concern is that it's an async method, so it might not be possible to use. I will take a look and try.

Tyrrrz: image

Looks like it can't. And `stackalloc` can't be assigned to `Memory<byte>`.

Tyrrrz:
> That's fine. Though I assume it would be a ValueTask and wouldn't allocate if the buffer had more data. The inner stream wouldn't be buffering a byte at a time.

We'd have to change the whole call chain to `ValueTask` as well to benefit from this, if I understand everything correctly.

> That'd be a bug and it's not really a big problem to read the whole stream into memory. We could cap the buffer and if we read more than X (500 KB?) header data we log and error and bail.

Can definitely do that.

> I'm replying this before I review. But I guess we could solve by creating a Stream and only letting it read up to length?

Probably, but likely no point in that. JSON.NET still does deserialization synchronously under the hood, so it will likely not provide any benefits to us.

> Stream has a CanSeek, it's part of the Stream abstraction that it can or not Read/Write/Seek

Yes, that is not the concern. My concern is in the API design, because when I see `Deserialize` I think that the input stream will be completely mapped to an in-memory representation. In our case, however, it actually does something completely different where it builds an abstraction on top of the stream and takes ownership of the stream itself. I just think a different name would probably be better, but maybe even use a constructor to emphasize that the stream is an integral part of the envelope. Of course, if we make all of this internal, it won't be a concern. Before we release, we can go through the types and mark as many of them as possible as internal (Envelope, for one is not exposed outside of Sentry).

src/Sentry/Internal/Json.cs
```diff
@@ -26,12 +26,22 @@ internal static class Json
             new StreamWriter(stream, Encoding, 1024, true)
         );
 
+        private static JsonTextReader CreateReader(Stream stream) => new JsonTextReader(
+            new StreamReader(stream, Encoding, false, 1024, true)
```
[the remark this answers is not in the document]
Tyrrrz: 1024 is also the default on `StreamReader` when you initialize it with another overload.

As for `stackalloc`, as mentioned, it doesn't actually work in async methods (I presume since the resulting method gets compiled into a state machine, with the state stored on the heap).

src/Sentry/Protocol/Envelope.cs
Tyrrrz: Good point. It does return `null` because Json.NET is nice like that.

Tyrrrz: Should `null` be an exception or coerced to an empty map?

src/Sentry/Internal/Extensions/StreamExtensions.cs
bruno-garcia: Fuck

src/Sentry/Internal/Json.cs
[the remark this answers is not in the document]
bruno-garcia: Fair about the default of the reader.

bummer with the stackalloc lets see what else we can do.

src/Sentry/Protocol/Envelope.cs
bruno-garcia: Good question. I dunno. `null` kind of sucks, if the header is invalid then maybe we should disconsider the payload instead. Empty map will lead us to keep going and try to read the rest of the stream as envelope item

bruno-garcia:
> We'd have to change the whole call chain to ValueTask as well to benefit from this, if I understand everything correctly.


Good point, we should do that actually. I believe if we're not cachign `Task` we should be defaulting to `ValueTask` these days

bruno-garcia:
> Probably, but likely no point in that. JSON.NET still does deserialization synchronously under the hood, so it will likely not provide any benefits to us.

Fair

bruno-garcia:
> Yes, that is not the concern. My concern is in the API design, because when I see Deserialize I think that the input stream will be completely mapped to an in-memory representation. In our case, however, it actually does something completely different where it builds an abstraction on top of the stream and takes ownership of the stream itself. I just think a different name would probably be better, but maybe even use a constructor to emphasize that the stream is an integral part of the envelope. Of course, if we make all of this internal, it won't be a concern. Before we release, we can go through the types and mark as many of them as possible as internal (Envelope, for one is not exposed outside of Sentry).


Makes sense to me.

Tyrrrz:
> Good point, we should do that actually. I believe if we're not cachign Task we should be defaulting to ValueTask these days

I agree. I will add a project item and do that after this PR.

src/Sentry/Protocol/Envelope.cs
Tyrrrz: Changed it to exception, looks cleaner now

src/Sentry/Protocol/Envelope.cs
```diff
@@ -97,7 +102,9 @@ public static Envelope FromEvent(SentryEvent @event)
                 lastByte = await stream.ReadByteAsync(cancellationToken).ConfigureAwait(false);
             }
 
-            return Json.DeserializeFromByteArray<Dictionary<string, object>>(buffer.ToArray());
+            return
+                Json.DeserializeFromByteArray<Dictionary<string, object>?>(buffer.ToArray()) ??
+                throw new InvalidOperationException("Envelope header is malformed.");
```
bruno-garcia, line 107: Could we keep the operator on the left, I always get confused why the lines are stacked.
```suggestion
                Json.DeserializeFromByteArray<Dictionary<string, object>?>(buffer.ToArray())
                  ?? throw new InvalidOperationException("Envelope header is malformed.");
```

src/Sentry/Protocol/EnvelopeItem.cs
```diff
@@ -163,7 +169,9 @@ public static EnvelopeItem FromEvent(SentryEvent @event)
                 lastByte = await stream.ReadByteAsync(cancellationToken).ConfigureAwait(false);
             }
 
-            return Json.DeserializeFromByteArray<Dictionary<string, object>>(buffer.ToArray());
+            return
+                Json.DeserializeFromByteArray<Dictionary<string, object>?>(buffer.ToArray()) ??
+                throw new InvalidOperationException("Envelope item header is malformed.");
```
bruno-garcia, line 174:
```suggestion
                Json.DeserializeFromByteArray<Dictionary<string, object>?>(buffer.ToArray())
                  ?? throw new InvalidOperationException("Envelope item header is malformed.");
```

Tyrrrz:
@bruno-garcia I improved performance by using async enumerators. This way the pooled buffer for longer. Also, just to clarify, there are no heap allocations if the pooled buffer already exists and is not used (which is often going to be the case). The type that represents it is a stack-allocated readonly struct.

src/Sentry/Protocol/Envelope.cs
Tyrrrz: image
How about this? Wink wink

Tyrrrz:
@bruno-garcia all yours, enjoy.

src/Sentry/Internal/BackgroundWorker.cs
```diff
@@ -152,6 +152,7 @@ private async Task WorkerAsync()
                         {
                             _ = _queue.TryDequeue(out _);
                             _ = Interlocked.Decrement(ref _currentItems);
+                            envelope.Dispose();
```
bruno-garcia, line 155: This better now throw. Should we guard this with a try/catch?

bruno-garcia:
Will take a look as soon as I can

Tyrrrz:
Should be good.

src/Sentry/Internal/Extensions/StreamExtensions.cs
```diff
@@ -0,0 +1,52 @@
...
+            CancellationToken cancellationToken = default)
+        {
+            using var buffer = new PooledBuffer<byte>(expectedLength);
+            var bytesRead = await stream.ReadAsync(buffer.Array, 0, expectedLength, cancellationToken)
+                .ConfigureAwait(false);
+
+            // The buffer is rented so we can't return it, plus it may be larger than needed.
+            // So we copy everything to a new buffer.
+            var result = new byte[bytesRead];
+            Array.Copy(buffer.Array, result, bytesRead);
```
bruno-garcia, line 36: nit: Buffer.BlockCopy

src/Sentry/Protocol/UserFeedback.cs
```diff
@@ -3,23 +3,19 @@
 using System.Threading;
 using System.Threading.Tasks;
 using Sentry.Internal;
-using Sentry.Protocol;
-using ISerializable = Sentry.Protocol.ISerializable;
 
-namespace Sentry
+namespace Sentry.Protocol
```
bruno-garcia, line 7: For discoverability, for most common API I'd like to avoid requiring 2 imports

```suggestion
namespace Sentry
```

bruno-garcia (approved):

PR #562, opened 2020-10-27
ref/fix reference url

lucas-zimerman:
This will PR will update all references that were pointing to the old documentation for the new one.


bruno-garcia (approved):
Thanks @lucas-zimerman !

PR #563, opened 2020-10-29
feat: Add support for ASP.NET Core gRPC

Mitch528:


samples/Sentry.Samples.AspNetCore.Grpc/GameService.cs
```diff
@@ -0,0 +1,42 @@
...
+
+        public GameService(ILogger<GameService> gameLogger) => _gameLogger = gameLogger;
+
+        public override async Task<NextPhaseResponse> FetchNextPhaseData(Empty request, ServerCallContext context)
+        {
+            _gameLogger.LogInformation("Fetching dungeons and mana level in parallel.");
+
+            var getDungeonsTask =
+                Task.Run(new Func<int>(() => throw new HttpRequestException("Failed to fetch available Dungeons")));
+            var getUserMana = Task.Run(new Func<int>(() => throw new DataException("Invalid Mana level: -10")));
```
Tyrrrz, line 24: Is specifying `Func<int>` necessary here?

samples/Sentry.Samples.AspNetCore.Grpc/Program.cs
```diff
@@ -0,0 +1,51 @@
+using System;
+using System.Net;
+using Microsoft.AspNetCore;
+using Microsoft.AspNetCore.Hosting;
+
+namespace Sentry.Samples.AspNetCore.Grpc
+{
+#pragma warning disable RCS1102
+    public class Program
```
Tyrrrz, line 9: Any reason why it's not `static`?

samples/Sentry.Samples.AspNetCore.Grpc/GameService.cs
Mitch528: @Tyrrrz I was following how it was in the MVC sample, but yes. It needs to return a `Task<int>`. Attempting to do `Task.Run<int>(() => throw ..)` would result in an ambiguous invocation, so I can't do that either.

samples/Sentry.Samples.AspNetCore.Grpc/Program.cs
Mitch528: Not really, no. I'm not sure why I did that. I'll make it static.

samples/Sentry.Samples.AspNetCore.Grpc/Program.cs
```diff
@@ -0,0 +1,49 @@
...
+            BuildWebHost(args).Run();
+        }
+
+        public static IWebHost BuildWebHost(string[] args) =>
+            WebHost.CreateDefaultBuilder(args)
+                .UseShutdownTimeout(TimeSpan.FromSeconds(10))
+                .UseStartup<Startup>()
+
+                // Example integration with advanced configuration scenarios:
+                .UseSentryGrpc(options =>
```
bruno-garcia, line 21: I'm not sure I follow the approach here. If I had a normal ASP.NET Core app, now I'd need to change my UseSentry to be UseSentryGrpc? 

Couldn't se hook through the container? Basically registering a `EventProcessor` so that the plain ASP.NET Core integration can pick it up?

src/Sentry.AspNetCore.Grpc/Extensions/DependencyInjection/ServiceCollectionExtensions.cs
```diff
@@ -0,0 +1,40 @@
...
+        /// Adds Sentry's services to the <see cref="IServiceCollection"/>
+        /// </summary>
+        /// <param name="services">The services.</param>
+        /// <returns></returns>
+        public static IServiceCollection AddSentryGrpc(this IServiceCollection services)
+        {
+            _ = services.AddSingleton<ISentryEventProcessor, AspNetCoreEventProcessor>();
+            services.TryAddSingleton<IUserFactory, DefaultUserFactory>();
+
+            _ = services.AddSingleton<IProtobufRequestPayloadExtractor, DefaultProtobufRequestPayloadExtractor>();
```
bruno-garcia, line 28:
```suggestion
            _ = services.AddSingleton<IProtobufRequestPayloadExtractor, DefaultProtobufRequestPayloadExtractor>();
```

Looks like instead of copying the whole ASP.NET Core integration, all we need is this line ^, and the one below:

```
    services.AddGrpc(options =>
            {
                options.Interceptors.Add<SentryGrpcInterceptor>();
            });
```

Besides these, the code that goes in the middleware etc...

This integration could be be done differently then, we could offer a single entrypoint, and add it like:

`UseSentry(o => o.AddGrpc()`

And through that extension method we find the right hooks, like adding stuff into DI and the middleware.

src/Sentry.Extensions.Protobuf/Sentry.Extensions.Protobuf.csproj
```diff
@@ -0,0 +1,15 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFrameworks>netcoreapp3.0;net5.0</TargetFrameworks>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Google.Protobuf" Version="3.13.0" />
```
bruno-garcia, line 8: Would it make sense to bundle the GPRC and Protobuf stuff into a single package? Sentry.Grpc?

I don't think we'll need (anytime soon) a Protobuf package stand alone and it would simplify things on. our end.

bruno-garcia (changes requested):
There's a lot of good stuff here @Mitch528 thank you.
The general architecture though I believe can't get merged in its current form. It's too much duplication from ASP.NET Core and I think we can redo this in a way that it hooks into the current ASP.NET Core integration. Without having to take over the hook `UseSentry`. The docs would then refer to: Add `Sentry.AspNetCore.Grpc` and add `AddGrpc` to the callback, or add something DI or whatever.

Would you be willing to make the changes necessary?

samples/Sentry.Samples.AspNetCore.Grpc/Program.cs
Mitch528: I suppose my initial approach was more under the assumption that it would be an either-or sort of thing.

The issue with just making it an `EventProcessor` would be that the plain ASP.NET Core integration does not catch the exceptions thrown by gRPC services. I'm not really sure why that is the case, but it seems to be by design (refer to https://docs.microsoft.com/en-us/aspnet/core/grpc/migration?view=aspnetcore-3.1#grpc-interceptors-vs-middleware).

src/Sentry.Extensions.Protobuf/Sentry.Extensions.Protobuf.csproj
Mitch528: Yep.

Mitch528:
@bruno-garcia I think I resolved the issues you had with the code, but let me know if you need me to make any more changes.

src/Sentry.AspNetCore.Grpc/Sentry.AspNetCore.Grpc.csproj
```diff
@@ -0,0 +1,21 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFrameworks>netcoreapp3.0;net5.0</TargetFrameworks>
```
bruno-garcia, line 4: Do we really need net5.0 here? Wouldn't netcoreapp3.0 be enough?

src/Sentry.AspNetCore.Grpc/SentryGrpcInterceptor.cs
```diff
@@ -0,0 +1,236 @@
...
+
+        private void CaptureException(IHub hub, Exception e)
+        {
+            var evt = new SentryEvent(e);
+
+            _logger.LogTrace("Sending event '{SentryEvent}' to Sentry.", evt);
+
+            var id = hub.CaptureEvent(evt);
+
+            _logger.LogInformation("Event '{id}' queued.", id);
```
bruno-garcia, line 233: Correct me if I'm wrong but this would be the first time we use `ILogger` in the SDK directly, right? So far we've stuck to `sentryOptions.Logger?.Log...`.
In ASP.NET Core (and MEL) we hook into the frameworks logging so you can actually see these logs everywhere you pipe them through, but only if you've opt-ed in to Sentry's logging in the first place (sentryOptions.Debug=true). With this approach you took Sentry's logs will always be send to the applications logs, even if Debug=false.

I propose we stick to `options.DiagnosticLogger?.LogXyz`

bruno-garcia (commented):
Yeah this looks great now! Thanks!
I left two nits out but we're almost there.

src/Sentry.AspNetCore.Grpc/Sentry.AspNetCore.Grpc.csproj
Mitch528: Probably should be good with just netcoreapp3.0. I'll change it.

src/Sentry.AspNetCore.Grpc/SentryGrpcInterceptor.cs
Mitch528: @bruno-garcia I actually took that part directly from the ASP.NET Core middleware: https://github.com/getsentry/sentry-dotnet/blob/main/src/Sentry.AspNetCore/SentryMiddleware.cs#L134

samples/Sentry.Samples.AspNetCore.Grpc/Program.cs
bruno-garcia: I see. Good to know, thanks for clarifying!

src/Sentry.AspNetCore.Grpc/SentryGrpcInterceptor.cs
bruno-garcia: I guess I was the first offender. Since we have presendece in the aspnet integration, looks good to keep it

bruno-garcia (approved):
This is a great addition! Thanks a lot @Mitch528

bruno-garcia:
https://www.nuget.org/packages/Sentry.AspNetCore.Grpc/3.0.0-alpha.4

Mitch528:
@bruno-garcia Awesome! 🎉

PR #571, opened 2020-11-02
Refactor envelopes

Tyrrrz:
@bruno-garcia 
This is what I was attempting to do, but I've realized making `Envelope` internal is actually impossible because `IBackgroundWorker`  and `ITransport` are public and reference `Envelope`.


Tyrrrz:
@bruno-garcia tag

src/Sentry/Protocol/Envelopes/EnvelopeItem.cs
```diff
@@ -7,46 +7,28 @@
...
 using Sentry.Internal.Extensions;
 
-namespace Sentry.Protocol
+namespace Sentry.Protocol.Envelopes
 {
-    /// <summary>
-    /// Envelope item.
-    /// </summary>
-    public class EnvelopeItem : IDisposable, ISerializable
+    internal class EnvelopeItem : ISerializable, IDisposable
```
bruno-garcia, line 12: If (when?) we end up making envelopes public again we'll need to add all these xml docs back.

Why remove it?

bruno-garcia (approved):
Generally much better, not having the protocol types implement serializable is really the way to go.

I'd keep the `HTTP`-related types outside of the `Sentry.Protocol` namespace though, have intenral transport related types inside `Sentry.Internal`, importing types from `Sentry.Protocol` makes more sense to me.

PR #572, opened 2020-11-04
feat: Enumerate all .NET Framework installations

### LINKED ISSUE #531: Enumerate all .NET Framework installations
PlatformAbstractions API is able to enumerate all .NET Framework installations.
An integration could be created to do so on .NET Framework.
The data would fit on a `Context` called `.NET Framework`

Basically: NetFrameworkInstallIntegration class that implement ISdkIntegration
and it'll call hub.AddEventProcessor(s => s.Contexts[".NET Framework"] = GetInstallations()
this method GetInstallations will returns an object that lists all installations.

The integration could be added by default on #if NET461 but there should be a way to opt-out (like `options.RemoveIntegration<NetFrameworkInstallIntegration>();`

lucas-zimerman:
Closes #531 

When running a .NET Application, Sentry will add a list of installed .NET Frameworks with the events.
The operation can be opted-out.


CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+## unreleased
+
+* Add a list of .Net Frameworks installed when avaliable. (#531) @lucas-zimerman
```
```suggestion
* Add a list of .NET Frameworks installed when available. (#531) @lucas-zimerman
```

src/Sentry/Internal/NetFxInstallationsEventProcessor.cs
```diff
@@ -0,0 +1,26 @@
...
+using Sentry.PlatformAbstractions;
+
+namespace Sentry.Internal
+{
+    internal class NetFxInstallationsEventProcessor : ISentryEventProcessor
+    {
+        internal static readonly string NetFxInstallationsKey = ".NET Framework";
+
+        private readonly Lazy<IEnumerable<FrameworkInstallation>> _netFxInstallations =
+            new Lazy<IEnumerable<FrameworkInstallation>>(() => FrameworkInfo.GetInstallations());
```
bruno-garcia, line 14: You need to consider multiple threads will capture events concurrently.

Use [ExecutionAndPublication](https://docs.microsoft.com/en-us/dotnet/api/system.threading.lazythreadsafetymode?view=netcore-3.1#System_Threading_LazyThreadSafetyMode_ExecutionAndPublication) here.

Another point is error handling. You need to make sure an exception thrown in `Lazy` ctor doesn't crash the whole process (or in this case since we'd handle it in the SDK, drop the whole event).

```
As stated earlier, a Lazy<T> object always returns the same object or value that it was initialized with, and therefore the Value property is read-only. If you enable exception caching, this immutability also extends to exception behavior. If a lazy-initialized object has exception caching enabled and throws an exception from its initialization method when the Value property is first accessed, that same exception is thrown on every subsequent attempt to access the Value property
```

We need to enable such caching, and when trying to access `Value` do that under a try catch.
If an exception is thrown, log that with `options.DiagnosticLogger?.LogError` and flip a flag in the object (a field) so that you don't try to access `Value` again. From that point a `LogDebug` is enough just to note that we skipped the integration.

test/Sentry.Tests/Internals/NetFxInstallationsEventProcessorTests.cs
```diff
@@ -0,0 +1,49 @@
...
+
+            //Act
+            _ = sut.Process(@event);
+
+            //Assert
+            _ = Assert.IsAssignableFrom<IEnumerable<FrameworkInstallation>>(@event.Contexts[NetFxInstallationsEventProcessor.NetFxInstallationsKey]);
+        }
+
+        [Fact]
+        public void Process_NetFxInstallationsKeyExist_UnchangedSentryEvent()
```
bruno-garcia, line 33: We need a test that verifies we can actually add something to the context. Maybe call `FrameworkInfo.GetInstallations()` and assert the items are there.

Ideally another that we can serialize the `contexts` once you add the items to that list.

bruno-garcia (commented):
Looking nice! Got a screenshot of what it looks like in Sentry?

bruno-garcia:
Also check if it returns anything because this code will run on Mono on Linux for example and there's no .NET Framework there

lucas-zimerman:
> Looking nice! Got a screenshot of what it looks like in Sentry?

Sure, for the next Commit :D 

image

bruno-garcia:
> EDIT: I don't think there'll be any problems with mono due to #if NETFX (and GetInstallations is using Win32 code)


`#if NETFX` is true on Mono

lucas-zimerman:
> > EDIT: I don't think there'll be any problems with mono due to #if NETFX (and GetInstallations is using Win32 code)
> 
> `#if NETFX` is true on Mono

Yeah, the test is failing because of that :c

src/Sentry/PlatformAbstractions/NetFxInstallationsEventProcessor.cs
```diff
@@ -0,0 +1,68 @@
...
+
+        private bool _netFxInstallationEnabled = true;
+
+        private readonly SentryOptions _options;
+
+        internal NetFxInstallationsEventProcessor(SentryOptions options)
+        {
+            Debug.Assert(options != null);
+            _options = options;
+        }
```
bruno-garcia, line 27: nullable reference types now are in

```suggestion
        internal NetFxInstallationsEventProcessor(SentryOptions options) => _options = options;
```

src/Sentry/PlatformAbstractions/NetFxInstallationsEventProcessor.cs
```diff
@@ -0,0 +1,68 @@
...
+
+        internal static Dictionary<string, string> GetInstallationsDictionary()
+        {
+            var versionsDictionary = new Dictionary<string, string>();
+            var installations = FrameworkInfo.GetInstallations();
+            foreach (var profile in installations.Select(p => p.Profile).Distinct())
```
bruno-garcia, line 33: Why `Distinct` here? Isn't this going to work on the object references?

src/Sentry/PlatformAbstractions/NetFxInstallationsEventProcessor.cs
```diff
@@ -0,0 +1,68 @@
...
+                    {
+                        _options.DiagnosticLogger?.LogError("Failed to add NetFxInstallations into event.", ex);
+                        //In case of any failure, this process function will be disabled to avoid throwing exceptions for future events.
+                        _netFxInstallationEnabled = false;
+                    }
+                }
+            }
+            else
+            {
+                _options.DiagnosticLogger.LogDebug("Process skipped due to an exception with this with the processor.");
```
bruno-garcia, line 62:
```suggestion
                _options.DiagnosticLogger.LogDebug("NetFxInstallation disabled due to previous error.");
```

test/Sentry.Tests/Integrations/NetFxInstallationsIntegrationTests.cs
```diff
@@ -0,0 +1,47 @@
...
+using Xunit;
+
+namespace Sentry.Tests.Integrations
+{
+    public class NetFxInstallationsIntegrationTests
+    {
+        [SkippableFact]
+        public void Register_IgnoredIfPlattformIsMono()
+        {
+            Skip.If(!Runtime.Current.IsMono());
```
bruno-garcia, line 17: Don't u need this off in order to test this?
```suggestion
```

test/Sentry.Tests/PlatformAbstractions/FrameworkInstallationExtensionsTests.cs
```diff
@@ -0,0 +1,79 @@
...
+using Xunit;
+
+namespace Sentry.Tests.PlatformAbstractions
+{
+    public class FrameworkInstallationExtensionsTests
+    {
+        [Fact]
+        public void GetVersionNumber_NullFrameworkInstallation_NullVersion()
+        {
+            FrameworkInstallation frameworkInstallation = null;
```
bruno-garcia, line 12: This compiles? What happened to nullable reference types?

test/Sentry.Tests/PlatformAbstractions/FrameworkInstallationExtensionsTests.cs
```diff
@@ -0,0 +1,79 @@
...
+            Assert.Null(version);
+        }
+
+        [Fact]
+        public void GetVersionNumber_ValidShortVersion_ShortVersion()
+        {
+            var frameworkInstallation = new FrameworkInstallation();
+            var expectedShortVersion = "v1.2.3";
+            frameworkInstallation.ShortName = expectedShortVersion;
+            //Arrange
```
bruno-garcia, line 41: Why is Arrange below the arrange?

Consider leaving these comments out and using the empty lines to split them.
Otherwise please move the comment above

lucas-zimerman: not sure why I did that, in any case, fixed.

test/Sentry.Tests/Integrations/NetFxInstallationsIntegrationTests.cs
lucas-zimerman: The test just checks when you are running mono that the event processor isn't added.

`public void Register_PlattformIsMono_NetFxInstallationsEventProcessorNotAdded()`
I suppose the function name caused some confusion.

lucas-zimerman: or Register_CurrentRuntimeIsMono_NetFxInstallationsEventProcessorNotAdded

test/Sentry.Tests/PlatformAbstractions/FrameworkInstallationExtensionsTests.cs
lucas-zimerman: It does, and I added the nullable type in the next commit.

src/Sentry/PlatformAbstractions/NetFxInstallationsEventProcessor.cs
lucas-zimerman: The goal of distinct is to avoid repeating keys to be inserted into the Dictionary and the foreach loop will then group the releases by profiles.

src/Sentry/PlatformAbstractions/FrameworkInstallationExtensions.cs
```diff
@@ -0,0 +1,10 @@
+namespace Sentry.PlatformAbstractions
+{
+    internal static class FrameworkInstallationExtensions
+    {
+        internal static string? GetVersionNumber(this FrameworkInstallation frameworkInstall)
+            => frameworkInstall?.ShortName
+                ?? (frameworkInstall?.Version != null ? $"v{frameworkInstall.Version}" : null);
+
```
bruno-garcia, line 8:
```suggestion
```

src/Sentry/PlatformAbstractions/NetFxInstallationsEventProcessor.cs
bruno-garcia: But you're doing `Distinct` on a reference type so unless `GetHashCode` and `Equals` was overriden, it will dedupe on the reference.

bruno-garcia (approved):
LGTM. Just clarifying if that `Distinct` does something or not

src/Sentry/PlatformAbstractions/FrameworkInfo.NetFx.cs
```diff
@@ -103,7 +103,7 @@ public static IEnumerable<FrameworkInstallation> GetInstallations()
                 if (version != null && versionKey.GetInt("Install") == 1)
                 {
                     // 1.0 to 3.5
-                    Version.TryParse(version, out var parsed);
+                    _ = Version.TryParse(version, out var parsed);
```
Tyrrrz, line 106: Is this some sort of analyzer we have that warns about this? I don't see it on my end and I'd say adding those `_` makes things more noisy.

bruno-garcia: I get the warnings too. Visual Studio

src/Sentry/PlatformAbstractions/NetFxInstallationsEventProcessor.cs
bruno-garcia: Confirmed offline that it fails with dupe key in Map otherwise

src/Sentry/PlatformAbstractions/FrameworkInfo.NetFx.cs
lucas-zimerman: image
The Sentry project is a bit noisy about unused return parameters, it compiles, but leaves a lot of error messages here. (I am running Visual Studio Community without any addon)

PR #573, opened 2020-11-05
UWP support

bruno-garcia:


lucas-zimerman:
I suggest that there'll be a PlatformIntegration.cs inside Integrations folder where it calls new Platform Integrations if they exist.
Next, I would add specific folders for each platform inside of PlatformAbstractions (or Integrations).

What do you think @bruno-garcia?

For example:

file: Integrations/PlatformAbstractions
```
    internal class PlatformIntegration : IInternalSdkIntegration
    {
        {
            #if WINDOWS_UWP
            options.AddIntegration(new UwpPlatformIntegration());
            #endif
            //Remove itself after registered?
        }

        public void Unregister(IHub hub)
        {
            ...
        }
    }
```

file: PlatformAbstractions/Uwp/UwpPlatformIntegration.cs
```
#if WINDOWS_UWP
    internal class UwpPlatformIntegration : IInternalSdkIntegration
    {
        {
            ...
        }

        public void Unregister(IHub hub)
        {
            ...
        }
    }
#endif
```

bruno-garcia:
I personally prefer the way I put it up so far, so if you add a new `TFM` the build will break for the missing `PlatformIntegration` and it's clear.

It also avoid `#ifdef` all over, they'd be clearly defined with the extension.

We can also change it in the future if that's the case

bruno-garcia:
@lucas-zimerman I'll close this since your PR works on top of this

PR #576, opened 2020-11-05
Caching

### LINKED ISSUE #557: Cache Envelopes
Multiple users have requested a way to store events to disk. Either because the device is always offline or because it can be offline for long periods of time (laptops in the field, oil rigs etc).

Now that we have Envelopes #544 we can more easily implement caching by **optionally** storing envelopes to disk. This feature will help desktop apps and also Xamarin/.NET Mobile, given that mobile often needs to persist to disk before attempting to send out an event. This is supported by Android and iOS too.

* Offline caching is optional
  * A path to serve as root path for storing things must be provided via `SentryOptions`.
* Max files must be capped, default could be 30 to match other SDKs.
* If the cache directory is full, delete the oldest item first.
  * Ideally using the file timestamp date
  * Consider that envelopes might be edited later (changes last modified date), when sessions are added, to track `init=true`.
  * Consider that this can run on Android, Windows, Linux, macOS, the file timestamp API expected to work on these platforms.
* The order of files cached should be respected.
  * If an envelope is captured and files are queued for submission on the disk, send the files from disk first.
* If a file is cached and subsequently sent to Sentry, the cached file then is deleted.
* Consideration for retry:
  * If sending to Sentry fails due to timeout or no connection at all, the file continues to be cached
  * If Sentry returns **any** status code, the file should be deleted: Don't retry an envelope that hit Sentry and caused a 500, 200, 400 or any status.

This will resolve: https://github.com/getsentry/sentry-dotnet/issues/80

Tyrrrz:
Closes #557


src/Sentry/SentryOptionsExtensions.cs
```diff
@@ -182,6 +183,15 @@ public static SentryOptions UseStackTraceFactory(this SentryOptions options, ISe
             return options;
         }
 
+        /// <summary>
+        /// Sets the cache directory.
+        /// </summary>
+        public static SentryOptions SetCacheDirectory(this SentryOptions options, string directoryPath)
```
bruno-garcia, line 189: What about we hold only a `string` in the options (to make it easier for folks to add the path, remove this method) and we create the DirectoryInfo internally?

src/Sentry/Internal/CachingBackgroundWorker.cs
```diff
@@ -0,0 +1,89 @@
...
+            while (TryGetNextEnvelopeFile() is { } envelopeFile)
+            {
+                Options.DiagnosticLogger?.LogDebug("Sending cached envelope: {0}", envelopeFile.FullName);
+
+                using var envelope = await Envelope.DeserializeAsync(
+                    envelopeFile.OpenRead(),
+                    cancellationToken
+                ).ConfigureAwait(false);
+
+                await Transport.SendEnvelopeAsync(envelope, cancellationToken).ConfigureAwait(false);
```
bruno-garcia, line 49: We default to deleting the file. If there was an exception while trying to send it, for example. Or if Sentry returned 400, 500.

The exception is when we can clearly identify we couldn't connect to Sentry (timeout, or offline).
In this case we keep the file around to try again later.

That said I believe we're missing a mechanism to communicate that here. And some try/finally(delete)

src/Sentry/Internal/CachingBackgroundWorker.cs
```diff
@@ -0,0 +1,89 @@
...
+            _cacheDirectory.Create();
+            using var stream = envelopeFile.Create();
+            await envelope.SerializeAsync(stream, cancellationToken).ConfigureAwait(false);
+        }
+
+        protected override async ValueTask ProcessEnvelopeAsync(
+            Envelope envelope,
+            CancellationToken cancellationToken = default)
+        {
+            await FlushCacheAsync(cancellationToken).ConfigureAwait(false);
```
bruno-garcia, line 77: Does it mean we're not storing the `envelope` received here until we deal with cached files?
I believe we should cache the first first of all, to guarantee in the event of a failure (device our of power, process killed, crashed) we didn't lose the event.

The behavior for mobile at least, is to alwauys write to disk first, then try to send. and delete if it succeeded. Also delete it if we got a failure that could happen again once we retry.

src/Sentry/Internal/CachingBackgroundWorker.cs
```diff
@@ -0,0 +1,89 @@
...
+
+                await Transport.SendEnvelopeAsync(envelope, cancellationToken).ConfigureAwait(false);
+
+                envelopeFile.Delete();
+            }
+        }
+
+        private async ValueTask CacheEnvelopeAsync(Envelope envelope, CancellationToken cancellationToken = default)
+        {
+            if (GetCachedEnvelopeFiles().Count() >= 50)
```
bruno-garcia, line 57: 30 I believe is the general default, this will need to propagate from Options.

```suggestion
            if (GetCachedEnvelopeFiles().Count() >= 30)
```

src/Sentry/Internal/CachingBackgroundWorker.cs
```diff
@@ -0,0 +1,89 @@
...
+            {
+                return;
+            }
+
+            var envelopeFile = new FileInfo(
+                Path.Combine(
+                    _cacheDirectory.FullName,
+                    $"{DateTimeOffset.Now}_{envelope.TryGetEventId()}_{envelope.GetHashCode()}")
```
bruno-garcia, line 65: You used a `.envelope` extension above, we need it here?

What about UtcNow so the file names are not subjective to daylight savings/timezone changes?

What if not event id in the envelope, maybe `?? ...`

src/Sentry/SentryOptions.cs
```diff
@@ -360,6 +361,12 @@ public bool Debug
         /// </remarks>
         public DeduplicateMode DeduplicateMode { get; set; } = DeduplicateMode.All ^ DeduplicateMode.InnerException;
 
+        /// <summary>
+        /// Path to the directory where events are stored locally for resilience purposes.
+        /// If set to <code>null</code>, caching will not be used.
+        /// </summary>
+        public DirectoryInfo? CacheDirectory { get; set; }
```
bruno-garcia, line 368: I think `string` is probably a better type, to for example bind from configuration files

bruno-garcia (commented):
I see it's a draft, but a first pass hopefully some points will be useful

src/Sentry/SentryOptionsExtensions.cs
Tyrrrz: I kinda like the explicitness of `DirectoryInfo` specifically for user-facing interface. I wouldn't use it internally.

src/Sentry/Internal/CachingBackgroundWorker.cs
Tyrrrz: That's true, I will update it

src/Sentry/Internal/CachingBackgroundWorker.cs
Tyrrrz: Good point, will update.

src/Sentry/Internal/CachingBackgroundWorker.cs
Tyrrrz: Yeah, I just put a random number for now

src/Sentry/Internal/CachingBackgroundWorker.cs
Tyrrrz: Good points. I didn't think this through very much yet.

src/Sentry/SentryOptions.cs
Tyrrrz: Hm, I didn't think about configuration files. That may indeed be a good reason to change it to `string`. Ah, unfortunate that C# discourages proper static typing so much.

src/Sentry/Internal/CachingBackgroundWorker.cs
Tyrrrz: Fixed.

If event ID is not set then it will be empty, so it will look like this:

```
123456789__91231293.envelope
```

Note the double underscore. Since the file names are not really meant to be read by users, I think it's okay. It will still be sorted by date and combination of unix time + hash code will ensure there are no collisions. Event ID is there for debugging purposes, mostly.

Tyrrrz:
@bruno-garcia addressed the issues you pointed out.  Going to start working on tests soon, so if there are any immediate design concerns, we should address them now :)

src/Sentry/Internal/BackgroundWorkerBase.cs
```diff
@@ -144,7 +125,7 @@ private async ValueTask WorkerAsync()
                         }
                         catch (Exception exception)
                         {
-                            _options.DiagnosticLogger?.LogError(
+                            Options.DiagnosticLogger?.LogError(
```
bruno-garcia, line 128: Seems the format on this one expects 3 arguments but if the first is Exception (not part of `params`) are we missing one?

src/Sentry/Internal/CachingBackgroundWorker.cs
```diff
@@ -0,0 +1,146 @@
...
+        public CachingBackgroundWorker(ITransport transport, SentryOptions options)
+            : base(transport, options)
+        {
+            _cacheDirectory = !string.IsNullOrWhiteSpace(options.CacheDirectoryPath)
+                ? _cacheDirectory = new DirectoryInfo(options.CacheDirectoryPath)
+                : throw new InvalidOperationException("Cache directory is not set.");
+        }
+
+        private IEnumerable<FileInfo> GetEnvelopeFiles() =>
+            _cacheDirectory.Exists
```
bruno-garcia, line 26: Should we check if it exists in the `ctor` and do some LogWarn or LogError?
I'm afraid this will be hard to debug

src/Sentry/Internal/CachingBackgroundWorker.cs
```diff
@@ -0,0 +1,146 @@
...
+
+                    // Delete the cache file in case of success
+                    envelopeFile.Delete();
+
+                    Options.DiagnosticLogger?.LogDebug(
+                        "Successfully sent cached envelope: {0}",
+                        envelope.TryGetEventId()
+                    );
+                }
+                catch (IOException ex)
```
bruno-garcia, line 69: I'm afraid we could have some I/O error that's not really recoverable and we'll be retrying that forever. Maybe we need to be more strict on what we consider OK to retry

src/Sentry/Internal/CachingBackgroundWorker.cs
```diff
@@ -0,0 +1,146 @@
...
+                {
+                    // Don't delete the cache file in case of transient exceptions,
+                    // i.e. loss of connection, failure to connect, etc.
+                    // Instead break to not violate the order and wait until
+                    // the next batch to attempt sending again.
+                    Options.DiagnosticLogger?.Log(
+                        SentryLevel.Error,
```
bruno-garcia, line 76:
```suggestion
                    Options.DiagnosticLogger?.LogError(
```

src/Sentry/Internal/CachingBackgroundWorker.cs
```diff
@@ -0,0 +1,146 @@
...
+                    break;
+                }
+                catch (Exception ex)
+                {
+                    // Delete the cache file for all other exceptions that could
+                    // indicate a successfully completed, but error response.
+                    envelopeFile.Delete();
+
+                    Options.DiagnosticLogger?.Log(
+                        SentryLevel.Error,
```
bruno-garcia, line 91:
```suggestion
                    Options.DiagnosticLogger?.LogError(
```

src/Sentry/Internal/CachingBackgroundWorker.cs
```diff
@@ -0,0 +1,146 @@
...
+
+            return envelopeFile;
+        }
+
+        protected override async ValueTask ProcessEnvelopeAsync(
+            Envelope envelope,
+            CancellationToken cancellationToken = default)
+        {
+            // Add this envelope to cache
+            await StoreEnvelopeAsync(envelope, cancellationToken).ConfigureAwait(false);
```
bruno-garcia, line 139: If we throw here, the worker thread above is supposed to log and continue, right?
I wonder if we'll drop the item and deal with it "gracefully".
We need to make sure we don't end up retrying something broken in a tight loop here on a worker thread or killing that thread and stopping all event submission

src/Sentry/Internal/CachingBackgroundWorker.cs
```diff
@@ -0,0 +1,146 @@
...
+using System.IO;
+using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
+using Sentry.Extensibility;
+using Sentry.Protocol.Envelopes;
+
+namespace Sentry.Internal
+{
+    internal class CachingBackgroundWorker : BackgroundWorkerBase
```
bruno-garcia, line 12: We're not going to be able to use this class later outside of the background thread..
What if we used composition here somehow?

The reason I ask is the Xamarin crash handler use case, where we need to write to disk but not sure we can do on a threadpool thread.

We can try that first before changing anything, and we can change it later too, just thinking ahead a bit

bruno-garcia (commented):
Looking awesome!

src/Sentry/Internal/BackgroundWorkerBase.cs
```diff
@@ -8,65 +8,42 @@
 
 namespace Sentry.Internal
 {
-    internal class BackgroundWorker : IBackgroundWorker, IDisposable
+    internal abstract class BackgroundWorkerBase : IBackgroundWorker, IDisposable
```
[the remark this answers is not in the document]
Tyrrrz: I find that the `Base` suffix is much more common for abstract classes that have base functionality, while abstract classes without functionality (basically interfaces) don't have that suffix. For example, `ControllerBase` from aspnetcore mvc.

src/Sentry/Internal/CachingBackgroundWorker.cs
Tyrrrz: That would probably be a bad idea. File system is very volatile, so even if it existed during construction, it may get deleted somehow after that, but before writing files. We should always treat working with FS as if it's in a constantly unpredictable state.

src/Sentry/Internal/CachingBackgroundWorker.cs
Tyrrrz: Unfortunately most transient errors we would want to retry on just throw base `IOException`. Some have inner exceptions (like `SocketException`) but not all and it's probably implementation-dependent. I guess we'd have to parse exception messages if we really wanted to be selective.

src/Sentry/Internal/CachingBackgroundWorker.cs
Tyrrrz: Yeah, this method is called from the worker which has like 5 layers of `try`/`catch` :)
How do you think the workflow should look like?

src/Sentry/Internal/CachingBackgroundWorker.cs
Tyrrrz: We may want to consolidate into one approach somehow, because otherwise we'll have to have two caching background workers, one for Xamarin and one for everything else.

src/Sentry/Internal/CachingBackgroundWorker.cs
bruno-garcia: I agree with everything you said. But I would expect to see some logs if the directory isn't there when the user explicitly set a path and expects offline caching to be working,

src/Sentry/Internal/CachingBackgroundWorker.cs
bruno-garcia: That's an option. For now we're good with IOException, the client returning 500 wouldn't throw so at least we know that if it hits Sentry it'll be deleted

src/Sentry/Internal/CachingBackgroundWorker.cs
bruno-garcia: > Yeah, this method is called from the worker which has like 5 layers of try/catch :)


Does this mean the worker won't crash? I believe that's the case.

src/Sentry/Internal/CachingBackgroundWorker.cs
bruno-garcia: I'll try your code now on Xamarin crash handler

src/Sentry/Internal/CachingBackgroundWorker.cs
Tyrrrz: Do you think it's unusual for someone to set a caching directory to something that doesn't exist? I personally would expect it to be normal.

I can add a log when we save a file, but the directory is not there.

src/Sentry/Internal/CachingBackgroundWorker.cs
Tyrrrz: Yeah it won't crash.

src/Sentry/Internal/BackgroundWorkerBase.cs
Tyrrrz: Can't figure out what it's missing :/
Probably some info got lost in transition from events to envelopes. I'll remove the extra placeholder for now.

Tyrrrz:
Updated @bruno-garcia.

Dispose on `BackgroundWorker` is so messy, I didn't bother trying to dispose the `ITransport` in it for the fear of introducing bugs. I had some major refactor done for `BackgroundWorker` in one of my previous attempts, but had to discard it because I switched to a different approach now.

Tyrrrz:
Work of art now.

Tyrrrz:
Hm, the tests are failed. This must be the work of an enemy stand. I'll see what's wrong.

lucas-zimerman:
@Tyrrrz I think the windows issue is related to the one that's happening with the UWP PR

Tyrrrz:
Added blocking on SentrySdk.Init

test/Sentry.Tests/Internals/Http/CachingTransportTests.cs
```diff
@@ -0,0 +1,227 @@
...
+            {
+                using var envelope = Envelope.FromEvent(new SentryEvent());
+                await transport.SendEnvelopeAsync(envelope);
+            }
+
+            await transport.FlushAsync();
+
+            // (transport stops failing)
+            innerTransport.ClearReceivedCalls();
+            isFailing = false;
```
bruno-garcia, line 170: Can u make sure it didn't do the job yet?

```suggestion
            _ = innerTransport.Received(0).SendEnvelopeAsync(Arg.Any<Envelope>(), Arg.Any<CancellationToken>());
            isFailing = false;
```

test/Sentry.Tests/Internals/Http/CachingTransportTests.cs
```diff
@@ -0,0 +1,227 @@
...
+            _ = innerTransport.Received(0).SendEnvelopeAsync(Arg.Any<Envelope>(), Arg.Any<CancellationToken>());
+        }
+
+        public void Dispose()
+        {
+            if (Directory.Exists(CacheDirectoryPath))
+            {
+                Directory.Delete(CacheDirectoryPath, true);
+            }
+        }
```
bruno-garcia, line 225: It just noops if there's no folder to delete

```suggestion
            Directory.Delete(CacheDirectoryPath, true);
        }
```

src/Sentry/Internal/Lock.cs
```diff
@@ -0,0 +1,21 @@
...
+
+namespace Sentry.Internal
+{
+    internal class Lock : IDisposable
+    {
+        private readonly Signal _signal;
+
+        public Lock() => _signal = new Signal(true);
+
+        public async ValueTask<IDisposable> ClaimAsync(CancellationToken cancellationToken = default)
```
bruno-garcia, line 13:
```suggestion
        public async ValueTask<IDisposable> AcquireAsync(CancellationToken cancellationToken = default)
```

src/Sentry/Internal/Http/CachingTransport.cs
```diff
@@ -0,0 +1,266 @@
...
+            _processingDirectoryPath = Path.Combine(_cacheDirectoryPath, "__processing");
+
+            // Processing directory may already contain some files left from previous session
+            // if the worker has been terminated unexpectedly.
+            // Move everything from that directory back to cache directory.
+            if (Directory.Exists(_processingDirectoryPath))
+            {
+                foreach (var filePath in Directory.EnumerateFiles(_processingDirectoryPath))
+                {
+                    Directory.CreateDirectory(_cacheDirectoryPath);
```
bruno-garcia, line 57: This could be created outside of the `foreach`.
It should no-op if it already exists.

src/Sentry/Internal/Http/CachingTransport.cs
```diff
@@ -0,0 +1,266 @@
...
+            _cacheDirectoryPath = !string.IsNullOrWhiteSpace(options.CacheDirectoryPath)
+                ? _cacheDirectoryPath = options.CacheDirectoryPath
+                : throw new InvalidOperationException("Cache directory is not set.");
+
+            _processingDirectoryPath = Path.Combine(_cacheDirectoryPath, "__processing");
+
+            // Processing directory may already contain some files left from previous session
+            // if the worker has been terminated unexpectedly.
+            // Move everything from that directory back to cache directory.
+            if (Directory.Exists(_processingDirectoryPath))
```
bruno-garcia, line 53: `_cacheDirectoryPath` is a subdirectory of `_cacheDirectoryPath` so creating `_cacheDirectoryPath` in the foreach isn't needed.

src/Sentry/Internal/Http/CachingTransport.cs
```diff
@@ -0,0 +1,266 @@
...
+
+            var targetFilePath = Path.Combine(_processingDirectoryPath, Path.GetFileName(filePath));
+
+            Directory.CreateDirectory(_processingDirectoryPath);
+            File.Move(filePath, targetFilePath);
+
+            return targetFilePath;
+        }
+
+        private async ValueTask ProcessCacheAsync(CancellationToken cancellationToken = default)
```
bruno-garcia, line 124: Please make this the first method (or at least `private` method) since this is what we should be reading first.

src/Sentry/Internal/Http/CachingTransport.cs
```diff
@@ -0,0 +1,266 @@
...
+        private async ValueTask StoreToCacheAsync(
+            Envelope envelope,
+            CancellationToken cancellationToken = default)
+        {
+            using var lockClaim = await _cacheDirectoryLock.ClaimAsync(cancellationToken).ConfigureAwait(false);
+
+            // If over capacity - remove oldest envelope file
+            while (
+                GetCacheFilePaths().Count() >= _options.MaxQueueItems &&
+                GetCacheFilePaths().FirstOrDefault() is { } oldestEnvelopeFilePath)
```
bruno-garcia, line 207: What about assigning the result of `GetCacheFilePaths` before the `FirstOrDefault`?
Maybe a local function would help

src/Sentry/Internal/Http/CachingTransport.cs
```diff
@@ -0,0 +1,266 @@
...
+            Directory.CreateDirectory(_cacheDirectoryPath);
+
+            using (var stream = File.Create(envelopeFilePath))
+            {
+                await envelope.SerializeAsync(stream, cancellationToken).ConfigureAwait(false);
+            }
+
+            // Tell the worker that there is work available
+            // (file stream MUST BE DISPOSED prior to this)
+            _workerSignal.Release();
```
bruno-garcia, line 233: If these operations above fail, we don't `Release`?
Are we missing a try finally here?

src/Sentry/Internal/Http/CachingTransport.cs
```diff
@@ -0,0 +1,266 @@
...
+            // Store the envelope in a file without actually sending it anywhere.
+            // The envelope will get picked up by the background thread eventually.
+            await StoreToCacheAsync(envelope, cancellationToken).ConfigureAwait(false);
+        }
+
+        public async ValueTask StopWorkerAsync()
+        {
+            // Stop worker and wait until it finishes
+            _workerCts.Cancel();
+            await _worker.ConfigureAwait(false);
```
bruno-garcia, line 250: Since we call this from the `Dispose` method, awaiting on a Task that fails throws. Should we try/catch here?

src/Sentry/SentryOptions.cs
```diff
@@ -51,6 +52,7 @@ public class SentryOptions : IScopeOptions
         /// </summary>
         internal ISdkIntegration[]? Integrations { get; set; }
 
+
```
lucas-zimerman, line 55: Is this extra line required?

src/Sentry/Internal/Http/CachingTransport.cs
```diff
@@ -0,0 +1,266 @@
...
+                await envelope.SerializeAsync(stream, cancellationToken).ConfigureAwait(false);
+            }
+
+            // Tell the worker that there is work available
+            // (file stream MUST BE DISPOSED prior to this)
+            _workerSignal.Release();
+        }
+
+        // This method asynchronously blocks until the envelope is written to cache, but not until it's sent
+        public async ValueTask SendEnvelopeAsync(
```
bruno-garcia, line 237: This should come right after the constructor

src/Sentry/Internal/Http/CachingTransport.cs
```diff
@@ -0,0 +1,266 @@
...
+                _options.DiagnosticLogger?.LogDebug(
+                    "Reading cached envelope: {0}",
+                    envelopeFilePath
+                );
+
+                var shouldRetry = false;
+
+                using (var envelopeFile = File.OpenRead(envelopeFilePath))
+                using (var envelope =
+                    await Envelope.DeserializeAsync(envelopeFile, cancellationToken).ConfigureAwait(false))
```
lucas-zimerman, line 139: If the file got corrupted, isn't deserialization going to throw an Exception and the remaining Envelopes are not going to be loaded?

src/Sentry/Internal/Http/CachingTransport.cs
```diff
@@ -0,0 +1,271 @@
...
+            _cacheDirectoryPath = !string.IsNullOrWhiteSpace(options.CacheDirectoryPath)
+                ? _cacheDirectoryPath = options.CacheDirectoryPath
+                : throw new InvalidOperationException("Cache directory is not set.");
+
+            _processingDirectoryPath = Path.Combine(_cacheDirectoryPath, "__processing");
+
+            // Processing directory may already contain some files left from previous session
+            // if the worker has been terminated unexpectedly.
+            // Move everything from that directory back to cache directory.
+            if (Directory.Exists(_processingDirectoryPath))
```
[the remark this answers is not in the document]
Tyrrrz: Are you sure it does? I'm pretty sure I remember it throwing in the past.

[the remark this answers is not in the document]
Tyrrrz: image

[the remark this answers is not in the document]
Tyrrrz: https://docs.microsoft.com/en-us/dotnet/api/system.io.directory.enumeratefiles?view=net-5.0

src/Sentry/Internal/Http/CachingTransport.cs
```diff
@@ -0,0 +1,266 @@
...
+                $"{envelope.TryGetEventId()}_" +
+                $"{envelope.GetHashCode()}" +
+                $".{EnvelopeFileExt}"
+            );
+
+            Directory.CreateDirectory(_cacheDirectoryPath);
+
+            using (var stream = File.Create(envelopeFilePath))
+            {
+                await envelope.SerializeAsync(stream, cancellationToken).ConfigureAwait(false);
```
lucas-zimerman, line 228: If an IOException like Disk full happens, it'll not release the workerSignal?

src/Sentry/Internal/Http/CachingTransport.cs
```diff
@@ -0,0 +1,266 @@
...
+        }
+
+        public async ValueTask FlushAsync(CancellationToken cancellationToken = default) =>
+            await ProcessCacheAsync(cancellationToken).ConfigureAwait(false);
+
+        public void Dispose()
+        {
+            StopWorkerAsync().GetAwaiter().GetResult();
+
+            _workerSignal.Dispose();
```
Tyrrrz, line 260: Dispose inner transport

src/Sentry/Internal/Http/CachingTransport.cs
```diff
@@ -0,0 +1,266 @@
...
+
+        private readonly CancellationTokenSource _workerCts = new CancellationTokenSource();
+        private readonly Task _worker;
+
+        public CachingTransport(ITransport innerTransport, SentryOptions options)
+        {
+            _innerTransport = innerTransport;
+            _options = options;
+
+            _cacheDirectoryPath = !string.IsNullOrWhiteSpace(options.CacheDirectoryPath)
```
Tyrrrz, line 44: Use a subdirectory: `{cacheDir}/Sentry/DSN`

Tyrrrz:
Mention: multiple instances of the same apps are not supported

src/Sentry/Internal/Http/CachingTransport.cs
```diff
@@ -0,0 +1,266 @@
...
+                        await _workerSignal.WaitAsync(_workerCts.Token).ConfigureAwait(false);
+                        await ProcessCacheAsync(_workerCts.Token).ConfigureAwait(false);
+                    }
+                    catch (OperationCanceledException)
+                    {
+                        // Ignore
+                    }
+                    catch (Exception ex)
+                    {
+                        _options.DiagnosticLogger?.LogError(
```
bruno-garcia, line 81: Since this is potentially dangerous (we will throw on the next iteration) lets slow things down:

```csharp
await Task.Delay(500);
```

src/Sentry/Internal/Http/CachingTransport.cs
Tyrrrz: No, it won't.

src/Sentry/Internal/Http/CachingTransport.cs
Tyrrrz: Good point.

src/Sentry/Internal/Http/CachingTransport.cs
```diff
@@ -0,0 +1,271 @@
...
+
+            var filePath = GetCacheFilePaths().FirstOrDefault();
+            if (string.IsNullOrWhiteSpace(filePath))
+            {
+                return null;
+            }
+
+            var targetFilePath = Path.Combine(_processingDirectoryPath, Path.GetFileName(filePath));
+
+            Directory.CreateDirectory(_processingDirectoryPath);
```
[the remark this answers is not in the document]
Tyrrrz: Yes, it's not a guarantee, but it's rather cheap if the directory already exists and it's not like we're running in a tight loop anyway. This gives a bit more resiliency because we don't care about previous state.

src/Sentry/Internal/Http/CachingTransport.cs
Tyrrrz: Why?

src/Sentry/Internal/Http/CachingTransport.cs
Tyrrrz: No. Why would we need to release it in that case?

test/Sentry.Tests/Internals/Http/CachingTransportTests.cs
```diff
@@ -0,0 +1,227 @@
...
+using Sentry.Protocol.Envelopes;
+using Sentry.Testing;
+using Xunit;
+
+namespace Sentry.Tests.Internals.Http
+{
+    public class CachingTransportTests : IDisposable
+    {
+        private string CacheDirectoryPath { get; } = Path.Combine(
+            Directory.GetCurrentDirectory(),
```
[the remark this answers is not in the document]
Tyrrrz: This lets us inspect the actual cache directory more easily. Plus I don't like polluting system directories if I have a choice.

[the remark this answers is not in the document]
Tyrrrz: Note: with the current setup, the cache directory ends up inside `bin/debug` directory in tests.

test/Sentry.Tests/Internals/Http/CachingTransportTests.cs
Tyrrrz: No it doesn't 😄

test/Sentry.Tests/Internals/Http/CachingTransportTests.cs
Tyrrrz: well it did call that method when trying (and failing) initially

test/Sentry.Tests/Internals/Http/CachingTransportTests.cs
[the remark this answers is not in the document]
lucas-zimerman: https://github.com/lucas-zimerman/ContribSentryDotNet/blob/master/ContribSentry.Testing/TempFolder.cs

Something like this would be nice for testing the cached folder, it'll delete the files once the test is finished with or without an exception.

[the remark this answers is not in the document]
Tyrrrz: That does not work if the test is canceled by xUnit timeout, but I think it looks nicer so I'll refactor to that

Tyrrrz:
@bruno-garcia should we keep the retry logic, given your experimentation with Xamarin?

src/Sentry/Internal/Signal.cs
```diff
@@ -0,0 +1,39 @@
...
+            // Make sure the semaphore does not go above 1
+            lock (_lock)
+            {
+                if (_semaphore.CurrentCount >= 1)
+                {
+                    return;
+                }
+            }
+
+            _semaphore.Release();
```
bruno-garcia: racey, sync the whole thing to make sure no 2 threads reach Release

bruno-garcia (approved):

PR #580, opened 2020-11-08
ref: TFMs and Dependencies

bruno-garcia:


src/Sentry.Log4Net/Sentry.Log4Net.csproj
```diff
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFrameworks>netstandard2.1;netstandard2.0</TargetFrameworks>
+    <TargetFrameworks>netstandard2.0;net461</TargetFrameworks>
```
bruno-garcia: to simplify dependency resolution, when targeting ns2.0 we should also target net461.

Also, that version of log4net doesn't support ns2.1 so there's no advantage to target ns2.1 here

src/Sentry.NLog/Sentry.NLog.csproj
```diff
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFrameworks>netstandard2.1;netstandard2.0;net461</TargetFrameworks>
+    <TargetFrameworks>netstandard2.0;net461</TargetFrameworks>
```
bruno-garcia: The version of NLog we depend here doesn't target ns2.1 and we have no code that needs it so dropping it.

src/Sentry.Log4Net/Sentry.Log4Net.csproj
```diff
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFrameworks>netstandard2.1;netstandard2.0</TargetFrameworks>
+    <TargetFrameworks>netstandard2.1;netstandard2.0;net461</TargetFrameworks>
```
bruno-garcia, line 4: when ns2.0, we should multi-target and add net461 to avoid dependency resolution issues

src/Sentry.Serilog/Sentry.Serilog.csproj
```diff
@@ -13,8 +13,12 @@
     <ProjectReference Include="../../src/Sentry/Sentry.csproj" />
   </ItemGroup>
 
-  <ItemGroup>
+  <ItemGroup Condition="$(TargetFramework) != 'netstandard2.1'">
     <PackageReference Include="Serilog" Version="2.7.1" />
   </ItemGroup>
 
+  <ItemGroup Condition="$(TargetFramework) == 'netstandard2.1'">
+    <PackageReference Include="Serilog" Version="2.10.0" />
```
bruno-garcia, line 21: Serilog 2.10 came up recently but has a ns2.1 target so lets try this.
A fallback would be to target Sentry.Serilog of ns2.0 if one can't bump Serilog to 2.10 for some reason.

lucas-zimerman:
Looks ok :)

bruno-garcia:
Codecovv doesn't say why:
> There was a problem loading the file:

bruno-garcia:
0.28% though, not worth the noise

riyad-hub:
More like

PR #589, opened 2020-11-16
feat: net5 for sentry.csproj

lucas-zimerman:
Add target framework NET5.0 on Sentry.csproj. 
Change the type of `Extra` where the value parameter becomes nullable.


lucas-zimerman:
@bruno-garcia  not sure why there were no tests on the other platforms (Ubuntu  and mac) but on my Branch PR it passed the Ubuntu and Mac test 🌵 
https://github.com/lucas-zimerman/sentry-dotnet/runs/1406493108?check_suite_focus=true

src/Sentry.Log4Net/SentryAppender.cs
```diff
@@ -127,7 +127,7 @@ protected override void Append(LoggingEvent loggingEvent)
             _ = Hub.CaptureEvent(evt);
         }
 
-        private static IEnumerable<KeyValuePair<string, object>> GetLoggingEventProperties(LoggingEvent loggingEvent)
+        private static IEnumerable<KeyValuePair<string, object?>> GetLoggingEventProperties(LoggingEvent loggingEvent)
```
bruno-garcia, line 130: Not sure it makes sense to make the value nullable makes sense.
We can just not add the key at all

bruno-garcia (changes requested):

src/Sentry/Protocol/BaseScope.cs
```diff
@@ -142,11 +142,11 @@ public User User
         /// </summary>
         public
 #if NET451
```
bruno-garcia, line 144: We don't really support NET451 so this can be deleted

src/Sentry/Sentry.csproj
```diff
@@ -35,4 +35,8 @@
     <PackageReference Include="Microsoft.Bcl.AsyncInterfaces" Version="1.1.1" />
   </ItemGroup>
 
-</Project>
+  <PropertyGroup Condition="'$(TargetFramework)' == 'net5.0'">
+    <DefineConstants>NET5_0;$(AdditionalConstants)</DefineConstants>
+  </PropertyGroup>
```
bruno-garcia, line 40: This one should be there out of the box
```suggestion
```

src/Sentry/SentryClient.cs
```diff
@@ -234,16 +234,18 @@ private bool CaptureEnvelope(Envelope envelope)
...
+                    @event?.AddBreadcrumb(
+                        "BeforeSend callback failed.",
+                        category: "SentryClient",
+                        data: new Dictionary<string, string>
+                        {
+                            {"message", e.Message},
+                            {"stackTrace", e.StackTrace}
+                        },
+                        level: BreadcrumbLevel.Error);
+                }
```
bruno-garcia, line 248: Why throw everything away if there's no stack trace?


```suggestion
                var data = new Dictionary<string, string>
                {
                    {"message", e.Message},
                };
                if (e.StackTrace is not null)
                {
                    data.Add("stackTrace", e.StackTrace);
                }
                @event?.AddBreadcrumb(
                    "BeforeSend callback failed.",
                    category: "SentryClient",
                    data: data,
                    level: BreadcrumbLevel.Error);
```

test/Sentry.Tests/Sentry.Tests.csproj
```diff
@@ -17,4 +17,8 @@
     <PackageReference Include="Xunit.SkippableFact" Version="1.4.13" />
   </ItemGroup>
 
+  <PropertyGroup Condition="'$(TargetFramework)' == 'net5.0'">
+    <DefineConstants>NET5_0;$(AdditionalConstants)</DefineConstants>
+  </PropertyGroup>
```
bruno-garcia, line 22: Should be added OOTB

```suggestion
```

bruno-garcia (changes requested):

src/Sentry.Log4Net/SentryAppender.cs
lucas-zimerman: Decided that the change will stay.

bruno-garcia (approved):

PR #590, opened 2020-11-17
Replace BaseScope with IScope

Tyrrrz:
Rearranging scope data and cleaning up the code. Required for implementing transactions.
Some side effects of the changes:
- Changed `Breadcrumbs` from `ConcurrentQueue<T>` to a lock-synchronized `List<T>`. This is because `ConcurrentQueue<T>` doesn't implement `ICollection<T>` or any other common abstraction that could be used in `IScope`. Synchronized list is less efficient, but for small collections that don't get updated on a hot path it's actually really good.
- Removed `null` defaults for collections. As I understand it, this was only used to avoid serializing additional properties if they are unset. Replaced this behavior with a special attribute and a custom contract resolver instead. Previous workaround was also inefficient (serializer still dynamically read those properties, forcing them to initialize) and lead to unnecessary `null`s appearing in the code base. After this change, the collections are now null-safe as an additional benefit. I didn't extend this on objects like `Contexts` and `Request` because it's more difficult and probably not worth it. Ultimately, manual serialization fixes all those problems in a much more elegant way.


Tyrrrz:
@bruno-garcia

src/Sentry/Internal/SynchronizedList.cs
```diff
@@ -0,0 +1,122 @@
...
+        public SynchronizedList(List<T> innerList) =>
+            _innerList = innerList;
+
+        public SynchronizedList() : this(new List<T>()) {}
+
+        public IEnumerator<T> GetEnumerator()
+        {
+            lock (_lock)
+            {
+                return _innerList.GetEnumerator();
```
bruno-garcia, line 34: This enumerator now potentially going to break when the underlying collection mutates.
The safe approach here would be to do return `new List(_innerList)`.

src/Sentry/Internal/Json.cs
```diff
@@ -1,14 +1,52 @@
...
+            {
+                var jsonProperty = base.CreateProperty(member, memberSerialization);
+                var property = jsonProperty.DeclaringType.GetProperty(jsonProperty.UnderlyingName);
+
+                // DontSerializeEmpty
+                if (jsonProperty.ShouldSerialize is null &&
+                    property?.GetCustomAttribute<DontSerializeEmptyAttribute>() is {})
+                {
+                    // Collections
+                    if (property.PropertyType.GetInterfaces().Any(i => i == typeof(IEnumerable)))
```
bruno-garcia, line 32: The previous code only serialized members annotated which were only the private, nullable fields, so it isn't true that the serializer called the properties accessors.

That said this is a lot more expensive than the previous code though it doesn't matter given it's anyway executed in the background thread and it's still negligent perf impact.

src/Sentry/Protocol/IScope.cs
```diff
@@ -129,29 +85,22 @@ public User User
...
         /// <example> { "fingerprint": ["{{ default }}", "http://example.com/my.url"] } </example>
-        public IEnumerable<string> Fingerprint => InternalFingerprint ?? Enumerable.Empty<string>();
+        IEnumerable<string> Fingerprint { get; set; }
 
         /// <summary>
         /// A trail of events which happened prior to an issue.
         /// </summary>
         /// <seealso href="https://docs.sentry.io/platforms/dotnet/enriching-events/breadcrumbs/"/>
-        public IEnumerable<Breadcrumb> Breadcrumbs => InternalBreadcrumbs ??= new ConcurrentQueue<Breadcrumb>();
+        ICollection<Breadcrumb> Breadcrumbs { get; }
```
bruno-garcia, line 94: Collection allows for Add, this is not expected to let people add to directly (we have those `AddBreadcrumbs` API.
```suggestion
        IEnumerable<Breadcrumb> Breadcrumbs { get; }
```

src/Sentry/Protocol/IScope.cs
```diff
@@ -129,29 +85,22 @@ public User User
...
 
         /// <summary>
         /// An arbitrary mapping of additional metadata to store with the event.
         /// </summary>
-        public IReadOnlyDictionary<string, object?> Extra
-            => InternalExtra ??= new ConcurrentDictionary<string, object?>();
+        IDictionary<string, object?> Extra { get; }
```
bruno-garcia, line 99: Same here, not expecting to let get a hold of the mutable reference (of course they could type cast but that's beside the point).

The Scope in other SDKs is not even readable. This was made readable but not mutable (except through the APIs we defined like `SetExtra`

src/Sentry/Protocol/SentryEvent.cs
```diff
@@ -117,6 +121,72 @@ public class SentryEvent : BaseScope
...
+            set => _user = value;
+        }
+
+        /// <inheritdoc />
+        [DataMember(Name = "environment", EmitDefaultValue = false)]
+        public string? Environment { get; set; }
+
+        /// <inheritdoc />
+        [DataMember(Name = "sdk", EmitDefaultValue = false)]
+        public SdkVersion? Sdk { get; internal set; } = new SdkVersion();
```
bruno-garcia, line 168: Actually this shouldn't be nullable:

```suggestion
        public SdkVersion Sdk { get; internal set; } = new SdkVersion();
```

src/Sentry/Scope.cs
```diff
@@ -78,11 +82,76 @@ internal SentryId LastEventId
...
+
+        /// <inheritdoc />
+        [DataMember(Name = "fingerprint", EmitDefaultValue = false)]
+        [DontSerializeEmpty]
+        public IEnumerable<string> Fingerprint { get; set; } = Enumerable.Empty<string>();
+
+        /// <inheritdoc />
+        [DataMember(Name = "breadcrumbs", EmitDefaultValue = false)]
+        [DontSerializeEmpty]
+        public ICollection<Breadcrumb> Breadcrumbs { get; } = new SynchronizedList<Breadcrumb>();
```
bruno-garcia, line 139: I'm not a fan of this custom "synchronized list". Doesnt' ConcurrentBag and List both implement IEnumerable? 

The extension method is internal and can type cast to the type we know we support in both implementation.

src/Sentry/Scope.cs
```diff
@@ -78,11 +82,76 @@ internal SentryId LastEventId
...
+
+        /// <inheritdoc />
+        [DataMember(Name = "extra", EmitDefaultValue = false)]
+        [DontSerializeEmpty]
+        public IDictionary<string, object?> Extra { get; } = new ConcurrentDictionary<string, object?>();
```
bruno-garcia, line 144: In the old implementation these lists were not allocated at all. Now we get these two lists instatiated for each new Scope pushed/poped even if nothing is ever added

bruno-garcia (commented):
Left a few notes

src/Sentry/Internal/Json.cs
Tyrrrz: True, I forgot it ignored non-annotated properties.

src/Sentry/Scope.cs
Tyrrrz: IEnumerable doesn't have .Add().
ConcurrentBag implements ICollection but doesn't guarantee order.

bruno-garcia: We don't want `Add` exposed to the user. We need `Add` ourselves on the `AddBreadcrumb` method (before we just typecast to the right type)

bruno-garcia: Right about the order. I don't see the reason to change the way this was done before tbqh

src/Sentry/Internal/SynchronizedList.cs
```diff
@@ -0,0 +1,123 @@
+using System.Collections;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace Sentry.Internal
+{
+    // Inefficient, but okay for our use case
+    internal class SynchronizedList<T> : IList<T>
```
bruno-garcia, line 8: Can we go back to the built in ConcurrentQueue instead of having this?

src/Sentry/Scope.cs
Tyrrrz: I wanted to avoid leaky abstraction from `IScope`, because then `AddBreadcrumb` would be coupled to both `SentryEvent` and `Scope` at runtime without a way to validate its assumptions statically.

Tyrrrz: Another approach is to add `AddBreadcrumb` directly to `IScope`, which would eliminate the abstraction issues and would probably be the cleanest solution.

Tyrrrz: Anyway, I don't really care about this too much and I guess it's safer to just keep things how they were. Reverted it back to `ConcurrentQueue` on Scope.

bruno-garcia: OK adding to the interface sounds OK

bruno-garcia (approved):

Issue #594, opened 2020-11-18
Fix BeginScope<T> data binding to SentryEvent


bruno-garcia:
To be built on top of the latest 2.x. We'll release a new 2.x
Merge it back into `main`.

MEL `BeginScope` data needs to be kept even on ASP.NET Core (where scope lock is in place, in the Middleware).

https://github.com/getsentry/sentry-dotnet/blob/2693b265176be37f4b33966481bed66e9ec171b4/src/Sentry/Internal/SentryScopeManager.cs#L59-L64
Here we should add code that takes TState and adds to the event (or call a user callback, we need a hook to replace our implementation), that is able to process TState and add to the event.

`ApplyScopeStateToEvent<TState>(SentryEvent @event)`

Today the strategy is hard coded as such:
https://github.com/getsentry/sentry-dotnet/blob/2693b265176be37f4b33966481bed66e9ec171b4/src/Sentry/Protocol/BaseScopeExtensions.cs#L326-L360

This should stay as the default but be replaceble.

https://github.com/getsentry/sentry-dotnet/blob/2693b265176be37f4b33966481bed66e9ec171b4/src/Sentry.Extensions.Logging/SentryLogger.cs#L35

We need a hook so customers can serialize `TState` by themselves.
https://github.com/getsentry/sentry-dotnet/blob/2693b265176be37f4b33966481bed66e9ec171b4/src/Sentry/Internal/SentryScopeManager.cs#L59-L64

Change the sample: https://github.com/getsentry/sentry-dotnet/blob/2693b265176be37f4b33966481bed66e9ec171b4/samples/Sentry.Samples.ME.Logging/Program.cs#L52-L81

To show case using the hook to serialize TState into the Event.

bruno-garcia:
Shipped on the latest stable 2.1.6

PR #596, opened 2020-11-19
Hook for BaseScope.Apply

Tyrrrz:
Related to #594

@bruno-garcia I've looked at how existing `Func<IEnumerable<ISentryEventProcessor>>[]` and etc. work, but they don't make sense to me. Can you guide me in regards to what I should do to make it usable from ASP.NET Core?

### DIFF
--- a/src/Sentry.Protocol/BaseScopeExtensions.cs
+++ b/src/Sentry.Protocol/BaseScopeExtensions.cs
@@ -2,7 +2,6 @@ using System;
 using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.ComponentModel;
-using System.Linq;
 using Sentry.Protocol;
 
 // ReSharper disable once CheckNamespace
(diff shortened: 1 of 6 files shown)


Tyrrrz:
Open question: what should happen in the event of `scope.IsLocked`? Should the apply be called on the current scope? Should it only happen if the custom hook was provided?

bruno-garcia:
> Open question: what should happen in the event of scope.IsLocked? Should the apply be called on the current scope? Should it only happen if the custom hook was provided?

yeah we just set the data to the current scope.

src/Sentry.Protocol/BaseScopeExtensions.cs
```diff
@@ -308,38 +307,8 @@ public static void Apply(this BaseScope from, BaseScope to)
...
-                    {
-                        scope.SetTag(tupleStringString.Item1, tupleStringString.Item2);
-                    }
-                    break;
-#endif
-                default:
-                    scope.SetExtra("state", state);
-                    break;
-            }
+            var processor = scope.ScopeOptions.SentryScopeProcessor ?? new DefaultSentryScopeProcessor();
```
bruno-garcia, line 310: We'll allocate this processor for each call to `Apply`, what about we set this instance to the Options ctor (and make it non null), users can set a new one and replace it though

bruno-garcia (commented):
Looking good!

Given that the high level API is `X<TScope>(T arg` I wonder if we could do the same here and dump the `object` arg

src/Sentry.Protocol/BaseScopeExtensions.cs
Tyrrrz: It is already like that on `SentryOptions`, but we can't guarantee it on the interface level (`IScopeOptions`) unless we unconditionally trust that user will respect NRTs. The RHS of `??` should not be executed majority of the time.

PR #605, opened 2020-11-22
Avoid deleting the cache if there's no connection.

lucas-zimerman:
Previously, cached envelopes were being deleted if there was a connection issue, for example, having no connection.

Example:

image
image

Test case: 
* Tested with the Xamarin Sample:  https://github.com/lucas-zimerman/Sentry.Samples.Mobile
* Airplane mode throws a HttpRequestException with a Socket inner exception.
* Connection with a proxy without connection throws HttpRequestException with Socket inner exception.


lucas-zimerman:
@bruno-garcia  new PR created https://github.com/getsentry/sentry-dotnet/pull/610

PR #613, opened 2020-11-22
fix: runtime current is equal

bruno-garcia:
The field is lazily instantiated but makes no sense to synchronize, so matching for Same here means a flaky test with little value

### DIFF
--- a/test/Sentry.Tests/PlatformAbstractions/RuntimeTests.cs
+++ b/test/Sentry.Tests/PlatformAbstractions/RuntimeTests.cs
@@ -6,9 +6,9 @@ namespace Sentry.Tests.PlatformAbstractions
     public class RuntimeTests
     {
         [Fact]
-        public void Current_SameInstance()
+        public void Current_Equal()
         {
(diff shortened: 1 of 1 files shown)


test/Sentry.Tests/PlatformAbstractions/RuntimeTests.cs
```diff
@@ -6,9 +6,9 @@ namespace Sentry.Tests.PlatformAbstractions
     public class RuntimeTests
     {
         [Fact]
-        public void Current_SameInstance()
+        public void Current_Equal()
         {
-            Assert.Same(Runtime.Current, Runtime.Current);
+            Assert.Equal(Runtime.Current, Runtime.Current);
```
Tyrrrz, line 11: From an outside perspective, this looks like a very superfluous test. What does it verify actually?

bruno-garcia: It verified the data was cached, but now with the change it doesn't do that so I might as well have deleted it

PR #616, opened 2020-11-22
deserialize event

bruno-garcia:
Resolves #614
Resolves #615

We need to figure out:

1. Why assigning to SentryException fail to deserialize
2. Deserializing `Contexts` is tricky since it correctly adds to the map, but it's unaware that a well known key such as `browser` should actually deserialize to `Browser`. We need a custom serializer.
3. Complete all Serialization tests into a two-way serialize->deserialize
4. Have snapshot tests to make sure that we know when we break compatibility with already serialized events (that will exist on devices once we ship this).

PR #620, opened 2020-11-23
show snapshot testing

SimonCropp:
figured i would do a small demo of how snapshot testing could work in this project.

Thoughts?

#skip-changelog


bruno-garcia:
@SimonCropp I wrote about that yesterday, it would be great to get some for the serialization stuff indeed.

Since I'm afraid testing a round trip only will mean we might break compatibility with already serialized data and wouldn't really catch in CI.

Thanks for raising this.

test/Sentry.Tests/SentryClientTests.CaptureEvent_BeforeEventThrows_ErrorToEventBreadcrumb.verified.txt
```diff
@@ -0,0 +1,14 @@
+[
+  {
+    timestamp: DateTimeOffset_1,
```
bruno-garcia, line 3: [We've identified some problems with deserialization of the protocol](https://github.com/getsentry/sentry-dotnet/pull/616/).

Not sure it's related to this, I assume `Verify` just compares the loaded json at the end and doesn't rely on the type being able to deserilize though.

SimonCropp:
no worries. also, where did u write about it?

do u want me to extend this pr to any other scenarios?

test/Sentry.Tests/SentryClientTests.CaptureEvent_BeforeEventThrows_ErrorToEventBreadcrumb.verified.txt
bruno-garcia: Verify serializes the type here to avoid having to match on dates?

bruno-garcia:
@SimonCropp What about a fork of that draft PR I mentioned, you can PR back into that?
There I've pointed out the places we need to cover round trip serialization. I guess adding the snapshot there in combination might be the way to go?

We're interested in the exact format the data serializes too (like millisecond precision etc)

bruno-garcia:
But any some examples are already highly appreciated.

I'm really hoping @Tyrrrz will be able to give me a hand with the serialization stuff since it's quite a bit of work to get it fixed it seems.

test/Sentry.Tests/SentryClientTests.CaptureEvent_BeforeEventThrows_ErrorToEventBreadcrumb.verified.txt
SimonCropp: Verify automatically scrubs any data that might be unstable between testruns/machines/build directory. all of these can be disabled if u want https://github.com/VerifyTests/Verify/blob/master/docs/serializer-settings.md#guids-are-scrubbed

SimonCropp: for guids and datetimes, they are detected and replaced with consistent values. so, for example, a given datetime will always be replaced with the same text.

bruno-garcia: Got it. Thanks for the pointer. I think we'll disable the scrubbing for these and make sure we use consistent values (we already do that actually)

Tyrrrz:
I was thinking earlier about snapshot testing when working on this:

https://github.com/getsentry/sentry-dotnet/blob/7e6d797b35ccbd9c99e07eef16dbf4394e94b210/test/Sentry.Tests/Protocol/SentryEventTests.cs#L45-L68

There are a couple tests like this that could really benefit from it.

bruno-garcia:
Agreed. We should delete all these hard coded things but for those tests we need the snapshot to save the exact serialized form of the object. The goal is to verify if we broke compatibility with already serialized files in devices using the SDK.

bruno-garcia:
Well also to make sure the format is valid in terms of the protocol but that the test itself won't do, we need to validate the json

Tyrrrz:
> Well also to make sure the format is valid in terms of the protocol but that the test itself won't do, we need to validate the json

Well, the test itself won't be any different from what's already there. The only difference is that the JSON string we compare to will be stored in a file rather than as a literal string in code. And that file will be updated semi-automatically. But other than that, the tests and what they accomplish, will be pretty much the same.

test/Sentry.Tests/SentryClientTests.CaptureEvent_BeforeEventThrows_ErrorToEventBreadcrumb.verified.txt
SimonCropp: i checked and the datetime is not consistent between runs. should it be?

SimonCropp:
> What about a fork of that draft PR I mentioned, you can PR back into that?

I think merge this. then pull these changes into that pr, then i can look at those scenarios with the snapshot testing in place

test/Sentry.Tests/SentryClientTests.CaptureEvent_BeforeEventThrows_ErrorToEventBreadcrumb.verified.txt
bruno-garcia: Yes, through the code base types have at least an internal overload that takes timestamp, guid etc just for testing sake.

Since we don't have round trip tests so far, it's possible it creates a UtcNow for example but then asserts against that. Going forward, for the sake of snapshot tests that approach can't be used anymore and a fixed date should be used (that includes some level of precision to cover the scenario of millseconds when relevant)

bruno-garcia:
> I think merge this. then pull these changes into that pr, then i can look at those scenarios with the snapshot testing in place

We could merge but the test didn't pass

Tyrrrz:
Looks like the failing tests are related to missing method issues with Newtonsoft.Json. Can you please rebase on latest `main` as we've recently got rid of Newtonsoft.Json? This should also fix those errors.

SimonCropp:
ok. i think i have ironed out all the issues.

lucas-zimerman:
The Danger file is happening because no changes were made on changelog. If it's not required you can add  #skip-changelog

 on the PR's first comment.

bruno-garcia (approved):
Thanks @SimonCropp for showing off `Verify` (and `ModuleInitializer`)!

SimonCropp:
ok what the next area you think would benefit from snapshot testing?

bruno-garcia:
Maybe @Tyrrrz has more ideas?
I was thinking more related to the serialization but there the use case is more literal, so what the snapshot value is would really matter. like actual date format etc

Tyrrrz:
Yeah, I can think of only the JSON tests. Not sure if this PR covered all of them, but there were definitely a lot last time I checked 😃 
Basically you can search for `ToJsonString()` in tests and find them: https://github.com/getsentry/sentry-dotnet/search?q=.ToJsonString

PR #621, opened 2020-11-24
Add IDiagnosticLogger that routes to ITestOutputHelper

### LINKED ISSUE #612: Capture IDiagnosticLogger during testing
Some of the tests are rather complex and test a mix of static state with filesystem.
It's hard to debug flaky tests without knowing what the SDK was doing.

We need a way to set the options.DiagnosticLogger to a type that will accumulate logs in memory and in the event of a failing test, dump them all ton the test output (via xunit `ITestOutputHelper` I'd assume).

Ideally nothing would go into `ITestOutputHelper` unless the test fails. If xunit had that control instead, it would be simpler (we map `IDiagnosticLogger` to `ITestOutputHelper` and let xunit only print logs for failing tests instead of spamming the build output)

Tyrrrz:
In some tests we use a fake/mocked logger so we might need to add an `AggregateDiagnosticLogger` abstraction to accommodate those. Anyway, this is a prototype.

Closes #612

### DIFF
--- a/test/Sentry.Testing/TestOutputDiagnosticLogger.cs
+++ b/test/Sentry.Testing/TestOutputDiagnosticLogger.cs
@@ -1,0 +1,32 @@
+using System;
+using Sentry.Extensibility;
+using Xunit.Abstractions;
+
+namespace Sentry.Testing
+{
+    public class TestOutputDiagnosticLogger : IDiagnosticLogger
+    {
+        private readonly ITestOutputHelper _testOutputHelper;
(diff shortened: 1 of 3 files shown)


bruno-garcia (approved):
sweet!!

bruno-garcia:
We could probly add to HubTests, SentrySdkTests, SentryClientTests?

bruno-garcia:
@Tyrrrz can we merge this?

Tyrrrz:
@bruno-garcia yes. We can add it to other places later.

PR #623, opened 2020-11-24
Manual serialization

Tyrrrz:
Gets rid of JSON.net, adds dependency on `System.Text.Json` (except on .NET 5.0 target; STJ is not included in .NET Standard).

Based on #616

#skip-changelog


src/Sentry/Internal/Extensions/JsonExtensions.cs
```diff
@@ -0,0 +1,32 @@
...
+namespace Sentry.Internal.Extensions
+{
+    internal static class JsonExtensions
+    {
+        public static void WriteDictionary(
+            this Utf8JsonWriter writer,
+            string propertyName,
+            IReadOnlyDictionary<string, string?>? dic)
+        {
+            writer.WritePropertyName(propertyName);
```
bruno-garcia, line 13: I guess avoiding the while object (not writing prop name) if there are not values is the behavior we want (so far hacked with the null backing fields, which for bonus avoideed allocations that weren't needed/lazy init)

src/Sentry/Protocol/SdkVersion.cs
```diff
@@ -62,6 +51,57 @@ public void AddPackage(string name, string version)
...
+            // Packages
+            var packages = new List<Package>();
+            foreach (var packageJson in json.GetProperty("packages").EnumerateArray())
+            {
+                var package = Package.FromJson(packageJson);
+                packages.Add(package);
+            }
+
+            // Name
+            var name = json.GetProperty("name").GetString() ?? "";
```
bruno-garcia, line 94: What about
```suggestion
            var name = json.GetProperty("name").GetString() ?? "dotnet.unknown";
```

src/Sentry/Protocol/SdkVersion.cs
```diff
@@ -62,6 +51,57 @@ public void AddPackage(string name, string version)
...
+
+            // Version
+            var version = json.GetProperty("version").GetString() ?? "";
```
bruno-garcia, line 97:
```suggestion
            var version = json.GetProperty("version").GetString() ?? "0.0.0";
```

bruno-garcia (commented):
This is looking promising.

One suggestion: Do some "Read" too first, so we can see what that looks like, before doing all Write methods for all classes

Tyrrrz:
Note for future: we should probably allow XML docs to be omitted, at least when building on CI from a non-master branch

src/Sentry/Internal/Extensions/EnumExtensions.cs
```diff
@@ -0,0 +1,12 @@
+using System;
+
+namespace Sentry.Internal.Extensions
+{
+    internal static class EnumExtensions
+    {
+        public static T ParseEnum<T>(this string str) where T : struct, Enum
+        {
+            return (T)Enum.Parse(typeof(T), str, true);
+        }
```
bruno-garcia, line 10: Gave up on this?

```suggestion
        public static T ParseEnum<T>(this string str) where T : struct, Enum =>
            (T)Enum.Parse(typeof(T), str, true);
```

bruno-garcia (commented):
@lucas-zimerman looks like we can try offline caching from this branch

🥳 👯 👯‍♀️ 😄

src/Sentry/Internal/Extensions/EnumExtensions.cs
Tyrrrz: No, just didn't think too much :)

Tyrrrz:
also note for the future: we should consider which protocol classes we may want to delete.
previously they were 1:1 to sentry data types, but that made working with them in code a bit annoying.
now, since we have manual serialization, we no longer need our class structure to reflect the object structure on sentry, so we can clean things up.

Tyrrrz:
This is still work in progress, but can be reviewed for the most part.

src/Sentry/Protocol/SdkVersion.cs
```diff
@@ -83,18 +85,15 @@ public void WriteTo(Utf8JsonWriter writer)
...
-                var package = Package.FromJson(packageJson);
-                packages.Add(package);
-            }
+            var packages =
+                json.GetPropertyOrNull("packages")?.EnumerateArray().Select(Package.FromJson).ToArray() ??
+                Array.Empty<Package>();
 
             // Name
-            var name = json.GetProperty("name").GetString() ?? "dotnet.unknown";
+            var name = json.GetPropertyOrNull("name")?.GetString() ?? "dotnet.unknown";
```
bruno-garcia, line 97: Not that you've set it on the property, maybe a private const to reuse the value (and avoid going out of sync in the future?)

src/Sentry/Protocol/UserFeedback.cs
```diff
@@ -43,5 +39,43 @@ public UserFeedback(SentryId eventId, string email, string comments, string? nam
...
+            }
+
+            writer.WriteEndObject();
+        }
+
+        public static UserFeedback FromJson(JsonElement json)
+        {
+            var eventId = json.GetPropertyOrNull("event_id")?.Pipe(SentryId.FromJson) ?? SentryId.Empty;
+            var name = json.GetPropertyOrNull("name")?.GetString() ?? "";
+            var email = json.GetPropertyOrNull("email")?.GetString() ?? "";
```
bruno-garcia, line 79: This is nullable, so rather let it be null than give it an empty string.

Also I'm not sure it makes sense to give it an instance full of empty strings if the vallues weren't there (likely a bug since u can't create such instance).

Maybe best to return null.

Tyrrrz: It's not nullable on the class

Tyrrrz:
@bruno-garcia tests are passing.
Please use this opportunity to look through the tests and see if you want to change/add something. I know you left them in a intermediate state on your branch when I checked out.

src/Sentry/PlatformAbstractions/RuntimeInfo.cs
```diff
@@ -83,7 +83,7 @@ internal static void SetNetCoreVersion(Runtime runtime)
             {
                 // https://github.com/dotnet/BenchmarkDotNet/issues/448#issuecomment-308424100
                 var assembly = typeof(System.Runtime.GCSettings).GetTypeInfo().Assembly;
-#if NET5_0
+#if NET5_0 || NETCOREAPP3_0
```
bruno-garcia, line 86: if not now, when net6 lands and we come to this line again, it'll make more sense to flip: `if !NETFX && !NETSTANDARD` :)

src/Sentry/Protocol/Context/Contexts.cs
```diff
@@ -78,5 +83,57 @@ internal void CopyTo(Contexts to)
                 to.TryAdd(kv.Key, value);
             }
         }
+
+        /// <inheritdoc />
+        public void WriteTo(Utf8JsonWriter writer)
```
bruno-garcia, line 88: =>

src/Sentry/Protocol/Context/Contexts.cs
```diff
@@ -78,5 +83,57 @@ internal void CopyTo(Contexts to)
...
+        {
+            writer.WriteDictionaryValue(this!);
```
bruno-garcia, line 90: Curious why does it need `!`?

src/Sentry/Protocol/Context/Device.cs
```diff
@@ -222,5 +191,202 @@ internal Device Clone()
...
+
+            writer.WriteEndObject();
+        }
+
+        /// <summary>
+        /// Parses from JSON.
+        /// </summary>
+        public static Device FromJson(JsonElement json)
+        {
+            var timezone = json.GetPropertyOrNull("timezone")?.GetString()?.Pipe(TimeZoneInfo.FindSystemTimeZoneById);
```
bruno-garcia, line 336: We really need a test for this given that tz information on linux is very different than windows. Even though we don't have a requirement now to get an envelope serialized on windows to be deserialized on linux, I just want to make sure it round trips on each OS independently

src/Sentry/Protocol/Exceptions/SentryStackTrace.cs
```diff
@@ -29,5 +29,38 @@ public IList<SentryStackFrame> Frames
...
+        /// </summary>
+        public static SentryStackTrace FromJson(JsonElement json)
+        {
+            var frames = json.GetPropertyOrNull("frames")?.EnumerateArray().Select(SentryStackFrame.FromJson).ToArray();
+
+            return new SentryStackTrace
+            {
+                InternalFrames = frames
+            };
+        }
```
bruno-garcia, line 64:
```suggestion
        public static SentryStackTrace FromJson(JsonElement json) =>
            new SentryStackTrace
            {
                InternalFrames = json.GetPropertyOrNull("frames")?.EnumerateArray().Select(SentryStackFrame.FromJson).ToArray();
            };
```

src/Sentry/Protocol/IJsonSerializable.cs
```diff
@@ -0,0 +1,115 @@
...
+        /// Writes the object as JSON.
+        /// </summary>
+        void WriteTo(Utf8JsonWriter writer);
+    }
+
+    internal static class JsonSerializableExtensions
+    {
+        public static string ToJsonString(this IJsonSerializable serializable)
+        {
+            using var stream = new MemoryStream();
```
bruno-garcia, line 23: This is just for testing, right?

src/Sentry/Protocol/IJsonSerializable.cs
```diff
@@ -0,0 +1,115 @@
...
+        {
+            value.WriteTo(writer);
+        }
+
+        public static void WriteSerializable(
+            this Utf8JsonWriter writer,
+            string propertyName,
+            IJsonSerializable value)
+        {
+            writer.WritePropertyName(propertyName);
```
bruno-garcia, line 44: Avoid the prop altogether?

I can't think of Sentry features that accept just the key doesn't complain that there's no value

src/Sentry/Protocol/Package.cs
```diff
@@ -37,5 +34,36 @@ public Package(string name, string version)
...
+
+            writer.WriteEndObject();
+        }
+
+        /// <summary>
+        /// Parses from JSON.
+        /// </summary>
+        public static Package FromJson(JsonElement json)
+        {
+            var name = json.GetProperty("name").GetString() ?? "";
```
bruno-garcia, line 63: I'm not a fan of these `""` ..

It'll be like: wtf is this empty string and where did it come from.

The type takes the values in the constructor, so a serialized value wtihout them is broken.
I rather we return null here and communicate there's no Package. If that isn't an option the values would be: "unknown" or something clear like that

src/Sentry/Protocol/SdkVersion.cs
```diff
@@ -62,6 +53,64 @@ public void AddPackage(string name, string version)
...
+
+        /// <summary>
+        /// Parses from JSON.
+        /// </summary>
+        public static SdkVersion FromJson(JsonElement json)
+        {
+            // Packages
+            var packages =
+                json.GetPropertyOrNull("packages")?.EnumerateArray().Select(Package.FromJson).ToArray() ??
+                Array.Empty<Package>();
```
bruno-garcia, line 100: O_O

```suggestion            json.GetPropertyOrNull("packages")?.EnumerateArray().Select(Package.FromJson).ToArray()
    ?? Array.Empty<Package>();
```

src/Sentry/Protocol/UserFeedback.cs
bruno-garcia: I mean `Name`.

These empty strings are not great. can't we return null and log an error here?
We got a broken file in this case. Sending `""` to Sentry because the value is required is such a hack

src/Sentry/Sentry.csproj
```diff
@@ -1,38 +1,41 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFrameworks>net5.0;netstandard2.1;netstandard2.0;net461</TargetFrameworks>
+    <TargetFrameworks>net5.0;netcoreapp3.0;netstandard2.1;netstandard2.0;net461</TargetFrameworks>
```
bruno-garcia, line 4: We need to make sure this TFM is tested now.

i.e:

```
  <ItemGroup Condition="'$(TargetFramework)' == 'netcoreapp3.1'">
    <ProjectReference Include="../../src/Sentry/Sentry.csproj" Properties="TargetFramework=netcoreapp3.0" />

and
  <ItemGroup Condition="'$(TargetFramework)' == 'netcoreapp3.0'">
    <ProjectReference Include="../../src/Sentry/Sentry.csproj" Properties="TargetFramework=netstandard2.1" />

```

If there's a way to get netcoreapp3.1 to run twice, once against 1 target and once against the other, we're good.

bruno-garcia (approved):
Should we `sealed` all protocol classes? They are all serializable and we won't serialize random inherited filds anymore

src/Sentry/Protocol/IJsonSerializable.cs
Tyrrrz: Currently yes, but I thought it would also be useful for debug logging (i.e. log the event content).

src/Sentry/Protocol/Context/Device.cs
Tyrrrz: There are tests that call `FromJson` on `Device` already.

src/Sentry/Protocol/Context/Contexts.cs
Tyrrrz: No idea :/

Tyrrrz: image

No idea. The receiving type is even more accepting. I think nullability covariance is just buggy in C#.

src/Sentry/Internal/Extensions/JsonExtensions.cs
```diff
@@ -0,0 +1,145 @@
...
+            }
+        }
+
+        public static void WriteDictionary(
+            this Utf8JsonWriter writer,
+            string propertyName,
+            IEnumerable<KeyValuePair<string, object?>>? dic)
+        {
+            writer.WritePropertyName(propertyName);
+            writer.WriteDictionaryValue(dic);
```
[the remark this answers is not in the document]
Tyrrrz: will reafctor later

PR #632, opened 2020-11-30
Increase codecov threshold

### LINKED ISSUE #622: Tell codecov to pass PR if diff is not -5% or less
image
Useless ^

Can we configure codecov to pass the build with a resonable value like anything above -5% or something?

Tyrrrz:
Closes #622

Not sure why `patch` was `false` previously. The codecov documentation doesn't show an example which sets that variable to `false`.

#skip-changelog

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -1,6 +1,11 @@
 coverage:
   status:
-    patch: false
     changes: false
+    project:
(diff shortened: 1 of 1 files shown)


bruno-garcia (approved):
awesome! thanks

Issue #502, opened 2020-09-07
Support for Release Health and Performance Monitoring


bddckr:
Please mark the type framework used:

- [ ] ASP.NET MVC
- [ ] ASP.NET Web API (OWIN)
- [ ] ASP.NET Core
- [ ] WPF
- [x] WinForms
- [x] Xamarin
- [ ] Other:

Please mark the type of the runtime used:

- [ ] .NET Framework
- [x] Mono
- [x] .NET Core

* Version: The latest stable for both, that is .NET Core 3.1.401 and Xamarin.Android 11.0.2.0.

Please mark the NuGet packages used:

- [ ] Sentry
- [x] Sentry.Serilog
- [ ] Sentry.NLog
- [ ] Sentry.Log4Net
- [x] Sentry.Extensions.Logging
- [ ] Sentry.AspNetCore

* Version: 2.1.6 of both.

---

It looks like both [Release Health](https://docs.sentry.io/product/releases/health/) and [Performance Monitoring](https://docs.sentry.io/product/performance/) aren't (directly?) supported.

Is there any way for me to utilize these platform features, perhaps with the underlying `Sentry.Protocol` package? I'm totally fine with having to identify and generate session details myself, as well as performance data.

Ultimately I'm interested in filling the Sentry release's user information, as well as gain some insight into our client apps' performance when it comes to hitting our servers.

bruno-garcia:
@bddckr we're planning to start working towards adding both release health and performance support to .NET soon.

@lucas-zimerman created already some support for both features on an external library: https://github.com/lucas-zimerman/sentry-dotnet-performance-addon

We'll use this issue to track progress.

bruno-garcia:
You've selected WinForms and Xamarin. What do you use on the server?

bddckr:
We're using Sentry on the server and our web frontend, too. I wasn't mentioning this as I'm primarily interested in device-only insight for now and I'm not yet sure on buy-in from the rest of the teams for any potential cross-team insight we might be able to get via Sentry.

For my purposes (and after just a quick glance for now) it looks like Lucas' library does exactly what I am after! Thanks for the pointer and great to hear it's on the roadmap.

bruno-garcia:
Just to clarify. You're looking to use this on a Xamarin and WinForms app then, correct?

To help us prioritize things, could you describe a bit the targets? The WinForms app, is it .NET Framework or core? And what versions? Xamarin iOS and Android or something else?

[...]

PR #633, opened 2020-11-30
Support performance

Tyrrrz:
#502 (not sure if this will enable release health as well?)


Tyrrrz:
I know the interface is really exposed here and ideally we would want to add some methods to make things nicer (i.e. set timestamps automatically, etc.).

For now, I want to wire this on Sentry.AspNetCore side first. I have never done that, any guidance on how should I approach this? @bruno-garcia  @lucas-zimerman

Tyrrrz:
@bruno-garcia

Sentry docs say:
> A Transaction is basically a Span combined with an Event.

Is it true in regards to our event structure? Or should it just implement `IScope`?

Tyrrrz:
As there any additional options that need to be added into `SentryOptions` for this?

lucas-zimerman:
> @bruno-garcia
> 
> Sentry docs say:
> 
> > A Transaction is basically a Span combined with an Event.
> 
> Is it true in regards to our event structure? Or should it just implement `IScope`?

True, the transaction it's basically 'IScope' with some extras (a list of pans, StartTimestamp and some functions for creating a child).
Now I'm unsure if the transaction is going to be consumed by the existing event processors + before send the event, like some platforms.

lucas-zimerman:
If it helps https://github.com/lucas-zimerman/ContribSentryDotNet/blob/ref/better_tracing/ContribSentry/Internals/SpanStatus.cs
Here's the list of Span/Transaction status (The exception part wasn't defined by the Doc so I was experimenting with it)

lucas-zimerman:
> As there any additional options that need to be added into `SentryOptions` for this?
> 
> Looks like Java SDK added these. Anything else?
> 
> image

for each transaction, I was adding a "transaction Breadcrumb", if you're also going to add the breadcrumb for every transaction, it would be nice to have an option to disable the transaction breadcrumb

Breadcrumb example: `SentrySdk.AddBreadcrumb(tracing.EventId.ToString(), "sentry.transaction");`

Tyrrrz:
@lucas-zimerman does your project also have AspNetCore integration for transactions?

lucas-zimerman:
> @lucas-zimerman does your project also have AspNetCore integration for transactions?

There's no middleware for it but it should work just fine if you manually call the StartTransaction where you would like to measure the performance.

All you need is an AsyncLocal control and a way for controlling when it starts and when it ends (with or without any error).

```
        public Task StartTransaction(string name, string method, Action<ISentryTracing> action)
            => Task.Run(() =>
            {
                ISentryTracing tracing = DisabledTracing.Instance;
                if(CurrentTransaction.Value is null)
                {
                    CurrentTransaction.Value = new SentryTracing(name, method);
                    tracing = CurrentTransaction.Value;
                    tracing.Finish();
                }
                try
                {
                    action(tracing);
                }
                catch(Exception ex)
                {
                    tracing.Finish(ex);
                    throw;
                }
            });
```
https://github.com/lucas-zimerman/ContribSentryDotNet/blob/d049f3fe630ffead4636c33b30c0dca7315cecae/ContribSentry/Internals/TransactionWorker.cs#L35

lucas-zimerman:
> #502 (not sure if this will enable release health as well?)

Only Performance, release health is another pack of new classes.

src/Sentry/Protocol/Span.cs
```diff
@@ -0,0 +1,105 @@
...
+        public SentryId TraceId { get; set; }
+        public DateTimeOffset StartTimestamp { get; set; }
+        public DateTimeOffset EndTimestamp { get; set; }
+        public string? Operation { get; set; }
+        public string? Description { get; set; }
+        public string? Status { get; set; }
+        public bool IsSampled { get; set; }
+
+        private ConcurrentDictionary<string, string>? _tags;
+        public IReadOnlyDictionary<string, string> Tags => _tags ??= new ConcurrentDictionary<string, string>();
```
bruno-garcia, line 23: I assume concurrency here might be more of an issue so this approach of instantiating on the getter won't work well here.

We might need to synchronize in order to make sure concurrent calls to `get` get the same instance. Maybe a double check lock will do

src/Sentry/Protocol/Transaction.cs
```diff
@@ -0,0 +1,112 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text.Json;
+using Sentry.Internal.Extensions;
+
+namespace Sentry.Protocol
+{
+    public class Transaction : ISpan, IJsonSerializable
```
bruno-garcia, line 9: Transaction is actually a whole Event :(

That's how it is today in Sentry, basically an event with a `contexts.trace` field and spans

We can try to go with `ISpan` for now and see what it looks like.

bruno-garcia (commented):
> #502 (not sure if this will enable release health as well?)

Let's focus on Performance only for now.

> Is it true in regards to our event structure? Or should it just implement IScope?

Yeah technically a Transaction is a SentryEvent but we don't want BeforeSend to run for example (and hopefully no EventProcessors too), but we might have to let EventProcessors run if we need to align with other SDKs that did that (by accident).

> As there any additional options that need to be added into SentryOptions for this?


Those two. We can add that later, lets go without sampling in this PR to make smaller chunks but keep that in mind.

src/Sentry/Protocol/Span.cs
lucas-zimerman: So Spans have tags? I was thinking that only Transactions could have tags

src/Sentry/Protocol/ISpan.cs
```diff
@@ -0,0 +1,32 @@
...
+    {
+        SentryId SpanId { get; set; }
+
+        SentryId? ParentSpanId { get; set; }
+
+        SentryId TraceId { get; set; }
+
+        DateTimeOffset StartTimestamp { get; set; }
+
+        DateTimeOffset EndTimestamp { get; set; }
```
lucas-zimerman, line 16: EndTimestamp is Timestamp unless it was changed recently.

bruno-garcia: `timestamp` is the serialized name, I'd name the property `EndTimestamp` not to suck so much from the API point of view

Tyrrrz: What do you mean recently?

src/Sentry/Protocol/Transaction.cs
Tyrrrz: What is `contexts.trace` actually? I've seen it referenced before but still not sure.

Also, "transaction is actually a whole Event", does that affect how we should design the class? We probably can't make use of inheritance anyway due to manual serialization. I'm thinking of doing `public class Transactions : ISpan, IScope`, would  that be enough?

src/Sentry/Protocol/ISpan.cs
bruno-garcia: nothing changed. `EndTimestamp` is the C# type, but serialize this as `timestamp` (which is the sentry event protocol for timestamp)

src/Sentry/Protocol/Span.cs
Tyrrrz: Maybe just eagerly initialize the instance? It will probably be a smaller overhead than a lock.

Tyrrrz:
@bruno-garcia regarding aspnetcore middleware, do we programmatically add one before endpoint routing middleware and then measure the execution time of `await next()`? Or something more clever?

src/Sentry/Protocol/Span.cs
bruno-garcia: one will incur overhead on each new span due to allocating the list then GC'ing it. The other one will cost when accessing it only and only the first time due to the double check lock.

Without measuring we're just guessing, the the transaction I'm happy with always having the list there since more often than not we'll use it, for individual Spans, I don't know

src/Sentry/Protocol/Transaction.cs
bruno-garcia: we could inherit and call the base serializer to add some stuff and the Transaction one to add the rest.

In Java it was split in 2 parts, SentryEvent (current one) and Transaction, and both inherit from some BaseEvent.

bruno-garcia:
> @bruno-garcia regarding aspnetcore middleware, do we programmatically add one before endpoint routing middleware and then measure the execution time of `await next()`? Or something more clever?

That sounds like it would work yeah

Tyrrrz:
@bruno-garcia do we create some spans automatically for nested operations using the middleware? Or just let the user do it themselves.

bruno-garcia:
@Tyrrrz any outgoing HTTP request, SQL queries would eventually need to become spans, and automatically. We'll get there eventually though, for now if we can instrument some areas like Controller action invocation, and HTTP requests only would bea great start

src/Sentry/Protocol/ISpan.cs
lucas-zimerman: oh right, I was too sleepy to understand code.

Tyrrrz: Ah, I misunderstood the original comment as well.

src/Sentry/Protocol/Transaction.cs
Tyrrrz: > we could inherit and call the base serializer to add some stuff and the Transaction one to add the rest.

This will be messy because it will have to add serialization logic before the `writer.WriteEndObject()`.

src/Sentry/Protocol/Span.cs
Tyrrrz: > The other one will cost when accessing it only and only the first time due to the double check lock.

Maybe I misunderstood your suggestions, but how will that work? From my understand, the lock has to be on the getter, which would mean it's acquired/checked every time the property is accessed.

bruno-garcia: https://en.wikipedia.org/wiki/Double-checked_locking

src/Sentry/Protocol/Transaction.cs
bruno-garcia: we could call a `protected virtual AddMoreStuff(writer)` and implement this in the derived type

src/Sentry/Protocol/Span.cs
Tyrrrz: @lucas-zimerman the docs say so, but I also found the docs to be incredibly unreliable too, so...
https://develop.sentry.dev/sdk/event-payloads/span/

bruno-garcia: If the docs are not good, we should fix it. Please open a PR there or at least raise a flag (open an issue asking for clarification).

Tyrrrz:
@bruno-garcia where should the current transaction be kept? Should it be kept on the current scope? Or should we create a new `AsyncLocal` for transaction?

Tyrrrz:
@bruno-garcia @lucas-zimerman can spans be infinitely nested within other spans? or is it just the transaction that can have spans?

lucas-zimerman:
> @bruno-garcia @lucas-zimerman can spans be infinitely nested within other spans? or is it just the transaction that can have spans?

Each Span could have N childrens
Example
image

Not sure if the Doc specifies some limit but I'd guess the only limit is the request size.

bruno-garcia:
Transaction class has a list of Spans, they are linked in Sentry with their Ids.. the SDK doesn't hold a tree structure.

The Scope can hold a Transaction. This was done in the Java SDK, you can peek the implementation there.

The Ruby PR is less than a 1k lines: https://github.com/getsentry/sentry-ruby/pull/1108

bruno-garcia:
We also need to consider that `AsyncLocal` will need to change later, to better serve Desktop and Mobile.

Tyrrrz:
@bruno-garcia 

> Transaction class has a list of Spans, they are linked in Sentry with their Ids.. the SDK doesn't hold a tree structure.

Ok, so the spans can also have spans inside, but the class representation itself doesn't keep track of the list, right?

> The Scope can hold a Transaction. This was done in the Java SDK, you can peek the implementation there.

Ok, cool.

> If the docs are not very clear, we need to improve the docs.

If you look at the transaction/span docs, they are all over the place. The list of fields does not match the example at the end. In docs it says `event_id` while in reality it's `span_id`. The example on transaction docs actually show the example from span docs. Etcetera.

Tyrrrz:
Also, @bruno-garcia 

`IScope` and `Scope` already have `Transaction`, which is a string. This looks like it corresponds to `Transaction.Name`. Should we remove this and replace with a transaction? Or do something else?

image

bruno-garcia:
> If you look at the transaction/span docs, they are all over the place. The list of fields does not match the example at the end. In docs it says `event_id` while in reality it's `span_id`. The example on transaction docs actually show the example from span docs. Etcetera.

Would be nice to get this fixed. Do you mind opening a PR?

bruno-garcia:
Today the `Transaction` field is used by the error product. In Java, it's still there for the Event:

https://github.com/getsentry/sentry-java/blob/3ed2ff105dd246cbf0242e802543c710bcb14a0b/sentry/src/main/java/io/sentry/SentryEvent.java#L89

As I said above, there are two classes: `SentryEvent` and `Transaction`. From Sentry's point of view, both are Events.

src/Sentry.AspNetCore/SentryMiddleware.cs
```diff
@@ -104,7 +104,9 @@ public async Task InvokeAsync(HttpContext context)
                     // event creation will be sent to Sentry
 
                     scope.OnEvaluating += (_, __) => PopulateScope(context, scope);
+                    scope.Transaction.StartTimestamp = DateTimeOffset.Now;
```
bruno-garcia, line 107: Here we shoudl be doing:

@HazAT @rhcarvalho @untitaker @maciejwalkowiak does the pseudo code below make sense?


```
var transactionName = GetName(); // we get the name somewhere because it's added today to the event.transaction
var transaction = hub.StartTransaction(transactionName, new SamplingContext())

// Ideally we'd add to some data bag of the context, so we can get from it in different parts of the framework like some MVC Filters or something, and are able to create more spans from it, without having to rely on `AsyncLocal`:
context.SomeDataBag["SentryTransaction"] = transaction;

// ALternatively we rely on AsyncLocal, or both:
Hub.ConfigureScope(s => s.SetTransaction(transaction));

try {
var middlewareSpan = transaction.startSpan("middleware") (or startChild or something);
await middleware();
middlewareSpan.end();
transaction.status(context.Response.Status.ToSentryStatus());
} catch {
transaction.status(exception);
}
finally {
transaction.finish()
}

```

src/Sentry.AspNetCore/SentryMiddleware.cs
```diff
@@ -122,6 +124,13 @@ public async Task InvokeAsync(HttpContext context)
 
                     ExceptionDispatchInfo.Capture(e).Throw();
                 }
+                finally
+                {
+                    hub.ConfigureScope(scope =>
```
bruno-garcia, line 129: I wouldn't rely on the thing being in the scope.

We can hold the reference to the transaction in the body of this method, to make sure we're finishign the same instance we started here.

bruno-garcia (commented):
Another quick pass..

src/Sentry.AspNetCore/SentryMiddleware.cs
maciejwalkowiak: `var transaction = hub.StartTransaction(transactionName, new SamplingContext())` should rather take `CustomSamplingContext` instead of `SamplingContext`.

`Hub.ConfigureScope(s => s.SetTransaction(transaction));` in Java once you start transaction it's set on the scoep automatically.

Other than that it does make sense.

rhcarvalho: If it can serve as inspiration, here's the equivalent middleware in Go: https://github.com/getsentry/sentry-go/blob/cca482cffbf600853b2f3abda730f9d5a97ee9aa/http/sentryhttp.go#L80-L101

1. The transaction name is stored in the `Scope` (as there was already a slot for it there, and Tracing support hijacks that event field)
2. The "some data bag" idea is represented by Go's `Context` type. Spans are stored there. There's no transaction type per-se. A transaction is a wire representation of a span tree (in fact, just a special type of Event).
3. ⚠️ Bruno your code creates both a transaction AND a child span. But if you represent the transaction as a thing on its own, it is already a span. So in your example there's one too many nesting levels. Either the middleware starts a transaction or a child span on an existing transaction. I don't like code needing to check whether it needs to call `StartTransaction` or `StartSpan`, that's why, among other reasons, there's only `StartSpan` in Go.
4. Apart from setting the span status, the Python SDK also sets a tag with the HTTP status. We should do that too.
[integrations/aiohttp.py#L112](https://github.com/getsentry/sentry-python/blob/ab3da0809d6c2c32adfa63917af03a58cd498fd3/sentry_sdk/integrations/aiohttp.py#L112)
[tracing.py#L319-L350](https://github.com/getsentry/sentry-python/blob/f6f3525f8812f60911573a8b7f71807ecf2e1052/sentry_sdk/tracing.py#L319-L350)
5. When starting a transaction/span, we pass data that modify the transaction/span (the so called "TransactionContext" and "SpanContext" in JS). The "SamplingContext" is something else, it originates in the SDK and is passed to a TracesSampler.

src/Sentry.AspNetCore/SentryMiddleware.cs
rhcarvalho: I agree. If the middleware starts a span/transaction, it should have a direct reference to it and use it to call finish when the other HTTP handlers it wraps terminate.
I'd expect something along the lines of what Bruno wrote in [his comment](https://github.com/getsentry/sentry-dotnet/pull/633/files/c99635da62183de25b1e29c0e82f8802181a41cf#r533866281).

src/Sentry.AspNetCore/SpanStatusMapper.cs
```diff
@@ -0,0 +1,23 @@
...
+            403 => SpanStatus.PermissionDenied,
+            404 => SpanStatus.NotFound,
+            409 => SpanStatus.AlreadyExists,
+            429 => SpanStatus.ResourceExhausted,
+            499 => SpanStatus.Cancelled,
+            500 => SpanStatus.InternalError,
+            501 => SpanStatus.Unimplemented,
+            503 => SpanStatus.Unavailable,
+            504 => SpanStatus.DeadlineExceeded,
+            _   => SpanStatus.UnknownError
```
rhcarvalho, line 20: Does this intentionally ignore the non-error case? 200 => Ok.

Compare with [tracing.py#L319-L350](https://github.com/getsentry/sentry-python/blob/f6f3525f8812f60911573a8b7f71807ecf2e1052/sentry_sdk/tracing.py#L319-L350) for reference.

src/Sentry/Protocol/ISpan.cs
```diff
@@ -0,0 +1,34 @@
...
+
+        SentryId? ParentSpanId { get; }
+
+        SentryId TraceId { get; set; }
+
+        DateTimeOffset StartTimestamp { get; set; }
+
+        DateTimeOffset EndTimestamp { get; set; }
+
+        string? Operation { get; set; }
```
rhcarvalho, line 18: While empty operation "works", SDKs would be more useful if they make it obvious that this field must be set. A span without "op" is defaulted during ingestion (in Relay); here's an example:

image

What I have done in Go is take an explicit `op` parameter in `StartSpan`.

src/Sentry.AspNetCore/SentryMiddleware.cs
rhcarvalho: One point to add:

6. The `op` for HTTP server integrations should be `"http.server"`, so we all follow a consistent naming pattern -- unless there's good reason to deviate from that.

maciejwalkowiak: > The op for HTTP server integrations should be "http.server", so we all follow a consistent naming pattern -- unless there's good reason to deviate from that.

Is this documented somewhere?

rhcarvalho: Possibly not :/
It started off prior industry specs and we use it for Python as in [integrations/aiohttp.py#L92](https://github.com/getsentry/sentry-python/blob/ab3da0809d6c2c32adfa63917af03a58cd498fd3/sentry_sdk/integrations/aiohttp.py#L92)

Tyrrrz:
> Today the `Transaction` field is used by the error product. In Java, it's still there for the Event:
> 
> https://github.com/getsentry/sentry-java/blob/3ed2ff105dd246cbf0242e802543c710bcb14a0b/sentry/src/main/java/io/sentry/SentryEvent.java#L89
> 
> As I said above, there are two classes: `SentryEvent` and `Transaction`. From Sentry's point of view, both are Events.

I see. But it's removed on the Scope though?

src/Sentry/Protocol/ISpan.cs
Tyrrrz: @rhcarvalho is the `Name` of Transaction also required?

rhcarvalho: Yes, here in this example the name is `GET /random.png`

image

image


For automatic instrumentation provided by the SDK for HTTP servers, the transaction name should be `METHOD /path/pattern` and the transaction `op` should be `"http.server"`

Tyrrrz: Awesome, thanks!

lucas-zimerman: Interesting, I always had in mind that the transaction name was the URL and op the HTTP type (get/post,...)

Tyrrrz:
@bruno-garcia 

> ... From Sentry's point of view, both are Events.

I'm confused what do you mean by this, because it seems that Sentry docs discriminate between Events and Transactions as different envelope items:

image

image

src/Sentry/Protocol/ISpan.cs
rhcarvalho: Sentry groups transactions by name, so the pattern `METHOD /path/pattern` normally gives you the right split.

We have/had some SDK releases that shipped with only `/path/pattern`, and we're moving away from that because either you have different handlers for different HTTP verbs, or simply a `GET /foo` request will have different performance characteristics than `POST /foo`.

In any case, SDKs should have a mechanism for users to provide a custom transaction name. A common use case is being able to translate things like `GET /users/123/token` into a more generic pattern `GET /users/{}/token`, if the SDK cannot do it by itself.

lucas-zimerman:
> @bruno-garcia
> 
> > ... From Sentry's point of view, both are Events.
> 
> I'm confused what do you mean by this, because it seems that Sentry docs discriminate between Events and Transactions as different envelope items:
> 
> image
> 
> image

lucas-zimerman:
oof sorry D:

Tyrrrz:
Another question: what is Sentry trace header for?

rhcarvalho:
> @bruno-garcia
> 
> > ... From Sentry's point of view, both are Events.
> 
> I'm confused what do you mean by this, because it seems that Sentry docs discriminate between Events and Transactions as different envelope items.

Let me try to help Bruno and clarify.

In memory, spans build-up a conceptual tree of timed operations. We call the whole span tree a transaction. So one can think of the transaction as either the whole tree or the root span of the tree (in inheritance terms, `Transaction` "is a (subclass of)" `Span`).

Over the wire, transactions are serialized to JSON as an augmented `Event`, and sent as envelopes. The different envelope types are for optimizing ingestion (so we can route "transaction events" differently than other events, mostly "error events").

In the Sentry UI, you can use Discover to look at all events regardless of type, and the Issues and Performance sections to dive into errors and transactions, respectively.

I hope that helps :)

rhcarvalho:
> Another question: what is Sentry trace header for?

The header is for trace propagation. https://develop.sentry.dev/sdk/unified-api/tracing#header-sentry-trace

Long story short, the SDK uses the header to continue traces from upstream services (incoming HTTP requests), and to propagate tracing information to downstream services (outgoing HTTP requests).

Tyrrrz:
> > @bruno-garcia
> > > ... From Sentry's point of view, both are Events.
> > 
> > 
> > I'm confused what do you mean by this, because it seems that Sentry docs discriminate between Events and Transactions as different envelope items.
> 
> Let me try to help Bruno and clarify.
> 
> In memory, spans build-up a conceptual tree of timed operations. We call the whole span tree a transaction. So one can think of the transaction as either the whole tree or the root span of the tree (in inheritance terms, `Transaction` "is a (subclass of)" `Span`).
> 
> Over the wire, transactions are serialized to JSON as an augmented `Event`, and sent as envelopes. The different envelope types are for optimizing ingestion (so we can route "transaction events" differently than other events, mostly "error events").
> 
> In the Sentry UI, you can use Discover to look at all events regardless of type, and the Issues and Performance sections to dive into errors and transactions, respectively.
> 
> I hope that helps :)

I see, yes that clarified a lot, thanks.

Do you differentiate between Event and Transaction at the SDK level (in the Go implementation) or do you just attach a Span to the current scope and then decide whether to encode it as event or transaction based on whether the Span is set?

Tyrrrz:
> > Another question: what is Sentry trace header for?
> 
> The header is for trace propagation. https://develop.sentry.dev/sdk/unified-api/tracing#header-sentry-trace
> 
> Long story short, the SDK uses the header to continue traces from upstream services (incoming HTTP requests), and to propagate tracing information to downstream services (outgoing HTTP requests).

Thanks, that makes sense!

rhcarvalho:
Good questions, thanks @Tyrrrz 

> Do you differentiate between Event and Transaction at the SDK level (in the Go implementation) or do you just attach a Span to the current scope and then decide whether to encode it as event or transaction based on whether the Span is set?

The choice is not between event or transaction, but event or nothing -- we don't send spans on their own.

```
T -
    \
    S1
     |- S11
    S2
```

Calling `Finish` on any span `S*` sets their end time. Calling `Finish` on the transaction / root span `T` sets its end time AND sends a "transaction event" (and event with `Event.Type = "transaction"` and other fields filled in) to Sentry.

Example in Go of how a root span / transaction gets converted into an `Event`: https://github.com/getsentry/sentry-go/blob/111eae57bd5a5953ed4df47d7ccb7d0d8cf8b3d4/tracing.go#L311-L338

---

Disclaimer: there is no concept of inheritance in Go (nor classes), so it is implemented different than JS and Python, for example.

Disclaimer 2: personally, I find it harder to write generic middleware / instrumentation code if it needs to decide whether to start a transaction or a span (depending on the pre-existence of a parent / span tree somewhere).

For the above reasons, in Go, there is only `Span` and `StartSpan`. Starting a span in an empty context makes it a "transaction" when it comes the time to serialize the tree as an `Event` (when the `Span.Finish` method is called). There is no `Transaction` type.

`Span.Finish` of the root span converts the span tree into an `Event` (as I linked above) and calls the existing `CaptureEvent` method to send the transaction event to Sentry. The transports deal with the envelope format and figuring out the right endpoint before sending out to Sentry based on `Event.Type`.

While the span is ongoing, it is stored in a Go-specific `Context` type, which plays a role similar to `Scope`, and was a more idiomatic choice. Unless there's something similarly idiomatic for .NET, the expectation is that the current span stays in the `Scope`.

kanadaj:
That might actually be a direction worth going for. The current state of Python instrumentation makes it so tracing code errors out if there isn't a transaction already, which is a bit of a pain when sharing code between a web end point and CLI, and that's probably something we should avoid in .NET

untitaker:
>tracing code errors out if there isn't a transaction already

spans are discarded when a transaction is not open, which is not exactly erroring out

> Starting a span in an empty context makes it a "transaction" when it comes the time to serialize the tree as an Event (when the Span.Finish method is called). There is no Transaction type.

This has the disadvantage that every db span will suddenly become its own transaction if there's no wrapping transaction for the endpoint. This is billing relevant.

bruno-garcia:
> > tracing code errors out if there isn't a transaction already
> 
> spans are discarded when a transaction is not open, which is not exactly erroring out
> 

I'd like to stick to this behavior (which was used in Java too). 
I understand @rhcarvalho's concern about having to diferentiate a startSpan and startTransaction but in a middleware, of our auto instrumentation integration, we can define that: Here we start a new transaction (not a span). If there's a transaction already open, we can then just add a span but that's a fallback scenario, unlikely to happen. I'm thinking here when do you have something that creates a Sentry transaction in a ASP.NET Core requests that runs before we hadnle the incoming HTTP request. i can't think of a legit use case for that. Even not doing anything if a transaction exists is OK (don't create a span, don't acknowledge that use case)


> > Starting a span in an empty context makes it a "transaction" when it comes the time to serialize the tree as an Event (when the Span.Finish method is called). There is no Transaction type.
> 
> This has the disadvantage that every db span will suddenly become its own transaction if there's no wrapping transaction for the endpoint. This is billing relevant.

Goes wiht the first comment, we should be explicit here if a transaction is being started or it's a span only.

src/Sentry/Protocol/Transaction.cs
```diff
@@ -56,6 +56,8 @@ public void Finish(SpanStatus status = SpanStatus.Ok)
         {
             EndTimestamp = DateTimeOffset.Now;
             Status = status;
+
+            SentrySdk.CaptureTransaction(this);
```
bruno-garcia, line 60: we wont' be able to unit test this unless we take `IHub` here.

Hub could pass `this` when creating the instance

Tyrrrz: Yeah, this is a hacky implementation.
Overall, I really dislike sending stuff from `Transaction.Finish`

bruno-garcia: Same here. At firs tin Java we didn't do that but rather you should call Hub.capture.

Tyrrrz:
@bruno-garcia regarding Transaction inheriting most of Event's fields:

- Does transaction also have exception? If yes, how does it work/ how it gets propagated?
- I imagine fields that are on scope get copied onto the transaction, but what about fields that are on event but not on scope? For example, on event, stuff like `event.Release` etc are set from event processors, but I assume they won't be called on transactions? or will they?

src/Sentry/Protocol/Envelopes/EnvelopeItem.cs
```diff
@@ -176,6 +177,20 @@ public static EnvelopeItem FromUserFeedback(UserFeedback sentryUserFeedback)
...
 
+        /// <summary>
+        /// Creates an envelope item from transaction.
+        /// </summary>
+        public static EnvelopeItem FromTransaction(Transaction transaction)
+        {
+            var header = new Dictionary<string, object?>(StringComparer.Ordinal)
+            {
+                // Transaction is an event
+                [TypeKey] = TypeValueEvent
```
bruno-garcia, line 188: As we figured out today this will need to be `transaction`.

src/Sentry/Protocol/Span.cs
```diff
@@ -0,0 +1,118 @@
...
+        public string Operation { get; }
+        public string? Description { get; set; }
+        public SpanStatus? Status { get; set; }
+        public bool IsSampled { get; set; }
+
+        private ConcurrentDictionary<string, string>? _tags;
+        public IReadOnlyDictionary<string, string> Tags => _tags ??= new ConcurrentDictionary<string, string>();
+
+        private ConcurrentDictionary<string, object>? _data;
+        public IReadOnlyDictionary<string, object> Data => _data ??= new ConcurrentDictionary<string, object>();
```
bruno-garcia, line 27: Maybe Lazy<T> with `PublicationOnly`

src/Sentry/Protocol/Span.cs
```diff
@@ -0,0 +1,118 @@
...
+
+        internal Span(SentryId? spanId = null, SentryId? parentSpanId = null, string operation = "unknown")
+        {
+            SpanId = spanId ?? SentryId.Create();
+            ParentSpanId = parentSpanId;
+            TraceId = SentryId.Create();
+            StartTimestamp = EndTimestamp = DateTimeOffset.Now;
```
bruno-garcia, line 34: Do we run the risk of sending a 0 duration span when in fact we should have sent a open span (no end time?)

What's the reason to initialize the value like this?

src/Sentry/Protocol/Span.cs
```diff
@@ -0,0 +1,118 @@
...
+
+        public void Finish(SpanStatus status = SpanStatus.Ok)
+        {
+            EndTimestamp = DateTimeOffset.Now;
+            Status = status;
+        }
+
+        public void WriteTo(Utf8JsonWriter writer)
+        {
+            writer.WriteStartObject();
```
bruno-garcia, line 48: Odd that this is a mutable object that will be accessed concurrently and we're reading its state here to serialize. We need to make sure that all members we're accessing can be accessed concurrently.

src/Sentry/Protocol/Span.cs
```diff
@@ -0,0 +1,118 @@
...
+            if (ParentSpanId is {} parentSpanId)
+            {
+                writer.WriteSerializable("parent_span_id", parentSpanId);
+            }
+
+            writer.WriteSerializable("trace_id", TraceId);
+            writer.WriteString("start_timestamp", StartTimestamp);
+            writer.WriteString("timestamp", EndTimestamp);
+
+            if (!string.IsNullOrWhiteSpace(Operation))
```
bruno-garcia, line 61: This is mutable so it's racy. It could be non empty here but suddenly empty on the next line

src/Sentry/Protocol/Span.cs
```diff
@@ -0,0 +1,118 @@
...
+using System.Linq;
+using System.Text.Json;
+using Sentry.Internal.Extensions;
+
+namespace Sentry.Protocol
+{
+    // https://develop.sentry.dev/sdk/event-payloads/span
+    public class Span : ISpan, IJsonSerializable
+    {
+        public SentryId SpanId { get; }
```
bruno-garcia, line 13: The protocol really has all of these members being mutable?

src/Sentry/Protocol/Span.cs
```diff
@@ -0,0 +1,118 @@
...
+        public SentryId? ParentSpanId { get; }
+        public SentryId TraceId { get; set; }
+        public DateTimeOffset StartTimestamp { get; set; }
```
bruno-garcia, line 16: I can't see why this has a setter, could you please elaborate?

src/Sentry/Protocol/SpanStatus.cs
```diff
@@ -0,0 +1,78 @@
...
+
+namespace Sentry.Protocol
+{
+    /// <summary>
+    /// Span status.
+    /// </summary>
+    public enum SpanStatus
+    {
+        /// <summary>The operation completed successfully.</summary>
+        [EnumMember(Value = "ok")]
```
bruno-garcia, line 11: Oh I thought you killed these `System.Runtime.Serialization` attributes

src/Sentry/Protocol/Transaction.cs
```diff
@@ -0,0 +1,179 @@
...
+                null,
+                null,
+                operation)
+        {
+        }
+
+        public ISpan StartChild(string operation)
+        {
+            var span = new Span(null, SpanId, operation);
+            (_children ??= new List<Span>()).Add(span);
```
bruno-garcia, line 64: This needs to be synchronized, if we're allowing the transaction instance to be passed to multiple threads where we'll be calling `StartChild`

lucas-zimerman: storing Spans on AsyncLocal works great for this case.

src/Sentry.AspNetCore/SentryMiddleware.cs
```diff
@@ -2,13 +2,15 @@
 using System.Diagnostics;
 using System.Runtime.ExceptionServices;
 using System.Threading.Tasks;
+using System.Transactions;
```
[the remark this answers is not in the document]
Tyrrrz: Yeah, looks like i accidentally imported the wrong transaction at some point

src/Sentry/Protocol/Span.cs
Tyrrrz: do you think the span should be frozen after finishing? or when serializing?

src/Sentry/Protocol/SpanStatus.cs
Tyrrrz: I decided to leave these and later make it so they're picked up by the new serialization

Tyrrrz: Currently they are ignored

src/Sentry/Protocol/Transaction.cs
Tyrrrz: > This needs to be synchronized...

Do you mean synchronizing the list creation specifically, or turning the list into a `ConcurrentBag`?

src/Sentry/Protocol/Span.cs
Tyrrrz: I didn't know spans can have no end time

src/Sentry/Protocol/Span.cs
Tyrrrz: Then we can't do deserialization

src/Sentry/Protocol/Span.cs
Tyrrrz: The protocol doesn't say anything about mutability. It doesn't even cover some of the fields at all.

src/Sentry/Protocol/Span.cs
Tyrrrz: Removed

src/Sentry/Protocol/Span.cs
Tyrrrz: Do you have a suggestion what we should do here? We could try "freezing" the span once it's finished, but it will be hard to guarantee that nothing gets written to the dictionaries.

src/Sentry.AspNetCore/SentryMiddleware.cs
```diff
@@ -94,6 +97,19 @@ public async Task InvokeAsync(HttpContext context)
...
                 }
 
+                var routeData = context.GetRouteData();
+                var controller = routeData.Values["controller"]?.ToString();
+                var action = routeData.Values["action"]?.ToString();
+                var area = routeData.Values["area"]?.ToString();
+
+                // TODO: What if it's not using controllers (i.e. endpoints)?
+
+                var transactionName = area == null
```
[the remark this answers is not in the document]
Tyrrrz: Updated so that it gets set

aleksihakli:
Hey! Just wanted to pop in and say that this PR and performance monitoring for .NET Core would be a very nice thing to have upstream. Is there anything you need help with in upstreaming this and getting a new release with the performance addon included?

bruno-garcia:
@aleksihakli thanks for chiming in. We're planning to get this merged and make a release soon.

Would you be willing to try a preview? We'll release a 3.0.0-alpha.X with this code as soon as @Tyrrrz merges this

src/Sentry/Protocol/Transaction.cs
bruno-garcia: two concurrent calls to `StartChild` would yield multiple `List<Span>`'s created, and due to CPU cache and other optimizations it's not guaranteed that the instance set first would be visible to all threads. This particular case could be solved with `volatile` but that wouldn't solve everything:

We'd have a non thread-safe list in the field, which would throw as we read spans in the scope and concurrently try to add to it (call StartSpan) again.

Without much looking into it the simplest way I can think is to indeed use `ConcurrentBag` (if order doesn't matter), but have that initialized in the constructor (or with a `Lazy<T>` with `PublicationOnly`). 

If you prefer we can hop on a call and brainstorm this together. I'm not familiar with the performance API to the level I can just drop a comment here and be confident we're addressing all the issues. /cc @marandaneto FYI

src/Sentry/Protocol/Span.cs
bruno-garcia: We should address that. Have a discussion and write it on the docs.
Ideally things would be immutable unless they need not to be.

src/Sentry/Protocol/Span.cs
bruno-garcia: It seems natural to me that once it's finished it should be frozen.

Curious what other SDKs did here.

src/Sentry/Protocol/Span.cs
bruno-garcia: Sorry, I don't follow, what do you mean?

The current approach is racy though. As concurrent access to this will yield multiple threads creating their own collection and assigning to the field, which is not immediately visible to all threads because the field isn't `volatile`. Perhaps a double check lock or `Interlocked.Exchange` could solve the problem here but I think `Lazy<T> with `PublicationOnly` is the simplest.

The simplest is to just have the instance from the start of the object but since this field is likely not going to be used often, paying the price to allocating it on each span, which will have dozens of instances per transaction, we're better off just using lazy or other mechanism

src/Sentry.AspNetCore/ScopeExtensions.cs
```diff
@@ -87,7 +87,9 @@ public static void Populate(this Scope scope, HttpContext context, SentryAspNetC
                         scope.SetTag("route.area", area);
                     }
 
-                    scope.Transaction = area == null ? $"{controller}.{action}" : $"{area}.{controller}.{action}";
+                    scope.TransactionName = area == null
```
bruno-garcia, line 90: Can we have this in some internal static method to make sure changing its stays aligned with the one on the middleware?

src/Sentry/Protocol/Envelopes/EnvelopeItem.cs
```diff
@@ -176,6 +177,19 @@ public static EnvelopeItem FromUserFeedback(UserFeedback sentryUserFeedback)
             return new EnvelopeItem(header, new JsonSerializable(sentryUserFeedback));
         }
 
+        /// <summary>
+        /// Creates an envelope item from transaction.
+        /// </summary>
+        public static EnvelopeItem FromTransaction(Transaction transaction)
+        {
+            var header = new Dictionary<string, object?>(StringComparer.Ordinal)
```
bruno-garcia, line 185: Does this change? Can we get away with a single instance that we pass the reference of to all envelope items?

src/Sentry/Protocol/IScope.cs
```diff
@@ -69,6 +58,17 @@ public interface IScope
...
+        /// <summary>
+        /// The name of the transaction in which there was an event.
+        /// </summary>
+        /// <remarks>
+        /// A transaction should only be defined when it can be well defined.
+        /// On a Web framework, for example, a transaction is the route template
+        /// rather than the actual request path. That is so GET /user/10 and /user/20
+        /// (which have route template /user/{id}) are identified as the same transaction.
+        /// </remarks>
+        string? TransactionName { get; set; }
```
bruno-garcia, line 70: Probably with a changelog item just for this change since it's a breaking one:
* Transaction on Scope and Event now is called TransactionName

src/Sentry/Protocol/SentryId.cs
```diff
@@ -29,20 +29,39 @@ namespace Sentry
...
+        public override string ToString() => _guid.ToString("n");
+
+        // Note: spans are sentry IDs with only 16 characters, rest being truncated.
+        // This is obviously a bad idea as it invalidates GUID's uniqueness properties
+        // (https://devblogs.microsoft.com/oldnewthing/20080627-00/?p=21823)
+        // but all other SDKs do it this way, so we have no choice but to comply.
+        /// <summary>
+        /// Returns a truncated ID.
+        /// </summary>
+        public string ToShortString() => ToString().Substring(0, 16);
```
bruno-garcia, line 41: Lets please discuss this before taking this approach.

Why do we need to do this, because other SDKs did? That's not a reason.

/cc @marandaneto

src/Sentry/Protocol/Span.cs
```diff
@@ -0,0 +1,152 @@
...
+
+        private ConcurrentDictionary<string, string>? _tags;
+
+        /// <inheritdoc />
+        public IReadOnlyDictionary<string, string> Tags => _tags ??= new ConcurrentDictionary<string, string>();
+
+        private ConcurrentDictionary<string, object?>? _data;
+
+        /// <inheritdoc />
+        public IReadOnlyDictionary<string, object?> Extra => _data ??= new ConcurrentDictionary<string, object?>();
```
bruno-garcia, line 51: Does a span really have to be thread-safe? Maybe only the transaction needs to be which would make our life's easier.

Otherwise this assignment again isn't thread-safe

src/Sentry/Protocol/Transaction.cs
```diff
@@ -0,0 +1,329 @@
...
+            TraceId = SentryId.Create();
+        }
+
+        /// <summary>
+        /// Initializes an instance of <see cref="Transaction"/>.
+        /// </summary>
+        /// <summary>
+        /// Child spans.
+        /// </summary>
+        public IReadOnlyList<Span> Children => _children ??= new List<Span>();
```
bruno-garcia, line 152: I think here we could go with ConcurrentBag and set the instance from the start, since we expect transactions to have spans

src/Sentry/Protocol/Transaction.cs
```diff
@@ -0,0 +1,329 @@
...
+
+        private Dictionary<string, object?>? _extra;
+
+        /// <inheritdoc />
+        public IReadOnlyDictionary<string, object?> Extra => _extra ??= new Dictionary<string, object?>();
+
+        private Dictionary<string, string>? _tags;
+
+        /// <inheritdoc />
+        public IReadOnlyDictionary<string, string> Tags => _tags ??= new Dictionary<string, string>();
```
bruno-garcia, line 124: Not thread-safe.

Same as other discussions, Lazy<T> or assigning Concurrent.X from instantiation, or double check lock, or Interlocked.. lets hop on a call and think this through to choose the right approach?

bruno-garcia (commented):
This is looking really good! We're getting closer.

There are some points about concurrency we should discuss, mainly all properties of Transaction will need to change, and Span (if that's expected to be accessed concurrently)

aleksihakli:
> @aleksihakli thanks for chiming in. We're planning to get this merged and make a release soon.
> 
> Would you be willing to try a preview? We'll release a 3.0.0-alpha.X with this code as soon as @Tyrrrz merges this

Hey Bruno! Sure, I think we have some .NET (Core) projects that I could try the alpha version in and give feedback on :)

src/Sentry/Protocol/Envelopes/EnvelopeItem.cs
Tyrrrz: Technically it should also have `event_id` but it's not currently exposed on `Transaction` and it works without it. Not sure if I should actually try to get that in.

src/Sentry/Protocol/SentryId.cs
Tyrrrz: Can we also not capture the transaction directly on `Transaction.Finish()`? 😄 
That's how other SDKs did it but it introduces a lot of complexity.

src/Sentry/Protocol/Transaction.cs
Tyrrrz: I see that the fields are eagerly-initialized on `Span` (which is accessed concurrently) so maybe we should do this here too
image

bruno-garcia:
> > @aleksihakli thanks for chiming in. We're planning to get this merged and make a release soon.
> > Would you be willing to try a preview? We'll release a 3.0.0-alpha.X with this code as soon as @Tyrrrz merges this
> 
> Hey Bruno! Sure, I think we have some .NET (Core) projects that I could try the alpha version in and give feedback on :) Once the PR is upstreamed and possibly released to package repository I'll try and get it running against the Sentry SaaS version. We also have a recent version of Sentr on-premises running if testing against both distributions is desirable.

That's awesome! Thanks.
One note for Sentry on-premise. In the 3.0.0 of this SDK we moved to the `/envelope` endpoint, so Sentry on-prem must be v20.6 or higher.

But testing only on SaaS is fine. It'sn really about how does it work for you, if you have any issues, and ultimately, does it add value? Does it help you improve the experience for your customers by identifying slowdowns and giving you insight to improve them.

[We're online on Sentry's Discord if you'd like to chat. On #dotnet](https://discord.gg/sentry)

src/Sentry.AspNetCore/ScopeExtensions.cs
```diff
@@ -68,29 +69,26 @@ public static void Populate(this Scope scope, HttpContext context, SentryAspNetC
...
-                        : $"{area}.{controller}.{action}";
+                    scope.SetTag("route.action", action);
                 }
+
+                if (area != null)
+                {
+                    scope.SetTag("route.area", area);
+                }
+
+                scope.TransactionName = context.TryGetRouteTemplate() ?? context.Request.Path;
```
bruno-garcia, line 91: when would we fallback to Path here? Maybe better to leave it out is what I've heard.
If we can't get a proper template, we shoudn't fallback to the whole path

src/Sentry.AspNetCore/SentryMiddleware.cs
```diff
@@ -119,14 +109,18 @@ public async Task InvokeAsync(HttpContext context)
                     scope.OnEvaluating += (_, __) => PopulateScope(context, scope);
                 });
 
-                var transaction = hub.CreateTransaction(transactionName, "http.server");
+                var transaction = hub.CreateTransaction(
+                    // Try to get the route template or fallback to the request path
+                    context.TryGetRouteTemplate() ?? context.Request.Path,
```
bruno-garcia, line 114: @marandaneto @HazAT what do we do when we can't figure out a route template for a transaction name? Don't create the transaction at all?

Falling back to the actual route would be noise I assune?

@Tyrrrz or this fallback means no template was used so it's safe?

src/Sentry.AspNetCore/ScopeExtensions.cs
Tyrrrz: We fallback to path if we can't get the route from endpoint routing or from MVC.
This makes it so that if you have a custom middleware that checks for request path (e.g. `if (context.Request.Path == "/health"`) then it at least shows up properly. I felt that it's more useful.

src/Sentry.AspNetCore/SentryMiddleware.cs
Tyrrrz: @bruno-garcia this checks for two known way to define routes, i.e. the routing middleware and the MVC middleware (legacy). Theoretically, the user can use some other custom routing middleware and then we won't be able to extract the route. But for regular use cases, this would mean that there is no template.

Tyrrrz:
Can't understand what failed windows build.

src/Sentry/Internal/Extensions/StringExtensions.cs
```diff
@@ -0,0 +1,10 @@
+using System.Text.RegularExpressions;
+
+namespace Sentry.Internal.Extensions
+{
+    internal static class StringExtensions
+    {
+        public static string ToSnakeCase(this string str) =>
```
aleksihakli, line 7: The purpose and use cases of snake casing could maybe be documented as a comment as well 🐍

src/Sentry/Protocol/Envelopes/EnvelopeItem.cs
aleksihakli: Maybe that should be documented as a TODO now, if it is not yet included but could be added to the implementation later on?

test/Sentry.Tests/Protocol/TransactionTests.cs
```diff
@@ -0,0 +1,37 @@
...
+        [Fact]
+        public void Serialization_Roundtrip_Success()
+        {
+            // Arrange
+            var transaction = new Transaction(DisabledHub.Instance, null)
+            {
+                Name = "my transaction",
+                Operation = "some op",
+                Description = "description",
+                Request = new Request {Method = "GET", Url = "https://test.com"},
```
aleksihakli, line 20: For examples the top level example domains such as example.com or e.g. sentry.example should always be used:

https://tools.ietf.org/html/rfc2606#page-2

test/Sentry.Tests/Protocol/TransactionTests.cs
```diff
@@ -0,0 +1,37 @@
...
+                User = new User {Email = "foo@bar.com", Username = "john"},
```
aleksihakli, line 21: Similarly, the email should be e.g. test@sentry.example as per RFC 2606

aleksihakli (approved):
C# and .NET aren't my strongest skills but a few notes from here and there. The PR looks clean and good as a whole, good job!

Tyrrrz:
Thanks @aleksihakli

src/Sentry/Extensibility/DisabledHub.cs
```diff
@@ -51,6 +52,17 @@ public void WithScope(Action<Scope> scopeCallback)
         {
         }
 
+        /// <summary>
+        /// Returns a dummy transaction.
+        /// </summary>
+        public Transaction CreateTransaction(string name, string operation) =>
+            new Transaction(this, null);
```
bruno-garcia, line 59: We should have a singleton NoOp instance here instead.
Having a disabled hub means we should reduce the instrumentations overhead

src/Sentry/Internal/Hub.cs
```diff
@@ -99,6 +101,39 @@ public void WithScope(Action<Scope> scopeCallback)
...
+                trans.Sdk.Name = Constants.SdkName;
+                trans.Sdk.Version = nameAndVersion.Version;
+            }
+
+            if (nameAndVersion.Version != null)
+            {
+                trans.Sdk.AddPackage(protocolPackageName, nameAndVersion.Version);
+            }
+
+            ConfigureScope(scope => scope.Transaction = trans);
```
bruno-garcia, line 126: I believe by default we won't do this. The snippets I saw from JS require the caller to decide whether the transaction should be bound to the scope or not. 

For now this looks good and lets try like this.

src/Sentry/Internal/MainSentryEventProcessor.cs
```diff
@@ -96,20 +96,17 @@ public SentryEvent Process(SentryEvent @event)
 
             @event.Platform = Protocol.Constants.Platform;
 
-            if (@event.Sdk != null)
+            // SDK Name/Version might have be already set by an outer package
+            // e.g: ASP.NET Core can set itself as the SDK
+            if (@event.Sdk.Version == null && @event.Sdk.Name == null)
```
bruno-garcia, line 101: Are transactions going through this at the end? Doesn't seem like it since this take SentryEvent.

So transactions will always be reported as `sentry.dotnet`?

src/Sentry/Protocol/Context/Trace.cs
```diff
@@ -0,0 +1,85 @@
...
+
+            writer.WriteEndObject();
+        }
+
+        /// <summary>
+        /// Parses trace context from JSON.
+        /// </summary>
+        public static Trace FromJson(JsonElement json)
+        {
+            var spanId = json.GetPropertyOrNull("span_id")?.Pipe(SpanId.FromJson) ?? SpanId.Empty;
```
bruno-garcia, line 69: Invalid json blobs will result in broken traces being sent (empty string ids etc)

I think we should be returning nullable and dealing with it.

src/Sentry/Protocol/Envelopes/Envelope.cs
```diff
@@ -106,6 +106,24 @@ public static Envelope FromUserFeedback(UserFeedback sentryUserFeedback)
...
         }
 
+        /// <summary>
+        /// Creates an envelope that contains a single transaction.
+        /// </summary>
+        public static Envelope FromTransaction(Transaction transaction)
+        {
+            var header = new Dictionary<string, object?>(StringComparer.Ordinal)
+            {
+                // TODO: this should include transaction's event id
```
bruno-garcia, line 116: 👍

src/Sentry/Protocol/Envelopes/EnvelopeItem.cs
bruno-garcia: True. Also if it's not needed then maybe we can get away with it. @jan-auer do we need the `event_id` in the envelope header afterall? it's working without it and we could avoid allocating a new map just to set the id (which is anyway on the payload)

src/Sentry/Protocol/SentryEvent.cs
```diff
@@ -169,6 +169,13 @@ public IEnumerable<string> Fingerprint
         /// <inheritdoc />
         public IReadOnlyDictionary<string, string> Tags => _tags ??= new Dictionary<string, string>();
 
+        // TODO: this is a workaround, ideally Event should not inherit from IScope
```
bruno-garcia, line 172: 👍

src/Sentry/Protocol/SpanId.cs
```diff
@@ -0,0 +1,84 @@
...
+        public override string ToString() => _value;
+
+        // Note: spans are sentry IDs with only 16 characters, rest being truncated.
+        // This is obviously a bad idea as it invalidates GUID's uniqueness properties
+        // (https://devblogs.microsoft.com/oldnewthing/20080627-00/?p=21823)
+        // but all other SDKs do it this way, so we have no choice but to comply.
+        /// <summary>
+        /// Generates a new Sentry ID.
+        /// </summary>
+        public static SpanId Create() => new(Guid.NewGuid().ToString("n").Substring(0, 16));
```
bruno-garcia, line 42: 🙄 

Lets get back to this then

src/Sentry/Protocol/SpanRecorder.cs
```diff
@@ -0,0 +1,40 @@
...
+        }
+
+        /// <summary>
+        /// Gets all recorded spans.
+        /// </summary>
+        public IReadOnlyList<Span> GetAll()
+        {
+            lock (_lock)
+            {
+                return _spans;
```
bruno-garcia, line 36: There's exclusive access to just return the reference, which now can be access outside of any locks by the caller which can iterate over it while another thread is adding to it, which would throw an exception.

We need a new list, with the current items:

```suggestion
                return new List<Span>(_spans);
```

src/Sentry/Protocol/SpanRecorder.cs
```diff
@@ -0,0 +1,40 @@
...
+
+        private readonly object _lock = new object();
+        private readonly List<Span> _spans = new List<Span>();
+
+        /// <summary>
+        /// Records a span.
+        /// </summary>
+        public void Add(Span span)
+        {
+            lock (_lock)
```
bruno-garcia, line 20: Taking a lock to add any and all Spans can possibly create contention if we have a bunch of spans getting added to the same transaction.

Lets revisit this later to add a capped list that won't require exclusive access on all `Add`

Maybe just a `Interlocked.Increment` so we keep track of the count of items in the `_span` and using `ConcurrentBag` would be an improvement. But we discuss/measure things later, this will do for now.

src/Sentry/Protocol/Transaction.cs
```diff
@@ -0,0 +1,327 @@
...
+
+        /// <inheritdoc />
+        public bool IsSampled
+        {
+            get => Contexts.Trace.IsSampled;
+            set => Contexts.Trace.IsSampled = value;
+        }
+
+        /// <inheritdoc />
+        public SentryLevel? Level { get; set; }
```
bruno-garcia, line 72: This probably doesn't make sense, right? What does "level" mean on a Transaction.

src/Sentry/SentryClient.cs
```diff
@@ -108,19 +97,53 @@ public void CaptureUserFeedback(UserFeedback userFeedback)
...
+            {
+                _options.DiagnosticLogger?.LogWarning("Transaction discarded due to one or more required fields missing.");
+                return;
+            }
+
+            if (_options.TraceSampleRate < 1)
+            {
+                if (Random.NextDouble() > _options.TraceSampleRate)
+                {
+                    _options.DiagnosticLogger?.LogDebug("Transaction sampled.");
```
bruno-garcia, line 141:
```suggestion
                    _options.DiagnosticLogger?.LogDebug("Transaction dropped due to random sampling.");
```

test/Sentry.Tests/SentryClientTests.cs
```diff
@@ -365,6 +366,48 @@ public void CaptureUserFeedback_DisposedClient_ThrowsObjectDisposedException()
...
+                Name = "test name",
+                Operation = "test operation"
+            });
+
+            // Assert
+            _ = sut.Worker.Received(1).EnqueueEnvelope(Arg.Any<Envelope>());
+        }
+
+        [Fact]
+        public void CaptureTransaction_InvalidTransaction_Ignored()
```
bruno-garcia, line 387: Why is it ignored?

Maybe: `CaptureTransaction_NoName_Ignored`

`CaptureTransaction_NoOperation_Ignored`

and one case per each?

bruno-garcia (approved):
There are a few things we can take on a new PR, but this is good to merge and put out on a new preview!

:shipit:

bruno-garcia:
@aleksihakli thanks for helping out with reviewing!

aleksihakli:
Took a look and saw that 3.0.0-alpha.7 is now available in package repositories.

I'm not sure if I have the time to test drive this before the Christmas holidays, but we'll most probably have the chance to start testing the performance side of things early next year in different environments.

Super excited to get this integrated upstream, thanks for the effort folks! Ever since OpBeat was acquired and merged into Elastic APM I've been looking for other viable performance tracking options and think that Sentry is an increasingly exciting option for projects of all size and has a good model with the FOSS and SaaS deployments available. Keep up the great work.

src/Sentry/Internal/MainSentryEventProcessor.cs
Tyrrrz: Yeah... The SDK is set separately when creating the transaction right now. Transactions also don't get all the other default metadata either, which is populated by event processors.

src/Sentry/Protocol/Transaction.cs
Tyrrrz: Yeah it might be one of those fields that have no place on transaction. It would be really nice to have some "official" list.

src/Sentry/Protocol/Span.cs
rhcarvalho: (Late to the party, commenting in case it is still useful -- cc @bruno-garcia @Tyrrrz )

The current behavior for "open spans" (no end time set) in other SDKs is such that they are filtered out when finishing/sending the transaction.

E.g.:
- https://github.com/getsentry/sentry-python/blob/b7816b0cc100a47082922b8dd3e058134ad75d7c/sentry_sdk/tracing.py#L537-L553
- https://github.com/getsentry/sentry-javascript/blob/a833f0ec0a186f0672a536ea8f7d3b129a71a4fe/packages/tracing/src/transaction.ts#L90-L105

aleksihakli:
Added the SDK to a minimal project and can confirm that the service is indeed sending metrics with 3.0.0-alpha.10 which seems good.

What kind of setups should be tested for a typical .NET Core web project? Entity Framework, API calls, library calls, anything interesting in there?

image

image

bruno-garcia:
Thanks for testing @aleksihakli 

Would you want to discuss this on Discord instead of this PR? It's been merged, I often skip reading messages on merged PR or close issues.

Issue #626, opened 2020-11-28
Upgrade NSubstitute


bruno-garcia:
NSubstitute v4 was released a while ago and dependabot keeps opening a PR.
But there are breaking changes and CI fails.

We need to bump to the latest version and fix the tests.

Tyrrrz:
Do you have a strong preference in NSubstitute? IMO, Moq is better (with its static API) as the code is much easier without extension methods on everything.

Tyrrrz:
https://twitter.com/Tyrrrz/status/1327245894475984896?s=20

bruno-garcia:
I would be happy to discuss this on a greenfield project and maybe go back to Moq (I used it between 2010 and 2015) but I wouldn't add a second library here. 

I'm assuming upgrading won't require a whole rewrite. NSubstitute has it's limitations (and Moq too, like out parameters that I can recall).

SimonCropp:
done by https://github.com/getsentry/sentry-dotnet/pull/634

PR #634, opened 2020-11-30
Tests updated to NSubstitute 4

kanadaj:
All tests now work with NSubstitute 4.2.2 instead of 3.x. 

`LogDebug_DefaultOptions_DoesNotRecordsBreadcrumbs` and `LogTrace_DefaultOptions_DoesNotRecordsBreadcrumbs` have had their `Args.Any<>()` replaced with `DidNotReceiveWithAnyArgs()` since this is a stricter condition anyway and the old version doesn't seem to work regardless of what I've tried.

#skip-changelog

### DIFF
--- a/test/Sentry.Extensions.Logging.Tests/SentryLoggerTests.cs
+++ b/test/Sentry.Extensions.Logging.Tests/SentryLoggerTests.cs
@@ -380,15 +380,6 @@ namespace Sentry.Extensions.Logging.Tests
                     default,
                     default);
 
-            // The below test is the original one, but for some reason NSubstitute 4 refuses to use the Arg.Any() parameters even though they should be correct and therefore throws an error
-            // _fixture.Hub.DidNotReceive()
-            //     .AddBreadcrumb(
-            //         _fixture.Clock,
-            //         Arg.Any<string>(),
-            //         _fixture.CategoryName,
-            //         BreadcrumbType,
-            //         null,
-            //         Arg.Any<BreadcrumbLevel>());
         }
 
         [Fact]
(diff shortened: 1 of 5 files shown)


kanadaj:
Addresses https://github.com/getsentry/sentry-dotnet/issues/626

test/Sentry.Extensions.Logging.Tests/SentryLoggerTests.cs
```diff
@@ -370,14 +370,25 @@ public void LogDebug_DefaultOptions_DoesNotRecordsBreadcrumbs()
...
+
+            // The below test is the original one, but for some reason NSubstitute 4 refuses to use the Arg.Any() parameters even though they should be correct and therefore throws an error
+            // _fixture.Hub.DidNotReceive()
+            //     .AddBreadcrumb(
+            //         _fixture.Clock,
+            //         Arg.Any<string>(),
+            //         _fixture.CategoryName,
+            //         BreadcrumbType,
+            //         null,
+            //         Arg.Any<BreadcrumbLevel>());
```
bruno-garcia, line 391: Makes sense
```suggestion
```

test/Sentry.Extensions.Logging.Tests/SentryLoggerTests.cs
```diff
@@ -387,14 +398,24 @@ public void LogTrace_DefaultOptions_DoesNotRecordsBreadcrumbs()
...
```
bruno-garcia, line 418:
```suggestion
```

bruno-garcia (approved):
Nice, that's all? :) I was under the impressino would be a lot more code.

Please clean up the dead code and we're good to go

Issue #588, opened 2020-11-16
Question: is there much difference/lacking features of the Serilog.Sentry library (vs main Sentry SDK)?


PureKrome:
G'Day 👋🏻 !

With the `Sentry.Serilog` library, is there much features missing .. versus using the main `Sentry` library?

Context: I prefer to pass `ILogger<T>` all around my application. In my `ConfigureServices`, I would prefer to just setup Serilog and then log whatever to wherever, via sinks, like `Sentry.Serilog`. 

So if I get an error, I would then log that to Sentry.

Is it safe to assume that the data which is logged to sentry via the Serilog sink is pretty much feature rich as manually using the Serilog library? Like, rich exception information, etc?

Or in other words, if I use the `Sentry.Serilog` library, I would be missing out of the following pieces of data:
- `<something>`
- `<another thing>`
- etc.

bruno-garcia:
For the main `Sentry` package, you're not missing anything.

Except that you can use the Sentry API to add more data to the event. Like `Contexts` or `tags` which get indexed.

The protocol has well defined stuff: https://develop.sentry.dev/sdk/event-payloads/
C# API Docs: https://getsentry.github.io/sentry-dotnet/api/Sentry.Protocol.html

But the integrations such as `UnhandledException` and `UnobservedTaskException` are hooked up automatically so even if you're using `Sentry.Serilog`, such errors would be captured.

For more context, check out the [Sentry Serilog Sink](https://github.com/getsentry/sentry-dotnet/blob/fdf24546995fd0a1e9c009668647c8648af9e3ec/src/Sentry.Serilog/SentrySink.cs#L58-L136), and you'll see what we **do use**. And if you check out the API docs, and the [Sentry Docs](https://docs.sentry.io/platforms/dotnet/enriching-events/context/) you'll what possible through the SDK itself (not limited by a logging integration).

If you're on a ASP.NET Core app, and don't install the `Sentry.AspNetCore` you'll be missing:

* Request information like headers
* User agent is used by Sentry to show browser, OS info
* Errors handled by the error page
* Exceptions thrown in the middleware pipeline
* User authenticated in ASP.NET Core is 
* .. other that I can't think of right now

Similar data for ASP.NET classic (which btw, is moving to a `Sentry.AspNet` dedicated package on Sentry SDK 3.0.0.

PureKrome:
Thanks @bruno-garcia for the help/reply.

So if i understand you right ...

[...]

Issue #257, opened 2019-08-07
How to use Serilog + Sentry + AspNetCore ?


PureKrome:
Hi 👋 
I'm trying to understand the steps involved to leverging Sentry.IO with an ASP.NET Core 2.2 website BUT using Serilog as the logging mechanism.

Currently, we're using Serilog to log Info+ messages to <some place>: Console or <some website>. Works great.

But we'd also like to use Sentry.io as a sink for our manual errors or (unexpected) exceptions that could get thrown.

What is the trick to setting this up please?

(for bonus points, we define our serilog settings in our `appsettings.xxxx.config` file(s) so we don't hardcode anything in the code but use the specific environment files to handle specific sinks and settings -per environment-).

xt0rted:
👋 I used the Serilog integration in a small test project (ended up sticking with the built-in logging) and found the sample project here a good starting point if you haven't seen it yet.

https://github.com/getsentry/sentry-dotnet/tree/master/samples/Sentry.Samples.AspNetCore.Serilog

PureKrome:
Looking at that example, I can't see where there's any specific ASPNET Core setup/settings?

```
               // Add Sentry integration with Serilog
                    // Two levels are used to configure it.
                    // One sets which log level is minimally required to keep a log message as breadcrumbs
                    // The other sets the minimum level for messages to be sent out as events to Sentry
                    .WriteTo.Sentry(s =>
                     {
                         s.MinimumBreadcrumbLevel = LogEventLevel.Debug;
                         s.MinimumEventLevel = LogEventLevel.Error;
                     }))
```

That's pretty generic I thought -> meaning I could do that on a console app / background tasks.

```
                // Add Sentry integration
                // It can also be defined via configuration (including appsettings.json)
                // or coded explicitly, via parameter like:
                // .UseSentry("dsn") or .UseSentry(o => o.Dsn = ""; o.Release = "1.0"; ...)
                .UseSentry()
```

Again ... this _feels like_ it's pretty agnostic and not specific to ASPNET Core.

My goal is to add ASPNET core *data* to any errors ... and define the sentry sink via configuration. (right now, i have sentry defined in the `appsettings.release.config` file and it works great .. but is just missing any ASPNET core stuff, like the request payload).

bruno-garcia:
Hey @PureKrome,

I'm dogfooding the SDK in NuGetTrends and there we got both Serilog and ASP.NET Core so hopefully a useful example.

The packages:

[...]

PR #637, opened 2020-12-02
Moved DefaultTags to SentryOptions.

### LINKED ISSUE #635: SentryOptions to offer DefaultTags and DefaultExtraData

- [ ] ASP.NET MVC
- [ ] ASP.NET Web API (OWIN)
- [X] ASP.NET Core
- [ ] WPF
- [ ] WinForms
- [ ] Xamarin
- [ ] Other:

- [ ] .NET Framework
- [ ] Mono
- [X] .NET Core

* Version:

- [X] Sentry
- [X] Sentry.Serilog
- [ ] Sentry.NLog
- [ ] Sentry.Log4Net
- [ ] Sentry.AspNetCore

* Version: Latest / 2.1.8

This issue is based upon a conversation I had with @bruno-garcia in #588 

BG: 
> Sentry.AspNetCore has appsettngs.json support for default tags. This is missing from Serilog.Sinks.Sentry. A PR to accept this would be considered.

and 

> We could move default tags to SentryOptions so all integrations could leverage.

So my original idea was to have `Sentry,Serilog` be able to send `Tag` and `ExtraData` up to Sentry. (of course, any other Sentry 'sinks' could leverage this, too).

This convo happens to also be linked to how MEL + Scopes creates (I believe) Sentry `LoggingEventProperties` which sinks like `Serilog.Sinks` could leverage. 

Finally, `Sentry.Serilog` should then be able to report back any `Tags` and `ExtraData` to Sentry.

## MAIN GOAL: 
- configuration defined in `appsettings.json` which Serilog reads/respects.
- bare minimum Sentry code in program.cs (settings defined via configuration files / environmental settings)
- App and nuget libraries all leverage MEL throughout the code (Sentry agnostic)
- Code uses MEL + Scopes. These 'scoped/temp' logging values should be sent up with their appropriate logging event, up to Sentry.

---

Having a quick look at the code, I've noticed that currently default Tags do exist, but here:

https://github.com/getsentry/sentry-dotnet/blob/e17703a77a4a5f20b44f233ecfa4e937d42542f8/src/Sentry.Extensions.Logging/SentryLoggingOptions.cs#L24-L30

I'm not familiar with the full Sentry code base, so I'm not sure of the reason for `DefaultTags` existing in `SentryLoggingOptions` .. but based on what @bruno-garcia suggested,  this should be moved to `SentryOptions`. This would also mean this code would also be moved:

https://github.com/getsentry/sentry-dotnet/blob/e17703a77a4a5f20b44f233ecfa4e937d42542f8/src/Sentry.Extensions.Logging/SentryLoggingOptionsExtensions.cs#L40-L52

Next, the `AspNetCoreEventProcessor` shouldn't be the only thing to process this, but _all_ EventProcessors:

https://github.com/getsentry/sentry-dotnet/blob/e17703a77a4a5f20b44f233ecfa4e937d42542f8/src/Sentry.AspNetCore/AspNetCoreEventProcessor.cs#L40

so I was thinking of moving that out to the `MainSentryEventProcessor`.

Now this is for `DefaultTags`. I was then going to (more or less) copy all of that and add `DefaultExtraData` (or should it be called `DefaultExtra`) which then allows the developer to have default extra data.

- Tags = indexed
- Extra Data = not indexed

Next, update maybe the `Samples.Console.Customized` and maybe `Samples.ME.Logging` to test this out. (NOTE: I cannot access that DSN bucket, so I'm assuming I would make my own and not PR that DNS).

Finally, test out Serilog with `Samples.Serilog` and make sure the data is getting added.

This is the way.

UPDATE:

The official docs state [that `Extra` should be deprecated away in favour of the word `Context`](https://docs.sentry.io/platforms/dotnet/enriching-events/context/). So if you're still happy with having the Default Context added, I can use that terminology instead?

PureKrome:
Fixes #635 

Other linked references: #257 and #588 

Items in this PR:

- [x] Move `DefaultTags` from `SentryLoggingOptions` to `SentryOptions` (as requested by @bruno-garcia )
- [X] Apply these default tags from `AspNetCoreEventProcessor` -> `MainSentryEventProcessor.cs`
- [X] Make sure the unit tests are moved.
- [X] Test MEL to make sure these default tags are applied
- [X] Test Serilog to make sure they can be defined in the `appsettings.json` 
- [x] Update `CHANGELOG.md`
- [ ] Any tests that I've forgotten? 
- [ ] Any documentation needs to be updated? (I think the Serilog documentation needs to be updated with [an example of how to set all the settings](https://github.com/getsentry/sentry-dotnet/blob/fd8a5ec0ef2a215a8c3a71ec9f0b1d2e405c1eff/src/Sentry.Serilog/SentrySinkExtensions.cs#L49-L83), via `appsettings.json`)

### DIFF
--- a/src/Sentry.Serilog/SentrySinkExtensions.cs
+++ b/src/Sentry.Serilog/SentrySinkExtensions.cs
@@ -308,7 +308,7 @@ namespace Serilog
 
             if (defaultTags?.Any() == true)
             {
-                foreach(var tag in defaultTags)
+                foreach (var tag in defaultTags)
                 {
                     sentrySerilogOptions.DefaultTags.Add(tag.Key, tag.Value);
                 }
--- a/src/Sentry.Serilog/SentrySinkExtensions.cs
+++ b/src/Sentry.Serilog/SentrySinkExtensions.cs
@@ -1,5 +1,7 @@
 using System;
+using System.Collections.Generic;
 using System.IO.Compression;
+using System.Linq;
 using System.Net;
 using Sentry;
 using Sentry.Internal;
@@ -42,6 +44,7 @@ namespace Serilog
         /// <param name="reportAssemblies">Whether or not to include referenced assemblies in each event sent to sentry. Defaults to <see langword="true"/>. <seealso cref="SentryOptions.ReportAssemblies"/></param>
         /// <param name="deduplicateMode">What modes to use for event automatic de-duplication. <seealso cref="SentryOptions.DeduplicateMode"/></param>
         /// <param name="initializeSdk">Whether to initialize this SDK through this integration. <seealso cref="SentrySerilogOptions.InitializeSdk"/></param>
+        /// <param name="defaultTags">Defaults tags to add to all events. <seealso cref="SentryOptions.DefaultTags"/></param>
         /// <returns><see cref="LoggerConfiguration"/></returns>
         /// <example>This sample shows how each item may be set from within a configuration file:
         /// <code>
@@ -75,7 +78,11 @@ namespace Serilog
         ///                     "diagnosticsLevel": "Debug",
         ///                     "reportAssemblies": false,
         ///                     "deduplicateMode": "All",
-        ///                     "initializeSdk": true
+        ///                     "initializeSdk": true,
+        ///                     "defaultTags": {
+        ///                         "key-1", "value-1",
+        ///                         "key-2", "value-2"
+        ///                     }
         ///                 }
         ///             }
         ///         ]
@@ -106,7 +113,8 @@ namespace Serilog
             SentryLevel? diagnosticsLevel = null,
             bool? reportAssemblies = null,
             DeduplicateMode? deduplicateMode = null,
-            bool? initializeSdk = null)
+            bool? initializeSdk = null,
+            Dictionary<string, string>? defaultTags = null)
         {
             return loggerConfiguration.Sentry(o => ConfigureSentrySerilogOptions(o,
                 dsn,
@@ -130,7 +138,8 @@ namespace Serilog
                 diagnosticsLevel,
                 reportAssemblies,
                 deduplicateMode,
-                initializeSdk));
+                initializeSdk,
+                defaultTags));
         }
 
         /// <summary>
@@ -159,8 +168,9 @@ namespace Serilog
         /// <param name="reportAssemblies">Whether or not to include referenced assemblies in each event sent to sentry. Defaults to <see langword="true"/>. <seealso cref="SentryOptions.ReportAssemblies"/></param>
         /// <param name="deduplicateMode">What modes to use for event automatic de-duplication. <seealso cref="SentryOptions.DeduplicateMode"/></param>
         /// <param name="initializeSdk">Whether to initialize this SDK through this integration. <seealso cref="SentrySerilogOptions.InitializeSdk"/></param>
+        /// <param name="defaultTags">Defaults tags to add to all events. <seealso cref="SentryOptions.DefaultTags"/></param>
         public static void ConfigureSentrySerilogOptions(
-            SentrySerilogOptions sentrySerilogOptions,
+        SentrySerilogOptions sentrySerilogOptions,
             string? dsn = null,
             LogEventLevel? minimumEventLevel = null,
             LogEventLevel? minimumBreadcrumbLevel = null,
@@ -182,7 +192,8 @@ namespace Serilog
             SentryLevel? diagnosticsLevel = null,
             bool? reportAssemblies = null,
             DeduplicateMode? deduplicateMode = null,
-            bool? initializeSdk = null)
+            bool? initializeSdk = null,
+            Dictionary<string, string>? defaultTags = null)
         {
             if (!string.IsNullOrWhiteSpace(dsn))
             {
@@ -293,6 +304,14 @@ namespace Serilog
             if (initializeSdk.HasValue)
             {
                 sentrySerilogOptions.InitializeSdk = initializeSdk.Value;
+            }
+
+            if (defaultTags?.Any() == true)
+            {
+                foreach(var tag in defaultTags)
+                {
+                    sentrySerilogOptions.DefaultTags.Add(tag.Key, tag.Value);
+                }
             }
         }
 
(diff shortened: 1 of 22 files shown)


PureKrome:
Here's a screen shot of me playing around with the `MEL Logging Sample` :

given this code in the initialization steps:


image

this is what is logged to Sentry:

image

So here

PureKrome:
### Serilog example

- ✔️ All configuration settings are handled via `appsettings.json`
- ❌ No configuration is coded.
- ✔️ `Serilog.Settings.Configuration` nuget is used.

First, lets define the app to use `Serilog.Settings.Configuration` as the logging mechanism.

image

And set this up in our app. I've just used a console app, so I'll also need to manually define an `appsettings.json` configuration source:

image

and finally, the money shot => our config file.

🌟 🌟 NOTE : i've added in 2 other not-related-to-this-repo _sinks_ to try and replicate a _real world scenario_ 👍🏻 
- `Console` : this is used for all localhost development. Possibly also for Containers.
- `Seq` : nice way to grab _all_ logging messages, not just errors. Dev and Prod
- `Sentry` : this sink we're testing.

```
{
  "Serilog": {
    "Using": [ "Serilog.Sinks.Seq", "Sentry.Serilog" ], // Make sure we link to the assemblies so Serilog knows what to do.
    "MinimumLevel": {
      "Default": "Debug"
    },
    "WriteTo": [
      {
        "Name": "Seq", 
        "Args": {
          "serverUrl": "http://localhost:5301",
          "restrictedToMinimumLevel": "Information"
        }
      },
      {
        "name": "Sentry",
        "Args": {
          "dsn": "https://4935ed5530614e04afb34982f68ce3cd@o176569.ingest.sentry.io/5541257",
          "AttachStacktrace": true,
          "SendDefaultPii": true,
          "debug": true,
          "IncludeRequestPayload": true,
          "IncludeActivityData": true,
          "DefaultTags": { // A Dictionary / key-value section of application tags....
            "key-1": "value-1",
            "key-2": "value-2",
            "key-3": "value-3"
          }
        }
      }
    ],
    "Enrich": [ "FromLogContext" ],
    "Properties": { // BONUS!!! Properties are stored as EXTRA DATA !!!
      "ApplicationName": "Accounts.Api"
    }
  }
}
```

and this is a sample Sentry report:

image

(and for that bonus `Properties` section) :

image

PureKrome:
@bruno-garcia Okay - ready for draft-review 🍰

src/Sentry.Serilog/SentrySinkExtensions.cs
```diff
@@ -294,6 +305,14 @@ public static class SentrySinkExtensions
             {
                 sentrySerilogOptions.InitializeSdk = initializeSdk.Value;
             }
+
+            if (defaultTags?.Any() == true)
+            {
+                foreach(var tag in defaultTags)
```
bruno-garcia, line 311:
```suggestion
                foreach (var tag in defaultTags)
```

src/Sentry/SentryOptions.cs
```diff
@@ -383,6 +383,14 @@ public bool Debug
         /// </remarks>
         public TimeSpan CacheFlushTimeout { get; set; } = TimeSpan.FromSeconds(1);
 
+        /// <summary>
+        /// Defaults tags to add to all events. (These are indexed by Sentry).
+        /// </summary>
+        /// <remarks>
+        /// If the key already exists in the event, it will not be overwritten by a default tag.
+        /// </remarks>
+        public Dictionary<string, string> DefaultTags { get; } = new Dictionary<string, string>();
```
bruno-garcia, line 392: To avoid allocating this map as a result of instantiating this class, could have a private backing field, nullable, and on the getter here we set instance?

This way we only create on the first use.

As this SDK becomes more general purpose I want to be mindful of the initial cost you pay just to `Init` Sentry on a basic use case.


Something like
```suggestion
        public Dictionary<string, string> DefaultTags => _defaultTags =?? new Dictionary();
```

bruno-garcia (commented):
It's far ahead of a draft :)

Thanks! Just left a note

src/Sentry/SentryOptions.cs
PureKrome: Ah yes! good use of the null-coalescing operator.

PureKrome:
🔧 Fixing compiler errors. plz hold.

PureKrome:
@bruno-garcia ok - good to review (except the changelog CI is having a cry - not sure why).

test/Sentry.Tests/Internals/MainSentryEventProcessorTests.cs
```diff
@@ -434,6 +434,24 @@ public void Process_CultureInfo_ValuesOnContext(Action<CultureInfo> setter, Func
             }
         }
 
+        [Fact]
+        public void Process_AppliesDefaultTags()
```
bruno-garcia, line 438: Could we have one test that makes sure default tags don't override the event tag please?

bruno-garcia: Looks good and I know we didnt have thst test to begin with but we should have it

bruno-garcia:
#skip-changelog

bruno-garcia:
The changelog thing chokes on external PRs because it requires the GH auth. Just ignore it, we'll merge even if that's red

test/Sentry.Tests/Internals/MainSentryEventProcessorTests.cs
PureKrome: 🔧

bruno-garcia (approved):

bruno-garcia:
Awesome! Thanks once again! @PureKrome

PureKrome:
@bruno-garcia Cheers! I do have one last question - what is the process to get the [Sentry Serilog documentation](https://docs.sentry.io/platforms/dotnet/guides/serilog/) updated to include information about how to set the settings, via the `appsettings.config` file?

bruno-garcia:
@PureKrome oh it would be awesome if you could help out there.
Docs for specific configurations of integrations are really lagging behind SDK developments.

There are [docs about contributing to the docs](https://docs.sentry.io/contributing/).

bruno-garcia:
@PureKrome you don't even need to build the docs locally, when you open a PR, the branch gets deployed and you can browser it and see what it looks like.

PureKrome:
Ta! I'll read and make a PR.
cheers!

bruno-garcia:
Oh that's awesome! Thanks a lot.

Feel free to @ me there

bruno-garcia:
If you have any problems with the docs, write me on Discord and I or other folks there can help you on the fly: https://discord.gg/sentry

PR #641, opened 2020-12-05
ref: rename CacheFlushTimeout to InitCacheFlushTimeout.

### LINKED ISSUE #638: Rename CacheFlushTimeout to something more descriptive
https://github.com/getsentry/sentry-dotnet/blob/429c58bbaa2aae2ad02c772d6f04c19e3039481a/src/Sentry/SentryOptions.cs#L384

Ideally it should clarify that it: Blocks initializing the SDK if offline caching is enabled and events are on disk to be sent. Something like `BlockInitForCacheFlushTimeout` but ideally less horrible.

Which makes me realize, before using the field, make sure the CacheDir isn't null.

lucas-zimerman:
closes #638 

🥟Tricky name to replace, it points out that it defines the timeout to Flush the Cache initialization.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## vNext
+
+* Ref renamed `CacheFlushTimeout` to `InitCacheFlushTimeout` (#638) @lucas-zimerman
+
 ## 3.0.0-alpha.6
 
 * Abandon ValueTask #611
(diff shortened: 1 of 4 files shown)


bruno-garcia:
Isn't this a good candidate to get a changelog item? :)

lucas-zimerman:
> Isn't this a good candidate to get a changelog item? :)

I had some doubts since it was an item introduced on 3.0.0 and renamed it in the same version.

bruno-garcia (approved):
👍

PR #644, opened 2020-12-08
ref: moved Sentry from namespace from Sentry.Protocol to Sentry (#643)

### LINKED ISSUE #643: Move SentryId to Sentry namespace
Move SentryId to Sentry instead of Sentry.Protocol to simplify [snippets such as the User Feedback](https://docs.sentry.io/platforms/dotnet/guides/aspnetcore/enriching-events/user-feedback/#integration):

image

lucas-zimerman:
With that, the user is not required to use the Sentry.Protocol namespace only for calling the user feedback feature.
closes #643.

### DIFF
--- a/.github/workflows/danger.yml
+++ b/.github/workflows/danger.yml
@@ -9,7 +9,6 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
-      - uses: actions/setup-node@v1
       - run: npx danger@10.5.2 ci
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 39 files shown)


lucas-zimerman:
@bruno-garcia Dange is going to complain because it expects the changelog to be in the following format.
`* Ref moved SentryId from namespace Sentry.Protocol to Sentry (#643) @lucas-zimerman ([#644](https://github.com/getsentry/sentry-dotnet/pull/644))`
But it's a bit verbose, I'd rather leave it as 
`* Ref moved SentryId from namespace Sentry.Protocol to Sentry (#643) @lucas-zimerman`
and refactor the dangerfile to follow our standard of the changelog.

bruno-garcia (approved):

bruno-garcia:
@lucas-zimerman OK, we gotta change that format then so it stops bugging us.
That rule was taken from getsentry/relay

PR #658, opened 2020-12-16
Don't use GUID for SpanId

Tyrrrz:
_#skip-changelog_

### DIFF
--- a/src/Sentry/Protocol/SpanId.cs
+++ b/src/Sentry/Protocol/SpanId.cs
@@ -1,8 +1,6 @@
 using System;
-using System.Linq;
 using System.Text;
 using System.Text.Json;
-using Sentry.Internal.Extensions;
 
 namespace Sentry.Protocol
 {
--- a/src/Sentry/Protocol/SpanId.cs
+++ b/src/Sentry/Protocol/SpanId.cs
@@ -82,12 +82,12 @@ namespace Sentry.Protocol
         public static bool operator !=(SpanId left, SpanId right) => !(left == right);
 
         /// <summary>
-        /// The <see cref="Guid"/> from the <see cref="SentryId"/>.
+        /// The <see cref="string"/> from the <see cref="SentryId"/>.
         /// </summary>
         public static implicit operator string(SpanId id) => id._value;
 
         /// <summary>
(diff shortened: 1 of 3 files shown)


Tyrrrz:
@bruno-garcia fixed a random null ref bug in tests that was unrelated to this

src/Sentry/Protocol/SpanId.cs
```diff
@@ -32,14 +42,14 @@ namespace Sentry.Protocol
...
-        // This is obviously a bad idea as it invalidates GUID's uniqueness properties
-        // (https://devblogs.microsoft.com/oldnewthing/20080627-00/?p=21823)
-        // but all other SDKs do it this way, so we have no choice but to comply.
         /// <summary>
         /// Generates a new Sentry ID.
         /// </summary>
-        public static SpanId Create() => new(Guid.NewGuid().ToString("n").Substring(0, 16));
+        public static SpanId Create() => new(
+            Enumerable.Range(0, 16)
+                .Select(_ => AllowedChars[Random.Next(0, AllowedChars.Length)])
```
bruno-garcia, line 50: `Random` isn't thread-safe. You can't call this here from the static method.

Also, I'm curious if you benchmark this vs a plain `for` what's the result compared to the enumerator, with ToArray then allocating a string for the char array.

Maybe there's a simpler way to do this?

bruno-garcia (changes requested):

src/Sentry/Protocol/SpanId.cs
Tyrrrz: What's the common workaround for syncing `Random`? I imagine using locks would not be desirable here.

`for` allocates less I'm sure, but I went with this approach because of its simplicity. I'll rewrite to for.

bruno-garcia: I'm starting to think the original approach is better here. Or at least we need more convincing of its limitations, to justify this since it's has its own shortcomings

Tyrrrz: Let me try to run some benchmarks to see how often it results in collisions.

Tyrrrz: Looks like the problem may indeed be overexaggerated.

image

image

PR #660, opened 2020-12-17
Make all fields on UserFeedback optional

### LINKED ISSUE #651: Issues with UserFeedback

- [ ] ASP.NET MVC
- [ ] ASP.NET Web API (OWIN)
- [ ] ASP.NET Core
- [ ] WPF
- [ ] WinForms
- [x] Xamarin
- [x] Other:

- [ ] .NET Framework
- [x] Mono
- [x] .NET Core

* Version:

- [x] Sentry
- [ ] Sentry.Serilog
- [ ] Sentry.NLog
- [ ] Sentry.Log4Net
- [ ] Sentry.AspNetCore

* Version:

1. When trying to register an user feedback with cache enabled, Sentry reports a serialization error and drops the envelope.
2. When sending the user feedback without cache, Sentry server seems to be dropping feedbacks that doesn't contain the username field.

EDIT: Validated with the Sentry.Samples.Console.Basic by setting the user as null on line 155.  (The output request is correct so I guess the issue is on Sentry backend?)

EDIT2: the cache issue can also be validated on the same project by setting the o.CacheDirectoryPath = "./Sentry";
the error might be related to             var name = json.GetProperty("name").GetStringOrThrow(); since the name could be null

Tyrrrz:
Also changed the order of the parameters on the constructor to be more natural. Will need to update the feedback docs to match.

Closes #651

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## vNext
+
+- Make all fields on UserFeedback optional. (#660) @Tyrrrz
+
 ## 3.0.0-alpha.7
 
 * Ref moved SentryId from namespace Sentry.Protocol to Sentry (#643) @lucas-zimerman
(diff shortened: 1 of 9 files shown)


bruno-garcia (approved):
I really hate the idea of sending junk to the server though

PR #662, opened 2020-12-17
Include parameters in stack frames

### LINKED ISSUE #647: Send full method signatures to Sentry instead of just the method name
For integration with https://github.com/getsentry/sentry-visualstudio (and as a good practice in general when dealing with method overloads) it would be the best if Sentry sent the full method signatures where available in the stacktrace.

The parameter list is available on the `MethodInfo` from the StackTrace Frame:
```cs
var stacktrace = new StackTrace(e);
foreach (var frame in stacktrace.GetFrames())
{
    var method = frame.GetMethod();
    var methodParams = method.GetParameters(); // These are our parameters
    foreach(var parameter in methodParams){
        var paramName = parameter.ParameterType.Name; // The name of the parameter type
        var fullName = parameter.ParameterType.FullName; // Full name including namespace
    }
}
```

Tyrrrz:
Closes #647

### DIFF
--- a/src/Sentry/Extensibility/SentryStackTraceFactory.cs
+++ b/src/Sentry/Extensibility/SentryStackTraceFactory.cs
@@ -124,13 +124,20 @@ namespace Sentry.Extensibility
         protected SentryStackFrame InternalCreateFrame(StackFrame stackFrame, bool demangle)
         {
             const string unknownRequiredField = "(unknown)";
+
             var frame = new SentryStackFrame();
             if (GetMethod(stackFrame) is { } method)
             {
                 // TODO: SentryStackFrame.TryParse and skip frame instead of these unknown values:
                 frame.Module = method.DeclaringType?.FullName ?? unknownRequiredField;
                 frame.Package = method.DeclaringType?.Assembly.FullName;
-                frame.Function = method.Name;
+
+                // Include parameters in the function name
+                frame.Function =
+                    method.Name +
+                    '(' +
+                    string.Join(", ", method.GetParameters().Select(p => $"{p.ParameterType.Name} {p.Name}")) +
+                    ')';
             }
 
             frame.InApp = !IsSystemModuleName(frame.Module);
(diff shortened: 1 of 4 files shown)


src/Sentry/Extensibility/SentryStackTraceFactory.cs
```diff
@@ -124,13 +124,20 @@ internal IEnumerable<SentryStackFrame> CreateFrames(StackTrace stackTrace, bool
...
                 // TODO: SentryStackFrame.TryParse and skip frame instead of these unknown values:
                 frame.Module = method.DeclaringType?.FullName ?? unknownRequiredField;
                 frame.Package = method.DeclaringType?.Assembly.FullName;
-                frame.Function = method.Name;
+
+                // Include parameters in the function name
+                frame.Function =
+                    method.Name +
+                    '(' +
+                    string.Join(", ", method.GetParameters().Select(p => $"{p.ParameterType.Name} {p.Name}")) +
```
kanadaj, line 139: Let's add a & after the type name if there is a modifier like in/out/ref. The type info should have this, right?

kanadaj (changes requested):
The standard .NET stacktrace method signature includes a & for all in/out/ref methods:
```cs
        public static void TestWithIn(string a, in int b)        
        {
            throw new NotImplementedException();
        }
```
=>
```cs
StacktraceRefOutTest.Program.TestWithIn(String a, Int32& b)
```
This shouldn't be too hard to implement if the type info contains any information about these keywords. It is possible that instead we need to look at some sort of reference type flag if the individual modifiers are not available.

Tyrrrz:
Before:

image

After:

image

Tyrrrz:
@bruno-garcia ready for review

src/Sentry/Extensibility/SentryStackTraceFactory.cs
Tyrrrz: Actually already done. `p.ParameterType.Name` includes the ampersand.

bruno-garcia (approved):
:shipit:

PR #667, opened 2020-12-20
Blazor Wasm JavaScript interop

bruno-garcia:
Exceptions thrown in a C# to JavaScript calls get reported by the .NET SDK.

JavaScript code that isn't directly called by .NET Wasm doesn't get captured without the JavaScript SDK. `setTimeout` or a call from the DOM to a JS function for example.


jamescrosswell:
Closing this as a cold case. We an reopen if necessary/appropriate.

PR #670, opened 2020-12-21
Attachments

Tyrrrz:
Includes a rework to `Scope` and `IScope` (i.e. the deletion of the latter). `IScope` and the leaky abstractions it created have been accumulating technical debt for a while. It was a source of pain when implementing transactions (and one of the reasons they were not implemented as elegantly as they should be) and it's now also impeding the implementation of attachments as well. So as a result, I decided to include this refactoring as part of the attachments PR.


src/Sentry/Protocol/Envelopes/Envelope.cs
```diff
@@ -73,18 +74,24 @@ public async Task SerializeAsync(Stream stream, CancellationToken cancellationTo
         /// <summary>
         /// Creates an envelope that contains a single event.
         /// </summary>
-        public static Envelope FromEvent(SentryEvent @event)
+        // TODO: use builders for this?
```
bruno-garcia, line 77: Can't use a builder because there are many ways this would be built wrong.

```suggestion
```

Can't have 2 events in. can't have 1 event and 1 transaction. Can have many sessions.
Can have stand alone attachment, but must include event id in the envelope header.

I'd continue with some fixed public methods that create it with the right/valid use cases. For now.

Tyrrrz: How is a standalone attachment sent?

bruno-garcia: @Tyrrrz you need to give a event id at least

Tyrrrz: No, I mean from the SDK perspective. I didn't see `Sentry.CaptureAttachment` in any other SDK.

bruno-garcia:
The approach of dealing with failures on Java:
https://github.com/getsentry/sentry-java/pull/1134

src/Sentry/Protocol/Envelopes/Envelope.cs
bruno-garcia: @Tyrrrz I don't think we have it, the idea is to only send an attachments to an event. So:

`CaptureEvent(@event, s => s.AddAttachment(...))`

Although we still don't have the overload to take a `Action<Scope>`.

Raised it: https://github.com/getsentry/sentry-dotnet/issues/675

Tyrrrz:
TODO: support attachment types

image

Tyrrrz:
image

Tyrrrz:
@bruno-garcia ready for review

src/Sentry/Protocol/Attachment.cs
```diff
@@ -0,0 +1,92 @@
...
+        /// <summary>
+        /// Standard attachment without special meaning.
+        /// </summary>
+        Default,
+
+        /// <summary>
+        /// Minidump file that creates an error event and is symbolicated.
+        /// The file should start with the <code>MDMP</code> magic bytes.
+        /// </summary>
+        MiniDump,
```
bruno-garcia, line 21: Minidump is one word:

```suggestion
        Minidump,
```

src/Sentry/ScopeExtensions.cs
```diff
@@ -146,109 +145,32 @@ public static bool HasUser(this IScope scope)
...
-            }
-        }
-
         /// <summary>
         /// Set all items as tags.
         /// </summary>
         /// <param name="scope">The scope.</param>
         /// <param name="tags"></param>
-        public static void SetTags(this IScope scope, IEnumerable<KeyValuePair<string, string>> tags)
+        public static void SetTags(this IEventLike scope, IEnumerable<KeyValuePair<string, string>> tags)
```
bruno-garcia, line 173: We should rename all variables too now then? Isn't this a refactor on the IDE?

same in other methods of this change
```suggestion
        public static void SetTags(this IEventLike eventLike, IEnumerable<KeyValuePair<string, string>> tags)
```

src/Sentry/ScopeExtensions.cs
```diff
@@ -257,108 +179,18 @@ public static void SetTags(this IScope scope, IEnumerable<KeyValuePair<string, s
         }
 
         /// <summary>
-        /// Removes a tag from the <see cref="IScope"/>.
+        /// Removes a tag from the <see cref="Scope"/>.
```
bruno-garcia, line 182:
```suggestion
        /// Removes a tag from the <see cref="IEventLike"/>.
```

src/Sentry/SentryClient.cs
```diff
@@ -113,7 +113,7 @@ public void CaptureTransaction(Transaction transaction)
                 throw new ObjectDisposedException(nameof(SentryClient));
             }
 
-            if (transaction.SpanId.Equals(SentryId.Empty))
+            if (transaction.SpanId.Equals(SpanId.Empty))
```
bruno-garcia, line 116: Random performance bug fixes in the middle? :)

bruno-garcia (approved):
Some nits but other than that lets merge!

src/Sentry/SentryClient.cs
Tyrrrz: Not performance, actually a logical bug that wasn't caught because `Equals` is not type-safe.

src/Sentry/ScopeExtensions.cs
Tyrrrz: Are you okay with `IEventLike` as the name? It was something I put without thinking, so maybe you have better ideas.

src/Sentry/SentryClient.cs
bruno-garcia: I meant "the performance product".
Anyway just nagging because there's a lot of different changes here but happy to get this in. Thanks!

Tyrrrz:
Looks like locally `git merge origin/main` works without conflicts, but on github it always shows conflicts in `CHANGELOG.md`. Maybe github's merge doesn't take `.gitattributes` into account somehow?

test/Sentry.NLog.Tests/SentryTargetTests.cs
```diff
@@ -390,24 +390,24 @@ public void Log_WithFormat_EventCaptured()
...
             var sut = _fixture.GetLogger();
 
             sut.Error("message {SourceContext}", "Sentry.NLog");
 
-            _fixture.Hub.DidNotReceive().ConfigureScope(Arg.Any<Action<IScope>>());
+            _fixture.Hub.DidNotReceive().ConfigureScope(Arg.Any<Action<Scope>>());
         }
 
-        [Fact]
+        [Fact(Skip = "Failing for unknown reason")]
```
bruno-garcia, line 403: https://github.com/getsentry/sentry-dotnet/issues/686

bruno-garcia (approved):

PR #672, opened 2020-12-22
wpf sample

bruno-garcia:


bruno-garcia:
Not surprisingly the solution won't build on Linux and macOS:
https://github.com/dotnet/wpf/issues/688

PR #679, opened 2020-12-24
Embed PDBs

### LINKED ISSUE #678: Drop snupkg in favor of embedded pdbs
We've been using `snupkg` during the 3.0.0 alphas and so far it seems to only add pain for the little benefit of a tiny smaller final nupkg. Seems like [there's a consensus at least among some folks](https://twitter.com/brungarc/status/1342176040504991744?s=20) that we should just embed the pdb in the dll.

`<DebugType>embedded</DebugType>`

Tyrrrz:
Closes #678

_#skip-changelog_


Tyrrrz:
> will increase coverage by 1.95%.

easy coverage

bruno-garcia (approved):
🥳

PR #680, opened 2020-12-24
Inject changelog into release notes

Tyrrrz:
Nuspec looks like this:

image

_#skip-changelog_

### DIFF
--- a/src/Directory.Build.props
+++ b/src/Directory.Build.props
@@ -23,7 +23,7 @@
     <PackageProjectUrl>https://sentry.io</PackageProjectUrl>
     <PackageLicenseExpression>MIT</PackageLicenseExpression>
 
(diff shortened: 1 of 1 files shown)


bruno-garcia:
hm.. kind of like this idea. But at the same time it'll be an ever growing changelog added to each package?
Isn't the link just straight forward?
Folks can click to see diff between versions etc

also, why CI broken?

>   Failed to restore D:\a\sentry-dotnet\sentry-dotnet\test\Sentry.Tests\Sentry.Tests.csproj (in 330 ms).

Tyrrrz:
> hm.. kind of like this idea. But at the same time it'll be an ever growing changelog added to each package?
Isn't the link just straight forward?

This shows the content right in the IDE, without forcing the user to open the browser.

> also, why CI broken?

As I mentioned earlier, Github Actions windows images are broken currently with .NET.

bruno-garcia (approved):
OK lets give it a try.

PR #682, opened 2020-12-24
Include assembly name in default release

### LINKED ISSUE #668: Change default, auto detected release format
By default in Sentry a release is organization wide.
That means that if you have two projects, and release it as `1.0`, they'll both show up under the same release in Sentry.

An issue was released on the docs to include in all snippets a prefix like `io.sentry.app@1.1.1+234` as used in Mobile: https://github.com/getsentry/sentry-docs/pull/2793

In .NET we use the Entry Assembly version but not the name. The change is:
Change the Release locator to append the Assembly.Name:

https://github.com/getsentry/sentry-dotnet/blob/0e93afe83546f8ba639c5f68ae8d9c3b72466ed5/src/Sentry/Internal/ReleaseLocator.cs#L11

Tyrrrz:
Closes #668


src/Sentry/Internal/ApplicationVersionLocator.cs
```diff
@@ -9,15 +9,20 @@ internal static class ApplicationVersionLocator
...
+            var version = asm.GetNameAndVersion().Version;
 
             return !string.IsNullOrEmpty(version)
                    // If it really was on of the following, app would need to be set explicitly since these are defaults.
                    && version != "0.0.0"
                    && version != "1.0.0"
                    && version != "0.0.0.0"
                    && version != "1.0.0.0"
-                ? version
+                ? $"{asm.GetName().Name}@{version}"
```
lucas-zimerman, line 25: I think you don't need to check if the version is 0.0.0, 1.0.0 anymore.
@bruno-garcia  ping

bruno-garcia: yeah, lets just alwayas send this.

need to update docs though

bruno-garcia: Needs fixing here: https://github.com/getsentry/sentry-dotnet/blob/main/docs/manual/manual.md#release

Actually that could be deleted. It was before proper Sentry docs was added.

Needs fixing here: https://github.com/getsentry/sentry-docs/edit/master/src/platforms/common/configuration/releases.mdx

bruno-garcia (changes requested):

bruno-garcia (approved):
Lets merge this and fix in a follow up PR

PR #695, opened 2020-12-28
Always send a default release

### LINKED ISSUE #687: Always send a default release
Follows: https://github.com/getsentry/sentry-dotnet/pull/682/

We don't report release name if default value exist like `1.0`. Let's change it to always report it.
Users can get the value of release tracking from the beginning. And learn more about how to set a version on their app if they are no familiar yet.

Needs fixing here: https://github.com/getsentry/sentry-dotnet/blob/main/docs/manual/manual.md#release

Actually that could be deleted. It was before proper Sentry docs was added.

Needs fixing here: getsentry/sentry-docs/edit/master/src/platforms/common/configuration/releases.mdx

Tyrrrz:
Closes #687


bruno-garcia (approved):
Thanks!

PR #702, opened 2020-12-29
Fix Windows CI/CD

Tyrrrz:
_#skip-changelog_


Tyrrrz:
Appveyor broke on a flaky test

bruno-garcia:
Why do we have so many flaky tests :~

bruno-garcia:
`MainSentryEventProcessor` has side effect because it relies on caching the release version

PR #703, opened 2020-12-29
fix: flaky tests

bruno-garcia:
One of the flaky tests in #702

#skip-changelog


Tyrrrz:
That looks like a very hacky fix, I think it's better to just set up proper environment isolation instead so that environment variables are not shared. Or, alternatively, use contexts to pass environment variables. Or, even more alternatively, use DI and `IEnvironment`, which I guess you would prefer more.

IMO, bad code is worse than flaky tests, so it's better to address this at the root.

bruno-garcia:
@Tyrrrz I agree, feel free to raise an issue and refactor it.
Testing the code against the env var isn't going to scale anyway.

until a better fix is in, taking the code with static side effect as a dependency for unit testing is the way to go. Better than flaky tests anyway and took a min to push

PR #705, opened 2020-12-30
Limit attachment size

### LINKED ISSUE #704: Disallow sending attachments over 20MiB in size
Core culprit of #688

And deal with error such as file not found, permission to read etc by catching and logging.

Simply ignore the envelope item. If no items in the envelope, don't send it at all (ideally) but it's OK to send an empty envelope worst case, as Sentry supports that.

Tyrrrz:
Closes #704


src/Sentry/Internal/Polyfills.cs
```diff
@@ -2,15 +2,16 @@
 
 // Polyfills to bridge the missing APIs in older versions of the framework/standard.
 // In some cases, these just proxy calls to existing methods but also provide a signature that matches .netstd2.1
-#if NET461 || NETSTANDARD2_0
 
+using System.Net.Http;
+using System.Threading;
+using System.Threading.Tasks;
```
bruno-garcia, line 8: Without having the using directive, when using "refactoring" it deletes them as it doesn't know it's used by the code currently not compiling due to #ifdef

ideally only what's needed is available

src/Sentry/Scope.cs
```diff
@@ -197,7 +197,24 @@ public void AddBreadcrumb(Breadcrumb breadcrumb)
...
         /// </summary>
-        public void AddAttachment(Attachment attachment) => _attachments.Add(attachment);
+        public void AddAttachment(Attachment attachment)
+        {
+            // Limit file size
+            if (attachment.Length > Options.MaxAttachmentSize)
+            {
+                Options.DiagnosticLogger?.LogWarning(
+                    "Attachment '{0}' dropped because it's too large.",
+                    attachment.FileName
```
bruno-garcia, line 207: How large?

```suggestion
                    "Attachment '{0}' dropped because it's too large {1}.",
                    attachment.FileName, attachment.Length
```

src/Sentry/ScopeExtensions.cs
```diff
@@ -281,11 +282,35 @@ public static void AddEventProcessors(this Scope scope, IEnumerable<ISentryEvent
...
+            this Scope scope,
+            Stream stream,
+            string fileName,
+            AttachmentType type = AttachmentType.Default,
+            string? contentType = null)
+        {
+            var length = stream.TryGetLength();
+            if (length is null)
+            {
+                scope.Options.DiagnosticLogger?.LogWarning(
```
bruno-garcia, line 304: Would be nice to document this. "If you're passing a stream .. "

src/Sentry/ScopeExtensions.cs
```diff
@@ -307,10 +332,21 @@ public static void AddEventProcessors(this Scope scope, IEnumerable<ISentryEvent
...
         {
-            var stream = File.OpenRead(filePath);
-            var fileName = Path.GetFileName(filePath);
+            try
+            {
+                var stream = File.OpenRead(filePath);
+                var fileName = Path.GetFileName(filePath);
 
-            scope.AddAttachment(stream, fileName, type, stream.Length, contentType);
+                scope.AddAttachment(stream, fileName, type, contentType);
```
bruno-garcia, line 340: If it throws here we don't call dispose on the stream.

src/Sentry/SentryOptions.cs
```diff
@@ -423,6 +423,15 @@ public double TraceSampleRate
         /// </summary>
         public ISentryTraceSampler? TraceSampler { get; set; }
 
+        /// <summary>
+        /// Maximum allowed file size of attachments, in bytes.
+        /// Attachments above this size will be discarded.
+        /// </summary>
+        /// <remarks>
+        /// Regardless of this setting, attachments are also limited to 20mb on Relay.
```
bruno-garcia, line 431:
```suggestion
        /// Regardless of this setting, attachments are also limited to 20mb (compressed) on Relay.
```

test/Sentry.Tests/Protocol/ScopeExtensionsTests.cs
```diff
@@ -547,6 +573,57 @@ public void AddAttachment_FromFile_ArgumentsResolvedCorrectly()
...
+            // Assert
+            Assert.Empty(scope.Attachments);
+            Assert.Contains(logger.Entries, e =>
+                e.Message == "Attachment '{0}' dropped because it's too large." &&
+                e.Args[0].ToString() == "MyFile.bin"
+            );
+        }
+
+        [Fact]
+        public void AddAttachment_FromFile_FileDoesNotExist_Drops()
```
bruno-garcia, line 608: Do we have a file that throws to read?

bruno-garcia (approved):
Some minor notes but LGTM!

src/Sentry/Internal/Polyfills.cs
Tyrrrz: I know, but I haven't been able to come up with a better way :/
There are two different #if regions and they have different usings.
Previously I was able to scope it to a class, but these polyfills need to be in global namespace because they fill intrinsic type members, which would normally not require adding a using on the calling side.

Tyrrrz: Can probably add a ReSharper ignore comment about unused directives? Maybe that will stop the refactor from touching this file.

src/Sentry/ScopeExtensions.cs
Tyrrrz: Good point

test/Sentry.Tests/Protocol/ScopeExtensionsTests.cs
Tyrrrz: That will just throw during envelope serialization.

src/Sentry/Internal/Polyfills.cs
Tyrrrz: Yeah, that appears to work.

test/Sentry.Tests/Protocol/ScopeExtensionsTests.cs
bruno-garcia: I meant: Do we have a test that covers that caase? Ia the whole envelope thrown away or can we skip just the item like in Java and Objective-C?

src/Sentry/ScopeExtensions.cs
```diff
@@ -307,10 +338,21 @@ public static void AddEventProcessors(this Scope scope, IEnumerable<ISentryEvent
             AttachmentType type = AttachmentType.Default,
             string? contentType = null)
         {
-            var stream = File.OpenRead(filePath);
-            var fileName = Path.GetFileName(filePath);
+            try
+            {
+                var stream = File.OpenRead(filePath);
```
bruno-garcia, line 343: Oh I just realized this.

So the stream stays open for the whole duration of the app?

In the other SDKs we read when the SDK is about to be sent. On Windows this will stop the file from being written to also.

bruno-garcia: We should probably be specific about this on the docs: If you give a path, we'll read only when the event is going out.

Tyrrrz: Yeah. That's the implication of streaming in envelopes.

Tyrrrz: Well not for the duration of the app, but until the envelope is sent.

bruno-garcia: Why can't we hold the path and open the team only if or when an event is sent?

bruno-garcia: Lets discuss this before merging

bruno-garcia: [Going into the docs btw](https://github.com/getsentry/sentry-docs/pull/2826/files#diff-624052c7e178a12fa0c1f439d8fd6913d3ae7bcac1cc3d939658f9f4cd0a874fR47-R50).

> The simplest way to create an attachment is to use a `path`. **The SDK is going to read the contents of the file each time** it is
preparing an event or transaction and is going to add the attachment to the same [envelope](https://develop.sentry.dev/sdk/envelopes/).
If the SDK can't read the file it logs an error message and drops the attachment.

test/Sentry.Tests/Protocol/ScopeExtensionsTests.cs
Tyrrrz: We have this test in `BackgroundWorkerTests`:

image

This verifies that an exception thrown when sending envelope is swallowed. It would go via `Envelope.SerializeAsync` -> `HttpTransport.SendEnvelopeAsync` -> `BackgroundWorker.EnqueueEnvelope`.

src/Sentry/ScopeExtensions.cs
Tyrrrz: > Why can't we hold the path and open the team only if or when an event is sent?

Hm, I guess we can probably do that. However, it's worth noting a scenario that may happen in such case:

1. User adds `log.txt` as attachment. At the time of calling `AddAttachment` it had certain content. The operation was successful (file exists, everything is ok).
2. Before the envelope is sent, the application changes `log.txt` somehow so it's now technically a different file. Or even deleted.
3. The attachment sent to Sentry is different than the one the user initially "added".

This is a concern because the user needs to understand the internals of Sentry SDK, i.e. the fact that the envelopes are sent later and the fact that `AddAttachment` doesn't actually add anything at the moment of calling. This may lead to unexpected results.

On the other hand, locking the file (by holding the stream) obviously prevents that.

Tyrrrz:
@bruno-garcia please check my latest comments.

src/Sentry/Internal/Polyfills.cs
bruno-garcia: you can just add `#if` and the right `using` under it. But no strong feelings about this, if it doesn't get dropped because of a refactor, it's OK.

src/Sentry/ScopeExtensions.cs
Tyrrrz: okay, makes sense

bruno-garcia (approved):
Approved again

PR #709, opened 2021-01-01
ref: Disable Mono StackTrace Factory.

lucas-zimerman:
As discussed on Discord, the MonoStackTraceFactory lacks any Assembly information, generating a different grouping compared to the previous stable release (2.1.8).

In addition, since there's no assembly info, the SDK doesn't exclude the known frames that are not part of the user code.
To conclude, the current SentryStackTraceFactory produces richer information when running on mono compared to MonoStackTraceFactory, but it's only lacking the additional info for symbolicate the frames. But since this code wasn't implemented on the server-side, I feel like it's better, for now, to stick to the current StackTraceFactory for mono and in the future, once the symbolization code is ready, enhance the current StackTrace with the needed information.

### DIFF
--- a/src/Sentry/SentryOptions.cs
+++ b/src/Sentry/SentryOptions.cs
@@ -436,10 +436,7 @@ namespace Sentry
                 () => ExceptionProcessors ?? Enumerable.Empty<ISentryEventExceptionProcessor>()
             };
 
-            SentryStackTraceFactory = Runtime.Current.IsMono()
-                // Also true for IL2CPP
-                ? new MonoSentryStackTraceFactory(this)
-                : new SentryStackTraceFactory(this);
+            SentryStackTraceFactory = new SentryStackTraceFactory(this);
 
             ISentryStackTraceFactory SentryStackTraceFactoryAccessor() => SentryStackTraceFactory;
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## Unreleased
+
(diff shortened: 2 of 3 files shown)


bruno-garcia:
> generating a different grouping compared to the previous stable release (2.1.8).

Shouldn't affect grouping though. Did you confirm it does?
It just doesn't show the assembly info UI.

bruno-garcia:
As we agreed on Discord, please just don't enabled it on `IsMono`?

The code that's there will likely still be needed on IL2CPP or some other Mono-like environment where the `new StackTrace(.., true)` won't work as expected.

src/Sentry/SentryOptions.cs
```diff
@@ -436,10 +436,7 @@ public SentryOptions()
                 () => ExceptionProcessors ?? Enumerable.Empty<ISentryEventExceptionProcessor>()
             };
 
-            SentryStackTraceFactory = Runtime.Current.IsMono()
-                // Also true for IL2CPP
-                ? new MonoSentryStackTraceFactory(this)
-                : new SentryStackTraceFactory(this);
+            SentryStackTraceFactory = new SentryStackTraceFactory(this);
```
bruno-garcia, line 439: This is the only change we need actually. Just don't add it by default on Mono but keep the code around for now until we're sure we won't need this for Unity

bruno-garcia (changes requested):
This is the only change we need actually. Just don't add it by default on Mono but keep the code around for now until we're sure we won't need this for Unity

lucas-zimerman:
Fixxd

bruno-garcia (approved):
Thanks @lucas-zimerman

Issue #716, opened 2021-01-04
Performance Monitoring ASP.NET Core Auto instrumentation


bruno-garcia:
This is a follow up of #715.

The goal is to ship GA of ASP.NET Core integration for performance monitoring.

The scope is requests into the process automatically capture a transaction to Sentry with timing information for the overall request. Additional spans will be added in follow up issues.

bruno-garcia:
@kanadaj raised some points about the current instrumentation on ASP.NET Core. We could add a second middleware just for performance monitoring btw, and inject it at the right place (after UseRouting or UseMVC are called). [Seems Andrew Lock blogged about this](https://andrewlock.net/inserting-middleware-between-userouting-and-useendpoints-as-a-library-author-part-1/).

We need to be able to add our middleware to start the transaction as early as possible but only once the route template exists. Addiing the transaction before that would require a mutable transaction and the sampling decision callback which runs when the transaction starts wouldn't have that information, which is the most important (I want to sample out `/health-check` for example).

kanadaj:
Small correction, should be either after `UseRouting` or `UseEndpoints`.

Additionally, legacy health checks don't need routing, so they may be inserted before `UseRouting`.

On the topic of inserting a middleware with `IStartupFilter`, I propose to either add a string option for the full name of the middleware to insert the middleware after, a bool predicate for the same purpose, or a simple bool to decide whether the middleware is inserted after `UseRouting`/`UseEndpoints` or at the very start of the pipeline in case of a custom middleware pipeline.

Tyrrrz:
@bruno-garcia 

>The scope is requests into the process automatically capture a transaction to Sentry with timing information for the overall request. Additional spans will be added in follow up issues.

This seems to be already done, except for additional spans. How should we proceed?

bruno-garcia:
It's out on 3.0.0 in preview, will GA soon y'all 🥳 

Thanks @Tyrrrz for all the work!

PR #717, opened 2021-01-04
fix: ASP.NET Core route template as transaction name

kanadaj:
…extracted

### DIFF
--- a/src/Sentry.AspNetCore/SentryMiddleware.cs
+++ b/src/Sentry.AspNetCore/SentryMiddleware.cs
@@ -109,7 +109,7 @@ namespace Sentry.AspNetCore
                 });
 
                 var transaction = hub.CreateTransaction(
-                    context.GetTransactionName(),
+                    context.Request.Path,
                     "http.server"
                 );
 
@@ -132,6 +132,7 @@ namespace Sentry.AspNetCore
                 }
                 finally
                 {
+                    transaction.Name = context.GetTransactionName();
                     transaction.Finish(
                         GetSpanStatusFromCode(context.Response.StatusCode)
                     );
--- a/src/Sentry.AspNetCore/SentryMiddleware.cs
+++ b/src/Sentry.AspNetCore/SentryMiddleware.cs
@@ -132,10 +132,13 @@ namespace Sentry.AspNetCore
                 }
                 finally
                 {
-                    transaction.Name = context.GetTransactionName();
-                    transaction.Finish(
-                        GetSpanStatusFromCode(context.Response.StatusCode)
-                    );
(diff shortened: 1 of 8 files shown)


src/Sentry/SentryOptions.cs
```diff
@@ -423,6 +423,11 @@ public double TraceSampleRate
         /// </summary>
         public ISentryTraceSampler? TraceSampler { get; set; }
 
+        /// <summary>
+        /// When set to true, Sentry captures transactions from middlewares even if the request has no legacy or Endpoint Routing features
+        /// </summary>
+        public bool TraceRequestsWithoutRouting { get; set; } = false;
```
bruno-garcia, line 429: Sounds like this belong to `SentryAspNetCoreOptions`?

src/Sentry.AspNetCore/SentryMiddleware.cs
```diff
@@ -132,9 +132,13 @@ public async Task InvokeAsync(HttpContext context)
                 }
                 finally
                 {
-                    transaction.Finish(
-                        GetSpanStatusFromCode(context.Response.StatusCode)
-                    );
+                    transaction.Name = context.GetTransactionName(out var routeTemplateExists);
+                    if (routeTemplateExists || _options.TraceRequestsWithoutRouting)
```
bruno-garcia, line 136: Sounds like the we should use ` _options.TraceRequestsWithoutRouting` when doing the sampling decision and mark the transaction as unsampled. Because this means the transaction will be created, outgoing HTTP requests will propagate it, but at the end we won't capture it here.

src/Sentry/SentryOptions.cs
kanadaj: Ohh oops

src/Sentry.AspNetCore/SentryMiddleware.cs
kanadaj: @bruno-garcia We cannot, because we don't have the name of the route until after the request

bruno-garcia: Discussed here: https://github.com/getsentry/sentry-dotnet/issues/716
We likely need a new middleware that will be added after the current one (once a route template exists). There we can start a transaction. but the error tracking part stays where it is.

bruno-garcia:
@kanadaj given your branch is gone, it's safe to assume this is abandoned? Let us know if there's something to be done here

kanadaj:
@bruno-garcia can't even remember what this does anymore or whether the change was implemented in another PR, I'd have to check the current code. This PR had some weird mixed stuff as well, likely by accident

PR #729, opened 2021-01-07
Adds to the existing User Other dict rather than replacing

brettjenkins:
@event.User.Other will never be null here, so haven't put in a null ref check as was suggested in the issue

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Unreleased
 
 - Disabled Mono StackTrace Factory. (#709) @lucas-zimerman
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 ## Unreleased
 
 - Disabled Mono StackTrace Factory. (#709) @lucas-zimerman
+- Adds to the existing User Other dict rather than replacing (#718)(#729) @brettjenkins
```
```suggestion
- Adds to the existing User Other dict rather than replacing (#729) @brettjenkins
```

bruno-garcia (approved):
Thanks @brettjenkins !

PR #731, opened 2021-01-09
lock file for nuget

bruno-garcia:
A year and a half later: #247

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 - Disabled Mono StackTrace Factory. (#709) @lucas-zimerman
 - Adds to the existing User Other dict rather than replacing (#729) @brettjenkins
+- Reproducible builds: lock file for nuget (#731) @bruno-garcia
 
 ## 3.0.0-alpha.9
 
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -8,6 +8,7 @@
     <SignAssembly>true</SignAssembly>
     <Deterministic>true</Deterministic>
     <Features>strict</Features>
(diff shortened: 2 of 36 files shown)


bruno-garcia:
image
always on Windows it break, why

bruno-garcia:
Still unanswered: https://stackoverflow.com/questions/57161512/nuget-lock-file-fails-to-restore-with-locked-mode

Tyrrrz:
What are we trying to achieve with this?

bruno-garcia:
@Tyrrrz lock the version of transitive dependencies so we always restore everything at the same version

Tyrrrz:
@bruno-garcia is the current behavior causing issues? Just curious because I've never had any issues that would be solved by a NuGet lock file.

bruno-garcia:
@Tyrrrz Not raising this as a reaction to a bug. Just want to guarantee reproducible builds to avoid any potential issues that arise from dependencies releasing new versions and further builds changing the dependency dependency tree. 

Some downsides to this I'm no aware of?

Tyrrrz:
> Some downsides to this I'm no aware of?

Well, committing a huge file to VCS for one ;)

bruno-garcia:
@Tyrrrz I thought you loved nodejs

Tyrrrz:
@bruno-garcia Not necessarily. NPM lockfiles need to be committed (which I absolutely hate) because NPM package resolution strategy is rather permissive when it comes to different versions (by default it's latest minor). That said, even if you run `npm ci` instead of `npm i`, it will always resolve to the exact versions found in `package.json`.

NuGet, on the other hand, never had that retarded design decision in the first place, so it's safe from such issues. 🙂 Thus, I don't really see a point in lockfiles here.

PR #732, opened 2021-01-10
Add Ben.Demystifier

bruno-garcia:
TODO:

- [x] Add Options enum: `StackTraceFormat` with values: `Original` (which should be the value before #662), `IncludeParameters` for the current approach (which will break current grouping), and `BenDemystifier` which will be default on 3.0.0
- [x] Check flag to fallback to the current stacktrace strategy
- [ ] Test on Unity (specially IL2CPP)
- [x] Add `patch` to make all types internal
- [x] Add `path` back and figure out the UI
- [x] 3.0.0 announcement draft: big bold statement that grouping will be affected, and this is an **opt-out**

# Before (minus file path)
image
image

# After 
* need to change the UI not to show path like that
* Arguments should be hidden and display on a click (similar to C++)

image


src/Directory.Build.props
```diff
@@ -47,7 +47,7 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.SourceLink.GitHub" Version="1.0.0" PrivateAssets="All" />
+    <PackageReference Include="Microsoft.SourceLink.GitHub" Version="1.1.0-beta-20204-02" PrivateAssets="All" />
```
bruno-garcia, line 50: Needed to be able to use relative git submodule (use url from checkout)

bruno-garcia:
This stuff could be refactored now to be honest. The abstractions added today was done by a contributor who wanted to add Ben.Demystifier and it's an odd mix of `virtual` methods and protected "internal":
```
        protected virtual StackTrace CreateStackTrace(Exception? exception)
        public virtual SentryStackTrace? Create(Exception? exception = null)
        protected virtual SentryStackFrame CreateFrame(StackFrame stackFrame, bool isCurrentStackTrace) 
        protected SentryStackFrame InternalCreateFrame(StackFrame stackFrame, bool demangle)
```

I've added checks for the "mode" at different locations. We might be better off refactoring in a way the mode is decided once and the right type is instantiated. Such type would override what makes sense (or use composition instead) for the default, shared behavior.

src/Sentry/Extensibility/SentryStackTraceFactory.cs
```diff
@@ -79,8 +79,17 @@ public class SentryStackTraceFactory : ISentryStackTraceFactory
...
-            var frames = stackTrace.GetFrames();
-            if (frames == null)
+            var frames = _options.StackTraceMode switch
+            {
+                StackTraceMode.Enhanced => EnhancedStackTrace.GetFrames(stackTrace).Select(p => p as StackFrame),
+                _ => stackTrace.GetFrames()
+// error CS8619: Nullability of reference types in value of type 'StackFrame?[]' doesn't match target type 'IEnumerable<StackFrame>'.
+#if NETCOREAPP3_0
+                                            .Where(f => f is not null)
+#endif
```
Tyrrrz, line 89: If we already have NET5.0 target, we can probably switch NET3.0 to annotations mode

src/Sentry/Extensibility/SentryStackTraceFactory.cs
```diff
@@ -132,17 +141,36 @@ protected SentryStackFrame InternalCreateFrame(StackFrame stackFrame, bool deman
...
+                    frame.InApp = false;
+                }
             }
 
-            frame.InApp = !IsSystemModuleName(frame.Module);
-            frame.FileName = stackFrame.GetFileName();
+            frame.InApp ??= !IsSystemModuleName(frame.Module);
+
+            // TODO: Uncomment before merging
+            // frame.FileName = stackFrame.GetFileName();
```
Tyrrrz, line 173: TODO ^

src/Sentry/StackTraceMode.cs
```diff
@@ -0,0 +1,30 @@
...
+        /// </summary>
+        AddParameters,
+        /// <summary>
+        /// Includes return type, arguments ref modifiers and more.
+        /// </summary>
+        /// <remarks>
+        /// This mode uses <see href="https://github.com/getsentry/Ben.Demystifier">Ben Adams' Demystifier library</see>.
+        /// </remarks>
+        Enhanced,
+    }
```
Tyrrrz, line 29: Do we really need both `AddParameters` and `Enhanced`?

Tyrrrz:
I have limited experience working with git submodules and all of it was horrible, so I'm a bit cautious of doing that. Maybe we can do a source package, even if Ben doesn't want to maintain it we can do it ourselves?

bruno-garcia:
@Tyrrrz I've had a bit of experience lately and it's working really well. One advantage is that you can edit code in there and use it. While the source package would mean fiddling with nuget packages in development. That said the nuget approach feels cleaner.

lucas-zimerman:
How well it works with .NET AOT? (Thinking about iOS right now).
If needed I could test on UWP .NET Native and so how it goes.

bruno-garcia:
No idea how it works on any AOT yet. `#Help wanted`

src/Sentry/Extensibility/SentryStackTraceFactory.cs
bruno-garcia: Arent' we getting the error because on 3.0 it was indeed annotated with nullable (`'StackFrame?[]`) so we need to compile in that `.Where(f => f is not null)` or am I missing something

Tyrrrz: Was it incorrectly annotated? Or it actually returns nulls on 3.0?

bruno-garcia: I just assumed it was rightly annotated, since it was .. annotated.
Since this can change between versions I rather we have the warnings turned on on both? I assume they should match 99% of the cases, we #ifdef when they don't to be sure the compiled thing accounts for it. 

I would be surprised if it went from non-nullable to nullable. But the other way around would be OK. So a nca3.1 DLL won't crash on net5 because didn't account for null, for example

src/Sentry/Extensibility/SentryStackTraceFactory.cs
```diff
@@ -128,20 +138,35 @@ protected SentryStackFrame InternalCreateFrame(StackFrame stackFrame, bool deman
...
+                if (_options.StackTraceMode == StackTraceMode.Enhanced && stackFrame is EnhancedStackFrame enhancedStackFrame)
+                {
+                    var sb = new StringBuilder();
+                    frame.Function = enhancedStackFrame.MethodInfo.Append(sb, false).ToString();
+
+                    if (enhancedStackFrame.MethodInfo.DeclaringType is { } declaringType)
+                    {
+                        sb.Clear();
+                        sb.AppendTypeDisplayName(declaringType);
+                        frame.Module = sb.ToString();
```
bruno-garcia, line 153: When I set `Module` and `Package` Sentry seems to ignore `Module` completely.

@matejminar maybe can chime in here? We're doing some changes in the .NET stack traces and it might need a bit of help on the UI to properly render.

matejminar: Sure, I'll gladly help.

Issue #663, opened 2020-12-18
Performance API follow up


bruno-garcia:
I've tried to add some docs, and in the process a few things came up.

Docs PR: https://github.com/getsentry/sentry-docs/pull/2790

Notes:
* StartTransaction vs CreateTransaction
Sentry's SDKs use `startTransaction` but .NET has `CreateTransaction`

Java requires only "name" and .NET requires "name" and "op".
Code snippets in Java set an Op right after creating it, should it be in the Start method instead?

Java has `Sentry.getSpan();` to get the current "span bound to the Hub". How do we do this in .NET?
Is this in JavaScript? If not, how do we do it there?

// C#:
`var innerSpan = span.StartChild("task");`

// Java
`Span innerSpan = span.startChild("task", item.getName());`

How do I set the span name in C#? Only operation in the method signature.

```
var transaction = SentrySdk.CreateTransaction("ProcessOrderBatch()", "batch");
try 
{
    procesOrderBatch();
    transaction.Finish();
}
catch (Exception e)
{
    // How do I set the exception? Or relate to an error event?
    // Java: transaction.setThrowable(e);
    transaction.Finish(SpanStatus.InternalError);

	// this
    SentrySdk.CaptureException(e); // How is the transaction and error linked?
	// or
	throw; // I expect the request is logged upstream so the reference to the Exception must have been given to the transaction earlier in this catch block
}
```
 
In Java we set the exception to the transaction, which is kept in a WeakRef so we can corelate the error event to the transaction.
How do we do this in C#?   

Java has:

```java
ISpan span = Sentry.getSpan();
if (span == null) {
    span = Sentry.startTransaction("task");
} else {
    span = span.startChild();
}
```

How do we do this? Should we change Java to something else. I believe Go doesn't distinct between StartTran or StartSpan (thinking in the future with single span ingestion, does it matter?)

Tyrrrz:
@bruno-garcia

> Java requires only "name" and .NET requires "name" and "op".

What should be the default value for `op` be? Also, in some SDKs the entire object is mutable, should we do the same thing here as well?

> Java has Sentry.getSpan(); to get the current "span bound to the Hub". How do we do this in .NET?

What does a "current span" mean? Last created transaction?

> How do I set the span name in C#? Only operation in the method signature.

Do spans actually have names? From what I know, they don't.

> // How do I set the exception? Or relate to an error event?
> // Java: transaction.setThrowable(e);

[...]

PR #737, opened 2021-01-14
Separate tracing middleware

### LINKED ISSUE #708: Performance doesn't merge MVC requests based on routes

- [ ] ASP.NET MVC
- [ ] ASP.NET Web API (OWIN)
- [X] ASP.NET Core
- [ ] WPF
- [ ] WinForms
- [ ] Xamarin
- [ ] Other:

- [ ] .NET Framework
- [ ] Mono
- [X] .NET Core

* Version: 5.0

- [ ] Sentry
- [X] Sentry.Serilog
- [ ] Sentry.NLog
- [ ] Sentry.Log4Net
- [X] Sentry.Extensions.Logging
- [X] Sentry.AspNetCore

* Version: 3.0.0-alpha9

Sentry doesn't seem to merge routes when used with ASP.NET MVC and configured inside Program.cs:

Configuration code:

```cs
        public static IHostBuilder CreateWebHostBuilder(string[] args)
        {
        Host.CreateDefaultBuilder(args)
              .ConfigureWebHostDefaults(webBuilder =>
              {
                  webBuilder.ConfigureKestrel(options => { options.AllowSynchronousIO = true; })
                      .UseSentry(ConfigureSentry)
                      .UseStartup<Startup>();
              });
        }

        private static void ConfigureSentry(SentryAspNetCoreOptions x)
        {
            x.SendDefaultPii = true;
            x.MaxRequestBodySize = RequestSize.Always;
            var assembly = Assembly.GetAssembly(typeof(Program));
            x.Release = assembly.GetVersion();
            x.AddEventProcessor(new OperationCanceledSentryFilter());
            x.AddEventProcessor(new FixHeaderDelimiterSentryEventProcessor());
            
            x.AddInAppExclude("MySql.Data.");
            x.AddInAppExclude("MySqlConnector.");
            x.AddInAppExclude("Pomelo.");
            x.AddInAppExclude("ElasticApm.");
            x.AddInAppExclude("Hangfire");
            x.AddInAppExclude("Serilog");
            x.AddInAppExclude("App.Metrics.");

            x.TraceSampleRate = 0.1;

            #if DEBUG
            x.TraceSampleRate = 1;
            #endif
            
            x.AddLogEntryFilter((category, level, eventId, exception) => eventId.ToString() == "Microsoft.EntityFrameworkCore.Infrastructure.SensitiveDataLoggingEnabledWarning" && string.Equals(category, "Microsoft.EntityFrameworkCore.Model.Validation", StringComparison.Ordinal));

            x.AddLogEntryFilter((category, level, eventId, exception) => eventId.Id == 10000 && string.Equals(eventId.Name, "Microsoft.EntityFrameworkCore.Update.SaveChangesFailed", StringComparison.Ordinal) && string.Equals(category, "Microsoft.EntityFrameworkCore.Database.Command", StringComparison.Ordinal));
        }
```

Tested with both legacy and endpoint routing; the route in particular is set up with attributes:
```cs
        [Route("Fiction/{id:int}/{slug}", Order = 0)]
        [Route("Fiction/{id:int}", Order = 1)]
```

Additionally, static files are also getting traced, which isn't ideal.

Tyrrrz:
Separated tracing into its own middleware.

The user needs to add `UseSentryTracing()` after `UseRouting()` so that it has access to routing. If they add it before, it will instead use the request path for transaction name, which is still better than not working at all.

Closes #708

### DIFF
--- a/samples/Sentry.Samples.AspNetCore.Basic/Program.cs
+++ b/samples/Sentry.Samples.AspNetCore.Basic/Program.cs
@@ -5,6 +5,7 @@ using Microsoft.AspNetCore.Hosting;
 using Microsoft.AspNetCore.Routing;
 using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.DependencyInjection;
+using Sentry.AspNetCore;
 
 namespace Sentry.Samples.AspNetCore.Basic
 {
@@ -32,6 +33,10 @@ namespace Sentry.Samples.AspNetCore.Basic
                     // An example ASP.NET Core middleware that throws an
                     // exception when serving a request to path: /throw
                     app.UseRouting();
+
+                    // Enable Sentry performance monitoring
+                    app.UseSentryTracing();
+
                     app.UseEndpoints(endpoints =>
                     {
                         // Reported events will be grouped by route pattern
--- a/src/Sentry.AspNetCore/SentryTracingMiddleware.cs
+++ b/src/Sentry.AspNetCore/SentryTracingMiddleware.cs
@@ -1,0 +1,88 @@
+﻿using System;
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Builder;
+using Microsoft.AspNetCore.Http;
+using Sentry.AspNetCore.Extensions;
+using Sentry.Protocol;
+
+namespace Sentry.AspNetCore
+{
+    /// <summary>
+    /// Sentry tracing middleware for ASP.NET Core
+    /// </summary>
+    public class SentryTracingMiddleware
+    {
+        private readonly RequestDelegate _next;
+        private readonly Func<IHub> _hubAccessor;
+
+        public SentryTracingMiddleware(RequestDelegate next, Func<IHub> hubAccessor)
+        {
+            _next = next;
+            _hubAccessor = hubAccessor;
+        }
+
+        /// <summary>
+        /// Handles the <see cref="HttpContext"/>.
+        /// </summary>
+        public async Task InvokeAsync(HttpContext context)
+        {
+            var hub = _hubAccessor();
+            if (!hub.IsEnabled)
+            {
+                await _next(context).ConfigureAwait(false);
+                return;
+            }
+
+            var transaction = hub.CreateTransaction(
+                context.GetTransactionName(),
+                "http.server"
+            );
+
+            try
+            {
+                await _next(context).ConfigureAwait(false);
+            }
+            finally
+            {
+                transaction.Finish(
+                    GetSpanStatusFromCode(context.Response.StatusCode)
+                );
+            }
+        }
+
+        private static SpanStatus GetSpanStatusFromCode(int statusCode) => statusCode switch
+        {
+            < 400 => SpanStatus.Ok,
+            400 => SpanStatus.InvalidArgument,
+            401 => SpanStatus.Unauthenticated,
+            403 => SpanStatus.PermissionDenied,
+            404 => SpanStatus.NotFound,
+            409 => SpanStatus.AlreadyExists,
+            429 => SpanStatus.ResourceExhausted,
+            499 => SpanStatus.Cancelled,
+            < 500 => SpanStatus.InvalidArgument,
+            500 => SpanStatus.InternalError,
+            501 => SpanStatus.Unimplemented,
+            503 => SpanStatus.Unavailable,
+            504 => SpanStatus.DeadlineExceeded,
+            < 600 => SpanStatus.InternalError,
+            _ => SpanStatus.UnknownError
+        };
+    }
+
+    /// <summary>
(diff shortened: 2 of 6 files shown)


Tyrrrz:
Note: I don't think we can get this working with `UseMvc()` because it doesn't provide us a slot where we can insert our middleware which would be after routing but before controller (which is exactly why `UseRouting()` was added). My personal opinion is that it's fine and users should be advised to migrate towards `UseRouting()` to make use of tracing.

kanadaj:
For `UseMvc()` the only way to do it would be to collect route info post-execution.

That is still not completely terrible though, since with a middleware, it's possible to use custom sampling logic by using `UseWhen()` or `MapWhen()`

samples/Sentry.Samples.AspNetCore.Basic/Program.cs
```diff
@@ -32,6 +33,10 @@ public static void Main(string[] args)
                     // An example ASP.NET Core middleware that throws an
                     // exception when serving a request to path: /throw
                     app.UseRouting();
+
+                    // Enable Sentry performance monitoring
+                    app.UseSentryTracing();
```
bruno-garcia, line 38: What about `UseSentryTransaction()`? Is it `tracing` that we're doing?
Tracing to me sounds like "distributed tracing" which isn't really added with this. We don't even have it yet and needs to patch outgoing HTTP requests and read incoming HTTP requests (which we can through this hook indeed).

Requested some comments: https://discord.com/channels/621778831602221064/621792783316942878/799409049778913352

src/Sentry.AspNetCore/SentryTracingMiddleware.cs
```diff
@@ -0,0 +1,88 @@
...
+            var transaction = hub.CreateTransaction(
+                context.GetTransactionName(),
+                "http.server"
+            );
+
+            try
+            {
+                await _next(context).ConfigureAwait(false);
+            }
+            finally
```
bruno-garcia, line 45: If an exception bubbles out of there, is the error in Sentry connected to this transaction?
Sounds like we're missing this connection (the error event should get the trace context data).

It was done in Java recently: https://github.com/getsentry/sentry-java/pull/1156

Raised an issue so we can track this on a follow up PR: https://github.com/getsentry/sentry-dotnet/issues/738

bruno-garcia (commented):
Just raised a discussion about the name of the middleware but other than that I think this is much better than the original approach.

One point though, I'm worried about `UseMvc`. Just telling people "tough luck" isn't really a good approach. If we go that route we need to at least be really clear on the docs that UseRouting MUST be used and what alternatives to use if those can't use it.

samples/Sentry.Samples.AspNetCore.Basic/Program.cs
kanadaj: `UseSentryPerformance()`?

Tyrrrz: I named it 'tracing' because Sentry docs often refer to it as tracing: https://develop.sentry.dev/sdk/unified-api/tracing/

I don't really like `UseSentryPerformance()` and `UseSentryTransaction()` tbh. Maybe `UseSentryPerformanceMonitoring()`, but that's too long.

src/Sentry.AspNetCore/SentryTracingMiddleware.cs
Tyrrrz: No, we don't support this currently. I also mentioned it here: https://github.com/getsentry/sentry-dotnet/issues/663#issuecomment-760227782

samples/Sentry.Samples.AspNetCore.Basic/Program.cs
bruno-garcia: Agree `SentryPerformance` isn't ideal. "Does it make my app faster?". `PerformanceMonitoring` does what it says but, too long also agree.

`SentryTransaction` does what it says and it isnt' too long. The docs you linked I have a PR renaming already. Remember you didn't even find it because you searched for performance?

I guess we keep it SentryTracing for now

bruno-garcia (approved):

samples/Sentry.Samples.AspNetCore.Basic/Program.cs
Tyrrrz: > SentryTransaction does what it says and it isnt' too long. The docs you linked I have a PR renaming already. Remember you didn't even find it because you searched for performance?

Didn't realize they were renamed. The reason I'm not too keen on `SentryTransaction` is because it doesn't really convey to me what it does. A transaction is a pretty generic concept that can mean different things. I'm not sure that if I were a Sentry user and saw `SentryTransaction` I would think about performance monitoring.

But of course we can rename it if you are confident that's the way to go.

PR #740, opened 2021-01-15
Sync transaction and transaction name on scope

Tyrrrz:
Currently, `Scope.Transaction` and `Scope.TransactionName` are completely separate. This is apparently incorrect, so this PR addresses that.

For context, check the discussion on this issue: https://github.com/getsentry/develop/issues/246

Note: still pending an answer to https://github.com/getsentry/develop/issues/246#issuecomment-761172858

### DIFF
--- a/src/Sentry/Scope.cs
+++ b/src/Sentry/Scope.cs
@@ -112,14 +112,12 @@ namespace Sentry
         /// <inheritdoc />
         public string? Environment { get; set; }
 
-        // TransactionName and Transaction need to coexist because
-        // SentryEvent still tracks the transaction as a string.
-        // If a started Transaction exists and the user sets TransactionName,
-        // it should update the name of that transaction.
-        // Otherwise, it should just keep track of the name separately, so that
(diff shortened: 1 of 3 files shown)


src/Sentry/Scope.cs
```diff
@@ -112,8 +112,33 @@ public User User
...
+        // Otherwise, it should just keep track of the name separately, so that
+        // it can be used in an event, in case a transaction doesn't exist.
+        // More context in this discussion:
+        // https://github.com/getsentry/develop/issues/246#issuecomment-760997974
+
+        private string? _fallbackTransactionName;
+
         /// <inheritdoc />
-        public string? TransactionName { get; set; }
+        public string? TransactionName
```
bruno-garcia, line 127: I know you're waiting for a reply from @rhcarvalho on that thread but if we're going ahead with this, there are a few test cases here right?

bruno-garcia (commented):
We all agree this is suboptimal and a result of adding performance into the error tracking SDK but if we all agree this will be the solution, we just need some tests here

PR #741, opened 2021-01-15
Bring Transaction a bit more inline with the design guidelines

Tyrrrz:
Some changes to make the performance implementation more similar to Java's.

Open questions:

- ~Do we need `scope.Transaction` to be public? It is public in Java: https://github.com/getsentry/sentry-java/blob/681c75963466a55a45709fe9d475b2d64b36fe09/sentry/src/main/java/io/sentry/Scope.java#L135-L137~
- ~Do we want to do `DisabledSpan` & `DisabledTransaction`, similarly to Java?~
- ~Is there a default value for `op` or can it just be `null`?~


src/Sentry/Extensibility/DisabledHub.cs
```diff
@@ -55,7 +55,7 @@ public void WithScope(Action<Scope> scopeCallback)
         /// <summary>
         /// Returns a dummy transaction.
         /// </summary>
-        public Transaction CreateTransaction(string name, string operation) => new(this);
+        public Transaction StartTransaction() => new(this);
```
Tyrrrz, line 58: Should we have `ITransaction` and return some kind of `DisabledTransaction` here? That's how Java does this, but I personally don't think it's warranted.

src/Sentry/Protocol/ISpanContext.cs
```diff
@@ -26,12 +26,12 @@ public interface ISpanContext
         /// <summary>
         /// Operation.
         /// </summary>
-        string Operation { get; }
+        string Operation { get; set; }
```
Tyrrrz, line 29: Should this be nullable? Or have some sort of default value?

src/Sentry/Extensibility/HubAdapter.cs
```diff
@@ -75,8 +75,8 @@ public void WithScope(Action<Scope> scopeCallback)
         /// Forwards the call to <see cref="SentrySdk"/>.
         /// </summary>
         [DebuggerStepThrough]
-        public Transaction CreateTransaction(string name, string operation)
-            => SentrySdk.CreateTransaction(name, operation);
+        public Transaction StartTransaction()
```
Tyrrrz, line 78: Note to myself: check if creating transaction without a name is ok

src/Sentry/SentryClient.cs
```diff
@@ -126,10 +126,17 @@ public void CaptureTransaction(Transaction transaction)
                 return;
             }
 
-            if (_options.TraceSampleRate < 1)
+            if (transaction.IsSampled)
             {
-                if (Random.NextDouble() > _options.TraceSampleRate)
+                _options.DiagnosticLogger?.LogDebug("Transaction dropped due to sampling.");
+                return;
+            }
```
Tyrrrz, line 133: The `if` condition needs to be inverted

src/Sentry/SentryClient.cs
```diff
@@ -126,10 +126,17 @@ public void CaptureTransaction(Transaction transaction)
                 return;
             }
 
-            if (_options.TraceSampleRate < 1)
+            if (transaction.IsSampled)
```
Tyrrrz, line 129:
```suggestion
            if (!transaction.IsSampled)
```

src/Sentry/Extensibility/DisabledHub.cs
bruno-garcia: In other SDKs, Python comes to mind (/cc @untitaker ) they decided to create a transaction even if it's disabled to have a trace_id in the scope and allow errors to be linked by trace_id. But that means we're paying for the allocations and the fact instrumentations will run (since there's a transaction on the scope). 

I'm unsure what's the best way here but I lean towards the NoOp approach and if we want to correlate things we could just add the id as a tag or another approach.

src/Sentry/Protocol/ISpanContext.cs
bruno-garcia: What's the use case for not having an operation? If we make it non-nullable are going to force users to set gibberish because they often have nothing to set?

Gut feeling is that we should always have an `operation` otherwise we're likely instrumenting something badly. Can we have it non-null (readonly also is not an option?) and we can change it later.

src/Sentry/Extensibility/HubAdapter.cs
bruno-garcia: I don't believe Sentry is OK with a transaction without name. Does it make sense to have one? How do we group them in the UI?

bruno-garcia:
> Do we need scope.Transaction to be public? It is public in Java: getsentry/sentry-java@681c759/sentry/src/main/java/io/sentry/Scope.java#L135-L137

Originally that was the `string` transaction, right? This is supposed to be a unfied solution I assume since it was part of the SDK spec. If we're dropping it or moving to the complex type Transaction (or Event).

> Do we want to do DisabledSpan & DisabledTransaction, similarly to Java?

It's either that or returning null which now with NRT isnt' that bad anymore. Probably better to go with NoOp for now (disabled) to stay aligned.

> Is there a default value for op or can it just be null?

What if we make it mandatory to provide an operation? We can always release an update with an overload, the other way around is a breaking change.

src/Sentry/Extensibility/HubAdapter.cs
untitaker: transaction without name cannot be used by product at all

src/Sentry/Protocol/ISpanContext.cs
untitaker: if you don't have an op you're saying you don't know what you're instrumenting. That can't be right

src/Sentry/Extensibility/DisabledHub.cs
untitaker: the main point of having noop transactions was that

1. you can start a child span without worrying whether you're in a transaction (the span will just be discarded)
2. you can do error correlation without enabling transaction sending

Tyrrrz: @bruno-garcia the challenge here is that we need to decide now, otherwise changing `ITransaction` to `Transaction` is technically a breaking change (albeit I would argue highly unlikely to actually break anyone)

Tyrrrz:
> Originally that was the string transaction, right? This is supposed to be a unfied solution I assume since it was part of the SDK spec. If we're dropping it or moving to the complex type Transaction (or Event).

Yes, originally it was string. My question is more of a, do we need to get or set `Transaction` through a property on scope? Or should we just direct users towards `StartTransaction` and `GetSpan` (not implemented yet) instead?

bruno-garcia:
> > Originally that was the string transaction, right? This is supposed to be a unfied solution I assume since it was part of the SDK spec. If we're dropping it or moving to the complex type Transaction (or Event).
> 
> Yes, originally it was string. My question is more of a, do we need to get or set `Transaction` through a property on scope? Or should we just direct users towards `StartTransaction` and `GetSpan` (not implemented yet) instead?

We need a way to have the transaction bound to the scope so integrations can find them later. Afaik this is how we do that.

In JS it's acutally `setSpan`:

`  hub.configureScope(scope => scope.setSpan(transaction));`

https://docs.sentry.io/platforms/javascript/guides/react/performance/capturing/manual/

Tyrrrz:
Spawned:

https://github.com/getsentry/sentry-dotnet/issues/752
https://github.com/getsentry/sentry-dotnet/issues/751
https://github.com/getsentry/sentry-dotnet/issues/749

src/Sentry/Protocol/Context/Trace.cs
```diff
@@ -23,28 +23,48 @@ public class Trace : ISpanContext, IJsonSerializable
         public SentryId TraceId { get; set; }
 
         /// <inheritdoc />
-        public string Operation { get; set; } = "unknown";
+        public string Operation { get; set; } = "";
```
Tyrrrz, line 26: Ideally this should be nullable, but it shares the same interface `ISpanContext` as `ISpan` and `ITransaction` :/

src/Sentry/Protocol/IHasExtra.cs
```diff
@@ -0,0 +1,20 @@
+﻿using System.Collections.Generic;
+
+namespace Sentry.Protocol
+{
+    /// <summary>
+    /// Implemented by objects that contain a map of untyped extra data.
+    /// </summary>
+    public interface IHasExtra
```
Tyrrrz, line 8: Separating these into these micro interfaces allows us to reuse the same extension methods across Event, Span, Transaction, and Scope.

Tyrrrz:
@bruno-garcia this is finally ready for review. The PR brings the design in line with the guidelines and Java SDK, fixes a bunch of bugs, adds some scary comments, and also tests. I tested to the best of my knowledge but I may have missed something. Let me know if I did.

The feature overall is still missing the following major things:
- Trace header propagation
- Exception on span
- Adaptive sampling
- scope.GetCurrentSpan()

src/Sentry/Protocol/IHasExtra.cs
Tyrrrz: Also note: the `extra` on a span is actually serialized as `data` in JSON (as required by protocol).

test/Sentry.Tests/SentryClientTests.cs
```diff
@@ -55,7 +54,7 @@ public void CaptureEvent_IdReturnedToString_NoDashes()
 
             var actual = sut.CaptureEvent(evt);
 
-            var hasDashes = actual.ToString().Contains('-');
+            var hasDashes = actual.ToString().Contains("-");
```
Tyrrrz, line 57: This previously used LINQ on net462 but not on netcore. Changed to string so it uses intrinsic string method in both cases.

src/Sentry/Protocol/Transaction.cs
```diff
@@ -55,14 +73,14 @@ public string Operation
...
-            private set => Contexts.Trace.Status = value;
+            set => Contexts.Trace.Status = value;
         }
 
         /// <inheritdoc />
         public bool IsSampled
         {
             get => Contexts.Trace.IsSampled;
-            set => Contexts.Trace.IsSampled = value;
+            internal set => Contexts.Trace.IsSampled = value;
```
bruno-garcia, line 83: `IsSampled` is nullable in Java

https://github.com/getsentry/sentry-java/blob/3c87a48442e536a41ed905d9e43bdc2f7b800176/sentry/src/main/java/io/sentry/SentryTraceHeader.java#L58

src/Sentry/Protocol/Transaction.cs
```diff
@@ -155,27 +195,32 @@ internal Transaction(IHub hub)
...
+        internal ISpan StartChild(SpanId parentSpanId, string operation)
         {
-            var span = new Span(_spanRecorder, null, SpanId, operation);
-            _spanRecorder.Add(span);
+            var span = new Span(this, parentSpanId, operation)
+            {
+                IsSampled = IsSampled
+            };
+
+            (_spans ??= new List<Span>()).Add(span);
```
bruno-garcia, line 205: Since `StartChild` can be called by multiple threads, this could be a problem

bruno-garcia (approved):
Lots of improvements! Great stuff!

src/Sentry/Protocol/Context/Trace.cs
```diff
@@ -69,16 +89,18 @@ public static Trace FromJson(JsonElement json)
             var spanId = json.GetPropertyOrNull("span_id")?.Pipe(SpanId.FromJson) ?? SpanId.Empty;
             var parentSpanId = json.GetPropertyOrNull("parent_span_id")?.Pipe(SpanId.FromJson);
             var traceId = json.GetPropertyOrNull("trace_id")?.Pipe(SentryId.FromJson) ?? SentryId.Empty;
-            var operation = json.GetPropertyOrNull("op")?.GetString() ?? "unknown";
+            var operation = json.GetPropertyOrNull("op")?.GetString() ?? "";
             var status = json.GetPropertyOrNull("status")?.GetString()?.Pipe(s => s.Replace("_", "").ParseEnum<SpanStatus>());
+            var isSampled = json.GetPropertyOrNull("sampled")?.GetBoolean() ?? true;
```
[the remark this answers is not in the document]
Tyrrrz: I will make it nullable in a follow up PR

src/Sentry/Protocol/Span.cs
```diff
@@ -31,51 +34,53 @@ public class Span : ISpan, IJsonSerializable
...
+        public ISpan StartChild(string operation) =>
+            _parentTransaction.StartChild(SpanId, operation);
 
         /// <inheritdoc />
-        public void Finish(SpanStatus status = SpanStatus.Ok)
-        {
-            EndTimestamp = DateTimeOffset.UtcNow;
-            Status = status;
-        }
+        public void Finish() => EndTimestamp = DateTimeOffset.UtcNow;
```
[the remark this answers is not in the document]
Tyrrrz: Nope, no issues so far.

src/Sentry/Protocol/Span.cs
```diff
@@ -144,12 +146,13 @@ public static Span FromJson(JsonElement json)
             var operation = json.GetPropertyOrNull("op")?.GetString() ?? "unknown";
             var description = json.GetPropertyOrNull("description")?.GetString();
             var status = json.GetPropertyOrNull("status")?.GetString()?.Pipe(s => s.Replace("_", "").ParseEnum<SpanStatus>());
-            var isSampled = json.GetPropertyOrNull("sampled")?.GetBoolean() ?? false;
+            var isSampled = json.GetPropertyOrNull("sampled")?.GetBoolean() ?? true;
```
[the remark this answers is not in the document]
Tyrrrz: > Spans have sampling decision in themselves?

Apparently they do: 
image

src/Sentry/Protocol/SpanId.cs
```diff
@@ -74,11 +92,22 @@ public static SpanId FromJson(JsonElement json)
         /// <summary>
         /// The <see cref="Guid"/> from the <see cref="SentryId"/>.
         /// </summary>
-        public static implicit operator string(SpanId id) => id._value;
-
+        public static implicit operator string(SpanId id) => id.ToString();
+
+        // Note: no implicit conversion from `string` to `SpanId` as that leads to serious bugs.
```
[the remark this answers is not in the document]
Tyrrrz: `<remarks>` is for the user, this is more for us.

src/Sentry/Protocol/Transaction.cs
Tyrrrz: can we make the list of spans an eagerly initialized `ConcurrentBag`? really don't want to deal with `Lazy<T>` or locks here.

bruno-garcia: Lets leave it out so far, we might just pull Transaction out of the Scope in favor of a list of Spans (that could then be concurrent). Lets discuss this later

PR #744, opened 2021-01-17
Ref: Add instruction_addr  to SentryStackFrame

lucas-zimerman:
https://develop.sentry.dev/sdk/event-payloads/stacktrace/
instruction_addr  was missing from SentryStackFrame and it's an important parameter for Native exceptions (allowing it to get symbolicated)

### DIFF
--- a/src/Sentry/Protocol/Exceptions/SentryStackFrame.cs
+++ b/src/Sentry/Protocol/Exceptions/SentryStackFrame.cs
@@ -1,6 +1,8 @@
 using System.Collections.Generic;
+using System.Globalization;
 using System.Linq;
 using System.Text.Json;
+using System.Text.RegularExpressions;
 using Sentry.Internal.Extensions;
 
 // ReSharper disable once CheckNamespace
@@ -116,6 +118,13 @@ namespace Sentry.Protocol
         /// We actually use the instruction address for symbolication but this can be used to calculate an instruction offset automatically.
         /// </summary>
         public long? SymbolAddress { get; set; }
+
+        /// <summary>
+        /// An optional instruction address for symbolication.<br/>
+        /// This should be a string with a hexadecimal number that includes a <b>0x</b> prefix.<br/>
+        /// If this is set and a known image is defined in the <see href="https://develop.sentry.dev/sdk/event-payloads/debugmeta/">Debug Meta Interface</see>, then symbolication can take place.<br/>
+        /// </summary>
+        public string? InstructionAddress { get; set; }
 
         /// <summary>
         /// The instruction offset.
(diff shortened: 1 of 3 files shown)


src/Sentry/Protocol/Exceptions/SentryStackFrame.cs
```diff
@@ -248,6 +257,13 @@ public void WriteTo(Utf8JsonWriter writer)
                 writer.WriteNumber("symbol_addr", symbolAddress);
             }
 
+            // Instruction address
+            if (InstructionAddress is { } instructionAddress &&
+                Regex.IsMatch(instructionAddress, @"\A\b0[xX][0-9a-fA-F]+\b\Z"))
```
bruno-garcia, line 262: if it doesn't match we just don't do anything? Should we match on the setter then?
Should this be a `long` instead?

src/Sentry/Protocol/Exceptions/SentryStackFrame.cs
```diff
@@ -248,6 +257,13 @@ public void WriteTo(Utf8JsonWriter writer)
...
+            {
+                writer.WriteString("instruction_addr", instructionAddress.ToLower());
+            }
```
bruno-garcia, line 265: I'd rather we dont' have such logic buried in the serialization code.

We can add that elsewhere, or nowhere really (let it hit the server and the error there will alert the user it was malformed. That will only affect users creating frames "by hand" and doing it wrong, not us. (Assuming we get it right :) )

Also, why `ToLower`? it'll allocate a string. Can we just make sure we get the value in its right format into the object?

```suggestion
            if (!string.IsNullOrWhiteSpace(instructionAddress))
            {
                writer.WriteString("instruction_addr", instructionAddress);
            }
```

bruno-garcia (changes requested):

src/Sentry/Protocol/Exceptions/SentryStackFrame.cs
lucas-zimerman: Alright, got rid of that code :D

src/Sentry/Protocol/Exceptions/SentryStackFrame.cs
```diff
@@ -248,6 +257,12 @@ public void WriteTo(Utf8JsonWriter writer)
                 writer.WriteNumber("symbol_addr", symbolAddress);
             }
 
+            // Instruction address
+            if (InstructionAddress is { } instructionAddress)
```
bruno-garcia, line 261: This doesn't check if it's `""` or `"  "`, why did you chose this vs string.IsNullOrEmpty?

bruno-garcia (approved):

PR #745, opened 2021-01-17
Only include latest release notes on NuGet

### LINKED ISSUE #742: Fix release notes from NuGet package
The whole release notes is added and nuget.org doesn't hide it under a toggle. 
It also has a character limit as it was pointed out on Discord.

Ideally only include the last release notes by parsing the content. But just pulling the last N lines is good enough.

image

Mitch528:
Resolves #742.

This PR uses an MSBuild task from https://github.com/sillsdev/SIL.BuildTasks. A caveat to using this would be that it follows the conventions of https://keepachangelog.com/en/1.0.0/.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## 3.0.0-alpha.11
 
+### Changed
+
 - Limit attachment size (#705)
 - Separate tracing middleware (#737)
 
@@ -45,7 +47,7 @@
 * Bump Microsoft.Bcl.AsyncInterfaces to 5.0.0 #618
 * Bump `Microsoft.Bcl.AsyncInterfaces` to 5.0.0 #618
 * `DefaultTags` moved from `SentryLoggingOptions` to `SentryOptions` (#637) @PureKrome
-* `Sentry.Serilog` can accept DefaultTags (#637) @PureKrome 
+* `Sentry.Serilog` can accept DefaultTags (#637) @PureKrome
 
 ## 3.0.0-alpha.5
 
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## 3.0.0-alpha.11
 
+### Changed
```
Tyrrrz, line 5: Is this required for the tool to work?

Mitch528: Yeah, it seems to require at least one subheading on the latest version, though it doesn't necessarily need to be that exact one.

CHANGELOG.md
```diff
@@ -104,9 +106,9 @@ We'd love to get feedback.
 
 ## 2.2.0-alpha
 
-Add nullable reference types support (Sentry, Sentry.Protocol) (#509) 
-fix: Use ASP.NET Core endpoint FQDN (#485) 
-feat: Add integration to TaskScheduler.UnobservedTaskException (#481) 
+Add nullable reference types support (Sentry, Sentry.Protocol) (#509)
```
bruno-garcia, line 109: I think you editor doesn't like the empty spaces, it changed all files you saved

CHANGELOG.md
bruno-garcia: It pulls the first subheading?

If so what if we had a first heading `#Changelog` and the subheading becomes the version?

bruno-garcia (commented):
Thanks for taking in this @Mitch528 !! ❤️ 

CI doesn't seem happy though

bruno-garcia:
OH nevermind it's Danger that fails, missing a changelog :D

CHANGELOG.md
Mitch528: @bruno-garcia The #Changelog part should be fine. It selects the version by using the regex `^## (.+)`. What I meant was that the latest version requires at least one heading within, for example:

```md
# Changelog
...

## 1.0.1

### Added:
- ...
- ...
```

CHANGELOG.md
Mitch528: oops, will change that back.

CHANGELOG.md
bruno-garcia: Got it. maybe a good reason to start breaking changes into ### Fixes, Feature, etc

bruno-garcia (approved):
🥳

bruno-garcia:
#skip-changelog

PR #753, opened 2021-01-19
Trace sampling

Tyrrrz:
Related: #663


src/Sentry/SentryClient.cs
```diff
@@ -122,24 +122,39 @@ public void CaptureTransaction(Transaction transaction)
...
-            // A transaction may have already been sampled somehow or the
-            // field may have been set directly. To be safe, we check that.
-            if (!transaction.IsSampled)
+            // If a transaction has been already sampled out (by adaptive sampling, at the time
+            // the transaction was created), then we just drop it here.
+            if (transaction.IsSampled == false)
             {
-                _options.DiagnosticLogger?.LogDebug("Transaction dropped due to sampling.");
+                _options.DiagnosticLogger?.LogDebug(
+                    "Transaction dropped due to adaptive sampling."
```
bruno-garcia, line 137: Was it due to adaptive sampling? Maybe better:

```suggestion
                    "Transaction dropped due to being marked as unsampled."
```

src/Sentry/SentryClient.cs
```diff
@@ -122,24 +122,39 @@ public void CaptureTransaction(Transaction transaction)
...
-            if (_options.TracesSampleRate < 1)
+            // If the sampling decision has not been made yet, we apply random sampling here.
+            if (transaction.IsSampled is null)
             {
-                if (Random.NextDouble() > _options.TracesSampleRate)
+                transaction.IsSampled =
+                    // If sample rate is 1, transaction is always sampled in
+                    _options.TracesSampleRate >= 1 ||
+                    // Otherwise roll the dice
+                    Random.NextDouble() > _options.TracesSampleRate;
```
bruno-garcia, line 150: > Random.NextDouble()

isn't thread safe. need to guard access to it

`lock(Random)` is probably good enough. If we don't do it for events, it's broken there too and I'm to blame.

Tyrrrz: wanna use eventId-based sampling?  :P
can add a test to ensure even distribution across supported platforms.

src/Sentry/Extensibility/DisabledHub.cs
```diff
@@ -55,10 +56,11 @@ public void WithScope(Action<Scope> scopeCallback)
         /// <summary>
         /// Returns a dummy transaction.
         /// </summary>
-        public ITransaction StartTransaction(string name, string operation) => new Transaction(this, name, operation)
-        {
-            IsSampled = false
-        };
+        public ITransaction StartTransaction(
+            ITransactionContext context,
+            IReadOnlyDictionary<string, object?> customSamplingContext) =>
```
Tyrrrz, line 61: Decided not to roll a custom type for this and just use `IReadOnlyDictionary<string, object?>` for custom sampling context. Thoughts @bruno-garcia?

src/Sentry/Internal/ThreadSafeRandom.cs
```diff
@@ -0,0 +1,20 @@
...
+    internal static class ThreadSafeRandom
+    {
+        private static readonly object Lock = new();
+        private static readonly Lazy<Random> RandomLazy = new();
+
+        private static Random Random => RandomLazy.Value;
+
+        public static double NextDouble()
+        {
+            lock (Lock)
```
bruno-garcia, line 14: `Random` is a private field here, why not use that for locking?

bruno-garcia (approved):

src/Sentry/Internal/ThreadSafeRandom.cs
Tyrrrz: It's an evaluated property here but I guess you can lock on the lazy instead. I always prefer a dedicate lock object for clarity of intent though.

src/Sentry/Internal/Hub.cs
```diff
@@ -121,6 +125,34 @@ public ITransaction StartTransaction(string name, string operation)
...
+            // Make a sampling decision
+            var samplingContext = new TransactionSamplingContext(
+                context,
+                customSamplingContext
+            );
+
+            var sampleRate =
+                // Custom sampler may not exist or may return null, in which case we fallback
+                // to the static sample rate.
+                _options.TracesSampler?.Invoke(samplingContext)
```
Tyrrrz, line 137: This is different from Java. I decided to allow the dynamic sampler to return `null` to fallback to the configured sample rate. I think it would be very useful, in case when you only want special sampling rules for a subset of transactions and default rules for everything else.

src/Sentry/Protocol/ISpan.cs
```diff
@@ -8,9 +8,22 @@ namespace Sentry.Protocol
     public interface ISpan : ISpanContext, IHasTags, IHasExtra
     {
         /// <summary>
-        /// Description.
+        /// Span description.
         /// </summary>
-        string? Description { get; set; }
+        // 'new' because it adds a setter.
+        new string? Description { get; set; }
```
Tyrrrz, line 14: `new` is yikes

src/Sentry/Protocol/Span.cs
```diff
@@ -33,17 +33,17 @@ public class Span : ISpan, IJsonSerializable
         /// <inheritdoc />
         public DateTimeOffset? EndTimestamp { get; private set; }
 
-        /// <inheritdoc />
+        /// <inheritdoc cref="ISpan.Operation" />
```
Tyrrrz, line 36: Side-effect of having `new`

src/Sentry/Protocol/SpanContext.cs
```diff
@@ -0,0 +1,46 @@
+﻿namespace Sentry.Protocol
+{
+    /// <summary>
+    /// Span metadata used for sampling.
+    /// </summary>
+    public class SpanContext : ISpanContext
```
Tyrrrz, line 6: I still don't know why we need `SpanContext` here though. Both guidelines and Java has it, but it's not used in sampling.

Tyrrrz:
@bruno-garcia are you happy with the API? I've commented on some places of interest.
If it's all good, I'll move on to tests.

src/Sentry/Internal/ThreadSafeRandom.cs
bruno-garcia: It's a good rule of thumb but this class is so simple, it's clear what we're doing here. Not locking on a public instance nor there are many instances, bug prone thing going on.

bruno-garcia: Are we saving up anything with the Lazy here? Maybe we just `new` up straight away? it's anyway a heap allocated object, or is the random ctor expensive?

src/Sentry/Internal/Hub.cs
bruno-garcia: It makes sense to me and I don't have a problem with that.
Maybe @marandaneto has opinions though.

Please make a note on the spec then that this is an alternative approach.

src/Sentry/Protocol/ISpan.cs
bruno-garcia: yeah, if the reference held is of a base type this `new` won't show up. And the wrong getter will run. But hey, not the worst thing we got going on to add this API right?

bruno-garcia (approved):
why did the submodule change?

Tyrrrz:
> why did the submodule change?

not sure, i just merged origin

src/Sentry/Internal/ThreadSafeRandom.cs
Tyrrrz: > Are we saving up anything with the Lazy here?

Nope, I just adapted the existing code which used lazy.

src/Sentry/Protocol/ISpan.cs
Tyrrrz: The getter will run correctly (since both interfaces still map to the same member on the class). However, if you implement explicitly then yeah you can end up with two different properties:

```csharp
ISpanContext.Description {get;set;}
ISpan.Description {get;set;}
```

Tyrrrz:
@bruno-garcia why is codecov not showing up? :(

Tyrrrz:
@bruno-garcia can you retry appveyor build? Don't know why it broke but I can't merge it because all checks need to pass.

src/Sentry/Internal/Hub.cs
marandaneto: that sounds a good use case, lets talk about this today in the meeting

marandaneto: we'll make this on Java too

PR #758, opened 2021-01-22
Integrate trace headers

Tyrrrz:
Related to #663 

Note, this doesn't include auto instrumentation for ASP.net Core, but only the core parts. Adding the instrumentation later would be easy once these are in (and won't be a breaking change).

### DIFF
--- a/src/Sentry/Internal/Hub.cs
+++ b/src/Sentry/Internal/Hub.cs
@@ -158,6 +158,7 @@ namespace Sentry.Internal
 
         public SentryTraceHeader? GetSentryTrace()
         {
+            // TODO:
             var (currentScope, _) = ScopeManager.GetCurrent();
             return currentScope.Transaction?.GetTraceHeader();
         }
--- a/src/Sentry/Internal/Hub.cs
+++ b/src/Sentry/Internal/Hub.cs
@@ -191,11 +191,20 @@ namespace Sentry.Internal
             }
         }
 
-        public void CaptureTransaction(Transaction transaction)
+        public void CaptureTransaction(ITransaction transaction)
         {
             try
             {
                 _ownedClient.CaptureTransaction(transaction);
+
+                // Clear the transaction from the scope
+                ScopeManager.WithScope(scope =>
+                {
+                    if (scope.Transaction == transaction)
+                    {
+                        scope.Transaction = null;
+                    }
+                });
             }
             catch (Exception e)
             {
--- a/src/Sentry/Internal/Hub.cs
+++ b/src/Sentry/Internal/Hub.cs
@@ -156,11 +156,10 @@ namespace Sentry.Internal
             return transaction;
         }
 
-        public SentryTraceHeader? GetSentryTrace()
+        public SentryTraceHeader? GetTraceHeader()
         {
-            // TODO:
             var (currentScope, _) = ScopeManager.GetCurrent();
-            return currentScope.Transaction?.GetTraceHeader();
+            return currentScope.GetSpan()?.GetTraceHeader();
         }
 
         public SentryId CaptureEvent(SentryEvent evt, Scope? scope = null)
--- a/src/Sentry/Protocol/SentryTraceHeader.cs
+++ b/src/Sentry/Protocol/SentryTraceHeader.cs
@@ -1,5 +1,7 @@
 using System;
+using System.Linq;
 using System.Net.Http;
+using System.Net.Http.Headers;
 
 namespace Sentry.Protocol
 {
@@ -82,5 +84,26 @@ namespace Sentry.Protocol
 
             return new SentryTraceHeader(traceId, spanId, isSampled);
         }
+
+        /// <summary>
+        /// Parses <see cref="SentryTraceHeader"/> from HTTP headers.
+        /// </summary>
+        public static SentryTraceHeader Parse(HttpHeaders headers)
+        {
+            if (!headers.TryGetValues(HeaderName, out var values) ||
+                !(values?.FirstOrDefault() is { } headerValue))
+            {
+                // TODO: we probably need a TryParse for this as it's gonna be common
+                throw new InvalidOperationException("Sentry trace header is not set.");
+            }
+
+            return Parse(headerValue);
+        }
+
+        /// <summary>
+        /// Parses <see cref="SentryTraceHeader"/> from an HTTP response.
+        /// </summary>
+        public static SentryTraceHeader Parse(HttpResponseMessage response) =>
+            Parse(response.Headers);
     }
 }
--- a/src/Sentry/Protocol/SentryTraceHeader.cs
+++ b/src/Sentry/Protocol/SentryTraceHeader.cs
@@ -1,4 +1,5 @@
 using System;
+using System.Net.Http;
 
 namespace Sentry.Protocol
 {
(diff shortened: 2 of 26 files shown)


src/Sentry/Internal/Hub.cs
```diff
@@ -158,6 +158,7 @@ public void WithScope(Action<Scope> scopeCallback)
 
         public SentryTraceHeader? GetSentryTrace()
         {
+            // TODO:
             var (currentScope, _) = ScopeManager.GetCurrent();
             return currentScope.Transaction?.GetTraceHeader();
         }
```
Tyrrrz, line 164: @bruno-garcia this previously relied that transaction is added to the current scope.
How should we get current transaction now? Or is it current span? Can you please explain how `Hub.GetSentryTrace()` should work?

src/Sentry/Protocol/SentryTraceHeader.cs
```diff
@@ -7,24 +8,59 @@ namespace Sentry.Protocol
...
+        /// <summary>
+        /// Injects trace information into the default headers of the specified HTTP client.
+        /// </summary>
+        public void Inject(HttpClient client)
+        {
+            var headerValue = ToString();
+
+            client.DefaultRequestHeaders.Remove(HeaderName);
+            client.DefaultRequestHeaders.Add(HeaderName, headerValue);
         }
```
Tyrrrz, line 58: It's either instance methods on `SentryTraceHeaders` or extension methods on `HttpRequestMessage` and `HttpClient` respectively. I have no preference. Thoughts?

src/Sentry/Protocol/SentryTraceHeader.cs
```diff
@@ -7,24 +10,60 @@ namespace Sentry.Protocol
...
+        /// Injects trace information into the headers of the specified HTTP request.
+        /// </summary>
+        public void Inject(HttpRequestMessage request) =>
+            Inject(request.Headers);
+
+        /// <summary>
+        /// Injects trace information into the default headers of the specified HTTP client.
+        /// </summary>
+        public void Inject(HttpClient client) =>
+            Inject(client.DefaultRequestHeaders);
```
Tyrrrz, line 61: This facilitates this type of workflow:

```csharp
using var client = new HttpClient();
using var request = new HttpRequestMessage(HttpMethod.Get, "https://example.com");
SentrySdk.GetTraceHeader().Inject(request);
```

Could make it as extension methods instead on `HttpHeaders`/`HttpRequestMessage`/`HttpClient` instead but not sure if worth.

src/Sentry/Internal/Hub.cs
bruno-garcia: I believe it needs to read the current span. In a world where we don't bind a whole transaction to the scope (also because it isnt' thread safe) we would only have a tree (in a list form) of spans. We could build this SentryTrace by getting the current span and with that the traceid and sampling decision. Isn't this spec'ed out? How is it in Java and Python?

/cc @marandaneto, @maciejwalkowiak, @untitaker

src/Sentry/Protocol/SentryTraceHeader.cs
bruno-garcia: I'm hoping we'll be able to run this through an integration instead of users having to do this manually on their `HttpClient`s. Maybe as a `MessageHandler` or even through ListenerHandler if that's possible:

https://github.com/open-telemetry/opentelemetry-dotnet/blob/2bc7924b065ddef8199bf69c2456724980522557/src/OpenTelemetry.Instrumentation.Http/Implementation/HttpHandlerDiagnosticListener.cs

https://github.com/open-telemetry/opentelemetry-dotnet/blob/2bc7924b065ddef8199bf69c2456724980522557/src/OpenTelemetry.Instrumentation.Http/Implementation/HttpInstrumentationEventSource.cs

src/Sentry/Internal/Hub.cs
untitaker: Python assembles the entire trace tree all the time. Regardless of sampling decision it propagates a trace header with outbound HTTP requests that is derived from the current span.

If you feel strongly about performance and don't want to create span IDs when the transaction sampling decision is 0, I think you could take something other than the current span as long as the trace ID is propagated properly.

>In a world where we don't bind a whole transaction to the scope (also because it isnt' thread safe)

This already deviates from Python behavior so fundamentally that I cannot give any advice here. Python scopes are basically threadsafe and the current span is always bound to the scope. This is useful because it means outgoing error events can grab the current span from the scope to attach a trace context that makes sense.

bruno-garcia: Yeah this "Transaction on the Scope" Seem to have originated on JS, then the Spec and now Java and .NET.

The topic wasn't about avoiding allocations in this case but really that, the transaction on the scope (as opposed to only spans).

Tyrrrz: @bruno-garcia all of that makes sense, I was more interested in the `current span` part. Considering it doesn't store the transaction anywhere when you call `StartTransaction`, where would it get the current span?

src/Sentry/Protocol/SentryTraceHeader.cs
Tyrrrz: These would also be useful from inside the handlers, I believe. I guess I could also add another method `GetHttpHandler()` that wraps an existing handler to make it easier for those use cases. I'll look into `ListenerHandler`, first time hearing about it.

src/Sentry/Internal/Hub.cs
bruno-garcia: If you want to make the transaction globally accessibly you must bind the transaction (or ideally the span list in a thread-ssafe manner instead, since we will have huge headaches with the Transaction thing on the scope!) to the Scope. It's optional. If it's not bound, it won't be available for integrations unless you programatically pass the value it's needed.

var tran = hub.StartTransction(..);

DoSomething("arg", tran);

void DoSomething(string arg, ITransaction tran)
{
 var client = new HttpClient(new SentryHandler(tran));
 client.Get()..
 // or
  var clietn = new HttpClient();
  var req = new HttpRequest();
  req.Headers.Add(tran.GetSentryTrace());
  client.Request(req);
}

Tyrrrz: Ok I see. So the code is right, it's just that it will rely on the user to set the transaction themselves.

Tyrrrz:
@bruno-garcia should be final, please do a final review

src/Sentry/SentryClient.cs
```diff
@@ -129,6 +129,17 @@ public void CaptureTransaction(ITransaction transaction)
                 return;
             }
 
+            if (!transaction.IsFinished)
+            {
+                _options.DiagnosticLogger?.LogWarning(
```
bruno-garcia, line 134: Loggin a warning makes sense.

Should we just finish the transaction and capture it instead?

How does it work in the other SDKs? Could be a behavior we align.

src/Sentry/SentryClient.cs
```diff
@@ -129,6 +129,17 @@ public void CaptureTransaction(ITransaction transaction)
...
+                    "Transaction dropped because it's not finished. " +
+                    "Don't call hub.CaptureTransaction(...) directly. " +
```
bruno-garcia, line 136:
```suggestion
                    "You don't need to call hub.CaptureTransaction(...) directly. " +
```

src/Sentry/SentryHttpMessageHandler.cs
```diff
@@ -0,0 +1,72 @@
...
+using System.Threading.Tasks;
+using Sentry.Extensibility;
+using Sentry.Protocol;
+
+namespace Sentry
+{
+    /// <summary>
+    /// Special HTTP message handler that can be used to propagate Sentry headers and other contextual information.
+    /// </summary>
+    public class SentryHttpMessageHandler : HttpMessageHandler
```
bruno-garcia, line 12: `DelegatingHandler` isn't a good option?

test/Sentry.Tests/SentryHttpMessageHandlerTests.cs
```diff
@@ -0,0 +1,40 @@
...
+using Sentry.Protocol;
+using Sentry.Testing;
+using Xunit;
+
+namespace Sentry.Tests
+{
+    public class SentryHttpMessageHandlerTests
+    {
+        [Fact]
+        public async Task SendAsync_SentryTraceHeaderNotSet_SetsHeader()
```
bruno-garcia, line 15: 1 test can really cover all edge cases this code can have?

bruno-garcia (approved):
I left some notes but it's good to merge. We can iterate with the other features etc.

src/Sentry/SentryClient.cs
marandaneto: that's a good point, we don't check it at all on Java

marandaneto: what's about logging it as `INFO` instead but finishing and capturing it?

marandaneto: Ive raised this issue https://github.com/getsentry/sentry-java/issues/1208

test/Sentry.Tests/SentryHttpMessageHandlerTests.cs
Tyrrrz: 2 tests for 2 branches, so I can't think of anything else ;)

src/Sentry/SentryClient.cs
Tyrrrz: Changed it so that it captures with a warning.

PR #759, opened 2021-01-23
ref: diagnosticLevel

bruno-garcia:
To align with Dart and Java:

https://github.com/getsentry/sentry-docs/pull/2945

### DIFF
--- a/src/Sentry/SentryOptions.cs
+++ b/src/Sentry/SentryOptions.cs
@@ -311,7 +311,7 @@ namespace Sentry
         /// Whether to log diagnostics messages
         /// </summary>
         /// <remarks>
-        /// The verbosity can be controlled through <see cref="DiagnosticsLevel"/>
+        /// The verbosity can be controlled through <see cref="DiagnosticLevel"/>
         /// and the implementation via <see cref="DiagnosticLogger"/>.
         /// </remarks>
         public bool Debug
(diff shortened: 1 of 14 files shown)


src/Sentry/SentryOptions.cs
```diff
@@ -326,7 +326,7 @@ public bool Debug
         /// <remarks>
         /// The <see cref="Debug"/> flag has to be switched on for this setting to take effect.
         /// </remarks>
-        public SentryLevel DiagnosticsLevel { get; set; } = SentryLevel.Debug;
+        public SentryLevel DiagnosticLevel { get; set; } = SentryLevel.Debug;
```
marandaneto, line 329: M: why don't we deprecate `DiagnosticsLevel ` and forwards getters/setters to `DiagnosticLevel`? so we won't break people's code, we could remove on the next version

marandaneto (approved):

Tyrrrz (approved):

src/Sentry/SentryOptions.cs
bruno-garcia: It's a breaking change and a niche option (most ppl are not fiddling with this).

Since it's a major bump and it's an easy to discover fix (even looking at intellisense ppl can figure out) I rather we just do it (optimize for our maintainalibiliy in this case).

If this was a rename on `BeforeSend` for example I'd surely not do this same thing.

PR #763, opened 2021-01-26
Add additional data to transactions

### LINKED ISSUE #762: Transactions should include all tags from the Scope
When a transaction is captured, it should include all tags from the scope.

More importantly, the SDK isn't adding `environment` and `release` to the transaction.

Tyrrrz:
Closes #762

### DIFF
--- a/src/Sentry/Internal/Hub.cs
+++ b/src/Sentry/Internal/Hub.cs
@@ -132,15 +132,15 @@ namespace Sentry.Internal
             }
 
             // Release information
-            transaction.Release = _options.Release ?? ReleaseLocator.GetCurrent();
+            transaction.Release ??= _options.Release ?? ReleaseLocator.GetCurrent();
 
             // Environment information
             var foundEnvironment = EnvironmentLocator.Locate();
-            transaction.Environment = string.IsNullOrWhiteSpace(foundEnvironment)
+            transaction.Environment ??= (string.IsNullOrWhiteSpace(foundEnvironment)
                 ? string.IsNullOrWhiteSpace(_options.Environment)
                     ? Constants.ProductionEnvironmentSetting
                     : _options.Environment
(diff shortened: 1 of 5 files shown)


Tyrrrz:
@bruno-garcia I made it so it applies scope and adds environment and release. Is that good/sufficient?

src/Sentry/Protocol/ITransaction.cs
```diff
@@ -18,6 +18,12 @@ public interface ITransaction : ISpan, ITransactionContext, IEventLike
         // 'new' because it adds a setter
         new string Name { get; set; }
 
+        // Should this be on IEventLike?
```
bruno-garcia, line 21: Yeah we should bump this so it's available on any Event (transaction or errors)

src/Sentry/Internal/Hub.cs
```diff
@@ -125,6 +131,17 @@ public void WithScope(Action<Scope> scopeCallback)
                 transaction.Sdk.AddPackage(protocolPackageName, nameAndVersion.Version);
             }
 
+            // Release information
+            transaction.Release ??= _options.Release ?? ReleaseLocator.GetCurrent();
+
+            // Environment information
+            var foundEnvironment = EnvironmentLocator.Locate();
```
bruno-garcia, line 138: Good for us to GA this on time but good reminder that we need to get back to this after GA.

bruno-garcia (approved):

src/Sentry/Protocol/ITransaction.cs
Tyrrrz: The problem is currently `IEventLike` also encompasses `Scope`, which does not have `Release` :/

bruno-garcia: We can add it, no problem

PR #767, opened 2021-01-27
Group unknown routes

kanadaj:
Creates a grouping for "Unknown Route" instead of using the path name for 404 responses.

Also exposes an `out` bool on `GetTransactionName` to make this simpler.

_#skip-changelog_

### DIFF
--- a/src/Sentry.AspNetCore/SentryTracingMiddleware.cs
+++ b/src/Sentry.AspNetCore/SentryTracingMiddleware.cs
@@ -34,7 +34,7 @@ namespace Sentry.AspNetCore
             }
 
             var transaction = hub.StartTransaction(
-                context.GetTransactionName(),
+                context.GetTransactionName(out var usesPathName),
                 "http.server"
             );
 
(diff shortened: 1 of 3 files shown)


bruno-garcia (approved):
LGTM. @Tyrrrz ?

src/Sentry.AspNetCore/SentryTracingMiddleware.cs
```diff
@@ -44,6 +44,8 @@ public async Task InvokeAsync(HttpContext context)
             }
             finally
             {
+                transaction.Name = context.GetTransactionName();
```
Tyrrrz, line 47: Let's add a comment here to explain why we get the transaction name again

Tyrrrz (approved):

src/Sentry.AspNetCore/SentryTracingMiddleware.cs
kanadaj: @Tyrrrz Not "again", if you check a few lines above I've replaced the default value with "Unknown Route"

Tyrrrz:
#skip-changelog

PR #792, opened 2021-02-03
changelog format craft auto

bruno-garcia:
Please lets start using:

### Changes

The sub heading is for the changelog generation in the nuget package that requires that. So unless there's a workaround, lets keep it. Unless we can split into `### Bug fixes` and `### Features`

And use `Unreleased` (and not anything else) so that I can use craft's `auto` mode

#skip-changelog

### DIFF
--- a/.craft.yml
+++ b/.craft.yml
@@ -2,7 +2,7 @@ minVersion: '0.13.2'
 github:
   owner: getsentry
   repo: sentry-dotnet
-changelogPolicy: simple
+changelogPolicy: auto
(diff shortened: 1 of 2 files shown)


lucas-zimerman (approved):
Look ok for me.
Later on, it would be nice to have an action for checking if the changelog is respecting that format.

bruno-garcia:
> Look ok for me.
> Later on, it would be nice to have an action for checking if the changelog is respecting that format.

PR welcome :)

PR #794, opened 2021-02-04
Try to re-enable flaky test

### LINKED ISSUE #600: Fix flaky test Init_WithCache_BlocksUntilExistingCacheIsFlushed
Error: Sentry.Tests.SentrySdkTests.Init_WithCache_BlocksUntilExistingCacheIsFlushed: Expected collection to be empty, but found {"/home/runner/work/sentry-dotnet/sentry-dotnet/test/Sentry.Tests/bin/Release/net461/56cb25c4-5538-4c98-b9cc-2cc487c63118/Sentry/E071FE7F4BE9BFA6EDB6089C375F386F4142AD01/__processing/1605994077_b2395428585a43baa8e7ea55afd4e08b_709824220.envelope"}.

Annoyingly breaking CI

Tyrrrz:
Closes #600 
_#skip-changelog_


bruno-garcia:
Based on the red X I assume it didn't work out?

Tyrrrz:
> Based on the red X I assume it didn't work out?

Nope :(

Tyrrrz:
@bruno-garcia I have a hunch that the test runner starts multiple tests in parallel, but somehow ends up executing only one or few of them at the time, maybe due to some thread contention issue or locking somewhere. Because of that, this test starts but doesn't finish in time and just times out. My theory is further supported by the fact that when I was playing with sampling tests, I could get them to run for a couple of seconds with large sample size, which would cause other tests to fail by timeout on CI (but not locally).

This Q/A seems relevant too: https://stackoverflow.com/a/56474681/2205454

PR #808, opened 2021-02-09
Ignore warning nca3.0

bruno-garcia:
#skip-changelog

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -10,6 +10,12 @@
     <Features>strict</Features>
   </PropertyGroup>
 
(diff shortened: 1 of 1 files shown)


Tyrrrz:
What about net461 and netstandard2.0? I believe those are also EOL or maybe soon.

Tyrrrz (approved):

bruno-garcia:
> What about net461 and netstandard2.0? I believe those are also EOL or maybe soon.

They're not warning at least. Let's opt-in once warnings creep in, if that happens.

PR #813, opened 2021-02-09
Fix tracing middleware

### LINKED ISSUE #805: Calling .UseSentryTracing() before .UseEndpoints() destroys ASP.NET exception handling
SDK: Sentry.AspNetCore 3.0.3

Calling `.UseSentryTracing()` before `.UseEndpoints()` in `Configure()` causes the ASP.NET exception handling flow to be messed up.

## Expected behaviour

Throwing an exception from an ASP.NET Core controller works normally, i.e. causes a HTTP 500 response to be returned, and any attached exception handler (`.UseExceptionHandler()`) will be triggered in the process.

## Actual behaviour

Sentry somehow swallows the exception and causes an empty HTTP 200 (!) response to be returned. Additionally, any attached exception handler is not triggered in the process.

## Workaround

Move the call to `.UseSentryTracing()` after `.UseEndpoints()` in `Configure()`.

Tyrrrz:
This should fix the following issues:

- Pipeline shortcircuit when `SentryTracingMiddleware` fails.
- `Failure to ConfigureScopeAsync` error due to attempt to configure local scope. Not exactly sure why it was caused in #805 but I assume it was that. @bruno-garcia can you check if my usage of `PushAndLockScope()` is valid here?
- `Unknown route` appearing in `TransactionSamplingContext`. Now it will attempt to get a name first, so there is a chance something useful will be available to the user.

Fixes #805

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## Unreleased
+
+- Fix tracing middleware (#813) @Tyrrrz
+
 ## 3.0.5
 
 ### Changes
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+### Changes
+
 - Fix tracing middleware (#813) @Tyrrrz
 
 ## 3.0.5
--- a/src/Sentry.AspNetCore/SentryTracingMiddleware.cs
+++ b/src/Sentry.AspNetCore/SentryTracingMiddleware.cs
@@ -14,6 +14,7 @@ namespace Sentry.AspNetCore
     internal class SentryTracingMiddleware
     {
         private const string UnknownRouteTransactionName = "Unknown Route";
+        private const string OperationName = "http.server";
 
         private readonly RequestDelegate _next;
         private readonly Func<IHub> _getHub;
@@ -72,12 +73,12 @@ namespace Sentry.AspNetCore
                 var transaction = traceHeader is not null
                     ? hub.StartTransaction(
                         transactionName,
-                        "http.server",
+                        OperationName,
                         traceHeader
                     )
                     : hub.StartTransaction(
                         transactionName,
-                        "http.server"
+                        OperationName
                     );
 
                 _options.DiagnosticLogger?.LogInfo(
--- a/src/Sentry.AspNetCore/SentryTracingMiddleware.cs
+++ b/src/Sentry.AspNetCore/SentryTracingMiddleware.cs
@@ -125,6 +125,7 @@ namespace Sentry.AspNetCore
                         // try to get the transaction name again.
                         if (context.TryGetTransactionName() is { } transactionName)
                         {
+                            _options.DiagnosticLogger?.LogDebug("Found transaction name after request pipeline executed: {0}.", transactionName);
                             transaction.Name = transactionName;
                         }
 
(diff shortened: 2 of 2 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+## Unreleased
```
bruno-garcia, line 3: Maybe we split in `### Feature`, `### Bug fixes`? Better than this crappy `### Changes`

```suggestion
## Unreleased

### Changes
```

src/Sentry.AspNetCore/SentryTracingMiddleware.cs
```diff
@@ -35,74 +38,102 @@ internal class SentryTracingMiddleware
                 return null;
             }
 
-            _options.DiagnosticLogger?.LogInfo("Received Sentry trace header: {0}", value);
+            _options.DiagnosticLogger?.LogInfo("Received Sentry trace header '{0}'.", value);
```
bruno-garcia, line 41: I wonder if debug is better here. It'll print out for all requests if it's properly integrated with a front end

```suggestion
            _options.DiagnosticLogger?.LogDebug("Received Sentry trace header '{0}'.", value);
```

src/Sentry.AspNetCore/SentryTracingMiddleware.cs
```diff
@@ -35,74 +38,102 @@ internal class SentryTracingMiddleware
...
+        {
+            var hub = _getHub();
+
+            if (!hub.IsEnabled)
+            {
+                await _next(context).ConfigureAwait(false);
+                return;
+            }
+
+            using (hub.PushAndLockScope())
```
bruno-garcia, line 112: isn't locking done by the error SDK? Do we want to do this here?

I think this whole locking was a bad idea. Better would have been to disconenct the MEL BeginScope with Sentry's PushScope

src/Sentry.AspNetCore/SentryTracingMiddleware.cs
```diff
@@ -35,74 +38,102 @@ internal class SentryTracingMiddleware
...
-                    if (context.TryGetTransactionName() is { } transactionName)
+                    if (transaction is not null)
                     {
-                        transaction.Name = transactionName;
-                    }
+                        // The routing middleware may have ran after ours, so
+                        // try to get the transaction name again.
+                        if (context.TryGetTransactionName() is { } transactionName)
+                        {
+                            transaction.Name = transactionName;
```
bruno-garcia, line 128:
```suggestion
                            _options.DiagnosticLogger?.LogDebug("Found transaction name after request pipeline executed: {0}.", transactionName);
                            transaction.Name = transactionName;
```

src/Sentry.AspNetCore/SentryTracingMiddleware.cs
Tyrrrz: Hm yeah. So should I use `ConfigureScopeAsync(...)` here instead? In the original issue, according to the logs, it failed for some reason. Maybe because the scope was locked?

bruno-garcia: I think you don't need to push a scope at all, instead just use the current scope available?

Though this would depend on the first middleware running. So maybe we just go ahead like this

bruno-garcia (approved):

PR #821, opened 2021-02-13
Fix: Disable Ben.Demystifier for UWP by default.

lucas-zimerman:
Ben Demystifier doesn't support .NET Native so we have to disable it for the UWP platform by default.

See benaadams/Ben.Demystifier#51 for more information.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Changes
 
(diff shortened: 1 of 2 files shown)


bruno-garcia (commented):
Thanks for the PR. Could you please revert the changes to the submodule?

bruno-garcia (approved):

PR #833, opened 2021-02-18
Ignore dirty submodule.

lucas-zimerman:
Based on https://stackoverflow.com/questions/3240881/git-can-i-suppress-listing-of-modified-content-dirty-submodule-entries-in-sta

This will stop adding garbage to commits with changes from Ben.Demystifier submodule.

#skip-changelog

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,4 @@
 [submodule "modules/Ben.Demystifier"]
 	path = modules/Ben.Demystifier
 	url = ../Ben.Demystifier.git
+    ignore = dirty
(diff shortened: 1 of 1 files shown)


.gitmodules
```diff
@@ -1,3 +1,4 @@
 [submodule "modules/Ben.Demystifier"]
 	path = modules/Ben.Demystifier
 	url = ../Ben.Demystifier.git
+	ignore = dirty
```
bruno-garcia, line 4: if you're making changes to it, how does it work?
It doens't show that you've got changes in it?

I believe the biggest question is: Why do you have local changes on the submodule?
```suggestion
	ignore = dirty

```

lucas-zimerman: changes are ignored, it could be that your submodule is outdated or updated compared to the main branch.

bruno-garcia: Not sure that's what we want though

Issue #818, opened 2021-02-12
Set tags from exception?


Lobosque:
We enrich exceptions adding context to the `Data` [dictionary](https://docs.microsoft.com/en-us/dotnet/api/system.exception.data?view=net-5.0)

Sentry does handle these and send them over, in the UI they appear as "Additional Data".

However, it is not possible to search errors in the UI using Additional data Key or Value, only "tags".

I wonder if it makes sense to somehow set tags using the Data dictionary? (maybe under a "tags" key).

Since our modules are not aware of Sentry (they get a Serilog logger instance only), I cannot see another way of settings specific tags on error handling time, but am welcome to suggestions.

bruno-garcia:
What if we special cased the key? As in `ex.Data["Sentry:Tag:Value"] = "tag value"`?
And when processing the exception data we could `StartsWith(Sentry:`, etc..

Either that or some mapping that can be provided via `init` but convention on key name sounds simpler.

lucas-zimerman:
I like the "Sentry:Tag:Value" idea, which would save me a lot of time on my app since I have to create an extra step before capturing the exception so that I extract the tags inside of the Exception Data.

bruno-garcia:
That could work to.

What about?

`Exception.Data["Sentry:Tag:123"] = "321"` results in 
`SentrySdk.SetTag("123", "321")`

And:
`Exception.Data["Sentry:Contexts:123"] = new { Obj = "312" }` results in
`SentrySdk.Contexts["123"] = new { Obj = "312" }`

@Tyrrrz thoughts?

bruno-garcia:
Another use case I came across:

Attaching a file by path, such as:

![image](https://user-images.githubusercontent.com/1633368/107900298-9da51880-6f0e-11eb-9a8d-7a396fe6f4ae.png)

Defining `path` explicitly to make room for attaching the actual payload or other forms in the future.

Makes sense?

Lobosque:
You guys implemented it before I could check the issue back again.

Awesome!

bruno-garcia:
This wasn't shipped yet though, it'll be on whatever comes after version 3.0.7 (probably 3.0.8)

Issue #822, opened 2021-02-14
Custom Context object serialized as string


bruno-garcia:
Adding an object to be serialized like:

```csharp
if (uname is { })
{
    s.Contexts["uname"] = new
    {
        uname.Machine,
        uname.Nodename,
        uname.Release,
        uname.Sysname,
        uname.Version
    };
}
```

Results in a broken contexts:

![image](https://user-images.githubusercontent.com/1633368/107865698-5d269b80-6e37-11eb-86f8-6e5be32f8be5.png)

Tyrrrz:
Specifically with anonymous objects?

bruno-garcia:
@Tyrrrz It can be an actual class defined by the user. We don't have a strategy of serializing objects via reflection, right?

Tyrrrz:
Nope. But more importantly, during deserialization that object will be resolved as a dictionary since there is no way to identify its type.

PR #834, opened 2021-02-18
feat: Additional filters to Exception Data

lucas-zimerman:
Filter exception data starting with sentry:tag and sentry:context to new tags and contexts. Those are going to be set by two new Extensions for Exception.
* AddSentryContext

Example:
```csharp
            var ex = new Exception("Test Exception data");
            ex.AddSentryTag("tag1", "some Data");
            ex.Data.AddSentryContext("My Cool Context", new Dictionary<string, object>
            {
                { "Data 1", new {a = 1, b = "hello world" } },
                { "Data 2", "Fish" }
            });
            _ = SentrySdk.CaptureException(ex);
```
is equivalent to:
```csharp
 SentrySDK.WithScope(scope => 
{
     scope.SetTag("tag1", "some Data");
     scope.Context.AddOrUpdate( "My Cool Context", new Dictionary<string, object>
    {
        { "Data 1", new {a = 1, b = "hello world" } },
        { "Data 2", "Fish" }
    });
    _ = SentrySdk.CaptureException(ex);
});
```

Closes https://github.com/getsentry/sentry-dotnet/issues/818


Tyrrrz:
> What's the best location to store the constant strings "sentry:tag:" and "sentry:context:" ?
> 
> Ideally the user would do something like:
> 
> ```cs
>  var x = new Exception();
>  x.Data.Add($"{SentryExceptionTag}Priority", "5");
> or
>  x.Data.Add($"{SentryExceptionTag}:Priority", "5");
> ```

Extension methods:

```csharp
public static void AddSentryTag(this Exception ex, string name, string value) =>
    ex.Data.Add($"sentry:tag:{name}", value);
```

src/Sentry/Internal/MainExceptionProcessor.cs
```diff
@@ -49,7 +56,23 @@ public void Process(Exception exception, SentryEvent sentryEvent)
...
+                        sentryException.Data[key] is object contextData &&
+                        key.Length > ExceptionDataContextKey.Length)
+                    {
+                        var keyName = key.Substring(ExceptionDataContextKey.Length);
+                        _ = sentryEvent.Contexts.AddOrUpdate(keyName, contextData, (_, value) => value);
+                    }
+                    else
+                    {
+                        sentryEvent.SetExtra($"Exception[{i}][{key}]", sentryException.Data[key]);
+                    }
```
Tyrrrz, line 75: How about first narrowing down by `sentry:` and then by type? Should make the conditions a bit simpler to read.
Also, make sure to use `StringComparison.Ordinal` or maybe even `StringComparison.OrdinalIgnoreCase` to avoid relying on current locale.

src/Sentry/Internal/MainExceptionProcessor.cs
```diff
@@ -49,7 +56,30 @@ public void Process(Exception exception, SentryEvent sentryEvent)
 
                 foreach (var key in sentryException.Data.Keys)
                 {
-                    sentryEvent.SetExtra($"Exception[{i}][{key}]", sentryException.Data[key]);
+                    if (key.StartsWith("sentry", StringComparison.OrdinalIgnoreCase))
```
Tyrrrz, line 59: Should probably be `sentry:` instead of `sentry` (i.e. to ignore something like `sentryButNotReally`). Maybe as a constant as well.

lucas-zimerman: Not sure about it being a constant since it'll only be used there (unlike the tag/context keys where they are being used by the tests and the Exception extension).

src/Sentry/Extensibility/ExceptionExtensions.cs
```diff
@@ -0,0 +1,30 @@
+using System;
+using System.Collections.Generic;
+using Sentry.Internal;
+
+namespace Sentry.Extensibility
```
bruno-garcia, line 5: Users will need to import this namespace to see the extension methods.
Better on `Sentry` namespace. `System` I wonder could be even better since we're naming the method all `XyzSentryQwe`

src/Sentry/Extensibility/ExceptionExtensions.cs
```diff
@@ -0,0 +1,30 @@
...
+        public static void AddSentryTag(this Exception ex, string name, string value)
+            => ex.Data.Add($"{MainExceptionProcessor.ExceptionDataTagKey}{name}", value);
+
+        /// <summary>
+        /// TODO
+        /// </summary>
+        /// <param name="ex"></param>
+        /// <param name="name"></param>
+        /// <param name="data"></param>
+        public static void AddSentryContext(this Exception ex, string name, Dictionary<string, object> data)
```
bruno-garcia, line 27: Does it make sense to take `IReadOnlyDictionary` instead?

```suggestion
        public static void AddSentryContext(this Exception ex, string name, IReadOnlyDictionary<string, object> data)
```

src/Sentry/Internal/MainExceptionProcessor.cs
```diff
@@ -9,6 +9,13 @@ namespace Sentry.Internal
 {
     internal class MainExceptionProcessor : ISentryEventExceptionProcessor
     {
+
+        /// <summary>
+        /// Ideally a public string
+        /// </summary>
```
bruno-garcia, line 15:
```suggestion
```

src/Sentry/Internal/MainExceptionProcessor.cs
bruno-garcia, line 59: Agree with @Tyrrrz 
```suggestion
                    if (key.StartsWith("sentry:", StringComparison.OrdinalIgnoreCase))
```

src/Sentry/Internal/MainExceptionProcessor.cs
```diff
@@ -49,7 +56,30 @@ public void Process(Exception exception, SentryEvent sentryEvent)
 
                 foreach (var key in sentryException.Data.Keys)
                 {
-                    sentryEvent.SetExtra($"Exception[{i}][{key}]", sentryException.Data[key]);
+                    if (key.StartsWith("sentry", StringComparison.OrdinalIgnoreCase))
+                    {
+                        if (key.StartsWith(ExceptionDataTagKey, StringComparison.OrdinalIgnoreCase))
```
bruno-garcia, line 61: Better then to iterate the map instead of only the keys though. You'd get back a KeyValuePair<T,V>.

src/Sentry/Internal/MainExceptionProcessor.cs
```diff
@@ -49,7 +56,30 @@ public void Process(Exception exception, SentryEvent sentryEvent)
...
+                        {
+                            var data = sentryException.Data[key] as string ?? "<invalid>";
```
bruno-garcia, line 63: I rather we drop the value (internal logging as error) since we expect this to be `string` (and nothing else, not calling `ToString()` either) since the API to get this key is `AddSentryTag(string,string)`

src/Sentry/Internal/MainExceptionProcessor.cs
```diff
@@ -49,7 +56,30 @@ public void Process(Exception exception, SentryEvent sentryEvent)
...
+                            //Set the key after the ExceptionDataTagKey string.
+                            sentryEvent.SetTag(key.Substring(ExceptionDataTagKey.Length), data);
+                        }
+                        else if (key.StartsWith(ExceptionDataContextKey, StringComparison.OrdinalIgnoreCase))
+                        {
+                            //Set the key after the ExceptionDataTagKey string.
+                            _ = sentryEvent.Contexts.AddOrUpdate(key.Substring(ExceptionDataContextKey.Length),
```
bruno-garcia, line 70: need to make sure the key isn't just `sentry:context:` or this would break.

src/Sentry/Internal/MainExceptionProcessor.cs
```diff
@@ -49,7 +56,30 @@ public void Process(Exception exception, SentryEvent sentryEvent)
...
+                                sentryException.Data[key] ?? "<empty>",
```
bruno-garcia, line 71: You can't use `<empty>` since `contexts` expects an object.

Since version 3.0.0 the SDK isn't serializing random objects anymore 
see: #822 

So we have a problem with this right now, but this must be an object, not a simple type.

src/Sentry/Internal/MainExceptionProcessor.cs
```diff
@@ -49,7 +56,30 @@ public void Process(Exception exception, SentryEvent sentryEvent)
...
+                                (_, value) => value);
+                        }
+                        else
+                        {
+                            sentryEvent.SetExtra($"Exception[{i}][{key}]", sentryException.Data[key]);
```
bruno-garcia, line 76: I believe Exception has a data bag now, this should go there instead. Not on this PR though.

On the mechanism:
https://develop.sentry.dev/sdk/event-payloads/exception/#exception-mechanism

image

bruno-garcia (commented):
This is good stuff!!

src/Sentry/Internal/MainExceptionProcessor.cs
lucas-zimerman: whoops I forgot to push 😦

src/Sentry/Extensibility/ExceptionExtensions.cs
lucas-zimerman: the root folder will end up with a lot of source files D:
but I agree the namespace should only be "Sentry"

src/Sentry/Internal/MainExceptionProcessor.cs
lucas-zimerman: users can set null to Contexts like
`SentrySdk.WithScope(s => { s.Contexts.AddOrUpdate("", null, (_, value) => value); });`

mmm, should I just ignore if the object is null?

src/Sentry/Extensibility/ExceptionExtensions.cs
Tyrrrz: Please don't put it on `System`, it's not good user experience. As a user, I don't want to see random extension methods appearing on `Exception` after I import `System` in my code.

Also, if you want to make a global extension method, using global namespace is better than assuming the user has `System` imported.

bruno-garcia: I thought about thr global namespace, agree it's better than System which has no benefits. It's not bad use rexperience to show the error tracking api on an exception instance. We're not polluting String or Int with random helpers.

src/Sentry/Internal/MainExceptionProcessor.cs
bruno-garcia: Key wirhiut a value we can log and drop. 

Value as string will just render an error is whst I said. It's not supported

src/Sentry/Internal/MainExceptionProcessor.cs
lucas-zimerman: It doesn't break from what I saw (+ added a test case for that rn)

src/Sentry/Internal/MainExceptionProcessor.cs
lucas-zimerman: In this case, I think the best scenario is to fall back to the original output as Exception[i][context] = data

src/Sentry/Extensibility/ExceptionExtensions.cs
lucas-zimerman: If you have a custom Exception, there's a risk the source file that caught the custom exception might not be using the System namespace, so it might be strange to the end-user that it requires to include the System namespace to use a Sentry extension instead of Sentry namespace.

Tyrrrz: > I thought about thr global namespace, agree it's better than System which has no benefits. It's not bad use rexperience to show the error tracking api on an exception instance. We're not polluting String or Int with random helpers.

I don't know, for me personally that's not a good UX. I would prefer to import `Sentry` or even `Sentry.Extensions` rather than have extension methods forced on me without any way to opt out. Then, when I will want to find out where that method comes from, I will see that it's defined in `System` namespace and that will be extremely confusing. Or worse, like @lucas-zimerman said, when I already have an instance of `Exception` and then have to import `System` to use Sentry-specific methods, also confusing.

src/Sentry/ExceptionExtensions.cs
```diff
@@ -0,0 +1,30 @@
+using System;
+using System.Collections.Generic;
+using Sentry.Internal;
+
+namespace Sentry
```
bruno-garcia, line 5: I think an extension method to `Exception` coming from your Crash reporting tool will not surprising and justifies adding to the global namespace. We need to continue to take special care with the name to avoid any potential conflict or ambiguity as you did in this PR (Having Sentry in the method name will be a must).

 ```suggestion
```

File: src/Sentry/Internal/MainExceptionProcessor.cs [hunk 12]
```diff
@@ -47,9 +51,33 @@ public void Process(Exception exception, SentryEvent sentryEvent)
...
+                foreach (var keyValue in sentryException.Data)
                 {
-                    sentryEvent.SetExtra($"Exception[{i}][{key}]", sentryException.Data[key]);
+                    if (keyValue.Key.StartsWith("sentry:", StringComparison.OrdinalIgnoreCase) &&
+                        keyValue.Value != null)
+                    {
+                        if (keyValue.Key.StartsWith(ExceptionDataTagKey, StringComparison.OrdinalIgnoreCase) &&
+                            keyValue.Value is string tagValue)
+                        {
+                            //Set the key after the ExceptionDataTagKey string.
```
bruno-garcia, line 62:
```suggestion
                            // Set the key after the ExceptionDataTagKey string.
```

src/Sentry/Internal/MainExceptionProcessor.cs
```diff
@@ -47,9 +51,33 @@ public void Process(Exception exception, SentryEvent sentryEvent)
...
+                            sentryEvent.SetTag(keyValue.Key.Substring(ExceptionDataTagKey.Length), tagValue);
+                        }
+                        else if (keyValue.Key.StartsWith(ExceptionDataContextKey, StringComparison.OrdinalIgnoreCase))
+                        {
+                            //Set the key after the ExceptionDataTagKey string.
```
bruno-garcia, line 67:
```suggestion
                            // Set the key after the ExceptionDataTagKey string.
```

src/Sentry/Internal/MainExceptionProcessor.cs
```diff
@@ -47,9 +51,33 @@ public void Process(Exception exception, SentryEvent sentryEvent)
...
+                            _ = sentryEvent.Contexts.AddOrUpdate(keyValue.Key.Substring(ExceptionDataContextKey.Length),
```
bruno-garcia, line 68: Curious why using AddOrUpdate here instead of plain: `[..] =` ?

The code still doesn't account for the key being an exact match to the prefix. It would use an empty string as key `""`. We should make sure the string is larger than the key at least.

bruno-garcia (commented):
This is in great shape, I had some final minor notes.

src/Sentry/Internal/MainExceptionProcessor.cs
lucas-zimerman: Not sure why, maybe I saw it at some old Sample/Doc.

src/Sentry/Extensibility/ExceptionExtensions.cs
bruno-garcia: `System` was a bad idea.

Global namespace is an excellent idea. The problem is discoverability. 

You'd need to known that you must `using Sentry` to see that. And Sentry is what you use to better instrument your app for errors, extending `Exception` is a great candidate.

I've seen extension methods such as `exception.Capture()` which I'm not a fan at all. But extension such as what @lucas-zimerman added sounds great.

CHANGELOG.md
```diff
@@ -3,7 +3,7 @@
 ## 3.0.7
 
 ### Changes
-
+- Add AddSentryTag and AddSentryContext Extensions for exception class (#834) @lucas-zimerman
```
bruno-garcia, line 6: Oh I just made a release, didn't see this PR got updated.

Can you please move this to the top:

```

## Unpublished

### Features

- Add AddSentryTag and AddSentryContext Extensions to Exception (#834) @lucas-zimerman


```

lucas-zimerman: @bruno-garcia  done

bruno-garcia (approved):

PR #848, opened 2021-02-23
Associate span exceptions with event exceptions

### LINKED ISSUE #738: Connect errors to the transaction (and span if possible) they happened in
Discussed in: https://github.com/getsentry/sentry-java/pull/1156

If an exception bubbles out of the transaction middleware, the error in Sentry should get connected to the transaction.

It was done in Java recently: https://github.com/getsentry/sentry-java/pull/1156

Tyrrrz:
Closes #738

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,9 +6,6 @@
 
 - Add AddSentryTag and AddSentryContext Extensions for exception class (#834) @lucas-zimerman
 - Associate span exceptions with event exceptions (#848) @Tyrrrz
-
-### Features
-
 - Move SentryHttpMessageHandlerBuilderFilter to Sentry.Extensions.Logging (#845) @Tyrrrz
 
 ## 3.0.7
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Features
 
 - Add AddSentryTag and AddSentryContext Extensions for exception class (#834) @lucas-zimerman
+- Associate span exceptions with event exceptions (#848) @Tyrrrz
 
 ## 3.0.7
 
--- a/src/Sentry/Internal/Hub.cs
+++ b/src/Sentry/Internal/Hub.cs
@@ -1,6 +1,7 @@
 using System;
 using System.Collections.Generic;
 using System.Diagnostics;
+using System.Runtime.CompilerServices;
 using System.Threading.Tasks;
 using Sentry.Extensibility;
 using Sentry.Integrations;
@@ -13,6 +14,8 @@ namespace Sentry.Internal
         private readonly SentryOptions _options;
         private readonly ISdkIntegration[]? _integrations;
         private readonly IDisposable _rootScope;
+
+        private readonly ConditionalWeakTable<Exception, ISpan> _exceptionToSpanMap = new();
 
         internal SentryScopeManager ScopeManager { get; }
 
@@ -170,6 +173,9 @@ namespace Sentry.Internal
             return transaction;
         }
 
+        public void BindException(Exception exception, ISpan span) =>
+            _exceptionToSpanMap.AddOrUpdate(exception, span);
+
         public ISpan? GetSpan()
         {
             var (currentScope, _) = ScopeManager.GetCurrent();
@@ -184,8 +190,19 @@ namespace Sentry.Internal
             {
                 var currentScope = ScopeManager.GetCurrent();
                 var actualScope = scope ?? currentScope.Key;
+
+                // Inject trace information from a linked span
+                if (evt.Exception is { } exception &&
+                    _exceptionToSpanMap.TryGetValue(exception, out var linkedSpan))
+                {
+                    evt.Contexts.Trace.SpanId = linkedSpan.SpanId;
+                    evt.Contexts.Trace.TraceId = linkedSpan.TraceId;
+                    evt.Contexts.Trace.ParentSpanId = linkedSpan.ParentSpanId;
+                }
+
                 var id = currentScope.Value.CaptureEvent(evt, actualScope);
                 actualScope.LastEventId = id;
+
                 return id;
             }
             catch (Exception e)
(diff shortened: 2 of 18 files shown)


src/Sentry/Transaction.cs
```diff
@@ -254,8 +257,13 @@ public void Finish()
         {
             EndTimestamp = DateTimeOffset.UtcNow;
 
+            if (Exception is {} exception)
```
bruno-garcia, line 260: This means that if the exception is logged before we finish the transaction, it won't be bound.

Ideally we'd be able to "connect the exception instance to a span" and if CaptureException happens right after that, it would be connected.

```
void method(ISpan tran) {
var child = tran.StartChild(..);
try {
throw null;
}
catch (e) {
child.Finish(e);
// Will know in which exact Span that the exception happened
// don't rethrow
}
`

File: src/Sentry/Span.cs [hunk 2]
```diff
@@ -87,7 +95,15 @@ public Span(Transaction parentTransaction, SpanId? parentSpanId, string operatio
...
 
         /// <inheritdoc />
-        public void Finish() => EndTimestamp = DateTimeOffset.UtcNow;
+        public void Finish()
+        {
+            EndTimestamp = DateTimeOffset.UtcNow;
+
+            if (Exception is {} exception)
+            {
+                _hub.BindException(exception, this);
```
bruno-garcia, line 104: What if we have an overload to `Finish` that takes an Exception instead of the Exception property? Or am I missing a use case where just setting the exception without finishing is needed?

bruno-garcia (commented):
I wonder if we can find a way without adding `BindException` to `IHub`

src/Sentry/Internal/Polyfills.cs
```diff
@@ -34,6 +35,17 @@ internal static partial class PolyfillExtensions
...
+
+    public static void AddOrUpdate<TKey, TValue>(
+        this ConditionalWeakTable<TKey, TValue> table,
+        TKey key,
+        TValue value)
+        where TKey : class
+        where TValue : class
+    {
+        table.Remove(key);
+        table.Add(key, value);
```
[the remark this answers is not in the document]
Tyrrrz: > should we lock on table and also lock before looking up on Hub?

Not sure I understood the second part. You mean a lock inside the polyfill?

> Can you use just table[key] = value instead?

Nope.

src/Sentry/SentrySdk.cs
```diff
@@ -357,6 +357,16 @@ public static ITransaction StartTransaction(string name, string operation, strin
...
             => _hub.StartTransaction(name, operation, traceHeader);
 
+        /// <summary>
+        /// Binds specified exception the specified span.
+        /// </summary>
+        /// <remarks>
+        /// This method is used internally and is not meant for public use.
+        /// </remarks>
+        [DebuggerStepThrough]
+        public static void BindException(Exception exception, ISpan span)
```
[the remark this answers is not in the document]
Tyrrrz: Well the `Span` and `Transaction` have a reference to `IHub`, not `Hub`.
Can change it to `Hub` but that will have implications in your unit tests.

src/Sentry/Transaction.cs
Tyrrrz: This is how it works in Java btw.
When should it happen then? When `transaciton.Exception` changes?

src/Sentry/Span.cs
Tyrrrz: We do have that overload, as an extension method.

src/Sentry/Internal/Polyfills.cs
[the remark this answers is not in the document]
bruno-garcia: > Not sure I understood the second part. You mean a lock inside the polyfill?

And also when reading from it on the other method. It's not a very elegant solution though

src/Sentry/Internal/Hub.cs
```diff
@@ -184,8 +193,19 @@ public SentryId CaptureEvent(SentryEvent evt, Scope? scope = null)
             {
                 var currentScope = ScopeManager.GetCurrent();
                 var actualScope = scope ?? currentScope.Key;
+
+                // Inject trace information from a linked span
+                if (evt.Exception is { } exception &&
```
bruno-garcia, line 198: I notice d we're not taking trace info from the scopeif there's a span kn the scope we can connect events to it through the span bound to the scope (or currently the transaction property). I guess separate PR for that. This is what other sdks do btw so we should have this on the sdk docs as a spec otherwise no connection between errors and transactions

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 ### Features
 
 - Add AddSentryTag and AddSentryContext Extensions for exception class (#834) @lucas-zimerman
+- Associate span exceptions with event exceptions (#848) @Tyrrrz
 
 ### Features
```
bruno-garcia, line 11:
```suggestion
### Features

- Add AddSentryTag and AddSentryContext Extensions for exception class (#834) @lucas-zimerman
- Associate span exceptions with event exceptions (#848) @Tyrrrz
```

src/Sentry/Span.cs
```diff
@@ -73,9 +76,11 @@ public class Span : ISpan, IJsonSerializable
         /// <summary>
         /// Initializes an instance of <see cref="Span"/>.
         /// </summary>
-        public Span(Transaction parentTransaction, SpanId? parentSpanId, string operation)
+        public Span(IHub hub, Transaction parentTransaction, SpanId? parentSpanId, string operation)
```
bruno-garcia, line 79: Once we go with the SentryTracer approach taken by @brustolin we can drop this requirement of having IHub in Span

src/Sentry/Span.cs
```diff
@@ -155,6 +175,8 @@ public void WriteTo(Utf8JsonWriter writer)
         /// </summary>
         public static Span FromJson(Transaction parentTransaction, JsonElement json)
         {
+            var hub = DisabledHub.Instance;
```
bruno-garcia, line 178:
```suggestion
            // A Span deserialized is just used as wire-protocol and doesn't need a Hub
            var hub = DisabledHub.Instance;
```

bruno-garcia (approved):

PR #877, opened 2021-03-05
Don't add version prefix in release if it's already set

### LINKED ISSUE #873: Only include release prefix if it doesn't already exist
This change https://github.com/getsentry/sentry-dotnet/pull/682/files introduced a regression for customers that already appended their own prefix: `app@`.

Before adding a prefix we should make sure one isn't there yet.

Tyrrrz:
Closes #873

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Features
 
 - Adding TaskUnobservedTaskExceptionIntegration to default integrations and method to remove it (#870) @FilipNemec
+- Don't add version prefix in release if it's already set (#877) @Tyrrrz
 
 ## 3.0.8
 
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 ### Features
 
 - Adding TaskUnobservedTaskExceptionIntegration to default integrations and method to remove it (#870) @FilipNemec
+- Don't add version prefix in release if it's already set (#877) @Tyrrrz
```
bruno-garcia, line 8:
```suggestion

### Fixes
  - Don't add version prefix in release if it's already set (#877) @Tyrrrz
```

bruno-garcia (approved):

Issue #631, opened 2020-11-30
Refactor manual JSON serialization


Tyrrrz:
- Generalize "don't write if value is null" into a single method
- Cleanup `JsonExtensions` into more meaningful methods
- ~~Make use of `EnumMember` attribute when serializing enums~~
- Use `JsonSerializer.Serializer` where it's more appropriate than `WriteDynamic` extension method
- ~~Potentially use `JsonConverter` instead of `IJsonSerializable`~~
- Type-check when deserializing `Dictionary<string, string>` (i.e. not nullable)

Related to #623

PR #880, opened 2021-03-11
Rework transactions

Tyrrrz:


Tyrrrz:
@bruno-garcia RFC

src/Sentry/Extensibility/DisabledHub.cs
```diff
@@ -72,7 +71,7 @@ public void BindException(Exception exception, ISpan span)
         /// <summary>
         /// Returns null.
         /// </summary>
-        public ISpan? GetSpan() => null;
+        public ISpanTracer? GetSpan() => null;
```
bruno-garcia, line 74: Ideally we'd keep this `ISpan`. Technically any span can be set to the scope, you can create a child and set to the scope so next startSpan would be two levels down from the transaction

src/Sentry/Extensibility/HubAdapter.cs
```diff
@@ -75,7 +74,7 @@ public void WithScope(Action<Scope> scopeCallback)
         /// Forwards the call to <see cref="SentrySdk"/>.
         /// </summary>
         [DebuggerStepThrough]
-        public ITransaction StartTransaction(
+        public ITransactionTracer StartTransaction(
```
bruno-garcia, line 77: This should return `ISpan`

src/Sentry/ISpanTracer.cs
```diff
@@ -0,0 +1,65 @@
...
+        /// <summary>
+        /// Span status.
+        /// </summary>
+        // 'new' because it adds a setter.
+        new SpanStatus? Status { get; set; }
+
+        /// <summary>
+        /// Starts a child span.
+        /// </summary>
+        ISpanTracer StartChild(string operation);
```
bruno-garcia, line 31: You should be able to create a child from a span. Why did this need to move to this interface?

bruno-garcia:
Turned out quite different than Java and Objective-C, could you please elaborate the motivations?

src/Sentry/Extensibility/DisabledHub.cs
Tyrrrz: `ISpan` doesn't have `StartChild()` though so we do need `ISpanTracer` here tho. Unless you have something else in mind?

Tyrrrz:
@bruno-garcia it's actually the same as Java. The only difference is that I added interfaces for the Transaction/Span tracers. Other than that it's identical.

src/Sentry/ISpanTracer.cs
Tyrrrz: Span effectively cannot be modified, as it represents the final state included in a transaction. It's the same as Java's `SentrySpan` vs `Span`.

src/Sentry/Extensibility/DisabledHub.cs
bruno-garcia: Can't we add StartChild on `ISpan`? A Span can have children so I don't see a reason not to:
Java has it too: https://github.com/getsentry/sentry-java/blob/7c3f88ad24c95af317c73a042951e1eabc4032be/sentry/src/main/java/io/sentry/Span.java#L67

If we really wanted to split the serialization format from what we have floating around at runtime it would be a reason not to add this. But `ISpanTracer` as a public type is a new concept that wasn't discussed with the wider team.

bruno-garcia: @marandaneto ^

src/Sentry/ISpanTracer.cs
bruno-garcia: Isn't SpanContext cannot be modified?

bruno-garcia:
This PR includes a lot of breaking changes so before we go ahead and merge this (and essentially sign up for releasing 4.0.0 already) lets make sure we get all the details about this right.

Tyrrrz:
@bruno-garcia how should we go around it? set up a meeting or wait for the usual sync?

bruno-garcia:
Ideally we keep this rolling if just async until we can make a decision. 
Let me ping folks on Discord

src/Sentry/Extensibility/DisabledHub.cs
marandaneto: yep, sounds like `ISpan` would do the job

src/Sentry/Extensibility/HubAdapter.cs
marandaneto: on Java, `StartTransaction` still returns `ITransaction` which extends `ISpan` to not do breaking changes, `getSpan` returns `ISpan`, and `startChild` aways returns `ISpan`

that's what makes the most sense for me, in the feature `StartTransaction` could also return `ISpan`

right now this introduces a breaking change by changing the return type, we've agreed to not do that, right? unless `ITransactionTracer` implements  `ITransaction`, didn't check yet

src/Sentry/ISpanTracer.cs
marandaneto: on Java, `Span` takes a `SentryTracer` as ctor param and when `startChild` is called, it calls `SentryTracer#startChild`

src/Sentry/ISpanTracer.cs
```diff
@@ -0,0 +1,65 @@
+﻿using System;
+
+namespace Sentry
+{
+    /// <summary>
+    /// Span tracer.
+    /// </summary>
+    public interface ISpanTracer : ISpan
```
marandaneto, line 8: ISpanTracer and ITransactionTracer add a suffix to the API, what was the inspiration for it?

src/Sentry/Span.cs
```diff
@@ -199,8 +180,8 @@ public static Span FromJson(Transaction parentTransaction, JsonElement json)
                 Description = description,
                 Status = status,
                 IsSampled = isSampled,
-                _tags = tags,
-                _data = data
+                _tags = tags!,
+                _extra = data!
```
marandaneto, line 184: I'm not a big fun of `!`, maybe check if its null and set a new Dic if null? just for the sake of nul safety

src/Sentry/TransactionTracer.cs
```diff
@@ -0,0 +1,271 @@
...
+        // var t = new Transaction(hub, "my transaction", "my op")
+        // {
+        //     Contexts = new Contexts
+        //     {
+        //         ["my ctx"] = myObj
+        //     }
+        // }
+        //
+        // The above code overwrites the contexts and, as a result, the "my op" value
+        // is lost completely.
```
marandaneto, line 31: I'm not entirely sure if it's actually helpful to have this in the docs, we already have an issue for that, right?
giving examples of what to NOT do is..a red flag.

src/Sentry/TransactionTracer.cs
```diff
@@ -0,0 +1,271 @@
...
+            }
+            finally
+            {
+                // Clear the transaction from the scope
+                _hub.ConfigureScope(scope =>
+                {
+                    if (scope.Transaction == this)
+                    {
+                        scope.Transaction = null;
+                    }
```
marandaneto, line 245: should we make this Atomic? it is on Java

marandaneto:
so my 2 cents are pretty much what @Tyrrrz said, looks the same as Java except for the other 2 interfaces, and the reasoning is how the code is structured, immutability or not, etc.
so @bruno-garcia your call to either keep as it is or refactoring to get rid of them and keep as Java with ISpan ITransaction like, as I don't know the code base, I can't really give more than that.

src/Sentry/TransactionTracer.cs
Tyrrrz: yeah, this was added before the issue existed.
i will remove it.

src/Sentry/ISpanTracer.cs
Tyrrrz: I think the main difference in this version is that

ISpan - immutable; Protocol.Span implements ISpan
ISpanTracer - ISpan + mutations (including StartChild); SpanTracer implements ISpanTracer; potentially NoOpSpanTracer would also implement it.

and then essentially the same with ITransaction/ITransactionTracer.

src/Sentry/Span.cs
Tyrrrz: I agree.
we can't fully guarantee that the map is not nullable in JSON, but we're responsible for both serialization and deserialization parts here, so we can reasonably guarantee it.
in general though, it would be better to add a proper check, but it's on our other list of improvements: #631

src/Sentry/TransactionTracer.cs
Tyrrrz: @marandaneto 
I noticed that the code inside the `clearTransaction()` method is locked, but the `if (transaction == this)` part is not:

https://github.com/getsentry/sentry-java/blob/6c371b38c494ba08d0f4fbe4823cde9f378f1b5f/sentry/src/main/java/io/sentry/SentryTracer.java#L106-L111

In C# we can do `Interlocked.CompareExchange(...)` to make an atomic compare + replace operation, but that requires a direct field access...

Tyrrrz:
@marandaneto @bruno-garcia I adjusted the naming a bit so that it's less of a breaking change now.

src/Sentry/TransactionTracer.cs
marandaneto: @Tyrrrz `scope.withTransaction` is fully locked, have a look at the `withTransaction` impl

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 ### Changes
 
+- Rework transactions (#880) @Tyrrrz
```
bruno-garcia, line 7: Please add a more descriptive entry such as `CaptureTransaction takes concrete Transaction class` or whatever else was done here worth noting to the users?

src/Sentry/Span.cs
bruno-garcia: We had a few experiences with deserializing something that we expect to be in a certain format only to find out it's not.

I agree with @marandaneto here and even though not the nicest thing, ideally we'd:

`tags ?? new Dictionary<>()` to avoid a null reference exception if we're accessing `tags` from  a deserialized instance.

src/Sentry/Transaction.cs
```diff
@@ -205,16 +180,29 @@ public Transaction(IHub hub, string name, string operation)
         /// <summary>
         /// Initializes an instance of <see cref="Transaction"/>.
         /// </summary>
-        public Transaction(IHub hub, ITransactionContext context)
-            : this(hub, context.Name)
+        public Transaction(ITransaction tracer)
+        : this(tracer.Name, tracer.Operation)
```
bruno-garcia, line 184: nit:

```suggestion
            : this(tracer.Name, tracer.Operation)
```

bruno-garcia (approved):
Great stuff! Thanks.
I left a couple of small notes before we merge

src/Sentry/Span.cs
Tyrrrz: @bruno-garcia it's not the `tags` that is null here. The error is raised because the second generic argument in dictionary is nullable here.

bruno-garcia: Oh, I see now. Not clear at all from just `!`. Lost chance to have a nicer syntax

`tags<>!` :P

bruno-garcia (approved):

bruno-garcia:
@Tyrrrz LGTM but CI has a test timing out after 7 seconds. Flaky or bug?

src/Sentry/Span.cs
Tyrrrz: Indeed, that would have been much better :P

Tyrrrz:
@bruno-garcia yeah the same flaky tests are failing :/
I remember we briefly discussed that they could be caused by inconsistent thread locking in xUnit. I think we should look into it as soon as we can, it has become more common lately, even though the tests have not changed.

PR #885, opened 2021-03-15
Add IsParentSampled to TransactionContext

### LINKED ISSUE #882: Align transaction sampling with spec
[The spec](https://develop.sentry.dev/sdk/performance/#sampling) defines that if a callback is defined it should be called even if a header exists with sampled=true. The sampling decision should be in sampling context.

We need to update samples so this is clear.

https://github.com/getsentry/sentry-dotnet/blob/2337261333d5743996d6d96ca3aa289d34a9d3ec/src/Sentry/Internal/Hub.cs#L114-L139

More context: https://github.com/getsentry/sentry-java/issues/1315#issuecomment-797071507

Tyrrrz:
Closes #882 

Please validate the logic I implemented:

- Sampling callback ALWAYS runs and has access to `transactionContext.IsParentSampled` that allows it to reject transactions started from header.
- Sampling callback may return `null` which just means it didn't make any decision. If the decision has already been made before it (via header) then it stays, otherwise random sampling is applied. This also means that returning `null` on transactions started from header will not fall back to random sampling, but instead will retain their sampling decision (docs need to be updated).
- Random sampling is applied ONLY if no sampling decision has been made yet, either via header or by sampler callback.
- With `tracesSampler = null`, transactions started from header will always inherit sampling of their parent regardless of static sampling (even if `tracesSampleRate = 0.0`).
- With `tracesSampler = _ => 0`, ALL transactions will be sampled out, including local ones and those started from header.

### DIFF
--- a/src/Sentry/Internal/Hub.cs
+++ b/src/Sentry/Internal/Hub.cs
@@ -108,7 +108,7 @@ namespace Sentry.Internal
         {
             var transaction = new Transaction(this, context);
 
-            // Dynamic sampling is ran regardless of whether a decision
+            // Tracing sampler callback runs regardless of whether a decision
             // has already been made, as it can be used to override it.
             if (_options.TracesSampler is { } tracesSampler)
             {
--- a/src/Sentry/Internal/Hub.cs
+++ b/src/Sentry/Internal/Hub.cs
@@ -123,7 +123,7 @@ namespace Sentry.Internal
                 }
             }
 
-            // Static sampling is ran only if the sampling decision hasn't
+            // Random sampling runs only if the sampling decision hasn't
             // been made already.
             transaction.IsSampled ??= SynchronizedRandom.NextBool(_options.TracesSampleRate);
 
--- a/src/Sentry/Internal/Hub.cs
+++ b/src/Sentry/Internal/Hub.cs
@@ -117,7 +117,7 @@ namespace Sentry.Internal
                     customSamplingContext
                 );
 
-                if (tracesSampler(samplingContext) is {} sampleRate)
+                if (tracesSampler(samplingContext) is { } sampleRate)
                 {
                     transaction.IsSampled = SynchronizedRandom.NextBool(sampleRate);
                 }
--- a/src/Sentry/Internal/Hub.cs
+++ b/src/Sentry/Internal/Hub.cs
@@ -109,34 +109,26 @@ namespace Sentry.Internal
         {
             var transaction = new Transaction(this, context);
 
-            // Transactions are not handled by event processors, so some things need to be added manually
+            var samplingContext = new TransactionSamplingContext(
+                context,
+                customSamplingContext
+            );
 
-            // Make a sampling decision if it hasn't been made already.
-            // It could have been made by this point if the transaction was started
-            // from a trace header which contains a sampling decision.
-            if (transaction.IsSampled is null)
+            var sampleRate =
+                // Custom sampler may not exist or may return null, in which case we fallback
+                // to the static sample rate.
+                _options.TracesSampler?.Invoke(samplingContext)
+                ?? _options.TracesSampleRate;
+
+            transaction.IsSampled = sampleRate switch
             {
-                var samplingContext = new TransactionSamplingContext(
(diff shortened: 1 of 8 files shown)


bruno-garcia:
> Closes #882
> 
> Please validate the logic I implemented:
> 
> * Sampling callback ALWAYS runs and has access to `transactionContext.IsParentSampled` that allows it to reject transactions started from header.
> * Sampling callback may return `null` which just means it didn't make any decision. If the decision has already been made before it (via header) then it stays, otherwise random sampling is applied. This also means that returning `null` on transactions started from header will not fall back to random sampling, but instead will retain their sampling decision (docs need to be updated).
> * Random sampling is applied ONLY if no sampling decision has been made yet, either via header or by sampler callback.
> * With `tracesSampler = null`, transactions started from header will always inherit sampling of their parent regardless of static sampling (even if `tracesSampleRate = 0.0`).
> * With `tracesSampler = _ => 0`, ALL transactions will be sampled out, including local ones and those started from header.

This makes sense to me.

src/Sentry/Internal/Hub.cs
```diff
@@ -109,41 +108,28 @@ public void WithScope(Action<Scope> scopeCallback)
         {
             var transaction = new Transaction(this, context);
 
-            // Transactions are not handled by event processors, so some things need to be added manually
-
-            // Make a sampling decision if it hasn't been made already.
-            // It could have been made by this point if the transaction was started
-            // from a trace header which contains a sampling decision.
-            if (transaction.IsSampled is null)
+            // Dynamic sampling is ran regardless of whether a decision
```
bruno-garcia, line 111: Lets not call this "dynamic sampling" because there's a feature in the works called Dynamic Sampling and it involved being able to change sampling on the server via configuration that affects Relays ingesting the data.

This is the trace sampler callback.

bruno-garcia, line 111:
```suggestion
            // Tracing sampler callback runs regardless of whether a decision
```

src/Sentry/Internal/Hub.cs
```diff
@@ -109,41 +108,28 @@ public void WithScope(Action<Scope> scopeCallback)
...
                 );
 
-                var sampleRate =
-                    // Custom sampler may not exist or may return null, in which case we fallback
-                    // to the static sample rate.
-                    _options.TracesSampler?.Invoke(samplingContext)
-                    ?? _options.TracesSampleRate;
-
-                transaction.IsSampled = sampleRate switch
+                if (tracesSampler(samplingContext) is {} sampleRate)
```
bruno-garcia, line 120: nit: Don't really care if it's `{}` or `{ }` but I wish a linter would just keep it consistent across the code base.

On line 113 you used `{ }` so:
```suggestion
                if (tracesSampler(samplingContext) is { } sampleRate)
```

src/Sentry/Internal/Hub.cs
```diff
@@ -109,41 +108,28 @@ public void WithScope(Action<Scope> scopeCallback)
...
-                    // Sample rate <= 0 means always sampled *out*
-                    <= 0 => false,
-                    // Otherwise roll the dice
-                    _ => SynchronizedRandom.NextDouble() < sampleRate
-                };
+                    transaction.IsSampled = SynchronizedRandom.NextBool(sampleRate);
+                }
             }
 
+            // Static sampling is ran only if the sampling decision hasn't
```
bruno-garcia, line 126:
```suggestion
            // Random sampling runs only if the sampling decision hasn't
```

src/Sentry/SentryOptions.cs
```diff
@@ -420,6 +420,11 @@ public bool Debug
         /// Values outside of this range are invalid.
         /// Default value is <code>0</code>, which means tracing is disabled.
         /// </summary>
+        /// <remarks>
+        /// Static sampling rate is only applied to transactions that don't already
```
bruno-garcia, line 424:
```suggestion
        /// Random sampling rate is only applied to transactions that don't already
```

bruno-garcia (approved):

PR #887, opened 2021-03-23
Add StartupTime and Device.BootTime

### LINKED ISSUE #866: Add `App.StartupTime` and `Device.BootTime`
iOS and Android send this data and we're [planning to extend it to show duration from it to event](https://github.com/getsentry/sentry/issues/24209).

Device.BootTime was done on SharpRaven actually, somehow never got added to `Sentry`. What I did there was:

```csharp
new DateTimeOffset(DateTime.UtcNow - TimeSpan.FromTicks(Stopwatch.GetTimestamp()),
                    TimeSpan.Zero);
```

Not sure if it's always accurate though, there might be a better approach.

AppStartup we could track the time from when `Init` was called and assume it's the start of the app, if there's no better way.
On RavenSharp I did:

```csharp
using (var proc = Process.GetCurrentProcess())
{
    app.StartTime = proc.StartTime.ToUniversalTime();
}
```

But again, is there a cost on `Process.GetCurrentProcess`? Is there a better way?

If it makes sense we could also add AppName (if not defined by the user in the scope/event) as done before too:

https://github.com/getsentry/raven-csharp/blob/678d83f8784f107bcb4acf6439aaee50fe4f8f5e/src/app/SharpRaven/Data/Context/App.cs#L125-L130

https://github.com/getsentry/raven-csharp/blob/678d83f8784f107bcb4acf6439aaee50fe4f8f5e/src/app/SharpRaven/Utilities/EntryAssemblyNameLocator.cs#L26-L29

Other Device info we got on SharpRaven but not here we can consider adding: https://github.com/getsentry/raven-csharp/blob/678d83f8784f107bcb4acf6439aaee50fe4f8f5e/src/app/SharpRaven/Data/Context/Device.cs#L191-L193

lucas-zimerman:
StartTime and Device.BootTime was not being implemented, were previously on Raven SDK had those fields.

This Pull Request aims to implement the two missing fields into Sentry .NET SDK.

Close #866.


src/Sentry/Internal/Enricher.cs
```diff
@@ -90,6 +91,12 @@ public void Apply(IEventLike eventLike)
                 eventLike.User.IpAddress ??= "{{auto}}";
             }
 
+            //Apply App startup time
+            eventLike.Contexts.App.StartTime ??= _options.StartupTime;
+
+            eventLike.Contexts.Device.BootTime ??= new DateTimeOffset(DateTime.UtcNow - TimeSpan.FromTicks(Stopwatch.GetTimestamp()),
```
Tyrrrz, line 97: Can this be instead:

```csharp
eventLike.Contexts.Device.BootTime ??= DateTimeOffset.UtcNow - TimeSpan.FromTicks(Stopwatch.GetTimestamp());
```

?

src/Sentry/SentryOptions.cs
```diff
@@ -95,6 +93,11 @@ public class SentryOptions
         /// <seealso href="https://docs.sentry.io/platforms/dotnet/guides/aspnet/configuration/options/#in-app-include"/>
         internal string[]? InAppInclude { get; set; }
 
+        /// <summary>
+        /// When the code was initialized/
+        /// </summary>
+        internal DateTime StartupTime { get; set; } = DateTime.UtcNow;
```
Tyrrrz, line 99: Should probably use `DateTimeOffset`

src/Sentry/SentryOptions.cs
```diff
@@ -95,6 +93,11 @@ public class SentryOptions
         /// <seealso href="https://docs.sentry.io/platforms/dotnet/guides/aspnet/configuration/options/#in-app-include"/>
         internal string[]? InAppInclude { get; set; }
 
+        /// <summary>
+        /// When the code was initialized/
```
Tyrrrz, line 97:
```suggestion
        /// When the code was initialized.
```

src/Sentry/SentryOptionsExtensions.cs
```diff
@@ -236,5 +237,30 @@ internal static void SetupLogging(this SentryOptions options)
...
+                    {
+                        options.StartupTime = proc.StartTime.ToUniversalTime();
+                    }
+                }
+                finally
+                {
+                    //Ignore any exception and stay with the fallback DateTime.UtcNow value.
+                }
+            }).ConfigureAwait(false);
+        }
```
Tyrrrz, line 264: Why do we need it on `SentryOptions`?
Can't we just have it lazily initialized on `Contexts.App` instead?

src/Sentry/SentryOptions.cs
bruno-garcia: Always

src/Sentry/Internal/Enricher.cs
```diff
@@ -90,6 +91,12 @@ public void Apply(IEventLike eventLike)
                 eventLike.User.IpAddress ??= "{{auto}}";
             }
 
+            //Apply App startup time
+            eventLike.Contexts.App.StartTime ??= _options.StartupTime;
```
bruno-garcia, line 95: Odd that this value is in the options. Can we have this is some internal type instead?

src/Sentry/SentryOptionsExtensions.cs
bruno-garcia: Also not a fan of extending SentryOptions, doesn't belong here.

Could be its own internal class, something like `ProcessInfo`.

It can have a property `DateTimeOffset StartupTime` that starts off with the `DateTimeOffset.UtcNow` (set inlined to the propert decalaration) but does the Task.Run in the  constructor to set the "better" value afterwards.

Please add a catch block and suppress all exceptions (so it doesn't go into UnobservedTaskException), ideally logging with `options.Logger?.LogError`.

You can add a `static ProcessInfo` field to `SentrySdk` that gets set when `Init` is called (so the UtcNow is read as early as possible: i.e: Now when the first event goes out) and increases the chance that an error early in the process uses the "worst" value.

Lastly, this gives a chance of introducing an opt-out flag on `SentryOptions` later if this becomes a problem (Task.Run on Unity or something), so we can check for this flag before instantiating this class during SentrySdk.Init.

bruno-garcia (changes requested):
Thanks for adding this! I think the general technical solution makes sense but where its layed out in the project needs adjusting

src/Sentry/SentryOptionsExtensions.cs
Tyrrrz: I wonder why we need the task at all, is it not possible to just use `Lazy<>`?

```csharp
var startupTime = new Lazy<DateTimeOffset>(() =>
{
      using var proc = Process.GetCurrentProcess();
});
```

lucas-zimerman:
All the code related to BootTime and StartupTime moved to ProcessInfo class.

src/Sentry/SentryOptionsExtensions.cs
lucas-zimerman: How would Lazy work in this scenario?
StartupTime is initially set with UtcNow. After that, it'll be set only once with a more precise Timer.

The task is there to save some ms from the calling thread since GetCurrentProcess takes ~80ms to be executed.
I'd agree that Lazy would make sense if StartupTime was only set with proc.StartTime.ToUniversalTime, but we are initializing it with DateTimeOffset.UtcNow and also the Process StartTime.

src/Sentry/Internal/ProcessInfo.cs
```diff
@@ -0,0 +1,57 @@
...
+            _ = Task.Run(() =>
+            {
+                try
+                {
+                    using (var proc = Process.GetCurrentProcess())
+                    {
+                        StartupTime = proc.StartTime.ToUniversalTime();
+                    }
+                }
+                catch(Exception e)
```
bruno-garcia:
```suggestion
                catch (Exception e)
```

src/Sentry/Internal/ProcessInfo.cs
```diff
@@ -0,0 +1,57 @@
...
+        {
+            _ = Task.Run(() =>
+            {
+                try
+                {
+                    using (var proc = Process.GetCurrentProcess())
+                    {
+                        StartupTime = proc.StartTime.ToUniversalTime();
+                    }
+                }
```
bruno-garcia, line 47:
```suggestion
                    using var proc = Process.GetCurrentProcess();
                    StartupTime = proc.StartTime.ToUniversalTime();
                }
```

src/Sentry/SentrySdk.cs
```diff
@@ -33,7 +33,8 @@ internal static IHub InitHub(SentryOptions options)
             // Side-effects in a factory function 🤮
             options.SetupLogging();
 
-            options.SetupStartupTime();
+            ProcessInfo.Instance ??= new ProcessInfo(options);
+            ProcessInfo.Instance.SetupStartupTime();
```
bruno-garcia, line 36: Best to call `SetupStartupTime` only if a new instance was created. Also best to use the local reference to call it since, in theory, `ProcessInfo.Instance` could be null between assigning the reference to it and calling `SetupStartupTime` in the event someone called SentrySdk.Close()` or any other place in the future.

```suggestion
            if (ProcessInfo.Instance is null)
            {
            	var processInfo = new ProcessInfo(options);
            	processInfo.SetupStartupTime();
            	ProcessInfo.Instance = processInfo;
          	}
```

src/Sentry/Internal/ProcessInfo.cs
```diff
@@ -0,0 +1,57 @@
...
+            BootTime = DateTimeOffset.UtcNow - TimeSpan.FromTicks(Stopwatch.GetTimestamp());
+        }
+
+        /// <summary>
+        /// By default, the StartupTime is set once SentryOptions is created.
+        /// That way is computationally cheap but not so precise.
+        /// You call this extension to give a better precision to the StartupTime at a cost of a small overhead to current
+        /// Thread.
+        /// </summary>
+        internal void SetupStartupTime()
```
bruno-garcia, line 37: What about something like:
```suggestion
        /// </summary>
        internal void StartAccurateStartupTime()
```

src/Sentry/SentrySdk.cs
```diff
@@ -124,6 +125,7 @@ internal static IDisposable UseHub(IHub hub)
         public static void Close()
         {
             var oldHub = Interlocked.Exchange(ref _hub, DisabledHub.Instance);
+            ProcessInfo.Instance = null;
```
bruno-garcia, line 127: Better do to this after disposing the Hub since on Dispose events are still being processed and flushed out to Sentry.

test/Sentry.Tests/Internals/ProcessInfoTests.cs
```diff
@@ -0,0 +1,51 @@
...
+using Sentry.Internal;
+using Sentry.Testing;
+using Xunit;
+
+namespace Sentry.Tests.Internals
+{
+    public class ProcessInfoTests
+    {
+        [Fact]
+        public async void SetupStartupTime_StartupTimeSet()
```
bruno-garcia, line 12: If you return `void` xunit can't await for your test.
```suggestion
        public async Task SetupStartupTime_StartupTimeSet()
```

Does it fail if you change `Assert.True` to `Assert.False`?

test/Sentry.Tests/Internals/ProcessInfoTests.cs
```diff
@@ -0,0 +1,51 @@
...
+            //Act
+            processInfo.SetupStartupTime();
+
+            //Assert
+            Assert.True(await func.WaitConditionAsync(true, TimeSpan.FromSeconds(1)));
+        }
+
+
+        [Fact]
+        public async void SetupStartupTime_MultipleCalls_DoesntCrash()
```
bruno-garcia, line 30:
```suggestion
        public async Task SetupStartupTime_MultipleCalls_DoesntCrash()
```

test/Sentry.Tests/Internals/ProcessInfoTests.cs
```diff
@@ -0,0 +1,51 @@
...
+
+            //Act
+            for (int i = 0; i < 10; i++)
+            { 
+                processInfo.SetupStartupTime();
+            }
+
+            //Assert
+            Assert.True(await func.WaitConditionAsync(true, TimeSpan.FromSeconds(1)));
+            await Task.Delay(200);
```
bruno-garcia, line 47:
```suggestion
```

test/Sentry.Tests/Internals/ProcessInfoTests.cs
```diff
@@ -0,0 +1,51 @@
...
+        }
+
```
bruno-garcia, line 49:
```suggestion
```

test/Sentry.Tests/Internals/ProcessInfoTests.cs
```diff
@@ -0,0 +1,51 @@
+using System;
+using System.Threading.Tasks;
+using Sentry.Internal;
+using Sentry.Testing;
+using Xunit;
+
+namespace Sentry.Tests.Internals
+{
+    public class ProcessInfoTests
```
bruno-garcia, line 9: Could you please add a test that asserts a new instance of ProcessInfo starts with a `StartupTime` that's not in the future and it's less than `UtcNow` by up to a few seconds?

src/Sentry/Internal/ProcessInfo.cs
lucas-zimerman: Indeed, StartupTime is already set, the original name doesn't make a lot of sense given the current setup.

test/Sentry.Tests/Internals/ProcessInfoTests.cs
lucas-zimerman: It should fail since it returned True.

lucas-zimerman: TIL

CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
+
+### Changes
```
```suggestion
### Features
```

src/Sentry/Internal/ProcessInfo.cs
```diff
@@ -0,0 +1,55 @@
...
+            _options = options;
+            StartupTime = DateTimeOffset.UtcNow;
+            BootTime = DateTimeOffset.UtcNow - TimeSpan.FromTicks(Stopwatch.GetTimestamp());
+        }
+
+        /// <summary>
+        /// By default, the StartupTime is set once SentryOptions is created.
+        /// That way is computationally cheap but not so precise.
+        /// You call this extension to give a better precision to the StartupTime at a cost of a small overhead to current
+        /// Thread.
```
bruno-garcia, line 35:
```suggestion
        /// StartupTime is set to UtcNow in this constructor.
        /// That's computationally cheap but not very precise.
        /// This method will give a better precision to the StartupTime at a cost
        /// of calling Process.GetCurrentProcess, on a thread pool thread.
```

src/Sentry/Internal/ProcessInfo.cs
```diff
@@ -0,0 +1,55 @@
...
+                    StartupTime = proc.StartTime.ToUniversalTime();
+                }
+                catch (Exception e)
+                {
+                    _options.DiagnosticLogger?.LogError("Failure to GetCurrentProcess", e);
+                    //Ignore any exception and stay with the fallback DateTime.UtcNow value.
+                }
+            }).ConfigureAwait(false);
+        }
+
```
bruno-garcia, line 53:
```suggestion
```

src/Sentry/SentryOptions.cs
```diff
@@ -153,7 +153,7 @@ public class SentryOptions
         /// </example>
         /// <see href="https://develop.sentry.dev/sdk/features/#event-sampling"/>
         private float? _sampleRate;
-
+
```
bruno-garcia, line 156:
```suggestion

```

src/Sentry/SentrySdk.cs
```diff
@@ -32,6 +32,14 @@ internal static IHub InitHub(SentryOptions options)
             // Side-effects in a factory function 🤮
             options.SetupLogging();
 
+            var processInfo = ProcessInfo.Instance;
+            if (processInfo is null)
+            {
+                processInfo = new ProcessInfo(options);
+                ProcessInfo.Instance = processInfo;
+          	}
+            processInfo.StartAccurateStartupTime();
```
bruno-garcia, line 41:
```suggestion
            var processInfo = ProcessInfo.Instance;
            if (processInfo is null)
            {
                processInfo = new ProcessInfo(options);
                processInfo.StartAccurateStartupTime();
                ProcessInfo.Instance = processInfo;
            }
```

test/Sentry.Testing/FuncExtensions.cs
```diff
@@ -0,0 +1,33 @@
...
+        {
+            var now = DateTime.UtcNow;
+            timeout ??= TimeSpan.FromSeconds(10);
+
+            while (now.Add(timeout.Value) >= DateTime.UtcNow && condition() != expectedResult) 
+            {
+                await Task.Delay(250);
+            }
+
+            if(condition() != expectedResult)
```
bruno-garcia, line 25:
```suggestion
                await Task.Delay(50);
            }

            if (condition() != expectedResult)
```

bruno-garcia (approved):

PR #908, opened 2021-03-29
Limit max spans in transaction to 1000

### LINKED ISSUE #907: Limit transactions to 1000 spans
Return no-op spans from `StartChild()` after 1000.

Via https://github.com/getsentry/sentry-java/issues/1356

Tyrrrz:
Closes #907

Initially I wanted to return `DisabledSpan`, but `ISpan` has a lot of properties that would all need to be implemented consistently because the user may want to inspect their values. So instead I opted for a simpler solution which is just to not add the created span to the list. This way it can still maintain its functionality, but it won't be sent to Sentry and will be eligible for immediate garbage collection.

~~Note that `span.Finish(exception)` will still bind the exception, which means some event may have a `traceId`/`spanId` reference to a span that doesn't actually exist. I suppose it's effectively the same as if the span was part of a sampled out transaction, so I guess it's fine.~~

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,11 +1,12 @@
 # Changelog
 
-## Unreleased
+## 3.3.0
 
 ### Features
 
 - Add StartupTime and Device.BootTime (#887) @lucas-zimerman
 - Link events to currently active span (#909) @Tyrrrz
+- Add useful contextual data to TransactionSamplingContext in ASP.NET Core integration (#910) @Tyrrrz
 
 ### Changes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Features
 
 - Add StartupTime and Device.BootTime (#887) @lucas-zimerman
+- Link events to currently active span (#909) @Tyrrrz
 
 ### Changes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 # Changelog
 
+## Unreleased
+
+### Fixes
+
+- Don't dispose the SDK when UnobservedTaskException is captured (#925) @bruno-garcia
+- Fix spans not inheriting TraceId from transaction (#922) @Tyrrrz
+
 ## 3.3.0
 
 ### Features
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@
 
 - Add StartupTime and Device.BootTime (#887) @lucas-zimerman
 
+### Changes
+
+- Limit max spans in transaction to 1000 (#908) @Tyrrrz
+
 ## 3.2.0
 
 ### Changes
--- a/src/Sentry/Internal/Hub.cs
+++ b/src/Sentry/Internal/Hub.cs
@@ -162,8 +162,13 @@ namespace Sentry.Internal
                 return spanBoundToException;
             }
 
-            // Otherwise just get the currently active span on the scope
-            return scope.GetSpan();
+            // Otherwise just get the currently active span on the scope (unless it's sampled out)
+            if (scope.GetSpan() is {IsSampled: not false} span)
+            {
+                return span;
+            }
+
+            return null;
         }
 
         public SentryId CaptureEvent(SentryEvent evt, Scope? scope = null)
--- a/src/Sentry/Internal/Hub.cs
+++ b/src/Sentry/Internal/Hub.cs
@@ -153,6 +153,19 @@ namespace Sentry.Internal
 
         public SentryTraceHeader? GetTraceHeader() => GetSpan()?.GetTraceHeader();
 
+        private ISpan? GetLinkedSpan(SentryEvent evt, Scope scope)
+        {
+            // Find the span which is bound to the same exception
+            if (evt.Exception is { } exception &&
+                _exceptionToSpanMap.TryGetValue(exception, out var spanBoundToException))
+            {
+                return spanBoundToException;
+            }
+
+            // Otherwise just get the currently active span on the scope
+            return scope.GetSpan();
+        }
+
         public SentryId CaptureEvent(SentryEvent evt, Scope? scope = null)
         {
             try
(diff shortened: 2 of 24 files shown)


bruno-garcia:
> Note that span.Finish(exception) will still bind the exception, which means some event may have a traceId/spanId reference to a span that doesn't actually exist. I suppose it's effectively the same as if the span was part of a sampled out transaction, so I guess it's fine.

Can we check `Span.Sampled` before keeping track of it?

bruno-garcia (commented):
Would be nice to cover this with tests

Tyrrrz:
Ok, changed it so now out-of-limit spans are also sampled out. Changed `BindException(...)` to ignore sampled out spans too.

Tyrrrz:
I also had to change `GetLinkedSpan(...)`'s behavior to also ignore sampled out spans, to make the linking consistent.

src/Sentry/Internal/Hub.cs
```diff
@@ -135,6 +135,12 @@ public void WithScope(Action<Scope> scopeCallback)
 
         public void BindException(Exception exception, ISpan span)
         {
+            // Don't bind on sampled out spans
+            if (span.IsSampled == false)
```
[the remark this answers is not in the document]
Tyrrrz: @bruno-garcia this approach does not compile ;)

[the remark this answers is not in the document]
bruno-garcia: Oh, it's nullable

src/Sentry/SentryOptions.cs
```diff
@@ -450,6 +450,27 @@ public double TracesSampleRate
...
         public Func<TransactionSamplingContext, double?>? TracesSampler { get; set; }
 
+        private int _maxSpans = 1000;
+
+        /// <summary>
+        /// Maximum spans allowed in a transaction.
+        /// Attempting to add more spans to a transaction than this limit dictates
+        /// will result in the excess spans being silently dropped.
+        /// </summary>
+        public int MaxSpans
```
bruno-garcia, line 460: Thanks for pointing out this is experimental in Python.
In hindsight we're be stuck with this setting even if we move towards collecting spans only where this would become a confusing configuration option. 

I propose we just hide this for now. @Tyrrrz @marandaneto what do you think? 

```suggestion
        internal int MaxSpans
```

bruno-garcia (commented):
Lets agree on the naming for the option, of if we add it at all given this is not really aligning to anything (experimental on Python) and the fact we want to move to collecting spans outside of a transaction.

Other than that LGTM

Tyrrrz:
@bruno-garcia I don't mind either way. One thing to note is that we need to inject `SentryOptions` to `TransactionTracer` in order to make this configurable. It is technically a breaking change, so probably the sooner we do it the better. However, I doubt anyone is newing up `TransactionTracer` by hand. Other than that, I don't really care.

As a user, I doubt I would want to configure it though. But I've not used the product enough to know for sure.

src/Sentry/SentryOptions.cs
marandaneto: on Java we'llll have an annotation for it.

https://github.com/JetBrains/java-annotations/blob/master/common/src/main/java/org/jetbrains/annotations/ApiStatus.java#L34-L54

which is pretty much the same as Python, but via annotation processing, can we do that instead?
I don't see being useful if its there but hidden, while adding such annotation let us know the possibility of removing later if it was a wrong move, wdyt?

Tyrrrz: I'm personally skeptical about adding public members just to annotate them with "don't use it, it's internal". I can see it as a way to workaround language limitations, like where you need to define a type/member as public because the language requires it to be, rather than because it actually needs to be used publicly. For example, C# doesn't allow public types to implement internal interfaces implicitly, which is a case where "public but not really" annotation would be nice.

In this case, I feel like we'd be overthinking/overengineering a solution to a non-existent problem. Are there any users who actually want to set `MaxSpans` today?

In summary: I think we should either make this option public (and agree on the name) or just not add it at all. In case we choose not to add it, we can still keep `SentryOptions` in `TransactionTracer` for the future.

bruno-garcia: I'm OK with getting rid of it

bruno-garcia:
Lets avoid the breaking change and leave this without a configuration option for now

Tyrrrz:
@bruno-garcia reverted

CHANGELOG.md
```diff
@@ -8,6 +8,10 @@
 - Link events to currently active span (#909) @Tyrrrz
 - Add useful contextual data to TransactionSamplingContext in ASP.NET Core integration (#910) @Tyrrrz
 
+### Changes
```
bruno-garcia, line 11: "feature" or "bug fix"? :)

```suggestion
### Features
```

src/Sentry/TransactionTracer.cs
```diff
@@ -194,12 +194,20 @@ public TransactionTracer(IHub hub, ITransactionContext context)
...
+                IsSampled = !isOutOfLimit
+                    ? IsSampled
+                    : false // sample out out-of-limit spans
             };
 
-            _spans.Add(span);
+            if (!isOutOfLimit)
+            {
+                _spans.Add(span);
+            }
```
bruno-garcia, line 210: Can we please log that this happened?
```suggestion
            }
            else
            {
            _hub?.Logger?.LogDebug("Span {0} dropped due to limit reached", span.Description);
```

It'll help users figure out what's going on if "my spans don't show up".

the `_hub?.Logger` is a suggestion, maybe best to just pass the options (since we can swap the logger instance at runtime)

bruno-garcia (commented):
LGTM.

Would be nice to log out things there this is a good candidate for customer issues to be raised.

bruno-garcia (approved):
Lets merge as-is but some logging could be beneficial on a follow up PR

bruno-garcia:
`develop` discussion about this: https://github.com/getsentry/develop/issues/298

PR #909, opened 2021-03-29
Link events to active span

### LINKED ISSUE #851: When capturing events with exception, link to current trace
Regarding: https://github.com/getsentry/sentry-dotnet/pull/848#discussion_r582439193

If the current Span is already mapped to a different exception (via `BindException`/`span.Finish(ex)`), what should happen?

Also, if the current span on scope is the root transaction, but the exception is bound to some nested Span (or maybe even Span that's not part of the bound Transaction), what should happen?

There are probably other similar edge cases.

@bruno-garcia

Tyrrrz:
Closes #851

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Features
 
 - Add StartupTime and Device.BootTime (#887) @lucas-zimerman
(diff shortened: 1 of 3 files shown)


bruno-garcia (approved):
Nice!

PR #910, opened 2021-03-30
Add useful contextual data to TransactionSamplingContext in ASP.NET Core integration

### LINKED ISSUE #814: Put useful request info into CustomSamplingContext
Currently, it's difficult to effectively run dynamic sampler in ASP.NET Core because the integration does not expose useful data such as the request URL in the `TransactionSamplingContext`.

See #776 and #799

The suggestion is to make it so that `SentryTracingMiddleware` will create a custom sampling context and put this data there. To make it easier to consume for the users, we can expose extension methods on `TransactionSamplingContext` such as `ctx.TryGetRequestUrl()`, etc.

Tyrrrz:
Closes #814 

I added only request path and route name, because those were the most commonly requested ones. If there are any other suggestions, I'm all ears.

It's also possible to add the whole `HttpContext`, however it becomes unusable once it's disposed. Given that the sampling should occur while the request is still underway, it may be fine, but I'm still weary of putting too much rather than too little in this case. Thoughts?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 
 - Add StartupTime and Device.BootTime (#887) @lucas-zimerman
 - Link events to currently active span (#909) @Tyrrrz
+- Add useful contextual data to TransactionSamplingContext in ASP.NET Core integration (#910) @Tyrrrz
 
 ## 3.2.0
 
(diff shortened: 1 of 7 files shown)


bruno-garcia (commented):
Cool stuff. Users will need some docs to see this in action.

Also, please add this to the one of more of the samples?

bruno-garcia (approved):

bruno-garcia:
Thanks @Tyrrrz !

PR #922, opened 2021-04-05
Inherit TraceId from transaction in spans

### LINKED ISSUE #920: SpanTracer crates a new Trace for each span:
https://github.com/getsentry/sentry-dotnet/blob/a4ebfeb3e63ff27a7b1f432761050e4c7ef448ac/src/Sentry/SpanTracer.cs#L77

Spans part of the same transaction should be part of the same trace.

A trace can contain one or more transactions.

image

Tyrrrz:
Fixes #920

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
+
+### Fixes
+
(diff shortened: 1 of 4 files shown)


bruno-garcia (approved):
Thanks!

PR #924, opened 2021-04-05
Move Description field from Transaction to Trace context

### LINKED ISSUE #913: `description` on `StartTransaction` results in a Sentry processing error
The overload:

`SentrySdk.StartTransaction("name", "op", "description");`

Results in  a server error:
image

Event doesn't have a top level field `description` so it makes sense. This could be a field in Span which we, trying to align 'startTransaction' with a 'transaction is only a span with a name' approach ended up adding to event/transation.

Tyrrrz:
Fixes #913

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
+
+### Changes
+
+- Move Description field from Transaction to Trace context (#924) @Tyrrrz
+
 ## 3.3.0
 
 ### Features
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
+
+### Changes
```
bruno-garcia, line 5: When versioning it's easier to see if we have only "bug fixes" or also "features" to chose to bump a minor or patch:

```suggestion
### Fixes
```

bruno-garcia (commented):
Looks good.
Could you please add a test to validate the code?

PR #928, opened 2021-04-07
Don't override Span/Transaction status on Finish(...) if status was not provided explicitly

### LINKED ISSUE #919: transaction.Finish() overrides the transaction status set via the public setter with the default parameter (Ok).
repro:

var tran = startTransaction();
tran.Status = SpanStatus.Error;
tran.Finish();

That's due to:
public void Finish(SpanStatus status = SpanStatus.Ok)

Suggestion: Have a proper parameterless method.
Also, don't set Ok if nothing was set. Sentry has a status: 'unknown' when nothing is set. Verify the spec/other SDKs if they default to OK or just send it without status.

Tyrrrz:
Fixes #919

### DIFF
--- a/test/Sentry.Tests/HubTests.cs
+++ b/test/Sentry.Tests/HubTests.cs
@@ -71,7 +71,7 @@ namespace NotSentry.Tests
             hub.ConfigureScope(s => Assert.False(s.Locked));
         }
 
-        [Fact]
+        [Fact(Skip = "Flaky")]
         public void CaptureMessage_AttachStacktraceFalse_DoesNotIncludeStackTrace()
         {
             // Arrange
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
+
+### Fixes
+
+- Don't override Span/Transaction status on Finish(...) if status was not provided explicitly (#928) @Tyrrrz
+
 ## 3.3.1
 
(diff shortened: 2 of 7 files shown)


test/Sentry.Tests/HubTests.cs
```diff
@@ -71,7 +71,7 @@ public void PushAndLockScope_DoesNotAffectOuterScope()
             hub.ConfigureScope(s => Assert.False(s.Locked));
         }
 
-        [Fact]
+        [Fact(Skip = "Flaky")]
```
bruno-garcia, line 74: Best to raise issues to address these later, if there are no issues yet

bruno-garcia (approved):
We need a big note on the release notes once we ship since this will affect transactions in Sentry. Currently users would be getting OK for simply calling `finish` now it'll be `unknown`.

A Migration note on the docs will be needed to: From version x to y: Before: Finish() -> Now you need to explicitly call Finish(Ok)

test/Sentry.Tests/HubTests.cs
Tyrrrz: There are issues about specific tests, maybe one of them can be generalized.

bruno-garcia:
can be merged?

bruno-garcia:
develop discussion about this: https://github.com/getsentry/develop/issues/298

Tyrrrz:
@bruno-garcia yes

Issue #941, opened 2021-04-15
Allow to serialize events


pchinery:
We are trying to adapt Sentry for our product, but we have a slightly different workflow than the usual "Sentry way", as we do not publish events immediately, but we allow our users to review them and send them afterwards. As this can be from within a service, we have to serialize the events in some way and then deserialize them to present them to the user as well as to send them to Sentry.

Right now, the `WriteTo` method from `IJsonSerializable` is marked as "do not use / only for internal reasons" and this warning probably is well justified.

We are using it anyway right now, as this is the only easy way to serialize this completely. To make it even worse, we have to parse the `Exception` to `SentryExceptions`. Due to `MainExceptionProcessor` being internal (again probably for good reasons), we have copied the class to our project.

I'd like to explore if opening these paths a little bit is something you think is worth exploring, as we would like to stick to this workflow and think this could be of benefit for others as well.

If you think this is something you would bring into the library, we would be happy to create a PR for this.

bruno-garcia:
Wouldn't it be a better hook for your to implement `ITransport` instead? That happens after these callbacks such as EventProcessor, so the actual data sent to the user will be there, complete.

In your custom transport you could serialize the envelope, to display, and if approved by the user, you can just get an envelope object back, from the serialized form and send it to the `SentryHttpTransport`.

pchinery:
Right now, this is beyond my understanding of the Sentry library, but it sounds quite promising. I will play a bit with this today. Thank you very much for pointing me in this direction, I will follow up here later.

pchinery:
After doing some research, this approach sounds very good, as it does not require to do any custom serialization and we would not have to use the Sentry lib in an unintended way. Right now, `ITransport` is internal though and the Transport option is internal as well. Would it be the way to go to make this public? Or do you prefer a factory instead, like with `SentryStackTraceFactory`?

bruno-garcia:
Seems like it was made internal here: https://github.com/getsentry/sentry-dotnet/commit/2e3a433039fbdf70d0c8b327cd02dc2a7afd1c12#diff-51893aa74e97f7920707e0e7eafac7c116d293e5bddbc3f2c229d608da562c35R10

This is designed to be public API so looks like an oversight on our end. @Tyrrrz thoughts?

Tyrrrz:
[...]

PR #944, opened 2021-04-19
Allow to configure the ITransport to support custom serialization scenarios

pchinery:
As discussed in #941, this is a proposal to make the `ITransport` configurable, which required to make the interface and some other classes along the path public.

The transport will be wrapped with caching, if configured.


src/Sentry/SentryOptions.cs
```diff
@@ -21,8 +22,11 @@ public class SentryOptions
     {
         private Dictionary<string, string>? _defaultTags;
 
-        // Override for tests
-        internal ITransport? Transport { get; set; }
+        /// <summary>
+        /// The <see cref="ITransport"/> to use for sending the events. If null, the <see cref="HttpTransport"/> will be used.
+        /// The Transport will be wrapped with a <see cref="CachingTransport"/>, if a cache directory is set.
```
bruno-garcia, line 27:
```suggestion
        /// The Transport will be wrapped with a <see cref="CachingTransport"/>, if a <see cref="CacheDirectoryPath"/> is set.
```

src/Sentry/Internal/SdkComposer.cs
```diff
@@ -12,17 +12,12 @@ internal class SdkComposer
         public SdkComposer(SentryOptions options)
         {
             _options = options ?? throw new ArgumentNullException(nameof(options));
-            if (options.Dsn is null) throw new ArgumentException("No DSN defined in the SentryOptions");
+            if (options.Dsn is null)
+                throw new ArgumentException("No DSN defined in the SentryOptions");
```
bruno-garcia, line 16:
```suggestion
            if (options.Dsn is null) throw new ArgumentException("No DSN defined in the SentryOptions");
```

src/Sentry/Envelopes/ISerializable.cs
```diff
@@ -7,7 +7,7 @@ namespace Sentry.Protocol.Envelopes
     /// <summary>
     /// Represents a serializable entity.
     /// </summary>
-    internal interface ISerializable
+    public interface ISerializable
```
bruno-garcia, line 10: I think this was another reason @Tyrrrz wanted to make things internal.
He had some plans to refactor this so types don't need to implement this in order to be serialized.

bruno-garcia (commented):
Code looks good but it's missing some tests to cover the new behavior.

One outstanding point is the fact we need to make `ISerializable` also public. @Tyrrrz won't be back for 3 weeks to chime in but I believe he wanted to refactor things before we make those public. So either we'll wait or we need to figure this out ourselves

pchinery:
I have now added a test to verify that custom transports a wrapped with a cache. The opposite (to show that no caching is used) would require either to make some properties public on the `BackgroundWorker` or would require some reflection to verify this. Do you have a preference for any of the two ways or is it good as it is?

src/Sentry/Envelopes/ISerializable.cs
pchinery: That's most likely something that goes beyond of my understanding of the code vision of the library right now. I'm happy to help, but if there was an idea how to do this already, I might not think in the direction @Tyrrrz wanted to bring this, so I think it's better to wait for him.

src/Sentry/Internal/SdkComposer.cs
```diff
@@ -34,16 +28,30 @@ private ITransport CreateTransport()
...
+            ITransport transport;
+
+            if (_options.Transport is not null)
+            {
+                transport = _options.Transport;
+            }
+            else
+            {
+                transport = CreateHttpTransport();
+            }
```
Tyrrrz, line 45:
```suggestion
            var transport = _options.Transport ?? CreateHttpTransport();
```

test/Sentry.Tests/SentrySdkTests.cs
```diff
@@ -270,6 +271,32 @@ public async Task Init_WithCache_BlocksUntilExistingCacheIsFlushed()
...
+            // Act
+
+            SentrySdk.CaptureMessage("some message");
+            await SentrySdk.FlushAsync(TimeSpan.FromSeconds(5));
+
+            // Assert
+            Directory
+                .EnumerateFiles(cacheDirectory.Path, "*", SearchOption.AllDirectories)
+                .ToArray()
+                .Should().NotBeEmpty();
```
Tyrrrz, line 297: These tests are failing, but if you merge in the latest main I think they should work

src/Sentry/Envelopes/ISerializable.cs
Tyrrrz: As a temp workaround to avoid exposing `ISerializable`, let's make:

- `EnvelopeItem.Payload` internal (or even private? can we do that?)
- `EnvelopeItem.ctor` internal (we won't expose a public constructor for envelope)

It's ugly but I think exposing `ISerializable` right now will constrain us too much. I don't like exposing `Envelope` either but there's no way around it if we want to expose `ITransport`.

Alexandr-Pletnev:
> As discussed in #941, this is a proposal to make the `ITransport` configurable, which required to make the interface and some other classes along the path public.
> 
> The transport will be wrapped with caching, if configured.

It's great feature! Thanks.

Could you help me in one use-case?

I have to develop a use-case:  Send the Sentry Events from a privite side throught RabbitMq to a public side and then send them to Sentry Server.  

What I did:
1. I developed own ITransport that send an Envelope thought RabbitMq.
2. I developed the RabbitMq consumer that recive the Envelope. Everything ok.
3. I need to send the Envelope to the Sentry Server. But i can't realise how i can do it.

I will plesant if you help me.

Tyrrrz:
> > As discussed in #941, this is a proposal to make the `ITransport` configurable, which required to make the interface and some other classes along the path public.
> > The transport will be wrapped with caching, if configured.
> 
> It's great feature! Thanks.
> 
> Could you help me in one use-case?
> 
> I have to develop a use-case: Send the Sentry Events from a private side throught RabbitMq to a public side and then send them to Sentry Server.
> 
> What I did:
> 
> 1. I developed own ITransport that send an Envelope thought RabbitMq.
> 2. I developed the RabbitMq consumer that receive the Envelope. Everything ok.
> 3. I need to send the Envelope to the Sentry Server. But i can't realise how i can do it.
> 
> I will plesant if you help me.

I don't think there's a straightforward way to achieve that (@bruno-garcia can correct me if I'm wrong).

I think your best bet is to use write the envelopes from RabbitMQ to a local directory, and pick them up via caching feature: https://github.com/getsentry/sentry-dotnet/blob/984b21063ca806df0129a89aeec011f95012eca5/src/Sentry/SentryOptions.cs#L385-L389

The files will need to be placed in a specific subdirectory as well. Assuming `SentryOptions.CacheDirectoryPath = "path/to/some/directory/"`, the files would need to be seeded in:

```
path/to/some/directory/Sentry/{your DSN}/
```

The file name doesn't matter, but the extension has to be `.envelope`. You can inspect the implementation here for more info: https://github.com/getsentry/sentry-dotnet/blob/984b21063ca806df0129a89aeec011f95012eca5/src/Sentry/Internal/Http/CachingTransport.cs

Cached envelopes are flushed automatically during `SentrySdk.Init(...)` but can also be flushed manually through `SentrySdk.FlushAsync(...)` (I think).

That said, all of this is very hacky and I'm not sure how well this will work. We probably need a proper solution for this, but currently all of the relevant interfaces and implementations are internal.

Alexandr-Pletnev:
Thank You for your reply. 

I decided to achieve this via reflections like that:

`FieldInfo fiHub = typeof(SentrySdk).GetField("_hub", BindingFlags.NonPublic | BindingFlags.Static);
object hub = fiHub.GetValue(null);

FieldInfo fiClient = hub.GetType().GetField("_ownedClient", BindingFlags.NonPublic | BindingFlags.Instance);
object client = fiClient.GetValue(hub);


object result = null;

var mi = client.GetType().GetMethod("CaptureEnvelope", BindingFlags.NonPublic | BindingFlags.Instance);
if (mi != null)
{
	result = mi.Invoke(client, new object[] { e });
}`

mattjohnsonpint:
Hi.  Sorry this PR is open for so long.  I'm taking a look now, and it seems like *most* of what it does has already been addressed by other PRs for other reasons, as follows:

- Making `Envelope`, `EnvelopeItem`, and `ISerializable` public was done with #1450, released in 3.14.0.
- Making `ITransport` and `options.Transport` public was *not* done, but rather we extracted most of the logic from `HttpTransport` into the public abstract `HttpTransportBase` class with #1560, released in 3.16.0.  This was done primarily to support Unity WebGL.
- Ensuring `CreateHttpTransport` correctly uses `options.Transport` when set, and wraps with `CachingTransport` properly was done with #1589 (unreleased as of today)

I think to achieve the stated goals, all that needs to be done is to make `ITransport` and `options.Transport` public.  That would allow others to create their own transports (extending from `HttpTransportBase` if desired).  Unity didn't need that, because they also needed to create their own implementation of `IBackgroundWorker` - which is public.  But we shouldn't require others to do that just to attach a transport.

I'll close this PR and open a new one with that.

mattjohnsonpint:
See #1602

PR #948, opened 2021-04-20
ref startup time detection

bruno-garcia:


src/Sentry/Internal/ProcessInfo.cs
```diff
@@ -12,43 +14,68 @@ internal class ProcessInfo
...
         {
             _options = options;
-            StartupTime = DateTimeOffset.UtcNow;
-            BootTime = DateTimeOffset.UtcNow - TimeSpan.FromTicks(Stopwatch.GetTimestamp());
-        }
+            _findPreciseStartupTime = findPreciseStartupTime ?? GetStartupTime;
+            if (options.DetectStartupTime == StartupTimeDetectionMode.None)
+            {
+                _options.DiagnosticLogger?.LogDebug("Not detecting startup time due to options: {0}",
+                    options.DetectStartupTime);
```
lucas-zimerman, line 39: Is it bad if you use _options.DetectStartupTime? _options on top of options look different 😅

bruno-garcia: Good point. Will fix

PR #960, opened 2021-04-28
test code to test asm

bruno-garcia:
Serializing it all into a memory stream and then getting a string out of it is test code and should live in the test assembly.

#skip-changelog


bruno-garcia:
#skip-changelog

PR #996, opened 2021-05-13
Add StartSentryTransaction() to Sentry.AspNet

### LINKED ISSUE #981: Performance and tracing for ASP.NET
We already have an example hooking into Global.asax.cs:

https://github.com/getsentry/examples/blob/master/dotnet/AspNetMvc5Ef6/Global.asax.cs#L53-L97

That's missing the inheritance of the trace-id and sampling decision. 

We also need to add that to the docs:

https://docs.sentry.io/platforms/dotnet/guides/aspnet/performance/included-instrumentation/

Tyrrrz:
Closes #981

### DIFF
--- a/src/Sentry.AspNet/HttpContextExtensions.cs
+++ b/src/Sentry.AspNet/HttpContextExtensions.cs
@@ -1,0 +1,48 @@
+﻿using System.Collections.Specialized;
+using System.Web;
+
+namespace Sentry.AspNet
+{
+    /// <summary>
+    /// Sentry extensions for <see cref="HttpContext"/>.
+    /// </summary>
+    public static class HttpContextExtensions
+    {
+        private static SentryTraceHeader? TryGetTraceHeader(NameValueCollection headers)
+        {
+            try
+            {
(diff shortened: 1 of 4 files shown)


src/Sentry.AspNet/HttpContextExtensions.cs
```diff
@@ -0,0 +1,48 @@
...
+    /// Sentry extensions for <see cref="HttpContext"/>.
+    /// </summary>
+    public static class HttpContextExtensions
+    {
+        private static SentryTraceHeader? TryGetTraceHeader(NameValueCollection headers)
+        {
+            try
+            {
+                var traceHeader = headers.Get(SentryTraceHeader.HttpHeaderName);
+                return SentryTraceHeader.Parse(traceHeader);
```
bruno-garcia, line 16: Next up is doing a TryParse because we can'ta ssume every request will come from an instrumented SDK so this will likely happen a lot (throw)

bruno-garcia (approved):
Nice!

PR #998, opened 2021-05-16
Add support for Serilog.Formatting.ITextFormatter

Genteure:
This feature is needed because sometimes a identifier of some kind (example: user id, task id) is added via `Serilog.ILogger.ForContext()` so it's not visible in the log message body.
Currently there's no option to control how sentry formats received log events.
Most Serilog sinks that output texts have a `Serilog.Formatting.ITextFormatter formatter` or `string outputTemplate` argument so additional information can be added to the result log text.

This pull request is adding more arguments to public methods thus should be considered as a breaking change.

### DIFF
--- a/src/Sentry.Serilog/SentrySink.cs
+++ b/src/Sentry.Serilog/SentrySink.cs
@@ -77,17 +77,7 @@ namespace Sentry.Serilog
 
             var exception = logEvent.Exception;
             var template = logEvent.MessageTemplate.Text;
-            string formatted;
-            if (_options.TextFormatter is { } formatter)
-            {
-                using var stringWriter = new StringWriter(new StringBuilder());
-                formatter.Format(logEvent, stringWriter);
-                formatted = stringWriter.ToString();
-            }
-            else
-            {
-                formatted = logEvent.RenderMessage(_options.FormatProvider);
-            }
+            var formatted = FormatLogEvent(logEvent);
 
             if (logEvent.Level >= _options.MinimumEventLevel)
             {
@@ -140,6 +130,20 @@ namespace Sentry.Serilog
                     context,
                     data: data,
                     level: logEvent.Level.ToBreadcrumbLevel());
+            }
+        }
+
+        private string FormatLogEvent(LogEvent logEvent)
+        {
+            if (_options.TextFormatter is { } formatter)
+            {
+                using var stringWriter = new StringWriter();
+                formatter.Format(logEvent, stringWriter);
+                return stringWriter.ToString();
+            }
+            else
+            {
+                return logEvent.RenderMessage(_options.FormatProvider);
             }
         }
 
--- a/src/Sentry.Serilog/SentrySink.cs
+++ b/src/Sentry.Serilog/SentrySink.cs
@@ -1,5 +1,7 @@
 using System;
 using System.Collections.Generic;
+using System.IO;
+using System.Text;
 using Sentry.Extensibility;
 using Sentry.Infrastructure;
 using Sentry.Protocol;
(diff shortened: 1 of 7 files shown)


src/Sentry.Serilog/SentrySinkExtensions.cs
```diff
@@ -213,6 +219,11 @@ public static class SentrySinkExtensions
                 sentrySerilogOptions.FormatProvider = formatProvider;
             }
 
+            if(textFormatter != null)
```
bruno-garcia, line 222: nit:
```suggestion
            if (textFormatter != null)
```

src/Sentry.Serilog/SentrySink.cs
```diff
@@ -74,8 +76,18 @@ public void Emit(LogEvent logEvent)
...
+            if (_options.TextFormatter is null)
+            {
+                formatted = logEvent.RenderMessage(_options.FormatProvider);
+            }
+            else
+            {
+                var stringWriter = new StringWriter(new StringBuilder());
+                _options.TextFormatter.Format(logEvent, stringWriter);
+                formatted = stringWriter.ToString();
+            }
```
bruno-garcia, line 90: since `TextFormatter` is mutable, another way to write this could be:
```suggestion
            if (_options.TextFormatter is { } formatter)
            {
                using var stringWriter = new StringWriter(new StringBuilder());
                formatter.Format(logEvent, stringWriter);
                formatted = stringWriter.ToString();
            }
            else
            {
                formatted = logEvent.RenderMessage(_options.FormatProvider);
            }
```

bruno-garcia (commented):
Thanks for adding this!
Before we can merge this could you please add a test that uses the `TextFormatter`?
There's also a sample in the repo: https://github.com/getsentry/sentry-dotnet/tree/main/samples/Sentry.Samples.Serilog

And one with ASP.NET Core: https://github.com/getsentry/sentry-dotnet/tree/main/samples/Sentry.Samples.AspNetCore.Serilog

Would you please also use it in one of the samples to help show its usage?

bruno-garcia (approved):
Looks great!

The only thing is that as a breaking change this should land on our next major.

@Tyrrrz thoughts?

bruno-garcia:
Oh @Genteure could you please add a changelog entry?

Genteure:
Sure. Which version should I put this change under? From `dangerfile.js` I think it should be "Unreleased"?

bruno-garcia:
Right, I don't know why it didn't write the comment here but it should be something like

```
## Unreleased

### Features

- Serilog: Add support for Serilog.Formatting.ITextFormatter [#998](https://github.com/getsentry/sentry-dotnet/pull/998)
```

src/Sentry.Serilog/SentrySink.cs
```diff
@@ -74,8 +76,18 @@ public void Emit(LogEvent logEvent)
...
+            if (_options.TextFormatter is { } formatter)
+            {
+                using var stringWriter = new StringWriter(new StringBuilder());
+                formatter.Format(logEvent, stringWriter);
+                formatted = stringWriter.ToString();
+            }
+            else
+            {
+                formatted = logEvent.RenderMessage(_options.FormatProvider);
+            }
```
Tyrrrz, line 90: Is there something like a `PlainTextFormatter` that we can use? So that we could:

```csharp
var formatter = _options.TextFormatter ?? PlainTextFormatter.Instance;
var formatted = ...
```

If not, maybe we can create one?

Finally, if none of those work, can we at least move this part to a separate function like `string Format(LogEvent logEvent)` and replace variable mutations with return statements instead?

Tyrrrz: Another suggestion is to create an internal extension method for `ITextFormatter`:

```csharp
static string Format(this LogEvent logEvent) { /* ... */ }
```

which would use `StringWriter`. The purpose is to move this infrastructural complexity out of this method to make things a bit simpler.

Tyrrrz:
Looks good, I only have one nit

src/Sentry.Serilog/SentrySink.cs
Genteure: I don't think there's a `PlainTextFormatter` available.
We can create a `PlainTextFormatter` or `FallbackTextFormatter` that take a `SentrySerilogOptions` in the constructor since `SentrySerilogOptions.FormatProvider` is also mutable.
Would you want me to make the changes?

Tyrrrz: It sounds like it would be too much work, let's go with just extracting into a separate method

Tyrrrz (approved):

Tyrrrz:
Thanks @Genteure 🎉

PR #1000, opened 2021-05-17
Adjust paramer type in AddBreadcrumb

jairbubbles:
Always better to use a concrete type, also it's now consistent with the other extension method.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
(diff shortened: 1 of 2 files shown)


Tyrrrz:
Looks good @jairbubbles.
Can you please add a quick changelog entry? Should be something like this (in `## Unreleased` section):

```
- Adjust paramer type in `AddBreadcrumb` to use `IReadOnlyDictionary<...>` instead of `Dictionary<...>` [#1000](https://github.com/getsentry/sentry-dotnet/pull/1000)
```

jairbubbles:
@Tyrrrz  Done!

Tyrrrz:
Thanks!

bruno-garcia:
Thanks @jairbubbles !

PR #1010, opened 2021-05-24
simplify ifdef

bruno-garcia:


src/Sentry/PlatformAbstractions/RuntimeInfo.cs
```diff
@@ -22,9 +22,9 @@ internal static Runtime GetRuntime()
 
             runtime ??= GetFromEnvironmentVariable();
 
-#if NETFX
+#if NET461
             SetNetFxReleaseAndVersion(runtime);
-#elif NETSTANDARD || NETCOREAPP // Possibly .NET Core
+#else
```
bruno-garcia, line 27: I think we had a bug here since `NETCOREAPP` was not defined in the csproj

src/Sentry/PlatformAbstractions/RuntimeInfo.cs
```diff
@@ -75,15 +75,15 @@ internal static void SetNetFxReleaseAndVersion(Runtime runtime)
         }
 #endif
 
-#if NETSTANDARD || NETCOREAPP // Possibly .NET Core
+#if !NET461
```
bruno-garcia, line 78: Also here

src/Sentry/DefaultSentryScopeStateProcessor.cs
```diff
@@ -38,7 +38,7 @@ public void Apply(Scope scope, object state)
 
                     break;
                 }
-#if HAS_VALUE_TUPLE
+#if !NET461
```
Tyrrrz, line 41: Note: we can bring `ValueTuple` to .NET461 with: https://www.nuget.org/packages/System.ValueTuple/
This will remove the need for this if/defs

Tyrrrz (approved):

src/Sentry/DefaultSentryScopeStateProcessor.cs
bruno-garcia: I had that originally and it cause a WORLD of pain (binding redirects), so I removed it

PR #1011, opened 2021-05-24
Further simplify transaction integration from legacy ASP.NET

Tyrrrz:


bruno-garcia (approved):
Great! thanks a lot!

bruno-garcia:
Sounds like we are ready to add it to the sample and finally the wizard 🥳

PR #1013, opened 2021-05-26
Client health

Tyrrrz:
Notes:
- [x] ~Increment session error count on all events (with or without exception, so includes `CaptureMessage(...)` too)~
- [x] Mark sessions as errored if a handled exception event went through. 
- [x] `StartSession()` should close previous session if it hasn't been closed
- [x] ~`IHub` can be separated from `ISentryClient` to make it easier to pass `Session` to `CaptureEvent(...)`~
- [x] On unhandled exceptions set state -> crashed.
- [ ] On restart, if an open session exists, and it wasn't a crash in the previous run: abnormal state

Dev docs: https://develop.sentry.dev/sdk/sessions/
User docs: https://docs.sentry.io/product/releases/health/

Cocoa reference: https://github.com/getsentry/sentry-cocoa/blob/master/Sources/Sentry/SentrySession.m

### DIFF
--- a/samples/Sentry.Samples.Console.Basic/Program.cs
+++ b/samples/Sentry.Samples.Console.Basic/Program.cs
@@ -1,16 +1,7 @@
-using System;
-using System.Collections.Generic;
-using System.Threading;
 using Sentry;
 
 using (SentrySdk.Init("https://eb18e953812b41c3aeb042e666fd3b5c@o447951.ingest.sentry.io/5428537"))
 {
-    SentrySdk.StartSession();
-
-    SentrySdk.CaptureMessage("test session");
-
-    SentrySdk.EndSession(SessionEndStatus.Crashed);
-
     // The following exception is captured and sent to Sentry
-    //throw null;
+    throw null;
 }
--- a/src/Sentry/Envelopes/Envelope.cs
+++ b/src/Sentry/Envelopes/Envelope.cs
@@ -82,7 +82,7 @@ namespace Sentry.Protocol.Envelopes
         public static Envelope FromEvent(
             SentryEvent @event,
             IReadOnlyCollection<Attachment>? attachments = null,
-            SessionUpdate? sessionSnapshot = null)
+            SessionUpdate? sessionUpdate = null)
         {
             var header = new Dictionary<string, object?>(StringComparer.Ordinal)
             {
@@ -99,9 +99,9 @@ namespace Sentry.Protocol.Envelopes
                 items.AddRange(attachments.Select(EnvelopeItem.FromAttachment));
             }
 
-            if (sessionSnapshot is not null)
+            if (sessionUpdate is not null)
             {
-                items.Add(EnvelopeItem.FromSessionSnapshot(sessionSnapshot));
+                items.Add(EnvelopeItem.FromSession(sessionUpdate));
             }
 
             return new Envelope(header, items);
@@ -146,11 +146,11 @@ namespace Sentry.Protocol.Envelopes
         /// <summary>
         /// Creates an envelope that contains a session snapshot.
         /// </summary>
-        public static Envelope FromSessionSnapshot(SessionUpdate sessionUpdate)
+        public static Envelope FromSession(SessionUpdate sessionUpdate)
         {
             var items = new[]
             {
-                EnvelopeItem.FromSessionSnapshot(sessionUpdate)
+                EnvelopeItem.FromSession(sessionUpdate)
             };
 
             return new Envelope(items);
--- a/src/Sentry/Envelopes/Envelope.cs
+++ b/src/Sentry/Envelopes/Envelope.cs
@@ -144,7 +144,7 @@ namespace Sentry.Protocol.Envelopes
         }
 
         /// <summary>
-        /// Creates an envelope that contains a session snapshot.
+        /// Creates an envelope that contains a session update.
         /// </summary>
         public static Envelope FromSession(SessionUpdate sessionUpdate)
         {
--- a/src/Sentry/Envelopes/Envelope.cs
+++ b/src/Sentry/Envelopes/Envelope.cs
@@ -84,7 +84,7 @@ namespace Sentry.Protocol.Envelopes
             IReadOnlyCollection<Attachment>? attachments = null,
             SessionUpdate? sessionUpdate = null)
         {
-            var header = new Dictionary<string, object?>(StringComparer.Ordinal)
+            var header = new Dictionary<string, object?>(1, StringComparer.Ordinal)
             {
                 [EventIdKey] = @event.EventId.ToString()
             };
@@ -112,7 +112,7 @@ namespace Sentry.Protocol.Envelopes
         /// </summary>
         public static Envelope FromUserFeedback(UserFeedback sentryUserFeedback)
         {
-            var header = new Dictionary<string, object?>(StringComparer.Ordinal)
+            var header = new Dictionary<string, object?>(1, StringComparer.Ordinal)
             {
                 [EventIdKey] = sentryUserFeedback.EventId.ToString()
             };
@@ -130,7 +130,7 @@ namespace Sentry.Protocol.Envelopes
         /// </summary>
         public static Envelope FromTransaction(Transaction transaction)
         {
-            var header = new Dictionary<string, object?>(StringComparer.Ordinal)
+            var header = new Dictionary<string, object?>(1, StringComparer.Ordinal)
             {
                 [EventIdKey] = transaction.EventId.ToString()
             };
--- a/src/Sentry/Extensibility/DisabledHub.cs
+++ b/src/Sentry/Extensibility/DisabledHub.cs
@@ -102,7 +102,7 @@ namespace Sentry.Extensibility
         /// <summary>
         /// No-Op.
         /// </summary>
-        public SentryId CaptureEvent(SentryEvent evt, Scope? scope = null, Session? session = null) => SentryId.Empty;
+        public SentryId CaptureEvent(SentryEvent evt, Scope? scope = null) => SentryId.Empty;
 
         /// <summary>
         /// No-Op.
--- a/src/Sentry/GlobalSessionManager.cs
+++ b/src/Sentry/GlobalSessionManager.cs
@@ -210,7 +210,7 @@ namespace Sentry
             return null;
         }
 
-        private SessionUpdate? EndSession(Session session, SessionEndStatus status)
+        private SessionUpdate EndSession(Session session, SessionEndStatus status)
         {
             session.End(status);
 
--- a/src/Sentry/GlobalSessionManager.cs
+++ b/src/Sentry/GlobalSessionManager.cs
@@ -116,10 +116,19 @@ namespace Sentry
                 TryGetPersistentInstallationId() ??
                 TryGetHardwareInstallationId();
 
-            _options.DiagnosticLogger?.LogDebug(
-                "Resolved installation ID '{0}'.",
-                id
-            );
+            if (!string.IsNullOrWhiteSpace(id))
+            {
+                _options.DiagnosticLogger?.LogDebug(
+                    "Resolved installation ID '{0}'.",
+                    id
+                );
+            }
+            else
+            {
+                _options.DiagnosticLogger?.LogDebug(
+                    "Failed to resolve installation ID."
+                );
+            }
 
             return _cachedInstallationId = id;
         }
--- a/src/Sentry/GlobalSessionManager.cs
+++ b/src/Sentry/GlobalSessionManager.cs
@@ -66,7 +66,9 @@ namespace Sentry
 
                 return id;
             }
-            catch (IOException ex)
+            // If there's no write permission or the platform doesn't support this, we handle
+            // and let the next installation id strategy kick in
+            catch (Exception ex)
             {
                 _options.DiagnosticLogger?.LogError(
                     "Failed to resolve persistent installation ID.",
--- a/src/Sentry/GlobalSessionManager.cs
+++ b/src/Sentry/GlobalSessionManager.cs
@@ -1,5 +1,5 @@
-﻿using System.Linq;
-using System.Net.NetworkInformation;
+﻿using System;
+using System.IO;
 using Sentry.Extensibility;
 using Sentry.Internal;
 
@@ -15,6 +15,30 @@ namespace Sentry
         public GlobalSessionManager(SentryOptions options)
         {
             _options = options;
+        }
+
+        private string GetInstallationId()
+        {
+            var filePath = Path.Combine(
+                Environment.GetFolderPath(Environment.SpecialFolder.LocalApplicationData),
+                "Sentry",
+                ".installation"
+            );
+
+            // Try to read existing
+            try
+            {
+                return File.ReadAllText(filePath);
+            }
+            catch (FileNotFoundException)
+            {
+            }
+
+            // Generate new
+            var id = Guid.NewGuid().ToString();
+            File.WriteAllText(filePath, id);
+
+            return id;
         }
 
         public Session? StartSession()
@@ -41,15 +65,7 @@ namespace Sentry
             }
 
             var environment = EnvironmentLocator.Resolve(_options);
-
-            // TODO: proper distinct id
-            var distinctId = NetworkInterface
-                .GetAllNetworkInterfaces()
-                .Where(nic =>
-                    nic.OperationalStatus == OperationalStatus.Up &&
-                    nic.NetworkInterfaceType != NetworkInterfaceType.Loopback)
-                .Select(nic => nic.GetPhysicalAddress().ToString())
-                .FirstOrDefault();
+            var distinctId = GetInstallationId();
 
             var session = new Session(distinctId, release, environment);
             CurrentSession = session;
--- a/src/Sentry/GlobalSessionManager.cs
+++ b/src/Sentry/GlobalSessionManager.cs
@@ -22,11 +22,19 @@ namespace Sentry
 
         private string GetInstallationId()
         {
-            // Prevent race conditions that could cause us to generate multiple IDs.
-            // Note: this has to be synchronized across multiple processes too.
+            // Installation ID could have already been resolved by this point.
+            // Try to retrieve it from cache to avoid expensive I/O and locking.
+            if (!string.IsNullOrWhiteSpace(_cachedInstallationId))
+            {
+                return _cachedInstallationId;
+            }
+
+            // Resolve installation ID in a locked manner to avoid race conditions.
+            // Note: in the future, this probably has to be synchronized across multiple processes too.
             lock (_lock)
             {
-                // Avoid IO if possible
+                // We may have acquired the lock after another thread has already resolved
+                // installation ID, so check the cache one more time.
                 if (!string.IsNullOrWhiteSpace(_cachedInstallationId))
                 {
                     return _cachedInstallationId;
@@ -44,7 +52,7 @@ namespace Sentry
                     Directory.CreateDirectory(directoryPath);
                 }
 
-                // Try to read existing
+                // Read installation ID stored in a file
                 try
                 {
                     return _cachedInstallationId = File.ReadAllText(filePath);
@@ -53,7 +61,7 @@ namespace Sentry
                 {
                 }
 
-                // Generate new
+                // Generate new installation ID and store it in a file
                 var id = Guid.NewGuid().ToString();
                 File.WriteAllText(filePath, id);
 
--- a/src/Sentry/GlobalSessionManager.cs
+++ b/src/Sentry/GlobalSessionManager.cs
@@ -25,7 +25,7 @@ namespace Sentry
                     "Starting a new session while an existing one is still active."
                 );
 
-                // End previous session (TODO: should this be abnormal instead?)
+                // End previous session
                 EndSession(SessionEndStatus.Exited);
             }
 
--- a/src/Sentry/GlobalSessionManager.cs
+++ b/src/Sentry/GlobalSessionManager.cs
@@ -17,6 +17,9 @@ namespace Sentry
         private string? _cachedInstallationId;
         private Session? _currentSession;
 
+        // Internal for testing
+        internal Session? CurrentSession => _currentSession;
+
         public GlobalSessionManager(SentryOptions options)
         {
             _options = options;
--- a/src/Sentry/GlobalSessionManager.cs
+++ b/src/Sentry/GlobalSessionManager.cs
@@ -26,10 +26,14 @@ namespace Sentry
         {
             try
             {
-                // Store in cache directory or fall back to appdata
-                var directoryPath = !string.IsNullOrWhiteSpace(_options.CacheDirectoryPath)
-                    ? _options.CacheDirectoryPath
-                    : Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.LocalApplicationData), "Sentry");
+                var directoryPath = Path.Combine(
+                    // Store in cache directory or fall back to appdata
+                    !string.IsNullOrWhiteSpace(_options.CacheDirectoryPath)
+                        ? _options.CacheDirectoryPath
+                        : Environment.GetFolderPath(Environment.SpecialFolder.LocalApplicationData),
+                    // Put under "Sentry" subdirectory
+                    "Sentry"
+                );
 
                 Directory.CreateDirectory(directoryPath);
 
@@ -147,7 +151,7 @@ namespace Sentry
             }
         }
 
-        public Session? StartSession()
+        public SessionUpdate? StartSession()
         {
             // Extract release
             var release = ReleaseLocator.Resolve(_options);
@@ -185,24 +189,25 @@ namespace Sentry
                 session.Id, session.DistinctId
             );
 
-            return session;
+            return session.CreateUpdate(true);
         }
 
-        public void ReportError()
+        public SessionUpdate? ReportError()
         {
             if (_currentSession is { } session)
             {
                 session.ReportError();
+                return session.CreateUpdate(false);
             }
-            else
-            {
-                _options.DiagnosticLogger?.LogError(
-                    "Failed to report an error on a session because there is none active."
-                );
-            }
+
+            _options.DiagnosticLogger?.LogDebug(
+                "Failed to report an error on a session because there is none active."
+            );
+
+            return null;
         }
 
-        private Session EndSession(Session session, SessionEndStatus status)
+        private SessionUpdate? EndSession(Session session, SessionEndStatus status)
         {
             session.End(status);
 
@@ -211,10 +216,10 @@ namespace Sentry
                 session.Id, session.DistinctId, status
             );
 
-            return session;
+            return session.CreateUpdate(false);
         }
 
-        public Session? EndSession(SessionEndStatus status)
+        public SessionUpdate? EndSession(SessionEndStatus status)
         {
             var session = Interlocked.Exchange(ref _currentSession, null);
             if (session is null)
--- a/src/Sentry/GlobalSessionManager.cs
+++ b/src/Sentry/GlobalSessionManager.cs
@@ -24,13 +24,6 @@ namespace Sentry
 
         private string? TryGetPersistentInstallationId()
         {
-            // We may have acquired the lock after another thread has already resolved
-            // installation ID, so check the cache one more time before proceeding with I/O.
-            if (!string.IsNullOrWhiteSpace(_cachedInstallationId))
-            {
-                return _cachedInstallationId;
-            }
-
             try
             {
                 // Store in cache directory or fall back to appdata
@@ -125,6 +118,13 @@ namespace Sentry
             // Note: in the future, this probably has to be synchronized across multiple processes too.
             lock (_lock)
             {
+                // We may have acquired the lock after another thread has already resolved
+                // installation ID, so check the cache one more time before proceeding with I/O.
+                if (!string.IsNullOrWhiteSpace(_cachedInstallationId))
+                {
+                    return _cachedInstallationId;
+                }
+
                 var id =
                     TryGetPersistentInstallationId() ??
                     TryGetHardwareInstallationId();
--- a/src/Sentry/GlobalSessionManager.cs
+++ b/src/Sentry/GlobalSessionManager.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.IO;
+using System.Threading;
 using Sentry.Extensibility;
 using Sentry.Internal;
 
@@ -10,7 +11,8 @@ namespace Sentry
     {
         private readonly SentryOptions _options;
 
-        public Session? CurrentSession { get; private set; }
+        private Session? _currentSession;
+        public Session? CurrentSession => _currentSession;
 
         public GlobalSessionManager(SentryOptions options)
         {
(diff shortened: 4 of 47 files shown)


src/Sentry/Internal/Hub.cs
```diff
@@ -230,6 +293,18 @@ public void CaptureTransaction(Transaction transaction)
             }
         }
 
+        public void CaptureSessionSnapshot(SessionSnapshot sessionSnapshot)
```
bruno-garcia, line 296: Can we get away without this being public?

src/Sentry/Session.cs
```diff
@@ -0,0 +1,61 @@
...
+                release,
+                null,
+                null,
+                null)
+        {
+        }
+
+        public void End(SessionEndState state) => EndState = state;
+
+        public void ReportError() => ErrorCount++;
```
bruno-garcia, line 57: needs a field and `Interlocked.Increment`

lucas-zimerman:
It would be great if Capture message with event-level Info and Debug wouldn't count as errors on Sessions.

src/Sentry/Internal/EnvironmentLocator.cs
```diff
@@ -1,17 +1,35 @@
...
-        public static string? Current => Environment.Value;
+        public static string? Current => FromEnvironmentVariable.Value;
 
-        internal static string? Locate() => System.Environment.GetEnvironmentVariable(Constants.EnvironmentEnvironmentVariable);
+        internal static string? LocateFromEnvironmentVariable() =>
+            Environment.GetEnvironmentVariable(Constants.EnvironmentEnvironmentVariable);
+
+        public static string Resolve(SentryOptions options)
+        {
+            // Changing from `LocateFromEnvironmentVariable()` to `Current` fails tests?
```
Tyrrrz, line 21: It sounds like the state of `Lazy<T>` persists across multiple tests because it's static.

src/Sentry/Internal/ReleaseLocator.cs
```diff
@@ -4,12 +4,13 @@ namespace Sentry.Internal
...
-            => Environment.GetEnvironmentVariable(Constants.ReleaseEnvironmentVariable)
-                ?? ApplicationVersionLocator.GetCurrent();
+        private static readonly Lazy<string?> CurrentLazy = new(ResolveFromEnvironment);
+
+        private static string? ResolveFromEnvironment() =>
+            Environment.GetEnvironmentVariable(Constants.ReleaseEnvironmentVariable)
+            ?? ApplicationVersionLocator.GetCurrent();
+
+        // Replacing `ResolveFromEnvironment()` with `CurrentLazy.Value` fails tests?
+        public static string? Resolve(SentryOptions options) => options.Release ?? ResolveFromEnvironment();
```
Tyrrrz, line 14: It sounds like the state of `Lazy<T>` persists across multiple tests because it's static.

bruno-garcia:
> It would be great if Capture message with event-level Info and Debug wouldn't count as errors on Sessions.

We defined in the protocol that any event captured counts as errored

lucas-zimerman:
> > It would be great if Capture message with event-level Info and Debug wouldn't count as errors on Sessions.
> 
> We defined in the protocol that any event captured counts as errored

Oh I see, sadly there'll be use cases where all sessions were errored by folks using Sentry as an Analytics tool (me for example)

src/Sentry/Internal/Hub.cs
Tyrrrz: Not to my knowledge

src/Sentry/Extensibility/DisabledHub.cs
```diff
@@ -88,7 +102,7 @@ public void BindClient(ISentryClient client)
         /// <summary>
         /// No-Op.
         /// </summary>
-        public SentryId CaptureEvent(SentryEvent evt, Scope? scope = null) => SentryId.Empty;
+        public SentryId CaptureEvent(SentryEvent evt, Scope? scope = null, Session? session = null) => SentryId.Empty;
```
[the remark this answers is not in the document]
Tyrrrz: I still don't understand how this would work for global session (client mode)

src/Sentry/Envelopes/Envelope.cs
```diff
@@ -74,7 +79,10 @@ public async Task SerializeAsync(Stream stream, CancellationToken cancellationTo
         /// <summary>
         /// Creates an envelope that contains a single event.
         /// </summary>
-        public static Envelope FromEvent(SentryEvent @event, IReadOnlyCollection<Attachment>? attachments = null)
+        public static Envelope FromEvent(
+            SentryEvent @event,
+            IReadOnlyCollection<Attachment>? attachments = null,
+            SessionSnapshot? sessionSnapshot = null)
```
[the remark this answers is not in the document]
Tyrrrz: `ISession` would then only be implemented by `SesssionSnapshot` anyway though. It will have to contain properties that are unique to the snapshot.

src/Sentry/Extensibility/DisabledHub.cs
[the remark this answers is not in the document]
bruno-garcia: Instead of passing the session here it goes into the scope. (the second argument)

src/Sentry/GlobalSessionManager.cs
```diff
@@ -0,0 +1,89 @@
...
+
+        public Session? StartSession()
+        {
+            if (CurrentSession is not null)
+            {
+                _options.DiagnosticLogger?.LogWarning(
+                    "Starting a new session while an existing one is still active."
+                );
+
+                // End previous session (TODO: should this be abnormal instead?)
```
bruno-garcia, line 28: I believe exited is OK:
```suggestion
```

src/Sentry/GlobalSessionManager.cs
```diff
@@ -0,0 +1,89 @@
...
+                _options.DiagnosticLogger?.LogError(
+                    "Failed to start a session because there is no release information."
+                );
+
+                return null;
+            }
+
+            var environment = EnvironmentLocator.Resolve(_options);
+
+            // TODO: proper distinct id
```
bruno-garcia, line 45: This is how cocoa/java does, the 'installation id': https://github.com/getsentry/sentry-cocoa/blob/master/Sources/Sentry/SentryInstallation.m

just a guid we write once

src/Sentry/GlobalSessionManager.cs
```diff
@@ -0,0 +1,89 @@
...
+            }
+
+            session.End(status);
+
+            _options.DiagnosticLogger?.LogInfo(
+                "Ended session (SID: {0}; DID: {1}) with state '{2}'.",
+                session.Id, session.DistinctId, status
+            );
+
+            CurrentSession = null;
```
bruno-garcia, line 84: this code would need to be synchronized. Either by locking the whole thing, or preferably just using Interlocked to swap references

src/Sentry/ISessionManager.cs
```diff
@@ -0,0 +1,11 @@
+﻿namespace Sentry
+{
+    internal interface ISessionManager
+    {
+        Session? CurrentSession { get; }
```
bruno-garcia, line 5: Exposing the reference will make it harder for us to synchronize things. Ideally mutations would come through method calls into this interface instead

src/Sentry/Internal/Hub.cs
```diff
@@ -230,6 +258,18 @@ public void CaptureTransaction(Transaction transaction)
             }
         }
 
+        public void CaptureSession(SessionUpdate sessionUpdate)
+        {
+            try
+            {
+                _ownedClient.CaptureSession(sessionUpdate);
```
bruno-garcia, line 265: Do we have `CaptureSession` in the spec?
I was hoping this would not be exposed. The client can take the Session from the scope (that can be internal) and add to an envelope if one was provided. I believe we do that for attachments too, no?

Tyrrrz: Attachments are not sent separately from events, but sessions are. So we need this as part of the interface.

src/Sentry/Internal/Hub.cs
```diff
@@ -188,9 +214,18 @@ public SentryId CaptureEvent(SentryEvent evt, Scope? scope = null)
...
+                _sessionManager.ReportError();
+
                 var id = currentScope.Value.CaptureEvent(evt, actualScope);
                 actualScope.LastEventId = id;
 
+                // If the event contains unhandled exception - end session as crashed
+                if (evt.SentryExceptions?.Any(e => !(e.Mechanism?.Handled ?? true)) ?? false)
+                {
+                    EndSession(SessionEndStatus.Crashed);
+                }
```
Tyrrrz, line 227: Is this good enough?

src/Sentry/Internal/Hub.cs
bruno-garcia: Another language limitation/design issue?

We don't need users to capture sessions, so it ideally stays an internal business. The API is: `StartSessions/EndSessions`.

Can we have an internal extension method to capture that which will wrap in an envelope?

Or as I originally suggested, `Scope.Session`, we just set the Session instance to the scope before passing down to the client (I believe I did that on iOS)

src/Sentry/GlobalSessionManager.cs
```diff
@@ -0,0 +1,151 @@
...
+        {
+            // Prevent race conditions that could cause us to generate multiple IDs.
+            // Note: this has to be synchronized across multiple processes too.
+            lock (_lock)
+            {
+                // Avoid IO if possible
+                if (!string.IsNullOrWhiteSpace(_cachedInstallationId))
+                {
+                    return _cachedInstallationId;
+                }
```
bruno-garcia, line 33: avoid taking the lock too: double check locking

```suggestion
            if (!string.IsNullOrWhiteSpace(_cachedInstallationId))
            {
                return _cachedInstallationId;
            }

            // Prevent race conditions that could cause us to generate multiple IDs.
            // Note: this has to be synchronized across multiple processes too.
            lock (_lock)
            {
                // Avoid IO if possible
                if (!string.IsNullOrWhiteSpace(_cachedInstallationId))
                {
                    return _cachedInstallationId;
                }
```

src/Sentry/GlobalSessionManager.cs
```diff
@@ -0,0 +1,151 @@
...
+
+                var filePath = Path.Combine(
+                    Environment.GetFolderPath(Environment.SpecialFolder.LocalApplicationData),
```
bruno-garcia, line 36: We can't really assume there'll be access to disk I'm afraid. We could have a logic such as:

Rely on `CacheDirectoryPath`. If it's null, lets try to get the app data directory (what you did here, though I'd be afraid we might need an opt out, and definitely deal with errors that might throw from attempting to do I/O here).

If that fails we could hash `Environment.MachineName` or something else else cheap.
Lastly we could finally fallback to something you doing before like the mac address from the first NIC.

Think this could run on Wasm, Mobile, Unity (IL2CPP) on xbox, ps4 or whatever else they do. We are not testing any of those platforms right now but we can take into account that the code must be as portable as possible.

src/Sentry/GlobalSessionManager.cs
```diff
@@ -0,0 +1,151 @@
...
+                }
+
+                // Try to read existing
+                try
+                {
+                    return _cachedInstallationId = File.ReadAllText(filePath);
+                }
+                catch (FileNotFoundException)
+                {
+                }
```
bruno-garcia, line 54: Would also be nice to have logging here to know we failed to read the file. Also that we're creating the directory.

And generally some error handling on the other I/O. CreateDirectory above would be the first to throw.
Today I spent one hour looking at a customer crash. That's the time yo appreciate good logs.

src/Sentry/GlobalSessionManager.cs
```diff
@@ -0,0 +1,151 @@
...
+            var distinctId = GetInstallationId();
+
+            // Create new session
+            var session = new Session(distinctId, release, environment);
+
+            // Set new session and check whether we ended up overwriting an active one in the process
+            var previousSession = Interlocked.Exchange(ref _currentSession, session);
+            if (previousSession is not null)
+            {
+                _options.DiagnosticLogger?.LogWarning(
```
bruno-garcia, line 94: sweeet!

src/Sentry/GlobalSessionManager.cs
```diff
@@ -0,0 +1,151 @@
...
+                "Ended session (SID: {0}; DID: {1}) with state '{2}'.",
+                session.Id, session.DistinctId, status
+            );
+
+            return session;
+        }
+
+        public Session? EndSession(SessionEndStatus status)
+        {
+            var session = Interlocked.Exchange(ref _currentSession, null);
```
bruno-garcia, line 138: nice!

src/Sentry/Internal/Hub.cs
bruno-garcia: I think this will do!
Lets worry about that native crash story later on. As we chat on the call we might turn off the .NET session support and use the native for Android and iOS anyway for MAUI, Unity at least

bruno-garcia (commented):
Another pass: It's getting in great shape!!

src/Sentry/GlobalSessionManager.cs
Tyrrrz: Makes sense. So how about:

1. Use cache directory
2. Use app data
3. Use MAC address

I don't think there's much difference between machine name and MAC address in terms of cheapness, we do it once per runtime anyway.

bruno-garcia: Yeah I agree that's a good approach

src/Sentry/Internal/Hub.cs
Tyrrrz: It's not a language limitation, just the fact that `ISentryClient` is public, which it is (I assume) to let users provide their own instance in tests. Users don't directly interact with `ISentryClient`, but they need it to be public. And because it's public, these methods need to be public to.

If we change `ISentryClient` to deal with envelopes directly, I fail to see how its purpose going to be different than what's already managed by `ITransport` (which works with envelopes directly).

> Or as I originally suggested, Scope.Session, we just set the Session instance to the scope before passing down to the client (I believe I did that on iOS)

Yeah, we do it.

Tyrrrz:
@bruno-garcia can you please go over and see if I missed anything? What's the definition of done for this task?

src/Sentry/SessionUpdate.cs
```diff
@@ -0,0 +1,137 @@
...
+        public DateTimeOffset Timestamp { get; }
+
+        /// <summary>
+        /// Duration of time since start of the session.
+        /// </summary>
+        public TimeSpan Duration => Timestamp - Session.Timestamp;
+
+        internal SessionUpdate(Session session, bool isInitial, DateTimeOffset timestamp)
+        {
+            Session = session;
```
bruno-garcia, line 34: Isn't this risky? Now we're holding a reference to a mutable object.
Aren't we better off just copying all fields here instead?

test/Sentry.Tests/GlobalSessionManagerTests.cs
```diff
@@ -0,0 +1,181 @@
...
+using FluentAssertions;
+using Sentry.Testing;
+using Xunit;
+
+namespace Sentry.Tests
+{
+    public class GlobalSessionManagerTests
+    {
+        [Fact]
+        public void StartSession_ReleaseSet_CreatesNewSession()
```
bruno-garcia, line 12: Where is release being set? Isn't clear by reading the code

test/Sentry.Tests/GlobalSessionManagerTests.cs
```diff
@@ -0,0 +1,181 @@
...
+            // Act
+            var session = sessionManager.StartSession();
+
+            // Assert
+            session.Should().NotBe(previousSession);
+            previousSession?.EndStatus.Should().Be(SessionEndStatus.Exited);
+        }
+
+        [Fact]
+        public void StartSession_InstallationId_FileCreated()
```
bruno-garcia, line 49: No tests for installation id without a file path?

test/Sentry.Tests/GlobalSessionManagerTests.cs
```diff
@@ -0,0 +1,181 @@
...
+            sessionManager.ReportError();
+            sessionManager.ReportError();
+
+            // Assert
+            session.Should().NotBeNull();
+            session?.ErrorCount.Should().Be(3);
+        }
+
+        [Fact]
+        public void ReportError_ActiveSessionDoesNotExist_IncrementsErrorCount()
```
bruno-garcia, line 107:
```suggestion
        public void ReportError_ActiveSessionDoesNotExist_LogsOutError()
```

src/Sentry/Session.cs
```diff
@@ -0,0 +1,107 @@
...
+
+        /// <summary>
+        /// Transitions the session to ended state.
+        /// </summary>
+        public void End(SessionEndStatus status) => EndStatus = status;
+
+        /// <summary>
+        /// Creates a snapshot of this session.
+        /// </summary>
+        public SessionUpdate CreateSnapshot(bool isInitial) => new(this, isInitial);
```
bruno-garcia, line 105: Having the flag sit only on the snapshot is a pretty good idea!

src/Sentry/GlobalSessionManager.cs
```diff
@@ -0,0 +1,218 @@
...
+                );
+            }
+            else
+            {
+                _options.DiagnosticLogger?.LogDebug(
+                    "Failed to resolve installation ID."
+                );
+            }
+
+            return _cachedInstallationId = id;
```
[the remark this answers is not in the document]
bruno-garcia: ON a second thought I think this installation id should be a different concern. You should take an abstraction as a dependency and test that you only poke that dependency once (value is cached). The abstraction would look like `IInstallationIdProvider`. We could have 2 implementations (the File and the NetworkCard one) but to keep this simple it would be fine to have a single one (i.e: `DefaultInstallationIdProvider` or something that includes the code you have here). `Lazy<T>` is a good approach on that to make sure you only read that value once (only do the I/O or sys call once).

test/Sentry.Tests/Protocol/Envelopes/EnvelopeTests.cs
```diff
@@ -492,6 +535,36 @@ public async Task Roundtrip_WithUserFeedback_Success()
...
+
+            // Act
+            await envelope.SerializeAsync(stream);
+            stream.Seek(0, SeekOrigin.Begin);
+
+            using var envelopeRoundtrip = await Envelope.DeserializeAsync(stream);
+
+            // Assert
+
+            // Can't compare the entire object graph because output envelope contains evaluated length,
```
bruno-garcia, line 559: Can we remove length and compare?

test/Sentry.Tests/SentrySdkTests.cs
```diff
@@ -447,10 +446,17 @@ public void CaptureMessage_SdkInitialized_IncludesScope()
...
+                .Select(m => m.ToString())
+                .ToArray();
+
+            var sentrySdkMembers = typeof(SentrySdk)
+                .GetMembers(BindingFlags.Public | BindingFlags.Static)
+                .Select(m => m.ToString())
+                .ToArray();
 
-            Assert.Empty(clientMembers.Select(m => m.ToString()).Except(sentrySdk.Select(m => m.ToString())));
+            sentrySdkMembers.Should().Contain(clientMembers);
```
bruno-garcia, line 459: Why was this refactor needed? Did it break?

src/Sentry/GlobalSessionManager.cs
```diff
@@ -0,0 +1,218 @@
...
+
+            _options.DiagnosticLogger?.LogInfo(
+                "Started new session (SID: {0}; DID: {1}).",
+                session.Id, session.DistinctId
+            );
+
+            return session;
+        }
+
+        public void ReportError()
```
bruno-garcia, line 177: I just learned that errors are no longer a counter. It's either 0 or 1.
It'll work as it is, but we can simplify things but removing it.

samples/Sentry.Samples.Console.Basic/Program.cs
```diff
@@ -1,13 +1,7 @@
...
-    static void Main()
-    {
-        using (SentrySdk.Init("https://eb18e953812b41c3aeb042e666fd3b5c@o447951.ingest.sentry.io/5428537"))
-        {
-            // The following exception is captured and sent to Sentry
-            throw null;
-        }
-    }
+    // The following exception is captured and sent to Sentry
+    throw null;
```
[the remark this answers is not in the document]
Tyrrrz: Hmm... What do you mean? I just reverted the sample to its pre-PR state.

src/Sentry/SessionUpdate.cs
Tyrrrz: Yeah, good point

test/Sentry.Tests/SentrySdkTests.cs
Tyrrrz: This test failed during development because I forgot to add a method to `SentrySdk`. The error message was indecipherable, however, so I improved the test so that it shows exactly which methods are missing.

src/Sentry/GlobalSessionManager.cs
Tyrrrz: Really? Even on client mode? Was this a recent change, because it's still a number in docs.

src/Sentry/Session.cs
Tyrrrz: Thank you thank you 😊

test/Sentry.Tests/GlobalSessionManagerTests.cs
Tyrrrz: It pulls it from assembly version. I'll make it explicit.

test/Sentry.Tests/Protocol/Envelopes/EnvelopeTests.cs
Tyrrrz: What do you mean? How?

src/Sentry/GlobalSessionManager.cs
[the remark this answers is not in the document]
Tyrrrz: Went with your first suggestion to keep it simpler for now

src/Sentry/Internal/Hub.cs
```diff
@@ -156,6 +163,45 @@ public void BindException(Exception exception, ISpan span)
...
+                    ex
+                );
+            }
+        }
+
+        public void EndSession(SessionEndStatus status = SessionEndStatus.Exited)
+        {
+            try
+            {
+                var session = _sessionManager.EndSession(status);
```
bruno-garcia, line 190: We need to sync between this and startSession and `ConfigureScope(s => s.Session`

bruno-garcia (commented):
Another pass

src/Sentry/Internal/ReleaseLocator.cs
```diff
@@ -4,12 +4,13 @@ namespace Sentry.Internal
...
-        public static string? GetCurrent()
-            => Environment.GetEnvironmentVariable(Constants.ReleaseEnvironmentVariable)
-                ?? ApplicationVersionLocator.GetCurrent();
+        private static readonly Lazy<string?> CurrentLazy = new(ResolveFromEnvironment);
+
+        private static string? ResolveFromEnvironment() =>
+            Environment.GetEnvironmentVariable(Constants.ReleaseEnvironmentVariable)
+            ?? ApplicationVersionLocator.GetCurrent();
+
+        // Replacing `ResolveFromEnvironment()` with `CurrentLazy.Value` fails tests?
```
bruno-garcia, line 13:
```suggestion
```

src/Sentry/Scope.cs
```diff
@@ -328,6 +330,7 @@ public void Apply(Scope other)
             Apply((IEventLike)other);
 
             other.Transaction ??= Transaction;
+            other.Session ??= Session;
```
bruno-garcia, line 333:
```suggestion
            other.SessionUpdate ??= SessionUpdate;
```

src/Sentry/Internal/Hub.cs
```diff
@@ -189,9 +235,21 @@ public SentryId CaptureEvent(SentryEvent evt, Scope? scope = null)
                     evt.Contexts.Trace.ParentSpanId = linkedSpan.ParentSpanId;
                 }
 
+                // Report an error on current session if contains an exception
+                if (evt.Exception is not null)
+                {
+                    _sessionManager.ReportError();
```
bruno-garcia, line 241:
```suggestion
                    // only returns an instance when it goes from 0 to 1
                    SessionUpdate? sessionUpdate = _sessionManager.MarkAsErrored();
                    actualScope.SessionUpdate = sessionUpdate;
```

src/Sentry/Internal/Hub.cs
```diff
@@ -156,6 +163,45 @@ public void BindException(Exception exception, ISpan span)
...
+        }
+
+        public void EndSession(SessionEndStatus status = SessionEndStatus.Exited)
+        {
+            try
+            {
+                var session = _sessionManager.EndSession(status);
+                if (session is not null)
+                {
+                    CaptureSession(session.CreateUpdate(false));
```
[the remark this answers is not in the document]
bruno-garcia: On second thought, the `SessionManager` should be the one holding the reference to the session. And operations on it are synchronized and will return a immutable snapshot: `SessionUpdate`.
For that reason we don't need `                    ConfigureScope(scope => scope.Session = session);` anymore. And `Scope.Session should become `Scope.SessionUpdate`

src/Sentry/Session.cs
```diff
@@ -0,0 +1,89 @@
...
+        /// <inheritdoc />
+        public string? UserAgent { get; }
+
+        /// <inheritdoc />
+        public SessionEndStatus? EndStatus { get; private set; }
+
+        private int _errorCount;
+
+        /// <inheritdoc />
+        public int ErrorCount => _errorCount;
```
bruno-garcia, line 39:
```suggestion
        // Sentry's implementaion changed to only care whether it's 0 or 1 (has error(s))
        public int ErrorCount => _errorCount;
```

src/Sentry/Session.cs
```diff
@@ -0,0 +1,89 @@
...
+                environment,
+                null,
+                null)
+        {
+        }
+
+        /// <summary>
+        /// Reports an error on the session.
+        /// </summary>
+        public void ReportError() => Interlocked.Increment(ref _errorCount);
```
bruno-garcia, line 77: This class doesn't need to be thread safe. The manager will make sure it accesses it safely

src/Sentry/SessionEndStatus.cs
```diff
@@ -0,0 +1,23 @@
...
+    /// </summary>
+    public enum SessionEndStatus
+    {
+        /// <summary>
+        /// Session ended normally.
+        /// </summary>
+        Exited,
+
+        /// <summary>
+        /// Session ended with an error.
```
bruno-garcia, line 14:
```suggestion
        /// Session ended with an unhandled exception.
```

bruno-garcia (commented):
Another bunch

test/Sentry.Tests/Protocol/Envelopes/EnvelopeTests.cs
```diff
@@ -457,6 +457,49 @@ public async Task Roundtrip_WithEvent_WithAttachment_Success()
...
+
+            // Assert
+            envelopeRoundtrip.Items.Should().HaveCount(3);
+
+            envelopeRoundtrip.Items[0].Payload.Should().BeOfType<JsonSerializable>()
+                .Which.Source.Should().BeEquivalentTo(@event);
+
+            envelopeRoundtrip.Items[1].Payload.Should().BeOfType<StreamSerializable>();
+
+            envelopeRoundtrip.Items[2].Payload.Should().BeOfType<JsonSerializable>();
```
bruno-garcia, line 500: Please check that what we serialized was in fact deserialized

test/Sentry.Tests/Protocol/Envelopes/EnvelopeTests.cs
bruno-garcia: What if we do:

Read all bytes from `envelope.GetStream()` it should set `Length`.

The goal is to make sure we are in fact deserializign the thing correctly. This test doesn't verify that.

Tyrrrz:
@bruno-garcia Okay I think I've resolved everything. Please skim over/double check.

src/Sentry/ISession.cs
```diff
@@ -0,0 +1,55 @@
...
+{
+    /// <summary>
+    /// Session metadata.
+    /// </summary>
+    public interface ISession
+    {
+        /// <summary>
+        /// Session auto-generated ID.
+        /// </summary>
+        string Id { get; }
```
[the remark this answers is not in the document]
Tyrrrz: I want to replace this with `SentryId` for consistency. Waiting on reply on Sentry discord.

src/Sentry/GlobalSessionManager.cs
```diff
@@ -0,0 +1,240 @@
...
+            );
+
+            return session.CreateUpdate(true);
+        }
+
+        public SessionUpdate? ReportError()
+        {
+            if (_currentSession is { } session)
+            {
+                session.ReportError();
```
bruno-garcia, line 202: Technically this could be an 'ended' session since another thread could have raced and closed/also called reported error on it. We either need to be able to flip the `errors` atomically or we need to sync all methods here to have a consistent view of things.

Considering worst case scenario is that we double report a session with the same state (errored+closed or errored+2) this isn't critical so we can merge as-is.

src/Sentry/Internal/Hub.cs
```diff
@@ -156,6 +163,44 @@ public void BindException(Exception exception, ISpan span)
...
 
+        public void StartSession()
+        {
+            try
+            {
+                var sessionUpdate = _sessionManager.StartSession();
+                if (sessionUpdate is not null)
+                {
+                    CaptureSession(sessionUpdate);
+                }
```
bruno-garcia, line 174: When can it return null here? Should we log Debug?

src/Sentry/Internal/Hub.cs
```diff
@@ -189,8 +234,27 @@ public SentryId CaptureEvent(SentryEvent evt, Scope? scope = null)
                     evt.Contexts.Trace.ParentSpanId = linkedSpan.ParentSpanId;
                 }
 
+                // Report an error on current session if contains an exception
+                var sessionUpdate = evt.Exception is not null
```
bruno-garcia, line 238: The edge case here is that if this was a capture event that came as a result of a manually crafted event or even deserializing an already proceeded event, this instance (`evt.Exception`) will be null but `SentryException` wouldn't be null so ideally this could be:

```suggestion
                var sessionUpdate = evt.Exception is not null || evt.SentryException?.Any() == true
```

src/Sentry/Internal/Hub.cs
```diff
@@ -189,8 +234,27 @@ public SentryId CaptureEvent(SentryEvent evt, Scope? scope = null)
...
+                    ? _sessionManager.ReportError()
+                    : null;
+
+                // Only set the session if the error count changed from 0 to 1.
```
bruno-garcia, line 242: Sweet

bruno-garcia (approved):
Thanks a lot! Left some nitpicks and some other suggestions but we're good to merge this and give it a try!

Huge effort to get here and it's a great addition to the Sentry SDK for .NET. Thanks!

src/Sentry/Internal/Hub.cs
Tyrrrz: If the `release` is unset for example. It's already logged inside session manager.

Issue #1001, opened 2021-05-17
Serilog + WorkerService + IHostBuilder


Vandersteen:
I've been searching for quite a while how to make Serilog and Sentry work together in a WorkerService environment (.net core > 3)
I've come to the conclusion that this is not easily done / possible right now due to some limitation:

### Known issue:
* There is no `IHostBuilder` extension method for sentry

### Unable to 'hack' together what I need by using internals / copy pasting
* The current `IApplicationBuilder` is marked internal (https://github.com/getsentry/sentry-dotnet/blob/351f264afd569ca46ed180ea87a885c0fc212740/src/Sentry.AspNetCore/ApplicationBuilderExtensions.cs#L25) 
* The current `IServiceCollectionExtension` method is marked internal (https://github.com/getsentry/sentry-dotnet/blob/351f264afd569ca46ed180ea87a885c0fc212740/src/Sentry.AspNetCore/Extensions/DependencyInjection/ServiceCollectionExtensions.cs#L14)
* Some classes used in these methods are also marked internal

### Work around
The current 'work-around' I have is using the Serilog sink and let it initialise the SDK, and add the following in `ConfigureServices`:

**WARNING** This is not production tested, just to make it work locally
```
services.TryAddTransient<IHub>((_) => HubAdapter.Instance);
services.TryAddTransient<ISentryClient>((_) => HubAdapter.Instance);
services.TryAddTransient<ISentryScopeManager>((_) => HubAdapter.Instance);
```

However this way you lose the extensibility of Sentry together with DI

There might be other ways, in which case feel free to add them :)

### Proposal

First an observation, `IWebHostBuilder` on a high level does 2 things (i know it's a bit reductive):
* (1) Integrates Sentry with the AspNetCore host (DI, Logging, HostLifetime, ...)
* (2) Decorates the request pipeline using a [middleware](https://github.com/getsentry/sentry-dotnet/blob/main/src/Sentry.AspNetCore/SentryMiddleware.cs)

Proposed changes:
* Add `IHostBuilder` extension method
    * `UseSentryCore` that handles (1)
* Add `IWebHostBuilder` extension method
    * `UseSentryWeb` that handles (2)
* Keep `IWebHostBuilder` extension method
    * `UseSentry` for backwards compatibility

### Analysis

This is after a short, preliminary & quick swoop through the code base:

Split following classes into core and web (or add methods to support them)
[...]

Issue #190, opened 2019-05-07
Add extensions for IHostBuilder


LordMike:
The current implementation for MS's DI is for `IWebHostBuilder`, but given the new Generic Hosts, I think more and more will use the `IHostBuilder` builder.

Please add a `.UseSentry()` for `IHostBuilder`. :)

bruno-garcia:
True this is pending. Would be great to get some help on this.

xt0rted:
I'm migrating a console app over to the new generic host and in order to use `appsettings.json` for my config with the `ILoggingBuilder.AddSentry()` method I needed to adjust my setup like so:

```diff
 var builder = Host.CreateDefaultBuilder()
     .ConfigureLogging((ctx, loggingBuilder) =>
     {
+        loggingBuilder.Services.Configure<SentryLoggingOptions>(ctx.Configuration.GetSection("Sentry"));
         loggingBuilder.AddSentry();
     })
```

To make this more reusable I've added a helper function:

```csharp
public static class SentryHostBuilderExtensions
{
    public static IHostBuilder UseSentry(this IHostBuilder builder) =>
        builder.ConfigureLogging((ctx, logging) =>
        {
            logging.Services.Configure<SentryLoggingOptions>(ctx.Configuration.GetSection("Sentry"));

            logging.AddSentry();
        });
}
```

And then I use it just like the asp.net core integration:

```csharp
var builder = Host.CreateDefaultBuilder()
    .UseSentry()
    .ConfigureServices((ctx, services) => { ... });
```

bruno-garcia:
This looks straightforward @xt0rted thanks.
Isn't there any other hooks we can add via `IHostBuilder`? With `IWebHostBuilder` we have the middleware pipeline so that's an obvious place we add code to. But I'm unaware of `IHostBuilder`.

Is there a way to capture exceptions of `BackgroundService`s or whatever else is usually setup?

xt0rted:
@bruno-garcia I'm not seeing any graceful exception handling for these (here's the [Host](https://github.com/aspnet/Extensions/blob/master/src/Hosting/Hosting/src/Internal/Host.cs) code). In the `StopAsync()` method it does throw an aggregate exception if anything was thrown inside of the `IHostedService`s, but that  just throws, it doesn't even call `logger.LogError(ex)` 😞 

Here's a question about this and the feature has been moved to the backlog https://github.com/aspnet/Extensions/issues/813.

bruno-garcia:
[...]

Issue #103, opened 2018-09-22
Failure to capture event: System.ObjectDisposedException: Cannot access a disposed object.


alesdvorakcz:
Hi I tried sentry in my asp.net core app. When I turn on Debug mode I can see this exception:

```
Sentry.ISentryClient[0]
      Failure to capture event: System.ObjectDisposedException: Cannot access a disposed object.
      Object name: 'SentryClient'.
         at Sentry.SentryClient.CaptureEvent(SentryEvent event, Scope scope) in C:\projects\sentry-dotnet\src\Sentry\SentryClient.cs:line 77
         at Sentry.Internal.Hub.CaptureEvent(SentryEvent evt, Scope scope) in C:\projects\sentry-dotnet\src\Sentry\Internal\Hub.cs:line 95
dbugdbug: Sentry.ISentryClient[0]
      Configuring the scope.
: Sentry.ISentryClient[0]
      Configuring the scope.
dbugfail: Sentry.ISentryClient[0]
      Configuring the scope.
: Sentry.ISentryClient[0]
      Failure to capture event: System.ObjectDisposedException: Cannot access a disposed object.
      Object name: 'SentryClient'.
         at Sentry.SentryClient.CaptureEvent(SentryEvent event, Scope scope) in C:\projects\sentry-dotnet\src\Sentry\SentryClient.cs:line 77
         at Sentry.Internal.Hub.CaptureEvent(SentryEvent evt, Scope scope) in C:\projects\sentry-dotnet\src\Sentry\Internal\Hub.cs:line 95
```

I dont think I have some special setup. Its dotnet core 2.1 web api with EF core and Autofac DI container.
Unfortunately I can post here whole project but if you need more info don't hesitate to ask.

bruno-garcia:
@alesdvorakcz 
This would happen in case the SDK got disposed.

Could you please clarify how do you initialize the SDK? Is it via `SentrySdk.Init`? If so, how do you handle the `IDisposable` returned?

Or do you use `UseSentry` on the `WebHostBuilder` like the [samples](https://github.com/getsentry/sentry-dotnet/blob/4c35e55da900802d568bb644e5939e0ee9a6d5f6/samples/Sentry.Samples.AspNetCore.Mvc/Program.cs#L21)?

Did you register `ISentryClient` or `IHub` with an Autofac module or something? It could be an issue with the lifetime used in the registration.

It would be great if you could have a small reproducible code to make it easier to troubleshoot.

[...]

PR #1015, opened 2021-05-27
Add IHostBuilder support

Vandersteen:
As discussed here: https://github.com/getsentry/sentry-dotnet/issues/1001

First quick and dirty commit

My first thought after beginning to add an implementation: would it be bad if it acted 100% like the IWebHostBuilder extension.

The user would have the choice between:

```
Host.CreateDefaultBuilder(args)
    .ConfigureWebHostDefaults(webBuilder =>
    {
        webBuilder.UseSentry()
        webBuilder.UseStartup<Startup>();
    });
```

And

```
Host.CreateDefaultBuilder(args)
    .ConfigureWebHostDefaults(webBuilder =>
    {
        webBuilder.UseStartup<Startup>();
    })
    .UseSentry();
```

Only 'caveat' right now is that it would add a few classes that would never be used in a WorkerService environment:
* IStartupFilter, SentryStartupFilter
* IRequestPayloadExtractor, FormRequestPayloadExtractor
* IRequestPayloadExtractor, DefaultRequestPayloadExtractor

However i'm not sure this is such an issue, thoughts ?

Also, I haven't taken https://github.com/getsentry/sentry-dotnet/issues/1001#issuecomment-846674053 into consideration


src/Sentry.AspNetCore/SentryHostBuilderExtensions.cs
```diff
@@ -0,0 +1,107 @@
...
+    [EditorBrowsable(EditorBrowsableState.Never)]
+    public static class SentryHostBuilderExtensions
+    {
+        /// <summary>
+        /// Uses Sentry integration.
+        /// </summary>
+        /// <param name="builder">The builder.</param>
+        /// <returns></returns>
+        public static IHostBuilder UseSentry(this IHostBuilder builder)
+            => UseSentry(builder, (Action<SentryAspNetCoreOptions>?)null);
```
Tyrrrz, line 25:
```suggestion
            => builder.UseSentry((Action<SentryAspNetCoreOptions>?)null);
```

Tyrrrz:
Can we converge on just one way of using `UseSentry()` somehow?

Vandersteen:
> Can we converge on just one way of using `UseSentry()` somehow?

What exactly do you mean ?

Tyrrrz:
> > Can we converge on just one way of using `UseSentry()` somehow?
> 
> What exactly do you mean ?

Sorry. I was referring to how having both the option to put `UseSentry()` inside `ConfigureWebHostDefaults(...)` and outside of it seems a bit confusing from user perspective. I wonder if we can do anything about that.

Vandersteen:
> > > Can we converge on just one way of using `UseSentry()` somehow?
> > 
> > 
> > What exactly do you mean ?
> 
> Sorry. I was referring to how having both the option to put `UseSentry()` inside `ConfigureWebHostDefaults(...)` and outside of it seems a bit confusing from user perspective. I wonder if we can do anything about that.

I'm not sure we can do anything about that.
IWebHostBuilder was the predecessor for IHostBuilder

I understand it could be confusing.

One way I suggested to 'tackle' that problem was explained in the original issue

* UseSentryWeb() on IWebHostBuilder
* UseSentryCore() on IHostBuilder

But that was also not desirable behaviour

Tyrrrz:
Let's see what @bruno-garcia thinks about this

bruno-garcia:
Not a fan of the two methods either.
Can we phase out the one on `IWebHostBuilder` then? Can we do everything through `IHostBuilder` instead?

Vandersteen:
My 2 cents, so you can make an informed decision

> Can we phase out the one on `IWebHostBuilder` then? 

That could indeed be an option, deprecate IWebHostBuilder extensions.
However, do note that `WebHost.CreateDefaultBuilder` will no longer be supported then, forcing users to use `Host.CreateDefaultBuilder`

While both still seem acceptable ways to create a Host from asp.net core's standpoint. There does not seem to be plans to deprecate/remove `WebHost`

```
// no longer supported
WebHost.CreateDefaultBuilder()
    .UseSentry()

// required
Host.CreateDefaultBuilder()
   .ConfigureWebHostDefaults(webBuilder =>
    {
        webBuilder.UseStartup<Startup>();
    })
    .UseSentry()
```

As far as I know, both of these are equivalent.
I don't see any signs that IWebHostBuilder will be deprecated as well (f.e. it is used in the callback of `ConfigureWebHostDefaults`)

>Can we do everything through `IHostBuilder` instead?

I can't see anything that you cannot achieve through the `IHostBuilder` that you can with the `IWebHostBuilder`

But don't take my word for it :)

bruno-garcia:
Sounds like we do need to keep both. So one way is: We add the infrastructure for `IHostBuilder` and update our docs to show only the use of this extension going forward.

We could add a note that `UseSentry` exists on `IWebHostBuilder` but is not needed if the one in `IHostBuilder` was already called.

All of that assumes that we can make them equivalent. If we can add "more" features through `IWebHostBuilder` this wouldn't really work.

Vandersteen:
Exactly, at the time of writing, the current `IHostBuilder` extensions in this PR are a 'literal' copy of the `IWebHostBuilder` with IWebHostBuilder replaced with IHostBuilder and some namespace / include changes.

bruno-garcia:
> Exactly, at the time of writing, the current `IHostBuilder` extensions in this PR are a 'literal' copy of the `IWebHostBuilder` with IWebHostBuilder replaced with IHostBuilder and some namespace / include changes.

Sounds like we could pull the code to a single place and forward the call from both places then?

Is the code reentrant already? We can expect that ppl will call it twice (from each interface).

Vandersteen:
> Sounds like we could pull the code to a single place and forward the call from both places then?

I looked into that, but IHostBuilder and IWebHostBuilder are not the same interface, so code sharing is not easily done

> Is the code reentrant already? We can expect that ppl will call it twice (from each interface).

No not yet, I will comment in code where I see issues

src/Sentry.AspNetCore/SentryHostBuilderExtensions.cs
```diff
@@ -0,0 +1,107 @@
...
+            _ = builder.ConfigureLogging((context, logging) =>
+            {
+                logging.AddConfiguration();
+
+                var section = context.Configuration.GetSection("Sentry");
+                _ = logging.Services.Configure<SentryAspNetCoreOptions>(section);
+
+                _ = logging.Services
+                    .AddSingleton<IConfigureOptions<SentryAspNetCoreOptions>, SentryAspNetCoreOptionsSetup>();
+                _ = logging.Services.AddSingleton<ILoggerProvider, SentryAspNetCoreLoggerProvider>();
```
Vandersteen, line 92: @bruno-garcia using TryAddSingleton here can cause issues, as it won't register it if let's say Serilog or other already exists, in this case you want the 'override' ability

src/Sentry.AspNetCore/SentryHostBuilderExtensions.cs
```diff
@@ -0,0 +1,107 @@
...
+        /// <param name="configureSentry">The Sentry builder.</param>
+        /// <returns></returns>
+        public static IHostBuilder UseSentry(
+            this IHostBuilder builder,
+            Action<HostBuilderContext, ISentryBuilder>? configureSentry)
+        {
+            // The earliest we can hook the SDK initialization code with the framework
+            // Initialization happens at a later time depending if the default MEL backend is enabled or not.
+            // In case the logging backend was replaced, init happens later, at the StartupFilter
+            _ = builder.ConfigureLogging((context, logging) =>
```
bruno-garcia, line 83: The code inside thsi callback could be moved out and shared by `IHostBuilderExtensions` and `IWebHostBuilderExtensions`

src/Sentry.AspNetCore/SentryHostBuilderExtensions.cs
```diff
@@ -0,0 +1,107 @@
...
+                logging.AddConfiguration();
+
+                var section = context.Configuration.GetSection("Sentry");
+                _ = logging.Services.Configure<SentryAspNetCoreOptions>(section);
+
+                _ = logging.Services
+                    .AddSingleton<IConfigureOptions<SentryAspNetCoreOptions>, SentryAspNetCoreOptionsSetup>();
+                _ = logging.Services.AddSingleton<ILoggerProvider, SentryAspNetCoreLoggerProvider>();
+
+                _ = logging.AddFilter<SentryAspNetCoreLoggerProvider>(
```
Vandersteen, line 94: There is no TryAddFilter, i'm not sure it is an issue if this is added twice thought

src/Sentry.AspNetCore/SentryHostBuilderExtensions.cs
```diff
@@ -0,0 +1,107 @@
...
+                    "Microsoft.AspNetCore.Diagnostics.ExceptionHandlerMiddleware",
+                    LogLevel.None);
+
+                var sentryBuilder = logging.Services.AddSentry();
+                configureSentry?.Invoke(context, sentryBuilder);
+            });
+
+            _ = builder.ConfigureServices((c, s) => _ = s.AddTransient<IStartupFilter, SentryStartupFilter>());
```
Vandersteen, line 102: TryAddTransient is also an issue here, if any IStartupFilter was already added (which most likely it will) this will not be registered

Vandersteen:
I dropped some comments here and there (see above)

In some code bases, I worked around this with a `private static bool` that keeps track if it is already called or not.
I could proceed and resolve those that way, so that calling it on both interfaces does not give strange issues

src/Sentry.AspNetCore/SentryHostBuilderExtensions.cs
bruno-garcia: That's true. Is there a way to check if the specific specification has been added? I believe it was possible to check the items in the collection so we can verify if specifically `SentryAspNetCoreLoggerProvider` was already added and make sure we only add it once.

src/Sentry.AspNetCore/SentryHostBuilderExtensions.cs
bruno-garcia: Not sure either, I guess something we'd need to test?  There's no way to look into the filters already added?

src/Sentry.AspNetCore/SentryHostBuilderExtensions.cs
bruno-garcia: Something lke:

```
            if (!s.Any(d =>  d.ServiceType == descriptor.ServiceType &&    d.GetImplementationType() == implementationType))

```

Vandersteen: Oh nice, I didn't think of that.
Just ran a quick test and yes that would work.

src/Sentry.AspNetCore/SentryHostBuilderExtensions.cs
Vandersteen: Looks like it boils down to: https://github.com/aspnet/Logging/blob/2d2f31968229eddb57b6ba3d34696ef366a6c71b/src/Microsoft.Extensions.Logging/FilterLoggingBuilderExtensions.cs#L75

bruno-garcia:
> In some code bases, I worked around this with a private static bool that keeps track if it is already called or not.
> I could proceed and resolve those that way, so that calling it on both interfaces does not give strange issues

This strategy won't work in the ASP.NET Core case because even on a normal run of it, 2 containers are created and the first time we'd add (before the static fields are set) would end up on a dead host, and the new one wouldn't have it.
See: https://github.com/bruno-garcia/repro-logger-provider-double-instantiation https://github.com/getsentry/sentry-dotnet/issues/103#issuecomment-433882680 https://github.com/aspnet/Hosting/issues/1499

src/Sentry.AspNetCore/SentryHostBuilderExtensions.cs
Vandersteen: Done

Vandersteen:
Took the comments into consideration, made  `AddSentry()` idempotent on

* IHostBuilder 
* IWebHostBuilder

src/Sentry.AspNetCore/Extensions/DependencyInjection/ServiceCollectionExtensions.cs
```diff
@@ -20,17 +22,55 @@ internal static class ServiceCollectionExtensions
...
+        }
+
+        /// <summary>
+        /// Adds Sentry's StartupFilter to the <see cref="IServiceCollection"/>
+        /// </summary>
+        /// <param name="serviceCollection"></param>
+        /// <returns></returns>
+        internal static IServiceCollection AddSentryStartupFilter(this IServiceCollection serviceCollection)
+        {
+            if(!serviceCollection.IsRegistered<IStartupFilter, SentryStartupFilter>())
```
bruno-garcia, line 68: nit: one day we'll have a linter that reformats and pushed to the branch :)

Could you please do this throughout the change?

```suggestion
            if (!serviceCollection.IsRegistered<IStartupFilter, SentryStartupFilter>())
```

bruno-garcia (commented):
I have just a nitpick but this already looks good to me!

Before we merge, It would be nice to have some coverage of this by some test, even an integration test of sorts.
I assume we have such test in the ASP.NET Core integration already then it could be copied over since it's the same thing.

It could also be used from one of the samples (Samples.ME.Logging or GenericHost actually)?
The samples make sure we have an easy way to run it and verify it works. And also to 'show ppl' some examples. And to have usage of the API to see what it looks like.

Vandersteen:
Fixed those `if (`'s I found 

I'll add some coverage and some samples

Vandersteen:
I had some trouble running it in .netcoreapp2.1 (IHostingEnvironment was unknown with HostBuilder)

I changed it to fix it, ran all (most) the aspnetcore samples, they still seem to work properly, however I can't check if the issues get to sentry

I changed the generic sample to use the new extension method.

Would you like me to make another sample for the Mvc's (using IHostBuilder instead), or shall I add it as a comment underneath ?

bruno-garcia:
Great stuff!

> Would you like me to make another sample for the Mvc's (using IHostBuilder instead), or shall I add it as a comment underneath ?

I wonder if we're just better off moving over to `IHostBuilder` on the `Sample.AspNetCore.Mvc` then. So we have one sample (`.Basic`) using `IWebHostBuilder` and the other using `IHostBuilder`. What do you think? 

Btw if you want to join us on Discord, we're online there often: http://discord.gg/sentry

Vandersteen:
hey @bruno-garcia i've tried to add some tests, but for some reason I can't seem to make them work.
I wasn't able to run these tests in debug mode either.

I keep getting: 

```
Diagnostic information:

Remaining (non-bound) argument specifications:
    any Action<IServiceCollection>

All argument specifications:
    any Action<IServiceCollection>

```

I don't understand how the `SentryWebHostBuilderTests` pass but not the `SentryHostBuilderTests`
I must be missing something silly.

Vandersteen:
Ok after a long debugging session, i finally found the issue.

IWebHostBuilder's interface contain
* ConfigureServices(Action<IServiceCollection> a);

While IHostBuilder does not, it has an extension method, that falls back to
* ConfigureServices(Action<HostBuilderContext, IServiceCollection> a)

Vandersteen:
@bruno-garcia @Tyrrrz Is there anything else I can do ?

Vandersteen:
Should've fixed failing tests

samples/Sentry.Samples.AspNetCore.Basic/Program.cs
```diff
@@ -70,5 +73,21 @@ public static void Main(string[] args)
...
+                // Add Sentry integration
+                // In this example, DSN and Release are set via environment variable:
+                // See: Properties/launchSettings.json
+                .UseSentry()
+                // It can also be defined via configuration (including appsettings.json)
+                // or coded explicitly, via parameter like:
+                // .UseSentry("dsn") or .UseSentry(o => o.Dsn = ""; o.Release = "1.0"; ...)
+                .ConfigureWebHost(webBuilder =>
+                {
+                    webBuilder.UseStartup<Startup>();
```
bruno-garcia, line 89: What about this? Sometimes I find in samples it's good to be very explicit about things:

```suggestion
                    // Note that it's NOT needed to also call UseSentry here. That's because it's already hooked through IHostBuilder
                    webBuilder.UseStartup<Startup>();
```

samples/Sentry.Samples.AspNetCore.Basic/Startup.cs
```diff
@@ -0,0 +1,70 @@
+using System;
```
bruno-garcia, line 1: I would prefer we remove this file. The idea of `Basic` is to have the simplest possible example.
I'd prefer we continued to use a single file for this example. We can even refactor this to use top level statements. 

We can remove the WebHost example and use only your new approach

samples/Sentry.Samples.GenericHost/Program.cs
```diff
@@ -18,12 +18,8 @@ public static Task Main()
...
             })
             .ConfigureServices((c, s) => { s.AddHostedService<SampleHostedService>(); })
-            .ConfigureLogging((c, l) =>
-            {
-                l.AddConfiguration(c.Configuration);
-                l.AddConsole();
-                l.AddSentry();
-            })
+            .ConfigureLogging(b => b.AddConsole())
+            .UseSentry()
```
bruno-garcia, line 22: Very nice!

samples/Sentry.Samples.GenericHost/Sentry.Samples.GenericHost.csproj
```diff
@@ -12,15 +12,15 @@
   </ItemGroup>
 
   <ItemGroup>
-    <ProjectReference Include="../../src/Sentry.Extensions.Logging/Sentry.Extensions.Logging.csproj" />
+    <ProjectReference Include="../../src/Sentry.AspNetCore/Sentry.AspNetCore.csproj" />
```
bruno-garcia, line 15: Do we need to change the dependency to ASP.NET Core integration? The new types are all on the S.E.L, right?
```suggestion
    <ProjectReference Include="../../src/Sentry.Extensions.Logging/Sentry.Extensions.Logging.csproj" />
```

src/Sentry.AspNetCore/SentryHostBuilderExtensions.cs
```diff
@@ -0,0 +1,93 @@
...
+    /// </summary>
+    [EditorBrowsable(EditorBrowsableState.Never)]
+    public static class SentryHostBuilderExtensions
+    {
+        /// <summary>
+        /// Uses Sentry integration.
+        /// </summary>
+        /// <param name="builder">The builder.</param>
+        /// <returns></returns>
+        public static IHostBuilder UseSentry(this IHostBuilder builder)
```
bruno-garcia, line 20: `IHostBuilder` lives in `Microsoft.Extensions.Hosting` so this could be added to `Sentry.Extensions.Logging`.

bruno-garcia (commented):
Thanks for this feature! There are a few minor things to address.
I'd even say it's likely good enough to merge and we can adjust things on our end. What would you say @Tyrrrz ?

src/Sentry.AspNetCore/SentryHostBuilderExtensions.cs
Vandersteen: I would've personally kept it in the asp.net core part, as it is the replacement of WebHostBuilder in the future.
Would you like me to move it ?

samples/Sentry.Samples.GenericHost/Sentry.Samples.GenericHost.csproj
Vandersteen: New types are not on S.E.L. right now, see comment below

samples/Sentry.Samples.AspNetCore.Basic/Startup.cs
Vandersteen: What do you think of the current version of basic ?

Tyrrrz:
I would also prefer to avoid having a dependency on `Sentry.AspNetCore` when one needs a generic host integration. It seems counter-intuitive. Sentry.Extensions.Logging is in a weird place right now, in a sense that it should probably be called Sentry.Shared or something, but I think it's a better place for the generic `IHostBuilder` stuff.

Vandersteen:
I understand

How do we proceed with these dependencies:

* Adding the Startupfilter / middleware 
* Calling the AddSentry() on servicecollection extensions
* SentryAspNetCoreOptions
* Microsoft.AspNetCore.Hosting

Tyrrrz:
I'm also okay with merging as is since the PR has been up for a while, and then improving on it as we go along 😉

Vandersteen:
I'll leave that up to you guys

bruno-garcia:
> I understand
> 
> How do we proceed with these dependencies:
> 
> * Adding the Startupfilter / middleware
> * Calling the AddSentry() on servicecollection extensions (and it's depedencies on aspnetcore)
> * SentryAspNetCoreOptions
> * Microsoft.AspNetCore.Hosting

Could we move all of that to `Sentry.Extensions.Logging`?
Oh I see, we'd need to add `Microsoft.AspNetCore.Hosting` as a dependency or:

Instead of AspNetCore.Hosting, isn't this `Microsoft.Extensions.Hosting`?

```
  <ItemGroup Condition="'$(TargetFramework)' != 'netstandard2.0'">
  </ItemGroup>
```

So it's either that or a new `Sentry.Extensions.Hosting` and have Sentry.AspNetCore depend on it?

Vandersteen:
I'll take a closer look tomorrow

Vandersteen:
Well I couldn't keep myself, took a quick gander.

Things like:
* IApplicationBuilder
* IStartupFilter
* HttpContext

All require AspNetCore ref

I believe you won't be able to substitute `IHostBuilder.UseSentry()` for `IWebHostBuilder.UseSentry()` completely if you go that path

`IHostBuilder.UseSentry()` could contain the minimum requirements in a `Sentry.Extensions.Hosting` but will not be a replacement for `IWebHostBuilder.UseSentry()`

I don't see anything wrong with that, if it's clear from the usage point of view.

Vandersteen:
@bruno-garcia Going forward, I think a new `Sentry.Extensions.Hosting` might be the best bet. Considering future .NET MAUI support

Do you wish for me to proceed on that path ?
Anything special I should know when setting up a new Project ?

bruno-garcia:
> @bruno-garcia Going forward, I believe a new `Sentry.Extensions.Hosting` might be the best bet. Considering future .NET MAUI support
> 
> Do you wish for me to proceed on that path ?
> Anything special I should know when setting up a new Project ?

Yeah sounds like we'll need it particularly because we're looking at adding support to MAUI as soon as it's out so we'd really appreciate your help with this!

Vandersteen:
Ok, thank you for the feedback.

I might have some time this week to continue working on this

mattjohnsonpint:
Thanks again for the time spent here, but we've decided not to do this.  See https://github.com/getsentry/sentry-dotnet/issues/190#issuecomment-1491065903

Thank you.

PR #1021, opened 2021-05-31
Hub.IsEnabled set to false when Hub disposed

semuserable:
`Hub.IsEnabled` is set to `false` when `Hub.Dispose()` is called. Related to https://github.com/getsentry/sentry-unity/issues/193 bug.

### DIFF
--- a/src/Sentry/Internal/Hub.cs
+++ b/src/Sentry/Internal/Hub.cs
@@ -249,6 +249,11 @@ namespace Sentry.Internal
         {
             _options.DiagnosticLogger?.LogInfo("Disposing the Hub.");
 
+            if (Interlocked.Exchange(ref _isEnabled, 0) != 1)
+            {
+                return;
+            }
+
             if (_integrations?.Length > 0)
             {
                 foreach (var integration in _integrations)
@@ -260,7 +265,6 @@ namespace Sentry.Internal
                 }
             }
 
-            Interlocked.Exchange(ref _isEnabled, 0);
             (_ownedClient as IDisposable)?.Dispose();
             _rootScope.Dispose();
             ScopeManager.Dispose();
--- a/src/Sentry/Internal/Hub.cs
+++ b/src/Sentry/Internal/Hub.cs
@@ -1,6 +1,7 @@
 using System;
 using System.Collections.Generic;
 using System.Runtime.CompilerServices;
+using System.Threading;
 using System.Threading.Tasks;
 using Sentry.Extensibility;
 using Sentry.Integrations;
@@ -19,7 +20,8 @@ namespace Sentry.Internal
 
         internal SentryScopeManager ScopeManager { get; }
 
-        public bool IsEnabled => true;
+        private int _isEnabled = 1;
+        public bool IsEnabled => _isEnabled == 1;
 
         internal Hub(ISentryClient client, SentryOptions options)
         {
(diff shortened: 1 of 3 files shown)


src/Sentry/Internal/Hub.cs
```diff
@@ -258,6 +260,7 @@ public void Dispose()
                 }
             }
 
+            Interlocked.Exchange(ref _isEnabled, 0);
```
Tyrrrz, line 263: Does this helps us in any way? I'm thinking because by this point we're already half way into disposing anyway, so if there was a race condition, it could have occurred already. Maybe just an `bool _isDisposed` field is enough? Which scenarios are we trying to protect against here?

bruno-garcia, line 263: It's just missing the check to avoid double dispose:

```suggestion
            if (Interlocked.Exchange(ref _isEnabled, 0) != 0)
            {
                return; // Already disposed
            }
```

bruno-garcia: But this must be the first thing (maybe after logging)

bruno-garcia:
Also, could you please add a test to validate that it in fact changes from true to false once u dispose?

src/Sentry/Internal/Hub.cs
semuserable: Oh, I see.

semuserable:
Updated. `Interlocked` moved up; Unit test added.

@bruno-garcia I added a test with `InlineData` to potentially check "over-disposing". I'm not sure it's needed for now, but still.

src/Sentry/Internal/Hub.cs
semuserable: Tbh, this is what we discussed with Bruno. My initial thought was the same, just `bool`, but Bruno insisted on `Interlocked`. From my understanding, the main scenario for this is that `Dispose` can be called by more than 1 thread.

test/Sentry.Tests/HubTests.cs
```diff
@@ -639,5 +639,26 @@ public void CaptureTransaction_AfterTransactionFinishes_ResetsTransactionOnScope
...
+            });
+
+            // Act
+            for (var i = 0; i < disposeCount; i++)
+            {
+                hub.Dispose();
+            }
+
+            // Assert
+            hub.IsEnabled.Should().BeFalse();
```
bruno-garcia, line 661: As discussed on our call:

```suggestion
            var hub = new Hub(new SentryOptions
            {
                Dsn = DsnSamples.ValidDsnWithSecret,
                
            });
            var client = Substibute.For<ISentryClient>();
            hub.BindClient(client);
            // Act
            for (var i = 0; i < disposeCount; i++)
            {
                hub.Dispose();
            }
            client.Received(1).Dispose();
            // Assert
            hub.IsEnabled.Should().BeFalse();
```

Another test would verify that `IsDispose` returns true at first, and after Dispose is called, it returns false

semuserable:
updated

cc @bruno-garcia @Tyrrrz

bruno-garcia (approved):

src/Sentry/Internal/Hub.cs
bruno-garcia: > but Bruno insisted on Interlocked

I don't recall a conversation where I insisted. 
What I recall is that I suggested a solution with `Interlocked` and there was no discussion of trade off.

Tyrrrz (approved):

PR #1022, opened 2021-05-31
Add non-breaking overloads for Serilog integration

Tyrrrz:


src/Sentry.Serilog/SentrySinkExtensions.cs
```diff
@@ -144,6 +144,112 @@ public static class SentrySinkExtensions
...
+        ///                         "key-2", "value-2"
+        ///                     }
+        ///                 }
+        ///             }
+        ///         ]
+        ///     }
+        /// }
+        /// </code>
+        /// </example>
+        public static LoggerConfiguration Sentry(
```
bruno-garcia, line 220: Can we add an obsolete to remove on the next major? Will require supressing warnings of the usage from tests
```suggestion
        [Obsolete("This method will be deleted on the next major. Please use the overload.")]
        public static LoggerConfiguration Sentry(
```

Tyrrrz:
Alright, looks like adding an overload causes a breaking change as well lol @bruno-garcia 
In this case, if the user doesn't provide any parameters to one of those methods, the call becomes ambiguous.

bruno-garcia:
Alright, thanks for trying

PR #1036, opened 2021-06-02
experimental wpf integration

bruno-garcia:
Starting from a sample to see what hooks we need.
The goal is to create a new package `Sentry.Wpf` that already hooks to `DispatcherUnhandledException`. Additionally it would set up offline caching by default, and create some transactions out of the box.

Once #1013 is done it can also emit release health stats out of the box.


jamescrosswell:
@SimonCropp just going through the old pull requests, looking to close any that are no longer active. Is this one still being worked on or should we close this?

PR #1037, opened 2021-06-02
Finish span with exception in SentryHttpMessageHandler

Tyrrrz:


bruno-garcia (commented):
Could you please add a test that verifies this behavior and avoid regression?

Tyrrrz:
> Could you please add a test that verifies this behavior and avoid regression?

Good point. Done.

bruno-garcia (approved):

PR #1039, opened 2021-06-03
Attempt to fix danger

### LINKED ISSUE #1004: Danger is broken
Danger doesn't work on external PRs, probably due to the token not being exposed.

image

Tyrrrz:
Not 100% sure what exactly is failing currently because the errors lack stacktraces, but looking at the previous version of the file (https://github.com/getsentry/sentry-dotnet/blob/04214d80af1c2be70450cc0c833828bc24ba164e/dangerfile.js) this seems to be the only thing that stands out.

Also can't really test it out until someone external makes a PR :/

image

#skip-changelog

Fixes #1004

### DIFF
--- a/dangerfile.js
+++ b/dangerfile.js
@@ -18,7 +18,7 @@ async function checkChangelog() {
     // Check if current PR has an entry in changelog
     const changelogContents = await danger.github.utils.fileContents(changelogFile);
 
-    const hasChangelogEntry = RegExp(`#${danger.github.pr.number}\\b`).testchangelogContents);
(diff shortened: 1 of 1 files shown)


lucas-zimerman:
forking the project and making the same PR wouldn't help for testing Danger behavior?

bruno-garcia (approved):

bruno-garcia:
@lucas-zimerman if I remove you from the repo, yes we can do that to test it out.
Or we can wait for the next external PR

PR #1045, opened 2021-06-05
Allow User with only IP address set

mlibk:
Currently the `DefaultUserFactory` requires `HttpContext.User.Identity` or `HttpContext.User.Claims` to be set in order to create the User object. If they're not set, the IP address of the web client isn't used and eventually gets overridden with `{{auto}}` which isn't a desirable default behaviour for ASP.NET Core as the server ip will be reported as the user IP.

This PR allows User to be created also if the client IP address is the only property available and this will result in a more desirable default behaviour for ASP.NET Core apps as client address should (almost?) always be set.

### DIFF
--- a/src/Sentry.AspNetCore/DefaultUserFactory.cs
+++ b/src/Sentry.AspNetCore/DefaultUserFactory.cs
@@ -9,7 +9,7 @@ namespace Sentry.AspNetCore
         public User? Create(HttpContext context)
         {
             var principal = context.User;
-            if (principal == null)
+            if (principal is null)
             {
                 return null;
             }
@@ -36,20 +36,21 @@ namespace Sentry.AspNetCore
             // Identity.Name Reads the value of: ClaimsIdentity.NameClaimType which by default is ClaimTypes.Name
             // It can be changed by the application to read a different claim though:
             var name = principal.Identity?.Name;
-            if (name != null && username != name)
+            if (name is not null && username != name)
             {
                 username = name;
             }
 
-            // Don't create a user if all we have is his IP address
-            return email == null && id == null && username == null
+            var ipAddress = context.Connection?.RemoteIpAddress?.ToString();
+
+            return email is null && id is null && username is null && ipAddress is null
                 ? null
                 : new User
                 {
                     Id = id,
                     Email = email,
                     Username = username,
-                    IpAddress = context.Connection?.RemoteIpAddress?.ToString()
+                    IpAddress = ipAddress,
                 };
         }
     }
--- a/src/Sentry.AspNetCore/DefaultUserFactory.cs
+++ b/src/Sentry.AspNetCore/DefaultUserFactory.cs
@@ -36,7 +36,7 @@ namespace Sentry.AspNetCore
             // Identity.Name Reads the value of: ClaimsIdentity.NameClaimType which by default is ClaimTypes.Name
             // It can be changed by the application to read a different claim though:
             var name = principal.Identity?.Name;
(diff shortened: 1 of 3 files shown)


src/Sentry.AspNetCore/DefaultUserFactory.cs
```diff
@@ -36,20 +36,21 @@ internal class DefaultUserFactory : IUserFactory
             // Identity.Name Reads the value of: ClaimsIdentity.NameClaimType which by default is ClaimTypes.Name
             // It can be changed by the application to read a different claim though:
             var name = principal.Identity?.Name;
-            if (name != null && username != name)
+            if (name is not null && username != name)
```
bruno-garcia, line 39: It's fine to do the refactor in this change, lets go all the way then for consistency sake:

```suggestion
            if (name is not null && username is not name)
```

bruno-garcia (commented):
Just a nit but other than that LGTM! 
Thanks

bruno-garcia (approved):

bruno-garcia:
Actually could you please add an entry to  `CHANGELOG.md`:

```
### Features

- Report user IP address for ASP.NET Core ([#1045](https://github.com/getsentry/sentry-unity/pull/1045))
```

src/Sentry.AspNetCore/DefaultUserFactory.cs
```diff
@@ -36,20 +36,21 @@ internal class DefaultUserFactory : IUserFactory
...
             var name = principal.Identity?.Name;
-            if (name != null && username != name)
+            if (name is not null && username is not name)
             {
                 username = name;
             }
 
-            // Don't create a user if all we have is his IP address
-            return email == null && id == null && username == null
+            var ipAddress = context.Connection?.RemoteIpAddress?.ToString();
```
bruno-garcia, line 44: If the server is behind a reverse proxy it would need to consider `X-Forwarded-For`.
We need to add something to the docs to mention how this works and how a user can 'override' this (basically using a `IUserFactory`

mlibk: I think the proper way to handle this is to add `UseForwardedHeaders()` middleware as per [asp.net core docs](https://docs.microsoft.com/en-us/aspnet/core/host-and-deploy/proxy-load-balancer?view=aspnetcore-5.0) which will update `HttpContext.Connection.RemoteIpAddress`. It might be worth mentioning it in the docs though.

bruno-garcia: Or to be more precise we should document that we get the IP Address through `HttpContext.Connection.RemoteIpAddress` and if they have a reverse proxy, they need to add a middleware to deal with that by linking to the docs:

https://docs.microsoft.com/en-us/aspnet/core/host-and-deploy/proxy-load-balancer?view=aspnetcore-5.0#forwarded-headers-middleware-order

src/Sentry.AspNetCore/DefaultUserFactory.cs
```diff
@@ -36,20 +36,21 @@ internal class DefaultUserFactory : IUserFactory
             // Identity.Name Reads the value of: ClaimsIdentity.NameClaimType which by default is ClaimTypes.Name
             // It can be changed by the application to read a different claim though:
             var name = principal.Identity?.Name;
-            if (name != null && username != name)
+            if (name is not null && username is not name)
```
bruno-garcia, line 39:
```suggestion
            if (name is not null && username != name)
```

PR #1048, opened 2021-06-08
Make debug diagnostic logger obsolete and add Trace diagnostic logger.

### LINKED ISSUE #1040: Usage of `DebugDiagnosticLogger` in the shipped SDK will be no-op
`DebugDiagnosticLogger` works when developing the SDK as we compile in debug mode.
Once we ship the SDK, we're building in Release. The `Debug` class won't be compiled so basically `DebugDiagnosticLogger` is no-op.

We need to revert #1012 and rethink the approach with shipping a logger that makes sense when `Console` isn't available.

lucas-zimerman:
DebugDiagnosticLogger only works with projects inside Sentry.NET.snl since Release builds remove the Debug code, making it useless to the end-user.
Another approach is to use Trace as a replacement, it's not as simple as the good old DebugDiagnosticLogger but it should work just fine even in Release mode.

Close #1040.

It would be great if I could simply use Trace.Write but it seems like the compiler gets rid of it if not called from the current process.

EDIT: Validated on
* .NET 5 Console
* .NET Core with IIS 
* .NET Native UWP

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## Unreleased
 
-- Make debug diagnostic logger obsolete and add Trace diagnostic logger ([#1048](https://github.com/getsentry/sentry-dotnet/pull/1048))
+- Introduce TraceDiagnosticLogger and obsolete DebugDiagnosticLogger ([#1048](https://github.com/getsentry/sentry-dotnet/pull/1048))
 
 ## 3.5.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## Unreleased
+
+- Make debug diagnostic logger obsolete and add Trace diagnostic logger ([#1048](https://github.com/getsentry/sentry-dotnet/pull/1048))
+
 ## 3.5.0
 
 ### Features
--- a/src/Sentry.AspNet/SentryAspNetOptionsExtensions.cs
+++ b/src/Sentry.AspNet/SentryAspNetOptionsExtensions.cs
@@ -22,7 +22,7 @@ namespace Sentry.AspNet
 
             var eventProcessor = new SystemWebRequestEventProcessor(payloadExtractor, options);
 
-            options.DiagnosticLogger ??= new DebugDiagnosticLogger(options.DiagnosticLevel);
+            options.DiagnosticLogger ??= new TraceDiagnosticLogger(options.DiagnosticLevel);
             options.Release ??= SystemWebVersionLocator.GetCurrent();
             options.AddEventProcessor(eventProcessor);
         }
--- a/src/Sentry/Infrastructure/DebugDiagnosticLogger.cs
+++ b/src/Sentry/Infrastructure/DebugDiagnosticLogger.cs
@@ -10,6 +10,7 @@ namespace Sentry.Infrastructure
     /// <remarks>
     /// Logger available when compiled in Debug mode. It's useful when debugging apps running under IIS which have no output to Console logger.
     /// </remarks>
+    [Obsolete("Logger doesn't work outside of Sentry SDK. Please use TraceDiagnosticLogger instead")]
     public class DebugDiagnosticLogger : IDiagnosticLogger
     {
         private readonly SentryLevel _minimalLevel;
--- a/src/Sentry/Infrastructure/TraceDiagnosticLogger.cs
+++ b/src/Sentry/Infrastructure/TraceDiagnosticLogger.cs
@@ -33,8 +33,12 @@ namespace Sentry.Infrastructure
             {
                 for (var i = 0; i < Trace.Listeners.Count; i++)
                 {
-                    Trace.Listeners[i].WriteLine($@"{logLevel,7}: {string.Format(message, args)}
-{exception}");
+                    var listener = Trace.Listeners[i];
+                    listener.WriteLine($"{logLevel,7}: {string.Format(message, args)}");
+                    if (exception != null)
+                    {
+                        listener.WriteLine(exception);
+                    }
                 }
             }
             catch (ArgumentOutOfRangeException)
--- a/src/Sentry/Infrastructure/TraceDiagnosticLogger.cs
+++ b/src/Sentry/Infrastructure/TraceDiagnosticLogger.cs
@@ -1,4 +1,5 @@
 using System;
+using System.Collections.Generic;
 using System.Diagnostics;
 using Sentry.Extensibility;
 namespace Sentry.Infrastructure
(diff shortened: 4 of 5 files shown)


src/Sentry.AspNet/SentryAspNetOptionsExtensions.cs
```diff
@@ -22,7 +22,7 @@ public static void AddAspNet(this SentryOptions options, RequestSize maxRequestB
 
             var eventProcessor = new SystemWebRequestEventProcessor(payloadExtractor, options);
 
-            options.DiagnosticLogger ??= new DebugDiagnosticLogger(options.DiagnosticLevel);
+            options.DiagnosticLogger ??= new TraceDiagnosticLogger(options.DiagnosticLevel);
```
lucas-zimerman, line 25: I'd replace the default logger with TraceDiagnosticLogger since it doesn't pollute the terminal screen if your application is console-based.

src/Sentry/Infrastructure/TraceDiagnosticLogger.cs
```diff
@@ -0,0 +1,47 @@
...
+        private Lazy<TraceListener?> _traceListener = new Lazy<TraceListener?>(() => GetFirstOrDefaultListener());
+
+        /// <summary>
+        /// Get the first TraceListener, usually it's the process debug output.
+        /// </summary>
+        /// <returns>The first TraceListener, if available</returns>
+        private static TraceListener? GetFirstOrDefaultListener()
+        {
+            var listeners = Trace.Listeners;
+            return listeners.Count > 0 ? listeners[0] : null;
```
lucas-zimerman, line 27: It works, not a fan of this line and linq doesn't work with listeners.

Tyrrrz: Maybe we should write to **all** listeners, not just one?

src/Sentry/Infrastructure/DebugDiagnosticLogger.cs
```diff
@@ -10,6 +10,7 @@ namespace Sentry.Infrastructure
     /// <remarks>
     /// Logger available when compiled in Debug mode. It's useful when debugging apps running under IIS which have no output to Console logger.
     /// </remarks>
+    [Obsolete("Logger doesn't work outside of Sentry SDK. Please use TraceDiagnosticLogger instead")]
```
Tyrrrz, line 13: Would it be fair to say...

```suggestion
    [Obsolete("This logger only works in debug mode. Please use TraceDiagnosticLogger instead.")]
```

Or does it not work even if consumed from a project built in debug?

src/Sentry/Sentry.csproj
```diff
@@ -7,7 +7,7 @@
     <RootNamespace>Sentry</RootNamespace>
     <Description>Official SDK for Sentry - Open-source error tracking that helps developers monitor and fix crashes in real time.</Description>
     <NoWarn Condition="$(TargetFramework) == 'netstandard2.0'">$(NoWarn);RS0017</NoWarn>
-    <DefineConstants>$(AdditionalConstants)</DefineConstants>
+    <DefineConstants>$(AdditionalConstants);TRACE</DefineConstants>
```
Tyrrrz, line 10: Does the trace logger not work without `TRACE` constant?

Tyrrrz (approved):

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+## Unreleased
+
+- Make debug diagnostic logger obsolete and add Trace diagnostic logger ([#1048](https://github.com/getsentry/sentry-dotnet/pull/1048))
```
bruno-garcia, line 5: We need a sub heading (due to the tool we use for changlog in nuget):
Also, better simply the message:

```suggestion
- Introduce TraceDiagnosticLogger and obsolete DebugDiagnosticLogger ([#1048](https://github.com/getsentry/sentry-dotnet/pull/1048))
```

src/Sentry.AspNet/SentryAspNetOptionsExtensions.cs
bruno-garcia: Logging is opt-in so nothing gets poluted unless u turn it on.

src/Sentry/Infrastructure/DebugDiagnosticLogger.cs
bruno-garcia: I believe the call to `Debug.Write` doesn't get compiled when we compile our SDk in Release mode so my guess is that it won't work. But curious if I'm just wrong here

src/Sentry/Sentry.csproj
bruno-garcia: I wonder if we're adding overhead to the SDK by setting this. We're not explicitly adding anything that depends on `TRACE` constant being defined so I assume it wouldn't matter but it does make me wonder. Does it affect the size of the binary at all for example?

src/Sentry/Infrastructure/DebugDiagnosticLogger.cs
Tyrrrz: You are right. I checked and it gets completely omitted by the compiler.

src/Sentry/Sentry.csproj
lucas-zimerman: Trace.Write/WriteLine doesn't work without the `TRACE` trace constant. Trace.Listeners are exposed with or without the TRACE constant, but in my opinion, it's better to use set the TRACE constant and using the build-in Trace.Write instead of reinventing the wheel.

The package sizes are:
1,076,098 bytes with the TRACE constant
1,076,158 bytes without the TRACE constant and without the Trace Logger file
1,077,102 bytes without it.
1,077,858 bytes without the trace constant with the addition of a loop to write to all listeners.
1,080,993 bytes adding the DEBUG constant (out of curiosity, this allows the debug diagnostic logger to work properly in release)

Tyrrrz (approved):

bruno-garcia:
@lucas-zimerman and I talked on a call and decided to drop the constant and use the listeners. This is a logger for debug purposes and we don't wanta to add a trace constant to a release build

src/Sentry/Infrastructure/TraceDiagnosticLogger.cs
```diff
@@ -0,0 +1,41 @@
...
+        public void Log(SentryLevel logLevel, string message, Exception? exception = null, params object?[] args)
+        {
+            lock (Trace.Listeners)
+            {
+                for (int index = 0; index < Trace.Listeners.Count; index++)
+                {
+                    Trace.Listeners[index].Write($@"{logLevel,7}: {string.Format(message, args)}
+{exception}");
+                }
+            }
```
bruno-garcia, line 38:
```suggestion
            try
            {
                for (var i = 0; i < Trace.Listeners.Count; i++)
                {
                    Trace.Listeners[i].WriteLine($@"{logLevel,7}: {string.Format(message, args)}
{exception}");
                }

            }
            catch (IndexOutOfRangeException)
            {
                // Trace.Listeners is a public mutable static property and listeners can be removed
                // from anywhere at anytime.
            }
```

lucas-zimerman: It throws ArgumentOutOfRangeException instead of IndexOutOfRangeException

bruno-garcia (approved):

PR #1057, opened 2021-06-15
Added SentryAspNetCoreOptions.AdjustStandardEnvironmentNameCasing to …

### LINKED ISSUE #999: Request: please don't change case of environment
In the .NET SDK, Sentry will change the case of the environment if the IHostingEnvironment.EnviromentName matches either "Production" or "Development". Any other environment names will not be changed. 

https://github.com/getsentry/sentry-dotnet/blob/a9304a0a4b4702d0e62e2703d55c66483d27a0e5/src/Sentry.AspNetCore/SentryAspNetCoreOptionsSetup.cs#L39-L56

I find this pretty strange, unexpected behaviour. I am migrating from .NET Framework to .NET Core and in so doing, want to stop explicitly specifying the environment and rely on the value specified in IHostingEnvironment. If I do this, due to this case change, I will then have the following environments listed in Sentry:

- Production (now unused)
- production
- Development (now unused)
- development
- Test

I can hide the unused environments though this could cause problems viewing older issues. It's also a bit frustrating that the casing on my environments is no longer consistent (considering the Test environment). I admit these are minor inconveniences and I can just explicitly specify the environment to avoid this, but they are frustrations all the same. I can't think of a good reason why this undocumented behaviour exists and think generally it's best to display environments as they have been specified by the developer/framework.

The comment says that Sentry prefers lowercase environment names (though I haven't seen this echoed in the documentation). Is there any reason for this preference?

ajbeaven:
…have sentry avoid modifying the case of the standard ASPNET_ENVIRONMENT names when determining the sentry environment.

Fixes #999.

### DIFF
--- a/src/Sentry.AspNetCore/SentryAspNetCoreOptions.cs
+++ b/src/Sentry.AspNetCore/SentryAspNetCoreOptions.cs
@@ -40,7 +40,7 @@ namespace Sentry.AspNetCore
         public TimeSpan FlushTimeout { get; set; } = TimeSpan.FromSeconds(2);
 
         /// <summary>
-        /// Controls whether the casing of the standard (Production, Development and Staging) environment name supplied by <see cref="Microsoft.AspNetCore.Hosting.HostingEnvironment" />
+        /// Controls whether the casing of the standard (Production, Development and Staging) environment name supplied by <see cref="Microsoft.AspNetCore.Hosting.IHostingEnvironment" />
         /// is adjusted when setting the Sentry environment. Defaults to true.
         /// </summary>
         /// <remarks>
--- a/src/Sentry.AspNetCore/SentryAspNetCoreOptions.cs
+++ b/src/Sentry.AspNetCore/SentryAspNetCoreOptions.cs
@@ -40,11 +40,11 @@ namespace Sentry.AspNetCore
         public TimeSpan FlushTimeout { get; set; } = TimeSpan.FromSeconds(2);
 
         /// <summary>
-        /// Controls whether the case of the standard (Production, Development and Staging) EnvironmentName supplied by HostingEnvironment
+        /// Controls whether the casing of the standard (Production, Development and Staging) environment name supplied by <see cref="Microsoft.AspNetCore.Hosting.HostingEnvironment" />
         /// is adjusted when setting the Sentry environment. Defaults to true.
         /// </summary>
         /// <remarks>
-        /// The default .NET Core EnvironmentNames include Production, Development and Staging (note Pascal casing), whereas sentry prefers
+        /// The default .NET Core environment names include Production, Development and Staging (note Pascal casing), whereas Sentry prefers
         /// to have its environment setting be all lower case.
         /// </remarks>
         public bool AdjustStandardEnvironmentNameCasing { get; set; } = true;
--- a/src/Sentry.AspNetCore/SentryAspNetCoreOptions.cs
+++ b/src/Sentry.AspNetCore/SentryAspNetCoreOptions.cs
@@ -40,6 +40,16 @@ namespace Sentry.AspNetCore
         public TimeSpan FlushTimeout { get; set; } = TimeSpan.FromSeconds(2);
 
         /// <summary>
+        /// Controls whether the case of the standard (Production, Development and Staging) EnvironmentName supplied by HostingEnvironment
+        /// is adjusted when setting the Sentry environment. Defaults to true.
+        /// </summary>
+        /// <remarks>
+        /// The default .NET Core EnvironmentNames include Production, Development and Staging (note Pascal casing), whereas sentry prefers
+        /// to have its environment setting be all lower case.
+        /// </remarks>
+        public bool AdjustStandardEnvironmentNameCasing { get; set; } = true;
+
+        /// <summary>
         /// Creates a new instance of <see cref="SentryAspNetCoreOptions"/>.
         /// </summary>
         public SentryAspNetCoreOptions()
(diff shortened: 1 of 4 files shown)


src/Sentry.AspNetCore/SentryAspNetCoreOptionsSetup.cs
```diff
@@ -38,31 +38,39 @@ public override void Configure(SentryAspNetCoreOptions options)
...
-                    //             Alternatively, developers might set this to a CUSTOM value, which we
-                    //             need to respect (especially the case-sensitivity).
-                    //             REF: https://docs.microsoft.com/en-us/aspnet/core/fundamentals/environments
-
-                    if (_hostingEnvironment.IsProduction())
-                    {
-                        options.Environment = Internal.Constants.ProductionEnvironmentSetting;
-                    }
-                    else if (_hostingEnvironment.IsStaging())
+                    if (options.AdjustStandardEnvironmentNameCasing)
```
Tyrrrz, line 41: Can we merge this with the above `else` into `else if`?

src/Sentry.AspNetCore/SentryAspNetCoreOptions.cs
```diff
@@ -39,6 +39,16 @@ public class SentryAspNetCoreOptions : SentryLoggingOptions
         /// </summary>
         public TimeSpan FlushTimeout { get; set; } = TimeSpan.FromSeconds(2);
 
+        /// <summary>
+        /// Controls whether the case of the standard (Production, Development and Staging) EnvironmentName supplied by HostingEnvironment
```
Tyrrrz, line 43:
```suggestion
        /// Controls whether the casing of the standard (Production, Development and Staging) environment name supplied by <see cref="Microsoft.AspNetCore.Hosting.HostingEnvironment" />
```

src/Sentry.AspNetCore/SentryAspNetCoreOptions.cs
```diff
@@ -39,6 +39,16 @@ public class SentryAspNetCoreOptions : SentryLoggingOptions
...
+        /// is adjusted when setting the Sentry environment. Defaults to true.
+        /// </summary>
+        /// <remarks>
+        /// The default .NET Core EnvironmentNames include Production, Development and Staging (note Pascal casing), whereas sentry prefers
```
Tyrrrz, line 47:
```suggestion
        /// The default .NET Core environment names include Production, Development and Staging (note Pascal casing), whereas Sentry prefers
```

test/Sentry.AspNetCore.Tests/SentryAspNetCoreOptionsSetupTests.cs
```diff
@@ -42,18 +42,24 @@ public void Filters_KestrelEventId1_WithException_NotFiltered()
...
+        [InlineData("", "Development", false, "Development")] // Custom - nothing set. Adjust standard casing - false. ASPNET_ENVIRONMENT is default DEV.
+        [InlineData(null, "Staging", false, "Staging")] // Custom - nothing set. Adjust standard casing - false. ASPNET_ENVIRONMENT is default STAGING.
+        [InlineData("", "Staging", false, "Staging")] // Custom - nothing set. Adjust standard casing - false. ASPNET_ENVIRONMENT is default STAGING.
+        [InlineData(null, "production", true, "production")] // Custom - nothing set. Adjust standard casing - true. ASPNET_ENVIRONMENT is custom (notice lowercase 'p').
+        [InlineData(null, "development", true, "development")] // Custom - nothing set. Adjust standard casing - true. ASPNET_ENVIRONMENT is custom (notice lowercase 'd').
+        [InlineData(null, "staging", true, "staging")] // Custom - nothing set. Adjust standard casing - true. ASPNET_ENVIRONMENT is custom (notice lowercase 's').
+        public void Filters_Environment_CustomOrASPNETEnvironment_Set(string environment, string hostingEnvironmentSetting, bool adjustStandardEnvironmentNameCasingSetting, string expectedEnvironment)
```
Tyrrrz, line 62: Can we move the new stuff to a separate test to keep it focused? Can also get away with much fewer inline data test cases too.

Tyrrrz:
Also please update `Changelog.md` with something like this:

```
- Added SentryAspNetCoreOptions.AdjustStandardEnvironmentNameCasing to control whether to transform the environment name to lower case or to keep it as is. [#1057](https://github.com/getsentry/sentry-dotnet/pull/1057)
```

src/Sentry.AspNetCore/SentryAspNetCoreOptions.cs
ajbeaven: Hmm, VS complains with this:
```
XML comment has cref attribute 'HostingEnvironment' that could not be resolved	
Sentry.AspNetCore (netcoreapp3.0), Sentry.AspNetCore (netstandard2.0)
```

test/Sentry.AspNetCore.Tests/SentryAspNetCoreOptionsSetupTests.cs
ajbeaven: I've pulled them out, but there are still a few test cases to cover every scenario. Happy for you to remove any you don't think are necessary.

src/Sentry.AspNetCore/SentryAspNetCoreOptions.cs
ajbeaven: A reference to `Microsoft.AspNetCore.Hosting.IHostingEnvironment` (which is more accurate anyway) worked, so I'll go with that.

Tyrrrz (approved):

Tyrrrz:
@bruno-garcia looks good?

bruno-garcia (approved):
Beautiful! Thanks a lot @ajbeaven

PR #1059, opened 2021-06-15
Revert changes to dangerfile.js

### LINKED ISSUE #1004: Danger is broken

Tyrrrz:
Revert danger to its original state, but update the comment message.
Hopefully fixes #1004.

#skip-changelog

### DIFF
--- a/dangerfile.js
+++ b/dangerfile.js
@@ -1,76 +1,97 @@
-async function checkDocs() {
-    if (danger.github.pr.title.startsWith("feat:")) {
-        message(
-            'Do not forget to update <a href="https://github.com/getsentry/sentry-docs">Sentry-docs</a> with your feature once the pull request gets approved.'
-        );
-    }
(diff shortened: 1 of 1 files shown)


dangerfile.js
```diff
@@ -1,76 +1,97 @@
...
+async function containsChangelog(path) {
+  const contents = await danger.github.utils.fileContents(path);
+  return contents.includes(PR_LINK);
+}
 
-- ${prTitleFormatted} ([#${danger.github.pr.number}](${danger.github.pr.html_url}))
-\`\`\`
+async function checkChangelog() {
+  const skipChangelog =
+    danger.github && (danger.github.pr.body + "").includes("#skip-changelog");
```
lucas-zimerman, line 54: It would be nice in the future that skip-change log would be more flexible, not being case sensitive and allowing the `_` symbol, so folks can write `#SKIP-CHANGELOG` or `#skip_changelog` but that could be done in another PR

PR #1060, opened 2021-06-16
Refactor JSON serialization

### LINKED ISSUE #631: Refactor manual JSON serialization

Tyrrrz:
Closes #631 
#skip-changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Changelog
 
-## Unreleased
+## 3.6.0-alpha.1
 
 ### Features
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,16 +1,19 @@
 # Changelog
 
-## Unreleased
+## 3.6.0-alpha.2
 
 ### Features
 
+- Extended Device and GPU protocol; public IJsonSerializable ([#1063](https://github.com/getsentry/sentry-dotnet/pull/1063))
 - ASP.NET Core: Option `AdjustStandardEnvironmentNameCasing` to opt-out from lower casing env name. [#1057](https://github.com/getsentry/sentry-dotnet/pull/1057)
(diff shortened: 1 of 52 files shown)


Tyrrrz:
@bruno-garcia I think this is as far as it goes

semuserable:
Can we make `JsonExtensions` public? We use some extensions in Unity. Copied them for now, but would be great if we just use the ones from Sentry .NET SDK.

cc @Tyrrrz @bruno-garcia

Tyrrrz:
> Can we make `JsonExtensions` public? We use some extensions in Unity. Copied them for now, but would be great if we just use the ones from Sentry .NET SDK.
> 
> cc @Tyrrrz @bruno-garcia

@bruno-garcia You mentioned that you've reached some conclusion regarding this

Tyrrrz:
@semuserable we had a discussion and decided to not make `JsonExtensions` public. The reason is that it will be too easy to accidentally add them in user code (because they will be visible in code completion drop down) and because that will introduce extra liability that we'd prefer to avoid. So for now copying is the safest approach.

Long term we may move these utility extensions along with some other stuff in a separate submodule that can be referenced in both the main SDK and Unity. /cc @bruno-garcia

semuserable:
> @semuserable we had a discussion and decided to not make `JsonExtensions` public. The reason is that it will be too easy to accidentally add them in user code (because as extension methods they will always be visible in code completion drop down) and because that will introduce extra liability that we'd prefer to avoid. So for now copying is the safest approach.
> 
> Long term we may move these utility extensions along with some other stuff in a separate submodule that can be referenced in both the main SDK and Unity. /cc @bruno-garcia

Got you. Thanks for an update!

PR #1068, opened 2021-06-18
Add auto session tracking

### LINKED ISSUE #1054: Application mode Sessions by default (between Init and Close)
Sentry Native introduced a session mode called "application mode session". It's defined by a session which starts when the SDK is initialized and it ends when the process exits.

Docs for it are here: https://docs.sentry.io/platforms/native/configuration/releases/#release-health

> typically just a single session from application start to quit

More on the spec: https://develop.sentry.dev/sdk/sessions/

We need to make sure that:

- [x] Add an option: `EnableAutoSessionTracking` with default `false` until our next major, where we'll flip the default to `true`.
- [x] Unhandled exceptions (marked as `handled: false`) finish the session as `crashed`.
- [x] When the SDK is disposed, sessions are closed

Tyrrrz:
Related to #1054
Implemented through `ISdkIntegration`.

Tests omitted.

Note the issue says:

> Add an option: EnableAutoSessionTracking with default false until our next major, where we'll flip the default to true.

Should we do that? I added it as `true` by default because I thought that was the consensus.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@
 - ASP.NET Core: Option `AdjustStandardEnvironmentNameCasing` to opt-out from lower casing env name. [#1057](https://github.com/getsentry/sentry-dotnet/pull/1057)
 - Sessions: Improve exception check in `CaptureEvent(...)` for the purpose of reporting errors in session ([#1058](https://github.com/getsentry/sentry-dotnet/pull/1058))
 - Introduce TraceDiagnosticLogger and obsolete DebugDiagnosticLogger ([#1048](https://github.com/getsentry/sentry-dotnet/pull/1048))
-- Add auto session tracking([#1068](https://github.com/getsentry/sentry-dotnet/pull/1068))
+- Add auto session tracking ([#1068](https://github.com/getsentry/sentry-dotnet/pull/1068))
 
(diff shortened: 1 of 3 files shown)


bruno-garcia:
> > Add an option: EnableAutoSessionTracking with default false until our next major, where we'll flip the default to true.

> Should we do that? I added it as true by default because I thought that was the consensus.

I thought it would be best if we only started sending session data on a major release (as we did for Android, JavaScript, iOS) and until then had it as a opt-in feature.

On a server application we'd definitely want that turned off so we need to add to the ASP.NET and ASP.NET Core packages as `false`.

We can always flip the flag later on, lets start with it as `false`.

bruno-garcia (commented):
Looks good other than we should default to `false` and again on aspnet and aspnetcore default to false again because we'll flip this to `true` on the core next major.

On the comment we should say: Default is `false` but will be changed to `true` on the next major release.

Tyrrrz:
Updated. Not sure why the tests are failing (seems unrelated but who knows). I don't really have time to debug it today, but I will look into it next week (unless you want to poke at it yourself).

bruno-garcia (approved):

bruno-garcia:
CI passed. thanks

PR #1069, opened 2021-06-18
Implement pause & resume session

### LINKED ISSUE #1061: Session tracking Pause and Resume
We'd like to expose from the core Sentry the Pause and Resume session operations.

It should rely on the `TimeSpan SessionTrackingInterval` to control whether a "Resume" should result in a new session being created (if the time paused is greater than the interval).

Start -> Pause -> Resume: with the time on the paused shorter than `SessionTrackingInterval` results in 1 session
Start -> Pause -> Resume: with the time on the paused longer than `SessionTrackingInterval` results in 2 session (last one open)
Resume -> without any start call, it should behave like Start
Pause -> Resume: Pause will noop, and resume like above ^

Please add a tag: `session.state`: `active` or `paused`.

Mobile SDKs also add a breadcrumb. But we decided this makes more sense to be added by the integrations that will call into these methods (Pause and Resume) so the crumb information can be more specific (i.e: Unity can say: `MonoBehavior.OnPause` called or say "backgrund and foreground" instead of just pause and resume.

Example crumb from Mobile SDKs:

image

Example JSON:

```
{
  "timestamp": 1623809481.848,
  "type": "navigation",
  "category": "app.lifecycle",
  "level": "info",
  "data": {
    "state": "foreground"
  }
```

Relates to: https://github.com/getsentry/sentry-dotnet/issues/1054 as it relies on auto sessions

Tyrrrz:
Closes #1061

### DIFF
--- a/src/Sentry/GlobalSessionManager.cs
+++ b/src/Sentry/GlobalSessionManager.cs
@@ -193,7 +193,7 @@ namespace Sentry
                 );
 
                 // End previous session
-                EndSession(previousSession, SessionEndStatus.Exited, _clock.GetUtcNow());
+                EndSession(previousSession, _clock.GetUtcNow(), SessionEndStatus.Exited);
             }
 
             _options.DiagnosticLogger?.LogInfo(
@@ -204,19 +204,17 @@ namespace Sentry
             return session.CreateUpdate(true);
         }
 
-        private SessionUpdate EndSession(Session session, SessionEndStatus status, DateTimeOffset timestamp)
+        private SessionUpdate EndSession(Session session, DateTimeOffset timestamp, SessionEndStatus status)
         {
-            session.End(status);
-
             _options.DiagnosticLogger?.LogInfo(
                 "Ended session (SID: {0}; DID: {1}) with status '{2}'.",
                 session.Id, session.DistinctId, status
             );
 
-            return session.CreateUpdate(false, timestamp);
+            return session.CreateUpdate(false, timestamp, status);
         }
 
-        public SessionUpdate? EndSession(SessionEndStatus status, DateTimeOffset timestamp)
+        public SessionUpdate? EndSession(DateTimeOffset timestamp, SessionEndStatus status)
         {
             var session = Interlocked.Exchange(ref _currentSession, null);
             if (session is null)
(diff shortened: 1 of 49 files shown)


src/Sentry/GlobalSessionManager.cs
```diff
@@ -234,7 +220,22 @@ private SessionUpdate EndSession(Session session, SessionEndStatus status)
...
-            return EndSession(session, status);
+            return EndSession(session, status, timestamp);
+        }
+
+        public SessionUpdate? ReportError()
+        {
+            if (_currentSession is { } session)
+            {
+                session.ReportError();
+                return session.CreateUpdate(false);
```
bruno-garcia, line 231: This could happen after the session is removed from the scope in order to be closed. So technically we could be sending session updates after a closed session.

We're probably safer using a full lock around any of these methods that return `SessionUpdate` to make sure once we close the session, no session updates are created for it.

bruno-garcia (commented):
shaping up nicely

src/Sentry/Internal/Hub.cs
```diff
@@ -182,11 +186,38 @@ public void StartSession()
...
+            {
+                _sessionPauseTimestamp = DateTimeOffset.Now;
+            }
+        }
+
+        public void ResumeSession()
+        {
+            lock (_sessionPauseLock)
+            {
+                var pauseDuration = (DateTimeOffset.Now - _sessionPauseTimestamp).Duration();
```
bitsandfoxes, line 201: Does it make sense to provide ISystemClock to the hub for testability, to fake the pause?

src/Sentry/GlobalSessionManager.cs
Tyrrrz: I'm not sure how a lock here would help, because session manager is not responsible for sending session updates, the Hub is.

bruno-garcia: All we can guarantee from here is that the order of calls into this type will return session updates that never result in a session update after a session close. We can do that with locks

src/Sentry/GlobalSessionManager.cs
```diff
@@ -20,11 +23,17 @@ internal class GlobalSessionManager : ISessionManager
         // Internal for testing
         internal Session? CurrentSession => _currentSession;
 
-        public GlobalSessionManager(SentryOptions options)
+        public GlobalSessionManager(ISystemClock clock, SentryOptions options)
```
bruno-garcia, line 26:
```suggestion
        public GlobalSessionManager(SentryOptions options, ISystemClock clock)
```

Tyrrrz:
@bruno-garcia ready for final review/merge btw

bruno-garcia (approved):
NOICE

PR #1079, opened 2021-06-23
Add ReportAssembliesMode in favor of ReportAssemblies

### LINKED ISSUE #1072: InformationalVersion for referenced assemblies
We really like the assemblies part where we can see what assembly versions are being used. However, it does not support pre-release versions (e.g. 2.0.0-alpha0027). They all show up as 2.0.0.0.

For the main app and sentry, it respects the `InformationalVersion` attribute.

Could this part be changed to prefer that instead of the version first?

https://github.com/getsentry/sentry-dotnet/blob/main/src/Sentry/Internal/MainSentryEventProcessor.cs#L109

GeertvanHorrik:
Fixes #1072

Introduces the InformationalVersionAttribute in favor of ReportAssemblies, covered by unit tests.

A few additional changes:

1. Added additional null check for assemblies
2. Moved retrieving the informational version attribute to try/catch since it was failing on the unit tests for NET461

Technically the GetName() can be optimized inside the foreach loop (only needed when value =  in the MainSentryEventProcessor, but wasn't sure if it was worth less-readable code vs this optimization. Let me know if you would like this to be changed. For now this value is also used to determine whether an assembly should be checked / added at all.

Looking forward to your review. Have a great day!

### DIFF
--- a/src/Sentry/Internal/MainSentryEventProcessor.cs
+++ b/src/Sentry/Internal/MainSentryEventProcessor.cs
@@ -4,6 +4,7 @@ using System.Globalization;
 using System.Linq;
 using System.Threading;
 using Sentry.Extensibility;
+using Sentry.Reflection;
 
 namespace Sentry.Internal
 {
@@ -96,19 +97,36 @@ namespace Sentry.Internal
                 }
             }
 
-            if (_options.ReportAssemblies)
+            if (_options.ReportAssembliesMode != ReportAssembliesMode.None)
             {
                 foreach (var assembly in AppDomain.CurrentDomain.GetAssemblies())
                 {
-                    if (assembly.IsDynamic)
+                    if (assembly is null ||
+                        assembly.IsDynamic)
                     {
                         continue;
                     }
 
                     var asmName = assembly.GetName();
-                    if (asmName.Name is not null && asmName.Version is not null)
+                    if (asmName.Name is not null)
                     {
-                        @event.Modules[asmName.Name] = asmName.Version.ToString();
+                        var asmVersion = string.Empty;
+
+                        switch (_options.ReportAssembliesMode)
+                        {
+                            case ReportAssembliesMode.Version:
+                                asmVersion = asmName.Version?.ToString() ?? string.Empty;
+                                break;
+
+                            case ReportAssembliesMode.InformationalVersion:
+                                asmVersion = assembly.GetNameAndVersion().Version ?? string.Empty;
+                                break;
+
+                            default:
+                                throw new ArgumentOutOfRangeException(nameof(_options.ReportAssembliesMode), $"Report assemblies mode '{_options.ReportAssembliesMode}' is not yet supported");
+                        }
+
+                        @event.Modules[asmName.Name] = asmVersion;
                     }
                 }
             }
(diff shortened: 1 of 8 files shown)


bruno-garcia (approved):
Perfect. Thanks!

bruno-garcia:
Please add an entry to `CHANGELOG.md` to the "Unreleased" section. Make sure the entry includes this PR's number.

Example:
```
## Unreleased

### Features

* Add ReportAssembliesMode in favor of ReportAssemblies([#1079](https://github.com/getsentry/sentry-dotnet/pull/1079))
```

GeertvanHorrik:
Done, thanks for the fast review!

bruno-garcia:
Seems like we have some failing tests @GeertvanHorrik 

image

GeertvanHorrik:
That's my bad, sorry. I though returning string.Empty is better than null to prevent potential null reference exceptions. Will fix.

GeertvanHorrik:
Turning the tests back into their original behavior, but I am not sure if they are correct. At the moment, the informational version returns string.Empty (""), and that is being respected as null version. But I think it should fallback to the real assembly version in such case?

Maybe that's something to think about for a separate ticket. I will revert the old behavior back and respect a string.empty return from the informational version.

src/Sentry/Internal/MainSentryEventProcessor.cs
```diff
@@ -96,19 +97,36 @@ public SentryEvent Process(SentryEvent @event)
...
+
+                            case ReportAssembliesMode.InformationalVersion:
+                                asmVersion = assembly.GetNameAndVersion().Version ?? string.Empty;
+                                break;
+
+                            default:
+                                throw new ArgumentOutOfRangeException(nameof(_options.ReportAssembliesMode), $"Report assemblies mode '{_options.ReportAssembliesMode}' is not yet supported");
+                        }
+
+                        @event.Modules[asmName.Name] = asmVersion;
```
bruno-garcia, line 129: To keep the behavior like before I think we need to make sure we're not adding it if we didn't resolve the version:

```suggestion
                        if (!string.IsNullOrWhiteSpace(asmVersion))
                        {
                            @event.Modules[asmName.Name] = asmVersion;
                        }
```

bruno-garcia (commented):
I spotted one thing we might need to change

src/Sentry/Reflection/AssemblyExtensions.cs
```diff
@@ -22,11 +22,28 @@ public static SdkVersion GetNameAndVersion(this Assembly asm)
...
+            }
+            catch
+            {
+            }
+
+            // Note: even though the informational version could be "invalid" (e.g. string.Empty), it should
+            // be used for versioning and the software should not fallback to the assembly version string.
+            //
+            // See https://github.com/getsentry/sentry-dotnet/pull/1079#issuecomment-866992216
+            if (asmVersion is null)
```
bruno-garcia, line 41: I believe we should fallback to `Version` though. If it's an empty string it could be a bug in the build process that patches the value for example. Having as an empty string is arguably worse than `""`

```suggestion
            if (string.IsNullOrWhiteSpace(asmVersion))
```

So if tests fails with this, I believe changing the tests is the right thing to do

bruno-garcia (commented):
Sorry for adding more comments but I think we should consider one more thing

bruno-garcia:
@GeertvanHorrik Thanks for your contributions. I'll change the minor items I suggested and release it

PR #1093, opened 2021-06-29
Change DSN resolving logic to throw if the DSN was neither provided directly through options nor through environment variable

Tyrrrz:
See my review comments. Also note that this is likely a behavioral breaking change.

### DIFF
--- a/src/Sentry/Internal/DsnLocator.cs
+++ b/src/Sentry/Internal/DsnLocator.cs
@@ -19,5 +19,8 @@ namespace Sentry.Internal
         /// <returns>DSN string or null if none found.</returns>
         internal static string? FindDsn(Assembly? asm = null)
             => (asm ?? Assembly.GetEntryAssembly())?.GetCustomAttribute<DsnAttribute>()?.Dsn;
+
+        internal static string? Resolve() =>
+            Environment.GetEnvironmentVariable(Constants.DsnEnvironmentVariable) ?? FindDsn();
     }
 }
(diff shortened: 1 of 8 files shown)


src/Sentry/Internal/DsnLocator.cs
```diff
@@ -19,5 +19,8 @@ internal static string FindDsnStringOrDisable(Assembly? asm = null)
         /// <returns>DSN string or null if none found.</returns>
         internal static string? FindDsn(Assembly? asm = null)
             => (asm ?? Assembly.GetEntryAssembly())?.GetCustomAttribute<DsnAttribute>()?.Dsn;
+
+        internal static string? Resolve() =>
+            Environment.GetEnvironmentVariable(Constants.DsnEnvironmentVariable) ?? FindDsn();
```
Tyrrrz, line 24: Note: `FindDsnStringOrDisable` is now completely unused. I'm not sure what its purpose is at this point (or was, for that matter, if the current behavior was expected before).

test/Sentry.Tests/SentrySdkTests.cs
```diff
@@ -139,7 +139,7 @@ public void Init_InvalidDsnEnvironmentVariable_Throws()
                 });
         }
 
-        [Fact]
+        [Fact(Skip = "Setting an environment variable to an empty string is the same as not having that variable set at all, so this is not testable")]
```
Tyrrrz, line 142: As mentioned in Discord, setting an environment variable to `""` has the same effect as not setting the environment variable at all (or removing it if it was set previously).

image

Works as expected with a non-empty string:

image

Tyrrrz:
Closing because it's a massive breaking change and we decided not to do it

PR #1094, opened 2021-06-30
fix: incorrect order when getting the last active span.

robersonliou:
Here is an active span order issue.

when I started multi active span in the same transaction, the span tree in sentry ui was not what I expected.
For example, it should be:

```text
--- http.server
      |
      --- controller span
            |
            --- service span
                  |
                  --- repository span
```

but what I got was:
```text
--- http.server
      |
      --- controller span
            |
            --- service span
            |
            --- repository span
```            
Here is the actual screenshot:
image

I've traced the source code, and found this issue was cause by `TransactionTrace.GetLastActiveSpan()`.
The `TransactionTrace` stored the ongoing span in the `ConcurrentBag<ISpan>`, it worked fine!
But the default order of `ConcurrentBag` is just like FILO, if I put several active spans sequentially, the result would like as following table:

 SpanName | IsFinished | StartTime 
--------------|--------|------
repository span | false | 2021/6/30 00:00:02 |
service span | false | 2021/6/30 00:00:01 | 
controller span | false | 2021/6/30 00:00:00 |

It could be solved by reordering the span.
```csharp
# before
public ISpan? GetLastActiveSpan() => Spans.LastOrDefault(s => !s.IsFinished);

# after
public ISpan? GetLastActiveSpan() => Spans.OrderBy(s => s.StartTimestamp).LastOrDefault(s => !s.IsFinished);
```

### DIFF
--- a/src/Sentry/TransactionTracer.cs
+++ b/src/Sentry/TransactionTracer.cs
@@ -253,7 +253,9 @@ namespace Sentry
             Finish(exception, SpanStatusConverter.FromException(exception));
 
         /// <inheritdoc />
-        public ISpan? GetLastActiveSpan() => Spans.OrderByDescending(x => x.StartTimestamp).FirstOrDefault(s => !s.IsFinished);
+        public ISpan? GetLastActiveSpan() => 
+            // We need to sort by timestamp because the order of ConcurrentBag<T> is not deterministic
+            Spans.OrderByDescending(x => x.StartTimestamp).FirstOrDefault(s => !s.IsFinished);
 
         /// <inheritdoc />
         public SentryTraceHeader GetTraceHeader() => new(
--- a/src/Sentry/TransactionTracer.cs
+++ b/src/Sentry/TransactionTracer.cs
@@ -253,7 +253,7 @@ namespace Sentry
             Finish(exception, SpanStatusConverter.FromException(exception));
 
         /// <inheritdoc />
-        public ISpan? GetLastActiveSpan() => Spans.LastOrDefault(s => !s.IsFinished);
+        public ISpan? GetLastActiveSpan() => Spans.OrderBy(x => x.StartTimestamp).LastOrDefault(s => !s.IsFinished);
 
         /// <inheritdoc />
         public SentryTraceHeader GetTraceHeader() => new(
--- a/src/Sentry/TransactionTracer.cs
+++ b/src/Sentry/TransactionTracer.cs
@@ -253,7 +253,7 @@ namespace Sentry
             Finish(exception, SpanStatusConverter.FromException(exception));
 
         /// <inheritdoc />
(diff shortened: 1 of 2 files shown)


Tyrrrz:
Thank you!

Can you please add a changelog entry inside Changelog.md/Unreleased/Fixes section?

```
- Fixed incorrect order when getting the last active span ([#1094](https://github.com/getsentry/sentry-dotnet/pull/1094))
```

src/Sentry/TransactionTracer.cs
```diff
@@ -253,7 +253,7 @@ public void Finish(Exception exception, SpanStatus status)
             Finish(exception, SpanStatusConverter.FromException(exception));
 
         /// <inheritdoc />
-        public ISpan? GetLastActiveSpan() => Spans.LastOrDefault(s => !s.IsFinished);
+        public ISpan? GetLastActiveSpan() => Spans.OrderBy(x => x.StartTimestamp).LastOrDefault(s => !s.IsFinished);
```
Tyrrrz, line 256:
```suggestion
        public ISpan? GetLastActiveSpan() =>
            // We need to sort by timestamp because the order of ConcurrentBag<T> is not deterministic
            Spans.OrderBy(x => x.StartTimestamp).LastOrDefault(s => !s.IsFinished);
```

robersonliou: Sure, I've updated the CHANGELOG , thanks!

Tyrrrz (approved):

src/Sentry/TransactionTracer.cs
bruno-garcia: Might turn out to be the same at runtime but arguably this is easier to reason about:

```suggestion
        public ISpan? GetLastActiveSpan() => Spans.OrderByDescending(x => x.StartTimestamp).FirstOrDefault(s => !s.IsFinished);
```

src/Sentry/TransactionTracer.cs
```diff
@@ -253,7 +253,7 @@ public void Finish(Exception exception, SpanStatus status)
             Finish(exception, SpanStatusConverter.FromException(exception));
 
         /// <inheritdoc />
-        public ISpan? GetLastActiveSpan() => Spans.LastOrDefault(s => !s.IsFinished);
+        public ISpan? GetLastActiveSpan() => Spans.OrderByDescending(x => x.StartTimestamp).FirstOrDefault(s => !s.IsFinished);
```
bruno-garcia, line 256:
```suggestion
        public ISpan? GetLastActiveSpan() => 
            // We need to sort by timestamp because the order of ConcurrentBag<T> is not deterministic
            Spans.OrderByDescending(x => x.StartTimestamp).FirstOrDefault(s => !s.IsFinished);
```

bruno-garcia:
Thanks @robersonliou !

robersonliou:
@bruno-garcia It seems nothing update in `v3.6.1` assembly.
image

PR #1095, opened 2021-06-30
Process isolation for cache directory

### LINKED ISSUE #1067: Offline caching should support concurrent process instances
Currently the `CacheDirectoryPath` property enabled offline caching of events, and the persistency of the "InstalationId" that uniquely identifies a device without any privacy concerns (it's a random ID so can't be related to any other information to identify a user across apps).

This path is used as a base, and a sub directory `Sentry` is created followed by a hash of the DSN (to make sure different apps using Sentry don't share any data).

One use case that is not covered is when multiple instances of the same app (so using the same DSN) are executed at the same time. Different SDK instances will race through accessing the caching directory.

The behavior we'd expect is:

* Installation Id is shared across all instances of the same app. But not of different apps (so must live inside the DSN hash folder)
* All Sentry related files are under a folder called `Sentry`
* Session state is not shared across different instances of the same app. In other words, if you have 2 notepad instances open, each will have its own session and hence its own cache file of it to survive app restarts

Tyrrrz:
Closes #1067

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -3,11 +3,14 @@
   <PropertyGroup>
     <Version>3.6.0</Version>
     <LangVersion>latest</LangVersion>
-    <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
     <AssemblyOriginatorKeyFile>../../.assets/Sentry.snk</AssemblyOriginatorKeyFile>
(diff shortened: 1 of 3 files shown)


Tyrrrz:
How the fuck did this pass the tests

PR #1101, opened 2021-07-02
Avoid Unhandled Exception on .NET 461 if the Registry Access threw an exception.

### LINKED ISSUE #969: Opt-out for .NET Framework installations detection
[From a Discord thread](https://discord.com/channels/621778831602221064/621783515423440927/839406941272014878): On **Dynamic CRM** the SDK can fail with:

```
Unhandled Exception: System.ServiceModel.FaultException`1[[Microsoft.Xrm.Sdk.OrganizationServiceFault, Microsoft.Xrm.Sdk, Version=9.0.0.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35]]: Échec de la demande d'autorisation de type 'System.Security.Permissions.SecurityPermission, mscorlib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089'.
Detail: <OrganizationServiceFault xmlns="http://schemas.microsoft.com/xrm/2011/Contracts" xmlns:i="http://www.w3.org/2001/XMLSchema-instance">
  <ActivityId>00000000-0000-0000-0000-000000000000</ActivityId>
  <ErrorCode>-2147220970</ErrorCode>
  <ErrorDetails xmlns:a="http://schemas.datacontract.org/2004/07/System.Collections.Generic">
    <KeyValuePairOfstringanyType>
      <a:key>CallStack</a:key>
      <a:value i:type="b:string" xmlns:b="http://www.w3.org/2001/XMLSchema">   à System.Security.CodeAccessSecurityEngine.Check(Object demand, StackCrawlMark&amp; stackMark, Boolean isPermSet)
   à System.Security.CodeAccessSecurityEngine.Check(CodeAccessPermission cap, StackCrawlMark&amp; stackMark)
   à System.Security.CodeAccessPermission.Demand()
   à Microsoft.Win32.RegistryKey.CheckUnmanagedCodePermission()
   à Microsoft.Win32.RegistryKey.OpenBaseKey(RegistryHive hKey, RegistryView view)
   à Sentry.PlatformAbstractions.FrameworkInfo.Get45PlusLatestInstallationFromRegistry()
   à Sentry.PlatformAbstractions.FrameworkInfo.GetLatest(Int32 clrVersion)
   à Sentry.PlatformAbstractions.RuntimeInfo.SetNetFxReleaseAndVersion(Runtime runtime)
   à Sentry.PlatformAbstractions.RuntimeInfo.GetRuntime()
   à Sentry.PlatformAbstractions.Runtime.get_Current()
   à Sentry.SentryOptions..ctor()
   à Sentry.SentrySdk.Init(Action`1 configureOptions)
   à Client.Xrm.Plugins.AppCode.Plugin.Execute(IServiceProvider serviceProvider)
   à PluginProfiler.Library.PluginAppDomainProxy.ExecuteCore(Stopwatch watch, ProfilerExecutionReport report, Object instance, Object executionParameter)
   à PluginProfiler.Library.AppDomainProxy.Execute(ProfilerExecutionConfiguration configuration, ProfilerExecutionReport report)</a:value>
    </KeyValuePairOfstringanyType>
  </ErrorDetails>
  <HelpLink i:nil="true" />
  <Message>Échec de la demande d'autorisation de type 'System.Security.Permissions.SecurityPermission, mscorlib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089'.</Message>
  <Timestamp>2021-05-05T07:35:46.8046979Z</Timestamp>
  <ExceptionRetriable>false</ExceptionRetriable>
  <ExceptionSource i:nil="true" />
  <InnerFault i:nil="true" />
  <OriginalException i:nil="true" />
  <TraceText i:nil="true" />
</OrganizationServiceFault>
```

This is through the integration added here:
https://github.com/getsentry/sentry-dotnet/blob/d25c97f0e5c77b520e30877444e93a9147afb9cd/src/Sentry/SentryOptions.cs#L517-L519

We need to add an opt-out to `SentryOptions`:
`bool IncludeNetFxInstallations { get; set; }` and check for that in here:

https://github.com/getsentry/sentry-dotnet/blob/d25c97f0e5c77b520e30877444e93a9147afb9cd/src/Sentry/Integrations/NetFxInstallationsIntegration.cs#L8-L14

Additionally to the opt-out, we could detect this is running on Dynamic CRM and automatically opt-out of this feature.
And/or inspect the CAS policy we're running under and avoid the call altogether (log out with diagnostic logger)

lucas-zimerman:
SentryOptions .Ctor calls Runtime.Current.Name and that parameter was unsafe since it calls SetNetFxReleaseAndVersion on NET461.
Where, inside of it, is called RegistryKey.OpenBaseKey, which may throw an Exception by executing it.

The goal of this PR is to block any access to the functions that rely on RegistryKey in case of failure, avoiding unhandled exceptions to the user.

My only question is what to do in case of an exception inside of GetLatest? it would be nice to log the error but we don't have access to SentryOptions from it.

Close #969

#skip-changelog


src/Sentry/SentryOptions.cs
```diff
@@ -527,6 +521,12 @@ public SentryOptions()
                 () => ExceptionProcessors ?? Enumerable.Empty<ISentryEventExceptionProcessor>()
             };
 
+            // from 3.0.0 uses Enhanced (Ben.Demystifier) by default which is a breaking change
+            // unless you are using .NET Native which isn't compatible with Ben.Demystifier.
+            StackTraceMode = Runtime.Current.Name == ".NET Native"
+                ? StackTraceMode.Original
+                : StackTraceMode.Enhanced;
+
```
Tyrrrz, line 529: Why did we need to move it?

src/Sentry/Integrations/NetFxInstallationsIntegration.cs
```diff
@@ -7,7 +7,7 @@ internal class NetFxInstallationsIntegration : ISdkIntegration
     {
         public void Register(IHub hub, SentryOptions options)
         {
-            if (!Runtime.Current.IsMono())
+            if (!Runtime.Current.IsMono() && !FrameworkInfo.RegistryLocked)
```
Tyrrrz, line 10: Do we need `!FrameworkInfo.RegistryLocked`? Can we not just have a try/catch inside `FrameworkInfo.GetLatest()` and be done with it?

src/Sentry/PlatformAbstractions/FrameworkInfo.NetFx.cs
```diff
@@ -81,10 +102,15 @@ public static partial class FrameworkInfo
...
+        /// <param name="registryLock">Blocks this function to access the System's registry.</param>
         /// <seealso href="https://docs.microsoft.com/en-us/dotnet/framework/migration-guide/how-to-determine-which-versions-are-installed#to-find-net-framework-versions-by-querying-the-registry-in-code-net-framework-1-4"/>
         /// <returns>Enumeration of installations</returns>
-        public static IEnumerable<FrameworkInstallation> GetInstallations()
+        public static IEnumerable<FrameworkInstallation> GetInstallations(bool registryLock = false)
         {
+            if (registryLock || RegistryLocked)
+            {
+                yield break;
+            }
```
Tyrrrz, line 113: What's the point of passing a parameter that just makes the function exit immediately? Wouldn't it be better to just _not call_ the function instead?

src/Sentry/PlatformAbstractions/FrameworkInfo.NetFx.cs
```diff
@@ -160,13 +186,17 @@ private static FrameworkInstallation GetFromV4(RegistryKey subKey, string subKey
...
         }
 
         // https://docs.microsoft.com/en-us/dotnet/framework/migration-guide/how-to-determine-which-versions-are-installed#to-find-net-framework-versions-by-querying-the-registry-in-code-net-framework-45-and-later
-        internal static int? Get45PlusLatestInstallationFromRegistry()
+        internal static int? Get45PlusLatestInstallationFromRegistry(bool registryLock)
         {
+            if (registryLock)
+            {
+                return null;
+            }
```
Tyrrrz, line 199: What's the point of passing a parameter that just makes the function exit immediately? Wouldn't it be better to _just not_ call the function instead?

src/Sentry/PlatformAbstractions/FrameworkInfo.NetFx.cs
```diff
@@ -160,13 +186,17 @@ private static FrameworkInstallation GetFromV4(RegistryKey subKey, string subKey
...
             using var ndpKey = RegistryKey.OpenBaseKey(RegistryHive.LocalMachine, RegistryView.Registry32)
                 .OpenSubKey(@"SOFTWARE\Microsoft\NET Framework Setup\NDP\v4\Full\");
```
Tyrrrz, line 201: Note: you're not passing a second parameter to `OpenSubKey` here (`bool writable`). I believe that parameter is `true` by default which could cause access exception if the user doesn't have required permissions. Read-only access, on the other hand, should be available in all cases. Of course it still might fail when running on non-Windows operating system.

https://docs.microsoft.com/en-us/dotnet/api/microsoft.win32.registrykey.opensubkey?view=net-5.0#microsoft-win32-registrykey-opensubkey(system-string-system-boolean)

src/Sentry/PlatformAbstractions/FrameworkInfo.NetFx.cs
```diff
@@ -32,7 +39,21 @@ public static partial class FrameworkInfo
 
             if (clrVersion == 4)
             {
-                var release = Get45PlusLatestInstallationFromRegistry();
+                int? release = null;
```
Tyrrrz, line 42: Please avoid mutation. Try to extract the try/catch block into a separate method and use `return` statements directly there.

src/Sentry/PlatformAbstractions/FrameworkInfo.NetFx.cs
```diff
@@ -32,7 +39,21 @@ public static partial class FrameworkInfo
...
+                {
+                    if (registryLock)
+                    {
+                        RegistryLocked = true;
+                    }
+                    release = Get45PlusLatestInstallationFromRegistry(RegistryLocked);
+                }
+                catch (Exception ex)
+                {
+                    _ = ex;
```
Tyrrrz, line 53: What are we achieving with this?

src/Sentry/PlatformAbstractions/FrameworkInfo.NetFx.cs
```diff
@@ -20,8 +26,9 @@ public static partial class FrameworkInfo
         /// CLR 4 => .NET 4.0, 4.5.x, 4.6.x, 4.7.x
         /// </remarks>
         /// <param name="clrVersion">The CLR version: 1, 2 or 4</param>
-        /// <returns>The framework installation or null if none is found.</returns>
-        public static FrameworkInstallation? GetLatest(int clrVersion)
+        /// <param name="registryLock">Blocks this function to access the System's registry.</param>
+        /// <returns>The framework installation or null if none is found, used for fine-tuning the Latest info.</returns>
+        public static FrameworkInstallation? GetLatest(int clrVersion, bool registryLock = false)
```
Tyrrrz, line 31: Again, can we just have the function return `null` in case registry is inaccessible? Why do we need a shared property and a method parameter for that?

Tyrrrz (changes requested):

src/Sentry/SentryOptions.cs
lucas-zimerman: To leave it next to 
```csharp
            SentryStackTraceFactory = new SentryStackTraceFactory(this);

            ISentryStackTraceFactory SentryStackTraceFactoryAccessor() => SentryStackTraceFactory;
```
but I'll undo it since the change doesn`t fix the problem.

src/Sentry/Integrations/NetFxInstallationsIntegration.cs
lucas-zimerman: I added this check to avoid adding an EventProcessor that wasn't going to work due to lack of Registry access. But I could remove it.

src/Sentry/PlatformAbstractions/FrameworkInfo.NetFx.cs
lucas-zimerman: Nothing 😅, I'll remove the _ = ex; 
Yet, I wonder if we should Log that the Registry access failed or not.

bruno-garcia: if you're not using the exception remove the whole `(Exception ex`) from the catch statement

src/Sentry/Integrations/NetFxInstallationsIntegration.cs
Tyrrrz: I think it's fine to keep the event processor, just make it not do anything if it can't access registry.

@bruno-garcia thoughts on this?

src/Sentry/PlatformAbstractions/FrameworkInfo.NetFx.cs
Tyrrrz: After looking at it again, it seems the second parameter is `false` by default, so it probably wasn't the issue.

src/Sentry/PlatformAbstractions/FrameworkInfo.NetFx.cs
Tyrrrz: > Yet, I wonder if we should Log that the Registry access failed or not.

I think it would be nice. We can inject an instance of `SentryOptions` inside the event processor through constructor and then inject it here, I think?

src/Sentry/PlatformAbstractions/FrameworkInfo.NetFx.cs
lucas-zimerman: The doc says if you only pass the name it'll try to open as read-only
https://docs.microsoft.com/en-us/dotnet/api/microsoft.win32.registrykey.opensubkey?view=net-5.0#Microsoft_Win32_RegistryKey_OpenSubKey_System_String_

Should I specify the second parameter as false just to be sure?

Tyrrrz: > Should I specify the second parameter as false just to be sure?

I think it's a good idea, yeah

src/Sentry/Integrations/NetFxInstallationsIntegration.cs
lucas-zimerman: So the flow that I made was to not throw/register any Exception while creating the Runtime in NET461, but to log it once NetFxInstallationsEventProcessor is Processed.

src/Sentry/PlatformAbstractions/FrameworkInfo.NetFx.cs
```diff
@@ -99,20 +78,35 @@ public static partial class FrameworkInfo
             return latest;
         }
 
+        internal static RegistryKey? TryOpenSubKey(string name, out Exception? exception)
```
Tyrrrz, line 83: Let's turn this into an extension method `TryOpenSubKey(this RegistryKey key, string name)` instead (with explicit base key).

We can then do:

```csharp
var throws = RegistryKey.OpenBaseKey(RegistryHive.LocalMachine, RegistryView.Registry32).OpenSubKey(name);

var noThrow = RegistryKey.OpenBaseKey(RegistryHive.LocalMachine, RegistryView.Registry32).TryOpenSubKey(name);
```

Then just use the throw variant instead of the current `TryOpenSubKey(name out _)`.

Also, is the above equivalent to this?

```csharp
var subKey = Registry.LocalMachine.TryOpenSubKey(name);
```

src/Sentry/SentryOptions.cs
Tyrrrz: Just curious if it serves a specific purpose. It wasn't obvious from the diffs 😉

src/Sentry/PlatformAbstractions/FrameworkInfo.NetFx.cs
Tyrrrz: Or, optionally, just wrap `GetInstallations()` and `Get45PlusLatestInstallationFromRegistry()` into full body try/catch to swallow an exception.

For example:

```csharp
internal static int? Get45PlusLatestInstallationFromRegistry()
        {
            try
            {
                using var ndpKey = Registry.LocalMachine.OpenSubKey(NetFxNdpFullRegistryKey, false);
                return ndpKey?.GetInt("Release");
             }
             catch
             {
                 return null;
             }
        }
```

src/Sentry/PlatformAbstractions/FrameworkInfo.NetFx.cs
```diff
@@ -80,31 +80,22 @@ public static partial class FrameworkInfo
...
         /// <seealso href="https://docs.microsoft.com/en-us/dotnet/framework/migration-guide/how-to-determine-which-versions-are-installed#to-find-net-framework-versions-by-querying-the-registry-in-code-net-framework-1-4"/>
         /// <returns>Enumeration of installations</returns>
         public static IEnumerable<FrameworkInstallation> GetInstallations()
         {
-            using var ndpKey = TryOpenSubKey(NetFxNdpRegistryKey, out _);
-            return GetInstallationsFromRegistryKey(ndpKey);
+            try
+            {
+                using var ndpKey = Registry.LocalMachine.OpenSubKey(NetFxNdpRegistryKey, false);
+                return GetInstallationsFromRegistryKey(ndpKey);
```
Tyrrrz, line 93: `ndpKey` is disposed by the time the enumerator actually enumerates. The issue is that this statements returns immediately, but the `foreach` inside `GetInstallationsFromRegistryKey()` kicks in after `ndpKey` is disposed.

You might need to do this:

```suggestion
                using var ndpKey = Registry.LocalMachine.OpenSubKey(NetFxNdpRegistryKey, false);
                foreach (var installation in GetInstallationsFromRegistryKey(ndpKey))
                {
                    yield return installation;
                }
```

Tyrrrz: Or, alternatively:

```suggestion
                using var ndpKey = Registry.LocalMachine.OpenSubKey(NetFxNdpRegistryKey, false);
                return GetInstallationsFromRegistryKey(ndpKey).ToArray();
```

lucas-zimerman: It's weird that it worked on my Machine(TM), but at least, the test pointed out it was broken 😅

Tyrrrz (approved):
Looks good :+1:

src/Sentry/PlatformAbstractions/FrameworkInfo.NetFx.cs
```diff
@@ -85,8 +87,22 @@ public static partial class FrameworkInfo
...
+            {
+                using var ndpKey = Registry.LocalMachine.OpenSubKey(NetFxNdpRegistryKey, false);
+                foreach (var installation in GetInstallationsFromRegistryKey(ndpKey))
+                {
+                    yield return installation;
+                }
+            }
+            catch
+            {
+                return null;
```
Tyrrrz, line 100: Build error:

```
/home/runner/work/sentry-dotnet/sentry-dotnet/src/Sentry/PlatformAbstractions/FrameworkInfo.NetFx.cs(100,17): error CS1622: Cannot return a value from an iterator. Use the yield return statement to return a value, or yield break to end the iteration. [/home/runner/work/sentry-dotnet/sentry-dotnet/src/Sentry/Sentry.csproj]
```

Probably best to just return an empty enumerator?

```suggestion
                yield break;
```

lucas-zimerman: Actually, you can't use yield inside of a try-catch block

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 ### Fixes
 
+- Avoid Unhandled Exception on .NET 461 if the Registry Access threw an exception([#1101](https://github.com/getsentry/sentry-dotnet/pull/1101))
```
bruno-garcia, line 7: Need to be added to `Unreleased`

src/Sentry/PlatformAbstractions/FrameworkInfo.NetFx.cs
bruno-garcia: Please resolve conversations once it's sorted out

src/Sentry/PlatformAbstractions/FrameworkInfo.NetFx.cs
```diff
@@ -85,8 +88,19 @@ public static partial class FrameworkInfo
...
         public static IEnumerable<FrameworkInstallation> GetInstallations()
         {
-            using var ndpKey = RegistryKey.OpenRemoteBaseKey(RegistryHive.LocalMachine, string.Empty)
-                .OpenSubKey(@"SOFTWARE\Microsoft\NET Framework Setup\NDP\");
+            try
+            {
+                using var ndpKey = Registry.LocalMachine.OpenSubKey(NetFxNdpRegistryKey, false);
+                return GetInstallationsFromRegistryKey(ndpKey).ToArray();
+            }
+            catch
```
bruno-garcia, line 96: Personally not a fan of just catch and hiding things. Can we either pass a logger and log the issue or let it bubble and handle this at the caller?

src/Sentry/PlatformAbstractions/FrameworkInfo.NetFx.cs
```diff
@@ -160,16 +174,24 @@ private static FrameworkInstallation GetFromV4(RegistryKey subKey, string subKey
...
-            return ndpKey?.GetInt("Release");
+            {
+                try
+                {
+                    using var ndpKey = Registry.LocalMachine.OpenSubKey(NetFxNdpFullRegistryKey, false);
+                    return ndpKey?.GetInt("Release");
+                }
+                catch
+                {
+                    return null;
```
bruno-garcia, line 192: Same as above (handler at caller or log here)

src/Sentry/PlatformAbstractions/NetFxInstallationsEventProcessor.cs
```diff
@@ -23,7 +24,8 @@ internal class NetFxInstallationsEventProcessor : ISentryEventProcessor
         internal static Dictionary<string, string> GetInstallationsDictionary()
         {
             var versionsDictionary = new Dictionary<string, string>();
-            var installations = FrameworkInfo.GetInstallations().ToArray();
+            using var registryKey = Registry.LocalMachine.OpenSubKey(FrameworkInfo.NetFxNdpRegistryKey, false);
+            var installations = FrameworkInfo.GetInstallationsFromRegistryKey(registryKey).ToArray();
```
bruno-garcia, line 28: This means we'lll be throwing and catch an exception on each call?
Ideally we'd cache the value of this here once and have the error handling/logging done here instead (remove `catch return null` from the code you've added)

Tyrrrz: Can probably just wrap the whole thing in a `Lazy<T>`

lucas-zimerman: Changed the flow so it'll only throw only once and log the exception when an Event is processed by NetFxInstallationsEventProcessor

src/Sentry/Internal/Hub.cs
```diff
@@ -57,7 +57,14 @@ internal Hub(ISentryClient client, ISystemClock clock, ISessionManager sessionMa
...
                 {
                     options.DiagnosticLogger?.LogDebug("Registering integration: '{0}'.", integration.GetType().Name);
-                    integration.Register(this, options);
+                    try
+                    {
+                        integration.Register(this, options);
+                    }
+                    catch (Exception ex)
+                    {
+                        options.DiagnosticLogger?.LogError("Failed to Register integration: '{0}'.", ex, integration.GetType().Name);
```
Tyrrrz, line 66:
```suggestion
                        options.DiagnosticLogger?.LogError("Failed to register integration: '{0}'.", ex, integration.GetType().Name);
```

lucas-zimerman:
Ok, error flow has changed:
* It'll no longer crash inside of SentryOptions since StackTraceMode isn't changed there.
* Because of that, the Runtime on NET461 will be created when NetFxInstallationsIntegration is registered.
** Because of that, the Registration will throw an exception and an additional try/catch had to be placed inside of hub to avoid _rootScope not being pushed here
https://github.com/getsentry/sentry-dotnet/blob/6cce6bd533cc9c15b6d4ac8a30fedfc0b2113c76/src/Sentry/Internal/Hub.cs#L64-L67
Resulting in a crash if Disposed (or another side effect untested)

* StackTraceMode inside of SentryOptions is safer, avoiding Exceptions when Accessing Runtime.Current and logging the errors if a DiagnosticLogger is available. 

* Lastly, the Current Runtime got split into two functions,  one that returns the runtime and another that sets SetNetFxReleaseAndVersion or SetNetFxReleaseAndVersion. that was made to avoid runtime being null whenever an exception happened inside of the second part and entering on a bad logic that would throw an exception any time the runtime was called.

src/Sentry/PlatformAbstractions/Runtime.cs
```diff
@@ -12,7 +12,18 @@ public class Runtime
         /// <value>
         /// The current runtime.
         /// </value>
-        public static Runtime Current => _runtime ??= RuntimeInfo.GetRuntime();
+        public static Runtime Current
+        {
+            get
+            {
+                if( _runtime is null)
```
Tyrrrz, line 19:
```suggestion
                if (_runtime is null)
```

src/Sentry/SentryOptions.cs
```diff
@@ -461,14 +461,40 @@ public double TracesSampleRate
...
-        public StackTraceMode StackTraceMode { get; set; }
+        public StackTraceMode StackTraceMode
+        {
+            get
+            {
+                if(_stackTraceMode  is { } stackTraceMode)
+                {
+                    return stackTraceMode;
+                }
+                try
```
Tyrrrz, line 481: Newline

```suggestion
                
                try
```

src/Sentry/SentryOptions.cs
```diff
@@ -461,14 +461,40 @@ public double TracesSampleRate
...
+                    // from 3.0.0 uses Enhanced (Ben.Demystifier) by default which is a breaking change
+                    // unless you are using .NET Native which isn't compatible with Ben.Demystifier.
+                    _stackTraceMode = Runtime.Current.Name == ".NET Native"
+                        ? StackTraceMode.Original
+                        : StackTraceMode.Enhanced;                    
+                }
+                catch (Exception ex)
+                {
+                    _stackTraceMode = StackTraceMode.Enhanced;
+                    DiagnosticLogger?.LogError("Failed to get runtime, setting StackTraceMode to {0} ", ex, StackTraceMode.Enhanced);
```
Tyrrrz, line 492:
```suggestion
                    DiagnosticLogger?.LogError($"Failed to get runtime, setting {nameof(StackTraceMode)} to {0} ", ex, _stackTraceMode);
```

src/Sentry/SentryOptions.cs
```diff
@@ -461,14 +461,40 @@ public double TracesSampleRate
...
+                }
+                return _stackTraceMode.Value;
```
Tyrrrz, line 494:
```suggestion

                return _stackTraceMode.Value;
```

Tyrrrz (approved):
Looks good! Some visual nits

samples/Sentry.Samples.Console.Basic/Program.cs
```diff
@@ -1,6 +1,6 @@
 using Sentry;
 
-using (SentrySdk.Init("https://eb18e953812b41c3aeb042e666fd3b5c@o447951.ingest.sentry.io/5428537"))
+using (SentrySdk.Init(o=> { o.Dsn = "https://eb18e953812b41c3aeb042e666fd3b5c@o447951.ingest.sentry.io/5428537"; o.Debug = true; o.DiagnosticLevel = SentryLevel.Debug; }))
```
bruno-garcia: It's debug by default:

Is it better to break lines here instead? Something like:

```suggestion
using (SentrySdk.Init(o=> 
    {
        o.Dsn = "https://eb18e953812b41c3aeb042e666fd3b5c@o447951.ingest.sentry.io/5428537"
        o.Debug = true;
     }))
```

src/Sentry/Internal/Hub.cs
```diff
@@ -57,7 +57,14 @@ internal Hub(ISentryClient client, ISystemClock clock, ISessionManager sessionMa
                 foreach (var integration in _integrations)
                 {
                     options.DiagnosticLogger?.LogDebug("Registering integration: '{0}'.", integration.GetType().Name);
-                    integration.Register(this, options);
+                    try
+                    {
+                        integration.Register(this, options);
```
bruno-garcia, line 62: Do we want to hide integrations that fail installing? Before the user at least would be aware and raise an issue

src/Sentry/PlatformAbstractions/FrameworkInfo.NetFx.cs
```diff
@@ -87,8 +86,10 @@ public static partial class FrameworkInfo
         /// <seealso href="https://docs.microsoft.com/en-us/dotnet/framework/migration-guide/how-to-determine-which-versions-are-installed#to-find-net-framework-versions-by-querying-the-registry-in-code-net-framework-1-4"/>
         /// <returns>Enumeration of installations</returns>
         public static IEnumerable<FrameworkInstallation> GetInstallations()
-        { 
-            using var ndpKey = Registry.LocalMachine.TryOpenLocalSubKey(NetFxNdpRegistryKey);
+        {
+            int i = 0; if (i == 0)
```
bruno-garcia: ?

src/Sentry/SentryOptions.cs
```diff
@@ -461,14 +461,40 @@ public double TracesSampleRate
...
         /// <remarks>
         /// Sentry groups events by stack traces. If you change this mode and you have thousands of groups,
         /// you'll get thousands of new groups. So use this setting with care.
         /// </remarks>
-        public StackTraceMode StackTraceMode { get; set; }
+        public StackTraceMode StackTraceMode
+        {
+            get
+            {
+                if(_stackTraceMode  is { } stackTraceMode)
```
bruno-garcia, line 477:
```suggestion
                if (_stackTraceMode  is { } stackTraceMode)
```

bruno-garcia (changes requested):

src/Sentry/PlatformAbstractions/RuntimeInfo.cs
```diff
@@ -21,15 +21,20 @@ internal static Runtime GetRuntime()
...
+        {
 #if NET461
             SetNetFxReleaseAndVersion(runtime);
 #else
             SetNetCoreVersion(runtime);
 #endif
-            return runtime;
         }
 
+
```
lucas-zimerman, line 36:
```suggestion
```

src/Sentry/SentryOptions.cs
```diff
@@ -461,14 +461,42 @@ public double TracesSampleRate
...
+                    // from 3.0.0 uses Enhanced (Ben.Demystifier) by default which is a breaking change
+                    // unless you are using .NET Native which isn't compatible with Ben.Demystifier.
+                    _stackTraceMode = Runtime.Current.Name == ".NET Native"
+                        ? StackTraceMode.Original
+                        : StackTraceMode.Enhanced;
+                }
+                catch (Exception ex)
+                {
+                    _stackTraceMode = StackTraceMode.Enhanced;
+                    DiagnosticLogger?.LogError($"Failed to get runtime, setting {nameof(StackTraceMode)} to {0} ", ex, _stackTraceMode);
```
bruno-garcia, line 493: Avoid the interpolation, the generic overloads are here to avoid allocations (unless you measured and confirmed the compiler does that already):

```suggestion
                    DiagnosticLogger?.LogError("Failed to get runtime, setting {0} to {1} ", ex, nameof(StackTraceMode), _stackTraceMode);
```

bruno-garcia (approved):
I have a couple of nits but otherwise LGTM

PR #1103, opened 2021-07-02
logging and additional fallback for installation ID

Tyrrrz:
I don't know what other logs I can add, this sequence is incredibly trivial. I did add `Environment.MachineName` as fallback, which should hopefully work x-platform.


src/Sentry/GlobalSessionManager.cs
```diff
@@ -145,7 +151,8 @@ public GlobalSessionManager(SentryOptions options)
 
                 var id =
                     TryGetPersistentInstallationId() ??
-                    TryGetHardwareInstallationId();
+                    TryGetHardwareInstallationId() ??
+                    Environment.MachineName;
```
bruno-garcia, line 155: Could you please you `SHA256` this or something? So we don't rely on the raw machine name?

Tyrrrz: Good point, missed that

Tyrrrz: Done

bruno-garcia (approved):

PR #1105, opened 2021-07-02
File system persistence for sessions

### LINKED ISSUE #1096: Cache current session information on disk.
Cache current session information on disk. Close open sessions on restart as Abnormal.

The use case is when a crash happens that doesn't go through the UnhandledException integration (i.e: Native crash). Or the device ran out of battery for example.

The next step after this will be to differentiate abnormal from crash (with the help of native SDKs).

If the session is paused, we should mark it as such and the timestamp. If it dies while paused, on restart we can use that timestamp to end the sessions (as Exited). It's possible that there was a native crash on a background thread while the app was on the background but that we would only be able to answer with support of a native SDK.

To support differentiating Abnormal and Crashed (or even Exited when the session was paused) we need native SDK integration.
From the .NET SDK point of view this is just a callback:

```
SentryOptions
Func<booL>? CrashedLastRun
```
[Similar to what we have on `sentry-cocoa`](https://github.com/getsentry/sentry-cocoa/search?q=crashedlastrun).

Use cases:

Without native SDK integration (callback is null):

1. Session starts, a file is written for the session state. App is put on the background. App dies (out of battery).
On restart Sentry will find the file and end that session as Exited.

2. Sentry SDK starts, an open session is found. Close that session as Abnormal.

With native SDK integration (callback is set to something):

1. SDK Init with an open session on disk. Ask native SDK (Through a callback): Did we crash on the last run? of so we close as Crashed otherwise as Abnormal.

Example Unity Android:

```
SentrySdk.Init(o => {
#if ANDROID
o.CrashedLastRun = () => SentryAndroid.didCrashLastRun();
#elif IOS
o.CrashedLastRun = () => SentryCocoa.didCrashLastRun();
#endif
});
```

In other words:

1. When the .NET SDK starts, it checks for open sessions.
2. If `CrashedLastRun` is null it fallback to the best of its knowlege: Abnormal if session was not paused, otherwise Exited.
3. If callback exists, rely on its value to differentiate between a native crash and Abnormal (Out battery? User killed the app?) and Exited (if was put on the background).

Tyrrrz:
Closes #1096


src/Sentry/SentryOptionsExtensions.cs
```diff
@@ -236,5 +238,25 @@ internal static void SetupLogging(this SentryOptions options)
...
+        {
+            if (string.IsNullOrWhiteSpace(options.CacheDirectoryPath))
+            {
+                return null;
+            }
+
+            return Path.Combine(
+                options.CacheDirectoryPath,
+                "Sentry",
+                options.Dsn?.GetHashString() ?? "no-dsn"
```
bruno-garcia, line 252: Should we throw here and end the world?

src/Sentry/Internal/Hub.cs
```diff
@@ -177,6 +178,27 @@ public void BindException(Exception exception, ISpan span)
...
+                    }
+                }
+                catch (Exception ex)
+                {
+                    _options.DiagnosticLogger?.LogError(
+                        "Failed to recover persisted session.",
+                        ex
+                    );
+                }
+            }
```
Tyrrrz, line 199: I put the recovery in `StartSession()` to avoid adding a new method to `IHub`. An alternative would be to put it in the constructor though, but I think it's better to avoid I/O in there. Thoughts?

src/Sentry/SentryOptionsExtensions.cs
Tyrrrz: You mean disable caching when DSN is not provided?

Tyrrrz:
Still missing support for special handling of paused sessions (will have to store additional data in the file) and `DidCrash` handler. But what do you think about the general direction @bruno-garcia?

src/Sentry/GlobalSessionManager.cs
```diff
@@ -161,7 +164,81 @@ public GlobalSessionManager(SentryOptions options)
...
+        private void DeletePersistedSession()
+        {
+            if (string.IsNullOrWhiteSpace(_persistanceDirectoryPath))
+            {
+                return;
+            }
+
+            try
+            {
+                File.Delete(
```
bruno-garcia, line 205:
```suggestion
                if (_logger.IsEnabled(Level.Debug))
                {
                    _logger.LogDebug("Deleting: {0}, File.ReadAllText(_..);
                }
                File.Delete(
```

src/Sentry/GlobalSessionManager.cs
```diff
@@ -161,7 +164,81 @@ public GlobalSessionManager(SentryOptions options)
...
+                    ex
+                );
+            }
+        }
+
+        public SessionUpdate? TryRecoverPersistedSession()
+        {
+            if (string.IsNullOrWhiteSpace(_persistanceDirectoryPath))
+            {
+                return null;
```
bruno-garcia, line 222:
```suggestion
                _logger.Debug(..);
                return null;
```

src/Sentry/GlobalSessionManager.cs
```diff
@@ -161,7 +164,81 @@ public GlobalSessionManager(SentryOptions options)
...
+            }
+
+            try
+            {
+                var data = File.ReadAllBytes(Path.Combine(_persistanceDirectoryPath, PersistedSessionFileName));
+                var update = SessionUpdate.FromJson(Json.Parse(data));
+
+                // Switch status to abnormal and initial to false
+                // TODO: crashed for paused sessions
```
bruno-garcia, line 231:
```suggestion
                // TODO: Exited for paused sessions
                // TODO: Crashed based on the result of a callback
```

src/Sentry/IJsonSerializable.cs
```diff
@@ -16,4 +17,18 @@ public interface IJsonSerializable
         /// </remarks>
         void WriteTo(Utf8JsonWriter writer);
     }
+
+    internal static class JsonSerializableExtensions
+    {
+        public static byte[] WriteToMemory(this IJsonSerializable serializable)
+        {
+            using var buffer = new MemoryStream();
```
bruno-garcia, line 25: Prefer FileStream through instead of buffering it all here

src/Sentry/SentryOptionsExtensions.cs
```diff
@@ -236,5 +238,25 @@ internal static void SetupLogging(this SentryOptions options)
                 options.DiagnosticLogger = null;
             }
         }
+
+        internal static string? TryGetDsnSpecificCacheDirectoryPath(this SentryOptions options)
+        {
+            if (string.IsNullOrWhiteSpace(options.CacheDirectoryPath))
+            {
+                return null;
```
bruno-garcia, line 246: Perhaps at the call site but please log out the path it resolved to or if there was no caching at all

Tyrrrz:
@bruno-garcia some significant changes.
I had to change how the pause/resume works so that it's more encapsulated within session manager, which would allow it to take ownership of the file persistence logic. As a result, it now returns multiple updates from `ResumeSession(...)` back to `Hub`. Ugly, but I don't think there are better options since `GlobalSessionManager` does not encapsulate a reference to `ISentryClient` (although that wouldn't be a horrible idea maybe?).
Evaluating race conditions and thread contention is a nightmare now, too many variables. Part of me just wants to wrap everything in locks.
What do you think?

src/Sentry/Internal/Json.cs
```diff
@@ -15,5 +16,12 @@ public static JsonElement Parse(string json)
             using var jsonDocument = JsonDocument.Parse(json);
             return jsonDocument.RootElement.Clone();
         }
+
+        public static JsonElement Load(string filePath)
+        {
+            using var file = File.OpenRead(filePath);
+            using var jsonDocument = JsonDocument.Parse(file);
+            return jsonDocument.RootElement.Clone();
```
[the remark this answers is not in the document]
Tyrrrz: `jsonDocument.RootElement` is a JSON element backed by a streaming/lazy `JsonDocument`. What `Clone()` does is buffer it all in memory so that the original stream can be safely disposed.

src/Sentry/GlobalSessionManager.cs
```diff
@@ -174,7 +183,141 @@ public GlobalSessionManager(SentryOptions options)
...
+
+            if (string.IsNullOrWhiteSpace(_persistenceDirectoryPath))
+            {
+                _options.DiagnosticLogger?.LogDebug("Persistence directory is not set, returning.");
+                return;
+            }
+
+            try
+            {
+                var filePath = Path.Combine(_persistenceDirectoryPath, PersistedSessionFileName);
```
bruno-garcia, line 242:
```suggestion
            var filePath = Path.Combine(_persistenceDirectoryPath, PersistedSessionFileName);
            try
            {
```

src/Sentry/GlobalSessionManager.cs
```diff
@@ -174,7 +183,141 @@ public GlobalSessionManager(SentryOptions options)
...
+                _options.DiagnosticLogger?.LogInfo(
+                    "Deleted persisted session file '{0}'.",
+                    filePath
+                );
+            }
+            catch (Exception ex)
+            {
+                _options.DiagnosticLogger?.LogError(
+                    "Failed to delete persisted session from the file system.",
+                    ex
```
bruno-garcia, line 275: Given the suggestion above, what about logging the path so we can peek into the filesystem to figure out why:

```suggestion
                _options.DiagnosticLogger?.LogError(
                    "Failed to delete persisted session from the file system: '{0}'",
                    ex,
                    filePath
```

src/Sentry/GlobalSessionManager.cs
```diff
@@ -174,7 +183,141 @@ public GlobalSessionManager(SentryOptions options)
...
+
+            if (string.IsNullOrWhiteSpace(_persistenceDirectoryPath))
+            {
+                _options.DiagnosticLogger?.LogDebug("Persistence directory is not set, returning.");
+                return null;
+            }
+
+            try
+            {
+                var filePath = Path.Combine(_persistenceDirectoryPath, PersistedSessionFileName);
```
bruno-garcia, line 292: Same as above we could life the path and use in the exception logging

bruno-garcia (approved):
Some small suggestions but LGTM

PR #1113, opened 2021-07-09
Add HTTP request breadcrumb

### LINKED ISSUE #1109: HTTP Request Breadcrumbs
> With #1178 we add auto performance instrumentation for HTTP requests. Having this in place it should be rather "easy" to add breadcrumbs for HTTP requests as this is an [expected](https://develop.sentry.dev/sdk/features/#http-client-integrations) feature for SDKs.

See: https://github.com/getsentry/sentry-cocoa/issues/1197#issuecomment-875631210

lucas-zimerman:
Add HTTP breadcrumbs once a request is finished

Close #1109

https://sentry.io/organizations/sentry-sdks/issues/2280281206/events/7a0ffe1d069d4f859c0b373849f44b28/?project=5428537


Tyrrrz (approved):

src/Sentry/SentryHttpMessageHandler.cs
```diff
@@ -63,16 +64,27 @@ public SentryHttpMessageHandler()
...
+                if (request.RequestUri?.AbsoluteUri is { } uri)
+                {
+                    var breadcrumbData = new Dictionary<string, string>
+                    {
+                        { "url", uri },
+                        { "method", requestMethod }
+                    };
+                    _hub.AddBreadcrumb(string.Empty, "http", "http", breadcrumbData);
+                }
+
```
bruno-garcia, line 87: `Uri.AbsoluteUri` Throws if it's not a AbsoluteUri:

https://github.com/dotnet/runtime/blob/65b472a0f31f2b878601b41a344754ec3e776d1d/src/libraries/System.Private.Uri/src/System/Uri.cs#L707

Also, `request.RequestUri` is already being red above so why `?.` here? The IDE should be warning this is redundant.

Finally, we're using `request.RequestUri` above when adding the span but `request.RequestUri.AbsoluteUri` here, why the inconsistency?

Should we be using only `request.RequestUri` here too?

bruno-garcia (changes requested):
There are some issues and no tests for this code. I left a comment

src/Sentry/SentryHttpMessageHandler.cs
```diff
@@ -61,18 +62,28 @@ public SentryHttpMessageHandler()
...
             InnerHandler ??= new HttpClientHandler();
 
+            var requestMethod = request.Method.Method.ToUpperInvariant();
+            var url = request.RequestUri?.ToString() ?? string.Empty;
+
             // Start a span that tracks this request
             // (may be null if transaction is not set on the scope)
             var span = _hub.GetSpan()?.StartChild(
-                "http.client",
+            "http.client",
```
bruno-garcia, line 71:
```suggestion
                "http.client",
```

src/Sentry/SentryHttpMessageHandler.cs
```diff
@@ -61,18 +62,28 @@ public SentryHttpMessageHandler()
...
             try
             {
                 var response = await base.SendAsync(request, cancellationToken).ConfigureAwait(false);
 
+                var breadcrumbData = new Dictionary<string, string>
+                {
+                    { "url", url },
+                    { "method", requestMethod }
+                };
+                _hub.AddBreadcrumb(string.Empty, "http", "http", breadcrumbData);
```
bruno-garcia, line 85: Instead of passing `Empty` which will result in a `""` being sent to Sentry, why not go "manual" and do `new Breadcrumb`?
```suggestion
                _hub.AddBreadcrumb(string.Empty, "http", "http", breadcrumbData);
```

src/Sentry/SentryHttpMessageHandler.cs
```diff
@@ -61,18 +62,28 @@ public SentryHttpMessageHandler()
...
             );
 
             try
             {
                 var response = await base.SendAsync(request, cancellationToken).ConfigureAwait(false);
 
+                var breadcrumbData = new Dictionary<string, string>
+                {
+                    { "url", url },
+                    { "method", requestMethod }
```
bruno-garcia, line 83:
```suggestion
                    { "method", requestMethod }
                    { "status", response.StatusCode }
```

src/Sentry/SentryHttpMessageHandler.cs
```diff
@@ -63,16 +64,27 @@ public SentryHttpMessageHandler()
 
             // Start a span that tracks this request
             // (may be null if transaction is not set on the scope)
+            var requestMethod = request.Method.Method.ToUpperInvariant();
+            var url = request.RequestUri?.ToString() ?? string.Empty;
+
```
lucas-zimerman, line 69:
```suggestion
            var requestMethod = request.Method.Method.ToUpperInvariant();
            var url = request.RequestUri?.ToString() ?? string.Empty;

            // Start a span that tracks this request
            // (may be null if transaction is not set on the scope)
```

bruno-garcia (approved):

PR #1123, opened 2021-07-14
TryRecoverPersistedSession logs debug if file doesnt exist

Tyrrrz:
Log a debug entry instead of an error because it's an expected error (`TryRecoverPersistedSession()` is called every time a session is started for the first time, but a left over file will actually exist very rarely).


bruno-garcia (commented):
Thank you

Issue #221, opened 2019-06-03
Custom tags missing in some contexts


mariusz-schimke-iteo:
I'm trying to use the Sentry package in a Topshelf Windows service with Serilog handling logging. The Topshelf package lets you install and start/stop your command line app as a service by simply running the executable with arguments, e.g.

`MyConsoleApp.exe install start` or `MyConsoleApp.exe stop`.

When you run it without arguments, it starts as a command line app.

When I start the app as a service (with the `start` command line argument), events generated by it have no custom tags and user ID in Sentry (please see scope configuration in the code below). When I start the app without any args, the custom tags are available in Sentry.

What I investigated so far, when starting the app with the `start` param, the started instance starts the same executable as a Windows service and exits right away. Events logged by the initially started instance are OK in Sentry, but those logged by the service miss custom tags.

I don't know how to explain and fix it. Would you have any suggestions in this context? I would be really grateful, even though I will probably just quit Topshelf in the meantime because I have already wasted too much time without success.

Here's my .NET Framework console app code:

```
using Sentry;
using Sentry.Protocol;
using Serilog;
using System;
using Topshelf;

namespace ConsoleAppSentry
{
    internal class Program
    {
        private static void Main(string[] args)
        {
            var sentry = SentrySdk.Init(o =>
            {
                o.Dsn = new Dsn("my dsn");
            });

            SentrySdk.ConfigureScope(scope =>
            {
                scope.User = new User() { Id = "TestUID" };
                scope.SetTag("app.name", "Console");
            });

            Log.Logger = new LoggerConfiguration()
                .WriteTo.Sentry()
                .CreateLogger();

            Log.Logger.Error($"MAIN (CUSTOM TAGS EXIST)");

            var rc = HostFactory.Run(x =>
            {
                x.UseSerilog();

                x.Service<WindowsService>(s =>
                {
                    s.ConstructUsing(name => new WindowsService());
                    s.WhenStarted(tc => tc.Start());
                    s.WhenStopped(tc => tc.Stop());
                });

                x.RunAsLocalSystem();
                x.StartManually();

                x.SetServiceName("topshelf_test");
            });

            sentry.Dispose();

            Environment.ExitCode = (int)Convert.ChangeType(rc, rc.GetTypeCode());
        }
    }

    public class WindowsService
    {
        public void Start()
        {
            Log.Logger.Error($"START (CUSTOM TAGS MISSING)");
        }

        public void Stop()
        {
[...]

Issue #505, opened 2020-09-11
Figure out a solution to state management


bruno-garcia:
Currently the SDK has a single `Hub` hanging off of an AsyncLocal static field.

This doesn't work in some cases, like through round trip through native code calling into C#.

Also, in Desktop and Mobile apps, it's more common to expect state to be shared across the app.
This is how our Android and Cocoa SDKs work.
Currenly the Hub will fork and data set on a copied AsyncLocal won't propagate to the root one. That works well for Server apps, but not for desktop/mobile.

See: #221, #448 

Relates to #628 (or just a dupe)

bruno-garcia:
Dupe of: #628

bruno-garcia:
"Global Mode" makes it so that any call to `SentrySdk.SetTag` or `AddBreadcrumb` or anything else that you do through `SentrySdk.ConfigureScope` mutates a single, static `Scope` object that affects **all threads of the app**. This is useful for Desktop and Mobile apps when you have a single user session and you add context to Sentry anywhere in the app and want that to be included if it crashes in any other thread.

You can opt-in to this mode through:

`options.IsGlobalModeEnabled = true`

To clarify: This options doesn't make sense in any sort of web server where each individual request takes a separate scope (You want each `SetTag("url", ...)` to take a different request URL, headers etc). Nothing needs to be done for server apps, it continues to work as it did in the past.

We plan to ship `Sentry.Wpf` and `Sentry.WinForms` in the future that would already flip `IsGlobalModeEnabled = true` for you. Until then we'll add to the docs that you can do that if you prefer to have a single mutable state.

PR #1124, opened 2021-07-14
Introduce scope stack abstraction

### LINKED ISSUE #628: Make Scope data global on desktop and mobile
This came up a few times, more recently on SO:

https://stackoverflow.com/questions/65066020/c-xamarin-forms-and-sentrysdk-context-info-not-making-it-to-sentry-errors

Currently the Hub is stored in a AsyncLocal field so data forks on new asynchronous flows.
This causes problems for Desktop and Mobile apps that want to push work to a back ground thread but would like that data available in case of an error anywhere.

This is how the Android and iOS SDKs of Sentry work, with a static global Hub.

An attempt to simulate that happened in this PR #448 but we the general approach needs to be discussed.

Resolves: #221

More context: https://forum.sentry.io/t/accessing-and-ensuring-global-shared-scope-in-net/11709/2

### LINKED ISSUE #1106: Sentry loses its scope inside the ASP.NET pipeline because it uses AsyncLocal for storage
[The official Sentry SDK docs for ASP.NET](https://docs.sentry.io/platforms/dotnet/guides/aspnet/) instructs one to start and finish tracing transactions like this in `Global.asax.cs`:

```csharp
protected void Application_BeginRequest()
{
    Context.StartSentryTransaction();
}

protected void Application_EndRequest()
{
    Context.FinishSentryTransaction();
}
```

This approach is buggy, as it will sometimes cause the Sentry scope to be lost inside the (non-Core) ASP.NET pipeline. The underlying problem is that the Sentry SDK stores its scopes in an `AsyncLocal` field in `SentryScopeManager.cs`. This is *not* a safe way to store such state in (non-Core) ASP.NET. The only safe method is likely to use `HttpContext.Items`. For more information, see:

* https://stackoverflow.com/questions/43391498/asynclocal-value-is-null-after-being-set-from-within-application-beginrequest
* https://devblogs.microsoft.com/dotnet/net-framework-4-7-1-asp-net-and-configuration-features/#asp-net-execution-step-feature
* http://piers7.blogspot.com/2005/11/threadstatic-callcontext-and_02.html

In my use case, I only care about the transaction reference, so I applied the following workaround to `Global.asax.cs`, based on the above mentioned blog post my Microsoft and how they themselves work around the issue in [Application Insights]( https://github.com/microsoft/ApplicationInsights-dotnet/blob/7c6589e2c5799edae5ffc30991b54364b2756835/WEB/Src/Web/Web/ApplicationInsightsHttpModule.cs#L150):

```csharp
public override void Init()
{
    base.Init();
    
    OnExecuteRequestStep((httpContextBase, step) =>
    {
        if (httpContextBase.CurrentNotification == RequestNotification.ExecuteRequestHandler && !httpContextBase.IsPostNotification)
        {
            {
                if (scope.Transaction == null)
                {
                    scope.Transaction = (ITransaction) httpContextBase.Items["__SentryTransaction"];
                }
            });
        }

        step();
    });
}
```

Tyrrrz:
Closes #1106
Closes #628

Related to #221, #448, #505

Not entirely sure how to test this. Ideas @bruno-garcia?


test/Sentry.Tests/Internals/ScopeStack/AsyncLocalScopeStackContainerTests.cs
```diff
@@ -0,0 +1,161 @@
...
+                Substitute.For<Scope>(),
+                Substitute.For<ISentryClient>()
+            );
+
+            // Act & assert
+            await Task.Run(async () =>
+            {
+                container.Stack.Should().BeNull();
+
+                await Task.Delay(10).ConfigureAwait(false);
```
bruno-garcia, line 133: Do you need to wait for 10ms or just yield?
```suggestion
                await Task.Yield().ConfigureAwait(false);
```

bruno-garcia (commented):
I like where this is going!

bruno-garcia:
> Not entirely sure how to test this. Ideas @bruno-garcia?

I think the direction you're going for a test for each approach is a good one. Let me take another look once it's out of draft to propose some integration/unit tests

Tyrrrz:
@bruno-garcia what do you think about the general implementation by the way? I know you said there are a lot of potential edge-cases/hidden issues, so I'm somewhat skeptical about my solution given how easy it was to implement 😅 

It could benefit to put some more responsibility inside `IScopeStackContainer`, but I decided to go with the easiest approach for now.

test/Sentry.Tests/Internals/ScopeStack/AsyncLocalScopeStackContainerTests.cs
Tyrrrz: I don't know. I thought adding a delay would introduce some potential race conditions which could be nice to test. But I guess yield is ok too.

bruno-garcia (approved):
Thanks!

bruno-garcia:
We need to ship this and add some docs

PR #1130, opened 2021-07-16
Fix `ParentSampledId` being reset on `Transaction`

### LINKED ISSUE #1097: sentry-trace ParentSpanId being lost from the ASP.NET Core Transaction
### Environment

How do you use Sentry?
Sentry SaaS (sentry.io) or self-hosted/on-premise (which version?)
SaaS

Which SDK and version?
e.g: JavaScript 5.11.1, .NET 1.2.0
.ASP NET Core SDK 3.6.0
### Steps to Reproduce

1. What
2. you
3. did.

### Expected Result

Connected Transactions.

### Actual Result

image
It was verified here https://github.com/getsentry/sentry-dotnet/blob/0f4358d1f37f664d7e144d33f6db5db842180726/src/Sentry.AspNetCore/SentryTracingMiddleware.cs#L94 that the information from sentry-trace was being set correctly.
But verifying at the end of the transaction lifecycle  https://github.com/getsentry/sentry-dotnet/blob/1d2efd50fea764c3d86790a97a82c66737812a07/src/Sentry/Envelopes/Envelope.cs#L142
I noticed that the ParentSpanId was removed while the TraceId was kept.
image

This results in an unconnected Transaction as shown in the UI
image
But if manually inserted on CaptureTransacton, the UI would be able to find the connected transaction as shown below:
image

Tyrrrz:
Closes #1097

### DIFF
--- a/test/Sentry.AspNetCore.Tests/SentryTracingMiddlewareTests.cs
+++ b/test/Sentry.AspNetCore.Tests/SentryTracingMiddlewareTests.cs
@@ -1,8 +1,6 @@
 ﻿#if !NETCOREAPP2_1
 using System;
 using System.Collections.Generic;
-using System.Diagnostics;
-using System.Linq;
 using System.Net.Http;
 using System.Threading.Tasks;
 using FluentAssertions;
@@ -15,7 +13,6 @@ using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.DependencyInjection.Extensions;
 using NSubstitute;
 using Sentry.AspNetCore.Tests.Utils.Extensions;
-using Sentry.Testing;
 using Xunit;
 
 namespace Sentry.AspNetCore.Tests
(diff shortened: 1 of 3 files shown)


test/Sentry.AspNetCore.Tests/SentryTracingMiddlewareTests.cs
```diff
@@ -169,11 +163,12 @@ public async Task Transaction_is_started_automatically_from_incoming_trace_heade
             await client.SendAsync(request);
 
             // Assert
-            transaction.Should().NotBeNull();
-            transaction?.Name.Should().Be("GET /person/{id}");
-            transaction.TraceId.Should().Be(SentryId.Parse("75302ac48a024bde9a3b3734a82e36c8"));
-            transaction.ParentSpanId.Should().Be(SpanId.Parse("1000000000000000"));
-            transaction.IsSampled.Should().BeFalse();
+            sentryClient.Received(1).CaptureTransaction(Arg.Is<Transaction>(t =>
+                t.Name == "GET /person/{id}" &&
```
bruno-garcia, line 167: The issue was `trasnaction?.` and transaction was null so assert wasn't being called?

bruno-garcia (approved):

test/Sentry.AspNetCore.Tests/SentryTracingMiddlewareTests.cs
Tyrrrz: No, it was fine. The problem is that the transaction was created correctly, it was not sent correctly. So I widened the test scope to cover that (by testing against `ISentryClient`).

PR #1143, opened 2021-07-26
Set error status if http has exception and ok status.

lucas-zimerman:
InvokeAsync middleware doesn't receive the correct status code during an error, setting a transaction with an error as "OK", this PR sets the default status code based on the received Exception and if the status code is Ok.

Before
image

After
image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
+
+### Fixes
+
+- Set error status to transaction if http has exception and ok status ([#1143](https://github.com/getsentry/sentry-dotnet/pull/1143))
+
 ## 3.8.3
 
 ### Features
--- a/src/Sentry.AspNetCore/SentryTracingMiddleware.cs
+++ b/src/Sentry.AspNetCore/SentryTracingMiddleware.cs
@@ -157,6 +157,9 @@ namespace Sentry.AspNetCore
                     {
                         transaction.Finish(status);
                     }
+                    // Status code not yet changed to 500 but an exception does exist
+                    // so lets avoid passing the misleading 200 down and close only with
+                    // the exception instance that will be inferred as errored.
                     else if (status == SpanStatus.Ok)
                     {
                         transaction.Finish(exception);
--- a/src/Sentry.AspNetCore/SentryTracingMiddleware.cs
+++ b/src/Sentry.AspNetCore/SentryTracingMiddleware.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Collections.Generic;
 using System.Runtime.ExceptionServices;
 using System.Threading.Tasks;
@@ -156,6 +156,10 @@ namespace Sentry.AspNetCore
                     if (exception is null)
                     {
                         transaction.Finish(status);
+                    }
+                    else if (status == SpanStatus.Ok)
+                    {
+                        transaction.Finish(exception);
                     }
                     else
                     {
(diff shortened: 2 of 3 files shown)


Tyrrrz:
Does this happen because the status code is set after the middleware?

Tyrrrz (approved):

lucas-zimerman:
> Does this happen because the status code is set after the middleware?

yup

CHANGELOG.md
```diff
@@ -2,8 +2,9 @@
 
 ## Unreleased
 
-## Fixes
+### Fixes
 
+- Set error status to transaction if http has exception and ok status ([#1143](https://github.com/getsentry/sentry-dotnet/pull/1143))
```
bruno-garcia, line 7:
```suggestion
- End transaction with error if exception is thrown in pipeline ([#1143](https://github.com/getsentry/sentry-dotnet/pull/1143))
```

src/Sentry.AspNetCore/SentryTracingMiddleware.cs
```diff
@@ -157,6 +157,10 @@ public async Task InvokeAsync(HttpContext context)
                     {
                         transaction.Finish(status);
                     }
+                    else if (status == SpanStatus.Ok)
+                    {
+                        transaction.Finish(exception);
```
bruno-garcia, line 162:
```suggestion
                    // Status code not yet changed to 500 but an exception does exist
                    // so lets avoid passing the misleading 200 down and close only with
                    // the exception instance that will be inferred as errored
                    else if (status == SpanStatus.Ok && exception is not null)
                    {
                        transaction.Finish(exception);
```

bruno-garcia (changes requested):
I think we need to check if we have an instance because passing `null` as exception could behave as `unknown` status or as `ok` or that can change in the future.

There are no tests for this so sounds like we can change this and nothing validates the change.

src/Sentry.AspNetCore/SentryTracingMiddleware.cs
lucas-zimerman: Doesn't the previous lines guarantee that you don't have a null exception?
```csharp
                    if (exception is null)
                    {
                        transaction.Finish(status);
                    }
                    else if (status == SpanStatus.Ok)
                    {
                        transaction.Finish(exception);
                    }
                    else
                    {
                        transaction.Finish(exception, status);
                    }
```

lucas-zimerman: Slightly edited the test Transaction_binds_exception_thrown for checking if the Span was set as Internal Error.
Without the changes in this PR, the test was returning a span with the Status Ok.

Tyrrrz: Yeah it already checks for null

bruno-garcia: OK, please add the comments

test/Sentry.AspNetCore.Tests/SentryTracingMiddlewareTests.cs
```diff
@@ -1,4 +1,4 @@
-﻿#if !NETCOREAPP2_1
+#if !NETCOREAPP2_1
```
bruno-garcia, line 1: BOM?

bruno-garcia (approved):

test/Sentry.AspNetCore.Tests/SentryTracingMiddlewareTests.cs
lucas-zimerman: not sure what has changes, at least, vs code doesn't spot it https://github1s.com/getsentry/sentry-dotnet/pull/1143

PR #1144, opened 2021-07-26
Server side release health

Tyrrrz:
for reference: https://github.com/getsentry/sentry-java/pull/1572


Tyrrrz:
@bruno-garcia I really don't see how we can adapt the existing `ISessionManager` interface for server mode.

Issues:

- Most methods on `ISessionManager` return `SessionUpdate`, which are completely irrelevant in server mode.
- In server mode, we need a way to get a `SessionAggregate` instead. Should we just add a separate method in `ISessionManager` for that? That will look really ugly, IMO.
- `Hub` will have different routines for `SessionUpdate`/`SessionAggregate` anyway because it needs to pack them in envelopes in distinct ways and then pass them to transport.

I'm struggling to see how to reconcile these conceptual differences.

Tyrrrz:
Note: reworked `GlobalSessionManager` to an approach similar to Java's.
`ServerSessionManager` is still pending.
Because `SessionAggregate` is a completely distinct structure from `SessionUpdate`, a new method was added to `ISentryClient` which needs to be propagated across different classes.

src/Sentry/ServerSessionManager.cs
```diff
@@ -0,0 +1,151 @@
...
+            SentryOptions options,
+            ISentryClient client,
+            ISystemClock clock)
+        {
+            _options = options;
+            _client = client;
+            _clock = clock;
+
+            // TODO: timer should be synced to run *on the minute*
+            _timer = new Timer {Interval = 60 * 1000, Enabled = true, AutoReset = true};
```
lucas-zimerman, line 39: This constructor is only compatible with .NET 5

lucas-zimerman: oh, Timer is implemented at two different places: System.Threading and System.Timers

Tyrrrz: Actually more like 4 different places :joy:

lucas-zimerman:
PR will be continued on #1162.

PR #1153, opened 2021-08-05
Add a notification layer to scope mutation

### LINKED ISSUE #1114: Add a notification layer to scope mutation
The Unity SDK requires synchronizing scope data to underlying SDKs.
Similar to what was done on Android->Native

https://github.com/getsentry/sentry-java/blob/3739f6775ea13bb3bfe01d66df2765bc35636e7e/sentry/src/main/java/io/sentry/Scope.java#L228-L231

Similar to what was done in Android:  

if in C# one calls SetUser, we call the observer:
https://github.com/getsentry/sentry-java/blob/ceccb11ca67af528531a4cfa8bb74963dd8230ac/sentry/src/main/java/io/sentry/Scope.java#L225-L233
 

We’d implement an observer per native SDK like: CocoaObserver, and AndroidObserver that implements the IScopeObserver interface and fires an extern call to the native layer

 https://github.com/getsentry/sentry-java/blob/main/sentry-android-ndk/src/main/java/io/sentry/android/ndk/NdkScopeObserver.java

lucas-zimerman:
Implements a one-way scope synchronization with the main SDK's scope with an external scope if EnableScopeSync is set to true.
Close #1114.


src/Sentry/IScopeObserver.cs
```diff
@@ -0,0 +1,54 @@
...
+
+        /// <summary>
+        /// Arbitrary key-value for this event.
+        /// </summary>
+        IReadOnlyDictionary<string, string> Tags { get; }
+
+        /// <summary>
+        /// An arbitrary mapping of additional metadata to store with the event.
+        /// </summary>
+        IReadOnlyDictionary<string, object?> Extra { get; }
```
bruno-garcia, line 24: I believe this interface should have only the methods to handle scope change and not expose any state
```suggestion
```

src/Sentry/IScopeObserver.cs
```diff
@@ -0,0 +1,54 @@
...
+        /// </summary>
+        void UnsetTag(string key);
+
+        /// <summary>
+        /// Gets the user information.
+        /// </summary>
+        /// <value>
+        /// The user.
+        /// </value>
+        User User { get; set; }
```
bruno-garcia, line 52: We can't return anything from here. We're just getting notified of changes:

```suggestion
        User User { set; }
```

src/Sentry/Scope.cs
```diff
@@ -99,13 +99,31 @@ public Contexts Contexts
             set => _contexts = value;
         }
 
+        // Internal for testing.
+        internal Action UserChanged => () =>
+        {
+            if (Options.EnableScopeSync &&
+                Options.ScopeObserver is { } observer)
+            {
+                observer.User = User;
```
bruno-garcia, line 108: Can we change this action to take the user as a reference? Technically the User being read here could be different than the one set if they raced (global hub mode)

src/Sentry/User.cs
```diff
@@ -11,37 +12,79 @@ namespace Sentry
     /// <see href="https://develop.sentry.dev/sdk/event-payloads/user/"/>
     public sealed class User : IJsonSerializable
     {
+        internal Action? PropertyChanged { get; set; }
```
bruno-garcia, line 15: Maybe this I mean:
```suggestion
        internal Action<User>? PropertyChanged { get; set; }
```

So you pass `PropertyChanged(this)`

bruno-garcia (commented):
Nice!

src/Sentry/Scope.cs
```diff
@@ -99,13 +99,31 @@ public Contexts Contexts
...
-            get => _user ??= new User();
-            set => _user = value;
+            get => _user ??= new User() { PropertyChanged = UserChanged };
+            set
+            {
+                _user = value;
+                if (_user != null)
+                {
+                    _user.PropertyChanged = UserChanged;
+                }
```
bruno-garcia, line 124:
```suggestion
                _user?.PropertyChanged = UserChanged;
```

src/Sentry/Scope.cs
```diff
@@ -99,13 +99,31 @@ public Contexts Contexts
...
+        };
+
         private User? _user;
 
         /// <inheritdoc />
         public User User
         {
-            get => _user ??= new User();
-            set => _user = value;
+            get => _user ??= new User() { PropertyChanged = UserChanged };
```
bruno-garcia, line 117:
```suggestion
            get => _user ??= new User { PropertyChanged = UserChanged };
```

test/Sentry.Tests/ScopeTests.cs
```diff
@@ -223,61 +224,215 @@ public void GetSpan_ActiveSpans_ReturnsSpan()
...
+            // Assert
+            observer.Received(expectedCount).SetUser(Arg.Is<User>((user)=> user.Email == expectedEmail));
+            observer.Received(expectedCount).SetUser(Arg.Is<User>((user) => user.Id == expectedid));
+            observer.Received(expectedCount).SetUser(Arg.Is<User>((user) => user.Username == expectedusername));
+        }
+
+        [Theory]
+        [InlineData(true)]
+        [InlineData(false)]
+        public void UserChanged_Observernull_Ignored(bool observerEnable)
```
bruno-garcia, line 327:
```suggestion
        public void UserChanged_ObserverNull_Ignored(bool observerEnable)
```

bruno-garcia (approved):
🐿️ 
I mean
:shipit:

PR #1112, opened 2021-07-08
Feat: EF Core performance

lucas-zimerman:
image

This PR provides automatic metrics for EF Core.

It creates spans for measuring the time required to connect to the database, execute the query and optimize it. 
If everything went well, the spans will receive the status OK, otherwise an internal error.

samples
OK Request https://sentry.io/organizations/sentry-sdks/discover/sentry-dotnet:7a363753fbc6446ba983de5006bb5809
On error: https://sentry.io/organizations/sentry-sdks/discover/sentry-dotnet:8806f87262324b45aaf1a9a5fd851f29
Another case of error: https://sentry.io/organizations/sentry-sdks/discover/sentry-dotnet:6424aec6c1684f658144ee29f320f279

PR #1161, opened 2021-08-23
fix: tests reaching the web

bruno-garcia:


src/Sentry/Internal/Http/CachingTransport.cs
```diff
@@ -215,9 +206,9 @@ private async Task ProcessCacheAsync(CancellationToken cancellationToken = defau
         // from disk could raise an IOException related to Disk I/O.
         // For that reason, we're not retrying IOException, to avoid any disk related exception from retrying.
         private static bool IsRetryable(Exception exception) =>
-            exception is OperationCanceledException // Timed-out or Shutdown triggered
-            || exception is HttpRequestException // Myriad of HTTP related errors
-            || exception is SocketException; // Network related
+            exception is OperationCanceledException
```
bruno-garcia, line 209:
```suggestion
            exception is OperationCanceledException // Timed-out or Shutdown triggered
```

lucas-zimerman:
It breaks on Ububty
```
Error: Sentry.AspNetCore.Tests.SentryHttpMessageHandlerBuilderFilterTests.Generated_client_sends_Sentry_trace_header_automatically: System.Net.Http.HttpRequestException : Name or service not known (fake.tld:443)
  Failed Sentry.AspNetCore.Tests.SentryHttpMessageHandlerBuilderFilterTests.Generated_client_sends_Sentry_trace_header_automatically [431 ms]
```

CHANGELOG.md
```diff
@@ -2,6 +2,11 @@
 
 ## Unreleased
 
+### Features
+
+- EF Core performance monitoring ([#1112](https://github.com/getsentry/sentry-dotnet/pull/1112))
```
lucas-zimerman, line 7: - EF Core performance monitoring ([#1112](https://github.com/getsentry/sentry-dotnet/pull/1112))

is not part of this PR but I suppose it'll be sorted out after merging the main.

src/Sentry/Internal/Http/CachingTransport.cs
```diff
@@ -67,10 +67,11 @@ private async Task CachedTransportBackgroundTask()
                 {
                     foreach (var filePath in Directory.EnumerateFiles(_processingDirectoryPath))
                     {
-                        File.Move(
-                            filePath,
-                            Path.Combine(_isolatedCacheDirectoryPath, Path.GetFileName(filePath))
-                        );
+                        var destinationPath = Path.Combine(_isolatedCacheDirectoryPath, Path.GetFileName(filePath));
+                        _options.DiagnosticLogger?.LogDebug("Moving unprocessed file back to cache: {0} to {1}.",
+                            filePath, destinationPath);
```
lucas-zimerman, line 72: nit-pick but I believe the below one will be slightly more clear to the end-user.
```suggestion
                        _options.DiagnosticLogger?.LogDebug("Moving unprocessed file back from cache {0} to {1}.",
                            filePath, destinationPath);

```

src/Sentry/Internal/Http/HttpTransport.cs
```diff
@@ -220,11 +226,19 @@ public async Task SendEnvelopeAsync(Envelope envelope, CancellationToken cancell
                         responseString
                     );
                 }
+
+                // If debug level, dump the whole envelope to the logger
+                if (_options.DiagnosticLogger?.IsEnabled(SentryLevel.Debug) is true)
+                {
+                    _options.DiagnosticLogger?.LogDebug("Failed envelope '{0}' has payload:\n{1}",
```
lucas-zimerman, line 233: maybe indicate when the payload start and where it ends, just to not mix with other logs.

```suggestion
                    _options.DiagnosticLogger?.LogDebug("Failed envelope '{0}' has payload:\n Payload star\n{1}\n Payload end.",
```

src/Sentry/Internal/SerializableExtensions.cs
```diff
@@ -4,18 +4,18 @@
...
         public static async Task<string> SerializeToStringAsync(
             this ISerializable serializable,
             CancellationToken cancellationToken = default)
         {
-#if !NET461 && !NETCOREAPP2_1
+#if !NET461 && !NETSTANDARD2_0
             await
 #endif
-            using var stream = new MemoryStream();
+                using var stream = new MemoryStream();
```
lucas-zimerman, line 18:
```suggestion
#if !NET461 && !NETSTANDARD2_0
            await using var stream = new MemoryStream();
#else
            using var stream = new MemoryStream();
#endif
```

src/Sentry/Internal/SerializableExtensions.cs
```diff
@@ -4,18 +4,18 @@
 using System.Threading.Tasks;
 using Sentry.Protocol.Envelopes;
 
-namespace Sentry.Tests.Helpers
+namespace Sentry.Internal
 {
     internal static class SerializableExtensions
```
lucas-zimerman, line 9: I may be wrong, but moving a static code from the Test project to the Main project isn't going to include an unused static extension to the Main SDK instead of limiting it to the Test projects.

test/Sentry.AspNetCore.Tests/SentryHttpMessageHandlerBuilderFilterTests.cs
```diff
@@ -74,11 +74,10 @@ public async Task Generated_client_sends_Sentry_trace_header_automatically()
                                 .GetRequiredService<IHttpClientFactory>()
                                 .CreateClient();
 
-                            await httpClient.GetAsync("https://example.com");
+                            await httpClient.GetAsync("https://fake.tld");
```
lucas-zimerman, line 77: flaky since https://example.com exists but not https://fake.tld.

lucas-zimerman (approved):
Great changes!. 
Left some comments but overall looks ok (except the broken test that needs to be fixed before merging)

CHANGELOG.md
bruno-garcia:
```suggestion
```

src/Sentry/Internal/Http/CachingTransport.cs
bruno-garcia: You suggestion adds another call to the logger? 
Code here is moving for "processing" back to the "cache" folder. So I think the original message makes more sense

src/Sentry/Internal/Http/HttpTransport.cs
bruno-garcia: There's a typo on `start` -> `star` so I can't 'just accept' your suggestion.
Also there's a space after the last line break.
```suggestion
                    _options.DiagnosticLogger?.LogDebug("Failed envelope '{0}' has payload:\n{1}\n",
```

src/Sentry/Internal/SerializableExtensions.cs
bruno-garcia: Why is it beneficial to add a `#else` block with the same statement over only adding `await`?

If we agreed on this new style lets do a single pass and replace it everything. The approach I used is already the standard and there are many usages

test/Sentry.AspNetCore.Tests/SentryHttpMessageHandlerBuilderFilterTests.cs
bruno-garcia: I don't follow. Why is it flaky?

The goal is to make sure we're not relying on a real DNS name and Internet to resolve it.

I ran these tests offline and the build failed due to such tests.

PR #1179, opened 2021-08-30
Fix SQLClient unplanned behaviors

lucas-zimerman:
This PR will fix the following issues found during dogfooding:
* The first Query span is lost.
: It happens because oftentimes the event got invoked before the Connection event.
: FIX: create the Query on the Transaction if a connection is not found, afterwards, set the correct parent once the query is closed.

* Missing `Connection Span`
:  The span was opened correctly but in case of data insertion or exclusion, the event related to closing the connection wasn't invoked, instead, another event gets invoked `WriteTransactionCommitAfter`.
: FIX: Close the Connection span once `WriteTransactionCommitAfter` is invoked.

* Many `Connection Spans` with the same `connection Id`:
:  Not sure why, perhaps since the connection might be shared, it closes to continue the execution of another query from another request and return to open the previous one once finished.
: FIX: Set the connection Id to the `Connection Span` if another Span with the same `Connection Id` isn't found.

PS: Had to change the setter from the `SpanTrace.parentId` to internal in order to update the `ParentID`.


src/Sentry.DiagnosticSource/Internals/DiagnosticSource/SentrySqlListener.cs
```diff
@@ -76,6 +84,15 @@ private void AddSpan(SentrySqlSpanType type, string operation, string? descripti
                     TryGetQuerySpan(scope, queryId) is { } querySpan)
                 {
                     span = querySpan;
+
+                    if (span.ParentSpanId == scope.Transaction?.SpanId &&
+                        span.Extra[ConnectionExtraKey] is Guid spanConnectionId &&
```
bruno-garcia, line 89: Lets assume the value could have been removed from the dictionary and be safe when reading:
```suggestion
                        span.Extra.TryGetValue("ConnectionExtraKey", out var spanConnectionId) && spanConnectionId is Guid && spanConnectionId &&
```

src/Sentry.DiagnosticSource/Internals/DiagnosticSource/SentrySqlListener.cs
```diff
@@ -148,6 +171,11 @@ public void OnNext(KeyValuePair<string, object?> value)
                     TrySetConnectionStatistics(connectionSpan, value);
                     connectionSpan.Finish(SpanStatus.Ok);
                 }
+                else if ((value.Key is SqlMicrosoftWriteTransactionCommitAfter || value.Key is SqlDataWriteTransactionCommitAfter) &&
+                    GetSpan(SentrySqlSpanType.Connection, null, value.GetSubProperty<Guid>("Connection", "ClientConnectionId")) is { } connectionSpan2)
```
bruno-garcia, line 175: Can this throw if props don't exist? `value.GetSubProperty<Guid>("Connection", "ClientConnectionId")`

lucas-zimerman: This is a safe call, if not found, it'll return the default value of T, in this case, an empty GUID.

lucas-zimerman:
I had to initialize by default the Extra parameter from SpanTracer due to the chance of data loss when inserting data when the parameter wasn't created.
That was the main issue with the test since it relies on the extra parameters to correctly link the spans.

src/Sentry.DiagnosticSource/Internals/DiagnosticSource/SentrySqlListener.cs
```diff
@@ -91,18 +139,27 @@ private void AddSpan(SentrySqlSpanType type, string operation, string? descripti
...
         {
             _hub.ConfigureScope(scope =>
             {
-                var span = scope.Transaction?.Spans.FirstOrDefault(span => TryGetKey(span.Extra, OperationExtraKey) is Guid id && id == operationId);
-                span?.SetExtra(ConnectionExtraKey, connectionId);
+                // We may have multiple Spans with different Operations for the same connection.
+                // So lets set the connection Id only if there are no connection spans with the same connectionId.
+                var connectionSpans = scope.Transaction?.Spans?.Where(span => span.Operation is "db.connection");
+                if (connectionSpans?.Any(span => TryGetConnectionId(span) == connectionId) is false &&
+                    connectionSpans.FirstOrDefault(span => TryGetOperationId(span) == operationId) is { } span)
```
bruno-garcia, line 159: Why iterate over connectionSpans twice inside this if?
this can have 1k items in it, looks highly inefficient with so many LINQ calls per execution

test/Sentry.Tests/SpanTracerTests.cs
```diff
@@ -0,0 +1,38 @@
...
+                var hub = Substitute.For<IHub>();
+                var transaction = new SpanTracer(hub, null, null, SentryId.Empty, "");
+
+                // Act
+                var tasks = Enumerable.Range(1, 4).Select((_) => Task.Run(() =>
+                {
+                    for (int i = 0; i < amount; i++)
+                    {
+
+                        transaction.SetExtra(Guid.NewGuid().ToString(), Guid.NewGuid());
```
bruno-garcia, line 27:
```suggestion
                        transaction.SetExtra(Guid.NewGuid().ToString(), Guid.NewGuid());
```

test/Sentry.Tests/SpanTracerTests.cs
```diff
@@ -0,0 +1,38 @@
...
+        {
+            // Run 20 times to avoid flacky tests scapping.
+            for (int amount = 0; amount < 20; amount++)
+            {
+                // Arrange
+                var hub = Substitute.For<IHub>();
+                var transaction = new SpanTracer(hub, null, null, SentryId.Empty, "");
+
+                // Act
+                var tasks = Enumerable.Range(1, 4).Select((_) => Task.Run(() =>
```
bruno-garcia, line 22: There's no guarantee of parallelism here because there's no synchronization.
It's possible/likely that the tasks will activate in series and their callback run also in series.

You need to synchronize (with the snippet I shared earlier) to block them all before the code you're attempting to test is ready to be executed by each task to get closer to parallelism here.

Have them all WaitOne before the for loop and unblock them all at once from the test thread once they are all ready for it.

src/Sentry.DiagnosticSource/Internals/DiagnosticSource/SentrySqlListener.cs
lucas-zimerman: The first iteration checks if there're any connection Spans with the given connection ID, if none is found it grabs the connection span with the given Operation ID and sets the connection ID.

I don't think you can perform both checks with only one iteration

bruno-garcia (approved):

PR #1185, opened 2021-09-08
remove some redundant async state machines

SimonCropp:
#skip-changelog

### DIFF
--- a/src/Sentry.AspNetCore/SentryMiddleware.cs
+++ b/src/Sentry.AspNetCore/SentryMiddleware.cs
@@ -88,10 +88,10 @@ namespace Sentry.AspNetCore
 
                 if (_options.FlushOnCompletedRequest)
                 {
-                    context.Response.OnCompleted(async () =>
+                    context.Response.OnCompleted(() =>
                     {
                         // Serverless environments flush the queue at the end of each request
-                        await hub.FlushAsync(timeout: _options.FlushTimeout).ConfigureAwait(false);
+                        return hub.FlushAsync(timeout: _options.FlushTimeout);
                     });
                 }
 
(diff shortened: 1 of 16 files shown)


src/Sentry/Internal/BackgroundWorker.cs
```diff
@@ -46,7 +46,7 @@ internal class BackgroundWorker : IBackgroundWorker, IDisposable
             _shutdownSource = shutdownSource ?? new CancellationTokenSource();
             _queuedEnvelopeSemaphore = new SemaphoreSlim(0, _maxItems);
 
-            WorkerTask = Task.Run(async () => await WorkerAsync().ConfigureAwait(false));
+            WorkerTask = Task.Run(() => WorkerAsync());
```
josh-degraw, line 49:
```suggestion
            WorkerTask = Task.Run(WorkerAsync);
```
Might as well simplify this further

lucas-zimerman:
Could you add `#skip-changelog` to the PR description?

src/Sentry.AspNetCore/SentryMiddleware.cs
```diff
@@ -88,10 +88,10 @@ public async Task InvokeAsync(HttpContext context)
 
                 if (_options.FlushOnCompletedRequest)
                 {
-                    context.Response.OnCompleted(async () =>
+                    context.Response.OnCompleted(() =>
```
bruno-garcia, line 91: Could be then:

```
// Serverless environments flush the queue at the end of each request
context.Response.OnCompleted(() => hub.FlushAsync(timeout: _options.FlushTimeout));
```

bruno-garcia (commented):
One downside of returning tasks instead of awaiting is the resulting stack trace in case of an error. It's easier to debug when we await on tasks since we can tell exactly where it failed.

bruno-garcia:
I would use the `return Task` approach only on tests and sample code but not sure it's ideal on code we'll need to debug through from customers reported stack traces

SimonCropp:
> not sure it's ideal on code we'll need to debug through from customers reported stack traces

I heave hear this fear from a few people. TBH i have never had any issue with returning a task in the 100+ projects i manage. And on a side not, all the MS libraries directly return a task where possible.

But i understand the reticence, so i will close this one

bruno-garcia:
https://twitter.com/haacked/status/1441454691087310859?s=20

PR #1188, opened 2021-09-09
Fix: Use fallback if RoutePattern is MVC.

### LINKED ISSUE #1189: Fix performance transaction name when using MVC routes
When using `MapControllerRoute`, the SDK is currently just using the route template regardless of the endpoint. I.e:

image

We need to change the logic:

https://github.com/getsentry/sentry-dotnet/blob/a12bb9f2e0748c4560b207815631bfd5be651ac5/src/Sentry.AspNetCore/Extensions/HttpContextExtensions.cs#L19-L39

To account for such template.
For example:

```csharp
if (pattern.Contains("{controller=")) { 
  Regex.Replace("\{controller=(.*)\}", controllerName)
}
if (pattern.Contains("{action=")) { 
  Regex.Replace("\{action=(.*)\}", controllerName)
}
if (pattern.Contains("{area=")) { 
  Regex.Replace("\{area=(.*)\}", controllerName)
}
```

Such as:
> area/{controller=MyController}/bla/{action=MyAction}/{id?}/blabla

Becomes: 
> area/Controller/bla/Action/{id?}/blabla

lucas-zimerman:
Despite a context feature having a RouteEndpoint, it doesn't mean we can use the RawText value of the RoutePattern since it may result in a default route for different paths.
For a simple fix, I opted to just use the fallback to get the MVC route since we didn't have any proper data of the Route Template on the given Endpoint.

Additionally, that the SDK is creating transactions for items like loading a CSS file, thus results in Unknown Routes. If desired, we could use context.Request.Path.Value as a last fallback for cases like that.

Sample:
image

Also, the issue can be easily reproduced on the sample Sentry.Samples.AspNetCore.Mvc

Close #1189.

### DIFF
--- a/src/Sentry.AspNetCore/Extensions/HttpContextExtensions.cs
+++ b/src/Sentry.AspNetCore/Extensions/HttpContextExtensions.cs
@@ -16,7 +16,7 @@ namespace Sentry.AspNetCore.Extensions
             var endpoint = context.Features.Get<IEndpointFeature?>()?.Endpoint as RouteEndpoint;
             var routePattern = endpoint?.RoutePattern.RawText;
 
-            // Skip route pattern and use legacy Mvc route e.g.
+            // Skip route pattern if it resembles to a MVC route or null  e.g.
             // {controller=Home}/{action=Index}/{id?}
             if (!string.IsNullOrWhiteSpace(routePattern) &&
                 routePattern.StartsWith("{controller=") is false)
--- a/src/Sentry.AspNetCore/Extensions/HttpContextExtensions.cs
+++ b/src/Sentry.AspNetCore/Extensions/HttpContextExtensions.cs
@@ -16,7 +16,10 @@ namespace Sentry.AspNetCore.Extensions
             var endpoint = context.Features.Get<IEndpointFeature?>()?.Endpoint as RouteEndpoint;
             var routePattern = endpoint?.RoutePattern.RawText;
 
-            if (!string.IsNullOrWhiteSpace(routePattern))
+            // Skip route pattern and use legacy Mvc route e.g.
+            // {controller=Home}/{action=Index}/{id?}
+            if (!string.IsNullOrWhiteSpace(routePattern) &&
+                routePattern.StartsWith("{controller=") is false)
             {
                 return routePattern;
             }
--- a/src/Sentry.AspNetCore/Extensions/HttpContextExtensions.cs
+++ b/src/Sentry.AspNetCore/Extensions/HttpContextExtensions.cs
@@ -22,7 +22,7 @@ namespace Sentry.AspNetCore.Extensions
             {
                 // Skip route pattern if it resembles to a MVC route or null  e.g.
                 // {controller=Home}/{action=Index}/{id?}
-                return RouteHasMvcParameters(routePattern) ? ReplaceMcvParameters(routePattern, context) : routePattern;
+                return RouteHasMvcParameters(routePattern) ? ReplaceMvcParameters(routePattern, context) : routePattern;
             }
 #endif
 
@@ -48,7 +48,7 @@ namespace Sentry.AspNetCore.Extensions
         }
 
         // Internal for testing.
-        internal static string ReplaceMcvParameters(string route, HttpContext? context)
+        internal static string ReplaceMvcParameters(string route, HttpContext? context)
         {
             // Return RouteData or Null, marking the HttpContext as nullable since the output doesn't
             // shows the nullable output.
@@ -75,6 +75,7 @@ namespace Sentry.AspNetCore.Extensions
         // Internal for testing.
         internal static bool RouteHasMvcParameters(string route)
             => route.Contains("{controller=") || route.Contains("{action=") || route.Contains("{area=");
+
         public static string? TryGetTransactionName(this HttpContext context)
         {
             var route = context.TryGetRouteTemplate();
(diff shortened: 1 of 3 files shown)


test/Sentry.AspNetCore.Tests/Extensions/HttpContextExtensionsTests.cs
```diff
@@ -0,0 +1,43 @@
...
+
+        [Theory]
+        [InlineData("{area=MyArea}/{controller=Home}/{action=Index}/{id?}", "theArea/house/about/{id?}", "house", "about", "theArea")]
+        [InlineData("{area=MyArea}/{controller=Home}/{action=Index}/{id?}", "{area=MyArea}/house/about/{id?}", "house", "about", null)]
+        [InlineData("{controller=Home}/{action=Index}/{id?}", "house/about/{id?}", "house", "about", null)]
+        [InlineData("{controller=Home}/{action=Index}", "house/about", "house", "about", null)]
+        [InlineData("{controller=Home}/{id?}", "house/{id?}", "house", "about", null)]
+        [InlineData("{action=Index}/{id?}", "about/{id?}", null, "about", null)]
+        [InlineData("not/mvc/", "not/mvc/", "house", "about", "area")]
+        public void ReplaceMcvParameters_ParsedParameters(string routeInput, string assertOutput, string context, string action, string area)
```
bruno-garcia, line 27: Nice approach for testing. But missing some fixed bits in the URL like `abc/{controller=}`

And perhaps also: `abc/{controller}` <- this isn't an MVC format, it's just a parameter called controller.

test/Sentry.AspNetCore.Tests/Extensions/HttpContextExtensionsTests.cs
```diff
@@ -0,0 +1,72 @@
+using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Routing;
+using Sentry.AspNetCore.Extensions;
+using Xunit;
+
+namespace Sentry.AspNetCore.Tests.Extensions
+{
+    public class HttpContextExtensionsTests
+    {
+#if !NETCOREAPP2_1
```
bruno-garcia, line 10: Can this be defined the whole file instead?

test/Sentry.AspNetCore.Tests/Extensions/HttpContextExtensionsTests.cs
```diff
@@ -0,0 +1,72 @@
...
+        public void ReplaceMcvParameters_ParsedParameters(string routeInput, string assertOutput, string context, string action, string area)
+        {
+            // Arrange
+            var httpContext = new DefaultHttpContext();
+            AddRouteValuesIfNotNull(httpContext.Request.RouteValues, "controller", context);
+            AddRouteValuesIfNotNull(httpContext.Request.RouteValues, "action", action);
+            AddRouteValuesIfNotNull(httpContext.Request.RouteValues, "area", area);
+
+            // Act
+            var filteredRoute = HttpContextExtensions.ReplaceMcvParameters(routeInput, httpContext);
```
bruno-garcia, line 38: Mcv -> Mvc

src/Sentry.AspNetCore/Extensions/HttpContextExtensions.cs
```diff
@@ -43,6 +47,34 @@ internal static class HttpContextExtensions
...
+            {
+                route = Regex.Replace(route, "\\{area=[^\\}]+\\}", area);
+            }
+
+            return route;
+        }
+
+        // Internal for testing.
+        internal static bool RouteHasMvcParameters(string route)
+            => route.Contains("{controller=") || route.Contains("{action=") || route.Contains("{area=");
```
bruno-garcia, line 77:
```suggestion
        internal static bool RouteHasMvcParameters(string route)
            => route.Contains("{controller=") || route.Contains("{action=") || route.Contains("{area=");

```

src/Sentry.AspNetCore/Extensions/HttpContextExtensions.cs
```diff
@@ -43,6 +47,34 @@ internal static class HttpContextExtensions
...
+            }
+
+            if (routeData?.Values["action"]?.ToString() is { } action)
+            {
+                route = Regex.Replace(route, "\\{action=[^\\}]+\\}", action);
+            }
+
+            if (routeData?.Values["area"]?.ToString() is { } area)
+            {
+                route = Regex.Replace(route, "\\{area=[^\\}]+\\}", area);
```
bruno-garcia, line 69: Please remove the `\\`:

```suggestion
                route = Regex.Replace(route, "{area=[^}]+}", area);
```

test/Sentry.AspNetCore.Tests/Extensions/HttpContextExtensionsTests.cs
lucas-zimerman: sure

src/Sentry.AspNetCore/Extensions/HttpContextExtensions.cs
```diff
@@ -11,14 +12,17 @@ internal static class HttpContextExtensions
...
             // Requires .UseRouting()/.UseEndpoints()
             var endpoint = context.Features.Get<IEndpointFeature?>()?.Endpoint as RouteEndpoint;
             var routePattern = endpoint?.RoutePattern.RawText;
 
             if (!string.IsNullOrWhiteSpace(routePattern))
             {
-                return routePattern;
+                // Skip route pattern if it resembles to a MVC route or null  e.g.
+                // {controller=Home}/{action=Index}/{id?}
+                return RouteHasMvcParameters(routePattern) ? ReplaceMvcParameters(routePattern, context) : routePattern;
```
bruno-garcia, line 25:
```suggestion
                return RouteHasMvcParameters(routePattern) 
                    ? ReplaceMvcParameters(routePattern, context)
                    : routePattern;
```

bruno-garcia (approved):

PR #1202, opened 2021-09-17
Update CONTRIBUTING.md, add file to solution

josh-degraw:
#skip-changelog

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -7,7 +7,7 @@ For big feature it's advised to raise an issue to discuss it first.
 
 * Install [the .NET Core SDK](https://dot.net/) 
 * To quickly get up and running, you can just run `dotnet build`
-* To run a full build, test, and run all the validation logic locally before pushing, run `./build.sh` or `./buld.ps1`
+* To run a full build and test locally before pushing, run `./build.sh` or `./buld.ps1`
 
 ## Dependencies
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -6,7 +6,8 @@ For big feature it's advised to raise an issue to discuss it first.
 ## TLDR: 
 
 * Install [the .NET Core SDK](https://dot.net/) 
-* run `dotnet build` 
+* To quickly get up and running, you can just run `dotnet build`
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -6,10 +6,10 @@ For big feature it's advised to raise an issue to discuss it first.
 ## TLDR: 
 
 * Install [the .NET Core SDK](https://dot.net/) 
-* run `./build.sh` on macOS/Linux or `powershell ./build.ps1` on Windows.
```
josh-degraw, line 9: I'm not sure if that's still recommended by you guys, but the `./build.ps1` did nothing good for me, it just blew away my nuget cache and then I couldn't build anything until closing the IDE and running `dotnet build`

bruno-garcia: The build script will compile in release at least so those warnings as error for missing code docs will alert you.

Why did it blow up your nuget cache? The build script is just doing dotnet build

lucas-zimerman: Because of this code inside of build.sh
`dotnet nuget locals all --clear`

josh-degraw: Yep that'd be why. I'm not sure what exactly the reason is, but I've actually had a bunch of weird issues building locally. I've found that I can _usually_ get it to work if I manually run `dotnet restore` before building, but not always. The most common issue I've had is that it seems to be including the `Nullable` attributes for some reason 🤷🏻‍♂️

bruno-garcia: That should definitely be removed

bruno-garcia: https://github.com/getsentry/sentry-dotnet/commit/5626477b2b558c56bbd078a9b4163e2c6f731a05

bruno-garcia (approved):
This script is what runs in CI so the goal here is to have that same validation locally before pushing to avoid finding out things are broken only through GHA. Mainly a Release vs a Debug build usually but other things like adding sourcelink and doing a `dotnet pack`

Other than that feel free to merge.

lucas-zimerman (approved):

josh-degraw:
Okay I updated the wording to clarify the purpose of the `build` scripts. Let me know if that's okay or needs to change.

bruno-garcia (approved):

PR #1208, opened 2021-09-22
FIx: Add Sentry.Google.Cloud.Functions to craft.yml

lucas-zimerman:
#skip-changelog.

for reference https://github.com/getsentry/sentry-release-registry/tree/master/packages/nuget/Sentry.Google.Cloud.Functions

Because of that, the `Google cloud functions` is not getting the version updated.

### DIFF
--- a/.craft.yml
+++ b/.craft.yml
@@ -15,3 +15,4 @@ targets:
       nuget:Sentry.Log4Net:
       nuget:Sentry.Serilog:
       nuget:Sentry.NLog:
+      nuget:Sentry.Google.Cloud.Functions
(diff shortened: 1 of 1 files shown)


.craft.yml
```diff
@@ -15,3 +15,4 @@ targets:
       nuget:Sentry.Log4Net:
       nuget:Sentry.Serilog:
       nuget:Sentry.NLog:
+      nuget:Sentry.Google.Cloud.Functions
```
lucas-zimerman, line 18:
```suggestion
      nuget:Sentry.Google.Cloud.Functions:
```

bruno-garcia (approved):

PR #1209, opened 2021-09-22
Process isolation in offline caching

josh-degraw:
Testing around with an implementation for #1067.  I took some guidance from the existing draft PR, and made some adjustments. I put this up to see how you feel about the general approach before it gets too far. Open to any and all feedback.


src/Sentry/Internal/Http/CachingTransport.cs
```diff
@@ -37,44 +43,157 @@ internal class CachingTransport : ITransport, IAsyncDisposable, IDisposable
...
+                        continue;
+                    }
+
+                    // Skip processes that are still alive
+                    if (_processInfo.IsProcessActive(processId))
+                    {
+                        continue;
+                    }
+
+
```
lucas-zimerman, line 106:
```suggestion

```

src/Sentry/Internal/ProcessInfo.cs
```diff
@@ -103,5 +103,6 @@ private static DateTimeOffset GetStartupTime()
             using var proc = Process.GetCurrentProcess();
             return proc.StartTime.ToUniversalTime();
         }
+
```
lucas-zimerman, line 106:
```suggestion
```

test/Sentry.Tests/Internals/Http/CachingTransportTests.cs
```diff
@@ -257,5 +258,69 @@ public async Task DoesNotDeleteCacheIfConnectionWithIssue()
...
+
+            var dsnDir = Directory.EnumerateDirectories(Path.Combine(cacheDirectory.Path, "Sentry")).Single();
+            var processDirectories = Directory.EnumerateDirectories(dsnDir).ToList();
+
+            var envelopeFiles = Directory.EnumerateFiles(cacheDirectory.Path, "*", SearchOption.AllDirectories).ToList();
+
+            processDirectories.Should().HaveCount(2);
+
+            envelopeFiles.Should().HaveCount(6);
+
```
lucas-zimerman, line 323:
```suggestion
```

test/Sentry.Tests/Internals/Http/CachingTransportTests.cs
```diff
@@ -257,5 +258,69 @@ public async Task DoesNotDeleteCacheIfConnectionWithIssue()
             Assert.Equal(exception, receivedException);
             Assert.True(Directory.EnumerateFiles(cacheDirectory.Path, "*", SearchOption.AllDirectories).Any());
         }
+
+        [Fact(Timeout = 7000)]
+        public async Task IsolatesCacheDirectories()
```
lucas-zimerman, line 263: Is it possible to rephrase this test name? from what I understand, it tests different processes generating envelopes and it expects to have all those envelopes to be stored on their given isolated process directory?

test/Sentry.Tests/Internals/Http/CachingTransportTests.cs
```diff
@@ -257,5 +258,69 @@ public async Task DoesNotDeleteCacheIfConnectionWithIssue()
...
+                EnableProcessIsolationForCaching = true
+            };
+
+            var fileSystem = Substitute.ForPartsOf<FileSystemStub>();
+            fileSystem.When(f => f.DeleteFile(Arg.Any<string>())).DoNotCallBase();
+            fileSystem.When(f => f.DeleteDirectory(Arg.Any<string>())).DoNotCallBase();
+
+            // Act
+            // Send some envelopes with a failing transport to make sure they all stay in cache
+            {
```
lucas-zimerman, line 281: I don't follow why why this process has braces while the below one doesn't

test/Sentry.Tests/Internals/Http/CachingTransportTests.cs
```diff
@@ -257,5 +258,69 @@ public async Task DoesNotDeleteCacheIfConnectionWithIssue()
...
+            var processInfo2 = Substitute.For<IActiveProcessInfo>();
+            processInfo2.GetCurrentProcessId().Returns(2);
+            processInfo2.IsProcessActive(1).ReturnsForAnyArgs(false);
+            processInfo2.IsProcessActive(2).ReturnsForAnyArgs(true);
+            await using var transport = new CachingTransport(innerTransport, options, processInfo2, fileSystem);
+            for (var i = 0; i < 3; i++)
+            {
+                using var envelope = Envelope.FromEvent(new SentryEvent());
+                await transport.SendEnvelopeAsync(envelope);
+            }
```
lucas-zimerman, line 309: since you're doing the same action as above, perhaps transform it into a function
```csharp
async Task CreateTransactions(int amount)
{
            for (var i = 0; i < amount; i++)
            {
                using var envelope = Envelope.FromEvent(new SentryEvent());
                await transport.SendEnvelopeAsync(envelope);
            }
}

await CreateTransactions(3);
```

src/Sentry/Internal/ActiveProcessInfo.cs
```diff
@@ -0,0 +1,30 @@
...
+
+namespace Sentry.Internal
+{
+    internal sealed class ActiveProcessInfo : IActiveProcessInfo
+    {
+        public static IActiveProcessInfo Instance { get; } = new ActiveProcessInfo();
+
+        public int GetCurrentProcessId()
+        {
+            using var process = Process.GetCurrentProcess();
```
SimonCropp, line 12: shall we use prep for the new net6 process apis? https://devblogs.microsoft.com/dotnet/new-dotnet-6-apis-driven-by-the-developer-community/#process-path-id
```
#if NET6_0_OR_GREATER
            return Environment.ProcessId
#else
            using var process = Process.GetCurrentProcess();
            return process.Id;
#endif
```

src/Sentry/Internal/Http/CachingTransport.cs
```diff
@@ -37,44 +43,156 @@ internal class CachingTransport : ITransport, IAsyncDisposable, IDisposable
...
+            {
+                // No cache directories, that's fine
+                _options.DiagnosticLogger?.LogDebug(
+                    "Cache directory doesn't exist yet. Not scanning for leftover cache files."
+                );
+            }
+            catch (Exception ex)
+            {
+                _options.DiagnosticLogger?.LogError(
+                    "Failed to scan cache directory for leftover files.",
```
SimonCropp, line 154: can the cache directory path be added to this?

src/Sentry/Internal/Http/CachingTransport.cs
```diff
@@ -37,44 +43,156 @@ internal class CachingTransport : ITransport, IAsyncDisposable, IDisposable
...
+                            dirPath
+                        );
+                    }
+                }
+            }
+            catch (DirectoryNotFoundException)
+            {
+                // No cache directories, that's fine
+                _options.DiagnosticLogger?.LogDebug(
+                    "Cache directory doesn't exist yet. Not scanning for leftover cache files."
```
SimonCropp, line 148: can the cache directory path be added to this?

src/Sentry/Internal/ActiveProcessInfo.cs
bruno-garcia: Seems like this is .NET 5 even: https://docs.microsoft.com/en-us/dotnet/api/system.environment.processid?view=net-5.0

test/Sentry.Testing/FakeActiveProcessInfo.cs
```diff
@@ -0,0 +1,16 @@
+using System;
+using Sentry.Internal;
+
+namespace Sentry.Testing
+{
+    internal class FakeActiveProcessInfo : IActiveProcessInfo
```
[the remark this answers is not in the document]
josh-degraw: I actually did that and just forgot to remove this lol

src/Sentry/Internal/Http/CachingTransport.cs
```diff
@@ -37,44 +43,128 @@ internal class CachingTransport : ITransport, IAsyncDisposable, IDisposable
...
         private readonly Task _worker;
+        private string TryGetProcessSpecificCacheDirectoryPath(SentryOptions options)
+        {
+            if (!options.CacheProcessIsolationEnabled)
+            {
+                return _dsnCacheDirectoryPath;
+            }
+
+            return Path.Combine(_dsnCacheDirectoryPath, _processInfo.GetCurrentProcessId().ToString());
+
```
lucas-zimerman, line 54:
```suggestion
```

lucas-zimerman:
I was thinking right now and maybe we could simply leave an open File stream for each process.
The idea is simple, each process will have an open file, let's call it "SentryWatcher", once an instance is opened, it'll try to delete this file, if the process is open, the FILE stream will stay open and it'll throw an IOException because the file is open by another process and cannot be opened.
And in case of the process is not running, you'll be able to delete it and with that, you'll be sure that the isolated process is no longer active.

Another approach would be to use FileSystemWatcher to make some sort of communication system across processes, The idea is the same but you don't keep your FILE stream open, instead, you react from it getting changed (for example getting deleted) and react by creating the watched file, the other process will notice the file is back with `externalWatcher.WaitForChanged(WatcherChangeTypes.Created, 1000)` and will assume the process is active.

Two ideas that could replace the current System.Diagnostics checks for inactive Processes.

PR #1215, opened 2021-09-24
Fix: Avoid replacing the Transaction Name by null or empty

lucas-zimerman:
This fix does two issues:
- Avoid calling TryGetTransactionName if you don't have a transaction or if it Transaction.Name is already set.
- Avoid replacing the Transaction Name with null.

The second fix is important since transactions with null or Empty names get dropped.
https://github.com/getsentry/sentry-dotnet/blob/74312a4684e96babdfa555df2013fcc7c2ff9291/src/Sentry/SentryClient.cs#L120-L127

### DIFF
--- a/src/Sentry.AspNetCore/ScopeExtensions.cs
+++ b/src/Sentry.AspNetCore/ScopeExtensions.cs
@@ -87,7 +87,14 @@ namespace Sentry.AspNetCore
                     scope.SetTag("route.area", area);
                 }
 
-                scope.TransactionName = context.TryGetTransactionName();
+                // TransactionName refers to Transaction.Name, so skip this code if Transaction is null.
+                if (scope.Transaction != null &&
+                    scope.TransactionName is SentryTracingMiddleware.UnknownRouteTransactionName or null)
+                {
+                    scope.TransactionName =
+                        context.TryGetTransactionName()
+                        ?? SentryTracingMiddleware.UnknownRouteTransactionName;
+                }
             }
             catch (Exception e)
             {
--- a/src/Sentry.AspNetCore/SentryTracingMiddleware.cs
+++ b/src/Sentry.AspNetCore/SentryTracingMiddleware.cs
@@ -14,7 +14,7 @@ namespace Sentry.AspNetCore
     /// </summary>
     internal class SentryTracingMiddleware
     {
-        private const string UnknownRouteTransactionName = "Unknown Route";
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unchanged
```
josh-degraw, line 3:
```suggestion
## Unreleased
```
Did you mean this?

josh-degraw (approved):

CHANGELOG.md
lucas-zimerman: True, ty for spotting it.

lucas-zimerman:
This transaction code is Driving Me Nuts 🤪

lucas-zimerman:
After further refactor, I made changes to account for the Event consumption process.
By doing the previous fix, it lent to another bug in regard to events having the transaction named "Unknown Route" where previously they were not inserted due to another "bug" :crazy:

In addition, I removed the `scope.OnEvaluating += (_, _) => PopulateScope(context, scope);` from SentryTravingMiddleWare since it's a duplicated code that is also called by SentryMiddleware
https://github.com/getsentry/sentry-dotnet/blob/0cd15849687dd13be0f73c7edb3cfe8ddc40dfe0/src/Sentry.AspNetCore/SentryMiddleware.cs#L98-L108


I also removed the code to get the transaction name on TryStartTransaction since not always it would return a path, and since we do the same process afterwards with a higher chance to capture the correct name, I opted to remove it.

Despite that, I left the following code as is
`[SamplingExtensions.KeyForHttpRoute] = context.TryGetRouteTemplate()` It's not the correct code since the route template could be null when starting and only be set afterwards, but I believe that's a task for a future PR.

And, 
https://github.com/getsentry/sentry-dotnet/blob/0cd15849687dd13be0f73c7edb3cfe8ddc40dfe0/src/Sentry.AspNetCore/ScopeExtensions.cs#L70
routeData is nullable almost all the time when running Sentry.Samples.AspNetCore.MVC and it keeps throwing exceptions all the time, I left it as is since it resulted in a weird side effect of it setting the wrong Transaction Name to events .

To conclude, as requested by @bruno-garcia I inverted the TransactionTracer finish logic, by first removing itself from the scope and then Capturing it. I left the code without any try/catch block since both Hub ConfigureScope and CaptureTransactions are already covered with a try/catch block.

Sorry for the long explanation 😅

SimonCropp (approved):

PR #1224, opened 2021-09-29
remove IsPackable>false since it is the default

SimonCropp:
#skip-changelog


lucas-zimerman:
The default value used to be true and now it's false?

bruno-garcia (approved):

bruno-garcia:
I believe if the project uses `Microsoft.NET.Test.Sdk` so maybe not the case for benchmarks?

SimonCropp:
@lucas-zimerman sorry. yes you are correct. i conflated IsPackable and GeneratePackageOnBuild

PR #1225, opened 2021-09-29
CheckEolTargetFramework=false also for netcore2.1

SimonCropp:
* remove some warnings about 2.1 being EOL
 * Not point limiting it to netcoreapp3.0 since i assume we dont care about when fw become obsolete, and will instead make explicit decisions as to when to remove support for each one

#skip-changelog


bruno-garcia (approved):

bruno-garcia:
`#skip-changelog` can be used on the description to ignore danger for such PRs

PR #1245, opened 2021-10-07
Add options logger

SimonCropp:
There is a common pattern where we `options.DiagnosticLogger?.Log*`. This is often done many time in a single method. When it is a complex method it (IMO) decreases the readability of said methods. So I added some `options.Log*` extension methods. This increases some duplication in the simple extension methods, but also decreases some noise in every method that uses them. 

no #skip-changelog since the methods are currently internal


lucas-zimerman:
`should these new extension methods be public?`
I believe they should be internal, at least for now.

PR #1248, opened 2021-10-08
Use api snapshot to verify api changes

### LINKED ISSUE #1055: Check PRs against breaking changes
We want to make sure no breaking change gets merged without bumping a major. 

Possible solution: https://github.com/PublicApiGenerator/PublicApiGenerator

The approach taken by the OTel SDK to check API compatibility: https://github.com/open-telemetry/opentelemetry-dotnet/pull/2075/files

See: https://devblogs.microsoft.com/dotnet/package-validation/

SimonCropp:
fixes #1055

 * allows any api change to be easily verified in on location in a diff
 * build a history of apl changes over time that allows diff'ing of api between versions
 * forces people to acknowledge api changes, and hence reduces the risk of accidental breaking changes

#skip-changelog

### DIFF
--- a/src/Sentry.AspNet/Internal/SystemWebRequestEventProcessor.cs
+++ b/src/Sentry.AspNet/Internal/SystemWebRequestEventProcessor.cs
@@ -37,7 +37,7 @@ namespace Sentry.AspNet.Internal
             }
             catch (HttpException)
             {
-                _options.DiagnosticLogger?.LogDebug("HttpException not available to retrieve context.");
+                _options.LogDebug("HttpException not available to retrieve context.");
                 return @event;
             }
 
@@ -53,7 +53,7 @@ namespace Sentry.AspNet.Internal
             {
                 // Ignored since it can throw on WCF on the first event.
                 // See #390
-                _options.DiagnosticLogger?.LogDebug("Ignored NRE thrown on System.Web.HttpContext.Request.QueryString");
+                _options.LogDebug("Ignored NRE thrown on System.Web.HttpContext.Request.QueryString");
             }
 
             foreach (var key in context.Request.Headers.AllKeys)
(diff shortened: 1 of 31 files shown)


lucas-zimerman:
Could you commit a test where you change an interface, for example, IEventLike, and adds an additional function/parameter, just for registering that the test will fail?

Also about the verified files, in case of a breaking change, you'll need to manually edit those verified files or run a script/build command?

SimonCropp:
@lucas-zimerman 

> Could you commit a test where you change an interface, for example, IEventLike, and adds an additional function/parameter, just for registering that the test will fail?

I added an extra member to `SentryEvent`, didnt accept the change, and pushed. you can see the [broken build](https://github.com/getsentry/sentry-dotnet/actions/runs/1322156147). The build output contains the failed test and the verified + received content
```
[xUnit.net 00:00:10.21]     Sentry.Tests.ApiApprovalTests.Run [FAIL]
Error: Sentry.Tests.ApiApprovalTests.Run: VerifyException : Results do not match.
  Failed Sentry.Tests.ApiApprovalTests.Run [545 ms]
  Error Message:
   VerifyException : Results do not match.
Differences:
Received: ApiApprovalTests.Run.DotNet5_0.received.txt
Verified: ApiApprovalTests.Run.DotNet5_0.verified.txt
Received Content:
[assembly: System.CLSCompliant(true)]
```
Since for some snapshot failures it can be difficult to compare in text, the mismatched received files are uploaded to the GH action artifacts:
image

> Also about the verified files, in case of a breaking change, you'll need to manually edit those verified files or run a script/build command?

There are a [variety of ways to accept a new snapshot](https://github.com/VerifyTests/Verify#snapshot-management). Many people use a combination of approaches. If you want to bypass the "many accept a change" part, you can use [AutoVerify](https://github.com/VerifyTests/Verify/blob/main/docs/verify-options.md#autoverify). This means any change in the api will automatically be written to the snapshot file, and the change will still be visible in your git diff tools and when viewing the commit/PR on github.

SimonCropp:
BTW here is the experience on my machine, using diffenginetray, when i made the above test fail:

image

PR #1252, opened 2021-10-11
Update Sentry.Samples.GenericHost

SimonCropp:
I tried the GenericHost sample and got a missingMethodException since the nugets were not in sync.

So this PR updates them to the current gen of nugets

#skip-changelog


SimonCropp:
replaces https://github.com/getsentry/sentry-dotnet/pull/1030/files

PR #1270, opened 2021-10-15
Data from Scope in options should be applied on each request

### LINKED ISSUE #968: aspnetcore: Data from Scope in options should be applied on each request
The `SentryMiddleware` class takes a `Func<IHub>` because that can change after initialization.

This means the code below can't be correct:

https://github.com/getsentry/sentry-dotnet/blob/d25c97f0e5c77b520e30877444e93a9147afb9cd/src/Sentry.AspNetCore/SentryMiddleware.cs#L59-L63

If after instantiation the func returns a new instance, that hub will not have the correct values.

lucas-zimerman:
The PR applies the configure scopes stored on Sentry options into the hub once and reapplies it if the hud changed.

Close #968


src/Sentry.AspNetCore/SentryMiddleware.cs
```diff
@@ -104,7 +101,16 @@ public async Task InvokeAsync(HttpContext context)
                     // In case of event, all data made available through the HTTP Context at the time of the
                     // event creation will be sent to Sentry
 
-                    scope.OnEvaluating += (_, _) => PopulateScope(context, scope);
+                    scope.OnEvaluating += (_, _) =>
+                    {
+                        if (!object.ReferenceEquals(hub, _previousHub))
```
bruno-garcia, line 106: The point of adding it here is there's no need to compare reference to Hub.

bruno-garcia (changes requested):

bruno-garcia (approved):

Issue #1283, opened 2021-10-21
Sentry doesn't correctly set error.unhandled, so issues don't appear on the Sentry project page


kanadaj:
### Environment

How do you use Sentry?
Self-hosted, 21.10.0

Which SDK and version?
.NET 3.10.0

### Steps to Reproduce

1. Set up Sentry on an ASP.NET Core application, trigger an uncaught exception in a controller or page

### Expected Result

Issue should appear in the Sentry project frequent issues list

### Actual Result

![image](https://user-images.githubusercontent.com/3247357/138334850-22fb93c7-1c6d-47bf-a1e6-ac310fd77609.png)

See that errors are in the chart, but issues don't show in the list

Do note that errors from inside a middleware show correctly:

![image](https://user-images.githubusercontent.com/3247357/138335154-7d3a77a0-4d42-47c1-8248-ae85037d8f9e.png)

PR #1284, opened 2021-10-22
Fix missing handled tag when events are logged via an ASP.NET Core pipeline logger

kanadaj:
Fixes https://github.com/getsentry/sentry-dotnet/issues/1283

This is a draft fix for an issue with the Serilog sink that causes issues not to be marked as handled.

Realistically this should have a test written for it, but I have no idea how to mock the entire ASP.NET Core pipeline, and testing this would require that. Also not entirely sure if this is the best way to handle this. Maybe an event processor could also do this instead?

@bruno-garcia @lucas-zimerman thoughts?


test/Sentry.Serilog.Tests/AspNetCoreIntegrationTests.cs
```diff
@@ -0,0 +1,37 @@
...
+using Sentry.Testing;
+using Xunit;
+
+namespace Sentry.Serilog.Tests
+{
+    [Collection(nameof(SentrySdkCollection))]
+    public class AspNetCoreIntegrationTests : AspNetSentrySdkTestFixture
+    {
+
+#if NET5_0_OR_GREATER || NETCOREAPP3_1_OR_GREATER
```
bruno-garcia, line 17: Should include NET5:

```suggestion
#if NETCOREAPP3_1_OR_GREATER
```

test/Sentry.Serilog.Tests/AspNetSentrySdkTestFixture.cs
```diff
@@ -0,0 +1,63 @@
...
+                    LastEvent = @event;
+                    return @event;
+                };
+            };
+
+            ConfigureApp = app =>
+            {
+                #if NET5_0_OR_GREATER || NETCOREAPP3_1_OR_GREATER
+                app.UseExceptionHandler("/error");
+                #endif
```
bruno-garcia, line 39:
```suggestion
#if NETCOREAPP3_1_OR_GREATER
                app.UseExceptionHandler("/error");
#endif
```

test/Sentry.Serilog.Tests/Utils/FakeSentryServer.cs
```diff
@@ -0,0 +1,43 @@
...
+    {
+        public static TestServer CreateServer(IReadOnlyCollection<RequestHandler> handlers)
+        {
+            var builder = new WebHostBuilder()
+                .UseDefaultServiceProvider(di => di.EnableValidation())
+                .Configure(app =>
+                {
+                    app.Use(async (context, next) =>
+                    {
+                        var handler = handlers.FirstOrDefault(p => p.Path == context.Request.Path);
```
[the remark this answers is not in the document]
bruno-garcia: Technically yes but this is used for testing. it would match if u setup your test with two handlers matching the same path

bruno-garcia (approved):

bruno-garcia:
@kanadaj Thanks for the PR.
Could you please add to changelog:

```
Fix missing handled tag when events are logged via an ASP.NET Core pipeline logger ([#1284](getsentry/sentry-dotnet/pull/1284))

```

PR #1298, opened 2021-11-03
move to c#10. keep c#9 for sentry project

SimonCropp:
#skip-changelog


bruno-garcia (commented):
Change makes sense but very oddly broke a single test on 3 platforms.

bruno-garcia:
Rerunning. I wonder if a new SDK version was installed (since we don't pin the SDK on global.json nor we install a specific version).

bruno-garcia:
Replacing it with #1311

PR #1310, opened 2021-11-04
Feat/tunnel ip forwarding

### LINKED ISSUE #1309: `Sentry.Tunnel` doesn't include `X-Forwarded-To`
### Summary
 
A high level overview of why we supply tunnelling in Sentry is described in [NextJS: Using the tunnel option](https://docs.sentry.io/platforms/javascript/guides/nextjs/troubleshooting/#using-the-tunnel-option).

I'm assuming we're doing something similar in our .NET backend... so to prevent ad blockers etc. getting in the way of sending data to Sentry, your Angular front end is sending these to your back end (which is sitting on host that's allowed by the CORS policy) and the back end then forwards those events to Sentry on behalf of the Angular SPA?

Assuming I've followed along so far, our tunnelling logic then needs to get the **_user's IP address_** (i.e. the ip:port of the machine calling the tunnelling endpoint) and use this value to set the `X-Forwarded-For` header when forwarding requests. 

### References

Similar to what's discussed here: https://forum.sentry.io/t/real-client-ip-with-sentry-nextjs-tunnel/15438/2

kanadaj:
Addresses #1309

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## Unreleased
 
-### Features
+### Fixes
 
 - Forward the IP of the client with whe tunnel middleware ([#1310](getsentry/sentry-dotnet/pull/1310))
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
+
+### Features
+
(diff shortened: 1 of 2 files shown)


bruno-garcia (approved):

CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
+
+### Features
```
```suggestion
### Fixes
```

PR #1317, opened 2021-11-08
Fix: Don't leave leftovers for SQLClient

lucas-zimerman:
All connection spans are closed with the correct Connection Id. Previously only one of the connection Spans with the same connection Id was going to be set. Now, this results in spans with deadlines instead of being ignored due to this PR #1296.

Might be hard to visualize, but this
image
Will now become 
image

The appearance will depend on the connectivity code used by the client, for some operations the appearance will remain the same, as shown below:
image


test/Sentry.DiagnosticSource.Tests/SentrySqlListenerTests.cs
```diff
@@ -402,28 +440,31 @@ void SimulateDbRequest(List<Guid> connnectionOperationIds, List<Guid> queryOpera
...
             });
+
             // Assert all Query spans have the correct ParentId Set and not the Transaction Id.
             Assert.All(querySpans, querySpan =>
             {
                 Assert.True(querySpan.IsFinished);
                 var connectionId = (Guid)querySpan.Extra[SentrySqlListener.ConnectionExtraKey];
-                var connectionSpan = connectionSpans.Where(span => span.Extra.ContainsKey(SentrySqlListener.ConnectionExtraKey) && (Guid)span.Extra[SentrySqlListener.ConnectionExtraKey] == connectionId);
-                Assert.Single(connectionSpan);
+                var connectionSpan = connectionSpans.First(span => span.SpanId == querySpan.ParentSpanId);
```
SimonCropp, line 464: can this be a single here? since First is ambiguous as to if more than one is valid

lucas-zimerman: Yes it could

PR #1318, opened 2021-11-08
Update X-Sentry-Auth header to include correct sdk name and version

### LINKED ISSUE #1285: Inconsistent reporting of SDK information
I noticed a couple of inconsistencies when comparing the behavior of the .NET SDK with others when looking at a sample app: https://github.com/getsentry/sentry-sdk-benchmark/pull/56

- [ ] SDK name in `X-Sentry-Auth` header differs from `sdk.name` in the event payload
- [ ] `X-Sentry-Auth` header missing SDK version

Spec reference: https://develop.sentry.dev/sdk/overview/#authentication

## .NET

```
X-Sentry-Auth: Sentry sentry_version=7,sentry_client=sentry.dotnet,sentry_key=sentry,sentry_timestamp=1634938521
...

  "sdk": {
    "packages": [
      {
        "name": "nuget:Sentry",
        "version": "3.9.2"
      },
      {
        "name": "nuget:Sentry.AspNetCore",
        "version": "3.9.2"
      }
    ],
    "name": "sentry.dotnet.aspnetcore",
    "version": "3.9.2"
  },
...
```

Notice in the header **`sentry_client=sentry.dotnet`**, expected **`sentry_client=sentry.dotnet.aspnetcore/3.9.2`**.

## Python

For comparison, this is what the Python SDK sends:

```
X-Sentry-Auth: Sentry sentry_key=sentry, sentry_version=7, sentry_client=sentry.python/1.3.0
...

  "sdk": {
    "name": "sentry.python",
    "version": "1.3.0",
    "packages": [
      {
        "name": "pypi:sentry-sdk",
        "version": "1.3.0"
      }
    ],
    "integrations": [...]
  },
...
```

## Node

And Node:

```
X-Sentry-Auth: Sentry sentry_version=7, sentry_client=sentry.javascript.node/6.11.0, sentry_key=sentry
...

    "sdk": {
        "integrations": [...],
        "name": "sentry.javascript.node",
        "version": "6.11.0",
        "packages": [
            {
                "name": "npm:@sentry/node",
                "version": "6.11.0"
            }
        ]
    },
...
```

josh-degraw:
I didn't know if it makes more sense to change what `SentryOptions.ClientVersion` points to, since that seems to be incorrect, so here is a fix just within the `HttpTransport`.

Closes #1285


src/Sentry/Internal/Http/HttpTransport.cs
```diff
@@ -259,6 +260,22 @@ await using (envelopeFile)
             }
         }
 
+        private string BuildAuthHeader(Dsn dsn)
+        {
+            var builder = new StringBuilder(100)
```
bruno-garcia, line 265: When I tested it out (11 years ago :x) using a StringBuilder on a fixed-size small string similar to this case actually performed worse than a simple string concat.

Optimizing here (I'm guessing, to be sure we need some benchmarks) would be to have the fixed part assembled first, then just concatenate the timestamp to the end of the string:

```
           _authHeaderPrefix =
                $"Sentry sentry_version={_options.SentryVersion}," +
                $"sentry_client={_options.ClientVersion}," +
                $"sentry_key={dsn.PublicKey}," +
                (dsn.SecretKey is { } secretKey ? $"sentry_secret={secretKey}," : null) +
                "sentry_timestamp=";
           ...
           return _authHeaderPrefix + _clock.GetUtcNow().ToUnixTimeSeconds(); 
```

bruno-garcia (commented):
Makes sense to just change in the transport. I have a suggestion about the approach

src/Sentry/Internal/Http/HttpTransport.cs
josh-degraw: Huh, interesting. I'll just switch it back, no need for premature optimization if it doesn't obviously provide obvious benefit 👍🏻

bruno-garcia:
Tried to fix the merge conflict and the test that was failing (not the best assert approach but hopefully I get the build to pass)

bruno-garcia (approved):

bruno-garcia:
OK sounds like some of the `using` was needed in there. Sorry

lucas-zimerman:
Superseded by #1333

PR #1320, opened 2021-11-09
try coverlet.collector instead coverlet.msbuild of to fix intermittent build issue

SimonCropp:
#skip-changelog


SimonCropp:
no longer produced coverage. so a failure

PR #1328, opened 2021-11-10
handle exception from crashedLastRun callback

bruno-garcia:
- [x] Add a test that verifies the event is captured even if the callback throws


lucas-zimerman:
Just to be sure:
` Add a test that verifies the event is captured even if the callback throws`
you wanna be sure that sessions are captured or events?

bruno-garcia:
> Just to be sure: ` Add a test that verifies the event is captured even if the callback throws` you wanna be sure that sessions are captured or events?

Both actually

SimonCropp (approved):

PR #1346, opened 2021-11-23
infer netfx version from min reg versions

SimonCropp:
For me SetReleaseAndVersionNetFx_OnNetFx_NonNullReleaseAndVersion was failing. This is dues to me having a newer net48 on my machine which has a reg key of 528449.

This PR changes the detection logic to infer by the [min versions](https://docs.microsoft.com/en-us/dotnet/framework/migration-guide/how-to-determine-which-versions-are-installed#minimum-version) instead of a hard coded dictionary, so it is more resilient moving forward

### DIFF
--- a/src/Sentry/PlatformAbstractions/FrameworkInfo.NetFx.cs
+++ b/src/Sentry/PlatformAbstractions/FrameworkInfo.NetFx.cs
@@ -12,6 +12,7 @@ namespace Sentry.PlatformAbstractions
     {
         internal const string NetFxNdpRegistryKey = @"SOFTWARE\Microsoft\NET Framework Setup\NDP\";
         internal const string NetFxNdpFullRegistryKey = @"SOFTWARE\Microsoft\NET Framework Setup\NDP\v4\Full\";
+
         /// <summary>
         /// Get the latest Framework installation for the specified CLR
         /// </summary>
@@ -174,9 +175,57 @@ namespace Sentry.PlatformAbstractions
 
         internal static Version GetNetFxVersionFromRelease(int release)
         {
-            _ = NetFxReleaseVersionMap.TryGetValue(release, out var version);
-            _ = Version.TryParse(version, out var parsed);
-            return parsed;
+            if (release >= 528040)
+            {
+                return new Version(4, 8);
+            }
+
+            if (release >= 461808)
+            {
+                return new Version(4, 7, 2);
+            }
+
+            if (release >= 461308)
+            {
+                return new Version(4, 7, 1);
+            }
+
+            if (release >= 460798)
+            {
+                return new Version(4, 7);
+            }
+
+            if (release >= 394802)
+            {
+                return new Version(4, 6, 2);
+            }
+
+            if (release >= 394254)
+            {
+                return new Version(4, 6, 1);
+            }
+
+            if (release >= 393295)
+            {
+                return new Version(4, 6);
+            }
+
+            if (release >= 379893)
+            {
+                return new Version(4, 5, 2);
+            }
+
+            if (release >= 378675)
+            {
+                return new Version(4, 5, 1);
+            }
+
+            if (release >= 378389)
+            {
+                return new Version(4, 5);
+            }
+
+            throw new($"Unknown version for NetFx registry version: {release}");
         }
     }
 }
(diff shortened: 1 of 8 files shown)


src/Sentry/PlatformAbstractions/FrameworkInfo.NetFx.cs
```diff
@@ -172,11 +173,60 @@ private static FrameworkInstallation GetFromV4(RegistryKey subKey, string subKey
             return ndpKey?.GetInt("Release");
         }
 
+        // https://docs.microsoft.com/en-us/dotnet/framework/migration-guide/how-to-determine-which-versions-are-installed#minimum-version
         internal static Version GetNetFxVersionFromRelease(int release)
         {
-            _ = NetFxReleaseVersionMap.TryGetValue(release, out var version);
-            _ = Version.TryParse(version, out var parsed);
-            return parsed;
+            if (release >= 528040)
```
bruno-garcia, line 179: Wasn't there a 4.8.1 going to come out? I'm afraid if they change their mind about 4.8 being the last we'll be falsely reporting 4.8 for a bunch of events

bruno-garcia: We have this in Relay too: https://github.com/getsentry/relay/blob/94a78b8c5965af9935e24ff8270968da07305432/relay-general/src/store/normalize/contexts.rs#L58-L63

bruno-garcia: Added the missing one: https://github.com/getsentry/relay/pull/1142

SimonCropp: > Wasn't there a 4.8.1 going to come out? I'm afraid if they change their mind about 4.8 being the last we'll be falsely reporting 4.8 for a bunch of events

that was the behavior of the previous code

bruno-garcia: Sorry, I don't follow. The old code relied on a map: https://github.com/getsentry/sentry-dotnet-platform-abstractions/blob/2e715e9ebe33c02a9ee9a9705d2dc1481ba90e59/src/Sentry.PlatformAbstractions/FrameworkInfo.cs#L19-L43

How would it report 4.8 for future version?

SimonCropp: sorry. i mis-spoke. it throws for an unknown version

bruno-garcia: Agree that's suboptimal. Could be replaced by TryGetValue and return `null` if value isn't known. 
If we agree that reporting nothing is better than reporting the wrong value

SimonCropp: how about i change it to also report the reg value. then for an unknown reg value it wont log a version. at least then we can debug what the new reg value means

bruno-garcia: That Sounds better but the only fear is that bogus data would result in high cardinality value being sent over to an indexed field. We could simple log with diagnostic logger a Warn level message with the reg value saying it doesn't match and report nothing. It's better than polluting Sentry with a bunch of new Runtime Version completely off the expected format and possibly adding lots of new values

SimonCropp: actually given there is no upper bound, this is not possible. ie given an arbitrary higher number than 528040, we can not tell if it should be 4.8 or 4.8.1

SimonCropp:
could we instead use the last part of the `System.Runtime.InteropServices.RuntimeEnvironment.GetRuntimeDirectory();` path?

bruno-garcia:
The docs don't have example values: https://docs.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.runtimeenvironment.getruntimedirectory?view=net-6.0 

What does it look like?

I ask because the value we return here is indexed in Sentry. Customers can have alerts and queries based on these values so we can't really change their format (I.e: `.NET 6.0.0` becomes `.net 6.0.0`)

src/Sentry/PlatformAbstractions/FrameworkInfo.NetFx.cs
bruno-garcia: We could also amend at the call site, that if we get `null` back from this, we use the reg value?
Maybe I'm splitting hairs here, if bogus values are coming in, the problem is at the call site, so happy with your proposal

lucas-zimerman:
I just noticed that it's also failing for me locally.

SimonCropp:
closin this. will add 528040 in the short term

PR #1371, opened 2021-12-05
remove await using noise in test project

SimonCropp:
before 

```
#if !NET461 && !NETCOREAPP2_1
        await
#endif
            using var originalStream = new MemoryStream();
```

after 
```
using var originalStream = new MemoryStream();
```

in tests there will be no noticeable perf impact. and the result is much easier to read

#skip-changelog


bruno-garcia:
I rather we stop doing such changes, reverting something that adds no value.

PR #1382, opened 2021-12-13
add unity config

SimonCropp:
#skip-changelog

so we can have custom config / apis for unity

### DIFF
--- a/src/Sentry/Sentry.csproj
+++ b/src/Sentry/Sentry.csproj
@@ -9,8 +9,15 @@
     <NoWarn Condition="$(TargetFramework) == 'netstandard2.0'">$(NoWarn);RS0017</NoWarn>
     <DefineConstants>$(AdditionalConstants)</DefineConstants>
     <Configurations>Debug;Release;Release_Unity;Debug_Unity</Configurations>
-    <DefineConstants Condition="$(Configuration) == 'Release_Unity'">$(DefineConstants);RELEASE;Unity</DefineConstants>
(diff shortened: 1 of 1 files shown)


SimonCropp:
@bruno-garcia if this looks good to u, can u merge and update the module in unity

bruno-garcia:
I believe we can define properties in the unity directory.build.props in the unity repo since we have the submodule as a sub folder. Would that be a better approach?

bruno-garcia:
On a second thought it won't because it already includes a top level file so we'd need to include manually a parent

lucas-zimerman (approved):

src/Sentry/Sentry.csproj
```diff
@@ -8,6 +8,16 @@
     <Description>Official SDK for Sentry - Open-source error tracking that helps developers monitor and fix crashes in real time.</Description>
     <NoWarn Condition="$(TargetFramework) == 'netstandard2.0'">$(NoWarn);RS0017</NoWarn>
     <DefineConstants>$(AdditionalConstants)</DefineConstants>
+    <Configurations>Debug;Release;Release_Unity;Debug_Unity</Configurations>
+  </PropertyGroup>
+
+  <PropertyGroup Condition="$(Configuration) == 'Release_Unity'">
```
bruno-garcia, line 14: I believe it's best we add an additional Property instead so we can continue to use Configurion=Release
The reasoning is that there (?) properties and behavior tied to Configuration `Release` and we might be missing those due to introducing a new configuration. 

Can we just pass an additional properly: `dotnet msbuild /p:Unity=true` ?

SimonCropp: but that wont work when building from and ide

PR #1395, opened 2021-12-15
add solution filters

SimonCropp:
#skip-changelog


bruno-garcia (approved):
If there was a way to say: Include everything except `/samples` it would go easier on the maintenance

SimonCropp:
@bruno-garcia not at the moment. here is the feature request https://developercommunity.visualstudio.com/t/solution-filter-should-allow-for-includeexclude-ba/1090914

PR #1396, opened 2021-12-16
android solution filters

SimonCropp:
#skip-changelog


buildMac.sh
```diff
@@ -8,7 +8,7 @@ if [ "$GITHUB_ACTIONS" == "true" ]
         testLogger="console"
 fi
 
-dotnet test -c Release -l $testLogger \
+dotnet test Sentry_Mac.slnf -c Release -l $testLogger \
```
bruno-garcia, line 11:
```suggestion
dotnet test $solutionFilterFile -c Release -l $testLogger \
```

.github/workflows/build.yml
```diff
@@ -20,11 +20,11 @@ jobs:
...
         include:
           - os: windows-latest
-            build_script: .\build.ps1
+            build_script: .\buildWindows.ps1
           - os: ubuntu-latest
-            build_script: ./build.sh
+            build_script: ./buildLinux.sh
           - os: macos-latest
-            build_script: ./build.sh
+            build_script: ./buildMac.sh
```
bruno-garcia, line 27:
```suggestion
            env: solutionFilterFile =Sentry_Mac.slnf
            build_script: ./buildMac.sh
```

Sentry_Linux.slnf
```diff
@@ -0,0 +1,49 @@
+{
+  "solution": {
+    "path": "Sentry.sln",
+    "projects": [
+      "benchmarks\\Sentry.Benchmarks\\Sentry.Benchmarks.csproj",
```
bruno-garcia, line 5: We could generate these files with a script by crawling **/*.csproj with some filters? Then we can just define which csproj's to escape for each platform, and in CI we build a fresh new filter for each run

SimonCropp: do we change/add/remove project regularly enough to justify wiring code to do this?

lucas-zimerman:
Ugh, I'm a bit lost here, PR mentions `Android` but I see no changes in regard to Android but script changes for Windows/Linux/Mac.

buildMac.sh
```diff
@@ -0,0 +1,15 @@
...
+set -e
+
+if [ "$GITHUB_ACTIONS" == "true" ]
+    then
+        testLogger="GitHubActions;report-warnings=false"
+    else
+        testLogger="console"
+fi
+
+dotnet test $solutionFilterFile -c Release -l $testLogger \
```
bruno-garcia, line 11: The idea here is to delete the `buildLinux` and `buildMac` and have a single script

Sentry_Linux.slnf
bruno-garcia: I wasn't thinking about how often we change, simply the fact it will be wrong eventually. Any new manual step we add should be considered carefully as it can lead to a lot of thing 'to keep in mind'.

It's just a shame these filter files are so limited that it would require scripting from our end. Feel free to merge once the duplicate script thing is removed (you merged the suggestions but ti's missing going back to a single .sh file

bruno-garcia: Once there's a way, we can change the filters: https://github.com/getsentry/sentry-dotnet/issues/1405

bruno-garcia (approved):

PR #1399, opened 2021-12-16
Report ThreadPool stats

### LINKED ISSUE #938: Report ThreadPool stats
At least:

SetMinThreads(Int32, Int32)
GetMaxThreads(Int32, Int32)
GetAvailableThreads(Int32, Int32)

Other useful data easy to include are: GC (server/workstation), current used mem etc:

[GetTotalMemory](https://docs.microsoft.com/en-us/dotnet/api/system.gc.gettotalmemory?view=net-5.0#System_GC_GetTotalMemory_System_Boolean_)
> Retrieves the number of bytes currently thought to be allocated. A parameter indicates whether this method can wait a short interval before returning, to allow the system to collect garbage and finalize objects.

[GetTotalAllocatedBytes](https://docs.microsoft.com/en-us/dotnet/api/system.gc.gettotalallocatedbytes?view=net-5.0#System_GC_GetTotalAllocatedBytes_System_Boolean_)
> Gets a count of the bytes allocated over the lifetime of the process.

`MaxGeneration`

SimonCropp:
fixes #938

### DIFF
--- a/src/Sentry/Internal/MainSentryEventProcessor.cs
+++ b/src/Sentry/Internal/MainSentryEventProcessor.cs
@@ -13,6 +13,7 @@ namespace Sentry.Internal
         internal const string CultureInfoKey = "Current Culture";
         internal const string CurrentUiCultureKey = "Current UI Culture";
         internal const string MemoryInfoKey = "Memory Info";
+        internal const string ThreadPoolInfoKey = "ThreadPool Info";
 
         private readonly Enricher _enricher;
 
@@ -56,6 +57,7 @@ namespace Sentry.Internal
             }
 
             AddMemoryInfo(@event.Contexts);
+            AddThreadPoolInfo(@event.Contexts);
             if (@event.ServerName == null)
             {
                 // Value set on the options take precedence over device name.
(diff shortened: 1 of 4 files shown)


src/Sentry/Internal/MainSentryEventProcessor.cs
```diff
@@ -56,6 +57,7 @@ public SentryEvent Process(SentryEvent @event)
             }
 
             AddMemoryInfo(@event.Contexts);
+            AddThreadPoolInfo(@event.Contexts);
```
lucas-zimerman, line 60: @bruno-garcia  we only want that information for Events?

SimonCropp: forgive my ignorance, where else would we have it?

bruno-garcia: I believe Lucas is referring to Transactions as the other possible place to add it.
We only run `SentryEvent` instances here so transactions won't be getting this information.

A few things go through when deciding: 
* Does this information help understand performance problems with the app?
* Is it cheap to retrieve this information (as in resource intensiveness)?

If the answer is 'yes' to both, it's a good candidate. Ideally we add information that's relative to the transaction itself (like what happened during the duration of that transaction). Like: How many GC gen0/1/2 ran during the transaction, for example. As 'point in time' information alone within a transaction might not be as valuable. 

Ultimately, this might be more useful only as Metrics. Which is something we'll look into adding in the future

PR #1400, opened 2021-12-17
fix: Recover from broken context/attachment

bitsandfoxes:
To throw without a logger available means we lose the whole event with `options.Debug = false;`.


bruno-garcia (approved):

bruno-garcia:
@bitsandfoxes CI won't pass because we're deleting the `.targets` here. We need a solution to get the two repos working together again. Maybe @SimonCropp can give us a hand

bruno-garcia:
Closing this in favor of #1401 
We still need to resolve the `.targets` issue on a separate PR

PR #1401, opened 2021-12-17
fix: Keep events in case of nonserializable context

bitsandfoxes:


bruno-garcia (approved):
Turns out the serializes does push data into the writer before it fails. So it can end up at a broken state when we handle the exception. Added some compensating code there to fix the JSON and some extra logging.

bruno-garcia:
Turns out what exception is thrown changed from netcore 2.1 and other targets.

Issue #245, opened 2019-07-21
Enhancements via netcoreapp3.0


bruno-garcia:
Milestone 3.0.0 for the SDK is intended to target `netcoreapp3.0` and get access to the new APIs introduced.

* [x] Send as Extra or Tags the values for `IsDynamicCodeCompiled` and `IsDynamicCodeSupported`.  [PR](https://github.com/getsentry/sentry-dotnet/pull/1418)
* [ ] IIS Server variables (accessible via ASP.NET Core now).
* [ ] Read request body via `System.IO.Pipelines`.
* [ ] Opt-in for `IConfigurationRoot.GetDebugView`.
* [x] GC info [GetMemoryInfo](https://docs.microsoft.com/en-us/dotnet/api/system.gc.getgcmemoryinfo?view=netcore-3.0)  and [GetTotalAllocatedBytes](https://docs.microsoft.com/en-us/dotnet/api/system.gc.gettotalallocatedbytes?view=netcore-3.0). See  [PR](https://github.com/getsentry/sentry-dotnet/pull/1337)

jeanpoelie:
Do you have a targeted release date for this ticket? We are looking into upgrading to .net core 3.0 and Sentry is one of the blocking packages right now. Thank you for your response!

bruno-garcia:
@jeanpoelie it should be working fine with .NET Core 3.0. What's blocking you?
This ticket is just for enhancements.
I hope to get around doing these sometime in December or early next year.

jeanpoelie:
@bruno-garcia the stable version (1.2.0) does not work with .net core 3.0, when I use the pre-release version (2.0.0-beta4) it does work, but generally I don't like using prerelease versions and am wondering when it will be into stable mode :)

Thank you for the answer, we will for now keep using the beta version since I have not run into any issues regarding it being unstable yet :)

bruno-garcia:
The only known bug we have right now (I'm working on it) also affects 1.2.0.

The reason 2.0.0-beta is not out of beta is not that it's unstable but that we want to add a few more breaking changes (likely not to affect most users) on the 2.0.0 release.
If you have issues with the latest preview please raise it and we'll help you as quick as we can with a fix.

SimonCropp:
@bruno-garcia didnt we decide BackgroundWorker, SentryClient, and Hub should not be disposable?

mattjohnsonpint:
Closing stale issue.

PR #1418, opened 2021-12-28
add IsDynamicCode* to events

SimonCropp:
related to #245

### DIFF
--- a/src/Sentry/Internal/MainSentryEventProcessor.cs
+++ b/src/Sentry/Internal/MainSentryEventProcessor.cs
@@ -2,6 +2,7 @@ using System;
 using System.Collections.Generic;
 using System.Globalization;
 using System.Linq;
+using System.Runtime.CompilerServices;
 using System.Threading;
 using Sentry.Extensibility;
 using Sentry.Reflection;
@@ -14,6 +15,8 @@ namespace Sentry.Internal
         internal const string CurrentUiCultureKey = "Current UI Culture";
         internal const string MemoryInfoKey = "Memory Info";
         internal const string ThreadPoolInfoKey = "ThreadPool Info";
+        internal const string IsDynamicCodeCompiledKey = "Is Dynamic Code Compiled";
+        internal const string IsDynamicCodeSupportedKey = "Is Dynamic Code Supported";
 
         private readonly Enricher _enricher;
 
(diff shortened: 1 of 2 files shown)


src/Sentry/Internal/MainSentryEventProcessor.cs
```diff
@@ -56,6 +59,11 @@ public SentryEvent Process(SentryEvent @event)
                 @event.Contexts[CurrentUiCultureKey] = currentUiCultureMap;
             }
 
+#if NETCOREAPP3_0_OR_GREATER
+            @event.Contexts[IsDynamicCodeCompiledKey] = RuntimeFeature.IsDynamicCodeCompiled;
+            @event.Contexts[IsDynamicCodeSupportedKey] = RuntimeFeature.IsDynamicCodeSupported;
```
lucas-zimerman, line 64: image
This will not work. And I understand that the context struct makes you believe it accepts an object as a parameter but in reality, it expects a dictionary 😅

I would suggest the following change to your code in order to attach the new data.

```suggestion
            @event.Contexts[IsDynamicCodeKey] = new Dictionary<string, bool>
            {
                { IsDynamicCodeCompiledKey, RuntimeFeature.IsDynamicCodeCompiled },
                { IsDynamicCodeSupportedKey, RuntimeFeature.IsDynamicCodeSupported }
            };
```

image

src/Sentry/Internal/MainSentryEventProcessor.cs
```diff
@@ -14,6 +15,8 @@ internal class MainSentryEventProcessor : ISentryEventProcessor
         internal const string CurrentUiCultureKey = "Current UI Culture";
         internal const string MemoryInfoKey = "Memory Info";
         internal const string ThreadPoolInfoKey = "ThreadPool Info";
+        internal const string IsDynamicCodeCompiledKey = "Is Dynamic Code Compiled";
+        internal const string IsDynamicCodeSupportedKey = "Is Dynamic Code Supported";
```
lucas-zimerman, line 19:
```suggestion
        internal const string IsDynamicCodeKey = "Dynamic Code";
        internal const string IsDynamicCodeCompiledKey = "Compiled";
        internal const string IsDynamicCodeSupportedKey = "Supported";
```

lucas-zimerman (changes requested):

src/Sentry/Internal/MainSentryEventProcessor.cs
SimonCropp: @lucas-zimerman thanks

Issue #1373, opened 2021-12-07
Provide a way for users to define a Transaction name (Avoid `Unknown Route `)


bruno-garcia:
Currently, unless the route is known via endpoint routing or MVC, the SDK creates transactions named `Unknown Route`. This is suboptiomal as there are many cases where the route should just. be the path (i.e: Static files).

The problem with defaulting to that is that it's possible some middleware is handling things dynamically like `/something/GUID` and those would break the grouping of transactions (it should be `/something/{id}` as a transaction name instead of a new transaction for each GUID).

We need to provide an API for users to define what should be defined as a transaction and what should fallback to Unknown Route.

lucas-zimerman:
My suggestion would be to alter the interface https://github.com/getsentry/sentry-dotnet/blob/main/src/Sentry.AspNetCore/ISentryRouteName.cs
to

```csharp
namespace Sentry.AspNetCore;

public interface ISentryRouteName
{
    string? GetRouteName(HttpContext context);
}
```

Then add an extension to `SentryAspNetCoreOptions` where the user is able to register his class.
Lastly, we could check here https://github.com/getsentry/sentry-dotnet/blob/41cf450f3000f7a2a0311d85011a4ca5bbacd56a/src/Sentry.AspNetCore/SentryTracingMiddleware.cs#L119 if the user registered the extension and invoke the following code:
```csharp
    httpContext.Features.Set<ISentryRouteName>(UserSentryRouteName);
```

bruno-garcia:
@lucas-zimerman not sure I follow. How much effort would it be to have a PoC for this so we can have an idea?
This keeps coming up. And even on my experience with nugettrends.com it's rather annoying to have those when I'd like to see the static files it served, for example.

Additionally, how would we support other frameworks that themselves would have routes? Like servicestack

mattjohnsonpint:
Completed in #1421

PR #1421, opened 2022-01-04
Add TransactionNameProvider to allow the name definition from Unknown transactions on ASP.NET Core

lucas-zimerman:
- The idea is to allow the user to set the transaction name on cases where the SDK isn't able to define the current Transaction name.

- I'm also altering the GCP Integration since it already uses the integration, but without relying on the HttpContext to set the name.

Solves #1373.


src/Sentry.AspNetCore/SentryAspNetCoreOptions.cs
```diff
@@ -43,6 +43,12 @@ public class SentryAspNetCoreOptions : SentryLoggingOptions
     /// </summary>
     public TimeSpan FlushTimeout { get; set; } = TimeSpan.FromSeconds(2);
 
+    /// <summary>
+    /// RouteName sets the transaction name that was created by the ASPNetCore integration
+    /// when the integration isn't able to define the transaction name..
+    /// </summary>
+    public ISentryRouteName? RouteName { get; set; }
```
bruno-garcia, line 50: Something that better reflects what this options does:

```suggestion
    public ITransactionNameProvider? TransactionNameProvider { get; set; }
```

src/Sentry.AspNetCore/SentryAspNetCoreOptions.cs
```diff
@@ -43,6 +43,12 @@ public class SentryAspNetCoreOptions : SentryLoggingOptions
     /// </summary>
     public TimeSpan FlushTimeout { get; set; } = TimeSpan.FromSeconds(2);
 
+    /// <summary>
+    /// RouteName sets the transaction name that was created by the ASPNetCore integration
+    /// when the integration isn't able to define the transaction name..
+    /// </summary>
```
bruno-garcia, line 49:
```suggestion
    /// <summary>
    /// The strategy to define the name of a transaction based on the HttpContext
    /// </summary>
    /// <remarks >
    /// The SDK can name transactions automatically when using MVC or Endpoint Routing. In other cases, like when serving static files, it fallback to Unknown Route. This hook allows custom code to define a transaction name given a HttpContext.
    /// </remarks>
```

src/Sentry.AspNetCore/SentryTracingMiddleware.cs
```diff
@@ -110,6 +110,11 @@ public async Task InvokeAsync(HttpContext context)
             var transaction = TryStartTransaction(context);
             var initialName = transaction?.Name;
 
+            if (_options.RouteName is { } route)
```
bruno-garcia, line 113: this needs to happen earlier

SimonCropp (approved):
LGTM. i pushed some very small cosmetic changes

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 ## Features
 
+- Add ITransactionNameProvider to allow the name definition from Unknown transactions on ASP.NET Core ([#1421](https://github.com/getsentry/sentry-dotnet/pull/1421))
```
bruno-garcia, line 7: I guess this needs to be reworded now

src/Sentry.AspNetCore/ITransactionNameProvider.cs
```diff
@@ -0,0 +1,11 @@
+using Microsoft.AspNetCore.Http;
```
lucas-zimerman, line 1: Filename needs to be renamed.

SimonCropp:
do we need a corresponding docs PR?

lucas-zimerman:
> do we need a corresponding docs PR?

We deffo need some docs but I believe we could work on it once we have a released version with this PR bundled.

bruno-garcia (approved):

PR #1432, opened 2022-01-11
create CachingTransport dirs in constructor

### LINKED ISSUE #1429: Perform a Dir exists check in CachingTransport.GetCacheFilePaths
image

SimonCropp:
fixes #1429

### DIFF
--- a/src/Sentry/Internal/Http/CachingTransport.cs
+++ b/src/Sentry/Internal/Http/CachingTransport.cs
@@ -2,7 +2,6 @@ using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
-using System.Net;
 using System.Net.Http;
 using System.Net.Sockets;
 using System.Threading;
@@ -52,7 +51,11 @@ namespace Sentry.Internal.Http
                 options.TryGetProcessSpecificCacheDirectoryPath() ??
                 throw new InvalidOperationException("Cache directory or DSN is not set.");
 
+            Directory.CreateDirectory(_isolatedCacheDirectoryPath);
             _processingDirectoryPath = Path.Combine(_isolatedCacheDirectoryPath, "__processing");
+
+            // Ensure that the processing directory exists
+            Directory.CreateDirectory(_processingDirectoryPath);
 
             _worker = Task.Run(CachedTransportBackgroundTaskAsync);
         }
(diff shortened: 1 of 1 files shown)


src/Sentry/Internal/Http/CachingTransport.cs
```diff
@@ -143,17 +146,9 @@ private void EnsureFreeSpaceInCache()
...
-                return Directory
-                    .EnumerateFiles(_isolatedCacheDirectoryPath, $"*.{EnvelopeFileExt}")
-                    .OrderBy(f => new FileInfo(f).CreationTimeUtc);
-            }
-            catch (DirectoryNotFoundException)
-            {
-                _options.LogWarning("Cache directory is empty.");
-                return Array.Empty<string>();
-            }
+            return Directory
```
bruno-garcia, line 149: I think it's best we keep the try catch. In the event the sir is deleted from underneath us we don't bubble out of here and possibly stop the worker thread. Should we recreate it in that case?

src/Sentry/Internal/Http/CachingTransport.cs
```diff
@@ -310,13 +301,13 @@ await using (stream.ConfigureAwait(false))
...
         // This method asynchronously blocks until the envelope is written to cache, but not until it's sent
-        public async Task SendEnvelopeAsync(
+        public Task SendEnvelopeAsync(
             Envelope envelope,
             CancellationToken cancellationToken = default)
         {
             // Store the envelope in a file without actually sending it anywhere.
             // The envelope will get picked up by the background thread eventually.
-            await StoreToCacheAsync(envelope, cancellationToken).ConfigureAwait(false);
+            return StoreToCacheAsync(envelope, cancellationToken);
```
bruno-garcia, line 310: It's best to await so we can debug this better by having this line in the stacktrace. Optimizing here doesn't pay off

src/Sentry/Internal/Http/CachingTransport.cs
lucas-zimerman, line 149: I'd suggest going  with 
```csharp
        private IEnumerable<string> GetCacheFilePaths()
            => Directory
                  .EnumerateFiles(_isolatedCacheDirectoryPath, $"*.{EnvelopeFileExt}")
                  .OrderBy(f => new FileInfo(f).CreationTimeUtc);
```
Since it's a "single line" function.

bruno-garcia (approved):

PR #1444, opened 2022-01-20
Use delegate for transaction name

SimonCropp:
what are peoples thoughts on using a delegate instead of an interface? it is less code for consumers and also more versatile since it can be defined inline


lucas-zimerman (approved):

bruno-garcia:
image

PR #1447, opened 2022-01-25
UnobservedTaskException is unhandled

bruno-garcia:


bruno-garcia:
just changed it through GitHub, still need a test to avoid regression

src/Sentry/Integrations/TaskUnobservedTaskExceptionIntegration.cs
```diff
@@ -35,6 +35,7 @@ internal void Handle(object? sender, UnobservedTaskExceptionEventArgs e)
         {
             if (e.Exception != null)
             {
+                e.Data[Mechanism.HandledKey] = false;
```
lucas-zimerman, line 38: What about
```suggestion
                e.Exception.Data[Mechanism.HandledKey] = false;
```

lucas-zimerman: Also what about using e.Observed as the value for `e.Exception.Data[Mechanism.HandledKey]` ?

https://docs.microsoft.com/en-us/dotnet/api/system.threading.tasks.unobservedtaskexceptioneventargs.observed?view=net-6.0#System_Threading_Tasks_UnobservedTaskExceptionEventArgs_Observed

lucas-zimerman (approved):

PR #1460, opened 2022-02-04
fix flaky test on .NET 461

lucas-zimerman:
`var persistedSessionUpdate = sut.TryRecoverPersistedSession();` could run in less than a tick, the fix fixes the time flow, by not rellying on UtcNow.

#skip-changelog.

### DIFF
--- a/test/Sentry.Tests/GlobalSessionManagerTests.cs
+++ b/test/Sentry.Tests/GlobalSessionManagerTests.cs
@@ -311,7 +311,11 @@ public class GlobalSessionManagerTests
         });
         persistedSessionUpdate!.IsInitial.Should().BeFalse();
         persistedSessionUpdate!.Timestamp.Should().BeAfter(sessionUpdate!.Timestamp);
-        persistedSessionUpdate!.Duration.Should().BeGreaterThan(sessionUpdate!.Duration);
+        persistedSessionUpdate!.Duration.Should().BeGreaterThan(sessionUpdate!.Duration,
+            "Ok what broke: duration ticks is even? {0}, what about {1} and {2}",
+            persistedSessionUpdate!.Duration.Ticks == sessionUpdate!.Duration.Ticks,
+            persistedSessionUpdate!.Duration,
+            sessionUpdate!.Duration);
         persistedSessionUpdate!.SequenceNumber.Should().Be(sessionUpdate!.SequenceNumber + 1);
     }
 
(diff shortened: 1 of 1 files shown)


test/Sentry.Tests/GlobalSessionManagerTests.cs
```diff
@@ -311,7 +311,11 @@ public void TryGetPersistentInstallationId_SessionStarted_ReturnsLastSession()
         });
         persistedSessionUpdate!.IsInitial.Should().BeFalse();
         persistedSessionUpdate!.Timestamp.Should().BeAfter(sessionUpdate!.Timestamp);
-        persistedSessionUpdate!.Duration.Should().BeGreaterThan(sessionUpdate!.Duration);
+        persistedSessionUpdate!.Duration.Should().BeGreaterThan(sessionUpdate!.Duration,
```
SimonCropp, line 314: this will be more readable if u assign the two Durations to a local variables

test/Sentry.Tests/GlobalSessionManagerTests.cs
```diff
@@ -292,12 +294,13 @@ public void TryGetPersistentInstallationId_SessionStarted_ReturnsLastSession()
         var sut = _fixture.GetSut();
 
         var sessionUpdate = sut.StartSession();
-
+        _fixture.Clock.GetUtcNow().Returns(DateTimeOffset.Now.AddSeconds(1));
```
bruno-garcia, line 297: It will still be broken if it runs with 0 ticks, as both calls would return the same value (whatever + 1)

bruno-garcia: keep the counter in scope, and `++` on it so each call to the method increments the value

bruno-garcia (approved):

test/Sentry.Tests/GlobalSessionManagerTests.cs
```diff
@@ -291,13 +293,21 @@ public void TryGetPersistentInstallationId_SessionStarted_ReturnsLastSession()
...
+        });
+
         var sessionUpdate = sut.StartSession();
 
         // Act
         var persistedSessionUpdate = sut.TryRecoverPersistedSession();
 
         // Assert
         sessionUpdate.Should().NotBeNull();
+        persistedSessionUpdate!.EndStatus.Should().Be(SessionEndStatus.Abnormal);
```
SimonCropp, line 310: perhaps move the `!` up to `var persistedSessionUpdate = sut.TryRecoverPersistedSession()!;`. it better represents the intent?

SimonCropp (approved):

PR #1472, opened 2022-02-06
Rename thread pool protocol keys to snake case

### LINKED ISSUE #1471: Rename thread pool protocol keys to snake case
Rename thread pool keys to snake case:

image

The Sentry protocol is snake case: https://develop.sentry.dev/sdk/event-payloads/
Please document this interface there so we can reuse this in other SDKs and add the proper titles in the Sentry front end

Documenting it and fixing it up on the front end allows us to have something like: 
image

And: 
image

On the protocol doc please clarify what the unit is (I believe all of those are 'counts'?) But in the event of bytes/dates etc, the  UI can display it nicely.

SimonCropp:
fixes #1471

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@
 - Reduced the logger noise from EF when not using Performance Monitoring ([#1441](https://github.com/getsentry/sentry-dotnet/pull/1441))
 - Create CachingTransport directories in constructor to avoid DirectoryNotFoundException ([#1432](https://github.com/getsentry/sentry-dotnet/pull/1432))
 - UnobservedTaskException is now considered as Unhandled ([#1447](https://github.com/getsentry/sentry-dotnet/pull/1447))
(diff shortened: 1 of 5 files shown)


SimonCropp:
@bruno-garcia so should these be snake case as well?

```
    internal class MainSentryEventProcessor : ISentryEventProcessor
    {
        internal const string CultureInfoKey = "Current Culture";
        internal const string CurrentUiCultureKey = "Current UI Culture";
        internal const string MemoryInfoKey = "Memory Info";
        internal const string ThreadPoolInfoKey = "ThreadPool Info";
        internal const string IsDynamicCodeKey = "Dynamic Code";
        internal const string IsDynamicCodeCompiledKey = "Compiled";
        internal const string IsDynamicCodeSupportedKey = "Supported";
```

bruno-garcia:
Either the way we want to show in the UI if jot partnof thr protocol, or snake case and added to develop docs and changes in sentry UI to render nicely.

The "UI" culture is a known protocol I believe, just the key there is different because we add two instances.

bruno-garcia (approved):
Should we have some snapshot or other type of tests to validate this?

PR #1482, opened 2022-02-10
fix dotnet format

bruno-garcia:
#skip-changelog


bruno-garcia:
@SimonCropp verified it locally and works fine. I mixed things up when I was testing here

PR #1493, opened 2022-02-22
expose IsSystemModuleName as a public static function.

lucas-zimerman:
#skip-changelog.

I don' believe this function belongs to SentryOptionsExtensions since it's not exactly an extension to SentryOptions so leaving it on SentryStackTraceFactory may be thet best place from what I say.

The goal of this PR is to expose IsSystemModuleName so a proper fix can be implemented for https://github.com/getsentry/sentry-unity/issues/575

### DIFF
--- a/src/Sentry/Exceptions/SentryStackFrame.cs
+++ b/src/Sentry/Exceptions/SentryStackFrame.cs
@@ -170,16 +170,18 @@ namespace Sentry
         {
             var parameterName = Module ?? Function;
             if (InApp != null)
-            { }
-            else if (string.IsNullOrEmpty(parameterName))
+            {
+                return;
+            }
+
+            if (string.IsNullOrEmpty(parameterName))
             {
                 InApp = true;
+                return;
             }
-            else
-            {
-                InApp = options.InAppInclude?.Any(include => parameterName.StartsWith(include, StringComparison.Ordinal)) == true ||
-                        options.InAppExclude?.Any(exclude => parameterName.StartsWith(exclude, StringComparison.Ordinal)) != true;
-            }
+
+            InApp = options.InAppInclude?.Any(include => parameterName.StartsWith(include, StringComparison.Ordinal)) == true ||
+                    options.InAppExclude?.Any(exclude => parameterName.StartsWith(exclude, StringComparison.Ordinal)) != true;
         }
 
         /// <summary>
--- a/src/Sentry/Exceptions/SentryStackFrame.cs
+++ b/src/Sentry/Exceptions/SentryStackFrame.cs
@@ -162,7 +162,7 @@ namespace Sentry
         }
 
         /// <summary>
-        /// Configures <see cref="InApp"/> based on the InApp defined on SentryOptions.
+        /// Configures <see cref="InApp"/> based on the <see cref="SentryOptions.InAppInclude"/> and <see cref="SentryOptions.InAppExclude"/> or <paramref name="options"/>.
         /// </summary>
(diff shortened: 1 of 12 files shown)


src/Sentry/Extensibility/SentryStackTraceFactory.cs
```diff
@@ -211,15 +211,21 @@ protected SentryStackFrame InternalCreateFrame(StackFrame stackFrame, bool deman
...
             => stackFrame.GetMethod();
 
-        private bool IsSystemModuleName(string? moduleName)
+        /// <summary>
+        /// Inform if the given module name belongs to the System or not.
+        /// </summary>
+        /// <param name="moduleName">The module name to be checked.</param>
+        /// <param name="options">The Sentry options.</param>
+        /// <returns>True if the module belongs to the System, false otherwise.</returns>
+        public static bool IsSystemModuleName(string? moduleName, SentryOptions options)
```
bruno-garcia, line 220: It's also an opportunity to rename this. I believe a more appropriate name is: `IsInApp` and take the `SentryStackFrame` instead. Keep the current implementation but reverse the current result.

Taking `SentryStackFrame` as opposed to `string` also to get some flexibilty in the future to decide InApp on other frame information other than "module"

How does that sound @bitsandfoxes @SimonCropp ?

lucas-zimerman: `frame.InApp ?? = frame.IsInApp(options); ` might sound a bit odd and `IsInApp` can diverge in value from the InApp set on the frame.

What about `frame.ConfigureAppFrame(_options);` where it sets the InApp value only if not previously set?

I'm checking in this case both `Module` and `Function` since  on the .NET SDK we are setting the `Module` parameter and on Unity we set the `Function` parameter.

SimonCropp: in my limited understanding of the conversation, i prefer `frame.ConfigureAppFrame(_options)`

src/Sentry/Exceptions/SentryStackFrame.cs
```diff
@@ -160,6 +161,27 @@ public void WriteTo(Utf8JsonWriter writer, IDiagnosticLogger? logger)
...
+            {
+                InApp = true;
+            }
+            else
+            {
+                InApp = options.InAppInclude?.Any(include => parameterName.StartsWith(include, StringComparison.Ordinal)) == true ||
+                        options.InAppExclude?.Any(exclude => parameterName.StartsWith(exclude, StringComparison.Ordinal)) != true;
+            }
+        }
+
```
SimonCropp, line 184: nitpick

```

        public void ConfigureAppFrame(SentryOptions options)
        {
            var parameterName = Module ?? Function;
            if (InApp != null)
            {
                return;
            }

            if (string.IsNullOrEmpty(parameterName))
            {
                InApp = true;
                return;
            }

            InApp = options.InAppInclude?.Any(include => parameterName.StartsWith(include, StringComparison.Ordinal)) == true ||
                    options.InAppExclude?.Any(exclude => parameterName.StartsWith(exclude, StringComparison.Ordinal)) != true;
        }
        
        
        
        ```

SimonCropp (approved):

SimonCropp:
i pushed a very minor xml doc improvement

lucas-zimerman:
Looks good for me :)

bruno-garcia:
There's a new public API which doesn't match the PR title and has no changelog entry.
Ideally we'd have user-facing changes in the changelog to help folks updating

PR #1510, opened 2022-03-07
Skip failing test on M1

### LINKED ISSUE #1508: Ctor_StartupTimeSimilarToUtcNow failing on Mac M1 on .NET Core
### Environment

MacBook Pro with M1 (Apple Silicon) hardware

### Steps to Reproduce

1. Install the .NET SDKs (6.0, 5.0, 3.1, 2.1)
2. Clone the sentry-dotnet repo
3. Built and test

### Expected Result
Test passes for all runtime versions we target in the test suite.

### Actual Result

Test fails for netcoreapp3.1, netcoreapp3.0, and netcoreapp2.1 *only*.

```
[xUnit.net 00:00:08.77]     Sentry.Tests.Internals.ProcessInfoTests.Ctor_StartupTimeSimilarToUtcNow [FAIL]
  Failed Sentry.Tests.Internals.ProcessInfoTests.Ctor_StartupTimeSimilarToUtcNow [9 ms]
  Error Message:
   diff isn't less expected 240 seconds: 45848.086737
Expected: True
Actual:   False
  Stack Trace:
     at Sentry.Tests.Internals.ProcessInfoTests.Ctor_StartupTimeSimilarToUtcNow() in /Users/matt/Code/sentry-dotnet/test/Sentry.Tests/Internals/ProcessInfoTests.cs:line 33
--- End of stack trace from previous location where exception was thrown ---
```

The diff time is > 12 hours, which is not possible when the test fails in a few seconds.

Initial investigation shows that the value of `System.Diagnostics.Process.GetCurrentProcess().StartTime` is erroneous, and thus our `Sentry.Internal.ProcessInfo.StartupTime` is bad.  This appears to be a regression of https://github.com/dotnet/runtime/issues/19928, which was resolved in .NET Core 2.0.  Perhaps this due to the M1 hardware?  If so, why does it not also fail on .NET 5.0 and 6.0?  Need to dive in deeper to see if something else changed within .NET that would affect this.

Setting `options.DetectStartupTime = StartupTimeDetectionMode.Fast` avoids this path to pass the test, but that just hides the issue.

mattjohnsonpint:
Resolves #1508 by skipping the failing test on M1 hardware for affected .NET runtime versions.

#skip-changelog

### DIFF
--- a/test/Sentry.Tests/Internals/ProcessInfoTests.cs
+++ b/test/Sentry.Tests/Internals/ProcessInfoTests.cs
@@ -1,10 +1,23 @@
+using System.Runtime.InteropServices;
+
 namespace Sentry.Tests.Internals;
 
 public class ProcessInfoTests
 {
-    [Fact]
+    [SkippableFact]
     public async Task Ctor_StartupTimeSimilarToUtcNow()
     {
+
+#if NETCOREAPP
+        // Skip this test when running on macOS with Apple Silicon (M1) hardware due to a bug in .NET with Process.StartTime.
+        // See https://github.com/getsentry/sentry-dotnet/issues/1508
+        // and https://github.com/dotnet/runtime/issues/66170
+        // TODO: When .NET resolves the issue, change the version number below to reflect the fixed version of 3.1.xxx
+        Skip.If(RuntimeInformation.IsOSPlatform(OSPlatform.OSX) &&
+                RuntimeInformation.OSDescription.Contains("ARM64") &&
+                Environment.Version < Version.Parse("5.0.0"));
+#endif
+
         //Arrange
         var options = new SentryOptions();
 


test/Sentry.Tests/Internals/ProcessInfoTests.cs
```diff
@@ -1,10 +1,23 @@
+using System.Runtime.InteropServices;
+
 namespace Sentry.Tests.Internals;
 
 public class ProcessInfoTests
 {
-    [Fact]
+    [SkippableFact]
```
bruno-garcia, line 7: What about also:

```suggestion
#if NETCOREAPP
    [SkippableFact]
#endif
```

bruno-garcia (approved):

test/Sentry.Tests/Internals/ProcessInfoTests.cs
mattjohnsonpint: Don't think that should matter all that much really.

PR #1513, opened 2022-03-08
feat: Add new Protocol for DebugImages and AddressMode

Swatinem:


bruno-garcia:
@SimonCropp is this PR just missing the verify snapshots?
@Swatinem is from the native team, could you please give him a hand?

SimonCropp:
@Swatinem sorry for the hassle. You can read about Verify here https://github.com/VerifyTests/Verify

For now i have accepted the changes in the pending snaphots.

 those api changes are configured to [autoverify](https://github.com/VerifyTests/Verify/blob/main/docs/verify-options.md#autoverify). so if you run all the tests, you should notice those files being updated in your git tooling.
 
 As for getting you setup and more familiar with snapshot testing. lets chat on discord

src/Sentry/Exceptions/SentryDebugImage.cs
```diff
@@ -0,0 +1,94 @@
...
+using System.Text.Json;
+using Sentry.Extensibility;
+using Sentry.Internal.Extensions;
+
+namespace Sentry
+{
+    /// <summary>
+    /// The Sentry Debug Meta Images interface.
+    /// </summary>
+    /// <see href="https://develop.sentry.dev/sdk/event-payloads/debugmeta#debug-images"/>
```
SimonCropp, line 12: this is not a valid xml doc. should it be inside a `<remarks>`  ?

src/Sentry/SentryEvent.cs
```diff
@@ -96,6 +96,12 @@ public sealed class SentryEvent : IEventLike, IJsonSerializable
             set => SentryThreadValues = value != null ? new SentryValues<SentryThread>(value) : null;
         }
 
+        /// <summary>
+        /// The Sentry Debug Meta Images interface.
+        /// </summary>
+        /// <see href="https://develop.sentry.dev/sdk/event-payloads/debugmeta#debug-images"/>
+        public List<SentryDebugImage>? SentryDebugImages { get; set; }
```
bruno-garcia, line 103: Property names don't need to be prefixed with `Sentry`:

```suggestion
        public List<SentryDebugImage>? DebugImages { get; set; }
```

src/Sentry/Exceptions/SentryDebugImage.cs
```diff
@@ -0,0 +1,94 @@
+using System.Collections.Generic;
+using System.Linq;
+using System.Text.Json;
+using Sentry.Extensibility;
+using Sentry.Internal.Extensions;
+
+namespace Sentry
```
bruno-garcia, line 7: The file should be on the root of the project (where this namespace represents).

src/Sentry/Exceptions/SentryDebugImage.cs
```diff
@@ -0,0 +1,94 @@
...
+{
+    /// <summary>
+    /// The Sentry Debug Meta Images interface.
+    /// </summary>
+    /// <see href="https://develop.sentry.dev/sdk/event-payloads/debugmeta#debug-images"/>
+    public sealed class SentryDebugImage : IJsonSerializable
```
bruno-garcia, line 13: Unlikely that `DebugImage` would conflict with default types (when `ImplicitUsings` is used) I'd say we can go with `DebugImage`:

```suggestion
    public sealed class DebugImage : IJsonSerializable
```

src/Sentry/SentryEvent.cs
```diff
@@ -242,6 +248,16 @@ public void WriteTo(Utf8JsonWriter writer, IDiagnosticLogger? logger)
             writer.WriteDictionaryIfNotEmpty("extra", _extra, logger);
             writer.WriteStringDictionaryIfNotEmpty("tags", _tags!);
 
+            if (SentryDebugImages != null && SentryDebugImages.Count > 0)
```
bruno-garcia, line 251: Does it work?

```suggestion
            if (SentryDebugImages?.Count > 0)
```

src/Sentry/SentryEvent.cs
Swatinem: Sure thing. I was following the `public IEnumerable<SentryThread>? SentryThreads` example. But maybe those are named that way to avoid conflicts with stdlib types probably?

src/Sentry/DebugImage.cs
```diff
@@ -0,0 +1,94 @@
...
+        /// Memory address, at which the image is mounted in the virtual address space of the process.
+        /// Should be a string in hex representation prefixed with "0x".
+        /// </summary>
+        public string? ImageAddress { get; set; }
+
+        /// <summary>
+        /// The size of the image in virtual memory.
+        /// If missing, Sentry will assume that the image spans up to the next image, which might lead to invalid stack traces.
+        /// </summary>
+        public long ImageSize { get; set; }
```
SimonCropp, line 30: should this be nullable?

SimonCropp: since below `writer.WriteNumberIfNotNull("image_size", ImageSize);` implies it can be null

src/Sentry/Exceptions/SentryStackFrame.cs
```diff
@@ -134,6 +134,14 @@ public sealed class SentryStackFrame : IJsonSerializable
         /// </remarks>
         public long? InstructionOffset { get; set; }
 
+        /// <summary>
+        /// Optionally changes the addressing mode. The default value is the same as
+        /// `"abs"` which means absolute referencing. This can also be set to
+        /// `"rel:DEBUG_ID"` or `"rel:IMAGE_INDEX"` to make addresses relative to an
+        /// object referenced by debug id or index.
+        /// </summary>
+        public string? AddressMode { get; set; }
```
SimonCropp, line 143: seems there is no real usage of AddressMode. should the be a test that uses it?

src/Sentry/SentryEvent.cs
Swatinem: > For the [comparison operators](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/operators/comparison-operators) <, >, <=, and >=, if one or both operands are null, the result is false.

That is good to know, I wasn’t sure about the nullability rules in C#.

Swatinem:
```
[xUnit.net 00:00:17.03]     Sentry.Tests.TaskUnobservedTaskExceptionIntegrationTests.Handle_UnobservedTaskException_CaptureEvent [FAIL]
Error: Assert.True() Failure
Expected: True
Actual:   False
  Failed Sentry.Tests.TaskUnobservedTaskExceptionIntegrationTests.Handle_UnobservedTaskException_CaptureEvent [4 s]
  Error Message:
   Assert.True() Failure
Expected: True
Actual:   False
  Stack Trace:
    at Sentry.Tests.TaskUnobservedTaskExceptionIntegrationTests.Handle_UnobservedTaskException_CaptureEvent () [0x000c2] in <0cb8deca1ea949418add5fbe2e94325a>:0 
  at (wrapper managed-to-native) System.Reflection.RuntimeMethodInfo.InternalInvoke(System.Reflection.RuntimeMethodInfo,object,object[],System.Exception&)
  at System.Reflection.RuntimeMethodInfo.Invoke (System.Object obj, System.Reflection.BindingFlags invokeAttr, System.Reflection.Binder binder, System.Object[] parameters, System.Globalization.CultureInfo culture) [0x0006a] in <533173d24dae460899d2b10975534bb0>:0 
```

Seems flaky? It used to pass before the merge.

src/Sentry/Exceptions/SentryStackFrame.cs
bruno-garcia: test added

bruno-garcia (approved):

PR #1522, opened 2022-03-14
exclude system and Microsoft from coverage

SimonCropp:


mattjohnsonpint (approved):
Seems fine, but why is it necessary?

PR #1531, opened 2022-03-16
Ignore zero properties for MemoryInfo

SimonCropp:
WriteNumberIfNotZero since on OS that dont implement all props, thos props are stubbed out to zero
unknown

### DIFF
--- a/src/Sentry/Internal/MemoryInfo.cs
+++ b/src/Sentry/Internal/MemoryInfo.cs
@@ -3,6 +3,7 @@
 using System;
 using System.Text.Json;
 using Sentry.Extensibility;
+using Sentry.Internal.Extensions;
 
 namespace Sentry
 {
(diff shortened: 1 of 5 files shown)


src/Sentry/Internal/MemoryInfo.cs
```diff
@@ -81,23 +82,24 @@ internal sealed class MemoryInfo : IJsonSerializable
...
-            writer.WriteNumber("pause_time_percentage", PauseTimePercentage);
-            writer.WriteNumber("index", Index);
-            writer.WriteNumber("generation", Generation);
-            writer.WriteNumber("finalization_pending_count", FinalizationPendingCount);
+            writer.WriteNumberIfNotZero("total_committed_bytes", TotalCommittedBytes);
+            writer.WriteNumberIfNotZero("promoted_bytes", PromotedBytes);
+            writer.WriteNumberIfNotZero("pinned_objects_count", PinnedObjectsCount);
+            writer.WriteNumberIfNotZero("pause_time_percentage", PauseTimePercentage);
+            writer.WriteNumberIfNotZero("index", Index);
+            writer.WriteNumberIfNotZero("generation", Generation);
```
bruno-garcia, line 101: Gen0 would be `0` here? What does `generation` usually return?

bruno-garcia: Actually does it make sense to add generation at all? https://docs.microsoft.com/en-us/dotnet/api/system.gcmemoryinfo.generation?view=net-6.0#system-gcmemoryinfo-generation

We're not triggering GC collection here, right? We just want GC info and this doesn't seem to be relevant/helpful

src/Sentry/Internal/MemoryInfo.cs
```diff
@@ -81,23 +82,24 @@ internal sealed class MemoryInfo : IJsonSerializable
...
+            writer.WriteNumberIfNotZero("finalization_pending_count", FinalizationPendingCount);
```
bruno-garcia, line 102: Would 0 be a valid value here? https://docs.microsoft.com/en-us/dotnet/api/system.gcmemoryinfo.finalizationpendingcount?view=net-6.0#system-gcmemoryinfo-finalizationpendingcount

bruno-garcia (commented):
I wonder if there's some API we can call that will tell us whether `GetGCMemoryInfo` returns meaningful data or not. I mean: When do we get all zeros/defaults?

src/Sentry/Internal/MemoryInfo.cs
SimonCropp: happy to remove Generation. but do u want that in diff PR with a diff release note?

SimonCropp:
@bruno-garcia 

> I wonder if there's some API we can call that will tell us whether GetGCMemoryInfo returns meaningful data or not. I mean: When do we get all zeros/defaults?

not from my research

mattjohnsonpint (approved):

PR #1534, opened 2022-03-17
ignore vs code

SimonCropp:
#skip-changelog


mattjohnsonpint:
There is *some* value in sharing VS Code settings, though not everyone uses the same settings.  I could go either way with this.  @bruno-garcia?

SimonCropp:
@mattjohnsonpint agreed. but in my experience it is not possible to distinguish between what is shared and what is custom. so to have any shared settings u need to give up any customisation

Issue #1443, opened 2022-01-20
Introduce a default Retry-After value if one is not present


bruno-garcia:
The SDK doesn't have a default value in case the response headers don't include a value for Retry-After:

https://github.com/getsentry/sentry-dotnet/blob/a9304a0a4b4702d0e62e2703d55c66483d27a0e5/src/Sentry/Internal/Http/RetryAfterHandler.cs#L80-L87

See: https://github.com/getsentry/develop/pull/492

bruno-garcia:
Update notion once this is done:

https://www.notion.so/sentry/Add-rate-limit-default-to-NET-fe4aa210986e475c952b2924640c4204

mattjohnsonpint:
Done.

bruno-garcia:
Nice! Thank you Matt!

PR #1537, opened 2022-03-18
Add default Retry-After value

mattjohnsonpint:
Completes #1443 

Refactored the logic for better readability as well.

### DIFF
--- a/src/Sentry/Internal/Http/RetryAfterHandler.cs
+++ b/src/Sentry/Internal/Http/RetryAfterHandler.cs
@@ -19,6 +19,7 @@ namespace Sentry.Internal.Http
         private readonly ISystemClock _clock;
 
         private const HttpStatusCode TooManyRequests = (HttpStatusCode)429;
+        internal static readonly TimeSpan DefaultRetryAfterDelay = TimeSpan.FromSeconds(60);
 
         private long _retryAfterUtcTicks;
         internal long RetryAfterUtcTicks => _retryAfterUtcTicks;
@@ -84,6 +85,11 @@ namespace Sentry.Internal.Http
                 {
                     var retryAfterSpan = TimeSpan.FromSeconds(retryAfterSeconds);
                     _ = Interlocked.Exchange(ref _retryAfterUtcTicks, _clock.GetUtcNow().AddTicks(retryAfterSpan.Ticks).UtcTicks);
+                }
+                else
+                {
+                    // No retry header was sent. Use the default retry delay.
+                    _ = Interlocked.Exchange(ref _retryAfterUtcTicks, _clock.GetUtcNow().Add(DefaultRetryAfterDelay).UtcTicks);
                 }
             }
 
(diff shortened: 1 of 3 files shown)


src/Sentry/Internal/Http/RetryAfterHandler.cs
```diff
@@ -19,6 +19,7 @@ internal class RetryAfterHandler : DelegatingHandler
         private readonly ISystemClock _clock;
 
         private const HttpStatusCode TooManyRequests = (HttpStatusCode)429;
+        internal static readonly TimeSpan DefaultRetryAfterDelay = TimeSpan.FromSeconds(60);
```
SimonCropp, line 22: `TimeSpan.FromMinutes(1)`

src/Sentry/Internal/Http/RetryAfterHandler.cs
```diff
@@ -63,31 +64,43 @@ internal RetryAfterHandler(HttpMessageHandler innerHandler, ISystemClock clock)
...
+                var retryAfterTimestamp = GetRetryAfterTimestamp(response);
+                _ = Interlocked.Exchange(ref _retryAfterUtcTicks, retryAfterTimestamp.UtcTicks);
+            }
+
+            return response;
+        }
+
+        private DateTimeOffset GetRetryAfterTimestamp(HttpResponseMessage response)
+        {
+            if (response.Headers != null)
```
SimonCropp, line 78: the code for header is 
```
public HttpResponseHeaders Headers => _headers ??= new HttpResponseHeaders();
```
so the null check would seem redundant?

src/Sentry/Internal/Http/RetryAfterHandler.cs
```diff
@@ -63,31 +64,43 @@ internal RetryAfterHandler(HttpMessageHandler innerHandler, ISystemClock clock)
...
-                        _ = Interlocked.Exchange(ref _retryAfterUtcTicks, response.Headers.RetryAfter.Date.Value.UtcTicks);
+                        return date;
                     }
                 }
+
                 // Sentry was sending floating point numbers which are not handled by RetryConditionHeaderValue
                 // To be compatible with older versions of sentry on premise: https://github.com/getsentry/sentry/issues/7919
-                else if (response.Headers.TryGetValues("Retry-After", out var values)
-                         && double.TryParse(values?.FirstOrDefault(), out var retryAfterSeconds))
+                if (response.Headers.TryGetValues("Retry-After", out var values)
```
SimonCropp, line 95: TryGetValues has a notnullwhentrue for value. so the following `values?` is redundant?

src/Sentry/Internal/Http/RetryAfterHandler.cs
mattjohnsonpint: Thanks, but I'd rather keep it in terms of seconds, to match the spec.

SimonCropp (approved):

Issue #1516, opened 2022-03-09
intermittent error in CaptureEvent_NonSerializableContextAndOfflineCaching_CapturesEventWithContextKey


SimonCropp:
https://github.com/getsentry/sentry-dotnet/runs/5485978713?check_suite_focus=true#step:6:1235

```
[xUnit.net 00:00:23.71]     NotSentry.Tests.HubTests.CaptureEvent_NonSerializableContextAndOfflineCaching_CapturesEventWithContextKey(offlineCaching: True) [FAIL]
Error: Un-serializable context key should exist
Expected: True
Actual:   False
  Failed NotSentry.Tests.HubTests.CaptureEvent_NonSerializableContextAndOfflineCaching_CapturesEventWithContextKey(offlineCaching: True) [894 ms]
  Error Message:
   Un-serializable context key should exist
Expected: True
Actual:   False
  Stack Trace:
    at NotSentry.Tests.HubTests.CaptureEvent_NonSerializableContextAndOfflineCaching_CapturesEventWithContextKey (System.Boolean offlineCaching) [0x00150] in <f3596d226acd48f8bc36ebc6a5233641>:0 
  at (wrapper managed-to-native) System.Reflection.RuntimeMethodInfo.InternalInvoke(System.Reflection.RuntimeMethodInfo,object,object[],System.Exception&)
  at System.Reflection.RuntimeMethodInfo.Invoke (System.Object obj, System.Reflection.BindingFlags invokeAttr, System.Reflection.Binder binder, System.Object[] parameters, System.Globalization.CultureInfo culture) [0x0006a] in <e068e2227ab74c1bb3d724ebaab0e3ff>:0 
  ```

mattjohnsonpint:
FWIW, I can't repro this locally.  Been running the test in a loop for awhile now, and it passes every time.  I saw it in CI though, reviewing #1515.  Reran the check jobs and that one passes now (but a different one failed).

mattjohnsonpint:
Appears to be on Mono (net4.6.1 target, Ubuntu and macOS only).

mattjohnsonpint:
I can't prove it's responsible, but I see a hard `.Result` instead of async/await here:

https://github.com/getsentry/sentry-dotnet/blob/b26a2fba91e2d6140444ffd906bef01c2cba8e4a/test/Sentry.Tests/HubTests.cs#L316

I'll fix that and check for any others elswhere.

mattjohnsonpint:
[...]

Issue #1438, opened 2022-01-14
Flakey test ShouldNotLogOperationCanceledExceptionWhenIsCancellationRequested


SimonCropp:
https://github.com/getsentry/sentry-dotnet/runs/4810834881?check_suite_focus=true#step:7:1068

```
[xUnit.net 00:00:16.86]     Sentry.Tests.Internals.Http.CachingTransportTests.ShouldNotLogOperationCanceledExceptionWhenIsCancellationRequested [FAIL]
Error: Sentry.Tests.Internals.Http.CachingTransportTests.ShouldNotLogOperationCanceledExceptionWhenIsCancellationRequested: Error or fatal message logged
  Failed Sentry.Tests.Internals.Http.CachingTransportTests.ShouldNotLogOperationCanceledExceptionWhenIsCancellationRequested [4 s]
  Error Message:
   Error or fatal message logged
Expected: False
Actual:   True
  Stack Trace:
    at Sentry.Tests.Internals.Http.CachingTransportTests.ShouldNotLogOperationCanceledExceptionWhenIsCancellationRequested () [0x0021f] in <8e4526e711384650ba6878eae1d68882>:0 
  at Sentry.Tests.Internals.Http.CachingTransportTests.ShouldNotLogOperationCanceledExceptionWhenIsCancellationRequested () [0x002ea] in <8e4526e711384650ba6878eae1d68882>:0 
```

PR #1539, opened 2022-03-18
Use async/await wherever possible in tests

mattjohnsonpint:
This is an attempt to resolve some flaky tests such as #1516, on the wild guess that blocking waits might be part of the reason they fail sometimes.  I updated a few others as well.

#skip-changelog

### DIFF
--- a/test/Sentry.Testing/EnvironmentVariableGuard.cs
+++ b/test/Sentry.Testing/EnvironmentVariableGuard.cs
@@ -3,21 +3,35 @@ namespace Sentry.Testing;
 public static class EnvironmentVariableGuard
 {
     // To allow different xunit collections use of this
-    private static readonly object Lock = new();
+    private static readonly SemaphoreSlim Lock = new(1, 1);
 
     public static void WithVariable(string key, string value, Action action)
     {
-        lock (Lock)
+        Lock.Wait();
+        Environment.SetEnvironmentVariable(key, value, EnvironmentVariableTarget.Process);
+        try
         {
-            Environment.SetEnvironmentVariable(key, value, EnvironmentVariableTarget.Process);
-            try
-            {
-                action();
-            }
-            finally
-            {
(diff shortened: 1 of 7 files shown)


mattjohnsonpint:
I'm going to re-run the CI build jobs several times for this PR to see if I can get it to bomb again.

mattjohnsonpint:
Still failed on #1438 [here](https://github.com/getsentry/sentry-dotnet/runs/5606345458?check_suite_focus=true#step:6:707) on Ubuntu.  I'll check a few more things there...

test/Sentry.Testing/EnvironmentVariableGuard.cs
```diff
@@ -3,21 +3,35 @@ namespace Sentry.Testing;
...
-            }
-            finally
-            {
-                Environment.SetEnvironmentVariable(key, null, EnvironmentVariableTarget.Process);
-            }
+            action();
+        }
+        finally
+        {
+            Environment.SetEnvironmentVariable(key, null, EnvironmentVariableTarget.Process);
```
bruno-garcia, line 18: I honest think if we have flaky tests due to env var, we have to take to be able to lock this env var read/write. Global state like this is often going to hunt us back

bruno-garcia (approved):

test/Sentry.Testing/EnvironmentVariableGuard.cs
mattjohnsonpint: On this one, I just didn't like the blocking `GetAwaiter().GetResult()` in the tests in `IntegrationMockedBackroundWorker`.  So I added `WithVariableAsync`, which uses a semaphore for locking since you can't `lock` around awaits.  The existing non-async was updated just to use the same semaphore.

But yeah, there's probably a way to do this without setting environment variables globally.

SimonCropp (approved):

PR #1541, opened 2022-03-19
remove code coverage from CI

SimonCropp:
since it is causing too much friction with so many failed builds

#skip-changelog


bruno-garcia:
We can use something else to collect coverage?

mattjohnsonpint (approved):

PR #1553, opened 2022-03-23
Make HttpTransport public (for other SDKs)

vaind:
This is required to be able to reuse the logic in `HttpTransport` while implementing a custom `HttpClient`.

Required for sentry-unity

#skip-changelog


SimonCropp:
`SentryOptions` exposes a `SentryHttpClientFactory` property which allows you to use a custom `HttpClient`

```
public interface ISentryHttpClientFactory
{
    /// <summary>
    /// Creates an HttpClient using the specified options.
    /// </summary>
    /// <param name="options">The options.</param>
    /// <returns><see cref="HttpClient"/>.</returns>
    HttpClient Create(SentryOptions options);
}
```

does this meet your requirements?

vaind:
> 

Unfortunately not. I've tried that (see approaches in the linked sentry-unity PR description), but it doesn't work in Unity webgl because the `HttpTransport` still uses `async/await` which requires multithreading and gets stuck, because there's none. What I had to so was override the async methods with unity's goroutine implementation and reuse the logic from the base class.

test/Sentry.Tests/Internals/AccessModifierTests.cs
```diff
@@ -14,7 +14,10 @@ public void TypesInInternalsNamespace_AreNotPublic()
 
         Assert.All(types, type =>
         {
-            Assert.False(type.IsPublic, $"Expected type {type.Name} to be internal.");
+            if (type.Name != "HttpTransport")
```
bruno-garcia, line 17: @mattjohnsonpint should we move this to `Sentry.Http` namespace (or other) or really open the door to have public types under the `Internal` namespace to 'signify' this won't follow semver. 

ASP.NET Core has public types under `Internal` namespaces, I know as I've relied on them without noticying in the past :D

src/Sentry/Internal/Http/HttpTransport.cs
```diff
@@ -143,10 +163,6 @@ public async Task SendEnvelopeAsync(Envelope envelope, CancellationToken cancell
             using var processedEnvelope = ProcessEnvelope(envelope, instant);
             if (processedEnvelope.Items.Count == 0)
             {
-                _options.LogInfo(
```
mattjohnsonpint, line 146: Looks like this log message was removed?

mattjohnsonpint:
I have some concerns.

It seems that while this would make some of the functionality in `HttpTransport` available to be used by `UnityWebRequestTransport` (in the linked Unity PR), *that* class would no longer be accurately implementing the `ITransport` interface.

I can also see that we would then end up with two places to remember to add features and fix things.  For example, I'm currently working on adding client reports (see draft PR #1556), which is adding more business logic to the background worker and the transport.  These wouldn't carry over to the Unity implementation so would have to be implemented there also.

I think rather than exposing `HttpClient` public, we'd likely be better off moving the common functionality to an abstract base class for them both to share, and making *that* public.  At least that way it's not implied that `UnityWebTransport` implements `ITransport`.  I also think we may need something similar for the background workers.  There may be a couple of new interfaces needed as well.

In short, exposing internals smells like we're missing some interfaces or shared base classes.  I'd rather refactor for that now and built on top of a stronger contract.

mattjohnsonpint:
Closing in favor of #1560 instead.

SimonCropp:
@mattjohnsonpint can this branch be deleted?

PR #1564, opened 2022-04-01
Add UseAsyncFileIO option

mattjohnsonpint:
Adds an option to control whether async file i/o is used when reading attachments from disk.

This was hard-coded previously, but created a problem for when adding support for Unity WebGL.

@vaind - I think this should unblock you?

### DIFF
--- a/src/Sentry/FileAttachmentContent.cs
+++ b/src/Sentry/FileAttachmentContent.cs
@@ -14,8 +14,16 @@ namespace Sentry
         /// Creates a new instance of <see cref="FileAttachmentContent"/>.
         /// </summary>
         /// <param name="filePath">The path to the file to attach.</param>
+        public FileAttachmentContent(string filePath) : this(filePath, true)
+        {
+        }
+
+        /// <summary>
+        /// Creates a new instance of <see cref="FileAttachmentContent"/>.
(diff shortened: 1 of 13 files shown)


src/Sentry/FileAttachmentContent.cs
```diff
@@ -8,11 +8,18 @@ namespace Sentry
...
         private readonly string _filePath;
+        private readonly bool _readFileAsynchronously;
 
         /// <summary>
         /// Creates a new instance of <see cref="FileAttachmentContent"/>.
         /// </summary>
-        public FileAttachmentContent(string filePath) => _filePath = filePath;
+        /// <param name="filePath">The path to the file to attach.</param>
+        /// <param name="readFileAsynchronously">Whether to use async file I/O to read the file.</param>
+        public FileAttachmentContent(string filePath, bool readFileAsynchronously = true)
```
SimonCropp, line 18: this is a runtime breaking change. eg if someone consuming a 3rd partly dll that uses this api

mattjohnsonpint: Fixed.  Thanks!

SimonCropp (approved):

PR #1571, opened 2022-04-05
Make Contexts properties more thread-safe

### LINKED ISSUE #1570: `Contexts` properties are not thread-safe
The [`Sentry.Contexts`](https://github.com/getsentry/sentry-dotnet/blob/main/src/Sentry/Protocol/Contexts.cs) object is derived from a `ConcurrentDictionary`, and thus operations against it are intended to be usable by multiple threads simultaneously.  Indeed, the object itself does that just fine.

However, in many places where we *use* that object on an object, we're instantiating it using a lazy initialization approach that is *not* thread safe.  We also provide a setter to replace the object with another instance, which would not be a thread-safe operation either.

For example:

https://github.com/getsentry/sentry-dotnet/blob/107cb5838b2e22144572dee0db271da3846aabe5/src/Sentry/SentryEvent.cs#L125-L132

The `??=` pattern in the getter is not thread-safe, and thus it is possible that two different threads are working with two different instances of a `Contexts` object.  Only one would be written to `_contexts` and the other would be discarded.  This may possibly be the cause of #1516, but could also be losing real data in the wild.

The setter really should not be there at all.  Reset+Copy pattern should be used instead.  Unfortunately, those setters are public.  They should probably be obsoleted where possible, but that is a secondary concern.

For the time being, I'll at least make sure all instances of `Contexts` objects are pre-initialized, and I'll see if I can reduce the number of places where we use the setter ourselves.  There's a few other minor associated changes.

(Thanks @lucas-zimerman for finding this!)

mattjohnsonpint:
Resolves #1570

### DIFF
--- a/test/Sentry.Tests/Protocol/BaseScopeTests.cs
+++ b/test/Sentry.Tests/Protocol/BaseScopeTests.cs
@@ -59,6 +59,7 @@ public class BaseScopeTests
         _sut.Contexts = expected;
 
         Assert.Equal(expected, _sut.Contexts);
+        Assert.NotSame(expected, _sut.Contexts);
     }
 
     [Fact]
--- a/test/Sentry.Tests/Protocol/BaseScopeTests.cs
+++ b/test/Sentry.Tests/Protocol/BaseScopeTests.cs
@@ -51,9 +51,14 @@ public class BaseScopeTests
     [Fact]
     public void Contexts_Settable()
     {
+        _sut.Contexts.App.Name = "Foo";
+
         var expected = new Contexts();
+        expected.App.Name = "Bar";
+
         _sut.Contexts = expected;
-        Assert.Same(expected, _sut.Contexts);
+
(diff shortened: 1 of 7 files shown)


test/Sentry.Tests/Protocol/BaseScopeTests.cs
```diff
@@ -51,9 +51,14 @@ public void Contexts_ByDefault_NotNull()
...
     {
+        _sut.Contexts.App.Name = "Foo";
+
         var expected = new Contexts();
+        expected.App.Name = "Bar";
+
         _sut.Contexts = expected;
-        Assert.Same(expected, _sut.Contexts);
+
+        Assert.Equal(expected, _sut.Contexts);
```
SimonCropp, line 61: should a `Assert.NotSame(expected, _sut.Contexts);` be added here?

SimonCropp (approved):

lucas-zimerman (approved):

PR #1578, opened 2022-04-20
Update Nugets

SimonCropp:
only tests and sample nugets

#skip-changelog


SimonCropp:
there is too much here. i am going to do them individually

PR #1589, opened 2022-04-23
Fix options internal reference to transport

mattjohnsonpint:
While working on #1556, I found out that while the internal property `SentryOptions.Transport` is available, it is null unless manually set by a unit test.  For client reports to work, I'll need it to be populated with the current transport so I can access the transport from `SentryClient`.

I also found that if a transport is set on the options, and a cache path is also set, caching doesn't work because it was just using the provided transport as-is without wrapping it in a `CachingTransport`.

This PR fixes both, and adds tests.  There should be no externally noticeable behavior changed, so #skip-changelog


bruno-garcia (approved):
👍

PR #1602, opened 2022-04-26
Make ITransport and SentryOptions.Transport public

mattjohnsonpint:
In #1560, we created the `HttpTransportBase` class, containing the "guts" of the `HttpTransport` assembled such that others could extend it to create their own transports without loss of features that are wired at the transport.  However, we didn't create an easy way to actually attach a custom transport at that time, because we were primarily focused on getsentry/sentry-unity#657, which was already building it's own implementation of `IBackgroundWorker` where it could connect its own custom transport.

For others, the ideal place to attach a custom transport is on `SentryOptions.Transport`, as we do in unit tests.  This PR makes that property, and the corresponding `ITransport` type public so that can happen.   (It also removes `IFlushableTransport` just for cleanup.  It wasn't adding anything.)

After this is released, the recommendation for custom transports is to implement similarly to the built-in [`HttpTransport`](https://github.com/getsentry/sentry-dotnet/blob/main/src/Sentry/Internal/Http/HttpTransport.cs).

```csharp
public class SomeCustomTransport : HttpTransportBase, ITransport
```

One *could* also just implement `ITransport` without implementing `HttpTransportBase`, but then they would be missing out on any functionality wired up in the transport such as processing event handlers, client reports, etc.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ### Features
 
-- Expose `ITransport` and `SentryOptions.Transport` public, to support using custom transports ([#1602](https://github.com/getsentry/sentry-dotnet/pull/1602)
+- Expose `ITransport` and `SentryOptions.Transport` public, to support using custom transports ([#1602](https://github.com/getsentry/sentry-dotnet/pull/1602))
 
 ### Fixes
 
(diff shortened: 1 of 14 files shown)


bruno-garcia (approved):
Thanks Matt

SimonCropp (approved):

PR #1606, opened 2022-04-27
consolidate sample TFs

SimonCropp:
#skip-changelog

In Samples:

 * Use net6.0 over net5.0/netcore
 * Use net48 over net472


mattjohnsonpint:
I'm ok with standardizing .NET Framework targets to `net48`, but I'm not so sure about the rest.  I believe that for any of the ASP.NET Core projects, upgrading the target framework also changes the underlying version of the `Microsoft.NET.Sdk.Web` used, and there are known to be breaking changes between versions.  At minimum, I'd want to test each one once that it still works.  But probably it would be best to re-create the samples from the latest templates supplied by `dotnet new`.

SimonCropp:
will revisit this on individual samples

PR #1607, opened 2022-04-28
update actions/checkout

SimonCropp:
#skip-changelog


mattjohnsonpint (approved):

mattjohnsonpint:
Actually I think this should have just been `@v3` rather than `@v3.0.2` - so we get the latest on that major.  I've got that and latest for the others in #1609.  Sorry I didn't catch it earlier.

PR #1609, opened 2022-04-28
Update all github actions to latest major

mattjohnsonpint:
Update all the things!

#skip-changelog


mattjohnsonpint:
Note that the main difference between these versions is that they run on Node 16 instead of Node 12.  A few have bugfixes and dependency updates also.

SimonCropp (approved):

Issue #1590, opened 2022-04-25
Remove 'Platform' from scope


bitsandfoxes:
### Problem Statement

SentryEvents get the Platform property already set in the internal Constructor https://github.com/getsentry/sentry-dotnet/blob/def623108063996cd86facb3bef43998cb2b8867/src/Sentry/SentryEvent.cs#L205
and the scope does not override it when it gets applied https://github.com/getsentry/sentry-dotnet/blob/def623108063996cd86facb3bef43998cb2b8867/src/Sentry/Scope.cs#L363
Does this mean setting the Platform when configuring the scope doesn't do anything?

### Solution Brainstorm

We might as well remove the 'Platform' from the scope.

bruno-garcia:
platform should be always `csharp` (even if code was written in VB or F#) so makes sense to get rid of it

mattjohnsonpint:
> platform should be always `csharp`

Yeah, probably it should have been called `dotnet`, but effectively it's the same for this purpose.

Though after reading [the spec for stack traces](https://develop.sentry.dev/sdk/event-payloads/stacktrace/), I believe we still need to let platform be set on a stack trace/frame so that we can override it if we bundle native SDKs in the future.

mattjohnsonpint:
We were wrong about this.  The platform name won't always be `"csharp"`.  With the new Android support, if an event is generated from the embedded Android SDK, its platform will be `"java"`.  This would be evident on events exposed in a `BeforeSend` delegate.   We shouldn't obsolete the property or remove it, as that will erroneously change the platform from `"java"` to `"csharp"` when roundtripping.

I'll revert the change.

bruno-garcia:
The protocol should allow different values to match the use case Matt described. But I believe there's still no need for Platform on the Scope though. So we could remove it from the scope.

PR #1610, opened 2022-04-28
Obsolete platform and always write csharp

SimonCropp:
related to #1590

### DIFF
--- a/src/Sentry/Exceptions/SentryStackFrame.cs
+++ b/src/Sentry/Exceptions/SentryStackFrame.cs
@@ -104,8 +104,7 @@ namespace Sentry
         /// <summary>
         /// This can override the platform for a single frame. Otherwise the platform of the event is assumed.
         /// </summary>
-        [Obsolete("Platform is always csharp, and should not be set by consuming code. This property will be removed in version 4.")]
-        public string? Platform { get; set; } = Constants.Platform;
+        public string? Platform { get; set; }
 
         /// <summary>
         /// Optionally an address of the debug image to reference.
@@ -161,7 +160,7 @@ namespace Sentry
             writer.WriteStringIfNotWhiteSpace("context_line", ContextLine);
             writer.WriteBooleanIfNotNull("in_app", InApp);
             writer.WriteStringIfNotWhiteSpace("package", Package);
-            writer.WriteString("platform", Constants.Platform);
+            writer.WriteStringIfNotWhiteSpace("platform", Platform);
             writer.WriteNumberIfNotNull("image_addr", ImageAddress.NullIfDefault());
             writer.WriteNumberIfNotNull("symbol_addr", SymbolAddress);
             writer.WriteStringIfNotWhiteSpace("instruction_addr", InstructionAddress);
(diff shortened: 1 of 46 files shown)


src/Sentry/Exceptions/SentryStackFrame.cs
```diff
@@ -160,7 +161,7 @@ public void WriteTo(Utf8JsonWriter writer, IDiagnosticLogger? logger)
             writer.WriteStringIfNotWhiteSpace("context_line", ContextLine);
             writer.WriteBooleanIfNotNull("in_app", InApp);
             writer.WriteStringIfNotWhiteSpace("package", Package);
-            writer.WriteStringIfNotWhiteSpace("platform", Platform);
+            writer.WriteString("platform", Constants.Platform);
```
SimonCropp, line 164: @bruno-garcia @mattjohnsonpint if we always write `csharp` for the root event, can `"platform"` be omitted from stackframe and transaction?

src/Sentry/Exceptions/SentryStackFrame.cs
```diff
@@ -104,7 +104,8 @@ public sealed class SentryStackFrame : IJsonSerializable
         /// <summary>
         /// This can override the platform for a single frame. Otherwise the platform of the event is assumed.
         /// </summary>
-        public string? Platform { get; set; }
```
mattjohnsonpint, line 107: This one is different, as it's an override that's only needed when the platform is different than the one in the event.  We shouldn't set it to `Constants.Platform`.

I also don't think we will obsolete it completely, as we'll need it if we bundle the native SDK (as we will likely need to do for MAUI).  Though, I could see setting it `internal` in the future, which would effectively obsolete it.

test/Sentry.Tests/Protocol/Exceptions/SentryStackFrameTests.cs
```diff
@@ -22,7 +22,6 @@ public void SerializeObject_AllPropertiesSetToNonDefault_SerializesValidObject()
             Vars = { { "var", "val" } },
             FramesOmitted = { 1, 2 },
             Package = "Package",
-            Platform = "Platform",
```
mattjohnsonpint, line 25: Lets keep this test as it was, to ensure we can override the platform for a single stack frame.

src/Sentry/Exceptions/SentryStackFrame.cs
mattjohnsonpint: It always needs to be sent to sentry, for all events including both errors and transactions.  But on stack frames, it should only be included if that frame originates from another platform.  So, if we end up bundling the native SDKs at some point, the stack frames captured by those SDKs would be different.  Thus, this line should be reverted to as it was, and we should keep `StackFrame.Platform` null by default.

mattjohnsonpint (changes requested):
Please update stack frame as indicated.  Thanks.

test/Sentry.Tests/Internals/MonoSentryStackTraceFactoryTests.cs
```diff
@@ -33,7 +33,6 @@ public void Create_UnityIl2Cpp_LeavesOutZeroedOutPath()
             Assert.Null(actual!.Frames[i].LineNumber);
             Assert.Null(actual!.Frames[i].InstructionOffset);
             Assert.Null(actual!.Frames[i].Package);
-            Assert.Null(actual!.Frames[i].Platform);
```
mattjohnsonpint, line 36: I think this one needs to be put back?

mattjohnsonpint (approved):

PR #1617, opened 2022-04-29
Update caching transport and tests

mattjohnsonpint:
Some improvements to the caching transport, mostly around tests:

- Most of the tests immediately stop the worker and then flush it manually later.  Add a boolean `startWorker` that can be set `false` on these tests to just not have a worker started to begin with.
- Several tests didn't need a cache worker, and had one previously.
- The `AwareOfExistingFiles` test was passing, but wasn't actually testing what we thought it was.  It does now.
- The `WorksInBackground` test was watching the directory to know when it was finished.  Rewrote to watch an event fired on the inner transport.  No need to watch the directory at all now.
- On the caching transport itself, recovery of files left in processing was being done with the worker start.  Moved this to initialization so that it always happens, even if the worker is shutting down or when there's not a worker.  This mostly is for the test to pass, but is a good idea anyway.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 
 - Workaround `System.Text.Json` issue with Unity IL2CPP. ([#1583](https://github.com/getsentry/sentry-dotnet/pull/1583))
 - Demystify stack traces for exceptions that fire in a `BeforeSend` callback. ([#1587](https://github.com/getsentry/sentry-dotnet/pull/1587))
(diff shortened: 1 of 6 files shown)


mattjohnsonpint:
Looks like there's still a race condition in the `AwareOfExistingFiles` test.  I'll come back to this...

mattjohnsonpint:
It's good at this point, but I see a few more related improvements to make.  More coming shortly. :wink:

mattjohnsonpint:
Good enough for now.  Ready for review.  Thanks.

SimonCropp (approved):

Issue #1014, opened 2021-05-26
Make `Unregister` part of the `ISdkIntegration` interface


bruno-garcia:
It's currently internal, and we'd like to consume that from Unity.

PR #1632, opened 2022-05-03
Changed `RemoveIntegration` access modifier to public

bitsandfoxes:
Resolves https://github.com/getsentry/sentry-dotnet/issues/1014
The Unity SDK adds its own integrations and currently can't remove them again due to the `Integrations` being internal. Making the `RemoveIntegration` public alleviates this problem.

#skip-changelog


SimonCropp:
i have no objections. might be worth seeing if @bruno-garcia has any historical knowledge

SimonCropp (approved):

bruno-garcia (approved):
Fair enough. This was 'designed' to be public so it can be extended. But at the end I left it internal originally and focused on giving dedicated extension methods for each integration built-in.

Issue #1640, opened 2022-05-04
System.AggregateException from Sentry assembly


caiofbpa:
### Package

Sentry

### .NET Flavor

Xamarin

### .NET Version

5.0.0

### OS

Android

### SDK Version

1.4.0

### Steps to Reproduce

1. Install Sentry.Xamarin in our Xamarin.Droid app
2. Serve tens of thousands of users
3. See random crashes coming out of Sentry.Xamarin itself

### Expected Result

No crash should come from Sentry.Xamarin assemblies.

### Actual Result

We're seeing different crashes being reported in Sentry in our Xamarin.Droid app:

### Network error

```
System.AggregateException
A Task's exception(s) were not observed either by Waiting on the Task or accessing its Exception property. As a result, the unobserved exception was rethrown by the finalizer thread. (No such host is known)
```

```
System.Net.Sockets.SocketException: No such host is known
  ?, in async ValueTask<ValueTuple<Socket, Stream>> ConnectHelper.ConnectAsync(string host, int port, CancellationToken cancellationToken)
System.Net.Http.HttpRequestException: No such host is known
  ?, in async ValueTask<ValueTuple<Socket, Stream>> ConnectHelper.ConnectAsync(string host, int port, CancellationToken cancellationToken)
  ?, in async ValueTask<ValueTuple<HttpConnection, HttpResponseMessage>> HttpConnectionPool.CreateConnectionAsync(HttpRequestMessage request, CancellationToken cancellationToken)
  ?, in async ValueTask<ValueTuple<HttpConnection, HttpResponseMessage>> HttpConnectionPool.WaitForCreatedConnectionAsync(ValueTask<ValueTuple<HttpConnection, HttpResponseMessage>> creationTask)
  ?, in async Task<HttpResponseMessage> HttpConnectionPool.SendWithRetryAsync(HttpRequestMessage request, bool doRequestAuth, CancellationToken cancellationToken)
  ?, in async Task<HttpResponseMessage> RedirectHandler.SendAsync(HttpRequestMessage request, CancellationToken cancellationToken)
  ?, in async Task<HttpResponseMessage> DecompressionHandler.SendAsync(HttpRequestMessage request, CancellationToken cancellationToken)
  ?, in async Task<HttpResponseMessage> GzipBufferedRequestBodyHandler.SendAsync(HttpRequestMessage request, CancellationToken cancellationToken)
[...]

PR #1644, opened 2022-05-06
Re-implement InitCacheFlushTimeout option and tests

### LINKED ISSUE #1642: Remove BlockCacheFlush and improve InitCacheFlushTimeout mechanism
The option `InitCacheFlushTimeout` controls how the [`BlockCacheFlush`](https://github.com/getsentry/sentry-dotnet/blob/fc5f464ff0245676613764cf7e45597e151dc6c3/src/Sentry/Internal/SdkComposer.cs#L55) method works.  Unfortunately, there are issues with this method, such as those raised in #1640.  Mostly this is caused by sync-over-async antipattern.

I was going to rewrite most of it to be async, but after reviewing carefully, I'm fairly certain we don't need this feature at all.  I'd like to remove it.

The only thing it does is pauses initialization on the calling thread for a certain amount of time while it tries to flush the cache in a background task.  Either the cache is flushed within the timeout specified, or the timeout is reached and the flush keeps occurring in the background.

The reason I don't think we need it is that the `CachingTransport` *already* works in the background.  As soon as it is created, it starts a background task and enters a loop, waiting for a signal that something is being sent.  Then it processes items (same as flushing) before reseting the signal.

I propose we simply invert the loop.  Always start with processing/flushing items, then wait for the signal that there's more items being sent.  That will have the same effect as `BlockFlushCache` was doing, but without any need to worry about the `InitCacheFlushTimeout` and without any delay in completing initialization.

Associated PR forthcoming.

mattjohnsonpint:
Fixes #1642 and #1640

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Changelog
 
+## Unreleased
+
+### Fixes
+- Rework how the `InitCacheFlushTimeout` option is implemented. ([#1644](https://github.com/getsentry/sentry-dotnet/pull/1644))
+
 ## 3.17.0
 
 **Notice:** If you are using self-hosted Sentry, this version and forward requires either Sentry version >= [21.9.0](https://github.com/getsentry/relay/blob/master/CHANGELOG.md#2190), or you must manually disable sending client reports via the `SendClientReports` option.
--- a/src/Sentry/Internal/Http/CachingTransport.cs
+++ b/src/Sentry/Internal/Http/CachingTransport.cs
@@ -92,6 +92,9 @@ namespace Sentry.Internal.Http
                     // or on timeout, whichever comes first.
                     _initCacheResetEvent.Wait(_options.InitCacheFlushTimeout);
                 }
+
+                // We're done with this. Set null to avoid object disposed exceptions on future processing calls.
+                _initCacheResetEvent = null;
             }
         }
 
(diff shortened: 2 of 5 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 # Changelog
 
+## Unreleased
+
+### Fixes
+- Rework how the `InitCacheFlushTimeout` option is implemented. ([#1644](https://github.com/getsentry/sentry-dotnet/pull/1644))
```
```suggestion

- Rework how the `InitCacheFlushTimeout` option is implemented. ([#1644](https://github.com/getsentry/sentry-dotnet/pull/1644))
```

src/Sentry/Internal/Http/CachingTransport.cs
```diff
@@ -80,6 +82,20 @@ private void Initialize(bool startWorker)
...
             _worker = startWorker ? Task.Run(CachedTransportBackgroundTaskAsync) : Task.CompletedTask;
+
+            // Wait for init timeout, if configured.  (Can't do this without a worker.)
+            if (startWorker && _options.InitCacheFlushTimeout > TimeSpan.Zero)
+            {
+                using (_initCacheResetEvent = new ManualResetEventSlim())
+                {
+                    // This will complete either when the first round of processing is done,
+                    // or on timeout, whichever comes first.
+                    _initCacheResetEvent.Wait(_options.InitCacheFlushTimeout);
```
bruno-garcia, line 93: Is it worth adding this within a `if` and logging the outcome?

bruno-garcia (approved):
Much simpler solution, thanks Matt!

src/Sentry/Internal/Http/CachingTransport.cs
mattjohnsonpint: Probably.  Will do. Thanks.

PR #1646, opened 2022-05-06
Dont unsubscribe from AppDomain.UnobservedTaskException on dispose

### LINKED ISSUE #321: on async Main, unhandled exception unregisters before capturing crash
```csharp
static void Main(string[] args)
{
    using var _ = SentrySdk.Init();
    throw null;
}
```

This exception is captured as expected by the unhandled exception handler.

```csharp
static async Task Main(string[] args)
{
    using var _ = SentrySdk.Init();
    throw null;
}
```

The code above doesn't work as expected. Before calling the `UnhandledException` handler the SDK goes out of the `using` "block" and unregisters the unhandled exception integration.

SimonCropp:
fixes #321

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## Unreleased
+
+- Dont unsubscribe from AppDomain.UnobservedTaskException on dispose ([#1646](https://github.com/getsentry/sentry-dotnet/pull/1646))
+
 ## 3.17.0
 
 **Notice:** If you are using self-hosted Sentry, this version and forward requires either Sentry version >= [21.9.0](https://github.com/getsentry/relay/blob/master/CHANGELOG.md#2190), or you must manually disable sending client reports via the `SendClientReports` option.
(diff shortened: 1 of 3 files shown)


bruno-garcia (commented):
Are calling Unregister on integrations when the Hub is disposed? I wonder if that could be an alternative approach.
Integrations can still unregister (even though we wouldn't be using that from SentrySdk.Close -> Hub.Dispose

Thoughts?

SimonCropp:
no objections here. should i close this?

@bruno-garcia can u show us what u mean via a PR

mattjohnsonpint:
@bruno-garcia writes code? 😄

lucas-zimerman:
Could we add the `Unregister` to `ISdkIntegration ` instead of `IInternalSdkIntegration`?

mattjohnsonpint:
After some discussion, it seems to makes sense to kill `Unregister` completely, as long as we don't break integrations.

mattjohnsonpint:
Done in #1656

PR #1654, opened 2022-05-11
Hide public extension method classes from Intellisense

mattjohnsonpint:
We were using `[EditorBrowsable(EditorBrowsableState.Never)]` in a few places, but not consistently. This PR is adding where it was missing.

This attribute makes it so that the *class name* containing the extension method does not appear in Intellisense in an IDE.  We should use it on any public class designed purely to contain extension methods.

#skip-changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
+
+### Fixes
+
+- Hide public extension method container classes from Intellisense. ([#1654](https://github.com/getsentry/sentry-dotnet/pull/1654))
+
 ## 3.17.1
 
 ### Fixes
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,11 +1,5 @@
 # Changelog
 
-## Unreleased
-
-### Fixes
-
-- Hide public extension method container classes from Intellisense. ([#1654](https://github.com/getsentry/sentry-dotnet/pull/1654))
-
 ## 3.17.1
 
 ### Fixes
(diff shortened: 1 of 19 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
+
+### Fixes
+
+- Hide public extension method container classes from Intellisense. ([#1654](https://github.com/getsentry/sentry-dotnet/pull/1654))
+
 ## 3.17.1
```
bruno-garcia, line 9: Is it worth adding this to the changelog? Somehow it doesn't feel like a fix

bruno-garcia (approved):
One downside of doing this that I've identified is that because the class has this, it's public methods don't show up in the API docs:

https://getsentry.github.io/sentry-dotnet/api/index.html

But that's a DocFX issue I'd say

mattjohnsonpint:
On the DocFX issue, I believe that is configurable.  I'll add that here in a minute.

SimonCropp:
AFAIK r# and rider ignore `[EditorBrowsable(EditorBrowsableState.Never)]` ?

mattjohnsonpint:
It works in Rider, it's just controlled by an option that's off by default.

`Editor` - `General` - `Code Completion` - `Filter members by [EditorBrowsable] attribute`

I assume its similar in ReSharper, but didn't check.

mattjohnsonpint:
Looks like we already configured DocFX for this correctly.

https://github.com/getsentry/sentry-dotnet/blob/a9304a0a4b4702d0e62e2703d55c66483d27a0e5/docs/filterConfig.yml#L2-L6

Indeed, it is working.  For example:

https://github.com/getsentry/sentry-dotnet/blob/c6a660b1affc894441c63df2695a995701671744/src/Sentry/HubExtensions.cs#L11-L12

Docs are generated here:
https://getsentry.github.io/sentry-dotnet/api/Sentry.HubExtensions.html

Issue #1651, opened 2022-05-09
.NET MAUI Support (Preview 1)


bruno-garcia:
This is a subset of the issue tracking the complete MAUI support: #1174

### Preview 1
- [x] Add Sentry.Samples.Maui project
- [x] Add Sentry.Maui project 
- [x] Add Sentry.Tests.Maui project
- [x] Extension method on `MauiAppBuilder` with `UseSentry`
- [x] Set `sentry.dotnet.maui` SDK version string
- [x] Disable `io.sentry.auto-init` automatically instead of in `AndroidManifest.xml`
- [x] Enable `IsGlobalModeEnabled`
- [x] Hook into MAUI events: #1706
- [x] PII Options: #1709
- [x] Support Android native crashes: https://github.com/getsentry/sentry-dotnet/issues/1659
- [x] Gather device data: https://github.com/getsentry/sentry-dotnet/pull/1713
- [x] Add `sentry.dotnet.maui` to release registry: https://github.com/getsentry/sentry-release-registry/pull/72
- [x] Initial Sentry Docs: https://github.com/getsentry/sentry-docs/pull/5144
- [x] Publish Sentry.Maui NuGet package

The above items have been completed, and are published to Nuget in version `3.18.0-preview.1`

See #1729 for the next batch!

bruno-garcia:
@mattjohnsonpint what about we close this issue as released on preview.1 and move the bottom items to a new issue?

PR #1663, opened 2022-05-19
Initial MAUI support

mattjohnsonpint:
Adding first set of projects for MAUI (see #1651).

- Captures unhandled exceptions from managed code only.
- Unit tests for `UseSentry` registration methods
- Sample app that throws an unhandled exception and sends to Sentry:

image

image


mattjohnsonpint:
Note, I need to update the build steps to pull in the maui workload, so I expect this to fail on CI builds at the moment.  Will update that next.

samples/Sentry.Samples.Maui/App.xaml.cs
```diff
@@ -0,0 +1,11 @@
+﻿namespace Sentry.Samples.Maui;
+
+public partial class App : Application
+{
+	public App()
+	{
+		InitializeComponent();
```
bruno-garcia, line 7: Tabs 😭

mattjohnsonpint: Came with the sample app template, lol.  Will update.

.github/workflows/build.yml
```diff
@@ -36,11 +37,6 @@ jobs:
           with:
             submodules: recursive
 
-        - name: "Set up Java: 11"
```
mattjohnsonpint, line 39: Note, JDK 11 is preinstalled on all the images we use.  We don't need to install it separately.

src/Directory.Build.props
```diff
@@ -45,7 +45,7 @@
 
   <PropertyGroup>
     <!-- Used by SIL.ReleaseTasks below -->
-    <ChangelogFile>$(SolutionDir)/CHANGELOG.md</ChangelogFile>
```
mattjohnsonpint, line 48: Note: This fixed an error caused when trying to pack MAUI, because we call `dotnet pack` on just the one csproj, so there's no solution dir.

samples/Directory.Build.props
```diff
@@ -5,6 +5,7 @@
 
   <PropertyGroup>
     <IsPackable>false</IsPackable>
+    <SignAssembly>false</SignAssembly>
```
mattjohnsonpint, line 8: Note: MAUI currently requires this because the MAUI dependencies are unsigned.  But it's probably good form to not be signing our sample apps anyway.

mattjohnsonpint:
MAUI just went GA.  I'll update and take one more pass at this before we merge.

SimonCropp:
so the sample seems overly large for just demonstrating MAUI. eg can we remove the fonts, the image, aboutassets.txt, and any other stuff to trim it down

SimonCropp:
can we remove the styles as well?

mattjohnsonpint:
For now, I just want to use the sample app that comes with the MAUI default app template.  We can trim it down later.

mattjohnsonpint:
Something's broken with the sample app after updating to the MAUI GA. I'm investigating...

mattjohnsonpint:
The Android app crashes immediately on startup if `Sentry.csproj` is compiled with the `net6.0-android` target.  This worked in MAUI RC3, but isn't in GA.  I'll investigate more tomorrow.  For now, I've temporarily disabled that target so the sample app should stop crashing.

mattjohnsonpint:
... That broke the `Sentry.Samples.Android` build, of course.

SimonCropp:
i assume we want to disable this ?

image

mattjohnsonpint:
> i assume we want to disable this ?
> ...

Sure.  I'll define that constant in the next MAUI update PR.

PR #1680, opened 2022-05-27
Improve timestamp precision of transactions and spans

### LINKED ISSUE #1678: Transaction and Span timestamp precision
For both transactions and spans, we are setting both start and end timestamps using `DateTimeOffset.UtcNow`.  We send both timestamps to Sentry, which later computes durations by subtraction of the two values.

This is fine for long running tasks, or for traces that span across multiple requests or come from different machines (assuming everything is NTP synchronized).   But for short-lived tasks running on a single machine, it may give imprecise results due to the nature of the system's realtime clock.  It's also affected by clock sync, drift, and by the drift correction algorithms baked in to most modern OS's, so it's possible (though rare) that some values could come through with the end before the start.

A better approach is to use `System.Diagnostics.Stopwatch` to measure duration.  A stopwatch would be started at the same time the starting timestamp is set.  The ending timestamp would be calculated based on the start time plus the elapsed duration from the stopwatch.

Thanks to @vaind for pointing this out. 👍

mattjohnsonpint:
Fixes #1678

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Fixes
 
 - Flatten AggregateException ([#1672](https://github.com/getsentry/sentry-dotnet/pull/1672))
+- Improve timestamp precision of transactions and spans ([#1680](https://github.com/getsentry/sentry-dotnet/pull/1680))
 
 ### Features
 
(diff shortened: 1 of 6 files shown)


mattjohnsonpint:
Note, there's very little that can be done to test the actual improvement, as precision varies across systems and target framework versions, both for `UtcNow` and for `Stopwatch`.

SimonCropp (approved):

PR #1689, opened 2022-06-03
Always include stack frames from samples in DEBUG configuration

mattjohnsonpint:
In some cases, an exception will be shown in sentry as "Crashed in non-app", as in the first line of this one:

image

In this case, it's because `"Sentry."` is in the `InAppExclude` list, and we're coming from code in `Sentry.Samples.`

Samples should be treated the same as a real application, so add them to the `InAppInclude` list.

#skip-changelog

### DIFF
--- a/src/Sentry/SentryOptions.cs
+++ b/src/Sentry/SentryOptions.cs
@@ -680,10 +680,15 @@ namespace Sentry
                     "ServiceStack"
             };
 
+#if DEBUG
             InAppInclude = new[]
             {
                 "Sentry.Samples."
             };
+#else
+            InAppInclude = Array.Empty<string>();
+#endif
+
         }
     }
 }
--- a/src/Sentry/SentryOptions.cs
+++ b/src/Sentry/SentryOptions.cs
@@ -680,7 +680,10 @@ namespace Sentry
                     "ServiceStack"
             };
 
-            InAppInclude = Array.Empty<string>();
+            InAppInclude = new[]
+            {
+                "Sentry.Samples."
+            };
         }
     }
 }


SimonCropp:
shouldnt this be a sample level config? ie not in the core lib?

mattjohnsonpint:
Yeah, I suppose that makes sense. Would be nice if it could be in a single place though, and not implied that the user had to do anything similar.

bruno-garcia:
It's in the core lib to avoid sentry's frames show up and become noise. E.g middleware, withscope, and other cases

mattjohnsonpint:
Looking at this closer, we shouldn't need to actually add this to `InAppInclude`.  The logic in `SentryStackTraceFactory` (`ConfigurAppFrame` and `InternalCreateFrame`) should automatically set the `InApp` to `true` on frames that are fired from methods that are part of the main application.

Indeed, if I try just the basic console sample, that is marked as in-app.  So it must be something specific to the environment I was testing with in my MAUI samples.  I'll dig a bit deeper and open a new PR when ready.  Thanks.

mattjohnsonpint:
Actually, I was wrong on my last comment.  After doing some debugging, I found the reasons the basic console sample didn't trigger the same behavior was twofold:

- It uses top-level statements, so is in the global namespace.  The module name compared against the include/exclude list thus doesn't start with `Sentry.Samples`, it's just `Program`.
- It only has one module as reported to sentry, so even if it's sent with in-app false, the UI doesn't render the "Crashed in non-app".

I can get it to behave the same as my other example if I just create another class that *is* namespaced and throw the exception from there instead.

Program.cs
```csharp
using (SentrySdk.Init("https://eb18e953812b41c3aeb042e666fd3b5c@o447951.ingest.sentry.io/5428537"))
{
    Sentry.Samples.Console.Basic.Foo.Throw();
}
```

Foo.cs
```csharp
namespace Sentry.Samples.Console.Basic;

public static class Foo
{
    public static void Throw()
    {
        throw null;
    }
}
```

image

src/Sentry/SentryOptions.cs
```diff
@@ -680,7 +680,10 @@ public SentryOptions()
                     "ServiceStack"
             };
 
-            InAppInclude = Array.Empty<string>();
+            InAppInclude = new[]
+            {
+                "Sentry.Samples."
+            };
```
bruno-garcia, line 686: It's fair to say users run our sample always in debug mode? In that case we can avoid having this ship out in production 

```suggestion
#if DEBUG
            InAppInclude = new[]
            {
                "Sentry.Samples."
            };
#endif
```

mattjohnsonpint: Great idea.  I was exploring a different path with  frames from `Assembly.GetEntryAssembly()` being treated as in-app, but it's a rabbit hole due to all the ways apps can be run.  I'll just take your idea instead.  Don't want to waste too much time on this.

mattjohnsonpint:
I took Bruno's suggestion to just add it to the core in debug mode only.  Putting it on each sample app would make it noisy, and also imply to the developer that they have to do the same in their apps.  Also, not all samples are using the options delegate.

bruno-garcia:
Thanks Matt

PR #1690, opened 2022-06-04
Use `sent_at` instead of `sentry_timestamp` to reduce clock skew

### LINKED ISSUE #1687: Add missing parameter `sent_at` to the envelope header
### Problem Statement

see: https://github.com/getsentry/sentry/issues/35312 and https://develop.sentry.dev/sdk/envelopes/#data-model

this is a recommended parameter to be sent in order to properly set the clock drift correction for transactions/events.

### Solution Brainstorm

_No response_

SimonCropp:
fixes #1687


SimonCropp:
does it need to also be deserialized and exposed as a member?

mattjohnsonpint:
What happens when caching is enabled? Does `sent_at` still get set when the event is being sent? Or does it happen when the event is written to disk?

bruno-garcia (commented):
@jan-auer is this field new? Should we add this everywhere we have disk caching? I believe we don't have this on Java and cocoa either.

SimonCropp:
@mattjohnsonpint @bruno-garcia review please

mattjohnsonpint:
Sorry, but I still have doubts about this.  If we are adding it, it should be as late as possible, but we *already* have that on the headers.  See discussion on #1687.  Thanks.

SimonCropp:
@mattjohnsonpint yep i took that discussion into account. but on re-evaluating the code, this PR will result in the sent_at being added at the point EnvelopeHttpContent does `SerializeToStreamAsync`.

bruno-garcia:
Turned into draft

mattjohnsonpint:
Investigation results: `sentry_timestamp` [was removed](https://github.com/getsentry/sentry/blob/ac7ec3fc33de32ac96e513662f23a825ed7459cf/CHANGES#L2263) from Sentry and Relay quite a long time ago, and was not used for clock skew correction anyway.  After discussion with the TSC, we will just remove it from the dev docs, and it can optionally be removed from all SDKs.  Several have already removed it long ago.

`sent_at` is the way to go, and should be added as close as possible to the event being sent.  I think there may be a better way to do that here.  I'll take a look.

mattjohnsonpint:
FYI, `sent_at` is supposed to be in the envelope header, not the payload.  It also applies to several other envelope types, so I'm just going to always add it to the envelope header.  Update coming to this PR shortly.

src/Sentry/Envelopes/Envelope.cs
```diff
@@ -47,8 +50,10 @@ public Envelope(IReadOnlyDictionary<string, object?> header, IReadOnlyList<Envel
                 ? new SentryId(guid)
                 : null;
 
-        private async Task SerializeHeaderAsync(Stream stream, IDiagnosticLogger? logger, CancellationToken cancellationToken)
+        private async Task SerializeHeaderAsync(Stream stream, IDiagnosticLogger? logger, ISystemClock clock,
+            CancellationToken cancellationToken)
```
SimonCropp, line 54: NIT: is this some ide setting on your machine? i am not a fan of the "wrap one parameter"

mattjohnsonpint: updated

mattjohnsonpint (approved):

mattjohnsonpint (approved):

PR #1693, opened 2022-06-07
Improve serialization perf and fix memory leak in SentryEvent

mattjohnsonpint:
I ran into an issue wherein the `SentryExceptions` and `SentryThreads` properties on `SentryEvent` can return open `JsonElement` objects instead of their actual properties.  If the `JsonDocument` holding those elements is not disposed, we get a memory leak.  But also, if the document *is* disposed, we can't deserialize the data without getting an `ObjectDisposedException`.  The root cause turned out to be an un-materialized `IEnumerable` on those properties coming from `SentryEvent.FromJson`.

The tests were hiding this, because we were using our `Json.Parse(string)` helper function, which was using `.Clone()`, so the elements wouldn't have been disposed with the document.  Since cloning is semi-costly, I revised the helper methods to avoid it, by passing in a factory method delegate instead and using it to return a concrete object instead of a `JsonElement`.

Only one test failed with the change to the JSON helpers, which was for the issue had already found on `SentryEvent`.  Adding `.ToList()` in the correct place made the test pass again.

So - part fix, part perf improvement.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,8 @@
 
 - Improve timestamp precision of transactions and spans ([#1680](https://github.com/getsentry/sentry-dotnet/pull/1680))
 - Flatten AggregateException ([#1672](https://github.com/getsentry/sentry-dotnet/pull/1672))
-  - NOTE: This can affect grouping. You can keep the original behavior by setting the option `KeepAggregateException` to `true`.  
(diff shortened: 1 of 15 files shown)


bruno-garcia (approved):
Excellent! Thank toy for finding and fixing this.

It could explain a an issue a user described on discord, without repro.

PR #1694, opened 2022-06-07
Set Java/Android SDK options

mattjohnsonpint:
This sets *most* of the Java/Android SDK options when using the `net6.0-android` target from the .NET SDK.

- Many options are shared between both SDKs, and are simply copied across.
- A few Java SDK options are hard-coded because they don't make sense any other way in this context.
- Several options available on the Java `SentryOptions` and `SentryAndroidOptions` are not available on ours at all, so I have exposed them on `SentryOptions.Android` and mirrored the defaults.
- Some to/from mapping functions are added where needed.  Some map directly, others go through JSON serialization.
- I also refactored quite a bit, and fixed several TODOs (and added a few more)

This is also relevant for MAUI.

Note, #1692 and #1693 need to be in place before this works.
Also, we currently have no way to write unit tests for this code, but it's all been run manually several times on my box.


src/Sentry/Android/JavaExtensions.cs
```diff
@@ -0,0 +1,163 @@
...
+                return SpanStatus.Aborted;
+            if (status == Java.SpanStatus.OutOfRange)
+                return SpanStatus.OutOfRange;
+            if (status == Java.SpanStatus.DataLoss)
+                return SpanStatus.DataLoss;
+
+            throw new ArgumentException(null, nameof(status));
+        }
+
+        public static SentryEvent ToSentryEvent(this Java.SentryEvent sentryEvent, Java.SentryOptions javaOptions)
```
SimonCropp, line 88: why the need to round trip via serialization? can the props of Java.SentryEvent be duplicated into a  SentryEvent  without the serialization?

src/Sentry/Android/JavaExtensions.cs
```diff
@@ -0,0 +1,163 @@
...
+
+        public static Java.SentryLevel ToJavaSentryLevel(this SentryLevel level) =>
+            level switch
+            {
+                SentryLevel.Debug => Java.SentryLevel.Debug!,
+                SentryLevel.Info => Java.SentryLevel.Info!,
+                SentryLevel.Warning => Java.SentryLevel.Warning!,
+                SentryLevel.Error => Java.SentryLevel.Error!,
+                SentryLevel.Fatal => Java.SentryLevel.Fatal!,
+                _ => throw new ArgumentException(null, nameof(level))
```
bruno-garcia, line 38: Good case for ArgumentOutOfRange?
Curious why `null` instead of level as an arg. Could we use it? Would be useful to know what level was given.

src/Sentry/Android/SentryOptions.cs
```diff
@@ -0,0 +1,61 @@
...
+
+            internal string[]? InAppExclude { get; set; }
+            internal string[]? InAppInclude { get; set; }
+
+            public void AddInAppExclude(string prefix) =>
+                InAppExclude = InAppExclude != null
+                    ? InAppExclude.Concat(new[] { prefix }).ToArray()
+                    : new[] { prefix };
+
+            public void AddInAppInclude(string prefix) =>
```
bruno-garcia, line 55: We'd document users to pass android packages here (in the format io.sentry.samples), right? So filter down in the native layer? It'll be in both layers but not a big deal

src/Sentry/Android/SentrySdk.cs
```diff
@@ -1,52 +1,162 @@
...
                 return evt;
             }));
 
-            Sentry.Android.SentryAndroid.Init(context, new JavaLogger(options),
-                new ConfigureOption(o =>
+            SentryAndroid.Init(context, new JavaLogger(options),
+                new OptionsConfigurationCallback(o =>
                 {
+                    // TODO: Should we set the DistinctId to match the one used by GlobalSessionManager?
+                    //o.DistinctId = ?
```
bruno-garcia, line 52: Good point, if we're initializing the dotnet layer first, we could pass down our installation id (will require a new api in java).
In unity we init android first. So during the unity init, we read the installation id from the Java layer and reuse.

src/Sentry/Android/SentrySdk.cs
```diff
@@ -1,52 +1,162 @@
...
+                    }
+
+                    if (options.BeforeSend != null)
+                    {
+                        o.BeforeSend = new BeforeSendCallback(options.BeforeSend, options.DiagnosticLogger, o);
+                    }
+
+                    if (options.TracesSampler != null)
+                    {
+                        o.TracesSampler = new TracesSamplerCallback(options.TracesSampler);
```
bruno-garcia, line 102: We often check for the existence of an instance here to make decisions (performance is enabled, add auto instrumentation) so this could have side effects

bruno-garcia (commented):
Looking great

src/Sentry/Android/JavaExtensions.cs
mattjohnsonpint: Great question.  So, I started down the path of mapping individual properties, similar to the `ToSentryLevel`/`ToJavaSentryLevel` methods.  I quickly found lots of problems:

- The mapping code was _ginormous_, due to the number of different objects that are part of the object graph. All of these objects would need to be mapped:  `SentryEvent`, `SentryMessage`, `SentryValues`, `SentryException`, `SentryStackTrace`, `SentryStackFrame`, `Mechanism`, `SentryThread`, `DebugImage`, `Contexts`, `User`, `SdkVersion`, `BreadCrumb`, `App`, `Browser`, `Device`, `DeviceOrientation`, `OperatingSystem`, `Runtime`, `Gpu`, `Trace`, `SpanId`, `SentryId`, `SpanStatus` (perhaps more).
- There are lots of places with collections of `object` typed properties.  They could end up containing usable primitives, but they could also contain references to Java objects and might be difficult to work with in .NET - especially since we have made all of Sentry's Java public API *internal* for .NET in the JNI transforms.
- The objects are not completely the same on both sides.  Trying to handle all the differences leads to fragile code that likely will miss updates.  Through serialization, we at least get a contract that both sides use the same way, as it's the same that is required for shipping data to Sentry.

I recognize there may be some performance disadvantages to this approach, so we may want to map *some* of the object graph, or perhaps choose to use the original object for the return value if its clone hasn't been modified (if we can detect that).  But for now, this works.

src/Sentry/Android/SentryOptions.cs
mattjohnsonpint: Yes.  These are separate include/exclude filters for the Android SDK only.  I'll add xml-docs to clarify.

src/Sentry/Android/JavaExtensions.cs
SimonCropp: ok. thanks for the clarification

mattjohnsonpint: I actually had an idea that might avoid the serialization cost. I'll try it out before going into details just yet.  Stay tuned...

mattjohnsonpint:
I had to revert the last commit.  I was attempting to map Java enums to C# enums using the recommended approach [documented here](https://docs.microsoft.com/en-us/xamarin/android/platform/binding-java-library/customizing-bindings/java-bindings-metadata#enumfieldsxml-and-enummethodsxml).  The library compiled fine, but generated the wrong method signature in the [Android Callable Wrapper](https://docs.microsoft.com/en-us/xamarin/android/platform/java-integration/android-callable-wrappers), so failed when building one of the sample apps.  I hunted for a workaround but found none.

I need to raise this as an upstream issue to Microsoft, but in the meantime we can just keep things as they were.

src/Sentry/Android/JavaExtensions.cs
mattjohnsonpint: I'll come back to the serialization issue in a separate PR.  This works for now, and doesn't even matter if you aren't using `BeforeSend` or `BeforeBreadcrumb` delegates.

mattjohnsonpint: I'll come back to the serialization issue in a separate PR.  This works for now, and doesn't even matter if you aren't using `BeforeSend` or `BeforeBreadcrumb` delegates.

mattjohnsonpint:
FYI - the enum issue is reported here: https://github.com/xamarin/java.interop/issues/991

src/Sentry/Android/SentrySdk.cs
mattjohnsonpint: I'll address this in a future PR.  Thanks.

src/Sentry/Android/SentrySdk.cs
mattjohnsonpint: I'll address this in a future PR.  Thanks.

src/Sentry/Android/JavaExtensions.cs
```diff
@@ -0,0 +1,158 @@
...
+            using var jsonReader = new Java.JsonObjectReader(streamReader);
+            using var deserializer = new Java.SentryEvent.Deserializer();
+            return deserializer.Deserialize(jsonReader, javaOptions.Logger);
+        }
+
+        public static Breadcrumb ToBreadcrumb(this Java.Breadcrumb breadcrumb, Java.SentryOptions javaOptions)
+        {
+            using var stream = new MemoryStream();
+            using var streamWriter = new JavaOutputStreamWriter(stream);
+            using var jsonWriter = new Java.JsonObjectWriter(streamWriter, javaOptions.MaxDepth);
```
SimonCropp, line 114: why does only one of these use `javaOptions.MaxDepth`

src/Sentry/Android/JavaLogger.cs
```diff
@@ -0,0 +1,21 @@
...
+
+        public JavaLogger(SentryOptions options) => _options = options;
+
+        public void Log(Java.SentryLevel level, string message, JavaObject[]? args) =>
+            _options.DiagnosticLogger?.Log(level.ToSentryLevel(), message, null, args?.Cast<object?>());
+
+        public void Log(Java.SentryLevel level, string message, Throwable? throwable) =>
+            _options.DiagnosticLogger?.Log(level.ToSentryLevel(), message, throwable);
+
+        public void Log(Java.SentryLevel level, Throwable? throwable, string message, params JavaObject[]? args) =>
```
SimonCropp, line 15: isnt the `?` on `params JavaObject[]? args`  redundant?

mattjohnsonpint: Possibly, but it's on the auto generated `Java.ILogger` interface, so it has to be here too or it fails with a compilation error.

PR #1697, opened 2022-06-07
Use android 6.0.0

mattjohnsonpint:
Update to the [Sentry-Android 6.0.0 GA release](https://github.com/getsentry/sentry-java/releases/tag/6.0.0)

Also cleaned up the changelog a bit.


bruno-garcia (approved):
Next step is to get these jar/aar out of the repo, and have a task that downloads them on demand. Otherwise we'll bloat the repo with these binaries.

Issue #1695, opened 2022-06-07
InvalidCastException when trying to add device context


IasonasNikolaou:
**Please mark the type framework used:**
- [x ] Other: Maui

**Please mark the type of the runtime used:**

- [x ] .NET Framework
* Version: 6.0.5

**Please mark the NuGet packages used:**

- [ x] Sentry
* Version: 3.17.1

**Steps to reproduce the issue or link to a repository with a small reproducible code.** 
Add sentry to a maui or probably even a .net project. I don't think thats a maui related issue tbh.

I've got this piece of code in my Main which worked fine until i decided i also needed to add device info
```
SentrySdk.ConfigureScope(scope =>
            {
                scope.User = new User
                {
                    Id = "1",
                    Username = "Testuser",
                };
                scope.Contexts["os"] = new
                {
                    name = AndroidInfo.platform,
                    version = AndroidInfo.version,
                };
                scope.Contexts["device"] = new
                {
                    name = AndroidInfo.deviceName,
                    family = AndroidInfo.manufacturer,
                    model = AndroidInfo.device,
                    manufacturer = AndroidInfo.manufacturer,
                }; 
            });
```
"os" and  "user" are working just fine on their own, however "device" returns this exception
```
[DOTNET] System.InvalidCastException: Specified cast is not valid.
[DOTNET]    at Sentry.Internal.Extensions.CollectionsExtensions.GetOrCreate[Device](ConcurrentDictionary`2 dictionary, String key) in /_/src/Sentry/Internal/Extensions/CollectionsExtensions.cs:line 13
[DOTNET]    at Sentry.Contexts.get_Device() in /_/src/Sentry/Protocol/Contexts.cs:line 30
[DOTNET]    at Sentry.Internal.MainSentryEventProcessor.Process(SentryEvent event) in /_/src/Sentry/Internal/MainSentryEventProcessor.cs:line 45
[DOTNET]    at Sentry.SentryClient.DoSendEvent(SentryEvent event, Scope scope) in /_/src/Sentry/SentryClient.cs:line 201
```
I should note that all of these are strings and do have values. 
If i try to add the device context as something else (ex:```scope.Contexts["deviceinfo"]```) it works just fine.
Is there any way to get around this error?

mattjohnsonpint:
@jaisonass - Thanks for reporting this issue with the `InvalidCastException`.

[...]

PR #1700, opened 2022-06-07
Add type checking in contexts try get value

SimonCropp:
mitigates #1695


SimonCropp:
@bruno-garcia @mattjohnsonpint a better fix would be for Contexts to not inherit from ConcurrentDictionary and instead wrap it. then we can check if people try and set reserved keys. but that would be a breaking change

mattjohnsonpint (approved):

PR #1706, opened 2022-06-11
Continue with adding MAUI support

mattjohnsonpint:
Continues #1651 

- Hook into MAUI events (adds breadcrumbs for now, will add spans for tracing support later)
  - We still will need to do more work here to reduce the number of breadcrumbs sent to those that are meaningful only, include additional data for some events where useful, and remove some data items that might contain PII.

Also:

- Hooks the `Microsoft.UI.Xaml.Application.UnhandledException` handler when targeting Windows
- Fixes the previous workaround in the sample app's csproj, so it runs on iOS and maccatalyst without crashing
- Registers a `Disposer` that will ensure the Sentry SDK is disposed with the `ServiceProvider`, so we flush events to Sentry during shutdown.


src/Sentry.Maui/Internal/Disposer.cs
```diff
@@ -0,0 +1,23 @@
+namespace Sentry.Maui.Internal;
+
+// This is a helper we register as a singleton on the service provider.
+// It allows us to register other items that should be disposed when the service provider disposes.
+// TODO: There might be something like this built-in to .NET already.  Investigate and replace if so.
+
+internal class Disposer : IDisposable
```
SimonCropp, line 7: no use case for async  dispose?

mattjohnsonpint: Not at the moment.  If we need it later we can add it.

CHANGELOG.md
```diff
@@ -19,6 +19,7 @@
 - Added a new `Sentry.Maui` integration library for the [.NET MAUI](https://dotnet.microsoft.com/apps/maui) platform:
   - Initial MAUI support ([#1663](https://github.com/getsentry/sentry-dotnet/pull/1663))
   - Continue with adding MAUI support ([#1670](https://github.com/getsentry/sentry-dotnet/pull/1670))
+  - ontinue with adding MAUI support ([#1706](https://github.com/getsentry/sentry-dotnet/pull/1706))
```
bruno-garcia: What about:

```suggestion
  - MAUI events become extra context in Sentry events ([#1706](https://github.com/getsentry/sentry-dotnet/pull/1706))
```

src/Sentry.Maui/Internal/Extensions.cs
```diff
@@ -0,0 +1,70 @@
...
+
+        if (element is ITitledElement { Title: { } } titledElement)
+        {
+            // TODO: Scrub PII ?
+            data.Add(prefix + nameof(titledElement.Title), titledElement.Title);
+        }
+
+        if (element is IText { Text: { } } textElement)
+        {
+            // TODO: Scrub PII ?
```
bruno-garcia, line 66: We're not adding PII scrubbing rules in SDK but this can be scrubbed on BeforeBreadcrumb if there's a risk.
Since we're adding this OOTB, it's less of an issue as if we added on a patch release or something. So folks can see what goes OOTB and add their own PII stripping rules on the backend, or BeforeBreadcrumb if needed

src/Sentry.Maui/Internal/SentryMauiInitializer.cs
```diff
@@ -0,0 +1,111 @@
...
+        // This will ensure that it gets disposed when the service provider is disposed.
+        // TODO: re-evaluate this with respect to MAUI app lifecycle events
+        disposer.Register(disposable);
+
+        // Bind MAUI events
+        var binder = services.GetRequiredService<MauiEventsBinder>();
+        binder.BindMauiEvents();
+
+#if IOS || MACCATALYST
+        // Workaround for https://github.com/xamarin/xamarin-macios/issues/15252
```
bruno-garcia, line 35: Possibly good candidate to go into the SentrySdk.Init for the mac/ios TFMs, once we get them in

bruno-garcia (approved):

src/Sentry.Maui/Internal/Extensions.cs
mattjohnsonpint: Honestly, I think I should just remove it.  It's too risky, IMHO.

mattjohnsonpint: Or put it behind a feature flag perhaps?

PR #1708, opened 2022-06-13
Send the .NET Runtime Identifier to Sentry

mattjohnsonpint:
This sends the value from `System.Runtime.InteropServices.RuntimeInformation.RuntimeIdentifier` to Sentry with all events, on the runtime context info.  This is helpful (for example) when working with MAUI events.

It shows up in the Sentry issues page under the Runtime section further down the page.

image

Note, that the `RuntimeIdentifier` property is only available on .NET 5 and greater, so it is not sent for other targets.

### DIFF
--- a/src/Sentry/Internal/Enricher.cs
+++ b/src/Sentry/Internal/Enricher.cs
@@ -17,6 +17,7 @@ namespace Sentry.Internal
             {
                 Name = current.Name,
                 Version = current.Version,
+                Identifier = current.Identifier,
                 RawDescription = current.Raw
             };
         });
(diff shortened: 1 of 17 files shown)


src/Sentry/Internal/Enricher.cs
```diff
@@ -82,6 +83,13 @@ public void Apply(IEventLike eventLike)
 
             // Default tags
             _options.ApplyDefaultTags(eventLike);
+
+            // Pass the runtime identifier as a tag also
+            const string ridTag = "runtime.identifier";
+            if (_runtimeLazy.Value.Identifier is { } ridValue && !eventLike.Tags.ContainsKey(ridTag))
+            {
+                eventLike.SetTag(ridTag, ridValue);
```
bruno-garcia, line 91: No risk of this becoming a high cardinality value?

Runtime info isn't sent as a tag from the SDK, ingest extracts the tag at ingestion time. @jan-auer brought this up in the past, lets see if he has any pointers.

mattjohnsonpint: Makes sense.  Let me know if there's a better place to put it.  Either way, it's very useful information that became available starting with .NET 5 that we haven't been including on events yet.

mattjohnsonpint:
Per discussion, it would be nice to have it appear as a tag, like this:

image

But we don't want to send it as a *custom* tag for various reasons.  Instead we'll send it just on the runtime context information, and consider having it extracted as a tag on the ingestion side.

bruno-garcia (approved):

PR #1709, opened 2022-06-13
Add options for PII breadcrumbs from MAUI events

mattjohnsonpint:
Disables automatically adding breadcrumbs from MAUI events that might contain PII, and adds options to enable them if desired.


bruno-garcia (approved):
Should we flip them all to true if SendDefaultPii is true?

mattjohnsonpint:
Nah, I think they shouldn't be considered "default" PII at all.  (Most people won't want them on in production, but may turn them on while troubleshooting.)

PR #1713, opened 2022-06-14
Add device information for MAUI and Android

mattjohnsonpint:
Continuing #1651

- For all MAUI platforms, add device information to the event context as given by MAUI
- For Android (with or without MAUI), augment with information gathered by the Sentry Android SDK

There are some gaps identified with `TODO` markers, but we can fill them in later.  I think this is good enough for now.

### DIFF
--- a/src/Sentry/Android/AndroidEventProcessor.cs
+++ b/src/Sentry/Android/AndroidEventProcessor.cs
@@ -6,7 +6,7 @@ namespace Sentry.Android;
 
 internal class AndroidEventProcessor : ISentryEventProcessor, IDisposable
 {
-    private readonly Java.IEventProcessor _androidProcessor;
+    private readonly Java.IEventProcessor? _androidProcessor;
     private readonly Java.Hint _hint = new();
 
     public AndroidEventProcessor(SentryAndroidOptions androidOptions)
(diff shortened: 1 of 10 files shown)


src/Sentry/Android/AndroidEventProcessor.cs
```diff
@@ -0,0 +1,38 @@
...
+
+internal class AndroidEventProcessor : ISentryEventProcessor, IDisposable
+{
+    private readonly Java.IEventProcessor _androidProcessor;
+    private readonly Java.Hint _hint = new();
+
+    public AndroidEventProcessor(SentryAndroidOptions androidOptions)
+    {
+        _androidProcessor = androidOptions.EventProcessors.OfType<JavaObject>()
+            .Where(x => x.Class.Name == "io.sentry.android.core.DefaultAndroidEventProcessor")
```
bruno-garcia, line 15: If proguard is enabled, it might not find the type. We can look into bundling proguard rules to avoid this type name from being minified.

Proguard rules are bundled in the Sentry Android SDK:
https://github.com/getsentry/sentry-java/pull/2031/files

But we likely can add our own

See; https://docs.microsoft.com/en-us/xamarin/android/deploy-test/release-prep/proguard?tabs=macos

src/Sentry/Android/AndroidEventProcessor.cs
```diff
@@ -0,0 +1,38 @@
...
+            .Cast<Java.IEventProcessor>()
+            .First();
```
bruno-garcia, line 17: Can we use FirstOrDefault and downgrade?

src/Sentry/Protocol/Device.cs
```diff
@@ -209,10 +209,12 @@ public sealed class Device : IJsonSerializable
         /// </summary>
         /// <example>
         /// iOS: UIDevice.identifierForVendor (UUID)
-        /// Android: md5 of ANDROID_ID
+        /// Android: The generated Installation ID
         /// Windows Store Apps: AdvertisingManager::AdvertisingId (possible fallback to HardwareIdentification::GetPackageSpecificToken().Id)
         /// Windows Standalone: hash from the concatenation of strings taken from Computer System Hardware Classes
         /// </example>
+        /// TODO: Investigate - Do ALL platforms now return a generated installation ID?
```
bruno-garcia, line 216: They should. ANDROID_ID isn't privacy aware, different apps can get the same value and 'link' a user across apps.

Our goal is not to use such ids. Apple now asks for permission for accessing the "advertising id" which we don't use in the iOS SDK.

bruno-garcia (approved):

src/Sentry/Android/AndroidEventProcessor.cs
mattjohnsonpint: Thanks.  Resolved with a bit of trickery.

PR #1717, opened 2022-06-15
Add OS information from MAUI device data

mattjohnsonpint:
Similar to #1716, this sets OS context info for all MAUI platforms.  For example on iOS:

image

image


bruno-garcia (approved):
🚀

PR #1727, opened 2022-06-16
iOS native support

bruno-garcia:
Hopefully one day: Resolves https://github.com/getsentry/sentry-dotnet/issues/1282

PR #1732, opened 2022-06-17
Download Android SDK during restore

mattjohnsonpint:
This downloads the Sentry Android SDK during the restore phase of the build, removing the need to check in the arr/jar files.

#skip-changelog


bruno-garcia (approved):
🔥

bruno-garcia:
if we don't go with cocoapods on #1727 , we can use this to download the release we need from github

PR #1733, opened 2022-06-18
Expose EnumerateChainedExceptions

SimonCropp:


src/Sentry/SentryExceptionExtensions.cs
```diff
@@ -18,6 +20,36 @@ public static class SentryExceptionExtensions
     public static void AddSentryTag(this Exception ex, string name, string value)
         => ex.Data.Add($"{MainExceptionProcessor.ExceptionDataTagKey}{name}", value);
 
+    /// <summary>
+    /// Recursively enumerates all <see cref="AggregateException.InnerExceptions"/> and <see cref="Exception.InnerException"/>
+    /// Not for public use.
+    /// </summary>
+    public static IEnumerable<Exception> EnumerateChainedExceptions(this Exception exception, SentryOptions options)
```
bruno-garcia, line 27: This not being for public use but part of the `global` namespace don't go well together.
Should we remove the note on the xml doc?

Or perhaps this is best not being an extension method of Exception? If it's under a different namespace it will already help not show up on the user's intellisense. The goal is to share the built-in behavior with the Unity IL2CPP support

src/Sentry/Internal/MainExceptionProcessor.cs
```diff
@@ -82,49 +82,34 @@ public void Process(Exception exception, SentryEvent sentryEvent)
...
-                }
-            }
-            else if (exception.InnerException != null)
-            {
-                foreach (var inner in CreateSentryException(exception.InnerException))
-                {
-                    yield return inner;
-                }
-            }
+            return exception.EnumerateChainedExceptions(_options)
```
bruno-garcia, line 85: I was thinking we could refactor out this class to have a base class, so that the code that follows it doesn't need to iterate on the exceptions once again:

https://github.com/getsentry/sentry-unity/pull/683/files#diff-d7c6f55c576cbd905ba5affcc1c578fd120655cdfcdc520994c837efeeebccc2R33

Then a third time to have the original + the Sentry exception side by side:

https://github.com/getsentry/sentry-unity/pull/683/files#diff-d7c6f55c576cbd905ba5affcc1c578fd120655cdfcdc520994c837efeeebccc2R47

SimonCropp: @bruno-garcia i dont really understand what you are proposing here. can we sync my tuesday morning?

bruno-garcia: I thought of re-thinking these hooks but didn't get that far, https://github.com/getsentry/sentry-dotnet/pull/1734

bruno-garcia (approved):
Unless @mattjohnsonpint disagrees, we can go ahead with this

src/Sentry/SentryExceptionExtensions.cs
mattjohnsonpint, line 27:
```suggestion
    [EditorBrowsable(EditorBrowsableState.Never)]
    public static IEnumerable<Exception> EnumerateChainedExceptions(this Exception exception, SentryOptions options)
```

mattjohnsonpint: I'm ok with it as is, since it includes a `SentryOptions` parameter - we're not likely to step on anyone.  We can hide it from intellisense with `EditorBrowsable`

mattjohnsonpint (approved):
Approved, with suggestions

PR #1734, opened 2022-06-20
Not to be merged Exp/stack trace rewrite

bruno-garcia:
The goal was to avoid iterating over things multiple times, and give the native integration a simpler hook:

https://github.com/getsentry/sentry-unity/pull/827/files#diff-d7c6f55c576cbd905ba5affcc1c578fd120655cdfcdc520994c837efeeebccc2R26

### DIFF
--- a/src/Sentry/Exceptions/ExceptionProcessor.cs
+++ b/src/Sentry/Exceptions/ExceptionProcessor.cs
@@ -1,0 +1,146 @@
+using System;
+using System.Collections.Generic;
+using Sentry.Extensibility;
+using Sentry.Protocol;
+
+namespace Sentry.Exceptions
+{
+    public abstract class ExceptionProcessor : ISentryEventExceptionProcessor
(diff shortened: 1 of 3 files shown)


src/Sentry/Exceptions/ExceptionProcessor.cs
```diff
@@ -0,0 +1,146 @@
...
+
+        protected ExceptionProcessor(SentryOptions options) => _options = options;
+
+        public void Process(Exception exception, SentryEvent sentryEvent)
+        {
+            _options.LogDebug("Running processor {0} on exception: {1}", GetType(), exception.Message);
+
+            var sentryExceptionList = new List<SentryException>();
+
+            foreach (var flattenedEx in Flatten(exception))
```
bruno-garcia, line 23: The goal was to iterate on the exception list once once, but not sure it's worth rewriting this whole thing

Issue #327, opened 2019-12-20
Send project root path with events


bruno-garcia:
IDE's (at least Rider) can open relative path and line:

`src/SymbolCollector.Android/Resources/layout/activity_main.axml:5`

<img width="680" alt="image" src="https://user-images.githubusercontent.com/1633368/71259960-f80e0d80-2339-11ea-9f9c-fdfa847e8e5b.png">

[New changes for .NET stack traces are incoming to Sentry](https://github.com/getsentry/sentry/pull/16199) and it includes an easy way to copy the path.

Currently, the rooted path is copied, which is often useless since code is built on a build server and that rooted path doesn't match the developer's file structure.

This task is to include in events the root path of the project so that in the UI we can trim the build server specific part of the path.

A line number can then be appended to it with `:` and pasting in the IDE as above will work.

kanadaj:
ReSharper can also do this in Visual Studio:
![image](https://user-images.githubusercontent.com/3247357/71261383-22a99780-2335-11ea-8ef7-b1ed5a14caed.png)

Visual Studio's `Go to All` can also do this, but not with the line number:
![image](https://user-images.githubusercontent.com/3247357/71261507-6ef4d780-2335-11ea-974a-a2420d3ba9fa.png)

![image](https://user-images.githubusercontent.com/3247357/71261518-74eab880-2335-11ea-9e49-fd02d4b0b750.png)

PR #1703, opened 2022-06-10
MAUI Android Device Tests

vaind:
Integrates MAUI TestUtils, as mentioned in https://github.com/dotnet/maui/issues/3552#issuecomment-1129717167 and https://github.com/dotnet/xharness/issues/738

Some tests still fail, see e.g. https://github.com/getsentry/sentry-dotnet/pull/1703#issuecomment-1174010891 so I've enabled `continue-on-error` for the Android test run and I suggest creating a follow-up issue to resolve these failures.

closes #1704 
closes #1662 

#skip-changelog

PR #1739, opened 2022-06-22
Change FileName to be relative to project root

SimonCropp:
hopefully helps with #327


SimonCropp:
@bruno-garcia this embeds the project and solution dir into the assembly that consumes Sentry. 

so where would i wire this in to help with https://github.com/getsentry/sentry-dotnet/issues/327

bruno-garcia (commented):
Nice! yeah, I think we need to figure out where that goes into the payload next.
@untitaker do you know about this? (see: https://github.com/getsentry/sentry-dotnet/issues/327)

untitaker:
What is the question?

bruno-garcia:
> What is the question?

We have `abs_path`

```
abs_path
The absolute path to the source file.
```

We currently have the rooted path there `D:\bla\something\else\app\file.cs`.
This PR allows us to know where `app`. 

Do we add the 'project root path' somewhere to the payload? Or do we just truncate the `D:\bla\something\else` out of the `abs_path` and just send the relevant 'project path'? More details on: https://github.com/getsentry/sentry-dotnet/issues/327#issue-541012498

untitaker:
You're correct, you want to put the relative path into filename. That's what the python sdk does at least

sorry for the slow response, i definetly replied via email but for some reason it didn't show up here

src/Sentry/buildTransitive/Sentry.props
```diff
@@ -0,0 +1,35 @@
...
+        <_Parameter1>Sentry.SolutionDirectory</_Parameter1>
+        <_Parameter2>$(SolutionDir)</_Parameter2>
+      </SentryAttributes>
+      <!-- Ensure not part of Compile, as a workaround for https://github.com/dotnet/sdk/issues/114 -->
+      <Compile Remove="$(SentryAttributesFilePath)" />
+    </ItemGroup>
+    <WriteCodeFragment AssemblyAttributes="@(SentryAttributes)"
+                       Language="$(Language)"
+                       OutputFile="$(SentryAttributesFilePath)">
+      <Output TaskParameter="OutputFile" ItemName="Compile" Condition="$(Language) != 'F#'" />
```
bruno-garcia, line 30: `Condition="$(Language) != 'F#'"` does this mean we want this for VB or a possible IronRuby or something if that's a thing?

test/Sentry.Tests/AttributeReaderTests.cs
```diff
@@ -0,0 +1,12 @@
+﻿using AttributeReader = Sentry.AttributeReader;
+
+public class AttributeReaderTests
+{
+    [Fact]
```
bruno-garcia, line 5: Cool, this should run on iOS AOT'ed so we would see how it behaves there. Assuming the tests are running there following #1703

mattjohnsonpint:
Thats odd.  It seems to be coming up blank on Mono in CI:

```
Error: Assert.Equal() Failure
          ↓ (pos 0)
Expected: Internals/SentryStackTraceFactoryTests.cs
Actual:   
          ↑ (pos 0)
```

But it works on my local machine, even on Mono.  hmmm

SimonCropp:
@mattjohnsonpint  r u running in release mode locally?

mattjohnsonpint:
Testing further, it doesn't matter if debug vs release mode.  It passes on all targets when run from Rider's test runner.  It fails on NET461 (Mono) when run from `dotnet test` on the command line.

I've seen before where I get filenames and line numbers in Rider, but that was when a debugger was attached.  This is in the test runner.  Whatever they are doing special, they're doing it for tests as well.

src/Sentry/Sentry.csproj
```diff
@@ -55,4 +55,9 @@
     <PackageReference Include="System.Text.Json" Version="5.0.2" />
   </ItemGroup>
 
+  <ItemGroup>
+    <Content Include="buildTransitive\Sentry.props" PackagePath="buildTransitive\Sentry.props" />
+    <Content Include="buildTransitive\Sentry.props" PackagePath="build\Sentry.props" />
```
mattjohnsonpint, line 60: I'm seeing two warnings in the build output related to these:

```
buildTransitive/Sentry.props : warning XA0101: @(Content) build action is not supported [/Users/matt/Code/getsentry/sentry-dotnet/src/Sentry/Sentry.csproj]
buildTransitive/Sentry.props : warning XA0101: @(Content) build action is not supported [/Users/matt/Code/getsentry/sentry-dotnet/src/Sentry/Sentry.csproj]
```

mattjohnsonpint:
FYI - even if I build and test (passing) in Rider, and then test without building so it's using the exact same code (`dotnet test test/Sentry.Tests --filter FileNameShouldBeRelative -f net461 --no-build`), it still fails.  So I don't believe it's a compiler flag, but something being done at runtime.

src/Sentry/Sentry.csproj
SimonCropp: fixed

src/Sentry/buildTransitive/Sentry.props
SimonCropp: for some reason the item name for F# is `CompileBefore` but for VB and c# it is `Compile`. 

for other languages (like IronRuby) i have not tested. do you want me to include any other languages? Currently they will be a no-op

mattjohnsonpint (approved):

untitaker:
You're correct, you want to put the relative path into filename. That's
what the python sdk does at least

PR #1743, opened 2022-06-23
chore: unity stack trace fix

vaind:
See https://github.com/getsentry/sentry-unity/issues/845 - we're implementing a workaround to demangle the stack trace.

Additionally, I've changed existing code to use pre-compiled regex too.

The added test doesn't actually test the new code, it's just something I've added while trying to reproduce it. In the end, I couldn't make a reproduction in the sentry-dotnet repo which led me to believe the issue is Unity specific.

#skip-changelog

### DIFF
--- a/test/Sentry.Tests/Internals/SentryStackTraceFactoryTests.cs
+++ b/test/Sentry.Tests/Internals/SentryStackTraceFactoryTests.cs
@@ -46,9 +46,9 @@ public class SentryStackTraceFactoryTests
     }
 
     [Theory]
-    [InlineData(StackTraceMode.Original, "Create_Async_CurrentStackTrace { <lambda> }")]
-    [InlineData(StackTraceMode.Enhanced, "void SentryStackTraceFactoryTests.Create_Async_CurrentStackTrace(StackTraceMode mode, string method)+() => { }")]
-    public void Create_Async_CurrentStackTrace(StackTraceMode mode, string method)
+    [InlineData(StackTraceMode.Original, "AsyncWithWait_StackTrace { <lambda> }")]
+    [InlineData(StackTraceMode.Enhanced, "void SentryStackTraceFactoryTests.AsyncWithWait_StackTrace(StackTraceMode mode, string method)+() => { }")]
+    public void AsyncWithWait_StackTrace(StackTraceMode mode, string method)
     {
         _fixture.SentryOptions.AttachStacktrace = true;
         _fixture.SentryOptions.StackTraceMode = mode;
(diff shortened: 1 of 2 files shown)


SimonCropp:
isnt this something that should be pushed into Ben.Demystifier ?

vaind:
> isnt this something that should be pushed into Ben.Demystifier ?

I have no experience in that area - isn't it only used with "enhanced" stack traces? [Unity doesn't use those](https://github.com/getsentry/sentry-unity/issues/675)

vaind:
> > isnt this something that should be pushed into Ben.Demystifier ?
> 
> This only really affects the "Original" mode
> 
> See tests:
> 
> ```
>         if (_fixture.SentryOptions.StackTraceMode == StackTraceMode.Original)
> ```
> 
> Which makes me wonder if we could apply the regex only when running on that mode

Already works like that:

https://github.com/getsentry/sentry-dotnet/blob/089648edb31ad13871338754234a88b3c962ae05/src/Sentry/Extensibility/SentryStackTraceFactory.cs#L199_L204

```c#
            if (demangle && _options.StackTraceMode != StackTraceMode.Enhanced)
            {
                DemangleAsyncFunctionName(frame);
                DemangleAnonymousFunction(frame);
                DemangleLambdaReturnType(frame);
            }
```

test/Sentry.Tests/Internals/SentryStackTraceFactoryTests.cs
```diff
@@ -45,6 +45,28 @@ public void Create_NoExceptionAndAttachStackTraceOptionOnWithOriginalMode_Curren
...
+    [InlineData(StackTraceMode.Original, "Create_Async_CurrentStackTrace { <lambda> }")]
+    [InlineData(StackTraceMode.Enhanced, "void SentryStackTraceFactoryTests.Create_Async_CurrentStackTrace(StackTraceMode mode, string method)+() => { }")]
+    public void Create_Async_CurrentStackTrace(StackTraceMode mode, string method)
+    {
+        _fixture.SentryOptions.AttachStacktrace = true;
+        _fixture.SentryOptions.StackTraceMode = mode;
+        var sut = _fixture.GetSut();
+
+        SentryStackTrace stackTrace = null!;
+        Task.Run(() => stackTrace = sut.Create()).Wait();
```
mattjohnsonpint, line 58: This isn't really async.  Perhaps you should change the test method from void to `async Task`, and do something like this instead?

```csharp
await Task.Run(async () =>
{
    await Task.Yield();
    return stackTrace = sut.Create();
});
```

Of course, that will change all your expected values.

mattjohnsonpint: Would that still test the case you were trying to capture?

vaind:
Can this be merged or are there any reservations still?

mattjohnsonpint:
Taking another look...

test/Sentry.Tests/Internals/SentryStackTraceFactoryTests.cs
mattjohnsonpint: I was able to update the regex to fix that.  Thanks.

mattjohnsonpint: Actually, after discussing with Bruno, my change would affect grouping and thus can't be done.  Yours is fine though, as we haven't seen that before until Unity.

PR #1744, opened 2022-06-24
Misc improvements

mattjohnsonpint:
A few minor improvements, mostly in tests.

#skip-changelog

### DIFF
--- a/src/Sentry/Internal/SerializableExtensions.cs
+++ b/src/Sentry/Internal/SerializableExtensions.cs
@@ -22,7 +22,10 @@ namespace Sentry.Internal
 #endif
             {
                 await serializable.SerializeAsync(stream, logger, cancellationToken).ConfigureAwait(false);
-                return Encoding.UTF8.GetString(stream.ToArray());
+
+                stream.Seek(0, SeekOrigin.Begin);
+                using var reader = new StreamReader(stream);
+                return await reader.ReadToEndAsync().ConfigureAwait(false);
             }
         }
 
@@ -30,7 +33,10 @@ namespace Sentry.Internal
         {
             using var stream = new MemoryStream();
             serializable.Serialize(stream, logger);
-            return Encoding.UTF8.GetString(stream.ToArray());
+
+            stream.Seek(0, SeekOrigin.Begin);
+            using var reader = new StreamReader(stream);
+            return reader.ReadToEnd();
         }
     }
 }
--- a/test/Sentry.Tests/Internals/Http/HttpTransportTests.cs
+++ b/test/Sentry.Tests/Internals/Http/HttpTransportTests.cs
@@ -10,6 +10,8 @@ namespace Sentry.Tests.Internals.Http;
 
 public class HttpTransportTests
 {
+    private static readonly IDiagnosticLogger DebugLogger = new TraceDiagnosticLogger(SentryLevel.Debug);
+
     [Fact]
     public async Task SendEnvelopeAsync_CancellationToken_PassedToClient()
     {
@@ -278,7 +280,7 @@ public class HttpTransportTests
             new SentryOptions
             {
                 Dsn = DsnSamples.ValidDsnWithSecret,
-                DiagnosticLogger = new TraceDiagnosticLogger(SentryLevel.Debug),
+                DiagnosticLogger = DebugLogger,
                 SendClientReports = false,
                 Debug = true
             },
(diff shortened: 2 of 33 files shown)


test/Sentry.Tests/Internals/Http/HttpTransportTests.cs
```diff
@@ -10,6 +10,8 @@ namespace Sentry.Tests.Internals.Http;
 
 public class HttpTransportTests
 {
+    private static readonly IDiagnosticLogger DebugLogger = new TraceDiagnosticLogger(SentryLevel.Debug);
```
bruno-garcia, line 13: The school of thought that drove the author or these tests is that tests should be as independent of external state as possible. Trade off is there's more duplication but often moving things around in this "test classes state" doesn't cause other tests to break.

Obviously harder to vary behavior by "centrally" changing setup. You'd need to find and replace or manually change each occurrence. Since in this pr no test changed the instance, neither a different type of diag logger nor a different level, I personally don't see a reason to make this change. That said, tests I write indeed have central state I mutate before each test so I'm most certainly not against the approach used here.

test/Sentry.Tests/Internals/MemoryInfoTests.WriteTo.Core3_1.verified.txt
```diff
@@ -0,0 +1,8 @@
+{
+  allocated_bytes: 1,
+  fragmented_bytes: 2,
+  heap_size_bytes: 3,
+  high_memory_load_threshold_bytes: 4,
```
bruno-garcia, line 5: Could this become flaky?

bruno-garcia (approved):
Lgtm! 2 nits/questions

test/Sentry.Tests/Internals/MemoryInfoTests.cs
```diff
@@ -1,27 +1,24 @@
...
-        writer.Flush();
-        var json = Encoding.UTF8.GetString(stream.ToArray());
-        Assert.NotNull(json);
-        Assert.NotEmpty(json);
-        //Validate json
-        var serializer = new Newtonsoft.Json.JsonSerializer();
-        serializer.Deserialize(new Newtonsoft.Json.JsonTextReader(new StringReader(json)));
+        var settings = new VerifySettings();
+        settings.UniqueForTargetFrameworkAndVersion();
+        return VerifyJson(json, settings);
```
SimonCropp, line 20:
```suggestion
        return VerifyJson(json).UniqueForTargetFrameworkAndVersion();
```

test/Sentry.Tests/Internals/MemoryInfoTests.WriteTo.Core3_1.verified.txt
SimonCropp: @bruno-garcia i think this test is hard coded values and testing the serialization

test/Sentry.Tests/Internals/Http/HttpTransportTests.cs
SimonCropp: is the instance of IDiagnosticLogger is not important for the tests. should we have a static no-op instance of IDiagnosticLogger  ?

mattjohnsonpint: `TraceDiagnosticLogger` is stateless other than the minimum level, which is the constant passed in the constructor, and it's immutable.  It just forwards everything to `Trace.Writeline`, so there's really no point in multiple instances floating around.  It was just a quick cleanup.

That said, I'm not sure why we're using it here.  xUnit stopped supporting `Trace` and `Debug` in v2.x. We should actually be using `ITestOutputHelper`.  See https://xunit.net/docs/capturing-output.  I'll make another adjustment for that.

test/Sentry.Tests/Internals/MemoryInfoTests.WriteTo.Core3_1.verified.txt
mattjohnsonpint: Yes. Just the values.  The previous test caught my eye because it was doing its own stream processing and then oddly calling Newtonsoft (which we only import transitively), and then it wasn't actually doing anything with the output. So it could have been sending garbage as long as it could serialize and deserialize and we'd never have known it.  Again, it was just a quick fix.  Nothing all that important here.

src/Sentry/Internal/SerializableExtensions.cs
```diff
@@ -22,15 +22,21 @@ await using (stream.ConfigureAwait(false))
...
+                using var reader = new StreamReader(stream);
+                return await reader.ReadToEndAsync().ConfigureAwait(false);
             }
         }
 
         public static string SerializeToString(this ISerializable serializable, IDiagnosticLogger logger)
         {
             using var stream = new MemoryStream();
             serializable.Serialize(stream, logger);
-            return Encoding.UTF8.GetString(stream.ToArray());
```
mattjohnsonpint, line 33: Just for clarity - this is the only real useful change.  We shouldn't call `MemoryStream.ToArray` because it creates an intermediate byte array.  So there were three copies of the data - one in the stream, one in the byte array, and one in the output string.  We only need two.

PR #1747, opened 2022-06-25
why the fixture / sut pattern

SimonCropp:
so i dont like the fixture/sut pattern. when reading a test i need so continually scan back to the top of the page to see what the fixture is doing. and in many cases it is doing nothing useful, often only an empty new. Also they set a field that some test can than mutate. 

IMO it is better for each test to, in most cases, to own its own construction + setup.

### DIFF
--- a/test/Sentry.Testing/DsnSamples.cs
+++ b/test/Sentry.Testing/DsnSamples.cs
@@ -16,3 +16,20 @@ public static class DsnSamples
     /// </summary>
     public const string InvalidDsn = "https://d4d82fc1c2c4032a83f3a29aa3a3aff@fake-sentry.io:65535/";
 }
+
+public class ValidDsnNoSecretOptions : SentryOptions
+{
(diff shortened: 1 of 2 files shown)


test/Sentry.Testing/DsnSamples.cs
```diff
@@ -16,3 +16,20 @@ public static class DsnSamples
...
+        Dsn = DsnSamples.ValidDsnWithoutSecret;
+    }
+}
+
+public class ValidDsnOptions : SentryOptions
+{
+    public static ValidDsnOptions Instance = new ValidDsnOptions();
+    public ValidDsnOptions()
+    {
+        Dsn = DsnSamples.ValidDsnWithSecret;
```
bruno-garcia, line 33: Secret has been deprecated so many years ago that we can just delete it at this point. The SDK requires the `/envelope` endpoint so doesn't support old Sentry's that would require a secret in the DSN.

PR #1748, opened 2022-06-25
dont allow SentryDiagnosticListenerIntegration to be added multiple times

### LINKED ISSUE #1707: DiagnosticSource can easily be added twice
### Package

Sentry.DiagnosticSource

### .NET Flavor

.NET

### .NET Version

all

### OS

Other

### SDK Version

all

_No response_

### Steps to Reproduce

Install Sentry.AspNet

Call AddAspNet and AddDiagnosticSourceIntegration

### Expected Result

Only add it once. 

### Actual Result

The integration is added twice.

https://github.com/getsentry/sentry-dotnet/blob/89cad6dd25fd9cb0df4051c7414fba8263b4f1ac/src/Sentry.DiagnosticSource/SentryOptionsDiagnosticExtensions.cs#L18

SimonCropp:
fixes #1707

### DIFF
--- a/src/Sentry.DiagnosticSource/SentryOptionsDiagnosticExtensions.cs
+++ b/src/Sentry.DiagnosticSource/SentryOptionsDiagnosticExtensions.cs
@@ -1,5 +1,6 @@
 using System.ComponentModel;
 using System.Linq;
+using Sentry.Extensibility;
 using Sentry.Internals.DiagnosticSource;
 
 namespace Sentry
@@ -15,7 +16,16 @@ namespace Sentry
         /// </summary>
         /// <param name="options">The Sentry options.</param>
         public static void AddDiagnosticSourceIntegration(this SentryOptions options)
-            => options.AddIntegration(new SentryDiagnosticListenerIntegration());
+        {
+            if (options.Integrations != null &&
+                options.Integrations.OfType<SentryDiagnosticListenerIntegration>().Any())
+            {
+                options.LogInfo($"{nameof(SentryDiagnosticListenerIntegration)} has already been added. The second call to {nameof(AddDiagnosticSourceIntegration)} will be ignored.");
+                return;
+            }
+
+            options.AddIntegration(new SentryDiagnosticListenerIntegration());
+        }
 
         /// <summary>
         /// Disables the integrations with Diagnostic source.
--- a/src/Sentry.DiagnosticSource/SentryOptionsDiagnosticExtensions.cs
+++ b/src/Sentry.DiagnosticSource/SentryOptionsDiagnosticExtensions.cs
@@ -20,7 +20,7 @@ namespace Sentry
             if (options.Integrations != null &&
                 options.Integrations.OfType<SentryDiagnosticListenerIntegration>().Any())
             {
-                options.LogInfo($"{nameof(SentryDiagnosticListenerIntegration)} has already been added. The second call to {nameof(AddDiagnosticSourceIntegration)} will be ignored.");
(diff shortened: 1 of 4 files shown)


src/Sentry.DiagnosticSource/SentryOptionsDiagnosticExtensions.cs
```diff
@@ -15,7 +16,16 @@ public static class SentryOptionsDiagnosticExtensions
         /// </summary>
         /// <param name="options">The Sentry options.</param>
         public static void AddDiagnosticSourceIntegration(this SentryOptions options)
-            => options.AddIntegration(new SentryDiagnosticListenerIntegration());
+        {
+            if (options.Integrations != null &&
+                options.Integrations.OfType<SentryDiagnosticListenerIntegration>().Any())
```
SimonCropp, line 21: @bruno-garcia @mattjohnsonpint  should i refactor this into a helper method in options?

src/Sentry.DiagnosticSource/SentryOptionsDiagnosticExtensions.cs
```diff
@@ -15,7 +16,16 @@ public static class SentryOptionsDiagnosticExtensions
...
+            {
+                options.LogInfo($"{nameof(SentryDiagnosticListenerIntegration)} has already been added. The second call to {nameof(AddDiagnosticSourceIntegration)} will be ignored.");
```
mattjohnsonpint, line 23:
```suggestion
                options.LogWarning($"{nameof(SentryDiagnosticListenerIntegration)} has already been added. The second call to {nameof(AddDiagnosticSourceIntegration)} will be ignored.");
```

src/Sentry.DiagnosticSource/SentryOptionsDiagnosticExtensions.cs
mattjohnsonpint, line 21:
```suggestion
            if (options.Integrations?.OfType<SentryDiagnosticListenerIntegration>().Any() is true)
```

mattjohnsonpint: I don't think that's necessary.

src/Sentry/SentryOptionsExtensions.cs
```diff
@@ -87,9 +87,16 @@ public static void DisableNetFxInstallationsIntegration(this SentryOptions optio
...
-                ? options.Integrations.Concat(new[] { integration }).ToArray()
-                : new[] { integration };
+        {
+            if (options.Integrations == null)
+            {
+                options.Integrations = new[] {integration};
+            }
+            else
+            {
+                options.Integrations = options.Integrations.Concat(new[] {integration}).ToArray();
```
mattjohnsonpint, line 97: Just a thought, but maybe we should leave `AddDiagnosticSourceIntegration` as it was, and check here more generally instead?   I mean, can we safely say that there should not be more than one integration of the same type?

SimonCropp: @mattjohnsonpint i asked @bruno-garcia the same question. he said it is valid to have more than one integration by type, but with different settings in each integration instance

bruno-garcia: Where did that communication happen?

The example was: `new FileWatcherIntegration("path1"), new FileWatcherIntegratoin("path2")`

mattjohnsonpint: LGTM

mattjohnsonpint (approved):

PR #1749, opened 2022-06-25
Log Warning when secret is detected in DSN

SimonCropp:
quoting @bruno-garcia 

> Secret has been deprecated so many years ago that we can just delete it at this point. The SDK requires the /envelope endpoint so doesn't support old Sentry's that would require a secret in the DSN

https://github.com/getsentry/sentry-dotnet/pull/1747#discussion_r906612700


test/Sentry.Tests/Protocol/DsnTests.cs
```diff
@@ -133,21 +126,15 @@ public void Ctor_NullDsn_ThrowsArgumentNull()
     }
 
     [Fact]
-    public void TryParse_SampleValidDsnWithoutSecret_Succeeds()
-    {
-        Assert.NotNull(Dsn.TryParse(DsnSamples.ValidDsnWithoutSecret));
-    }
-
-    [Fact]
-    public void TryParse_SampleValidDsnWithSecret_Succeeds()
```
mattjohnsonpint, line 142: Shouldn't we still keep the Parse and TryParse tests? Though we don't use the secret ourself, it's still valid input from the user.

mattjohnsonpint:
Perhaps we should add some logic that logs a warning message if they pass a DSN that includes a secret?

bruno-garcia (approved):
Much simpler. Thanks!

test/Sentry.Tests/Protocol/DsnTests.cs
SimonCropp: @bruno-garcia is this correct. from your statement that spured this, i cant tell if secret is still valid?

bruno-garcia: we don't need to validate secret anymore. Test can be deleted.

bruno-garcia:
> Perhaps we should add some logic that logs a warning message if they pass a DSN that includes a secret?

There's no problem with having the secret, it's just not going to be used for anything on the server.

test/Sentry.Tests/Protocol/DsnTests.cs
SimonCropp: re added the tests for with secret

SimonCropp:
i added a warning when secret is detected

mattjohnsonpint:
Fixed merge with #1758

PR #1750, opened 2022-06-25
Catch Android/MAUI permission exceptions

mattjohnsonpint:
Battery and Network device info is not available unless the application has specific permissions.  Swallow those permission errors.

Neither is reading the process start time.  We may be able to find a better way to collect this later, but for now just use "Fast" mode instead of "Best".


bruno-garcia (approved):
I wonder if it's worth compiling out the Best on Android. Or compile an Obsolete so it warns if they use it?

mattjohnsonpint:
> I wonder if it's worth compiling out the Best on Android. Or compile an Obsolete so it warns if they use it?

Makes sense.  Added a commit for that.

PR #1753, opened 2022-06-26
Enable offline caching in MAUI

mattjohnsonpint:
- Enables offline caching in MAUI
- Adds `WebException` to the exceptions that indicate network failure in the caching transport, as that's what thrown by Android when there is no Internet on the device.
- Adds an optional "network status listener" interface and option, which allows the caching transport to be more intelligent about whether it will attempt to send items from the cache to Sentry, based on network status.
- Tests for all the above.


src/Sentry.Maui/Internal/MauiNetworkStatusListener.cs
```diff
@@ -0,0 +1,68 @@
...
+        catch (PermissionException)
+        {
+            _hasNetworkStatusPermission = false;
+            options.DiagnosticLogger?.LogDebug("No network status permission.  Will assume device is online.");
+        }
+    }
+
+    public bool Online => !_hasNetworkStatusPermission || TreatAsOnline(_connectivity.NetworkAccess);
+
+    public async Task WaitForNetworkOnlineAsync(CancellationToken cancellationToken)
```
bruno-garcia, line 29: Can we verify the behavior when the token provided is already cancelled? Since we call `Register` after it was cancelled, it won't fire so I wonder if it'll hang

mattjohnsonpint: Good point.  Though in that case I'll just return immediately.

src/Sentry.Maui/Internal/MauiNetworkStatusListener.cs
```diff
@@ -0,0 +1,68 @@
...
+                tcs.TrySetResult();
+            }
+        }
+
+        cancellationToken.Register(() =>
+        {
+            _connectivity.ConnectivityChanged -= OnConnectivityChanged;
+            tcs.TrySetCanceled();
+        });
+
```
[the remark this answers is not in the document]
mattjohnsonpint: Actually, you're right that there's still a race condition in my code.  If the token was already cancelled, the event handler is deregistered first (a no-op), and then it is registered and never actually de-registered.

I'll fix it, and add a test.

bruno-garcia (approved):

PR #1755, opened 2022-06-27
Build Android on Linux

mattjohnsonpint:
- Build Android targets on Linux
- Remove OS solution filters (.slnf files).  We don't need them now.
- Update build scripts and yaml

#skip-changelog


bruno-garcia:
> * Remove OS solution filters (.slnf files).  We don't need them now.

Do we need them for iOS? And Windows-only stuff like WinForms and WPF sample? #1036

.github/workflows/build.yml
```diff
@@ -61,38 +54,29 @@ jobs:
...
-          run: dotnet workload install maui-android maui-windows
+          run: 'dotnet workload install maui-android maui-windows --temp-dir "${{ runner.temp }}"'
         - name: Install .NET Workloads (macOS)
           if: startsWith(matrix.os, 'macos')
-          run: dotnet workload install maui-android maui-ios maui-maccatalyst
+          run: 'dotnet workload install maui-android maui-ios maui-maccatalyst --temp-dir "${{ runner.temp }}"'
 
         - name: Build
-          run: dotnet build ${{ matrix.slnf }} -c Release /p:CopyLocalLockFileAssemblies=true
+          run: dotnet build -c Release /p:CopyLocalLockFileAssemblies=true
```
bruno-garcia, line 71: Could we set this property here: https://github.com/getsentry/sentry-dotnet/blob/main/Directory.Build.props#L20-L22

This way we don't need to remember to pass it in different places: https://github.com/getsentry/sentry-dotnet/search?q=p%3ACopyLocalLockFileAssemblies

mattjohnsonpint: No, because we don't want to do that in the CodeQL build.

mattjohnsonpint:
> > * Remove OS solution filters (.slnf files).  We don't need them now.
> 
> Do we need them for iOS? And Windows-only stuff like WinForms and WPF sample? #1036

I'm investigating into another way to deal with that.  I don't like using solution filters for this.

SimonCropp (approved):

PR #1756, opened 2022-06-27
Don't allow newlines in diagnostic logger messages

mattjohnsonpint:
Addresses CodeQL alert https://github.com/getsentry/sentry-dotnet/pull/1755/checks?check_run_id=7079180169

This appears to have been raised twice before:
- https://github.com/getsentry/sentry-dotnet/security/code-scanning/5
- https://github.com/getsentry/sentry-dotnet/security/code-scanning/6

Not sure why it was marked fixed and suddenly reappeared, but doesn't look like it was ever actually addressed.

### DIFF
--- a/src/Sentry/Infrastructure/ConsoleDiagnosticLogger.cs
+++ b/src/Sentry/Infrastructure/ConsoleDiagnosticLogger.cs
@@ -1,5 +1,4 @@
 using System;
-using Sentry.Extensibility;
 
 namespace Sentry.Infrastructure
 {
@@ -9,25 +8,16 @@ namespace Sentry.Infrastructure
     /// <remarks>
     /// The default logger, usually replaced by a higher level logging adapter like Microsoft.Extensions.Logging.
     /// </remarks>
-    public class ConsoleDiagnosticLogger : IDiagnosticLogger
+    public class ConsoleDiagnosticLogger : DiagnosticLogger
     {
-        private readonly SentryLevel _minimalLevel;
-
         /// <summary>
         /// Creates a new instance of <see cref="ConsoleDiagnosticLogger"/>.
         /// </summary>
-        public ConsoleDiagnosticLogger(SentryLevel minimalLevel) => _minimalLevel = minimalLevel;
+        public ConsoleDiagnosticLogger(SentryLevel minimalLevel) : base(minimalLevel)
+        {
+        }
 
-        /// <summary>
-        /// Whether the logger is enabled to the defined level.
-        /// </summary>
-        public bool IsEnabled(SentryLevel level) => level >= _minimalLevel;
-
-        /// <summary>
-        /// Log message with level, exception and parameters.
-        /// </summary>
-        public void Log(SentryLevel logLevel, string message, Exception? exception = null, params object?[] args)
-            => Console.Write($@"{logLevel,7}: {string.Format(message, args)}
-{exception}");
+        /// <inheritdoc />
+        protected override void LogMessage(string message) => Console.WriteLine(message);
     }
 }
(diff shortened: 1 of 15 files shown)


src/Sentry/Infrastructure/ConsoleDiagnosticLogger.cs
```diff
@@ -9,25 +8,16 @@ namespace Sentry.Infrastructure
...
-
         /// <summary>
         /// Creates a new instance of <see cref="ConsoleDiagnosticLogger"/>.
         /// </summary>
-        public ConsoleDiagnosticLogger(SentryLevel minimalLevel) => _minimalLevel = minimalLevel;
-
-        /// <summary>
-        /// Whether the logger is enabled to the defined level.
-        /// </summary>
-        public bool IsEnabled(SentryLevel level) => level >= _minimalLevel;
```
SimonCropp, line 24: was this not used? and isnt this a breaking change?

SimonCropp: ahh i see. base class. ignore me

src/Sentry/Infrastructure/DiagnosticLogger.cs
```diff
@@ -0,0 +1,48 @@
+using System;
+using System.Linq;
+using Sentry.Extensibility;
+
+namespace Sentry.Infrastructure
+{
+    /// <summary>
+    /// Base class for diagnostic loggers.
+    /// </summary>
+    public abstract class DiagnosticLogger : IDiagnosticLogger
```
SimonCropp, line 10: i think moving methods to a base class is a runtime breaking change.

src/Sentry/Infrastructure/DiagnosticLogger.cs
```diff
@@ -0,0 +1,48 @@
...
+
+        /// <summary>
+        /// Log message with level, exception and parameters.
+        /// </summary>
+        public void Log(SentryLevel logLevel, string message, Exception? exception = null, params object?[] args)
+        {
+            // Note, linefeed and newline chars are removed to guard against log injection attacks.
+            // See https://github.com/getsentry/sentry-dotnet/security/code-scanning/5
+
+            var formattedMessage = new string(string.Format(message, args)
```
SimonCropp, line 32: for a log message of 
```
The log
message
```

wont this result in `The logmessage` ? Is that what we want?

src/Sentry/Infrastructure/DiagnosticLogger.cs
mattjohnsonpint: Not according to https://docs.microsoft.com/dotnet/core/compatibility/#members

> - ✔️ ALLOWED: Moving a member into a class higher in the hierarchy than the type from which it was removed

Unless I'm misinterpreting?

src/Sentry/Infrastructure/DiagnosticLogger.cs
mattjohnsonpint: Do you think it would be better to do something like replacing `"\n"` with `"\\n"`, etc.?

src/Sentry/Infrastructure/DiagnosticLogger.cs
SimonCropp: consider me educated.

src/Sentry/Infrastructure/DiagnosticLogger.cs
SimonCropp: i was thinking replacing with a space would be good enough

src/Sentry/Infrastructure/ConsoleDiagnosticLogger.cs
mattjohnsonpint: np. :)

src/Sentry/Infrastructure/DiagnosticLogger.cs
mattjohnsonpint: Done. Thanks.

SimonCropp (approved):

SimonCropp:
:shipit:

PR #1762, opened 2022-06-30
Test projects cleanup

mattjohnsonpint:
- Refactored to remove ASP.NET Core dependencies from Sentry.Testing, which unblocks #1703
- Re-enables .NET Framework tests for ASP.NET Core 2.1.  We were testing the OS using a deprecated technique, so these tests weren't actually running before - even on Windows.  They are now, and restoring them also generated a missing API verification file.
- Updates all the tests to use the latest minor release of their various dependencies.
  - Note, we need to come up with a better way to stay evergreen with test dependencies.  While we want our libraries to use minimum compatible versions, the *tests* should use latest.

#skip-changelog


test/Sentry.AspNetCore.Tests/Sentry.AspNetCore.Tests.csproj
```diff
@@ -1,12 +1,7 @@
...
 
-  <!-- Running these tests on Mono just hangs -->
-  <PropertyGroup Condition="'$(OS)' == 'Windows_NT'">
-    <TargetFrameworks>net461;$(TargetFrameworks)</TargetFrameworks>
-  </PropertyGroup>
-
   <PropertyGroup>
-    <TargetFrameworks>net6.0;net5.0;netcoreapp3.1;netcoreapp2.1</TargetFrameworks>
+    <TargetFrameworks>net6.0;net5.0;netcoreapp3.1;netcoreapp2.1;net461</TargetFrameworks>
   </PropertyGroup>
```
mattjohnsonpint, line 5: Note that `Condition="'$(OS)' == 'Windows_NT'"` is unsupported.  The correct way is `Condition="$([MSBuild]::IsOSPlatform('windows'))"`.  In this case though, we can just include it directly.  The comment about the tests hanging on Mono appears to be not the case anymore.

test/Sentry.EntityFramework.Tests/Sentry.EntityFramework.Tests.csproj
```diff
@@ -5,12 +5,12 @@
   </PropertyGroup>
 
   <!-- Running these tests on Mono fail -->
-  <PropertyGroup Condition="'$(OS)' == 'Windows_NT'">
```
mattjohnsonpint, line 8: Here's another case where we were checking incorrectly.  If `OS` isn't defined, then the tests don't run for this target at all on any OS.

test/Sentry.EntityFramework.Tests/Sentry.EntityFramework.Tests.csproj
```diff
@@ -5,12 +5,12 @@
...
+  <PropertyGroup Condition="$([MSBuild]::IsOSPlatform('windows'))">
```
mattjohnsonpint, line 8: Here's another case where we were checking incorrectly.  If `OS` wasn't defined, then the tests din't run for this target at all on any OS.

test/Sentry.AspNetCore.Tests/Sentry.AspNetCore.Tests.csproj
```diff
@@ -1,12 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
-  <!-- Running these tests on Mono just hangs -->
-  <PropertyGroup Condition="'$(OS)' == 'Windows_NT'">
-    <TargetFrameworks>net461;$(TargetFrameworks)</TargetFrameworks>
-  </PropertyGroup>
-
   <PropertyGroup>
-    <TargetFrameworks>net6.0;net5.0;netcoreapp3.1;netcoreapp2.1</TargetFrameworks>
+    <TargetFrameworks>net6.0;net5.0;netcoreapp3.1;netcoreapp2.1;net461</TargetFrameworks>
```
mattjohnsonpint, line 4: We can just include `net461` directly now.  The comment about the tests hanging on Mono appears to be not the case anymore.  But note that it *wasn't* being included on Windows before because of the way we were checking.

`Condition="'$(OS)' == 'Windows_NT'"` is unsupported.  The correct way is `Condition="$([MSBuild]::IsOSPlatform('windows'))"`.

SimonCropp (approved):

PR #1763, opened 2022-07-01
Add tests for stack of a generic method 

SimonCropp:
#skip-changelog


SimonCropp:
@mattjohnsonpint @bruno-garcia ok i need soe help on this one

u can see in the output the failing snapshot. for the enhanced stack 
```
Compare Result:   {
-   FileName: {ProjectDirectory}Internals/SentryStackTraceFactoryTests.cs,
-   Function: void SentryStackTraceFactoryTests.GenericMethodThatThrows<T>(T value),
+   FileName: ,
+   Function: void SentryStackTraceFactoryTests.GenericMethodThatThrows<int>(int value),
```

So this means that on Windows

 * the stack frame has the filename
 * the function has the genetic `<T`>

On mac and linux

 * the stack frame has no filename
 * the function has the generic `<int>`

I debugged into the windows one, and it seems there is no way to get the `<int>` from the stack trace. So given this would effect grouping, do we want to change it to change the mac/linux implementation to match?

Can one of u debug locally and see if u get a FileName on a mac?

mattjohnsonpint:
Appears to be only happening on Mono.  Works for all other targets.  I added a skip for now.  We can investigate further later if needed.

mattjohnsonpint (approved):

SimonCropp:
@mattjohnsonpint thanks

PR #1764, opened 2022-07-02
Verify on device

SimonCropp:
#skip-changelog


SimonCropp:
@vaind  i fixed the OS issue in Verify. but now the tests fail with no obvious failure message. Any ideas?

vaind:
> @vaind i fixed the OS issue in Verify. but now the tests fail with no obvious failure message. Any ideas?

that's not caused by this change - already happens on the main MAUI test PR (and no, I have no idea why).

What do you think about updating this on the main branch instead, I'll pull it into the maui test afterwards

SimonCropp:
yep i can push it to main branch today . it shouldn’t cause any conflicts for u

PR #1767, opened 2022-07-03
remove System.Runtime.Versioning.TargetFramework  from api snapshot

SimonCropp:
After an update to some runtime, i noticed the value for `System.Runtime.Versioning.TargetFramework` had a bug fix that added the framework name to the attribute. but given we dont care about that attribute for API purposes, i removed it

#skip-changelog


bruno-garcia (approved):
I wouldn't mind if that stayed either

SimonCropp:
@bruno-garcia if u re-run those API tests on a mac,  do u get the `FrameworkDisplayName=""` having a value or being empty? My concern was since none of the mac devs noticed it, it was a change only on windows

PR #1780, opened 2022-07-15
feat: Pin TFM for Unity

bitsandfoxes:
There might be a smarter/better condition but I was not able to read properties from the actual Unity `.props` inside the `Sentry.csproj`. But when building it is aware of which solution it is part of.

#skip-changelog


mattjohnsonpint:
@bitsandfoxes  - I updated this to just change the tfm. Can you try building `Sentry.Unity` against the latest commit here?  I just want to check that `$(SolutionName)` is passed through correctly.

If not, another approach would be to add a custom parameter to the unity build such as `/p:SentryUnity=true` and test for that instead.

mattjohnsonpint (approved):

PR #1782, opened 2022-07-15
Fix test failing on Windows

mattjohnsonpint:
For some unknown reason, my local Windows machine was giving project paths with `c:\` instead of `C:\`, which was causing `FileNameShouldBeRelative` to fail.  I'm still hunting down why, but this will prevent it.

Related to #1739 

#skip-changelog

### DIFF
--- a/src/Sentry/AttributeReader.cs
+++ b/src/Sentry/AttributeReader.cs
@@ -1,5 +1,7 @@
-﻿using System.Linq;
+﻿using System.IO;
+using System.Linq;
 using System.Reflection;
+using System.Runtime.InteropServices;
 
 namespace Sentry
 {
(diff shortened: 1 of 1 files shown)


mattjohnsonpint:
The test only failed on `net6.0`, so I thought it might be because I had 6.0.302 locally, while CI is still on 6.0.301.

So I uninstalled 6.0.302 and installed 6.0.301 (since they no longer let you install them both side-by-side).  The test passed.

But then I re-installed 6.0.302, and I expected the test to fail, but it still passed - even with a full git clean and rebuild.

So either I had a bad install, or it reset something else? Not sure.  Anyway, I still think we should merge this.

SimonCropp:
IMO it would be easier to do a case insensitive startswith later in the processing

mattjohnsonpint:
It's used here (on line 190)

https://github.com/getsentry/sentry-dotnet/blob/f3c7949023581f566067207157779c21cda42ed5/src/Sentry/Extensibility/SentryStackTraceFactory.cs#L181-L194

Unfortunately, `string.Replace` doesn't have an overload with comparer parameter in `netstandard2.0` or `net461`.  We would need a shim.

mattjohnsonpint:
Either way is fine with me, if you want to take another go at it, @SimonCropp ?

mattjohnsonpint:
Close in favor of #1791

PR #1784, opened 2022-07-16
Fix slow tests and race condition

mattjohnsonpint:
This should speed up several of the tests.  It also improves some test debug output, and fixes a minor race condition in the caching transport (in the handling of `InitCacheFlushTimeout`) that was only noticeable when shutting down very close to starting (like we do in tests).


mattjohnsonpint:
Looks like a have a little more to do for `Init_WithCache_BlocksUntilExistingCacheIsFlushed`.  Leaving as a draft for now.

SimonCropp (approved):

PR #1785, opened 2022-07-19
feat: Capture Debug Images and StackFrame fields for Portable PDB symbolication

Swatinem:
This adds the new event properties defined in [RFC 0013](https://github.com/getsentry/rfcs/pull/13), namely InstructionAddress / FunctionId + Relative Indexing, and a list of `pe_dotnet` DebugImages.

The new information allows symbolicating .NET stack traces on symbolicator, using uploaded Portable PDB debug files.

Example events that were captured with this PR and symbolicated serverside:
- https://sentry.io/organizations/sentry-sdks/issues/3323330137/events/07c45ddfa9bd4ebbbcf0b56b2a7d0b4c/?project=5428537
- https://sentry.io/organizations/sentry-sdks/issues/2442781715/events/a4f4c59a7d1a4b06bd98c0d1dcfa7e8c/?project=5428537

The second event might be an example where it could be possible to load debug symbols from the Microsoft Symbol Server which is not possible yet, but we might add this later.

---

I open this PR with the hope of either delegating it to the .NET team, or at the very least get some dedicated mentoring to finish it up.
The problem I see right now are:
- The PR depends on default interface methods which I believe we can’t use because of backwards compatibility?
- The way that stack frame factories and event / exception processors are implemented means that we currently have no way to process frames while at the same time have "atomic" access to the list of debug images. The relative indexing that we need depends on having full control over the index of the referenced debug images in the DebugImages list. Currently the different factories and processors can change that list as they please, which makes it quite fragile.

### DIFF
--- a/src/Sentry/DebugImage.cs
+++ b/src/Sentry/DebugImage.cs
@@ -33,6 +33,11 @@ public sealed class DebugImage : IJsonSerializable
     public string? DebugId { get; set; }
 
     /// <summary>
+    /// Checksum of the companion debug file.
+    /// </summary>
+    public string? DebugChecksum { get; set; }
+
+    /// <summary>
     /// Path and name of the debug companion file.
     /// </summary>
     public string? DebugFile { get; set; }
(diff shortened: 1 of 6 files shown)


vaind:
I've rebased this PR but in the end decided to replace the PR as the approach may end up different, due to the lack of default interface method support this PR relied on. Also, want to keep the new PR as a draft to avoid spamming everyone before it's ready.

replaced by #2050

PR #1788, opened 2022-07-20
fix: Use the OFFSET_UNKNOWN to check for valid ILOffset

Swatinem:
A value of 0 is perfectly valid as ILOffset, the sentinel value is OFFSET_UNKNOWN
which we now explicitly check for.

#skip-changelog I doubt this is worth a changelog entry :-D


bruno-garcia (approved):

bruno-garcia:
thanks @Swatinem !

mattjohnsonpint:
Yes, thanks!  Makes sense, because [`OFFSET_UNKNOWN`](https://docs.microsoft.com/dotnet/api/system.diagnostics.stackframe.offset_unknown) == `-1`.

This would be worthy of a unit test, IMHO.

PR #1794, opened 2022-07-21
Don't dispose hub in unhandled exception handler

### LINKED ISSUE #1772: Exceptions on background thread are not handled and reported correctly
### Environment

**How do you use Sentry?**
sentry.io

**Which version of the SDK?**
0.20.1

**How did you install the package?**
UPM, git-url

**Which version of Unity?**
2021.3.6f1 LTS

**Is this happening in Unity (editor) or on a player like Android, iOS, Windows?**
Reproduced in Windows Editor and Windows Player

### Steps to Reproduce

1. Create a new empty Unity project.
2. Import sentry-unity 0.20.1 in UPM.
3. Attach the code below to a GameObject
4. Start the game
5. Throw some exceptions in a background thread (click the in-game button).

### Expected Result

Exceptions are reported as events to sentry.io

### Actual Result

Only the first exception is handled and reported correctly.

Something seems to go wrong within the SDK, because after the first exception, subsequent exceptions do not result in any debug logs from Sentry, except a single "Sentry: (Info) Disposing the Hub", which is directly followed by:

> exception inside UnhandledException handler: (null) assembly:C:\Program Files\Unity\Hub\Editor\2021.3.6f1\Editor\Data\MonoBleedingEdge\lib\mono\unityjit-win32\mscorlib.dll type:Exception member:(null)

### Code

```csharp
using System;
using System.Threading;
using UnityEngine;

public class SentryTest : MonoBehaviour
{
	void OnGUI()
	{
		if (GUILayout.Button("Throw Exception On Background Thread"))
		{
			Debug.Log("Pressed button");
			var thread = new Thread(() =>
			{
				Thread.Sleep(1000);
				Debug.Log("Will throw exception");
				throw new Exception("Testing Background Thread Crash 2");
			});
			thread.Start();	
		}
	}
}
```

### Any logs or screenshots

[editor-log.txt](https://github.com/getsentry/sentry-unity/files/9079766/editor-log.txt)

mattjohnsonpint:
Fixes #1772

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 - Fix error with `ConcurrentHashMap` on Android <= 9 ([#1761](https://github.com/getsentry/sentry-dotnet/pull/1761))
 - Minor improvements to `BackgroundWorker` ([#1773](https://github.com/getsentry/sentry-dotnet/pull/1773))
 - Make GzipRequestBodyHandler respect async ([#1776](https://github.com/getsentry/sentry-dotnet/pull/1776))
+- Fix exceptions on background thread not reported in Unity ([#1794](https://github.com/getsentry/sentry-dotnet/pull/1794))
 
 ## 3.19.0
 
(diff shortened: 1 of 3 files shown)


mattjohnsonpint:
@bitsandfoxes - can you try your repro of the issue with this change?  Thanks.

SimonCropp (approved):

vaind:
> can you try your repro of the issue with this change? Thanks.

yes it fixes the issue in Unity

PR #1801, opened 2022-07-22
Update build

mattjohnsonpint:
Now that [the GitHub Actions macOS virtual images have .NET 6 preinstalled](https://github.com/actions/virtual-environments/issues/5896#issuecomment-1191803465), we can simplify our workflow.

Also, dependency caching doesn't quite work the way I thought it would for workloads, so just remove that part for now.

Also fixed the warning about overwriting `trx` files when there are multiple target frameworks.

Also consolidated environment setup into a composite action.

#skip-changelog


.github/workflows/vulnerabilities.yml
```diff
@@ -29,32 +29,19 @@ jobs:
...
-        - name: Setup .NET SDK (macOS)
-          if: startsWith(matrix.os, 'macos')
-          uses: actions/setup-dotnet@v2
-          with:
-            dotnet-version: |
-              2.1.818
-              6.0.x
-        - name: Setup .NET SDK (Ubuntu)
-          if: startsWith(matrix.os, 'ubuntu')
+        - name: Install .NET 2.1 SDK
```
bruno-garcia, line 32: We could use an "include" to share the setup steps across workflows

bruno-garcia:
Simpler is nicer 👌Thanks

SimonCropp (approved):

PR #1811, opened 2022-07-25
Re-enable test parallelization

mattjohnsonpint:
We should now be able to run tests in parallel again.

A few test collections updated (and one test moved).  xUnit uses test collections to control parallelization.  Tests in the same collection run sequentially, and thus won't step on each other.
See https://xunit.net/docs/running-tests-in-parallel

#skip-changelog


mattjohnsonpint:
Need to look into `Sentry.Tests.HubTests.CaptureEvent_NonSerializableContextAndOfflineCaching_CapturesEventWithContextKey` failing in these before continuing.

PR #1821, opened 2022-07-27
Remove extraneous API approvals

mattjohnsonpint:
We shouldn't be generating extra copies of the API approvals for `Sentry` within the `Sentry.DiagnosticSource` tests

#skip-changelog


bruno-garcia (approved):
good catch

Issue #1701, opened 2022-06-08
Failed to extract body - NullReferenceException


msxdan:
### Package

Sentry.AspNetCore

### .NET Flavor

.NET

### .NET Version

6.0.201

### OS

Windows

### SDK Version

3.16.0

### Steps to Reproduce

We could not reproduce the issue, it happens randomly and seldomly but it makes the server crash

### Expected Result

The error should be triggered only once

### Actual Result

It seems that NLog logs properly but when sending the log to Sentry it enters in a infinite loop.

```
2022-06-06 10:25:29.7338|ERROR|App.Controllers.StudyController|Unhandled controller exception: Object reference not set to an instance of an object.|System.NullReferenceException: Object reference not set to an instance of an object.
   at Microsoft.AspNetCore.Http.DefaultHttpRequest.get_BodyReader()
   at Microsoft.AspNetCore.Http.Features.FormFeature.InnerReadFormAsync(CancellationToken cancellationToken)
   at Microsoft.AspNetCore.Mvc.ModelBinding.FormValueProviderFactory.AddValueProviderAsync(ValueProviderFactoryContext context)
   at Microsoft.AspNetCore.Mvc.ModelBinding.CompositeValueProvider.CreateAsync(ActionContext actionContext, IList`1 factories)
   at Microsoft.AspNetCore.Mvc.ModelBinding.CompositeValueProvider.TryCreateAsync(ActionContext actionContext, IList`1 factories)
   at Microsoft.AspNetCore.Mvc.Controllers.ControllerBinderDelegateProvider.<>c__DisplayClass0_0.<<CreateBinderDelegate>g__Bind|0>d.MoveNext()
--- End of stack trace from previous location ---
   at Microsoft.AspNetCore.Mvc.Infrastructure.ControllerActionInvoker.<InvokeInnerFilterAsync>g__Awaited|13_0(ControllerActionInvoker invoker, Task lastTask, State next, Scope scope, Object state, Boolean isCompleted)
   at Microsoft.AspNetCore.Mvc.Infrastructure.ResourceInvoker.<InvokeNextExceptionFilterAsync>g__Awaited|26_0(ResourceInvoker invoker, Task lastTask, State next, Scope scope, Object state, Boolean isCompleted)
2022-06-06 10:25:29.8449|ERROR|Sentry.ISentryClient|Failed to extract body.|System.NullReferenceException: Object reference not set to an instance of an object.
[...]

PR #1824, opened 2022-07-28
Logging loop with n log sentry

### LINKED ISSUE #1741: Logging loop with NLog-Sentry
### Package

Sentry.AspNetCore

### .NET Flavor

.NET Core

### .NET Version

6.0.2

### OS

Windows

### SDK Version

3.17.1

_No response_

### Steps to Reproduce

From what I've read and what I see in the traces, seems that we are trying to log something when the HttpContext has been already disposed and that seems to happen on our app when a detached task fails (using Task.Run()) and tries to log the error.

The problem is not a simple exception that we could try to identify and fix (in this case, some bad data was sent to an API in the background task, a case we are fixing) but that it seems that NLog/Sentry enter in a loop that ends up exhausting app resources.

I attach some log files and traces to show the behavior.

One is a log file from NLog (there are a lot more errors but is essentially the same) and the other is a stack trace of the error from a memory dump.

[stack_trace.txt](https://github.com/getsentry/sentry-dotnet/files/8966424/stack_trace.txt)
[error_log.txt](https://github.com/getsentry/sentry-dotnet/files/8966425/error_log.txt)

### Expected Result

The error to be logged just once. It's acceptable that there's some internal error accessing the context as at this point everything is more or less broken, but with a single log that we can see in Sentry/local log files is enough.

### Actual Result

It seems that NLog tries to log the Sentry error trying to access a disposed object and then tries to send this error to Sentry again, ending up in a loop.

SimonCropp:
* [x] Repro
 * [x] Fixes #1741 #1701

although this is not the exact scenario as #1741 as #1701, it is the same root cause. essentially when inside logging we have anything that causes a log event it results in an infinite loop. in #1701

 * `NLog.SentryTarget.Write` occurs
 * `Scope.Evaluate` occurs
 * triggering `AspNetCore.SentryMiddleware` which is failing
 * triggering   `NLog.SentryTarget.Write`
 * repeat


bruno-garcia (approved):
Thank you for fixing this Simon!

SimonCropp:
i work around this with an asynclocal. can anyone think of a better approach?

mattjohnsonpint:
> i work around this with an asynclocal. can anyone think of a better approach?

Yes.  We can use NLog's built-in context for this.  See my update, and LMK if you agree.  Thanks.

mattjohnsonpint (approved):

SimonCropp:
@mattjohnsonpint much nicer. thanks

snakefoot:
Think it is a bad idea to "abuse" the NLog ScopeContext to protect against bad calling pattern. The NLog ScopeContext is for the user to capture application-state, not for secret infrastructure problems. Just like you hopefully won't "abuse" the Serilog LogContext for injecting special context values.

If you really insist on abusing the NLog NestedDiagnosticsLogicalContext, then you should render the Sentry-LogEvent outside this secret-scope. As layouts depending on [${ndlc}](https://github.com/NLog/NLog/wiki/NDLC-Layout-Renderer) will have side-effects from this usage.

But I recommend that you keep the secret infrastructe problems secret and use a local AsyncLocal-member-variable.

mattjohnsonpint:
@snakefoot that's a good point.

@SimonCropp - let's go back to the async local approach you had.  Thanks.

SimonCropp:
done https://github.com/getsentry/sentry-dotnet/pull/1830

PR #1834, opened 2022-08-02
Enable Scope Sync for iOS

mattjohnsonpint:


src/Sentry/Platforms/iOS/Extensions/UserExtensions.cs
```diff
@@ -0,0 +1,72 @@
...
+            foreach (var item in user.Other)
+            {
+                userData[item.Key] = NSObject.FromObject(item.Value);
+            }
+        }
+
+        return new SentryCocoa.SentryUser
+        {
+            Email = user.Email,
+            UserId = user.Id ?? "",
```
bruno-garcia, line 66: Ia user ID not nullable? I wonder if we'd be sending a user object with an empty if when instead we should be omitting the user object in that case

mattjohnsonpint: https://github.com/getsentry/sentry-cocoa/blob/master/Sources/Sentry/Public/SentryUser.h#L12

mattjohnsonpint: https://develop.sentry.dev/sdk/event-payloads/user/

mattjohnsonpint: Seems like it should be nullable in Sentry Cocoa, but isn't.

bruno-garcia: https://github.com/getsentry/sentry-cocoa/issues/2035

src/Sentry/Platforms/iOS/Extensions/UserExtensions.cs
```diff
@@ -0,0 +1,72 @@
...
+        else
+        {
+            otherData = data
+                .ToDictionary(
+                    x => (string)x.Key,
+                    x =>
+                    {
+                        try
+                        {
+                            return JsonSerializer.Serialize(x.Value);
```
bruno-garcia, line 27: How successful is this in general? I have no experience with getting an NSObject serialized through S.T.J

In other SDKs we opt'ed to omit synchronizing this property due to risk/complexity.

bruno-garcia (approved):

src/Sentry/Platforms/iOS/Extensions/UserExtensions.cs
mattjohnsonpint: I'll just make leave it uninitialized in case of null, so the the default constructor can do its thing.

src/Sentry/Platforms/iOS/Extensions/UserExtensions.cs
mattjohnsonpint: Good callout.  I'll investigate.

mattjohnsonpint: Indeed, there is a better way to handle this.  I'll do that in a separate PR shortly.  Thanks.

PR #1836, opened 2022-08-02
Add and use serialization and dictionary extensions

mattjohnsonpint:
Following up feedback on #1834 

- Uses Apple's `NSJsonSerialization` for JSON serialization of `NSObject` and derived types.
- Consolidates `NSDictionary` conversions into extension methods.

#skip-changelog

### DIFF
--- a/src/Sentry/Platforms/iOS/Extensions/BreadcrumbExtensions.cs
+++ b/src/Sentry/Platforms/iOS/Extensions/BreadcrumbExtensions.cs
@@ -1,43 +1,26 @@
+using Sentry.Extensibility;
+
 namespace Sentry.iOS.Extensions;
 
 internal static class BreadcrumbExtensions
 {
-    public static Breadcrumb ToBreadcrumb(this SentryCocoa.SentryBreadcrumb breadcrumb)
-    {
-        var items = breadcrumb.Data as IEnumerable<KeyValuePair<NSString, NSObject?>>;
-        var data = items?.ToDictionary(
-            x => (string)x.Key,
-            x => x.Value?.ToString() ?? "");
-
-        return new Breadcrumb(
+    public static Breadcrumb ToBreadcrumb(this SentryCocoa.SentryBreadcrumb breadcrumb, IDiagnosticLogger? logger) =>
+        new(
             breadcrumb.Timestamp?.ToDateTimeOffset(),
             breadcrumb.Message,
             breadcrumb.Type,
-            data,
+            breadcrumb.Data.ToNullableStringDictionary(logger),
             breadcrumb.Category,
             breadcrumb.Level.ToBreadcrumbLevel());
-    }
 
-    public static SentryCocoa.SentryBreadcrumb ToCocoaBreadcrumb(this Breadcrumb breadcrumb)
-    {
-        NSDictionary<NSString, NSObject>? data = null;
-        if (breadcrumb.Data is { } breadcrumbData)
-        {
-            data = new NSDictionary<NSString, NSObject>();
-            foreach (var item in breadcrumbData)
-            {
-                data[item.Key] = NSObject.FromObject(item.Value);
-            }
-        }
-
-        return new SentryCocoa.SentryBreadcrumb
+    public static SentryCocoa.SentryBreadcrumb ToCocoaBreadcrumb(this Breadcrumb breadcrumb) =>
+        new()
         {
             Timestamp = breadcrumb.Timestamp.ToNSDate(),
             Message = breadcrumb.Message,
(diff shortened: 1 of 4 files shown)


src/Sentry/Platforms/iOS/Extensions/BreadcrumbExtensions.cs
```diff
@@ -1,43 +1,26 @@
...
-        var items = breadcrumb.Data as IEnumerable<KeyValuePair<NSString, NSObject?>>;
-        var data = items?.ToDictionary(
-            x => (string)x.Key,
-            x => x.Value?.ToString() ?? "");
-
-        return new Breadcrumb(
+    public static Breadcrumb ToBreadcrumb(this SentryCocoa.SentryBreadcrumb breadcrumb, IDiagnosticLogger? logger) =>
+        new(
             breadcrumb.Timestamp?.ToDateTimeOffset(),
             breadcrumb.Message,
```
SimonCropp, line 10: not that u changed this. but i get null warnings for message and type 
image

should we default them to something?

src/Sentry/Platforms/iOS/Extensions/CocoaExtensions.cs
```diff
@@ -1,8 +1,110 @@
...
+    {
+        if (dict == null)
+        {
+            return new Dictionary<string, string>();
+        }
+
+        var result = new Dictionary<string, string>(capacity: (int)dict.Count);
+        foreach (var key in dict.Keys)
+        {
+            var value = dict[key];
```
SimonCropp, line 60:
```suggestion
        foreach (var keyValue in dict)
        {
            var value = keyValue.Value;
            var key = keyValue.Key;
```

no need to do a dictionary lookup

src/Sentry/Platforms/iOS/Extensions/CocoaExtensions.cs
```diff
@@ -1,8 +1,110 @@
...
+                d.Values.ToArray(),
+                d.Keys.ToArray());
+    }
+
+    public static NSDictionary<NSString, NSObject>? ToNullableNSDictionary<TValue>(
+        this IEnumerable<KeyValuePair<string, TValue>> dict)
+    {
+        var d = dict.ToNSDictionary();
+        return d.Count == 0 ? null : d;
+    }
```
SimonCropp, line 109:
```suggestion
    public static NSDictionary<NSString, NSObject>? ToNullableNSDictionary<TValue>(
        this ICollection<KeyValuePair<string, TValue>> dict)
    {
        if (dict.Count == 0)
        {
            return null;
        }

        return dict.ToNSDictionary();
    }
```

src/Sentry/Platforms/iOS/Extensions/CocoaExtensions.cs
```diff
@@ -1,8 +1,110 @@
...
+        IDiagnosticLogger? logger = null)
+        where TValue : NSObject
+    {
+        if (dict == null)
+        {
+            return new Dictionary<string, string>();
+        }
+
+        var result = new Dictionary<string, string>(capacity: (int)dict.Count);
+        foreach (var key in dict.Keys)
```
SimonCropp, line 58: u cant enumerate key-value pairs in `NSDictionary`?

src/Sentry/Platforms/iOS/Extensions/CocoaExtensions.cs
```diff
@@ -1,8 +1,110 @@
...
+        return result;
+    }
+
+    public static Dictionary<string, string>? ToNullableStringDictionary<TValue>(
+        this NSDictionary<NSString, TValue>? dict,
+        IDiagnosticLogger? logger = null)
+        where TValue : NSObject
+    {
+        var d = dict?.ToStringDictionary(logger);
+        return d?.Count is null or 0 ? null : d;
```
SimonCropp, line 82:
```suggestion
        if (dict is null || dict.Length == 0)
        {
            return null;
        }

        return dict.ToStringDictionary(logger);
```

src/Sentry/Platforms/iOS/Extensions/BreadcrumbExtensions.cs
mattjohnsonpint: They are nullable.  Intellisense failure.

src/Sentry/Platforms/iOS/Extensions/CocoaExtensions.cs
mattjohnsonpint: Not easily.

SimonCropp (approved):

PR #1838, opened 2022-08-04
add  --nologo to dotnet

SimonCropp:
remove some of the noise when looking at test results in the logs

#skip-changelog

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -28,15 +28,15 @@ jobs:
           uses: ./.github/actions/environment
 
         - name: Build
-          run: dotnet build -c Release /p:CopyLocalLockFileAssemblies=true
+          run: dotnet build -c Release --nologo /p:CopyLocalLockFileAssemblies=true
 
         - name: Test
-          run: dotnet test -c Release --no-build -l GitHubActions -l "trx;LogFilePrefix=testresults_${{ runner.os }}"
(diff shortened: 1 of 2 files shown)


mattjohnsonpint:
We already set it via environment variable globally

https://github.com/getsentry/sentry-dotnet/blob/e62013ca63436d2a786fb229570e96f42c7c4044/.github/actions/environment/action.yml#L11

If it's not working, we should figure out why.

mattjohnsonpint:
Looks like it's coming through to me.  For example:

https://github.com/getsentry/sentry-dotnet/runs/7672016918?check_suite_focus=true#step:5:6

image

mattjohnsonpint:
Looks like these do two different things.  We can have them both.

https://docs.microsoft.com/en-us/dotnet/core/tools/dotnet-build

> `--nologo`
> Doesn't display the startup banner or the copyright message.


https://docs.microsoft.com/dotnet/core/tools/dotnet-environment-variables#dotnet_nologo

> `DOTNET_NOLOGO`
> Specifies whether .NET welcome and telemetry messages are displayed on the first run. Set to true to mute these messages (values true, 1, or yes accepted) or set to false to allow them (values false, 0, or no accepted). If not set, the default is false and the messages will be displayed on the first run. This flag does not affect telemetry (see DOTNET_CLI_TELEMETRY_OPTOUT for opting out of sending telemetry).

mattjohnsonpint (approved):

PR #1839, opened 2022-08-04
add EventProcessorTests

SimonCropp:
and ensure non default contexts appear in snapshots

#skip-changelog


SimonCropp:
@mattjohnsonpint any ideas why this would only fail on windows in CI?

mattjohnsonpint (approved):

PR #1842, opened 2022-08-05
Add API for deliberately crashing an app

mattjohnsonpint:
This adds a new public API, `SentrySdk.CauseCrash(CrashType)`, and several crash types - some of which are platform specific.  This makes demonstrating how Sentry handles such crashes much easier.

The main Sentry library for the `net6.0-android` target will now include some Java and C code that was previously used in the Android sample.  The code and build scripts have been cleaned up a bit.  We will still keep the binaries checked in so contributors don't necessarily need to re-build these dependencies.  However we will rebuild them in CI to make sure the ones we are always clean and secure.

For the `net6.0-ios` target, we already have code in the Sentry Cocoa SDK that will cause a native crash, so we just call that.


src/Sentry/Platforms/iOS/SentrySdk.cs
```diff
@@ -6,11 +6,6 @@ namespace Sentry;
 
 public static partial class SentrySdk
 {
-    /// <summary>
-    /// Causes a native crash.  Use for testing purposes only.
-    /// </summary>
-    public static void CauseNativeCrash() => SentryCocoa.SentrySDK.Crash();
```
mattjohnsonpint, line 12: Note: Removing this wasn't a breaking change, because we just recently added it and it was never deployed.

SimonCropp (approved):

PR #1849, opened 2022-08-12
Add and configure options for the iOS SDK

mattjohnsonpint:
- Sets the default options for the Cocoa SDK
- Added several internal extension methods to deal with various mappings between .NET and Objective-C data types
- Connect `BeforeBreadcrumb` on the Cocoa SDK
- Prevented duplicate events on unhandled managed crash by filtering out the corresponding native SIGABRT event

A few minor fixes along the way:
- Fixed the MAUI sample app crashing on startup on iOS
- Updated some `#if` statements
- Fixed some typos
- Set defaults explicitly for the Android options
- Fixed a bug with `DefaultTags` being duplicated in the Android SDK
- Aliased a few common types

Next to do (separate PRs coming for these)
- Finish connecting `BeforeSend` and `OnCrashedLastRun`, both which need `SentryEvent` to/from working.
  - Started on this - code is commented out in this PR.
- Gather device and context info in `IosScopeObserver`
- Need to move the `Distribution` option into the main SDK.

### DIFF
--- a/src/Sentry/Platforms/Android/SentryOptions.cs
+++ b/src/Sentry/Platforms/Android/SentryOptions.cs
@@ -250,7 +250,7 @@ public partial class SentryOptions
         /// events that originate from the embedded Android SDK. The default value is <c>false</c> (disabled).
         /// </summary>
         /// <remarks>
-        /// This is an experimental feature and is imperefct, as the .NET SDK and the embedded Android SDK don't
+        /// This is an experimental feature and is imperfect, as the .NET SDK and the embedded Android SDK don't
         /// implement all of the same features that may be present in the event graph. Some optional elements may
         /// be stripped away during the roundtripping between the two SDKs.  Use with caution.
         /// </remarks>
--- a/src/Sentry/Platforms/Android/SentryOptions.cs
+++ b/src/Sentry/Platforms/Android/SentryOptions.cs
@@ -204,8 +204,8 @@ public partial class SentryOptions
 
         // ---------- Other ----------
 
-        internal string[]? InAppExclude { get; set; }
-        internal string[]? InAppInclude { get; set; }
+        internal List<string>? InAppExcludes { get; private set; }
+        internal List<string>? InAppIncludes { get; private set; }
 
         /// <summary>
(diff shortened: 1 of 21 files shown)


src/Sentry/Platforms/iOS/Extensions/SentryEventExtensions.cs
```diff
@@ -0,0 +1,45 @@
...
+//         envelope.Serialize(stream, options.DiagnosticLogger);
+//         stream.Seek(0, SeekOrigin.Begin);
+//
+//         using var data = NSData.FromStream(stream)!;
+//         var cocoaEnvelope = SentryCocoa.PrivateSentrySDKOnly.EnvelopeWithData(data);
+//
+//         var cocoaEvent = (SentryCocoa.SentryEvent) cocoaEnvelope.Items[0];
+//         return cocoaEvent;
+//     }
+// }
```
mattjohnsonpint, line 45: Ignore this.  Leaving in as commented because I'll be returning to it soon.

mattjohnsonpint: Same goes for the few other blocks commented that will end up using this when ready.

src/Sentry/Platforms/Android/SentryOptions.cs
```diff
@@ -33,7 +33,7 @@ public class AndroidOptions
         /// <remarks>
         /// See https://docs.sentry.io/platforms/android/configuration/app-not-respond/
         /// </remarks>
-        public bool AnrReportInDebug { get; set; }
+        public bool AnrReportInDebug { get; set; } = false;
```
SimonCropp, line 36: why adding the redundant ` = false;`?

mattjohnsonpint: To be explicit about matching the defaults that are set in the native SDKs.  Easy to miss an `= true` if there's a whole bunch of falses in a row. Better in this case to have a place for the default on every property.

SimonCropp (approved):

PR #1851, opened 2022-08-13
Add `Distribution` properties to go along with `Release` info

mattjohnsonpint:
From https://develop.sentry.dev/sdk/event-payloads/#optional-attributes

image

We had `release`, but not `dist` because it's usually used only in mobile apps.  For one example usage, see our [React Native docs](https://docs.sentry.io/platforms/react-native/sourcemaps/#using-custom-release-and-distribution).

We need this now, since we have the embedded Android and iOS SDKs.  However, for various reasons, it was much easier to make it an optional value available to all targets.


mattjohnsonpint:
Env vars are leaking.  I'll fix soon.

src/Sentry/Internal/DistributionLocator.cs
```diff
@@ -0,0 +1,19 @@
+using System;
+
+namespace Sentry.Internal
+{
+    internal static class DistributionLocator
```
SimonCropp, line 5: wouldnt this be simpler with a `GetEnvironmentVariable` in the static constructor and allow the resulting prop to be settable for testing purposes?

src/Sentry/Internal/MainSentryEventProcessor.cs
```diff
@@ -93,6 +95,11 @@ public SentryEvent Process(SentryEvent @event)
                 @event.Release = Release;
             }
 
+            if (@event.Distribution == null)
+            {
+                @event.Distribution = Distribution;
+            }
```
SimonCropp, line 101:
```
@event.Distribution ??= Distribution;
```

and the same on the two above?

src/Sentry/SentryOptions.cs
```diff
@@ -223,6 +223,20 @@ public bool IsGlobalModeEnabled
         /// <seealso href="https://docs.sentry.io/platforms/dotnet/configuration/releases/"/>
         public string? Release { get; set; }
 
+        /// <summary>
+        /// The distribution of the application, associated with the release set in <see cref="Release"/>.
+        /// </summary>
+        /// <example>
+        /// TODO: add examples
```
SimonCropp, line 230: given this will appear in intellisense,  probs better to have an issue to track it and remove the todo?

src/Sentry/Internal/DistributionLocator.cs
mattjohnsonpint: It sure would.  I was copying the existing pattern from the other "locator" classes, but that would be much simpler.  I'll do that next.

mattjohnsonpint: We can do much better with testing on environment variables anyway, and those locators can be collapsed to a single class.

mattjohnsonpint: On second thought, I'm going to do that work in a separate PR first and then update this to use it later.  We'll hold this one for now.

mattjohnsonpint:
This is ready.  Default values for release and distribution for Android and iOS will come in a separate PR after this is merged.

SimonCropp (approved):

PR #1852, opened 2022-08-13
Fix nullability warnings with IsNull extension method

mattjohnsonpint:
This is just a cleanup PR.

We have a lot of places in our code where we do this sort of check:

```csharp
// Not to throw on code that ignores nullability warnings.
// ReSharper disable once ConditionIsAlwaysTrueOrFalse
if (other is null)
{
    return null;
}
```

That particular ReSharper disable comment doesn't quite catch everything.  To do that, we'd have to expand them to this mess:

```csharp
#if NET461 || NETSTANDARD2_0
    // ReSharper disable once ConditionIsAlwaysTrueOrFalse
    // ReSharper disable HeuristicUnreachableCode
    if (other is null)
    {
        return;
    }
    // ReSharper restore HeuristicUnreachableCode
#else
    // ReSharper disable once ConditionIsAlwaysTrueOrFalseAccordingToNullableAPIContract
    if (other is null)
    {
        return;
    }
#endif
```

Instead, we can use a one-line extension method that doesn't need any of those comments:

```csharp
internal static class MiscExtensions
{
    public static bool IsNull(this object? o) => o is null;
}
```

#skip-changelog


bruno-garcia:
is the ide going to warn users to replace is null with this extension method? Seems like it'll be an item folks need to remember in pr reviews and point out to any new contributor

mattjohnsonpint:
No. It's just a workaround for when we want a runtime check for null when the IDE says it's not needed. We don't (and shouldn't) use it everywhere, and nothing will prompt you to use it. 

If we could make all consumers of our library require nullabilty checks, we wouldn't need it at all.

mattjohnsonpint:
It's internal anyway.

mattjohnsonpint:
Note also that the current implementation is problematic, because it only takes effect on `net461` and `netstandard2.0` targets.  On other targets like `net6.0`, you still get the yellow warning but also the actual effect of `ConditionIsAlwaysTrueOrFalse` being disabled passes further down the code.  In theory, it could disable something we wanted to see.

bruno-garcia:
The warning is about checking for null when not needed. Is it possible disabling the warning completely is a better route?

mattjohnsonpint:
No. That would hurt us in many other ways.  This is just about when we've specifically decided to make a guard for null at runtime on a public api that says null is not allowed.  The IDE will till us that the guard isn't required, which only holds true if the caller is also using nullabilty checks on their side.  So we've decided in several places to keep the guard in place.  I'm not suggesting we change anything in that regard.

All this does is fix the fact that our current ReSharper ignore comment only works on net461 and netstandard20 targets.  Most of the time I have net6.0 set in the editor and all of these show yellow.

mattjohnsonpint:
Some example screenshots of this code:

https://github.com/getsentry/sentry-dotnet/blob/40ec0d87fd1b8b41c04a3c774bcb9bc4b3e5dfd4/src/Sentry/Scope.cs#L324-L329

In .NET 6.0:

image

In .NET Framework 4.6.1.:

image

Resolve in .NET 6.0 by updating the ignore comment

image

... but that doesn't work for .NET 4.6.1.

image


This covers both but is big and ugly:

.NET 6.0:

image

.NET 4.6.1:

image



The extension method get the same effect without any ugliness

image

mattjohnsonpint:
Going back to your original question:

> is the ide going to warn users to replace is null with this extension method? 

The IDE warns that the null check isn't required.  The only thing it recommends is to delete the check completely.  We've decided to do it anyway, because it's on a public facing API and we don't know for certain that the consumer has nullability checks enabled.  So we need a workaround, which this provides.

SimonCropp:
Yuck. but seems the least worst option given how many legacy frameworks we still support

SimonCropp (approved):

PR #1853, opened 2022-08-15
Improve how settings from environment variables are resolved internally

mattjohnsonpint:
We had three separate locator classes: `DsnLocator`, `EnvironmentLocator`, and `ReleaseLocator`.  They mostly did the same thing, with some different default values.  All of them use environment variables and are static classes, so we had to protect ourselves during testing by using a semaphore in `EnvironmentVariableGuard`.

The problem with that approach is that lots of code still hits environment variables, and we can't put the guard on *every* test, so we sometimes get flaky test results elsewhere.

This new approach attaches a single `SettingLocator` instance as an internal property on `SentryOptions`.  It has consolidated all of the logic from the previous three locators, and gives us a clean places to add new settings if we need them.  (By "setting" here, I mean an option that might come from an environment variable or assembly attribute when not set explicitly in code or config file.)

At testing time, there's a clean `FakeSettingLocator` and extension methods to access it.  That allows us to test fake environment variables for a specific test without ever setting any real ones that could leak out of the test.

#skip-changelog (because there is no external impact)

(contains a few minor unrelated nullability annotation fixes also)


bruno-garcia (approved):
Great improvement, thanks @mattjohnsonpint !

PR #1854, opened 2022-08-17
fix: Skip attachment if stream is empty

bitsandfoxes:
Background: In Unity, we attach a screenshot attachment to the scope so when `GetStream` gets called we return a MemoryStream. In some circumstances, i.e. on a background thread, we cannot actually capture the screen and we return an empty MemoryStream. 

To avoid adding an empty `EnvelopeItem` we check for it first and warn.

### DIFF
--- a/src/Sentry/Envelopes/Envelope.cs
+++ b/src/Sentry/Envelopes/Envelope.cs
@@ -183,7 +183,7 @@ namespace Sentry.Protocol.Envelopes
                         // We pull the stream out here so we can length check
                         // to avoid adding an invalid attachment
                         var stream = attachment.Content.GetStream();
-                        if (stream.Length != 0)
+                        if (stream.TryGetLength() != 0)
                         {
                             items.Add(EnvelopeItem.FromAttachment(attachment, stream));
                         }
--- a/src/Sentry/Envelopes/Envelope.cs
+++ b/src/Sentry/Envelopes/Envelope.cs
@@ -180,14 +180,17 @@ namespace Sentry.Protocol.Envelopes
                 {
                     try
                     {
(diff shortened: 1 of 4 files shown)


src/Sentry/Envelopes/Envelope.cs
```diff
@@ -180,7 +180,18 @@ internal void Serialize(Stream stream, IDiagnosticLogger? logger, ISystemClock c
...
+                        // We pull the stream out here so we can length check
+                        // to avoid adding an invalid attachment
+                        var stream = attachment.Content.GetStream();
+                        if (stream.Length != 0)
+                        {
+                            items.Add(EnvelopeItem.FromAttachment(attachment, stream));
+                        }
+                        else
+                        {
+                            logger?.LogWarning("Did not add '{0}' to envelope because the stream was empty.",
```
bruno-garcia, line 192: For streams that are not buffered, could this return `0` before you actually do the first `Read`?

mattjohnsonpint: That's a really good point.  To address that, we may have to buffer the stream here, which feels wrong...  Thinking...

mattjohnsonpint: Or we can just type check.  Feels like cheating, but will work

mattjohnsonpint: Oh, we already have an extension for this, we can just use that.

mattjohnsonpint: FYI - We double checked a non-buffered stream.  As long as its seekable, the length will return correctly.  Specifically, a `FileStream` will give the length of the file even if you haven't read it yet - which is probably what people will be using the most on this API.

PR #1857, opened 2022-08-17
Allow some mobile options to be modified from defaults

mattjohnsonpint:
On Android, iOS, and MacCatalyst (and Windows if using MAUI):

- `DetectStartupTime` should default to `Fast` (not `Best`), but should be allowed to set to `None`
- `AutoSessionTracking` should be `true` by default, but should be allowed to set to `false`
- `IsGlobalModeEnabled` should always be `true` and not allowed to be changed.

Additionally:
- Consolidated some conditional compilation flags
  - note: `__IOS__` covers both iOS and MacCatalyst, `__MOBILE__` covers those plus Android
- Deprecated `SentrySdk.Init` methods that took an Android `Context` parameter, since we can get the application context directly.

### DIFF
--- a/samples/Sentry.Samples.Android/MainActivity.cs
+++ b/samples/Sentry.Samples.Android/MainActivity.cs
@@ -6,7 +6,7 @@ public class MainActivity : Activity
 {
     protected override void OnCreate(Bundle? savedInstanceState)
     {
-        SentrySdk.Init(this, o =>
+        SentrySdk.Init(o =>
         {
             o.Dsn = "https://eb18e953812b41c3aeb042e666fd3b5c@o447951.ingest.sentry.io/5428537";
             o.SendDefaultPii = true; // adds the user's IP address automatically
--- a/samples/Sentry.Samples.Maui/MainPage.xaml.cs
+++ b/samples/Sentry.Samples.Maui/MainPage.xaml.cs
@@ -23,7 +23,7 @@ public partial class MainPage
         JavaCrashBtn.IsVisible = false;
 #endif
 
-#if !(ANDROID || IOS || MACCATALYST)
+#if !__MOBILE__
         NativeCrashBtn.IsVisible = false;
 #endif
         base.OnAppearing();
@@ -73,7 +73,7 @@ public partial class MainPage
 
     private void OnNativeCrashClicked(object sender, EventArgs e)
     {
-#if ANDROID || IOS || MACCATALYST
+#if __MOBILE__
         SentrySdk.CauseCrash(CrashType.Native);
 #endif
     }
--- a/src/Sentry/Platforms/Android/SentrySdk.cs
+++ b/src/Sentry/Platforms/Android/SentrySdk.cs
@@ -3,7 +3,6 @@ using Android.OS;
 using Sentry.Android;
 using Sentry.Android.Callbacks;
 using Sentry.Android.Extensions;
-using Sentry.Extensibility;
 using Sentry.Protocol;
 
 // ReSharper disable once CheckNamespace
@@ -11,7 +10,7 @@ namespace Sentry;
 
 public static partial class SentrySdk
 {
-    private static AndroidContext? AndroidContext;
+    private static AndroidContext AppContext { get; set; } = Application.Context;
 
     /// <summary>
     /// Initializes the SDK for Android, with an optional configuration options callback.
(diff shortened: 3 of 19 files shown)


samples/Sentry.Samples.Android/MainActivity.cs
```diff
@@ -6,7 +6,7 @@ public class MainActivity : Activity
 {
     protected override void OnCreate(Bundle? savedInstanceState)
     {
-        SentrySdk.Init(this, o =>
+        SentrySdk.Init(o =>
```
bruno-garcia, line 9: interesting, things work alright like this? I thought this worked on MAUI for example because there's some Init call going on there

samples/Sentry.Samples.Maui/MainPage.xaml.cs
```diff
@@ -73,7 +73,7 @@ private void OnJavaCrashClicked(object sender, EventArgs e)
 
     private void OnNativeCrashClicked(object sender, EventArgs e)
     {
-#if ANDROID || IOS || MACCATALYST
+#if __MOBILE__
```
bruno-garcia, line 76: worth adding a `#else` with a log Info "hey this doesn't do anything here"

src/Sentry/SentryOptions.cs
```diff
@@ -26,16 +26,28 @@ public partial class SentryOptions
...
+        /// Always <c>true</c> for mobile targets.
         /// </summary>
         public bool IsGlobalModeEnabled
         {
-            get => ScopeStackContainer is GlobalScopeStackContainer;
-            set => ScopeStackContainer = value ? new GlobalScopeStackContainer() : new AsyncLocalScopeStackContainer();
+            get => true;
+            set
+            {
+                if (value == false)
```
bruno-garcia, line 39: nit, how about:
```suggestion
                if (value is false)
```

src/Sentry/SentryOptions.cs
```diff
@@ -573,13 +585,19 @@ public StackTraceMode StackTraceMode
         public long MaxAttachmentSize { get; set; } = 20 * 1024 * 1024;
 
         /// <summary>
-        /// Whether the SDK should attempt to detect the app's and device's startup time.
+        /// The mode that the SDK should user when attempting to detect the app's and device's startup time.
```
bruno-garcia, line 588:
```suggestion
        /// The mode that the SDK should use when attempting to detect the app's and device's startup time.
```

bruno-garcia (approved):
awesome stuff

samples/Sentry.Samples.Android/MainActivity.cs
mattjohnsonpint: Originally, yes.  But when I added iOS that changed because there's nothing to pass.

Instead we do this, after we validate the DSN in the main `Init`

https://github.com/getsentry/sentry-dotnet/blob/7ff846f2c15c5d77d3f54e4a7c78a6ddae4abe74/src/Sentry/SentrySdk.cs#L54-L59

samples/Sentry.Samples.Maui/MainPage.xaml.cs
mattjohnsonpint: I already make the button invisible, so no need really.

mattjohnsonpint:
I see the failure on the current build.  Will fix shortly...

SimonCropp:
i like it. but did we not need to add/change any tests related to this?

src/Sentry/Platforms/Android/SentrySdk.cs
```diff
@@ -179,7 +165,7 @@ private static void InitSentryAndroidSdk(SentryOptions options)
                 o.AddIgnoredExceptionForType(JavaClass.ForName("android.runtime.JavaProxyThrowable"));
             }));
 
-        // Set options for the managed SDK that depend on the Android SDK
+        // Set options for the managed SDK that depend on the Android SDK. (The user will not be able to modify these.)
```
SimonCropp, line 168: > The user will not be able to modify these

does this mean "there is no API what the user can use to modify these" or "we ignore/overwrite the options the user sets"?

mattjohnsonpint:
> i like it. but did we not need to add/change any tests related to this?

Not until we have ability to write tests for ios/android

src/Sentry/Platforms/Android/SentrySdk.cs
mattjohnsonpint: It means, we set these options ourselves during init, ignoring whatever the user provided.  These should be limited to what is needed to make the native and managed SDKs work together.

PR #1861, opened 2022-08-18
Fix environment name casing issue

### LINKED ISSUE #1267: Regression in changing the case of environment
### Environment

How do you use Sentry? Sentry.AspNetCore

Which SDK and version?
3.9.4

### Steps to Reproduce
There seems to be a regression for this issue: https://github.com/getsentry/sentry-dotnet/issues/999
In particular, I noticed this morning that I get the environment lower-cased even if I set the option "AdjustStandardEnvironmentNameCasing" to false.

~~It might be due to the changes in this file: https://github.com/ajbeaven/sentry-dotnet/blob/53323f0ea6b788a06aeb2cecf693fdf23e1b71a1/src/Sentry.AspNetCore/SentryAspNetCoreOptionsSetup.cs which seems to be different from the version that actually fixed the issue: https://github.com/getsentry/sentry-dotnet/pull/1057/files~~

EDIT My bad, there doesn't seem to be any change in the files above. 

### Expected Result

Environment not lowercased

### Actual Result

Environment lowercased (from Staging to staging)

SimonCropp:
Move environment casing to later in configuration to fix environment name casing issue

fixes #1267

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@
 - Fix logging loop with Serilog sentry ([#1828](https://github.com/getsentry/sentry-dotnet/pull/1828))
 - Skip attachment if stream is empty ([#1854](https://github.com/getsentry/sentry-dotnet/pull/1854))
 - Allow some mobile options to be modified from defaults ([#1857](https://github.com/getsentry/sentry-dotnet/pull/1857))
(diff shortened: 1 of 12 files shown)


mattjohnsonpint (approved):

mattjohnsonpint:
Made a few minor changes, but generally happy with this.  Thanks!

I also snuck in a new solution filter to improve workflow with aspnet core stuff. 😉

PR #1862, opened 2022-08-19
Add ISentryTransactionProcessor

### LINKED ISSUE #1468: ISentryEventProcessor doesn't do anything for transactions
### Environment

How do you use Sentry?
Sentry SaaS (sentry.io)

Which SDK and version?
Sentry.AspNetCore 3.13 used in dotnetcoreapp2.2 API project

### Steps to Reproduce

I created an ISentryEventProcessor like the example one
```
    public class ExampleEventProcessor : ISentryEventProcessor
    {
        private readonly IHttpContextAccessor _httpContext;

        public ExampleEventProcessor(IHttpContextAccessor httpContext) => _httpContext = httpContext;

        public SentryEvent Process(SentryEvent @event)
        {
            // Here I can modify the event, while taking dependencies via DI

            @event.SetExtra("Response:HasStarted", _httpContext.HttpContext?.Response.HasStarted);
            return @event;
        }
    }
```

Registered in my Startup.cs like so
```
        // This method gets called by the runtime. Use this method to add services to the container.
        public void ConfigureServices(IServiceCollection services)
        {
            services.AddTransient<ISentryEventProcessor, ExampleEventProcessor>();
...            
```

When I call and endpoint I see the transaction in Sentry but breakpoints in the EventProcessor are never triggered and it doesn't run.

I would like to use this to override properties of the event.

What am I doing wrong?

### Expected Result

EventProcessor is triggered for a transaction when I call an endpoint in the API

### Actual Result

Nothing happened

SimonCropp:
fixes #1468


src/Sentry/Internal/DiscardReason.cs
```diff
@@ -6,6 +6,7 @@ internal record DiscardReason(string Value) : Enumeration(Value)
         public static DiscardReason BeforeSend = new("before_send");
         public static DiscardReason CacheOverflow = new("cache_overflow");
         public static DiscardReason EventProcessor = new("event_processor");
+        public static DiscardReason TransactionProcessor = new("transaction_processor");
```
mattjohnsonpint, line 9: We can't define new discard reasons at this time.  That would require coordination with the backend.  From their perspective, an event discarded from our "transaction processor" would just use `event_processor`, since they don't distinguish between an error event and a transaction event.
https://develop.sentry.dev/sdk/client-reports/#envelope-item-payload

src/Sentry/Internal/Hub.cs
```diff
@@ -342,12 +342,27 @@ public void CaptureTransaction(Transaction transaction)
...
                 _enricher.Apply(transaction);
 
+                if (transaction.IsSampled != false)
+                {
+                    foreach (var processor in scope.GetAllTransactionProcessors())
+                    {
+                        processor.Process(transaction);
+                        if (transaction.IsSampled == false)
+                        {
+                            _options.ClientReportRecorder.RecordDiscardedEvent(DiscardReason.TransactionProcessor, DataCategory.Error);
```
mattjohnsonpint, line 359:
```suggestion
                            _options.ClientReportRecorder.RecordDiscardedEvent(DiscardReason.EventProcessor, DataCategory.Transaction);
```

src/Sentry/Internal/Hub.cs
```diff
@@ -342,12 +342,27 @@ public void CaptureTransaction(Transaction transaction)
...
+                var scope = currentScope.Key;
+                scope.Evaluate();
+                scope.Apply(transaction);
 
                 // Apply enricher
                 _enricher.Apply(transaction);
 
+                if (transaction.IsSampled != false)
+                {
+                    foreach (var processor in scope.GetAllTransactionProcessors())
```
mattjohnsonpint, line 354: We apply event and exception processors in `SentryClient`.  Why did you choose to apply transaction processors in `Hub`?

src/Sentry/Extensibility/ISentryTransactionProcessor.cs
```diff
@@ -0,0 +1,14 @@
...
+    /// <summary>
+    /// Process a <see cref="Transaction"/> during the prepare phase.
+    /// </summary>
+    public interface ISentryTransactionProcessor
+    {
+        /// <summary>
+        /// Process the <see cref="Transaction"/>
+        /// </summary>
+        /// <param name="transaction">The Transaction to process</param>
+        void Process(Transaction transaction);
```
mattjohnsonpint, line 12: Are we sure we don't want to use a nullable return value, like we do for event processors?  With this approach, you can only mutate the existing transaction - you can't create a new one.

src/Sentry/Internal/Hub.cs
```diff
@@ -342,12 +342,27 @@ public void CaptureTransaction(Transaction transaction)
...
-                currentScope.Key.Evaluate();
-                currentScope.Key.Apply(transaction);
+                var scope = currentScope.Key;
+                scope.Evaluate();
+                scope.Apply(transaction);
 
                 // Apply enricher
                 _enricher.Apply(transaction);
 
+                if (transaction.IsSampled != false)
```
mattjohnsonpint, line 352: I'm concerned that if we rely on the user setting `IsSampled = false` in a transaction processor to indicate that they want to discard the transaction, we might mix up our discard reasons for client reports.

If we used the nullable return value pattern like we do for event processors, we wouldn't have that concern.

src/Sentry/SentryOptionsExtensions.cs
```diff
@@ -210,6 +210,36 @@ public static void AddEventProcessorProvider(this SentryOptions options, Func<IE
...
 
+        /// <summary>
+        /// Adds an transaction processor which is invoked when creating a <see cref="Transaction"/>.
+        /// </summary>
+        /// <param name="options">The SentryOptions to hold the processor.</param>
+        /// <param name="processor">The transaction processor.</param>
+        public static void AddTransactionProcessor(this SentryOptions options, ISentryTransactionProcessor processor)
+            => options.TransactionProcessors = options.TransactionProcessors != null
+                ? options.TransactionProcessors.Concat(new[] { processor }).ToArray()
+                : new[] { processor };
```
mattjohnsonpint, line 221: Ughh..  maybe we should just use lists for these (and the existing ones)?

test/Sentry.Tests/TransactionProcessorTests.cs
```diff
@@ -0,0 +1,40 @@
+﻿namespace Sentry.Tests;
+
+[UsesVerify]
+public class TransactionProcessorTests
```
mattjohnsonpint, line 4: Need some more tests:

- Discarding a transaction in a transaction processor works as expected
- Discarding a transaction in a transaction processor logs the correct client report reason and category
- A transaction that is sampled out is not run through a transaction processor

src/Sentry/Internal/Hub.cs
SimonCropp: because it needs access to the `scope`. which would be a breaking change in `SentryClient`

SimonCropp: i would like to keep the approach consistent if u can see a way to do it in a non breaking way?

SimonCropp:
@mattjohnsonpint does `options.TransactionProcessors` need to be added to the bindings?

mattjohnsonpint:
> @mattjohnsonpint does `options.TransactionProcessors` need to be added to the bindings?

I think you're asking if we need to connect custom transaction processors to native iOS/Android events?  At the moment, we don't support that for custom event processors, so I don't expect that for transaction processors either.  When/if we do that, we'll do it for both at the same time.

src/Sentry/Internal/DelegateTransactionProcessor.cs
```diff
@@ -0,0 +1,21 @@
+﻿using System;
+using Sentry.Extensibility;
+
+namespace Sentry.Internal
+{
+    internal class DelegateTransactionProcessor : ISentryTransactionProcessor
+    {
+        private readonly Action<Transaction> _action;
+
+        public DelegateTransactionProcessor(Action<Transaction> action)
```
mattjohnsonpint, line 10: Should probably be `Func<Transaction, Transaction?>` yes?

src/Sentry/Internal/Hub.cs
mattjohnsonpint: It seems that `CaptureEvent` takes an optional scope parameter, where the default `null` will use the current scope.  Since `CaptureTransaction` doesn't take a `scope` parameter, but `Hub.CaptureTransaction` gets the current scope from the scope manager, that you could also get the current scope in `SentryClient.CaptureTransaction` and evaluate the transaction processors there.

Alternatively, we could just only allow transaction processors to be added on `SentryOptions` globally, and not on a specific scope.  I don't quite understand the use case for adding an event/transaction processor only to a scope anyway.

I must admit I don't quite understand the need for all these layers of indirection we have... Something perhaps to re-evaluate later.

SimonCropp: @bruno-garcia can u weigh in here? is there any use case for scoped transaction processors? and, historically, what was the use case of scoped EventProcessors and ExceptionProcessors

mattjohnsonpint: Ping @bruno-garcia

test/Sentry.Tests/TransactionProcessorTests.cs
```diff
@@ -0,0 +1,95 @@
...
+
+    public class TrackingProcessor : ISentryTransactionProcessor
+    {
+        public Transaction Process(Transaction transaction)
+        {
+            Called = true;
+            return transaction;
+        }
+
+        public static bool Called;
```
mattjohnsonpint, line 57: Avoid static in unit tests where possible.  This could just be an instance property.

Looks good otherwise.  Thanks.

test/Sentry.Tests/TransactionProcessorTests.cs
SimonCropp: 👍🏻

mattjohnsonpint (approved):

PR #1863, opened 2022-08-19
Apply WinUI 3 exception handler in Sentry core

mattjohnsonpint:
Previously we were hooking up the WinUI 3 unhandled exception handler in Sentry.Maui only.

This moves that to an internal integration in Sentry core, so that it can also be used by WinUI 3 apps directly, or by other non-MAUI platforms that can leverage WinUI 3, such as [Uno Platform](https://platform.uno/docs/articles/uwp-vs-winui3.html).

Note that this doesn't apply for UWP (WinUI 2) apps.  Those will still need to hook up the handler directly, per [our configuration guide](https://docs.sentry.io/platforms/dotnet/guides/uwp/)

I also added a button to the MAUI sample that causes a crash on a background thread, which I used to verify that both types of crashes work.  One is caught by this integration, the other is caught by the existing AppDomain integration.

### DIFF
--- a/src/Sentry/Integrations/WinUIUnhandledExceptionIntegration.cs
+++ b/src/Sentry/Integrations/WinUIUnhandledExceptionIntegration.cs
@@ -32,9 +32,8 @@ namespace Sentry.Integrations
 
     internal class WinUIUnhandledExceptionIntegration : ISdkIntegration
     {
-        private static readonly Assembly? WinUIAssembly = GetWinUIAssembly();
-
         private static readonly byte[] WinUIPublicKeyToken = Convert.FromHexString("de31ebe4ad15742b");
+        private static readonly Assembly? WinUIAssembly = GetWinUIAssembly();
 
         private Exception? _lastFirstChanceException;
         private IHub _hub = null!;
--- a/src/Sentry/Integrations/WinUIUnhandledExceptionIntegration.cs
+++ b/src/Sentry/Integrations/WinUIUnhandledExceptionIntegration.cs
@@ -1,0 +1,133 @@
+#if NET5_0_OR_GREATER 
+using System;
+using System.Reflection;
+using System.Runtime.InteropServices;
+
+namespace Sentry.Integrations
+{
+    // This integration hooks unhandled exceptions in WinUI 3.
+    // The primary hook is Microsoft.UI.Xaml.Application.Current.UnhandledException
+    //
+    // There are some quirks to be aware of:
+    //
+    // - By default, important details (message, stack trace, etc.) are stripped away.
+    //   We can work around this by catching first-chance exceptions.
+    //   See: https://github.com/microsoft/microsoft-ui-xaml/issues/7160
+    //
+    // - Exceptions from background threads are not caught here.
+    //   However, they are caught by System.AppDomain.CurrentDomain.UnhandledException,
+    //   which we already hook in our AppDomainUnhandledExceptionIntegration
+    //   See: https://github.com/microsoft/microsoft-ui-xaml/issues/5221
+    //
+    // Note that we use reflection in this integration to get at WinUI code.
+    // If we ever add a Windows platform target (net6.0-windows, etc.), we could refactor to avoid reflection.
+    //
+    // This integration is for WinUI 3.  It does NOT work for UWP (WinUI 2).
+    // For UWP, the calling application will need to hook the event handler.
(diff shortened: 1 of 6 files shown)


src/Sentry/Integrations/WinUIUnhandledExceptionIntegration.cs
```diff
@@ -0,0 +1,133 @@
...
+                // first check by name
+                var assemblyName = x.GetName();
+                if (assemblyName.Name != WinUIAssemblyName)
+                {
+                    return false;
+                }
+
+                // check the public key token also
+                var token = assemblyName.GetPublicKeyToken();
+                return token != null && string.Equals(Convert.ToHexString(token), WinUiAssemblyPublicKeyToken, StringComparison.OrdinalIgnoreCase);
```
SimonCropp, line 91: would be better to convert `WinUiAssemblyPublicKeyToken` to bytes once at startup. then to a byte array compare

src/Sentry/Integrations/WinUIUnhandledExceptionIntegration.cs
```diff
@@ -0,0 +1,133 @@
...
+            });
+        }
+
+        private void AttachEventHandler()
```
SimonCropp, line 95: perhaps a bit more defensive here?

```
            catch (Exception e)
            {
                throw new($"In assembly {WinUIAssembly!.GetName().Name} expected to find the type {ApplicationTypeFullName} with a property {CurrentPropertyName} and an event {UnhandledExceptionEventName}");
            }
```

src/Sentry/Integrations/WinUIUnhandledExceptionIntegration.cs
```diff
@@ -0,0 +1,133 @@
...
+            var application = applicationType.GetProperty(CurrentPropertyName)!.GetValue(null);
+            var eventInfo = applicationType.GetEvent(UnhandledExceptionEventName)!;
+            var typedHandler = Delegate.CreateDelegate(eventInfo.EventHandlerType!, handler.Target, handler.Method);
+            eventInfo.AddEventHandler(application, typedHandler);
+        }
+
+        private void WinUIUnhandledExceptionHandler(object sender, object e)
+        {
+            var eventArgsType = e.GetType();
+            var handled = (bool)eventArgsType.GetProperty(HandledPropertyName)!.GetValue(e)!;
```
SimonCropp, line 111: similar defensive  approach here?

src/Sentry/Integrations/WinUIUnhandledExceptionIntegration.cs
```diff
@@ -0,0 +1,133 @@
...
+        private SentryOptions _options = null!;
+
+        // Constants used for reflection
+        private const string WinUIAssemblyName = "Microsoft.WinUI";
+        private const string WinUiAssemblyPublicKeyToken = "de31ebe4ad15742b";
+        private const string WinUINamespace = "Microsoft.UI.Xaml";
+        private const string ApplicationTypeName = "Application";
+        private const string UnhandledExceptionEventName = "UnhandledException";
+        private const string CurrentPropertyName = "Current";
+        private const string HandledPropertyName = "Handled";
```
SimonCropp, line 46: have to say i have never been a fan of single use constants disconnected from the context of the use. 

when reading

```
 var handled = (bool)eventArgsType.GetProperty(HandledPropertyName)!.GetValue(e)!;
```

the constant is just more obscured (and more verbose) compared to 

```
 var handled = (bool)eventArgsType.GetProperty("Handled")!.GetValue(e)!;
```

SimonCropp:
forgive my ignorance, i can bever remember all the windows UI approaches.  but isnt `uap` the tfm for winui? and wouldnt that make this easier?

mattjohnsonpint:
> forgive my ignorance, i can bever remember all the windows UI approaches.  but isnt `uap` the tfm for winui? and wouldnt that make this easier?

That's for UWP (WinUI 2).  This is for WinUI 3.

Also, having a Windows target would mean a lot of build and CI changes, for repackaging,  etc.  Better to avoid it if possible.

SimonCropp:
> Also, having a Windows target would mean a lot of build and CI changes, for repackaging, etc. Better to avoid it if possible.

ok. i was assuming u had already considered it. thanks

SimonCropp:
few minor suggestions above. but nothing that "needs" to be done.

SimonCropp (approved):

PR #1866, opened 2022-08-22
Move Array option over to use List<T>

SimonCropp:
@mattjohnsonpint as requested

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ## Features
 
+- Move Array option over to use List<T> ([#1866](https://github.com/getsentry/sentry-dotnet/pull/1866))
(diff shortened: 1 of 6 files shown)


SimonCropp:
@mattjohnsonpint how do i regen the bindings?

mattjohnsonpint:
Why?

mattjohnsonpint (approved):

mattjohnsonpint:
To clarify - In general, we don't regenerate the iOS bindings - they were generated once and then manually modified.  If they need updating, we'd do that manually.  The Android bindings are regenerated automatically at compile time, and are controlled by the [Metadata.xml](https://github.com/getsentry/sentry-dotnet/blob/main/src/Sentry/Platforms/Android/Transforms/Metadata.xml) file.  

Regardless, there shouldn't be any need to touch the Android or iOS bindings with this change.  They are updated when the public APIs we are calling the APIs in the iOS/Android SDKs change - not when our internals change (generally).

PR #1873, opened 2022-08-23
Try GNU tar on Windows to speed up dependency caching

mattjohnsonpint:
Per https://github.com/actions/cache/issues/752#issuecomment-1222415717

#skip-changelog


mattjohnsonpint:
FYI, it did switch over to GNU tar, but it did not make any difference in performance.

Issue #1835, opened 2022-08-02
Pre-flight Transactions being registered as Unknown


lucas-zimerman:
### Package

Sentry.AspNetCore

### .NET Flavor

.NET Core

### .NET Version

3.1.27

### OS

Linux

### SDK Version

3.11.1

### Steps to Reproduce

Pre-flight requests are being registered as transactions by the SDK, not sure what's the ideal case here, but, they are all grouped as Unknown Route with OPTIONS as http method.

### Expected Result

A formatted transaction name. (or discarded transaction)

### Actual Result

![image](https://user-images.githubusercontent.com/8229322/182371817-3f111bd1-144a-4658-abff-cb40769fe54a.png)

mattjohnsonpint:
Any chance we could get a minimal repro?  Thanks.

SimonCropp:
@lucas-zimerman  the repro could perhaps use a similar approach to https://github.com/getsentry/sentry-dotnet/blob/main/test/Sentry.AspNetCore.Tests/VersioningTests.cs#L22

lucas-zimerman:
@mattjohnsonpint  @SimonCropp 
I was only able to repro it within .NET Core 3.1 (Only tested on .NET 6 and .NET Core 3.1).

![image](https://user-images.githubusercontent.com/8229322/184058417-585b1af5-bf7b-4d2d-902e-d1143f9f659f.png)

It seems flaky as I shown on the picture, at some runs it stores the transaction name, but there is other times where it gives a Unknown Route as a path, it also seems to only be happening on the Option Route.


Sample: https://github.com/getsentry/sentry-dotnet/tree/sample/option-tracing-issue

I amrunning the following test 
sentry-dotnet\test\Sentry.AspNetCore.Tests\VersioningTests.cs -> SimpleOptionRoute

I didn't validate if Verify is working or not, all I did was add a breakpoint to check the payload content on line 132.

lucas-zimerman:
Ok so the issue seems related to how the Request is being received:
```
curl \
 -X OPTIONS \
 --compressed \
 -H "Accept: */*" \
 -H "Accept-Encoding: gzip" \
 -H "Accept-Language: pt-BR,pt;q=0.9,en-US;q=0.8,en;q=0.7, en-AL;q=0.1" \
 -H "Access-Control-Request-Headers: authorization,content-type" \
 -H "Access-Control-Request-Method: GET" \
 -H "Cdn-Loop: cloudflare" \
[...]

PR #1880, opened 2022-08-28
fallback to request path for route

SimonCropp:
potential fix for #1835


SimonCropp:
@mattjohnsonpint @bruno-garcia @lucas-zimerman thoughts on this as a solution for pre-flight requests?

bruno-garcia (changes requested):
Pleas don't use the raw URL unless the new "source: raw" is provided. Transactions are grouped by name and raw URL can cause issues

lucas-zimerman:
I think this is more of a business rule on what todo with pre-flight requests? Sentry UI should mention it's a pre-flight request so users don't get confused with requests with the `wrong` method  succeeding.

What if we just don't create a new transaction for pre-flight requests? I see them just as a transaction quota burner than giving any important info to the end user.

mattjohnsonpint:
> Pleas don't use the raw URL unless the new "source: raw" is provided. Transactions are grouped by name and raw URL can cause issues

I think @bruno-garcia is referring to https://develop.sentry.dev/sdk/event-payloads/transaction/#transaction-annotations

```json
{
  "transaction_info": {
    "source": "url"
  }
}
```

mattjohnsonpint:
> I think this is more of a business rule on what todo with pre-flight requests? Sentry UI should mention it's a pre-flight request so users will get confused with requests with the `wrong` method succeeding.
> 
> What if we just don't create a new transaction for pre-flight requests? I see them just as a transaction quota burner than giving any important info to the end user.

@lucas-zimerman - Preflight requests are indeed HTTP requests that get made from the client to the server.  It's possible they could be a source of slowness, depending on how the server is configured to handle them.  I think we should not filter them out.

mattjohnsonpint:
Per internal discussion, we've decided that OPTIONS requests do not need to be sent to Sentry after all.

PR #1885, opened 2022-08-31
Fix detection of .NET Framework 4.8.1

mattjohnsonpint:
Builds are failing because Github Actions runners [now have](https://github.com/actions/runner-images/pull/6157) .NET Framework 4.8.1 installed, and we don't have that in our version map.

Example failure:

https://github.com/getsentry/sentry-dotnet/runs/8105415270?check_suite_focus=true#step:6:556

```
Error: Expected string not to be <null>.

  Failed Sentry.Tests.PlatformAbstractions.RuntimeInfoTests.SetReleaseAndVersionNetFx_OnNetFx_NonNullReleaseAndVersion [13 ms]
  Error Message:
   Expected string not to be <null>.
  Stack Trace:
     at FluentAssertions.Execution.XUnit2TestFramework.Throw(String message)
   at FluentAssertions.Execution.AssertionScope.FailWith(Func`1 failReasonFunc)
   at FluentAssertions.Primitives.ReferenceTypeAssertions`2.NotBeNull(String because, Object[] becauseArgs)
   at Sentry.Tests.PlatformAbstractions.RuntimeInfoTests.SetReleaseAndVersionNetFx_OnNetFx_NonNullReleaseAndVersion() in D:\a\sentry-dotnet\sentry-dotnet\test\Sentry.Tests\PlatformAbstractions\RuntimeInfoTests.cs:line 51
```


mattjohnsonpint:
FYI, according to [the official docs](https://docs.microsoft.com/en-us/dotnet/framework/migration-guide/how-to-determine-which-versions-are-installed#detect-net-framework-45-and-later-versions), the only build number we should expect is `533325`.  However, in [this comment](https://devblogs.microsoft.com/dotnet/announcing-dotnet-framework-481/#comment-16270) a user mentions `533320` is possible also.   Not sure how to validate that, but there doesn't seem to be any harm in adding both.

SimonCropp (approved):

PR #1913, opened 2022-09-10
Run most unit tests on Android

mattjohnsonpint:
This PR enables `Sentry.Tests` and `Sentry.Extensions.Logging.Tests` to run as part of the Android device test (along with existing `Sentry.Maui.Tests`)

A few notes:

- `Verify` tests are still disabled due to absence of verification files on the device at runtime.
- Because `AutoSessionTracking` is true by default for mobile, it needed to be disabled in a lot of places for existing tests to pass.  Otherwise, just initializing the SDK creates session messages that can interfere with the intent of the test.
- `IsGlobalMode` was forced `true` previously.  I changed it just so `true` is the *default* for mobile, but it's still possible to set it `false` now.  This allows tests that require global mode disabled to pass.
- Many tests had live transports connected unnecessarily.  Since the DSN is fake, that can lead to timeouts while the domain tries to resolve.  I've substituted them where appropriate.

There was a lot of trial and error with this, but I believe it's reliable now.

#skip-changelog


test/Sentry.Tests/Internals/JsonTests.cs
```diff
@@ -175,37 +175,6 @@ public void WriteDynamicValue_ClassWithAssembly_SerializedClassWithNullAssembly(
         Assert.Equal(expectedSerializedData, serializedString);
     }
 
-    [Fact]
```
mattjohnsonpint, line 178: Note, I deleted this test because it behaves differently on android, but also because we don't ever serialize a `TimeZoneInfo` in this way.  We serialize only the `Id` and `DisplayName` - separately.

test/Sentry.Tests/Protocol/DsnTests.cs
```diff
@@ -30,14 +30,14 @@ public void Ctor_SampleValidDsnWithSecret_CorrectlyConstructs()
     public void Ctor_NotUri_ThrowsUriFormatException()
     {
         var ex = Assert.Throws<UriFormatException>(() => Dsn.Parse("Not a URI"));
-        Assert.Equal("Invalid URI: The format of the URI could not be determined.", ex.Message);
+        ex.Message.Should().BeOneOf("net_uri_BadFormat", "Invalid URI: The format of the URI could not be determined.");
```
mattjohnsonpint, line 33: These strings are the pre-resolved string resource names from .NET.  They come through on Android in the emulators used in CI, but not locally.  I couldn't figure out why - but it didn't seem important to fix so I just allow them.

test/Sentry.Tests/SentryClientTests.cs
```diff
@@ -322,106 +332,48 @@ public void CaptureEvent_SamplingNull_DropsEvent()
         Assert.Same(@event, received);
     }
 
-    [Fact]
-    public void CaptureEvent_Sampling50Percent_EqualDistribution()
+    [Theory]
+    [InlineData(0.25f)]
+    [InlineData(0.50f)]
+    [InlineData(0.75f)]
+    public void CaptureEvent_WithSampleRate_AppropriateDistribution(float sampleRate)
```
mattjohnsonpint, line 339: I consolidated these three tests into one, and then simplified it.  I also had to add some retry due to randomness of results.

src/Sentry/SentryOptions.cs
```diff
@@ -37,24 +37,29 @@ public partial class SentryOptions
...
-            get => true;
+            get => _isGlobalModeEnabled;
             set
             {
-                if (value is false)
+                _isGlobalModeEnabled = value;
+                if (!value)
                 {
-                    _diagnosticLogger?.LogWarning("Cannot disable Global Mode on {0}", DeviceInfo.PlatformName);
+                    _diagnosticLogger?.LogWarning("Global Mode should usually be enabled on {0}", DeviceInfo.PlatformName);
```
bruno-garcia, line 54: nit:

```suggestion
                   // The use case to expose this are integration tests
                    _diagnosticLogger?.LogWarning("Global Mode should usually be enabled on {0}", DeviceInfo.PlatformName);
```

bruno-garcia (approved):
Awesome sauce! Thanks Matt

PR #1919, opened 2022-09-15
Use URL path in transaction names instead of "Unknown Route"

mattjohnsonpint:
Now that we have `TransactionNameSource` (see #1910), we no longer need to send "Unknown Route" if we can't identify the route for a web request.  Instead, we'll use the URL path, and set `TransactionNameSource.Url`.

Also, if the route is set via `SentryAspNetCoreOptions.TransactionNameProvider`, then we'll continue to use that - but will set `TransactionNameSource.Custom`.

Only when we've actually got a *route* should we set `TransactionNameSource.Route`.

### DIFF
--- a/src/Sentry.AspNetCore/SentryTracingMiddleware.cs
+++ b/src/Sentry.AspNetCore/SentryTracingMiddleware.cs
@@ -12,7 +12,6 @@ namespace Sentry.AspNetCore
     /// </summary>
     internal class SentryTracingMiddleware
     {
-        internal const string UnknownRouteTransactionName = "Unknown Route";
         private const string OperationName = "http.server";
 
         private readonly RequestDelegate _next;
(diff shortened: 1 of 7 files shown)


test/Sentry.AspNetCore.Tests/WebIntegrationTests.PreFlightIgnoresTransaction.verified.txt
```diff
@@ -23,8 +23,8 @@
   Payloads: [
     {
       Source: {
-        Name: Unknown Route,
-        NameSource: Route,
+        Name: OPTIONS /Target,
+        NameSource: Url,
```
mattjohnsonpint, line 27: Note, that this essentially would handle #1835, except that we've decided to no longer sent pre-flight requests.  The work to remove such transactions will come separately by @SimonCropp.

src/Sentry.AspNetCore/SentryTracingMiddleware.cs
```diff
@@ -161,10 +160,26 @@ public async Task InvokeAsync(HttpContext context)
...
-                    // If no Name was found for Transaction, fallback to UnknownRoute name.
+                    // If no Name was found for Transaction, then we don't have the route.
                     if (transaction.Name == string.Empty)
                     {
-                        transaction.Name = UnknownRouteTransactionName;
+                        var method = context.Request.Method.ToUpperInvariant();
+
+                        // If we've set a TransactionNameProvider, use that here
+                        var customTransactionName = context.TryGetCustomTransactionName();
+                        if (!string.IsNullOrEmpty(customTransactionName))
```
SimonCropp, line 170: can we invert this to avoid the negation

mattjohnsonpint: I'd prefer not to on this one, as it keeps it in the sequence that they would apply.

SimonCropp (approved):

PR #1922, opened 2022-09-15
Fix missing details when aggregate exception is filtered out

### LINKED ISSUE #1917: Stack traces missing from inner exceptions of aggregate exceptions
### SDK Version

3.21.0

### Steps to Reproduce

```cs
using Sentry;

using (SentrySdk.Init(o =>
       {
           o.Dsn = "https://eb18e953812b41c3aeb042e666fd3b5c@o447951.ingest.sentry.io/5428537";
           o.Debug = true;
       }))
{
    throw new AggregateException(new MyException());
}

public class MyException : Exception
{
}
```

### Expected Result

The `AggregateException` should be filtered out, and Sentry should capture the unhandled `MyException`, retaining the stack trace.

### Actual Result

The `AggregateException` is filtered out, but the stack trace is gone also.

We need to copy over the stack trace when filtering out aggregate exceptions.

mattjohnsonpint:
Fixes #1917

A few minor code cleanup items also.

### DIFF
--- a/src/Sentry/Internal/MainExceptionProcessor.cs
+++ b/src/Sentry/Internal/MainExceptionProcessor.cs
@@ -25,9 +25,7 @@ namespace Sentry.Internal
         {
             _options.LogDebug("Running processor on exception: {0}", exception.Message);
 
-            var sentryExceptions = CreateSentryException(exception)
-                // Otherwise realization happens on the worker thread before sending event.
-                .ToList();
+            var sentryExceptions = CreateSentryException(exception);
 
             MoveExceptionExtrasToEvent(sentryEvent, sentryExceptions);
 
@@ -81,7 +79,7 @@ namespace Sentry.Internal
             }
         }
 
-        internal IEnumerable<SentryException> CreateSentryException(Exception exception)
+        internal List<SentryException> CreateSentryException(Exception exception)
         {
             var exceptions = exception
                 .EnumerateChainedExceptions(_options)
--- a/src/Sentry/Internal/MainExceptionProcessor.cs
+++ b/src/Sentry/Internal/MainExceptionProcessor.cs
@@ -25,7 +25,7 @@ namespace Sentry.Internal
         {
             _options.LogDebug("Running processor on exception: {0}", exception.Message);
 
-            var sentryExceptions = CreateSentryException(exception);
+            var sentryExceptions = CreateSentryExceptions(exception);
 
             MoveExceptionExtrasToEvent(sentryEvent, sentryExceptions);
 
@@ -79,7 +79,7 @@ namespace Sentry.Internal
             }
         }
 
-        internal List<SentryException> CreateSentryException(Exception exception)
+        internal List<SentryException> CreateSentryExceptions(Exception exception)
         {
             var exceptions = exception
                 .EnumerateChainedExceptions(_options)
--- a/src/Sentry/Internal/MainExceptionProcessor.cs
+++ b/src/Sentry/Internal/MainExceptionProcessor.cs
@@ -107,8 +107,8 @@ namespace Sentry.Internal
         {
             var sentryEx = new SentryException
             {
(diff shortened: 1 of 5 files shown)


src/Sentry/Internal/MainExceptionProcessor.cs
```diff
@@ -80,18 +79,34 @@ public void Process(Exception exception, SentryEvent sentryEvent)
...
         {
-            return exception.EnumerateChainedExceptions(_options)
-                .Select(BuildSentryException);
+            var exceptions = exception
+                .EnumerateChainedExceptions(_options)
+                .Select(BuildSentryException)
+                .ToList();
+
+            // If we've filtered out the aggregate exception, we'll need to copy over details from it.
+            if (exception is AggregateException && !_options.KeepAggregateException)
```
SimonCropp, line 90: it feels weird that now we have "is AggregateException" and KeepAggregateException logic in two places. both EnumerateChainedExceptions and here. can we refactor to have them co-located?

test/Sentry.Tests/Internals/MainExceptionProcessorTests.cs
```diff
@@ -295,18 +338,18 @@ public void Process_HasUnsupportedExceptionValue_ValueSetAsExtra()
     {
         //Assert
         var sut = _fixture.GetSut();
-        var InvalidData = new KeyValuePair<string, string>("sentry:attachment:filename", "./path");
-        var InvalidData2 = new KeyValuePair<string, int?>("sentry:unsupported:value", null);
+        var invalidData = new KeyValuePair<string, string>("sentry:attachment:filename", "./path");
```
SimonCropp, line 341: can we have this as `invalidData1`

test/Sentry.Tests/Internals/MainExceptionProcessorTests.cs
```diff
@@ -208,9 +251,9 @@ public void Process_HasInvalidExceptionTagsValue_TagsSetWithInvalidValue()
     {
         //Assert
         var sut = _fixture.GetSut();
-        var InvalidTag = new KeyValuePair<string, int>("Tag1", 1234);
-        var InvalidTag2 = new KeyValuePair<string, int?>("Tag2", null);
-        var InvalidTag3 = new KeyValuePair<string, string>("", "abcd");
+        var invalidTag = new KeyValuePair<string, int>("Tag1", 1234);
```
SimonCropp, line 254: `invalidTag1`

test/Sentry.Tests/Internals/MainExceptionProcessorTests.cs
```diff
@@ -169,17 +169,60 @@ public Task CreateSentryException_Aggregate_Keep()
...
-            new Exception("Inner message1"),
-            new Exception("Inner message2"));
+        try
+        {
+            // Throwing will put a stack trace on the exception
+            throw new AggregateException(
+                new Exception("Inner message1"),
+                new Exception("Inner message2"));
+        }
+        catch (AggregateException ex)
```
SimonCropp, line 218: i personally disklike the `ex` convention. can we go with `exception`?

SimonCropp:
some minor optional feedback. but will approve anyway and let u pick which of those to address

SimonCropp (approved):

src/Sentry/Internal/MainExceptionProcessor.cs
mattjohnsonpint: The logic is different between those two.
- In `EnumerateChainedExceptions`, we still unwind aggregate exceptions regardless of the option.  The option only controls whether we include any aggregate exceptions in the result.
- In `CreateSentryExceptions`, we're specifically looking at the root exception because that's where the details we need to copy are.  The option just tells us that we can expect it to be filtered.

So they're not really able to be refactored together.

PR #1923, opened 2022-09-15
Cleanup SentryClientTests

mattjohnsonpint:
#skip-changelog

### DIFF
--- a/test/Sentry.Tests/SentryClientTests.cs
+++ b/test/Sentry.Tests/SentryClientTests.cs
@@ -13,18 +13,24 @@ public class SentryClientTests
     {
         public SentryOptions SentryOptions { get; set; } = new()
         {
-            AttachStacktrace = false
+            AttachStacktrace = false,
+            AutoSessionTracking = false
         };
 
         public IBackgroundWorker BackgroundWorker { get; set; } = Substitute.For<IBackgroundWorker, IDisposable>();
-        public IClientReportRecorder ClientReportRecorder { get; set; } = Substitute.For<IClientReportRecorder>();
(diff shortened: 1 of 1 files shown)


test/Sentry.Tests/SentryClientTests.cs
```diff
@@ -191,9 +194,6 @@ public void CaptureEvent_BeforeEvent_RejectEvent_RecordsDiscard()
     {
         _fixture.SentryOptions.BeforeSend = _ => null;
 
-        var transport = Substitute.For<ITransport>();
-        _fixture.SentryOptions.Transport = transport;
-
```
mattjohnsonpint, line 196: Note that with the background worker substituted already, there's no purpose in substituting the transport.  It will never get invoked.

test/Sentry.Tests/SentryClientTests.cs
```diff
@@ -746,29 +724,4 @@ public void Ctor_WrapsCustomTransportWhenCachePathOnOptions()
         var cachingTransport = Assert.IsType<CachingTransport>(_fixture.SentryOptions.Transport);
         _ = Assert.IsType<FakeTransport>(cachingTransport.InnerTransport);
     }
-
-    [Fact]
-    public async Task SentryClient_WithCachingTransport_RecordsDiscardedEvents()
```
mattjohnsonpint, line 751: This test was irrelevant, as the event gets discarded when captured, not when passed to the transport.  The transport is never invoked anyway, as we substituted the background worker.

It's also redundant, as we already test that we record discarded events correctly.

SimonCropp (approved):

PR #1924, opened 2022-09-15
Exception filters should consider child exceptions of an `AggregateException`

### LINKED ISSUE #1916: Nested inner exceptions are not filtered correctly
### SDK Version

3.21.0

### Steps to Reproduce

```cs
using Sentry;

using (SentrySdk.Init(o =>
       {
           o.Dsn = "https://eb18e953812b41c3aeb042e666fd3b5c@o447951.ingest.sentry.io/5428537";
           o.Debug = true;
           o.AddExceptionFilterForType<MyException>();
       }))
{
    throw new AggregateException(new MyException());
}

public class MyException : Exception
{
}
```

### Expected Result

No exception should be sent, because it's been configured to be filtered out.

### Actual Result

The exception is sent anyway, because it's an inner exception of another exception.

We already unwind inner exceptions and discard aggregates, but we appear to be filtering beforehand.  We probably *should* still filter before, because we'd want to discard any child exceptions of a filtered exception also.  We just need to walk the entire tree during the filtering process.

I haven't checked, but this probably applies to exception processors as well.

mattjohnsonpint:
Fixes #1916

### DIFF
--- a/src/Sentry/SentryClient.cs
+++ b/src/Sentry/SentryClient.cs
@@ -235,19 +235,27 @@ namespace Sentry
             var filters = _options.ExceptionFilters;
             if (exception == null || filters == null || filters.Count == 0)
             {
+                // There was nothing to filter.
                 return null;
             }
 
-            // Note that _options.KeepAggregateException is not relevant here.  Even if we want to keep aggregate
-            // exceptions, we would still never send one if all of its children are supposed to be filtered.
-            if (exception is AggregateException aggregate)
+            if (filters.Any(f => f.Filter(exception)))
             {
-                return aggregate.InnerExceptions.All(e => ApplyExceptionFilters(e) != null)
-                    ? aggregate.InnerExceptions
-                    : null;
+                // The event should be filtered based on the given exception
+                return new[] {exception};
             }
 
-            return filters.Any(f => f.Filter(exception)) ? new[] {exception} : null;
+            if (exception is AggregateException aggregate &&
+                aggregate.InnerExceptions.All(e => ApplyExceptionFilters(e) != null))
+            {
+                // All inner exceptions of the aggregate matched a filter, so the event should be filtered.
+                // Note that _options.KeepAggregateException is not relevant here.  Even if we want to keep aggregate
+                // exceptions, we would still never send one if all of its children are supposed to be filtered.
+                return aggregate.InnerExceptions;
+            }
+
+            // The event should not be filtered.
+            return null;
         }
 
         /// <summary>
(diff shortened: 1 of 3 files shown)


src/Sentry/SentryClient.cs
```diff
@@ -168,15 +168,13 @@ private SentryId DoSendEvent(SentryEvent @event, Scope? scope)
                 }
             }
 
-            if (@event.Exception != null && _options.ExceptionFilters?.Count > 0)
```
SimonCropp, line 171: given we have

```
internal List<IExceptionFilter>? ExceptionFilters { get; set; } = new();
```

and it is internal. can we mark is as not nullable?

mattjohnsonpint: Good catch.  There's some inconsistency about this in other properties on `SentryOptions` as well, so I'll make a separate PR.

src/Sentry/SentryClient.cs
```diff
@@ -232,6 +230,26 @@ private SentryId DoSendEvent(SentryEvent @event, Scope? scope)
...
+                return null;
+            }
+
+            // Note that _options.KeepAggregateException is not relevant here.  Even if we want to keep aggregate
+            // exceptions, we would still never send one if all of its children are supposed to be filtered.
+            if (exception is AggregateException aggregate)
+            {
+                return aggregate.InnerExceptions.All(e => ApplyExceptionFilters(e) != null)
+                    ? aggregate.InnerExceptions
+                    : null;
```
SimonCropp, line 247: doesnt this ignore the scenario where someone has a filter explicitly for the AggregateException ?

SimonCropp: seems this should be 
```
    if (exception is AggregateException aggregate)
    {
        if (aggregate.InnerExceptions.All(e => ApplyExceptionFilters(e) != null))
        {
            return aggregate.InnerExceptions;
        }
    }
```

src/Sentry/SentryClient.cs
SimonCropp: 👍🏻

src/Sentry/SentryClient.cs
mattjohnsonpint: That would cause the unfiltered case to fall through, which would invoke the filters again just to come up with the same `null` result.

In case not clear here - it's either an all or nothing.  Either all inner exceptions match the filters and the event is filtered out, or the event is not filtered out (null)

mattjohnsonpint: > doesnt this ignore the scenario where someone has a filter explicitly for the AggregateException ?

Oh, I missed that actually.  That would be strange, but yeah we should handle it.  And yeah, your proposal would do it.

SimonCropp (approved):

Issue #1332, opened 2021-11-19
invalid event envelope. Error causes: unexpected end of file. If Context is null


lucas-zimerman:
### Environment

How do you use Sentry?
SaaS and Self Hosted.

Which SDK and version?
latest.

### Steps to Reproduce
Only happens with the caching feature.

run the snippet
```csharp
using System;
using System.Globalization;
using System.Text;
using System.Threading;
using System.Threading.Tasks;
using Sentry;
using Sentry.Infrastructure;

SentrySdk.Init(o =>
{
    o.Dsn = "https://eb18e953812b41c3aeb042e666fd3b5c@o447951.ingest.sentry.io/5428537";
    o.Debug = true;
    o.DiagnosticLogger = new TraceDiagnosticLogger(SentryLevel.Debug); //only used for revealing the data on the IDE.
    o.CacheDirectoryPath = "D:/bob";
});


SentrySdk.ConfigureScope(scope =>
{
    scope.Contexts["MyCoolContext"] = null;
    scope.Contexts["MyCoolContext2"] = null;
});
SentrySdk.CaptureMessage("Hello Wôrld");
await Task.Delay(1000);
```

### Expected Result
An event sent to Sentry.

### Actual Result

```
  Error: Sentry rejected the envelope 3e2862cff51d43b3860e4513ef522764. Status code: BadRequest. Error detail: invalid event envelope. Error causes: unexpected end of file.
  Debug: Failed envelope '3e2862cff51d43b3860e4513ef522764' has payload:
{"sdk":{"name":"Sentry","version":"3.12.0-alpha.1"},"event_id":"3e2862cff51d43b3860e4513ef522764"}
{"type":"event","length":1965}
[...]

PR #1932, opened 2022-09-19
Support sending null Context values

SimonCropp:
potential fix for #1332


mattjohnsonpint:
Let's go with the other approach instead.  See my notes on the issue.  Thanks.

Issue #1350, opened 2021-11-25
Suspicious warning messages with tracing enabled (Trying to close a span that was already garbage collected)


ZlobnyiSerg:
### Environment
```
<PackageReference Include="Sentry" Version="3.11.1" />
<PackageReference Include="Sentry.AspNetCore" Version="3.11.1" />
<PackageReference Include="Sentry.Extensions.Logging" Version="3.11.1" />
.net 6
```

### Steps to Reproduce

1. Configure aspnet project to use Sentry tracing (`app.UseSentryTracing()`)
2. Create controller that creates EF DB context and queries something from DB
3. Call this controller

### Expected Result
No warnings in log file from Sentry

### Actual Result
Lot of warning messages coming from Sentry:
```
[11:40:07 WRN] Trying to close a span that was already garbage collected. Connection
[11:40:07 WRN] Trying to close a span that was already garbage collected. QueryExecution
[11:40:07 WRN] Trying to close a span that was already garbage collected. QueryCompiler
```

I'm not sure if this is critical for functioning, but this makes noise in logs (with default configuration). I had to raise level for Sentry to Error.
If this is intended behaviour, please close this issue.

bruno-garcia:
Thanks for raising @ZlobnyiSerg 
Those spans should have been closed before so something we need to look at.

lucas-zimerman:
#1368 will resolve this problem.

bruno-garcia:
Should be fixed on 3.12.1.
Please reopen otherwise.

ZlobnyiSerg:
I've updated to 3.12.1, now we have less warnings but they're still there. For example:
```
[19:52:55 INF] Route matched with {action = "Update", controller = "Customers", page = ""}. Executing controller action with signature System.Threading.Tasks.Task`1[Microsoft.AspNetCore.Mvc.IActionResult] Update(Int32, Model, System.Threading.CancellationToken) on controller Checkpoint.Web.Api.Controllers.CustomersController (Checkpoint.Web.Api).
[19:52:55 WRN] Trying to close a span that was already garbage collected. Connection
[19:52:55 INF] Envelope '65dc851d09894611b90e90e9d2abef20' successfully received by Sentry.
[19:52:56 WRN] Trying to close a span that was already garbage collected. Connection
[19:52:56 WRN] Trying to close a span that was already garbage collected. Connection
[19:52:56 WRN] Trying to close a span that was already garbage collected. Connection
[...]

PR #1933, opened 2022-09-19
respect IsSampled in SqlListener

SimonCropp:


src/Sentry.DiagnosticSource/Internals/DiagnosticSource/SentrySqlListener.cs
```diff
@@ -90,7 +90,12 @@ private void AddSpan(SentrySqlSpanType type, string operation, KeyValuePair<stri
         {
             _hub.ConfigureScope(scope =>
             {
-                if (scope.Transaction is not { } transaction)
+                if (scope.Transaction is not {} transaction)
+                {
+                    return;
+                }
+
+                if (transaction.IsSampled != true)
```
lucas-zimerman, line 98: Just wondering, why not 
```suggestion
                if (scope.Transaction is not {} transaction || transaction.IsSampled != true)
```

SimonCropp:
@lucas-zimerman yep. good point

src/Sentry.DiagnosticSource/Internals/DiagnosticSource/SentrySqlListener.cs
```diff
@@ -103,7 +103,7 @@ private void AddSpan(SentrySqlSpanType type, string operation, KeyValuePair<stri
                 }
 
                 if (type == SentrySqlSpanType.Execution &&
-                    value.GetProperty<Guid>(ConnectionKey) is { } connectionId)
+                    value.GetProperty<Guid>(ConnectionKey) is var connectionId)
```
mattjohnsonpint, line 106: `is var` is just an assignment, not a null check.  But also `Guid` is a struct, so it will never be null.  You could assign and then compare to `default` or `Guid.Empty` if you that's a valid use case, but as-is, there's no point in it being in the `if` expression because it will always be true.

src/Sentry.DiagnosticSource/Internals/DiagnosticSource/SentrySqlListener.cs
```diff
@@ -90,7 +90,7 @@ private void AddSpan(SentrySqlSpanType type, string operation, KeyValuePair<stri
         {
             _hub.ConfigureScope(scope =>
             {
-                if (scope.Transaction is not { } transaction)
+                if (scope.Transaction is not {} transaction || transaction.IsSampled != true)
```
mattjohnsonpint, line 93: I'm not certain, but should we actually be doing this check here? I though the sampling decision was made later.

src/Sentry.DiagnosticSource/Internals/DiagnosticSource/SentrySqlListener.cs
mattjohnsonpint: I should be clearer.

If `Guid.Empty` is valid, then do:

```csharp
if (type == SentrySqlSpanType.Execution) {
  var connectionId = value.GetProperty<Guid>(ConnectionKey);
  // ...
```

If `Guid.Empty` is not valid, then do:

```csharp
if (type == SentrySqlSpanType.Execution &&
    value.GetProperty<Guid>(ConnectionKey) is var connectionId &&
    connectionId != default)
{
    // ...
```

mattjohnsonpint:
Not sure I understand the reason for this PR.  Please add a better description to the top of this PR, or reference an issue if there is one open.  Thanks.

SimonCropp:
when looking into fixing https://github.com/getsentry/sentry-dotnet/issues/1350. i noticed that `SentryEFCoreListener` does a `IsSampled` check that that point. there are even explicit tests for that check. given `SentrySqlListener` should follow the same pattern, i assumed it was missing. 

did i assume wrong?

src/Sentry.DiagnosticSource/Internals/DiagnosticSource/SentrySqlListener.cs
mattjohnsonpint: I did some research and I was mistaken.  The sampling decision happens as the transaction is *started*.  So checking it here is an optimization, to not go to the effort of adding spans when we already know the transaction won't be sent anyway.

mattjohnsonpint, line 93:
```suggestion
                if (scope.Transaction is not {IsSampled: true} transaction)
```

mattjohnsonpint:
After looking into this further, I see that the sampling decision is originally made in `StartTransaction`, and then later it's used to filter which get sent to Sentry.  So anything in between is an optimization, in that there's no purpose in capturing a span if it's part of a transaction that is going to be discarded.  So this doesn't fix anything in particular, but it is still good to have.

mattjohnsonpint (approved):
One minor suggestion, but looks good otherwise.

src/Sentry.DiagnosticSource/Internals/DiagnosticSource/SentrySqlListener.cs
mattjohnsonpint: We can revisit this separately.  There's a few other parts of this file that have the same issue.  Thanks.

Issue #1944, opened 2022-09-23
Support 'long' built-in type for message template parameter in Sentry.Extensions.Logging


babyjohn42:
### Problem Statement

When using one of the message template overloads of `ILogger` (e.g. `logger.Log(LogLevel.Error, "I'm a {message}", "user");`), the Sentry logger handles the template parameters by adding tags to the event for each parameter if it is one of several data types. Today, the data types supported are `string`, `int`, `float`, `double`, and `Guid`. However, the `long` data type is commonly used for database identifiers of large tables, and it would be really nice to use them as is rather than have to call `.ToString()` on them throughout a codebase.

Right now, this code will not add a tag for "myId" to the Sentry event:
```
long id = 42;
logger.Log(LogLevel.Error, "Problem related to {myId}", id);
```
but this will:
```
long id = 42;
logger.Log(LogLevel.Error, "Problem related to {myId}", id.ToString());
```

### Solution Brainstorm

In `Sentry.Extensions.Logging.SentryLogger`, the following code added to the `Log` method would add this functionality if desired:

```
        // omitted for brevity

        else if (property.Value is int integerTagValue)
        {
            @event.SetTag(property.Key, integerTagValue.ToString());
        }
        // new else if block below
        else if (property.Value is long longIntegerTagValue)
        {
            @event.SetTag(property.Key, longIntegerTagValue.ToString());
        }
        // end new code
        else if (property.Value is float floatTagValue)
        {
            @event.SetTag(property.Key, floatTagValue.ToString());
        }

        // omitted for brevity

PR #1945, opened 2022-09-23
Support more types for message template tags in SentryLogger

babyjohn42:
Adds support for all primitive types (long integers, etc.), and enums when converting objects to strings to be used for tags.  Applies to the `SentyLogger` used by `Sentry.Extensions.Logging` and packages that depend on it such as ASP.NET Core.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@
 - Exception filters should consider child exceptions of an `AggregateException` ([#1924](https://github.com/getsentry/sentry-dotnet/pull/1924))
 - Add Blazor WASM detection to set IsGlobalModeEnabled to true ([#1931](https://github.com/getsentry/sentry-dotnet/pull/1931))
 - Respect Transaction.IsSampled in SqlListener ([#1933](https://github.com/getsentry/sentry-dotnet/pull/1933))
+- Support `long` for message template tags in SentryLogger ([#1945](https://github.com/getsentry/sentry-dotnet/pull/1945))
 
 ## 3.21.0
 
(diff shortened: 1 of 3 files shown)


babyjohn42:
Implements #1944 if desired

SimonCropp:
@babyjohn42 thanks. this will be included in the next release

SimonCropp (approved):

mattjohnsonpint (approved):

mattjohnsonpint:
I added a changelog entry.  Looks good otherwise.  Thanks!

SimonCropp:
> I added a changelog entry. Looks good otherwise. Thanks!

@mattjohnsonpint lol. u just beat me t it

SimonCropp:
@mattjohnsonpint should we also support short and the unsigned variants?

mattjohnsonpint:
> @mattjohnsonpint should we also support short and the unsigned variants?

Actually, now that I think about it, it should probably support any primitive or enum value.  It shouldn't support other object types though, because they are tags.  Complex objects don't make sense for tags.  I'll make a small adjustment to this PR.

babyjohn42:
Thanks all, I definitely agree that the primitive types make a ton of sense to be handled as tags in this way. I just wasn't sure if there were concerns with how Sentry.Extensions.Logging handled primitive types versus the other logging integrations, so I only included `long` for now. Thanks for the feedback and adjustment!

mattjohnsonpint:
Having some issues with our PR validation pipeline, but this will be merged once resolved.  Thanks.

PR #1949, opened 2022-09-26
tags should not differ based on current culture

SimonCropp:
I assume?

### DIFF
--- a/src/Sentry.Extensions.Logging/SentryLogger.cs
+++ b/src/Sentry.Extensions.Logging/SentryLogger.cs
@@ -1,3 +1,4 @@
+using System.Globalization;
 using Microsoft.Extensions.Logging;
 using Sentry.Infrastructure;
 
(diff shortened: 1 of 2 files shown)


SimonCropp:
@mattjohnsonpint i refactored out the `CreateEvent` method to do simple asserts since the the NSubstitute approach was giving me useless errors on build server

https://github.com/getsentry/sentry-dotnet/actions/runs/3124331669/jobs/5067590943
```
  Failed Sentry.Extensions.Logging.Tests.SentryLoggerTests.Culture_does_not_effect_tags [18 ms]
  Error Message:
   NSubstitute.Exceptions.ReceivedCallsException : Expected to receive exactly 1 call matching:
	CaptureEvent(e => (((e.Tags.get_Item("fooInteger") == "12345") AndAlso (e.Tags.get_Item("fooDouble") == "12345.123")) AndAlso (e.Tags.get_Item("fooFloat") == "12345.123")), <null>)
Actually received no matching calls.
Received 1 non-matching call (non-matching arguments indicated with '*' characters):
	CaptureEvent(*SentryEvent*, <null>)
```

note it give `*SentryEvent*` and none of the actual tags that were received. no i have it working i am happy to revert back the `CreateEvent` method. or would you prefer i simplify the other `SentryLoggerTests` to use `CreateEvent`.  or is there an NSubstitute  tweak we can use to make the exception message better

mattjohnsonpint:
I updated this to use `Convert.ToString(property.Value, CultureInfo.InvariantCulture)` to merge with change made in #1945.

The refactor is fine. 

Thanks!

mattjohnsonpint:
I see now why you were using "R".  However, the docs recommend "G9" for float and "G17" for double instead.  Updated accordingly.

mattjohnsonpint:
Actually, after seeing more test fails and thinking this through better - it would be inappropriate to do anything other than the default generic representation.  Otherwise we'd be tagging with lots of extra decimal places just due to floating point issues.

Ultimately, the test for `12345.123` isn't a fair test for single precision floating point, since only 7 digits of precision are guaranteed.  The actual representation is `12345.12305`.

I removed the extra checks and updated the test.

SimonCropp:
@mattjohnsonpint thanks for finishing up this one

Issue #1840, opened 2022-08-04
Dynamic Sampling support


bruno-garcia:
### Problem Statement

Similar to the Java issue: https://github.com/getsentry/sentry-java/issues/2205

https://develop.sentry.dev/sdk/performance/dynamic-sampling-context/

### Solution Brainstorm

* Add support for baggage propagation for HTTP libraries (HttpMessageHandler).
* Read incoming baggage from requests in ASP.NET and ASP.NET Core
* Make sure envelopes sent to Sentry include the trace context info needed for DS.

bruno-garcia:
A few notes:

* There's no need to implement any of the `freeze` specs that came up earlier. That was trying to solve problems in JS where the transaction name changes in the middle of the transaction and result in transactions downstream containing different transactions name.
* Need to add `tracePropagationTargets` to control what endpoints we should send the trace context

The spec should be up-to-date with all changes

mattjohnsonpint:
Just FYI for anyone who may have been following Sentry's product announcements about dynamic sampling - yes, we are working on adding it to the .NET SDK.  It should be available in the next release.  You can watch this issue for updates.  Thanks for your patience! 😃

mattjohnsonpint:
This will be included in the 3.22.0 release (coming soon).

PR #1953, opened 2022-09-28
Support Dynamic Sampling

mattjohnsonpint:
Adding .NET SDK support for dynamic sampling, per #1840 

Includes:
- `DynamicSamplingContext`
- `BaggageHeader`
- `TracePropagationTarget`
- Outbound headers sent with `SentryHttpMessageHandler`
- Inbound headers respected in ASP.NET and ASP.NET Core tracing integrations
- Lots of tests


src/Sentry/BaggageHeader.cs
```diff
@@ -0,0 +1,135 @@
...
+        internal const string SentryKeyPrefix = "sentry-";
+
+        // https://www.w3.org/TR/baggage/#baggage-string
+        // "Uniqueness of keys between multiple list-members in a baggage-string is not guaranteed."
+        // "The order of duplicate entries SHOULD be preserved when mutating the list."
+
+        public IReadOnlyList<KeyValuePair<string, string>> Members { get; }
+
+        private BaggageHeader(IEnumerable<KeyValuePair<string, string>> members) =>
+            Members = members.ToList().AsReadOnly();
```
SimonCropp, line 25: given Members is internal. and we dont need to care about people bypassing the `IReadOnlyList` contract. we can omit the `.AsReadOnly()`

src/Sentry/BaggageHeader.cs
```diff
@@ -0,0 +1,135 @@
...
+#else
+                    g => g.Key.Substring(SentryKeyPrefix.Length),
+#endif
+                    g => g.First());
+
+        /// <summary>
+        /// Creates the baggage header string based on the members of this instance.
+        /// </summary>
+        /// <returns>The baggage header string.</returns>
+        public override string ToString()
```
SimonCropp, line 45: where is the ToString used?

src/Sentry/BaggageHeader.cs
```diff
@@ -0,0 +1,135 @@
...
+            var items = baggage.Split(',', StringSplitOptions.RemoveEmptyEntries);
+            var members = new List<KeyValuePair<string, string>>(items.Length);
+
+            foreach (var item in items)
+            {
+                // Per baggage spec, the value may contain = characters, so limit the split to 2 parts.
+                var parts = item.Split('=', 2);
+                if (parts.Length != 2)
+                {
+                    // malformed, missing separator, key, or value
```
SimonCropp, line 80: should we log for the two malformed cases?

src/Sentry/BaggageHeader.cs
```diff
@@ -0,0 +1,135 @@
...
+        // "The order of duplicate entries SHOULD be preserved when mutating the list."
+
+        public IReadOnlyList<KeyValuePair<string, string>> Members { get; }
+
+        private BaggageHeader(IEnumerable<KeyValuePair<string, string>> members) =>
+            Members = members.ToList().AsReadOnly();
+
+        // We can safely return a dictionary of Sentry members, as we are in control over the keys added.
+        // Just to be safe though, we'll group by key and only take the first of each one.
+        public IReadOnlyDictionary<string, string> GetSentryMembers() =>
```
SimonCropp, line 29: given this is a new instance every time, could return a` Dictionary<string, string>`

src/Sentry/DynamicSamplingContext.cs
```diff
@@ -0,0 +1,143 @@
...
+        public IReadOnlyDictionary<string, string> Items { get; }
+
+        public bool IsEmpty => Items.Count == 0;
+
+        private DynamicSamplingContext(IReadOnlyDictionary<string, string> items) => Items = items;
+
+        /// <summary>
+        /// Gets an empty <see cref="DynamicSamplingContext"/> that can be used to "freeze" the DSC on a transaction.
+        /// </summary>
+        public static readonly DynamicSamplingContext Empty = new(new Dictionary<string, string>().AsReadOnly());
```
SimonCropp, line 23: remove the `.AsReadOnly()`

src/Sentry/BaggageHeader.cs
SimonCropp, line 45: not a fain of overriding string. would prefer to make it explicit as to the use case. eg `ToHeaderString`

test/Sentry.Tests/BaggageHeaderTests.cs
```diff
@@ -0,0 +1,172 @@
...
+        };
+
+        Assert.Equal(expected, merged.Members);
+    }
+
+    private static SettingsTask VerifyHeader(BaggageHeader header)
+    {
+        return Verify(header.Members)
+            .DontScrubGuids()
+            .AddExtraSettings(x => x.Converters.Add(new SentryIdConverter()));
```
SimonCropp, line 162: i assume we could make this a global settign?

test/Sentry.AspNetCore.Tests/SentryTracingMiddlewareTests.cs
```diff
@@ -148,6 +156,202 @@ public async Task Transaction_is_started_automatically_from_incoming_trace_heade
...
+            .Configure(app =>
+            {
+                app.UseRouting();
+                app.UseSentryTracing();
+
+                app.UseEndpoints(routes =>
+                {
+                    routes.Map("/person/{id}", _ =>
+                    {
+                        transaction = hub.GetSpan() as TransactionTracer;
```
SimonCropp, line 330: can this be a direct cast?

SimonCropp:
would any of this be useful https://github.com/VerifyTests/Verify.Http ?

src/Sentry/BaggageHeader.cs
mattjohnsonpint: Generally we want to not add to the result, so I feel `IReadOnlyDictionary` is warranted.

src/Sentry/BaggageHeader.cs
mattjohnsonpint: I hear ya, but I think I'll keep it on `ToString`, as it helps when debugging and also we already are doing that for `SentryTraceHeader` and a few other similar objects.  Thanks.

src/Sentry/DynamicSamplingContext.cs
mattjohnsonpint: Why?  It ensures it can't ever be mutated, and also there's only one instance so there's not much overhead.

src/Sentry/HubExtensions.cs
```diff
@@ -191,5 +192,14 @@ public static SentryId CaptureException(this IHub hub, Exception ex, Action<Scop
...
                     },
                     configureScope);
+
+        internal static ITransaction StartTransaction(
+            this IHub hub,
+            ITransactionContext context,
+            IReadOnlyDictionary<string, object?> customSamplingContext,
+            DynamicSamplingContext? dynamicSamplingContext)
+            => hub is Hub fullHub
+                ? fullHub.StartTransaction(context, customSamplingContext, dynamicSamplingContext)
```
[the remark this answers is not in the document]
mattjohnsonpint: There's only one condition being checked, and only one method being called on each path. So I'll disagree.

The one above it, definitely.  but I'll save that for future refactoring. 😉

mattjohnsonpint:
> would any of this be useful https://github.com/VerifyTests/Verify.Http ?

In general, yes.  But not at this time.  Thanks.

test/Sentry.AspNetCore.Tests/allsettings.json
```diff
@@ -8,8 +8,14 @@
...
-    "SampleRate": "1",
+    "SampleRate": 1.0,
     "Release": "7f5d9a1",
-    "Environment": "Staging"
+    "Environment": "Staging",
+    "TracesSampleRate": 1.0,
+    "TracePropagationTargets": [
+      "foo",
+      "bar",
+      "regex:^abc.*ghi$"
```
mattjohnsonpint, line 18: I hope this `regex:` prefix makes sense?  I didn't see any other simple way to add these to a json config file and still be able to distinguish between regex and substring.

src/Sentry.AspNet/HttpContextExtensions.cs
```diff
@@ -37,8 +64,9 @@ public static ITransaction StartSentryTransaction(this HttpContext httpContext)
     {
         var method = httpContext.Request.HttpMethod;
         var path = httpContext.Request.Path;
+        var options = SentrySdk.CurrentOptions;
```
marandaneto, line 67: Can `CurrentOptions` return null? on Java and friends, we have a NoOpOptions in case there's nothing bound to the Hub or no Hub at all, so we don't need to deal with NPEs.
Since .NET has explicitly nullability, that should be fine.

src/Sentry.AspNetCore/SentryTracingMiddleware.cs
```diff
@@ -50,6 +50,30 @@ internal class SentryTracingMiddleware
             }
         }
 
+        private BaggageHeader? TryGetBaggageHeader(HttpContext context)
```
marandaneto, line 53: Is there any difference between this one and the one in `HttpContextExtensions`? If not, maybe we can extract to an extension method and reuse the `BaggageHeader` in different places?

src/Sentry/BaggageHeader.cs
```diff
@@ -0,0 +1,150 @@
...
+
+        private static bool IsValidKey(string key)
+        {
+            if (key.Length == 0)
+            {
+                return false;
+            }
+
+            // The rules are the same as for HTTP headers.
+            // TODO: Is this public somewhere in .NET we can just call?
```
marandaneto, line 142: Do we know that already?

src/Sentry/Protocol/Envelopes/Envelope.cs
```diff
@@ -17,10 +17,6 @@ namespace Sentry.Protocol.Envelopes
     /// </summary>
     public sealed class Envelope : ISerializable, IDisposable
     {
-        private const string SdkKey = "sdk";
-        private const string EventIdKey = "event_id";
-        private const string SentAtKey = "sent_at";
```
marandaneto, line 22: Just out of curiosity, why not a `const` anymore?

src/Sentry/SentryOptions.cs
```diff
@@ -574,6 +574,16 @@ public double TracesSampleRate
...
 
+        /// <summary>
+        /// A customizable list of <see cref="TracePropagationTarget"/> objects, each containing either a
+        /// substring or regular expression pattern that can be used to control which outgoing HTTP requests
+        /// will have the <c>sentry-trace</c> and <c>baggage</c> headers propagated, for purposes of distributed tracing.
+        /// The default value is <c>null</c>, which indicates that the headers will be propagated to ALL targets.
+        /// To disable propagation completely, set this value to an empty collection.
+        /// </summary>
+        /// <seealso href="https://develop.sentry.dev/sdk/performance/#tracepropagationtargets"/>
+        public ICollection<TracePropagationTarget>? TracePropagationTargets { get; set; } = null;
```
marandaneto, line 585: Other SDKs just do a plain list of Strings that can be either a String or a Regex.
This solution is nice because forces the user to tell, whether this is a String or a Regex, so the SDK does not need to handle it blindly.
Downside is more types, because of bundle size, always, trade-offs =D

test/Sentry.AspNetCore.Tests/allsettings.json
marandaneto: Can you rather do an internal try catch to figure out if its a valid Regex or not?
On Dart, we can do catch the `FormatException` type.

marandaneto: I actually didn't handle that myself https://github.com/getsentry/sentry-dart/pull/1043/files

mattjohnsonpint: Per our discussion, I will update the behavior to:

- Remove need for prefix
- Always try substring matching first
- If fails substring matching, try regex matching

src/Sentry/SentryOptions.cs
mattjohnsonpint: Bigger downside is that we couldn't take advantage of [`RegexOptions.Compiled`](https://learn.microsoft.com/en-us/dotnet/standard/base-types/regular-expression-options#compiled-regular-expressions).  We'd be essentially re-parsing the regex pattern for every request.

I have adjusted the implementation to more closely resemble the others, but I'll still keep the custom type so I can build the compiled regex once on app startup.

src/Sentry/Protocol/Envelopes/Envelope.cs
mattjohnsonpint: This was part of a refactoring not specifically to dynamic sampling, but came up when adding the `"trace"` key.  I noted several existing inconsistencies, such as not having constants for `"name"` and `"version"` keys.  The decision was either add them for all, or none.  In general, we've decided to inline most key names in other parts of the SDK, so made sense to do it here as well.

src/Sentry/BaggageHeader.cs
mattjohnsonpint: I could only find it internally (on the .NET source link given).  Not on any publicly available method.

src/Sentry.AspNet/HttpContextExtensions.cs
mattjohnsonpint: Yes, it is marked as `SentryOptions?` and can return null. It's passed down mostly for logging purposes, so in the case of null, just no debug logs will be written.  That would only happen in testing.

src/Sentry.AspNetCore/SentryTracingMiddleware.cs
mattjohnsonpint: This one is for ASP.NET Core, and the other one is for ASP.NET (framework).

The only real difference between them is in how we retrieve the header from the request, which differs between the two flavors of ASP.NET.

However, we don't currently share code between them in a way that would be useful here.

SimonCropp (approved):

src/Sentry/SentryOptions.cs
marandaneto: TIL compiled regexes, Dart is compiled in the ctor.

Issue #1523, opened 2022-03-14
Sentry. Serilog. Envelopes which are constructed from"Serilog style" log messages cause 400 response from server


vlaterz:
### Environment

How do you use Sentry?
Sentry SaaS

Which SDK and version?
.NET 5, Sentry.Serilog 3.14.1, Sentry sdk 3.13.0

### Steps to Reproduce

I use Serilog with Sentry to write real time logs with caching (if connection is lost). 
1. I configure Sentry sink:
```
x.WriteTo.Sentry(o =>
                        {
                            o.MinimumBreadcrumbLevel = Serilog.Events.LogEventLevel.Debug;
                            o.MinimumEventLevel = Serilog.Events.LogEventLevel.Error;
                            o.Dsn = "http://e87a7650076d4287adf5fdd654bdeb64@10.0.225.69:9000/2";
                            o.AttachStacktrace = true;
                            o.TracesSampleRate = 1.0;
                            o.SampleRate = 1.0f;
                            o.MaxBreadcrumbs = 5;
                            o.DeduplicateMode = Sentry.DeduplicateMode.All;
                            o.MaxQueueItems = 10;
                            o.MaxCacheItems = 100;
                            o.InitCacheFlushTimeout = new TimeSpan(5000);
                            o.CacheDirectoryPath = Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.CommonApplicationData), "blah_blah", "SentryCachedLogs");
                            o.SendDefaultPii = true;
                        })
```
2. I try to log some info in Serilog style:
```
for (int i = 0; i < numOfRepeat; i++)
            {
                    _loggerError.Error(@"TestError:  a{i}, {time}", i, DateTime.Now);
            }
```
3. After running program logs are cached by Sentry and cleared from cache, as they have been sent, but they don`t get to server. Server responds with 400 status code 'Bad request'. 
Full error:  Error: Sentry rejected the envelope afc2b865ade74406bf7aa914c5095a95. Status code: BadRequest. Error detail: invalid event envelope. Error causes: missing newline after header or payload. 
At the same time Sentry Debug says  Debug: Successfully sent cached envelope: afc2b865ade74406bf7aa914c5095a95.

I tested sending simple strings as messages and  envelopes were successfully received by server.
Like that:
```
for (int i = 0; i < numOfRepeat; i++)
            {
                    _loggerError.Error($"TestError:  a{i}, {DateTime.Now}");
            }
```

### Expected Result

[...]

PR #1957, opened 2022-09-30
Always recalculate payload length

### LINKED ISSUE #1956: Envelope rejected when event has an `object` containing a `DateTime` or `DateTimeOffset` with a positive offset from UTC
This is a simplification of #1523

### Steps to Reproduce

```csharp
using var _ = SentrySdk.Init(o =>
{
    o.Dsn = // dsn
    o.Debug = true;
    o.CacheDirectoryPath = cacheDir;
});

var dto = new DateTimeOffset(2020, 1, 1, 0, 0, 0, TimeSpan.FromHours(1));
var evt = new SentryEvent {Message = "test"};
evt.SetExtra("foo", dto);

SentrySdk.CaptureEvent(evt);

Console.Read();
```

### Expected Result

Message should be sent to sentry successfully.

### Actual Result

```
Error: Sentry rejected the envelope 5243bf81806c4ff3ab7751749aced136. Status code: BadRequest. Error detail: invalid event envelope. Error causes: missing newline after header or payload.
```

This will occur anywhere we expose adding custom information in an `object` type, such as extras, custom context, message template params, etc.  It does not occur when the field in our object model is strongly typed as a `DateTime` or `DateTimeOffset`.  It doesn't occur with other data types (AFAIK).

mattjohnsonpint:
Fixes #1956 and #1523

### DIFF
--- a/src/Sentry/Protocol/Envelopes/EnvelopeItem.cs
+++ b/src/Sentry/Protocol/Envelopes/EnvelopeItem.cs
@@ -273,7 +273,7 @@ namespace Sentry.Protocol.Envelopes
             return new EnvelopeItem(header, new JsonSerializable(report));
         }
 
-        private static async Task<IReadOnlyDictionary<string, object?>> DeserializeHeaderAsync(
+        private static async Task<Dictionary<string, object?>> DeserializeHeaderAsync(
             Stream stream,
             CancellationToken cancellationToken = default)
         {
@@ -297,11 +297,6 @@ namespace Sentry.Protocol.Envelopes
             {
                 throw new InvalidOperationException("Envelope item header is malformed.");
             }
-
-            // Always remove the length header on deserialization so it will get re-calculated if later serialized.
-            // We cannot trust the length to be identical when round-tripped.
-            // See https://github.com/getsentry/sentry-dotnet/issues/1956
-            header.Remove(LengthKey);
 
             return header;
         }
@@ -403,6 +398,11 @@ namespace Sentry.Protocol.Envelopes
                     break;
                 }
             }
+
+            // Always remove the length header on deserialization so it will get re-calculated if later serialized.
+            // We cannot trust the length to be identical when round-tripped.
+            // See https://github.com/getsentry/sentry-dotnet/issues/1956
+            header.Remove(LengthKey);
 
             return new EnvelopeItem(header, payload);
         }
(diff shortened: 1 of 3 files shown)


src/Sentry/Protocol/Envelopes/EnvelopeItem.cs
```diff
@@ -320,9 +292,13 @@ await foreach (var curByte in stream.ReadAllBytesAsync(cancellationToken).Config
                 prevByte = curByte;
             }
 
-            return
-                Json.Parse(buffer.ToArray(), JsonExtensions.GetDictionaryOrNull)
-                ?? throw new InvalidOperationException("Envelope item header is malformed.");
+            var header = Json.Parse(buffer.ToArray(), JsonExtensions.GetDictionaryOrNull);
```
bruno-garcia, line 295: Is this a refactor or something changed?

src/Sentry/Protocol/Envelopes/EnvelopeItem.cs
```diff
@@ -128,77 +128,49 @@ await using (writer.ConfigureAwait(false))
...
+        public async Task SerializeAsync(Stream stream, IDiagnosticLogger? logger,
+            CancellationToken cancellationToken = default)
         {
-            // Length is known
-            if (TryGetLength() != null)
-            {
-                // Header
-                await SerializeHeaderAsync(stream, logger, cancellationToken).ConfigureAwait(false);
-                await stream.WriteByteAsync((byte)'\n', cancellationToken).ConfigureAwait(false);
+            // Always calculate the length of the payload, as Sentry will reject envelopes that have incorrect lengths
```
bruno-garcia, line 135: If debug mode is enabled should we assert the values we recalculated and the original match and otherwise log? It can help us find discrepancies in round trip serialization

src/Sentry/Protocol/Envelopes/EnvelopeItem.cs
mattjohnsonpint: Artifact between commits.  I'll undo.  Thanks

src/Sentry/Protocol/Envelopes/EnvelopeItem.cs
mattjohnsonpint: No, because it's *valid* for there to be such discrepancies.  That's what this issue is about, essentially.

SimonCropp (approved):

Issue #1657, opened 2022-05-15
CaptureException returns empty SentryId but exception is sent to Sentry


georgiosd:
### Package

Sentry.AspNetCore

### .NET Flavor

.NET Core

### .NET Version

6.0.5

### OS

Linux

### SDK Version

3.16.0

### Steps to Reproduce

No local repro

### Expected Result

SentryId should not be 00000000000000000000000000000000

### Actual Result

SentryId is 00000000000000000000000000000000

SimonCropp:
CaptureException will return an empty id if `SentrySdk.Init` has not been called.

georgiosd:
@SimonCropp I'm going to try but it's unlikely to not work in a minimal repro. In fact it's intermittent even now :-\

SimonCropp:
could you try enabling Diagnostic Logger https://docs.sentry.io/platforms/dotnet/configuration/diagnostic-logger/ 
see FileAppenderDiagnosticLogger  in that page as an example
and persist error and above.

if you need a help writing a custom DiagnosticLogger that targets serilog/nlog/log4net etc let me know

SimonCropp:
also, if you are occasionally getting it on your local dev machine when debugging, can u try enabling "break on all exceptions"

mattjohnsonpint:
In many cases you may have other logging enabled, and can just set `options.Debug = true` and send us any relevant output.

There shouldn't be any case where `SentryId.Empty` (all zeros)  gets sent to Sentry.  We should add a guard for this.

mattjohnsonpint:
Given this was reported for version `3.16.0`, it's possible we may have already fixed this, but still I think we should guard somewhere.

georgiosd:
Sorry for the quiet - I'm hammered right now but will retry to do the above steps before end of August.

georgiosd:
I'm only seeing this on older builds so I'll presume it's fixed! but like @mattjohnsonpint said, it would be a good idea to guard against it.

mattjohnsonpint:
@SimonCropp please add some tests to ensure we don't encounter this again. (or see if we already have them. )

thanks

mattjohnsonpint:
Closing, since we believe the bug is fixed already.  But we'll still proceed with adding some more tests and safeguards.

georgiosd:
@mattjohnsonpint I'm seeing this again for the first time since you closed the issue. 

Can we debug it collaboratively somehow?

[...]

Issue #307, opened 2019-11-12
System.TypeLoadException: Could not load type 'Microsoft.AspNetCore.Http.Internal.BufferingHelper'


CCC598:
- [ ] ASP.NET Core 3.0
- [ ] .NET Core 3.0
- [ ] Sentry.AspNetCore 1.2.0

In ASP.NET Core 3.0:
System.TypeLoadException: Could not load type 'Microsoft.AspNetCore.Http.Internal.BufferingHelper' from assembly 'Microsoft.AspNetCore.Http, Version=3.0.0.0, Culture=neutral, PublicKeyToken=adb9793829ddae60'.
   at Sentry.AspNetCore.SentryMiddleware.InvokeAsync(HttpContext context)
   at System.Runtime.CompilerServices.AsyncMethodBuilderCore.Start[TStateMachine](TStateMachine& stateMachine)
   at Sentry.AspNetCore.SentryMiddleware.InvokeAsync(HttpContext context)
   at Microsoft.AspNetCore.HostFiltering.HostFilteringMiddleware.Invoke(HttpContext context)
   at Microsoft.AspNetCore.Hosting.HostingApplication.ProcessRequestAsync(Context context)
   at Microsoft.AspNetCore.Server.IIS.Core.IISHttpContextOfT1.ProcessRequestAsync()

I look source code, this class is internal. 
https://source.dot.net/#Microsoft.AspNetCore.Http/Internal/BufferingHelper.cs

bruno-garcia:
Please use the `2.0.0-beta6`, it's fixed there.

It's not `2.0.0` yet because we'll fit a few more "breaking changes" into it, which unlikely will affect you, and if it does, the compiler would tell you.

bruno-garcia:
I'll close this since it's a dupe of #249 and it's fixed in the latest beta

PR #1958, opened 2022-10-03
Add Debug.Asserts for SentryId.Empty

SimonCropp:
fixes: #1657

@mattjohnsonpint i wasnt sure about putting explicit throws in given in 1657 it was successfully registered server side

note this doesnt really fix the problem. but will hopefully mean we notice this type of bug when in development


src/Sentry/SentryId.cs
```diff
@@ -31,22 +36,37 @@ namespace Sentry
...
         public bool Equals(SentryId other) => _guid.Equals(other._guid);
 
         /// <inheritdoc />
         public override bool Equals(object? obj) => obj is SentryId other && Equals(other);
 
         /// <inheritdoc />
-        public override int GetHashCode() => _guid.GetHashCode();
+        public override int GetHashCode()
+        {
+            AssertNotEmpty();
```
mattjohnsonpint, line 54: I believe this would interfere with comparison and equality operations.

mattjohnsonpint:
Need tests to pass.

SimonCropp:
@mattjohnsonpint is `API-32` an intermittent failure?

src/Sentry/SentryId.cs
mattjohnsonpint: Did some research and you are correct - it would assert when used in things like `HashSet<SentryId>` or `Dictionary<SentryId, object>`, etc.

I don't want to prevent adding `SentryId.Empty` to a dictionary - there may be a valid use for that now or in the future. We mainly want to avoid sending them to Sentry, not using them internally.

Also, I found: https://learn.microsoft.com/dotnet/api/system.object.gethashcode#notes-to-inheritors
Which says (among other things):

> The `GetHashCode()` method should not throw exceptions.

And a few various StackOverflow posts about that as well.

Let's leave it out of `GetHashCode`.

src/Sentry/SentryId.cs
```diff
@@ -14,12 +15,16 @@ namespace Sentry
...
-        public static readonly SentryId Empty = Guid.Empty;
+        public static readonly SentryId Empty = new();
 
         /// <summary>
         /// Creates a new instance of a Sentry Id.
         /// </summary>
-        public SentryId(Guid guid) => _guid = guid;
+        public SentryId(Guid guid)
+        {
+            Debug.Assert(guid != Guid.Empty, "Dont use this API with Guid.Empty. Instead use SentryId.Empty");
```
mattjohnsonpint, line 25: Since this is a struct, the result would be the same.  `SentryId.Empty == new SentryId(Guid.Empty)` via value equality, not reference equality.

mattjohnsonpint:
Thinking through this more, I think we should avoid this approach entirely.  `SentryId.Empty` is a valid value for a `SentryId`.  We just don't want to send them on outbound events.  Lets focus on #307 instead.

Thanks.

SimonCropp:
@mattjohnsonpint ok. no worries

PR #1959, opened 2022-10-03
more defensive on eventid header response

SimonCropp:
related to #1657

#skip-changelog

for the `event_id` header:

 * if missing return null (i am surprised this is valid?)
 * if null log error (throw in debug) and return null
 * if not a string log error (throw in debug) and return null. IMO this should always throw a exception
 * if fails to parse as a guid log error (throw in debug)  and return null. IMO this should always throw a exception 
 * if guid is empty log error (throw in debug) and return `SentryId.Empty`.  IMO this should always throw a exception 
 * else return new SentryId with the guid


src/Sentry/Protocol/Envelopes/Envelope.cs
```diff
@@ -43,12 +44,54 @@ public Envelope(IReadOnlyDictionary<string, object?> header, IReadOnlyList<Envel
...
-            Header.TryGetValue(EventIdKey, out var value) &&
-            value is string valueString &&
-            Guid.TryParse(valueString, out var guid)
-                ? new SentryId(guid)
-                : null;
+        [Obsolete]
+        public SentryId? TryGetEventId()
+        {
+            return TryGetEventId(null);
+        }
```
mattjohnsonpint, line 51:
```suggestion
        public SentryId? TryGetEventId()
        {
            var logger = SentrySdk.CurrentOptions?.DiagnosticLogger;
            return TryGetEventId(logger);
        }
```

(no need to obsolete)

src/Sentry/Protocol/Envelopes/Envelope.cs
```diff
@@ -43,12 +44,54 @@ public Envelope(IReadOnlyDictionary<string, object?> header, IReadOnlyList<Envel
...
+
+        /// <summary>
+        /// Attempts to extract the value of "event_id" header if it's present.
+        /// </summary>
+        public SentryId? TryGetEventId(IDiagnosticLogger? logger)
```
mattjohnsonpint, line 56:
```suggestion
        internal SentryId? TryGetEventId(IDiagnosticLogger? logger)
```

test/Sentry.Tests/Internals/Http/HttpTransportTests.cs
```diff
@@ -5,6 +5,7 @@
 using Sentry.Internal.Http;
 using Sentry.Testing;
 using Sentry.Tests.Helpers;
+#pragma warning disable CS0612
```
mattjohnsonpint, line 8: remove

CHANGELOG.md
```diff
@@ -24,6 +24,7 @@
 - Ignore null Context values ([#1942](https://github.com/getsentry/sentry-dotnet/pull/1942))
 - Tags should not differ based on current culture ([#1949](https://github.com/getsentry/sentry-dotnet/pull/1949))
 - Always recalculate payload length ([#1957](https://github.com/getsentry/sentry-dotnet/pull/1957))
+- More defensive on `event_id` header response ([#1959](https://github.com/getsentry/sentry-dotnet/pull/1959))
```
mattjohnsonpint, line 27: remove

mattjohnsonpint:
RE:
> if missing return null (i am surprised this is valid?)

Not all envelopes contain "events". In particular, a [session update](https://develop.sentry.dev/sdk/sessions/#session-update-payload) will not have an `event_id` of its own.  If a session is sent by itself, such as at the start of a session, then `TryGetEventId` should return `null`

mattjohnsonpint:
We *could* improve debug messages for such cases.  For example, the start of a session looks like this in the debug output:

```
  Debug: Envelope '' sent successfully. Payload: {"sdk":{"name":"sentry.dotnet","version":"3.21.0"},"sent_at":"2022-10-06T18:11:36.785535+00:00"} {"type":"session","length":293} {"sid":"6a09b56686834e67b5d8042c60ec2799","did":"4e01e65c-8951-4610-ac6b-7805e72269d9","init":true,"started":"2022-10-06T11:11:36.548873-07:00","timestamp":"2022-10-06T18:11:36.549708+00:00","seq":0,"duration":0,"errors":0,"attrs":{"release":"2022-10-06T11:11:36","environment":"development"}}
```

The message `Envelope '' sent successfully.` looks like a bug, but it's just that the envelope contains no event id.

Also the wording "Payload: " is wrong, because it's not just the payload of the envelope, but the entire serialized envelope including header and payload.

src/Sentry/Http/HttpTransportBase.cs
```diff
@@ -280,31 +280,56 @@ private void ExtractRateLimits(HttpHeaders responseHeaders)
...
+                else
+                {
+                    _options.LogInfo("Envelope '{0}' successfully received.", eventId);
+                }
+            }
+            else
+            {
+                if (eventId == null)
+                {
+                    _options.LogDebug("Envelope for session successfully received. Content:\n{1}", eventId, payload);
```
mattjohnsonpint, line 327: The `\n` will be replaced with ` ` due to #1756.  Might as well not put them.

Also, I believe they should all say `sent`, not `received`.

mattjohnsonpint: And I'm not entirely sure that `eventId` being absent *only* happens with sessions.  We should just omit `for session`.  No need to check.

SimonCropp: > being absent only happens with sessions

i did wonder about that. 

> We should just omit for session

done

SimonCropp: > I believe they should all say sent, not received.

this comes from the `HandleResponse`. so received seems more correct?? (happy to change it)

src/Sentry/Internal/BackgroundWorker.cs
```diff
@@ -142,8 +143,10 @@ private async Task DoWorkAsync()
                         }
                     }
 
-                    if (_queue.TryPeek(out var envelope)) // Work with the envelope while it's in the queue
+                    // Work with the envelope while it's in the queue
+                    if (_queue.TryPeek(out var envelope))
                     {
+                        var eventId = envelope.TryGetEventId(_options.DiagnosticLogger);
```
mattjohnsonpint, line 149: Probably should return null if no diagnostic logger to avoid extra parsing?

mattjohnsonpint: Should we use it on line 160 as well?

src/Sentry/Internal/BackgroundWorker.cs
```diff
@@ -317,7 +320,7 @@ private async Task SendFinalClientReportAsync(CancellationToken cancellationToke
                 catch (Exception exception)
                 {
                     _options.LogError("Error while sending final client report (event ID: '{0}').",
-                        exception, envelope.TryGetEventId());
+                        exception, envelope.TryGetEventId(_options.DiagnosticLogger));
```
mattjohnsonpint, line 323: And also here?

src/Sentry/Internal/BackgroundWorker.cs
SimonCropp: eventId  is now reused.

but i dont understand "Probably should return null if no diagnostic logger to avoid extra parsing"

mattjohnsonpint (approved):

src/Sentry/Internal/BackgroundWorker.cs
mattjohnsonpint: I meant given that `TryGetEventId` has work to do to get the event id out of the header, then we could skip it if not logging.  Nevermind though, not super important.

PR #1977, opened 2022-10-08
Bump test/sample dependency versions

mattjohnsonpint:
Brings all packages current in `/test` and `/samples` projects.

Additionally, resolves vulnerably package warnings from test and sample projects, as seen [here](https://github.com/getsentry/sentry-dotnet/actions/runs/3201317337/jobs/5229157184#step:5:1) (expand "List vulnerable packages")

Excludes changes required to resolve transitive vulnerability warnings in `Sentry.Google.Cloud.Functions` (source, samples, and tests).  Those will be fixed in a separate PR after fixed upstream.  See https://github.com/GoogleCloudPlatform/functions-framework-dotnet/issues/260

Also note our workflow should have failed on these.  Fix for that is in #1976.

#skip-changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,5 @@
 # Changelog
 
-## Unreleased
-
 ## 3.22.0
 
 ## Features
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,11 +2,6 @@
 
 ## Unreleased
 
-## Fixes
-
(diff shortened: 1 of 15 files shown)


mattjohnsonpint:
Looks like we'll have a fix shortly for the GCP functions issue, so holding off for just a bit.  Will revise when that's available.

SimonCropp (approved):

Issue #318, opened 2019-12-02
Unable to use ISentryEventExceptionProcessor with Scoped DI


udlose:
Framework used: ASP.NET Core

Runtime used: .NET Core Version: 2.2.7

NuGet packages used: Sentry.AspNetCore Version: 1.2.0

I get a Startup error with the following exception when I register `ISentryEventExceptionProcessor` as a Scoped instance.  **This works correctly if I register it as Transient or Singleton.**

**Exception:**

```
 An error occurred while starting the application.
 InvalidOperationException: Cannot resolve 'System.Collections.Generic.IEnumerable`1[Sentry.Extensibility.ISentryEventExceptionProcessor]' from root provider because it requires scoped service 'Sentry.Extensibility.ISentryEventExceptionProcessor'.
 Microsoft.Extensions.DependencyInjection.ServiceLookup.CallSiteValidator.ValidateResolution(Type serviceType, IServiceScope scope, IServiceScope rootScope)
 
 InvalidOperationException: Cannot resolve 'System.Collections.Generic.IEnumerable`1[Sentry.Extensibility.ISentryEventExceptionProcessor]' from root provider because it requires scoped service 'Sentry.Extensibility.ISentryEventExceptionProcessor'.
 Microsoft.Extensions.DependencyInjection.ServiceLookup.CallSiteValidator.ValidateResolution(Type serviceType, IServiceScope scope, IServiceScope rootScope)
 Microsoft.Extensions.DependencyInjection.ServiceProvider.Microsoft.Extensions.DependencyInjection.ServiceLookup.IServiceProviderEngineCallback.OnResolve(Type serviceType, IServiceScope scope)
 Microsoft.Extensions.DependencyInjection.ServiceLookup.ServiceProviderEngine.GetService(Type serviceType, ServiceProviderEngineScope serviceProviderEngineScope)
 Microsoft.Extensions.DependencyInjection.ServiceLookup.ServiceProviderEngine.GetService(Type serviceType)
 Microsoft.Extensions.DependencyInjection.ServiceProvider.GetService(Type serviceType)
 Microsoft.Extensions.DependencyInjection.ServiceProviderServiceExtensions.GetService<T>(IServiceProvider provider)
 Microsoft.AspNetCore.Builder.ApplicationBuilderExtensions.UseSentry(IApplicationBuilder app) in ApplicationBuilderExtensions.cs
[...]

PR #1979, opened 2022-10-08
Allow custom processors to be added as a scoped dependency

chrisza4:
Refer from issue #318 

We can fixed this by turning `SentryMiddleware` to be a [factory-based middleware](https://learn.microsoft.com/en-us/aspnet/core/fundamentals/middleware/extensibility?view=aspnetcore-6.0). But this also required some interfaces changes and now all custom event processors will be injected using `PopulateScope` during `OnEvaluation`.

One concern is that this changes might break behaviour of public method `GetAllEventProcessors` since now it can only be used after scope is evaluated. I'm not sure if that could be a concern.

Alternatively, we can move out from `onEvaluating` hook but I don't know if that is more useful so I just put it there first to make it easier to write test in same pattern (testing `PopulateScope`). Let me know if you have other opinion.


test/Sentry.AspNetCore.Tests/IntegrationsTests.EventProcessor.cs
```diff
@@ -41,10 +39,9 @@ public async Task SentryEventExceptionProcessor_Transient_LifetimeRespected()
         _ = await HttpClient.GetAsync("/throw");
         _ = await HttpClient.GetAsync("/throw");
 
-        // First resolve only to decided if worth patching SentryOptions
-        Assert.Equal(3, exceptionProcessorsResolved.Count);
-        exceptionProcessorsResolved[0].DidNotReceive().Process(Arg.Any<Exception>(), Arg.Any<SentryEvent>());
+        Assert.Equal(2, exceptionProcessorsResolved.Count);
+        // exceptionProcessorsResolved[0].DidNotReceive().Process(Arg.Any<Exception>(), Arg.Any<SentryEvent>());
```
chrisza4, line 43: Oops

mattjohnsonpint:
Thanks for this! Looks promising on first glance.  I'll review in more detail next week.

SimonCropp:
wouldn't this result in: processors explicitly added to options being ignored?

chrisza4:
Few test failed. Let me take a look.


@SimonCropp 
> wouldn't this result in: processors explicitly added to options being ignored?

As I read into [Sentry SDK doc](https://docs.sentry.io/platforms/dotnet/guides/aspnetcore/#adding-event-and-exception-processors), I thought we can only use processors via dependency injection. If there is other way then let me know where and I can take a look.

mattjohnsonpint:
During initialization, event processors can be added to the options:

```csharp
.UseSentry(options =>
{
    options.AddEventProcessor(new MyEventProcessor());
}
```

They can also be added for a given scope:

```csharp
{
    scope.AddEventProcessor(new MyEventProcessor());
});
```

Both of those would need to continue to work.

mattjohnsonpint:
Looking over the PR, the main issue I see is now *all* processors are registered on the scope.

Instead, let's control this based on how the processor was registered with DI.

- If registered using `AddSingleton`, then it should be applied globally - as if registered directly on the options during initialization.  This is the existing behavior.
- If registered using `AddScoped`, then it should be applied for the current scope - as if registered with `SentrySdk.ConfigureScope`.

mattjohnsonpint:
Actually, ignore my last comments.  There's no need to separate those paths out, as the effect would be the same as the approach you're currently taking.  In other words, registering the same instance of an event processor on each scope is identical to registering it once on options - since we gather both at the same time when we run them.

mattjohnsonpint:
I see we are doing this for event processors and exception processors.  Please also do the same for transaction processors added in #1861.  Looks like we missed this back then.  Thanks.

src/Sentry.AspNetCore/SentryMiddleware.cs
```diff
@@ -162,6 +165,17 @@ private void SyncOptionsScope(IHub newHub)
...
+        {
+            var toAdd = _eventProcessors.Except(scope.GetAllEventProcessors());
+            scope.AddEventProcessors(toAdd);
+        }
+        if (_eventExceptionProcessors != null)
+        {
+            var toAdd = _eventExceptionProcessors.Except(scope.GetAllExceptionProcessors());
+            scope.AddExceptionProcessors(toAdd);
+        }
+
```
mattjohnsonpint, line 178: The null checks are not necessary.

```suggestion
        scope.AddEventProcessors(_eventProcessors.Except(scope.GetAllEventProcessors()));
        scope.AddExceptionProcessors(_eventExceptionProcessors.Except(scope.GetAllExceptionProcessors()));
```

mattjohnsonpint (changes requested):

mattjohnsonpint:
@SimonCropp

> wouldn't this result in: processors explicitly added to options being ignored?

No, because `scope.GetAllEventProcessors()` returns both:

https://github.com/getsentry/sentry-dotnet/blob/77ccdb2dc01b85c1dc53604f9186f715f66b9ff8/src/Sentry/ScopeExtensions.cs#L18-L33

mattjohnsonpint:
Needs transaction processors added, and a changelog entry, then LGTM.  Thanks!

mattjohnsonpint (approved):

mattjohnsonpint:
I added the changelog entry.  We'll do transaction processors in a separate PR.  Thanks for your contribution!

PR #1996, opened 2022-10-17
Mark Transaction as aborted when unhandled exception occurs

### LINKED ISSUE #1975: Closing Unfinished Transactions
Consider:

```csharp
public void Main()
{
    SentrySdk.Init(...);
    var transaction = SentrySdk.StartTransaction("foo", "bar");
    DoSomething();
    transaction.Finish();
}
```

If `DoSomething` throws an unhandled exception, the transaction is never finished and thus not sent to Sentry.  We should consider whether we should catch this in and finish the transactions with an error status, so that information is not lost.

We *could* just do that in our unhandled exception handlers, or when capturing an exception that is marked unhandled (so it works with all integrations).  However, that doesn't take care of this possibility:

```csharp
public void Main()
{
    SentrySdk.Init(...);

    try
    {
        var transaction = SentrySdk.StartTransaction("foo", "bar");
        DoSomething();
        transaction.Finish();
    }
    catch (Exception e)
    {
    }
}
```

Of course, that should probably be written as:

```csharp
public void Main()
{
    SentrySdk.Init(...);

    try
    {
        var transaction = SentrySdk.StartTransaction("foo", "bar");
        DoSomething();
        transaction.Finish();
    }
    catch (Exception e)
    {
        transaction.Finish(SpanStatus.UnknownError);
    }
}
```

But we could also do that if we made it `IDisposable`, such as:

```csharp
public void Main()
{
    SentrySdk.Init(...);

    try
    {
        using var transaction = SentrySdk.StartTransaction("foo", "bar");
        DoSomething();
        transaction.Finish();
    }
    catch (Exception e)
    {
    }
}
```

... which then simplifies the original problem to:

```csharp
public void Main()
{
    SentrySdk.Init(...);

    using var transaction = SentrySdk.StartTransaction("foo", "bar");
    DoSomething();
    transaction.Finish();
}
```

Unfortunately, that would be a breaking change - and one that is easily missed.

An alternative API that we could add without breaking anything would be:

```csharp
public void Main()
{
    SentrySdk.Init(...);

    SentrySdk.WithTransaction("foo", "bar", transaction => {
        DoSomething();
        transaction.Finish();
    });
}
```

SimonCropp:
Fixes #1975


mattjohnsonpint:
@SimonCropp - not sure if you wanted to add anything else here, or if you're good with my changes.  Please review.  Thanks. 😄

SimonCropp:
rebased on main, and one small thing fixed. out of draft now

mattjohnsonpint (approved):

PR #2007, opened 2022-10-24
Add Order of execution tests

SimonCropp:
#skip-changelog

for several scenarios, records the order of execution of events, and the corresponding envelopes


SimonCropp:
@mattjohnsonpint ok need a second pair of eyes. why is this only failing in ci?

mattjohnsonpint:
Device tests will fail for Verify tests unless they are excluded with `[Trait("Category", "Verify")]`.  I need to find a more natural way to exclude them, but that's how it is presently.

For the others, I'm not sure - seems like sometimes you're getting events and sometimes you get none.  Maybe they're being sampled out when you don't expect them to be?  I'd have to debug further.

SimonCropp:
@mattjohnsonpint are you happy to take over this one?

mattjohnsonpint:
Yes

PR #2013, opened 2022-10-26
Add Custom Measurements API

### LINKED ISSUE #1951: Set custom measurements on transactions
See https://github.com/getsentry/team-mobile/issues/51

Includes https://github.com/getsentry/team-mobile/issues/31

mattjohnsonpint:
Adds support for Custom Performance Metrics
https://docs.sentry.io/product/performance/metrics/#custom-performance-metrics

Usage:

```csharp
// integer values can be int, long, or ulong
transaction.SetMeasurement("some_time", 50, MeasurementUnit.Duration.Millisecond);

// floating point values are passed as double
transaction.SetMeasurement("file_size", 1.21, MeasurementUnit.Information.Gigabyte);

// set measurement without unit is the same as passing MeasurementUnit.None
transaction.SetMeasurement("num_widgets", 10);

// custom units can be used as well
transaction.SetMeasurement("foo", 5000, MeasurementUnit.Custom("bar"));
```

Resolves #1951

### DIFF
--- a/src/Sentry/Transaction.cs
+++ b/src/Sentry/Transaction.cs
@@ -1,10 +1,12 @@
 using System;
 using System.Collections.Generic;
+using System.ComponentModel;
 using System.Linq;
 using System.Text.Json;
 using Sentry.Extensibility;
 using Sentry.Internal;
 using Sentry.Internal.Extensions;
+using Sentry.Protocol;
 
 namespace Sentry
 {
@@ -12,7 +14,7 @@ namespace Sentry
     /// <summary>
     /// Sentry performance transaction.
     /// </summary>
-    public class Transaction : ITransactionData, IJsonSerializable, IHasDistribution, IHasTransactionNameSource
+    public class Transaction : ITransactionData, IJsonSerializable, IHasDistribution, IHasTransactionNameSource, IHasMeasurements
     {
         /// <summary>
         /// Transaction's event ID.
@@ -184,6 +186,12 @@ namespace Sentry
         /// </summary>
         public IReadOnlyCollection<Span> Spans => _spans;
 
+        // Not readonly because of deserialization
+        private Dictionary<string, Measurement> _measurements = new();
+
+        /// <inheritdoc />
+        public IReadOnlyDictionary<string, Measurement> Measurements => _measurements;
+
         /// <inheritdoc />
         public bool IsFinished => EndTimestamp is not null;
 
@@ -261,6 +269,7 @@ namespace Sentry
             {
                 SampleRate = transactionTracer.SampleRate;
                 DynamicSamplingContext = transactionTracer.DynamicSamplingContext;
+                _measurements = transactionTracer.Measurements.ToDictionary();
             }
         }
 
@@ -279,6 +288,11 @@ namespace Sentry
         /// <inheritdoc />
         public void UnsetTag(string key) =>
             _tags.Remove(key);
+
+        /// <inheritdoc />
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public void SetMeasurement(string name, Measurement measurement) =>
+            _measurements[name] = measurement;
 
         /// <inheritdoc />
         public SentryTraceHeader GetTraceHeader() => new(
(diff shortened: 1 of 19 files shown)


src/Sentry/Transaction.cs
```diff
@@ -360,20 +379,19 @@ public static Transaction FromJson(JsonElement json)
...
-                _spans = json
-                    .GetPropertyOrNull("spans")?
-                    .EnumerateArray()
-                    .Select(Span.FromJson)
-                    .ToArray() ?? Array.Empty<Span>()
+                _breadcrumbs = breadcrumbs,
+                _extra = extra,
+                _tags = tags,
+                _measurements = measurements,
+                _spans = spans
```
mattjohnsonpint, line 391: Note, there's some formatting changes and refactoring here also.  The only new part is adding measurements.

marandaneto:
@mattjohnsonpint a small note:

```
// set measurement without unit is the same as passing MeasurementUnit.None
transaction.SetMeasurement("num_widgets", 10);
```

If no `unit` is given, it should not have a `unit` rather than defaulting to `MeasurementUnit.None`.
[Relay will set them to "none" as a default.](https://develop.sentry.dev/sdk/event-payloads/transaction/)

One should be able to `transaction.SetMeasurement("num_widgets", 10, MeasurementUnit.None);`

mattjohnsonpint:
@marandaneto - Within the SDK, `MeasuementUnit.None.ToString() == ""`, and `""` will omit sending the unit field during json serialization.  It won't ever send `"unit":"none"` or `"unit":""` in the json either way.

mattjohnsonpint:
Interesting... I read it differently I guess.  I will double check.  Thanks for highlighting that! 👍

mattjohnsonpint:
Hmmm...  I'm not so sure.

In https://getsentry.github.io/relay/relay_metrics/enum.MetricUnit.html#variant.None

> None
> Untyped value without a unit ("").

But in https://develop.sentry.dev/sdk/event-payloads/transaction/

> Custom measurements need units to be specified; if the units are missing, Relay will set them to "none" as a default. 

Both are docs for Relay, so one of them must be in error. 🤔

mattjohnsonpint:
Looking at the source code, it appears either `"none"` or `""` are both treated as `None`.

https://github.com/getsentry/relay/blob/91250fc1b52e66418830e7b6051887dbfea8662a/relay-common/src/constants.rs#L639

Thus, I conclude all of these would be identical:

```json
{"value":"123"}
{"value":"123","unit":""}
{"value":"123","unit":"none"}
```

marandaneto:
> Looking at the Relay source code, it appears either `"none"` or `""` are both treated as `None`.
> 
> [getsentry/relay@`91250fc`/relay-common/src/constants.rs#L639](https://github.com/getsentry/relay/blob/91250fc1b52e66418830e7b6051887dbfea8662a/relay-common/src/constants.rs#L639)
> 
> Thus, I conclude all of these would be identical:
> 
> ```json
> {"value":"123"}
> {"value":"123","unit":""}
> {"value":"123","unit":"none"}
> ```
> 
> (For the .NET SDK, I always use the first one.)

Yep, but that's what I got from the Ingest folks, Relay is a good citizen and accepts everything but the SDK is;

Prefer `{"value":"123"}` over `{"value":"123","unit":"none"}`, the later only if the user set `MeasuementUnit.None`.
`{"value":"123","unit":""}` isn't allowed by the SDKs, but it's accepted by Relay because of back-compatibility reasons most likely.
You can see that I've removed accepting `empty` on this PR https://github.com/getsentry/sentry-docs/pull/5598/files L54

mattjohnsonpint:
So I think the differences boil down to:

> the later only if the user set `MeasuementUnit.None`

I still don't understand why it would matter though.  Given two ways to serialize the data where one is 14 chars longer and they both have the same effect, I prefer the shorter form.

marandaneto:
> MeasuementUnit.None

`MeasuementUnit.None` is a valid unit and it should be sent as such.
Omitting the `unit` defaults to `none`, yes, but maybe tomorrow it'd be something else and this would be a breaking change for the former use case if doing what you've suggested.
The other reason is to align with all the other SDKs.
There was a DACI for this, so I'd not spend too much though on it.

mattjohnsonpint:
@marandaneto - I updated as you advised, to align with the behavior of the other SDKs.

- `transaction.SetMeasurement("foo", 123)` will send `"foo":{"value":"123"}`
- `transaction.SetMeasurement("foo", 123, MeasurementUnit.None)` will send `"foo":{"value":"123","unit":"none"}`

bitsandfoxes (approved):
🔥

PR #2014, opened 2022-10-26
Add `ISpan.GetTransaction` convenience method

mattjohnsonpint:
This adds a convenience method for when you have an `ISpan` and need to get the `ITransaction` it belongs to.

For example:

```csharp
void DoSomething(ISpan span)
{
    var transaction = span.GetTransaction();
}
```

This also works well anywhere we have a `GetSpan` method, such as `IHub` and `SentrySdk`

```csharp
var hub = serviceProvider.GetRequiredService<IHub>();
var transaction = hub.GetSpan()?.GetTransaction();
```

or

```csharp
var transaction = SentrySdk.GetSpan()?.GetTransaction();
```

### Motivation

With custom measurements added in #2013, we have this:

```csharp
transaction.SetMeasurement("some_time", 50, MeasurementUnit.Duration.Millisecond);
```

Since sometimes we only have a span, we need to do this:

```csharp
span.GetTransaction().SetMeasurement("some_time", 50, MeasurementUnit.Duration.Millisecond);
```

The alternative would have been to attach `SetMeasurement` extension methods on `ISpan` as well, but that would imply that measurement data is set on individual spans - which it is not.

This seems cleaner and has other usages as well.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 - Support DI for custom transaction processors ([#1993](https://github.com/getsentry/sentry-dotnet/pull/1993))
 - Mark Transaction as aborted when unhandled exception occurs ([#1996](https://github.com/getsentry/sentry-dotnet/pull/1996))
 - Build Windows and Tizen targets for `Sentry.Maui` ([#2005](https://github.com/getsentry/sentry-dotnet/pull/2005))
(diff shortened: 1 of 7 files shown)


mattjohnsonpint:
Some minor formatting changes in the PR as well.

SimonCropp (approved):

PR #2016, opened 2022-10-26
Update nuget dependencies

mattjohnsonpint:
Update package references to use [floating versions](https://learn.microsoft.com/nuget/concepts/dependency-resolution#floating-versions) in tools, tests, and samples.  For example, `1.*` instead of `1.2.3` (or `0.1.*` instead of `0.1.2`).  This should reduce the amount of updates we need to do to keep healthy.

We will still use specific minimal versions for projects under `/src`, as those affect consumers of our nuget packages.

Also bumps a few more transitive dependencies to resolve all security warnings.

#skip-changelog


mattjohnsonpint:
@SimonCropp - We can bump Verify to the next major in a separate PR.

SimonCropp (approved):

PR #2020, opened 2022-10-27
Use C# 10 everywhere

mattjohnsonpint:
We had the core of Sentry pinned to C# 9, while using C# 10 in the rest of the solution.  The comments say that's for Unity, but my current understanding is that shouldn't matter.

This PR updates the solution to use C# 10 throughout, and applies the following features:

- File-scoped namespace declarations, which reduce indenting (review with whitespace changes hidden)
- Implicit usings (just the ones we had already, removing extra)

I also took this opportunity to change all our `#if NET461` statements to `#if NETFRAMEWORK` for clarity and forward compatibility.

Some other minor formatting changes as well.  No significant code changes.

#skip-changelog


mattjohnsonpint:
Discussed with @bitsandfoxes, and he tested this against the current sentry-unity build.  It works fine.

We speculate that the original intent may have been to compile the Sentry code within the Unity editor itself, which [requires C# 9](https://docs.unity3d.com/Manual/CSharpCompiler.html).  However, we don't actually do that.  We build the entire solution with the regular .NET compiler and pull in the resulting assemblies as DLLs.  Thus, we can use any version of C# that is supported by the regular compiler.  That's C# 10 now, and leaves the door open for C# 11 in the future.

We still cannot use any language features from C# 8+ that require runtime support, such as default interface methods.

bitsandfoxes (approved):

PR #2030, opened 2022-11-03
Simplify API for flushing events

mattjohnsonpint:
This is mostly just for convenience, simplifying usage and documentation.

- Move the `FlushTimeout` option to `SentryOptions`.
  - Previously this was defined separately on `SentryAspNetCoreOptions` and `SentryNLogOptions` only.
  - Note, this is not a breaking change, as those classes inherit from `SentryOptions` already.
  - We default to 2 seconds, except for NLog where we preserve the existing 15 second default. 

- Add `FlushAsync()` APIs where we previously had `FlushAsync(timeout)`.
  - When a timeout is not specified, the value from the `FlushTimeout` option is used.

- Added synchronous versions, `Flush()` and `Flush(timeout)`.  They are blocking sync-over-async, but that's what we've been doing already anyway.  This reduces error and gets `.Wait()` or `.GetAwaiter().GetResult()` out of samples and docs.

- Updated existing usages throughout the solution
- Added tests


samples/Sentry.Samples.AspNetCore.Blazor.Wasm/Program.cs
```diff
@@ -28,6 +28,6 @@
 catch (Exception e)
 {
     SentrySdk.CaptureException(e);
-    await SentrySdk.FlushAsync(TimeSpan.FromSeconds(2));
+    await SentrySdk.FlushAsync();
```
bitsandfoxes, line 31: Since we're showcasing it here, does it make sense to add setting the timeout to the `Init`?

src/Sentry/ISentryClient.cs
```diff
@@ -45,9 +45,9 @@ public interface ISentryClient
     void CaptureSession(SessionUpdate sessionUpdate);
 
     /// <summary>
-    /// Flushes events queued up.
+    /// Flushes the queue of captured events until the timeout is reached.
     /// </summary>
-    /// <param name="timeout">How long to wait for flush to finish.</param>
+    /// <param name="timeout">The amount of time allow for flushing.</param>
```
bitsandfoxes, line 50:
```suggestion
    /// <param name="timeout">The amount of time allowed for flushing.</param>
```

src/Sentry/SentryClientExtensions.cs
```diff
@@ -61,12 +61,49 @@ public static void CaptureUserFeedback(this ISentryClient client, SentryId event
...
+    /// Blocks synchronously. Prefer <see cref="FlushAsync"/> in async code.
+    /// </remarks>
+    public static void Flush(this ISentryClient client) =>
+        client.FlushAsync().GetAwaiter().GetResult();
+
+    /// <summary>
+    /// Flushes the queue of captured events until the timeout is reached.
+    /// </summary>
+    /// <param name="client">The Sentry client.</param>
+    /// <param name="timeout">The amount of time allow for flushing.</param>
```
bitsandfoxes, line 79:
```suggestion
    /// <param name="timeout">The amount of time allowed for flushing.</param>
```

src/Sentry/SentryClientExtensions.cs
```diff
@@ -61,12 +61,49 @@ public static void CaptureUserFeedback(this ISentryClient client, SentryId event
...
+
+    /// <summary>
+    /// Flushes the queue of captured events until the timeout set in <see cref="SentryOptions.FlushTimeout"/>
+    /// is reached.
+    /// </summary>
+    /// <param name="client">The Sentry client.</param>
+    /// <returns>A task to await for the flush operation.</returns>
+    public static Task FlushAsync(this ISentryClient client)
+    {
+        var options = client.GetSentryOptions() ?? new SentryOptions();
```
bitsandfoxes, line 94: Is that the intended way? The new options would not have a DNS and I think that would get caught somewhere in the transport?

src/Sentry/SentrySdk.cs
```diff
@@ -116,8 +116,38 @@ internal static IDisposable UseHub(IHub hub)
...
+    /// is reached.
+    /// </summary>
+    /// <returns>A task to await for the flush operation.</returns>
+    [DebuggerStepThrough]
+    public static Task FlushAsync() => CurrentHub.FlushAsync();
+
+    /// <summary>
+    /// Flushes the queue of captured events until the timeout is reached.
+    /// </summary>
+    /// <param name="timeout">The amount of time allow for flushing.</param>
```
bitsandfoxes, line 149:
```suggestion
    /// <param name="timeout">The amount of time allowed for flushing.</param>
```

bitsandfoxes (approved):

samples/Sentry.Samples.AspNetCore.Blazor.Wasm/Program.cs
mattjohnsonpint: The flush is part being showcased, not the timeout.  In general, the point with most of this PR is that the developer either should set the timeout once in options and not think about it again, or have a very specific reason to deviate from the default.

src/Sentry/SentryClientExtensions.cs
mattjohnsonpint: That instance is only used temporarily, to retrieve the default timeout value in the rare case that the `GetSentryOptions` method returns null.  That would only happen if `ISentryClient` was something other than expected, such as a test substitute, and only if the `SentryOptionsForTestingOnly` property wasn't set.  So just a final check to get back the 2 second default without putting that in two places.

PR #1998, opened 2022-10-17
Update verify

SimonCropp:
#skip-changelog

PR #2031, opened 2022-11-04
Sort context during serialization

mattjohnsonpint:
In #1998 I updated Verify, and it highlighted an inconsistent sort order when serializing our `Contexts` object.  I speculate that's because it inherits from `ConcurrentDictionary`.

This PR fixes the problem, adds a test, and restores the part of the verify test I previously removed.

This only matters for snapshot testing - Sentry doesn't care what order contexts are sent in.

#skip-changelog

### DIFF
--- a/src/Sentry/Contexts.cs
+++ b/src/Sentry/Contexts.cs
@@ -111,8 +111,11 @@ public sealed class Contexts : ConcurrentDictionary<string, object>, IJsonSerial
     }
 
     /// <inheritdoc />
-    public void WriteTo(Utf8JsonWriter writer, IDiagnosticLogger? logger) =>
-        writer.WriteDictionaryValue(this!, logger, includeNullValues: false);
+    public void WriteTo(Utf8JsonWriter writer, IDiagnosticLogger? logger)
+    {
+        var contexts = this.OrderBy(x => x.Key, StringComparer.Ordinal);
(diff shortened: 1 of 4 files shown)


src/Sentry/Contexts.cs
```diff
@@ -111,8 +111,11 @@ internal void CopyTo(Contexts to)
     }
 
     /// <inheritdoc />
-    public void WriteTo(Utf8JsonWriter writer, IDiagnosticLogger? logger) =>
-        writer.WriteDictionaryValue(this!, logger, includeNullValues: false);
+    public void WriteTo(Utf8JsonWriter writer, IDiagnosticLogger? logger)
+    {
+        var contexts = this.OrderBy(x => x.Key, StringComparer.Ordinal);
```
bitsandfoxes, line 116: I assume the cost associated with sorting it all is negligent?

bitsandfoxes (approved):
LGTM!

src/Sentry/Contexts.cs
mattjohnsonpint: Verify test became flaky after updating, but this was the root cause.  Verify appearing to be be flaky just highlighted that we weren't getting consistent sort order here in the first place.

Cost is negligent - yes.  Contexts are pretty small anyway.

PR #2034, opened 2022-11-08
Update unobserved task exception integration

mattjohnsonpint:
We had a few extraneous null checks in the `TaskUnobservedTaskExceptionIntegration`.

- From https://github.com/dotnet/runtime/issues/32454 it appears that the exception will never be null.
- The `_hub` variable will never be null within `Handle`, because that can only occur after registering the integration.

Also, we have been calling this `TaskUnobservedTaskExceptionIntegration` with the word `Task` twice.  The exception raised is called `UnobservedTaskException`, so the leading word `Task` is redundant.  Renaming all usage to `UnobservedTaskExceptionIntegration`.

This rename makes one public change. The `DisableTaskUnobservedTaskExceptionCapture` extension method is now obsolete in favor of `DisableUnobservedTaskExceptionCapture`.

### DIFF
--- a/.github/workflows/device-tests.yml
+++ b/.github/workflows/device-tests.yml
@@ -47,7 +47,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        api-level: [27, 28, 29, 30, 31, 32, 33]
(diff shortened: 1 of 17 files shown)


bitsandfoxes (approved):
That's a nice little cleanup!

PR #1952, opened 2022-09-27
WIP: Add .NET 7 to test projects

mattjohnsonpint:
Per #1676, start testing on .NET 7 RC1.

No changes to targets of any Sentry libraries yet.

#skip-changelog

PR #2035, opened 2022-11-09
Fix trace propagation targets setter

mattjohnsonpint:
While testing with .NET 7 (in #1952), I found that configuration bindings for lists have changed.  It used to be that an existing list would have its `Add` method called to incorporate items during binding.  Now it appears the whole list is replaced.

With `SentryOptions.TracePropagationTargets`, we use `[".*"]` by default.  If the user provides a different list via config, or adds to the existing list in code, we have to clear the wildcard automatically.

With these changes, we now do this correctly for both adding to this _or replacing_ the list - so it works correctly on all targets and we don't have to make any assumptions about configuration binding behavior per version.

### DIFF
--- a/src/Sentry/SentryOptions.cs
+++ b/src/Sentry/SentryOptions.cs
@@ -579,6 +579,10 @@ namespace Sentry
         /// </remarks>
         public Func<TransactionSamplingContext, double?>? TracesSampler { get; set; }
 
+        // The default propagation list will match anything, but adding to the list should clear that.
+        private IList<TracePropagationTarget> _tracePropagationTargets = new AutoClearingList<TracePropagationTarget>
+            (new[] {new TracePropagationTarget(".*")}, clearOnNextAdd: true);
+
         /// <summary>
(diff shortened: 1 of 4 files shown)


src/Sentry/SentryOptions.cs
```diff
@@ -590,11 +594,35 @@ public double TracesSampleRate
         /// <remarks>
         /// Adding an item to the default list will clear the <c>.*</c> value automatically.
         /// </remarks>
-        public IList<TracePropagationTarget> TracePropagationTargets { get; set; } =
-            new AutoClearingList<TracePropagationTarget>
+        public IList<TracePropagationTarget> TracePropagationTargets
+        {
+            // NOTE: During configuration binding, .NET 6 and lower used to just call Add on the existing item.
+            //       .NET 7 changed this to call the setter with an array that already starts with the old value.
+            //       We have to handle both cases.
```
bitsandfoxes, line 601: That sounds like a *huge* gotcha.

PR #2040, opened 2022-11-09
Fix CI build detection for mobile tests

mattjohnsonpint:
Detecting CI for mobile tests needs to be done at build time.  At runtime, we're executing in an emulator, so the environment variables used by `BuildServerDetector` are not available.

#skip-changelog

### DIFF
--- a/test/Sentry.Tests/UnobservedTaskExceptionIntegrationTests.cs
+++ b/test/Sentry.Tests/UnobservedTaskExceptionIntegrationTests.cs
@@ -33,7 +33,7 @@ public class UnobservedTaskExceptionIntegrationTests
     public void Handle_UnobservedTaskException_CaptureEvent()
     {
 #if __MOBILE__ && CI_BUILD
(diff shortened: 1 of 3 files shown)


test/Sentry.Tests/UnobservedTaskExceptionIntegrationTests.cs
```diff
@@ -36,8 +32,8 @@ public void Handle_WithException_CaptureEvent()
     [SkippableFact]
     public void Handle_UnobservedTaskException_CaptureEvent()
     {
-#if __MOBILE__
-        Skip.If(BuildServerDetector.Detected, "Test is flaky on mobile in CI.");
+#if __MOBILE__ && CI_BUILD
+        Skip.If(true, "Test is flaky on mobile in CI.");
```
bitsandfoxes, line 36:
```suggestion
        throw new Xunit.SkipException("Test is flaky on mobile in CI.");
```
What about this?

bitsandfoxes (approved):

PR #2050, opened 2022-11-15
feat: debug images for Portable PDB

### LINKED ISSUE #1740: Line number for .NET stack traces with server-side PDB support
### Problem statement:

.NET since its inception had the default Debug and Release configurations emit debug information that was used in production, so that at runtime, exceptions could include stack traces with function names, line numbers and file paths.

That requires apps to ship `pdb`s together with the executable, which is common on server apps, and most desktop apps. Particularly internal ones where the download size and reverse engineering are not big concerns.
The only motivation for server-side `pdb` support in those days were from desktop apps (such as WinForms and WPF) that wanted to have a smaller installer, or were concerned with facilitating reverse engineering of their IP.

In 2011 mobile support for .NET was introduced through Xamarin, in mobile app size is critical to stay as small as possible, so Xamarin didn't include debug information in the final app. And to this day, Xamarin users don't have line numbers on stack traces in Sentry, or any other competitor (to the best of my knowledge).

This feature was requested back in [2015](https://github.com/getsentry/sentry-dotnet/issues/193) but the reason this became a bigger priority now is due to [.NET MAUI](https://github.com/getsentry/sentry-dotnet/issues/1174). And also because ASP.NET Core since .NET 3.1 (or 5?) does not include PDBs anymore in the runtime installation. So 'system' frames no longer have line numbers on stack traces. Symbols (portable `pdb`s) are made available on the nuget.org's symbol server.

####  Support portable PDB (ppdb) with the focus of giving line numbers for InApp frames
1. For that, SDK needs to include data required by the server to look up data in portable PDBs
2. `sentry-cli` will need to understand and upload ppdb's
3. Sentry's .NET SDK can include msbuild tasks that [use the wizard API](https://github.com/getsentry/sentry-unity/pull/780/) for onboarding, and invoke `sentry-cli`

#### Line number of other frames such as system frames. 
This requires fetching symbols from nuget.org

#### [Rely on source link](https://github.com/getsentry/sentry-dotnet/issues/400) to render a link to the right sha/file:line

1. At this point we could fetch the source and [show it as Source Context](https://github.com/getsentry/sentry-dotnet/issues/646).

### Technical details

.NET tooling, by default, generates `exe` or `dll` with an accompanying `pdb`.
At runtime, when you use an API to retrieve a _line number_ or a _file path_, the framework makes a lookup to the `pdb` on the local directory, and find the appropriate line number and path through the pdb.

https://github.com/getsentry/sentry-dotnet/blob/610dee87287bfcb0138090f040511a154f6aa9db/src/Sentry/Internal/MonoSentryStackTraceFactory.cs#L58-L60

https://github.com/getsentry/sentry-dotnet/blob/610dee87287bfcb0138090f040511a154f6aa9db/src/Sentry/Extensibility/SentryStackTraceFactory.cs#L178-L188

```
dotnet Sentry.Samples.Console.Basic.dll
Unhandled exception. System.NullReferenceException: Object reference not set to an instance of an object.
   at Program.<<Main>$>g__AnotherMethod|0_1() in /Users/bruno/git/sentry-dotnet/samples/Sentry.Samples.Console.Basic/Program.cs:line 14
   at Program.<<Main>$>g__SomeMethod|0_0() in /Users/bruno/git/sentry-dotnet/samples/Sentry.Samples.Console.Basic/Program.cs:line 9
   at Program.<Main>$(String[] args) in /Users/bruno/git/sentry-dotnet/samples/Sentry.Samples.Console.Basic/Program.cs:line 4
```

Outside of this, there's no (to my knowledge) straightforward way to get from an `Exception` instance to an over-the-wire format that a standard system can be plugged in to give line numbers and paths. In other words, `Exception.ToString()` will just result in the same stack trace, but without any line numbers.

```
rm -rf Sentry.Samples.Console.Basic.pdb 
dotnet Sentry.Samples.Console.Basic.dll
Unhandled exception. System.NullReferenceException: Object reference not set to an instance of an object.
   at Program.<<Main>$>g__AnotherMethod|0_1()
   at Program.<<Main>$>g__SomeMethod|0_0()
   at Program.<Main>$(String[] args)
```

The way [Mono](https://github.com/mono/mono) solved this in the past was through [`mono-symbolicate`](https://github.com/mono/mono/tree/main/mcs/tools/mono-symbolicate).
The runtime generates an exception in a different format than above. It encodes the `mvid` (module version id which is the id of the pdb) and the `aotid` which was added when the code as AOT (ahead of time compiled, for example on _Xamarin.iOS_). _Mono symbolicate_ on the 'server' needs to prepare code in a way that is sort of a symbol server lookup protocol, by moves the pdbs to folders named by their `mvid`.

[This PR added support to parse](https://github.com/getsentry/sentry-dotnet/pull/578) that in the Sentry SDK for .NET.

In order to get this done at Sentry, we'll need to change the Sentry SDK for .NET and add the data required by the backend to find the correct `pdb`, and lookup within it. 

The data that has to be read at runtime, to add to the payload can be found in this PoC: https://github.com/bruno-garcia/simbolo/blob/d7340e46c45c87e4598b76f350336cb695fa00f1/Simbolo/Client.cs#L19
And an example of the lookup done this way (with .NET library): https://github.com/bruno-garcia/simbolo/blob/d7340e46c45c87e4598b76f350336cb695fa00f1/Simbolo.Backend/Symbolicate.cs#L14
The PoC above does call from native code, since it was supposed to serve as an example where we use FFI from Rust into .NET on the backend: https://github.com/bruno-garcia/simbolo/blob/d7340e46c45c87e4598b76f350336cb695fa00f1/Simbolo.NativeLib/app.c

The PoC relied on `[Mono.Cecil](https://github.com/jbevain/cecil)`. This library is very popular, widely adopted and allows for modifying the IL. On the other hands it allocates memory when reading debug info. Alternatively, [`System.Reflection.Metadata`](https://github.com/dotnet/runtime/tree/main/src/libraries/System.Reflection.Metadata) can only read things, but does so in a much more optimized way. [`Simbolo` has a branch using `System.Reflection.Metadata`](https://github.com/bruno-garcia/simbolo/tree/ref/use-System.Reflection.Metadata).

The symbol lookup straight to portable `pdb` will be a solution to customers that upload their . NET `pdbs` to Sentry.
But this won't solve the use case for libraries published to NuGet. There, we'll need to make a symbol server lookup to nuget.org to fetch the relevant debug symbols. All .NET libraries (installed with .NET) have source link information so we know the commit sha, git repo link so we can render a link to the exact line number.

### Resources

Relates to:
* https://github.com/getsentry/sentry-dotnet/issues/193
* https://github.com/getsentry/sentry-dotnet/issues/604

PPDB 
* docs: https://docs.microsoft.com/en-us/windows-hardware/drivers/debugger/symbols-portable-pdb
* design: https://github.com/dotnet/core/blob/main/Documentation/diagnostics/portable_pdb.md
* spec: https://github.com/dotnet/runtime/blob/main/docs/design/specs/PortablePdb-Metadata.md

For Sentry employees: Internal video intro about this problem: https://drive.google.com/file/d/1nXOtB-ChTcuCj1fqgPrbHhx83A0bvLEw/view

vaind:
This adds the new event properties defined in [RFC 0013](https://github.com/getsentry/rfcs/pull/13), namely InstructionAddress / FunctionId + Relative Indexing, and a list of `pe_dotnet` DebugImages.

- replaces the original PR #1785 with a different approach to avoid adding new interface methods to `ISentryStackTraceFactory`
- closes #1740

## TODOs & followups
* [x] make sure we know how this affects existing issues grouping - seems to be creating new issues **if users upload symbols** meaning all ignored issues would be resurrected too.
* [x] add changelog, depending on the previous item
* [ ] docs, if needed
* [x] Android doesn't contain assemblies as DLLs
* [ ] ~generate PDBs for Sentry libraries (see the asp.net (or Android) event below)~ (see https://github.com/getsentry/sentry-dotnet/issues/2073)
* [x] unit test `MergeDebugImagesInto()`
* [x] unit test for stack traces to contain (correct) debug images
* [ ] maybe device tests that debug images are actually sent? Locally, this seems covered because we're testing it through the verifier.

## Event examples
Notice the presence of the "Images loaded" section and the line numbers in (full) stack traces. All apps were built with the following commands before executing:

```shell-script
cd samples/<sample-name>
git clean -ffxd . # after a branch switch
dotnet build --configuration Release 
sentry-cli upload-dif --project=sentry-dotnet --org=sentry-sdks --wait bin/Release/
rm -vf ./**/*.pdb # Remove-Item ./* -Recurse -Include "*.pdb"
```

* Basic CLI 
   - [main branch](https://sentry.io/organizations/sentry-sdks/issues/3664608251/events/afabbadeb9c84351980e60c00392c80c/?project=5428537)
   - [this PR](https://sentry.io/organizations/sentry-sdks/issues/3323330137/events/2771dd4647584990955a430660b77a1b/?project=5428537)
* Complex CLI - longer stack trace (click "Full stack trace")
   - [main branch](https://sentry.io/organizations/sentry-sdks/issues/3744064963/events/edda14f2171547f1bc2e65c418b39f63/?project=5428537)
   - [this PR](https://sentry.io/organizations/sentry-sdks/issues/3744054261/events/c90d291ce14a4d718968b24b85c96c36/?project=5428537) - this also successfully fetched `System.Private.CoreLib.dll` from a MS symbol server
* Android MAUI
   - [main branch](https://sentry.io/organizations/sentry-sdks/issues/3755224873/events/1dd7271fcf034b609f232de76dc78707/?project=5428537)
   - [this PR](https://sentry.io/organizations/sentry-sdks/issues/3764994593/events/83d884bf05a34c0eaa7745cc01149b53/?project=5428537)
* ASP.NET core MVC - multiple exceptions in a single event (also click "Full stack trace")
   - [main branch](https://sentry.io/organizations/sentry-sdks/issues/3744089646/events/c481e05e11ea453693658c55242b134d/?project=5428537)
   - [this PR](https://sentry.io/organizations/sentry-sdks/issues/3744250864/events/9cbaab1c74be425db3e636d26da9bc0a/?project=5428537)
* MacCatalyst
   - [main branch](https://sentry.io/organizations/sentry-sdks/issues/3749181901/events/b8cf3777d1ed46d0abd421b4e749b98e/?project=5428537)
   - [this PR](https://sentry.io/organizations/sentry-sdks/issues/3749129049/events/90ffb07a76614fbea2be05fb2bf78b7e/?project=5428537)
* iOS
   - [main branch](https://sentry.io/organizations/sentry-sdks/issues/3749339831/events/381af6b032f542f0a33a23de627f9aa6/?project=5428537)
   - [this PR](https://sentry.io/organizations/sentry-sdks/issues/3672147213/events/2536519d7ed34da690f0d3cef71717ff/?project=5428537)
   - [this PR, MAUI](https://sentry.io/organizations/sentry-sdks/issues/3496490412/events/a7b7c158382842f886197211244b4a3b/?project=5428537)


Swatinem:
Thanks @vaind for picking this up ❤️ 

Serverside everything should be ready, you should be able to upload portable pdb files via sentry-cli, and serverside symbolication should pick up the new every properties. Feel free to reach out to me anytime if you need help.

scripts/accept-verifier-changes.ps1
```diff
@@ -0,0 +1,23 @@
+# Similar to `dotnet verify accept` but doesn't create new runtime-specific ".verified" files if a common one exists.
```
bruno-garcia, line 1: Not clear why do we need this?
Should we have an entry on CONTRIBUTING.md describing when/why to use it?

scripts/update-test-apks.ps1
```diff
@@ -0,0 +1,51 @@
...
+Set-StrictMode -Version Latest
+$ErrorActionPreference = "Stop"
+
+$repoRoot = "$PSScriptRoot/.."
+$apkPrefix = "$repoRoot/test/Sentry.Tests/Internals/android"
+
+function BuildAndroidSample([bool] $UseAssemblyStore, [bool] $UseAssemblyCompression)
+{
+    $sampleDir = "$repoRoot/samples/Sentry.Samples.Android"
+    $framework = 'net6.0-android'
```
bruno-garcia, line 12: Does this value need to align with what's inside `Sentry.Samples.Android.csproj`?
I'm thinking of maintaining this, will we need to bump this line when we bump other files/stay in sync?

src/Sentry/Extensibility/SentryStackTraceFactory.cs
```diff
@@ -36,287 +21,18 @@ public class SentryStackTraceFactory : ISentryStackTraceFactory
...
 
-        return Create(CreateStackTrace(exception), isCurrentStackTrace);
-    }
-
-    /// <summary>
-    /// Creates a s<see cref="StackTrace"/> from the <see cref="Exception"/>.
-    /// </summary>
-    /// <param name="exception">The exception.</param>
-    /// <returns>A StackTrace.</returns>
-    protected virtual StackTrace CreateStackTrace(Exception? exception) =>
```
bruno-garcia, line 57: This is a breaking change. Can we get around it? Otherwise we'll need to land this in 4.0.

IIRC this was added to get this package out there: https://www.nuget.org/packages/Tuhu.Sentry.AsyncStackTrace

This is not really needed since Sentry 3.0.0 since we bundled Ben.Demystifier's behaviour already.
So we could also decide to go ahead with the change still on 3.0.0

src/Sentry/Extensibility/SentryStackTraceFactory.cs
```diff
@@ -36,287 +21,18 @@ public class SentryStackTraceFactory : ISentryStackTraceFactory
...
-        //   System.Threading.Tasks.Task`1 in InnerInvoke`
-        //   or System.Collections.Generic.List`1 in get_Item
-        var match = RegexAsyncReturn.Match(frame.Module);
-        if (match.Success && match.Groups.Count == 2)
-        {
-            frame.Module = match.Groups[1].Value;
-        }
+        var stackTrace = exception is null ? new StackTrace(true) : new StackTrace(exception, true);
+        var result = DebugStackTrace.Create(_options, stackTrace, isCurrentStackTrace);
+        _options.LogDebug($"Created {typeof(DebugStackTrace).Name} with {0} frames.", result.Frames.Count);
```
bruno-garcia, line 35: The goal was to avoid allocations if Debug is turned off (the default behaviour).

That's why originally all calls were to a nullable reference (`_options.Logger?.Log(..)`) so if `Debug=false`, not even a call was done.
Now there's an extension method so the call is going to be made, we can avoid the interpolation at least

```suggestion
        _options.LogDebug("Created {0} with {1} frames.", typeof(DebugStackTrace).Name, result.Frames.Count);
```

src/Sentry/Sentry.csproj
```diff
@@ -67,4 +68,8 @@
     <None Include="buildTransitive\Sentry.props" Pack="true" PackagePath="build\Sentry.props" />
   </ItemGroup>
 
+  <ItemGroup Condition="$(TargetFramework.StartsWith('net6.0'))">
+    <!-- TODO we only need LZ4 decompression only for Android, but we want to run all tests during development on net6.0/net7.0 directly on dev machine -->
+    <PackageReference Include="K4os.Compression.LZ4" Version="1.*" />
```
bruno-garcia, line 73: This is a show stopper though. Can we add this dep only on the test project instead?

src/Sentry/SentryStackTrace.cs
```diff
@@ -27,6 +27,9 @@ public IList<SentryStackFrame> Frames
         set => InternalFrames = value;
     }
 
+    // This class is sealed externally, but can be extended internally.
+    internal SentryStackTrace() { }
```
bruno-garcia, line 31: I'd just remove the `sealed` and let it be:

```suggestion
```

test/Sentry.Tests/Internals/AndroidAssemblyReaderTests.cs
```diff
@@ -0,0 +1,97 @@
...
+        _logger = new(output);
+    }
+
+    private IAndroidAssemblyReader GetSut(bool isAssemblyStore, bool isCompressed)
+    {
+#if ANDROID
+        // On Android, this tests the current app APK.
+        var apkPath = Environment.CommandLine;
+        // AndroidBuild not available here - We can't use the usual code
+        // var supportedAbis = AndroidBuild.SupportedAbis ?? new List<string> { AndroidBuild.CpuAbi ?? "" };
```
bruno-garcia, line 24: How about an internal helper method on `Sentry` TFM `net*-android` ?

bruno-garcia (commented):
First pass. I'll come back to review those AssemblyReaders and other bits

bruno-garcia:
Forgot to thank you for doing this. This is a huge step for Sentry's .NET support. 🙏

src/Sentry/Internal/AndroidAssemblyReader.cs
```diff
@@ -0,0 +1,710 @@
...
+                    EntriesByHash32.Add(entry.Hash32, entry);
+                }
+
+                if (entry.Hash64 != 0)
+                {
+                    EntriesByHash64.Add(entry.Hash64, entry);
+                }
+            }
+        }
+        // public readonly Dictionary<string, AssemblyStoreManifestEntry?> EntriesByName = new();
```
Swatinem, line 430: I believe you can remove this commented out code

Swatinem (approved):
I haven’t reviewed the AssemblyReader in too much detail.
lgtm, though I would rather defer proper review to @bruno-garcia

scripts/accept-verifier-changes.ps1
vaind: added an entry to contrib docs - it's useful when you have many (intended) changes in `Verify` snapshots

scripts/update-test-apks.ps1
vaind: In the end, we may not need this because there's only one target framework, I've removed it from the script.

src/Sentry/Sentry.csproj
vaind: I know, that's why there's the TODO above - this needs to be part of the Android build for runtime & also part of the test project - I don't know how to do that though... Open to suggestions how to make it work. When I tried adding to the test project, the Sentry project didn't compile because of the dependency. Likely, the whole AndroidAssemblyReader will need to move out of that main project. It's a bit unclear to me due to my limited knowledge of the solution structure and what integrates with what.

src/Sentry/Internal/DebugStackTrace.cs
```diff
@@ -0,0 +1,494 @@
...
+        }
+    }
+
+#if ANDROID
+    private IAndroidAssemblyReader? _assemblyReader;
+    private bool _assemblyReaderInitialized = false;
+
+    private void InitializeAssemblyReader()
+    {
+        if (_assemblyReaderInitialized)
```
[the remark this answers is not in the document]
vaind: you're right, added a lock

src/Sentry/Internal/DebugStackTrace.cs
```diff
@@ -0,0 +1,494 @@
...
+    protected readonly List<DebugImage> _debugImages = new();
+    private readonly Dictionary<Guid, int> _debugImageIndexByModule = new();
+    private const int DebugImageMissing = -1;
+    private bool _debugImagesMerged = false;
+
+    /*
+     *  NOTE: While we could improve these regexes, doing so might break exception grouping on the backend.
+     *        Specifically, RegexAsyncFunctionName would be better as:  @"^(.*)\+<(\w*|<\w*>b__\d*)>d(?:__\d*)?$"
+     *        But we cannot make this change without consequences of ignored events coming back to life in Sentry.
+     */
```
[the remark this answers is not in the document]
vaind: They're the default in Unity though

BTW about 50 % of this class is just moved from SentryStrackTraceFactory

src/Sentry/Internal/DebugStackTrace.cs
```diff
@@ -0,0 +1,494 @@
...
+            // Matching PDB ID is stored in the #Pdb stream of the .pdb file.
+            // See https://github.com/dotnet/runtime/blob/main/docs/design/specs/PE-COFF.md#codeview-debug-directory-entry-type-2
+            debugId = string.Format("{0}-{1:x}", codeView.Guid, entry.Stamp);
+            debugFile = codeView.Path;
+        }
+
+        // well, we are out of luck :-(
+        if (debugId == null)
+        {
+            _options.LogDebug("Skipping DebugImage for module '{0}' because DebugId couldn't be determined", module.Name);
```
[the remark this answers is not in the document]
vaind: * [x] Changed to warning

I don't remember seeing it myself - maybe @Swatinem knows?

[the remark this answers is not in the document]
Swatinem: This should be extremely rare. Usually all the images should have a (deterministic / portable pdb) debugid. the checksum however is missing from a lot of them. So an image without a debug id is not really usable, also not on the backend.

src/Sentry/Sentry.csproj
mattjohnsonpint: You can move the package reference  to `/src/Sentry/Platforms/Android/Sentry.Android.props`.  It is already set up to be included only for the Android target.

`AndroidAssemblyReader.cs` should probably move under `/src/Sentry/Platforms/Android` also.  Everything in that folder is only included for Android targets.

mattjohnsonpint: To be clear though - we are taking a true dependency on the LZ4 package at runtime for Android?  Is there another way?

mattjohnsonpint: To clarify, I mean do we need the LZ4 decompression to actually occur in the user's shipping application? Or just at build time?  If the latter, then the package reference could go in `/src/Sentry/buildTransitive/Sentry.props`, with something like:

```xml
<ItemGroup Condition="$([MSBuild]::GetTargetPlatformIdentifier('$(TargetFramework)')) == 'android'">
    <PackageReference Include="K4os.Compression.LZ4" Version="1.*" PrivateAssets="All" />
</ItemGroup>
```

src/Sentry/Internal/DebugStackTrace.cs
```diff
@@ -0,0 +1,508 @@
...
+        }
+    }
+#endif
+
+    private PEReader? TryReadAssembly(string assemblyName)
+    {
+#if ANDROID
+        InitializeAssemblyReader();
+        return _assemblyReader?.TryReadAssembly(assemblyName);
+#else
```
vaind, line 423: This doesn't work in the Unity SDK when building for Android - we always get the non-android variant. I assume it's because we pre-compile a single DLL before including in the game. Looks like we would need runtime checks instead? That also plays into the approach to take about including AndroidAssemblyReader and the LZ4 dependency.

cc @bitsandfoxes since you haven't been active in this PR yet

bruno-garcia: The Android support done in .NET doesn't really apply to Unity. They have their own Android support stuff, so we only use `netstandard2.0` target (and DLL) from this code base.

test/Sentry.Tests/ApiApprovalTests.Run.Core3_1.verified.txt
```diff
@@ -1206,10 +1207,6 @@ namespace Sentry.Extensibility
     {
         public SentryStackTraceFactory(Sentry.SentryOptions options) { }
         public virtual Sentry.SentryStackTrace? Create(System.Exception? exception = null) { }
-        protected virtual Sentry.SentryStackFrame CreateFrame(System.Diagnostics.StackFrame stackFrame, bool isCurrentStackTrace) { }
-        protected virtual System.Diagnostics.StackTrace CreateStackTrace(System.Exception? exception) { }
-        protected virtual System.Reflection.MethodBase? GetMethod(System.Diagnostics.StackFrame stackFrame) { }
-        protected Sentry.SentryStackFrame InternalCreateFrame(System.Diagnostics.StackFrame stackFrame, bool demangle) { }
```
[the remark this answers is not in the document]
bruno-garcia: After some thought, I suggest we go ahead with this on 3.x
@mattjohnsonpint OK with that?

src/Sentry/Sentry.csproj
bruno-garcia: If this goes into the `netX.0-android`-only target I think it's reasonable. We could go ahead

src/Sentry/Internal/DebugStackTrace.cs
vaind: In this case it does apply (or rather it would if it worked), because this would be executed for Mono builds where it's would help get line numbers. See https://github.com/getsentry/sentry-unity/pull/1106

vaind:
> Lets move that new dependency to the Android TFM only and go ahead?

That's probably blocked by this unresolved discussion https://github.com/getsentry/sentry-dotnet/pull/2050#pullrequestreview-1208105473

Additionally, It would be nice to preserve the tests that run locally on dev machine (with a locally build APK) that are currently present. Those wouldn't work with the dependency being limited to running it as part of the android device tests. It was very useful also during development so I guess it would be again when there are changes that need to be done.


Honestly, I don't really know how to get around this. Maybe you can suggest how we could do it @bruno-garcia or @mattjohnsonpint?

mattjohnsonpint:
Thanks for all of this!  I'll start reviewing soon and get to your questions thereafter.  Thanks.

src/Sentry/Sentry.csproj
mattjohnsonpint: Ok.  I will see if I can figure out if there's a way to access them in memory.  If we need to stay with the LZ4, then the place to add it is in `/src/Sentry/Platforms/Android/Sentry.Android.props`.  Just make a simple `<PackageReference>` with no extra tags, and do use an exact version rather than floating (we use floating versions in tests and samples only as a convenience to reduce update frequency, but they pose a security risk for distributable components.).

vaind:
Hey @mattjohnsonpint how is it going with the review?

src/Sentry/Sentry.csproj
mattjohnsonpint: I've resolved this by moving the LZ4 library and the `AndroidAssemblyReader` to compile for android only.  Separately the files are manually linked into a new `Sentry.Android.Tests` project that compile for regular .NET.

src/Sentry/Internal/DebugStackTrace.cs
mattjohnsonpint: Anything we have in sentry-unity for Android doesn't use the Android support in sentry-dotnet.  Sentry-Unity only uses the `netstandard2.0` target from sentry-dotnet.  If we want to consolidate platform support, we'll have to look at that separately.

mattjohnsonpint (approved):

mattjohnsonpint:
Almost done here. Some failures to investigate. Overall looking pretty good. 🙂

src/Sentry/DebugImage.cs
```diff
@@ -47,6 +52,8 @@ public sealed class DebugImage : IJsonSerializable
     /// </summary>
     public string? CodeFile { get; set; }
 
+    internal Guid? ModuleVersionId { get; set; }
```
mattjohnsonpint, line 55: Any reason this needs to be internal instead of public?

mattjohnsonpint:
There's an issue with Verify getting in the way here.  Addressing that in #2092.

PR #2053, opened 2022-11-16
Update bundled Cocoa SDK to version 7.31.1

mattjohnsonpint:
Updates the Sentry Cocoa SDK (for iOS/MacCatalyst targets) from 7.28.0 to 7.31.1 [[diff]](https://github.com/getsentry/sentry-cocoa/compare/7.28.0...7.31.1) [[changelog]](https://github.com/getsentry/sentry-cocoa/blob/master/CHANGELOG.md).

Also changes how we acquire the SDK.  We now use a submodule, and build it ourselves as part of building `Sentry.Bindings.Cocoa`.  This should make it easier to update going forward.


.github/actions/buildcocoasdk/action.yml
```diff
@@ -0,0 +1,45 @@
...
+      uses: actions/cache@v3
+      with:
+        path: modules/Carthage/.build/debug/carthage
+        key: carthage-${{ env.CARTHAGE_SHA }}
+
+    - name: Build Carthage
+      if: ${{ steps.cache-carthage.outputs.cache-hit != 'true' }}
+      shell: bash
+      run: make all
+      working-directory: modules/Carthage
```
vaind, line 23: Why do we need this? It's available on macOS

vaind (commented):
I haven't had a proper look yet, just wondering whether it would make sense to grab what we do in Unity (and Unreal) instead:
* reusable workflow: https://github.com/getsentry/sentry-unity/blob/f3f805106695832381dbf405449f4babfa2ba3b4/.github/workflows/sdk.yml
* actual call - we would just take one, not the whole matrix - https://github.com/getsentry/sentry-unity/blob/f3f805106695832381dbf405449f4babfa2ba3b4/.github/workflows/ci.yml#L28

.github/actions/buildcocoasdk/action.yml
mattjohnsonpint: It's a custom build so we can create xcframeworks for Mac Catalyst properly.  See comments in the `build-sentry-cocoa.sh` script. https://github.com/getsentry/sentry-dotnet/pull/2053/files#diff-6be5d334c8d46a713f0d9a5b40af751ffc613d6e52405e4a0f72fa06c2d91ed6R6-R9

mattjohnsonpint: (The Carthage submodule is tracking their `xcframework-catalyst` branch.)

mattjohnsonpint: Happy to kill it if you can find any other solution.  I figure there must be a way to skip Carthage and just use `xcodebuild`, but I lack expertise in this area and couldn't figure it out.

I already looked into using either Swift SPM or CocoaPods, but the matching tooling on the Xamarin side just isn't up to par (docs talk about `sharpie pod`, but that is very broken and MS doesn't seem to be in a hurry to fix it).

However it is done, ultimately we need the build output to look like the following to bind correctly in .NET:

```
.
└── Sentry.xcframework
    ├── Info.plist
    ├── ios-arm64
    │   ├── Sentry.framework
    │   └── dSYMs
    ├── ios-arm64_x86_64-maccatalyst
    │   ├── Sentry.framework
    │   └── dSYMs
    └── ios-arm64_x86_64-simulator
        ├── Sentry.framework
        └── dSYMs
```

We get that by using `carthage build --use-xcframeworks --no-skip-current --platform ios,macCatalyst` but only when using the custom build can we ask for `macCatalyst`.

Only reason this matters is because .NET MAUI uses Mac Catalyst to build macOS apps.

mattjohnsonpint:
> wondering whether it would make sense to grab what we do in Unity

I took a lot of inspiration from there, but I believe this will work better for this repo.

We have a few different concerns.  Besides just the custom build of Carthage, we also have msbuild complexities due to multi-targeting that don't arise in the Unity SDK.  Some of that you can see in this PR here: https://github.com/getsentry/sentry-dotnet/pull/2053/files#diff-71d304df430c892a78a35eb87b1dcbd39ece3e782e3900129d28c9bb559672cfR29-R35

.github/actions/buildcocoasdk/action.yml
```diff
@@ -0,0 +1,45 @@
...
+    - name: Get Carthage SHA
+      shell: bash
+      run: echo "CARTHAGE_SHA=$(git rev-parse HEAD)" >> $GITHUB_ENV
+      working-directory: modules/Carthage
+
+    - name: Cache Carthage
+      id: cache-carthage
+      uses: actions/cache@v3
+      with:
+        path: modules/Carthage/.build/debug/carthage
```
bruno-garcia, line 16: can we go without this cache since we're caching below at `Cache Sentry Cocoa SDK`?

bruno-garcia (approved):
LGTM but if Ivan has ideas to simplify both this repo + Unity somehow, that could work too.
Otherwise, this works pretty well for .NET it seems

.github/actions/buildcocoasdk/action.yml
mattjohnsonpint: I split them into separate caches because Carthage takes a very long time to build, and it's not likely we'll need to update it as often as we do the Cocoa SDK.  I also restore it every time to keep it active, otherwise GHA will prune it from the cache over time.

PR #2059, opened 2022-11-17
Cleanup MAUI initialization

mattjohnsonpint:
Cleanup basically dead code.  All the reasons I made `SentryMauiInitializer` have already been resolved.  It now is no better than letting Sentry.Extensions.Logging do the initialization.

#skip-changelog


mattjohnsonpint:
Nevermind, I was wrong.  We still need to do it the current way.

PR #2062, opened 2022-11-23
iOS + MAUI tests

mattjohnsonpint:
This adds iOS tests to CI.  It also fixes several tests that were failing on iOS, and quiets some that were making network calls during the test. 

It also adds an option used only in testing to disable initialiazation of the iOS or Android SDKs when needed.

It also separates verify tests cleanly into partial classes that are excluded in mobile. 

Lots more to come still, but this much is read to merge.

#skip-changelog


test/Sentry.Tests/Protocol/MeasurementTests.verify.cs
```diff
@@ -0,0 +1,29 @@
+#if !__MOBILE__
```
bruno-garcia, line 1: How about excluding *verify.cs* files in the directory build props for mobile projects/targets? Would avoid the ifdef in all files

bruno-garcia (approved):
Two nits but otherwise lgtm 🚀

PR #2067, opened 2022-11-29
Fix missing stack trace on UnobservedTaskException

### LINKED ISSUE #2065: UnobservedTaskExceptionIntegration could provide a stack trace (but doesn't)
### Package

Sentry.AspNetCore

### .NET Flavor

.NET

### .NET Version

6.0.11

### OS

Windows

### SDK Version

3.24.0

_No response_

### Steps to Reproduce

1. Cause an UnobservedTaskException in a Blazor event handler, e.g.:
```
protected void CauseUnobservedTaskException()
	{
		// Task.Run(() => throw new ArgumentException("Test unobserved task exception from Blazor poke page"));
		var task = Task.Factory.StartNew(() =>
		{
			Thread.Sleep(1000); // emulate some calculation
			throw new Exception("Test unobserved task exception from Blazor poke page");
		});

			Thread.Sleep(2000); // emulate some calculation
		task = null;
		GC.Collect();
		GC.WaitForPendingFinalizers();
		GC.Collect();
	}
```
2. Allow Sentry's default handling to report the error

### Expected Result

Issue in Sentry should have a useful stack trace to the cause of the error

### Actual Result

Issue in Sentry doesn't have a stack trace.

If I turn off Sentry's handling, e.g.:
```
webBuilder.UseSentry(sentryOptions =>
{
   sentryOptions.Dsn = ...;
   sentryOptions.DisableUnobservedTaskExceptionCapture();
});
```
... and add my own code like this:
```
TaskScheduler.UnobservedTaskException += (sender, e) =>
{
	// Report the inner exception, to get the stack trace
	SentrySdk.CaptureException(e.Exception.InnerException);
};
```
... then I get a stack trace in Sentry. Could the built-in UnobservedTaskException do this, or something else clever to report the stack trace? Otherwise it's really difficult to debug these issues.

mattjohnsonpint:
Fixes #2065

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
+
+### Fixes
+
+- Fix missing stack trace on UnobservedTaskException ([#2067](https://github.com/getsentry/sentry-dotnet/pull/2067))
+
 ## 3.24.0
 
 ### Features
--- a/src/Sentry/Integrations/UnobservedTaskExceptionIntegration.cs
+++ b/src/Sentry/Integrations/UnobservedTaskExceptionIntegration.cs
@@ -30,7 +30,7 @@ internal class UnobservedTaskExceptionIntegration : ISdkIntegration
 #if NET5_0_OR_GREATER
         var ex = e.Exception;
 #else
-            var ex = e.Exception!;
+        var ex = e.Exception!;
 #endif
         ex.Data[Mechanism.HandledKey] = false;
         ex.Data[Mechanism.MechanismKey] = "UnobservedTaskException";
--- a/src/Sentry/Internal/MainExceptionProcessor.cs
+++ b/src/Sentry/Internal/MainExceptionProcessor.cs
@@ -90,9 +90,12 @@ internal class MainExceptionProcessor : ISentryEventExceptionProcessor
 
             // Exceptions are sent from oldest to newest, so the details belong on the LAST exception.
             var last = exceptions.Last();
-            last.Stacktrace = original.Stacktrace;
             last.Mechanism = original.Mechanism;
             original.Data.TryCopyTo(last.Data);
+
+            // In some cases the stack trace is already positioned on the inner exception.
+            // Only copy it over when it is missing.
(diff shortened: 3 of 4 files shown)


bruno-garcia (approved):

christallire:
Could we get this released ASAP? I have too many UnobservedTaskException without stack traces

mattjohnsonpint:
Yes, there will be a release very soon.  Thanks for your patience.

mattjohnsonpint:
@christallire - 3.24.1 has been release which includes this fix.  Please update and let us know how it goes.  Thanks.

brnbs:
@mattjohnsonpint I am not sure if it's related, but I am using Sentry.AspNetCore 3.29.1 nuget and all the issues I get via UnobservedTaskException mechanism have a very short stack trace:
```
OpenQA.Selenium.WebDriverException: Cannot start the driver service on http://localhost:61414/
  ?, in void DriverService.Start()
  File "Task.cs", line 2381, in void Task.ExecuteWithThreadLocal(ref Task currentTaskSlot, Thread threadPoolThread)
```
```
System.Net.Sockets.SocketException: The I/O operation has been aborted because of either a thread exit or an application request.
  File "Socket.Tasks.cs", line 1451, in void AwaitableSocketAsyncEventArgs.ThrowException(SocketError error, CancellationToken cancellationToken)
  File "ValueTask.cs", line 271, in void ValueTaskSourceAsTask.s_completionAction(object state)
```
Is there any way to fix this?

mattjohnsonpint:
@brnbs - That's not related to this.

You appear to have a legitimate unobserved task exception related to Selenium web driver not being able to start the driver service.  That particular exception is thrown [from here](https://github.com/SeleniumHQ/selenium/blob/43318cb969347d90007f4cbc0827a8de8a8ac937/dotnet/src/webdriver/DriverService.cs#L299-L300).

As to why it's unobserved, I can't exactly say.  At first glance doesn't appear to be in an async code path.  Given the short stack trace, I can only guess that something like `Task.Run(() => SomeDriverService.Start())` is being called somewhere.

At any rate, I don't believe there's any Sentry issue here.  Thanks.

PR #2071, opened 2022-12-03
Attach db connections to child spans correctly

### LINKED ISSUE #2023: Why db.connection is always connected to root transaction and not visible under proper ChildSpan?
### Problem Statement

I looking into your implementation for sentry-dotnet and I had big issue with this particular line.
https://github.com/getsentry/sentry-dotnet/blob/e85b2e9ed17e870f956a9fb7f649e5b9e72c18d2/src/Sentry.DiagnosticSource/Internal/DiagnosticSource/SentryEFCoreListener.cs#L65
Is there any reason why you use always Transaction as root for db.connection rather to connect to Child which was created before and finished after connection? Did it is connected with that, that connection could be opened and in meantime we could finish one span and open another one and then close conection?

### Solution Brainstorm

_No response_

mattjohnsonpint:
Ensures that db connection spans are placed on the last active non-db span, rather than the transaction root.

Fixes #2023 

Replaces #1671 

Some minor refactoring/cleanup included as well.

### DIFF
--- a/src/Sentry/ISpan.cs
+++ b/src/Sentry/ISpan.cs
@@ -76,4 +76,20 @@ public static class SpanExtensions
             SpanTracer tracer => tracer.Transaction,
             _ => throw new ArgumentOutOfRangeException(nameof(span), span, null)
         };
+
+    /// <summary>
+    /// Gets the parent span for database operations. This is the last active non-database span, which might be the
+    /// transaction root, or it might be some other child span of the transaction (such as a web request).
+    /// </summary>
+    /// <remarks>
+    /// Used by EF, EF Core, and SQLClient integrations.
+    /// </remarks>
+    internal static ISpan GetDbParentSpan(this ISpan span)
+    {
+        var transaction = span.GetTransaction();
+        return transaction.Spans
+                   .OrderByDescending(x => x.StartTimestamp)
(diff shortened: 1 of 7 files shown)


src/Sentry/ISpan.cs
```diff
@@ -76,4 +76,20 @@ span switch
...
+    /// transaction root, or it might be some other child span of the transaction (such as a web request).
+    /// </summary>
+    /// <remarks>
+    /// Used by EF, EF Core, and SQLClient integrations.
+    /// </remarks>
+    internal static ISpan GetDbParentSpan(this ISpan span)
+    {
+        var transaction = span.GetTransaction();
+        return transaction.Spans
+                   .OrderByDescending(x => x.StartTimestamp)
```
bruno-garcia, line 91: Is it possible in, global mode, this list can be mutated while we're iterating here?

mattjohnsonpint: Perhaps, but if so then we already have that problem.

https://github.com/getsentry/sentry-dotnet/blob/de3cf7288aec646842f32d59ef88eec9d859fbf1/src/Sentry/TransactionTracer.cs#L298-L300

This one does the same thing, just with the additional filter to make sure we don't grab a connection span in the case of multiple connections on separate threads.

bruno-garcia (approved):

PR #2089, opened 2022-12-16
Improve MAUI event bindings and add tests

mattjohnsonpint:
These tests validate that we bind correctly to MAUI UI events, adding breadcrumbs as appropriate.

Along the way, improved some of the event breadcrumbs and the way that the events are bound.


mattjohnsonpint:
build failures fixed with #2090

bruno-garcia (approved):
🚀

PR #2094, opened 2022-12-21
Fix hash string case

mattjohnsonpint:
In #2050 I refactored some code to consolidate the `ToHexString` internal extension method, and made it return the string in lower-case.

However, the existing `GetHashString` method was using upper case previously.  Switching to lower case created a problem (observed on iOS) where the caching transport failed to create its directory because another directory cased differently already existed.

This PR adds an option for the casing.  The default is lower-case, except for `GetHashString` where the default is upper-case to preserve previous behavior.

#skip-changelog because #2050 where the bug was introduced has not yet been released.

### DIFF
--- a/src/Sentry/Internal/Extensions/HashExtensions.cs
+++ b/src/Sentry/Internal/Extensions/HashExtensions.cs
@@ -2,11 +2,11 @@ namespace Sentry.Internal.Extensions;
 
 internal static class HashExtensions
 {
-    public static string GetHashString(this string str)
+    public static string GetHashString(this string str, bool upperCase = true)
     {
         var bytes = Encoding.UTF8.GetBytes(str);
         using var sha = SHA1.Create();
         var hash = sha.ComputeHash(bytes);
-        return hash.ToHexString();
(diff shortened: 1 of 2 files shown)


mattjohnsonpint:
FYI - I reported the error I got with the case mismatch here:
https://github.com/dotnet/runtime/issues/79889

Appears to affect iOS only.

bruno-garcia (approved):

src/Sentry/Internal/Extensions/HashExtensions.cs
```diff
@@ -2,11 +2,11 @@ namespace Sentry.Internal.Extensions;
 
 internal static class HashExtensions
 {
-    public static string GetHashString(this string str)
+    public static string GetHashString(this string str, bool upperCase = true)
```
bitsandfoxes, line 5: Why is it `true` by default?

mattjohnsonpint: Because that was the previous behavior before refactoring.

Issue #2074, opened 2022-12-05
Bundle `sentry-cli` in `Sentry` NuGet package


bruno-garcia:
[Sentry CLI](https://docs.sentry.io/product/cli/) offers many features.
The main features we use during build phase of projects are:

1. [Uploading debug files](https://docs.sentry.io/product/cli/dif/#uploading-files) including [ProGuard on Android](https://docs.sentry.io/product/cli/dif/#proguard-mapping-upload) and [JS Source Maps](https://docs.sentry.io/product/cli/releases/#sentry-cli-sourcemaps)
2. [Creating source bundles](https://docs.sentry.io/product/cli/dif/#creating-source-bundles)
3. Uploading [commit info](https://docs.sentry.io/product/cli/releases/#sentry-cli-commit-integration), to unlock [suspect commits](https://docs.sentry.io/product/issues/suspect-commits/)
4. [Uploading release details](https://docs.sentry.io/product/cli/releases/), to enable features like [Release Health](https://docs.sentry.io/product/releases/health/), resolving [issue by release](https://docs.sentry.io/product/releases/#track-release-health) and "First seeing in Release X".

While many dev ecosystems are used to pull a binary and run via the CLI. In others, build system automation abstracts these. We integrate with many, such as [Gradle](https://github.com/getsentry/sentry-android-gradle-plugin), [Fastlane](https://github.com/getsentry/sentry-fastlane-plugin) and [Webpack](https://github.com/getsentry/sentry-webpack-plugin).

In the current integrations we have a mock server that simulates Sentry and validates things. This helps us integration test things such as symbol upload.

## Goals

- [x] Bundle the different builds of `sentry-cli` in the final `Sentry` NuGet package - #2107
  * Pin the version in a file that can be bumped, triggering download of the newer version.([don't check-in the binary](https://github.com/getsentry/sentry-fastlane-plugin/issues/165))
- [ ] ~[GitHub workflow to automatically open PR](https://github.com/getsentry/github-workflows) with updates for Sentry CLI~
  - (Moved to #2141)
- [x] On Release build, upload debug files: PPDB and native - #2107
- [ ] ~Upload Release info to Sentry~ 
  - (Moved to #2142)
- [x] Gracefully downgrade in case Sentry is offline: https://github.com/getsentry/team-mobile/issues/3
  - Already handled via `SENTRY_ALLOW_FAILURE` env var
[...]

PR #2107, opened 2023-01-06
Use Sentry CLI after build to upload symbols

mattjohnsonpint:
As part of #2074, this PR adds the following:
- Download and bundle the Sentry CLI with the `Sentry` nuget package
- Call the Sentry CLI to upload debug information files (symbols, etc.) after a successful build.

Currently requires setting the following environment variables manually before running the build.
- `SENTRY_AUTH_TOKEN`
- `SENTRY_ORG`
- `SENTRY_PROJECT`
- `SENTRY_URL` (if self-hosted)

See https://docs.sentry.io/product/cli/configuration/

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -56,4 +56,10 @@
     <Compile Include="$(MSBuildThisFileDirectory)GlobalUsings.cs" />
   </ItemGroup>
 
+  <!-- Set the version and local path for Sentry CLI (downloaded in the restore phase of Sentry.csproj) -->
+  <PropertyGroup Condition="'$(SolutionName)' != 'Sentry.Unity'">
+    <SentryCLIVersion>2.11.0</SentryCLIVersion>
+    <SentryCLIDirectory>$(MSBuildThisFileDirectory)tools\sentry-cli\$(SentryCLIVersion)\</SentryCLIDirectory>
+  </PropertyGroup>
+
 </Project>
--- a/src/Sentry/buildTransitive/Sentry.props
+++ b/src/Sentry/buildTransitive/Sentry.props
@@ -32,4 +32,27 @@
               ItemName="FileWrites" />
     </WriteCodeFragment>
   </Target>
+
+  <Target Name="UploadSymbolsToSentry" AfterTargets="AfterBuild" Condition="'$(PkgSentry)' != '' Or '$(SentryCLIDirectory)' != ''">
+    <PropertyGroup>
+      <!-- Sentry CLI comes from the Sentry Nuget package when installed. -->
+      <SentryCLIDirectory Condition="'$(PkgSentry)' != ''">$(PkgSentry)\tools\</SentryCLIDirectory>
+
+      <!-- We bundle Sentry CLI for these architectures (in Sentry.csproj) -->
+      <_OSArchitecture>$([System.Runtime.InteropServices.RuntimeInformation]::OSArchitecture)</_OSArchitecture>
+      <SentryCLI Condition="$([MSBuild]::IsOSPlatform('OSX')) And $(_OSArchitecture) == 'Arm64'">$(SentryCLIDirectory)osx-arm64\sentry-cli</SentryCLI>
+      <SentryCLI Condition="$([MSBuild]::IsOSPlatform('OSX')) And $(_OSArchitecture) == 'X64'">$(SentryCLIDirectory)osx-x64\sentry-cli</SentryCLI>
+      <SentryCLI Condition="$([MSBuild]::IsOSPlatform('Linux')) And $(_OSArchitecture) == 'Arm64'">$(SentryCLIDirectory)linux-arm64\sentry-cli</SentryCLI>
+      <SentryCLI Condition="$([MSBuild]::IsOSPlatform('Linux')) And $(_OSArchitecture) == 'X64'">$(SentryCLIDirectory)linux-x64\sentry-cli</SentryCLI>
+      <SentryCLI Condition="$([MSBuild]::IsOSPlatform('Windows'))">$(SentryCLIDirectory)windows-x64\sentry-cli.exe</SentryCLI>
+    </PropertyGroup>
+
+    <Message Importance="High" Text="Preparing to upload debug symbols to Sentry for $(MSBuildProjectName) ($(Configuration)/$(TargetFramework))" />
+
+    <Exec
+      Condition="'$(SENTRY_AUTH_TOKEN)' != '' And '$(SENTRY_ORG)' != '' And '$(SENTRY_PROJECT)' != '' And '$(SentryCLI)' != '' And Exists('$(SentryCLI)')"
+      Command="'$(SentryCLI)' upload-dif '$(IntermediateOutputPath)'" />
+
+  </Target>
+
 </Project>
--- a/.gitignore
+++ b/.gitignore
@@ -23,3 +23,4 @@ docs/docfx.zip
 mono_crash.*.json
 test_output/
 test/**/*.apk
+/tools/
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Features
+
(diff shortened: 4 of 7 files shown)


src/Sentry/buildTransitive/Sentry.props
```diff
@@ -32,4 +32,27 @@
               ItemName="FileWrites" />
     </WriteCodeFragment>
   </Target>
+
+  <Target Name="UploadSymbolsToSentry" AfterTargets="AfterBuild" Condition="'$(PkgSentry)' != '' Or '$(SentryCLIDirectory)' != ''">
```
bruno-garcia, line 36: On Gradle we only upload on Release builds by default. This makes the dev loop faster and avoid uploading tons of symbols that won't go to production.
We did have an option to opt-in debug build uploads

Directory.Build.props
```diff
@@ -56,4 +56,10 @@
     <Compile Include="$(MSBuildThisFileDirectory)GlobalUsings.cs" />
   </ItemGroup>
 
+  <!-- Set the version and local path for Sentry CLI (downloaded in the restore phase of Sentry.csproj) -->
+  <PropertyGroup Condition="'$(SolutionName)' != 'Sentry.Unity'">
+    <SentryCLIVersion>2.11.0</SentryCLIVersion>
```
bruno-garcia, line 61: One additional thing we could add here for security reason is to take besides the version, the checksum to verify what we downloaded is what we want.

Doesn't seem like we're doing this on the fastlane plugin anymore: https://github.com/getsentry/sentry-fastlane-plugin/blob/9e694cb721ccb0f4e084a54da305226cabd94de3/script/sentry-cli-download.sh#L15 or grade.

@vaind did the sentry-cli PR bump automation, maybe he knows about this

mattjohnsonpint:
Note, the verification snapshots previously had `.../` in file paths because the `buildTransitive\Sentry.props` file was only imported to `Sentry.Tests.csproj`, so the project path assembly metadata attribute wasn't being added to the other test projects.  Now that it's imported for all projects, it works correctly and strips the paths as intended.

mattjohnsonpint:
Another verification failure was due to the way we were scrubbing attributes on the API verification tests.  On CI, the path string was sometimes long enough that the attribute was printed on two lines and we only scrubbed one of them.  Fixed.

Directory.Build.props
mattjohnsonpint: I saw the checksum verification in the sentry-cli download bash script.  It uses python.  I think I can do something similar with C# or pure MSBuild so we don't have a python dependency.

src/Sentry/buildTransitive/Sentry.props
mattjohnsonpint: Added `SentryUploadSymbols` MSBuild property.  It's true by default for release builds, and false by default otherwise.  End user can set it in their csproj to override if desired.

Directory.Build.props
mattjohnsonpint: Was able to do this easily with the [`VerifyFileHash`](https://learn.microsoft.com/visualstudio/msbuild/verifyfilehash-task) msbuild task.

mattjohnsonpint:
Added the file hash verifications, and also went ahead and put the i686 (32-bit) versions in there too.  Nuget size is 43MB, which is manageable.

mattjohnsonpint:
Going to do some refactoring before completing this, as we'll want to separate the Sentry CLI setup from the symbol upload part, such that we can re-use it for other Sentry CLI tasks like creating releases, etc.

src/Sentry/buildTransitive/Sentry.props
```diff
@@ -32,4 +32,69 @@
...
+
+    <ItemGroup>
+      <SentryCLIInfo Include="Sentry CLI Configuration:" />
+    </ItemGroup>
+    <Exec Command="'$(SentryCLI)' info" Condition="'$(SentryCLI)' != ''" IgnoreExitCode="true" ConsoleToMsBuild="true" StandardOutputImportance="Low">
+      <Output TaskParameter="ExitCode" PropertyName="_SentryCLIStatusCode" />
+      <Output TaskParameter="ConsoleOutput" ItemName="SentryCLIInfo" />
+    </Exec>
+
+    <Message Importance="High" Condition="'$(_SentryCLIStatusCode)' != '0'" Text="The Sentry CLI is not fully configured with authentication, organization, and project." />
```
mattjohnsonpint, line 88: I've been debating whether this should be a message or a warning.  If it were a warning, then right after updating to the version that includes this, people will start getting the warning in release builds.  That would be good from the perspective that they now have signal to either go configure their auth, org, and project, or to disable symbol upload with `<SentryUploadSymbols>false</SentryUploadSymbols>`.  But it would also be bad from the perspective that some people treat warnings as errors and we'd break their build.  Also if they disable symbol upload now, they might never turn it back on.

So for now I think we stick with an informational message.  We can consider switching to warning later, perhaps.

src/Sentry/buildTransitive/Sentry.props
```diff
@@ -32,4 +32,69 @@
...
+    <PropertyGroup Condition="'$(_SentryCLIStatusCode)' != '0'">
+      <SentryCLI />
+    </PropertyGroup>
+  </Target>
+
+  <!-- Upload symbols (and any other debug information files) to Sentry after the build. -->
+  <Target Name="UploadSymbolsToSentry" AfterTargets="AfterBuild" DependsOnTargets="_PrepareSentryCLI" Condition="'$(SentryUploadSymbols)' == 'true' And '$(SentryCLI)' != ''">
+    <Message Importance="High" Text="Preparing to upload debug symbols to Sentry for $(MSBuildProjectName) ($(Configuration)/$(TargetFramework))" />
+    <Exec Command="'$(SentryCLI)' upload-dif $(SentryCLIOptions) '$(IntermediateOutputPath)'" />
+  </Target>
```
mattjohnsonpint, line 98: After this last refactoring, other Sentry CLI usage can be added with small targets like this one.  Everything else is handled ahead of time.

src/Sentry/buildTransitive/Sentry.props
bruno-garcia: Agree with the thought process. And turning into a Warning could go on next major.
But the `!= 0` is the right thing? Do we have an error code specific for missing auth or (different one?) org or project?

I wonder if we'd be giving this message if Sentry CLI was just erroring out for some other reason (can't reach Sentry right now).

bruno-garcia (commented):
Generally seems good to go. Only a couple notes to clarify

src/Sentry/buildTransitive/Sentry.props
mattjohnsonpint: Oh, actually I need to do it slightly different if org and project are configured manually.  (It has a `--no-defaults` option to only check the authentication part.)

But also, yes it appears we need to check the actual output to determine if the Sentry server is unreachable.  The exit code alone is insufficient.  (It returns 1.)
Should the build warn/error in such a case?

src/Sentry/buildTransitive/Sentry.props
```diff
@@ -32,4 +32,69 @@
               ItemName="FileWrites" />
     </WriteCodeFragment>
   </Target>
+
+  <PropertyGroup>
+    <!-- Defaults for Sentry CLI are set here.  Any of these properties can be set in the application to override the defaults. -->
+    <SentryUploadSymbols Condition="'$(SentryUploadSymbols)' == '' And '$(Configuration)' == 'Release'">true</SentryUploadSymbols>
```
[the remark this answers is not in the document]
mattjohnsonpint: Great idea!

src/Sentry/buildTransitive/Sentry.props
mattjohnsonpint: I believe I've addressed this with the latest changes. Thanks.

bruno-garcia (approved):

PR #2113, opened 2023-01-12
WIP: Create release in Sentry after build

mattjohnsonpint:


mattjohnsonpint:
TODO: Setting release at build time should also set it as the default on `SentryOptions.Release` at runtime.

We can autogenerate another assembly metadata attribute for this (perhaps `Sentry.Release`), in the same way that we are currently doing for `Sentry.ProjectDirectory` [here](https://github.com/getsentry/sentry-dotnet/blob/52ef5f2575eef20cc1d9a4f56348b633eb950053/src/Sentry/buildTransitive/Sentry.props#L3-L34)

bitsandfoxes:
Closing this as stale and to be picked up again with https://github.com/getsentry/sentry-dotnet/issues/2142

PR #2130, opened 2023-01-27
add craft target Sentry.Android.AssemblyReader

bruno-garcia:


bruno-garcia:
Oh, I don't think we need this in the registry.

PR #2133, opened 2023-01-29
Set Sentry.Scope.User to be nullable

kanadaj:
Sentry.Scope.User was written to take a `null` value correctly (and even the documentation suggests to do so) but the actual annotation marked it as non-nullable until now.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+- Mark `Sentry.Scope.User` as nullable, as per its implementation and the documentation
+
 ### Features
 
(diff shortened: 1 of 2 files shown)


mattjohnsonpint:
Thanks, but this would be a breaking change, and would require many other breaking changes due to how the interfaces are currently set up.

Also, the property is correctly non-nullable with regard to the `get`.  The `set` will accept a null, so the correct way to annotate this is to add an `[AllowNull]` attribute.  There are several other properties that should have this as well.  I'll make a new PR to address them all.  I also want to get nullability annotations working properly for our unit test projects at the same time.

Thanks.

PR #2139, opened 2023-01-30
Use 4-core Linux for Android device tests

mattjohnsonpint:
When running the Android device tests, lets run the Android emulator on a 4-core Linux ["larger runner"](https://docs.github.com/en/actions/using-github-hosted-runners/using-larger-runners) instead of the macOS runner.

For Linux, this only works on larger runners, because the 2-core default Linux runners are on hardware that does not support nested virtualization.  We were running them previously on macOS, which do support nested virtualization but for some reason are still flaky when trying to use it.

This is in hopes that we have fewer intermittent test failures related to the emulator failing to boot.

It also happens to be [cheaper](https://docs.github.com/en/billing/managing-billing-for-github-actions/about-billing-for-github-actions#per-minute-rates).

#skip-changelog


mattjohnsonpint:
https://github.blog/changelog/2023-02-23-hardware-accelerated-android-virtualization-on-actions-windows-and-linux-larger-hosted-runners/

mattjohnsonpint:
It works!  Had to change the cache key to not use the AVDs created from the macOS runner.  Other than that, was pretty straightforward.

Also tested if hardware GPU acceleration (`-gpu host`) works, but it does not.  We'll stick with `-gpu swiftshader_indirect`.

mattjohnsonpint:
Seems to be working well now.  This is a representative run (uses cached avds from prior run):
https://github.com/getsentry/sentry-dotnet/actions/runs/4308804445

image

May be possible to optimize further, but this is pretty good.  Seems to be stable so far also.

mattjohnsonpint:
Holding temporarily for internal review, but it does all work.

bruno-garcia:
> Seems to be working well now. This is a representative run (uses cached avds from prior run): https://github.com/getsentry/sentry-dotnet/actions/runs/4308804445
> 
> image
> 
> May be possible to optimize further, but this is pretty good. Seems to be stable so far also.

That [compares to](https://github.com/getsentry/sentry-dotnet/actions/runs/4314106188): 
image
?

mattjohnsonpint:
When they _work_, the perf is about the same.  It does vary somewhat between runs though.

The problem is that they often don't work.  If the emulator fails to boot, the test eventually times out.  Then we have to re-run the failed jobs.  The hope is that running on Linux will be more stable.  Either way, they're definitely cheaper.

bruno-garcia (approved):

PR #2155, opened 2023-02-03
Log aborted HTTP requests as debug instead of error

### LINKED ISSUE #2132: BadHttpRequestException when accessing request body is logged as error
### Problem Statement

When a HTTP request is cancelled before the request body is fully sent, Sentry cannot read the body of the request, triggering the following error log: https://github.com/getsentry/sentry-dotnet/blob/main/src/Sentry.AspNetCore/ScopeExtensions.cs#L56

While it's great that this is logged in case we need to debug an issue related to this, that implies the log level should be Debug as well, since it's an expected outcome of any aborted request.

This issue was raised by @mattjohnsonpint as possibly related: https://github.com/dotnet/aspnetcore/issues/23949

However, as per https://github.com/dotnet/aspnetcore/issues/23949#issuecomment-658746290 "The exception needs to be thrown from the request body APIs so you don't think the request finished gracefully, but the caller should catch it."

In this case, Sentry is the caller, so we should handle this gracefully instead of logging an error.

### Solution Brainstorm

Ideally, we should check the `HttpContext.RequestAborted` `CancellationToken` and if it's set, we should only log this as a lower log level. Anywhere from Debug to Warning might be appropriate here I feel.

mattjohnsonpint:
Fixes #2132

Reorganized and cleaned up some nearby tests also.


mattjohnsonpint:
@kanadaj FYI

test/Sentry.AspNetCore.Tests/ScopeExtensionsTests.cs
```diff
@@ -0,0 +1,478 @@
...
+            .Returns(new[] { extractor });
+
+        _sut.Populate(_httpContext, SentryAspNetCoreOptions);
+
+        extractor.Received(1).ExtractPayload(Arg.Any<IHttpRequest>());
+    }
+
+    [Theory]
+    [MemberData(nameof(InvalidRequestBodies))]
+    public void Populate_PayloadExtractors_DoesNotConsiderInvalidResponse(object expected)
```
mattjohnsonpint, line 310: Note, this was the only test modified.  The others were just moved to this file and cleaned up a bit.

bruno-garcia (approved):

Issue #2161, opened 2023-02-06
ViewHierarchy for MAUI


mattjohnsonpint:
Investigate if we can provide MAUI-specific view hierarchy.

Example in iOS doc: https://docs.sentry.io/platforms/apple/guides/ios/enriching-events/viewhierarchy/

![image](https://github.com/getsentry/sentry-dotnet/assets/1633368/2483afef-7834-44fc-9b5e-f54be78bf34e)

PR #2163, opened 2023-02-07
feat: View Hierarchy

bitsandfoxes:
Relates to  https://github.com/getsentry/sentry-dotnet/issues/2161
Based on this [RFC](https://github.com/getsentry/rfcs/pull/33/files).

Originally, I tried keeping this within the Unity SDK but the View Hierarchy requires changes to the `AttachmentType`.


src/Sentry/ViewHierarchy.cs
```diff
@@ -0,0 +1,57 @@
+using Sentry.Extensibility;
+using Sentry.Internal.Extensions;
+
+namespace Sentry
+{
+    public class ViewHierarchy : IJsonSerializable
+    {
+        public string RenderingSystem { get; set; } = string.Empty;
+        public List<IJsonSerializable>? Children { get; set; }
```
mattjohnsonpint, line 9:
```suggestion
        public List<ViewHierarchyNode> Children { get; set; }
```

src/Sentry/ViewHierarchy.cs
```diff
@@ -0,0 +1,57 @@
...
+
+            writer.WriteEndObject();
+        }
+    }
+
+    public class ViewHierarchyNode : IJsonSerializable
+    {
+        public string Type { get; set; } = string.Empty;
+        public string Identifier { get; set; } = string.Empty;
+        public List<IJsonSerializable>? Children { get; set; }
```
mattjohnsonpint, line 35:
```suggestion
        public List<ViewHierarchyNode>? Children { get; set; }
```

src/Sentry/ViewHierarchy.cs
```diff
@@ -0,0 +1,57 @@
...
+        public string RenderingSystem { get; set; } = string.Empty;
+        public List<IJsonSerializable>? Children { get; set; }
+
+        public void WriteTo(Utf8JsonWriter writer, IDiagnosticLogger? logger)
+        {
+            writer.WriteStartObject();
+
+            writer.WriteStringIfNotWhiteSpace("rendering_system", RenderingSystem);
+
+            if (Children is {} children)
```
mattjohnsonpint, line 17: I think safe to say there will always be children here?

src/Sentry/ViewHierarchy.cs
```diff
@@ -0,0 +1,57 @@
...
+            {
+                writer.WriteStartArray("windows");
```
mattjohnsonpint, line 19: Should the array be called `Windows` to match this (instead of `Children`)?

src/Sentry/ViewHierarchy.cs
```diff
@@ -0,0 +1,57 @@
...
+                    child.WriteTo(writer, logger);
+                }
+                writer.WriteEndArray();
+            }
+
+            writer.WriteEndObject();
+        }
+    }
+
+    public class ViewHierarchyNode : IJsonSerializable
```
mattjohnsonpint, line 31: Wondering if `ViewHierarchytNode` is useful on its own, or if it will always need to be overridden with a more specialized class?  If the latter, then perhaps it should be `abstract`?

src/Sentry/ViewHierarchy.cs
```diff
@@ -0,0 +1,57 @@
...
+    {
+        public string Type { get; set; } = string.Empty;
+        public string Identifier { get; set; } = string.Empty;
+        public List<IJsonSerializable>? Children { get; set; }
+
+        public virtual void WriteTo(Utf8JsonWriter writer, IDiagnosticLogger? logger)
```
mattjohnsonpint, line 37: Perhaps a better way we could do this would be to not make _this_ method virtual, but instead create a new method that is just for serializing the extra properties.  Something like `protected virtual WritePropertiesTo ...`.  Then the inheriting class wouldn't have to worry about the base properties or children.

mattjohnsonpint: (Or `abstract`, if we decide the class should be abstract)

bruno-garcia (approved):
LGTM

PR #2176, opened 2023-02-11
Workaround issue with Verify on Windows in Rider on .NET 7

mattjohnsonpint:
Two Verify tests are failing, only on Windows, only on Rider, and only for .NET 7.

image

```
Received: SentryStackTraceFactoryTests.MethodGeneric_mode=Enhanced.DotNet7_0.received.txt
Verified: SentryStackTraceFactoryTests.MethodGeneric_mode=Enhanced.verified.txt
Compare Result:
  {
    FileName: Internals/SentryStackTraceFactoryTests.cs,
    Function: void SentryStackTraceFactoryTests.GenericMethodThatThrows<T>(T value),
-   AbsolutePath: {ProjectDirectory}Internals/SentryStackTraceFactoryTests.cs,
+   AbsolutePath: {UserProfile}/Code/getsentry/sentry-dotnet/test/Sentry.Tests/Internals/SentryStackTraceFactoryTests.cs,
    InApp: true,
    AddressMode: rel:0
  }
```

On closer inspection, it appears that the `ProjectDir` and `SolutionDir` MSBuild properties are being passed with a lower-case drive letter `c:\` instead of `C:\`.  These get written to assembly metadata attributes, and then fails the replacement that Verify does when scrubbing.

Will add an issue to Verify for @SimonCropp to investigate, but for now I'm working around it by forcing the drive letter to be upper-cased.

@SeanFeldman - This will fix the test fail you had while onboarding.

#skip-changelog

### DIFF
--- a/test/Directory.Build.props
+++ b/test/Directory.Build.props
@@ -9,6 +9,17 @@
     <IsTestProject>true</IsTestProject>
   </PropertyGroup>
 
+  <!--
+    Workaround for Verify issue with scrubbing when running in Rider on Windows.
+    Ensures that the volume label is upper cased ("C:" instead of "c:"), which is read
+    from ProjectDir and SolutionDir properties during compilation and written to the
+    assembly metadata attributes.
+  -->
+  <PropertyGroup Condition="$([MSBuild]::IsOSPlatform('Windows'))">
+    <ProjectDir Condition="'$(ProjectDir)' != ''">$(ProjectDir.Substring(0, 1).ToUpper())$(ProjectDir.Substring(1))</ProjectDir>
+    <SolutionDir Condition="'$(SolutionDir)' != ''">$(SolutionDir.Substring(0, 1).ToUpper())$(SolutionDir.Substring(1))</SolutionDir>
+  </PropertyGroup>
+
   <ItemGroup>
     <Using Include="Sentry" />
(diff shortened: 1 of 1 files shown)


bruno-garcia (approved):

SeanFeldman (approved):
I was supposed to reduce the load on Matt, not add it 😂

SimonCropp:
sorry for the hassle peoples. it is weird though, the tests target net7 and i run rider and have not seen this. can u provide a minimal repro?

mattjohnsonpint:
@SimonCropp - If I can make a minimal repro, I'll open an issue on Verify.  Playing a bit more with it, I don't think it's a Rider bug necessarily, but it manifests if you clone the `sentry-dotnet` repo clean (or `git clean -dfx` to clean an existing clone), then `dotnet build SentryCore.slnf` - the `Sentry.Tests.dll` assembly that gets built has the lower-cased drive letter.  Then open `SentryCore.slnf` in Rider and run tests, and those two will fail.

Tried various things already to reproduce minimally, and I never get the lower-cased drive letter.

(Using Rider's "Assembly Explorer" to check the metadata on the assembly is pretty neat, BTW)

mattjohnsonpint:
But honestly, the workaround took about 10 minutes tops, so not sure how much time I want to invest here.

PR #2191, opened 2023-02-19
test: add symbol upload test

### LINKED ISSUE #2152: Integration test for symbol upload
### Problem Statement

sentry-cli was bundled and now through msbuild we can have symbols uploaded to Sentry
No integration tests exist for this step

### Solution Brainstorm

See

* https://github.com/getsentry/sentry-unity/pull/763
* https://github.com/getsentry/sentry-fastlane-plugin/pull/130

vaind:
closes #2152 

#skip-changelog


test/sentry-cli-integration/server-output-Linux.expected
```diff
@@ -0,0 +1,586 @@
...
+     System.Diagnostics.TraceSource.dll.so
+     libaot-Microsoft.Extensions.Logging.dll.so
+     libsentry-android.so
+     Microsoft.Maui.dll.so
+     libaot-System.Collections.NonGeneric.dll.so
+     Xamarin.AndroidX.Navigation.UI.dll.so
+     libaot-Xamarin.AndroidX.AppCompat.AppCompatResources.dll.so
+POST /api/0/projects/org/project/reprocessing/ HTTP/1.1
+GET /STOP HTTP/1.1
+HTTP server stopping!
```
vaind, line 586: Before I finish this, I think we should have a look if we should really upload all those intermediary symbols. I know this came up before but I think it wasn't really resolved. Uploading 800+ files doesn't look like the greatest idea ever if we're not using them at all.

BTW it's possible to specify the `DifType` we want to upload when calling CLI `debug-files upload`:
```
    -t, --type <TYPE>                Only consider debug information files of the given type.  By
                                     default, all types are considered. [possible values:
                                     bcsymbolmap, dsym, elf, pe, pdb, portablepdb, sourcebundle,
                                     breakpad, proguard, wasm]
```

If we know we only need PPDB, let's only use that.

@mattjohnsonpint 
@bruno-garcia

mattjohnsonpint: Agree on the tons of files, and yeah - we primarily need the `portablepdb` type.  Or the `pe` type if the symbols are embedded.  The problem is that we don't have a good sense of which debug files will be useful during a native crash on each platform.  If we were to be more selective, we'd need to specify different types and paths based on target platform and various other build properties.  We'd likely miss something.  Certainly it could be done, but would take a lot of experimentation.

Also, I think we'd need the ability to specify more than one type.  Not sure if that's already supported or not.

bruno-garcia: catching everything form bin/Release should be fine for all the managed code (ppdb) but if we're building a mobile app, there'll be some native libraries that never make it to the bin/Release folder. So maybe we need different invocations. One for the .NET managed. And other(s) for the native stuff that only made somewhere into the `obj` folder

vaind: BTW this is what's in the release folder
image

bruno-garcia: Makes sense. That looks good.

If we only collect that folder, we'll miss out on debug files that were tripped off of native code that got into the `.apk`

vaind: I've created a follow-up issue and disabled the test part that matches the expected file list temporarily. https://github.com/getsentry/sentry-dotnet/issues/2260

mattjohnsonpint:
Looks pretty good.  It seems that #2260 will need to be done before we can merge this though.  Then the resulting files may change depending on build platform.

Also, can you add at least one test for `UploadSources=true;UploadSymbols=false`?  Thanks.

vaind:
> Looks pretty good. It seems that #2260 will need to be done before we can merge this though. Then the resulting files may change depending on build platform.

I don't see why. Part of the PR resolving #2260 can be updating the list of resulting files - that's actually the goal

> Also, can you add at least one test for `UploadSources=true;UploadSymbols=false`? Thanks.

mattjohnsonpint:
Aren't the `return` with `todo` statements in the script showing that the only thing being tested presently is that the build didn't fail?

vaind:
> Aren't the `return` with `todo` statements in the script showing that the only thing being tested presently is that the build didn't fail?

Well yes, but also I've tested these locally and they worked, it just didn't make sense to me to add the platform-specific files to the list and then remove them when that issue is implemented. But feel free to keep out of the `main` branch if you don't like merging it in this state. I just didn't want to let `Sentry.targets` diverse again and having to merge all the changes.

mattjohnsonpint:
Ok.  I'll merge this as-is, and then update the list of files while fixing #2260.  Thanks for the suggestion. 👍

mattjohnsonpint (approved):

PR #2195, opened 2023-02-22
WIP: Remove floating versions

mattjohnsonpint:
Floating versions in tests and samples have reduced frequency of deps-only PRs, but they have occasionally broken the CI build at inconvenient times.  We should go back to using regular version numbers.  In cases where we need to not take latest, we should use semver ranges (such as `[1.2.3,2)`).  We will need to update the dependency updating scripts and/or dependabot to support these.

#skip-changelog

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -51,9 +51,9 @@
   </PropertyGroup>
 
   <ItemGroup>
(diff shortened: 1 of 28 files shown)


bitsandfoxes:
Superseded by https://github.com/getsentry/sentry-dotnet/pull/2465

PR #2197, opened 2023-02-24
Allow SentryUploadSources to work even when not uploading symbols

### LINKED ISSUE #2196: Detach SentryUploadSources from SentryUploadSymbols
We have two [MSBuild properties](https://docs.sentry.io/platforms/dotnet/configuration/msbuild/#optional-msbuild-properties):

- `SentryUploadSymbols` - On by default for release builds, but off for debug builds unless set `true`.
- `SentryUploadSources` - Off by default always.

Currently, `SentryUploadSources` works by adding the `--include-sources` option while uploading symbols.  If symbols aren't being uploaded, then `SentryUploadSources` does nothing.

We don't actually have to bind them in this way.  Sentry CLI has separate commands for creating source bundles and uploading them separately from other debug files.  See https://docs.sentry.io/product/cli/dif/#creating-source-bundles

We should use that to allow `SentryUploadSources` to work regardless of the `SentryUploadSymbols` setting.  (Sources are still useful in debug builds.)

mattjohnsonpint:
Resolves #2196 

Caveat: Symbols still need to be *created*, even if they're not uploaded.  If symbols are not created, we can't resolve the files needed to generate the source bundle.

Additionally, if symbols are *embedded* (using `<DebugType>embedded</DebugType`), sources won't be uploaded due to https://github.com/getsentry/sentry-cli/issues/1491.  Once that is fixed, the subsequent version of Sentry CLI will remove this limitation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ### Features
 
-- Allow SentryUploadSources to work even when not uploading symbols ([#2197](https://github.com/getsentry/sentry-dotnet/pull/2197))
+- Allow `SentryUploadSources` to work even when not uploading symbols ([#2197](https://github.com/getsentry/sentry-dotnet/pull/2197))
 
 ### Fixes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Features
+
+- Allow SentryUploadSources to work even when not uploading symbols ([#2197](https://github.com/getsentry/sentry-dotnet/pull/2197))
+
 ### Fixes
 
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Features
+
+- Allow SentryUploadSources to work even when not uploading symbols ([#2197](https://github.com/getsentry/sentry-dotnet/pull/2197))
```
SeanFeldman, line 7:
```suggestion
- Allow `SentryUploadSources` to work even when not uploading symbols ([#2197](https://github.com/getsentry/sentry-dotnet/pull/2197))
```

SeanFeldman (approved):
One small nit

PR #2206, opened 2023-03-02
feat: Profiling

### LINKED ISSUE #1955: Support for profiling
Possibly through the same apis used by `dotnet trace`

vaind:
Part of #1955 

This uses [Diagnostic IPC](https://github.com/dotnet/diagnostics/blob/main/documentation/design-docs/ipc-protocol.md) to connect to the CLR in-process, capture `Microsoft-DotNETCore-SampleProfiler` trace events and transforms them to the format Sentry understands. 

In its current state, the feature is in a good shape for alfa testing but there's a major followup to resolve and that is how `TraceLog`, which is a tool to process raw events and get figure out stack traces,  works, i.e. that it requires files and needs to be saved & then loaded from file to be actually usable (see [here](https://github.com/microsoft/perfview/issues/1829)). I'm starting to lean towards a fork/copy of relevant parts where we gut it to get what we need and drop the rest. That way we could also do without the rundown provider that gives as all past .net runtime events when we start a profile. Functionally (from the user perspective), the profile should look the same, but the implementation should be more efficient than it is now, especially the processing that happens after the profiling stops and also it would be most noticeable for short transactions if we could avoid requesting all past CLR runtime events (for simple app, it's about 3 MB of data).

### TODOs

* [x] test full SDK setup & finish right after a transaction is finished to make sure the processing is waited-for in the BackgroundWorker FlushAsync() and the profile appears in the envelope.
* [x] handle exceptions gracefully
* [x] limit long running transactions (30s)
* [x] maybe make processing asynchronous, or run the whole thing on a different thread
* [x] fix to support interleaving transactions
* [x] downsample during processing (currently downsampled while converting nettrace to ETLX)
* [x] configure in-app

### Followups
* [ ] ~maybe filter-out threads that don't run any user code?~ marked as in-app=false
* [ ] ~should we filter-out sentry-specific/profiling-specific samples?~ marked as in-app=false
* [ ] provide `debug_meta` images if we want to show line numbers  - this may need to be opt-in as it will likely add some runtime overhead
* [ ] downsample during ingest, if possible - currently, the runtime samples at 1000 Hz (1 ms) [without any way to configure that yet](https://github.com/dotnet/runtime/issues/82939), but we want it to be closer to 101 Hz - we should be able to filter samples directly in the `_session.EventStream` before even transforming to ETLX.
* [ ] filter out repeating stack_id for a given thread (i.e. don't send new timestamps unless the stack trace changes) - this may be a follow-up and should also be considered for other SDKs because it may significantly reduce the payload size in multithreaded environments. For example, this thread could be significantly reduced: https://sentry-sdks.sentry.io/profiling/profile/sentry-dotnet/3f359ed615d94111bd927e821917d1c8/flamechart/?colorCoding=by+symbol+name&query=&sorting=call+order&tid=2&view=top+down&xAxis=transaction 
* [ ] [it may be possible](https://medium.com/criteo-engineering/spying-on-net-garbage-collector-with-traceevent-f49dc3117de) to handle GCstart & GCend events to show garbage collection
* [ ] collect `Microsoft-Windows-DotNETRuntime` rundown provider (loaded libs, etc) in a separate session and merge with actual samples later during processing, if possible (merging nettrace files, or having a custom provider for TraceLog). 
* [ ] Or maybe we don't even need to collect these because we're in the same process? can we hack our way around and get the libraries from the current process instead? Probably not so easy because it also somehow stack walking if I'm not mistaken.
* [ ] Get rid of writing the intermediary files if at all possible - doesn't seem to be doable at the moment, raised https://github.com/microsoft/perfview/issues/1829
* [ ] Figure out thread names (we only have native thread IDs though so it seems to be a problem)
* [ ] Consider capturing only the current "activity", i.e. not all threads but only what belongs to the transaction.
* [ ] profile & optimize the capture and processing


src/Sentry/Protocol/Envelopes/Envelope.cs
```diff
@@ -285,11 +284,16 @@ public static Envelope FromTransaction(Transaction transaction)
...
 
-        var items = new[]
+        var items = new List<EnvelopeItem>
         {
             EnvelopeItem.FromTransaction(transaction)
         };
 
+        if (transaction.TransactionProfiler is { } profiler)
+        {
+            items.Add(EnvelopeItem.FromProfileInfo(profiler.Collect(transaction)));
```
bruno-garcia, line 294: If it throws here? And if it doesn't, do we can an empty profile?
Worth commenting the behavior here at least

src/Sentry/Protocol/Envelopes/EnvelopeItem.cs
```diff
@@ -200,6 +201,20 @@ public static EnvelopeItem FromTransaction(Transaction transaction)
...
+    /// </summary>
+    internal static EnvelopeItem FromProfileInfo(Task<ProfileInfo?> profileInfo)
+    {
+        var header = new Dictionary<string, object?>(1, StringComparer.Ordinal)
+        {
+            [TypeKey] = TypeValueProfile
+        };
+
+        var jsonSerializable = profileInfo.ContinueWith(task => task.Result as IJsonSerializable);
+        return new EnvelopeItem(header, new AsyncJsonSerializable(jsonSerializable));
```
bruno-garcia, line 215: This is a task, right?

```suggestion
        var jsonSerializableTask = profileInfo.ContinueWith(task => task.Result as IJsonSerializable);
        return new EnvelopeItem(header, new AsyncJsonSerializable(jsonSerializableTask));
```

src/Sentry/Protocol/ProfileInfo.cs
```diff
@@ -0,0 +1,86 @@
...
+        //  https://github.com/getsentry/relay/blob/master/relay-profiling/src/sample.rs#L117
+        writer.WriteStartObject("device");
+        writer.WriteString("architecture", _contexts.Device.Architecture ?? "");
+        writer.WriteStringIfNotWhiteSpace("manufacturer", _contexts.Device.Manufacturer);
+        writer.WriteStringIfNotWhiteSpace("model", _contexts.Device.Model);
+        writer.WriteEndObject();
+
+        // TODO writer.WriteSerializable("os", _contexts.OperatingSystem, logger);
+        //  https://github.com/getsentry/relay/blob/master/relay-profiling/src/sample.rs#L102
+        var rawOS = _contexts.OperatingSystem.RawDescription?.Replace("Microsoft Windows", "Windows")?.Split(' ', 2);
```
bruno-garcia, line 65: We parse this in Relay: https://github.com/getsentry/relay/blob/1a7f256f65a51da069959adf3eb5619c63660af4/relay-general/src/store/normalize/contexts.rs#L330

samples/Sentry.Samples.AspNetCore.Mvc/Sentry.Samples.AspNetCore.Mvc.csproj
```diff
@@ -14,6 +14,7 @@
 
   <ItemGroup>
     <ProjectReference Include="..\..\src\Sentry.AspNetCore\Sentry.AspNetCore.csproj" />
+    <ProjectReference Include="..\..\src\Sentry.Extensions.Profiling\Sentry.Extensions.Profiling.csproj" />
```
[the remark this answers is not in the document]
vaind: OK, renamed

src/Sentry.Profiling/TraceLogProcessor.cs
```diff
@@ -0,0 +1,678 @@
...
+
+namespace Sentry.Profiling;
+
+// A list of frame indexes.
+using SentryProfileStackTrace = HashableGrowableArray<int>;
+
+/// <summary>
+/// Processes TraceLog to produce a SampleProfile.
+/// Based on https://github.com/microsoft/perfview/blob/ff6e8e6c4118a26521515f41ae77f15981d68c53/src/TraceEvent/Computers/SampleProfilerThreadTimeComputer.cs
+/// </summary>
```
bruno-garcia, line 18:
```suggestion
/// Note that several parts of the code below are commented out as we intend to use this in follow up PRs
/// </summary>
```

bruno-garcia:
Outcome of a chat between @vaind and I:

* There's a lot more we can/want to do but this PR is at a good state to get something `alpha` and feedback.
* Ideally we don't want to publish `Sentry.Profiling`. We can add it to `Sentry` but that's be only under `netX.0` (.NET Core) target we support. We need to communicate this clearly in docs. 
* The option to turn it on should be available on all TFMs but with a `[Obsolete` or whatever gives a warning/error at compile time that the target is not supported to avoid support request (why option not available).
* We'll use GlobalMode to set the behavior: all threads in process (global Mode) or only activity/request (server mode)
* There are changes that need to be done on the TraceEvent so we can have a fork. But we need to figure out if we can have 0 dependencies at the end (right now we have 2 dependencies on Sentry.Profiling) do we need 2 forks or just vendor 2 files? The motivation: We want to possibly have profiling on by default if tracing is turned on. Same approach we use for Ben.Demystifier with changing types to `internal` so we can change approach without breaking change.
* We should support the lowest .NET Core version possible but for as long as we have no friction to do so. If supporting .NET 6 is the simplest/what we get OOTB, we should take that route. It's a new feature and we expect new customers/those upgrading to take advantage of. in other words: No .NET Framework or unsupported .NET Core as a goal.
* To be clear: This is only for client mode right now (desktop/cli). Server mode (Global mode/aspnet) will come in follow up PRs. Mobile isn't supported either since it's not based on CoreCLR. We can revisit this in the future.

vaind:
@mattjohnsonpint have you had a chance to review this yet? Any reservations against merging to the main package as described by Bruno last Monday? Any other inputs?

mattjohnsonpint:
Sorry, not just yet.  Will do soon.  Thanks.

vaind:
> Sorry, not just yet. Will do soon. Thanks.

What's the status? When should I expect the info? Just to make it clear, there's currently nothing I can do on this PR at this time (without blowing it up and making even less manageable), so I'm blocked until the questions from the previous comment are answered.

mattjohnsonpint:
Sorry for the delay.  I'm reviewing it now.  Thanks.

src/Sentry/Internal/ITransactionProfiler.cs
```diff
@@ -0,0 +1,30 @@
...
+{
+    /// <summary>
+    /// Called when the transaction ends - this should stop profile samples collection.
+    /// </summary>
+    void OnTransactionFinish(DateTimeOffset now);
+
+    /// <summary>
+    /// Process and collect the profile.
+    /// </summary>
+    Task<ProfileInfo?> Collect(Transaction transaction);
```
mattjohnsonpint, line 29:
```suggestion
    Task<ProfileInfo?> CollectAsync(Transaction transaction);
```

src/Sentry.Profiling/SamplingTransactionProfiler.cs
```diff
@@ -0,0 +1,285 @@
...
+                if (_cancellationToken.IsCancellationRequested)
+                {
+                    return null;
+                }
+
+                _options.LogDebug("Converting profile to Sentry format.");
+
+                var processor = new TraceLogProcessor(_options, traceLog)
+                {
+                    MaxTimestampMs = (ulong)(_endTime.Value - _startTime).TotalMilliseconds
```
mattjohnsonpint, line 156: This seems like we should be using a stopwatch directly.

src/Sentry/Internal/ITransactionProfiler.cs
```diff
@@ -0,0 +1,30 @@
...
+
+/// <summary>
+/// Factory to create/attach profilers when a transaction starts.
+/// </summary>
+internal interface ITransactionProfilerFactory
+{
+    /// <summary>
+    /// Called during transaction start to start a new profiler, if applicable.
+    /// </summary>
+    ITransactionProfiler? OnTransactionStart(ITransaction tracer, DateTimeOffset now, CancellationToken cancellationToken);
```
mattjohnsonpint, line 13: Why are we passing `DateTimeOffset` for start/finish methods?  Could the factory create a `Stopwatch` or `SentryStopwatch` internally?

If it would help, I can update `SentryStopwatch` to be stoppable.

mattjohnsonpint (changes requested):
I'll have more soon, but there's enough to start with here.  Thanks.

src/Sentry/Internal/ITransactionProfiler.cs
vaind: We need the same stopwatch as the transaction  runs with, otherwise, we may end up with a different timestamp, if the system time changes.

However, after having a second look at `SentryStopwatch`, I see that it's actually not guaranteed to be monotonic anyway 🤔

src/Sentry/Protocol/SampleProfile.cs
```diff
@@ -0,0 +1,54 @@
...
+
+/// <summary>
+/// Sentry sampling profiler output profile
+/// </summary>
+internal sealed class SampleProfile : IJsonSerializable
+{
+    public GrowableArray<Sample> Samples = new(10000);
+    public GrowableArray<SentryStackFrame> Frames = new(100);
+    public GrowableArray<SentryProfileStackTrace> Stacks = new(100);
+    public List<SentryThread> Threads = new(10);
```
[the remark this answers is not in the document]
vaind: Is there any actual benefit in changing to properties, with this being an `internal` class? I thought it's only useful if you need to change the implementation in the future without affecting the interface. I'm fine with it even if it's "only" a convention in this case, just wanted to understand if I'm missing something useful that it would bring.

test/Sentry.Tests/Protocol/ProfilerTests.verify.cs
```diff
@@ -0,0 +1,101 @@
+namespace Sentry.Tests.Protocol;
+
+[UsesVerify]
+public partial class ProfilerTests
```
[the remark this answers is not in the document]
vaind: Actually it is a "verify" test for both `SampleProfile` and `ProfileInfo`, so renaming makes it even less clear. I'd need to split which seems like an overkill for a single test case. But let me know if you want me to do that (either of the options) anyway

src/Sentry/Internal/GrowableArray.cs
```diff
@@ -0,0 +1,426 @@
...
+}
+
+/// <summary>
+/// A cheap version of List(T). The idea is to make it as cheap as if you did it 'by hand' using an array and
+/// an int which represents the logical charCount. It is a struct to avoid an extra pointer dereference, so this
+/// is really meant to be embedded in other structures.
+/// </summary>
+/// <remarks>
+/// Adapted to null-safety from the original version licensed under MIT and located at:
+/// https://github.com/microsoft/perfview/blob/050c303943e74ff51ce584b2717e578d96684e85/src/FastSerialization/GrowableArray.cs
```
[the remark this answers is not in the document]
vaind: Thanks for the pointer, I wasn't aware. Leaving as a TODO for later, after the decision about forking stuff from perfview repo.

src/Sentry.Profiling/SamplingTransactionProfiler.cs
mattjohnsonpint: I think `SamplingTransactionProfiler` should keep an internal stopwatch.  It shouldn't ask for the start/stop times (or another stopwatch) from the transaction.  `DateTimeOffset now` can be removed from the `Start` and `Finish` methods on the interface.  There's really no need (as far as I can tell) for the profiler to care about the date and time-of-day, only the elapsed duration.

I see where the profile's `StartTimeStamp` is serialized, but that can come from the transaction directly.

From https://develop.sentry.dev/sdk/profiles/ 

> Ideally, this will be the same as the field start_timestamp on the transaction so all relative timestamps for transaction and profile can be both from the same reference point and we don't need to correct relative timestamps.

mattjohnsonpint (changes requested):

src/Sentry/Protocol/Envelopes/Envelope.cs
```diff
@@ -285,11 +282,17 @@ public static Envelope FromTransaction(Transaction transaction)
...
-        var items = new[]
+        var items = new List<EnvelopeItem>
         {
             EnvelopeItem.FromTransaction(transaction)
         };
 
+        if (transaction.TransactionProfiler is { } profiler)
+        {
+            // Profiler.CollectAsync() must not throw but it may complete with null in which case the EnvelopeItem won't serialize.
+            items.Add(EnvelopeItem.FromProfileInfo(profiler.CollectAsync(transaction)));
```
mattjohnsonpint, line 293: In case of `null`, the `Payload` of the `EnvelopeItem` will serialize empty, but `EnvelopeItem` will still serialize.  It will have an item header showing `{"length": 0}`, and and then an empty item.  Take a close look at `EnvelopeItem.SerializeAsync`.

While you're there, you'll want to read the comments about payload buffering being synchronous.  I think for `AsyncJsonSerializable` payloads, it makes sense to go back to asynchronous buffering.  Otherwise the profile collection is actually a blocking call.

Also, we'll want some way to ensure that `Envelope.SerializeAsync` doesn't append extraneous newlines when there was nothing serialized for each item.

src/Sentry/Protocol/ProfileInfo.cs
```diff
@@ -0,0 +1,90 @@
...
+    public string? Environment { get; set; }
+
+    public string? Platform { get; set; } = Constants.Platform;
+
+    public string? Release { get; set; }
+
+    public Transaction? Transaction { get; set; }
+
+
+    public void WriteTo(Utf8JsonWriter writer, IDiagnosticLogger? logger)
```
mattjohnsonpint, line 41: So much duplication here with the serialization we already have on each context class.  There's also a lot more information available there, not being serialized here.  Is that intentional?

mattjohnsonpint (changes requested):

src/Sentry.Profiling/SampleProfilerSession.cs
```diff
@@ -0,0 +1,46 @@
...
+using Microsoft.Diagnostics.Tracing.EventPipe;
+using Microsoft.Diagnostics.Tracing.Parsers;
+
+namespace Sentry.Profiling;
+
+internal class SampleProfilerSession
+{
+    private DiagnosticsClient _client;
+    private EventPipeSession _session;
+    private MemoryStream _stream = new MemoryStream();
```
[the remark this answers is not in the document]
mattjohnsonpint: Specifically, consider long-running transactions.  The `EventPipeSession` uses a circular buffer, but we're copying continuously from the buffer to this memory stream.  How big will it grow?  What if there's just one transaction for the lifetime of the application?  Seems like we need to impose some limitations?

src/Sentry/Internal/ITransactionProfiler.cs
vaind: Understood & changed.

samples/Sentry.Samples.Console.Profiling/Program.cs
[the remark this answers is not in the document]
vaind: I've also considered removing the rest, but it also provides some value when working on the profiler to see other stack traces.

src/Sentry.Profiling/SampleProfilerSession.cs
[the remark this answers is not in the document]
vaind: We're limiting the profile collection to 30 seconds (from https://develop.sentry.dev/sdk/profiles/#validation) in `SamplingTransactionProfiler` - creating a delayed task that stops the profiler:
```c#
        Task.Delay(timeoutMs, _cancellationToken).ContinueWith(_ =>
        {
            if (Stop(TimeSpan.FromMilliseconds(timeoutMs)))
            {
                _options.LogDebug("Profiling is being cut-of after {0} ms because the transaction takes longer than that.", timeoutMs);
            }
        }, CancellationToken.None);
```

src/Sentry/Protocol/ProfileInfo.cs
vaind: Unfortunately yes, the deserialization on ingest is implemented manually and doesn't match these contexts. For example, OS name and version are required values while in our OS context, we don't set them at all (only raw description is set). Without these values, the profile will be dropped during ingest because it will raise an error when deserializing.

src/Sentry/Protocol/Envelopes/JsonSerializable.cs
```diff
@@ -37,3 +37,50 @@ public void Serialize(Stream stream, IDiagnosticLogger? logger)
...
+#else
+        await using (writer.ConfigureAwait(false))
+#endif
+        {
+            if (await Source.ConfigureAwait(false) is { } source)
+            {
+                source.WriteTo(writer, logger);
+                await writer.FlushAsync(cancellationToken).ConfigureAwait(false);
+            }
+        }
```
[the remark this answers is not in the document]
vaind: Results in `error CA2007: Consider calling ConfigureAwait on the awaited task` on the `await using var writer = new Utf8JsonWriter(stream);` line. I didn't figure out what to do about that.

src/Sentry/Protocol/Envelopes/Envelope.cs
vaind: > In case of `null`, the `Payload` of the `EnvelopeItem` will serialize empty, but `EnvelopeItem` will still serialize. It will have an item header showing `{"length": 0}`, and and then an empty item. Take a close look at `EnvelopeItem.SerializeAsync`.

I'm not sure what to do about that `ISerializable` is public so can't easily change that and matching specifically for `AsyncJsonSerializable` doesn't seem like the best idea... How about changing the `public async Task<ProfileInfo?> CollectAsync(...)` to a non-null task completion value `public async Task<ProfileInfo> CollectAsync(...)` and returning a default object in case of any of the invalid states? It should be an uncommon scenario anyway.

> While you're there, you'll want to read the comments about payload buffering being synchronous. I think for `AsyncJsonSerializable` payloads, it makes sense to go back to asynchronous buffering. Otherwise the profile collection is actually a blocking call.

It is blocking, but doing so in the background worker - that should be fine, shouldn't it? Bruno was OK with that back when we had a call but if you feel we should change that, please let me know (ideally with a proposal).

src/Sentry.Profiling/SampleProfilerSession.cs
```diff
@@ -0,0 +1,46 @@
...
+            new EventPipeProvider("Microsoft-Windows-DotNETRuntime", EventLevel.Informational, (long)ClrTraceEventParser.Keywords.Default),
+            new EventPipeProvider("Microsoft-DotNETCore-SampleProfiler", EventLevel.Informational),
+            new EventPipeProvider("System.Threading.Tasks.TplEventSource", EventLevel.Informational, (long)TplEtwProviderTraceEventParser.Keywords.Default)
+        };
+
+        // The size of the runtime's buffer for collecting events in MB, same as the current default in StartEventPipeSession().
+        var circularBufferMB = 256;
+        _session = _client.StartEventPipeSession(providers, true, circularBufferMB);
+        _copyTask = _session.EventStream.CopyToAsync(_stream, _cancellationToken);
+        _cancellationTokenStopRegistration = _cancellationToken.Register(() => _session.Stop(), false);
```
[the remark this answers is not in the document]
vaind: > If I understand correctly, this is going to start a profiling session and start copying data to the stream immediately.
> 
> Is there risk that `_copyTask` will complete prematurely? Or is it guaranteed to keep streaming until it is awaited?

I'm not aware either way - it's the standard `Stream.CopyToAsync()`. What would we need to change either way?

> Also, I tend to prefer not starting actual work from within a constructor. Constructors should generally create objects without having side effects (starting tasks, performing file or network i/o, etc.). We have a few violations of this already (`BackgroundWorker`, for example) - but I don't want to continue this anti-pattern. The better approach is to use a static factory method. Thanks.

OK, changed added `static SampleProfilerSession.StartNew()` and `SamplingTransactionProfiler.Start()`

src/Sentry/Protocol/Envelopes/Envelope.cs
mattjohnsonpint: If it's uncommon, then have it be non-nullable and throw a `SerializationException` if it can't be serialized instead of returning null.  Then try/catch/log that exception in envelope serialization to avoid adding the extra newline and continue with the rest of the envelope.

vaind: > https://develop.sentry.dev/sdk/envelopes/  
> Length-prefixed payloads must terminate with \n or EOF. The newline is not considered part of the payload. Any other character, including whitespace, means the Envelope is malformed.

That implies the new line should be there regardless of the payload length.

mattjohnsonpint: On the async payload serialization - I made an error in switching it to always buffer synchronously.  It should only be doing so for `JsonSerializable` payloads.  Others such as `StreamSerializable` (used for attachments) and this new `AsyncJsonSerializable` should still buffer asynchronously.  I'll make a PR for that separately so you don't need to deal with it here.

mattjohnsonpint: Async buffering adjusted in #2297.

mattjohnsonpint: > That implies the new line should be there regardless of the payload length.

Correct.  I'm saying that if the profile info cannot be serialized because it doesn't exist, then don't serialize an envelope item for it at all, which would include the newline.

src/Sentry.Profiling/SampleProfilerSession.cs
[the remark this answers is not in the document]
mattjohnsonpint: Ok.  That makes sense.  And I also see where the stream is ultimately disposed (in `CreateTraceLogAsync`).  But where is the `EventPipeSession` disposed?

src/Sentry.Profiling/SampleProfilerSession.cs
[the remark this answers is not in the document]
mattjohnsonpint: The factory methods looks great.  Thanks!

On `CopyToAsync` - that will copy until reaching the end of the stream.  But I'm not sure if the end of the stream means when it read zero bytes from the source stream, or if it will keep at it until the source stream is closed.  I'll check and get back to you.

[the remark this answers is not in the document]
mattjohnsonpint: Yeah, I think the copy is problematic.  There are two problems:

- As suspected, copying will stop at the end of the stream, but that is determined by zero bytes remaining to be read from the source stream - not by the stream being closed.
- Copying starts based on the position of the stream at the time the copy method is invoked.  Writing bytes to the source stream will move the position to the end.

Consider this small test program:

```csharp
using var source = new MemoryStream();
using var destination = new MemoryStream();

// this byte will copy to the destination stream
source.WriteByte(1);

// ... but only if the position is reset before copying
source.Position = 0;

var copyTask = source.CopyToAsync(destination);
Console.WriteLine(copyTask.Status);

// this will not copy, because the task has already ran to completion
source.WriteByte(2);

await copyTask;

Dump(source);
Dump(destination);

void Dump(MemoryStream stream)
{
    var bytes = stream.GetBuffer()[..(int) stream.Length];
    Console.WriteLine($"[{string.Join(',', bytes)}]");
}
```

src/Sentry/Protocol/SampleProfile.cs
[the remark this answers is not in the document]
mattjohnsonpint: In this particular case, not really - because the class is internal, so the fields are the same as internal even though they're public.  The only risk is that if the class ever gets made public, the fields would become part of the public API.  You could mark them internal now to avoid that.

We generally don't want fields exposed on a public API because it restricts how they can evolve over time.  Rule of thumb is private fields, public properties.  Internals are sorta meh. 😄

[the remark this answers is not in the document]
vaind: changed to properties

src/Sentry.Profiling/SampleProfilerSession.cs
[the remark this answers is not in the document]
vaind: While this applies to `MemoryStream`, it doesn't imply the same is true for the type of stream that is `_session.EventStream`. The `CopyToAsync()` method there seems to be implemented differently. Also see the same pattern here: https://github.com/dotnet/diagnostics/blob/main/documentation/diagnostics-client-library-instructions.md#4-trigger-a-cpu-trace-for-given-number-of-seconds

src/Sentry.Profiling/SampleProfilerSession.cs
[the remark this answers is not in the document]
vaind: I've added an explicit `Dispose()` to `FinishAsync()`

src/Sentry.Profiling/SampleProfilerSession.cs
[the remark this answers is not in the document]
mattjohnsonpint: Ok.  That looks to be a `NamedPipeClientStream`, which isn't seekable - so that addresses the position issue.  I think the rest of it is working as a side effect of the `EventPipeSession` starting first.  It must start writing to the stream immediately and continuously.  So when copying will continue until after the session is stopped, and all the data from that session is read.

This will work fine I think.  Thanks for walking through it with me. 👍

src/Sentry.Profiling/SamplingTransactionProfiler.cs
```diff
@@ -0,0 +1,200 @@
...
+        typeof(TraceLog)
+            .GetMethod(
+                "CreateFromEventPipeEventSources",
+                BindingFlags.Public | BindingFlags.NonPublic | BindingFlags.Static,
+                new Type[] { typeof(TraceEventDispatcher), typeof(string), typeof(TraceLogOptions) })?
+            .Invoke(null, new object[] { source, etlxPath, new TraceLogOptions() { ContinueOnError = true } });
+
+        if (!File.Exists(etlxPath))
+        {
+            throw new FileNotFoundException($"Profiler failed at CreateFromEventPipeEventSources() - temproary file doesn't exist", etlxPath);
```
mattjohnsonpint, line 193:
```suggestion
            throw new FileNotFoundException($"Profiler failed at CreateFromEventPipeEventSources() - temporary file doesn't exist", etlxPath);
```

src/Sentry/Protocol/Envelopes/Envelope.cs
vaind: I've made the change to make FinishAsync() throw & also handle it during Envelope serialization so it won't have an item in that case

mattjohnsonpint:
Looks like something in the latest changes broke things.  Everything else looks good, so after resolving that all that's needed is the license file and and then I will approve.  Thanks!!!

mattjohnsonpint:
Oh, and a changelog entry also.  Let's go with "Initial work to support profiling in a future release." - since we're not making it available just yet.

And we should also add a `<IsPackable>false</IsPackable>` to `Sentry.Profiling.csproj` so we don't attempt to publish a nupkg for it yet.

mattjohnsonpint (approved):

mattjohnsonpint:
Merged.  Thanks!!!

filipnavara:
> Mobile isn't supported either since it's not based on CoreCLR.

JFYI MonoVM does have the identical sampling profiler mechanism.

PR #2211, opened 2023-03-05
Adjust Android AVD images

mattjohnsonpint:
This is a test to see if we get better perf running on the `aosp-atd` images where available.  Also test using `x86` instead of `x86_64` where available.  (Note API 30 only comes in x86 for ATD images anyway.)

#skip-changelog

### DIFF
--- a/.github/workflows/device-tests-android.yml
+++ b/.github/workflows/device-tests-android.yml
@@ -55,7 +55,7 @@ jobs:
     runs-on: ubuntu-latest-4-cores
 
     # We currently test on:
-    #  Android  5 (API 21) - because it's the oldest version we can support with .NET
+    #  Android 5.1 (API 22) - because it's the oldest version we can support without the emulator crashing on boot
     #  Android 10 (API 29) - because it's the oldest version still in active support
     #  Android 11 (API 30) - because it has an ATD image, designed for testing
     #  Android 13 (API 33) - because it's the latest stable version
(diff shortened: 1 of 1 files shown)


mattjohnsonpint:
Hmmm.. While this didn't hurt anything, it didn't make it much better.  Each android run takes anywhere between 2 to 6 minutes, and it seems rather random.

When it's slow, one thing that stands out is when the runner recursively sets permissions with `chown -R`.  Here it took about two minutes:

image

But sometimes this is really fast.  Here it only took 15 seconds to do the same thing:

image

There doesn't seem to be much pattern to when it's slow or fast.

It may not be necessary to set permissions recursively on the entire Android SDK tree like that.  Some further investigation could help.  It would require fixing upstream in the repo for the Reactive Circus emulator action.  We can do this some other time.

As for the images, either the old ones or the new ones work fine.  The x86 images seem slightly smaller, so maybe worth keeping this.  The aosp-atd image for Android 30 is quite nice also.

mattjohnsonpint:
Also, I tried the aosp-atd image for Android 31, which is only on the canary channel.  It worked (xharness tests passed) but it spewed a ton of errors to the logs like these:

image

See test run here:
https://github.com/getsentry/sentry-dotnet/actions/runs/4337786128/jobs/7574269604

So until there's a stable (non-canary) release of that, I don't think we should use it.

jamescrosswell:
Closing this as a cold case. We can reopen if necessary/appropriate.

PR #2216, opened 2023-03-06
Make `SentryUploadSymbols` strictly opt-in

mattjohnsonpint:
The `<SentryUploadSymbols>` MSBuild property previously defaulted to `true` for projects compiled in `Release` configuration.  This led to some confusion when combining it with other options.  It is now `false` by default.

Going forward, all features enabled by MSBuild properties will be disabled by default, and strictly opt-in.

To continue uploading symbols, opt-in by setting `<SentryUploadSymbols>true</SentryUploadSymbols>`.  We recommend you do this for `Release` configuration only (because it will slow down local debug builds), but that is at your discretion.

See the [MSBuild Setup](https://docs.sentry.io/platforms/dotnet/configuration/msbuild/) docs for further details.

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -14,7 +14,16 @@
     <NoWarn>$(NoWarn);CS8002</NoWarn>
   </PropertyGroup>
 
-  <!-- Configure Sentry CLI -->
+  <!--
+    Configure Sentry CLI
+    NOTE: Authentication is not set here.  Instead:
+    - In CI, `SENTRY_AUTH_TOKEN` is set as an environment variable via a GitHub Actions secret.
+    - In dev, you can either set `SENTRY_AUTH_TOKEN` manually, or you can set an auth token
+      in your ~/.sentryclirc file.  See https://docs.sentry.io/product/cli/configuration/
+
+    There is also a a `SentryAuthToken` msbuild property, but it should be used sparingly.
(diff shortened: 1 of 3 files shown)


bruno-garcia (approved):

Directory.Build.props
```diff
@@ -12,12 +12,14 @@
...
+  </PropertyGroup>
 
-    <!-- Configure Sentry CLI -->
+  <!-- Configure Sentry CLI -->
+  <PropertyGroup Condition="'$(Configuration)' == 'Release'">
     <SentryOrg>sentry-sdks</SentryOrg>
     <SentryProject>sentry-dotnet</SentryProject>
-    <SentryUploadSources Condition="'$(Configuration)' == 'Release'">true</SentryUploadSources>
-
+    <SentryUploadSymbols>true</SentryUploadSymbols>
```
bruno-garcia, line 21:
```suggestion
    <!-- <SentryAuth>This value is set via environment variable but could be set here</SentryAuth > -->
    <SentryUploadSymbols>true</SentryUploadSymbols>
```

bruno-garcia: How about something like this so it's clear one value is "missing" but is set elsewhere?

mattjohnsonpint: For us, sure.  Just not for the public recommendation in the corresponding docs.

PR #2222, opened 2023-03-08
Get Debug ID for Full PDB format on Windows

### LINKED ISSUE #2221: Debug ID not generated for Full PDB format on Windows
### Package

Sentry

### .NET Flavor

Any

### .NET Version

Any

### OS

Windows

### SDK Version

3.29.0

### Steps to Reproduce

If a project has `<DebugType>full</DebugType>`, and it is compiled on Windows, it will have a full Windows PDB debug file generated.  Sentry CLI knows how to deal with these, but we still need to be able to send a `debug_id` at runtime to match.

Presently, we are not sending any `debug_id` except for *portable* PDB format.  We need to also send it when a *full* PDB was built.

This is reproducible on Sentry.Samples.Console.

```
dotnet run --project samples\Sentry.Samples.Console.Basic -f net48 -p:DebugType=full
```

Same for `net6.0`.  (It's not specific to .NET Framework.)

### Expected Result

The `debug_id` should be sent.  Symbolication and source context should work.

### Actual Result

The debug output will contain:

```
Info: Skipping DebugImage for module 'Sentry.Samples.Console.Basic.exe' because DebugId couldn't be determined
```

The event is missing a `debug_id`.  Symbolication and source context does not work.

mattjohnsonpint:
Fixes #2221.

Also applied some refactoring and cleanup.

### DIFF
--- a/src/Sentry/Internal/DebugStackTrace.cs
+++ b/src/Sentry/Internal/DebugStackTrace.cs
@@ -93,7 +93,7 @@ internal class DebugStackTrace : SentryStackTrace
                 {
                     if (i != j)
                     {
-                        relocations.Add(DebugStackTrace.GetRelativeAddressMode(i), DebugStackTrace.GetRelativeAddressMode(j));
+                        relocations.Add(GetRelativeAddressMode(i), GetRelativeAddressMode(j));
                     }
                     found = true;
                 }
@@ -101,7 +101,7 @@ internal class DebugStackTrace : SentryStackTrace
 
             if (!found)
             {
-                relocations.Add(DebugStackTrace.GetRelativeAddressMode(i), DebugStackTrace.GetRelativeAddressMode(@event.DebugImages.Count));
+                relocations.Add(GetRelativeAddressMode(i), GetRelativeAddressMode(@event.DebugImages.Count));
                 @event.DebugImages.Add(DebugImages[i]);
             }
         }
@@ -213,7 +213,7 @@ internal class DebugStackTrace : SentryStackTrace
 
             if (AddDebugImage(method.Module) is { } moduleIdx && moduleIdx != DebugImageMissing)
             {
-                frame.AddressMode = DebugStackTrace.GetRelativeAddressMode(moduleIdx);
+                frame.AddressMode = GetRelativeAddressMode(moduleIdx);
 
                 try
                 {
@@ -369,9 +369,9 @@ internal class DebugStackTrace : SentryStackTrace
         }
     }
 
-    private PEReader? TryReadAssembly(string assemblyName)
+    private static PEReader? TryReadAssembly(string assemblyName, SentryOptions options)
     {
-        if (_options.AssemblyReader is { } reader)
+        if (options.AssemblyReader is { } reader)
         {
             return reader.Invoke(assemblyName);
         }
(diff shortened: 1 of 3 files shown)


src/Sentry/Internal/DebugStackTrace.cs
```diff
@@ -382,77 +382,116 @@ private static void DemangleLambdaReturnType(SentryStackFrame frame)
...
+                        // Portable PDB Format
+                        // Version Major=any, Minor=0x504d
+                        debugId = $"{codeView.Guid}-{entry.Stamp:x8}";
+                    }
+                    else
+                    {
+                        // Full PDB Format (Windows only)
+                        // Version Major=0, Minor=0
+                        debugId = $"{codeView.Guid}-{codeView.Age}";
+                    }
```
mattjohnsonpint, line 463: This is the format we did not previously have.  Everything else in the PR is refactoring and optimization.

test/Sentry.Tests/HubTests.CaptureEvent_ActiveTransaction_UnhandledExceptionTransactionEndedAsCrashed.Net4_8.verified.txt
```diff
@@ -50,6 +50,16 @@
...
+              {
+                Type: pe_dotnet,
+                ImageAddress: null,
+                ImageSize: null,
+                DebugId: ________-____-____-____-____________-_,
+                DebugChecksum: null,
+                DebugFile: mscorlib.pdb,
+                CodeId: ______________,
+                CodeFile: .../mscorlib.dll
+              },
```
mattjohnsonpint, line 62: Hey look - now that are collecting it, we also get a debug id for mscorlib on .NET Framework. (This should light up some framework symbolication.)

src/Sentry/Internal/DebugStackTrace.cs
```diff
@@ -382,77 +382,116 @@ private static void DemangleLambdaReturnType(SentryStackFrame frame)
...
+            return null;
+        }
+
+        idx = DebugImages.Count;
+        DebugImages.Add(debugImage);
+        _debugImageIndexByModule.Add(id, idx);
+
+        return idx;
+    }
+
```
mattjohnsonpint, line 404: Note this is also just refactoring.  The `AddDebugImage` method now is entirely responsible for managing the cache, while `GetDebugImage` can focus on the hard parts.  (We can later add some tests on `GetDebugImage`.)

src/Sentry/Internal/DebugStackTrace.cs
```diff
@@ -382,77 +382,116 @@ private static void DemangleLambdaReturnType(SentryStackFrame frame)
...
+    internal static DebugImage? GetDebugImage(Module module, SentryOptions options)
```
mattjohnsonpint, line 405: Note this is also just refactoring.  The `AddDebugImage` method now is entirely responsible for managing the cache, while `GetDebugImage` can focus on the hard parts.  (We can later add some tests on `GetDebugImage`.)

src/Sentry/Internal/DebugStackTrace.cs
mattjohnsonpint, line 463: This is the format we did not previously have.  Everything else in the PR is refactoring and optimization.

bruno-garcia (commented):
What if we write something like `[assembly: SentrySourceContext]` if `<UploadSources>` or `<EmbedSources>` is on. In other words figure out at runtime that we in fact uploaded sources during build time?

And only then add debug_meta?

vaind:
> What if we write something like `[assembly: SentrySourceContext]` if `<UploadSources>` or `<EmbedSources>` is on. In other words figure out at runtime that we in fact uploaded sources during build time?
> 
> And only then add debug_meta?

Sounds great though it would break uses using sentry-cli manually.

mattjohnsonpint:
Another problem is, we'd have to write the attribute to the assembly during compilation before the upload was attempted.  So if the upload fails, it's to late to change the attribute.  I'll keep thinking about it though.  It might be ok.

Either way, it's a separate concern from this PR.  This one is specifically about getting the `debug_id` when the PDB is the full format instead of the portable one.  (other than the refactoring)

bruno-garcia:
> Sounds great though it would break uses using sentry-cli manually.

I would be surprised if people are using sentry cli manually. Is there a reason we need to support that? I feel the trade off here (vast majority will not opt in to this feature) indicates it's best to avoid the overhead of adding all this debug id's and extra debug_meta to events unless we know we need server side symbolication or source context. We could also get source context client side if sources are embedded in the pdb and released with the app (server use cases) avoiding the need to upload anything.

bruno-garcia (approved):

mattjohnsonpint:
Let's keep this issue focused on the full-pdb debug id. Moving discussion about whether debug images should be sent or not to #2227.  Thanks.

PR #2224, opened 2023-03-08
Fix debug files not uploading for `packages.config` nuget 

### LINKED ISSUE #2223: MSBuild integration doesn't work with packages.config
### Package

Sentry

### .NET Flavor

.NET Framework

### .NET Version

4.8

### OS

Windows

### SDK Version

3.29.0

_No response_

### Steps to Reproduce

- On Windows in Visual Studio 2022 latest, create a new .NET Framework console application.
- Add Sentry via the Nuget GUI in Visual Studio.  Note that it will use `packages.config`, not `<PackageReference>`.
- Setup Sentry's [MSBuild configuration](https://docs.sentry.io/platforms/dotnet/configuration/msbuild/). (auth token, org, project, upload symbols).
- Build the application.

### Expected Result

Debug files should be sent to Sentry during the build.

### Actual Result

No debug files are sent.

The `Sentry.targets` file uses `$(SentryPkg)` to find the path to where the nuget package was installed, so it can run the bundled Sentry CLI.  This is set automatically when a `<PackageReference>` is used, but not for the older `packages.config` style.

mattjohnsonpint:
Fixes #2223

### DIFF
--- a/src/Sentry/buildTransitive/Sentry.targets
+++ b/src/Sentry/buildTransitive/Sentry.targets
@@ -40,20 +40,31 @@
   </Target>
 
   <Target Name="PrepareSentryCLI" AfterTargets="CheckSentryCLI" Condition="'$(UseSentryCLI)' == 'true'">
+
+    <!--
+      The Sentry configuration can be set manually in MSBuild properties if desired.
+      Otherwise the default configuration will be used, as reported by "sentry-cli info".
+      The defaults can be set either via config file, or environment variables, per: https://docs.sentry.io/product/cli/configuration/
+    -->
     <PropertyGroup>
-      <!--
-        The Sentry configuration can be set manually in MSBuild properties if desired.
-        Otherwise the default configuration will be used, as reported by "sentry-cli info".
-        The defaults can be set either via config file, or environment variables, per: https://docs.sentry.io/product/cli/configuration/
-      -->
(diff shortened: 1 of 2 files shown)


src/Sentry/buildTransitive/Sentry.targets
```diff
@@ -40,20 +40,31 @@
...
       <SentryCLIOptions Condition="'$(SentryProject)' != ''">$(SentryCLIOptions) --project $(SentryProject)</SentryCLIOptions>
+    </PropertyGroup>
+
+    <!--
+      If Sentry was added via PackageReference, PkgSentry will already be set to the path of the Sentry nuget package.
+      However, older-style projects using packages.config will need us to resolve the path manually.
+    -->
+    <PropertyGroup Condition="'$(SentryCLIDirectory)' == '' And '$(PkgSentry)' == ''">
+      <PkgSentry Condition="$([System.String]::new('%(Reference.Identity)').StartsWith('Sentry,'))">$([System.IO.Directory]::GetParent(%(Reference.HintPath)).Parent.Parent.FullName)</PkgSentry>
+    </PropertyGroup>
```
mattjohnsonpint, line 63: Yeah, it's ugly - but it works.  Tested on Windows with VS 2022.

SeanFeldman (approved):

PR #2231, opened 2023-03-13
Serialize objects as JSON (with @ prefix)

### LINKED ISSUE #556: Serialize objects as JSON (with @ prefix).
Please mark the type framework used: ASP.NET Core

* Version: 2.1

Please mark the type of the runtime used: .NET Framework

* Version: 4.7.2

Please mark the NuGet packages used: Sentry.AspNetCore

* Version: 2.1.6

---

I suggest adding functionality to prefix log message parameter names with `@` to force serialization of parameters as JSON.

Sentry displays objects logged through the `ILogger` interface by calling `ToString` on the object. This works well for anonymous objects - since the compiler generates a JSON-like `ToString` override - but not for "proper classes".

Example:
```
class TestClass {
    public string TestProp { get; set; } = "Value";
}

_logger.LogError(
    "Test - {@testObj} - {@anonObj}",
    new TestClass(),
    new { SomeProp = "SomeValue" }
);
```
"Expected" result:  
`Test - { "TestProp": "Value" } - { "SomeProp": "SomeValue" }`  
Actual result:  
`Test - Namespace+TestClass - { SomeProp = SomeValue }`

Other logging libraries (like NLog and Serilog) make use of an `@` in the parameter name to force serialization of the parameter as JSON.  
https://github.com/NLog/NLog/wiki/How-to-use-structured-logging#formatting-of-the-message  
https://github.com/serilog/serilog/wiki/Structured-Data#preserving-object-structure

"Workarounds":
* Manually serialize the object as JSON when logging (e.g. `_logger.LogError("{obj}", JsonConvert.SerializeObject(obj));`)
* Create anonymous object with relevant data (e.g. `_logger.LogError("{obj}", new { Val1 = obj.SomeProp, Val2 = obj.OtherProp.SubProp });`)

(Original question on Stack Overflow: https://stackoverflow.com/questions/64317618/send-object-as-json-to-sentry-io )

SeanFeldman:
Fixes #556

### DIFF
--- a/src/Sentry.Extensions.Logging/SentryLogger.cs
+++ b/src/Sentry.Extensions.Logging/SentryLogger.cs
@@ -6,21 +6,23 @@ namespace Sentry.Extensions.Logging;
 internal sealed class SentryLogger : ILogger
 {
     private readonly IHub _hub;
+    private readonly SentryLoggerFormatter _sentryLoggerFormatter;
     private readonly ISystemClock _clock;
     private readonly SentryLoggingOptions _options;
 
     internal string CategoryName { get; }
 
-    internal SentryLogger(
-        string categoryName,
+    internal SentryLogger(string categoryName,
         SentryLoggingOptions options,
         ISystemClock clock,
-        IHub hub)
+        IHub hub,
+        SentryLoggerFormatter sentryLoggerFormatter)
     {
         CategoryName = categoryName;
         _options = options;
         _clock = clock;
         _hub = hub;
+        _sentryLoggerFormatter = sentryLoggerFormatter;
     }
 
     public IDisposable BeginScope<TState>(TState state) => _hub.PushScope(state);
@@ -43,10 +45,9 @@ internal sealed class SentryLogger : ILogger
             return;
         }
 
-        // TBD: If Sentry options contains an override for the formatter, use it instead of the default passed in
-        //var message = formatter?.Invoke(state, exception);
-        var sentryFormatter = new SentryLoggerFormatter();
-        var message = sentryFormatter.Invoke(state);
+        var message = _options.SupportObjectDestructuring
+            ? _sentryLoggerFormatter.Invoke(state)
+            : formatter?.Invoke(state, exception);
 
         if (ShouldCaptureEvent(logLevel, eventId, exception))
         {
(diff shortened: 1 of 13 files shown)


SeanFeldman:
@mattjohnsonpint, this is a quick spike to see how log statements with parameters that are objects, primitives, and anonymous types would get logged. If this approach is alright, I'll need to complete the following:

1. Add Sentry logging option extension to enable (opt-in) object serialization. I.e. using a Sentry formatter.
2. Update the logging provider to potentially inject Sentry formatted upon logger creation rather than at every log write invocation.
3. Add tests.
4. Review the implementation for performance issues.

src/Sentry.Extensions.Logging/SentryLoggerFormatter.cs
```diff
@@ -0,0 +1,54 @@
...
+
+internal class SentryLoggerFormatter
+{
+    public static readonly SentryLoggerFormatter Instance = new();
+
+    private SentryLoggerFormatter() {}
+
+    public string Invoke<TState>(TState state)
+    {
+        var dictionary = (state as IEnumerable<KeyValuePair<string, object>>)?.ToDictionary(i => i.Key, i => i.Value) ?? new Dictionary<string, object>();
```
mattjohnsonpint, line 11: These should be `<string, object>`

mattjohnsonpint: I don't believe we can guarantee that `TState` will be castable to `IEnumerable<KeyValuePair<string, object?>>` in all cases.  The contract for `TState` doesn't constrain it.  In theory, any caller to `ILogger.Log` could pass a different type of data, as long as the formatter they supply can handle it.

src/Sentry.Extensions.Logging/SentryLogger.cs
```diff
@@ -43,7 +45,9 @@ public bool IsEnabled(LogLevel logLevel)
             return;
         }
 
-        var message = formatter?.Invoke(state, exception);
+        var message = _options.SupportObjectDestructuring
+            ? _sentryLoggerFormatter.Invoke(state)
+            : formatter?.Invoke(state, exception);
```
mattjohnsonpint, line 50: Per our discussion, we can avoid having a custom formatter by creating new `state` instead, then call the original formatter.

SeanFeldman:
Adding support for this feature is not straightforward, given the missing support in the .NET logging API.
Refer to the original issue for more details and the future of this feature.

PR #2234, opened 2023-03-14
fix: normalize frame in-app resolution for modules & function prefixes

vaind:
Originally, I was concerned about the in-app resolution [implementation ](https://github.com/getsentry/sentry-dotnet/blob/c8ad13d7a96c85726fec02cc2ef9dec4430aab4a/src/Sentry/SentryStackFrame.cs#L197-L198) not using binary prefix search especially since it would be used in profiling where there are easily thousands of frames. I've added a micro-benchmark to measure the current implementation but it looks good enough to me at the moment. 

Then, I noticed some frames in profiling are recognized in-app even though they're coming from Sentry so I've added test cases to cover that and updated the resolution code.

Forgive me for mixing up concerns in this PR, I just didn't want to add it to the Profiling one to make it even less digestable.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
+
+### Fixes
+
+- Normalize StackFrame in-app resolution for modules & function prefixes ([#2234](https://github.com/getsentry/sentry-dotnet/pull/2234))
+
 ## 3.29.1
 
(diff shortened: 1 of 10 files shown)


mattjohnsonpint:
LGTM.  Please update the tests and changelog to fix check failures.

mattjohnsonpint (approved):

PR #2254, opened 2023-03-26
Update polyfills

mattjohnsonpint:
Switch to https://github.com/SimonCropp/Polyfill.  Adjust where needed.  Fix a few related things.

Thanks @SimonCropp

#skip-changelog


SeanFeldman (approved):

SimonCropp:
r there any common apis i should be adding polyfills for?

mattjohnsonpint:
@SimonCropp - Depends on how crazy you want to get with this. 😆 

Your library allowed me to remove `IsExternalInit` and `ModuleInitializerAttributes` which we had in separate files.  It also allowed me to remove `string.Contains`, `Stream.CopyToAsync`, and `Deconstruct<TKey, TValue>` from our `Polyfills.cs` file. 

Feel free to migrate any of the others from:
https://github.com/getsentry/sentry-dotnet/blob/main/src/Sentry/Internal/Polyfills.cs

The only downside to this source-only library approach is that any change is potentially a breaking change, but we can live with that.

SimonCropp:
> The only downside to this source-only library approach is that any change is potentially a breaking change

why is that?

SimonCropp:
> Well, for example, say you decided to add one or more of the polyfills we already have. When we update to that version, we'd get a compile error because a class or method has already been implemented, or a warning because an extension method call is ambiguous.

Ahh yes. sorry that is obvious.

PR #2259, opened 2023-03-28
Fix Sentry CLI arguments when using custom URL or auth token parameters

mattjohnsonpint:
Authentication and URL parameters must be passed to Sentry CLI _before_ the specific command.

I discovered this while testing against a local Sentry instance for development purposes.  However, it will also be broken for Self-Hosted Sentry, if passing the URL to Sentry via our `<SentryUrl>` MSBuild property.


SeanFeldman (approved):

PMExtra:
😭 It bothered me so a long time.

PR #2284, opened 2023-04-03
Add Scope.Clear and Scope.ClearBreadcrumbs methods

### LINKED ISSUE #2272: Add `Scope.Clear` and `Scope.ClearBreadcrumbs`
### Problem Statement

We need a way to clear the current scope entirely.

We also need a way to just clear breadcrumbs from the current scope.

See #1660 for use case examples.

### Solution Brainstorm

As per https://develop.sentry.dev/sdk/unified-api/#scope, we can add `Clear` and `ClearBreadcrumbs` methods to the existing `Scope` object.

jamescrosswell:
Resolves #2272

### DIFF
--- a/src/Sentry/Scope.cs
+++ b/src/Sentry/Scope.cs
@@ -312,7 +312,7 @@ public class Scope : IEventLike, IHasDistribution
     /// <summary>
     /// Resets all the properties and collections within the scope to their default values. 
     /// </summary>
-    internal void Clear()
+    public void Clear()
     {
         Level = default;
         Request = new();
@@ -343,8 +343,10 @@ public class Scope : IEventLike, IHasDistribution
 #endif
     }
 
-    /// <inheritdoc />
-    internal void ClearBreadcrumbs()
+    /// <summary>
+    /// Removes all Breadcrumbs from the scope.
+    /// </summary>
+    public void ClearBreadcrumbs()
     {
 #if NETSTANDARD2_0 || NETFRAMEWORK
         // No Clear method on ConcurrentQueue for these target frameworks
--- a/src/Sentry/Scope.cs
+++ b/src/Sentry/Scope.cs
@@ -195,7 +195,11 @@ public class Scope : IEventLike, IHasDistribution
     /// <inheritdoc />
     public IReadOnlyList<string> Fingerprint { get; set; } = Array.Empty<string>();
 
+#if NETSTANDARD2_0 || NETFRAMEWORK
+    private ConcurrentQueue<Breadcrumb> _breadcrumbs = new(); 
+#else
     private readonly ConcurrentQueue<Breadcrumb> _breadcrumbs = new();
+#endif
 
     /// <inheritdoc />
     public IReadOnlyCollection<Breadcrumb> Breadcrumbs => _breadcrumbs;
@@ -306,6 +310,28 @@ public class Scope : IEventLike, IHasDistribution
     public void AddAttachment(Attachment attachment) => _attachments.Add(attachment);
 
     /// <summary>
+    /// Resets all the properties and collections within the scope to their default values. 
+    /// </summary>
+    internal void Clear()
+    {
+        Level = default;
+        Request = new();
+        Contexts.Clear();
+        User = new();
+        Platform = default;
+        Release = default;
+        Distribution = default;
+        Environment = default;
+        TransactionName = default;
+        Transaction = default;
+        Fingerprint = Array.Empty<string>();
+        ClearBreadcrumbs();
+        _extra.Clear();
+        _tags.Clear();
+        ClearAttachments();
+    }
+
+    /// <summary>
     /// Clear all Attachments.
     /// </summary>
     public void ClearAttachments()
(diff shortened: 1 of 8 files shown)


test/Sentry.Tests/ScopeTests.cs
```diff
@@ -232,6 +233,51 @@ public void AddAttachment_AddAttachments()
         scope.Attachments.Should().Contain(attachment2, "Attachment2 was not found.");
     }
 
+    [Fact]
+    public void Clear_SetsPropertiesToDefaultValues()
+    {
+        // Arrange
```
jamescrosswell, line 239: Could use something like Bogus/AutoBogus here to generate fakes.

mattjohnsonpint:
Thanks for the PR!  I'll review soon and get back to you. 🙂

src/Sentry/Scope.cs
```diff
@@ -305,6 +309,28 @@ public void UnsetTag(string key)
     /// </summary>
     public void AddAttachment(Attachment attachment) => _attachments.Add(attachment);
 
+    /// <summary>
+    /// Resets all the properties and collections within the scope to their default values. 
+    /// </summary>
+    internal void Clear()
```
mattjohnsonpint, line 315:
```suggestion
    public void Clear()
```

src/Sentry/Scope.cs
```diff
@@ -317,6 +343,18 @@ public void ClearAttachments()
 #endif
     }
 
+    /// <inheritdoc />
+    internal void ClearBreadcrumbs()
```
mattjohnsonpint, line 347:
```suggestion
    public void ClearBreadcrumbs()
```

src/Sentry/Scope.cs
```diff
@@ -317,6 +343,18 @@ public void ClearAttachments()
 #endif
     }
 
+    /// <inheritdoc />
```
mattjohnsonpint, line 346: We'll need a real summary here instead of `inheritdoc`.

test/Sentry.Tests/ScopeTests.cs
mattjohnsonpint: Neat idea! But let's think about that as a separate future endeavor. 😃

test/Sentry.Tests/ScopeTests.cs
```diff
@@ -232,6 +233,51 @@ public void AddAttachment_AddAttachments()
...
+        _sut.SetExtra("extra", "extra to be cleared");
+        _sut.SetTag("tag", "tag to be cleared");
+        _sut.AddAttachment(new Attachment(default, default, default, "attachment to be cleared"));
+
+        // Act
+        _sut.Clear();
+
+        // Assert
+        var newScope = new Scope();
+        using (new AssertionScope())
```
mattjohnsonpint, line 261: Nice use of `AssertionScope`. 👍

test/Sentry.Tests/ScopeTests.cs
mattjohnsonpint, line 1: Let's add one more test:
- Set some values on the scope
- Push a new scope
- Ensure the values are there
- Clear the scope
- Ensure the values are gone
- Pop the scope
- Ensure the value have returned

mattjohnsonpint (changes requested):
Looks pretty good overall!  A few things to address as indicated inline.

Also:
- After you switch the methods public, re-run tests locally and you'll see the public API snapshots have been updated.  Commit those.  (You can use `dotnet test test/Sentry.Tests --filter ApiApprovalTests` if you want to do it fast.  Or just run all tests.)
- Add an entry to `CHANGELOG.md`, pointing back at this PR.

Thanks!

src/Sentry/Scope.cs
jamescrosswell: Added a new commit to address this and the missing docs on the ClearBreadcrumbs method

test/Sentry.Tests/ScopeTests.cs
jamescrosswell: Hmm... Interesting. It feels like that would either be an integration test or one for SentryScopeManagerTests. Do you think there's a risk that the implementation of Scope.Clear or Scope.ClearBreadcrumbs might affect how scopes get pushed/popped to/from the stack? 

I probably need to do a bit more reading to understand how pushing/popping scopes in the Hub/ScopeManager are related here. I did find the [concurrency  docs](https://develop.sentry.dev/sdk/unified-api#concurrency) docs and also took a peek at the implementation of SentryScopeManager.Push, but haven't had any lightbulb moments yet. 

One thing I did glean from that source code is that there may have been an easier way to implement the Clean method by using something like `(new Scope()).Apply(this)`. That would avoid having very similar code in both the Apply and the Clean methods as well, so is maybe a better solution for the implementation?

mattjohnsonpint: Mainly I just want to ensure that clearing the current scope doesn't affect the parent scope.  Any other way you can think to do that would be fine.  Thanks!

jamescrosswell: Added a new unit test for this in the most recent commit. 

To avoid copying/pasting code, I ended up refactoring out some of the original test into a couple of extension methods. Be good to check if you're OK with how I've done that as well.

mattjohnsonpint (approved):

mattjohnsonpint:
Nice job.  Thanks!!!

PR #2287, opened 2023-04-04
Rework handling of `AggregateException` to align with Sentry Exception Groups

### LINKED ISSUE #2240: Rework handling of `AggregateException`
### Problem Statement

In #1672, we implemented logic for dealing with `AggregateException` by flattening the tree of inner exceptions and their children into a single list of exceptions.  At the same time, we added an option `KeepAggregateException` that controls whether the `AggregateException` objects would remain in that  list or be stripped out.  The default is `false`, meaning to strip them out.

Since then, we also ran in to several bugs caused by this approach:
- #1916
  - fixed by #1924
- #1917
  - fixed by #1922
- #2065
  - fixed by #2067

Despite those fixes, we still have several problems with these workarounds not fitting the Sentry design - which is that multiple exceptions represent a single chain of events.

Consider this example:

- `AggregateException`
  - `A`
    - `1`
  - `B`
    - `2`

Currently, we'd flatten that in to one big list and remove the aggregate (by default).  Sentry reverses the ordering, and presently we sort the inner exceptions in ascending order, so the list sent would be `[1,A,2,B]`.  An issue would be titled with the details of `B`, showing `2` as the cause of `B` (correct), then `A` as the cause of `2` (incorrect), then `1` as the cause of `A` (correct).

Besides the incorrect ordering, this also has other more serious consequences:
- An issue will not get created for `A`.  If it happened to be a different exception type than `B`, it could go unnoticed.
- The issue grouping rules will take the whole chain of exceptions into account.  Thus if there's another error raised independently for `B`, it wouldn't be grouped with the instance of `B` that was part of the aggregate.
- Similarly, the *quantity* of exceptions can affect the grouping.  Say one has code that performs parallel tasks, but might have a different number of tasks each time.  `[A,A]` would be grouped separately from `[A,A,A]` or `[A,A,A,A,A,A]` (etc.) - each issue having an identical title and cause.  This also messes up any alerting rules, and can resurrect issues that were intentionally ignored.
- Within the SDK itself, we have issues with processing aggregate exceptions:
  - Exception filters will only filter out an `AggregateException if *all* of the inner exceptions were supposed to be filtered.
  - Exception processors, event processors, and `BeforeSend` all present interfaces that require special handling of `AggregateException` in each implementation.
  - When we strip away an aggregate exception, if there was no stack trace on the inner exception then we copy the one from the aggregate.  This is usually appropriate - so that one can have some sense of where to start debugging.  However, the stack trace now no longer applies to the exception being viewed - so it's a bit disingenuous.

Ultimately, the applied workaround is not very robust.  It works best when there's only one inner exception in the aggregate, but has problems when there are multiple.

Without the workarounds, the above example would be sent as `[1,A,AggregateException]`. All issues with an aggregate would be titled the same, and we'd lose any information about `B`.  So this is also not a good plan.

### Solution Brainstorm

The latest version of Python (3.11) had a similar feature added called `ExceptionGroup`.  There's also `AggregateError` in JavaScript, and several others.  Together these have increased the demand for a properly designed solution.

We're tracking the new design in https://github.com/getsentry/sentry/issues/37716.  When ready, we should undo the previous .NET workarounds and implement the new unified design. Please add feedback about the new design over there.

We'll leave this issue open to track the work in .NET, and to comment on any other issues specifically related to `AggregateException` or the workarounds we've currently implemented.

mattjohnsonpint:
This PR updates the .NET SDK to align with the new "exception groups" feature of Sentry.  See #2240, and [Sentry RFC 79](https://github.com/getsentry/rfcs/blob/main/text/0079-exception-groups.md) for details.

Aligns with full changes listed at https://github.com/getsentry/sentry/issues/37716#status-update

This changes how `AggregateException` is handled.  Instead of filtering them out client-side, the SDK marks them as an "exception group",  and adds includes data that represents the hierarchical structure of inner exceptions. Sentry now recognizes this server-side, improving the accuracy of the issue detail page.

Accordingly, the `KeepAggregateException` option is now obsolete and does nothing.  Please remove any usages of `KeepAggregateException`.

Closes #2240.


mattjohnsonpint:
Note: I'll rebase this on main after #2294 is merged.  Then it will be a bit clearer on the exact changes needed for exception groups.  (Currently it contains both.)

mattjohnsonpint:
NOTE: This PR is ready for review but should not be merged until the corresponding back-end changes to Sentry are complete (per [RFC 79](https://github.com/getsentry/rfcs/blob/main/text/0079-exception-groups.md)).

src/Sentry/Internal/Extensions/MiscExtensions.cs
```diff
@@ -82,4 +82,19 @@ public static void CancelAfterSafe(this CancellationTokenSource cts, TimeSpan ti
         TKey key,
         TValue value) =>
         collection.Add(new KeyValuePair<TKey, TValue>(key, value));
+
+    internal static string GetRawMessage(this AggregateException exception)
```
SeanFeldman, line 86: Are there any concerns with this assuming there's only one level nested exception?

For example, if an exception `B` nested in exception `A` and `A` is used to throw an `AggregateException`, the returned result will be neither `B`'s nor `A`'s message, but rather `"One or more errors occurred."`

src/Sentry/Internal/MainExceptionProcessor.cs
```diff
@@ -30,6 +30,68 @@ public void Process(Exception exception, SentryEvent sentryEvent)
...
+            {
+                // No need to convey an empty mechanism.
+                exceptions[0].Mechanism = null;
+            }
+        }
+
+        return exceptions;
+    }
+
+    private class Counter
```
SeanFeldman, line 54: Could `Counter` be a `struct`?

src/Sentry/Internal/MainExceptionProcessor.cs
```diff
@@ -30,6 +30,68 @@ public void Process(Exception exception, SentryEvent sentryEvent)
...
+    {
+        private int _value;
+
+        public int GetNextValue() => _value++;
```
SeanFeldman, line 58: I got confused by this. The counter starts at zero. So when the first time `GetNextValue()` is invoked, should it be `1`? But it will return 0. I'm assuming this is by design but is worth asking 😊

src/Sentry/SentryOptions.cs
```diff
@@ -776,10 +776,14 @@ public StackTraceMode StackTraceMode
...
-    /// The default behaviour is to only log <see cref="AggregateException.InnerExceptions"/> and not include the root <see cref="AggregateException"/>.
-    /// Set KeepAggregateException to true to include the root <see cref="AggregateException"/>.
+    /// This property is no longer used.  It will be removed in a future version.
     /// </summary>
+    /// <remarks>
+    /// All exceptions are now sent to Sentry, including <see cref="AggregateException"/>s.
+    /// The issue grouping rules in Sentry have been updated to accomodate "exception groups",
+    /// such as <see cref="AggregateException"/> in .NET.
+    /// </remarks>
+    [Obsolete("This property is no longer used.  It will be removed in a future version.")]
```
SeanFeldman, line 786: To we have an obsoletion policy? Like "remove in major versions only". Would it be obsoleted with an error first, then removed, or removed?

SeanFeldman (approved):

src/Sentry/Internal/MainExceptionProcessor.cs
mattjohnsonpint: Intentionally needs to be a class, so the same instance is passed by reference throughout all recursice calls.

src/Sentry/Internal/MainExceptionProcessor.cs
mattjohnsonpint: By design. It's a postfix, so the value returned is before the increment.

https://learn.microsoft.com/dotnet/csharp/language-reference/operators/arithmetic-operators#increment-operator-

Ultimately this class is just an easy way to assign an ID to each item, starting with 0.

src/Sentry/SentryOptions.cs
mattjohnsonpint: Obsoleted like this, then left there until next major version (4.0.0).

If there were an equavlience, we'd connect it. But in this case there isn't any.

src/Sentry/Internal/Extensions/MiscExtensions.cs
mattjohnsonpint: "One or more errors occurred." is actually what we're after here.  Rather than "One or more errors occurred. (error 1) (error 2)"

Only reason not to take just the default message is that technically it can be customized.  Even then, we'd just want the custom message, before the inner exceptions messages were appended.

SeanFeldman: In that case please ignore my comment 🙂

jamescrosswell (approved):

mattjohnsonpint:
This can be shipped as soon as https://github.com/getsentry/sentry/pull/48653 is completed and released to production.  

Also, for the best experience with exception groups, self-hosted Sentry users should take the corresponding update when it is available.

PR #2294, opened 2023-04-07
Improvements to exception mechanism data

### LINKED ISSUE #1480: Add `handled:true` when an exception has been handled
The SDK captures unhandled exceptions by flagging those to `handled: false`. The opposite isn't the case. Simply no option is sent.
The issue is when customers are trying to create alerts. Or query by "Handled: true". To help with this, the SDK should always include the Handled property, and default it to 'true'

### LINKED ISSUE #2288: Exception mechanism type unset
### Package

Sentry

### SDK Version

3.29.1

### Steps to Reproduce

```csharp
using Sentry;

using var _ = SentrySdk.Init(o =>
{
    o.Dsn = "...";
    o.Debug = true;
});

SentrySdk.CaptureException(new Exception("test"));
```

### Expected Result

Per https://develop.sentry.dev/sdk/event-payloads/exception/#exception-mechanism:

> The exception mechanism is an optional field ...

and also

> The `type` attribute is required to send any exception mechanism attribute, even if the SDK cannot determine the specific mechanism. In this case, set the type to `generic`.

Thus, I either the `event.mechanism` should not exist on the event, or the `event.mechanism.type` field should be populated.

### Actual Result

The mechanism exists, but it is an empty object.

```json
"exception": {
  "values": [
    {
      "type": "System.Exception",
      "value": "test",
      "module": "System.Private.CoreLib, Version=6.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e",
      "thread_id": 1,
      "mechanism": {}
    }
  ]
},
```

### LINKED ISSUE #2289: Move exception data to mechanism
### Problem Statement

Consider:

```csharp
SentrySdk.CaptureException(new Exception("test")
{
    Data =
    {
        ["foo"] = 123,
        ["bar"] = 456
    }
});
```

The data gets put in the "Additional Details" section on the issue details page:

image

It should instead be rendered with the exception, such as:

image

### Solution Brainstorm

During processing, extra data added to an exception's `Data` property is moved to the `extra` section of the Sentry event.

https://github.com/getsentry/sentry-dotnet/blob/8093e26ecbd805d32a59b5708ef6772056b3d791/src/Sentry/Internal/MainExceptionProcessor.cs#L32-L35

It should instead be moved to the exception's `mechanism.data` collection.  Then it will be rendered as shown above.

### LINKED ISSUE #2290: Unused mechanism fields
### Problem Statement

Reference: https://develop.sentry.dev/sdk/event-payloads/types/#mechanism

Compare to our [`Mechanism`](https://github.com/getsentry/sentry-dotnet/blob/8093e26ecbd805d32a59b5708ef6772056b3d791/src/Sentry/Protocol/Mechanism.cs) class:

- `Description` is defined but unused
- `Data` is defined but unused
- `Meta` is defined but unused
- `Synthetic` is not defined

### Solution Brainstorm

- `Description` can be populated with a short description of the integration that caught it.  This will help users.
- `Data` can be used for the `Exception.Data` (see #2289)
- `Meta` we can't really use it presently.
  -  We _could_ consider updating Sentry to use this for `Exception.HResult` and/or `Win32Exception.NativeErrorCode`, but it would need a new meta key.  None of the current ones are compatible.
- `Synthetic` should be added to our class.

Even if unused, both `Meta` and `Synthetic` should be defined for purposes of serialization compatibility with native SDKs.

mattjohnsonpint:
This PR address several items related to the [`exception.mechanism`](https://develop.sentry.dev/sdk/event-payloads/exception/#exception-mechanism) data sent to Sentry.

- Prevents an empty mechanism.  Either `exception.mechanism` is omitted, or the `exception.mechanism.type` field will be set.  Defaults the mechanism type to `"generic"` when not set by another method.
  Fixes #2288.
- Moves extra data added to `System.Exception.Data` to the `exception.mechanism.data` collection, instead of placing it in `exception.extra`.  This places it near the exception in the issue details page, and renders it correctly.
  Fixes #2289.
- Adds descriptions to exception mechanisms for built-in integrations, and a new `SetSentryMechanism` extension method for convenience. Also adds a property for the `synthetic` field.  Omits it unless true.  The default is false.
  Fixes #2290.
- Sets `exception.mechanism.handled` to `true` in the case of an exception that was thrown, caught, and manually captured.  As an example:

  ```csharp
  try
  {
      // ... something that throws an exception
  }
  catch (Exception ex)
  {
      // this will now be correctly marked as handled:true
      SentrySdk.CaptureException(ex);
  }
  ```

  Fixes #1480.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,8 @@
 
 - Add `FileDiagnosticLogger` to assist with debugging the SDK ([#2242](https://github.com/getsentry/sentry-dotnet/pull/2242))
 - Attach stack trace when events have captured an exception without a stack trace ([#2266](https://github.com/getsentry/sentry-dotnet/pull/2266))
-- Add Scope.Clear and Scope.ClearBreadcrumbs methods ([#2284](https://github.com/getsentry/sentry-dotnet/pull/2284))
+- Add `Scope.Clear` and `Scope.ClearBreadcrumbs` methods ([#2284](https://github.com/getsentry/sentry-dotnet/pull/2284))
+- Improvements to exception mechanism data ([#2294](https://github.com/getsentry/sentry-dotnet/pull/2294))
 
 ### Fixes
 
--- a/src/Sentry.AspNet/SentryHttpServerUtilityExtensions.cs
+++ b/src/Sentry.AspNet/SentryHttpServerUtilityExtensions.cs
@@ -21,8 +21,12 @@ public static class SentryHttpServerUtilityExtensions
     {
         if (server.GetLastError() is { } exception)
         {
-            exception.Data[Mechanism.HandledKey] = false;
-            exception.Data[Mechanism.MechanismKey] = "HttpApplication.Application_Error";
+            exception.SetSentryMechanism(
+                "HttpApplication.Application_Error",
+                "This exception was caught by the ASP.NET global error handler. " +
+                "The web server likely returned a 5xx error code as a result of this exception.",
+                handled: false);
+
             return hub.CaptureException(exception);
         }
         return SentryId.Empty;
--- a/src/Sentry.AspNetCore/SentryMiddleware.cs
+++ b/src/Sentry.AspNetCore/SentryMiddleware.cs
@@ -120,8 +120,13 @@ internal class SentryMiddleware : IMiddleware
                 var exceptionFeature = context.Features.Get<IExceptionHandlerFeature?>();
                 if (exceptionFeature?.Error != null)
                 {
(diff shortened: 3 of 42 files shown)


mattjohnsonpint:
Note, many of these items are prerequisites for #2240 - coming soon.

mattjohnsonpint:
Here are some screenshots how this looks in Sentry:

Caught and handled exception:

image

Uncaught exception (hovering over the mechanism for description popup):

image

Captured exception that was never thrown:

image

PR #1965, opened 2022-10-05
Fix issues with envelope deserialization

mattjohnsonpint:
Improves deserialization by reading the envelope header line in a buffer instead of streaming one character at a time, which should hopefully fix #1963.

Also makes a few other optimizations in nearby code for both serialization and deserialization.

No new tests because we already have a lot of tests that validate it works correctly.

PR #2297, opened 2023-04-12
Buffer payloads asynchronously when appropriate

mattjohnsonpint:
During serialization, we buffer payloads of `EnvelopeItems` into memory so we can calculate lengths to include in headers.

As part of #1965 (released in 3.22.0), I switched to synchronous buffering of these payloads - because serialization of a fully materialized object to a `MemoryStream` gains no advantage by being async, and doing so adds some minor overhead.

That is still true, however I didn't consider that not all payloads are fully materialized.  In particular, a `StreamSerializable` could be representing an attachment, in which case it would be reading from a `FileStream` or any other type of `Stream` (depending on how it was attached).  Also, we may need other types of async materialization in the future.

Thus, we should defer to synchronous buffering only when we _know_ the object is fully materialized - which is what our `JsonSerializable` class represents.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
+
+### Fixes
+
+- Buffer payloads asynchronously when appropriate ([#2297](https://github.com/getsentry/sentry-dotnet/pull/2297))
+
 ## 3.30.0
 
 ### Features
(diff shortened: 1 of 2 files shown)


SeanFeldman (approved):

SeanFeldman:
@mattjohnsonpint, it looks like there's a lot of work taking place with `MemoryStream`, and this is a hot path. Would there be any benefit in taking advantage of [`RecyclableMemoryStream`](https://github.com/microsoft/Microsoft.IO.RecyclableMemoryStream)?

vaind (approved):

mattjohnsonpint:
@SeanFeldman - `RecyclableMemoryStream` looks useful indeed.  Created #2299 to track.  Thanks!

Issue #1050, opened 2021-06-08
Exceptions not captured since updating to 3.4.0 or above


Camble:
### Environment

How do you use Sentry?
Sentry SaaS

Which SDK and version?
Sentry.AspNetCore 3.4.0 & 3.5.0

### Steps to Reproduce

Installed & configured Sentry SDK 3.5.0 on a new ASP.NET Core 3.1 Web API. 
Using webBuilder.UseSentry() in Program.cs
Configuration in appsettings.json, but also tried hardcoding values.

### Expected Result

Test exception to be captured

### Actual Result

No exceptions captured.
Rolled back to known working version (3.3.4) and this resolved the issue.

Tyrrrz:
Hi! Where was the exception thrown? Inside an endpoint or somewhere else?

bruno-garcia:
@Camble Running the sample in the repo works fine:
![image](https://user-images.githubusercontent.com/1633368/121392391-57d8e980-c91d-11eb-94c0-aa5d1ce4e834.png)


Could you please share a repro so we can investigate? If you set the SDK in Debug mode, anything shows on the log to indicate the problem?

Camble:
@Tyrrrz Inside an endpoint, but I tried capturing a test message from various places using the SentrySdk static class.

@bruno-garcia I'm unable to share the exact repo, but I can try to reproduce in a new project. I'll investigate further. Thanks.

bruno-garcia:
Thanks @Camble . With a repro in hand we can debug it to find out what's going on. But the tests we did with our samples, we couldn't reproduce it.

jouni-kantola:
Same problem here; here's a repro:
[SentryUnhandledExceptionsNotCapturedRepro.zip](https://github.com/getsentry/sentry-dotnet/files/6657457/SentryUnhandledExceptionsNotCapturedRepro.zip)

The unhandled exception is not captured when the following project config is used:
```xml
<Project Sdk="Microsoft.NET.Sdk.Web">
  <PropertyGroup>
    <TargetFramework>netcoreapp3.1</TargetFramework>
  </PropertyGroup>

  <ItemGroup>
    <PackageReference Include="Sentry.AspNetCore" Version="3.5.0" />
  </ItemGroup>
</Project>
```

The unhandled exception is captured with either of these combos:
- `Sentry.AspNetCore@3.3.4` works with `netcoreapp3.1`
- `Sentry.AspNetCore@3.5.0` works with `net5.0`

[...]

PR #2302, opened 2023-04-16
Restore `System.Reflection.Metadata` dependency for .NET Core 3

mattjohnsonpint:
We accidentally removed the `System.Reflection.Metadata` dependency for .NET Core 3 targets in 3.30.0.  This restores it.

See https://github.com/getsentry/sentry-dotnet/issues/1050#issuecomment-1510505001

Thanks @SirSalamandra

### DIFF
--- a/src/Sentry/Sentry.csproj
+++ b/src/Sentry/Sentry.csproj
@@ -29,7 +29,9 @@
       <Link>%(RecursiveDir)\%(Filename)%(Extension)</Link>
     </Compile>
   </ItemGroup>
-  <ItemGroup Condition="$(TargetFramework.StartsWith('netstandard')) or $(TargetFramework.StartsWith('net4'))">
+
+  <!-- Ben.Demystifier also needs System.Reflection.Metadata 5.0.0 or higher on all platforms. -->
+  <ItemGroup Condition="$(TargetFramework.StartsWith('netstandard')) or $(TargetFramework.StartsWith('net4')) or $(TargetFramework.StartsWith('netcoreapp'))">
     <PackageReference Include="System.Reflection.Metadata" Version="5.0.0" />
   </ItemGroup>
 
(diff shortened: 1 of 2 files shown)


src/Sentry/Sentry.csproj
```diff
@@ -29,7 +29,9 @@
       <Link>%(RecursiveDir)\%(Filename)%(Extension)</Link>
     </Compile>
   </ItemGroup>
-  <ItemGroup Condition="$(TargetFramework.StartsWith('netstandard')) or $(TargetFramework.StartsWith('net4'))">
+
+  <!-- Ben.Demystifier also needs System.Reflection.Metadata 5.0.0 or higher on all platforms. -->
+  <ItemGroup Condition="$(TargetFramework.StartsWith('netstandard')) or $(TargetFramework.StartsWith('net4')) or $(TargetFramework.StartsWith('netcoreapp'))">
     <PackageReference Include="System.Reflection.Metadata" Version="5.0.0" />
```
SeanFeldman, line 35: Just a suggestion 🙂
One of the challenges with standalone comments is that those are not linked/associated with the package reference. An accidental re-arrangement in the CSPROJ can sever the connection. `PackageReference` has a `Label` property that could be used to keep the connection between the comment and the package.

```suggestion
  <ItemGroup Condition="$(TargetFramework.StartsWith('netstandard')) or $(TargetFramework.StartsWith('net4')) or $(TargetFramework.StartsWith('netcoreapp'))">
    <PackageReference Include="System.Reflection.Metadata" Version="5.0.0" Label="Ben.Demystifier also needs System.Reflection.Metadata 5.0.0 or higher on all platforms." />
```

SeanFeldman (approved):

src/Sentry/Sentry.csproj
mattjohnsonpint: Thanks, but I believe `Label` isn't appropriate for lengthy comments.  It becomes metadata on the item that can be used by other `Condition` expressions.  The MS docs are poor on this also - they've documented `ItemGroup.Label`, but not `Item.Label`.  Either way, examples I've seen would use this for shorter strings like `Label="Something"`, rather than comments.

PR #2303, opened 2023-04-16
Improve `WithScope` and add `WithScopeAsync`

### LINKED ISSUE #2300: Enable WithScope and WithScopeAsync
### Problem Statement

In #675, we added overloads to `Capture` methods to take an `Action<Scope>` parameter.  This was good, and should remain.  This is the preferred method to capture using one-off scope-based configuration.

At the same time, we decided to obsolete `SentrySdk.WithScope`.  IMHO, this wasn't necessary, and should be undone.

Justification:

- `WithScope` is still useful when many things are going to be done within a single new scope, such as adding breadcrumbs, starting a transaction, performing work that may need to capture multiple events, etc.
- Comments on the original issue indicated some friction with the obsoletion.  See https://github.com/getsentry/sentry-dotnet/issues/675#issuecomment-1123805220
- `WithScope` would be easier than `PushScope` + `ConfigureScope` for #2274 
- `ISentryScopeManager.WithScope` wasn't obsoleted, so `hub.WithScope` still works.  Generally, anything that works on an `IHub` instance should also be available statically on `SentrtySdk`.

### Solution Brainstorm

Remove `Obsolete` attribute from `SentrySdk.WithScope`.  Add some comments to suggest using `Capture` overloads in the original use case of only needing to configure scope for a single event.

Also, we should expand this to include `WithScopeAsync`.  Overloads should also exist for returning values (ie: `T WithScope<T>` and `Task<T> WithScopeAsync<T>`).

mattjohnsonpint:
- Un-obsoletes `WithScope`
- Adds `WithScope<T>`, `WithScopeAsync`, and `WithScopeAsync<T>`

Resolves #2300

### DIFF
--- a/src/Sentry/Internal/SentryScopeManager.cs
+++ b/src/Sentry/Internal/SentryScopeManager.cs
@@ -115,7 +115,7 @@ internal sealed class SentryScopeManager : IInternalScopeManager
         using (PushScope())
         {
             var scope = GetCurrent();
-            await scopeCallback.Invoke(scope.Key);
+            await scopeCallback.Invoke(scope.Key).ConfigureAwait(false);
         }
     }
 
@@ -124,7 +124,7 @@ internal sealed class SentryScopeManager : IInternalScopeManager
         using (PushScope())
         {
             var scope = GetCurrent();
-            return await scopeCallback.Invoke(scope.Key);
+            return await scopeCallback.Invoke(scope.Key).ConfigureAwait(false);
         }
     }
 
--- a/src/Sentry/Internal/SentryScopeManager.cs
+++ b/src/Sentry/Internal/SentryScopeManager.cs
@@ -101,6 +101,33 @@ internal sealed class SentryScopeManager : IInternalScopeManager
         }
     }
 
+    public T? WithScope<T>(Func<Scope, T?> scopeCallback)
+    {
+        using (PushScope())
+        {
+            var scope = GetCurrent();
+            return scopeCallback.Invoke(scope.Key);
+        }
+    }
+
+    public async Task WithScopeAsync(Func<Scope, Task> scopeCallback)
+    {
+        using (PushScope())
+        {
+            var scope = GetCurrent();
+            await scopeCallback.Invoke(scope.Key);
+        }
+    }
+
+    public async Task<T?> WithScopeAsync<T>(Func<Scope, Task<T?>> scopeCallback)
+    {
+        using (PushScope())
+        {
+            var scope = GetCurrent();
+            return await scopeCallback.Invoke(scope.Key);
+        }
+    }
+
     public void BindClient(ISentryClient? client)
     {
         _options.LogDebug("Binding a new client to the current scope.");
--- a/src/Sentry/SentrySdk.cs
+++ b/src/Sentry/SentrySdk.cs
@@ -272,7 +272,7 @@ public static partial class SentrySdk
         => CurrentHub.AddBreadcrumb(clock, message, category, type, data, level);
 
     /// <summary>
-    /// Runs the callback with a new scope which gets dropped at the end.
+    /// Runs the callback within a new scope.
     /// </summary>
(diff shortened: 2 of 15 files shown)


src/Sentry/SentrySdk.cs
```diff
@@ -275,16 +275,63 @@ public void Dispose()
...
     /// <see href="https://docs.sentry.io/platforms/dotnet/enriching-events/scopes/#local-scopes"/>
     /// <param name="scopeCallback">The callback to run with the one time scope.</param>
-    [Obsolete("This method is deprecated in favor of overloads of CaptureEvent, CaptureMessage and CaptureException " +
-              "that provide a callback to a configurable scope.")]
     [DebuggerStepThrough]
     public static void WithScope(Action<Scope> scopeCallback)
         => CurrentHub.WithScope(scopeCallback);
 
+    /// <summary>
+    /// Runs the callback with a new scope which gets dropped at the end.
```
SeanFeldman, line 289: To align the language with .NET terminology

```suggestion
    /// Runs the callback with a new scope which gets disposed of at the end.
```

src/Sentry/SentrySdk.cs
```diff
@@ -275,16 +275,63 @@ public void Dispose()
...
+    /// </remarks>
+    /// <see href="https://docs.sentry.io/platforms/dotnet/enriching-events/scopes/#local-scopes"/>
+    /// <param name="scopeCallback">The callback to run with the one time scope.</param>
+    /// <returns>The result from the callback.</returns>
+    [DebuggerStepThrough]
+    public static T? WithScope<T>(Func<Scope, T?> scopeCallback)
+        => CurrentHub.WithScope(scopeCallback);
+
+    /// <summary>
+    /// Runs the asynchronous callback with a new scope which gets dropped at the end.
```
SeanFeldman, line 304:
```suggestion
    /// Runs the asynchronous callback with a new scope which gets disposed of at the end.
```

src/Sentry/SentrySdk.cs
```diff
@@ -275,16 +275,63 @@ public void Dispose()
...
+    /// </remarks>
+    /// <see href="https://docs.sentry.io/platforms/dotnet/enriching-events/scopes/#local-scopes"/>
+    /// <param name="scopeCallback">The callback to run with the one time scope.</param>
+    /// <returns>An async task to await the callback.</returns>
+    [DebuggerStepThrough]
+    public static Task WithScopeAsync(Func<Scope, Task> scopeCallback)
+        => CurrentHub.WithScopeAsync(scopeCallback);
+
+    /// <summary>
+    /// Runs the asynchronous callback with a new scope which gets dropped at the end.
```
SeanFeldman, line 320:
```suggestion
    /// Runs the asynchronous callback with a new scope which gets disposed of at the end.
```

src/Sentry/Internal/SentryScopeManager.cs
```diff
@@ -101,6 +101,33 @@ public void WithScope(Action<Scope> scopeCallback)
...
+            return scopeCallback.Invoke(scope.Key);
+        }
+    }
+
+    public async Task WithScopeAsync(Func<Scope, Task> scopeCallback)
+    {
+        using (PushScope())
+        {
+            var scope = GetCurrent();
+            await scopeCallback.Invoke(scope.Key).ConfigureAwait(false);
```
SeanFeldman, line 118: Do we want to use something like [`ConfigureAwaitChecker`](https://github.com/cincuranet/ConfigureAwaitChecker) analyzer for development to avoid missing `ConfigureAwait()`?

SeanFeldman (commented):
Minor suggestions

src/Sentry/Internal/SentryScopeManager.cs
mattjohnsonpint: See #2308.  Thanks.

PR #2311, opened 2023-04-18
Use C# 11 for raw string literals

mattjohnsonpint:
We can update the project to use C# 11.  Here's a list of the new features available:

https://learn.microsoft.com/dotnet/csharp/whats-new/csharp-11

While we don't necessarily need to use all of these, we should have the ability to use them where it makes sense.

I've started this with using raw string literals in our tests where we have JSON strings for input or expected output.  This makes them much easier to read, and to copy/paste for validation.  I've also added support for producing indented JSON in our `ToJsonString` test helper method, so that we can use multi-line raw literal strings.

#skip-changelog - because there are no features or bugfixes here.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,5 +1,5 @@
 * text=auto
-*.cs text=auto diff=csharp
+*.cs text=auto eol=lf diff=csharp
 *.sln text=auto eol=crlf
 *.sh eol=lf
 CHANGELOG.md merge=union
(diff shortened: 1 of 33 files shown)


SeanFeldman (approved):
Love this feature!

mattjohnsonpint:
Looks like there's some `\n` vs `\r\n` issues with this on Windows.  I'll look into that shortly.

mattjohnsonpint:
Switched to using `\n` only (LF, not CRLF) for `.cs` files on all OS's (including Windows), enforced via `.gitattributes`.  There's not really any other clean way to control how a newline within a raw string literal is interpreted.

Also updated other test snapshots affected by that change.

PR #2317, opened 2023-04-19
Update console sample

mattjohnsonpint:
Updates the basic console sample to cover the key features we think most users will want to use.

Also uses an MSBuild-based approach to setting the default DSN for sample projects.  This allows a `SENTRY_DSN` environment variable to override the default without having to change any code.  We'll apply the same technique to the other samples as we clean them up one at a time.

See comments inline.

#skip-changelog

### DIFF
--- a/samples/Sentry.Samples.Console.Basic/Program.cs
+++ b/samples/Sentry.Samples.Console.Basic/Program.cs
@@ -10,8 +10,8 @@
 
 using Sentry;
 
-// Initialize the Sentry SDK, disposing when the application is shutting down.
-using var _ = SentrySdk.Init(options =>
+// Initialize the Sentry SDK.  (It is not necessary to dispose it.)
+SentrySdk.Init(options =>
 {
     // A Sentry Data Source Name (DSN) is required.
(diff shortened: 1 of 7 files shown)


mattjohnsonpint:
Also made a few minor changes to the "customized" sample (and the copy of it currently used by profiling).  This was just to get the build passing.  We will need to review those samples with more scrutiny to get them in better shape later.

samples/Sentry.Samples.Console.Basic/Program.cs
```diff
@@ -1,13 +1,91 @@
...
+// Optionally, you can try/catch the exception, and call transaction.Finish(exception) on failure.
+transaction.Finish();
+
+void FirstFunction()
+{
+    // This shows how you might instrument a particular function.
+    var span = transaction.StartChild("function", nameof(FirstFunction));
+
+    // Simulate doing some work
+    Thread.Sleep(100);
```
SeanFeldman, line 58: Most of the code these days is async. Should we turn these functions into asynchronous and use `await Task.Delay(100)` instead?

SeanFeldman (approved):

jamescrosswell (approved):

Issue #2217, opened 2023-03-07
Create a Sentry event for failed HTTP requests


bruno-garcia:
### Problem Statement

Failed HTTP requests don't result in Sentry events

### Solution Brainstorm

* Spec: https://develop.sentry.dev/sdk/features/#http-client-errors

Main issue with description
* https://github.com/getsentry/team-mobile/issues/38

mattjohnsonpint:
Applies to `HttpClient` - we already have `SentryHttpMessageHandler`, so this should be fairly trivial to add there.

We should think about what stacktrace (if any) belongs with the error.

mattjohnsonpint:
This would be part of our [`SentryHttpMessageHandler`](
https://github.com/getsentry/sentry-dotnet/blob/main/src/Sentry/SentryHttpMessageHandler.cs)

jamescrosswell:
I'm looking into this now and in particular the [docs regarding FailedRequestTargets](https://develop.sentry.dev/sdk/features/#http-client-errors).

We already have the [TracePropagationTarget class](https://github.com/getsentry/sentry-dotnet/blob/efb5dc8acea1450de9fe1dd8561ad31c4ef069e4/src/Sentry/TracePropagationTarget.cs), which appears to do exactly what we want. About the only thing wrong with it is the name. If it wasn't public I'd suggest we simply rename it to `PropagationTarget` and reuse it for `SentryOptions.FailedRequestTargets` as well. 

Two alternatives I can think of then (and there may be more):
1. Refactor key functionality to a new `PropagationTarget` class that `TracePropagationTarget` inherits from... 
2. Refactor into a Utils/Helper class, similar to the [PropogationTargetsUtils class](https://github.com/getsentry/sentry-java/blob/b79a165be139ac8f7dc2b56c598831b9d3d8601c/sentry/src/main/java/io/sentry/util/PropagationTargetsUtils.java) in the Java SDK. 
3. Something else entirely. 

Are there any strong opinions about which approach to take? I was leaning towards option 1 personally... but it may have drawbacks I haven't considered.

mattjohnsonpint:
Agreed, the name is too specific.  Even `PropogationTarget` is too specific, IMHO.  What do you think of `SubstringOrRegexPattern`?

jamescrosswell:
`SubstringOrRegexPattern` works... quite literal ;-)

jamescrosswell:
[...]

Issue #1469, opened 2022-02-04
Add `Hint` support to beforeSend and Event processors


lawrence-laz:
Hi, Sentry is great!
I am however missing a way to add an attachment to an outgoing event in the `BeforeSend` callback.
For example:
```csharp
using var host = Host.CreateDefaultBuilder()
    .ConfigureLogging((context, builder) =>
    {
        builder.AddSentry(options =>
        {
            options.BeforeSend = sentryEvent =>
            {
                sentryEvent.AddBreadcrumb("Hello"); // I can do this
                sentryEvent.AddAttachment("Foo.txt"); // But I cannot do this
                return sentryEvent;
            };
        };
    })
    .Build();
```
Am I missing something, or this is a missing feature?

bruno-garcia:
One solution to this is adding `Hint` which is part of the unified API spec: https://develop.sentry.dev/sdk/unified-api/#hints
Recently the Java SDK for hint as Map<string,object>: https://github.com/getsentry/sentry-java/pull/1929

pistoleta:
It would be great to have this

bruno-garcia:
Blocks https://github.com/getsentry/sentry-unity/issues/492

SimonCropp:
@bruno-garcia can u share an example how u expect the api usage to look if we use the hint approach

mattjohnsonpint:
Note, this also applies to `BeforeBreadcrumb`, and our docs already (incorrectly) state that hints are supported.

https://docs.sentry.io/platforms/dotnet/configuration/options/#before-breadcrumb

mattjohnsonpint:
Because we've currently implemented `BeforeSend` as a property that accepts a delegate function, there's no clean way to add a `Hint` parameter to the function without making a breaking change that would require a major version bump.  Methods are overloadable, but properties are not.

I'd like to handle this as follows:

```csharp
public class SentryOptions
{
    // ...

    internal Func<SentryEvent, Hint, SentryEvent?>? _beforeSend;

    public void SetBeforeSend(Func<SentryEvent, SentryEvent?> beforeSend)
    {
        _beforeSend = (e, _) => beforeSend(e);
    }

    public void SetBeforeSend(Func<SentryEvent, Hint, SentryEvent?> beforeSend)
    {
        _beforeSend = beforeSend;
    }

    [Obsolete("This property will be removed in a future version. Use SetBeforeSend instead.")]
[...]

PR #2320, opened 2023-04-20
Create a Sentry event for failed HTTP requests

jamescrosswell:
Should eventually address #2217 Create a Sentry event for failed HTTP requests

Still very much WIP. 

Will comment inline to create discussion.


src/Sentry/SentryFailedRequestHandler.cs
```diff
@@ -0,0 +1,100 @@
...
+using Sentry.Protocol;
+
+namespace Sentry
+{
+    internal class SentryFailedRequestHandler : ISentryFailedRequestHandler
+    {
+        private readonly IHub _hub;
+        private readonly SentryOptions? _options;
+
+        public string MechanismType { get => "SentryFailedRequestHandler"; }
```
mattjohnsonpint, line 15:
```suggestion
        public const string MechanismType = "SentryFailedRequestHandler";
```

src/Sentry/HttpStatusCodeRange.cs
```diff
@@ -0,0 +1,46 @@
+namespace Sentry
```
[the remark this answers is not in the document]
mattjohnsonpint: Same for other new files please.

src/Sentry/SentryFailedRequestHandler.cs
```diff
@@ -0,0 +1,100 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using Sentry.Protocol;
+
```
mattjohnsonpint, line 7:
```suggestion
```

The `System` namespaces are already being imported globally, and I don't think you need `Sentry.Protocol` here.  In general, we try to reduce `using` statements to  the minimum needed to compile.

src/Sentry/SentryFailedRequestHandler.cs
```diff
@@ -0,0 +1,100 @@
...
+    internal class SentryFailedRequestHandler : ISentryFailedRequestHandler
+    {
+        private readonly IHub _hub;
+        private readonly SentryOptions? _options;
+
+        public string MechanismType { get => "SentryFailedRequestHandler"; }
+
+        /// <summary>
+        /// Initializes an instance of <see cref="SentryFailedRequestHandler"/>.
+        /// </summary>
```
mattjohnsonpint, line 19: We only require XML summary comments for public members.  I'd omit them if they don't add any useful information.

src/Sentry/SentryFailedRequestHandler.cs
```diff
@@ -0,0 +1,100 @@
...
+        {
+            _hub = hub;
+            _options = options;
+        }
+
+        /// <summary>
+        /// Automaticlaly capture HTTP Client errors when captureFailedRequests
+        /// </summary>
+        /// <param name="request">The HttpRequestMessage sent</param>
+        /// <param name="response">The HttpResponse received</param>
```
mattjohnsonpint, line 30: Some typos here also.  But again, you can just remove it since the class is internal.

src/Sentry/SentryFailedRequestHandler.cs
```diff
@@ -0,0 +1,100 @@
...
+        public void CaptureEvent(HttpRequestMessage request, HttpResponseMessage response)
+        {
+            if (request is null)
+                return;
```
mattjohnsonpint, line 34: The `request` will never be null.  But this would be a good place to do a sanity check:

```csharp
if (response.IsSuccessStatusCode)
{
    return;
}
```

Regardless of the range provided, we'd not want to do anything for success status.

src/Sentry/SentryHttpClientException.cs
mattjohnsonpint, line 1: We can eliminate this and just use `HttpRequestException`.

src/Sentry/SentryFailedRequestHandler.cs
```diff
@@ -0,0 +1,100 @@
...
+using System.Text;
+using System.Threading.Tasks;
+using Sentry.Protocol;
+
+namespace Sentry
+{
+    internal class SentryFailedRequestHandler : ISentryFailedRequestHandler
+    {
+        private readonly IHub _hub;
+        private readonly SentryOptions? _options;
```
mattjohnsonpint, line 13:
```suggestion
        private readonly SentryOptions _options;
```

I don't think this class can operate without options.  Let's make it non-nullable.  Adjust throughout please.

src/Sentry/SentryHttpMessageHandler.cs
```diff
@@ -18,12 +19,14 @@ public SentryHttpMessageHandler(IHub hub)
     {
         _hub = hub;
         _options = hub.GetSentryOptions();
+        _failedRequestHandler = new SentryFailedRequestHandler(_hub, _options);
     }
 
-    internal SentryHttpMessageHandler(IHub hub, SentryOptions options)
+    internal SentryHttpMessageHandler(IHub hub, ISentryFailedRequestHandler failedRequestHandler, SentryOptions options)
```
mattjohnsonpint, line 25:
```suggestion
    internal SentryHttpMessageHandler(IHub hub, SentryOptions options, ISentryFailedRequestHandler? failedRequestHandler = null)
```

src/Sentry/SentryHttpMessageHandler.cs
```diff
@@ -18,12 +19,14 @@ public SentryHttpMessageHandler(IHub hub)
     {
         _hub = hub;
         _options = hub.GetSentryOptions();
+        _failedRequestHandler = new SentryFailedRequestHandler(_hub, _options);
```
mattjohnsonpint, line 22:
```suggestion
        if (_options != null)
        {
            _failedRequestHandler = new SentryFailedRequestHandler(_hub, _options);
        }
```

src/Sentry/SentryHttpMessageHandler.cs
```diff
@@ -35,8 +38,8 @@ public SentryHttpMessageHandler(HttpMessageHandler innerHandler, IHub hub)
         InnerHandler = innerHandler;
     }
 
-    internal SentryHttpMessageHandler(HttpMessageHandler innerHandler, IHub hub, SentryOptions options)
-        : this(hub, options)
+    internal SentryHttpMessageHandler(HttpMessageHandler innerHandler, ISentryFailedRequestHandler failedRequestHandler, IHub hub, SentryOptions options)
+        : this(hub, failedRequestHandler, options)
```
mattjohnsonpint, line 42:
```suggestion
    internal SentryHttpMessageHandler(HttpMessageHandler innerHandler, IHub hub, SentryOptions options, ISentryFailedRequestHandler? failedRequestHandler = null)
        : this(hub, options, failedRequestHandler)
```

src/Sentry/SentryFailedRequestHandler.cs
```diff
@@ -0,0 +1,100 @@
...
+                return;
+
+            // Capture the event
+            var exception = new SentryHttpClientException($"HTTP Client Error with status code: {response.StatusCode}");
+            exception.SetSentryMechanism(MechanismType);
+            var @event = new SentryEvent(exception);
+            _hub.CaptureEvent(@event);
+
+            /*
+             * Copied from SentryOkHttpInterceptorTest.kt in the Java SDK for reference
```
mattjohnsonpint, line 61: Ok for now, but don't forget to remove before completing this. 😉

src/Sentry/SentryFailedRequestHandler.cs
```diff
@@ -0,0 +1,100 @@
...
+
+            // Ignore requests that don't match the FailedRequestTargets
+            if (_options?.FailedRequestTargets.ContainsMatch(requestString) is false)
+                return;
+
+            // Capture the event
+            var exception = new SentryHttpClientException($"HTTP Client Error with status code: {response.StatusCode}");
+            exception.SetSentryMechanism(MechanismType);
+            var @event = new SentryEvent(exception);
+            _hub.CaptureEvent(@event);
```
mattjohnsonpint, line 58:
```suggestion
            try
            {
                response.EnsureSuccessStatusCode();
            }
            catch (HttpRequestException exception)
            {
                exception.SetSentryMechanism(MechanismType);
                _hub.CaptureException(exception);
            }
```

src/Sentry/SentryFailedRequestHandler.cs
```diff
@@ -0,0 +1,100 @@
...
+                return;
+
+            // Don't capture events for successful requets
+            if (_options?.FailedRequestStatusCodes.Any(range => range.Contains(response.StatusCode)) is false)
+                return;
+
+            // Ignore requests to the Sentry DSN
+            var uri = request.RequestUri;
+            var requestString = uri?.OriginalString ?? "";
+            if (_options?.Dsn is { } dsn && new SubstringOrRegexPattern(dsn).IsMatch(requestString))
```
mattjohnsonpint, line 47: Just compare the hosts.

```suggestion
            var uri = response.RequestMessage.RequestUri;
            if (_options.Dsn is { } dsn && new Uri(dsn).Host.Equals(uri.Host, StringComparison.OrdinalIgnoreCase))
```

src/Sentry/SentryFailedRequestHandler.cs
mattjohnsonpint: Also please use braces on all if-statement bodies throughout.  Thanks.

src/Sentry/SentryOptions.cs
```diff
@@ -522,6 +522,41 @@ public bool ReportAssemblies
     /// </summary>
     public string? CacheDirectoryPath { get; set; }
 
+    /// <summary>
+    /// The SDK will only capture HTTP Client errors if it is enabled.
+    ///
```
mattjohnsonpint, line 527: nit: remove extra blank lines in the doc comments.

mattjohnsonpint:
Good approach.  👍

test/Sentry.Tests/SentryFailedRequestHandlerTests.cs
```diff
@@ -0,0 +1,294 @@
...
+
+        // Assert
+        using (new AssertionScope())
+        {            
+            @event.Should().NotBeNull();
+
+            @event.Request.Method.Should().Be(HttpMethod.Post.ToString());
+            @event.Request.Url.Should().Be(url);
+            @event.Request.QueryString.Should().Be(queryString);
+            // Not sure how we want to handle fragments in .NET... potentially we add a Fragment
```
jamescrosswell, line 177: The comment says it all here:
Not sure how we want to handle fragments in .NET... potentially we add a Fragment property to the Request class, but that might be a breaking change.

Other options:

We don't capture the fragment specifically (just leave it part of the queryString)... but that would be inconsistent with the Java API
We could put the fragment in the Request.Other collection (as shown in the code that's commented out below).

I think the cleanest and most consistent option would be to add a Fragment property to the Request class but that's a losely held opinion.

src/Sentry/SentryHttpMessageHandler.cs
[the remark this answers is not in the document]
mattjohnsonpint: I can see how that would be useful, but I don't think it's critical at this time.  Let's keep it in mind for the future though.  Feel free to write an enhancement issue for it so we don't lose track.  Thanks.

test/Sentry.Tests/SentryFailedRequestHandlerTests.cs
mattjohnsonpint: Fragments are constrained to the browser by design of URLs. They are not sent over HTTP at all.

mattjohnsonpint: I see it in the Java code.  TBH, I have no clue why they did that there.  Maybe something specific to Android or Java?  AFAIK, fragments don't go over HTTP - period.  They're for the browser only.

test/Sentry.Tests/SentryFailedRequestHandlerTests.cs
```diff
@@ -0,0 +1,294 @@
...
+            // property to the Request class, but that might be a breaking change.
+            // @event.Request.Other.Should().Contain(other => other.Key == "fragment" && other.Value == fragment);
+
+            throw new NotImplementedException("Test doesn't yet check for the response is set") ;
```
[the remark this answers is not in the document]
mattjohnsonpint: Hints are coming in #1469.  For now, let's skip over the part where the spec doc says to add the request and response as hints.  We can add that in later.  Thanks.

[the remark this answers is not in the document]
mattjohnsonpint: Added a comment to https://github.com/getsentry/sentry-dotnet/issues/1469#issuecomment-1518509806

If we can manage Hints around the same time as this, then we can do them both.  Thanks.

test/Sentry.Tests/SentryFailedRequestHandlerTests.cs
mattjohnsonpint: I'll check with the Java team, but I think their scrubbing of fragments was unnecessary.  They should already be absent.  Per the HTTP spec, HTTP works with URLs.  Fragments are part of URIs, but not URLs, and they're disallowed over HTTP.

test/Sentry.Tests/SentryFailedRequestHandlerTests.cs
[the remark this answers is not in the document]
mattjohnsonpint: > I see the HttpTransportBase class has a HandleFailure method that captures a failed HttpResponseMessage ...

That's specifically for our SDK failing to communicate to Sentry.  We would not want to do the same for this.

In this particular case, we're not actually going to be sending the failed HTTP request or response messages to Sentry, but rather we'd just be exposing them on a hint, such that the user has them available in methods such as `BeforeSend`, and in custom event processors, etc.  They can be ignored, or they can use them to decide whether to filter or augment the events, or they could manually copy them as attachments or extras to the event at that time.  But we're not doing it automatically.  Main reason being that the risk of capturing sensitive information is too high - even if protected by the PII option.

src/Sentry/SubstringOrRegexPattern .cs
mattjohnsonpint, line 1: You have an extra space in the filename.

test/Sentry.Tests/SentryFailedRequestHandlerTests.cs
mattjohnsonpint: Confirmed. We don't need to worry about fragments.  They will never be present.  This was a misunderstanding, and shouldn't be in the Java SDK either, nor in the Sentry protocol (`event.request.fragment`). I'll look to have it deprecated.

src/Sentry/SentryFailedRequestHandler.cs
jamescrosswell: Sounds good. I think VS generated those comments automatically when I added the class... but I'll strip them out.

src/Sentry/SentryFailedRequestHandler.cs
```diff
@@ -0,0 +1,95 @@
...
+        {
+            response.EnsureSuccessStatusCode();
+        }
+        catch (HttpRequestException exception)
+        {
+            exception.SetSentryMechanism(MechanismType);
+
+            var @event = new SentryEvent(exception);
+
+            var sentryRequest = new Request
```
jamescrosswell, line 66: @mattjohnsonpint excepting hints, this implementation now mimics [what the Java SDK is doing](https://github.com/getsentry/sentry-java/blob/0be0d10a9fc61369a551c3b90134fe36c050a584/sentry-android-okhttp/src/main/java/io/sentry/android/okhttp/SentryOkHttpInterceptor.kt#L183). The Java SDK sets the `event.Request` and `event.Contexts["response"]` properties. However it only includes cookies and headers if `SendDefaultPii == true`.

test/Sentry.Tests/ApiApprovalTests.Run.DotNet7_0.verified.txt
```diff
@@ -371,6 +388,17 @@ namespace Sentry
         public void WriteTo(System.Text.Json.Utf8JsonWriter writer, Sentry.Extensibility.IDiagnosticLogger? logger) { }
         public static Sentry.Request FromJson(System.Text.Json.JsonElement json) { }
     }
+    public sealed class ResponseContext : Sentry.IJsonSerializable
+    {
+        public ResponseContext() { }
+        public long? BodySize { get; set; }
+        public string? Cookies { get; set; }
+        public System.Collections.Generic.IDictionary<string, string> Headers { get; }
```
jamescrosswell, line 396: @mattjohnsonpint I see `VerifyExtensions.IgnoreStandardSentryMembers` ignores the Dictionary properties of the Request class. I wondered if we should be doing the same for ResponseContext.Headers... but I'm not familiar with Verify so wanted to understand why we'd want to do this before configuring additional `IgnoreMembers` in there.

src/Sentry/ResponseContext.cs
jamescrosswell, line 1: @mattjohnsonpint I hesitated about whether to call this `Response` (which would match our `Sentry.Request` class nicely) or ResponseContext (which is [what it is named in the docs](https://develop.sentry.dev/sdk/event-payloads/types/#responsecontext)). 

LMK.

mattjohnsonpint: Call it `Response` and move it to `Sentry.Protocols`.  Also update `Contexts.cs` to follow pattern used by other context types.

In v4, we can consider renaming all of these types to match (ie, `ResponseContext`, `GpuContext`, etc.). But I don't want them mismatched in the meantime.  Thanks.

test/Sentry.Tests/ApiApprovalTests.Run.DotNet7_0.verified.txt
mattjohnsonpint: We should only ignore members that vary significantly every time they're run, or vary when run on different machines.  The point is to have snapshots that work everywhere.  If they already do, then no need to add more ignores.

src/Sentry/SentryFailedRequestHandler.cs
mattjohnsonpint: Sounds good.  Thanks!

src/Sentry/ResponseContext.cs
jamescrosswell: Should I move `Request` into the `Sentry.Protocol` folder while I'm at it?

>Also update Contexts.cs to follow pattern used by other context types.

Not sure what you mean here... I haven't (yet) touched `Contexts.cs` - although I'm using it to store the `Response` obviously. 

>In v4, we can consider renaming all of these types to match (ie, ResponseContext, GpuContext, etc.). But I don't want them mismatched in the meantime. Thanks.

Ah, so the docs are kind of future focused ;-)

mattjohnsonpint: No, leave `Request` where it is.  It's not a "context" object, because it goes directly on its own property on the event.

> I haven't (yet) touched Contexts.cs...

Open `src/Sentry/Contexts.cs`.  Your `Response` object should be similar to the others there, like `App`, `Browser`, `Device`, etc.

mattjohnsonpint: The [Event Type Definition](https://develop.sentry.dev/sdk/event-payloads/types/) docs are auto-generated.  The type names there are from the corresponding Rust sources code in https://github.com/getsentry/relay.  They are usually good type names, but ultimately it's only the serialization that matters.

jamescrosswell:
OK I think this is working now... @mattjohnsonpint see [sample event](https://sentry-sdks.sentry.io/issues/4131074301/?project=5428537).

.editorconfig
```diff
@@ -194,4 +194,7 @@ dotnet_diagnostic.RCS1118.severity = silent
 dotnet_diagnostic.RCS1080.severity = silent
 
 # Declare type inside namespace
-dotnet_diagnostic.RCS1110.severity = silent
\ No newline at end of file
+dotnet_diagnostic.RCS1110.severity = silent
+csharp_using_directive_placement = outside_namespace:silent
```
mattjohnsonpint, line 198:
```suggestion
csharp_using_directive_placement = outside_namespace:suggestion
```

.editorconfig
mattjohnsonpint, line 1: Let's start a new PR for `.editorconfig` changes.  There are probably others.

mattjohnsonpint: Started #2341.  (and went with `outside_namespace:suggestion` instead of `outside_namespace:silent`)

src/Sentry/SentryFailedRequestHandler.cs
```diff
@@ -0,0 +1,100 @@
...
+
+            var responseContext = new Response {
+                StatusCode = (short)response.StatusCode
+            };
+
+            // .NET 4.8 doesn't set the Content-Length header
+            // https://github.com/dotnet/runtime/issues/16162
+#if NET5_0_OR_GREATER
+            responseContext.BodySize = response.Content?.Headers?.ContentLength;
+#endif
```
mattjohnsonpint, line 86: Despite their bug, we probably shouldn't restrict in this way.  If it returns `null`, there's no harm to us - as that means we couldn't collect it.  (If they fix this bug, we'd not get that benefit.)

Also, this particular conditional compilation constant would exclude .NET Standard and earlier .NET core targets as well.

So unless it is throwing an exception, then just let it return null.  Also the extra null checks aren't required.

```suggestion
            responseContext.BodySize = response.Content.Headers.ContentLength;
```

mattjohnsonpint: I couldn't reproduce that bug anyway.  Perhaps they've fixed it and forgot to update GitHub.  Either way, let's just take the value or null.

mattjohnsonpint (approved):

src/Sentry/SentryFailedRequestHandler.cs
```diff
@@ -0,0 +1,109 @@
...
+        // Starting with .NET 5, the content and headers are guaranteed to not be null.
+        var bodySize = response.Content.Headers.ContentLength;
+#else
+        // We have to get the content body size before calling EnsureSuccessStatusCode,
+        // because older implementations of EnsureSuccessStatusCode disposes the content.
+        // See https://github.com/dotnet/runtime/issues/24845
+
+        // The ContentLength might be null (but that's ok).
+        // See https://github.com/dotnet/runtime/issues/16162
+        var bodySize = response.Content?.Headers?.ContentLength;
```
jamescrosswell, line 65: @mattjohnsonpint this was the approach I'd originally tried. However it's not that it returns null - when running tests against the net43 target it sometimes throws a System.ObjectDisposedException... so I don't think this code is safe. 

I also tried variants like `response.Content?.Headers.?.Contains("Content-Length")` as well, which avoided the exception on net43 but also prevented us from capturing the BodySize in later versions of .NET where it could be calculated but wasn't stored in the header. 

It looks like [HttpContent.TryComputeLength](https://learn.microsoft.com/en-us/dotnet/api/system.net.http.httpcontent.trycomputelength?view=net-8.0) (which I think is where they fixed all this) was introduced in .NET 4.5... so possibly `NET5_0_OR_GREATER` is the wrong cut-off point... 

An alternative I toyed with was `try... catch(ObjectDisposedException ex) {}`. Possibly that's a better solution but it would definitely need to be accompanied by a code comment explaining why it was there... we'd need to keep using an `#if` directive in the unit tests, but it would mean we'd be capturing the BodySize in all circumstances where this was possible (failing only when an exception was thrown by older versions of the .NET Framework).

mattjohnsonpint: It's safe now, as it's been moved before `EnsureSuccessStatusCode` which used to dispose the content (which is why you got `ObjectDisposedException`).  See my comment just above this one.

As for specifically checking `"Content-Length"` - we could add that if `bodySize == null`, but I'd want to reproduce the issue first.  I tried locally (on both NetFx and Mono) and the value was always present if it came from an actual `HttpClient` instance.  And if it came just from `ResponseMessage` (like in the tests), then `Headers` was null, so there'd be nothing to check against.

Either way, I'm not too worried about it.  The disposed exception is the main thing, which is taken care of.

jamescrosswell: Ah, interesting... How did you work out where Dispose was being called? 

I also just reaslised that relying on EnsureSuccessStatusCode to throw an exception means it's not possible to configure `FailedRequestStatusCodes` in the 200-299 range... nothing in that range will result in an exception being thrown here that will be captured by Sentry. 

I can't think why someone would want to do that... Something to be aware of though.

PR #2332, opened 2023-04-26
Include Dynamic Sampling Context with error events, when there's a transaction

### LINKED ISSUE #2330: Send DSC on error events
### Problem Statement

We currently send the Dynamic Sampling Context (DSC) on transactions, as per the original design.

This has recently changed in the product, and it is now beneficial to also send the DSC for errors as well.

### Solution Brainstorm

- Ensure that the DSC is sent on errors.
- In ASP.NET and ASP.NET Core, ensure that the DSC is populated from incoming baggage - even if tracing is disabled.
  - This may require some reworking of middleware components, as baggage is currently collected in `SentryTracingMiddleware`, not `SentryMiddleware`.

mattjohnsonpint:
Resolves #2330

### DIFF
--- a/src/Sentry/SentryEvent.cs
+++ b/src/Sentry/SentryEvent.cs
@@ -197,6 +197,8 @@ public sealed class SentryEvent : IEventLike, IJsonSerializable, IHasDistributio
         ) ?? false;
     }
 
+    internal DynamicSamplingContext? DynamicSamplingContext { get; set; }
+
     /// <summary>
     /// Creates a new instance of <see cref="T:Sentry.SentryEvent" />.
     /// </summary>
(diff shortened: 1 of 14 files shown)


bruno-garcia:
Relates to: https://github.com/getsentry/team-replay/issues/70

src/Sentry/SentryEvent.cs
```diff
@@ -197,6 +197,8 @@ internal bool HasTerminalException()
         ) ?? false;
     }
 
+    internal DynamicSamplingContext? DynamicSamplingContext { get; set; }
```
bruno-garcia, line 200: This means it doesn't work with offline mode? That's OK btw, just checking.

bruno-garcia (approved):
🚀

src/Sentry/SentryEvent.cs
mattjohnsonpint: It still works.  We cache envelopes, including the headers - so that includes the `trace` item header where the DSC is written.

PR #2339, opened 2023-04-28
Fix ASP.NET Core issue with missing context when using capture methods that configure scope

### LINKED ISSUE #2337: HTTP info not applies to local scope
### Package

Sentry.AspNetCore

### SDK Version

3.30.0

### Steps to Reproduce

Description: 
When trying to addBreadcrumbs to local scope, HTTP info/context is not applied to events unless you capture an event to the global scope prior to applying breadcrumbs to local scope. Test utilizing `UseSentry()` and attached repro app - [MyWebApp.zip](https://github.com/getsentry/sentry-dotnet/files/11348017/MyWebApp.zip)

Steps to repro:
1. Add DSN
2. Navigate to Index.cshtml.cs -> Run with local scope -> event has breadcrumbs with missing HTTP context
3. Uncomment first try/catch -> run -> find event with breadcrumbs now has HTTP context
4. Comment first try/catch  -> Uncomment last try/catch -> run -> now eventa with local scope has breadcrumbs with missing HTTP context

### Expected Result

Issues to have HTTP context when addingBreadcrumbs to the local scope when initializing with UseSentry()

### Actual Result

Only when adding the global scope first, is the HTTP context applied for local scope

mattjohnsonpint:
Fixes #2337

### DIFF
--- a/src/Sentry.AspNetCore/SentryMiddleware.cs
+++ b/src/Sentry.AspNetCore/SentryMiddleware.cs
@@ -8,7 +8,6 @@ using Microsoft.AspNetCore.Http;
 using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Options;
 using Sentry.Extensibility;
-using Sentry.Protocol;
 using Sentry.Reflection;
 
 namespace Sentry.AspNetCore;
@@ -99,10 +98,13 @@ internal class SentryMiddleware : IMiddleware
                 // sent to Sentry. This avoid the cost on error-free requests.
                 // In case of event, all data made available through the HTTP Context at the time of the
                 // event creation will be sent to Sentry
-                scope.OnEvaluating += (_, _) =>
+
+                // Important: The scope that the event is attached to is not necessarily the same one that is active
+                // when the event fires.  Use `activeScope`, not `scope` or `hub`.
+                scope.OnEvaluating += (_, activeScope) =>
                 {
-                    SyncOptionsScope(hub);
-                    PopulateScope(context, scope);
+                    SyncOptionsScope(activeScope);
+                    PopulateScope(context, activeScope);
                 };
             });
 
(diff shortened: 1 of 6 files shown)


mattjohnsonpint:
Fixed some formatting issues as well, so you may want to review while hiding whitespace.

src/Sentry/Scope.cs
```diff
@@ -461,6 +464,8 @@ public void Apply(Scope other)
     public Scope Clone()
     {
         var clone = new Scope(Options);
+        clone.OnEvaluating = OnEvaluating;
```
mattjohnsonpint, line 467: This was required to ensure the event handler attached in the middleware is still called.  Otherwise, the new scope has no handlers attached to this event.

src/Sentry.AspNetCore/SentryMiddleware.cs
```diff
@@ -99,10 +98,13 @@ public async Task InvokeAsync(HttpContext context, RequestDelegate next)
...
-                scope.OnEvaluating += (_, _) =>
+
+                // Important: The scope that the event is attached to is not necessarily the same one that is active
+                // when the event fires.  Use `activeScope`, not `scope` or `hub`.
+                scope.OnEvaluating += (_, activeScope) =>
                 {
-                    SyncOptionsScope(hub);
-                    PopulateScope(context, scope);
+                    SyncOptionsScope(activeScope);
+                    PopulateScope(context, activeScope);
```
mattjohnsonpint, line 107: This was required so that we are using the scope that the event fired on - which isn't necessarily the same one it's attached to (after we copy over the event handlers in the `Clone` method below).

src/Sentry.AspNetCore/SentryMiddleware.cs
```diff
@@ -167,11 +169,11 @@ void CaptureException(Exception e, SentryId evtId, string mechanism, string desc
         }
     }
 
-    private void SyncOptionsScope(IHub newHub)
+    private void SyncOptionsScope(Scope scope)
     {
         foreach (var callback in _options.ConfigureScopeCallbacks)
         {
-            newHub.ConfigureScope(callback);
+            callback.Invoke(scope);
```
mattjohnsonpint, line 176: We want the callbacks to fire on the scope the event was fired on.

SeanFeldman (approved):

PR #2346, opened 2023-04-29
Azure Functions (Isolated Worker/Out-of-Process) support

### LINKED ISSUE #373: Improved integration for Azure Functions
[Azure functions support DI](https://docs.microsoft.com/en-us/azure/azure-functions/functions-dotnet-dependency-injection).

Right now the entry point is on `IWebHost`. This is not expose on serverless environments. Which means users need to `Init` and `Dispose` the SDK on each execution or some other work around that at `Init` once but `await SentrySdk.FlushAsync` by hand.

This task is to improve the story so that it's easy to plug Sentry's SDK with Azure Functions via DI (i.e: single Hook into `IFunctionsHostBuilder`).

Related to #730  #529

SeanFeldman:
Adds support for Azure Functions ([Isolated Worker/Out-of-Process SDK](https://learn.microsoft.com/en-us/azure/azure-functions/dotnet-isolated-process-guide))

Closes #373


src/Sentry.AzureFunctions.Worker/Sentry.AzureFunctions.Worker.csproj
```diff
@@ -0,0 +1,15 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+    <PropertyGroup>
+      <TargetFrameworks>net7.0;net6.0</TargetFrameworks>
```
mattjohnsonpint, line 4: Unless there's something special we need that's only open to .NET 7, then please target .NET 6 only in the source libraries.  Thanks.

src/Sentry.AzureFunctions.Worker/Sentry.AzureFunctions.Worker.csproj
```diff
@@ -0,0 +1,15 @@
...
+      <PackageTags>$(PackageTags);Azure;Functions;Worker</PackageTags>
+      <Description>Official Azure Functions Worker SDK integration for Sentry - Open-source error tracking that helps developers monitor and fix crashes in real time.</Description>
+    </PropertyGroup>
+
+    <ItemGroup>
+      <PackageReference Include="Microsoft.Azure.Functions.Worker" Version="1.13.0" />
+      <PackageReference Include="Microsoft.Azure.Functions.Worker.Extensions.Http" Version="3.0.12" />
+      <PackageReference Include="Sentry.Extensions.Logging" Version="3.30.0" />
```
mattjohnsonpint, line 12: We'll want this to be a project reference.

src/Sentry.AzureFunctions.Worker/SentryFunctionsWorkerMiddleware.cs
```diff
@@ -0,0 +1,81 @@
...
+        try
+        {
+            var transactionName = await GetTransactionNameAsync(context) ?? context.FunctionDefinition.Name;
+
+            _hub.ConfigureScope(scope =>
+            {
+                var transaction = _hub.StartTransaction(transactionName, "function");
+                scope.Transaction = transaction;
+
+                context.CancellationToken.Register(() => scope.SetExtra("aborted", true));
```
mattjohnsonpint, line 26: I think this is unnecessary, since we're already setting the transaction aborted here:

https://github.com/getsentry/sentry-dotnet/blob/b1bfe1efc04eb4c911a85f1cf4cd2e5a176d7c8a/src/Sentry/Internal/Hub.cs#L360-L366

That's presuming a `OperationCanceledException` (or `TaskCanceledException`) is already being thrown during cancellation.  If not, then I think we should just do `context.CancellationToken.ThrowIfCancellationRequested();` here, rather than registering.

SeanFeldman: I'm assuming we still want to start a transaction and then call `context.CancellationToken.ThrowIfCancellationRequested();` so Sentry captures it as part of the transaction. Right, @mattjohnsonpint?

mattjohnsonpint: Yes, exactly.  I just meant registering for cancellation was probably unnecessary.

SeanFeldman:
Looking at the supported .NET versions, .NET 5 and .NET Standard 2.0 are [supported](https://www.nuget.org/packages/Microsoft.Azure.Functions.Worker#supportedframeworks-body-tab) by `Microsoft.Azure.Functions.Worke`. Should the Sentry package support `net5.0` and `netstandard2.0` instead of [the current](https://github.com/getsentry/sentry-dotnet/pull/2346/files#diff-a6e901f0b28e2b0ea5c1c3eb48e7e5236727f0785ce4b5094319f2bffb17cc48R4) `net6.0`, @mattjohnsonpint?

SeanFeldman:
@mattjohnsonpint, keeping it as a draft, but you can start tearing this apart 🙂

src/Sentry.AzureFunctions.Worker/.editorconfig
```diff
@@ -0,0 +1,5 @@
+# https: //docs.microsoft.com/en-us/visualstudio/ide/editorconfig-code-style-settings-reference
+[*.cs]
+
+# Justification: Azure Functions (single thread) project
+dotnet_diagnostic.CA2007.severity = none
```
mattjohnsonpint, line 5: I'm not so sure about this.

`ConfigureAwait(false)` (CA2007) isn't about being single threaded or not (and I'm not sure that's accurate anyway).

Rather, `ConfigureAwait` is about _synchronization context_.  The `false` is assigned to the `continueOnCapturedContext` parameter,  which defaults to `true`.  However, if there's no synchronization context, then it doesn't matter what this setting is.

For example, ASP.NET Core applications don't have a synchronization context, but ASP.NET Framework and desktop applications do.  The usual advice for _libraries_ is to always set `ConfigureAwait(false)`.

As for Azure Functions, I can find very little concrete information about this with regard to the isolated worker model.  I only found a bit about [durable functions](https://learn.microsoft.com/azure/azure-functions/durable/durable-functions-code-constraints#net-threading-apis) having a synchronization context, and requiring `ConfigureAwait(true)` (or omitting for the default).  Even then, this would apply to the orchestrator functions themselves, not necessarily any libraries they are using.

One thing that took me a long time to wrap my head around is that the placement of `ConfigureAwait` affects the function in a way that is not what you might think it is.

```csharp
var result = await DoSomethingAsync().ConfigureAwait(false);
```

You might think it's interpreted like this:

```csharp
var result = (await DoSomethingAsync()).ConfigureAwait(false);
```

But actually it's like this:

```csharp
var result = await (DoSomethingAsync().ConfigureAwait(false));
```

So the `ConfigureAwait` is affecting the `await` itself - which controls the step in the state machine at this particular await.  Thus in the following:

```csharp
var result = await DoSomethingAsync();

async Task<T> DoSomethingAsync()
{
    return await DoSomethingElseAsync().ConfigureAwait(false);
}
```

... there is an implicit `ConfigureAwait(true)` in the first call, so the `ConfigureAwait(false)` in the inner function only applies to the `await` in _that_ function.  This is why we set false in library code - because we don't want to create any _additional_ synchronization requirements.  It doesn't affect the synchronization of the caller.

I can't think of why someone would _need_ synchronization context in an Azure functions isolated worker, but I can see that there might still be one present.  Perhaps only with a .NET Framework implementation?  Not sure.  Still - as a library, we should still set `ConfigureAwait(false)` as to not create additional synchronization requirements.

src/Sentry.AzureFunctions.Worker/SentryFunctionsWorkerMiddleware.cs
```diff
@@ -0,0 +1,82 @@
...
+    public SentryFunctionsWorkerMiddleware(IHub hub)
+    {
+        _hub = hub;
+    }
+
+    public async Task Invoke(FunctionContext context, FunctionExecutionDelegate next)
+    {
+        try
+        {
+            var transactionName = await GetTransactionNameAsync(context) ?? context.FunctionDefinition.Name;
```
mattjohnsonpint, line 19:
```suggestion
            var transactionName = await GetTransactionNameAsync(context).ConfigureAwait(false)
                ?? context.FunctionDefinition.Name;
```

src/Sentry.AzureFunctions.Worker/SentryFunctionsWorkerMiddleware.cs
```diff
@@ -0,0 +1,82 @@
...
+
+            _hub.ConfigureScope(scope =>
+            {
+                var transaction = _hub.StartTransaction(transactionName, "function");
```
mattjohnsonpint, line 23: This starts the transaction, but we need to finish it later, and set status correctly.  See `src/Sentry.AspNetCore/SentryTracingMiddleware.cs` and follow a similar process.

src/Sentry.AzureFunctions.Worker/SentryFunctionsWorkerMiddleware.cs
```diff
@@ -0,0 +1,82 @@
...
+                scope.Transaction = transaction;
+
+                scope.UnsetTag("AzureFunctions_FunctionName");
+                scope.UnsetTag("AzureFunctions_InvocationId");
```
mattjohnsonpint, line 27: A comment here would be useful, describing how these were set in `SentryLogger` because they were passed in as `state` when the functions runtime was logging.

Or maybe a better way to handle this would be to prevent them from being added in the first place?  We could make an option on `SentryLoggingOptions` for tag filters, and then use them in the `SentryLogger`.  Then our Azure Functions library could just set these as tag filters during initialization.

src/Sentry.AzureFunctions.Worker/SentryFunctionsWorkerMiddleware.cs
```diff
@@ -0,0 +1,82 @@
...
+
+        // Find the HTTP Trigger attribute via reflection
+        var assembly = Assembly.LoadFrom(context.FunctionDefinition.PathToAssembly);
+        var entryPointName = context.FunctionDefinition.EntryPoint;
+
+        var typeName = entryPointName[..entryPointName.LastIndexOf('.')];
+        var methodName = entryPointName[(typeName.Length + 1)..];
+        var attribute = assembly.GetType(typeName)?.GetMethod(methodName)?.GetParameters()
+            .Select(p => p.GetCustomAttribute<HttpTriggerAttribute>())
+            .FirstOrDefault(a => a is not null);
```
mattjohnsonpint, line 70: Since we have no choice but to use reflection - we should probably keep a cache so we don't have to do it repeatedly.

samples/Sentry.Samples.AzureFunctions.Worker/.gitignore
mattjohnsonpint, line 1: I don't think we need another full `.gitignore`.  If there are new things to ignore, add them to the one in our root.

samples/Sentry.Samples.AzureFunctions.Worker/local.settings.json
```diff
@@ -0,0 +1,7 @@
+{
+    "IsEncrypted": false,
+    "Values": {
+        "AzureWebJobsStorage": "UseDevelopmentStorage=true",
```
mattjohnsonpint, line 4: Do we need _any_ storage for this sample?

src/Sentry.AzureFunctions.Worker/Sentry.AzureFunctions.Worker.csproj
```diff
@@ -0,0 +1,23 @@
...
+      <Version>0.1.0-beta</Version>
+    </PropertyGroup>
+
+    <ItemGroup>
+      <ProjectReference Include="..\Sentry.Extensions.Logging\Sentry.Extensions.Logging.csproj" />
+    </ItemGroup>
+
+    <ItemGroup>
+      <PackageReference Include="Microsoft.Azure.Functions.Worker" Version="1.13.0" />
+      <PackageReference Include="Microsoft.Azure.Functions.Worker.Extensions.Http" Version="3.0.13" />
```
mattjohnsonpint, line 16: Please set these to the _minimum_ (non-preview) version that our library requires.  (Unless any are marked as having vulnerabilities.)

mattjohnsonpint: Probably this is `1.0.0` and `3.0.12`, but bump higher if those don't work.

src/Sentry.AzureFunctions.Worker/.editorconfig
SeanFeldman: Thank you for the elaborate comment. I was unaware of how `await` is applied with `ConfigureAwait()` in place. 
For Functions, I remember from the in-process SDK that there was no synchronization context. Isolated Worker SDK continued that. [This answer](https://stackoverflow.com/questions/55542016/is-configureawaitfalse-needed-beneficial-when-awaiting-async-calls-in-azure-fu) from Stephen Cleary mentions what I remembered. I can also try and get a concrete answer for the Isolated Worker SDK to confirm it.

SeanFeldman: Raised a [question](https://github.com/Azure/azure-functions-dotnet-worker/issues/1520).

samples/Sentry.Samples.AzureFunctions.Worker/local.settings.json
SeanFeldman: Part of the template. Running locally with HTTP and timer trigger, the setting is not needed. I'll remove it.

samples/Sentry.Samples.AzureFunctions.Worker/.gitignore
SeanFeldman: Oops 🙂

src/Sentry.AzureFunctions.Worker/SentryFunctionsWorkerMiddleware.cs
SeanFeldman: I didn't bother with that as I wanted to see if anything was coming soon. Since it's not, I'll add something, 👍

src/Sentry.AzureFunctions.Worker/SentryFunctionsWorkerMiddleware.cs
SeanFeldman: Ah, I missed the `transaction.Finish()` call. @mattjohnsonpint, if I get it right, a transaction should be finished regardless of the outcome. And if it's a failure, the transaction should be finished with the exception that was thrown. Did I get that part right?

SeanFeldman: Also, unlike ASP.NET Core, Functions middleware will handle multiple types of triggers. Should it still handle the [span status](https://github.com/getsentry/sentry-dotnet/blob/0110385baedc3527d03652b4b6a9dbd8fe459029/src/Sentry.AspNetCore/SentryTracingMiddleware.cs#L200) or not?

src/Sentry.AzureFunctions.Worker/SentryFunctionsWorkerMiddleware.cs
SeanFeldman: > Or maybe a better way to handle this would be to prevent them from being added in the first place?  We could make an option on `SentryLoggingOptions` for tag filters, and then use them in the `SentryLogger`.  Then our Azure Functions library could just set these as tag filters during initialization.

I actually like this idea a lot more. The concert for those tags shouldn't be with the Functions middleware. Plus, it could benefit other Sentry integrations that would like to filter out other tags, right?

So I'm assuming the actual filtering would happen in `SentryLogger` [here](https://github.com/getsentry/sentry-dotnet/blob/0110385baedc3527d03652b4b6a9dbd8fe459029/src/Sentry.Extensions.Logging/SentryLogger.cs#L104). Should this be part of this PR or do you think it should be a follow-up PR? I'm inclined to include it in this work.

SeanFeldman:
@mattjohnsonpint, the test I didn't know how to add was related to the two tags I'm removing (`AzureFunctions_FunctionName` and `AzureFunctions_InvocationId`). With the suggestion you've made, I guess I don't need to write that test anymore as long as it would get be added to the general `SentryLogger` tests to verify tags can be filtered out. So that would be a test I'll need to add to the `SentryLogger` test project.

src/Sentry.AzureFunctions.Worker/.editorconfig
mattjohnsonpint: Yeah, I'd still like to err on the side of safety and consistency.  For us that means, anything under `/src` should be using `ConfigureAwait(false)`.  It sounds like it won't hurt to have it either way.

Thanks for raising the question to confirm though. 👍

src/Sentry.AzureFunctions.Worker/SentryFunctionsWorkerMiddleware.cs
mattjohnsonpint: Yes, to both.

For an HTTP trigger, we should use the same approach as ASP.NET Core - taking the status from the outgoing HTTP response code.

For other trigger types, we can use `transaction.Finish()` on success, or `transaction.Finish(exception)` on failure.

src/Sentry.AzureFunctions.Worker/SentryFunctionsWorkerMiddleware.cs
mattjohnsonpint: It would be further up, [here](https://github.com/getsentry/sentry-dotnet/blob/0110385baedc3527d03652b4b6a9dbd8fe459029/src/Sentry.Extensions.Logging/SentryLogger.cs#LL91C43-L91C43) based on the property key (which becomes the tag name).

It's small enough, but since it's not specific to Functions, it should be in it's own PR so we can call it out as a separate feature.  You can either do it first and merge it in for usage here, or you can leave this one with the current approach and fix it when you do that later.

src/Sentry.AzureFunctions.Worker/Sentry.AzureFunctions.Worker.csproj
SeanFeldman: Can't use `Microsoft.Azure.Functions.Worker` `1.0.0` as it [didn't target `netstandard2.0`](https://www.nuget.org/packages/Microsoft.Azure.Functions.Worker/1.0.0#supportedframeworks-body-tab). The support for `netstandard2.0` was [added in `1.8.0`](https://www.nuget.org/packages/Microsoft.Azure.Functions.Worker/1.8.0#supportedframeworks-body-tab). But the `CancellationToken` support on `FunctionContext` was only added in [`1.9.0`](https://github.com/Azure/azure-functions-dotnet-worker/blame/1.9.0/src/DotNetWorker.Core/Context/FunctionContext.cs). Moving `Microsoft.Azure.Functions.Worker` to `1.9.0`.

src/Sentry.AzureFunctions.Worker/SentryFunctionsWorkerMiddleware.cs
SeanFeldman: Regarding the cached transaction name and the lookup key. Suppose I go with a `context.FunctionDefinition.PathToAssembly` + `context.FunctionDefinition.EntryPoint` + HTTP method combo; the result will look something like this:

```
<path>\sentry-dotnet\samples\Sentry.Samples.AzureFunctions.Worker\bin\Debug\net6.0\Sentry.Samples
.AzureFunctions.Worker.dll-Sentry.Samples.AzureFunctions.Worker.SampleHttpTrigger.RunAsync-GET
```

This feels somewhat redundant, @mattjohnsonpint. I think `context.FunctionDefinition.EntryPoint` + HTTP method combo should suffice, which will generate 

```
Sentry.Samples.AzureFunctions.Worker.SampleHttpTrigger.RunAsync-GET
```

`Sentry.Samples.AzureFunctions.Worker` - the assembly name containing the trigger
`SampleHttpTrigger.RunAsync` - the class name + the entry method
`GET` - the HTTP method.

What do you think, @mattjohnsonpint?

samples/Sentry.Samples.AzureFunctions.Worker/local.settings.json
SeanFeldman: > Do we need any storage for this sample?

Yes. The storage connection string is needed for the timer trigger. The timer uses a storage queue with a visibility timeout. Otherwise, the trigger doesn't work. It shouldn't be a problem, given that both VS2022 and Rider have support for Storage Emulator. Alternatively, there's Azurite that, is often used locally, with the identical development connection string `"UseDevelopmentStorage=true"`.

src/Sentry.AzureFunctions.Worker/SentryFunctionsWorkerMiddleware.cs
SeanFeldman: Note to self - remove `UnsetTags` and validate it's working.

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@
 - Build .NET Standard 2.1 for Unity ([#2328](https://github.com/getsentry/sentry-dotnet/pull/2328))
 - Add `RemoveExceptionFilter`, `RemoveEventProcessor` and `RemoveTransactionProcessor` extension methods on `SentryOptions` ([#2331](https://github.com/getsentry/sentry-dotnet/pull/2331))
 - Include Dynamic Sampling Context with error events, when there's a transaction ([#2332](https://github.com/getsentry/sentry-dotnet/pull/2332))
+- Azure Functions (Isolated Worker/Out-of-Process) support ([#2346](https://github.com/getsentry/sentry-dotnet/pull/2346))
```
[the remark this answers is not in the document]
SeanFeldman: @mattjohnsonpint, not sure how to resolve this. The version for `Sentry.AzureFunctions.Worker` is going to be `0.1.0-beta`. But the stable version is going to be different. How should the `CHANGELOG.md` be handled in such a case?

src/Sentry.AzureFunctions.Worker/SentryFunctionsWorkerMiddleware.cs
mattjohnsonpint: That's fine. Just go with the entrypoint. We can always adjust it later if we get any complaints.

CHANGELOG.md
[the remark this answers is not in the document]
SeanFeldman: For now fixed my error, moved the PR under `## Unreleased` section.

src/Sentry.AzureFunctions.Worker/SentryFunctionsWorkerMiddleware.cs
SeanFeldman: The necessary change will be completed in #2360

src/Sentry.AzureFunctions.Worker/SentryFunctionsWorkerApplicationBuilderExtensions.cs
```diff
@@ -0,0 +1,24 @@
...
+
+/// <summary>
+/// Sentry extension methods for Azure Functions with Isolated Worker SDK
+/// </summary>
+public static class SentryFunctionsWorkerApplicationBuilderExtensions
+{
+    /// <summary>
+    /// Configure Azure Functions to use Sentry
+    /// </summary>
+    public static IFunctionsWorkerApplicationBuilder UseSentry(this IFunctionsWorkerApplicationBuilder builder, Action<SentryOptions>? optionsConfiguration)
```
mattjohnsonpint, line 16:
```suggestion
    public static IFunctionsWorkerApplicationBuilder UseSentry(this IFunctionsWorkerApplicationBuilder builder, Action<SentryLoggingOptions>? optionsConfiguration)
```

src/Sentry.AzureFunctions.Worker/Sentry.AzureFunctions.Worker.csproj
```diff
@@ -0,0 +1,23 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+    <PropertyGroup>
+      <TargetFrameworks>net6.0;netstandard2.0</TargetFrameworks>
+      <PackageTags>$(PackageTags);Azure;Functions;Worker</PackageTags>
+      <Description>Official Azure Functions Worker SDK integration for Sentry - Open-source error tracking that helps developers monitor and fix crashes in real time.</Description>
+      <Version>0.1.0-beta</Version>
```
mattjohnsonpint, line 7:
```suggestion
      <Version>$(Version)-beta.1</Version>
```

src/Sentry.AzureFunctions.Worker/SentryFunctionsWorkerApplicationBuilderExtensions.cs
mattjohnsonpint, line 16:
```suggestion
    public static IFunctionsWorkerApplicationBuilder UseSentry(this IFunctionsWorkerApplicationBuilder builder, Action<SentryLoggingOptions>? optionsConfiguration)
```

src/Sentry.AzureFunctions.Worker/.editorconfig
mattjohnsonpint, line 1: remove

mattjohnsonpint:
Add API approval tests please.

samples/Sentry.Samples.AzureFunctions.Worker/Program.cs
```diff
@@ -0,0 +1,15 @@
+using Microsoft.Extensions.Hosting;
+using Sentry.AzureFunctions.Worker;
+
+var host = new HostBuilder()
+    .ConfigureFunctionsWorkerDefaults((host, builder) =>
```
SeanFeldman, line 5: @mattjohnsonpint, notice the change here - I had to use the overload with the `HostBuilder` and `AzureFunctionsBuilder` to get access to `IConfiguration`.

mattjohnsonpint (approved):

PR #2349, opened 2023-05-01
Improve debug file upload handling

### LINKED ISSUE #2260: Upload symbols from primary output (bin) folder, not intermediate output (obj) folder
There's a lot of intermediate debug files being uploaded to Sentry during build. From my past testing, these do not actually provide any value to symbolication and shouldn't be uploaded. See some more details here: https://github.com/getsentry/sentry-dotnet/pull/2191#pullrequestreview-1307895569

I propose we only upload final files from the Release folder during a build.

After this is resolved, enable the tests here: `test/sentry-cli-integration.Tests.ps1`

mattjohnsonpint:
- Upload from `bin` folder rather than `obj`.  This uploads fewer unnecessary files, and also ensures that symbols from dependencies are uploaded.  Fixes #2260.
- Complete the integration tests started in #2191 - they actually check the list of upload files now.
- Expand the tests to cover various combination of input flags and project types
- Upload source bundles separately from symbols so we can always see the source bundle being uploaded in the tests.  It also simplifies the implementation.

### DIFF
--- a/test/sentry-cli-integration.Tests.ps1
+++ b/test/sentry-cli-integration.Tests.ps1
@@ -50,9 +50,9 @@ BeforeAll {
     }
 }
 
-Describe 'CLI-integration' {
+$exe = ([RuntimeInformation]::IsOSPlatform([OSPlatform]::Windows) ? '.exe' : '')
 
-    $exe = [RuntimeInformation]::IsOSPlatform([OSPlatform]::Windows) ? '.exe' : '';
+Describe 'CLI-integration' {
 
     It "uploads symbols and sources for a console app build" {
         $result = DotnetBuild 'Sentry.Samples.Console.Basic' $True $True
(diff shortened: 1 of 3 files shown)


test/sentry-cli-integration.Tests.ps1
```diff
@@ -8,15 +10,23 @@ if (!(Test-Path env:CI ))
...
-    function DotnetBuild([string]$Sample)
+    function DotnetBuild([string]$Sample, [bool]$Symbols, [bool]$Sources, [string]$TargetFramework = '')
     {
         $rootDir = "$(Get-Item $PSScriptRoot/../../)"
 
+        if ($TargetFramework -ne '') {
+            $framework = @('-f', $TargetFramework)
+        } else {
+            $framework = ''
+        }
```
SeanFeldman, line 21: If we can control the version of PS on the build server, with PS 7 we could do
```
$framework = $TargetFramework -ne '' ? @('-f', $TargetFramework) : ''
```

SeanFeldman (approved):

PR #2351, opened 2023-05-02
Add Hint support

### LINKED ISSUE #1469: Add `Hint` support to beforeSend and Event processors

jamescrosswell:
Adds new APIs that provide callbacks containing `Hint` object:

- `SentryOptions.SetBeforeSend`
- `SentryOptions.SetBeforeSendTransaction`
- `SentryOptions.SetBeforeBreadcrumb`
- `ISentryEventProcessorWithHint`
- `ISentryTransactionProcessorWithHint`

NOTE: Previously the first three of those were properties: `BeforeSend`, `BeforeSendTransaction`, `BeforeBreadcrumb`.  These properties have been obsoleted in favor of methods, so that they can be overloaded when hints are desired.

Hints provide two specific functions:
- They allow you to add or remove attachments during any of the mentioned callbacks. (Resolves #1469)
- They allow you to provide arbitrary objects (such as context, etc.) as additional items at time of capture, that are later present in the callbacks - but _not_ sent to sentry.


jamescrosswell:
With respect to `BeforeBreadcrumb` support, I see the only place `BeforeBreadcrumb` gets used is in `Scope.AddBreadcrumb`. 

`SentryEvent`, `Transaction` and `TransactionTracer` implement `IHasBreadcrumbs`, which also has an `AddBreadcrumb` method... but none of those call `BeforeBreadcrumb` before adding a breadcrumb. I assume that's all by design but what is the rationale for being able to filter breadcrumbs before they get added to the Scope vs when they get added to the other IEventLike implementations?

mattjohnsonpint:
I fixed the iOS build issue.  The Cocoa SDK doesn't have hints yet, and as you pointed out in comments, the Android SDK doesn't really let us propagate hints.  But this is fine for now.  We can address later with some coordination with the other teams if we think it will add value.

mattjohnsonpint:
We need to do the same with `BeforeSendTransaction` that we did with `BeforeSend` and `BeforeBreadcrumb`.  That should be fairly straightforward.

But also we need to supply hints to the various "processor" abstractions:
- `ISentryEventProcessor`
- `ISentryTransactionProcessor`
- `ISentryEventExceptionProcessor`
- `SentryEventExceptionProcessor` (abstract class)

Preventing breaking changes is going to be tough with this one.  We can use default interface implementation, but it won't work on .NET Framework or .NET Standard 2.0, so we'll just not offer hints on those for now.  We can consolidate in the next major v4.0.0 release.

Example:

```csharp
#if NETFRAMEWORK || NETSTANDARD2_0
public interface ISentryEventProcessor
{
    SentryEvent? Process(SentryEvent @event);
}
#else
public interface ISentryEventProcessor : ISentryEventProcessorWithHint
{
    SentryEvent? Process(SentryEvent @event);
    SentryEvent? ISentryEventProcessorWithHint.Process(SentryEvent @event, Hint hint) => Process(@event);
}

public interface ISentryEventProcessorWithHint
{
    SentryEvent? Process(SentryEvent @event, Hint hint);
}
#endif
```

src/Sentry/Hint.cs
```diff
@@ -0,0 +1,143 @@
...
+namespace Sentry;
+
+/// <summary>
+/// A hint that can be provided when capturing a <see cref="SentryEvent"/> or adding a <see cref="Breadcrumb"/>.
+/// Hints can be used to filter or modify events or breadcrumbs before they are sent to Sentry.
+/// </summary>
+public class Hint : ICollection, IEnumerable<KeyValuePair<string, object?>>
+{
+    private readonly Dictionary<string, object?> _internalStorage = new();
+    private readonly List<Attachment> _attachments = new();
```
mattjohnsonpint, line 15: I think this class is really blurry, as it encapsulates both a dictionary of items and a list of attachments.  The interfaces it provides are only about the items.  So methods like `Clear`, `Count`, CopyTo`, etc. don't consider the attachments.

I'd rather go with a cleaner data structure:

```csharp
public class Hint
{
    public IDictionary<string, object?> Items { get; } = new Dictionary<string, object?>();
    public IList<Attachment> Attachments { get; } = new List<Attachment>();
}
```

It doesn't need to implement any interfaces of its own.  One will need to be more explicit when using it (ie. `hint.Items["foo"] = bar` - but I think that's ok.

test/Sentry.Tests/ScopeTests.cs
```diff
@@ -309,6 +309,27 @@ public void AddBreadcrumb__AddBreadcrumb_RespectLimits(int initialCount, int max
...
+            return breadcrumb;
+        });
+        var scope = new Scope(options);
+
+        // Act
+        var expectedHint = new Hint();
+        scope.AddBreadcrumb(new Breadcrumb(), expectedHint);
+
+        // Assert
+        receivedHint.Should().Equal(expectedHint);
```
mattjohnsonpint, line 330:
```suggestion
        Assert.Same(expectedHint, receivedHint);
```

or if you prefer

```suggestion
        receivedHint.Should().BeSameAs(expectedHint);
```

src/Sentry/Hint.cs
mattjohnsonpint: This also removes the need for us to have any of our own add/get/remove methods.

src/Sentry/Hint.cs
```diff
@@ -0,0 +1,143 @@
+using System;
+using System.Collections.Generic;
+using System.Net.Mail;
+using Sentry.Internal.Extensions;
+
```
mattjohnsonpint, line 5: Remove

```suggestion
```

src/Sentry/Hint.cs
jamescrosswell: Yeah I'm OK with that. I pretty much mimicked the Java implementation but agree, it's not clear what things like the "Clear" method do without inspecting their implementation...

jamescrosswell:
> I was thinking that we should use default interface methods for this, but I think a cleaner approach is to implement abstract classes instead. Specifically:
> 
> ```cs
> public abstract class SentryEventProcessor : ISentryEventProcessor
> {
>     SentryEvent ISentryEventProcessor.Process(SentryEvent @event)
>     {
>         throw new NotImplementedException();
>     }
> 
>     public abstract SentryEvent? Process(SentryEvent @event, Hint hint);
> }
> 
> public abstract class SentryTransactionProcessor : ISentryTransactionProcessor
> {
>     Transaction ISentryTransactionProcessor.Process(Transaction transaction)
>     {
>         throw new NotImplementedException();
>     }
> 
>     public abstract Transaction? Process(Transaction transaction, Hint hint);
> }
> ```

If the abstract class doesn't have any implemented methods, it's not doing anything for us that an interface wouldn't. 

Is this to avoid [breaking our public API contract](https://learn.microsoft.com/en-gb/dotnet/core/compatibility/library-change-rules)? I see that guidance includes:
* ❓ REQUIRES JUDGMENT: Adding a new interface implementation to a type
* ✔️ ALLOWED: Adding an abstract member to a public type that has no accessible (public or protected) constructors, or the type is [sealed](https://learn.microsoft.com/en-gb/dotnet/csharp/language-reference/keywords/sealed)

And would descendants of the abstract `Processor` classes be passed in as dependencies to constructors or relevant methods?

src/Sentry/Hint.cs
```diff
@@ -0,0 +1,88 @@
...
+    /// <summary>
+    /// Gets or sets a Screenshot for the Hint
+    /// </summary>
+    public Attachment? Screenshot { get; set; }
+
+    /// <summary>
+    /// Gets or sets a ViewHierarchy for the Hint
+    /// </summary>
+    public Attachment? ViewHierarchy { get; set; }
+
```
mattjohnsonpint, line 69: Let's remove these properties.  Reasons:

- They aren't part of the `Attachments` collection.
- They assume there can only be one screenshot or view hierarchy - but there could be multiple (multiple windows, multi-screen devices, etc.)
- We only have limited support for view hierarchy in the base sdk.  Presently only `sentry-unity` uses it.
- We don't have any support for screenshots in the base sdk yet.

jamescrosswell:
>But also we need to supply hints for event and transaction processors (not exception processors). 

@mattjohnsonpint this morning we discussed copying information (like attachments etc.) across from the Scope into the hint so that this would be available in event/transaction processes (and the Before* call backs). 

### Events

`ISentryEventProcessor.Process(SentryEvent @event)` gets called by the SentryClient when processing all the events. The call flow looks something like this:
```
SentryClient.CaptureEvent(SentryEvent? @event, Hint? hint, Scope? scope = null);
...
DoSendEvent(@event, hint, scope);
...
processor.Process(@event.Exception, @event);
```

So rather than copying everything over from the Scope to a Hint, it would probably be easier just to pass both the scope and the hint into the processors. 

The key difference between the `Hint` and the `Scope` appears to be that the `Scope` contains information that will be sent with the `SentryEvent` to the Sentry server... whereas a `Hint` might contain information that you wouldn't necessarily want to send (e.g. the `HttpResponseMessage` that we put in there on the FailedRequestHandler). 

So it seems we do need a Hint class to store that information (we can't put that kind of thing in the Scope). However I don't think we need to copy all of the information from the Scope over to the Hint before calling processors/callbacks. 

### Transactions

All of the above largely applies to Transactions as well. It has an extra layer of indirection. `ISentryTransactionProcessor` gets called by the Hub, which in turn just passes the grunt work off to the [current Scope/Client](https://github.com/getsentry/sentry-dotnet/blob/0f701eb0dabae0f5b2606f5c2fdecb143b3ec8fc/src/Sentry/Internal/Hub.cs#L413). The other difference with transactions is that SDK users aren't supposed to call `CaptureTransaction` directly so the only way Hints would end up in the Transaction processors or `BeforeSendTransaction` callback would be if our SDK provided these hints... and there may be use cases for this (information we want to provide to processors but that we don't want to put in the Scope because we're not sure if it's safe to send to the Sentry server). We could adopt the same approach with Transactions as we do with Events and just pass both the Scope and a Hint into the processors... that would be simple/consistent. 

### Breadcrumbs

`AddBreadcrumb` is a [method on the Scope](https://github.com/getsentry/sentry-dotnet/blob/0f701eb0dabae0f5b2606f5c2fdecb143b3ec8fc/src/Sentry/Scope.cs#L254), so obviously no reason to pass the scope in as a parameter to this. Although we could easily supply the scope as context to the `BeforeBreadcrumb` callback, if that might be useful.

### Final thoughts

An alternative to passing Scope and Hint into all of these processors/callbacks might be to hold the Scope as a property of the Hint. The current implementaions of `CaptureEvent` and `CaptureTransaction` accept a Scope parameter, and we wouldn't be able to get rid of that until v4 but eventually it could be deprecated in favour of methods taking a Hint parameter instead (that would hold the Scope plus other stuff that may or may not be safe to send to Sentry). I don't think that's a good idea though. It appeals to my OCD tendency to tidy up overloads and method signatures, but it couples Scope and Hint unnecessarily and doesn't do much for us... seems like that would just be fluffing around.

### How to move forward

For the time being, I'll assume we want to provide the `Scope` along with a `Hint` to processors & callbacks for Events, Transactions and Breadcrumbs. Let me know if you can see an issue with that approach though.

jamescrosswell:
> We need to do the same with `BeforeSendTransaction` that we did with `BeforeSend` and `BeforeBreadcrumb`. That should be fairly straightforward.
> 
> But also we need to supply hints for event and transaction processors (_not_ exception processors). I was thinking that we should use default interface methods for this, but I think a cleaner approach is to implement abstract classes instead. Specifically:
> 
> ```cs
> public abstract class SentryEventProcessor : ISentryEventProcessor
> {
>     SentryEvent ISentryEventProcessor.Process(SentryEvent @event)
>     {
>         throw new NotImplementedException();
>     }
> 
>     public abstract SentryEvent? Process(SentryEvent @event, Hint hint);
> }
> 
> public abstract class SentryTransactionProcessor : ISentryTransactionProcessor
> {
>     Transaction ISentryTransactionProcessor.Process(Transaction transaction)
>     {
>         throw new NotImplementedException();
>     }
> 
>     public abstract Transaction? Process(Transaction transaction, Hint hint);
> }
> ```
> 
> So if one wants hints in processors, they implement the abstract class. We can remove the current interfaces in the next major release (v4.0.0).

It seems like all the abstract base class is doing for us here is letting the `ISentryClient` know whether it should call `Process(SentryEvent @event)` or `Process(SentryEvent @event, Hint hint);`. 

I think I prefer an interface in this case. 
- People can use the interace without being forced to descend from our base class (more flexible)
- Our base class doesn't contain any useful logic (i.e. no better than an interface)

So I can see pros for interfaces and no cons. 

I'll go with an interface for the time, but open to being persuaded otherwise if there are good reasons to go with an abstract base class.

src/Sentry/Hint.cs
jamescrosswell: Done.

jamescrosswell:
I think I've taken this as far as I can. Let me know when you get a chance to review... and @mattjohnsonpint would be great to get your wisdom on the CI Build error since the most recent merge from `main`.

mattjohnsonpint (approved):

src/Sentry/SentryClient.cs
```diff
@@ -276,9 +276,9 @@ private SentryId DoSendEvent(SentryEvent @event, Hint? hint, Scope? scope)
             return SentryId.Empty;
         }
 
-        return CaptureEnvelope(Envelope.FromEvent(processedEvent, _options.DiagnosticLogger, scope.Attachments, scope.SessionUpdate))
-            ? processedEvent.EventId
-            : SentryId.Empty;
+        var attachments = hint.Attachments.ToList();
+        var envelope = Envelope.FromEvent(processedEvent, _options.DiagnosticLogger, attachments, scope.SessionUpdate);
+        return CaptureEnvelope(envelope) ? processedEvent.EventId : SentryId.Empty;
```
jamescrosswell, line 281: @mattjohnsonpint just so I understand... are attachments that get added to hints supposed to be sent to Sentry?

PR #2354, opened 2023-05-08
Require XML comments on public types for all build configurations

mattjohnsonpint:
We used to only require XML comments for public types be authored while in release configuration, but that makes it easy to miss them at development time.  Waiting for CI to fail is more friction than just requiring the docs in the first place, so restoring to default behavior.

#skip-changelog

### DIFF
--- a/src/Directory.Build.props
+++ b/src/Directory.Build.props
@@ -6,9 +6,6 @@
     <!--Generate xml docs for all projects under 'src'-->
     <GenerateDocumentationFile>true</GenerateDocumentationFile>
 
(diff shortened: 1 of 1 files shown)


SeanFeldman (approved):
Should we also turn it on in `.editorconfig`?

`dotnet_diagnostic.CS1591.severity = error`

mattjohnsonpint:
> Should we also turn it on in `.editorconfig`?
> 
> `dotnet_diagnostic.CS1591.severity = error`

I don't think that's needed.  It already errors without it.

image

PR #2364, opened 2023-05-11
Allow setting the active span on the scope

### LINKED ISSUE #1679: Option to link Http Request Spans to the Root Transaction?
### Package

Sentry.AspNetCore

### .NET Flavor

.NET Core

### .NET Version

6.0.0

### OS

Windows

### SDK Version

3.17.1

_No response_

### Steps to Reproduce

Sample project: https://github.com/lucas-zimerman/httpclient-stack-sample
- run and once you get information on the browser an event will be sent to `Sentry SDKs` org on the .NET project.

- but it should be replicable if you call an httpclient in parallel inside of a request. 

### Expected Result

Looking at the code, you may expect that each span generated on the ForEach loop would have the parentId as `GetAnimals`
image

### Actual Result

https://sentry.io/organizations/sentry-sdks/discover/sentry-dotnet:dc044830d1994a2685d81dd22eb3cb7c/?field=title&field=event.type&field=project&field=user.display&field=timestamp&name=All+Events&query=&sort=-timestamp&statsPeriod=1h&yAxis=count%28%29

image
Due to the defined behavior the HttpClient will try to get the last opened span and due to the parallel behavior, you'll end up with multiple `GetDog` spans inside of each other.

One idea is to set an option that affects spans not generated by the user where you can define if the parent will be the last opened Span or always the scope Transaction, (or maybe even an user defined logic?).

EDIT: A brainstorm. Spans could have a parameter that would hide it from `GetSpan`, let's name it `AllowChildren`. in specific cases like a http request, we could easily assume it's the last span from a stack so we set `AllowChildren` as false on HttpClient spans, with that, when you call `GetSpan` you'll always get the last open span that could have a children.
The same idea could be applied to the DiagnosticSource Spans or even into parallel code from users.

### LINKED ISSUE #1845: Allow defining the "Active Span"
### Problem Statement

Integrations such as `SentryHttpMesageHandler` rely on

```
_hub.GetSpan()?.StartChild
```

to create a span. This on the Hub today is wired as:

```
        public ISpan? GetSpan() => Transaction?.GetLastActiveSpan() ?? Transaction;
```

This means I can't set a specific span to get returned from `GetSpan`. When parallelizing work, the parent/child relationship becomes messy since there are multiple spans "active" and relying on the root transaction isn't possible to find out which one it is.

### Solution Brainstorm

When we're propagating the span explicitly downstream. A user can push a scope and set that new span as the active one. So integrations can find the correct parent span.

For example `_hub.SetActiveSpan(span)`

mattjohnsonpint:
With this change,  the `GetSpan` method on the `Scope` class has been obsoleted, and replaced with a `Span` property that has both a getter and a setter.

Previous usage:

```csharp
{
    var span = scope.GetSpan();
});
```

New usage:

```csharp
{
    var span = scope.Span;
});
```

That's the boring part.  The new interesting part is that you can now _set_ the span if desired.  This can be used to pin an active span as the root span before starting complex tasks such as doing work in parallel.  This allows each subsequent child span to attach to the root, rather than to each other.

For example:

```csharp
var span = SentrySdk.GetSpan();
Parallel.ForEach(items, item =>
{
    SentrySdk.WithScope(scope =>
    {
        scope.Span = span;

        // Do some work that might generate new child spans, etc.
        // They'll all be created from the span set on the scope.
    });
});
```

Or asynchronously:

```csharp
var span = SentrySdk.GetSpan();
await Parallel.ForEachAsync(items, async (item, cancellationToken) =>
{
    await SentrySdk.WithScopeAsync(async scope =>
    {
        scope.Span = span;

        // Do some async work that might generate new child spans, etc.
        // They'll all be created from the span set on the scope.
    });
});
```

Or asynchronously with tasks:

```csharp
var span = SentrySdk.GetSpan();
await Task.WhenAll(items.Select(async item =>
{
    await SentrySdk.WithScopeAsync(async scope =>
    {
        scope.Span = span;

        // Do some async work that might generate new child spans, etc.
        // They'll all be created from the span set on the scope.
    });
}));
```

Resolves #1845
Resolves #1679

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@
 ### Features
 
 - Add tag filters to SentryLoggingOptions ([#2360](https://github.com/getsentry/sentry-dotnet/pull/2360))
+- Allow setting the active span on the scope ([#2364](https://github.com/getsentry/sentry-dotnet/pull/2364))
+  - Note: Obsoletes the `Scope.GetSpan` method in favor of a `Scope.Span` property (which now has a setter as well).
 
 ### Dependencies
 
--- a/src/Sentry/Internal/Hub.cs
+++ b/src/Sentry/Internal/Hub.cs
@@ -182,11 +182,7 @@ internal class Hub : IHubEx, IDisposable
         _ = ExceptionToSpanMap.GetValue(exception, _ => span);
     }
 
-    public ISpan? GetSpan()
-    {
-        var (currentScope, _) = ScopeManager.GetCurrent();
-        return currentScope.GetSpan();
-    }
+    public ISpan? GetSpan() => ScopeManager.GetCurrent().Key.Span;
 
     public SentryTraceHeader? GetTraceHeader() => GetSpan()?.GetTraceHeader();
 
@@ -286,7 +282,7 @@ internal class Hub : IHubEx, IDisposable
         }
 
         // Otherwise just get the currently active span on the scope (unless it's sampled out)
-        if (scope.GetSpan() is { IsSampled: not false } span)
+        if (scope.Span is { IsSampled: not false } span)
         {
             return span;
         }
--- a/src/Sentry/Scope.cs
+++ b/src/Sentry/Scope.cs
@@ -516,10 +516,34 @@ public class Scope : IEventLike, IHasDistribution
     }
 
     /// <summary>
-    /// Gets the currently ongoing (not finished) span or <c>null</c> if none available.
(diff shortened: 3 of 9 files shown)


mattjohnsonpint:
Using the example from #1679, without this feature the spans would sometimes link to each other (because more than one parallel request is active at a time).

image

With this feature, the request spans are all rooted correctly:

image

SeanFeldman (approved):

Issue #2078, opened 2022-12-06
Sanitize sensitive data from URLs sent to Sentry


mattjohnsonpint:
Per https://github.com/getsentry/team-webplatform-meta/issues/2 and https://github.com/getsentry/team-mobile/issues/70, ensure all captured URLs are sanitized according to the new spec in https://github.com/getsentry/develop/pull/773 (when it is ready)

## Description

When we make HTTP requests to third party services we create a breadcrumb with the URL and also create a span that has the URL as a description (and also breadcrumbs including this URL are created.)

Make sure that all integrations that record outgoing or incoming HTTP request structure the data like described in the spec linked above.

### References

* [RFC 0038 - Scrubbing sensitive data](https://github.com/getsentry/rfcs/blob/main/text/0038-scrubbing-sensitive-data.md)
* [Data Handling - Structure Data (Sentry Docs)](https://develop.sentry.dev/sdk/data-handling/#structuring-data)
* https://github.com/getsentry/sentry-java/pull/2366
* https://github.com/getsentry/sentry-python/pull/1876

mattjohnsonpint:
Rendered spec: https://develop.sentry.dev/sdk/data-handling/

Issue #773, opened 2021-01-29
Description is nullable but ctor param is not


marandaneto:
as we agreed, Description is nullable, but ctor requires a non nullable value

https://github.com/getsentry/sentry-dotnet/blob/8bc640f1c5873d9acc10d8a1b05c4ee58e2d4fc2/src/Sentry/Protocol/SpanContext.cs#L21
https://github.com/getsentry/sentry-dotnet/blob/8bc640f1c5873d9acc10d8a1b05c4ee58e2d4fc2/src/Sentry/Protocol/SpanContext.cs#L37

I believe we should match them, or was it intended?

Tyrrrz:
Ah, on the call I thought we were talking about `TraceContext` (I keep confusing all of these things).
This actually looks like an oversight, it should be nullable on the constructor.

PR #2365, opened 2023-05-12
Remove authority from URLs sent to Sentry

jamescrosswell:
Fixes https://github.com/getsentry/sentry-dotnet/issues/2078

# Description

Any URLs that are captured by the SentrySDK are sanitized per the spec (see refs below) before being sent to Sentry. 

When we make HTTP requests to third party services we create a breadcrumb with the URL and also create a span that has the URL as a description. Additionally, our failed outbound HTTP requests containing UserInfo can be captured and sent to Sentry - PII in these also needs to be redacted. 

Make sure that all integrations that record outgoing or incoming HTTP request structure the data like described in the 
spec linked above.

#### References

* [Specification: Better scrubbing of sensitive data #773](https://github.com/getsentry/develop/pull/773)
* [RFC 0038 - Scrubbing sensitive data](https://github.com/getsentry/rfcs/blob/main/text/0038-scrubbing-sensitive-data.md)
* [Data Handling - Structure Data (Sentry Docs)](https://develop.sentry.dev/sdk/data-handling/#structuring-data)
* https://github.com/getsentry/sentry-java/pull/2366
* https://github.com/getsentry/team-webplatform-meta/issues/2
* https://github.com/getsentry/team-mobile/issues/70
* https://github.com/getsentry/sentry-dotnet/issues/2369 

## Approach

The `SentryHttpMessageHandler` does things like this to add a breadcrumb:
```
        var breadcrumbData = new Dictionary<string, string>
        {
            {"url", url},
            {"method", method},
            {"status_code", ((int) response.StatusCode).ToString()}
        };
        _hub.AddBreadcrumb(string.Empty, "http", "http", breadcrumbData);
```

`SentryHttpMessageHandler` also creates a span as follows:
```
       // Start a span that tracks this request
        // (may be null if transaction is not set on the scope)
        // e.g. "GET https://example.com"
        var span = _hub.GetSpan()?.StartChild("http.client", $"{method} {url}");
```

Breadcrumbs and various other bits of data reside on the `Scope` most of the time. This scope data get's ***applied*** 
to `IEventLike` objects (which includes `Transaction` and `SentryEvent`) before they get sent to Sentry by the
`SentryClient`.

Spans are created directly on the `Transaction`. 

To ensure Personally Identifiable Information (PII) does not get with SentryEvents, if `SendDefaultPii == false` we 
redact any PII in the `SentryClent.DoSendEvent` method, just before events are sent to Sentry.

Similarly, we redact PII from Transactions in the `SentryClient.CaptureTransaction` method just before the transaction
is stuffed in an Envelope.

Both `SentryEvent` and `Transaction` now have an `internal void Redact()` method that redacts any PII data in their 
member fields or recursively calls `Redact()` on any complex members (e.g. `SentryEvent.Breadcrumbs` and 
`Transaction.Breadcrumbs`). 

---

## Java implementation

The [Java SDK implementation](https://tinyurl.com/3zy7v9yy) has a [urlWithAuthRemoved](https://tinyurl.com/fusud7vr) 
method they use to filter out the Auth portion of URLs:

```
  private static final @NotNull Pattern AUTH_REGEX = Pattern.compile("(.+://)(.*@)(.*)");

  private static @NotNull String urlWithAuthRemoved(final @NotNull String url) {
    final @NotNull Matcher userInfoMatcher = AUTH_REGEX.matcher(url);
    if (userInfoMatcher.matches() && userInfoMatcher.groupCount() == 3) {
      final @NotNull String userInfoString = userInfoMatcher.group(2);
      final @NotNull String replacementString =
          userInfoString.contains(":") ? "[Filtered]:[Filtered]@" : "[Filtered]@";
      return userInfoMatcher.group(1) + replacementString + userInfoMatcher.group(3);
    } else {
      return url;
    }
  }
```

And [here are the unit tests](https://tinyurl.com/4e49vn86).


SeanFeldman:
> It feels like this would be a cleaner and more robust approach. It'll take a bit more effort up front but I suspect it will be more scalable and save us a bunch of effort further down the road.

I understand that the first option may not be ideal, but I have concerns that using a serializer approach would result in dual responsibility, sanitization and serialization, which could be difficult to manage.

src/Sentry/Sentry.csproj
```diff
@@ -92,35 +92,17 @@
...
+      <SentryCLIDownload Condition="'$(CI_PUBLISHING_BUILD)' == 'true' Or ($([MSBuild]::IsOSPlatform('Linux')) And $(_OSArchitecture) == 'X86')" Include="sentry-cli-Linux-i686" FileHash="142824382e394f06a1a9b0bcb2276dd29fed17310fc01a24706724e6979749ff" />
+      <SentryCLIDownload Condition="'$(CI_PUBLISHING_BUILD)' == 'true' Or ($([MSBuild]::IsOSPlatform('Linux')) And $(_OSArchitecture) == 'X64')" Include="sentry-cli-Linux-x86_64" FileHash="8200b8f0831535d5c21adfde947ca6d30930619eae36a650cbcf1005c68cd6dd" />
+      <SentryCLIDownload Condition="'$(CI_PUBLISHING_BUILD)' == 'true' Or ($([MSBuild]::IsOSPlatform('Windows')) And $(_OSArchitecture) == 'X86')" Include="sentry-cli-Windows-i686.exe" FileHash="718209c411305dc4d86676dba6bef09e7c69bce03ce457c09f3b05ae08e1126e" />
+      <SentryCLIDownload Condition="'$(CI_PUBLISHING_BUILD)' == 'true' Or ($([MSBuild]::IsOSPlatform('Windows')) And $(_OSArchitecture) != 'X86')" Include="sentry-cli-Windows-x86_64.exe" FileHash="4ee6d34cce58bed122acc1602d327c28c9b82fa9e64bcc73100012e644e6537a" />
```
SeanFeldman, line 101: These shouldn't be changing, right?

jamescrosswell: They don't need to. Rider seems to want to format them as one line each. We could:
1. Accept Rider's reformating (assuming it doesn't conflict with VS)... which is what I did.
2. Manually ignore this changed file every time we try to check in the solution
3. Try to find some way to disable this in Rider's config.

jamescrosswell:
> I have concerns that using a serializer approach would result in dual responsibility, sanitization and serialization

Yeah, that would be a fair call if the JsonSerializer was actually a performing serialization. In fact all it's doing is creating a `UtfJsonWriter` and passing that into the `WriteTo` method of whatever `IJsonSerializable` instance it's dealing with ([see here](https://github.com/getsentry/sentry-dotnet/blob/3ec26ffd4287bc15c0ea049ab150e5a8deee8f7b/src/Sentry/Protocol/Envelopes/JsonSerializable.cs#L27)). The `WriteTo` method gets the json writer to serialize each of it's members. So the real serializer is the `UtfJsonWriter`. 

I think I'll probably just add another method to the IJsonSerializable interface with an overload of the [WriteTo method](https://github.com/getsentry/sentry-dotnet/blob/3ec26ffd4287bc15c0ea049ab150e5a8deee8f7b/src/Sentry/Transaction.cs#L302) that accepts a `bool sanitizeData` parameter or something.

jamescrosswell:
> I understand that the first option may not be ideal, but I have concerns that using a serializer approach would result in dual responsibility, sanitization and serialization, which could be difficult to manage.

@SeanFeldman finally I think you were right about this. When I dug into it futher, most of the serialization happens in a bunch of extension methods we've written in `JsonSerializableExtensions`... so I've reverted to the original approach.

test/Sentry.Tests/Protocol/SentryEventTests.cs
```diff
@@ -80,4 +82,88 @@ public void Modules_Getter_NotNull()
...
+            ServerName = serverName,
+            Release = release,
+            Distribution = distribution,
+            TransactionName = transactionName,
+            Request = new Request()
+            {
+                Method   = "GET",
+                Url = requestUrl
+            },
+            User = new User()
```
jamescrosswell, line 124: @mattjohnsonpint should these three `User` properties also be redacted? e.g. bob@foo.com -> [Filtered]@foo.com?

Same again the in the Transaction (which also has a user)...

src/Sentry/Internal/PiiExtensions.cs
```diff
@@ -0,0 +1,55 @@
...
+    internal const string RedactedText = "[Filtered]";
+
+    /// <summary>
+    /// Searches for URLs in text data and redacts any PII data from these, as required.
+    /// </summary>
+    /// <param name="data">The data to be searched</param>
+    /// <returns>
+    /// The data, if no PII data is present or a copy of the data with PII data redacted otherwise
+    /// </returns>
+    public static string? RedactUrl(this string? data)
```
mattjohnsonpint, line 18: Generally, prefer to let the _caller_ deal with nullability on extension methods.

```suggestion
    public static string RedactUrl(this string data)
```

So usage would be like `foo = foo?.RedactUrl()` when `foo` is a `string?`, or `foo = foo.RedactUrl()` if it's just a `string`.

mattjohnsonpint: ... then update usages.

src/Sentry/Internal/PiiExtensions.cs
```diff
@@ -0,0 +1,55 @@
...
+    private static string RedactAuth(string data)
+    {
+        // ^ matches the start of the string. (?i)(https?://) gives a case-insensitive matching of the protocol.
+        // (.*@) matches the username and password (authentication information). (.*)$ matches the rest of the URL.
+        var userInfoMatcher = new Regex(@"^(?i)(https?://)(.*@)(.*)$").Match(data);
+        if (userInfoMatcher.Success && userInfoMatcher.Groups.Count == 4)
+        {
+            var userInfoString = userInfoMatcher.Groups[2].Value;
+            var replacementString = userInfoString.Contains(":") ? "[Filtered]:[Filtered]@" : "[Filtered]@";
+            return userInfoMatcher.Groups[1].Value + replacementString + userInfoMatcher.Groups[3].Value;
```
mattjohnsonpint, line 48: Why not using `regex.Replace` here?

src/Sentry/Request.cs
```diff
@@ -1,4 +1,5 @@
 using Sentry.Extensibility;
+using Sentry.Internal;
```
mattjohnsonpint, line 2:
```suggestion
```

mattjohnsonpint (changes requested):

src/Sentry/Breadcrumb.cs
```diff
@@ -39,7 +50,17 @@ public sealed class Breadcrumb : IJsonSerializable
...
     /// </remarks>
-    public IReadOnlyDictionary<string, string>? Data { get; }
+    public IReadOnlyDictionary<string, string>? Data
+    {
+        get => _sendDefaultPii
+            ? _data
+            : _data?.ToDictionary(
+                x => x.Key,
+                x => x.Value?.RedactUrl()
+                )!
```
SeanFeldman, line 60: Curious why do we have to specify that the redacted result is not null if the returned type is a nullable dictionary.

SeanFeldman (approved):

src/Sentry/Breadcrumb.cs
mattjohnsonpint: That was because of the nullability on the extension method.  Now that's been updated, you can simplify the usage here to:

```csharp
x => x.Value.RedactUrl());
```

mattjohnsonpint (approved):

PR #2367, opened 2023-05-12
Move tags filtering from SentryLogger to Scope

SeanFeldman:
Move tags filtering from SentryLogger to Scope.
By doing so, tags that are set via `Scope.SetTag(key, value)` can be filtered out.


mattjohnsonpint (approved):

mattjohnsonpint:
@SeanFeldman - This feature needs docs. 🙂

SeanFeldman:
> This feature needs docs. 🙂

On its way 🙂 https://github.com/getsentry/sentry-docs/pull/7000

PR #2368, opened 2023-05-13
Fix `EnableTracing` option conflict with `TracesSampleRate`

### LINKED ISSUE #2366: Conflict between EnableTracing and TracesSampleRate
### Package

Sentry

### SDK Version

3.31.0

### Steps to Reproduce

```csharp
var options = new SentryOptions();
options.TracesSampleRate = options.TracesSampleRate;
options.EnableTracing = true;
Console.WriteLine($"Traces Sample Rate: {options.TracesSampleRate}");
```

### Expected Result

```
Traces Sample Rate: 1
```

### Actual Result

```
Traces Sample Rate: 0
```

Assigning a property to itself should generally be a no-op, especially for options.  So the `EnableTracing = true` should set the sample rate to `1.0`.

This is a problem when configuration binding occurs, such as with an `appsettings.json` file, when used with the ASP.NET Core integration.  During binding, some properties may get reset to their default values.  So even if you've set `EnableTracing=True`, tracing might have been disabled.

mattjohnsonpint:
Fixes #2366 by having `TracesSampleRate` and `EnableTracing` properties have simple get/set implementations (except for some existing validation).  This issue was mainly seen during configuration binding in ASP.NET Core applications, and then - only in the case where `EnableTracing` was set `true` and `TracesSampleRate` was unset.

Unfortunately, the only way to implement the correct fix was to make `TracesSampleRate` a nullable `double?`.  Previously, it was a non-nullable `double`.  (The default value is now `null` instead of `0.0`.)  Normally I'd consider that to be a breaking change that could not be done without a major version bump.  But in this case, it's unlikely to affect anyone, because this property is typical _set_, but never retrieved externally.

There are no behavioral changes.  Both values work as they were originally intended.

I also improved the xmldocs, and fixed the exception type thrown when an invalid sample rate is set. (`ArgumentOutOfRangeException`, is more appropriate than `InvalidOperationException` for that).

### DIFF
--- a/test/Sentry.Tests/SentryOptionsTests.cs
+++ b/test/Sentry.Tests/SentryOptionsTests.cs
@@ -1,8 +1,3 @@
-#if NETFRAMEWORK
-using Sentry.PlatformAbstractions;
-using Xunit.Sdk;
-#endif
-
 namespace Sentry.Tests;
 
 public class SentryOptionsTests
@@ -43,6 +38,20 @@ public class SentryOptionsTests
     }
 
     [Fact]
+    public void TracesSampleRate_Default_Null()
+    {
+        var sut = new SentryOptions();
+        Assert.Null(sut.TracesSampleRate);
+    }
+
+    [Fact]
+    public void TracesSampler_Default_Null()
+    {
+        var sut = new SentryOptions();
+        Assert.Null(sut.TracesSampler);
+    }
+
+    [Fact]
     public void IsTracingEnabled_Default_False()
     {
         var sut = new SentryOptions();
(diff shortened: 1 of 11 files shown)


SeanFeldman (approved):

test/Sentry.Tests/SentryOptionsTests.cs
```diff
@@ -50,39 +59,119 @@ public void IsTracingEnabled_Default_False()
...
-    public void EnableTracing_WhenTrue()
+    public void IsTracingEnabled_TracesSampleRate_Zero()
     {
         var sut = new SentryOptions
         {
-            EnableTracing = true
+            TracesSampleRate = 0.0
+        };
+
+        Assert.False(sut.IsTracingEnabled);
```
jamescrosswell, line 91: It seems like the SentryOptions.EnableTracing property is redundant... tracing could be implicitly disabled by setting `SentryOptions.TracesSampleRate` to `0` and implicitly enabled by setting `SentryOptions.TracesSampleRate` to something `> 0`. 

Under the hood, maybe we don't need a backing variable for EnableTracing property? Possibly something for v4?

SeanFeldman: I think `EnableTracing` was added later, after `TracesSampleRate` was there.

mattjohnsonpint: Yes. It is specifically redundant, and done across all SDKs. Intended to be a simplified shorthand for enabling tracing without thinking about sample rate.

PR #2379, opened 2023-05-19
Update CI Solution Filters

mattjohnsonpint:
Follow-up from  #2346 and #2206 to update the solution filters.

Also addresses some minor test fixes we would have caught had the filter changes been in place.

#skip-changelog


SeanFeldman (approved):
Thank you for taking care of that approval 🙂

Issue #2316, opened 2023-04-19
.NET Profiling for ASP.NET Core


bruno-garcia:
Follow up from:

* https://github.com/getsentry/sentry-dotnet/issues/2315
* https://github.com/getsentry/sentry-dotnet/issues/1955

* [ ] collect `Microsoft-Windows-DotNETRuntime` rundown provider (loaded libs, etc) in a separate session and merge with actual samples later during processing, if possible (merging nettrace files, or having a custom provider for TraceLog). 
* [ ] Or maybe we don't even need to collect these because we're in the same process? can we hack our way around and get the libraries from the current process instead? Probably not so easy because it also somehow stack walking if I'm not mistaken.
* [ ] We'll use GlobalMode to set the behavior: all threads in process (global Mode) or only activity/request (server mode)
* [ ] Consider capturing only the current "activity", i.e. not all threads but only what belongs to the transaction.

ericsampson:
Very exciting. How long do you think it'll be before external users are able to give it a go, assuming that we acknowledge that it's "alpha/beta caveat emptor" :)

mattjohnsonpint:
First we need to get through #2315, which is blocked by https://github.com/microsoft/perfview/issues/1829

Then we can better estimate this one.  It will likely be a while.

bruno-garcia:
Blazor Wasm profiling available on .NET 10, I wonder if we can leverage that.
Also:
* https://github.com/getsentry/sentry-dotnet/issues/3523#issuecomment-2877930809

PR #2385, opened 2023-05-22
profiling streaming TraceLog

vaind:
Addresses multiple issues with our current TraceLog integration:
1. originally, we processed profiles by writing them to a file and then loading that, processing events and removing - this PR ensures all events are processed as they happen
2. originally, we started a new DiagnositcsClient profiler session on each transaction - this is very slow (100+ms) so this PR changes the logic to have a single profiler session running all the time and attach/detach our processing logic for each transaction

Additionally, this PR fixes some bugs in the profiler & adds it as a submodule, with a ProjectReference dependency on it. Currently, the submodule points to https://github.com/microsoft/perfview/pull/1867

#skip-changelog because this isn't public-facing


mattjohnsonpint:
Looking forward to getting this in.  Thanks so much for your diligence! 🎉

bruno-garcia:
branched from this into this: https://github.com/getsentry/sentry-dotnet/pull/2555/

benchmarks/Sentry.Benchmarks/ProfilingBenchmarks.cs
```diff
@@ -125,9 +139,9 @@ public void DiagnosticsSessionStartCopyStop(bool rundown, string provider)
     {
         EventPipeProvider[] providers = provider switch
         {
-            "runtime" => new[] { new EventPipeProvider("Microsoft-Windows-DotNETRuntime", EventLevel.Informational, (long)ClrTraceEventParser.Keywords.Default) },
+            "runtime" => new[] { new EventPipeProvider("Microsoft-Windows-DotNETRuntime", EventLevel.Informational, (long)BenchmarkDotNetTransientTraceEvent::Microsoft.Diagnostics.Tracing.Parsers.ClrTraceEventParser.Keywords.Default) },
```
bruno-garcia, line 142: We'll need to fix this in a follow up PR. Since our customers will experience the same thing. We need to change namespaces, or make stuff internal

src/Sentry.Profiling/SampleProfilerSession.cs
```diff
@@ -1,70 +1,85 @@
...
-        new EventPipeProvider("Microsoft-DotNETCore-SampleProfiler", EventLevel.Informational),
-        new EventPipeProvider("System.Threading.Tasks.TplEventSource", EventLevel.Informational, (long) TplEtwProviderTraceEventParser.Keywords.Default)
+        // TODO replace Keywords.Default with a subset. Currently it is:
+        //   Default = GC | Type | GCHeapSurvivalAndMovement | Binder | Loader | Jit | NGen | SupressNGen
+        //                | StopEnumeration | Security | AppDomainResourceManagement | Exception | Threading | Contention | Stack | JittedMethodILToNativeMap
+        //                | ThreadTransfer | GCHeapAndTypeNames | Codesymbols | Compilation,
+        new EventPipeProvider(ClrTraceEventParser.ProviderName, EventLevel.Informational, (long) ClrTraceEventParser.Keywords.Default),
+        new EventPipeProvider(SampleProfilerTraceEventParser.ProviderName, EventLevel.Informational),
+        // new EventPipeProvider(TplEtwProviderTraceEventParser.ProviderName, EventLevel.Informational, (long) TplEtwProviderTraceEventParser.Keywords.Default)
```
bruno-garcia, line 40: This isnt' needed now because there's no file involved?
```suggestion
```

src/Sentry.Profiling/Sentry.Profiling.csproj
```diff
@@ -5,14 +5,12 @@
...
     <Description>Performance profiling support for Sentry - Open-source error tracking that helps developers monitor and fix crashes in real time.</Description>
-    <IsPackable>false</IsPackable>
   </PropertyGroup>
 
   <ItemGroup>
     <ProjectReference Include="..\..\src\Sentry\Sentry.csproj" />
-
     <PackageReference Include="Microsoft.Diagnostics.NETCore.Client" Version="0.2.421201" />
-    <PackageReference Include="Microsoft.Diagnostics.Tracing.TraceEvent" Version="3.1.2" />
+    <ProjectReference Include="../../modules/perfview/src/TraceEvent/TraceEvent.csproj" />
```
bruno-garcia, line 13: I imagine we'll need to figure out distributing this, also OK on a follow up PR since profilign isn't packaged yet

src/Sentry/Internal/FastSerialization/GrowableArray.cs
```diff
@@ -347,11 +347,6 @@ private void Realloc(int minSize)
     // Implementation of IEnumerable<T>.
     public IEnumerator<T> GetEnumerator() => new GrowableArrayEnumerator(this);
 
-    internal void Add(object value)
```
bruno-garcia, line 350: This GrowableArray is part of the `perfview` module. Can we use that implementation isntead?

src/Sentry/Internal/SparseArray.cs
```diff
@@ -34,6 +36,6 @@ public SparseScalarArray(T uninitializedValue, int capacity = 0)
 
     public bool ContainsKey(int key)
     {
-        return key > 0 && key < _items.Count && !_uninitializedValue.Equals(_items[key]);
+        return key >= 0 && key < _items.Count && !_uninitializedValue.Equals(_items[key]);
```
bruno-garcia, line 39: Is this bug int he original implementation?

vaind: yes and there were no tests (fixed that as well of course)

src/Sentry.Profiling/Sentry.Profiling.csproj
vaind: yes, that was my thinking too

src/Sentry.Profiling/SampleProfilerSession.cs
vaind: There were some experiments with this but it didn't give that much value in the past. However, the idea is that it would become necessary for [profiling individual activities](#2316). As is, it cannot be used because the "activity computer" in perfview (that was used to handle these events) isn't yet updated to support streaming event pipe. It would have made the perfview PR too large to hope for a merge so I didn't do it yet. Hoping to get back to it once all this shakes down.

bruno-garcia:
Added 11f2b34a9ec24ddabad6e45ccbc09bb4c9244edd to deal with: https://github.com/microsoft/perfview/issues/1900

If we need more packages from that feed, we need to opt-in by adding the specific name to the `packageSource` node

benchmarks/Sentry.Benchmarks/ProfilingBenchmarks.cs
vaind: related to https://github.com/getsentry/sentry-dotnet/pull/2385#discussion_r1301976435

src/Sentry/Internal/FastSerialization/GrowableArray.cs
vaind: We can't because it's only depended up in `Sentry.Profiling` project and this is part of `Sentry` (so it can be used directly in protocol classes)

bruno-garcia (approved):
I'd say this is good to merge, and we have follow-ups. It's still not getting packed for NuGet so we are free to iterate on main

Issue #2247, opened 2023-03-22
Support Native AOT on .NET 7+


mattjohnsonpint:
## Problem Statement

.NET 7 introduced ["Native AOT"](https://learn.microsoft.com/dotnet/core/deploying/native-aot/) deployments for Windows and Linux and .NET 8 adds Native AOT support for macOS.
- This is different than ".NET Native", that is used with UWP
- It is also different than the Mono AOT compilation used on .NET iOS apps

### Initial investigation

Testing on Windows, a .NET console app with Sentry publishing with Native AOT gives these warnings:

```
C:\Users\mattj\.nuget\packages\sentry\3.29.1\lib\net6.0\Sentry.dll : warning IL2104: Assembly 'Sentry' produced trim warnings. For more information see https://aka.ms/dotnet-illink/libraries [C:\Users\mattj\Code\Temp\AotConsoleApp\AotConsoleApp.csproj]
C:\Users\mattj\.nuget\packages\sentry\3.29.1\lib\net6.0\Sentry.dll : warning IL3053: Assembly 'Sentry' produced AOT analysis warnings. [C:\Users\mattj\Code\Temp\AotConsoleApp\AotConsoleApp.csproj]
```

Trying to capture an exception shows internal serialization failures related AOT trimming:

```
Error: Failed to serialize object for property 'Dynamic Code'. Original depth: 2, current depth: 2
System.NotSupportedException: 'System.Text.Json.Serialization.Converters.DictionaryOfTKeyTValueConverter`3[System.Collections.Generic.Dictionary`2[System.String, System.Boolean], System.String, System.Boolean]' is missing native code or metadata. This can happen for code that is not compatible with trimming or AOT. Inspect and fix trimming and AOT related warnings that were generated when the app was published. For more information see https://aka.ms/nativeaot-compatibility
   at System.Reflection.Runtime.General.TypeUnifier.WithVerifiedTypeHandle(RuntimeConstructedGenericTypeInfo, RuntimeTypeInfo[]) + 0x88
   at System.Text.Json.Serialization.Converters.IEnumerableConverterFactory.CreateConverter(Type, JsonSerializerOptions) + 0x6a2
   at System.Text.Json.Serialization.JsonConverterFactory.GetConverterInternal(Type, JsonSerializerOptions) + 0x15
   at System.Text.Json.JsonSerializerOptions.ExpandConverterFactory(JsonConverter, Type) + 0x32
[...]

Issue #2144, opened 2023-01-30
DB Connection spans presented poorly


mattjohnsonpint:
### SDK Version

3.26.2

### Steps to Reproduce

Using code from NugetTrends [here](https://github.com/dotnet/nuget-trends/blob/fc196aee7904b9747a58605b0bf37f115444d1e7/src/NuGetTrends.Scheduler/DailyDownloadWorker.cs#L232-L299), or from scratch, use EF Core to send multiple update queries to the database within one db context.

### Expected Result

The recorded spans should be easy to interpret, and should correctly reflect connection time vs query time.

### Actual Result

We capture a new connection span for each query, though there is likely only one real connection.  We never close the connection span, so it appears that they are running forever (in reality, they close automatically with the transaction)

![image](https://user-images.githubusercontent.com/1396388/215621720-9c54fd0d-f20b-4124-b7a9-c66076067f0b.png)

mattjohnsonpint:
There clearly is a bug here, in that we should be closing the connection span.  But even then, it will still be confusing because there's actually not a separate connection per query.  Rather, EF Core and SQL Client are pooling the connections in the background.

More importantly the only thing we really are interested in seeing with regard to connection is any slowness caused when actually connecting to the database.  Thus, I propose changing how we show the connection span.  Rather than it being the parent of the query, it should be a sibling that comes first.  But also, it shouldn't be the time from connection connected to closed, but rather the time from _connecting_ to _connected_.

J-Swift:
Just came across this when trying to figure out how to interpret `db.connection` vs `db.query`. Would really appreciate this being more straightforward to interpret.

jamescrosswell:
I'm not able to reproduce this. This is what I'm seeing using the same DB context to issue multiple SQL Queries in a single controller method (each with it's own span):
![image](https://github.com/getsentry/sentry-dotnet/assets/728212/b9711244-3835-4dbb-8769-ce045ac4b793)

pinkpawkitty:
@jamescrosswell Those traces seem to be missing the ef core traces.
Yours are called "db query", I assume written by you? The ef core traces are called "db.query" and contain the exeuted SQL as a description

jamescrosswell:
Hm... I had three kinds of queries being run in the transaction above:

1. `_context.Database.SqlQuery`
2. `_context.Database.ExecuteSqlAsync`
3. An `IQueriable` constructed by appending some conditions (like Take(10)) to one of the EF models

[...]

PR #2398, opened 2023-05-29
Fix/db spans not finishing

jamescrosswell:
## Root cause
Sentry listens for EF related diagnostic data in the `SentryEFCoreListener` class. Typically those events come in pairs - e.g. `"Microsoft.EntityFrameworkCore.Database.Connection.ConnectionOpening"` and `"Microsoft.EntityFrameworkCore.Database.Connection.ConnectionClosed"`. 

Previously, when we received one of the "opening" events we added a span to the `TransactionTrace` and stored a weak reference to this in `SentryEFCoreListener` (using `AsyncLocal<WeakReference<ISpan>>`). However for some reason the `ConnectionClosed` event was causing the thread context to change, so in the context that those events were processed, the `AsyncLocal` would be holding a null reference... meaning we couldn't track down the original `ISpan` and finish it. 

## High level solution

We no longer store references to the spans that get added in the `SentryEFCoreListener` class. Instead we use the correlation ids that are provided with the diagnostic events. 

The `ConnectionOpening` event contains a `ConnectionId` that we store in `ISpan.Extra`.  `ConnectionClosed` events also contain a `ConnectionId`, so when we receive these, we can extract this, find the span with the matching `ConnectionId` and finish it.

For queries, the solution is very similar except that the correlation id is called `CommandId`. 

**Note:** For `QueryCompilationStarting`, `QueryModelCompiling` and `QueryExecutionPlanned` there is no appropriate correlation id and it also wasn't possible to use the description of the `QueryExpressionEventData` (since this differs between the opening and closing events). In that specific instance, we simply run a FIFO... so the first expressions to start being compiled are assumed to be the first to finish. 

## SQL Listener

The SQL Listener implementation was already using the correlation id technique so didn't suffer from this problem. I haven't touched that. It's worth noting that the implementation of `SentrySqlListener` is slightly different in that it all sits in a single class... whereas for Entity Framework, the logic has been factored out into helper classes. That's because I built most of the solution for the problem with EF before I checked how it was done in `SentrySqlListener`. In hindsight, that was a mistake... If we think this is a major issue, we could rewrite one or the other of these implementations so that they both used the same code pattern.

## Result

DB Spans are now finishing correctly!

image


mattjohnsonpint:
We won't be able to take a package dependency on `Microsoft.EntityFrameworkCore.Relational`, because it would require anyone using Sentry to also be using EFCore.  We don't currently publish a separate EFCore-specific package where we can take such dependencies.

If all we need is the `ConnectionId`, then consider getting it through reflection - which won't require any dependencies.

```csharp
private Guid? ConnectionId =>
    DiagnosticSourceValue?.GetType().FullName == "Microsoft.EntityFrameworkCore.Diagnostics.ConnectionEventData"
        ? DiagnosticSourceValue.GetGuidProperty("ConnectionId")
        : null;
```

mattjohnsonpint:
... Though I'm surprised that we can't get this value directly from diagnostics data.  Are you sure it isn't exposed some other way?

mattjohnsonpint:
If we _do_ have to resort to reflection, then that is ok for now, but would be a strong reason to make a separate `Sentry.EntryFrameworkCore` package when we bump to the next major release - because we need to remove reflection to for Native AOT support (#2247).

jamescrosswell:
> We won't be able to take a package dependency on `Microsoft.EntityFrameworkCore.Relational`, because it would require anyone using Sentry to also be using EFCore.

Aha... makes sense. 

>Though I'm surprised that we can't get this value directly from diagnostics data. Are you sure it isn't exposed some other way

The only place I've seen it is in the `Value` of the KeyPair we're getting from the Diagnostic events... and that value is a type that is defined in `Microsoft.EntityFrameworkCore.Relational`. 

Can you think of somewhere else we'd get this?

>If all we need is the ConnectionId, then consider getting it through reflection - which won't require any dependencies.

The reflection idea is smart... that means our solution will work for all target frameworks as well! 

The values we need are:
1. `ConnectionEventData.ConnectionId` (to correlate start/finish on connection events)
2. `CommandEventData.ConnectionId` and `CommandEventData.CommandId` (to identify the parent and correlate start/finish on command events)
3. [Possibly] `QueryExpressionEventData.EventId` to correlate start/finish on compilation events... but need to test (it wasn't 100% clear from the docs whether this would be suitable).

Just have to be cautious that those types may have changed over time between different releases of EF - which we don't detect via static typing when using Reflection. Need to make sure we test it with different versions of EF then (or dig through the history of that repo on GitHub to check if there have been any changes that would break our reflection).

src/Sentry.DiagnosticSource/Internal/DiagnosticSource/EFDiagnosticSourceHelper.cs
```diff
@@ -0,0 +1,117 @@
...
+    ///    makes the resulting waterfall diagram hard to read.
+    /// 2. Sentry has a performance to errors feature which detects n + 1 problems on OM frameworks... but this only
+    ///    works if all of the "n" queries have the same parent span in the transaction. Again, since queries ofen run
+    ///    on their own connection, if we wrap each query in a connection span, it breaks the performance error handling
+    /// </summary>
+    private ISpan GetParentSpan(ITransaction transaction) => transaction.GetDbParentSpan();
+
+    protected virtual void SetSpanReference(ISpan span)
+    {
+        Options.LogDebug($"No Span reference found when adding {Operation} Span.");
```
jamescrosswell, line 109: `LogWarning` instead?

src/Sentry/Sentry.csproj
mattjohnsonpint, line 1: Undo unrelated changes please.

src/Sentry.DiagnosticSource/Internal/DiagnosticSource/SentryEFCoreListener.cs
```diff
@@ -52,85 +45,54 @@ public SentryEFCoreListener(IHub hub, SentryOptions options)
...
+        new(_hub, _options, _spansQueryLocal, diagnosticsSourceValue);
 
     public void OnNext(KeyValuePair<string, object?> value)
     {
         try
         {
+            // Because we have to support the .NET framework, we can't get at strongly typed diagnostic source events.
+            // We do know they're objects, that can be converted to strings though... and we can get the correlation
+            // data we need from there by parsing the string. Not as reliable, but works against all frameworks.
+            var diagnosticSourceValue = value.Value?.ToString();
```
mattjohnsonpint, line 68: This variable is unused.

src/Sentry.DiagnosticSource/Internal/DiagnosticSource/EFCommandDiagnosticSourceHelper.cs
```diff
@@ -0,0 +1,48 @@
...
+internal class EFCommandDiagnosticSourceHelper : EFDiagnosticSourceHelper
+{
+    internal EFCommandDiagnosticSourceHelper(IHub hub, SentryOptions options, AsyncLocal<WeakReference<ISpan>> spanLocal, object? diagnosticSourceValue)
+        : base(hub, options, diagnosticSourceValue)
+    {
+    }
+
+    protected override string Operation => "db.query";
+    protected override string Description => FilterNewLineValue(DiagnosticSourceValue) ?? string.Empty;
+    private Guid? ConnectionId => DiagnosticSourceValue?.GetGuidProperty("ConnectionId");
```
mattjohnsonpint, line 15: `ConnectionId` is unused

src/Sentry.DiagnosticSource/Internal/DiagnosticSource/EFConnectionDiagnosticSourceHelper.cs
```diff
@@ -0,0 +1,47 @@
+using Sentry.Internal.Extensions;
+
+namespace Sentry.Internal.DiagnosticSource;
+
+// ReSharper disable once InconsistentNaming
+internal class EFConnectionDiagnosticSourceHelper : EFDiagnosticSourceHelper
+{
+    internal EFConnectionDiagnosticSourceHelper(IHub hub, SentryOptions options, AsyncLocal<WeakReference<ISpan>> spanLocal, object? diagnosticSourceValue)
```
mattjohnsonpint, line 8: `spanLocal` is unused

src/Sentry.DiagnosticSource/Internal/DiagnosticSource/EFCommandDiagnosticSourceHelper.cs
```diff
@@ -0,0 +1,48 @@
+using Sentry.Internal.Extensions;
+
+namespace Sentry.Internal.DiagnosticSource;
+
+// ReSharper disable once InconsistentNaming
```
mattjohnsonpint, line 5: I don't think these ReSharper comments are needed.  They flag as unneeded in Rider on my machine anyway.  Maybe just update abbreviations locally if they're flagging for you.  (same on other classes.)

src/Sentry.DiagnosticSource/Internal/DiagnosticSource/EFConnectionDiagnosticSourceHelper.cs
```diff
@@ -0,0 +1,47 @@
+using Sentry.Internal.Extensions;
```
mattjohnsonpint, line 1: import not needed

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 
 - SentryHttpMessageHandler added when AddHttpClient is before UseSentry ([#2390](https://github.com/getsentry/sentry-dotnet/pull/2390))
 - Set the native sdk name for Android ([#2389](https://github.com/getsentry/sentry-dotnet/pull/2389))
+- Fix/db spans not finishing ([#2398](https://github.com/getsentry/sentry-dotnet/pull/2398))
```
mattjohnsonpint, line 9:
```suggestion
- Fix db connection spans not finishing ([#2398](https://github.com/getsentry/sentry-dotnet/pull/2398))
```

src/Sentry.DiagnosticSource/Internal/DiagnosticSource/EFQueryCompilerDiagnosticSourceHelper.cs
```diff
@@ -0,0 +1,28 @@
+using Sentry.Internal.Extensions;
+
+namespace Sentry.Internal.DiagnosticSource;
+
+// ReSharper disable once InconsistentNaming
+internal class EFQueryCompilerDiagnosticSourceHelper : EFDiagnosticSourceHelper
+{
+    internal EFQueryCompilerDiagnosticSourceHelper(IHub hub, SentryOptions options, AsyncLocal<WeakReference<ISpan>> spanLocal, object? diagnosticSourceValue)
```
mattjohnsonpint, line 8: `spanLocal` is unused

src/Sentry.DiagnosticSource/Internal/DiagnosticSource/EFDiagnosticSourceHelper.cs
```diff
@@ -0,0 +1,117 @@
+using Sentry.Extensibility;
+
+namespace Sentry.Internal.DiagnosticSource;
+
+// ReSharper disable once InconsistentNaming
+internal abstract class EFDiagnosticSourceHelper
+{
+    internal const string ConnectionExtraKey = "db.connection_id";
+    internal const string CommandExtraKey = "db.command_id";
+    internal const string CompiledQueryExtraKey = "db.compiled_query";
```
mattjohnsonpint, line 10: `CompiledQueryExtraKey` is unused

src/Sentry.DiagnosticSource/Internal/DiagnosticSource/EFDiagnosticSourceHelper.cs
```diff
@@ -0,0 +1,117 @@
...
+    /// will both have the same parent as those of <see cref="EFConnectionDiagnosticSourceHelper"/>.
+    ///
+    /// We've done this for two reasons:
+    /// 1. We could show these underneath the relevant connection, but each command often gets it's own connection which
+    ///    makes the resulting waterfall diagram hard to read.
+    /// 2. Sentry has a performance to errors feature which detects n + 1 problems on OM frameworks... but this only
+    ///    works if all of the "n" queries have the same parent span in the transaction. Again, since queries ofen run
+    ///    on their own connection, if we wrap each query in a connection span, it breaks the performance error handling
+    /// </summary>
+    private ISpan GetParentSpan(ITransaction transaction) => transaction.GetDbParentSpan();
```
mattjohnsonpint, line 105: Other than a place to put the comment, I'm not sure this function is adding anything useful.  If we need to keep it, then it can simplify to:

```suggestion
    private static ISpan GetParentSpan(ISpan span) => span.GetDbParentSpan();
```

But I think it could just be inlined.  There's only one usage.

src/Sentry.DiagnosticSource/Internal/DiagnosticSource/EFDiagnosticSourceHelper.cs
mattjohnsonpint: I think it would be better to move these to the child classes so the messages can be more specific.  "Span reference" isn't really it - it's "Connection ID" or "Command ID".

Then they can be warnings, yes.

These methods could be made abstract then also.

src/Sentry.DiagnosticSource/Internal/DiagnosticSource/SentryEFCoreListener.cs
```diff
@@ -52,85 +45,54 @@ public SentryEFCoreListener(IHub hub, SentryOptions options)
...
             switch (value.Key)
             {
                 // Query compiler span
                 case EFQueryStartCompiling or EFQueryCompiling:
-                    AddSpan(SentryEFSpanType.QueryCompiler, "db.query.compile", FilterNewLineValue(value.Value));
+                    QueryCompilerDiagnosticSourceHelper(value.Value).AddSpan();
```
mattjohnsonpint, line 73: These factory methods allocate new instances of the helper classes every time they are called.  That's going to be a lot of allocations under load.  Please refactor to either use static helpers, or a single shared instance.

I don't think there is state shared in the helper, because we're getting a different helper instance for each diagnostic event.  Right?

src/Sentry.DiagnosticSource/Internal/DiagnosticSource/EFCommandDiagnosticSourceHelper.cs
```diff
@@ -0,0 +1,48 @@
...
+                    TryGetCommandId(span) == commandId)
+            : base.GetSpanReference(transaction);
+
+    protected override void SetSpanReference(ISpan span)
+    {
+        if (CommandId is { })
+        {
+            SetCommandId(span, CommandId);
+            return;
+        }
```
mattjohnsonpint, line 45:
```suggestion
        if (CommandId is { } commandId)
        {
            SetCommandId(span, commandId);
            return;
        }
```

mattjohnsonpint (changes requested):
See feedback inline.  Thanks.

mattjohnsonpint:
The approach seems sound.  Thanks for the work here!

Looking at the screenshot, it seems we are now placing connections and commands side-by-side in a sibling relationship rather than parent-child.  That's good, I think.  However I'm not sure why it looks like we have a different connection ID every time. If it's connection pooling, then shouldn't the connection ID be the same on each one?  And if so, then we could eliminate the extra ones and just show one connection span at the top with all the related command spans under it?

Also, what's this gap about?

image

src/Sentry.DiagnosticSource/Internal/DiagnosticSource/EFDiagnosticSourceHelper.cs
```diff
@@ -0,0 +1,117 @@
...
+    /// <param name="value">the query to be parsed value</param>
+    /// <returns>the filtered query</returns>
+    internal static string? FilterNewLineValue(object? value)
+    {
+        var str = value?.ToString();
+        return str?[(str.IndexOf('\n') + 1)..];
+    }
+
+    /// <summary>
+    /// Finds an appropriate parent for EF diagnostic spans. Note that in our implementaion we're "flattening" these.
```
mattjohnsonpint, line 94:
```suggestion
    /// Finds an appropriate parent for EF diagnostic spans. Note that in our implementation we're "flattening" these.
```

src/Sentry.DiagnosticSource/Internal/DiagnosticSource/EFDiagnosticSourceHelper.cs
```diff
@@ -0,0 +1,117 @@
...
+    /// Spans from <see cref="EFQueryCompilerDiagnosticSourceHelper"/> and <see cref="EFCommandDiagnosticSourceHelper"/>
+    /// will both have the same parent as those of <see cref="EFConnectionDiagnosticSourceHelper"/>.
+    ///
+    /// We've done this for two reasons:
+    /// 1. We could show these underneath the relevant connection, but each command often gets it's own connection which
+    ///    makes the resulting waterfall diagram hard to read.
+    /// 2. Sentry has a performance to errors feature which detects n + 1 problems on OM frameworks... but this only
+    ///    works if all of the "n" queries have the same parent span in the transaction. Again, since queries ofen run
```
mattjohnsonpint, line 102:
```suggestion
    /// 2. Sentry has a performance to errors feature which detects n + 1 problems on ORM frameworks... but this only
    ///    works if all of the "n" queries have the same parent span in the transaction. Again, since queries often run
```

mattjohnsonpint:
Just to clarify on my previous comment...

We used to have:

- `db.query.compile`
- `db.connection` - 1111111
  - `db.query`
- `db.connection` - 2222222
  - `db.query`
- `db.connection` - 3333333
  - `db.query`

... and the connections were hanging open.

Now it's _better_ with:

- `db.query.compile`
- `db.connection` - 1111111
- `db.query`
- `db.connection` - 2222222
- `db.query`
- `db.connection` - 3333333
- `db.query`

But what we're aiming for is more like:

- `db.query.compile`
- `db.connection` - 1111111
- `db.query`
- `db.query`
- `db.query`

... where the same connection id is used for the three queries (via pooling), and the logical connection span ends when all the queries end - even if the connection is still technically open, but returned to the pool.

src/Sentry.DiagnosticSource/Internal/DiagnosticSource/EFCommandDiagnosticSourceHelper.cs
jamescrosswell: Hm, no idea how to configure that. I'll remove the comments and just learn to live with the squiggly lines I think.

src/Sentry.DiagnosticSource/Internal/DiagnosticSource/EFConnectionDiagnosticSourceHelper.cs
jamescrosswell: That one is needed for the `GetGuidProperty` extension that's used to read the ConnectionId.

jamescrosswell:
> it seems we are now placing connections and commands side-by-side in a sibling relationship rather than parent-child. That's good, I think. 

True... that was intentional ;-) The `SentrySqlListener` was already doing that so this means we now have consistency between both the `SentrySqlListener` and the `SentryEFCoreListener`.

> But what we're aiming for is more like:

Yeah, I think we can do that. Looking at what comes through to Sentry, all of the connections have the same ConnectionId. Something to look at as part of https://github.com/getsentry/sentry-dotnet/issues/2144 I think.

mattjohnsonpint (approved):

mattjohnsonpint:
Thanks!!!

PR #2404, opened 2023-06-01
Reduce debug files uploaded

mattjohnsonpint:
When `<SentryUploadSymbols>` is `true`, we use `sentry-cli` to upload _any_ valid debug info file from the output directory.  Unfortunately, that also picks up every `.so`, `.dll` and `.exe` because they can contain native stack unwind details and symbol tables.  For most .NET projects, we only need the `.pdb` files.  We also need the `dsym` files (`.dylib`, etc.) on iOS/MacCatalyst.

This PR updates our msbuild target to only include the symbol files that are needed per platform.


mattjohnsonpint:
Test failed in CI.  Will investigate.

SeanFeldman (approved):

SeanFeldman (approved):

mattjohnsonpint:
Note, the tests don't really show it, but on a WinUI app, this reduces from hundreds of debug files down to just a handful - because there are lots of native `.dll` files in the output directory.

PR #2424, opened 2023-06-14
[Spike] file based lease

SeanFeldman:
Relates to: https://github.com/getsentry/sentry-dotnet/issues/1067

### DIFF
--- a/test/Sentry.Tests/Internals/FileSystemTests.cs
+++ b/test/Sentry.Tests/Internals/FileSystemTests.cs
@@ -6,19 +6,6 @@ public class FileSystemTests
 {
     private const string Filename = "lock.txt";
 
-    [Fact(Skip = "Run manually if needed")]
-    public void Can_get_lease_after_it_was_consumed()
-    {
-        FileSystem.Instance.DeleteFile(Filename);
-
-        _ = FileSystem.Instance.GetLeaseFile(Filename);
-
-        Assert.Throws<IOException>(() =>
-        {
-            using (FileSystem.Instance.GetLeaseFile(Filename)) { }
-        });
-    }
-
     [Fact]
     public void Leased_file_cannot_be_leased_again()
     {
@@ -35,5 +22,35 @@ public class FileSystemTests
         {
             using (fileSystem.GetLeaseFile(path)) { }
         });
+    }
+
+    [Fact]
+    public void RealFileSystem_Leased_file_is_accessible_after_process_crush()
+    {
+        FileSystem.Instance.DeleteFile(Filename);
+
+        var thread = new Thread(() =>
+        {
+            _ = FileSystem.Instance.GetLeaseFile(Filename);
+        });
+
+        thread.Start();
+        thread.Join(200);
+
+        var keepTrying = true;
+
+        while (keepTrying)
+        {
+            try
+            {
+                keepTrying = false;
+
+                _ = FileSystem.Instance.GetLeaseFile(Filename);
+            }
+            catch (IOException)
+            {
+                // Immediate access might fail, retry
+            }
+        }
     }
 }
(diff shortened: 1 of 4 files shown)


test/Sentry.Tests/Internals/FileSystemTests.cs
```diff
@@ -0,0 +1,56 @@
...
+        Assert.Throws<IOException>(() =>
+        {
+            using (fileSystem.GetLeaseFile(path)) { }
+        });
+    }
+
+    [Fact]
+    public void RealFileSystem_Leased_file_is_accessible_after_process_crush()
+    {
+        FileSystem.Instance.DeleteFile(Filename);
```
bruno-garcia, line 30: We're going to have this test run concurrently once per each TFM in the test project.
Ideally we use a `guid` as a file name here and make sure we delete the file in a `finally` block in the test to avoid leaving trash behind

test/Sentry.Tests/Internals/FileSystemTests.cs
```diff
@@ -0,0 +1,56 @@
...
+        IDictionary<string, MockFileData> files  = new Dictionary<string, MockFileData>
+        {
+            { path, new MockFileData(Filename) { AllowedFileShare = FileShare.None } }
+        };
+
+        IFileSystem fileSystem = new FakeFileSystem(files);
+
+        Assert.Throws<IOException>(() =>
+        {
+            using (fileSystem.GetLeaseFile(path)) { }
```
bruno-garcia, line 23: I don't understand what we're testing here if this file system is getting mocked. Could you please expand?

test/Sentry.Tests/Internals/FileSystemTests.cs
```diff
@@ -0,0 +1,56 @@
...
+    {
+        FileSystem.Instance.DeleteFile(Filename);
+
+        var thread = new Thread(() =>
+        {
+            _ = FileSystem.Instance.GetLeaseFile(Filename);
+        });
+
+        thread.Start();
+        thread.Join(200);
```
bruno-garcia, line 38: Could this just be?

```suggestion
Task.Run(() => FileSystem.Instance.GetLeaseFile(Filename)).Wait();
```

test/Sentry.Tests/Internals/FileSystemTests.cs
```diff
@@ -0,0 +1,56 @@
...
+
+        var keepTrying = true;
+
+        while (keepTrying)
+        {
+            try
+            {
+                keepTrying = false;
```
bruno-garcia, line 46: If we're flipping this flag here, do we ever retry?

test/Sentry.Tests/Internals/FileSystemTests.cs
SeanFeldman: The direction was originally to verify this against TestFileSystem to avoid what was described as issues in the past.
Sorry, I don't have the full context. I did look into this, and this test was an attempt to see if it's feasible, but it may not help as the OS is the one that releases a file, and a real OS would be needed.

test/Sentry.Tests/Internals/FileSystemTests.cs
SeanFeldman: In theory, yes. I wanted to ensure it's a different thread. IIRC, tasks don't have to be on a different thread. They could be on the same thread, which would defeat the purpose of the validation.

test/Sentry.Tests/Internals/FileSystemTests.cs
SeanFeldman: 🤦‍♂️ 
It should have been after the attempt to get the lease 😀

Issue #2362, opened 2023-05-10
Debug meta info not sent when using `PublishSingleFile=true`


mattjohnsonpint:
### Package

Sentry

### .NET Flavor

.NET

### .NET Version

7.0.2013

### OS

Any (not platform specific)

### SDK Version

3.31.0

### Steps to Reproduce

Create a simple console app:

```csharp
using Sentry;

SentrySdk.Init(options =>
{
    options.Dsn = "...";
    options.Debug = true;
});

try
{
    throw new Exception("Test");
}
catch (Exception exception)
{
    SentrySdk.CaptureException(exception);
}
```

In the csproj, configure to upload symbols and sources to Sentry (and authenticate with `sentry-cli login`).
```xml
<PropertyGroup>
    <SentryOrg>...</SentryOrg>
    <SentryProject>...</SentryProject>
    <SentryUploadSymbols>true</SentryUploadSymbols>
    <SentryUploadSources>true</SentryUploadSources>
</PropertyGroup>
```

Compile and publish with:
```
dotnet publish -c Release -p:PublishSingleFile=true
```

Run the app from its published folder:
```
./bin/Release/net7.0/osx-arm64/publish/MyConsoleApp
```

### Expected Result

The event generated and shown in the console debug output should contain a `debug_meta` section, including the `debug_id`.

In Sentry, the source context should be visible, and the debug images section should show the symbols were found.

### Actual Result

The event is missing the `debug_meta` section when `-p:PublishSingleFile=true` is used.  Thus, source context is not shown.

Line numbers will still be shown if the `.pdb` file is present (which it is by default in the `publish` folder), but if you delete it - or ship the executable app without the pdb file, then client-side symbolication won't occur.  Server-side symbolication also won't occur because the `debug_meta` section is missing from the event.

mattjohnsonpint:
The code that resolves the assembly debug information is here:

https://github.com/getsentry/sentry-dotnet/blob/00299c7a950f7229b31153db5cff82001527b0d7/src/Sentry/Internal/DebugStackTrace.cs#L413-L422

https://github.com/getsentry/sentry-dotnet/blob/00299c7a950f7229b31153db5cff82001527b0d7/src/Sentry/Internal/DebugStackTrace.cs#L380-L388

I presume that we're not able to read the assembly directly from a file when that file is a single-file-executable.  We need to figure out how to get a `PEReader` in that environment.

[...]

PR #2425, opened 2023-06-15
Symbolication for Single File Apps

jamescrosswell:
## Background research

There's quite a bit of background in the details of https://github.com/getsentry/sentry-dotnet/issues/2362.  

Specifically the following comments:
- https://github.com/getsentry/sentry-dotnet/issues/2362#issuecomment-1580178118
- https://github.com/getsentry/sentry-dotnet/issues/2362#issuecomment-1588849442
- https://github.com/getsentry/sentry-dotnet/issues/2362#issuecomment-1589120319

## High level solution

We've taken the minimum possible subset of code from ILSpy that we need for our purposes. The stuff that was copied from or inspired by ILSpy is sitting in the `/src/Sentry/Internal/ILSpy/` folder along with an [ATTRIBUTION.txt](https://github.com/getsentry/sentry-dotnet/blob/34b096f73c5d8bb623d23749a52f2c921bbbeb64/src/Sentry/Internal/ILSpy/ATTRIBUTION.txt) file. 

`SingleFileBundle.cs` was copied almost verbatim (changed namespaces and a couple of other things to satisfy our linters). 

`SingleFileApp.cs` we've adapted from the `LoadedPackage.cs`, `LoadedAssembly.cs` and `PEFile.cs` classes in ILSpy. All of those would come with a web of dependencies we don't want or need.

## Testing

The `/test/SingleFileTestApp/SingleFileTestApp.csproj` file includes a `<PublishSingleFile>true</PublishSingleFile>` property, which tells msbuild to publish this as a single file app. Ordinarily that would only happen when you run `dotnet publish` but we've also configured `DefaultTargets="Build;Publish"`... which means that whenever you run `dotnet build` or (importantly) `dotnet test` the single file test app gets published. As such, we get our Single File App whenever running our tests in CI (hooray!). 

Additionally we set the `RuntimeIdentifier` that is used when publishing the `SingleFileTestApp` based on the OS/Architecture of the machine running the tests... so we can test this on various different platforms.

```
  <PropertyGroup>
    <_OSArchitecture>$([System.Runtime.InteropServices.RuntimeInformation]::OSArchitecture)</_OSArchitecture>
    <!-- Set the appropriate RuntimeIdentifier. See: https://learn.microsoft.com/en-us/dotnet/core/rid-catalog -->
    <RuntimeIdentifier Condition="$([MSBuild]::IsOSPlatform('OSX')) And ($(_OSArchitecture) == 'Arm64')">osx-arm64</RuntimeIdentifier>
    <RuntimeIdentifier Condition="$([MSBuild]::IsOSPlatform('OSX')) And ($(_OSArchitecture) == 'X64')">osx-x64</RuntimeIdentifier>
    <RuntimeIdentifier Condition="$([MSBuild]::IsOSPlatform('Linux')) And ($(_OSArchitecture) == 'Arm64')">linux-arm64</RuntimeIdentifier>
    <RuntimeIdentifier Condition="$([MSBuild]::IsOSPlatform('Linux')) And ($(_OSArchitecture) == 'X86')">linux-x86</RuntimeIdentifier>
    <RuntimeIdentifier Condition="$([MSBuild]::IsOSPlatform('Linux')) And ($(_OSArchitecture) == 'X64')">linux-x64</RuntimeIdentifier>
    <RuntimeIdentifier Condition="$([MSBuild]::IsOSPlatform('Windows')) And ($(_OSArchitecture) == 'X86')">win-x86</RuntimeIdentifier>
    <RuntimeIdentifier Condition="$([MSBuild]::IsOSPlatform('Windows')) And ($(_OSArchitecture) != 'X86')">win-x64</RuntimeIdentifier>
  </PropertyGroup>
```

Our tests in `Sentry.Tests.Internals.ILSpy` then use some similar logic at runtime to determine where to find and load this single file app from.

### Remaining challenges

I haven't yet found an easy way to target multiple frameworks for this project so currently it only targets `net6.0`. 

The way we typically target multiple frameworks is to set the `<TargetFrameworks>` MSBuild property... and that works, when building. When publishing, however, that property is ignored - publish needs a `<TargetFramework>` property. 

There might be a way to do this with some "after build" hook or something, but this is probably only really important if we want to support (and test) .NET Core 3.1 as that does single file apps a bit differently.


src/Sentry/Sentry.csproj
```diff
@@ -52,6 +53,7 @@
   -->
   <ItemGroup>
     <PackageReference Include="Polyfill" Version="1.23.0" PrivateAssets="all" />
+    <PackageReference Include="System.IO.MemoryMappedFiles" Version="4.3.0" />
```
mattjohnsonpint, line 56: This would be a concern.  Though it appears it's not necessary.  The code still compiles if removed.

mattjohnsonpint:
The approach looks great!  Let's tidy this up, and add a test or two if possible.

I suggest we also wrap the calling code in conditional compilation to avoid using it in targets where `PublishSingleFile` isn't a feature.  Probably `#if NETCOREAPP3_0_OR_GREATER && PLATFORM_NEUTRAL`

(`PLATFORM_NEUTRAL` is ours - and keeps it out of the ios/android targets)

mattjohnsonpint:
We should also test with both a .NET Core 3.1 app and a .NET 6 app - both with `PublishSingleFile=true`.  The former will use the extract-to-disk approach, which should be trivial to find the assembly compared to what we're having to do here.

jamescrosswell:
> Let's tidy this up, and add a test or two if possible.

Yeah I scratched my head about how to test all of this. See comments about the "Sample App" in the PR description above. I guess that's more of an integration test. 

Currently the way the code is written, it assumes it needs to find an embedded module in the same application that it's running in. Potentially we could add a thin wrapper around the code that passes in a memory mapped file for the current assembly. However, for testing, we could load a pre-published "Hello World" app that we'd built and pass that in, instead. It would mean adding a fully compiled/bundled single file app to our repo... which is a bit ugly. 

Alternatively, we could compile/publish a single file app on the fly. It is quite slow to do that though, so would slow down our CI runs considerably. 

Any other ideas?

jamescrosswell:
> I suggest we also wrap everything in conditional compilation to avoid using it in targets where `PublishSingleFile` isn't a feature. Probably `#if NETCOREAPP3_0_OR_GREATER && PLATFORM_NEUTRAL`

I think that should be done. The only thing that isn't (and we could potentially omit) is `<AllowUnsafeBlocks>true</AllowUnsafeBlocks>` in the `Sentry.csproj` file. My MSBuild is a bit rusty - I can do a bit of googling/remembering if you'd like though.

src/Sentry/Sentry.csproj
```diff
@@ -93,35 +94,17 @@
       <_OSArchitecture>$([System.Runtime.InteropServices.RuntimeInformation]::OSArchitecture)</_OSArchitecture>
     </PropertyGroup>
     <ItemGroup>
-      <SentryCLIDownload
```
mattjohnsonpint, line 96: Undo unrelated whitespace changes please.  (We really should figure out how to keep this from happening.)

test/SingleFileTestApp/SingleFileTestApp.csproj
```diff
@@ -0,0 +1,13 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <IsTestProject>false</IsTestProject>
+    <OutputType>Exe</OutputType>
+    <TargetFramework>net7.0</TargetFramework>
+    <RuntimeIdentifier>win-x64</RuntimeIdentifier>
```
mattjohnsonpint, line 7: Let's also test on .NET Core 3.1 - as it will extract its files to disk and require a slightly different code path.

For the runtime identifier, let's choose it based on the OS and architecture the test is running on.  That way we'll get CI to test on Windows, Linux, and MacOS.

You can use MSBuild conditions in the same way we do for Sentry CLI in the Sentry.csproj file.  (Unless there's some magic to get the default RuntimeIdentifier automatically.)

jamescrosswell:
@bruno-garcia it might be good to get your eyes across this and also confirm what we want to do about .NET Core 3.1. From what Matt said, there aren't a lot of people clamouring for this to be released... so possibly we circle back to .NET Core 3.1 support if/when we see some appetite for this from customers.

src/Sentry/Internal/DebugStackTrace.cs
```diff
@@ -412,94 +421,40 @@ private static void DemangleLambdaReturnType(SentryStackFrame frame)
...
-        if (debugId == null)
-        {
-            options.LogInfo("Skipping debug image for module '{0}' because the Debug ID couldn't be determined", module.Name);
-            return null;
+            options.LogDebug("Got debug image for '{0}' having Debug ID: {1}", moduleName, debugImage.DebugId);
+            return debugImage;
         }
 
-        var debugImage = new DebugImage
+#if NET6_0_OR_GREATER && PLATFORM_NEUTRAL
```
jamescrosswell, line 439: Maybe this should be `NET5_0_OR_GREATER` actually...

bruno-garcia (approved):
Looks great! I did a quick pass and left a few suggestions but nothing big (didn't get in detail with the testing challenges yet)

> @bruno-garcia it might be good to get your eyes across this and also confirm what we want to do about .NET Core 3.1. From what Matt said, there aren't a lot of people clamouring for this to be released... so possibly we circle back to .NET Core 3.1 support if/when we see some appetite for this from customers.

We can leave .NET Core 3.1 out since it's out of support by the .NET team. Just clearly state on the docs that it's .NET 5 (also out of support) or latter.

bruno-garcia:
btw I can dogfood this through: https://github.com/bruno-garcia/cloudflare-dynamic-dns

Though I'd need to rebuild/publish and it has been running flawlessly for 2 years. I think I'd do it once I can swap the current transaction-based alerting (_transaction volume less than X_ mentioned here https://github.com/bruno-garcia/cloudflare-dynamic-dns#alerts) with Crons.

src/Sentry/Sentry.csproj
```diff
@@ -117,11 +118,7 @@
     </ItemGroup>
 
     <!-- Download the files -->
-    <DownloadFile
-      SourceUrl="https://downloads.sentry-cdn.com/sentry-cli/$(SentryCLIVersion)/%(SentryCLIDownload.Identity)"
-      DestinationFolder="$(SentryCLIDirectory)"
-      Condition="!Exists('$(SentryCLIDirectory)%(Identity)')"
-      Retries="3">
+    <DownloadFile SourceUrl="https://downloads.sentry-cdn.com/sentry-cli/$(SentryCLIVersion)/%(SentryCLIDownload.Identity)" DestinationFolder="$(SentryCLIDirectory)" Condition="!Exists('$(SentryCLIDirectory)%(Identity)')" Retries="3">
```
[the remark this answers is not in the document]
jamescrosswell: Nah it's just Rider likes to reformat this every time any other changes are made to the csproj file. I haven't worked out why yet... 

Will roll back.

test/Sentry.Tests/Internals/ILSpy/SingleFileAppTests.cs
```diff
@@ -0,0 +1,112 @@
...
+/// SingleFileTestApp is included in the solution that dotnet test is running against. Otherwise, tests are skipped.
+/// </summary>
+public class SingleFileAppTests
+{
+    private static readonly string ValidBundleFile;
+    private static readonly string InValidBundleFile;
+    static SingleFileAppTests()
+    {
+        var currentDirectory = Path.GetDirectoryName(Assembly.GetExecutingAssembly().Location)!;
+        var debugConfigMarker = "/bin/Debug/".OsAgnostic();
```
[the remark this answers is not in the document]
jamescrosswell: It works yes... I've move to a StringExtensions.cs class in our Sentry.Testing project (that seems to include various other little utils like this).

test/SingleFileTestApp/SingleFileTestApp.csproj
```diff
@@ -0,0 +1,24 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <IsTestProject>false</IsTestProject>
+    <OutputType>Exe</OutputType>
+    <TargetFramework>net6.0</TargetFramework>
+    <LangVersion>11</LangVersion>
+    <ImplicitUsings>true</ImplicitUsings>
```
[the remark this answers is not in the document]
jamescrosswell: We do... for this particular project (which isn't really a "Test" project per say) it has it's own `Directory.Build.props` file that intentionally does not import the parent.

```
<Project DefaultTargets="Build;Publish">
  <!-- Ignore Directory.Build.props from parent -->
</Project>
```

The reason being that this isn't really a Test project - it simply creates an artefact that's used by some of the tests. If it inherited all the same Build props as the other test projects, it would need an unnecessary reference to the Sentry project... which would simply slow down building/publishing this single file app. 

Ultimately it's a Hello World app, so I figured if it sits forever on LanVersion 11, it's no biggy.

jamescrosswell:
> btw I can dogfood this through: https://github.com/bruno-garcia/cloudflare-dynamic-dns
> 
> Though I'd need to rebuild/publish and it has been running flawlessly for 2 years. I think I'd do it once I can swap the current transaction-based alerting (_transaction volume less than X_ mentioned here https://github.com/bruno-garcia/cloudflare-dynamic-dns#alerts) with Crons.

Nice. So you had an "early version" of crons running years ago 😉

PR #2434, opened 2023-06-21
OpenTelemetry support

bruno-garcia:
PR from work done by Matt

## TODO

- [ ] Update the GH discussion: https://github.com/getsentry/sentry/discussions/40712

## Not planned
- [GenerateSentryErrorsFromOtelSpan](https://develop.sentry.dev/sdk/performance/opentelemetry/#step-7-define-generatesentryerrorsfromotelspan) is **Blocked** by https://github.com/open-telemetry/opentelemetry-dotnet/issues/2439#issuecomment-1577314568. 

This is something only our Javascript SDK has. None of the other Sentry SDKs support it. 

It's not critical as we already capture errors through the normal exception handlers. It would only be an issue if a particular Otel instrumentation decided to not let exceptions bubble up, and only to report them as Otel events.


jamescrosswell:
I noticed some `TryGetBaggageHeader` and `TryGetSentryTraceHeader` methods [in `Sentry.AspNet.HttpContextExtensions`](https://github.com/getsentry/sentry-dotnet/blob/ea1c0cf5bb6a843c84930511a5278d3522c9d081/src/Sentry.AspNet/HttpContextExtensions.cs#L13). 

I also noticed almost the same [in `SentryTracingMiddleware` in the AspNetCore project](https://github.com/getsentry/sentry-dotnet/blob/ea1c0cf5bb6a843c84930511a5278d3522c9d081/src/Sentry.AspNetCore/SentryTracingMiddleware.cs#L29).

They're slightly different because they're targeting different HttpContext abstractions.
```csharp
        var value = context.Request.Headers.Get(BaggageHeader.HttpHeaderName); // AspNet
        var value = context.Request.Headers.GetValueOrDefault(BaggageHeader.HttpHeaderName); // AspNetCore

```

It turns out, we also need something very similar in the [SentryPropagator for Otel](https://github.com/getsentry/sentry-dotnet/blob/6fc599f8cc28945a56c37e1ebffda60e2a073edf/src/Sentry.OpenTelemetry/SentryPropagator.cs#L25).

A couple of questions:
- Are we trying to target AspNet and AspNetCore with our Open Telemetry efforts?
- If we're not, is it OK for the `Sentry.OpenTelemetry` project to take a reference to `Sentry.AspNetCore` to use those extensions?
- Those extensions do use `SentryOptions.LogDebug`. I'm not sure if we can inject `SentryOptions` into the `SentryPropagator`. Maybe all moot in that case. It just makes me itchy to use copy/paste as a form of code reuse 🙈

bruno-garcia:
> If we're not, is it OK for the Sentry.OpenTelemetry project to take a reference to Sentry.AspNetCore to use those extensions?

I'd say rather not take a dependency on AspNetCore. Any use of OTel outside ASP.NET would have that dep and also would add another dependency to deal with versioning and breaking changes.

Those two methods seem quite small. I'd say we could be pragmatic and copy them over but if I understand correctly the goal is to get the AspNetCore HTTP context object so we can read from it? If that's the case we need to consider introducing a new package like: `Sentry.OpenTelemetry.AspNetCore` instead. Open has: 
https://www.nuget.org/packages/OpenTelemetry.Instrumentation.AspNetCore/

That does take dep on `OpenTelemetry` and AspNetCore:

image

In any regard better to break the work up in parts, if we can do core bits first, we can follow up with AspNetCore after (we can leave AspNet out for the time being, possibly forever)

jamescrosswell:
> In any regard better to break the work up in parts, if we can do core bits first, we can follow up with AspNetCore after (we can leave AspNet out for the time being, possibly forever)

I think we could support Otel span processors without any dependency on `HttpContext`. 

Propagation has an inherent dependency on the concept of "Http Headers", since that's how the trace id gets passed around (at least in Sentry's code - I'm not sure how it gets passed into things like DB servers). So I don't think we can include a solution for propagation without a dependency on `Microsoft.AspNetCore.Http.Abstractions` and/or `System.Web` (for AspNet).

Options for that:
1. We put it all in a `Sentry.OpenTelemetry` package with directives to select the right dependencies, based on the target framework. This would be a bit of a bummer for people wanting to use Otel in non web based projects though
2. We pull it out to separate `Sentry.OpenTelemetry.Propagation.AspNet` and `Sentry.OpenTelemetry.Propagation.AspNetCore` assemblies
3. Something else?

src/Sentry.OpenTelemetry/SentryPropagator.cs
```diff
@@ -0,0 +1,142 @@
...
+
+        var otelSpanContext = new ActivityContext(
+            sentryTraceHeader.TraceId.AsActivityTraceId(),
+            sentryTraceHeader.SpanId.AsActivitySpanId(),
+            sentryTraceHeader.IsSampled is true ? ActivityTraceFlags.Recorded : ActivityTraceFlags.None,
+            null,
+            true
+            );
+
+        // TODO: Understand why this is needed in the Java implementation
```
jamescrosswell, line 59: @bruno-garcia do you have any idea why they've done [this in the Java implementation](https://github.com/getsentry/sentry-java/blob/cb2fb895cc8d31a4be7025bce2ea536fb7d4d055/sentry-opentelemetry/sentry-opentelemetry-core/src/main/java/io/sentry/opentelemetry/SentryPropagator.java#L81)?

From [the source code for otel's java sdk](https://github.com/open-telemetry/opentelemetry-java/blob/bcec7e9380c943be3c3764efa22fbb10ddd6473f/api/all/src/main/java/io/opentelemetry/api/trace/Span.java#L79):
```
  /**
   * Returns a non-recording {@link Span} that holds the provided {@link SpanContext} but has no
   * functionality. It will not be exported and all tracing operations are no-op, but it can be used
   * to propagate a valid {@link SpanContext} downstream.
   */
  static Span wrap(SpanContext spanContext) {
    if (spanContext == null) {
      ApiUsageLogger.log("context is null");
      return getInvalid();
    }
    if (!spanContext.isValid()) {
      return getInvalid();
    }
    return PropagatedSpan.create(spanContext);
  }
```

The only reason I can think why they're doing this is to get a Span.noOp... and they appear to have littered the [rest of their codebase](https://github.com/getsentry/sentry-java/pull/2361/files) with checks like `span != null && !span.isNoOp ` instead of `span != null` to determine whether tracing gets performed... so I think they're using this as a way to disable Sentry's built in tracing mechanism when Otel tracing is being used, but wanted to double check I'd understood correctly.

bruno-garcia: > so I think they're using this as a way to disable Sentry's built in tracing mechanism when Otel tracing is being used, but wanted to double check I'd understood correctly.

That's right, my understanding is that either one or the other gets active. cc @AbhiPrasad to confirm

AbhiPrasad: Yup we either choose one or the other. See https://develop.sentry.dev/sdk/performance/opentelemetry/#step-6-add-instrumenter-option-to-sentry-sdk for more details.

jamescrosswell:
I believe the propagator is working now (see below)... just need to add a few tests and then we should be done.

image

jamescrosswell:
# The Dynamic Sampling Context Problem

This is the bit that Matt was wrestling with when I started working on the PR, so probably worth documenting, as it's really not that obvious. The following is an explanation of what's going on in the Java implementation. We may have to get creative to solve in .NET as we don't have the same building blocks to work with that they did in Java. 

## Java implementation
Issue: [Java SDK basic OTEL Support sentry-java#2327](https://github.com/getsentry/sentry-java/issues/2327)
Pull requests:
- [Add SpanProcessor for OTEL sentry-java#2357](https://github.com/getsentry/sentry-java/pull/2357)
- [Add contexts to SentryTracer and add otel context to it sentry-java#2361](https://github.com/getsentry/sentry-java/pull/2361)

### Extracting

[SentryPropagator.extract](https://github.com/getsentry/sentry-java/blob/93b327f83955d8413abcf518c4e5ce8398c7a944/sentry-opentelemetry/sentry-opentelemetry-core/src/main/java/io/sentry/opentelemetry/SentryPropagator.java#L78)
```
    Baggage baggage = Baggage.fromHeader(baggageString);
    modifiedContext = modifiedContext.with(SentryOtelKeys.SENTRY_BAGGAGE_KEY, baggage);
```

.. which indirectly calls [this overload](https://github.com/getsentry/sentry-java/blob/dc67004db625e4a09c3392d4ac3b7b38a9498d28/sentry/src/main/java/io/sentry/Baggage.java#L81C3-L84C37),
the only important detail of which is that `includeThirdPartyValues==false` so basically we just extract any values from
the `baggage` header that are prefixed with `sentry-`, 

```
public static Baggage fromHeader(
      final @Nullable String headerValue,
      final boolean includeThirdPartyValues,
      final @NotNull ILogger logger)
```

All the `sentry-*` values from the baggage header get stuffed back in the context (as can be seen from the very first
code snippet above) using [modifiedContext.with](https://github.com/open-telemetry/opentelemetry-java/blob/ddb47e14c814a1f3e114cbe0c6db50638e9cdf14/context/src/main/java/io/opentelemetry/context/Context.java#L173-L175) which doesn't exist in the .NET OTEL SDK:
```
return value.storeInContext(this);
```

And that puts the data in something implementing [`ContextStorage`](https://github.com/open-telemetry/opentelemetry-java/blob/ddb47e14c814a1f3e114cbe0c6db50638e9cdf14/context/src/main/java/io/opentelemetry/context/ContextStorage.java#L63)... The concrete class is probably not relevant. It'll be stored somewhere so it can be used later. The key question then is where does this DSC get used later?

`SentryPropagator.inject` below gets everything it needs for DSC injection from the Transaction associated with the
Sentry Span... And it puts all of those things directly in the `baggage` header of the `carrier` (e.g. an HttpRequest 
header). So the stuff in `ContextStorage` is not used for injection.

Where it does appear to be used is in [`SpanProcessor.onStart`](https://github.com/getsentry/sentry-java/blob/93b327f83955d8413abcf518c4e5ce8398c7a944/sentry-opentelemetry/sentry-opentelemetry-core/src/main/java/io/sentry/opentelemetry/SentrySpanProcessor.java#L54), which calls [SentrySpanProcessor.getTraceData](https://github.com/getsentry/sentry-java/blob/93b327f83955d8413abcf518c4e5ce8398c7a944/sentry-opentelemetry/sentry-opentelemetry-core/src/main/java/io/sentry/opentelemetry/SentrySpanProcessor.java#L186)
```
baggage = parentContext.get(SentryOtelKeys.SENTRY_BAGGAGE_KEY);
```

When creating a new Sentry Span representing a Transaction, if that information is available then it's [used to create 
the new Transaction](https://github.com/getsentry/sentry-java/blob/93b327f83955d8413abcf518c4e5ce8398c7a944/sentry-opentelemetry/sentry-opentelemetry-core/src/main/java/io/sentry/opentelemetry/SentrySpanProcessor.java#L82C18-L82C18)
```
  TransactionContext.fromSentryTrace(
      transactionName,
      transactionNameSource,
      op,
      traceData.getSentryTraceHeader(),
      traceData.getBaggage(),
      spanId);
```

### Injecting DSC

[SentryPropagator.inject](https://github.com/getsentry/sentry-java/blob/93b327f83955d8413abcf518c4e5ce8398c7a944/sentry-opentelemetry/sentry-opentelemetry-core/src/main/java/io/sentry/opentelemetry/SentryPropagator.java#L48-L52)
```
    final @Nullable BaggageHeader baggageHeader = sentrySpan.toBaggageHeader(Collections.emptyList());
    if (baggageHeader != null) {
      setter.set(carrier, baggageHeader.getName(), baggageHeader.getValue());
    }
```

.. which eventually translates into:

[SentryTracer.updateBaggageValues](https://github.com/getsentry/sentry-java/blob/dc67004db625e4a09c3392d4ac3b7b38a9498d28/sentry/src/main/java/io/sentry/SentryTracer.java#L510-L511)
```
    baggage.setValuesFromTransaction(
        this, userAtomicReference.get(), hub.getOptions(), this.getSamplingDecision());
```

... and finally into:

[baggage.setValuesFromTransaction](https://github.com/getsentry/sentry-java/blob/dc67004db625e4a09c3392d4ac3b7b38a9498d28/sentry/src/main/java/io/sentry/Baggage.java#L338-L353)
```
    
    public void setValuesFromTransaction(
      final @NotNull ITransaction transaction,
      final @Nullable User user,
      final @NotNull SentryOptions sentryOptions,
      final @Nullable TracesSamplingDecision samplingDecision
      ) 
    {
        setTraceId(transaction.getSpanContext().getTraceId().toString());
        setPublicKey(new Dsn(sentryOptions.getDsn()).getPublicKey());
        setRelease(sentryOptions.getRelease());
        setEnvironment(sentryOptions.getEnvironment());
        setUserSegment(user != null ? getSegment(user) : null);
        setTransaction(isHighQualityTransactionName(transaction.getTransactionNameSource()) ? transaction.getName() : null);
        setSampleRate(sampleRateToString(sampleRate(samplingDecision)));
    }
```

src/Sentry.OpenTelemetry/SentryPropagator.cs
```diff
@@ -0,0 +1,126 @@
...
+        {
+            Debug.WriteLine("No SentryTraceHeader present in carrier");
+            return result;
+        }
+
+        Debug.WriteLine($"Extracted SentryTraceHeader from carrier: {sentryTraceHeader}");
+
+        var activityContext = new ActivityContext(
+            sentryTraceHeader.TraceId.AsActivityTraceId(),
+            sentryTraceHeader.SpanId.AsActivitySpanId(),
```
jamescrosswell, line 71: This is possibly incorrect. The sentryTraceHeader.SpanId would be the Id of the upstream span right? Should we be using this to set the `SpanTracer.ParentSpanId` or`TransactionTracer.ParentSpanId` in the `SentrySpanProcessor` instead?

jamescrosswell: It seems to be [what the Java code is doing](https://github.com/getsentry/sentry-java/blob/dc67004db625e4a09c3392d4ac3b7b38a9498d28/sentry-opentelemetry/sentry-opentelemetry-core/src/main/java/io/sentry/opentelemetry/SentrySpanProcessor.java#L103) though... 

```java
      final SpanId spanId = new SpanId(traceData.getSpanId());
```

jamescrosswell:
@bruno-garcia suggested creating a new PR for this so that I can get a review from him (rather than the other way around)... I'll close this PR and create a new one with all the relevant history/context then.

jamescrosswell:
These are the scripts for the two python nodes that I was using for testing:

### main.py - The entry script
```
import requests
import sentry_sdk
#
sentry_sdk.init(
    dsn="...",

    # Set traces_sample_rate to 1.0 to capture 100%
    # of transactions for performance monitoring.
    # We recommend adjusting this value in production.
    traces_sample_rate=1.0,
    debug=True
)

def eat_slice(slice):
    with sentry_sdk.start_span(description="Eat Slice") as span:
        print(f'span.trace_id: {span.trace_id}')
        print(f'span.span_id: {span.span_id}')
        span.set_tag("slice.topping", slice)
        response = requests.get('http://localhost:5092/hello')
        print(response.status_code)

def eat_pizza():
    while True:
        user_input = input("What kind of pizza do you want to eat?")
        if user_input.lower() == "quit":
            break
        else:
            with sentry_sdk.start_transaction(op="task", name="Eat Pizza") as transaction:
                print(f'transaction.trace_id: {transaction.trace_id}')
                eat_slice(user_input)

eat_pizza()
```

### ferver.py - a hello world Flask server
```
# Quick and dirty Flask server to act as a third component in the experiment
import sentry_sdk
from flask import Flask

sentry_sdk.init(
    dsn="...",
    traces_sample_rate=1.0,
)

app = Flask(__name__)

@app.route("/")
def hello_world():
    return "<p>Hello, World!</p>"
```

PR #2436, opened 2023-06-21
ios: missing binding method

bruno-garcia:
Building a new checkout of `main` reruns the script to create bindings and makes a change.

In CI, do we break if we end up with a dirty workspace? It's probably a good idea to catch cases such as this.

Additionally, `modules/sentry-cocoa` was dirty, with 2 files deleted:

```
-Subproject commit d277532e1c8af813981ba01f591b15bbdd735615
+Subproject commit d277532e1c8af813981ba01f591b15bbdd735615-dirty
..
        modified:   modules/sentry-cocoa (modified content)
..
        deleted:    Sentry.xcodeproj/xcshareddata/xcschemes/SentryPrivate.xcscheme
        deleted:    Sentry.xcodeproj/xcshareddata/xcschemes/SentrySwiftUI.xcscheme
```

`d277532e1c8af813981ba01f591b15bbdd735615` is 8.8.0 so nothing wrong on the .NET end, @brustolin were these deleted after the release? Or is it something with the way we build things with Carthage?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,11 @@
 
 ## Unreleased
 
+### Features
+
+- Symbolication for Single File Apps ([#2425](https://github.com/getsentry/sentry-dotnet/pull/2425))
+- Add binding to `SwiftAsyncStacktraces` on iOS ([#2436](https://github.com/getsentry/sentry-dotnet/pull/2436))
+
 ### Fixes
 
 - Fix Sentry logger options for MAUI and Azure Functions ([#2423](https://github.com/getsentry/sentry-dotnet/pull/2423))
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,11 @@
 
 ## Unreleased
 
+### Features
```
bruno-garcia, line 5: Rather start with Feature, than Fixes

brustolin:
> were these deleted after the release? Or is it something with the way we build things with Carthage?

How we build things with Carthage. The build cocoa script delete this files in order for Carthage to work.

jamescrosswell (approved):
I'm assuming you know what you're doing with the `src/Sentry.Bindings.Cocoa/ApiDefinitions.cs` so happy to approve 😉

bruno-garcia:
> > were these deleted after the release? Or is it something with the way we build things with Carthage?
> 
> How we build things with Carthage. The build cocoa script delete this files in order for Carthage to work.

How do we resolve this? On your CI you end up with these changes pending and that doesnt' break the build?

brustolin:
We don't do this changes in cocoa CI, only dotnet specific Carthage requires this.

I dont get why you need to resolve this. The workflow pull sentry-cocoa, modified it and build it, copy the binaries to another place to be able to test it, and then, discard everything no? 

But, I just open a PR that will solve this: #2439

PR #2443, opened 2023-06-23
fix: Pinning dependency versions in tests

bitsandfoxes:
Based on the most frustrating of all development experiences. New versions of packages were released but could not be downloaded. This resulted in the greatest of confusions and broke my local environment. (not for the first time)
I don't see a good reason, especially for tests, why versions of dependencies cannot be pinned.

I'd rather have them pinned and add a bumping mechanism instead of getting surprised.

#skip-changelog

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -5,6 +5,12 @@ updates:
   schedule:
     interval: monthly
   open-pull-requests-limit: 2
+
+- package-ecosystem: nuget
+  directory: "/test"
+  schedule:
+    interval: weekly
+  open-pull-requests-limit: 2 
   
 - package-ecosystem: "github-actions"
   directory: "/"
--- a/test/Sentry.AspNetCore.Grpc.Tests/Sentry.AspNetCore.Grpc.Tests.csproj
+++ b/test/Sentry.AspNetCore.Grpc.Tests/Sentry.AspNetCore.Grpc.Tests.csproj
@@ -5,9 +5,9 @@
   </PropertyGroup>
 
   <ItemGroup>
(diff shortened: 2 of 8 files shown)


SeanFeldman:
> I'd rather have them pinned and add a bumping mechanism instead of getting surprised.

Would Dependabot or Renovate help? In general, having the latest of the same major version is better for testing to identify issues earlier.

bitsandfoxes:
> > I'd rather have them pinned and add a bumping mechanism instead of getting surprised.
> 
> Would Dependabot or Renovate help? In general, having the latest of the same major version is better for testing to identify issues earlier.

Super valid point!

SeanFeldman:
> Super valid point!

With Dependabot already [in place](https://github.com/getsentry/sentry-dotnet/pulls?q=is%3Apr+is%3Aclosed+dependabot), what else I do not see that would require pinning for test projects?

.github/dependabot.yml
```diff
@@ -5,6 +5,12 @@ updates:
   schedule:
     interval: monthly
   open-pull-requests-limit: 2
+
+- package-ecosystem: nuget
```
bitsandfoxes, line 9: @jamescrosswell I'm not too familiar with dependabot but that should do the trick?

SeanFeldman: @bitsandfoxes yes. I'm unsure about the number of PRs that can be opened. Two sounds very constraining for a weekly period.

bruno-garcia: On the other hand not very busy/distracting, we can adjust as we go

bruno-garcia (approved):

PR #2452, opened 2023-06-27
Support transaction finishing automatically with 'idle timeout'

### LINKED ISSUE #1074: Support transaction finishing automatically with 'idle timeout'
The idea is to have transactions that do not have a specified ending but a timeout instead. As if we had a mode for finishing transactions automatically once no new spans are added to it for N milliseconds, or a general timeout.

This would also be used by Unity, for example, when a scene loads. Potentially other UI frameworks like MAUI, WinForms, Uno, Avalonia and WPF. In all cases, user interaction can be a trigger. Without a specified end time, it runs until 'enough' has been tracked. 

We need to define the spec in the develop.sentry.dev docs and align with other implementations of it. This is an approach used by the JS on the browser (and React Native). Similar behavior exists in different SDKs such as JavaScript and [Android](https://github.com/getsentry/sentry-java/issues/1523).

In terms of Span operation names, some have been defined already. For example, [under _Mobile_ here](https://develop.sentry.dev/sdk/performance/span-operations/#mobile) we see: `ui.action.click` which often rely on this behavior of completing based on a sliding expiration instead of a 'finish' call into the SDK.

### Solution Brainstorm

This could be achieved by extending `TransactionTracer` with an optional parameter for the timeout. This should work independently of `wait for children` and simply trigger a finish after a set time period.

Other SDKs implementation of this: 
- Cocoa: https://github.com/getsentry/sentry-cocoa/blob/d2076a999d23fff77ff5736d6ae7c6734e49b8e0/Sources/Sentry/SentryTracer.m#L97
- Dart: https://github.com/getsentry/sentry-dart/blob/a5e904f5a26ffdfc8b6065a68c6152ea936b0644/dart/lib/src/sentry_tracer.dart#L43
- Java: https://github.com/getsentry/sentry-java/blob/fa3886ffaae668ef2e5d57783df89bfd8bb0c6bb/sentry/src/main/java/io/sentry/SentryTracer.java#L100

Spec/SDK docs for this: https://develop.sentry.dev/sdk/performance/ui-event-transactions/

SeanFeldman:
Fixes #1074

Add support to specify transaction idle timeout to finish an idling transaction.

- ~~Add `SentryOptions` documentation under [common options](https://docs.sentry.io/platforms/dotnet/configuration/options/#common-options)~~
- [x] Raise an issue to internalize public `TransactionTracer` constructors for testing in [milestone for v4](https://github.com/getsentry/sentry-dotnet/milestone/6) --> #2481


test/Sentry.Tests/Protocol/TransactionTests.cs
```diff
@@ -504,4 +504,17 @@ public void ISpan_GetTransaction_FromSpan()
         // Assert
         Assert.Same(transaction, result);
     }
+
+    [Fact]
+    public async Task Idle_transaction_should_finish_with_idle_timeout_specified()
+    {
+        // Arrange
+        var transaction = new TransactionTracer(DisabledHub.Instance, "my name", "my op", TransactionNameSource.Custom, TimeSpan.FromMilliseconds(10));
```
SeanFeldman, line 512: @bitsandfoxes, I chose a 10 milliseconds timeout and ensured it's invoked with a delay of 15 ms.
Thinking about it, the idle timeout and the delay could have been even less than that. On the flip side, this is not a significantly substantial amount of time. What do you think, keep it as-is or reduce it?

src/Sentry/TransactionTracer.cs
```diff
@@ -178,21 +180,35 @@ public IReadOnlyList<string> Fingerprint
...
-        : this(hub, name, operation, TransactionNameSource.Custom)
+        : this(hub, name, operation, TransactionNameSource.Custom, null)
     {
     }
 
     /// <summary>
     /// Initializes an instance of <see cref="Transaction"/>.
     /// </summary>
-    public TransactionTracer(IHub hub, string name, string operation, TransactionNameSource nameSource)
+    public TransactionTracer(IHub hub, string name, string operation, TransactionNameSource nameSource, TimeSpan? idleTimeout)
```
SeanFeldman, line 190: @bitsandfoxes, do you know why there's a 3rd constructor that is not chained? 
https://github.com/getsentry/sentry-dotnet/blob/1074-transacton-idle-timeout/src/Sentry/TransactionTracer.cs#L217

jamescrosswell: `TransactionContext` contains a superset of the properties being passed in the other constructors... maybe everything should ultimately chain up to that? It's the one that gets used [in the Hub](https://github.com/getsentry/sentry-dotnet/blob/5401fed146c2219f2eae25cf26b96419e59b804e/src/Sentry/Internal/Hub.cs#L119).

SeanFeldman: I got confused by the 3 constructors
1.  `public TransactionTracer(IHub hub, string name, string operation)` - calling the 2nd constructor
2. `public TransactionTracer(IHub hub, string name, string operation, TransactionNameSource nameSource, TimeSpan? idleTimeout)`
3. `public TransactionTracer(IHub hub, ITransactionContext context)`

The [1st constructor](https://github.com/getsentry/sentry-dotnet/blob/7db3f1b9e8569986d3929e27310f41140fa19723/src/Sentry/TransactionTracer.cs#L180) is chaining to the [2nd constructor](https://github.com/getsentry/sentry-dotnet/blob/7db3f1b9e8569986d3929e27310f41140fa19723/src/Sentry/TransactionTracer.cs#L188).
The [3rd constructor](https://github.com/getsentry/sentry-dotnet/blob/7db3f1b9e8569986d3929e27310f41140fa19723/src/Sentry/TransactionTracer.cs#L201), using `ITransactionContext` is not chaining to anything.

The 1st constructor is mainly used for testing purposes. It also solely uses the 2nd constructor.
This supports what you're suggesting, @jamescrosswell. 

Why do we need the 1st (and the 2nd) constructor as a public constructor, given it's for testing only and could be made internal? Potentially, those two could be made internal in the next major.

jamescrosswell: Sounds reasonably. I'm not sure if people using the SDK fiddle with TransactionTracers directly at all tbh - I imagine they start transactions and create child spans from the SDK. The Tracers seem like an implementation detail (it's what we wrap those transactions/spans in so that we've got somewhere to track performance data).

SeanFeldman: I'd like to leave the 1st and the 2nd contractors disconnected from the 3rd one, which the Hub uses anyway. If those two are solely for testing purposes, perhaps some comment could be made to clarify their purpose, along with a TODO for version 4 to make those internal.

test/Sentry.Tests/Protocol/TransactionTests.cs
SeanFeldman: Changed to lower values.

jamescrosswell:
>This could be achieved by extending TransactionTracer with an optional parameter for the timeout. 

I imagine SDK users would want to do this by passing in a timeout via one of the [Hub.StartTransaction extensions](https://github.com/getsentry/sentry-dotnet/blob/b653afc5165e463c11f1223ffd601e20a0a1560e/src/Sentry/HubExtensions.cs#L16).

We probably need those additional extension methods before this functionality would be usable then right?

@SeanFeldman maybe worth showing how this would work in one of the sample projects, as a bit of a smoke test to make sure this is all wired up?

SeanFeldman:
> I imagine SDK users would want to do this by passing in a timeout via one of the [Hub.StartTransaction extensions](https://github.com/getsentry/sentry-dotnet/blob/b653afc5165e463c11f1223ffd601e20a0a1560e/src/Sentry/HubExtensions.cs#L16).

I don't know if it should be just an option or also an explicit option via extensions. For example, Java has it on [`SentryOptions`](https://github.com/getsentry/sentry-java/blob/dc67004db625e4a09c3392d4ac3b7b38a9498d28/sentry/src/main/java/io/sentry/SentryOptions.java#L376)

@bruno-garcia or @bitsandfoxes, do you have some context to share? Thank you.

SeanFeldman:
> We probably need those additional extension methods before this functionality would be usable then right?

@jamescrosswell, yep. That will be the last step once I know everything that's needed is in place.

src/Sentry/TransactionTracer.cs
```diff
@@ -198,9 +200,15 @@ public TransactionTracer(IHub hub, string name, string operation, TransactionNam
     /// <summary>
     /// Initializes an instance of <see cref="TransactionTracer"/>.
     /// </summary>
-    public TransactionTracer(IHub hub, ITransactionContext context)
+    public TransactionTracer(IHub hub, ITransactionContext context) : this (hub, context, null) {}
+
+    /// <summary>
+    /// Initializes an instance of <see cref="TransactionTracer"/>.
+    /// </summary>
+    public TransactionTracer(IHub hub, ITransactionContext context, TimeSpan? idleTimeout)
```
bruno-garcia, line 208: Can we read the options through `hub` and avoid adding the overload?

It's internal for that reason: https://github.com/getsentry/sentry-dotnet/blob/69a92b54fb9254aca451bf9ff038154f053f6b69/src/Sentry/Internal/Hub.cs#L28

src/Sentry/TransactionTracer.cs
```diff
@@ -6,9 +6,11 @@ namespace Sentry;
 /// <summary>
 /// Transaction tracer.
 /// </summary>
-public class TransactionTracer : ITransaction, IHasDistribution, IHasTransactionNameSource, IHasMeasurements
+public class TransactionTracer : ITransaction, IHasDistribution, IHasTransactionNameSource, IHasMeasurements, IDisposable
```
bruno-garcia, line 9: Is adding IDisposable is a "breaking change" as we expect anyone bumping the SDK to now deal with disposing this. How do we expect the caller to deal with this? 
IIRC we return this as `ITransaction` in places, right? We'd need to document/expect users to try cast to IDisposable and call it if not null.

Also, Dispose could only be called after `Finish` which defeats the purpose of idle transaction (which is completed from the background thread)

src/Sentry/TransactionTracer.cs
```diff
@@ -260,6 +281,8 @@ internal ISpan StartChild(SpanId parentSpanId, string operation)
     /// <inheritdoc />
     public void Finish()
     {
+        _idleTimer?.Change(Timeout.InfiniteTimeSpan, Timeout.InfiniteTimeSpan);
```
bruno-garcia, line 284: There'a s race condition here where Finish could be called by the user and the background thread at the same time. 

A cheap `Interlocked.Exchange` could help here, just let the first one run through and the other thread just bails immediately

bruno-garcia (changes requested):

bruno-garcia:
> > We probably need those additional extension methods before this functionality would be usable then right?

> @jamescrosswell, yep. That will be the last step once I know everything that's needed is in place.

I believe these calls from a sample are import from the get go. Looking at what the API call will look like helps us go the right direction

src/Sentry/TransactionTracer.cs
bruno-garcia: > The 1st constructor is mainly used for testing purposes. 
> Why do we need the 1st (and the 2nd) constructor as a public constructor, given it's for testing only and could be made internal? Potentially, those two could be made internal in the next major.

Agree! These should be `internal`, we can use them for testing through `InternalsVisibleTo` and they don't polute the public API

SeanFeldman: @bruno-garcia, aside from leaving a to-do comment in the code, is there anything else that would be used to indicate future deprecation?

src/Sentry/TransactionTracer.cs
SeanFeldman: Yes, that could be done.
`IHub` has more than one implementation but only `Hub` has the internal `SentryOptions`. Is the transaction tracer guaranteed to be invoked by `Hub` and be able to cast to `Hub` to access `SentryOptions` or not? And if not, should the failover be idle timeout as null? I can see that working.

SeanFeldman: Implementations of `IHub` are

1. [`Hub`](https://github.com/getsentry/sentry-dotnet/blob/af5aac5a3d0e5faddda57d68a6ff7d4f169aab88/src/Sentry/Internal/Hub.cs#L7C5-L7C5)
2. [`DisabledHub`](https://github.com/getsentry/sentry-dotnet/blob/af5aac5a3d0e5faddda57d68a6ff7d4f169aab88/src/Sentry/Extensibility/DisabledHub.cs#L6C14-L6C25)
3. [`HubAdapter`](https://github.com/getsentry/sentry-dotnet/blob/af5aac5a3d0e5faddda57d68a6ff7d4f169aab88/src/Sentry/Extensibility/HubAdapter.cs#L15)

Only `Hub` has the SentryOptions. The other two implementations are used for testing.
I've decided to cast `IHub` to `Hub explicitly` and, if not `null`, access the options. This should be a fair assumption.

SeanFeldman:
> > > We probably need those additional extension methods before this functionality would be usable then right?
> 
> > @jamescrosswell, yep. That will be the last step once I know everything that's needed is in place.
> 
> I believe these calls from a sample are import from the get go. Looking at what the API call will look like helps us go the right direction
> 
> > I'm fine with having it as an explicit option. Maybe I'm missing something but what would be the benefit of putting it on the extension?
> 
> We have this as a default value in the options. Users shouldn't have to be concerned with that for the most part (and in fact, that's mainly used by us, on our auto instrumentation for UI frameworks). It's cheap to add overloads later but unless we're confident folks will need them to solve real use cases, default to leaving them out (API design 101: when in doubt leave them out)

@bruno-garcia, to confirm, you're saying we **shouldn't** add an extension method and stick to the initial `SentryOption` until the need emerges in something else. Is that accurate?

src/Sentry/TransactionTracer.cs
SeanFeldman: > Is adding IDisposable is a "breaking change" as we expect anyone bumping the SDK to now deal with disposing this. How do we expect the caller to deal with this?
> IIRC we return this as `ITransaction` in places, right? We'd need to document/expect users to try cast to IDisposable and call it if not null.

You're correct. It would be considered a breaking change.
`IDisposable` is there to dispose of the `Timer`. But I wonder, since we control the `Finish` and stop the timer there, the timer could be disposed of w/o the need for the user to do anything. I.e. dispose of the timer in the `Finish` and remove the breaking `IDisposable` on the `TransactionTracer`, removing the need for the user to handle the disposing.

src/Sentry/TransactionTracer.cs
SeanFeldman: Addressed

SeanFeldman:
All conflicts should be resolved.
Looks like PRs service is currently degraded 
image

Eventual consistency FTW.

src/Sentry/TransactionTracer.cs
```diff
@@ -210,6 +215,19 @@ public TransactionTracer(IHub hub, ITransactionContext context)
         Description = context.Description;
         Status = context.Status;
         IsSampled = context.IsSampled;
+
+        if (idleTimeout != null)
```
bitsandfoxes, line 219: Instead of just grabbing it from the options, can we optionally pass the idle timeout to the tracer somehow? The idea is to have `some` transactions that can be idle and time out on their own. The primary use case for this feature right now would be auto-instrumentation for things like UI interaction with no properly defined end.

Right now, with the `IdleTimeout` set on the options, all transactions, even the ones created by users, turn out to be idle.

bruno-garcia:
some rebase problem? 
image

SeanFeldman:
> some rebase problem?

Yep. Working on the same files as OTEL branch touched.

bruno-garcia:
> > some rebase problem?
> 
> Yep. Working on the same files as OTEL branch touched.

That's merged in `main` which is your target. They shouldn't be showing in the diff here.

SeanFeldman:
> > > some rebase problem?
> > 
> > 
> > Yep. Working on the same files as OTEL branch touched.
> 
> That's merged in `main` which is your target. They shouldn't be showing in the diff here.

Correct. I'll address it so that only my changes show up 🙂

SeanFeldman:
> > > > some rebase problem?
> > > 
> > > 
> > > Yep. Working on the same files as OTEL branch touched.
> > 
> > 
> > That's merged in `main` which is your target. They shouldn't be showing in the diff here.
> 
> Correct. I'll address it so that only my changes show up 🙂

I've merged `main` rather than rebasing. Since the PR is squashed merged, it won't matter much. Now it looks a little better, @bruno-garcia

image

src/Sentry/TransactionTracer.cs
```diff
@@ -201,6 +205,7 @@ public TransactionTracer(IHub hub, string name, string operation, TransactionNam
     public TransactionTracer(IHub hub, ITransactionContext context)
     {
         _hub = hub;
+        var idleTimeout = (hub as Hub)?.Options.IdleTimeout;
```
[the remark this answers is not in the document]
SeanFeldman: That's a handy extension method. Thank you, @bitsandfoxes.

src/Sentry/TransactionTracer.cs
SeanFeldman: @bruno-garcia, should I raise an issue for v4 to mark those methods as internal? Or is there a registry for v4 changes this could be added to? /cc @bitsandfoxes

src/Sentry/TransactionTracer.cs
SeanFeldman: > Instead of just grabbing it from the options, can we optionally pass the idle timeout to the tracer somehow? The idea is to have `some` transactions that can be idle and time out on their own. The primary use case for this feature right now would be auto-instrumentation for things like UI interaction with no properly defined end.
> 
> Right now, with the `IdleTimeout` set on the options, all transactions, even the ones created by users, turn out to be idle.

Makes sense. I have dropped the idea of the extension method due to [this comment](https://github.com/getsentry/sentry-dotnet/pull/2452#issuecomment-1611620938), which sounded as questioning the need for an extension method. So was [this comment](https://github.com/getsentry/sentry-dotnet/pull/2452#issuecomment-1612147758).

But I can see when a user would like to have a transaction with an idle timeout w/o necessarily having all transactions configured with an idle timeout.

Should this be done via [`HubExtensions`](https://github.com/getsentry/sentry-dotnet/blob/main/src/Sentry/HubExtensions.cs)? If yes, I think that could be a wrap-up for this PR? Plus, a sample, maybe.

jamescrosswell:
@SeanFeldman one thing to note is that I [added this to the Otel impmentation](https://github.com/getsentry/sentry-dotnet/pull/2453#discussion_r1250927718). 

For context, Otel is using "Activities" and we listen to OnStart and OnEnd events for Activities that we use to convert those into Sentry Transactions/Spans. Otel creates Activities for the messages that we're sending to Sentry as well and unfortunately we're not able to identify these until the OnEnd event that we get from Otel... which means we can potentially create Transactions/Spans for Activities that we don't actually want to capture and send to Sentry. We're using the same strategy as all the other SDKs that have run into this same problem, which is simply never to finish these transactions/spans.

For this PR that you're working on then, it might be worth checking if a Transaction is one of these Otel generated transactions that we don't want to send, before automatically finishing it. 

If you look for references to `IsSentryRequest` you'll see how/where I've set that for spans.

src/Sentry.OpenTelemetry/SentrySpanProcessor.cs
```diff
@@ -105,10 +105,20 @@ public override void OnEnd(Activity data)
...
-                spanTracerToRemove.IsSentryRequest = true;
+                if (removed is SpanTracer spanTracerToRemove)
+                {
+                    spanTracerToRemove.IsSentryRequest = true;
+                }
+
+                if (removed is TransactionTracer transactionTracer)
+                {
+                    transactionTracer.IsSentryRequest = true;
+                }
```
jamescrosswell, line 119: You could do a switch here... I'm not sure if it's any more/less readable (up to you). 
```csharp
                switch (removed)
                {
                    case SpanTracer spanTracerToRemove:
                        spanTracerToRemove.IsSentryRequest = true;
                        break;
                    case TransactionTracer transactionTracer:
                        transactionTracer.IsSentryRequest = true;
                        break;
                }
```

src/Sentry/TransactionTracer.cs
jamescrosswell: @bruno-garcia suggested we add these extensions as/when customers had need of them (so as not to create a bunch of stuff we're maintaining but nobody is using).

@bitsandfoxes would you see a use case for having different timeouts for specific transactions on the Unity side, for example? If the default timeout was 3s then I could imagine someone wanting to bump it higher for a particular operation they knew was likely to take longer (e.g. a slow. performing query in some analytics database)... but that's just a hypothetical from me (it's not a real use case from an actual customer).

test/Sentry.Tests/HubTests.cs
```diff
@@ -1376,4 +1376,32 @@ public async Task WithScopeAsyncT_Works()
     }
 
     private static Scope GetCurrentScope(Hub hub) => hub.ScopeManager.GetCurrent().Key;
+
+    [Fact]
+    public async Task IdleTimeout_auto_finishes_transaction()
```
jamescrosswell, line 1381: Does this test something we're not already testing in `TransactionTests.Finish_SentryRequestTransactionGetsIgnored`?

jamescrosswell (commented):
Looking good!

src/Sentry/TransactionTracer.cs
bitsandfoxes: I don't see Unity needing anything additional timeouts.

test/Sentry.OpenTelemetry.Tests/SentrySpanProcessorTests.cs
```diff
@@ -282,19 +280,44 @@ public void OnEnd_FinishesTransaction()
...
+        sut.OnEnd(data);
+
+        // Assert
+        if (span is not TransactionTracer transaction)
+        {
+            Assert.Fail("Span is not a transaction tracer");
+            return;
         }
+
+        transaction.Contexts.ContainsKey("otel").Should().BeFalse();
```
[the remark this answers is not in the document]
SeanFeldman: Absolutely! I was too eager to declare that a done 😀

src/Sentry.OpenTelemetry/SentrySpanProcessor.cs
SeanFeldman: `switch` works as well. I'd go for it if it would be more than 2 cases. For 2 options, I'll keep it as is.

test/Sentry.Tests/HubTests.cs
SeanFeldman: Yes, it's redundant. Will remove.

bruno-garcia:
I'll be honest, I'm concerned this is opened for 3 weeks and is still in draft. But I haven't really looked at it since my first round of comments. @bitsandfoxes what are the next steps here?

test/Sentry.Tests/Protocol/TransactionTests.cs
```diff
@@ -526,4 +562,70 @@ public void ISpan_GetTransaction_FromSpan()
...
+            SpanStatus.Ok,
+            null,
+            true,
+            TransactionNameSource.Component
+        );
+
+        var transaction = new TransactionTracer(hub, context);
+
+        // Act
+        await Task.Delay(20);
```
jamescrosswell, line 593: @SeanFeldman is there any reason to use `await` here? Isn't the goal to make sure that the timer has had a chance to expire **_before_** continuing with the check in your test? You don't want this happening in the background right?

Maybe just `Thread.Sleep(20)` instead.

test/Sentry.Tests/Protocol/TransactionTests.cs
```diff
@@ -526,4 +562,70 @@ public void ISpan_GetTransaction_FromSpan()
...
+            SpanStatus.Ok,
+            null,
+            true,
+            TransactionNameSource.Component
+        );
+
+        var transaction = new TransactionTracer(hub, context, TimeSpan.FromMilliseconds(50));
+
+        // Act
+        await Task.Delay(5);
```
jamescrosswell, line 626: Same again... I think you need `Thread.Sleep` here instead of running this `async`.

test/Sentry.Tests/Protocol/TransactionTests.cs
jamescrosswell, line 593: @SeanFeldman [this stack overflow](https://stackoverflow.com/questions/31742521/accuracy-of-task-delay) might be relevant here.

src/Sentry/TransactionTracer.cs
```diff
@@ -234,6 +240,8 @@ public TransactionTracer(IHub hub, ITransactionContext context)
             _instrumenter = transactionContext.Instrumenter;
         }
 
+        var idleTimeout = idleTimeoutOverride ?? hub.GetSentryOptions()?.IdleTimeout;
```
bitsandfoxes, line 243:
```suggestion
```
When the transaction gets created and there is no idle timeout the transaction does not time out.

src/Sentry/TransactionTracer.cs
```diff
@@ -215,10 +215,16 @@ public TransactionTracer(IHub hub, string name, string operation, TransactionNam
...
     /// </summary>
-    public TransactionTracer(IHub hub, ITransactionContext context)
+    public TransactionTracer(IHub hub, ITransactionContext context) : this(hub, context, null)
+    {
+    }
+
+    /// <summary>
+    /// Initializes an instance of <see cref="TransactionTracer"/>.
+    /// </summary>
+    internal TransactionTracer(IHub hub, ITransactionContext context, TimeSpan? idleTimeoutOverride)
```
bitsandfoxes, line 225:
```suggestion
    internal TransactionTracer(IHub hub, ITransactionContext context, TimeSpan? idleTimeout = null)
```
By default, Transactions do not time out. But with this we have the option of providing a timeout.

src/Sentry/SentryOptions.cs
```diff
@@ -1008,6 +1008,11 @@ public bool JsonPreserveReferences
     /// </summary>
     internal bool InitNativeSdks { get; set; } = true;
 
+    /// <summary>
+    /// Idling transaction timeout after which transaction will be finished automatically.
+    /// </summary>
+    public TimeSpan? IdleTimeout { get; set; }
```
bitsandfoxes, line 1014:
```suggestion
    internal TimeSpan? IdleTimeout { get; set; }
```
I think we can keep this internal for now. Once we have an integration that actually uses this we can make it public.

bitsandfoxes (commented):
This looks good to me!

test/Sentry.Tests/Protocol/TransactionTests.cs
SeanFeldman: Validation based on time is tricky. I'm not sure `Thread.Sleep()` needs to replace `Task.Delay()`, @jamescrosswell.

Looking at the [SO answer by Stephen Cleary](https://stackoverflow.com/a/20084603/59563), I need a logical delay, not block the current thread. I will try to increase the delay from milliseconds to seconds to see if the issue is with such a low wait time.

src/Sentry/TransactionTracer.cs
SeanFeldman: @bitsandfoxes, isn't this problematic? If a `SentryOptions` has an IdleTimeout specified, it will be ignored as with this change we no longer take it into consideration.

test/Sentry.Tests/Protocol/TransactionTests.cs
SeanFeldman: Bumping time has worked. Can't use ms.

src/Sentry/TransactionTracer.cs
```diff
@@ -240,9 +240,11 @@ internal TransactionTracer(IHub hub, ITransactionContext context, TimeSpan? idle
             _instrumenter = transactionContext.Instrumenter;
         }
 
-
-        if (idleTimeout != null)
+        // Set idle timer only if idle timeout has been provided directly or via options
```
jamescrosswell, line 243: Can we simplify a bit?
```csharp
        if ((idleTimeout ?? hub.GetSentryOptions()?.IdleTimeout) is { } idleTimeoutToUse)
        {
            _idleTimer = new Timer(state =>
            {
                if (state is not TransactionTracer transactionTracer)
                {
                    return;
                }

                transactionTracer.Finish(Status ?? SpanStatus.Ok);
            }, this, idleTimeoutToUse, Timeout.InfiniteTimeSpan);
        }
```

jamescrosswell (approved):
Looks good to me!

test/Sentry.Tests/Protocol/TransactionTests.cs
```diff
@@ -388,6 +388,42 @@ public void Finish_SentryRequestSpansGetIgnored()
...
+            TransactionNameSource.Component
+        );
+
+        var transaction = new TransactionTracer(hub, context)
+        {
+            IsSentryRequest = true
+        };
+
+        // Act
+        await Task.Delay(5);
```
bitsandfoxes, line 421:
```suggestion
        await Task.Delay(TimeSpan.FromMilliseconds(5));
```
Nitpick just for consistency.

src/Sentry/SentryOptions.cs
```diff
@@ -1008,6 +1008,11 @@ public bool JsonPreserveReferences
     /// </summary>
     internal bool InitNativeSdks { get; set; } = true;
 
+    /// <summary>
+    /// Idling transaction timeout after which transaction will be finished automatically.
+    /// </summary>
+    internal TimeSpan? IdleTimeout { get; set; }
+
```
bitsandfoxes, line 1015:
```suggestion
```
This seems to cause a lot of confusion. Let's remove it and whenever auto instrumentation gets implemented we'll add it back in.

src/Sentry/TransactionTracer.cs
```diff
@@ -215,12 +234,25 @@ public TransactionTracer(IHub hub, ITransactionContext context)
...
         IsSampled = context.IsSampled;
-        StartTimestamp = _stopwatch.StartDateTimeOffset;
 
-        if (context is TransactionContext transactionContext)
+		if (context is TransactionContext transactionContext)
         {
             _instrumenter = transactionContext.Instrumenter;
         }
+
+        // Set idle timer only if idle timeout has been provided directly or via options
```
bitsandfoxes, line 243:
```suggestion
        // Set idle timer only if an idle timeout has been provided directly
```

test/Sentry.Tests/Protocol/TransactionTests.cs
```diff
@@ -388,6 +388,42 @@ public void Finish_SentryRequestSpansGetIgnored()
...
 
+    [Fact]
+    public async Task Finish_SentryRequestTransactionGetsIgnored()
+    {
+        // Arrange
+        var client = Substitute.For<ISentryClient>();
+        var options = new SentryOptions
+        {
+            Dsn = ValidDsn,
+            IdleTimeout = TimeSpan.FromMilliseconds(2)
```
bitsandfoxes, line 399:
```suggestion
```

test/Sentry.Tests/Protocol/TransactionTests.cs
```diff
@@ -388,6 +388,42 @@ public void Finish_SentryRequestSpansGetIgnored()
...
+            "my name",
+            "my operation",
+            "description",
+            SpanStatus.Ok,
+            null,
+            true,
+            TransactionNameSource.Component
+        );
+
+        var transaction = new TransactionTracer(hub, context)
```
bitsandfoxes, line 415:
```suggestion
        var transaction = new TransactionTracer(hub, context, TimeSpan.FromMilliseconds(2))
```

bitsandfoxes (changes requested):

test/Sentry.OpenTelemetry.Tests/SentrySpanProcessorTests.cs
```diff
@@ -282,19 +280,44 @@ public void OnEnd_FinishesTransaction()
...
-                transaction.Status.Should().Be(SpanStatus.Ok);
-            }
+                { "attributes", tags }
+            });
+            transaction.Status.Should().Be(SpanStatus.Ok);
+        }
+    }
+
+    [Fact]
+    public void OnEnd_does_not_finish_transaction_marked_as_IsSentryRequest()
```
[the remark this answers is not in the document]
jamescrosswell: >Ideally, where applicable, we'd like testnames to follow the scheme of `Method_Context_Expectation`

Aha... I was trying to work out what the convention was! This is a lightbulb moment for me 😄

src/Sentry/TransactionTracer.cs
```diff
@@ -215,12 +234,25 @@ public TransactionTracer(IHub hub, ITransactionContext context)
...
+        if ((idleTimeout ?? hub.GetSentryOptions()?.IdleTimeout) is { } idleTimeoutToUse)
```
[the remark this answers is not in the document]
SeanFeldman: Then what's the purpose of sentry options having `IdleTimeout`?
If it's never used, it should be removed, no?

test/Sentry.OpenTelemetry.Tests/SentrySpanProcessorTests.cs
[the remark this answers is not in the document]
SeanFeldman: 2nd that!

test/Sentry.Tests/Protocol/TransactionTests.cs
```diff
@@ -526,4 +562,72 @@ public void ISpan_GetTransaction_FromSpan()
...
+            SpanStatus.Ok,
+            null,
+            true,
+            TransactionNameSource.Component
+        );
+
+        var transaction = new TransactionTracer(hub, context);
+
+        // Act
+        await Task.Delay(TimeSpan.FromSeconds(2));
```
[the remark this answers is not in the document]
SeanFeldman: It was necessary to ensure the idle timeout is completed. But with such a short idle time, might not even need this. I'll verify it.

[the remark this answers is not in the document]
SeanFeldman: It is still needed. W/o it, the test goes through too fast, w/o the idle timer kicking in.

bitsandfoxes (approved):

bitsandfoxes:
🚀

src/Sentry/TransactionTracer.cs
```diff
@@ -278,6 +310,18 @@ private void AddChildSpan(SpanTracer span)
     /// <inheritdoc />
     public void Finish()
     {
+        if (Interlocked.Exchange(ref _idleTimerStopped, 1) == 0)
+        {
+            _idleTimer?.Change(Timeout.InfiniteTimeSpan, Timeout.InfiniteTimeSpan);
```
bruno-garcia, line 315: debug log here too is a good idea

jamescrosswell: See https://github.com/getsentry/sentry-dotnet/pull/2547

PR #2457, opened 2023-07-03
chore: Use `latest` runners for CI

bitsandfoxes:
Latest! The "only" caveat: `net48` running on Mono does not seem to be too happy, breaking the fixture due to `System.Exception : Library e_sqlite3 not found`. But I think skipping those should be fine for now.

#skip-changelog

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -56,7 +56,7 @@ jobs:
           SENTRY_AUTH_TOKEN: ${{ secrets.SENTRY_AUTH_TOKEN }}
 
       - name: Test
-        run: dotnet test Sentry-CI-Test.slnf -c Release --no-build --nologo -l GitHubActions -l "trx;LogFilePrefix=testresults_${{ runner.os }}"
+        run: dotnet test Sentry-CI-Test.slnf -r linux-x64 -c Release --no-build --nologo -l GitHubActions -l "trx;LogFilePrefix=testresults_${{ runner.os }}"
 
       - name: Upload broken test project
         if: failure()
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -15,10 +15,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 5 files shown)


jamescrosswell:
> Let's see what happens! :)

You broke Linux 😉 
```
Error: System.TypeInitializationException : The type initializer for 'Microsoft.Data.Sqlite.SqliteConnection' threw an exception.
---- System.Reflection.TargetInvocationException : Exception has been thrown by the target of an invocation.
-------- System.Exception : Library e_sqlite3 not found
```

SeanFeldman (approved):

.github/workflows/build.yml
```diff
@@ -15,10 +15,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        # Using ubuntu-20.04 because ubuntu-latest is 22.04, which is fails Sentry.DiagnosticSource.Tests due to a SQLite on Mono issue.
-        # Using macos-12 because we need Xcode 13.3 or later to build Sentry.Samples.Maui. (macos-latest currently points at macos-11 which uses Xcode 13.2)
-        # Using windows-2019 because windows-latest is much slower and we don't need anything in particular from it.
-        os: [ubuntu-20.04, windows-2019, macos-12]
+        os: [ubuntu-latest, windows-latest, macos-latest]
```
jamescrosswell, line 18: @bitsandfoxes does this run the risk of breaking our builds without us changing any of our code (since we can't be sure which os image will be used)? Maybe worth pinning to the "current" latest version of all of these things and deliberately testing/upgrading these in a branch when we do want to upgrade?

bitsandfoxes: That's a fair point. And I agree, especially in the tests or samples. But even the pinned ones change underneath without our knowledge (i.e the removal of .net core 3.1 on all OS)
I'd rather stay on top of it (there are PRs here daily) and pin/downgrade in case something breaks, rather than be forced to upgrade and fix something on top of it.

Issue #2480, opened 2023-07-14
Chore: Include Sentry.Samples.OpenTelemetry.AspNet in Sentry-CI-Build-Windows.slnf


jamescrosswell:
Currently `Sentry.Samples.OpenTelemetry.AspNet.csproj` isn't run by CI so we wouldn't know if bumping the version on dependencies or any other changes we make breaks this. We can't include it in `Sentry.sln` without breaking the developer experience for Mac users but we can have it in `SentryFull.sln` and the Windows CI solution filter.

PR #2482, opened 2023-07-14
chore: Added `OTel.AspNet` sample Windows filter

bitsandfoxes:
Resolves https://github.com/getsentry/sentry-dotnet/issues/2480

#skip-changelog


jamescrosswell:
@bitsandfoxes I'm happy to remove this from all our solutions... it sounds like it's more effort than it's worth. I opened https://github.com/getsentry/sentry-dotnet/issues/2488 for that so I think either we accept/commit that PR or this one (but not both)?

bitsandfoxes:
Superseded by https://github.com/getsentry/sentry-dotnet/pull/2489

PR #2487, opened 2023-07-17
Fixes baggage propagation when an exception is thrown from middleware

### LINKED ISSUE #2441: Baggage propagation doesn't seem to work when Exception captured in middleware
I'm updating NuGet Trends with the latest JavaScript (Angular) and .NET SDKs.

One of the things since the last update is the introduction of this feature: https://changelog.getsentry.com/announcements/connect-session-replays-to-backend-errors

Unfortunately it doesn't seem to work:

I throw an exception from my C# controller, but only the Frontend error shows on the Replay: 
image

Using the Replay Network details to capture the `baggage` header, I can confirm the JS SDK is sending it:

image

Value (from [replay](https://nugettrends.sentry.io/replays/spa:8f9bf18820be4450b62868dc460be8d8/?f_n_type=resource.xhr&n_detail_row=1&query=&referrer=%2Freplays%2F&statsPeriod=14d&t_main=network&yAxis=count%28%29)):
```
sentry-environment=local,sentry-transaction=%2F,sentry-public_key=85a592e835c64ca3a97d93776c12e947,sentry-trace_id=20b4f41dc98f49e2b25b5f88ffae1901,sentry-sample_rate=1,sentry-replay_id=8f9bf18820be4450b62868dc460be8d8
```

The exception captured by the .NET SDK included the envelope header:

```
[23:11:08 DBG] Envelope 'c0736a5e8f594176a65a9c0129b05cdd' successfully sent. Content: {"sdk":{"name":"sentry.dotnet","version":"3.33.1"},"event_id":"c0736a5e8f594176a65a9c0129b05cdd","sent_at":"2023-06-23T03:11:08.488511+00:00"}
{"type":"event","length":19132}
{"modules":{"System.Private.CoreLib":"7.0.0.0",.....
```

Note that replay_id isn't included.

## Investigation

The PR that added support here is: 
* https://github.com/getsentry/sentry-dotnet/pull/2332

Note that we look for a Span from an exception to support the case where a transaction is closed when unwinding the stack due to an exception bubbling up. Which is the case in ASP.NET Core where the root middleware captures the exception but further in the call chain we start/stop transactions:

https://github.com/getsentry/sentry-dotnet/blob/060480e1dfb42ca2044a8445c1d8e71b84b23f73/src/Sentry/Internal/Hub.cs#L275-L290

But on this PR adding support for DSC on exceptions, we don't consider the lookup, and only expect the transaction on the scope:

https://github.com/getsentry/sentry-dotnet/blob/060480e1dfb42ca2044a8445c1d8e71b84b23f73/src/Sentry/Internal/Hub.cs#L163-L166

I tried to validate that by not throwing from the controller, but instead just calling `CaptureMessage`. With a transaction on the scope, thing should work. That wasn't the case.

The exception included the DSC details, but no `replay_id`:

```
[23:44:21 DBG] Envelope '60ce0fe4d9294f27acbb429d3a7c858f' successfully sent. Content: {"sdk":{"name":"sentry.dotnet","version":"3.33.1"},"event_id":"60ce0fe4d9294f27acbb429d3a7c858f","trace":{"trace_id":"0c6188d56e9e49ffba45e31e5b697b3c","public_key":"57331596a25b4c3da49750b292299e09","sample_rate":"1","release":"NuGetTrends.Web@1.0.0\u002Ba678a0d59db0bcef71578b7c341c10c48ac5c80f","environment":"development","transaction":"GET api/package/history/{id}"},"sent_at":"2023-06-23T03:44:21.916599+00:00"}

```

jamescrosswell:
Fixes #2441

### DIFF
--- a/src/Sentry/HubExtensions.cs
+++ b/src/Sentry/HubExtensions.cs
@@ -1,3 +1,4 @@
+using Sentry.Extensibility;
 using Sentry.Infrastructure;
 using Sentry.Internal;
 using Sentry.Internal.Extensions;
@@ -227,10 +228,12 @@ public static class HubExtensions
         this IHub hub,
         ITransactionContext context,
         IReadOnlyDictionary<string, object?> customSamplingContext,
-        DynamicSamplingContext? dynamicSamplingContext)
-        => hub is Hub fullHub
-            ? fullHub.StartTransaction(context, customSamplingContext, dynamicSamplingContext)
-            : hub.StartTransaction(context, customSamplingContext);
+        DynamicSamplingContext? dynamicSamplingContext) => hub switch
+        {
+            Hub fullHub => fullHub.StartTransaction(context, customSamplingContext, dynamicSamplingContext),
+            HubAdapter adapter => adapter.StartTransaction(context, customSamplingContext, dynamicSamplingContext),
+            _ => hub.StartTransaction(context, customSamplingContext)
+        };
 
     internal static ITransaction? GetTransaction(this IHub hub)
     {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Fixes
+
+- Fixes baggage propagation when an exception is thrown from middleware ([#2487](https://github.com/getsentry/sentry-dotnet/pull/2487))
+
 ### Dependencies
 
 - Bump Cocoa SDK from v8.8.0 to v8.9.0 ([#2479](https://github.com/getsentry/sentry-dotnet/pull/2479))
--- a/src/Sentry/Extensibility/HubAdapter.cs
+++ b/src/Sentry/Extensibility/HubAdapter.cs
@@ -79,6 +79,16 @@ public sealed class HubAdapter : IHub, IHubEx
     /// Forwards the call to <see cref="SentrySdk"/>.
     /// </summary>
     [DebuggerStepThrough]
+    internal ITransaction StartTransaction(
+        ITransactionContext context,
+        IReadOnlyDictionary<string, object?> customSamplingContext,
+        DynamicSamplingContext? dynamicSamplingContext)
+        => SentrySdk.StartTransaction(context, customSamplingContext, dynamicSamplingContext);
+
+    /// <summary>
+    /// Forwards the call to <see cref="SentrySdk"/>.
+    /// </summary>
+    [DebuggerStepThrough]
     public void BindException(Exception exception, ISpan span) =>
         SentrySdk.BindException(exception, span);
 
--- a/src/Sentry/Internal/Hub.cs
+++ b/src/Sentry/Internal/Hub.cs
@@ -344,6 +344,11 @@ internal class Hub : IHubEx, IDisposable
                 evt.Contexts.Trace.SpanId = linkedSpan.SpanId;
                 evt.Contexts.Trace.TraceId = linkedSpan.TraceId;
                 evt.Contexts.Trace.ParentSpanId = linkedSpan.ParentSpanId;
+                // Try to read the DSC from the transaction associated with the linked span
+                if (linkedSpan?.GetTransaction() is TransactionTracer transactionTracer)
+                {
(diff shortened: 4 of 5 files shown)


src/Sentry/HubExtensions.cs
```diff
@@ -227,10 +228,12 @@ public LockedScope(IHub hub)
...
         IReadOnlyDictionary<string, object?> customSamplingContext,
-        DynamicSamplingContext? dynamicSamplingContext)
-        => hub is Hub fullHub
-            ? fullHub.StartTransaction(context, customSamplingContext, dynamicSamplingContext)
-            : hub.StartTransaction(context, customSamplingContext);
+        DynamicSamplingContext? dynamicSamplingContext) => hub switch
+        {
+            Hub fullHub => fullHub.StartTransaction(context, customSamplingContext, dynamicSamplingContext),
+            HubAdapter adapter => adapter.StartTransaction(context, customSamplingContext, dynamicSamplingContext),
+            _ => hub.StartTransaction(context, customSamplingContext)
```
bitsandfoxes, line 235: I don't get what's happening here. Could you explain?
Do we still need that in the first place? It looks to me like we can drop the whole extension method. What other types besides `Hub` and `HubAdapter` are we expecting?

bruno-garcia: I wonder why does it matter the type at all? Should the overload with DS be in the public interface? Or that information be part of `customSamplingContext` (even if `internal`)?

jamescrosswell: > I don't get what's happening here. Could you explain?

There's a bit of an [explanation here](https://github.com/getsentry/sentry-dotnet/issues/2441#issuecomment-1637207121). Essentially the DSC was being dropped because of the codepath that extension forced things down.

jamescrosswell: > I wonder why does it matter the type at all? Should the overload with DS be in the public interface? Or that information be part of `customSamplingContext` (even if `internal`)?

It's pretty unlikely any of our SDK users would want to use this overload... which is why I steered away from putting it on the `IHub` interface. 

We could put it on `IHubEx`, which is an internal interface only implemented by `Hub` and `HubAdapter` (not by `DisabledHub`). 

Potentially we could have `DisabledHub` implement `IHubEx` as well... instead of having a bunch of `switch` expressions all over the place. The `HubAdapter` seems to pass quite a lot of work off to the `SentrySdk`, which in turn just pushes it through to the `CurrentHub`. [`HubExtensions`](https://github.com/getsentry/sentry-dotnet/blob/78d33f8d3052927152ea09cc856b63d369275db5/src/Sentry/HubExtensions.cs#L190) and  [`SentrySDK`](https://github.com/getsentry/sentry-dotnet/blob/78d33f8d3052927152ea09cc856b63d369275db5/src/Sentry/SentrySdk.cs#L350) both employ switch expression in multiple locations to determine whether to call a **_real_** internal method or the public **_disabled_** method (that pipes to `/dev/null` or returns `default`). That could probably be achieved more simply by having `DisabledHub` do the same directly if it implemented that internal `IHubEx` interface. 

I can take a look at trying to refactor that if you like, but probably as part of a separate issue to this, as it's not really related to fixing this particular issue. 

As an aside, I was a bit surprised to encounter an instance of `HubAdapter` when running an ASP.NET Core application. [The XML docs](https://github.com/getsentry/sentry-dotnet/blob/78d33f8d3052927152ea09cc856b63d369275db5/src/Sentry/Extensibility/HubAdapter.cs#L6C1-L15C46) for that class indicate it was created to help with testing... it seems to offload all it's work to the `SentrySdk`, which uses a single static Hub instance. Outside of testing, in what scenarios would we expect people to be using the `SentrySdk.CurrentHub` vs scoped/transient instances of a `Hub`?

jamescrosswell:
> The one point we're discussing isn't related to this change (the type check was there to begin with) so I'm happy to merge this fix and address that oddness in a follow up PR

Created [Chore: Refactor HubExtensions and SentrySdk #2490](https://github.com/getsentry/sentry-dotnet/issues/2490) to follow up.

PR #2493, opened 2023-07-18
feat: Tracing without performance

bitsandfoxes:
Resolves https://github.com/getsentry/sentry-dotnet/issues/2447
Implementations in other SDKs:
- Java https://github.com/getsentry/sentry-java/pull/2788
- PHP https://github.com/getsentry/sentry-php/pull/1516

The idea here is:
The scope now has a `SentryPropagationContext`. The context contains minimally a `traceID` and a `spanID`.  `ContinueTrace` overwrites the context on the current scope, making it possible to propagate the IDs without the performance feature enabled.
Relevant top-level methods: `GetTraceHeader`, `GetBaggage`. The methods first look for active spans before falling back to creating their return value from the PropagationContext. The same is true for `CaptureEvent` where we check for linked or active spans first.


src/Sentry/SentrySdk.cs
```diff
@@ -614,9 +614,34 @@ public static void BindException(Exception exception, ISpan span)
     /// Gets the Sentry trace header.
     /// </summary>
     [DebuggerStepThrough]
+    // [Obsolete("This method will be removed in a future major version. Use the `GetTraceParent` instead.")]
```
jamescrosswell, line 617: @bitsandfoxes I'm wondering why `GetTraceParent` is a better name than `GetTraceHeader`. It looks like this method is returning a trace header... that header does include the parentid, so I guess it's the parent. Maybe `GetParentTraceHeader` is a more explicit name for what this is then (vs the TraceHeader that we'll be propagating downstream)?

src/Sentry/DynamicSamplingContext.cs
```diff
@@ -127,6 +127,14 @@ public static DynamicSamplingContext CreateFromTransaction(TransactionTracer tra
...
             transactionName);
     }
+
+    public static DynamicSamplingContext CreateFromPropagationContext(SentryPropagationContext propagationContext, SentryOptions options)
+    {
+        var publicKey = Dsn.Parse(options.Dsn!).PublicKey;
+        var traceId = propagationContext.TraceId;
+
+        return new DynamicSamplingContext(traceId, publicKey, 1.0f);
+    }
```
jamescrosswell, line 137: @bitsandfoxes I think you'll want to propagate/set the [new `sampled` property](https://github.com/getsentry/sentry-dotnet/blob/55d4f22e81861a2e0982cdf96a79c4135e8f03e1/src/Sentry/DynamicSamplingContext.cs#L25) in the DSC as well.

src/Sentry/SentrySdk.cs
bitsandfoxes: I could not find `GetTraceHeader` in other SDKs so this is something unique to the .NET SDK. Right now I'm just following the unified API to keep things consistent with the requirements. But maybe we opt into reusing the method to avoid the "rename". 
Also, the function changes slightly: `GetTraceHeader` would return the trace header of the last active span or null if there is no present. Since there's now always the `PropagationContext` available on the scope `GetTraceParent` will always a trace header.

src/Sentry/DynamicSamplingContext.cs
bitsandfoxes: We do. With this change, all errors and transactions sent should always contain a DSC.

src/Sentry/Internal/Hub.cs
```diff
@@ -360,9 +409,31 @@ SentryId IHubEx.CaptureEventInternal(SentryEvent evt, Hint? hint, Scope? scope)
...
+            {
+                // Inject trace information from a linked span
+                if (GetLinkedSpan(evt, actualScope) is { } linkedSpan)
+                {
+                    evt.Contexts.Trace.SpanId = linkedSpan.SpanId;
+                    evt.Contexts.Trace.TraceId = linkedSpan.TraceId;
+                    evt.Contexts.Trace.ParentSpanId = linkedSpan.ParentSpanId;
+                }
+
+                // When a transaction is present, copy its DSC to the event.
```
bruno-garcia, line 423: This isn't super helpful, but I suggested comments above could. To clarify: Maybe it's my lack of understanding of `IsTracingEnabled`. Is this meaning `IsSentryPeformanceMonitoringWithTrasnactions` instead? Because what both sides of the `if` do are tracing

src/Sentry/Sentry.csproj
```diff
@@ -94,27 +94,13 @@
...
-      <SentryCLIDownload
-        Condition="'$(CI_PUBLISHING_BUILD)' == 'true' Or ($([MSBuild]::IsOSPlatform('Linux')) And $(_OSArchitecture) == 'X64')"
-        Include="sentry-cli-Linux-x86_64" FileHash="a9fb79e44c5bae6ca8dfd2c66ac918c7e0405e3456edeb100d698961842f057f" />
-      <SentryCLIDownload
-        Condition="'$(CI_PUBLISHING_BUILD)' == 'true' Or ($([MSBuild]::IsOSPlatform('Windows')) And $(_OSArchitecture) == 'X86')"
-        Include="sentry-cli-Windows-i686.exe" FileHash="db51f4ce5cddaea762b54cd5527166f0045842e4558e315fc971f63687850cb6" />
-      <SentryCLIDownload
-        Condition="'$(CI_PUBLISHING_BUILD)' == 'true' Or ($([MSBuild]::IsOSPlatform('Windows')) And $(_OSArchitecture) != 'X86')"
-        Include="sentry-cli-Windows-x86_64.exe" FileHash="769cace3b97e583a185b849b4a47bd9590d0ecfaea94022135301b8783558aa7" />
+      <SentryCLIDownload Condition="'$(CI_PUBLISHING_BUILD)' == 'true' Or ($([MSBuild]::IsOSPlatform('OSX')) And $(_OSArchitecture) == 'Arm64')" Include="sentry-cli-Darwin-arm64" FileHash="a2bd6322b774b9ecfdc0adab53d9d76914bb10eab2ec7d2380b25c0300c428f8" />
```
bruno-garcia, line 97: is this related to the change? Can't tell if anythign changed

src/Sentry/SentryPropagationContext.cs
```diff
@@ -0,0 +1,58 @@
...
+namespace Sentry;
+
+internal class SentryPropagationContext
+{
+    public SentryId TraceId { get; }
+    public SpanId SpanId { get; }
+    public SpanId? ParentSpanId { get; }
+    public DynamicSamplingContext? DynamicSamplingContext { get; set; }
+
+    internal static IDiagnosticLogger? Logger { get; set; } = SentrySdk.CurrentOptions?.DiagnosticLogger;
```
bruno-garcia, line 12: Can you pass explicitly the logger instead of statically accessing it? The whole design avoided static access to make things easier to reason about and unit test.

The class is internal so we could just take the logger as an argument on methods that need it, have the caller provide it?

I might be too pedantic here since this is all internal code we can change anytime so if it works, it works, but I'd at least suggest:

```suggestion
    internal static IDiagnosticLogger? Logger => SentrySdk.CurrentOptions?.DiagnosticLogger;
```

So it picks up on changes to the logger implementation at runtime.

src/Sentry/SentrySdk.cs
bruno-garcia: also: https://www.w3.org/TR/trace-context/#traceparent-header

bruno-garcia (commented):
Great work! left some notes

src/Sentry/BaggageHeader.cs
```diff
@@ -8,7 +8,7 @@ namespace Sentry;
 /// </summary>
 /// <seealso href="https://develop.sentry.dev/sdk/performance/dynamic-sampling-context"/>
 /// <seealso href="https://www.w3.org/TR/baggage"/>
-internal class BaggageHeader
+public class BaggageHeader
```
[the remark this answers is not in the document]
jamescrosswell: See [Activity.Baggage](https://learn.microsoft.com/en-gb/dotnet/api/system.diagnostics.activity.baggage?view=net-7.0)... so there's no specific class for this in .NET - just an IEnumerable of string KVPs:

```csharp
public IEnumerable<KeyValuePair<string,string?>> Baggage { get; }
```

src/Sentry/Internal/Hub.cs
```diff
@@ -360,9 +409,31 @@ SentryId IHubEx.CaptureEventInternal(SentryEvent evt, Hint? hint, Scope? scope)
...
+                transaction = actualScope.Transaction as TransactionTracer;
+                evt.DynamicSamplingContext = transaction?.DynamicSamplingContext;
```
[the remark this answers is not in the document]
jamescrosswell: That was one of the changes we made in this [recent fix to baggage propagation](https://github.com/getsentry/sentry-dotnet/commit/d6d531cb7d3699f7c4f1f7225433928fbcdab00d) (see changes to Hub.cs in that commit).

src/Sentry/Sentry.csproj
jamescrosswell: Nah, Rider seems to want to remove the Carriage Returns. I've been trying to undo this in every commit that has other (unrelated) changes to the project file since I started working on this project. It'd be great if we could either find a way to stop Rider from doing that or just go with the flow. 

FWIW I actually prefer the way Rider wants this (all on one line) since it makes it easy to compare/contrast across multiple tags/lines.

src/Sentry/SentrySdk.cs
jamescrosswell: Hm... that's even more confusing. Our `SentryPropagationContext` is propagating the `sentry-trace` header right (not the W3C `traceparent` header)?

jamescrosswell (commented):
Looking good! Not too much to add to the comments Bruno left already.

src/Sentry/Internal/Hub.cs
bitsandfoxes: Yes. Effectively, the scope  now contains a traceID in its `PropagationContext`. This turns `IsTracingEnabled` into `IsSentryPeformanceMonitoringWithTrasnactions`.
This also means that `Tracing without Performance` is not controlled by any option.

src/Sentry/Internal/Hub.cs
```diff
@@ -195,7 +195,64 @@ public void BindException(Exception exception, ISpan span)
...
+            return propagationContext.DynamicSamplingContext.ToBaggageHeader();
+        }
+
+        return null;
+    }
+
+    public TransactionContext? ContinueTrace(string? sentryTraceHeader, string? baggageHeaders)
+    {
+        var propagationContext = SentryPropagationContext.CreateFromHeaders(sentryTraceHeader, baggageHeaders);
+        ConfigureScope(scope => scope.PropagationContext = propagationContext);
```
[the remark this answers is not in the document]
bitsandfoxes: I think so. The prop splits where the actual work is going to get done: Within the the "performance" feature or outside of it.

src/Sentry/Internal/Hub.cs
```diff
@@ -195,7 +195,64 @@ public void BindException(Exception exception, ISpan span)
...
+
+        if (_options.IsTracingEnabled)
```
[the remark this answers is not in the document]
bitsandfoxes: I don't think so. The above is the minimum to make `Tracing without Performance` work. The part guarded is the performance part.

src/Sentry/Internal/Hub.cs
jamescrosswell: Just checking my understanding:
* `IsTracingEnabled == true` => Tracing: `Enabled`; Sentry Performance Monitoring: `Enabled`
* `IsTracingEnabled == false` => Tracing: `Enabled`; Sentry Performance Monitoring: `Disabled`

Is that correct? If so then Tracing is now always enabled and that flag in the options probably needs to be renamed to be less confusing (maybe `IsPerformanceMonitoringEnabled`). We could mark the old one as obsolete and introduce a new one instead.

src/Sentry.AspNetCore/SentryMiddleware.cs
```diff
@@ -18,6 +18,10 @@ namespace Sentry.AspNetCore;
 /// </summary>
 internal class SentryMiddleware : IMiddleware
 {
+    internal const string TraceHeaderItemKey = "sentry-trace-header";
```
jamescrosswell, line 21: Why do we have a separate header name here to [the one used in SentryTraceHeader](https://github.com/getsentry/sentry-dotnet/blob/dc0e0223aa64555966ff5c9ab194a535dc737b72/src/Sentry/SentryTraceHeader.cs#L8)?

src/Sentry.AspNetCore/SentryMiddleware.cs
```diff
@@ -91,12 +95,14 @@ public async Task InvokeAsync(HttpContext context, RequestDelegate next)
...
-            {
-                var traceHeader = context.TryGetSentryTraceHeader(_options);
-                var baggageHeader = context.TryGetBaggageHeader(_options);
-                hub.ContinueTrace(traceHeader, baggageHeader);
-            }
+            var traceHeader = context.TryGetSentryTraceHeader(_options);
+            var baggageHeader = context.TryGetBaggageHeader(_options);
+            var transactionContext = hub.ContinueTrace(traceHeader, baggageHeader);
+
+            // Adding the headers and thw TransactionContext to the context to be picked up by the Sentry tracing middleware
```
jamescrosswell, line 102:
```suggestion
            // Adding the headers and the TransactionContext to the context to be picked up by the Sentry tracing middleware
```

src/Sentry.AspNet/HttpContextExtensions.cs
```diff
@@ -65,13 +85,19 @@ public static ITransaction StartSentryTransaction(this HttpContext httpContext)
...
         var transactionName = $"{method} {path}";
         const string transactionOperation = "http.server";
 
-        var transactionContext = traceHeader is not null
-            ? new TransactionContext(transactionName, transactionOperation, traceHeader, TransactionNameSource.Url)
-            : new TransactionContext(transactionName, transactionOperation, TransactionNameSource.Url);
+        var transactionContext = SentrySdk.ContinueTrace(traceHeader, baggageHeader, transactionName, transactionOperation);
+        if (transactionContext is null)
+        {
+            throw new NullReferenceException("Failed to create a transaction context. " + "Tracing was " +
```
jamescrosswell, line 96:
```suggestion
            throw new NullReferenceException("Failed to create a transaction context. Tracing was " +
```

src/Sentry.AspNetCore/SentryMiddleware.cs
```diff
@@ -91,12 +95,14 @@ public async Task InvokeAsync(HttpContext context, RequestDelegate next)
...
+            context.Items.Add(TraceHeaderItemKey, traceHeader);
+            context.Items.Add(BaggageHeaderItemKey, baggageHeader);
+            context.Items.Add(TransactionContextItemKey, transactionContext);
```
jamescrosswell, line 105: This is kind of an interesting (ab)use of `HttpContext.Items`... we're sort of using it a bit like a Post Office Box here to send stuff between our various Middleware. This will probably work without any issues (as long as the Key's you're picking never conflict with anything), but I don't think that's [what it was designed for](https://learn.microsoft.com/en-us/dotnet/api/system.web.httpcontext.items?view=netframework-4.8.1). 

Another (I think cleaner) option would be [ConditionalWeakTable](https://learn.microsoft.com/dotnet/api/system.runtime.compilerservices.conditionalweaktable-2). You could store the the `HttpContext` as the Key in that table and maybe a tupple holding the three values that you want to pass as the Value. 

You can see an [example of this in our codebase here](https://github.com/getsentry/sentry-dotnet/blob/dc0e0223aa64555966ff5c9ab194a535dc737b72/src/Sentry/Internal/Hub.cs#L21).

src/Sentry/Internal/Hub.cs
jamescrosswell: OK, I think worth doing that in this pull request then... it's certain to cause confusion otherwise.

src/Sentry.AspNetCore/SentryMiddleware.cs
bitsandfoxes: Based on [this article](https://learn.microsoft.com/en-us/aspnet/core/fundamentals/app-state?view=aspnetcore-7.0#httpcontextitems) the `items` seem to be the mechanism with which to pass data between middlewares.
I guess we could use objects to use as keys to be extra sure not to have any conflicting keys?

src/Sentry/Extensibility/DisabledHub.cs
```diff
@@ -87,11 +87,15 @@ public void BindException(Exception exception, ISpan span)
...
-    public TransactionContext? ContinueTrace(
+    public TransactionContext ContinueTrace(
         SentryTraceHeader? traceHeader,
         BaggageHeader? baggageHeader,
         string? name = null,
-        string? operation = null) => null;
+        string? operation = null)
+    {
+        // Transactions from DisabledHub are always sampled out
+        return new TransactionContext(string.Empty, string.Empty, false);
```
jamescrosswell, line 97: I think we could do the same [thing we've done for the `DisabledHub.Instance`](https://github.com/getsentry/sentry-dotnet/blob/dfc4274fc79cdd1472649387bee6feee713343cf/src/Sentry/Extensibility/DisabledHub.cs#L11) here, to avoid creating a new object every time this method is called.

Not sure where to put that... maybe `TransactionContext.Disabled` or `TransactionContext.DisabledInstance`?

src/Sentry/SentryPropagationContext.cs
```diff
@@ -7,20 +7,33 @@ internal class SentryPropagationContext
...
+        get => _dynamicSamplingContext;
+        set
+        {
+            if (_dynamicSamplingContext is null)
+            {
+                _dynamicSamplingContext = value;
+            }
+            else
+            {
+                throw new Exception("should not do that.");
```
jamescrosswell, line 23:
```suggestion
                throw new Exception("Attempted to set the DynamicSamplingContext but the context exists already.");
```

jamescrosswell (commented):
I didn't do a full review - just the files that have changed since I last checked a few days ago. All looking good. 

My only question is what we plan to do with the [EnableTracing option](https://github.com/getsentry/sentry-dotnet/pull/2493#discussion_r1274148072).

Otherwise, once the Check failures from CI are addressed, I think it's looking good.

jamescrosswell:
A bit annoying, but you have to run a build on a Windows box to get the verify tests running (and the file you need to commit) to fix this:

image

jamescrosswell (approved):

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 ### Features
 
+- Distributed tracing now workes independently of the performance feature. This allows you to connect errors to other Sentry instrumented applications ([#2493](https://github.com/getsentry/sentry-dotnet/pull/2493))
```
```suggestion
- Distributed tracing now works independently of the performance feature. This allows you to connect errors to other Sentry instrumented applications ([#2493](https://github.com/getsentry/sentry-dotnet/pull/2493))
```

PR #2496, opened 2023-07-19
Align SDK with docs regarding session update for dropped events

jamescrosswell:
Completes [Align SDK with docs regarding session update for dropped events #1599](https://github.com/getsentry/sentry-dotnet/issues/1599).

All of this now happens in the `SentryClient.DoSendEvent` method in the .NET SDK. 

### .NET

The [CaptureEvent_Processing_Order](https://github.com/getsentry/sentry-dotnet/blob/46865c6919f8c54528b17aa36894203a929f6808/test/Sentry.Tests/SentryClientTests.cs#L617) ensures .NET SDK processing happens in the same order.

### DIFF
--- a/src/Sentry/SentryClient.cs
+++ b/src/Sentry/SentryClient.cs
@@ -211,16 +211,6 @@ public class SentryClient : ISentryClient, IDisposable
     // TODO: this method needs to be refactored, it's really hard to analyze nullability
     private SentryId DoSendEvent(SentryEvent @event, Hint? hint, Scope? scope)
     {
-        if (_options.SampleRate != null)
-        {
-            if (!_randomValuesFactory.NextBool(_options.SampleRate.Value))
-            {
-                _options.ClientReportRecorder.RecordDiscardedEvent(DiscardReason.SampleRate, DataCategory.Error);
-                _options.LogDebug("Event sampled.");
-                return SentryId.Empty;
-            }
-        }
-
         var filteredExceptions = ApplyExceptionFilters(@event.Exception);
         if (filteredExceptions?.Count > 0)
         {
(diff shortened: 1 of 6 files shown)


src/Sentry/SentryClient.cs
```diff
@@ -281,6 +267,34 @@ private SentryId DoSendEvent(SentryEvent @event, Hint? hint, Scope? scope)
...
+            {
+                _options.ClientReportRecorder.RecordDiscardedEvent(DiscardReason.SampleRate, DataCategory.Error);
+                _options.LogDebug("Event sampled.");
+                return SentryId.Empty;
+            }
+        }
+        else
+        {
+            _options.LogDebug("Event not sampled.");
+        }
```
bitsandfoxes, line 296:
```suggestion
```
Alternatively, this block could be moved one line up?

test/Sentry.Tests/SentryClientTests.cs
```diff
@@ -612,6 +613,68 @@ public void CaptureEvent_WithSampleRate_AppropriateDistribution(float sampleRate
         });
     }
 
+    [Fact]
+    public void CaptureEvent_Processing_Order()
```
bitsandfoxes, line 617: That looks like it's been fun figuring out!

bitsandfoxes (approved):

bitsandfoxes:
LGTM!

src/Sentry/SentryClient.cs
jamescrosswell: It's mainly there for testing... so that we can ensure the order of processing when the sample rate is null. 

The problem with putting it in the block where the sample rate is not null is that it would make the tests non-deterministic (you're rolling dice to see whether the tests pass then). 

Or did you have something else in mind?

test/Sentry.Tests/SentryClientTests.cs
jamescrosswell: Yes, deep in the weeds of NSubstitute!

PR #2503, opened 2023-07-21
Use 'HttpMessageHandler' instead of 'HttpClientHandler' to support NSUrlSessionHandler on iOS

### LINKED ISSUE #2502: SentryOptions should use 'HttpMessageHandler' instead of 'HttpClientHandler' to support NSUrlSessionHandler on iOS
### Problem Statement

The SentryOptions should accept the more general HttpMessageHandler instead of HttpClientHandler to provide to ability to use a custom-configured handler for the HttpClient.

Background:
It is recommended to use NSUrlSessionHandler on iOS (TLS 1.2, performance), but NSUrlSessionHandler only implements HttpMessageHandler.

Since NSUrlSessionHandler inherits from HttpMessageHandler and DefaultSentryHttpClientFactory also mainly uses this class to create the HttpClient, changes are going to be minor.

During creation of the HttpClient the DefaultSentryHttpClientFactory can take care of if it is a HttpClientHandler, a HttpMessageHandler or a NSUrlSessionHandler (which also accepts proxy and decompression properties.

### Solution Brainstorm

Going to file a PR.

AlexSchuetz:
fixing #2502 

Breaking change:
 - SentryOptions.HttpClientHandlerFactory moved to SentryOptions.HttpMessageHandlerFactory

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 - Fixed baggage propagation when an exception is thrown from middleware ([#2487](https://github.com/getsentry/sentry-dotnet/pull/2487))
 - Fix Durable Functions preventing orchestrators from completing ([#2491](https://github.com/getsentry/sentry-dotnet/pull/2491))
 - Re-enable HubTests.FlushOnDispose_SendsEnvelope ([#2492](https://github.com/getsentry/sentry-dotnet/pull/2492))
+- Fixes #2502 SentryOptions should use 'HttpMessageHandler' instead of 'HttpClientHandler' to support NSUrlSessionHandler on iOS  ([#2503](https://github.com/getsentry/sentry-dotnet/pull/2503))
 
 ### Dependencies
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,7 @@
 - Added a MSBuild property `SentryUploadAndroidProguardMapping` to automatically upload the Proguard mapping file when targeting Android ([#2455](https://github.com/getsentry/sentry-dotnet/pull/2455))
 - Symbolication for Single File Apps ([#2425](https://github.com/getsentry/sentry-dotnet/pull/2425))
 - Add binding to `SwiftAsyncStacktraces` on iOS ([#2436](https://github.com/getsentry/sentry-dotnet/pull/2436))
+- Support transaction finishing automatically with 'idle timeout' ([#2452](https://github.com/getsentry/sentry-dotnet/pull/2452))
 
 ### Fixes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 - Fix Durable Functions preventing orchestrators from completing ([#2491](https://github.com/getsentry/sentry-dotnet/pull/2491))
 - Re-enable HubTests.FlushOnDispose_SendsEnvelope ([#2492](https://github.com/getsentry/sentry-dotnet/pull/2492))
 - Fixes #2502 SentryOptions should use 'HttpMessageHandler' instead of 'HttpClientHandler' to support NSUrlSessionHandler on iOS  ([#2503](https://github.com/getsentry/sentry-dotnet/pull/2503))
+- Fixed the SDK failing to report issues via Blazor WebAssembly due to a `PlatformNotSupportedException` ([#2506](https://github.com/getsentry/sentry-dotnet/pull/2506))
 
 ### Dependencies
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,9 +17,9 @@
 - Bump Java SDK from v6.25.1 to v6.25.2 ([#2484](https://github.com/getsentry/sentry-dotnet/pull/2484))
   - [changelog](https://github.com/getsentry/sentry-java/blob/main/CHANGELOG.md#6252)
   - [diff](https://github.com/getsentry/sentry-java/compare/6.25.1...6.25.2)
-- Bump Cocoa SDK from v8.8.0 to v8.9.1 ([#2479](https://github.com/getsentry/sentry-dotnet/pull/2479), [#2483](https://github.com/getsentry/sentry-dotnet/pull/2483))
-  - [changelog](https://github.com/getsentry/sentry-cocoa/blob/main/CHANGELOG.md#891)
-  - [diff](https://github.com/getsentry/sentry-cocoa/compare/8.8.0...8.9.1)
+- Bump Cocoa SDK from v8.8.0 to v8.9.2 ([#2479](https://github.com/getsentry/sentry-dotnet/pull/2479), [#2483](https://github.com/getsentry/sentry-dotnet/pull/2483), [#2500](https://github.com/getsentry/sentry-dotnet/pull/2500))
+  - [changelog](https://github.com/getsentry/sentry-cocoa/blob/main/CHANGELOG.md#892)
+  - [diff](https://github.com/getsentry/sentry-cocoa/compare/8.8.0...8.9.2)
 
 ## 3.34.0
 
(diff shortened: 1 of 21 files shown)


bitsandfoxes:
That's a great catch, thanks for contributing!
We'd like to avoid introducing breaking changes outside of mayor releases and the next one is still a bit out.
I'm thinking, could this be added as an additional option and we mark the old one as `obsolete`?

AlexSchuetz:
Of course. That's a good idea. I will adjust it shortly. 👍

jamescrosswell:
Thanks heaps @AlexSchuetz, the commit all looks great!

There are a couple of things that need doing to get things through our CI servers though.

### Changelog

Could you add the following to the CHANGELOG.md in the Fixes section?

```
- Fixes #2502 SentryOptions should use 'HttpMessageHandler' instead of 'HttpClientHandler' to support NSUrlSessionHandler on iOS  ([#2503](https://github.com/getsentry/sentry-dotnet/pull/2503))
```

### Update verify expectations
We're using [Verify](https://github.com/VerifyTests/Verify) to make sure that any changes to the API surface area are deliberate. That means when the API does change, we need to update the relevant verified.txt files. The easiest way to create updated versions of those is to run `./build.sh` or `build.cmd` (depending on your development platform) from the repository root folder. So if you're developing on MacOS, for example, you'd run `./build.sh`. Once you've done that, I think you should see changes to commit for the following files:
- test/Sentry.Tests/ApiApprovalTests.Run.Net4_8.verified.txt
- test/Sentry.Tests/ApiApprovalTests.Run.Core3_1.verified.txt
- test/Sentry.Tests/ApiApprovalTests.Run.DotNet6_0.verified.txt
- test/Sentry.Tests/ApiApprovalTests.Run.DotNet7_0.verified.txt

You can add those to a new commit with a description like "updated verify files" or something like that and push.

**_Note:_** You can ignore any errors the build shows relating to profiling... that's a long story and unrelated to this PR.

### Failed Checks
There are a [couple of other checks that are failing](https://github.com/getsentry/sentry-dotnet/actions/runs/5623842449/job/15273647829?pr=2503). I'm not actually sure what's going on there. @bitsandfoxes do you have any context for that?

AlexSchuetz:
I did the required commit and also had a look at the error. NSUrlSessionHandler does not support a proxy, but I want to include the code for setting it anyway, because if it ever does noone will remember adjusting it here.
Thus I implemented it and placed it inside of a try-catch. So nothing should go wrong. I now adjusted my implementation and check NSUrlSessionHandler.SupportsProxy. Hopefully this works.

AlexSchuetz:
I investigated a bit more. It is a code analysis error. Since NSUrlSessionHandler.Proxy is not supported.
I added a pragma for this one line to disable this issue, since I want to support it as soon as NSUrlSesseionHandler does.

If you think I should remove the code in question and just print out that a proxy for NSUrlSessionHandler is not supported, that would be fine as well. I don't know if the Proxy-property will ever be implemented. I think iOS and MAC require a proxy to be defined elsewhere.

src/Sentry/Internal/Http/DefaultSentryHttpClientFactory.cs
```diff
@@ -20,25 +20,77 @@ public HttpClient Create(SentryOptions options)
...
+                        options.LogInfo("Using Proxy: {0}", options.HttpProxy);
+                    }
+                    catch (PlatformNotSupportedException)
+                    {
+                        supportsProxy = false;
+                    }
+                }
+                if (!supportsProxy)
+                {
+                    options.LogDebug("No proxy supported by NSUrlSessionHandler.");
```
bruno-garcia, line 69: User was explicit in setting a proxy, lets warn them it didn't work:

```suggestion
                    options.LogWarning("No proxy supported by NSUrlSessionHandler.");
```

src/Sentry/SentryOptions.cs
```diff
@@ -542,8 +542,19 @@ public int MaxCacheItems
 
     /// <summary>
     /// Creates the inner most <see cref="HttpClientHandler"/>.
+    /// Deprecated in favor of <see cref="CreateHttpMessageHandler"/>.
     /// </summary>
-    public Func<HttpClientHandler>? CreateHttpClientHandler { get; set; }
+    [Obsolete("Use CreateHttpMessageHandler instead")]
+    public Func<HttpClientHandler>? CreateHttpClientHandler
+    {
+        get => CreateHttpMessageHandler as Func<HttpClientHandler>;
```
bruno-garcia, line 550: Does this typecast work? If not, it'll just return `null`

src/Sentry/Internal/Http/DefaultSentryHttpClientFactory.cs
```diff
@@ -20,25 +20,77 @@ public HttpClient Create(SentryOptions options)
...
+                    compressionSupported = true;
+                }
+            }
+            catch (PlatformNotSupportedException)
+            {
+                compressionSupported = false;
+            }
+            if (!compressionSupported)
+            {
+                options.LogDebug("No response compression supported by NSUrlSessionHandler.");
```
bruno-garcia, line 90: Ideally we'd be aware of the intention of the user, as in, did they set a non-default value to `DecompressionMethods`? If so we might want to `Warn` that things didn't go as they expected instead. But at least LogInfo is a better choice then Debug here:

```suggestion
                options.LogInfo("No response compression supported by NSUrlSessionHandler.");
```

bruno-garcia (approved):
Left some minor suggestions but generally lgtm

src/Sentry/SentryOptions.cs
AlexSchuetz: Unfortunately you are right. This doesn't work. I introduced a private field to hold the `Func<HttpClientHandler>?`. So when the setter is used, the getter will work. if the new `CreateHttpMessageHandler` is used instead `CreateHttpClientHandler`will return null.

AlexSchuetz:
@bruno-garcia I followed your suggestions. Am I supposed to resolve those conversations now?

bitsandfoxes:
> Am I supposed to resolve those conversations now?

Yes. Everything that has been addressed can be marked as resolved.

src/Sentry/SentryOptions.cs
```diff
@@ -540,10 +540,30 @@ public int MaxCacheItems
...
+
     /// <summary>
     /// Creates the inner most <see cref="HttpClientHandler"/>.
+    /// Deprecated in favor of <see cref="CreateHttpMessageHandler"/>.
+    /// </summary>
+    [Obsolete("Use CreateHttpMessageHandler instead")]
+    public Func<HttpClientHandler>? CreateHttpClientHandler
+    {
+        get => _createClientHandler;
+        set =>
```
bitsandfoxes, line 556:
```suggestion
        set
```

CHANGELOG.md
```diff
@@ -11,6 +11,7 @@
 - Fixed baggage propagation when an exception is thrown from middleware ([#2487](https://github.com/getsentry/sentry-dotnet/pull/2487))
 - Fix Durable Functions preventing orchestrators from completing ([#2491](https://github.com/getsentry/sentry-dotnet/pull/2491))
 - Re-enable HubTests.FlushOnDispose_SendsEnvelope ([#2492](https://github.com/getsentry/sentry-dotnet/pull/2492))
+- Fixes #2502 SentryOptions should use 'HttpMessageHandler' instead of 'HttpClientHandler' to support NSUrlSessionHandler on iOS  ([#2503](https://github.com/getsentry/sentry-dotnet/pull/2503))
```
bitsandfoxes, line 14:
```suggestion
- Introduced `HttpMessageHandler` in favor of the now deprecated `HttpClientHandler` on the options. This allows the SDK to support NSUrlSessionHandler on iOS ([#2503](https://github.com/getsentry/sentry-dotnet/pull/2503))
```

bitsandfoxes (approved):
This looks good to me! Thanks again for contributing!

AlexSchuetz:
@bitsandfoxes @bruno-garcia  You are welcome. Thank's for your support and review.

PR #2505, opened 2023-07-22
Add MinimumEventLevel to Sentry.Log4Net and convert events below it to breadcrumbs

IhateTrains:
Both NLog and Serilog integrations already feature `MinimumEventLevel`. I attempted to also add it for log4net, hope its implementation is acceptable to you.

If `MinimumEventLevel` is specified, every log4net LoggingEvent above `threshold` and below `MinimumEventLevel` will be converted to a breadcrumb.

Heavily inspired by this solution from SO:
https://stackoverflow.com/a/62923187/10249243

### DIFF
--- a/src/Sentry.Log4Net/LevelMapping.cs
+++ b/src/Sentry.Log4Net/LevelMapping.cs
@@ -16,7 +16,7 @@ internal static class LevelMapping
         };
     }
 
-    public static BreadcrumbLevel? ToBreadcrumbLevel(this LoggingEvent loggingLevel)
+    public static BreadcrumbLevel ToBreadcrumbLevel(this LoggingEvent loggingLevel)
     {
         return loggingLevel.Level switch
         {
@@ -26,7 +26,7 @@ internal static class LevelMapping
             var l when l == Level.Notice || l == Level.Info => BreadcrumbLevel.Info,
             var l when l == Level.Debug || l == Level.Verbose || l == Level.Trace || l == Level.Finer || l == Level.Finest ||
                        l == Level.Fine => BreadcrumbLevel.Debug,
-            _ => null
+            _ => BreadcrumbLevel.Info
         };
     }
 }
--- a/src/Sentry.Log4Net/LevelMapping.cs
+++ b/src/Sentry.Log4Net/LevelMapping.cs
@@ -16,7 +16,7 @@ internal static class LevelMapping
         };
     }
 
-    public static BreadcrumbLevel ToBreadcrumbLevel(this LoggingEvent loggingLevel)
+    public static BreadcrumbLevel? ToBreadcrumbLevel(this LoggingEvent loggingLevel)
     {
         return loggingLevel.Level switch
         {
@@ -26,7 +26,7 @@ internal static class LevelMapping
             var l when l == Level.Notice || l == Level.Info => BreadcrumbLevel.Info,
             var l when l == Level.Debug || l == Level.Verbose || l == Level.Trace || l == Level.Finer || l == Level.Finest ||
                        l == Level.Fine => BreadcrumbLevel.Debug,
-            _ => BreadcrumbLevel.Info
+            _ => null
         };
     }
 }
--- a/src/Sentry.Log4Net/LevelMapping.cs
+++ b/src/Sentry.Log4Net/LevelMapping.cs
@@ -15,4 +15,18 @@ internal static class LevelMapping
             _ => null
         };
     }
+
+    public static BreadcrumbLevel? ToBreadcrumbLevel(this LoggingEvent loggingLevel)
+    {
+        return loggingLevel.Level switch
+        {
+            var l when l == Level.Fatal || l == Level.Emergency || l == Level.All => BreadcrumbLevel.Critical,
+            var l when l == Level.Alert || l == Level.Critical || l == Level.Severe || l == Level.Error => BreadcrumbLevel.Error,
+            var l when l == Level.Warn => BreadcrumbLevel.Warning,
+            var l when l == Level.Notice || l == Level.Info => BreadcrumbLevel.Info,
+            var l when l == Level.Debug || l == Level.Verbose || l == Level.Trace || l == Level.Finer || l == Level.Finest ||
+                       l == Level.Fine => BreadcrumbLevel.Debug,
+            _ => null
+        };
+    }
 }
--- a/src/Sentry.Log4Net/LevelMapping.cs
+++ b/src/Sentry.Log4Net/LevelMapping.cs
@@ -20,12 +20,12 @@ internal static class LevelMapping
     {
         return loggingLevel.Level switch
         {
-            var l when l == Level.Fatal || l == Level.Emergency || l == Level.All => BreadcrumbLevel.Critical,
+            var l when l == Level.Fatal || l == Level.Emergency => BreadcrumbLevel.Critical,
             var l when l == Level.Alert || l == Level.Critical || l == Level.Severe || l == Level.Error => BreadcrumbLevel.Error,
             var l when l == Level.Warn => BreadcrumbLevel.Warning,
             var l when l == Level.Notice || l == Level.Info => BreadcrumbLevel.Info,
             var l when l == Level.Debug || l == Level.Verbose || l == Level.Trace || l == Level.Finer || l == Level.Finest ||
-                       l == Level.Fine => BreadcrumbLevel.Debug,
+                       l == Level.Fine || l == Level.All => BreadcrumbLevel.Debug,
             _ => null
         };
     }
--- a/src/Sentry.Log4Net/LevelMapping.cs
+++ b/src/Sentry.Log4Net/LevelMapping.cs
@@ -16,17 +16,13 @@ internal static class LevelMapping
         };
     }
 
-    public static BreadcrumbLevel? ToBreadcrumbLevel(this LoggingEvent loggingLevel)
+    public static BreadcrumbLevel? ToBreadcrumbLevel(this LoggingEvent loggingLevel) => loggingLevel.Level switch
     {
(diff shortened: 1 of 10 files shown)


src/Sentry.Log4Net/LevelMapping.cs
```diff
@@ -15,4 +15,18 @@ internal static class LevelMapping
             _ => null
         };
     }
+
+    public static BreadcrumbLevel? ToBreadcrumbLevel(this LoggingEvent loggingLevel)
+    {
+        return loggingLevel.Level switch
+        {
+            var l when l == Level.Fatal || l == Level.Emergency || l == Level.All => BreadcrumbLevel.Critical,
```
jamescrosswell, line 23: Having `Level.All` map to `BreadcrumbLevel.Critical` seems a bit odd. [According to the NLog Docs](https://logging.apache.org/log4net/release/sdk/html/F_log4net_Core_Level_All.htm):
>The All level designates the lowest level possible.

Would  `null`  be more suitable?

I see the code just above that `All` is being mapped to `SentryLevel.Fatal`... I imagine you tried to respect the same mappings (thanks). This is maybe more of a question for the other code owners then (@SeanFeldman @bitsandfoxes ).

test/Sentry.Log4Net.Tests/SentryAppenderTests.cs
```diff
@@ -125,6 +130,19 @@ public void Append_NullEvent_NoOp()
...
+    public void Append_BelowThreshold_NoOp()
+    {
+        var sut = _fixture.GetSut();
+        sut.Threshold = Level.Warn;
+        var evt = new LoggingEvent(new LoggingEventData
+        {
+            Level = Level.Info
+        });
+
+        sut.DoAppend(evt);
```
jamescrosswell, line 143: I'm not too sure what this test is checking. The only thing it really seems to test is whether an exception is thrown by `DoAppend`, in which case maybe we should rename it `Append_BelowThreshold_DoesNotThrow`.

test/Sentry.Log4Net.Tests/SentryAppenderTests.cs
```diff
@@ -291,6 +309,32 @@ public void Append_ConfiguredEnvironment()
...
 
+    [Fact]
+    public void MinimumEventLevel_DefaultsToNull()
+    {
+        var appender = new SentryAppender();
+        Assert.Null(appender.MinimumEventLevel);
+    }
+
+    [Fact]
+    public void MinimumEventLevel_ConvertsEventsBelowToBreadcrumbs()
```
jamescrosswell, line 320: For [naming unit tests](https://github.com/getsentry/sentry-dotnet/blob/main/CONTRIBUTING.md#naming-tests), we try to use the convention `Method_Context_Expectation`. For this one then, maybe `DoAppend_BelowMinimumEventLevel_AddsBreadcrumb`.

jamescrosswell, line 320: It'd be great to add two more unit tests to be explicit about what we expect:
1.  When `MinimumEventLevel is null`
2. When the Level is above `MinimumEventLevel` (make sure an event gets created in that case)

src/Sentry.Log4Net/SentryAppender.cs
```diff
@@ -79,6 +85,27 @@ protected override void Append(LoggingEvent loggingEvent)
...
+
+            var level = loggingEvent.ToBreadcrumbLevel();
+            if (level is null)
+            {
+                _hub.AddBreadcrumb(message, category, type: null, data);
+            }
+            else
+            {
+                _hub.AddBreadcrumb(message, category, type: null, data, level.Value);
+            }
```
jamescrosswell, line 105: I think this if/else block can be simplified to:
```csharp
_hub.AddBreadcrumb(message, category, type: null, data, level ?? default);
```

jamescrosswell (changes requested):
@IhateTrains thank you so much for the contribution. 

I added a few comments but generally looking pretty good 👍🏻

src/Sentry.Log4Net/LevelMapping.cs
IhateTrains: Indeed, I tried to make it consistent with `ToSentryLevel`. But I agree, mapping All to Fatal is rather weird.

> Would null be more suitable?

I think it should map to Debug. Typically such a low level should be below the appender's threshold, but if a user actually wants it, they should be able to use it for breadcrumbs. Does that make sense?

src/Sentry.Log4Net/SentryAppender.cs
IhateTrains: Thanks! I think I've never used `default` like this before. You learn something new every day.

test/Sentry.Log4Net.Tests/SentryAppenderTests.cs
IhateTrains: I wasn't sure how to write a test that verifies that no event has been sent. But now after searching through other tests in the solution I changed the assertion to:
```cs
_fixture.Hub.DidNotReceiveWithAnyArgs().CaptureEvent(Arg.Any<SentryEvent>());
```
Also renamed the test to `Append_BelowThreshold_DoesNotSendEvent`. Is it OK now?

test/Sentry.Log4Net.Tests/SentryAppenderTests.cs
IhateTrains: Renamed ✅

IhateTrains: Added `DoAppend_NullMinimumEventLevel_AddsEvent` and `DoAppend_AboveMinimumEventLevel_AddsEvent`, as in both suggested cases an event is sent.

src/Sentry.Log4Net/LevelMapping.cs
jamescrosswell: Hm. If `null` is returned then the [`HubExtensions.AddBreadcrumb` extension](https://github.com/getsentry/sentry-dotnet/blob/470567c4882b2c13ce265ee1daa68256dc58deed/src/Sentry/HubExtensions.cs#L63) would be called without any `BreadcrumbLevel` specified and would use the default `BreadcrumbLevel` (which it turns out is [Info](https://github.com/getsentry/sentry-dotnet/blob/470567c4882b2c13ce265ee1daa68256dc58deed/src/Sentry/BreadcrumbLevel.cs#L18))

Arguably the `default` value is a good pick when no specific value has been specified. But I don't really have a strong - other than that `Critical` probably isn't the right level. I'm OK with your code returning either null (implicitly Info) or Debug.

jamescrosswell: btw: I also raised https://github.com/getsentry/sentry-dotnet/issues/2521, which we can address in a separate commit.

test/Sentry.Log4Net.Tests/SentryAppenderTests.cs
jamescrosswell: Ah, that makes heaps more sense 😄 The docs for NSubstitute are great by the way - it's a super powerful mocking framework.

test/Sentry.Log4Net.Tests/SentryAppenderTests.cs
jamescrosswell: Perfect!

jamescrosswell:
This is looking great @IhateTrains ! Just the small tweak to be made to LevelMappings and then getting the build checks passing. Let me know if you need any guidance/help working out the problems with the builds. 

ℹ️  One tip in that respect is search for the text `Failed!`, since that's a marker for any failed unit tests...

src/Sentry.Log4Net/LevelMapping.cs
IhateTrains: Mapped to BreadcrumbLevel.Debug ✅

IhateTrains:
@jamescrosswell turns out the `Hub.IsEnabled.Returns(true);` I copied from another test class' fixture was a problem; after removing it all checks are passing.

src/Sentry.Log4Net/LevelMapping.cs
```diff
@@ -15,4 +15,18 @@ internal static class LevelMapping
...
+        {
+            var l when l == Level.Fatal || l == Level.Emergency => BreadcrumbLevel.Critical,
+            var l when l == Level.Alert || l == Level.Critical || l == Level.Severe || l == Level.Error => BreadcrumbLevel.Error,
+            var l when l == Level.Warn => BreadcrumbLevel.Warning,
+            var l when l == Level.Notice || l == Level.Info => BreadcrumbLevel.Info,
+            var l when l == Level.Debug || l == Level.Verbose || l == Level.Trace || l == Level.Finer || l == Level.Finest ||
+                       l == Level.Fine || l == Level.All => BreadcrumbLevel.Debug,
+            _ => null
+        };
+    }
```
SeanFeldman, line 31: I think this could be simplified with the following:

```suggestion
    public static BreadcrumbLevel? ToBreadcrumbLevel(this LoggingEvent loggingLevel) => loggingLevel.Level switch
    {
        Level.Fatal or Level.Emergency => BreadcrumbLevel.Critical,
        Level.Alert or Level.Critical or Level.Severe or Level.Error => BreadcrumbLevel.Error,
        Level.Warn => BreadcrumbLevel.Warning,
        Level.Notice or Level.Info => BreadcrumbLevel.Info,
        Level.Debug or Level.Verbose or Level.Trace or Level.Finer or Level.Finest or Level.Fine or Level.All => BreadcrumbLevel.Debug,
        _ => null
    };
```

jamescrosswell (approved):
Nice work @IhateTrains ! Thanks heaps for the contribution 👍🏻

PR #2511, opened 2023-07-25
Fixes Transaction name not reporting correctly when using UseExceptionHandler

### LINKED ISSUE #2433: Transaction name not reporting correctly when using UseExceptionHandler(string errorHandlingPath)
### Package

Sentry.AspNetCore

### .NET Flavor

.NET

### .NET Version

7.0.5

### OS

macOS

### SDK Version

3.33.1

_No response_

### Steps to Reproduce

1. With an ASP.NET Core MVC project, configure the pipeline with `UseExceptionHandler("/errors")` somewhere before `MapDefaultControllerRoute`.
2. Create the controller `ErrorsController` with an action `Index` that returns a view.
3. In some other controller action, throw an exception.

I have a minimal project to reproduce here: https://github.com/goelze/sentry-error-example

### Expected Result

If the exception was thrown in the Index action on the HomeController, the expected transaction name reported in Sentry should be `GET /Home/Index/{id?}`

### Actual Result

The actual transaction name is `GET /Errors/Index/{id?}`
sentry error screenshot

jamescrosswell:
Fixes #2433

### Problem

The `UseExceptionHandler` extension in ASP.NET (and Core) configures a route that all uncaught exceptions get redirected to. It does that by catching uncaught exceptions (similar to how Sentry does this I imagine) and then redirecting the request along with some context about the exception to the error route that users have configured.

Since the exception never gets re-thrown, Sentry's regular exception handling mechanism doesn't pick this up. We detect it after the request has been redirected to the error route by checking to see if ASP.NET's global exception handler left any evidence of an exception in the call to `context.Features.Get<IExceptionHandlerFeature?>()`.

At that point in time, the route values on the `HttpContext` have already been changed from something like `MyController/Throw` to `ErrorController/Index` or whatever the original and error routes are. 

### Solution 

On .NET 6 and later, we have almost enough information in `IExceptionHandlerFeature` to reconstruct the original request route. The only bit we're missing really is the HTTP Method, which we capture/store on the way into the middleware before the request gets handled and the exception gets caught by ASP.NET's global exception handler. 

_Note:_ Versions of `IExceptionHandlerFeature` prior to .NET 6 shipped with much more limited information and we don't yet have a solution for those targets. 

Where the [route data gets applied is in `SentryClient.DoSendEvent`](https://github.com/getsentry/sentry-dotnet/blob/13faf4d2359b7cb8e094a0a974873d7554f04fb2/src/Sentry/SentryClient.cs#L227). At that point in time we don't have access to any of this information (there's no `HttpContext` to read the `IExceptionHandlerFeature` off, for example). 

[The way `Scope.Apply` is implemented](https://github.com/getsentry/sentry-dotnet/blob/13faf4d2359b7cb8e094a0a974873d7554f04fb2/src/Sentry/Scope.cs#L415C1-L431C51) though, it never overwrites values that have been explicitly provided with  the `SentryEvent` with differing values from the `Scope`... so we just store the route information on the Event in our SentryMiddleware before calling `hub.CaptureEvent(evt)`... that way the correct route information makes it's way through to Sentry. 

### Implementation notes

When the `UseExceptionHandler` middleware is used, we need to read route information from `IExceptionHandlerFeature` rather than `HttpContext`. I refactored a bunch of logic that we had in `HttpContextExtensions` out into a new `RouteUtils` class, where it is also leveraged a new `ExceptionHandlerFeatureExtensions` class.


src/Sentry.AspNetCore/SentryMiddleware.cs
```diff
@@ -151,14 +180,15 @@ public async Task InvokeAsync(HttpContext context, RequestDelegate next)
...
+            // preventing OnCompleted flush from working.
             Task FlushBeforeCompleted() => hub.FlushAsync(_options.FlushTimeout);
 
-            void CaptureException(Exception e, SentryId evtId, string mechanism, string description)
+            void CaptureException(Exception e, SentryId evtId, string mechanism, string description, ExceptionHandlerFeatureDetails? exceptionHandlerDetails = null)
             {
                 e.SetSentryMechanism(mechanism, description, handled: false);
 
                 var evt = new SentryEvent(e, eventId: evtId);
+                exceptionHandlerDetails?.ApplyOriginalRouteValues(evt);
```
bitsandfoxes, line 191: If I read that correctly, then [`context.GetRouteData()`](https://github.com/getsentry/sentry-dotnet/blob/0714fd20df8991dfc47ab066a63dbfd495dc009b/src/Sentry.AspNetCore/ScopeExtensions.cs#L73) is returning the wrong route when we first populate the scope? Same for the transaction name. Is there a way to correct it there already instead of overwriting the tags on the event?

jamescrosswell: Kind of. The `UseExceptionHandler` extension in ASP.NET (and Core) configures a route that all uncaught exceptions get redirected to. It does that by catching uncaught exceptions (similar to how Sentry does this I imagine) and then redirecting the request along with some context about the exception to the error route that users have configured.

Since the exception never gets re-thrown, Sentry's regular exception handling mechanism doesn't pick this up. We detect it after the fact (after the redirect to the error route). So if the original route is `/throw` and that route throws an exception, we don't find out immediately. We work it out by checking to see if ASP.NET's global exception handler left any evidence of an exception in the call to `context.Features.Get<IExceptionHandlerFeature?>()`.

At that point in time, the route values on the `HttpContext` have already been changed from something like `MyController/Throw` to 'ErrorController/Index` or whatever the original and error routes are. We do have almost enough information in `IExceptionHandlerFeature` to reconstruct the original request route though (at least on .NET 6 and later we do... earlier versions of that interface shipped with much more limited information). The only bit we're missing really is the HTTP Method (which you notice I'm capturing/storing on the way into the middleware pipeline before the request gets handled and the exception gets caught by ASP.NET's global exception handler). 

Where the [route data gets applied is in `SentryClient.DoSendEvent`](https://github.com/getsentry/sentry-dotnet/blob/13faf4d2359b7cb8e094a0a974873d7554f04fb2/src/Sentry/SentryClient.cs#L227). At that point in time we don't have access to any of this information (there's no `HttpContext` to read the `IExceptionHandlerFeature` off, for example). 

[The way `Scope.Apply` is implemented](https://github.com/getsentry/sentry-dotnet/blob/13faf4d2359b7cb8e094a0a974873d7554f04fb2/src/Sentry/Scope.cs#L415C1-L431C51) though, it never overwrites values that have been explicitly provided with  the `SentryEvent` with differing values from the `Scope`... so we just store these in the Event early.

I should probably put all of that context in the summary for the PR actually 😜

jamescrosswell: OK, see [summary above](https://github.com/getsentry/sentry-dotnet/pull/2511#issue-1820062274) - it's probably better written.

jamescrosswell: @bitsandfoxes an alternative solution would be to call `hub.CaptureEvent(evt, scope => applyOriginalRoutealues(scope))` instead of `hub.CaptureEvent(evt)` here in the middleware... that way when `DoSendEvent` calls `scope.Apply(@event)` it would be applying the correct values. Are there pros/cons to either of these approaches?

bitsandfoxes: I'd prefer the correction of the scope before the capture over the correction of the event during the capture.
Right before we capture the event here we already know that there is another mechanism at play. https://github.com/getsentry/sentry-dotnet/blob/30e93cbd848f66c0f4d089c45bdbb95d58f0f1c2/src/Sentry.AspNetCore/SentryMiddleware.cs#L129
Can we simply correct what we know is wrong (the transaction name in that case) via configureScope and let the SDK do the rest?

jamescrosswell: @bitsandfoxes I had a crack at implementing your suggestion (after modifying the `ExceptionHandlerFeatureExtensions` to operate on a `Scope` rather than a `SentryEvent`):
```csharp
// When exceptions get caught by the UseExceptionHandler feature we reset the TransactionName and
// Tags to reflect the original route (not the global error handling route)
_logger.LogTrace("Applying original route values for ExceptionHandlerFeature");
hub.ConfigureScope(scope =>
{
    exceptionFeature.ApplyTransactionName(scope, originalMethod);
    exceptionFeature.ApplyRouteTags(scope);
});
CaptureException(exceptionFeature.Error, eventId, "IExceptionHandlerFeature", description);
```

I thought it would be that simple but the integration tests were failing and it turns out [`Scope.Evaluate` is called immediately before `Scope.Apply` in the `SentryClient.DoSendEvent`](https://github.com/getsentry/sentry-dotnet/blob/30e93cbd848f66c0f4d089c45bdbb95d58f0f1c2/src/Sentry/SentryClient.cs#L226). Under the hood, `Evaluate` invokes the `OnEvaluating` callback (which is defined on the middleware) and this in turn calls `Scope.Populate`, which overwrites/populates the route values stored in `Scope.Tags` from the `HttpContext` (which, at this point, holds the route values of the global exception handler).

It seems that the order of precedence then is:
1. Event values (these trump everything)
2. Values from the HttpContext (these overwrite whatever is sitting on the Scope)
3. Scope values (used only if these don't conflict with existing values from either of the above)

That leaves two possible solutions:
1. We either leave the fix as is... setting the values in the Event before it gets passed down the chain
2. Since `Scope.Populate` has access to the HttpContext, we could potentially check here again for the presence of the `IExceptionHandlerFeature` and, if it exists, read the route values from this instead of getting them from the `HttpContext`. The [current logic to read them from the context is here](https://github.com/getsentry/sentry-dotnet/blob/30e93cbd848f66c0f4d089c45bdbb95d58f0f1c2/src/Sentry.AspNetCore/ScopeExtensions.cs#L71-L104). 

About the only challenge I see with the second approach is working out what the HTTP Method was on the original request (since this isn't stored in the `IExceptionHandlerFeature` details). Otherwise, I don't have any strong preference for either solution (one solution splits the logic for handling `IExceptionHandlerFeature` across two classes... the other solution splits the logic for extracting the route data across two classes - so they both seem to have pros and cons). 

I think I've got a preference for the existing solution now that it's written though, purely because it's the least effort at this point 😜

jamescrosswell: Adding some more context from a discord chat with @bitsandfoxes 
> There are just a couple things that stick out to me.
> - processing of an event in an integration but outside of an event processor
> - the setting of tags.
>
>Especially the settings of tags were "a huge issue" has an issue for our ingestion/storage people. 
> So I'd like to keep all that in one place

I'll have a crack at deferring the logic to happen in an event processor instead then.

bitsandfoxes (approved):
This looks great!

PR #2513, opened 2023-07-26
fix(android): Fix internal Android SDK API changes break build

markushi:
Starting with our Sentry Java/Android SDK `6.26.0` `SentryEvent.serialize()` expects a parameter which implements the `ObjectWriter` interface.
`JsonObjectWriter` implements this `ObjectWriter` interface in Java, but somehow this is not reflected in the generated binding.

This workaround replaces the interface with the `JsonObjectWriter` type in the `SentryEvent.serialize()`. Definitely not an elegant solution, if you have something better in mind or know how fix the binding generation itself please let me know 😊

### DIFF
--- a/src/Sentry/Platforms/Android/Extensions/SentryEventExtensions.cs
+++ b/src/Sentry/Platforms/Android/Extensions/SentryEventExtensions.cs
@@ -21,7 +21,8 @@ internal static class SentryEventExtensions
         using var streamWriter = new JavaOutputStreamWriter(stream);
         using var jsonWriter = new JavaSdk.JsonObjectWriter(streamWriter, javaOptions.MaxDepth);
         sentryEvent.Serialize(jsonWriter, javaOptions.Logger);
-        jsonWriter.Flush();
+
+        streamWriter.Flush();
         stream.Seek(0, SeekOrigin.Begin);
 
         using var json = JsonDocument.Parse(stream);
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,9 +17,9 @@
 
 ### Dependencies
 
-- Bump Java SDK from v6.25.1 to v6.25.2 ([#2484](https://github.com/getsentry/sentry-dotnet/pull/2484))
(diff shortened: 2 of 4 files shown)


src/Sentry/Platforms/Android/Extensions/SentryEventExtensions.cs
```diff
@@ -21,7 +21,8 @@ public static SentryEvent ToSentryEvent(this JavaSdk.SentryEvent sentryEvent, Ja
         using var streamWriter = new JavaOutputStreamWriter(stream);
         using var jsonWriter = new JavaSdk.JsonObjectWriter(streamWriter, javaOptions.MaxDepth);
```
bitsandfoxes, line 22: So we no longer need the json writer?

src/Sentry/Platforms/Android/Extensions/SentryEventExtensions.cs
```diff
@@ -21,7 +21,8 @@ public static SentryEvent ToSentryEvent(this JavaSdk.SentryEvent sentryEvent, Ja
...
         sentryEvent.Serialize(jsonWriter, javaOptions.Logger);
-        jsonWriter.Flush();
+
+        streamWriter.Flush();
```
bruno-garcia, line 25: `jsonWriter.Flush();` doesn't exist anymore? I'm not sure if we ca skip flushing this and only flush the `streamWriter`

markushi: Yes, this should be safe as `JsonObjectWriter` and the underlying vendored JSON itself don't hold flush-able state. `sentry-java` in some places flushes the writer` as well (e.g. [see this one](https://github.com/getsentry/sentry-java/blob/87b3774676dde4ad1f5e031bc461e26b9921ad77/sentry/src/main/java/io/sentry/JsonSerializer.java#L222-L225))
But I kinda agree, this feels brittle and error-prone - let me re-add it in the upcoming `7.0.0` release of `sentry-java`. Ideally `JsonObjectWriter` should own the stream and take care of flushing itself.

bitsandfoxes (approved):

Issue #2514, opened 2023-07-26
Error status of OpenTelemetry Spans is not being correctly reflected in Sentry spans.


jamescrosswell:
See https://stackoverflow.com/questions/76742975/capturing-exceptions-in-opentelemetry-net-and-sentry-io

This can be reproduced by modifying the throw endpoint on the OpenTelemetry sample to look something like this:

```csharp
app.MapGet("/throw", () =>
    {
        using var activity = Telemetry.ActivitySource.StartActivity("Some sketchy business");
        try
        {
            throw new Exception("test with activity.RecordException");
        }
        catch (Exception e)
        {
            activity?.RecordException(e);
            activity?.SetStatus(Status.Error.WithDescription(e.Message));
        }
        return "Catch";
    }
);
```

At the very least, the span processors should identify the otel span as having `otel.status_code == ERROR`... and even if we [can't create a fully symbolicated error report from the otel exception](https://github.com/open-telemetry/opentelemetry-dotnet/issues/2439#issuecomment-1577314568) yet, we can get some exception details from otel.

jamescrosswell:
https://github.com/open-telemetry/opentelemetry-dotnet/discussions/4703

jamescrosswell:
![image](https://github.com/getsentry/sentry-dotnet/assets/728212/a420b364-0903-433a-8e40-d128b5bbc57e)

@bitsandfoxes is currently checking on this.

jamescrosswell:
Resurrecting this issue to see if we can get some resolution. The original behaviour described in the StackOverflow thread is still present. 

The .NET SDK is sending the span as:
```json
  {
    "span_id": "79ca9f101de64c64",
    "parent_span_id": "549459431681a0c1",
    "trace_id": "f5f184955e25c44a989428eb9613ae7f",
    "op": "Some sketchy business",
    "description": "Some sketchy business",
    "status": "unknown_error",
    "start_timestamp": "2025-07-14T21:53:13.321662+00:00",
    "timestamp": "2025-07-14T21:53:13.351699+00:00",
    "data": {
      "otel.status_description": "test with activity.RecordException",
      "otel.status_code": "ERROR",
      "otel.kind": "Internal"
    }
  }
```

So the status is quite clearly `unknown_error` and the data contains `"otel.status_code": "ERROR"`. 

However the UI in Sentry shows a status of `unknown`, which is a bit more ambiguous. 

[...]

PR #2515, opened 2023-07-27
Fixed Error status of OpenTelemetry Spans when using Activity.RecordException

jamescrosswell:
Fixes [#2514](https://github.com/getsentry/sentry-dotnet/issues/2514)

It's not perfect, as we don't have the full exception details available to us via the OpenTelemetry API... all we have is exception type, message and "stack trace"... although it's not really a stack trace - it's actually just `Exception.ToString()`. 

That might look something like this:
```csharp
Program+MyException: Attempted to divide by zero.
    ---> System.DivideByZeroException: Attempted to divide by zero.
    at Program.Main()
    --- End of inner exception stack trace ---
    at Program.Main()   
```

It could be more complicated in the case of an `AggregateException`. It might be possible to parse this into an exception graph that could be then be used in our `MainExceptionProcessor` to create one or more SentryExceptions. There would be no Debug info.

That all sounds like a lot of work though and it probably makes more sense to push to get the underlying [issue in the Otel codebase](https://github.com/open-telemetry/opentelemetry-dotnet/issues/2439#issuecomment-1577314568) resolved instead. 

For the time being, we're just putting the **_stack trace_** provided by OpenTelemetry in an `otel.stack_trace` attribute that gets sent with the event for the exception. So people can see the stack trace provided by OpenTelemetry if they want.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,7 @@
 - Fixed SDK not sending exceptions via Blazor WebAssembly due to a `PlatformNotSupportedException` ([#2506](https://github.com/getsentry/sentry-dotnet/pull/2506))
 - Align SDK with docs regarding session update for dropped events ([#2496](https://github.com/getsentry/sentry-dotnet/pull/2496))
 - Introduced `HttpMessageHandler` in favor of the now deprecated `HttpClientHandler` on the options. This allows the SDK to support NSUrlSessionHandler on iOS ([#2503](https://github.com/getsentry/sentry-dotnet/pull/2503))
-- Fixed Error status of OpenTelemetry Spans when using Activity.RecordException ([#2515](https://github.com/getsentry/sentry-dotnet/pull/2515))
+- Using `Activity.RecordException` now correctly updates the error status of OpenTelemetry Spans ([#2515](https://github.com/getsentry/sentry-dotnet/pull/2515))
 
 ### Dependencies
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@
 - Fixed SDK not sending exceptions via Blazor WebAssembly due to a `PlatformNotSupportedException` ([#2506](https://github.com/getsentry/sentry-dotnet/pull/2506))
 - Align SDK with docs regarding session update for dropped events ([#2496](https://github.com/getsentry/sentry-dotnet/pull/2496))
 - Introduced `HttpMessageHandler` in favor of the now deprecated `HttpClientHandler` on the options. This allows the SDK to support NSUrlSessionHandler on iOS ([#2503](https://github.com/getsentry/sentry-dotnet/pull/2503))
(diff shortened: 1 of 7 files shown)


src/Sentry.OpenTelemetry/SentrySpanProcessor.cs
```diff
@@ -286,36 +295,57 @@ private static SpanStatus GetErrorSpanStatus(IDictionary<string, object?> attrib
...
+            catch
+            {
+                _options?.DiagnosticLogger?.LogError($"Failed to create poor man's exception for type : {exceptionType}");
+                continue;
+            }
+
+            // TODO: Validate that our `DuplicateEventDetectionEventProcessor` prevents this from doubling exceptions
+            // that are also caught by other means, such as our AspNetCore middleware, etc.
+            // (When options.RecordException = true is set on AddAspNetCoreInstrumentation...)
+            // Also, in such cases - how will we get the otel scope and trace context on the other one?
```
bitsandfoxes, line 336: The PR looks good to me. The TODO here sounds like quite the gotcha.

bitsandfoxes (approved):

CHANGELOG.md
```diff
@@ -14,6 +14,7 @@
 - Fixed SDK not sending exceptions via Blazor WebAssembly due to a `PlatformNotSupportedException` ([#2506](https://github.com/getsentry/sentry-dotnet/pull/2506))
 - Align SDK with docs regarding session update for dropped events ([#2496](https://github.com/getsentry/sentry-dotnet/pull/2496))
 - Introduced `HttpMessageHandler` in favor of the now deprecated `HttpClientHandler` on the options. This allows the SDK to support NSUrlSessionHandler on iOS ([#2503](https://github.com/getsentry/sentry-dotnet/pull/2503))
+- Fixed Error status of OpenTelemetry Spans when using Activity.RecordException ([#2515](https://github.com/getsentry/sentry-dotnet/pull/2515))
```
bitsandfoxes, line 17:
```suggestion
- Using `Activity.RecordException` now correctly updates the error status of OpenTelemetry Spans ([#2515](https://github.com/getsentry/sentry-dotnet/pull/2515))
```

src/Sentry.OpenTelemetry/SentrySpanProcessor.cs
jamescrosswell: It does. It occurs if you do something like this:
```csharp
        using var activity = Telemetry.ActivitySource.StartActivity("Funny business");
        try
        {
            throw new Exception("test activity.RecordException");
        }
        catch (Exception e)
        {
            activity?.RecordException(e);
            activity?.SetStatus(Status.Error.WithDescription(e.Message));
        }
        return "Catch";
```

The deduplication doesn't work at all (since it matches by reference). Again, we could come up with convoluted ways of working around this but I think the ultimate solution is to address the issue in the OpenTelemetry codebase rather than try to work around it in ours. 

I'll create an issue for the above in our repository saying as much though.

jamescrosswell: https://github.com/getsentry/sentry-dotnet/issues/2519

PR #2526, opened 2023-08-01
WIP: GraphQL

jamescrosswell:
# Summary

GraphQL supports OpenTelemetry, so at the moment just exploring what we get out of the box vs what we need.

### Requirements

- [GraphQL Client Integrations](https://develop.sentry.dev/sdk/features/#graphql-client-integrations)
- [GraphQL Performance](https://develop.sentry.dev/sdk/features/#graphql-performance)
  - [Server](https://develop.sentry.dev/sdk/features/#server)
  - [Client](https://develop.sentry.dev/sdk/features/#client)
  - [Breadcrumbs](https://develop.sentry.dev/sdk/features/#breadcrumbs-1)
- [Example implementation for Appolo3](https://docs.sentry.io/platforms/android/configuration/integrations/apollo3/#graphql-client-errors)

### Reference Material
- [GraphQL .NET](https://github.com/graphql-dotnet/graphql-dotnet)
- [GraphQL Client](https://github.com/graphql-dotnet/graphql-client)

### Server 
[graphql-dotnet](https://github.com/graphql-dotnet/graphql-dotnet) can be used to create a GraphQL Server. This package supports OpenTelemetry, which we can leverage. 

We will need to to customise and enrich the spans that get created out of the box. Possibly we introduce the idea of "SpanEnrichers" or something in the `SentrySpanProcessor` .

### Client

[graphql-client](https://github.com/graphql-dotnet/graphql-client) doesn't include any OpenTelemetry instrumentation. 

We can instrument this by configuring a custom [GraphQLHttpClientOptions.HttpMessageHandler](https://github.com/graphql-dotnet/graphql-client/blob/344a3e9f2ff2d110d331d377c25611e0306ed042/src/GraphQL.Client/GraphQLHttpClientOptions.cs#L31).

### Client

To instrument for `graphql-client` we've created `SentryGraphQLHttpMessageHandler` and  `SentryGraphQLFailedRequestHandler`, which are very similar to `SentryHttpMessageHandler` and `SentryFailedRequestHandler`.

The `graphql-client` serialises the payload content using an `IGraphQLWebsocketJsonSerializer` that can be customised when initialising the graphql  client... which unfortunately means a dependency on `GraphQL.Client.Serializer.SystemTextJson` and that means we probably have to add a separate Sentry.GraphQL.Client package 😞 

We pass an `IGraphQLWebsocketJsonSerializer` into our custom GraphQL message handlers so that we can get at the deserialised payload and use this to fulfil our various requirements. 

#### Challenges:
- GraphQL messages can be passed over Websockets and we still have to work out what to do in that case. The above only works when using HTTP Post operations to send GraphQL requests.
- The inheritance means we're doing things to satisfy technical language requirements rather than to enable API capability as well, which isn't very nice.
  - **TODO:** Having a shared base class for `SentryGraphQLHttpMessageHandler` and `SentryHttpMessageHandler` turns out to be a bit messy. Maybe try pulling some of the common logic out into a helper class that's used by two handlers that don't share the same ancestor instead.


jamescrosswell:
Closing as this will be done in https://github.com/getsentry/sentry-dotnet/pull/2538 and https://github.com/getsentry/sentry-dotnet/pull/2572 instead.

PR #2528, opened 2023-08-03
ref: Internal `IsTracingEnabled` to `IsPerformanceMonitoringEnabled`

bitsandfoxes:
With tracing without performance in the SDK this should start separating intentions.
#skip-changelog


src/Sentry/SentryOptions.cs
```diff
@@ -719,7 +719,7 @@ public bool ReportAssemblies
     /// Indicates whether tracing is enabled, via any combination of
```
jamescrosswell, line 719: Don't forget to update the XML docs

jamescrosswell (approved):

PR #2538, opened 2023-08-10
GraphQL over HTTP client integration

jamescrosswell:
# Summary

To auto-instrument GraphQL, there are two separate packages we're looking at supporting (initially):
1. [GraphQL .NET](https://github.com/graphql-dotnet/graphql-dotnet) (for building GraphQL servers)
2. [GraphQL Client](https://github.com/graphql-dotnet/graphql-client) (for building GraphQL clients)

This pull requests enables an integration with GraphQL Client.

### Requirements

- [GraphQL Client Integrations](https://develop.sentry.dev/sdk/features/#graphql-client-integrations)
- [GraphQL Performance](https://develop.sentry.dev/sdk/features/#client)
- [Example implementation for Appolo3](https://docs.sentry.io/platforms/android/configuration/integrations/apollo3/#graphql-client-errors)

## Overview 

[graphql-client](https://github.com/graphql-dotnet/graphql-client) doesn't include any OpenTelemetry instrumentation so we've created a `SentryGraphQLHttpMessageHandler` and  `SentryGraphQLFailedRequestHandler` to instrument this.

`graphql-client` implements [GraphQL over HTTP](https://github.com/graphql/graphql-over-http) so we shouldn't need to take any dependencies on the `graphql-client` nuget packages... we've done everything using standard JSON serializers.

[See here](https://sentry-sdks.sentry.io/issues/4392798909/?project=5428537&query=is%3Aunresolved&referrer=issue-stream&statsPeriod=1h&stream_index=0) for an example of a trace/transaction with a GraphQL Error.

#### Limitations:
- GraphQL messages can be passed over Websockets and we still have to work out what to do in that case. This PR is limited to support for GraphQL over HTTP.

## Example

graphql-client-errors


CHANGELOG.md
```diff
@@ -1,12 +1,19 @@
...
+- Graphql client ([#2538](https://github.com/getsentry/sentry-dotnet/pull/2538))
+
 ## 3.35.0
 
 ### Features
 
 - Distributed tracing now works independently of the performance feature. This allows you to connect errors to other Sentry instrumented applications ([#2493](https://github.com/getsentry/sentry-dotnet/pull/2493))
 - Added Sampling Decision to Trace Envelope Header ([#2495](https://github.com/getsentry/sentry-dotnet/pull/2495))
 - Add MinimumEventLevel to Sentry.Log4Net and convert events below it to breadcrumbs ([#2505](https://github.com/getsentry/sentry-dotnet/pull/2505))
+- GraphQL Integration ([#2526](https://github.com/getsentry/sentry-dotnet/pull/2526))
```
bitsandfoxes, line 16:
```suggestion
```

jamescrosswell:
@bitsandfoxes the [expected features](https://develop.sentry.dev/sdk/features/#graphql-client-integrations) contain the following guidance:

image

That requires the addition of two properties:
- `string? Sentry.Request.ApiTarget`
- `object? Sentry.Response.Data`

Both `Request` and `Response` are `public sealed`... so I'm wondering if this is a breaking change. Reading through [how Microsoft deal with .NET API changes](https://learn.microsoft.com/en-gb/dotnet/core/compatibility/library-change-rules), curiously there's no mention of "adding a member to a public class". There are multiple references to adding abstract members, but nothing about vanilla members. 

I'll build the functionality in for the time being but would be good to understand if this is a binary compatible change.

test/Sentry.Tests/ApiApprovalTests.Run.Core3_1.verified.txt
```diff
@@ -1623,6 +1635,7 @@ namespace Sentry.Protocol
         public Response() { }
         public long? BodySize { get; set; }
         public string? Cookies { get; set; }
+        public object? Data { get; set; }
```
jamescrosswell, line 1638: @bitsandfoxes let me know if we can do this...

test/Sentry.Tests/ApiApprovalTests.Run.Core3_1.verified.txt
```diff
@@ -404,6 +404,7 @@ namespace Sentry
     public sealed class Request : Sentry.IJsonSerializable
     {
         public Request() { }
+        public string? ApiTarget { get; set; }
```
jamescrosswell, line 407: @bitsandfoxes let me know if we can do this...

jamescrosswell:
graphql-client-errors

src/Sentry/Internal/ObjectExtensions.cs
```diff
@@ -0,0 +1,41 @@
...
+{
+    private static ConditionalWeakTable<object, Dictionary<string, object?>> Map { get; } = new();
+
+    private static Dictionary<string, object?> AssociatedProperties(this object source) =>
+        Map.GetValue(source, _ => new Dictionary<string, object?>());
+
+    internal static void SetFused(this object source, string propertyName, object? value) =>
+        AssociatedProperties(source)[propertyName] = value;
+
+    internal static void SetFused<T>(this object source, T value)
```
bitsandfoxes, line 18:
```suggestion
    public static void SetFused<T>(this object source, T value)
```
The setter and getter are supposed to be used, right? And the `internal Fused` is for testing?

src/Sentry/Internal/ObjectExtensions.cs
```diff
@@ -0,0 +1,41 @@
...
+    {
+        var propertyName = typeof(T).Name;
+        source.AssociatedProperties()[propertyName] = value;
+    }
+
+    internal static T? GetFused<T>(this object source, string? propertyName = null)
```
bitsandfoxes, line 24:
```suggestion
    public static T? GetFused<T>(this object source, string? propertyName = null)
```

src/Sentry.GraphQl/SentryGraphQlHttpFailedRequestHandler.cs
```diff
@@ -0,0 +1,93 @@
...
+    internal SentryGraphQlHttpFailedRequestHandler(IHub hub, SentryOptions options, GraphQlContentExtractor? extractor = null)
+        : base(hub, options)
+    {
+        _hub = hub;
+        _options = options;
+        _extractor = extractor ?? new GraphQlContentExtractor(options);
+        _httpFailedRequestHandler = new SentryHttpFailedRequestHandler(hub, options);
+    }
+
+    private static readonly Regex ErrorsRegex = new ("(?i)\"errors\"\\s*:\\s*\\[", RegexOptions.Compiled);
```
bitsandfoxes, line 20: Should that get used somewhere?

src/Sentry.GraphQl/SentryGraphQlHttpFailedRequestHandler.cs
```diff
@@ -0,0 +1,93 @@
...
+
+internal class SentryGraphQlHttpFailedRequestHandler : SentryFailedRequestHandler
+{
+    private readonly IHub _hub;
+    private readonly SentryOptions _options;
+    private readonly GraphQlContentExtractor _extractor;
+    internal const string MechanismType = "SentryGraphQLHttpFailedRequestHandler";
+    private readonly SentryHttpFailedRequestHandler _httpFailedRequestHandler;
+
+    internal SentryGraphQlHttpFailedRequestHandler(IHub hub, SentryOptions options, GraphQlContentExtractor? extractor = null)
```
bitsandfoxes, line 11:
```suggestion
    public SentryGraphQlHttpFailedRequestHandler(IHub hub, SentryOptions options, GraphQlContentExtractor? extractor = null)
```
That's the constructor to be used, right?

src/Sentry.GraphQl/SentryGraphQlHttpFailedRequestHandler.cs
```diff
@@ -0,0 +1,93 @@
...
+
+            @event.Request = sentryRequest;
+            @event.Contexts[Response.Type] = responseContext;
+            if (requestContent is not null)
+            {
+                @event.Fingerprint = new[]
+                {
+                    requestContent.OperationNameOrFallback(),
+                    requestContent.OperationTypeOrFallback(),
+                    ((int)response.StatusCode).ToString()
```
bitsandfoxes, line 87:
```suggestion
                    response.StatusCode.ToString()
```
I don't think we need to cast the enum first?

src/Sentry/SentryHttpFailedRequestHandler.cs
```diff
@@ -0,0 +1,79 @@
+using Sentry.Internal;
+using Sentry.Protocol;
+
+namespace Sentry;
+
+internal class SentryHttpFailedRequestHandler : SentryFailedRequestHandler
+{
+    public const string MechanismType = "SentryHttpFailedRequestHandler";
+
+    internal SentryHttpFailedRequestHandler(IHub hub, SentryOptions options)
```
bitsandfoxes, line 10:
```suggestion
    public SentryHttpFailedRequestHandler(IHub hub, SentryOptions options)
```

src/Sentry.GraphQl/GraphQlContentExtractor.cs
```diff
@@ -0,0 +1,90 @@
...
+
+        Stream contentStream;
+        try
+        {
+            await content.LoadIntoBufferAsync().ConfigureAwait(false);
+            contentStream = await content.ReadAsStreamAsync().ConfigureAwait(false);
+        }
+        catch (Exception exception)
+        {
+            _options?.LogDebug($"Unable to read GraphQL content stream: {exception.Message}");
```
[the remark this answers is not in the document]
jamescrosswell: I think the GraphQL library itself will throw an exception if the content is unacceptable. Initially I had this code here before I'd added the call to:
```
await content.LoadIntoBufferAsync().ConfigureAwait(false);
```

You'd get an exception here if some other middleware was configured that wasn't being a good citizen and wasn't calling `LoadIntoBufferAsync` (which enables the buffer content to be read multiple times - if you don't call this then the content stream gets closed/disposed after whichever middleware calls `ReadToEndAsync` first and then all the other middlewares that want to read the content stream fall over.

Admittedly in that scenario the GraphQL client will fall over as well. For people who don't understand what's going on, it's probably best they don't think Sentry has broken their application though (instead, the GraphQL library will throw an exception whenever they're using that other middleware, which is actually the culprit).

src/Sentry.GraphQl/GraphQlContentExtractor.cs
```diff
@@ -0,0 +1,90 @@
...
+        }
+
+        var originalPosition = (contentStream.CanSeek) ? contentStream.Position : 0;
+        try
+        {
+            TrySeek(contentStream, 0);
+            // https://github.com/dotnet/corefx/blob/master/src/Common/src/CoreLib/System/IO/StreamReader.cs#L186
+            // Default parameters other than 'leaveOpen'
+            using var reader = new StreamReader(contentStream, Encoding.UTF8, true,
+                1024, leaveOpen: true);
```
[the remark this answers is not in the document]
jamescrosswell: I copied this from [`DefaultRequestPayloadExtractor`](https://github.com/getsentry/sentry-dotnet/blob/d4dda091cd4bc957d2b5f582fd3b3e83f9b157ab/src/Sentry/Extensibility/DefaultRequestPayloadExtractor.cs#L24).

We could maybe pull it out to a StreamPayloadExtractor helper that both classes use? That might be a good idea in fact, since I've already improved it by adding the call to `LoadIntoBufferAsync`.

src/Sentry.GraphQl/GraphQlContentExtractor.cs
```diff
@@ -0,0 +1,90 @@
...
+            return reader.ReadToEndAsync().GetAwaiter().GetResult();
+        }
+        catch (Exception exception)
+        {
+            _options?.LogDebug($"Unable to extract GraphQL content: {exception.Message}");
+            return null;
+        }
+        finally
+        {
+            TrySeek(contentStream, originalPosition);
```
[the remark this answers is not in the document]
jamescrosswell: Again, I copied/tweaked this from the `DefaultRequestPayloadExtractor`. My best guess is that by resetting the position to it's original position once we're done (most likely zero) we wouldn't break anything that **_assumes_** the position won't have been changed.

src/Sentry.GraphQl/SentryGraphQlHttpFailedRequestHandler.cs
jamescrosswell: Ah no... that was from the requirements but we're using the functionality on JsonElement to see if there is an `errors` property present (seemed cleaner to me). I'll get rid of the regex.

src/Sentry.GraphQl/SentryGraphQlHttpFailedRequestHandler.cs
jamescrosswell: It is, but the only place it would ever get used is in the internal constructor for [`SentryGraphQlHttpMessageHandler`](https://github.com/getsentry/sentry-dotnet/blob/ef28a1a82856f767ae667de25d02b60eac91210b/src/Sentry.GraphQl/SentryGraphQlHttpMessageHandler.cs#L34), which is basically just there for testings. 

We don't presently allow SDK users to pass in their own failed request handler to our message handlers.

One reason for keeping the failed request handlers separate from the message handlers themselves (since they might look, at first glance, tightly coupled) is that they are a bit more composable this way... and you can see I've actually used that composability [here](https://github.com/getsentry/sentry-dotnet/blob/ef28a1a82856f767ae667de25d02b60eac91210b/src/Sentry.GraphQl/SentryGraphQlHttpFailedRequestHandler.cs#L36). If the `GraphQLFailedRequestHandler` doesn't find any graphql specific errors in the response content, it passes the response for processing by the `SentryHttpFailedRequestHandler` to check if we received a status code that would indicate some other kind of failure (e.g. a 500 internal server error).

src/Sentry.GraphQl/SentryGraphQlHttpFailedRequestHandler.cs
```diff
@@ -0,0 +1,93 @@
...
+
+    protected internal override void DoEnsureSuccessfulResponse([NotNull]HttpRequestMessage request, [NotNull]HttpResponseMessage response)
+    {
+        JsonElement? json = null;
+        try
+        {
+            json = _extractor.ExtractResponseContentAsync(response).Result;
+            if (json is { } jsonElement)
+            {
+                if (jsonElement.TryGetProperty("errors", out var errorsElement))
```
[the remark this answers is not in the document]
jamescrosswell: That gets returned by the GraphQL Server... it's part of the GraphQL protocol. You can see it in the Response.Data context in the screenshot above and the server uses it to communicate some problems with the graphql. Although it was syntactically valid GraphQL (I think the client could pick up on most errors like that before sending anything through to the server) the field `test` was not a valid field for a Notes query so the server didn't know how to serve it up.

Unlike REST, GraphQL wasn't built specifically to sit on top of HTTP so it doesn't rely on HTTP Status codes to communicate error states... it puts that stuff in the payload instead.

One question I had actually is whether we wanted to use the GraphQL Error message as the description for the event in the case of an error, since this is probably more informative than showing 'Sentry.GraphQL.SentryGraphQLHttpFailedRequestException`.

And that triggers another question, which is what we should do if there are multiple errors communicated by the GraphQL Server in the payload? Currently we capture a single exception... which I think makes sense (one request, one stack trace, one exception). What would we use as the description of the error if there were multiple exceptions though? We could generate some kind of Aggregate exception in this case (`SentryGraphQLHttpFailedRequestException` is our own fabricated exception in any case). 

And that triggers yet another question: I noticed `SentryGraphQLHttpFailedRequestException` is the only custom exception in the whole solution... which makes me nervous. Do we have a reason for avoiding our own exception types? Is there some other way preferred way to handle this?

src/Sentry/SentryHttpFailedRequestHandler.cs
jamescrosswell: That one [was/is internal in our main branch as well](https://github.com/getsentry/sentry-dotnet/blob/d4dda091cd4bc957d2b5f582fd3b3e83f9b157ab/src/Sentry/SentryFailedRequestHandler.cs#L12).

In `SentryOptions` people can configure `CaptureFailedRequests`, `FailedRequestStatusCodes` and `FailedRequestTargets` but we don't (yet) provide a mechanism for people to bring their own implementation of `ISentryFailedRequestHandler`. I think if we do that, it should be part of a separate PR... but I'm not sure if we want to or not.

src/Sentry.GraphQl/GraphQlContentExtractor.cs
```diff
@@ -0,0 +1,90 @@
+namespace Sentry.GraphQl;
```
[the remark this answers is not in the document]
jamescrosswell: Done.

src/Sentry.GraphQl/GraphQlContentExtractor.cs
```diff
@@ -0,0 +1,90 @@
...
+
+/// <summary>
+/// Helper class to extract the content from GraphQL requests and responses
+/// </summary>
+internal class GraphQlContentExtractor
+{
+    private readonly SentryOptions? _options;
```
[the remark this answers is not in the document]
jamescrosswell: Thanks, that looks heaps better 👍🏻

samples/Sentry.Samples.GraphQL.Server/Program.cs
```diff
@@ -0,0 +1,119 @@
...
+
+        builder.Services.AddOpenTelemetry()
+            .WithTracing(tracerProviderBuilder =>
+                tracerProviderBuilder
+                    .AddSource(Telemetry.ActivitySource.Name)
+                    .ConfigureResource(resource => resource.AddService(Telemetry.ServiceName))
+                    .AddAspNetCoreInstrumentation()
+                    .AddHttpClientInstrumentation()
+                    .AddSentry()
+                ); // <-- Configure OpenTelemetry to send traces to Sentry
```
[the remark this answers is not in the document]
jamescrosswell: Nice catch!

samples/Sentry.Samples.GraphQL/Sentry.Samples.GraphQL.csproj
```diff
@@ -0,0 +1,17 @@
...
+    </PropertyGroup>
+
+    <ItemGroup>
+      <PackageReference Include="GraphQL" Version="7.5.0" />
+      <PackageReference Include="GraphQL.DataLoader" Version="7.5.0" />
+      <PackageReference Include="GraphQL.MemoryCache" Version="7.5.0" />
+      <PackageReference Include="GraphQL.SystemTextJson" Version="7.5.0" />
+    </ItemGroup>
+
+</Project>
```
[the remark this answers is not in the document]
jamescrosswell: It is not... I'll delete it 👍🏻

src/Sentry.GraphQl/SentryGraphQlHttpMessageHandler.cs
```diff
@@ -0,0 +1,105 @@
...
+        var breadcrumbData = new Dictionary<string, string>
+        {
+            {"url", url},
+            {"method", method},
+            {"status_code", ((int) response.StatusCode).ToString()}
+        };
+        AddIfExists(breadcrumbData, "request_body_size", response.RequestMessage?.Content?.Headers.ContentLength?.ToString());
+        AddIfExists(breadcrumbData, "response_body_size", response.Content.Headers.ContentLength?.ToString());
+        AddIfExists(breadcrumbData, "operation_name", graphqlInfo?.OperationName); // The GraphQL operation name
+        AddIfExists(breadcrumbData, "operation_type", graphqlInfo?.OperationType); // i.e. `query`, `mutation`, `subscription`
```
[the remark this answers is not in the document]
jamescrosswell: Can do 👍🏻

src/Sentry/Internal/ObjectExtensions.cs
jamescrosswell: We could get rid of the `Fused` extension... I'd copied that over [from here](https://github.com/mentaldesk/fuse) (I was looking for an excuse to build/publish a nuget package from scratch, so that I could better learn all the moving parts of Sentry's CI process).

src/Sentry.GraphQl/GraphQlContentExtractor.cs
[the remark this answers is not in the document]
jamescrosswell: Ah, in fact that doesn't work. `LoadIntoBufferAsync` only exists on the `HttpContent` type (and the `DefaultRequestPayloadExtractor` is operating on the `Request.Body`, which is of type `Stream?`). 

It might be best to leave them separate as `GraphQlContentExtractor` can then return an async Task (and hopefully save some memory rather than read the whole string onto the stack before returning). I'll update that now.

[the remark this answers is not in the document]
jamescrosswell: To answer your original question though, I'm not sure why 1024. The default buffer size for StreamReader is 128 so we could just go with that. We have no way of knowing how big people's GraphQL requests are going to be or what an appropriate buffer size would be, so maybe we just run with the default.

src/Sentry.GraphQl/SentryGraphQlHttpMessageHandler.cs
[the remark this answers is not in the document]
jamescrosswell: Actually, that doesn't look great in the case of breadcrumbs. If the user hasn't specified an operation name, it's confusing to see something arbitrary like `graphql`... I think it makes sense for the Fingerprint but it doesn't really work for the Breadcrumbs.

For example, the breadcrumb below relates to this query (which doesn't have a query name):
```
{ test { id } }
```

Our breadcrumb suggests the actual query that was sent through is this:
```
query graphql { test { id } }
```

image

test/Sentry.Tests/ApiApprovalTests.Run.Core3_1.verified.txt
bitsandfoxes: I'm quoting @bruno-garcia here: Adding optional fields to a class is not a breaking change. Adding it to an interface would be.

src/Sentry/Internal/ObjectExtensions.cs
bitsandfoxes: That's a super nifty little library!

src/Sentry.GraphQl/SentryGraphQlHttpFailedRequestHandler.cs
bitsandfoxes: TIL. Thanks!

bitsandfoxes (approved):
This looks great!
I'm just going to drop a couple of follow-ups here:
- Custom sentry-exceptions for GraphQL errors. I think that's fine. We might want to think about the naming. `Sentry.SentryGraphQLHttpFailedRequestException` is quite the mouthful on the actual event
- Putting the error details in the exception instead of `"GraphQL Error"` would improve the way the event is displayed on Sentry
- Aggregating GraphQL errors. As you mentioned, going with the first error as the description sounds reasonable to me.
- We could capture the full request from Request.Data

PR #2558, opened 2023-08-20
fix: binding missing method

bruno-garcia:
Latest release of the cocoa SDK has some API changes: https://github.com/getsentry/sentry-cocoa/pull/3194

It seems that our CI builds a new binding, but doesn't break if the world changes.

We need to pin the cocoa SDK version and make sure we re-run the binding generation code in the bump PR. And if building the binding project results in a dirty repository, we need to break the build.

#skip-changelog

### DIFF
--- a/src/Sentry.Bindings.Cocoa/Sentry.Bindings.Cocoa.csproj
+++ b/src/Sentry.Bindings.Cocoa/Sentry.Bindings.Cocoa.csproj
@@ -60,7 +60,7 @@
   </Target>
   <Target Name="_InnerGenerateSentryCocoaBindings">
     <Exec Command="pwsh ../../scripts/generate-cocoa-bindings.ps1" />
-    <Exec Command="pwsh ../../scripts/dirty-check.ps1" />
+    <Exec Condition="'$(GITHUB_ACTIONS)' == 'true'" Command="pwsh ../../scripts/dirty-check.ps1" />
   </Target>
 
   <!-- Workaround for https://github.com/xamarin/xamarin-macios/issues/15299 -->
--- a/src/Sentry.Bindings.Cocoa/Sentry.Bindings.Cocoa.csproj
+++ b/src/Sentry.Bindings.Cocoa/Sentry.Bindings.Cocoa.csproj
@@ -54,12 +54,12 @@
   </Target>
 
   <!-- Generate bindings -->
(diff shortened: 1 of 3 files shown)


jamescrosswell:
See https://remarkablemark.org/blog/2017/10/12/check-git-dirty/

We should be able to use `git diff --stat` to see if anything has changed.

jamescrosswell (approved):

src/Sentry.Bindings.Cocoa/Sentry.Bindings.Cocoa.csproj
```diff
@@ -54,12 +54,13 @@
   </Target>
 
   <!-- Generate bindings -->
-  <Target Name="_GenerateSentryCocoaBindings" AfterTargets="_BuildSentryCocoaSDK" Condition="$([MSBuild]::IsOSPlatform('OSX'))"
-    Inputs="..\..\modules\sentry-cocoa\Carthage\.built-from-sha" Outputs="ApiDefinitions.cs;StructsAndEnums.cs">
+  <Target Name="_GenerateSentryCocoaBindings" AfterTargets="_BuildSentryCocoaSDK" Condition="$([MSBuild]::IsOSPlatform('OSX'))">
```
bruno-garcia, line 57: Oh, just noticed the input/output got removed. This means it'll run on each build, even if no changes were done. Was that the intention? @jamescrosswell ?

Was the `git submodule update` approach slower?

jamescrosswell: It was intentional yes. It will make the runs somewhere between 0.5s and 2s slower, which I figured wasn't consequential. We could potentially run `git submodule update --recursive` before all of this... I'm not sure if we'd continue to run into (other) problems though and it seems like a lot of complexity to add just to shave ~1s off the build time though. I'd rather remove the optimsation and be done with it.

bruno-garcia: Sounds good then. Surprised it takes only 2 seconds to run obj sharpie.

jamescrosswell: Yeah it's taking about 2.5s of user time on my machine today (although only 0.67s of system time):
image

PR #2570, opened 2023-08-21
restore submodule on each build

bruno-garcia:
#skip-changelog

### DIFF
--- a/before.Sentry.sln.targets
+++ b/before.Sentry.sln.targets
@@ -1,7 +1,6 @@
 <Project InitialTargets="RestoreSubmodule">
   <!-- If Ben.Demystifer is not found, restore git submodules -->
   <Target Name="RestoreSubmodule"
-    Condition="!Exists('modules\Ben.Demystifier\Ben.Demystifier.sln')">
     <Message Importance="High" Text="Ben.Demystifer not found. Restoring git submodules." />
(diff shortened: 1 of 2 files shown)


before.Sentry.sln.targets
```diff
@@ -1,7 +1,6 @@
 <Project InitialTargets="RestoreSubmodule">
   <!-- If Ben.Demystifer is not found, restore git submodules -->
   <Target Name="RestoreSubmodule"
-    Condition="!Exists('modules\Ben.Demystifier\Ben.Demystifier.sln')">
     <Message Importance="High" Text="Ben.Demystifer not found. Restoring git submodules." />
```
bruno-garcia, line 4:
```suggestion
```

before.Sentry.sln.targets
```diff
@@ -1,8 +1,6 @@
 <Project InitialTargets="RestoreSubmodule">
   <!-- If Ben.Demystifer is not found, restore git submodules -->
```
bruno-garcia, line 2:
```suggestion
  <!-- Always restore, to apply changes when we change branches. If no changes happened, this is pretty quick so no need to optimize by adding a Condition -->
```

jamescrosswell (approved):

Issue #2560, opened 2023-08-20
Discussion about Solution files, filters and CI


bruno-garcia:
I found it hard to get my head around this:
![image](https://github.com/getsentry/sentry-dotnet/assets/1633368/5e7cd548-c1b5-4a37-8fd2-c26d0f20c42f)

Mainly because:

1. `Sentry.sln` vs `Sentry.Full.sln` was not very intuitive to me. Shouldn't `Sentry.sln` already be "full"?
In that case, could it be `Sentry.sln` with everything, and just the filters on top? I understand there's some logic based on the solution file to avoid compiling the mobile targets:

https://github.com/getsentry/sentry-dotnet/blob/b3f81f1cf18c877029076e184ddef4b04acdef7e/Directory.Build.props#L18-L20

Is there another approach?

3. `Sentry-CI-Pack.slnf`: What does give us? Since `pack` happens on the same workflow where things are built, I imagine this saves from looking what projects need to be backed (`<IsPackable>true..`)? Samples are all marked as not packable already:

https://github.com/getsentry/sentry-dotnet/blob/b3f81f1cf18c877029076e184ddef4b04acdef7e/samples/Directory.Build.props#L6
Same with tests:
https://github.com/getsentry/sentry-dotnet/blob/b3f81f1cf18c877029076e184ddef4b04acdef7e/test/Directory.Build.props#L8

Is the 'speed up' worth having to maintain this additional filter file?

4. Too many `slfn` files? `SentryNoSamples.slnf`, `SentryAzureFunctions.slnf`, etc. Could this be just `Unload project` instead or are these used often enough?

The biggest problem I see is that everytime we add a new project, we need to deal with 3 solutions and all subsequent slnf because there's no way to use glob patterns, or negate (like everything **except** `/samples`, or for the CodeQL filter, just add everything under `src`, to name a couple).

5. `Sentry.Samples.OpenTelemetry.AspNet` classic msbuild project in the repo. Because `dotnet build` doesn't work here, there was an effort to keep these out, since the creation of this repo. It doesn't seem like this is built in CI (and that's probably good because it'd be slow, and require a separate build script/flow). There was another repo (https://github.com/getsentry/sentry-dotnet-samples and later https://github.com/getsentry/examples/tree/master/dotnet) to keep such samples that are rarely used and don't need to be built in CI on each PR.

[...]

PR #2571, opened 2023-08-22
NIT: Change build.ps1 to build.cmd

lucas-zimerman:
Build.ps1 was replaced by build.cmd but the reference wasn't changed on the source project: https://github.com/getsentry/sentry-dotnet/pull/2355

This PR adds the correct reference to build.cmd and removes build.ps1 

#skip-changelog


bruno-garcia:
How about we remove those solution items. Just acccess the files from the filesystem:

See: https://github.com/getsentry/sentry-dotnet/issues/2560

PR #2572, opened 2023-08-22
Fixed GraphQL server sample

jamescrosswell:
#skip-changelog

# Summary

We investigated whether we could implement everything outlined in Sentry's developer docs regarding graphql-dotnet servers. It turns out most of what is described in the developer docs isn't possible/useful with graphql-dotnet. As such, this PR simply updates our GraphQL Server sample to show how to instrument this correctly with OpenTelemetry and to illustrate how people can manually augment this to capture additional information, if they want. 

# Keeping for posterity

This is a summary of what we looked at doing before parring back to just updating our sample to illustrate what we get today out of the box with OpenTelemetry.

To auto-instrument GraphQL, there are two separate packages we're looking at supporting (initially):
1. [GraphQL .NET](https://github.com/graphql-dotnet/graphql-dotnet) (for building GraphQL servers)
2. [GraphQL Client](https://github.com/graphql-dotnet/graphql-client) (for building GraphQL clients)

This PR  pertains to the GraphQL.NET server integration.

### Requirements

- [GraphQL Performance](https://develop.sentry.dev/sdk/features/#graphql-performance)
  - [Server](https://develop.sentry.dev/sdk/features/#server)
  - [Breadcrumbs](https://develop.sentry.dev/sdk/features/#breadcrumbs-1)

## Overview 

[graphql-dotnet](https://github.com/graphql-dotnet/graphql-dotnet) can be used to create a GraphQL Server. This package supports OpenTelemetry, which we have leveraged and this gives fairly good instrumentation of the box:
image

## Limitations

Sentry's developer docs described various things that are difficult, impossible or just not particularly useful via graphql-dotnet's OpenTelemetry. 

### Request and Response context

And although we can see the query that gets sent to graphql-dotnet and also detect errors sent in response (see the sample application) the raw request/response payloads aren't accessible to us, so we can't set these in the Request or Response contexts as per the developer docs.

As such, currently we just illustrate how these things can be sent to Sentry as Extra tags.

### Spans for Resolvers and Data Loaders

We didn't find an appropriate way to capture spans for resolvers and data loaders, per the developer docs. 

One way might be to [enable metrics and use ApolloTracing](https://graphql-dotnet.github.io/docs/getting-started/metrics/), but the [ApolloTracing repo](https://github.com/apollographql/apollo-tracing) is archived due to server performance concerns.

Another option would be to make a pull request to the graphql-dotnet library, instrumenting these via OpenTelemetry there. 

For the time being, our .NET SDK isn't capturing spans for resolvers and data loaders for graphql-dotnet.


bitsandfoxes:
> If we're creating spans (and not transactions) for graphql requests on the server then, should these also omit the Request and Response interfaces? Or should we try to walk up the chain and set this information on the parent transaction that represents the HttpRequest and set the Request/Response interfaces on that span instead?

We should walk up the chain and set this on the parent transaction.

bitsandfoxes:
Regarding the error handling: I don't think we should intercept and report handled exceptions. (until we find a good reason or usecase for it)

jamescrosswell:
@bitsandfoxes regarding how we handle/display spans for GraphQL Client Queries that failed, if we're following the [OTEL guidelines to the letter](https://opentelemetry.io/docs/specs/otel/trace/semantic_conventions/http/#name):
* HTTP server span names SHOULD be {http.request.method} {http.route}
* For HTTP status codes in the 4xx range span status MUST be left unset in case of SpanKind.SERVER

That is a bit of a pain since there's no easy way to see in Sentry whether a client query that has been received by the server was successful or not, without digging into the event details and looking at the OpenTelemetry attributes (where you can see the http status on the response is something other than 200 OK). 

Here, for example, some of these `POST /graphql` operations resulted in 200 OK and others in 400 Bad Request... but they all look fine on the dashboard:
image

Digging into one that resulted in a client error, it all looks good up top:
image

But if you scroll down a bit, you can see the problem with the client query tucked in the OpenTelemetry attributes:
image

That seems really hidden to me. It will show up as a full blown client error in the client instrumentation... and with tracing enabled you can match those things up (as long as you control both the client and the server). 

What if you you just control the server though and you get an email from one of your customers saying, "Hey when I'm calling your graphql API I'm getting an error back"... and you want to track down the request that the client is referring to? There's currently no easy way to do it that I could find.

One solution would be to make it easy for people to search for these spans. If you could plug something like `http.status_code: 400` into the **_search transactions_** box  in the Sentry dashboard and have it locate the appropriate transaction, I think that would be workable. For my own part, I have no idea how to use that  transaction search box - I tried all sorts of combinations and [the docs](https://docs.sentry.io/product/sentry-basics/search/) don't seem that helpful (they talk about properties for errors and events, but not how to search for performance transactions/spans).

Any idea how to search for transactions/spans? Is it possible to search by otel attributes?

jamescrosswell:
> > If we're creating spans (and not transactions) for graphql requests on the server then, should these also omit the Request and Response interfaces? Or should we try to walk up the chain and set this information on the parent transaction that represents the HttpRequest and set the Request/Response interfaces on that span instead?
> 
> We should walk up the chain and set this on the parent transaction.

It turns out we can't really capture the information we'd like to put in the Request/Response objects anyway. I [tried this](https://github.com/getsentry/sentry-dotnet/blob/5b72cad61e1d674e1ec64af19464211fa8350c5e/src/Sentry.OpenTelemetry/Enrichers.cs) but when instrumenting with OpenTelemetry, the original Request and Response objects aren't available to us. The only things we can get at are parsed/structured information (such as the "document"... which is actually only one of various key value pairs that actually gets sent as a dictinoary object in the underlying request). I think it's misleading to use that as the RequestContext.Data (since in Sentry this gets presented as the `Body` of the HttpRequest):

image

The developer documents want us to capture this:

> The request.data should be set with the raw GraphQL request payload, if possible. This should be guarded by an opt-in flag, e.g. sendDefaultPii.
>
> The contexts.response.data should be set with the raw GraphQL response payload, only if there were errors. This should be guarded by an opt-in flag, e.g. sendDefaultPii and maxResponseBodySize.

So there's no tidy way to do this. What I might do instead is extend the sample for the server to show how people can capture things like errors from the response in custom attributes, if they want to.

bitsandfoxes (approved):
Thanks!

PR #2574, opened 2023-08-22
fix: upload proguard mapping

bruno-garcia:
1. fix: proguard upload command invocation
2. fix: allow enable only proguard command (without enabling debug file upload)
3. Check command exit code and print message if failed
4. enabled R8 in our sample

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@
 
 - Graphql client ([#2538](https://github.com/getsentry/sentry-dotnet/pull/2538))
 
+###  Fixes
+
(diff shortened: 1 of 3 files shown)


jamescrosswell (approved):

bruno-garcia:
I opened a follow up ticket for integration testing this: https://github.com/getsentry/sentry-dotnet/issues/2575
Maybe @vaind can help?

I tried running the script locally to add the test, but it fails to run completely with

```
Line |
  12 |  BeforeAll {
     |  ~~~~~~~~~
     | The term 'BeforeAll' is not recognized as a name of a cmdlet, function, script file, or executable program. Check the spelling of the name, or if a path was included, verify that the path is correct and try again.
```

Not sure what's wrong on my local dev right now

Issue #2549, opened 2023-08-17
Unable to add attachments to events with User Feedback in BeforeSend or SetBeforeSend in v3.35.1


Fwang36:
### Package

Sentry

### .NET Flavor

.NET

### .NET Version

7.01

### OS

macOS

### SDK Version

3.35.1

### Steps to Reproduce

**Create event with User Feedback like this -**
```
var eventId = SentrySdk.CaptureMessage("An event that will receive user feedback.");
SentrySdk.CaptureUserFeedback(eventId, "fran@gmail", "help.", "The User");
```

**From within BeforeSend, add an attachment to the event with configureScope()**
```
    options.BeforeSend = @event =>
    {
        SentrySdk.ConfigureScope(scope =>
        {

            scope.AddAttachment("logger.txt");

        });
        return @event;
    };
```

### Expected Result

Attachment gets attached to the event in beforeSend. 

### Actual Result

Attachment does not get attached.  

Downgrading to 3.31.0 causes this method to work again.  In 3.35.1, normal captureMessages without User Feedback will also work.

**This is an issue from my org which is mix of 3.35.0 and 3.31.0 events. Note that the recent 3.35.0/1 events do not have attachments.  The initial 3 attachments for this issue are 3.35.0 but these were attached using Local Scope and not in BeforeSend**

[User Feedback Issue From my org](https://francisorg.sentry.io/issues/4399139246/events/latest/?query=is%3Aunresolved&referrer=latest-event&stream_index=0)

bitsandfoxes:
Hey there, sorry to see you run into this issue.
There seems to be a gap in our documentation regarding how to utilize `hints` that were introduced in `3.32.0`. We'll fix that. 

In short, the scope gets applied to the event before it hits the `BeforeSend` callback which means modifying the scope within the callback does not affect the event itself.

Instead of the deprecated `BeforeSend` you can (and should) use the `SetBeforeSend` method on the options that allows you to add the attachment to its `hint`.

```
options.SetBeforeSend((@event, hint) =>
{
    hint.Attachments.Add(new Attachment(
        AttachmentType.Default,
        new FileAttachmentContent("<filepath>", true),
        Path.GetFileName("<filepath>"),
        null));
    return @event;
});
```

Admittedly, this is not the most straightforward of user experiences. We'll see if we can make this smoother.

PR #2585, opened 2023-08-29
feat: Added `AddAttachment` to the `Hint`

bitsandfoxes:
Resolves: https://github.com/getsentry/sentry-dotnet/issues/2549

Taken from the API we have on the scope to add an attachment. So instead of 
```
options.SetBeforeSend((@event, hint) =>
{
    hint.Attachments.Add(new Attachment(
        AttachmentType.Default,
        new FileAttachmentContent("<filepath>", true),
        Path.GetFileName("<filepath>"),
        null));
    return @event;
});
```

we can do 
```
options.SetBeforeSend((@event, hint) =>
{
    hint.AddAttachment("<filepath>");
    return @event;
});
```

### DIFF
--- a/test/Sentry.Tests/HintTests.cs
+++ b/test/Sentry.Tests/HintTests.cs
@@ -19,7 +19,7 @@ public class HintTests : IDisposable
         File.Create(attachmentPath1);
         File.Create(attachmentPath2);
 
-        var hint = new Hint();
+        var hint = new Hint(new SentryOptions());
 
         // Act
         hint.AddAttachment(attachmentPath1);
--- a/test/Sentry.Tests/HintTests.cs
+++ b/test/Sentry.Tests/HintTests.cs
@@ -1,18 +1,29 @@
 namespace Sentry.Tests;
 
-public class HintTests
+public class HintTests : IDisposable
 {
+    private readonly string _testDirectory;
+
+    public HintTests()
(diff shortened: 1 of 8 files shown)


jamescrosswell:
@bitsandfoxes do hints get sent to Sentry? I thought they were just there for context prior to capturing an event (so useful in the BeforeSend callback but not afterwards)...

bitsandfoxes:
> @bitsandfoxes do hints get sent to Sentry? I thought they were just there for context prior to capturing an event (so useful in the BeforeSend callback but not afterwards)...

Yes they do. `Hints` have a bit of a mixed responsibility but the attachment part is supposed to be the answer to "how do I add attachments in the `BeforeSend` callback?".

test/Sentry.Tests/HintTests.cs
jamescrosswell, line 1: Looks like we have good basic tests to ensure attachments can be added to the Hints. 

Maybe worth adding a test to the Hub to ensure Hint.Attachments make their way through to Sentry when capturing an event?

jamescrosswell (approved):

test/Sentry.Tests/HintTests.cs
bitsandfoxes: Good catch! Thanks for the callout!

Issue #2591, opened 2023-09-01
DynamicSamplingContext Exceptions: Attempted to set the DynamicSamplingContext but the context exists already


ian-a-anderson:
### Package

Sentry.AspNetCore

### .NET Flavor

.NET Core

### .NET Version

6.0.0

### OS

Linux

### SDK Version

3.36.0

### Steps to Reproduce

These errors seem to have begun at the beginning of August, after the following PR landed:
https://github.com/getsentry/sentry-dotnet/pull/2493

They fire intermittently as the result of cross-api backend calls where we are forwarding sentry headers along to another request (for distributed tracing). As a result I don't have a consistent repro. Mainly I would like a way to disable these from throwing until a resolution can be found, but since this is an exception coming from Sentry internals I'm not sure how best to avoid it.

It appears this functionality is tied to [Dynamic Sampling](https://develop.sentry.dev/sdk/performance/dynamic-sampling-context/) so I assumed that disabling Performance tracing would prevent the issue. But it appears that you can't disable backend/API performance tracing if it is enabled on an incoming request from a frontend client that has it enabled (for distributed tracing). 

### Expected Result

No exception thrown (possibly a logged error instead?)

### Actual Result

<img width="785" alt="image" src="https://github.com/getsentry/sentry-dotnet/assets/2657265/da62ce82-9fd0-4b39-aaba-af5fa1cecb13">

bitsandfoxes:
Sorry about that. The SDK definitely should not just throw. Fix is coming asap.

PR #2592, opened 2023-09-01
fix: 

bitsandfoxes:
Resolves https://github.com/getsentry/sentry-dotnet/issues/2591

I don't understand how it even gets there. The property is guarded by null-coalescing operators. But the SDK should not just throw regardless.

So now we have a getter that just creates the context in case it's `null`.

### DIFF
--- a/src/Sentry/SentryPropagationContext.cs
+++ b/src/Sentry/SentryPropagationContext.cs
@@ -10,7 +10,16 @@ internal class SentryPropagationContext
 
     internal DynamicSamplingContext? _dynamicSamplingContext;
 
-    public DynamicSamplingContext GetDynamicSamplingContext(SentryOptions options) => _dynamicSamplingContext ??= this.CreateDynamicSamplingContext(options);
+    public DynamicSamplingContext GetDynamicSamplingContext(SentryOptions options)
+    {
+        if (_dynamicSamplingContext is null)
+        {
+            options.LogDebug("Creating the Dynamic Sampling Context from the Propagation Context");
+            _dynamicSamplingContext = this.CreateDynamicSamplingContext(options);
+        }
+
+        return _dynamicSamplingContext;
+    }
 
     internal SentryPropagationContext(
         SentryId traceId,
--- a/src/Sentry/SentryPropagationContext.cs
+++ b/src/Sentry/SentryPropagationContext.cs
@@ -8,22 +8,9 @@ internal class SentryPropagationContext
     public SpanId SpanId { get; }
     public SpanId? ParentSpanId { get; }
 
-    private DynamicSamplingContext? _dynamicSamplingContext;
-    public DynamicSamplingContext? DynamicSamplingContext
-    {
-        get => _dynamicSamplingContext;
-        set
-        {
-            if (_dynamicSamplingContext is null)
-            {
-                _dynamicSamplingContext = value;
-            }
-            else
-            {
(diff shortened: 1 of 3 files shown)


jamescrosswell (commented):
@bitsandfoxes I'm not sure I get this. Other than adding logging (when creating the dynamic sampling context from the propagation context) what has changed here?

bitsandfoxes:
The SDK is now no longer throwing exceptions. Instead of being opinionated regarding the setter being called only once we defensively handle the dynamic sampling context creation in the getter.

jamescrosswell:
> The SDK is now no longer throwing exceptions. Instead of being opinionated regarding the setter being called only once we defensively handle the dynamic sampling context creation in the getter.

Ah I see, there's no setter for DynamicSamplingContext anymore...

src/Sentry/SentryPropagationContext.cs
```diff
@@ -33,13 +29,19 @@ internal class SentryPropagationContext
         TraceId = traceId;
         SpanId = SpanId.Create();
         ParentSpanId = parentSpanId;
-        DynamicSamplingContext = dynamicSamplingContext;
+        _dynamicSamplingContext = dynamicSamplingContext;
     }
 
-    public SentryPropagationContext()
+    public SentryPropagationContext() : this(null)
+    { }
```
jamescrosswell, line 36: When this constructor gets called in the `CreateFromHeaders` method, won't the TraceId and SpanId be empty? Is there ever a scenario where we'd want that to be the case? 

Even if we don't get it from upstream, I think we always want a TraceId to link this span to downstream events right?

test/Sentry.Tests/SentryPropagationContextTests.cs
```diff
@@ -2,6 +2,43 @@ namespace Sentry.Tests;
...
+        var propagationContext = new SentryPropagationContext();
+        var firstDynamicSamplingContext = propagationContext.GetOrCreateDynamicSamplingContext(options);
+
+        var secondDynamicSamplingContext = propagationContext.GetOrCreateDynamicSamplingContext(options);
+
+        Assert.Same(firstDynamicSamplingContext, secondDynamicSamplingContext);
+    }
+
     [Fact]
     public void CreateFromHeaders_TraceHeaderNullButBaggageExists_CreatesPropagationContextWithoutDynamicSamplingContext()
```
jamescrosswell, line 43: I think we need another test where the traceheader is null and there is no baggage, to make explicit what our expectations are in that scenario.

test/Sentry.Tests/ScopeTests.cs
```diff
@@ -98,7 +98,8 @@ public void Clone_NewScope_IncludesPropagationContext()
 
         var clone = sut.Clone();
 
-        Assert.Same(propagationContext, clone.PropagationContext);
+        Assert.Equal(propagationContext.TraceId, clone.PropagationContext.TraceId);
+        Assert.Equal(propagationContext.SpanId, clone.PropagationContext.SpanId);
```
jamescrosswell, line 102: Maybe also assert that the references are not eqal (i.e. this is a clone... not the same object)?

jamescrosswell (changes requested):
Looks good generally. I just had some questions about the SentryPropagationContext constructor but otherwise good to go I think.

jamescrosswell (approved):
The code itself looks good... once our CI build environment is healthy again 👍🏻

Issue #2597, opened 2023-09-06
Can't create BaggageHeader for SentrySdk.ContinueTrace(), it's not public


j2ghz:
### Package

Sentry

### .NET Flavor

.NET Core

### .NET Version

7.0.0

### OS

Windows

### SDK Version

3.36.0

### Steps to Reproduce

- Try using SentrySdk.ContinueTrace(), where the second parameter is BaggageHeader
- See that there's no way to create an instance of BaggageHeader

### Expected Result

I expect that BaggageHeader should have a public contructor or a public Parse/TryParse method

### Actual Result

BaggageHeader doesn't have any public methods or constructors (expect an instance .ToString() method)
See: https://github.com/getsentry/sentry-dotnet/blob/3.36.0/src/Sentry/BaggageHeader.cs

bitsandfoxes:
Thanks for raising this. Fair point. That's an oversight but I'm not sure how to proceed here yet. 

I'd like to avoid making the `HttpContextExtensions` methods public. I think we should rather provide overloads to the `ContinueTrace` to consume the headers more generically i.e. letting the SDK extract them from the `System.Web.HttpContext` and `Microsoft.AspNetCore.Http`.

j2ghz:
> i.e. letting the SDK extract them from the `System.Web.HttpContext` and `Microsoft.AspNetCore.Http`.

I don't think that would help in my case, because I'm trying to set up tracing for a transport that isn't HTTP (in my case it's Service Fabric TCP transport). I took inspiration from https://github.com/getsentry/sentry-dotnet/blob/3.36.0/src/Sentry.AspNetCore/SentryTracingMiddleware.cs, and on the transmitting side I can use BaggageHeader.ToString(), but there's nothing available on the receiving side. For comparison, SentryTraceHeader.Parse is public.

bitsandfoxes:
Thanks for the additional input! We're going to add an overload to the `ContinueTrace` that accepts the headers as strings.

bruno-garcia:
@j2ghz did this change resolve it for you? https://github.com/getsentry/sentry-dotnet/pull/2601

j2ghz:
@bruno-garcia I haven't gotten around to testing it, but it sounds like it should be exactly what I needed. Thanks.

PR #2601, opened 2023-09-07
fix: Overload for `ContinueTrace` to accept headers as `string`

bitsandfoxes:
Resolves https://github.com/getsentry/sentry-dotnet/issues/2597


src/Sentry/Internal/Hub.cs
```diff
@@ -219,6 +219,27 @@ public BaggageHeader GetBaggage()
...
     }
 
+    public TransactionContext ContinueTrace(
+        string? traceHeader,
+        string? baggageHeader,
+        string? name = null,
+        string? operation = null)
+    {
+        SentryTraceHeader? sentryTraceHeader = null;
+        if (traceHeader != null)
```
bruno-garcia, line 229: we need a linter to help us move towards:

```suggestion
        if (traceHeader is not null)
```

bruno-garcia:
LGTM basicaly but missing tests I imagine is the reason to be a draft?

src/Sentry/Internal/Hub.cs
```diff
@@ -219,6 +219,27 @@ public BaggageHeader GetBaggage()
...
+        string? operation = null)
+    {
+        SentryTraceHeader? sentryTraceHeader = null;
+        if (traceHeader is not null)
+        {
+            sentryTraceHeader = SentryTraceHeader.Parse(traceHeader);
+        }
+
+        BaggageHeader? sentryBaggageHeader = null;
+        if (baggageHeader != null)
```
bitsandfoxes, line 235:
```suggestion
        if (baggageHeader is not null)
```

bruno-garcia (approved):

Issue #2202, opened 2023-02-28
Sentry tracing middleware should be added automatically


mattjohnsonpint:
### Problem Statement

In an ASP.NET Core app, enabling Sentry tracing currently requires two parts:
- Enabling tracing in general, using `EnableTracing`, `TracesSampleRate`, or `TracesSampler`
- Enabling automatic instrumentation, using `app.UseSentryTracing()` to add the ASP.NET Core tracing middleware

It's easy to forget adding one step or the other.  It's also easy to place the middleware at the wrong location, since it has to be after `UseRouting` and before `UseEndpoints`.

### Solution Brainstorm

If tracing is enabled, we can add the tracing middleware automatically at the correct location.  We could then deprecate `UseSentryTracing`.

There's a detailed example of how to do this in the following blog posts:
1. https://andrewlock.net/inserting-middleware-between-userouting-and-useendpoints-as-a-library-author-part-1/
2. https://andrewlock.net/inserting-middleware-between-userouting-and-useendpoints-as-a-library-author-part-2/

### Alternative solution

Andrew's solution is quite ingenious... a simpler partial solution would be to create an extension that people would use instead of `UseRouting`:

```csharp
    /// <summary>
    /// Adds Sentry's tracing middleware to the pipeline.
    /// Call this instead of <c>builder.UseRouting()</c>.
    /// </summary>
    public static IApplicationBuilder UseSentryTraceRouting(this IApplicationBuilder builder)
    {
        builder.UseRouting();
        builder.UseMiddleware<SentryTracingMiddleware>();
        return builder;
    }
```

And then when people are wiring up their middleware they could just do this (see commented code... which is what gets replaced):

```csharp
    .Configure(app =>
    {
        // app.UseRouting();
        // app.UseSentryTracing();
        app.UseSentryTraceRouting();

        app.UseEndpoints(routes =>
        {
            routes.Map("/person/{id}", async ctx =>
            {
                var id = ctx.GetRouteValue("id") as string;
                await ctx.Response.WriteAsync($"Person #{id}");
            });
        });
    });
```

This doesn't solve the problem of also requiring `EnableTracing`, `TracesSampleRate`, or `TracesSampler` to be set on the options however, so Tracing still needs to be configured in two locations with this solution.

mattjohnsonpint:
We think this is an important item, because it helps customers fall into the pit of success with setting up Sentry for ASP.NET Core.

mattjohnsonpint:
[...]

PR #2602, opened 2023-09-08
Configure Sentry tracing middleware automatically

jamescrosswell:
Addresses [Sentry tracing middleware should be added automatically #2202](https://github.com/getsentry/sentry-dotnet/issues/2202).

With this PR, it's almost never necessary for SDK users to call `UseSentryTracing` anymore, which will hopefully make it way easier for SDK users to get started with Sentry Tracing.

## Approach

We've used a much simpler variant of Andrew Lock's solution:
- We register `SentryTracingStartupFilter` as a `IStartupFilter`
  - That filter wraps the `IApplicationBuilder` in a `SentryTracingBuilder` so that we can intercept all of the calls to `builder.Use<T>`
- On each call to `Use<T>` our custom builder checks to see if the `__EndpointRouteBuilder` property has been set on the builder... which is a marker for `UseRouting`. As soon as that property is set, we know `UseRouting` has been called and we can then register our tracing middleware
- One other thing we do, before registering the middleware, is check that `SentryAspNetCoreOptions.Instrumenter == Instrumeter.Sentry`. If not, we don't register the tracing middleware... this would happen if folks were using OpenTelemetry rather than Sentry to auto-instrument their AspNET Core applications. 

## Caveats

[Sentry.Google.Cloud.Functions](https://github.com/getsentry/sentry-dotnet/blob/3554dabf6f9b0cc79ac6b8f0f5984c057de7a80d/src/Sentry.Google.Cloud.Functions/SentryStartup.cs#L85) is one of the few places I saw where I think users would still need to make the call to UseSentryTracing. There might be some other outlier scenarios like that but for the bulk of users, using vanilla ASP.NET Core, the developer experience should be way less complicated.

### DIFF
--- a/src/Sentry.AspNetCore/SentryTracingBuilder.cs
+++ b/src/Sentry.AspNetCore/SentryTracingBuilder.cs
@@ -41,8 +41,8 @@ internal class SentryTracingBuilder : IApplicationBuilder
     {
         var builder = InnerBuilder.Use(middleware);
         return (Properties.ContainsKey(EndpointRouteBuilder) && !this.IsSentryTracingRegistered())
-            // UseRouting sets a property on the builder that we can to detect when UseRouting has been added... we then
-            // register SentryTracing immediately afterwards (the best place for it)
+            // UseRouting sets a property on the builder that we use to detect when UseRouting has been added and we
+            // register SentryTracing immediately afterwards
             // https://github.com/dotnet/aspnetcore/blob/8eaf4b51f73ae2b0ed079e4f8253afb53e96b703/src/Http/Routing/src/Builder/EndpointRoutingApplicationBuilderExtensions.cs#L58-L62
             ? builder.UseSentryTracing()
             : builder;
--- a/src/Sentry.AspNetCore/SentryTracingBuilder.cs
+++ b/src/Sentry.AspNetCore/SentryTracingBuilder.cs
@@ -1,0 +1,50 @@
+using Microsoft.AspNetCore.Builder;
+using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Http.Features;
+
+namespace Sentry.AspNetCore;
+
+/// <summary>
+/// Delegates/wraps the inner builder, so that we can intercept calls to <see cref="Use"/> and add our middleware
+/// </summary>
+internal class SentryTracingBuilder : IApplicationBuilder
+{
+    private const string EndpointRouteBuilder = "__EndpointRouteBuilder";
+    public SentryTracingBuilder(IApplicationBuilder inner)
+    {
+        InnerBuilder = inner;
+    }
+
+    private IApplicationBuilder InnerBuilder { get; }
+
+    /// <inheritdoc />
+    public IServiceProvider ApplicationServices
+    {
+        get => InnerBuilder.ApplicationServices;
+        set => InnerBuilder.ApplicationServices = value;
+    }
+
+    /// <inheritdoc />
+    public IDictionary<string, object?> Properties => InnerBuilder.Properties;
+
+    /// <inheritdoc />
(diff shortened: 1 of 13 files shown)


src/Sentry.AspNetCore/SentryTracingBuilder.cs
```diff
@@ -0,0 +1,61 @@
...
+    /// <inheritdoc />
+    public IApplicationBuilder New() => new SentryTracingBuilder(InnerBuilder.New());
+
+    /// <inheritdoc />
+    public IApplicationBuilder Use(Func<RequestDelegate, RequestDelegate> middleware)
+    {
+        // UseRouting sets a property on the builder that we use to detect when UseRouting has been added and we
+        // register SentryTracing immediately afterwards
+        // https://github.com/dotnet/aspnetcore/blob/8eaf4b51f73ae2b0ed079e4f8253afb53e96b703/src/Http/Routing/src/Builder/EndpointRoutingApplicationBuilderExtensions.cs#L58-L62
+        if (Properties.ContainsKey(EndpointRouteBuilder) && this.ShouldRegisterSentryTracing())
```
bitsandfoxes, line 48: Do I get that correctly: Since the `UseRouting` sets a flag and we check for that flag before forwarding to the InnerBuilder, Sentry will add the `TracingMiddleware` with the `UseX` that follows after `UseRouting`?

src/Sentry.AspNetCore/SentryTracingMiddlewareExtensions.cs
```diff
@@ -9,12 +11,40 @@ namespace Microsoft.AspNetCore.Builder;
...
     /// Adds Sentry's tracing middleware to the pipeline.
     /// Make sure to place this middleware after <c>UseRouting(...)</c>.
     /// </summary>
     public static IApplicationBuilder UseSentryTracing(this IApplicationBuilder builder)
     {
+        // Don't register twice
+        if (builder.IsSentryTracingRegistered())
+        {
+            return builder;
+        }
```
bitsandfoxes, line 45: That's a good catch. And that means users are not forced to update their existing setup. 👍

bitsandfoxes (approved):
I take it the `Sentry.sln.DotSettings` is not needed? Do we need to add it to the `.gitignore`?

This is awesome! More out-of-the-box, less documentation needed and one pitfall! With one PR! 🔥

src/Sentry.AspNetCore/SentryTracingBuilder.cs
jamescrosswell: Yeah this is what happens in the dotnet repo:

https://github.com/dotnet/aspnetcore/blob/8eaf4b51f73ae2b0ed079e4f8253afb53e96b703/src/Http/Routing/src/Builder/EndpointRoutingApplicationBuilderExtensions.cs#L58-L62

You can see they set that property immediately before calling `UseMiddleware` (which is then intercepted by `SentryTracingBuilder`). As soon as we see this property on the builder then, we know a call has been made to `UseRouting` and we can safely register our tracing middleware. 

It shouldn't matter whether something comes after UseRouting or not... the way it sits at the moment, we would detect the call to UseRouting, the property would already have been set at that stage and we'd register both the UseRouting middleware and the UseSentryTracing middleware. If anything gets registered after UseRouting or not doesn't matter much then.

Initially I was making the call to `InnerBuilder.Use(middleware)` before checking for the existence of the property... just in case Microsoft ever changed the order in which they made the call to UseMiddleware and set the build property. However that would result in an extra (potentially unecessary) operation.

jamescrosswell:
> I take it the `Sentry.sln.DotSettings` is not needed? Do we need to add it to the `.gitignore`?

That change was to add `instrumenter` to our shared dictionary actually... that file is where you can configure settings that are shared across the whole team, so it's supposed to be included in the git repo.

Issue #2616, opened 2023-09-13
Identifying Users with OpenTelemetry doesn't work


ldalonzo:
### Package

Sentry

### .NET Flavor

.NET Core

### .NET Version

7.0.0

### OS

Windows

### SDK Version

3.39.1

### Steps to Reproduce

1. Instrument an ASP.NET Core with Sentry.OpenTelemetry
2. Configure Sentry to opt-in for `SendDefaultPii`
3. Register a custom `IUserFactory` to populate the sentry username

### Expected Result

The custom user factory gets called and the username is reported to Sentry

### Actual Result

The custom `IUserFactory` is never called. Only the IP address is reported to Sentry.
A similar behaviour applies when setting the username through the hub's `ConfigureScope` method.

bruno-garcia:
For context, came up from: https://stackoverflow.com/questions/58858092/sentry-sdk-dotnet-core-not-using-iuserfactory/58868882?noredirect=1#comment135908613_58868882

jamescrosswell:
Hi @ldalonzo, thanks for raising a ticket for this. I can reproduce this easily enough, so we should be able to come up with a solution.

It doesn't necessarily require a custom `IUserFactory` either... we see the same behaviour using the `DefaultUserFactory`.

When not using OpenTelemetry, this is enabled via our `SentryMiddleware`:
https://github.com/getsentry/sentry-dotnet/blob/55ca61eb90b7a269b950d95500502e154f16787d/src/Sentry.AspNetCore/SentryMiddleware.cs#L107-L123

`Sentry.OpenTelemetry` is agnostic of the application type being deployed... so it doesn't know about things like `Microsoft.AspNetCore.Http.HttpContext`, that only exist in ASP.NET Core. 

@bruno-garcia @bitsandfoxes fyi - a bit of a pain but  I think to make this work I'll have to add a `Sentry.OpenTelemery.AspNetCore` package that takes dependencies on `Microsoft.AspNetCore.Http` (so we can get at the `HttpContext`.

PR #2618, opened 2023-09-14
Fixes: Identifying Users with OpenTelemetry doesn't work

jamescrosswell:
Fixes [Identifying Users with OpenTelemetry doesn't work #2616](https://github.com/getsentry/sentry-dotnet/issues/2616).

### Key implementation details
- The addition of: https://github.com/getsentry/sentry-dotnet/blob/ad6968c04ecdc2775692dddf48852da7090aef2b/src/Sentry.AspNetCore/Extensions/DependencyInjection/ServiceCollectionExtensions.cs#L25
- Added `ISentryUserFactory` to the `DefaultUserFactory` (eventually this should replace `IUserFactory`). This enables us to get the user data _without having to add any dependencies_ to the `Sentry.OpenTelemetry` package. See: https://github.com/getsentry/sentry-dotnet/blob/ad6968c04ecdc2775692dddf48852da7090aef2b/src/Sentry.AspNetCore/IUserFactory.cs#L10-L11
- Introduced an `AspNetCoreEnricher`: https://github.com/getsentry/sentry-dotnet/blob/ad6968c04ecdc2775692dddf48852da7090aef2b/src/Sentry.OpenTelemetry/TracerProviderBuilderExtensions.cs#L37-L42

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@
 
 ### Fixes
 
-- Identifying Users with OpenTelemetry doesn't work ([#2618](https://github.com/getsentry/sentry-dotnet/pull/2618))
+- Resolved issue identifying users with OpenTelemetry ([#2618](https://github.com/getsentry/sentry-dotnet/pull/2618))
 
 ### Dependencies
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,11 +6,16 @@
 
 - Sentry tracing middleware now gets configured automatically ([#2602](https://github.com/getsentry/sentry-dotnet/pull/2602))
 
+### Fixes
+
+- Identifying Users with OpenTelemetry doesn't work ([#2618](https://github.com/getsentry/sentry-dotnet/pull/2618))
+
 ### Dependencies
 
 - Bump CLI from v2.20.6 to v2.20.7 ([#2604](https://github.com/getsentry/sentry-dotnet/pull/2604))
   - [changelog](https://github.com/getsentry/sentry-cli/blob/master/CHANGELOG.md#2207)
   - [diff](https://github.com/getsentry/sentry-cli/compare/2.20.6...2.20.7)
+
 ## 3.39.1
 
 ### Fixes
--- a/src/Sentry.AspNetCore/IUserFactory.cs
+++ b/src/Sentry.AspNetCore/IUserFactory.cs
@@ -3,7 +3,13 @@ using Microsoft.AspNetCore.Http;
 namespace Sentry.AspNetCore;
 
 /// <summary>
+/// <para>
 /// Sentry User Factory
+/// </para>
+/// <para>
+/// Note: This interface is tightly coupled to AspNetCore and Will be removed in version 4.0.0. Please consider using
+/// <see cref="ISentryUserFactory"/> with <see cref="IHttpContextAccessor"/> instead.
+/// </para>
 /// </summary>
 public interface IUserFactory
 {
--- a/src/Sentry.AspNetCore/IUserFactory.cs
+++ b/src/Sentry.AspNetCore/IUserFactory.cs
@@ -3,14 +3,9 @@ using Microsoft.AspNetCore.Http;
 namespace Sentry.AspNetCore;
 
 /// <summary>
-/// <para>
 /// Sentry User Factory
-/// </para>
-/// <para>
-/// Note: This interface is tightly coupled to AspNetCore and Will be removed in version 4.0.0. Please consider using
-/// <see cref="ISentryUserFactory"/> with <see cref="IHttpContextAccessor"/> instead.
-/// </para>
 /// </summary>
+[Obsolete("This interface is tightly coupled to AspNetCore and will be removed in version 4.0.0. Please consider using ISentryUserFactory with IHttpContextAccessor instead.")]
 public interface IUserFactory
 {
     /// <summary>
--- a/src/Sentry.OpenTelemetry/AspNetCoreEnricher.cs
+++ b/src/Sentry.OpenTelemetry/AspNetCoreEnricher.cs
@@ -1,0 +1,26 @@
+﻿namespace Sentry.OpenTelemetry;
+
+internal class AspNetCoreEnricher : IOpenTelemetryEnricher
+{
+    private readonly ISentryUserFactory _userFactory;
+
+    internal AspNetCoreEnricher(ISentryUserFactory userFactory)
+    {
+        _userFactory = userFactory;
+    }
+
+    public void Enrich(ISpan span, Activity activity, IHub hub, SentryOptions? options)
+    {
+        var sendPii = options?.SendDefaultPii ?? false;
+        if (sendPii)
+        {
+            hub.ConfigureScope(scope =>
+            {
+                if (!scope.HasUser() && _userFactory.Create() is {} user)
+                {
+                    scope.User = user;
+                }
+            });
+        }
+    }
+}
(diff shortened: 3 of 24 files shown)


src/Sentry.AspNetCore/IUserFactory.cs
```diff
@@ -3,7 +3,13 @@
 namespace Sentry.AspNetCore;
 
 /// <summary>
+/// <para>
 /// Sentry User Factory
+/// </para>
+/// <para>
+/// Note: This interface is tightly coupled to AspNetCore and Will be removed in version 4.0.0. Please consider using
```
bruno-garcia, line 10: If we plan to remove it on 4.0.0, should this be marked as `[Obsolete]`?
You can `#pragma` ignore the usage on `DefaultUserFactory` until we get rid of it

src/Sentry.OpenTelemetry/AspNetCoreEnricher.cs
```diff
@@ -0,0 +1,26 @@
...
+
+    internal AspNetCoreEnricher(ISentryUserFactory userFactory)
+    {
+        _userFactory = userFactory;
+    }
+
+    public void Enrich(ISpan span, Activity activity, IHub hub, SentryOptions? options)
+    {
+        var sendPii = options?.SendDefaultPii ?? false;
+        if (sendPii)
```
bruno-garcia, line 15: nit:

```suggestion
        if (options?.SendDefaultPii is true)
```

src/Sentry.OpenTelemetry/AspNetCoreEnricher.cs
```diff
@@ -0,0 +1,26 @@
+﻿namespace Sentry.OpenTelemetry;
+
+internal class AspNetCoreEnricher : IOpenTelemetryEnricher
+{
+    private readonly ISentryUserFactory _userFactory;
+
+    internal AspNetCoreEnricher(ISentryUserFactory userFactory)
+    {
+        _userFactory = userFactory;
+    }
```
bruno-garcia, line 10: nit:
```suggestion
    internal AspNetCoreEnricher(ISentryUserFactory userFactory) => _userFactory = userFactory;
```

src/Sentry.OpenTelemetry/SentrySpanProcessor.cs
```diff
@@ -27,9 +29,14 @@ public SentrySpanProcessor() : this(SentrySdk.CurrentHub)
     /// <summary>
     /// Constructs a <see cref="SentrySpanProcessor"/>.
     /// </summary>
-    public SentrySpanProcessor(IHub hub)
+    public SentrySpanProcessor(IHub hub) : this(hub, null)
+    {
+    }
+
+    internal SentrySpanProcessor(IHub hub, List<IOpenTelemetryEnricher>? enrichers)
```
bruno-garcia, line 36: Do we need a list here or just something we can enumerate?
```suggestion
    internal SentrySpanProcessor(IHub hub, IEnumerable<IOpenTelemetryEnricher>? enrichers)
```

src/Sentry.OpenTelemetry/TracerProviderBuilderExtensions.cs
```diff
@@ -29,6 +30,18 @@ public static TracerProviderBuilder AddSentry(this TracerProviderBuilder tracerP
...
         defaultTextMapPropagator ??= new SentryPropagator();
         Sdk.SetDefaultTextMapPropagator(defaultTextMapPropagator);
-        return tracerProviderBuilder.AddProcessor<SentrySpanProcessor>();
+        return tracerProviderBuilder.AddProcessor(services =>
+        {
+            List<IOpenTelemetryEnricher> enrichers = new();
+
+            // AspNetCoreEnricher
+            var userFactory = services.GetService<ISentryUserFactory>();
+            if (userFactory != null)
```
bruno-garcia, line 39: Would be nice to get a linter that suggests this instead. Since `!=` is an operator

```suggestion
            if (userFactory is not null)
```

bruno-garcia (commented):
Awesome work! Happy you found a way around without having to introduce a new package or add AspNetCore as a dep to Sentry.OTel!!

src/Sentry.AspNetCore/IUserFactory.cs
jamescrosswell: Done.

CHANGELOG.md
```diff
@@ -6,11 +6,16 @@
 
 - Sentry tracing middleware now gets configured automatically ([#2602](https://github.com/getsentry/sentry-dotnet/pull/2602))
 
+### Fixes
+
+- Identifying Users with OpenTelemetry doesn't work ([#2618](https://github.com/getsentry/sentry-dotnet/pull/2618))
```
bitsandfoxes, line 11:
```suggestion
- Resolved issue identifying users with OpenTelemetry ([#2618](https://github.com/getsentry/sentry-dotnet/pull/2618))
```

yudielcurbelo:
ISentryUserFactory is an internal interface, so how are we supposed to use it?

```csharp
internal interface ISentryUserFactory
{
    public User? Create();
}
```

```csharp
public class UserFactory : ISentryUserFactory
{
    public User Create()
    {
        ...
    }
}
```

image

jamescrosswell:
> ISentryUserFactory is an internal interface, so how are we supposed to use it?

Hey @yudielcurbelo, `ISentryUserFactory` is just an implementation detail that let's the SentrySdk access this information when using OpenTelemetry. Under the hood, all it does is let us get at the `HttpContext.User`:

https://github.com/getsentry/sentry-dotnet/blob/b833430b982b938fda2738a2e1ea44af85e5d036/src/Sentry.AspNetCore/DefaultUserFactory.cs#L24

So you shouldn't need to use this directly. You'd use whatever Authorization scheme you're using already and Sentry will be able to pull details for the user from the HttpContext.

ErlendSB:
How should we rewrite the example stated here: https://docs.sentry.io/platforms/dotnet/guides/aspnetcore/#capturing-the-affected-user

`public void ConfigureServices(IServiceCollection services)
{
    services.AddSingleton<IUserFactory, MyUserFactory>();
}`

jamescrosswell:
> How should we rewrite the example stated here: https://docs.sentry.io/platforms/dotnet/guides/aspnetcore/#capturing-the-affected-user
> 
> `public void ConfigureServices(IServiceCollection services) { services.AddSingleton<IUserFactory, MyUserFactory>(); }`

Ah, I see. You want to create your own User Factory that gets used by the OpenTelemetry integration? In that case you're right... we'd need to make`ISentryUserFactory` public.

Thanks for calling this out.

I've created a new ticket to track this: https://github.com/getsentry/sentry-dotnet/issues/2714

src/Sentry/ISentryUserFactory.cs
```diff
@@ -0,0 +1,6 @@
+namespace Sentry;
+
+internal interface ISentryUserFactory
```
jeremy-allocate, line 3: @jamescrosswell I'm hitting the IUserFactory warning in our builds, and I went to do the cuover to silence them, but this interface is marked internal so I can't implement it in my projects.  Was this unintentional?

jamescrosswell: Hey @jeremy-allocate , see https://github.com/getsentry/sentry-dotnet/pull/2618#issuecomment-1757367606 above. 

If it's an emergency and you just need to get a build out, you can disable the warning with a pragma, as we've done here:

https://github.com/getsentry/sentry-dotnet/blob/bbe4fa9d72b9c08e53b5bc33d07e2d52b2409b4f/src/Sentry.AspNetCore/Extensions/DependencyInjection/ServiceCollectionExtensions.cs#L26-L28

jeremy-allocate: All good, thanks @jamescrosswell.  We'll live with the warnings until #2714 is released.

jamescrosswell:
As an FYI, we just made [Release 3.40.1](https://github.com/getsentry/sentry-dotnet/releases/tag/3.40.1) that addresses this:
* ISentryUserFactory is now public so users can register their own implementations via DI (https://github.com/getsentry/sentry-dotnet/pull/2719)

jeremy-allocate:
Awesome, many thanks!

PR #2619, opened 2023-09-14
ref: SpanId creation

bitsandfoxes:
In an effort to improve the memory footprint: There is no technical reason to keep the Id value as `string`.
The spanId value of `0` will be treated as `empty`.

```
BenchmarkDotNet=v0.13.5, OS=macOS Ventura 13.4 (22F66) [Darwin 22.5.0]
Apple M1 Pro, 1 CPU, 10 logical and 10 physical cores

OLD:
|              Method |     Mean |    Error |  StdDev |   Gen0 | Allocated |
|-------------------- |---------:|---------:|--------:|-------:|----------:|
| 'Creates a Span ID' | 722.1 ns | 63.51 ns | 3.48 ns | 0.0687 |     144 B |

NEW:
|              Method |     Mean |     Error |    StdDev | Allocated |
|-------------------- |---------:|----------:|----------:|----------:|
| 'Creates a Span ID' | 7.924 ns | 0.3097 ns | 0.0170 ns |         - |
```

#skip-changelog

### DIFF
--- a/src/Sentry/SpanId.cs
+++ b/src/Sentry/SpanId.cs
@@ -1,4 +1,5 @@
 using Sentry.Extensibility;
+using Sentry.Internal;
 
 namespace Sentry;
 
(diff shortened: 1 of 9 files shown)


src/Sentry/SpanId.cs
```diff
@@ -7,58 +8,60 @@ namespace Sentry;
...
+            Random.NextBytes(buf);
+            random = BitConverter.ToInt64(buf, 0);
+        } while (random == 0);
+
+        return new SpanId(random);
+    }
 
     /// <inheritdoc />
-    public void WriteTo(Utf8JsonWriter writer, IDiagnosticLogger? _) => writer.WriteStringValue(GetNormalizedValue());
+    public void WriteTo(Utf8JsonWriter writer, IDiagnosticLogger? _) => writer.WriteStringValue(ToString());
```
bruno-garcia, line 64: here we're still allocating the string before writing, is there a better way? Like writing each of the 8 bytes directly to the json writer instead?

src/Sentry/SpanId.cs
```diff
@@ -7,58 +8,60 @@ namespace Sentry;
 /// </summary>
 public readonly struct SpanId : IEquatable<SpanId>, IJsonSerializable
 {
-    private readonly string _value;
+    private static readonly RandomValuesFactory Random = new SynchronizedRandomValuesFactory();
```
bruno-garcia, line 11: is this possibly going to create contention on a server scenario?

src/Sentry/SpanId.cs
```diff
@@ -7,58 +8,60 @@ namespace Sentry;
...
-    // This is obviously a bad idea as it invalidates GUID's uniqueness properties
-    // (https://devblogs.microsoft.com/oldnewthing/20080627-00/?p=21823)
-    // but all other SDKs do it this way, so we have no choice but to comply.
     /// <summary>
     /// Generates a new Sentry ID.
     /// </summary>
-    public static SpanId Create() => new(Guid.NewGuid().ToString("n")[..16]);
+    public static SpanId Create()
+    {
+        var buf = new byte[8];
```
bruno-garcia, line 51: Would this work/improve things?
```suggestion
        Span<byte> buf = stackalloc byte[8];
```

bitsandfoxes: It does! A lot! 🔥

src/Sentry/SpanId.cs
```diff
@@ -7,58 +8,82 @@ namespace Sentry;
...
-    // all fields initialized to default values.
-    // Effectively, using this method instead of just referencing `_value`
-    // makes the behavior more consistent, for example:
-    // default(SpanId).ToString() -> "0000000000000000"
-    // default(SpanId) == SpanId.Empty -> true
-    private string GetNormalizedValue() => !string.IsNullOrWhiteSpace(_value)
-        ? _value
-        : EmptyValue;
+    // TODO: mark as internal in version 4
+    public SpanId(string value) => long.TryParse(value, NumberStyles.HexNumber, null, out _value);
```
bruno-garcia, line 27: You could add an obsolete attrbiute to this

src/Sentry/SpanId.cs
```diff
@@ -7,58 +8,82 @@ namespace Sentry;
...
+    /// Creates a new instance of a Sentry span Id.
+    /// </summary>
+    /// <param name="value"></param>
+    public SpanId(long value) => _value = value;
 
     /// <inheritdoc />
-    public bool Equals(SpanId other) => StringComparer.Ordinal.Equals(
-        GetNormalizedValue(),
-        other.GetNormalizedValue());
+    public bool Equals(SpanId other) => GetValue() == other.GetValue();
```
bruno-garcia, line 36: `==` is an operator. Since u're on Equals here, stick to it
```suggestion
    public bool Equals(SpanId other) => GetValue().Equals(other.GetValue());
```

src/Sentry/SpanId.cs
```diff
@@ -91,20 +116,4 @@ public static SpanId FromJson(JsonElement json)
...
-    // transaction.StartChild(string operation, string description)
-    // The following code:
-    // transaction.StartChild("foo", "bar")
-    // Will resolve to the first method and not the second, which is incorrect.
-
-    /*
-    /// <summary>
-    /// A <see cref="SentryId"/> from a <see cref="Guid"/>.
-    /// </summary>
-    public static implicit operator SpanId(string value) => new(value);
```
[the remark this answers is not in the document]
bruno-garcia: `    public static implicit operator SpanId(string value) => new(value);` this isn't a comment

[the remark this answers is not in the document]
bruno-garcia: oh am i missing a `/*` ?

bruno-garcia (approved):

PR #2622, opened 2023-09-15
build with dotnet 8 sdk

bruno-garcia:
When trying to run tests on .NET 8 (https://github.com/getsentry/sentry-dotnet/pull/2617/) it was clear that it was too much of a jump. Mobile support issues came up (Android and iOS, different issues). .NET 6 Android support was completely removed from .NET 8, so we need to chose between the two: https://github.com/xamarin/xamarin-android/pull/7900

This PR attempts to do the minimum changes required to build the repo with the .NET 8 SDK

This can be reverted possibly (as fixed in https://github.com/xamarin/xamarin-macios/pull/18793)
* https://github.com/getsentry/sentry-dotnet/blob/2f11b3073e4169a60cc8143b8d5bd6016a3b826a/samples/Directory.Build.props#L38

## Depends on:
* [x] https://github.com/getsentry/sentry-dotnet/pull/2624
* ~https://github.com/dotnet/sdk/issues/34496#issuecomment-1721875830~
* [x] https://github.com/getsentry/sentry-dotnet/issues/2639
* [x] https://github.com/xamarin/xamarin-macios/pull/18793 (should be part of .NET 8 RC2)

Since there's no customer-facing change:

#skip-changelog

PR #2625, opened 2023-09-16
macOS native SDK integration

bruno-garcia:
An experiment to introduce specific mac target and bundle the native SDK (cocoa or sentry-native)

TODO: 
1. Build `sentry-cocoa` for desktop and remove the bundled dylib
2. Compile the bridge code as part of CI
3. Make sure native symbol upload include the native bits needed
4. Portable PDB upload works: https://github.com/getsentry/sentry-cli/issues/1746
5. test with NativeAOT (on a Mac, requires .NET 8 https://github.com/getsentry/sentry-dotnet/pull/2622)

First successful crash captured natively: https://sentry-sdks.sentry.io/issues/4483237642/?project=5428537&query=is%3Aunresolved&referrer=issue-stream&statsPeriod=90d&stream_index=0


Directory.Build.props
```diff
@@ -18,12 +18,14 @@
   </PropertyGroup>
 
   <!-- NO_MOBILE can be passed in via an environment variable or a build property to disable all mobile targets -->
-  <PropertyGroup Condition="'$(NO_MOBILE)' == 'true'">
```
jamescrosswell, line 21: We probably don't want to check this particular change into source control

Directory.Build.props
```diff
@@ -18,12 +18,14 @@
...
+  <PropertyGroup Condition="'$(NO_MOBILE)' == ''">
```
jamescrosswell, line 21: We probably don't want this checked into source control

lib/sentry-native/linux/Sentry/libsentry.dbg.so
jamescrosswell, line 1: Is there a way to avoid having compiled code in source control?

bruno-garcia: yeah we'll need to build this, and order CI in a way where the final/packaging job gets the artifacts from previous jobs. We did that on the Unity repo.

Directory.Build.props
bruno-garcia: We might need a second `sln` after all. Unles we find another way to avoid this.

Can we have a symlink? `Sentry.NoMobile.sln` is a symlink to `Sentry.sln`? So you can add the checks back on the msbuild files based on sln file name?

vaind:
I've originally wanted to update this PR and make the fixes but it's not possible to build macos native AOT apps at this time (before .net 8 RC2), so I'm going to fork this further and only do Windows+sentry-native in the first step.

vaind:
Although #2767 brings native AOT debug images by using sentry-native for macOS (because it's much easier then getting cocoa working and bundled properly), I'll leave this open for the overall approach taken with all the scope sync, etc. We may revisit this later when there are features only sentry-cocoa can fulfill.

bruno-garcia:
Replaced by #2770

PR #2633, opened 2023-09-19
Memory optimisation for GetLastActiveSpan

jamescrosswell:
#skip-changelog

# Overview

When load testing an ASP.NET Core application that creates a lot of spans, `GetLastActiveSpan` was responsible for a significant proportion of memory allocation:

```csharp
                    // The endpoint being load tested
                    endpoints.MapGet("/spans", async context =>
                    {
                        for (int i = 0; i < 1000; i++)
                        {
                            var span = SentrySdk.GetSpan()?.StartChild($"child {i}");
                            span?.Finish();
                        }
                        await context.Response.WriteAsync("Hello Spans!");
                    });

```

image

This appears to be due to the call to `Spans.OrderByDescending`:
https://github.com/getsentry/sentry-dotnet/blob/bbe4fa9d72b9c08e53b5bc33d07e2d52b2409b4f/src/Sentry/TransactionTracer.cs#L381 

## Iteration 1

By replacing this with a for loop, memory consumption in our benchmark dropped by **65.68%** from 2109.85 KB to 724.02 KB

### Before 

|                          Method | SpanCount |        Mean |       Error |     StdDev |     Gen0 |    Gen1 |  Allocated |
|-------------------------------- |---------- |------------:|------------:|-----------:|---------:|--------:|-----------:|
| 'Create spans for scope access' |         1 |    45.39 us |    13.71 us |   0.752 us |   5.6152 |  1.7700 |   19.77 KB |
| 'Create spans for scope access' |        10 |   205.46 us |   276.93 us |  15.179 us |  23.9258 |  4.3945 |   85.51 KB |
| 'Create spans for scope access' |       100 | 4,055.16 us | 5,006.05 us | 274.399 us | 601.5625 | 85.9375 | 2109.85 KB |

### After

|                          Method | SpanCount |        Mean |       Error |    StdDev |     Gen0 |    Gen1 | Allocated |
|-------------------------------- |---------- |------------:|------------:|----------:|---------:|--------:|----------:|
| 'Create spans for scope access' |         1 |    52.05 us |   242.80 us | 13.309 us |   5.2490 |  1.6479 |  18.92 KB |
| 'Create spans for scope access' |        10 |   164.61 us |    55.00 us |  3.015 us |  12.2070 |  4.1504 |  63.88 KB |
| 'Create spans for scope access' |       100 | 2,287.86 us | 1,098.36 us | 60.205 us | 183.5938 | 50.7813 | 724.02 KB |

## Iteration 2 

By instead keeping all the spans on a Stack and inspecting the most recent one(s) to return the last active span when required we were able to reduce memory usage for the 100 spans scenario drops by a further **63.25%** from 724.02 KB to 266.09 KB. So about an **~87%** reduction vs the original benchmarks. 

### After

|                          Method | SpanCount |      Mean |     Error |   StdDev |    Gen0 |    Gen1 | Allocated |
|-------------------------------- |---------- |----------:|----------:|---------:|--------:|--------:|----------:|
| 'Create spans for scope access' |         1 |  28.45 us | 13.270 us | 0.727 us |  4.6387 |  1.5869 |  16.64 KB |
| 'Create spans for scope access' |        10 |  84.98 us |  6.361 us | 0.349 us |  9.3994 |  2.5635 |  39.17 KB |
| 'Create spans for scope access' |       100 | 657.54 us | 54.581 us | 2.992 us | 57.6172 | 16.6016 | 266.09 KB |


benchmarks/Sentry.Benchmarks/LastActiveSpanBenchmarks.cs
```diff
@@ -0,0 +1,54 @@
...
+        var transaction = SentrySdk.StartTransaction(Name, Operation);
+        SentrySdk.WithScope(scope =>
+        {
+            scope.Transaction = transaction;
+            for (var i = 0; i < SpanCount; i++)
+            {
+                // Simulates a scenario where TransactionTracer.GetLastActiveSpan will be called frequently
+                // See: https://github.com/getsentry/sentry-dotnet/blob/c2a31b4ead03da388c2db7fe07f290354aa51b9d/src/Sentry/Scope.cs#L567C1-L567C68
+                var span = transaction.StartChild(Operation);
+                if (rnd.Next(2) < 1)
```
bruno-garcia, line 35: Ideally we don't use random in the benchmark. So we can get consistent values for each run

src/Sentry/TransactionTracer.cs
```diff
@@ -313,9 +313,16 @@ private void AddChildSpan(SpanTracer span)
         if (!isOutOfLimit)
         {
             _spans.Add(span);
+            _sortedSpanCache = null; // invalidate cache
```
bruno-garcia, line 316: I believe without a memory barrier this won't be visible to all threads at once. the `Interlocked` API can solve that.

Or mark the field as `volatile`

src/Sentry/TransactionTracer.cs
```diff
@@ -313,9 +313,16 @@ private void AddChildSpan(SpanTracer span)
...
         }
     }
 
+    private IOrderedEnumerable<ISpan>? _sortedSpanCache;
+    /// <inheritdoc />
+    public ISpan? GetLastActiveSpan() =>
+        // We need to sort by timestamp because the order of ConcurrentBag<T> is not deterministic
+        (_sortedSpanCache ??= Spans.OrderByDescending(x => x.StartTimestamp)).FirstOrDefault(s => !s.IsFinished);
```
bruno-garcia, line 324: Can we replace these LINQ usages with loops? That'll avoid allocations too

bruno-garcia: Also: Isnt' TransactionTrace supposed to be thread safe? In a GlobalMode=true scenario, multiple threads can be adding spans. So iterating on this collection will throw at runtime.

benchmarks/Sentry.Benchmarks/LastActiveSpanBenchmarks.cs
jamescrosswell: Good call. I got rid of the "sometimes test `GetLastActiveSpan` and sometimes don't" logic from the benchmark entirely... The impact of this optimization will depend on how frequently customers' applications call `GetLastActiveSpan` but we can't really know (or similate) the likelihood of that in a benchmark... so currently our benchmarks give us something clsoe to the maximum possible performance gain a customer could get, if their code was structured exactly like our benchmark (pretty slim chance of that).

src/Sentry/TransactionTracer.cs
jamescrosswell: Cool, that answers the question I raised in Discord. Thanks Bruno!

src/Sentry/TransactionTracer.cs
jamescrosswell: Replacing it with a for loop gave even better results... sometimes I guess it's good to go back to basics!

jamescrosswell:
# Important Note

I merged main into this branch and annoyingly it seems to contain a bunch of the changes from https://github.com/getsentry/sentry-dotnet/pull/2637... I'm not sure why but it makes this PR really hard to review. 

As such, I created a new branch and PR for these changes... will close this PR in favour of that branch. 

See https://github.com/getsentry/sentry-dotnet/pull/2642 instead.

PR #2646, opened 2023-09-22
feat: Public headers utility

bitsandfoxes:
With the `ContinueTrace` now accepting the headers as `string` I think we should also provide the keys to them.
I added a public `TryParse` to the Baggage because
1. SentryTraceHeader has a public `TryParse`
2. `ContinueTrace` accepts a Baggage but there is no way for users to create a Baggage object.
Since the parsing part is not supposed to be a general tool but only deal with Sentry it defaults to `onlySentry = true`.

How it's supposed to be used:
```
// Grab the trace and baggage header i.e. from the context's request
var traceHeader = context.Request.Headers.GetValueOrDefault(SentryTraceHeader.HttpHeaderName);
var baggage = context.Request.Headers.GetValueOrDefault(BaggageHeader.HttpHeaderName);

var transactionContext = SentrySdk.ContinueTrace(sentryTraceHeader, baggageHeader);
var transaction = Sentry.startTransaction(transactionContext);
```

### DIFF
--- a/src/Sentry/BaggageHeader.cs
+++ b/src/Sentry/BaggageHeader.cs
@@ -10,7 +10,11 @@ namespace Sentry;
 /// <seealso href="https://www.w3.org/TR/baggage"/>
 public class BaggageHeader
 {
-    internal const string HttpHeaderName = "baggage";
+    /// <summary>
+    /// The HTTP header name for baggage
+    /// </summary>
+    public const string HttpHeaderName = "baggage";
+
     internal const string SentryKeyPrefix = "sentry-";
 
     internal static IDiagnosticLogger? Logger { get; set; } = SentrySdk.CurrentOptions?.DiagnosticLogger;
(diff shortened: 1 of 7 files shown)


bruno-garcia:
> ```
> // Grab the trace and baggage header i.e. from the context's request
> var traceHeader = context.Request.Headers.GetValueOrDefault(SentryTraceHeader.HttpHeaderName);
> var baggage = context.Request.Headers.GetValueOrDefault(BaggageHeader.HttpHeaderName);
>
> var transactionContext = SentrySdk.ContinueTrace(sentryTraceHeader, baggageHeader);
> var transaction = Sentry.startTransaction(transactionContext);
> ```

Alternatively the API could be:

```
var transactionContext = context.Request.Headers.GetSentryTransactionContext();

var transaction = SentrySdk.StartTransaction(transactionContext);
```

Where `GetSentryTransactionContext` is an extension method that does the two first steps of the originally proposed API.

bitsandfoxes:
Doesn't make much sense to me anymore.

PR #2648, opened 2023-09-25
Added http.request.method and http.response.status_code to span data/extras

jamescrosswell:
#skip-changelog

## Overview

This is to implement https://github.com/getsentry/sentry-dotnet/issues/2606 and https://github.com/getsentry/sentry-dotnet/issues/2471

The requirements are a little hazy. I believe the goal is to start storing various things in Spans that were prevoiusly stored in Transactions, to pave the way for eventually deprecating Transactions. The goal of this particular ticket is to store the `http.request.method` and `http.response.status_code` on Spans (specifically, in the Span.Data attributes - currently named Span.Extra in the dotnet SDK although we'll be renaming those to be consistent with other SDKs when we release v4.0.0). 

There are two scenarios in which we'd want to capture the http request method and response status on spans:

1. Outboud Http Requests
2. Inbound Http Requests

### Outbound Requests

Of the two scenarios, outbound requests are the most straight forward. These are currently implemented in two concrete `SentryMessageHandler` classess: `SentryHttpMessageHandler` and `SentryGraphQLHttpMessageHandler`. Both of these message handlers have a `ProcessRequest` method (where a `Span` representing the outbound HTTP request gets created) and a `HandleResponse` method, where the Span has additional information added to it and is finally finished. 

About the only design decision is when to add the information. Technically we could set the `http.request.method` Span.Data attribute already in the `ProcessRequest` method before the outbound HTTP Request is made... since we already know at that point what kind of request is going to be made. 

On the other hand, we won't know the `http.response.status_code` until after the HTTP Request has completed, in the `HandleResponse` method. 

We've opted to make this data available on the span as soon as possible, so the request method is being set in the `ProcessRequest` method and the response status code is being captured in the `HandleResponse` method. This also matches what happens for Inbound Requests (described below). 

### Inbound Requests

Inbound requests are slightly more complicated and this is where the requirements are a bit hazy. The most common sceanrio for inbound requests will be for Web or API servers running ASP.NET Core or the older ASP.NET framework equivalent.

In this scenario, there is no obvious span where we would/should add these attributes. The most logical span would be the one that wraps the whole HTTP Request - however in our ASP.NET and ASP.NET Core integrations that span is a Transaction... and it already holds both the request method and the response status code in the Request and Response interfaces. 

That gives us various different possible ways we could implement this:
1. We duplicate this information in the Transaction Span that we already have... preparing for a future  where this Transaction becomes a plain on Span (presumably the Request/Response interfaces get dropped at that point in favour of the more generic Span.Data attributes). 
2. We add a new "special" child Span inside the Transaction, which will basically be a duplicate fo the Transaction but only holding the subset of data from the Transaction that has already been migrated to Spans
3. We arbitrarily pick the first Span underneath the transaction and apply the new Span.Data attributes there
4. We apply the new Span.Data attributes to all the child spans of the Transaction

No. 3 and 4 feel like weird solutions. Depending on what the application is doing, we'd end up with these attributes on random spans that may represent something completely different. 

No. 2 could work, but it would result in some extraneous new spans being introduced to Sentry.io that:
a) Might be a bit confusing for users, as these spans will feel like (and will actually be) duplicates of the Transaction spans that already exist
b) Won't have much meat on them, at least initially 

As such, the first solution above seems like the best approach. It means we won't be adding any additional spans that we're not already capturing so will probably be the smoothest transition from Transactions to Spans for users. About the only down side is that it will mean there will be some duplication in Transactions for a while... The Request Method will be in both the Request interface on the Transaction and stored as an attribute in the Transaction.Data... same again for the Http Response Status Code.


jamescrosswell:
@bitsandfoxes worth noting that this information is duplicated in the `transaction.method` attribute (see screenshot). I assume we will want to deprecate this at some stage?

src/Sentry/SentryHttpMessageHandler.cs
```diff
@@ -85,7 +86,12 @@ protected internal override void HandleResponse(HttpResponseMessage response, IS
         _failedRequestHandler?.HandleResponse(response);
 
         // This will handle unsuccessful status codes as well
-        var status = SpanStatusConverter.FromHttpStatusCode(response.StatusCode);
-        span?.Finish(status);
+        if (span is not null)
+        {
+            span.SetExtra(OtelSemanticConventions.AttributeHttpRequestMethod, method);
+            span.SetExtra(OtelSemanticConventions.AttributeHttpResponseStatusCode, (int)response.StatusCode);
```
jamescrosswell, line 92: Note that I thought about putting this on the scope. In a scenario where the HttpClient is inside an ASP.NET Core Endpoint though, the Request.Method and Response.StatusCode already refer to the method and status code of the HttpRequest to the webserver. 

Another option might be to push a scope when the span gets created in `SentryMessageHandler.SendAsync`, set the Request and Response info on the Scope as context there, and then pop the scope once the span finishes. 

Maybe that's a cleaner solution? It would mean we could consolidate the logic to set these attributes for outbound requests in the `SentryMessageHandler` as well (rather than having it duplicated, as it currently is, across `SentryHttpMessageHandler` and `SentryGraphQLHttpMessageHandler`).

src/Sentry/Internal/Hub.cs
```diff
@@ -486,6 +486,10 @@ public void CaptureTransaction(Transaction transaction, Hint? hint)
             var scope = currentScopeAndClient.Key;
             scope.Evaluate();
             scope.Apply(transaction);
+            foreach (var span in transaction.Spans)
+            {
+                scope.ApplyToSpan(span);
+            }
```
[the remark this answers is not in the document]
jamescrosswell: Aha... that wasn't my intention. I had meant to apply the scope to all of the spans (not just the immediate children of the Transaction). I'll take another look.

src/Sentry/SentryHttpMessageHandler.cs
jamescrosswell: Yeah agreed, I'll move it.

jamescrosswell: Ah, actually... this is one reason it can't be done when the span gets created:
```
span.SetExtra(OtelSemanticConventions.AttributeHttpResponseStatusCode, (int)response.StatusCode);
```

We don't have the response code until later:

https://github.com/getsentry/sentry-dotnet/blob/b5ab31abfe4e9d0a891d260edb91faf03fecc922/src/Sentry/SentryMessageHandler.cs#L88-L99

I'll move it into the `SentryMessageHandler` though. If the logic ever diverges, between the different concrete message handlers, we can always push it back out to the concrete classes.

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Features
+
+- Added attribute http.request.method to span data ([#2648](https://github.com/getsentry/sentry-dotnet/pull/2648))
```
bruno-garcia, line 7: Why does this matter? Thinking as a user, reading this: Why would I want to update the SDK?
I believe this will show up in Sentry on the new performance view? perhaps poiniting that out will motivate ppl to upgrade

src/Sentry/Scope.cs
```diff
@@ -382,6 +384,19 @@ public void ClearBreadcrumbs()
...
+    internal void ApplyToSpan(ISpanData spanData)
+    {
+        if (Request.Method is {} method)
+        {
+            spanData.SetExtra(OtelSemanticConventions.AttributeHttpRequestMethod, method);
+        }
+
+        if (Contexts.Response.StatusCode is {} statusCode)
+        {
+            spanData.SetExtra(OtelSemanticConventions.AttributeHttpResponseStatusCode, statusCode);
```
jamescrosswell, line 396: This will currently it will overwrite these values, if they've already been set (e.g. by a `SentryMessageHandler`). Need to add a TryAdd method to IHasExtra and use this instead.

src/Sentry/SentryHttpMessageHandler.cs
jamescrosswell: Actually, these things are already a bit different... I think better to just leave them separate.

This is what the SentryHttpMessageHandler is doing:
https://github.com/getsentry/sentry-dotnet/blob/b5ab31abfe4e9d0a891d260edb91faf03fecc922/src/Sentry/SentryHttpMessageHandler.cs#L91-L94

And this is what the SentryGraphQLMessageHandler is doing:
https://github.com/getsentry/sentry-dotnet/blob/b5ab31abfe4e9d0a891d260edb91faf03fecc922/src/Sentry/SentryGraphQLHttpMessageHandler.cs#L91-L96

Not enough in common, I don't think, to justify refactoring to a common base method.

CHANGELOG.md
jamescrosswell: I've updated the description for this PR to clarify what we're doing here and why. 

I don't think users will care about the this change until we phase out Transactions, to be honest. We're not providing any information that they didn't already have. This is really just an internal change to prepare us for the move to Spans. 

So I've put #skip-changelog on the description for the time being.

src/Sentry/Internal/Hub.cs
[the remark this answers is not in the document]
jamescrosswell: OK, I've removed this entirely... opting instead to capture on the span directly in the SentryTracingMiddleware.

bitsandfoxes (approved):
LGTM!

PR #2651, opened 2023-09-25
Removed legacy group

jamescrosswell:
#skip-changelog

Removed `legacy` group from the solution filters as both the `Mobile` and `NoMobile` `slnf` files (which use this) should include the `Sentry.AspNet.csproj` file.


bitsandfoxes (approved):
Nice!

PR #2659, opened 2023-09-26
Merged IHasMeasurements into ISpanData

jamescrosswell:


src/Sentry/ITransactionData.cs
```diff
@@ -1,8 +1,59 @@
...
+    /// The measurements that have been set on the transaction.
+    /// </summary>
+    IReadOnlyDictionary<string, Measurement> Measurements { get; }
+
+    /// <summary>
+    /// Sets a measurement on the transaction.
+    /// </summary>
+    /// <param name="name">The name of the measurement.</param>
+    /// <param name="measurement">The measurement.</param>
+    [EditorBrowsable(EditorBrowsableState.Never)]
```
bruno-garcia, line 20: Why do we want this not to be visible in the IDE?

bruno-garcia (commented):
Given we're adding this soon to the SDK: https://github.com/getsentry/sentry-python/blob/master/sentry_sdk/metrics.py

Do we gain from keeping things split from transactions? We'll be sending Spans only soon too (outside transactions).

jamescrosswell:
> Given we're adding this soon to the SDK: https://github.com/getsentry/sentry-python/blob/master/sentry_sdk/metrics.py
> 
> Do we gain from keeping things split from transactions? We'll be sending Spans only soon too (outside transactions).

So put it on ISpanData instead already?

src/Sentry/ITransactionData.cs
jamescrosswell: I think that was copied across verbatim. I'll get rid of it.

src/Sentry/SpanTracer.cs
```diff
@@ -30,6 +31,16 @@ public class SpanTracer : ISpan
     /// <inheritdoc />
     public bool IsFinished => EndTimestamp is not null;
 
+    // Not readonly because of deserialization
+    private readonly Dictionary<string, Measurement> _measurements = new();
+
+    /// <inheritdoc />
+    public IReadOnlyDictionary<string, Measurement> Measurements => _measurements;
```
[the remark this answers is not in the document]
jamescrosswell: I can do it easily enough in the `SpanTracer`. 

However, the `Span` and `Transaction` classes assign the measurements via a constructor... which makes the use of the Lazy class itself problematic. It would be possible to wire something up manually but not sure if this is worth doing since Span/Transaction are generally only ever created just before serializing the transaction and sending this over the wire. I guess it might save some allocations for events that get sampled out?

src/Sentry/SpanTracer.cs
```diff
@@ -30,6 +31,16 @@ public class SpanTracer : ISpan
     /// <inheritdoc />
     public bool IsFinished => EndTimestamp is not null;
 
+    // Not readonly because of deserialization
+    private readonly Lazy<Dictionary<string, Measurement>> _measurements = new();
```
bruno-garcia, line 35: when are we using `Lazy` vs `==?` ? what's the criteria?

src/Sentry/Span.cs
```diff
@@ -88,6 +100,7 @@ public Span(ISpan tracer)
         Status = tracer.Status;
         IsSampled = tracer.IsSampled;
         _extra = tracer.Extra.ToDictionary();
+        _measurements = tracer.Measurements.ToDictionary();
```
bitsandfoxes, line 103:
```suggestion
        InternalMeasurements = tracer.Measurements?.ToDictionary();
```
So then we can keep it `null` if there is nothing to convert over.

src/Sentry/SpanTracer.cs
jamescrosswell: In SpanTracer I used it because it was convenient... measurements only ever gets initialised from one place and it doesn't get assigned any items when it's being initialized (as above). 

In `Span` and `Transaction` it was a bit trickier because there are two different constructors. One of them doesn't include measurements and so could potentially be initialised as above. The other has measurements passed in:
https://github.com/getsentry/sentry-dotnet/blob/feat/4.0.0-remove-ihasmeasurements/src/Sentry/Span.cs#L103

It was messy to try to use `Lazy<T>` here - it seemed easier to use `??=`. I guess it could be done by passing a function to the Lazy<T> constructor instead:

```
_measurements = new Lazy<Dictionary<string, Measurement>>(() => tracer.Measurements.ToDictionary());
```

Maybe that's cleaner in hindsight.

jamescrosswell: OK, I've just gone with Lazy<T> everywhere. I think I was just too lazy to do this in the first place 😜

src/Sentry/Span.cs
```diff
@@ -28,6 +29,17 @@ public class Span : ISpanData, IJsonSerializable
...
+    // Not readonly because of deserialization
+    private Dictionary<string, Measurement>? _measurements;
+    private Dictionary<string, Measurement> LazyMeasurements => _measurements ??= new();
+
+    /// <inheritdoc />
+    public IReadOnlyDictionary<string, Measurement> Measurements => LazyMeasurements;
+
+    /// <inheritdoc />
+    public void SetMeasurement(string name, Measurement measurement) =>
+        LazyMeasurements[name] = measurement;
```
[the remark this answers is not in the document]
jamescrosswell: I ended up just going with Lazy<T> everywhere...

bitsandfoxes (approved):

PR #2668, opened 2023-09-26
Merged ISpanContext into ITraceContext

jamescrosswell:
This interface contained no properties or methods... and only impleneted `ITraceContext` so is entirelyreplacable by `ITraceContext`. 

### Migration Notes

Methods and properties now available on `ITraceContext` (although likely soon to be merged into `ISpanData` - so we'll probably tell users to look for the methods and properties on `ISpanData` instead).


CHANGELOG.md
```diff
@@ -14,6 +14,10 @@ without native/platform specific bindings and SDKs. See [this ticket for more de
 
 - Drop .NET 6 Mobile in favor of .NET 7 ([#2624](https://github.com/getsentry/sentry-dotnet/pull/2604))
 
+API Changes:
+
+- Removed ISpanContext ([#2668](https://github.com/getsentry/sentry-dotnet/pull/2668))
```
bruno-garcia, line 19: Does this affect the end user in any way? Doesn't seem so, but worth being clear here.

At the end of the day we'll [need a **migration guide**](https://docs.sentry.io/platforms/dotnet/migration/) so users know what steps users need to take in order to upgrade the SDK. So when we say a type was removed, they might expect: "Ok what do I need to do to replace it?". In this case it's possibly just "This type had no documented use and if somehow you used it, it can be replaced with `ITraceContext`" or something like that?

bruno-garcia (approved):
Just a note on release notes/migration guide.

CHANGELOG.md
jamescrosswell: OK, I'll try to paraphrase that in the changelog for now... and put a bit more detail in the PR notes that we can use when preparing the migration guide proper.

PR #2669, opened 2023-09-26
Feat/4.0.0 remove IHasBreadcrumbs

jamescrosswell:
Moved `IHasBreadcrumbs` properties and methods to `IEventLike` as every class that implemented one also implemented the other.


jamescrosswell:
Dang... started with the wrong base branch.

PR #2672, opened 2023-09-26
fix: Replacing `Carthage` with `xcodebuild`

bitsandfoxes:
It's a bit messy but it puts the SDK back in a buildable state. Removed the submodule and replaced `Carthage` with building the individual targets via `xcodebuild`.
Because the actual binding generation is relying on it I kept the output Carthage specific for now.

#skip-changelog


scripts/build-sentry-cocoa.sh
```diff
@@ -1,77 +1,67 @@
...
     fi
 
     # Remove anything we don't want to bundle in the nuget package.
     find Carthage/Build* \( -name Headers -o -name PrivateHeaders -o -name Modules \) -exec rm -rf {} +
-
-    rollbackSchemes
 fi
 
-popd > /dev/null
+popd > /dev/null
```
bruno-garcia, line 67:
```suggestion
popd > /dev/null

```

bruno-garcia (approved):
:shipit:

jamescrosswell:
Awesome - thanks @bitsandfoxes ! 🕺🏻

PR #2675, opened 2023-09-27
IHasTags and IHasExtra replaced with ITagsAndData

jamescrosswell:
`IHasTags` and `IHasExtra` can be merged as anything that implements one also implements the other. 
SDK users should use `ITagsAndData` instead.


bitsandfoxes:
That one seems a bit odd with those two small and distinct things now bundled up in one interface.
The interfaces get used in `IEventLike` and `ISpanData` but actually accessed very differently. I.e.`ISpanData` does not get passed through the Enricher to have the defaults set but has them accessible for users.
Instead of merging them, can we get rid of them?

jamescrosswell:
@bitsandfoxes yes, putting an `And` in the name of any class/interface is a bit of a smell... it usually indicates you have two unrealted things. I think in this case, we probably need to think about why these things might be sitting together on the same interface and what the name for that "single" purpose is. I opened a related issue: https://github.com/getsentry/sentry-dotnet/issues/2676 but maybe we think about renaming these immediately (before merging this pull request). Essentially I'm suggesting a better name for `ITagsAndData` might be `IScopeData`... but that depends whether I'm on the same wavelength as everyone else wrt `IEventLike` and how the shift from Transactions to Spans will play out.

bruno-garcia (changes requested):
doesn't seem worth the change but seems like you and Stefan also got to that conclusion

PR #2682, opened 2023-09-28
ref: InternalTags autoproperty

bitsandfoxes:
#skip-changelog

### DIFF
--- a/src/Sentry/SpanTracer.cs
+++ b/src/Sentry/SpanTracer.cs
@@ -50,20 +50,18 @@ public class SpanTracer : ISpan
     /// <inheritdoc />
     public bool? IsSampled { get; internal set; }
 
-    private ConcurrentDictionary<string, string>? _tags;
-
-    internal ConcurrentDictionary<string, string>? InternalTags => _tags;
+    internal ConcurrentDictionary<string, string>? InternalTags { get; private set; }
 
     /// <inheritdoc />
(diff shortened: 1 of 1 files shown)


src/Sentry/SpanTracer.cs
```diff
@@ -50,20 +50,18 @@ public class SpanTracer : ISpan
...
     public bool? IsSampled { get; internal set; }
 
-    private ConcurrentDictionary<string, string>? _tags;
-
-    internal ConcurrentDictionary<string, string>? InternalTags => _tags;
+    internal ConcurrentDictionary<string, string>? InternalTags { get; private set; }
 
     /// <inheritdoc />
-    public IReadOnlyDictionary<string, string> Tags => _tags ??= new ConcurrentDictionary<string, string>();
+    public IReadOnlyDictionary<string, string> Tags => InternalTags ??= new ConcurrentDictionary<string, string>();
```
jamescrosswell, line 56: So I guess a pretty similar discussion to the one we just had about Measurements. Any reason not to use `Lazy<T>` here instead of implementing it ourselves?

bitsandfoxes: `Lazy<T>` is used for delayed initialization of resource-intensive or large things. `Tags` are neither. Also, `Lazy<T>` would indicate that we will definitely need the property, just not right now. Which we don't. It's perfectly fine for the tags to remain `null` for the whole duration of the object's lifetime.

jamescrosswell (approved):

Issue #2644, opened 2023-09-20
HTTP Method column missing as OTEL attribute doesn't appear to parse


JLuse:
### Package

Sentry

### .NET Flavor

.NET

### .NET Version

7.0.11

### OS

Android

### SDK Version

3.39.1

### Steps to Reproduce

**Description**
Reported from a customer integrating OTEL, that when viewing their performance page, the HTTP Method column contains all empty values. However, the OTEL attribute associated with the event has the method included:
<img width="672" alt="Screenshot 2023-09-20 at 2 55 26 PM" src="https://github.com/getsentry/sentry-dotnet/assets/8451987/18a7dc93-148e-4c22-af38-f0ebc309961a">

Customer support ticket: https://sentry.zendesk.com/agent/tickets/103357
Steps to reproduce:
1. Instrument OTEL - https://docs.sentry.io/platforms/dotnet/performance/instrumentation/opentelemetry/
2. Send transaction -> view in Sentry project

### Expected Result

The span processor should adjust by setting the request context on the event, populating the HTTP methods

### Actual Result

Column is empty:
<img width="576" alt="Screenshot 2023-09-20 at 3 13 13 PM" src="https://github.com/getsentry/sentry-dotnet/assets/8451987/b3963311-0930-4ac9-954a-64d15ed21662">

jamescrosswell:
Possibly related to https://github.com/getsentry/sentry-dotnet/issues/2606

jamescrosswell:
Thanks @JLuse , I can reproduce this so we should be able to get to the bottom of it. 

For some reason, the transaction.Request context isn't being set correctly. 

Normally our Middleware hooks things up so the Scope gets populated when it's evaluated:
https://github.com/getsentry/sentry-dotnet/blob/55ca61eb90b7a269b950d95500502e154f16787d/src/Sentry.AspNetCore/SentryMiddleware.cs#L118-L122

And our Hub evalulates and applies this when capturing a transaction:
https://github.com/getsentry/sentry-dotnet/blob/0ae99b14d8bf90a994a64c114f8c1e93c0d0417b/src/Sentry/Internal/Hub.cs#L484-L488

This isn't working. `Hub.CatpureTransaction` gets called but, the scope that it retrieves from the ScopeManager doesn't have any `Scope.OnEvaluating` callback configured... which makes me think it's not the same scope that was configured by our Middleware. 

Needs some investigation.

jamescrosswell:
[...]

PR #2690, opened 2023-10-03
Fix: Scope not being applied to OpenTelemetry spans in ASP.NET Core

jamescrosswell:
# Problem

Fixes https://github.com/getsentry/sentry-dotnet/issues/2644... however `Transaction.Request.Method` was just one of a number of attributes missing that would ordinarily be applied to Transactions from the Scope:

https://github.com/getsentry/sentry-dotnet/blob/0ae99b14d8bf90a994a64c114f8c1e93c0d0417b/src/Sentry/Internal/Hub.cs#L484-L488

### SentryMiddleware and SentrySpanProcessor don't share a ScopeAndClientStack

Part of the problem is that OpenTelemetry spans get created as a result of System.Diagnostic events that Otel subscribes to... and those events are received/processed in a different AsyncLocal context to the rest of our middleware pipeline. As such, our `SentryMiddleware` and our `SentrySpanProcessor` don't share the same ScopeAndClientStack:

https://github.com/getsentry/sentry-dotnet/blob/efb5dc8acea1450de9fe1dd8561ad31c4ef069e4/src/Sentry/Internal/SentryScopeManager.cs#L24-L27

### Scope popped before SentrySpanProcessor.OnEnd is called

Initially I tried to work around the problem above by using a `ScopeAndClientStack` that was bound to `Activity.Current.RootId` rather than to the current `AsyncLocal` context... and that was possible. However it revealed another problem, which is that by the time `SentrySpanProcessor.OnEnd` is called, the Scope that was pushed by our middleware had already been popped from the `ScopeAndClientStack`... The fact that Otel is using a System.Diagnostics activity event subscription means that processing of these events happens async (kind of like events on a queue).

# Solution

The second of those two problems forced me to come up with a fairly simple solution in the end.. which is to:
1. Save a clone of the scope from the SentryMiddleware, once the request has finished
2. Restore that saved scope to the ScopeAndClientStack inside the SentrySpanProcessor prior to finishing and capturing the transaction

https://github.com/getsentry/sentry-dotnet/blob/d189a78c2e5eb5a5ccab15a205c8cbd96b7c21ef/src/Sentry.AspNetCore/SentryMiddleware.cs#L136-L141

https://github.com/getsentry/sentry-dotnet/blob/2e4e4b50336fdee1e97a26be10b07eddf2ce8d85/src/Sentry.OpenTelemetry/SentrySpanProcessor.cs#L161-L165


bitsandfoxes:
Nice findings! Thanks for digging through that!
If I understand this correctly the `RestoreScope` basically pushes a "too soon popped" scope back on the stack so the whole "apply scope to transaction" logic down the line works with `GetCurrentScope`?
But does that mean that the whole stack gets copied with every request? 
You've already found a way to link the activity with the scope. Can we reuse the OtelEnricher or a transaction processor for applying the scope instead?

src/Sentry/Internal/SentryScopeManager.cs
```diff
@@ -91,40 +87,52 @@ public IDisposable PushScope<TState>(TState? state)
         ScopeAndClientStack = newScopeAndClientStack;
         return scopeSnapshot;
     }
+    public void RestoreScope(Scope savedScope)
```
bruno-garcia, line 90:
```suggestion
    public void RestoreScope(Scope savedScope)
```

src/Sentry/Internal/SentryScopeManager.cs
```diff
@@ -91,40 +87,52 @@ public IDisposable PushScope<TState>(TState? state)
...
+    {
+        var currentScopeAndClientStack = ScopeAndClientStack;
```
bruno-garcia, line 92: This method isn't safe when IsGlobalMode=true. It doesn't seem like it would be called the way things are right now, but best to guard it in here to avoid surprises

Similar to this: https://github.com/getsentry/sentry-dotnet/blob/407dc28dbd7f1c29059fc509a886b114a2c75b0b/src/Sentry/Internal/SentryScopeManager.cs#L55-L59

src/Sentry/SentrySdk.cs
```diff
@@ -240,6 +240,17 @@ public void Dispose()
     [DebuggerStepThrough]
     public static IDisposable PushScope<TState>(TState state) => CurrentHub.PushScope(state);
 
+    /// <summary>
+    /// Restores a previously saved/cloned scope to the top of the stack
+    /// </summary>
+    public static void RestoreScope(Scope savedScope)
+    {
+        if (CurrentHub is IHubEx hub)
```
bruno-garcia, line 248: This type checks are odd, and if the type isn't what we expect we'll just not observe the behavior we expect?
How about a `else debug.log` at least?

src/Sentry/SentrySdk.cs
```diff
@@ -240,6 +240,17 @@ public void Dispose()
     [DebuggerStepThrough]
     public static IDisposable PushScope<TState>(TState state) => CurrentHub.PushScope(state);
 
+    /// <summary>
+    /// Restores a previously saved/cloned scope to the top of the stack
+    /// </summary>
+    public static void RestoreScope(Scope savedScope)
```
bruno-garcia, line 246: It feels backwards that we're adding a public API to the static class but not exposing that on `ISentryClient` or `IHub`. This static API was designed for convinience, and only mapping to one of the two interfaces. And code we write relies on the abstractions for testability sake.

We can see this design choice is odd when we're doing type check for a specific type of Hub before we can call the method.

Can we avoid these public methods altogether?

jamescrosswell:
> If I understand this correctly the `RestoreScope` basically pushes a "too soon popped" scope back on the stack so the whole "apply scope to transaction" logic down the line works with `GetCurrentScope`?

Kind of yeah. currently our middleware includes this line... note the `using` - so dispose will get called on the return value when it falls out of scope.:
https://github.com/getsentry/sentry-dotnet/blob/7d05542210f62c48f585f1be6c4e040ea7542fac/src/Sentry.AspNetCore/SentryMiddleware.cs#L86

That makes sense, for the most part, since the ScopeManager we use in ASP.NET Core is using an `AsyncLocalScopeStackContainer`... our scope/client stack is designed to be kept "per request" and we're using the AsyncLocal context as a proxy for "per request". That's a fair assumption - I looked at the implementation for `IHttpContextAccessor` and it also uses AsyncLocal.

All good, until we want to start tracing stuff to do with the request via Diagnostic events that probably get emmited in the context of that AsyncLocal middleware pipeline but don't get received/consumed in that same AsyncLocal context. By the time the OpenTelemetry .NET SDK (and therefore our `SentrySpanProcessor`) receive those diagnostic events, the ScopeManager no longer holds a reference to that Scope (and it may have been garbage collected).

Looking at [what happens in the Dispose method](https://github.com/getsentry/sentry-dotnet/blob/549a0d921c04ead40a010e243a39ac5ec4e01e31/src/Sentry/Internal/SentryScopeManager.cs#L168-L184), it looks like the Scope itself doesn't get disposed - rather the scope/client call stack gets reset to it's previous state (which doesn't include the new scope that was cloned/pushed to the top of the stack). So if we retained a reference to that scope, we could prevent it from being garbage collected. Maybe we could avoid having to clone it then - I'll play around and see if I can get that working.

src/Sentry/SentrySdk.cs
jamescrosswell: I removed from the public interfaces/sdk entirely...

jamescrosswell:
> But does that mean that the whole stack gets copied with every request?

It was never the whole stack (just the top scope on the stack) but actually the Scope itself doesn't get disposed explicitly, so we can avoid even that (see latest commit). No cloning required then!

jamescrosswell:
> Can we reuse the OtelEnricher or a transaction processor for applying the scope instead?

@bitsandfoxes I'm not sure how easy/practical that would be. We'd essentially have to replicate all of the [logic here](https://github.com/getsentry/sentry-dotnet/blob/d189a78c2e5eb5a5ccab15a205c8cbd96b7c21ef/src/Sentry/Internal/Hub.cs#L492C1-L516C14) somehow and execute it in our SpanProcessor... then we probably don't want to run that logic again in the `Hub.CaptureTransaction` method.  Seems like a much more major refactoring, so if we want to do that I reckon we do it outside this bug fix PR.

bitsandfoxes (approved):
Thanks!

PR #2696, opened 2023-10-04
Reordered parameters to TransactionContext and SpanContext constructors

jamescrosswell:
Consolidated multiple TransactionContext constructors into a single constructor with default values for most of the arguments... so all the convience of some overloads and all the power of others. 

# SDK Change Notes

Both the `TransactionContext` and `SpanContext` classes now have default values for most of the values that can be passed to their constructors, which should mean you no longer need to provide unecessary `null` values etc. for arguments you don't care about. 

However in order to make this change, the order of parameters had to be changed so that the non-optional parameters come first in the constructor method signatures. If you're constructing instances of these classes in your own code (which is unlikely), you will need to adjust the order in which you pass parameters to these. [See the PR changes](https://github.com/getsentry/sentry-dotnet/pull/2696/files) for details.


src/Sentry/SpanContext.cs
```diff
@@ -37,17 +37,17 @@ public class SpanContext : ITraceContext
     /// Initializes an instance of <see cref="SpanContext"/>.
     /// </summary>
     public SpanContext(
-        SpanId spanId,
+        SpanId? spanId,
```
bitsandfoxes, line 40: Thanks for weeding out the constructors.
Since this is public and user-facing, can we make the final one even more user-friendly? If I were to create a new instance without propagating anything it would look like 
`var s = new SpanContext(null, null, null, "my-operation", null, null, null);`
Can we use defaults like with the new `TransactionContext` constructor?

bitsandfoxes (approved):
Thanks for pulling all those together!

bruno-garcia (commented):
didn't really look at the change, sorry but another nit on the changelog: this is a breaking change and ppl become afraid of taking them sometimes. It's helpful when we give folks a migration path that's clear to follow

PR #2702, opened 2023-10-05
fix build on mac m1

vaind:
#skip-changelog


bruno-garcia (changes requested):
lets just go with net8

vaind:
> lets just go with net8

would love to but we need to fix branch 4.0.0 **now** and we cannot switch to .net 8 before RC2 which fixes build on mac in general (see #2622)

bruno-garcia (approved):

bruno-garcia:
got it. I thought that was merged already.

Issue #2664, opened 2023-09-26
Use Version as fallback


jamescrosswell:
https://github.com/getsentry/sentry-dotnet/blob/e75d537c14ded617c99eb33f19cf4ce69ca4de36/src/Sentry/Reflection/AssemblyExtensions.cs#L44-L48

bruno-garcia:
Also please note there is an initative to have a default `release` in all SDKs, based on running `git` at compile time

bruno-garcia:
Also another thing we can do on 4.0: https://github.com/getsentry/sentry-javascript/discussions/9195#discussioncomment-7211748

Lets have more fallbacks where we get the SHA from one of those env vars at compile time? We have something that at compile timer adds an `[assembly: ..` attribute already I believe

bruno-garcia:
Oh @AbhiPrasad found it: 

https://github.com/getsentry/develop/issues/755

jamescrosswell:
> That is, unless newest versions of .NET have the `AssemblyInformationVersion` populated automatically, in which case we'd be reinventing the wheel

Whatever people put in the [AssemblyInformationalVersionAttribute](https://learn.microsoft.com/en-us/dotnet/api/system.reflection.assemblyinformationalversionattribute?view=net-7.0#remarks) can be obtained at run time from the [Application.ProductVersion](https://learn.microsoft.com/en-us/dotnet/api/system.windows.forms.application.productversion?view=net-7.0) property.
> Typically, a version number displays as major number.minor number.build number.private part number. You can set it explicitly by setting the assembly version within your assembly manifest. 

Potentially folks can add a SHA in the `private part number` by passing `/p:SourceRevisionId=foo` when building the assembly ([Hanselman describes this](https://www.hanselman.com/blog/adding-a-git-commit-hash-and-azure-devops-build-number-and-build-id-to-an-aspnet-website)).

We could use a RegEx or something to check whether something resembling a commit hash has been embedded in the private part number then.

bruno-garcia:
> Whatever people put in the [AssemblyInformationalVersionAttribute](https://learn.microsoft.com/en-us/dotnet/api/system.reflection.assemblyinformationalversionattribute?view=net-7.0#remarks) can be obtained at run time from the [Application.ProductVersion](https://learn.microsoft.com/en-us/dotnet/api/system.windows.forms.application.productversion?view=net-7.0) property.

That's a .NET Framework API. Not available on .NET

[...]

Issue #1538, opened 2022-03-18
Release version not being read from assembly and defaulting to 1.0.0


carlin-q-scott:
### Environment

How do you use Sentry?
Sentry SaaS

Which SDK and version?

- Sentry.AspNetCore 3.13.0
- Microsoft.AspNetCore 5.0.0
- .NET 5.0

### Steps to Reproduce

1. Set my Website AssemblyVersion and AssemblyFileVersion in my .csproj file
2. Ran my project in Debug mode from Visual Studio
3. Triggered a Sentry Event

### Expected Result

I expected my release version to be set to either of the versions I set in my csproj.

### Actual Result

My release version is 1.0.0. The Assembly name is correct.

carlin-q-scott:
This is how I get my version at runtime for health check API:
- AssemblyVersion: `System.Reflection.Assembly.GetExecutingAssembly().GetName().Version.ToString()`
- AssemblyFileVersion: `System.Diagnostics.FileVersionInfo.GetVersionInfo(System.Reflection.Assembly.GetExecutingAssembly().Location).FileVersion.ToString()`

I noticed that this is not how sentry-dotnet does it. For one thing, you're using the EntryAssembly rather than the ExecutingAssembly. Is that because the executing assembly would be Sentry.DotNetCore?

bruno-garcia:
I believe this changed recently. Might have been a regression

SimonCropp:
while we look into this. a workaround is to set `SentryOptions.Release` to a custom value.

mattjohnsonpint:
It would appear this behavior changed in #1079, release with 3.6.0.  A code comment was added at that time:

> ```csharp
> // Note: even though the informational version could be "invalid" (e.g. string.Empty), it should
> // be used for versioning and the software should not fallback to the assembly version string.
> // See https://github.com/getsentry/sentry-dotnet/pull/1079#issuecomment-866992216
> ```

Then during a refactoring in #1083, another code comment was added:

> ```csharp
> // TODO: Lets change this in a new major to return the Version as fallback
> ```

However, in [the docs](https://docs.sentry.io/platforms/dotnet/configuration/releases/#setting-a-release) we still say this:

> The SDK will first look at the entry assembly's `AssemblyInformationalVersionAttribute`, which accepts a string as value and is often used to set a GIT commit hash.
>
> If that returns null, it'll look at the default `AssemblyVersionAttribute` which accepts the numeric version number.

[...]

PR #2705, opened 2023-10-06
Fallback to `AssemblyVersion` if `InformalVersion` is `NullOrEmpty`

bitsandfoxes:
Resolves https://github.com/getsentry/sentry-dotnet/issues/2664
This removed the comments and todo.
Based on this comment: https://github.com/getsentry/sentry-dotnet/issues/1538#issuecomment-1105929550 the `AssemblyInformationalVersionAttribute` is automatically created and defaults to `<Version>`.
This can be disabled with the `GenerateAssemblyInformationalVersionAttribute` leading to the informal version returning `null`.
In that case, the SDK falls back to the assembly version.

#skip-changelog


bruno-garcia (approved):
Tests need to be addressed or are they uncovering something we didn't think about?
Otherwise: lgtm

PR #2709, opened 2023-10-09
feat: blocking detection

bruno-garcia:
Just playing around, but after a thread on twitter (https://twitter.com/brungarc/status/1711484812702171309). Based in the [Ben Adams BlockingDetector](https://github.com/benaadams/Ben.BlockingDetector?tab=readme-ov-file#caveats).

## Suppressing Blocking Detection

If users are intentionally making blocking calls in their code for some reason, the blocking detection can be suppressed (temporarily) via:

```csharp
using (new SuppressBlockingDetection())
{
    // Some blocking code
}
```

## Volume Concerns

If blocking calls exist on a hot path, we may end up sending lots of events.

We did create a solution to this in https://github.com/getsentry/sentry-dotnet/pull/3174 but decided to roll this back in this PR due to concerns about using metrics. 

## Top Frame Decision

Bruno had experimented, in the initial code, with the number of frames to skip when capturing a stack trace. The thinking was that it might be nice to see one or two calls into the Sentry blocking detection code. This turns out to be [problematic when none of the stack frames are InApp](https://github.com/getsentry/sentry-dotnet/pull/2709#issuecomment-1945437802), since Sentry doesn't know which frame to highlight. So I've reverted to Ben Adams' original code that leaves the culprit blocking call at the top of the stack. This makes it easy to find regardless of whether there are InApp frames present in the stack trace or not.  

## Grouping

There were concerns about grouping and fingerprinting... By default Sentry uses the stack trace as a fingerprint and we can get different stack traces, depending on how the async state machine schedules the task. However this is also related to the top frame decision and given that we're skipping any frames related to the TPL now, the grouping problem has gone away. 

# Resources
- [It's all about the SynchronizationContext](https://learn.microsoft.com/en-us/archive/msdn-magazine/2011/february/msdn-magazine-parallel-computing-it-s-all-about-the-synchronizationcontext) - Stephen Cleary
- [ASP.NET SynchronizationContext](https://blog.stephencleary.com/2017/03/aspnetcore-synchronization-context.html) - Stephen Cleary
- [ExecutionContext vs SynchronizationContext](https://devblogs.microsoft.com/pfxteam/executioncontext-vs-synchronizationcontext/) - Stephen Toub

Aparently you have to be called Stephen to understand Synchronization 😜


bruno-garcia:
since I haven't had a change to look into this for a long time, some notes I have:

> Stack trace in SentryEvent.Thread and also in Sentry.Exception? Seems like it's done so the UI renders in some way
> Some thread info only shows if there are more than 1 thread in the array (Create 1 manually to see it in the UI)

> If all `InApp=false` groups by the whole thing and creates new issues for each event. Group by URL in that case? (also risky because of parameterized URL)

jamescrosswell:
@bruno-garcia are we sure we want this as the call stack?
image

I know you intentially changed skipframes from `3:6` to `0:3` but it seems weird to have some Sentry code at the top of the stack (rather than highlighting the culprit code - i.e. the `In App` frame).

bruno-garcia:
Braindump during chat with @jamescrosswell 

* Concern with volume: If there's a blocking call on a hot path (like on `GET /product`) this will fire on each request and burn through quota.
  * We could keep state on the client and fire it once. 
  * Folks can use error sample rate, or sample on `beforeSend` specifically for this error. 
* Use metrics and code locations? But here we really need the full stack trace.
* What should be the top frame? In other words, how many frames to skip? (original code `DetectionSource.SynchronizationContext ? 3 : 6`). 
  * Sentry focuses (expands) on the first InApp frame so it's OK to include some framework code that's called from it. But not clear how many frames/where to go
* The SDK itself triggers at least 2 events on the first request. Only happen once but if we can, we need to fix these before going ahead.
  * If this is intentional, we need to opt-out of reporting this one.
* We need a mechanism to drop them outside giving folks code to write (e.g; beforeSend snippets), see point above. We need an API to drop certain errors.
* `SentryThreads` vs `SentryException`. I found it odd looking at events that we had stack traces on both (duplicated). And the UI rendered different depending on which one was used. I tried to capture this in this note (https://github.com/getsentry/sentry-dotnet/pull/2709#issuecomment-1927400061)
  * Not really related to blocking detector but came up while doing this. 
  * To better understand this, uncomment the code in the PR (`SentryThreads = new []{new SentryThread`) and see how the UI changes. Also inspect JSON and note that we have 1 dupe stack trace, matching by ThreadId IIRC.
* Grouping problems. We'd need to make sure each blocking detected groups properly. If all frames are inApp=false, the async stack trace machinery can change and result in different groups. Not sure how to solve this, might need to Fingerprint as BlockingDetectorNotInAPp or group by URL in that case? (also risky because of parameterized URL, `transaction` might work if not raw url based).
* Tests - From the lib might help but mainly how it groups, etc. How does it work within Sentry?
* Documentation: Not all things are detected. As Ben Adams listed in details on the original code's repo

jamescrosswell:
> * Grouping problems. We'd need to make sure each blocking detected groups properly. If all frames are inApp=false, the async stack trace machinery can change and result in different groups. Not sure how to solve this, might need to Fingerprint as BlockingDetectorNotInAPp or group by URL in that case? (also risky because of parameterized URL, `transaction` might work if not raw url based).

@bruno-garcia how could someone end up with all of the frames `InApp==false`? I can hack this by adding `frame.InApp = false;` in here:
https://github.com/getsentry/sentry-dotnet/blob/85509fd463a995dba12f6638e767d442e6e88a21/src/Sentry/Internal/DebugStackTrace.cs#L57-L60

And yeah, the grouping goes out the window when I do that. I wanted to create a more realistic stack trace scenario though.

bruno-garcia:
> @bruno-garcia how could someone end up with all of the frames InApp==false? 

The whole stack trace was made only of framework stuff

Here's an example: https://sentry-sdks.sentry.io/issues/4754111791/?project=5428537&query=is%3Aunresolved&referrer=issue-stream&statsPeriod=90d&stream_index=2

The events I created while testing are still in Sentry (1 month left before they expire): https://sentry-sdks.sentry.io/issues/?project=5428537&query=message%3A%22Blocking+call+detected%22&referrer=issue-list&statsPeriod=90d

jamescrosswell:
> > @bruno-garcia how could someone end up with all of the frames InApp==false?
> 
> The whole stack trace was made only of framework stuff

OK cool... I can simulate that easily enough by putting a blocking call in our `SentryMiddleware` 😉 

This is actually one situation in which it might make sense to skip a few more frames (per Ben's original code) as without Sentry's smarts to expand the first InApp frame, it's hard to work out which frame is the problem and where the blocking call is coming from. 

[Here](https://sentry-sdks.sentry.io/issues/4979323571/?project=6601807&query=is%3Aunresolved&referrer=issue-stream&statsPeriod=5m&stream_index=5) it's the third frame down (the call from the middleware). And I know that because I just intentionally wrote the code. If it was a blocking call event out of the blue for someone else's code though, it'd be harder to isolate.
image

In terms of how to group them:
1. If there are InApp frames, I think the default grouping works well (it groups by stack trace)
2. If there aren't InApp frames, maybe we go by code location (assuming we can correctly isolate the line of code that's the culprit) plus the mechanism? I'll play with that and see how it works (unless you've got a better suggestion).

jamescrosswell:
I think the grouping issues when there are no InApp frames stems from variations in the state machine after the blocking call. The latest commit removes all frames related to the BlockingDetector itself and now the grouping seems to work well just based on the call stack (even when there are no InApp frames).

jamescrosswell:
> > Stack trace in SentryEvent.Thread and also in Sentry.Exception? Seems like it's done so the UI renders in some way
> > Some thread info only shows if there are more than 1 thread in the array (Create 1 manually to see it in the UI)

Not sure I know what this is supposed to be doing. What I'm seeing is this:

image

And then this (if I change the thread):

image

So yes, seeing the stack trace on both threads... but at least the stack trace is consistent. 

Should it only show for one of those threads do you think @bruno-garcia? Is this maybe a UX issue?

bruno-garcia:
> Should it only show for one of those threads do you think @bruno-garcia? Is this maybe a UX issue?

So we're only missing the thread-name that could be useful, right? Nothing else there is really useful it seems? 
Probably OK to ignore this then

bruno-garcia:
>             // Skip frames relating to the async state machine
>            || frameInfo?.StartsWith("System.Threading") == true;

Won't this potentially hide useful information about where the problem is? We could mark these on the backend as grouping=false although I don't have a lot of details on how to do that. @adinauer I believe did some stuff for Java and might know (I believe https://github.com/getsentry/sentry/pull/45185)

jamescrosswell:
> Won't this potentially hide useful information about where the problem is? We could mark these on the backend as grouping=false although I don't have a lot of details on how to do that. @adinauer I believe did some stuff for Java and might know (I believe [getsentry/sentry#45185](https://github.com/getsentry/sentry/pull/45185))

These frames all come after the blocking call so I don't think they're relevant to the event.

adinauer:
@bruno-garcia @jamescrosswell glancing at the code here this seems different from what I changed for Java SDK. We removed some auto generated IDs from class names in the stacktrace so they don't mess up grouping in the issue you linked.

src/Sentry/Ben.BlockingDetector/StaticRecursionTracker.cs
```diff
@@ -0,0 +1,11 @@
+namespace Sentry.Ben.BlockingDetector;
+
+internal class StaticRecursionTracker : IRecursionTracker
+{
+    [ThreadStatic] private static int RecursionCount;
```
bruno-garcia, line 5: does this need to be AsyncLocal instead?

jamescrosswell: In the original code it's a ThreadSatic. See [here](https://github.com/benaadams/Ben.BlockingDetector/blob/fce6c534dfcee6ba8cb3ef74d246211b81f5cbeb/src/Ben.BlockingDetector/BlockingMonitor.cs#L10-L11) and [here](https://github.com/benaadams/Ben.BlockingDetector/blob/fce6c534dfcee6ba8cb3ef74d246211b81f5cbeb/src/Ben.BlockingDetector/TaskBlockingListener.cs#L12-L13)... I think the intent is to report blocking only on the first occurrence. So if you have some code that does:
```
- async stuff 1
  -  async stuff 2
    - blocking stuff 3
      - blocking stuff 4
```
... you only get a blocking detection alert for blocking stuff 3.

Arguably, someone could possibly do something like this:

```
- async stuff 1
  - blocking stuff 2
    -  async stuff 3
      - blocking stuff 4
```

And there's maybe some chance blocking stuff 4 gets scheduled on the same thread as blocking stuff 2 in this case. It's a bit of an outlier though and would require a complete redesign of the blocking detector. Maybe we tackle that if it ever becomes an issue (and if we get some good engagement with this first draft of the feature).

bruno-garcia:
> @bruno-garcia @jamescrosswell glancing at the code here this seems different from what I changed for Java SDK. We removed some auto generated IDs from class names in the stacktrace so they don't mess up grouping in the issue you linked.

yes, it's a totally different use case. But the goal here is to send frames that we'd like to show, but not group by. Which can be achieved by changing the server the similarly to how you did it.

bruno-garcia:
@vaind added u as reviewer since @jamescrosswell wrote the code since the initial push I made so I guess both of us are like "not sure I should review or merge this" :D

src/Sentry/Internal/MainExceptionProcessor.cs
```diff
@@ -155,7 +155,7 @@ private SentryException BuildSentryException(Exception exception, int id, int? p
             sentryEx.Mechanism = mechanism;
         }
 
-        sentryEx.Stacktrace = SentryStackTraceFactoryAccessor().Create(exception);
+        sentryEx.Stacktrace ??= SentryStackTraceFactoryAccessor().Create(exception);
```
bitsandfoxes, line 158: What's the implication here? Previously, the stacktrace would get overwritten. Why is this change necessary? Where would the stacktrace come from?

src/Sentry/Internal/MainSentryEventProcessor.cs
```diff
@@ -86,7 +86,9 @@ public SentryEvent Process(SentryEvent @event)
         // if there's no exception with a stack trace, then get the current stack trace
         if (@event.Exception?.StackTrace is null)
         {
-            var stackTrace = SentryStackTraceFactoryAccessor().Create();
+            var stackTrace = @event.SentryExceptions?.FirstOrDefault() is { } sentryEx
```
bitsandfoxes, line 89: This means that there are now events with no exception but a SentryException?

src/Sentry/Ben.BlockingDetector/BlockingMonitor.cs
```diff
@@ -0,0 +1,107 @@
...
+                                Type = "BlockingCallDetector",
+                                Handled = false,
+                                Description = "Blocking calls can cause ThreadPool starvation."
+                            },
+                            Type = "Blocking call detected",
+                            Stacktrace = stackTrace,
+                        }
+                    },
+                };
+                evt.SetTag("DetectionSource", detectionSource.ToString());
```
bitsandfoxes, line 81:
```suggestion
```
In general, the SDK should not set tags automatically. We could put the detection source on the context and extract it from there.

src/Sentry/Ben.BlockingDetector/BlockingMonitor.cs
```diff
@@ -0,0 +1,107 @@
...
+            _options = options;
+            _recursionTracker = recursionTracker;
+        }
+
+        private static bool ShouldSkipFrame(string? frameInfo) =>
+            frameInfo?.StartsWith("Sentry.Ben") == true
+            // Skip frames relating to the TaskBlockingListener
+            || frameInfo?.StartsWith("System.Diagnostics") == true
+            // Skip frames relating to the async state machine
+            || frameInfo?.StartsWith("System.Threading") == true;
```
bitsandfoxes, line 30: I think we can skip doing this and improve grouping on the server side for this kind of stuff in general.

src/Sentry/Ben.BlockingDetector/BlockingMonitor.cs
bitsandfoxes, line 81:
```suggestion
                                Description = "Blocking calls can cause ThreadPool starvation.",
                                Source = detectionSource.ToString()
                            },
                            Type = "Blocking call detected",
                            Stacktrace = stackTrace,
                        }
                    },
                };
```
The SDK should not set tags automatically. But we could put it on the mechanism source?

src/Sentry/Internal/MainExceptionProcessor.cs
bruno-garcia: the stack trace is already set in the event. We get it from the blocking detection logic so we won't want it overwriten

src/Sentry/Internal/MainSentryEventProcessor.cs
jamescrosswell: @bitsandfoxes yes. When we detect a blocking call, we capture that as an Event that has a SentryException but not System.Exception (since none was thrown). See [here](https://github.com/getsentry/sentry-dotnet/blob/45c67b95174c637687a2b038d4b14f5f1fabbce3/src/Sentry/Ben.BlockingDetector/BlockingMonitor.cs#L53-L80) for the code. 

@bruno-garcia not sure I follow - I'm not sure what the server requirements are.

src/Sentry/Ben.BlockingDetector/BlockingMonitor.cs
jamescrosswell: @bitsandfoxes could you link to the associated PR where this logic gets changed on the server?

Also, if we do skip this here, we'll need to make some changes to set a custom fingerprint that excludes all this stuff (since by default the entire stacktrace is used as the fingerprint).

bitsandfoxes: I'll follow up on this. The blocking detection is opt-in and the improvements to the grouping are not blocking this PR.

src/Sentry/Internal/MainExceptionProcessor.cs
bitsandfoxes:
```suggestion
        sentryEx.Stacktrace = SentryStackTraceFactoryAccessor().Create(exception);
```

The `sentryEx` gets newly instantiated in this method and the `StackTrace` has to be set explicitly. I think this change does nothing.

> We get it from the blocking detection logic so we won't want it overwriten

That should all happen in the event processor.

src/Sentry/Internal/MainSentryEventProcessor.cs
bitsandfoxes: > capture that as an Event that has a SentryException but not System.Exception (since none was thrown)

That makes sense. Thanks

bitsandfoxes (approved):
This looks really good to me!

bruno-garcia:
Thanks folks!

src/Sentry.AspNetCore/SentryMiddleware.cs
```diff
@@ -132,7 +145,24 @@ public async Task InvokeAsync(HttpContext context, RequestDelegate next)
             try
             {
                 var originalMethod = context.Request.Method;
-                await next(context).ConfigureAwait(false);
+                if (_options.CaptureBlockingCalls && _monitor is not null)
+                {
+                    var syncCtx = SynchronizationContext.Current;
+                    SynchronizationContext.SetSynchronizationContext(syncCtx == null ? _detectBlockingSyncCtx : new DetectBlockingSynchronizationContext(_monitor, syncCtx));
```
jamescrosswell, line 151: @bruno-garcia won't the SynchronizationContext in ASP.NET Core always be null? Is this code here just in case SDK users have implemented a custom sync context for some reason?

Issue #699, opened 2020-12-28
Diagnostic logger change order of args


bruno-garcia:
The `LogError` should take `Exception` as the first arg. So it doesn't get mixed up with the `<TArg` overloads.
There shouldn't be a nullable Exception parameter, either takes an exception or it doesnt.

SimonCropp:
@bruno-garcia it looks like this is done?

```
       public static void LogError(this ILogger logger, Exception exception, string message, params object[] args)
        {
            logger.Log(LogLevel.Error, exception, message, args);
        }

```

bitsandfoxes:
Closed by [#2715](https://github.com/getsentry/sentry-dotnet/pull/2715)

PR #2715, opened 2023-10-11
Change signature order in `DiagnosticLogger` for exceptions

bitsandfoxes:
Resolves https://github.com/getsentry/sentry-dotnet/issues/699

### DIFF
--- a/src/Sentry/Extensibility/DiagnosticLoggerExtensions.cs
+++ b/src/Sentry/Extensibility/DiagnosticLoggerExtensions.cs
@@ -211,8 +211,15 @@ public static class DiagnosticLoggerExtensions
     /// </summary>
     public static void LogError(
         this IDiagnosticLogger logger,
-        string message,
-        Exception? exception = null)
+        string message)
+        => logger.LogIfEnabled(SentryLevel.Error, null, message);
+
+    /// <summary>
+    /// Log an exception with an error message.
+    /// </summary>
+    public static void LogError(this IDiagnosticLogger logger,
+        Exception exception,
+        string message)
         => logger.LogIfEnabled(SentryLevel.Error, exception, message);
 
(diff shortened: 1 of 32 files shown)


src/Sentry/Extensibility/DiagnosticLoggerExtensions.cs
```diff
@@ -211,68 +211,77 @@ public static class DiagnosticLoggerExtensions
     /// </summary>
     public static void LogError(
         this IDiagnosticLogger logger,
-        string message,
-        Exception? exception = null)
+        string message)
+        => logger.LogIfEnabled(SentryLevel.Error, null, message);
```
jamescrosswell, line 215: Could potentially merge this with the next one, making `Exception` an optional parameter with a default value of `null`... makes for one less hop when debugging and tracing into things (and possibly one less entry on the call stack, depending on how the compiler optimises things).

src/Sentry/Extensibility/DiagnosticLoggerExtensions.cs
```diff
@@ -211,68 +211,77 @@ public static class DiagnosticLoggerExtensions
...
 
     /// <summary>
     /// Log a error message.
     /// </summary>
     internal static void LogError(
         this SentryOptions options,
-        string message,
-        Exception? exception = null)
+        string message)
+        => options.DiagnosticLogger?.LogIfEnabled(SentryLevel.Error, null, message);
```
jamescrosswell, line 231: Same again, these two methods could be merged.

jamescrosswell (commented):
Looks good. I left a couple of comments about consolidating some of the overloads... there are more opportunities than the two I pointed out but figured that was obvious 😉

jamescrosswell (approved):

Issue #1585, opened 2022-04-21
Obsolete unnecessary APIs


mattjohnsonpint:
This issue will keep track of APIs that we want to remove in the next major release (4.0.0).  Will keep this open and update as we find them.  See https://github.com/getsentry/sentry-dotnet/blob/main/test/Sentry.Tests/ApiApprovalTests.Run.DotNet6_0.verified.txt for list from the main Sentry project.

- [x] https://github.com/getsentry/sentry-dotnet/issues/1590
- [ ] https://github.com/getsentry/sentry-dotnet/issues/2718
- [ ] https://github.com/getsentry/sentry-dotnet/pull/2717

mattjohnsonpint:
> Does scope.Platform qualify? #1590

Yes, I think so.

bruno-garcia:
Other good candidates: `WIthScope` API. It was reverted from being marked obsolete due to what I believe was a misunderstanding, since it's broken on global mode. And a new one with `async` was introduced, Another one worth considering making obsolete is the original; ConfigureScopeAsync, but this is less obvious:

That's because the SDK avoid calling callbacks if disabled as an optimization. While people put their core business logic inside our callbacks.  So the async call can be spared if the sole reason to run it is to fetch data to set to the scope. But WithScope has logic that happens inside and folks expect to always run regardless of Sentry being configured.

Local scope works in global mode and local mode and is advised instead.

jamescrosswell:
We now have separate issues for each of the items mentioned here and we already have the feat/4.0.0 pull request as an umbrella to sit across these (plus all the others)... so closing this issue as redundant.

PR #2717, opened 2023-10-11
Removed `WithScope` and `WithScopeAsync`

bitsandfoxes:
Relates to https://github.com/getsentry/sentry-dotnet/issues/1585


CHANGELOG.md
```diff
@@ -28,6 +28,7 @@ API Changes:
 - Enable `CaptureFailedRequests` by default ([2688](https://github.com/getsentry/sentry-dotnet/pull/2688))
 - Additional constructors removed from `TransactionTracer`. ([#2694](https://github.com/getsentry/sentry-dotnet/pull/2694))
 - Removed the `Scope.Platform` property as it was never applied ([#2695](https://github.com/getsentry/sentry-dotnet/pull/2695))
+- `WithScope` and `WithScopeAsync` have been proven to not work correctly in desktop contexts when using a global scope. They are have been removed in favor of the overloads of `CaptureEvent`, `CaptureMessage`, and `CaptureException`. Those methods provide a callback to a configurable scope. ([#2717](https://github.com/getsentry/sentry-dotnet/pull/2717))
```
bruno-garcia, line 31:
```suggestion
- `WithScope` and `WithScopeAsync` have been proven to not work correctly in desktop contexts when using a global scope. Removing them in favor of the overloads of `CaptureEvent`, `CaptureMessage`, and `CaptureException`. Those methods provide a callback to a configurable scope that doesn't . ([#2717](https://github.com/getsentry/sentry-dotnet/pull/2717))
```

I think it would be nice to expand. Like: Before, After

WIthScope(..); // Becomes:

CaptureEvent(..).

jamescrosswell: > That doesn't

That doesn't what? That doesn't not work correctly? If so, I'd take out the double negative and go with "that work correctly when using a global scope" or something like that.

jamescrosswell (approved):

jamescrosswell:
Nice! 🧹

Issue #2721, opened 2023-10-12
AspNetCore chaining UseRouting().UseEndpoints crash with InvalidOperationException


Julien-Mialon:
### Package

Sentry.AspNetCore

### .NET Flavor

.NET

### .NET Version

7.0.12

### OS

Any (not platform specific)

### SDK Version

3.40.0 & 3.40.1

### Steps to Reproduce

Made a small repo for repro here : https://github.com/Julien-Mialon/SentryRoutingEndpointIssue/

But basically, create an AspNet API project, add Sentry, use this line of code in Startup Configure method
```csharp
app.UseRouting().UseEndpoints(endpoints => endpoints.MapControllers());
```
And it will crash with exception : 
```
System.InvalidOperationException: The EndpointRoutingMiddleware and EndpointMiddleware must be added to the same IApplicationBuilder instance. To use Endpoint Routing with 'Map(...)', make sure to call 'IApplicationBuilder.UseRouting' before 'IApplicationBuilder.UseEndpoints' for each branch of the middleware pipeline.
```

This is due to app being a SentryTracingBuilder but app.UseRouting() returning the default ApplicationBuilder which is normally wrapped in SentryTracingBuilder.

I'm not understanding every impact of this but maybe the Use method in SentryTracingBuilder should return itself in all case to avoid that ?

### Expected Result

No crash

### Actual Result

Exception above

bitsandfoxes:
Hey @Julien-Mialon and thanks for raising this and providing a repro. I suspect https://github.com/getsentry/sentry-dotnet/pull/2602 being the culprit.

jamescrosswell:
Thanks @Julien-Mialon - great catch!

I've tracked down the problem and put together a PR to resolve this. I'll put details in the PR about root cause and solution.

PR #2726, opened 2023-10-15
Calls to SentryTracingBuilder can now be chained correctly

jamescrosswell:
Resolves https://github.com/getsentry/sentry-dotnet/issues/2721

# Problem

The dotnet `UseRouting` and `UseEndpoints` methods are just extensions on the builder... they set properties etc. on the builder before adding middleware. The only part of what they do that gets passed to the `SentryTracingBuilder.InnerBuilder` then is the delegate that they return on the last line of their implementation (e.g. [this](https://github.com/dotnet/aspnetcore/blob/8eaf4b51f73ae2b0ed079e4f8253afb53e96b703/src/Http/Routing/src/Builder/EndpointRoutingApplicationBuilderExtensions.cs#L113)).

Previously we were calling that delegate against the InnerBuilder and returning the result directly (which implicitly returns a reference to the InnerBuilder). Chained calls would then be against the InnerBuilder rather than the SentryTracingBuilder. 

This caused issues since `UseRouting` sets properties on the `SentryTracingBuilder` and then (in the chained call) `UserEndpoints` tries to follow up with some matching logic... only it's not using the same builder! It's using the InnerBuilder at this stage 😢 

# Solution

We've separated using the InnerBuilder to call the delegate from the return value in our SentryTracingBuilder so that for chained calls, the same builder will always be used.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Fixes
+
+- Fixed chaining on the IApplicationBuilder for methods like UseRouting and UseEndpoints ([#2726](https://github.com/getsentry/sentry-dotnet/pull/2726))
+
 ### Dependencies
 
 - Bump Cocoa SDK from v8.13.0 to v8.13.1 ([#2722](https://github.com/getsentry/sentry-dotnet/pull/2722))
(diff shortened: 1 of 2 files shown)


bitsandfoxes (approved):
🔥

Issue #804, opened 2021-02-07
Rename Span.extras to Span.data


marandaneto:
Context: https://github.com/getsentry/sentry-cocoa/pull/932#discussion_r570807351

https://develop.sentry.dev/sdk/event-payloads/span/

if it's a breaking change, fine to do in the next major bump.

bruno-garcia:
We could add `data` and set `[Obsolete]` on `extra` and remove it on the next major.

Tyrrrz:
Currently `extras` are also copied from current scope. Should something like this happen to `data`? We don't have `data` on scope.

@bruno-garcia  @marandaneto

marandaneto:
looks like `data` is missing on Java, not sure if it's been removed or not, see https://github.com/getsentry/sentry-java/blob/bc10b2b70d328241b7c4a5dba32d97ad42d7687c/sentry/src/main/java/io/sentry/SpanContext.java
@maciejwalkowiak knows maybe?

iirc it's not being added because it'd be exposed to the transaction interface which does not have a `data` field.

@Tyrrrz `Span.data` != `Scope.extras` afaik, if they are being copied over, It's a bug probably.

`extra` is part of the Event payload https://develop.sentry.dev/sdk/event-payloads/

Tyrrrz:
@marandaneto ah, it was a brainfart on my part, the `extras` are only copied from scope on transactions, not spans. Scratch my question.

marandaneto:
no worries, the Java issue is https://github.com/getsentry/sentry-java/issues/1233

Tyrrrz:
Another question:
Since we currently have `ITransaction` and since `ITransaction` extends `ISpan`, what should we do? Adding `data` to `ISpan` will make it also appear on `ITransaction`. Considering we're trying to make this non-breaking by also not removing `extra` on `ISpan`, this will result in both `extra` and `data` appearing on both `ITransaction` and `ISpan`.

I think we should wait with this change for when we drop `ITransaction` entirely (v4.0?). Otherwise, it will be very confusing in my opinion.

marandaneto:
@Tyrrrz that's exactly the reason why we've not added to Java, if its the case on .NET too, we can do it once we get rid of ITransaction

Tyrrrz:
Yep, I think we should wait.

jamescrosswell:
We're not going to do this.

PR #2730, opened 2023-10-16
Rename Span.extras to Span.data

jamescrosswell:
Resolves https://github.com/getsentry/sentry-dotnet/issues/804

# On Hold

Putting this on pause. The current requirements we know of:
- Scope, event and transaction should have have an `Extra` property
- Span should have `Data`

However we've got an [RFC](https://github.com/getsentry/rfcs/pull/116/files) to flatten stuff on the span... so we're holding off on any changes here until that RFC goes through, at which point `Span.Extra` and `Span.Data` might disappear completely.

### DIFF
--- a/.github/actions/environment/action.yml
+++ b/.github/actions/environment/action.yml
@@ -17,24 +17,15 @@ runs:
         distribution: 'temurin'
         java-version: '11'
 
-    # .NET 3.1 has been removed from all OS due to EOL
-    # https://github.com/actions/runner-images/issues/7667
-    - name: Install .NET 3.1 SDK
+    - name: Install .NET SDK
       uses: actions/setup-dotnet@v3
       with:
-        dotnet-version: 3.1.x
-
-    - name: Install .NET 7 SDK
-      uses: actions/setup-dotnet@v3
-      with:
-        dotnet-version: 7.0.203   # switch back to 7.x.x after resolving the below Tizen issue.
-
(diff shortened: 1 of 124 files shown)


bitsandfoxes (changes requested):
The `Scope` [needs to retain](https://docs.sentry.io/platforms/dotnet/enriching-events/context/#additional-data) `Extra`.

jamescrosswell:
> The `Scope` [needs to retain](https://docs.sentry.io/platforms/dotnet/enriching-events/context/#additional-data) `Extra`.

Are you sure? Those docs say otherwise:
> Additional Data
> Additional Data is deprecated in favor of structured contexts.
>
> Sentry used to support adding unstructured "Additional Data" via SetExtra.

It looks like maybe we should be removing `Scope.Data` as well. I'm not sure what value having `SentryEvent.Extra` gives in that case since it would never come from the Scope, or a Transaction. 

More generally, I can do whatever, but we do need clarity on the actual requirements... which I'm not getting from the links on the [original issue](https://github.com/getsentry/sentry-dotnet/issues/804).

bruno-garcia:
@jamescrosswell can we remove this branch?

jamescrosswell:
> @jamescrosswell can we remove this branch?

Sure, seems unlikely we'd ever use it now... too hard to merge for too little gain.

Issue #2676, opened 2023-09-27
Rename ISpan and ITransaction


jamescrosswell:
`ISpan` might be more appropriately named `ISpanTracer` since it represents properties and methods that are only implemented by `SpanTracer` and `TransactionTracer`. 

For the same reason, `ITransaction` could be more appropriately named `ITransactionTracer`. 

See entity relationship diagram below:

```mermaid
erDiagram
    NoOpSpan ||--|| ISpan : implements

    SpanTracer ||--|| ISpan : implements
    SpanTracer {
        bool IsSentryRequest
        TransactionTracer Transaction
        void Unfinish()
    }


    ISpan ||--|| ISpanData : implements
    ISpan {
        string Description
        string Operation
        SpanStatus Status
        ISpan StartChild()
        void Finish()
    }

    Span ||--|| ISpanData : implements

    ISpanData ||--|| ITraceContext : implements
    ISpanData ||--|| IHasTags : implements
    ISpanData ||--|| IHasExtra : implements
    ISpanData {
        DateTimeOffset StartTimestamp
        DateTimeOffset EndTimestamp
        bool IsFinished
        SentryTraceHeader GetTraceHeader()
    }

    NoOpTransaction ||--|| ITransaction : implements

    TransactionTracer ||--|| ISpan : implements
    TransactionTracer ||--|| ITransaction : implements
    TransactionTracer ||--|| ITransactionData : implements
    TransactionTracer ||--|| ISpan : implements
    TransactionTracer {
        bool IsSentryRequest
        double SampleRate
        IReadOnlyCollection Spans
        DynamicSamplingContext DynamicSamplingContext
        ITransactionProfiler TransactionProfiler
        void AddChildSpan()
        ISpan GetLastActiveSpan()
    }

    ITransaction ||--|| ITransactionData : implements
    ITransaction ||--|| ISpan : implements
    ITransaction {
        string Name
        bool IsParentSampled
        IReadOnlyCollection Spans
        ISpan GetLastActiveSpan()
    }

    Transaction ||--|| ITransactionData : implements
    Transaction {
        SentryId EventId
        double SampleRate
        DynamicSamplingContext DynamicSamplingContext
        ITransactionProfiler TransactionProfiler
        IReadOnlyCollection Spans
        void Redact()
        void WriteTo()
        Transaction FromJson()
    }

    ITransactionData ||--||  ISpanData : implements
    ITransactionData ||--|| ITransactionContext : implements
[...]

PR #2731, opened 2023-10-17
Renamed ISpan and ITransaction to ISpanTracer and ITransactionTracer

jamescrosswell:
Resolves https://github.com/getsentry/sentry-dotnet/issues/2676


src/Sentry/ITransactionTracer.cs
```diff
@@ -1,9 +1,9 @@
 namespace Sentry;
 
 /// <summary>
-/// Transaction.
+/// TransactionTracer interface
```
bitsandfoxes, line 4:
```suggestion
/// TransactionTracer interface.
```

src/Sentry/ISpanTracer.cs
```diff
@@ -3,9 +3,9 @@
 namespace Sentry;
 
 /// <summary>
-/// Span.
+/// SpanTracer interface
```
bitsandfoxes, line 6:
```suggestion
/// SpanTracer interface.
```

src/Sentry/Span.cs
```diff
@@ -88,7 +88,7 @@ public Span(SpanId? parentSpanId, string operation)
     /// <summary>
     /// Initializes an instance of <see cref="Span"/>.
     /// </summary>
-    public Span(ISpan tracer)
+    public Span(ISpanTracer tracer)
```
bitsandfoxes, line 91: Idk why but this suddenly makes so much more sense to me.

bitsandfoxes (approved):
Should `NoOpSpan` be renamed to `NoOpSpanTracer`? Same for the `NoOpTransaction`?

PR #2732, opened 2023-10-17
Trimming for .NET 7+

jamescrosswell:
# Summary
One of the pre-requisites for supporting AOT for Sentry to support trimming. 

In the Sentry.csproj core package, there are 3 primary challenges with this:
1. BenDemystifier
2. WinUIUnhandledExceptionIntegration
3. System.Text.Json.JsonSerializer

## Ben.Demystifier ✅ 

`Ben.Demystifier` uses reflection heavily to provide enhanced stack traces.

### Approach / Solution

When setting `<IsAotCompatible>true</IsAotCompatible>` in `Sentry.csproj`,  we get a bunch of errors/warnings about Ben.Demystifier not being compatible with trimming. 

To solve this, in our fork of Ben.Demystifier we surface those via `RequiresUnreferencedCode` attributes on the relavant classes. That way Ben.Demystifier itself compiles when included in a solution with `IsAotCompatible` set, but only if any code leveraging it handles the warnings (either be not using the Ben.Demystifier functionality in question or by supressing the warnings). 
- https://github.com/getsentry/Ben.Demystifier/pull/1 includes the necessary changes to `Ben.Demystifier`
- Then we set `UnconditionalSuppressMessage` on:
   - `DebugStackTrace.CreateFrames` and 
   - `DebugStackTrace.TryCreateManagedFrame` and
   - `DebugStackTrace.TryCreateManagedFrame` and

## WinUIUnhandledExceptionIntegration ✅ 

`WinUIUnhandledExceptionIntegration` uses reflection to set `Microsoft.UI.Xaml.Application.Current.UnhandledException` to Sentry's `WinUIUnhandledExceptionHandler` for customers using `Microsoft.UI.Xaml` (.NET MAUI targeting Windows or WinUI 3 applications). The relection code it users generates the following errors with `IsTrimming==true`:

### Approach / Solution

- For the time being, if `IsTrimmable == true` we're not registering our WinUiUnhandledExceptionHandler automatically and we'll need to provide some guidance to users who are both targeting WinUI and wanting to enable AOT compilation.

## System.Text.Json.JsonSerializer ✅

When using Trimming, serialization can't be performed using old reflection based technique (which utilise `JsonSerializer` overloads accepting `JsonSerializerOptions` parameters). 

### Approach / Solution

Instead we need to use source generated serializers that can be supplied to overloads `JsonSerializer` overloads accepting `JsonSerializerContext` arguments. See [`System.Text.Json` source generators](https://devblogs.microsoft.com/dotnet/try-the-new-system-text-json-source-generator/) for details. 

### Bumping the .netcoreapp3 version

In order to use the new source generated JSON serializers, we had to add a PackageReference  to `System.Text.Json" Version="6.0.8"` for version of .NET prior to .net6.0. That resulted in a new error: `System.Runtime.CompilerServices.Unsafe doesn't support netcoreapp3.0. Consider updating your TargetFramework to netcoreapp3.1 or later.`... As a results, we've had to bump the target framework in our Sentry packages from `netcoreapp3.0` to `netcoreapp3.1`.

# Next Steps

Beyond Sentry.csproj many of the other projects can just be added as is to SentryAot.slnf and they compile and tests run without problem. Some have  challenges however. One that was obvious during initial exploration is the use of `ConfigureFromConfigurationOptions<TOptions>` classes to initialize a number of the integrations... which doesn't work with Trimming enabled. This affects:
    - `Sentry.Extensions.Logging`
    - `Sentry.Maui`
    - `Sentry.Azure.Functions.Worker`
    - `Sentry.AspNetCore`

I think we should deal with that in a separate pull request though. 

# Additional Reference material:
- [The minimal API AOT compilation template](https://andrewlock.net/exploring-the-dotnet-8-preview-the-minimal-api-aot-template/)
- [Intro to Trim warnings](https://learn.microsoft.com/en-us/dotnet/core/deploying/trimming/fixing-warnings)
- [Intro to AOT warnings](https://learn.microsoft.com/en-us/dotnet/core/deploying/native-aot/fixing-warnings)
- [Creating a SourceGenerator](https://andrewlock.net/series/creating-a-source-generator/)
- [System.Text.Json source generators](https://devblogs.microsoft.com/dotnet/try-the-new-system-text-json-source-generator/)
- [Packaging Generators in Libraries](https://github.com/dotnet/roslyn/blob/main/docs/features/source-generators.cookbook.md#package-a-generator-as-a-nuget-package)
- [The new configuration binder source generator](https://andrewlock.net/exploring-the-dotnet-8-preview-using-the-new-configuration-binder-source-generator/)


jamescrosswell:
Currently it's only tested/working for `Sentry.csproj`. Also, as a hack (to enable me to test this), if you're targeting .NET 8, trimming always is enabled:
(<https://github.com/getsentry/sentry-dotnet/blob/4c766e56e151eb41f6b7ca67d2274c8d348b1ce6/src/Sentry/Sentry.csproj#L25-L29>). 

There are probably a couple of different ways we could solving that properly. Some I can think of:
1. We let IsTrimmable be passed in via the command line (e.g. when running dotnet publish). 
   - If the `IsTrimmable` property is passed in, we set a `TRIMMABLE` directive that's used for all the conditional compilation. 
   - For .NET 8 test targets, we pass an additional property to the csproj files when we include them (`IsTrimmable`)
2. We could modify the script that we currently use to generate solution filters so that:
   - It copies `Sentry.sln` to `Sentry-AOT.sln`
   - We have a `Sentry-AOT.slnf` solution filter that is based off `Sentry-AOT.sln`
   - When building, if the solution is `Sentry-AOT.sln`, we enable AOT/Trimming
   - Then on CI, we just run tests and publish tasks against the new `Sentry-AOT.slnf` solution filter as well as against the existing `Sentry-CI-*.slnf` files
 
I think the second of these two solutions is a bit more robust as that way we can test against the .NET 8 targets when building managed code packages as well as native packages. As a side note, if it works, we could use the same hack to ressurect a `Sentry-NoMobile.sln` file that is used to set the `NO_MOBILE` flag:
https://github.com/getsentry/sentry-dotnet/blob/34d65a27e1c5bacb8c15da4c97156558e6e3ea95/Directory.Build.props#L21-L21

jamescrosswell:
@vaind I've finally got a proper test setup going with .NET 8 and Trimming and various of the tests that are failing are related to `DebugStackTrace.cs`. [Process_AggregateException](https://github.com/getsentry/sentry-dotnet/blob/2a8813ab6725caac4bc3a76f71af542ffce7f588/test/Sentry.Tests/Internals/MainExceptionProcessorTests.cs#L112), for example, fails because `stackFrame.HasNativeImage()` returns false in the `TryCreateNativeAOTFrame` method. 

https://github.com/getsentry/sentry-dotnet/blob/21aab6d7661406be5bb16dda7cbe56ce3a3f440c/src/Sentry/Internal/DebugStackTrace.cs#L237-L242

Diving into `RealStackFrame`, this just calls:
https://github.com/getsentry/sentry-dotnet/blob/7000964395cddd5107b954131bdc1d74970c3ff4/src/Sentry/Internal/StackFrame.cs#L128

Oddly, in the .NET Source code, that will always resolve to `false`:
* [HasNativeImage](https://github.com/dotnet/runtime/blob/main/src/libraries/System.Private.CoreLib/src/System/Diagnostics/StackFrameExtensions.cs#L9-L14)
* [GetNativeImageBase](https://github.com/dotnet/runtime/blob/main/src/libraries/System.Private.CoreLib/src/System/Diagnostics/StackFrameExtensions.cs#L38-L41)

Any idea what's going on here? 

More philisophically (not really inspired by the code) if we're compiling `Sentry.csproj` with Trimming enabled but we're not compiling `Sentry.Tests.csproj`  (which is where the exception gets raised) with Trimming enabled, there wouldn't ever be a Native image available right? No native image would ever get created for `Sentry.Tests.csproj` because it remains JIT.

src/Sentry/Internal/DebugStackTrace.cs
```diff
@@ -489,6 +496,13 @@ private static void DemangleLambdaReturnType(SentryStackFrame frame)
         }
     }
 
+#if TRIMMABLE
```
bruno-garcia, line 499: how about having a single method declaration and only ifdef the body?

bruno-garcia (commented):
I looked at 10% if it, pretty scary :D

jamescrosswell:
> I looked at 10% if it, pretty scary :D

Yeah, and that's just for `Sentry.csproj`. There will be more to come when we look at the other projects. I think we've got the builds/tooling setup OK now though, so hopefully just grunt work from here on out...

src/Sentry/Internal/DebugStackTrace.cs
jamescrosswell: I did try that initially... since the ifdef wraps another ifdef, it wasn't super readable, tbh. Also, this version will result in ever so slightly less code in an AOT app. 

I'm happy to go either way though if we'd prefer putting the ifdef in the body.

.github/workflows/device-tests-android.yml
```diff
@@ -29,6 +29,12 @@ jobs:
           distribution: 'temurin'
           java-version: '11'
 
+      - name: Install .NET 8
```
vaind, line 32: Note: this has been fixed by #2742

src/Sentry/Internal/Extensions/CollectionsExtensions.cs
```diff
@@ -29,9 +29,14 @@ internal static class CollectionsExtensions
         }
     }
 
-#if !NET8_0_OR_GREATER
-    public static Dictionary<TKey, TValue> ToDictionary<TKey, TValue>(
+    // Note this was renamed from ToDictionary to ToDict to avoid conflicts with System.Linq.Enumerable.ToDictionary.
+    // on .NET 8 and later. Previously this code was simply omitted for .NET 8 targets and later, which compiles fine
+    // but breaks at runtime (where it's still looking for  CollectionExtensions.ToDictionary method).
```
vaind, line 34: 🥇

vaind (commented):
This is getting pretty large. Do you think it would be reasonable to split to make it more digestible to review? E.g. JSON stuff, ben.demystifier stuff, etc. Or is it really just one set of changes?

src/Sentry/Internal/ModuleExtensions.cs
```diff
@@ -0,0 +1,19 @@
...
+    /// The Module.Name for Modules that are embedded in SingleFileApps will be null
+    /// or &lt;Unknown&gt;, in that case we can use Module.ScopeName instead
+    /// </summary>
+    /// <param name="module">A Module instance</param>
+    /// <returns>module.Name, if this is available. module.ScopeName otherwise</returns>
+    public static string? GetNameOrScopeName(this Module module) =>
+#if TRIMMABLE
+        module?.ScopeName;
+#else
+        (module?.Name is null || module.Name.Equals("<Unknown>"))
```
lucas-zimerman, line 15: if module is null, isn't `module.Name` going to to throw null exception?

jamescrosswell:
> This is getting pretty large. Do you think it would be reasonable to split to make it more digestible to review? E.g. JSON stuff, ben.demystifier stuff, etc. Or is it really just one set of changes?

At the moment, it's the minimum to enable AOT for Sentry.csproj,  get it to compile and pass all the tests. I think we should look at merging this into the 4.0.0 branch before we try to do the same for all the other projects in the solution - we could probably tick them off one by one, each in a separate PR, as we slowly add more projects to the `SentryAot.slnf` file.

src/Sentry/Internal/ModuleExtensions.cs
jamescrosswell: If `module?.Name is null` then the expression evaluates to `true`, so `module.Name` never gets evaluated.

src/Sentry/PlatformAbstractions/RuntimeInfo.cs
```diff
@@ -82,6 +82,8 @@ internal static Runtime WithAdditionalProperties(this Runtime runtime)
                 version = frameworkInstallation?.Version?.ToString();
             }
         }
+#elif TRIMMABLE
+    private static string? GetNetCoreVersion(Runtime runtime) => null; // There is no runtime for AOT apps
```
vaind, line 86: The comment seems misleading; there is runtime info, see

https://sentry-sdks.sentry.io/issues/4537780610/events/fae662afea604b7593d35d3ef7ae09af/?project=5428537&query=is%3Aunresolved+os.name%3AWindows&referrer=previous-event&statsPeriod=14d&stream_index=1#context-runtime

src/Directory.Build.props
```diff
@@ -35,6 +35,11 @@
     <CodeAnalysisRuleSet>$(MSBuildThisFileDirectory)CodeAnalysis.ruleset</CodeAnalysisRuleSet>
   </PropertyGroup>
 
+  <PropertyGroup Condition="'$(SolutionName)' == 'Sentry.Aot' and $(TargetFramework.StartsWith('net8'))">
+    <PublishAot>true</PublishAot>
```
vaind, line 39: I'm not sure about defining this property as it may imply that we're publishing AOT libraries, which we're not. Instead, I'd leave it up to specific sample apps to specify the property as needed (and as supported by the current OS).

src/Directory.Build.props
```diff
@@ -35,6 +35,11 @@
...
+    <DefineConstants>$(DefineConstants);TRIMMABLE</DefineConstants>
```
vaind, line 40: Is the target-framework net8 condition necessary here? Maybe we could instead define "Sentry.Aot" solution itself to target net8 (or net7 which, from my experience, seems to work fine too)?

src/Sentry/Sentry.csproj
```diff
@@ -23,21 +23,24 @@
...
-    Ben.Demystifier is compiled directly into Sentry.
-    Note: It uses Microsoft.Bcl.AsyncInterfaces, which we get transitively from System.Text.Json.
-  -->
-  <ItemGroup>
-    <Compile Include="..\..\modules\Ben.Demystifier\src\**\*.cs">
-      <Link>%(RecursiveDir)\%(Filename)%(Extension)</Link>
-    </Compile>
-    <Compile Remove="..\..\modules\Ben.Demystifier\**\obj\**" />
-  </ItemGroup>
+  <Target Name="IncludeBenDemystifier" BeforeTargets="BeforeBuild" Condition="'$(PublishAot)' != 'true'">
```
vaind, line 26: `PublishAot` here in a library package is confusing to me because we're not actually publishing Sentry as a native AOT library. Or at least I don't think we need to? 🤔 

Maybe we could define a condition based on the solution name instead?

src/Sentry/Sentry.csproj
```diff
@@ -8,7 +8,7 @@
   </PropertyGroup>
 
   <PropertyGroup Condition="'$(SolutionName)' != 'Sentry.Unity'">
-    <TargetFrameworks>net6.0;net5.0;netcoreapp3.0;netstandard2.1;netstandard2.0;net461</TargetFrameworks>
+    <TargetFrameworks>net8.0;net6.0;net5.0;netcoreapp3.0;netstandard2.1;netstandard2.0;net461</TargetFrameworks>
     <TargetFrameworks Condition="'$(NO_ANDROID)' == ''">$(TargetFrameworks);net7.0-android</TargetFrameworks>
     <TargetFrameworks Condition="'$(NO_IOS)' == '' And $([MSBuild]::IsOSPlatform('OSX'))">$(TargetFrameworks);net7.0-ios</TargetFrameworks>
     <TargetFrameworks Condition="'$(NO_MACCATALYST)' == '' And $([MSBuild]::IsOSPlatform('OSX'))">$(TargetFrameworks);net7.0-maccatalyst</TargetFrameworks>
```
vaind, line 14: Just out of curiosity, why do we need to target net8.0 specifically? How about we limit this to the AOT solution? something like:

```suggestion
    <TargetFrameworks>net6.0;net5.0;netcoreapp3.0;netstandard2.1;netstandard2.0;net461</TargetFrameworks>
    <TargetFrameworks Condition="'$(NO_ANDROID)' == ''">$(TargetFrameworks);net7.0-android</TargetFrameworks>
    <TargetFrameworks Condition="'$(NO_IOS)' == '' And $([MSBuild]::IsOSPlatform('OSX'))">$(TargetFrameworks);net7.0-ios</TargetFrameworks>
    <TargetFrameworks Condition="'$(NO_MACCATALYST)' == '' And $([MSBuild]::IsOSPlatform('OSX'))">$(TargetFrameworks);net7.0-maccatalyst</TargetFrameworks>

    <!-- For AOT, we only build a library targeting at least .net 7.0 -->
    <TargetFrameworks Conditon="'$(SolutionName)' == 'Sentry.Aot'">net7.0</TargetFrameworks>
```

SentryAot.slnf
```diff
@@ -0,0 +1,12 @@
+{
+  "solution": {
+    "path": "Sentry.Aot.sln",
+    "projects": [
+      "samples\\Sentry.Samples.Console.Basic\\Sentry.Samples.Console.Basic.csproj",
+      "src\\Sentry\\Sentry.csproj",
```
vaind, line 6: Is there already a TODO/issue somewhere that we would need to combine this into the main solution nuget package? Hopefully using the following works: 
* https://learn.microsoft.com/en-us/nuget/create-packages/supporting-multiple-target-frameworks 
* and https://github.com/dotnet/runtime/blob/main/src/libraries/Microsoft.NETCore.Platforms/src/runtime.json#L1043

src/Sentry/Sentry.csproj
```diff
@@ -23,21 +23,24 @@
...
-  </ItemGroup>
+  <Target Name="IncludeBenDemystifier" BeforeTargets="BeforeBuild" Condition="'$(PublishAot)' != 'true'">
+    <!--
+      Ben.Demystifier is compiled directly into Sentry.
+      Note: It uses Microsoft.Bcl.AsyncInterfaces, which we get transitively from System.Text.Json.
+    -->
+    <ItemGroup>
+      <Compile Include="..\..\modules\Ben.Demystifier\src\**\*.cs">
+<!-- TODO: Work out why we get ' Error NETSDK1022 : Duplicate 'Compile' items were included.' if we uncomment this -->
+<!--        <Link>%(RecursiveDir)\%(Filename)%(Extension)</Link>-->
```
vaind, line 34: Can this be addressed before merging so that it doesn't get lost? Or has this always been the case and it has nothing to do with this PR?

src/Sentry/Internal/StackFrame.cs
```diff
@@ -113,7 +113,12 @@ public RealStackFrame(StackFrame frame)
 
     public int GetILOffset() => _frame.GetILOffset();
 
-    public MethodBase? GetMethod() => _frame.GetMethod();
+    public MethodBase? GetMethod() =>
+#if TRIMMABLE
+        null;
+#else
+        _frame.GetMethod();
+#endif
```
vaind, line 121: Why conditional though? This has worked before (may have returned null but at least it's not a custom code on our end.

test/Sentry.Tests/ModuleInit.cs
```diff
@@ -2,8 +2,17 @@
...
     [ModuleInitializer]
+    [SuppressMessage("Usage", "CA2255:The \'ModuleInitializer\' attribute should not be used in libraries")]
     public static void Init()
     {
+#if TRIMMABLE
+        Verifier.DerivePathInfo((sourceFile, projectDirectory, type, method) => new(
+            directory: projectDirectory,
+            typeName: type.Name,
+            // This ensures a unique name for our verify files when trimming is enabled
+            methodName: method.Name + ".AOT"
```
vaind, line 13: Does this need we have to have all these new "AOT.verified" files? Do we really need that or can we get away to have these files if necessary,  the same way we do when there are variants for each .net framework version? 

As is in the current state of this PR, it's too hard to see whether there are changes in these files between AOT and non-AOT version - it would require the reviewer to individually open each of the new .verified file and compare it to the previous version committed to the repository, without that suffix.

I believe it is quite important to see whether there are differences between AOT and non-AOT outputs so I'd strongly prefer if we could not add new files unless the changes are there.

vaind (commented):
This looks very good, though there are some open questions.

src/Sentry/PlatformAbstractions/RuntimeInfo.cs
jamescrosswell: But that's just the runtime that things were compiled with right? And if we bake that into our library, all it tells us is the runtime that the Sentry NuGet packages were compiled with (the user application that is making use of our SentrySDK might be built using a more recent runtime). the "actual" runtime is just native assembly for AOT apps right?

vaind: I don't believe so. AFAICT the sentry.dll is built as a standard library and is only later packaged into the AOT published app.

src/Directory.Build.props
vaind: yes, that makes sense.

src/Sentry/Sentry.csproj
jamescrosswell: We could try to make it work for .NET 7 but there's quite a bit of stuff that was added in .NET 8 to help with things like serialization. I'm not sure how much of that can be used in .NET 7 simply by including the new .NET 8 nuget packages as dependencies. For `version 1` of this though, to make things simple, I figured I'd try to get it working with .NET 8 and we could extend that to see if we can make it work for .NET 7 at some point further down the road (once we've also got all the other sentry packages supporting AOT).

src/Directory.Build.props
vaind: See the following log, specifically the sentry.dll compilation part. Warnings removed for brewity because this log is from my branch which doesn't fix the warnings yet.

```
dotnet publish ./samples/Sentry.Samples.Console.Basic -c release -f net8.0                
MSBuild version 17.8.0+6cdef4241 for .NET
/usr/local/share/dotnet/sdk/8.0.100-rc.2.23502.2/Sdks/Microsoft.NET.Sdk/targets/Microsoft.NET.RuntimeIdentifierInference.targets(311,5): message NETSDK1057: You are using a preview version of .NET. See: https://aka.ms/dotnet-support-policy [/Users/ivan/dev/sentry-dotnet/samples/Sentry.Samples.Console.Basic/Sentry.Samples.Console.Basic.csproj::TargetFramework=net8.0]
  Sentry -> /Users/ivan/dev/sentry-dotnet/src/Sentry/bin/release/net6.0/Sentry.dll
  Sentry.Samples.Console.Basic -> /Users/ivan/dev/sentry-dotnet/samples/Sentry.Samples.Console.Basic/bin/release/net8.0/osx-arm64/Sentry.Samples.Console.Basic.dll
  Generating native code
....
  Sentry.Samples.Console.Basic -> /Users/ivan/dev/sentry-dotnet/samples/Sentry.Samples.Console.Basic/bin/release/net8.0/osx-arm64/publish/
```

src/Sentry/Sentry.csproj
vaind: makes sense

SentryAot.slnf
jamescrosswell: I'm not sure I understand. `Sentry.Aot.sln` is a copy of `Sentry.sln`. The only reason it exists is so that we can  do things like this:
```
<PropertyGroup Condition="'$(SolutionName)' == 'Sentry.Aot'">
```

The only other ways I could think of to do that would be:
- We always build/test AOT for some specific version of .net (e.g. `net8.0`)
  - Not ideal, as then we couldn't also have managed tests against the `net8.0` 
- Pass in a flag from the command line or an environment variable
  - Would work OK in CI but not ideal for developers in an IDE
  
Do you think we could use some combination of the things you've referenced above to achieve the same, more simply?

src/Sentry/Sentry.csproj
jamescrosswell: No, I introduced that. It took me about a day to find that and I don't understand it at all. Logically, that section of the csproj file shouldn't be included when `PublishAot == true`.. but it is. I assume it relates to some arcane details of MS Build that I don't know about and I couldn't find any explanation in my googling attempts. 
1. I'd like to know what the implication of commenting this out would be. 
2. I'd like to know why we need to comment it out to get things to build properly. 
   - Possibly with a minimum reproducible example, we could get an answer from the team that works on MSBuild...

test/Sentry.Tests/ModuleInit.cs
jamescrosswell: That sounds great if we could find a way to do that. However Verify doesn't have any explicit support for AOT (in the same way that [it does for different target frameworks and architectures etc.](https://github.com/VerifyTests/Verify/blob/main/docs/naming.md#uniquefor)).

We could [follow up with Simon](https://github.com/orgs/VerifyTests/discussions/1029).

SentryAot.slnf
vaind: I think we're talking about orthogonal problems. What you're saying is all true and I generally agree with the solution. What I've meant by the comment is that we'll need a followup to make sure the sentry nuget package includes sentry.dll for both solutions. I think we'll need to do this manually (i.e. have a script)

src/Sentry/GraphQLRequestContent.cs
```diff
@@ -82,3 +81,66 @@ public GraphQLRequestContent(string? requestContent, SentryOptions? options = nu
     /// </summary>
     public string OperationTypeOrFallback() => OperationType ?? "graphql.operation";
 }
+
+/// <summary>
+/// Adapted from https://github.com/graphql-dotnet/graphql-dotnet/blob/42a299e77748ec588bf34c33334e985098563298/src/GraphQL.SystemTextJson/GraphQLRequestJsonConverter.cs#L64
+/// </summary>
```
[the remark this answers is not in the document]
jamescrosswell: Yeah I've done that before for things like ILSpy where we'd made minimal changes to the code (namespaces mostly). 

In this case, the original code is quite different. My main reason for referencing the original code was more incase it changed substantially and our GraphQL integration broke than about attribution. If we think the two bits of code are similar enough to warrant attribution, I can definitely do that though.

test/Sentry.Tests/ModuleInit.cs
vaind: Can we annotate test cases that need to have AOT specific golden files with a custom annotation and then check it here with `method.GetCustomAttribute` or some such?

SentryAot.slnf
jamescrosswell: Ah I see... because of all the #ifdef stuff we've got going on, we can't have a single package that is both trimmable and not trimmable. Do you know how NuGet would know which dll to select?

test/Sentry.Tests/ModuleInit.cs
jamescrosswell: That should be possible yes... which would at least reduce the number of `AOT.verified` files to just those that are required/different from their managed counterparts. 

I think there's also another issue we need to address at the same time, which is that CI doesn't currently run `dotnet test SentryAot.slnf` so it doesn't check the `AOT.verified` files at the moment. That's a simple tweak to our `build.yml` workflow.

SentryAot.slnf
vaind: I think nuget would pick the right one if we structure the libraries correctly, i.e. put it in `/runtimes/aot/net8.0` or somewhat like that. But that's just my best guess from the documentation, I haven't tested it yet.

src/Sentry/GraphQLRequestContent.cs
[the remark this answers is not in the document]
jamescrosswell: I've added an attribution for this.

src/Directory.Build.props
jamescrosswell: OK I'm not sure if using compiler directives to include/exclude stuff based on `#TRIMMABLE` will work at all. 

I tried ripping out the property group above and put this instead in `buildTransitive/Sentry.targets`:
```
<Target Name="SetTrimmableConstant" BeforeTargets="BeforeBuild" Condition="'$(PublishAot)' != 'true'">
  <PropertyGroup Condition="'$(PublishAot)' != 'true'">
    <DefineConstants>$(DefineConstants);TRIMMABLE</DefineConstants>
  </PropertyGroup>
</Target>
```

Then I packed that NuGet package and included it in a shinny new Console App (from a [local Nuget store](https://weblog.west-wind.com/posts/2022/Sep/11/Referencing-a-Local-Private-NuGet-Package-in-your-Solution)) with `<PublishAot>true</PublishAot>`. When I published that console app I got a whole bunch of trimming warnings from the Sentry.dll 😞 
 
We could, of course, take out all the `#TRIMMABLE` directives and just make those code paths the only code paths... i.e. Sentry is always trimmable. There are quite a few implications to that though:
- No enhanced stack traces
- No serializing anonymous/dynamic types in the Sentry protocol
- No automatic wiring up unhandled exception handlers for for WinUI applications
- Presumably a bunch more once we take trimming/aot beyond Sentry.csproj

🤔

vaind: It should not be in buildTransitive. The way it was before in this PR made sense and should work. You'll get a sentry.dll with the compiler directive set correctly and then use that DLL to compile the final app

vaind: I believe it's enough, for this PR, to concentrate on the sample apps included in this repo. Even just the console ones. If we get them to publish without warnings, we can make it work with nuget too.

jamescrosswell: That makes sense. I'll resolve this conversation and create a new issue from the [comment below](https://github.com/getsentry/sentry-dotnet/pull/2732#discussion_r1368297025).

jamescrosswell: Created https://github.com/getsentry/sentry-dotnet/issues/2759 for this.

SentryAot.slnf
jamescrosswell: Created https://github.com/getsentry/sentry-dotnet/issues/2759 for this.

src/Sentry/Internal/StackFrame.cs
jamescrosswell: Hm... OK this does actually cause us problems.

Because this now always returns `null` when `#TRIMMABLE` is set, we no longer get managed stack frames returned when running tests in `SentryAot.slnf`... which is kind of what we expected - in that context (compiling trimmable code designed to be used with applications being published to AOT, we're not going to be able to use methods like this that rely on reflection so our managed stack traces won't work). 

However, we're also not able to get Native stack frames. When running tests in `SentryAot.slnf`, this returns null :
https://github.com/getsentry/sentry-dotnet/blob/e6bedf6b3959264a5822dc8ea5af9dd61bd0b401/src/Sentry/Internal/DebugStackTrace.cs#L237-L242

Which in turn is because here we just delegate to the dotnet extension method:
https://github.com/getsentry/sentry-dotnet/blob/7000964395cddd5107b954131bdc1d74970c3ff4/src/Sentry/Internal/StackFrame.cs#L128

And that [will always return false](https://github.com/dotnet/runtime/blob/main/src/libraries/System.Private.CoreLib/src/System/Diagnostics/StackFrameExtensions.cs):
```csharp
    public static partial class StackFrameExtensions
    {
        public static bool HasNativeImage(this StackFrame stackFrame)
        {
            return stackFrame.GetNativeImageBase() != IntPtr.Zero;
        }
        ...
        public static IntPtr GetNativeImageBase(this StackFrame stackFrame)
        {
            return IntPtr.Zero;
        }
    }
```

I think that's also reasonable. Although we've set the `#TRIMMABLE` constant, the xUnit test runner that is running this test won't actually be published to AOT and compiled to native code... so I would expect `HasNativeImage` to return false in that context. 

I think utlimately there are aspects of the fuctionality that we're building which cannot be tested from xUnit. The only way to test this stuff would be to publish an AOT app that ran various tests (most likely not using xUnit, as I don't think it was designed for that. 

I can't think of a way to [get this working](https://github.com/getsentry/sentry-dotnet/blob/2a8813ab6725caac4bc3a76f71af542ffce7f588/test/Sentry.Tests/Internals/MainExceptionProcessorTests.cs#L120) in xUnit, for example. 🤔 
```suggestion
    public MethodBase? GetMethod() =>
#if TRIMMABLE
        null;
#else
        _frame.GetMethod();
#endif
```

src/Sentry/Internal/DebugStackTrace.cs
jamescrosswell: Actually, not sure what I was talking about wrt the other ifdefs. Have just updated 👍🏻

jamescrosswell: Have added a separate issue to refactor the `DebugStackTrace` class: https://github.com/getsentry/sentry-dotnet/issues/2761

src/Sentry/Sentry.csproj
jamescrosswell: Found a workaround to this.

src/Sentry/Internal/StackFrame.cs
vaind: I've linked this from the other PR for Trimming for net7.0 but let's reiterate: currently,  there's no xUnit runner to support AOT so no way to test this in unit tests. There's a tracking issue: https://github.com/dotnet/runtime/issues/91069

I second that for now we need to do an integration test for AOT stack frames and rely on unit tests on the managed solution.

src/Sentry/Internal/Extensions/JsonExtensions.cs
```diff
@@ -14,40 +14,84 @@ internal static class JsonExtensions
...
+        }
+        foreach (var converter in CustomConverters)
+        {
+            options.Converters.Add(converter);
+        }
+
+        return options;
+    }
+
+#if TRIMMABLE
```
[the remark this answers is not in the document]
vaind: worth pulling another tracking issue?

scripts/generate-solution-filters-config.yaml
```diff
@@ -170,3 +176,13 @@ filterConfigs:
...
+
+  - outputPath: SentryAot.slnf
+    solution: Sentry.Aot.sln
+    include:
+      patterns:
+        - "**/Sentry.Samples.Console.Basic.csproj"
+        - "**/Sentry.csproj"
+        - "**/Sentry.Testing.csproj"
+        - "**/Sentry.Testing.CrashableApp.csproj"
+        - "**/Sentry.Tests.csproj"
```
bruno-garcia, line 188:
```suggestion
        - "**/Sentry.Tests.csproj"

```

scripts/generate-solution-filters.ps1
```diff
@@ -138,7 +138,9 @@ foreach($filter in $config.filterConfigs){
...
-# Update solution files from Sentry.sln
-$source = Join-Path $repoRoot "Sentry.sln"
-$destination = Join-Path $repoRoot "Sentry.NoMobile.sln"
-Copy-Item -Path $source -Destination $destination -Force
+# Copy the Core solution to each of the required build solutions
+$source = Join-Path $repoRoot $config.coreSolution
+foreach ($buildSolution in $config.buildSolutions) {
+  $destination = Join-Path $repoRoot $buildSolution
+  Copy-Item -Path $source -Destination $destination -Force
+}
```
bruno-garcia, line 146:
```suggestion
}

```

src/Directory.Build.props
bruno-garcia: net9 comes out next year, and net10 after that. So checking versionOrLater is ideal

src/Sentry/Integrations/WinUIUnhandledExceptionIntegration.cs
```diff
@@ -73,6 +75,11 @@ public void Register(IHub hub, SentryOptions options)
         });
     }
 
+#if !TRIMMABLE
+    /// <summary>
```
bruno-garcia, line 79: This might have been fixed since this hack was done. @lucas-zimerman might know if there's a non-reflection way around this.

bruno-garcia: I think we'll need to document what features are not supported in an AOT scenario then:

* WinUI apps won't have unhandled exceptions?

bruno-garcia: OK just realized this is on the PR desc, so it'll be easier to write on the CHANGELOG and later on the docs

bruno-garcia (commented):
Ivan's note make sense to me, we can't conditionally do things for AOT in Sentry's build since we're shipping only IL.
The final consumer of this library might be targeting JIT or AOT.

Which brings all of the `#ifdef` in this PR in question, since all of those are compile flags and we're shopping only a single version. Or if we provided sources, to be compiled by the end user

src/Sentry/Sentry.csproj
```diff
@@ -8,7 +8,7 @@
   </PropertyGroup>
 
   <PropertyGroup Condition="'$(SolutionName)' != 'Sentry.Unity'">
-    <TargetFrameworks>net6.0;net5.0;netcoreapp3.0;netstandard2.1;netstandard2.0;net461</TargetFrameworks>
+    <TargetFrameworks>net8.0;net6.0;net5.0;netcoreapp3.0;netstandard2.1;netstandard2.0;net461</TargetFrameworks>
```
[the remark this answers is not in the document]
vaind: Also, do we need to add net8.0 at all? Because I think this was a recent addition in this PR so if it's just for sake of adding it, I'd prefer not to because it makes builds even slower.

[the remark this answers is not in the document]
vaind: I've removed this temporarily because of test failures in CI. If we need it (in this PR), feel free to bring it back.

test/CommonModuleInit.cs
```diff
@@ -2,7 +2,29 @@
...
+#if TRIMMABLE
+        PathInfo DeriveAotPathInfo(string sourceFile, string projectDirectory, Type type, MethodInfo method)
+        {
+            var uniqueForAot = method.GetCustomAttribute<UniqueForAotAttribute>()
+                            ?? type.GetCustomAttribute<UniqueForAotAttribute>();
+            var derivedMethodName = (uniqueForAot != null)
+                ? method.Name + ".AOT"
+                : method.Name;
+            return new PathInfo(
+                directory: projectDirectory,
```
vaind, line 17: This seems wrong - all ".AOT.verified" files are put in the root test directory instead of next to tests.

test/Sentry.Tests/ContextCopyTests.CopyTo_Basic.AOT.verified.txt
```diff
@@ -0,0 +1,15 @@
+﻿{
```
vaind, line 1: these look the same to me as the ones without AOT

vaind (commented):
1. I've fixed the issues causing CLI integration test failures. Main takeaway is we cannot pass /p:PublishAot=true in the integration test itself because it applies to the whole build, including dependencies (i.e. Sentry)
2. I've noticed you've added the new verifier annotation for AOT but the random files I've picked and checked are still the same as non-AOT .verified files, can you please clean this up so we can review?
3. I think we need to actually check whether we can package two versions of the same compiled library, one for aot and one for non-aot into a nugget package. Otherwise the whole multi-solution approach here won't work

test/CommonModuleInit.cs
vaind: example: ContextCopyTests

vaind:
> 3\. I think we need to actually check whether we can package two versions of the same compiled library, one for aot and one for non-aot into a nugget package. Otherwise the whole multi-solution approach here won't work

I've tried doing this (based on the links in #2759) but didn't get the app to work (yet). I've created a nuget package manually by copying the "AOT" variant to `runtimes/lib/net6.0/sentry.dll` (and adding `ref/` so that it compiles. While it does compile, it fails during app startup with:
```
Unhandled Exception: System.IO.FileNotFoundException: Could not find file 'Sentry'.
File name: 'Sentry'
   at Internal.Runtime.TypeLoaderExceptionHelper.CreateFileNotFoundException(ExceptionStringID, String) + 0x43
   at Internal.Runtime.CompilerHelpers.ThrowHelpers.ThrowFileNotFoundException(ExceptionStringID, String) + 0x9
   at Program.<<Main>$>d__0.MoveNext() + 0x15
   at System.Runtime.CompilerServices.AsyncMethodBuilderCore.Start[TStateMachine](TStateMachine&) + 0x3f
   at Program.<Main>$(String[]) + 0x2a
   at Program.<Main>(String[] args) + 0x9
   at Sentry.Samples.Console.Basic!<BaseAddress>+0xd1290
PS C:\dev\temp\Sentry.Samples.Console.Basic> 
```

I'll reiterate: I really think we should reduce scope here and just make JSON work, no #ifdef, just make it **not throw exceptions** when running in AOT. Even if users get warnings at compile time, Sentry would work. We can than fix warnings in the next PR.

jamescrosswell:
At the moment we have this in the src/Directory.Build.props file:

```xml
  <!-- AOT not supported for osx-arm64 on .NET 7 -->
  <PropertyGroup Condition="'$(SolutionName)' == 'Sentry.Aot' and $(TargetFramework.StartsWith('net8'))">
    <PublishAot>true</PublishAot>
  </PropertyGroup>
  <!-- Kept separate in case PublishAot is passed in as a build parameter from the command line -->
  <PropertyGroup Condition="'$(PublishAot)' == 'true'">
    <DefineConstants>$(DefineConstants);TRIMMABLE</DefineConstants>
  </PropertyGroup>
```

This enables us to retain/tests two separate "branches" of our code - one supporting AOT and the other which would only work with JIT compilation (where reflection is still available).

This has ben useful for testing and has enabled us to delay making any big calls about dropping features from
Sentry that don't work with AOT (e.g. Enhanced Stack Traces) but it's not how Microsoft intended for library authors to enable their packages for AOT. If we retained the above, we'd have to compile, publish maintain and support two separate versions of the Sentry packages, for as long as we retain the `#ifdef` statements 😢 

What we probably should be using instead is the `<IsAotCompatible>true</IsAotCompatible>` attribute:
* [AOT-compatibility analyzers](https://learn.microsoft.com/en-us/dotnet/core/deploying/native-aot/?tabs=net8plus%2Cwindows#aot-compatibility-analyzers)

This attribute would set all of these things to true:
- IsTrimmable
- EnableTrimAnalyzer
- EnableSingleFileAnalyzer
- EnableAotAnalyzer

Essentially we'd get all the same compiler warnings and errors that we get when defining `PublishAot`... if we can
compile Sentry and all our unit tests pass with `<IsAotCompatible>true</IsAotCompatible>` it should work for customers
using Sentry in applications that use AOT compilation.

If we don't want to compile/maintain/support two separate versions of Sentry (which I think would be pretty difficult),
realistically we've only got a few options:
1. Customers using Sentry can't use AOT compilation with their apps (straw man - I don't think we want this)
2. We drop support for features that are incompatible with AOT compilation (not great, but last resort)
3. We try to execute AOT compatible code and fall back to JIT where this is possible
     - For serialization this should be fairly straight forward. We might have to disable some warnings for code that we  know won't be executed when using AOT compilation.
     - For things like Enhanced Stack Traces, we'll have to play around a bit to see if we can find a similar solution...

vaind:
> if we can compile Sentry and all our unit tests pass with `<IsAotCompatible>true</IsAotCompatible>` it should work for customers using Sentry in applications that use AOT compilation.

We don't necessarily need to do this for customers' code to work - it will still work even if the library triggers warnings, as long as we can make sure the runtime code the warnings are about works fine (e.g. with the try-catch with fallbacks)

jamescrosswell:
> > if we can compile Sentry and all our unit tests pass with `<IsAotCompatible>true</IsAotCompatible>` it should work for customers using Sentry in applications that use AOT compilation.
> 
> We don't necessarily need to do this for customers' code to work - it will still work even if the library triggers warnings, as long as we can make sure the runtime code the warnings are about works fine (e.g. with the try-catch with fallbacks)

That's true but I'd definitely rather rely on the compiler to detect issues. Given that we can't use unit tests to detect these, the only other option would be integration tests, which are very expensive to build and maintain. So if there's any way we can leverage that `IsAotCompatible` property, I think we should.

vaind:
> So if there's any way we can leverage that `IsAotCompatible` property, I think we should.

I agree completely

vaind (commented):
I've looked into this somewhat. Do we still need ifdefs? I thought we're getting rid of those at first to just get the JSON working.

jamescrosswell:
> I've looked into this somewhat. Do we still need ifdefs? I thought we're getting rid of those at first to just get the JSON working.

We don't... I'd left them in the tests as a reminder for where we needed integration tests. I've removed the ifdefs and replaced these with comments in the test projects.

src/Sentry/Sentry.csproj
```diff
@@ -70,9 +77,9 @@
     <Reference Include="System.Net.Http" />
   </ItemGroup>
 
-  <!-- System.Text.Json is already included with .NET 5 and higher. Add a direct reference for other targets. -->
-  <ItemGroup Condition="!$(TargetFramework.StartsWith('net5')) and !$(TargetFramework.StartsWith('net6'))">
-    <PackageReference Include="System.Text.Json" Version="5.0.2" />
+  <!-- Ensure at least version 6 of System.Text.Json so we have JsonSerializationContext available -->
+  <ItemGroup Condition="!$(TargetFramework.StartsWith('net6')) and !$(TargetFramework.StartsWith('net7')) and !$(TargetFramework.StartsWith('net8'))">
+    <PackageReference Include="System.Text.Json" Version="6.0.8" />
```
vaind, line 82: How come the original dependency is now not necessary for .net framework **lower than** net5? That's whet the original piece of the code was fore, right? I understand the new code, just don't get why we could remove the previous one. I'd have expected we need both.

test/Directory.Build.props
```diff
@@ -7,6 +7,13 @@
     <SuppressTfmSupportBuildWarnings>true</SuppressTfmSupportBuildWarnings>
     <IsPackable>false</IsPackable>
     <IsTestProject>true</IsTestProject>
+    <!-- Ignore warnings for generated code -->
```
vaind, line 10:
```suggestion
    <!-- Ignore warnings for JSON generated code -->
```

test/Directory.Build.props
```diff
@@ -7,6 +7,13 @@
...
+    <NoWarn>$(NoWarn);SYSLIB0005;SYSLIB0012</NoWarn>
+  </PropertyGroup>
+
+  <!-- Never use Sentry CLI for test projects. -->
+  <PropertyGroup>
+    <UseSentryCLI>false</UseSentryCLI>
   </PropertyGroup>
```
vaind, line 17: This has been removed on 4.0.0 branch in the meantime, must be a merge issue.

```suggestion
```

test/Sentry.Tests/Internals/DebugStackTraceTests.verify.cs
```diff
@@ -17,6 +17,8 @@ private class Fixture
     private readonly Fixture _fixture = new();
     private static readonly string ThisNamespace = typeof(SentryStackTraceFactoryTests).Namespace!;
 
+// TODO: Create integration test to test this behaviour when publishing AOT apps
+// See https://github.com/getsentry/sentry-dotnet/pull/2732#discussion_r1371006441
```
vaind, line 21:
```suggestion
// See https://github.com/getsentry/sentry-dotnet/issues/2772
```

test/Sentry.Tests/Internals/DebugStackTraceTests.verify.cs
```diff
@@ -28,6 +30,8 @@ public void CreateSentryStackFrame_AppNamespace_InAppFrame()
         Assert.True(actual?.InApp);
     }
 
+// TODO: Create integration test to test this behaviour when publishing AOT apps
+// See https://github.com/getsentry/sentry-dotnet/pull/2732#discussion_r1371006441
```
vaind, line 34:
```suggestion
// See https://github.com/getsentry/sentry-dotnet/issues/2772
```

test/Sentry.Tests/Internals/DebugStackTraceTests.verify.cs
```diff
@@ -40,6 +44,8 @@ public void CreateSentryStackFrame_AppNamespaceExcluded_NotInAppFrame()
         Assert.False(actual?.InApp);
     }
 
+// TODO: Create integration test to test this behaviour when publishing AOT apps
+// See https://github.com/getsentry/sentry-dotnet/pull/2732#discussion_r1371006441
```
vaind, line 48:
```suggestion
// See https://github.com/getsentry/sentry-dotnet/issues/2772
```

test/Sentry.Tests/Internals/DebugStackTraceTests.verify.cs
```diff
@@ -62,6 +68,8 @@ public void CreateSentryStackFrame_SystemType_NotInAppFrame(bool useEnhancedStac
...
```
vaind, line 72:
```suggestion
// See https://github.com/getsentry/sentry-dotnet/issues/2772
```

test/Sentry.Tests/Internals/DebugStackTraceTests.verify.cs
```diff
@@ -206,6 +214,8 @@ public void ParseNativeAOTToString()
         Assert.Null(frame.Package);
     }
 
+// TODO: Create integration test to test this behaviour when publishing AOT apps
+// See https://github.com/getsentry/sentry-dotnet/pull/2732#discussion_r1371006441
```
vaind, line 218:
```suggestion
// See https://github.com/getsentry/sentry-dotnet/issues/2772
```

test/Sentry.Tests/Internals/MainExceptionProcessorTests.cs
```diff
@@ -117,6 +117,8 @@ public void Process_AggregateException()
         sut.Process(BuildAggregateException(), evt);
 
         var last = evt.SentryExceptions!.Last();
+// TODO: Create integration test to test this behaviour when publishing AOT apps
+// See https://github.com/getsentry/sentry-dotnet/pull/2732#discussion_r1371006441
```
vaind, line 121:
```suggestion
// See https://github.com/getsentry/sentry-dotnet/issues/2772
```

test/Sentry.Tests/SerializationTests.verify.cs
```diff
@@ -1,3 +1,7 @@
+#if NET7_0_OR_GREATER
```
vaind, line 1: Wouldn't net6 work here? That's what the following added code from above made me believe:

```
 <!-- Ensure at least version 6 of System.Text.Json so we have JsonSerializationContext available -->
  <ItemGroup Condition="!$(TargetFramework.StartsWith('net6')) and !$(TargetFramework.StartsWith('net7')) and !$(TargetFramework.StartsWith('net8'))">
    <PackageReference Include="System.Text.Json" Version="6.0.8" />
```

samples/Directory.Build.props
```diff
@@ -33,4 +33,8 @@
     <MtouchUseLlvm>false</MtouchUseLlvm>
   </PropertyGroup>
 
+  <!-- AOT not supported for osx-arm64 on .NET 7 -->
+  <PropertyGroup Condition="'$(SolutionName)' == 'Sentry.Aot' and (!$([MSBuild]::IsOSPlatform('OSX')) or '$(TargetFramework)' == 'net8.0')">
+    <PublishAot>true</PublishAot>
+  </PropertyGroup>
```
vaind, line 39: I still believe this should be done in each sample individually as needed. Not all samples support AOT publishing nor do we care since we don't normally run `dotnet publish` on them - we only do that in the CLI integration test for the basic console sample.
```suggestion
```

src/Directory.Build.props
```diff
@@ -35,6 +35,10 @@
     <CodeAnalysisRuleSet>$(MSBuildThisFileDirectory)CodeAnalysis.ruleset</CodeAnalysisRuleSet>
   </PropertyGroup>
 
+  <PropertyGroup Condition="'$(SolutionName)' == 'Sentry.Aot' and ($(TargetFramework.StartsWith('net6')) or $(TargetFramework.StartsWith('net7')) or $(TargetFramework.StartsWith('net8')))">
+    <IsAotCompatible>true</IsAotCompatible>
+  </PropertyGroup>
+
```
vaind, line 41: This seems unused at the moment
```suggestion
```

vaind (commented):
Some minor comments/change suggestions.

I'd still prefer the new solution to be removed (at least until we need it, which we don't in this PR) but that's just me.

I think someone else should review this PR too. @bruno-garcia @bitsandfoxes

.github/workflows/build.yml
```diff
@@ -94,3 +94,59 @@ jobs:
...
+    name: ${{ matrix.os }} AOT
+    runs-on: ${{ matrix.os }}
+    env:
+      AOT_SOLUTION: SentryAot.slnf
+
+    strategy:
+      fail-fast: false
+      matrix:
+        # Pinning `macos-13` because Microsoft.iOS 16.4 requires Xcode 14.3 which is only built-in in 13
+        os: [macos-13] # just macos at the moment
```
bruno-garcia, line 108: Should we build this also on Windows and Linux?

src/Directory.Build.props
```diff
@@ -35,6 +35,11 @@
     <CodeAnalysisRuleSet>$(MSBuildThisFileDirectory)CodeAnalysis.ruleset</CodeAnalysisRuleSet>
   </PropertyGroup>
 
+  <!-- Only compile with IsAotCompatible on .NET 6.0 and later. Our Device Tests fail for net7.0-android with this enabled as well -->
+  <PropertyGroup Condition="'$(TargetFramework)' != 'net7.0-android' AND $([System.Text.RegularExpressions.Regex]::IsMatch('$(TargetFramework)', '^(?!net5\.|net4\d{2}|netcoreapp|netstandard)net\d+'))">
```
vaind, line 39: Maybe the regexp could do without the negative lookahead
```suggestion
  <PropertyGroup Condition="'$(TargetFramework)' != 'net7.0-android' AND $([System.Text.RegularExpressions.Regex]::IsMatch('$(TargetFramework)', '^net[6-9]'))">
```

src/Sentry/Sentry.csproj
```diff
@@ -23,21 +23,32 @@
...
-  -->
-  <ItemGroup>
-    <Compile Include="..\..\modules\Ben.Demystifier\src\**\*.cs">
-      <Link>%(RecursiveDir)\%(Filename)%(Extension)</Link>
-    </Compile>
-    <Compile Remove="..\..\modules\Ben.Demystifier\**\obj\**" />
-  </ItemGroup>
+  <PropertyGroup Condition="'$(FrameworkSupportsAot)' == 'true'">
+    <IsAotCompatible>true</IsAotCompatible>
+  </PropertyGroup>
```
vaind, line 28: I did not expect we could set this - I've expected trimming errors so this is nice. 
Should we propagate this property to all projects under `/src/`?

vaind (approved):
LGTM with minor comments/ideas for improvement

vaind (changes requested):
Oh wait, we also need changelog entries for the breaking changes, right?

jamescrosswell:
> Oh wait, we also need changelog entries for these breaking changes, right?

Yeah we do... it's a bit tricky off the basis of this PR only. I suspect we might want to mention something about the symbolication and debug stack traces, and that might require details from some of the PRs you've been working on @vaind.

I've put something basic in there for now. It will need to be adapted once https://github.com/getsentry/sentry-dotnet/issues/2778 has been done and we might want to adapt it further and add references to other PRs relevant to AOT.

vaind (approved):

PR #2733, opened 2023-10-17
CI cleanup

bitsandfoxes:
The actual fix on top of the initial "make it go green".
Also, added the `SelfContained` from [here](https://learn.microsoft.com/en-us/dotnet/core/compatibility/sdk/8.0/runtimespecific-app-default) to the appropriate projects to get rid of:
```
warning NETSDK1201: For projects targeting .NET 8.0 and higher, specifying a RuntimeIdentifier will no longer produce a self contained app by default. To continue building self-contained apps, set the SelfContained property to true or use the --self-contained argument.
```
#skip-changelog

### DIFF
--- a/.github/actions/environment/action.yml
+++ b/.github/actions/environment/action.yml
@@ -17,14 +17,17 @@ runs:
         distribution: 'temurin'
         java-version: '11'
 
+    # https://github.com/actions/runner-images/issues/7667
+    # .NET 3.1 has been removed from all OS due to EOL
+    # .NET 6 and .NET 8 are not built-in with macos-13
     - name: Install .NET SDK
       uses: actions/setup-dotnet@v3
       with:
         dotnet-version: |
           3.1.x
-          6.0.x
+          6.0.x 
           7.0.x
-          8.0.100-rc.2.23502.2
+          8.0.x
         
     - name: Dependency Caching
       uses: actions/cache@v3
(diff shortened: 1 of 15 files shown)


samples/Sentry.Samples.Ios/Sentry.Samples.Ios.csproj
```diff
@@ -6,6 +6,7 @@
     <Nullable>enable</Nullable>
     <ImplicitUsings>true</ImplicitUsings>
     <SupportedOSPlatformVersion>10.0</SupportedOSPlatformVersion>
+    <SelfContained>true</SelfContained>
```
jamescrosswell, line 9: What's the thinking behind this?

.github/actions/environment/action.yml
```diff
@@ -44,17 +47,4 @@ runs:
...
-    - name: Update workloads
-      shell: bash
-      run: dotnet workload update
-
-    - name: Log what we have in the workloads
-      shell: bash
-      run: dotnet workload update
-      
\ No newline at end of file
+          --temp-dir "${{ runner.temp }}"
```
jamescrosswell, line 50: Nice! We don't need rollback.json anymore?

jamescrosswell (approved):
Looks good. Questions to satisfy my curiosity.

samples/Sentry.Samples.Ios/Sentry.Samples.Ios.csproj
bitsandfoxes: Apparently, there was an implicit thing going on that `RuntimeIdetifier` would lead to a self-contained app (due to being runtime-specific). With .NET 8 that is no longer the case and requires the `SelfContained` property.

.github/actions/environment/action.yml
bitsandfoxes: Yess!

bruno-garcia:
This could ahve been called "drop tizen support"

PR #2736, opened 2023-10-18
Fix no mobile

jamescrosswell:
##skip-changelog

When we started [generating solution filters automatically](https://github.com/getsentry/sentry-dotnet/pull/2576) from a single `Sentry.sln` solution file, we lost one quite important convenience. Our previous build process maintained separate solution files that were used in our solution filters to determine whether or not mobile targets should be enabled:

image

This PR re-enables that same behaviour but **_still lets us maintain a single Sentry.sln solution file_**. 

Once this PR is merged into main, we should be able to merge it into the `feat/4.0.0` branches and use a similar technique to maintain solution filters that can be used to test and build AOT packages by setting `<PublishAot>true</PublishAot>` in the build props if the solution is `Sentry.Native.sln` (or something like that).

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -33,7 +33,7 @@ jobs:
 
       - name: Ensure Solution Filters are up to date
         shell: pwsh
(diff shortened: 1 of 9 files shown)


bitsandfoxes (approved):
🔥

PR #2739, opened 2023-10-18
Feat: native AOT debug info upload

### LINKED ISSUE #2692: Native AOT - upload debug files and sources
We should actually upload the contents of the "publish" directory, not the intermediary PDBs

vaind:
closes #2692

This is mostly done in `Sentry.targets` - I've merged the two jobs (one uploading debug files, the second one uploading sources) because it was getting unwieldy when trying to handle both native AOT and managed symbols (and referenced sources) in both jobs. 

I've also had to rework the PrepareCLI workflow a bit so that we get consistent integration test runs. Ppreviously, the integration test build would also upload symbols for the Sentry project itself, which is not what would happen when building actual user projects, because that project would have come from a nuget package.

### TODO: 
* [ ] force some source bundles to be uploaded for managed apps? it seems all sources are embedded with .net 8 SDK... How can we do it though? I'm open to suggestions
* [x] test on Linux
* [x] test on macOS (although the actual native integration will only come as a part of #2625)

#skip-changelog (This is part of a bigger task to support Native AOT)

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -50,11 +50,6 @@
     <SentryUploadSources>true</SentryUploadSources>
   </PropertyGroup>
 
-  <!-- If running in CI and auth token is not set, disable SentryCLI completely (to avoid logging warnings). -->
-  <PropertyGroup Condition="'$(GITHUB_ACTIONS)' == 'true' And '$(SENTRY_AUTH_TOKEN)' == ''">
-    <UseSentryCLI>false</UseSentryCLI>
-  </PropertyGroup>
-
   <!--
     Note: The following platform-specific properties need to be set in both Directory.Build.props and DirectoryBuild.targets.
     TODO: Figure out how to consolidate to a single location.
--- a/samples/Sentry.Samples.Console.Basic/Sentry.Samples.Console.Basic.csproj
+++ b/samples/Sentry.Samples.Console.Basic/Sentry.Samples.Console.Basic.csproj
@@ -4,7 +4,7 @@
     <OutputType>Exe</OutputType>
     <ImplicitUsings>enable</ImplicitUsings>
     <Nullable>enable</Nullable>
-    <TargetFramework>net7.0</TargetFramework>
+    <TargetFrameworks>net7.0;net8.0</TargetFrameworks>
     <PublishAot>true</PublishAot>
   </PropertyGroup>
 
--- a/samples/Sentry.Samples.Console.Basic/Sentry.Samples.Console.Basic.csproj
+++ b/samples/Sentry.Samples.Console.Basic/Sentry.Samples.Console.Basic.csproj
@@ -7,6 +7,7 @@
     <TargetFrameworks>net7.0;net8.0</TargetFrameworks>
 
     <!-- AOT not supported for osx-arm64 on .NET 7 -->
+	<!-- TODO, consider updating after changes from #2732 come into effect -->
     <PublishAot Condition="!$([MSBuild]::IsOSPlatform('OSX')) or '$(TargetFramework)' == 'net8.0'">true</PublishAot>
   </PropertyGroup>
 
--- a/samples/Sentry.Samples.Console.Basic/Sentry.Samples.Console.Basic.csproj
+++ b/samples/Sentry.Samples.Console.Basic/Sentry.Samples.Console.Basic.csproj
@@ -4,14 +4,6 @@
     <OutputType>Exe</OutputType>
     <ImplicitUsings>enable</ImplicitUsings>
     <Nullable>enable</Nullable>
-  </PropertyGroup>
-
-  <!-- Remove after switching to .NET 8 SDK -->
-  <PropertyGroup Condition="$([MSBuild]::IsOSPlatform('OSX'))">
-    <TargetFramework>net6.0</TargetFramework>
-    <PublishAot>false</PublishAot>
-  </PropertyGroup>
-  <PropertyGroup Condition="!$([MSBuild]::IsOSPlatform('OSX'))">
     <TargetFramework>net7.0</TargetFramework>
     <PublishAot>true</PublishAot>
   </PropertyGroup>
--- a/samples/Sentry.Samples.Console.Basic/Sentry.Samples.Console.Basic.csproj
+++ b/samples/Sentry.Samples.Console.Basic/Sentry.Samples.Console.Basic.csproj
@@ -5,7 +5,9 @@
     <ImplicitUsings>enable</ImplicitUsings>
     <Nullable>enable</Nullable>
     <TargetFrameworks>net7.0;net8.0</TargetFrameworks>
-    <PublishAot>true</PublishAot>
+
+    <!-- AOT not supported for osx-arm64 on .NET 7 -->
+    <PublishAot Condition="!$([MSBuild]::IsOSPlatform('OSX')) or '$(TargetFramework)' == 'net8.0'">true</PublishAot>
   </PropertyGroup>
 
   <PropertyGroup>
--- a/src/Sentry/buildTransitive/Sentry.targets
+++ b/src/Sentry/buildTransitive/Sentry.targets
@@ -149,9 +149,11 @@
     <SentryCLIUploadSymbolType Include="dsym" Condition="$([MSBuild]::GetTargetPlatformIdentifier('$(TargetFramework)')) == 'maccatalyst'" />
   </ItemGroup>
 
-  <!-- Upload symbols to Sentry after the build. -->
+  <!-- Upload symbols to Sentry after the build. 
+       We're explicitly skipping uploads for Sentry project because it would interfere with integration tests. 
+       Also, it only ever gets triggered for Sentry within this repository. -->
   <Target Name="UploadDebugInfoToSentry" AfterTargets="$(SentryCLIUploadAfterTargets)" DependsOnTargets="PrepareSentryCLI"
-    Condition="'$(SentryCLI)' != '' and ('$(SentryUploadSymbols)' == 'true' or '$(SentryUploadSources)' == 'true')">
+    Condition="'$(MSBuildProjectName)' != 'Sentry' and '$(SentryCLI)' != '' and ('$(SentryUploadSymbols)' == 'true' or '$(SentryUploadSources)' == 'true')">
 
     <!-- if (UploadSymbols && UploadSources) { -->
       <Message Importance="High"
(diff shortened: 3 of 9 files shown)


Directory.Build.props
```diff
@@ -50,11 +50,6 @@
     <SentryUploadSources>true</SentryUploadSources>
   </PropertyGroup>
 
-  <!-- If running in CI and auth token is not set, disable SentryCLI completely (to avoid logging warnings). -->
-  <PropertyGroup Condition="'$(GITHUB_ACTIONS)' == 'true' And '$(SENTRY_AUTH_TOKEN)' == ''">
-    <UseSentryCLI>false</UseSentryCLI>
-  </PropertyGroup>
```
vaind, line 56: This (and other similar cases below) is now handled by a condition in `Sentry.targets` property `UseSentryCLI`

samples/Sentry.Samples.Console.Basic/Sentry.Samples.Console.Basic.csproj
```diff
@@ -4,16 +4,10 @@
...
-  <!-- Remove after switching to .NET 8 SDK -->
-  <PropertyGroup Condition="$([MSBuild]::IsOSPlatform('OSX'))">
-    <TargetFramework>net6.0</TargetFramework>
-    <PublishAot>false</PublishAot>
-  </PropertyGroup>
-  <PropertyGroup Condition="!$([MSBuild]::IsOSPlatform('OSX'))">
-    <TargetFramework>net7.0</TargetFramework>
-    <PublishAot>true</PublishAot>
+    <!-- AOT not supported for osx-arm64 on .NET 7 -->
+    <PublishAot Condition="!$([MSBuild]::IsOSPlatform('OSX')) or '$(TargetFramework)' == 'net8.0'">true</PublishAot>
```
jamescrosswell, line 10: @vaind maybe worth putting a TODO in here. Once we get this branch merged with the net8-trimming branch, I think a better way to enable/disable AOT is simply to check whether we're buliding in the context of the `SentryAot.slnf` solution filter, like we do here:
https://github.com/getsentry/sentry-dotnet/blob/731c6829235022c936e7dd53cdcbfbfec389ba76/src/Directory.Build.props#L38-L41

... and here:
https://github.com/getsentry/sentry-dotnet/blob/731c6829235022c936e7dd53cdcbfbfec389ba76/test/Directory.Build.props#L17-L22

We should be able to do this once in the `Directory.Build.props` file in the samples directory.

src/Sentry/buildTransitive/Sentry.targets
```diff
@@ -34,8 +34,13 @@
...
 
-      <!-- This property controls if the Sentry CLI is to be used at all.  Setting false will disable all Sentry CLI usage. -->
-      <UseSentryCLI Condition="'$(UseSentryCLI)' == '' And ('$(SentryUploadSymbols)' == 'true' Or '$(SentryUploadSources)' == 'true' Or $(SentryUploadAndroidProguardMapping) == 'true')">true</UseSentryCLI>
+      <!-- This property controls if the Sentry CLI is to be used at all. Setting false will disable all Sentry CLI usage.
+           We're explicitly skipping uploads for Sentry projects (except samples) because they interfere with CLI integration test asserts. -->
+      <UseSentryCLI Condition="
+        '$(UseSentryCLI)' == ''
+        and ('$(SentryUploadSymbols)' == 'true' or '$(SentryUploadSources)' == 'true' or $(SentryUploadAndroidProguardMapping) == 'true')
+        and ('$(MSBuildProjectName)' != 'Sentry')
+        and (!$(MSBuildProjectName.StartsWith('Sentry.')) or '$(MSBuildProjectName)' == '$(SentryCLIIntegrationTestProject)')">true</UseSentryCLI>
```
jamescrosswell, line 43: `!$(MSBuildProjectName.StartsWith('Sentry.')` doesn't skip our samples?

src/Sentry/buildTransitive/Sentry.targets
```diff
@@ -120,53 +125,88 @@
...
-      <SentryCLIUploadSymbolType Include="dsym" Condition="$([MSBuild]::GetTargetPlatformIdentifier('$(TargetFramework)')) == 'maccatalyst'" />
-    </ItemGroup>
-
-    <Exec Command="$(SentryCLIDebugFilesUploadCommand) @(SentryCLIUploadSymbolType -> '-t %(Identity)', ' ') $(OutputPath)" IgnoreExitCode="true" ContinueOnError="WarnAndContinue">
-      <Output TaskParameter="ExitCode" PropertyName="_SentryCLIExitCode" />
-    </Exec>
+  <!-- Native AOT publishing
+       While '_IsPublishing' looks a bit "private", it's also OK if it suddenly stops working and this step runs anyway.
+       See https://github.com/dotnet/sdk/issues/26324#issuecomment-1169236993 -->
+  <PropertyGroup Condition="'$(PublishDir)' != '' and '$(PublishAot)' == 'true' and '$(_IsPublishing)' == 'true'">
```
jamescrosswell, line 131: I see we already have a `CI_PUBLISHING_BUILD` property... I take it this doesn't work because it's only ever set by by our github actions (and we want this to work even in the context of builds on local development machines)?

samples/Sentry.Samples.Console.Basic/Sentry.Samples.Console.Basic.csproj
vaind: As I've mentioned in the other PR, I am not sure about setting PublishAot globally, i.e. not only for sample apps but also for libraries. Let's see how that pans out. 

In the meantime, I'll add a TODO here but this PR is still addressing the 4.0.0 branch so can't really change any code here.

```suggestion
	<!-- TODO, consider updating after changes from #2732 come into effect -->
    <PublishAot Condition="!$([MSBuild]::IsOSPlatform('OSX')) or '$(TargetFramework)' == 'net8.0'">true</PublishAot>
```

src/Sentry/buildTransitive/Sentry.targets
vaind: Yes, it does and it should. We don't want to upload samples to sentry.io just because we (sentry-dotnet SDK contributors) have SENTRY_AUTH_TOKEN env var set locally, which would enable CLI.

Instead, it will only ever upload any project starting with`Sentry.` if explicitly configured by `/p:SentryCLIIntegrationTestProject=.....`. That's what the integration test script passes now.

src/Sentry/buildTransitive/Sentry.targets
vaind: this is a transitive build target, so one that gets triggered during lib-consumer builds.

jamescrosswell (approved):

src/Sentry/buildTransitive/Sentry.targets
jamescrosswell: Ah OK. In that case maybe [the comment](https://github.com/getsentry/sentry-dotnet/blob/1dfe043e9df4c277c3a66eb06f440d03150cbc00/src/Sentry/buildTransitive/Sentry.targets#L38-L37) needs to be updated. Currently it reads:
> We're explicitly skipping uploads for Sentry projects (except samples)

vaind: right, that was an in-betwen solution, thanks

src/Sentry/buildTransitive/Sentry.targets
```diff
@@ -34,8 +34,13 @@
       <SentryUploadSymbols Condition="'$(SentryUploadSymbols)' == ''">false</SentryUploadSymbols>
       <SentryUploadSources Condition="'$(SentryUploadSources)' == ''">false</SentryUploadSources>
 
-      <!-- This property controls if the Sentry CLI is to be used at all.  Setting false will disable all Sentry CLI usage. -->
-      <UseSentryCLI Condition="'$(UseSentryCLI)' == '' And ('$(SentryUploadSymbols)' == 'true' Or '$(SentryUploadSources)' == 'true' Or $(SentryUploadAndroidProguardMapping) == 'true')">true</UseSentryCLI>
+      <!-- This property controls if the Sentry CLI is to be used at all. Setting false will disable all Sentry CLI usage.
+           We're explicitly skipping uploads for Sentry projects (except samples) because they interfere with CLI integration test asserts. -->
```
vaind, line 38:
```suggestion
           We're explicitly skipping uploads for Sentry projects because they interfere with CLI integration test asserts. -->
```

src/Sentry/buildTransitive/Sentry.targets
jamescrosswell: Aha... cunning. I had no idea we were doing that (or that it was even possible - never had a need before). Is that enabled simply by this?
https://github.com/getsentry/sentry-dotnet/blob/d32f05ba46ca3a022224dae039b27f13691ac06c/src/Sentry/Sentry.csproj#L78-L86

I think that's all we need to do right (reading [the docs](https://learn.microsoft.com/en-us/nuget/concepts/msbuild-props-and-targets)).

vaind: yes, that's it

vaind: and I've learned about this a couple of weeks ago myself, for the same reason :)

src/Sentry/buildTransitive/Sentry.targets
```diff
@@ -120,53 +125,88 @@
...
+    <SentryCLIUploadSymbolType Include="portablepdb" />
+    <SentryCLIUploadSymbolType Include="dsym" Condition="$([MSBuild]::GetTargetPlatformIdentifier('$(TargetFramework)')) == 'ios'" />
+    <SentryCLIUploadSymbolType Include="dsym" Condition="$([MSBuild]::GetTargetPlatformIdentifier('$(TargetFramework)')) == 'maccatalyst'" />
+  </ItemGroup>
 
+  <!-- Upload symbols to Sentry after the build. -->
+  <Target Name="UploadDebugInfoToSentry" AfterTargets="$(SentryCLIUploadAfterTargets)" DependsOnTargets="PrepareSentryCLI"
+    Condition="'$(SentryCLI)' != '' and ('$(SentryUploadSymbols)' == 'true' or '$(SentryUploadSources)' == 'true')">
+
+    <!-- if (UploadSymbols && UploadSources) { -->
```
bitsandfoxes, line 161: Thanks for those!

vaind: yeah they're nice although at one point I've found myself just copying a piece of a code from one if branch to the other, forgetting to actually update the relevant condition 😅

bitsandfoxes (approved):
:shipit:

Issue #2713, opened 2023-10-10
Single CaptureEvent with optional parameters Hint, Scope


jamescrosswell:
We've got a bunch of methods like this scattered throughout the SDK:

https://github.com/getsentry/sentry-dotnet/blob/0714fd20df8991dfc47ab066a63dbfd495dc009b/src/Sentry/SentryClient.cs#L62-L67

I suspect we created the overloads when we introduced hints. We don't really need multiple overloads - we could just have one method signature with optional parmeters.

bruno-garcia:
Most people don't care about hints. Can we avoid this change?

PR #2749, opened 2023-10-23
Consolidated CaptureEvent API methods

jamescrosswell:
Resolves https://github.com/getsentry/sentry-dotnet/issues/2713


bitsandfoxes (approved):
`IHubEx` be gone! 🔥 
I suspect the order `scope, hint` for the optional parameters is due to us expecting that scopes are much more likely to be used than hints?
I like that none of the samples had to be touched.

PR #2750, opened 2023-10-23
feat: cocoa SDK integration for macOS

### LINKED ISSUE #2700: Native AOT - attach debug images
Native AOT builds are really a native app and the stack traces come with instruction addresses. Therefore, to symbolicate this on Sentry, we need to append a list of debug images currently loaded, including the load address. We can either build this list manually by calling appropriate native platform-specific endpoints or use existing functionality from `sentry-native`. The latter seems like a better idea because in the future we may also want to use sentry-native for crash handling.

Related: #2247 and #2692 

* [x] linux & windows (sentry-native): #2704 
* [x] macos (sentry-native): #2767

vaind:
Follows up #2704 by enabling the Cocoa SDK for macOS. This is loosely based on #2625 but approaches the integration from a different angle: instead of adding another build and binding code, we try to leverage objective-sharpie generated bindings.

Closes #2700

#skip-changelog - we'll have a single item for #2247

### DIFF
--- a/.github/actions/buildcocoasdk/action.yml
+++ b/.github/actions/buildcocoasdk/action.yml
@@ -4,23 +4,17 @@ runs:
   using: composite
 
   steps:
-
-    - name: Fetch version tags for Sentry Cocoa SDK
+  
+    - name: Get submodule status
       shell: bash
-      run: git fetch --tags --quiet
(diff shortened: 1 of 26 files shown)


vaind:
replaced by #2767

PR #2755, opened 2023-10-24
Spike: Cocoa bindings fails dirty-check

jamescrosswell:
##skip-changelog


vaind (commented):
The issue is caused by the lack of Xamarin on the new macos-13 runners. We can install it using `brew install --cask xamarin-ios`

vaind (approved):

Issue #2666, opened 2023-09-26
Replace TracePropagationTarget with SubstringOrRegexPattern


jamescrosswell:
https://github.com/getsentry/sentry-dotnet/blob/7513600e2f9a3da153b349ad623447de95f3ef74/src/Sentry/TracePropagationTarget.cs#L3-L12

PR #2763, opened 2023-10-26
Remove TracePropagationTarget

jamescrosswell:
Resolves https://github.com/getsentry/sentry-dotnet/issues/2666

# Motivation

`TracePropagationTarget` currently doesn't add any functionality over/above `SubstringOrRegexPattern`, so there's no need to retain both of these:
```csharp
[TypeConverter(typeof(TracePropagationTargetTypeConverter))]
public class TracePropagationTarget: SubstringOrRegexPattern
{
    public TracePropagationTarget(string substringOrRegexPattern, StringComparison comparison = StringComparison.OrdinalIgnoreCase)
        : base(substringOrRegexPattern, comparison)
    {
    }

    public TracePropagationTarget(Regex regex) : base (regex)
    {
    }
}
```

`SubstringOrRegexPattern` is the less specific type that is also used for `FailedRequestTargets`:
https://github.com/getsentry/sentry-dotnet/blob/f2524bdf758d3af0192e7ba91d24abb21a1a562d/src/Sentry/SentryOptions.cs#L678-L680


src/Sentry/TracePropagationTarget.cs
```diff
@@ -1,60 +0,0 @@
-namespace Sentry;
-
-// TODO: Replace TracePropagationTarget with SubstringOrRegexPattern in next major version.
```
bruno-garcia, line 3: It doesn't really explain why.
This type at least from what its named is part of Sentry's protocol and easy to find in docs, etc.

It's getting removed just because it's 'the same as SubstringOrRegexPattern'?

CHANGELOG.md
```diff
@@ -40,22 +40,23 @@ API Changes:
     - `SentrySdk.CaptureEvent(SentryEvent @event, Action<Scope> scopeCallback)`
     - `SentrySdk.CaptureMessage(string message, Action<Scope> scopeCallback)`
     - `SentrySdk.CaptureException(Exception exception, Action<Scope> scopeCallback)`
+- The TracePropagationTarget class has been removed. Use the SubstringOrRegexPattern class instead. ([#2763](https://github.com/getsentry/sentry-dotnet/pull/2763))
```
bitsandfoxes, line 43: The last changelog entry should be the one regarding `WithScope`. The `#### Before` is the migration guide for it.

src/Sentry/TracePropagationTarget.cs
jamescrosswell: I put some context in the PR description. I think the `TracePropagationTargets` property will still be easy enough to find and is well documented and having it's own type wouldn't make it any more usable:
https://github.com/getsentry/sentry-dotnet/blob/f2524bdf758d3af0192e7ba91d24abb21a1a562d/src/Sentry/SentryOptions.cs#L824-L842

bitsandfoxes (approved):
Thanks!

bruno-garcia (approved):

PR #2765, opened 2023-10-30
chore: include debug info in sentry-native static build

vaind:
See https://github.com/getsentry/sentry-native/issues/895 for context.
In short we need to pass additional flags so that the sentry-native static lib contains debug info

#skip-changelog


jamescrosswell (approved):
I'll trust you on this one - I haven't had much to do with the native bindings yet.

Issue #2665, opened 2023-09-26
Consider removing SentryValues (or making this internal at least)


jamescrosswell:
https://github.com/getsentry/sentry-dotnet/blob/e75d537c14ded617c99eb33f19cf4ce69ca4de36/src/Sentry/SentryValues.cs#L6-L10

PR #2771, opened 2023-10-31
Made Sentry.Values<T> internal

jamescrosswell:
Resolves https://github.com/getsentry/sentry-dotnet/issues/2665

Sentry.Values<T> is only used internally. Publicly this is exposed in two places and in both cases as `IEnumerable<T>`:
https://github.com/getsentry/sentry-dotnet/blob/56e5b06f0aee72e24ee392d3394d25d52a6becd1/src/Sentry/SentryEvent.cs#L74-L95


vaind (approved):

vaind:
Danger was complaining because the changelog entry was missing a `#` before the number. See https://github.com/getsentry/sentry-dotnet/commit/a448ab363a82d0a7c27b1e10487218a7168f5758

PR #2777, opened 2023-11-01
refactor: merge Sentry.Bindings.Native into Sentry

vaind:


src/Sentry/Sentry.csproj
```diff
@@ -21,7 +21,7 @@
   <!-- Platform-specific props included here -->
   <Import Project="Platforms\Android\Sentry.Android.props" Condition="'$(TargetPlatformIdentifier)' == 'android'" />
   <Import Project="Platforms\iOS\Sentry.iOS.props" Condition="'$(TargetPlatformIdentifier)' == 'ios' Or '$(TargetPlatformIdentifier)' == 'maccatalyst'" />
-  <Import Project="Platforms\Native\Sentry.Native.props" Condition="('$(TargetFramework)' == 'net7.0' Or '$(TargetFramework)' == 'net8.0')" />
+  <Import Project="Platforms\Native\Sentry.Native.targets" Condition="'$(TargetFramework)' == 'net6.0'" /> <!-- TODO use net7.0 instead, if we add that as a TF. -->
```
bitsandfoxes, line 24: Is that because we just can?

bitsandfoxes (approved):

src/Sentry/Sentry.csproj
vaind: I don't get the question.

we need to condition this on net6.0 because that's the highest TF version we're currently targeting in the Senttry.csproj. If/when we add net7.0 or net8.0, we can change this condition because in reality it doesn't make sense to build this for net6.0 anymore (since anyone targeting AOT would be on net7.0/net8.0 and then their build would pick the higher build of Sentry.dll)

Issue #2772, opened 2023-10-31
Native AOT stack trace integration test


vaind:
We can't unit test Native AOT implementation of `DebugStackTrace` (there's no xUnit runner to support AOT so no way to test this in unit tests. There's a tracking issue: https://github.com/dotnet/runtime/issues/91069). Instead, we need to have an integration test that captures app requests. 

Possible solutions
* use something like the Sentry CLI integration test pwsh script and mock the sentry server
* use something like the SingleFileTestApp and run through unit tests: https://github.com/getsentry/sentry-dotnet/blob/feat/4.0.0/test/Sentry.Tests/Internals/ILSpy/SingleFileAppTests.cs

jamescrosswell:
@vaind are there any gaps on this issue that I can help with? I see you've already got various integration tests [here](https://github.com/getsentry/sentry-dotnet/pull/2781/files#diff-9b2f8e99ffe82a8a1ac2e7da3abe82efdbf06e172cd597778eba32fe82fb5df8). I'm not sure if that covers all the scenarios we wanted to cover though... e.g. all the [TODO items in DebugStackTraceTests.verify.cs](https://github.com/getsentry/sentry-dotnet/blob/feat/4.0.0/test/Sentry.Tests/Internals/DebugStackTraceTests.verify.cs).

I figured integration tests is probably the most important thing still sitting on the backlog besides what you're doing.

vaind:
There are some basic runtime tests for AOT in  [`integration-test/runtime.Tests.ps`](https://github.com/getsentry/sentry-dotnet/blob/79695bb9692698eae180c2cabf4f1a53bd5dc7ab/integration-test/runtime.Tests.ps1#L79) added by #2781

I've had a look at the places marked with TODO linking this issue & integration testing it would be quite a bit of work so I hope we could wait for https://github.com/dotnet/runtime/issues/91069

ivanpovazan:
> For future investigation: I've had an idea to run the MAUI devicetest app with a NativeAOT published build. I can't try it out at the moment because we need to update device-tests to net8 first, which is blocked (#2854).

FWIW, we will be standing up MAUI device tests with NativeAOT which is being tracked here: https://github.com/dotnet/maui/issues/19817

PR #2781, opened 2023-11-02
feat: packaging native runtime libs

vaind:
Part of #2247 and #2772

Besides packaging sentry-native static lib properly, this adds integration testing capabilities where now we create sample apps with `dotnet new` and test against actual nuget packages, instead of relying on all those build props/targets that are implied on Sample projects inside our repo.

Followup: #2809 - MAUI test is currently disabled to unblock this PR

#skip-changelog - there will be a single item for Native AOT support


scripts/build-sentry-native.ps1
```diff
@@ -0,0 +1,76 @@
...
+
+    if ($Clean)
+    {
+        rm -rf $buildDir
+    }
+
+    cmake `
+        -S $submodule `
+        -B $buildDir `
+        -D CMAKE_BUILD_TYPE=RelWithDebInfo `
```
bruno-garcia, line 48: Debug info will be bundled in case of mac/linux? And on Windows we'll get dlls?

Should we upload this through Symbol Collector and avoid bundling that in the NuGet Package? It's already extremely large with all the Sentry CLI per platform

vaind: We're compiling static libraries, therefore the debug info will only be created once the final app is compiled (with addresses relative to the final executable). 

The best we can do to reduce package size is stop packaging CLI and let it be downloaded later during user project's build, as needed. 

BTW, I've also reduced sentry CLi binary size recently (should come out with the latest release: https://github.com/getsentry/sentry-cli/pull/1794)

bruno-garcia: it's all good, sorry I forgot about it being a static lib

integration-test/cli.Tests.ps1
```diff
@@ -0,0 +1,178 @@
...
+
+    It "uploads nothing when disabled" {
+        $result = RunDotnetWithSentryCLI 'build' 'console-app' $False $False $framework
+        $result.UploadedDebugFiles() | Should -BeNullOrEmpty
+    }
+}
+
+Describe 'Console apps (<framework>) - native AOT publish' -ForEach @(
+    @{ framework = "net7.0" },
+    @{ framework = "net8.0" }
```
bruno-garcia, line 44: Adding a new framework will be done here?
Could you please write some doc/instruction on how this works/how to keep it maintained or add new frameworks?

bruno-garcia (approved):

integration-test/cli.Tests.ps1
vaind: I've written some intro in CONTRIBUTING.md (https://github.com/getsentry/sentry-dotnet/pull/2781/files#diff-eca12c0a30e25b4b46522ebf89465a03ba72a03f540796c979137931d8f92055R138).

And you've guessed right so the question of adding frameworks should be clear then :) 

But if you have questions that you don't know the answer to (or anyone else does), feel free to post here and I'll update the docs.

.github/workflows/build.yml
```diff
@@ -151,20 +151,20 @@ jobs:
           sparse-checkout: |
```
jamescrosswell, line 151: I see... is it maybe dangerous for us to use sparse-checkout in CI at all? It seems like this introduces the possibility of differences between running jobs on CI and running things locally (differences that are really hard to spot/debug)? 

If we're going to optimize our CI for something, I'd rather go with reliability over speed (since fast, unreleable CI costs us lots of human time... which is way more expensive and constrained than github actions compute time for us).

vaind: > It seems like this introduces the possibility of differences between running jobs on CI and running things locally (differences that are really hard to spot/debug)?

Yes, it does. The only alternative I can think of (that also makes sure all these .props and .targets and other files I can't even think of are not picked up from parent directories by dotnet tooling) is moving all the sources of the current repo to another directory and having the integration test as a sibling, not child of that root directory. 

Well no, not the only one, the other alternative is building in a directory outside the project root (e.g. in %TEMP%) which is not great for local development either because then you have a hard time manually checking the generated project files if something doesn't work.

They all have their costs ¯\_(ツ)_/¯

vaind: I'm open to suggestions how to achieve the isolation during tests (at least in CI, as it is now) that would improve local development experience

vaind: Should I update the comment in any way? That's what it was trying to get across so not sure if you've missed it (I usually ignore skip comments...) or it's unclear.

jamescrosswell: I hadn't seen the comment actually (I was looking at the commit diff in github - not sure if that even showed). Maybe worth adding `(e.g. Directory.Build.props etc.)` just to be clear about what is meant by local files impacting integration tests.

vaind: done, thx for the feedback.

PR #2784, opened 2023-11-03
perf: Speed up SDK init

jairbubbles:
In some scenarios the time spent in initializing the SDK can be too important. What is tricky is that the init of the SDK is one of the first thing the app do and we're JITing almost everything. Creating a `HttpClient`, a `Regex` is costly...

In this pull request I'm experimenting with lazy instanciation so all that JIT time for these types will be spent afterwards.

I'm disabling almost everything to be in the best scenario possible:
```csharp
await Task.Run(() =>
{
    SentrySdk.Init(o =>
    {
        o.Dsn = "<sentry_dsn>";
        o.Release = "1.0.0";
        o.DisableAppDomainProcessExitFlush();
        o.DisableAppDomainUnhandledExceptionCapture();
        o.DisableUnobservedTaskExceptionCapture();
        o.DisableDuplicateEventDetection();
        o.DisableDiagnosticSourceIntegration();
        o.DisableWinUiUnhandledExceptionIntegration(); // this one is new!
        o.AutoSessionTracking = false;
        o.IsGlobalModeEnabled = true;
        o.DetectStartupTime = StartupTimeDetectionMode.Fast;
        o.TracesSampleRate = 0;
    });
});
```

Here is the current status:

image

Compared to main branch:

image

Let me know what you think!


src/Sentry/SentryOptions.cs
```diff
@@ -59,12 +59,18 @@ public bool IsGlobalModeEnabled
...
+
     /// <summary>
     /// Specifies whether to use global scope management mode.
     /// Should be <c>true</c> for client applications and <c>false</c> for server applications.
     /// The default is <c>false</c>. The default for Blazor WASM, MAUI, and Mobile apps is <c>true</c>.
     /// </summary>
-    public bool IsGlobalModeEnabled { get; set; } = Runtime.Current.IsBrowserWasm();
+    public bool IsGlobalModeEnabled
+    {
+        get => _isGlobalModeEnabled ?? Runtime.Current.IsBrowserWasm();
```
bruno-garcia, line 71: I think this is called often so we could cache the value here on first access:

```suggestion
        get => _isGlobalModeEnabled ??= Runtime.Current.IsBrowserWasm();
```

src/Sentry/SentryOptionsExtensions.cs
```diff
@@ -387,7 +351,12 @@ public static IEnumerable<ISentryEventExceptionProcessor> GetAllExceptionProcess
     /// <param name="sentryStackTraceFactory">The stack trace factory.</param>
     public static SentryOptions UseStackTraceFactory(this SentryOptions options, ISentryStackTraceFactory sentryStackTraceFactory)
     {
-        options.SentryStackTraceFactory = sentryStackTraceFactory ?? throw new ArgumentNullException(nameof(sentryStackTraceFactory));
+        if (sentryStackTraceFactory is null)
```
bruno-garcia, line 354: no functional change here, just style?

src/Sentry/Internal/Http/LazyHttpTransport.cs
```diff
@@ -0,0 +1,22 @@
...
+    private HttpTransport? _httpTransport;
+
+    public LazyHttpTransport(SentryOptions options)
+    {
+        _options = options;
+    }
+
+    public Task SendEnvelopeAsync(Envelope envelope, CancellationToken cancellationToken = default)
+    {
+        _httpTransport ??= new HttpTransport(_options, (_options.SentryHttpClientFactory ?? new DefaultSentryHttpClientFactory()).Create(_options));
```
bruno-garcia, line 18: How about moving this all under a `Lazy<>` and access it via `_lazyTransport.Value` ?

src/Sentry.DiagnosticSource/SentryOptionsDiagnosticExtensions.cs
```diff
@@ -15,7 +15,7 @@ public static class SentryOptionsDiagnosticExtensions
     /// <param name="options">The Sentry options.</param>
     public static void AddDiagnosticSourceIntegration(this SentryOptions options)
     {
-        if (options.Integrations?.OfType<SentryDiagnosticListenerIntegration>().Any() is true)
+        if (options.Integrations.Any(integration => integration.Type == typeof(SentryDiagnosticListenerIntegration)))
```
bruno-garcia, line 18: Now with the type, we prob should go with a dictionary Type:instance
We're doing allocations in each of the LINQ calls, and iterating the lists

bruno-garcia (commented):
Nice! Thanks for this PR, the idea makes sense. I realize it's a draft but left some notes already.

Could you please add some benchmarks in https://github.com/getsentry/sentry-dotnet/tree/main/benchmarks ?

src/Sentry/SentryOptionsExtensions.cs
jairbubbles: It's because of successive refactor. I'll fix it.

src/Sentry/Internal/Http/LazyHttpTransport.cs
jairbubbles: Sure, there's not enough `Lazy` in my PR 😅

src/Sentry.DiagnosticSource/SentryOptionsDiagnosticExtensions.cs
jairbubbles: That was my previous modification but I've had issues in unit test where we use substitutes as they were sharing the same "RuntimeType"  type.

Also, it was not clear to me if we can or not have several times the same types.

At the end, I use `List<Tuple>` as I felt like it was more close to previous code.

jairbubbles:
> Could you please add some benchmarks in https://github.com/getsentry/sentry-dotnet/tree/main/benchmarks ?

I struggled a bit to add the benchmarks as I needed a specific "ColdStart" configuration but here we go:

|            Method |     Mean |    Error |   StdDev |      Min |      Max |   Median | Methods JITted | Methods Tiered | JIT allocated memory | Allocated |
|------------------ |---------:|---------:|---------:|---------:|---------:|---------:|---------------:|---------------:|---------------------:|----------:|
| 'Init SDK (Full)' | 68.88 ms | 63.86 ms | 16.59 ms | 58.46 ms | 98.09 ms | 63.95 ms |          1,160 |            110 |               236 KB |  44.35 KB |
| 'Init SDK (Slim)' | 44.72 ms | 52.51 ms | 13.64 ms | 35.17 ms | 66.58 ms | 36.53 ms |            900 |              1 |               165 KB |  28.34 KB |

jairbubbles:
Some notes:
- `Dns.Parse` is still taking a bit of time (20ms), maybe we could do a simple verification in the constructor and lazy init it?
- Some features which I disabled in my use case were taking a lot of time : in particular AutoSession is writing / reading a file for the persistant id.

vaind:
> Dns.Parse is still taking a bit of time (20ms),

that's not a bit of time, that's a lot of time for such a simple operation (seemingly). Could use a bit more investigation.

jairbubbles:
> > Dns.Parse is still taking a bit of time (20ms),
> 
> that's not a bit of time, that's a lot of time for such a simple operation (seemingly). Could use a bit more investigation.

Yeah but it's due to JIT.

bruno-garcia:
> > > Dns.Parse is still taking a bit of time (20ms),
> > 
> > 
> > that's not a bit of time, that's a lot of time for such a simple operation (seemingly). Could use a bit more investigation.
> 
> Yeah but it's due to JIT.

We looked at it with @bitsandfoxes when profiling this and IIRC it was mainly the `Uri` class, most time spent.
If that's the case we could prob stop using it and take a simpler approach.
We parse it in other SDKs, e.g in `C`: https://github.com/getsentry/sentry-native/blob/406d41881111d15b1144b443d51760c1415c8029/src/sentry_utils.c#L216-L270

src/Sentry.DiagnosticSource/SentryOptionsDiagnosticExtensions.cs
bruno-garcia: Fair, Tuple is the way to go unless we'd want to break that behavior.
Still dropping the LINQ usage is potential area of improvement if we're looking for more.

jairbubbles: Still, overall it feels a bit complex. How would you feel if those default integrations would be stored in an enum flag? During the config you would just need to remove the flags in the `DisableXXX` methods. It requires a bit more work though.

jairbubbles:
```csharp
[Flags]
internal enum DefaultIntegrations
{
    AutoSessionTrackingIntegration = 1 << 0,
    AppDomainUnhandledExceptionIntegration = 1 << 1,
    AppDomainProcessExitIntegration = 1 << 2,
    UnobservedTaskExceptionIntegration = 1 << 3,
    NetFxInstallationsIntegration = 1 << 4,
    SentryDiagnosticListenerIntegration = 1 << 5,
    WinUiUnhandledExceptionIntegration = 1 << 6,
}
```

jairbubbles: I gave it a shot in my last commit.

src/Sentry/SentryOptions.cs
```diff
@@ -1157,4 +1226,38 @@ public SentryOptions()
...
     }
+
+    internal void AddIntegration(ISdkIntegration integration)
+    {
+        _integrations.Add(integration);
+    }
+
+    internal void RemoveIntegration<TIntegration>()
+    {
+        // TODO: For backward compatibility, check if it's a default integration type?
```
bruno-garcia, line 1237: All default integrations are internal, by finding its usages we could spot any removal of default integrations 

```suggestion
    // Note: Not removing default integrations
```

We have usage in Unity (which access internal API) but no default integration there: https://github.com/search?q=repo%3Agetsentry%2Fsentry-unity%20RemoveIntegration&type=code


A comment on the method stating it doesn't apply to default integrations would be good enough I imagine

src/Sentry/Internal/Hub.cs
```diff
@@ -62,7 +62,7 @@ internal class Hub : IHubEx, IDisposable
         _enricher = new Enricher(options);
 
         // An integration _can_ deregister itself, so make a copy of the list before iterating.
-        var integrations = options.Integrations?.ToList() ?? Enumerable.Empty<ISdkIntegration>();
+        var integrations = options.Integrations.ToList();
```
bruno-garcia, line 65: Could we get around iterating on `Integrations` below without allocating the list?

bruno-garcia (commented):
This is quite promising.Once it's out of draft we can test it against `feat/4.0.0` to make sure it works fine with AOT

bruno-garcia:
> But it's not so convincing when running it on main to compare:

Would be more noticible if we had some "application code" right after Init. That now would get a chance to run before some gets sent into Sentry vs before where it would need everything to bootstrap

src/Sentry/Internal/Hub.cs
jairbubbles: My first draft didn't have the `ToList()` and some test was failing. One integration was removing another integration.

It's probably this one (which will work fine now with the enum flags):
```csharp
internal class SentryDiagnosticListenerIntegration : ISdkIntegration
{
    {
        if (!options.IsPerformanceMonitoringEnabled)
        {
            options.Log(SentryLevel.Info, "DiagnosticSource Integration is disabled because tracing is disabled.");
            options.DisableDiagnosticSourceIntegration();
            return;
        }

        var subscriber = new SentryDiagnosticSubscriber(hub, options);
        DiagnosticListener.AllListeners.Subscribe(subscriber);
    }
}
```

jairbubbles: In fact no it won't be working, there is a duality with `SentryDiagnosticListenerIntegration` as it's internaly provided or provided by `Sentry.DiagnosticSource`. 

I feel like it's useless to remove itself from the options. I'm removing the `ToList`, I just adapted a test to check the log instead.

jairbubbles:
With latest changes, here are the benchmarks results:

|            Method |     Mean |    Error |   StdDev |      Min |      Max |   Median | Methods JITted | Methods Tiered | JIT allocated memory | Allocated |
|------------------ |---------:|---------:|---------:|---------:|---------:|---------:|---------------:|---------------:|---------------------:|----------:|
| 'Init SDK (Full)' | 42.91 ms | 1.679 ms | 1.110 ms | 41.51 ms | 44.63 ms | 42.52 ms |          1,028 |              1 |               186 KB |  43.05 KB |
| 'Init SDK (Slim)' | 25.45 ms | 0.981 ms | 0.649 ms | 24.51 ms | 26.25 ms | 25.56 ms |            891 |              1 |               161 KB |  27.22 KB |

bruno-garcia (approved):
Great stuff! Thanks @jairbubbles 
Prob needs @bitsandfoxes @vaind or @jamescrosswell as some extra check before we can merge but lgtm

bitsandfoxes:
Those are some serious gains! I'll follow up with a review/merge this tomorrow.

bitsandfoxes (approved):
Let me run this against `4.0.0` before merging to avoid any heartache.

bruno-garcia:
Lets merge this and figure out AOT after

bruno-garcia:
Thanks @jairbubbles

bitsandfoxes:
Already tested. Works like a charm. No hickups there!

PR #2787, opened 2023-11-04
remove deprecated frameworks

bruno-garcia:
I thought of dropping netstandard2.0 from Sentry.AspNetCore even though technically still supported via ASP.NET Core 2.1 on .NET Framework because frankly, not much use there and big maintenance burden.

But turns out Azure Workers support netstandard2.0, and our integration is built on top of our ASP.NET Core package. So lets keep it in this major and prob goes away once that drops. Then we can stop testing aspnetcore on .net framework too

[Grpc.AspNetCore.Server min version](https://www.nuget.org/packages/Grpc.AspNetCore.Server/2.58.0#supportedframeworks-body-tab) is net6.0, dropping 3.1


vaind (approved):

vaind:
Related discussion: https://github.com/getsentry/sentry-dotnet/discussions/2776

PR #2791, opened 2023-11-05
test on net8.0

bruno-garcia:
#skip-changelog

having some issues with Verify

image


SimonCropp:
can u see if https://github.com/getsentry/sentry-dotnet/pull/2798 helps?

bruno-garcia:
Last standing snapshot testing challenge:
https://stackoverflow.com/questions/77442233/verify-snapshot-test-writes-all-frameworks-to-single-file

SimonCropp:
have a look at the `UniqueFor` options https://github.com/VerifyTests/Verify/blob/main/docs/naming.md#uniquefor

jamescrosswell:
> have a look at the `UniqueFor` options https://github.com/VerifyTests/Verify/blob/main/docs/naming.md#uniquefor

@SimonCropp I tried using `UniqueFor` but same result. 

https://github.com/getsentry/sentry-dotnet/blob/18c7ede71c9f1c68366cdd88b58186312e9eda22/test/Sentry.AspNetCore.Tests/WebIntegrationTests.verify.cs#L66-L69

I also tried manually renaming files like `WebIntegrationTests.Versioning.DotNet7_0.received.txt` to `WebIntegrationTests.Versioning.DotNet7_0.verified.txt`. On the next run however, verfiy wants to reroute the verified files for multiple different target frameworks back to the target agnostic `WebIntegrationTests.Versioning.Dot.verified.txt` file. See below (these are the diffs verify opens automatically in Rider... they both want to reroute to the same file but they are different - look at the last few characters on line 30:

image

bruno-garcia:
@jamescrosswell I think that diff just says: OK move to the right (accept change) and it'll create the new files.

bruno-garcia:
Today we have 1 test -> N runtimes -> 1 file
Now it'll be 1 test -> N runtimes -> N files

You need to commit the new files

jamescrosswell:
> EDIT: Oh I see the new ones are not with the runtime names yet? My bad, looks wrong

Yeah, it's trying to reroute net6.0 and net7.0 to the "main" verified file (because these are the same)... but then it tries to do the same with the net8.0 one (which is different). So not detecting that there are actually differences for the net8.0 framework and putting that one in a runtime specific verified file...

jamescrosswell:
No luck upgrading to Verify.Xunit v22.2.0-beta.7 and Xunit v2.6.1 either 😞

jamescrosswell:
@SimonCropp I've created a minimal reproduction of the issue in https://github.com/jamescrosswell/VerifyTestProject...

vaind (approved):
LGTM after removing the unnecessary file.

bruno-garcia:
Thanks!!

Issue #2780, opened 2023-11-02
Extend Trimming to all Sentry packages


jamescrosswell:
In https://github.com/getsentry/sentry-dotnet/pull/2732 we added the `IsAotCompatible` property to `Sentry.csproj`. 

## Not planned (legacy or .net framework stuff)
- src/Sentry.Log4Net/Sentry.Log4Net.csproj
- src/Sentry.EntityFramework/Sentry.EntityFramework.csproj

## Low hanging fruit

Beyond Sentry.csproj many of the other projects can have this added as well with very little modification. We can leverage the same technique to mark these as AOT compatible in the respective csproj files:
```
  <PropertyGroup Condition="'$(FrameworkSupportsAot)' == 'true'">
    <IsAotCompatible>true</IsAotCompatible>
  </PropertyGroup>
```

Also, if they don't already, these projects will have to target `.net6.0` or later (since `FrameworkSupportsAot` doesn't have any effect otherwise). 

- https://github.com/getsentry/sentry-dotnet/pull/2793
  - src/Sentry.Serilog/Sentry.Serilog.csproj
  - src/Sentry.Profiling/Sentry.Profiling.csproj
  - src/Sentry.OpenTelemetry/Sentry.OpenTelemetry.csproj
  - src/Sentry.NLog/Sentry.NLog.csproj
  - src/Sentry.Log4Net/SentryAppender.cs

## ConfigureFromConfigurationOptions<TOptions>

Some  of our integrations use `ConfigureFromConfigurationOptions<TOptions>`, which isn't compatible with Trimming and we need to use the new [configuration binding source generator](https://learn.microsoft.com/en-us/dotnet/core/whats-new/dotnet-8#configuration-binding-source-generator): 

This affects:

- `Sentry.Extensions.Logging`
- `Sentry.Maui`
- `Sentry.Azure.Functions.Worker`
- `Sentry.AspNetCore`

This needs some investigation.

## Sentry.DiagnosticSource

We don't need to do anything for `Sentry.DiagnosticSource` as it's included in the Sentry core package for all versions of .NET that support AOT:
https://github.com/getsentry/sentry-dotnet/blob/4080ac6b894ff263da6e01805bc91df1b3f84f2e/src/Sentry/Sentry.csproj#L53-L61

bruno-garcia:
Doesn't really need to be on all.
It should be on stuff that makes most sense. Like `Sentry`. `Sentry.AspNetCore` All the serverless stuff.

[...]

PR #2793, opened 2023-11-06
IsAotCompatible for uncomplicated projects

jamescrosswell:
Added IsAotCompatible to all the integrations where this is very straight forward (low hanging fruit described in https://github.com/getsentry/sentry-dotnet/issues/2780).


src/Sentry.EntityFramework/Sentry.EntityFramework.csproj
```diff
@@ -1,11 +1,15 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFrameworks>net462;netstandard2.1</TargetFrameworks>
+    <TargetFrameworks>net6.0;net462;netstandard2.1</TargetFrameworks>
     <Description>Official Entity Framework 6 integration for Sentry - Open-source error tracking that helps developers monitor and fix crashes in real time.</Description>
     <PackageTags>$(PackageTags);EntityFramework;EF;EF6</PackageTags>
   </PropertyGroup>
 
+  <PropertyGroup Condition="'$(FrameworkSupportsAot)' == 'true'">
```
vaind, line 9: Is there any reason why we must define this conditionally instead of just for all frameworks?

jamescrosswell: Under the hood it sets things like `IsTrimmable`. I don't think you can do that on a .NET Framework application or a .NET 5.0 application as it was only introduced in .NET 6.

Our Device tests fall over if we enable it for `net7.0-android` as well... so currently we can't turn this on for all frameworks.

vaind: You can sat whatever properties you want. They won't just have any effect.

vaind: > Our Device tests fall over if we enable it for `net7.0-android` as well... so currently we can't turn this on for all frameworks.

I was talking specifically about the added property, i.e. `<IsAotCompatible>`. Does it trip tests or can we just add it in general?

vaind: I mean I don't mind this too much, it's just that I prefer avoiding conditional properties unless we know they're necessary, otherwise it makes reasoning about things harder (for me).

jamescrosswell: > I was talking specifically about the added property, i.e. `<IsAotCompatible>`. Does it trip tests or can we just add it in general?

Yeah, that was the reason I added conditions like targetframework != net7.0-android... in that case, it compiles, but our tests fail (it is [still experimental](https://learn.microsoft.com/en-us/dotnet/core/deploying/native-aot/?tabs=net8plus%2Cwindows#platformarchitecture-restrictions) there I think). 

From memory, if you try to enable IsTrimmable for all the frameworks, the compiler bleats for .net framework and netcoreapp3.1... but it was such a game of wack-a-mole, my memory may not be very accurate. I think my brain blocked some of it out as traumatic. 

Regardless, until the issues with android are resolved, I don't think we can enable it unconditionally. At least we have a single place where we can modify the conditionality though.

vaind: thx for the explanatino

src/Sentry.EntityFramework/Sentry.EntityFramework.csproj
```diff
@@ -21,8 +25,8 @@
     <Using Include="Sentry.Integrations" />
   </ItemGroup>
 
-  <ItemGroup Condition="'$(TargetFramework)' == 'net462'">
-    <PackageReference Include="EntityFramework" Version="6.2.0" />
+  <ItemGroup Condition="'$(TargetFramework)' == 'net6.0'">
```
vaind, line 28: the highest supported version of EntityFramework is net45 - why do we build for net6.0 at all?

src/Sentry.EntityFramework/ErrorProcessors/DbConcurrencyExceptionProcessor.cs
```diff
@@ -11,6 +11,6 @@ public class DbConcurrencyExceptionProcessor : SentryEventExceptionProcessor<DBC
     protected internal override void ProcessException(DBConcurrencyException exception, SentryEvent sentryEvent)
     {
         sentryEvent.SetExtra("Row Count", exception.RowCount);
-        sentryEvent.SetExtra("Row Error", exception.Row.RowError);
+        sentryEvent.SetExtra("Row Error", exception.Row?.RowError);
```
bruno-garcia, line 14:
```suggestion
        if (exception.Row is { } row)
        {
            sentryEvent.SetExtra("Row Error", row.RowError);
        }
```

src/Sentry.EntityFramework/Sentry.EntityFramework.csproj
```diff
@@ -1,11 +1,15 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFrameworks>net462;netstandard2.1</TargetFrameworks>
+    <TargetFrameworks>net6.0;net462;netstandard2.1</TargetFrameworks>
```
bruno-garcia, line 4: Do we need to add this new target? This package isn't something evolving so best to leave it with the old targets and eventually move it out of this repo

src/Sentry.Log4Net/Sentry.Log4Net.csproj
```diff
@@ -1,12 +1,16 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFrameworks>netstandard2.0;net462</TargetFrameworks>
+    <TargetFrameworks>net6.0;netstandard2.0;net462</TargetFrameworks>
```
bruno-garcia, line 4: Lets not add net6.0 for `log4net` if we can't get away without it

Another library mainly legacy, I expect most apps using this are on .NET Framework or not really evolving quickly

src/Sentry.EntityFramework/Sentry.EntityFramework.csproj
jamescrosswell: > the highest supported version of EntityFramework (and log4net) is net45 - why do we target net6.0 at all?

Because I didn't know that 😉 I spent too long as a manager, I think... I missed the whole bit between .net core and .net 7.

src/Sentry.Log4Net/Sentry.Log4Net.csproj
jamescrosswell: > Lets not add net6.0 for `log4net` if we can't get away without it

If we want to turn on `IsAotCompatible` for this integration then it needs to target `net6.0`, since one of the conditions for AOT compatibility is the target framework being `net6.0` or higher.

src/Sentry.EntityFramework/Sentry.EntityFramework.csproj
jamescrosswell: Just checking, so we don't want to turn on `IsAotCompatible` for `Sentry.EntityFramework `? That's the implication of not targeting `net6.0`. People may still be able to include it in applications that they publish AOT, but we wouldn't ever get any compiler assurance that that was the case. By turning on `IsAotCompatible` (which would require targeting `net6.0`) we get compiler warnings/errors if anything changes in this library such that it would break AOT compatibility.

src/Sentry.EntityFramework/Sentry.EntityFramework.csproj
jamescrosswell: See https://github.com/getsentry/sentry-dotnet/pull/2793#discussion_r1383992723

src/Sentry.EntityFramework/Sentry.EntityFramework.csproj
jamescrosswell: Hm, unlikely the MS package would work with AOT. I'll revert this one.

bruno-garcia: Yeah EntityFramework (aka EF6) is "old". EF Core is the "new" thing. We don't need to invest in AOT for EF 6

bruno-garcia (approved):
Other than changing EF6, I'm good with this.
AspNetCore is the largest opportunity for us with AOT. And I imagine it'll be the hardest thing?

jamescrosswell:
> Other than changing EF6, I'm good with this. AspNetCore is the largest opportunity for us with AOT. And I imagine it'll be the hardest thing?

I've reverted Sentry.EntityFramework to match how it is in the `feat/4.0.0` branch. 

wrt AspNetCore, the only issue I know of yet is that is uses `ConfigureFromConfigurationOptions<TOptions>`. Beyond that, I'm not sure. We'll need some integration tests at the very least. Maybe we should try to dogfood it though... rather than wait for the community to spot really basic stuff.

Issue #2792, opened 2023-11-05
Add cmake dependency to CONTRIBUTING.md


jamescrosswell:
I'm not sure under what conditions this needs to be installed. I had to install this manually on my mac:
```
brew install cmake
```

I think if folks are using Visual Studio (most likely on Windows) it would be installed with certain components for Unity, C++, Windows API development etc.

So maybe we put a note in CONTRIBUTING.md saying people either need to install those components with their visual studio installation, on Windows, or install cmake separately (e.g. using `brew` on macOS)?

PR #2795, opened 2023-11-06
Added CMake as a dependency to CONTRIBUTING.md

jamescrosswell:
Closes https://github.com/getsentry/sentry-dotnet/issues/2792

#skip-changelog

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -21,9 +21,13 @@ For big feature it's advised to raise an issue to discuss it first.
 
   *If using an M1 ("Apple silicon") processor, read [the special instructions below](#special-instructions-for-apple-silicon-cpus).*
 
-* On Windows: [.NET Framework](https://dotnet.microsoft.com/download/dotnet-framework) 4.6.2 or higher.
+* On Windows: 
+  - [.NET Framework](https://dotnet.microsoft.com/download/dotnet-framework) 4.6.2 or higher.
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -21,9 +21,13 @@ For big feature it's advised to raise an issue to discuss it first.
 
   *If using an M1 ("Apple silicon") processor, read [the special instructions below](#special-instructions-for-apple-silicon-cpus).*
 
-* On Windows: [.NET Framework](https://dotnet.microsoft.com/download/dotnet-framework) 4.6.2 or higher.
+* On Windows: 
+  - [.NET Framework](https://dotnet.microsoft.com/download/dotnet-framework) 4.6.2 or higher.
+  - [C++ CMake tools for Windows](https://learn.microsoft.com/en-us/cpp/build/cmake-projects-in-visual-studio?view=msvc-170#installation)
```
vaind, line 26: note: I don't mind having this text, just saying it's not accurate - you just need `cmake` in path, it doesn't need to come from VS

vaind (approved):

CONTRIBUTING.md
jamescrosswell: I'm happy to reword to make it more accurate. Mainly I just wanted to help new contributors understand:
1. They need CMake installed (and on the path)
2. How to install it if they don't have it

We could say, "you'll need it on the path... if you don't have it, one way to get it is by installing the C++ CMake tools for Windows". If there's an easier or lighter weight approach we could recommend, happy with that as well.

Issue #2273, opened 2023-03-30
Improve `SentryEvent` public API


mattjohnsonpint:
### Problem Statement

The `SentryEvent` is generally allowed to be mutated.  Most properties have setters, or other mechanisms for modification.  This is useful when creating a new event to capture manually, and also when modifying an event via `BeforeSend`.

However, we have some inconsistencies with regards to treatment of dictionaries.

Specifically:

- `Breadcrumbs` can only have items added via `AddBreadcrumb`.  The API doesn't provide a direct way to modify or remove breadcrumbs once they are on the event. (`Breadcrumb` objects are immutable.)
  - `BeforeBreadcrumb` can be used to keep them off the event, but there may be need to modify in `BeforeSend` as well, based on other criteria specific to the event.
  - One could cast to `List<Breadcrumb>` or `IList<Breadcrumb>` manually, but that is brittle and shouldn't be required.
- `Extra` can only have items added via `SetExtra`.
  - You can set `null` to overwrite an existing item, but the null will still end up in the data.
  - One could cast to `Dictionary<string, object?>` or `IDictionary<string, object?>` manually, but that is brittle and shouldn't be required.
- `Tags` has both `SetTag` and `UnsetTag` methods, so it is mutable for all practical reasons.  However, it's not clear why these methods are needed, rather than just exposing `IDictionary<string, string>`.  It also makes it cumbersome to clear all entries at once.
  - One can cast to `Dictionary<string, string>` or `IDictionary<string, string>` manually, but that is brittle and shouldn't be required.

### Solution Brainstorm

In next major release (4.0.0), change the public API of `SentryEvent` to expose all collection and dictionary properties on `SentryEvent` as `IList<T>` or `IDictionary<TKey, TValue>`.

We can then remove the `AddBreadcrumb`, `SetExtra`, `SetTag`, and `UnsetTag` methods.

The only other non-mutable properties would be:

- `EventId`, `Timestamp`, and `Sdk` - which should probably remain immutable.
- `Exception`, which we might consider adding a setter for.

### Investigation

On the Scope, Breadcrumbs are stored internally in a `ConcurrentQueue<Breadcrumb>`... presumably because once we hit `SentryOptions.MaxBreadcrumbs` we start throwing away the old ones. For this reason, I think it makes sense to leave Breadcrumbs as they are. 

[...]

PR #2796, opened 2023-11-06
Tags and Extra exposed as IDictionary (instead of IReadOnlyDictionary)

jamescrosswell:
Resolves https://github.com/getsentry/sentry-dotnet/issues/2273


src/Sentry/Internal/ScopeSyncingDictionary.cs
```diff
@@ -0,0 +1,116 @@
...
+{
+    private readonly IDictionary<K, V> _innerDictionary;
+    protected readonly SentryOptions Options;
+
+    public ScopeSyncingDictionary(
+        IDictionary<K, V> innerDictionary,
+        SentryOptions options
+        )
+    {
+        _innerDictionary = innerDictionary ?? throw new ArgumentNullException(nameof(innerDictionary));
```
bitsandfoxes, line 13: This is never getting caught, right? Should we log this instead?

bruno-garcia (changes requested):
Lets please not to this. `SetTag` is the API in all Sentry SDKs, and it's what we exposed for 5 and a half years here. We gain nothing in changing this. In fact I'd argue we lose in exposing the IDictionary API which exposes things in a way we might not have expected/planned

bruno-garcia:
We had a bit of a race condition here but I close this issue 2 days ago too: https://github.com/getsentry/sentry-dotnet/issues/2273

jamescrosswell:
> Lets please not to this. `SetTag` is the API in all Sentry SDKs, and it's what we exposed for 5 and a half years here. We gain nothing in changing this. In fact I'd argue we lose in exposing the IDictionary API which exposes things in a way we might not have expected/planned

I'm OK with that. It would be good to make these calls before writing the solution/implementation though. I was basically just going through the remaining issues on the 4.0.0 milestone... we could maybe do a bit of "backlog grooming" every now and then to avoid writing code that will never get used (like this).

PR #2799, opened 2023-11-06
fix: Native AOT resolution

vaind:
#skip-changelog

### DIFF
--- a/src/Sentry/Internal/AotHelper.cs
+++ b/src/Sentry/Internal/AotHelper.cs
@@ -11,18 +11,14 @@ internal static class AotHelper
 
     internal static bool IsAot { get; }
 
+    [UnconditionalSuppressMessage("Trimming", "IL2026:Members annotated with 'RequiresUnreferencedCodeAttribute' require dynamic access otherwise can break functionality when trimming application code", Justification = AotHelper.SuppressionJustification)]
     static AotHelper()
     {
-        try
-        {
-            // GetMethod should throw an exception if Trimming is enabled
-            var type = typeof(AotTester);
-            _ = type.GetMethod(nameof(AotTester.Test));
-        }
-        catch
-        {
-            IsAot = true;
-        }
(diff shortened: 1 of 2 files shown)


src/Sentry/Internal/AotHelper.cs
```diff
@@ -11,18 +11,14 @@ private class AotTester
...
-        {
-            // GetMethod should throw an exception if Trimming is enabled
-            var type = typeof(AotTester);
-            _ = type.GetMethod(nameof(AotTester.Test));
-        }
-        catch
-        {
-            IsAot = true;
-        }
+#if NET6_0_OR_GREATER   // TODO NET7 once we target it
```
bruno-garcia, line 17: We can keep this forever technically, since it'll include NET7 and in a way it helps document what's the min version this works from. Also avoid us having to go through all the code base and change stuff that won't do any change (I did that a bit on my last PR dropping stuff)

```suggestion
#if NET6_0_OR_GREATER
```

src/Sentry/Internal/AotHelper.cs
```diff
@@ -11,18 +11,14 @@ private class AotTester
...
+        var stackTrace = new StackTrace(false);
```
bruno-garcia, line 18: This is kind of expensive I believe. Even though the Exception approach is definitely not cheap and kind of dirty. And fires `FirstChanceException` so definitely prefer not to use that approach

bruno-garcia (approved):

jamescrosswell:
Sorry, @bruno-garcia I saw you'd approved this so I merged it in as it was related to another bug @vaind found and I wanted the changes... in hindsight, I should have left this open a while longer. Basically, the work continues here though: https://github.com/getsentry/sentry-dotnet/pull/2802

bruno-garcia:
My bad, I thought it made sense, didn't run the code

Issue #2788, opened 2023-11-04
Default JSON TypeInfoResolver


vaind:
Out of the box, Sentry.init() doesn't work for apps compiled with native AOT. The following console app fails to serialize with the error later below:
```c#
using Sentry;

// Initialize the Sentry SDK.  (It is not necessary to dispose it.)
SentrySdk.Init(options =>
{
    options.Dsn = Environment.GetCommandLineArgs()[1];
    options.Debug = true;
});

throw new ApplicationException("Something happened!");
```

```
System.InvalidOperationException: Reflection-based serialization has been disabled for this application. Either use the source generator APIs or explicitly configure the 'JsonSerializerOptions.TypeInfoResolver' property.
   at System.Text.Json.ThrowHelper.ThrowInvalidOperationException_JsonSerializerIsReflectionDisabled() + 0x2b
   at System.Text.Json.JsonSerializerOptions.ConfigureForJsonSerializer() + 0x31
   at System.Text.Json.JsonSerializer.GetTypeInfo(JsonSerializerOptions, Type) + 0x2a
   at System.Text.Json.JsonSerializer.GetTypeInfo[T](JsonSerializerOptions) + 0x25
   at Sentry.Internal.Extensions.JsonExtensions.InternalSerializeToUtf8Bytes(Object) + 0x6a
   at Sentry.Internal.Extensions.JsonExtensions.WriteDynamicValue(Utf8JsonWriter, Object, IDiagnosticLogger) + 0x41b
   at Sentry.Internal.Extensions.JsonExtensions.WriteDynamic(Utf8JsonWriter, String, Object, IDiagnosticLogger) + 0x4b
```

vaind:
Actually the AotHelper doesn't work at all. Fixed here: https://github.com/getsentry/sentry-dotnet/pull/2781/commits/300db5a39cd4ed79b6169b9467aa946a221fc364

vaind:
Actually, this is an issue, regardless of the AotHelper.

In JIT, the serialization throws above exception. This can be easily verified by running:
```
cd samples/Sentry.Samples.Console.Basic
dotnet run --framework net8.0
```

jamescrosswell:
Interesting. It doesn't occur when running `Sentry.Samples.Console.Basic` from Rider as part of the `Sentry-CI-Build-macOS` solution filter. 🤔 

[...]

PR #2802, opened 2023-11-07
Fixed JsonSerialization issues when Publishing to AOT

jamescrosswell:
Resolves https://github.com/getsentry/sentry-dotnet/issues/2788

Turns out our second revision of the AotHelper wasn't working either. I figured this time I'd try/catch something that was clearly impossible with serialization and AOT/Trimming and use that to determine whether we were AOT or JIT. 

This works with `Sentry.Samples.Console.Basic`. @vaind it'd be good to know if it also works in the scenario that you'd tested the previous logic (using stack traces) with... to make sure we're not just breaking one thing to fix another.

#skip-changelog

### DIFF
--- a/src/Sentry/Internal/AotHelper.cs
+++ b/src/Sentry/Internal/AotHelper.cs
@@ -1,3 +1,5 @@
+using Sentry.Internal.Extensions;
+
 namespace Sentry.Internal;
 
 internal static class AotHelper
@@ -11,12 +13,23 @@ internal static class AotHelper
 
     internal static bool IsAot { get; }
 
-    [UnconditionalSuppressMessage("Trimming", "IL2026:Members annotated with 'RequiresUnreferencedCodeAttribute' require dynamic access otherwise can break functionality when trimming application code", Justification = AotHelper.SuppressionJustification)]
+    [UnconditionalSuppressMessage("AOT", "IL3050:Calling members annotated with 'RequiresDynamicCodeAttribute' may break functionality when AOT compiling.", Justification = SuppressionJustification)]
+    [UnconditionalSuppressMessage("Trimming", "IL2026:Members annotated with 'RequiresUnreferencedCodeAttribute' require dynamic access otherwise can break functionality when trimming application code", Justification = SuppressionJustification)]
     static AotHelper()
     {
 #if NET6_0_OR_GREATER   // TODO NET7 once we target it
-        var stackTrace = new StackTrace(false);
-        IsAot = stackTrace.GetFrame(0)?.GetMethod() is null;
+        try
+        {
+            var json = JsonSerializer.Serialize(
+                new { a = "1" },
+                JsonExtensions.SerializerOptions
+            );
+            IsAot = !string.IsNullOrEmpty(json);
+        }
+        catch
+        {
+            IsAot = true;
+        }
 #else
         IsAot = false;
 #endif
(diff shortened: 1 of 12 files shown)


src/Sentry/Internal/AotHelper.cs
```diff
@@ -11,12 +13,23 @@ private class AotTester
...
+            var _ = JsonSerializer.Serialize(
+                new { a = "1" },
+                JsonExtensions.SerializerOptions
+            );
+            IsAot = false;
+        }
+        catch
+        {
+            IsAot = true;
+        }
```
vaind, line 32: > Turns out our second revision of the AotHelper wasn't working either.

In what scenario have you found it didn't work? I've tested the "fixed" code with the console run and it reported AOT vs JIT correctly, based on either publishing (AOT) or dotnet run or build (JIT).

jamescrosswell: > In what scenario have you found it didn't work? I've tested the "fixed" code with the console run and it reported AOT vs JIT correctly, based on either publishing (AOT) or dotnet run or build (JIT).

That was the scenario I ran as well 😕

vaind: OK, let's wait for #2781 which enables integration testing before moving along with this change.

jamescrosswell: It definitely fixes it though. Try running `dotnet run Sentry.Samples.Console.Basic.csproj --framework net8.0` in the `feat/4.0.0` branch and then try the same in this branch. The first one will complain about JsonSerialization issues (when I stepped through it in a debugger, AotHelper.IsAot was false, so it wasn't using the new JsonSerializers) and the second one all the serliazation works, because it uses this correctly:
https://github.com/getsentry/sentry-dotnet/blob/6679952d9dc846f109dbfc96512981854ae8f16b/src/Sentry/Internal/Extensions/JsonExtensions.cs#L887

vaind: >  when I stepped through it in a debugger, AotHelper.IsAot was false

that is always expected unless you're using a native debugger (gdb) with a published executable.

jamescrosswell: > that is always expected unless you're using a native debugger (gdb) with a published executable.

Bloody twilight zone. I'm sure when I stepped through the new code in a debugger 60 minutes ago, it returned true... and now not. In any case, I see what you're saying. 

> Also, the new code can report IsAot=true for any exception thrown by the serialization, regardless of being AOT.

It could... but what is the scenario where it throws an exception in JIT mode? We could add an Xunit test to make sure it always returns false in JIT if we were worried about that. 

Otherwise, if we don't think the serialization test is going to be reliable enough, for some reason, there may be other's we can use. For example, [`Type. MakeGenericType` won't work with Trimming enabled](https://learn.microsoft.com/en-us/dotnet/core/deploying/native-aot/fixing-warnings#examples-of-aot-warnings), but I'm not sure I see a compelling reason to change the existing solution (which works).

vaind: I've found the issue and fixed it in #2810

bruno-garcia:
I think we can close this now that https://github.com/getsentry/sentry-dotnet/pull/2810 is merged?

src/Sentry/Internal/AotHelper.cs
jamescrosswell: ⭐ Nice! Where did you find that @vaind ? Docs, Blogs or trawling through source code? In any event, way better to have some platform capability we can rely on than to wire up our own stuff.

Issue #2276, opened 2023-03-31
Consistent transaction name in ASP.NET Core projects


SeanFeldman:
A transaction name with ASP.NET Core should follow the following structure: `HTTP-VERB /path`.

For example, for a controller, resulting in a URL https://localhost/UserStatus/GetCurrentStatus

```
[ApiController]
[Route("[controller]")]
public class UserStatusController : ControllerBase
{
    [HttpGet]
    [Route("[action]")]
    public Task<string> GetCurrentStatus(string userId)
    {
        throw new Exception("oops");
    }
}
```

The transaction name should be **`GET /UserStatus/GetCurrentStatus`**, not `GET UserStatus/GetCurrentStatus`.

For an endpoint, resulting in a URL https://localhost/sad-endpoint

```
// program.cs, minimal API with .NET 6
app.MapGet("sad-endpoint/", _ => throw new Exception("oy vey"));
```

The transaction name should be **`GET /sad-endpoint`**, not `GET sad-endpoint/` as it is today.

The inconsistent naming is happening with Sentry.AspNetCore 3.29.1

mattjohnsonpint:
I agree this should be consistent, and the leading slash should definitely be there.  I'm not so sure about the trailing slash.

Our developer docs shows a slash on both ends:
https://develop.sentry.dev/sdk/event-payloads/transaction/#transaction-annotations

The screenshot in our product docs shows transactions from other SDKs with the leading slash, but not the trailing slash
https://docs.sentry.io/product/performance/

Either way, changing them from whatever they presently are would affect issue grouping for the majority of users, so we'll need to wait until a major release (v4.0) to make this change.

Thanks!

PR #2808, opened 2023-11-08
Consistent names for Transactions representing ASP.NET Core routes

jamescrosswell:
Resolves https://github.com/getsentry/sentry-dotnet/issues/2276

### DIFF
--- a/src/Sentry.AspNetCore/RouteUtils.cs
+++ b/src/Sentry.AspNetCore/RouteUtils.cs
@@ -36,7 +36,8 @@ internal static class RouteUtils
         }
 
         // Force a leading slash (if there isn't already one present)
-        return $"/{builder.ToString().TrimStart('/')}";
+        var url = builder.ToString();
+        return url.Length>0 && url[0] == '/' ? url : $"/{url}";
     }
 
     // Internal for testing.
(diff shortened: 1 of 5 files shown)


src/Sentry.AspNetCore/RouteUtils.cs
```diff
@@ -35,7 +35,8 @@ internal static class RouteUtils
             builder.Append(routePattern);
         }
 
-        return builder.ToString();
+        // Force a leading slash (if there isn't already one present)
+        return $"/{builder.ToString().TrimStart('/')}";
```
bruno-garcia, line 39: hm this seems lazy, can we check if `/` is there instead of adding and trimming?

bruno-garcia, line 39: something like: 
```suggestion
        var url = builder.ToString();
        return url.Length>0 && url[0] == '/' ? url : $"/{url}";
```

jamescrosswell: Looks computationally less expensive

src/Sentry.AspNetCore/RouteUtils.cs
```diff
@@ -35,7 +35,9 @@ internal static class RouteUtils
             builder.Append(routePattern);
         }
 
-        return builder.ToString();
+        // Force a leading slash (if there isn't already one present)
+        var url = builder.ToString();
+        return url.Length>0 && url[0] == '/' ? url : $"/{url}";
```
bruno-garcia, line 40:
```suggestion
        return url.Length >0 && url[0] == '/' ? url : $"/{url}";
```

bruno-garcia (approved):

PR #2810, opened 2023-11-08
fix: net8 JSON serialization

### LINKED ISSUE #2788: Default JSON TypeInfoResolver

vaind:
closes #2788

See comments in the linked issue for details (and also https://devblogs.microsoft.com/dotnet/system-text-json-in-dotnet-8/). In short, we must check `JsonSerializer.IsReflectionEnabledByDefault` instead of relying on "IsAOT" and also it's only available since net8.0 so we must add a new TF.

- fix: json serialzation on JIT net8 when PublishAot=true (requires adding a net8.0 target for Sentry)
- refactor: rename IsAot to IsNativeAot

TODO: tests (depends on #2791)

#skip-changelog - this hasn't been released yet

### DIFF
--- a/src/Sentry/Internal/Extensions/JsonExtensions.cs
+++ b/src/Sentry/Internal/Extensions/JsonExtensions.cs
@@ -551,15 +551,18 @@ internal static class JsonExtensions
 
     private static byte[] InternalSerializeToUtf8Bytes(object value)
     {
-        return AotHelper.IsAot
-            ? AotSerializeToUtf8Bytes()
-            : JitSerializeToUtf8Bytes();
-
+#if NET8_0_OR_GREATER
         byte[] AotSerializeToUtf8Bytes()
         {
             var context = GetSerializerContext(value.GetType());
             return JsonSerializer.SerializeToUtf8Bytes(value, value.GetType(), context);
         }
+        return JsonSerializer.IsReflectionEnabledByDefault
+            ? JitSerializeToUtf8Bytes()
+            : AotSerializeToUtf8Bytes();
+#else
+        return JitSerializeToUtf8Bytes();
+#endif
 
         [UnconditionalSuppressMessage("Trimming", "IL2026:Members annotated with 'RequiresUnreferencedCodeAttribute' require dynamic access otherwise can break functionality when trimming application code", Justification = AotHelper.SuppressionJustification)]
         [UnconditionalSuppressMessage("AOT", "IL3050:Calling members annotated with 'RequiresDynamicCodeAttribute' may break functionality when AOT compiling.", Justification = AotHelper.SuppressionJustification)]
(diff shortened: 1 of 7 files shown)


bruno-garcia (approved):

Daniel15:
Is there a nightly build I can use before this fix is published to production?

vaind:
> Is there a nightly build I can use before this fix is published to production? Trying to enable AOT for one of my apps and I hit this issue :)

There's a prerelease on nuget https://www.nuget.org/packages/Sentry/4.0.0-beta.0

src/Sentry/Internal/Extensions/JsonExtensions.cs
```diff
@@ -551,15 +551,18 @@ private static JsonSerializerContext GetSerializerContext(Type type, bool preser
 
     private static byte[] InternalSerializeToUtf8Bytes(object value)
     {
-        return AotHelper.IsAot
-            ? AotSerializeToUtf8Bytes()
-            : JitSerializeToUtf8Bytes();
-
+#if NET8_0_OR_GREATER
```
jamescrosswell, line 554: @vaind I think this implicitly means we need to change when we support AOT compilation:
https://github.com/getsentry/sentry-dotnet/blob/ebaed3e131d7346ce15ad28a11d37bcd30a99b0a/src/Directory.Build.props#L38-L41

If the serialization only works on net8.0 or greater then the framework only supports AOT on net8.0 or greater as well right (due to trimming)?

bruno-garcia: Ah that's right AOT is around since .NET 7, .NET 8 added macOS support

vaind: This condition must be for net8 because of a new API it uses.

bruno-garcia: Good to know we have net8-only code already to validate the need to bump it.
That alone is a reason to add net8 to main packages that can be added alone (and hence bringing the net8 version of Sentry)

PR #2815, opened 2023-11-09
Refactor: move DebugImage and DebugMeta to Sentry.Protocol

vaind:
Reasoning: some of the protocol classes (those sent to Sentry, mostly POCO) are residing directly in the Sentry namespace. Maybe we should move them to Protocol, although, it's a case-by-case decision because some may be used too frequently by SDK users so the migration wouldn't be worth any small benefit from the cleanup done here.

Let me know if you think any of these should move, or if I should abandon this completely. The list is a simple search of `: IJsonSerializable`
```
C:\dev\dotnet\src\Sentry\Breadcrumb.cs
  11,32: public sealed class Breadcrumb : IJsonSerializable

C:\dev\dotnet\src\Sentry\Package.cs
  9,29: public sealed class Package : IJsonSerializable

C:\dev\dotnet\src\Sentry\PersistedSessionUpdate.cs
  6,39: internal class PersistedSessionUpdate : IJsonSerializable

C:\dev\dotnet\src\Sentry\Request.cs
  28,29: public sealed class Request : IJsonSerializable

C:\dev\dotnet\src\Sentry\SdkVersion.cs
  11,32: public sealed class SdkVersion : IJsonSerializable

C:\dev\dotnet\src\Sentry\SentryMessage.cs
  19,35: public sealed class SentryMessage : IJsonSerializable

C:\dev\dotnet\src\Sentry\SentryStackFrame.cs
  10,38: public sealed class SentryStackFrame : IJsonSerializable

C:\dev\dotnet\src\Sentry\SentryStackTrace.cs
  43,31: public class SentryStackTrace : IJsonSerializable

C:\dev\dotnet\src\Sentry\SentryThread.cs
  10,34: public sealed class SentryThread : IJsonSerializable

C:\dev\dotnet\src\Sentry\SentryValues.cs
  9,39: internal sealed class SentryValues<T> : IJsonSerializable

C:\dev\dotnet\src\Sentry\User.cs
  10,26: public sealed class User : IJsonSerializable

C:\dev\dotnet\src\Sentry\UserFeedback.cs
  9,34: public sealed class UserFeedback : IJsonSerializable

C:\dev\dotnet\src\Sentry\ViewHierarchy.cs
  9,39:     public sealed class ViewHierarchy : IJsonSerializable

C:\dev\dotnet\src\Sentry\ViewHierarchyNode.cs
  8,41: public abstract class ViewHierarchyNode : IJsonSerializable
  ```

### Note
There are also these types that implement IJsonSerializable in addition to other interfaces:
```
public sealed class Contexts : IDictionary<string, object>, IJsonSerializable
public sealed class SentryEvent : IEventLike, IJsonSerializable
public readonly struct SentryId : IEquatable<SentryId>, IJsonSerializable
public sealed class SentryStackFrame : IJsonSerializable
public class SessionUpdate : ISession, IJsonSerializable
public class Span : ISpanData, IJsonSerializable
public readonly struct SpanId : IEquatable<SpanId>, IJsonSerializable
public class Transaction : ITransactionData, IJsonSerializable
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -79,6 +79,7 @@ Additionally, we're dropping support for some of the old target frameworks, plea
 - `DebugImage.ImageAddress` changed to `long?`. ([#2725](https://github.com/getsentry/sentry-dotnet/pull/2725))
 - Contexts now inherits from `IDictionary` rather than `ConcurrentDictionary`. The specific dictionary being used is an implementation detail. ([#2729](https://github.com/getsentry/sentry-dotnet/pull/2729))
 - Transaction names for ASP.NET Core are now consistently named `HTTP-VERB /path` (e.g. `GET /home`). Previously the leading forward slash was missing for some endpoints. ([#2808](https://github.com/getsentry/sentry-dotnet/pull/2808))
+- `DebugImage` and `DebugMeta` moved to `Sentry.Protocol` namespace. ([#2815](https://github.com/getsentry/sentry-dotnet/pull/2815))
 
 ### Fixes
 
--- a/src/Sentry/Internal/DebugStackTrace.cs
+++ b/src/Sentry/Internal/DebugStackTrace.cs
@@ -2,6 +2,7 @@ using Sentry.Internal.Extensions;
 using Sentry.Extensibility;
 using Sentry.Native;
 using Sentry.Internal.ILSpy;
+using Sentry.Protocol;
 
 namespace Sentry.Internal;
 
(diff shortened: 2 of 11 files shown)


jamescrosswell:
Maybe a naive question, but what's the purpose/benefit of the Protocol namespace? Is it just to help organise/identify the all the classes that get sent across the wire to Sentry? If so then yeah, makes sense to move all of the above... hopefully it's just global usings changes for SDK users,  the most part.

bruno-garcia:
I'd prefer we optmize for the end user DX. Having to import another namespace just to send an event for example is not ideal, what's why some stuff was put in the root namespace, so `import Sentry` by itself is all you need.

bruno-garcia:
That said DebugMeta and stuff isn't really used a lot so it's fine to move, but we need to document a "before/after" so folks know what to when it breaks after bumping the SDK

bitsandfoxes (approved):

PR #2821, opened 2023-11-13
fix: don't add WinUI integration on mobile

vaind:
I've noticed we do this when testing NativeAOT on iOS

### DIFF
--- a/src/Sentry/Integrations/WinUIUnhandledExceptionIntegration.cs
+++ b/src/Sentry/Integrations/WinUIUnhandledExceptionIntegration.cs
@@ -1,4 +1,4 @@
-#if NET5_0_OR_GREATER
+#if NET5_0_OR_GREATER && !__MOBILE__
 using Sentry.Extensibility;
 using Sentry.Internal;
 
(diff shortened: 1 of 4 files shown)


src/Sentry/Integrations/WinUIUnhandledExceptionIntegration.cs
jamescrosswell, line 1: I think this should be OK. Technically you can [target WinUI 3 from a .NET MAUI app](https://learn.microsoft.com/en-us/windows/apps/windows-dotnet-maui/) - I'm not sure what the value of `__MOBILE__` will hold in that case. It doesn't sound, logically, like a mobile app but possibly worth double checking.

vaind: I've tried running the MAUI sample on Widnows but it doesn't seem to work for me.

[Docs](https://learn.microsoft.com/en-us/xamarin/cross-platform/app-fundamentals/building-cross-platform-applications/platform-divergence-abstraction-divergent-implementation) say __MOBILE__ is defined for iOS and Android

jamescrosswell (approved):

PR #2824, opened 2023-11-13
chore: cleanup v4 changelog  

vaind:
I suggest you don't review the diff but the rendered changelog instead: https://github.com/getsentry/sentry-dotnet/blob/chore/changelog-cleanup/CHANGELOG.md#400-alpha0

#skip-changelog


bitsandfoxes (approved):
Bless you for this!

Issue #2744, opened 2023-10-20
Setting CaptureFailedRequests only affects the .NET layer


rodolfoBee:
### Package

Sentry.Maui

### .NET Flavor

.NET

### .NET Version

7.0.0

### OS

iOS

### SDK Version

3.40.1

### Steps to Reproduce

Create a simple app that has HTTP errors (400, 404...) on the ios layer
set `options.CaptureFailedRequests = false` in the .NET SDK init. 

### Expected Result

The expectation when setting CaptureFailedRequests = false on the .NET sdk is that this option will be passed on the native SDKs, thus the cocoa SDK will also not send HTTP failed requests events.

In other words, it will automatically set `enableCaptureFailedRequests = false` in the underlaying cocoa SDK

### Actual Result

The option is not propagated to cocoa SDK and HTTP failed requests that happen on the ios layer are sent to Sentry.

jamescrosswell:
Hi @rodolfoBee, thanks for creating an issue. That all makes sense. It might take us a little while to get around to this as we're currently preparing things for the .NET 8 release but we'll definitely look into this.

carmichaelalonso:
Hi, I created a PR to fix this. We initially reported the issue to @rodolfoBee via Sentry support, but noticed today that it was a simple fix. PR: https://github.com/getsentry/sentry-dotnet/pull/2826

PR #2826, opened 2023-11-13
fix: Support for CaptureFailedRequests on iOS

carmichaelalonso:
Fixes https://github.com/getsentry/sentry-dotnet/issues/2744

Setting either `CaptureFailedRequests` or `FailedRequestStatusCodes` previously only affected the .NET layer and was not passed to sentry-cocoa.

In production, this meant that failed requests were being tracked despite the setting being set to false in the .NET layer, which quickly added up in tracked errors on a hosted Sentry plan. This PR ensures we pass this value at init, and properly support failed status code ranges.


bitsandfoxes (approved):

bitsandfoxes:
This looks great, thanks for the fix!

PR #2831, opened 2023-11-13
fix: NativeAOT inApp setting

### LINKED ISSUE #2817: Native AOT in-app frames
Check what we sent as in-app from the runtime and how we show it on the server.

The thing is, we don't really know at runtime whether the frame is in-app or not.

If there's any change, we should try to do it prior to 4.0.0 release because it would be breaking in grouping.

vaind:
closes #2817 

#skip-changelog the original code hasn't been released yet (outside of alpha)

### DIFF
--- a/src/Sentry/SentryStackFrame.cs
+++ b/src/Sentry/SentryStackFrame.cs
@@ -182,7 +182,7 @@ public sealed class SentryStackFrame : IJsonSerializable
         {
             ConfigureAppFrame(options, Function, mustIncludeSeparator: true);
         }
-        else
+        else if (ImageAddress is null or 0 && InstructionAddress is null or 0) // Leave InApp=null on NativeAOT
         {
             InApp = true;
         }
--- a/integration-test/runtime.Tests.ps1
+++ b/integration-test/runtime.Tests.ps1
@@ -81,7 +81,7 @@ internal class FakeTransport : ITransport
     }
 
     It "sends stack trace native addresses" {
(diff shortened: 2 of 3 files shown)


src/Sentry/SentryStackFrame.cs
```diff
@@ -182,7 +182,7 @@ public void ConfigureAppFrame(SentryOptions options)
         {
             ConfigureAppFrame(options, Function, mustIncludeSeparator: true);
         }
-        else
+        else if (ImageAddress is null or 0 && InstructionAddress is null or 0) // Leave InApp=null on NativeAOT
```
jamescrosswell, line 185: Just curious... why does this condition imply it's not a NativeAOT app?

vaind: https://github.com/getsentry/sentry-dotnet/blob/feat/4.0.0/src/Sentry/Internal/DebugStackTrace.cs#L237

bitsandfoxes: I'm not all the way up to speed but a simple `0` works? The framework returns an `IntPtr`, the `InstructionAddress` is a `nint`.

bitsandfoxes (approved):

src/Sentry/SentryStackFrame.cs
vaind: it's nullable so I have to check for that first

PR #2837, opened 2023-11-14
chore: fixup changelog

vaind:
#skip-changelog

I've noticed the changelog link on the recent item points to an issue, not the PR so fixed that.

And while at it, I've applied the fixes to other linter issues that have been popping up forever.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ### Fixes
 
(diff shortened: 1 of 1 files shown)


bitsandfoxes (approved):

carmichaelalonso:
Sorry, my bad, only realized it had to be the PR link after it was merged. Thanks for cleaning it up!

Issue #972, opened 2021-05-07
Transaction should be enriched from the client


bruno-garcia:
A SentryClient can be used stand-alone. All event data is applied through the client.

This rule is broken for transactions:

https://github.com/getsentry/sentry-dotnet/blob/d25c97f0e5c77b520e30877444e93a9147afb9cd/src/Sentry/Internal/Hub.cs#L223

Tyrrrz:
To fix this, we need to expose `Scope` as a parameter to `CaptureTransaction(...)`, similarly to how it's done with `CaptureEvent(...)`:

https://github.com/getsentry/sentry-dotnet/blob/d25c97f0e5c77b520e30877444e93a9147afb9cd/src/Sentry/SentryClient.cs#L105

https://github.com/getsentry/sentry-dotnet/blob/d25c97f0e5c77b520e30877444e93a9147afb9cd/src/Sentry/SentryClient.cs#L63

bruno-garcia:
Yeah the client should be getting a scope in order to merge it too, like it does for CaptureEvent already.
Sadly it's a breaking change. This will affect only customers using the client directly and since they can't create transactions on it (only through SentrySdk.StartTransaction) this is low priority indeed.

bruno-garcia:
Lets add it to the next major yeah

mattjohnsonpint:
Before waiting for 4.0.0, we should also validate whether scoped data (ex: tags) makes its way through to transactions at all.  If not, that's a p1 bug to solve now.

PR #2838, opened 2023-11-14
feat: Enrich transactions

bitsandfoxes:
Fixes https://github.com/getsentry/sentry-dotnet/issues/972


src/Sentry/Internal/Hub.cs
```diff
@@ -473,34 +470,11 @@ public void CaptureTransaction(Transaction transaction, Hint? hint)
...
-                    if (processedTransaction == null)
-                    {
-                        _options.ClientReportRecorder.RecordDiscardedEvent(DiscardReason.EventProcessor, DataCategory.Transaction);
-                        _options.LogInfo("Event dropped by processor {0}", processor.GetType().Name);
-                        return;
-                    }
-                }
-            }
+            var (currentScope, client) = ScopeManager.GetCurrent();
+            // scope ??= currentScope;
```
bruno-garcia, line 474: delete?

src/Sentry/Internal/Hub.cs
```diff
@@ -473,34 +470,11 @@ public void CaptureTransaction(Transaction transaction, Hint? hint)
...
 
-            client.CaptureTransaction(processedTransaction, hint);
+            client.CaptureTransaction(transaction, currentScope, hint);
+            Console.WriteLine("HALP!");
```
bruno-garcia, line 477: we don't want this

bruno-garcia (commented):
quick pass

bruno-garcia (commented):
quick pass

src/Sentry/SentryClient.cs
```diff
@@ -137,7 +138,33 @@ public void CaptureTransaction(Transaction transaction, Hint? hint)
...
+
+        _options.LogInfo("Capturing transaction.");
+
+        scope.Evaluate();
+        scope.Apply(transaction);
+
+        _enricher.Apply(transaction);
+
+        var processedTransaction = transaction;
+        if (transaction.IsSampled != false)
```
bruno-garcia, line 153: is this nullable? The fact it's not `!IsSampled` indicates so and it means if value is `null` we'll also go in the block.

I see it's the original code copied but wonder

bruno-garcia (approved):

src/Sentry/SentryClient.cs
bitsandfoxes: It's not nullable.

PR #2839, opened 2023-11-14
ci: cache supplemental build

vaind:
While I'm not convinced this stuff is needed to be separate binaries, at least it shouldn't slow down CI (on Windows it's sometimes slow) because it never changes.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -84,14 +84,6 @@ jobs:
           fail-on-cache-miss: true
           enableCrossOsArchive: true
 
-      - name: Rebuild Solution Filters
-        shell: pwsh
-        run: scripts/generate-solution-filters.ps1
-
-      - name: Ensure Solution Filters are up to date
-        shell: pwsh
-        run: scripts/dirty-check.ps1 -PathToCheck ./*.sln* -GuidanceOnFailure "Uncommitted changes to the solution filters detected. Run `scripts/generate-solution-filters.ps1` locally and commit changes."
-
       - name: Setup Environment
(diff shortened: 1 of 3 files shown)


.github/workflows/build.yml
```diff
@@ -84,14 +84,6 @@ jobs:
...
           enableCrossOsArchive: true
 
-      - name: Rebuild Solution Filters
-        shell: pwsh
-        run: scripts/generate-solution-filters.ps1
-
-      - name: Ensure Solution Filters are up to date
-        shell: pwsh
-        run: scripts/dirty-check.ps1 -PathToCheck ./*.sln* -GuidanceOnFailure "Uncommitted changes to the solution filters detected. Run `scripts/generate-solution-filters.ps1` locally and commit changes."
-
```
vaind, line 94: somehow this got merged back but there's also a separate job that runs the code...

scripts/build-sentry-cocoa.sh
```diff
@@ -10,7 +10,6 @@ rm -rf Carthage
 sdks=$(xcodebuild -showsdks)
 ios_sdk=$(echo "$sdks" | awk '/iOS SDKs/{getline; print $NF}')
 ios_simulator_sdk=$(echo "$sdks" | awk '/iOS Simulator SDKs/{getline; print $NF}')
-macos_sdk=$(echo "$sdks" | awk '/macOS SDKs/{getline; print $NF}')
```
vaind, line 13: unused code

bitsandfoxes (approved):

PR #2840, opened 2023-11-15
Removed `IUserFactory` in favor of `ISentryUserFactory`

bitsandfoxes:
If I got that right:
Sentry's `ServiceCollectionExtension` adds the `HttpContextAccessor`.
https://github.com/getsentry/sentry-dotnet/blob/ad6968c04ecdc2775692dddf48852da7090aef2b/src/Sentry.AspNetCore/Extensions/DependencyInjection/ServiceCollectionExtensions.cs#L25

That way the `DefaultUserFactory` is able to access the context and no longer requires it to be passed down.

Also resolves https://github.com/getsentry/sentry-dotnet/issues/2844


jamescrosswell:
> That way the `DefaultUserFactory` is able to access the context and no longer requires it to be passed down.

Exactly, yeah. 

Might be worth checking if this is related, while we're making changes here:
* https://github.com/getsentry/sentry-dotnet/issues/2844

bitsandfoxes:
> Might be worth checking if this is related, while we're making changes here:

Yes. It coincidentally fixes this issue.

vaind (approved):

PR #2841, opened 2023-11-15
Removed obsolete options

bitsandfoxes:
Removed options marked as `[Obsolete]`.

### DIFF
--- a/src/Sentry/Platforms/Android/SentryOptions.cs
+++ b/src/Sentry/Platforms/Android/SentryOptions.cs
@@ -270,8 +270,8 @@ public partial class SentryOptions
         public bool EnableAndroidSdkTracing { get; set; } = false;
 
         /// <summary>
-        /// Gets or sets a value that indicates if the <see cref="o:SetBeforeSend"/> callback will be invoked for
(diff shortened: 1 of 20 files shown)


src/Sentry/Platforms/Android/SentryOptions.cs
```diff
@@ -270,7 +270,7 @@ public void AddInAppExclude(string prefix)
         public bool EnableAndroidSdkTracing { get; set; } = false;
 
         /// <summary>
-        /// Gets or sets a value that indicates if the <see cref="BeforeSend"/> callback will be invoked for
+        /// Gets or sets a value that indicates if the <see cref="o:SetBeforeSend"/> callback will be invoked for
```
vaind, line 273: IMO, It's still the "BeforeSend" callback, but it's set with `SetBeforeSend`.
```suggestion
        /// Gets or sets a value that indicates if the <see cref="o:BeforeSend"/> callback will be invoked for
```

vaind (approved):

jamescrosswell (approved):
Nice! So good to delete code 😁

PR #2852, opened 2023-11-16
feat: Add NET8 to major packages

vaind:
#skip-changelog

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -15,6 +15,9 @@
 
     <!-- Allow references to unsigned assemblies (like MAUI) from signed projects -->
     <NoWarn>$(NoWarn);CS8002</NoWarn>
+
+    <!-- Platform-agnostic TFMs that packages should target. -->
+    <SentryMainTargetFrameworks>net8.0;net6.0;netstandard2.1;netstandard2.0;net462</SentryMainTargetFrameworks>
   </PropertyGroup>
 
   <PropertyGroup Condition="'$(SolutionName)' == 'Sentry.NoMobile'">
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -15,9 +15,6 @@
 
     <!-- Allow references to unsigned assemblies (like MAUI) from signed projects -->
     <NoWarn>$(NoWarn);CS8002</NoWarn>
-
-    <!-- Platform-agnostic TFMs that packages should target. -->
-    <SentryMainTargetFrameworks>net8.0;net6.0;netstandard2.1;netstandard2.0;net462</SentryMainTargetFrameworks>
   </PropertyGroup>
 
   <PropertyGroup Condition="'$(SolutionName)' == 'Sentry.NoMobile'">
--- a/samples/Sentry.Samples.Console.Basic/Sentry.Samples.Console.Basic.csproj
+++ b/samples/Sentry.Samples.Console.Basic/Sentry.Samples.Console.Basic.csproj
@@ -4,7 +4,7 @@
     <OutputType>Exe</OutputType>
     <ImplicitUsings>enable</ImplicitUsings>
     <Nullable>enable</Nullable>
-    <TargetFrameworks>$(SentryMainTargetFrameworks)</TargetFrameworks>
+    <TargetFrameworks>net8.0;net6.0;netstandard2.1;netstandard2.0;net462</TargetFrameworks>
     <PublishAot Condition="$(TargetFramework.StartsWith('net8'))">true</PublishAot>
   </PropertyGroup>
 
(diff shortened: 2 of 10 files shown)


Directory.Build.props
```diff
@@ -15,6 +15,9 @@
 
     <!-- Allow references to unsigned assemblies (like MAUI) from signed projects -->
     <NoWarn>$(NoWarn);CS8002</NoWarn>
+
+    <!-- Platform-agnostic TFMs that packages should target. -->
+    <SentryMainTargetFrameworks>net8.0;net6.0;netstandard2.1;netstandard2.0;net462</SentryMainTargetFrameworks>
```
bruno-garcia, line 20: hm, I find a lot of value in seeing the TFMs on the projects instead of having to refer to this.
We usually don't need to think of aligning across many, it's a per TFM thing

src/Sentry.Log4Net/Sentry.Log4Net.csproj
```diff
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFrameworks>netstandard2.0;net462</TargetFrameworks>
+    <TargetFrameworks>$(SentryMainTargetFrameworks)</TargetFrameworks>
```
bruno-garcia, line 4: This is bad because now u're adding a bunch of new TFMs on a "lagacy" package

bruno-garcia (changes requested):

Directory.Build.props
vaind: My understanding was that when packages are being imported by consumers as a primary dependency, it needs to have the right TFMs. By defining these in a single place, we won't have an issue where a lower TFM from Sentry package would be pulled-in transitively.

src/Sentry.Log4Net/Sentry.Log4Net.csproj
vaind:
```suggestion
    <TargetFrameworks>netstandard2.0;net462</TargetFrameworks>
```

Directory.Build.props
bitsandfoxes: If we could keep the `MainTFMs` on all of them that'd be great. But because of the mono repo with the legacy stuff mixed in we have to special-case projects individually anyway. Let's be explicit and keep them on the projects.

vaind: ok, changed to explicit TFMs

bitsandfoxes (approved):

Directory.Build.props
bruno-garcia: Thank you

samples/Sentry.Samples.Console.Basic/Sentry.Samples.Console.Basic.csproj
```diff
@@ -4,8 +4,8 @@
     <OutputType>Exe</OutputType>
     <ImplicitUsings>enable</ImplicitUsings>
     <Nullable>enable</Nullable>
-    <TargetFramework>net8.0</TargetFramework>
-    <PublishAot>true</PublishAot>
+    <TargetFrameworks>net8.0;net6.0;netstandard2.1;netstandard2.0;net462</TargetFrameworks>
```
bruno-garcia, line 7: Was this a mistake? netstandard on an executable doesn't make sense to me

src/Sentry.NLog/Sentry.NLog.csproj
```diff
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFrameworks>net6.0;netstandard2.0;net462</TargetFrameworks>
+    <TargetFrameworks>net8.0;net6.0;netstandard2.1;netstandard2.0;net462</TargetFrameworks>
```
bruno-garcia, line 4: I was hoping we would get rid of ns2.1 if we can, but avoid adding it to more libraries. 
It's fair that the lib supports it and so does the Sentry core SDK. But the package is fairly lower in terms of usage: https://nugettrends.com/packages?months=72&ids=Sentry.NLog&ids=Sentry

And most people are updating their runtimes. .NET Standard 2.1 came to help bridge .NET Framework and .NET Core 3.1. We had .NET 5,6,7,8 now so hoping we can get rid of those netstandard

How about we remove this?

bruno-garcia: NLog doesn't have ns2.1: https://www.nuget.org/packages/NLog#dependencies-body-tab

samples/Sentry.Samples.Console.Basic/Sentry.Samples.Console.Basic.csproj
vaind: I've removed those on the feat/4.0.0 branch

PR #2857, opened 2023-11-17
chore: cleanup obsolete pragmas

vaind:
#skip-changelog

### DIFF
--- a/test/Sentry.Tests/SentrySdkTests.cs
+++ b/test/Sentry.Tests/SentrySdkTests.cs
@@ -80,22 +80,6 @@ public class SentrySdkTests : IDisposable
     }
 
     [Fact]
-    public void Init_ValidDsnWithSecret_EnablesSdk()
-    {
-        using var _ = SentrySdk.Init(o =>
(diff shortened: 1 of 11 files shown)


test/Sentry.Tests/SentrySdkTests.cs
```diff
@@ -79,22 +79,6 @@ public void Init_ValidDsn_EnablesSdk()
         Assert.True(SentrySdk.IsEnabled);
     }
 
-    [Fact]
-    public void Init_ValidDsnWithSecret_EnablesSdk()
```
bruno-garcia, line 83: This makes sure we don't regress on old DSNs. Which could still be out there

bruno-garcia (approved):
note on a test but honestly, it's been 5 years so don't mind removing it

test/Sentry.Tests/SentrySdkTests.cs
vaind: I was conflicted about this too, but it should go at some point, so...

PR #2862, opened 2023-11-18
fix: log only if transaction exists

bruno-garcia:
I see this on the log output:

```
  Debug: Ending transaction as Aborted, due to unhandled exception.
```

When no transaction is going on. Got me quite confused at first

#skip-changelog


vaind (approved):
Nice find

PR #2873, opened 2023-11-21
fix: aspnetcore dependency

bruno-garcia:
On `main` we don't have any dependencies, so this was introduced on net8.0:

https://github.com/getsentry/sentry-dotnet/blob/4d239a5156550b20b294e9b92968920f50f43240/src/Sentry.AspNetCore/Sentry.AspNetCore.csproj#L28

If that's the case, we don't need to add when running on net6.0/7.0.

If not, we should define the lowest version per TFM

#skip-changelog

### DIFF
--- a/src/Sentry.AspNetCore/Sentry.AspNetCore.csproj
+++ b/src/Sentry.AspNetCore/Sentry.AspNetCore.csproj
@@ -31,6 +31,13 @@
 
   <ItemGroup Condition="'$(TargetFramework)' != 'netstandard2.0'">
     <FrameworkReference Include="Microsoft.AspNetCore.App" />
+  </ItemGroup>
+
+  <!-- <ItemGroup Condition="'$(TargetFramework)' == 'net6.0'">
+    <PackageReference Include="Microsoft.Extensions.Configuration.Binder" Version="6.0.0" />
+  </ItemGroup> -->
+
+  <ItemGroup Condition="'$(TargetFramework)' == 'net8.0'">
     <PackageReference Include="Microsoft.Extensions.Configuration.Binder" Version="8.0.0" />
   </ItemGroup>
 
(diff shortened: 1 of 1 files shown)


src/Sentry.AspNetCore/Sentry.AspNetCore.csproj
```diff
@@ -31,6 +31,13 @@
 
   <ItemGroup Condition="'$(TargetFramework)' != 'netstandard2.0'">
     <FrameworkReference Include="Microsoft.AspNetCore.App" />
+  </ItemGroup>
+
+  <!-- <ItemGroup Condition="'$(TargetFramework)' == 'net6.0'">
+    <PackageReference Include="Microsoft.Extensions.Configuration.Binder" Version="6.0.0" />
+  </ItemGroup> -->
+
```
bruno-garcia, line 39: If CI works without this, just delete it:

```suggestion
  <!-- <ItemGroup Condition="'$(TargetFramework)' == 'net6.0'">
    <PackageReference Include="Microsoft.Extensions.Configuration.Binder" Version="6.0.0" />
  </ItemGroup> -->

```

jamescrosswell: Hm, very mysterious. I am wondering how CI can possibly work without that, as it depends on the source generated configuration bindings from `Microsoft.Extensions.Configuration.Binder`:
https://github.com/getsentry/sentry-dotnet/blob/89301cb5b90535e55972bf6aec197ec862afa91c/src/Sentry.AspNetCore/SentryAspNetCoreOptionsSetup.cs#L70

They are only used at design time to generate the `*.g.cs` classes I guess... still doesn't make sense to me though.

bruno-garcia: .NET 6 prob used reflection, and .NET 8 added source generator (get make this trimmable/AOT friendly)

jamescrosswell: That particular method (`Bind`) isnt' available via reflection (it's a generated method) and the codepath that uses it is when `!NETSTANDARD2_0` so it still doesn't compute for me 🤷🏻‍♂️

Issue #2872, opened 2023-11-20
Tests to ensure Bindable classes are consisten with Options classes


jamescrosswell:
To make the source generated configuration binders work, each of our Options classes has a matching BindableOptions class (e.g. `SentryOptions` ➡️ `BindableSentryOptions`). The source generated binders pull anything that's bindable into the Bindable classes, which take care of applying those config settings to the relevant Options classes. 

This works OK, but it's easy for these things to get out of sync. For example, we just [added a new property to SentryAspNetCoreOptions](https://github.com/getsentry/sentry-dotnet/pull/2871) in the main branch. If/when that gets merged into the feat/4.0.0 branch (or vice versa) all of our tests will pass and everything will look fine... but the `BindableSentryAspNetCoreOptions` class doesn't know anything about the new property and so we will have just broken our configuration binders 😿 

To prevent this from happening, we'd like to add some unit tests that loop through all the properties in the options classes and check that either these are not bindable (e.g. all the delegate properties) or that corresponding properties exist in the relevant BindableOptions class. 

We should do this in the 4.0.0 branch and then try merging in from main (once the PR above gets accepted and merged into main) and check to make sure our tests break to catch the issue.

PR #2875, opened 2023-11-21
Add Tests to ensure BindableOptions are consistent with Options

jamescrosswell:
#skip-changelog

Resolves https://github.com/getsentry/sentry-dotnet/issues/2872

I also added some reusable logic to test values get loaded from config and applied correctly to actual options classes by the bindingoptions classes... those tests are quite painful to write and maintain by hand otherwise.

### DIFF
--- a/test/Sentry.Testing/BindableTests.cs
+++ b/test/Sentry.Testing/BindableTests.cs
@@ -23,12 +23,15 @@ public abstract class BindableTests<TOptions>
 
     protected TextFixture Fixture { get; } = new();
 
-    private static IEnumerable<PropertyInfo> GetBindableProperties()
+    protected virtual IEnumerable<string> SkipProperties => Enumerable.Empty<string>();
+
+    private IEnumerable<PropertyInfo> GetBindableProperties()
     {
         return typeof(TOptions).GetProperties()
             .Where(p =>
                 !p.PropertyType.IsSubclassOf(typeof(Delegate)) // Exclude delegate properties
                 && !p.PropertyType.IsInterface // Exclude interface properties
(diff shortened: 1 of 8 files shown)


bitsandfoxes (approved):
Thanks for these tests! That will give us a lot more confidence when adding/changing the options! 🔥

test/Sentry.Testing/BindableTests.cs
```diff
@@ -0,0 +1,121 @@
+#if !NETFRAMEWORK
+using Microsoft.Extensions.Configuration;
+
+namespace Sentry.Testing;
+
+public abstract class BindableTests<TOptions>(params string[] skipProperties)
```
bitsandfoxes, line 6: Not gonna lie, that class made me scratch my head in a couple of places..

jamescrosswell: Yeah, originally the code was quite readable but there was a lot of duplication (the was a tonne of code for the unit tests themselves)... so I pulled a bunch of it back into the base class. That makes it easier to maintain (you should hardly ever have to touch the actual tests) but harder to understand... bit of a trade off. 

If this wasn't for testing, the actual test methods wouldn't exist in the derived classes (that could all be done in the base class) but then we can't surface any methods to decorate with `[Fact]` attributes... so you end up with a kind of weird base class that you have to extend in ways that wouldn't normally make sense (it's just to expose some stuff we can decorate with `[Fact]`). 

Additionally, I wanted to push some stuff into the Fixture, which can only be done in the constructor (so you can't use virtual/overloadable methods/properties for that).

Apologies, if you can find a nicer way to do it, I'm open to suggestions.

PR #2876, opened 2023-11-21
fix: upload Android native symbols

### LINKED ISSUE #2860: Android native crashes don't symbolicate - Upload native symbols via msbuild
Building an Android app uploads Portable PDBs only. So a native crash looks like this:

image

Manually running `sentry-cli debug-files upload .` on the proj root uploads (A LOT) of files including some needed to get better stack traces:

image

vaind:
fixes #2860

Sample event symbolicated after the upload hasa been fixed: https://sentry-sdks.sentry.io/issues/4654919326/events/6e0ef2059d4045e03f6549ef6b34a310/

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,13 @@
 
 ## Unreleased - 4.x
 
+### API breaking Changes
+
 - ISpanTracer has been renamed back again to ISpan, to make it easier to upgrade from v3.x to v4.x ([#2870](https://github.com/getsentry/sentry-dotnet/pull/2870))
+
+### Fixes
+
+- Android native symbol upload ([#2876](https://github.com/getsentry/sentry-dotnet/pull/2876))
 
 ## 4.0.0-beta.1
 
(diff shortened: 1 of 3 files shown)


bitsandfoxes (approved):
Nice! Thanks!

CHANGELOG.md
```diff
@@ -2,8 +2,14 @@
 
 ## Unreleased - 4.x
 
+### API breaking Changes
```
bruno-garcia, line 5: Technically this reverted the breakage, we can fix the final changelog before GA

bruno-garcia:
Thanks for fixing this!!

CHANGELOG.md
jamescrosswell: Yeah since we had separated out the changes into `-beta.1` etc. it was hard to know what to do... I didn't want to remove the breakage from the earlier alpha release as that would be misleading. But yeah, we can consolidate it all when we GA. 👍🏻

PR #2878, opened 2023-11-21
fix: Skipping package signing check during workload installation

bitsandfoxes:
Apparently, this is the way to go: https://github.com/dotnet/maui/issues/9573#issuecomment-1224546639
#skip-changelog

### DIFF
--- a/.github/actions/environment/action.yml
+++ b/.github/actions/environment/action.yml
@@ -43,7 +43,7 @@ runs:
           ${{ runner.os == 'macOS' && 'maui-ios maui-maccatalyst maui-windows maui-tizen' || '' }} \
           ${{ runner.os == 'Windows' && 'maui-windows' || '' }} \
           --temp-dir "${{ runner.temp }}" --from-rollback-file rollback.json \
-          -skip-sign-check
+          --skip-sign-check
 
     # We build Sentry.Maui for every supported MAUI target so we can access platform-specific features.
     # That includes Tizen. We don't need the entire Tizen SDK, but we do need the base Tizen workload.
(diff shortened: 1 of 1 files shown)


.github/actions/environment/action.yml
```diff
@@ -42,7 +42,8 @@ runs:
           maui-android \
           ${{ runner.os == 'macOS' && 'maui-ios maui-maccatalyst maui-windows maui-tizen' || '' }} \
           ${{ runner.os == 'Windows' && 'maui-windows' || '' }} \
-          --temp-dir "${{ runner.temp }}" --from-rollback-file rollback.json
+          --temp-dir "${{ runner.temp }}" --from-rollback-file rollback.json \
+          --skip-sign-check
```
bruno-garcia, line 46:
```suggestion
          --skip-sign-check
          # Skipping signing check above because: https://github.com/dotnet/maui/issues/9573
```

Unrelated but didnt' we remove rollback.json ?

bruno-garcia (commented):
that was on .NET 6, I'm surprised if this "hack" is needed in .NET 8

jamescrosswell (approved):

.github/actions/environment/action.yml
bitsandfoxes: We did on the `feat/4.0.0` branch

jamescrosswell: I wonder if this has something to do with the `--from-rollback-file` parameter. We weren't experiencing these problems on the `feat/4.0.0` branch where that parameter had been removed.

Issue #2866, opened 2023-11-19
Sentry.Serilog throws if a DSN is not passed even though SentrySdk should already be initialized


kanadaj:
### Package

Sentry

### .NET Flavor

.NET

### .NET Version

8.0.0

### OS

Any (not platform specific)

### SDK Version

4.0.0-beta.1

### Steps to Reproduce

1. Add Sentry.Serilog and Sentry.AspNetCore to the app
2. Init Sentry with `builder.ConfigureWebHostDefaults(webBuilder => webBuilder.UseSentry(...));```
3. Try adding serilog with `builder.Services.CongureLogging()`
4. Use `.WriteTo.Sentry()` with no arguments

### Expected Result

The Serilog `.WriteTo.Sentry()` call shouldn't throw if SentrySdk is already initialized in the HostBuilder with `.UseSentry()`

### Actual Result

Throws during startup with `ArgumentNullException`

bruno-garcia:
Should be an easy fix with SentrySdk.IsEnabled before throwing

kanadaj:
Do note that passing an empty string to `WriteTo.Sentry()` doesn't actually fix this since it uses the following code:
```cs
        if (!string.IsNullOrWhiteSpace(dsn))
        {
            sentrySerilogOptions.Dsn = dsn;
        }
```

jamescrosswell:
Relates to:
- https://github.com/getsentry/sentry-dotnet/pull/2655

jamescrosswell:
@kanadaj I think in this scenario, where you have already initialized Sentry via the generic host (and a call to `UseSentry`), the best option would be to pass a value of `false` in the `initializeSdk` argument when hooking Sentry up to Serilog. So something like:
```
builder.Host.UseSerilog(new LoggerConfiguration()
    .WriteTo.Sentry(initializeSdk: false)
    .CreateLogger());
```

That way, even though you're not supplying a DSN, it won't try to use that (null) value to initialise the SDK anyway. 

I will make a change to the code such that if you do provide a disabled DSN (rather than a null DSN) it won't throw... but in that scenario you'd be initializing Sentry with a DisabledHub and nothing would get sent through to Sentry (which I don't think is what you want). 

I'll also create another ticket to see if we can be a bit smart about how we set the default value of `initializeSdk`... If a call had already been made SentrySdk.Init or UseSentry, for example, we could default this to `false` and conversely if those methods hadn't already been called we could leave the default of `true`.

Issue #2494, opened 2023-07-18
Sentry should raise an exception if the DSN has not been configured


jamescrosswell:
### Problem Statement

At the moment, if the DSN is empty, you get a DisabledHub. But that's because we've decided to use ["" as the DSN for a disabled hub](https://github.com/getsentry/sentry-dotnet/blob/d6d531cb7d3699f7c4f1f7225433928fbcdab00d/src/Sentry/Constants.cs#L12).

This means the SentrySDK is unable to tell if someone forgot to supply the DSN or if they intentionally left it blank in order to disable Sentry.

### Solution Brainstorm

If the `DisableSdkDsnValue` was a non-empty string (e.g. `"disabled"`) then people could still disable Sentry via code/config. and we could then provide some useful guidance when initialising the SDK in cases where they forgot to supply the DSN (e.g. "Error: You forgot to supply a DSN. Sentry needs a DSN to do anything useful."). 

This is a breaking change, so would need to wait until the next major release.

bruno-garcia:
This `""` behavior is 10 years old and set in all SDks. Not sure this is worth doing. Usualy not passing the DSN means `null` is passed (DSN should be nullable). And in this case, throwing would be OK.

PR #2883, opened 2023-11-22
fix: Sentry.Serilog throws with a disabled DSN

jamescrosswell:
Resolves https://github.com/getsentry/sentry-dotnet/issues/2866

Passing a disabled DSN when initialising Sentry via the Serilog integration results in an `ArgumentNullException`:
```csharp
Log.Logger = new LoggerConfiguration().WriteTo.Sentry(dsn: "").CreateLogger();
``` 

## Note

I think the fix/workaround to the initial bug/issue that triggered this change is to pass `initializeSdk: false` into the `WriteTo.Sentry` extension method. Ideally this wouldn't be necessary and I've created a new issue https://github.com/getsentry/sentry-dotnet/issues/2884 to try to solve that.

### DIFF
--- a/src/Sentry.Serilog/SentrySinkExtensions.cs
+++ b/src/Sentry.Serilog/SentrySinkExtensions.cs
@@ -1,4 +1,7 @@
 // ReSharper disable once CheckNamespace - Discoverability
+
+using Constants = Sentry.Constants;
+
 namespace Serilog;
 
 /// <summary>
@@ -82,7 +85,7 @@ public static class SentrySinkExtensions
     /// </example>
     public static LoggerConfiguration Sentry(
         this LoggerSinkConfiguration loggerConfiguration,
-        string? dsn = null,
+        string? dsn = Constants.DisableSdkDsnValue,
         LogEventLevel minimumBreadcrumbLevel = LogEventLevel.Information,
         LogEventLevel minimumEventLevel = LogEventLevel.Error,
         IFormatProvider? formatProvider = null,
@@ -189,7 +192,7 @@ public static class SentrySinkExtensions
         bool? initializeSdk = null,
         Dictionary<string, string>? defaultTags = null)
     {
-        if (!string.IsNullOrWhiteSpace(dsn))
+        if (dsn != null)
         {
             sentrySerilogOptions.Dsn = dsn;
         }
--- a/src/Sentry.Serilog/SentrySinkExtensions.cs
+++ b/src/Sentry.Serilog/SentrySinkExtensions.cs
@@ -1,7 +1,5 @@
 // ReSharper disable once CheckNamespace - Discoverability
 
-using Constants = Sentry.Constants;
-
 namespace Serilog;
 
 /// <summary>
--- a/src/Sentry.Serilog/SentrySinkExtensions.cs
+++ b/src/Sentry.Serilog/SentrySinkExtensions.cs
@@ -85,7 +85,7 @@ public static class SentrySinkExtensions
     /// </example>
     public static LoggerConfiguration Sentry(
         this LoggerSinkConfiguration loggerConfiguration,
-        string? dsn = Constants.DisableSdkDsnValue,
+        string? dsn = null,
         LogEventLevel minimumBreadcrumbLevel = LogEventLevel.Information,
         LogEventLevel minimumEventLevel = LogEventLevel.Error,
         IFormatProvider? formatProvider = null,
--- a/src/Sentry.Serilog/SentrySinkExtensions.cs
+++ b/src/Sentry.Serilog/SentrySinkExtensions.cs
@@ -192,7 +192,7 @@ public static class SentrySinkExtensions
         bool? initializeSdk = null,
         Dictionary<string, string>? defaultTags = null)
     {
-        if (dsn != null)
+        if (dsn is not null)
         {
             sentrySerilogOptions.Dsn = dsn;
         }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@
 ### Fixes
 
 - Android native symbol upload ([#2876](https://github.com/getsentry/sentry-dotnet/pull/2876))
-- Sentry.Serilog no longer throws if a DSN is not specified ([#2883](https://github.com/getsentry/sentry-dotnet/pull/2883))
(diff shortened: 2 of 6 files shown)


src/Sentry.Serilog/SentrySinkExtensions.cs
```diff
@@ -189,7 +192,7 @@ public static class SentrySinkExtensions
         bool? initializeSdk = null,
         Dictionary<string, string>? defaultTags = null)
     {
-        if (!string.IsNullOrWhiteSpace(dsn))
+        if (dsn != null)
```
bruno-garcia, line 195: prefer pattern matching over operator

```suggestion
        if (dsn is not null)
```

bruno-garcia (commented):
downside of this fix is that adding Sentry on Serilog without specifying a DSN it will silently do nothing. While before (before we started changing this) you needed to specify `""` explicitly.

vaind:
> downside of this fix is that adding Sentry on Serilog without specifying a DSN it will silently do nothing. While before (before we started changing this) you needed to specify `""` explicitly.

Maybe we should revert the original PR that broke this, i.e. throwing for an empty DSN. I now see that the issue was an internal suggestion with a strong counterargument #2494 that doesn't seem to be resolved, at least not in the issue.

jamescrosswell:
> > downside of this fix is that adding Sentry on Serilog without specifying a DSN it will silently do nothing. While before (before we started changing this) you needed to specify `""` explicitly.
> 
> Maybe we should revert the original PR that broke this, i.e. throwing for an empty DSN. I now see that the issue was an internal suggestion with a strong counterargument #2494 that doesn't seem to be resolved, at least not in the issue. It is still time to revert this without user headaches because a final release isn't out.

I think that goes against the principle of fail early. 

We got frequent support requests from people who couldn't get Sentry working because they hadn't supplied a DSN... so not only did we not fail early, but people got all the way passed "should I give up on this" to contact our support channel for something that should/would have been really obvious if we had have just thrown an exception for what is obviously a misconfiguration. 

I get that this will be a change for people who've been using the API for years, but for anyone who is new to Sentry, I think it's a small but big improvement. 

The alternative is to go back to failing silently and leaving people to scratch their heads and either give up or reach out to us on Discord/GitHub (taking hours or days to resolve something that could have been resolved in minutes).

jamescrosswell:
> downside of this fix is that adding Sentry on Serilog without specifying a DSN it will silently do nothing. While before (before we started changing this) you needed to specify `""` explicitly.

Yeah agreed, I reversed that particular part of the change. I think the solution is not to provide a fake/disabled DSN but to supply `initializeSdk: false` in the arguments if Sentry has already been initialised.

bruno-garcia:
@kanadaj thoughts?

kanadaj:
Any reason not to just print a warning or error if someone runs the code without a DSN configured?

bruno-garcia:
> I get that this will be a change for people who've been using the API for years, but for anyone who is new to Sentry, I think it's a small but big improvement.

We can turn back on people using this for 10 years or anything between now and then. The `""` behavior is in all SDKs and well documented understood. `""` isnt' default anywhere so in shares the intention of disabling things, without having to recompile code or whatever sometimes via config file.

> The alternative is to go back to failing silently and leaving people to scratch their heads and either give up or reach out to us on Discord/GitHub (taking hours or days to resolve something that could have been resolved in minutes).

Logging can help here. Even if Debug=false we could print here, without a way to turn it off.

jamescrosswell:
> Logging can help here. Even if Debug=false we could print here, without a way to turn it off.

We could... so make it a `LogError` rather than a `LogDebug`. I'm wondering why we'd do that instead of throwing an exception though. 

There are two scenarios:
1. Someone has supplied a disabled DSN: `""`... in this case they have deliberately configured a disabled DSN
2. No DSN has been supllied... in this case, it's _possible_ (if they're a long time user of Sentry) that they know the .NET SDK also treats that as a disabled DSN and this is a lazy way to provide it. However it's also possible (even likely, especially for new Sentry users) that they left out he DSN by mistake. In this scenario we can
    - Either silently fail... reverting to a disabled hub. Everything seems fine. The app runs but Sentry doesn't work. _If_ the user is smart enough to turn on `SentryOptions.Debug` and run the app again and check their logs (which pretty much only the experienced Sentry users know to do) they could work through that issue. More often though, either people give up or they reach out to us... at which point we tell them how to turn `SentryOptions.Debug` on and/or ask what value they're supplying for the DSN.
    - Or fail early... throwing an exception, with an error message explaining a DSN hasn't been supplied

Other than "existing users don't need to change anything", I'm not seeing any advantages of the silent fail approach over fail early. It seems like it's designed specifically to trip new users up and make it less likely their first experience with Sentry will be a positive one. 

We could do LogError instead of throwing an exception. But that error would never make it through to Sentry.io and it would kind of water down the whole fail early approach. Once again, we'd only be failing there in a way that people would notice if they checked their Debug output. Some people don't do that. Everyone notices when the application fails to start with an exception though. 

We can change this back, but I think it would be a step backwards for the SDK usability overall.

bruno-garcia:
I'm fine with throwing if `null` (I even reviewed and approved the original change). The issue was that the SDK was already initialized or so at least the user thought so. And this is making now obvious what we didn't realize: That the first call to Init (UseSentry) actually isn't the first one to run. And the DSN should be provided to Serilog instead of M.E.L.

How to communicate this to users is the challenge now. Should we bake this whole context in the exception message? Perhaps if we check loaded modules and see some Sentry.Extensions.Logging also loaded? And if only Sentry.Serilog then it's fine to crash as it's not this edge case?

jamescrosswell:
Yeah that all makes sense.

> How to communicate this to users is the challenge now.

I've added a link to your comment in the related issue where we hopefully deal with that.

bruno-garcia (approved):

kanadaj:
@jamescrosswell I just want to highlight the fact that `dsn` is an optional argument of `WriteTo.Sentry()`: https://github.com/getsentry/sentry-dotnet/blob/0ea0f0ace66ddf2a7d76235c590d36edd068d740/src/Sentry.Serilog/SentrySinkExtensions.cs#L86

So supplying the dsn is required, but the dsn argument is not? It's kind of super confusing as a user of the library.

PR #2887, opened 2023-11-22
Feat: native crash reporting

### LINKED ISSUE #2770: Native AOT - sentry-native crash reporting
### Problem Statement

We now integrate sentry-native for Native AOT builds (since v4.0.0). We could also initialize the native SDK to enable crash reporting. This should be done together with scope sync.

### Solution Brainstorm

We already did this (in c#) in sentry-unity and [this PR](https://github.com/getsentry/sentry-dotnet/pull/2625) tried to bring it here but it was trying to do a bit much so was ultimately split up.

vaind:
Closes #2770

## TODO:

* [ ] initial scope sync (followup issue?) - we cannot sync scope during init because all the contexts are only filled at the time a new .NET event is processed. 
* [ ] what do we want to do about the cache directory path? Defaults to the current working directory. Can we figure out a more reasonable default or do we want to document this? Or even warn unless explicitly specified?

Sample event: https://sentry-sdks.sentry.io/issues/4677655767/events/025a40ef472d40b2fa0c3011bf3c68bf/

Sample log output (second run)

```
  Debug: Setting native CacheDirectoryPath on Windows: C:\dev\dotnet\.sentry-native
  Debug: Logging enabled with ConsoleDiagnosticLogger and min level: Debug
  Debug: Setting Debug: True
  Debug: Disabling native auto session tracking
  Debug: Setting native CacheDirectoryPath on Windows: C:\dev\dotnet\.sentry-native
  Debug: Setting a native logger
  Debug: Initializing sentry native
   Info: [native] using database path "C:\dev\dotnet\.sentry-native"
  Debug: [native] starting transport
  Debug: [native] starting background worker thread
  Debug: [native] starting backend
  Debug: [native] background worker thread started
  Debug: [native] processing and pruning old runs
  Debug: [native] sending envelope
  Debug: [native] submitting task to background worker thread
  Debug: [native] executing task on worker thread
  Debug: [native] sending request using winhttp to "https://o447951.ingest.sentry.io:443/api/5428537/envelope/": x-sentry-auth:Sentry sentry_key=eb18e953812b41c3aeb042e666fd3b5c, sentry_version=7, sentry_client=sentry.native.dotnet/0.6.7 content-type:application/x-sentry-envelope content-length:10364
  Debug: Native SDK reported: 'crashedLastRun': 'True'
  Debug: Initializing Hub for Dsn: 'https://eb18e953812b41c3aeb042e666fd3b5c@o447951.ingest.sentry.io/5428537'.
  Debug: This looks like a NativeAOT application build.
  Debug: Starting BackgroundWorker.
  Debug: New scope pushed.
  Debug: Registering integration: 'AutoSessionTrackingIntegration'.
  Debug: Registering integration: 'AppDomainUnhandledExceptionIntegration'.
  Debug: Registering integration: 'AppDomainProcessExitIntegration'.
  Debug: Registering integration: 'UnobservedTaskExceptionIntegration'.
  Debug: Registering integration: 'SentryDiagnosticListenerIntegration'.
   Info: DiagnosticSource Integration is disabled because tracing is disabled.
  Debug: Registering integration: 'WinUIUnhandledExceptionIntegration'.
  Debug: BackgroundWorker Started.
  Debug: [native] received response: HTTP/1.1 200 OK Date: Tue, 28 Nov 2023 13:52:24 GMT Via: 1.1 google Content-Length: 41 Content-Type: application/json Server: nginx Vary: origin,access-control-request-method,access-control-request-headers access-control-allow-origin: * access-control-expose-headers: x-sentry-error,x-sentry-rate-limits,retry-after cross-origin-resource-policy: cross-origin x-envoy-upstream-service-time: 0 Strict-Transport-Security: max-age=31536000; includeSubDomains; preload Alt-Svc: h3=":443"; ma=2592000,h3-29=":443"; ma=2592000
  Debug: [native] request handled in 216ms
  Debug: Triggering a deliberate exception because SentrySdk.CauseCrash(CrashType.Native) was called
   Info: [native] entering signal handler
  Debug: [native] captured backtrace from ucontext with 17 frames
  Debug: [native] captured backtrace with 17 frames
  Debug: [native] merging scope into event
  Debug: [native] adding attachments to envelope
  Debug: [native] sending envelope
  Debug: [native] serializing envelope into buffer
   Info: [native] crash has been captured
```


src/Sentry/SentrySdk.cs
```diff
@@ -52,17 +48,40 @@ internal static IHub InitHub(SentryOptions options)
...
+
+        // We init the hub after native SDK in case the native init needs to adapt some options.
+        var hub = new Hub(options);
+
+        // Do the initial scope sync, if configured by the native integration.
+        if (options.NativeContextWriter is { } contextWriter)
+        {
+            hub.ConfigureScope((scope) =>
+            {
+                Task.Run(() => contextWriter.Write(scope)).ContinueWith(t =>
```
bruno-garcia, line 73: This time worth adding some context so we don't scratch our heads why we did Task.Run here:


```suggestion
                // Write context asynchronously to reduce overhead on `Init`. 
                // Any exception is logged to avoid UnobservedTaskException
                Task.Run(() => contextWriter.Write(scope)).ContinueWith(t =>
```

src/Sentry/SentrySdk.cs
```diff
@@ -52,17 +48,40 @@ internal static IHub InitHub(SentryOptions options)
...
+                {
+                    if (t.Exception is not null)
+                    {
+                        options.DiagnosticLogger?.LogWarning(
+                            "Failed to synchronize scope to the native SDK: {0}", t.Exception);
+                    }
+                });
```
bruno-garcia, line 80: Also I believe there's a continuation that only happens if the task failed. 
```suggestion
                }, TaskContinuationOptions.OnlyOnFaulted);
```

src/Sentry/Platforms/Native/Sentry.Native.targets
```diff
@@ -59,6 +59,7 @@
     Condition="'$(TargetFramework)' == 'net8.0' and '$(CI)' != 'true'"
     Inputs="$(SentryNativeBuildInputs)"
     Outputs="$(SentryNativeBuildOutputs)">
-    <Exec Command="pwsh $(SentryNativeBuildScript)" />
+    <!-- We want a "-Clean" because if the build script changes, previous cmake cache may contain invalid defines. -->
+    <Exec Command="pwsh $(SentryNativeBuildScript) -Clean" />
```
bruno-garcia, line 63: We need to callout on `CONTRIBUTING.md` that PowerShell Core, in particular having `pwsh` on path is a requirement

src/Sentry/SentrySdk.cs
```diff
@@ -626,29 +657,47 @@ public static void CauseCrash(CrashType crashType)
...
                 break;
+#elif NET8_0_OR_GREATER
+            case CrashType.Native:
+                if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
+                {
+                    NativeStrlenMSVCRT(IntPtr.Zero);
+                }
+                else
+                {
+                    NativeStrlenLibC(IntPtr.Zero);
```
bruno-garcia, line 683: should we compile the `.c` lib for Windows/macOS/Linux and have a crash that's also explicit "Crashed on purpose because ABC" ?

It's also nice to see the line number of the native function in the crash report (we also test that we upload debug files to make it work)

src/Sentry/Platforms/Native/Sentry.Native.targets
vaind: added the line

src/Sentry/SentrySdk.cs
vaind: I'm not sure that having these is worth all the work needed to make it work:
* c code
* compilation in CI and locally
* packaging
* linking for native aot

vaind: If anything, I'd say let's try to get sentry-native to include that code

bruno-garcia: that's a good idea, just add a Crash function to `sentry-native` like we have on `sentry-cocoa`

bruno-garcia:
>  what do we want to do about the cache directory path? Defaults to the current working directory. Can we figure out a more reasonable default or do we want to document this? Or even warn unless explicitly specified?

This is a problem on a Mac where the OS asks the user to give permission if the file is on Documents and others.
I wonder if using a consistent path name under `Path.GetTempPath` is an alternative. 

It's a problem on Unity today btw ^ so we could align the solution here

ALso a problem the fact we don't have paths by default. And the .NET SDK doesn't support having more than 1 instance of the app running at the same time (afaik not fixed) as worker transports would race to pick work off of the directory

samples/Sentry.Samples.Console.Native/Sentry.Samples.Console.Native.csproj
```diff
@@ -0,0 +1,38 @@
...
+    <SentryOrg>...your org...</SentryOrg>
+    <SentryProject>...your project...</SentryProject>
+    -->
+
+    <!--
+    After the above properties are configured, you can use the following features.
+    Uploading sources to Sentry during the build will enable Source Context in the Sentry issue details page.
+    Uploading symbols to Sentry will enable server-side symbolication (i.e. when the PDB is not present at runtime, or for AOT published programs).
+    -->
+    <SentryUploadSources>true</SentryUploadSources>
```
bruno-garcia, line 29: Is it working? I tested this recently and didn't work (maybe didn't open an issue) but I only got sources with the .NET flag to bundle sources in the PDB

src/Sentry/Internal/ContextWriter.cs
```diff
@@ -0,0 +1,78 @@
+namespace Sentry.Internal;
+
+/// <summary>
+/// Allows synchronizing Context from .NET to native layers.
+/// We're providing a single method that the implementations should override.
+/// They can choose to either have the single method directly in native using p/invoke,
+/// or use a more fine-grained interface, whatever is best for the platform.
+/// </summary>
+/// <remarks>
+/// WriteScope() is called in a new Task (background thread from a pool).
```
bruno-garcia, line 10:
```suggestion
/// WriteScope() is called in a new Task (background thread from a pool).
/// It's also only used once. For changes in the scope done during the execution, ScopeObserver is used
```

src/Sentry/Platforms/Native/CFunctions.cs
```diff
@@ -149,6 +239,17 @@ internal static void SetValueIfNotNull(sentry_value_t obj, string key, double? v
     internal static extern void sentry_remove_extra(string key);
 
     internal static Dictionary<long, DebugImage> LoadDebugImages(IDiagnosticLogger? logger)
+    {
+        // It only makes sense to load them once because they're cached on the native side anyway. We could force
+        // native to reload the list by calling sentry_clear_modulecache() when a dynamic library is loaded, but
+        // there's currently no way for us to know when that should happen.
+        DebugImages ??= LoadDebugImagesOnce(logger);
```
bruno-garcia, line 246: We should document this. And if this becomes a problem maybe we hook into the .NET assembly loading and invalidate any time that happens (as it could reference new native libs)?

src/Sentry/Platforms/Native/SentrySdk.cs
```diff
@@ -0,0 +1,60 @@
...
+public static partial class SentrySdk
+{
+    private static readonly Dictionary<string, bool> PerDirectoryCrashInfo = new();
+
+    private static void InitNativeSdk(SentryOptions options)
+    {
+        if (!C.Init(options))
+        {
+            options.DiagnosticLogger?
+                .LogWarning("Sentry native initialization failed - native crashes are not captured.");
```
bruno-garcia, line 16: We should look into if this happens

```suggestion
                .LogError("Sentry native initialization failed - native crashes will not captured. Managed code continues normally.");
```

src/Sentry/Scope.cs
```diff
@@ -127,13 +127,16 @@ public User User
         };
         set
         {
-            _user = value;
-            if (_user is not null)
+            if (_user != value)
```
bruno-garcia, line 130: lets keep it consistent (using pattern matching below and operator here):

```suggestion
            if (_user is not value)
```

src/Sentry/User.cs
```diff
@@ -26,8 +26,11 @@ public sealed class User : IJsonSerializable
         get => _id;
         set
         {
-            _id = value;
-            PropertyChanged?.Invoke(this);
+            if (_id != value)
```
bruno-garcia, line 29:
```suggestion
            if (_id is not value)
```

src/Sentry/User.cs
```diff
@@ -39,8 +42,11 @@ public sealed class User : IJsonSerializable
         get => _username;
         set
         {
-            _username = value;
-            PropertyChanged?.Invoke(this);
+            if (_username != value)
```
bruno-garcia, line 45:
```suggestion
            if (_username is not value)
```

src/Sentry/User.cs
```diff
@@ -52,8 +58,11 @@ public sealed class User : IJsonSerializable
         get => _email;
         set
         {
-            _email = value;
-            PropertyChanged?.Invoke(this);
+            if (_email != value)
```
bruno-garcia, line 61: We need a linter

```suggestion
            if (_email is not value)
```

src/Sentry/User.cs
```diff
@@ -65,8 +74,11 @@ public sealed class User : IJsonSerializable
         get => _ipAddress;
         set
         {
-            _ipAddress = value;
-            PropertyChanged?.Invoke(this);
+            if (_ipAddress != value)
```
bruno-garcia, line 77:
```suggestion
            if (_ipAddress is not value)
```

src/Sentry/User.cs
```diff
@@ -78,8 +90,11 @@ public sealed class User : IJsonSerializable
         get => _segment;
         set
         {
-            _segment = value;
-            PropertyChanged?.Invoke(this);
+            if (_segment != value)
```
bruno-garcia, line 93:
```suggestion
            if (_segment is not value)
```

src/Sentry/User.cs
```diff
@@ -91,8 +106,11 @@ public sealed class User : IJsonSerializable
         get => _other ??= new Dictionary<string, string>();
         set
         {
-            _other = value;
-            PropertyChanged?.Invoke(this);
+            if (_other != value)
```
bruno-garcia, line 109:
```suggestion
            if (_other is not value)
```

samples/Sentry.Samples.Console.Native/Program.cs
```diff
@@ -0,0 +1,35 @@
...
+
+// Initialize the Sentry SDK.  (It is not necessary to dispose it.)
+SentrySdk.Init(options =>
+{
+    // A Sentry Data Source Name (DSN) is required.
+    // See https://docs.sentry.io/product/sentry-basics/dsn-explainer/
+    // You can set it in the SENTRY_DSN environment variable, or you can set it in code here.
+    // options.Dsn = "... Your DSN ...";
+
+    // When debug is enabled, the Sentry client will emit detailed debugging information to the console.
```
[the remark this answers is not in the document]
vaind: Native backend uses offline caching to store the error at the time of the crash and only sends it later on the second app run.

samples/Sentry.Samples.Console.Native/Sentry.Samples.Console.Native.csproj
vaind: yes, it works AFAIK
image
image

src/Sentry/User.cs
```diff
@@ -42,7 +42,7 @@ public sealed class User : IJsonSerializable
         get => _username;
         set
         {
-            if (_username != value)
+            if (_username is not value)
```
bruno-garcia, line 45: Oh I read this as `null` checks somehow.

So we're checking `==` operator on the value we have stored + value set to spare the call to the native layer?

vaind: yes
I'll roll back

samples/Sentry.Samples.Console.Native/Program.cs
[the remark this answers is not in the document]
jamescrosswell: So is this also really an issue with SentryNative?
> what do we want to do about the cache directory path? Defaults to the current working directory.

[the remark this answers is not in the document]
vaind: If I get your question right, @jamescrosswell, then yes, this is only an issue with native crashes that are captured by sentry-native

src/Sentry/SentrySdk.cs
vaind: https://github.com/getsentry/sentry-native/issues/913

src/Sentry/Platforms/Native/CFunctions.cs
vaind: https://github.com/getsentry/sentry-dotnet/issues/2922

PR #2888, opened 2023-11-22
delete global.json

bruno-garcia:
craft checks out the branch it publishes and pinning 7.0 fails because we already bumped 8.0 on it.

Removing global.json from main

https://github.com/getsentry/publish/actions/runs/6962570877/job/18946542580

#skip-changelog

### DIFF
--- a/global.json
+++ b/global.json
@@ -1,7 +1,0 @@
-{
-  "sdk": {
-    "version": "7.0.100",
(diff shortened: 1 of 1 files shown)


bruno-garcia:
Best if install net7.0 on the craft docker image: https://github.com/getsentry/craft/pull/506

Issue #2886, opened 2023-11-22
FrameworkSupportsAot should only be set for .net8 and higher


jamescrosswell:
@vaind I think this implicitly means we need to change when we support AOT compilation:
https://github.com/getsentry/sentry-dotnet/blob/ebaed3e131d7346ce15ad28a11d37bcd30a99b0a/src/Directory.Build.props#L38-L41

If the serialization only works on net8.0 or greater then the framework only supports AOT on net8.0 or greater as well right (due to trimming)?

_Originally posted by @jamescrosswell in https://github.com/getsentry/sentry-dotnet/pull/2810#discussion_r1398510528_

vaind:
what is this "FrameworkSupportsAot" anyway? google doesn't turn up anything

vaind:
Personally, while I've been an advocate of trying to support net7, I'm all for switching of all the net7 AOT support in favor of net8+ because of this and other differences. We haven't supported NativeAOT for net7 when it came out so no reason to try to add support now that net8 is out. My assumption is that people running NativeAOT on net7 are going to try to update to net8 to get all the optimizations introduced there.

bruno-garcia:
That works, given that it's a major version, new feature (for us), .NET 8 **is LTS**

jamescrosswell:
> what is this "FrameworkSupportsAot" anyway? google doesn't turn up anything
> 
> update: we set `IsAotCompatible` based on that, so that's what this issue is about.

Yeah, it's just a way to centralise the logic for when we enable IsAotCompatible, so we don't have to copy/paste this all over the place every time it changes (like now). Once the AOT stuff settles down, we might be able to remove it but so far it's been good to have the logic for this defined in one place. 

I'll add some docs to the property to make this clear.

PR #2889, opened 2023-11-22
Only set FrameworkSupportsAot for net8.0+

jamescrosswell:
#skip-changelog

Resolves https://github.com/getsentry/sentry-dotnet/issues/2886

No Changelog required as this is already consistent with our description for AOT support in the Changelog.

### DIFF
--- a/src/Directory.Build.props
+++ b/src/Directory.Build.props
@@ -35,8 +35,8 @@
     <CodeAnalysisRuleSet>$(MSBuildThisFileDirectory)CodeAnalysis.ruleset</CodeAnalysisRuleSet>
   </PropertyGroup>
 
-  <!-- Only compile with IsAotCompatible on .NET 6.0 and later. Our Device Tests fail for net7.0-android with this enabled as well -->
-  <PropertyGroup Condition="'$(TargetFramework)' != 'net7.0-android' AND $([System.Text.RegularExpressions.Regex]::IsMatch('$(TargetFramework)', '^(?!net5\.|net4\d{2}|netstandard)net\d+'))">
+  <!-- Only compile with IsAotCompatible on .NET 8.0 and later -->
+  <PropertyGroup Condition="$([System.Text.RegularExpressions.Regex]::IsMatch('$(TargetFramework)', 'net8\.|net9\.|net\d{2,}\.'))">
     <FrameworkSupportsAot>true</FrameworkSupportsAot>
   </PropertyGroup>
 
--- a/src/Directory.Build.props
+++ b/src/Directory.Build.props
@@ -37,6 +37,11 @@
 
   <!-- Only compile with IsAotCompatible on .NET 8.0 and later -->
   <PropertyGroup Condition="$([System.Text.RegularExpressions.Regex]::IsMatch('$(TargetFramework)', 'net8\.|net9\.|net\d{2,}\.'))">
+    <!-- `FrameworkSupportsAot` is a custom property we use to centralise this logic for enabling `IsAotCompatible`. As
+    of very recently, this logic is fairly simple (net8.0 or later) but it's historically been more complex and very
+    volatile (as we learn more about the constraints of targeting AOT compilation). We may be able to remove this custom
+    property once we have confidence that this logic is unlikely to change again.
+    -->
     <FrameworkSupportsAot>true</FrameworkSupportsAot>
   </PropertyGroup>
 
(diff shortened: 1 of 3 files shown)


src/Sentry.Extensions.Logging/Sentry.Extensions.Logging.csproj
```diff
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFrameworks>net6.0;netstandard2.0</TargetFrameworks>
+    <TargetFrameworks>net8.0;net6.0;netstandard2.0</TargetFrameworks>
```
bruno-garcia, line 4: oh big save! we need this one. 

Best do a review of all projects before we close 4.0

src/Sentry.Extensions.Logging/Sentry.Extensions.Logging.csproj
```diff
@@ -20,12 +20,15 @@
     <PackageReference Include="Microsoft.Extensions.Http" Version="2.1.0" />
   </ItemGroup>
 
-  <ItemGroup Condition="'$(TargetFramework)' == 'net6.0'">
-    <PackageReference Include="Microsoft.Extensions.Configuration.Binder" Version="8.0.0" />
+  <ItemGroup Condition="'$(TargetFramework)' != 'netstandard2.0'">
```
bruno-garcia, line 23:
```suggestion
  <ItemGroup Condition="'$(TargetFramework)' == 'net6.0'">
```

You want this one to be specific to 6.0.0 (since those are the min version u're supporting there).

But on the 8.0 block add the same libraries but with 8.0.0 as a min version.

src/Sentry.Extensions.Logging/Sentry.Extensions.Logging.csproj
jamescrosswell: Currently the following projects **do not** target `net8.0`:

- Sentry.Android.AssemblyReader.csproj
- Sentry.AspNet.csproj
- Sentry.AspNetCore.Grpc.csproj
- Sentry.Bindings.Android.csproj
- Sentry.Bindings.Cocoa.csproj
- Sentry.DiagnosticSource.csproj
- Sentry.EntityFramework.csproj
- Sentry.Google.Cloud.Functions.csproj
- Sentry.Log4Net.csproj
- Sentry.Maui.csproj

I think we've had this conversation before... some of those are legacy (e.g. AspNet, Log4Net, and EntityFramework). Anything that needs to support AOT will need a net8.0 target though - including the bindings presumably? 

I'll create a new issue from this comment for the 4.0.0 milestone.

bruno-garcia (approved):

src/Directory.Build.props
```diff
@@ -35,8 +35,13 @@
     <CodeAnalysisRuleSet>$(MSBuildThisFileDirectory)CodeAnalysis.ruleset</CodeAnalysisRuleSet>
   </PropertyGroup>
 
-  <!-- Only compile with IsAotCompatible on .NET 6.0 and later. Our Device Tests fail for net7.0-android with this enabled as well -->
-  <PropertyGroup Condition="'$(TargetFramework)' != 'net7.0-android' AND $([System.Text.RegularExpressions.Regex]::IsMatch('$(TargetFramework)', '^(?!net5\.|net4\d{2}|netstandard)net\d+'))">
+  <!-- Only compile with IsAotCompatible on .NET 8.0 and later -->
+  <PropertyGroup Condition="$([System.Text.RegularExpressions.Regex]::IsMatch('$(TargetFramework)', 'net8\.|net9\.|net\d{2,}\.'))">
```
vaind, line 39: net9 and netXY? Aren't we getting ahead of ourselves? 😆

vaind (commented):
Thx. I'll update NativeAOT conditions

src/Directory.Build.props
jamescrosswell: Yeah probably... I was trying to write it so it wouldn't need to be changed next year when net9.0 is released... and then I added another 30 years or so runway after that (vast overkill) 😛 In reality, we're probably going to have to change it in a week or two when we discover it doesn't work with net8.0-android or something.

jamescrosswell:
> Thx. I'll update NativeAOT conditions

Yeah you can see I removed `'$(TargetFramework)' != 'net7.0-android'` from the Condition... work works at the moment because the bindings projects aren't targeting net8.0 but I'm not sure if the Device Tests will all pass on Android with AOT enabled, once we bump these to net8.0.

PR #2892, opened 2023-11-23
chore: support NativeAOT only on net8+

vaind:
Follows up on #2889 getting rid of all the conditions to enable partial support of NativeAOT on net7. The support hasn't been out other than on a pre-release and .net7 is not an LTS release and it's unlikely many projects would choose to keep using it with NativeAOT now that net8 is out and improves it greatly. Also see discussions on #2886

#skip-changelog

### DIFF
--- a/integration-test/cli.Tests.ps1
+++ b/integration-test/cli.Tests.ps1
@@ -4,7 +4,6 @@ $ErrorActionPreference = 'Stop'
 . $PSScriptRoot/common.ps1
 
 Describe 'Console apps (<framework>) - normal build' -ForEach @(
-    @{ framework = "net7.0" },
     @{ framework = "net8.0" }
 ) {
     BeforeAll {
@@ -41,7 +40,6 @@ Describe 'Console apps (<framework>) - normal build' -ForEach @(
 }
 
 Describe 'Console apps (<framework>) - native AOT publish' -ForEach @(
-    @{ framework = "net7.0" },
     @{ framework = "net8.0" }
 ) {
     BeforeAll {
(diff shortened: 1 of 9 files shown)


integration-test/cli.Tests.ps1
```diff
@@ -52,10 +50,10 @@ Describe 'Console apps (<framework>) - native AOT publish' -ForEach @(
...
     }
 
-    It "uploads symbols and sources" -Skip:($IsMacOS -and $framework -eq 'net7.0') {
+    It "uploads symbols and sources" {
         $result = RunDotnetWithSentryCLI 'publish' 'console-app' $True $True $framework
         $result.ScriptOutput | Should -AnyElementMatch "Preparing upload to Sentry for project 'console-app'"
-        if ($IsWindows -or ($IsLinux -and $framework -eq 'net7.0'))
+        if ($IsWindows)
         {
             $extension = $IsLinux ? '' : '.pdb'
```
jamescrosswell, line 58: I guess the confidtional is redundant now eh (inside this block it's always windows so always a pdb)?

vaind: right, I've missed the linux bit, thanks

src/Sentry/Internal/AotHelper.cs
```diff
@@ -9,16 +9,17 @@ private class AotTester
...
     static AotHelper()
     {
-#if NET6_0_OR_GREATER   // TODO NET7 once we target it
         var stackTrace = new StackTrace(false);
         IsNativeAot = stackTrace.GetFrame(0)?.GetMethod() is null;
+    }
 #else
-        IsNativeAot = false;
+    // This is a compile-time const so that the irrelevant code is removed during compilation.
+    internal const bool IsNativeAot = false;
```
jamescrosswell, line 23: Better!

jamescrosswell (approved):
Looks all good, presuming all the tests pass 👍🏻

Issue #2863, opened 2023-11-18
Crash free rate is broken


bruno-garcia:
![image](https://github.com/getsentry/sentry-dotnet/assets/1633368/5016afea-132f-480b-ad98-4d99531bbf6c)

I clearly have 1 crash event (unhandled) but still Sentry tells me 100% crash free.

In the logs I see one session update with `"status":"abnormal"` right after the crash, and without `init=true` and one without `status` at all (when `init=true` so might default to `ok` as it's a new session automatically started after the crash).

It's odd we automatically restart it because the process is crashing. We should check if it's an unhandled exception and not automatically start a fresh session.

I used options:

```
// Crash free run statistics (sends a ping for healthy and for crashes sessions)
o.AutoSessionTracking = true;

o.Debug = true;

// Caches files in the event device is offline
// Sentry creates a 'sentry' sub directory, no need to concat here
o.CacheDirectoryPath = appFolderInfo.GetAppDataPath();
```

And threw an exception right after Sentry was initialized

Relevant debug log bits:

```
  Debug: Ending session as Crashed, due to unhandled exception.
Warning: Failed to end session because there is none active.
[Fatal] ConsoleApp: EPIC FAIL! 
...
[v10.0.0.32241] System.Exception: test mechanism
   at NzbDrone.Console.ConsoleApp.Main(String[] args) in ./Sonarr.Console/ConsoleApp.cs:line 44
...
  Debug: Envelope successfully sent. Content: {"sdk":{"name":"sentry.dotnet","version":"4.0.0-beta.0"},"sent_at":"2023-11-18T22:54:46.450585+00:00"} {"type":"session","length":315} {"sid":"e5f3daf681bf47c793afb5bb3c771ba9","did":"82d98c16-cd54-46da-bb69-2b6ae4b9359d","init":false,"started":"2023-11-18T17:43:15.001553-05:00","timestamp":"2023-11-18T22:54:45.550827+00:00","seq":1,"duration":690,"errors":0,"status":"abnormal","attrs":{"release":"10.0.0.31593-Debug-dev","environment":"testing"}}
  Debug: Reading cached envelope: /Users/bruno/.config/Sonarr/Sentry/6B5D58608E674AC26BAF3CF583889982F0C7847F/__processing/1700348085_6036__51269976.envelope
  Debug: Sending cached envelope:
[...]

Issue #2551, opened 2023-08-19
Crash Free rate in MAUI seems broken


bruno-garcia:
SDK: 3.35.1

New .NET MAUI, Added Sentry. Added button to crash (took from sample in this repo).
Running on Android.

Crashed the app a few times. Crash free rate says 100%

https://github.com/getsentry/sentry-dotnet/assets/1633368/096275dc-391e-4dff-8391-45dc8e577335

Project: https://bruno-test.sentry.io/issues/?project=4505687395270656&query=release%3Acom.companyname.mauiapp2%401.0%2B1+error.unhandled%3Atrue&sort=freq

The iOS Crashes seem to work properly: 
![image](https://github.com/getsentry/sentry-dotnet/assets/1633368/585902c9-d588-4f43-8da2-28a4219633d7)

jamescrosswell:
Possibly something to do with [Session Reporting](https://develop.sentry.dev/sdk/sessions/).

Needs investigation.

bitsandfoxes:
Resolves by #2895

bruno-garcia:
Still missing tests? 

https://github.com/getsentry/sentry-dotnet/pull/2895#issuecomment-1824955349

This regressed a few times already

PR #2895, opened 2023-11-23
fix: Sessions not getting finished

bitsandfoxes:
Fixes https://github.com/getsentry/sentry-dotnet/issues/2863 and https://github.com/getsentry/sentry-dotnet/issues/2551

With https://github.com/getsentry/sentry-dotnet/pull/2496 we moved updating the session to the `client` to be able to align with the rest of the SDKs regarding the order of filtering and whether dropped events affect the session.

But we forgot to pass the `GlobalSessionManager` from the `hub` to the `client`. So both the `hub` and the `client` had their own `GlobalSessionManager`.

Ideally, I'd like to fix this properly by keeping the SessionManager solely on the `hub` and passing the session at the time down to the scope during capture instead of holding the state in the client. But with https://github.com/getsentry/rfcs/pull/122 on the horizon, this might be a waste of time and effort.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Fixes
+
+- The SDK no longer fails to finish sessions while capturing an event. This fixes broken crash-free rates ([#2895](https://github.com/getsentry/sentry-dotnet/pull/2895))
+
 ### Dependencies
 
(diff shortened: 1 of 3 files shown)


bruno-garcia (approved):

bruno-garcia:
We could probably benefit some integration tests eventually. But a fix should go out asap for this

Issue #2885, opened 2023-11-22
Remove SentrySinkExtensions.ConfigureSentrySerilogOptions


jamescrosswell:
This method is redundant:
https://github.com/getsentry/sentry-dotnet/blob/65a7bba00a73db6c619aa5d4a44ca4da2092b738/src/Sentry.Serilog/SentrySinkExtensions.cs#L165

It just shuffles parameters between two of the other extension methods (one that accepts specific options as parameters and another that accepts an `Action<SentrySerilogOptions>`. 

It's public, requires lots of documented properties and will have to change anytime one of the other extension methods changes... so it's just a pain to maintain. 

At the very least, we could make this method internal prior to going live with v4.0.0. That way, we can deal with it after the v4.0.0 release without it being a breaking change. 

It's a relatively small change though (mostly refactoring some unit tests to use NSubstitute)... so maybe we can squeeze it in.

PR #2902, opened 2023-11-27
Made SentrySinkExtensions.ConfigureSentrySerilogOptions internal

jamescrosswell:
The first part of https://github.com/getsentry/sentry-dotnet/issues/2885 - just makes this internal so that later (when we have time post the 4.0.0 release) we can remove this and refactor the relevant unit tests. 

There are 3 extension methods in [this class](https://github.com/getsentry/sentry-dotnet/blob/b9ffcfff5dfe9a069ffe60b145219c705aedf11b/src/Sentry.Serilog/SentrySinkExtensions.cs). They're all variants to allow the SDK user to set one or more properties on the `SentrySerilogOptions`. 
1. The first method let's you pass each option as an argument/parameter to the extension method
2. The second method does the same but instead of passing the properties separately, you pass a `ConfigureSentrySerilogOptions` class that contains each of the properties as nullables (instead of optional named arguments)
3. The third method takes an `Action<SentrySerilogOptions>` so you can do whatever you want

There's nothing you can do with the first two methods that can't be done with the 3rd... Although it's maybe nice to be able to call `WriteTo().Sentry("mydsn")` instead of having to wrap that in an action, so I can see an argument for retaining No. 1 and No. 3. The second extension method does nothing useful... the only place it's used is in our unit tests (and we can easily refactor that code)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+- SentrySinkExtensions.ConfigureSentrySerilogOptions is now internal. This change shouldn't impact SDK users as the method was likely only ever used for testing. ([#2902](https://github.com/getsentry/sentry-dotnet/pull/2902)) 
+
 ## 4.0.0-beta.2
 
 ### Fixes
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## Unreleased
 
-- SentrySinkExtensions.ConfigureSentrySerilogOptions is now internal. This change shouldn't impact SDK users as the method was likely only ever used for testing. ([#2902](https://github.com/getsentry/sentry-dotnet/pull/2902)) 
+- SentrySinkExtensions.ConfigureSentrySerilogOptions is now internal. If you were using this method, please use one of the `SentrySinkExtensions.Sentry` extension methods instead. ([#2902](https://github.com/getsentry/sentry-dotnet/pull/2902)) 
 
 ## 4.0.0-beta.2
 
--- a/src/Sentry.Serilog/SentrySinkExtensions.cs
+++ b/src/Sentry.Serilog/SentrySinkExtensions.cs
@@ -135,35 +135,7 @@ public static class SentrySinkExtensions
             defaultTags));
     }
 
-    /// <summary>
-    /// Configure the Sentry Serilog Sink.
-    /// </summary>
-    /// <param name="sentrySerilogOptions">The logger configuration to configure with the given parameters.</param>
-    /// <param name="dsn">The Sentry DSN. <seealso cref="SentryOptions.Dsn"/></param>
-    /// <param name="minimumEventLevel">Minimum log level to send an event. <seealso cref="SentrySerilogOptions.MinimumEventLevel"/></param>
-    /// <param name="minimumBreadcrumbLevel">Minimum log level to record a breadcrumb. <seealso cref="SentrySerilogOptions.MinimumBreadcrumbLevel"/></param>
-    /// <param name="formatProvider">The Serilog format provider. <seealso cref="IFormatProvider"/></param>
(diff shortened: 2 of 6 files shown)


bitsandfoxes (commented):
This was added in https://github.com/getsentry/sentry-dotnet/pull/380.
What changed that we no longer want to provide `ConfigureSentrySerilogOptions`?

bruno-garcia (commented):
Can we have the motivation to this change in the description? I'm not sure why it was added but it's a public API so saying "only used in testing" that's from our point of view, not sure what use cases it solved adding it.

WIth better understanding of what this helps solve, it'll be easier to communicate why we're changing to someone who might be affected

jamescrosswell:
There are 3 extension methods in [this class](https://github.com/getsentry/sentry-dotnet/blob/b9ffcfff5dfe9a069ffe60b145219c705aedf11b/src/Sentry.Serilog/SentrySinkExtensions.cs). They're all variants to allow the SDK user to set one or more properties on the `SentrySerilogOptions`. 
1. The first method let's you pass each option as an argument/parameter to the extension method
2. The second method does the same but instead of passing the properties separately, you pass a `ConfigureSentrySerilogOptions` class that contains each of the properties as nullables (instead of optional named arguments)
3. The third method takes an `Action<SentrySerilogOptions>` so you can do whatever you want

There's nothing you can do with the first two methods that can't be done with the 3rd... Although it's maybe nice to be able to call `WriteTo().Sentry("mydsn")` instead of having to wrap that in an action, so I can see an argument for retaining No. 1 and No. 3. The second extension method does nothing useful, that I can see...

jamescrosswell:
> Can we have the motivation to this change in the description?

Done. If anyone is using this method, they can easily use either of the other two extension methods in this class instead.

jamescrosswell:
> This was added in #380 and served a purpose other than tests.
> What changed that we no longer want to provide `ConfigureSentrySerilogOptions`?

@bitsandfoxes in the original PR, the properties that you could set via the two different extension methods were different (and only a small subset of the properties that can be set today). So originally, I see why Jim added a second extension method. There have been quite a few [changes made to this class since then](https://github.com/getsentry/sentry-dotnet/commits/main/src/Sentry.Serilog/SentrySinkExtensions.cs) though and today, there's no reason to have two separate methods.

bruno-garcia:
> > This was added in #380 and served a purpose other than tests.
> > What changed that we no longer want to provide `ConfigureSentrySerilogOptions`?
> 
> @bitsandfoxes in the original PR, the properties that you could set via the two different extension methods were different (and only a small subset of the properties that can be set today). So originally, I see why Jim added a second extension method. There have been quite a few [changes made to this class since then](https://github.com/getsentry/sentry-dotnet/commits/main/src/Sentry.Serilog/SentrySinkExtensions.cs) though and today, there's no reason to have two separate methods.

This makes sense. But the changelog says to me:

* This thing is internal
* Prob not a problem, used for testing

If I bump the SDK and it breaks because I was using that, CMD + F find the current changelog, I'd be very unhappy.
Could we point users to the method that should be used instead? Our believes that it's prob for testing are irrelevant for the end user I believe

CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+- SentrySinkExtensions.ConfigureSentrySerilogOptions is now internal. This change shouldn't impact SDK users as the method was likely only ever used for testing. ([#2902](https://github.com/getsentry/sentry-dotnet/pull/2902))
```
```suggestion
- SentrySinkExtensions.ConfigureSentrySerilogOptions was made internal. If you relied on it, please use XYZ instead. All the same arguments are available on it. ([#2902](https://github.com/getsentry/sentry-dotnet/pull/2902)) 
```

jamescrosswell:
> Could we point users to the method that should be used instead? Our believes that it's prob for testing are irrelevant for the end user I believe

Done 👌🏻

bruno-garcia (approved):

bruno-garcia:
Thanks James!!

PR #2906, opened 2023-11-27
CI Ignores changes to `.md` files

bitsandfoxes:
#skip-changelog

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -2,12 +2,12 @@ name: build
 
 on:
   push:
-    paths-ignore:
-      - '**.md'
     branches:
       - main
       - release/*
   pull_request:
+    paths-ignore:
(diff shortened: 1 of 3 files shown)


.github/workflows/build.yml
```diff
@@ -2,6 +2,8 @@ name: build
 
 on:
   push:
+    paths-ignore:
+      - '**.md'
```
vaind, line 6: we don't normally push to those branches so no point in ignoring here - maybe you wanted to update the `pull_request` conf?

bitsandfoxes: Ahh.. he who can read and all.. Thanks!

vaind (approved):

vaind:
LGTM, maybe also device-tests_

Issue #2738, opened 2023-10-18
Sentry.OpenTelemetry: The Sentry SDK has not been initialised. To use Sentry with OpenTelemetry tracing you need to initialize the Sentry SDK.


sinanbozkus:
### Package

Other

### .NET Flavor

.NET

### .NET Version

7.0.12

### OS

Linux

### SDK Version

7.0.0

### Steps to Reproduce

When I was using "Sentry.OpenTelemetry" 3.39.1 or older everything was ok. After updating 3.40.0 or 3.40.1 I got this error.

`The Sentry SDK has not been initialised. To use Sentry with OpenTelemetry tracing you need to initialize the Sentry SDK.
`

```
builder.Services.AddOpenTelemetry()
    .WithTracing(tracerProviderBuilder =>
            tracerProviderBuilder
                .AddSource("Development-Application.Api")
                .ConfigureResource(resource => resource.AddService("Development-Application.Api"))
                .AddAspNetCoreInstrumentation()
                .AddHttpClientInstrumentation()
                .AddSentry() // <-- Configure OpenTelemetry to send trace information to Sentry
    );

builder.WebHost.UseSentry(o =>
{
    o.Dsn = "XXXXXXXXX";
    o.TracesSampleRate = 1.0;
    o.UseOpenTelemetry();
});
```

### Expected Result

No error

### Actual Result

The Sentry SDK has not been initialised. To use Sentry with OpenTelemetry tracing you need to initialize the Sentry SDK.initialized

jamescrosswell:
Thanks for the report @sinanbozkus !

Are you able to try switching those statments, so the Sentry registration comes first?
```csharp
builder.WebHost.UseSentry(o =>
{
    o.Dsn = "XXXXXXXXX";
    o.TracesSampleRate = 1.0;
    o.UseOpenTelemetry();
});

builder.Services.AddOpenTelemetry()
    .WithTracing(tracerProviderBuilder =>
            tracerProviderBuilder
                .AddSource("Development-Application.Api")
                .ConfigureResource(resource => resource.AddService("Development-Application.Api"))
                .AddAspNetCoreInstrumentation()
                .AddHttpClientInstrumentation()
                .AddSentry() // <-- Configure OpenTelemetry to send trace information to Sentry
    );
```

sinanbozkus:
> , so the Sentry registration comes first

```
services.AddOpenTelemetry()
            .WithTracing(builder =>
                {
                    builder.AddSentry();
[...]

PR #2911, opened 2023-11-28
Fixed Sentry SDK has not been initialised when using ASP.NET Core, Serilog and OpenTelemetry

jamescrosswell:
Resolves https://github.com/getsentry/sentry-dotnet/issues/2738

# Problem

When using Sentry in an ASP.NET Core application, using SeriLog for logging and OpenTelemetry for tracing, the `SentrySpanProcessor` would throw an exception:
https://github.com/getsentry/sentry-dotnet/blob/84925a9da90197e88ef41013e80e028358b9951f/src/Sentry.OpenTelemetry/SentrySpanProcessor.cs#L44-L46

# Solution

When the logging backend is replaced by SeriLog, `SentrySdk.CurrentHub` doesn't get initialized until after `TracerProviderBuilderExtensions.AddSentry` gets run. I don't fully understand all the mechanics, to be honest - as far as I can tell, our ConfigureLogging code does actually get called, then it gets nuked by the SeriLog initialization, and finally our SDK gets initialized in by the `SentryStartupFilter`.

In any event, retrieving an `IHub` from DI rather than pulling it off the SentrySdk static class ensures we are always able to resolve this (regardless of whether the MEL backend has been replaced or not). 

I'd love a better explanation of what's going on here, if anyone else can provide one. 

# Backport
- https://github.com/getsentry/sentry-dotnet/pull/2918

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,10 @@
 
 - SentrySinkExtensions.ConfigureSentrySerilogOptions is now internal. If you were using this method, please use one of the `SentrySinkExtensions.Sentry` extension methods instead. ([#2902](https://github.com/getsentry/sentry-dotnet/pull/2902)) 
 
+### Fixes
+
+- Fixed Sentry SDK has not been initialised when using ASP.NET Core, Serilog and OpenTelemetry ([#2911](https://github.com/getsentry/sentry-dotnet/pull/2911))
+
 ### Dependencies
 
 - Bump Cocoa SDK from v8.16.1 to v8.17.0 ([#2910](https://github.com/getsentry/sentry-dotnet/pull/2910))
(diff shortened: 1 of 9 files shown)


CHANGELOG.md
```diff
@@ -2,8 +2,14 @@
 
 ## Unreleased
 
+### API breaking Changes
+
 - SentrySinkExtensions.ConfigureSentrySerilogOptions is now internal. If you were using this method, please use one of the `SentrySinkExtensions.Sentry` extension methods instead. ([#2902](https://github.com/getsentry/sentry-dotnet/pull/2902)) 
 
+### Fixes
+
+- Fixed Sentry SDK has not been initialised when using ASP.NET Core, SeriLog and OpenTelemetry ([#2911](https://github.com/getsentry/sentry-dotnet/pull/2911))
```
bruno-garcia, line 11: nit: 
image

```suggestion
- Fixed Sentry SDK has not been initialised when using ASP.NET Core, Serilog and OpenTelemetry ([#2911](https://github.com/getsentry/sentry-dotnet/pull/2911))
```

bruno-garcia (approved):

Issue #2847, opened 2023-11-15
Remove ConfigureFromConfigurationOptions<T> from Sentry.Azure.Functions.Worker


jamescrosswell:
Breaking https://github.com/getsentry/sentry-dotnet/issues/2822 down into more bite sized chunks. 

Potentially also relevant:
- https://github.com/getsentry/sentry-dotnet/issues/2525

Issue #2890, opened 2023-11-23
Review projects that should target `net8.0` prior to release


jamescrosswell:
Currently the following projects **do not** target `net8.0`:

## Need to target net8.0???
- Sentry.Android.AssemblyReader.csproj
- Sentry.AspNetCore.Grpc.csproj
- Sentry.Bindings.Android.csproj
- Sentry.Bindings.Cocoa.csproj
- Sentry.DiagnosticSource.csproj
- Sentry.Google.Cloud.Functions.csproj

## Legacy (won't target net8.0)
- Sentry.EntityFramework.csproj
- Sentry.AspNet.csproj
- Sentry.Log4Net.csproj

I think we've had this conversation before... some of those are legacy (e.g. AspNet, Log4Net, and EntityFramework). Anything that needs to support AOT will need a net8.0 target though - including the bindings presumably? 

I'll create a new issue from this comment for the 4.0.0 milestone.

_Originally posted by @jamescrosswell in https://github.com/getsentry/sentry-dotnet/pull/2889#discussion_r1402827468_

PR #2920, opened 2023-11-28
Removed ConfigureFromConfigurationOptions from Sentry.Azure.Functions.Worker

jamescrosswell:
#skip-changelog

Resolves https://github.com/getsentry/sentry-dotnet/issues/2847


src/Sentry.Azure.Functions.Worker/Sentry.Azure.Functions.Worker.csproj
```diff
@@ -6,6 +6,11 @@
       <Description>Official Azure Functions Worker SDK integration for Sentry - Open-source error tracking that helps developers monitor and fix crashes in real time.</Description>
     </PropertyGroup>
 
+    <PropertyGroup Condition="'$(FrameworkSupportsAot)' == 'true'">
+      <IsAotCompatible>true</IsAotCompatible>
+      <EnableConfigurationBindingGenerator>true</EnableConfigurationBindingGenerator>
```
bruno-garcia, line 11: Do we always set this when IsAotCompatible is true? If so, should we set this in the Directory.Build.Props where we define IsAotCompatible instead?

src/Sentry.Azure.Functions.Worker/SentryFunctionsWorkerMiddleware.cs
```diff
@@ -94,7 +96,11 @@ private async Task<TransactionContext> StartOrContinueTraceAsync(FunctionContext
         var httpMethod = requestData.Method.ToUpperInvariant();
         var transactionNameKey = $"{context.FunctionDefinition.EntryPoint}-{httpMethod}";
 
-        if (!TransactionNameCache.TryGetValue(transactionNameKey, out transactionName))
+        // Note that, when Trimming is enabled, we can't use reflection to read route data from the HttpTrigger
+        // attribute. In that case the route name will always be /api/<FUNCTION_NAME>
+        // If this is ever a problem for customers, we can potentially see if there are alternate ways to get this info
+        // from route tables or something. We're not even sure if anyone will use this functionality for now though. 
+        if (!AotHelper.IsNativeAot && !TransactionNameCache.TryGetValue(transactionNameKey, out transactionName))
```
bruno-garcia, line 103: For this AotHelper, instead of:

https://github.com/getsentry/sentry-dotnet/blob/217c11cea611a5f26e4a1abfbb7df37f34695722/src/Sentry/Internal/AotHelper.cs#L18-L19

Could we instead write an attribute at compile time? I imagine it's a lot cheaper (worth measuring) than trying to get a stack trace on a static constructor:

Something like this:
https://github.com/getsentry/sentry-dotnet/blob/217c11cea611a5f26e4a1abfbb7df37f34695722/src/Sentry/buildTransitive/Sentry.targets#L8-L29


Here's a PR that added writing the attribute at build time as a reference: https://github.com/getsentry/sentry-dotnet/pull/1739

src/Sentry.Azure.Functions.Worker/SentryFunctionsWorkerMiddleware.cs
```diff
@@ -94,7 +96,11 @@ private async Task<TransactionContext> StartOrContinueTraceAsync(FunctionContext
         var httpMethod = requestData.Method.ToUpperInvariant();
         var transactionNameKey = $"{context.FunctionDefinition.EntryPoint}-{httpMethod}";
 
-        if (!TransactionNameCache.TryGetValue(transactionNameKey, out transactionName))
+        // Note that, when Trimming is enabled, we can't use reflection to read route data from the HttpTrigger
+        // attribute. In that case the route name will always be /api/<FUNCTION_NAME>
```
bruno-garcia, line 100: We should add this as a note to the docs under AOT support

test/Sentry.Azure.Functions.Worker.Tests/BindableSentryAspNetCoreOptionsTests.cs
```diff
@@ -0,0 +1,30 @@
+#if NET8_0_OR_GREATER
+using Microsoft.Extensions.Configuration;
+
+namespace Sentry.Azure.Functions.Worker.Tests;
```
bitsandfoxes, line 4: I think this file needs renaming.

src/Sentry.Azure.Functions.Worker/Sentry.Azure.Functions.Worker.csproj
jamescrosswell: We only set it for NuGet packages that we intend to support AOT. At the moment that's:

- [ ] Sentry.csproj
- [ ] Sentry.AspNetCore.csproj
- [ ] Sentry.Azure.Functions.Worker.csproj
- [ ] Sentry.Extensions.Logging.csproj
- [ ] Sentry.Maui.csproj
- [ ] Sentry.NLog.csproj
- [ ] Sentry.OpenTelemetry.csproj
- [ ] Sentry.Profiling.csproj
- [ ] Sentry.Serilog.csproj

Still missing are:

- [ ] Sentry.Android.AssemblyReader.csproj
- [ ] Sentry.AspNet.csproj (N/A Legacy)
- [ ] Sentry.AspNetCore.Grpc.csproj (TODO?)
- [ ] Sentry.Bindings.Android.csproj (???)
- [ ] Sentry.Bindings.Cocoa.csproj (???)
- [ ] Sentry.DiagnosticSource.csproj (N/A Legacy?)
- [ ] Sentry.EntityFramework.csproj (N/A Legacy)
- [ ] Sentry.Google.Cloud.Functions.csproj (TODO?)
- [ ] Sentry.Log4Net.csproj (N/A Legacy)

I think this is related to https://github.com/getsentry/sentry-dotnet/issues/2890. Once everything that targets net8.0 is AotCompatible, we could put this in the `Directory.Build.props` but I don't think we can do that yet (and not in this PR).

src/Sentry.Azure.Functions.Worker/SentryFunctionsWorkerMiddleware.cs
jamescrosswell: Hm, we could look into that... I'll create a ticket from your comment.

test/Sentry.Azure.Functions.Worker.Tests/BindableSentryAspNetCoreOptionsTests.cs
jamescrosswell: Good catch 😉

bruno-garcia (approved):

src/Sentry.Azure.Functions.Worker/Sentry.Azure.Functions.Worker.csproj
bruno-garcia: Also we can add the `false` explicitly where needed

Issue #2884, opened 2023-11-22
Simplify how users configure Sentry with Serilog


jamescrosswell:
Currently it's extremely confusing to configure Sentry to work with Serilog. See [this StackOverflow question](https://stackoverflow.com/questions/77454233/sentry-with-generic-host-and-serilog-net-7/77463866) to get some flavour, and also this issue/PR: 
- https://github.com/getsentry/sentry-dotnet/pull/2883 

Currently you can provide a Sentry DSN and various other settings via:
1. The `SentrySdk.Init` method in an application that's not using a Generic Host
2. The `SentryWebHostBuilderExtensions.UseSentry` extension method, when configuring the generic host
3. And/Or the `SentrySinkExtensions.Sentry` extension method when configuring Serilog

This is maybe an issue with the default value of the `initializeSdk` parameter that gets passed when initializing via Serilog. 

At the moment, this defaults to `null`:
https://github.com/getsentry/sentry-dotnet/blob/cf2d206612c5096efaf3f8e853ec4cfcf1fdbab0/src/Sentry.Serilog/SentrySinkExtensions.cs#L107

And that means whatever was in the `SentrySerilogOptions` before won't be overridden:
https://github.com/getsentry/sentry-dotnet/blob/cf2d206612c5096efaf3f8e853ec4cfcf1fdbab0/src/Sentry.Serilog/SentrySinkExtensions.cs#L302-L306

And with a shinny new set of `SentrySerilogOptions` that will be `true`:
https://github.com/getsentry/sentry-dotnet/blob/2285a7f3b805136ab858d4f61e456f30b8a1e2c4/src/Sentry.Serilog/SentrySerilogOptions.cs#L7-L12

So by default, if you initialize Sentry using `SentrySdk.Init` or `UsingSentry` and then you wire up Serilog to use Sentry via the `SentrySinkExtensions.Sentry` extension method we provide, without supplying any parameters to that method, you get an exception at runtime 😢  

## Brainstorming

One possible solution would be to detect, in the call to `SentrySinkExtensions.Sentry`, whether a call to initialize Sentry had already been made or not and, if so, use a default value of `false` for `initializeSdk`. 

Ordinarily, we could check SentrySdk.IsEnabled to work this out... but it turns out when using `SentryWebHostBuilderExtensions.UseSentry` initialization of the SDK is delayed:
[...]

PR #2928, opened 2023-11-30
Added non-initialising overload to SerilogSinkExtensions

jamescrosswell:
Resolves https://github.com/getsentry/sentry-dotnet/issues/2884

# Problem statement

Previously configuring Sentry to work with Serilog was extremely error prone. If you initialized Sentry using `SentrySdk.Init` or `UsingSentry` and then you wire up Serilog to use Sentry via the `SentrySinkExtensions.Sentry` extension method we provide, without supplying any parameters to that method, you get an exception at runtime because it tries to initialize the SDK without a DSN 😢  

A simple example is:

```csharp
var builder = WebApplication.CreateBuilder(args);

// This will eventually initialize Sentry... but it gets delayed until after the MEL backend has been replaced
builder.WebHost.UseSentry(o =>
{
    o.Dsn = "https://b887218a80114d26a9b1a51c5f88e0b4@o447951.ingest.sentry.io/6601807";
    o.Debug = true;
});
builder.Host.UseSerilog(new LoggerConfiguration()
    .WriteTo.Sentry() // <-- Still fails trying to initialize Sentry with no DSN... 
    .CreateLogger());

var app = builder.Build();
```

Whilst it looks like that should work, the initialization code in the call to `WriteTo.Sentry()` would be executed before the initialization code in the call to `UseSentry` (where a DSN is actually provided) and the user would get an error that was not obvious to resolve.  

# Solution

We have changed one of the `WriteTo.Sentry` overrides and added another. 

```csharp
    public static LoggerConfiguration Sentry(
        this LoggerSinkConfiguration loggerConfiguration,
        string dsn, // <-- Make the DSN required... use this override if you want to initialize Sentry
        ...
        );

    // Implicitly this one does not initialize the SDK so only takes the logging specific parameters
    public static LoggerConfiguration Sentry(
        this LoggerSinkConfiguration loggerConfiguration,
        LogEventLevel? minimumEventLevel = null,
        LogEventLevel? minimumBreadcrumbLevel = null,
        IFormatProvider? formatProvider = null,
        ITextFormatter? textFormatter = null
        );
```

The first override has been changed such that the `dsn` parameter is no longer optional. This is the override that should be used when users do want to initialize the Sentry SDK at the same time as setting up a Sentry Serilog Sink. 

The second override doesn't take any `dsn` parameter... indeed it only accepts the limited set of arguments concerned with configuring the Sentry Serilog Sink itself. This is the variant that will be used if no arguments are supplied. Thus the sample code above will work as expected.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@
 
 ### API breaking Changes
 
+- The method used to configure a Sentry Sink for Serilog now has an additional overload. Calling `WriteTo.Sentry()` with no arguments will no longer attempt to initialize the SDK (it has optional arguments to configure the behavrious of the Sink only). If you want to initialize Sentry at the same time you configure the Sentry Sink then you will need to use the overload of this method that accepts a DSN as the first parameter (e.g. `WriteTo.Sentry("https://d4d82fc1c2c4032a83f3a29aa3a3aff@fake-sentry.io:65535/2147483647")`). ([#2928](https://github.com/getsentry/sentry-dotnet/pull/2928))
+
 #### Removed APIs
 
 - SentrySinkExtensions.ConfigureSentrySerilogOptions is now internal. If you were using this method, please use one of the `SentrySinkExtensions.Sentry` extension methods instead. ([#2902](https://github.com/getsentry/sentry-dotnet/pull/2902)) 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@
 
 ### API breaking Changes
 
-- The method used to configure a Sentry Sink for Serilog now has an additional overload. Calling `WriteTo.Sentry()` with no arguments will no longer attempt to initialize the SDK (it has optional arguments to configure the behavrious of the Sink only). If you want to initialize Sentry at the same time you configure the Sentry Sink then you will need to use the overload of this method that accepts a DSN as the first parameter (e.g. `WriteTo.Sentry("https://d4d82fc1c2c4032a83f3a29aa3a3aff@fake-sentry.io:65535/2147483647")`). ([#2928](https://github.com/getsentry/sentry-dotnet/pull/2928))
+- The method used to configure a Sentry Sink for Serilog now has an additional overload. Calling `WriteTo.Sentry()` with no arguments will no longer attempt to initialize the SDK (it has optional arguments to configure the behaviour of the Sink only). If you want to initialize Sentry at the same time you configure the Sentry Sink then you will need to use the overload of this method that accepts a DSN as the first parameter (e.g. `WriteTo.Sentry("https://d4d82fc1c2c4032a83f3a29aa3a3aff@fake-sentry.io:65535/2147483647")`). ([#2928](https://github.com/getsentry/sentry-dotnet/pull/2928))
 
 #### Removed APIs
 
--- a/src/Sentry.Serilog/SentrySinkExtensions.cs
+++ b/src/Sentry.Serilog/SentrySinkExtensions.cs
@@ -132,7 +132,11 @@ public static class SentrySinkExtensions
     }
 
     /// <summary>
-    /// Adds a Sentry Sink for Serilog.
+    /// <para>Adds a Sentry Sink for Serilog.</para>
+    /// <remarks>
+    /// Note this overload doesn't initialize Sentry for you, so you'll need to have already done so. Alternatively you
+    /// can use use the overload of this extension method, passing a DSN string in the first argument.
+    /// </remarks>
     /// </summary>
     /// <param name="loggerConfiguration">The logger configuration .<seealso cref="LoggerSinkConfiguration"/></param>
     /// <param name="minimumEventLevel">Minimum log level to send an event. <seealso cref="SentrySerilogOptions.MinimumEventLevel"/></param>
@@ -163,10 +167,10 @@ public static class SentrySinkExtensions
     /// </example>
     public static LoggerConfiguration Sentry(
         this LoggerSinkConfiguration loggerConfiguration,
-        LogEventLevel? minimumEventLevel,
-        LogEventLevel? minimumBreadcrumbLevel,
-        IFormatProvider? formatProvider,
-        ITextFormatter? textFormatter
+        LogEventLevel? minimumEventLevel = null,
+        LogEventLevel? minimumBreadcrumbLevel = null,
+        IFormatProvider? formatProvider = null,
+        ITextFormatter? textFormatter = null
         )
     {
         return loggerConfiguration.Sentry(o => ConfigureSentrySerilogOptions(o,
--- a/src/Sentry.Serilog/SentrySinkExtensions.cs
+++ b/src/Sentry.Serilog/SentrySinkExtensions.cs
@@ -207,40 +207,118 @@ public static class SentrySinkExtensions
         DeduplicateMode? deduplicateMode = null,
         Dictionary<string, string>? defaultTags = null)
     {
-        sentrySerilogOptions.Dsn = dsn ?? sentrySerilogOptions.Dsn;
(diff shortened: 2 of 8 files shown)


CHANGELOG.md
```diff
@@ -8,6 +8,8 @@
 
 ### API breaking Changes
 
+- The method used to configure a Sentry Sink for Serilog now has an additional overload. Calling `WriteTo.Sentry()` with no arguments will no longer attempt to initialize the SDK (it has optional arguments to configure the behavrious of the Sink only). If you want to initialize Sentry at the same time you configure the Sentry Sink then you will need to use the overload of this method that accepts a DSN as the first parameter (e.g. `WriteTo.Sentry("https://d4d82fc1c2c4032a83f3a29aa3a3aff@fake-sentry.io:65535/2147483647")`). ([#2928](https://github.com/getsentry/sentry-dotnet/pull/2928))
```
bitsandfoxes, line 11:
```suggestion
- The method used to configure a Sentry Sink for Serilog now has an additional overload. Calling `WriteTo.Sentry()` with no arguments will no longer attempt to initialize the SDK (it has optional arguments to configure the behavior of the Sink only). If you want to initialize Sentry at the same time you configure the Sentry Sink then you will need to use the overload of this method that accepts a DSN as the first parameter (e.g. `WriteTo.Sentry("https://d4d82fc1c2c4032a83f3a29aa3a3aff@fake-sentry.io:65535/2147483647")`). ([#2928](https://github.com/getsentry/sentry-dotnet/pull/2928))
```

src/Sentry.Serilog/SentrySinkExtensions.cs
```diff
@@ -9,10 +9,10 @@ namespace Serilog;
 public static class SentrySinkExtensions
 {
     /// <summary>
-    /// Add Sentry Serilog Sink.
+    /// Initialize Sentry and add the SentrySink for Serilog.
     /// </summary>
     /// <param name="loggerConfiguration">The logger configuration .<seealso cref="LoggerSinkConfiguration"/></param>
-    /// <param name="dsn">The Sentry DSN. <seealso cref="SentryOptions.Dsn"/></param>
+    /// <param name="dsn">The Sentry DSN (required) <seealso cref="SentryOptions.Dsn"/></param>
```
bitsandfoxes, line 15:
```suggestion
    /// <param name="dsn">The Sentry DSN (required). <seealso cref="SentryOptions.Dsn"/></param>
```

src/Sentry.Serilog/SentrySinkExtensions.cs
```diff
@@ -83,9 +81,9 @@ public static class SentrySinkExtensions
     /// </example>
     public static LoggerConfiguration Sentry(
         this LoggerSinkConfiguration loggerConfiguration,
-        string? dsn = null,
-        LogEventLevel minimumBreadcrumbLevel = LogEventLevel.Information,
-        LogEventLevel minimumEventLevel = LogEventLevel.Error,
+        string dsn,
+        LogEventLevel? minimumBreadcrumbLevel = null,
+        LogEventLevel? minimumEventLevel = null,
```
bitsandfoxes, line 86: Why are we doing away with the default values?

src/Sentry.Serilog/SentrySinkExtensions.cs
```diff
@@ -159,132 +205,45 @@ public static class SentrySinkExtensions
...
-        if (reportAssembliesMode.HasValue)
-        {
-            sentrySerilogOptions.ReportAssembliesMode = reportAssembliesMode.Value;
-        }
-
-        if (deduplicateMode.HasValue)
-        {
-            sentrySerilogOptions.DeduplicateMode = deduplicateMode.Value;
-        }
+        sentrySerilogOptions.Dsn = dsn ?? sentrySerilogOptions.Dsn;
```
bitsandfoxes, line 210: Is this just to be "in line" and look more concise or am I missing something with overwriting?

src/Sentry.Serilog/SentrySinkExtensions.cs
jamescrosswell: They have no impact... They're the same as the default values for those properties when you create the new SentrySerilogOptions.

src/Sentry.Serilog/SentrySinkExtensions.cs
bruno-garcia: It's trying to "only set `dsn` as a value if isn't `null` otherwise keep `option.Dsn` as-is.
So looks like the original code, just refactored.

I personally don't prefer this, before it was pretty clear (if not null, set). Now it's set! Either this value or the other. Properties could have side effects, etc. Just doing some work so it can be 1 line instead of 3 per property`

src/Sentry.Serilog/SentrySinkExtensions.cs
```diff
@@ -159,132 +205,45 @@ public static class SentrySinkExtensions
...
+        sentrySerilogOptions.MinimumEventLevel = minimumEventLevel ?? sentrySerilogOptions.MinimumEventLevel;
+        sentrySerilogOptions.MinimumBreadcrumbLevel = minimumBreadcrumbLevel ?? sentrySerilogOptions.MinimumBreadcrumbLevel;
+        sentrySerilogOptions.FormatProvider = formatProvider ?? sentrySerilogOptions.FormatProvider;
+        sentrySerilogOptions.TextFormatter = textFormatter ?? sentrySerilogOptions.TextFormatter;
+        sentrySerilogOptions.SendDefaultPii = sendDefaultPii ?? sentrySerilogOptions.SendDefaultPii;
+        sentrySerilogOptions.IsEnvironmentUser = isEnvironmentUser ?? sentrySerilogOptions.IsEnvironmentUser;
+        sentrySerilogOptions.ServerName = NonNullOrWhiteSpaceOrDefault(serverName, sentrySerilogOptions.ServerName);
```
bruno-garcia, line 217: Then need to introduce helper function in an attempt to reduce the amount of code here, when it was just 3 lines (if, then, assign)

src/Sentry.Serilog/SentrySinkExtensions.cs
```diff
@@ -159,132 +205,45 @@ public static class SentrySinkExtensions
...
+        sentrySerilogOptions.InitializeSdk = dsn is not null;
+        if (defaultTags?.Count > 0)
         {
             foreach (var tag in defaultTags)
             {
                 sentrySerilogOptions.DefaultTags.Add(tag.Key, tag.Value);
             }
         }
+
+        string? NonNullOrWhiteSpaceOrDefault(string? value, string? defaultValue)
```
bruno-garcia, line 243: Looks like it could be static?

```suggestion
        static string? NonNullOrWhiteSpaceOrDefault(string? value, string? defaultValue)
```

bruno-garcia (commented):
There's a ton of refactoring mixed with the change in behavior which makes code review really difficult. Ideally try to keep refactoring to a minimum in PRs that are fixing or adding a feature. PRs done just for refactoring are much simpler to review

src/Sentry.Serilog/SentrySinkExtensions.cs
jamescrosswell: Yeah logically it should do exactly the same thing as the old code (assuming the setters don't have any side effects)... but it does result in an unecessary `sentrySerilogOptions.Dsn = sentrySerilogOptions.Dsn` when `sentrySerilogOptions.Dsn is not null`. 

I can revert this if you guys want. I liked it because it 1/3 the amount of code (visually) and I just generally like stuff concise.

jamescrosswell:
> There's a ton of refactoring mixed with the change in behavior which makes code review really difficult.

I removed almost all the refactoring... the only one I left in there is to remove redundant default values for `minimumBreadcrumbLevel` and `minimumEventLevel`...

PR #2930, opened 2023-11-30
iOS native profiling

vaind:
This uses sentry-cocoa profiler to capture native profiles. It works for both NativeAOT and MonoAOT. The same approach has been previously used in the Flutter SDK

* NativeAOT: https://sentry-sdks.sentry.io/profiling/profile/sentry-dotnet/d4570e7308f24adfadbf41d56455a361/flamegraph/
    image

* MonoAOT: https://sentry-sdks.sentry.io/profiling/profile/sentry-dotnet/232b1e4a2f7f4d0598b8b2cc07dccf9c/flamegraph/
   image


src/Sentry.Profiling/SamplingTransactionProfiler.cs
```diff
@@ -96,11 +96,11 @@ public void Finish()
     }
 
     /// <inheritdoc />
-    public async Task<ProfileInfo> CollectAsync(Transaction transaction)
+    public async object Collect(Transaction transaction)
```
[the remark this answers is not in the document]
vaind: good point, thanks

src/Sentry/Platforms/Cocoa/CocoaProfiler.cs
```diff
@@ -0,0 +1,54 @@
...
+using Sentry.Cocoa.Facades;
+
+namespace Sentry.Cocoa;
+
+internal class CocoaProfiler : ITransactionProfiler
+{
+    private readonly SentryOptions _options;
+    private readonly SentryId _traceId;
+    private readonly CocoaSdk.SentryId _cocoaTraceId;
+    private readonly ulong _starTimeNs;
```
filipnavara, line 14:
```suggestion
    private readonly ulong _startTimeNs;
```

src/Sentry/Platforms/Cocoa/CocoaProfiler.cs
```diff
@@ -0,0 +1,54 @@
...
+    private ulong _endTimeNs;
+    private readonly SentryStopwatch _stopwatch = SentryStopwatch.StartNew();
+
+    public CocoaProfiler(SentryOptions options, ulong starTimeNs, SentryId traceId, CocoaSdk.SentryId cocoaTraceId)
```
filipnavara, line 18:
```suggestion
    public CocoaProfiler(SentryOptions options, ulong startTimeNs, SentryId traceId, CocoaSdk.SentryId cocoaTraceId)
```

src/Sentry/Platforms/Cocoa/CocoaProfiler.cs
```diff
@@ -0,0 +1,54 @@
...
+    {
+        _options = options;
+        _starTimeNs = starTimeNs;
```
filipnavara, line 21:
```suggestion
        _startTimeNs = startTimeNs;
```

src/Sentry/Platforms/Cocoa/CocoaProfiler.cs
```diff
@@ -0,0 +1,54 @@
...
+        _cocoaTraceId = cocoaTraceId;
+        _options.LogDebug("Trace {0} profile start timestamp: {1} ns", _traceId, _starTimeNs);
+    }
+
+    /// <inheritdoc />
+    public void Finish()
+    {
+        if (_endTimeNs == 0)
+        {
+            _endTimeNs = _starTimeNs + (ulong)_stopwatch.ElapsedNanoseconds;
```
filipnavara, line 32:
```suggestion
            _endTimeNs = _startTimeNs + (ulong)_stopwatch.ElapsedNanoseconds;
```

src/Sentry/Platforms/Cocoa/CocoaProfiler.cs
```diff
@@ -0,0 +1,54 @@
...
+            _options.LogDebug("Trace {0} profile end timestamp: {1} ns", _traceId, _endTimeNs);
+        }
+    }
+
+    public ISerializable Collect(Transaction transaction)
+    {
+        // TODO change return type of CocoaSDKs CollectProfileBetween to NSMutableDictionary
+        var payload = SentryCocoaHybridSdk.CollectProfileBetween(_starTimeNs, _endTimeNs, _cocoaTraceId)?.MutableCopy() as NSMutableDictionary;
```
filipnavara, line 40:
```suggestion
        var payload = SentryCocoaHybridSdk.CollectProfileBetween(_startTimeNs, _endTimeNs, _cocoaTraceId)?.MutableCopy() as NSMutableDictionary;
```

src/Sentry/Platforms/Cocoa/CocoaProfiler.cs
vaind: thx

filipnavara (approved):

src/Sentry/SentryOptions.cs
```diff
@@ -746,6 +746,12 @@ public IList<SubstringOrRegexPattern> FailedRequestTargets
         true => TracesSampler is not null || TracesSampleRate is > 0.0 or null
     };
 
+    /// <summary>
+    /// Indicates whether profiling is enabled, via any combination of
+    /// <see cref="EnableTracing"/>, <see cref="TracesSampleRate"/>, or <see cref="TracesSampler"/>.
+    /// </summary>
+    internal bool IsProfilingEnabled => IsPerformanceMonitoringEnabled && ProfilesSampleRate > 0.0;
```
bitsandfoxes, line 753: Might be a weird hill to die on but could we go with `EnableProfiling` for consistency?

src/Sentry/Platforms/Cocoa/CocoaProfiler.cs
```diff
@@ -0,0 +1,54 @@
...
+    private ulong _endTimeNs;
+    private readonly SentryStopwatch _stopwatch = SentryStopwatch.StartNew();
+
+    public CocoaProfiler(SentryOptions options, ulong startTimeNs, SentryId traceId, CocoaSdk.SentryId cocoaTraceId)
+    {
+        _options = options;
+        _startTimeNs = startTimeNs;
+        _traceId = traceId;
+        _cocoaTraceId = cocoaTraceId;
+        _options.LogDebug("Trace {0} profile start timestamp: {1} ns", _traceId, _startTimeNs);
```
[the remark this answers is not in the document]
vaind: I understand, but this is what the native profiler works with so when you're actually "debugging" based on the log output (that's why it's `LogDebug()`), you see what is actually passed. I was going to do a conversion to DateTime but it seems it doesn't support nanosecond precision.

src/Sentry/Platforms/Cocoa/CocoaProfiler.cs
```diff
@@ -0,0 +1,54 @@
...
+    }
+
+    /// <inheritdoc />
+    public void Finish()
+    {
+        if (_endTimeNs == 0)
+        {
+            _endTimeNs = _startTimeNs + (ulong)_stopwatch.ElapsedNanoseconds;
+            _options.LogDebug("Trace {0} profile end timestamp: {1} ns", _traceId, _endTimeNs);
```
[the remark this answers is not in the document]
vaind: same as above

src/Sentry.Profiling/SamplingTransactionProfiler.cs
```diff
@@ -96,11 +96,14 @@ public void Finish()
...
-    public async Task<ProfileInfo> CollectAsync(Transaction transaction)
+    public Protocol.Envelopes.ISerializable Collect(Transaction transaction)
+        => Protocol.Envelopes.AsyncJsonSerializable.CreateFrom(CollectAsync(transaction));
+
+    internal async Task<ProfileInfo> CollectAsync(Transaction transaction)
     {
         if (!_stopped)
         {
-            throw new InvalidOperationException("Profiler.CollectAsync() called before Finish()");
+            throw new InvalidOperationException("Profiler.Collect() called before Finish()");
```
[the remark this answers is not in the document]
vaind: It's caught down the road when serializing. I've added test case to check

CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 ## Unreleased
 
+### Features
+
+- iOS profiling support (alpha). ([#2930](https://github.com/getsentry/sentry-dotnet/pull/2930))
```
bruno-garcia, line 7:
```suggestion
- iOS profiling support (alpha). ([#2930](https://github.com/getsentry/sentry-dotnet/pull/2930))

```

src/Sentry.Profiling/ProfilingIntegration.cs
```diff
@@ -10,6 +10,8 @@ public class ProfilingIntegration : ISdkIntegration
     /// <inheritdoc/>
     public void Register(IHub hub, SentryOptions options)
     {
-        options.TransactionProfilerFactory = SamplingTransactionProfilerFactory.Create(options);
+        if (options.IsProfilingEnabled) {
```
bruno-garcia, line 13: I'll buy 10 beers to the first person who adds something like `dotnet format` that works in this repo :)

```suggestion
        if (options.IsProfilingEnabled)
        {
```

src/Sentry/Internal/Hub.cs
```diff
@@ -154,7 +154,9 @@ public async Task ConfigureScopeAsync(Func<Scope, Task> configureScope)
                 transaction.SampleRate = sampleRate;
             }
 
-            if (transaction.IsSampled is true && _options.TransactionProfilerFactory is { } profilerFactory)
+            if (transaction.IsSampled is true &&
+                _options.TransactionProfilerFactory is { } profilerFactory &&
+                _randomValuesFactory.NextBool(_options.ProfilesSampleRate ?? 0.0))
```
bruno-garcia, line 159: Can we avoid the null coalescing -> random generation if we don't have a sample rate? 
Might even be able to do something like `_options.ProfilesSampleRate is { } sampleRate && sampleRate > 0.0
```suggestion
            if (transaction.IsSampled is true &&
                _options.TransactionProfilerFactory is { } profilerFactory &&
                _options.ProfilesSampleRate is { } sampleRate &&
                sampleRate > 0.0 &&
                _randomValuesFactory.NextBool(sampleRate))
```

src/Sentry/Platforms/Cocoa/CocoaProfiler.cs
```diff
@@ -0,0 +1,62 @@
...
+namespace Sentry.Cocoa;
+
+internal class CocoaProfiler : ITransactionProfiler
+{
+    private readonly SentryOptions _options;
+    private readonly SentryId _traceId;
+    private readonly CocoaSdk.SentryId _cocoaTraceId;
+    private readonly ulong _startTimeNs;
+    private ulong _endTimeNs;
+    private readonly SentryStopwatch _stopwatch = SentryStopwatch.StartNew();
```
bruno-garcia, line 15: just FYI this will be the last thing to run in the contructor (iirc). I'm pretty sure it doesn't matter but if u wanted this to be the first to include time spent in the ctor (logger is the only thing that would matter), add to it explicitlty

src/Sentry/Platforms/Cocoa/CocoaProfiler.cs
```diff
@@ -0,0 +1,62 @@
...
+        if (_endTimeNs == 0)
+        {
+            _endTimeNs = _startTimeNs + (ulong)_stopwatch.ElapsedNanoseconds;
+            _options.LogDebug("Trace {0} profile end timestamp: {1} ns", _traceId, _endTimeNs);
+        }
+    }
+
+    public ISerializable? Collect(Transaction transaction)
+    {
+        // TODO change return type of CocoaSDKs CollectProfileBetween to NSMutableDictionary
```
bruno-garcia, line 38: do we need a ticket there not to forget this?

src/Sentry/Platforms/Cocoa/CocoaProfiler.cs
```diff
@@ -0,0 +1,62 @@
...
+
+        payload["timestamp"] = transaction.StartTimestamp.ToString("o", CultureInfo.InvariantCulture).ToNSString();
+
+        var payloadTx = payload["transaction"]?.MutableCopy() as NSMutableDictionary;
+        if (payloadTx is null)
+        {
+            _options.LogWarning("Trace {0} collected profile payload doesn't have transaction information", _traceId);
+            return null;
+        }
+
```
bruno-garcia, line 55: I can't spot if we need the timestamp earlier. Since we bail in case payload is null, move it down?
```suggestion
        var payloadTx = payload["transaction"]?.MutableCopy() as NSMutableDictionary;
        if (payloadTx is null)
        {
            _options.LogWarning("Trace {0} collected profile payload doesn't have transaction information", _traceId);
            return null;
        }

        payload["timestamp"] = transaction.StartTimestamp.ToString("o", CultureInfo.InvariantCulture).ToNSString();
```

src/Sentry/Platforms/Cocoa/CocoaProfilerFactory.cs
```diff
@@ -0,0 +1,26 @@
...
+    {
+        _options = options;
+    }
+
+    /// <inheritdoc />
+    public ITransactionProfiler? Start(ITransactionTracer tracer, CancellationToken cancellationToken)
+    {
+        var traceId = tracer.TraceId.ToCocoaSentryId();
+        var startTime = SentryCocoaHybridSdk.StartProfilerForTrace(traceId);
+        if (startTime == 0)
```
bruno-garcia, line 20: Is this a normal behavior of the profiler? 
I wonder if this will be hard to diagnose when "profiling isn't working" and it's hitting this return.
Do we have logs coming in from somewhere else already?

src/Sentry/Platforms/Cocoa/Facades/SerializableNSObject.cs
```diff
@@ -0,0 +1,33 @@
...
+{
+    private readonly NSObject _value;
+
+    public SerializableNSObject(NSObject value)
+    {
+        _value = value;
+    }
+
+    public Task SerializeAsync(Stream stream, IDiagnosticLogger? logger, CancellationToken cancellationToken = default)
+        => Task.Run(() => Serialize(stream, logger)); // TODO do we need a better implementation?
```
bruno-garcia, line 16: If the code is synchronous we could call it synchronously and just return `Task.CompletedTask`?
This is called on the background worker class which is already a background task so no need to spawn another one with Task.Run

bruno-garcia (approved):

src/Sentry/Internal/Hub.cs
vaind: I've had it like that but:
1. NextBool short-circuits if the given value is 0.0
2. it actually never is because options.TransactionProfilerFactory is null if profiling is disabled (as checked by the iOS & .net profiler integrations

src/Sentry/Platforms/Cocoa/CocoaProfiler.cs
vaind: thx, I'll change that

src/Sentry/Platforms/Cocoa/CocoaProfiler.cs
vaind: Sure. I mean it's not going to be null unless sentry-cocoa changes, but I don't have a problem with the change.

src/Sentry/Platforms/Cocoa/Facades/SerializableNSObject.cs
vaind: thx, updated

src/Sentry/Platforms/Cocoa/CocoaProfilerFactory.cs
vaind: It should never be zero, I'll just remove the check, it's pointless

src/Sentry/Platforms/Cocoa/Facades/SerializableNSObject.cs
```diff
@@ -0,0 +1,33 @@
...
+        var obj = _value is CocoaSdk.ISentrySerializable serializable
+            ? serializable.Serialize()
+            : _value;
+
+        // Now we will use Apple's JSON Serialization functions.
+        // See https://developer.apple.com/documentation/foundation/nsjsonserialization
+        // TODO can we pipe NSOutputStream directly? It can be passed as a second argument
+        // TODO how do we check if the error happened? Is it non-null? Then we can rethrow as NSErrorException?
+        var data = NSJsonSerialization.Serialize(obj, 0, out NSError error);
+        using var dataStream = data.AsStream();
```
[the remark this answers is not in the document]
bruno-garcia: Sounds like we can't use async then

src/Sentry/SentryOptions.cs
bitsandfoxes: That'd be even better.

Issue #2935, opened 2023-12-02
Move LogCat prop to SentryOptions


bruno-garcia:
https://github.com/getsentry/sentry-dotnet/pull/2926/files#diff-3db3897634fba1f00490c5adec65ac48a341016f4a0b4c0675115ef7094ff39cR14-R15

Ended up on `option.Android.LogCat..`
Lets move it to `options.LogCat..`

`options.Android` exists for exposing the binding options from the native Android SDK (see at the end of https://github.com/getsentry/sentry-dotnet/issues/2934 about naming)

PR #2937, opened 2023-12-02
Move LogCat prop to SentryOptions

kanadaj:
Addresses https://github.com/getsentry/sentry-dotnet/issues/2935 (Move LogCat prop to SentryOptions)

### DIFF
--- a/src/Sentry/Internal/LogCatAttachmentEventProcessor.cs
+++ b/src/Sentry/Internal/LogCatAttachmentEventProcessor.cs
@@ -1,3 +1,4 @@
+#if ANDROID
 using Sentry.Extensibility;
 using Runtime = Java.Lang.Runtime;
 
@@ -103,3 +104,5 @@ internal class LogCatAttachmentEventProcessor : ISentryEventProcessorWithHint
         return Process(@event, new Hint());
     }
 }
+
+#endif
--- a/src/Sentry/Internal/LogCatAttachmentEventProcessor.cs
+++ b/src/Sentry/Internal/LogCatAttachmentEventProcessor.cs
@@ -1,7 +1,7 @@
 using Sentry.Extensibility;
 using Runtime = Java.Lang.Runtime;
 
-namespace Sentry.Android;
+namespace Sentry.Internal;
 
 internal class LogCatAttachmentEventProcessor : ISentryEventProcessorWithHint
 {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,13 @@
 # Changelog
 
+## Unreleased
+
+### Fixes
+
+### Features
+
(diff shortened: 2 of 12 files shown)


src/Sentry/Internal/LogCatAttachmentEventProcessor.cs
```diff
@@ -1,7 +1,8 @@
+#if ANDROID
```
bruno-garcia, line 1: This could stay in Sentry/Platform/Android, no need to move it up.

bruno-garcia (commented):
There's a partial class defined in Platform/Android for SentryOptions so the properties can stay there (just not under the native binding `options.Android` that map to the native SDK options:

[main/src/Sentry/Platforms/Android/SentryOptions.cs](https://github.com/getsentry/sentry-dotnet/blob/main/src/Sentry/Platforms/Android/SentryOptions.cs)

All Android specific code that can go there is better located under `Platform/Android` (similar to a MAUI app structure).
We can avoid #ifdef at the top level stuff a lot this way.

bruno-garcia (approved):
LGTM but looking at the sample made me realize that having it under `Android` had the advantage of not requiring additional #ifdef from the client app. Since the whole property is by definition Android specific.
The fact it breaks the simple explanation that's "binding the native Android SDK" is the challenge. 

So I'm happy making this change, but at the same feel less strongly we must, because I see the trade off of the original approach vs this.

cc @bitsandfoxes @jamescrosswell @vaind for thoughts

kanadaj:
It still requires #ifdef on MAUI so long as the property is only available on Android. At the same time, of we don't do that, it will get confusing that the property is there on platforms without logcat

bruno-garcia:
We are renaming `Android` to `Native` which frees up adding a new property `Android` (lol?) that would only only Android specific things.

As I mentioned before I don't know if this is the best way forward, but wanted to share these thoughts. @vaind wdyt?

jamescrosswell:
> My learning from seeing this PR is that the usage of the Android-only option is a bit awkward if sitting directly on `option` when compared to having it under `options.Android.Abc`

I agree. I wouldn't move it, tbh. 

If we do want to make it always available (so the client code doesn't need `#if` directives) I think it's inconsistent with how we surface all the other platform specific options. So either we we do or we don't, but let's not do it for some options but not others.

bruno-garcia:
> > My learning from seeing this PR is that the usage of the Android-only option is a bit awkward if sitting directly on `option` when compared to having it under `options.Android.Abc`
> 
> I agree. I wouldn't move it, tbh.
> 
> If we do want to make it always available (so the client code doesn't need `#if` directives) I think it's inconsistent with how we surface all the other platform specific options. So either we we do or we don't, but let's not do it for some options but not others.

Now that the binding option became `option.Native`, we can create a `SentryAndroidOptions` class to add Android-only properties, and keep the original location.

Basically pulling `main` into this PR will force you into that set up since the `Android` to `Native` renaming was merged already

PR #2941, opened 2023-12-04
Refactor ConfigureSentrySerilogOptions

jamescrosswell:
#skip-changelog

Potential refactor of `ConfigureSentrySerilogOptions` for discussion. 

I pulled this out of [simplify_serilog_2884](https://github.com/getsentry/sentry-dotnet/tree/simplify_serilog_2884) to keep the refactoring separate from the change. Not sure if it's worth it. It does make the code way more compact. (takes ~150 lines and makes it about ~50 lines)... on the flip side, it's a little bit more complex. 🤷🏻

### DIFF
--- a/src/Sentry.Serilog/SentrySinkExtensions.cs
+++ b/src/Sentry.Serilog/SentrySinkExtensions.cs
@@ -132,7 +132,11 @@ public static class SentrySinkExtensions
     }
 
     /// <summary>
-    /// Adds a Sentry Sink for Serilog.
+    /// <para>Adds a Sentry Sink for Serilog.</para>
+    /// <remarks>
+    /// Note this overload doesn't initialize Sentry for you, so you'll need to have already done so. Alternatively you
+    /// can use use the overload of this extension method, passing a DSN string in the first argument.
+    /// </remarks>
     /// </summary>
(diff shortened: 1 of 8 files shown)


src/Sentry.Serilog/SentrySinkExtensions.cs
```diff
@@ -159,130 +205,54 @@ public static class SentrySinkExtensions
...
-        if (maxQueueItems.HasValue)
-        {
-            sentrySerilogOptions.MaxQueueItems = maxQueueItems.Value;
-        }
-
-        if (shutdownTimeout.HasValue)
-        {
-            sentrySerilogOptions.ShutdownTimeout = shutdownTimeout.Value;
-        }
+        SetIfNotNull(dsn, value => sentrySerilogOptions.Dsn = value);
```
bruno-garcia, line 210: I don't necessarily find this easier to maintain. The original code is easy/simple enough to read, now there's a function call and a lambda and `!`'s everywhere which raise alarm (even though the SetIfNotNull validates it should be safe)

jamescrosswell: All good, let's close this one then.

Issue #2880, opened 2023-11-22
Add Metrics API


bruno-garcia:
Sentry is building a new metrics product: https://github.com/getsentry/sentry/discussions/58584

**_Note:_** We need to be clear that the API itself would be preview, perhaps through xml docs.

## Developer Docs
https://develop.sentry.dev/sdk/metrics/

## Progress in other SDKs
- **Python**:  
    - trailblazer to dogfood/build the product...
    - [Initial implementation](https://github.com/getsentry/sentry-python/pull/2385/files)
    - [Metric Span Summaries](https://github.com/getsentry/sentry-python/pull/2522)
- **PHP**:  https://github.com/getsentry/sentry-php/pull/1619 (draft)
- **Javascript**: https://github.com/getsentry/sentry-javascript/issues/9691 (WIP)
    - https://github.com/getsentry/sentry-javascript/pull/9304
    - https://github.com/getsentry/sentry-javascript/pull/9698
    - Browser: https://github.com/getsentry/sentry-javascript/pull/9794
- **Rust**: https://github.com/getsentry/sentry-rust/pull/618 

## Testing / Dogfooding in .NET

Would be great to dogfood this through [Symbol Collector](https://github.com/getsentry/symbol-collector) which today uses StatsD and emits metrics

The mobile app also has some metrics but it just dumps the `Extra` on events (can see it anywhere other than individual events).

bruno-garcia:
There are actually develop docs on this: https://develop.sentry.dev/delightful-developer-metrics/sending-metrics-sdk/

Please note this is very early stage so not worth adding a ton of tests, we could have this on a branch and ship `alpha`'s from it for example

bruno-garcia:
see also https://github.com/getsentry/sentry-python/pull/2522

bruno-garcia:
JS: https://github.com/getsentry/sentry-javascript/issues/9691

bruno-garcia:
https://github.com/getsentry/sentry-rust/pull/618

jamescrosswell:
## Open Telemetry

Eventaully we might be able to build a bridge between Sentry's metrics and [OpenTelemetry Metrics](https://opentelemetry.io/docs/specs/otel/metrics/api/), if people wanted to use OpenTelemetry for instrumentation (similar to what we've done for Tracing)

### Metric Types

For the following [metric types](https://develop.sentry.dev/delightful-developer-metrics/#metric-types):
[...]

Issue #1006, opened 2021-05-21
Consider concurrency testing with `coyote`


bruno-garcia:
Explore https://github.com/microsoft/coyote

bruno-garcia:
Experimented with it in here: https://github.com/getsentry/sentry-dotnet/pull/1019

It's blocked due to: https://github.com/microsoft/coyote/issues/163

bruno-garcia:
Came up on: https://github.com/getsentry/sentry-dotnet/pull/2949#pullrequestreview-1777350589

> Experimented with it in here: https://github.com/getsentry/sentry-dotnet/pull/1019

> It's blocked due to: https://github.com/microsoft/coyote/issues/163

This was resolved in version [v1.6.0](https://www.nuget.org/packages/Microsoft.Coyote/1.6.0)

PR #2949, opened 2023-12-06
Metrics Support - alpha

jamescrosswell:
Resolves https://github.com/getsentry/sentry-dotnet/issues/2880

image

# Dogfooding

Once we have a nuget package, we can use the [Symbol Collector](https://github.com/getsentry/symbol-collector/blob/16facc91ffddce74918f452be19423b3b4fd105f/src/SymbolCollector.Server/MetricsPublisher.cs#L7) to dogfood this.


src/Sentry/DelegatingMetricAggregator.cs
```diff
@@ -0,0 +1,16 @@
+namespace Sentry;
+
+internal class DelegatingMetricAggregator(IMetricAggregator innerAggregator) : IMetricAggregator
+{
+    public void Increment(string key, double value = 1, MeasurementUnit? unit = null, IDictionary<string, string>? tags = null,
```
mitsuhiko, line 5: Should probably call this `Incr` for consistency with the other SDKs.

src/Sentry/DisabledMetricAggregator.cs
```diff
@@ -0,0 +1,28 @@
+namespace Sentry;
+
+internal class DisabledMetricAggregator : IMetricAggregator
+{
+    public void Increment(string key, double value = 1, MeasurementUnit? unit = null, IDictionary<string, string>? tags = null,
```
mitsuhiko, line 5: Replying here for no good reason, but can `MeasurementUnit` hold an arbitrary non defined unit? What is this, is it an enum?

src/Sentry/DelegatingMetricAggregator.cs
bruno-garcia: .NET folks spell things out and since no one writes C# without an IDE even if we documented incorrectly Incr, they'd find Increment

I'd say it's fine to deviate here but don't feel strongly on this

bruno-garcia: > .NET folks spell things out 

In other words: like like Objective-C where it defaults to complete names instead of abbreviation. That's why everything is so verbose :)

jamescrosswell: This one seems like the odd one out in the original Python implementation as well... all the other metrics just use the metric type name as the method that is used to emit the metric (e.g. `Set`, `Distribution`, `Gauge`). For some reason `Incr` is used instead of `Counter` though. Would it makes sense to rename `Incr` to `Counter`? 

If this is a nouns and verbs thing, it's also not consistent at the moment. Something like `AddCounter`, `AddSet`, `AddGauge` etc. might work for consistent verbs names.

src/Sentry/MetricAggregator.cs
```diff
@@ -0,0 +1,273 @@
...
+        try
+        {
+            while (!shutdownTimeout.IsCancellationRequested)
+            {
+                // If the cancellation was signaled, run until the end of the queue or shutdownTimeout
+                if (!shutdownRequested)
+                {
+                    try
+                    {
+                        var delay = Task.Delay(_flushInterval, shutdownTimeout.Token).ConfigureAwait(false);
```
Giorgi, line 154: Is `delay` used for anything?

src/Sentry/MetricAggregator.cs
```diff
@@ -0,0 +1,273 @@
...
+
+    /// <summary>
+    /// Returns the keys for any buckets that are ready to be flushed (i.e. are for periods before the cutoff)
+    /// </summary>
+    /// <returns>
+    /// An enumerable containing the keys for any buckets that are ready to be flushed
+    /// </returns>
+    internal IEnumerable<long> GetFlushableBuckets()
+    {
+        var cutoff = MetricBucketHelper.GetCutoff();
```
bruno-garcia, line 221: Reasoning below

```suggestion
        if (_buckets is null)
        {
            return Enumerable.Empty<long>();
        }
        var cutoff = MetricBucketHelper.GetCutoff();
```

src/Sentry/MetricAggregator.cs
```diff
@@ -0,0 +1,273 @@
...
+                }
+
+                // Work with the envelope while it's in the queue
+                foreach (var key in GetFlushableBuckets())
+                {
+                    // TODO: Check if a shutdown request has been made
+                    try
+                    {
+                        _options.LogDebug("Flushing metrics for bucket {0}", key);
+                        var bucket = Buckets[key];
```
bruno-garcia, line 186: You have a race condition here that can cause data loss. Happens between this operation and line 201 where you remove it, other threads can write new metrics on this bucket.

You should replace this indexer access with `TryRemove` you used below, and use the reference you extracted out (now discarded with `_`) of the map to do the work here. This also means the `finally` block is redundant

src/Sentry/MetricAggregator.cs
```diff
@@ -0,0 +1,273 @@
...
+                        _captureMetrics(bucket.Values);
+                    }
+                    catch (OperationCanceledException)
+                    {
+                        _options.LogInfo("Shutdown token triggered. Time to exit.");
+                        return;
+                    }
+                    catch (Exception exception)
+                    {
+                        _options.LogError(exception, "Error while processing metric aggregates.");
```
bruno-garcia, line 196: When can we get here? I wonder if this could result in something throwing on each iteration and the code not stopping

src/Sentry/MetricAggregator.cs
```diff
@@ -0,0 +1,273 @@
...
+            _options.LogDebug("Already disposed MetricAggregator.");
+            return;
+        }
+
+        _disposed = true;
+
+        try
+        {
+            // Request the LoopTask stop.
+            _shutdownSource.Cancel();
```
bruno-garcia, line 251: I don't see this being used anywhere. Is the top CancellationTokenSource supposed to be created out of composing with this one? (and if so, should we consider the options.ShutdownTimeout here too when calling cancel?

src/Sentry/MetricBucketHelper.cs
```diff
@@ -0,0 +1,25 @@
+namespace Sentry;
+
+internal static class MetricBucketHelper
+{
+    private const int RollupInSeconds = 10;
+
+    private static readonly DateTime EpochStart = new(1970, 1, 1, 0, 0, 0, 0, DateTimeKind.Utc);
```
bruno-garcia, line 7: There's a `DateTime.UnixEpoch`, can we use that?

https://learn.microsoft.com/en-us/dotnet/api/system.datetime.unixepoch?view=net-8.0

src/Sentry/Protocol/Envelopes/Envelope.cs
```diff
@@ -320,6 +321,18 @@ public static Envelope FromTransaction(Transaction transaction)
         return new Envelope(eventId, header, items);
     }
 
+    /// <summary>
+    /// Creates an envelope that contains one or more Metrics
+    /// </summary>
+    internal static Envelope FromMetrics(IEnumerable<Metric> metrics)
```
bruno-garcia, line 327: if all we do with this is `ToArray` maybe we should take an array here instead of IEnumerable?

src/Sentry/Protocol/Envelopes/Envelope.cs
```diff
@@ -320,6 +321,18 @@ public static Envelope FromTransaction(Transaction transaction)
...
+    {
+        var header = DefaultHeader;
+
+        var items = metrics.Select(EnvelopeItem.FromMetric).ToArray();
```
bruno-garcia, line 331: bonus point to remove this LINQ method call with a loop to avoid an allocations 😅

src/Sentry/Protocol/Metrics/Metric.cs
```diff
@@ -0,0 +1,126 @@
...
+        writer.WriteStartObject();
+        writer.WriteString("type", GetType().Name);
+        writer.WriteSerializable("event_id", EventId, logger);
+        writer.WriteString("name", Key);
+        writer.WriteString("timestamp", Timestamp);
+        if (Unit.HasValue)
+        {
+            writer.WriteStringIfNotWhiteSpace("unit", Unit.ToString());
+        }
+        writer.WriteStringDictionaryIfNotEmpty("tags", Tags!);
```
bruno-garcia, line 46: Looks like Tags is never null. Do we need `!` here?

src/Sentry/Protocol/Metrics/Metric.cs
```diff
@@ -0,0 +1,126 @@
...
+    internal static string SanitizeValue(string input) => Regex.Replace(input, @"[^\w\d_:/@\.\{\}\[\]$-]+", "_");
+
+    public async Task SerializeAsync(Stream stream, IDiagnosticLogger? logger, CancellationToken cancellationToken = default)
+    {
+        /*
+         * We're serializing using the statsd format here: https://github.com/b/statsd_spec
+         */
+        var metricName = SanitizeKey(Key);
+        await Write(metricName).ConfigureAwait(false);
+        await Write("@").ConfigureAwait(false);
```
bruno-garcia, line 63: is there a synchronous API for `Write` instead?

Seems quite wasteful to going async/await machinery for single characters

src/Sentry/Protocol/Metrics/Metric.cs
```diff
@@ -0,0 +1,126 @@
...
+    internal static string SanitizeKey(string input) => Regex.Replace(input, @"[^a-zA-Z0-9_/.-]+", "_");
+    internal static string SanitizeValue(string input) => Regex.Replace(input, @"[^\w\d_:/@\.\{\}\[\]$-]+", "_");
+
+    public async Task SerializeAsync(Stream stream, IDiagnosticLogger? logger, CancellationToken cancellationToken = default)
+    {
+        /*
+         * We're serializing using the statsd format here: https://github.com/b/statsd_spec
+         */
+        var metricName = SanitizeKey(Key);
+        await Write(metricName).ConfigureAwait(false);
```
bruno-garcia, line 62: afaik `.ConfigureAwait(false);` after the first one of these it's already didn't capture ExecutionContext so we could remove all the ones below. We do have some linter that yells at us but in thsi case since there are so many might be easier to disable it for the rest of the method instead of complying to it by adding this everywhere.

src/Sentry/MetricAggregator.cs
```diff
@@ -0,0 +1,273 @@
...
+            MetricType.Set => (string _) => new SetMetric(key, (int)value, unit.Value, tags, timestamp),
+            _ => throw new ArgumentOutOfRangeException(nameof(type), type, "Unknown MetricType")
+        };
+
+        timeBucket.AddOrUpdate(
+            GetMetricBucketKey(type, key, unit.Value, tags),
+            addValuesFactory,
+            (_, metric) =>
+            {
+                metric.Add(value);
```
bruno-garcia, line 132: Might look like we have exclusive access to `metric` here but you're also reading from it to flush on another thread and while you're iterating on things there it might crash if mutate it here.

src/Sentry/MetricAggregator.cs
jamescrosswell: It is not... will remove 👍🏻

src/Sentry/MetricAggregator.cs
```diff
@@ -0,0 +1,273 @@
...
+    /// <summary>
+    /// Returns the keys for any buckets that are ready to be flushed (i.e. are for periods before the cutoff)
+    /// </summary>
+    /// <returns>
+    /// An enumerable containing the keys for any buckets that are ready to be flushed
+    /// </returns>
+    internal IEnumerable<long> GetFlushableBuckets()
+    {
+        var cutoff = MetricBucketHelper.GetCutoff();
+        foreach (var bucket in Buckets)
```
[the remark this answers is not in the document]
jamescrosswell: If metrics aren't being used (haven't been explicitly enabled in the options), this class never gets instantiated - a `DisabledMetricAggregator` gets created instead:
https://github.com/getsentry/sentry-dotnet/blob/c83e9c91840d0f78b070df426d318b43493f0cca/src/Sentry/SentryClient.cs#L78-L85

src/Sentry/MetricAggregator.cs
jamescrosswell: Ah, I see... so if metrics are enabled but no metrics have yet been emitted. Cool, good idea (I'll fix up the syntax so that it works with the `yield return` and the `Lazy<T>`.

src/Sentry/MetricAggregator.cs
jamescrosswell: I thought about that. Given how the cutoff works, nothing will have been written to the bucket for at least 10-20 seconds by the time the bucket gets flushed... so no race condition.

I figured this way if there was some issue with the `_captureMetrics` call, the bucket wouldn't have been removed so we could maybe try again. Thinking on that now though, we don't do anything to prevent a potential infinite loop in this case... so maybe better off removing it from the Buckets collection and just making a best effort to deliver the metrics to Sentry (we have retry logic etc. in the Transport/Envelope handling logic anyway).

src/Sentry/MetricAggregator.cs
jamescrosswell: The `_captureMetrics` Action (which is the only thing that could trhow) is a dependency that gets passed into the `MetricAggregator`. Typically that would be this:
https://github.com/getsentry/sentry-dotnet/blob/c83e9c91840d0f78b070df426d318b43493f0cca/src/Sentry/SentryClient.cs#L244-L248

The most likely thing to go wrong is probably the [`Metric.SerializeAsync` call](https://github.com/getsentry/sentry-dotnet/blob/c83e9c91840d0f78b070df426d318b43493f0cca/src/Sentry/Protocol/Metrics/Metric.cs#L56) that happens when creating a `statsd` envelope item. 

Making the change on what you thought was a race condition above will prevent an infinite loop... and I think having the try catch here makes sense to make sure we don't crash the user's application. So I think once we always remove a bucket from the `Buckets` collection before trying to send the metrics through, we should be good.

src/Sentry/Protocol/Metrics/Metric.cs
jamescrosswell: If I remove it, we get a compiler error due to the signature of the ExtensionMethod being called:
> Argument of type 'System.Collections.Generic.IDictionary<string,string>' cannot be used for parameter 'dic' of type 'System.Collections.Generic.IEnumerable<System.Collections.Generic.KeyValuePair<string,string?>>?' in 'Sentry.Internal.Extensions.JsonExtensions.WriteStringDictionaryIfNotEmpty' because of differences in the nullability of reference types

I can't think of an elegant way aroud this. I could disable the warning with a pragma and include a comment?

src/Sentry/Protocol/Metrics/Metric.cs
jamescrosswell: There is, but it takes an offset as a parameter... so not easy to mix and match the async and the sync calls to write to the stream. I'll try to consolidate some of the writes so that we're not writing out individual characters asynchronously.

src/Sentry/MetricAggregator.cs
jamescrosswell: Is that possible, given the [cutoff logic](https://github.com/getsentry/sentry-dotnet/blob/19dc74aa20d4d76fe6c94b3f78289b35517fe6af/src/Sentry/MetricAggregator.cs#L226-L225)?

This is a `ConcurrentDictionary` so all of the add/remove operations are atomic. The Flush logic will only retrieve buckets prior to the cuttoff (which means no mutations will have happened for `_flushShift` seconds. In an extreme case the flush shift could be set to a very small time value, in which case maybe at the end of a bucket a metric gets emitted and there's some chance of overlap?

I could create a lock, I guess, when flushing (so `metric.Add` operations would queue up while flushing). I think this is what the Python code does (although there's no ConcurrentDictionary in Python). 

Alternatively, the flushShift could be offset to guarantee all metric emits to have completed (e.g. 50-100ms). 

Thoughts? 🤔

bruno-garcia: `ConcurrentDictionary` guarantees the callback on Add/Update won't run concurrently, but if you remove the reference (as it's done below) any access to the reference after is outside any locks. Now that it's `TryRemove` below (after the last commit) it might be OK, need to check the logic with iterating over the concurrentDic (which is safe), won't cause a situation where we find the key here (and run the update callback) while the iterator is alive with that reference. Does it block the `MoveNext()` call on the iterator?

I'll take another look (it's 10 PM on a Sunday) later on the IDE to see if I can spot anything but might be worth testing this with some code in a tight loop adding metrics and making the flush code run every 100ms or something to see if u can cause a crash. Let it run for a few hours for example.

bruno-garcia (commented):
Quick second pass

src/Sentry/MetricAggregator.cs
```diff
@@ -0,0 +1,278 @@
...
+    {
+        var typePrefix = type switch
+        {
+            MetricType.Counter => "c",
+            MetricType.Gauge => "g",
+            MetricType.Distribution => "d",
+            MetricType.Set => "s",
+            _ => throw new ArgumentOutOfRangeException(nameof(type), type, null)
+        };
+        var serializedTags = tags?.ToUtf8Json() ?? string.Empty;
```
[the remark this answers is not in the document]
jamescrosswell: Yeah, this is just replicating how the buckets work in Python... Within each time bucket, a "metric bucket" is a set of metrics for a unique combination of:
1. Metric type
2. Key provided by the user
3. Measurement unit type
4. Tags

The python code also serializes the tags to turn these into part of the composite key.

Potentially we could use a Hash of the JSON string... that would save space but increase computation. 

Alternatively, we could concatenate all the values/keys, using some kind of delimeter (which would have to be escaped if it was present in any of the values/keys). That might be slightly cheaper both computationally and in terms of storage.

src/Sentry/MetricAggregator.cs
```diff
@@ -0,0 +1,278 @@
...
+    // The key for this dictionary is the Timestamp for the bucket, rounded down to the nearest RollupInSeconds... so it
+    // aggregates all of the metrics data for a particular time period. The Value is a dictionary for the metrics,
+    // each of which has a key that uniquely identifies it within the time period
+    internal ConcurrentDictionary<long, ConcurrentDictionary<string, Metric>> Buckets => _buckets.Value;
+    private readonly Lazy<ConcurrentDictionary<long, ConcurrentDictionary<string, Metric>>> _buckets
+        = new(() => new ConcurrentDictionary<long, ConcurrentDictionary<string, Metric>>());
+
+    private Task LoopTask { get; }
+
+    public MetricAggregator(SentryOptions options, Action<IEnumerable<Metric>>? captureMetrics = null, CancellationTokenSource? shutdownSource = null, bool disableLoopTask = false)
```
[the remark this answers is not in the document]
jamescrosswell: > Reading this signature makes me confused if it's done to make it testable or for another reason?

The `disableLoopTask` parameter was just to make testing easier. 

Taking `Action<IEnumerable<Metric>>? captureMetrics` as a dependency was more specific than taking, for example, an `ISentryClient` or something (we don't need most of the methods on `ISentryClient` - just the one that captures metrics). But yeah, I'll make this non nullable 👍🏻 .

src/Sentry/Protocol/Metrics/Metric.cs
jamescrosswell: Yeah, it's a bit weird. Another way to make the warning go away would be an explicit cast:
image

You can see the first call to `WriteStringDictionary` has no warning... where the second does. It's a lot more verbose than a Null forgiving operator but 🤷🏻

jamescrosswell:
I created a unit test (commented out as I don't think we'll keep this or want it running on CI) to help find any issues:
https://github.com/getsentry/sentry-dotnet/blob/54af5c28aeeb05d341a2ae3b4c902d837f79b7d3/test/Sentry.Tests/MetricAggregatorTests.cs#L168-L193

No exceptions and no buckets left over at the end. That's 10 minutes (not two hours) but good that it didn't fall over immediately I guess. 

Other than letting it run longer, anything else you can think of to trip it up?

src/Sentry/MetricAggregator.cs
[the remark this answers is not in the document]
jamescrosswell: Added comments for this.

src/Sentry/MetricAggregator.cs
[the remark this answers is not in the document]
bruno-garcia: > Potentially we could use a Hash of the JSON string... that would save space but increase computation.

I was hoping we could avoid JSON encoding altogether. A cheaper way to stringify it might be enough for as long there's no chance of collision (`obj.GetHashCode()` is not a good choice for example)

> Alternatively, we could concatenate all the values/keys, using some kind of delimeter (which would have to be escaped if it was present in any of the values/keys). That might be slightly cheaper both computationally and in terms of storage.

Yeah something like this could work. We can look at this later, it's not a blocker here at all, but we need to benchmark this stuff to avoid regressing on performance after we had to situations recently of folks pointing out high CPU/Mem usage with Sentry.

bruno-garcia:
> Other than letting it run longer, anything else you can think of to trip it up?

That's a good validating for now, we can get this out and start testing on Symbol Collector, NuGet Trends etc

bruno-garcia:
> I created a unit test (commented out as I don't think we'll keep this or want it running on CI) to help find any issues:
> 
> https://github.com/getsentry/sentry-dotnet/blob/54af5c28aeeb05d341a2ae3b4c902d837f79b7d3/test/Sentry.Tests/MetricAggregatorTests.cs#L168-L193
> 
> No exceptions and no buckets left over at the end. That's 10 minutes (not two hours) but good that it didn't fall over immediately I guess.
> 
> Other than letting it run longer, anything else you can think of to trip it up?

Oh I just read the code and that's not really trying to cause much concurrency. It's spawning a task and awaiting on it on each for.

You'd want to start some threads and let them all run concurrently (1 per CPU core is enough), and in each of them a tight loop.
So you have all of your cores accessing the same metrics bucket trying to write, while the worker (in a higher frequency) captures.

jamescrosswell:
OK... looks like there is a problem with concurrency (good to know!). I'll get the tests to pass then share corrected code and new tests.

jamescrosswell:
Fixed the concurrency issue... I found a [good explanation of it here](https://www.dotnetcurry.com/dotnet/1360/concurrent-programming-dotnet-core#:~:text=Concurrent%20Collections,when%20used%20incorrectly.) for reference:
> Naively, one might even want to use the following:
> `counters.AddOrUpdate(key, 1, (oldKey, oldValue) => oldValue + 1);`
> However, the update delegate in the above method is executed outside the critical section. Therefore a second thread could still read the same old value as the first thread, before the first one has updated it, effectively overwriting the first thread’s update with its own value and losing one increment. Even concurrent collections are not immune to multithreading issues when used incorrectly.

src/Sentry/MetricAggregator.cs
jamescrosswell: So turns out this was nothing to do with the Flush operation... the `ConcurrentDictionary<K,V>.AddOrUpdate` method is a bit misleading. See [here](https://github.com/getsentry/sentry-dotnet/pull/2949#issuecomment-1851463606).

vaind (commented):
Maybe worth looking into https://github.com/microsoft/coyote for thread-safety testing?

jamescrosswell:
> Maybe worth looking into https://github.com/microsoft/coyote for thread-safety testing?

It definitely looks interesting... although there's a lot of moving parts that would add complexity to tests/build. 

I'm using this as an alternative solution to the problem they outlined in their docs:
https://github.com/getsentry/sentry-dotnet/blob/0f8615870c97063fc3fac6ac4051abc3e13a9c17/test/Sentry.Tests/MetricAggregatorTests.cs#L200-L203

I think that's easier to understand and maintain, unless there's something else coyote can do that we need.

vaind:
> I think that's easier to understand and maintain, unless there's something else coyote can do that we need.

I haven't worked with it myself but this looks very useful (from README.md):

> The awesome thing is that once a bug is found, Coyote gives you a trace through the engine.TestReport API that you can use to reliably reproduce the bug as many times as you want, making debugging and fixing the issue significantly easier.

bruno-garcia:
> Maybe worth looking into [microsoft/coyote](https://github.com/microsoft/coyote) for thread-safety testing?

* https://github.com/getsentry/sentry-dotnet/issues/1006

jamescrosswell:
> > Maybe worth looking into [microsoft/coyote](https://github.com/microsoft/coyote) for thread-safety testing?
> 
> * [Consider concurrency testing with `coyote` #1006](https://github.com/getsentry/sentry-dotnet/issues/1006)

@vaind suggested this as well (see [my reply above](https://github.com/getsentry/sentry-dotnet/pull/2949#issuecomment-1852818584)). It looks interesting but would definitely add complexity to the tests and the build:
> - you first convert the above test to a concurrency unit test using the Coyote TestingEngine API
> - Next, you run the coyote rewrite command from the CLI (typically as a post-build task) to automatically rewrite the IL of your test and production binaries. 
> - You can then run the concurrent unit test from your favorite unit testing framework (such as [xUnit](https://xunit.net/))

If we do do that, I think let's do it as part of a separate commit.

bruno-garcia:
@jamescrosswell renamed the PR because we dropped the `DDM` internal name in favor of "Metrics"

src/Sentry/SentryClient.cs
```diff
@@ -431,11 +465,12 @@ private bool CaptureEnvelope(Envelope envelope)
     /// <summary>
     /// Disposes this client
     /// </summary>
-    /// <inheritdoc />
     public void Dispose()
     {
         _options.LogDebug("Flushing SentryClient.");
 
+        Metrics.FlushAsync().GetAwaiter().GetResult();
```
bruno-garcia, line 472: If this throws (e.g: OperationCancelled) the worker task below won't complete.

I think best way here is:

// start tasks
var metricFlushTask = Metrics.FlushAsync()
var workerFlushTask = Worker.FlushAsync(_options.ShutdownTimeout)
try
Task.WhenAll(metricsFlushTask, wrokerFlushTask);
catch
catch anything here and log.

src/Sentry/Protocol/Metrics/DistributionMetric.cs
```diff
@@ -0,0 +1,35 @@
...
+    }
+
+    public DistributionMetric(string key, double value, MeasurementUnit? unit = null,
+        IDictionary<string, string>? tags = null, DateTime? timestamp = null)
+        : base(key, unit, tags, timestamp)
+    {
+        Value = new List<double>() { value };
+    }
+
+    public IList<double> Value { get; set; }
```
bruno-garcia, line 23: We offer a Add method but we expose `IList` which has Add/Remove etc. Do we really want this as a public property?
Even `set` is public

src/Sentry/Protocol/Metrics/DistributionMetric.cs
```diff
@@ -0,0 +1,35 @@
...
+    public override void Add(double value)
+    {
+        Value.Add(value);
+    }
+
+    protected override void WriteValues(Utf8JsonWriter writer, IDiagnosticLogger? logger) =>
+        writer.WriteArrayIfNotEmpty<double>("value", Value, logger);
+
+    protected override IEnumerable<IConvertible> SerializedStatsdValues()
+        => Value.Select(v => (IConvertible)v);
```
bruno-garcia, line 34: if we'll have a lot of those, a `foreach` here instead avoids an allocation I believe (but we can benchmark this and improve in a follow up PR)

src/Sentry/Protocol/Envelopes/Envelope.cs
```diff
@@ -325,6 +326,35 @@ public static Envelope FromTransaction(Transaction transaction)
...
 
+    /// <summary>
+    /// Creates an envelope that contains one or more <see cref="CodeLocations"/>
+    /// </summary>
+    internal static Envelope FromCodeLocations(CodeLocations codeLocations)
+    {
+        var header = DefaultHeader;
+
+        List<EnvelopeItem> items = new();
+        items.Add(EnvelopeItem.FromCodeLocations(codeLocations));
```
bruno-garcia, line 337: Does `Add` only allocates an array with 1 element? I believe adding the first item u can:

var items = new List<EnvelopeItem>(1) {EnvelopeItem.FromCodeLocations(};

Though it's anyway calling `Add` just a syntactical sugar. But the `(1)` in there already gives you the list in the size we'll use.

src/Sentry/MetricHelper.cs
```diff
@@ -0,0 +1,40 @@
...
+    /// also apply independent jittering.
+    /// </summary>
+    /// <remarks>Internal for testing</remarks>
+    internal static double FlushShift = new Random().Next(0, 1000) * RollupInSeconds;
+    internal static DateTime GetCutoff() => DateTime.UtcNow
+        .Subtract(TimeSpan.FromSeconds(RollupInSeconds))
+        .Subtract(TimeSpan.FromMilliseconds(FlushShift));
+
+    internal static string SanitizeKey(string input) => Regex.Replace(input, @"[^a-zA-Z0-9_/.-]+", "_");
+    internal static string SanitizeValue(string input) => Regex.Replace(input, @"[^\w\d_:/@\.\{\}\[\]$-]+", "_");
```
bruno-garcia, line 39: These will run a lot, how about compiling them?

src/Sentry/MetricAggregator.cs
```diff
@@ -0,0 +1,455 @@
...
+        {
+            _flushLock.Dispose();
+            _shutdownSource.Dispose();
+            LoopTask.Dispose();
+        }
+    }
+
+    public void Dispose()
+    {
+        DisposeAsync().GetAwaiter().GetResult();
```
bruno-garcia, line 453: if the task throws (e.g cancelled) it'll throw. Best handle here (dispose shouldn't throw)

Also, should we just not have a sync dipose then>? Go only async

bruno-garcia (commented):
Ranout of gas, didn't review `MetricAggregator` really and only did around 30/52 files.

bruno-garcia (commented):
Only got as far as 30/52 files and didn't go deep on `MetricAggregator` and other major pieces but sending some bits as it's what I managed to cover

samples/Sentry.Samples.Console.Metrics/Sentry.Samples.Console.Metrics.csproj
```diff
@@ -0,0 +1,17 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+    <PropertyGroup>
+        <OutputType>Exe</OutputType>
+        <TargetFramework>net8.0</TargetFramework>
+    </PropertyGroup>
+
+    <PropertyGroup Condition="'$(Configuration)' == 'Debug'">
+      <DebugType>full</DebugType>
```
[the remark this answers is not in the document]
jamescrosswell: We can probably get rid of this actually. I copied this across from profiling when I was struggling to get line numbers for the code locations... but that turned out to be due to "unrelated" issues 😉

samples/Sentry.Samples.Console.Metrics/Program.cs
```diff
@@ -0,0 +1,111 @@
...
+    {
+        // Enable the SDK
+        using (SentrySdk.Init(options =>
+               {
+                   options.Dsn =
+                       // NOTE: ADD YOUR OWN DSN BELOW so you can see the events in your own Sentry account
+                       "https://eb18e953812b41c3aeb042e666fd3b5c@o447951.ingest.sentry.io/5428537";
+
+                   options.Debug = true;
+                   options.IsGlobalModeEnabled = true;
```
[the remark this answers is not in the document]
jamescrosswell: No I copied that across from the profiling sample as well...

src/Sentry/Protocol/Metrics/DistributionMetric.cs
jamescrosswell: Yeah good call... I'll wrap it.

src/Sentry/Protocol/Metrics/DistributionMetric.cs
jamescrosswell: I'll change it to Cast<T> actually... that's all the select clause is doing anyway.

src/Sentry/MetricAggregator.cs
jamescrosswell: I did try going Async... .NET Framework doesn't yet have the `IAsyncDisposable` interface so was complaining about missinc the Bcl.Interfaces package or something. It is just an interface so we could potentially conditionally shim it, but it seemed like a lot of complexity for very little gain, so I rolled it all back.

I'll try/catch the above though.

jamescrosswell:
> Only got as far as 30/52 files and didn't go deep on `MetricAggregator` and other major pieces but sending some bits as it's what I managed to cover

Awesome - thank you! The meat of it is all in the MetricAggregator so that one might take a little while.

src/Sentry/MetricHelper.cs
```diff
@@ -0,0 +1,50 @@
...
+    /// also apply independent jittering.
+    /// </summary>
+    /// <remarks>Internal for testing</remarks>
+    internal static double FlushShift = new Random().Next(0, 1000) * RollupInSeconds;
+    internal static DateTime GetCutoff() => DateTime.UtcNow
+        .Subtract(TimeSpan.FromSeconds(RollupInSeconds))
+        .Subtract(TimeSpan.FromMilliseconds(FlushShift));
+
+#if NET7_0_OR_GREATER
+    [GeneratedRegex(@"[^a-zA-Z0-9_/.-]+", RegexOptions.Compiled)]
```
bruno-garcia, line 39: Nice!!

src/Sentry/MetricHelper.cs
```diff
@@ -0,0 +1,50 @@
...
+    private static partial Regex InvalidKeyCharacters();
+    internal static string SanitizeKey(string input) => InvalidKeyCharacters().Replace(input, "_");
+
+    [GeneratedRegex(@"[^\w\d_:/@\.\{\}\[\]$-]+", RegexOptions.Compiled)]
+    private static partial Regex InvalidValueCharacters();
+    internal static string SanitizeValue(string input) => InvalidValueCharacters().Replace(input, "_");
+#else
+    internal static string SanitizeKey(string input) => Regex.Replace(input, @"[^a-zA-Z0-9_/.-]+", "_", RegexOptions.Compiled);
```
bruno-garcia, line 47: This is going to compile the regex on each use? Do we need to assign the regex to a field and reuse the field to leverage the compiled regex?

src/Sentry/Protocol/Envelopes/Envelope.cs
```diff
@@ -325,6 +326,34 @@ public static Envelope FromTransaction(Transaction transaction)
...
     }
 
+    /// <summary>
+    /// Creates an envelope that contains one or more <see cref="CodeLocations"/>
+    /// </summary>
+    internal static Envelope FromCodeLocations(CodeLocations codeLocations)
+    {
+        var header = DefaultHeader;
+
+        List<EnvelopeItem> items = [ EnvelopeItem.FromCodeLocations(codeLocations) ];
```
bruno-garcia, line 336: Does this create a List with capacity=1 ? If not the optimization I mentioned before didn't make it through (though this syntax is much nicer)

src/Sentry/Protocol/Metrics/DistributionMetric.cs
```diff
@@ -0,0 +1,37 @@
...
+    }
+
+    public DistributionMetric(string key, double value, MeasurementUnit? unit = null,
+        IDictionary<string, string>? tags = null, DateTime? timestamp = null)
+        : base(key, unit, tags, timestamp)
+    {
+        _value = new List<double>() { value };
+    }
+
+    private readonly List<double> _value;
```
bruno-garcia, line 23: fields above constructor?

src/Sentry/Protocol/Metrics/DistributionMetric.cs
```diff
@@ -0,0 +1,37 @@
...
+
+    public IReadOnlyList<double> Value => _value;
+
+    public override void Add(double value)
+    {
+        _value.Add(value);
+    }
```
bruno-garcia, line 30: wdyt?
```suggestion
    public override void Add(double value) => _value.Add(value);
```

src/Sentry/Protocol/Metrics/Metric.cs
```diff
@@ -0,0 +1,119 @@
...
+{
+    protected Metric() : this(string.Empty)
+    {
+    }
+
+    protected Metric(string key, MeasurementUnit? unit = null, IDictionary<string, string>? tags = null, DateTime? timestamp = null)
+    {
+        Key = key;
+        Unit = unit;
+        Tags = tags ?? new Dictionary<string, string>();
```
bruno-garcia, line 17: can we set the tags lazily instead? I imagine a lots of metrics will have no tags so we can spare the dic allocations

src/Sentry/Timing.cs
```diff
@@ -0,0 +1,108 @@
...
+                MeasurementUnit.Duration.Minute => _stopwatch.Elapsed.TotalMinutes,
+                MeasurementUnit.Duration.Second => _stopwatch.Elapsed.TotalSeconds,
+                MeasurementUnit.Duration.Millisecond => _stopwatch.Elapsed.TotalMilliseconds,
+                MeasurementUnit.Duration.Microsecond => _stopwatch.Elapsed.TotalMilliseconds * 1000,
+                MeasurementUnit.Duration.Nanosecond => _stopwatch.Elapsed.TotalMilliseconds * 1000000,
+                _ => throw new ArgumentOutOfRangeException(nameof(_unit), _unit, null)
+            };
+            _hub.Metrics.Timing(_key, value, _unit, _tags, _startTime);
+        }
+        catch(Exception e)
```
bruno-garcia, line 99: nit:
```suggestion
        catch (Exception e)
```

test/Sentry.Tests/MetricAggregatorTests.cs
```diff
@@ -0,0 +1,229 @@
...
+        resetEvent.WaitOne();
+        await sut.FlushAsync();
+
+        // Assert
+        sent.Should().Be(numThreads * numThreadIterations);
+    }
+
+    [Fact]
+    public void TestGetCodeLocation() {
+        // Arrange
```
bruno-garcia, line 218:
```suggestion
    public void TestGetCodeLocation()
    {
        // Arrange
```

bruno-garcia (commented):
Now at 48/52 (only what really matters is left lol)

Maybe on a follow up PR we should rename `MeasurementUnit`, it's quite verbose

src/Sentry/MetricAggregator.cs
```diff
@@ -0,0 +1,466 @@
...
+        }
+        finally
+        {
+            _codeLocationLock.Release();
+        }
+    }
+
+    internal SentryStackFrame? GetCodeLocation(int stackLevel)
+    {
+        var stackTrace = new StackTrace(true);
```
bruno-garcia, line 224: this is a rather expensive API so i wonder if we can leverage https://learn.microsoft.com/en-us/dotnet/api/system.runtime.compilerservices.callermembernameattribute?view=net-8.0 instead

src/Sentry/MetricAggregator.cs
```diff
@@ -0,0 +1,466 @@
...
+            }
+        }
+    }
+
+    private Dictionary<long, Dictionary<MetricResourceIdentifier, SentryStackFrame>> FlushableLocations()
+    {
+        _codeLocationLock.Wait();
+        try
+        {
+            var result = _pendingLocations;
```
bruno-garcia, line 373: You can do this atomically with Interlocked afaik, CompareExchange iirc

src/Sentry/MetricHelper.cs
jamescrosswell: Ah yes... my bad. I thought that was a static property but actually it's got parameters 😉

src/Sentry/Protocol/Envelopes/Envelope.cs
jamescrosswell: Apparently not... just tried it out and the deafult capacity is 4. I'll use the longer syntax that allocates less memory though.

src/Sentry/Protocol/Metrics/DistributionMetric.cs
jamescrosswell: We can. I've often been putting the fields next to the properties they back but happy to put them above the constructor.

src/Sentry/MetricAggregator.cs
jamescrosswell: I don't see an easy way to do so. There's no way to supply a stackLevel to the `CallerMemberName`, that I can see.

src/Sentry/MetricAggregator.cs
jamescrosswell: Hm, the `_codeLocationLock` is used elsewhere in the MetricAggregator to lock on different operations against _pendingLocations though. For example, there's this:
https://github.com/getsentry/sentry-dotnet/blob/1c3b2757cf352540472a97794ab7804b062177e2/src/Sentry/MetricAggregator.cs#L210-L214

src/Sentry/MetricAggregator.cs
```diff
@@ -0,0 +1,459 @@
...
+                return metric;
+            });
+
+        if (_options.ExperimentalMetrics is { EnableCodeLocations: true })
+        {
+            RecordCodeLocation(type, key, unit.Value, stackLevel + 1, timestamp.Value);
+        }
+    }
+
+    private readonly SemaphoreSlim _codeLocationLock = new(1,1);
```
bruno-garcia, line 173: please move this up next to the other fields

src/Sentry/MetricAggregator.cs
```diff
@@ -0,0 +1,459 @@
...
+        try
+        {
+            // Request the LoopTask stop.
+            await _shutdownSource.CancelAsync().ConfigureAwait(false);
+
+            // Now wait for the Loop to stop.
+            // NOTE: While non-intuitive, do not pass a timeout or cancellation token here.  We are waiting for
+            // the _continuation_ of the method, not its _execution_.  If we stop waiting prematurely, this may cause
+            // unexpected behavior in client applications.
+            LoopTask.Wait();
```
bruno-garcia, line 426: this is a blocking call, did you mean:
```suggestion
            await LoopTask;
```
?

src/Sentry/MetricAggregator.cs
```diff
@@ -0,0 +1,459 @@
...
+        }
+
+        foreach (var dailyValues in _seenLocations.Keys.ToArray())
+        {
+            if (dailyValues < today)
+            {
+                _seenLocations.TryRemove(dailyValues, out _);
+            }
+        }
+        lastClearedStaleLocations = today;
```
bruno-garcia, line 401: is this a field? it's hard now with some fields prefixed with `_` and some not. Is that due to the new constructor syntax?

src/Sentry/MetricAggregator.cs
```diff
@@ -0,0 +1,459 @@
...
+    /// </summary>
+    private void ClearStaleLocations()
+    {
+        var now = DateTimeOffset.UtcNow;
+        var today = now.GetDayBucketKey();
+        if (lastClearedStaleLocations == today)
+        {
+            return;
+        }
+        // Allow 60 seconds for all code locations to be sent at the transition from one day to the next
```
bruno-garcia, line 387: Is this what you mean? Reading the code below it allows for `< 1` so `Minute` has to be `0`. Allowing for the `Seconds` property to go from `0` to `59`

```suggestion
        // Allow less than 60 seconds for all code locations to be sent at the transition from one day to the next
```

src/Sentry/MetricAggregator.cs
```diff
@@ -0,0 +1,459 @@
...
+
+    // The key for this dictionary is the Timestamp for the bucket, rounded down to the nearest RollupInSeconds... so it
+    // aggregates all of the metrics data for a particular time period. The Value is a dictionary for the metrics,
+    // each of which has a key that uniquely identifies it within the time period
+    internal ConcurrentDictionary<long, ConcurrentDictionary<string, Metric>> Buckets => _buckets.Value;
+
+    private readonly Lazy<ConcurrentDictionary<long, ConcurrentDictionary<string, Metric>>> _buckets
+        = new(() => new ConcurrentDictionary<long, ConcurrentDictionary<string, Metric>>());
+
+    private long lastClearedStaleLocations = DateTimeOffset.UtcNow.GetDayBucketKey();
```
bruno-garcia, line 27:
```suggestion
    private long _lastClearedStaleLocations = DateTimeOffset.UtcNow.GetDayBucketKey();
```

src/Sentry/MetricAggregator.cs
```diff
@@ -0,0 +1,459 @@
...
+        MeasurementUnit unit,
+        int stackLevel,
+        DateTimeOffset timestamp
+    )
+    {
+        var startOfDay = timestamp.GetDayBucketKey();
+        var metaKey = new MetricResourceIdentifier(type, key, unit);
+
+        var seenToday = _seenLocations.GetOrAdd(startOfDay,_ => []);
+        if (seenToday.Contains(metaKey))
```
bruno-garcia, line 187: You're accessing the map outside of the lock and this will throw if one thread is on line 195 and another is here.

bruno-garcia (changes requested):
There's a race condition in `RecordCodeLocation`

jamescrosswell:
> There's a race condition in `RecordCodeLocation`

Do you mean this?
https://github.com/getsentry/sentry-dotnet/blob/4cb9ef746a0cf78b565f66186d4753938de53469/src/Sentry/MetricAggregator.cs#L187-L195

I'd considered putting another check for the `metaKey` inside the lock, but `seenToday` is a `HashSet`. Rider reckoned the code path could/would never get executed. 

I'll change it from a SemaphoreSlim to a ReadWriteSlim and upgrade to a write lock where necessary anyway... that way there's no doubt.

src/Sentry/MetricAggregator.cs
jamescrosswell: It checks if the minute is zero (less than 1) and if so it exists. So during the first minute (zero based) of every day, it does NOT try to clear the stale locations... then it'll clear them out.

src/Sentry/MetricAggregator.cs
jamescrosswell: No I think this just got overlooked. Maybe it wasn't private before or something... have changed.

src/Sentry/DelegatingMetricAggregator.cs
```diff
@@ -0,0 +1,31 @@
+namespace Sentry;
+
+internal class DelegatingMetricAggregator(IMetricAggregator innerAggregator) : IMetricAggregator
```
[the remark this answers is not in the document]
jamescrosswell: All of the classes that implement ISentryClient have an `IMetricAggregator Metrics` property. The DelegatingMetricAggregator is used by the Hub to delegate any calls to it's owned client:
https://github.com/getsentry/sentry-dotnet/blob/8e7f0ce5baa655d9c641919cb9f84d19efb98de8/src/Sentry/Internal/Hub.cs#L62

Now I look at it though, I see what you're saying. Maybe that line could be replaced with:
```
Metrics = _ownedClient.Metrics;
```

In which case, we wouldn't need the `DelegatingMetricAggregator`...

src/Sentry/MetricHelper.cs
```diff
@@ -0,0 +1,52 @@
...
+    }
+
+    /// <summary>
+    /// The aggregator shifts it's flushing by up to an entire rollup window to avoid multiple clients trampling on end
+    /// of a 10 second window as all the buckets are anchored to multiples of ROLLUP seconds.  We randomize this number
+    /// once per aggregator boot to achieve some level of offsetting across a fleet of deployed SDKs.  Relay itself will
+    /// also apply independent jittering.
+    /// </summary>
+    /// <remarks>Internal for testing</remarks>
+    internal static double FlushShift = new Random().Next(0, 1000) * RollupInSeconds;
```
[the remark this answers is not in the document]
jamescrosswell: Not sure I follow:
> we need the same instance to have proper pseudo-randomness

We need the same instance where? This is ensuring this instance of Sentry is using a different flush shift to other instances being deployed in a fleet. Within this process, this code is being called (effectively) in a static constructor... which should be guaranteed to be called only once right? I thought that was one of the guarantees of static constructors?

I can use `Random.Shared`, but at the moment I don't quite understand why we want/need to do that...

src/Sentry/Protocol/Metrics/CodeLocations.cs
```diff
@@ -0,0 +1,43 @@
...
+internal class CodeLocations(long timestamp, Dictionary<MetricResourceIdentifier, SentryStackFrame> locations)
+    : IJsonSerializable
+{
+    /// <summary>
+    /// Uniquely identifies a code location using the number of seconds since the UnixEpoch, as measured at the start
+    /// of the day when the code location was recorded.
+    /// </summary>
+    public long Timestamp { get; set; } = timestamp;
+
+    public Dictionary<MetricResourceIdentifier, SentryStackFrame> Locations { get; set; } = locations;
```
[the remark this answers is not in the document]
jamescrosswell: Yeah I think we can make it all immutable 👍🏻

src/Sentry/Protocol/Metrics/CodeLocations.cs
```diff
@@ -0,0 +1,43 @@
...
+namespace Sentry.Protocol.Metrics;
+
+internal class CodeLocations(long timestamp, Dictionary<MetricResourceIdentifier, SentryStackFrame> locations)
+    : IJsonSerializable
+{
+    /// <summary>
+    /// Uniquely identifies a code location using the number of seconds since the UnixEpoch, as measured at the start
+    /// of the day when the code location was recorded.
+    /// </summary>
+    public long Timestamp { get; set; } = timestamp;
```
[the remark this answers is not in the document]
jamescrosswell: According to [the docs](https://learn.microsoft.com/en-us/dotnet/csharp/whats-new/tutorials/primary-constructors):
> Primary constructor parameters don't become properties, except in [record](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/record) types.

There's some ambiguity though... later when describing one of the examples there's this:
> In the previous example, the primary constructor properties are accessed in a method. Therefore the compiler creates hidden fields to represent each parameter. 

The example they're referring to mutates the value of the primary constructor parameter values... which we never do in `CodeLocations`, so it's hard to say without looking at the IL whether a redundant backing field is being created or not in the case of `CodeLocations`.

I'll get rid of the field for the Locations... `Timestamp` is used outside the class though so I need to expose that as a read only field at least.

src/Sentry/MetricAggregator.cs
jamescrosswell: I've expanded the scope of the lock... should be OK now I think.

src/Sentry/Protocol/Metrics/CodeLocations.cs
```diff
@@ -0,0 +1,42 @@
+using Sentry.Extensibility;
+using Sentry.Internal.Extensions;
+
+namespace Sentry.Protocol.Metrics;
+
+
```
bitsandfoxes, line 6:
```suggestion
```

src/Sentry/IMetricAggregator.cs
```diff
@@ -0,0 +1,107 @@
...
+    /// <param name="key">A unique key identifying the metric</param>
+    /// <param name="value">The value to be added</param>
+    /// <param name="unit">An optional <see cref="MeasurementUnit"/></param>
+    /// <param name="tags">Optional Tags to associate with the metric</param>
+    /// <param name="timestamp">
+    ///     The time when the metric was emitted. Defaults to the time at which the metric is emitted, if no value is provided.
+    /// </param>
+    /// <param name="stackLevel">Optional number of stacks levels to ignore when determining the code location</param>
+    void Set(string key,
+        double value = 1.0,
```
bitsandfoxes, line 75: Should this be an `int` instead? The `SetMetric` only deals with integers so this might be a bit misleading.

src/Sentry/Protocol/Envelopes/Envelope.cs
```diff
@@ -325,6 +326,35 @@ public static Envelope FromTransaction(Transaction transaction)
...
+    /// </summary>
+    internal static Envelope FromMetrics(IEnumerable<Metric> metrics)
+    {
+        var header = DefaultHeader;
+
+        List<EnvelopeItem> items = new();
+        foreach (var metric in metrics)
+        {
+            items.Add(EnvelopeItem.FromMetric(metric));
+        }
```
bitsandfoxes, line 353:
```suggestion
        var items = metrics.Select(EnvelopeItem.FromMetric).ToList();
```
We could give LINQ a go. :D

src/Sentry/SentryClient.cs
```diff
@@ -225,6 +241,24 @@ public void CaptureTransaction(Transaction transaction, Scope? scope, Hint? hint
         return transaction;
     }
 
+    /// <summary>
+    /// Captures one or more metrics to be sent to Sentry.
+    /// </summary>
+    internal void CaptureMetrics(IEnumerable<Metric> metrics)
+    {
+        _options.LogDebug($"Capturing metrics");
```
bitsandfoxes, line 249:
```suggestion
        _options.LogDebug("Capturing metrics.");
```

bitsandfoxes (commented):
This looks amazing!

src/Sentry/IMetricAggregator.cs
```diff
@@ -0,0 +1,107 @@
...
+    void Timing(string key,
+        double value,
+        MeasurementUnit.Duration unit = MeasurementUnit.Duration.Second,
+        IDictionary<string, string>? tags = null,
+        DateTimeOffset? timestamp = null,
+        int stackLevel = 1);
+
+    /// <summary>
+    /// Flushes any flushable metrics and/or code locations.
+    /// If <paramref name="force"/> is true then the cutoff is ignored and all metrics are flushed.
```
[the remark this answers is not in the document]
jamescrosswell: That's from here:
https://github.com/getsentry/sentry-dotnet/blob/705d131195f13b70e60f5dff8b1573ac9313891f/src/Sentry/MetricAggregator.cs#L398-L406

At the moment it's hard wired for 10s buckets (which is the same behaviour as the Python implementation). So every 10 seconds, all the metrics from the last 10 seconds get sent through to Sentry. I'm not sure if users should be able to configure this - the buckets are probably more about managing load on Sentry's servers than anything.

src/Sentry/IMetricAggregator.cs
```diff
@@ -0,0 +1,107 @@
...
+    /// </summary>
+    /// <param name="force">Forces all buckets to be flushed, ignoring the cutoff</param>
+    /// <param name="cancellationToken">A <see cref="CancellationToken"/></param>
+    /// <returns>False if a shutdown is requested during flush, true otherwise</returns>
+    Task FlushAsync(bool force = true, CancellationToken cancellationToken = default);
```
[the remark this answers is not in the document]
jamescrosswell: In the concrete implementation, a shutdown token can be passed in when that's the reason for flushing, yes:
https://github.com/getsentry/sentry-dotnet/blob/705d131195f13b70e60f5dff8b1573ac9313891f/src/Sentry/MetricAggregator.cs#L310

src/Sentry/MetricAggregator.cs
```diff
@@ -0,0 +1,508 @@
...
+    /// <param name="captureMetrics">The callback to be called to transmit aggregated metrics</param>
+    /// <param name="captureCodeLocations">The callback to be called to transmit new code locations</param>
+    /// <param name="shutdownSource">A <see cref="CancellationTokenSource"/></param>
+    /// <param name="disableLoopTask">
+    /// A boolean value indicating whether the Loop to flush metrics should run, for testing only.
+    /// </param>
+    /// <param name="flushInterval">An optional flushInterval, for testing only</param>
+    public MetricAggregator(SentryOptions options, Action<IEnumerable<Metric>> captureMetrics,
+        Action<CodeLocations> captureCodeLocations, CancellationTokenSource? shutdownSource = null,
+        bool disableLoopTask = false, TimeSpan? flushInterval = null)
```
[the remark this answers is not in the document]
jamescrosswell: Yeah, I can make the constructor internal to make that more apparent though.

src/Sentry/Internal/Hub.cs
```diff
@@ -20,6 +20,9 @@ internal class Hub : IHub, IDisposable
 
     internal IInternalScopeManager ScopeManager { get; }
 
+    /// <inheritdoc cref="IMetricAggregator"/>
+    public IMetricAggregator Metrics { get; }
```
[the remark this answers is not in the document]
jamescrosswell: Yeah, it's on the interface:

https://github.com/getsentry/sentry-dotnet/blob/c65355dede946090646be759d5a39f3df19fd4fb/src/Sentry/ISentryClient.cs#L75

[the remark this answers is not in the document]
jamescrosswell: I think it has to be on the interface, otherwise the SentrySdk can't get at it here:
https://github.com/getsentry/sentry-dotnet/blob/36fe0291aa87ca4e561953c47d3c75daf5ad19fe/src/Sentry/SentrySdk.cs#L621-L623

src/Sentry/IMetricAggregator.cs
jamescrosswell: Yeah, good call. I'll remove the default value as well (there's no reason for this to be arbitrarily 1 or 0 or anything else).

src/Sentry/Protocol/Envelopes/Envelope.cs
jamescrosswell: I think in an earlier version of the code I was. Bruno suggested removing linq for performance reasons... it's a pity - linq is easier on the eyes (just harder on the CPU).

bruno-garcia: I'd avoid LINQ as a plague in hot paths

bruno-garcia: but worth measuring when it matters, I just find a loop is pretty simple and often default to that if it's hot path

src/Sentry/Protocol/Metrics/CodeLocations.cs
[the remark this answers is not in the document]
bruno-garcia: > Primary constructor parameters don't become properties, except in [record](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/record) types.

TIL, sorry not familiar with this (well deserved RTFM 😅 )

src/Sentry/MetricHelper.cs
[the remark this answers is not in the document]
bruno-garcia: I think I misread this and saw the random instance getting created on each access (PascalCase read as a prop but static internal named follows that so makes sense as-is.

The right-side of the assignment will run in the static constructor so I believe there's no concern with concurrency and it's fine as-is. Unless we have this approach elsewhere in the code where both Shifts could come out with the same number (on .NET Framework). That's the reason to use a single instance throughout the app for randomness as opposed to creating new instances

src/Sentry/Timing.cs
```diff
@@ -0,0 +1,100 @@
...
+        _key = key;
+        _unit = unit;
+        _tags = tags;
+        _stopwatch.Start();
+
+        ITransactionTracer? currentTransaction = null;
+        hub.ConfigureScope(s => currentTransaction = s.Transaction);
+        _span = currentTransaction is {} transaction
+            ? transaction.StartChild("metric.timing", key)
+            : hub.StartTransaction("metric.timing", key);
```
bruno-garcia, line 57: Just spotted this (sorry if it was here from the start). But are we creating a transaction named `metric.timing` that we don't attach to the scope to hold this metric? Curious if we do that in Python too

src/Sentry/SentryClient.cs
```diff
@@ -225,6 +241,24 @@ public void CaptureTransaction(Transaction transaction, Scope? scope, Hint? hint
...
+        CaptureEnvelope(Envelope.FromMetrics(metrics));
+    }
+
+    /// <summary>
+    /// Captures one or more <see cref="CodeLocations"/> to be sent to Sentry.
+    /// </summary>
+    internal void CaptureCodeLocations(CodeLocations codeLocations)
+    {
+        _options.LogDebug($"Capturing code locations for period: {codeLocations.Timestamp}");
+        CaptureEnvelope(Envelope.FromCodeLocations(codeLocations));
```
bruno-garcia, line 259: Are code locations not coupled with metrics? In which case they should be on the same envelope ideally

bruno-garcia (approved):
There are a few comments that I might have dropped outside a review that are pending. Could you please take a look on the files view: https://github.com/getsentry/sentry-dotnet/pull/2949/files

I don't see any blockers so I'm happy to merge this and ship a `beta` so we can test. It's just too hard to review this change now and we can do improvements/fixes in follow up PRs now.

src/Sentry/Internal/Hub.cs
```diff
@@ -520,7 +525,11 @@ public void Dispose()
             return;
         }
 
-        _ownedClient.Flush(_options.ShutdownTimeout);
+        var disposeTasks = new List<Task> {
+            _ownedClient.Metrics.FlushAsync(),
+            _ownedClient.FlushAsync(_options.ShutdownTimeout)
```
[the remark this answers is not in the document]
jamescrosswell: I figured they could run in parallel... so we could start sending envelopes immediately.

src/Sentry/SentryClient.cs
```diff
@@ -225,6 +241,24 @@ public void CaptureTransaction(Transaction transaction, Scope? scope, Hint? hint
...
+        _options.LogDebug($"Capturing metrics");
+        CaptureEnvelope(Envelope.FromMetrics(metrics));
+    }
+
+    /// <summary>
+    /// Captures one or more <see cref="CodeLocations"/> to be sent to Sentry.
+    /// </summary>
+    internal void CaptureCodeLocations(CodeLocations codeLocations)
+    {
+        _options.LogDebug($"Capturing code locations for period: {codeLocations.Timestamp}");
```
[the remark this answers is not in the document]
jamescrosswell: I can make this change... not sure I understand the comment though. 

I know it's useful not to use string interpolation if we're asserting that something was logged in tests but what's the deal with Debug:false and string interpolation?

src/Sentry/Protocol/Metrics/Metric.cs
```diff
@@ -0,0 +1,119 @@
...
+            await stream.WriteAsync(Encoding.UTF8.GetBytes(content), cancellationToken).ConfigureAwait(false);
+        }
+    }
+
+    public void Serialize(Stream stream, IDiagnosticLogger? logger)
+    {
+        SerializeAsync(stream, logger).GetAwaiter().GetResult();
+    }
+
+    private string StatsdType =>
```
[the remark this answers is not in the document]
jamescrosswell: We do? I can't find it.

src/Sentry/Protocol/Metrics/Metric.cs
```diff
@@ -0,0 +1,119 @@
...
+
+        async Task Write(string content)
+        {
+            await stream.WriteAsync(Encoding.UTF8.GetBytes(content), cancellationToken).ConfigureAwait(false);
+        }
+    }
+
+    public void Serialize(Stream stream, IDiagnosticLogger? logger)
+    {
+        SerializeAsync(stream, logger).GetAwaiter().GetResult();
```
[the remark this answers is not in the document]
jamescrosswell: It's required to implement the `Sentry.Protocol.Envelopes.ISerializable` interface.

src/Sentry/SentryClient.cs
jamescrosswell: I think I did that because of this:
https://github.com/getsentry/sentry-python/blob/67c963d9c8d5e7e9de6347aee0edcf0c58d9fb24/sentry_sdk/metrics.py/#L580-L581

src/Sentry/Timing.cs
jamescrosswell: Yeah that's lifted from here:
https://github.com/getsentry/sentry-python/blob/67c963d9c8d5e7e9de6347aee0edcf0c58d9fb24/sentry_sdk/metrics.py/#L790

Python doesn't do anything special with the scope but is that due to fundamental differences between the Python SDK and the .NET SDK? Should we be setting the scope transaction in .NET?

src/Sentry/MetricAggregator.cs
```diff
@@ -0,0 +1,508 @@
...
+            options.LogDebug("Starting MetricsAggregator.");
+            LoopTask = Task.Run(RunLoopAsync);
+        }
+    }
+
+    internal static string GetMetricBucketKey(MetricType type, string metricKey, MeasurementUnit unit,
+        IDictionary<string, string>? tags)
+    {
+        var typePrefix = type.ToStatsdType();
+        var serializedTags = GetTagsKey(tags);
```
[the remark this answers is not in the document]
jamescrosswell: That's just mimicing the Python implementation. It might be more computationally expensive to generate a hash than to serialize these... although it would be less expensive in terms of memory footprint, so could go either way on this.

src/Sentry/Protocol/Metrics/Metric.cs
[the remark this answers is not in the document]
jamescrosswell: Ah I see... yeah there's one that switches on the `MetricType` enum and another that switches on the Metric class type... potentially we could add an abstract MetricType property to the Metric base class and use that instead of a switch on the class type - then we could get rid of one of these.

bitsandfoxes (approved):

src/Sentry/Timing.cs
bruno-garcia: > Should we be setting the scope transaction in .NET?
No lets not, it's fine then, we're just creating a local transaction to be a container for the metric. Let see how this works

PR #2952, opened 2023-12-06
Verify that test server recieves proguard mapping file

Giorgi:
#skip-changelog


integration-test/assets/artifact.json
```diff
@@ -0,0 +1,10 @@
+{
```
vaind, line 1: I think it would be better to leave these in the action repo. The whole point of that action is to avoid having a copy in each repository that runs cli integration tests.

vaind (changes requested):

vaind (approved):

Issue #2954, opened 2023-12-06
MAUI MacCatalyst Apps have a `default.profraw` generated inside the .app bundle when the app exits


Redth:
### Package

Sentry

### .NET Flavor

.NET

### .NET Version

8.0.0

### OS

macOS

### SDK Version

4.0.0-beta.4

### Steps to Reproduce

1. dotnet new maui
2. Add Sentry, configure it for a MAUI project
3. Build/Debug the MacCatalyst target
4. Quit the app (IMPORTANT: Quit it gracefully, do not terminate it)
5. Observe in the `bin/Debug/net8.0-maccatalyst/MauiApp1.app` bundle there should be a `default.profraw` file now
6. Make a change to some c# and/or xaml in the app
7. Build/Debug MacCatalyst target again
8. Observe a build error which complains about codesigning and unsigned contents in bundle, ultimately pointing at the `default.profraw` file.

### Expected Result

This file should not get created at least in this location.

### Actual Result

File gets created, causing subsequent incremental builds to fail.

Video:
https://github.com/getsentry/sentry-dotnet/assets/271950/bae02c29-aaa3-42e2-86b0-8322ec208d0f

Sample/Repro project: 
[Archive.zip](https://github.com/getsentry/sentry-dotnet/files/13601458/Archive.zip)

bruno-garcia:
I tried:
1. Created a new app
2. Built through `dotnet build -f net8.0-maccatalyst` and spawned the app from finder: CMD+O
3. Validated no `default.profraw`
4. Added Sentry 4.0.0-beta.4
5. Built through `dotnet build -f net8.0-maccatalyst` and spawned the app from finder: CMD+O
6. Validated no `default.profraw`
7. Changed some code, and `dotnet build -f net8.0-maccatalyst` and spawned the app from finder: CMD+O
8. Still no `default.profraw`

Changed code again and tried to build/launch via `dotnet build -t:Run` instead of via finder. Still couldn't repro.

This is what I see in the bundle:
![image](https://github.com/getsentry/sentry-dotnet/assets/1633368/4a469ed4-875d-4ed3-bfb6-c17448585c16)

akoeplinger:
Seems this is enabled via the `-fprofile-instr-generate` flag when compiling LLVM https://clang.llvm.org/docs/SourceBasedCodeCoverage.html#compiling-with-coverage-enabled

[...]

PR #2960, opened 2023-12-07
Use xcodebuild archive to build sentry-cocoa

akoeplinger:
Fixes https://github.com/getsentry/sentry-dotnet/issues/2954

Without the "archive" command xcodebuild will build with code coverage enabled (`-fprofile-instr-generate` is passed to clang) and that links in the compiler profiling runtime. This is evidenced by the `__llvm_profile_runtime` symbol being present in the `Sentry` binary with `nm`. The static initializer will hook up profiling and we get a `default.profraw` file when the app exits.

The fix is to use the "archive" command instead, similar to how upstream sentry-cocoa uses it here: https://github.com/getsentry/sentry-cocoa/blob/ddb47781bffd76406f31ab43e66760362585fcd7/Makefile#L82-L83

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Fixes
+
+- Stop Sentry for MacCatalyst from creating `default.profraw` in the app bundle using xcodebuild archive to build sentry-cocoa ([#2960](https://github.com/getsentry/sentry-dotnet/pull/2960))
+
 ### Dependencies
 
 - Bump CLI from v2.22.3 to v2.23.0 ([#2956](https://github.com/getsentry/sentry-dotnet/pull/2956))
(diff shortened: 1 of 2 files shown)


bruno-garcia (approved):

bruno-garcia:
@akoeplinger 

Could add to `CHANGELOG.md` below `## Unreleased`:


```
### Fixes

- Stop Sentry for MacCatalyst from creating `default.profraw` in the app bundle using xcodebuild archive to build sentry-cocoa ([#2960](https://github.com/getsentry/sentry-dotnet/pull/2960))
```

bruno-garcia:
Thanks a lot for finding the culprit and fixing the problem!!

akoeplinger:
Hmm looks like it's is now building for macOS instead of Catalyst :)

I'll have another look tomorrow.

bruno-garcia:
> Hmm looks like it's is now building for macOS instead of Catalyst :)
> 
> I'll have another look tomorrow.

Oh, we build [`sentry-native`](https://github.com/getsentry/sentry-native) for macOS (+Win/Linux). And [`sentry-cocoa`](https://github.com/getsentry/sentry-cocoa) for iOS and Catalyst (and use ObjSharpie etc)

akoeplinger:
Figured out the issue, we need to use `generic/` prefix for the destination arg...

vaind (approved):

Issue #1754, opened 2022-06-26
Screenshots for Android, iOS and MAUI


bruno-garcia:
### Problem Statement

Sentry supports screenshots for UI frameworks. This is adopted in Sentry's iOS, Android, Unity and Xamarin SDKs.

![image](https://user-images.githubusercontent.com/1633368/175831725-e33344cd-a399-41dd-b72e-63ef281d7728.png)

This potentially could be made available on desktop (WPF/WinForms/Avalonia/Uno) too.

### Solution Brainstorm

SDK dev docs: https://develop.sentry.dev/sdk/features/#screenshots

Done on Symbol Collector for target `net6.0-android` but not sure this approach will work unless the final app has `UseMauiEssentials`:

[getsentry/symbol-collector@`6a8458a` (#129)](https://github.com/getsentry/symbol-collector/pull/129/commits/6a8458a98883f7101bf802461a89be0fe2c7c462)

Related issues:

* https://github.com/getsentry/team-mobile/issues/17

PR #2965, opened 2023-12-11
MAUI screenshot

Giorgi:
Resolves: https://github.com/getsentry/sentry-dotnet/issues/1754


CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Features
+
+- Add an option to attach screenshots to crashes for MAUI apps. Configurable via `SentryMauiOptions.AttachScreenshots` ([#2965](https://github.com/getsentry/sentry-dotnet/pull/2965))
```
bruno-garcia, line 7: More 'catchy' like this:

```suggestion
- MAUI Screenshot support. You can opt-in via `SentryMauiOptions.AttachScreenshots` ([#2965](https://github.com/getsentry/sentry-dotnet/pull/2965))
```

src/Sentry.Maui/BindableSentryMauiOptions.cs
```diff
@@ -9,6 +9,7 @@ internal class BindableSentryMauiOptions : BindableSentryLoggingOptions
     public bool? IncludeTitleInBreadcrumbs { get; set; }
     public bool? IncludeBackgroundingStateInBreadcrumbs { get; set; }
     public bool? CreateElementEventsBreadcrumbs { get; set; } = false;
+    public bool? AttachScreenshots { get; set; } = false;
```
bruno-garcia, line 12: If we default to `false`, why is it nullable? @jamescrosswell might know since he did this bindable bits?

src/Sentry.Maui/Internal/ScreenshotAttachment.cs
```diff
@@ -0,0 +1,39 @@
...
+    }
+}
+
+internal class ScreenshotAttachmentContent : IAttachmentContent
+{
+    public Stream GetStream()
+    {
+        if (Screenshot.Default.IsCaptureSupported)
+        {
+            var screen = Screenshot.Default.CaptureAsync().ConfigureAwait(false).GetAwaiter().GetResult();
```
bruno-garcia, line 30: Off there's no way to make this `async` instead? The transport is `async` but I imagine we'd want the screenshot done at the time of the error and not later on in the transport once "the relevant point in time is gone".

Q:
1. Is there a risk of deadlocking here?
2. Does this need to run on the UI thread? What happens if it's called on a backgorund thread (happens if the error is captured on a background thread)

Finally, this is actually only going to run when we call `GetStream` which is indeed on the background worker, could you confirm?
We'd want this to run synchronously when `Capture` is called into the SDK to get the screen at that moment.

src/Sentry.Maui/Internal/SentryMauiInitializer.cs
```diff
@@ -15,6 +15,12 @@ public void Initialize(IServiceProvider services)
         // Initialize the Sentry SDK.
         var disposable = SentrySdk.Init(options);
 
+        if (options.AttachScreenshots)
+        {
+            SentrySdk.ConfigureScope(s =>
```
bruno-garcia, line 20: Does this mean if I call `scope.Clear()` the screenshot is gone? I believe we had issues with that on Unity cc @bitsandfoxes

bruno-garcia (commented):
We want the screenshot to show up on the `Hint` so users can remove it in some conditions (for example if `transaction==payment` or something.

bruno-garcia:
One last thought: Can this be decoupled from MAUI?

The native Android and iOS SDKs also have screenshots so we could rely on those at least on mobile. And for MAUI outside mobile we use this one you're adding

src/Sentry.Maui/BindableSentryMauiOptions.cs
Giorgi: Removed the default value.

src/Sentry.Maui/Internal/SentryMauiInitializer.cs
Giorgi: Modified to use Hint so it isn't cleared any more.

src/Sentry.Maui/Internal/ScreenshotAttachment.cs
Giorgi: Just tried it and it's deadlocking on UWP. Shouldn't be an issue on other platforms because on other platforms it runs synchronously and returns `Task.FromResult`

src/Sentry.Maui/BindableSentryMauiOptions.cs
bruno-garcia: The value above has a default, should this have a default? Should we remove it everywhere (where is default source of truth)?

Still hoping @jamescrosswell can chime in he might know these answers

src/Sentry.Maui/Internal/ScreenshotAttachment.cs
bruno-garcia: Could you please add a comment block with pointers? I think it'll help put ppl's mind at ease: 

```
// This actually runs synchronously (returning Task.FromResult) on the following platforms:
// Android: https://github.com/dotnet/android/.../Bla.cs
// iOS/macOS: https://github.com/dotnet/macios/.../Bla.cs

```

What's the strategy on UWP? Do we skip with `#ifdef` ?

src/Sentry.Maui/Internal/SentryMauiInitializer.cs
bruno-garcia: `SentryMauiScreenshotProcessor` is in the scope, if I `scope.Clear()` happens the processor is gone

Giorgi: Should be resolved now.

src/Sentry.Maui/Internal/ScreenshotAttachment.cs
Giorgi: Add comments and skip on Windows.

bruno-garcia:
I gave two passes while it was draft, I is it ready for review? if so please mark it as such

Giorgi:
I haven't added tests yet, otherwise I think it's ready.

src/Sentry.Maui/Internal/ScreenshotAttachment.cs
```diff
@@ -0,0 +1,52 @@
...
+    }
+}
+
+internal class ScreenshotAttachmentContent : IAttachmentContent
+{
+    public Stream GetStream()
+    {
+        try
+        {
+#if !WINDOWS
```
bruno-garcia, line 30: So this won't compile it on `net8.0-windows` targets because it deadlocks, right?
This is very useful context worth adding here:

```suggestion
// Not including this on Windows specific build because on WinUI this can deadlock.
#if !WINDOWS
```

src/Sentry.Maui/Internal/ScreenshotAttachment.cs
```diff
@@ -0,0 +1,52 @@
...
+            if (Screenshot.Default.IsCaptureSupported)
+            {
+                // This actually runs synchronously (returning Task.FromResult) on the following platforms:
+                // Android: https://github.com/dotnet/maui/blob/3c7b65264d2f341a48db32263a271fd8718cfd23/src/Essentials/src/Screenshot/Screenshot.android.cs#L49
+                // iOS: https://github.com/dotnet/maui/blob/3c7b65264d2f341a48db32263a271fd8718cfd23/src/Essentials/src/Screenshot/Screenshot.ios.cs#L49
+                var screen = Screenshot.Default.CaptureAsync().ConfigureAwait(false).GetAwaiter().GetResult();
```
bruno-garcia, line 36: Very helpful! Thanks. These blocking calls raise alarm bells (since they can deadlock as they do on WIndows) so helpful to set the reader at ease here.

src/Sentry.Maui/Internal/ScreenshotAttachment.cs
```diff
@@ -0,0 +1,52 @@
...
+            }
+#endif
+        }
+        //In some cases screen capture can throw, for example on Android if the activity is marked as secure.
+        catch (Exception)
+        {
+            return MemoryStream.Null;
+        }
+
+        return MemoryStream.Null;
```
bruno-garcia, line 50: No need to go through the `try/catch` on Windows if `try` doesn't do anything.

Also, you can take the `SentryOptions` via the constructor so you can log out problems to ease debugging.
Folks will raise tickets: "Why the screenshot isn't getting attached" and we need to have tools to help us debug

Finally, wrt Android Secure Actitivies is there a way to check? If so, you could do something like: 
```
#if ANDROID
if (Android.Activity.IsSecure) { return
```

```suggestion
        }
        //In some cases screen capture can throw, for example on Android if the activity is marked as secure.
        catch (Exception ex)
        {
            _options.LogError(ex, "Failed to capture screenshot");
            return MemoryStream.Null;
        }
#else
        return MemoryStream.Null;
#endif
```

src/Sentry.Maui/Internal/ScreenshotAttachment.cs
bruno-garcia, line 36: Very helpful! Thanks. These blocking calls raise alarm bells (since they can deadlock as they do on WIndows) so helpful to set the reader at ease here.

src/Sentry.Maui/Internal/SentryMauiInitializer.cs
bruno-garcia: You did so by adding to the options instead? (a clarification on how you resolves (as opposed to "should be resolved") helps the reviewer understand how.

I see this is added to both the scope and the options now, is this intended? Are we taking 2 screenshots possibly?

test/Sentry.Maui.Tests/SentryMauiOptionsTests.cs
```diff
@@ -49,4 +49,11 @@ public void CacheDirectoryPath_Default()
         Assert.Equal(expected, options.CacheDirectoryPath);
 #endif
     }
+
+    [Fact]
+    public void AttachScreenshots_Default()
+    {
+        var options = new SentryMauiOptions();
+        Assert.False(options.AttachScreenshots);
```
bruno-garcia, line 57: Some other tests that would help here:

1. Scope.clear doesn't affect screenshot capture
2. Setting it to true and an event is capture actually ends up with a screenshot as attachment (can verify on the transport level)
3. Remove screenshot from hint on beforeSend makes sure it doesn't show up on the transport later

src/Sentry.Maui/Internal/ScreenshotAttachment.cs
Giorgi: Added logging. As for the secure check, the code can throw in some other cases too so it's safer to have try-catch

src/Sentry.Maui/Internal/SentryMauiInitializer.cs
Giorgi: Yeah, I forgot to remove it from scope :)

src/Sentry.Maui/Internal/ScreenshotAttachment.cs
```diff
@@ -0,0 +1,68 @@
...
+                else
+                {
+                    _options.DiagnosticLogger?.Log(SentryLevel.Warning, "Capturing screenshot not supported");
+                    return Stream.Null;
+                }
+            }
+            //In some cases screen capture can throw, for example on Android if the activity is marked as secure.
+            catch (Exception ex)
+            {
+                _options.DiagnosticLogger?.Log(SentryLevel.Error, "Error capturing screenshot", ex);
```
bruno-garcia, line 58: there are extension method direclty on option:

```suggestion
                    _options.LogWarning("Capturing screenshot not supported");
                    return Stream.Null;
                }
            }
            //In some cases screen capture can throw, for example on Android if the activity is marked as secure.
            catch (Exception ex)
            {
                _options.LogError(ex, "Error capturing screenshot");
```

src/Sentry.Maui/Internal/ScreenshotAttachment.cs
```diff
@@ -0,0 +1,68 @@
...
+    public ScreenshotAttachmentContent(SentryMauiOptions options)
+    {
+        _options = options;
+    }
+
+    public Stream GetStream()
+    {
+        // Not including this on Windows specific build because on WinUI this can deadlock.
+#if !WINDOWS
+        var stream = MainThread.InvokeOnMainThreadAsync(async () =>
```
bruno-garcia, line 37: Does this run synchronously if called on the UI thread?
I ask that trying to answer:

Does this take the screenshot at the point in time when the error happens or after? In other words does the UI re-render before we actually take the screenshot?

src/Sentry.Maui/Internal/ScreenshotAttachment.cs
```diff
@@ -0,0 +1,68 @@
...
+                    return Stream.Null;
+                }
+            }
+            //In some cases screen capture can throw, for example on Android if the activity is marked as secure.
+            catch (Exception ex)
+            {
+                _options.DiagnosticLogger?.Log(SentryLevel.Error, "Error capturing screenshot", ex);
+                return Stream.Null;
+            }
+        }).ConfigureAwait(false).GetAwaiter().GetResult();
```
bruno-garcia, line 61: Before you had this call and was confident it wasn't blocking anything (u linked source of how it was `Task.CompletedTask`.  
But now this is doing a blocking call on async task?

It looks like now we don't run it synchronously anymore (see above), which is not ideal here. We need to see what was on the screen at the time of the error.

test/Sentry.Maui.Tests/SentryMauiScreenshotTests.cs
```diff
@@ -0,0 +1,87 @@
...
+        var options = app.Services.GetRequiredService<IOptions<SentryMauiOptions>>().Value;
+        var transport = options.Transport as FakeTransport;
+
+        var envelope = transport.GetSentEnvelopes().FirstOrDefault(e => e.TryGetEventId() == sentryId);
+        envelope.Should().NotBeNull();
+
+        var envelopeItem = envelope!.Items.FirstOrDefault(item => item.TryGetType() == "attachment");
+
+        // Assert
+#if __MOBILE__
```
bruno-garcia, line 53: else we don't assert anything, is this test supposed to only exist in mobile then? Looks lke we're missing something here for non mobile

test/Sentry.Maui.Tests/SentryMauiScreenshotTests.cs
```diff
@@ -0,0 +1,87 @@
...
+        envelopeItem.Should().NotBeNull();
```
bruno-garcia, line 54: `Stream.Null` isn't `null`: https://learn.microsoft.com/en-us/dotnet/api/system.io.stream.null?view=net-8.0

So this isnt' verifying you have what you expect here.
Can we check this is a valid `jpg` or `png` or whatever this is? Or at least read the bytes and make sure it's more than 1 or something smarter that checks there's a screenshot here

test/Sentry.Maui.Tests/SentryMauiScreenshotTests.cs
Giorgi: There is nothing to verify for mobile, I can wrap the whole test in #ifdef

src/Sentry.Maui/SentryMauiOptions.cs
```diff
@@ -56,4 +56,12 @@ public SentryMauiOptions()
...
     /// </summary>
     public bool CreateElementEventsBreadcrumbs { get; set; } = false;
+
+    /// <summary>
+    /// Automatically attaches a screenshot of the app at the time of the event capture.
+    /// </summary>
+    /// <remarks>
+    /// Be aware PII might be included by activating this feature.
+    /// </remarks>
+    public bool AttachScreenshots { get; set; }
```
bruno-garcia, line 66: The property name should be `AttachScreenshot`. As it's on the native iOS/Android.

Also please make sure to pass the value to the native layer in both platforms (prob MacCatalyst too) since that controls a screenshot if a native crash happens, for example. 

That also means making sure we don't create two screenshots in some weird situation.

bruno-garcia, line 66:
```suggestion
    public bool AttachScreenshot { get; set; }
```

The property name should be `AttachScreenshot`. As it's on the native iOS/Android.

Also please make sure to pass the value to the native layer in both platforms (prob MacCatalyst too) since that controls a screenshot if a native crash happens, for example. 

That also means making sure we don't create two screenshots in some weird situation.

src/Sentry.Maui/Internal/ScreenshotAttachment.cs
lucas-zimerman: interesting, it has the same behaviour as the old UWP

src/Sentry.Maui/Internal/ScreenshotAttachment.cs
```diff
@@ -0,0 +1,89 @@
...
+        {
+            try
+            {
+                if (Screenshot.Default.IsCaptureSupported)
+                {
+                    var screen = await Screenshot.Default.CaptureAsync().ConfigureAwait(true);
+
+                    var stream = await screen.OpenReadAsync(ScreenshotFormat.Jpeg).ConfigureAwait(true);
+
+#if __ANDROID__ && DEBUG
```
bruno-garcia, line 55: why ` && DEBUG`?. In CI we test release builds for example

```suggestion
#if __ANDROID__
```

src/Sentry.Maui/Internal/ScreenshotAttachment.cs
```diff
@@ -0,0 +1,89 @@
...
+                    Log.Info("Sentry", "Got screenshot stream with length: " + stream.Length);
```
bruno-garcia, line 56: Is this "logcat"? if so, doesn't our `options.Log` also go to logcat by default?

src/Sentry.Maui/Internal/ScreenshotAttachment.cs
```diff
@@ -0,0 +1,89 @@
...
+                    return stream;
+                }
+                else
+                {
+#if __ANDROID__ && DEBUG
+                    //On Android screen capture fails sometimes: https://github.com/dotnet/maui/issues/19450
+                    Log.Warn("Sentry", "Capturing screenshot not supported");
+#endif
+                    CaptureFailed = true;
+                    _options.LogWarning("Capturing screenshot not supported");
```
bruno-garcia, line 68: It's just a platform/support thing, not worthy of warning I'd say

```suggestion
                    _options.LogDebug("Capturing screenshot not supported");
```

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 ### Features
 
+- MAUI Screenshot support. You can opt-in via `SentryMauiOptions.AttachScreenshots` ([#2965](https://github.com/getsentry/sentry-dotnet/pull/2965))
```
bruno-garcia, line 7: need to verify which platforms it works:

```suggestion
- MAUI Screenshot support. You can opt-in via `SentryMauiOptions.AttachScreenshots` ([#2965](https://github.com/getsentry/sentry-dotnet/pull/2965))
   - Supports Android, iOS and Mac Catalyst only. Windows is not supported.
```

src/Sentry.Maui/Internal/ScreenshotAttachment.cs
```diff
@@ -0,0 +1,89 @@
...
+    public ScreenshotAttachmentContent(SentryMauiOptions options)
+    {
+        _options = options;
+    }
+
+    public Stream GetStream()
+    {
+        // Not including this on Windows specific build because on WinUI this can deadlock.
+#if !WINDOWS
+        var stream = MainThread.InvokeOnMainThreadAsync(async () =>
```
bruno-garcia, line 45: Check if on UI thread. If yes, do as u did originally. Blocking (with commets to source showing it's not really blocking).
This guarantees there's no yelding back and possibly redrawing the screen.

Now if we're on teh background thread, and on iOS only (that's when it fails?) then you do this approach u did here. And we document that, saying that in this case it could be that the screen changed from the time of the error.

src/Sentry.Maui/Internal/ScreenshotAttachment.cs
```diff
@@ -0,0 +1,89 @@
...
+
+    public ScreenshotAttachmentContent(SentryMauiOptions options)
+    {
+        _options = options;
+    }
+
+    public Stream GetStream()
+    {
+        // Not including this on Windows specific build because on WinUI this can deadlock.
+#if !WINDOWS
```
bruno-garcia, line 44: To support Windows (on a new PR, future) we can return a stream decorator and pass the `Task` as an argument. On ReadAsync we first await on that task.

This approach suffers from the same problem of doing async `InvokeOnMainThreadAsync` where it could change the UI (redraw) before we get the screenshot

src/Sentry.Maui/Internal/ScreenshotAttachment.cs
```diff
@@ -0,0 +1,89 @@
...
+                    CaptureFailed = false;
+                    return stream;
+                }
+                else
+                {
+#if __ANDROID__ && DEBUG
+                    //On Android screen capture fails sometimes: https://github.com/dotnet/maui/issues/19450
+                    Log.Warn("Sentry", "Capturing screenshot not supported");
+#endif
+                    CaptureFailed = true;
```
bruno-garcia, line 67: can we remove this hack and use `        var logger = new InMemoryDiagnosticLogger();` instead on tests and assert u logged an error (for that specific ticket)

src/Sentry.Maui/Internal/ScreenshotAttachment.cs
```diff
@@ -0,0 +1,84 @@
...
+#else
+            stream = MainThread.InvokeOnMainThreadAsync(async () =>
+            {
+                try
+                {
+                    var screen = await Screenshot.Default.CaptureAsync().ConfigureAwait(true);
+    
+                    return await screen.OpenReadAsync(ScreenshotFormat.Jpeg).ConfigureAwait(true);
+                }
+                //In some cases screen capture can throw, for example on Android if the activity is marked as secure.
```
bruno-garcia, line 72: > for example on Android

But this code doesn't run on Android (see ifdef above)

Also, we're not catching exceptions on Android/when on the UI thread

src/Sentry.Maui/SentryMauiOptions.cs
```diff
@@ -56,4 +56,12 @@ public SentryMauiOptions()
     /// Use caution when enabling, as depending on your application this might incur a performance overhead.
     /// </summary>
     public bool CreateElementEventsBreadcrumbs { get; set; } = false;
+
+    /// <summary>
+    /// Automatically attaches a screenshot of the app at the time of the event capture.
+    /// </summary>
+    /// <remarks>
+    /// Be aware PII might be included by activating this feature.
```
bruno-garcia, line 64:
```suggestion
    /// Make sure to only enable this feature if no sensitive data, such as PII, can be visible in the screen.
    /// Screenshots can be removed from some specific events during BeforeSend through the Hint.
```

test/Sentry.Maui.Tests/SentryMauiScreenshotTests.cs
bruno-garcia: Yes please add the whole test then, otherwise it's running half a test and not asserting anything, I don't see the goal

test/Sentry.Maui.Tests/SentryMauiScreenshotTests.cs
```diff
@@ -0,0 +1,97 @@
...
+        var options = app.Services.GetRequiredService<IOptions<SentryMauiOptions>>().Value;
+        
+        var envelope = _fixture.Transport.GetSentEnvelopes().FirstOrDefault(e => e.TryGetEventId() == sentryId);
+        envelope.Should().NotBeNull();
+
+        var envelopeItem = envelope!.Items.FirstOrDefault(item => item.TryGetType() == "attachment");
+
+        // Assert
+#if __MOBILE__
+        if (_fixture.Logger.Entries.Any(entry => entry.Level == SentryLevel.Error && entry.Exception is NullReferenceException))
```
bruno-garcia, line 58: This is ignoring that flaky issue on Android, right? so please be clearn:

```suggestion
        #if ANDROID
        // On Android this can fail due to XYZ, see issue https://github.com/maui/etc
        if (_fixture.Logger.Entries.Any(entry => entry.Level == SentryLevel.Error && entry.Exception is NullReferenceException))
```

src/Sentry.Maui/Internal/ScreenshotAttachment.cs
Giorgi: I removed try catch and let the caller log the exception.

src/Sentry.Maui/Internal/ScreenshotAttachment.cs
Giorgi: What we can also do on Windows is check if the app is in the foreground (what we already do in the other PR) and if so use PInvoke to grab a screenshot of that window. It will work for MAUI, Winforms, WPF, and probably other UI frameworks too.

src/Sentry.Maui/Internal/ScreenshotAttachment.cs
Giorgi: Where do I document the iOS behavior?

bruno-garcia: in Sentry docs, when you document the whole feature

src/Sentry.Maui/Internal/ScreenshotAttachment.cs
bruno-garcia: > PInvoke to grab a screenshot of that window. It will work for MAUI, Winforms, WPF, and probably other UI frameworks too.

I was hoping originally we'd have this feature work beyond MAUI with such approach. That means having the option on the core SentryOptions object and doing P/Invoke in other platforms but we would need to be sure we only get a screenshot of the app-window (not the whole desktop), can we do that?

bruno-garcia: Prob best to merge this PR first as-is and follow up with support on more platforms (create a ticket: Screenshot for Desktop apps for example)

bruno-garcia: Prob best to merge this PR first as-is and follow up with support on more platforms (create a ticket: Screenshot for Desktop apps for example)

Giorgi: On Windows we can do that. I'm not familiar with Linux/macOS apis for taking screenshot.

test/Sentry.Maui.Tests/SentryMauiScreenshotTests.cs
```diff
@@ -0,0 +1,98 @@
...
+            envelopeItem!.TryGetFileName().Should().Be("screenshot.jpg");
+        }
+    }
+#endif
+
+    [Fact]
+    public async Task CaptureException_RemoveScreenshot_NotContainsScreenshotAttachmentAsync()
+    {
+        // Arrange
+        var builder = _fixture.Builder.UseSentry(options => options.SetBeforeSend((e, hint) =>
```
bruno-garcia, line 75: This test doesn't have screenshot enabled, so would Hint even have a screeshot?

bruno-garcia (approved):

test/Sentry.Maui.Tests/SentryMauiScreenshotTests.cs
Giorgi: Screenshot is enabled in the `Fixture` setup.

PR #2968, opened 2023-12-13
.NET profiler improvements & test fixes

vaind:
* Make profiler startup non-blocking by default, with an option to block up to the given timeout.
* Tune tests to reduce flakiness
* null-check on TraceEvent thread()
* if option profiling sampling is on, but no integration was added or ios aot, diagnostic warn user telling them how to fix. 

#skip-changelog


src/Sentry.Profiling/SampleProfilerSession.cs
```diff
@@ -52,15 +53,43 @@ private SampleProfilerSession(SentryStopwatch stopwatch, EventPipeSession sessio
...
+            var session = client.StartEventPipeSession(Providers, requestRundown: false, CircularBufferMB);
 
-        // Process() blocks until the session is stopped so we need to run it on a separate thread.
-        Task.Factory.StartNew(eventSource.Process, TaskCreationOptions.LongRunning);
+            var stopWatch = SentryStopwatch.StartNew();
+            var eventSource = TraceLog.CreateFromEventPipeSession(session, TraceLog.EventPipeRundownConfiguration.Enable(client));
 
-        return new SampleProfilerSession(stopWatch, session, eventSource, logger);
+            // Process() blocks until the session is stopped so we need to run it on a separate thread.
+            Task.Factory.StartNew(eventSource.Process, TaskCreationOptions.LongRunning);
```
bruno-garcia, line 68: Do we need to observe any exception of this and log? I think right this it'll end up on UnobservedExceptionHandler

src/Sentry.Profiling/SamplingTransactionProfiler.cs
```diff
@@ -80,6 +80,7 @@ private void OnThreadSample(TraceEvent data)
             }
             else
             {
+                _options.LogDebug("Profiler received a sample past the end time; stopping.");
```
bruno-garcia, line 83: Is thsi something we should look into?
If so, could be a warn?

```suggestion
                _options.LogWarning("Profiler received a sample past the end time; stopping.");
```

src/Sentry.Profiling/SamplingTransactionProfilerFactory.cs
```diff
@@ -41,10 +45,24 @@ private SamplingTransactionProfilerFactory(SentryOptions options, SampleProfiler
...
+            if (!_sessionTask.IsCompleted)
+            {
+                _options.LogDebug("Cannot start a sampling profiler, the session hasn't started yet.");
+                _inProgress = FALSE;
+                return null;
+            }
+
+            if (!_sessionTask.IsCompletedSuccessfully)
+            {
+                _options.LogDebug("Cannot start a sampling profiler, the session startup has not been successful.");
```
bruno-garcia, line 57: level warn?

src/Sentry.Profiling/SamplingTransactionProfiler.cs
vaind: it's not unexpected, I'll reword

src/Sentry.Profiling/SampleProfilerSession.cs
```diff
@@ -52,15 +53,43 @@ private SampleProfilerSession(SentryStopwatch stopwatch, EventPipeSession sessio
...
+
+            return new SampleProfilerSession(stopWatch, session, eventSource, logger);
+        }
+        catch (Exception ex)
+        {
+            logger?.LogWarning("Error during sampler profiler EventPipeSession startup.", ex);
```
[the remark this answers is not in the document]
vaind: > Did we reorder the logger to take Exception as the first arg

It has been like this for a while:
https://github.com/getsentry/sentry-dotnet/blob/a9e8acd030efb232bb08ec3745c9ef9a479fca8d/src/Sentry/Extensibility/DiagnosticLoggerExtensions.cs#L167-L171

src/Sentry.Profiling/SamplingTransactionProfilerFactory.cs
```diff
@@ -60,6 +78,6 @@ private SamplingTransactionProfilerFactory(SentryOptions options, SampleProfiler
 
     public void Dispose()
     {
-        _session.Dispose();
+        _sessionTask.ContinueWith((_session) => _session.Dispose());
```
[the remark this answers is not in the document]
vaind: we don't really care when is it disposed so I wouldn't

src/Sentry.Profiling/ProfilingIntegration.cs
```diff
@@ -7,12 +7,32 @@ namespace Sentry.Profiling;
...
+    /// This is useful for applications that need to start quickly. The profiler will start in the background
+    /// and will be ready to capture transactions that have started after the profiler has started.
+    ///
+    /// If given a non-zero timeout, profiling startup blocks up to the given amount of time. If the timeout is reached
+    /// and the profiler session hasn't started yet, the execution is unblocked and behaves as the async startup,
+    /// i.e. transactions will be profiled only after the session is eventually started.
+    /// </param>
+    public ProfilingIntegration(TimeSpan startupTimeout = default)
+    {
+        Debug.Assert(TimeSpan.Zero == default);
```
jamescrosswell, line 26: Curious to know what the purpose of this line is... is that in case Microsoft ever change the default value for `TimeSpan` (which would implicitly break our functionality)?

src/Sentry.Profiling/SamplingTransactionProfilerFactory.cs
```diff
@@ -41,10 +45,24 @@ private SamplingTransactionProfilerFactory(SentryOptions options, SampleProfiler
...
         if (Interlocked.Exchange(ref _inProgress, TRUE) == FALSE)
         {
+            if (!_sessionTask.IsCompleted)
+            {
+                _options.LogWarning("Cannot start a sampling profiler, the session hasn't started yet.");
+                _inProgress = FALSE;
+                return null;
+            }
+
+            if (!_sessionTask.IsCompletedSuccessfully)
```
jamescrosswell, line 55: In these scenarios there's no profile associated with a transaction and so `TransactionProfiler?.Finish();` won't be called when the Transaction finishes right?
https://github.com/getsentry/sentry-dotnet/blob/175623c91c7c7a660ed3a5bafb6d4bd31ba86cb7/src/Sentry/TransactionTracer.cs#L368

What does happen to the profile though? Does it get cleaned up?

test/Sentry.Profiling.Tests/SamplingTransactionProfilerTests.cs
```diff
@@ -71,69 +73,83 @@ private SampleProfile CaptureAndValidate(ITransactionProfilerFactory factory)
...
+    public void Profiler_SingleProfile_Works(int startTimeoutSeconds)
+    {
+        using var factory = new SamplingTransactionProfilerFactory(_testSentryOptions, TimeSpan.FromSeconds(startTimeoutSeconds));
+        // in the async startup case, we need to wait before collecting
+        if (startTimeoutSeconds == 0)
         {
-            var frame = profile.Frames[profileToVerify.Stacks[0][i]];
-            frame.Module = frame.Module.Replace("System.Private.CoreLib.il", "System.Private.CoreLib");
-            profileToVerify.Frames.Add(frame);
+            factory._sessionTask.Wait(60_000);
```
jamescrosswell, line 111: Yikes... is this test flaky if we wait less than a minute for the _sessionTask?

src/Sentry.Profiling/SamplingTransactionProfilerFactory.cs
```diff
@@ -15,24 +15,28 @@ internal class SamplingTransactionProfilerFactory : IDisposable, ITransactionPro
...
-        _options = options;
-        _session = session;
+            // This can block indefinitely.
+            await session.WaitForFirstEventAsync().ConfigureAwait(false);
+
+            return session;
+        });
+
+        Debug.Assert(TimeSpan.FromSeconds(0) == TimeSpan.Zero);
+        if (startupTimeout != TimeSpan.Zero && !_sessionTask.Wait(startupTimeout))
```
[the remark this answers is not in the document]
vaind: Ok, added try-catch to `ProfilingIntegration`

src/Sentry.Profiling/ProfilingIntegration.cs
vaind: It communicates the assumptions made later in the code. For me, it wasn't 100 % clear what a default value for `TimeSpan` is so this clarifies it for anyone else too (and for me later when I forget :) )

src/Sentry.Profiling/SamplingTransactionProfilerFactory.cs
vaind: > What does happen to the profile though? Does it get cleaned up?

There's no profile being collected if we don't create an instance of `SamplingTransactionProfiler`. Yes, the session is running in the background but it drops all samples unless there's a profiler..

> I'm thinking particularly since some of the stuff that happens during initialization can "block indefinitely"...

That's OK, we just won't collect any profiles.

>Is this the answer?  ` // TODO cancellation token based on Hub being closed? `

Nope, that's another improvement we could do in case a transaction is left unfinished while a hub is being closed. In that case, we don't really care about continuing the profiler (even until up to the given 30 second timeout) and we could stop it immediately. This isn't handled by the hub at the moment, though.

test/Sentry.Profiling.Tests/SamplingTransactionProfilerTests.cs
vaind: That's a new code required due to an introduced change: session is now started asynchronously. So in case it is, we need to wait before actually trying to collect profiles. 60 seconds is as good as any other time,

src/Sentry.Profiling/SampleProfilerSession.cs
vaind: it's logged below on line 74

I've added a test case `Profiler_ThrowingOnSessionStartup_DoesntBreakSentryInit` to make sure an exception thrown inside profiler session init doesn't break Sentry init

test/Sentry.Profiling.Tests/SamplingTransactionProfilerTests.cs
vaind: And yet it can still fail to start up if the device is really overloaded... 
image

I don't really know what to do with this other than allowing a startup failure when running in CI in which case the test would get skipped...

src/Sentry.Profiling/SampleProfilerSession.cs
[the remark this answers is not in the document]
bruno-garcia: We had a PR to reorder this recently but I guess it didn't get to these extensions, maybe only whats hanging on SentryOptions

src/Sentry.Profiling/SampleProfilerSession.cs
```diff
@@ -50,17 +51,57 @@ private SampleProfilerSession(SentryStopwatch stopwatch, EventPipeSession sessio
 
     public TraceLog TraceLog => _eventSource.TraceLog;
 
+#if DEBUG
```
bruno-garcia, line 54: We run tests in CI in release builds

src/Sentry.Profiling/SampleProfilerSession.cs
bruno-garcia: Sorry I don't get it (line 74 is above this). But I mean if the callback running concurrently throws (eventSource.Process`, if that throws? We can `ContinuesWith` at the end of `StartNew` for example, or have a callback with a trycatch before we call eentSource.Process?

test/Sentry.Profiling.Tests/SamplingTransactionProfilerTests.cs
bruno-garcia: Waits up to a minute for one test and still times out? This is def far from ideal. Skipping in CI is OK but is this going to be flaky locally? 
Hard to believe it can take more than 1 minute to startup

test/Sentry.Profiling.Tests/SamplingTransactionProfilerTests.cs
```diff
@@ -231,6 +248,67 @@ async Task VerifyAsync(HttpRequestMessage message)
         }
     }
 
+    [SkippableFact]
+    public async Task Profiler_ThrowingOnSessionStartup_DoesntBreakSentryInit()
+    {
+#if DEBUG
+        SampleProfilerSession.ThrowOnNextStartupForTests = true;
+#else
+        Skip.If(true, "This test only works in DEBUG mode.");
```
bruno-garcia, line 257: Mentioned in a different comment but also means we don't run the test in CI

src/Sentry.Profiling/SampleProfilerSession.cs
vaind: Oh, I get what you mean now. I was confused at first because this bit of code hasn't actually changed so I thought you're talking about sth else.

In case `Process()` throws: yes, I think it'd get handled by UnobservedExceptionHandler.

I've added ContinueWith()

src/Sentry.Profiling/SampleProfilerSession.cs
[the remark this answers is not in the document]
vaind: https://github.com/getsentry/sentry-dotnet/issues/2986

test/Sentry.Profiling.Tests/SamplingTransactionProfilerTests.cs
vaind: changed to run all the time. the idea was to only have the code present when running in debug mode..

test/Sentry.Profiling.Tests/SamplingTransactionProfilerTests.cs
vaind: I've made the change to skip tests on particular failing calls. [`44fec5a` (#2968)](https://github.com/getsentry/sentry-dotnet/pull/2968/commits/44fec5a3488cfba6dd75abfe0fe6d57f19534e63)

src/Sentry.Profiling/SampleProfileBuilder.cs
```diff
@@ -173,15 +171,20 @@ private SentryStackFrame CreateStackFrame(CodeAddressIndex codeAddressIndex)
         }
         else
         {
-            // native frame
-            frame.InApp = false;
-        }
+            // Fall back if the method info is unknown, see more info on Symbol resolution in
+            // https://github.com/microsoft/perfview/blob/031250ffb4f9fcadb9263525d6c9f274be19ca51/src/PerfView/SupportFiles/UsersGuide.htm#L7745-L7784
```
bruno-garcia, line 175: Our fork needs to live for this product to live so safer to point to it:

```suggestion
            // https://github.com/getsentry/perfview/blob/031250ffb4f9fcadb9263525d6c9f274be19ca51/src/PerfView/SupportFiles/UsersGuide.htm#L7745-L7784
```

src/Sentry.Profiling/SampleProfilerSession.cs
```diff
@@ -50,17 +51,60 @@ private SampleProfilerSession(SentryStopwatch stopwatch, EventPipeSession sessio
...
+
     public static SampleProfilerSession StartNew(IDiagnosticLogger? logger = null)
     {
-        var client = new DiagnosticsClient(Process.GetCurrentProcess().Id);
-        var session = client.StartEventPipeSession(Providers, requestRundown: false, CircularBufferMB);
-        var stopWatch = SentryStopwatch.StartNew();
-        var eventSource = TraceLog.CreateFromEventPipeSession(session, TraceLog.EventPipeRundownConfiguration.Enable(client));
+        try
+        {
+            var client = new DiagnosticsClient(Process.GetCurrentProcess().Id);
```
bruno-garcia, line 60: There's in .NET 6 and newer a new API for this that's cheaper. it's used here:

https://github.com/getsentry/sentry-dotnet/blob/725c830167d84b1bab38e349df3e8aa3339fdd7e/src/Sentry/Internal/ProcessInfo.cs#L29C9-L35

Could you rename that to `GetProcessId` and use this:
```suggestion
            var procId = _options.ProcessInfo.GetProcessId();
            var client = new DiagnosticsClient(procId);
```

src/Sentry.Profiling/SampleProfilerSession.cs
```diff
@@ -50,17 +51,60 @@ private SampleProfilerSession(SentryStopwatch stopwatch, EventPipeSession sessio
...
+            var stopWatch = SentryStopwatch.StartNew();
+            var eventSource = TraceLog.CreateFromEventPipeSession(session, TraceLog.EventPipeRundownConfiguration.Enable(client));
+
+            // Process() blocks until the session is stopped so we need to run it on a separate thread.
+            Task.Factory.StartNew(eventSource.Process, TaskCreationOptions.LongRunning)
+                .ContinueWith(_ =>
+                {
+                    if (_.Exception?.InnerException is { } e)
+                    {
+                        logger?.LogWarning("Error during sampler profiler EventPipeSession processing.", e);
```
bruno-garcia, line 81: Will conflict now I imagine, since we merged ur PR reordering

test/Sentry.Profiling.Tests/SamplingTransactionProfilerTests.cs
```diff
@@ -1,16 +1,42 @@
...
+    {
+        if (TestEnvironment.IsGitHubActions)
+        {
+            try
+            {
+                checks.Invoke();
+            }
+            catch (Exception e)
+            {
+                _testOutputLogger.LogWarning("Caught an exception in a test block that is allowed to fail when in CI.", e);
```
bruno-garcia, line 30: Do we know what this exception will be? That would allow us to validate this in CI (timeout is OK but else it's not?)

src/Sentry.Profiling/SampleProfilerSession.cs
vaind: Actually, I should have looked for usages on the other PR too... All the usages passing an exception got picked up by the generic method so they didn't fail. I've fixed it in this branch with hope it gets merged soon also those fixes are unrelated. If it doesn't get merged, I can cherry-pick the commit to a new PR.

src/Sentry.Profiling/SampleProfilerSession.cs
vaind: since we target only net6 and greater in this project I'm going to use `Environment.ProcessId` directly.

bruno-garcia (approved):

Issue #2978, opened 2023-12-18
Send frame `filename` including project root


bruno-garcia:
Source link doesn't work on .NET right now. We send `abs_path` with the full path to root, plus a `filename` that has only literally the name, without path.
 
* https://github.com/getsentry/sentry/issues/61730

But here:

https://github.com/getsentry/sentry-dotnet/blob/2e00a5c5509744a3f4fdb830952cb5aa176f3d0c/src/Sentry/buildTransitive/Sentry.targets

We create at build time an attribute `Sentry.ProjectDirectory` that contains the path to the project root. 
With that we can, truncate off the root path that's irrelevant (build server paths etc).

This can help unblocking **DDM code locations** but also Source Link without the fix for `abs_path` (that hopefully happens so all these years of SDK shipped also work). 

## References
 
* See [Frame Attributes](https://develop.sentry.dev/sdk/event-payloads/stacktrace/#frame-attributes)

jamescrosswell:
I played around with this briefly. I don't think the `Sentry.ProjectDirectory` will work for us... what we really need is the `RepoRoot`.

See attached PR.

bitsandfoxes:
Closed by product! It works like a charm now!

PR #2980, opened 2023-12-19
Repository Relative Filenames in Stack Frames

jamescrosswell:
Resolves to https://github.com/getsentry/sentry-dotnet/issues/2978

The idea is to strip everything above the repository root from the absolute path that we report in stack traces... so if someone has a git repo for their project in `/Users/Bobby/code/myPetProject/` the filename paths would be reported as:
- `/Users/Bobby/code/myPetProject/readme.md` -> `readme.md`
- `/Users/Bobby/code/myPetProject/src/Program.cs` -> `src/Program.cs`

Combining that with a remote URL (via source link) and a commit hash, we'd be able to take people directly to the source code.

### DIFF
--- a/src/Sentry/Internal/DebugStackTrace.cs
+++ b/src/Sentry/Internal/DebugStackTrace.cs
@@ -349,7 +349,7 @@ internal class DebugStackTrace : SentryStackTrace
             if (AttributeReader.TryGetProjectDirectory(method.Module.Assembly) is { } projectPath
                 && frameFileName.StartsWith(projectPath, StringComparison.OrdinalIgnoreCase))
             {
-                frame.AbsolutePath = GetRepositoryRelativePath(method.Module.Assembly, frameFileName);
+                frame.AbsolutePath = GetRepositoryRelativePath(method.Module.Assembly, frameFileName, _options);
                 frameFileName = frameFileName[projectPath.Length..];
             }
             frame.FileName = frameFileName;
@@ -358,15 +358,21 @@ internal class DebugStackTrace : SentryStackTrace
         return frame;
     }
 
-    internal static string GetRepositoryRelativePath(Assembly moduleAssembly, string frameFileName)
+    internal static string GetRepositoryRelativePath(Assembly moduleAssembly, string frameFileName,
+        SentryOptions options)
     {
-        if (AttributeReader.TryGetProjectDirectory(moduleAssembly) is { } projectDirectory && frameFileName.StartsWith(projectDirectory))
+        if (AttributeReader.TryGetRepositoryRoot(moduleAssembly) is not { } repoRoot ||
+            !frameFileName.StartsWith(repoRoot))
         {
-            frameFileName = frameFileName.Substring(projectDirectory.Length);
-            if (frameFileName.StartsWith("/"))
-            {
-                frameFileName = frameFileName.Substring(1);
-            }
+            options.LogDebug("No repository root found. Using absolute file path.");
+            return frameFileName;
+        }
+
+        options.LogDebug($"Filenames relative to repository root: {repoRoot}");
+        frameFileName = frameFileName.Substring(repoRoot.Length);
+        if (frameFileName.StartsWith("/"))
+        {
+            frameFileName = frameFileName.Substring(1);
         }
         return frameFileName;
     }
--- a/src/Sentry/Internal/DebugStackTrace.cs
+++ b/src/Sentry/Internal/DebugStackTrace.cs
@@ -346,11 +346,13 @@ internal class DebugStackTrace : SentryStackTrace
 
         if (stackFrame.GetFileName() is { } frameFileName)
         {
-            if (AttributeReader.TryGetProjectDirectory(method.Module.Assembly) is { } projectPath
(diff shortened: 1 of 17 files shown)


src/Sentry/Internal/DebugStackTrace.cs
```diff
@@ -358,6 +358,25 @@ internal static SentryStackFrame ParseNativeAOTToString(string info)
         return frame;
     }
 
+    internal static string GetRepositoryRelativePath(Assembly moduleAssembly, string frameFileName,
+        SentryOptions options)
+    {
+        if (AttributeReader.TryGetRepositoryRoot(moduleAssembly) is not { } repoRoot ||
+            !frameFileName.StartsWith(repoRoot))
```
jamescrosswell, line 365: This might be problematic if the code was built on a Windows machine (with `\` path separators) and run on a *nix variant (with `/` path separators). Let's make this more robust.

test/Sentry.Tests/Internals/SentryStackTraceFactoryTests.cs
```diff
@@ -191,7 +191,7 @@ public void FileNameShouldBeRelative()
         var path = Path.Combine("Internals", "SentryStackTraceFactoryTests.cs");
         Assert.Equal(path, frame.FileName);
 
-        var fullPath = GetThisFilePath();
+        var fullPath = "test/Sentry.Tests/Internals/SentryStackTraceFactoryTests.cs";
```
jamescrosswell, line 194: @vaind could this have any side effects? I think this test was more about the `frame.FileName` property above (making sure this was a combination of Internals + filename) but I'm not sure why that's important or what the intent was here.

src/Sentry/Internal/DebugStackTrace.cs
```diff
@@ -346,18 +346,28 @@ internal static SentryStackFrame ParseNativeAOTToString(string info)
 
         if (stackFrame.GetFileName() is { } frameFileName)
         {
-            if (AttributeReader.TryGetProjectDirectory(method.Module.Assembly) is { } projectPath
-                && frameFileName.StartsWith(projectPath, StringComparison.OrdinalIgnoreCase))
+            var assembly = method.Module.Assembly;
+            var root = TryGetRoot(assembly, frameFileName, AttributeReader.TryGetRepositoryRoot)
+                ?? TryGetRoot(assembly, frameFileName, AttributeReader.TryGetProjectDirectory);
```
jamescrosswell, line 351: @vaind I'm not sure if this is a good idea or not. My thinking was we'd try to resolve a git repository relative path and, if we couldn't do that, we'd try a project relative path (the later being "more specific").

A repo relative path would be useful for source link and code locations (in the delightful developer metrics stuff). 

What was the motivation for the original code that you wrote though, using a project relative path for the `frame.FileName`? Is what I've done here going to break anything there?

src/Sentry/buildTransitive/Sentry.targets
```diff
@@ -12,12 +12,17 @@
           Outputs="$(IntermediateOutputPath)$(SentryAttributesFile)">
     <PropertyGroup>
       <SentryAttributesFilePath>$(IntermediateOutputPath)$(SentryAttributesFile)</SentryAttributesFilePath>
+      <RepoRoot>$([System.IO.Path]::GetDirectoryName($([MSBuild]::GetPathOfFileAbove('.gitignore', '$(ProjectDir)'))))</RepoRoot>
```
bruno-garcia, line 15: any folder can have `.gitignore` and in monorepos this can be more common.
How about calling this `GitIgnorePath` instead?

bruno-garcia:
@jamescrosswell we might be able to drop this PR, @malwilley mentioned he's planning on adding support based on `abs_path` which will unblock this for all current .NET (and cocoa I believe) customers without the need to update SDK 🙏

vaind:
> @jamescrosswell we might be able to drop this PR, @malwilley mentioned he's planning on adding support based on `abs_path` which will unblock this for all current .NET (and cocoa I believe) customers without the need to update SDK 🙏

Overall this looks a bit more problematic and I'd need to probably look a bit deeper before being able to do a proper PR so let me know if we should go ahead with it or it's getting closed.

jamescrosswell:
All good. @vaind it would still be good to understand the original motivation for having project relative paths for the filename (just for future context).

jamescrosswell:
> @jamescrosswell we might be able to drop this PR, @malwilley mentioned he's planning on adding support based on `abs_path` which will unblock this for all current .NET (and cocoa I believe) customers without the need to update SDK 🙏

Is there an issue or something that we can track for that on our original issue (instead of this PR)?

bruno-garcia:
> > @jamescrosswell we might be able to drop this PR, @malwilley mentioned he's planning on adding support based on `abs_path` which will unblock this for all current .NET (and cocoa I believe) customers without the need to update SDK 🙏
> 
> Is there an issue or something that we can track for that on our original issue (instead of this PR)?

yes:
* https://github.com/getsentry/sentry/issues/61730

bruno-garcia:
Lets close this for now, since the solution should be fallign back to `abs_path`

Issue #2938, opened 2023-12-02
Set ip_address to {{auto}} by default, even if sendDefaultPII is disabled


bruno-garcia:
Similar to: 

* https://github.com/getsentry/sentry-java/pull/2860

This is the behavior of the JavaScript and Apple (Cocoa) SDK, and now of the Android SDK (v7) which we already merged into `main` and shipped on 4.0.0-beta.3

bruno-garcia:
@bitsandfoxes there was a thread on Slack confirming this. It's a small change code-wise but large in impact, we need to be clear on the changelog. cc @Giorgi

smino-dev:
Hey Sentry, we came across this change and are a bit worried about this. Our software needs to be GDPR compliant and for now set sendDefaultPII always to false, because we do not want to send the IP to sentry. How do you filter out those IP Addresses now if the option is on sentry.io? Does the IP get sent to sentry and it gets then discarded there?

Or how can we achieve to not sent the IP Address to sentry at all? Do we have to set sendDefaultPII to true and create our own `SentryUserFactory` where set the IP to `null`?

jamescrosswell:
> How do you filter out those IP Addresses now if the option is on sentry.io?

Hey @smino-dev, 

There's a description of this [in the 4.0.0 changelog](https://github.com/getsentry/sentry-dotnet/blob/c8330f10afd241171d0657e91e9b0cfdf22b1b2c/CHANGELOG.md#significant-change-in-behavior:~:text=The%20User.IpAddress,control%20this%20instead)

> The "Prevent Storing of IP Addresses" option in the "Security & Privacy" project settings on sentry.io can be used to control this instead

jamescrosswell:
> On our API or somewhere on a sentry server?

That I'm not sure of... maybe best to ask on [Discord](https://discord.com/channels/621778831602221064/621778831602221072) (referencing this chat here).

bruno-garcia:
> or exception is sent to sentry, where does this IP Address gets filtered out? 

`{{auto}}` on the event instructs Relay to use the IP address of the connection.

PR #2981, opened 2023-12-20
Default User.IpAddress to {{auto}}

jamescrosswell:
Resolves https://github.com/getsentry/sentry-dotnet/issues/2938

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,11 @@
 
 ## Unreleased
 
+### API breaking Changes
+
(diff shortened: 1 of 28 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,11 @@
 
 ## Unreleased
 
+### API breaking Changes
```
bruno-garcia, line 5: How about a different section like this: (since it's not an API change);

```suggestion
### Significant change in behavior
```

bruno-garcia (approved):

Issue #2945, opened 2023-12-05
Add In Foreground to App context


bruno-garcia:
Relevant for all UI-based apps (WPF, WInForms, MAUI, Xamarin, etc)

See: 

* https://github.com/getsentry/team-mobile/issues/36
* https://github.com/getsentry/sentry/issues/44251

Same as:
* https://github.com/getsentry/sentry-dart/pull/1260

bruno-garcia:
@Giorgi:

>  In the Add In Foreground to App context ticket https://github.com/getsentry/sentry-dotnet/issues/2945  in case of Windows apps, does in foreground mean if the app has focus? Or only check if the app is in foreground?


hm, how about in that case we don't set `in_foreground` at all, and instead set `in_focus` ? We can document the difference, and if searching and you care about either you can `in_foreground`=true && `is_focus`=true?

Giorgi:
We can do that but I think from a Windows developer perspective the naming is a bit confusing. In Windows, foreground window is the windows that has focus: [GetForegroundWindow](https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-getforegroundwindow)

bitsandfoxes:
I assume the the definition `if it's visible it's in the foreground` comes from the mobile perspective. In our case, I would use `in_focus` and `in_foreground` synonymously and stick with `in_foreground`.

What we care about with this is: Was the user actively interacting with the application vs was it just floating around somewhere/just being visible on the monitor?

bruno-garcia:
> I assume the the definition if it's visible it's in the foreground comes from the mobile perspective. In our case, I would use in_focus and in_foreground synonymously and stick with in_foreground.

works for me, just worth being clear on the docs about all these terminologies

PR #2983, opened 2023-12-20
Set in_foreground app context

Giorgi:
Resolves: https://github.com/getsentry/sentry-dotnet/issues/2945

### DIFF
--- a/src/Sentry.Maui/SentryMauiAppBuilderExtensions.cs
+++ b/src/Sentry.Maui/SentryMauiAppBuilderExtensions.cs
@@ -99,13 +99,26 @@ public static class SentryMauiAppBuilderExtensions
 
                     var platformApplication = application.Delegate as IPlatformApplication;
                     platformApplication?.HandleMauiEvents(bind: false);
+
+                    SentryMauiEventProcessor.InForeground = false;
                 });
+
+                lifecycle.OnActivated(application => SentryMauiEventProcessor.InForeground = true);
+
+                lifecycle.DidEnterBackground(application => SentryMauiEventProcessor.InForeground = false);
+                lifecycle.OnResignActivation(application => SentryMauiEventProcessor.InForeground = false);
             });
 #elif ANDROID
             events.AddAndroid(lifecycle =>
             {
                 lifecycle.OnApplicationCreating(application => (application as IPlatformApplication)?.HandleMauiEvents());
                 lifecycle.OnDestroy(application => (application as IPlatformApplication)?.HandleMauiEvents(bind: false));
+
+                lifecycle.OnResume(activity => SentryMauiEventProcessor.InForeground = true);
+                lifecycle.OnStart(activity => SentryMauiEventProcessor.InForeground = true);
+
+                lifecycle.OnStop(activity => SentryMauiEventProcessor.InForeground = false);
+                lifecycle.OnPause(activity => SentryMauiEventProcessor.InForeground = false);
             });
 #elif WINDOWS
             events.AddWindows(lifecycle =>
--- a/src/Sentry.Maui/SentryMauiAppBuilderExtensions.cs
+++ b/src/Sentry.Maui/SentryMauiAppBuilderExtensions.cs
@@ -100,6 +100,8 @@ public static class SentryMauiAppBuilderExtensions
                     var platformApplication = application.Delegate as IPlatformApplication;
                     platformApplication?.HandleMauiEvents(bind: false);
 
+                    //According to https://developer.apple.com/documentation/uikit/uiapplicationdelegate/1623111-applicationwillterminate#discussion
+                    //WillTerminate is called: in situations where the app is running in the background (not suspended) and the system needs to terminate it for some reason. 
                     SentryMauiEventProcessor.InForeground = false;
                 });
 
--- a/src/Sentry/Internal/Enricher.cs
+++ b/src/Sentry/Internal/Enricher.cs
@@ -85,19 +85,8 @@ internal class Enricher
         //Apply App startup and Boot time
         eventLike.Contexts.App.StartTime ??= ProcessInfo.Instance?.StartupTime;
         eventLike.Contexts.Device.BootTime ??= ProcessInfo.Instance?.BootTime;
-
-        try
-        {
-            if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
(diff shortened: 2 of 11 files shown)


src/Sentry/Internal/ProcessInfo.cs
```diff
@@ -25,6 +26,24 @@ internal Task PreciseAppStartupTask
...
+#if NET6_0_OR_GREATER
+        return Environment.ProcessId;
+#else
+        try
+        {
+            return Process.GetCurrentProcess().Id;
+        }
+        catch (Exception)
+        {
+            return null;
```
bruno-garcia, line 42: we should options.LogWarning(ex, "..") here
We'd want to know what went wrong

src/Sentry/Internal/ProcessInfo.cs
```diff
@@ -25,6 +26,24 @@ internal Task PreciseAppStartupTask
         private set => _preciseAppStartupTask = value;
     }
 
+    public int? Id => _id ??= GetCurrentProcessId();
```
bruno-garcia, line 29: This code path can end up on the P/Invoke below expecting `user32.dll`and will crash outside Windows.
We need to have runtime OS check before running that stuff.

Also, CI is green which indicates to me we don't have code calling this, need some tests.

src/Sentry/Internal/ProcessInfo.cs
```diff
@@ -100,4 +119,32 @@ private static DateTimeOffset GetStartupTime()
         return proc.StartTime.ToUniversalTime();
     }
 #endif
+
+    public bool? ApplicationIsActivated(SentryOptions options)
+    {
+        try
+        {
+            var activatedHandle = GetForegroundWindow();
```
bruno-garcia, line 127: Will crash on mac/linux, right?

src/Sentry/Internal/Enricher.cs
```diff
@@ -86,6 +86,19 @@ public void Apply(IEventLike eventLike)
...
+        try
+        {
+            if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
+            {
+                eventLike.Contexts.App.InForeground = ProcessInfo.Instance?.ApplicationIsActivated(_options);
+            }
+        }
+        catch
+        {
+            //If RuntimeInformation.IsOSPlatform(OSPlatform.Windows) throws we are not on Windows.
```
bruno-garcia, line 98: but there's a guard, so why would it throw?

Please catch the exception and log, lets not make assumptions and hide it,makes it hard to debug later

src/Sentry.Maui/SentryMauiAppBuilderExtensions.cs
```diff
@@ -99,13 +99,26 @@ private static void RegisterMauiEventsBinder(this MauiAppBuilder builder)
 
                     var platformApplication = application.Delegate as IPlatformApplication;
                     platformApplication?.HandleMauiEvents(bind: false);
+
+                    SentryMauiEventProcessor.InForeground = false;
```
bruno-garcia, line 103: Is it true that at this point it's not on the foreground anymore?
If it crashes right after, would be correct to say it wasn't on the foreground?

src/Sentry.Maui/SentryMauiAppBuilderExtensions.cs
```diff
@@ -99,13 +99,26 @@ private static void RegisterMauiEventsBinder(this MauiAppBuilder builder)
...
+                lifecycle.DidEnterBackground(application => SentryMauiEventProcessor.InForeground = false);
+                lifecycle.OnResignActivation(application => SentryMauiEventProcessor.InForeground = false);
             });
 #elif ANDROID
             events.AddAndroid(lifecycle =>
             {
                 lifecycle.OnApplicationCreating(application => (application as IPlatformApplication)?.HandleMauiEvents());
                 lifecycle.OnDestroy(application => (application as IPlatformApplication)?.HandleMauiEvents(bind: false));
+
+                lifecycle.OnResume(activity => SentryMauiEventProcessor.InForeground = true);
```
bruno-garcia, line 117: Shouldn't MAUI deal with firing these evens on the right platforms or not? I expect this to work on iOS/MacCatalyst too

bruno-garcia (commented):
It's missing device tests or some way to verify that if the events fire, the flags flip, on the right platforms

src/Sentry.Maui/SentryMauiAppBuilderExtensions.cs
Giorgi: MAUI fires lifecycle events for each platform. iOS callbacks are also handled here and this works on iOS too. Docs don't say anything about MacCatalyst: https://learn.microsoft.com/en-us/dotnet/maui/fundamentals/app-lifecycle?view=net-maui-8.0#cross-platform-lifecycle-events

src/Sentry.Maui/SentryMauiAppBuilderExtensions.cs
Giorgi: From Apple docs: 

> For apps that support background execution, this method is generally not called when the user quits the app because the app simply moves to the background in that case. However, this method may be called in situations where the app is running in the background (not suspended) and the system needs to terminate it for some reason.

src/Sentry/Internal/Enricher.cs
Giorgi: On lime 42 in the same file, it says that: _RuntimeInformation.* throws on .NET Framework on macOS/Linux_

src/Sentry/Internal/ProcessInfo.cs
Giorgi: Yes, but it is called only on Windows but I'll move OS check inside this method.

src/Sentry/Internal/ProcessInfo.cs
Giorgi: `GetCurrentProcessId` doesn't use PInvoke I believe.

src/Sentry.Maui/SentryMauiAppBuilderExtensions.cs
bruno-garcia: worth adding some links to those docs from the code

bruno-garcia (approved):

PR #2991, opened 2023-12-22
fix: remove aot warn

bruno-garcia:
This will stop the warning but raises the question: Why do we have a const? 
Might `#ifdef` everywhere NET8 or newer at the callsite?

#skip-changelog

### DIFF
--- a/src/Sentry/Internal/DebugStackTrace.cs
+++ b/src/Sentry/Internal/DebugStackTrace.cs
@@ -499,6 +499,7 @@ internal class DebugStackTrace : SentryStackTrace
     {
 #if !NET8_0_OR_GREATER
 #pragma warning disable 0162
+#endif
         if (AotHelper.IsNativeAot)
         {
             // Only unreachable outside NET8_0_OR_GREATER
@@ -506,6 +507,7 @@ internal class DebugStackTrace : SentryStackTrace
             assemblyName = null;
             return null;
         }
+#if !NET8_0_OR_GREATER
 #pragma warning restore 0162
 #endif
         assemblyName = module.FullyQualifiedName;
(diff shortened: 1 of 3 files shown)


src/Sentry/Internal/StackFrame.cs
```diff
@@ -114,7 +114,15 @@ public RealStackFrame(StackFrame frame)
...
+#if !NET8_0_OR_GREATER
+#pragma warning disable CS0162 // Unreachable code detected
+#endif
+        // Only unreachable outside NET8_0_OR_GREATER
+        // ReSharper disable once HeuristicUnreachableCode
         ? null
+#if !NET8_0_OR_GREATER
+#pragma warning restore CS0162 // Unreachable code detected
+#endif
         : _frame.GetMethod();
```
vaind, line 126: How about the following? It's a bit more readable to me:

```suggestion
    public MethodBase? GetMethod() 
#if NET8_0_OR_GREATER
				=> AotHelper.IsNativeAot ? null : _frame.GetMethod();
#else
				=> _frame.GetMethod();
#endif
```

src/Sentry/Internal/DebugStackTrace.cs
```diff
@@ -497,14 +497,19 @@ private static void DemangleLambdaReturnType(SentryStackFrame frame)
     [UnconditionalSuppressMessage("SingleFile", "IL3002:Avoid calling members marked with 'RequiresAssemblyFilesAttribute' when publishing as a single-file", Justification = AotHelper.SuppressionJustification)]
     private static PEReader? TryReadAssemblyFromDisk(Module module, SentryOptions options, out string? assemblyName)
     {
+#if !NET8_0_OR_GREATER
+#pragma warning disable 0162
```
vaind, line 501: Where do you get these warnings? Also, I'm not a fan of having negative one line ifdefs like this, makes the code quite hard to read.

vaind (commented):
> Why do we have a const?

to completely remove the code on build it doesn't apply to.

src/Sentry/Internal/DebugStackTrace.cs
bruno-garcia: Rider gets all red without this for me, since statically knows that `null` is impossible. Including ( believe, didn't check yet) on ur suggestion above

vaind: but doesn't it respect `#pragma warning disable CS0162`?

vaind: looks like it should since 2023.1 
https://youtrack.jetbrains.com/issue/RSRP-490648

vaind: I've tested this myself and created an alternative PR: #2994

Issue #2998, opened 2024-01-02
OpenTelemetry causes sentry transactions loop


kerenkhatiwada:
### Package

Sentry

### .NET Flavor

.NET

### .NET Version

7.0

### OS

Browser

### SDK Version

3.35.0
Issue also occurs in 4.0.0-beta.6

### Steps to Reproduce

Use `OpenTelemetry.Instrumentation.Http`, `OpenTelemetry.Instrumentation.AspNetCore`, and `OpenTelemetry.Extensions.Hosting` version `1.7.0`

Observe an opentelemetry span become a transaction,
that transaction creates a POST request to Sentry with the event,
that POST request adds a new span in opentelemetry,
https://org2-w7.sentry.io/performance/dotnet-aspnetcore-mywebapp:cc08e21292b4408cabb04a1cc41bb292/?project=4504492492390400&query=&referrer=performance-transaction-summary&statsPeriod=1h&transaction=POST&unselectedSeries=p100%28%29&unselectedSeries=avg%28%29
![Screenshot 2024-01-02 at 4 16 14 PM](https://github.com/getsentry/sentry-dotnet/assets/99350819/45255484-23ca-469c-8690-adc36d0b97bb)

that new span becomes a new Sentry transaction,
repeat

https://org2-w7.sentry.io/discover/homepage/?end=2024-01-02T20%3A50%3A59&field=title&field=event.type&field=project&field=user.display&field=timestamp&field=replayId&name=All+Events&project=4504492492390400&query=transaction%3APOST&sort=-timestamp&start=2024-01-02T20%3A26%3A00&yAxis=count%28%29

https://github.com/getsentry/sentry-dotnet/assets/99350819/904711ae-fb92-497e-80f6-541e9a097b77

### Expected Result

The POST requests made to Sentry are not reported

### Actual Result

Sentry recursively generates events for its calls

bitsandfoxes:
Hey, thanks for reaching out. The SDK should most definitely not instrument itself calling home. We'll be fixing that ASAP.

knd775:
> Hey, thanks for reaching out. The SDK should most definitely not instrument itself calling home. We'll be fixing that ASAP.

Thanks for fixing it so quickly, this was a fun bug 😅 
![image](https://github.com/getsentry/sentry-dotnet/assets/1235055/ce3ff419-eab8-4784-9935-0edfb6d1b62c)

bitsandfoxes:
Ooof.. Sorry about that 😓

PR #3001, opened 2024-01-03
fix: OTel spans for Sentry requests

bitsandfoxes:
Fixes https://github.com/getsentry/sentry-dotnet/issues/2998

Looks like the key changed.

### DIFF
--- a/src/Sentry.OpenTelemetry/SentrySpanProcessor.cs
+++ b/src/Sentry.OpenTelemetry/SentrySpanProcessor.cs
@@ -113,7 +113,12 @@ public class SentrySpanProcessor : BaseProcessor<Activity>
         // Make a dictionary of the attributes (aka "tags") for faster lookup when used throughout the processor.
         var attributes = data.TagObjects.ToDict();
 
-        if (attributes.TryGetTypedValue("url.full", out string? url) && (_options?.IsSentryRequest(url) ?? false))
+        var url =
+            attributes.TryGetTypedValue(OtelSemanticConventions.AttributeUrlFull, out string? tempUrl) ? tempUrl
+            : attributes.TryGetTypedValue("http.url", out string? fallbackUrl) ? fallbackUrl    // Falling back to a convention from pre-1.5.0
+            : null;
+
+        if (!string.IsNullOrEmpty(url) && (_options?.IsSentryRequest(url) ?? false))
         {
             _options?.DiagnosticLogger?.LogDebug($"Ignoring Activity {data.SpanId} for Sentry request.");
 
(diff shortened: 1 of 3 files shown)


src/Sentry.OpenTelemetry/SentrySpanProcessor.cs
```diff
@@ -113,7 +113,7 @@ public override void OnEnd(Activity data)
         // Make a dictionary of the attributes (aka "tags") for faster lookup when used throughout the processor.
         var attributes = data.TagObjects.ToDict();
 
-        if (attributes.TryGetTypedValue("http.url", out string? url) && (_options?.IsSentryRequest(url) ?? false))
+        if (attributes.TryGetTypedValue("url.full", out string? url) && (_options?.IsSentryRequest(url) ?? false))
```
bruno-garcia, line 116: this approach to detect if it's a Sentry request seems rather fragile

bruno-garcia (approved):

src/Sentry.OpenTelemetry/SentrySpanProcessor.cs
jamescrosswell, line 116: We should probably use the consts defined in:
https://github.com/getsentry/sentry-dotnet/blob/bf4e0b3db4df28447d055cb569093c4307e85cf7/src/Sentry/Internal/OpenTelemetry/SemanticConventions.cs#L128

As you can see from the comments, however, the key will depend on the version of OTel that was used to generate the span... and I'm not sure if we have any control over that. So we might need to check for both the old and the new value. 

I think it was changed in v1.5.1-beta.1 of OpenTelemetry and we take a dependency on 1.5.0 at the moment:
https://github.com/getsentry/sentry-dotnet/blob/dd9cfac482c7759f9742db495e2e3750967449f5/src/Sentry.OpenTelemetry/Sentry.OpenTelemetry.csproj#L18-L20

jamescrosswell (changes requested):

src/Sentry.OpenTelemetry/SentrySpanProcessor.cs
jamescrosswell: Other ideas?

jamescrosswell (changes requested):
Sorry, should have put this in the original review... might be good to change this test from a Fact to a Theory and test it with both the old and the new:
https://github.com/getsentry/sentry-dotnet/blob/41714ae283854a31955cb31fb5b68326806cd075/test/Sentry.OpenTelemetry.Tests/SentrySpanProcessorTests.cs#L381-L382

At the moment, it only tests the old attributes.

jamescrosswell (approved):

src/Sentry.OpenTelemetry/SentrySpanProcessor.cs
bruno-garcia: AsyncLocal?

PR #3017, opened 2024-01-08
chore: make use of the new NSMutableDictionary return type in iOS profiler

vaind:
Follow up on changes from the latest Cocoa SDK update.

#skip-changelog

### DIFF
--- a/src/Sentry/Platforms/Cocoa/CocoaProfiler.cs
+++ b/src/Sentry/Platforms/Cocoa/CocoaProfiler.cs
@@ -36,8 +36,7 @@ internal class CocoaProfiler : ITransactionProfiler
 
     public ISerializable? Collect(Transaction transaction)
     {
-        // TODO change return type of CocoaSDKs CollectProfileBetween to NSMutableDictionary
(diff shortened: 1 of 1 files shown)


src/Sentry/Platforms/Cocoa/CocoaProfiler.cs
```diff
@@ -36,8 +36,7 @@ public void Finish()
 
     public ISerializable? Collect(Transaction transaction)
     {
-        // TODO change return type of CocoaSDKs CollectProfileBetween to NSMutableDictionary
-        var payload = SentryCocoaHybridSdk.CollectProfileBetween(_startTimeNs, _endTimeNs, _cocoaTraceId)?.MutableCopy() as NSMutableDictionary;
+        var payload = SentryCocoaHybridSdk.CollectProfileBetween(_startTimeNs, _endTimeNs, _cocoaTraceId) as NSMutableDictionary;
```
jamescrosswell, line 39: Is the purpose of this PR just to remove the TODO/comment?

vaind: and `?.MutableCopy()`

vaind: and actually, the cast is also unnecessary, I think.
removing now, let's see what the CI says

Issue #3033, opened 2024-01-10
Flaky `MetricAggregatorTests.Distribution_AggregatesMetrics`


vaind:
See https://github.com/getsentry/sentry-dotnet/actions/runs/7471895579?pr=3026

```
System.Collections.Generic.KeyNotFoundException : Arg_KeyNotFoundWithKey, 40
at System.Collections.Concurrent.ConcurrentDictionary`2[[System.Int64, System.Private.CoreLib, Version=7.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e],[System.Collections.Concurrent.ConcurrentDictionary`2[[System.String, System.Private.CoreLib, Version=7.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e],[Sentry.Protocol.Metrics.Metric, Sentry, Version=4.0.0.0, Culture=neutral, PublicKeyToken=fba2ec45388e2af0]], System.Collections.Concurrent, Version=7.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a]].ThrowKeyNotFoundException(Int64 )
   at System.Collections.Concurrent.ConcurrentDictionary`2[[System.Int64, System.Private.CoreLib, Version=7.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e],[System.Collections.Concurrent.ConcurrentDictionary`2[[System.String, System.Private.CoreLib, Version=7.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e],[Sentry.Protocol.Metrics.Metric, Sentry, Version=4.0.0.0, Culture=neutral, PublicKeyToken=fba2ec45388e2af0]], System.Collections.Concurrent, Version=7.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a]].get_Item(Int64 )
   at Sentry.Tests.MetricAggregatorTests.Distribution_AggregatesMetrics()
   at System.Reflection.MethodInvoker.InterpretedInvoke(Object , Span`1 , BindingFlags )
```

jamescrosswell:
I think this might be the call to `ConcurrentDictionary.GetOrAdd`:
https://github.com/getsentry/sentry-dotnet/blob/0fde2e34300a961fd62fcfd77614ec51b11275e3/src/Sentry/MetricAggregator.cs#L186

The docs indicate:
[...]

PR #3036, opened 2024-01-11
Replaced ConcurrentDictionary with ReaderWriterLockSlim in MetricAggregator

jamescrosswell:
#skip-changelog

Checking to see if this helps with https://github.com/getsentry/sentry-dotnet/issues/3033... I think we merge this in, once reviewed, and monitor to see if we have any more test failures over the next couple of weeks before resolving the original issue.

### DIFF
--- a/src/Sentry/MetricAggregator.cs
+++ b/src/Sentry/MetricAggregator.cs
@@ -223,27 +223,23 @@ internal class MetricAggregator : IMetricAggregator
             {
                 return existingBucket;
             }
-            else
+
+            _bucketsLock.EnterWriteLock();
+            try
             {
-                _bucketsLock.EnterWriteLock();
-                try
+                // Check again in case another thread added the bucket while we were waiting for the write lock
+                if (Buckets.TryGetValue(bucketKey, out existingBucket))
                 {
-                    // Check again in case another thread added the bucket while we were waiting for the write lock
-                    if (Buckets.TryGetValue(bucketKey, out existingBucket))
-                    {
-                        return existingBucket;
-                    }
-                    else
-                    {
-                        var timeBucket = new ConcurrentDictionary<string, Metric>();
-                        Buckets[bucketKey] = timeBucket;
-                        return timeBucket;
-                    }
+                    return existingBucket;
                 }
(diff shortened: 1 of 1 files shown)


src/Sentry/MetricAggregator.cs
```diff
@@ -216,6 +214,44 @@ static string EscapeString(string input, params char[] charsToEscape)
...
+                        var timeBucket = new ConcurrentDictionary<string, Metric>();
+                        Buckets[bucketKey] = timeBucket;
+                        return timeBucket;
+                    }
+                }
+                finally
+                {
+                    _bucketsLock.ExitWriteLock();
+                }
+            }
```
vaind, line 247: nit: maybe we could decrease the depth here a bit, e.g. 
```suggestion

            _bucketsLock.EnterWriteLock();
            try
            {
                // Check again in case another thread added the bucket while we were waiting for the write lock
                if (Buckets.TryGetValue(bucketKey, out existingBucket))
                {
                    return existingBucket;
                }
                
                var timeBucket = new ConcurrentDictionary<string, Metric>();
                Buckets[bucketKey] = timeBucket;
                return timeBucket;
            }
            finally
            {
                _bucketsLock.ExitWriteLock();
            }
```

src/Sentry/MetricAggregator.cs
```diff
@@ -334,9 +370,27 @@ public async Task FlushAsync(bool force = true, CancellationToken cancellationTo
...
+                        continue;
+                    }
+#else
+                    if (!Buckets.ContainsKey(key))
+                    {
+                        continue;
+                    }
+                    bucket = Buckets[key];
+                    Buckets.Remove(key);
+#endif
```
vaind, line 389: nit? Maybe just keep a single variant of the code? Should work in any .net version, I think?
```suggestion
                    if (!Buckets.ContainsKey(key))
                    {
                        continue;
                    }
                    bucket = Buckets[key];
                    Buckets.Remove(key);
```

vaind (approved):

bitsandfoxes (approved):

PR #3040, opened 2024-01-11
chore: update to .NET 8.0.101

vaind:
#skip-changelog

### DIFF
--- a/global.json
+++ b/global.json
@@ -1,6 +1,6 @@
 {
   "sdk": {
-    "version": "8.0.100",
+    "version": "8.0.101",
     "rollForward": "latestMinor",
     "allowPrerelease": false
   }


global.json
```diff
@@ -1,6 +1,6 @@
 {
   "sdk": {
-    "version": "8.0.100",
+    "version": "8.0.101",
     "rollForward": "latestMinor",
```
bruno-garcia, line 4: it should roll forward already, do we need this?

vaind: hmm, guess not then

SimonCropp: it only rolls forward if people have it installed. 

I would think you would prefer not to build from an sdk with a CVE?

bruno-garcia: By reading the CVE it doesn't seem to affect building a library. It's a runtime thing

That said I don't mind pinning it to a newer version. I just reviewed it as-is without the context on why (usually we don't bump for patch releases on global.json so any version from the base is OK to make things work)

Issue #1491, opened 2022-02-18
Rename types likely to clash to include `ImplicitUsings` by default


bruno-garcia:
See: https://github.com/getsentry/sentry-dotnet/issues/1487

mattjohnsonpint:
Clear example, `Sentry.Span` conflicting with `System.Span<T>`.

Before simply renaming everything, we should re-think our type names within the context of implict/global usings.  Does the whole .NET ecosystem need to re-think this?  Are these conversations already happening, and can we align to what others might be doing?

mattjohnsonpint:
Discussed.  I think we should do this for the next major (4.0), especially in light of global usings. Another good example is `Transaction`, which should be `SentryTransaction`.

Lets do this as part of reviewing the public API surface for 4.0.

mattjohnsonpint:
See also #1585

bruno-garcia:
We should give this one a try now on 4.0.0. 

Adding `ImplicitUsings`  to a project should allow you do do a basic `SentrySdk.Init` call and basic `SentrySdk.CaptureException` without having to do any `using Sentry` or `Sentry.Anything`

Wdyt @bitsandfoxes @jamescrosswell ?

jamescrosswell:
> Wdyt @bitsandfoxes @jamescrosswell ?

From memory, we initially pushed this back to 5.0.0 from 4.0.0 to descope the 4.0 release a bit and make sure we'd be able to ship it around the same time as the .net8.0 release. 

I think we've derisked that now so could potentially put this one back in there. It'd be an easy first issue to pick up.

bruno-garcia:
Since the 3.x SDK works fine on .NET 8, releasing 4.0.0 is not time sensitive. So if we want to extend the scope/time to take advantage of the major bump, we can

bruno-garcia:
> What about `Scope`?

Since that'll be more prevalent going forward, to keep it in sync with most SDKs, I think it's fine we keep it as-is.

Issue #1487, opened 2022-02-16
Revert adding Sentry to global usings


jhartmann123:
The latest Release 3.14.0 comes with the following feature:

> Add Sentry to global usings when ImplicitUsings is enabled (<ImplicitUsings>true</ImplicitUsings>) (https://github.com/getsentry/sentry-dotnet/pull/1398)

We do have implicit usings enabled, with the intention was that only the often used "standard"-namespaces are added there.
Sentry on the other hand is a namespace which we use exactly once in our ASP.NET Core API projects: when configuring Sentry on startup.

For me as an "end user" of this library it does make absolutely no sense to have Sentry in the global usings.

**Main issue:**  
The main conflict here is `Sentry.User`. I guess a `User` is not a really uncommon name for a class. Now, with the latest update we have a lot of ambiguous references between Sentry.User and our Domain-User. As a workaround I'd have to explicitly reference our Domain-User in a ton of classes over multiple projects. Which I really don't want to do.

So I vote for removing Sentry from global usings again.

bruno-garcia:
`User` is definitely high chance of conflict. I suggest we take a pragmatic approach:

1. asap: release a patch release with a opt-out for ImplicitUsings (e.g: `<SentryImplicitUsings>false`)
2. Name Sentry.User to `SentryUser` on the next major change. Do such renames on other types that are high risk of conflict

davidroth:
IMO adding Sentry to global usings was a mistake. This library is used only in the composition root in 99% of the cases. The potential for real naming clashes (`User`) and similar types (i.e. `Sentry.Span` vs `System.Span<T>`) is real and not worth it IMO.

bruno-garcia:
We discussed and agreed to revert this feature, until we release 4.0 that will include renaming some types prone to conflict.
Lets spin off a new issue to track the class renames and adding this feature back in.

PR #3043, opened 2024-01-11
add sentry to global usings

bruno-garcia:
Now that this is complete:
* https://github.com/getsentry/sentry-dotnet/issues/1491

We can enable namespace `Sentry` by default

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,8 +4,13 @@
 
 ### Significant change in behavior
 
-- Added `Sentry` namespace to global usings when `ImplicitUsings is enabled ([#3043](https://github.com/getsentry/sentry-dotnet/pull/3043))
-  - To opt-out, use `<SentryImplicitUsings>false</SentryImplicitUsings>` to your project.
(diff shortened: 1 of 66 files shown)


bruno-garcia:
Actually the opt-out could be simpler, as Simon said here:

https://github.com/getsentry/sentry-dotnet/issues/1487#issuecomment-1043318689

bruno-garcia:
Added back `SentryImplicitUsings` for opt-out because on generated code, it causes compilation error, so the `ItemGroup Using Remove=Sentry` approach didn't work.

bitsandfoxes (approved):

frankbuckley:
This clashes all over the place for me - and let alone our own types, obvious core library types - e.g.:

```
error CS0104: 'ISession' is an ambiguous reference between 'Microsoft.AspNetCore.Http.ISession' and 'Sentry.ISession'
```

It is a **huge** assumption to think that library users want `using Sentry` inkected into every source file in their projects.

Issue #3027, opened 2024-01-09
Log a warning if `Debug=true` and `Environment=production`


bruno-garcia:
SDK diagnostic logger, specially in the default level (debug) can cause performance degradations: lots of allocations, and printing to console can slow down the app.

We should log out that and link to the diagnostic logs docs page (where it talks about level)

PR #3045, opened 2024-01-12
fix: Log warning in case of `debug` in `production`

bitsandfoxes:
Fixes https://github.com/getsentry/sentry-dotnet/issues/3027

#skip-changelog

### DIFF
--- a/src/Sentry/SentryOptionsExtensions.cs
+++ b/src/Sentry/SentryOptionsExtensions.cs
@@ -373,9 +373,8 @@ public static class SentryOptionsExtensions
 
             if (options.SettingLocator.GetEnvironment().Equals("production", StringComparison.OrdinalIgnoreCase))
             {
-                options.DiagnosticLogger.LogWarning("The DiagnosticLogger, specifically on the diagnostic level " +
-                                                    "`debug` (default) can cause performance degradations. " +
-                                                    "Many allocations, and printing to console can slow down your app." +
+                options.DiagnosticLogger.LogWarning("Sentry option 'Debug' is set to true while Environment is production. " +
+                                                    "Be aware this can cause performance degradation and is not advised. " +
                                                     "See https://docs.sentry.io/platforms/dotnet/configuration/diagnostic-logger " +
                                                     "for more information");
             }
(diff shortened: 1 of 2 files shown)


src/Sentry/SentryOptionsExtensions.cs
```diff
@@ -370,6 +370,15 @@ internal static void SetupLogging(this SentryOptions options)
...
                     options.DiagnosticLevel);
             }
+
+            if (options.SettingLocator.GetEnvironment().Equals("production", StringComparison.OrdinalIgnoreCase))
+            {
+                options.DiagnosticLogger.LogWarning("The DiagnosticLogger, specifically on the diagnostic level " +
+                                                    "`debug` (default) can cause performance degradations. " +
+                                                    "Many allocations, and printing to console can slow down your app." +
+                                                    "See https://docs.sentry.io/platforms/dotnet/configuration/diagnostic-logger " +
+                                                    "for more information");
```
bruno-garcia, line 380: I think most users don't know what Diagnostic Logger is. All they know is `Debug=true` or not even (because they copied the code we gave them, which includes this).

Worth mentioning that. Possibly simplifying the whole message (details can be in the link).
```suggestion
                options.DiagnosticLogger.LogWarning("Sentry option 'Debug' is set to true while Environment is production. " +
                                                    "Be aware this can cause performance degradation and is not advised. " +
                                                    "See https://docs.sentry.io/platforms/dotnet/configuration/diagnostic-logger " +
                                                    "for more information");
```

bruno-garcia (approved):
Looks good! Any detail about overhead and levels could be added to: https://docs.sentry.io/platforms/dotnet/configuration/diagnostic-logger/

For example Debug:true adds level:Debug which shows the whole enveloeps serialized on the console. That's obviously expensive. Changing the level to 'Info' will reduce the overhead and still give some context. Even less overhead with less context on Warn and Error. But we suggest not having Debug true in prod unless needed to figure out problems with the SDK and verifying the overhead is acceptable before doing so.

bruno-garcia:
Some tests rely on the logs and are failing:

```
Actually received 2 matching calls:
	Log(Warning, "Sentry option 'Debug' is set to true while Environment is production. Be aware this can cause performance degradation and is not advised. See https://docs.sentry.io/platforms/dotnet/configuration/diagnostic-logger for more information", <null>, Object[])
	Log(Warning, "Sentry option 'Debug' is set to true while Environment is production. Be aware this can cause performance degradation and is not advised. See https://docs.sentry.io/platforms/dotnet/configuration/diagnostic-logger for more information", <null>, Object[])
```

jamescrosswell (approved):

Issue #3020, opened 2024-01-09
Add default metrics


jamescrosswell:
There are various metrics that have already been instrumented in .NET which users might be interested in sending to Sentry:

- [Built In Metrics](https://learn.microsoft.com/en-us/dotnet/core/diagnostics/built-in-metrics) that are accessible via the [System.Diagnostics.Metrics](https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.metrics) API.
- [Well known Counters](https://learn.microsoft.com/en-us/dotnet/core/diagnostics/available-counters) 
- [Polly](https://github.com/App-vNext/Polly) resilience metrics
- [Available Threads](https://learn.microsoft.com/en-us/dotnet/api/system.threading.threadpool.getavailablethreads?view=net-8.0)
- FirstChanceException
- [ObservableInstruments](https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.metrics.meterlistener.recordobservableinstruments?view=net-8.0#system-diagnostics-metrics-meterlistener-recordobservableinstruments)

We'll likely resolve this issue in multiple Pull Requests:

- [ ] https://github.com/getsentry/sentry-dotnet/pull/3052
- [ ] https://github.com/getsentry/sentry-dotnet/pull/3100

bruno-garcia:
> Furthermore, the Polly.Extensions package now provides seamless integration with IServiceCollection and facilitates the export of Polly’s telemetry events into .NET metrics.

https://devblogs.microsoft.com/dotnet/building-resilient-cloud-services-with-dotnet-8/

maybe something interesting to tap into

jamescrosswell:
See [Built in Metrics](https://learn.microsoft.com/en-us/dotnet/core/diagnostics/built-in-metrics)

bruno-garcia:
Added `FirstChanceException` to the list (I recall we talked about it but I couldn't find a ticket)

jamescrosswell:
> Added `FirstChanceException` to the list (I recall we talked about it but I couldn't find a ticket)

Yeah I remember discussing it. Were we just thinking of a counter for that one (# of first chance exceptions)?

optical:
[...]

PR #3052, opened 2024-01-16
Capture built in metrics from System.Diagnostics.Metrics API

jamescrosswell:
The PR makes a start on implementing:
- https://github.com/getsentry/sentry-dotnet/issues/3020

For this first integration we'd like to make it easy for Sentry SDK Users to be able to capture any of the [Built In Metrics](https://learn.microsoft.com/en-us/dotnet/core/diagnostics/built-in-metrics) that are accessible via the [System.Diagnostics.Metrics](https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.metrics) API.

SDK users can use [SentryOptions.ExperimentalMetrics.CaptureInstruments](https://github.com/getsentry/sentry-dotnet/blob/4686e0641efd3d0bff34f8dc20061010ca2bd054/src/Sentry/ExperimentalMetricsOptions.cs#L16-L36) to configure one or more `System.Diagnostics.Metrics.Instrument` that they are interested in. The `SystemDiagnosticsMetricsIntegration` (if enabled) will then capture any metrics for those instruments and send them to Sentry.

Currently the integration supports the following Instruments:
* [Counter<T>](https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.metrics.counter-1?view=net-8.0)
* [UpDownCounter<T>](https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.metrics.updowncounter-1?view=net-8.0)
* [ObservableCounter<T>](https://learn.microsoft.com/en-us/dotnet/api/System.Diagnostics.Metrics.ObservableCounter-1?view=net-8.0)
* [ObservableUpDownCounter<T>](https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.metrics.observableupdowncounter-1?view=net-8.0)
* [Histogram<T>](https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.metrics.histogram-1?view=net-8.0)
* [ObservableGauge<T>](https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.metrics.observablegauge-1?view=net-8.0)

## Dependencies

This integration needs `System.Diagnostics.DiagnosticSource` (version 8 I think) to work... Applications that target .NET 8+ include this reference by default so, for the time being, this integration is only available for .NET 8 or greater. I'm not sure if there's a way to enable the integration on previous TFMs if people are OK to add `System.Diagnostics.DiagnosticSource`... I was thinking maybe a separate NuGet package (with that as a dependency) for older TFMs but have it built in for .NET 8 or greater. Two things:
1. I'm not quite sure how to do this... some MS Build trickery now doubt.
2. Would we even want to do this? It would be another NuGet package to maintain. Maybe we wait to see if anyone actually wants this in the first place?

At the very least, I think let's not do it in this PR (we can create a separate issue and PR for this if/when we decide to do it).


src/Sentry/Integrations/SystemDiagnosticsMetricsIntegration.cs
```diff
@@ -0,0 +1,83 @@
...
+        }
+
+        if (SentryMeterListener is not null)
+        {
+            options.Log(SentryLevel.Info,
+                "System.Diagnostics.Metrics Integration has already been registered.");
+            return;
+        }
+
+        lock (InitLock)
```
bruno-garcia, line 29: Nothing in Init is thread safe, do we need this?

bruno-garcia (commented):
quick pass

src/Sentry/Integrations/SystemDiagnosticsMetricsIntegration.cs
jamescrosswell: I was wary about initializing it twice and double counting metrics... but yeah, as it's just about static initialization I'll switch out the double lock for a Lazy<T>.

bruno-garcia:
> This integration needs System.Diagnostics.DiagnosticSource (version 8 I think) to work... Applications that target .NET 8+ include this reference by default so, for the time being, this integration is only available for .NET 8 or greater. I'm not sure if there's a way to enable the integration on previous TFMs if people are OK to add System.Diagnostics.DiagnosticSource... I was thinking maybe a separate NuGet package (with that as a dependency) for older TFMs but have it built in for .NET 8 or greater. Two things:

>    I'm not quite sure how to do this... some MS Build trickery now doubt.
>    Would we even want to do this? It would be another NuGet package to maintain. Maybe we wait to see if anyone actually wants this in the first place?

> At the very least, I think let's not do it in this PR (we can create a separate issue and PR for this if/when we decide to do it)

We have already a package, that's only used for older frameworks:

https://github.com/getsentry/sentry-dotnet/tree/main/src/Sentry.DiagnosticSource#when-shouldnt-i-include-this-package

We could bump the min version it depends on:

https://github.com/getsentry/sentry-dotnet/blob/feabf090918b551d8e7c8c24d27c752881623c37/src/Sentry.DiagnosticSource/Sentry.DiagnosticSource.csproj#L17

Seems to work on older frameworks but they do a trick of copying older DLLs for old targets sometimes to the new package:
image

So worth testing it out.

samples/Sentry.Samples.Console.Metrics/Program.cs
```diff
@@ -18,31 +28,53 @@ private static void Main()
...
+                        HatsSold.Add(Roll.Next(0, 1000));
+                        break;
+                    case 4:
+                        // Here we demonstrate collecting some built in metrics for HTTP requests... this works because
+                        // we've configured ExperimentalMetricsOptions.CaptureInstruments to match "http.client.*"
+                        //
+                        // See https://learn.microsoft.com/en-us/dotnet/core/diagnostics/built-in-metrics-system-net#systemnethttp
+                        var httpClient = new HttpClient();
+                        var url = "https://api.sampleapis.com/coffee/hot";
+                        var result = httpClient
```
bruno-garcia, line 68: you could change `void Main` to `async Task Main` and await here instead

src/Sentry/ExperimentalMetricsOptions.cs
```diff
@@ -0,0 +1,37 @@
...
+    /// </para>
+    /// <para>
+    /// These can be either custom System.Diagnostics.Metrics that you have instrumented in your
+    /// application or any of the built in metrics that are available.
+    /// </para>
+    /// <para>
+    /// See https://learn.microsoft.com/en-us/dotnet/core/diagnostics/built-in-metrics for more information.
+    /// </para>
+    /// </summary>
+    public IList<SubstringOrRegexPattern> CaptureInstruments
```
bruno-garcia, line 29: is `Instruments` the best name? 

Also: Will we have some built-in ones here? And if so, we can't get code locations for those?

src/Sentry/ExperimentalMetricsOptions.cs
```diff
@@ -0,0 +1,37 @@
...
+    {
+        // NOTE: During configuration binding, .NET 6 and lower used to just call Add on the existing item.
+        //       .NET 7 changed this to call the setter with an array that already starts with the old value.
+        //       We have to handle both cases.
+        get => _captureInstruments;
+        set => _captureInstruments = value.SetWithConfigBinding();
```
bruno-garcia, line 35: `SetXyb()` but it says set and takes no arguments. It's hard to parse. Is  `WIthConfigBinding`  a better choice? Or what is this doing actually since both prop and field have the same type

src/Sentry/ExperimentalMetricsOptions.cs
jamescrosswell: I used Instruments since the `System.Diagnostics.Metrics` API uses [Instrument](https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.metrics.instrument?view=net-8.0) as the base class for all System Diagnostic Metrics. 

Potentially `CaptureSystemDiagnosticsMetrics` would be more obvious/explicit?

In terms of defaults, we could capture none/some/all of [these build in metrics](https://learn.microsoft.com/en-us/dotnet/core/diagnostics/built-in-metrics). 

Are there quotas that we need to be mindful of (so are we potentially going to chew through someone's entire Sentry Quota if we start collecting [routing match attempt](https://learn.microsoft.com/en-us/dotnet/core/diagnostics/built-in-metrics-aspnetcore#metric-aspnetcoreroutingmatch_attempts) metrics?

If not, all of those built in metrics are potentially useful/interesting to people so we could enable these.

That raises another issue about the current API for enabling/disabling metrics. Currently this property is a bunch of SubstringOrRegexPatterns that match against the Instrument.Name. This is how you manually instrument stuff:
```
    private static readonly Meter HatsMeter = new(name: "HatCo.HatStore", "1.0.0");
    private static readonly Counter<int> HatsSold = HatsMeter.CreateCounter<int>(
        name: "hats-sold",
        unit: "Hats",
        description: "The number of hats sold in our store");
```

In the built in metrics you can see that related instruments often share the same prefix, so we could have a bunch of regex expresses to match those. 

Alternatively, those instruments were all created by the same meter, so we could also offer the ability to match against the meter name... that would make it easier for us to configure broad categories of built in metrics that we want to listen to (and implicitly easier for users to do the same) so sounds like a good idea.

I think it'd be an inclusive/or match (so we listen to the metric **either** if we match on the `Instrument` name **or** if we match on the `Meter` name). Possibly we'd want a separate options for each of those, in which case `CaptureSystemDiagnosticsMetrics` wouldn't be explicit enough... we'd probably want separate `CaptureSystemDiagnosticsInstruments` and `CaptureSystemDiagnosticsMeters` options. 

As I write this, I'm persuading myself that this is a good idea and those are probably good names 😉

bruno-garcia: > Are there quotas that we need to be mindful of (so are we potentially going to chew through someone's entire Sentry Quota if we start collecting routing match attempt metrics?

Very good point, but we can have some built in metrics that don't count towards quota perhaps, if we don't have high cadinality, basically 0 tags.

> Potentially CaptureSystemDiagnosticsMetrics would be more obvious/explicit?

Not a bad idea, but maybe ur original one is better. dunno @bitsandfoxes thoughts?

> Possibly we'd want a separate options for each of those,

That makes sense. 

How about:

```
SystemDiagnosticsMetrics.Instruments = [".."];
SystemDiagnosticsMetrics.Meters = [".."];
```
Two properties and have `SystemDiagnosticsMetrics` be non nullable?

src/Sentry/ExperimentalMetricsOptions.cs
bruno-garcia: extension method on IList? seems to me this could live elsewhere then, I feel we've abusing extension methods recently

src/Sentry/Internal/SystemDiagnosticsMetricsListener.cs
```diff
@@ -0,0 +1,86 @@
...
+
+    internal void RecordMeasurement<T>(
+        Instrument instrument,
+        T measurement,
+        ReadOnlySpan<KeyValuePair<string, object?>> tags,
+        object? _)
+        where T: struct
+    {
+        var unit = MeasurementUnit.Parse(instrument.Unit);
+        var tagDict = tags.ToImmutableArray().ToImmutableDictionary(
```
bruno-garcia, line 59: 👁️ it's almost midnight I need to get back to this and do a proper review 😅

src/Sentry/ExperimentalMetricsOptions.cs
```diff
@@ -0,0 +1,61 @@
...
+    /// matches one of the items in this list will be collected and reported to Sentry.
+    /// </para>
+    /// <para>
+    /// These can be either custom Instruments that you have created or any of the built in metrics that are available.
+    /// </para>
+    /// <para>
+    /// See https://learn.microsoft.com/en-us/dotnet/core/diagnostics/built-in-metrics for more information.
+    /// </para>
+    /// </summary>
+    public IList<SubstringOrRegexPattern> CaptureSystemDiagnosticsInstruments
```
jamescrosswell, line 30: In terms of defaults, we could capture none/some/all of [these build in metrics](https://learn.microsoft.com/en-us/dotnet/core/diagnostics/built-in-metrics). 

Are there quotas that we need to be mindful of (so are we potentially going to chew through someone's entire Sentry Quota if we start collecting [routing match attempt](https://learn.microsoft.com/en-us/dotnet/core/diagnostics/built-in-metrics-aspnetcore#metric-aspnetcoreroutingmatch_attempts) metrics?

If not, all of those built in metrics are potentially useful/interesting to people so we could enable these.

jamescrosswell:
> We have already a package, that's only used for older frameworks:
> 
> https://github.com/getsentry/sentry-dotnet/tree/main/src/Sentry.DiagnosticSource#when-shouldnt-i-include-this-package

OK, interesting. So we have:

Sentry DiagnosticSource integration
- Baked in on `net6.0` and later at the moment

System DiagnosticsMetrics integration
- Baked in on `net8.0` and later at the moment

Assuming we can bump the `System.Diagnostics.DiagnosticSource` version from `4.5.0` to `8.0.0`, I could a `.src/Sentry.DiagnosticSource/Internal/DiagnosticsMetrics` folder containing the Metrics integration... include that in the root project for `net8.0` and later and then people could grab the Sentry.DiagnosticSource **either** if they wanted the DiagnosticSource integration for netstandard/netframework **or** if they wanted the Metrics integration for `net6.0` or below.

I am wondering whether that might create a clash however. If someone was running `net6.0` and added a reference to `Sentry.DiagnosticSource` so that they could use the metrics integration... presumably their Sentry.nupkg would already include everything from `src/Sentry.DiagnosticSource/Internal/DiagnosticSource` but now so would the Sentry.DiagnosticSource.nupkg that they referred to as well. 

Seems like actually these would have to exist as separate packages then... So we'd need a new Sentry.DiagnosticSourceMetrics package that people could add if they're on net6.0 or below, and we just do the same trick for that that we've done already for Sentry.DiagnosticSource.

test/Sentry.Tests/SentryOptionsTests.verify.cs
```diff
@@ -18,9 +18,12 @@ public Task Integrations_default_ones_are_properly_registered()
...
         Hub _ = new(options, Substitute.For<ISentryClient>());
 
-        return Verify(logger.Entries)
-            .UniqueForOSPlatform()
+        var settingsTask = Verify(logger.Entries)
+            .UniqueForTargetFrameworkAndVersion()
             .UniqueForRuntime()
             .AutoVerify(includeBuildServer: false);
+        if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
+            settingsTask = settingsTask.UniqueForOSPlatform();
```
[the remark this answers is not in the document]
jamescrosswell: We it's only unique in that instance. If it was ever unique on any of the other platforms, our verify tests would start failing on those platforms. Until then though, they all drop back to the "default" verify received file. We could make it unique for all OS Platforms but it would just results in a bunch of extra/redundant verify files (that were all the same).

jamescrosswell:
> Seems like actually these would have to exist as separate packages then... So we'd need a new Sentry.DiagnosticSource.Metrics package that people could add if they're on net6.0 or below, and we just do the same trick for that that we've done already for Sentry.DiagnosticSource.

I've added a separate issue to track that work:
- https://github.com/getsentry/sentry-dotnet/issues/3090

src/Sentry/ExperimentalMetricsOptions.cs
jamescrosswell: I've updated this PR to include all of the built in metrics by default.

src/Sentry/BuiltInSystemDiagnosticsMeters.cs
```diff
@@ -0,0 +1,170 @@
...
+    private static partial Regex SystemNetHttpRegex();
+#else
+    public static readonly SubstringOrRegexPattern SystemNetHttp = new Regex(SystemNetHttpPattern, RegexOptions.Compiled);
+#endif
+
+    /// <summary>
+    /// Matches all built in metrics
+    /// </summary>
+    /// <returns></returns>
+    public static IList<SubstringOrRegexPattern> All = [
```
bruno-garcia, line 157: How about Lazy<T>?

I wonder if we can avoid allocating this list and all regex's if the user is not even setting the default metrics. Or if they are setting specific ones here instead of leaving the default All

src/Sentry/ExperimentalMetricsOptions.cs
bruno-garcia: Since this is experimental I think it's fine to include it by default but definitely worth keeping an eye on impact.
Cost is usually driven by cardinality so if the metrics have no tags it could be OK. cc @bitsandfoxes to own this conversation internally

src/Sentry/Internal/SystemDiagnosticsMetricsListener.cs
bruno-garcia: Worth profiling this to see the impact. This being ReadOnlySpan<T> and us allocating immutableArray then dictionary from it seems like could show up depending how often this is called.

src/Sentry/Internal/SystemDiagnosticsMetricsListener.cs
```diff
@@ -0,0 +1,87 @@
...
+            kvp => kvp.Key,
+            kvp => kvp.Value?.ToString() ?? string.Empty
+        );
+        var doubleMeasurement = Convert.ToDouble(measurement);
```
bruno-garcia, line 64: `where T: struct`

This seems very unsafe. How do we guarantee this won't blow up?
Is it because all data types you set as `T` above are convertable to Double? if so we need some comments here and above to clarify that.
Main concern is someone adding a new value as `T` above and causing issues here

bruno-garcia (approved):
No blockers so approving, we can follow up with improvements before GA

src/Sentry/ExperimentalMetricsOptions.cs
jamescrosswell: The metrics definitely have tags. For example, [`http.server.request.duration`](https://learn.microsoft.com/en-us/dotnet/core/diagnostics/built-in-metrics-aspnetcore#metric-httpserverrequestduration) has 7 attributes, which add dimensionality represented as tags in Sentry. In most situations, that will result in 1 different metric per route configured in the ASP.NET Core application... but that's just one example. 

Basically anything that's documented as an Attribute in the built in metrics documentation turns up as a tag in Sentry.

src/Sentry/Internal/SystemDiagnosticsMetricsListener.cs
jamescrosswell: Yeah I think something got lost when we made this public (and then reverted to internal). I can add another constraint here `where: struct, IConvertible`. We couldn't do that when it was public becuase `IConvertible` isn't CLS Compliant (can't have it on public interfaces targeting netstandard).

src/Sentry/Internal/SystemDiagnosticsMetricsListener.cs
jamescrosswell: I've added a separate issue to track this:
- https://github.com/getsentry/sentry-dotnet/issues/3093

Issue #3050, opened 2024-01-13
Skip WinUI outside Windows


bruno-garcia:
Running Sentry on a Raspberry Pi I see on the logs:

```
  Debug: Registering integration: 'WinUIUnhandledExceptionIntegration'.
```

Can we skip adding this integration if not running on Windows?

PR #3055, opened 2024-01-17
fix: Add `WinUIUnhandledExceptionIntegration` on Windows only

bitsandfoxes:
Fixes https://github.com/getsentry/sentry-dotnet/issues/3050

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,7 @@ If you have conflicts, you can opt-out by adding the following to your `csproj`:
 
 ### Fixes
 
-- The SDK no longer adds the `WinUIUnhandledExceptionIntegration` on non Windows platforms ([#3043](https://github.com/getsentry/sentry-dotnet/pull/3043))
+- The SDK no longer adds the `WinUIUnhandledExceptionIntegration` on non Windows platforms ([#3055](https://github.com/getsentry/sentry-dotnet/pull/3055))
 
 ### Dependencies
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,10 @@ If you have conflicts, you can opt-out by adding the following to your `csproj`:
 </PropertyGroup>
 ```
 
+### Fixes
+
+- The SDK no longer adds the `WinUIUnhandledExceptionIntegration` on non Windows platforms ([#3043](https://github.com/getsentry/sentry-dotnet/pull/3043))
+
 ### Dependencies
 
 - Bump Java SDK from v7.1.0 to v7.2.0 ([#3049](https://github.com/getsentry/sentry-dotnet/pull/3049))
--- a/src/Sentry/SentryOptions.cs
+++ b/src/Sentry/SentryOptions.cs
@@ -186,7 +186,7 @@ public class SentryOptions
             }
 #endif
 
-#if NET5_0_OR_GREATER && !__MOBILE__
+#if NET5_0_OR_GREATER && WINDOWS
             if ((_defaultIntegrations & DefaultIntegrations.WinUiUnhandledExceptionIntegration) != 0)
             {
                 yield return new WinUIUnhandledExceptionIntegration();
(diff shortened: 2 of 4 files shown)


CHANGELOG.md
```diff
@@ -12,6 +12,10 @@ If you have conflicts, you can opt-out by adding the following to your `csproj`:
 </PropertyGroup>
 ```
 
+### Fixes
+
+- The SDK no longer adds the `WinUIUnhandledExceptionIntegration` on non Windows platforms ([#3043](https://github.com/getsentry/sentry-dotnet/pull/3043))
```
bitsandfoxes, line 17:
```suggestion
- The SDK no longer adds the `WinUIUnhandledExceptionIntegration` on non Windows platforms ([#3055](https://github.com/getsentry/sentry-dotnet/pull/3055))
```

src/Sentry/SentryOptions.cs
```diff
@@ -186,7 +186,7 @@ internal IEnumerable<ISdkIntegration> Integrations
             }
 #endif
 
-#if NET5_0_OR_GREATER && !__MOBILE__
+#if NET5_0_OR_GREATER && WINDOWS
```
bruno-garcia, line 189: Does this mean building for `net8.0` for example won't have this handler and is that by design?
Sounds right to me, but checking, if a WInUI app has to always be built with `net8.0-windowsXYZ` or something.

bruno-garcia: @lucas-zimerman might know

bruno-garcia: Note: we don't have a `net8.0-windows` target in the SDK so I think this isn't getting what you expect

bitsandfoxes: This is a big TIL. So we have to check during runtime whether to add it depending on the current platform.

lucas-zimerman: to be honest I don't follow about the support of UnhandledException on .NET 8.

bruno-garcia: > This is a big TIL. So we have to check during runtime whether to add it depending on the current platform.

yeah all the compiler flags affect the DLLs we output. And `net8.0` is platform agnostic so ANDROID/WINDOWS/etc wil not be set.

test/Sentry.Tests/SentryOptionsTests.verify.cs
```diff
@@ -3,9 +3,12 @@ namespace Sentry.Tests;
 [UsesVerify]
 public partial class SentryOptionsTests
 {
-    [Fact]
+    [SkippableFact]
     public Task Integrations_default_ones_are_properly_registered()
     {
+        // Windows additionally adds `WinUIUnhandledExceptionIntegration`
+        Skip.If(RuntimeInformation.IsOSPlatform(OSPlatform.Windows));
```
bruno-garcia, line 10: should we have two tests then? one that tests we have this on Windows? Instead of skipping

bitsandfoxes: I had that but it messed with the verify. I guess I need to run this on windows then too to generate the expected file?

bruno-garcia: yeah you need to have a file per platform, just saw this on another PR:
image

bruno-garcia: one of those per OS I imagine

bruno-garcia (approved):
approving to unblock but best finding out a way to test it instead of skipping on Windows

bruno-garcia:
> approving to unblock but best finding out a way to test it instead of skipping on Windows

@bitsandfoxes is there a ticket to track this? good to link to this PR

bitsandfoxes:
> > approving to unblock but best finding out a way to test it instead of skipping on Windows
> 
> @bitsandfoxes is there a ticket to track this? good to link to this PR

Follor up PR is here https://github.com/getsentry/sentry-dotnet/pull/3063

PR #3065, opened 2024-01-19
fix: Sanitize value in metrics

bitsandfoxes:
Fixes https://github.com/getsentry/symbol-collector/issues/175

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ If you have conflicts, you can opt-out by adding the following to your `csproj`:
 
 ### Fixes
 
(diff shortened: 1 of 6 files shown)


bruno-garcia (commented):
if we had tests we wouldn't have had this bug in prod, can u add a test?

jamescrosswell:
> if we had tests we wouldn't have had this bug in prod, can u add a test?

We probably need tests more generally for the metrics capability. We intentionally didn't do much testing just to put a POC together but I think we need to circle back to this:
- https://github.com/getsentry/sentry-dotnet/issues/3066

jamescrosswell (approved):

Issue #3061, opened 2024-01-18
Enabling Open Telemetry Breaks Query Tracking (4.0.0-beta.8)


Silvenga:
### Package

Sentry.AspNetCore

### .NET Flavor

.NET

### .NET Version

8.0.0

### OS

Any (not platform specific)

### SDK Version

4.0.0-beta.8

### Steps to Reproduce

When configuring the Sentry SDK via the `SentryWebHostBuilderExtensions.UseSentry` method, use the `SentryOptionsExtensions.UseOpenTelemetry` method to setup Open Telemetry support.

```csharp
(IWebHostBuilder).UseSentry((context, options) =>
{
  options.UseOpenTelemetry();
});
```

### Expected Result

I expect Open Telemetry support to be enabled, but I also expect native Sentry instrumentation to work as well (Open Telemetry for EFCore is still in flux, however, I mostly want to use Sentry's native integration to ensure the UI operates correctly around slow queries). 

I would expect output similar to this in the logs (shown with `UseOpenTelemetry` disabled):

```logs
[2024-01-18 09:38:55.3291 DEBUG ISentryClient] Registering integration: 'SentryDiagnosticListenerIntegration'.
[2024-01-18 09:38:55.3291 DEBUG ISentryClient] Registering SQL Client integration.
[2024-01-18 09:38:55.3291 DEBUG ISentryClient] Registering EF Core integration

...

[2024-01-18 09:39:48.9569 DEBUG ISentryClient] New scope pushed.
[2024-01-18 09:39:48.9569 DEBUG ISentryClient] Creating a propagation context from headers.
[2024-01-18 09:39:48.9569 INFO ISentryClient] Sentry trace header is null. Creating new Sentry Propagation Context.
[2024-01-18 09:39:48.9569 INFO LoggingMiddleware] Request 'GET /api/v1/users/me' from ::1.
[2024-01-18 09:39:48.9569 INFO ISentryClient] Started transaction with span ID 'a16dcc289a805ff6' and trace ID 'd87a6ddf1afd40f3ba7b05e98ef3f47b'.
[2024-01-18 09:39:48.9766 DEBUG ISentryClient] (Sentry add span db.query.compile)
[2024-01-18 09:39:48.9766 DEBUG ISentryClient] Transaction Spans
[2024-01-18 09:39:48.9766 DEBUG ISentryClient] -----------------
[...]

PR #3072, opened 2024-01-22
Set the scope transaction for Otel transactions

jamescrosswell:
Resolves https://github.com/getsentry/sentry-dotnet/issues/3061

## Problem Summary

When using OpenTelemetry this is returning null:
https://github.com/getsentry/sentry-dotnet/blob/84925a9da90197e88ef41013e80e028358b9951f/src/Sentry.DiagnosticSource/Internal/DiagnosticSource/SentrySqlListener.cs#L75-L79

When instrumenting with OpenTelemetry, the `SentrySpanProcessor` wasn't assigning these to the `Scope.Transaction`... so the integrations weren't able to create DB Spans under the DB Root.

### DIFF
--- a/src/Sentry/TransactionTracer.cs
+++ b/src/Sentry/TransactionTracer.cs
@@ -276,14 +276,6 @@ public class TransactionTracer : ITransactionTracer
     internal ISpan StartChild(SpanId? spanId, SpanId parentSpanId, string operation,
         Instrumenter instrumenter = Instrumenter.Sentry)
     {
-        if (instrumenter != _instrumenter)
-        {
-            _options?.LogWarning(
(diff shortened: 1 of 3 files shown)


src/Sentry/TransactionTracer.cs
```diff
@@ -275,14 +275,6 @@ internal TransactionTracer(IHub hub, ITransactionContext context, TimeSpan? idle
     internal ISpan StartChild(SpanId? spanId, SpanId parentSpanId, string operation,
         Instrumenter instrumenter = Instrumenter.Sentry)
     {
-        if (instrumenter != _instrumenter)
```
bitsandfoxes, line 278: Why is this no longer needed?

bitsandfoxes (approved):

Issue #3074, opened 2024-01-23
Add StartTimer extension method to MetricsAggrigator


jamescrosswell:
> The API is `SentrySdk.Metrics.StartTimer()` (or Timing or Timed as you said but that feels awkward to me, is that how .NET'sn new metric API does it? statsd from JustEats calls it StartTimer I believe (it's used in this repo)

_Originally posted by @bruno-garcia in https://github.com/getsentry/symbol-collector/pull/178#discussion_r1462087645_

PR #3075, opened 2024-01-23
WIP: Added StartTimer extension method to IMetricAggregator

jamescrosswell:
Resolves https://github.com/getsentry/sentry-dotnet/issues/3074

This is a simple way to enable something like `SentrySdk.Metrics.StartTimer()`

One thing that bugs me a bit is that, unlike other instruments, `Timing` creates a span/transaction... and to do that it needs an `IHub`. The `MetricAggregator` can't supply that dependency so under the hood this extension method uses `SentrySdk.CurrentHub` to create spans/transactions for Timings. 

It would be possible to create an internal overload of the extension method that accepts an IHub instance as a parameter, for testing purpose... which is pretty much how the existing/previous `Timing` constructor worked, so I guess it holds together.

We've discussed previously the possibility of putting the MetricAggregator on `IHub` rather than on `ISentryClient`. I think that would make the code significantly more complex. It would couple the MetricAggregator with the Hub (where currently those things are decoupled) and add another layer of indirection for simple things like capturing metric events and transmitting them to Sentry. So I'm not a fan of that solution.

In any event, this solution enables the API interface we discussed. The only other thing we might want to do is make the `Timing` class internal so that people can't instanciate that directly (they have to use the `IMetricAggregator.StartTimer()` syntax).

### DIFF
--- a/src/Sentry/IMetricAggregator.cs
+++ b/src/Sentry/IMetricAggregator.cs
@@ -105,3 +105,23 @@ public interface IMetricAggregator: IDisposable
     /// <returns>False if a shutdown is requested during flush, true otherwise</returns>
     Task FlushAsync(bool force = true, CancellationToken cancellationToken = default);
 }
+
+/// <summary>
+/// Extension methods for <see cref="IMetricAggregator"/>.
+/// </summary>
+public static class MetricsAggregatorExtensions
+{
+    /// <summary>
+    /// Measures the time it takes to run a given code block and emits this as a metric.
+    /// </summary>
+    /// <example>
+    /// using (SentrySdk.Metrics.StartTimer("my-operation"))
+    /// {
+    ///     ...
+    /// }
+    /// </example>
+    public static IDisposable StartTimer(this IMetricAggregator aggregator, string key,
+        MeasurementUnit.Duration unit = MeasurementUnit.Duration.Second, IDictionary<string, string>? tags = null,
+        int stackLevel = 1)
+        => new Timing(SentrySdk.CurrentHub, key, unit, tags, stackLevel + 1);
+}
--- a/samples/Sentry.Samples.Console.Metrics/Program.cs
+++ b/samples/Sentry.Samples.Console.Metrics/Program.cs
@@ -60,9 +60,10 @@ internal static class Program
     {
         var solution = new[] { 3, 5, 7, 11, 13, 17 };
 
-        // The Timing class creates a distribution that is designed to measure the amount of time it takes to run code
+        // StartTimer creates a distribution that is designed to measure the amount of time it takes to run code
         // blocks. By default it will use a unit of Seconds - we're configuring it to use milliseconds here though.
-        using (new Timing("bingo", MeasurementUnit.Duration.Millisecond))
+        // The return value is an IDisposable and the timer will stop when the timer is disposed of.
+        using (SentrySdk.Metrics.StartTimer("bingo", MeasurementUnit.Duration.Millisecond))
         {
             for (var i = 0; i < attempts; i++)
             {
@@ -78,7 +79,7 @@ internal static class Program
 
     private static void CreateRevenueGauge(int sampleCount)
     {
-        using (new Timing(nameof(CreateRevenueGauge), MeasurementUnit.Duration.Millisecond))
+        using (SentrySdk.Metrics.StartTimer(nameof(CreateRevenueGauge), MeasurementUnit.Duration.Millisecond))
         {
             for (var i = 0; i < sampleCount; i++)
             {
@@ -92,7 +93,7 @@ internal static class Program
 
     private static void MeasureShrimp(int sampleCount)
     {
-        using (new Timing(nameof(MeasureShrimp), MeasurementUnit.Duration.Millisecond))
+        using (SentrySdk.Metrics.StartTimer(nameof(MeasureShrimp), MeasurementUnit.Duration.Millisecond))
         {
             for (var i = 0; i < sampleCount; i++)
             {
--- a/src/Sentry/Timing.cs
+++ b/src/Sentry/Timing.cs
@@ -4,8 +4,7 @@ using Sentry.Protocol.Metrics;
 namespace Sentry;
 
 /// <summary>
-/// Measures the time it takes to run a given code block and emits this as a metric. The <see cref="Timing"/> class is
-/// designed to be used in a <c>using</c> statement.
+/// Measures the time it takes to run a given code block and emits this as a metric.
 /// </summary>
 /// <example>
 /// using (var timing = new Timing("my-operation"))


bitsandfoxes (approved):
That looks much nicer to use! Thanks!

bruno-garcia:
> One thing that bugs me a bit is that, unlike other instruments, Timing creates a span/transaction... and to do that it needs an IHub. The MetricAggregator can't supply that dependency so under the hood this extension method uses SentrySdk.CurrentHub to create spans/transactions for Timings.

Remind me again why isn't Metrics on the Hub instead? That's where we keep all state last I saw

src/Sentry/IMetricAggregator.cs
```diff
@@ -105,3 +105,23 @@ public interface IMetricAggregator: IDisposable
...
+    /// <example>
+    /// using (SentrySdk.Metrics.StartTimer("my-operation"))
+    /// {
+    ///     ...
+    /// }
+    /// </example>
+    public static IDisposable StartTimer(this IMetricAggregator aggregator, string key,
+        MeasurementUnit.Duration unit = MeasurementUnit.Duration.Second, IDictionary<string, string>? tags = null,
+        int stackLevel = 1)
+        => new Timing(SentrySdk.CurrentHub, key, unit, tags, stackLevel + 1);
```
bruno-garcia, line 126: using `SentrySdk.CurrentHub` seems a problem to me. Since the whole SDK design was based on instances except `SentrySdk` and `HubAdapter`
```suggestion
        => new Timing(SentrySdk.CurrentHub, key, unit, tags, stackLevel + 1);
```

bruno-garcia (dismissed):

jamescrosswell:
> Remind me again why isn't Metrics on the Hub instead? That's where we keep all state last I saw

I can't find the old conversation (too many threads on the original PR and it's really hard to search in resolved conversations).
 
However, basically it's on the SentryClient as that's where the BackgroundWorker lives... and the aggregator is very similar to the BackgroundWorker. It also doesn't have any dependencies on the Hub - all it needs is a way to send metrics and a way to send code locations (which are two methods on `ISentryClient`). It would be quite a bit of work, at this stage, to move it over to the Hub, so would need a fairly compelling reason.

> That's where the scope lives and where other state lives too like session state I believe, no?

Kind of... it's a bit of a mixed bag. 

The Scope really lives in the `ScopeManager` which holds the `Scope` alongside a `SentryClient`. 
https://github.com/getsentry/sentry-dotnet/blob/941d9dc294312c537401fec6e7ab0bc2520cda63/src/Sentry/Internal/SentryScopeManager.cs#L12-L16

So the Scope state is loosely and indirectly associated with the SentryClient... although normally as an AsyncLocal - unless we're in Global mode - so arguably it's tied to the Thread rather than either the Hub or the Client. The ConfigureScope API that is used to read/mutate this lives on the Hub though... so in that respect it feels like a Hub thing.  

For Session state, both the Hub and the SentryClient hold references to this:
https://github.com/getsentry/sentry-dotnet/blob/2a41c48c2185d910be95bbe850b8979a1a3c571b/src/Sentry/Internal/Hub.cs#L50
https://github.com/getsentry/sentry-dotnet/blob/2a41c48c2185d910be95bbe850b8979a1a3c571b/src/Sentry/SentryClient.cs#L54

Although when using a Hub, the `Hub._ownedClient` will have it's session manager passed in by the Hub. The SentryClient also doesn't provide any methods for working with the session, whereas the Hub has Start/Pause/Resume/End session methods so again, in terms of how you access it, it seems to be via the Hub. 

With the current design, you can also do everything you want to for Metrics via the Hub as well:
https://github.com/getsentry/sentry-dotnet/blob/2a41c48c2185d910be95bbe850b8979a1a3c571b/src/Sentry/Internal/Hub.cs#L24

That's because all of the functionality related to metrics is pretty much encapsulated in the MetricAggregator... the only exception being the functionality to send those metrics to Sentry (which obviously should be a SentryClient thing). 

We could have the MetricAggregator take a hard dependency on IHub (so that you couldn't send metrics without a Hub)... At the moment, you could technically use Metrics and SentryClient independently (without any Hub)... although with this PR the `StartTimer` method introduces a dependency on the `SentrySdk.CurrentHub` so we're already starting to break that I guess.

src/Sentry/SentryClient.cs
```diff
@@ -471,14 +453,12 @@ public void Dispose()
 
         try
         {
-            Metrics.FlushAsync().ContinueWith(_ =>
-                // Worker should empty it's queue until SentryOptions.ShutdownTimeout
-                Worker.FlushAsync(_options.ShutdownTimeout)
-            ).ConfigureAwait(false).GetAwaiter().GetResult();
+            // Worker should empty it's queue until SentryOptions.ShutdownTimeout
```
bitsandfoxes, line 456:
```suggestion
            // Worker should empty its queue until SentryOptions.ShutdownTimeout
```

src/Sentry/Internal/Hub.cs
```diff
@@ -486,6 +494,52 @@ public void CaptureTransaction(SentryTransaction transaction, Scope? scope, Hint
...
+        {
+            if (enumerable is null)
+            {
+                _options.LogError(e, "Failure to enumerate metrics for capture");
+            }
+            else
+            {
+                foreach (var metric in enumerable)
+                {
+                    _options.LogError(e, "Failure to capture metric: {0}", metric.EventId);
```
jamescrosswell, line 520: Maybe better to log a single AggregateException?

src/Sentry/Protocol/Metrics/CounterMetric.cs
```diff
@@ -26,7 +26,7 @@ public CounterMetric()
     protected override void WriteValues(Utf8JsonWriter writer, IDiagnosticLogger? logger) =>
         writer.WriteNumber("value", Value);
 
-    protected override IEnumerable<IConvertible> SerializedStatsdValues()
+    protected override IEnumerable<object> SerializedStatsdValues()
```
jamescrosswell, line 29: IConvertible isn't  CLS Compliant so it can't be in any public signatures for libraries targeting netstandard 2.x.

src/Sentry/Protocol/Metrics/Metric.cs
```diff
@@ -73,7 +114,7 @@ public async Task SerializeAsync(Stream stream, IDiagnosticLogger? logger, Cance
 
         foreach (var value in SerializedStatsdValues())
         {
-            await Write($":{value.ToString(CultureInfo.InvariantCulture)}");
+            await Write($":{((IConvertible)value).ToString(CultureInfo.InvariantCulture)}");
```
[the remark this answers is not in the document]
jamescrosswell: If user's implemented a custom Metric instrument and tried to serlize this, using values that don't implement `IConvertible` then they could break it. I can't see users doing that though, since Sentry.io doesn't support custom Metric instruments... there's be no reason to create such a class. 

Ideally we'd use `IConvertible` in the interface but it's not CLS Compliant.

src/Sentry/Internal/Hub.cs
```diff
@@ -486,6 +494,52 @@ public void CaptureTransaction(SentryTransaction transaction, Scope? scope, Hint
...
+            return;
+        }
+
+        try
+        {
+            _ownedClient.CaptureCodeLocations(codeLocations);
+        }
+        catch (Exception e)
+        {
+            _options.LogError(e, "Failure to capture code locations: {0}", codeLocations.Timestamp);
```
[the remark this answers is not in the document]
jamescrosswell: Only in so much as the timestamp identifies the codelocations bucket... but yeah, realistically, you'd need a debugger open to have any hope of knowing what's going on here. We could probably remove the timestamp from the log then.

src/Sentry/DisabledMetricAggregator.cs
```diff
@@ -37,6 +37,22 @@ internal class DisabledMetricAggregator : IMetricAggregator
...
+            // No Op
+        }
+    }
+
+    public IDisposable StartTimer(string key, MeasurementUnit.Duration unit = MeasurementUnit.Duration.Second,
+        IDictionary<string, string>? tags = null,
+        int stackLevel = 1)
+    {
+        // No Op
+        return new NoOpDisposable();
```
bruno-garcia, line 53: we can have a singleton and avoid allocations

src/Sentry/ISentryClient.cs
```diff
@@ -52,6 +54,16 @@ public interface ISentryClient
     [EditorBrowsable(EditorBrowsableState.Never)]
     void CaptureTransaction(SentryTransaction transaction, Scope? scope, Hint? hint);
 
+    /// <summary>
+    /// Captures one or more metrics to be sent to Sentry.
+    /// </summary>
+    void CaptureMetrics(IEnumerable<Metric> metrics);
```
bruno-garcia, line 60: could we get around these public methods by making the envelope outside (on hub) and just calling capture envelopes?

src/Sentry/Protocol/Metrics/CounterMetric.cs
bruno-garcia: We could drop CLS compliance on 4.0. But I'm not sure IConvertible is really used anymore anyway is it?

src/Sentry/Protocol/Metrics/Metric.cs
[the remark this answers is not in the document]
bruno-garcia: It feels like a pitfall. Auers will for sure return random types here and break it. Can we define a type ourselves to represent this?

src/Sentry/ISentryClient.cs
jamescrosswell: Simple but cunning... why did I not think of that 😌

src/Sentry/IMetricHub.cs
bitsandfoxes, line 1: That looks like the thing with `IHubEx` all over again. Couldn't `MetricsAggregator` and `Hub` communicate via `Lazy<Envelope>` or somesuch instead? And the hub just passes that down to the client?

jamescrosswell: Bruno suggested we consider making SentryClient.CaptureEnvelope public and adding it to the ISentryClient interface. This method is public already in the other SDKs (like Java).

CHANGELOG.md
```diff
@@ -12,6 +12,10 @@ If you have conflicts, you can opt-out by adding the following to your `csproj`:
 </PropertyGroup>
 ```
 
+### Features
```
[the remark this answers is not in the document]
jamescrosswell: We're not... the non-CLS compliant types aren't used in our API interfaces anywhere.

PR #3079, opened 2024-01-23
fix: vsnprintf usage on macOS

vaind:
`vsnprintf` was used incorrectly on macOS (used the windows version) because this code was copied from sentry-unity where it didn't run on mac, only Windows and Linux.


jamescrosswell (changes requested):
Could we get some context in the PR summary for what this change is about? Does it relate to an Issue? What was the problem and how does this solve it?

vaind:
says it all in the PR title, really, but OK, I'll add the same to the description

jamescrosswell:
> says it all in the PR title, really, but OK, I'll add the same to the description

That's helpful. I wasn't aware that vsnprintf was windows only (I've never heard of it before) or that the code we had was copied from Unity. So the whole PR was just mysterious to me.

jamescrosswell (approved):

vaind:
> . I wasn't aware that vsnprintf was windows only

It's not, but we have a windows-only variant defined in the class (IIRC it has different arguments), and since the code switched on linux vs rest of the world (which was just Windows in Unity), macOS was trying to use that windows variant too.

Issue #2275, opened 2023-03-30
Improve guidance for ASP.NET Core


mattjohnsonpint:
### Problem Statement

The [ASP.NET Core documentation page](https://docs.sentry.io/platforms/dotnet/guides/aspnetcore/) could be much clearer.

So could the samples.

### Solution Brainstorm

- [x] Lead with the current version of ASP.NET Core.
- [x] Add sections for older versions, either collapsed or in tabs.
- [x] Reduce the list of features so it doesn't take up so much of the first page.
- [x] Make sure `EnableTracing` and `UseSentryTracing` are visible in all examples.
- [x] Make sure its clear what order `UseSentryTracing` goes in the middleware (after `UseRouting`, but why and what if that's not present?)
- [ ] Update samples to use the current ASP.NET Core.  Start from templates `dotnet new web` / `webapi` / `mvc`, etc.
  - No need to show every feature of Sentry.   Just show what's needed to get started in ASP.NET Core.
  - [ ] Sentry.Samples.AspNetCore.Basic
  - [ ] Sentry.Samples.AspNetCore.Blazor.Server
  - [ ] Sentry.Samples.AspNetCore.Blazor.Wasm
  - [ ] Sentry.Samples.AspNetCore.Grpc
  - [x] https://github.com/getsentry/sentry-dotnet/pull/3083
  - [ ] Sentry.Samples.AspNetCore.Serilog
  - [ ] An ASP.NET Core AOT sample - [New project template](https://learn.microsoft.com/en-us/aspnet/core/release-notes/aspnetcore-8.0?view=aspnetcore-8.0#new-project-template)
- [ ] Add a few limited samples for older versions of ASP.NET Core, but only where drastically different from current.

jamescrosswell:
Closing this issue as it was always a bit of an epic and lots of it has been done already. If there's specific stuff we still want to do, we can create separate issues for these.

PR #3083, opened 2024-01-25
Updated ASP.NET Core MVC sample to use the latest dotnet template

jamescrosswell:
First of various sample updates for https://github.com/getsentry/sentry-dotnet/issues/2275

#skip-changelog


bitsandfoxes (approved):
I'm kind of torn here. It's not like the "old samples" don't apply anymore, just, that there is a "new" way of doing it, right? So if I'm a user and I'm working on something older I might not even be familiar with what's going on here.
On the other hand, if I call `dotnew new _whatevertemplate_` to try out Sentry then this is what I get..

jamescrosswell:
> I call `dotnew new _whatevertemplate_` to try out Sentry then this is what I get..

Yeah exactly... I think that's what we need to optimise for. If a new user comes along and wants to see how Sentry works, they could go to our docs which would say "create a new project and then do this". At the moment, if you do that using the latest IDE, it can be quite confusing since what you end up with doesn't look like our sample. 

We Do need to give guidance on how to integrate Sentry with legacy projects (which we also do in the docs).

Possibly we could create additional samples (one for each version of ASP.NET Core)... it does take a while to compile all of these when building the solution though so 🤷🏻

PR #3085, opened 2024-01-26
GitHubActions(build job): relax branch restrictions

knocte:
Restricting CI to certain branch prevents contributors to know the CI status of their contributions (e.g. the result from building and running unit tests after pushing their work to their fork) before they propose a PullRequest; because normally people create branches when they want to start to work on something (instead of having to use the 'main' branch of their fork).

#skip-changelog

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -3,8 +3,7 @@ name: build
 on:
   push:
     branches:
-      - main
-      - release/*
+      - '**'
   pull_request:
     paths-ignore:
       - "**.md"


bitsandfoxes:
Hey @knocte, thank you for opening this up. Unfortunately, I don't think this will work tho. The workflows right now need to be approved before running. So even if the restrictions are relaxed I don't think it would change anything?

jamescrosswell:
@knocte what you could do is create a branch in your own fork of Sentry, rather than working in the main branch. You could then create a PR to merge into main in your fork (which would trigger CI on your own GitHub CI instances). Once you were happy, merge into main and then create a PR to the upstream repo (`getsentry/sentry-dotnet`).

knocte:
> Unfortunately, I don't think this will work tho. The workflows right now need to be approved before running. So even if the restrictions are relaxed I don't think it would change anything?

It will. The approval is only needed when PR is created. But before it is created, a contributor can push to his own fork (with the 'push' event, not 'pull_request') and runners won't be restricted.

> what you could do is create a branch in your own fork of Sentry, rather than working in the main branch. You could then create a PR to merge into main in your fork (which would trigger CI on your own GitHub CI instances). Once you were happy, merge into main and then create a PR to the upstream repo (getsentry/sentry-dotnet).

@jamescrosswell but that's not usually how you contribute to opensource projects. If you did that, you could only have one PR open at a time instead of more than one. Just check in any opensource project in GitHub that has external contributors (so, people that open PRs from their forks, not from branches of the repo like you guys do), you will see that most PRs come from branches that are not called master or main. It is a good practice to name your branch with something related to the work you're performing.

jamescrosswell:
Whilst at first glance, I can't a problem with this... we don't want CI to run multiple times for the same code. CI builds are costly (particularly the macos agents and the device tests) so we should avoid running these unecessarily. 

Reading through the docs for GitHub Actions, it looks like that might be the case:
> [Using multiple events](https://docs.github.com/en/actions/using-workflows/triggering-a-workflow#using-multiple-events)
You can specify a single event or multiple events. For example, a workflow with the following on value will run when a push is made to any branch in the repository or when someone forks the repository:
> 
> on: [push, fork]
If you specify multiple events, only one of those events needs to occur to trigger your workflow. If multiple triggering events for your workflow occur at the same time, multiple workflow runs will be triggered.

The Pull Request might have a different SHA in the event of a merge conflict, but otherwise I think we'd end up running additional/redundant workflows on GitHub Actions if we did this.

For that reason, I don't think I'm in favour of this change. If people really want/need to run CI on their own GitHub CI instances prior to making a PR to the upstream repo (which I think is a bit of an edge case... not many people are setting up their own CI infrastructure to build the dotnet client), [the above workaround/hack](https://github.com/getsentry/sentry-dotnet/pull/3085#issuecomment-1914364538) is available.

knocte:
> CI builds are costly (particularly the macos agents and the device tests) 

If someone forks your repo and pushes to his fork, they would be using GitHubActions' free CI agents, not your agents. Public repositories have unlimited CI on GitHub, so what exactly do you mean by the word "costly" here?

antonpirker:
The way I do it is: I create my branch and work on it, when I have *something* I create a draft pull request. This signals that this is work in progress. But whenever I update my branch, the CI is triggered. I guess this should also work for forks of the repo (Maybe someone from Sentry needs to trigger the CI runs thou. I am in the Sentry org and working on Python SDK, so I am not sure how this is exactly setup in .NET)

bruno-garcia:
It doesn't really matter since we're not pushing a bunch of branches all the time that would affect our CI runs so I'd say it's OK to merge this

knocte:
> Another example of why: one of the build tasks is to cancel previous CI builds if one is already running on a non-release branch... so we only have one CI build at a time per branch. That would mean when we push and then create a pull request, we'd start a CI run (for the push), then cancel it and start another one (for the pull)... which just results in a bunch of notifications to delete on every PR.

Mmm, sorry, not sure I understood well your workflow here, but it sounds like the `concurrency` feature of GitHubActions could help with that: https://docs.github.com/en/actions/using-jobs/using-concurrency

jamescrosswell (approved):

PR #3092, opened 2024-01-29
feat: `Metrics.Set` now accepts `string` as value

bitsandfoxes:
From the updated dev docs: https://develop.sentry.dev/sdk/metrics/#sets `Set should accept strings as value`.
With this PR we're vendoring in part of [force's Crc32.NET](https://github.com/force-net/Crc32.NET) implementation because there is not anything built-in that I could find.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,10 @@ If you have conflicts, you can opt-out by adding the following to your `csproj`:
 </PropertyGroup>
 ```
 
+### Features
+
+- `SentrySdk.Metrics.Set` now additionally accepts `string` as value ([#3092](https://github.com/getsentry/sentry-dotnet/pull/3092))
+
 ### Fixes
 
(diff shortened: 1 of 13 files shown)


bruno-garcia (approved):

jamescrosswell:
@bitsandfoxes did you see [System.IO.Hashing.Crc32](https://learn.microsoft.com/en-us/dotnet/api/system.io.hashing.crc32?view=dotnet-plat-ext-8.0)?

jamescrosswell (approved):

bitsandfoxes:
> @bitsandfoxes did you see [System.IO.Hashing.Crc32](https://learn.microsoft.com/en-us/dotnet/api/system.io.hashing.crc32?view=dotnet-plat-ext-8.0)?
> 
> I guess that comes with a dependency on the [System.IO.Hashing NuGet package](https://www.nuget.org/packages/System.IO.Hashing/)...

Yes. Not wanting to take on the dependency is the reason why we're vendoring it in.

Issue #734, opened 2021-01-13
Investigate counters


bruno-garcia:
https://devblogs.microsoft.com/dotnet/net-5-new-networking-improvements/#telemetry
https://github.com/dotnet/runtime/issues/37428#issuecomment-674597327

bruno-garcia:
This will be great with DDM!

jamescrosswell:
@bitsandfoxes I think this one is going on the back burner for a little bit so we could change the priority. I think most of the research has been done (see attached PR)... just some Development grunt work to get this over the line once Sentry's core metrics product is a bit more mature. 

In the mean time, we already support OpenTelemetry / System.Diagnostics.Metrics instruments.

jamescrosswell:
@Flash0ver is this still relevant do you think (in the context of the new metrics product)?

The Spike might be worth looking at:
- https://github.com/getsentry/sentry-dotnet/pull/3100

PR #3100, opened 2024-01-31
Spike: Capture well known EventSource counters

jamescrosswell:
Resolves https://github.com/getsentry/sentry-dotnet/issues/734

EventSource based metrics are a bit different to the OpenTelemetry metrics (and Sentry Metrics), so some consideration needs to be given to the API. 

Initially what I've built is quite naive. You can define one or more EventSourceMatcher objects that are used to define which EventSources should be captured by our listener. The matchers consist of a `SubstringOrRegEx` expression that matches against the `EventSource.Name` and also a couple of other filters (for `Level` and something called `EventKeywords`, which is a flags enum) that will be applied to the listener as well): 
https://github.com/getsentry/sentry-dotnet/blob/bbde463fb25a5aaf80210eba64362f020b28f051/src/Sentry/Internal/SystemDiagnosticsEventSourceListener.cs#L58-L62

The `Arguments.Value` supplied above configures events to be captured every 10 seconds, which aligns with our bucketing strategy... this would be one way to ensure we don't overcount metrics which are already pre-aggregated in the EventCounter.

## Challenges

So far, so good. Where our implementation is a bit naive however is when we report the metrics to Sentry. This happens in the `OnEventWritten` method of our listener:
https://github.com/getsentry/sentry-dotnet/blob/bbde463fb25a5aaf80210eba64362f020b28f051/src/Sentry/Internal/SystemDiagnosticsEventSourceListener.cs#L76-L96

There are a couple of challenges with this implementation.

Firstly, it assumes every Counter will be an incrementing counter and that the increment count will always be `1`. This is rarely the case. 

Secondly, most of the interesting data reported by counters is sent in the [EventWrittenEventArgs.Payload property](https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.tracing.eventwritteneventargs.payload?view=net-8.0#system-diagnostics-tracing-eventwritteneventargs-payload)... this is unfortunately a `ReadOnlyCollection<object?>?` so, yeah, not strongly typed and specific to each EventCounter.

## Potential Solution

Whilst our current implementation of the `OnEventWritten` method might work for some EventCounters, other counters will need to be handled differently. Effectively we have one "EventCounterProcessor" that could be registered to process certain counters for which it was appropriate, but we really need to allow SDK users to supply their own EventCounterProcessors (to handle whatever counters they might want to create themselves) and we should probably implement various built in processors to handle all of the (useful) [Well Known counters](https://learn.microsoft.com/en-us/dotnet/core/diagnostics/available-counters).

For example, the code below gives a feel for how the `System.Net.Sockets[bytes-sent]` Eventcounter might be parsed (this code isn't very resilient but demonstrates how we'd need to pull info out of the Payload):
```csharp
    protected override void OnEventWritten(EventWrittenEventArgs eventData)
    {
        if (eventData.EventName.Equals("EventCounters"))
        {
            foreach (var payload in eventData.Payload)
            {
                var counterPayload = (IDictionary<string, object>)payload;
                var name = (string)counterPayload["Name"];
                if (name == "bytes-sent")
                {
                    Console.WriteLine($"{name}: {counterPayload["Mean"]}");
                }
            }
        }
    }
```

That particular counter stores a bunch of stuff as an `IDictionary<string, object>` in the payload... with a bit of grunt work we could build some logic to pull all the useful info out and report it to Sentry... we could then encapsulate that logic in a processor which would (hopefully) be able to be reused with other EventCounters that stored their payloads similarly (and differed only in name). 

# Alternatives

It looks like there's an [OpenTelemetry wrapper around counters](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/tree/main/src/OpenTelemetry.Instrumentation.EventCounters)... perhaps we could leverage this?


jamescrosswell:
Discussed with @bitsandfoxes and we agreed to put this on pause for a bit until the Metrics capability in Sentry itself is a bit more mature.

PR #3103, opened 2024-01-31
add alpha version suffix to profiling package

bruno-garcia:
#skip-changelog


jamescrosswell (approved):

bruno-garcia:
I made some misstake here because it didn't work:

image


cc @bitsandfoxes 

I guess at this point doesn't matter, better add it to the docs the feature is beta

PR #3120, opened 2024-02-06
chore: Fix flaky android device tests

bitsandfoxes:
Hack to see if it affects the flake.
Based on https://github.com/ReactiveCircus/android-emulator-runner/issues/362 the emulator getting stuck while shutting down might be due to jobs running in parallel. 
Let's see if running in sequence helps. Otherwise we might just disable AVD caching for the time being.

#skip-changelog

### DIFF
--- a/.github/workflows/device-tests-android.yml
+++ b/.github/workflows/device-tests-android.yml
@@ -78,30 +78,30 @@ jobs:
 
       # Cached AVD setup per https://github.com/ReactiveCircus/android-emulator-runner/blob/main/README.md
 
-      - name: AVD cache
-        uses: actions/cache@v4
-        id: avd-cache
-        with:
-          path: |
-            ~/.android/avd/*
-            ~/.android/adb*
-          key: ${{ runner.os }}-avd-api${{ matrix.api-level }}
+      # - name: AVD cache
+      #   uses: actions/cache@v4
+      #   id: avd-cache
+      #   with:
+      #     path: |
+      #       ~/.android/avd/*
+      #       ~/.android/adb*
+      #     key: ${{ runner.os }}-avd-api${{ matrix.api-level }}
 
-      - name: Create AVD and generate snapshot for caching
-        if: steps.avd-cache.outputs.cache-hit != 'true'
-        timeout-minutes: 30
-        uses: reactivecircus/android-emulator-runner@6b0df4b0efb23bb0ec63d881db79aefbc976e4b2 # pin@v3
-        with:
-          api-level: ${{ matrix.api-level }}
-          # We don't need the Google APIs, but the default images are not available for 32+
-          target: ${{ matrix.api-level >= 32 && 'google_apis' || 'default' }}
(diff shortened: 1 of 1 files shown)


.github/workflows/device-tests-android.yml
```diff
@@ -80,21 +76,68 @@ jobs:
...
+          disk-size: 4096M
+          emulator-options: -no-window -accel on -gpu swiftshader_indirect -noaudio -no-boot-anim -camera-back none
+          disable-animations: false
+          script: echo "Generated AVD snapshot for caching."
+
+      - name: Run Tests
+        timeout-minutes: 30
+        uses: reactivecircus/android-emulator-runner@6b0df4b0efb23bb0ec63d881db79aefbc976e4b2 # pin@v3
+        with:
+          api-level: ${{ env.API-LEVEL-MIN }}
```
bruno-garcia, line 106: We'll run it for 2 API levels in sequence now? That'll slow down a lot the build. 

Are you sure it's due to parallelism? I don't get it why this would be

vaind: Makes zero sense to me either. strategy-based jobs are completely different VMs, they don't influence one another in any implicit way.

vaind: Sorry I didn't click the link. if it's due to caching, have you tried disabling it? Does it actually do much in terms of job runtime?

bitsandfoxes: Doing it right now.

bitsandfoxes: Sorry, this should have been a draft. I was trying to compare disabling AVD caching and putting them in sequence. And now I'm not sure what's going on. They run just as fast without caching?

bitsandfoxes: Alright, the 27 like to take its time.

bitsandfoxes: Let's go with disabling caching (takes 2 minutes longer) and we'll track enabling it in a separate issue

.github/workflows/device-tests-android.yml
```diff
@@ -78,30 +78,30 @@ jobs:
...
+      #     api-level: ${{ matrix.api-level }}
+      #     # We don't need the Google APIs, but the default images are not available for 32+
+      #     target: ${{ matrix.api-level >= 32 && 'google_apis' || 'default' }}
+      #     force-avd-creation: false
+      #     ram-size: 2048M
+      #     arch: x86_64
+      #     disk-size: 4096M
+      #     emulator-options: -no-window -accel on -gpu swiftshader_indirect -noaudio -no-boot-anim -camera-back none
+      #     disable-animations: false
+      #     script: echo "Generated AVD snapshot for caching."
```
bruno-garcia, line 104: it's easy to revert based on history, lets delete instead of dead-code-it
```suggestion
```

bruno-garcia (approved):

Issue #2239, opened 2023-03-16
Support for Crons on Hangfire 


bruno-garcia:
### Problem Statement

Crons is in Beta: https://docs.sentry.io/product/crons/

Other SDKs:
* Celery beat PR: https://github.com/getsentry/sentry-python/pull/1935
* Laravel: https://github.com/getsentry/sentry-laravel/pull/659

### Solution Brainstorm

Spec is pending, this can be on hold until then.

@kanadaj already created something for Hangfire (and Quartz) that can be brought in or used as inspiration: https://github.com/kanadaj/Sentry.Monitor

mattjohnsonpint:
Hangfire is mostly about fire-and-forget background tasks.  We'd be mostly interested their recurrent tasks feature.
https://docs.hangfire.io/en/latest/background-methods/performing-recurrent-tasks.html

We should also look at [Quartz.NET](https://www.quartz-scheduler.net/), which is primarily aimed at scheduling.

Also, there are several cloud-based schedulers offered by Azure, AWS & GCP - but I think those should be broader initiatives, not specific for .NET.

mattjohnsonpint:
Also, this has already been started externally, by @kanadaj at https://github.com/kanadaj/Sentry.Monitor

bruno-garcia:
I use it as a recurring job runner on: https://github.com/dotnet/nuget-trends and plan to dogfood there

bruno-garcia:
Deleted Quartz from the title. Lets scope this to Hangfire only. if there's demand for Quartz we can create another issue and track separately.

kanadaj:
@bruno-garcia My implementation already has both Quartz and Hangfire and seems to work fine unless the API changes for some reason.

bruno-garcia:
> @bruno-garcia My implementation already has both Quartz and Hangfire and seems to work fine unless the API changes for some reason.

Edited the title to note that your code has both and we can talk about moving it in to this repo 🙏 Thanks!

kanadaj:
For recordkeeping, if we want to make these more generic libraries (Sentry.Hangfire and Sentry.Quartz) we'd probably want to also add facilities for tracing the jobs. This could be done for non-recurring jobs too.

bruno-garcia:
Related to Crons (but not Hangfire, an alternative): https://www.nuget.org/packages/ncrontab

pinkpawkitty:
[...]

PR #3128, opened 2024-02-07
feat: Crons support (Hangfire)

bitsandfoxes:
Resolves https://github.com/getsentry/sentry-dotnet/issues/2239

This PR accomplishes the following things:
- It adds the `SentrySdk.CaptureCheckIn` API for manual capture
- It adds a Hangfire integration

The Hangfire integration works as follows:
It adds an extension to the hangfire config allowing the SDK to add its filter for all jobs.
```
builder.Services.AddHangfire(configuration => configuration
    .UseSentry() // <- Add Sentry to automatically send check-ins
);
```

For the integration to work it requires a `monitor slug` that needs to be present. The monitor slug can be added via our own HangfireAttribute on the job running in the background.
```
public class SomeJob
{
    [SentryMonitorSlug("first-job")]
    public void Execute() { /* do something */ }
}
```

with no changes required to the call into Hangfire
```
BackgroundJob.Enqueue<SomeJob>(job => job.Execute());
```


src/Sentry.Hangfire/Sentry.Hangfire.csproj
```diff
@@ -0,0 +1,24 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <Description>Official Hangfire integration for Sentry - Open-source error tracking that helps developers monitor and fix crashes in real time.</Description>
+    <PackageTags>$(PackageTags);OpenTelemetry</PackageTags>
```
bruno-garcia, line 5:
```suggestion
    <PackageTags>$(PackageTags);Hangfire</PackageTags>
```

src/Sentry.Hangfire/Sentry.Hangfire.csproj
```diff
@@ -0,0 +1,24 @@
...
+    <TargetFrameworks>net8.0;net6.0;netstandard2.1;netstandard2.0;net462</TargetFrameworks>
```
bruno-garcia, line 6: We can always add things later, how about we go with only this to start:
```suggestion
    <TargetFrameworks>net8.0;net6.0;net462</TargetFrameworks>
```

src/Sentry.Hangfire/Sentry.Hangfire.csproj
```diff
@@ -0,0 +1,24 @@
...
+  </PropertyGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="..\Sentry\Sentry.csproj" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Hangfire.Core" Version="1.8.7" />
+    <PackageReference Include="Hangfire.MemoryStorage" Version="1.8.0" />
+    <PackageReference Include="Hangfire.AspNetCore" Version="1.8.7" />
```
bruno-garcia, line 21: If this is going to depend on their ASP.NET Core package, we can even remove net462 because it's not going to work on .NET Framework.

Sucks to add more packages but ideally we'd have `Sentry.Hangfire` and `Sentry.Hangfire.AspNetCore` the latter depending on Sentry.AspNetCore if that helps with anything (like extension method `AddHangfire` on `SentryAspNetCoreOptions` or something to make it easier to add

src/Sentry/SentryCheckIn.cs
```diff
@@ -0,0 +1,87 @@
+using Sentry.Extensibility;
+using Sentry.Internal.Extensions;
+
+namespace Sentry;
+
+/// <summary>
+/// Checkin metadata.
+/// </summary>
+public interface ICheckin
```
bruno-garcia, line 9: Do we need an interface for Check-ins?

src/Sentry/SentryCheckIn.cs
```diff
@@ -0,0 +1,87 @@
...
+    /// The Checkin errored
+    /// </summary>
+    Error
+}
+
+/// <summary>
+/// Sentry Checkin
+/// </summary>
+// https://develop.sentry.dev/sdk/check-ins/
+public class SentryCheckIn : ICheckin, ISentryJsonSerializable
```
bruno-garcia, line 52: The casing isn't aligned between the class and the interface. In Java they capitalized: `CheckIn`: https://github.com/getsentry/sentry-java/blob/a537f8a45828b751272940034c4a64ee5f8fd278/sentry/src/main/java/io/sentry/CheckIn.java#L14

Should we align?

src/Sentry/SentryCheckIn.cs
```diff
@@ -0,0 +1,87 @@
...
+    }
+
+    /// <inheritdoc />
+    public void WriteTo(Utf8JsonWriter writer, IDiagnosticLogger? logger)
+    {
+        writer.WriteStartObject();
+
+        writer.WriteSerializable("check_in_id", Id, logger);
+        writer.WriteString("monitor_slug", MonitorSlug);
+        writer.WriteString("status", Status.ToString().ToSnakeCase());
```
bruno-garcia, line 83: instead of paying for 2 allocations on `Status.ToString().ToSnakeCase()` on each use how about a switch enum -> string already in the write format? no allocations

```suggestion
        writer.WriteString("status", Status.ToString().ToSnakeCase());
```

src/Sentry.Hangfire/Sentry.Hangfire.csproj
bruno-garcia: This was resolved but we ended up with both netstandard TFMs, is that by design?

src/Sentry.Hangfire/SentryJobFilter.cs
```diff
@@ -0,0 +1,87 @@
...
+
+    private readonly IHub _hub;
+    private readonly IDiagnosticLogger? _logger;
+
+    public SentryJobFilter() : this(() => HubAdapter.Instance)
+    { }
+
+    internal SentryJobFilter(Func<IHub> hubAccessor)
+    {
+        _hub = hubAccessor.Invoke();
```
bruno-garcia, line 19: why do we need the accessor?

Originally it was added because the hub could change (e.g: aspnetcore double init). so we called on the accessor on each use to make sure we were on the latest Hub instance

this patterns seems to defeat that purpose

src/Sentry.Hangfire/SentryJobFilter.cs
```diff
@@ -0,0 +1,87 @@
...
+        OnPerformedInternal(context, GetJobParameter, HasException);
+    }
+
+    internal void OnPerformingInternal(PerformingContext? context,
+        Func<PerformContext?, string, object?> getJobParameterFunc,
+        Action<PerformContext?, string, SentryId> setJobParameter)
+    {
+        if (getJobParameterFunc(context, SentryMonitorSlugKey) is not string monitorSlug)
+        {
+            var jobName = context?.BackgroundJob.Job.ToString();
```
bruno-garcia, line 39: Does Job have a Name property or something? Relying on `ToString` here is less optimal if that's the case becauase it could change

src/Sentry.Hangfire/SentryJobFilter.cs
```diff
@@ -0,0 +1,87 @@
...
+        Func<PerformedContext?, bool> hasExceptionFunc)
+    {
+        if (getJobParameterFunc(context, SentryMonitorSlugKey) is not string monitorSlug)
+        {
+            return;
+        }
+
+        if (getJobParameterFunc(context, SentryCheckInIdKey) is not SentryId checkInId)
+        {
+            return;
```
bruno-garcia, line 61: are these situations expected? Or a Debug log might be useful to find something misconfigured?

src/Sentry.Hangfire/SentryJobFilter.cs
```diff
@@ -0,0 +1,87 @@
...
+        {
+            status = CheckInStatus.Error;
+        }
+
+        _ = _hub.CaptureCheckIn(new SentryCheckIn(monitorSlug, status, checkInId));
+    }
+
+    private static object? GetJobParameter(PerformContext? context, string key)
+    {
+        return context?.GetJobParameter<object>(key);
```
bruno-garcia, line 75: are these all type object? or are we boxing here potentially? (e.g: `SentryId` is a struct, no?)

test/Sentry.Hangfire.Tests/Sentry.Hangfire.Tests.csproj
```diff
@@ -0,0 +1,16 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFrameworks>net8.0;net7.0;net6.0;net48</TargetFrameworks>
```
bruno-garcia, line 4: do these match what we're targeting in the lib? like net6 is there to cover ns2.1, but ns2.0 has no coverage?

src/Sentry.Hangfire/SentryJobFilter.cs
bitsandfoxes: `Job` has a `type` and `method` and `ToString` concatenates this to `Type.Method`. 
Users opt-in to have their jobs monitored when they add `.UseSentry`. If the job does not have a monitor slug the warning tells them which specific type and method is missing the attribute.

src/Sentry.Hangfire/SentryJobFilter.cs
bitsandfoxes: We would run into those if the `SentryMonitorSlug` attribute has not been set. We log this type of misconfiguration when the job gets started. Imo, doing it here too would add no benefit.

bruno-garcia: if it's guaranteed to go on both places, no value

src/Sentry.Hangfire/SentryJobFilter.cs
bruno-garcia: Not sure I follow. My point is that we're relying on ToString where it might be safer to rely on a different property which indicates why it exists. ToString can become a longer thing that includes other stuff, later for other reasons

src/Sentry.Hangfire/SentryServerFilter.cs
```diff
@@ -0,0 +1,57 @@
...
+        {
+            var jobType = context.BackgroundJob.Job.Type;
+            var jobMethod = context.BackgroundJob.Job.Method;
+            _logger?.LogWarning("Skipping creating a check-in for '{0}.{1}'. " +
+                                "Failed to find Monitor Slug the job. You can set the monitor slug " +
+                                "by setting the 'SentryMonitorSlug' attribute.", jobType, jobMethod);
+            return;
+        }
+
+        var checkInId = _hub.CaptureCheckIn(new SentryCheckIn(monitorSlug, CheckInStatus.InProgress));
```
bruno-garcia, line 36: APIs where we need to do `method(new Something()` feel janky, can we have an overload that takes the two arguments directly?

test/Sentry.Hangfire.Tests/HangfireTests.cs
```diff
@@ -0,0 +1,82 @@
...
+
+    [Fact]
+    public async void ExecuteJobWithAttribute_CapturesCheckInInProgressAndOk()
+    {
+        var sentryId = SentryId.Create();
+        _fixture.Hub.CaptureCheckIn(Arg.Any<SentryCheckIn>()).Returns(sentryId);
+
+        BackgroundJob.Enqueue<TestJob>(job => job.ExecuteJobWithAttribute());
+
+        await Task.Delay(100);
```
bruno-garcia, line 23: can we rewrite this without any delay? relying on this stuff causes tests to be flaky.
Ideally you share a signal and wait on one end and signal from the other.

You could signal from the `CaptureCheckIn` method even, and wait below on the signal with a timeout so if the test fails it doesn't hang forever:

example; https://github.com/getsentry/sentry-dotnet/blob/c19fb14d5c2a56cbe162f1ef5e30af4a8149b5fa/test/Sentry.Tests/Integrations/AppDomainAdapterTests.cs#L14-L23

test/Sentry.Hangfire.Tests/HangfireTests.cs
```diff
@@ -0,0 +1,82 @@
...
+
+    [Fact]
+    public async void ExecuteJobWithException_CapturesCheckInInProgressAndError()
+    {
+        var sentryId = SentryId.Create();
+        _fixture.Hub.CaptureCheckIn(Arg.Any<SentryCheckIn>()).Returns(sentryId);
+
+        BackgroundJob.Enqueue<TestJob>(job => job.ExecuteJobWithException());
+
+        await Task.Delay(1000);
```
bruno-garcia, line 42: 😭

test/Sentry.Hangfire.Tests/HangfireTests.cs
bitsandfoxes: Yessir!

samples/Sentry.Samples.Hangfire/Program.cs
```diff
@@ -0,0 +1,91 @@
...
+    {
+        Console.WriteLine($"Starting to do some heavy work at: {DateTime.Now}");
+        Task.Delay(1000).Wait();
+        Console.WriteLine($"Finished doing some heavy work at: {DateTime.Now}");
+    }
+}
+
+public class SecondJob
+{
+    [SentryMonitorSlug("job-that-throws")]
```
bruno-garcia, line 72: can we use the method name by default? so adding this attribute becomes optional?

src/Sentry.Hangfire/SentryServerFilter.cs
bruno-garcia: this was resolved but it's still doing `new CheckIn` ?

src/Sentry/Extensibility/DisabledHub.cs
```diff
@@ -180,6 +180,17 @@ public void CaptureSession(SessionUpdate sessionUpdate)
     {
     }
 
+    /// <summary>
+    /// No-Op
+    /// </summary>
+    public SentryId CaptureCheckIn(SentryCheckIn checkIn) => SentryId.Empty;
```
bruno-garcia, line 186: do we need this method?

test/Sentry.Hangfire.Tests/Sentry.Hangfire.Tests.csproj
```diff
@@ -0,0 +1,17 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFrameworks>net8.0;net6.0</TargetFrameworks>
```
bruno-garcia, line 4: Not testing the .NET Framework build?

bruno-garcia (commented):
Only thing I'd consider 'blocking' is agreeing if we need 2 public APIs instead of only 1 (with 3 arguments)

src/Sentry.Hangfire/GlobalConfigurationExtensions.cs
```diff
@@ -0,0 +1,34 @@
...
+{
+    /// <summary>
+    /// Uses Sentry
+    /// </summary>
+    /// <param name="configuration"></param>
+    /// <returns></returns>
+    public static IGlobalConfiguration UseSentry(this IGlobalConfiguration configuration)
+    {
+        configuration.UseFilter(new SentryServerFilter());
+        return configuration;
```
jamescrosswell, line 19: It might be better to chain these constructors by using `: this(null, null)` here. Otherwise every time you make changes to one, you have to remember to make equivallent changes to the other, which is quite dangerous when one of them is used for testing and the other for production. 

It means you'd need the `hub` parameter on the overload nullable and maybe do something like:
```csharp
hub ??= HubAdapter.Instance;
```

Similarly, IDiagnosticLogger would have to be nullable but the `SentryServerFilter` is able to deal with that anyway.

jamescrosswell (approved):
LGTM - just a couple of comments.

test/Sentry.Hangfire.Tests/Sentry.Hangfire.Tests.csproj
bitsandfoxes: I forgot to add the explainer: Hangfire is not strong naming their assemblies and CI chokes on Windows on net48. Due to the viral nature and the use of `InternalsVisibleTo` I could not get it to opt out.

samples/Sentry.Samples.Hangfire/Program.cs
bitsandfoxes: That would cause the SDK to send a `CheckIn` with every Hangfire job. Is this something we'd be okay with?

Issue #3127, opened 2024-02-07
No service for type 'Sentry.IHub' has been registered


jamescrosswell:
When configuring Sentry with the OpenTelemetry integration, our extension method tries to resolve `IHub` from DI:
https://github.com/getsentry/sentry-dotnet/blob/91724e3d076a691c72299c5c972880872c21d0b1/src/Sentry.OpenTelemetry/TracerProviderBuilderExtensions.cs#L44

This doesn't work if Sentry has been initialised via `SentrySdk.Init` as that initialization method doesn't register `IHub` or anything else with DI. 

Even running our own sample `samples/Sentry.Samples.OpenTelemetry.Console/Sentry.Samples.OpenTelemetry.Console.csproj` will fail 😢

PR #3129, opened 2024-02-07
Fix: No service for type 'Sentry.IHub' has been registered

jamescrosswell:
Resolves https://github.com/getsentry/sentry-dotnet/issues/3127


bitsandfoxes (approved):
We don't have any integration tests, do we? I feel like we should start having some.

PR #3133, opened 2024-02-08
Added OpenTelemetry.Instrumentation.Runtime metrics to built in metrics

jamescrosswell:
I noticed the Aspire starter template includes these metrics so we should probably make them part of our built in metrics also.

There is a bit of [documentation available for these here](https://opentelemetry.io/docs/languages/net/automatic/instrumentations/#metrics-instrumentations).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ### Features
 
-- Added OpenTelemetry.Instrumentation.Runtime to built in metrics ([#3133](https://github.com/getsentry/sentry-dotnet/pull/3133))
+- The SDK now automatically collects metrics coming from `OpenTelemetry.Instrumentation.Runtime` ([#3133](https://github.com/getsentry/sentry-dotnet/pull/3133))
 
 ### Fixes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Features
+
+- Added OpenTelemetry.Instrumentation.Runtime to built in metrics ([#3133](https://github.com/getsentry/sentry-dotnet/pull/3133))
+
 ### Fixes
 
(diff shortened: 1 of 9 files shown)


samples/Sentry.Samples.OpenTelemetry.AspNetCore/Program.cs
```diff
@@ -21,10 +32,15 @@
 
 builder.WebHost.UseSentry(options =>
 {
-    //options.Dsn = "...Your DSN...";
+    options.Dsn = "...Your DSN...";
```
jamescrosswell, line 35:
```suggestion
    // options.Dsn = "...Your DSN...";
```

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Features
+
+- Added OpenTelemetry.Instrumentation.Runtime to built in metrics ([#3133](https://github.com/getsentry/sentry-dotnet/pull/3133))
```
bitsandfoxes, line 7: Can we make that more.. spicy/explicit? What is it users will now get OOTB?

bitsandfoxes:
```suggestion
- The SDK now automatically collects metrics coming from `OpenTelemetry.Instrumentation.Runtime` ([#3133](https://github.com/getsentry/sentry-dotnet/pull/3133))
```
Something like that?

bitsandfoxes (approved):

CHANGELOG.md
jamescrosswell: It's [these guys here](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/Instrumentation.Runtime-1.7.0/src/OpenTelemetry.Instrumentation.Process/README.md#metrics)... although you do have to do this as well (so it only works if they've configured OpenTelemetry to collect/emit these metrics in the first place):
https://github.com/getsentry/sentry-dotnet/blob/88729d097dee078c4100bef203112258e4fd9a16/samples/Sentry.Samples.OpenTelemetry.AspNetCore/Program.cs#L17

PR #3138, opened 2024-02-09
feat: PowerShell support

vaind:
This brings initial integration for PowerShell.

closes #3117

### DIFF
--- a/src/Sentry.PowerShell/assemblies-loader.ps1
+++ b/src/Sentry.PowerShell/assemblies-loader.ps1
@@ -1,0 +1,57 @@
+$dir = Join-Path (Split-Path -Parent $MyInvocation.MyCommand.Path) 'lib'
+
+# TODO move libs to this package and remove the following code
+$dir = Join-Path $dir '../../Sentry/bin/Debug'
+
+function GetTFM
+{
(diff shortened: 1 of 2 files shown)


src/Sentry.PowerShell/assemblies-loader.ps1
```diff
@@ -0,0 +1,57 @@
...
+    }
+    elseif ($PSVersionTable.PSVersion -ge '7.2')
+    {
+        return 'net6.0'
+    }
+    elseif ($PSVersionTable.PSVersion -ge '6.0')
+    {
+        return 'netstandard2.0'
+    }
+    else
```
bruno-garcia, line 29: should this be a guard to a specific min version and `else` just error out?

vaind:
Closed in favor of https://github.com/getsentry/sentry-powershell

src/Sentry.PowerShell/assemblies-loader.ps1
vaind: I was thinking of letting it try to load the lib since checking the actual available version of .NET framework when running Windows Powershell doesn't seem to be so trivial.

Issue #3136, opened 2024-02-09
Migration guide: How to disable Sentry when no DSN supplied


jamescrosswell:
## Problem

A couple of our customers have a setup where they enable Sentry by configuring the DSN in an environment variable. When this environment variable is not set, however, they want Sentry to be disabled. 

This is tricky to do since the configuration files take precedence... then code, then environment variables. So if a disabled DSN is configured in XML/JSON/Code then it can't be conditionally overwritten by an environment variable only on those machines where customers want Sentry enabled. 

However they can get Sentry to do what they want by configuring the Disabled DSN in an assembly attribute... which is the very last place that our `SettingsLocator` checks (even after having checked environment variables):
https://github.com/getsentry/sentry-dotnet/blob/0aed96926720b59f20c6c3e781dce68df99c9e27/src/Sentry/Internal/SettingLocator.cs#L33-L47

So you can cadd a file called `AssembyInfo.cs` to your project with the following contents:

```csharp
using Sentry;

[assembly: Dsn(SentryConstants.DisableSdkDsnValue)]
```

That way you end up with the same behaviour as you had in 3.x (which is that if no DSN is provided anywhere, you get a disabled DSN by default).

_Originally posted by @jamescrosswell in https://github.com/getsentry/sentry-dotnet/issues/3134#issuecomment-1935459702_
            
## TODO

We should summarize the solution above and add it to the 4.0.0 migration guide.

ericsampson:
that's kinda awkward.

What about also introducing a new setting and env var like `SENTRY_ENABLED` instead of messing around with fragile in-band signaling via DSN

jamescrosswell:
> What about also introducing a new setting and env var like `SENTRY_ENABLED` instead of messing around with fragile in-band signaling via DSN

It is already possible to enable/disable sentry by setting the `SENTRY_DSN` environment variable to either `""` (i.e. disabled DSN) or `some_valid_dsn`. For people getting started with sentry-dotnet 4.x, this is probably the right way to go.

This suggested addition to the migration guide is really for folks who want to simulate the behaviour of the 3.x SDK which was that if no DSN at all was provided (e.g. if no `SENTRY_DSN` environment variable was present and that was the only way the DSN was ever configured) then Sentry would be disabled.

ericsampson:
hmm ok, that's just different than the long-standing Unix convention of "environment variables that unset or blank should result in the same behavior"

[...]

Issue #3140, opened 2024-02-10
Should config values from Environment variables override other settings?


jamescrosswell:
@slonopotamus [suggested in another issue](https://github.com/getsentry/sentry-dotnet/issues/3134#issuecomment-1936099357) that setting the DSN in an environment variable should override an empty string (i.e. disabled) value that has been set via other bindings (like appsettings.json). 

This would be more consistent with [how things are done in ASP.NET Core](https://learn.microsoft.com/en-us/aspnet/core/fundamentals/configuration/?view=aspnetcore-8.0#non-prefixed-environment-variables):

> Using the [default](https://learn.microsoft.com/en-us/aspnet/core/fundamentals/configuration/?view=aspnetcore-8.0#default) configuration, the [EnvironmentVariablesConfigurationProvider](https://learn.microsoft.com/en-us/dotnet/api/microsoft.extensions.configuration.environmentvariables.environmentvariablesconfigurationprovider) loads configuration from environment variable key-value pairs after reading appsettings.json, appsettings.{Environment}.json, and [user secrets](https://learn.microsoft.com/en-us/aspnet/core/security/app-secrets?view=aspnetcore-8.0). Therefore, key values read from the environment override values read from appsettings.json, appsettings.{Environment}.json, and user secrets.

I think if we do this, it shouldn't be just for the DSN and shouldn't be just for the disabled DSN value... but that env variables trump other provider sources more generally. 

However this is likely a breaking changes as we may have many customers who have their production environments configured based on the current behaviour.

bitsandfoxes:
See https://github.com/getsentry/sentry-dotnet/issues/3143#issuecomment-1970955309

PR #2655, opened 2023-09-26
Throw ArgumentNullException if DSN is null

jamescrosswell:
Completes https://github.com/getsentry/sentry-dotnet/issues/2494

## Migration notes

In Sentry.NET SDK 3.x.x and below, it was possible to initialise Sentry with a `DisabledHub` by setting the DSN to either `String.Empty` or `null`. However this sometimes tripped up people who were new to Sentry and/or had forgotten to configure the DSN in their Sentry options. 

As such, starting with version 4.x.x of the SDK, if you want to configure a `DisabledHub` then you **_have to_** set the DSN to `String.Empty`. If you ommit the DSN altogether or forget to supply this, Sentry will throw an exception during initialization to let you know you've made a mistake.

PR #3147, opened 2024-02-13
fix: Allow a DSN `string.Empty` to be overwritten by the environment

bitsandfoxes:
## TLDR;
This PR aims to let users control the SDK via environment variables by overwriting the DSN set to `string.Empty`.
`This "" behavior is 10 years old and set in all SDks.` https://github.com/getsentry/sentry-dotnet/issues/2494#issuecomment-1722098718

## What do?
Setting `options.Dsn = "";` disables the SDK except for when a DSN is present in the environment. In that case the DSN from the environment gets used.

## Context
[We've introduced](https://github.com/getsentry/sentry-dotnet/pull/2655/files) some friction by letting the SDK throw if there was no DSN provided during initialization. The idea was to make the onboarding process more explicit and let new users know immediately that there was something wrong with their current configuration. https://github.com/getsentry/sentry-dotnet/issues/2494

The change made it very awkward to control the SDK's behavior through the environment since now not setting the DSN via code implicitly gets treated as a configuration error.

The current order of priority:
- Code
- Binding through config files
- Environment
- AssemblyAttribute

2. and 3. are now following the convention of letting the environment overwrite config files but that's OK for now. We can [track](https://github.com/getsentry/sentry-dotnet/issues/3140) and document this.

Fixes https://github.com/getsentry/sentry-dotnet/issues/3136,

### DIFF
--- a/src/Sentry/Internal/SettingLocator.cs
+++ b/src/Sentry/Internal/SettingLocator.cs
@@ -1,3 +1,4 @@
+using Sentry.Extensibility;
 using Sentry.Internal.Extensions;
 
 namespace Sentry.Internal;
@@ -35,14 +36,31 @@ internal class SettingLocator
         // For DSN only
         // An empty string set on the option should NOT be converted to null because it is used
         // to indicate the the SDK is disabled.
-        _options.Dsn ??= GetEnvironmentVariable(Constants.DsnEnvironmentVariable)
-                         ?? AssemblyForAttributes?.GetCustomAttribute<DsnAttribute>()?.Dsn;
+
+        // If the DSN has not been set
         if (_options.Dsn is null)
         {
-            throw new ArgumentNullException("You must supply a DSN to use Sentry." +
-                                            "To disable Sentry, pass an empty string: \"\"." +
-                                            "See https://docs.sentry.io/platforms/dotnet/configuration/options/#dsn");
+            _options.Dsn = GetEnvironmentVariable(Constants.DsnEnvironmentVariable)
+                           ?? AssemblyForAttributes?.GetCustomAttribute<DsnAttribute>()?.Dsn;
+
+            if (_options.Dsn is null)
+            {
+                throw new ArgumentNullException("You must supply a DSN to use Sentry." +
+                                                "To disable Sentry, pass an empty string: \"\"." +
+                                                "See https://docs.sentry.io/platforms/dotnet/configuration/options/#dsn");
+            }
         }
+        else if (_options.Dsn.Equals(string.Empty))
+        {
+            var dsn = GetEnvironmentVariable(Constants.DsnEnvironmentVariable)
+                           ?? AssemblyForAttributes?.GetCustomAttribute<DsnAttribute>()?.Dsn;
+
+            if (dsn is not null)
+            {
+                _options.Dsn = dsn;
+            }
+        }
+
         return _options.Dsn;
     }
 
--- a/src/Sentry/Internal/SettingLocator.cs
+++ b/src/Sentry/Internal/SettingLocator.cs
@@ -34,10 +34,9 @@ internal class SettingLocator
     public string GetDsn()
     {
         // For DSN only
-        // An empty string set on the option should NOT be converted to null because it is used
-        // to indicate the the SDK is disabled.
 
-        // If the DSN has not been set
+        // If the DSN has not been set at all require it to be on either the environment or an AssemblyAttribute. If
+        // neither can be found: throw
         if (_options.Dsn is null)
         {
             _options.Dsn = GetEnvironmentVariable(Constants.DsnEnvironmentVariable)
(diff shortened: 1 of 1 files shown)


src/Sentry/Internal/SettingLocator.cs
```diff
@@ -33,16 +33,34 @@ public SettingLocator(SentryOptions options)
     public string GetDsn()
     {
         // For DSN only
-        // An empty string set on the option should NOT be converted to null because it is used
-        // to indicate the the SDK is disabled.
-        _options.Dsn ??= GetEnvironmentVariable(Constants.DsnEnvironmentVariable)
```
slonopotamus, line 38: I'd just do

```cs
if (String.IsNullOrEmpty(_options.Dsn))
{
    _options.Dsn = GetEnvironmentVariable(Constants.DsnEnvironmentVariable)
                         ?? AssemblyForAttributes?.GetCustomAttribute<DsnAttribute>()?.Dsn;
}
```

bitsandfoxes: Yeah, I thought about that.
But if we want to keep the "fix" for the "Hey, I'm a new user, I put `.UseSentry` somewhere but nothing happens. What do?` experience, then we need to check for those separately. Otherwise, it'll still throw if you explicitly set `options.Dsn = string.Empty` and there is nothing in the environment.

slonopotamus: I'm not sure I follow you. null will remain null if neither env nor assembly attribute was set.

bruno-garcia (approved):
with these tests passing I believe we have a good logic to deal with this 
other than nitpicking on the code itself this LGTM

jamescrosswell:
Broadly I'm in favour of this change, as it brings us a little closer to aligning with the Microsoft priorities but just be aware that this is going to be a change in the behaviour of the SDK. 

Prevsiously, when using version 3.x or version 4.x of the SDK, if you configured a disabled DSN in appsettings.json, this would not be overwritten by any environment variable configuration (nulls would, but disabled wouldn't - that's the whole point of this change). 

After this change, a disabled DSN configuration via appsettings.json will be overwritten by environment variable configuration. 

I have no idea how many customers (if any) were relying on the assumption that what they configured in appsettings.json would not be overwritten by environment variables. Just a risk to be aware of.

slonopotamus:
You're _already_ different from 3.x. This PR makes code different in another direction that is supposed to fix currently broken usecases. And yes, it will involve adjustments on user side (we need to pass empty string).

jamescrosswell:
> You're _already_ different from 3.x. This PR makes code different in another direction that is supposed to fix currently broken usecases. And yes, it will involve adjustments on user side (we need to pass empty string to `UseSentry`).

That's true. The other workaround I'd proposed (relying on assembly attributes) would not be breaking things for customers who were relying on them being able to configure a valid or disabled DSN in appsettings that would not be overwritten by env variables though. This PR would break things for those customers. 

As I say, I'm not sure how many (if any) customers have their CI/CD setup that way, but it's a risk to be aware of. We could be inadvertedly breaking things for a bunch of customers here.

slonopotamus:
> We could be inadvertedly breaking things for a bunch of customers here.

Hah. I am an opinionated person here, because this PR will fix _my_ workflow :D But I believe that cases "we have `SENTRY_DSN` defined in env, but expect it to be ignored" are nonexistent.

src/Sentry/Internal/SettingLocator.cs
```diff
@@ -33,17 +33,32 @@ public SettingLocator(SentryOptions options)
...
         }
-        return _options.Dsn;
+
+        // Overwriting the `string.Empty` with the DSN found in the environment
+        if (dsn is not null)
+        {
+            _options.Dsn = dsn;
+        }
+
+        return _options.Dsn!;
```
bruno-garcia, line 61: We're `!` here so lets be clear this can't happen here:

```suggestion
        Debug.Assert(_options.Dsn != null, "Dsn can't be null at this point based on the rules above");
        return _options.Dsn!;
```

bruno-garcia (approved):

bruno-garcia:
> Hah. I have a bias here, because this PR will fix my workflow :D But I believe that cases "we have SENTRY_DSN defined in env, but expect it to be ignored" are nonexistent.

Yeah env var should trump all other values and it's how a lot of software works so I feel OK about this change

jamescrosswell:
> Yeah env var should trump all other values and it's how a lot of software works so I feel OK about this change

Do we just want to do it for the DSN property, or should we make this change instead (it would render this PR irrelevant)?
- https://github.com/getsentry/sentry-dotnet/issues/3143

bruno-garcia:
> > Yeah env var should trump all other values and it's how a lot of software works so I feel OK about this change
> 
> Do we just want to do it for the DSN property, or should we make this change instead (it would render this PR irrelevant)?
> 
>     * [Align configuration binding precedence with Microsoft priorities #3143](https://github.com/getsentry/sentry-dotnet/issues/3143)

makes sense to change the behavior across all env vars

ericsampson:
thanks for reconsidering the original direction all, it's an appreciated sign of customer focus <3

Maybe you can partner with some poweruser companies/people who would be a "beta test" partner group? I'd be happy to, just LMK. If you gave me a heads-up when there are design questions or a beta version to check out, then I'd be happy to give it a whirl and provide feedback. As long as it was at like 3 or 4  weeks before the planned SDK GA, to give time to schedule the change/deploy n our side.

Cheers

Issue #3155, opened 2024-02-19
OpenTelemetry integration throws when Sentry is disabled


jamescrosswell:
`GetSentryOptions` currently returns null if Sentry is disabled:

https://github.com/getsentry/sentry-dotnet/blob/1007ba32ec5efb62ad107ec1cc0b3cb3bcca167c/src/Sentry/SentryClientExtensions.cs#L99-L106

This causes issues with the `SentrySpanProcessor` when using the OpenTelemetry integration... that span processor assumes that if `SentryOptions == null` then the Sentry DSN is also null.

PR #3156, opened 2024-02-19
OpenTelemetry integration throws an exception when Sentry is disabled

jamescrosswell:
Fixes https://github.com/getsentry/sentry-dotnet/issues/3155

Ultimately I've fixed this in the `TraceProviderBuilderExtensions`:
https://github.com/getsentry/sentry-dotnet/blob/3b0918797367e84cc006923e3765857849129b48/src/Sentry.OpenTelemetry/TracerProviderBuilderExtensions.cs#L45-L47

I also refactored those slighly for testability.

### Note

We didn't pick up on this because we configure the DSN via an attribute in our sample projects... so even if you tried to force the DSN to be disabled when using our Otel sample, it would use a Valid DSN 😢 

https://github.com/getsentry/sentry-dotnet/blob/6f037bd94f10b094b8beab3ae0799228d3b92af3/samples/Directory.Build.props#L25-L29

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@
 ### Fixes
 
 - Metric unit names are now sanitized correctly. This was preventing some built in metrics from showing in the Sentry dashboard ([#3151](https://github.com/getsentry/sentry-dotnet/pull/3151))
-- The Sentry OpenTelemetry integration threw an exception when Sentry was disabled ([#3156](https://github.com/getsentry/sentry-dotnet/pull/3156))
+- The Sentry OpenTelemetry integration no longer throws an exception with the SDK disabled ([#3156](https://github.com/getsentry/sentry-dotnet/pull/3156))
 
 ## 4.1.1
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Fixes
 
 - Metric unit names are now sanitized correctly. This was preventing some built in metrics from showing in the Sentry dashboard ([#3151](https://github.com/getsentry/sentry-dotnet/pull/3151))
+- The Sentry OpenTelemetry integration threw an exception when Sentry was disabled ([#3156](https://github.com/getsentry/sentry-dotnet/pull/3156))
 
 ## 4.1.1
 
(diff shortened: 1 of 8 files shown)


bitsandfoxes:
> We didn't pick up on this because we configure the DSN via an attribute in our sample projects...

Maybe it's time to remove the attribute again? I wasn't even aware we were doing this until last week and ran headfirst into a wall with the assumption that we don't.

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 ### Fixes
 
 - Metric unit names are now sanitized correctly. This was preventing some built in metrics from showing in the Sentry dashboard ([#3151](https://github.com/getsentry/sentry-dotnet/pull/3151))
+- The Sentry OpenTelemetry integration threw an exception when Sentry was disabled ([#3156](https://github.com/getsentry/sentry-dotnet/pull/3156))
```
bitsandfoxes, line 8:
```suggestion
- The Sentry OpenTelemetry integration no longer throws an exception with the SDK disabled ([#3156](https://github.com/getsentry/sentry-dotnet/pull/3156))
```

bitsandfoxes (dismissed):
Thanks for the fix!

src/Sentry.OpenTelemetry/DisabledSpanProcessor.cs
```diff
@@ -0,0 +1,24 @@
...
+namespace Sentry.OpenTelemetry;
+
+internal class DisabledSpanProcessor : BaseProcessor<Activity>
+{
+    private static readonly Lazy<DisabledSpanProcessor> LazyInstance = new();
+    internal static DisabledSpanProcessor Instance => LazyInstance.Value;
+
+    public DisabledSpanProcessor()
+    {
+        Console.WriteLine("Sentry is disabled so no OpenTelemetry spans will be sent to Sentry.");
```
bitsandfoxes, line 12: I think this should go through a diagnostic logger?

jamescrosswell: > I think this should go through a diagnostic logger?

The DisabledSpanProcessor only gets used if Sentry is disabled, so none of the Sentry logging will work... Do you think there's a system diagnostic logging class that would be more useful than Console.WriteLine?

bitsandfoxes (approved):

PR #3176, opened 2024-02-27
Fix: replace invalid metric tag values with an empty string instead of _

markushi:
[According to our spec](https://develop.sentry.dev/sdk/metrics/#normalization) invalid tag values should be replaced with no string.

> Tag Values: regex [^\w\d_:/@\.{}\[\]$-]+ with no replacement character.

I cross checked with [PHP](https://github.com/getsentry/sentry-php/blob/6f8cce5441da92829e84c3c2549625860f11787f/src/Serializer/EnvelopItems/MetricsItem.php#L59) and [JS](https://github.com/getsentry/sentry-javascript/blob/2a8ef4094ff283a207d3522db3e4c3357d73bb30/packages/core/src/metrics/utils.ts#L65): both of them also use no replacement character.

Relevant Python PR: https://github.com/getsentry/sentry-python/pull/2773

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Fixes
+
+- Empty strings are used instead of underscores to replace invalid metric tag values ([#3176](https://github.com/getsentry/sentry-dotnet/pull/3176))
+
 ### Dependencies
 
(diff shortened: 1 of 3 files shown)


jamescrosswell:
Thanks @markushi !

bitsandfoxes (approved):

bruno-garcia:
this is higher chance to cause conflict no?

abc&&a
abc*a
abc::::a

all become abca?

Issue #3178, opened 2024-02-29
Could not load file or assembly Microsoft.Bcl.AsyncInterfaces when building Sentry on Windows


jamescrosswell:
When building Sentry-CI-Build-Windows.slnf on a clean copy of the sentry-dotnet respository, 4 of the unit tests for Log4Net fail with: 
```
System.IO.FileNotFoundException : Could not load file or assembly 'Microsoft.Bcl.AsyncInterfaces, Version=6.0.0.0, Culture=neutral, PublicKeyToken=cc7b13ffcd2ddd51'. The system cannot find the file specified.
```

I believe this is because `CachingTransport` implements [IAsyncDisposable](https://learn.microsoft.com/en-us/dotnet/api/system.iasyncdisposable?view=net-8.0), which is only supported in netstandard2.1 and later, where Log4Net targets netstandard2.0.

It's a bit odd, since the tests pass for the net8.0 and netfx targets... but fail for net6.0 and net7.0.

PR #3179, opened 2024-02-29
Removed IAsyncDisposable from CachingTransport

jamescrosswell:
Resolves https://github.com/getsentry/sentry-dotnet/issues/3178

#skip-changelog

### DIFF
--- a/src/Sentry/Internal/Http/CachingTransport.cs
+++ b/src/Sentry/Internal/Http/CachingTransport.cs
@@ -4,7 +4,17 @@ using Sentry.Protocol.Envelopes;
 
 namespace Sentry.Internal.Http;
 
-internal class CachingTransport : ITransport, IAsyncDisposable, IDisposable
+/// <summary>
+/// A transport that caches envelopes to disk and sends them in the background.
+/// </summary>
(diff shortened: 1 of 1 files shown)


bitsandfoxes:
I'm not sure I get what's going on there. Is this due to us dropping ns2.0? Is removing the interface the right move here, especially since it seems to work on different TFWs? Can we `#if` around this if this is a target framework issue?

jamescrosswell:
> I'm not sure I get what's going on there. Is this due to us dropping ns2.0? Is removing the interface the right move here, especially since it seems to work on different TFWs? Can we `#if` around this if this is a target framework issue?

I'm not sure what's going on either... technically that interface should be included in net5.0 and later. The tests fail on net6.0 and net7.0. I think it's something to do with the version of `Microsoft.Bcl.AsyncInterfaces` that is expected. 

However, we don't use or need that interface. If we drop it, it has no impact. If we keep it, our tests fail on a Windows clean build. I figured in the interests of simplicity then, we just drop it. It gives us no benefits and conversely it creates problems we don't need to solve.

bitsandfoxes (approved):

Issue #3175, opened 2024-02-27
Metrics Improvement


bitsandfoxes:
The following parts have been added to the dev docs since the initial implementation
- [x] [Trace Seeking](https://develop.sentry.dev/sdk/metrics/#trace-seeking) - Attach Tags
- [x] [Span Aggregation](https://develop.sentry.dev/sdk/metrics/#span-aggregation)

jamescrosswell:
> [x] Timings don't need to create a span themselves. Which means the whole `IMetricHub` can probably go?

@bitsandfoxes are you sure about this? The [reference implementation creates spans for Timings](https://github.com/getsentry/sentry-python/blob/0901953c93071e858f4da67c1e864766ae19c002/sentry_sdk/metrics.py#L800) and the RFC still contains this:

>  Most of the metrics APIs are span seeking which means that they record a measurement in relation to that span. Some APIs (such as metrics.timing when used with a code block) will instead create a span and bind it.

bitsandfoxes:
Looks like I missed that. So spans stay!

PR #3191, opened 2024-03-04
Metrics improvements

jamescrosswell:
Resolves https://github.com/getsentry/sentry-dotnet/issues/3175

## Span Aggregates

Emitted metrics are now displayed for spans (e.g. [here](https://sentry-sdks.sentry.io/performance/aspnet-core:ad14c6a8526040db98d3d16ee117b3e6/?project=6601807&query=&referrer=performance-transaction-summary&statsPeriod=24h&transaction=GET+%2Fhello&unselectedSeries=p100%28%29&unselectedSeries=avg%28%29)):

image

## Automatic Tags 

Tags are set on metrics for Release, Environment and Transaction (where these are present):

image


src/Sentry/SentrySpan.cs
```diff
@@ -104,6 +105,10 @@ public SentrySpan(ISpan tracer)
         {
             _measurements = spanTracer.InternalMeasurements?.ToDict();
             _tags = spanTracer.InternalTags?.ToDict();
+            if (spanTracer.HasMetrics)
+            {
+                _metricsSummary = spanTracer.MetricsSummary;
+            }
```
bitsandfoxes, line 111: So the child inherits the summary from the parent?

src/Sentry/Protocol/Metrics/LocalAggregate.cs
```diff
@@ -0,0 +1,54 @@
+using Sentry.Extensibility;
+using Sentry.Internal.Extensions;
+
+namespace Sentry.Protocol.Metrics;
+
+/// <summary>
+/// Gauges track a value that can go up and down.
+/// </summary>
+internal class LocalAggregate
```
bitsandfoxes, line 9: I know it is internal but can we get `Metric` in the name somewhere?

src/Sentry/SpanTracer.cs
```diff
@@ -13,6 +13,10 @@ public class SpanTracer : ISpan
 
     internal TransactionTracer Transaction { get; }
 
+    private readonly Lazy<LocalAggregator> _metricsSummary = new();
```
bitsandfoxes, line 16: Thanks for the `lazy`.

src/Sentry/LocalAggregator.cs
```diff
@@ -0,0 +1,71 @@
+using Sentry.Extensibility;
+using Sentry.Protocol.Metrics;
+
+namespace Sentry;
+
+internal class LocalAggregator : ISentryJsonSerializable
```
bitsandfoxes, line 6: I know it is internal but can we get `Metric` in the name somewhere?

src/Sentry/SentrySpan.cs
jamescrosswell: Not quite... I think I've mixed responsibilities here a little. 

I've refactored this now, splitting the LocalAggregator into two separate classes with separate responsibilities: 
1. MetricsSummaryAggregator to track totals
2. MetricsSummary for serialization across the wire

This makes the serialization logic easier to understand as well since the way the metrics are stored either has to be optimized to be easy to mutate or easy to serialize (no easy way to have both, due to the requirement to group by exportKey for serialization).

jamescrosswell: > So the child inherits the summary from the parent?

btw That mirrors the difference between SpanTracer and SentrySpan now as well (the first is mutable, designed to be used when actively tracking a span and the second is an immutable DTO for serialization).

src/Sentry/Protocol/Metrics/MetricsSummary.cs
```diff
@@ -0,0 +1,47 @@
...
+
+    public void WriteTo(Utf8JsonWriter writer, IDiagnosticLogger? logger)
+    {
+        writer.WriteStartObject();
+
+        foreach (var (exportKey, value) in _measurements)
+        {
+            writer.WritePropertyName(exportKey);
+            writer.WriteStartArray();
+            foreach (var metric in value.OrderBy(x => MetricHelper.GetMetricBucketKey(x.MetricType, x.Key, x.Unit, x.Tags)))
```
jamescrosswell, line 38: btw: I'd love to get rid of the OrderBy here. This only exists to ensure the order of things in JSON stays exactly the same, so our Verify tests pass. If I could work out how to have Verify do a Deep object comparison rather than a string compare, we could get rid of the OrderBy here and the ToImmutableSortedDictionary call above.

src/Sentry/SentrySpan.cs
bitsandfoxes: Oh nice. The split makes it way easier to wrap my head around it. Thanks!

bitsandfoxes (approved):
:shipit: LGTM!
The `IMetricsHub` still makes me feel itchy.

PR #3195, opened 2024-03-05
chore: merge SentryOptionsExtensions to SentryOptions

vaind:
In PowerShell, extension classes are just standard classes with static methods, so you had to do

`[Sentry.SentryOptionsExtensions]::AddEventProcessor($options, [EventUpdater]::new())`

with this PR, you can now:

`$options.AddEventProcessor( [EventUpdater]::new())`


src/Sentry/SentryOptions.cs
```diff
@@ -1292,17 +1293,321 @@ public SentryOptions()
         );
     }
 
-    internal void AddIntegration(ISdkIntegration integration)
```
jamescrosswell, line 1295: Is this related to the change from Extensions to ordinary class methods?

jamescrosswell: Ah, I see... there was an equivalent public extension method 😵‍💫

jamescrosswell (approved):

src/Sentry/SentryOptions.cs
vaind: yes indeed. the method was already there on the extension class as public and then again here as internal 🤷

Issue #3197, opened 2024-03-06
Http.HttpRequestOut OpenTelemetry spans not finished if filtered


jamescrosswell:
# Problem

For background context see:
- https://github.com/getsentry/sentry-dotnet/discussions/3166#discussioncomment-8669832

If OpenTelemetry is configured with filtering - e.g. as follows:
```csharp
AddHttpClientInstrumentation(o => o.FilterHttpRequestMessage = _ => false)
```
... then our SentrySpanProcessor receives `OnStart` events for any outgoing requests from HttpClient but never receives a corresponding `OnEnd` event. 

This leaves a bunch of `SentrySpan` instances accumulating in `SpanProcessor._map`, which never get cleaned up. 

## Investigation

- Created a [discussion in the OpenTelemetry repo](https://github.com/open-telemetry/opentelemetry-dotnet/discussions/5415)

PR #3198, opened 2024-03-06
Fix memory leak when OpenTelemetry spans get filtered

jamescrosswell:
Fixes https://github.com/getsentry/sentry-dotnet/issues/3197

## Solution

For the Sentry spans we hold in `SentrySpanProcessor._map`, we need a way to detect when the associated Activities have been filtered out. The OpenTelemetry SDK doesn't provide a mechanism for I had to get a bit creative. 

Previously we only held on to the Activity.Id for OpenTelemetry spans that we'd processed and there's no way to lookup an Activity by Id. 

With this PR we now retain weak references to the Activities themselves (not just the Activity.Id)... and we periodically loop through each of the spans in our `_map` to check if the associated Activities have been filtered out. When Activities get filtered, `IsAllDataRequested` gets set and the `ActivityTraceFlags.Recorded` flag gets removed ([see code](https://github.com/open-telemetry/opentelemetry-dotnet/blob/5784b45c4e310fc54d84278891bb22d6b93a1246/src/OpenTelemetry.Instrumentation.Http/Implementation/HttpHandlerDiagnosticListener.cs#L149-L150)). When that happens, we know we're not going to receive anymore OnStart/OnEnd events relating to the Activity, so we can remove it from our map.

Additionally, when finishing transactions that were instrumented using OpenTelemetry, we now surgically remove any spans that were filtered... since although those spans were removed from the `_map` they still exist in the `SentryTransaction.Spans`. 

## Testing Manually

To test this manually, [this line in our sample project](https://github.com/getsentry/sentry-dotnet/blob/088557ec20c2478578a6d1642d2bb82b7fc1a3e2/samples/Sentry.Samples.OpenTelemetry.AspNetCore/Program.cs#L23) can be changed to:
```csharp
            .AddHttpClientInstrumentation(o => o.FilterHttpRequestMessage = _ => false)
```

That basically filters every outgoing HTTP request. You can then see in resulting traces that these have magically disappeared (there's a suspicious gap in the timings, but there's no trace event for these). 

Before this PR, this was not the case... we were seeing:
a) A bunch of unfinished spans relating to outgoing HTTP requests that were filtered by OTEL
b) An accumulation of those spans in our SentrySpanProcessor, resulting in ever increasing memory consumption

### DIFF
--- a/src/Sentry.OpenTelemetry/SentrySpanProcessor.cs
+++ b/src/Sentry.OpenTelemetry/SentrySpanProcessor.cs
@@ -88,6 +88,9 @@ public class SentrySpanProcessor : BaseProcessor<Activity>
 
             var span = (SpanTracer)parentSpan.StartChild(context);
             span.StartTimestamp = data.StartTimeUtc;
+            // Used to filter out spans that are not recorded when finishing a transaction.
+            span.SetFused(data);
+            span.IsFiltered = () => span.GetFused<Activity>() is { IsAllDataRequested: false, Recorded: false };
             _map[data.SpanId] = span;
         }
         else
@@ -113,7 +116,19 @@ public class SentrySpanProcessor : BaseProcessor<Activity>
                 );
             transaction.StartTimestamp = data.StartTimeUtc;
             _hub.ConfigureScope(scope => scope.Transaction = transaction);
+            transaction.SetFused(data);
             _map[data.SpanId] = transaction;
+        }
+
+        // Clean up items that may have been filtered out (OnEnd never gets called for these but IsRecorded flips)
+        foreach (var mappedItem in _map)
+        {
+            var (spanId, span) = mappedItem;
+            var activity = span.GetFused<Activity>();
+            if (activity is { Recorded: false, IsAllDataRequested: false })
+            {
+                _map.TryRemove(spanId, out _);
+            }
         }
     }
 
--- a/src/Sentry.OpenTelemetry/SentrySpanProcessor.cs
+++ b/src/Sentry.OpenTelemetry/SentrySpanProcessor.cs
@@ -19,9 +19,8 @@ public class SentrySpanProcessor : BaseProcessor<Activity>
     private readonly SentryOptions? _options;
     private readonly Lazy<IDictionary<string, object>> _resourceAttributes;
 
-    private static readonly TimeSpan PruningInterval = TimeSpan.FromSeconds(5);
-    internal DateTimeOffset _lastPruned = DateTimeOffset.MinValue;
-    private readonly object _pruningLock = new();
+    private static readonly long PruningInterval = TimeSpan.FromSeconds(5).Ticks;
+    internal long _lastPruned = 0;
 
     /// <summary>
     /// Constructs a <see cref="SentrySpanProcessor"/>.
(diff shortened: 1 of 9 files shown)


src/Sentry.OpenTelemetry/SentrySpanProcessor.cs
```diff
@@ -196,6 +207,44 @@ public override void OnEnd(Activity data)
...
+            if (activity is { Recorded: false, IsAllDataRequested: false })
+            {
+                _map.TryRemove(spanId, out _);
+            }
+        }
+    }
+
+    private bool NeedsPruning()
+    {
+        lock (_pruningLock)
```
bruno-garcia, line 239: This reads as potential contention code and since we're just doing some path with the date is there a way we can do something atomic here instead with Interlocked?

bruno-garcia: suggestion: unix timestamp as seconds and interlocked.compareExchange

jamescrosswell: I've replaced lock with some rough equivalent using Interlocked... In the Benchmarks I ran, I wasn't able to detect a significant performance difference between the two, but that might have been a shortcoming in the Benchmark itself. I'll have another crack at it as intuitively the two should behave quite differently.

jamescrosswell: OK, I put together some new benchmarks... interestingly, `lock` actually comes out a bit faster than either `Interlocked` or `SemaphoreSlim`. The difference isn't wild and I'm not sure how much I trust these - I'm worried there's some error in my design of the benchmark tests themselves. [Code here](https://github.com/jamescrosswell/LockBenchmark/blob/main/BenchmarkDotNet.Artifacts/results/LockBenchmark.LockBenchmarks-report-github.md) however.

bitsandfoxes (approved):
Looks like we're not alone in this https://github.com/getsentry/sentry-python/issues/2722
It might be worth checking back how it gets solved there.

jamescrosswell:
> Looks like we're not alone in this https://github.com/getsentry/sentry-python/issues/2722
> It might be worth checking back how it gets solved there.

Interesting... But this fix doesn't affect the public API so no need to align with the Python SDK, I don't think 

Be good to review and merge ASAP as it's quite a serious issue for anyone using filtering with OTEL.

src/Sentry.OpenTelemetry/SentrySpanProcessor.cs
bruno-garcia: > interestingly, lock actually comes out a bit faster than either Interlocked or SemaphoreSlim. 

I really don't trust the benchmark then :)

bruno-garcia: the link 404's to me

bruno-garcia (approved):
If it fixes the problem lets go with it, possibly lock contetion is better than certain OOM crash

src/Sentry.OpenTelemetry/SentrySpanProcessor.cs
jamescrosswell: Whoops... that repo was private. Just made it public.

PR #3215, opened 2024-03-13
fix: missing stack trace on exceptions

vaind:
Fixes a regression introduced by https://github.com/getsentry/sentry-dotnet/pull/2709/files#diff-21cc5c4d89fdc7636600f7c8368d895929cf0da4ebb36d117f2e02444dfb9201

### DIFF
--- a/src/Sentry/Internal/MainSentryEventProcessor.cs
+++ b/src/Sentry/Internal/MainSentryEventProcessor.cs
@@ -86,9 +86,8 @@ internal class MainSentryEventProcessor : ISentryEventProcessor
         // if there's no exception with a stack trace, then get the current stack trace
         if (@event.Exception?.StackTrace is null)
         {
-            var stackTrace = @event.SentryExceptions?.FirstOrDefault() is { } sentryEx
-                ? sentryEx.Stacktrace
-                : SentryStackTraceFactoryAccessor().Create();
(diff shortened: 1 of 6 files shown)


src/Sentry/Internal/MainSentryEventProcessor.cs
```diff
@@ -86,9 +86,8 @@ public SentryEvent Process(SentryEvent @event)
         // if there's no exception with a stack trace, then get the current stack trace
         if (@event.Exception?.StackTrace is null)
         {
-            var stackTrace = @event.SentryExceptions?.FirstOrDefault() is { } sentryEx
-                ? sentryEx.Stacktrace
-                : SentryStackTraceFactoryAccessor().Create();
+            var stackTrace = @event.SentryExceptions?.FirstOrDefault()?.Stacktrace
+                            ?? SentryStackTraceFactoryAccessor().Create();
```
bitsandfoxes, line 90: I don't get how the pattern-matching fails here..

bitsandfoxes: Oh the `Stacktrace` itself is `null`!

bitsandfoxes (approved):

jamescrosswell:
Awesome - thanks @vaind! 🙏🏻

PR #3216, opened 2024-03-13
Revert "GitHubActions(build job): relax branch restrictions (#3085)"

vaind:
This reverts PR #3085. Since the PR, there's a lot of duplicate CI runs. 

The goal of the PR was to allow 3rd party contributors easier access to CI. As it has been explained in the original CI, that is already possible with either creating a draft PR targeting this upstream repo from the fork (even before it is ready for review), or creating a PR inside the fork itself. 

I've had a look at the GH Action docs whether there's a better option, e.g. to filter the `push` trigger to only run on forks, but it doesn't seem to be the case. 
Also, I've checked ou some of the most-starred GH repos and haven't seen a better solution - it seems they're either doing the same as we do here or having manual workflow triggers.

If there's a better solution that would not incurr duplicate CI runs, we can follow up in a new issue/PR. For the time being, I suggest we fix the duplicate CI by going beck to the working solution.

cc @knocte  

#skip-changelog

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -3,7 +3,8 @@ name: build
 on:
   push:
     branches:
-      - '**'
+      - main
+      - release/*
   pull_request:
     paths-ignore:
       - "**.md"


knocte:
> e.g. to filter the push trigger to only run on forks, but it doesn't seem to be the case.

I disagree. This can be done easily.

bitsandfoxes:
> I disagree. This can be done easily.

Could you help us out here on how?

knocte:
> Could you help us out here on how?

You can use an `if` element on your CI job that checks the GITHUB_REPOSITORY_OWNER variable (against `getsentry` value of course).

knocte:
More info: https://docs.github.com/en/actions/learn-github-actions/variables

jamescrosswell:
> You can use an `if` element on your CI job that checks the GITHUB_REPOSITORY_OWNER variable (against `getsentry` value of course).

If I've understood, you're saying we could add some conditionality so that builds run on a push on all branches, only if the repo owner is not `getsentry`... otherwise push would only trigger a build on the `main` and `release/*` branches.

It looks like that [if statement in GitHub can be applied to Jobs but not to Workflows](https://docs.github.com/en/actions/using-workflows/triggering-a-workflow#using-conditionals).

So in order for this to work we'd need to leave the workflow configuration as:
```yaml
  push:
    branches:
      - '**'
``` 

Then we'd need to do some fancy footwork in all 4 of the jobs within that workflow. We do something similar for the test-solution-filters job already:
https://github.com/getsentry/sentry-dotnet/blob/d43c1def7c8b504839ad144109d53e37333b410c/.github/workflows/build.yml#L202

That particular conditional would need to be a bit more complex... taking into account
1. The repo owner
2. The branch name
3. The trigger for the event (push vs pull)

It seems like it could be done. It also seems like it introduces a lot of complexity (something we're in no short supply of already). 

I'm maybe biased from previous discussions on this topic though.

knocte:
> It seems like it could be done. It also seems like it introduces a lot of complexity (something we're in no short supply of already).

Sorry, a single `if` line of code is "a lot of complexity"? Because yes, it can be a single line if you just create a dummy 1st job with it, and make the 2nd job depend on it with a `needs:` clause. (And other jobs depend on 2nd, like a chain, or the 1st in case you want to run them on parallel).

And I still don't see what's the issue anyway with having "duplicated jobs". CI for opensource projects is free.

jamescrosswell:
> Sorry, a single `if` line of code is "a lot of complexity"?

Whilst I appreciate the rationale, I don't much appreciate the sarcasm. We're just talking about software here. Let's keep it friendly.

Issue #3220, opened 2024-03-14
Scopes not being applied to Spans when using ASP.NET Core with OTel


jamescrosswell:
This code:
```
app.MapGet("/hello", async context =>
{
   SentrySdk.ConfigureScope(scope =>
   {
       scope.AddBreadcrumb("Saying hello...");
       scope.SetTag("Greeting", "Hello");
       using var task = SampleTelemetry.ActivitySource.StartActivity("Greet");
       task?.SetTag("Answer", 42);
       Thread.Sleep(100); // simulate some work
   });


   await context.Response.WriteAsync("Hey bro!");
});
```

Creates this trace:
<img width="1630" alt="image" src="https://github.com/getsentry/sentry-dotnet/assets/728212/d9f8ac23-aed0-4422-8f62-95cdb749d590">

Neither the Tags nor the Breadcrumbs flow through to Sentry. 

This is because the `_hub` being resolved here is a `HubAdapter` rather than a plain old `Hub`:
https://github.com/getsentry/sentry-dotnet/blob/a448cc45beef2109a7f5d6d628b4b4dce43f4496/src/Sentry.OpenTelemetry/SentrySpanProcessor.cs#L195-L198

So that logic for resolving the Hub needs to be improved.

PR #3221, opened 2024-03-14
Fixed issue resolving scope in ASP.NET Core with OTel

jamescrosswell:
Resolves https://github.com/getsentry/sentry-dotnet/issues/3220

This is hard to test as the Hub that we weren't resolving before (but now do) is SentrySdk.CurrentHub, which is static... We could possibly add an integration test using a TestServer. 

Here's the resulting trace though (includes breadcrumbs and tags now):

image

### DIFF
--- a/src/Sentry.OpenTelemetry/SentrySpanProcessor.cs
+++ b/src/Sentry.OpenTelemetry/SentrySpanProcessor.cs
@@ -21,6 +21,7 @@ public class SentrySpanProcessor : BaseProcessor<Activity>
 
     private static readonly long PruningInterval = TimeSpan.FromSeconds(5).Ticks;
     internal long _lastPruned = 0;
+    private readonly Lazy<Hub?> _realHub;
 
     /// <summary>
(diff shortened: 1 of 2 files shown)


src/Sentry.OpenTelemetry/SentrySpanProcessor.cs
```diff
@@ -211,6 +213,21 @@ public override void OnEnd(Activity data)
...
+    {
+        if (_hub is Hub hub)
+        {
+            return hub;
+        }
+
+        if (_hub is HubAdapter && SentrySdk.CurrentHub is Hub sdkHub)
+        {
+            return sdkHub;
+        }
```
vaind, line 226: Maybe this could be done in the constructor instead of each time?

jamescrosswell: Good call. I've refactored it into a lazy lambda as well.

vaind (approved):

Issue #3228, opened 2024-03-18
Implicit OpenTelemetry initialization


jamescrosswell:
Related to:
- https://github.com/getsentry/sentry-dotnet/issues/3159

## Overview

If we're going to build upon OpenTelemetry and leverage this behind the scenes for Tracing, we don't want to add complexity for people who are currently using Sentry for Tracing. One important aspect of that is initialization.

At the moment if you want to use Sentry with OpenTelemetry for Tracing, initialization is a two step process.
1. You need to Initialize OpenTelemetry... e.g.:
https://github.com/getsentry/sentry-dotnet/blob/bd62cf0421cc7a5d7b7804f8461fae00d47d43b8/samples/Sentry.Samples.OpenTelemetry.Console/Program.cs#L23-L30
2. You need to configure Sentry:
https://github.com/getsentry/sentry-dotnet/blob/bd62cf0421cc7a5d7b7804f8461fae00d47d43b8/samples/Sentry.Samples.OpenTelemetry.Console/Program.cs#L16-L21

And the stuff you configure in step 1 has to match the stuff you configure in Step 2... You can't call `UseOpenTelemetry` when initializing Sentry if you haven't called `UseSentry` when initializing OpenTelemetry. 

Ideally, particularly if users don't even know (or care) that OpenTelemetry is being used behind the scenes, this would be a seamless one step process.

jamescrosswell:
We didn't find any straight forward way to do this.

PR #3229, opened 2024-03-18
Experimenting with implicit OpenTelemetry initialization

jamescrosswell:
Playing around with one idea to implement:
- https://github.com/getsentry/sentry-dotnet/issues/3228

This is a total hack at the moment, just to get an idea of the kinds of challenges involved. 

## Approach

OpenTelemetry Tracing is configured via a [Static API](https://github.com/open-telemetry/opentelemetry-dotnet/blob/680115f9dca33d8d730093cfd14c7917262c34e2/src/OpenTelemetry/Sdk.cs#L92-L95) that returns a `TracerProviderBuilderBase`. 

In the context of a Console application especially (where there's no common `IServiceCollection` for DI), it's relatively difficult to work out whether OpenTelemetry has already been initialized or not. The OpenTelemetry static API doesn't provide an `Sdk.Current` property or anything similar.

However one possible solution might be to leverage our own AddSentry extension method:
https://github.com/getsentry/sentry-dotnet/blob/818e4043303a83f32d554c3e72f1ffd2f3d3c3b3/src/Sentry.OpenTelemetry/TracerProviderBuilderExtensions.cs#L30-L35

If we **_always_** [initialise OpenTelemetry when initializing Sentry](https://github.com/getsentry/sentry-dotnet/blob/01beb42931717d3d0893e443c7ec1afeabf8b0a2/src/Sentry.OpenTelemetry/SentryOptionsExtensions.cs#L58), and we do so using the above extension method... we can guarantee it will be called _**at least once**_... If this extension method gets called twice, we know the initialization we made internally was unecessary and can either [skip our own initialization](https://github.com/getsentry/sentry-dotnet/blob/01beb42931717d3d0893e443c7ec1afeabf8b0a2/src/Sentry.OpenTelemetry/TracerProviderBuilderExtensions.cs#L43-L46) of the Otel SDK or [dispose of any `TracerProvider` we created previously](https://github.com/getsentry/sentry-dotnet/blob/01beb42931717d3d0893e443c7ec1afeabf8b0a2/src/Sentry.OpenTelemetry/InternalTracerProvider.cs#L18-L21).

Broadly speaking, this [seems to work](https://github.com/getsentry/sentry-dotnet/blob/20f807a13219859d8688e3f8ea9fe13a236020e4/samples/Sentry.Samples.OpenTelemetry.Console/Program.cs#L22-L24)... there are a few challenges to making this production ready however.

## Challenges

- This POC uses the [`Sdk.CreateTracerProviderBuilder` method](https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/src/OpenTelemetry/Sdk.cs#L82). In an ASP.NET Core app we would need to use the [OpenTelemetryBuilder extensions](https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/src/OpenTelemetry.Extensions.Hosting/OpenTelemetryServicesExtensions.cs) instead. 
  - If we knew OpenTelemetry was being used as **_the_** trace implementation for Sentry then we could wire this up in the SentryOptions class and override it in SentryAspNetCoreOptions etc. to be platform specific.
- If we're automatically instrumenting OpenTelemetry we have to make some assumptions about the ActivitySource names to listen to. We'd probably need to create our own ActivitySources and use these internally (e.g. in the implementation of IHub.StartTransaction) to start new OpenTelemetry activities instead of creating Sentry Spans. 
- Wiring this up has a dependency on OpenTelemetry, so we'd have to move a bunch of the Tracing stuff to the Sentry.OpenTelemetry package... if people wanted Tracing, they'd need to add that NuGet package to their project...
- If we did want to continue to support both Sentry's proprietary tracing and OpenTelemetry tracing, we'd need to create some common interface and have an implementation of this for Sentry plus another for OpenTelemetry.
- We'd need to work out what to do with our own integrations that are part crash reporting and part tracing. In some cases the tracing part would be obsolete... for example our ASP.NET Core integration (the tracing has already been done in OpenTelemetry by Microsoft). In other cases, we'd need to work out what to do however. This needs a bit of an audit of our integrations and a plan for what to do about them.

### DIFF
--- a/samples/Sentry.Samples.OpenTelemetry.Console/Program.cs
+++ b/samples/Sentry.Samples.OpenTelemetry.Console/Program.cs
@@ -6,25 +6,47 @@
  */
 
 using OpenTelemetry;
-using OpenTelemetry.Resources;
 using OpenTelemetry.Trace;
 using Sentry.OpenTelemetry;
 
-var serviceName = "Sentry.Samples.OpenTelemetry.Console";
-var serviceVersion = "1.0.0";
+var activitySource = Sentry.OpenTelemetry.TracerProviderBuilderExtensions.DefaultActivitySource;
 
 SentrySdk.Init(options =>
 {
     // options.Dsn = "... Your DSN ...";
+    options.Dsn = "https://eb18e953812b41c3aeb042e666fd3b5c@o447951.ingest.sentry.io/5428537";
     options.TracesSampleRate = 1.0;
     options.UseOpenTelemetry(); // <-- Configure Sentry to use OpenTelemetry trace information
 });
 
-using var tracerProvider = Sdk.CreateTracerProviderBuilder()
-    .AddSource(serviceName)
-    .ConfigureResource(resource =>
-        resource.AddService(
-            serviceName: serviceName,
-            serviceVersion: serviceVersion))
-    .AddSentry() // <-- Configure OpenTelemetry to send traces to Sentry
-    .Build();
+// using var tracerProvider = Sdk.CreateTracerProviderBuilder()
+//     .AddSentry() // <-- Configure OpenTelemetry to send traces to Sentry
+//     .Build();
+
+Console.WriteLine("Hello World!");
+Question();
+Console.WriteLine("Goodbye cruel world...");
+
+void Question()
+{
+    using var disposable = SentrySdk.PushScope();
+    SentrySdk.ConfigureScope(scope =>
+    {
+        scope.AddBreadcrumb("Asking the question...");
+        scope.SetTag("Question", "Meaning of life");
+        using var task = activitySource.StartActivity("Question");
+        Thread.Sleep(100); // simulate some work
+        Answer();
+    });
+}
+
+void Answer()
+{
+    SentrySdk.ConfigureScope(scope =>
+    {
+        scope.AddBreadcrumb("Giving the answer...");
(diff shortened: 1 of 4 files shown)


bruno-garcia:
> We'd need to work out what to do with our own integrations that are part crash reporting and part tracing. In some cases the tracing part would be obsolete... for example our ASP.NET Core integration (the tracing has already been done in OpenTelemetry by Microsoft). In other cases, we'd need to work out what to do however. This needs a bit of an audit of our integrations and a plan for what to do about them.

this could be a good thing before we commit to anything. 

A list of things we'd be losing and getting by changing for example. Suggestion migration path. etc. 

It's a big enough change the would benefit having good upfront research and detailed documentation of what this would mean to us. It's 4 years worth of installations literally tens of thousands of apps out there at minimum that rely on this

jamescrosswell:
> > Wiring this up has a dependency on OpenTelemetry, so we'd have to move a bunch of the Tracing stuff to the Sentry.OpenTelemetry package... if people wanted Tracing, they'd need to add that NuGet package to their project...
> 
> No way around this? We'd essentially be tying the sdk to OTels SDK and it's a no return thing with lots of things to be aware of

I'll look into abstracting tracing activities (like creating/finishing spans) out into an interface with the specific implementation of that interface (Sentry vs OTEL) as something that gets registered during Init.

PR #3236, opened 2024-03-23
add net48 without system.reflection.metadata

bruno-garcia:


src/Sentry/Sentry.csproj
```diff
@@ -8,7 +8,7 @@
   </PropertyGroup>
 
   <PropertyGroup Condition="'$(SolutionName)' != 'Sentry.Unity'">
-    <TargetFrameworks>net8.0;net6.0;netstandard2.1;netstandard2.0;net462</TargetFrameworks>
+    <TargetFrameworks>net8.0;net6.0;netstandard2.1;netstandard2.0;net462;net48</TargetFrameworks>
```
jamescrosswell, line 11: Is this supposed to be a new target or to replace the net462 target?

jamescrosswell:
I'm not sure I get it. [Sentry](<https://www.nuget.org/packages/Sentry#dependencies-body-tab>) has a dependency on [System.Reflection.Metadata](https://www.nuget.org/packages/System.Reflection.Metadata/) for .NETFramework 4.6.2, .NETStandard 2.0 and .NETStandard 2.1.

Surely one of those is being resolved when compiling for net48. What does having Sentry target net48 specifically do for us?

src/Sentry/Sentry.csproj
bruno-garcia: No, we need to keep supporting the older one, but we can at least improve for apps running on the latest version

bruno-garcia:
> I'm not sure I get it. [Sentry](https://www.nuget.org/packages/Sentry#dependencies-body-tab) has a dependency on [System.Reflection.Metadata](https://www.nuget.org/packages/System.Reflection.Metadata/) for .NETFramework 4.6.2, .NETStandard 2.0 and .NETStandard 2.1.
> 
> Surely one of those is being resolved when compiling for net48. What does having Sentry target net48 specifically do for us?

If using .NET 4.8, it should pick the net462 of our SDK which does have the dependency. BUT this is breaking, because.. .NET Framework and these problems with dependency resolution (is my assumption).

With 2 customers I debugged this with, I suggested they try adding the dependency directly to their app (S.R.M that is) and it resolved the issue. I didn't have access (nor I have a WIndows machine) to debug the issue myself

jamescrosswell:
> If using .NET 4.8, it should pick the net462 of our SDK which does have the dependency. BUT this is breaking, because.. .NET Framework and these problems with dependency resolution (is my assumption).
> 
> With 2 customers I debugged this with, I suggested they try adding the dependency directly to their app (S.R.M that is) and it resolved the issue. I didn't have access (nor I have a WIndows machine) to debug the issue myself

OK, with this commit, we're basically dropping the following dependencies for people using .net48 : 
```xml
  </ItemGroup>
<PackageReference Include="System.Runtime.InteropServices.RuntimeInformation" Version="4.3.0" />
```

Is that because these packages are included as part of .net48 and later? I couldn't find this documented anywhere but maybe I'm just not looking in the right places...

bitsandfoxes:
> I couldn't find this documented anywhere but maybe I'm just not looking in the right places...

I think this is part of the problem. It "should" work. But we don't know. I'd like to hold off on this for now.

Issue #3159, opened 2024-01-23
.NET SDK powered by OTel


smeubank:
### Description:

See [parent ticket](https://github.com/getsentry/team-sdks/issues/4) for more project background. 

We would like to investigate using OpenTelemetry's .NET SDK to replace Sentry performance package in our own SDK. This would allow our users to benefit from all of the instrumentation of the open standard's SDK, while allowing us to tailor the experience. 

### Requirement:
The OTel SDK should run the show in terms of instrumentations and collecting telemetry, our SDK should only do some fine tuning to ensure data is flagged for ingestion properly (OTel attributed as  tags for example)

Open Point:
- Sentry will have a product ready OTLP endpoint, with this, we have an opportunity to exclude the transaction model all together from our implementation. It should be investigated if we need to adapt our existing trxn model.
- Does OTLP give us a simpler route?

### Information:

- [OTel SDK Docs](https://opentelemetry.io/docs/languages/net/)
  - [OTel .NET SDK](https://github.com/open-telemetry/opentelemetry-dotnet)
  - [Getting Started](https://opentelemetry.io/docs/languages/net/getting-started/)
- [Microsoft OTel Docs](https://learn.microsoft.com/en-us/dotnet/core/diagnostics/observability-with-otel)

# Updated requirements 

See [Discord chat](https://discord.com/channels/621778831602221064/1225041001765146644/1225427049804595382) for context.

- When Sentry is initialized using OpenTelemetry instrumentation, calls to .StartSpan() or .StartChild() on the Sentry SDK should result in spans that fit seamlessly into the auto-instrumentation provided by OpenTelemetry.
- We also need to ensure the correct trace id is used regardless of whether the backend is otel or sentry only

## Sentry Transactions with OTEL Child Spans

Currently leads to unexpected outcomes. 

```csharp
var transaction = SentrySdk.StartTransaction("Program", "Main");
try
{
   Console.WriteLine("Hello World!");
   using (var task = activitySource.StartActivity("Task 1"))
   {
       task?.SetTag("Answer", 42);
       Thread.Sleep(100); // simulate some work
   }


   using (var task = activitySource.StartActivity("Task 2"))
   {
       task?.SetTag("Question", "???");
       Thread.Sleep(100); // simulate some more work
   }


   Console.WriteLine("Goodbye cruel world...");
}
finally
{
   transaction.Finish();
}
```

The Sentry transaction gets ignored. If we turn on Debug mode we see the following warning:
```
[...]

PR #3238, opened 2024-03-25
WIP: System.Diagnostic wrappers

jamescrosswell:
Another experiment related to https://github.com/getsentry/sentry-dotnet/issues/3159

The plan here is to leverage `System.Diagnostics.DiagnosticSource` under the hood for tracing (this is the technology underpinning OTel in .NET). This has been part of .NET since version 5.0.0 so we'll enable it by default for users targeting net5.0 and later. 

For users targeting .NET Framework or .NET Core 3.1 and earlier this will be opt in via the `Sentry.DiagnosticSource` package. 

## TODO

- [x] **Sentry**: `SentryHttpMessageHandler` and `SentryGraphQLHttpMessageHandler` create spans and propagates context
    - [ ] Tests
- [ ] **Sentry.AspNet**: `HttpContextExtensions` creates spans and propagates context
    - [ ] Tests
- [ ] **Sentry.AspNetCore**: `SentryTracingMiddleware` creates spans and propagates context
    - [ ] Tests
- [ ] **Sentry.Azure.Functions.Worker**: `SentryFunctionsWorkerMiddleware` creates spans and propagates context
    - [ ] Tests
- [ ] **Sentry.DiagnosticSource**: `SentrySqlListener` and `EFDiagnosticSourceHelper` create spans
    - [ ] Tests
- [ ] **Sentry.EntityFramework**:  `SentryQueryPerformanceListener` creates spans
    - [ ] Tests
- [ ] Investigate some equivalent of `SentryPropagator`. That extends `OpenTelemetry.Context.Propagation.BaggagePropagator` so we'll need to extract the relevant bits to some helper class that is used by both BaggagePropagator and `ActivitySpanProcessor`. Need to work out when/why this gets leveraged in the OpenTelemetry code so that we can mimic this as well.
- [ ] Do we need to wire up [some equivalent of] the `OpenTelemetryTransactionProcessor`? Possibly rename this to `ActivityTransactionProcessor` and move it to `Sentry.DiagnosticSource`...
- [ ] Check uses of `Hub.GetSpan()`, such as `Hub.GetTraceHeader` and `Hub.GetBaggage`... how will these work with Activity.Current? How will propagation etc. work?
- [ ] More tests... this is quite a big change and we want good test coverage of all the changes

#### Out of scope

`Sentry.OpenTelemetry` uses OpenTelemetry instrumentation (not the new tracing interfaces). 

The following integrations are also out of scope as they don't include any tracing functionality:
- Sentry.Android.AssemblyReader
- Sentry.AspNetCore.Grpc
- Sentry.Bindings.Android  
- Sentry.Bindings.Cocoa    
- Sentry.Extensions.Logging
- Sentry.Google.Cloud.Functions
- Sentry.Hangfire          
- Sentry.Log4Net        
- Sentry.Maui   
- Sentry.NLog              
- Sentry.Profiling         
- Sentry.Serilog


CHANGELOG.md
```diff
@@ -10,7 +10,8 @@
 ### API changes
 
 - Removed `SentryOptionsExtensions` class - all the public methods moved directly to `SentryOptions` ([#3195](https://github.com/getsentry/sentry-dotnet/pull/3195))
-
+- Sentry now uses System.Diagnostics.DiagnosticSource internally for tracing. Any traces created by Sentry can now be captured via OpenTelemetry. If you are using .NET Core 3.1 or earlier, you will need to add the `Sentry.DiagnosticSource` package to your solution to enable tracing. ([#3238](https://github.com/getsentry/sentry-dotnet/pull/3238))
```
jamescrosswell, line 13: The only way to avoid the new dependency for netstandard and netfx would be to continue to use our existing tracing mechanism for those targets for tracing in all of our integrations. 

This would require quite a bit of extra work to build and maintain... but I guess we'd be making the new OTel behind the scenes instrumentation opt-in only for netfx and netstandard at that point. The only reason I can think why they might do this is so that they could capture Sentry spans from OTel (and show them in something like their Aspire dashboard).

bruno-garcia: > The only way to avoid the new dependency for netstandard and netfx would be to continue to use our existing tracing mechanism for those targets for tracing in all of our integrations.

We don't need to add anything to netstandard at this point, it's just here to have some extra time to phase out. 

But .NET Framework is here to stay. So we need to consider that 

> This would require quite a bit of extra work to build and maintain... but I guess we'd be making the new OTel behind the scenes instrumentation opt-in only for netfx and netstandard at that point.

Yeah makes sense to me.

Seems like it's OK to keep support only to .NET (6+) then. And .NET Framework continues as-is?

jamescrosswell: k, I've created two concrete implementations of the Tracing interfaces:
1. One using the `System.Diagnostics.DiagnosticSource` classes (`ActivitySource`, `Activity` etc.)
2. Another that uses Sentry's tracing classes

The `Activity` classes will be used by default for `net5.0` targets and later... otherwise they're opt in (via the `Sentry.DiagnosticsSource` package). 

So no additional dependencies for anyone, unless they explicitly opt in via `Sentry.DiagnosticsSource`

jamescrosswell:
Unfortunately in some of our code the details of our internal tracing bleed out via a public interface in the `SentryMessageHandler` class:
https://github.com/getsentry/sentry-dotnet/blob/0ee390446894f82b89e92bc7aae4a83122d73f6a/src/Sentry/SentryMessageHandler.cs#L59-L75

Both the `ProcessRequest` and `HandleResponse` methods include an `ISpan?` in their signature... and both of these methods are `protected internal` (which means available internally within any assembly that implements the abstract base class they're part of). 🤔

To avoid a breaking change then, it might be necessary to create copies of these classes that use the new `ITraceSpan` interface in their method signatures instead.

Issue #3261, opened 2024-04-04
Metrics Normalization


cleptric:
Project description:
https://github.com/getsentry/team-sdks/issues/80

jamescrosswell:
See https://github.com/getsentry/sentry-dotnet/pull/3271#discussion_r1559072847

bruno-garcia:
rather, resolved by? https://github.com/getsentry/sentry-dotnet/pull/3271

PR #3271, opened 2024-04-09
Metric normalization

jamescrosswell:
Resolves https://github.com/getsentry/sentry-dotnet/issues/3261

### DIFF
--- a/src/Sentry/MetricHelper.cs
+++ b/src/Sentry/MetricHelper.cs
@@ -68,12 +68,12 @@ internal static partial class MetricHelper
 
     private static readonly Lazy<KeyValuePair<string, string>[]> LazyTagValueReplacements = new(() =>
     [
-        new KeyValuePair<string, string>( "\n", "<LF>" ),
-        new KeyValuePair<string, string>( "\r", "<CR>" ),
-        new KeyValuePair<string, string>( "\t", "<HT>" ),
-        new KeyValuePair<string, string>( @"\", @"\\" ),
-        new KeyValuePair<string, string>( "|", "\u007c" ),
-        new KeyValuePair<string, string>( ",", "\u002c")
+        new KeyValuePair<string, string>("\n", "<LF>"),
+        new KeyValuePair<string, string>("\r", "<CR>"),
+        new KeyValuePair<string, string>("\t", "<HT>"),
+        new KeyValuePair<string, string>(@"\", @"\\"),
+        new KeyValuePair<string, string>("|", "\u007c"),
+        new KeyValuePair<string, string>(",", "\u002c")
     ]);
     private static KeyValuePair<string, string>[] TagValueReplacements => LazyTagValueReplacements.Value;
     internal static string SanitizeTagValue(string input)
--- a/src/Sentry/MetricHelper.cs
+++ b/src/Sentry/MetricHelper.cs
@@ -9,7 +9,7 @@ internal static partial class MetricHelper
     private const int RollupInSeconds = 10;
     private const string InvalidKeyCharactersPattern = @"[^\w\-.\/]+";
     // See https://docs.sysdig.com/en/docs/sysdig-monitor/integrations/working-with-integrations/custom-integrations/integrate-statsd-metrics/#characters-allowed-for-statsd-metric-names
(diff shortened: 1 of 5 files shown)


src/Sentry/MetricHelper.cs
```diff
@@ -44,28 +44,47 @@ internal static long GetTimeBucketKey(this DateTimeOffset timestamp)
...
 
     [GeneratedRegex(InvalidMetricUnitCharactersPattern, RegexOptions.Compiled)]
     private static partial Regex InvalidMetricUnitCharacters();
-    internal static string SanitizeMetricUnit(string input) => InvalidMetricUnitCharacters().Replace(input, "_");
+    internal static string SanitizeMetricUnit(string input) => InvalidMetricUnitCharacters().Replace(input, "");
 #else
-    private static readonly Regex InvalidKeyCharacters = new(InvalidKeyCharactersPattern, RegexOptions.Compiled);
-    internal static string SanitizeKey(string input) => InvalidKeyCharacters.Replace(input, "_");
+    private static readonly Regex InvalidMetricKeyOrNameCharacters = new(InvalidMetricKeyOrNameCharactersPattern, RegexOptions.Compiled);
+    internal static string SanitizeMetricKeyOrName(string input) => InvalidMetricKeyOrNameCharacters.Replace(input, "_");
```
bitsandfoxes, line 60: Just double checking: We're keeping the `_` for MetricKeyOrName only?

src/Sentry/MetricHelper.cs
```diff
@@ -44,28 +44,47 @@ internal static long GetTimeBucketKey(this DateTimeOffset timestamp)
...
     private static readonly Regex InvalidMetricUnitCharacters = new(InvalidMetricUnitCharactersPattern, RegexOptions.Compiled);
-    internal static string SanitizeMetricUnit(string input) => InvalidMetricUnitCharacters.Replace(input, "_");
+    internal static string SanitizeMetricUnit(string input) => InvalidMetricUnitCharacters.Replace(input, "");
 #endif
 
+    private static readonly Lazy<KeyValuePair<string, string>[]> LazyTagValueReplacements = new(() =>
+    [
+        new KeyValuePair<string, string>("\n", "<LF>"),
+        new KeyValuePair<string, string>("\r", "<CR>"),
+        new KeyValuePair<string, string>("\t", "<HT>"),
```
bitsandfoxes, line 73: Are those the right way around?

src/Sentry/MetricHelper.cs
jamescrosswell: Yeah, that's per [the new docs](https://develop.sentry.dev/sdk/metrics/#normalization)... slight difference between the allowed characters for each different key or value type as well (e.g. tag keys allow a `/` but metric keys don't).

src/Sentry/MetricHelper.cs
jamescrosswell: I think so. [The docs](https://develop.sentry.dev/sdk/metrics/#normalization) say `Reserved characters for Tag Values are \n, \r, \t, \, |, ,. `

It's a bit confusing because directly underneath that statement is a table where sometimes things from that list are on the left hand side of the table and other times they're on the right hand side of the table. I believe this is a problem with the docs so I opened this PR to correct it:
- https://github.com/getsentry/develop/pull/1218

bitsandfoxes (approved):

src/Sentry/MetricHelper.cs
cleptric: The implementation is not correct. If there is a tab in the tag value, we want a string of `\t` to end up on the tag value. Not `<HT>`.

cleptric: Yes, these three.

jamescrosswell: Since this PR is already merged, I've addressed in a separate PR:
- https://github.com/getsentry/sentry-dotnet/pull/3281

bruno-garcia:
Did this PR resolve this ticket?!:

jamescrosswell:
> Did this PR resolve this ticket?
> 
> * [Metrics Normalization #3261](https://github.com/getsentry/sentry-dotnet/issues/3261)

Kind of, then we worked out we'd done it wrong and had to ammend with this:
- https://github.com/getsentry/sentry-dotnet/pull/3281

Issue #3244, opened 2024-03-27
Add support for catching native crashes in WinUI3


tipa:
### Package

Sentry

### .NET Flavor

.NET

### .NET Version

8.0.2

### OS

Windows

### SDK Version

4.2.1

### Steps to Reproduce

1. Create WinUI3 / WinAppSDK app
2. `Sentry.SentrySdk.CauseCrash(Sentry.CrashType.Native)`

### Expected Result

Crash is being reported to sentry.io

### Actual Result

No crash is being reported (but app crashes with `System.AccessViolationException: 'Attempted to read or write protected memory. This is often an indication that other memory is corrupt`).
Happens both in debug as well as in Release mode

tipa:
Sure, I see these logs before the crash, during initialization:
  Debug: Initializing Hub for Dsn: 'https://197d9a4bd6134f518f21fc19940321c0@o383261.ingest.us.sentry.io/6089453'.
```
  Debug: This looks like a standard JIT/AOT application build.
  Debug: Starting BackgroundWorker.
  Debug: Registering integration: 'AutoSessionTrackingIntegration'.
  Debug: BackgroundWorker Started.
  Debug: Registering integration: 'AppDomainUnhandledExceptionIntegration'.
  Debug: Registering integration: 'AppDomainProcessExitIntegration'.
  Debug: Registering integration: 'SentryDiagnosticListenerIntegration'.
   Info: DiagnosticSource Integration is disabled because tracing is disabled.
  Debug: Registering integration: 'SystemDiagnosticsMetricsIntegration'.
   Info: System.Diagnostics.Metrics Integration is disabled because no listeners are configured.
```

And then when I trigger the crash:

```
  Debug: Triggering a deliberate exception because SentrySdk.CauseCrash(CrashType.Native) was called
An unhandled exception of type 'System.AccessViolationException' occurred in Sentry.dll
Attempted to read or write protected memory. This is often an indication that other memory is corrupt.

```

You can reproduce it easily using this example: [TestException.zip](https://github.com/getsentry/sentry-dotnet/files/14789470/TestException.zip)

bitsandfoxes:
Thanks for the repro and logs! Much appreciated!

jamescrosswell:
I can definitely reproduce this... managed exceptions get caught by the global exception handler that Sentry wires up. This particularly native exception crashes the app, bypassing the global exception handler (and any other exception handling I try to setup). 

[...]

PR #3273, opened 2024-04-09
chore: Adding AOT check to demo crash

bitsandfoxes:
Resolves https://github.com/getsentry/sentry-dotnet/issues/3244
The native crash support requires AOT. 

Can/should we link some additional info in the log? @jamescrosswell blogpost?

#skip-changelog

### DIFF
--- a/src/Sentry/SentrySdk.cs
+++ b/src/Sentry/SentrySdk.cs
@@ -707,6 +707,12 @@ public static partial class SentrySdk
                 break;
 #elif NET8_0_OR_GREATER
             case CrashType.Native:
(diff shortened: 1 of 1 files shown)


jamescrosswell:
Awesome, thanks @bitsandfoxes.

> Can/should we link some additional info in the log?

We could. It's less about the stuff I covered in [that blog post](https://sentry.engineering/blog/should-you-could-you-aot) and more about the hooks into Sentry Native that @vaind added I think. Possibly there is already a good explanation in the PR(s) for the changes Ivan made?

jamescrosswell (approved):

Issue #3248, opened 2024-03-28
Metrics Rate Limits


stephanie-anderson:
Project description:
https://github.com/getsentry/team-sdks/issues/77

bitsandfoxes:
https://develop.sentry.dev/sdk/rate-limiting/

PR #3276, opened 2024-04-10
Added Metrics rate limits

jamescrosswell:
Resolves https://github.com/getsentry/sentry-dotnet/issues/3248

### DIFF
--- a/src/Sentry/Internal/Http/RateLimit.cs
+++ b/src/Sentry/Internal/Http/RateLimit.cs
@@ -34,6 +34,7 @@ internal class RateLimit
                 continue;
             }
 
+            // Response header looking like this: X-Sentry-Rate-Limits: 2700:metric_bucket:organization:quota_exceeded:custom
             namespaces = components.Length > 4 ? components[4].Split(';') : null;
             break;
         }
--- a/src/Sentry/Internal/Http/RateLimit.cs
+++ b/src/Sentry/Internal/Http/RateLimit.cs
@@ -4,14 +4,19 @@ internal class RateLimit
 {
     public IReadOnlyList<RateLimitCategory> Categories { get; }
 
+    public IReadOnlyList<string>? Namespaces { get; }
+
+    internal bool IsDefaultNamespace =>
+        Namespaces is null ||
+        (Namespaces.Count == 1 && string.Equals(Namespaces[0], "custom", StringComparison.OrdinalIgnoreCase));
+
     public TimeSpan RetryAfter { get; }
 
-    public RateLimit(
-        IReadOnlyList<RateLimitCategory> categories,
-        TimeSpan retryAfter)
+    public RateLimit(TimeSpan retryAfter, IReadOnlyList<RateLimitCategory> categories, IReadOnlyList<string>? namespaces = null)
     {
-        Categories = categories;
         RetryAfter = retryAfter;
+        Categories = categories;
+        Namespaces = namespaces;
     }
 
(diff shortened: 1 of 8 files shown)


test/Sentry.Tests/Internals/Http/HttpTransportTests.cs
```diff
@@ -276,13 +276,16 @@ public async Task SendEnvelopeAsync_ResponseNotOkNoMessage_LogsError()
         ).Should().BeTrue();
     }
 
-    [Fact]
-    public async Task SendEnvelopeAsync_ItemRateLimit_DropsItem()
+    [Theory]
+    [InlineData("2700:metric_bucket:organization:quota_exceeded:custom", true)] // Default namespace... we back off
+    [InlineData("2700:metric_bucket:organization:quota_exceeded", true)] // No namespace... we back off
```
cleptric, line 281: Let's add

```c#
[InlineData("2700:metric_bucket:organization:", true)] // Empty namespace... we back off
```

src/Sentry/Internal/Http/RateLimitCategory.cs
```diff
@@ -17,13 +17,18 @@ public bool Matches(EnvelopeItem item)
...
             return false;
         }
 
-        return string.Equals(Name, type, StringComparison.OrdinalIgnoreCase);
+        // The envelope item type for metrics is `statsd`, but the client report category is `metric_bucket`
+        var reportCategory = EnvelopeItem.TypeValueMetric == type
+            ? "metric_bucket"
+            : type;
+
+        return string.Equals(Name, reportCategory, StringComparison.OrdinalIgnoreCase);
```
bitsandfoxes, line 31: Can we be a bit more opinionated here? We're now allocating a new string, comparing multiple times in different ways just to special case the `metric_bucket -> statsd` conversion.
And that happens for every category.

src/Sentry/Internal/Http/RateLimit.cs
```diff
@@ -21,10 +26,19 @@ public static RateLimit Parse(string rateLimitEncoded)
...
         var categories = components[1].Split(';').Select(c => new RateLimitCategory(c)).ToArray();
+        string[]? namespaces = null;
+        foreach (var category in categories)
+        {
+            if (!string.Equals(category.Name, "metric_bucket", StringComparison.OrdinalIgnoreCase))
+            {
+                continue;
+            }
+
+            namespaces = components.Length > 4 ? components[4].Split(';') : null;
```
bitsandfoxes, line 37:
```suggestion
            // Response header looking like this: X-Sentry-Rate-Limits: 2700:metric_bucket:organization:quota_exceeded:custom
            namespaces = components.Length > 4 ? components[4].Split(';') : null;
```

src/Sentry/Internal/Http/RateLimitCategory.cs
jamescrosswell: I've replaced with a switch that removes the string allocation.

src/Sentry/Internal/Http/RateLimitCategory.cs
```diff
@@ -17,13 +17,20 @@ public bool Matches(EnvelopeItem item)
             return true;
         }
 
-        var type = item.TryGetType();
+        var type = item.TryGetType()?.ToLower();
```
bitsandfoxes, line 20:
```suggestion
        var type = item.TryGetType();
```
I think this is now redundant.

bitsandfoxes (approved):
Thanks!

PR #3281, opened 2024-04-10
Fixed normalization for metric tag values

jamescrosswell:
Resolves https://github.com/getsentry/sentry-dotnet/issues/3261

See:
- https://github.com/getsentry/sentry-dotnet/pull/3271#discussion_r1559072847

### DIFF
--- a/src/Sentry/MetricHelper.cs
+++ b/src/Sentry/MetricHelper.cs
@@ -68,16 +68,17 @@ internal static partial class MetricHelper
 
     private static readonly Lazy<KeyValuePair<string, string>[]> LazyTagValueReplacements = new(() =>
     [
-        new KeyValuePair<string, string>("\n", "<LF>"),
-        new KeyValuePair<string, string>("\r", "<CR>"),
-        new KeyValuePair<string, string>("\t", "<HT>"),
-        new KeyValuePair<string, string>(@"\", @"\\"),
+        new KeyValuePair<string, string>("\n", @"\n"),
+        new KeyValuePair<string, string>("\r", @"\r"),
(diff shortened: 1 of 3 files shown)


src/Sentry/MetricHelper.cs
```diff
@@ -68,16 +68,17 @@ internal static long GetTimeBucketKey(this DateTimeOffset timestamp)
 
     private static readonly Lazy<KeyValuePair<string, string>[]> LazyTagValueReplacements = new(() =>
     [
-        new KeyValuePair<string, string>("\n", "<LF>"),
-        new KeyValuePair<string, string>("\r", "<CR>"),
-        new KeyValuePair<string, string>("\t", "<HT>"),
-        new KeyValuePair<string, string>(@"\", @"\\"),
+        new KeyValuePair<string, string>("\n", @"\n"),
```
bruno-garcia, line 71: too bad we can't use `char` here instead of strings.

Everything on the left is `char` though, could it be

`new KeyValuePair<char, string>('\n', @"\r"),`

prob on the right too actually

jamescrosswell: The first three replacements are replacing a single control character with a string of length 2 (so `'/r'` gets replaced with `@"/r"`, for example). 

So yes, potentially we could store the `LazyTagValueReplacements` as a list of `KVP<char, string>` but there's no overload of `String.Replace` that takes `char` as the first argument and `string` as the second... so we'd need to do an unecessary conversion every time the method was called, in that case... which doesn't seem very efficient. 

I figured I'd use the types that were going to be used by method that makes use of them then.

bitsandfoxes (approved):

PR #3288, opened 2024-04-12
Allow OpenTelemetry child spans for Sentry Transactions

jamescrosswell:
Resolves https://github.com/getsentry/sentry-dotnet/issues/3159

## Caveats

It's currently not possible to determine the best parent for OTel spans when mixing OTel and Sentry instrumentation. 

As an example, someone might instrument their code as follows:

```csharp
using var main = activitySource.StartActivity("Main")
{
  var wrapper = transaction.StartChild("Wrapper");

  using (var child = activitySource.StartActivity("Child"))
  {
    ...
  }

  wrapper.Finish();
}
```

Looking at how the code is structured, perhaps what the SDK user wants is for the `child.Parent` to be `wrapper`. However consider the code snippet below:

```csharp
using var main = activitySource.StartActivity("Main")
{
  var wrapper = transaction.StartChild("Wrapper");

  using (var child = activitySource.StartActivity("Child", ActivityKind.Internal, main!.Id))
  {
    ...
  }

  wrapper.Finish();
}
```

In this second code snippet `child.ParentId` has been set explicity to `main.Id` so it's clear that the SDK user wants `main` to be the parent. 

Both code snippets above result in the same inputs to the `SentrySpanProcessor`, which makes it impossible for us to know what the parent really should be.

A similar problem arises if `child` is [created explicitly as a new Root span](https://opentelemetry.io/docs/languages/net/shim/#creating-new-root-spans).

The only way to resolve this would be to leverage OTel instrumentation under the hood in the Sentry SDK itself... but that would imply an additional dependency on `System.Diagnostics.DiagnosticSource` for anyone targeting net5.0 and earlier (including .NET Framework), as we discovered in https://github.com/getsentry/sentry-dotnet/pull/3238.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,9 +4,9 @@
 
 ### Features
 
-- Sentry Transactions now work with OpenTelemetry child spans ([#3288](https://github.com/getsentry/sentry-dotnet/pull/3288))
-### Fixes
+- The SDK's performance API now works in conjunction with OpenTelemetry's instrumentation. This means that SentrySpans and OTel spans now show up in the same span-tree. ([#3288](https://github.com/getsentry/sentry-dotnet/pull/3288))
 
+### Fixes
 -  `HttpResponse.Content` is no longer disposed by when using `SentryHttpFailedRequestHandler` on .NET Framework, which was causing an ObjectDisposedException when using Sentry with NSwag ([#3306](https://github.com/getsentry/sentry-dotnet/pull/3306))
 
 ## 4.4.0
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 ## Unreleased
 
+### Features
+
+## Unreleased
+
+- Sentry Transactions now work with OpenTelemetry child spans ([#3288](https://github.com/getsentry/sentry-dotnet/pull/3288))
+
 ### Dependencies
 
 - Bump Java SDK from v7.7.0 to v7.8.0 ([#3275](https://github.com/getsentry/sentry-dotnet/pull/3275))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
+
+### Features
+
+- Sentry Transactions now work with OpenTelemetry child spans ([#3288](https://github.com/getsentry/sentry-dotnet/pull/3288))
+
 ## 4.4.0
 
 ### Features
@@ -9,12 +15,6 @@
 ### Fixes
 
 - Fixed normalization for metric tag values for carriage return, line feed and tab characters ([#3281](https://github.com/getsentry/sentry-dotnet/pull/3281))
-
-### Features
-
-## Unreleased
-
-- Sentry Transactions now work with OpenTelemetry child spans ([#3288](https://github.com/getsentry/sentry-dotnet/pull/3288))
 
 ### Dependencies
 
--- a/src/Sentry.OpenTelemetry/SentrySpanProcessor.cs
+++ b/src/Sentry.OpenTelemetry/SentrySpanProcessor.cs
@@ -87,6 +87,10 @@ public class SentrySpanProcessor : BaseProcessor<Activity>
             // Explicit ParentSpanId of another activity that we have already mapped
             CreateChildSpan(data, mappedParent, data.ParentSpanId);
         }
+        // Note if the current span on the hub is OTel instrumented and is not the parent of `data` then this may be
+        // intentional (see https://opentelemetry.io/docs/languages/net/instrumentation/#creating-new-root-activities)
+        // so we explicitly exclude OTel instrumented spans from the following check.
+        // of Sentry
         else if (_hub.GetSpan() is IBaseTracer { IsOtelInstrumenter: false } inferredParent)
         {
             // When mixing Sentry and OTel instrumentation and we infer that the currently active span is the parent.
--- a/src/Sentry.OpenTelemetry/SentrySpanProcessor.cs
+++ b/src/Sentry.OpenTelemetry/SentrySpanProcessor.cs
@@ -82,58 +82,77 @@ public class SentrySpanProcessor : BaseProcessor<Activity>
     /// <inheritdoc />
     public override void OnStart(Activity data)
     {
-        if (data.ParentSpanId != default && _map.TryGetValue(data.ParentSpanId, out var parentSpan))
+        if (data.ParentSpanId != default && _map.TryGetValue(data.ParentSpanId, out var mappedParent))
         {
-            // We can find the parent span - start a child span.
-            var context = new SpanContext(
-                data.OperationName,
-                data.SpanId.AsSentrySpanId(),
-                data.ParentSpanId.AsSentrySpanId(),
-                data.TraceId.AsSentryId(),
-                data.DisplayName,
-                null,
-                null)
-            {
-                Instrumenter = Instrumenter.OpenTelemetry
-            };
-
-            var span = (SpanTracer)parentSpan.StartChild(context);
-            span.StartTimestamp = data.StartTimeUtc;
-            // Used to filter out spans that are not recorded when finishing a transaction.
-            span.SetFused(data);
-            span.IsFiltered = () => span.GetFused<Activity>() is { IsAllDataRequested: false, Recorded: false };
-            _map[data.SpanId] = span;
+            // Explicit ParentSpanId of another activity that we have already mapped
+            CreateChildSpan(data, mappedParent, data.ParentSpanId);
+        }
+        else if (_hub.GetSpan() is IBaseTracer { IsOtelInstrumenter: false } inferredParent)
+        {
+            // When mixing Sentry and OTel instrumentation and we infer that the currently active span is the parent.
+            var inferredParentSpan = (ISpan)inferredParent;
+            CreateChildSpan(data, inferredParentSpan, inferredParentSpan.SpanId);
         }
         else
         {
-            // If a parent exists at all, then copy its sampling decision.
-            bool? isSampled = data.HasRemoteParent ? data.Recorded : null;
-
-            // No parent span found - start a new transaction
-            var transactionContext = new TransactionContext(data.DisplayName,
-                data.OperationName,
-                data.SpanId.AsSentrySpanId(),
-                data.ParentSpanId.AsSentrySpanId(),
-                data.TraceId.AsSentryId(),
-                data.DisplayName, null, isSampled, isSampled)
-            {
-                Instrumenter = Instrumenter.OpenTelemetry
-            };
-
-            var baggageHeader = data.Baggage.AsBaggageHeader();
-            var dynamicSamplingContext = baggageHeader.CreateDynamicSamplingContext();
-            var transaction = (TransactionTracer)_hub.StartTransaction(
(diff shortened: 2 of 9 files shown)


smeubank:
> Allow OpenTelemetry child spans for Sentry Transactions

exciting stuff! I am just curious from the title.Does this mean that sentry transactions are always the parent span? Is there ever a case where we construct a "Sentry Transaction" based upon entirely OTel spans with OTel parent span?

jamescrosswell:
> Does this mean that sentry transactions are always the parent span? 

Hey @smeubank, Sentry transactions won't always be the parent span no. 

If you've using Sentry's OpenTelemetry integration in ASP.NET Core, for example, an OpenTelemetry span gets created for each http request before the middleware pipeline gets kicked off - so the "transaction" (i.e. span with no parent) in that case is going to be an OTel span. You could use the Sentry.DiagnosticSource integration to then capturing child spans for SqlClient or Entity Framework operations. This is something that's been possible for quite some time. 

The opposite (an OTel span inside a Sentry transaction) wasn't possible prior to this PR however. 

> Is there ever a case where we construct a "Sentry Transaction" based upon entirely OTel spans with OTel parent span?

If you're using the Sentry.OpenTelemetry package then you could perform all of your trace instrumentation using [ActivitySource](https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.activitysource?view=net-8.0), yes. That has been possible ever since the Sentry.OpenTelemetry integration was released.

Is that something you've been wanting to do? Is that just to avoid vendor lock in or some other reason?

src/Sentry/IBaseTracer.cs
bitsandfoxes, line 1: Do we need this as its own interface? Looks like we can put that on the `ISpan`?

bitsandfoxes, line 1: Do we need this as its own interface? Looks like we can put that on the `ISpan`?

test/Sentry.OpenTelemetry.Tests/SentrySpanProcessorTests.cs
```diff
@@ -183,12 +182,56 @@ public void OnStart_WithParentSpanId_StartsChildSpan()
         }
     }
 
+    [Fact]
+    public void OnStart_WithSentryParentSpanId_StartsChildSpan()
```
bitsandfoxes, line 186: That should work the other way around as well, right?

src/Sentry.OpenTelemetry/SentrySpanProcessor.cs
```diff
@@ -82,59 +82,78 @@ _hub switch
...
-            var span = (SpanTracer)parentSpan.StartChild(context);
-            span.StartTimestamp = data.StartTimeUtc;
-            // Used to filter out spans that are not recorded when finishing a transaction.
-            span.SetFused(data);
-            span.IsFiltered = () => span.GetFused<Activity>() is { IsAllDataRequested: false, Recorded: false };
-            _map[data.SpanId] = span;
+            // Explicit ParentSpanId of another activity that we have already mapped
+            CreateChildSpan(data, mappedParent, data.ParentSpanId);
+        }
+        else if (_hub.GetSpan() is IBaseTracer { IsOtelInstrumenter: false } inferredParent)
```
bitsandfoxes, line 90: I'm confused by this check.
`OnStart` handles now:
1. If the parent has already been mapped: Create a ChildSpan
2. The parent has not been mapped, we check the current span on the hub, and if it's not an Otel span (which means there has been a SentrySpan created somewhere, then we add a child to that.
3. Else we create a root span

Since this is strictly for mapping activities to spans, can we run into a scenario where the current span is an Otel span?

src/Sentry.OpenTelemetry/SentrySpanProcessor.cs
```diff
@@ -82,59 +82,78 @@ _hub switch
...
+    {
+        // We can find the parent span - start a child span.
+        var context = new SpanContext(
+            data.OperationName,
+            data.SpanId.AsSentrySpanId(),
+            parentSpanId,
+            description: data.DisplayName
+        )
+        {
+            Instrumenter = Instrumenter.OpenTelemetry
```
bitsandfoxes, line 118: I wonder if we can get rid of the `Instumenter`? Could we handle spans agnostically from their instrumenter?

src/Sentry/IBaseTracer.cs
jamescrosswell: `IBaseTracer` is internal. I didn't want to expose the new property publicly.

src/Sentry.OpenTelemetry/SentrySpanProcessor.cs
jamescrosswell: I figured there were three scenarios.
1. Someone creates an OTel Span and then creates an OTel child span... all using OTel instrumentation. In this case, the ParentSpanId will be set on the child span. We'll find that parent span in our map (since it will have already passed through the span processor
2. Someone creates a Sentry transaction... then creates an OTel span. There won't be anything in the map that matches (as a parent) but we are operating in the context of a \[Sentry\] transaction... so we create a child of that transaction, even though we're instrumenting with OTel at this point
3. If neither of the above is the case, what we have is a new root span (aka transaction). 

The reason I check to make sure the current span is not OTel instrumented is that if there is an OTel instrumented span active and it's not explicitly marked as the parent of the Activity that our span processor is handling, then it's not supposed to be the parent span. They've intentionally instrumented things to start a new span tree at this point. I think [this](https://opentelemetry.io/docs/languages/net/instrumentation/#creating-new-root-activities) is the only way they could do that, but it is a logical possibility.

```csharp
public static void DoWork()
{
    var previous = Activity.Current;
    Activity.Current = null;

    var newRoot = MyActivitySource.StartActivity("NewRoot");

    // Re-set the previous Current Activity so the trace isn't messed up
    Activity.Current = previous;
}
```

jamescrosswell: I'll add a comment in the code to hint at that.

test/Sentry.OpenTelemetry.Tests/SentrySpanProcessorTests.cs
jamescrosswell: It does yes... I'll add another test.

src/Sentry.OpenTelemetry/SentrySpanProcessor.cs
jamescrosswell: See [comment above](https://github.com/getsentry/sentry-dotnet/pull/3288#discussion_r1569611590) - I don't think so. 

I wondered if we could get rid of it from our options as well, but we use it there to work out whether we enable the SentryTracingMiddleware for ASP.NET Core.

src/Sentry/SpanTracer.cs
```diff
@@ -6,11 +6,15 @@ namespace Sentry;
...
 /// </summary>
-public class SpanTracer : ISpan
+public class SpanTracer : IBaseTracer, ISpan
 {
     private readonly IHub _hub;
     private readonly SentryStopwatch _stopwatch = SentryStopwatch.StartNew();
 
+    private readonly Instrumenter _instrumenter = Instrumenter.Sentry;
+
+    bool IBaseTracer.IsOtelInstrumenter => _instrumenter == Instrumenter.OpenTelemetry;
```
bitsandfoxes, line 16: An internal property on an internal interface. So that's why it looks to wonky to me!

CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
+
+### Features
+
+- Sentry Transactions now work with OpenTelemetry child spans ([#3288](https://github.com/getsentry/sentry-dotnet/pull/3288))
```
```suggestion
- The SDK's performance API now works in conjunction with OpenTelemetry's instrumentation. This means that SentrySpans and captured Activities now show up in the same span-tree. ([#3288](https://github.com/getsentry/sentry-dotnet/pull/3288))
```

bitsandfoxes (commented):
This is already looking really really good!
I think there is only one case missing to be covered: Wrapping an OTel activity by a span within another OTel activity. I made the following changes to the console sample:
```
var activitySource = new ActivitySource("Sentry.Samples.OpenTelemetry.Console");

SentrySdk.Init(options =>
{
    // options.Dsn = "... Your DSN ...";
    options.UseOpenTelemetry(); // <-- Configure Sentry to use OpenTelemetry trace information
});

using var tracerProvider = Sdk.CreateTracerProviderBuilder()
    .AddSource(activitySource.Name)
    .AddSentry() // <-- Configure OpenTelemetry to send traces to Sentry
    .Build();

var transaction = SentrySdk.StartTransaction("an_otel_transaction", "sentry_span");

// Finally we can use OpenTelemetry to instrument our code. These activities will be captured as a Sentry transaction.
using var activity = activitySource.StartActivity("Main");
Console.WriteLine("Hello World!");
using (var task = activitySource.StartActivity("Task 1"))
{
    task?.SetTag("Answer", 42);
    Thread.Sleep(100); // simulate some work
    Console.WriteLine("Task 1 completed");
    task?.SetStatus(Status.Ok);
}

var span = transaction.StartChild("an_otel_activity_wrapping_span");

using (var task = activitySource.StartActivity("Task 2"))
{
    task?.SetTag("Question", "???");
    Thread.Sleep(100); // simulate some more work
    Console.WriteLine("Task 2 unresolved");
    task?.SetStatus(Status.Error);
}

span.Finish();
transaction.Finish();

Console.WriteLine("Goodbye cruel world...");
```

And the wrapping span ends up as a proper child of the `Main` activity but should also be parent of `Task 2`.

jamescrosswell:
> And the wrapping span ends up as a proper child of the `Main` activity but should also be parent of `Task 2`.

Hm, that's kind of the normal mechanics of OTel tracing. When you create an activity, the parent of that activity is inferred from`Activity.Current` (from the `System.Diagnostics.DiagnosticSource` namespace).

I'm wondering how we could infer whether the parent should be the current sentry span or the current activity in that scenario, without getting into trouble. We could use the start date on SentrySdk.GetSpan vs Activity.Current to try to infer which one should be the parent... but is that really what should happen? I can imagine a scenario like this for example:
```
Request:
  -> Start Transaction for request (e.g. OTel)
     -> Request Handler
        -> Call some async functionality instrumented using OTEL
        -> Start Span manually (using Sentry instrumentation)
           -> Outbound integration has some further instrumentation (e.g. Database stuff)
        -> Do some work in the request handler (this is what is supposed to be instrumented by the manual span)
        -> Outbound request completes
        -> Finish manual span
```

In that scenario, I think the "further instrumentation" (the database stuff) would end up as a child of the manual sentry span (just because of the order of async execution)... when actually it should be a child of the otel instrumentation. 

Or would it? Are we maybe starting to get into the whole hubs and scopes discussion that the JS guys ended up in? 

I'll play around with some things here locally and see how this behaves with the async local scope manager.

jamescrosswell:
@bitsandfoxes after a bit of ruminating, it doesn't look like there's any way to solve for the scenario you described without potentially breaking the trace hierarchy in other senarios... see **Caveats** that I just added to the [PR description](https://github.com/getsentry/sentry-dotnet/pull/3288#issue-2239037125).

I don't think we'll get seemless integration with OTel without a major version bump then (and a new dependency for SDK users targeting net5.0 and earlier).

bitsandfoxes (approved):
The caveats do make sense. We have no way of knowing if that relationship got created explicitly by the user. We're going to have to document this limitation.

Issue #3259, opened 2024-04-04
CaptureCheckIn always report production environment, even if SDK environment differs


digital88:
### Package

Sentry

### .NET Flavor

.NET

### .NET Version

5.0.17

### OS

Windows

### SDK Version

4.2.1

### Self-Hosted Sentry Version

24.1.2

### Steps to Reproduce

This is what I see in Crons Tab

![image](https://github.com/getsentry/sentry-dotnet/assets/10961828/268290a6-122e-4969-a0e6-bbc43a80b913)

This is one of transactions with correct Environment

![image](https://github.com/getsentry/sentry-dotnet/assets/10961828/c4f2e279-1c26-4f33-8fc5-22c6cd17686b)

Code:

appsettings.json

```
{
  "Sentry": {
    "Dsn": ...,
    "Environment": "test",
    ...
  },
  "Serilog": {
    "WriteTo": [
      {
        "Name": "Sentry",
        "Args": {
          "InitializeSdk": false,
          "MinimumBreadcrumbLevel": "Information",
          "MinimumEventLevel": "Warning"
        }
      }
    ]
  }
}
```

Sentry Init Code (simplified)

```
public static IHostBuilder CreateHostBuilder(string[] args) =>
	Host.CreateDefaultBuilder(args)
		.ConfigureAppConfiguration((_, builder) =>
		{
			builder.AddEnvironmentVariables(...);
			builder.AddJsonFile("appsettings.json", optional: false);
		})
		.ConfigureServices((hostContext, services) =>
		{
			services.AddHostedService<MyService>();
			services.RegisterDependencies(hostContext.Configuration);
			SentrySettings.ConfigureSentry(hostContext.Configuration);
		})
		.UseWindowsService()
		.UseSystemd()
		.UseSerilog((ctx, config) => config.ReadFrom.Configuration(ctx.Configuration));

.. SentrySettings class
public static void ConfigureSentry(IConfiguration configuration)
{
	if (!bool.TryParse(configuration.GetSection("SentryEnabled").Value, out bool sentryEnabled)) return;
	if (!sentryEnabled) return;

	var opts = configuration.GetSection("Sentry").Get<SentryOptions>();
	var proxySettings = configuration.GetSection("Proxy").Get<ProxySettings>();
	if (proxySettings != null && !string.IsNullOrWhiteSpace(proxySettings.Url))
	{
		opts.HttpProxy = new WebProxy()
		{
[...]

PR #3320, opened 2024-04-22
feat: Added properties to `CheckIns`

bitsandfoxes:
Fixes https://github.com/getsentry/sentry-dotnet/issues/3259

Based on our [dev docs](https://develop.sentry.dev/sdk/check-ins/) check-ins now have 
```
- CheckIn ID
- Monitor Slug
- Status
- Duration
- Release
- Environment
- Contexts -> trace -> trace ID
```

`Release` and `Environment` are getting applied by the Enricher.
`Duration` is an optional parameter and currently gets set by the Hangfire Integration when a task is finished.
The `Trace ID` gets taken from the scope. Either from the propagation context or, if present, from the current span.

### DIFF
--- a/src/Sentry/Internal/Hub.cs
+++ b/src/Sentry/Internal/Hub.cs
@@ -566,7 +566,8 @@ internal class Hub : IHub, IMetricHub, IDisposable
         }
     }
 
-    public SentryId CaptureCheckIn(string monitorSlug, CheckInStatus status, SentryId? sentryId = null, TimeSpan? duration = null)
+    public SentryId CaptureCheckIn(string monitorSlug, CheckInStatus status, SentryId? sentryId = null,
+        Scope? scope = null, TimeSpan? duration = null)
     {
         if (!IsEnabled)
         {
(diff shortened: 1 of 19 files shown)


src/Sentry/Internal/Hub.cs
```diff
@@ -567,7 +569,14 @@ public SentryId CaptureCheckIn(string monitorSlug, CheckInStatus status, SentryI
         try
         {
             _options.LogDebug("Capturing '{0}' check-in for '{1}'", status, monitorSlug);
-            return _ownedClient.CaptureCheckIn(monitorSlug, status, sentryId);
+
+            if (scope is null)
+            {
+                ScopeManager.GetCurrent().Deconstruct(out var currentScope, out _);
```
jamescrosswell, line 575:
```suggestion
                var (currentScope, _) = ScopeManager.GetCurrent();
```

src/Sentry/SentryCheckIn.cs
```diff
@@ -40,11 +40,32 @@ public class SentryCheckIn : ISentryJsonSerializable
     /// </summary>
     public string MonitorSlug { get; }
 
+
```
jamescrosswell, line 43:
```suggestion
```

jamescrosswell (approved):
LGTM... suggested a couple of cosmetic changes.

Issue #3158, opened 2024-02-19
Allow regular expressions in the InAppIncludes and InAppExcludes


jamescrosswell:
### Problem Statement

One of our customers wants to be able to simply add their own namespace as an include and exclude everything else (see [Discord thread](https://discord.com/channels/621778831602221064/621783515423440927/1209267123554099250)).

Currently this doesn't work as the logic to determine whether a frame is InApp or not is:
https://github.com/getsentry/sentry-dotnet/blob/5fab09efd4204fd07d95c818df79f7aa71d61e88/src/Sentry/SentryStackFrame.cs#L218

So if you don't have any excludes configured or if the namespace doesn't match one of more of the  prefixes configured as an exclude then frame is determined as InApp. This seems way too greedy. That means if you have no includes configured but you have these excludes configured:
```
- Foo
- Bar
```

Then the namespace `Bar` will be considered InApp (because it doesn't match `Foo`). Given there's an explicit exclude configured for `Bar`, this seems counter-intuitive. I think there's a bug in this logic then.

The logic should instead be "If it matches any of the Includes and it doesn't match any of the excludes, then it's InApp". 

### Solution Brainstorm

If `SentryOptions.InAppInclude` and `SentryOptions.InAppExclude ` were lists of `SubstringOrRegex` rather than just lists of prefixes, it would be possible to configure an Exclude of `.*`... which would enable the customer to do what they want. 

We also need to fix the logic for matching exclusions from "doesn't match any single exclusion" to "doesn't match any exclusion".

PR #3321, opened 2024-04-23
Regular expressions can now be used for InApp includes/excludes

jamescrosswell:
Resolves https://github.com/getsentry/sentry-dotnet/issues/3158


test/Sentry.Tests/Internals/SentryStackTraceFactoryTests.MethodGeneric_mode=Enhanced.verified.txt
```diff
@@ -2,6 +2,6 @@
   FileName: Internals/SentryStackTraceFactoryTests.cs,
   Function: void SentryStackTraceFactoryTests.GenericMethodThatThrows<T>(T value),
   AbsolutePath: {ProjectDirectory}Internals/SentryStackTraceFactoryTests.cs,
-  InApp: true,
```
bitsandfoxes, line 5: Where is this change coming from?

bitsandfoxes (dismissed):
That is a really nice QOL improvement.

test/Sentry.Tests/Internals/SentryStackTraceFactoryTests.MethodGeneric_mode=Enhanced.verified.txt
jamescrosswell: The `parameter` being matched in this case is`Other.Tests.Internals.SentryStackTraceFactoryTests` (that's basically the namespace for our test). 

There's only one `InAppInclude`: `Sentry.Samples` so that doesn't match. 

And there is an exclude that matches: `Sentry` (which is a substring of our namespace). 

So, we deem it `InApp: false`.

src/Sentry/SentryStackFrame.cs
```diff
@@ -204,19 +205,43 @@ public void ConfigureAppFrame(SentryOptions options)
...
+            {
+                options.LogDebug("Error matching prefixOrPattern `{0}` with parameter `{1}`: {2}", prefixOrPattern, parameter, ex.Message);
             }
+
             return false;
         };
-        InApp = options.InAppInclude?.Any(resolver) == true || options.InAppExclude?.Any(resolver) != true;
+        var matchesInclude = options.InAppInclude?.Any(resolver) == true;
+        var matchesExclude = () => options.InAppExclude?.Any(resolver) ?? false;
+        InApp = matchesInclude || !matchesExclude();
```
vaind, line 244: Constructing a new SubstringOrRegexPattern (which parses a regex from string) on each stack frame doesn't seem like a right thing to do. How about changing the include/exclude list to already consist of `SubstringOrRegexPattern`s? 
Also, I've never been a fan of creating a lambda here for each frame, especially considering profiling, where each sample has a stack trace that needs to be processed (although just once for each unique frame). How about we get rid of this and make InAppInclude/Exclude a function instead? Taking string parameter and returning bool.

vaind (changes requested):

src/Sentry/SentryStackFrame.cs
jamescrosswell: We can easily change the lambda to a function 👍🏻 

There's a TODO on the options to change the way these get stored from `List<string>` to `List<SubstringOrRegex>` (which would mean the regex expressions would only be parsed once per program execution) as that would technically be a breaking change to the API. I'll see if I can do something to leave the public API as it is but back it with a `List<SubstringOrRegex>`.

jamescrosswell: Oh... in fact I see [both of those properties are internal](https://github.com/getsentry/sentry-dotnet/blob/b74a161f6b7821c6fe45dd114c8ae6dc3a76ec9a/src/Sentry/SentryOptions.cs#L237-L264) 🎉

test/Sentry.Tests/Internals/SentryStackTraceFactoryTests.MethodGeneric_mode=Enhanced.verified.txt
jamescrosswell: Actually, look at this with fresh eyes, I think I might have broken something. By changing from `string` to `SubstringOrRegexPattern` the matching logic inadvertedly got changed from `StartsWith` to `Substring`, which matches all kinds of excludes that it shouldn't. 

Fixing now...

src/Sentry/SentryStackFrame.cs
jamescrosswell: I ended up having to create a new class for this as we actually wanted `PrefixOrRegexPattern` not `SubstringOrRegexPattern`. 

I've done away with the lambdas as well (these are now inline functions).

vaind: > I've done away with the lambdas as well (these are now inline functions).
> 
> The tests seem to pass now. 🎈

So indeed it seems like it has been the performance issue. Maybe we should check check that we have some benchmarks for these critical paths and introduce https://github.com/marketplace/actions/continuous-benchmark to the repo.

src/Sentry/PrefixOrRegexPattern.cs
```diff
@@ -0,0 +1,99 @@
+using System;
+using Sentry.Internal;
+
+namespace Sentry;
+
+/// <summary>
+/// Provides a pattern that can be used to match against namespaces as either a prefix or regular expression.
+/// </summary>
+internal class PrefixOrRegexPattern
```
vaind, line 9: Any reason this needs to be a single class that handles both cases? How about an interface (e.g. `StringFilter`) with `IsMatch()` (one variant with separator arg and another one without it to really clean this up). Each implementation can just have its own code for prefix check/regex check (and substring check). Then we can leave the construction of the right implementation outside

src/Sentry.Profiling/SampleProfileBuilder.cs
```diff
@@ -64,9 +65,11 @@ internal void AddSample(TraceEvent data, double timestampMs)
         var stackIndex = AddStackTrace(callStackIndex);
         if (stackIndex < 0)
         {
+            _options.DiagnosticLogger?.LogDebug("Invalid stackIndex for Profiler Sample. Skipping.");
             return;
         }
 
+        _options.DiagnosticLogger?.LogDebug("Adding profile sample.");
```
vaind, line 72: Let's not forget to get rid of this before merging

src/Sentry/SentryOptions.cs
```diff
@@ -1358,14 +1358,38 @@ public void AddInAppExclude(string prefix)
...
+    /// </summary>
+    /// <param name="pattern">The regular expression to use to match stacktrace frames</param>
+    /// <remarks>
+    /// Sentry by default filters the stacktrace to display only application code.
+    /// A user can optionally click to see all which will include framework and libraries.
+    /// If a frame has not already been determined to be InApp per any of the prefixes or
+    /// patterns configured in <see cref="InAppInclude"/>, it will be considered InApp only
+    /// if it does not match any of the prefixes or patterns in <see cref="InAppExclude"/>.
+    /// </remarks>
+    public void AddInAppExcludeRegex(string pattern)
```
vaind, line 1380: Instead of parsing ourselves, I'd suggest adding `AddInAppExclude(Regex pattern)`

test/Sentry.Profiling.Tests/TraceLogProcessorTests.ProfileInfo_Serialization_Works.verified.txt
```diff
@@ -645,7 +645,7 @@
       {
         function: System.Reactive.Linq.Observable.Merge(class System.IObservable`1<!!0>[]),
         module: system.reactive,
-        in_app: true
+        in_app: false
```
vaind, line 648: interesting that it was `true` before

src/Sentry/SentryStackFrame.cs
vaind: BTW there already is a benchmark so we would have probably caught this if we had the action in place:
https://github.com/getsentry/sentry-dotnet/blob/24c159fac119315e21df6138dc04bfe3ccd573c2/benchmarks/Sentry.Benchmarks/StackFrameBenchmarks.cs#L40

src/Sentry/SentryOptions.cs
jamescrosswell: That was originally how I had it... it means reminding people they should use the right options to get decent performance. When parsing ourselves, we ensure the RegEx is created with sensible options:
```
new Regex(pattern, RegexOptions.Compiled | RegexOptions.CultureInvariant)
```

If these Regex instances were going to be used in ways we couldn't anticipate, we'd  definitely be better off giving people more flexibility. Since we know that they will only be used to match namespaces (i.d. perdiod delimited [C# identifiers](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/language-specification/lexical-structure#643-identifiers)) I wondered if we could make the API simpler - so give people less rope to hang themselves but still be powerful enough to do what they need. 

I was a bit worried about the use of `RegexOptions.CultureInvariant`. But the [docs says](https://learn.microsoft.com/en-us/dotnet/standard/base-types/regular-expression-options#compare-using-the-invariant-culture) that only affects case insensitive matches (namespaces would be case sensitive).

test/Sentry.Profiling.Tests/TraceLogProcessorTests.ProfileInfo_Serialization_Works.verified.txt
jamescrosswell: Yes, I think there was a bug in the logic we had previously to identify InApp frames.

src/Sentry/SentryStackFrame.cs
jamescrosswell: > BTW there already is a benchmark so we would have probably caught this if we had the action in place:

Interesting. And how would that typically integrate with CI? Is it a bit like code coverage where you're comparing to previous runs or something?

src/Sentry/PrefixOrRegexPattern.cs
jamescrosswell: Frustratingly, this is actually a really good suggestion 😝 We do eventually have a bunch of other issues to close but I think this is a worthwhile refactoring to make. It should make the code a bit simpler and more composable. It should allows us to get rid of the `SubstringOrRegexPattern` class in the next major version bump, for example (deleting code 🎉 )

jamescrosswell: OK done - thanks @vaind, I think it looks nicer now.

src/Sentry/SentryStackFrame.cs
jamescrosswell: For info on benchmarking options, see:
- https://github.com/getsentry/sentry-dotnet/issues/3332

vaind: > And how would that typically integrate with CI? Is it a bit like code coverage where you're comparing to previous runs or something?

Yes, I'd expect a comment on the PR with previous and current results. I haven't worked with the action myself though so maybe it doesn't have that (yet)

vaind (commented):
It's a bit more complicated solution than what I had in mind in my proposal (albeit more flexible too, I guess), but as long as the performance is OK, I'm fine with that. Have you checked the original vs current version in benchmarks manually? Can you please post the results?

vaind (approved):
great effort & thanks for the benchmarks

Issue #3282, opened 2024-04-10
Merge upstream AOT improvements into the Sentry Ben.Demystifier fork


jamescrosswell:
See:
-  https://github.com/benaadams/Ben.Demystifier/pull/222

PR #3322, opened 2024-04-24
Improve native AOT compatibility in Ben.Demystifier

jamescrosswell:
Resolves https://github.com/getsentry/sentry-dotnet/issues/3282

As Ben Adams appears to be awol, even though it hasn't been merged into main on the upstream repo yet I've patched our fork of Ben.Demystifier with:
- https://github.com/benaadams/Ben.Demystifier/pull/222 

Nothing to note in the changelog until we actually leverage the new code for AOT compilation. We'll need to address the other things Michal mentioned in that PR before doing that however, which we can do in:
- https://github.com/getsentry/sentry-dotnet/issues/3323

#skip-changelog


bitsandfoxes (approved):
Thanks for taking care of this!

Issue #3310, opened 2024-04-17
Add logs to `metrics`


bitsandfoxes:
Trying to help a user figure out why they don't see any metrics on the dashboard made me realize that I had to go through all the HttpTransport logs to look for `type = statsd`. Which was not the most pleasant user expierence.

PR #3324, opened 2024-04-24
Add logging to Metrics

jamescrosswell:
Resolves https://github.com/getsentry/sentry-dotnet/issues/3310

#skip-changelog

### DIFF
--- a/src/Sentry/Internal/Hub.cs
+++ b/src/Sentry/Internal/Hub.cs
@@ -63,11 +63,11 @@ internal class Hub : IHub, IMetricHub, IDisposable
 
         if (options.ExperimentalMetrics is not null)
         {
+            options.LogDebug("Registering integration: Metrics");
             Metrics = new MetricAggregator(options, this);
         }
         else
         {
-            options.LogDebug("Metrics disabled in SentryOptions");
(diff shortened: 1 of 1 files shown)


jamescrosswell:
@bitsandfoxes I had a bit of a scan through the `MetricAggregator` class but didn't see anywhere obvious to add logging... what additional logging did you want here?

jamescrosswell:
> > @bitsandfoxes I had a bit of a scan through the `MetricAggregator` class but didn't see anywhere obvious to add logging... what additional logging did you want here?
> 
> Initially, I thought that just emitting a metric could leave a log but I can see this getting quite noisy. Then I thought maybe just during flushing but there are already logs there. Now I think the issue came through Unity's logger not being able to handle the logs coming from the background.

OK, should we bother with this PR or close it do you think? It just adds a log if metrics are disabled... not sure how useful that is. 

We'll need to fix the broken verify tests if we want to merge.

bitsandfoxes:
> It just adds a log if metrics are disabled

How about we do it the other way around and treat is like an integration that logs if it got registerd.

bitsandfoxes (approved):

Issue #3209, opened 2024-03-12
Failed to send envelopes persist on disk


bitsandfoxes:
## Problem Summary

First raised through Unity SDK [here](https://discord.com/channels/621778831602221064/1215954412384817292). 

Repro with an attachment bigger than 50mb
```
SentrySdk.CaptureEvent(new SentryEvent(), scope =>
{
    scope.AddAttachment("/path/to/something/big");
});
```

What ends up happening: The SDK tries to send the envelope, fails but keeps the envelope around indefinitely.

jamescrosswell:
> What ends up happening: The SDK tries to send the envelope, fails but keeps the envelope around indefinitely.

What do we want to happen instead @bitsandfoxes ?

We could maintain a list in memory of envelopes that we've failed to send along with a retry count. When the count hits X, we could just log the failure and throw it away?

bitsandfoxes:
I think we should handle this differently based on the reason the envelope got rejected. The attachment being too big won't change during a retry so we should log and delete the envelope.
Alternatively, we could think about dropping the attachment and sending the envelope without it?

PR #3328, opened 2024-04-25
Fixed envelopes with large attachments persist on disk

jamescrosswell:
Resolves https://github.com/getsentry/sentry-dotnet/issues/3209

## Problem summary

This is only a problem when using the `CachingTransport`.  The issue is in two parts. Firstly here:
https://github.com/getsentry/sentry-dotnet/blob/59f31642c10f8c8159d8294e0a48536b14ad4cd0/src/Sentry/Http/HttpTransportBase.cs#L129-L130

If the `length` header has not been set then the envelope item is not correctly identified as being too big... the result is that we only get an exception when we try to send the attachment in the envelope to Sentry.

The second factor contributing to the issue is that when deserializing an envelope from the cache, this header gets removed:
https://github.com/getsentry/sentry-dotnet/blob/59f31642c10f8c8159d8294e0a48536b14ad4cd0/src/Sentry/Protocol/Envelopes/EnvelopeItem.cs#L446

### DIFF
--- a/src/Sentry/Protocol/Envelopes/EnvelopeItem.cs
+++ b/src/Sentry/Protocol/Envelopes/EnvelopeItem.cs
@@ -76,6 +76,21 @@ public sealed class EnvelopeItem : ISerializable, IDisposable
             var value => Convert.ToInt64(value) // can be int, long, or another numeric type
         };
 
+    internal long? TryGetOrRecalculateLength()
+    {
+        if (TryGetLength() is { } headerLength)
+        {
+            return headerLength;
+        }
+
+        if (Payload is StreamSerializable streamSerializable)
+        {
+            return streamSerializable.Source.TryGetLength();
+        }
+
+        return null;
+    }
+
     /// <summary>
     /// Returns the file name or null if no name exists.
     /// </summary>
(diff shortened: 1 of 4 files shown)


src/Sentry/Protocol/Envelopes/EnvelopeItem.cs
```diff
@@ -76,6 +76,21 @@ public EnvelopeItem(IReadOnlyDictionary<string, object?> header, ISerializable p
             var value => Convert.ToInt64(value) // can be int, long, or another numeric type
         };
 
+    internal long? TryGetOrRecalculateLength()
```
bitsandfoxes, line 79: I don't understand how we get here or how that fixes the issue - even if it does (which I believe). 
The attachment size did not get read correctly from the header? How/why not? Should it have been set in the first place? And if we're recalculating it here, should it not end up being set now?

jamescrosswell: I'd put all of that in the details for the issue itself... but I just moved it into the details of this PR.

bitsandfoxes: So the length ends up in the header during serialization and gets removed during deserialization because of trust-issues. Fair enough.
But is this not a bug in the caching transport then? Should the length not be set back on the header during serialization?

bitsandfoxes: I'm not trying to bikeshed here, I genuinely don't get how it works.

jamescrosswell: Yeah you have to read Matt's PR where he removes the header upon deserialisation in order to get it. It's linked to in the comment above the line where it gets removed.

bitsandfoxes (approved):

PR #3330, opened 2024-04-25
feat: Check-In Upserts

bitsandfoxes:
Using this would look like this
```
var id = SentrySdk.CaptureCheckIn("my_monitor", CheckInStatus.InProgress, configureMonitorOptions: options =>
{
    options.Interval("0 * * * *");
    // or
    options.Interval(1, MeasurementUnit.Duration.Day);

    options.FailureIssueThreshold = 3;
    options.MaxRuntime = TimeSpan.FromMinutes(5);
    options.Timezone = "America/Los_Angeles";
    options.Owner = "me@there.com";
});
```

I wish I could use `TimeZoneInfo` but Windows has its own thing going on and supports IANA Timezone IDs out of the box starting with .NET 6 ([source](https://stackoverflow.com/a/17348822))
Checking with the other SDKs I don't really see the reason to build our own timezone converter for .NET 5 and older if the user has to specify via `string` which timezone anyway.


src/Sentry/SentryMonitorConfig.cs
```diff
@@ -0,0 +1,115 @@
...
+        switch (Type)
+        {
+            case SentryMonitorScheduleType.Crontab:
+                writer.WriteStringIfNotWhiteSpace("value", Crontab);
+                break;
+            case SentryMonitorScheduleType.Interval:
+                writer.WriteNumberIfNotNull("value", Interval);
+                writer.WriteStringIfNotWhiteSpace("unit", Unit.ToString()?.ToLower());
+                break;
+        }
```
bitsandfoxes, line 92: Admittely, on the hacky-spectrum this might be a bit high, doing some "logic" in the writer but it's just data. I could put this into its own type `MonitorSchedule`?

bruno-garcia:
```
var monitorConfig = SentryMonitorConfig.CreateCronMonitorConfig("0 * * * *");
monitorConfig.FailureIssueThreshold = 3;
monitorConfig.MaxRuntime = TimeSpan.FromMinutes(5);
monitorConfig.Timezone = TimeZoneInfo.FindSystemTimeZoneById("Pacific Standard Time");
monitorConfig.Owner = "me@there.com";

```

This reads like an old C# API and seems very ackward to me. 
Are all those properties optional? Threshold, MaxRuntime, TimeZone? (also casing in C# for this is [TimeZone](https://learn.microsoft.com/en-us/dotnet/api/system.timezone?view=net-8.0) as its two words, as seen on `TimeZoneInfo` above too. Please validate casing before locking in a new PR (as in, merging a PR))

Why do we need an email for owner? I imagine that's optional? Since we use the DSN the cron job gets tied to a project which has a team/owner already.

Consider what's required value, and what's optional.
Take what's required, on the first set of args. And follow by optional parameters.
Additionally you can consider an overloap that takes a full object.

But finally, my understanding of upset was that it would take the same API we use to make a normal check-in. Can we figure out interval through that? The key feature/ask here is that I don't need to create something upfront. It'll take a new check-in it'll create a monitor for it. Maybe it requires a few check-ins to figure out the interval this job is running at before it can start alerting if check-ins are failing. cc @gaprl does it do that by any chance?

If we need some full blown API to create the cron job, do we run that once each time the app starts? If so, should that be on `SentrySdk.Init(o => o.Crons(interval, threshold, etc)` ?

src/Sentry/SentrySdk.cs
```diff
@@ -529,23 +529,34 @@ public static void CaptureSession(SessionUpdate sessionUpdate)
...
         TimeSpan? duration = null,
-        SentryMonitorConfig? monitorConfig = null)
-            => CurrentHub.CaptureCheckIn(
-                monitorSlug,
-                status,
-                sentryId,
-                duration: duration,
-                monitorConfig: monitorConfig);
+        Scope? scope = null,
+        Action<SentryMonitorOptions>? configureMonitorOptions = null)
```
jamescrosswell, line 543: 👍🏻

src/Sentry/SentryMonitorOptions.cs
```diff
@@ -26,6 +26,11 @@ public class SentryMonitorOptions : ISentryJsonSerializable
     /// <param name="cronTab"></param>
     public void Interval(string cronTab)
     {
+        if (_type is not SentryMonitorScheduleType.None)
+        {
+            throw new ArgumentException("You tried to set the interval twice. The Check-Ins interval is supposed to be set only once.");
+        }
+
         _type = SentryMonitorScheduleType.Crontab;
         _crontab = cronTab;
```
jamescrosswell, line 35: According to [our docs](https://develop.sentry.dev/sdk/check-ins/#using-crontab):
> The interval unit. Should be one of year, month, week, day, hour, minute.

I think we should design our API so that people can't misconfigure this then. One way to validate the `cronTab` argument would be to use regular expressions. Alternatively, it doesn't necessarily have to be a string... we have to send a string to Sentry (as part of the protocol) but that might not be the best way for users to configure this via our SDK.

src/Sentry/SentryMonitorOptions.cs
```diff
@@ -37,6 +42,11 @@ public void Interval(string cronTab)
     /// <param name="unit"></param>
     public void Interval(int interval, MeasurementUnit.Duration unit)
     {
+        if (_type is not SentryMonitorScheduleType.None)
+        {
+            throw new ArgumentException("You tried to set the interval twice. The Check-Ins interval is supposed to be set only once.");
+        }
+
         _type = SentryMonitorScheduleType.Interval;
         _interval = interval;
```
jamescrosswell, line 51: Similar to the `cronTab` argument to the overload above, we should either check that the interval + unit here are consistent with the accepted values in the Sentry protocol. Again, although we need to serialize this as a string in order to communicate it to Sentry, internally we could have a more structured representation of this, if that made it easier for the user to express and for our SDK to validate.

jamescrosswell (changes requested):
As we discussed, I think we need to add some more argument validation to ensure users can only configure crontab values that Sentry will accept but otherwise looking good.

src/Sentry/SentryMonitorOptions.cs
```diff
@@ -0,0 +1,140 @@
...
+    /// </summary>
+    public string? Owner { get; set; }
+
+    internal SentryMonitorOptions() { }
+
+    /// <inheritdoc />
+    public void WriteTo(Utf8JsonWriter writer, IDiagnosticLogger? logger)
+    {
+        Debug.Assert(_type != SentryMonitorScheduleType.None, "The Monitor Options do not contain a valid interval." +
+                                                              "Please update your monitor options by setting the Interval.");
```
jamescrosswell, line 93: Shouldn't this be guarded against by the constructors?

bitsandfoxes: The constructor cannot guard against that because the object gets created and only then get's the schedule type set by the user.

src/Sentry/SentryMonitorOptions.cs
bitsandfoxes: Fair point. I added some validation to the crontab.

src/Sentry/SentryMonitorOptions.cs
```diff
@@ -0,0 +1,188 @@
...
+    // Breakdown of the validation
+    // ^(\*|([0-5]?\d))                 Minute  0 - 59
+    // (\s+)(\*|([01]?\d|2[0-3]))       Hour    0 - 23
+    // (\s+)(\*|([1-9]|[12]\d|3[01]))   Day     1 - 31
+    // (\s+)(\*|([1-9]|1[0-2]))         Month   1 - 12
+    // (\s+)(\*|([0-7]))                Weekday 0 - 7
+    // $                                End of string
+    private readonly Regex _crontabValidation = new(
+        @"^(\*|([0-5]?\d))(\s+)(\*|([01]?\d|2[0-3]))(\s+)(\*|([1-9]|[12]\d|3[01]))(\s+)(\*|([1-9]|1[0-2]))(\s+)(\*|([0-7]))$",
+        RegexOptions.Compiled | RegexOptions.CultureInvariant);
```
jamescrosswell, line 68: On .NET 8 we could use [Source Generated regular expressions](https://learn.microsoft.com/en-us/dotnet/standard/base-types/regular-expression-source-generators?pivots=dotnet-8-0) for better performance.

We do this for the [BuiltInSystemDiagnosticsMeters](https://github.com/getsentry/sentry-dotnet/blob/e382fb080102cf654f86566cd53816b208746e62/src/Sentry/BuiltInSystemDiagnosticsMeters.cs#L7), for example.

src/Sentry/SentryMonitorOptions.cs
```diff
@@ -0,0 +1,188 @@
...
+        _type = SentryMonitorScheduleType.Crontab;
+        _crontab = cronTab;
+    }
+
+    /// <summary>
+    /// Set Interval
+    /// </summary>
+    /// <param name="interval"></param>
+    /// <param name="unit"></param>
+    public void Interval(int interval, SentryMonitorInterval unit)
```
jamescrosswell, line 96: Given we have this method, is there any need for the more fragile overload that allows a `string cronTab` argument? Could we just have this more robust way to create intervals (guaranteed by compile time static analysis)?

jamescrosswell (changes requested):
Looking very near ready. See my comment/question about whether we need the two overloads and/or I'd recommend we use a source generated regex in net8.0 if we do retain both overloads.

src/Sentry/SentryMonitorOptions.cs
bitsandfoxes: I'd like to avoid having to reimplement something that lets me run and expect a job every Thursday when crontabs are a somewhat standardized way to set these up.

jamescrosswell (approved):
LGTM. Just the source generators for regex, if you feel like making it as performant as possible.

src/Sentry/SentryMonitorOptions.cs
```diff
@@ -0,0 +1,224 @@
...
+                throw new ArgumentException("RecoveryThreshold has to be set to a number greater than 0.");
+            }
+            _recoveryThreshold = value;
+        }
+    }
+
+    /// <summary>
+    /// A tz database string representing the timezone which the monitor's execution schedule is in (i.e. "America/Los_Angeles").
+    /// </summary>
+    public string? Timezone { get; set; }
```
bruno-garcia, line 167: IIRC I raised this before but it's `TimeZone`: not Timezone: https://learn.microsoft.com/en-us/dotnet/api/system.timezone?view=net-8.0

PR #3333, opened 2024-04-26
chore: iOS device tests

bitsandfoxes:


bruno-garcia (approved):

scripts/device-test.ps1
```diff
@@ -59,7 +59,7 @@ try
         if (!(Get-Command xharness -ErrorAction SilentlyContinue))
         {
             Push-Location ($CI ? $env:RUNNER_TEMP : $IsWindows ? $env:TMP : $IsMacos ? $env:TMPDIR : '/temp')
-            dotnet tool install Microsoft.DotNet.XHarness.CLI --global --version '1.*-*' `
+            dotnet tool install Microsoft.DotNet.XHarness.CLI --global --version '9.*-*' `
```
jamescrosswell, line 62: Nice! How/where did you find this @bitsandfoxes ?

bitsandfoxes:
```
This is ten percent luck, twenty percent skill
Fifteen percent concentrated power of will
Five percent pleasure, fifty percent pain
```

Something had to have changed on the runner itself to start being incompatible with whatever we were trying to do. So I was looking for anything we had pinned down that might have started to act up like that.

Issue #3240, opened 2024-03-21
Just migrated my UWP app to Sentry from AppCenter but the release build is crashing on startup. Debug build runs fine


dpaulino:
### Package

Sentry

### .NET Flavor

.NET

### .NET Version

UWP

### OS

Windows

### SDK Version

4.2.1

### Steps to Reproduce

1. Get ambie white noise's code: https://github.com/jenius-apps/ambie 
2. Edit appsettings.resw > TelemetryApiKey. Add your sentry DSN in that field
3. Deploy debug build. Observe it runs fine
4. But release build crashes

### Expected Result

No crashes on startup

### Actual Result

Crashes on startup

bruno-garcia:
> I'm unblocked but it's not ideal. I want to maximize performance using trimming. AppCenter didn't have issues like this, so I was hoping Sentry would be the same.

yeah checking if u're unblocked as that's the first step. [We made the whole code base AOT friend for Native AOT as I mentioend on twitter](https://sentry.engineering/blog/should-you-could-you-aot). But regarding UWP I'm not sure how much that was tested. Sentry uses only Mac's so Windows is mainly tested in CI and @jamescrosswell, @lucas-zimerman and @vaind use Windows too but are not building UWP apps.

So we'll need a bit of your support figuring out what we need here to sort out. It sounds like it's including the `rd.xml` with those rules as part of the SDK, does that sound right?

jamescrosswell:
@dpaulino maybe you can help me out. I'm having difficulty reproducing this. I basically ran through the steps you outlined:
1. Cloned ambie
2. Set the TelemetryApiKey
3. Ran it in Debug (all good)
4. Ran it in Release mode (again, all good)

What command are you running to build a release version of the application (and so what should I run here to reproduce the issue)?

tipa:
The problem can be reproduce as follows:
- make new UWP app from template in VS22
- comment out this line in the Default.rd.xml file: `<Assembly Name="*Application*" Dynamic="Required All" />`
- add Sentry nuget and call Init, set a valid DSN: `Sentry.SentrySdk.Init(o => o.Dsn = "https://....@.....ingest.sentry.io/....");`
- run in either Release or Debug with csproj property `UseDotNetNativeToolchain` set to `true`
Here's an example project: [sentry_test.zip](https://github.com/getsentry/sentry-dotnet/files/14863466/sentry_test.zip)

vaind:
[...]

PR #3343, opened 2024-05-01
Added warning when attempting to use Sentry with UWP .NET Native

jamescrosswell:
Resolves https://github.com/getsentry/sentry-dotnet/issues/3240

Until we can carve some time out to enable Sentry to work with .NET Native, this PR logs a warning when trying to use Sentry with UWP and `<UseDotNetNativeToolchain>true</UseDotNetNativeToolchain>`.

### DIFF
--- a/src/Sentry/SentrySdk.cs
+++ b/src/Sentry/SentrySdk.cs
@@ -31,8 +31,11 @@ public static partial class SentrySdk
 #if NETSTANDARD
         if (AotHelper.IsDotNetNative)
         {
-            options.LogWarning("Sentry doesn't yet support .NET Native compilation. The SDK will be disabled.");
-            return DisabledHub.Instance;
+            throw new Exception(
+                "Sentry does not support .NET Native compilation. " +
+                "To use Sentry, you must disable the .NET Native toolchain. For further information see: " +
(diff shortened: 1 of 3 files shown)


src/Sentry/SentrySdk.cs
```diff
@@ -29,6 +28,14 @@ internal static IHub InitHub(SentryOptions options)
     {
         options.SetupLogging();
 
+#if NETSTANDARD
+        if (AotHelper.IsDotNetNative)
+        {
+            options.LogWarning("Sentry doesn't yet support .NET Native compilation. The SDK will be disabled.");
```
jamescrosswell, line 34: @bitsandfoxes I'm not sure whether `options.LogWarning` is the right approach... it can easily go unnoticed. Perhaps we should be throwing an exception here instead (but one with a self explanatory error message)?

bitsandfoxes: Yeah. I think we agreed to throw. The reasons were
1. It's in init. So fail fast.
2. It's going to crash anyway. So I'd rather have 1. happen.

bitsandfoxes:
After our sync today we agreed on the following:
If we were to merge this the likelihood of us actually fixing UWP decreases. Instead, we're tracking adding full UWP support in https://github.com/getsentry/sentry-dotnet/issues/3347 and opting to fix it up PR by PR.

Issue #3350, opened 2024-05-06
Missing AddAttachment from memory in SentryHint


tranb3r:
### Problem Statement

`Scope` has two methods to add an attachment: from file or from memory.
`SentryHint` only has one method: from file.

### Solution Brainstorm

Would it be possible to add the missing `AddAttachment(byte[] data...)` to `SentryHint`?

bitsandfoxes:
Sounds like a fair request to me! Is coming!

PR #3352, opened 2024-05-07
feat: Hints now accept `byte[]` as attachment

bitsandfoxes:
Fixes https://github.com/getsentry/sentry-dotnet/issues/3350

### DIFF
--- a/src/Sentry/SentryHint.cs
+++ b/src/Sentry/SentryHint.cs
@@ -59,7 +59,7 @@ public class SentryHint
     internal void AddAttachmentsFromScope(Scope scope) => _attachments.AddRange(scope.Attachments);
 
     /// <summary>
-    /// Creates a new Hint with one or more attachments.
+    /// Takes a path and adds the file as an attachment to the hint.
     /// </summary>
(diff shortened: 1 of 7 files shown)


src/Sentry/SentryHint.cs
```diff
@@ -80,6 +80,30 @@ public SentryHint(string key, object? value)
...
+    /// <param name="fileName">The filename for the attachment</param>
+    /// <param name="type">The type of attachment.</param>
+    /// <param name="contentType">The content type of the attachment.</param>
+    public void AddAttachment(
+        byte[] data,
+        string fileName,
+        AttachmentType type = AttachmentType.Default,
+        string? contentType = null)
+    {
+        if (_options is not null)
```
jamescrosswell, line 96: Do we need to check if options is not null here? The options aren't used in this overload right?

jamescrosswell (approved):
LGTM... just one small bit of redundant code to remove, I think.

src/Sentry/SentryHint.cs
bitsandfoxes: Fair point! Thanks!

Issue #2516, opened 2023-07-27
SentryEvent memory leak with large exception object graphs


danports:
### Package

Sentry

### .NET Flavor

.NET Core

### .NET Version

7.0

### OS

Linux

### SDK Version

3.34.0

### Steps to Reproduce

I'm having an issue where a process in which I've enabled the Sentry Microsoft.Extensions.Logging integration is consistently leaking memory. The issue seems to be most obvious in cases where the `SentryEvent` exception instance references a large object graph - say, a `Microsoft.EntityFrameworkCore.DbUpdateConcurrencyException` that references a `DbContext` with a large number of entities loaded.

### Expected Result

I would expect the `SentryEvent` instances to be processed/sent/etc. and eventually freed.

### Actual Result

Based on the process memory consumption (and memory dumps) it seems like the `SentryEvent` instances are never freed:
![image](https://github.com/getsentry/sentry-dotnet/assets/8728855/6f080d8e-9691-437e-8c82-e2b31991e7c0)

I'm wondering if the Sentry SDK is trying to serialize and send the `DbUpdateConcurrencyException` along with _everything_ in the associated `DbContext`, fails due to limitations on event size, and ends up retrying forever.

A typical `gcroot` from `dotnet-dump analyze` looks like this:
```
          -> 7fa86fe3bc78     System.Threading.Thread
          -> 7fa8421108a0     System.Threading.ExecutionContext
          -> 7fa842110870     System.Threading.AsyncLocalValueMap+TwoElementAsyncLocalValueMap
          -> 7fa8420ab168     System.Collections.Generic.KeyValuePair<Sentry.Scope, Sentry.ISentryClient>[]
          -> 7fa84209c4e0     Sentry.SentryClient
          -> 7fa84209c5f0     Sentry.Internal.BackgroundWorker
          -> 7fa8420a72f8     System.Collections.Concurrent.ConcurrentQueue<Sentry.Protocol.Envelopes.Envelope>
          -> 7fa8420a7338     System.Collections.Concurrent.ConcurrentQueueSegment<Sentry.Protocol.Envelopes.Envelope>
          -> 7fa8420a7420     System.Collections.Concurrent.ConcurrentQueueSegment<Sentry.Protocol.Envelopes.Envelope>+Slot[]
          -> 7fa8663d47d0     Sentry.Protocol.Envelopes.Envelope
[...]

PR #3355, opened 2024-05-08
Replace ConcurrentQueue<T> in BackgroundWorker

jamescrosswell:
Resolves https://github.com/getsentry/sentry-dotnet/issues/2516

That issue has quite a few messages in the history but see here for what finally gave rise to this PR:
- https://github.com/getsentry/sentry-dotnet/issues/2516#issuecomment-2099585468

### Before this PR

image

### After this PR

image

### Performance

I was a worried about performance using a bespoke concurrent queue but we seem to have pretty comparable performance to `ConcurrentQueue<T>` in our specific use case (a bit better in fact). 

| Method                   | N    | Mean     | Error     | StdDev    | Gen0      | Gen1     | Gen2     | Allocated |
|------------------------- |----- |---------:|----------:|----------:|----------:|---------:|---------:|----------:|
| ConcurrentQueueLiteAsync | 1000 | 4.106 ms | 0.0962 ms | 0.2835 ms | 1074.2188 | 507.8125 |  23.4375 |   2.18 MB |
| ConcurrentQueueAsync     | 1000 | 5.073 ms | 0.1282 ms | 0.3781 ms | 1062.5000 | 515.6250 | 125.0000 |   2.21 MB |

I tried a few different variations of the code (using a `LinkedList<T>` and `Queue<T>` for the internal queue) bit plain old `List<T>` turned out to be the fastest by quite a margin (and use the least memory).

### DIFF
--- a/benchmarks/Sentry.Benchmarks/ConcurrentQueueBenchmarks.cs
+++ b/benchmarks/Sentry.Benchmarks/ConcurrentQueueBenchmarks.cs
@@ -22,7 +22,7 @@ public class ConcurrentQueueBenchmarks
             }));
             tasks.Add(Task.Run(() =>
             {
-                if (queue.TryPeek(out var item))
+                while (queue.TryPeek(out var item))
                 {
                     queue.TryDequeue(out _);
                 }
@@ -45,7 +45,7 @@ public class ConcurrentQueueBenchmarks
             }));
             tasks.Add(Task.Run(() =>
             {
-                if (queue.TryPeek(out var item))
+                while (queue.TryPeek(out var item))
                 {
                     queue.TryDequeue(out _);
                 }
--- a/benchmarks/Sentry.Benchmarks/ConcurrentQueueBenchmarks.cs
+++ b/benchmarks/Sentry.Benchmarks/ConcurrentQueueBenchmarks.cs
@@ -1,0 +1,57 @@
+using BenchmarkDotNet.Attributes;
+using Sentry.Internal;
+using Sentry.Protocol.Envelopes;
+
+namespace Sentry.Benchmarks;
+
+public class ConcurrentQueueBenchmarks
+{
+    [Params(1000)]
+    public int N;
+
+    [Benchmark]
+    public async Task ConcurrentQueueLiteAsync()
+    {
+        ConcurrentQueueLite<Envelope> queue = new();
+        List<Task> tasks = new();
+        for (var i = 0; i < N; i++)
+        {
+            tasks.Add(Task.Run(() =>
+            {
+                queue.Enqueue(Envelope.FromEvent(new SentryEvent()));
+            }));
+            tasks.Add(Task.Run(() =>
+            {
+                if (queue.TryPeek(out var item))
+                {
+                    queue.TryDequeue(out _);
+                }
+            }));
+        }
+        await Task.WhenAll(tasks);
+        queue.Clear();
+    }
+
+    [Benchmark]
+    public async Task ConcurrentQueueAsync()
+    {
+        ConcurrentQueue<Envelope> queue = new();
+        List<Task> tasks = new();
+        for (var i = 0; i < N; i++)
+        {
+            tasks.Add(Task.Run(() =>
+            {
+                queue.Enqueue(Envelope.FromEvent(new SentryEvent()));
+            }));
+            tasks.Add(Task.Run(() =>
+            {
+                if (queue.TryPeek(out var item))
+                {
+                    queue.TryDequeue(out _);
+                }
+            }));
+        }
+        await Task.WhenAll(tasks);
+        queue.Clear();
+    }
+}
--- a/src/Sentry/Internal/BackgroundWorker.cs
+++ b/src/Sentry/Internal/BackgroundWorker.cs
@@ -9,13 +9,13 @@ internal class BackgroundWorker : IBackgroundWorker, IDisposable
 {
     private readonly ITransport _transport;
     private readonly SentryOptions _options;
-    // private readonly ConcurrentQueue<Envelope> _queue;
     private readonly ConcurrentQueueLite<Envelope> _queue;
     private readonly int _maxItems;
     private readonly CancellationTokenSource _shutdownSource;
     private readonly SemaphoreSlim _queuedEnvelopeSemaphore;
 
     private volatile bool _disposed;
+    private volatile bool _fatalExceptionEncountered;
     private int _currentItems;
 
     internal event EventHandler? OnFlushObjectReceived;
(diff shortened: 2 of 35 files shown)


src/Sentry/Internal/BackgroundWorker.cs
```diff
@@ -191,6 +191,7 @@ private async Task DoWorkAsync()
         }
         catch (Exception e)
         {
+            _queue.Clear();
```
bitsandfoxes, line 194: Why is clearing it necessary? And wouldn't that work with the `ConcurrentQueue` too?

test/Sentry.Tests/Internals/ConcurrentQueueLiteTests.cs
```diff
@@ -0,0 +1,141 @@
...
+    }
+
+    [Fact]
+    public void IsEmpty_Test()
+    {
+        // Arrange
+        var queue = new ConcurrentQueueLite<int>();
+
+        // Act & Assert
+        queue.IsEmpty.Should().BeTrue();
```
vaind, line 65:
```suggestion
        queue.IsEmpty.Should().BeTrue();
        queue.Enqueue(1);
        queue.IsEmpty.Should().BeFalse();
        queue.TryDequeue(out var dequeuedItem);
        queue.IsEmpty.Should().BeTrue();
```

benchmarks/Sentry.Benchmarks/ConcurrentQueueBenchmarks.cs
```diff
@@ -0,0 +1,57 @@
...
+            tasks.Add(Task.Run(() =>
+            {
+                queue.Enqueue(Envelope.FromEvent(new SentryEvent()));
+            }));
+            tasks.Add(Task.Run(() =>
+            {
+                if (queue.TryPeek(out var item))
+                {
+                    queue.TryDequeue(out _);
+                }
```
vaind, line 28:
```suggestion
                while (queue.TryPeek(out var item))
                {
                    queue.TryDequeue(out _);
                }
```

benchmarks/Sentry.Benchmarks/ConcurrentQueueBenchmarks.cs
```diff
@@ -0,0 +1,57 @@
...
```
vaind, line 51:
```suggestion
                while (queue.TryPeek(out var item))
                {
                    queue.TryDequeue(out _);
                }
```

src/Sentry/Internal/BackgroundWorker.cs
vaind: I assume this is because after the next line, the background worker task ends so the queue would never be processed. This could use a comment if we want to add this feature. Also, if that's the case, adding tasks should also be a no-op after the task has finished, shouldn't it? Although, both of these changes are functional changes so I'd prefer not doing them in this PR and only reserve making the change if they're actually the cause of the problem at hand (leak)

vaind (commented):
Looks pretty good to me except for the tests that are a bit sparse. I've added som suggestions although they're not exhaustive.

src/Sentry/Internal/BackgroundWorker.cs
jamescrosswell: Ah yes, that's an artefact of a change that I (mostly) backed out relating to [this comment](https://github.com/getsentry/sentry-dotnet/issues/2516#issuecomment-2099609891). I was thinking of clearing the queue and refusing to add anything else to the queue if that catch block ever got hit... but the only reason that catch block will ever get hit is if `ConcurrentQueueLite<T>.TryPeek` throws, so it seemed like it would be better just to ensure that never happened.

src/Sentry/Internal/ConcurrentQueueLite.cs
```diff
@@ -0,0 +1,64 @@
...
+                return true;
+            }
+        }
+        item = default;
+        return false;
+    }
+
+    public int Count => _listCounter;
+
+    public bool IsEmpty => _listCounter == 0;
```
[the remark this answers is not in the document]
jamescrosswell: No, good catch. It gets updated in locks but read outside of a lock. I'll remove entirely I think as the List maintains an internal counter anyway (if we need to lock to read this, we may as well just use the `List.Count` property).

src/Sentry/Internal/BackgroundWorker.cs
jamescrosswell: Removed.

modules/sentry-cocoa
jamescrosswell, line 1: This was changed by the Format bot... I think we need to roll this back.

vaind: Yes, can you please do so?

vaind (approved):
LGTM except for the unrelated sentry-cocoa change.

Also, should we change breadcrumbs too in a followup PR? That's the only other place we use concurent queue in (other than tests)

modules/sentry-cocoa
jamescrosswell: I can't no... wierdly the commit hash for the submodule in this branch is the same as that in the main branch. So it shouldn't even be showing up in the diff in this PR 😕

vaind: here you are https://github.com/getsentry/sentry-dotnet/pull/3355/commits/1e3b393eaa040f54102ea31a819af7eccce7d53b

vaind:
```
PS /Users/ivan/dev/sentry-dotnet> cd modules
PS /Users/ivan/dev/sentry-dotnet/modules> git status        
On branch fix/concurrentqueue-memory-leak
Your branch is up to date with 'origin/fix/concurrentqueue-memory-leak'.

Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
        modified:   sentry-cocoa (new commits)

no changes added to commit (use "git add" and/or "git commit -a")
PS /Users/ivan/dev/sentry-dotnet/modules> git checkout main -- ./sentry-cocoa/                                   
PS /Users/ivan/dev/sentry-dotnet/modules> git status                          
On branch fix/concurrentqueue-memory-leak
Your branch is up to date with 'origin/fix/concurrentqueue-memory-leak'.

Changes to be committed:
  (use "git restore --staged <file>..." to unstage)
        modified:   sentry-cocoa

Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
        modified:   sentry-cocoa (new commits)

PS /Users/ivan/dev/sentry-dotnet/modules> git add sentry-cocoa
PS /Users/ivan/dev/sentry-dotnet/modules> git commit -m "fixup sentry-cocoa"  
[fix/concurrentqueue-memory-leak 1e3b393e] fixup sentry-cocoa
 1 file changed, 1 insertion(+), 1 deletion(-)
PS /Users/ivan/dev/sentry-dotnet/modules> git push
```

jamescrosswell: Very odd. I tried:
```
> cd modules/sentry-cocoa 
> git checkout 0da9c56
```

But after that `git status` gave me no files changed. 

In any case, thank you for fixing from your end!

bruno-garcia:
Can we release this asap?

bruno-garcia:
It's out on: https://github.com/getsentry/sentry-dotnet/releases/tag/4.6.0

Issue #3334, opened 2024-04-26
Make span eligible for `Requests`


bitsandfoxes:
https://docs.sentry.io/product/performance/requests/#span-eligibility

### SentryHttpMessageHandler / SentryGraphQLHttpMessageHandler

- ✅ The op field is set to "http.client"
- ✅ The description field contains the HTTP method and the full URL of the request (e.g., "GET http://my.service.io/some-data").
- ❌ If the request is to a relative URL (e.g., "GET /data.json"), the server.address span data value should be set to the server address (e.g., "my.service.io").
- ✅ The http.response.status_code contains the HTTP response status code (e.g., "200").
  - We appear to store this as "Extra" data rather than in the structured contexts, but the spans still show on the Requests page in Sentry 
    https://github.com/getsentry/sentry-dotnet/blob/24c159fac119315e21df6138dc04bfe3ccd573c2/src/Sentry/SentryHttpMessageHandler.cs#L70

    https://github.com/getsentry/sentry-dotnet/blob/bf4e0b3db4df28447d055cb569093c4307e85cf7/src/Sentry/SentryGraphQLHttpMessageHandler.cs#L57

### OpenTelemetry instrumentation

- ❌ The op field is set to `http.client`
  - The operation name is being set by OpenTelemetry to `Microsoft.AspNetCore.Hosting.HttpRequestIn`
- ❌ The description field contains the HTTP method and the full URL of the request (e.g., "GET http://my.service.io/some-data").
  - Only contains the HTTP method:
     https://github.com/getsentry/sentry-dotnet/blob/af2dd7214bf60eb73630f040c7a75e992c4e2d66/src/Sentry.OpenTelemetry/SentrySpanProcessor.cs#L344
- ✅ If the request is to a relative URL (e.g., "GET /data.json"), the server.address span data value should be set to the server address (e.g., "my.service.io").
- ✅ The http.response.status_code contains the HTTP response status code (e.g., "200").

jamescrosswell:
@bitsandfoxes added some details to the issue description.

jamescrosswell:
So it turns out this scenario isn't actually possible in .NET:
> ❌ If the request is to a relative URL (e.g., "GET /data.json"), the server.address span data value should be set to the server address (e.g., "my.service.io").

The only way to use Relative URLs with HttpClient is like this:
```csharp
using var client = new HttpClient(sentryHandler);
client.BaseAddress = new Uri("http://localhost");
[...]

PR #3357, opened 2024-05-09
Ensure outgoing spans show on the Request dashboard in Sentry

jamescrosswell:
Resolves https://github.com/getsentry/sentry-dotnet/issues/3334

Tested using both `Sentry.Samples.Console.Basic` and `Sentry.Samples.OpenTelemetry.Console`. There's sometimes a bit of a delay between transactions showing in the Performance dashboard and the relevant spans showing in the Requests dashboard, but they are showing as expected (with the correct domain name).

### DIFF
--- a/samples/Sentry.Samples.Console.Basic/Sentry.Samples.Console.Basic.csproj
+++ b/samples/Sentry.Samples.Console.Basic/Sentry.Samples.Console.Basic.csproj
@@ -35,7 +35,7 @@
   <ItemGroup>
     <ProjectReference Include="..\..\src\Sentry\Sentry.csproj" />
   </ItemGroup>
-  <ItemGroup>
+  <ItemGroup Condition="'$(TargetFramework)' == 'net462'">
     <PackageReference Include="System.Net.Http" Version="4.3.4" />
   </ItemGroup>
 
--- a/samples/Sentry.Samples.Console.Basic/Sentry.Samples.Console.Basic.csproj
+++ b/samples/Sentry.Samples.Console.Basic/Sentry.Samples.Console.Basic.csproj
@@ -35,5 +35,8 @@
   <ItemGroup>
     <ProjectReference Include="..\..\src\Sentry\Sentry.csproj" />
   </ItemGroup>
+  <ItemGroup>
+    <PackageReference Include="System.Net.Http" Version="4.3.4" />
+  </ItemGroup>
 
 </Project>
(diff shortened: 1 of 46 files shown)


samples/Sentry.Samples.Console.Basic/Sentry.Samples.Console.Basic.csproj
```diff
@@ -35,5 +35,8 @@
   <ItemGroup>
     <ProjectReference Include="..\..\src\Sentry\Sentry.csproj" />
   </ItemGroup>
+  <ItemGroup>
+    <PackageReference Include="System.Net.Http" Version="4.3.4" />
```
bruno-garcia, line 39: Is this needed for modern .NET or just .NET Framework?

bitsandfoxes:
This looks really great! Thanks!
I'm just hesitant to merge this as I'm not sure about the implications: Will this affect alerts or users' dashboards?
Let me check with the team.

src/Sentry/SentryHttpMessageHandler.cs
```diff
@@ -69,6 +69,10 @@ internal SentryHttpMessageHandler(IHub? hub, SentryOptions? options, HttpMessage
             $"{method} {url}" // e.g. "GET https://example.com"
             );
         span?.SetExtra(OtelSemanticConventions.AttributeHttpRequestMethod, method);
+        if (!string.IsNullOrWhiteSpace(request.RequestUri?.Host))
+        {
+            span?.SetExtra(OtelSemanticConventions.AttributeServerAddress, request.RequestUri!.Host);
```
bruno-garcia, line 74: How about this? This way we don't need the `!` (assumign this compiled)

```suggestion
        if (request.RequestUri is not null && span is not null && !string.IsNullOrWhiteSpace(request.RequestUri.Host))
        {
            span.SetExtra(OtelSemanticConventions.AttributeServerAddress, request.RequestUri.Host);
```

bruno-garcia (approved):

samples/Sentry.Samples.Console.Basic/Sentry.Samples.Console.Basic.csproj
jamescrosswell: You're right... I can add `Condition="'$(TargetFramework)' == 'net462'"` here.

bitsandfoxes (approved):

Issue #3315, opened 2024-04-18
Improve OutOfMemory insight


bruno-garcia:
Today a OOM exception isn't really actionable. It shows me where it tried to allocated and blew up, which is helpful.<br>But I dunno how much memory it tried to allocate. How much was available, etc.

![image](https://uploads.linear.app/3ed206af-f87a-40aa-82a0-150bf83e43e6/e4552590-58b0-4fc0-87ad-3244ce6b71df/177bcc6b-d4f6-4888-9119-1cd8b2a4fcd2?signature=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJwYXRoIjoiLzNlZDIwNmFmLWY4N2EtNDBhYS04MmEwLTE1MGJmODNlNDNlNi9lNDU1MjU5MC01OGIwLTRmYzAtODdhZC0zMjQ0Y2U2YjcxZGYvMTc3YmNjNmItZDRmNi00ODg4LTkxMTktMWNkOGIyYTRmY2QyIiwiaWF0IjoxNzc5MTUzNDAwLCJleHAiOjE4MTA3MjM5NjB9.4k7v8llqqxDWy5bxXa-T-uqlI0cGKxeeYK42b4gXRdQ)

([example](<https://sentry.sentry.io/issues/5041230607/?project=1886021&query=is%3Aunresolved+issue.priority%3A%5Bhigh%2C+medium%5D&referrer=issue-stream&statsPeriod=7d&stream_index=5>))

But we have memory information further down in the event detail. It doesn't get displayed very nicely (I created a ticket for that here):

* [https://github.com/getsentry/sentry/issues/69233](<https://github.com/getsentry/sentry/issues/69233>)

[...]

Issue #3199, opened 2024-03-06
Increased CPU and memory overhead while profiling


InterstellarStella:
### Package

Sentry.AspNetCore

### .NET Flavor

.NET

### .NET Version

.Net 7 / .Net 8

### OS

Any (not platform specific)

### SDK Version

4.1.2

### Steps to Reproduce

Enable profiling through the following setup:
``` C#
SentrySdk.Init(options =>
{
    options.Dsn = sentryConfig.Dsn;
    options.TracesSampleRate = 0.1;
    options.ProfilesSampleRate = 0.1;
    options.Environment = sentryConfig.Environment;
    options.DiagnosticLevel = diagnosticLevel;
    options.AddIntegration(new ProfilingIntegration( TimeSpan.FromMilliseconds(500)));
});
```

### Expected Result

The CPU overhead is about +5 to 10%.

### Actual Result

The CPU overhead is around around +70%, which is seen through google cloud monitoring system.
The following Sentry packages are used:
Sentry 4.1.2
Sentry.AspNetCore 4.1.2
Sentry.Profiling 4.1.2

[Slack conversation](https://sentry.slack.com/archives/C02MYEFL2Q7/p1709634118504879)
[Zendesk ticket](https://sentry.zendesk.com/agent/tickets/115025)

bruno-garcia:
cc @vaind 

Could we have some details about what the app is doing? With a repro we can look exactly what's going on and make improvements but with the currently provided information it's hard for us to figure out where the problem is.

Is this a web server? How many requests in parallel are running?

If this has high thrughput, changing sample rate to:

```
    options.TracesSampleRate = 0.001;
    options.ProfilesSampleRate = 0.001;
```

Does that reduce the CPU overhead considerably?

saschash:
@bruno-garcia I tried to put sample rates to 0.001 and there is still quite a significant spike (from 7 % CPU -> 40 % CPU)
In fact, it was about the same as with sampling rate 1.0, which is weird...
The behaviour is, it first jumps to 60% and then stabilises at 40%.
It is a web server, but in sandbox environment, and we are running 20 workers in parallel

I have mentioned before that we are using hang fire to queue background jobs and I have noticed call like that: Sentry.AspNetCore.SentryTracingMiddleware.InvokeAsync(HttpContext context). But when it is a hangfire background job HttpContext will be null, so that might somehow explain what is going on

bruno-garcia:
[...]

PR #3366, opened 2024-05-14
Spike: memory dumps for high memory consumption

jamescrosswell:
POC relates to https://github.com/getsentry/sentry-dotnet/issues/3315

Currently all the code is in one of the samples, but we could battle harden this and initialise it from some SentryOptions if we agreed this makes sense for the SDK.

### DIFF
--- a/samples/Sentry.Samples.OpenTelemetry.Console/GarbageCollectionMonitor.cs
+++ b/samples/Sentry.Samples.OpenTelemetry.Console/GarbageCollectionMonitor.cs
@@ -1,0 +1,36 @@
+namespace Sentry.Samples.OpenTelemetry.Console;
+
+internal class GarbageCollectionMonitor
+{
+    private static Action? OnGarbageCollected;
+
+    ~GarbageCollectionMonitor()
+    {
+        if (Environment.HasShutdownStarted)
+        {
+            return;
+        }
+
+        if (AppDomain.CurrentDomain.IsFinalizingForUnload())
+        {
+            return;
+        }
+
+        CreateDanglingMonitor();
+        OnGarbageCollected?.Invoke();
+    }
+
+    private static void CreateDanglingMonitor()
+    {
+        // ReSharper disable once ObjectCreationAsStatement
+#pragma warning disable CA1806
+        new GarbageCollectionMonitor();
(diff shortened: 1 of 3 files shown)


jamescrosswell:
This issue here is another really good use case for this capability. He has an issue every 90 minutes but can't easily get a memory dump because he's running in Azure:
- https://github.com/getsentry/sentry-dotnet/issues/3199#issuecomment-2146360512

> I can't get a memory dump because Azure reasons. I will now redeploy with SignalR disabled and see what happens.

bruno-garcia:
@jamescrosswell if we bundled this binary in the Sentry SDK, could we shell out and invoke it from within the proc?

jamescrosswell:
> @jamescrosswell if we bundled this binary in the Sentry SDK, could we shell out and invoke it from within the proc?

Yes, indeed. The only thing it needs is for `dotnet-gcdump` to be installed... I think implicitly that means it won't work on Mobile (since you wouldn't have dotnet installed on Mobile) but could be a great diagnostic tool for server apps.

bruno-garcia:
> > @jamescrosswell if we bundled this binary in the Sentry SDK, could we shell out and invoke it from within the proc?
> 
> Yes, indeed. The only thing it needs is for `dotnet-gcdump` to be installed... I think implicitly that means it won't work on Mobile (since you wouldn't have dotnet installed on Mobile) but could be a great diagnostic tool for server apps.

That's what I meant by bundling it. If we add the exe to our package, and with msbuild copy to output when app is published. We'd have it there and know its path. Would that work?

jamescrosswell:
> That's what I meant by bundling it. If we add the exe to our package, and with msbuild copy to output when app is published. We'd have it there and know its path. Would that work?

Ah, I see... yes it looks like that should be possible. The [dotnet-gcdump package](https://www.nuget.org/packages/dotnet-gcdump#supportedframeworks-body-tab) doesn't have any dependencies other than the `net6.0` runtime (or higher). 

People targeting net fx/standard would need to ensure they installed the net6.0 runtime then, but it would work ootb of everyone else.

jamescrosswell:
Replaced by:
- https://github.com/getsentry/sentry-dotnet/pull/3667

Issue #3249, opened 2024-03-28
UWP: "In foreground" context warning + not working


tipa:
### Package

Sentry

### .NET Flavor

.NET

### .NET Version

.NET Standard 2.0

### OS

Windows

### SDK Version

4.2.1

### Steps to Reproduce

- Use Sentry in a UWP app
- Build in Release mode
- trigger crash while app window is in foreground

### Expected Result

1. no warning during build
2. "In Foreground" context shows as "True" in Sentry

### Actual Result

1. Warning during build:

```
C:\Users\user\.nuget\packages\microsoft.net.native.compiler\2.2.12-rel-31116-00\tools\Microsoft.NetNative.targets(809,5): warning :
       MCG : warning MCG0007: Unresolved P/Invoke method 'user32.dll!GetWindowThreadProcessId' for method 'System.Int32 Sentry.Internal.ProcessInfo.GetWindowThreadProcessId(System.IntPtr, System.Int32)'. 
Calling this method would throw exception at runtime. 
Please make sure the P/Invoke either points to a Windows API allowed in UWP applications, or a native DLL that is part of the package. 
If for some reason your P/Invoke does not satisfy those requirements, please use [DllImport(ExactSpelling=true) to indicate that you understand the implications of using non-UWP A 
```

2. "In Foreground" shows as "false" in Sentry

tipa:
Some more context:

The code was introduced in this PR: https://github.com/getsentry/sentry-dotnet/pull/2983/commits/c0e9a3d2a32520d2ed3e534de01877f583cd5352

I copied the code and executed it in both a UWP and a WinUI3 / WinAppSDK app:
- in the WinUI3 app, the code worked as expected, the `ApplicationIsActivated` method returned `true` when the window was in the foreground and `false`, if it was not
- In the UWP app, the `ApplicationIsActivated` method always returned false

Perhaps, on UWP the `Window.Current.Visible` property could serve as a workaround, however it returns true even when the window is visible but not focused. Another option is to use `CoreWindow.GetForCurrentThread().ActivationMode` which returns `ActivatedInForeground` when the app is in foreground and focused

jamescrosswell:
Thanks so much for the sleuthing @tipa ! `CoreWindow.GetForCurrentThread().ActivationMode` sounds promising...

jamescrosswell:
[...]

PR #3372, opened 2024-05-16
Suppress P/Invoke warning for GetWindowThreadProcessId in UWP

jamescrosswell:
Resolves https://github.com/getsentry/sentry-dotnet/issues/3249

### DIFF
--- a/src/Sentry/Internal/ProcessInfo.cs
+++ b/src/Sentry/Internal/ProcessInfo.cs
@@ -149,6 +149,9 @@ internal class ProcessInfo
     [DllImport("user32.dll", CharSet = CharSet.Auto, ExactSpelling = true)]
     private static extern IntPtr GetForegroundWindow();
 
-    [DllImport("user32.dll", CharSet = CharSet.Auto, SetLastError = true)]
+    // GetWindowThreadProcessId is only available in the Windows SDK, so trying to call this from UWP apps will fail.
+    // We wrap use of this in a try/catch as a workaround. However, we need `ExactSpelling = true` here  to suppress
+    // warnings about the use of this API when compiling UWP applications.
+    [DllImport("user32.dll", CharSet = CharSet.Auto, SetLastError = true, ExactSpelling = true)]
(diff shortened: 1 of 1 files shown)


src/Sentry/Internal/ProcessInfo.cs
```diff
@@ -149,6 +149,9 @@ private static DateTimeOffset GetStartupTime()
     [DllImport("user32.dll", CharSet = CharSet.Auto, ExactSpelling = true)]
     private static extern IntPtr GetForegroundWindow();
 
-    [DllImport("user32.dll", CharSet = CharSet.Auto, SetLastError = true)]
+    // GetWindowThreadProcessId is only available in the Windows SDK, so trying to call this from UWP apps will fail.
+    // We wrap use of this in a try/catch as a workaround. However, we need `ExactSpelling = true` here  to suppress
+    // warnings about the use of this API when compiling UWP applications.
+    [DllImport("user32.dll", CharSet = CharSet.Auto, SetLastError = true, ExactSpelling = true)]
```
bitsandfoxes, line 155: I get that the warning stated 
```
Please make sure the P/Invoke either points to a Windows API allowed in UWP applications, or a native DLL that is part of the package. 
If for some reason your P/Invoke does not satisfy those requirements, please use [DllImport(ExactSpelling=true) to indicate that you understand the implications of using non-UWP A 
```
but isn't adding `ExactSpelling=true` more a "Trust me, I know what I'm doing" and not a suppression of a warning?
But does it make a difference? The comment explains it.

bitsandfoxes (approved):

src/Sentry/Internal/ProcessInfo.cs
jamescrosswell: I published a new Sentry.nupkg file after making this change, referenced it in a UWP solution and rebuilt... yes the warnings go away if you do that. It's a very bizarre way to suppress warnings if you ask me (thus the code comment).

Issue #3359, opened 2024-05-10
Add SentrySdk.RunAsyncVoid helper method to Sentry SDK


jamescrosswell:
See https://www.jamescrosswell.dev/posts/catching-async-void-exceptions/

For Sentry, by default we could capture the exception and send it to sentry... and optionally invoke any user defined callback (if one was supplied). 

I just wanted something like this in a button click handler in Maui (that needed to call some async methods on HttpClient - no blocking versions of those methods exist).

PR #3379, opened 2024-05-21
Async void helper

jamescrosswell:
Resolves https://github.com/getsentry/sentry-dotnet/issues/3359

### DIFF
--- a/src/Sentry/AsyncVoid.cs
+++ b/src/Sentry/AsyncVoid.cs
@@ -1,0 +1,44 @@
+using Sentry.Internal;
+
+namespace Sentry;
+
+/// <summary>
+/// Utility class for running `async void` methods safely.
+/// </summary>
+public static class AsyncVoid
+{
+    /// <summary>
+    /// Runs an `async void` method safely.
+    /// </summary>
+    /// <param name="task">Typically either a method group or an async lambda that executes some async void code</param>
+    /// <param name="handler">
+    /// An optional callback that will be run if an exception is thrown. If no callback is provided then by default the
+    /// exception will be captured and sent to Sentry.
+    /// </param>
+    /// <example>
+    /// <code>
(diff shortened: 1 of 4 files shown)


src/Sentry/AsyncVoid.cs
bitsandfoxes, line 1: That's pretty great stuff! I think it should all go through our `SentrySdk.cs` tho. The API should be discoverable from there for one.
Would it make sense to put it behind a "module" like metrics? I.e. `SentrySdk.AsyncVoid.CaptureIfItExplodes`?

jamescrosswell: Sounds like a good idea - people are more likely to find it that way.

As it's just a static method, I didn't bother wrapping it in a module or anything - I've just added it directly as `SentrySdk.RunAsyncVoid`.

bitsandfoxes (approved):
That's awesome!

bruno-garcia:
This is a killer feature! Congtats!

I wonder if we have this on the docs anywhere already, or there's a changelog entry or a blog post to let the world know about it :)

jamescrosswell:
> I wonder if we have this on the docs anywhere already, or there's a changelog entry or a blog post to let the world know about it :)

It would definitely be good to have it in the docs. Although, it doesn't really fit neatly into the current doc structure. 
> * Configuration
> * Usage
> * Unit Testing
> * Enriching Events
> * Data Management
> * Setup [various features]
> * Migration Guide
> * Legacy SDK
> * Troubleshooting

It's not really configuration and it's also not related to any of the other features. The best place I can think to put it is under `Usage`. That elevates it next to some pretty key characteristics of the SDK however... 

I've created a [pull request](https://github.com/getsentry/sentry-docs/pull/10169) to add some docs under **_Usage_** and we can always tweak this if we find a better place to put it.

Issue #3380, opened 2024-05-21
Deprecate `EnableTracing`


bitsandfoxes:
Internal discussion: https://www.notion.so/sentry/DACI-Deprecate-enableTracing-in-SDKs-cae2732d252d41d3ac5cf65038ed78a3

TLDR; The premise was to live in a world of dynamic sampling and users no longer having to think about sampling rates. This turned out not to be the case. So having an additional flag that is not super clear whether is a prerequisite to get performance working is just a pain.

PR #3381, opened 2024-05-23
Marked EnableTracing as Obsolete

jamescrosswell:
Resolves https://github.com/getsentry/sentry-dotnet/issues/3380

### DIFF
--- a/src/Sentry/SentryOptions.cs
+++ b/src/Sentry/SentryOptions.cs
@@ -753,12 +753,14 @@ public class SentryOptions
     /// Indicates whether the performance feature is enabled, via any combination of
     /// <see cref="EnableTracing"/>, <see cref="TracesSampleRate"/>, or <see cref="TracesSampler"/>.
     /// </summary>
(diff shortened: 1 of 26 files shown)


src/Sentry/SentryOptions.cs
```diff
@@ -795,6 +797,7 @@ public IList<SubstringOrRegexPattern> FailedRequestTargets
     ///   </item>
     /// </list>
     /// </summary>
+    [Obsolete("Use TracesSampleRate or TracesSampler instead")]
```
bitsandfoxes, line 800: Does it make sense to be more descriptive here?

bitsandfoxes (approved):

src/Sentry/SentryOptions.cs
jamescrosswell: I played around with the wording on this. There's quite a bit of detail though - I think people will be able to work it out from the docs on the `TracesSampelRate` and `TracesSampler` options.

PR #3382, opened 2024-05-23
chore: update perfview

### LINKED ISSUE #3375: Windows Service memory grows indefinitely with Tracing and Profiling enabled.
From jforward5 on [Discord](https://discord.com/channels/621778831602221064/1240369245779525632/1240369245779525632).

> I've added Sentry to my .Net core project and see the project using an infinite amount of memory.  My question is, how do i make it stop using all the memory?
> 
> If I comment out this code my app runs normally, and only uses about 24MB of memory, but uncommenting causes the memory to grow endlessly by 2-3MB per second:
> builder.ConfigureLogging(loggingBuilder =>
>     {
>         loggingBuilder.AddSentry(o =>
>         {
>             o.Dsn = "https://my_dsn/";
>             o.Debug = true;
>             o.TracesSampleRate = 1.0;
>             o.ProfilesSampleRate = 1.0;
>             o.AddIntegration(new ProfilingIntegration(
>                 TimeSpan.FromMicroseconds(500)
>             ));
>         });
>     });
> 
> I should also mention this is a Windows Service using .Net Core 8

## Demo project to reproduce

> i've attached a sample app that reproduces the issue.  The app is a plain, "empty" windows service.  It is using .Net 8 and Sentry.  Run the app from VS and watch your task manager, you will see the process memory grow as long as the service is running.  If you turn off the profiling and tracing it no longer consumes memory, and stabilizes.  You'll need to add your own "DSN" since i removed mine, for security reasons.

[Memory Test Worker Service.zip](https://github.com/getsentry/sentry-dotnet/files/15340965/Memory.Test.Worker.Service.zip)

vaind:
fixes #3375

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@
 
 - P/Invoke warning for GetWindowThreadProcessId no longer shows when using Sentry in UWP applications ([#3372](https://github.com/getsentry/sentry-dotnet/pull/3372))
 
+### Dependencies
+
+- Update Perfview/TraceEvent to v3.1.10 (patched) ([#3382](https://github.com/getsentry/sentry-dotnet/pull/3382))
+
 ## 4.6.2
 
 ### Fixes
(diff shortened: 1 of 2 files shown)


jamescrosswell (approved):

jamescrosswell:
@bitsandfoxes we should probably make a release including this fix ASAP.

PR #3386, opened 2024-05-24
feat: Minimal Blazor support

### LINKED ISSUE #2021: Sentry support for Blazor WebAssembly
### Problem Statement

This is a feature request for support for Blazor WebAssembly. According to #317, there seems to be at least some level of integration supported, but I couldn't find any installation tutorial. I would love to see support including features like User Feedback Prompt or Tunneling.

### Solution Brainstorm

_No response_

bitsandfoxes:
Resolves #2021 

Docs: https://github.com/getsentry/sentry-docs/pull/10344


.generated.NoMobile.sln
```diff
@@ -180,6 +180,8 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "root", "root", "{233D34AB-9
 		CONTRIBUTING.md = CONTRIBUTING.md
 	EndProjectSection
 EndProject
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Sentry.AspNetCore.Blazor", "src\Sentry.AspNetCore.Blazor\Sentry.AspNetCore.Blazor.csproj", "{7D797C46-EF23-40DD-A8AC-EB1DB3708EB0}"
```
bruno-garcia, line 183: Is this specific to Wasm or this project will contain code for Server too?
If so we might need to name it `Sentry.AspNetCore.Blazor.Wasm` or `Sentry.AspNetCore.Blazor.WebAssembly` (whatever aligns with the upstream names, in .NET itself).

bruno-garcia: Looks like `WebAssembly` is what .NET uses

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Features
+
+- Added a new package `Sentry.AspNetCore.Blazor`. This packages provides you with an extension to `WebAssemblyHostBuilder` to allow SDK configuration via the builder pattern. This package gives us an entry point and the ability to extend the SDKs support and out-of-the-box offering. You can follow the progress and leave feedback either ([here](https://github.com/getsentry/sentry-dotnet/issues/2329)) for extending the support for Blazor Server or ([here](https://github.com/getsentry/sentry-dotnet/issues/2021)) for Blazor WebAssembly support ([#3386](https://github.com/getsentry/sentry-dotnet/pull/3386))
```
bruno-garcia, line 7: This is excellent!

samples/Sentry.Samples.AspNetCore.Blazor.Wasm/Program.cs
```diff
@@ -1,20 +1,19 @@
...
-    Debug.Assert(o.IsGlobalModeEnabled);
-});
 try
 {
     var builder = WebAssemblyHostBuilder.CreateDefault(args);
+    builder.UseSentry(options =>
+    {
+        options.Dsn = "https://eb18e953812b41c3aeb042e666fd3b5c@o447951.ingest.sentry.io/5428537";
+        options.Debug = true;
+        options.IsGlobalModeEnabled = true;
```
bruno-garcia, line 11: Should this be set automagically? Ideally this isnt' a decision the user needs to make

src/Sentry.AspNetCore.Blazor/Sentry.AspNetCore.Blazor.csproj
```diff
@@ -0,0 +1,16 @@
...
+    <VersionSuffix>-preview.1</VersionSuffix>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="..\Sentry.Extensions.Logging\Sentry.Extensions.Logging.csproj" />
+    <ProjectReference Include="..\Sentry\Sentry.csproj" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Microsoft.AspNetCore.Components.WebAssembly" Version="6.0.30" />
```
bruno-garcia, line 14: This leads me to think we might need to do `.WebAssembly` on the project name

src/Sentry.AspNetCore.Blazor/WebAssemblyHostBuilderExtensions.cs
```diff
@@ -0,0 +1,41 @@
...
+    /// <param name="configureOptions"></param>
+    /// <returns></returns>
+    public static WebAssemblyHostBuilder UseSentry(this WebAssemblyHostBuilder builder, Action<SentryBlazorOptions> configureOptions)
+    {
+        builder.Logging.AddSentry<SentryBlazorOptions>(blazorOptions =>
+        {
+            configureOptions(blazorOptions);
+
+            // System.PlatformNotSupportedException: System.Diagnostics.Process is not supported on this platform.
+            blazorOptions.DetectStartupTime = StartupTimeDetectionMode.Fast;
```
bruno-garcia, line 26: Should it be also Global Mode here?

bruno-garcia (commented):
Love the idea of this entry point to start the project!!

samples/Sentry.Samples.AspNetCore.Blazor.Wasm/Program.cs
bitsandfoxes: Good point!

src/Sentry.AspNetCore.Blazor.WebAssembly/WebAssemblyHostBuilderExtensions.cs
```diff
@@ -0,0 +1,43 @@
...
+    /// <param name="builder"></param>
+    /// <param name="configureOptions"></param>
+    /// <returns></returns>
+    public static WebAssemblyHostBuilder UseSentry(this WebAssemblyHostBuilder builder, Action<SentryBlazorOptions> configureOptions)
+    {
+        builder.Logging.AddSentry<SentryBlazorOptions>(blazorOptions =>
+        {
+            configureOptions(blazorOptions);
+
+            // System.PlatformNotSupportedException: System.Diagnostics.Process is not supported on this platform.
```
jamescrosswell, line 25: Why are these comments here?

jamescrosswell (changes requested):
@bitsandfoxes it looks like the even that comes through to Sentry is being marked as Handled even though the exception is [not caught](https://github.com/getsentry/sentry-dotnet/blob/113310f4c65bdc769be22701e05f3e62296853ca/samples/Sentry.Samples.AspNetCore.Blazor.Wasm/Pages/Thrower.razor#L21-L25)... but I think we have a separate issue to track this as it's an issue with our logging integration rather than our blazor wasm integration:
- https://github.com/getsentry/sentry-dotnet/issues/3394

The other thing that is a bit confusing is that the [Nav menu](https://github.com/getsentry/sentry-dotnet/blob/113310f4c65bdc769be22701e05f3e62296853ca/samples/Sentry.Samples.AspNetCore.Blazor.Wasm/Shared/NavMenu.razor#L8) is titled "Throw, Catch and Capture" even though the code is just doing a throw.

image

I had to look at the code to work out that this was a nav menu actually... initially I though that clicking on that link would throw, catch and capture and exception (one with `handled=true`) and that clicking on the button would throw and unhandled exception... demonstrating that the integration will catch both kinds of exceptions. 

So overall, the integration itself looks great... but the Sample project is quite confusing. I think it's worth tweaking the UI in the sample project to make this less confusing.

bitsandfoxes:
Super fair point with the sample! I've updated it and added some buttons and some navigation.
It can now `throw` and it shows up as `unhandled` 
Screenshot 2024-06-18 at 23 27 37

and `throw and capture` and it shows up as `handled`.
Screenshot 2024-06-18 at 23 27 44

jamescrosswell (approved):

jamescrosswell:
> I've updated it and added some buttons and some navigation.

Cool, I added a couple of styles to make the nav a bit more apparent and somewhat "Sentry" looking 😉 

image

Issue #3387, opened 2024-05-27
Instrumenting failed AWS SDK requests throws due to AWS's use of UriCreationOptions.DangerousDisablePathAndQueryCanonicalization


tloten:
Please mark the type framework used:
- [x] ASP.NET Core

Please mark the type of the runtime used:
- [x] .NET Core **8.0**

Please mark the NuGet packages used:
- [x] Sentry 4.6.2
- [x] Sentry.AspNetCore 4.6.2

After updating .Net 7 -> 8, we started getting a few errors when doing AWS S3 accesses:
```
System.InvalidOperationException
GetComponents() may not be used for Path/Query on a Uri instance created with UriCreationOptions.DangerousDisablePathAndQueryCanonicalization.
```

Specifically, it happens when S3 500's, and Sentry attempts to instrument/inspect the error. It doesn't seem to occur on 40x responses from S3.

In .Net 8+, the AWS SDK explicitly sets `DangerousDisablePathAndQueryCanonicalization` when constructing URIs. [See this line in the AWS codebase for details](https://github.com/aws/aws-sdk-net/blob/22a1ec96f18a31b72c4fe1bc79f8a37f2690c7a1/sdk/src/Core/Amazon.Runtime/AmazonServiceClient.cs#L644). Sentry then later calls `.GetComponents()` on these URIs, which is evidently not allowed.

I've created a minimal repro here: https://gist.github.com/tloten/6e3c69fa19fba91773f944ada6de24ae

Full stack trace:
```
System.InvalidOperationException: GetComponents() may not be used for Path/Query on a Uri instance created with UriCreationOptions.DangerousDisablePathAndQueryCanonicalization.
  File "UriExt.cs", line 335, in string Uri.GetComponents(UriComponents components, UriFormat format)
    throw new InvalidOperationException(SR.net_uri_GetComponentsCalledWhenCanonicalizationDisabled);
  ?, in void SentryHttpFailedRequestHandler.DoEnsureSuccessfulResponse(HttpRequestMessage request, HttpResponseMessage response)
  ?, in void SentryHttpMessageHandler.HandleResponse(HttpResponseMessage response, ISpan span, string method, string url)
  ?, in async Task<HttpResponseMessage> SentryMessageHandler.SendAsync(HttpRequestMessage request, CancellationToken cancellationToken)
[...]

PR #3393, opened 2024-05-27
Fixed exception when DangerousDisablePathAndQueryCanonicalization = true

jamescrosswell:
Resolves https://github.com/getsentry/sentry-dotnet/issues/3387

### DIFF
--- a/src/Sentry/Internal/Extensions/UriExtensions.cs
+++ b/src/Sentry/Internal/Extensions/UriExtensions.cs
@@ -1,0 +1,11 @@
+namespace Sentry.Internal.Extensions;
+
+internal static class UriExtensions
+{
+    public static string HttpRequestUrl(this Uri uri)
+    {
+        // This is a safe way to get the HttpRequestUrl, even when DisablePathAndQueryCanonicalization is true
+        // See https://github.com/getsentry/sentry-dotnet/issues/3387
+        return new UriBuilder(uri).Uri.GetComponents(UriComponents.HttpRequestUrl, UriFormat.Unescaped);
+    }
+}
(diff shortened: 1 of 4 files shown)


src/Sentry/Internal/Extensions/UriExtensions.cs
```diff
@@ -0,0 +1,11 @@
+namespace Sentry.Internal.Extensions;
+
+internal static class UriExtensions
+{
+    public static string HttpRequestUrl(this Uri uri)
+    {
+        // This is a safe way to get the HttpRequestUrl, even when DisablePathAndQueryCanonicalization is true
+        // See https://github.com/getsentry/sentry-dotnet/issues/3387
+        return new UriBuilder(uri).Uri.GetComponents(UriComponents.HttpRequestUrl, UriFormat.Unescaped);
```
bitsandfoxes, line 9: I thought calling `.GetComponents()` is the part of the issue?

bitsandfoxes: I see, because it's wrapped in the `UriBuilder`.

bitsandfoxes (approved):
There is no way to do this conditionally? So we don't always have to create a new builder object?

jamescrosswell:
> There is no way to do this conditionally? So we don't always have to create a new builder object?

Unfortunately not. The flags and properties indicating `DangerousDisablePathAndQueryCanonicalization == true` are all private or internal.

The only other option is a try/catch... which would be cheaper if/when this error does not occur but would be very expensive when this error does occur. We could do that - it would avoid allocation of a `UriBuilder` for most folks.

Issue #3395, opened 2024-05-28
ASP.NET Core Telemetry: Failed invoking event handler in anonymous routes


renanrcp:
### Package

Sentry.AspNetCore

### .NET Flavor

.NET

### .NET Version

8.0.0

### OS

Linux

### SDK Version

4.7.0

### Steps to Reproduce

After following the [telemetry tutorial](https://docs.sentry.io/platforms/dotnet/guides/aspnetcore/performance/instrumentation/opentelemetry/#aspnet-core-applications), I started receiving the following error when accessing unauthenticated routes:

```css
[May 28 - 01:16:38] [EROR] [Sentry.ISentryClient][0]
Failed invoking event handler.
System.ArgumentNullException: Value cannot be null. (Parameter 'provider')
   at System.ThrowHelper.Throw(String paramName)
   at Microsoft.Extensions.DependencyInjection.ServiceProviderServiceExtensions.GetService[T](IServiceProvider provider)
   at Sentry.AspNetCore.ScopeExtensions.Populate(Scope scope, HttpContext context, SentryAspNetCoreOptions options)
   at Sentry.AspNetCore.SentryMiddleware.PopulateScope(HttpContext context, Scope scope)
   at Sentry.AspNetCore.SentryMiddleware.<>c__DisplayClass16_0.<InvokeAsync>b__5(Object _, Scope activeScope)
   at Sentry.Scope.Evaluate()
```

After investigation, I discovered that the error is triggered [here in the .NET runtime](https://github.com/dotnet/runtime/blob/5535e31a712343a63f5d7d796cd874e563e5ac14/src/libraries/Microsoft.Extensions.DependencyInjection.Abstractions/src/ServiceProviderServiceExtensions.cs#L23).

This suggests that there is likely an issue when using HttpContext.RequestServices with unauthenticated routes.

Through some tests, I found that the error does not occur if:

- I disable SendDefaultPii.
- I remove AddAspNetCoreInstrumentation.
- I use only routes with mandatory authentication.

Here is what I use in Program.cs to add telemetry:

```cs
builder.WebHost.UseSentry(options =>
{
    options.UseOpenTelemetry();

    options.ExperimentalMetrics = new ExperimentalMetricsOptions
    {
        EnableCodeLocations = true
    };

    options.AddIntegration(new ProfilingIntegration(
        // During startup, wait up to 500ms to profile the app startup code.
        // This could make launching the app a bit slower so comment it out if you
[...]

PR #3401, opened 2024-05-31
Fixed null IServiceProvider in anonymous routes with OpenTelemetry

jamescrosswell:
Resolves https://github.com/getsentry/sentry-dotnet/issues/3395

### DIFF
--- a/src/Sentry.AspNetCore/ScopeExtensions.cs
+++ b/src/Sentry.AspNetCore/ScopeExtensions.cs
@@ -38,7 +38,10 @@ public static class ScopeExtensions
 
         if (options.SendDefaultPii && !scope.HasUser())
         {
-            var userFactory = context.RequestServices.GetService<ISentryUserFactory>();
+            // Although the compiler suggests context.RequestServices cannot be null, in practice it can be, so we
+            // have a null check here.
+            // See: https://github.com/getsentry/sentry-dotnet/issues/3395
+            var userFactory = context.RequestServices?.GetService<ISentryUserFactory>();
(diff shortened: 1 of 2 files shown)


src/Sentry.AspNetCore/ScopeExtensions.cs
```diff
@@ -38,7 +38,10 @@ public static void Populate(this Scope scope, HttpContext context, SentryAspNetC
 
         if (options.SendDefaultPii && !scope.HasUser())
         {
-            var userFactory = context.RequestServices.GetService<ISentryUserFactory>();
+            // Although the compiler suggests context.RequestServices cannot be null, in practice it can be, so we
+            // have a null check here.
+            // See: https://github.com/getsentry/sentry-dotnet/issues/3395
+            var userFactory = context.RequestServices?.GetService<ISentryUserFactory>();
```
bitsandfoxes, line 44: Fair enough. But how do we end up in a state where we cannot locate an `ISentryuserFactory`?

jamescrosswell: If we had a service collection reference, I think we'd find an `ISentryUserFactory` registered on it. I'm not sure what set of circumstances leads to `context.RequestServices` being null. That would require debugging some of the MS code, I think (or possibly the OpenTelemetry codebase). 

Since, in this case, the reference will only be null when nobody is logged in, it doesn't present any difficulty to us... we leave the user context empty in this case (exactly what we want).

bitsandfoxes (approved):

Issue #3312, opened 2024-04-18
Remove FirstChanceException workaround


jamescrosswell:
See comment here:
- https://github.com/microsoft/microsoft-ui-xaml/issues/7160#issuecomment-2060522795

Since the original bug has been fixed, we shouldn't need the workaround anymore.

bitsandfoxes:
Coming through a support ticket:
```
System.NullReferenceException: Object reference not set to an instance of an object.
   at void SentryExceptionExtensions.SetSentryMechanism(Exception ex, string type, string description, bool? handled) in SentryExceptionExtensions.cs:line 38
      ex.Data[Mechanism.MechanismKey] = type;
   at void WinUIUnhandledExceptionIntegration.WinUIUnhandledExceptionHandler(object sender, object e) in WinUIUnhandledExceptionIntegration.cs:line 134
      exception.SetSentryMechanism("Microsoft.UI.Xaml.UnhandledException", description, handled);
   at Delegate EventState.GetEventInvoke()+(object sender, UnhandledExceptionEventArgs e) => { }?
   at int UnhandledExceptionEventHandler.Do_Abi_Invoke(IntPtr thisPtr, IntPtr sender, IntPtr e)
```

Since `Data` is an [empty collection by default](https://arc.net/l/quote/iigibqer) I suspect the exception itself is null.
https://github.com/getsentry/sentry-dotnet/blob/800774ccfe84847afcaa15c61b14a786ed0377ef/src/Sentry/SentryExceptionExtensions.cs#L38

Going up the callstack we get to this:

https://github.com/getsentry/sentry-dotnet/blob/800774ccfe84847afcaa15c61b14a786ed0377ef/src/Sentry/Integrations/WinUIUnhandledExceptionIntegration.cs#L121-L125

Since this workaround now sports a bug we should re-prioritize this.

PR #3411, opened 2024-06-06
chore: Removed `FirstChanceException` workaround for WinUI

bitsandfoxes:
Fixes https://github.com/getsentry/sentry-dotnet/issues/3312

Supposedly, the bug has been fixed and the `Application.UnhandledException` does not strip the message and the stacktrace.
How do we end up in a state where the `.stacktrace` is still `null` and the workaround `fristChanceException` is also `null`?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@
 
 - Fixed null IServiceProvider in anonymous routes with OpenTelemetry ([#3401](https://github.com/getsentry/sentry-dotnet/pull/3401))
 
+### API Changes
+- Removed `FirstChanceException` workaround for WinUI ([#3411](https://github.com/getsentry/sentry-dotnet/pull/3411))
+
 ### Dependencies
 
 - Bump CLI from v2.31.2 to v2.32.1 ([#3398](https://github.com/getsentry/sentry-dotnet/pull/3398))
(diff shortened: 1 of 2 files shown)


jamescrosswell (approved):
LGTM. The only thing I'm confused about is the comment in the PR description:
> How do we end up in a state where the .stacktrace is still null and the workaround fristChanceException is also null?

Assuming that's not blocking this though, all good to merge.

bitsandfoxes:
My question is: The workaround was put there to fix the missing message and stacktrace. Now a user reports a bug where, for this to fail, the stacktrace has to be null (to trigger the workaround) - which is the reason we've got the workaround in there in the first place. And in addition to that the `_lastFirstChanceException` is also null.
I think I'll wait for a repro to check this.

jamescrosswell:
> I think I'll wait for a repro to check this.

I think Matt had a way to reproduce this:
- https://github.com/microsoft/microsoft-ui-xaml/issues/7160#issuecomment-1140542075

I believe that was what originally triggered the introduction of the code that captured the FirstChanceException.

bruno-garcia:
Still open because it requires validation on a Windows machine

jamescrosswell:
Confirmed this is working - receiving exceptions with stack traces from WinUI3 applications.

PR #3420, opened 2024-06-13
fix: Fixed duration calculation for Sentry.Hangfire crons checkIn

vzhurintulaco:
In the current implementation `duration` is wrong and even sometimes negative. As a result, it will be ignored on `Sentry` and crons will fail by timeout or Sentry will display the wrong value.


bitsandfoxes (approved):
Thanks for that!

Issue #3431, opened 2024-06-18
Memory issue/leak with Tracing enabled


jamescrosswell:
Running the following program with tracing enabled results in increasing memory usage which does not appear to be cleaned up by GC.

### Sample program

```csharp
SentrySdk.Init(o =>
{
    o.Dsn = "https://d4d82fc1c2c4032a83f3a29aa3a3aff@fake-sentry.io:65535/2147483647";
    o.Release = "release";
    o.TracesSampleRate = 1.0;
});

while(true)
{
    DoWork();
}

void DoWork()
{
    var transaction = SentrySdk.StartTransaction("name", "operation");

    for (var i = 0; i < 10; i++)
    {
        var span = transaction.StartChild("operation");
        Console.WriteLine($"{i}: {Fibonacci(i)}");
        span.Finish();
    }

    transaction.Finish();

    int Fibonacci(int n) => n switch
    {
        0 => 0,
        1 => 1,
        _ => Fibonacci(n - 1) + Fibonacci(n - 2)
    };
}
```

### Memory profile

Targeting `net6.0`...

<img width="1872" alt="image" src="https://github.com/getsentry/sentry-dotnet/assets/728212/dfd02119-7b44-4b12-bf11-023392c264f6">

PR #3432, opened 2024-06-18
Fix memory leak when tracing is enabled

jamescrosswell:
Resolves https://github.com/getsentry/sentry-dotnet/issues/3431

## Memory Profile prior to this PR

image

## Memory Profile after this PR

image

### DIFF
--- a/src/Sentry/TransactionTracer.cs
+++ b/src/Sentry/TransactionTracer.cs
@@ -167,7 +167,11 @@ public class TransactionTracer : IBaseTracer, ITransactionTracer
     /// <inheritdoc />
     public IReadOnlyDictionary<string, string> Tags => _tags;
 
+#if NETSTANDARD2_1_OR_GREATER
     private readonly ConcurrentBag<ISpan> _spans = new();
+#else
+    private ConcurrentBag<ISpan> _spans = new();
+#endif
 
     /// <inheritdoc />
     public IReadOnlyCollection<ISpan> Spans => _spans;
@@ -337,6 +341,14 @@ public class TransactionTracer : IBaseTracer, ITransactionTracer
                 return null;
             }
         }
+
+        public void Clear()
+        {
+            lock (_lock)
+            {
+                TrackedSpans.Clear();
+            }
+        }
     }
     private readonly LastActiveSpanTracker _activeSpanTracker = new LastActiveSpanTracker();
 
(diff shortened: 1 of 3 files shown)


src/Sentry/TransactionTracer.cs
```diff
@@ -395,4 +410,14 @@ public void Finish(Exception exception, SpanStatus status)
 
     /// <inheritdoc />
     public SentryTraceHeader GetTraceHeader() => new(TraceId, SpanId, IsSampled);
+
+    private void ReleaseSpans()
```
jamescrosswell, line 414: Argh... this won't work actually. Once the tracer has finished it needs to be converted to an actual Transaction 😢

jamescrosswell:
In a follow up PR we might want to replace `ConcurrentBag<T>` with a hand crafted `ConcurrentBagLite<T>` or `ConcurrentCollection<T>`... as I don't think `ConcurrentBag<T>` is the right collection to be used here... see [this stackoverflow thread](https://stackoverflow.com/a/5354468). I think the bag is appropriate when you plan to take out all the items you put into it... so kind of like an any-in-any-out data structure.

We're using `ConcurrentBag<T>` in quite a few other places as well... and I suspect we don't need to - a much simpler data structure could serve our purposes.

src/Sentry/TransactionTracer.cs
```diff
@@ -373,6 +385,9 @@ public void Finish()
 
         // Client decides whether to discard this transaction based on sampling
         _hub.CaptureTransaction(new SentryTransaction(this));
+
+        // Release tracked spans
+        ReleaseSpans();
```
bitsandfoxes, line 390: Do you know why this is required? Is something holding the reference to the `TransactionTracer`? Does that reference need to be released too?

bitsandfoxes (approved):
That seems like quite the gotcha! Thanks for fixing that!
I'm not sure I understand where that leak is coming from tho. The fix implies that the SDK doesn't get rid of the `TransactionTracer`?

src/Sentry/TransactionTracer.cs
jamescrosswell: I'm not sure I fully understand yet, no. The stack overflow thread [linked above](https://github.com/getsentry/sentry-dotnet/pull/3432#issuecomment-2175807165) seems to imply `ConcurrentBag<T>` stores stuff in a thread local and leaves it there, unless you remove it. It seems improbably that we wouldn't have picked up on something so obvious years ago if that was the case though. That would mean massive memory leaks all over our SDK (we use `ConcurrentBag<T>` in a number of places). 

I'll merge this PR, create a release, and then do a bit more digging on potentially replacing `ConcurrentBag<T>` with a light weight custom type that isn't so mysterious. That will require some benchmarking and testing though, as it would be replacing something quite foundational.

Issue #3384, opened 2024-05-23
Envelopes stuck in `__processing` until restart


bitsandfoxes:
When offline, the `CachingTransport` will leave envelopes stuck in `__processing`. Even if the client goes online again and starts sending new events, the old ones remain there.
Only during restart will those envelopes get moved back to the cache and then sent by the worker.

bitsandfoxes:
I might be missing something but how is the retry supposed to work
https://github.com/getsentry/sentry-dotnet/blob/5bd8c8a4297804a90520ca6e59e8e478d46189ea/src/Sentry/Internal/Http/CachingTransport.cs#L162-L166
if we're waiting for a signal here before doing any processing?
https://github.com/getsentry/sentry-dotnet/blob/5bd8c8a4297804a90520ca6e59e8e478d46189ea/src/Sentry/Internal/Http/CachingTransport.cs#L147-L152

jamescrosswell:
Lines 162-170 wait for either 500ms or the cancellation token (so basically just delays 500ms in normal circumstances). 

... then back at the start of the loop, lines 149-151 execute. 

It looks like the only time the `_workerSignal` gets released is here:
https://github.com/getsentry/sentry-dotnet/blob/d64358e1036e9fb7c20e88a10047f7eb09c1871e/src/Sentry/Internal/Http/CachingTransport.cs#L448

Possibly we could call that back on line 166 as well, after delaying. That would try every 500ms when offline though... so some kind of progressive backoff might be good here... 500ms after the first successive failure, 1000ms after the second etc. up to some maximum backoff that gets reset back to 500ms after a successful iteration.

Issue #80, opened 2018-08-16
Offline caching


jwillmer:
I am using sentry in an offshore environment and depending on the position of the vessel we don't have an internet connection. How can I cache the messages and send as soon as the internet is back on?

I saw that other implementations have same needs (JS with offline first progressive web apps). So please think about handling this as a feature request - workaround is also welcome.

bruno-garcia:
@jwillmer Thanks for letting us know of your use case.

We'll take it into account when prioritising.
In fact, we'll need this to support Xamarin.

cpwood:
@jwillmer I have vaguely similar requirement: caching data until the user gives their approval for it to be sent to sentry.io in accordance with our privacy policy.

I ended up implementing my own simplistic ISentryClient which stuck everything in a `ConcurrentQueue<Tuple<SentryEvent, Scope>>`. This then gets bound via `HubAdapter.Instance.BindClient()`.

Then, upon the user giving consent (or, in your case, the internet connection being established), it creates an instance of `SentryClient` within a `using` block and sends the data to sentry.io by calling `CaptureEvent()` for each queued item.

Maybe a similar approach could work for you?

bruno-garcia:
@cpwood that's an interesting usage of the API.
One thing to note though is that `SentryClient` internally hold the `HttpTransport` which in turn has an `HttpClient`.

That means that `SentryClient` should be re-used throughout the lifetime of the app and only disposed when the app is about to terminate (which will flush any queued events).

For caching in disk, one options is refactoring `BackgroundWorker` so that the in-memory queue becomes a strategy which I can replace with something that will store to disk and read from disk before calling the transport. The downside of this is the API would have to offer `async` methods now which would be a leaking abstraction.
Another options is to implement a new `ITransport` that will do the caching. Since that contract is `async`.

So today what we have is:

> `SentryClient` -> `BackgroundWorker` -> `HttpTransport` -> `HttpClient`

Where the first two interfaces offer sync APIs and the two latter offer async APIs. The BackgroundWorker is the one with an internal `Task` calling the transport and awaiting response.

What I propose is to create a decorator `Transport` that will write the `SentryEvent` to disk upon `CaptureEventAsync` and will internally hold a `BackgroundWorker` holding a standard `HttpTransport`.

Looking like this:

> `SentryClient` -> `BackgroundWorker` -> `CacheTransport` -> `BackgroundWorker` -> `HttpTransport` -> `HttpClient`

What do you think?

[...]

PR #3438, opened 2024-06-20
Envelopes that fail to send are now flushed when the transport recovers

jamescrosswell:
Resolves https://github.com/getsentry/sentry-dotnet/issues/3384

## Analysis

Previously, when envelopes failed to send (e.g. due to network availability) these were left in a `__processing` directory. The only way to get them back into the cache directory was to reinitialise the SentrySDK (at which point the SDK checks to see if there are any envelopes left in `__processing` from a previous execution that might have terminated unexpectedly, without being able to flush these out properly. 

## Approach / Solution

We'll detect when the network goes offline the first time we try to send an envelope. We now have to implementations of `INetworkStatusListener`:
1. [MauiNetworkStatusListener](https://github.com/getsentry/sentry-dotnet/blob/main/src/Sentry.Maui/Internal/MauiNetworkStatusListener.cs)
2. [PollingNetworkStatusListener](https://github.com/getsentry/sentry-dotnet/blob/f3591402095f0b4ade96190777fc82cca33e8bf5/src/Sentry/Internal/PollingNetworkStatusListener.cs)

If any errors occur when sending envelopes, regardless of whether we're in the context of a MAUI application or not, one of those network listeners is now used to detect when the network comes back online and to send any pending envelopes. 

The following little console app is useful to test all of this (while monitoring the files in the cache directory in the Finder or Explorer):

```csharp
using System.Diagnostics;
using Sentry.Infrastructure;

SentrySdk.Init(options =>
{
    // options.Dsn = "... Your DSN ...";
    options.DiagnosticLogger = new DebugDiagnosticLogger(SentryLevel.Debug);
    options.CacheDirectoryPath = "cache";
    options.Debug = true;
    options.IsGlobalModeEnabled = true;
});

var i = 0;
ConsoleKeyInfo input;
do
{
    SentrySdk.ConfigureScope(scope => scope.SetExtra("Loop Count", ++i));
    SentrySdk.CaptureMessage("Hello, Sentry!");

    Console.WriteLine("Press y to continue or any other key to quit...");
    input = Console.ReadKey();
    Console.WriteLine();
} while (input.Key == ConsoleKey.Y);

Console.WriteLine("Goodbye!");

public class DebugDiagnosticLogger : DiagnosticLogger
{
    public DebugDiagnosticLogger(SentryLevel minimalLevel) : base(minimalLevel) {}
    protected override void LogMessage(string message) => Debug.WriteLine(message);
}

```


src/Sentry/Internal/Http/CachingTransport.cs
```diff
@@ -278,9 +281,24 @@ private async Task ProcessCacheAsync(CancellationToken cancellation)
...
+            {
+                // Retry envelopes that got stuck in processing while the transport was failing
+                MoveUnprocessedFilesBackToCache();
+                _workerSignal.Release();
+            }
+
             // Signal that we can continue with initialization, if we're using _options.InitCacheFlushTimeout
             _initCacheResetEvent?.Set();
         }
+        catch (Exception e)
```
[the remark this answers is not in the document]
jamescrosswell: > Should we add a `when` clause here? Otherwise, we'd potentially run into a situation where a corrupted envelope gets moved between `processing` and `cache` back and forth?

Yeah, I think we need some mechanism to prevent envelopes that are failing for non-network related reasons from getting stuck in an endless loop. @bruno-garcia [mentioned that](https://github.com/getsentry/sentry-dotnet/issues/80#issuecomment-728352819) in the original discussion around building the `CachingTransport`. See more detailed discussion below...

> But that still does not fix the broken retry-logic we have on 168, right?

By 168 you the task delay here?
https://github.com/getsentry/sentry-dotnet/blob/d2d074bb08ed7c354224136eded29228d3a6057b/src/Sentry/Internal/Http/CachingTransport.cs#L161-L169

I'm not entirely sure why a delay was added here to be honest. The loop won't run again until the `_workerSignal` is released, which (prior to this PR) only happened when an envelope was queued to be sent:
https://github.com/getsentry/sentry-dotnet/blob/d2d074bb08ed7c354224136eded29228d3a6057b/src/Sentry/Internal/Http/CachingTransport.cs#L466

In this PR I tweaked it so that **_if_** the transport failed to send an envelope **_and subsequently_** the transport was able to send some other envelope (a fairly imprecise mechanism for detecting when a network issue might have manifested and then subsequently been resolved) then we ressurrect any envelopes that might have been stuck in processing purgatory and have another go at sending them.

## Comments / Suggested Improvements

### Detecting network errors

We need a better way to detect when envelopes fail to send due to network issues. At the moment, if we have one "bad" envelope that fails to send because because of something freaky about the way it serializes, followed by good envelopes forever then that bad envelope would currently go back and forth between processing and cache forever (just taking up non-disk system resources unnecessarily). 

Maybe, as you suggest, we could use the `Exception.Type` to infer this.

### Detecting network recovery

The trigger for resending an envelope is currently that another envelope is successfully sent (since implicitly, if we were able to send some other envelope, the network must be back online).

A better solution might be some kind of progressive backoff/retry timer to check periodically whether the network is back online... that way the resend would be guaranteed even if no other envelopes get enqueued after the network connection recovers.

### Dealing with non-network errors

Assuming we find a reliable mechanism to detect when there is a network issue and also to detect when the network reconnects again, currently when "retrying" we just copy all of the files that are stuck in processing back into the cache. If 97 of those failed due to non-network related reasons and 3 of them failed due to network related reasons, there's no reason to try to send all 100 of them again. Ideally we'd only resend the envelopes that failed because the network was unavailable when we last tried to send them. 

I think we need a list of those envelopes that we think deserve a second chance, vs those that failed for other reasons.... 

Those that failed for other reasons should probably go in some kind of dead letter queue. I don't see how we can automatically recover from other kinds of failure - these really require human intervention. So I think that might be the ultimate solution to that problem.

jamescrosswell:
So in terms of Detecting network errors, I'm wondering how to get an exhaustive list of all the exceptions types that could be symptomatic of a network failure. I see we have this for Maui:
https://github.com/getsentry/sentry-dotnet/blob/d2d074bb08ed7c354224136eded29228d3a6057b/test/Sentry.Tests/Internals/Http/CachingTransportTests.cs#L535-L542

However Maui stuff is abstracted via the [IConnectivity Interface](https://learn.microsoft.com/en-us/dotnet/api/microsoft.maui.networking.iconnectivity?view=net-maui-8.0&viewFallbackFrom=net-8.0) so there are no guarantees that will work outside of Maui.

I'll use that as a starter for 10 though and cobble something together something that works for a basic console app on my development machine. We can then have a think about how to test to find out what kinds of exceptions might be thrown in "other" scenarios (there are quite a few permutations though so maybe it's just something we improve gradually over time).

src/Sentry/Internal/Http/CachingTransport.cs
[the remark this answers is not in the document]
jamescrosswell: So I finally realised there was a bunch of logic already in the `InnerProcessCacheAsync` method that models out most of what we want. [In our main branch](https://github.com/getsentry/sentry-dotnet/blob/main/src/Sentry/Internal/Http/CachingTransport.cs#L291-L360) (i.e. prior to this PR) that catches any operation cancelled exceptions and re-throws these, then it catches any network exceptions and logs these (but doesn't delete the envelope)... finally it catches any other exceptions, logs these and deletes the envelope. 

So the desired behaviour for failing to send an envelope appears to be simply to discard the envelope. I rolled back plans to create a dead letter queue then as I don't think this is how we want the SDK to behave... we wouldn't ever have any plan to be able to successfully send these messages so that makes sense. 

All I'm doing now then is flagging when we run into a network error and then using that flag after sending an envelope successfully to detect when a previously unavailable network is available again. When that happens, we try to resend any envelopes that failed to send due to the network being unavailable. 

Given the broader logic for the `InnerProcessCacheAsync` method, we know we'll only be resending envelopes that failed due to network connectivity issues (other envelopes would have been deleted). 

So I think this is a pretty good solution. 

The only small improvement that might be made is to throw some kind of timer in there to check, periodically, for network connectivity and resend envelopes as soon as the network comes back on line again... rather than waiting for some other envelope to be sent successfully (which is currently now the network coming back on line gets detected). That feels more like a nice to have than a need to have though.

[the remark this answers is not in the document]
bitsandfoxes: I'm SO looking forward digging deeper there! Thanks for all the sleuthing!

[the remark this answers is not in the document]
jamescrosswell: > I'm SO looking forward digging deeper there! Thanks for all the sleuthing!

You probably want to grab the sample code in the issue description to test this - I've basically been firing up that console app and opening the cache directory in Finder... then playing around with turning the wifi on/off. Functionally, everything should be working now. What people think of my coding style is another matter 😜

src/Sentry/Internal/Http/CachingTransport.cs
```diff
@@ -278,6 +286,8 @@ private async Task ProcessCacheAsync(CancellationToken cancellation)
                 await InnerProcessCacheAsync(file, cancellation).ConfigureAwait(false);
             }
 
+            SendFilesQueuedWhileOffline();
```
bitsandfoxes, line 289: Is this not already covered by the backoff function of your `NetworkConnectivityMonitor`?

src/Sentry/Internal/Http/CachingTransport.cs
```diff
@@ -327,9 +344,16 @@ await using (stream.ConfigureAwait(false))
...
+                    catch (Exception ex) when (IsNetworkError(ex))
                     {
+                        if (Interlocked.Exchange(ref _networkIsUnavailable, 1) == 0)
+                        {
+                            // Initialize a monitor to check when the network comes back online
+                            lock (_connectivityMonitorLock)
+                            {
+                                _connectivityMonitor ??= new NetworkConnectivityMonitor(Host.Value, SendFilesQueuedWhileOffline);
+                            }
+                        }
```
bitsandfoxes, line 356: I get it, and it's kinda cool, but we've now got duplicated functionality on 310-3150 and this. Can we merge this? I.e. set the `NetworkConnectivityMonitor` as a default on the options.NetworkStatusListener?

src/Sentry/Internal/Http/CachingTransport.cs
```diff
@@ -359,6 +383,26 @@ await using (stream.ConfigureAwait(false))
...
+            {
+                _connectivityMonitor?.Dispose();
+                _connectivityMonitor = null;
+            }
+
+            // Retry envelopes that got stuck in processing while the transport was failing
+            MoveUnprocessedFilesBackToCache();
+            _workerSignal.Release();
+        }
+    }
```
bitsandfoxes, line 404: If we pause processing, then we don't need to move he files back to the cache on resume, right?

bitsandfoxes (changes requested):

src/Sentry/Internal/Http/CachingTransport.cs
jamescrosswell: It is but that backoff function is eventually only checking every 32 seconds (and it could be configured to check even less frequently than that if we wanted).

This provides an alternative mechanism that detects network recovery immediately after any envelope is successfully enqueued and sent.

src/Sentry/Internal/Http/CachingTransport.cs
jamescrosswell: Hm, the only `INetworkStatusListener` implementation we currently have is [MauiNetworkStatusListener](https://github.com/getsentry/sentry-dotnet/blob/8056c5064993cc74a021eefb928c4a8e700d3793/src/Sentry.Maui/Internal/MauiNetworkStatusListener.cs#L5) (which depends on stuff in `Microsoft.Maui.Networking`). 

But we could tweak the `NetworkConnectivityMonitor` to implement `INetworkStatusListener`. then if folks are building Maui apps, we register a `MauiNetworkStatusListener` and if they're building non-MAUI apps we register this new thing (maybe we call it a `PollingNetworkStatusListener` or something).

src/Sentry/Internal/Http/CachingTransport.cs
jamescrosswell: Not sure about that... I think the first message that fails to send due to a network connectivity error will end up stuck in the `__processing` folder - but will test.

src/Sentry/Internal/Http/CachingTransport.cs
bitsandfoxes: It's probably worth taking a peek at https://github.com/orgs/getsentry/projects/181/views/1?pane=issue&itemId=45150504 as well.

jamescrosswell: > The network connectivity check could be done earlier (perhaps in HttpTransportBase, or perhaps in BackgroundWorker, or both) such that events will queue until network status is restored, rather than being sent and lost (when not caching).

I think this work is best done separately. It's about how messages are handled when caching is disabled. It could make use of the same `INetworkStatusListener` implementations but otherwise it's probably separate logic.

> We should be able to make other implementations of INetworkStatusListener that are appropriate for each environment (desktop/server, non-MAUI mobile, etc.).

From the links Matt posted, there are a few different approaches:
1. Ping (as we've used in this PR)
2. A Windows API call, which obviously would only work on Windows. 
3. [Network change events](https://learn.microsoft.com/en-us/dotnet/fundamentals/networking/network-info#network-change-events).


#### Windows API

I'd rather not create anything Windows specific, if we can come up with a reliable generic approach.

#### Network change events

The network change events are interesting but I'm not sure how we can use them. Even if the status of one or more network interfaces changes, that doesn't necessarily mean that we can or cannot access the Sentry server.

These could maybe provide an additional trigger to check the network connectivity again using some other mechanism (like ping)... instead of waiting for the timeout. So it might let us more quickly detect when the network becomes available again. I'm not sure how valuable that is to us, however.

jamescrosswell:
On reflection the most recent changes we introduced might be problematic. For non-MAUI applications, we now have this code running using a `PollingNetworkStatusListener`:
https://github.com/getsentry/sentry-dotnet/blob/ea929661f55ab5707471ec1801b26abf39fcbe3f/src/Sentry/Internal/Http/CachingTransport.cs#L307-L312

That will wait, potentially indefinitely, until our `PollingNetworkStatusListener` detects that the network is online again (i.e. until it can successfully ping the Sentry host). Some particularly paranoid network administrators might disable ICMP on their network routers... so there's no guarantee our ping check will work on all networks. In that scenario, as soon as the network is unavailable, our transport will stop sending envelopes and won't start sending these again until the app restarts.

Maybe instead of using ICMP/Ping, we should simply try to establish an HTTP or HTTPS connection with Sentry then... I'll muck around and see what the easiest way is to do that from .NET.

jamescrosswell:
> Maybe instead of using ICMP/Ping, we should simply try to establish an HTTP or HTTPS connection with Sentry then... I'll muck around and see what the easiest way is to do that from .NET.

OK, turned out to be quite trivial... plain old TCP connect. Can't get much more reliable than that!

src/Sentry/Internal/PollingNetworkStatusListener.cs
```diff
@@ -0,0 +1,69 @@
...
+        var uri = new Uri(_options!.Dsn!);
+        return new TcpPing(uri.DnsSafeHost, uri.Port);
+    });
+    private IPing Ping => LazyPing.Value;
+
+    public bool Online
+    {
+        get => Interlocked.Read(ref _networkIsUnavailable) == 0;
+        set => Interlocked.Exchange(ref _networkIsUnavailable, value ? 0 : 1);
+    }
```
jamescrosswell, line 51: I think this is unnecessary... reading or modifying a bool should be atomic.

src/Sentry/Internal/PollingNetworkStatusListener.cs
```diff
@@ -0,0 +1,69 @@
...
+        _backoffFunction = backoffFunction ?? (x => x * 2);
+    }
+
+    private Lazy<IPing> LazyPing => new(() =>
+    {
+        if (_testPing != null)
+        {
+            return _testPing;
+        }
+        var uri = new Uri(_options!.Dsn!);
```
bitsandfoxes, line 42: `!` makes me nervous.
But I guess if we get to this point it's safe to assume it's there?

src/Sentry/Internal/TcpPing.cs
```diff
@@ -0,0 +1,25 @@
...
+
+internal class TcpPing(string hostToCheck, int portToCheck = 443) : IPing
+{
+    private readonly Ping _ping = new();
+
+    public async Task<bool> IsAvailableAsync()
+    {
+        try
+        {
+            using var tcpClient = new TcpClient();
```
bitsandfoxes, line 16: I take it the new TcpClient every 500ms is not an issue?

src/Sentry/Internal/Http/CachingTransport.cs
```diff
@@ -271,6 +272,12 @@ private async Task ProcessCacheAsync(CancellationToken cancellation)
             // Signal that we can start waiting for _options.InitCacheFlushTimeout
             _preInitCacheResetEvent?.Set();
 
+            // Make sure no files got stuck in the processing directory
+            if (_options.NetworkStatusListener is { Online: false } listener)
```
bitsandfoxes, line 276: Is this necessary? Doesn't it await getting back online in the `InnerPorcessCacheAsync`?

bitsandfoxes (approved):

src/Sentry/Internal/TcpPing.cs
jamescrosswell: It won't be every 500ms due to the progressive backoff. There would be a first attempt, a subsequent attempts after 500ms, then 1000ms then 2000ms etc.

We could try to use the same TcpClient but it would need to be reset to be usable again after each unsuccessful connection attempt. I could do some benchmarks to check those two different approaches. It's worth noting that there will only be one `PollingNetworkStatusListener` instance in the application though, so I don't think this is really a hot path.

src/Sentry/Internal/Http/CachingTransport.cs
jamescrosswell: The very first envelope that gets sent after the network fails ends up in the `__processing` directory because we bail out of processing [here](https://github.com/getsentry/sentry-dotnet/blob/2cd230206a79c8a2581dd338e7bf712fde1cc683/src/Sentry/Internal/Http/CachingTransport.cs#L342).

We have to move it back to the cache directory somewhere then.

We could potentially move it back to the cache here instead (immediately after setting `Online = false`):
https://github.com/getsentry/sentry-dotnet/blob/ea929661f55ab5707471ec1801b26abf39fcbe3f/src/Sentry/Internal/Http/CachingTransport.cs#L344-L349

I think the reason I didn't do that is that it requires more major restructuring of some of the tests (like [this one](https://github.com/getsentry/sentry-dotnet/blob/a57ed924468dd149018a311f2c61b4c29cab5542/test/Sentry.Tests/Internals/Http/CachingTransportTests.cs#L546))... but it's probably a tidier solution so I'll revisit.

jamescrosswell: On reflection, I'm not sure this is a better solution. The first attempt to send an envelope takes a while to timeout before the network error is raised. While that's happening, the envelope gets moved back and forth between the cache and processing directories in a loop, which is a lot of unnecessary I/O.

We'd also have a similar issue if/when the operation was cancelled (since that's another reason why envelopes can get stuck in the `__processing` directory). 

So I think it's better the way is is currently. I'll add a comment to the code to provide some context for why we do this.

src/Sentry/Internal/PollingNetworkStatusListener.cs
jamescrosswell: I've added a comment to explain why this is OK.

src/Sentry/Internal/Http/CachingTransport.cs
jamescrosswell: Added some comments in the details for the other issue with relevant info from here, so that we can resolve this conversation.

src/Sentry/Internal/PollingNetworkStatusListener.cs
```diff
@@ -0,0 +1,70 @@
...
+    private IPing Ping => LazyPing.Value;
+
+    private volatile bool _online = true;
+    public bool Online
+    {
+        get => _online;
+        set => _online = value;
+    }
+
+    public async Task WaitForNetworkOnlineAsync(CancellationToken cancellationToken = default)
```
bruno-garcia, line 53: This argument has a default value but I don't see it being tested when a cancellation token is passed.

If one is passed, and triggered, `Delay` would throw an exception that isnt' handled anywhere here. Is that considered?

I'd recommend remove the argument if we're not using anywhere (at least I can't see it being used).

If we're using it, add a test that does that to verify the behavior

src/Sentry/SentryOptions.cs
```diff
@@ -1297,6 +1297,10 @@ public SentryOptions()
                 UriComponents.SchemeAndServer,
                 UriFormat.Unescaped)
         );
+
+#if PLATFORM_NEUTRAL
```
bruno-garcia, line 1301: What does this mean in practice? Did we document in what environments this will work and which ones it wont?

jamescrosswell: It's the flip side to this:
https://github.com/getsentry/sentry-dotnet/blob/b76d536f15cfad9cf19ad0e218e1914b58a9e403/src/Sentry.Maui/Internal/SentryMauiOptionsSetup.cs#L54-L57

With `PLATFORM_NEUTRAL` being [defined here](https://github.com/getsentry/sentry-dotnet/blob/199cda3e3f06baeae3f5e7b286c1b7edc0996dd2/Directory.Build.props#L55-L67).

In practice, I think that means if you lose network connectivity and then regain it again, and you're targeting one of those platforms specifically but you're not using MAUI, then any envelopes that you tried to send while the network was down won't be sent until you restart the application again. 

Another option would be to **_always_** initialise the `SentryOptions.NetworkStatusListener` to a `new PollingNetworkStatusListener` but let the `SentryMauiOptionsSetup` override this... or some variant of that. Technically `SentryOptions.NetworkStatusListener` is public so the user could set this to some custom listener that they've crafted. We don't currently do anything to make sure we don't override the user's listener if they've already set one.

src/Sentry/Internal/PollingNetworkStatusListener.cs
jamescrosswell: Good call. Added test and modified code to ensure a graceful exit if the task is cancelled:
- https://github.com/getsentry/sentry-dotnet/pull/3506

src/Sentry/SentryOptions.cs
jamescrosswell: Have created https://github.com/getsentry/sentry-dotnet/issues/3510 to track this conversation in any case (this PR is closed/merged so easy for the conversation to get lost here).

bitsandfoxes: > We don't currently do anything to make sure we don't override the user's listener if they've already set one.

I think that's the crux. But if we set it up in the option's constructor the user would overwrite it again and we'd be fine, right?

PR #3447, opened 2024-06-25
chore: Replace `<Version>` with `<VersionPrefix>` to allow use of `<VersionSuffix>`

bitsandfoxes:
Turns out, setting `<Version>` sets the version in stone: https://github.com/dotnet/docs/pull/25751/files
And this is why the profiling package slipped through during the `4.0.0`.

```
If you want to use `--version-suffix`, specify `VersionPrefix` and not `Version` in the project file.
```

I've updated the bump-script as well. So craft should be happy with the change.

#skip-changelog

### DIFF
--- a/src/Sentry.AspNetCore.Blazor.WebAssembly/Sentry.AspNetCore.Blazor.WebAssembly.csproj
+++ b/src/Sentry.AspNetCore.Blazor.WebAssembly/Sentry.AspNetCore.Blazor.WebAssembly.csproj
@@ -2,7 +2,7 @@
 
   <PropertyGroup>
     <TargetFramework>net6.0</TargetFramework>
-    <VersionSuffix>-preview.1</VersionSuffix>
+    <VersionSuffix>preview.1</VersionSuffix>
     <RootNamespace>Sentry.AspNetCore.Blazor.WebAssembly</RootNamespace>
   </PropertyGroup>
 
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -1,7 +1,7 @@
 <Project>
 
   <PropertyGroup>
-    <Version>4.7.0</Version>
+    <VersionPrefix>4.7.0</VersionPrefix>
     <LangVersion>12</LangVersion>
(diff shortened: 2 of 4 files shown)


src/Sentry.AspNetCore.Blazor.WebAssembly/Sentry.AspNetCore.Blazor.WebAssembly.csproj
```diff
@@ -2,7 +2,7 @@
 
   <PropertyGroup>
     <TargetFramework>net6.0</TargetFramework>
-    <VersionSuffix>-preview.1</VersionSuffix>
+    <VersionSuffix>preview.1</VersionSuffix>
```
bruno-garcia, line 5: it appends `-` automatically?

bruno-garcia (approved):
looks good, but worth checking the output of the `dotnet pack` to validate everything comes out as expected

bruno-garcia:
@bitsandfoxes also consider releases where we add `-alpha` directly to the version, does that work in this case? (other than the obvious case where the package with VersionSuffix would end up with `-alpha.1-alpha.1`

src/Sentry.AspNetCore.Blazor.WebAssembly/Sentry.AspNetCore.Blazor.WebAssembly.csproj
bitsandfoxes: Yes it does!

bitsandfoxes:
> also consider releases where we add -alpha directly to the version, does that work in this case? (other than the obvious case where the package with VersionSuffix would end up with -alpha.1-alpha.1

For that we might just go with a different name for EA packages. Like we call the `Blazor.Webassembly` `preview`. Then we end up with `5.0.0-alpha.1` for regular packages and `5.0.0-alpha.1-preview.1` for the actual preview ones. Which is actually correct?
Alternatively, we can update the bumpscript to deal with `VersionSuffix`.
But yes, definitely something to be aware of.

PR #3449, opened 2024-06-25
fix: Allow trimming in sample

bitsandfoxes:
Based on the [.NET Docs](https://learn.microsoft.com/en-us/dotnet/core/deploying/trimming/trimming-options?pivots=dotnet-8-0#root-assemblies): `If an assembly is not trimmed, it's considered "rooted", which means that it and all of its statically understood dependencies will be kept. Additional assemblies can be "rooted" by name <TrimmerRootAssembly Include="MyAssembly" />` but that's not what we want to test, right?

#skip-changelog

### DIFF
--- a/samples/Sentry.Samples.Console.Basic/Sentry.Samples.Console.Basic.csproj
+++ b/samples/Sentry.Samples.Console.Basic/Sentry.Samples.Console.Basic.csproj
@@ -38,9 +38,6 @@
   <ItemGroup>
     <ProjectReference Include="..\..\src\Sentry\Sentry.csproj" />
   </ItemGroup>
-  <ItemGroup>
-    <TrimmerRootAssembly Include="Sentry" />
-  </ItemGroup>
   <ItemGroup Condition="'$(TargetFramework)' == 'net462'">
     <PackageReference Include="System.Net.Http" Version="4.3.4" />
   </ItemGroup>


samples/Sentry.Samples.Console.Basic/Sentry.Samples.Console.Basic.csproj
```diff
@@ -38,9 +38,6 @@
   <ItemGroup>
     <ProjectReference Include="..\..\src\Sentry\Sentry.csproj" />
   </ItemGroup>
-  <ItemGroup>
-    <TrimmerRootAssembly Include="Sentry" />
```
jamescrosswell, line 42: This comes from the [Trimming Libraries](https://learn.microsoft.com/en-us/dotnet/core/deploying/trimming/prepare-libraries-for-trimming?pivots=dotnet-8-0#show-all-warnings-with-test-app) documentation. 

> **TrimmerRootAssembly** ensures that every part of the library is analyzed. It tells the trimmer that this assembly is a "root". A "root" assembly means the trimmer analyzes every call in the library and traverses all code paths that originate from that assembly.

Without that build item, we don't see all the trimming errors. We can remove it and then the trimming warnings/errors go away... but I think we should resolve the underlying warnings instead. 

One issue I had is that I can't build that Sample.Console.Basic trimmed when it has a `ProjectReference` reference to Sentry. If it has a `PackageReference` to Sentry, everything works as expected (and I get all the useful warnings from the analyzer). I haven't managed to get to the bottom of that rabbit hole yet (I got dev tooling complexity fatigue and switched to something else for a few days, just to rebuild some sense that I was making progress).

jamescrosswell (changes requested):
I'm not sure we should be reversing this change. What is the motivation?

samples/Sentry.Samples.Console.Basic/Sentry.Samples.Console.Basic.csproj
bitsandfoxes: Maybe I'm misunderstanding what it does. This reads differently to me than what the docs I'm linking to describe.

> What is the motivation?

My intention is to have a working sample that has all the bells and whistles enabled without special treatment for the `Sentry` package. If it doesn't work OOTB and we need to add things then we need to reevaluate our approach.

jamescrosswell: > My intention is to have a working sample that has all the bells and whistles enabled without special treatment for the `Sentry` package.

We could remove this from the sample.

We'll need some other (very similar) project to surface any trimming warnings though... possibly we could build that as an integration test or something.

jamescrosswell (approved):

Issue #3440, opened 2024-06-20
Missing Trace Root


kdma:
### Package

Sentry

### .NET Flavor

.NET Core

### .NET Version

6.0

### OS

Windows

### SDK Version

4.7.0

### Steps to Reproduce

Following the samples I have integrated sentry 
```
webHostBuilder.UseSentry(conf =>
 {
     conf.Dsn = _configuration.GetValue<string>("Sentry:Dsn");
     conf.TracesSampleRate = _configuration.GetValue<double>("Sentry:SampleRate");
 });
```

### Expected Result

I am not sure about the expected result, since this looks like a misconfiguration of some sorts

### Actual Result

 The problem is that when I am looking at the tracing window, for every controller route hit, I see this warning about missing the trace root
  
![image](https://github.com/getsentry/sentry-dotnet/assets/2982255/7da78a61-2f31-4ca0-ad45-f6cc0c79ac17)

jamescrosswell:
Thanks for reaching out @kdma .

I'm seeing this too for any traces from our ASP.NET Core MVC sample:
<img width="1781" alt="image" src="https://github.com/getsentry/sentry-dotnet/assets/728212/43eccba1-1a98-4fa9-b5a2-d16aec0ddf05">

@bitsandfoxes  do you know what the requirements are for traces to show up correctly in the new Traces dashboard?

ericsampson:
I'm also seeing this, it's really not great. Thanks!

bitsandfoxes:
> @bitsandfoxes do you know what the requirements are for traces to show up correctly in the new Traces dashboard?

Not yet but I'm working on getting them!

patrickklaeren:
We just started using this and are amazed at what Sentry are building.

I can see a fix for this was merged 2 days ago. Is this already in circulation or is there a planned release?

jamescrosswell:
> I can see a fix for this was merged 2 days ago. Is this already in circulation or is there a planned release?

Change #3453 was included in the 4.8.0 release. See changelog for details:
- https://github.com/getsentry/sentry-dotnet/blob/main/CHANGELOG.md#480

ericsampson:
@jamescrosswell dang, I was really hoping to get that fix.

Would it be possible for y'all to create a quick-turn 4.9.0 with it?

bitsandfoxes:
> Would it be possible for y'all to create a quick-turn 4.9.0 with it?

We'll follow up with a release tomorrow!

[...]

PR #3453, opened 2024-06-26
Fix/trace root

bitsandfoxes:
Fixes https://github.com/getsentry/sentry-dotnet/issues/3440

Creating a `SentryTraceHeader` out of convenience to create a `TransactionContext` seems wrong in the first place and this time caused an actual bug. The newly created `TransactionContext` automatically sets the spanId of the provided `TraceHeader` as a parent, causing the SDK to end up in a situation without a trace root (a transaction without a parent).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 
 ### Fixes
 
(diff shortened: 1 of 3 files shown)


bruno-garcia (approved):

jamescrosswell (approved):
LGTM

Issue #3058, opened 2024-01-17
Allow uploading commits through `set-commits` 


bruno-garcia:
`sentry-cli` is bundled in the SDK adn invoked via msbuild
* https://github.com/getsentry/sentry-dotnet/issues/2074

The ticket was resolved but 2 items were not completed:

* Uploading [commit info](https://docs.sentry.io/product/cli/releases/#sentry-cli-commit-integration), to unlock [suspect commits](https://docs.sentry.io/product/issues/suspect-commits/)
* [Uploading release details](https://docs.sentry.io/product/cli/releases/), to enable features like [Release Health](https://docs.sentry.io/product/releases/health/), resolving [issue by release](https://docs.sentry.io/product/releases/#track-release-health) and "First seeing in Release X".

But we don't upload any commit information. This would be helpful when doing a CI run for publishing a new version. For example:

```sh
dotnet publish -c Release /p:SentrySetCommits
```

Associating commits allows Sentry to link to source code at the correct commit sha.
![image](https://github.com/getsentry/sentry-dotnet/assets/1633368/caa27b44-6627-4ecc-9972-03bf1f33511f)

As well as suggest suspect commits, and other features.

If we can land this on 4.0.0 we could default to making this call when msbuild `publish` target is running

jamescrosswell:
[Source Link included in the .NET SDK](https://learn.microsoft.com/en-us/dotnet/core/compatibility/sdk/8.0/source-link) for version 8.0 and later.

PR #3462, opened 2024-07-04
Allow creating releases and uploading commits from MSBuild

jamescrosswell:
Resolves https://github.com/getsentry/sentry-dotnet/issues/3058

## Details

There are a couple of extra build properties now:
1. `SentryCreateRelease`
2. `SentrySetCommits`

### SentryCreateRelease

If you set this to true then the Sentry SDK attempts to use the Sentry CLI to automatically [create a release](https://docs.sentry.io/cli/releases/#creating-releases) when building your project. ~~The release *version* can be supplied via a separate `SentryRelease` build property. If no `SentryRelease` property is provided then~~ the SDK will infer the version automatically from the ApplicationId, AppManifest, InformationalVersion, AssembyIdentity or the commit hash.

For example:
```xml
  <PropertyGroup>
    <SentryUrl>...your Sentry URL if self-hosted, or omit this line if using sentry.io...</SentryUrl>
    <SentryOrg>...your org...</SentryOrg>
    <SentryProject>...your project...</SentryProject>
    <SentryCreateRelease>true</SentryCreateRelease>
  </PropertyGroup>
```

### SentrySetCommits

If this property is set then the Sentry SDK will automatically [associate commits with the release](https://docs.sentry.io/cli/releases/#commit-integration). By default, the SDK will set commits with the `--auto` flag but you can override this to provide whatever flags you want via the optional `SentrySetCommitOptions` property. 

For example:
```xml
  <PropertyGroup>
    <SentryUrl>...your Sentry URL if self-hosted, or omit this line if using sentry.io...</SentryUrl>
    <SentryOrg>...your org...</SentryOrg>
    <SentryProject>...your project...</SentryProject>
    <SentrySetCommits>true</SentrySetCommits>
    <SentrySetCommitOptions>--local</SentrySetCommitOptions>
  </PropertyGroup>
```

### DIFF
--- a/src/Sentry/buildTransitive/Sentry.targets
+++ b/src/Sentry/buildTransitive/Sentry.targets
@@ -334,6 +334,11 @@
       <SentryRelease>%(MainAssemblyIdentity.Name)@%(Version)</SentryRelease>
     </PropertyGroup>
 
+    <Exec ConsoleToMSBuild="true"  Condition="'$(SentryRelease)' == ''"
+          Command="sentry-cli releases propose-version">
+      <Output TaskParameter="ConsoleOutput" PropertyName="SentryRelease"/>
+    </Exec>
+
     <Message Importance="High" Text="Sentry Release: $(SentryRelease)" />
   </Target>
 
--- a/src/Sentry/buildTransitive/Sentry.targets
+++ b/src/Sentry/buildTransitive/Sentry.targets
@@ -37,12 +37,14 @@
       <!-- Set defaults for the Sentry properties. -->
       <SentryUploadSymbols Condition="'$(SentryUploadSymbols)' == ''">false</SentryUploadSymbols>
       <SentryUploadSources Condition="'$(SentryUploadSources)' == ''">false</SentryUploadSources>
+      <SentryCreateRelease Condition="'$(SentryCreateRelease)' == ''">false</SentryCreateRelease>
+      <SentrySetCommits Condition="'$(SentrySetCommits)' == ''">false</SentrySetCommits>
 
       <!-- This property controls if the Sentry CLI is to be used at all. Setting false will disable all Sentry CLI usage.
            We're explicitly skipping uploads for Sentry projects because they interfere with CLI integration test asserts. -->
       <UseSentryCLI Condition="
         '$(UseSentryCLI)' == ''
-        and ('$(SentryUploadSymbols)' == 'true' or '$(SentryUploadSources)' == 'true' or $(SentryUploadAndroidProguardMapping) == 'true')
+        and ('$(SentryUploadSymbols)' == 'true' or '$(SentryUploadSources)' == 'true' or $(SentryUploadAndroidProguardMapping) == 'true' or $(SentryCreateRelease) == 'true' or $(SentrySetCommits) == 'true')
         and '$(MSBuildProjectName)' != 'Sentry'
         and !$(MSBuildProjectName.StartsWith('Sentry.'))">true</UseSentryCLI>
     </PropertyGroup>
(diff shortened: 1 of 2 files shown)


bitsandfoxes (commented):
This looks really great! What I don't get: How do we make sure the release reported by the SDK at runtime matches the release created at built time?
Users can set the `<SentryRelease>` but that has no affect on the build artifact, right? Should it?

src/Sentry/buildTransitive/Sentry.targets
```diff
@@ -242,5 +250,119 @@
     <Warning Condition="'$(_SentryCLIExitCode)' != '0'" Text="Sentry CLI could not upload proguard mapping." />
   </Target>
 
+  <UsingTask TaskName="SentryGetReleaseFromInformationalVersion" TaskFactory="RoslynCodeTaskFactory" AssemblyFile="$(MSBuildToolsPath)\Microsoft.Build.Tasks.Core.dll">
```
bitsandfoxes, line 253: Is there a way for us to reuse the logic we use during runtime?

jamescrosswell: I did a bit of experimenting and both the android build properties and the ios/macCatalyst information properties end up in the `AssemblyInformationVersionAttribute` anyway, so we can simply remove the target specific logic for those. 

I've also removed the code that tried to [get the version from the main assembly identity](https://github.com/getsentry/sentry-dotnet/blob/61cca84b41e2cdfecf04ec7cc6b103a1f2282bab/src/Sentry/buildTransitive/Sentry.targets#L330-L335). I'm not actually sure if/how that differs from `Assembly.GetName().Version?.ToString()` anyway (I think that is the main assembly identity) 

Finally, at run time we try to read the release from an environment variable so I changed the build targets to mimic this behaviour. 

That leaves us with:
| At build   | At runtime |
| ------------- | ------------- |
| `$(SentryRelease)` build property  |   |
|   | `SentryOptions.Release`  |
|  `SENTRY_RELEASE` Environment variable | `SENTRY_RELEASE` Environment variable  |
| `AssemblyInformationVersionAttribute`  | `AssemblyInformationVersionAttribute`  |
| `Assembly Name Version`  | `Assembly Name Version`  |

So they now both behave the same unless you override the release via the `$(SentryRelease)` build property (at build time) or via the `SentryOptions` (at runtime).

bitsandfoxes: Oh that's really really nice!

bitsandfoxes:
Can we remove the leading `--` for the commit options? It doesn't seem very intuitive having to add those "manually" from a users POV.

bitsandfoxes:
Just to confirm this: If I were to set a `SentryRelease`, how'd that end up on the event getting sent at runtime? Other than creating a new release, what does it to?

jamescrosswell:
> Can we remove the leading `--` for the commit options? It doesn't seem very intuitive having to add those "manually" from a users POV.

Possibly. I wanted people to be able to pass anything/everything that they could [via the CLI](https://docs.sentry.io/cli/releases/#commit-integration), which could be:
- `--auto`
- `--local`
- `--commit "repo-owner/repo-name@deadbeef"`
- `--commit "repo-owner/repo-name@from..to"`

So yeah, at least all of those examples start with a `--` that we could supply automatically.

One other possibility is they want to supply multiple parameters though... so:
- `--auto --ignore-missing`

In that case, what would we do? From the build settings would we drop the `--` from the first parameter but make them supply it for the second? That seems inconsistent. Or we try to do some fancy parsing to work out which of the args needs to be prefixed with a `--`? That seems fragile.

I figured it was better if we just let them supply the arguments as they would if they were using the CLI.

jamescrosswell:
> Just to confirm this: If I were to set a `SentryRelease`, how'd that end up on the event getting sent at runtime? 

I don't think it does. Similarly, if you set `SentryOptions.Release` at runtime, this won't end up being used to create a release at build time. So those two mechanisms to manually control either the Release that gets created at build or the Release that gets sent as context with events operate independently of one another. 

If you want those things to match, it's probably best not to set the `SentryRelease` build property or `SentryOptions.Release` and instead use one of the mechanisms that ensures these things match (environment variable or one of the multiple options for deriving the release from the assembly). 

>Other than creating a new release, what does it to?

Just what's done by the CLI basically. 

We could rename the build property to `__SentryRelease` if we didn't want people attempting to set this manually. I think we still need some way for them to be able to control the `SentryReleaseOptions` and `SentrySetCommitOptions` used during the build though...

jamescrosswell:
Renamed the `SentryRelease` build property to `_SentryRelease` to make it apparent this is not meant to be fiddled with. I also moved it out of the group of properties that people are meant to fiddle with.

bitsandfoxes (approved):
That's looking awesome!

bruno-garcia:
Is this documented? 
image


If not, how would a user discover it?

jamescrosswell:
> Is this documented? 
> 
> If not, how would a user discover it?

No not documented yet. It is demonstrated in the samples but `SentryReleaseOptions` and `SentrySetCommitOptions` in particular would benefit from some docs... as would an explanation of the order of precedence for where the release gets resolved from.

PR #3464, opened 2024-07-07
Add net8.0 support to sample projects

Tony-Liou:
Upgrade some sample projects' target framework to .NET 8

#skip-changelog

### DIFF
--- a/samples/Sentry.Samples.AspNetCore.Basic/Program.cs
+++ b/samples/Sentry.Samples.AspNetCore.Basic/Program.cs
@@ -1,68 +1,48 @@
-using Microsoft.AspNetCore;
+var builder = WebApplication.CreateBuilder(args);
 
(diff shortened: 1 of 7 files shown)


bitsandfoxes:
Thanks a lot, @Tony-Liou for contributing. Just out of curiosity, were you running into any issues that the samples did not provide answers for? What prompted the update to .NET8?

Tony-Liou:
Recently, I've been trying to add Sentry to my worker service project, which uses the latest project template. I used the generic host sample as a reference, but I missed some necessary code for setting up Sentry. I ignored some of the sample code because it seemed redundant, and I thought it might be outdated. The code I selected wasn't enough to set up Sentry correctly, and it took me a lot of time to find the right way to get it running.
So, I ended up refactoring it to align with the newer project template and more concise in my opinion.

bitsandfoxes (approved):
Again, thanks a lot for this!

PR #3466, opened 2024-07-09
Removed RunAsyncVoid method

jamescrosswell:
After further testing, we can't find any use case for this extension method so we're removing it from the SDK.

Technically this is a breaking change. Alternatively, we could mark it obsolete instead and remove it in the next major release but it's highly improbably that anyone is using this method.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@
 
 - Client reports now include dropped spans ([#3463](https://github.com/getsentry/sentry-dotnet/pull/3463))
 
+### API Changes
+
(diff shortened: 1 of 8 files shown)


bitsandfoxes:
Could you add a comment here on why we're removing it?

jamescrosswell:
> Could you add a comment here on why we're removing it?

I can... it's embarrassing but I had initially thought there were scenarios in which you couldn't catch an exception from an async void method which this extension could handle. It turns out there are no practical scenarios where that's true. The code in UI event handlers can just be wrapped in a try..catch and then there's no need for this extension method. 

It probably doesn't need any more detail than that since it would be the detail of an imaginary problem (not a real one) to which this extension method would be a solution.

bitsandfoxes (approved):
Since apparently it didn't do anything, let's just remove it.

PR #3476, opened 2024-07-11
Resolve Processors from ServiceProvider for Background Services in ASP.NET Core

jamescrosswell:
Resolves https://github.com/orgs/getsentry/projects/181/views/1?pane=issue&itemId=45150310

Currently WIP (sharing for comment).

It kind of works. I need to find a way to deduplicate these since the `Except` that is used currently only compares object references (which is too specific):
https://github.com/getsentry/sentry-dotnet/blob/30bb7b444dd8a6bfee4c06a03931e6391bf9bee6/src/Sentry.AspNetCore/SentryMiddleware.cs#L244-L246


bitsandfoxes:
Thanks for the detailed writeup!

PR #3480, opened 2024-07-12
chore: Make `samples` be standalone

bitsandfoxes:
The idea is to make our `samples` be as standalone and without any automagic as possible.
This makes it easier to copy sample projects, drop provided repro-projects in there, and change the package import from a nuget package to the `.csproj`.

#skip-changelog

### DIFF
--- a/samples/Sentry.Samples.AspNetCore.Blazor.Server/Sentry.Samples.AspNetCore.Blazor.Server.csproj
+++ b/samples/Sentry.Samples.AspNetCore.Blazor.Server/Sentry.Samples.AspNetCore.Blazor.Server.csproj
@@ -5,6 +5,15 @@
     <Nullable>enable</Nullable>
     <ImplicitUsings>enable</ImplicitUsings>
   </PropertyGroup>
+
+  <PropertyGroup Condition="'$(Configuration)' == 'Release'">
+    <SentryOrg>sentry-sdks</SentryOrg>
+    <SentryProject>sentry-dotnet</SentryProject>
+    <SentryUploadSymbols>true</SentryUploadSymbols>
+    <SentryUploadSources>true</SentryUploadSources>
+    <SentryCreateRelease>true</SentryCreateRelease>
+  </PropertyGroup>
+
   <ItemGroup>
     <ProjectReference Include="..\..\src\Sentry.AspNetCore.Blazor.WebAssembly\Sentry.AspNetCore.Blazor.WebAssembly.csproj" />
     <ProjectReference Include="..\..\src\Sentry.AspNetCore\Sentry.AspNetCore.csproj" />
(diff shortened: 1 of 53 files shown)


samples/Sentry.Samples.Console.Customized/Sentry.Samples.Console.Customized.csproj
```diff
@@ -5,6 +5,39 @@
...
+    See https://docs.sentry.io/cli/releases/#creating-releases
+    -->
+    <SentryCreateRelease>true</SentryCreateRelease>
+
+    <!--
+    To associate commits with releases
+    See https://docs.sentry.io/cli/releases/#commit-integration
+    -->
+    <!--    <SentrySetCommits>true</SentrySetCommits>-->
+    <!--    <SentrySetCommitOptions>&#45;&#45;local</SentrySetCommitOptions>-->
```
jamescrosswell, line 38: Is that supposed to be `&#45;&#45;`? Shouldn't it be `--`?

samples/Sentry.Samples.AspNetCore.Blazor.Server/Sentry.Samples.AspNetCore.Blazor.Server.csproj
```diff
@@ -5,6 +5,15 @@
     <Nullable>enable</Nullable>
     <ImplicitUsings>enable</ImplicitUsings>
   </PropertyGroup>
+
+  <PropertyGroup Condition="'$(Configuration)' == 'Release'">
+    <SentryOrg>sentry-sdks</SentryOrg>
+    <SentryProject>sentry-dotnet</SentryProject>
+    <SentryUploadSymbols>true</SentryUploadSymbols>
+    <SentryUploadSources>true</SentryUploadSources>
+    <SentryCreateRelease>true</SentryCreateRelease>
```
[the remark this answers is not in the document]
bitsandfoxes: Good idea!

samples/Sentry.Samples.Console.Customized/Sentry.Samples.Console.Customized.csproj
bitsandfoxes: You cannot have the `--` within a comment. In that case, the IDE automatically converts `--` into the character entity reference. When uncommenting, the IDE turns it back into hyphens. But it'd work either way.

jamescrosswell: OK, bizarre. I would have assumed the escape sequence was `<!--` and so a plain old `--` would be OK but anyway, if it's not allowed then it's not allowed.

jamescrosswell (approved):
LGTM

Issue #3465, opened 2024-07-08
[Android] Crash with `EnableBeforeSend` enabled


akravch:
### Package

Sentry

### .NET Flavor

.NET

### .NET Version

8.0.6

### OS

Android

### SDK Version

4.8.1

### Steps to Reproduce

1. Create an Android app and install `Sentry` NuGet package
2. Configure Sentry so that `EnableBeforeSend` is enabled and `BeforeSend` callback is set to something:

   ```c#
   SentrySdk.Init(options =>
   {
       options.Dsn = "https://****.ingest.us.sentry.io/****";
       options.Native.EnableBeforeSend = true;
       options.SetBeforeSend(@event => @event);
   }
   ```
4. Crash the app with a native crash, e.g. using `SentrySdk.CauseCrash(CrashType.Native)`
5. Restart the app

### Expected Result

`BeforeSend` callback is fired, the crash is sent, the app works normally.

### Actual Result

Crash with the following stack trace:

```
System.FormatException: One of the identified items was in an invalid format.
   at System.Text.Json.JsonElement.GetInt16()
   at Sentry.Protocol.Device.FromJson(JsonElement json)
   at Sentry.SentryContexts.FromJson(JsonElement json)
   at Sentry.Internal.Extensions.MiscExtensions.Pipe[JsonElement,SentryContexts](JsonElement input, Func`2 pipe)
   at Sentry.SentryEvent.FromJson(JsonElement json, Exception exception)
   at Sentry.Android.Extensions.SentryEventExtensions.ToSentryEvent(SentryEvent sentryEvent, SentryOptions javaOptions)
   at Sentry.Android.Callbacks.BeforeSendCallback.Execute(SentryEvent e, Hint h)
   at Sentry.JavaSdk.SentryOptions.IBeforeSendCallbackInvoker.n_Execute_Lio_sentry_SentryEvent_Lio_sentry_Hint_(IntPtr jnienv, IntPtr native__this, IntPtr native_p0, IntPtr native_p1)
   at Android.Runtime.JNINativeWrapper.Wrap_JniMarshal_PPLL_L(_JniMarshal_PPLL_L callback, IntPtr jnienv, IntPtr klazz, IntPtr p0, IntPtr p1) in /Users/runner/work/1/s/xamarin-android/src/Mono.Android/Android.Runtime/JNINativeWrapper.g.cs:line 231
```

[...]

PR #3485, opened 2024-07-15
fix: `FormatException` when reading `BatteryLevel`

bitsandfoxes:
Fixes https://github.com/getsentry/sentry-dotnet/issues/3465

The mobile SDKs have `battery_level` as `float`, causing a `FormatException` when the .NET SDK tries to deserialize the event. We should remove the workaround with the next major.

### DIFF
--- a/src/Sentry/Protocol/Device.cs
+++ b/src/Sentry/Protocol/Device.cs
@@ -361,6 +361,8 @@ public sealed class Device : ISentryJsonSerializable, ICloneable<Device>, IUpdat
         writer.WriteStringIfNotWhiteSpace("model", Model);
         writer.WriteStringIfNotWhiteSpace("model_id", ModelId);
         writer.WriteStringIfNotWhiteSpace("arch", Architecture);
+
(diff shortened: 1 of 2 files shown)


src/Sentry/Protocol/Device.cs
```diff
@@ -426,7 +426,18 @@ public static Device FromJson(JsonElement json)
         var model = json.GetPropertyOrNull("model")?.GetString();
         var modelId = json.GetPropertyOrNull("model_id")?.GetString();
         var architecture = json.GetPropertyOrNull("arch")?.GetString();
-        var batteryLevel = json.GetPropertyOrNull("battery_level")?.GetInt16();
+
+        // TODO: For next mayor: Remove this and change batteryLevel from short to float
```
vaind, line 430:
```suggestion
        // TODO: For next major: Remove this and change batteryLevel from short to float
```

vaind (approved):

jamescrosswell (approved):

Issue #3407, opened 2024-06-04
Memory Leak when using Sentry Profiling


Skyppid:
### Package

Sentry

### .NET Flavor

.NET

### .NET Version

8.0.6

### OS

Windows

### SDK Version

4.7.0

### Steps to Reproduce

Enable profiling like this in the `UseSentry` configuration delegate:

```
o.AddIntegration(new ProfilingIntegration());
if (builder.Environment.IsProduction())
{
    o.ProfilesSampleRate = 0.05;
    o.TracesSampleRate = 0.2;
}
else
{
    o.ProfilesSampleRate = 0.2;
    o.TracesSampleRate = 0.5;
}
```

Run app and wait. It constantly leaks memory even if no requests occur at all.

### Expected Result

No memory leak

### Actual Result

![image](https://github.com/getsentry/sentry-dotnet/assets/2773378/8b718dae-1791-4335-9207-347a377724b7)

As you see the app is completely idle. No requests happening. App just started and did nothing yet. Constantly increases memory.
On our staging cluster the app scaled to 10 pods in only a few minutes due to the memory leak.

Disabling the profiling fixes the problem.

**Further data:**
![image](https://github.com/getsentry/sentry-dotnet/assets/2773378/99d78891-2d76-427f-8aa9-6d2ad511d605)
![image](https://github.com/getsentry/sentry-dotnet/assets/2773378/081fb0e6-d97d-40bc-8889-8fcc04be5620)

As you can see the memory constantly grows. The snapshots show the difference between when the app is fully started and ~2 minutes of idle.

EDIT: This screenshot is after 50m of idle runtime.
![image](https://github.com/getsentry/sentry-dotnet/assets/2773378/f70870f4-b0f9-4a74-b962-d945d90d4850)

bitsandfoxes:
Hey @Skyppid, thanks for reaching out. Do you have a repro-sample for us to easily reproduce this?

klemmchr:
I can confirm huge memory problems with profiling in .NET in general too. We've activated it for our testing system and memory went through the roof while at the same time responses were incredibly slow.

bitsandfoxes:
It looks like we've got duplicate issues going on regarding memory consumption/leak and CPU requirements. I'm closing this in favor of https://github.com/getsentry/sentry-dotnet/issues/3199
Thanks for the detailed reports and the repro @Skyppid 
[...]

PR #3491, opened 2024-07-16
fix: decrease runtime circular buffer for diagnostics client

vaind:
This decreases memory usage (doesn't solve #3407 though)

As for the actual number, it is arbitrary as I haven't found a way to determine a number small enough for all scenarios (number of events per sec/a.k.a. load) that we may encounter.


bitsandfoxes (approved):

bitsandfoxes:
Any idea of the actual impact? Other than "better"?

vaind:
Diagnostic client memory usage grows slowly (based on the cadence of events) by the size of the circullar buffer. So 240 MB lower now

PR #3492, opened 2024-07-16
deps: update perfview

vaind:
Updated perfview dependency after our changes and other fixes were merged
- https://github.com/microsoft/perfview/pull/2068
- https://github.com/microsoft/perfview/pull/1867
- https://github.com/microsoft/perfview/pull/2114


bitsandfoxes (approved):
It's be amazing if that did the trick! 🔥

bruno-garcia (approved):
ideally we add to changelog what we get with the bump. On the submodule bumps it links to diffs at least.  ut here since it was manual we can just mention what we expect to improve

PR #3501, opened 2024-07-25
feat: Added .NET8 as TFM for `Blazor.WebAssembly` package

bitsandfoxes:
Addresses https://github.com/getsentry/sentry-dotnet/issues/2021#issuecomment-2248384178

Since the package has a dependency on `Microsoft.AspNetCore.Components.WebAssembly` adding .NET 8 as TFM allows users to resolve the actual .NET 8 dependency chain.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Features
 
 - Users can now automatically create releases and associated commits via sentry-cli and MSBuild properties ([#3462](https://github.com/getsentry/sentry-dotnet/pull/3462))
+- `Sentry.AspNetCore.Blazor.WebAssembly` now targets .NET 8 specifically, allowing for proper dependency resolution ([#3501](https://github.com/getsentry/sentry-dotnet/pull/3501))
 
 ### Fixes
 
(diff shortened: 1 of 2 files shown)


jamescrosswell (approved):
LGTM

klemmchr:
Should this be done for other libraries like `Sentry.AspNetCore.Grpc` as well?

bitsandfoxes:
> Should this be done for other libraries like Sentry.AspNetCore.Grpc as well?

As far as I can tell, either none of those have TFM version specific dependendencies or they are (like Sentry.AspNetCore) already targeting `.NET8`.

klemmchr:
At least the mentioned package `Sentry.AspNetCore.Grpc` has a dependency on `Sentry.AspNetCore` which has TFM version specific dependencies. I didn't check all packages though.

bitsandfoxes:
And we're building versions for those
https://github.com/getsentry/sentry-dotnet/blob/5ff8acc6710136297ae436967b081278a10a4eae/src/Sentry.AspNetCore/Sentry.AspNetCore.csproj#L4

klemmchr:
Sorry, I was unclear about what I mean. Since Sentry.AspNetCore has TFM version specific dependencies, dependant packages (like [Sentry.AspNetCore.Grpc](https://www.nuget.org/packages/Sentry.AspNetCore.Grpc) and [Sentry.Google.Cloud.Functions](https://www.nuget.org/packages/Sentry.Google.Cloud.Functions)) should build `net8.0` as well but they don't at the moment. Unless I misunderstand the argument made for `Blazor.WebAssembly`.

klemmchr:
From my understanding this does not only apply to direct dependencies but also to downstream dependencies (see https://github.com/getsentry/sentry-dotnet/issues/2021#issuecomment-2215490280). This would affect all libraries that have a dependency on `Sentry.AspNetCore` too.

bitsandfoxes:
Fair enough. Created an issue to track this: https://github.com/getsentry/sentry-dotnet/issues/3598

Issue #3484, opened 2024-07-15
.NET 9 Support


atomsk-0:
### Package

Sentry

### .NET Flavor

.NET

### .NET Version

9.0.100-preview.6.24328.19

### OS

Windows

### SDK Version

4.9.0

### Steps to Reproduce

1. add `<PublishAot>true</PublishAot>` to csproj
2. run `dotnet publish -c Release -r win-x64`
3. try run program

### Expected Result

Program starting up normally

### Actual Result

Throws this exception when trying to init sentry sdk. (Works normally without PublishAot flag enabled
```
System.DllNotFoundException: Unable to load DLL 'sentry-native' or one of its dependencies: The specified module could not be found.
   at System.Runtime.InteropServices.NativeLibrary.LoadLibErrorTracker.Throw(String) + 0x50
   at Internal.Runtime.CompilerHelpers.InteropHelpers.FixupModuleCell(InteropHelpers.ModuleFixupCell*) + 0x10e
   at Internal.Runtime.CompilerHelpers.InteropHelpers.ResolvePInvokeSlow(InteropHelpers.MethodFixupCell*) + 0x2f
   at Sentry.Native.C.sentry_options_new() + 0x32
   at Sentry.Native.C.Init(SentryOptions) + 0x2c
   at Sentry.SentrySdk.InitNativeSdk(SentryOptions) + 0x2b
   at Sentry.SentrySdk.InitHub(SentryOptions) + 0x1d2
   at Sentry.SentrySdk.Init(Action`1) + 0x37
   at Sensum.Utils.SentryClient.Setup() + 0x4e
   at Sensum.Program.Main(String[] args) + 0x16c
```

bitsandfoxes:
Hey @atomsk-0, sorry to see you run into troubles. I wonder if this is related to some changes in .NET 9.

atomsk-0:
ye could be, seems to work fine in .NET 8 project published with NativeAot enabled

bitsandfoxes:
Thanks for confirming! I guess it's time to check out the .NET 9 preview!

bruno-garcia:
Also thanks for trying .NET 9 preview and reporting bugs to Sentry!! this is huge help

andrewmd5:
I'm experiencing this issue even with .NET 8 - does anyone have recommendations for a workaround?

bitsandfoxes:
> I'm experiencing this issue even with .NET 8

[...]

PR #3511, opened 2024-07-31
fix: Sentry.Native.targets

andrewmd5:
I've updated the conditions to include both 'Exe' and 'WinExe' for the OutputType.

This is necessary because:

  1. When `OutputType` is set to `Exe`, it is automatically changed to `WinExe` for WPF and Windows Forms apps that target any framework version, including .NET Framework. (see this [Breaking Change](https://learn.microsoft.com/en-us/dotnet/core/compatibility/sdk/5.0/automatically-infer-winexe-output-type)).

  2. Avalonia uses 'WinExe' when targeting every platform (not just Windows).
  
Previously, when AOT compiling a WinForms or Avalonia app, sentry-native couldn't be found.

This change ensures that sentry-native is correctly linked for these types of applications during AOT compilation, fixing https://github.com/getsentry/sentry-dotnet/issues/3484#issuecomment-2254925496

Here is an Avalonia app that has been AOT compiled using Sentry Native successfully 

Screenshot 2024-07-31 at 11 37 48 AM

### DIFF
--- a/src/Sentry/Platforms/Native/buildTransitive/Sentry.Native.targets
+++ b/src/Sentry/Platforms/Native/buildTransitive/Sentry.Native.targets
@@ -7,21 +7,21 @@
 
   Note:Target framework conditions should be kept synchronized with src/Sentry/buildTransitive/Sentry.Native.targets -->
 <Project>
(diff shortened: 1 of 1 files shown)


bitsandfoxes (approved):
Thanks for the fix!

PR #3513, opened 2024-07-31
feat: `.NET 9` support

bitsandfoxes:
Fixes https://github.com/getsentry/sentry-dotnet/issues/3484#issuecomment-2254925496

# TODO

## iOS Device Tests

Around 700 tests failing, all for a similar reason:

```xml
      <test name="Sentry.Tests.Internals.Extensions.ClientReportExtensionsTests.EnvelopeContainsTransaction_ReportsDroppedSpans" type="Sentry.Tests.Internals.Extensions.ClientReportExtensionsTests" method="EnvelopeContainsTransaction_ReportsDroppedSpans" time="0.0011648" result="Fail">
        <failure exception-type="System.PlatformNotSupportedException">
          <message><![CDATA[System.PlatformNotSupportedException : PlatformNotSupported_ReflectionEmit]]></message>
          <stack-trace><![CDATA[   at System.Reflection.Emit.AssemblyBuilder.EnsureDynamicCodeSupported()
   at System.Reflection.Emit.RuntimeAssemblyBuilder..ctor(AssemblyName n, AssemblyBuilderAccess access)
   at System.Reflection.Emit.AssemblyBuilder.DefineDynamicAssembly(AssemblyName name, AssemblyBuilderAccess access)
   at Castle.DynamicProxy.ModuleScope.CreateModule(Boolean signStrongName)
</stack-trace>
        </failure>
      </test>
```

From what I can tell, [System.Reflection.Emit isn't supported on iOS](https://github.com/castleproject/Core/issues/405)... which begs the question, how was this working previously when targeting net7.0 for our device tests?

## Android Device Tests - proxy not accessible

On Android we're seeing a bunch of errors that are variants of:

> System.ArgumentException : Can not create proxy for type Sentry.ISessionManager because it is not accessible. Make it public, or internal and mark your assembly with [assembly: InternalsVisibleTo("DynamicProxyGenAssembly2, PublicKey=0024000004800000940000000602000000240000525341310004000001000100c547cac37abd99c8db225ef2f6c8a3602f3b3606cc9891605d02baa56104f4cfc0734aa39b93bf7852f7d9266654753cc297e7d2edfe0bac1cdcf9f717241550e0a7b191195b7667bb4f64bcb8e2121380fd1d9d46ad2d92d2d15605093924cceaf74c4861eff62abf69b9291ed0a340e113be11e6a7d3113e92484cf7045cc7")] attribute, because assembly Sentry is strong-named. Arg_ParamName_Name, additionalInterfacesToProxy

In the Directory.Build.props we have the following already ([per the docs](https://nsubstitute.github.io/help/how-nsub-works/#internal-members-and-types)):
https://github.com/getsentry/sentry-dotnet/blob/9b3a07cefdc5acbe4bba65eaebd44dc2a680eba7/Directory.Build.props#L87-L90

It's not clear why this is no longer working with `net9.0`. I've tried, and failed, to recreate the issue with a minimum test application:
- https://github.com/jamescrosswell/net8-mock-internal

## Update bindings

There are some hacks in our bindings script that we may be able to remove, if the fixes have made it into `net9.0`:
https://github.com/getsentry/sentry-dotnet/blob/4c903cfd6ba5dd083b545132f39654230fc3fa39/scripts/generate-cocoa-bindings.ps1#L158-L159

https://github.com/getsentry/sentry-dotnet/blob/4c903cfd6ba5dd083b545132f39654230fc3fa39/scripts/generate-cocoa-bindings.ps1#L170-L171


bruno-garcia:
Looks like it just needs some explicit packagereferences. Same thing happened when I bumped symbol collector to .NET 8
Which references to add are in the error message itself. Might be only on the sample even, not the actualy package

jamescrosswell:
> Looks like it just needs some explicit packagereferences.

Oh it'll be easy (he says)! 😱 20+ commits later...

src/Sentry.Bindings.Android/Sentry.Bindings.Android.csproj
```diff
@@ -12,6 +12,7 @@
   <ItemGroup>
     <None Remove="$(MSBuildThisFileDirectory)..\..\README.md" />
     <None Include="$(MSBuildThisFileDirectory)README.md" Pack="true" PackagePath="" />
+    <None Include="$(MSBuildThisFileDirectory)build/Sentry.Bindings.Android.targets" Pack="true" PackagePath="build/Sentry.Bindings.Android.targets" />
```
jamescrosswell, line 15: @bitsandfoxes I had to add this to prevent [this warning](https://github.com/getsentry/sentry-dotnet/actions/runs/10822485750/job/30026411627?pr=3513):
> error NU5129: Warning As Error: - At least one .targets file was found in 'build/', but 'build/Sentry.Bindings.Android.targets' was not. 

It looks like the NuGet packager expects the targets file to be in the `build/` directory whereas our project file copies/copied it into the `build/$(TargetFramework)$(TargetPlatformVersion)/` folder.

For the time being, I've left it being copied to both destinations since I don't have enough context to know/understand why it was originally being copied to a TargetFramework/PlatformVersion specific folder... something to do with proguard, I'm assuming???

global.json
```diff
@@ -1,7 +1,7 @@
 {
   "sdk": {
-    "version": "8.0.100",
+    "version": "9.0.100-preview.7.24407.12",
```
bruno-garcia, line 3:
```suggestion
    "version": "9.0.100-rc.1.24452.12",
```

samples/Sentry.Samples.GenericHost/Sentry.Samples.GenericHost.csproj
```diff
@@ -21,6 +21,7 @@
   <ItemGroup>
     <ProjectReference Include="..\..\src\Sentry.Extensions.Logging\Sentry.Extensions.Logging.csproj" />
     <PackageReference Include="Microsoft.Extensions.Hosting" Version="8.0.0" />
+    <!-- https://github.com/advisories/GHSA-hh2w-p6rv-4g7w -->
+    <PackageReference Include="System.Text.Json" Version="8.0.4" />
```
bruno-garcia, line 25: This is in the sample. Should we bump in the main SDK since it's a major? Could be time.

could cause diamond dependency with apps running .NET 6 that are pinned on S.T.J 6.0.x though so not sure

src/Sentry.Hangfire/Sentry.Hangfire.csproj
```diff
@@ -14,6 +14,8 @@
   <ItemGroup>
     <ProjectReference Include="..\Sentry\Sentry.csproj" />
     <PackageReference Include="Hangfire.Core" Version="1.8.12" />
+    <!-- https://github.com/advisories/GHSA-5crp-9r3c-p9vr -->
+    <PackageReference Include="Newtonsoft.Json" Version="13.0.3" />
```
bruno-garcia, line 18: Do we need a dependency on N.J? If we can compile and ship the lib without it, it's up to the final app to add this override.

We should suppress the warning here

test/Sentry.DiagnosticSource.IntegrationTests/LocalDbFixture.cs
```diff
@@ -4,6 +4,19 @@ public sealed class LocalDbFixture : IDisposable
...
+            "SqlListenerTests4";
+        #elif NET6_0
+            "SqlListenerTests6";
+        #elif NET7_0
+            "SqlListenerTests7";
+        #elif NET8_0
+            "SqlListenerTests8";
+        #else
+            throw new NotImplementedException("Needs a version specific name to prevent the tests from tripping over one another when running in parallel.");
+        #endif
```
bruno-garcia, line 18:
```suggestion
#if NETFRAMEWORK
            "SqlListenerTests4";
#elif NET6_0
            "SqlListenerTests6";
#elif NET7_0
            "SqlListenerTests7";
#elif NET8_0
            "SqlListenerTests8";
#else
            throw new NotImplementedException("Needs a version specific name to prevent the tests from tripping over one another when running in parallel.");
#endif
```

samples/Sentry.Samples.GenericHost/Sentry.Samples.GenericHost.csproj
jamescrosswell: I think maybe we leave it in the samples but roll back similar fixes that I made to SDK... that way we support as far back as possible and pick up on any issues people (who most likely are bumping to address these advisories) might have as a result of conflicts between their own apps and our SDK.

src/Sentry.Hangfire/Sentry.Hangfire.csproj
jamescrosswell: It's a transitive dependency from `Hangfire.Core` and the version `Hangfire.Core` depends on has advisory warnings. I've tweaked this to add a reference to the same versions as Hangfire.Core and to suppress the warning:
https://github.com/getsentry/sentry-dotnet/blob/328a95eb00f7a50906e01c808511c29060392646/src/Sentry.Hangfire/Sentry.Hangfire.csproj#L19-L33

bruno-garcia: I don't think we should add this though. We could force a consumer into a diamond dependency situation.
Can we skip the warning?

src/Sentry.Hangfire/Sentry.Hangfire.csproj
```diff
@@ -16,6 +16,22 @@
     <PackageReference Include="Hangfire.Core" Version="1.8.12" />
   </ItemGroup>
 
+  <ItemGroup Condition="'$(TargetFramework)' == 'net462'">
+    <PackageReference Include="Newtonsoft.Json" Version="[5.0.1,)">
```
bruno-garcia, line 20: Oh but we're still adding a new dependency here. Couldn't we just leave it out and suppress the warning?

Let the final app deal with it, since we don't depend on it directly.

jamescrosswell: > Couldn't we just leave it out and suppress the warning?

If I try to add `<NoWarn>NU1903</NoWarn>` at the project level (no specific reference to this package) then we get another error message: `CS8002 Referenced assembly does not have a strong name.`. 

We also have this, but I think it's only relevant when calling `dotnet pack` (not `dotnet build`):
https://github.com/getsentry/sentry-dotnet/blob/c7b966b9a70ff504e2f06d88f34c69f1aa25b8b8/Directory.Build.props#L19-L20

The only way that I've found to suppress the warning is to add it specifically within the `PacakgeReference`... and yeah, not ideal, since we end up with an additional explicit dependency here (even though it's actually a transitive dependency).

bruno-garcia: we need to open a ticket with Microsoft then. We shouldn't add a dependency as a workaround to disable a warning.
Rather disable warning as error on this project and keep the warning for the time being

jamescrosswell: I'm an idiot... I was adding this: `<NoWarn>NU1903</NoWarn>` when I should have been adding `<NoWarn>$(NoWarn);NU1903</NoWarn>`. 

As a result, this was being overwritten, which was causing the `CS8002`:
https://github.com/getsentry/sentry-dotnet/blob/c7b966b9a70ff504e2f06d88f34c69f1aa25b8b8/Directory.Build.props#L16-L17

Anyway, fixed now 😝

Directory.Build.props
```diff
@@ -2,6 +2,8 @@
 
   <PropertyGroup>
     <VersionPrefix>4.12.0</VersionPrefix>
+    <!-- TODO: Remove Suffix for 5.0.0 release   -->
+    <VersionSuffix>alpha</VersionSuffix>
```
bruno-garcia, line 6: Should we just add this by hand when releasing>?

test/MauiTestUtils/Directory.Build.props
```diff
@@ -2,6 +2,7 @@
 
   <PropertyGroup>
     <LangVersion>11</LangVersion>
+    <Nullable>disable</Nullable>
```
bruno-garcia, line 5: This is new?

test/Sentry.Tests/Internals/ILSpy/SingleFileAppTests.cs
```diff
@@ -30,7 +30,9 @@ static SingleFileAppTests()
         InValidBundleFile = Path.Combine(testRoot, invalidBundle);
     }
 
-#if NET8_0
+#if NET9_0
+    private static string TargetFramework => "net9.0";
```
bruno-garcia, line 34: A revew for line 45. Instead of `throw` which will cause it to compile and fail at runtime, use:

```
#error "Target Framework not yet supported for single file apps"
```

bruno-garcia (commented):
reviewed the file throughout and LGTM btw

bruno-garcia:
@jamescrosswell worth trying RC2 https://github.com/dotnet/announcements/issues/330#issue-2574195812

bruno-garcia:
We need to squash this for sure so all this code will goes as the PR author. It's a lot of @jamescrosswell code at this point so does it make sense for @jamescrosswell to take over this as he'll be the name popping on blame to do maitenance after anyway?

jamescrosswell:
> We need to squash this for sure so all this code will goes as the PR author. It's a lot of @jamescrosswell code at this point so does it make sense for @jamescrosswell to take over this as he'll be the name popping on blame to do maitenance after anyway?

Replaced by #3699 - that'll get rid of the noise of almost 90 commits as well!

Issue #3524, opened 2024-08-06
Installation ID is constantly being reevaluated


klemmchr:
### Package

Sentry

### .NET Flavor

.NET

### .NET Version

8.0.0

### OS

Linux

### SDK Version

4.9.0

___

I'm observing repeated log errors in my ASP.NET Core app.

```
fail: Sentry.ISentryClient[0]
      => SpanId:1423812b5249bd8d, TraceId:11f7f8ef457099059955f79b83b98f6f, ParentId:0000000000000000 => ConnectionId:0HN5LM26FI76P => RequestPath:/healthz RequestId:0HN5LM26FI76P:00000001
      Failed to resolve persistent installation ID.
      System.UnauthorizedAccessException: Access to the path '/app/Sentry' is denied.
       ---> System.IO.IOException: Permission denied
         --- End of inner exception stack trace ---
         at System.IO.FileSystem.CreateParentsAndDirectory(String fullPath, UnixFileMode unixCreateMode)
         at System.IO.FileSystem.CreateDirectory(String fullPath, UnixFileMode unixCreateMode)
         at System.IO.Directory.CreateDirectory(String path)
         at Sentry.InstallationIdHelper.TryGetPersistentInstallationId()
```

Looking at the corresponding code I cannot explain this behavior.

https://github.com/getsentry/sentry-dotnet/blob/4617aeea07649fa6ef781b24a787c1c5bc09d84b/src/Sentry/InstallationIdHelper.cs#L6-L91

It seems that the machine ID is being recomputed every time a transaction is being send.

On a side note: failing to get the persistent installation ID should not output a fail log. When using non-root .NET Docker containers you typically have no write access to the app root directory since you're using a different user.

jamescrosswell:
> Looking at the corresponding code I cannot explain this behavior.

Hi @klemmchr , 

The only reason I can think why you'd see that repeatedly is:
1. Either if the log errors related to separate instances of your app running (either on a single device or multiple devices)
2. Or if Sentry was also unable to resolve a HardwareInstallationId and this code failed:
    https://github.com/getsentry/sentry-dotnet/blob/009241d1fb662ae6ad88d5bffc854ca613c2ec72/src/Sentry/InstallationIdHelper.cs#L122-L124

[...]

PR #3529, opened 2024-08-12
Fix: InstallationId now gets evaluated only once per application execution

jamescrosswell:
Resolves https://github.com/getsentry/sentry-dotnet/issues/3524


src/Sentry/Internal/InstallationIdHelper.cs
```diff
@@ -1,20 +1,12 @@
...
 {
     private readonly object _installationIdLock = new();
     private string? _installationId;
-    private readonly SentryOptions _options;
-    private readonly string? _persistenceDirectoryPath;
-
-    public InstallationIdHelper(SentryOptions options)
-    {
-        _options = options;
-        _persistenceDirectoryPath = options.CacheDirectoryPath ?? options.TryGetDsnSpecificCacheDirectoryPath();
```
bitsandfoxes, line 16: Fair enough. Looks like `TryGetDsnSpecificCacheDirectoryPath` returns `null` in case of `CacheDirectoryPath` being `null` anyway. If null then null. Sweet.

src/Sentry/Internal/InstallationIdHelper.cs
```diff
@@ -72,20 +64,20 @@ public InstallationIdHelper(SentryOptions options)
...
-            _options.LogDebug("Saved installation ID '{0}' to file '{1}'.", id, filePath);
+            options.LogDebug("Saved installation ID '{0}' to file '{1}'.", id, filePath);
             return id;
         }
         // If there's no write permission or the platform doesn't support this, we handle
         // and let the next installation id strategy kick in
         catch (Exception ex)
         {
-            _options.LogError(ex, "Failed to resolve persistent installation ID.");
+            options.LogWarning(ex, "Failed to resolve persistent installation ID.");
```
bitsandfoxes, line 80: What's the reasoning for the downgrade in log level? Failing to create a persistent installation ID sounds like a pretty major deal to me.
AFAIK, the whole sessions/release health part relies on it.

src/Sentry/SentryOptions.cs
```diff
@@ -1182,6 +1181,7 @@ public bool JsonPreserveReferences
     public SentryOptions()
     {
         SettingLocator = new SettingLocator(this);
+        _lazyInstallationId = new(() => new InstallationIdHelper(this).TryGetInstallationId());
```
bitsandfoxes, line 1184: And because it's `readonly` now it's ensured to be set only once in the future as well.

bitsandfoxes (commented):
I just wonder about the log level.

jamescrosswell:
> I just wonder about the log level.

Summary from [this conversation thread](https://github.com/getsentry/sentry-dotnet/issues/3524#issuecomment-2272236024) (see comments from Christian Klemm):
- When using non-root .NET Docker containers you typically have no write access to the app root directory since you're using a different user.
- I'm not sure whether an unprivileged user has access to network cards

We could revert to logging an error. At least now people would only get one error per application run.

src/Sentry/Internal/InstallationIdHelper.cs
jamescrosswell: Summary from [this conversation thread](https://github.com/getsentry/sentry-dotnet/issues/3524#issuecomment-2272236024) (see comments from Christian Klemm):
- When using non-root .NET Docker containers you typically have no write access to the app root directory since you're using a different user.
- I'm not sure whether an unprivileged user has access to network cards

We could revert to logging an error. At least now people would only get one error per application run.

src/Sentry/SentryOptions.cs
jamescrosswell: The main problem before was actually this:
```csharp
    private Lazy<string?> LazyInstallationId => new(() => new InstallationIdHelper(this).TryGetInstallationId());
```

It's subtle but that was actually declaring a method that creates a new Lazy (not a get only property), so every time you call it a new InstallationIdHelper gets created and nothing gets cached. 

The new code declares a backing field and assigns it to a (readonly) lazy instance. Since it's using non-static fields/properties in the initialization, it has to go in the constructor. But yeah, the readonly is nice reassurance.

src/Sentry/Internal/InstallationIdHelper.cs
bitsandfoxes: I think we should keep it at error. 
The SDK is attempting something that's a prerequesite for a feature and fails to set it up. That's an error. Non-root .NET Docker container users can opt-out of i.e. auto-session-tracking.

bitsandfoxes, line 80:
```suggestion
            options.LogError(ex, "Failed to resolve persistent installation ID.");
```

src/Sentry/Internal/InstallationIdHelper.cs
```diff
@@ -105,15 +97,15 @@ public InstallationIdHelper(SentryOptions options)
 
             if (string.IsNullOrWhiteSpace(installationId))
             {
-                _options.LogError("Failed to find an appropriate network interface for installation ID.");
+                options.LogWarning("Failed to find an appropriate network interface for installation ID.");
```
bitsandfoxes, line 100:
```suggestion
                options.LogError("Failed to find an appropriate network interface for installation ID.");
```

src/Sentry/Internal/InstallationIdHelper.cs
```diff
@@ -105,15 +97,15 @@ public InstallationIdHelper(SentryOptions options)
...
                 return null;
             }
 
             return installationId;
         }
         catch (Exception ex)
         {
-            _options.LogError(ex, "Failed to resolve hardware installation ID.");
+            options.LogWarning(ex, "Failed to resolve hardware installation ID.");
```
bitsandfoxes, line 108:
```suggestion
            options.LogError(ex, "Failed to resolve hardware installation ID.");
```

bitsandfoxes (approved):
Awesome! That detail in the `lazy`!

src/Sentry/Internal/InstallationIdHelper.cs
```diff
@@ -77,7 +77,7 @@ internal class InstallationIdHelper(SentryOptions options)
         // and let the next installation id strategy kick in
         catch (Exception ex)
         {
-            options.LogWarning(ex, "Failed to resolve persistent installation ID.");
+            options.LogError(ex, "Failed to resolve persistent installation ID.");
```
jamescrosswell, line 80: I do wonder if all three of these log statements need to be `LogError`. If we can't store a file with a generated id, it's not a huge problem as long as we can read an machine identifier from the network card.

It feels like this would be an error only if none of the 3 techniques we try to resolve an installation id are successful.

Issue #3540, opened 2024-08-16
[Android] FormatException when reading "processor_frequency"


TimBurik:
### Package

Sentry

### .NET Flavor

.NET

### .NET Version

8.0.303

### OS

Android

### SDK Version

4.10.0

### Steps to Reproduce

1. Create an Android app and install Sentry NuGet package
2. Configure Sentry so that `EnableBeforeSend` is enabled and `BeforeSend` callback is set
3. Crash the app using SentrySdk.CauseCrash(CrashType.JavaBackgroundThread)

### Expected Result

BeforeSend callback is fired, the `java.lang.RuntimeException: This exception was caused deliberately by SentrySdk.CauseCrash(CrashType.JavaBackgroundThread)` crash is reported.

### Actual Result

The following crash is reported:

```
System.FormatException: One of the identified items was in an invalid format.
  File "ThrowHelper.cs", line 594, in void ThrowHelper.ThrowFormatException()
    throw new FormatException { Source = ExceptionSourceValueToRethrowAsJsonException };
  File "JsonElement.cs", line 660, in int JsonElement.GetInt32()
    ThrowHelper.ThrowFormatException();
  File "Device.cs", line 459, in Device Device.FromJson(JsonElement json)
    var processorFrequency = json.GetPropertyOrNull("processor_frequency")?.GetInt32();
  File "SentryContexts.cs", line 148, in SentryContexts SentryContexts.FromJson(JsonElement json)
    result[name] = Device.FromJson(value);
  File "MiscExtensions.cs", line 5, in SentryContexts MiscExtensions.Pipe<JsonElement, SentryContexts>(JsonElement input, Func<JsonElement, SentryContexts> pipe)
    public static TOut Pipe<TIn, TOut>(this TIn input, Func<TIn, TOut> pipe) => pipe(input);
  File "SentryEvent.cs", line 305, in SentryEvent SentryEvent.FromJson(JsonElement json, Exception exception)
    var contexts = json.GetPropertyOrNull("contexts")?.Pipe(SentryContexts.FromJson);
  File "SentryEventExtensions.cs", line 29, in SentryEvent SentryEventExtensions.ToSentryEvent(SentryEvent sentryEvent, SentryOptions javaOptions)
    return SentryEvent.FromJson(json.RootElement, sentryEvent.Throwable);
  File "BeforeSendCallback.cs", line 26, in SentryEvent BeforeSendCallback.Execute(SentryEvent e, Hint h)
[...]

PR #3541, opened 2024-08-16
fix: `FormatException` when reading `ProcessorFrequency`

bitsandfoxes:
Fixes https://github.com/getsentry/sentry-dotnet/issues/3540

### DIFF
--- a/src/Sentry/Protocol/Device.cs
+++ b/src/Sentry/Protocol/Device.cs
@@ -431,7 +431,7 @@ public sealed class Device : ISentryJsonSerializable, ICloneable<Device>, IUpdat
         // The Java and Cocoa SDK report the battery as `float`
         // Cocoa https://github.com/getsentry/sentry-cocoa/blob/e773cad622b86735f1673368414009475e4119fd/Sources/Sentry/include/SentryUIDeviceWrapper.h#L18
         // Java  https://github.com/getsentry/sentry-java/blob/25f1ca4e1636a801c17c1662f0145f888550bce8/sentry/src/main/java/io/sentry/protocol/Device.java#L231-L233
-        var batteryLevel = json.GetPropertyOrNull("battery_level")?.TryGetDouble(out var level) == true
+        var batteryLevel = json.GetPropertyOrNull("battery_level")?.TryGetDouble(out var level) is true
             ? (short)level
             : (short?)null;
 
@@ -457,7 +457,7 @@ public sealed class Device : ISentryJsonSerializable, ICloneable<Device>, IUpdat
         // TODO: For next major: Remove this and change ProcessorFrequency from int to float
         // The Java SDK reports the processorFrequency as `double`
         // Java https://github.com/getsentry/sentry-java/blob/9762f09afa51944b40a9b77e116a55e54636e6c5/sentry/src/main/java/io/sentry/protocol/Device.java#L130
-        var processorFrequency = json.GetPropertyOrNull("processor_frequency")?.TryGetDouble(out var frequency) == true
+        var processorFrequency = json.GetPropertyOrNull("processor_frequency")?.TryGetDouble(out var frequency) is true
             ? (int)frequency
             : (int?)null;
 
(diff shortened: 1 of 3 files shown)


src/Sentry/Protocol/Device.cs
```diff
@@ -456,7 +456,17 @@ public static Device FromJson(JsonElement json)
         var bootTime = json.GetPropertyOrNull("boot_time")?.GetDateTimeOffset();
         var processorCount = json.GetPropertyOrNull("processor_count")?.GetInt32();
         var cpuDescription = json.GetPropertyOrNull("cpu_description")?.GetString();
-        var processorFrequency = json.GetPropertyOrNull("processor_frequency")?.GetInt32();
+
+        // TODO: For next major: Remove this and change ProcessorFrequency from int to float
+        // The Java SDK reports the processorFrequency as `double`
+        // Java https://github.com/getsentry/sentry-java/blob/9762f09afa51944b40a9b77e116a55e54636e6c5/sentry/src/main/java/io/sentry/protocol/Device.java#L130
+        int? processorFrequency = null;
+        var processorFrequencyProperty = json.GetPropertyOrNull("processor_frequency");
```
jamescrosswell, line 464: Should we change one of our tests as well (to make sure it works when deserialising a double)?

jamescrosswell (changes requested):
LGTM but would be good to update our unit tests to match as well.

src/Sentry/Protocol/Device.cs
bitsandfoxes: AFAIK, we don't have any tests for `Device` as it's not doing anything. But the type-mismatch is obvious from looking at the Java code.

bitsandfoxes: I don't want to argue against writing tests. I've added it to https://github.com/getsentry/sentry-dotnet/issues/3542

src/Sentry/Protocol/Device.cs
```diff
@@ -456,7 +456,17 @@ public static Device FromJson(JsonElement json)
...
+        if (processorFrequencyProperty.HasValue)
+        {
+            processorFrequency = (int)processorFrequencyProperty.Value.GetDouble();
```
bruno-garcia, line 467: Can we use an API to check the type? like try casting stuff instead?

src/Sentry/Protocol/Device.cs
```diff
@@ -456,7 +453,14 @@ public static Device FromJson(JsonElement json)
         var bootTime = json.GetPropertyOrNull("boot_time")?.GetDateTimeOffset();
         var processorCount = json.GetPropertyOrNull("processor_count")?.GetInt32();
         var cpuDescription = json.GetPropertyOrNull("cpu_description")?.GetString();
-        var processorFrequency = json.GetPropertyOrNull("processor_frequency")?.GetInt32();
+
+        // TODO: For next major: Remove this and change ProcessorFrequency from int to float
+        // The Java SDK reports the processorFrequency as `double`
+        // Java https://github.com/getsentry/sentry-java/blob/9762f09afa51944b40a9b77e116a55e54636e6c5/sentry/src/main/java/io/sentry/protocol/Device.java#L130
+        var processorFrequency = json.GetPropertyOrNull("processor_frequency")?.TryGetDouble(out var frequency) == true
```
bruno-garcia, line 460: Should we use pattern matching instead of `==` operator?
```suggestion
        var processorFrequency = json.GetPropertyOrNull("processor_frequency")?.TryGetDouble(out var frequency) is true
```

src/Sentry/Protocol/Device.cs
```diff
@@ -427,16 +427,13 @@ public static Device FromJson(JsonElement json)
...
         // The Java and Cocoa SDK report the battery as `float`
         // Cocoa https://github.com/getsentry/sentry-cocoa/blob/e773cad622b86735f1673368414009475e4119fd/Sources/Sentry/include/SentryUIDeviceWrapper.h#L18
         // Java  https://github.com/getsentry/sentry-java/blob/25f1ca4e1636a801c17c1662f0145f888550bce8/sentry/src/main/java/io/sentry/protocol/Device.java#L231-L233
-        short? batteryLevel = null;
-        var batteryProperty = json.GetPropertyOrNull("battery_level");
-        if (batteryProperty.HasValue)
-        {
-            batteryLevel = (short)batteryProperty.Value.GetDouble();
-        }
+        var batteryLevel = json.GetPropertyOrNull("battery_level")?.TryGetDouble(out var level) == true
```
bruno-garcia, line 434: same here `is` instead of `==`

bruno-garcia (approved):
Just missing some tests that cover cases int, double and something invalid like string.

jamescrosswell (approved):

Issue #2136, opened 2023-01-30
Feature: Session Replay support for .NET MAUI


bruno-garcia:
### Problem Statement

Add support for Sentry's Session Replay: https://sentry.io/for/session-replay/

### Solution Brainstorm

Since this is native, it'll likely be a wrapper over the solution built for:

- Android: [https://github.com/getsentry/sentry-java/issues/2476](https://github.com/getsentry/sentry-java/issues/2476)
- iOS: [https://github.com/getsentry/sentry-cocoa/issues/2616](https://github.com/getsentry/sentry-cocoa/issues/2616)

bruno-garcia:
Waitlist and discussion about the feature: https://github.com/getsentry/sentry/discussions/63138

a-martsineuski:
@bruno-garcia do we have any updates on that? I suppose it is possible to do a native binding for MAUI support, but not sure if you are working on that...

bruno-garcia:
@a-martsineuski we'd like to add this but it's not a WIP yet.
I got an experiment for Android here:
* #3552 

As you said, it's all about bindings.

If you could make a PR for iOS, that'd be super helpful. But our bindings right now are a bit messy given there's a script we run to adjust the output of Objective Sharpie, and the script isn't re-entrant. Any help here much appreciated.

bruno-garcia:
To clarify we don't have a date on this yet.
There was a PR attempting support but we didn't manage to add this yet.

Syed-RI:
Hi everyone... I see the the MAUI PR was closed with unmerged commits. Is there any planning to bring this to MAUI yet? Appreciate you guys might be busy with other core products

Flash0ver:
We currently do support Session Replay for _Android_ only.
Please see our MAUI example:
https://github.com/getsentry/sentry-dotnet/blob/aba82c2858ace9eb552e0d44f9b254e85f96358b/samples/Sentry.Samples.Maui/MauiProgram.cs#L47-L61

We are currently busy with upgrading to and testing compatibility with .NET 10, as well as considering some _breaking changes_ (to improve general usability of the Sentry SDK), to be ready to release `v6.0.0` alongside GA of the .NET 10 SDK.
We are planning to resume supporting Session Replay for iOS afterwards.

Syed-RI:
awesome thanks @Flash0ver 🚀

Flash0ver:
We now also got (basic) _experimental_ support for Session Replay on iOS (#4664).
[...]

PR #3552, opened 2024-08-23
Session Replay: Recording on Android

bruno-garcia:
image

Recording and redaction works fine on MAUI because it's all native under the hood. 
To complete this we need to get breadcrumbs from C#, trace/replay id to the C# layer and other features. 

### References

* [Developer Docs](https://develop.sentry.dev/sdk/telemetry/replays/)
* Implementation in other SDKs 
  * [React Native](https://github.com/getsentry/sentry-react-native/pull/3830)
  * [Dart/Flutter](https://github.com/getsentry/sentry-dart/tree/main/flutter/lib/src/replay)

Relates to:
* https://github.com/getsentry/sentry-dotnet/issues/2136


bruno-garcia:
I'd love to test this out on Symbol Collector. How about we make a `-alpha` with this, so I can add it there?

it'll run on hundreds of devices.

jamescrosswell:
> I'd love to test this out on Symbol Collector. How about we make a `-alpha` with this, so I can add it there?

You mean make a release from this branch? We could definitely do that... just mark it `-replayAndroid1` or something like that so it's clear on nuget.org that it's an experimental/prerelease.

samples/Sentry.Samples.Maui/MauiProgram.cs
```diff
@@ -22,8 +22,13 @@ public static MauiApp CreateMauiApp()
 
                 options.Debug = true;
                 options.SampleRate = 1.0F;
+#if ANDROID
+                options.Native.ExperimentalOptions.SessionReplay.OnErrorSampleRate = 1.0;
+                options.Native.ExperimentalOptions.SessionReplay.SessionSampleRate = 1.0;
+                options.Native.ExperimentalOptions.SessionReplay.RedactAllImages = false;
+                options.Native.ExperimentalOptions.SessionReplay.RedactAllText = false;
+#endif
```
vaind, line 30: In the future, we should probably have a facade that configures options for the current platform (Android/iOS), instead of letting the user specify them multiple times.

vaind (approved):
LGTM as an early alfa

samples/Sentry.Samples.Maui/MauiProgram.cs
jamescrosswell: > In the future, we should probably have a facade that configures options for the current platform (Android/iOS), instead of letting the user specify them multiple times.

If the options had the same name for both iOS and Android (`OnErrorSampleRate`, `SessionSampleRate` etc.) then presumably SDK users could write something like this, even without any facade:

```
#if ANDROID || IOS
                options.Native.ExperimentalOptions.SessionReplay.OnErrorSampleRate = 1.0;
                options.Native.ExperimentalOptions.SessionReplay.SessionSampleRate = 1.0;
                options.Native.ExperimentalOptions.SessionReplay.RedactAllImages = false;
                options.Native.ExperimentalOptions.SessionReplay.RedactAllText = false;
#endif
```

When targeting android, `options.Native` would resolve to the appropriate platform specific options at build no?

bruno-garcia:
cc @bricefriha something we'd like to push through once the open issues are under control

bruno-garcia:
would love to try this too on Symbol Collector. cc @jamescrosswell

jamescrosswell:
> would love to try this too on Symbol Collector. cc @jamescrosswell

Is there anything blocking that? Can't we just make a pre-release from this branch? Or is it more about having someone integrated it into the Symbol Collector?

bruno-garcia:
woo it's green! THanks @jamescrosswell 

can we merge this in? (it's an experimental feature so not too worried about lack of docs for example)

Issue #3560, opened 2024-08-25
Replace SubstringOrRegexPattern with an IStringOrRegexMatcher


jamescrosswell:
We have a better abstraction for this now:
https://github.com/getsentry/sentry-dotnet/blob/34b55980cd9f84d3922f80ff70c87b76d38ae01d/src/Sentry/SubstringOrRegexPattern.cs#L5-L10

bitsandfoxes:
Closed by #3566

PR #3566, opened 2024-08-26
Added SubstringOrPatternMatcher

jamescrosswell:
Resolves https://github.com/getsentry/sentry-dotnet/issues/3560

# Summary

Previously we had a `SubstringOrRegexPattern` class that could store either a string or a regular expression and contained some logic to test either whether that string was a substring or whether the regular expression matched against some other target string.

Instead we now have a `StringOrRegex` class that takes care of holding the string or regular expression, the ability to bind those values to configuration storage and provides various extensions.

The logic to test that string or regex against a target has been refactored into a `SubstringOrPatternMatcher` class implementing `IStringOrRegexMatcher`.


bitsandfoxes (commented):
We don't have anything on the samples using this?

jamescrosswell:
> We don't have anything on the samples using this?

It would be used indirectly via various properties on the `SentryOptions` (`TagFilters`, `InAppExclude`, `FailedRequestTargets` and `TracePropagationTargets`). We don't seem to demo how to use any of those in the Samples. 

It's also used in the `ExperimentalMetricsOptions`, which we do have some samples for - e.g .here:
https://github.com/getsentry/sentry-dotnet/blob/43e28adab2859d8defc0c67692e3e00b4b2060bc/samples/Sentry.Samples.OpenTelemetry.AspNetCore/Program.cs#L50-L59

But nothing showing how people can define their own string or regex to assign to any of those options.

Issue #3542, opened 2024-08-16
Fix and align types in `Device`


bitsandfoxes:
## BatteryLevel
https://github.com/getsentry/sentry-dotnet/blob/de216fa6506852621a65e0d49c01f24132db08b4/src/Sentry/Protocol/Device.cs#L430-L436
- [ ] Change Device.BatteryLevel from short to float
- [ ] Tests to avoid regressions (i.e. Java changes the underlying types)
- [ ] https://github.com/getsentry/sentry-dotnet/pull/3485

## ProcessorFrequency
https://github.com/getsentry/sentry-dotnet/blob/de216fa6506852621a65e0d49c01f24132db08b4/src/Sentry/Protocol/Device.cs#L457-L462
- [ ] Change Device.ProcessorFrequency from int to float
- [ ] Tests to avoid regressions (i.e. Java changes the underlying types)
- [ ] https://github.com/getsentry/sentry-dotnet/pull/3541

PR #3567, opened 2024-08-27
Align device types with Cocoa and Java SDKs

jamescrosswell:
Resolves https://github.com/getsentry/sentry-dotnet/issues/3542


bitsandfoxes (approved):
Thanks for that one!

Issue #829, opened 2021-02-16
DisabledTransaction or DisabledSpan if SDK is disabled, eg empty DSN


marandaneto:
Right now the .NET SDK is allocating on every performance API call even if the SDK is non-functional (aka disabled), ideally, we'd do the NoOp pattern as we do on Java.

This SDK already does the NoOp pattern (eg Disabled) for many things, not yet for the performance API.

https://github.com/getsentry/sentry-dotnet/blob/37eb4a54cee682bec0d63435ce9959a67fb4656e/src/Sentry/Extensibility/DisabledHub.cs#L47-L54

Originally conflicted with #747 however we now have `NoOpSpan` and `NoOpTransaction` classes available.

@brustolin we'd need on Cocoa too.
@untitaker Python also does not do it, was there a reason or just because it's not a common thing in the Python ecosystem?
@bruno-garcia wanted to confirm that before going ahead and doing the changes.

untitaker:
does not do what?

marandaneto:
> does not do what?

eg:
`capture_exception` checks if `Client` is `None`  and bails out
https://github.com/getsentry/sentry-python/blob/master/sentry_sdk/hub.py#L362-L363

`start_transaction` does not do it, and always create a new `Transaction` even if the SDK is non functional (eg disabled)
https://github.com/getsentry/sentry-python/blob/master/sentry_sdk/hub.py#L479-L531

the idea is to introduce the NoOp pattern, as some languages are not null-safety like Java, we cannot return `null` if the SDK is disabled.

we don't want people checking the return types for null values, its too much boilerplate, we also don't want to allocate and affect memory footprint even if the SDK is disabled.

untitaker:
Yup ok. Creating a noop transaction type is fine. Python does not do it because I didn't get around to it, but I think the transactions don't have a span recorder attached, at least.

>we don't want people checking the return types for null values

yeah that's what I wanted to solve

bruno-garcia:
Dupe: https://github.com/getsentry/sentry-dotnet/issues/747 ?

marandaneto:
> Dupe: #747 ?

maybe, but the reason behind #747 was another one, to still be able to propagate the traceId across server-side and client-side SDKs, and this one is specifically of being NoOp only if SDK is disabled, which is what we've discussed during our call.

PR #3574, opened 2024-08-28
DisabledHub.StartTransaction now returns a NoOpTransaction

jamescrosswell:
Resolves https://github.com/getsentry/sentry-dotnet/issues/829


bitsandfoxes (approved):
I mean, yes! But why do we need to put that on the 5.0 branch? Because it's functionally a breaking change? Even tho nothing gets sent either way?

jamescrosswell:
> But why do we need to put that on the 5.0 branch?

I can't see any good reason. I'll cherry pick it out into a new branch targeting main instead.

jamescrosswell:
Superceded by https://github.com/getsentry/sentry-dotnet/pull/3581

PR #3581, opened 2024-08-30
DisabledHub.StartTransaction now returns NoOpTransaction

### LINKED ISSUE #829: DisabledTransaction or DisabledSpan if SDK is disabled, eg empty DSN

jamescrosswell:
Resolves #829 

Replaces https://github.com/getsentry/sentry-dotnet/pull/3574 (cherry picked commits to target `main` branch instead of the `version-5.0.0` branch.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,9 @@
 ### Fixes
 - On mobile devices, the SDK no longer throws a `FormatException` for `ProcessorFrequency` when trying to report native events ([#3541](https://github.com/getsentry/sentry-dotnet/pull/3541))
 
+### API Changes
+- When the Sentry SDK is disabled, `SentrySdk.StartTransaction()` now returns a `NoOpTransaction`, which avoids unnecessary memory allocations ([#3581](https://github.com/getsentry/sentry-dotnet/pull/3581))
+
 ### Dependencies
 
 - Bump Cocoa SDK from v8.35.0 to v8.36.0 ([#3570](https://github.com/getsentry/sentry-dotnet/pull/3570), [#3575](https://github.com/getsentry/sentry-dotnet/pull/3575))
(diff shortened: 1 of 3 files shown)


bitsandfoxes (approved):

bitsandfoxes:
macOS is not happy today

jamescrosswell:
> macOS is not happy today

Yeah I'm not even seeing any errors in the logs. It runs `dotnet restore` and then just stops logging after a while.
- We can't turn on diagnostic logging either, since then we run out of disk space 😢 
- We can't even upgrade to a larger runner since the [larger macOs runners](https://docs.github.com/en/actions/using-github-hosted-runners/using-larger-runners/about-larger-runners#specifications-for-general-larger-runners) just come with more memory (not more disk space) 😭 

That leaves us in a bit of a pickle.

bitsandfoxes (approved):
LGTM!

Joshhua5:
Can we expand on this and use the NoOpTransaction based on a sampling rate at the start of a transaction and not the end? 
Sentry is close to half of all our memory allocations and reducing the sample rate isn't effective at reducing that.

bitsandfoxes:
> Can we expand on this and use the NoOpTransaction based on a sampling rate at the start of a transaction and not the end?

That's a very fair request that I put in https://github.com/getsentry/sentry-dotnet/issues/3636. We're moving towards something of a spans-only world but maybe we can sneak this in somewhere in between to make life easier?

Issue #3383, opened 2024-05-23
`Mechanism.Handled`


bitsandfoxes:
The `MainExceptionProcessor` handles setting the mechanism here: https://github.com/getsentry/sentry-dotnet/blob/c30052e99d981aec3d6ff1bda42c3395a2376fe7/src/Sentry/Internal/MainExceptionProcessor.cs#L177-L188

So this code ends up with `Handled = true`
```
try
{
    throw new Exception();
}
catch (Exception ex)
{
    // will be marked as handled 
    SentrySdk.CaptureException(ex);
}
```

And this code ends up with `Handled --` in the issue's `Highlights`
```
// will not be marked as anything
SentrySdk.CaptureException(new Exception());
```
![Screenshot 2024-05-23 at 13 32 53](https://github.com/getsentry/sentry-dotnet/assets/25725783/619b7781-ba73-4175-8d71-4b4233e95ac2)

Interestingly, captured messages are `Handled --` too.

Which, personally, I find confusing. If an exception is not `unhandled`, should we not consider it `handled` implicitly? I'm not talking technical correctness here but user expectations.

But I can still filter issues based on `error.handled:false` so maybe it doesn't matter?

bitsandfoxes:
We should considered other scenarios where the stacktrace ends up being `null`. Just to make sure we're not falsely flagging handled exceptions as `--`.

ericsampson:
it feels to me like the only entity who knows for sure if an exception sent to `CaptureException` is actually "handled" or not is the caller. So should this be controlled by an optional input parameter of that method?
(I don't know how that aligns with the cross-language use of this type of Sentry method)

jamescrosswell:
> it feels to me like the only entity who knows for sure if an exception sent to `CaptureException` is actually "handled" or not is the caller. So should this be controlled by an optional input parameter of that method?

It's possible for the caller to set handled to true/false by setting the "mechanism" for the exception before capturing it... which is how the Sentry middleware sets this in the ASP.NET Core integration:
https://github.com/getsentry/sentry-dotnet/blob/d1e5efcdf3af763ad49f11cd2426cc14a315a901/src/Sentry.AspNetCore/SentryMiddleware.cs#L219

It looks like the main exception processor would overwrite any value that was set explicitly via the mechanism though... which probably isn't a good idea. If someone has set this mechanism explicitly, we shouldn't be overwriting it right?

bitsandfoxes:
[...]

PR #3583, opened 2024-08-30
Fix CaptureException to respect the current value of the exception's Sentry:Handled property, if set by the caller.

ericsampson:
Fix for bug mentioned here:
https://github.com/getsentry/sentry-dotnet/issues/3383#issuecomment-2201331266

### DIFF
--- a/src/Sentry.AspNetCore/SentryMiddleware.cs
+++ b/src/Sentry.AspNetCore/SentryMiddleware.cs
@@ -217,9 +217,19 @@ internal class SentryMiddleware : IMiddleware
 
             void CaptureException(Exception e, SentryId evtId, string mechanism, string description)
             {
+                // Don't overwrite any values that the user explicitly set on the Exception:
                 var userHandledSet = e.Data.Contains(Mechanism.HandledKey);
                 var userHandledValue = e.Data[Mechanism.HandledKey] as bool?;
-                e.SetSentryMechanism(mechanism, description, handled: userHandledSet ? userHandledValue : false);
+                var handled = userHandledSet ? userHandledValue : false;
+                var userMechanismSet = e.Data.Contains(Mechanism.MechanismKey);
+                var userMechanismValue = e.Data[Mechanism.MechanismKey] as string;
+                // See Mechanism.Type setter re special handling of "" here:
+                var type = userMechanismSet ? userMechanismValue ?? "" : mechanism;
+                var userDescriptionSet = e.Data.Contains(Mechanism.DescriptionKey);
+                var userDescriptionValue = e.Data[Mechanism.DescriptionKey] as string;
+                var desc = userDescriptionSet ? userDescriptionValue : description;
+
+                e.SetSentryMechanism(type, desc, handled);
 
                 var evt = new SentryEvent(e, eventId: evtId);
 
--- a/src/Sentry.AspNetCore/SentryMiddleware.cs
+++ b/src/Sentry.AspNetCore/SentryMiddleware.cs
@@ -6,6 +6,7 @@ using Sentry.AspNetCore.Extensions;
 using Sentry.Ben.BlockingDetector;
 using Sentry.Extensibility;
 using Sentry.Internal;
+using Sentry.Protocol;
 using Sentry.Reflection;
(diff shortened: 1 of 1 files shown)


src/Sentry.AspNetCore/SentryMiddleware.cs
```diff
@@ -216,7 +217,9 @@ public async Task InvokeAsync(HttpContext context, RequestDelegate next)
 
             void CaptureException(Exception e, SentryId evtId, string mechanism, string description)
             {
-                e.SetSentryMechanism(mechanism, description, handled: false);
+                var userHandledSet = e.Data.Contains(Mechanism.HandledKey);
+                var userHandledValue = e.Data[Mechanism.HandledKey] as bool?;
+                e.SetSentryMechanism(mechanism, description, handled: userHandledSet ? userHandledValue : false);
```
jamescrosswell, line 222:
```suggestion
                if (!e.Data.Contains(Mechanism.HandledKey)
                {
                    e.SetSentryMechanism(mechanism, description, handled: false);
                }
```

jamescrosswell (commented):
Thanks heaps for the PR @ericsampson !

src/Sentry.AspNetCore/SentryMiddleware.cs
jamescrosswell, line 222: I'm thinking, since we're going to the effort of not overwriting Mechanism.Handled when this has been set by the user, maybe we should also avoid overwriting the Mechanism.Type and Mechanism.Description as well?

ericsampson: Good question, I'll take a peek.

ericsampson: @jamescrosswell take a look again thanks.

jamescrosswell:
@ericsampson I took a look at this again and the SentryMiddleware code that you've modified in this PR is explicitly designed to catch exceptions that users's have caught themselves... these exceptions have, either because they were never caught or because they were rethrown, ended up in Sentry's "unhandled exceptions" handler. 

There are a number of these, including the one you've modified: [SentryMiddleware](https://github.com/getsentry/sentry-dotnet/blob/56ade60da9b81d351a8ccd92b95eb11abcd87728/src/Sentry.AspNetCore/SentryMiddleware.cs#L204),  [CaptureLastError](https://github.com/getsentry/sentry-dotnet/blob/24c159fac119315e21df6138dc04bfe3ccd573c2/src/Sentry.AspNet/SentryHttpServerUtilityExtensions.cs#L23-L27), [SentryFunctionsWorkerMiddleware](https://github.com/getsentry/sentry-dotnet/blob/3e70c3126515c87b1851b6a9a7afd99ffbb9aeb7/src/Sentry.Azure.Functions.Worker/SentryFunctionsWorkerMiddleware.cs#L50-L53), etc.

I don't think the SDK user should override the mechanism for these exceptions. The mechanism is set explicitly to indicate that this was caught by the `"SentryMiddleware.UnhandledException"` and the description provides some extra context for this, `""This exception was captured by the Sentry ASP.NET Core middleware, and then re-thrown. The web server likely returned a 5xx error code as a result of this exception."`. Finally, in that context, it also doesn't make sense for handled to ever be true in this case (the fact that we're handling it in an unhandled exception handler means it's not handled). 

To your comment regarding the two different scenarios where users might want to override handled:

### Catch and throw
```csharp
catch ex
{
sentry.CaptureException(ex); // `handled` is default `false` if unspecified.
throw // we're in the weeds, abandon ship!
} 
```

If you capture the exception here, even though it's re-thrown, when it gets captured again in the unhandled exception handler it will be identified as a duplicate and ignored/discarded. So this is the only Exception that will be reported to Sentry and you can put whatever you like in the mechanism, the exception data etc.

### Catch 
```csharp
catch ex
{
  // a bunch of code here to correct the exceptional situation and return the system state to correctness
  sentry.CaptureException(ex, handled: true);
}
```

This makes sense to me. You can do this currently by setting handled on the Mechanism and since you don't re-throw the exception, it will never get bubbled up to the `SentryMiddleware` and whatever values you set on the mechanism won't be overridden by Sentry... so this should work today.

ericsampson:
Closing as this was a misunderstanding of the actual issue.

PR #3584, opened 2024-09-02
All exceptions are now added as breadcrumbs on future events

### LINKED ISSUE #3444: SentrySdk.CaptureException should add a breadcrumb for future exceptions
### Problem Statement

It appears that when you call SentrySdk.CaptureException, it does not add a breadrcrumb of itself, which means the next SentrySdk.CaptureException call does not show the previous exception in the dashboard. This seems like critical information to have by default, as its quite common for one exception to trigger another if things aren't handled well, and you will lose that contextual awareness that this exception was caused by a different exception happening.

### Solution Brainstorm

_No response_

jamescrosswell:
Resolves #3444

This is a bit redundant if tracing is enabled. In that case there is a notification at the top of the event informing you of any other exceptions(s) that have been encountered within the same trace:

image

### DIFF
--- a/src/Sentry/Internal/Hub.cs
+++ b/src/Sentry/Internal/Hub.cs
@@ -381,6 +381,42 @@ internal class Hub : IHub, IMetricHub, IDisposable
 
     public bool CaptureEnvelope(Envelope envelope) => CurrentClient.CaptureEnvelope(envelope);
 
+    private void AddBreadcrumbForException(SentryEvent evt, Scope scope)
+    {
+        try
+        {
+            if (!IsEnabled || evt.Exception is not { } exception)
+            {
+                return;
+            }
+
+            var exceptionMessage = exception.Message ?? "";
+            var formatted = evt.Message?.Formatted;
+
+            string breadcrumbMessage;
+            Dictionary<string, string>? data = null;
+            if (string.IsNullOrWhiteSpace(formatted))
+            {
+                breadcrumbMessage = exceptionMessage;
+            }
+            else
+            {
+                breadcrumbMessage = formatted;
+                // Exception.Message won't be used as Breadcrumb message
+                // Avoid losing it by adding as data:
+                data = new Dictionary<string, string>
+                {
+                    {"exception_message", exceptionMessage}
+                };
+            }
+            scope.AddBreadcrumb(breadcrumbMessage, "Exception", data: data, level: BreadcrumbLevel.Critical);
+        }
+        catch (Exception e)
+        {
+            _options.LogError(e, "Failure to store breadcrumb for exception event: {0}", evt.EventId);
+        }
+    }
+
     public SentryId CaptureEvent(SentryEvent evt, Action<Scope> configureScope)
         => CaptureEvent(evt, null, configureScope);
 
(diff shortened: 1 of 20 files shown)


src/Sentry/Internal/Hub.cs
```diff
@@ -394,7 +396,10 @@ public SentryId CaptureEvent(SentryEvent evt, SentryHint? hint, Action<Scope> co
             var clonedScope = CurrentScope.Clone();
             configureScope(clonedScope);
 
-            return CaptureEvent(evt, clonedScope, hint);
+            // Although we clone a temporary scope for the configureScope action, for the second scope
+            // argument (the breadcrumbScope) we pass in the current scope... this is because we want
+            // a breadcrumb to be left on the current scope for exception events
```
bitsandfoxes, line 401: I guess adding the breadcrumb here is not an option. They are both public and the other `CaptureEvent` is the last down the chain, right? 🥲

src/Sentry/Internal/Hub.cs
```diff
@@ -403,9 +408,10 @@ public SentryId CaptureEvent(SentryEvent evt, SentryHint? hint, Action<Scope> co
         }
     }
 
-    public bool CaptureEnvelope(Envelope envelope) => CurrentClient.CaptureEnvelope(envelope);
-
     public SentryId CaptureEvent(SentryEvent evt, Scope? scope = null, SentryHint? hint = null)
+        => CaptureEvent(evt, scope ?? CurrentScope, scope ?? CurrentScope, hint);
+
+    private SentryId CaptureEvent(SentryEvent evt, Scope eventScope, Scope breadcrumbScope, SentryHint? hint)
```
bitsandfoxes, line 414: Oh man, this is already giving me a headache, `eventScope` vs `breadcrumbScope`?

jamescrosswell: > Oh man, this is already giving me a headache, `eventScope` vs `breadcrumbScope`?

Yeah I didn't like that much either, but I couldn't see an easier way to do it. The only alternative is to duplicate the code that adds the breadcrumb and insert this in the multiple public overloads of this method.

bitsandfoxes: What about putting it into a static helper? All it needs to do is modify a scope that's getting passed in, right?
Rather than juggling two different types of scopes during capture.

jamescrosswell: Yeah, that looks a bit tidier 👍🏻

src/Sentry/Internal/Hub.cs
```diff
@@ -379,6 +379,44 @@ private void ApplyTraceContextToEvent(SentryEvent evt, SentryPropagationContext
...
+            else
+            {
+                breadcrumbMessage = formatted;
+                // Exception.Message won't be used as Breadcrumb message
+                // Avoid losing it by adding as data:
+                data = new Dictionary<string, string>
+                {
+                    {"exception_message", exceptionMessage}
+                };
+            }
```
bitsandfoxes, line 411: Just double-checking: If I've got my log-level set to `Warning` and the SDK creates an event out of a log warning:
- The minimum level is reached and the integration captures an event, skipping the breadcrumb creation
- From the logging integration we get here
- There was no exception - we bail early
- Warnings don't get added as breadcrumbs by neither the logging integration nor the core SDK?

jamescrosswell: Nice catch... have pushed a new commit to fix (the `SentryLogger` and `SentrySink` now **_do_** create breadcrumbs for events, if no exceptions are associated with those events). 

I think that's the correct behaviour since `CaptureEvent` is also used by `CaptureMessage` (and I assume we don't want to automatically add breadcrumbs for calls to `CaptureMessage`):
https://github.com/getsentry/sentry-dotnet/blob/f22c2d479028cc883bf20dc58828a312856bb2dc/src/Sentry/HubExtensions.cs#L221

jamescrosswell:
@bitsandfoxes I just noticed this in the NLog options:
https://github.com/getsentry/sentry-dotnet/blob/57feab36cabc09723fd33f50f7dccf27c44faec1/src/Sentry.NLog/SentryNLogOptions.cs#L47-L51

Which makes me wonder whether we've taken the correct approach with MEL and SeriLog... possibly we want to make this configurable rather than always on?

PR #3588, opened 2024-09-02
chore: Fix macOS runner

bitsandfoxes:
The verbosity does not need to be `diagnostic`.

#skip-changelog


jamescrosswell (approved):

jamescrosswell:
btw: if we run into this again in the future, here are some [options to free even more disk space](https://github.com/actions/runner-images/issues/10511#issuecomment-2329762800).

PR #3596, opened 2024-09-05
Scope.Transaction is now AsyncLocal

### LINKED ISSUE #3590: SentryScope.Transaction should always be AsyncLocal
_Originally posted by @jamescrosswell in [#3546](https://github.com/getsentry/sentry-dotnet/issues/3546#issuecomment-2325391037)_

## Problem

When `IsGlobalModeEnabled == true`, spans from the UI can get mixed up with transactions in background services, and vice versa. 

Originally Global Mode was added so that tags and context that get set in the UI also get applied to exceptions that get thrown in background threads. This makes sense for Events/Exceptions. However, unless they share an AsyncLocal context, it doesn't make sense for spans from one thread to end up under transactions that were created in another thread.

## Solution

The "Current" transaction should always be stored as an AsyncLocal, regardless of what kind of scope stack manager is configured for the application more generally. 

## Compatibility with OpenTelemetry

This would be the same as the behaviour for OpenTelemetry traces, since [Activity.Current](https://github.com/dotnet/runtime/blob/5535e31a712343a63f5d7d796cd874e563e5ac14/src/libraries/System.Diagnostics.DiagnosticSource/src/System/Diagnostics/Activity.Current.net46.cs#L10) is declared as `static AsyncLocal<Activity>`.

jamescrosswell:
Resolves #3590
- https://github.com/getsentry/sentry-dotnet/issues/3590

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@
 - `Device.BatteryLevel` and `Device.ProcessorFrequency` are now stored as floats rather than ints, to align with the Cocoa and Java SDKs ([#3567](https://github.com/getsentry/sentry-dotnet/pull/3567))
 - `SentryOptions.EnableTracing` has been removed. Instead, tracing should be enabled or disabled by setting the `SentryOptions.TracesSampleRate` or by using `SentryOptions.TracesSampler` to configure a sampling function ([#3569](https://github.com/getsentry/sentry-dotnet/pull/3569))
 - The `FailedRequestTargets`, `TagFilters` and `TracePropagationTargets` options have all been changed from `SubstringOrRegexPattern` to `IList<StringOrRegex>` ([#3566](https://github.com/getsentry/sentry-dotnet/pull/3566))
+- Scope.Transaction is now always stored as an AsyncLocal, even in Global Mode, to prevent auto-instrumented spans from the UI ending up parented to transactions from a background task (or vice versa) ([#3596](https://github.com/getsentry/sentry-dotnet/pull/3596))
+
 ## Unreleased
 
 ### Fixes
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@
 - `Device.BatteryLevel` and `Device.ProcessorFrequency` are now stored as floats rather than ints, to align with the Cocoa and Java SDKs ([#3567](https://github.com/getsentry/sentry-dotnet/pull/3567))
 - `SentryOptions.EnableTracing` has been removed. Instead, tracing should be enabled or disabled by setting the `SentryOptions.TracesSampleRate` or by using `SentryOptions.TracesSampler` to configure a sampling function ([#3569](https://github.com/getsentry/sentry-dotnet/pull/3569))
 - The `FailedRequestTargets`, `TagFilters` and `TracePropagationTargets` options have all been changed from `SubstringOrRegexPattern` to `IList<StringOrRegex>` ([#3566](https://github.com/getsentry/sentry-dotnet/pull/3566))
-- Scope.Transaction is now always stored as an AsyncLocal, even in Global Mode, to prevent auto-instrumented spans from the UI ending up parented to transactions from a background task (or vice versa) ([#3596](https://github.com/getsentry/sentry-dotnet/pull/3596))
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -10,6 +10,8 @@
 - `Device.BatteryLevel` and `Device.ProcessorFrequency` are now stored as floats rather than ints, to align with the Cocoa and Java SDKs ([#3567](https://github.com/getsentry/sentry-dotnet/pull/3567))
 - `SentryOptions.EnableTracing` has been removed. Instead, tracing should be enabled or disabled by setting the `SentryOptions.TracesSampleRate` or by using `SentryOptions.TracesSampler` to configure a sampling function ([#3569](https://github.com/getsentry/sentry-dotnet/pull/3569))
 - The `FailedRequestTargets`, `TagFilters` and `TracePropagationTargets` options have all been changed from `SubstringOrRegexPattern` to `IList<StringOrRegex>` ([#3566](https://github.com/getsentry/sentry-dotnet/pull/3566))
+- Scope.Transaction is now always stored as an AsyncLocal, even in Global Mode, to prevent auto-instrumented spans from the UI ending up parented to transactions from a background task (or vice versa) ([#3596](https://github.com/getsentry/sentry-dotnet/pull/3596))
```
bruno-garcia, line 13: how about:
```suggestion
- `Scope.Transaction` is now always stored as an `AsyncLocal` also in [Global Mode](https://docs.sentry.io/platforms/dotnet/configuration/options/#is-global-mode-enabled), to prevent auto-instrumented spans from the UI ending up parented to transactions from a background task (or vice versa). ([#3596](https://github.com/getsentry/sentry-dotnet/pull/3596))
```

src/Sentry/Scope.cs
```diff
@@ -180,15 +185,52 @@ public string? TransactionName
...
+    /// GlobalScopeStackContainer, anything you store on the scope will be applied to all events that get sent to Sentry
+    /// (no matter which thread they are sent from).
+    /// </para>
+    /// <para>
+    /// Transactions are an exception, however. We don't want spans from threads created on the UI thread to be added as
+    /// children of Transactions/Spans that get created on the background thread, or vice versa. As such,
+    /// Scope.Transaction is always stored as an AsyncLocal, regardless of the ScopeStackContainer implementation.
+    /// </para>
+    /// <para>
+    /// See https://github.com/getsentry/sentry-dotnet/issues/3590 for more information.
```
bruno-garcia, line 200: cc @getsentry/team-mobile 

Should we have this behavior on all mobile SDKs? I think global mode and transactions is suboptimal across the board, right?

src/Sentry/Scope.cs
```diff
@@ -738,6 +780,24 @@ public void AddAttachment(string filePath, AttachmentType type = AttachmentType.
...
+    /// We need this lock to prevent a potential race condition in <see cref="ResetTransaction"/>.
+    /// </summary>
+    private readonly ReaderWriterLockSlim _transactionLock = new();
+
+    internal void ResetTransaction(ITransactionTracer? expectedCurrentTransaction)
+    {
+        _transactionLock.EnterWriteLock();
+        try
+        {
+            if (_transaction.Value == expectedCurrentTransaction)
```
bruno-garcia, line 793: Instead of `==` (operator) can we go with `object.ReferenceEquals(_transaction.Value, expectedCurrentTransaction);`? (iirc that's the API)

bruno-garcia (approved):

src/Sentry/Scope.cs
jamescrosswell: Done... out of interest, why is this preferable? Is this in case someone messes with the equality operator on their `ITransactionTracer` implementation)?

src/Sentry/Scope.cs
jamescrosswell: Yeah I think Scope.Transaction should be AsyncLocal always/everywhere. I can't think of a scenario where it will get us into trouble (famous last words...)

src/Sentry/Scope.cs
bruno-garcia: yeah one is an operator that can be overridden, and the other simply has no doubt of what it is or how it works.

Issue #3607, opened 2024-09-13
Resolving `InstallationId` writes to disk by default


bitsandfoxes:
# Problem

Writing to disk on restricted platforms - i.e. Nintento Switch - causes the game to crash: https://github.com/getsentry/sentry-unity/issues/1804

# What's happening

We're fetching the `installationId` lazily 
https://github.com/getsentry/sentry-dotnet/blob/9ad2caa0c1e21296a90242c06e564d079a59020f/src/Sentry/SentryOptions.cs#L1184

and during event enrichment that `installationId` gets resolved
https://github.com/getsentry/sentry-dotnet/blob/9ad2caa0c1e21296a90242c06e564d079a59020f/src/Sentry/Internal/Enricher.cs#L83

and this causes the InstallationIdHelper to attempt to create a persistent `installationId` on disk
https://github.com/getsentry/sentry-dotnet/blob/9ad2caa0c1e21296a90242c06e564d079a59020f/src/Sentry/Internal/InstallationIdHelper.cs#L30-L33
https://github.com/getsentry/sentry-dotnet/blob/9ad2caa0c1e21296a90242c06e564d079a59020f/src/Sentry/Internal/InstallationIdHelper.cs#L48-L56

# Solutions

1. ~~Change the accessor for `InstallationId` from `internal` to `public`. That way the Unity SDK can overwrite the lazy initialization during options construction.~~ (Hacky)
https://github.com/getsentry/sentry-dotnet/blob/9ad2caa0c1e21296a90242c06e564d079a59020f/src/Sentry/SentryOptions.cs#L45
2. Add controls over writing to disk. That could be a new option. Or we reuse `CacheDirectoryPath`. With the `CacheDirectoryPath` set to `null` we skip all operations writing to disk. We've already got `FileSystem` in place. 
https://github.com/getsentry/sentry-dotnet/blob/9ad2caa0c1e21296a90242c06e564d079a59020f/src/Sentry/Internal/FileSystem.cs#L19
We could replace all instances of the SDK directly calling `Directory.Create` for `File.Write` going through `FileSystem` and make that adhere to the options.

bitsandfoxes:
Going with solution 2. It's been like that for some time so there's no need to hack something together. Let's introduce a new options i.e. `DisableFileWrite` and route all writing to disk through our `FileSystem`.
Todo: Make sure reading from disk actually works on those restricted platforms.

PR #3614, opened 2024-09-18
fix: Route file writing through `FileSystem`

bitsandfoxes:
Fixes https://github.com/getsentry/sentry-dotnet/issues/3607, https://github.com/getsentry/sentry-unity/issues/1804

## Problem

Writing to disk is restricted on platforms like the Switch, causing crashes. There's currently no way to properly opt-out of this. I.e. the SDK creates and tries to persist an ID to disk out-of-the-box.

## Solution

- [x] Adding a new flag to the options: `DisableFileWrite`.
- [x] We've already got `IFileSystem` in place. Making sure to make it is respecting that flag
- [x] Find all write instances, i.e. `File.Create` and `Directory.Create` and replace them with `IFileSystem`.
- [x] Tests

## TODO

- [ ] Add automatic detection/linting of any use of `create`, `move, and `delete` to prevent regressions


src/Sentry/GlobalSessionManager.cs
```diff
@@ -53,14 +53,20 @@ private void PersistSession(SessionUpdate update, DateTimeOffset? pauseTimestamp
 
         try
         {
-            Directory.CreateDirectory(_persistenceDirectoryPath);
+            _options.LogDebug("Creating persistence directory for session file at '{0}'.", _persistenceDirectoryPath);
 
-            _options.LogDebug("Created persistence directory for session file '{0}'.", _persistenceDirectoryPath);
+            if (_options.FileSystem.CreateDirectory(_persistenceDirectoryPath))
```
bruno-garcia, line 58: ~if it returns true it means it failed? seems counter intuitive to me.~

I see below that it does return `false` if it fails, so there's a bug here

```suggestion
            if (!_options.FileSystem.CreateDirectory(_persistenceDirectoryPath))
```

src/Sentry/GlobalSessionManager.cs
```diff
@@ -53,14 +53,20 @@ private void PersistSession(SessionUpdate update, DateTimeOffset? pauseTimestamp
...
+            {
+                _options.LogInfo("Failed to create persistent directory for session file.");
```
bruno-garcia, line 60:
```suggestion
                _options.LogError("Failed to create persistent directory for session file.");
```

src/Sentry/GlobalSessionManager.cs
```diff
@@ -53,14 +53,20 @@ private void PersistSession(SessionUpdate update, DateTimeOffset? pauseTimestamp
...
+                return;
+            }
+
+            _options.LogDebug("Created persistence directory for session file");
```
bruno-garcia, line 64: We could get rid of this one since we have the entry one, and one if it failed.

src/Sentry/Internal/FileSystem.cs
```diff
@@ -16,38 +21,96 @@ public IEnumerable<string> EnumerateFiles(string path, string searchPattern) =>
...
 #endif
+        return true;
     }
 
-    public void DeleteFile(string path) => File.Delete(path);
+    public bool DeleteFile(string path)
+    {
+        if (_options?.DisableFileWrite is false)
+        {
+            _options?.LogWarning("Skipping deleting file. Writing to file system has been explicitly disabled.");
```
bruno-garcia, line 80:
```suggestion
            _options?.LogDebug("Skipping deleting file. Writing to file system has been explicitly disabled through DisableFileWrite");
```

src/Sentry/Sentry.csproj
```diff
@@ -59,6 +59,14 @@
     </Compile>
   </ItemGroup>
 
+  <!--
+    We use TestableIO to be able to mock the FileSystem during tests
+    https://github.com/TestableIO/System.IO.Abstractions
+  -->
+  <ItemGroup>
+    <PackageReference Include="TestableIO.System.IO.Abstractions.TestingHelpers" Version="20.0.4" PrivateAssets="all" />
```
bitsandfoxes, line 67: See Notes in description. We're not surfacing this to the public API but use it exclusively internaly.
Helps us with keeping a clean cache during tests due to mocking a filesystem in memory.

src/Sentry/GlobalSessionManager.cs
```diff
@@ -53,14 +53,18 @@ private void PersistSession(SessionUpdate update, DateTimeOffset? pauseTimestamp
 
         try
         {
-            Directory.CreateDirectory(_persistenceDirectoryPath);
+            _options.LogDebug("Creating persistence directory for session file at '{0}'.", _persistenceDirectoryPath);
 
-            _options.LogDebug("Created persistence directory for session file '{0}'.", _persistenceDirectoryPath);
+            if (!_options.FileSystem.CreateDirectory(_persistenceDirectoryPath))
+            {
+                _options.LogError("Failed to create persistent directory for session file.");
```
bruno-garcia, line 60: do we return false if the option was set to false? if so we'll polite the log

bruno-garcia:
Just realize the strategy with returning bool has an issue. since the caller doesn't know if it failed, or if it's an option picked by the user and will log a bunch of error with the sdk debug logger

You might not be able to hide that all in the file writer abstraction, leaving the knowledge of the option to the callsites and hence skipping the operation altogether. 

Or change the interface to return some enum where it can be failure or disabled. Or return null boolean to be simple. and document that null means disable and true/false is the real output

bitsandfoxes:
I'd like to keep the `DisableFileWrite` check in one place if possible. I think it's going to be much easier to maintain instead of having to check for it in every place the SDK is attempting to write something to disk.
The "failure to write" should be handled regardless of reason by the caller. But I'm adding the `null` as return value to allow for a more granular handling, i.e. retry or somesuch.

src/Sentry/Internal/FileSystem.cs
```diff
@@ -16,38 +19,112 @@ public IEnumerable<string> EnumerateFiles(string path, string searchPattern) =>
...
         Directory.EnumerateFiles(path, searchPattern, searchOption);
 
-    public void CreateDirectory(string path) => Directory.CreateDirectory(path);
+    public DirectoryInfo? CreateDirectory(string path)
+    {
+        if (!_options?.DisableFileWrite is false)
+        {
+            _options?.LogDebug("Skipping creating directory. Writing to file system has been explicitly disabled.");
+            return null;
+        }
```
jamescrosswell, line 28: This works but it does make the `FileSystem` class a bit harder to read and also means we're frequently checking the value of `DisableFileWrite` (even though this never changes, once initialised). 

An alternative would be to have two different implementations of `IFileSystem`:
1. SentryReadWriteFileSystem
2. SentryReadOnlyFileSystem

`SentryOptions.DisableFileWrite` could be inspected once and used to determine which concrete type to use when initialising when initialising `SentryOptions.FileSystem`.

That's a code style choice though - up to you.

src/Sentry/Internal/FileSystem.cs
```diff
@@ -16,38 +19,112 @@ public IEnumerable<string> EnumerateFiles(string path, string searchPattern) =>
     public IEnumerable<string> EnumerateFiles(string path, string searchPattern, SearchOption searchOption) =>
         Directory.EnumerateFiles(path, searchPattern, searchOption);
 
-    public void CreateDirectory(string path) => Directory.CreateDirectory(path);
+    public DirectoryInfo? CreateDirectory(string path)
+    {
+        if (!_options?.DisableFileWrite is false)
```
jamescrosswell, line 24:
```suggestion
        if (_options?.DisableFileWrite is true)
```

src/Sentry/Internal/FileSystem.cs
```diff
@@ -16,38 +19,112 @@ public IEnumerable<string> EnumerateFiles(string path, string searchPattern) =>
...
+            _options?.LogDebug("Skipping creating directory. Writing to file system has been explicitly disabled.");
+            return null;
+        }
+
+        return Directory.CreateDirectory(path);
+    }
+
+    public bool? DeleteDirectory(string path, bool recursive = false)
+    {
+        if (!_options?.DisableFileWrite is false)
```
jamescrosswell, line 35:
```suggestion
        if (_options?.DisableFileWrite is true)
```

src/Sentry/Internal/FileSystem.cs
```diff
@@ -16,38 +19,112 @@ public IEnumerable<string> EnumerateFiles(string path, string searchPattern) =>
...
+    }
 
     public bool DirectoryExists(string path) => Directory.Exists(path);
 
     public bool FileExists(string path) => File.Exists(path);
 
-    public void MoveFile(string sourceFileName, string destFileName, bool overwrite = false)
+    public bool? MoveFile(string sourceFileName, string destFileName, bool overwrite = false)
     {
+        if (!_options?.DisableFileWrite is false)
```
jamescrosswell, line 51:
```suggestion
        if (_options?.DisableFileWrite is true)
```

src/Sentry/Internal/FileSystem.cs
```diff
@@ -16,38 +19,112 @@ public IEnumerable<string> EnumerateFiles(string path, string searchPattern) =>
...
+            return false;
+        }
+
+        return true;
     }
 
-    public void DeleteFile(string path) => File.Delete(path);
+    public bool? DeleteFile(string path)
+    {
+        if (!_options?.DisableFileWrite is false)
```
jamescrosswell, line 81:
```suggestion
        if (_options?.DisableFileWrite is true)
```

src/Sentry/Internal/FileSystem.cs
```diff
@@ -16,38 +19,112 @@ public IEnumerable<string> EnumerateFiles(string path, string searchPattern) =>
...
+        {
+            _options?.LogDebug("Skipping creating directory. Writing to file system has been explicitly disabled.");
+            return null;
+        }
+
+        return Directory.CreateDirectory(path);
+    }
+
+    public bool? DeleteDirectory(string path, bool recursive = false)
```
jamescrosswell, line 33: I don't think this method can ever return null... so could be `bool` return type.

src/Sentry/Internal/FileSystem.cs
```diff
@@ -16,38 +19,112 @@ public IEnumerable<string> EnumerateFiles(string path, string searchPattern) =>
...
 
     public bool FileExists(string path) => File.Exists(path);
 
-    public void MoveFile(string sourceFileName, string destFileName, bool overwrite = false)
+    public bool? MoveFile(string sourceFileName, string destFileName, bool overwrite = false)
     {
+        if (!_options?.DisableFileWrite is false)
+        {
+            _options?.LogDebug("Skipping moving file. Writing to file system has been explicitly disabled.");
+            return null;
```
jamescrosswell, line 54: This is inconsistent with the behaviour in `DeleteDirectory`... In that method, if `DisableFileWrite` is enabled then a value of `false` is returned. Whereas, this method returns `null` when `DisableFileWrite` is enabled. 

Whatever we're deciding null and false represent, I think we should be consistent. 

Given that, possibly worth changing the return value to be an enum (Succeeded, Failed, Disabled) so that there is no ambiguity over what happened or why.

src/Sentry/Internal/ReadWriteFileSystem.cs
jamescrosswell, line 1: I think this is heaps easier to read although now there is quite a bit of duplication between `ReadOnlyFilesystem` (should that be `ReadOnlyFileSystem` with a capital S??) and `ReadWriteFileSystem`... all of the common (read) methods could be factored out into an abstract `BaseFileSystem` class.

src/Sentry/Internal/IFileSystem.cs
```diff
@@ -8,14 +15,21 @@ internal interface IFileSystem
...
+    string? ReadAllTextFromFile(string file);
     Stream OpenFileForReading(string path);
-    Stream CreateFileForWriting(string path);
+    Stream OpenFileForReading(string path,
+        bool useAsync,
+        FileMode fileMode = FileMode.Open,
+        FileAccess fileAccess = FileAccess.Read,
+        FileShare fileShare = FileShare.ReadWrite,
+        int bufferSize = 4096);
+    (FileOperationResult, Stream) CreateFileForWriting(string path);
```
jamescrosswell, line 33: I wonder if it would be better to return the Stream as an out parameter here and have this method return a `FileOperationResult`.

test/Sentry.Testing/FakeFileSystem.cs
```diff
@@ -1,54 +1,95 @@
-using MockFileSystem = System.IO.Abstractions.TestingHelpers.MockFileSystem;
+using System.IO.Abstractions.TestingHelpers;
 
-namespace Sentry.Testing;
+namespace Sentry.Internal;
```
jamescrosswell, line 3: Shouldn't the namespace match the assembly and folder? If I see a class in the Sentry.Internal namespace, my assumption is that it lives in the Sentry package...

test/Sentry.Tests/ReadWriteFileSystemTests.cs
jamescrosswell, line 1: TODO?

src/Sentry/Http/HttpTransportBase.cs
```diff
@@ -386,9 +386,18 @@ private void HandleFailure(HttpResponseMessage response, Envelope envelope)
                 var destination = Path.Combine(destinationDirectory, "envelope_too_large",
                     (eventId ?? SentryId.Create()).ToString());
 
-                Directory.CreateDirectory(Path.GetDirectoryName(destination)!);
+                if (_options.FileSystem.CreateDirectory(Path.GetDirectoryName(destination)!) is not FileOperationResult.Success)
+                {
+                    _options.LogError("Failed to create directory to store the envelope.");
+                    return;
+                }
```
jamescrosswell, line 393:
```suggestion
                var createDirectoryResult = _options.FileSystem.CreateDirectory(Path.GetDirectoryName(destination)!);
                if (createDirectoryResult != FileOperationResult.Success)
                {
                    _options.LogError($"Failed to create directory to store the envelope: {createDirectoryResult}");
                    return;
                }
```

Just so you can easily tell in the logs whether it failed or was disabled.

src/Sentry/Http/HttpTransportBase.cs
```diff
@@ -386,9 +386,18 @@ private void HandleFailure(HttpResponseMessage response, Envelope envelope)
...
 
-                var envelopeFile = File.Create(destination);
+                var (result, envelopeFile) = _options.FileSystem.CreateFileForWriting(destination);
+                if (result is not FileOperationResult.Success)
+                {
+                    _options.LogError("Failed to create envelope file.");
```
jamescrosswell, line 398: Same again here... be good to include the result in the log

src/Sentry/Http/HttpTransportBase.cs
```diff
@@ -442,9 +451,19 @@ private async Task HandleFailureAsync(HttpResponseMessage response, Envelope env
...
+                if (_options.FileSystem.CreateDirectory(Path.GetDirectoryName(destination)!) is not FileOperationResult.Success)
+                {
+                    _options.LogError("Failed to create directory to store the envelope.");
+                    return;
+                }
+
+                var (result, envelopeFile) = _options.FileSystem.CreateFileForWriting(destination);
+                if (result is not FileOperationResult.Success)
+                {
+                    _options.LogError("Failed to create envelope file.");
```
jamescrosswell, line 463: And here...

src/Sentry/Internal/ReadWriteFileSystem.cs
bitsandfoxes: Agreed. I was just playing around and the split into `ReadWrite` and `ReadOnly` really helped getting rid of the circular dependency between the `FileSystem` and the `options`.

jamescrosswell (approved):
LGTM

src/Sentry/ISentryJsonSerializable.cs
```diff
@@ -19,12 +20,18 @@ public interface ISentryJsonSerializable
...
+        if (result is not FileOperationResult.Success)
+        {
+            return;
+        }
+
         using var writer = new Utf8JsonWriter(file);
 
         serializable.WriteTo(writer, logger);
         writer.Flush();
+        file.Dispose();
```
bruno-garcia, line 35: this couldn’t have `using` anymore? I’m concerned an exception will keep dangling filesa

Issue #3597, opened 2024-09-05
Remove Metrics API


bitsandfoxes:
More about the what and why: https://sentry.zendesk.com/hc/en-us/articles/26369339769883-Metrics-Beta-Coming-to-an-End

Metrics have been marked as `obsolete` in https://github.com/getsentry/sentry-dotnet/pull/3619

jamescrosswell:
Resolved by:
- https://github.com/getsentry/sentry-dotnet/pull/3718

PR #3619, opened 2024-09-19
fix: Marked `SentrySdk.Metrics` as obsolete

bitsandfoxes:
Relates to Metrics Deprecation https://github.com/getsentry/sentry-dotnet/issues/3597

While marking the API as obsolete I also removed the shutdown flushing. I think it's fairly low risk and it fixes the deadlocking situation in https://github.com/getsentry/sentry-unity/issues/1779

### DIFF
--- a/src/Sentry/Internal/Hub.cs
+++ b/src/Sentry/Internal/Hub.cs
@@ -659,9 +659,7 @@ internal class Hub : IHub, IMetricHub, IDisposable
 
         try
         {
-            Metrics.FlushAsync().ContinueWith(_ =>
-                CurrentClient.FlushAsync(_options.ShutdownTimeout).ConfigureAwait(false).GetAwaiter().GetResult()
-            ).ConfigureAwait(false).GetAwaiter().GetResult();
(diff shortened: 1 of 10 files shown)


src/Sentry/Internal/Hub.cs
```diff
@@ -659,9 +659,7 @@ public void Dispose()
 
         try
         {
-            Metrics.FlushAsync().ContinueWith(_ =>
-                CurrentClient.FlushAsync(_options.ShutdownTimeout).ConfigureAwait(false).GetAwaiter().GetResult()
-            ).ConfigureAwait(false).GetAwaiter().GetResult();
+            CurrentClient.FlushAsync(_options.ShutdownTimeout).ConfigureAwait(false).GetAwaiter().GetResult();
```
jamescrosswell, line 662: Are we marking it as obsolete or removing the functionality in this PR? This change is removing metrics functionality right?

jamescrosswell: Ah, I see in the PR description:
> While marking the API as obsolete I also removed the shutdown flushing. I think it's fairly low risk and it fixes the deadlocking situation in https://github.com/getsentry/sentry-unity/issues/1779

jamescrosswell (approved):

PR #3628, opened 2024-09-25
Replace MauiTestUtils with DeviceRunners nuget package

jamescrosswell:
The device tests use a device runner Test App, which is a very simple MAUI application containing project references to all of our test projects. XHarness can be used to install and launch the test app on iOS and Android physical devices or simulators, run any xUnit tests referenced and then capture both test results and application logs as artefacts.

The mechanism used to discover and run xUnit tests within the test app is called Device Runners. Originally Microsoft made the Device Runners available as some sample code in the`dotnet/maui` repository, which we vendored into (i.e. copied into) the `test/MauiTestUtils` folder of the `sentry/sentry-dotnet` repo. 

Since then, that code appears to have been turned into a semi-official [Device Runners NuGet package](https://github.com/mattleibow/DeviceRunners), which is being actively maintained.

This PR removes our vendored in code - using a package reference to the most recent version of the NuGet package instead. 

Hopefully this will make it easier for us to upgrade the device tests to net8.0 in a separate PR.

## Notes to Reviewers

It looks like a lot to review but mostly it's deleting code:
- If you collapse the `tests/MauiTestUtils` directory, all of which is being deleted, it starts to look less overwhelming
- The `test/Sentry.Maui.Device.TestApp/Platforms/Android/SentryInstrumentation.cs` class is worth calling out... it's a custom instrumentation class that enables us to intercept launching of the device tests on Android (so we can inject environment variables that have been set on CI - mainly so we can skip certain flaky tests in CI)
- Most of the other changes are addressing flaky tests

## Notes on Flaky Tests

There are a few tests that I had to disable on iOS and some others I had to disable on Android. On Android, it was only when targeting API 31 or 33 (the only two I tried other than 27). When this happens, the error message from Xharness is a bit misleading - it indicates the tests are timing out after 900 seconds.

This may be related to [Restrictions on non-SDK interfaces](https://developer.android.com/guide/app-compatibility/restrictions-non-sdk-interfaces) but, if so, I don't understand how.

You can [download a CSV](https://dl.google.com/developers/android/vic/non-sdk/hiddenapi-flags.csv) describing ~656k API interfaces and the availability of each... so it's possible some of the APIs we're using in our device tests are not supported on API 33. How to know whether we're using any of the blocked/obsolete APIs though??? Maybe that information is in the adb logs if you know how to look for it but our adb logs are about 33k lines so that's a bit like trying to finding a needle in a haystack. 

If we could work out what was going on there, we might be able to re-enable the Android tests. I don't have any leads for  the iOS tests. 

#skip-changelog


test/Sentry.Extensions.Logging.Tests/LoggingTests.cs
```diff
@@ -14,6 +14,11 @@ public class LoggingTests
     [InlineData(LogLevel.Trace)]
     public void Log_CapturesEvent(LogLevel logLevel)
     {
+#if __IOS__
+        Skip.If(true, "Flaky on iOS");
+#endif
+        Skip.If(TestEnvironment.IsGitHubActions, "Flaky on CI");
```
bruno-garcia, line 20: This is flaky in CI and on iOS? should we get rid of this test?

test/Sentry.Extensions.Logging.Tests/LoggingTests.cs
```diff
@@ -55,6 +60,10 @@ public void Log_CapturesEvent(LogLevel logLevel)
     [InlineData(LogLevel.Trace)]
     public void Log_AddsBreadcrumb(LogLevel logLevel)
     {
+#if __IOS__
+        Skip.If(true, "Flaky on iOS");
```
bruno-garcia, line 64: Any ticket or pointer we can link from here? Curious if we should try to fix it or what not

test/Sentry.Maui.Device.TestApp/Platforms/Android/SentryInstrumentation.cs
```diff
@@ -0,0 +1,46 @@
...
+                Environment.SetEnvironmentVariable("GITHUB_ACTIONS", "true");
+            }
+            else
+            {
+                Console.WriteLine("CI build not detected - no environment variables set");
+            }
+        }
+        catch (Exception e)
+        {
+            Console.WriteLine(e);
```
bruno-garcia, line 39: Failure here would only be calling `IsGitHubActions` or `SetEnvironmentVariable` ?
Is this common? since we supresss the exception and continue, I wonder as it'll happen without us noticing unless we read the right part of the build output

test/Sentry.Tests/SentrySdkTests.cs
```diff
@@ -424,9 +424,13 @@ public void ConfigureScope_Sync_CallbackNeverInvoked()
         Assert.False(invoked);
     }
 
-    [Fact]
+    [SkippableFact]
     public async Task ConfigureScope_OnTask_PropagatedToCaller()
     {
+#if __ANDROID__
+        Skip.If(true, "Flaky on Android");
```
bruno-garcia, line 431: Android-only flaky tests too? 😬

bruno-garcia (approved):
a bit concerned with the Skip calls. otherwise LGTM

test/Sentry.Extensions.Logging.Tests/LoggingTests.cs
jamescrosswell: Yeah the stuff I surrounded in directives is flaky on my local dev machine as well as CI for that target platform. So in this case it's always flaky on iOS, but only flaky in CI on Android (runs reliably on Android on my local dev machine)... worth pointing out that they're only flaky on Android targeting API levels 28+ (reliable when targeting API level 27). 

There are some comments in the PR description about a potential way to resolve on Android, so it might be worth leaving the tests in here (disabled on CI), in the hope that we could circle back to this.

test/Sentry.Tests/SentrySdkTests.cs
jamescrosswell: Yep, it's the whole mixed bag of unreliable don't really trust these device tests. To be fair, that's not just this PR/branch - our Device Tests have been unreliable pretty much ever since I joined Sentry. Initially we were running them against 6 different API levels on Android and we disabled API levels 28, 29, 30, 32 because they were so unreliable. 

I think step one is to disable the flaky tests (we just ignore those failures and rerun the tests - so they're kind of meaningless) until we can come back and make them proper tests we trust again.

test/Sentry.Extensions.Logging.Tests/LoggingTests.cs
jamescrosswell: No ticket... the tests in this class just failed consistently on iOS. All works fine on other platforms. 

I do think we should circle back and try to find the root causes of these failures eventually. I've created this ticket so that doesn't get lost or forgotten about:
- https://github.com/getsentry/sentry-dotnet/issues/3649

test/Sentry.Maui.Device.TestApp/Platforms/Android/SentryInstrumentation.cs
jamescrosswell: We could probably remove this now 👍🏻 

It was always throwing an exception until I added this:
https://github.com/getsentry/sentry-dotnet/blob/4780038a8bf422d0190c0c1e75f9c9dcf38c02e0/test/Sentry.Maui.Device.TestApp/Sentry.Maui.Device.TestApp.csproj#L19-L20

test/Sentry.Extensions.Logging.Tests/LoggingTests.cs
bruno-garcia: > There are some comments in the PR description about a potential way to resolve on Android, so it might be worth leaving the tests in here (disabled on CI), in the hope that we could circle back to this.

Worth adding the notes on a ticekt since whatever is left here will be forgotten (unless a ticket links back to it)

bruno-garcia:
Feel free to merge once u're comfortable with maintaining this, from my PoV looks good

test/Sentry.Extensions.Logging.Tests/LoggingTests.cs
jamescrosswell: Done in https://github.com/getsentry/sentry-dotnet/issues/3649

PR #3631, opened 2024-09-26
Fixed duplicate key exception for Hangfire jobs with AutomaticRetry

### LINKED ISSUE #3626: Hangfire job failures result in SentryCheckInIdKey error
### Package

Other

### .NET Flavor

.NET

### .NET Version

8.0.8

### OS

Windows

### SDK Version

4.11.0

24.9.0

### Steps to Reproduce

1. Create a hangfire background job. In my case it's hourly with 5 retries.
2. Make said job fail repeatedly

Relevant function:
```
    [Hangfire.Queue("redacted")]
    [Hangfire.AutomaticRetry(Attempts = 5)]
    [Hangfire.DisableConcurrentExecution(timeoutInSeconds: 2 * 60)]
    [SentryMonitorSlug("redacted-auth-refresh")]
    public async Task DoAuthRefreshAsync(CancellationToken token)
    {
        var result = await this.apiClient.RefreshAuthAsync();

        if (result.IsError)
        {
            throw new Exception("TASK - Redacted: AuthRefresh: Failed to refresh auth. " + result.FirstError.Code + " - " + result.FirstError.Description);
        }
        else
        {
            Log.Information("TASK - Redacted: AuthRefresh: Redacted auth token checked and refreshed if needed.");
        }
    }
```

### Expected Result

A hangfire issue for the actual exception that was thrown. (Which actually happens the first time an exception occurs)

### Actual Result

The screenshot of the latest report in Sentry.
Image

Since I'm unable to share the actual sentry URL, This is the JSON for one of the many instances of this error.
https://gist.github.com/vrelk-net/e6175fda5ef90839e1a149769b2db14a

When the first event comes in, the report is actually helpful. After it continues, the original issue gets buried.
These breadcrumbs come from one of the first few events to come in.
Image

jamescrosswell:
Resolves #3626

@bitsandfoxes I'm not super familiar with the Hangfire integration. Is this sane? My guess is that `OnPerforming` is being called multiple times with the same context because @vrelk-net decorated the job with `[Hangfire.AutomaticRetry(Attempts = 5)]`... so all OK on the first attempt but on subsequent attempts a `SentryCheckInIdKey` is already on the context (from the first attempt).


bitsandfoxes (approved):
> Is this sane?

This looks correct to me!

PR #3641, opened 2024-09-30
fix: Simplified `IFileSystem`

bitsandfoxes:
## TLDR;

The responsibility to check whether a system is allowed to write to disk or not falls onto the system itself. The `ReadOnlyFileSystem` is only there to prevent regressions and crashes on platforms like the Switch.

Any attempts to write to disk and fail can now simply logged as `Error` because that's what they are.

## Context

The idea was to have `IFileSystem` be the abstraction layer and be swapped between `ReadOnly` and `ReadWrite` and have it be clever enough to log whether the operation failed or was disallowed.

Having it smart enough is not really feasable. The caller is still interested whether it failed or was disallowed for logging purposes. Having the caller simply attempt to write to disk and then not care whether it was allowed to do so is not possible.

So we end up checking the success/failure/disallowed with every call. That is noisy. So we're checking whether we're allowed to write to disk before even attempting. This making the `FileOperationResult` pointless. We can just log and skip when we know we can't write to disk in the first place.

### DIFF
--- a/src/Sentry/GlobalSessionManager.cs
+++ b/src/Sentry/GlobalSessionManager.cs
@@ -51,29 +51,42 @@ internal class GlobalSessionManager : ISessionManager
             return;
         }
 
+        if (_options.DisableFileWrite)
+        {
+            _options.LogInfo("File write has been disabled via the options. Skipping persisting session.");
+            return;
+        }
+
         try
         {
             _options.LogDebug("Creating persistence directory for session file at '{0}'.", _persistenceDirectoryPath);
 
-            var result = _options.FileSystem.CreateDirectory(_persistenceDirectoryPath);
-            if (result is not FileOperationResult.Success)
+            if (_options.FileSystem.CreateDirectory(_persistenceDirectoryPath) is not true)
             {
-                if (result is FileOperationResult.Disabled)
-                {
-                    _options.LogInfo("Persistent directory for session file has not been created. File-write has been disabled via the options.");
-                }
-                else
-                {
-                    _options.LogError("Failed to create persistent directory for session file.");
-                }
-
+                _options.LogError("Failed to create persistent directory for session file.");
                 return;
             }
 
             var filePath = Path.Combine(_persistenceDirectoryPath, PersistedSessionFileName);
 
             var persistedSessionUpdate = new PersistedSessionUpdate(update, pauseTimestamp);
-            persistedSessionUpdate.WriteToFile(_options.FileSystem, filePath, _options.DiagnosticLogger);
+            if (_options.FileSystem.CreateFileForWriting(filePath, out var file) is not true)
+            {
+                _options.LogError("Failed to persist session file.");
+                return;
+            }
+
+            using var writer = new Utf8JsonWriter(file);
+
+            try
+            {
+                persistedSessionUpdate.WriteTo(writer, _options.DiagnosticLogger);
+                writer.Flush();
+            }
+            finally
+            {
+                file.Dispose();
+            }
 
             _options.LogDebug("Persisted session to a file '{0}'.", filePath);
         }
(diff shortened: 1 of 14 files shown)


src/Sentry/ISentryJsonSerializable.cs
```diff
@@ -17,21 +17,3 @@ public interface ISentryJsonSerializable
...
-            return;
-        }
-
-        using var writer = new Utf8JsonWriter(file);
-
-        serializable.WriteTo(writer, logger);
-        writer.Flush();
-        file.Dispose();
-    }
-}
```
bitsandfoxes, line 37: This is only used once in the `GlobalSessionsManager`. Success/Failure is getting handled there and does not need to be abstracted away into an extension.

bruno-garcia:
Follow up to:
* https://github.com/getsentry/sentry-dotnet/pull/3614

src/Sentry/GlobalSessionManager.cs
```diff
@@ -91,6 +104,12 @@ private void DeletePersistedSession()
             return;
         }
 
+        if (_options.DisableFileWrite)
+        {
+            _options.LogInfo("File write has been disabled via the options. Skipping deleting persisted session.");
```
bruno-garcia, line 109: "skipping deletion if session files"

when we're reading the debug level log we're not sure how individual messages relate so could be helpful to be explicit

src/Sentry/Internal/Http/CachingTransport.cs
```diff
@@ -78,6 +78,13 @@ private CachingTransport(ITransport innerTransport, SentryOptions options, bool
             options.TryGetProcessSpecificCacheDirectoryPath() ??
             throw new InvalidOperationException("Cache directory or DSN is not set.");
 
+        // Sanity check: This should never happen in the first place.
+        // We check for `DisableFileWrite` before creating the CachingTransport.
+        if (_options.DisableFileWrite)
```
bruno-garcia, line 83: if this never happens consider Debug.Assert. 

Adding `throw` has side effects such as the method can no longer be inlined.

src/Sentry/Internal/IFileSystem.cs
```diff
@@ -1,14 +1,12 @@
...
-{
-    Success,
-    Failure,
-    Disabled
-}
-
 internal interface IFileSystem
 {
+    // Note: You are responsible for handling success/failure when attempting to write to disk.
+    // You are required to check for `Options.FileWriteDisabled` whether you are allowed to call any writing operations.
```
bruno-garcia, line 6: This is a good call out. Adding a sexton like <example> or <note> (can't recall the options) above methods in the interface with a snippet might make it even more obvious

src/Sentry/Internal/ReadOnlyFilesystem.cs
```diff
@@ -2,20 +2,24 @@ namespace Sentry.Internal;
 
 internal class ReadOnlyFileSystem : FileSystemBase
 {
-    public override FileOperationResult CreateDirectory(string path) => FileOperationResult.Disabled;
+    // Note: You are responsible for handling success/failure when attempting to write to disk.
```
bruno-garcia, line 5: this won't show up in the IDE at call site. Needs <summary>

bruno-garcia (approved):
Having integration tests related to e2e and with file I/O (sessions etc) have two runs, one with this option set to true and other to false, would really help us avoid regression

src/Sentry/Internal/ReadOnlyFilesystem.cs
jamescrosswell: These interfaces and classes are all internal, so it looks like the only place to document this is in some remarks or something on `SentryOptions.DisableFileWrite`.

jamescrosswell:
> Having the caller simply attempt to write to disk and then not care whether it was allowed to do so is not possible.

@bitsandfoxes can you provide a bit more context around when/why that's not possible?

bitsandfoxes:
> @bitsandfoxes can you provide a bit more context around when/why that's not possible?

Because the caller typically would like to log whether the failed write attempt failed because of a failure or because writing has been disabled. So I ended up with constructs like this
```
var result = _options.FileSystem.CreateDirectory(_persistenceDirectoryPath);
if (result is not FileOperationResult.Success)
{
    if (result is not FileOperationResult.Disabled)
    {
        _options.LogDebug("Disabled.");
    }
    else
    {
        _options.LogError("Failed.");    
    }
    
    return;
}
```
all over the place. So I could not escape checking for `Options.FileWriteDisabled`, either directly or indirectly anyway. I think it's simpler to be explicit about it.

Issue #3353, opened 2024-05-07
[MAUI] Make it possible to programmatically decide whether to attach a screenshot or not


vecalion:
### Problem Statement

Something like `BeforeCaptureCallback` in the Android version (https://docs.sentry.io/platforms/android/enriching-events/screenshots/).

There are two reasons why I'm interested in this feature:
1) Screenshots of some parts of the app may contain private information that is not supposed to be seen by the team monitoring issues in Sentry.
2) There is no need to have screenshots for handled exceptions because in our case, they don't add any new information. However, since they are quite heavy, they generate a lot of unnecessary traffic for users.

jamescrosswell:
Hi @vecalion ,

If you don't want screenshots at all, it is possible to [turn these off in the options](https://docs.sentry.io/platforms/dotnet/guides/maui/enriching-events/screenshots/#enabling-screenshots) when initialising the SDK.

Alternatively, the screenshots for an event are available (and can be altered) via the `Hint` parameter in [the SetBeforeSend callback](https://docs.sentry.io/platforms/dotnet/guides/maui/configuration/filtering/#using-):
```csharp
                options.AttachScreenshot = true;
                options.SetBeforeSend((@event, hint) =>
                {
                    hint.Attachments.Clear();
                    return @event;
                });
```

bitsandfoxes:
Thanks for the workarounds @jamescrosswell.
If we wanted to prevent the overhead of actually creating the screenshot and only then make a decision to discard it I think we need to add a `BeforeCaptureScreenshot` callback to the options and invoke it here:
https://github.com/getsentry/sentry-dotnet/blob/9884fdc7ec13caae0c5c9029dfec2258293eb429/src/Sentry.Maui/Internal/SentryMauiScreenshotProcessor.cs#L19-L23

bricefriha:
#3661 solved this issue

PR #3661, opened 2024-10-07
feat: BeforeCaptureScreenshot option

bricefriha:
Relevant to #3353, I added a option to have a callback just before attaching a screenshot.

### Note
I didn't had a closing keyword to this PR, as I don't believe that change alone fixes the core issue. However this change was demanded


bricefriha:
I did add a change log, maybe I did it wrong 🤔
image

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Features
+
+ - add `BeforeCaptureScreenshot` callback option | #3353
```
```suggestion
 - Added an `BeforeCaptureScreenshot` callback to the options ([#3661](https://github.com/getsentry/sentry-dotnet/pull/3661))
```
The number relates to the actual PR, not the issue.

src/Sentry.Maui/Internal/SentryMauiScreenshotProcessor.cs
```diff
@@ -18,6 +18,9 @@ public SentryMauiScreenshotProcessor(SentryMauiOptions options)
 
     public SentryEvent? Process(SentryEvent @event, SentryHint hint)
     {
+        // Call back before taking the screenshot if the callback is mentioned and if a Screenshot is attached
+        _options.BeforeCaptureScreenshot();
```
bitsandfoxes, line 22: We probably want to validate the return value here to allow opting-out of the screenshot taking. Otherwise the callback doesn't do much. 
That way users gain some control - i.e. they know that they do not want a screenshot taken due to PII at that exact moment.

bitsandfoxes (changes requested):
Fwiw, PR titles we loosely follow a `feat:`, `fix:`, `chore:` convention. No need to add the issue number.

src/Sentry.Maui/Internal/SentryMauiScreenshotProcessor.cs
bricefriha: I think you're right! 
In my head I thought they would validate the return while setting the callback, but it makes very little sense, indeed, in retrospect.
Thanks, I'll do that.

samples/Sentry.Samples.Maui/MauiProgram.cs
```diff
@@ -22,6 +24,11 @@ public static MauiApp CreateMauiApp()
 
                 options.Debug = true;
                 options.SampleRate = 1.0F;
+                options.BeforeCaptureScreenshot ( (SentryEvent @event, SentryHint hint) =>
```
bruno-garcia, line 27: We rather avoid adding dedicated callbacks like that.
For that reason, we added a `Hint` to the API. So we can add stuff like screenshots to the hint, and users can delete the screenshot if they prefer not to capture it.

bruno-garcia (changes requested):

bruno-garcia:
@romtsn how do we deal with this on Android for example?

bruno-garcia:
Looks like we tried binding from Android but we never added Attachment on the .NET hint:

https://github.com/getsentry/sentry-dotnet/blob/9b036457546fe31094e21134127cfa661b3ec357/src/Sentry/Platforms/Android/Extensions/HintExtensions.cs#L14

bruno-garcia:
From in this PR:
* https://github.com/getsentry/sentry-dotnet/pull/2351

This:
> Hints provide two specific functions:

> They allow you to add or remove attachments during any of the mentioned callbacks. (Resolves https://github.com/getsentry/sentry-dotnet/issues/1469)


But unfortunately no docs anywhere to be found

Sentry.sln
bruno-garcia, line 1: Feel free to make a PR just for this file. Since it's unrelated to the other functionality

src/Sentry.Maui/SentryMauiOptions.cs
```diff
@@ -65,4 +65,35 @@ public SentryMauiOptions()
     /// Screenshots can be removed from some specific events during BeforeSend through the Hint.
     /// </remarks>
     public bool AttachScreenshot { get; set; }
+    private Func<SentryEvent, SentryHint, SentryEvent?>? _beforeSend;
+    /// <summary>
+    /// Action performed before attaching a screenshot
+    /// </summary>
+    internal Func<SentryEvent, SentryHint, SentryEvent?>? BeforeCaptureScreenshotInternal => _beforeSend;
```
jamescrosswell, line 72: I got a little bit confused that the internal member is called `_beforeSend` but the public members that set it are called `BeforeCaptureScreenshot`. I'd probably rename the internal member `_beforeCaptureScreenshot` and it could probably be made private.

test/Sentry.Maui.Tests/SentryMauiOptionsTests.cs
```diff
@@ -56,4 +58,35 @@ public void AttachScreenshots_Default()
...
+    {
+        // Arrange
+        Action emptyAction = () => { };
+        var options = new SentryMauiOptions();
+
+        // Act
+        //bool result = options.BeforeCaptureScreenshot == emptyAction;
+
+        // Assert 
+        //Assert.p();
```
jamescrosswell, line 73: TODO???

jamescrosswell (commented):
The code looks good. I made a few styling suggestions. 

I'm not entirely sure what this new option allows SDK users to do that they can't already do with `SentryOptions.SetBeforeSend`... The [Android docs](https://docs.sentry.io/platforms/android/enriching-events/screenshots/#customize-screenshot-capturing) state:
> Because capturing screenshots can be a resource-intensive operation on Android, it's limited to one screenshot every 2 seconds using a debouncing mechanism. This behavior can be overruled if you supply a BeforeCaptureCallback for screenshots in the SentryAndroidOptions.
>
> The BeforeCaptureCallback also allows you to customize the behavior based on event data, so you can decide when to capture a screenshot and when not to. For example, you can decide to only capture screenshots of crashed and fatal events:

But it doesn't look like in the .NET implementation we're using the result of this callback in any way to determine whether a screenshot is added or not in [SentryMauiScreenshotProcessor.cs](https://github.com/getsentry/sentry-dotnet/pull/3661/files#diff-5c60099b287b39d4c18c0f7622784843956be0f820ac4fdb3100a964cbfc8d9d). 

I think it'd be worth adding an example of the new API to one of our samples, demonstrating how you can do something with it that can't already be achieved using SetBeforeSend. That'd be a good sanity check for us, as maintainers (to ensure we're building something meaningful), and would help demonstrate the value of the new functionality to SDK users.

samples/Sentry.Samples.Maui/MauiProgram.cs
bricefriha: Would you like me to remove it from the sample code then?

test/Sentry.Maui.Tests/SentryMauiOptionsTests.cs
bricefriha: oops, i was meant to remove that actually. I thought I would implement it but I figured there is no point, so I wanted to remove it but it seems like I forgot.

thanks for noticing it

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Features
+
+ - Added an `BeforeCaptureScreenshot` callback to the options: allowing the user to set an action before the screenshot is being taken ([#3661](https://github.com/getsentry/sentry-dotnet/pull/3661))
```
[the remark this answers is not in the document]
bricefriha: sorry about that 😅

src/Sentry.Maui/SentryMauiOptions.cs
bricefriha: I think you are right, let's rename it

romtsn:
> @romtsn how do we deal with this on Android for example?

we've added this callback because `BeforeSend` is already too late to determine whether we want to capture a screenshot or not, at that point the screenshot is already there and the performance cost is already paid.  The callback is mainly targeting users who want to avoid capturing screenshots e.g. for errors (default behavior), and only do it for crashes.

FWIW, our callback does not return an event but a boolean that defines whether the screenshot should be captured or not, see: https://github.com/getsentry/sentry-java/blob/b5b093e5f805d0d02c7be344403804e7a7605398/sentry-android-core/src/main/java/io/sentry/android/core/ScreenshotEventProcessor.java#L77-L88

Sentry.sln
bricefriha: PR is here #3665

samples/Sentry.Samples.Maui/MauiProgram.cs
bruno-garcia: I meant to say I don't think we need another BeforeSend style callback because we have already BeforeSend, and the screenshot should be on the Hint. 

That's said we might solve the user problem with the callback Roman described. Where the user returns bool whether they want to capture or not. 

A separate PR can add the screenshot to the Hint so they user can also remove it if they don't want it sent at all

src/Sentry.Maui/Internal/SentryMauiScreenshotProcessor.cs
```diff
@@ -18,6 +18,12 @@ public SentryMauiScreenshotProcessor(SentryMauiOptions options)
 
     public SentryEvent? Process(SentryEvent @event, SentryHint hint)
     {
+        // Call back before taking the screenshot if the callback is not null
+        if (_options.BeforeCaptureScreenshotInternal != null)
```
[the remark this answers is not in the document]
jamescrosswell: @bruno-garcia I've noticed this before. Is this just because it's easier to read (convention for this repo) or is there a technical reason?

test/Sentry.Maui.Tests/SentryMauiOptionsTests.cs
```diff
@@ -56,4 +58,21 @@ public void AttachScreenshots_Default()
...
+        options.AttachScreenshot = true;
+
+        // Act
+        options.BeforeCaptureScreenshot((SentryEvent @event) =>
+        {
+            return @event;
+        });
+
+        // Assert
+        Assert.True(options.BeforeCaptureScreenshotInternal != null);
```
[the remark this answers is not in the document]
bricefriha: makes more sense indeed

test/Sentry.Maui.Tests/SentryMauiOptionsTests.cs
bricefriha, line 1: @bruno-garcia I forgot that XUNIT doesn't have a definition for Assert.IsNotNull(), that might have been why I chose True() instead

test/Sentry.Maui.Tests/SentryMauiOptionsTests.cs
```diff
@@ -56,4 +58,21 @@ public void AttachScreenshots_Default()
...
+        options.AttachScreenshot = true;
+
+        // Act
+        options.BeforeCaptureScreenshot((SentryEvent @event) =>
+        {
+            return @event;
+        });
+
+        // Assert
+        Assert.IsNotNull(options.BeforeCaptureScreenshotInternal);
```
bruno-garcia, line 75:
```suggestion
        Assert.NotNull(options.BeforeCaptureScreenshotInternal);
```

test/Sentry.Maui.Tests/SentryMauiOptionsTests.cs
bruno-garcia: NUnit -> `IsNotNull`
xUnit: -> `NotNull`

I fixed it. But, the PR won't be merged as-is because of the discussion of API signature, and overhead (callback should fire before screenshot is taken)

src/Sentry.Maui/SentryMauiOptions.cs
```diff
@@ -65,4 +65,35 @@ public SentryMauiOptions()
...
+
+    /// <summary>
+    /// Configures a callback function to be invoked before taking a screenshot
+    /// </summary>
+    /// <remarks>
+    /// The event returned by this callback will be sent to Sentry. This allows the
+    /// application a chance to inspect and/or modify the event before it's sent. If the
+    /// event should not be sent at all, return null from the callback.
+    /// </remarks>
+    public void BeforeCaptureScreenshot(Func<SentryEvent, SentryEvent?> beforeSend)
```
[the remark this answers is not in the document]
bricefriha: I'm planning on tackling the changes tomorrow

bricefriha:
@bruno-garcia Per the latest changes, we can bypass the screenshot process similarly to Android, by setting `skipScreenshot` to true. (by default this value is false). In addition to that, the callback is executed before the screenshot is being taken

## Skip screenshot
image

## Default
image


Guys, let me know what you all think.

bricefriha:
I just removed the SentryEvent argument

image

jamescrosswell (approved):
Just a couple of notes about comments but otherwise looks good, as long as CI is happy 👍🏻 

It would be good to add some docs on this to the [Sentry Docs](https://github.com/getsentry/sentry-docs/) as well.

bricefriha:
### Update

The callback works exactly the same as on the Android SDK. The callback returns a boolean that determines whether we ignore the capture or not. 

| MAUI Sdk | [Android Sdk](https://docs.sentry.io/platforms/android/enriching-events/screenshots/)
|---|---|
| image| image

 ([details](https://github.com/getsentry/sentry-java/blob/b5b093e5f805d0d02c7be344403804e7a7605398/sentry-android-core/src/main/java/io/sentry/android/core/ScreenshotEventProcessor.java#L77-L88))

test/Sentry.Maui.Tests/SentryMauiOptionsTests.cs
```diff
@@ -56,4 +58,40 @@ public void AttachScreenshots_Default()
...
+            return true;
+        });
+
+        // Assert
+        Assert.NotNull(options.BeforeCaptureInternal);
+
+    }
+
+    [Fact]
+    public void BeforeCaptureScreenshot_Skip_Screenshot()
```
jamescrosswell, line 81: These two tests both seem to do basically the same thing... which is to ensure that when you call SetBeforeScreenshotCapture the internal field is not null. 

I think it'd be better to have one test where you set a value (which is what you have) and another test where you don't set a value... and make sure you have the expected behaviour in that second scenario as well.

src/Sentry.Maui/Internal/SentryMauiScreenshotProcessor.cs
```diff
@@ -18,6 +18,11 @@ public SentryMauiScreenshotProcessor(SentryMauiOptions options)
 
     public SentryEvent? Process(SentryEvent @event, SentryHint hint)
     {
+        if (!_options.BeforeCaptureInternal?.Invoke(@event, hint) ?? false)
+        {
+            return @event;
+        }
+
```
jamescrosswell, line 25: Probably the most meaningful code to test in this PR would be this bit... so an additional couple of `SentryMauiScreenshotProcessorTest` tests. 

Basically we want to ensure screenshots do get captured if the BeforeCapture callback hasn't been set or returns true... and that a screenshot doesn't get captured if it's set and returns false.

jamescrosswell (changes requested):
Functionally, looks good. I think just tweak the tests...

jamescrosswell (approved):
LGTM!

bruno-garcia (approved):

bitsandfoxes (approved):
Looks great to me!

src/Sentry.Maui/Internal/SentryMauiScreenshotProcessor.cs
bruno-garcia: > if the BeforeCapture callback hasn't been set

We don't have a test for this. Only tests with callback set

bricefriha: I'll add this now

bricefriha: however, `CaptureException_WhenAttachScreenshots_ContainsScreenshotAttachmentAsync` is already testing when BeforeCapture is not set

PR #3662, opened 2024-10-07
Added warning about XCode 16 depenedency to changelog

jamescrosswell:
Draft changes to changelog.

#skip-changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,9 +4,11 @@
 
 ### Support for XCode 16.0 (BREAKING CHANGE)
 
+If you are using XCode 16.0, you will need to update to `Sentry 4.12.0` or later. If you are still using XCode 15.4 then you should continue to use `Sentry 4.11.0` or earlier. 
+
 Using `XCode 16` to build .NET applications targeting iOS and Mac Catalyst requires [iOS SDK 18.0.8303](https://github.com/xamarin/xamarin-macios/releases/tag/dotnet-8.0.1xx-xcode16.0-8303). We [built Sentry 4.12.0 using XCode 16](https://github.com/getsentry/sentry-dotnet/pull/3635/files) in order to support this scenario.
 
-If you are using `Sentry 4.12.x` and targeting iOS or macCatalyst then you will need to install and use `iOS SDK 18.0.8303` (.NET 9 will also support XCode 16, when it is released in November).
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,15 @@
...
 
+## Unreleased
+
+### Support for XCode 16.0 (BREAKING CHANGE)
+
+If you are using XCode 16.0, you will need to update to `Sentry 4.12.0` or later. If you are still using XCode 15.4 then you should continue to use `Sentry 4.11.0` or earlier. 
+
+Using `XCode 16` to build .NET applications targeting iOS and Mac Catalyst requires [iOS SDK 18.0.8303](https://github.com/xamarin/xamarin-macios/releases/tag/dotnet-8.0.1xx-xcode16.0-8303). We [built Sentry 4.12.0 using XCode 16](https://github.com/getsentry/sentry-dotnet/pull/3635/files) in order to support this scenario.
+
+As such, if you are using `Sentry 4.12.x` and targeting iOS or macCatalyst then you will need to install and use XCode 16 and `iOS SDK 18.0.8303` (.NET 9 will also support XCode 16, when it is released in November).
```
bruno-garcia, line 11: Some notes: 
* Changed "November" to "next month" to give a sense of relative time, but also added month and year since "which novmember"?
* Added explicitly the fact we, unfortunately, broke people.
* "Sentry 4" can be confusing because Sentry is often thought as `sentry.io`, and this is "the SDK"
* "iOS SDK 8" can be confusing because iOS does have and SDK, as part of Xcode. We're referring to the .NET workload here, right?

Nits:  :)
* Brand name is `Xcode` not Xcode.
* It's Mac, macOS, and Mac Catalyst.

```suggestion
### Support for XCode 16.0 (BREAKING CHANGE)

If you are using Xcode 16.0, you will need to update the SDK to version `4.12.0` or later. 
If you are still using xCode 15.4 or earlier, you need to continue to use version `4.11.0` or earlier. 

Using Xcode 16 to build .NET applications targeting iOS and Mac Catalyst requires [.NET workload for iOS SDK version 18.0.8303](https://github.com/xamarin/xamarin-macios/releases/tag/dotnet-8.0.1xx-xcode16.0-8303). We [built the SDK version 4.12.0 using Xcode 16](https://github.com/getsentry/sentry-dotnet/pull/3635/files) in order to support this scenario. That, unfortunately, breaks folks using older version of xcode.

As such, if you are using SDK version `4.12.x` and targeting iOS or Mac Catalyst, you will need to install and use Xcode 16 and `workload iOS SDK 18.0.8303` 

Note that .NET 9 will also support Xcode 16, when it is released next month (Nov 2024).
```

bruno-garcia (approved):

PR #3665, opened 2024-10-08
fix: MAUI samples config - enable `deploy`

bricefriha:
Enabling `deploy` in the configuration for the MAUI sample allows the app to be run on VS2022 (on Windows).
image

see: [here](https://developercommunity.visualstudio.com/t/project-not-selected-to-build-for-this-solution-co-2/1533329)

### Note
Linux and Mac, in my experience, are not affected.

#skip-changelog

### DIFF
--- a/Sentry.sln
+++ b/Sentry.sln
@@ -1,184 +1,183 @@
-
 Microsoft Visual Studio Solution File, Format Version 12.00
 # Visual Studio Version 17
 VisualStudioVersion = 17.0.31903.59
 MinimumVisualStudioVersion = 10.0.40219.1
 Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "benchmarks", "benchmarks", "{7D4D7A6A-3F5C-4B4C-A198-AC51F9220231}"
 EndProject
-Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Sentry.Benchmarks", "benchmarks\Sentry.Benchmarks\Sentry.Benchmarks.csproj", "{8328B70C-B808-4ED1-BB16-8555B2752CB6}"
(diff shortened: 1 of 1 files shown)


bricefriha:
Do we have to update the change log for this?  Considering it's not something that would change the user experience

jamescrosswell:
> Do we have to update the change log for this? Considering it's not something that would change the user experience

Nah I think we can skip the changelog on this one.

Sentry.sln
```diff
@@ -276,6 +275,7 @@ Global
 		{4E0DC405-C372-4396-A5DF-F6AA108DA01C}.Release|Any CPU.Build.0 = Release|Any CPU
 		{9B175EC8-6B64-4345-A158-091CB8876077}.Debug|Any CPU.ActiveCfg = Debug|Any CPU
 		{9B175EC8-6B64-4345-A158-091CB8876077}.Debug|Any CPU.Build.0 = Debug|Any CPU
+		{9B175EC8-6B64-4345-A158-091CB8876077}.Debug|Any CPU.Deploy.0 = Debug|Any CPU
```
jamescrosswell, line 278: This looks like the only relevant change... all the rest of it is just messing with GUIDs. Could we possibly make this change manually rather than using Visual Studio, to avoid all the unnecessary messing about that it's done with the solution file?

Eventually we should probably move to slnx files (once the [tooling supports these](https://github.com/dotnet/sdk/issues/40913))

jamescrosswell (changes requested):

bruno-garcia (approved):

bitsandfoxes (approved):

bitsandfoxes:
To avoid `Danger` being annoying you can add `#skip-changelog` to your PR description.

Sentry.sln
bruno-garcia: it's a generated file so not sure it's worth doing any manual changes. Whatever VS does, we just push and merge

bruno-garcia:
> To avoid `Danger` being annoying you can add `#skip-changelog` to your PR description.

Use #skip-changelog if your change doesn't have any user-facing or user-impactful change. For example this PR. Otherwise please add an entry to the changelog that **helps users understand what's in the change. If it's worth them bumping or if they can or should wait for some reason"

jamescrosswell (approved):

Issue #3643, opened 2024-10-01
Add analyzer for `FileSystem`


bitsandfoxes:
## Problem
Without a strict enforcement mechanism, we risk a regression where file system access might bypass the `IFileSystem`.

## Context
Certain platforms, such as the Nintendo Switch, impose strict limitations on disk access. To support these platforms (i.e. allow the Unity SDK to fallback on preventing disk access OOTB) and prevent user projects from crashing, we introduced an option to opt-out of all disk-writing operations.

## Current Implementation
We now route file system access through our `IFileSystem` wrapper. Depending on the configuration in options, the SDK uses either a `ReadOnly` or `ReadWrite` filesystem. This mechanism acts as a fallback to prevent regressions. SDK developers must explicitly check `Options.FileWriteDisable` before attempting any disk write operation. If a write attempt is made with disk access disabled, the filesystem simply returns `false` without further context.

- [PR #3614](https://github.com/getsentry/sentry-dotnet/pull/3614)
- [PR #3641](https://github.com/getsentry/sentry-dotnet/pull/3641)

## Solution
We could/should create an analyzer to ensure that all file system access continues to be routed through the SDK’s custom wrappers. 
For reference on creating custom analyzers: [How to write a C# analyzer](https://learn.microsoft.com/en-us/dotnet/csharp/roslyn-sdk/tutorials/how-to-write-csharp-analyzer-code-fix)

PR #3667, opened 2024-10-09
Automatically capture heap dumps

### LINKED ISSUE #2580: Automatic Heap dump collection
Specify a memory threshold or other trigger (and API) to automatically capture a heap dump

Sentry attachment preview can be extended to load: https://github.com/1hub/dotnet-heapview/tree/import-memory-dump

jamescrosswell:
Resolves #2580

## Basic usage

`samples/Sentry.Samples.Console.Basic/Sentry.Samples.Console.Basic.csproj` includes an example of how to configure automatic heap dumps:

https://github.com/getsentry/sentry-dotnet/blob/ffad1356b5eef1571650bc6c0826d49db36b5b22/samples/Sentry.Samples.Console.Basic/Program.cs#L41-L50

That sample demonstrates using a built in trigger that we provide, that triggers heap dumps when memory usage exceeds a certain threshold (5% in that example). 

Alternatively there is an  override for the `EnableHeapDumps` method that SDK users can call to provide a custom trigger:
https://github.com/getsentry/sentry-dotnet/blob/265e397e63ee3381906186e0cb0e968fcb6f230f/src/Sentry/SentryOptions.cs#L561

### dotnet-gcdump

The SDK relies on [dotnet-gcdump](https://learn.microsoft.com/en-us/dotnet/core/diagnostics/dotnet-gcdump) to capture the heap dumps. 

The `dotnet-gcdump` can be installed globally on the machine or container where the heap dumps will be captured by running:
```
dotnet tool install --global dotnet-gcdump
```

### net6.0

[dotnet-gcdump requires .NET 6 or later](https://www.nuget.org/packages/dotnet-gcdump#supportedframeworks-body-tab).

## Analysing the Dump File
A couple of different options:
1. [dotnet-gcdump report](https://learn.microsoft.com/en-us/dotnet/core/diagnostics/dotnet-gcdump#dotnet-gcdump-report-gcdump_filename)
2. Open the dump file with [Perfview](https://github.com/microsoft/perfview) or [Visual Studio](https://devblogs.microsoft.com/dotnet/collecting-and-analyzing-memory-dumps/#analyzing-gc-dumps-in-visual-studio)... provides richer functionality but is only available on Windows.
3. [dotnet-heapview](https://github.com/1hub/dotnet-heapview/)

See [Stefan Geiger's blog](https://www.stefangeiger.ch/2023/10/03/dotnet-memory-analysis-linux.html) for a bit of info on the first two options.

## Possible improvements

- [ ] According to the docs, the static GC.* methods we use to be notified of garbage collection events [throw an exception when concurrent garbage collection is enabled](https://learn.microsoft.com/en-us/dotnet/api/System.GC.RegisterForFullGCNotification?view=net-8.0#exceptions) so this may not currently work with [Background GC mode](https://learn.microsoft.com/en-us/dotnet/core/runtime-config/garbage-collector#background-gc)
    - If/when server mode is enabled, we may have to revert to either creating a dangling object to detect GC events (not great as it only detects G0 collections) or alternatively some kind of timer
    - Particularly important as [DATAS is the new default mode in net9.0](https://learn.microsoft.com/en-us/dotnet/core/whats-new/dotnet-9/runtime#garbage-collection) and we're [not guaranteed to have a single heap in applications with DATAS configured](https://learn.microsoft.com/en-us/dotnet/standard/garbage-collection/datas#feature-description)
    - When experimenting with this, I didn't observer any `InvalidOperationException` when I attempted to enable Background GC... so we may need to see if any of our users are running into this problem (and can help us reproduce it) before we can fix it. Hopefully this is the kind of feedback we can gather during alpha testing.


jamescrosswell:
I've got to the point where `dotnet-gcdump` can be bundled with the Sentry NuGet package... however I'm not sure that's a good idea. 

We do something similar with the SentryCLI but we only use that at build/compile time... so it's sufficient to ship it as part of our NuGet and then it's available in our transitive build script. 

For `dotnet-gcdump`, we'd actually need it to be shipped as part of the SDK user's application and then extract it to the file system on the ultimate user's machine... That sounds complex to do. We could probably embed it in the SDK users's app via our transitive build script but extracting it from there to the ultimate user's machine would be something best done when installing the app... and we don't really know how our SDK users are going to be installing their apps. This also might be quite unexpected (and meddlesome) from the point of view of both our SDK users and their users. 

I think we might be better off just making it clear that installing `dotnet-gcdump` is a dependency for this feature to work.

src/Sentry/Internal/MemoryMonitor.cs
```diff
@@ -0,0 +1,164 @@
...
+            if (process.StandardOutput.ReadLine() is { } line)
+            {
+                _options.LogDebug(line);
+            }
+        }
+
+        if (!_options.FileSystem.FileExists(dumpFile))
+        {
+            // if this happens, hopefully there would be more information in the standard output from the process above
+            _options.LogError("Unexpected error creating memory dump. Check debug logs for more information.");
```
bruno-garcia, line 117:
```suggestion
            _options.LogError("Unexpected error creating memory dump. Use debug-level to see output of dotnet-gcdump.");
```

src/Sentry/Internal/MemoryMonitor.cs
```diff
@@ -0,0 +1,164 @@
...
+        var command = File.Exists(bundledToolPath)
+            ? $"dotnet {bundledToolPath} {arguments}"
+            : $"dotnet-gcdump {arguments}";
+        var startInfo = new ProcessStartInfo
+        {
+            FileName = RuntimeInformation.IsOSPlatform(OSPlatform.Windows) ? "cmd.exe" : "/bin/bash",
+            Arguments = $"-c \"{command}\"",
+            RedirectStandardOutput = true,
+            UseShellExecute = false,
+            CreateNoWindow = true,
```
bruno-garcia, line 101: Don't we want this as `false`? This will have a popup show up, no?

src/Sentry/Internal/MemoryMonitor.cs
```diff
@@ -0,0 +1,164 @@
...
+        };
+
+        var process = new Process { StartInfo = startInfo };
+        process.Start();
+        while (!process.StandardOutput.EndOfStream)
```
bruno-garcia, line 106: Do we need to subscribe also to stderr in addition to stdout?

src/Sentry/Internal/MemoryMonitor.cs
```diff
@@ -0,0 +1,164 @@
...
+            }
+        }
+
+        if (!_options.FileSystem.FileExists(dumpFile))
+        {
+            // if this happens, hopefully there would be more information in the standard output from the process above
+            _options.LogError("Unexpected error creating memory dump. Check debug logs for more information.");
+        }
+
+        _onDumpCollected(dumpFile);
```
bruno-garcia, line 120: should we try/catch here and leave our internal method without error handling logic instead?

ideally we let things bubble out and capture at the top-most location (unless we want to recover and carry on)

src/Sentry/Internal/MemoryMonitor.cs
```diff
@@ -0,0 +1,164 @@
...
+    }
+
+    internal string? TryGetDumpLocation()
+    {
+        try
+        {
+            var rootPath = _options.CacheDirectoryPath ??
+                           Environment.GetFolderPath(Environment.SpecialFolder.LocalApplicationData);
```
bruno-garcia, line 128: I believe (could be wrong) this can actually create a directory.

See: https://github.com/dotnet/runtime/blob/dfc2b85a00f3baf5ac52d7615bf857e6217011c8/src/libraries/System.Private.CoreLib/src/System/Environment.GetFolderPathCore.Unix.cs#L44

So we'd need to guard this under the `IFileSystem` abstraction, and/or check `options.DisableFileWrite` option before calling this.

src/Sentry/Internal/MemoryMonitor.cs
```diff
@@ -0,0 +1,164 @@
...
+        {
+            _options.LogDebug($"Using bundled version of dotnet-gcdump from: {bundledToolPath}");
+        }
+        else
+        {
+            _options.LogDebug("Using global version of dotnet-gcdump");
+        }
+
+        var arguments = $"collect -p {processId} -o '{dumpFile}'";
+        var command = File.Exists(bundledToolPath)
```
bruno-garcia, line 92: Spare 1 file I/O by assigning the return of the first check 

And, lets always use `_options.FileSystem.FileExists`. 
Ideally we'd have an analyzer because I only realized the `File.Exists` usage because I saw the second pattern further down.

* https://github.com/getsentry/sentry-dotnet/issues/3643

```suggestion
        var bundledFileExists = _options.FileSystem.FileExists(bundledToolPath);
        if (bundledFileExists)
        {
            _options.LogDebug($"Using bundled version of dotnet-gcdump from: {bundledToolPath}");
        }
        else
        {
            _options.LogDebug("Using global version of dotnet-gcdump");
        }

        var arguments = $"collect -p {processId} -o '{dumpFile}'";
        var command = bundledFileExists
            ? $"dotnet {bundledToolPath} {arguments}"
            : $"dotnet-gcdump {arguments}";
```

src/Sentry/buildTransitive/Sentry.targets
```diff
@@ -9,6 +9,12 @@
     <SentryAttributesFile>Sentry.Attributes$(MSBuildProjectExtension.Replace('proj', ''))</SentryAttributesFile>
   </PropertyGroup>
 
+  <PropertyGroup Condition="'$(PkgSentry)' == ''">
+    <!-- If Sentry was added via PackageReference, PkgSentry will already be set to the path of the Sentry nuget package.
+         However, older-style projects using packages.config will need us to resolve the path manually. -->
```
bruno-garcia, line 14: `packages.config` is this a thing on .NET Core apps? TIL

test/Sentry.Testing/SentryOptionsExtensions.cs
```diff
@@ -0,0 +1,70 @@
+namespace Sentry.Testing;
+
+/// <summary>
+/// We most frequently call IDiagnosticLogger via extension methods on the SentryOptions class, which obscures what
+/// we're expecting to receive via mocks when writing tests. This class provides some test extensions that mirror the
+/// extensions provided for SentryOptions to make it easier to write tests that match the code they're testing.
+/// </summary>
+public static class SentryOptionsExtensions
+{
+    private static SentryOptions DidNotReceiveReceiveLog(this SentryOptions substitute, SentryLevel level)
```
bruno-garcia, line 10: Is it named ReceivedReceived on purpose?
```suggestion
    private static SentryOptions DidNotReceiveLog(this SentryOptions substitute, SentryLevel level)
```

bruno-garcia (commented):
If the new File I/O options is set to false, we should prob not do any of this work.

Although not a real concern right now since that flag is used by Unity only, which is on netstandard2.0 (and 2.1 in the near future) but we want to make sure an eventual Unity on .NET Core won't cause crashes on Switch and other platforms

src/Sentry/GCDump.props
```diff
@@ -0,0 +1,30 @@
...
+    <!-- This is just where the relevant files are stored in the dotnet-gcdump Nuget package -->
+    <GCDumpIncludeFiles>$(GCDumpTempDirectory).store\dotnet-gcdump\$(GCDumpVersion)\dotnet-gcdump\$(GCDumpVersion)\tools\net6.0\any\</GCDumpIncludeFiles>
+  </PropertyGroup>
+
+  <Target Name="DownloadGCDump" BeforeTargets="CollectPackageReferences" Condition="'$(PlatformSupportsGcDump)' == 'true'">
+    <Message Importance="normal" Text="Found dotnet-gcdump - skipping download... " Condition="Exists('$(GCDumpIncludeFiles)')" />
+    <Message Importance="normal" Text="dotnet-gcdump not found... downloading to $(GCDumpTempDirectory)" Condition="!Exists('$(GCDumpIncludeFiles)')" />
+
+    <!-- NuGet doesn't have any "download" functionality, so we'll install it locally to a temporary folder   -->
+    <Exec Command="dotnet tool install dotnet-gcdump --version $(GCDumpVersion) --tool-path $(GCDumpTempDirectory)" Condition="!Exists('$(GCDumpIncludeFiles)')" IgnoreExitCode="true" />
```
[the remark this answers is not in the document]
jamescrosswell: `dotnet tool install` does have various arguments that let you control that:
```
 --version <VERSION>          The version of the tool package to install.
 --framework <FRAMEWORK>      The target framework to install the tool for.
-a, --arch <arch>            The target architecture.
```

So if need be, we could install/download it multiple times for different framework and architecture combinations. Anything package that supports net6.0 should also support net7.0+ though right?

It looks like the package includes everything required for multiple platforms:
```
~/code/sentry-dotnet/tools/gcdump/8.0.547301/.store/dotnet-gcdump/8.0.547301/dotnet-gcdump/8.0.547301/tools/net6.0/any git:[auto-heap-dump]
find . -type f
./dotnet-gcdump.pdb
./OSExtensions.dll
./Microsoft.Diagnostics.NETCore.Client.pdb
./Microsoft.Extensions.Logging.Abstractions.dll
./Microsoft.Diagnostics.FastSerialization.dll
./dotnet-gcdump.dll
./System.CommandLine.Rendering.dll
./amd64/msdia140.dll
./amd64/vcruntime140.dll
./amd64/KernelTraceControl.dll
./amd64/vcruntime140_1.dll
./amd64/msvcp140.dll
./Microsoft.Diagnostics.NETCore.Client.dll
./TraceReloggerLib.dll
./dotnet-gcdump.deps.json
./dotnet-gcdump.runtimeconfig.json
./System.CommandLine.dll
./shims/win-arm/dotnet-gcdump.exe
./shims/osx-x64/dotnet-gcdump
./shims/win-x86/dotnet-gcdump.exe
./shims/win-x64/dotnet-gcdump.exe
./Microsoft.Diagnostics.NETCore.Client.xml
./x86/msdia140.dll
./x86/vcruntime140.dll
./x86/KernelTraceControl.dll
./x86/KernelTraceControl.Win61.dll
./x86/msvcp140.dll
./Microsoft.Diagnostics.Tracing.TraceEvent.dll
./DotnetToolSettings.xml
./arm64/msdia140.dll
./arm64/vcruntime140.dll
./arm64/KernelTraceControl.dll
./arm64/vcruntime140_1.dll
./arm64/msvcp140.dll
./Dia2Lib.dll
```

I'll test it out on Windows for a start. Should be able to build something for a docker containers as well, to make sure it runs on at least one linux variant.

samples/Sentry.Samples.Console.Basic/Program.cs
```diff
@@ -33,66 +36,133 @@
...
-        // Simulate doing some work
-        await Task.Delay(100);
-
-        // This is an example of an unhandled exception.  It will be captured automatically.
-        throw new InvalidOperationException("Something happened that crashed the app!");
-    }
-    finally
-    {
-        span.Finish();
+        var array = new byte[2_000_000_000];
```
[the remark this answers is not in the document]
jamescrosswell: We'd have to do it in a different sample since this one sets `<PublishAot>true</PublishAot>`... i.e. no reflection. I could create a separate demo specifically for heap dumps. Worth it?

src/Sentry/Debouncer.cs
```diff
@@ -0,0 +1,108 @@
...
+            case DebouncerInterval.Minute:
+                return TimeSpan.FromMinutes(1);
+            case DebouncerInterval.Hour:
+                return TimeSpan.FromHours(1);
+            case DebouncerInterval.Day:
+                return TimeSpan.FromDays(1);
+            case DebouncerInterval.ApplicationLifetime:
+                return TimeSpan.MaxValue;
+            default:
+                throw new ArgumentOutOfRangeException(nameof(_intervalType));
```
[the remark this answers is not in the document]
jamescrosswell: I would like to say this would only be on Init... but the class and the static factory methods in it are public - so people could try to use it in unusual ways (e.g. to debounce events in their apps that have nothing to do with Heap Dumps). In those instances though, we're no longer dealing with the Sentry SDK proper, but their application code... and we do want to throw exceptions for problems if this class is used in their application code right?

src/Sentry/GCDump.props
```diff
@@ -0,0 +1,30 @@
...
+    <GCDumpTempDirectory>$(GCDumpRootDirectory)$(GCDumpVersion)\</GCDumpTempDirectory>
+    <!-- This is just where the relevant files are stored in the dotnet-gcdump Nuget package -->
+    <GCDumpIncludeFiles>$(GCDumpTempDirectory).store\dotnet-gcdump\$(GCDumpVersion)\dotnet-gcdump\$(GCDumpVersion)\tools\net6.0\any\</GCDumpIncludeFiles>
+  </PropertyGroup>
+
+  <Target Name="DownloadGCDump" BeforeTargets="CollectPackageReferences" Condition="'$(PlatformSupportsGcDump)' == 'true'">
+    <Message Importance="normal" Text="Found dotnet-gcdump - skipping download... " Condition="Exists('$(GCDumpIncludeFiles)')" />
+    <Message Importance="normal" Text="dotnet-gcdump not found... downloading to $(GCDumpTempDirectory)" Condition="!Exists('$(GCDumpIncludeFiles)')" />
+
+    <!-- NuGet doesn't have any "download" functionality, so we'll install it locally to a temporary folder   -->
```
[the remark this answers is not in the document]
jamescrosswell: Aha... weird - that wasn't working the other day for gcdump (was getting redirections etc.). Possibly just a service interruption but you're right - we can download it directly from https://www.nuget.org/api/v2/package/dotnet-gcdump/8.0.547301

src/Sentry/HeapDumpTriggers.cs
```diff
@@ -0,0 +1,27 @@
...
+/// <returns><see langword="true"/> if the heap dump should be triggered; otherwise, <see langword="false"/>.</returns>
+public delegate bool HeapDumpTrigger(long usedMemory, long totalMemory);
+
+internal static class HeapDumpTriggers
+{
+    internal static HeapDumpTrigger MemoryPercentageThreshold(int memoryPercentageThreshold)
+    {
+        if (memoryPercentageThreshold is < 0 or > 100)
+        {
+            throw new ArgumentException("Must be a value between 0 and 100", nameof(memoryPercentageThreshold));
```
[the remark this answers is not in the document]
jamescrosswell: That'd be when calling SentrySdk.Init as well...

src/Sentry/Internal/GarbageCollectionMonitor.cs
```diff
@@ -0,0 +1,28 @@
...
+    public static void Start(Action onGarbageCollected, CancellationToken cancellationToken) =>
+        Task.Run(() => MonitorGarbageCollection(onGarbageCollected, cancellationToken), cancellationToken);
+
+    private static void MonitorGarbageCollection(Action onGarbageCollected, CancellationToken cancellationToken)
+    {
+        GC.RegisterForFullGCNotification(MaxGenerationThreshold, LargeObjectHeapThreshold);
+
+        while (!cancellationToken.IsCancellationRequested)
+        {
+            if (GC.WaitForFullGCComplete(1000) == GCNotificationStatus.Succeeded)
```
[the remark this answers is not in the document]
jamescrosswell: Done - but that overload is net8.0+ only.

src/Sentry/Internal/GarbageCollectionMonitor.cs
```diff
@@ -0,0 +1,28 @@
...
+            {
+                onGarbageCollected?.Invoke();
+            }
+        }
+
+        GC.CancelFullGCNotification();
```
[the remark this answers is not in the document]
jamescrosswell: Yeah not only the callback can throw actually - even the `GC.CancelFullGCNotification()` can throw if concurrent garbage collection is enabled!

src/Sentry/Internal/Hub.cs
```diff
@@ -653,6 +692,10 @@ public void Dispose()
             return;
         }
 
+#if NET6_0_OR_GREATER && !(IOS || ANDROID)
+        _memoryMonitor?.Dispose();
```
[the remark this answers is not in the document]
jamescrosswell: With the current implementation I don't think there's any way for it to throw, but I've added a comment to the code to make sure we think about this if it ever gets more complex in there:
```csharp
    public void Dispose()
    {
        // Important no exceptions can be thrown from this method as it's called when disposing the Hub
        _cancellationTokenSource.Cancel();
    }
```

src/Sentry/Internal/MemoryMonitor.cs
```diff
@@ -0,0 +1,164 @@
...
+    {
+        _options = options;
+        _dumpTrigger = options.HeapDumpTrigger
+                       ?? throw new ArgumentException("No heap dump trigger configured on the options", nameof(options));
+        _onDumpCollected = onDumpCollected;
+        _onCaptureDump = onCaptureDump ?? CaptureMemoryDump;
+
+        _totalMemory = GC.GetGCMemoryInfo().TotalAvailableMemoryBytes;
+
+        GarbageCollectionMonitor.Start(CheckMemoryUsage, _cancellationTokenSource.Token);
```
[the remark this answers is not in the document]
jamescrosswell: We could and I even started playing around with it... but the earliest we can await the task is during MemoryMonitor.Dispose (after calling `token.Cancel()`). That's not ideal as we'd be logging the exception when shutting down the application (rather than when the exception actually occurred). 

Trying to fail early then, I went with your suggestion of using `Task.ContinueWith`... which means the error gets logged as soon as possible... albeit kind of behind the scenes in some task pool thread.

src/Sentry/Internal/MemoryMonitor.cs
jamescrosswell: It's a confusing parameter - it's called Create**No**Window (not CreateWindow)... so setting this to true suppresses a popup.

src/Sentry/Internal/MemoryMonitor.cs
jamescrosswell: We could, if we wanted to log/format errors differently... in that case we could set `ProcessStartInfo.RedirectStandardError = true` and we'd get errors in a separate stream. 

Reading output from multiple streams comes with quite a bit of extra complexity though (reading [the docs](https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.processstartinfo.redirectstandarderror?view=net-8.0#remarks))... and probably not worth it in this case. If there are any problems, we'll detect these below, because the dump file won't be created correctly... and the reason why will be in the log just prior to logging that error (assuming Debug level diagnostic logging is enabled).

src/Sentry/Internal/MemoryMonitor.cs
jamescrosswell: `_onDumpCollected(dumpFile)` is the callback once the dump has already successfully been created. It's mainly exposed to make it easier to unit test. In practice, this gets hooked up the `Hub.CaptureHeapDump` method which sends the heap dump to Sentry... and that does wrap it all in try/catch/log.

We also have `_onCaptureDump` which can be used to mock/intercept calls to `CaptureMemoryDump` in testing:
https://github.com/getsentry/sentry-dotnet/blob/2e6a0f131317fff04658f93aa0815d585b31f7d6/src/Sentry/Internal/MemoryMonitor.cs#L32

Ultimately that gets called via:
 `GarbageCollectionMonitor.MonitorGarbageCollection` -> `onGarbageCollected` (maps to `MemoryMonitor.CheckMemoryUsage`) -> `_onCaptureDump` (maps to `MemoryMonitor.CaptureMemoryDump`)

Any errors arising in that chain will ultimately bubble up to this now (set in the constructor for the MemoryMonitor):
https://github.com/getsentry/sentry-dotnet/blob/2dfd4f404ce15a3d16b993eef3ef1fee843b77de/src/Sentry/Internal/MemoryMonitor.cs#L36-L42

src/Sentry/Internal/MemoryMonitor.cs
jamescrosswell: Damned side affects - the F# community would be ashamed. 

In any event, I've added this in the Hub (when instantiating the MemoryMonitor):
https://github.com/getsentry/sentry-dotnet/blob/2e6a0f131317fff04658f93aa0815d585b31f7d6/src/Sentry/Internal/Hub.cs#L71-L78

src/Sentry/buildTransitive/Sentry.targets
jamescrosswell: Ah, right you are... it's a netframework thing only.

samples/Sentry.Samples.Console.Basic/Program.cs
```diff
@@ -33,66 +36,133 @@
...
+    // This determines the level of heap dump events that are sent to Sentry
+    options.HeapDumpEventLevel = SentryLevel.Warning;
 
-// Do some work. (This is where you'd have your own application logic.)
-await FirstFunction();
-await SecondFunction();
-await ThirdFunction();
+    // A debouncer can be configured to tell Sentry how frequently to send heap dumps. In this case we've configured it
+    // to capture a maximum of 3 events per day and to wait at least 1 hour between each event.
+    options.HeapDumpDebouncer = Debouncer.PerDay(3, TimeSpan.FromHours(1));
```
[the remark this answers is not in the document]
jamescrosswell: I played around with different APIs. I wanted to minimise the confusion between the ThreasholdTriggerPercentage amd the TriggerFunction... to make it really clear you either provided one or the other. So rather than exposing an object that has both of those properties, there are now two overloads of SentryOptions.EnableHeapdumps that basically wrap that options class with what is relevant for each variant.

src/Sentry/GCDump.props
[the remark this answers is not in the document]
jamescrosswell: Testing on two fresh ubuntu containers... 

On the first I installed net6.0, downloaded the NuGet package via curl and unzipped to get exactly the same files we bundle... then ran `dotnet dotnet-gcdump.dll` to ensure this runs correctly. 

On the second I did the same only with net8.0. 

All seems to work fine.

src/Sentry/GCDump.props
```diff
@@ -0,0 +1,30 @@
+<Project>
+  <!-- Check if we're building for a target framework that supports dotnet-gcdump -->
+  <PropertyGroup>
+    <PlatformIsLegacy Condition="$(TargetFramework.StartsWith('net4')) or $(TargetFramework.StartsWith('netstandard'))">true</PlatformIsLegacy>
+    <PlatformIsMobile Condition="$(TargetFramework.EndsWith('android')) or $(TargetFramework.EndsWith('ios')) or $(TargetFramework.EndsWith('maccatalyst'))">true</PlatformIsMobile>
+    <PlatformSupportsGcDump Condition="!($(PlatformIsLegacy) == 'true' or $(PlatformIsMobile) == 'true')">true</PlatformSupportsGcDump>
+
+    <GCDumpVersion>8.0.547301</GCDumpVersion>
```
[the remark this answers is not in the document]
jamescrosswell: I tested on two fresh ubuntu containers...

On the first I installed net6.0, downloaded the NuGet package via curl and unzipped to get exactly the same files we bundle... then ran dotnet dotnet-gcdump.dll to ensure this runs correctly.

On the second I did the same only with net8.0.

All seems to work fine.

Directory.Build.props
```diff
@@ -1,7 +1,7 @@
 <Project>
 
   <PropertyGroup>
-    <VersionPrefix>4.12.1</VersionPrefix>
+    <VersionPrefix>4.12.2-alpha</VersionPrefix>
```
bruno-garcia, line 4: Best to avoid adding this here, lets just version it during the release flow.

samples/Sentry.Samples.Console.Basic/Program.cs
```diff
@@ -4,14 +4,15 @@
  * - Performance Tracing (Transactions / Spans)
  * - Release Health (Sessions)
  * - MSBuild integration for Source Context (see the csproj)
+ * - Heap Dumps (.NET 6.0 or later)
```
bruno-garcia, line 7: The heap dump example was removed but this comment stayed?

samples/Sentry.Samples.Console.Basic/runtimeconfig.template.json
```diff
@@ -0,0 +1,6 @@
+{
+  "configProperties": {
```
bruno-garcia, line 2: do we need this file if there's a new sample?

test/Sentry.Tests/Internals/MemoryMonitorTests.cs
```diff
@@ -0,0 +1,220 @@
...
+    [Fact]
+    public void CaptureMemoryDump_DisableFileWrite_DoesNotCapture()
+    {
+        // Arrange
+        _fixture.Options.EnableHeapDumps(AlwaysTrigger);
+        _fixture.Options.DisableFileWrite = true;
+        using var sut = _fixture.GetSut();
+
+        // Act
+        sut.CaptureMemoryDump();
```
bruno-garcia, line 109: Any tests where `CaptureMemoryDump` throw? More concerned about the caller actually. Since we don't want to crash the app, do we handle every Task we create?

bruno-garcia (commented):
How about we make an alpha from this branch and test on Symbol Collector?

Directory.Build.props
jamescrosswell: That's actually not part of this PR - it comes from the version-5.0.0 branch that we're targeting for this PR. I think you made the same comment on the net9.0 PR (which is also a PR targeting version-5.0.0).

test/Sentry.Tests/Internals/MemoryMonitorTests.cs
jamescrosswell: The Task that calls that is ultimately this:
https://github.com/getsentry/sentry-dotnet/blob/58d9e556a6ab4bb8d760aab9dd5fed321e0d421a/src/Sentry/Internal/MemoryMonitor.cs#L40-L46

I played around with various things and eventually settled on a FireAndForget with a Continuation. The alternative of holding a reference to the task wasn't ideal because you only become aware of any errors when you await the task... which here would have to happen when disposing of the MemoryMonitor (application shutdown). I wanted to log any errors as soon as they occur however.

I've refactored the `GarbageCollectionMonitor` and added some `GarbageCollectionMonitorTests` just to make sure cancellation happens as we expect (without bubbling any exceptions up that might disrupt the calling code). 

I think ultimately what you're wanting to test is [that task continuation](https://github.com/getsentry/sentry-dotnet/blob/58d9e556a6ab4bb8d760aab9dd5fed321e0d421a/src/Sentry/Internal/MemoryMonitor.cs#L40-L46) but the fire and forget nature of that code makes it difficult to test. I did test it manually in a console app, when I was investigating how to detect any non-cancellation errors as soon as they happen.

samples/Sentry.Samples.Console.Basic/runtimeconfig.template.json
jamescrosswell: We need it in the new sample actually... I've moved it across - tnks!

Directory.Build.props
jamescrosswell: No idea how that happened. Reverted in any case.

jamescrosswell:
> How about we make an alpha from this branch and test on Symbol Collector?

We can. Currently the merge target is the version-5.0.0 branch so an alternative would be to make a `5.0.0-alpha1` release from there, once this PR has been approved/merged.

bruno-garcia:
> > How about we make an alpha from this branch and test on Symbol Collector?
> 
> We can. Currently the merge target is the version-5.0.0 branch so an alternative would be to make a `5.0.0-alpha1` release from there, once this PR has been approved/merged.

Lets make the release then, before we approve (lets not merge this accidently)

jamescrosswell:
> Lets make the release then, before we approve (lets not merge this accidently)

Just created this:
https://github.com/getsentry/publish/issues/4530

- Releasing version 14.12.1-dump1 from auto-heap-dump
  - merges back into the auto-heap-dump branch when it completes

bruno-garcia:
@jamescrosswell feel free to add it to https://github.com/getsentry/symbol-collector if it helps test this

samples/Sentry.Samples.Console.HeapDump/Program.cs
```diff
@@ -0,0 +1,72 @@
...
+WriteLine();
+
+WriteLine("Hogging memory...");
+
+// Sentry checks memory usage every time a full garbage collection occurs. It might take a while to trigger this,
+// although we've configured some ridiculously aggressive settings in the runtimeconfig.template.json file to make
+// this happen more quickly, for the purposes of this demo... definitely don't do this in production!
+while (cts.Token.IsCancellationRequested == false)
+{
+    var array = new byte[2_000_000_000];
```
bruno-garcia, line 67: Would still love to see a bunch of different types instantiated so the dump looks more interesting.

src/Sentry/Internal/Hub.cs
```diff
@@ -63,6 +67,17 @@ internal Hub(
...
         }
 
+#if MEMORY_DUMP_SUPPORTED
+        if (options.HeapDumpOptions is not null)
+        {
+            if (_options.DisableFileWrite)
+            {
+                _options.LogError("Automatic Heap Dumps cannot be used with file write disabled.");
+            }
+            _memoryMonitor = new MemoryMonitor(options, CaptureHeapDump);
```
bruno-garcia, line 77: Should we skip instantiating this if file write is disabled?

src/Sentry/Internal/MemoryMonitor.cs
```diff
@@ -0,0 +1,172 @@
...
+        {
+            _options.LogError(ex, "Failed to resolve appropriate memory dump location.");
+            return null;
+        }
+    }
+
+    public void Dispose()
+    {
+        // Important no exceptions can be thrown from this method as it's called when disposing the Hub
+        _cancellationTokenSource.Cancel();
```
bruno-garcia, line 168: CancellationTokenSource has to be disposed too (if you combine tokens with it specially, I recall a memory leak at NRGS because of that :) 

took a while to track down. Ironically I imagine this feature would help with that lol.

bruno-garcia (approved):
I'll be honest I didn't look in detail as much as this change might require but it does seem good to go.

Looking at the [preview package](https://nuget.info/packages/Sentry/14.12.1-dump1) though, it does have a LOT of files. I wonder if we have to be concerned about the increase in size (sentry-cli is huge already)?

image

bruno-garcia:
also maybe @filipnavara might be interested in this feature. So might be willing to review the PR too 🙏

samples/Sentry.Samples.Console.HeapDump/Program.cs
jamescrosswell: You saw my comments on that right? I played around with it but instantiating a bunch of random types led to unexpected errors.

src/Sentry/Internal/MemoryMonitor.cs
jamescrosswell: Ah, good to know... but yes, it would be ironic if the tool we used to help detect memory leaks had a memory leak.

.github/actions/environment/action.yml
```diff
@@ -42,8 +42,7 @@ runs:
       shell: bash
       run: >
         dotnet workload install \
-          wasm-tools wasm-tools-net8 maui-android \
-          ${{ runner.os == 'macOS' && 'maui-ios maui-maccatalyst maui-windows macos' || '' }} \
-          ${{ runner.os == 'Windows' && 'maui-windows' || '' }} \
+          wasm-tools wasm-tools-net8 maui-android maui-ios maui-windows \
+          ${{ runner.os == 'macOS' && 'maui-maccatalyst macos' || '' }} \
```
jamescrosswell: Made this change in the wrong branch...

samples/Sentry.Samples.Console.HeapDump/Program.cs
bruno-garcia: sorry I missed the comment

bruno-garcia:
Is CI on Windows really taking over 1 hour and 20 minutes? or this is hanging? https://github.com/getsentry/sentry-dotnet/actions/runs/11659928344/job/32500819174?pr=3667

The other day I killed a build that seemed to be handing but didn't take note of whcih agent

jamescrosswell:
> Is CI on Windows really taking over 1 hour and 20 minutes? or this is hanging?

Seems to be hanging since merging in the latest updates from the version-5.0.0 branch.

jamescrosswell:
> Is CI on Windows really taking over 1 hour and 20 minutes? or this is hanging?

If I enable diagnostics by running `dotnet test` with the `-v diag --diag {some.log}` arguments, then the tests don't hang and they pass on Windows.

Without diagnostics, the tests hang on Windows. 

Theories:
- Adding diagnostics changes the timings on some things and avoids a deadlock that otherwise occurs. That might also explain why I'm unable to reproduce the issue on my local machine. 
- Is it possible the runner was caching something and got into a funk? When I added diagnostics, I added this as a separate build step that is only run on `windows-latest`... 

I'll try commenting out the diagnostics stuff and running another build.

.github/workflows/build.yml
```diff
@@ -119,7 +119,19 @@ jobs:
       - name: Build
         run: dotnet build Sentry-CI-Build-${{ runner.os }}.slnf -c Release --no-restore --nologo -v:minimal -flp:logfile=build.log -p:CopyLocalLockFileAssemblies=true
 
+#      - name: Test Diagnostics
+#        if: matrix.os == 'windows-latest'
+#        run: dotnet test Sentry-CI-Build-${{ runner.os }}.slnf -c Release --no-build --nologo -l GitHubActions -l "trx;LogFilePrefix=testresults_${{ runner.os }}" --collect "XPlat Code Coverage" -v diag --diag "${{ github.workspace }}/diagnostics.log"
```
bruno-garcia, line 124: dead code?

bruno-garcia:
image
CI has issues

bruno-garcia:
@jamescrosswell so we prob shouldn't bundle stuff after all, lets document folks adding the CLI to the output?
Or can we download as part of the build or release process, directly from NuGet? (assuming that executable is inside a nuget package?)

test/Sentry.Tests/Internals/GarbageCollectionMonitorTests.cs
```diff
@@ -2,9 +2,11 @@ namespace Sentry.Tests.Internals;
 
 public class GarbageCollectionMonitorTests
 {
-    [Fact]
+    [SkippableFact]
     public async Task MonitorGarbageCollection_TaskCancelled_CancelsFullGCNotification()
     {
+        Skip.If(!RuntimeInformation.IsOSPlatform(OSPlatform.Windows), "These tests may be hanging in CI on Windows");
```
filipnavara, line 8: Isn't the condition reversed? (Also, what TFM is targeted in tests? Can this use `OperatingSystem.IsWindows()`?)

jamescrosswell: Hah - good spotting. I just put that in last night before going to sleep...

jamescrosswell:
> @jamescrosswell so we prob shouldn't bundle stuff after all, lets document folks adding the CLI to the output?
> Or can we download as part of the build or release process, directly from NuGet? (assuming that executable is inside a nuget package?)

OK, we could download it as part of the transient build file, but we'd still be shipping it along with customer's applications in that case... which I don't think resolves any legal concerns. 

I think we just advise customer that they need to install the tool globally on any machines where they want to capture heap dumps... and of course we can log a warning/error if `dotnet-gcdump` isn't installed.

bruno-garcia:
> I think we just advise customer that they need to install the tool globally on any machines where they want to capture heap dumps... and of course we can log a warning/error if dotnet-gcdump isn't installed.

OK that's reasonable. And also a good first step. We can always improve from there

test/Sentry.Tests/Internals/GarbageCollectionMonitorTests.cs
bruno-garcia: If CI was passing with that, then I guess tests are not that flakey after all 😅 

Thanks @filipnavara as always

bruno-garcia:
Releasing this branch as `5.0.0-alpha.2`, merging back on itself:

https://github.com/getsentry/publish/issues/4608

bruno-garcia:
Need to wait for this to be merged, and a new craft release:
* https://github.com/getsentry/craft/pull/564

jamescrosswell:
> If/when server mode is enabled, we may have to revert to either creating a dangling object to detect GC events (not great as it only detects G0 collections) or alternatively some kind of timer

If we do need to do this, the DanglingGarbageCollectionMonitor is fairly simple:
```csharp
namespace Sentry.Internal;

/// <summary>
/// Simple class to determine when GarbageCollection occurs
/// </summary>
internal sealed class DanglingGarbageCollectionMonitor
{
    private readonly CancellationToken _cancellationToken;
    private readonly Action _onGarbageCollected;

    private DanglingGarbageCollectionMonitor(Action onGarbageCollected, CancellationToken cancellationToken)
    {
        _cancellationToken = cancellationToken;
        _onGarbageCollected = onGarbageCollected;
    }

    ~DanglingGarbageCollectionMonitor()
    {
        // If monitoring has stopped or the App is shutting down, stop monitoring.
        var stopped = _cancellationToken.IsCancellationRequested;
        if (stopped || Environment.HasShutdownStarted || AppDomain.CurrentDomain.IsFinalizingForUnload())
        {
            return;
        }

        // Every time this class gets cleaned up by the GC, we create another dangling instance of the class (which will
        // be cleaned up in the next GC cycle)... so we keep creating dangling references as long as we want to keep
        // monitoring.
        CreateDanglingMonitor(_onGarbageCollected, _cancellationToken);
        _onGarbageCollected.Invoke();
    }

    private static void CreateDanglingMonitor(Action onGarbageCollected, CancellationToken cancellationToken)
    {
        // ReSharper disable once ObjectCreationAsStatement
#pragma warning disable CA1806
        new DanglingGarbageCollectionMonitor(onGarbageCollected, cancellationToken);
#pragma warning restore CA1806
    }

    public static void Start(Action onGarbageCollected, CancellationToken cancellationToken = default) =>
        CreateDanglingMonitor(onGarbageCollected, cancellationToken);
}
```

And we could alter the `GarbageCollectionMonitor` to leverage this like so:
```csharp
namespace Sentry.Internal;

/// <summary>
/// Simple class to detect when Full Garbage Collection occurs
/// </summary>
internal sealed class GarbageCollectionMonitor
{
    private const int MaxGenerationThreshold = 10;
    private const int LargeObjectHeapThreshold = 10;

    public static Task Start(Action onGarbageCollected, CancellationToken cancellationToken, IGCImplementation? gc = null) =>
        Task.Run(() => MonitorGarbageCollection(onGarbageCollected, cancellationToken, gc), cancellationToken);

    private static void MonitorGarbageCollection(Action onGarbageCollected, CancellationToken cancellationToken, IGCImplementation? gc = null)
    {
        try
        {
            MonitorGarbageCollectionNotifications(onGarbageCollected, cancellationToken, gc);
        }
        catch (OperationCanceledException) when (cancellationToken.IsCancellationRequested)
        {
            // Ignore
        }
        catch (InvalidOperationException)
        {
            // If concurrent GC is enabled, methods like GC.RegisterForFullGCNotification will throw an
            // <see cref="InvalidOperationException"/> so we need to use another mechanism to monitor the GC
            DanglingGarbageCollectionMonitor.Start(onGarbageCollected, cancellationToken);
        }
    }

    private static void MonitorGarbageCollectionNotifications(Action onGarbageCollected, CancellationToken cancellationToken, IGCImplementation? gc = null)
    {
        gc ??= new SystemGCImplementation();
        try
        {
            gc.RegisterForFullGCNotification(MaxGenerationThreshold, LargeObjectHeapThreshold);
            while (!cancellationToken.IsCancellationRequested)
            {
                if (gc.WaitForFullGCComplete(TimeSpan.FromSeconds(1)) == GCNotificationStatus.Succeeded)
                {
                    onGarbageCollected?.Invoke();
                }
            }
        }
        catch (OperationCanceledException) when (cancellationToken.IsCancellationRequested)
        {
            // Ignore
        }
        finally
        {
            gc.CancelFullGCNotification();
        }
    }
}
```

jamescrosswell:
There may be an even better way to detect GC events, that only triggers for G2 events, using [GC.ReRegisterForFinalize](https://learn.microsoft.com/en-us/dotnet/api/system.gc.reregisterforfinalize?view=net-8.0).

There's an example of how to use this in the dotnet runtime source code:
https://github.com/dotnet/runtime/blob/2ac0743b18cf70cc6047f0ec35d3a7a6a4428354/src/libraries/System.Private.CoreLib/src/System/Gen2GcCallback.cs

PR #3674, opened 2024-10-12
drop preview from blazor.wasm package

bruno-garcia:
Over 3k installs. No issues (afaik?)

Seems like we can GA this.

It doesn't do much other than set the right properties so things don't blow up on init, so seems pretty safe


jamescrosswell (approved):
This part all looks fine. 

From what I can tell, we still haven't merged any PRs releated to embedding the JS SDK... which would presumably be the most complex thing to implement and maintain.

bruno-garcia:
> This part all looks fine.
> 
> From what I can tell, we still haven't merged any PRs releated to embedding the JS SDK... which would presumably be the most complex thing to implement and maintain.

yeah but if we haver another alpha bit, we can append the alpha again before a release.
Just gotta be aware that if we add `-beta` on the whole release, we'll be appending something twice (I think there's a way around it actually)

Issue #3654, opened 2024-10-06
Report correct SDK name on native (NDK) Android crashes


bruno-garcia:
Crashes coming form `sentry.native` under the Android SDK should be reported as:

`sentry.native.android.dotnet`

Currently, it's reported as `sentry.native.android`, which is the same as standalone Android apps.

On the Unity SDK, we override it at compile time, and here we don't recompile the SDK so not sure if it can be easily done:

https://github.com/getsentry/sentry-unity/blob/e4d3d4d13ed0ab3f4e2a05534bfb8592c92e319d/Directory.Build.targets#L190

jamescrosswell:
Yeah it looks like in the Java repo it's baked in at build time:
https://github.com/getsentry/sentry-java/commit/66140be425f581aaf57584f2f927b67402a3fca7

I think we'd need some other mechanism to be made available to set this for the .NET SDK.

vaind:
Runtime setting:

https://github.com/getsentry/sentry-java/blob/main/sentry-android-core/src/main/java/io/sentry/android/core/SentryAndroidOptions.java#L145

jamescrosswell:
> Currently, it's reported as `sentry.native.android`, which is the same as standalone Android apps.

Hm, something is off here. We set this in the .NET SDK already:
https://github.com/getsentry/sentry-dotnet/blob/b0a19968ceb0ec5103f3e1887afe2c865351aca3/src/Sentry/Platforms/Android/SentrySdk.cs#L170

So whilst we could change that to `sentry.native.android.dotnet`, if it's coming through as `sentry.native.android` at the moment then that mechanism isn't working as it should at the moment.

bruno-garcia:
> > Currently, it's reported as `sentry.native.android`, which is the same as standalone Android apps.
> 
> Hm, something is off here. We set this in the .NET SDK already:
> 
> [sentry-dotnet/src/Sentry/Platforms/Android/SentrySdk.cs](https://github.com/getsentry/sentry-dotnet/blob/b0a19968ceb0ec5103f3e1887afe2c865351aca3/src/Sentry/Platforms/Android/SentrySdk.cs#L170)
> 
> Line 170 in [b0a1996](/getsentry/sentry-dotnet/commit/b0a19968ceb0ec5103f3e1887afe2c865351aca3)
> 
>              o.NativeSdkName = "sentry.native.dotnet"; 
[...]

PR #3682, opened 2024-10-14
fix: sdk name for native on android

bruno-garcia:
We append `.platform` when 'wrapping' another SDK.

`sentry.native` on Android, because of its custom unwinder and other significant changes (it's also built through sentry-java repo). Has a different SDK name:

https://github.com/getsentry/sentry-native/blob/b555a978986f882abb3161bd1c848339f9a2cc1a/include/sentry.h#L28

Resolves:
* https://github.com/getsentry/sentry-dotnet/issues/3654#issuecomment-2395681843

### DIFF
--- a/src/Sentry/Platforms/Android/SentrySdk.cs
+++ b/src/Sentry/Platforms/Android/SentrySdk.cs
@@ -167,7 +167,7 @@ public static partial class SentrySdk
             o.EnableExternalConfiguration = false;
             o.EnableDeduplication = false;
             o.AttachServerName = false;
-            o.NativeSdkName = "sentry.native.dotnet";
+            o.NativeSdkName = "sentry.native.android.dotnet";
 
             // These options are intentionally not expose or modified
             //o.MaxRequestBodySize   // N/A for Android apps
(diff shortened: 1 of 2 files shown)


src/Sentry/Platforms/Android/SentrySdk.cs
```diff
@@ -167,7 +167,7 @@ private static void InitSentryAndroidSdk(SentryOptions options)
             o.EnableExternalConfiguration = false;
             o.EnableDeduplication = false;
             o.AttachServerName = false;
-            o.NativeSdkName = "sentry.native.dotnet";
+            o.NativeSdkName = "sentry.native.android.dotnet";
```
jamescrosswell, line 170: I take it this doesn't address whatever the more fundamental problem with https://github.com/getsentry/sentry-dotnet/issues/3654#issuecomment-2395681843 is?

jamescrosswell (approved):
LGTM

src/Sentry/Platforms/Android/SentrySdk.cs
bitsandfoxes: Yes it does.

PR #3693, opened 2024-10-17
chore: Comment why `format-code` runs on `macos-latest`

bitsandfoxes:
We're currently forced to run on macos for `format-code` because restoring the workloads on `Ubuntu` causes errors for `Sentry.Samples.Ios.csproj`, `Sentry.Samples.MacCatalyst.csproj`, and `Sentry.Samples.MacOS.csproj`:

```
The project depends on the following workload packs that do not exist in any of the workloads available in this installation: Microsoft.iOS.Sdk.net8.0_18.0
You may need to build the project on another operating system or architecture, or update the .NET SDK. 
```

#skip-changelog


bruno-garcia (approved):
Thank you!

PR #3694, opened 2024-10-18
fix: Android - make sure logcat.log is attached to unhandled event (SIGSEGV Segfault)

### LINKED ISSUE #3461: logcat.log is not attached to unhandled event (SIGSEGV Segfault)
### Package

Sentry.Maui

### .NET Flavor

.NET

### .NET Version

8.0.6

### OS

Android

### SDK Version

4.7.0

_No response_

### Steps to Reproduce

```
options.Android.LogCatIntegration = LogCatIntegrationType.Unhandled;
```

I don't know how to provoke `SIGSEGV Segfault`.

### Expected Result

It is expected that the error event has an attachment `logcat.log`.

### Actual Result

There are no attachments to the event.

bricefriha:
From [sentry-native#1027](https://github.com/getsentry/sentry-native/pull/1027), we had to add the handler strategy to the .NET SDK, from the sentry-java ([7.15.0-alpha1](https://github.com/getsentry/sentry-java/releases/tag/7.15.0-alpha.1)).

This is also meant to fix #3461. I tested this, and it seems to be fixed now.

More context at:
* https://github.com/dotnet/android/issues/9055#issuecomment-2261347912


src/Sentry.Bindings.Android/Sentry.Bindings.Android.csproj
```diff
@@ -1,9 +1,9 @@
-<Project Sdk="Microsoft.NET.Sdk">
+﻿<Project Sdk="Microsoft.NET.Sdk">
```
bruno-garcia, line 1: some BOM or setting in your IDE is adding these to the diff, please address that

src/Sentry.Bindings.Android/Sentry.Bindings.Android.csproj
```diff
@@ -37,24 +37,9 @@
...
-      Condition="!Exists('$(SentryAndroidSdkDirectory)sentry-android-ndk-$(SentryAndroidSdkVersion).aar')"
-      Retries="3"
-    />
-    <DownloadFile
-      SourceUrl="https://repo1.maven.org/maven2/io/sentry/sentry/$(SentryAndroidSdkVersion)/sentry-$(SentryAndroidSdkVersion).jar"
-      DestinationFolder="$(SentryAndroidSdkDirectory)"
-      Condition="!Exists('$(SentryAndroidSdkDirectory)sentry-$(SentryAndroidSdkVersion).jar')"
-      Retries="3"
-    />
+    <DownloadFile SourceUrl="https://repo1.maven.org/maven2/io/sentry/sentry-android-core/$(SentryAndroidSdkVersion)/sentry-android-core-$(SentryAndroidSdkVersion).aar" DestinationFolder="$(SentryAndroidSdkDirectory)" Condition="!Exists('$(SentryAndroidSdkDirectory)sentry-android-core-$(SentryAndroidSdkVersion).aar')" Retries="3" />
```
bruno-garcia, line 40: I don't understand what's changing here, is this some IDE messing up?

src/Sentry/NdkHandlerStrategy.cs
```diff
@@ -0,0 +1,25 @@
...
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Sentry
+{
+    /// <summary>
+    /// Handler strategy
+    /// </summary>
+    /// 
+    public enum NdkHandlerStrategy
```
bruno-garcia, line 13: This is Android specific, but it was added to the whole SDK regardless of platform. Please add under `/Android/`

It should only show up if you're `#if ANDROID`

src/Sentry/SentryOptions.cs
```diff
@@ -77,6 +77,16 @@ public bool IsGlobalModeEnabled
...
 #endif
+    private NdkHandlerStrategy _handlerStrategy = NdkHandlerStrategy.SENTRY_HANDLER_STRATEGY_DEFAULT;
+    /// <summary>
+    /// Handler Strategy
+    /// </summary>
+    public NdkHandlerStrategy HandlerStrategy
+    {
+        get { return _handlerStrategy; }
+        set { _handlerStrategy = value; }
+
```
bruno-garcia, line 88: nit
```suggestion
```

src/Sentry/SentryOptions.cs
```diff
@@ -77,6 +77,16 @@ public bool IsGlobalModeEnabled
         set => _isGlobalModeEnabled = value;
     }
 #endif
+    private NdkHandlerStrategy _handlerStrategy = NdkHandlerStrategy.SENTRY_HANDLER_STRATEGY_DEFAULT;
+    /// <summary>
+    /// Handler Strategy
```
bruno-garcia, line 82: Please describe further what this is for.
Consider adding `<seealso href="https://github.com/dotnet/android/issues/9055#issuecomment-2261347912">`

src/Sentry/SentryOptions.cs
```diff
@@ -77,6 +77,16 @@ public bool IsGlobalModeEnabled
         set => _isGlobalModeEnabled = value;
     }
 #endif
+    private NdkHandlerStrategy _handlerStrategy = NdkHandlerStrategy.SENTRY_HANDLER_STRATEGY_DEFAULT;
```
bruno-garcia, line 80: Why do we need the private backing field?

Couldn't this be a 1 liner?

```csharp
public NdkHandlerStrategy HandlerStrategy {get;set;} = NdkHandlerStrategy.SENTRY_HANDLER_STRATEGY_DEFAULT;
```

test/Sentry.Maui.Tests/SentryMauiOptionsTests.cs
```diff
@@ -56,4 +56,30 @@ public void AttachScreenshots_Default()
         var options = new SentryMauiOptions();
         Assert.False(options.AttachScreenshot);
     }
+    [Fact]
```
bruno-garcia, line 59: nit:
we add lines between test
```suggestion

    [Fact]
```

test/Sentry.Maui.Tests/SentryMauiOptionsTests.cs
```diff
@@ -56,4 +56,30 @@ public void AttachScreenshots_Default()
...
+        var expected = NdkHandlerStrategy.SENTRY_HANDLER_STRATEGY_DEFAULT;
+
+        // Assert
+        // testing default state here so nothing to assert
+
+        // Act
+        Assert.Equal(expected, options.HandlerStrategy);
+    }
+    [Fact]
+    public void HandlerStrategy_set()
```
bruno-garcia, line 73: we're just testing setters here, can we add some integration test somehow?

Can we repro this null reference to SIGSEGV issue? We have tests in the repo where we have the SDK capture something and we grab the event from the transport and asset on what was created.

We'd add tests and `#if ANDROID`. cc @jamescrosswell @vaind are DeviceTests alive? I know we had issues
If so, any points you could give @bricefriha on how to do this?

src/Sentry/Platforms/Android/SentrySdk.cs
```diff
@@ -59,6 +59,11 @@ private static void InitSentryAndroidSdk(SentryOptions options)
 
         // Define the configuration for the Android SDK
         SentryAndroidOptions? nativeOptions = null;
+
+        // Convert NdkHandlerStrategy .net to Java Android
```
bruno-garcia, line 63: nit: inconsistent. You're capitalizing Java but not .NET
Either go with Java and .NET or, java and .net
```suggestion
        // Convert the .NET NdkHandlerStrategy to Java / Android
```

src/Sentry/Platforms/Android/SentrySdk.cs
```diff
@@ -59,6 +59,11 @@ private static void InitSentryAndroidSdk(SentryOptions options)
...
+        var handlerStrat = (int)options.HandlerStrategy;
+        JavaSdk.Android.Core.NdkHandlerStrategy? handlerStrategy = JavaSdk.Android.Core.NdkHandlerStrategy.Values()?[handlerStrat];
```
bruno-garcia, line 65: I'm struggling to parse. Why would `Values()` return null?
Can we use `var`
```suggestion
        var handlerStrategy = JavaSdk.Android.Core.NdkHandlerStrategy.Values()[(int)options.HandlerStrategy];
```

An extension method on `NdkHandlerStrategy` is probably a good solution:

```csharp
ToJava()
```

src/Sentry.Bindings.Android/Sentry.Bindings.Android.csproj
bricefriha: yes it's VS that updates the indentation. I should live this out of git staging next time

src/Sentry/Platforms/Android/SentrySdk.cs
bricefriha: I'll try the `ToJava()` thanks for the suggestion

src/Sentry.Bindings.Android/Sentry.Bindings.Android.csproj
```diff
@@ -1,9 +1,9 @@
-<Project Sdk="Microsoft.NET.Sdk">
+﻿<Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
     <TargetFramework>net7.0-android</TargetFramework>
     <!-- BG8605 and BG8606 happen because there's a missing androidx.lifecycle dependency, but we don't need it here.  (The native Android Sentry SDK will use it if it exists.) -->
     <NoWarn>$(NoWarn);BG8605;BG8606</NoWarn>
-    <SentryAndroidSdkVersion>7.14.0</SentryAndroidSdkVersion>
+    <SentryAndroidSdkVersion>7.15.0-alpha.1</SentryAndroidSdkVersion>
```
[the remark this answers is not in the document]
bricefriha: it was more so for the sake of testing locally.
e.g. if you pull the branch and want to run it, you just have to rebuild the android bindings and you should be good to go

src/Sentry.Bindings.Android/Sentry.Bindings.Android.csproj
bricefriha: i'm just gonna be more meticulous with git staging next time

src/Sentry/SentryOptions.cs
bricefriha: absolutely, I don't know why I did it this way. 
I think I was going for another approach at first

src/Sentry/Platforms/Android/SentrySdk.cs
bricefriha: > Why would Values() return null?

@bruno-garcia In my experience, it's always better to assume that a method can return null, even if it seems impossible. I quite feel strongly about it. 
We would rather get it to default to `SENTRY_HANDLER_STRATEGY_DEFAULT` than throw an unhandled exception which would cause things to break somewhere else that would be harder to troubleshoot in 10 months. `Values()?` doesn't have any cost compared to `Values()`. 

Now `JavaSdk.Android.Core.NdkHandlerStrategy?` is silly, let's do a JavaSdk.Android.Core.NdkHandlerStrategy.Values()?[handlerStrat] ?? JavaSdk.Android.Core.NdkHandlerStrategy.SENTRY_HANDLER_STRATEGY_DEFAULT

src/Sentry/NdkHandlerStrategy.cs
```diff
@@ -0,0 +1,32 @@
...
+        SENTRY_HANDLER_STRATEGY_CHAIN_AT_START,
+
+
+    }
+    /// <summary>
+    /// Extension class of strategy containing relevant its actions 
+    /// </summary>
+    public static class  HandlerStrategyExtension
+    {
+#if ANDROID
```
bruno-garcia, line 25: This means we end up with an empty class on platforms other than Android. Why would we do that?

src/Sentry/Platforms/Android/SentrySdk.cs
```diff
@@ -87,6 +90,7 @@ private static void InitSentryAndroidSdk(SentryOptions options)
             o.ServerName = options.ServerName;
             o.SessionTrackingIntervalMillis = (long)options.AutoSessionTrackingInterval.TotalMilliseconds;
             o.ShutdownTimeoutMillis = (long)options.ShutdownTimeout.TotalMilliseconds;
+            o.SetNativeHandlerStrategy(handlerStrategy ?? JavaSdk.Android.Core.NdkHandlerStrategy.SentryHandlerStrategyDefault);
```
bruno-garcia, line 93: Why do we need the null coalesce operator if `ToJava` doesn't return null?

Besides complexity an additional issue with this is that now we have 2 places dealing with null and defining a default. Had we even needed to change the default we'd need to 'remember' to change these two spots

src/Sentry/Platforms/Android/SentrySdk.cs
```diff
@@ -185,11 +189,11 @@ private static void InitSentryAndroidSdk(SentryOptions options)
         }
         else
         {
-            SentryAndroid.Init(AppContext, configuration);
+            SentryAndroid.Init(AppContext);
```
bruno-garcia, line 192: This change doesn't seem related to the goal of this PR. Or is it?

Why was `configuration` arg removed?

src/Sentry/Platforms/Android/SentrySdk.cs
```diff
@@ -185,11 +189,11 @@ private static void InitSentryAndroidSdk(SentryOptions options)
...
         }
 
         // Set options for the managed SDK that depend on the Android SDK. (The user will not be able to modify these.)
-        options.AddEventProcessor(new AndroidEventProcessor(nativeOptions!));
+        options.AddEventProcessor(new AndroidEventProcessor((SentryAndroidOptions)nativeOptions!));
```
bruno-garcia, line 196: Could you clarify why we need this cast now?

src/Sentry/Platforms/Android/SentrySdk.cs
bricefriha: mistake my bad

src/Sentry/Platforms/Android/SentrySdk.cs
bruno-garcia: > We would rather get it to default to SENTRY_HANDLER_STRATEGY_DEFAULT than throw an unhandled exception which would cause things to break somewhere else that would be harder to troubleshoot in 10 months. Values()? doesn't have any cost compared to Values().

It's more nuanced than that. With the `??` approach we also will scratch our heads when things are not working the way we expected: "I set the handler to A, why is B happening"?

You should add in code or comments that intention you describe in this comment then. Why would it be null in this case? Is some possible bindings problem? Is it the caller could be on a version of C# pre-nullability?
Once you clarify why that could be the case, you can use our debug logger. Try to read the value and if it's null you use `LogError("expected a value set for handler since a default is hard coded but got null")

IMO that's overkill, and throwing `??` all over the code base "just in case" is far from ideal.

That said I share the sentiment that we can't foresee all issues and this is a crash reporting SDK so we need ot be careful. The way we go about that is by using error handling and our debug logging. You could do that on `ToJava` only, and avoid the other `??` usage. `ToJava` would require the internal `ILogger` as an argument to properly log out the error. But unless there's a reasonable suspicion (based on experience or otherwise) that reading that value could be null even though we're hard coding values all over the place, it's enough to have a higher level try/catch deal (debug logging, always) with it and avoid the app from crashing.

bricefriha: fair enough, you're right

src/Sentry/Platforms/Android/NdkHandlerStrategy.cs
```diff
@@ -0,0 +1,30 @@
+﻿namespace Sentry
+{
```
jamescrosswell, line 2: We prefer File-scoped namespace declaration over Block-scoped namespace declaration in this repo.

test/Sentry.Maui.Tests/SentryMauiOptionsTests.cs
jamescrosswell: > cc @jamescrosswell @vaind are DeviceTests alive? I know we had issues

They're working fine on the main branch. We've got issues with them when we try to make them target net8.0-android or net8.0-ios... we don't have to do that until we bump to net9.0 though.

src/Sentry/SentryOptions.cs
```diff
@@ -77,6 +77,20 @@ public bool IsGlobalModeEnabled
...
+    /// SentrySdk.Init(options =>
+    /// {
+    /// Sentry
+    /// // ...
+    /// options.HandlerStrategy = NdkHandlerStrategy.SENTRY_HANDLER_STRATEGY_DEFAULT;
+    /// // ... other options
+    /// }
+    /// </code>
+    /// <seealso href="https://github.com/dotnet/android/issues/9055#issuecomment-2261347912"/>
+    public NdkHandlerStrategy HandlerStrategy { get; set; } = NdkHandlerStrategy.SENTRY_HANDLER_STRATEGY_DEFAULT;
```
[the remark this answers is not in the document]
bricefriha: Yes, now that you mention it, I think it would make more sense to have `SENTRY_HANDLER_STRATEGY_CHAIN_AT_START ` as a default as a constant, as I don't see any scenario where we wouldn't use it. Especially since the issue comes from dotnet-generated machine code

bricefriha:
@jamescrosswell it seems like the wrong version of the Java SDK is set there. 
Which causes the CI to crash.  
image

Is there a way to set this version to 7.15.0-alpha.1 without updating Sentry.Bindings.Android.csproj?

jamescrosswell:
> Is there a way to set this version to `7.15.0-alpha.1` without updating `Sentry.Bindings.Android.csproj`

I think we do want to update the Sentry.Bindings.Android.csproj right? How would we be using the new Native functionality otherwise?

src/Sentry.Bindings.Android/Sentry.Bindings.Android.csproj
```diff
@@ -3,7 +3,7 @@
     <TargetFramework>net7.0-android</TargetFramework>
     <!-- BG8605 and BG8606 happen because there's a missing androidx.lifecycle dependency, but we don't need it here.  (The native Android Sentry SDK will use it if it exists.) -->
     <NoWarn>$(NoWarn);BG8605;BG8606</NoWarn>
-    <SentryAndroidSdkVersion>7.15.0</SentryAndroidSdkVersion>
+    <SentryAndroidSdkVersion>7.15.0-alpha.1</SentryAndroidSdkVersion>
```
bricefriha, line 6: We changed the SdkVersion to `7.15.0-alpha.1` just for the alpha-release and to make it easy for CI. We can remove it just before merging to `main`

[more](https://github.com/getsentry/sentry-dotnet/pull/3694#issuecomment-2430134945)

jamescrosswell: OK, is the plan to wait for this to be shipped in the main branches of the native and android libraries first then? Have we confirmed this resolves the original issue that it was meant to address?

jamescrosswell: OK cool. So maybe we don't merge this into main just yet... we can make an alpha release from this branch "so folks can try that out".

Thanks heaps @bricefriha - nice work!

bruno-garcia: with this we're markign only this package as alpha. How do we release from the branch? If I set the whole repo's version `-alpha`, this package will be `-alpha0.1-alpha0.1`.

jamescrosswell: @bruno-garcia I thought this indicated we depend on an alpha of the Sentry Android SDK (it doesn't mark any of the packages in the sentry-dotnet repo as alpha). Is that not the case?

bruno-garcia: lets make an alpha of this branch alone (the whole SDK can be alpha).
otherwise we need to merge into `main` and if we make previews from there, we'll have teh issue I mentioend

test/Sentry.Maui.Tests/SentryMauiOptionsTests.cs
```diff
@@ -52,12 +52,32 @@ public void CacheDirectoryPath_Default()
...
     [Fact]
-    public void AttachScreenshots_Default()
+    public void HandlerStrategy_Default()
     {
+        // Arrange
+        var expected = Android.LogCatIntegrationType.None;
+        var options = new SentryMauiOptions();
+
+        // Assert
+        Assert.Equal(expected, options.Android.LogCatIntegration);
```
bruno-garcia, line 64: Why do we have logcat tests on the NDK PR?

bricefriha: As mentioned in the description, this PR is mostly to fix #3461. The issue was that the logcat file was not attached.

So, it's crucial to test that the logcat file is being attached properly

jamescrosswell: @bricefriha I think [that comment](https://github.com/dotnet/android/issues/9055#issuecomment-2261347912) describes a specific situation in which logcat logs aren't sent. More generally, as long as we're not dealing with SEGSEGV exceptions like that, we should be getting logcat reports come through though, so it should be possible to test this fix without mixing the two PRs right?

CHANGELOG.md
```diff
@@ -2,15 +2,17 @@
 
 ## Unreleased
 
+### Fixes
+
+- Android: Fix for SIGSEGV reported by Sentry when a NullReferenceException is thrown in Release mode ([#3694](https://github.com/getsentry/sentry-dotnet/pull/3694))
+  - To turn it on, set: `options.HandlerStrategy = NdkHandlerStrategy.SENTRY_HANDLER_STRATEGY_CHAIN_AT_START;`
```
jamescrosswell, line 8:
```suggestion
```

This isn't an option right? We always use `SENTRY_HANDLER_STRATEGY_CHAIN_AT_START` when initialising the integration from the .NET SDK?

CHANGELOG.md
```diff
@@ -2,15 +2,17 @@
...
+- Events from NDK on Android will report sdk.name `sentry.native.android.dotnet` ([#3682](https://github.com/getsentry/sentry-dotnet/pull/3682))
```
jamescrosswell, line 9: Isn't this a comment from another PR?

CHANGELOG.md
```diff
@@ -2,15 +2,17 @@
 
 ## Unreleased
 
+### Fixes
+
+- Android: Fix for SIGSEGV reported by Sentry when a NullReferenceException is thrown in Release mode ([#3694](https://github.com/getsentry/sentry-dotnet/pull/3694))
```
jamescrosswell, line 7: Is that this PR or another PR? The title talks about attachments but the changelog talks about fixing a `NullReferenceException`...

CHANGELOG.md
bricefriha: It isn't anymore indeed, I forgot to update it. thanks for noticing this

CHANGELOG.md
bricefriha: Not an open one at least. 
this is on [main](https://github.com/getsentry/sentry-dotnet/blob/main/CHANGELOG.md#L12) for sure

CHANGELOG.md
bricefriha: it's not from here, I have no idea how it ended up here

bricefriha: similarly it's coming from me

either way it's fixed now

CHANGELOG.md
jamescrosswell: Ah I see, it's just been moved from below the `###Features` in this branch. I was wondering why it was showing as being added in this PR.

bricefriha: yeah I see

do you think I should change that?

jamescrosswell: Historically we've put the features before the Fixes... just a question of what we want to draw people's attention to first. Not a big deal.

jamescrosswell:
Requested [this release](https://github.com/getsentry/publish/issues/4545) that we can share with testers.

bruno-garcia:
Note: https://github.com/getsentry/publish/issues/4545#issuecomment-2452007323

bruno-garcia:
which issue to reach out for feedback? anyone trying this out to help validate?

jamescrosswell:
> which issue to reach out for feedback? anyone trying this out to help validate?

See comment:
- https://github.com/getsentry/sentry-dotnet/issues/3461#issuecomment-2453766597

bricefriha:
> anyone trying this out to help validate?

@bruno-garcia @jamescrosswell  I created a repo to test this alpha release and it works perfectly https://github.com/bricefriha/Sentry-3694

bruno-garcia:
> > anyone trying this out to help validate?
> 
> @bruno-garcia @jamescrosswell I created a repo to test this alpha release and it works perfectly [bricefriha/Sentry-3694](https://github.com/bricefriha/Sentry-3694)

Awesome! thank you for that.
Shoudl we merge?

bricefriha:
>Shoudl we merge?

We could, but the issue is that we need to use the Android SDK version 7.15.0-alpha.1 until the NDKHandler setting is available on the full release. So shouldn't we wait for that first?

bruno-garcia:
> > Shoudl we merge?
> 
> We could, but the issue is that we need to use the Android SDK version 7.15.0-alpha.1 until the NDKHandler setting is available on the full release. So shouldn't we wait for that first?

If it's validated, we should reach out to folks on Android to merge that to `main` and ship a new release.

We can merge ours (after bumping to that version) then.

Also: Should that option be the new default? So users don't need to do anything?

bricefriha:
>We can merge ours (after bumping to that version) then.

sounds good 

>Should that option be the new default? So users don't need to do anything?

yes it is, so basically users don't need to change anything at all

bruno-garcia:
What's the next steps for this?

bricefriha:
> What's the next steps for this?

I let the Java team know that we are ready to merge so they are working on merging their PRs as well

once they have the update on the main release, we can merge

bruno-garcia:
merged! https://github.com/getsentry/sentry-native/pull/1027#event-15314896031

We'd like to wait for their release though, so we can checkout the release tag instead of a random commit sha

bricefriha:
> We'd like to wait for their release though

The release is out now!
https://github.com/getsentry/sentry-native/releases/tag/0.7.13

We just need an update from `sentry-java`

jamescrosswell:
> The release is out now! https://github.com/getsentry/sentry-native/releases/tag/0.7.13
> 
> We just need an update from `sentry-java`

@markushi any timeframe on getting the fix we tested in [7.15.0-alpha.1](https://github.com/getsentry/sentry-java/releases/tag/7.15.0-alpha.1) into an official release? I'd love to roll this out in the .NET SDK but we've got a bit of a dependency chain.

Looks like we'll have to do something similar for iOS as well, and it would be nice to reference this solution when discussing with the cocoa team.

markushi:
> > The release is out now! [getsentry/sentry-native@`0.7.13` (release)](https://github.com/getsentry/sentry-native/releases/tag/0.7.13)
> > We just need an update from `sentry-java`
> 
> @markushi any timeframe on getting the fix we tested in [7.15.0-alpha.1](https://github.com/getsentry/sentry-java/releases/tag/7.15.0-alpha.1) into an official release? I'd love to roll this out in the .NET SDK but we've got a bit of a dependency chain.
> 
> Looks like we'll have to do something similar for iOS as well, and it would be nice to reference this solution when discussing with the cocoa team.

If it's tested and we know the change works, we can merge and get this released at any time. Let me get the [PR](https://github.com/getsentry/sentry-java/pull/3671) up-to-date first.

jamescrosswell:
Looks like the fix was included with the [7.19.0 release of the Java SDK](https://github.com/getsentry/sentry-java/blob/main/CHANGELOG.md#7190). We just need to bump the version in this PR and we should be good to go!

jamescrosswell (approved):

PR #3695, opened 2024-10-18
fix: MAUI samples config - enable iOS/MAC deployment and debug for Windows

bricefriha:
As mentioned on Discord, I think it would be handy to allow Windows machines to deploy the iOS sample app using a paired Mac for testing purposes.

### Resources
in case you need some guidance to Pair a Mac to your Windows PC, Microsoft has some excellent [documentation for this](https://learn.microsoft.com/en-us/dotnet/maui/ios/pair-to-mac?view=net-maui-8.0)

#skip-changelog

### DIFF
--- a/samples/Sentry.Samples.Maui/Sentry.Samples.Maui.csproj
+++ b/samples/Sentry.Samples.Maui/Sentry.Samples.Maui.csproj
@@ -6,8 +6,9 @@
       On Mac, we'll also build for iOS and MacCatalyst.
       On Windows, we'll also build for Windows 10.
     -->
-    <TargetFrameworks>net8.0-android;net8.0-ios;net8.0-maccatalyst</TargetFrameworks>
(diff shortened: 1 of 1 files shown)


jamescrosswell:
@bricefriha would this then require people using the solution on Windows to have a paired Mac? We probably wouldn't want to do that as it would limit the number of people who could contribute to the repo... so it'd be good to "allow" people to work on Windows with a paired Mac but not "require" a paired Mac when working on Windows.

bricefriha:
> so it'd be good to "allow" people to work on Windows with a paired Mac but not "require" a paired Mac when working on Windows

sorry for the confusion, you don't require to pair it to a mac with this config. The changes allows people to pair a mac but they don't have to.

e.g: I have a project with a similar configuration and I was able to work on it while in Germany away from my Mac

bruno-garcia (approved):
LGTM

bruno-garcia:
Thanks. @jamescrosswell unless this breaks your workflow somehow, otherwise please merge

jamescrosswell (approved):

Issue #3703, opened 2024-10-20
logcat attachment can't be previewed in Sentry


bruno-garcia:
### Integration

sentry-android

### Build System

Other

### AGP Version

Not sure, using .NET build tools

### Proguard

Enabled

### Version

7.14.0 

### Steps to Reproduce

Logcat is a text file. So even without building a customer viewer for it (suggested on the issue below), we could at least show it as text.

* https://github.com/getsentry/sentry/issues/61005

[Example event](https://sentry.sentry.io/issues/4891888819/events/142923af2aaa4c24a69d138424ee5d71/)

### Expected Result

![Image](https://github.com/user-attachments/assets/6379ef5a-ef12-4ef0-9f51-689c31247b3e)

### Actual Result

See the logs without the need to download it.

bricefriha:
> this seems to stem from the symbol collector attaching this manually and not providing a proper mime-type

The `mime-type` we are attaching is `text/logcat` 
```csharp
new ByteAttachmentContent(bytes), "logcat.log", "text/logcat"); 
```
([LogCatAttachmentEventProcessor.cs, line 84](https://github.com/getsentry/sentry-dotnet/blob/5e5dece224da202cf8b7674ae5a2693c3f5cbbfb/src/Sentry/Platforms/Android/LogCatAttachmentEventProcessor.cs))

I think we could just change the `mime-type` to `text/plain` since Sentry's previewer [supports it](https://github.com/getsentry/sentry/blob/d4f5584da02f71f4c2a1a00792b9558c22df61db/static/app/components/events/eventAttachments.tsx#L38-L56).

Would there be any limitation in doing that?

bitsandfoxes:
Fixed by https://github.com/getsentry/sentry-dotnet/pull/3711

PR #3711, opened 2024-10-25
Feat: allow logcat attachment to be previewed in Sentry

bricefriha:
## Context

As issue #3703 reports, Logcats files currently attached to Sentry issues are not previewable.

image

This issue is due to the `mime-type` sent with the file, as it currently has its own type, `text/logcat`. However, Sentry's preview component does not support this type.
This PR aims to tackle just that by changing the mime-type we send from `text/logcat` to `text/plain`.

image

## Caviat
I personally believe that the user should have a more console-like experience when looking at the preview on Sentry. To tackle this, an [issue in Sentry's repo](https://github.com/getsentry/sentry/issues/61005) is currently open. So, the PR would offer more of a temporary fix than a complete solution.
Therefore, when/if `text/logcat` gets support, we could change the mime-type back on our end


src/Sentry.Bindings.Android/Sentry.Bindings.Android.csproj
```diff
@@ -3,7 +3,7 @@
     <TargetFramework>net7.0-android</TargetFramework>
     <!-- BG8605 and BG8606 happen because there's a missing androidx.lifecycle dependency, but we don't need it here.  (The native Android Sentry SDK will use it if it exists.) -->
     <NoWarn>$(NoWarn);BG8605;BG8606</NoWarn>
-    <SentryAndroidSdkVersion>7.16.0</SentryAndroidSdkVersion>
+    <SentryAndroidSdkVersion>7.15.0-alpha.1</SentryAndroidSdkVersion>
```
bricefriha, line 6: using `7.15.0-alpha.1` to change the NDK handler strategy and be able to get the logcat file attached see #3694.

To be changed before emerging to `main`

src/Sentry/Platforms/Android/LogCatAttachmentEventProcessor.cs
```diff
@@ -81,7 +81,7 @@ public SentryEvent Process(SentryEvent @event, SentryHint hint)
             output.Seek(0, SeekOrigin.Begin);
             var bytes = output.ToArray();
 
-            hint.Attachments.Add(new SentryAttachment(AttachmentType.Default, new ByteAttachmentContent(bytes), "logcat.log", "text/logcat"));
+            hint.Attachments.Add(new SentryAttachment(AttachmentType.Default, new ByteAttachmentContent(bytes), "logcat.log", "text/plain"));
```
bruno-garcia, line 84: Is this single line the fix for the preview issue?
Why branch off of another feature?

bricefriha: @Bruno-garcia
>Is this single line the fix for the preview issue?

Yes, that's just this line

>Why branch off of another feature?

Initially I wanted to make a pr just for this line. 
But I was also planning on writing some device tests so the CI can confirm the changes, and without PR #3694 the logcat files are not attached, so the CI would have automatically failed.

That's why we branch this off #3694.

jamescrosswell: @bricefriha I think the other PR addresses a specific situation in which logcat logs aren't sent. More generally, as long as we're not dealing with SEGSEGV exceptions, we should be getting logcat reports come through though, so it should be possible to test this fix without mixing the two PRs right?

bricefriha: @jamescrosswell You're absolutely right; I don't know why, after working on #3461, in my head I thought the logcat files weren't sent at all. 
Let's reset the branch before the first commit of #3694 and add a commit of the `mime-type` change

bricefriha:
I can't find a way to get `mime-type` during device tests.

But it's not like we **need** to test this since the `mime-type` been hardcoded anyways

jamescrosswell:
> I can't find a way to get `mime-type` during device tests.
> 
> But it's not like we **need** to test this since the `mime-type` been hardcoded anyways

@bricefriha the device tests are just unit tests that get run on Android or iOS by our device runner here:
https://github.com/getsentry/sentry-dotnet/blob/a2bdc824e133fee51fafcca6d792b466c456dca6/test/Sentry.Maui.Device.TestApp/Startup.cs#L22-L29

So you don't need to do anything special... just write some unit tests for `LogCatAttachmentEventProcessor.Process`.

test/Sentry.Maui.Tests/SentryMauiLogcatsTests.cs
```diff
@@ -0,0 +1,66 @@
...
+        // Arrange
+        var builder = _fixture.Builder.UseSentry();
+        using var app = builder.Build();
+        SentryHint hint = null;
+        var options = app.Services.GetRequiredService<IOptions<SentryMauiOptions>>().Value;
+        var lcEvtProcessor = new LogCatAttachmentEventProcessor(options.DiagnosticLogger, options.Android.LogCatIntegration, options.Android.LogCatMaxLines)
+        var scope = new Scope(options);
+
+        // Act
+        lcEvtProcessor.Process(Arg.Any<SentryEvent>(), Arg.Do<SentryHint>(h => hint = h)).Returns(new SentryEvent());
```
bricefriha, line 55: @jamescrosswell So here I'm testing the `LogCatAttachmentEventProcessor.Process()` method as you advised me, and making sure the log is attached as a `text/plain`. What do you think? Do you think it's enough, or is there something else you would test?

test/Sentry.Maui.Tests/SentryMauiLogcatsTests.cs
```diff
@@ -0,0 +1,72 @@
...
+        // Act
+        processor.Process(Arg.Any<SentryEvent>(), Arg.Do<SentryHint>(h => hint = h)).Returns(new SentryEvent());
+        options.AddEventProcessor(processor);
+
+        _ = new SentryClient(options).CaptureEvent(new SentryEvent(), scope);
+
+
+        // Assert
+        hint.Should().NotBeNull();
+        hint.Attachments.First().ContentType.Should().Be("text/plain", hint.Attachments.First().ContentType);
```
bricefriha, line 69: @jamescrosswell here. I'm testing  `LogCatAttachmentEventProcessor.Process` and found a way to check if the logcat attachment has the right `mime-type`. I didn't see the need to refactor `Process()` since I can still get `hint` from it this way

test/Sentry.Maui.Tests/SentryMauiLogcatsTests.cs
```diff
@@ -0,0 +1,72 @@
...
+                options.FlushTimeout = TimeSpan.FromSeconds(10);
+            });
+            Builder = builder;
+        }
+    }
+
+    private readonly Fixture _fixture = new();
+
+#if __MOBILE__
+    [SkippableFact]
```
jamescrosswell, line 39:
```suggestion
#if ANDROID
    [Fact]
```

The test is only relevant on ANDROID right? We could simplify this and get rid of a bunch of the other directive checks below...

test/Sentry.Maui.Tests/SentryMauiLogcatsTests.cs
jamescrosswell: Ah nice - I was thinking of a unit tests on the `LogCatAttachmentEventProcessor.Process` method (which would have required some refactoring), but this tests the same thing - so all good!

test/Sentry.Maui.Tests/SentryMauiLogcatsTests.cs
bricefriha: it makes more sense yes, now it might be there is gonna be a bit of a conflict with the other PR but nothing we can't handle

test/Sentry.Maui.Tests/SentryMauiLogcatsTests.cs
```diff
@@ -0,0 +1,72 @@
...
+
+    private readonly Fixture _fixture = new();
+
+#if ANDROID
+    [Fact]
+    public void CaptureException_CheckLogcatType()
+    {
+#if __IOS__
+        Skip.If(true, "Doesn't support logcats");
+#endif
```
jamescrosswell, line 44:
```suggestion
```

Can't be `ANDROID` and `__IOS__`

test/Sentry.Maui.Tests/SentryMauiLogcatsTests.cs
```diff
@@ -0,0 +1,72 @@
...
+        var builder = _fixture.Builder.UseSentry(options =>
+        {
+#if ANDROID
+            options.Android.LogCatIntegration = Android.LogCatIntegrationType.All;
+#endif
```
jamescrosswell, line 49:
```suggestion
            options.Android.LogCatIntegration = Android.LogCatIntegrationType.All;
```

This is covered by the `#if` above.

jamescrosswell (changes requested):

jamescrosswell (commented):
Looks good - thanks @bricefriha !

Just one question - have you actually tested this in an android app to make sure the logs come through OK and can be previewed in Sentry?

bricefriha:
> have you actually tested this in an android app to make sure the logs come through OK and can be previewed in Sentry?

I have tested with the sample we have and it can be previewed in Sentry (as you can see on the screenshot in the description)

jamescrosswell (approved):

PR #3721, opened 2024-10-31
ci: show msbuild performance summary

vaind:


bruno-garcia:
any chance this can show up in the PR as a comment or in the build summary if we open it?

vaind:
> any chance this can show up in the PR as a comment or in the build summary if we open it?

have you seen anything actually useful in there?

I guess we could parse the build.log and add it as a job summary, although I'm not sure how it behaves with the existing summaries coming from .net tests. I understand they would overwrite one-another...

PR #3725, opened 2024-11-02
Filesystem usage analyzer

### LINKED ISSUE #3643: Add analyzer for `FileSystem`

RRode:
Resolves #3643

Implements analyzer that searches for file system usage and suggests IFileSystem usage. Currently checks for usages of File, FileInfo, Directory and DirectoryInfo. No warning is produced, if these are used within IFileSystem implementation.

Analyzer was added to Sentry project and finds usage of file system outside of IFileSystem implementations, which then breaks the build. Usages found in PortablePdbReader.cs, CFunctions.cs and FileDiagnosticLogger.cs.

@bitsandfoxes Is this roughly what you had in mind? 

#skip-changelog


bruno-garcia:
This is really awesome and promising. Thank you!

Is there a way to ignore some places where we it would be safe to call from, for example the CFunction one perhaps will be

RRode:
Is there a reason why you could not use `#pragma warning disable / enable`? IMO, with additional comment, this would make it explicit and well documented why the warning should be ignored. But, if you want, I could add some exceptions to the rule. Just let me know what you need.
Also is it OK that I placed the new projects under src and test? For me the analyzer project feels a bit odd there, since it's not a deliverable. I thought about putting both projects under modules, but that seemed a bit off too, since those are git submodules that get pulled in.
Other thing is, would it make sense to put the analyzer into Directory.Build.props in src, so it is added in all src files? I assume that you want to enforce the warning in those projects as well.

bruno-garcia:
> Is there a reason why you could not use #pragma warning disable / enable? IMO, with additional comment, this would make it explicit and well documented why the warning should be ignored. But, if you want, I could add some exceptions to the rule. Just let me know what you need.

That makes sense, `#pragma` sounds like the right solution

Also is it OK that I placed the new projects under src and test? For me the analyzer project feels a bit odd there, since it's not a deliverable. I thought about putting both projects under modules, but that seemed a bit off too, since those are git submodules that get pulled in.

I also think it's kind of odd, but it's not a big deal. Alternatively it could be a new top level directory: `analyzers` or `build-src` I dunno. I think as you did it's fine. But worth adding a README.md on the project directory to clarify this analyizer is only for consumption from `Sentry` itself, for development of the SDK.

Other thing is, would it make sense to put the analyzer into Directory.Build.props in src, so it is added in all src files? I assume that you want to enforce the warning in those projects as well.

We only really need this on `Sentry` since this is a requirement coming from `Sentry.Unity`. So, technically we'd want that analyizer there too.

We'd include that here too: https://github.com/getsentry/sentry-unity/tree/main/src/Sentry.Unity
Using relative paths to the sentry-dotnet submodule

src/Sentry.Analyzers/Resources.resx
```diff
@@ -0,0 +1,30 @@
...
+        <value>1.3</value>
+    </resheader>
+    <resheader name="reader">
+        <value>System.Resources.ResXResourceReader, System.Windows.Forms, Version=2.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
+    </resheader>
+    <resheader name="writer">
+        <value>System.Resources.ResXResourceWriter, System.Windows.Forms, Version=2.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
+    </resheader>
+  <data name="SN0001Description" xml:space="preserve">
+    <value>Route file system access through IFileSystem wrapper.</value><comment>An optional longer localizable description of the diagnostic.</comment>
```
bruno-garcia, line 22: If it's optional, can we got without them? If not, I suggest taking more details from the ticket that introduced this requirement (tldr: Switch file system requirements)

```suggestion
    <value>Route file system access through IFileSystem wrapper.</value>
```

src/Sentry/Sentry.csproj
```diff
@@ -49,6 +49,10 @@
     <PackageReference Include="System.Reflection.Metadata" Version="5.0.0" />
   </ItemGroup>
 
+  <ItemGroup>
+    <ProjectReference Include="..\Sentry.Analyzers\Sentry.Analyzers.csproj" OutputItemType="Analyzer" ReferenceOutputAssembly="false" />
```
bruno-garcia, line 53: Does this need `PrivateAssets>all` ?

RRode: If I understand the [docs](https://learn.microsoft.com/en-us/nuget/consume-packages/package-references-in-project-files#controlling-dependency-assets) correctly, it could have even been fine, but I added it anyway. I also added private assets in analyzer references.

src/Sentry.Analyzers/Resources.resx
RRode: I completely dropped the resx and hardcoded the warning, since I assume that no warning translations will be needed.

RRode:
> That makes sense, `#pragma` sounds like the right solution

Can you provide some comments or should I just disable the warning in currently offending files?

> I also think it's kind of odd, but it's not a big deal. Alternatively it could be a new top level directory: `analyzers` or `build-src` I dunno. I think as you did it's fine. But worth adding a README.md on the project directory to clarify this analyizer is only for consumption from `Sentry` itself, for development of the SDK.

Then I will keep it simple for now and leave it as it is. I added the README, but was not able to remove the one added via Directory.Build.props to the project, so I hope that is OK. 

> We only really need this on `Sentry` since this is a requirement coming from `Sentry.Unity`. So, technically we'd want that analyizer there too.
>
> We'd include that here too: https://github.com/getsentry/sentry-unity/tree/main/src/Sentry.Unity
Using relative paths to the sentry-dotnet submodule

I guess it makes sense to first merge it here and then add it in the other repo, right?

src/Sentry.Analyzers/Sentry.Analyzers.csproj
```diff
@@ -0,0 +1,43 @@
...
+        <TargetFramework>netstandard2.0</TargetFramework>
+        <IsPackable>false</IsPackable>
+        <Nullable>enable</Nullable>
+        <LangVersion>latest</LangVersion>
+
+        <EnforceExtendedAnalyzerRules>true</EnforceExtendedAnalyzerRules>
+        <IsRoslynComponent>true</IsRoslynComponent>
+
+        <RootNamespace>Sentry.Analyzers</RootNamespace>
+        <AssemblyName>Sentry.Analyzers</AssemblyName>
```
bruno-garcia, line 13: afaik these default to the project name so are redundant 
```suggestion
```

bruno-garcia (commented):
Looks good. 

Doesn't seem to be draft anymore, ready for a final pass and merge?

bruno-garcia:
> Can you provide some comments or should I just disable the warning in currently offending files?

The current ones can be skipped, thanks

> Then I will keep it simple for now and leave it as it is. I added the README, but was not able to remove the one added via Directory.Build.props to the project, so I hope that is OK.

Sounds good

> I guess it makes sense to first merge it here and then add it in the other repo, right?

Definitely, that was just a note for reference, doesn't affect this PR. Can be a follow up if you'd like to contribute further 🙏

src/Sentry.Analyzers/Sentry.Analyzers.csproj
RRode: I removed them

RRode:
> The current ones can be skipped, thanks

I added the pragmas in all the places, but this requires changes to [this submodule](https://github.com/getsentry/Ben.Demystifier/pull/6). If I understand correctly this needs to be merged first to main and then reference in this PR adjusted? I am not used to this submodule workflow, so I hope I am doing it right. 😅
 
> Definitely, that was just a note for reference, doesn't affect this PR. Can be a follow up if you'd like to contribute further 🙏

I will give it a try. 🙂

src/Sentry/Infrastructure/FileDiagnosticLogger.cs
```diff
@@ -30,7 +30,10 @@ public FileDiagnosticLogger(string path, bool alsoWriteToConsole = false)
     public FileDiagnosticLogger(string path, SentryLevel minimalLevel, bool alsoWriteToConsole = false)
         : base(minimalLevel)
     {
+        // Allow direct file system usage
+#pragma warning disable SN0001
```
bruno-garcia, line 34: Ah good point, @bitsandfoxes anything we need to do to make this "Not available on Switch" ? 

Edge case since someoen needs to get out of their way to use this and its' just for debugging, but this analyzier really help us make these calls now.

Not doing anything about this (just ignoring it) is fine by me too but we should add a note to the docs that this thing isn't supported on Switch anyway:

https://docs.sentry.io/platforms/unity/configuration/diagnostic-logger/#FileDiagnosticLogger

src/Sentry/Platforms/Native/CFunctions.cs
```diff
@@ -158,8 +158,11 @@ internal static string GetCacheDirectory(SentryOptions options)
     {
         if (options.CacheDirectoryPath is null)
         {
+            // Allow direct file system usage
```
bruno-garcia, line 161:
```suggestion
            // Allow direct file system usage - Not application for Nintendo Switch
```

bruno-garcia (approved):
I love this! thank you so much

bruno-garcia:
I'll leave it for @jamescrosswell to do the merging as he's the maintainer but LGTM! 🙏

jamescrosswell:
> I added the pragmas in all the places, but this requires changes to [this submodule](https://github.com/getsentry/Ben.Demystifier/pull/6). If I understand correctly this needs to be merged first to main and then reference in this PR adjusted? I am not used to this submodule workflow, so I hope I am doing it right. 😅

Hey @RRode, up until now we've been maintaining the Sentry fork of Ben.Demystifier purely for use in the sentry-dotnet repository so the workflows for changing code in there are a bit obscure and not very contributor friendly (the code used by `sentry-dotnet` isn't in the `main` branch, for exaple). I'll do a bit of head scratching to see if we could make this easier.

Worst case, I can cherry pick the change from your PR and get it in there but I'd rather tidy up the submodule so it's easier to accept contributions. Give me a couple days.

jamescrosswell:
> Worst case, I can cherry pick the change from your PR and get it in there but I'd rather tidy up the submodule so it's easier to accept contributions. Give me a couple days.

Hi @RRode, I ended up just cherry pick your changes and push these up.

jamescrosswell:
> @jamescrosswell I pushed the missing warning disable for AndroidHelpers

Thanks @RRode - apologies in advance if the build breaks on this one. GitHub just changed their macos-14 runners (no longer support XCode 16.0) so we need to bump to macos-15 but that's missing a bunch of stuff we need. Once this is merged, we can pull the changes into this branch and the build should work again:
- https://github.com/getsentry/sentry-dotnet/pull/3754

RRode:
@jamescrosswell no problem. What about this format code check? Is there some doing from my side? If I understand this correctly, git fails to do something with my branch.

jamescrosswell:
> @jamescrosswell no problem. What about this format code check? Is there some doing from my side? If I understand this correctly, git fails to do something with my branch.

Hm, possibly it's getting confused because there are multiple remotes defined for the Ben.Demystifier submodule (getsentry/Ben.Demystifier and benadams/Ben.Demystifier). The benadams (upstream) seems to be dead - it hasn't been updated forever and we can't get in touch with Ben so maybe we just remove that remote. 

I'll look into it once we've fixed the macos builds.

jamescrosswell:
> @jamescrosswell no problem. What about this format code check? Is there some doing from my side? If I understand this correctly, git fails to do something with my branch.

Hm, actually I think it might be an issue with `scripts/commit-formatted-code.sh`:
https://github.com/getsentry/sentry-dotnet/blob/133e1a5fce1c80a43c79505e2c9f534a118ae6b0/scripts/commit-formatted-code.sh#L13

In your case, the PR is coming from a fork so it's not finding the `filesystem-usage-analyzer` branch (it needs to checkout `RRode:filesystem-usage-analyzer`).

I'm in GMT+12 so it's kind of late on a Friday night here, but I'll look into it on the weekend or early next week.

jamescrosswell:
Finally got it over the line. Thanks for your patience and huge thank you for the PR @RRode !

RRode:
Happy to help! 🙂

Issue #3724, opened 2024-11-01
Exception on shutdown when using Sentry & OpenTelemetry


petedishman:
### Package

Other

### .NET Flavor

.NET Core

### .NET Version

8.0.0

### OS

Any (not platform specific)

### SDK Version

4.12.1

### Steps to Reproduce

I have a minimal repo available at https://github.com/petedishman/SentryShutdownBug

This is a minimal aspnetcore app configured with Sentry & OpenTelemetry together

Can repro the issue with:
1. clone the above repo
2. configure a Sentry DSN in `program.cs`
3. Run the app with `dotnet run`
4. Browse to `http://localhost:5005`
5. Close the app with `CTRL+C`
6. exception should be logged

It appears to be dependent on the `TracesSampleRate` setting and this needs to be discarding traces for the error to occur, otherwise if it's set to `1.0` it will exit without error.

### Expected Result

Application should shutdown without an exception being thrown

### Actual Result

This exception is logged, then app continues to shutdown

```
fail: Sentry.ISentryClient[0]
      Error while sending final client report (event ID: '(null)').
      System.InvalidCastException: Unable to cast object of type 'Sentry.Internal.NoOpTransaction' to type 'Sentry.TransactionTracer'.
         at Sentry.OpenTelemetry.SentrySpanProcessor.CreateRootSpan(Activity data)
         at Sentry.OpenTelemetry.SentrySpanProcessor.OnStart(Activity data)
         at OpenTelemetry.CompositeProcessor`1.OnStart(T data)
         at OpenTelemetry.Trace.TracerProviderSdk.<.ctor>b__12_1(Activity activity)
         at System.Diagnostics.SynchronizedList`1.EnumWithAction(Action`2 action, Object arg)
         at System.Diagnostics.Activity.Start()
         at System.Net.Http.DiagnosticsHandler.SendAsyncCore(HttpRequestMessage request, Boolean async, CancellationToken cancellationToken)
         at System.Net.Http.RedirectHandler.SendAsync(HttpRequestMessage request, Boolean async, CancellationToken cancellationToken)
         at System.Net.Http.DecompressionHandler.SendAsync(HttpRequestMessage request, Boolean async, CancellationToken cancellationToken)
         at Sentry.Internal.Http.GzipBufferedRequestBodyHandler.SendAsync(HttpRequestMessage request, CancellationToken cancellationToken)
[...]

PR #3726, opened 2024-11-04
Fixed: processing spans after SDK is closed (OTel and ASP.NET Core)

jamescrosswell:
Resolves https://github.com/getsentry/sentry-dotnet/issues/3724

Fixes and issue when using ASP.NET Core and OpenTelemetry instrumentation, where the SDK might be closed before the last OpenTelemetry spans get fully processed by the `SentrySpanProcessor`.

## Problem

When the AppBuilderExtensions initialise the SDK a lifetime callback is also registered to close the SDK when the app shuts down:
https://github.com/getsentry/sentry-dotnet/blob/e382fb080102cf654f86566cd53816b208746e62/src/Sentry.AspNetCore/ApplicationBuilderExtensions.cs#L45

However this can be called before all telemetry events are received/processed by the `SentrySpanProcessor`, which was then trying to create new spans for that telemetry using a disabled hub.

## Solution

The OnEnd and OnStart events for the SentrySpanProcessor now check to see whether the hub is still active before doing any work.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
+
+### Fixes
+
+- When using OTel and ASP.NET Core the SDK could try to process OTel spans after the SDK had been closed ([#3726](https://github.com/getsentry/sentry-dotnet/pull/3726))
+
 ## 4.12.2
 
 ### Features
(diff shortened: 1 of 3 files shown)


bricefriha (approved):
looks good to me

PR #3727, opened 2024-11-05
Use pre-built version of sentry-cocoa SDK

jamescrosswell:
This is to address difficulties we've been having bumping the sentry-cocoa version in this repo (e.g [here](https://github.com/getsentry/sentry-dotnet/pull/3716#issuecomment-2445376208)). 

In this PR, instead of including the sentry-cocoa repo as a git submodule and building this ourselves, we're downloading the most recent release (similar to what is being done [in the Unity repo](https://github.com/getsentry/sentry-unity/blob/8de84e71c6439d43ece603880a95d4daa15188a5/Directory.Build.targets#L162-L169)).


src/Sentry.Bindings.Cocoa/ApiDefinitions.cs
```diff
@@ -1635,6 +1658,11 @@ interface SentrySDK
     [Export ("isEnabled")]
     bool IsEnabled { get; }
 
+    // @property (readonly, nonatomic, class) SentryReplayApi * _Nonnull replay;
+    [Static]
+    [Export ("replay")]
+    SentryReplayApi Replay { get; }
```
bruno-garcia, line 1664: woooo!! ❤️

src/Sentry.Bindings.Cocoa/Sentry.Bindings.Cocoa.csproj
```diff
@@ -44,18 +45,53 @@
...
-    Inputs="..\..\.git\modules\modules\sentry-cocoa\HEAD" Outputs="..\..\modules\sentry-cocoa\Carthage\.built-from-sha">
-    <MSBuild Projects="$(MSBuildProjectFile)" Targets="_InnerBuildSentryCocoaSDK" Properties="TargetFramework=once" />
+  <!-- Downloads and sets up the Cocoa SDK: dotnet msbuild /t:setupCocoaSDK src/Sentry.Bindings.Cocoa/Sentry.Bindings.Cocoa.csproj -->
+  <Target Name="SetupCocoaSDK"
+          Condition="$([MSBuild]::IsOSPlatform('OSX')) And !Exists('$(SentryCocoaCache)Sentry-Dynamic.xcframework')"
+          BeforeTargets="BeforeBuild">
+
+    <PropertyGroup>
+      <PropertiesContent>$([System.IO.File]::ReadAllText("../../modules/sentry-cocoa.properties"))</PropertiesContent>
+      <CocoaVersion>$([System.Text.RegularExpressions.Regex]::Match($(PropertiesContent), 'version\s*=\s*(\d+\.\d+\.\d+)').Groups[1].Value)</CocoaVersion>
```
bruno-garcia, line 55: this regex will match beta releases but extract only the `n.n.n` and not the `-xyz`

bruno-garcia (changes requested):

src/Sentry.Bindings.Cocoa/Sentry.Bindings.Cocoa.csproj
jamescrosswell: Ah, nice catch. @bitsandfoxes you'll want to [fix that in the Unity repo](https://github.com/getsentry/sentry-unity/blob/8de84e71c6439d43ece603880a95d4daa15188a5/Directory.Build.targets#L149) as well.

I've gone with dead simple - which is to match any sequence of non-whitespace characters:
https://github.com/getsentry/sentry-dotnet/blob/bbd1f64417731aab4d780afda0902066a0129a11/src/Sentry.Bindings.Cocoa/Sentry.Bindings.Cocoa.csproj#L55

bricefriha (approved):
As far as I'm concerned, 
looks fine!

src/Sentry.Bindings.Cocoa/Sentry.Bindings.Cocoa.csproj
jamescrosswell: @bruno-garcia the issue you found in the regex has been addressed. Any other "changes requested" or can we merge?

bruno-garcia (approved):

bruno-garcia:
I was a big fan of the flexibility of modifying the underlying library, pushing a branch and having a feature cut through that way but I realize it's a lot of overhead for this very rarely used benefit.

PR #3738, opened 2024-11-08
improve windows config for Cocoa development

bricefriha:
In the same veins as PR #3695, since I'm on windows there is a few things I need to adjust when contributing to issues related to cocoa. And since I'm using Visual Studio, these configs are so needed that I have keep them in git stash.

Then, I thought maybe it would be useful to have them on the repo to reduce the friction for new contributors who are on Windows. 

### Please not
Like for the MAUI sample config changes (#3695), none these configs don't require a mac to be paired.

#skip-changelog

### DIFF
--- a/src/Sentry.Bindings.Cocoa/Sentry.Bindings.Cocoa.csproj
+++ b/src/Sentry.Bindings.Cocoa/Sentry.Bindings.Cocoa.csproj
@@ -13,7 +13,7 @@
   </PropertyGroup>
 
   <!-- Build empty assemblies when not on macOS, to pass the solution build. -->
-  <ItemGroup Condition="!$([MSBuild]::IsOSPlatform('OSX'))">
+  <ItemGroup Condition="!($([MSBuild]::IsOSPlatform('OSX')) Or $([MSBuild]::IsOSPlatform('windows')))">
     <Compile Remove="*" />
     <Using Remove="*" />
   </ItemGroup>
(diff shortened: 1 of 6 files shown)


src/Sentry.Bindings.Cocoa/Sentry.Bindings.Cocoa.csproj
```diff
@@ -13,7 +13,7 @@
   </PropertyGroup>
 
   <!-- Build empty assemblies when not on macOS, to pass the solution build. -->
-  <ItemGroup Condition="!$([MSBuild]::IsOSPlatform('OSX'))">
+  <ItemGroup Condition="!($([MSBuild]::IsOSPlatform('OSX')) Or $([MSBuild]::IsOSPlatform('windows')))">
```
jamescrosswell, line 16: I think this is going to be problematic in CI. With these files included, it's going to attempt to build the source for Sentry.Bindings.Cocoa on the Windows runner... and that will fail.

Rather than checking for the operating system as a second parameter, perhaps you could check for the existence of an environment variable (`EnableRemoteMacBuilds` or something) so that anyone developing on Windows with a remote Mac available could turn this on if they want, but it won't mess with the CI builds?

bricefriha: That's a good idea!

bricefriha:
@jamescrosswell I'm trying to figure out a way to have this included only if the windows machine is connected to a pair. This way we can even have it run a build of Sentry.Bindings.Cocoa. so essentially, if a mac is paired the machine, it would be the same experience as setting up the environment on a mac. **(issue is that the windows instance that the CI run won't be connected to a mac)**

 it would be technically feasible. the question is would it be worth the effort? 
because for sure it would make the project more accessible.

What do you reckon?

jamescrosswell:
> it would be technically feasible. the question is would it be worth the effort? because for sure it would make the project more accessible.
> 
> What do you reckon?

Depends how much effort it is really.

It doesn't affect Bruno, Stefan or me since we're mainly developing on mac OS. We haven't had any customers or other contributors ask about this... so it's mainly for your own productivity at this point. If you can do it fairly quickly/easily without breaking CI then it might be worth it.

PR #3742, opened 2024-11-11
Only regenerate the cocoa bindings when they change

jamescrosswell:
Currently the `ApiDefinitions.cs` and `StructsAndEnums.cs` files are being regenerated every time the `src/Sentry.Bindings.Cocoa/Sentry.Bindings.Cocoa.csproj` project is built... changing the timestamp on these and showing as "changes" that aren't really changes in git. 

This PR uses the [MS Build Incremental Build](https://learn.microsoft.com/en-us/visualstudio/msbuild/how-to-build-incrementally) features to ensure these files only get regenerated if/when the CocoaSDK version changes. 

#skip-changelog

### DIFF
--- a/src/Sentry.Bindings.Cocoa/Sentry.Bindings.Cocoa.csproj
+++ b/src/Sentry.Bindings.Cocoa/Sentry.Bindings.Cocoa.csproj
@@ -91,6 +91,7 @@
   <!-- Generate bindings -->
   <Target Name="_GenerateSentryCocoaBindings" AfterTargets="SetupCocoaSDK"
           Condition="$([MSBuild]::IsOSPlatform('OSX')) and Exists('$(SentryCocoaFrameworkHeaders)')"
+          Inputs="../../modules/sentry-cocoa.properties"
           Outputs="ApiDefinitions.cs;StructsAndEnums.cs">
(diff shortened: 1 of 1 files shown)


src/Sentry.Bindings.Cocoa/Sentry.Bindings.Cocoa.csproj
```diff
@@ -91,6 +91,7 @@
   <!-- Generate bindings -->
   <Target Name="_GenerateSentryCocoaBindings" AfterTargets="SetupCocoaSDK"
           Condition="$([MSBuild]::IsOSPlatform('OSX')) and Exists('$(SentryCocoaFrameworkHeaders)')"
+          Inputs="../../modules/sentry-cocoa.properties"
           Outputs="ApiDefinitions.cs;StructsAndEnums.cs">
     <MSBuild Projects="$(MSBuildProjectFile)" Targets="_InnerGenerateSentryCocoaBindings" Properties="TargetFramework=once" />
```
vaind, line 96: The target launches `generate-cocoa-bindings.ps1` so we should have that as another input - when the script changes it's quite likely we need to regenerate the code.

Issue #3698, opened 2024-10-21
Trimming support for NLog (Bump to NLog 5.2.2)


jamescrosswell:
NLog 5.2.2 added trimming support. With growing adoption of MAUI, more and more users are trying to trim their apps. 

We need to:
- Bump to 5.2.2
- Provide enough information about the custom configuration type used with NLog that it either doesn't try to load certain properties or the linker preserves any types required to load those properties. 

Worth investigating @snakefoot 's comment about the complexity of the `SentryNLogOptions` class. If people are only using the **_NLog_** integration then ideally they'd be putting all of their Sentry config in one place, but maybe worth challenging that as it creates additional complexity (like this) elsewhere.

 _Originally posted by @snakefoot in [#3680](https://github.com/getsentry/sentry-dotnet/issues/3680#issuecomment-2427262823)_

> Think the [Sentry.NLog](https://github.com/getsentry/sentry-dotnet/blob/main/src/Sentry.NLog/Sentry.NLog.csproj) project should update to NLog v5.2.2 (or newer). 
> - See also: https://nlog-project.org/2023/05/30/nlog-5-2-trim-warnings.html
> - 
> 
> And then you should provide a registration-method, that marks the relevant types in Sentry.NLog to not be trimmed (This already happens for all types inside NLog):
> ```c#
> NLog.LogManager.Setup().SetupExtensions(ext => {
> 	ext.RegisterTarget<Sentry.NLog.SentryTarget>();
> 	ext.RegisterType<Sentry.NLog.SentryNLogOptions>();
> });
> ```
> 
> You should probably avoid mixing options that are not intended to be assigned from NLog-configuration-file. Maybe SentryNLogOptions  should not inherit from SentryOptions. But should instead just expose the properties that are intended to be assigned from NLog-configuration-file. Then you could have a property in SentryNLogOptions marked like this:
> ```c#
> [NLogConfigurationItem]
> public class SentryNLogOptions 
> {
>     [NLogConfigurationIgnoreProperty]
>     public SentryOptions SentryOptions { get; } = new SentryOptions();
> 
>     public int ShutdownTimeoutSeconds
>     {
>         get => (int)SentryOptions.ShutdownTimeout.TotalSeconds;
>         set => SentryOptions.ShutdownTimeout = TimeSpan.FromSeconds(value);
>     }
> }
> ```
[...]

PR #3743, opened 2024-11-11
fix: `Release` Builds with Trimming enabled crashing on launch with NLog Integration

### LINKED ISSUE #3680: Release Builds Crashing on Launch when using NLog Integration and FailedRequestStatusCodes Options
### Package

Sentry

### .NET Flavor

.NET

### .NET Version

8.0.0

### OS

Android

### SDK Version

4.12.1

_No response_

### Steps to Reproduce

Context:
I have been having issues with my Maui app crashing on launch when building for release. Eventually I was able to trace the issue to a few configuration options that reproduce the issue in a simple sample project. These include setting up Sentry and NLog and then running a release build on Android or iOS. If trimming is enabled for the project in the Android project settings (or linking for iOS) on debug builds then the issue seems to reproduce in debug builds as well and I get this error: System.ArgumentException: 'Arg_GetMethNotFnd' when calling LoadConfiguration on the NLog ISetupBuilder object.

Interestingly one of the configuration options I have to set to reproduce the error is setting sentryOptions.FailedRequestStatusCodes to a non-null value. Both an empty list or specified range seems to "work" (cause the crash). It doesn't seem to mater if SentryOptions.CaptureFailedRequests is enabled or not, so I have not included that option in my repro steps.

As a work around, the following snippet can be added to the project file to prevent trimming of the Sentry assembly.
`    <ItemGroup>
        <TrimmerRootAssembly Include="Sentry" RootMode="library"/>
    </ItemGroup>`

**Repro in a new/existing Maui project**
1. Add nuget packages Sentry.Maui 4.12.1 and Sentry.NLog 4.12.1
2. Add nuget packages NLog 5.3.4, NLog.Extensions.Logging 5.3.14, NLog.Targets.MauiLog 8.0.0
3. Add the the following configuration to MauiProgram.CreateMauiApp: 
```
.UseSentry(options =>
                {
                    options.Dsn = "<Your DSN here>";
                    options.Release = "releaseString";
                    options.Distribution = "distributionString";
                    options.Environment = "environmentString";
                    options.Debug = true;
                    options.FailedRequestStatusCodes = new List<HttpStatusCodeRange>() {};
                })
                .Logging.ClearProviders().AddNLog()
```
4. Add this code to setup NLog after setting up the app builder but before calling builder.Build():
```
            ISetupBuilder setup = LogManager.Setup();
            setup.RegisterMauiLog();
            setup.LoadConfiguration(Configure);
```
5. Add this configure method inside of MauiProgram:
```
        private static void Configure(ISetupLoadConfigurationBuilder configBuilder)
        {
            configBuilder.Configuration.AddSentry(options =>
            {
                options.InitializeSdk = false;
                options.MinimumBreadcrumbLevel = NLog.LogLevel.Info;
                options.MinimumEventLevel = NLog.LogLevel.Error;
            });
        }
```
6. Deploy a release build, or run a debug build with trimming (android) or linking (iOS) enabled.
7. Launch the app.

**Repro with Sample Project**
1. Download reproducing sample project: https://github.com/mgr44/Maui-Sentry-TrimmingSample
2. Add a valid DSN to MauiProgram.cs line 22
3. Deploy to an Android device with debug or release build (this project has trimming enabled for android debug builds)
4. Launch the App

### Expected Result

The app launches normally and Sentry and NLog are functional.

### Actual Result

The app crashes during the launch

bricefriha:
This PR aims to fix #3680, an issue that causes a crash when using NLog with FailedRequestStatusCodes options in a Maui app with Trimming enabled in `Release`.

## Note 
we needed to update Nlog to 5.2 to be able to use `RegisterType` for this fix

worth mentioning: https://github.com/getsentry/sentry-dotnet/issues/3698


bricefriha:
@jamescrosswell About #3698, do you think my PR would be problematic for other platforms? 
I mean, in a way that would make it safer to simplify `SentryNLogOptions`'s complexity first. 

Or can we go ahead with this change for now?

src/Sentry.NLog/Sentry.NLog.csproj
```diff
@@ -12,7 +12,7 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="NLog" Version="5.0.5" />
+    <PackageReference Include="NLog" Version="5.2.4" />
```
bricefriha, line 15: Need 5.2 for `RegisterType` 

worth mentioning: #3698

jamescrosswell:
> @jamescrosswell About #3698, do you think my PR would be problematic for other platforms? I mean, in a way that would make it safer to simplify `SentryNLogOptions`'s complexity first.
> 
> Or can we go ahead with this change for now?

Not sure I follow. This PR is to address #3698 right? I can't see anything in the PR (yet) that is platform specific.

Have you had a chance to test this out yet, to make sure it resolves the issue? It's another one that's hard to test in unit tests due to the need to enable trimming.

bricefriha:
@jamescrosswell Sorry, about that.

The PR is for #3680, but you mentioned in #3698 that we needed to bump the NLog and I did this here as part of the fix. 
#3698 is about changing the complexity of the options, so I was just wondering if you think it would be better to wait for this to be trackle before merging the PR?

> Have you had a chance to test this out yet, to make sure it resolves the issue?

Yes it's all good, I'm just waiting for your input to mark this PR as reviewable

jamescrosswell:
> Yes it's all good, I'm just waiting for your input to mark this PR as reviewable

If you're waiting for input on this PR, you should mark it ready for review... or is there another discussion I'm neglecting somewhere?

jamescrosswell:
> #3698 is about changing the complexity of the options, so I was just wondering if you think it would be better to wait for this to be trackle before merging the PR?

Ah I see, no I don't think we need to change the way the options are structured if this PR resolves the original issue. Changing the options would be quite complex - definitely couldn't be done in a point release (would have to happen in a major version bump) so best to keep separate from this PR.

bricefriha:
> so best to keep separate from this PR.

Sounds good. Well, there you have it! 😁

CHANGELOG.md
```diff
@@ -4,7 +4,8 @@
 
 ### Fixes
 
-- Fixed ArgumentNullException in FormRequestPayloadExtractor when handling invalid form data on ASP.NET ([#3734](https://github.com/getsentry/sentry-dotnet/pull/3734))
+- ArgumentNullException in FormRequestPayloadExtractor when handling invalid form data on ASP.NET ([#3734](https://github.com/getsentry/sentry-dotnet/pull/3734))
+- Crash when using NLog with FailedRequestStatusCodes options in a Maui app with Trimming enabled ([#3743](https://github.com/getsentry/sentry-dotnet/pull/3743))
```
bruno-garcia, line 8: Can you add a test that repros this crash, so we make sure this PR fixes it?

jamescrosswell: > Can you add a test that repros this crash, so we make sure this PR fixes it?

@bruno-garcia there's no easy way to test this. Enabling trimming breaks NSubstitute on the mobile targets... so the only way to test this would be an integration test (e.g. write a powershell script to create and compile an app that demonstrates this error). It seemed like more effort than it was worth so I told @bricefriha not to bother.

jamescrosswell (approved):
Looks great. Thanks @bricefriha !

CHANGELOG.md
```diff
@@ -26,7 +26,8 @@
...
```
bruno-garcia, line 30: MAUI is all caps

PR #3699, opened 2024-10-22
Add net9.0 support

jamescrosswell:
Fixes https://github.com/getsentry/sentry-dotnet/issues/3484#issuecomment-2254925496

## Notes for Reviewers

Mostly just a version bump, but with various domino effects:
- We started running out of space on our runners, so includes various workarounds to free up disk space on ubuntu and macos
- The analysers in net9.0 raised various new vulnerability warnings that had to be dealt with either by bumping dependencies or ignoring warnings
- Had to [work around some contention issues with LocalDb](https://github.com/getsentry/sentry-dotnet/blob/3da79c23a70f04f989b70a7bd66a333ac1471186/test/Sentry.DiagnosticSource.IntegrationTests/LocalDbFixture.cs#L7-L20)

### Device Tests
- When targeting net8.0-android, [trimming has to be disabled](https://github.com/getsentry/sentry-dotnet/blob/ec2335e5d34a7c6e93e991e0bc02d273dcc777bc/scripts/device-test.ps1#L62-L63). We had it enabled implicitly by setting `IsAotCompatible = true` for net8.0+ when `FrameworkSupportsAot == true`. That property has been renamed to `EnableAot` and we now override this and set it to `false` when running the android device tests. 
- To get the iOS device tests running on arm, an [x64 runtime identifier had to be used](https://github.com/getsentry/sentry-dotnet/blob/196d00d73e46b7a93ef9b9b901cf9a9192c6cab6/test/Sentry.Maui.Device.TestApp/Sentry.Maui.Device.TestApp.csproj#L59-L60)
- To get the iOS device tests to run against net8.0-ios, [the interpreter has to be enabled explicitly](https://github.com/getsentry/sentry-dotnet/blob/196d00d73e46b7a93ef9b9b901cf9a9192c6cab6/test/Sentry.Maui.Device.TestApp/Sentry.Maui.Device.TestApp.csproj#L42-L45) (presumably the default for this was changed between net7.0-ios and net8.0-ios
- Disabled some flaky tests

PR #3748, opened 2024-11-12
MAUI Sample Android - allow for Debug on physical devices

bricefriha:
Added Android runtime identifiers in the Maui Sample config to allow debugging from a physical Android device.
When debugging, it's often more handy to use a physical device over an emulator. Until now, it was only possible to do it with an iOS device. 
It's only logical to add this for Android as well.

#skip-changelog

### DIFF
--- a/samples/Sentry.Samples.Maui/Sentry.Samples.Maui.csproj
+++ b/samples/Sentry.Samples.Maui/Sentry.Samples.Maui.csproj
@@ -7,7 +7,6 @@
       On Windows, we'll also build for Windows 10.
     -->
     <TargetFrameworks>net8.0-android</TargetFrameworks>
-    <RuntimeIdentifiers Condition="$(TargetFramework.Contains('-android'))">android-arm;android-arm64;android-x86;android-x64</RuntimeIdentifiers>
     <TargetFrameworks Condition="$([MSBuild]::IsOSPlatform('windows'))">$(TargetFrameworks);net8.0-windows10.0.19041.0;net8.0-ios;net8.0-maccatalyst</TargetFrameworks>
     <TargetFrameworks Condition="$([MSBuild]::IsOSPlatform('OSX'))">$(TargetFrameworks);net8.0-ios;net8.0-maccatalyst</TargetFrameworks>
     <OutputType>Exe</OutputType>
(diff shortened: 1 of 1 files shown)


samples/Sentry.Samples.Maui/Sentry.Samples.Maui.csproj
```diff
@@ -7,6 +7,7 @@
       On Windows, we'll also build for Windows 10.
     -->
     <TargetFrameworks>net8.0-android</TargetFrameworks>
+    <RuntimeIdentifiers Condition="$(TargetFramework.Contains('-android'))">android-arm;android-arm64;android-x86;android-x64</RuntimeIdentifiers>
```
jamescrosswell, line 10: Isn't that what this group of properties is doing?

https://github.com/getsentry/sentry-dotnet/blob/ed63749b2a7e6404aaed19ae12810863016a9926/samples/Sentry.Samples.Maui/Sentry.Samples.Maui.csproj#L65-L72

bricefriha: Yes, my bad I didn't notice the line ` <RuntimeIdentifier Condition="'$(TargetPlatformIdentifier)' == 'android'      And '$(OSArchitecture)' == 'Arm64'">android-arm64</RuntimeIdentifier> ` for some reasons

However it doesn't work at all on my side

jamescrosswell: Doesn't work?

bricefriha: not on windows at least

jamescrosswell: Any context? Like error messages etc.?

Might also be worth trying with the version-5.0.0 branch since there are quite a few changes ([to support net9.0](https://github.com/getsentry/sentry-dotnet/pull/3699)) in that branch.

bricefriha: >Any context? Like error messages etc.?

I'm getting ADB0020: Mono.AndroidTools.IncompatibleCpuAbiException: The package does not support the CPU architecture of this device.

>Might also be worth trying with the version-5.0.0 branch since there are quite a few changes (https://github.com/getsentry/sentry-dotnet/pull/3699) in that branch.

I'm gonna try that

jamescrosswell: The only difference between what you're trying to do in this PR and what already happens in the Samples project is you're adding `android-arm` and `android-x86`... so possibly just need to tweak the existing code block to cater to whichever of those allows you to run the sample of the particular physical you're trying to deploy to. 

Would be good to keep this in the same styles as all the others.

samples/Sentry.Samples.Maui/Sentry.Samples.Maui.csproj
```diff
@@ -67,7 +67,7 @@
     <RuntimeIdentifier Condition="'$(TargetPlatformIdentifier)' == 'ios'          And '$(OSArchitecture)' == 'Arm64' And '$(_IsPublishing)' != 'true'">iossimulator-arm64</RuntimeIdentifier>
     <RuntimeIdentifier Condition="'$(TargetPlatformIdentifier)' == 'maccatalyst'  And '$(OSArchitecture)' == 'Arm64'">maccatalyst-arm64</RuntimeIdentifier>
 
-    <RuntimeIdentifier Condition="'$(TargetPlatformIdentifier)' == 'android'      And '$(OSArchitecture)' == 'x64'">android-x64</RuntimeIdentifier>
+    <RuntimeIdentifiers Condition="'$(TargetPlatformIdentifier)' == 'android'     And '$(OSArchitecture)' == 'x64'">android-arm;android-arm64;android-x86;android-x64</RuntimeIdentifiers>
```
jamescrosswell, line 70: @bricefriha why can we not do this like all the other runtimeidentirier elements here? So why can't we have a separate line for x64 and x86? And why is the `android-arm` identifier required?

bricefriha: >  So why can't we have a separate line for x64 and x86?

It seems to have been a mistake. thanks for pointing this out

> why is the android-arm identifier required?

To be fair, we could remove it; it's more for backward compatibility. I don't think there are many people who would debug using a 32-bit device, so we could remove this

jamescrosswell: I see, yeah let's just add what we need to keep it as simple as possible. We can add it if/when it's ever needed in the future.

bricefriha: Sounds good

PR #3753, opened 2024-11-13
Upgrade to .NET 9.0.100

jamescrosswell:
#skip-changelog

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -47,8 +47,8 @@
   -->
   <PropertyGroup>
     <TargetPlatformIdentifier>$([MSBuild]::GetTargetPlatformIdentifier('$(TargetFramework)'))</TargetPlatformIdentifier>
-    <SupportedOSPlatformVersion Condition="'$(TargetPlatformIdentifier)' == 'ios'">11.0</SupportedOSPlatformVersion>
-    <SupportedOSPlatformVersion Condition="'$(TargetPlatformIdentifier)' == 'maccatalyst'">13.1</SupportedOSPlatformVersion>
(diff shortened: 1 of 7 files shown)


Directory.Build.props
```diff
@@ -47,8 +47,8 @@
   -->
   <PropertyGroup>
     <TargetPlatformIdentifier>$([MSBuild]::GetTargetPlatformIdentifier('$(TargetFramework)'))</TargetPlatformIdentifier>
-    <SupportedOSPlatformVersion Condition="'$(TargetPlatformIdentifier)' == 'ios'">11.0</SupportedOSPlatformVersion>
-    <SupportedOSPlatformVersion Condition="'$(TargetPlatformIdentifier)' == 'maccatalyst'">13.1</SupportedOSPlatformVersion>
+    <SupportedOSPlatformVersion Condition="'$(TargetPlatformIdentifier)' == 'ios'">12.2</SupportedOSPlatformVersion>
+    <SupportedOSPlatformVersion Condition="'$(TargetPlatformIdentifier)' == 'maccatalyst'">15.0</SupportedOSPlatformVersion>
```
bitsandfoxes, line 51: Are those coming with/from the .NET 9 release?

bitsandfoxes (approved):

Directory.Build.props
jamescrosswell: They're the minimal versions yes. We can't build unless we bump these as there are targets files that ship with net9.0 that result in build errors if we specify earlier versions.

bruno-garcia (approved):

Issue #3545, opened 2024-08-20
Add managed attachments to native Android events


TimBurik:
### Package

Sentry

### .NET Flavor

.NET

### .NET Version

8.0.303

### OS

Android

### SDK Version

4.10.1

### Steps to Reproduce

1. Create an Android app and install Sentry NuGet package;
2. Configure Sentry so that reported events should have attachments (both logcat and custom attachments are affected by the issue):
```
SentrySdk.Init(options =>
{
    options.Dsn = "...";
    options.Android.LogCatIntegration = LogCatIntegrationType.All;
});
SentrySdk.ConfigureScope(scope =>
{
    scope.AddAttachment("path/to/existing/file.txt", AttachmentType.Default, MediaTypeNames.Text.Plain);
});
```
3. Generate crashes using `SentrySdk.CauseCrash()` with different parameters

### Expected Result

All reported events should have `logcat.log` attachment with logcat logs and custom `file.txt` file attachment.

### Actual Result

- Crash generated with parameter `CrashType.Native` is reported with mechanism `signalhandler` and contains no attachments;
- Crash generated with parameter `CrashType.JavaBackgroundThread` is reported twice: one with mechanism `UncaughtExceptionHandler` and contains no attachments, and another with mechanism `AppDomain.UnhandledException` and contains both attachments;
- All other crash types (`CrashType.Managed`, `CrashType.ManagedBackgroundThread`, `CrashType.Java`) are reported with mechanism `AppDomain.UnhandledException` and contain both attachments.

This issue might be related to [#3461](https://github.com/getsentry/sentry-dotnet/issues/3461)

Reproduction sample: [SampleProject.zip](https://github.com/user-attachments/files/16675998/SampleProject.zip)

TimBurik:
The issue is also reproduced when `ApplicationNotResponding` is reported, no files are attached despite both logcat and custom attachments are configured.

jamescrosswell:
> Crash generated with parameter CrashType.Native is reported with mechanism signalhandler and contains no attachments;

I think this is similar to:
- https://github.com/getsentry/sentry-dotnet/issues/3549

In both cases, Native crash reporting essentially skips all the .NET SDK processing.

jamescrosswell:
Possibly relates to this (so maybe something that hasn't yet been implemented):
[...]

PR #3756, opened 2024-11-13
fix: prevent event duplication when `JavaBackgroundThread` crashes occur 

bricefriha:
Reported on #3545 , this bug yielded two events when `JavaBackgroundThread` crashes occurred: one `UnhandledException` with no Logcat attachments and no scope attachments and one `AppDomain.UnhandledException` with all expected attachments.

This change prevents the `UnhandledException` event to popup. 
The reason why we are choosing the event with `UnhandledException` mechanism over `AppDomain.UnhandledException` is for consistency purposes, as other Java crashes send a single `AppDomain.UnhandledException` event.


src/Sentry/Internal/Extensions/JsonExtensions.cs
```diff
@@ -156,7 +156,21 @@ public static void Deconstruct(this JsonProperty jsonProperty, out string name,
...
+                    result[name] = val.ToString();
+                    break;
+                case JsonValueKind.True:
+                case JsonValueKind.False:
+                    result[name] = value.ValueKind.ToString();
+                    break;
+                default:
             result[name] = value.GetString();
+                    break;
+            }
```
bricefriha, line 173: I changed these because, while not directly linked to the issue at hand: as I was calling `SentrySdk.CauseCrash(CrashType.JavaBackgroundThread);` many times, I noticed that a `The requested operation requires an element of type 'String', but the target element has type 'Number'.' when doing jsonElement.GetString` many times. 

I strongly feel like this belongs in this PR because this fix could provoke it more frequently

jamescrosswell: @bricefriha where does the reference to `GetStringDictionaryOrNull` come from when you're calling` `SentrySdk.CauseCrash`?

It seems odd to to an implicit conversion from into to string in `GetStringDictionaryOrNull` (but only for that type, to make a very specific scenario work). I'm wondering if it's more appropriate to replace that call with `GetDictionaryOrNull` and then make an explicit conversion in the calling code.

bricefriha:
Since it's on java crashes, I couldn't find a way to add tests

src/Sentry/Platforms/Android/SentrySdk.cs
```diff
@@ -101,6 +101,11 @@ private static void InitSentryAndroidSdk(SentryOptions options)
             {
                 o.BeforeSend = new BeforeSendCallback(beforeSend, options, o);
             }
+            else
+            {
+                // Prevent JavaBackgroundThread crashes from creating duplicated event: one UnhandledException one AppDomain.UnhandledException
+                o.BeforeSend = new BeforeSendCallback(((_, _) => null), options, o);
```
jamescrosswell, line 107: Doesn't this just tell the Android SDK to discard all exceptions? Are there no circumstances in which we want the Android SDK to catch exceptions?

bricefriha:
> how did you test it manually?

I added an attachment at scope configuration and generated a `JavaBackgroundThread` crash and checked Sentry if we have only one event and all the attachments 

I was trying to make a device test, but that would crash the entire test, since it's an unhandled exception

CHANGELOG.md
```diff
@@ -26,6 +26,7 @@
 
 ### Fixes
 
+- prevent event duplication and missing attachements when `JavaBackgroundThread` crashes occur ([#3756](https://github.com/getsentry/sentry-dotnet/pull/3756))
```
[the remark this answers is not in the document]
bricefriha: No, the issue we are fixing is #3545, which reported the missing attachments.

Technically, we are not directly fixing the missing attachments. There was one event with the attachments and one without. All we are doing is preventing the latter from occurring.

Should we only mention the event duplication in the change logs?

src/Sentry/Platforms/Android/SentrySdk.cs
bricefriha: > Are there no circumstances in which we want the Android SDK to catch exceptions?

In this case, they can set an `options.Native.EnableBeforeSend` to true. But I can't think of  any use case in which they would since the event thrown by dotnet has a more detailed set of info

CHANGELOG.md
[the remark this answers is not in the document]
jamescrosswell: It feels like the missing attachment is a secondary issue, if we prevent the duplicate event that is missing the attachments from even being sent.

But probably doesn't matter too much either way. If you'd rather keep it in the changelog, that's fine.

src/Sentry/Platforms/Android/SentrySdk.cs
jamescrosswell: But what about java exceptions that don't come from a `JavaBackgroundThread`? This would discard those too right? It's just discarding all and any events that get captured by the Java SDK.

And if the user has set any custom BeforeSend callback, this workaround wouldn't work right???

bricefriha: All Java event throws a dotnet event, as far as I can tell. 

> if the user has set any custom BeforeSend callback, this workaround wouldn't work right???

it would work as long as `Native.EnableBeforeSend` is not set to true. if it's set to true, it means they technically want the Java event to appear

jamescrosswell: > All Java exceptions event throws a dotnet event, as far as I can tell.

I think this wouldn't:
```
try 
{
    // Do something crazy
}
catch(Exception e)
{
}
```

But I think this would:
```
try 
{
    // Do something crazy
}
catch(Exception e)
{
    throw e;
}
```


So we can't assume that all java exceptions will result in a corresponding .NET exception and we can't just throw away all the Java exceptions.

bruno-garcia: > All Java exceptions event throws a dotnet event, as far as I can tell.

If I write in Java/Kotlin `Sentry.captureException` does that go through the C# BeforeSend? I'd be surprised.

A NDK crash, does it go through the C# BeforeSend? Also would surprise me.

PR #3745, opened 2024-11-12
Prevent NullReferenceException in SentryTraceHeader Parsing

Victorvhn:
Hey y'all,

I recently encountered an issue where nearly every API request threw an exception after integrating Sentry with OpenTelemetry in my APIs. This issue only affected my environment, likely due to having **Just My Code** disabled in debugging options.

Upon investigation, I discovered that a null value was being passed to the `Parse` method in `SentryTraceHeader.cs`. The method was attempting to call `.Split()` on this null value, leading to an `NullReferenceException`.

To address this, I added a null check in the Parse method before calling `.Split()`.

Please review the changes when you have a chance and let me know if there are any additional adjustments or improvements you’d like to see. Thank you!

PR #3757, opened 2024-11-13
Prevent NullReferenceException in SentryTraceHeader Parsing

Victorvhn:
Hey y'all,

I recently encountered an issue where nearly every API request threw an exception after integrating Sentry with OpenTelemetry in my APIs. This issue only affected my environment, likely due to having **Just My Code** disabled in debugging options.

Upon investigation, I discovered that a null value was being passed to the `Parse` method in `SentryTraceHeader.cs`. The method was attempting to call `.Split()` on this null value, leading to an `NullReferenceException`.

To address this, I added a null check in the `Parse` method before calling `.Split()`.

Please review the changes when you have a chance and let me know if there are any additional adjustments or improvements you’d like to see. Thank you!

---

Moved from [PR](https://github.com/getsentry/sentry-dotnet/pull/3745)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,13 +19,13 @@
 
 ### Fixes
 
-- Fixed ArgumentNullException in FormRequestPayloadExtractor when handling invalid form data on ASP.NET ([#3734](https://github.com/getsentry/sentry-dotnet/pull/3734))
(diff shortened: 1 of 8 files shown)


bruno-garcia (commented):
Any particular reason you picked v5 branch for this fix? Would be great to put this out there for v4 too.

Oh, because it's a breaking change it seems?

jamescrosswell (approved):
Big thank you for the PR @Victorvhn and for fussing around with the branches. Much appreciated!

PR #3780, opened 2024-11-22
ci: change updater PR strategy to update

vaind:
The default (which was the previous behaviour when the udpater was created) creates new branch & PR for every release.

The "update" strategy updates a single PR per dependency.

### DIFF
--- a/.github/workflows/update-deps.yml
+++ b/.github/workflows/update-deps.yml
@@ -26,5 +26,6 @@ jobs:
     with:
       name: ${{ matrix.name }}
       path: ${{ matrix.path }}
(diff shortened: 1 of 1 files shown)


bruno-garcia (approved):

jamescrosswell:
Not sure I get this. Why does this matter? So what problem does changing to update strategy solve?

jamescrosswell:
> You can leave the PR open and it will get updated the next time there's a release on the dependency, instead of creating multiple PRs, one for each version.

Ahhh, I see. Cool, thanks @vaind !

Issue #3627, opened 2024-09-24
JNI DETECTED ERROR IN APPLICATION when reporting a crash in .NET MAUI Android App


dinisvieira-xo:
### Package

Sentry.Maui

### .NET Flavor

.NET

### .NET Version

8.0.8

### OS

Android

### SDK Version

4.11.0

### Steps to Reproduce

This might be an hard to reproduce issue outside of our app. Basically I started having **"random" crashes after 5 to 15 minutes of launching the app**.
During this time the app is reporting crashes normally but suddenly crashes with something like "`JNI DETECTED ERROR IN APPLICATION: JNI ERROR (app bug): jobject is an invalid global reference: 0x979a`" followed by a native crash that didn't offer a lot of information.

I enabled [GREF Logging](https://learn.microsoft.com/en-us/previous-versions/xamarin/android/deploy-test/environment) and reproduced the issue to find more information on where the issue is occurring. After reproducing a few times the issue always seemed to point at `Sentry.Maui.Internal.MauiDeviceData.ApplyMauiDeviceData` as a related cause of this.

**Important:** If I comment the code that calls the `SentrySdk.CaptureException` the app runs with no crashes for hours.

**Additional Info:**
- MAUI Version: 8.0.90
- Also happened on Sentry SDK 4.10.2
- Occurs in both Debug and Release
- Occurs both with AttachScreenshot true or false
- With GREF enabled (app running slower) the issue is even easier to reproduce (2 to 5 minutes)
- For the specific app I'm running the app is reporting crashes at about 2 per minute (for testing purposes)
- I'm using Prism.Pluggin.Logging to call Sentry, but I've already tried without Prism (simply calling Sentry directly to report) and the crash still occurs.

**Sentry Options:**
```
Debug = false;
AutoSessionTracking = true;
IsGlobalModeEnabled = false;
StackTraceMode = StackTraceMode.Enhanced;
DiagnosticLevel = SentryLevel.Debug;
TracesSampleRate = 1.0;
SampleRate = 1.0F;
MaxBreadcrumbs = 200;
CreateElementEventsBreadcrumbs = false;
IncludeBackgroundingStateInBreadcrumbs = false;
IncludeTextInBreadcrumbs = false;
IncludeTitleInBreadcrumbs = false;
SendDefaultPii = false;
AttachScreenshot = false;
```

### Expected Result

App should not crash and continue reporting/logging to Sentry normally.

### Actual Result

[...]

PR #3791, opened 2024-11-25
fix: MAUI Android - prevent JNI DETECTED ERROR IN APPLICATION crashes when too many events have to be processed at once

bricefriha:
This PR offers a fix for #3627.
When calling SentrySdk SentrySdk.CaptureException(ex) in multiple tight successions, Android apps would crash with a JNI DETECTED ERROR IN APPLICATION error message.

### DIFF
--- a/src/Sentry.Maui/Internal/MauiDeviceData.cs
+++ b/src/Sentry.Maui/Internal/MauiDeviceData.cs
@@ -54,7 +54,7 @@ internal static class MauiDeviceData
                 logger?.LogDebug("No permission to read battery state from the device.");
             }
 
-            device.IsOnline = networkStatusListener!.Online;
+            device.IsOnline ??= networkStatusListener!.Online;
 
 #if MACCATALYST || IOS
             if (MainThread.IsMainThread)
--- a/src/Sentry.Maui/Internal/MauiDeviceData.cs
+++ b/src/Sentry.Maui/Internal/MauiDeviceData.cs
@@ -57,7 +57,7 @@ internal static class MauiDeviceData
 #if MACCATALYST || IOS
             if (MainThread.IsMainThread)
             {
-                CaptureDisplayInfo();
+                CaptureDisplayInfo(displayInfo);
             }
             else
             {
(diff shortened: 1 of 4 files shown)


src/Sentry.Maui/Internal/MauiDeviceData.cs
```diff
@@ -6,29 +6,22 @@ namespace Sentry.Maui.Internal;
 
 internal static class MauiDeviceData
 {
-    public static void ApplyMauiDeviceData(this Device device, IDiagnosticLogger? logger)
+    public static void ApplyMauiDeviceData(this Device device, IDiagnosticLogger? logger, INetworkStatusListener? networkStatusListener, string deviceName, string deviceManufacturer, string deviceModel, string deviceType, bool? deviceSimulator, DevicePlatform devicePlatform, DisplayInfo displayInfo, bool supportsVibration, bool supportsAccelerometer, bool supportsGyroscope)
```
bricefriha, line 9: This is not the best when it comes to readability, and I'm not sure it makes much sense. (it works as is but it does mean it's the ultimate way to go)

so I need to find a better way

src/Sentry.Maui/Internal/MauiDeviceData.cs
```diff
@@ -6,29 +6,22 @@ namespace Sentry.Maui.Internal;
...
-            device.Manufacturer ??= deviceInfo.Manufacturer;
-            device.Model ??= deviceInfo.Model;
-            device.DeviceType ??= deviceInfo.Idiom.ToString();
-            device.Simulator ??= deviceInfo.DeviceType switch
-            {
-                DeviceType.Virtual => true,
-                DeviceType.Physical => false,
-                _ => null
-            };
+            device.Name ??= deviceName;
```
bricefriha, line 20: Passing a `DeviceInfo` parameter to `ApplyMauiDeviceData()` and calling deviceInfo.Name causes the crash to happen on this very line.

I'm trying to find a solution to avoid a crash here, and avoid the [above problem](https://github.com/getsentry/sentry-dotnet/pull/3791/files#r1858416894)

src/Sentry.Maui/Internal/MauiDeviceData.cs
```diff
@@ -22,8 +20,8 @@ public static void ApplyMauiDeviceData(this Device device, IDiagnosticLogger? lo
             device.Name ??= deviceInfo.Name;
             device.Manufacturer ??= deviceInfo.Manufacturer;
             device.Model ??= deviceInfo.Model;
-            device.DeviceType ??= deviceInfo.Idiom.ToString();
-            device.Simulator ??= deviceInfo.DeviceType switch
+            device.DeviceType ??= deviceIdiom;
```
bricefriha, line 23: doing `deviceInfo.Idiom.ToString()` causes the crash to occur

Issue #3777, opened 2024-11-21
Pin older iOS versions when building Sentry.Maui


jamescrosswell:
See [this discord conversation](https://discord.com/channels/732297728826277939/732297808148824115/1309180395140681830). 

It appears if we simply specify `net8.0-ios` and `net9.0-ios`, we'll get different build results depending on what the latests workload installed on the machine running the build is.

Rolf's recommendation (which might enable us to create more deterministic builds, so we get the same results on CI and dev boxes):
> set TargetFramework=net8.0-ios17.0 in your library project, that'll work for all .NET 8 iOS workloads and TargetFramework=net9.0-ios18.0 for .NET 9 (although net8.0-ios17.0 works in .NET 9 as well)

bricefriha:
Sounds good, I'll try that

bruno-garcia:
It's what they do here too: https://www.nuget.org/packages/CommunityToolkit.Maui.Core

jamescrosswell:
> But it's not clear to me yet why we should do that.

From [OS version in TFMs](https://learn.microsoft.com/en-us/dotnet/standard/frameworks#os-version-in-tfms):

> You can also specify an optional OS version at the end of an OS-specific TFM, for example, net6.0-ios15.0. The version indicates which APIs are available to your app or library. It doesn't control the OS version that your app or library supports at run time. It's used to select the reference assemblies that your project compiles against, and to select assets from NuGet packages. Think of this version as the "platform version" or "OS API version" to disambiguate it from the run-time OS version.

And further on:
> On Apple platforms (iOS, macOS, tvOS, and Mac Catalyst) in .NET 8 and earlier, the default TPV is the latest supported version in the currently installed workload. That means that updating the iOS workload in .NET 8, for example, might result in a higher default TPV, if support for a new version of iOS has been added in that workload. In the preceding table, the default TPV is the one in the initial release for the stated .NET version.

Consequently, when new workloads come out, we're implicitly bumping the version of APIs that our app supports by not specifying a TPV in our TFM. 

I think what happened in the last couple of days is that a new MAUI workload for ios was made available that implicitly bumped our TFM from `net9.0-18.0` to `net9.0-18.1`... and the APIs targeted by `net9.0-18.1` are only installed/available with Xcode 16.1... whereas our CI runner only had XCode 16.0 installed. 

[...]

PR #3798, opened 2024-11-27
Pin Target Platform Version in mobile TFM monkiers

jamescrosswell:
Resolves https://github.com/getsentry/sentry-dotnet/issues/3777:
- https://github.com/getsentry/sentry-dotnet/issues/3777

#skip-changelog

### DIFF
--- a/.github/workflows/device-tests-android.yml
+++ b/.github/workflows/device-tests-android.yml
@@ -39,7 +39,7 @@ jobs:
         with:
           name: device-test-android
           if-no-files-found: error
-          path: test/Sentry.Maui.Device.TestApp/bin/Release/net8.0-android/android-x64/io.sentry.dotnet.maui.device.testapp-Signed.apk
+          path: test/Sentry.Maui.Device.TestApp/bin/Release/net8.0-android34.0/android-x64/io.sentry.dotnet.maui.device.testapp-Signed.apk
 
   android:
     needs: [build]
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -87,7 +87,7 @@
   <!-- dotnet-gcdump needs net6+ and won't work on mobile. -->
   <PropertyGroup>
     <PlatformIsLegacy Condition="$(TargetFramework.StartsWith('net4')) or $(TargetFramework.StartsWith('netstandard'))">true</PlatformIsLegacy>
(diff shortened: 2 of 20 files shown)


Directory.Build.props
```diff
@@ -87,7 +87,7 @@
   <!-- dotnet-gcdump needs net6+ and won't work on mobile. -->
   <PropertyGroup>
     <PlatformIsLegacy Condition="$(TargetFramework.StartsWith('net4')) or $(TargetFramework.StartsWith('netstandard'))">true</PlatformIsLegacy>
-    <PlatformIsMobile Condition="$(TargetFramework.EndsWith('android')) or $(TargetFramework.EndsWith('ios')) or $(TargetFramework.EndsWith('maccatalyst'))">true</PlatformIsMobile>
+    <PlatformIsMobile Condition="$([MSBuild]::GetTargetPlatformIdentifier('$(TargetFramework)')) == 'android' or $([MSBuild]::GetTargetPlatformIdentifier('$(TargetFramework)')) == 'ios' or $([MSBuild]::GetTargetPlatformIdentifier('$(TargetFramework)')) == 'maccatalyst'">true</PlatformIsMobile>
```
vaind, line 90: See `TargetPlatformIdentifier` defined above

jamescrosswell: Interestingly, I wonder whether by calling it `TargetPlatformIdentifier` we could potentially break things:
- https://github.com/dotnet/msbuild/issues/7359

Some older build scripts might be relying on that property being set to `windows` on older TFMs (`netstandard2.0` and `netstandard2.1` are the only ones that would affect us, I think).

Unrelated to this PR though and we haven't run into any issues yet (that I know of)... but we could rename it just to avoid any potential clash.

.github/workflows/device-tests-android.yml
```diff
@@ -39,7 +39,7 @@ jobs:
         with:
           name: device-test-android
           if-no-files-found: error
-          path: test/Sentry.Maui.Device.TestApp/bin/Release/net8.0-android/android-x64/io.sentry.dotnet.maui.device.testapp-Signed.apk
+          path: test/Sentry.Maui.Device.TestApp/bin/Release/net8.0-android34.0/android-x64/io.sentry.dotnet.maui.device.testapp-Signed.apk
```
bruno-garcia, line 42: if you use -o when building to set the path, you can have a directory that's not going to change and avoid having to keep up with tfm changes

jamescrosswell: Ah, good idea... next time we touch this we can do that instead.

PR #3800, opened 2024-11-27
fix: cocoa bindings generated code line endings

vaind:
Fixes line endings in the generated code

on macOS git diff says warning: in the working copy of 'src/Sentry.Bindings.Cocoa/ApiDefinitions.cs', CRLF will be replaced by LF the next time Git touches it 

#skip-changelog


jamescrosswell (approved):
LGTM

PR #3822, opened 2024-12-04
Don't strip symbols for device tests

jamescrosswell:
When we run into issues running device tests, this makes it easier to make sense of any stack traces accompanying errors. 

#skip-changelog


bruno-garcia (approved):

bruno-garcia:
> When we run into issues running device tests, this makes it easier to make sense of any stack traces accompanying errors.

This should be default tbqh

PR #3850, opened 2024-12-16
fix: Ensure span ids are 16 characters

knd775:
Storing span ids as longs causes leading zeroes to be lost. This restores them.

### DIFF
--- a/src/Sentry/SpanId.cs
+++ b/src/Sentry/SpanId.cs
@@ -41,7 +41,7 @@ public readonly struct SpanId : IEquatable<SpanId>, ISentryJsonSerializable
     public override int GetHashCode() => StringComparer.Ordinal.GetHashCode(_value);
 
     /// <inheritdoc />
-    public override string ToString() => _value.ToString("x8");
+    public override string ToString() => _value.ToString("x8").PadLeft(16, '0');
 
     /// <summary>
     /// Generates a new Sentry ID.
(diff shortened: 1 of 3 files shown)


knd775:
Context:
https://discord.com/channels/621778831602221064/1318308956094992414

src/Sentry/SpanId.cs
```diff
@@ -41,7 +41,7 @@ namespace Sentry;
     public override int GetHashCode() => StringComparer.Ordinal.GetHashCode(_value);
 
     /// <inheritdoc />
-    public override string ToString() => _value.ToString("x8");
+    public override string ToString() => _value.ToString("x8").PadLeft(16, '0');
```
jamescrosswell, line 44: Is it possible to create a unit test that reproduces the issue (prior to this fix) and then apply this fix to make the test pass? That would prevent any regressions in future.

jamescrosswell (changes requested):
Thanks for the PR @knd775 ! Looks like a pretty straight forward fix... I just left one comment about preventing a regression.

btw: `test/Sentry.Tests/Protocol/Context/TraceTests.cs` might be a good place for the test(s).

src/Sentry/SpanId.cs
knd775: Done!

jamescrosswell (approved):
Fantastic - thanks @knd775 🙏🏻

PR #3856, opened 2024-12-18
Prepare 5.0.0 release

jamescrosswell:
#skip-changelog

- Tidy up the changelog
- Mark Auto Heap Dump functionality experimental (so we can change this after getting feedback)

### DIFF
--- a/src/Sentry/HeapDumpTriggers.cs
+++ b/src/Sentry/HeapDumpTriggers.cs
@@ -1,4 +1,8 @@
+using Sentry.Infrastructure;
+
 namespace Sentry;
+
+#if MEMORY_DUMP_SUPPORTED
 
 /// <summary>
 /// Delegate that determines whether a heap dump should be triggered or not.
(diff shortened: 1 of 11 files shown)


src/Sentry/HeapDumpTriggers.cs
```diff
@@ -1,11 +1,16 @@
+using Sentry.Infrastructure;
+
 namespace Sentry;
 
+#if MEMORY_DUMP_SUPPORTED
```
bruno-garcia, line 5: to be sure, we set this when we build right?

bruno-garcia (approved):

src/Sentry/HeapDumpTriggers.cs
jamescrosswell: Yeah that gets set here:
https://github.com/getsentry/sentry-dotnet/blob/ed0fd8a1c4383c70797b3b543b23e4f2aa692405/Directory.Build.props#L90-L97

Issue #3690, opened 2024-10-16
Add [GeneratedRegex] properties


jamescrosswell:
We've got lots of regexes defined as properties, with backing methods specifically because we had to use methods to leverage the source generated regexes. For example:
https://github.com/getsentry/sentry-dotnet/blob/3e70c3126515c87b1851b6a9a7afd99ffbb9aeb7/src/Sentry/Internal/OriginHelper.cs#L8-L18

We should be able to tidy that code up now that [[GeneratedRegex] can also be added to properties](https://learn.microsoft.com/en-us/dotnet/core/whats-new/dotnet-9/libraries#generatedregex-on-properties).

PR #3870, opened 2025-01-06
Task 3690: Use [GeneratedRegex] Properties

phunkeler:
This is meant to address #3690

As you can see, this doesn't really clean things up -- mainly to preserve regex source generation for `net8.0` (_Option 1_).

If we're fine using dynamic regex compilation at runtime for `net8.0` (_Option 2_) things get much simpler but it may come with performance changes.

<details>
  <summary>OriginHelper.cs (As-is)</summary>

```csharp
#if NET8_0_OR_GREATER 
     [GeneratedRegex(ValidOriginPattern, RegexOptions.Compiled)] 
     private static partial Regex ValidOriginRegEx(); 
     private static readonly Regex ValidOrigin = ValidOriginRegEx(); 
 #else 
     private static readonly Regex ValidOrigin = new (ValidOriginPattern, RegexOptions.Compiled); 
 #endif 

```

</details>

<details>
  <summary>OriginHelper.cs (Option 1)</summary>

```csharp
#if NET9_0_OR_GREATER
    [GeneratedRegex(ValidOriginPattern)]
    private static partial Regex ValidOrigin { get; }
#elif NET8_0
    [GeneratedRegex(ValidOriginPattern)]
    private static partial Regex ValidOriginRegex();
    private static readonly Regex ValidOrigin = ValidOriginRegex();
#else
    private static readonly Regex ValidOrigin = new(ValidOriginPattern, RegexOptions.Compiled);
#endif

```

</details>

<details>
  <summary>OriginHelper.cs (Option 2)</summary>

```csharp
#if NET9_0_OR_GREATER
    [GeneratedRegex(ValidOriginPattern)]
    private static partial Regex ValidOrigin { get; }
#else
    private static readonly Regex ValidOrigin = new(ValidOriginPattern, RegexOptions.Compiled);
#endif

```

</details>

Thoughts?

#skip-changelog

### DIFF
--- a/src/Sentry/BuiltInSystemDiagnosticsMeters.cs
+++ b/src/Sentry/BuiltInSystemDiagnosticsMeters.cs
@@ -24,10 +24,15 @@ public static partial class BuiltInSystemDiagnosticsMeters
     /// <summary>
     /// Matches the built-in Microsoft.AspNetCore.Hosting metrics
     /// </summary>
-#if NET8_0_OR_GREATER
+#if NET9_0_OR_GREATER
+    public static readonly StringOrRegex MicrosoftAspNetCoreHosting = MicrosoftAspNetCoreHostingRegex;
+
+    [GeneratedRegex(MicrosoftAspNetCoreHostingPattern)]
+    private static partial Regex MicrosoftAspNetCoreHostingRegex { get; }
+#elif NET8_0
     public static readonly StringOrRegex MicrosoftAspNetCoreHosting = MicrosoftAspNetCoreHostingRegex();
 
-    [GeneratedRegex(MicrosoftAspNetCoreHostingPattern, RegexOptions.Compiled)]
+    [GeneratedRegex(MicrosoftAspNetCoreHostingPattern)]
     private static partial Regex MicrosoftAspNetCoreHostingRegex();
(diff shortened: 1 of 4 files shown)


jamescrosswell:
Thanks heaps for the PR @phunkeler !

> As you can see, this doesn't really clean things up

It's a pity. I'd thought what enabled `[GeneratedRegex]` properties was the new support for partial properties in C#13... Presumably `GeneratedRegex` has also changed so that this can be applied to properties and not methods - which I'm guessing needs the net9.0 runtime. 

> Thoughts?

I'd say stick with Option 1 for the time being, so that people targeting `net8.0` also get the performance improvements of generated regex. It makes the code a bit more verbose in the short term but when `net8.0` reaches EOL we can simplify it and so we'll eventually end up with Option 2.

jamescrosswell:
btw: The builds on Ubuntu are failing periodically because of [this](https://github.com/getsentry/sentry-dotnet/pull/3866#issuecomment-2571947409) - I'm working on a fix in that branch, which you should get for free once that's all green and merged into main.

jamescrosswell (approved):
Thanks heaps for the PR @phunkeler !

Issue #3843, opened 2024-12-12
Requests proxied via YARP are missing in the traces


saithis:
### Package

Sentry

### .NET Flavor

.NET

### .NET Version

8.0.0

### OS

Any (not platform specific)

### SDK Version

4.13.0

### Steps to Reproduce

1. Create a ASP.NET Core project 
2. Add Sentry.AspNetCore and wire it up (.UseSentry + config)
```    
"Sentry": {
        "Dsn": "...",
        "IncludeActivityData": true,
        "MaxRequestBodySize": "None",
        "TracesSampleRate": 1.0
    },
``` 
3. Add Yarp.ReverseProxy
4. configure a route, that just echos the request back
5. configure a route, that takes the request and forwards it to the echo route via YARP (in production it would be another api for us, but for local debugging i tested it with the echo endpoint)
6. call the route via postman/etc.

 

### Expected Result

The proxied request should show up in the trace as a child 

### Actual Result

It doesn't
![Image](https://github.com/user-attachments/assets/64fc50f5-66c3-4bef-88a7-fba94684c995)

bruno-garcia:
Is the proxy forwarding all headers? I imagine that might be the culprit.

Doesn't sound like a Sentry issue tbqh, there's something added in the middle that must be aware of headers and propagate them.

If the proxy is .NET, and runs Sentry, for example `Sentry.AspNetCore`, our SDK does the propagation automatically.

Marioo1357:
@saithis When it started not work for you? From .Net 9 update or Sentry release?

saithis:
@bruno-garcia YARP should be forwarding all headers, yes. Our app is .NET, uses Sentry via Sentry.AspNetCore and uses the YARP nuget package for proxying the requests.

@Marioo1357 We just started using Sentry and I'm just trying to get it to work. We had no previous version with sentry.

I tried it with Sentry.OpenTelemetry now. It shows the missing traces then, but some things appear twice:

![Image](https://github.com/user-attachments/assets/c5d4b3a5-40e5-4807-874a-9c4c5130288a)
Seems like every http.client request is recorded via Sentry.AspNetCore and via Sentry.OpenTelemetry

bruno-garcia:
@jamescrosswell is the duplicated span when adding Sentry.OTel a known issue?

jamescrosswell:
> [@jamescrosswell](https://github.com/jamescrosswell) is the duplicated span when adding Sentry.OTel a known issue?

[...]

PR #3879, opened 2025-01-09
Add SentryOptions.DisableSentryHttpMessageHandler

jamescrosswell:
Addresses the duplicate trace problem described in https://github.com/getsentry/sentry-dotnet/issues/3843:
- https://github.com/getsentry/sentry-dotnet/issues/3843#issuecomment-2579087325

Essentially we allow SDK users to disable the `SentryHttpMessageHandler` if they are already using the `OpenTelemetry.Instrumentation.Http` package to trace outbound requests... to avoid duplicate instrumentation.  

This PR does not address any issues around YARP - that would require separate exploration.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## Unreleased
+
+- Added `SentryOptions.DisableSentryHttpMessageHandler` ([#3879](https://github.com/getsentry/sentry-dotnet/pull/3879))
+
 ## 5.0.1
 
 ### Fixes
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## Unreleased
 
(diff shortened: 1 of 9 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+## Unreleased
+
+- Added `SentryOptions.DisableSentryHttpMessageHandler` ([#3879](https://github.com/getsentry/sentry-dotnet/pull/3879))
```
```suggestion
- Added `SentryOptions.DisableSentryHttpMessageHandler`. Useful if you're using `OpenTelemetry.Instrumentation.Http` and ending up with duplicate spans. ([#3879](https://github.com/getsentry/sentry-dotnet/pull/3879))
```

bruno-garcia (approved):
Can we flip this to `false` automagically if `OpenTelemetry.Instrumentation.Http` is installed? If so, do we need a public options or would `internal` with `InternalsVisibleTo` be enough?

Issue #3887, opened 2025-01-13
Filtering Activities via Process is not respected by Sentry Exporter


CezaryKlus:
### Package

Sentry

### .NET Flavor

.NET

### .NET Version

8.0.0

### OS

Linux

### SDK Version

5.0.1

### Steps to Reproduce

1. Add filtering Processor, here an example of filtering out activities that do not belong to an ancestor that is an incoming Request or Message Consumer, or Hangfire Job

```
internal sealed class FilteringProcessor : BaseProcessor<Activity>
{
    private static bool HasParent(Activity activity, Func<Activity, bool> condition)
    {
        Activity? current = activity;

        while (current != null)
        {
            if (condition(current))
            {
                return true;
            }
            current = current.Parent;
        }

        return false;
    }

    // Called when an activity ends
    // Filters out activities that are not coming from ASP.NET Core request, MassTransit or Hangfire job
    public override void OnEnd(Activity activity)
    {
        var activityKinds = new ActivityKind[] { ActivityKind.Server, ActivityKind.Consumer };

        // If the activity does not have a parent that matches the specified kinds or tags, mark it as not recorded
        if (!HasParent(activity, a =>
            activityKinds.Any(y => a.Kind == y)
            || a.Tags.Any(y => y.Key == "job.id")))
        {
            Console.WriteLine($"Filtering out activity: {activity.OperationName} {activity.DisplayName}");
            activity.ActivityTraceFlags &= ~ActivityTraceFlags.Recorded;
            activity.IsAllDataRequested = false;
        }
    }
}
```

2. Register filtering Processor
```
 services.AddOpenTelemetry()
     .ConfigureResource(resource => resource
         .AddService(serviceName: ctx.HostingEnvironment.ApplicationName, serviceVersion: ServiceVersion.FromTypeAssembly<TStartup>())
         .AddHostDetector()
         )
     .WithTracing(tracerProviderBuilder =>
         tracerProviderBuilder
             .AddAspNetCoreInstrumentation(o =>
             {
                 // filter out health checks
                 o.Filter = (httpContext) => !httpContext.Request.Path.ToString().Contains(".probe", StringComparison.InvariantCultureIgnoreCase);
                 o.RecordException = true;
             }) // <-- Adds ASP.NET Core telemetry sources
             .AddHttpClientInstrumentation(
[...]

PR #3890, opened 2025-01-15
Filtered OTel activities should not be sent to Sentry

jamescrosswell:
Resolves https://github.com/getsentry/sentry-dotnet/issues/3887:
- https://github.com/getsentry/sentry-dotnet/issues/3887

## Manual Testing

Can be tested manually by altering our OTel Console sample to look something like this:

```csharp
var activitySource = new ActivitySource("Sentry.Samples.OpenTelemetry.Console");

SentrySdk.Init(options =>
{
    // You can set here in code, or you can set it in the SENTRY_DSN environment variable.
    // See https://docs.sentry.io/product/sentry-basics/dsn-explainer/
    options.Dsn = "https://eb18e953812b41c3aeb042e666fd3b5c@o447951.ingest.sentry.io/5428537";

    options.Debug = true;
    options.UseOpenTelemetry(); // <-- Configure Sentry to use OpenTelemetry trace information
});

using var tracerProvider = Sdk.CreateTracerProviderBuilder()
    .AddSource(activitySource.Name)
    .AddHttpClientInstrumentation()
    .AddProcessor(new FilteringProcessor()) // <-- Adds a processor to filter out unwanted activities
    .AddSentry() // <-- Configure OpenTelemetry to send traces to Sentry
    .Build();

Console.WriteLine("Hello World!");

// Finally we can use OpenTelemetry to instrument our code. This activity will be captured as a Sentry transaction.
using (var activity = activitySource.StartActivity("Main"))
{
    // This creates a span called "Task 1" within the transaction
    using (var task = activitySource.StartActivity("Task 1"))
    {
        task?.SetTag("Answer", 42);
        Thread.Sleep(100); // simulate some work
        Console.WriteLine("Task 1 completed");
        task?.SetStatus(Status.Ok);
    }

    // Since we use `AddHttpClientInstrumentation` when initializing OpenTelemetry, the following Http request will also
    // be captured as a Sentry span
    var httpClient = new HttpClient();
    var html = await httpClient.GetStringAsync("https://example.com/");
    Console.WriteLine(html);
}

using (var activity = activitySource.StartActivity("Hide Me"))
{
    // This activity will be filtered out by the processor
    Console.WriteLine("This activity will be filtered out");
    Thread.Sleep(100); // simulate some work
}

Console.WriteLine("Goodbye cruel world...");

internal sealed class FilteringProcessor : BaseProcessor<Activity>
{
    // Filters out some activities
    public override void OnEnd(Activity activity)
    {
        // If the activity does not have a parent that matches the specified kinds or tags, mark it as not recorded
        if (activity.DisplayName == "Hide Me")
        {
            Console.WriteLine($"Filtering out activity: {activity.OperationName} {activity.DisplayName}");
            activity.ActivityTraceFlags &= ~ActivityTraceFlags.Recorded;
            activity.IsAllDataRequested = false;
        }
    }
}
```


src/Sentry.OpenTelemetry/SentrySpanProcessor.cs
```diff
@@ -181,6 +181,14 @@ public override void OnEnd(Activity data)
             return;
         }
 
+        // Skip any activities that are not recorded.
+        if (data is { Recorded: false, IsAllDataRequested: false })
```
bitsandfoxes, line 185: It needs both to be "not recorded"?

jamescrosswell: I did that subconsciously to be honest. I have a vague memory of the OTel docs saying `IsAllDataRequested` should always be false when `Recorded` is false, but for our purposes I think you're right - we can just check `Recorded`. 

Note: The [docs for IsAllDataRequested](https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.activity.isalldatarequested?view=net-9.0) say:
> indicates whether this activity should be populated with all the propagation information, as well as all the other properties, such as links, tags, and events

src/Sentry.OpenTelemetry/SentrySpanProcessor.cs
```diff
@@ -181,6 +181,14 @@ public override void OnEnd(Activity data)
             return;
         }
 
+        // Skip any activities that are not recorded.
+        if (data is { Recorded: false })
+        {
+            _options?.DiagnosticLogger?.LogDebug($"Ignoring unrecorded Activity {data.SpanId}.");
```
jamescrosswell, line 187: Should pass in data.SpanId as a separate parameter. This avoids a string.Format and an allocation in scenarios where debugging is disabled.

aritchie (approved):

Issue #3902, opened 2025-01-16
Managed NullReferenceException causes EXC_BAD_ACCESS and SIGSEGV errors on iOS and Android 


jamescrosswell:
## Problem

When the Sentry Native SDK is used in apps that are AOT compiled (for example when publishing Android or iOS apps in release mode), `SIGSEGV` reports are captured by the NDK for managed `NullReferenceException` exceptions, even when the managed code in question catches and handles such exceptions. 

## Steps to reproduce

- Create a MAUI app
- Add a button with some code that will generate a `NullReferenceException` in the event handler... e.g.:
  ```chsarp
          string s = null;
        _logger.LogInformation("The length of the string is {StringLength}", s!.Length);
  ```
- Run the application in Release mode on a Simulator

**Expected Result**

A managed `NullReferenceException` should be captured by the .NET SDK

**Actual Result**

In addition to a `NullReferenceException` being captured by the .NET SDK, `SIGSEGV` signal error is captured by the NDK (on Android) or the CocoaSDK (on iOS).

## Analysis

Full analysis from @supervacuus here:

- https://github.com/dotnet/android/issues/9055#issuecomment-2261347912

On the basis of the analysis above,  we attempted to invert the order in which the Native and Managed signal handlers get run in this PR:
- https://github.com/getsentry/sentry-dotnet/pull/3694/files#diff-382226eacb7ee000748a4a5ea92c800b032a7225426a84d1e945e5aff7f20011

However this led to users' applications crashing (rather than just reporting erroneous `SIGSEGV` errors). So this change was reverted in:
- https://github.com/getsentry/sentry-dotnet/pull/3871

#### Gist

📄 [NullReference SIGSEGV Exception.pdf](https://github.com/user-attachments/files/18445952/NullReference.SIGSEGV.Exception.pdf)

## Current State

> We don't have the required resources available on the Native team to solve this end-to-end at the moment, so coming back to https://github.com/getsentry/sentry-dotnet/issues/3861#issuecomment-2574581724, we need to recommend deactivating the native crash reporters on Android and iOS for MAUI for the time being to mitigate the issue of duplicate reports.
> 
> However we're in the process of getting more resources on the downstream team(s) which means we'll hopefully be able to tackle the root cause some time in the next few months.  

 _Originally posted by @kahest in [#3861](https://github.com/getsentry/sentry-dotnet/issues/3861#issuecomment-2586751817)_

# Workaround (short term)

[...]

Issue #3898, opened 2025-01-16
.Net-Android - <unknown> SIGSEGV Segfaults


IainS1986:
We've only recently integrated Sentry into our project as a replacement for AppCenter.

We are *not* a MAUI project, we are just a straight .net-android and .net-ios native projects.

While testing internally, we are seeing a fair few android SIGSEGV errors coming through which show little to no real information for tracking down. I'm wondering if theres a symbolication issue we haven't got setup properly and wonder if theres some guidance on that.

The sorts of stack traces we are getting are...

```
OS Version: Android 14 (UKQ1.231108.001 release-keys)
Report Version: 104

Exception Type: Unknown (SIGSEGV)

Application Specific Information:
Segfault

Thread 0 Crashed:
0   libmonosgen-2.0.so              0x7d9b7e8efc        <unknown> + 539479674620
1   <unknown>                       0x7e07349ff0        <unknown>
```

Or

```
OS Version: Android 14 (AP2A.240605.024)
Report Version: 104

Exception Type: Unknown (SIGSEGV)

Application Specific Information:
Segfault

Thread 0 Crashed:
0   split_config.arm64_v8a.apk      0x77ad1a0efc        <unknown> + 514005274364
1   <unknown>                       0x1                 <unknown>
```

With the image viewer showing missing images...

![Image](https://github.com/user-attachments/assets/8c8b390d-99dd-4e2a-a542-207527e9fad7)

![Image](https://github.com/user-attachments/assets/bed8ccac-3abe-4b7a-8685-b5618dae41eb)

The breadcrumbs show some in similar 'places' but theres also a bit of a random array of times the crashes appeared to occur.

There also seem to be more happening than people are really reporting too me.

Is there anyway to get more information for these stack traces?
We upload the mappings.txt file for proguard, but we don't upload symbolication symbols. Where would they 'be' once built and how do we upload them?

We do all our uploading via DevOps and the API - so I'm assuming we would upload something like so,
'sentry-cli debug-files upload -o $(sentry.org.slug) -p $(sentry.project.android) $(Pipeline.Workspace)/$(artifact.name)/somesymbolfile'

But I'm not 100% on what symbol files there are than can be uploaded.

IainS1986:
So brief googling i've seen some talks about the Concurrent GC being an issue in .net-android 
https://github.com/dotnet/maui/issues/21479 and
https://github.com/dotnet/runtime/issues/105389

[...]

Issue #3837, opened 2024-09-29
C# Maui.Android Release Mode  SIGSEGV


gwise-vision:
### Environment

SaaS (https://sentry.io/)

### Steps to Reproduce

When deploying in release mode in C# Maui.Android, Sentry crashes every time. The Android app does not crash, but only the Sentry log is as follows:

![Image](https://github.com/user-attachments/assets/7cdff7df-57f5-4123-8e0b-9542e2b1424f)

jamescrosswell:
Possibly related to:
- Either https://github.com/getsentry/sentry-dotnet/pull/3694
- Or https://github.com/getsentry/sentry-dotnet/issues/3828

We have a fix for the first in the pipeline. The second one I haven't had a chance to look at yet.

jamescrosswell:
@gwise-vision we just released version 5.0.0 of the SDK, which includes the fix from https://github.com/getsentry/sentry-dotnet/pull/3694. Are you able to update to version 5 to see if that fix is related to the issue you're experiencing?

The other thing to check would be that you're not enabling AOT compilation as this [isn't yet supported](https://github.com/dotnet/runtime/issues/109284#issuecomment-2441804211) on Android.

jamescrosswell:
@gwise-vision just following up on this. It looks like this is due to the way the Native SDK for Sentry interacts with the .NET SDK for Sentry. As a workaround, for the time, you can disable the Native SDK when initialising Sentry on Android:

```csharp
#if ANDROID
             options.Native.EnableNdk = false;
#endif
```

The default value for `EnableNdk` is true. We'll make a new release to change that default value until we can properly work through the issues with the Native SDK. See [#3902 ](https://github.com/getsentry/sentry-dotnet/issues/3902) for further information.

gwise-vision:
> [@gwise-vision](https://github.com/gwise-vision) just following up on this. It looks like this is due to the way the Native SDK for Sentry interacts with the .NET SDK for Sentry. As a workaround, for the time, you can disable the Native SDK when initialising Sentry on Android:
> 
> #if ANDROID
>              options.Native.EnableNdk = false;
> #endif
> The default value for `EnableNdk` is true. We'll make a new release to change that default value until we can properly work through the issues with the Native SDK. See [#3902 ](https://github.com/getsentry/sentry-dotnet/issues/3902) for further information.

OK~

gwise-vision:
#if ANDROID
             options.Native.EnableNdk = false;
[...]

PR #3903, opened 2025-01-17
Suppress Native SIGSEGV signal errors on Android

jamescrosswell:
See https://github.com/getsentry/sentry-dotnet/issues/3902:
- https://github.com/getsentry/sentry-dotnet/issues/3902

Resolves [#3837](https://github.com/getsentry/sentry-dotnet/issues/3837)
Resolves [#3898](https://github.com/getsentry/sentry-dotnet/issues/3898)
Resolves [#3935](https://github.com/getsentry/sentry-dotnet/issues/3935)

This PR uses the Native BeforeSend callback on Android to intercept and suppress SIGSEGV errors, as these duplicate managed NullReferenceExceptions.

There's a small risk we're suppressing native SIGSEGV errors that are not wrapped by managed exceptions, but there's no easy way around that for the time being. This is the short term workaround described in https://github.com/getsentry/sentry-dotnet/issues/3902

In case SDK users see this as a problem though, we've added an option to let them control whether or not to suppress SIGSEGV Segfault errors. 

# Usage

```csharp
        var builder = MauiApp.CreateBuilder()
            .UseMauiApp<App>()
            .UseSentry(options =>
            {
                // All the standard options

#if ANDROID
                // Suppress native SIGSEGV Segfault errors
                options.Android.SuppressSegfaults = true;
#endif
            })
```

This should be documented. See:
- https://github.com/getsentry/sentry-dotnet/issues/3943


CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 ## Unreleased
 
+### Significant change in behavior
+- The NDK is now disabled by default on Android, to prevent SIGSEGV errors resulting from managed NullReferenceExceptions ([#3903](https://github.com/getsentry/sentry-dotnet/pull/3903))
```
bruno-garcia, line 6: "Crash detection in native code (NDK) is now disabled..."

bruno-garcia (commented):
It's quite unfortunate since now users will be blind of native crashes.

bruno-garcia:
@filipnavara any chance you can think of a better way?

filipnavara:
> @filipnavara any chance you can think of a better way?

I'll think about it. That said, I observed it and never really considered it a bug... NRE in .NET is almost never intentional, so this essentially just caused double reporting. Depending on .NET runtime used (NativeAOT vs MonoVM) this may contain some useful information.

jamescrosswell:
> NRE in .NET is almost never intentional, so this essentially just caused double reporting. Depending on .NET runtime used (NativeAOT vs MonoVM) this may contain some useful information.

Thanks @filipnavara - I think that's a perfectly valid perspective.

Currently this PR simply suppresses all SIGSEGV exceptions... which runs the risk of suppressing some that come from native code (and aren't reported as NREs). We've got two sub-optimal options at this point:
1. Let the SIGSEGV errors bubble up (possibly duplicating NREs)
2. Suppress the SIGSEGV errors... prevents duplicates but possibly suppresses genuine native errors

I think we have to pick one of those as the default behaviour, at the very least. But potentially we could expose an option to let SDK users override the default choice we make, depending on whether they prefer to catch all/every error (but possibly some twice) or whether they're happy to let a few slip through the cracks but prevent duplicates.

aritchie:
What about providing an option to control whether or not segfaults are suppressed.  If they are annoying and not providing value to the user, disable them otherwise leave them on.  In my experience, segfaults were a rare thing so I would still want to see them in most cases.

aritchie (approved):

Issue #3776, opened 2024-11-21
Sentry reports handled NRE as EXC_BAD_ACCESS signal


TimBurik:
### Package

Sentry

### .NET Flavor

.NET

### .NET Version

9.0.100

### OS

iOS

### SDK Version

4.13.0

### Steps to Reproduce

1. Create empty dotnet iOS project, add Sentry nuget package;
2. In `FinishedLaunching()`, initialize SentrySdk:
```
SentrySdk.Init(options =>
 {
    options.Dsn = "...";
    options.Debug = true;
    options.SampleRate = 1.0f;
});
```
3. After Sentry is initialized, execute a code like this:
```
try
{
    object o = null;
    Console.WriteLine(o.ToString()); //throws NullReferenceException
}
catch (Exception e)
{
    Console.WriteLine(e);
}
```

### Expected Result

Since exception is handled, there should be no reports in Sentry.

### Actual Result

Sentry writes down the envelope and sends it in the next session. Example of the crashlog: [crashlog.txt](https://github.com/user-attachments/files/17847559/crashlog.txt)

In production, we also see examples of double reports - when unhandled managed NRE and native SIGSEGV are reported from the same device at the same time with very similar stacktraces. But it seems to be the result of the same issue.

I seems that this issue is related to the order of signal handlers and has the same root cause as https://github.com/dotnet/android/issues/9055 (seems to have a fix on the Sentry side https://github.com/getsentry/sentry-native/issues/1026). Also it might be related to this issue: https://github.com/getsentry/sentry-dotnet/issues/3678

jamescrosswell:
Hi @TimBurik,

Thanks for reporting this and thanks for tracking down related issues.

> I seems that this issue is related to the order of signal handlers and has the same root cause as https://github.com/dotnet/android/issues/9055 (seems to have a fix on the Sentry side https://github.com/getsentry/sentry-native/issues/1026).

I suspect you're right. Support for alternative ordering of the signal handlers  has already been added in [sentry-native 0.7.13](https://github.com/getsentry/sentry-native/releases/tag/0.7.13). 

This PR is meant to address this for Android:
- https://github.com/getsentry/sentry-dotnet/pull/3694

It's [pending implementation of the same in the Java SDK](https://github.com/getsentry/sentry-dotnet/pull/3694#issuecomment-2483041757).

It looks like we'd need to do something similar for iOS eh @bricefriha?  Currently the fix is only applied in the Android/SentrySdk:
[...]

PR #3909, opened 2025-01-21
Prevent Native EXC_BAD_ACCESS signal for NullRefrenceExceptions

jamescrosswell:
Resolves https://github.com/getsentry/sentry-dotnet/issues/3776:
- https://github.com/getsentry/sentry-dotnet/issues/3776

# Analysis

The following all appear to impact error reporting behaviour for iOS applications:
1. The runtime (determined by whether AOT compilation is enabled or not)
2. The `ObjCRuntime.MarshalManagedExceptionMode`
3. Whether or not we try to suppress EXC_BAD_ACCESS errors (as in this PR)
4. Whether or not the exception is caught/ignored by managed code

### MONO vs CLR Runtime

When AOT compilation is not used, .NET iOS apps use the Mono runtime. When AOT compilation is used, .NET iOS apps use the CLR runtime. Managed exceptions are [marshalled](https://learn.microsoft.com/en-us/previous-versions/xamarin/ios/platform/exception-marshaling) differently by the two different runtimes and [unwinding native exceptions is currently not supported by the CLR](https://github.com/xamarin/xamarin-macios/blob/0a81da28e4217624a54ccb83541e1eb8ba651fca/runtime/runtime.m#L2480-L2483).

### MarshalManagedExceptionMode

In earlier versions of .NET, the `AppDomain.CurrentDomain.UnhandledException` was not triggered correctly unless  the marshalling mode for managed exceptions was set to `MarshalManagedExceptionMode.UnwindNativeCode`. (see https://github.com/xamarin/xamarin-macios/issues/15252 and [the workaround in Sentry](https://github.com/getsentry/sentry-dotnet/blob/81452c164d25925f4e796849f844e46ea9d3adfc/src/Sentry/Platforms/Cocoa/SentrySdk.cs#L13-L16)).

This has been [fixed in net9.0 on the MONO runtime](https://github.com/xamarin/xamarin-macios/issues/15252#issuecomment-2349301905).

### Scenarios

I tested the various different permutations of the above, with the following results:
* [EXC_BAD_ACCESS Exception scenario results](https://docs.google.com/spreadsheets/d/e/2PACX-1vTBRKHYlFtu1-Zt-0y9Nw4Ts4byd0mdDBuwQKg_bpoJpeh-rZeJxhgTa1KUMCDhtG2lRahDWzpK8EgA/pubhtml?gid=0&single=true)

# Solution

### AOT / CLR

- [x] Fix the issue of not getting NREs (unhandled exception handler not firing) in AOT Apps
  - Possibly, per Rolf's suggestion, we could [intercept different events to achieve this](https://github.com/xamarin/xamarin-macios/issues/15252#issuecomment-1154075753)
- [x] Suppress `EXC_BAD_ACCESS` errors, to avoid duplication (this PR)

### MONO

- https://github.com/getsentry/sentry-dotnet/pull/3912
  - Setting MarshalManagedExceptionMode.UnwindNative has no effect when the app is running on MONO in net9.0
  - We can't set MarshalManagedExceptionMode.UnwindNative in AOT (See Note 3)
  - **Note**: PR3912 may be rendered obsolete by this PR... since we no longer use `AppDomain.UnhandledException`
- [x] Also suppress EXC_BAD_ACCESS errors on MONO (this PR)

### Notes on EXC_BAD_ACCESS suppression 

We don't know, from the stack trace, whether the EXC_BAD_ACCESS comes from native or managed code. This workaround **_assumes_** it comes from managed code and will be reported separately as a managed exception. 

Prior to this PR the behaviour was the opposite. All EXC_BAD_ACCESS errors were assumed to be native exceptions and reported by the native SDK (so often users would get a EXC_BAD_ACCESS, that was a bit confusing, in addition to an NRE that makes more sense to .NET devs)

**Potential alternative**: add an option to let users decide whether to suppress these native errors or not.


src/Sentry/Platforms/Cocoa/SentrySdk.cs
```diff
@@ -154,20 +154,35 @@ private static void InitSentryCocoaSdk(SentryOptions options)
...
+                    return null!;
+                }
+
+                // In the case of NullReferenceExceptions, Xamarin's unhandled exception handler doesn't seem to catch
+                // it... so we filter these explicitly irrespective of the stack trace. This is possibly a bit dangerous
+                // as we could be filtering exceptions from native code blocks that don't get captured by our managed
+                // SDK. We don't have any easy way to know whether the exception is managed code (compiled to native)
+                // or bona fide native code though.
+                // See: https://github.com/getsentry/sentry-dotnet/issues/3776
+                if (ex.Type == "SIGABRT" && ex.Value.Contains("Attempted to dereference null pointer."))
```
jamescrosswell, line 183: This may not work on non-english versions of iOS. We should look at matching some of the other characteristics of the exception instead.

bruno-garcia (approved):
This seems to only remove the events we'd like to drop. @filipnavara might have tips on how to do this better but ai think it's good enough

jamescrosswell:
> This seems to only remove the events we'd like to drop. @filipnavara might have tips on how to do this better but ai think it's good enough

Kind of. 

If it's a `SIGABRT` that results from something that has already been captured and translated to a managed `NullReferenceException` then this will remove an event we're not interested in (we've already processed it in the managed SDK). 

If the `SIGABRT` results from a null pointer reference in some native code that didn't get translated to a managed `NullReferenceException` though, we're dropping an exception here and the customer won't have any visibility of this. 

Those are our two options at the moment though: either have duplicate events (one of them confusing) or risk dropping legitimate native exceptions. 

Unfortunately there's nothing in the stack trace we can use for these exceptions, like we did [here](https://github.com/getsentry/sentry-dotnet/blob/c20d16245c7e2d2f9056a49780593a391e59eda1/src/Sentry/Platforms/Cocoa/SentrySdk.cs#L168-L167).

src/Sentry/Platforms/Cocoa/RuntimeMarshalManagedExceptionIntegration.cs
```diff
@@ -0,0 +1,45 @@
+using ObjCRuntime;
+using Sentry.Extensibility;
+using Sentry.Integrations;
+
+namespace Sentry.Cocoa;
+
+/// <summary>
+/// When AOT Compiling iOS applications, the AppDomain UnhandledExceptionHandler doesn't fire. So instead we intercept
```
bruno-garcia, line 8: oh we've been missing out on a category of errors?

src/Sentry/Platforms/Cocoa/RuntimeMarshalManagedExceptionIntegration.cs
```diff
@@ -0,0 +1,45 @@
...
+        _runtime.MarshalManagedException += Handle;
+    }
+
+    // Internal for testability
+    [SecurityCritical]
+    internal void Handle(object sender, MarshalManagedExceptionEventArgs e)
+    {
+        _options?.LogDebug("Runtime Marshal Managed Exception");
+
+        if (e.Exception is { } ex)
```
bruno-garcia, line 33: `e` has: https://learn.microsoft.com/en-us/dotnet/api/objcruntime.marshalmanagedexceptionmode?view=xamarin-ios-sdk-12

We should add that detail as it can be useful context

src/Sentry/Platforms/Cocoa/RuntimeMarshalManagedExceptionIntegration.cs
```diff
@@ -0,0 +1,45 @@
...
+        {
+            ex.SetSentryMechanism(
+                "Runtime.MarshalManagedException",
+                "This exception was caught by the .NET Runtime Marshal Managed Exception global error handler. " +
+                "The application may have crashed as a result of this exception.",
```
bruno-garcia, line 38: It's not certain that it did crash?

src/Sentry/Platforms/Cocoa/SentrySdk.cs
```diff
@@ -154,20 +149,30 @@ private static void InitSentryCocoaSdk(SentryOptions options)
...
+
+                // Thankfully, sometimes we can see Xamarin's unhandled exception handler on the stack trace, so we can filter
+                // them out. Here is the function that calls abort(), which we will use as a filter:
+                // https://github.com/xamarin/xamarin-macios/blob/c55fbdfef95028ba03d0f7a35aebca03bd76f852/runtime/runtime.m#L1114-L1122
                 if (ex.Type == "SIGABRT" && ex.Value == "Signal 6, Code 0" &&
                     ex.Stacktrace?.Frames.Any(f => f.Function == "xamarin_unhandled_exception_handler") is true)
                 {
-                    // Don't sent it
+                    // Don't send it
+                    return null!;
```
bruno-garcia, line 167: would we want to have a debug level log in this case?

test/Sentry.Tests/Platforms/iOS/RuntimeMarshalManagedExceptionIntegrationTests.cs
```diff
@@ -0,0 +1,70 @@
+# if IOS
```
bruno-garcia, line 1:
```suggestion
#if IOS
```

bruno-garcia (approved):
Can we have a device test that verifies this? I imagine it would only be possible if we don't actually crash the app though

src/Sentry/Platforms/Cocoa/RuntimeMarshalManagedExceptionIntegration.cs
jamescrosswell: Essentially, yes. Matt created this workaround to the problem:
https://github.com/getsentry/sentry-dotnet/blob/c20d16245c7e2d2f9056a49780593a391e59eda1/src/Sentry/Platforms/Cocoa/SentrySdk.cs#L13-L17

However that marshalling mode isn't supported when AOT compiling iOS apps. It [causes xamarin to throw an assertion error](https://github.com/xamarin/xamarin-macios/blob/907081f787315704a01c940cf28b46b47db23df0/runtime/runtime.m#L2452-L2462), which is unrelated to the error from the user's code.

src/Sentry/Platforms/Cocoa/RuntimeMarshalManagedExceptionIntegration.cs
jamescrosswell: I'm not sure about that no. We're using this event instead of `AppDomain.Current.UnhandledException` because `Runtime.MarshalManagedException` fires even in AOT compiled applications. 

When testing, it doesn't fire for exceptions that are caught by managed code (in a `try..catch` block) so I I think it ordinarily would crash the application. I don't know enough about the interface between .NET and ObjectiveC to be sure of this though.

jamescrosswell:
> Can we have a device test that verifies this? I imagine it would only be possible if we don't actually crash the app though

Tricky... the fix using the native.OnBeforeSend handler.

An integration test might be possible but those are quite painful to build and maintain. Ivan basically has some pwsh scripts that create new MAUI apps with csharp source code, build and run these and then watch for stuff in the logs. I think they're good for smoke testing (making sure we can build and run JIT and AOT apps that init the SDK successfully) but not a good idea to try to build more sophisticated tests that way.

src/Sentry/Platforms/Cocoa/RuntimeMarshalManagedExceptionIntegration.cs
bruno-garcia: But regardless of that bool, if that fires (on your tests) the app crashes? If so, I'd indicate that's the case. But keeping it vague in here is fine, I was just wondering how this works

PR #3917, opened 2025-01-24
chore: update perfview

vaind:


jamescrosswell (approved):

jamescrosswell:
Oh, I see this is still draft. @vaind is there something else needs doing on this PR?

vaind:
> is there something else needs doing on this PR?

Maybe. There's [an issue with net8](https://github.com/getsentry/sentry-dotnet/issues/3924) I was hoping to fix just by upgrading and that didn't work out. I've raised an issue upstream and we'll see if that leads to a new perfview release in which case we'd get that instead. For now, I'd leave this as a draft and not merge.

PR #3936, opened 2025-02-03
Emit transaction.data inside contexts.trace.data

### LINKED ISSUE #3873: Emit `transaction.data` inside `contexts.trace.data`
Relates to https://github.com/getsentry/team-sdks/issues/95
See https://github.com/getsentry/team-sdks/issues/95#issuecomment-2575548026 for more details.

## Current Behavior
- We currently map `extra` to `data` in our SDK
- When serializing `SentryTransaction`, the extra data gets written to `extra`. The `ITransactionTracer` inherits from `IHasExtra` so this makes sense.
https://github.com/getsentry/sentry-dotnet/blob/7024f37efa663bbf054913b5a4659b9b407ab47d/src/Sentry/SentryTransaction.cs#L404

## Problem
The current inheritance structure prevents us from correctly mapping:
- Transaction data to `contexts.trace.data`
- Span data to `span.data`

## Proposed Solution
Instead of modifying the inheritance structure for transactions and spans (which would be hacky), we should:
1. Add `SetData` methods to both `TransactionTracer` and `SpanTracer`
2. Keep existing `SetExtra` functionality
   - This allows users to quickly add contexts without creating a full context
   - Note: We do not support arbitrary key-value pairs on the context

## Technical Details
- Transaction data should map to: `contexts.trace.data`
  - Implementation via [ITraceContext](https://github.com/getsentry/sentry-dotnet/blob/main/src/Sentry/Protocol/ITraceContext.cs) so this gets written in the correct place automatically with
https://github.com/getsentry/sentry-dotnet/blob/7024f37efa663bbf054913b5a4659b9b407ab47d/src/Sentry/SentryTransaction.cs#L398
- Span data should map to: `span.data`

aritchie:
Resolves #3873 

Removed serialization of "Extras" on SentryTransaction 

Unit tests included


test/Sentry.Tests/SerializationTests.Serialization_TransactionAndSpanData.verified.txt
```diff
@@ -0,0 +1,44 @@
...
+  SampleRate: 0.0,
+  Request: {},
+  Contexts: {
+    trace: {
+      SpanId: {},
+      ParentSpanId: {},
+      Operation: operation,
+      Description: ,
+      IsSampled: false,
+      Extra: {
```
jamescrosswell, line 20: @aritchie this is suppsed to be `Data` now right (not `Extra`)?

test/Sentry.Tests/SerializationTests.Serialization_TransactionAndSpanData.verified.txt
```diff
@@ -0,0 +1,44 @@
...
+        transaction1: transaction_value
+      }
+    }
+  },
+  User: {},
+  Sdk: {},
+  Extra: {
```
jamescrosswell, line 27: We presumably don't want to send this anymore as it would simply duplicate the Contexts.trace.data...

test/Sentry.Tests/SerializationTests.Serialization_TransactionAndSpanData.verified.txt
```diff
@@ -0,0 +1,44 @@
...
+  },
+  Spans: [
+    {
+      SpanId: {},
+      ParentSpanId: {},
+      StartTimestamp: DateTimeOffset_2,
+      IsFinished: false,
+      Operation: childop,
+      IsSampled: false,
+      Extra: {
```
jamescrosswell, line 38: Same again - I think this is suppose to be serialised as a data property on spans now right?

jamescrosswell (changes requested):
@aritchie awesome work getting a PR done on day 1! I wasn't necessarily even expecting you to be able to build the repo without some help/questions. 

See comments about the name of the field though - I think the purpose of this PR is to change that from Extra to Data for protocol compatibility with the other SDKs.

test/Sentry.Tests/SerializationTests.Serialization_TransactionAndSpanData.verified.txt
aritchie: This appears to be the result of letting verify do the serialization.  I did this in order to let it scrub the dates.  The built-in ToJsonString functions are generating the data outputs.  I'll see what I can do with the test.

aritchie: I've moved the tests to just json path walk to ensure the tree is serialized properly

test/Sentry.Tests/SerializationTests.Serialization_TransactionAndSpanData.verified.txt
aritchie: I've moved the tests to just json path walk to ensure the tree is serialized properly

test/Sentry.Tests/SerializationTests.Serialization_TransactionAndSpanData.verified.txt
aritchie: I've moved the tests to just json path walk to ensure the tree is serialized properly

src/Sentry/Protocol/Trace.cs
```diff
@@ -7,7 +7,7 @@ namespace Sentry.Protocol;
 /// <summary>
 /// Trace context data.
 /// </summary>
-public class Trace : ITraceContext, ISentryJsonSerializable, ICloneable<Trace>, IUpdatable<Trace>
+public class Trace : ITraceContext, IHasExtra, ISentryJsonSerializable, ICloneable<Trace>, IUpdatable<Trace>
```
jamescrosswell, line 10: @aritchie if I'm not way off base with [this comment](https://github.com/getsentry/sentry-dotnet/issues/3873#issuecomment-2635291329), I think `Trace` doesn't need to implement `IHasExtra`... we can just add some SetData and GetData methods to this (or, more idiomatic for .NET, just add a Data property).

src/Sentry/Protocol/Trace.cs
```diff
@@ -63,7 +78,8 @@ internal set
         Operation = Operation,
         Origin = Origin,
         Status = Status,
-        IsSampled = IsSampled
+        IsSampled = IsSampled,
+        _data = _data.ToDict()
```
jamescrosswell, line 82:
```suggestion
```

I think this is redundant right? It already happens here:
https://github.com/getsentry/sentry-dotnet/blob/eda625a36696264bd6751ada2003c061d22dc847/src/Sentry/Protocol/Trace.cs#L53

src/Sentry/Protocol/Trace.cs
```diff
@@ -50,6 +50,21 @@ internal set
     /// <inheritdoc />
     public bool? IsSampled { get; internal set; }
 
+    private Dictionary<string, object?> _data = new();
```
jamescrosswell, line 53: Could potentially back this by a `Lazy<T>` to avoid allocations when no Trace.Data is set.

We try to do this wherever possible - particularly for the trace classes as there can potentially be thousands of these allocated at any one time in larger applications.

src/Sentry/SentryTransaction.cs
```diff
@@ -339,6 +339,7 @@ public void AddBreadcrumb(Breadcrumb breadcrumb) =>
         _breadcrumbs.Add(breadcrumb);
 
     /// <inheritdoc />
+    [Obsolete("Add metadata to Contexts.Trace.SetData")]
```
jamescrosswell, line 342: Should we be doing this? I thought Extra was supposed to stay the way it is... or is this something you've seen it the docs or other SDKs?

jamescrosswell (changes requested):
Thanks @aritchie, couple of comments but Trace.Data is looking good now. 

We also need to add Span.Data 

Note that there is a bit of an equivalence between `SpanTracer` and `SentrySpan`... the `SpanTracer` class is the concrete implementation of `ISpan` that is used when traces are active. The `SentrySpan` class is the (mostly immutable) DTO that is used to serialise span data when sending events to Sentry. 

Similarly, there are both `TransactionTracer` and `SentryTransaction` classes. 

... just in case that wasn't obvious.

bruno-garcia:
@AbhiPrasad I believe you know why we're doing this and the expected outcome, right? Could you help with a review?

Also tagged @adinauer as this is quite similar to the Java implementation, maybe you have thoughts?

src/Sentry/SentryTransaction.cs
aritchie: Not sure what the answer is here?   @bruno-garcia

test/Sentry.Tests/SerializationTests.cs
```diff
@@ -0,0 +1,50 @@
...
+        var context = new TransactionContext("name", "operation", new SentryTraceHeader(SentryId.Empty, SpanId.Empty, false));
+        var transactionTracer = new TransactionTracer(hub, context);
+        var span = transactionTracer.StartChild("childop");
+        span.SetExtra("span1", "value1");
+
+        var transaction = new SentryTransaction(transactionTracer)
+        {
+            IsSampled = false
+        };
+        transaction.Contexts.Trace.SetData("transaction1", "transaction_value");
```
bruno-garcia, line 27: This could stay here but my understanding is that we're "replacing" `SetExtra` on Transaction (Not on Event) with `SetData`. But it doesn't serialize on the top level document like `extra` does, it goes under `contexts.trace.data`

src/Sentry/Protocol/Trace.cs
aritchie: It's a clone to ensure dictionary updates in the clone are not hitting the original collection.

src/Sentry/Protocol/Trace.cs
aritchie: So lazy is an allocation on its own and there is no computation here.  I can add this.  I could also set the initial size of the dictionary to 0 which is likely better.  Thoughts?

test/Sentry.Tests/SerializationTests.cs
aritchie: My original PR redirected SetExtra to Contexts.Trace.Data so this is easy to do.  Just not sure what you guys are looking for here.

src/Sentry/SentryTransaction.cs
cleptric: Deprecate extra and add `SetData` to `Span` and `Transaction` please.

src/Sentry/Protocol/Trace.cs
jamescrosswell: We could initialise the dictionary only if it's accessed... something like:
```csharp
Dictionary<string, object?>? _data = null;
Dictionary<string, object?> Data => _data ??= new();
```

src/Sentry/Protocol/Trace.cs
jamescrosswell: Argh... tricked by GitHub's review interface. It looked like this was in the SetData method:
image

I see it's actually in the `Clone` method!

test/Sentry.Tests/SerializationTests.cs
adinauer: Could we simply deprecate `setExtra` and add `setData` in the current major version?

Regarding changing where this goes customers may have filtering logic in place, e.g. in `beforeSend`/`beforeSendTransaction`. Since for Java the change coincided with us working on a major version (v8), we opted for changing this in a major so customers are more careful when upgrading and don't end up sending PII by accident. Here's the Java PR: https://github.com/getsentry/sentry-java/pull/3735 in case you want to compare something.

jamescrosswell: > Could we simply deprecate `setExtra` and add `setData` in the current major version?

This would be [a breaking change](https://learn.microsoft.com/en-gb/dotnet/core/compatibility/library-change-rules#members)... adding a new member to an interface would break any existing classes implementing that interface - and this would involve modifying multiple interfaces (for historical reasons). 

I think if it was a real showstopper, we could consider doing it. But in this case, changing the names of those methods from SetExtra to SetData doesn't really change anything except how SDK users use our SDK to add arbitrary data to traces and spans. It doesn't enable delivering any features that we want/need for Sentry. So it feels like a lot of work for very little gain.

I'd rather we do this in the next major release (v6.0) unless there is any compelling reason to do it now.

jamescrosswell: That's still a breaking change. `ISpanData : IHasData` still adds members to the `ISpanData` interface (which is implemented all over the place).

Again, if there's any compelling reason to rush this through now (like if it enables us to delivers some features Sentry needs) we can do it... but otherwise, why not just keep this in the backlog until version 6.0.0? From what I can see though, this is purely cosmetic.

So to avoid all confusion, we do this:
- **[Now]** Ensure additional data for Transactions gets transmitted via Transaction.Contexts.Trace.Data in the protocol 
- **[Now]** Additional data for Spans is already sent via Span.Data
- _[Later]_ Rename the methods that set that additional data from SetExtra to SetData

aritchie: @bruno-garcia @jamescrosswell  Bruno & I had a discussion about this.  Bruno agreed to the proposed workflow.  
Technically, it doesn't break the signatures for existing calls.  There is a new marker-ish interface to deal with the new methods.  

I'm fine with whatever.  I just want to put this "pr to bed"

jamescrosswell: @aritchie something we could do, which would not be a breaking change, would be to add a `SetData` extension method to `IHasExtra`, which simply called `SetExtra`. 

Eventually in version 6.0 we could make those methods on the class rather than extension methods.

aritchie: I did try that.  The issue was how much that interface touched.  It meant adding all of those methods to a lot of other classes.  Changing IHasExtra truly is a breaking change though because users that do implement it, would have to make updates to their models.  In this case, they only have to update their Spans (3 classes at the moment)

jamescrosswell: Not an interface method/member... An extension method. It would basically just be an alternate syntax for calling Set Extra (without changes to any interfaces or classes).

aritchie: They want the collection called Data as well.  I can't use an extension for the property.

src/Sentry/SentryTransaction.cs
```diff
@@ -179,12 +179,6 @@ public IReadOnlyList<string> Fingerprint
     /// <inheritdoc />
     public IReadOnlyCollection<Breadcrumb> Breadcrumbs => _breadcrumbs;
 
-    // Not readonly because of deserialization
-    private Dictionary<string, object?> _extra = new();
-
-    /// <inheritdoc />
-    public IReadOnlyDictionary<string, object?> Extra => _extra;
```
bruno-garcia, line 186: this will require a major bump. Can we just keep this instead?

src/Sentry/SpanTracer.cs
```diff
@@ -82,13 +82,21 @@ public void SetTag(string key, string value) =>
     public void UnsetTag(string key) =>
         (InternalTags ??= new ConcurrentDictionary<string, string>()).TryRemove(key, out _);
 
-    private readonly ConcurrentDictionary<string, object?> _data = new();
+    private Dictionary<string, object?>? _data;
 
     /// <inheritdoc />
-    public IReadOnlyDictionary<string, object?> Extra => _data;
+    public IReadOnlyDictionary<string, object?> Data =>
```
bruno-garcia, line 88: This will be accessed concurrently so needs synchronization. Keep the original implementation of map?

aritchie: Yup - that's what I get for mass copy/pasting on this one.  Thanks for the spot.

bruno-garcia (approved):

jamescrosswell (approved):

Issue #3924, opened 2025-01-29
Profiling - `unknown` stacks on net8


vaind:
On the current `main` branch, our `Sentry.Samples.Console.Profiling` gets captured with partially missing stack trace information. 

I've debugged EventPipe capture & TraceLog and bisected historical releases and it turns out this is caused by an upgrade to `net8`. 
When I try the sample on v4.13.0 branch it works fine. After changing the target framework in the sample to net8, it breaks and we get these missing stack traces. 
Additionally, I've found out that increasing the `CircularBufferMB` in `StartEventPipeSession` to a higher value helps with the issue on the 4.13 branch, indicating it's caused by the initial rundown being too large so some information is dropped. However, the same "fix" doesn't work on 5.0 or `main` so I'm continuing investigation.

vaind:
Looks like something changed in net8 that broke our usage of EventPipe. Raised an issue here: https://github.com/microsoft/perfview/issues/2155 to try and figure out how to resolve this.

PR #3941, opened 2025-02-06
chore: profiling improvements

vaind:
Some improvements I've done while investigating #3924 - the only visible thing is deduplicating stack frames that do have a different code address but are actually the same method so as far as the UI is concerned, they're the same frame.


jamescrosswell:
> deduplicating stack frames that do have a different code address but are actually the same method so as far as the UI is concerned

@vaind any idea how/why the same method ends up with multiple frames on the stack? That implies the method actually gets called twice right (unless one of them is a wrapper)?

jamescrosswell (approved):

vaind:
> why the same method ends up with multiple frames on the stack

Not on the same stack.
On different samples, we get stacks with different code addresses at the top but they map to the same method, just different lines.

PR #3942, opened 2025-02-06
fix: net8 unknown stack trace methods for JIT methods

### LINKED ISSUE #3924: Profiling - `unknown` stacks on net8

vaind:
Fixes #3924

### DIFF
--- a/src/Sentry.Profiling/SampleProfilerSession.cs
+++ b/src/Sentry.Profiling/SampleProfilerSession.cs
@@ -16,14 +16,16 @@ internal class SampleProfilerSession : IDisposable
     private readonly IDiagnosticLogger? _logger;
     private readonly SentryStopwatch _stopwatch;
     private bool _stopped = false;
+    private Task _processing;
 
-    private SampleProfilerSession(SentryStopwatch stopwatch, EventPipeSession session, TraceLogEventSource eventSource, IDiagnosticLogger? logger)
+    private SampleProfilerSession(SentryStopwatch stopwatch, EventPipeSession session, TraceLogEventSource eventSource, Task processing, IDiagnosticLogger? logger)
     {
         _session = session;
         _logger = logger;
         EventSource = eventSource;
         _sampleEventParser = new SampleProfilerTraceEventParser(EventSource);
         _stopwatch = stopwatch;
+        _processing = processing;
     }
 
     // Exposed only for benchmarks.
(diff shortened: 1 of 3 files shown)


src/Sentry.Profiling/SampleProfilerSession.cs
```diff
@@ -22,21 +21,21 @@ private SampleProfilerSession(SentryStopwatch stopwatch, EventPipeSession sessio
...
     internal static EventPipeProvider[] Providers = new[]
     {
-        // Note: all events we need issued by "DotNETRuntime" provider are at "EventLevel.Informational"
-        // see https://learn.microsoft.com/en-us/dotnet/fundamentals/diagnostics/runtime-events
-        // TODO replace Keywords.Default with a subset. Currently it is:
-        //   Default = GC | Type | GCHeapSurvivalAndMovement | Binder | Loader | Jit | NGen | SupressNGen
-        //                | StopEnumeration | Security | AppDomainResourceManagement | Exception | Threading | Contention | Stack | JittedMethodILToNativeMap
-        //                | ThreadTransfer | GCHeapAndTypeNames | Codesymbols | Compilation,
-        new EventPipeProvider(ClrTraceEventParser.ProviderName, EventLevel.Informational, (long) ClrTraceEventParser.Keywords.Default),
+        new EventPipeProvider(ClrTraceEventParser.ProviderName, EventLevel.Verbose, (long) (
```
vaind, line 32: `EventLevel.Verbose` is the actual fix here. The rest of the change is just an improvement so we don't receive things we don't need.

jamescrosswell (approved):
👍🏻

bruno-garcia:
This got reverted but it's in the changelog: https://github.com/getsentry/sentry-dotnet/compare/7faf221b6fbf3c293052c30429b38b75564380c4...refs/heads/release/5.1.1

vaind:
> This got reverted but it's in the changelog: [7faf221...refs/heads/release/5.1.1](https://github.com/getsentry/sentry-dotnet/compare/7faf221b6fbf3c293052c30429b38b75564380c4...refs/heads/release/5.1.1)

yeah these changelog items got accidently pulled back in when James was updating the revert PR and pulled changes from main and while I've noticed this yesterday and fixed it  in another PR, I didn't check whether there's already a release triggered.

Issue #3921, opened 2025-01-28
Implement Sampling Seed Propagation


bitsandfoxes:
See https://develop.sentry.dev/sdk/telemetry/traces/#propagated-random-value

The goal is to propagate the random value in the context of distributed tracing and aims to improve trace quality when using a custom `tracesSampler`.

### Motivation 

A trace is complete when all of its members are sampled. A "sub-trace" is complete when all of its descendents are sampled.

Ordinarily, Trace and Logging SDKs configure parent-based samplers which decide to sample based on the Context, because it leads to completeness. However, when non-root spans or logs make independent sampling decisions, instead of using the parent-based approach, incompleteness may result.

Consistent probability sampling requires that for any span in a given trace, if a Sampler with lesser sampling probability selects the span for sampling, then the span would also be selected by a Sampler configured with greater sampling probability.

This is achieved by propagating not only the sampling decision, but also the inputs used to make that decision, in the Dynamic Sampling Context so that downstream nodes can make consistent sampling decisions. 

### Example

Imagine a trace from machines A -> B -> C.

The respective sample rates are 0.5 -> 0.1 -> 0.3

On machine A (the start of the trace) a random number is generated in the range [0, 1) and assigns this to `sample_rand`, which is then propagated and used by all members of the trace for their sampling decision. Let's assume that number is 0.2345. 
Machine A: Sets `sampled = 0.2345 < 0.5` = `true`
Machine B: Sets `sampled = 0.2345 < 0.1` = `false`
Machine C: Sets `sampled = 0.2345 < 0.3` = `true`

Propagating the `sample_rand` and having all members use this in sampling decisions ensures that whenever a member with a lower sampling rate (like machine C in this case) sample in, members with higher sample rates (like machine A in this case) also sample in. 

In this example, Machine B didn't sample in and so we don't get a complete trace. However if machine B did sample in, Machines A and C would also be guaranteed to... so we always get complete traces whenever the member with the lowest sample rate samples in.

The only other way to guarantee complete traces is by having all members sample at the same rate (e.g. 0.5 or 0.1)... which is basically what you get when the parent decides for everyone. Seed propagation doesn't guarantee complete traces - it maximises the chances of a complete trace when different members participating in a trace are all maxing their own sampling decisions (either with a custom sample rate or a custom trace sampling function).

### References
- https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/processor/probabilisticsamplerprocessor/README.md

[...]

PR #3951, opened 2025-02-10
Propagate Sampling Seed

jamescrosswell:
Resolves https://github.com/getsentry/sentry-dotnet/issues/3921

Implementation in other SDKs:
- Dart: https://github.com/getsentry/sentry-dart/pull/2629
- PHP: https://github.com/getsentry/sentry-php/pull/1793
- Python: https://github.com/getsentry/sentry-python/pull/4016

### Progress
- [x] Implement hash algorithm to convert Trace ID to a seed that can be used with .NET's Random class 
- [x] Ensure `sentry-sample_rand` is read from the DSC and PropagationContext (TwP)
- [x] Generate `sample_rand` when none is provided in BaggageHeader
- [x] Generate/store `sample_rand` when creating new Transactions
- [x] Use `sample_rand` in sampling decision
- [x] Ensure `sentry-sample_rand` gets propagated in trace headers


CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 ## Unreleased
 
+### Features
+- The sample seed used for sampling decisions is now propagated, for use in downstream custom trace samplers ([#3951](https://github.com/getsentry/sentry-dotnet/pull/3951))
```
```suggestion

- The sample seed used for sampling decisions is now propagated, for use in downstream custom trace samplers ([#3951](https://github.com/getsentry/sentry-dotnet/pull/3951))
```

src/Sentry/DynamicSamplingContext.cs
```diff
@@ -44,7 +46,12 @@ private DynamicSamplingContext(
             throw new ArgumentOutOfRangeException(nameof(sampleRate));
         }
 
-        var items = new Dictionary<string, string>(capacity: 7)
+        if (sampleRand is < 0.0 or >= 1.0)
+        {
+            throw new ArgumentOutOfRangeException(nameof(sampleRand));
```
bruno-garcia, line 51:
```suggestion
            throw new ArgumentOutOfRangeException(nameof(sampleRand), "Arg invalid if < 0.0 or >= 1.0");
```

src/Sentry/DynamicSamplingContext.cs
```diff
@@ -61,6 +68,11 @@ private DynamicSamplingContext(
             items.Add("sample_rate", sampleRate.Value.ToString(CultureInfo.InvariantCulture));
         }
 
+        if (sampleRand is not null)
+        {
+            items.Add("sample_rand", sampleRand.Value.ToString("N4", CultureInfo.InvariantCulture));
```
bruno-garcia, line 73: we want this even if we have `sample_rate` in? see line 68

src/Sentry/DynamicSamplingContext.cs
```diff
@@ -111,6 +123,27 @@ private DynamicSamplingContext(
             return null;
         }
 
+        // See https://develop.sentry.dev/sdk/telemetry/traces/#propagated-random-value
+        if (items.TryGetValue("sample_rand", out var sampleRand))
```
bruno-garcia, line 127: odd that we validate this to be within a range as a float but then we make it a string then try to parse it back out.
Can we rely on the original value and avoid the parsing instead?

src/Sentry/DynamicSamplingContext.cs
```diff
@@ -111,6 +123,27 @@ private DynamicSamplingContext(
...
+        {
+            if (!double.TryParse(sampleRand, NumberStyles.Float, CultureInfo.InvariantCulture, out var rand) ||
+                 rand is < 0.0 or >= 1.0)
```
bruno-garcia, line 130: didn't we validate this above before adding it to `items`? or it could already be in items when the method was caled? in which case wouldn't we be adding it twice? (I add Add so assume it's not a dictionary)

bruno-garcia (commented):
only got as far as 2 files

src/Sentry/DynamicSamplingContext.cs
jamescrosswell: We do. This is strictly required for the DSC to be valid: https://develop.sentry.dev/sdk/telemetry/traces/dynamic-sampling-context/#dsc-specification

jamescrosswell: I believe so yes. The docs developer docs indicate:
> If sample_rate (inside baggage) and the sampling decision (trailing -1 or -0 from the sentry-trace header) are present in the incoming trace, create sample_rand so that when compared to the incoming sample_rate it would lead to the same sampling decision that is in the sentry-trace header. 

Which we're doing... and then we propagate the `sample_rand` so that if any downstream nodes can use this in their sampling decision.

src/Sentry/DynamicSamplingContext.cs
jamescrosswell: This method is creating a DSC from the baggage header. The BaggageHeader is a pretty simple class that just pulls stuff directly off a certain HTTP header, without any strong opinions about what the items in that header should be. Other vendors might be putting things on that header so we just propagate it, pretty much verbatim (we might add things to it but we never remove things from it).

But we can't control what will be on that baggage header in this SDK. A request could be made to our SDK with a malformed sentry-sample_rand item in the BaggageHeader.

src/Sentry/DynamicSamplingContext.cs
jamescrosswell: I'm not sure I follow. It may or may not be present as a string in the baggage header. If it's present then here we try to cast it to a float and ensure it's in the correct range. 

If it's present but malformed then we don't create the DSC from the baggage header.

bruno-garcia:
added @cleptric and @buenaflor to check alignment with the other SDKs

aritchie (approved):

src/Sentry/BaggageHeader.cs
```diff
@@ -26,7 +26,7 @@ private BaggageHeader(IEnumerable<KeyValuePair<string, string>> members) =>
 
     // We can safely return a dictionary of Sentry members, as we are in control over the keys added.
     // Just to be safe though, we'll group by key and only take the first of each one.
-    internal IReadOnlyDictionary<string, string> GetSentryMembers() =>
+    internal Dictionary<string, string> GetSentryMembers() =>
```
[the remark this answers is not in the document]
jamescrosswell: I changed it because I need to add the sample_rand to the dictionary if it's not present already:
https://github.com/getsentry/sentry-dotnet/blob/724880d31c075b292a32a6cbe6a5848e03948e18/src/Sentry/DynamicSamplingContext.cs#L145

Publicly, it's still exposed as a ReadOnlyDictionary.

PR #3956, opened 2025-02-11
Fix #2710 - Return macOS instead of 'Darwin' and return proper version

### LINKED ISSUE #2710: macOS version not reported
example event: https://sentry-sdks.sentry.io/issues/4536804443/?project=5428537&query=is%3Aunresolved&referrer=issue-stream&statsPeriod=1h&stream_index=1

image

Seems like macOS version has regressed

aritchie:
Resolves #2710 

Enricher will now check if running in a mac environment (runtime as we don't target mac targetframework) and switch "Darin" to "macOS" as well as set the operating system version properly

NOTES:
- The native library was using compiler checks.  This was an easier route for the dotnet repo instead of pinvoking. 
- Adds a PlatformFact for xunit, so we can skip tests based on platform


CHANGELOG.md
```diff
@@ -8,6 +8,7 @@
 - Fixed envelopes with oversized attachments getting stuck in __processing ([#3938](https://github.com/getsentry/sentry-dotnet/pull/3938))
 - Unknown stack frames in profiles on .NET 8+ ([#3942](https://github.com/getsentry/sentry-dotnet/pull/3942))
 - Deduplicate profiling stack frames ([#3941](https://github.com/getsentry/sentry-dotnet/pull/3941))
+- OperatingSystem will now return macOS as OS name instead of 'Darwin' as well as the proper version. ([#2710](https://github.com/getsentry/sentry-dotnet/issues/2710)
```
jamescrosswell, line 11: @aritchie the convention in this change log is to reference the PR (not the issue)... which is why the Danger check fails.

jamescrosswell (commented):
LGTM as soon as all the CI checks are green. 

Thanks @aritchie !

bruno-garcia (approved):
ideally would go on a major bump because technically breaks saved search and tag distribution based on that OS tag. But I'd say unlikely to be a real disruption and the value of the change is worth the risk.

Issue #735, opened 2021-01-13
Add onCrashedLastRun


philipphofmann:
Add a callback onCrashedLastRun to SentryOptions that is called by the SDK passing the event
when Sentry is initialized and the last program execution terminated with a crash.

PR in Cocoa: getsentry/sentry-cocoa#808
Docs in Cocoa: https://docs.sentry.io/platforms/apple/enriching-events/user-feedback/
The issue on Java: https://github.com/getsentry/sentry-java/issues/1125

dalux-era:
Hi. 
Could this feature get a prioritization bump? A lot of people are switching from AppCenter to Sentry. It would definitely help migrating for all the .NET projects that had this feature in AppCenter - Crashes.HasCrashedInLastSessionAsync().

bruno-garcia:
@aritchie this could be added just for iOS and Android if that makes it easier, as it just binds to the native sdks. 

If you have a chance to peek at the code, any idea of the effort to add this? 


@vaind @bitsandfoxes did we get this in Unity to be a reference?

aritchie:
@bruno-garcia I'm actually already working on this over at #3958 for iOS.  It already exists in part on Android.

jamescrosswell:
> [@bruno-garcia](https://github.com/bruno-garcia) I'm actually already working on this over at [#3958](https://github.com/getsentry/sentry-dotnet/pull/3958) for iOS. It already exists in part on Android.

Just bumped into [this](https://github.com/dotnet/maui/discussions/653#discussioncomment-12116908) btw - which might be an interesting workaround if we don't have access to the file system. Although, I'm not sure what happens to performance if we try to store an envelope as a preference.

vaind:
>  did we get this in Unity to be a reference?

Yeah, for example on iOS there's a bool function exposed so it's very simple to do: https://github.com/getsentry/sentry-unity/blob/0059c77876e3746d50998af5b9ca8c6adbff6f52/package-dev/Plugins/iOS/SentryNativeBridge.m#L49

Issue #3944, opened 2025-02-06
Add option to enable/disable EXC_BAD_ACCESS suppression


jamescrosswell:
In https://github.com/getsentry/sentry-dotnet/pull/3909 we added some code to suppress `EXC_BAD_ACCESS` mach exceptions in the CocoaSDK, as these frequently duplicate a managed `NullReferenceException`.

We did some similar work in https://github.com/getsentry/sentry-dotnet/pull/3903 on Android. However for Android we added an option to let SDK users enable/disable suppression. For consistency, I think we should add a similar option to control `EXC_BAD_ACCESS` suppression.

As with the Android functionality, this will also need documenting. Possibly we could create some shared documentation explaining the necessity for the existence of this functionality and then some platform specific stuff detailing how to enable/disable it on iOS/Android.

aritchie:
Completed with PR https://github.com/getsentry/sentry-dotnet/pull/3958

jamescrosswell:
@aritchie let's keep the issue open until the PR is merged.

jamescrosswell:
Github has some strange logic for closing issues. I think this one is only closed once both related PRs are merged... reopening until #3998 gets merged.

PR #3958, opened 2025-02-11
Map iOS Native BeforeSend & OnCrashedLastEvent to SentryOptions.Native

### LINKED ISSUE #2102: Support iOS BeforeSend and OnCrashedLastRun
For iOS and MacCatalyst targets, the bundled Sentry Cocoa SDK has the `BeforeSend` and `OnCrashLastRun` options, both of which should probably be exposed via .NET.  They require conversion between the .NET SDK's `SentryEvent` and the Cocoa SDK's `SentryEvent` objects.

For Android, we do this via serialization/deserialization.  However the Cocoa SDK doesn't currently offer deserialization for `SentryEvent`, so a similar approach would not work.

The object graphs are quite large, so implementing the conversions manually will take time and a lot of tests.

This was originally planned for MAUI GA, but is deferred for now.

# Related issues

- https://github.com/getsentry/sentry-dotnet/issues/3621
- https://github.com/getsentry/sentry-dotnet/issues/3854

aritchie:
Resolves #2102


src/Sentry/Platforms/Cocoa/SentrySdk.cs
```diff
@@ -88,27 +89,44 @@ private static void InitSentryCocoaSdk(SentryOptions options)
...
-
-        // if (options.Native.OnCrashedLastRun is { } onCrashedLastRun)
-        // {
-        //     nativeOptions.OnCrashedLastRun = evt =>
-        //     {
-        //         var sentryEvent = evt.ToSentryEvent(nativeOptions);
-        //         onCrashedLastRun(sentryEvent);
-        //     };
-        // }
+        if (options.Native.BeforeSend is { } beforeSend)
```
jamescrosswell, line 92: I think this needs reconciling/merging with:
https://github.com/getsentry/sentry-dotnet/blob/17790c77a123806f732d7563543ecd3d5ae84cfc/src/Sentry/Platforms/Cocoa/SentrySdk.cs#L169-L201

aritchie: Done - I made sure this code runs before the user BeforeSend

aritchie:
@jamescrosswell @bruno-garcia Native exceptions obviously teardown the MAUI apps.  I can't unit test this properly due to needing to shutdown the app.  Thoughts?  Any internal mechanism I can call in native to have it "pretend" an issue occurred previously and force run BeforeSend/CrashedLastRun

bitsandfoxes:
Over at https://github.com/getsentry/sentry-unity/ we're covering this via smoketests.
Two runs:
1. Running and crashing on purpose
2. Running again and expecting the `crashedLastRun` to be `true`

jamescrosswell:
> @jamescrosswell @bruno-garcia Native exceptions obviously teardown the MAUI apps. I can't unit test this properly due to needing to shutdown the app. Thoughts? Any internal mechanism I can call in native to have it "pretend" an issue occurred previously and force run BeforeSend/CrashedLastRun

I think the `CrashedLastRun` mechanism will ultimately be checking to see if there's an envelope for a crash report somewhere right? So that we could fake/mock?

aritchie:
> I think the CrashedLastRun mechanism will ultimately be checking to see if there's an envelope for a crash report somewhere right? So that we could fake/mock?

Can't really fake/mock the native layer though.  This is smoke tested in the iOS sample app.

src/Sentry/Platforms/Cocoa/Extensions/SentryEventExtensions.cs
```diff
@@ -18,29 +19,62 @@ internal static class SentryEventExtensions
...
+            return null;
 
         using var json = JsonDocument.Parse(stream);
         var exception = sentryEvent.Error == null ? null : new NSErrorException(sentryEvent.Error);
         var ev = SentryEvent.FromJson(json.RootElement, exception);
         return ev;
     }
 
-    public static CocoaSdk.SentryEvent ToCocoaSentryEvent(this SentryEvent sentryEvent, SentryOptions options, SentryCocoaSdkOptions nativeOptions)
+    public static CocoaSdk.SentryEvent ToCocoaSentryEvent(this SentryEvent sentryEvent, SentryOptions options)
```
jamescrosswell, line 34: Possibly a good use case for [Mapperly](https://github.com/riok/mapperly)?

It's kind of like AutoMapper but uses Source Generators. So would basically, I hope, generate code like what you have below for us... and keep this maintained if/when new properties get added. 

Hard to tell whether it would save us time or add complexity and take more time though. If we had **_a lot_** of mapping in the SDK then I'd say definitely check it out... just for one or two situations, maybe we're better off hand coding it and writing up some tests to spot when things need updating.

aritchie: I have used mapperly and its good for lots situations.  I don't believe these contracts change enough or are big enough to warrant it.  Some of the native types of tricky anyhow.  My thinking is to just "pound this one out"

aritchie:
This PR also now resolves #735 and #3944

CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 ## Unreleased
 
+### Fixes
+
+- Map iOS native events - OnCrashedLastRun and BeforeSend ([#2102](https://github.com/getsentry/sentry-dotnet/pull/3958))
```
jamescrosswell, line 7: @aritchie this is what we did to implement this feature. What does it mean for the user though? Why should they care?

This isn't necessarily a fix either. It's just new functionality, AFAIK. I'd probably put under `Features`.

CHANGELOG.md
```diff
@@ -25,6 +28,8 @@
 - Native SIGSEGV errors resulting from managed NullReferenceExceptions are now suppressed on Android ([#3903](https://github.com/getsentry/sentry-dotnet/pull/3903))
 - OTel activities that are marked as not recorded are no longer sent to Sentry ([#3890](https://github.com/getsentry/sentry-dotnet/pull/3890))
 - Fixed envelopes with oversized attachments getting stuck in __processing ([#3938](https://github.com/getsentry/sentry-dotnet/pull/3938))
+- Unknown stack frames in profiles on .NET 8+ ([#3942](https://github.com/getsentry/sentry-dotnet/pull/3942))
```
jamescrosswell, line 31: Why are we seeing this in this PR?

samples/Sentry.Samples.Ios/Sentry.Samples.Ios.csproj
```diff
@@ -1,7 +1,7 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFramework>net9.0-ios18.0</TargetFramework>
+    <TargetFramework>net9.0-ios</TargetFramework>
```
jamescrosswell, line 4: Although technically we could do this in our Samples, it might break our builds if the Workloads change. Is there any strong reason to remove the `ios18.0` suffix?

src/Sentry/Platforms/Cocoa/BindableSentryOptions.cs
```diff
@@ -26,6 +26,7 @@ public class NativeOptions
         public bool? EnableUIViewControllerTracing { get; set; }
         public bool? EnableUserInteractionTracing { get; set; }
         public bool? EnableTracing { get; set; }
+        public bool? SuppressSegfaults { get; set; }
```
jamescrosswell, line 29: I think the SuppressSegfaults stuff should go in separate PR:
- It needs it's own entry in the changelog (letting people know about the option, and the fact that the default is changing vs 5.1.1)
- Easier to review both PRs 
- Easier to work out what had to change to enable a feature/fix , if someone comes back to this PR in 12-18 months
- Easier to revert, if we need to

It could be done as a fork of this branch, and we review/merge after this branch is reviewed/merged... but definitely tidier if it's in it's own PR.

src/Sentry/Platforms/Cocoa/Extensions/CocoaExtensions.cs
```diff
@@ -169,6 +169,24 @@ public static NSDictionary<NSString, NSObject> ToNSDictionary<TValue>(
                 d.Keys.ToArray());
     }
 
+    public static NSDictionary<NSString, NSString> ToNSDictionaryStrings(
```
jamescrosswell, line 172: Maybe worth writing some simple unit tests for this?

src/Sentry/Platforms/Cocoa/Extensions/CocoaExtensions.cs
```diff
@@ -224,4 +242,101 @@ public static object ToObject(this NSNumber n)
                 $"NSNumber \"{n.StringValue}\" has an unknown ObjCType \"{n.ObjCType}\" (Class: \"{n.Class.Name}\")")
         };
     }
+
+
+    public static SentryEvent? ToSentryEvent(this CocoaSdk.SentryEvent sentryEvent)
```
jamescrosswell, line 247: Unit tests?

src/Sentry/Platforms/Cocoa/SentrySdk.cs
```diff
@@ -35,6 +35,7 @@ private static void InitSentryCocoaSdk(SentryOptions options)
         nativeOptions.SendDefaultPii = options.SendDefaultPii;
         nativeOptions.SessionTrackingIntervalMillis = (nuint)options.AutoSessionTrackingInterval.TotalMilliseconds;
 
+        // nativeOptions.BeforeSend
```
jamescrosswell, line 38:
```suggestion
```

src/Sentry/Platforms/Cocoa/SentrySdk.cs
```diff
@@ -86,29 +87,111 @@ private static void InitSentryCocoaSdk(SentryOptions options)
...
+                return evt;
+
+            try
+            {
+                var sentryEvent = evt.ToSentryEvent();
+                if (sentryEvent == null)
+                    return evt;
+
+                var result = options.BeforeSendInternal(sentryEvent, null!);
+                // returning null is fine - the native binding even has this set, but the tooling doesn't want to obey, so the pragma was necessary
```
jamescrosswell, line 138:
```suggestion
```

src/Sentry/Platforms/Cocoa/SentrySdk.cs
```diff
@@ -86,29 +87,111 @@ private static void InitSentryCocoaSdk(SentryOptions options)
...
+                if (result == null)
+                    return null!;
+
+                evt.ServerName = result.ServerName;
```
jamescrosswell, line 142: I think the code on the following lines updates the subset of properties that might be modified by the `OnBeforeSend` callback right? Maybe worth adding that context as a comment (I'm only aware of it because I listened in on the chat with @bruno-garcia).

We do try to let the code speak for itself and have/require as few comments as possible. In this case, I think it's some non-obvious requirements we landed on after a bit of back and forth though.

src/Sentry/SentryEvent.cs
```diff
@@ -290,7 +290,7 @@ internal static SentryEvent FromJson(JsonElement json, Exception? exception)
     {
         var modules = json.GetPropertyOrNull("modules")?.GetStringDictionaryOrNull();
         var eventId = json.GetPropertyOrNull("event_id")?.Pipe(SentryId.FromJson) ?? SentryId.Empty;
-        var timestamp = json.GetPropertyOrNull("timestamp")?.GetDateTimeOffset();
+        var timestamp = json.GetSafeDateTimeOffset("timestamp"); // Native sentryevents are serialized to epoch timestamps
```
jamescrosswell, line 293: Given this is a FromJson method, are we sure it will always be reading JSON that's been serialised by the native SDK? Might it not also be used to read JSON serialised by the managed SDK?

jamescrosswell (commented):
Generally looking good. 

I think we might be able to refactor the code dealing with the OnBeforeSend event into a BeforeSendWrapper, [like what was done for Android](https://github.com/getsentry/sentry-dotnet/blob/4e7c3e44d7055041a4840be01993cee56701ed04/src/Sentry/Platforms/Android/SentrySdk.cs#L173-L191). That would make it easier/possible to write unit tests for all of that logic (and prevent regressions in the future).

src/Sentry/SentryEvent.cs
aritchie: I handle both the native and managed cases in that new method.  This method was being used by the native deserialization calls.

samples/Sentry.Samples.Ios/Sentry.Samples.Ios.csproj
aritchie: Apps do not obey the versioning here.  This is only need for libraries.  This also breaks rider detection of the app.

src/Sentry/SentryEvent.cs
jamescrosswell: Shouldn't we handle both here as well? It's a public method.

aritchie: GetSafeDateTimeOffset handles both epoch timestamp and standard string based datetimes

src/Sentry/Platforms/Cocoa/BindableSentryOptions.cs
jamescrosswell: Moved to https://github.com/getsentry/sentry-dotnet/pull/3998:
- https://github.com/getsentry/sentry-dotnet/pull/3998

test/Sentry.Tests/Platforms/iOS/NativeSerializationTests.cs
```diff
@@ -0,0 +1,106 @@
...
+using Foundation;
+using Sentry.Cocoa.Extensions;
+using Xunit;
+
+namespace Sentry.Tests.Platforms.iOS;
+
+public class NativeSerializationTests
+{
+    [Fact]
+    public void Managed_To_Native()
```
jamescrosswell, line 13:
```suggestion
    public void CopyToCocoaSentryEvent_CopiesProperties()
```

See contributor docs on [naming tests](https://github.com/getsentry/sentry-dotnet/blob/main/CONTRIBUTING.md#naming-tests)... just a convention for this repo basically.  I'm not sure exactly what you want to call this test - the above is just a suggestion/idea.

test/Sentry.Tests/Platforms/iOS/NativeSerializationTests.cs
```diff
@@ -0,0 +1,106 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using Foundation;
+using Sentry.Cocoa.Extensions;
+using Xunit;
+
+namespace Sentry.Tests.Platforms.iOS;
+
+public class NativeSerializationTests
```
jamescrosswell, line 10:
```suggestion
public class CocoaExtensionsTests
```

Just to keep things tidy, we usually have the name of the test classes match the class exposing the functionality (only with a `Tests` suffix). That way it's easy to determine whether any tests exist for a particular class/method or not.

src/Sentry/Internal/Extensions/JsonExtensions.cs
```diff
@@ -213,6 +213,30 @@ public static void Deconstruct(this JsonProperty jsonProperty, out string name,
         return double.Parse(json.ToString()!, CultureInfo.InvariantCulture);
     }
 
+    /// <summary>
+    /// Safety value to deal with native serialization - allows datetimeoffset to come in as a long or string value
+    /// </summary>
+    /// <param name="json"></param>
+    /// <param name="propertyName"></param>
+    /// <returns></returns>
+    public static DateTimeOffset? GetSafeDateTimeOffset(this JsonElement json, string propertyName)
```
jamescrosswell, line 222: This one would be easy to unit test right?

jamescrosswell (changes requested):
Almost ready to merge I think. The only thing I'd like to see is for [this block of code](https://github.com/getsentry/sentry-dotnet/blob/63326ecb9c23501bf192da6b40dc56b8ecfc8ab4/src/Sentry/Platforms/Cocoa/SentrySdk.cs#L89-L152) to be more testable.

It should be pretty trivial to write unit tests for the new `JsonExtensions` as well (and I think we should).

I can push up a commit to address the tests, if necessary. I think it's better if we write tests for our own code though.

src/Sentry/SentryEvent.cs
```diff
@@ -286,6 +286,24 @@ public void WriteTo(Utf8JsonWriter writer, IDiagnosticLogger? logger)
     /// </summary>
     public static SentryEvent FromJson(JsonElement json) => FromJson(json, null);
 
+    private static SentryLevel? SafeLevelFromJson(JsonElement json)
+    {
+        var levelString = json.GetPropertyOrNull("level")?.GetString();
+        if (levelString == null)
+            return null;
+
+        // Native SentryLevel.None does not exist in dotnet
```
jamescrosswell, line 295: Do we know what the behaviour of SentryLevel.None is? Does it mean nothing gets logged? Is that was setting the level = null does?

src/Sentry/SentryEvent.cs
```diff
@@ -286,6 +286,24 @@ public void WriteTo(Utf8JsonWriter writer, IDiagnosticLogger? logger)
...
+        return levelString.ToLowerInvariant() switch
+        {
+            "debug" => SentryLevel.Debug,
+            "info" => SentryLevel.Info,
+            "warning" => SentryLevel.Warning,
+            "fatal" => SentryLevel.Fatal,
+            "error" => SentryLevel.Error,
+            _ => null
```
jamescrosswell, line 303: I take it this is intentional (we don't want to throw an exception but we can't conveniently log the irregularity anywhere)?

PR #3962, opened 2025-02-12
revert: profiling changes

vaind:
There are some crashes in CI, needs further investigation what caused this. For now, we revert so that we can release.

- **Revert "Skip flaky test (EventPipeSession_ReceivesExpectedCLREvents) (#3949)"**
- **Revert "fix: net8 unknown stack trace methods for JIT methods (#3942)"**
- **Revert "chore: profiling improvements (#3941)"**

#skip-changelog

PR #3967, opened 2025-02-13
fix: profiling - unknown stacktraces on net8

### LINKED ISSUE #3924: Profiling - `unknown` stacks on net8

vaind:
Fixes #3924 

This re-lands an updated version #3942 after it was reverted in #3962 
I believe the cause of the crash in CI was the new test case. I've updated it and it seems stable when testing locally.


jamescrosswell (approved):

CHANGELOG.md
```diff
@@ -30,6 +36,7 @@
 ### Fixes
 
 - Fix mismapped breadcrumb levels coming in from native to dotnet SDK ([#3993](https://github.com/getsentry/sentry-dotnet/pull/3993))
+- Deduplicate profiling stack frames ([#3969](https://github.com/getsentry/sentry-dotnet/pull/3969))
```
bruno-garcia, line 39: 3969 landed on 5.2.0 based on this diff: 

https://github.com/getsentry/sentry-dotnet/compare/5.1.1...5.2.0

bruno-garcia (approved):

PR #3970, opened 2025-02-13
perf: subscribe only to necessary CLR events

vaind:
This re-lands a part of #3942 after it was reverted in #3962 

Depends on #3967 so merge that first please and then I'll update this & resolve the merge conflict. Otherwise that PR would look weird if we added "Verbose" level in this one first.


bruno-garcia (approved):
Got merged so unblcoked: https://github.com/getsentry/sentry-dotnet/pull/3967

PR #3972, opened 2025-02-17
WIP: Unsampled Transactions to reduce memory pressure

jamescrosswell:
WIP... really needs to be done after https://github.com/getsentry/sentry-dotnet/pull/3951 gets merged as it's fiddling with the same sections of code (and needs to take into account the sample_rand etc. as well). 

Not sure if this is worth finishing since we eventually plan to get rid of transactions entirely.

### DIFF
--- a/src/Sentry/Internal/Hub.cs
+++ b/src/Sentry/Internal/Hub.cs
@@ -127,8 +127,6 @@ internal class Hub : IHub, IDisposable
         IReadOnlyDictionary<string, object?> customSamplingContext,
         DynamicSamplingContext? dynamicSamplingContext)
     {
-        var transaction = new TransactionTracer(this, context);
-
         // If the hub is disabled, we will always sample out.  In other words, starting a transaction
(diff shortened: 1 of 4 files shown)


jamescrosswell:
> One concern I have is; Does this affect our ability to create and propagate trace ids downstream? or to continue a trace (on aspnetcore) if a transaction wasn't sampled?

I don't think so no... I thought about that when creating the UnsampledTransaction class (making sure trace propagation still works). Some further testing could be done to double/triple check this. 

I guess it's lots of questions like this that make me a bit nervous - it's the sort of change that really requires thorough testing before rolling it out. In an ideal world, that would all be covered by unit tests, but I'm not sure that's the case.

bruno-garcia:
We should really ship this btw

jamescrosswell:
> We should really ship this btw

I'll bump the priority and pick it back up again... probably/hopefully just needs to be updated from main 🤞🏻

Issue #3605, opened 2024-09-10
Support for Feedback Events


jamescrosswell:
_Originally posted by @jamescrosswell in [#3604](https://github.com/getsentry/sentry-dotnet/issues/3604#issuecomment-2342062443)_

The .NET SDK is still implementing User Feedback using a `USER_REPORT_ENVELOPE` which means feedback has to be associated with exception events (an eventid is required in the envelope).

Since then a new [Feedback Events](https://develop.sentry.dev/application/feedback-architecture/#feedback-events) format has been introduced that allows User Feedback to be sent independently of exceptions. The .NET SDK.  should also support Feedback Events. 

This involved a new function: `CaptureFeedback`. 

This function is the base functionality needed to allow capturing user feedback while including tags, `replay_id`, `trace_id`, etc.
It's part of the 'new user feedback' product (https://github.com/getsentry/sentry/issues/64671) that was shipped for Web. That included a built-in widget, that allows screenshot attachment, and under the hood relies on this new function that deprecates the old `CaptureUserFeedback` which required an `event_id`. The new function optionally takes a related `event_id` and will contain relevant context stand alone.

This means a new envelope format for feedback was introduced. 
The new Feedback product architecture has thoroughly documented by @aliu39 here:

* https://develop.sentry.dev/application/feedback-architecture/

This new function will allow outgoing feedbacks to be tied to Session Replay, which is currently Open Beta and we're working on GA: 
* https://github.com/getsentry/sentry/issues/74441

The `develop` docs have not been added yet, but the ticket with details about it is here:
* https://github.com/getsentry/develop/issues/1206

Relates to:
* https://github.com/getsentry/sentry-react-native/issues/3859
* https://github.com/getsentry/sentry-java/issues/3599
* https://github.com/getsentry/sentry-go/issues/842
* https://github.com/getsentry/sentry-dart/issues/1593
* https://github.com/getsentry/sentry-dart/pull/2230
* https://github.com/getsentry/sentry-dart/pull/2369
* https://github.com/getsentry/sentry-javascript/issues/10953
* https://github.com/getsentry/sentry/issues/68481 - `trace_id` allows the product to show all related telemetry collected during that trace

bruno-garcia:
Relates to:

* https://github.com/getsentry/sentry-cocoa/issues/4187

PR #3981, opened 2025-02-19
Support sending User Feedback without errors/exceptions

jamescrosswell:
Resolves https://github.com/getsentry/sentry-dotnet/issues/3605:
- https://github.com/getsentry/sentry-dotnet/issues/3605

### Related 
- https://github.com/getsentry/sentry-docs/pull/12846

## Example

image


bruno-garcia:
Adding @aliu39 and @billyvg from the Replay team for review

src/Sentry/Protocol/Envelopes/Envelope.cs
```diff
@@ -248,31 +248,78 @@ public static Envelope FromEvent(
...
+    /// <summary>
+    /// Creates an envelope that contains a single feedback event.
+    /// </summary>
+    public static Envelope FromFeedback(
+        SentryEvent @event,
+        IDiagnosticLogger? logger = null,
+        IReadOnlyCollection<SentryAttachment>? attachments = null,
+        SessionUpdate? sessionUpdate = null)
+    {
+        if (@event.Contexts.Feedback == null)
```
bruno-garcia, line 301: Curious why feedback lives on the contexts?

src/Sentry/Protocol/Envelopes/Envelope.cs
```diff
@@ -248,31 +248,78 @@ public static Envelope FromEvent(
...
+        var items = new List<EnvelopeItem>
+        {
+            EnvelopeItem.FromFeedback(@event)
+        };
+
+        if (attachments is { Count: > 0 })
+        {
+            if (attachments.Count > 1)
+            {
+                logger?.LogWarning("Feedback can only contain one attachment. Discarding {0} additional attachments.",
```
bruno-garcia, line 318: Really? Why can it contain only one?

src/Sentry/SentryClient.cs
```diff
@@ -82,6 +82,42 @@ public SentryId CaptureEvent(SentryEvent? @event, Scope? scope = null, SentryHin
...
+        {
+            _options.LogWarning("Feedback dropped due to empty message.");
+            return;
+        }
+
+        scope ??= new Scope(_options);
+        hint ??= new SentryHint();
+        hint.AddAttachmentsFromScope(scope);
+
+        _options.LogInfo("Capturing event.");
```
bruno-garcia, line 97:
```suggestion
        _options.LogInfo("Capturing feedback: '{0}'.", feedback.Message);
```

src/Sentry/SentryClient.cs
```diff
@@ -82,6 +82,42 @@ public SentryId CaptureEvent(SentryEvent? @event, Scope? scope = null, SentryHin
...
+
+        var evt = new SentryEvent { Level = SentryLevel.Info };
+        evt.Contexts.Feedback = feedback;
```
bruno-garcia, line 100: I find this odd. Are we trying to find a way to attach the feedback to the event somehow? 
Could the feedback inherit from SentryEvent instead?

src/Sentry/SentryFeedback.cs
```diff
@@ -0,0 +1,101 @@
...
+    /// </summary>
+    public string? Name { get; set; }
+
+    /// <summary>
+    /// Optional ID of the Replay session associated with the feedback.
+    /// </summary>
+    public string? ReplayId { get; set; }
+
+    /// <summary>
+    /// The name of the user.
```
bruno-garcia, line 38: Name of the user or a URL? What URL would it be?

afaik a feedback is an event. Should we just derive from it?

src/Sentry/Protocol/Envelopes/Envelope.cs
jamescrosswell: According to [the developer docs](https://develop.sentry.dev/application-architecture/feedback-architecture/#attachments), yes:
> We only use attachments for the widget’s screenshot feature, which allows users to submit at most 1 screenshot per feedback.

src/Sentry/Protocol/Envelopes/Envelope.cs
jamescrosswell: Not sure - I was just following [the spec](https://develop.sentry.dev/application-architecture/feedback-architecture/#feedback-events):

> The user's submission is wrapped in a context object:
```
event[”contexts”][”feedback”] = {
	"name": <user-provided>,
	"contact_email": <user-provided>,
	"message": <user-provided>,
	"url": <referring web page>,
	"source": <developer-provided, ex: "widget">,
	"associated_event_id": <developer-provided, should be a valid error event in the same project>
}

// All fields are optional except message.
// The widget can be configured to require a non-empty email and/or name.
```

Presumably this would eventually allow feedback to be mixed and matched with other event types. For example, it would technically be possible to capture an exception with feedback in the context.

It looks like this decision was made some time ago, in any case, as we're one of the last SDKs to be implementing the new Feedback APIs.

samples/Sentry.Samples.Console.Customized/Program.cs
```diff
@@ -142,12 +142,18 @@ await SentrySdk.ConfigureScopeAsync(async scope =>
...
             var eventId = SentrySdk.CaptureMessage("Some warning!", SentryLevel.Warning);
 
-            // Send an user feedback linked to the warning.
+            // Send feedback linked to the warning.
             var timestamp = DateTime.Now.Ticks;
             var user = $"user{timestamp}";
             var email = $"user{timestamp}@user{timestamp}.com";
-
-            SentrySdk.CaptureUserFeedback(new UserFeedback(eventId, user, email, "this is a sample user feedback"));
+            SentrySdk.CaptureFeedback(new SentryFeedback(
```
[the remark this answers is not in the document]
jamescrosswell: Good call. I'll add an extension method to the interface and an overload to the static SentrySdk class.

src/Sentry/Internal/Hub.cs
```diff
@@ -506,6 +506,18 @@ private SentryId CaptureEvent(SentryEvent evt, SentryHint? hint, Scope scope)
...
+    public void CaptureFeedback(SentryFeedback feedback, Scope? scope = null, SentryHint? hint = null)
+    {
+        if (!IsEnabled)
+        {
+            return;
+        }
+
+        scope ??= CurrentScope;
+        CurrentClient.CaptureFeedback(feedback, scope, hint);
+        scope.SessionUpdate = null;
```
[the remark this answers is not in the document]
jamescrosswell: The docs described Feedback as being basically like SentryEvent only with a type of `feedback` in the header and some feedback in the Contexts array. 

But you're right... The session stuff happens in DoSendEvent (which only executes when capturing events - we capture envelopes directly when capturing feedback). 

I'll get rid of it.

src/Sentry/SentryClient.cs
jamescrosswell: Not without changing all the other SDKs. This is just how the spec says we should be doing it.

src/Sentry/Protocol/Envelopes/Envelope.cs
bruno-garcia: I think this is a JS implementation thing. Or the FE is only showing one attachment. But ingestion wise we possibly support more than one (afaik we do in error events). @billyvg do you have context?

billyvg: I know on the SDK/widget side of the UI, we wanted to limit it to 1 for simplicity, but I'm unsure about ingest cc @aliu39?

aliu39: Ingest supports more than 1, but IIRC feedback UI only allows 1? Which is why I put that snippet in dev docs

aritchie (approved):

src/Sentry/Protocol/Envelopes/Envelope.cs
michellewzhang: we have the ability to show more than 1 attachment, the ui is set up to handle displaying multiple

codecat:
Can there be a `Action<Scope> configureScope` parameter on `CaptureFeedback`? [The documentation](https://docs.sentry.io/platforms/dotnet/enriching-events/scopes/#using-scope-callback-parameter) even says:
> In the following example we use **the scope callback parameter that is available for all capture methods** to attach a level and a tag to only one specific error:

PR #3993, opened 2025-02-21
Fix fromjson serialization of breadcrumbs from android

### LINKED ISSUE #3851: [Android] ArgumentException while processing Breadcrumbs
### Package

Sentry

### .NET Flavor

.NET

### .NET Version

9.0.100

### OS

Android

### SDK Version

4.12.2

_No response_

### Steps to Reproduce

1. Initialize Sentry, enable BeforeSend feature
2. Use the app, so that a breadcrumb with `level = FATAL` is generated
3. Try to send a SentryEvent

### Expected Result

Sentry event is processed and send to server

### Actual Result

Sentry event is not processed, exception is thrown:
```
System.ArgumentException: Requested value 'fatal' was not found.
   at System.Enum.TryParseByName[UInt32](RuntimeType enumType, ReadOnlySpan`1 value, Boolean ignoreCase, Boolean throwOnFailure, UInt32& result)
   at System.Enum.TryParseByValueOrName[Int32,UInt32](RuntimeType enumType, ReadOnlySpan`1 value, Boolean ignoreCase, Boolean throwOnFailure, Int32& result)
   at System.Enum.TryParse(Type enumType, ReadOnlySpan`1 value, Boolean ignoreCase, Boolean throwOnFailure, Object& result)
   at System.Enum.Parse(Type enumType, String value, Boolean ignoreCase)
   at Sentry.Internal.Extensions.EnumExtensions.ParseEnum[BreadcrumbLevel](String str)
   at Sentry.Breadcrumb.FromJson(JsonElement json)
   at System.Linq.Enumerable.SelectEnumerableIterator`2[[System.Text.Json.JsonElement, System.Text.Json, Version=8.0.0.0, Culture=neutral, PublicKeyToken=cc7b13ffcd2ddd51],[Sentry.Breadcrumb, Sentry, Version=4.12.2.0, Culture=neutral, PublicKeyToken=fba2ec45388e2af0]].MoveNext()
   at System.Collections.Generic.List`1[[Sentry.Breadcrumb, Sentry, Version=4.12.2.0, Culture=neutral, PublicKeyToken=fba2ec45388e2af0]]..ctor(IEnumerable`1 collection)
   at System.Linq.Enumerable.ToList[Breadcrumb](IEnumerable`1 source)
   at Sentry.SentryEvent.FromJson(JsonElement json, Exception exception)
   at Sentry.Android.Extensions.SentryEventExtensions.ToSentryEvent(SentryEvent sentryEvent, SentryOptions javaOptions)
   at Sentry.Android.Callbacks.BeforeSendCallback.Execute(SentryEvent e, Hint h)
   at Sentry.JavaSdk.SentryOptions.IBeforeSendCallbackInvoker.n_Execute_Lio_sentry_SentryEvent_Lio_sentry_Hint_(IntPtr jnienv, IntPtr native__this, IntPtr native_p0, IntPtr native_p1)
```

The issue seems to be caused by the mismatch in Breadcrumb.level types:
In sentry-java:
https://github.com/getsentry/sentry-java/blob/90eb1e3c65f721893b17bce4d286b482afccacb8/sentry/src/main/java/io/sentry/SentryLevel.java#L8-L13
In sentry-dotnet:
https://github.com/getsentry/sentry-dotnet/blob/1ecab829c0c7d6533ba0cbd24dac9ac1e7caf6ba/src/Sentry/BreadcrumbLevel.cs#L6-L40

aritchie:
Resolves #3851 should also prevent issues from cocoa SDK as well

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+- Fix mismapped breadcrumb levels coming in from native to dotnet SDK ([#3993](https://github.com/getsentry/sentry-dotnet/pull/3993))
+
 ### Features
 
 - Serilog scope properties are now sent with Sentry events ([#3976](https://github.com/getsentry/sentry-dotnet/pull/3976))


jamescrosswell (approved):
Looks good. Would be great if we could add some unit tests. I think these would only run on the Android Device tests but that's OK.

aritchie:
@jamescrosswell I'm not sure how we would unit test these.  Thoughts?  The value coming from native is a string.  I don't see anyway to simulate all error conditions from native at this time either.

jamescrosswell:
> @jamescrosswell I'm not sure how we would unit test these. Thoughts? The value coming from native is a string. I don't see anyway to simulate all error conditions from native at this time either.

We not have bulletproof unit tests but perfect is the enemy of good, as they say. 

Ultimately this method takes a JsonElement as input. It should be possible to test the method with various different inputs. 

In the most basic case, that could be a JsonElement representing some JSON produced by the sentry-dotnet SDK when we serialize a breadcrumb. 

For this particular change, it would also be good to test it with the JSON for a breadcrumb that had been serialised using the Native SDK. Worst case, we simply run one of our sample apps and capture that JSON in the console/debug logs then copy/paste into a test. Ideally we could run the most recent version of the Native code to serialise one or more breadcrumbs to JSON and then parse that... which would ensure we were always testing against the JSON produced by the same version of the Native SDK that we embed with sentry-dotnet.

PR #3998, opened 2025-02-24
3944 iOS native error suppression

### LINKED ISSUE #3944: Add option to enable/disable EXC_BAD_ACCESS suppression

aritchie:
Follow up to #3958 

Resolves #3944


test/Sentry.Tests/SentrySdkTests.cs
```diff
@@ -831,6 +843,82 @@ public void InitHub_DebugEnabled_DebugLogsLogged()
...
+            }],
+            new Foundation.NSDictionary<Foundation.NSString, Foundation.NSString>()
+        );
+        ex.Stacktrace = st;
+        evt.Exceptions = [ex];
+        var result = SentrySdk.ProcessOnBeforeSend(options, evt);
+
+        if (suppressSegfaults)
+        {
+            result.Should().BeNull();
```
jamescrosswell, line 879: We could also ensure any custom beforesend handler isn't called in this scenario.

jamescrosswell (approved):
LGTM. I just left one small suggestion about the unit test.

test/Sentry.Tests/SentrySdkTests.cs
aritchie: Done

CHANGELOG.md
```diff
@@ -2,6 +2,11 @@
 
 ## Unreleased
 
+### Fixes
+
+- Using SentryOptions.SuppressSegfaults, users can now block duplicate errors from native due to dotnet NullReferenceExceptions - Defaults to false ([#3998](https://github.com/getsentry/sentry-dotnet/pull/3998))
+- Native iOS events are now exposed to the dotnet layer for users to hook through SentryOptions.BeforeSend and SentryOptions.OnCrashedLastRun ([#2102](https://github.com/getsentry/sentry-dotnet/pull/3958))
```
jamescrosswell, line 8: I take it this fixes an erroneous changelog entry from a previous PR?

It would be tidier to have that in it's own PR (it's not related to this change)... but as long as this gets merged fairly quickly, it probably doesn't matter too much.

src/Sentry/Platforms/Cocoa/SentrySdk.cs
```diff
@@ -250,4 +187,69 @@ private static CocoaSdk.SentryHttpStatusCodeRange[] GetFailedRequestStatusCodes(
...
     }
+
+    internal static CocoaSdk.SentryEvent? ProcessOnBeforeSend(SentryOptions options, CocoaSdk.SentryEvent evt)
+    {
+        // When we have an unhandled managed exception, we send that to Sentry twice - once managed and once native.
+        // The managed exception is what a .NET developer would expect, and it is sent by the Sentry.NET SDK
+        // But we also get a native SIGABRT since it crashed the application, which is sent by the Sentry Cocoa SDK.
+
+        // There should only be one exception on the event in this case
+        if (options.Native.SuppressSegfaults && evt.Exceptions?.Length == 1)
```
jamescrosswell, line 198: Interestingly, in the case of iOS, there are two different kinds of Segfaults that we suppress.

We always suppressed those with `xamarin_unhandled_exception_handler` in the stacktrace, since we were certain those were handed by managed code.

As of about a month ago, we [can] now also suppress `EXC_BAD_ACCESS` mach exceptions... for these we're less certain whether they are duplicates of a managed `NullReferenceException` or whether they originate from native code (and won't be reported as an NRE). 

I could see some users wanting to suppress the first type but not the second type.... so it might be worth making `options.Native.SuppressSegfaults` a Flag enum rather than a boolean - to allow more granular control of what gets suppressed and what doesn't. 

I think by default we should suppress the first kind (which we know have a managed exception counterpart) but not the EXC_BAD_ACCESS exceptions (which we're less sure of).

CHANGELOG.md
aritchie: The changelog shifting on me during the last merge.  This fixes both and could be merged today

jamescrosswell (approved):

PR #4006, opened 2025-02-27
Fix missing MAUI `Shell` navigation breadcrumbs on iOS

### LINKED ISSUE #3999: MAUI Shell breadcrumbs not working on iOS
### Package

Sentry.Maui

### .NET Flavor

.NET

### .NET Version

8.0.405

### OS

iOS

### SDK Version

5.1.1

_No response_

### Steps to Reproduce

- Create a MAUI app with Shell
- Set up a standard shell navigation and invoke a `GoToAsync` via button
- Once in the new page push a button backed by SentrySdk crash utility
- Restart the app
- Look at Sentry crash report

### Expected Result

`AppShell` logs in breadcrumbs

### Actual Result

No `AppShell` logs in breadcrumbs

albyrock87:
Fixes #3999

### DIFF
--- a/src/Sentry.Maui/Internal/MauiEventsBinder.cs
+++ b/src/Sentry.Maui/Internal/MauiEventsBinder.cs
@@ -34,7 +34,16 @@ internal class MauiEventsBinder : IMauiEventsBinder
     {
         if (bind)
         {
-            // Attach element events to all descendents as they are added to the application.
+            // Attach element events to all existing descendants (skip the application itself)
+            foreach (var descendant in application.GetVisualTreeDescendants().Skip(1))
+            {
+                if (descendant is VisualElement element)
+                {
+                    OnApplicationOnDescendantAdded(application, new ElementEventArgs(element));
+                }
+            }
+
+            // Attach element events to all descendants as they are added to the application.
             application.DescendantAdded += OnApplicationOnDescendantAdded;
             application.DescendantRemoved += OnApplicationOnDescendantRemoved;
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 ### Fixes
 
 - Unknown stack frames in profiles on .NET 8+ ([#3967](https://github.com/getsentry/sentry-dotnet/pull/3967))
+- Missing MAUI `Shell` navigation breadcrumbs on iOS ([#4006](https://github.com/getsentry/sentry-dotnet/pull/4006))
 
 ## 5.3.0
 
--- a/test/Sentry.EntityFramework.Tests/IntegrationTests.verify.cs
+++ b/test/Sentry.EntityFramework.Tests/IntegrationTests.verify.cs
@@ -5,7 +5,7 @@ namespace Sentry.EntityFramework.Tests;
 public class IntegrationTests
 {
     // needs to be a variable to stop EF from inlining it as a constant
-    static string shouldNotAppearInPayload = "SHOULD NOT APPEAR IN PAYLOAD";
+    private static string shouldNotAppearInPayload = "SHOULD NOT APPEAR IN PAYLOAD";
 
     [SkippableFact]
     public async Task Simple()
--- a/test/Sentry.Maui.Tests/MauiEventsBinderTests.Application.cs
+++ b/test/Sentry.Maui.Tests/MauiEventsBinderTests.Application.cs
@@ -145,6 +145,36 @@ public partial class MauiEventsBinderTests
         Assert.Single(_fixture.Scope.Breadcrumbs);
     }
 
+    [Fact]
+    public void Application_HandleApplicationEvents_TracksExistingDescendants()
+    {
+        // Arrange
+        var application = MockApplication.Create();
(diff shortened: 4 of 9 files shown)


src/Sentry.Maui/Internal/MauiEventsBinder.cs
```diff
@@ -45,8 +54,6 @@ public void HandleApplicationEvents(Application application, bool bind = true)
             }
 
             // Navigation events
-            application.PageAppearing += OnApplicationOnPageAppearing;
```
albyrock87, line 48: We're already subscribing to each `Page` methods below so this creates duplicate breadcrumbs for the same kind of event.

Should we remove the other one instead so we can reduce the number of subscribed events?

src/Sentry.Maui/Internal/MauiEventsBinder.cs
```diff
@@ -34,7 +34,16 @@ public void HandleApplicationEvents(Application application, bool bind = true)
     {
         if (bind)
         {
-            // Attach element events to all descendents as they are added to the application.
+            // Attach element events to all existing descendants.
+            foreach (var descendant in application.GetVisualTreeDescendants().Skip(1))
```
jamescrosswell, line 38: Just wondering why the `Skip(1)`? Maybe worth adding a comment to explain that...

jamescrosswell:
Thanks for the PR @albyrock87 ! This looks great ❤️ 

Can think of any way we could add automated tests for this? I'm thinking something like [these](https://github.com/albyrock87/sentry-dotnet/blob/a2752921456ddc5e19ba2d185f470a0fa1ce57f6/test/Sentry.Maui.Tests/MauiEventsBinderTests.Application.cs#L6)... not sure how easy/feasible it is to test this particular case though.

The only other thing is we need a small description in the CHANGELOG.md file to explain to other users what has changed. In this case, I think the change is a Feature... some people might argue it's a Fix. I'm OK with either of those - it's all just how you squint at it.

albyrock87:
@jamescrosswell I'm not able to compile the solution locally, but I've done my best to add a unit test which should pass and cover this PR entirely.

image

aritchie:
@albyrock87 Looks like the git submodules didn't pull for you.  I'll take a look and push this through.  Thanks for all the help on this.

bruno-garcia:
> @jamescrosswell I'm not able to compile the solution locally, but I've done my best to add a unit test which should pass and cover this PR entirely.
> 
> image

could you try: `git submodule update --init --recursive` ?

bruno-garcia:
@albyrock87 looks like CI is unhappy with:

> Error: /home/runner/work/sentry-dotnet/sentry-dotnet/test/Sentry.Maui.Tests/MauiEventsBinderTests.Application.cs(154,9): error CS0618: 'Application.MainPage.set' is obsolete: 'This property is deprecated. Initialize your application by overriding Application.CreateWindow rather than setting MainPage. To modify the root page in an active application, use Windows[0].Page for applications with a single window. For applications with multiple windows, use Application.Windows to identify and update the root page on the correct window.  Additionally, each element features a Window property, accessible when it's part of the current window.' [/home/runner/work/sentry-dotnet/sentry-dotnet/test/Sentry.Maui.Tests/Sentry.Maui.Tests.csproj::TargetFramework=net9.0]

If you could help with this bit so CI is OK we could merge the PR already it seems

albyrock87:
@bruno-garcia I've:
- rebased PR onto `main`
- added changelog 
- fixed unit test project build

`git submodule update --init --recursive` fixed the build, indeed, but now I see these two folders in my git changes, so I think those folders should be added to `.gitignore`.

image

bruno-garcia:
> git submodule update --init --recursive fixed the build, indeed, but now I see these two folders in my git changes, so I think those folders should be added to .gitignore.

The submodules don't need to go into .gitignore. It takes sometime to get used to working with it to be honest, yesterday I spent way too much time on that and I worked with these for years.

If you run `git diff`, does it show a change in commit sha? If so, you need to revert that.
If the diff shows a removed sha and the old one, copy the old one, go into the submodule director (which is essentially a different git repo) and `git checkout old-sha`.
Once you `cd ..` to get back to `sentry-dotnet`, if you do `git status` there won't be any diff anymore to correct.

jamescrosswell:
So this PR is actually ready to merge. 

The only thing stopping it now appears to be an issue with `dotnet format`, which may be related to this:
- https://github.com/dotnet/sdk/issues/46780#issuecomment-2719961093

I can't see how this would break in CI though as we're still using `9.0.100` in CI.

Additionally, when I use `9.0.201` locally I still have issues.

... the joys of fragile tooling 😞

jamescrosswell (approved):

jamescrosswell:
Managed to get it over the line this time simply by running `dotnet format` from a windows machine. It looks like they broke something with `LF` line endings at some point. If this pops up again, one possible workaround might be to change the format workflow to run on Windows instead of macOS. Ideally this would get fixed however.

@albyrock87 big thank you for the PR - it's very much appreciated!

PR #4025, opened 2025-03-05
feat: Added `SentrySdk.CrashedLastRun`

bitsandfoxes:
This came up through user feedback looking for functionality similar to AppCenter's `Crashes.GetLastSessionCrashReportAsync()`.

Snippet on how to consume the API
```
// This assumes that the SDK is already initialized and that he SDK has already set 
// up the native scope sync - it will return `Unknown` otherwise
var lastRunState = SentrySdk.CrashedLastRun();

switch (lastRunState)
{
    case Sentry.LastRunState.Crashed:
        // Perform recovery actions or show user message.
        break;
        
    case Sentry.LastRunState.DidNotCrash:
        // Operate normally
        break;
        
    case Sentry.LastRunState.Unknown:
        // The SDK was either not initialized or does not have access to the native layer
        break;
}
```

Also resolves https://github.com/getsentry/sentry-unity/issues/741

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+## Features
+
+- Added `SentrySdk.CrashedLastRun()`. Users can now retrieve the `LastRunState` ([#4025](https://github.com/getsentry/sentry-dotnet/pull/4025))
+
 ### Fixes
 
 - Using SentryOptions.Native.SuppressExcBadAccess and SentryOptions.Native.SuppressSignalAborts, users can now block duplicate errors from native due to dotnet NullReferenceExceptions - Defaults to false ([#3998](https://github.com/getsentry/sentry-dotnet/pull/3998))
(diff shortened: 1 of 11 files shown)


bruno-garcia (changes requested):
Do we return `true` if the app crashed with an unhandled C# exception? (On Unity it keeps going, on .NET it crashes)

At first I thought it made sense adding this here too but now thinking to the fact we have this SDK running on UWP and all other weird runtimes where we can't confidently report on this status. It's prob best to keep it on Unity only for now.

jamescrosswell:
So this would be a superset of the functionality provided by the native libraries (and wrapped in https://github.com/getsentry/sentry-dotnet/pull/3958)? As in, this would work for vanilla .NET applications as well?

Also, presumably not relevant for server side apps (like ASP.NET Core)?

bruno-garcia:
> Also, presumably not relevant for server side apps (like ASP.NET Core)?

The server could OOM and crash. And we'd say `DidNotCrash`

PR #4026, opened 2025-03-05
feat: Allow `IScopeObserver` to observe trace

bitsandfoxes:
## The Problem

Currently, errors on different layers - i.e. Unity C# -> Java -> C/C++ are not connected via trace ID. The Java and the native SDK allow for Hybrid SDKs to `SetTrace(SentryId traceId, SpanId parentSpanId)`. From the Unity SDK's point of view, there is no mechanism right now that allows the .NET SDK to automatically propagate the trace to the native layers.

## Proposal

Since the trace is held on the scope's [PropagationContext](https://github.com/getsentry/sentry-dotnet/blob/main/src/Sentry/SentryPropagationContext.cs) we can leverage the already existing TwP functionality. Additionally, we can use the `IScopeObserver` to "observe" the relevant `TraceID` and `ParentSpanID`.

### Allow the Unity SDK to `SetTrace`
There are multiple ways I can think of achieving this:
1. ~~Make `PropagationContext` public~~
2. ~~Add an accessors to the `Scope` to update the `PropagationContext~~
3. ~~Add an `InternalSentrySdk` class where we can put `SetTrace` (and other things that regular users are not supposed to go). We have similar solutions like [`PrivateSentrySDKOnly`](https://github.com/getsentry/sentry-cocoa/blob/main/Sources/Sentry/PrivateSentrySDKOnly.mm) for Cocoa and [`InternalSentrySdk`](https://github.com/getsentry/sentry-java/blob/main/sentry-android-core/src/main/java/io/sentry/android/core/InternalSentrySdk.java) in Java.~~
4. Make `InternalsVisible` to `Sentry.Unity`.

### Update the `IScopeObserver` 
This `IScopeObserver` can then be implemented by the respective native scope observer on the Unity SDK. 
https://github.com/getsentry/sentry-unity/pull/1997
This will also work for Android and iOS when building MAUI apps since. We'll just have to bump the native SDKs there.

### Unity SDK
The end result looks like this
Screenshot 2025-03-28 at 17 23 58

### .NET SDK
This also works on platforms where the .NET SDK brings support via `sentry-native`.
Since the SDK also has observers for iOS and Android, once those SDKs are updated, it'll work there as well.

## Context

Relates to https://github.com/getsentry/sentry-unity/pull/1997/
Relevant change to `sentry-java`: https://github.com/getsentry/sentry-java/pull/4137
Relevant change to `sentry-native`: https://github.com/getsentry/sentry-native/pull/1137


src/Sentry/IScopeObserver.cs
```diff
@@ -29,4 +29,9 @@ public interface IScopeObserver
     /// Sets the user information.
     /// </summary>
     void SetUser(SentryUser? user);
+
+    /// <summary>
+    /// Sets the current trace
+    /// </summary>
+    void SetTrace(SentryId traceId, SpanId parentSpanId);
```
bruno-garcia, line 36: spam should be optional no?
if don't have a sample rate set for tracing, we still pass a trace i'd around to link errors together right?

TwP project (trace without performance)?

bitsandfoxes: The way tracing without performance works is it expects a `SpanId` to create a hierarchy. Since we're leveraging the `PropagationContext` it requires that `parent span ID`.

src/Sentry/Platforms/Android/AndroidScopeObserver.cs
```diff
@@ -99,4 +99,9 @@ public void SetUser(SentryUser? user)
             _innerObserver?.SetUser(user);
         }
     }
+
+    public void SetTrace(SentryId traceId, SpanId parentSpanId)
+    {
+        // TODO: This requires sentry-java 8.4.0
```
bruno-garcia, line 105: Do we need to follow up later with this once this is added to the Java and Apple SDKs?
If so ideally we create a ticket so folks here can add it later

src/Sentry/Sentry.csproj
```diff
@@ -16,8 +16,14 @@
 
   <PropertyGroup Condition="'$(SolutionName)' == 'Sentry.Unity'">
     <TargetFrameworks>netstandard2.0;netstandard2.1</TargetFrameworks>
+    <!-- To be able to make the internals visible, since we're not signing the assembly for Unity -->
+    <SignAssembly>false</SignAssembly>
```
bruno-garcia, line 20: Does this affect things like assembly aliasing? Since before we did have signing and now we wouldn't?

bruno-garcia (approved):
I'm approving to unblock but give @jamescrosswell and @aritchie a chance to review this if they have time in the next day or two unless you need this out there asap

src/Sentry/Platforms/Android/AndroidScopeObserver.cs
bitsandfoxes: https://github.com/getsentry/sentry-dotnet/issues/4074

src/Sentry/Sentry.csproj
bitsandfoxes: I did some extensive checking and we're good!

PR #4029, opened 2025-03-08
Add Exception HResult to mechanism data

### LINKED ISSUE #3939: Also report exception HResult
### Problem Statement

Some exceptions, especially `COMException` or `Win32Exception`, sometimes do not provide a good exception message which makes it hard to analyze issues on Sentry. `COMException` even often returns an empty exception message.

However, in .NET every `Exception` object has a `HResult` property. Especially for the cases mentioned above, this value would be worth a lot as it contains the concrete error type (e.g. -2147417842 -> RPC_E_WRONG_THREAD). This `HResult` property is currently not sent with a Sentry event.

Therefore I would like to request the `HResult` property of an Exception to also be uploaded in a `SentryEvent`, and that its value is also visible in the Sentry web UI.

### Solution Brainstorm

According to your documentation, such information could be stored in the Exception.Mechanism.Meta field (https://develop.sentry.dev/sdk/data-model/event-payloads/exception/#meta-information).

However, there are currently only Linux/Apple specific definitions listed, like `signal` or `mach_exception`. A definition for a Windows HResult, or more generally HResult (as they may also be used in cross-platform .NET apps), should be added there.

Using `ISentryEventExceptionProcessor` I already tried manually adding some `HResult` field to the `Meta`. However, it was not displayed in the Sentry web UI. Only when I used `errno` to "fake" a HResult, it was showing up. So I assume you must also add server-side handling.

For the .NET side, the change should be fairly easy. In the [MainExceptionProcessor](https://github.com/getsentry/sentry-dotnet/blob/main/src/Sentry/Internal/MainExceptionProcessor.cs#L162), something alike should be added:

```
mechanism.InternalMeta = new()
{
    { "HResult", exception.HResult }
};
```

Lightczx:
Resolves #3939

### Example

This is an example of the event that gets captured by the Sentry.Samples.Console.Basic demo. You can see the `HResult` along with the other mechanism data at the bottom of the screenshot:
image

### DIFF
--- a/src/Sentry/Internal/MainExceptionProcessor.cs
+++ b/src/Sentry/Internal/MainExceptionProcessor.cs
@@ -199,6 +199,9 @@ internal class MainExceptionProcessor : ISentryEventExceptionProcessor
             exception.Data.Remove(Mechanism.DescriptionKey);
         }
 
+        // Add HResult to mechanism data before adding exception data, so that it can be overridden.
+        mechanism.Data["HResult"] = $"0x{exception.HResult:X8}";
+
         // Copy remaining exception data to mechanism data.
         foreach (var key in exception.Data.Keys.OfType<string>())
         {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Features
 
 - User Feedback can now be captured without errors/exceptions. Note that these APIs replace the older UserFeedback APIs, which have now been marked as obsolete (and will be removed in a future major version bump) ([#3981](https://github.com/getsentry/sentry-dotnet/pull/3981))
+- Exception.HResult is now included in the mechanism data for all exceptions ([#4029](https://github.com/getsentry/sentry-dotnet/pull/4029))
 
 ### Fixes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,11 +1,16 @@
 # Changelog
 
+## Unreleased
+
+### Features
+
+- Exception.HResult is now included in the mechanism data for all exceptions ([#4029](https://github.com/getsentry/sentry-dotnet/pull/4029))
+
 ## 5.3.0
 
 ### Features
 
 - User Feedback can now be captured without errors/exceptions. Note that these APIs replace the older UserFeedback APIs, which have now been marked as obsolete (and will be removed in a future major version bump) ([#3981](https://github.com/getsentry/sentry-dotnet/pull/3981))
-- Exception.HResult is now included in the mechanism data for all exceptions ([#4029](https://github.com/getsentry/sentry-dotnet/pull/4029))
 
 ### Fixes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,10 +1,15 @@
 # Changelog
 
-## 5.4.0
+## Unreleased
 
 ### Features
 
 - Exception.HResult is now included in the mechanism data for all exceptions ([#4029](https://github.com/getsentry/sentry-dotnet/pull/4029))
+
+## 5.4.0
+
+### Enhancements
+
 - Profiling: improve performance by subscribing only to necessary CLR events ([#3970](https://github.com/getsentry/sentry-dotnet/pull/3970))
 
 ### Fixes
--- a/test/Sentry.DiagnosticSource.IntegrationTests/SqlListenerTests.RecordsEfAsync.DotNet8_0.verified.txt
+++ b/test/Sentry.DiagnosticSource.IntegrationTests/SqlListenerTests.RecordsEfAsync.DotNet8_0.verified.txt
@@ -10,7 +10,15 @@
       SentryExceptions: [
         {
           Type: System.Exception,
-          Value: my exception
+          Value: my exception,
+          Mechanism: {
+            Type: generic,
+            Synthetic: false,
+            IsExceptionGroup: false,
+            Data: {
+              HResult: 0x80131500
+            }
+          }
         }
       ],
       Level: error,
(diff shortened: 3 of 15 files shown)


src/Sentry/Internal/MainExceptionProcessor.cs
```diff
@@ -199,6 +199,9 @@ private static Mechanism GetMechanism(Exception exception, int id, int? parentId
             exception.Data.Remove(Mechanism.DescriptionKey);
         }
 
+        // Add HResult to mechanism data before adding exception data, so that it can be overridden.
+        mechanism.Data["HResult"] = $"0x{exception.HResult:X8}";
```
bruno-garcia, line 203: Could you add a test too please?

We have tests already which you base off of:

https://github.com/getsentry/sentry-dotnet/blob/47df9824191f1e71c1a0a78e1e928cc509dff693/test/Sentry.Tests/Internals/MainExceptionProcessorTests.cs#L15-L25

bruno-garcia (approved):
Thank you for this! 1 review note only

src/Sentry/Internal/MainExceptionProcessor.cs
bruno-garcia: Oh one more note: https://learn.microsoft.com/en-us/dotnet/api/system.exception.hresult?view=net-9.0

How about we don't add the property if HResult is not `default`? 

Otherwise we'll be getting `HResult: 0` for lots of stuff

bruno-garcia:
> Provide an option to disable/enable this behavior?

No need for an option. If HResult is not `0` we should always add the value

src/Sentry/Internal/MainExceptionProcessor.cs
whiskhub: The chance for HResult being 0 is effectively 0 😅 because 0 means S_OK, so success.
The default HResult in C# is 0x80131500 COR_E_EXCEPTION.

Lightczx: Yes, unless someone manually change the HResult to 0, the value will always be like 0x800... . Every .NET BCL Exception have different default HResult.

So we can assume 0 is also meaningful that something is going wrong here (Deliberately set by developer or library author, although almost no one will do so).

Lightczx: > Could you add a test too please?

Not familiar about that, maybe someone else can help?

bruno-garcia: > The chance for HResult being 0 in an exception is effectively 0 😅 because 0 means S_OK, so success.

I see: https://learn.microsoft.com/en-us/windows/win32/seccrypto/common-hresult-values

This seems specific to Win32/COM but this property is on `Exception` not COMException for example. 

I'm wondering what's the value if I just throw new exception:

```
Console.WriteLine("Hello, World!");

try
{
    throw new Exception("what's the HResult?");
}
catch (Exception e)
{
    Console.WriteLine(e.HResult);
    throw;
}
// Results in

Hello, World!
-2146233088
```

Does `-2146233088` mean something?
If not, how do we only add this property to the exception if the value is relevant?

Would it be testing for some exception types? Or perhaps checking the value is within some range? (for example [within this range](https://learn.microsoft.com/en-us/windows/win32/seccrypto/common-hresult-values)?)

Lightczx: `-2146233088` is `0x80131500` which is COR_E_EXCEPTION

https://github.com/dotnet/runtime/blob/d17420560540e4184a78f431de8341f747068c4f/src/coreclr/pal/prebuilt/inc/corerror.h#L227

If a Exception is throw by BCL, then it may or may not be mapped/wrapped.
Especially for HttpRequestException, if the inner exception is IOException, the HResult is different from which inner exception is ScoketException

whiskhub: I would avoid adding any filtering here. Any HResult value may give some information. I don‘t think it will confuse anyone, since already today the HResult value can be seen on any exception in the debugger.

whiskhub: I've written a short test. Feel free to commit it:

```c#
[Fact]
public void Process_ExceptionWithHResult_MechanismDataIsPopulated()
{
	var sut = _fixture.GetSut();
	var evt = new SentryEvent();
	const int RPC_E_WRONG_THREAD = -2147417842;
	const string RPC_E_WRONG_THREAD_MSG = "The application called an interface that was marshalled for a different thread.";
	var ex = new System.Runtime.InteropServices.COMException(RPC_E_WRONG_THREAD_MSG, RPC_E_WRONG_THREAD);

	sut.Process(ex, evt);

	var sentryException = evt.SentryExceptions!.Single();
	var data = sentryException.Mechanism!.Data;

	Assert.Contains(data, pair => pair.Key == "HResult" && pair.Value.Equals("0x8001010E"));
}
```

bruno-garcia:
OK I'm convinced. Thanks @Lightczx and @whiskhub 
@jamescrosswell do you agree with the reasoning to keep it as-is?

And would you be able to take over this? Add the test I mean and merge it in?

```
Received: SqlListenerTests.RecordsSqlAsync.DotNet9_0.received.txt
Verified: SqlListenerTests.RecordsSqlAsync.verified.txt
```
test failure seems unrelated?

src/Sentry/Internal/MainExceptionProcessor.cs
bitsandfoxes: > -2146233088 is 0x80131500 which is COR_E_EXCEPTION
https://github.com/dotnet/runtime/blob/d17420560540e4184a78f431de8341f747068c4f/src/coreclr/pal/prebuilt/inc/corerror.h#L227

@bruno-garcia does it make sense to pull that mapping into the issues details page?

Lightczx: > pull that mapping into the issues details page

Not quite, that's not the full set of the HRESULT, and COM/WinRT/Win32 components can have custom values.

jamescrosswell:
> @jamescrosswell do you agree with the reasoning to keep it as-is?
> 
> And would you be able to take over this? Add the test I mean and merge it in?

Yep all good. No reason not to send this information really (and I can't think what logic we'd apply to know when it was or wasn't relevant). 

We already had tests - so I've updated the expectations on these.

jamescrosswell (approved):

Issue #3985, opened 2025-02-20
[.NET MAUI] JSON showing unknown_image and Android event lacks line and column numbers


souredoutlook:
### Package

Sentry

### .NET Flavor

Xamarin

### .NET Version

9.02

### OS

Android

### SDK Version

5.1.1

### Self-Hosted Sentry Version

saas

### Steps to Reproduce

with this csproj config:

```
<PropertyGroup Condition="'$(Configuration)|$(Platform)'=='AppStore|AnyCPU'"> 
 <Debugger>Xamarin</Debugger> 
 <DebugType>portable</DebugType> 
 <DebugSymbols>true</DebugSymbols> 
 <EmbedAllSources>true</EmbedAllSources> 
 <EnableSourceLink>true</EnableSourceLink> 
 <RunAOTCompilation>False</RunAOTCompilation> 
 <PublishTrimmed>False</PublishTrimmed> 
 <AndroidUseAapt2>True</AndroidUseAapt2> 
 <AndroidCreatePackagePerAbi>False</AndroidCreatePackagePerAbi> 
 <AndroidPackageFormat>aab</AndroidPackageFormat> 
 <AndroidLinkMode>None</AndroidLinkMode> 
 <Optimize>True</Optimize> 
 <EmbedAssembliesIntoApk>True</EmbedAssembliesIntoApk> 
 <UseSentryCLI>true</UseSentryCLI> 
 <SentryOrg>vesigo-studios</SentryOrg> 
 <SentryProject>ontime_mobile</SentryProject> 
 <SentryUploadSymbols>true</SentryUploadSymbols> 
 <SentryUploadAndroidProguardMapping>true</SentryUploadAndroidProguardMapping> 
</PropertyGroup>
```

There are no line or column numbers being shown in the events - native events are working as expected.

### Expected Result

Events are paired up with uploaded images

### Actual Result

Not symbolicating correctly, example:

https://vesigo-studios.sentry.io/issues/6309002628/events/2905a706b763497ca82e1933ae93ee6b/?project=4508716749684736

jamescrosswell:
Thanks @souredoutlook - I take it this is the same issue as https://github.com/getsentry/sentry-dotnet/issues/3983, in which case, see [the comment](https://github.com/getsentry/sentry-dotnet/issues/3983#issuecomment-2670330762) I made in that issue.

[...]

Issue #3983, opened 2025-02-19
Unable to get exception line numbers to show


nk-alex:
### Package

Sentry.Maui

### .NET Flavor

.NET

### .NET Version

9.0

### OS

Android

### SDK Version

14.0

### Steps to Reproduce

I initialize sentry in my **MauiProgram.cs** like this:

```
.UseSentry(options => {
	options.Dsn = "";
	options.Debug = true;
	options.AttachStacktrace = true;
})
```

Then I try to capture exceptions like this:

```
try
{
 ...
}
catch (Exception ex)
{
	SentrySdk.CaptureException(ex);
	SentrySdk.CaptureMessage(ex.ToString(), SentryLevel.Info);
}
```

I thought with CaptureException would have been enough. But since no line numers are showing, I also tried with CaptureMessage. No luck:

![Image](https://github.com/user-attachments/assets/57c393dd-fbc5-4dd6-b6e7-04e78feec77b)

Then I thought I could get line numbers information by uploading Debug Information Files. Which I did, but exceptions are still not showing exception line number information.

![Image](https://github.com/user-attachments/assets/610e4ec4-f379-473d-a57a-5c67b31a5645)

Is sentry able to give me this information?
Am I missing anything?

I'm unable to find the problem without the line number information.

### Expected Result

I would expect to find the line number where the exception is produced. I'm unable to find the problem without that information.

### Actual Result

Posted above

jamescrosswell:
Older (perhaps related) similar issue:
- https://github.com/getsentry/sentry-dotnet/issues/2177

jamescrosswell:
Hey @nk-alex,

It looks like you've uploaded some symbols and sources to Sentry so if you're still not seeing source context for exceptions, it seems likely Sentry isn't able to resolve to correct symbols/source. 

One thing to ensure is that the version of the code that is running when exceptions are captured and sent to Sentry is exactly the same version that was built when you uploaded the symbols/source to Sentry. Otherwise you might end up with a different debug_id in the PE file for the application to the debug_id that is included in the symbols/source being uploaded. [See here](https://docs.sentry.io/platforms/dotnet/data-management/debug-files/identifiers/#pdb-age-mismatches) for more details.

If you're doing this on a local development machine then basically you'd:
- Build the solution
- Upload the symbols/source files (if this isn't done automatically as part of your build process)
[...]

Issue #3761, opened 2024-11-17
Review Android AssemblyReader


bruno-garcia:
[Came up on Discord](https://discordapp.com/channels/732297728826277939/732297837953679412/1307521828302094337):

Our current version of Android AssemblyReader was taken from this sha:

* https://github.com/getsentry/sentry-dotnet/blob/cbe1af4c722b22eb6ee0cbe5c70b25a8d7056d41/src/Sentry.Android.AssemblyReader/ATTRIBUTION.txt#L2

Upstream, as the history shows:

* https://github.com/dotnet/android/commits/main/tools/assembly-store-reader/AssemblyStoreExplorer.cs

A change was done to it since:

* https://github.com/dotnet/android/pull/8694/files#diff-d705beccc7479fb87d9203224767b0bd417e49e3b8de238ec0cd0811179d1de6

I didn't look into it but might be related to making assemblies RID specific, changing the internal layout of the apk:
* https://github.com/dotnet/android/pull/8478

bruno-garcia:
@jamescrosswell @bricefriha one way to test if things are working, is getting a Release built Android app built on .NET 9, and see if we still get all C# line numbers in exceptions and crashes

jamescrosswell:
> [@jamescrosswell](https://github.com/jamescrosswell) [@bricefriha](https://github.com/bricefriha) one way to test if things are working, is getting a Release built Android app built on .NET 9, and see if we still get all C# line numbers in exceptions and crashes

Argh... this obviously slipped through the cracks.

It looks like all of the changes that have taken place to the store format have been implemented in `dotnet/android/tools/assembly-store-reader-mk2` (vs `dotnet/android/tools/assembly-store-reader`, which is the code we vendored in to Sentry). 

From what I can tell, in net9.0 they introduced a new store format and it's incompatible with the older format... so there are now two tools:
- `assembly-store-reader` for reading v1 stores (used in net8.0-android and earlier)
- `assembly-store-reader-mk2` for reading v2 stores (used in net9.0-android and later)

The relvant commits appear to be:
- https://github.com/dotnet/android/pull/8478
- https://github.com/dotnet/android/pull/9154

[...]

PR #4033, opened 2025-03-13
Enable symbolication and source context for net9.0-android

jamescrosswell:
Closes https://github.com/getsentry/sentry-dotnet/issues/3761:
- https://github.com/getsentry/sentry-dotnet/issues/3761

Fixes https://github.com/getsentry/sentry-dotnet/issues/3983: 
- https://github.com/getsentry/sentry-dotnet/issues/3983

Fixes https://github.com/getsentry/sentry-dotnet/issues/3985:
- https://github.com/getsentry/sentry-dotnet/issues/3985

## Summary

.NET 9 does introduced a new AssemblyStore format for APK files. Reading the new format is demonstrated in [assembly-store-reader-mk2](https://github.com/dotnet/android/blob/5ebcb1dd1503648391e3c0548200495f634d90c6/tools/assembly-store-reader-mk2/Main.cs). This PR adds support for reading the new format to the `AndroidAssemblyReaderFactory`. 

Also, the reason we didn't pick this up when we moved to `net9.0` is that the device tests are/were only running against `net8.0` targets. In this PR we'll try to run the device tests on `net8.0-android/ios` and `net9.0-android/ios`. 

## Notes for Reviewers

There's a lot of code here. However most of it is simply copied/vendored in from two external repositories.

### ELFSharp

**For the purposes of review, you can ignore _all_ of the files in the "src/Sentry.Android.AssemblyReader/ELFSharp" folder.** 

The V2 format for APKs stores all assemblies in in an RID specific folder but Android requires all files in that folder to be ELF files... so during the build process any exe, dll, pdb or config files stored in there get "packed" into ELF files... we need to use ElfSharp to unpack these. 

To avoid adding any dependencies, the code for ELFSharp has been copied in to the Sentry.Android.AssemblyReader project (with appropriate attribution). 

### dotnet/android

**For the purposes of review, you can ignore _most_ of the files in the "src/Sentry.Android.AssemblyReader/V2" folder.** 

`src/Sentry.Android.AssemblyReader/V2/AndroidAssemblyStoreReaderV2.cs` and `AndroidAssemblyDirectoryReaderV2` are classes I wrote (although `AndroidAssemblyDirectoryReaderV2.ArchiveAssemblyHelper` was vendored in). Otherwise, all of the files in `src/Sentry.Android.AssemblyReader/V2` have been copied from the `dotnet/android` repo, with appropriate attribution at the top of each file.

No functional changes were made to any of that code. The only alterations were:
- Visibility changed from `public `to `internal` 
- Unused code from files that we vendored in has been removed
- I removed the dependency on `Xamarin.LibZipSharp` (using the built in [ZipFile](https://learn.microsoft.com/en-us/dotnet/api/System.IO.Compression.ZipFile?view=net-8.0) class instead)
- A little bit of `Ctrl + .` in resharper to get the code to conform to our coding conventions

## References

Relates to:
- https://github.com/getsentry/sentry-dotnet/pull/2050
- https://github.com/dotnet/android/pull/8478
- https://elfsharp.it/


jamescrosswell:
@vaind wondering if you'd be able to look at the Notes for Reviewers on this PR. 

In particular, wondering if/how you dealt with ELF files in the first version of the AndroidAssemblyReader?

vaind:
With all the vendored-in code this is pretty hard to review. Can you link a diff showing your changes to the vendored code?

jamescrosswell:
> With all the vendored-in code this is pretty hard to review. Can you link a diff showing your changes to the vendored code?

I can't think of any simple way to do that. [This earlier commit](https://github.com/getsentry/sentry-dotnet/pull/4033/commits/339f4ef2bcec8b2710a29e50b5e2b921806eb6d4) represents the minimal work to get most of the vendored in code copied into our project without breaking the build... before any of our code actually uses that new code. So you could diff individual vendored in files with their equivalent in that commit. At that stage I was only trying to get the AssemblyStoreExplorer working though (so none of the code required to get `src/Sentry.Android.AssemblyReader/V2/AndroidAssemblyDirectoryReaderV2.cs` working). 

I've added some details to the PR description for reviewers, that hopefully makes things a little easier. 

For the most part, I think you can ignore the vendored (treat it kind of like a 3rd party library) as well as any changes I made to that code as my changes weren't functional in nature (I just removed stuff and tidied it up a little). The only exception **_might_** be the [change I made to remove `Xamarin.LibZipSharp`](https://github.com/getsentry/sentry-dotnet/pull/4033/commits/fc3e17a2b64b1e55d1f8e3a81f6254c82cdfea16)... potentially that's related to the [1 unit test that is still failing](https://github.com/getsentry/sentry-dotnet/blob/45769f95be5f4d7fd5b77acdb00ffa5e782214cb/test/Sentry.Android.AssemblyReader.Tests/AndroidAssemblyReaderTests.cs#L87)... although I don't immediately see how. [This one](https://github.com/getsentry/sentry-dotnet/blob/45769f95be5f4d7fd5b77acdb00ffa5e782214cb/test/Sentry.Android.AssemblyReader.Tests/AndroidAssemblyReaderTests.cs#L85) passes, after all. It's odd that we can read `Mono.Android.dll` from an archive that doesn't use the assembly store but we can't do the same for `System.Runtime.dll`.

scripts/device-test.ps1
```diff
@@ -21,13 +22,16 @@ $CI = Test-Path env:CI
 Push-Location $PSScriptRoot/..
 try
 {
-    $tfm = 'net8.0-'
+    if (!$Tfm)
+    {
+        $Tfm = 'net8.0'
```
bruno-garcia, line 27: once we clear all callsites to pass a tfm, we should delete this and throw an error instead.

src/Sentry.Android.AssemblyReader/ELFSharp/MachO/Dylib.cs
```diff
@@ -0,0 +1,38 @@
+using System;
+using System.IO;
+using System.Text;
+using ELFSharp.Utilities;
+
+namespace ELFSharp.MachO
+{
+    internal abstract class Dylib : Command
```
bruno-garcia, line 8: do we need this? this is a mac thing

src/Sentry.Android.AssemblyReader/ELFSharp/MachO/FatArchiveReader.cs
```diff
@@ -0,0 +1,32 @@
+using System.Collections.Generic;
+using System.IO;
+using ELFSharp.Utilities;
+
+namespace ELFSharp.MachO
```
bruno-garcia, line 5: Do we need antyhing from `.MachO` namespace?  not relevant for Android

bruno-garcia (commented):
I just skimmed this change. Ideally someone can do a proper review

scripts/device-test.ps1
jamescrosswell: This is just a default argument, if you don't specify a target framework. We could try to change the default to `latest` or something like that - would need a little bit of extra scripting to work out what latest was.

src/Sentry.Android.AssemblyReader/ELFSharp/MachO/Dylib.cs
jamescrosswell: Possibly not - I didn't make any attempt to trim off unused code from ELFSharp (it's had the minimal changes required to get our project to build with it). 

We could try to shrink it down. The linker probably removes unused code paths when publishing on Android but I guess we could slightly improve linker times by removing the code altogether.

Worth it?

vaind (approved):

Issue #3755, opened 2024-11-13
Session Replay- Android - `DSN is required.` crash on launch when a valid DSN is specified


bricefriha:
### Package

Sentry

### .NET Flavor

.NET

### .NET Version

8

### OS

Android

### SDK Version

4.13.0

### Steps to Reproduce

1. Create an Android project (not MAUI)
2. Add the DSN in the options:
options.Dsn = "..."; // Enter a valid DSN
options.Debug = true;

Alternatively, from the sentry-dotnet just run the sample app `Sentry.Samples.Android`

## Note
Make sure the SDK version is 4.13.0. versions below work just fine

### Expected Result

The app should launch without a crash

### Actual Result

Crashes on launch, telling us the DSN is missing

![Image](https://github.com/user-attachments/assets/f4a6a0d1-f2b9-463b-a42a-18bc81a0c618)

bricefriha:
To confirm, This affects all versions of the dotnet SDK above `4.12.2`

jamescrosswell:
> To confirm, This affects all versions of the dotnet SDK above `4.12.2`

Yikes, it looks like `samples/Sentry.Samples.Android/Sentry.Samples.Android.csproj` is filtered out of the `Sentry-CI-Build-macOS.slnf` as well, so this isn't' being detected in CI (which it should be).

bricefriha:
It seems like the issue is specifically related to the Android replay package => `sentry-android-replay-7.16.0.aar`

bricefriha:
**Update on this**

It seems like the configuration callback is never invoked via `void OptionsConfigurationCallback.Configure(JavaObject optionsObject)`.

I investigated, and here's what I observed:
- On MAUI Android: `OptionsConfigurationCallback.Configure(JavaObject optionsObject)` is invoked as soon as `SentryAndroid.Init();` is called
- On vanilla Android with the `4.12.2` SDK, the configuration callback is called from the code Android bindings generate. 
```Java
__this.Configure (p0!);
``` 
Line 93 of `Sentry.Java.Sentry.cs`
- On vanilla Android with `4.13` SDK: the app crashes before we get the chance to invoke the configuration method is even called

For the later two, `OptionsConfigurationCallback.Configure(JavaObject optionsObject)` is basically never called

It seems like this has something has to be fixed on `Sentry.Android.Bindings`, but the thing is that the 3 cases I mentioned all share the same code...

jamescrosswell:
[...]

PR #4075, opened 2025-04-02
chore(deps): update Java SDK to v8.6.0

bitsandfoxes:
Superseds https://github.com/getsentry/sentry-dotnet/pull/3911

The `sentry-native` NDK stuff [was moved](https://github.com/getsentry/sentry-java/pull/3189) from `sentry-java` to `sentry-native` and is no longer part of `android-ndk`. `sentry-native-ndk` has its own version and is now a dependency. This means we need to:
1. Download the accompanying `.pom` file for `android-ndk` from Maven
2. Read the dependency version
3. Download the correct version of `sentry-native-ndk`.

PR #4079, opened 2025-04-03
Detect Sentry Java SDK dependencies during build

jamescrosswell:
Resolves Java dependencies for all of the existing Maven repositories we depend on. 

Relates to:
- https://github.com/getsentry/sentry-dotnet/issues/3755#issuecomment-2555847331 

To fix that particular issue, we need to add an `AndroidMavenLibrary` reference to the `sentry-android-replay` ArtefactId when implementing:
- https://github.com/getsentry/sentry-dotnet/issues/4080 

# Summary

.NET 9 adds some features that allow us to [detect Java dependencies](https://learn.microsoft.com/en-us/dotnet/android/binding-libs/advanced-concepts/java-dependency-verification), with build warnings for any that are missing 🎉 

Previously we downloaded java/aar files for Sentry's Java SDK from Maven and then referenced these directly with `AndroidLibrary` XML tags. The new features in .NET 9 allow us to include a Manifest/POM file when referencing *.java files (the POM file is what describes the dependencies of the java library) or to reference maven repositories directly with `AndroidMavenLibrary` tags which takes care of downloading the java and POM files and referencing these.

## Solution

For all of this to work, `Sentry.Bindings.Android.csproj` has to target `net9.0-android`. So I've added a net9 target and it turns out we were missing various dependencies:
```
1>Xamarin.Android.Bindings.JavaDependencyVerification.targets(27,5): Error XA4242 : Java dependency 'androidx.lifecycle:lifecycle-process:2.2.0' is not satisfied. Microsoft maintains the NuGet package 'Xamarin.AndroidX.Lifecycle.Process' that could fulfill this dependency.
1>Xamarin.Android.Bindings.JavaDependencyVerification.targets(27,5): Error XA4242 : Java dependency 'androidx.lifecycle:lifecycle-common-java8:2.2.0' is not satisfied. Microsoft maintains the NuGet package 'Xamarin.AndroidX.Lifecycle.Common.Java8' that could fulfill this dependency.
1>Xamarin.Android.Bindings.JavaDependencyVerification.targets(27,5): Error XA4242 : Java dependency 'androidx.core:core:1.3.2' is not satisfied. Microsoft maintains the NuGet package 'Xamarin.AndroidX.Core' that could fulfill this dependency.
```

Although the tooling only identifies these when targeting net9, I believe we were missing the same dependencies when targeting net8.0. This may explain various "missing class" errors that users have been reporting at runtime. 

In addition to adding the tooling to ensure we get appropriate warnings then, this PR also [includes appropriate versions of the missing libraries](https://github.com/getsentry/sentry-dotnet/blob/18d17b2b5374973d9df4798955a1a5befc6ed00d/src/Sentry.Bindings.Android/Sentry.Bindings.Android.csproj#L23-L39).

# References
- https://learn.microsoft.com/en-us/dotnet/android/binding-libs/advanced-concepts/android-maven-library
- https://learn.microsoft.com/en-us/dotnet/android/binding-libs/advanced-concepts/resolving-java-dependencies

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
+
+### Fixes
+
+- Sentry Java SDK dependencies are now detected and included in the Android bindings ([#4079](https://github.com/getsentry/sentry-dotnet/pull/4079))
+
 ## 5.5.0
 
 ### Features
@@ -563,7 +569,6 @@ Unable to load DLL sentry-native or one of its dependencies
 ### Fixes
 
 - Dynamic Sampling Context not propagated correctly for HttpClient spans ([#3208](https://github.com/getsentry/sentry-dotnet/pull/3208))
-- Sentry Java SDK dependencies are now detected and included in the Android bindings ([#4079](https://github.com/getsentry/sentry-dotnet/pull/4079))
 
 ## 4.2.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -563,6 +563,7 @@ Unable to load DLL sentry-native or one of its dependencies
 ### Fixes
 
 - Dynamic Sampling Context not propagated correctly for HttpClient spans ([#3208](https://github.com/getsentry/sentry-dotnet/pull/3208))
+- Sentry Java SDK dependencies are now detected and included in the Android bindings ([#4079](https://github.com/getsentry/sentry-dotnet/pull/4079))
 
 ## 4.2.0
 
--- a/src/Sentry.Bindings.Android/Sentry.Bindings.Android.csproj
+++ b/src/Sentry.Bindings.Android/Sentry.Bindings.Android.csproj
@@ -1,10 +1,9 @@
 <Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
+<!--    <TargetFrameworks>net8.0-android34.0</TargetFrameworks>-->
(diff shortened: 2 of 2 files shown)


bitsandfoxes:
I've got this one ready to go https://github.com/getsentry/sentry-dotnet/pull/4079
Which ever one of us gets to merge first? 😅

jamescrosswell:
> I've got this one ready to go #4075 Which ever one of us gets to merge first? 😅

Interesting. Thanks heaps for bumping to 8.x!

TLDR; I think we need both PRs.

Full explanation, this PR doesn't bump the version at all. It exists to detect and add missing Java dependencies:
https://github.com/getsentry/sentry-dotnet/blob/18d17b2b5374973d9df4798955a1a5befc6ed00d/src/Sentry.Bindings.Android/Sentry.Bindings.Android.csproj#L23-L39

The tooling that uses the POM files to create errors for missing java dependencies was only added in net9.0 so I had to add a net9.0 target to get the errors... then trawl the net9.0-android dependencies for [MAUI core](https://www.nuget.org/packages/Microsoft.Maui.Core/10.0.0-preview.2.25165.1) to identify the appropriate minimum versions for these. 

I've got a couple of questions about your PR so will start a discussion there.

PS: yes, we're going to have to deal with merge conflicts on whichever PR is the second one to be merged. I think we should leave them separate though as they both address separate issues.

aritchie (approved):

PR #4085, opened 2025-04-06
Fix UWP Net Native compilation

jdswcz:
Sentry DotNet since version 5.2.0 (and until the most recent 5.5.0) can't be used in UWP projects because they fail to compile with following error:

`Sentry : Gatekeeper error GK0018 : GK0018 Value type 'Sentry.Internal.FnvHash' has an explicit default constructor which is unsupported. Use an initialization function instead.`

This happens with .NET Native compilation and static analysis enabled. Versions 5.1.1 and older aren't producing this issue as they don't contain FnvHash struct.

Similar error was solved in the past with other objects: https://github.com/getsentry/sentry-dotnet/pull/2415

Please test my changes before merging. As a customer, I want this issue quickly fixed, so I am helping you. However, I don't want to spend hours getting compilation of sentry working and testing it within my package. 💯 Thank you for your understanding.

### DIFF
--- a/src/Sentry/Internal/FnvHash.cs
+++ b/src/Sentry/Internal/FnvHash.cs
@@ -10,34 +10,39 @@ namespace Sentry.Internal;
 /// </remarks>
 internal struct FnvHash
 {
-    public FnvHash()
+    // Provide a constructor that takes a parameter (so it’s not the default one).
+    // This lets you explicitly set your initial hash code:
+    public FnvHash(int _)
     {
+        _hashCode = Offset;
     }
 
     private const int Offset = unchecked((int)2166136261);
     private const int Prime = 16777619;
 
-    private int HashCode { get; set; } = Offset;
+    // Field (no initializer here—set it in the constructor)
+    private int _hashCode;
 
     private void Combine(byte data)
     {
         unchecked
         {
-            HashCode ^= data;
-            HashCode *= Prime;
+            _hashCode ^= data;
+            _hashCode *= Prime;
         }
     }
 
-    private static int ComputeHash(byte[] data)
+    private static int ComputeHashInternal(byte[] data)
     {
-        var result = new FnvHash();
+        // Create FnvHash via the constructor, which initializes _hashCode to Offset
+        var result = new FnvHash(0);
         foreach (var b in data)
         {
             result.Combine(b);
         }
-
-        return result.HashCode;
+        return result._hashCode;
     }
 
-    public static int ComputeHash(string data) => ComputeHash(Encoding.UTF8.GetBytes(data));
+    public static int ComputeHash(string data)
+        => ComputeHashInternal(Encoding.UTF8.GetBytes(data));
 }
--- a/src/Sentry/Internal/FnvHash.cs
+++ b/src/Sentry/Internal/FnvHash.cs
@@ -5,44 +5,23 @@ namespace Sentry.Internal;
 ///
 /// See https://en.wikipedia.org/wiki/Fowler%E2%80%93Noll%E2%80%93Vo_hash_function#FNV_hash_parameters
 /// </summary>
-/// <remarks>
-/// We use a struct to avoid heap allocations.
-/// </remarks>
-internal struct FnvHash
+internal static class FnvHash
 {
-    // Provide a constructor that takes a parameter (so it’s not the default one).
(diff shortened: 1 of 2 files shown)


jamescrosswell:
Thanks for the PR @jdswcz !

I'm happy to get this to go green in CI. However ideally I'd also like to confirm the code fixes the problem and to do that I need to be able to reproduce the problem. 

It looks like to reproduce it static analysis needs to be enabled for .NET Native. How are you doing that? I tried adding `<EnableNETAnalyzers>true</EnableNETAnalyzers>` to a UWP application here and building it in release mode (using Sentry 5.5.0) but didn't get any warnings or errors from an analyser.

jdswcz:
> Thanks for the PR @jdswcz !
> 
> I'm happy to get this to go green in CI. However ideally I'd also like to confirm the code fixes the problem and to do that I need to be able to reproduce the problem.
> 
> It looks like to reproduce it static analysis needs to be enabled for .NET Native. How are you doing that? I tried adding `<EnableNETAnalyzers>true</EnableNETAnalyzers>` to a UWP application here and building it in release mode (using Sentry 5.5.0) but didn't get any warnings or errors from an analyser.

image

These are our Release build settings that produce this error.

jamescrosswell:
Thanks @jdswcz - if I use those build setting in Visual Studio then I get _an error_ from gatekeeper when trying to run the app in release mode... but it's not clear what the error is:
> Error		ILT0005: 'C:\Program Files (x86)\Microsoft SDKs\UWPNuGetPackages\runtime.win10-x64.microsoft.net.native.compiler\2.2.12-rel-31116-00\tools\x64\ilc\ilc.exe --gatekeeper @"C:\code\App1\obj\x64\Release\ilc\intermediate\gkargs.rsp"' returned exit code 1	App1				

I'm wondering how you got the more detailed/actionable error message?

src/Sentry/Internal/FnvHash.cs
```diff
@@ -5,44 +5,23 @@ namespace Sentry.Internal;
...
 /// </summary>
-/// <remarks>
-/// We use a struct to avoid heap allocations.
-/// </remarks>
-internal struct FnvHash
+internal static class FnvHash
 {
-    // Provide a constructor that takes a parameter (so it’s not the default one).
-    // This lets you explicitly set your initial hash code:
-    public FnvHash(int _)
```
jamescrosswell: @jdswcz this felt a bit weird... having a parameter to a constructor that we ignore.

On the other hand, the code we had before (using a struct that has no public properties) is also a bit weird.

There's not actually any reason to have a class or a struct here, so I simplified it all to a static method.

jdswcz: I understand your point and of course, it is up to you, how you resolve this issue. Many thanks for trying to fix the issue we have. 👍

jamescrosswell (approved):

jdswcz:
> Thanks @jdswcz - if I use those build setting in Visual Studio then I get _an error_ from gatekeeper when trying to run the app in release mode... but it's not clear what the error is:
> 
> > Error		ILT0005: 'C:\Program Files (x86)\Microsoft SDKs\UWPNuGetPackages\runtime.win10-x64.microsoft.net.native.compiler\2.2.12-rel-31116-00\tools\x64\ilc\ilc.exe --gatekeeper @"C:\code\App1\obj\x64\Release\ilc\intermediate\gkargs.rsp"' returned exit code 1	App1
> 
> I'm wondering how you got the more detailed/actionable error message?

Apologies for later response. You probably figured out, however, you can just open command line and run the command mentioned in error. Something like:

`C:\Program Files (x86)\Microsoft SDKs\UWPNuGetPackages\runtime.win10-x64.microsoft.net.native.compiler\2.2.12-rel-31116-00\tools\x64\ilc\ilc.exe --gatekeeper @"C:\code\App1\obj\x64\Release\ilc\intermediate\gkargs.rsp" >> output.txt`

Then, in the output.txt, you'll see the gatekeeper log. You may see some warnings and info messages. However, only line marked as 'Error' cause compilation to fail.

jdswcz:
Just tested Sentry nuget package v 5.1.1 and I can confirm the issue was resolved.

Thank you very much for your effort.

jamescrosswell:
Awesome, thanks @jdswcz for raising this and for making the PR! 🙏🏻

PR #4096, opened 2025-04-09
Pin iOS 18.0 in Samples and Test apps

jamescrosswell:
Pinning `TargetPlatformVersion` 18.0 for iOS targets in samples and test apps so that our builds are deterministic. Otherwise they break frequently when new versions of Xcode/Microsoft.iOS are released. 

For example the error from [this build run](https://github.com/getsentry/sentry-dotnet/actions/runs/14363626452/job/40271121336?pr=4084):
> ILLINK : error MT0180: This version of Microsoft.iOS requires the iOS 18.4 SDK (shipped with Xcode 16.3). Either upgrade Xcode to get the required header files or set the managed linker behaviour to Link Framework SDKs Only in your project's iOS Build Options > Linker Behavior (to try to avoid the new APIs). [/Users/runner/work/sentry-dotnet/sentry-dotnet/test/Sentry.Maui.Device.TestApp/Sentry.Maui.Device.TestApp.csproj::TargetFramework=net9.0-ios]

#skip-changelog

### DIFF
--- a/test/Sentry.Maui.Device.TestApp/Sentry.Maui.Device.TestApp.csproj
+++ b/test/Sentry.Maui.Device.TestApp/Sentry.Maui.Device.TestApp.csproj
@@ -4,7 +4,8 @@
     <TargetFrameworks></TargetFrameworks>
     <TargetFrameworks Condition="'$(NO_ANDROID)' == ''">$(TargetFrameworks);net8.0-android;net9.0-android</TargetFrameworks>
     <TargetFrameworks Condition="'$(NO_IOS)' == '' And $([MSBuild]::IsOSPlatform('OSX'))">$(TargetFrameworks);net8.0-ios;net9.0-ios</TargetFrameworks>
-    <!-- Pin target iOS version so that our tests don't break when new versions of Xcode are released ('net8.0-ios' floats up otherwise) -->
+    <!-- Pin target iOS version so that our tests don't break when new versions of Xcode are released.
+      'net8.0-ios' resolves the latest version of the iOS SDK otherwise. -->
     <TargetPlatformVersion Condition="$([MSBuild]::GetTargetPlatformIdentifier('$(TargetFramework)')) == 'ios'">18</TargetPlatformVersion>
     <UseMaui>true</UseMaui>
     <!-- Currently broken on .NET 7, see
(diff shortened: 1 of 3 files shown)


test/Sentry.Maui.Device.TestApp/Sentry.Maui.Device.TestApp.csproj
```diff
@@ -4,6 +4,9 @@
     <TargetFrameworks></TargetFrameworks>
     <TargetFrameworks Condition="'$(NO_ANDROID)' == ''">$(TargetFrameworks);net8.0-android;net9.0-android</TargetFrameworks>
     <TargetFrameworks Condition="'$(NO_IOS)' == '' And $([MSBuild]::IsOSPlatform('OSX'))">$(TargetFrameworks);net8.0-ios;net9.0-ios</TargetFrameworks>
+    <!-- Pin target iOS version so that our tests don't break when new versions of Xcode are released.
+      'net8.0-ios' resolves the latest version of the iOS SDK otherwise. -->
+    <TargetPlatformVersion Condition="$([MSBuild]::GetTargetPlatformIdentifier('$(TargetFramework)')) == 'ios'">18</TargetPlatformVersion>
```
Flash0ver, line 9: question: Do we need to think about different minor (or even patch)`TargetPlatformVersion`s between `net8.0-ios` and `net9.0-ios`?

jamescrosswell: When MAUI 9 was released it targeted iOS 18.0... In the projects that we have which represent NuGet packages you'll see we [include the API version in the TFM](https://github.com/getsentry/sentry-dotnet/blob/cb0a35cc8550ab3eaffabea1eeac022a4d3f16b9/src/Sentry.Maui/Sentry.Maui.csproj#L10-L11) like `net9.0-ios18.0`. We do this specifically to pin the version so that it doesn't float upward. This makes our CI builds deterministic and it means our SDK targets the minimum possible API level.

However that only works for libraries... for application projects the `18.0` bit seems to get ignored and adding it causes problems for IDEs like Rider. So for our Test/Sample apps we use the more generic TFM `net9.0-ios`.  The problem with that is it resolves to "whatever the most recent version of Microsoft.iOS is" and that changes from one day to the next. A couple of days ago Microsoft must have made a new version of Microsoft.iOS available for use with Xcode 16.3 and that broke all our CI builds because to get iOS 18.4 you need to install Xcode 16.3 (and [we currently have Xcode 16.2 pinned in our builds](https://github.com/getsentry/sentry-dotnet/blob/72ad46933c0ce9ea4e845d62482b0d2dad294a66/.github/actions/environment/action.yml#L18) - again, to give more deterministic builds). 

Eventually we might bump that to Xcode 16.3 (once this is GA) but we don't want that to happen randomly without our having made any changes to the source code. 

So for this PR I think we're fine to lock in the `TargetPlatformVersion` to 18.0... it could be 18.1, 18.2 or 18.3 as well. We don't need/use any of the new APIs in any of those versions of iOS so we don't really care... all we want to ensure is that the build doesn't try to use the "latest" version of iOS.

Flash0ver: Got it. Thanks for the details.

Flash0ver (approved):

aritchie (approved):

chojuninengu:
This is awesome @Flash0ver

PR #4101, opened 2025-04-10
Build Variables Source Generator

### LINKED ISSUE #2927: Generate AotHelper properties at compile time
A couple of different options

# Use built in attributes

The [RuntimeFeature.IsDynamicCodeSupported Property](https://learn.microsoft.com/en-us/dotnet/api/system.runtime.compilerservices.runtimefeature.isdynamiccodesupported?view=net-8.0) looks like it might do what we want in many cases (I think we're using [`AotHelper.IsNativeAot`](https://github.com/getsentry/sentry-dotnet/blob/27ede88985fcc92ce2f7f8ae53ba11cfd805b2f5/src/Sentry/Internal/AotHelper.cs#L3-L20) where we should probably be checking for Trimming compatibility instead in some cases). Need to do a search for uses of `AotHelper.IsNativeAot` and a bit of a review of these. 

# Generate these at compile time

For this AotHelper, instead of:

https://github.com/getsentry/sentry-dotnet/blob/217c11cea611a5f26e4a1abfbb7df37f34695722/src/Sentry/Internal/AotHelper.cs#L18-L19

Could we instead write an attribute at compile time? I imagine it's a lot cheaper (worth measuring) than trying to get a stack trace on a static constructor:

Something like this:
https://github.com/getsentry/sentry-dotnet/blob/217c11cea611a5f26e4a1abfbb7df37f34695722/src/Sentry/buildTransitive/Sentry.targets#L8-L29

Here's a PR that added writing the attribute at build time as a reference: https://github.com/getsentry/sentry-dotnet/pull/1739

_Originally posted by @bruno-garcia in https://github.com/getsentry/sentry-dotnet/pull/2920#discussion_r1409457272_
            

# References

- More info available here: https://devblogs.microsoft.com/dotnet/creating-aot-compatible-libraries/

aritchie:
Resolves #2927 

This is the initial implementation of proposed AOT build checks

Features
- Generates a precompiled list of msbuild properties we were interested in based on various issues within our repo
- Currently, we only look at PublishTrimmed and PublishAot in our AotHelper
- Generates module initializer only on runnable targets 
- Ability to disable source generator if user encounters issue 

TODO
- [x] Unit Tests 
- [x] Document ability to disable source generator


aritchie:
Before we release this, I would like to do a beta test to ensure all of the nuget stuff has zero issues.  CC: @jamescrosswell

modules/sentry-native
jamescrosswell, line 1: This seems odd. I'm assuming this PR shouldn't be messing with sentry-native?

src/Sentry/SentrySdk.cs
```diff
@@ -17,8 +22,27 @@ public static partial class SentrySdk
...
+
     internal static SentryOptions? CurrentOptions => CurrentHub.GetSentryOptions();
 
+
+    /// <summary>
+    /// This is called by a Sentry Source-Generator module initializers to help us determine things like
+    ///     Is your app AOT
+    /// </summary>
+    /// <param name="variables"></param>
+    public static void InitializeBuildVariables(Dictionary<string, string> variables)
```
Flash0ver, line 38: question: Should we "bury" the new public property and method in a new type that we "hide" a little bit?

Indeed - technically - user code cannot do too much "harm" here since it's likely that `AotHelper.IsTrimmed` is already evaluated and cached, even when user code calls this method.
Yet, I am wondering if we should move this new method and property to a different type, which is `public`, but in a more "hidden" namespace.

E.g. `Sentry.CompilerServices.BuildHelpers`

To avoid it being "accidentally discovered" by user code. And also to slightly declutter the `SentrySdk` type.
What do you think?

test/Sentry.Tests/Sentry.Tests.csproj
```diff
@@ -43,4 +43,10 @@
     <PackageReference Include="Microsoft.Extensions.Configuration.Binder" Version="8.0.1" />
   </ItemGroup>
 
+  <ItemGroup Condition="'$(TargetFramework)' == 'net9.0' OR '$(TargetFramework)' == 'net8.0'">
```
Flash0ver, line 46: question: Should we create a separate `Sentry.SourceGenerators.Tests` project?

To be consistent with the rest of our `src`-`test`-Project-pairs?

test/Sentry.Tests/BuildPropertySourceGeneratorTests.cs
```diff
@@ -0,0 +1,94 @@
...
+
+        var dict = buildProperties.ToDictionary(x => "build_property." + x.Key, x => x.Value, comparer: StringComparer.InvariantCultureIgnoreCase);
+        var provider = new MockAnalyzerConfigOptionsProvider(dict);
+
+        var driver = CSharpGeneratorDriver.Create([generator], optionsProvider: provider);
+        return driver.RunGenerators(compilation);
+    }
+}
+
+file class MockAnalyzerConfigOptionsProvider(Dictionary<string, string> buildProperties) : AnalyzerConfigOptionsProvider
```
Flash0ver, line 78: praise: I love the use of file-scoped types here.

src/Sentry/buildTransitive/Sentry.targets
```diff
@@ -5,6 +5,14 @@
     <Using Include="Sentry" />
   </ItemGroup>
 
+  <ItemGroup>
+    <CompilerVisibleProperty Include="PublishAot" />
+    <CompilerVisibleProperty Include="UseDotNetNativeToolchain" />
+    <CompilerVisibleProperty Include="PublishTrimmed" />
+    <CompilerVisibleProperty Include="DisableSentrySourceGenerator" />
```
Flash0ver, line 12: suggestion: Change to e.g. `SentrySourceGenerator` or `SentryBuildPropertySourceGenerator` or similar.

So that all our properties, that are intended to be set by user code, start with `Sentry`.
And offer a similar opt-out semantic via `"false"` as the existing `<SentryImplicitUsings>false</SentryImplicitUsings>`.

Then I am wondering if the name `SentrySourceGenerator` is too broad, should we add more Generators later.
But this thought could be a bit premature and we can think about this issue once it actually arises.
note: e.g. for Sentry Logs we **may** add a Source Generator later on, similar to .NET's `Microsoft.Extensions.Logging`-Source-Generator. But we could solve high-performance structured logging via an `[InterpolatedStringHandler]` instead ... not sure yet.
So perhaps `<SentrySourceGenerator>false</SentrySourceGenerator>` (`<SentrySourceGenerator>disable</SentrySourceGenerator>` is just fine to start with.

What's your preference?

src/Sentry/buildTransitive/Sentry.targets
```diff
@@ -5,6 +5,14 @@
...
+    <CompilerVisibleProperty Include="Configuration" />
```
Flash0ver, line 13: question: Do we need `"UseDotNetNativeToolchain"` and `"Configuration"`?

Flash0ver (changes requested):

src/Sentry/buildTransitive/Sentry.targets
aritchie: Yes - there were other stories linked to source generator outside of AOT for UWP/WinUI stuff.  

Configuration could help in the future.  It costs us nothing.

test/Sentry.Tests/Sentry.Tests.csproj
aritchie: I pushed it into Sentry.Tests because it is technically bundled as part of Sentry.  There were also various references I need already found within Sentry.Tests

src/Sentry.SourceGenerators/Sentry.SourceGenerators.csproj
```diff
@@ -0,0 +1,26 @@
...
+    <IsRoslynComponent>true</IsRoslynComponent>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Microsoft.CodeAnalysis.Analyzers" Version="3.3.4">
+      <PrivateAssets>all</PrivateAssets>
+      <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
+    </PackageReference>
+    <PackageReference Include="Microsoft.CodeAnalysis.CSharp" Version="4.3.0" PrivateAssets="all"/>
+    <PackageReference Include="Microsoft.CodeAnalysis.CSharp.Workspaces" Version="4.3.0" PrivateAssets="all"/>
```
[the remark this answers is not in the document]
aritchie: It was set as a private asset so it isn't a big deal.  I've removed in sake of getting something out

src/Sentry/buildTransitive/Sentry.targets
aritchie: I've removed it for now.  The source generator here is too small to even bother with a debugger attach

src/Sentry/SentrySdk.cs
aritchie: DOne

modules/sentry-native
aritchie: You're right.  I wasn't even sure how it got here because it wasn't in any of my commits.  Check this commit out > https://github.com/getsentry/sentry-dotnet/pull/4101/commits/c034685695b5d1d7b288a0d01b30e836cd29dcc0

test/Sentry.Tests/Sentry.Tests.csproj
bruno-garcia, line 46: Can this be something that will automatically add future versions? something akin to NET8_OR_GREATER but for msbuild conditions? I think we have this somewhere in there repo

bruno-garcia (commented):
lots of stuff new to me so I don't have much to add, just a few suggestions

test/Sentry.Tests/Sentry.Tests.csproj
Flash0ver: https://github.com/getsentry/sentry-dotnet/blob/1bf37620fe5955defb7eca8aa1218e3a969d9413/src/Sentry/Platforms/Native/buildTransitive/Sentry.Native.targets#L23

src/Sentry.SourceGenerators/BuildPropertySourceGenerator.cs
```diff
@@ -0,0 +1,92 @@
...
+
+    /// <summary>
+    /// Execute the source gen
+    /// </summary>
+    public void Execute(GeneratorExecutionContext context)
+    {
+        var opts = context.AnalyzerConfigOptions.GlobalOptions;
+        var properties = opts.Keys.Where(x => x.StartsWith("build_property.")).ToList();
+        if (properties.Count == 0)
+            return;
```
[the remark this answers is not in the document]
aritchie: Updated

jamescrosswell:
@aritchie you'll need to update from main to get the fix for the dependency errors... should help with CI.

Flash0ver (approved):

bruno-garcia (approved):

test/Sentry.SourceGenerators.Tests/BuildPropertySourceGeneratorTests.cs
```diff
@@ -2,16 +2,17 @@
...
+namespace Sentry.SourceGenerators.Tests;
 
 
 public class BuildPropertySourceGeneratorTests
 {
-    [Fact]
+    [SkippableFact]
     public Task RunResult_Success()
     {
+        Skip.If(RuntimeInformation.IsOSPlatform(OSPlatform.Windows));
```
jamescrosswell, line 14: Do we know if the source generator works on Windows? Any idea why this test behaves differently on Windows?

samples/Sentry.Samples.Android/Sentry.Samples.Android.csproj
```diff
@@ -12,6 +12,9 @@
   </PropertyGroup>
 
   <ItemGroup>
+    <ProjectReference Include="..\..\src\Sentry.SourceGenerators\Sentry.SourceGenerators.csproj"
```
jamescrosswell, line 15: Would SDK users need to reference this in order for the new functionality around detecting AOT etc. to work?

src/Sentry/Internal/AotHelper.cs
```diff
@@ -13,10 +13,31 @@ static AotHelper()
...
 
+
     [UnconditionalSuppressMessage("Trimming", "IL2026: RequiresUnreferencedCode", Justification = AvoidAtRuntime)]
     private static bool CheckIsTrimmed()
     {
+        if (Check("publishtrimmed"))
+            return true;
+
+        if (Check("publishaot"))
+            return true;
```
jamescrosswell, line 24: What if `Check` does find a build property and is able to parse it but the value is `false` (not `true`)? Wouldn't this logic ignore that value and proceed to the fallback logic (which really only detects whether publishTrimmed is true)?

test/Sentry.SourceGenerators.Tests/BuildPropertySourceGeneratorTests.cs
aritchie: It has issues with the spacing.  It wasn't worth going through the exercise of fixing that

aritchie: It being verify.

samples/Sentry.Samples.Android/Sentry.Samples.Android.csproj
aritchie: No.  It gets built in via a None itemgroup in the Sentry.csproj.  I also had to add the targets file here manually in order to get the source gen to run in this project.  Users who get this via nuget don't have to worry about any of this

src/Sentry/Internal/AotHelper.cs
aritchie: Agreed.  I will make this update

test/Sentry.SourceGenerators.Tests/BuildPropertySourceGeneratorTests.cs
jamescrosswell: Does it produce different results every time on Windows? Or just different to the results that you get on Linux/macOS?

If it's just different to the results you get on macOS then you can tweak verify on a per test basis with stuff like this:
https://github.com/getsentry/sentry-dotnet/blob/3364647c890802ff858ff5b15fcd54035cea9b8c/test/Sentry.Testing/ApiExtensions.cs#L29

src/Sentry/Internal/AotHelper.cs
aritchie: Done

jamescrosswell:
> Before we release this, I would like to do a beta test to ensure all of the nuget stuff has zero issues. CC: @jamescrosswell

Makes sense. We can do a pre-release (like the 5.7.0 beta that we just made).

PR #4103, opened 2025-04-10
Bump to cocoa sdk v8.46.0

jamescrosswell:
Bumps the Cocoa SDK to version 8.46, which is the oldest version which would be stable for implementing Session Replay (needed to implement https://github.com/getsentry/sentry-dotnet/issues/2136). 

I also tidied up the `scripts/generate-cocoa-bindings.ps1` script a bit. It was starting to get quite large, mostly because we were using it to add lots of API contracts  from the `Sentry-Swift.h` file (rather than tweak contracts generated by objective sharpie). There didn't seem much point in using PowerShell to write static C# code so I moved that stuff out to the following files (where we at least get syntax checking and analysers from the IDE):
- `src/Sentry.Bindings.Cocoa/SwiftApiDefinitions.cs`
- `src/Sentry.Bindings.Cocoa/SwiftStructsAndEnums.cs`

This hopefully makes the cocoa bindings script a bit easier to read and maintain.


aritchie (approved):
The deeper we go with this, the more it seems we should consider a slim binding in the near future

jamescrosswell:
> The deeper we go with this, the more it seems we should consider a slim binding in the near future

Possibly yes. If we had slim bindings, I wouldn't have had to add all these methods to satisfy  the "missing class" errors for classes referred to by the new objective sharpie code.

On the other hand, once we have slim bindings, we don't have any easy mechanism to keep these up to date if/when the CocoaSDK API's change... or I guess we would - that would be our unit tests. So we're _really_ need to make sure we had good coverage in our device/integration tests for all the bindings code in that case.

PR #4106, opened 2025-04-11
Ensure mechanism data is not fully overwritten

### LINKED ISSUE #4027: Exception.Data missing
When handling some exceptions I have the following:

```csharp
try {
  await PotentiallyDangerousCodeAsync();
} catch (SomeKindOfException ex) {
  ex.Data.Add("FieldToHelpMeDebug", "This is why");
  throw;
}
```

However, it looks like the `Data` fields are not showing in the Sentry UI.
Is there any way to get those fields to show up?

aritchie:
Resolves #4027 

ASP.NET Core's AspNetCoreExceptionProcessor was overwriting the Mechanism and thereby clearing any exception data that may have been recorded.

### DIFF
--- a/src/Sentry.AspNetCore/AspNetCoreExceptionProcessor.cs
+++ b/src/Sentry.AspNetCore/AspNetCoreExceptionProcessor.cs
@@ -14,11 +14,9 @@ internal class AspNetCoreExceptionProcessor : ISentryEventExceptionProcessor
             {
                 foreach (var ex in @event.SentryExceptions)
                 {
-                    ex.Mechanism = new Mechanism
-                    {
-                        Type = "ExceptionHandlerMiddleware",
-                        Handled = false
-                    };
+                    ex.Mechanism ??= new Mechanism();
+                    ex.Mechanism.Type = "ExceptionHandlerMiddleware";
+                    ex.Mechanism.Handled = false;
                 }
             }
         }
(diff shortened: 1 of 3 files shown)


aritchie:
@jamescrosswell how come this was turned into a draft?  The CI failures here don't have anything to do with the PR itself.  I've merged in the latest main in case I'm missing something needed for the PR to be finished

src/Sentry.AspNetCore/AspNetCoreExceptionProcessor.cs
jamescrosswell, line 1: We should probably add a test for this (nothing fancy - just a unit test for the `AspNetCoreExceptionProcessor.Process` method).

jamescrosswell (changes requested):
Looks good... I think we should update the expectations for our unit tests to match the code changes though. 

There's this one - could may be tweak it and/or add another test case:
https://github.com/getsentry/sentry-dotnet/blob/24cfe5beb3b7eac455f3a292754d40fc02e742b9/test/Sentry.Serilog.Tests/AspNetCoreIntegrationTests.cs#L6-L25

aritchie:
@jamescrosswell What's your thinking for the tweak though?  The code doesn't change the behaviour.  It just prevents the previous data from being completely overwritten.  I'm not even sure what I would be adjusting here.

jamescrosswell:
> @jamescrosswell What's your thinking for the tweak though? The code doesn't change the behaviour. It just prevents the previous data from being completely overwritten. I'm not even sure what I would be adjusting here.

At the moment, the delegate for handler in that test just throws an exception. If instead we throw, catch, add some exception.Data and rethrow, then we would expect the exception data to be retained in the SentryException.Mechanism (not overwritten).

aritchie:
@jamescrosswell I just added some basic tests directly around the AspNetCoreExceptionProcessor

jamescrosswell (approved):

Issue #4102, opened 2025-04-10
A way to disable "sentry-native"


maxkatz6:
### Problem Statement

We are developing a NativeAOT application. And as for now, native dependencies are linked into the main executable. It also adds additional dependencies to the application, which otherwise are not used.

Normally, it wouldn't be a problem, but we are getting reports from users with less common distributives, like Alt Linux:
```
/lib64/libcurl.so.4: version `CURL_OPENSSL_4' not found
```

Sidenote, I haven't found any information on what native dependency does. But I also haven't checked the source code yet. 
If `sentry-native` is used as a cornerstone for sending message, this request might not be worth it. But having a fully managed sentry implementation is still nice. 

### Solution Brainstorm

MSBuild property like `DisableSentryNative` would be nice. But I don't know how it affects runtime usage yet.

jamescrosswell:
Interesting. We leverage the Native SDK to capture errors with meaningful stack traces for native code.

It is possible to disable the NDK for Android applications. We don't yet have any way to disable the Native SDK for other platforms though.

Disabling the Native SDK might be the nuclear option... and we could look at adding a way to do that so that people could use it as a workaround in similar situations to this. It might be throwing the baby out with the bathwater though. 

A more optimal solution might be to try to resolve the missing dependency on libcurl. 

The [Platform and Feature Support docs](https://github.com/getsentry/sentry-native?tab=readme-ov-file#platform-and-feature-support) for the Native SDK mention this dependency:
> HTTP Transport is currently only supported on Windows and platforms that have the curl library available. On other platforms, library users need to implement their own transport, based on the function transport API.

Potentially we could plug in the transport from the managed SDK then (so that libcurl was no longer required). @vaind does that sound feasible (just looking for a sanity check)?

maxkatz6:
Another benefit of having a MSBuild parameter like `DisableSentryNative` is integration with [feature switches](https://medium.com/@malarsharmila/feature-switches-in-net-9-73b1cb96c13f). When such switch is disabled compile time, with trimming/aot enabled, compiler will remove all code blocks using this condition. Potentially removing a lot of unused code.

maxkatz6:
> However, @maxkatz6, I'd suggest not to include the option by default in your builds (as sentry-native is useful if users use native libraries and those crash) and only suggest it to users as an escape hatch if they fail to compile the library

[...]

PR #4107, opened 2025-04-13
Add `DisableSentryNative` build property

maxkatz6:
This PR adds a new `DisableSentryNative` build property, allowing to opt-out from linking and using Sentry.Native on .NET 8+ builds. By default, `DisableSentryNative` is false, so old behavior is kept unchanged.

Two main changes of this PR:
1. `DirectPInvoke`/`NativeLibrary` are only set when `DisableSentryNative` is false. 
2. `SentryNative.IsAvailable` returns false, if `DisableSentryNative` was set to true.

Runtime switches (`RuntimeHostConfigurationOption` and `AppContext.TryGetSwitch`) were used to pass build property to the runtime library. `FeatureSwitchDefinition` used should make this switch into a compile-time variable.

Closes https://github.com/getsentry/sentry-dotnet/issues/4102

NOTE: while it should work, I haven't tested the build yet. Need to figure out how to prepare a local build with native bits and everything included.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Features
+
+- Option to disable the SentryNative integration ([#4107](https://github.com/getsentry/sentry-dotnet/pull/4107))
+
 ### Fixes
 
 - Prevent users from disabling AndroidEnableAssemblyCompression which leads to untrappable crash ([#4089](https://github.com/getsentry/sentry-dotnet/pull/4089))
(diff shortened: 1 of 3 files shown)


jamescrosswell:
> NOTE: while it should work, I haven't tested the build yet. Need to figure out how to prepare a local build with native bits and everything included.

@maxkatz6 the PR where @vaind added `SentryNative.IsAvailable` might be helpful to look at. There are some integration tests build in PowerShell in that PR that demonstrate at least the conditions for SentryNative being available:
- https://github.com/getsentry/sentry-dotnet/pull/3363

Your code is looking at a couple of different scenarios/variants within that, I imagine (when it's enabled/disabled).

jamescrosswell (approved):
LGTM. 

Thank you so much for the PR @maxkatz6 ❤️

Issue #3915, opened 2025-01-23
CI Builds failing due to SSL Errors


jamescrosswell:
We've been seeing these for a while now but they're getting more and more frequent... to the point that it's difficult to get any PRs to pass (having to retry multiple jobs, multiple times to get one that works). 

We're getting SSL Errors when running `dotnet restore` and/or when trying to [install the Android dependencies](https://github.com/getsentry/sentry-dotnet/blob/b22155fc964dbbdd852636ee9b2854558e28b455/.github/workflows/build.yml#L222-L232):
```
Determining projects to restore...
  Restored /Users/runner/work/sentry-dotnet/sentry-dotnet/integration-test/maui-app/maui-app.csproj (in 2.35 sec).
/Users/runner/.dotnet/packs/Microsoft.Android.Sdk.Darwin/34.0.145/tools/Xamarin.Installer.Common.targets(12,3): error : Failed to download Xamarin Repository manifest from https://aka.ms/AndroidManifestFeed/d17-12. System.AggregateException: One or more errors occurred. (The SSL connection could not be established, see inner exception.) [/Users/runner/work/sentry-dotnet/sentry-dotnet/integration-test/maui-app/maui-app.csproj::TargetFramework=net8.0-android34.0]
/Users/runner/.dotnet/packs/Microsoft.Android.Sdk.Darwin/34.0.145/tools/Xamarin.Installer.Common.targets(12,3): error :  ---> System.Net.Http.HttpRequestException: The SSL connection could not be established, see inner exception. [/Users/runner/work/sentry-dotnet/sentry-dotnet/integration-test/maui-app/maui-app.csproj::TargetFramework=net8.0-android34.0]
/Users/runner/.dotnet/packs/Microsoft.Android.Sdk.Darwin/34.0.145/tools/Xamarin.Installer.Common.targets(12,3): error :  ---> System.Security.Authentication.AuthenticationException: The remote certificate is invalid because of errors in the certificate chain: RevocationStatusUnknown [/Users/runner/work/sentry-dotnet/sentry-dotnet/integration-test/maui-app/maui-app.csproj::TargetFramework=net8.0-android34.0]
[...]

PR #4110, opened 2025-04-14
chore: Bump to dotnet SDK 9.0.203

jamescrosswell:
Bump the .NET SDK version to 9.0.203. Possibly impacts https://github.com/getsentry/sentry-dotnet/issues/3915. See:
- https://github.com/getsentry/sentry-dotnet/issues/3915#issuecomment-2800227556

#skip-changelog

### DIFF
--- a/.github/actions/environment/action.yml
+++ b/.github/actions/environment/action.yml
@@ -60,7 +60,7 @@ runs:
       with:
         dotnet-version: |
           8.0.x
-          9.0.202
+          9.0.203
 
     - name: Install .NET Workloads
       shell: bash
--- a/global.json
+++ b/global.json
@@ -1,6 +1,6 @@
 {
   "sdk": {
-    "version": "9.0.202",
+    "version": "9.0.203",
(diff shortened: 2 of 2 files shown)


Flash0ver (approved):
question (non-blocking): Do we want to keep the `bug_report.yml` GitHub Issue Template in sync?

https://github.com/getsentry/sentry-dotnet/blob/87919d1b6c15e1c84ec2016f9c34ad1713abc074/.github/ISSUE_TEMPLATE/bug_report.yml?plain=1#L48C1-L48C51

I can imagine we don't want to update the template with every patch release of the .NET SDK, but maybe with feature band and/or major releases. What do you think?

aritchie (approved):

jamescrosswell:
> Do we want to keep the `bug_report.yml` GitHub Issue Template in sync?

I think it's OK to have these diverge. That's just a template and SDK users might well be using a different version of the .NET SDK to the one we use to compile our NuGet packages.

PR #4124, opened 2025-04-17
MAUI Gesture Recognizer Auto-Breadcrumbs

aritchie:
This is not a current enhancement request.  This is being done to see how well we can "dig" into the view hierarchy as we look for auto instrumentation on things like command execution.

This is here for discussion to decide if we should bring it into the product.

TODOs
- [x] Unit Tests - like other event binders, this will be....difficult


bruno-garcia:
@aritchie should we add this in?

I think we might want to disable this functionality from the native SDKs if we do, I believe they have a similar thing going there, but since we don't sync those from native to C# (only the other way around), if we add this here we get them in all evnets, C# and native ones.

samples/Sentry.Samples.Maui/MainPage.xaml.cs
```diff
@@ -100,4 +100,8 @@ private async void OnFeedbackClicked(object sender, EventArgs e)
     {
         await Navigation.PushModalAsync(new SubmitFeedback());
     }
+
+    private void TapGestureRecognizer_OnTapped(object sender, TappedEventArgs e)
+    {
```
jamescrosswell, line 105: This method doesn't do anything. Is it required?

src/Sentry.Maui/Internal/MauiGestureRecognizerEventsBinder.cs
```diff
@@ -0,0 +1,197 @@
+namespace Sentry.Maui.Internal;
+
+/// <summary>
+/// Detects and breadcrumbs any gesture recognizers attached to the visual element
```
jamescrosswell, line 4:
```suggestion
/// Detects and adds breadcrumbs for any gesture recognizers attached to the visual element
```

samples/Sentry.Samples.Maui/MainPage.xaml.cs
aritchie: Yes - the hook is required

src/Sentry.Maui/Internal/MauiGestureRecognizerEventsBinder.cs
```diff
@@ -0,0 +1,197 @@
...
+        nameof(PointerGestureRecognizer.PointerEntered),
+        ToPointerData(e)
+    ));
+
+    private static IEnumerable<(string Key, string Value)> ToPointerData(PointerEventArgs e) =>
+    [
+        // some of the data here may have some challenges being pulled out
+        #if ANDROID
+        ("MotionEventAction", e.PlatformArgs?.MotionEvent.Action.ToString() ?? string.Empty)
+        //("MotionEventActionButton", e.PlatformArgs?.MotionEvent.ActionButton.ToString() ?? String.Empty)
```
jamescrosswell, line 147: Should we remove this? If it's a TODO item, we could put it in a separate issue...

src/Sentry.Maui/Internal/MauiGestureRecognizerEventsBinder.cs
```diff
@@ -0,0 +1,197 @@
...
+        #elif IOS
+        ("State", e.PlatformArgs?.GestureRecognizer.State.ToString() ?? string.Empty),
+        //("ButtonMask", e.PlatformArgs?.GestureRecognizer.ButtonMask.ToString() ?? String.Empty)
```
jamescrosswell, line 150: Again, we should probably remove this... don't want loads of commented out code in the solution.

src/Sentry.Maui/Internal/MauiGestureRecognizerEventsBinder.cs
aritchie: I'd like to see if we can expand on this in the near future.  I just didn't want to get too crazy with the tests yet (especially since they're flaky on xharness)

test/Directory.Build.props
```diff
@@ -11,8 +11,10 @@
     <NoWarn>$(NoWarn);SYSLIB0005;SYSLIB0012</NoWarn>
     <!-- Ignore "Naming rule violation: Missing suffix: 'Async'" -->
     <NoWarn>$(NoWarn);IDE1006</NoWarn>
-  </PropertyGroup>
 
+    <!--set this to true to enable the visual runner in the MAUI device tests - make sure it is false for CI-->
+    <EnableMauiDeviceTestVisualRunner>false</EnableMauiDeviceTestVisualRunner>
```
jamescrosswell, line 16:
```suggestion
    <EnableMauiDeviceTestVisualRunner Condition="'$(GITHUB_ACTIONS)' != 'true'">true</EnableMauiDeviceTestVisualRunner>
<EnableMauiDeviceTestVisualRunner Condition="'$(GITHUB_ACTIONS)' == 'true'">false</EnableMauiDeviceTestVisualRunner>
```

Do we want to do this? Or are there scenarios where we'd want to disable the visual tests locally?

test/Sentry.Maui.Device.TestApp/Sentry.Maui.Device.TestApp.csproj
```diff
@@ -1,17 +1,21 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
+  <PropertyGroup Condition="'$(EnableMauiDeviceTestVisualRunner)' == 'true'">
```
jamescrosswell, line 3: We could reverse this... only enable the maui device test visual runner property if the TFM is net9 or above.

That way we wouldn't have to fiddle with the TFMs in this csproj.

I think that would be cleaner and easier to reason about.

test/Sentry.Maui.Device.TestApp/Startup.cs
jamescrosswell, line 1: 🚀

src/Sentry.Maui/Internal/MauiGestureRecognizerEventsBinder.cs
aritchie: Removed comments

src/Sentry.Maui/Internal/MauiGestureRecognizerEventsBinder.cs
aritchie: Done

test/Sentry.Maui.Tests/MauiEventsBinderTests.GestureRecognizers.cs
```diff
@@ -0,0 +1,145 @@
+using Sentry.Maui.Internal;
+
+namespace Sentry.Maui.Tests;
+
+public partial class MauiEventsBinderTests
+{
+#if VISUAL_RUNNER
+    [Fact]
+#else
+    [Fact(Skip = "Flaky Test in XHarness")]
```
jamescrosswell, line 10: These are unit tests right, so they don't get run by the visual runner? They just get run by XHarness along with the other device tests? In which case, whether we skip these or not probably isn't dependent on whether the visual runner is enabled or not so much as whether we're running the tests on a device or not.

In that case, it'd be better if the directive was `#if !ANDROID && !IOS`... to make it clear this has nothing to do with the visual runner.

jamescrosswell (commented):
Some question, some suggestions. Looks like we're getting close! Will be great to have some visual runners finally.

test/Sentry.Maui.Tests/MauiEventsBinderTests.GestureRecognizers.cs
aritchie: They get run by the visual runner only.  The gesture recognizers are crazy flaky in the headless setup.  They run perfectly in the visual runner and full scenarios.

test/Directory.Build.props
aritchie: This flag is something I think is best to knowingly turn on/off.  I could force it to be off if CI to safety it, but I've got a note on it

aritchie: I also like to run xharness locally.  The headless runner isn't a perfect runner.

test/Sentry.Maui.Device.TestApp/Sentry.Maui.Device.TestApp.csproj
jamescrosswell: Yeah, I was thinking we might have a compound condition for `EnableMauiDeviceTestVisualRunner`... something like:
```
Condition="'$(GITHUB_ACTIONS)' == 'true' and $([MSBuild]::IsTargetFrameworkCompatible('net9.0', $(TargetFramework)))"
```

So the visual tests would only be run locally and only when targeting net9.0 and later...

This csproj file wouldn't have to change then. It could target net8.0 and net9.0.

aritchie: I really don't like overcomplicating stuff on flags that only we're going to touch.  It will be forced to be net9.0 because of the TFMs in the target.  I can't use the original TFM setup in the device tests project.  I think we need to leave this one for now.

aritchie: Furthermore, if the visual runner flag gets left as true... the xharness device tests will fail on CI.... so we've got safety anyhow.

jamescrosswell: It feels leaky to me. This is a bare bones test project... it's just a harness to run stuff. We shouldn't have logic in here related to to a specific assembly that it runs tests for.

jamescrosswell: > I really don't like overcomplicating stuff

Yeah, that's what I'm trying to avoid as well. The above seems very complicated to me - and when combined with all the other flags we have I can see it being quite hard to grasp in the future. So I'm looking at ways to simplify it. 

I'll play around with it locally to see if what I'm proposing would even work and whether it would actually make things simpler or more complex (sometimes these things sound good in theory but the reality is something different).

jamescrosswell: OK, I think I got it going... but man MS Build is painful to work with sometimes 😢 

This is hopefully a bit neater and it means you don't have to manually enable visual tests when running locally (they're always enabled locally when running the device tests against net9.0 and later targets).

jamescrosswell (approved):

Issue #3181, opened 2024-02-29
Mobile Vitals


bitsandfoxes:
What are [Mobile Vitals](https://docs.sentry.io/product/performance/mobile-vitals/)
- [ ] App Start
- [ ] Slow and Frozen Frames
- [ ] Time to Initial Display
- [ ] Time to Full Display

Can we pull those from the native SDKs?

de-preut:
Hey @bitsandfoxes any updates on that topic? These mobile vitals would really helpful for our MAUI app use case.

bitsandfoxes:
Sadly, I don't have an ETA for these just yet. But your timing is great. We're looking into prioritizing this and having some idea how much benefit this would give helps out greatly.

de-preut:
@bitsandfoxes any updates here? We would really benefit from the mobile vitals in our MAUI app!

bitsandfoxes:
@de-preut thanks for asking! As a "next feature" we do have that still quite high on the roadmap but I can't give you an ETA, unfortunately.

de-preut:
@bitsandfoxes sorry to bother you with this again, but this feature would increase the value of the sentry integration in our mobile app dramatically. Any ETA yet?

bitsandfoxes:
No worries! Feature requests are so much more happy than issues because something broke!
That'd be a prime example of catching up with the rest of the mobile offering.

@bruno-garcia @bricefriha for priorization.

bruno-garcia:
We have prior art if you look at our [Flutter](https://github.com/getsentry/sentry-dart) or [React Native SDKs](https://github.com/getsentry/sentry-react-native/), all that interacts with native is there so we just need to replicate here.

That said we should first add some auto instrumentation, for example:
* #1837

bricefriha:
> We have prior art if you look at our Flutter or React Native SDKs, all that interacts with native is there so we just need to replicate here

Should be relatively easy then, I reckon

de-preut:
any updates here?

bricefriha:
> Don't think @bricefriha got started yet. Did you @bricefriha ?

I haven't, but I can make it my next priority

bruno-garcia:
For the record, got a few [Feedbacks](https://docs.sentry.io/product/user-feedback/) in Sentry from MAUI customers asking about App Starts and general Mobile Vitals support.

jamescrosswell:
[...]

PR #4125, opened 2025-04-17
MAUI CommunityToolkit MVVM auto instrumentation of async commands

aritchie:
This is a POC to try to insert auto instrumentation of async operations that occur in commands.  The [MVVM Community Toolkit](https://learn.microsoft.com/en-us/dotnet/communitytoolkit/mvvm/) is insanely popular in .NET MAUI.  This will be a key component to making #3181 awesome.

## TODO
- [ ] Documentation

## Performance considerations

There was quite a bit of [back and forth](https://github.com/getsentry/sentry-dotnet/pull/4125#discussion_r2088858937) before merging this PR about the performance implications of building a service provider to resolve the options during service registration itself (since there's a bit of inception going on wiring up the MauiEventBinders on startup here). 

To put the issue to bed, I created some benchmarks. Basically I temporarily changed the implementation of the UseSentry app builder extension to take an optional bool param and do this with it:
```csharp
        // Resolve the configured options and register any element event binders from these
        SentryMauiOptions options = null!;
        if (resolveOptionsWithServiceProvider)
        {
            IServiceProvider serviceProvider = services.BuildServiceProvider();
            options = serviceProvider.GetRequiredService<IOptions<SentryMauiOptions>>().Value;
            services.TryAddSingleton<SentryOptions>(options); // Ensure this doesn't get resolved again in AddSentry
        }
        else
        {
            options = new SentryMauiOptions();
            configureOptions?.Invoke(options);
        }
 ```

So if`resolveOptionsWithServiceProvider == true` we use the existing code from this PR to get the options... otherwise we use the "fake options" hack/workaround. 

#### Benchmark results 

| Method           | ResolveOptionsWithServiceProvider | Mean     | Error      | StdDev   | Gen0    | Gen1   | Allocated |
|----------------- |---------------------------------- |---------:|-----------:|---------:|--------:|-------:|----------:|
| 'Build MAUI App' | False                             | 819.0 us | 1,542.4 us | 84.54 us | 11.7188 | 2.9297 |  99.63 KB |
| 'Build MAUI App' | True                              | 601.4 us | 1,048.6 us | 57.48 us | 11.7188 | 3.9063 |  99.45 KB |

So the hack/alternative would be about 33% slower and allocates a fraction more memory. 

As such, we left the original code in the PR (that builds a service provider to resolve the options during init). 

For transparency, here's the code I used for the benchmark:
```csharp
[ShortRunJob]
public class MvvmBenchmarks
{
    private MauiAppBuilder AppBuilder;

    [Params(true, false)]
    public bool ResolveOptionsWithServiceProvider;

    [GlobalSetup]
    public void Setup()
    {
        AppBuilder = MauiApp.CreateBuilder()
            // This adds Sentry to your Maui application
            .UseSentry(options =>
            {
                // The DSN is the only required option.
                options.Dsn = DsnSamples.ValidDsn;
                // Automatically create traces for async relay commands in the MVVM Community Toolkit
                options.AddCommunityToolkitIntegration();
            }, ResolveOptionsWithServiceProvider);
    }

    [Benchmark(Description = "Build MAUI App")]
    public void BuildMauiAppBenchmark()
    {
        AppBuilder.Build();
    }
}
```


src/Sentry.Maui.CommunityToolkitMvvm/Sentry.Maui.CommunityToolkitMvvm.csproj
bruno-garcia, line 1: Should the package name be `Sentry.CommunityToolkit.Mvvm`? We usually follow the convention of `Sentry.PackageName` following the docs of it (exception I think is Sentry.Extensions.Logging instead of `Sentry.Microsoft.Extensions...`

bruno-garcia: Or we have `Sentry.Maui.Xyz` because we tie this to our `Sentry.Maui`? I see the project dependency below now. In that case the naming names space (I'd just add the `.` between CommunityToolkit and Mvvm to match what the actual package does

src/Sentry.Maui.CommunityToolkitMvvm/CtMvvmMauiElementEventBinder.cs
```diff
@@ -0,0 +1,151 @@
...
+
+    private static List<IAsyncRelayCommand> _refs = [];
+    private static void TryBindTo(ICommand? command, bool bind)
+    {
+        if (command is IAsyncRelayCommand relayCommand)
+        {
+            if (bind)
+            {
+                // necessary for collectionview buttons
+                if (!_refs.Contains(relayCommand))
```
bruno-garcia, line 115: Is this just used to look up if it was already added? If so shouldn't we use a map here instead? 
right now this is O(n)

Another question: Is the same reference being pass in twice? Or can we rely on `IAsyncRelayCommand` having some `Equals` and `GetHashCode` overridden?

src/Sentry.Maui.CommunityToolkitMvvm/CtMvvmMauiElementEventBinder.cs
```diff
@@ -0,0 +1,151 @@
...
+
+    private static void RelayCommandOnPropertyChanged(object? sender, PropertyChangedEventArgs e)
+    {
+        if (e.PropertyName == nameof(IAsyncRelayCommand.IsRunning))
+        {
+            var relay = (IAsyncRelayCommand)sender!;
+
+            if (relay.IsRunning)
+            {
+                var transaction = SentrySdk.StartTransaction("ctmvvm", "asynccommand");
```
bruno-garcia, line 139: for testability we usually take `IHub` in an internal constructor, passing `HubAdapter.Instance` from the public one not to have to expose this to the caller.

```suggestion
                var transaction = _hub_.StartTransaction("ctmvvm", "asynccommand");
```

bruno-garcia, line 139: Could the name of the command be used as transaction name instead? 

Something like:

```suggestion
                var transaction = SentrySdk.StartTransaction(relay.CommandName, "async.command");
```

The UI will group the transactions based on name. This way you can compare the duration and operations under each command separately

src/Sentry.Maui.CommunityToolkitMvvm/CtMvvmMauiElementEventBinder.cs
```diff
@@ -0,0 +1,151 @@
...
+                var span = transaction.StartChild("run");
```
bruno-garcia, line 140: Since the span and the transaction have the same start and end time, it seems like we could get away without the span.

A transaction _is like a span_ and we'll measure timing the same way at the transaction level

src/Sentry.Maui.CommunityToolkitMvvm/MauiAppBuilderExtensions.cs
```diff
@@ -0,0 +1,32 @@
...
+        return builder;
+    }
+
+
+    /// <summary>
+    /// Installs necessary services to auto-instrument CommunityToolkit.Mvvm commands
+    /// </summary>
+    /// <param name="services">The Service Collection</param>
+    /// <returns>The Service Collection</returns>
+    public static IServiceCollection UseSentryCommunityToolkitIntegration(this IServiceCollection services)
```
bruno-garcia, line 27: Do we expect folks to call this one? Seems like the public API is just `builder.UseSentryCommunityToolkitIntegration()` which calls this one, so we could hide it from the user:


```suggestion
    internal static IServiceCollection UseSentryCommunityToolkitIntegration(this IServiceCollection services)
```

bruno-garcia:
This is quite popular: https://nugettrends.com/packages?ids=CommunityToolkit.Mvvm&months=24 

Worth adding it in

image

src/Sentry.Maui.CommunityToolkitMvvm/CtMvvmMauiElementEventBinder.cs
aritchie: I agree that normally that is a good path.  What I do worry about is allocating more pointers per control.  Using the static events keeps our footprint lower in large pages with many controls.  This probably isn't going to be a huge issue, but I wanted to start on the happy path first.  Automated testing with this thing is difficult at best.  Happy to change this, but let me know your thoughts.

src/Sentry.Maui.CommunityToolkitMvvm/CtMvvmMauiElementEventBinder.cs
aritchie: When the CollectionView is rebinding cells, this fires off events that we pick up - and WE think we're rebinding.  The issue is that collectionview is going to give you recycled cells (as it should).  We end up X binding the object.  I had a thought over the weekend of NOT storing a reference at all.  My thinking is to simply force an event unhook (-=) even if we haven't hooked it before installing a new hook.  This will prevent double binding as well as us having to look up references.  

Second question: it would be the same issue I mentioned above.  It technically is the same cell reference because of cell recycling.  In any case, with the above potential solution.... none of this matters.  

Thoughts?

src/Sentry.Maui.CommunityToolkitMvvm/CtMvvmMauiElementEventBinder.cs
aritchie: There is no command name to use.  I was looking.  I was thinking of looking up the binding it is pointed at, but that comes at a higher cost than I would like.  Even if I do get the binding, very few people add x:Name to the control, so I have to do reflection to see what the actual property name is on the viewmodel (which also kills AOT).  It's a rock & hard place.... was hoping to start this as an MVP and build on it from there.

src/Sentry.Maui.CommunityToolkitMvvm/CtMvvmMauiElementEventBinder.cs
aritchie: No problems.  You're the dude that has been telling me to use spans "cause transactions are going away".  I can whatever here :)

src/Sentry.Maui.CommunityToolkitMvvm/MauiAppBuilderExtensions.cs
aritchie: Because this is an external integration/secondary package, I can't install it without the user acting on it in their code.  We could consider "help" with this in the future using source generators.

src/Sentry.Maui.CommunityToolkitMvvm/CtMvvmMauiElementEventBinder.cs
bruno-garcia: Right, use spans if you expect those operations to be attached to a transaction at some point. E.g: An HTTP integration, attaches spans to a transaction when the transaction happens. Eventually we'll send them stand alone, when we add support for that. But today they require a transaction.

But if you're starting the transaction somewhere, you have a 'stopwatch' there already. You can have child spans for it, but if they have the same start/end time as the transaction, you can skip that child span, the transction itself works as a stop watch.

src/Sentry.Maui.CommunityToolkitMvvm/CtMvvmMauiElementEventBinder.cs
aritchie: I've added the instance based methods and ihub

aritchie: There is nothing called name or like name on the async command implementation or interfaces.  Retrieving the binding info from the control is not pretty and requires reflection.  With gesture recognizers (coming soon) and all of the other events potentially auto-breadcrumbing in the future, the user should have a pretty good idea of what's going on.  I can pass through the control type & control name (if set), but that also comes with some costs.  

I really think we should start with an MVP and let users decide what/if anything is missing.

src/Sentry.Maui.CommunityToolkitMvvm/CtMvvmMauiElementEventBinder.cs
jamescrosswell: If/when we get rid of transactions, we've going to have to go through all our automatic instrumentation and replace any transactions that we have with parentless spans... until then though, the only way to create a span without a parent and send this to Sentry is to create a transaction.

src/Sentry.Maui.CommunityToolkit.Mvvm/CtMvvmMauiElementEventBinder.cs
```diff
@@ -130,6 +130,8 @@ private void RelayCommandOnPropertyChanged(object? sender, PropertyChangedEventA
             if (relay.IsRunning)
             {
                 var transaction = hub.StartTransaction("ctmvvm", "relay.command");
+                hub.ConfigureScope(x => x.Transaction ??= transaction);
```
jamescrosswell: This is a bit curious. Maybe it's what you're wanting to do but maybe not. 

The logic you have coded currently will always create a new transaction and will sometimes set that transaction as active on the scope. That means other spans may get recorded as children of this transaction or may be recorded as children of another transaction. That seems odd to me.

By setting `scope.Transaction` you make this available for other code that would add child spans to that transaction. So, for example, our DiagnosticSource integration [checks if there is an active/sampled transaction](https://github.com/getsentry/sentry-dotnet/blob/3e70c3126515c87b1851b6a9a7afd99ffbb9aeb7/src/Sentry.DiagnosticSource/Internal/DiagnosticSource/EFDiagnosticSourceHelper.cs#L34). If so, it creates child spans under that transaction representing database operations. If not, it doesn't create any spans at all. 

Alternatively the `SentrySpanProcessor` checks to see if there is an active transaction on the scope and it [creates either a new transaction or a new child span](https://github.com/getsentry/sentry-dotnet/blob/9b29d1f7b6fe1f2638d344c5cbcb60281e16739c/test/Sentry.OpenTelemetry.Tests/SentrySpanProcessorTests.cs#L238), as appropriate. 

It might make sense for the code here to follow one of those two patterns instead then?

aritchie: Most of the fixtures I found check the transaction on the scope (for assertions and general usage).  I asked Bruno about this and he was like "ya that sounds right to use the scope".  I was then told not to create a span and a transaction.  I just want to record that something happened and the finished followed up by a successful unit test.  This has been hell trying to figure out

samples/Sentry.Samples.Maui/CtMvvmViewModel.cs
```diff
@@ -0,0 +1,15 @@
+using CommunityToolkit.Mvvm.ComponentModel;
+using CommunityToolkit.Mvvm.Input;
+
+namespace Sentry.Samples.Maui;
+
+public partial class CtMvvmViewModel : ObservableObject
+{
+    // we auto-instrument how
```
jamescrosswell, line 8: `how`? Is this left over from before the solution was implemented?

Maybe worth expanding on this for SDK users:
// Sentry automatically creates spans for code like XYZ in Mvvm apps

samples/Sentry.Samples.Maui/MauiProgram.cs
```diff
@@ -53,6 +54,10 @@ public static MauiApp CreateMauiApp()
                 fonts.AddFont("OpenSans-Semibold.ttf", "OpenSansSemibold");
             });
 
+        // additional feature that hooks into the Sentry MAUI integration for auto-instrumenting async relay command
```
jamescrosswell, line 57:
```suggestion
        // Automatically create traces for async relay commands
```

samples/Sentry.Samples.Maui/CtMvvmViewModel.cs
jamescrosswell, line 8:
```suggestion
    // Sentry automatically creates spans for async relay commands
```

scripts/device-test.ps1
```diff
@@ -24,7 +24,8 @@ try
 {
     if (!$Tfm)
     {
-        $Tfm = 'net8.0'
+        $Tfm = 'net9.0'
+        #$Tfm = 'net8.0'
```
jamescrosswell, line 28: This changes the default TFM from net8.0 to net9.0... what that intentional? Any reason to leave the old one commented out? This is a parameter that gets passed to the script anyway so it's easy to switch between one tfm and the other simply by passing a different parameter.

src/Sentry.Maui.CommunityToolkit.Mvvm/CtMvvmMauiElementEventBinder.cs
```diff
@@ -0,0 +1,157 @@
...
+    /// Unbinds from the element
+    /// </summary>
+    /// <param name="element"></param>
+    /// <exception cref="NotImplementedException"></exception>
+    public void UnBind(VisualElement element) => Iterate(element, false);
+
+
+    private void Iterate(VisualElement element, bool bind)
+    {
+        switch (element)
```
jamescrosswell, line 28: Hm, is this switch statement likely to grow over time? It seems to essentially map from a visual element to a command. Potentially we could store this default map in some options and that way people could easily add custom mappings. I'm not sure if that's necessary or over engineering for a problem that doesn't exist though.

src/Sentry.Maui.CommunityToolkit.Mvvm/CtMvvmMauiElementEventBinder.cs
```diff
@@ -0,0 +1,157 @@
...
+                TryBindTo(tap.Command, bind);
+                break;
+
+            case SwipeGestureRecognizer swipe:
+                TryBindTo(swipe.Command, bind);
+                break;
+
+            // no commands
+            //case PinchGestureRecognizer pinch
+            //case PanGestureRecognizer pan
```
jamescrosswell, line 92: Is the commented code included as an explanation or a todo or could we remove it?

test/Sentry.Maui.Device.TestApp/Sentry.Maui.Device.TestApp.csproj
```diff
@@ -1,12 +1,13 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFrameworks></TargetFrameworks>
-    <TargetFrameworks Condition="'$(NO_ANDROID)' == ''">$(TargetFrameworks);net8.0-android;net9.0-android</TargetFrameworks>
-    <TargetFrameworks Condition="'$(NO_IOS)' == '' And $([MSBuild]::IsOSPlatform('OSX'))">$(TargetFrameworks);net8.0-ios;net9.0-ios</TargetFrameworks>
+<!--    <TargetFrameworks>$(TargetFrameworks);net8.0-android</TargetFrameworks>-->
```
jamescrosswell, line 4: I think we want to test both net8.0 and net9.0 targets. 

At the moment we only run device tests for both tfms for Android, which I added in [this PR](https://github.com/getsentry/sentry-dotnet/commit/13d4a9f2038e142f9dc8c7799df06edfe17a2dd4#diff-c709796952b06c9c4ae93d17583aacd1d473093651a6de9b275ce15a2ddca0f8). It was necessary to test/implement that functionality but I didn't want the PR to bloat out to 2000 lines so I left iOS out of it. 

We should really setup the iOS device tests to run in a matrix in the same way though (rather than arbitrarily testing against one tfm or the other).

test/Sentry.Maui.Device.TestApp/Sentry.Maui.Device.TestApp.csproj
```diff
@@ -1,12 +1,13 @@
...
+<!--    <TargetFrameworks>$(TargetFrameworks);net8.0-ios</TargetFrameworks>-->
+    <TargetFrameworks>$(TargetFrameworks);net9.0-android</TargetFrameworks>
+    <TargetFrameworks>$(TargetFrameworks);net9.0-ios</TargetFrameworks>
     <!-- Pin target iOS version so that our tests don't break when new versions of Xcode are released.
       'net8.0-ios' resolves the latest version of the iOS SDK otherwise. -->
-    <TargetPlatformVersion Condition="$([MSBuild]::GetTargetPlatformIdentifier('$(TargetFramework)')) == 'ios'">18</TargetPlatformVersion>
+
```
jamescrosswell, line 10: erm... won't this break our tests when a new version of Xcode is released (per the comment, that hasn't been removed here)?

test/Sentry.Maui.Device.TestApp/Startup.cs
```diff
@@ -30,6 +32,19 @@ public static MauiApp CreateMauiApp()
                 conf.AddXunit();
             });
 
+        //             .UseVisualTestRunner(conf =>
```
jamescrosswell, line 35: Should this be commented out?

jamescrosswell (changes requested):
From what I can tell, the functionality being implemented looks OK (depending on whether you want to create parallel transactions or child spans if/when there is already an active transaction on the scope). 

It looks like there are some (perhaps unintentional) changes to other parts of the solution though, which I've added specific comments on. I think we need to get to the bottom of all of those and, obviously, the CI tests would need to pass before we can merge.

test/Sentry.Maui.Device.TestApp/Startup.cs
aritchie: Yes - It caused issues for CI.  We definitely need it when debugging unit tests in MAUI though.  This was a god send today

src/Sentry.Maui.CommunityToolkit.Mvvm/CtMvvmMauiElementEventBinder.cs
jamescrosswell: If you are creating a transaction, I think it makes sense to store this on the scope. 

My question is whether you want to create a transaction or not though? If there isn't any transaction stored on the scope then I think you do want to create one right? If there is a transaction stored on the scope though then maybe you just create a child span.

So maybe something like:
```
var span = hub.StartSpan("ctmvvm", "relay.command");
if (span is ITransactionTracer transaction)
{
    hub.ConfigureScope(x => x.Transaction ??= transaction);
}
```

test/Sentry.Maui.Device.TestApp/Sentry.Maui.Device.TestApp.csproj
aritchie: It was moved, not removed.  Line 40 is where it moved to.  It sits with the other versioning values.

src/Sentry.Maui.CommunityToolkit.Mvvm/CtMvvmMauiElementEventBinder.cs
aritchie: I don't want to overcomplicate this right now.  There is unlikely to be a lot of "out of the box" command bindings in the next few major MAUI releases.

src/Sentry.Maui.CommunityToolkit.Mvvm/CtMvvmMauiElementEventBinder.cs
aritchie: It was left as somewhat of a note for implementors in the future.  There are no commands on those, so I wanted to make it known why they were missing.  I've updated the note and moved it to the bottom of the switch

test/Sentry.Maui.Device.TestApp/Sentry.Maui.Device.TestApp.csproj
aritchie: I actually think we should remove net8 for several reasons
- dotnet android & ios are out of support on net8 already
- It takes a ton of extra time to build those additional targets
- It was really causing grief for the visual runner which I would really REALLY like to use going forward

Thoughts?

scripts/device-test.ps1
aritchie: Done

src/Sentry.Maui.CommunityToolkit.Mvvm/CtMvvmMauiElementEventBinder.cs
aritchie: IHub does not have a StartSpan.  Do I still Finish the span if I'm setting it on the scope?

aritchie: Updated - Added StartSpan as an extension (taken from internal Hub class).  Span is NOT put on scope at all.  Span (or transaction) was finished when async relay command completes

test/Sentry.Maui.Device.TestApp/Sentry.Maui.Device.TestApp.csproj
jamescrosswell: Hm, tricky to remove for Android. We'd have no way to test the older `AssemblyReader` (unless we drop support for that as well).

Could we conditionally run the visual tests only on net9.0 (to avoid whatever issues we're having with these in net8.0)?

Is so then maybe we could drop support for net8.0 in the next major release (it seems like a pretty major change to the SDK).

aritchie: I don't know how we're using a global.json where we're multitargeting different major sdks and workload sets that are tied to a specific major version.  I don't know how our device tests are even running at the moment

test/Sentry.Maui.Device.TestApp/Startup.cs
jamescrosswell: When will/would we uncomment it? When running locally? If so, we need to work out how to install the necessary dependencies to get this working on CI. 

Is there any way we can do that via directives instead?

test/Sentry.Maui.Tests/MauiEventsBinderTests.cs
```diff
@@ -14,21 +18,29 @@ private class Fixture
 
         public Fixture()
         {
-            var hub = Substitute.For<IHub>();
-            hub.When(h => h.ConfigureScope(Arg.Any<Action<Scope>>()))
+            Hub = Substitute.For<IHub>();
+            Hub.When(h => h.ConfigureScope(Arg.Any<Action<Scope>>()))
                 .Do(c => c.Arg<Action<Scope>>()(Scope));
 
+            Hub.StartSpan(default, default).ReturnsForAnyArgs(args =>
```
jamescrosswell, line 25: I don't think you can do this with extension methods. You can mock the stuff they're using in their implementation but extension methods themselves aren't virtual.

test/Sentry.Maui.Device.TestApp/Sentry.Maui.Device.TestApp.csproj
aritchie: I've brought over an improved setup I did in another pending PR for gesture recognizers.  This should hopefully fix all of this stuff up for us and makes it easy to bring in the visual runner when we need it.

test/Sentry.Maui.Device.TestApp/Startup.cs
aritchie: Updated from gestures branch that has all of the flags we need to make this proper/better

test/Sentry.Maui.Tests/MauiEventsBinderTests.cs
aritchie: Fixed

CHANGELOG.md
```diff
@@ -10,6 +13,7 @@
 ### Features
 
 - New source generator allows Sentry to see true build variables like PublishAot and PublishTrimmed to properly adapt checks in the Sentry SDK ([#4101](https://github.com/getsentry/sentry-dotnet/pull/4101))
+- Auto breadcrumbs now include all .NET MAUI gesture recognizer events ([#4124](https://github.com/getsentry/sentry-dotnet/pull/4124))
```
jamescrosswell, line 16: This is from another PR right?

aritchie: Yes.  I needed the work for the visual runner here

CHANGELOG.md
```diff
@@ -4,7 +4,9 @@
 
 ### Features
 
+- .NET MAUI integration with CommunityToolkit.Mvvm Async Relay Commands can now be auto spanned with the new package Sentry.Maui.CommunityToolkit.Mvvm ([#4125](https://github.com/getsentry/sentry-dotnet/pull/4125))
 - New source generator allows Sentry to see true build variables like PublishAot and PublishTrimmed to properly adapt checks in the Sentry SDK ([#4101](https://github.com/getsentry/sentry-dotnet/pull/4101))
+- Auto breadcrumbs now include all .NET MAUI gesture recognizer events ([#4124](https://github.com/getsentry/sentry-dotnet/pull/4124))
```
jamescrosswell, line 9: Wrong PR?

src/Sentry/HubExtensions.cs
```diff
@@ -51,6 +51,19 @@ public static ITransactionTracer StartTransaction(
         SentryTraceHeader traceHeader) =>
         hub.StartTransaction(new TransactionContext(name, operation, traceHeader));
 
+    /// <summary>
+    /// Starts a span or transaction if a transaction is not already active on the scope
+    /// </summary>
+    public static ISpan StartSpan(this IHub hub, string operation, string description)
```
Flash0ver, line 57: question: Could we keep this method `internal` for now, to not increase the public surface area?

(since the `class Hub` it was moved from is also `internal`)

src/Sentry.Maui.CommunityToolkit.Mvvm/Sentry.Maui.CommunityToolkit.Mvvm.csproj
```diff
@@ -0,0 +1,27 @@
...
+    <Description>MAUI and Community Toolkit integration for Sentry - Open-source error tracking that helps developers monitor and fix crashes in real time.</Description>
+    <TargetFrameworks>net9.0;net8.0</TargetFrameworks>
+
+    <ImplicitUsings>enable</ImplicitUsings>
+    <Nullable>enable</Nullable>
+    <UseMaui>true</UseMaui>
+
+    <!-- Skip warnings about including Microsoft.Maui.Controls... that should be done from the application -->
+    <SkipValidateMauiImplicitPackageReferences>true</SkipValidateMauiImplicitPackageReferences>
+    <RootNamespace>Sentry.Maui.CommunityToolkitMvvm</RootNamespace>
```
Flash0ver, line 13: question: Should the root namespace match the project/package name?

I.e. missing `.` before `Mvvm`.
Or is the difference intentional?

src/Sentry.Maui.CommunityToolkit.Mvvm/MauiAppBuilderExtensions.cs
```diff
@@ -0,0 +1,32 @@
...
+        return builder;
+    }
+
+
+    /// <summary>
+    /// Installs necessary services to auto-instrument CommunityToolkit.Mvvm commands
+    /// </summary>
+    /// <param name="services">The Service Collection</param>
+    /// <returns>The Service Collection</returns>
+    public static IServiceCollection UseSentryCommunityToolkitIntegration(this IServiceCollection services)
```
Flash0ver, line 27: question: Naming: `Use` vs `Add`.

Should we prefix this method (or even both) with `Add` instead of `Use`, since it's on the `IServiceCollection`?
To be fair, I'm a bit unfamiliar with the MAUI conventions.

src/Sentry.Maui.CommunityToolkit.Mvvm/MauiAppBuilderExtensions.cs
```diff
@@ -0,0 +1,32 @@
+using Sentry.Maui.CommunityToolkitMvvm;
```
Flash0ver, line 1: question: Namespace name:

```suggestion
using Sentry.Maui.CommunityToolkit.Mvvm;
```

Or is the "missing `.`" intentional?

src/Sentry.Maui.CommunityToolkit.Mvvm/CtMvvmMauiElementEventBinder.cs
```diff
@@ -0,0 +1,149 @@
+using System.Windows.Input;
+using CommunityToolkit.Mvvm.Input;
+
+namespace Sentry.Maui.CommunityToolkitMvvm;
+
+/// <summary>
+/// Scans all elements for known commands that are implement
+/// </summary>
+public class CtMvvmMauiElementEventBinder(IHub hub) : IMauiElementEventBinder
```
Flash0ver, line 9: question(naming): is `CtMvvm` a well-known abbreviation for `CommunityToolkit.Mvvm`?

To me this type name doesn't read like idiomatic .NET.
Unless `CtMvvm` is a super common abbreviation used by the community ... but I personally have never encountered it so far.

On the other hand, a full `CommunityToolkitMvvmMauiElementEventBinder` is a bit unwieldy.

Could we get away with making this type `internal` for now?
Since I believe it's meant to be added to DI via the `MauiAppBuilder`-Extension method.
Or does user code need to refer to it?

Also:
From what I see, the other `IMauiElementEventBinder`-derived types we have are prefixed with `Maui`.
Does it make sense to go with `MauiCommunityToolkitMvvmEventsBinder` instead?
And/or make the type `internal`?

What do you think?

src/Sentry.Maui.CommunityToolkit.Mvvm/MauiAppBuilderExtensions.cs
aritchie: MAUI doesn't generally use the concept of builder vs post build like aspnet.  The convention is use on the maui sside

src/Sentry/HubExtensions.cs
aritchie: I feel this method is a good thing in general.  I plan to use it externally to Sentry.

src/Sentry.Maui.CommunityToolkit.Mvvm/CtMvvmMauiElementEventBinder.cs
aritchie: Renamed - I don't want to debate naming.  Let's just get it done.

samples/Sentry.Samples.Maui/MauiProgram.cs
```diff
@@ -53,6 +54,10 @@ public static MauiApp CreateMauiApp()
                 fonts.AddFont("OpenSans-Semibold.ttf", "OpenSansSemibold");
             });
 
+        // Automatically create traces for async relay commands
+        // in the MVVM Community Toolkit
+        builder.Services.UseSentryCommunityToolkitIntegration();
```
jamescrosswell, line 59: TLDR; can we make `UseSentryCommunityToolkitIntegration` an extension method on `SentryMauiOptions` instead of `IServiceCollection`?

All of our other integrations get initialised via the SentryOptions when initialising the Sentry SDK. In this case, maybe it's not an `ISdkIntegration` like those but it'd be nice if it was done via the `SentryMauiOptions` in the call to `UseSentry`.

We already [register a bunch of others](https://github.com/getsentry/sentry-dotnet/blob/08a82bc9490e2e2337e2815f8f80124f968a2dbc/src/Sentry.Maui/SentryMauiAppBuilderExtensions.cs#L58-L62) in that method...

src/Sentry.Maui.CommunityToolkit.Mvvm/Sentry.Maui.CommunityToolkit.Mvvm.csproj
```diff
@@ -0,0 +1,27 @@
...
+    <Nullable>enable</Nullable>
+    <UseMaui>true</UseMaui>
+
+    <!-- Skip warnings about including Microsoft.Maui.Controls... that should be done from the application -->
+    <SkipValidateMauiImplicitPackageReferences>true</SkipValidateMauiImplicitPackageReferences>
+    <RootNamespace>Sentry.Maui.CommunityToolkit.Mvvm</RootNamespace>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="CommunityToolkit.Mvvm" Version="8.4.0"/>
```
jamescrosswell, line 17: Do we need to use the latest and greatest [version of this package](https://www.nuget.org/packages/CommunityToolkit.Mvvm#versions-body-tab)? Ideally we'd depend on the oldest version of this package that works functionally... that way SDK users don't have to upgrade this package to use Sentry (but they could use the latest and greatest as well if they wanted).

samples/Sentry.Samples.Maui/MauiProgram.cs
aritchie: I need dependency injection and IServiceCollection.  SentryMauiOptions does not have that.  You really need a builder pattern if you want to create plugins like that.

src/Sentry/HubExtensions.cs
jamescrosswell: Yeah I think this is a pretty common operation for SDK users 👍🏻

jamescrosswell (changes requested):
It's looking pretty good. I'd like to find a way to enable this integration from within the call to `UseSentry` is all. 

Wherever possible, we try to keep the initialisation of Sentry to a single line of code. There are rare ocassions where this can't be done but it shouldn't be too hard here I don't think.

src/Sentry.Maui.CommunityToolkit.Mvvm/Sentry.Maui.CommunityToolkit.Mvvm.csproj
aritchie: I wanted to make sure I didn't have a range of versions to deal with.  I do this with open source as well.  If people have a need for lesser versions, it can be obviously be discussed, but we have exactly 1 feature requests for this right now - me :)

samples/Sentry.Samples.Maui/MauiProgram.cs
jamescrosswell: How about if the call to `UseSentryCommunityToolkitIntegration` stores something in on internal property like `SentryMauiOptions.CustomEventBinders` or something like that. 

Then the call to `UseSentry`, in addition to explicitly registering the event binders it already registers, also checks if there are any `CustomEventBinders` in the options and registers those as well.

That should work right?

src/Sentry.Maui.CommunityToolkit.Mvvm/Sentry.Maui.CommunityToolkit.Mvvm.csproj
jamescrosswell: That optimises for us as the maintainers (it makes our job really easy if we only have one potential version to deal with).

As a general principal, as SDK authors we try to target the oldest versions of things that we possibly can. The reason being that that generally makes it easier for our users. If we give them the greatest possible set of options about what versions of dependencies they need, we constrain them less in the complex system that they're building.

If this PR is genuinely just for an audience of 1, let's not waste any more time on it. Assuming it's for a wider audience than that though, the principal above as served us well in the past.

src/Sentry.Maui.CommunityToolkit.Mvvm/MauiCommunityToolkitMvvmEventsBinder.cs
```diff
@@ -0,0 +1,149 @@
...
+    private void RelayCommandOnPropertyChanged(object? sender, PropertyChangedEventArgs e)
+    {
+        if (e.PropertyName == nameof(IAsyncRelayCommand.IsRunning))
+        {
+            var relay = (IAsyncRelayCommand)sender!;
+
+            if (relay.IsRunning)
+            {
+                var span = hub.StartSpan(SpanName, SpanOp);
+                if (span is ITransactionTracer transaction)
```
jamescrosswell, line 136: There's something very unusual going on here. The first time this event triggers, there's no transaction on the scope and so a transaction gets created here.

Further down on line 145 (after the transaction is finished) the scope has been correctly reset and there's no transaction on the scope anymore.

So far so good...

However, the second time this event triggers, for some reason there's a transaction on the scope... and so a span gets created here (rather than a transaction). 

The result is that this integration currently only ever creates one trace per AsyncRelayCommand! 🐞 

No running theories currently - needs investigation.

jamescrosswell: Ah, I think I see what's going on here. Most stuff on the Scope is global in MAUI cause we IsGlobalMode is always true for MAUI apps. Transactions are the exception:
https://github.com/getsentry/sentry-dotnet/blob/8abc247566132a4b284d7fe24c44a93d84d32076/src/Sentry/Scope.cs#L203

And in this particular instance, we're creating spans/transactions in for an AsyncRelayCommand... it seems to start on the main thread but the completion happens on a different thread so on the first run through, everything looks fine (it looks like we reset the transaction on the scope). On the second run through we're in a different asynclocal context and so the transaction hasn't been reset on the scope (it's Scope.Transaction is `AsyncLocal<T>`). 

Tricky...

src/Sentry.Maui.CommunityToolkit.Mvvm/MauiCommunityToolkitMvvmEventsBinder.cs
```diff
@@ -0,0 +1,149 @@
...
+    private void TryBindTo(ICommand? command, bool bind)
+    {
+        if (command is IAsyncRelayCommand relayCommand)
+        {
+            // since events can retrigger binding pickups, we want to ensure we unhook any previous event handlers
+            // instead of storing a ref to know if we've already bound to an event or not
+            relayCommand.PropertyChanged -= RelayCommandOnPropertyChanged;
+            if (bind)
+            {
+                relayCommand.PropertyChanged += RelayCommandOnPropertyChanged;
```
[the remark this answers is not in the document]
jamescrosswell: Should be addressed in commit [52ef07a](https://github.com/getsentry/sentry-dotnet/pull/4125/commits/52ef07ab887b302f5b5cc1d718a0796d48f2a255)

src/Sentry.Maui.CommunityToolkit.Mvvm/MauiCommunityToolkitMvvmEventsBinder.cs
jamescrosswell: I think the simplest solution is not to store transactions representing AsyncRelayCommands on the Scope... 

This means other integrations (like DB integrations or our SentryMessageHandler) won't create child spans for these.

I can't think of any other solutions off the top of my head though.

src/Sentry.Maui.CommunityToolkit.Mvvm/MauiCommunityToolkitMvvmEventsBinder.cs
```diff
@@ -0,0 +1,149 @@
...
+            }
+        }
+    }
+
+
+    private readonly ConcurrentDictionary<IAsyncRelayCommand, ISpan> _contexts = new();
```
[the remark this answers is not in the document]
bruno-garcia: Curious if that works OK since I dunno where those objects come from. If they are kept around or somehow when events fire new instances come out?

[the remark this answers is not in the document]
jamescrosswell: We're basically looking for a way associate AsyncRelayCommands with Spans right? 

This line uses a ConditionalWeakTable to make that association:
https://github.com/getsentry/sentry-dotnet/blob/52ef07ab887b302f5b5cc1d718a0796d48f2a255/src/Sentry.Maui.CommunityToolkit.Mvvm/MauiCommunityToolkitMvvmEventsBinder.cs#L152

And then when the RelayCommand is finished, this bit looks up the associated Span and finishes it:
https://github.com/getsentry/sentry-dotnet/blob/52ef07ab887b302f5b5cc1d718a0796d48f2a255/src/Sentry.Maui.CommunityToolkit.Mvvm/MauiCommunityToolkitMvvmEventsBinder.cs#L154-L157

So the same thing as what we had with the `_contexts` before, only using a ConditionalWeak reference, so if the `AsyncRelayCommand` gets disposed of before the PropertyChange event fires, the item gets dropped from the table (no memory leak).

src/Sentry.Maui/SentryMauiAppBuilderExtensions.cs
```diff
@@ -55,10 +56,16 @@ public static MauiAppBuilder UseSentry(this MauiAppBuilder builder,
         services.AddSingleton<IConfigureOptions<SentryMauiOptions>, SentryMauiOptionsSetup>();
         services.AddSingleton<Disposer>();
 
-        services.AddSingleton<IMauiElementEventBinder, MauiButtonEventsBinder>();
-        services.AddSingleton<IMauiElementEventBinder, MauiImageButtonEventsBinder>();
-        services.AddSingleton<IMauiElementEventBinder, MauiGestureRecognizerEventsBinder>();
-        services.AddSingleton<IMauiElementEventBinder, MauiVisualElementEventsBinder>();
+        // Resolve the configured options and register any element event binders from these
+        IServiceProvider serviceProvider = services.BuildServiceProvider();
```
aritchie, line 60: This will murder startup performance for MAUI apps.  Building the container is one of the most expensive operations in startup.  We're building all of the users services here as well.

test/Sentry.Maui.Tests/MauiEventsBinderTests.CtMvvmAsyncRelayCommands.cs
```diff
@@ -0,0 +1,37 @@
...
+        _fixture.Scope.Transaction.Should().NotBeNull("transaction should have been created");
+    }
+}
+
+public partial class TestCtMvvmViewModel : ObservableObject
+{
+    [RelayCommand]
+    private async Task Test()
+    {
+        await Task.Delay(TimeSpan.FromSeconds(1));
```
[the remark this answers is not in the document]
aritchie: I wanted to make sure there was a span.  A yield wasn't always allowing the transaction to take place in time due to the long task.  

> It's definitely not audience of 1. Its got [15M downloads over the last few years](https://nugettrends.com/packages?months=72&ids=CommunityToolkit.Mvvm). Compared to

You have no specific users asking for this right now.  I asked for this for mobile vitals.  My point is that we could release something and refine it if users decide to start using this.

test/Sentry.Maui.Tests/MauiEventsBinderTests.CtMvvmAsyncRelayCommands.cs
```diff
@@ -0,0 +1,37 @@
...
+        {
+            Command = vm.TestCommand
+        };
+        var binder = new MauiCommunityToolkitMvvmEventsBinder(_fixture.Hub);
+        binder.Bind(button, _ => { });
+
+        button.Command.Execute(null!);
+
+        // we need to wait for async command to run
+        await Task.Delay(TimeSpan.FromSeconds(1.1));
```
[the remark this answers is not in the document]
aritchie: Agreed.  I could use a time provider here but I didn't want to make this test even more complicated.

src/Sentry.Maui.CommunityToolkit.Mvvm/MauiCommunityToolkitMvvmEventsBinder.cs
```diff
@@ -0,0 +1,159 @@
...
+
+/// <summary>
+/// Scans all elements for known commands that are implement
+/// </summary>
+internal class MauiCommunityToolkitMvvmEventsBinder(IHub hub) : IMauiElementEventBinder
+{
+    private const string SpanName = "ctmvvm";
+    private const string SpanOp = "relay.command";
+
+    private readonly object _propertyChangedLock = new();
```
jamescrosswell, line 15: We probably don't need this... we can just lock against the target AsyncRelayCommand instead.

src/Sentry.Maui/SentryMauiAppBuilderExtensions.cs
aritchie: This will build/assemble all services registered up to that point which will be numerous MAUI services since they get first crack at service collection.  Profiling this won't matter because it depends on the app.  You really only ever want to build the container once

jamescrosswell: Surely not. There's no point in building transient services in advance... even singletons I'd expect it do do lazily (build them as needed). I could see it maybe building a dependency graph in advance... 

In any event, probably need to dig into the internals of ServiceCollection and ServiceProvider to find out.

src/Sentry.Maui.CommunityToolkit.Mvvm/Sentry.Maui.CommunityToolkit.Mvvm.csproj
```diff
@@ -0,0 +1,27 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <Description>MAUI and Community Toolkit integration for Sentry - Open-source error tracking that helps developers monitor and fix crashes in real time.</Description>
+    <TargetFrameworks>net9.0;net8.0</TargetFrameworks>
```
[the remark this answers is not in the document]
jamescrosswell: We could drop that target yes (there's no point in targeting out of date TFMs for the initial launch of something). 

We'll need to refactor the tests though. Currently all of the tests for `MauiCommunityToolkitMvvmEventsBinder` are in the `Sentry.Maui.Tests` project, which needs to target both `net8.0` and `net9.0`. We really should have a separate test project for these anyway (same as all of our other integrations - one to one match between the NuGet/Assembly and the Test Project).

test/Sentry.Maui.Tests/MauiEventsBinderTests.CtMvvmAsyncRelayCommands.cs
```diff
@@ -0,0 +1,37 @@
...
+
+        // Assert
+        _fixture.Scope.Transaction.Should().NotBeNull("transaction should have been created");
```
jamescrosswell, line 26: This test will always pass, even when `button.Command.Execute()` isn't called... since the transaction gets initialised to a null value in the Test Fixture.

src/Sentry.Maui.CommunityToolkit.Mvvm/MauiCommunityToolkitMvvmEventsBinder.cs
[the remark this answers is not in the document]
jamescrosswell: Ah - I see what you're saying... looking at [the source code](https://github.com/CommunityToolkit/dotnet/blob/657c6971a8d42655c648336b781639ed96c2c49f/src/CommunityToolkit.Mvvm/Input/AsyncRelayCommand.cs#L197), the instance of `relay` that is passed into the matching calls to the PropertyChanged event will be consistent... so the technique above should work in theory.

More importantly, it actually works in our sample app:
image

src/Sentry.Maui/SentryMauiAppBuilderExtensions.cs
jamescrosswell: So I checked the source code. Calling `BuildServiceProvider` doesn't, in itself, create any of the services that are registered. It creates an instance of `ServiceProvider`, passing in a bunch of ServiceDescriptors (which come from the the `ServiceCollection`). 

Instances of dependencies get created when you try to resolve them via `IServiceProvider.GetService` and `IServiceProvider.GetRequiredService` etc.

If we're really worried about this, I can use an alternative solution here and/or try to run some benchmarks... from what I can tell, it's not a real problem though.

bruno-garcia: Measuring it will end the debate. 
Even a simple stopwatch just to get an idea of this might help clarify the real impact

bruno-garcia (approved):
worth profiling the Init code just to see if building the container has any significant bottleneck (or anything else we added there).

But outside of that,lgtm

aritchie:
People using MAUI tend to change the container library.  It is extremely common.  You want to profile all of them?

src/Sentry.Maui/SentryMauiAppBuilderExtensions.cs
jamescrosswell: k - I created some proper benchmarks and the alternative solution to this would be about 33% slower... so merging as is. 

Details in the PR description, so they don't get lost when resolving this discussion.

Syed-RI:
docs on usage please?

src/Sentry.Maui/SentryMauiAppBuilderExtensions.cs
Redth: @jamescrosswell @bruno-garcia this is really not good practice, especially in .NET MAUI where we fight very hard for every millisecond of startup performance we can reduce on mobile devices.  Adding more overhead here seems to be unnecessary if it's just about trying to auto instrument another library, where you could instead add another API call to avoid the hit.  Please reconsider this change.

aritchie: https://github.com/getsentry/sentry-dotnet/pull/4208

jamescrosswell:
> docs on usage please?

Wow - I think you must have asked for this at the precise moment this new integration was released 😄 

There aren't any docs yet but there's an example of how to enable the integration in Sentry.Samples.Maui:
https://github.com/getsentry/sentry-dotnet/blob/34d849038e3e23868a12d56ea256c36b0a7f3a61/samples/Sentry.Samples.Maui/MauiProgram.cs#L29-L35

Once enabled, the new integration automatically creates traces for any MVVM AsyncRelayCommands in the application.

This new integration is pretty minimal at the moment but we're hoping to get feedback from SDK about it's usefulness and whether/how it might be extended or tweaked to be even better.

jamescrosswell:
> Better yet, how do I use this in my Avalonia app?

@BrycensRanch this particular integration is MAUI specific. It leverages event hooks in the [Microsoft.Maui.Controls.Application](https://learn.microsoft.com/en-us/dotnet/api/microsoft.maui.controls.application?view=net-maui-9.0&viewFallbackFrom=net-9.0) class to detect when an AsynReplayCommand starts or stops and start/finish a tracing span that represents that relay command. I think Avalonia would require a separate solution.

Syed-RI:
> > docs on usage please?
> 
> Wow - I think you must have asked for this at the precise moment this new integration was released 😄
> 
> There aren't any docs yet but there's an example of how to enable the integration in Sentry.Samples.Maui:
> 
> https://github.com/getsentry/sentry-dotnet/blob/34d849038e3e23868a12d56ea256c36b0a7f3a61/samples/Sentry.Samples.Maui/MauiProgram.cs#L29-L35
> 
> Once enabled, the new integration automatically creates traces for any MVVM AsyncRelayCommands in the application.
> 
> This new integration is pretty minimal at the moment but we're hoping to get feedback from SDK about it's usefulness and whether/how it might be extended or tweaked to be even better.

Yes I did ask the question right after I have received notification of a new release 😄 .

Just to be clear -

- To use and add `options.AddCommunityToolkitIntegration(); ` you will have to add `Sentry.Maui.CommunityToolkit.Mvvm` package?
- Without it, the SDK will continue to work as it was before without impacting the app startup (concern raise by @Redth https://github.com/getsentry/sentry-dotnet/pull/4125#discussion_r2101064546) ? Im asking dumb question to understand whether updating to 5.8.0 without the `Sentry.Maui.CommunityToolkit.Mvvm` package will have any impact on the app start as everyone is quite touchy about it.

jamescrosswell:
> * To use and add `options.AddCommunityToolkitIntegration(); ` you will have to add `Sentry.Maui.CommunityToolkit.Mvvm` package?

Yes, indeed. That's where that extension on the SentryOptions is defined. Sorry, I should have mentioned that.

> Without it, the SDK will continue to work as it was before without impacting the app startup (concern raise by @Redth [MAUI CommunityToolkit MVVM auto instrumentation of async commands #4125 (comment)](https://github.com/getsentry/sentry-dotnet/pull/4125#discussion_r2101064546)) ? 

No, the comments on potential performance impact relate to code changes in the `Sentry.Maui` package. However in the benchmarks that I've run (which I just updated about 10 minutes ago) the the performance impact would be in the vicinity of 6-10 milliseconds (although that is admittedly running on a Mac M2 - not on a crappy Android device). 

I pushed up a [branch containing those benchmarks](https://github.com/getsentry/sentry-dotnet/pull/4213). I had to break the software a bit in order to build them, so we'll never merge that branch but it means other folks can run the benchmarks themselves to reproduce the results... so we can make decisions based on data rather than strong opinions. Data tends to argue less 😜 

@Redth @aritchie @bruno-garcia on the back of the benchmarks, it looks like we could shave 10 ms off startup by creating the config options and invoking the callback ourselves. So we could do that in another PR. It would be a trivial change.

Syed-RI:
> it could be more impactful on a really old/slow mobile device)

That could be potentially dealbreaker as enterprises devices like Zebra and Honeywell (most of our target audience) tend to run on archaic chips with low memory but I dont have an way of benchmarking them at the moment to compare Apple with Apple.

jamescrosswell:
> That could be potentially dealbreaker

@Syed-RI I spotted a potential issue with the benchmarks... I think a lot of what needed to be measured was hidden in the GlobalSetup method of the benchmark tests. I corrected this and re-ran them. The new benchmarks actually show the existing code/solution to be the fastest option

Feel free to [check the benchmarks out for yourself](https://github.com/getsentry/sentry-dotnet/pull/4213).

Again, the difference between all of the three options is less than 12 millionths of a second, which on a machine that is 1000 times slower than mine would be around 10 milliseconds (not something a human being would perceive)

jamescrosswell:
Closing this particular PR for comments as the discussion is getting quite long (and off topic). 

If you've got any performance concerns, please add comments on the related Benchmarking PR instead (so that we can use data to drive the discussion):
https://github.com/getsentry/sentry-dotnet/pull/4213

aritchie:
> I pushed up a https://github.com/getsentry/sentry-dotnet/pull/4213. I had to break the software a bit in order to build them, so we'll never merge that branch but it means other folks can run the benchmarks themselves to reproduce the results... so we can make decisions based on data rather than strong opinions. Data tends to argue less 😜

@jamescrosswell  - you're benchmarks don't account for real world situations. The "strong opinions" isn't a fair comment.  We're all professionals and some of us have lots of experience in different areas.  So let me summarize some additional reasoning for why I'm 1000% against doing what this PR is doing  

1. .NET MAUI developers don't just use the MSFT extensions DI.  There are projects out there that I've actually worked with that use DryIoc (prism - though a mutable container) and autofac (this is the killer - the performance for building the container is awful) which verifies the dependency references regardless of the buildservicecontainer flag.  I don't think we want to get into the market of testing every DI container here. 

2. Your tests assume a minimal set of dependencies.  Depending on where UseSentry is in the setup phase, depends on how large of a chain you actually end up building the container with.  Your test isn't up against anything but UseSentry.  Real world scenario may have 100s of services registered. Let's take this pig of a method from a real world project. Note where sentry is in this chain

```csharp
public static MauiApp CreateMauiApp() => MauiApp
        .CreateBuilder()
        .UseMauiApp<App>()
        .ConfigureServices(x =>
        {
            // this method is about 40 lines longer, so just a sample here is fine
            x.AddMauiBlazorWebView();
            x.AddMudServices();
            x.AddSingleton<ILogEnricher, Redact>();
            x.AddSingleton(TimeProvider.System);
            x.AddSingleton<SqliteConnection>();
            x.AddJob(typeof(Redact));
            x.AddBluetoothLE();
            x.AddGps<Redact>();
            x.AddGeofencing<Redact>();
            x.AddHttpTransfers<Redact>();
            #if DEBUG
            x.AddBlazorWebViewDeveloperTools();
            #endif
        })
        .UseShiny()
        .UseBarcodeScanning()
        .UseMauiCommunityToolkit()
        .UseMauiCommunityToolkitMediaElement()
        .UseMauiCommunityToolkitMarkup()
        .UseBottomSheet()
        .UseMauiCameraView()
        .UseBiometricAuthentication()
        .UseSettingsView()
        .UsePrism(
            new DryIocContainerExtension(), // different DI container because prism requires mutable containers 
            prism => prism.CreateWindow("redacted")
        )
        .UseMauiMaps()
        .UseCardsView()
        .AddRemoteConfigurationMaui("https://redacted")
        .UseSentry(options =>
        {
            options.Dsn = "redacted";
        })
        .ConfigureFonts(fonts =>
        {
            //.... redact
        })
        .Build();
```

3. What I've proposed in the following PR is small, easy for users to implement, and has zero risks of anything @redth & myself have stated previously and in this note directly.  https://github.com/getsentry/sentry-dotnet/pull/4208 

CC: @bruno-garcia

jamescrosswell:
@aritchie I'm happy to look at your alternative solution to the problem once that PR is passing all the CI tests.

Do you think you can also put together some benchmarks illustrating the scenario you're describing (maybe something using Autofac and/or emulating a real world / pig setup method)?

PR #4130, opened 2025-04-22
chore: `InternalVisibleTo` Unity SDK tests

bitsandfoxes:
I'd like to have some internals accessible in the tests.
Additionally, I removed the `internal SetTrace`. It is just a wrapper for `scope.SetPropagationContext` and the Unity SDK now has access to `internal` and the wrapper is no longer needed.

#skip-changelog

### DIFF
--- a/src/Sentry/SentrySdk.cs
+++ b/src/Sentry/SentrySdk.cs
@@ -210,13 +210,6 @@ public static partial class SentrySdk
     }
 
     /// <summary>
-    /// Allows to set the trace
-    /// </summary>
-    internal static void SetTrace(SentryId traceId, SpanId parentSpanId) =>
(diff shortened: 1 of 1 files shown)


Flash0ver (approved):

Flash0ver (approved):

src/Sentry/SentrySdk.cs
```diff
@@ -209,13 +209,6 @@ internal static IDisposable UseHub(IHub hub)
         return new DisposeHandle(hub);
     }
 
-    /// <summary>
-    /// Allows to set the trace
-    /// </summary>
-    internal static void SetTrace(SentryId traceId, SpanId parentSpanId) =>
```
Flash0ver, line 215: note: was added in #4026

PR #4133, opened 2025-04-23
feat: Associate replays with errors and traces on Android

jamescrosswell:
Part of #2136

Associate Errors and Traces with the active Session Replay (if one exists) on Android.

## Protocol

Although there is a new [Replay context](https://develop.sentry.dev/sdk/data-model/event-payloads/contexts/#replay-context), I found that just setting the replay_id on that context (e.g. for an event) did not result in the event showing in the errors pane for session replays. 

What was recommended, and work better, is to store the `replay_id` on the DSC (which will accompany all events) and let Relay unpack it from there. The [DSC specification](https://develop.sentry.dev/sdk/telemetry/traces/dynamic-sampling-context/#dsc-specification) mentions `replay_id` as an optional property on the dynamic sampling context.

## Getting the Replay ID

There are three ways the DSC can be created:
1. From the currently active transaction
2. From the propagation context
3. From a BaggageHeader (in distributed tracing scenarios)

In all 3 cases, we check to see if there's a Session Replay actively being recorded. If there is, we either set (or overwrite) the `replay_id` that int he DSC. If there's not, we just propagate whatever is in the header.

This is a bit different to what we normally do with the baggage headers, so there are quite a few new scenarios to unit test.


src/Sentry/Internal/Hub.cs
```diff
@@ -471,12 +473,9 @@ private SentryId CaptureEvent(SentryEvent evt, SentryHint? hint, Scope scope)
         {
             // We get the span linked to the event or fall back to the current span
             var span = GetLinkedSpan(evt) ?? scope.Span;
-            if (span is not null)
+            if (span is not null && span.IsSampled is not false)
```
jamescrosswell, line 476: I'm not actually sure if we want the `&& span.IsSampled is not false` bit in this condition. Potentially we always apply the transaction context to the event, even if the transaction is not sampled? What if people are using complex sampling strategies?

That's probably an issue for another PR though. This PR corrects that problem with no context being applied to the event (trace or propagation)... to guarantee the replay_id will be sent.

bruno-garcia: @AbhiPrasad will know the right way here

My understanding is that we always propagate this, so that downstream they know our decision.

AbhiPrasad: > Potentially we always apply the transaction context to the event, even if the transaction is not sampled

I assume you're referring to the `trace` context here, which yes we should always apply regardless if there is a sampled transaction.

src/Sentry/Internal/ReplaySession.cs
```diff
@@ -0,0 +1,33 @@
+#if __ANDROID__
+using Sentry.Android.Extensions;
+#endif
+
+namespace Sentry.Internal;
+
+internal interface IReplaySession
```
Flash0ver, line 7: question: Is "(I)ReplaySession" the right name?

To be fair, I'm not too familiar with Session-Replay yet. And also, it's an `internal` API.
But I am uncertain whether I got confused here a little.
Is this indeed representing a "Replay-Session" ... or is this more of an "Replay-Session-Accessor" ... I apparently think about the `IHttpContextAccessor`, which the ASP.NET Core team admitted isn't the best pattern either.

TL;DR; not sure if there would be a better name?
Or is this name matching native concepts (which which I'm admittedly not too familiar with yet)

src/Sentry/Protocol/Replay.cs
```diff
@@ -0,0 +1,88 @@
+using Sentry.Extensibility;
+using Sentry.Internal;
+using Sentry.Internal.Extensions;
+
+namespace Sentry.Protocol;
+
+/// <summary>
+/// Sentry Replay context interface.
+/// </summary>
+/// <example>
```
Flash0ver, line 10: nitpick: Would you mind wrapping the code in the example with `<code> .. </code>`?

src/Sentry/Protocol/Replay.cs
```diff
@@ -0,0 +1,88 @@
...
+/// <see href="https://develop.sentry.dev/sdk/data-model/event-payloads/contexts/#replay-context"/>
+public sealed class Replay : ISentryJsonSerializable, ICloneable<Replay>, IUpdatable<Replay>
+{
+    /// <summary>
+    /// Tells Sentry which type of context this is.
+    /// </summary>
+    public const string Type = "replay";
+
+    /// <summary>
+    /// The name of the runtime.
```
Flash0ver, line 28: issue: The comment seems wrong to me.

src/Sentry/Protocol/Replay.cs
```diff
@@ -0,0 +1,88 @@
...
+        }
+    }
+
+    /// <inheritdoc />
+    public void WriteTo(Utf8JsonWriter writer, IDiagnosticLogger? logger)
+    {
+        writer.WriteStartObject();
+
+        writer.WriteString("type", Type);
+        writer.WriteSerializableIfNotNull("replay_id", ReplayId, logger);
```
Flash0ver, line 71: question: Could it be problematic if `ReplayId` does not have a value?

`replay_id` is _required_.
But we only write it if it has a value.

Could we run into this issue, or this this guarded against somewhere else?

test/Sentry.Tests/TransactionProcessorTests.verify.cs
```diff
@@ -1,3 +1,5 @@
+using Sentry.Tests.Internals;
```
Flash0ver, line 1: question: Is this required? Or a leftover?

src/Sentry/Protocol/Replay.cs
jamescrosswell: It doesn't appear to cause any trouble. In fact, Relay doesn't seem to care what is in the Replay context - it reads the value from the DSC and populates the Replay context automatically based on what it finds there. 

I see what you're saying though. According to [the docs](https://develop.sentry.dev/sdk/data-model/event-payloads/contexts/#replay-context) it's required. The docs also say:
> This should not be sent by SDKs in most cases, as the Dynamic Sampling Context will be plucked in relay, but for current versions of python and PHP these SDKs will send this context as they do not currently support the envelope endpoint.

So probably we should remove the Replay context altogether.

src/Sentry/Internal/ReplaySession.cs
jamescrosswell: I just made this name up. All it does is get the ReplayId (from the Java SDK in the case of Android applications and from some mock in the case of unit tests). I figured when it was accessed it would read like:
```csharp
var replayId = ReplaySession.ActiveReplayId
```

So it gets the `ActiveReplayId` for the "current replay session". If there is not replay session then the `ActiveReplayId` will be null. Seemed like it would read well.

src/Sentry/DynamicSamplingContext.cs
```diff
@@ -184,18 +209,20 @@ public static DynamicSamplingContext CreateFromPropagationContext(SentryPropagat
...
+            replaySession: replaySession
+            );
     }
 }
 
 internal static class DynamicSamplingContextExtensions
 {
-    public static DynamicSamplingContext? CreateDynamicSamplingContext(this BaggageHeader baggage)
-        => DynamicSamplingContext.CreateFromBaggageHeader(baggage);
+    public static DynamicSamplingContext? CreateDynamicSamplingContext(this BaggageHeader baggage, IReplaySession? replaySession)
```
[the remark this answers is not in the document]
jamescrosswell: Initially I made it a required parameter so that I could discover all the different call sites (and had to deliberately deal with them)... that way I didn't miss anything out. 

Now that the code is built, it would be possible to give the `replaySession` parameter a default value of `null`... so no additional overload required and the ASP.NET and ASP.NET Core call sites could ignore it.

src/Sentry/Internal/Hub.cs
jamescrosswell: Cool, I'll update accordingly. Thanks @AbhiPrasad !

Flash0ver (commented):
LGTM, but I don't dare to approve yet ... I still need to get more familiar with _Android_.

src/Sentry/DynamicSamplingContext.cs
```diff
@@ -144,10 +161,18 @@ private DynamicSamplingContext(
             }
             items.Add("sample_rand", rand.ToString("N4", CultureInfo.InvariantCulture));
         }
+
+        if (replaySession?.ActiveReplayId is { } replayId)
+        {
+            // Overwrite any existing value - the DSC is simply used as a transport mechanism so that SDKs can
+            // communicate the replayId to Sentry Relay (SDKs don't need to propagate the replayId to each other).
```
bruno-garcia, line 168: The SDKs do send the replay_id to each other. Through the `baggage` header. So that downstream SDKs can tag events with the replay_id (no 'joins done through trace-id).

The receiving SDK (the server) won't have a replay_id of itself, since it doesn't have a frontend. So we'll just use whatever id we receive in the incoming connection.

src/Sentry/Internal/ReplaySession.cs
```diff
@@ -0,0 +1,33 @@
...
+    public SentryId? ActiveReplayId
+    {
+        get
+        {
+#if __ANDROID__
+            // Check to see if a Replay ID is available
+            var replayId = JavaSdk.ScopesAdapter.Instance?.Options?.ReplayController?.ReplayId?.ToSentryId();
+            return (replayId is { } id && id != SentryId.Empty) ? id : null;
+#else
+            return null;
```
bruno-garcia, line 29: Is the idea to add iOS here later? potentially Web for example if we were brave enough?

bruno-garcia (approved):
approving to unblock, lets see if @vaind has comments then we can fix in follow up PRs

src/Sentry/Internal/ReplaySession.cs
jamescrosswell: Yeah, I was planning for at least iOS...

src/Sentry/DynamicSamplingContext.cs
jamescrosswell: Probably the comment needs rewording. 

We do propagate any upstream replay_id _as long as_ the current process hasn't started it's own replay session. On Android/iOS the chance of that are so remote as not to worry about. If we ever did session replay in a web based app, it might be more likely. 

In any case, if the current process has started it's own replay session then we overwrite any replay_id in the DSC, if there is one... since we don't have any other choice (you can only associate one replay_id with the DSC and you need to use the DSC to communicate the replay_id to Sentry). 

I'll see if I can find a better way to word the comment.

jamescrosswell, line 168:
```suggestion
            // Any upstream replay_id will be propagated only if the current process hasn't started it's own replay session.
            // Otherwise we have to overwrite this as it's the only way to communicate the replayId to Sentry Relay.
            // In Mobile apps this should never be a problem.
```

src/Sentry/DynamicSamplingContext.cs
```diff
@@ -184,18 +210,20 @@ public static DynamicSamplingContext CreateFromPropagationContext(SentryPropagat
...
 
-    public static DynamicSamplingContext CreateDynamicSamplingContext(this TransactionTracer transaction, SentryOptions options)
-        => DynamicSamplingContext.CreateFromTransaction(transaction, options);
+    public static DynamicSamplingContext CreateDynamicSamplingContext(this TransactionTracer transaction, SentryOptions options, IReplaySession? replaySession)
+        => DynamicSamplingContext.CreateFromTransaction(transaction, options, replaySession);
 
-    public static DynamicSamplingContext CreateDynamicSamplingContext(this SentryPropagationContext propagationContext, SentryOptions options)
-        => DynamicSamplingContext.CreateFromPropagationContext(propagationContext, options);
+    public static DynamicSamplingContext CreateDynamicSamplingContext(this SentryPropagationContext propagationContext, SentryOptions options, IReplaySession? replaySession)
+        => DynamicSamplingContext.CreateFromPropagationContext(propagationContext, options, replaySession);
```
vaind, line 228: Instead of adding replay to all the factories, how about we use just the `WithReplayId()`?

vaind: for reference/example, in dart, updating from scope is done like this: https://github.com/getsentry/sentry-dart/blob/e401af3bfe7268f6ad94bbf592fc81525eda264b/dart/lib/src/sentry_baggage.dart#L96-L113

vaind (approved):
Looks reasonable to me. I'm a bit out of context on how replay is implemented in .net so if you want me to review/check this more in-depth, let me know.

src/Sentry/DynamicSamplingContext.cs
jamescrosswell: I'd rather avoid `WithReplayId` where possible as it has to recreate the `DynamicSamplingContext` in order to add the replay id... that would result in us creating the `DynamicSamplingContext` twice for each of the factories if there was an active replay session 😞 

The other option would be to change `DynamicSamplingContext.Items` from a read only dictionary to a mutable dictionary. Most of the stuff in there really shouldn't be messed with once it's created (we're supposed to be propagating stuff verbatim), so until now it's been nice to have this be immutable. 

All of this stuff is internal, so we can easily refactor in a separate PR whenever we like, if we want to go with a different coding style.

PR #4148, opened 2025-04-28
fix: include `Data` set via `ITransactionTracer` in `SentryTransaction`

### LINKED ISSUE #4140: `TransactionTracer.SetData` is not serialized to `contexts.trace.data`
- `TransactionTracer.SetData` is **not** serialized to `contexts.trace.data`
- `SentryTransaction.SetData` **is** serialized to `contexts.trace.data`

### Steps to Reproduce

```csharp
SentrySdk.Init(options =>
{
    options.Dsn = "..";

    options.SetBeforeSendTransaction(transaction =>
    {
        transaction.SetData("setbeforesendtransaction", "transaction-setdata");
        transaction.Contexts.Trace.SetData("setbeforesendtransaction", "transaction-contexts-trace-setdata");
        return transaction;
    });
});

var transaction = SentrySdk.StartTransaction("my-transaction", "start-transaction");
transaction.SetData("transaction-data-key", "transaction-data-value");
transaction.SetData("transaction-array", new[] {42, 43, 44});
// Do some work.
transaction.Finish();
```

### Expected Result

Image

```json
{
  "contexts": {
    "trace": {
      "data": {
        "setbeforesendtransaction": "transaction-contexts-trace-setdata",
        "transaction-array": [42, 43, 44],
        "transaction-data-key": "transaction-data-value"
      }
    }
  }
}
```

### Actual Result

Image

```json
{
  "contexts": {
    "trace": {
      "data": {
        "setbeforesendtransaction": "transaction-contexts-trace-setdata"
      }
    }
  }
}
```

### Additional Notes

- Uncovered through getsentry/sentry-docs#13457

Flash0ver:
fixes #4140

relates to #3936

Changes
- remove backing store of `SentryTransaction.Data`
- reuse `Data` of `SentryContexts.Trace` instead

Added links to discussions from the original PR (if I understood all the context correctly).
Added questions where I am uncertain whether it's the correct way of fixing the issue.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@
 ### Fixes
 
 - Common tags such as `Environment` and `Release` and custom event processors are all now correctly applied to CaptureFeedback events ([#4942](https://github.com/getsentry/sentry-dotnet/pull/4942))
-- Include `Data` set via `ITransactionTracer` in `SentryTransaction`([#4148](https://github.com/getsentry/sentry-dotnet/pull/4148))
+- Include `Data` set via `ITransactionTracer` in `SentryTransaction` ([#4148](https://github.com/getsentry/sentry-dotnet/pull/4148))
 
 ## 6.2.0
 


src/Sentry/TransactionTracer.cs
```diff
@@ -159,15 +159,12 @@ public IReadOnlyList<string> Fingerprint
...
-
     /// <inheritdoc />
     [Obsolete("Use Data")]
-    public IReadOnlyDictionary<string, object?> Extra => _data;
+    public IReadOnlyDictionary<string, object?> Extra => _contexts.Trace.Data;
 
     /// <inheritdoc />
-    public IReadOnlyDictionary<string, object?> Data => _data;
-
+    public IReadOnlyDictionary<string, object?> Data => _contexts.Trace.Data;
```
Flash0ver, line 167: question: reference Contexts in `SentryTransaction` vs copy Contexts to `SentryTransaction`

When creating a `SentryTransaction` from `ITransactionTracer`, we reference the `SentryContexts`:
https://github.com/getsentry/sentry-dotnet/blob/3745efe4720e76fca6a6f16cddc21f813d2ad911/src/Sentry/SentryTransaction.cs#L246

Should we instead copy the Contexts (via `SentryContexts.Clone`?)

Or is that an indicator that this fix here is not quite right, technically?
Should we maybe, rather than have `SetData` writing into `SentryContexts.Trace.Data` directly, instead when creating the `SentryTransaction` from the `ITransactionTracer` copy the `ITransactionTracer.Data` (backed by it's own ``Dictionary`2`` over to the Context of the new `SentryTransaction`.
E.g. something like
```csharp
class SentryTransaction
{
  public SentryTransaction(ITransactionTracer tracer) : this(tracer.Name, tracer.NameSource)
  {
    Contexts = tracer.Contexts;
    foreach (KeyValuePair<string, object?> data in tracer.Data)
    {
      Contexts.Trace.SetData(data.Key, data.Value);
    }
  }
}
```
or similar.
But that would have a side-effect on the `Contexts` of `TransactionTracer`, which I don't think is expected when creating a new `SentryTransaction` from `TransactionTracer`.
Should we perhaps indeed make a copy of the `Trace` and/or `SentryContext`?

Relates to https://github.com/getsentry/sentry-dotnet/pull/3936#discussion_r1946898731

test/Sentry.AspNetCore.Tests/WebIntegrationTests.Versioning.DotNet8_0.verified.txt
```diff
@@ -22,7 +22,11 @@
             Origin: auto.http.aspnetcore,
             Description: ,
             Status: Ok,
-            IsSampled: true
+            IsSampled: true,
+            Data: {
+              http.request.method: GET,
+              http.response.status_code: 200
+            }
```
Flash0ver, line 29: note: from
- https://github.com/getsentry/sentry-dotnet/blob/3745efe4720e76fca6a6f16cddc21f813d2ad911/src/Sentry.AspNetCore/SentryTracingMiddleware.cs#L89
- https://github.com/getsentry/sentry-dotnet/blob/3745efe4720e76fca6a6f16cddc21f813d2ad911/src/Sentry.AspNetCore/SentryTracingMiddleware.cs#L173

test/Sentry.Tests/Protocol/SentryTransactionTests.cs
```diff
@@ -244,6 +244,30 @@ public void SerializeObject_AllPropertiesSetToNonDefault_SerializesValidObject()
 
             return o;
         });
+
+        Assert.Contains($$"""
```
Flash0ver, line 248: note: I kept this test similar to
https://github.com/getsentry/sentry-dotnet/blob/3745efe4720e76fca6a6f16cddc21f813d2ad911/test/Sentry.Tests/Protocol/SentryEventTests.verify.cs#L6
which also does a serialize/deserialize-roundtrip test
plus an additional "Contains-JSON" check on the serialized string

src/Sentry/TransactionTracer.cs
```diff
@@ -278,10 +275,10 @@ internal TransactionTracer(IHub hub, ITransactionContext context, TimeSpan? idle
 
     /// <inheritdoc />
     [Obsolete("Use SetData")]
-    public void SetExtra(string key, object? value) => _data[key] = value;
+    public void SetExtra(string key, object? value) => _contexts.Trace.SetData(key, value);
 
     /// <inheritdoc />
-    public void SetData(string key, object? value) => _data[key] = value;
+    public void SetData(string key, object? value) => _contexts.Trace.SetData(key, value);
```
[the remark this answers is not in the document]
bruno-garcia: `SentryTransaction` shouldn't be accessed concurrently since it represents "a point in time" instance that gets serialized to Sentry.

The TransactionTracer is potentially accessed concurrently (we set it onto the Scope which has concurrent access in both global mode (every `ConfigureScope` call access the same scope instance) or non global mode (since even though there's a single Scope instance per thread, we do access/read things in order to make clones for new threads).

Throwing Concurrent around is not always the solution. In this case for sure not since we want SentryTrasaction to be a snapshot, so we need a *deep clone* here. We don't want just a clone of the collection if the items in the collection are mutable. I wrote a note above about this.

If things on the context are immutable, things are easier though we could have data loss if we have concurrent updates without synchronization.

src/Sentry/TransactionTracer.cs
jamescrosswell: I think at this point the tracer is immutable - we're capturing the transaction trace so the trace should be finished and won't change.

src/Sentry/TransactionTracer.cs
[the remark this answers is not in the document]
jamescrosswell: Hm... I see what you're saying. In the original code the TrancactionTracer stored additional data in `private readonly ConcurrentDictionary<string, object?> _data = new();`

In the new code that data instead gets stored in `_contexts.Trace._data` which is `private Dictionary<string, object?> _data = new();`.

So yes, I think we should probably make the this a concurrent dictionary:
https://github.com/getsentry/sentry-dotnet/blob/4169cb739534e28b64a322a3a3938ee116cf7e6b/src/Sentry/Protocol/Trace.cs#L53

test/Sentry.AspNetCore.Tests/WebIntegrationTests.Versioning.DotNet10_0.verified.txt
jamescrosswell, line 1: 🎉

Issue #4074, opened 2025-04-02
Propagate `Trace ID` to native layer


bitsandfoxes:
## Context

With https://github.com/getsentry/sentry-dotnet/pull/4026 the SDK now accepts `SetTrace` from other "Head SDKs" like the Unity SDK. This builds on top of TwP and allows them to directly set the `PropagationContext` without having to go through the `ContinueTrace`.

With this change we've also extended the `IScopeObserver` to observe changes to the `PropagationContext` and pass then on to the native layer.

## Goals

The goal is to have the .NET SDK pass on the trace ID to the native SDK via the respective ScopeObserver.

### Native ✅ 
Platforms that are covered by `sentry-native` got this functionality with #4026 and the observer observes.
https://github.com/getsentry/sentry-dotnet/blob/438ae83fd12beee596d64be3b5876eabb81c94e0/src/Sentry/Platforms/Native/NativeScopeObserver.cs#L43-L44

### Android
Supporting this on Android is currently blocked by https://github.com/getsentry/sentry-dotnet/pull/3911. The feature requires at least `8.4.0`.
https://github.com/getsentry/sentry-dotnet/blob/438ae83fd12beee596d64be3b5876eabb81c94e0/src/Sentry/Platforms/Android/AndroidScopeObserver.cs#L103-L106

### Cocoa
The Cocoa SDK is missing the functionality to have the trace ID passed. Once that's done we'll need to update the CocoaScopeObserver 
https://github.com/getsentry/sentry-dotnet/blob/438ae83fd12beee596d64be3b5876eabb81c94e0/src/Sentry/Platforms/Cocoa/CocoaScopeObserver.cs#L111-L114

jamescrosswell:
@bitsandfoxes it looks like it would currently only be possible to implement this for sentry-native.
- Is that correct?
- Is that sufficient or do we need it on all three platforms?
  - If so, are there tracking issues in the other repos for this yet that we can add as related/blocking issues?

PR #4153, opened 2025-04-29
feat: Observe transaction on scope

bitsandfoxes:
### Problem Statement

Underlying SDKs should be able to observe changes to the transaction, so they use the same trace context.

### Proposal

1. Leverage the existing `SetTrace` and call it whenever a transaction gets set/unset on the scope.
2. When the transaction finishes, the trace context needs to be updated again. Since there is no active transaction, this falls back to the propagation context. To preserve some sort of "time continuity", and not have new events contain a trace ID that is "older" than the transaction that just finished, the propagation context gets regenerated.

### DIFF
--- a/src/Sentry/Scope.cs
+++ b/src/Sentry/Scope.cs
@@ -225,6 +225,20 @@ public class Scope : IEventLike
             try
             {
                 _transaction.Value = value;
+
+                if (Options.EnableScopeSync)
+                {
+                    if (_transaction.Value != null)
+                    {
+                        // If there is a transaction set we propagate the trace to the native layer
+                        Options.ScopeObserver?.SetTrace(_transaction.Value.TraceId, _transaction.Value.SpanId);
+                    }
+                    else
+                    {
+                        // If the transaction is being removed from the scope, reset and sync the trace as well
+                        Options.ScopeObserver?.SetTrace(PropagationContext.TraceId, PropagationContext.SpanId);
+                    }
+                }
             }
             finally
             {
@@ -802,6 +816,11 @@ public class Scope : IEventLike
             if (ReferenceEquals(_transaction.Value, expectedCurrentTransaction))
             {
                 _transaction.Value = null;
+                if (Options.EnableScopeSync)
+                {
+                    // We have to restore the trace on the native layers to be in sync with the current scope
+                    Options.ScopeObserver?.SetTrace(PropagationContext.TraceId, PropagationContext.SpanId);
+                }
             }
         }
         finally
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
+
+## Features
+
+- When setting a transaction on the scope, the SDK will attempt to sync the transaction's trace with the SDK on the native layer ([#4153](https://github.com/getsentry/sentry-dotnet/pull/4153)) 
+
 ## 5.6.0
 
 ### Features
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Changelog
 
-## Unreleased
+## 5.6.0
 
 ### Features
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ## Features
 
-- When setting a transaction on the scope, the SDK will attempt to sync the transaction's trace with the SDK on the native layer ([#4153](https://github.com/getsentry/sentry-dotnet/pull/4153)) 
+- When setting a transaction on the scope, the SDK will attempt to sync the transaction's trace context with the SDK on the native layer. Finishing a transaction will now also start a new trace ([#4153](https://github.com/getsentry/sentry-dotnet/pull/4153)) 
 - Added `CaptureFeedback` overload with `configureScope` parameter ([#4073](https://github.com/getsentry/sentry-dotnet/pull/4073))
 - Custom SessionReplay masks in MAUI Android apps ([#4121](https://github.com/getsentry/sentry-dotnet/pull/4121))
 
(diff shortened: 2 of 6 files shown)


src/Sentry/Scope.cs
```diff
@@ -225,6 +225,20 @@ public ITransactionTracer? Transaction
             try
             {
                 _transaction.Value = value;
+
+                if (Options.EnableScopeSync)
+                {
+                    if (_transaction.Value != null)
+                    {
+                        // If there is a transaction set we propagate the trace to the native layer
+                        Options.ScopeObserver?.SetTrace(_transaction.Value.TraceId, _transaction.Value.SpanId);
```
bruno-garcia, line 234: I'm concerned about calling this callback under a lock. We could create contention if this callbackcan be slow (which if it's P/Invoke it can be relatively speaking) 

But I imagine we don't have an issue concurrent access to `Transaction` right? it's at most 1 auto instrumentation and then a competing access by a user doing their manual transaction stuff?

src/Sentry/Scope.cs
```diff
@@ -225,6 +225,20 @@ public ITransactionTracer? Transaction
...
+                    }
+                    else
+                    {
+                        // If the transaction is being removed from the scope, reset and sync the trace as well
+                        Options.ScopeObserver?.SetTrace(PropagationContext.TraceId, PropagationContext.SpanId);
```
bruno-garcia, line 239: hm here we go from the transaction trace-id "back" to whatever we had before?

So could be:

TwP Trace-1 .. A (telemetry sent), B, C
- Stop, transaction happening:
Transaction: Trace-2 .. A
- OK Transaction over, back to trace-1
TwP Trace-1 D .. E 

I find this odd. 
Some ideas:
1. The transaction could stay on the same trace that's existent. Leaving the creating of trace to be done by the other system that exists for "trace without performance"
2. Once the transaction ends, we continue with that same trace id. But we override it if a new trace id is set (by the TwP mechanism, such as scene load or whatnot). 

So if we do `n2` it'd be:

App starts, TwP Trace-1: A, B ...
Transaction starts, new trace created: Trace-2: A, B, C...
Transaction ends. 
Trace-2: D, E, F (data will be tied to the previous transaction by trace id, but they are different entitties
Some new trace is set (say Release hooks or something) or new Transaction happens: New trace id:
Trace-3: A, B, C .. etc

Does this make sense? (if not, it's because it's late :)

bruno-garcia: Another alternative: Trace Links.

When you create the transaction, can that trace id by linked to the outer, "Trace Context" trace?
Doesn't make sense for the two things to be disconnected just because we had a transaction with a new trace-id for some time, and stuff that hapened before and after be on the same trace (but not connected to the transaction)

bitsandfoxes: > hm here we go from the transaction trace-id "back" to whatever we had before?

The `ScopeObserver` attempts to keep the scope in sync between all layers. After the transaction finishes, all future events fall back and contain the `PropagationContext`'s trace ID. Events from the native layer should use the same ID.

The `TransactionTracer` is the one that clears the transaction off the scope from
https://github.com/getsentry/sentry-dotnet/blob/1bf37620fe5955defb7eca8aa1218e3a969d9413/src/Sentry/TransactionTracer.cs#L395-L396
to
https://github.com/getsentry/sentry-dotnet/blob/1bf37620fe5955defb7eca8aa1218e3a969d9413/src/Sentry/Scope.cs#L797

How about the `TransactionTracer` also updates the `PropagationContext` and "starts" a new trace when it finishes? That'd create some continuity and be in line with what trace ID generating integrations (i.e. in Unity) do. Since there was a transaction, this is also not even related to TwP anymore.

src/Sentry/Scope.cs
bitsandfoxes: > But I imagine we don't have an issue concurrent access to Transaction right? it's at most 1 auto instrumentation and then a competing access by a user doing their manual transaction stuff?

Correct, that is the current worst-case scenario.

jamescrosswell, line 234: Is it just the TraceId and SpanId that need to be synced? Should the sample rate and sample rand also be synced (we normally put that on the DSC to ensure consistent sampling decisions)? Or will this just be used to create spans (not transactions) and so no sampling decisions need to be made?

src/Sentry/Scope.cs
```diff
@@ -802,6 +816,11 @@ internal void ResetTransaction(ITransactionTracer? expectedCurrentTransaction)
             if (ReferenceEquals(_transaction.Value, expectedCurrentTransaction))
             {
                 _transaction.Value = null;
+                if (Options.EnableScopeSync)
```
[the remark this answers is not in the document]
jamescrosswell: Hm, I wonder if we need the locking at all. It's an `AsyncLocal<ITransactionTracer?>` so in what scenarios would we see contention for reading/writing this? It has to be code from the same AsyncLocal context, which implicitly means it's on the same thread right? We shouldn't ever see multiple threads trying to access this concurrently?

@bruno-garcia sanity check?

src/Sentry/Scope.cs
bitsandfoxes: In the future, they should. But both the iOS and Android ScopeObservers are missing their implementation. See https://github.com/getsentry/sentry-dotnet/issues/4074

It needs bumping the native SDKs to specific versions and then mess around with the bindings (especially, since the API is in private/internal parts of the Cocoa/Java SDK)

src/Sentry/Scope.cs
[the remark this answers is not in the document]
bruno-garcia: if it's always AsyncLocal we don't need any locking

bruno-garcia (approved):

src/Sentry/Scope.cs
bruno-garcia: Please raise a ticket in the relevant repos so we don't forget about this

Issue #4132, opened 2025-04-22
Sentry Structured Logging for .NET


AbhiPrasad:
https://develop.sentry.dev/sdk/telemetry/logs/

Sentry is adding support for [structured logging](https://github.com/getsentry/sentry/discussions/86804). Let's add it to the .NET SDK!

1. Define the [logs protocol](https://develop.sentry.dev/sdk/telemetry/logs/#logs-protocol) and log envelope item in the SDK
1. Add the Public API (SDK options and methods) as per the [docs](https://develop.sentry.dev/sdk/telemetry/logs/#public-api)
1. Make sure the the SDK follows the [documented behavior](https://develop.sentry.dev/sdk/telemetry/logs/#sdk-behavior)
1. Attach [default attributes](https://develop.sentry.dev/sdk/telemetry/logs/#default-attributes) to the SDK as per docs
1. Instrument popular logging libraries to send logs to Sentry. SDK maintainers can best decide what they should support, but we should definitely support the same integrations that we do in the SDK (`NLog`, `Serilog`, etc.)
1. Create a GH discussion that contains instructions for setting up the SDK in your repo. Eventually this will be moved into the primary docs. See the [JS SDK's GH discussion](https://github.com/getsentry/sentry-javascript/discussions/15916) for inspiration.

bruno-garcia:
The idea is to:

1. Ship the barebone Sentry API into the core `Sentry` package
2. Make this work from `Microsoft.Extensions.Logging`
3. Make this work from `Serilog`
4. Make this work from `NLog`
5. Make this work from `log4net

The order above allows us to try/test things incrementally.
I'll date to say that no one in .NET will use "Sentry's logging API". I know I wouldn't. But that's not the point.
The point is to have the API in the main SDK, so that the logging libraries integrations only pipe data to it.


All our current logging integrations do:
1. Check level, if higher than `MininumEventLevel` it records an event
2. Check level, if higher than `MinimumBreadcrumbLevel` it adds a breadcrumb

e.g:

https://github.com/getsentry/sentry-dotnet/blob/084e3b4bf2c043b42ad6b11892a604e027b3d1a5/src/Sentry.Extensions.Logging/SentryLogger.cs#L39-L86

The crumb goes after, so that in a subsequent event, it will include the breadcrumb of the previous event.

This works well for error monitoring. But with the new logging integration, we'll send all logs to Sentry.
[...]

PR #4158, opened 2025-04-30
WIP: Sentry Logs

Flash0ver:
Draft PR: Work in progress

Early feedback appreciated on
- public APIs

see also Question-Comments in the draft


src/Sentry/Experimental/SentrySeverity.cs
```diff
@@ -0,0 +1,74 @@
+using Sentry.Infrastructure;
+
+#pragma warning disable CS1591 // Missing XML comment for publicly visible type or member
+
+namespace Sentry.Experimental;
+
+//TODO: QUESTION: not sure about the name
+// this is a bit different to Sentry.SentryLevel and Sentry.BreadcrumbLevel
```
bruno-garcia, line 8: Does it need to be different? I wonder if we can use SentryLevel instead?
Because we don't have `Trace` there? :~

We could add it as `-1` 😅

src/Sentry/Experimental/System.Diagnostics.CodeAnalysis.ExperimentalAttribute.cs
```diff
@@ -0,0 +1,29 @@
...
+                AttributeTargets.Struct |
+                AttributeTargets.Enum |
+                AttributeTargets.Constructor |
+                AttributeTargets.Method |
+                AttributeTargets.Property |
+                AttributeTargets.Field |
+                AttributeTargets.Event |
+                AttributeTargets.Interface |
+                AttributeTargets.Delegate, Inherited = false)]
+internal sealed class ExperimentalAttribute : Attribute
```
bruno-garcia, line 18: @SimonCropp has a shimming library that might have this, no?

src/Sentry/SentryLogger.cs
```diff
@@ -0,0 +1,68 @@
...
+
+/// <summary>
+/// Creates and sends logs to Sentry.
+/// </summary>
+[Experimental(DiagnosticId.ExperimentalSentryLogs, UrlFormat = UrlFormats.ExperimentalSentryLogs)]
+public sealed class SentryLogger
+{
+    //TODO: QUESTION: Trace vs LogTrace
+    // Trace() is from the Sentry Logs feature specs. LogTrace() would be more .NET idiomatic
+    public void Trace(string template, object[]? parameters = null, Action<SentryLog>? configureLog = null)
```
bruno-garcia, line 18: Hm, could we have overloads with `T` instead to avoid boxing cost when logging is disabled?

That's the approach taken on the diagnostic logger extensions

src/Sentry/SentryLogger.cs
```diff
@@ -0,0 +1,68 @@
...
+
+    //TODO: consider ReadOnlySpan for TFMs where Span is available
+    //  or: utilize a custom [InterpolatedStringHandler] for modern TFMs
+    //      with which we may not be able to enforce on compile-time to only support string, boolean, integer, double
+    //      but we could have an Analyzer for that, indicating that Sentry does not support other types if used in the interpolated string
+    //  or: utilize a SourceGen, similar to the Microsoft.Extensions.Logging [LoggerMessage]
+    //      with which we could enforce on compile-time to only support string, boolean, integer, double
+    private static void CaptureLog(SentrySeverity level, string template, object[]? parameters, Action<SentryLog>? configureLog)
+    {
+        string message = String.Format(template, parameters ?? []);
```
bruno-garcia, line 61: prefer reserved keyword instead of class name:

```suggestion
        string message = string.Format(template, parameters ?? []);
```

src/Sentry/SentryLogger.cs
```diff
@@ -0,0 +1,68 @@
...
+        CaptureLog(SentrySeverity.Fatal, template, parameters, configureLog);
+    }
+
+    //TODO: consider ReadOnlySpan for TFMs where Span is available
+    //  or: utilize a custom [InterpolatedStringHandler] for modern TFMs
+    //      with which we may not be able to enforce on compile-time to only support string, boolean, integer, double
+    //      but we could have an Analyzer for that, indicating that Sentry does not support other types if used in the interpolated string
+    //  or: utilize a SourceGen, similar to the Microsoft.Extensions.Logging [LoggerMessage]
+    //      with which we could enforce on compile-time to only support string, boolean, integer, double
+    private static void CaptureLog(SentrySeverity level, string template, object[]? parameters, Action<SentryLog>? configureLog)
```
bruno-garcia, line 59: Are we going to do batching here?

jamescrosswell:
@Flash0ver can we reference some developer docs and/or an issue in the PR description to help reviewers get an overview of the change before looking at code?

src/Sentry/Infrastructure/DiagnosticId.cs
```diff
@@ -8,4 +8,13 @@ internal static class DiagnosticId
     /// </summary>
     internal const string ExperimentalFeature = "SENTRY0001";
 #endif
+
+    //TODO: QUESTION: Should we re-use the above for all [Experimental] features or have one ID per experimental feature?
+    internal const string ExperimentalSentryLogs = "SENTRY0002";
```
jamescrosswell, line 13: I think it's OK to just have one... the main purpose here is to make sure people are aware/deliberate about opting into our experimental APIs. They'll get a separate warning in their code each time they use one/any of them. 

Potentially they could be disabling those in a directory.build.props file for their solution... in which case separate diagnostic ids would mean they'd need to disable warnings for our Heap Dumps feature and the Logs feature separately... seems like overkill.

src/Sentry/Internal/Extensions/JsonExtensions.cs
```diff
@@ -472,6 +472,17 @@ public static void WriteSerializable(
         writer.WriteSerializableValue(value, logger);
     }
 
+    public static void WriteSerializable<TValue>(
```
jamescrosswell, line 475: How is this different to the overload defined on line 465? The only difference I can see is the struct constraint in the where clause (but is this important or used for anything)?

src/Sentry/SentryLogger.cs
```diff
@@ -0,0 +1,68 @@
...
+public sealed class SentryLogger
+{
+    //TODO: QUESTION: Trace vs LogTrace
+    // Trace() is from the Sentry Logs feature specs. LogTrace() would be more .NET idiomatic
+    public void Trace(string template, object[]? parameters = null, Action<SentryLog>? configureLog = null)
+    {
+        CaptureLog(SentrySeverity.Trace, template, parameters, configureLog);
+    }
+
+    //TODO: QUESTION: parameter name "template" vs "format"
```
jamescrosswell, line 23: I feel like it's less mental gymnastics for us if we go with the sentry naming conventions (when we're reading code from other eepos etc. to see how they make changes that we need to mimic). So I'd probably go with `template` personally.

src/Sentry/SentryLogger.cs
jamescrosswell: `CaptureLogs` as well? 

The spec allows for multiple envelope items... which would be more efficient in apps that are emitting loads of logs. We'd need to build some fancy logic to queue/flush logs in that case. Probably worth it from a performance point of view though since logs are the kind of thing people might send oodles of.

src/Sentry/SentryLogger.cs
jamescrosswell: @Flash0ver do you get squiggly lines for that stuff in Rider? The analysers should spot that stuff automatically.

test/Sentry.DiagnosticSource.IntegrationTests/SqlListenerTests.verify.cs
```diff
@@ -155,7 +155,7 @@ public void ShouldIgnoreAllErrorAndExceptionIds()
         foreach (var field in eventIds)
         {
             var eventId = (EventId)field.GetValue(null)!;
-            var isEfExceptionMessage = SentryLogger.IsEfExceptionMessage(eventId);
+            var isEfExceptionMessage = Sentry.Extensions.Logging.SentryLogger.IsEfExceptionMessage(eventId);
```
jamescrosswell, line 158: Hm, we've got two types called `SentryLogger` now? `Sentry.SentryLogger` and also `Sentry.Extensions.Logging.SentryLogger`? I can see that being a pain over time. I reckon we rename one of them. 

`Sentry.Extensions.Logging.SentryLogger` is `internal sealed` so no issues renaming that if we have to. `SentryMelLogger` maybe?

More painful would be things like [SentryLoggingOptions](https://github.com/getsentry/sentry-dotnet/blob/549bcd653c94b72e59e791a4503849188b2a48ca/src/Sentry.Extensions.Logging/SentryLoggingOptions.cs), since that's public. Definitely want to plan in advance to minimise ambiguity over naming once all of this is completed though.

jamescrosswell:
Looking good so far 👍🏻

src/Sentry/Internal/Extensions/JsonExtensions.cs
Flash0ver: The overload above
https://github.com/getsentry/sentry-dotnet/blob/8978bc4bf2b8c0de53d5cbedd63de176adfacb3d/src/Sentry/Internal/Extensions/JsonExtensions.cs#L465-L469
takes an `ISentryJsonSerializable`, and is boxing struct-based `ISentryJsonSerializable` to the interface type.

My original motivation was to add a new overload that avoids boxing for struct-based `ISentryJsonSerializable` to not binary-break existing call sites, but just realize that we have 11 existing call sites binding to the new overload.

IIRC we're not too concerned about binary compatibility, but definitely about source compat, since user's usually only consume one Sentry-package, or at least keep the Sentry-Package-Versions in sync, so I'm thinking about changing the existing `WriteSerializable` to be generic.

Also, I just realize that the only usages of the method is within the `Sentry` assembly, so there are not even binary breaking concerns at all.

src/Sentry/Infrastructure/DiagnosticId.cs
Flash0ver: SGTM
I am thinking about adding a single `UrlFormat` string, used by both/all experimental features, linking to a dedicated "Experimental Features" page in our _docs_. But will follow up with that in a follow-up PR.

src/Sentry/SentryLogger.cs
Flash0ver: Yes, batching is amongst my TODOs ... I updated the description with open work during the _draft_ status.

src/Sentry/SentryLogger.cs
Flash0ver: Yes, I do get ... it was an oversight on my end. Old habits.
Also, I need to start using `var` over explicit typing.

src/Sentry/SentryLogger.cs
Flash0ver: Yes, it's also on my bucket list ... I updated the description to reflect required work, too.
I'd not only like to add up to 3 `TArg` overloads per severity, but also - in a follow-up PR - add a custom `[InterpolatedStringHandler]`, which I'm going to explore over the weekend.

src/Sentry/Experimental/System.Diagnostics.CodeAnalysis.ExperimentalAttribute.cs
Flash0ver: It is indeed!
I only checked the `README` where I didn't find it, but since `v1.32.0` it is included (I'll PR `Polyfill`'s docs later).

Now I got two questions:
- Should we update to `v1.32.0`, or the latest version of `Polyfill` (`v7.27.0`).
- Depending on the version we chose: should we update this in a separate PR?
  - e.g. `v1.34.0` adds the `System.Runtime.CompilerServices.DefaultInterpolatedStringHandler` which the compiler then chooses for _interpolated strings_
    - this should have a positive effect on performance when using _interpolated strings_ in our codebase targeting TFMs less than `net6.0`

src/Sentry/Experimental/SentrySeverity.cs
Flash0ver: No, not necessarily.
And adding it as `-1` would not break any existing behavior - but may warrant some documentation - like:
https://github.com/getsentry/sentry-dotnet/blob/8978bc4bf2b8c0de53d5cbedd63de176adfacb3d/src/Sentry.Extensions.Logging/LogLevelExtensions.cs#L21

---

But reusing `Sentry.SentryLevel` could be confusing:
https://github.com/getsentry/sentry-dotnet/blob/8978bc4bf2b8c0de53d5cbedd63de176adfacb3d/src/Sentry/SentryLevel.cs#L6
uses the `EnumMemberAttribute`, with e.g. `[EnumMember(Value = "warning")] Warning,`. But the OTEL `SeverityText` field and our Protocol expects `"warn"` instead. Well, we would still have the separate `ToSentryProtocolLevel` (or similar) extension method that does the right thing, but the re-use may cause confusion.

---

Also, this would have a side effect onto existing usages, via`SentryOptions.DiagnosticLevel`:
https://github.com/getsentry/sentry-dotnet/blob/8978bc4bf2b8c0de53d5cbedd63de176adfacb3d/src/Sentry/Infrastructure/DiagnosticLogger.cs#L8
Do we want to extend the "Log-Level" for existing internal/diagnostic logging?

---

On the other hand, adding yet another "Log-Level" enum, that is similar, but not quite the same, is similarly confusing.

---

And now I am thinking if we could get away with making this new enum type `internal` instead, since we do have separate `Log{Level}()` methods, and application code could still change the severity via the `int`-based `SeverityNumber` through the `BeforeSendLog`-callback. (see [Log Severity Number](https://develop.sentry.dev/sdk/telemetry/logs/#log-severity-number))
To, for now, reduce the public surface area, and we could think about which approach to chose when/if the issue arises later.

test/Sentry.DiagnosticSource.IntegrationTests/SqlListenerTests.verify.cs
bruno-garcia: Are all of these public? `Sentry.Extensions.Logging.SentryLogger` 

https://github.com/getsentry/sentry-dotnet/blob/f765f991da77ecffd8813b772ddba55acc27d473/src/Sentry.Extensions.Logging/SentryLogger.cs#L6

This is internal so less of a concern

src/Sentry/Experimental/System.Diagnostics.CodeAnalysis.ExperimentalAttribute.cs
Flash0ver: see #4167

Flash0ver: done

test/Sentry.DiagnosticSource.IntegrationTests/SqlListenerTests.verify.cs
Flash0ver: We "leak" the internal `SentryLogger` name through the public `Sentry.Extensions.Logging.SentryLoggerProvider` and `Microsoft.Extensions.Logging.SentryLoggerFactoryExtensions` types. Well ... not really "leaking", but to some degree "implying".
So I would like to change the name of the new Logger.

Although I do like `SentryStructuredLogger`, ``M.E.L.ILogger`1`` also supports structured logging.

I go with `SentrySdkLogger` for now, definitely embracing the `[Experimental]`-Attribute, potentially changing it later when integrating it into `Sentry.Extensions.Logging` and using it from both `Sentry.AspNetCore` and `Sentry.Maui`, which I hope then makes the right name more obvious.

jamescrosswell: I think `SentryStructuredLogger` is fine - it's our implementation of structured logging and it's not going to conflict with any other packages.

samples/Sentry.Samples.Console.Basic/Program.cs
```diff
@@ -9,6 +9,7 @@
  */
 
 using System.Net.Http;
+using Sentry.Protocol;
```
bruno-garcia, line 12: Lets move the logger log level to the root namaespace to avoid having to import this?

samples/Sentry.Samples.Console.Basic/Program.cs
```diff
@@ -33,8 +34,37 @@
...
+{
+    log.SetAttribute("string-attribute", "value");
+    log.SetAttribute("boolean-attribute", true);
+    log.SetAttribute("integer-attribute", long.MaxValue);
+    log.SetAttribute("double-attribute", double.MaxValue);
+};
+
+SentrySdk.Logger.LogTrace("Hello, World!", null, configureLog);
+SentrySdk.Logger.LogDebug("Hello, .NET!", null, configureLog);
+SentrySdk.Logger.LogInfo("Information", null, configureLog);
```
bruno-garcia, line 60: the `null` in the public API isn't ideal, lets see if we can do something about that

samples/Sentry.Samples.Console.Basic/Program.cs
```diff
@@ -33,8 +34,37 @@
...
+SentrySdk.Logger.LogWarning("Warning with one {0}", ["parameter"], configureLog);
+SentrySdk.Logger.LogError("Error with {0} {1}", [2, "parameters"], configureLog);
+SentrySdk.Logger.LogFatal("Fatal {0} and {1}", [true, false], configureLog);
+
+await Task.Delay(TimeSpan.FromSeconds(5));
```
bruno-garcia, line 65: Do we need this wait?
If u want to flush, there's:

```suggestion
await SentrySdk.FlushAsync(TimeSpan.FromSeconds(5));
```

samples/Sentry.Samples.Console.Basic/Program.cs
```diff
@@ -33,8 +34,37 @@
...
+
+/*
```
bruno-garcia, line 67: this by design?

src/Sentry/Extensibility/DisabledHub.cs
```diff
@@ -21,6 +22,7 @@ public class DisabledHub : IHub, IDisposable
 
     private DisabledHub()
     {
+        Logger = new SentryStructuredLogger(this);
```
bruno-garcia, line 25: Should this be a `NoOp` instance?

Since you don't pass `options` and all public method have a gate behind IsEnabled, it's probably fine

src/Sentry/Protocol/Envelopes/Envelope.cs
```diff
@@ -445,6 +445,21 @@ internal static Envelope FromClientReport(ClientReport clientReport)
         return new Envelope(header, items);
     }
 
+    [Experimental(DiagnosticId.ExperimentalFeature)]
```
bruno-garcia, line 448:
```suggestion
    // TODO: This is temporary. We don't expect single log messages to become an envelope by themselves since batching is needed
    [Experimental(DiagnosticId.ExperimentalFeature)]
```

src/Sentry/Protocol/SentryLog.cs
```diff
@@ -0,0 +1,280 @@
...
+    private readonly SentryLogLevel _level;
+
+    [SetsRequiredMembers]
+    internal SentryLog(DateTimeOffset timestamp, SentryId traceId, SentryLogLevel level, string message)
+    {
+        Timestamp = timestamp;
+        TraceId = traceId;
+        Level = level;
+        Message = message;
+        _attributes = new Dictionary<string, SentryAttribute>(7);
```
bruno-garcia, line 24:
```suggestion
        // 7 is the number of built-in attributes, so we start with that.
        _attributes = new Dictionary<string, SentryAttribute>(7);
```

src/Sentry/Protocol/SentryLog.cs
```diff
@@ -0,0 +1,280 @@
...
+    /// The severity level of the log.
+    /// <para>This API is experimental and it may change in the future.</para>
+    /// </summary>
+    [Experimental(DiagnosticId.ExperimentalFeature)]
+    public required SentryLogLevel Level
+    {
+        get => _level;
+        init
+        {
+            SentryLogLevelExtensions.ThrowIfOutOfRange(value);
```
bruno-garcia, line 54: Ideally we don't throw from public APIs (and also properties ideally shouldnt' throw either, they 'look like fields').
Possibly better approach here is to debug log if diagnostic logging is on, and default to some level like `debug`.

src/Sentry/Protocol/SentryLog.cs
```diff
@@ -0,0 +1,280 @@
...
+    /// </summary>
+    [Experimental(DiagnosticId.ExperimentalFeature)]
+    public void SetAttribute(string key, double value)
+    {
+        _attributes[key] = new SentryAttribute(value, "double");
+    }
+
+    internal void SetAttributes(SentryOptions options)
+    {
+        var environment = options.SettingLocator.GetEnvironment();
```
bruno-garcia, line 213: I wonder wondering if we're reading env vars on each call, but value is cached:

https://github.com/getsentry/sentry-dotnet/blob/bd026c4aea0554fdacd114b91c874638846ceec7/src/Sentry/Internal/SettingLocator.cs#L69-L88

so we're good ;)

src/Sentry/Protocol/SentryLog.cs
```diff
@@ -0,0 +1,280 @@
...
+        {
+            SetAttribute("sentry.release", release);
+        }
+
+        if (ParentSpanId.HasValue)
+        {
+            SetAttribute("sentry.trace.parent_span_id", ParentSpanId.Value.ToString());
+        }
+
+        SetAttribute("sentry.sdk.name", Constants.SdkName);
```
bruno-garcia, line 227: what would we report on `Unity`? 

We have ways to let outer SDKs to override the name, we need that somehow to reflect here.

We might want to improve how we take on the SDK name, today I think it's at event creation time we set the value, then during event building/processing in the SDK we don't override it. And in EventProcessor we set the value if it's null.

Perhaps we want some `internal` property on options that let the value be changed by outer SDKs? not sure, worth checking with @vaind how he did this on Unity/Android/etc

src/Sentry/SentryStructuredLogger.cs
```diff
@@ -0,0 +1,212 @@
...
+    {
+        if (IsEnabled)
+        {
+            CaptureLog(SentryLogLevel.Fatal, template, parameters, configureLog);
+        }
+    }
+
+    private void CaptureLog(SentryLogLevel level, string template, object[]? parameters, Action<SentryLog>? configureLog)
+    {
+        Debug.Assert(_options is not null);
```
bruno-garcia, line 119: why not IsEnabled check here? and by pass all the public method checks

src/Sentry/SentryStructuredLogger.cs
```diff
@@ -0,0 +1,212 @@
...
+        try
+        {
+            configureLog?.Invoke(log);
+        }
+        catch (Exception e)
+        {
+            //TODO: change to Diagnostic Logger (if enabled)
+            // see https://github.com/getsentry/sentry-dotnet/issues/4132
+            Console.WriteLine(e);
+            return;
```
bruno-garcia, line 148: this is prob the safer approach, which ideally is documented under the configureLog xml doc and sentry docs

samples/Sentry.Samples.Console.Basic/Program.cs
Flash0ver: I reworked the sample a bit to be less verbose.

src/Sentry/Extensibility/DisabledHub.cs
Flash0ver: I made it more explicit that we are creating a "disabled" logger here, and added a test confirming that this logger does not capture logs, even when `Options.EnableLogs = true`.

src/Sentry/Protocol/SentryLog.cs
Flash0ver: removed all related throw helpers and added tests to ensure fallback behavior when out of range

src/Sentry/SentryStructuredLogger.cs
Flash0ver: I wanted to avoid the `CaptureLog` method invocation if logging is not enabled.
This assert is to let null-flow analysis know that options are not null here, because we already checked in the calling method.

src/Sentry/SentryStructuredLogger.cs
Flash0ver: Added XML doc comments to all `Log{Level}` messages and `SentryOptions.SetBeforeSendLog`, indicating that the log will not be captured when an `Exception` occurs during callback invocation.
Also changed to log an `Error` via the `IDiagnosticLogger` (if enabled).

src/Sentry/Experimental/SentrySeverity.cs
Flash0ver: We currently still have a new `enum`.
Renamed to `SentryLogLevel`. Matches the `sentry-java` SDK.
Also moved to `Sentry` namespace.

src/Sentry/SentryLogger.cs
Flash0ver: Will do in follow-up: https://github.com/getsentry/sentry-dotnet/issues/4132#issuecomment-2883482296

test/Sentry.DiagnosticSource.IntegrationTests/SqlListenerTests.verify.cs
Flash0ver: renamed to `SentryStructuredLogger`
and moved to namespace `Sentry`

samples/Sentry.Samples.Console.Basic/Program.cs
Flash0ver: moved both new types, `SentryLog` and `SentryLogLevel`

samples/Sentry.Samples.Console.Basic/Program.cs
Flash0ver: Will add appropriate overloads in a follow-up, to not bloat this one too much: https://github.com/getsentry/sentry-dotnet/issues/4132#issuecomment-2883482296

src/Sentry/Protocol/SentryLog.cs
Flash0ver: Will fix in immediate follow-up when integrating via `Sentry.Extensions.Logging`
See https://github.com/getsentry/sentry-dotnet/issues/4132#issuecomment-2883482296

src/Sentry/Extensibility/DiagnosticLoggerExtensions.cs
jamescrosswell, line 1: It's almost worth writing a source generator to maintain this file... so much boilerplate. Not for this PR obviously.

src/Sentry/SentryStructuredLogger.cs
```diff
@@ -0,0 +1,237 @@
...
+    /// Creates and sends a structured log to Sentry, with severity <see cref="SentryLogLevel.Trace"/>, when enabled and sampled.
+    /// <para>This API is experimental and it may change in the future.</para>
+    /// </summary>
+    /// <param name="template">A formattable <see langword="string"/>. When incompatible with the <paramref name="parameters"/>, the log will not be captured. See <see href="https://learn.microsoft.com/dotnet/api/system.string.format">System.String.Format</see>.</param>
+    /// <param name="parameters">The arguments to the <paramref name="template"/>. See <see href="https://learn.microsoft.com/dotnet/api/system.string.format">System.String.Format</see>.</param>
+    /// <param name="configureLog">A delegate to set attributes on the <see cref="SentryLog"/>. When the delegate throws an <see cref="Exception"/> during invocation, the log will not be captured.</param>
+    [Experimental(DiagnosticId.ExperimentalFeature)]
+    public void LogTrace(string template, object[]? parameters = null, Action<SentryLog>? configureLog = null)
+    {
+        if (IsEnabled)
```
jamescrosswell, line 49: It looks like all of the public methods have this `IsEnabled` check and then delegate all the work to the `CaptureLog` method.

We could centralise this logic by putting the check for `IsEnabled` at the top of the `CaptureLog` method instead.

src/Sentry/IHub.cs
```diff
@@ -17,6 +17,20 @@ public interface IHub : ISentryClient, ISentryScopeManager
...
+    /// </summary>
+    /// <remarks>
+    /// Available options:
+    /// <list type="bullet">
+    /// <item><see cref="Sentry.SentryOptions.EnableLogs"/></item>
+    /// <item><see cref="Sentry.SentryOptions.SetBeforeSendLog(System.Func{SentryLog, SentryLog})"/></item>
+    /// </list>
+    /// </remarks>
+    [Experimental(Infrastructure.DiagnosticId.ExperimentalFeature)]
+    public SentryStructuredLogger Logger { get; }
```
jamescrosswell, line 32: Just thinking about testing of the Hub classes in general, the `Logger` is kind of a dependency of the hub and currently it's not very easy to inject/mock.

Maybe worth creating an `ISentryStructuredLogger` interface and having this member be of type `ISentryStructuredLogger` rather than `SentryStructuredLogger`. A mock Logger could then optionally be injected via the constructor for test purposes.

src/Sentry/SentryLog.cs
```diff
@@ -0,0 +1,281 @@
+using Sentry.Extensibility;
+using Sentry.Infrastructure;
+using Sentry.Internal;
+using Sentry.Protocol;
+
+namespace Sentry;
+
+/// <summary>
+/// Represents the Sentry Log protocol.
```
jamescrosswell, line 9: Is there any reason this isn't in the `Sentry.Protocol` namespace?

src/Sentry/SentryLog.cs
```diff
@@ -0,0 +1,281 @@
...
+    }
+
+    /// <summary>
+    /// The timestamp of the log.
+    /// <para>This API is experimental and it may change in the future.</para>
+    /// </summary>
+    /// <remarks>
+    /// Sent as seconds since the Unix epoch.
+    /// </remarks>
+    [Experimental(DiagnosticId.ExperimentalFeature)]
```
jamescrosswell, line 35: Possibly overkill decorating every property with this... the class is already decorated with `Experimental`.

I think it's enough if we put the option to enable this feature under `SentryOptions.Experimenta` and then decorate any public methods in the API (the stuff on `IHub` I'm guessing).

Potentially we could decorate public classes like this as well, although I can't see anybody using these classes independently of the experimental APIs they support.

src/Sentry/SentryLogLevel.cs
```diff
@@ -0,0 +1,75 @@
+using Sentry.Extensibility;
+using Sentry.Infrastructure;
+
+namespace Sentry;
+
+/// <summary>
+/// The severity of the structured log.
+/// <para>This API is experimental and it may change in the future.</para>
+/// </summary>
+/// <seealso href="https://develop.sentry.dev/sdk/telemetry/logs/"/>
```
jamescrosswell, line 10: Maybe worth adding something to the comments here about the significance of the `int` values assigned to each of the enum members. I was initially a bit surprised we were assigning values here... only worked out why by reading these docs:
- https://develop.sentry.dev/sdk/telemetry/logs/#log-severity-level

src/Sentry/SentryStructuredLogger.cs
jamescrosswell: Alternatively, if we did add an `ISentryStructuredLogger` interface, we could have two implementations of that interface: `SentryStructuredLogger` and `DisabledSentryStructuredLogger`... the implementation for the later would just be a bunch of no-ops. 

One advantage of that approach is that we could mark the dependencies for this implementation as non nullable. I'm not sure if that would simplify things enough to be worth while though (it's a pretty simple class in the first place). 

Your call I think.

test/Sentry.Tests/SentryLogTests.cs
```diff
@@ -0,0 +1,260 @@
...
+        envelope.Serialize(stream, _output, Clock);
+        stream.Seek(0, SeekOrigin.Begin);
+        using var reader = new StreamReader(stream);
+        var header = JsonDocument.Parse(reader.ReadLine()!);
+        var item = JsonDocument.Parse(reader.ReadLine()!);
+        var payload = JsonDocument.Parse(reader.ReadLine()!);
+
+        reader.EndOfStream.Should().BeTrue();
+
+        header.ToIndentedJsonString().Should().Be($$"""
```
jamescrosswell, line 47: This is the sort of thing Verify is good at testing... might be a good excuse to get familiar with that.

jamescrosswell (commented):
Looking great. 

The main suggestion I've left is about testability... which I think you'd bump into if you tried to create some unit tests for the new methods that have been added to the`Hub` class.

samples/Sentry.Samples.Console.Basic/Program.cs
```diff
@@ -33,6 +33,19 @@
 
     // This option tells Sentry to capture 100% of traces. You still need to start transactions and spans.
     options.TracesSampleRate = 1.0;
+
+    // This option enables the (experimental) Sentry Logs.
+    options.EnableLogs = true;
```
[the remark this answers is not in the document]
bruno-garcia: I guess because they have `[Experimental(Infrastructure.DiagnosticId.ExperimentalFeature)]` it could be OK without it?

samples/Sentry.Samples.Console.Basic/Program.cs
```diff
@@ -33,6 +33,19 @@
...
+    options.SetBeforeSendLog(static log =>
+    {
+        if (log.TryGetAttribute("suppress", out bool attribute) && attribute)
```
bruno-garcia, line 41:
```suggestion
        // A demonstration of how you can drop logs based on some attribute they have
        if (log.TryGetAttribute("suppress", out bool attribute) && attribute)
```

samples/Sentry.Samples.Console.Basic/Program.cs
```diff
@@ -88,6 +104,8 @@ async Task ThirdFunction()
         // Simulate doing some work
         await Task.Delay(100);
 
+        SentrySdk.Logger.LogFatal("Crash imminent!", [], static log => log.SetAttribute("suppress", true));
```
bruno-garcia, line 107: Odd that we must pass an empty array as a second argument.
There's no way we can set up the overloads in a way that this is not needed?

src/Sentry/Protocol/SentryAttribute.cs
```diff
@@ -0,0 +1,107 @@
...
+
+    public object Value { get; }
+    public string Type { get; }
+}
+
+internal static class SentryAttributeSerializer
+{
+    internal static void WriteAttribute(Utf8JsonWriter writer, string propertyName, SentryAttribute attribute)
+    {
+        Debug.Assert(attribute.Type is not null);
```
bruno-garcia, line 20: do we need this assert on all calls?
it's odd to have it here but no on the others
Or only on the method these all fall into: `WriteAttributeValue`?

bruno-garcia (commented):
just went half the PR this time but leaving some notes for now

samples/Sentry.Samples.Console.Basic/Program.cs
jamescrosswell: We might be able to have an overload that doesn't take the second `parameters` argument right?

src/Sentry/SentryStructuredLogger.cs
Flash0ver: made `SentryStructuredLogger` abstract, and moved the `IsEnabled`-check to one of the implementation's `CaptureLog` override ... so there now is only a single `IsEnabled`-check

samples/Sentry.Samples.Console.Basic/Program.cs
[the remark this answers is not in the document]
Flash0ver: The `[Experimental]`Attribute only works for C# 12 compatible compilers, and is not included in our .NET Framework nor .NET Standard targets. I believe we need to change this to the `.Experimental.` section of the `SentryOptions` indeed.

Question: Can we do the same (an `.Experimental.` section) in our `BindableSentryOptions` and derived (for `Microsoft.Extensions.Logging` and `ASP.NET Core` Options) as well?
Or should we skip the Bindable-Options for now?

src/Sentry/SentryStructuredLogger.cs
Flash0ver: Changed the `SentryStructuredLogger` to be `abstract`, and centralized all the checks to a single check in the derived `DefaultSentryStructuredLogger` override.

src/Sentry/IHub.cs
Flash0ver: I am a bit hesitant about an interface ... since adding to it is breaking.
Well ... to be fair ... we offer no facility for the user to inject a custom `SentryStructuredLogger` type, so implementing that interface is not a scenario.

I did make the `SentryStructuredLogger` now abstract, with a `Default..` and a `Disabled..` implementation, where we could later add a `Test..` implementation in `Sentry.Testing` which logs to an in-memory-collection offering assertions/verifications.

src/Sentry/SentryLog.cs
Flash0ver: `SentryLog` is used in the `SetBeforeSendLog`-provided callback.
Should user code prefer explicit typing, then `Sentry.Protocol` would need to be included.

Alternatively, we could maybe split these, in a `SentryLog` type which is user facing and basically a user-editable property bag, which is further down the pipeline used to create the Envelope and Envelope-Item via a type that is in `Sentry.Protocol`.

What do you think is best?

src/Sentry/SentryLog.cs
Flash0ver: `SentryLog` is used in callbacks, and when user code utilizes implicit typing, no warning would appear when invoking any member.

But yes ... I still need to move the Options to the experimental section, as the Attribute does not work for our .NET Standard nor .NET Framework targets.

src/Sentry/SentryLog.cs
jamescrosswell: I'd only split them if there was some functional difference... otherwise we're just making unnecessary work for ourselves. 

I guess we have these already:
https://github.com/getsentry/sentry-dotnet/blob/9beef00b325bd79c2cf43829f069142461aec113/src/Sentry/SentryOptions.cs#L466-L469
https://github.com/getsentry/sentry-dotnet/blob/9beef00b325bd79c2cf43829f069142461aec113/src/Sentry/SentryOptions.cs#L479-L482

Neither `SentryEvent` nor `SentryTransaction` are in the Protocol namespace. They could be (arguably) since they are DTOs. I'm not sure if we've been very consistent about what sits in one namespace or the other. 

In any event, not something to worry about in this PR.

src/Sentry/Protocol/SentryAttribute.cs
Flash0ver: Not necessarily ... but I thought it was a nice safeguard of "never-should-happen" for `DEBUG` builds:
Since it's a `struct`, we could technically `default` it, in which case it's members will be `null`, which Nullable-Flow-Analysis does not guard against.

The other overloads don't take that struct, but reference types directly, for which Nullable-Flow-Analysis will create a warning-as-error for us when passing a nullable reference type.

I added a `message` to clarify the intent.

src/Sentry/SentryLogLevel.cs
Flash0ver: Added a `table` in `remarks`, and a note that the lowest value of the range per severity is used for the named constants of the `enum`, as per specification.

And also a test to assert the spec.

samples/Sentry.Samples.Console.Basic/Program.cs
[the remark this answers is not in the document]
Flash0ver: Moved the Log-related options to an `Experimental` for both the Sentry-Options and the Bindable-Sentry-Options,
because the `[ExperimentalAttribute]` is not published in .NET Standard nor .NET Framework.

src/Sentry/SentryLog.cs
Flash0ver: - Options moved to
  - `SentryOptions.Experimental.EnableLogs`
  - `SentryOptions.Experimental.SetBeforeSendLog(Func<SentryLog, SentryLog?>)`

Also moved the Logger:
```diff
-Sentry.SentrySdk.Logger.LogWarning(..)
+Sentry.SentrySdk.Experimental.Logger.LogWarning(..)
```

Also ... added Tests for the Hub-Implementations.

test/Sentry.Tests/SentryLogTests.cs
Flash0ver: More "default-attributes" have been added to the Logs-specification, which I will add in a follow-up PR.
During that, will refactor to use Verify here as well.

samples/Sentry.Samples.Console.Basic/Program.cs
Flash0ver: I'll add the convenience overloads in an immediate follow-up PR.
In this one, I'd like to get the "base-shape" reviewed.

CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
+
+### Features
+
+- Add experimental support for _Sentry Structured Logging_ via `SentrySdk.Experimental.Logger` ([#4158](https://github.com/getsentry/sentry-dotnet/pull/4158))
```
```suggestion
- Add experimental support for [_Sentry Structured Logging_](https://docs.sentry.io/product/explore/logs/) via `SentrySdk.Experimental.Logger` ([#4158](https://github.com/getsentry/sentry-dotnet/pull/4158))
```

src/Sentry/Internal/Hub.cs
```diff
@@ -66,6 +66,8 @@ internal Hub(
             PushScope();
         }
 
+        Logger = new DefaultSentryStructuredLogger(this, ScopeManager, options, _clock);
```
jamescrosswell: If `options.Experimental.EnableLogs == false` we could just instantiate a `DisabledSentryStructuredLogger` here.

That way `SentryStructuredLogger.CaptureLog` would be a NoOp rather than [evaluating this](https://github.com/getsentry/sentry-dotnet/blob/6bd4c96ad66c69360e7042c4307ce6f9d3de2d1f/src/Sentry/Internal/DefaultSentryStructuredLogger.cs#L28-L31) every time it's called. And we could even remove that if enabled check from the Default logger since, implicitly, if the logger had been instantiated that means that logging is enabled (could maybe change it to a `Debug.Assert` if we really wanted).

src/Sentry/Protocol/SentryAttribute.cs
```diff
@@ -0,0 +1,107 @@
...
+        {
+            writer.WriteNumber("value", int32);
+            writer.WriteString("type", "integer");
+        }
+        else if (value is long int64)
+        {
+            writer.WriteNumber("value", int64);
+            writer.WriteString("type", "integer");
+        }
+        else if (value is double float64)
```
jamescrosswell: I tried running this in C# interactive:
```
> var value = 13.3F;
> Console.WriteLine("{0}", value is double float64 ? float64 : null);
> (1,53): error CS8121: An expression of type 'float' cannot be handled by a pattern of type 'double'.
```

I know the [docs](https://develop.sentry.dev/sdk/telemetry/logs/#log-envelope-item-payload) only refer to double, but I think we should also match floats here (even if we send them over the wire with a `type` of `double`). It would be better than sending them with a type of `string`.

src/Sentry/BindableSentryOptions.cs
```diff
@@ -53,6 +53,15 @@ internal partial class BindableSentryOptions
     public bool? EnableSpotlight { get; set; }
     public string? SpotlightUrl { get; set; }
 
+    [Experimental(Infrastructure.DiagnosticId.ExperimentalFeature)]
+    public BindableSentryExperimentalOptions Experimental { get; set; } = new();
```
jamescrosswell, line 57:
```suggestion
    internal BindableSentryExperimentalOptions Experimental { get; set; } = new();
```

src/Sentry/Internal/DefaultSentryStructuredLogger.cs
```diff
@@ -0,0 +1,131 @@
...
+        if (hub.GetSpan() is { } span && span.ParentSpanId.HasValue)
+        {
+            parentSpanId = span.ParentSpanId;
+            return true;
+        }
+
+        if (scopeManager is not null)
+        {
+            var currentScope = scopeManager.GetCurrent().Key;
+            parentSpanId = currentScope.PropagationContext.ParentSpanId;
```
jamescrosswell: This method may return true with an out value for `parentSpanId` of `null`, if the `PropagationContext` was created using this constructor:
https://github.com/getsentry/sentry-dotnet/blob/2e1fff191a83a50dd113de845a4689a2d40fd654/src/Sentry/SentryPropagationContext.cs#L36-L40

src/Sentry/SentryStructuredLogger.cs
```diff
@@ -0,0 +1,95 @@
...
+/// <para>This API is experimental and it may change in the future.</para>
+/// </summary>
+[Experimental(DiagnosticId.ExperimentalFeature)]
+public abstract class SentryStructuredLogger
+{
+    private protected SentryStructuredLogger()
+    {
+    }
+
+    private protected abstract void CaptureLog(SentryLogLevel level, string template, object[]? parameters, Action<SentryLog>? configureLog);
```
jamescrosswell, line 24: Eventually, if the Sentry SDK itself will also create logs, we probably want to make this internal rather than private and add an optional [origin](https://develop.sentry.dev/sdk/telemetry/logs/#sdk-integration-attributes) argument (that defaults to null)... or alternatively we add a bunch of internal overloads to the `LogTrace`, `LogDebug` etc. methods, accepting such a parameter. 

I'm not sure whether we plan to set that origin attribute for integrations like our NLog or Serilog integrations. Those logs do technically come from an integration but ultimately I think they're logs that come from the user.

jamescrosswell (commented):
Looking good. I left a couple of comments that probably need to be resolved before merging. 

Also, what's the plan with breadcrumbs? Various of our existing integrations create breadcrumbs automatically (e.g. our Maui integration creates breadcrumbs for lots of UI events). If structured logging is enabled, would we also want those to be sent as logs?

src/Sentry/Internal/DefaultSentryStructuredLogger.cs
```diff
@@ -0,0 +1,131 @@
...
+        {
+            message = string.Format(CultureInfo.InvariantCulture, template, parameters ?? []);
+        }
+        catch (FormatException e)
+        {
+            _options.DiagnosticLogger?.LogError(e, "Template string does not match the provided argument. The Log will be dropped.");
+            return;
+        }
+
+        SentryLog log = new(timestamp, traceId, level, message)
```
bruno-garcia: Should this be closer to the end? 
```suggestion
        var timestamp = _clock.GetUtcNow();
        SentryLog log = new(timestamp, traceId, level, message)
```

src/Sentry/Internal/DefaultSentryStructuredLogger.cs
```diff
@@ -0,0 +1,131 @@
...
+        {
+            //TODO: enqueue in Batch-Processor / Background-Worker
+            // see https://github.com/getsentry/sentry-dotnet/issues/4132
+            _ = _hub.CaptureEnvelope(Envelope.FromLog(configuredLog));
+        }
+    }
+
+    private static bool TryGetTraceId(IHub hub, IInternalScopeManager? scopeManager, out SentryId traceId)
+    {
+        if (hub.GetSpan() is { } span)
```
bruno-garcia: We also have `hub.Span` which takes precedence I believe?
The logic to extract the trace-id needs to stay in a single place somehow so we avoid different behavior throughout the SDK

src/Sentry/Internal/DefaultSentryStructuredLogger.cs
```diff
@@ -0,0 +1,131 @@
...
+        }
+
+        var timestamp = _clock.GetUtcNow();
+
+        if (!TryGetTraceId(_hub, _scopeManager, out var traceId))
+        {
+            _options.DiagnosticLogger?.LogWarning("TraceId not found");
+        }
+
+        _ = TryGetParentSpanId(_hub, _scopeManager, out var parentSpanId);
```
bruno-garcia: Trace-id above has a parent-id. I'm confused why this is split in 2 operations. We want one trace id but possibly different parents?

src/Sentry/Internal/DefaultSentryStructuredLogger.cs
bruno-garcia: Seems odd that we can get 1 trace, but the parent from some other place.

Are these not on the same trace? We need to have 1 single place where we get a trace from. 
The trace context propagation has to be 'transparent' as right now it seems we pull from "here or there, part from here and part from there" if that makes sense

src/Sentry/Protocol/SentryAttribute.cs
```diff
@@ -0,0 +1,107 @@
...
+        {
+            writer.WriteNumber("value", (long)value);
+            writer.WriteString("type", type);
+        }
+        else if (type == "double")
+        {
+            writer.WriteNumber("value", (double)value);
+            writer.WriteString("type", type);
+        }
+        else
```
bruno-garcia, line 73: Is this expected? If not, should we debug log?
Will be worse to get some reports that "I'm getting Namespace.MyType in the logs"?

bruno-garcia (commented):
did 30% or so only so far

src/Sentry/Internal/DefaultSentryStructuredLogger.cs
jamescrosswell: There's no `hub.Span`... there's a `Scope.Span` though (which is what this current code is implicitly reading):
https://github.com/getsentry/sentry-dotnet/blob/6e50eaa87b3aa24a921824c4494566b01f0640f0/src/Sentry/Internal/Hub.cs#L244

And [that prioritises the internal _span property](https://github.com/getsentry/sentry-dotnet/blob/34d849038e3e23868a12d56ea256c36b0a7f3a61/src/Sentry/Scope.cs#L617-L629) that you're talking about @bruno-garcia.

src/Sentry/Internal/Hub.cs
Flash0ver: - moved creation of `SentryStructuredLogger` to factory method
  - create "enabled" logger when options are enabled
  - create "disabled" looger when options are disabled
- remove all "enabled-checks" from the Default-Logger
  - `Debug.Assert` in the `DefaultSentryStructuredLogger` to developer-assert that the options are meant to be enabled when it's `.ctor` is invoked
- ~also renamed from `DisabledSentryStructuredLogger` to `NoOpSentryStructuredLogger`~
  - ~to match `NoOpSpan`~
  - ~to match `NoOpTransaction`~

src/Sentry/BindableSentryOptions.cs
Flash0ver: accepted ... but reverted:
This change fails `BindableSentryOptionsTests`, as they do reflection over public properties only.

src/Sentry/Internal/DefaultSentryStructuredLogger.cs
Flash0ver: My intention was to keep this as close to the user-code that invokes the creation of the log as possible, to get an (almost) precise time of the creation.

Should we do this as late as feasible instead, to avoid the cost should the log be discarded later via the `BeforeSendLog` user callback?

src/Sentry/Internal/DefaultSentryStructuredLogger.cs
Flash0ver: replaced with `Hub.GetTraceHeader()`

src/Sentry/Internal/DefaultSentryStructuredLogger.cs
Flash0ver: thanks for the hint
replaced with `Hub.GetTraceHeader()`

src/Sentry/Internal/DefaultSentryStructuredLogger.cs
Flash0ver: replaced with `Hub.GetTraceHeader()`

src/Sentry/Internal/DefaultSentryStructuredLogger.cs
Flash0ver: replaced with `Hub.GetTraceHeader()`

src/Sentry/Protocol/SentryAttribute.cs
Flash0ver: It is possible ... a path that we could warn with a Sentry-Analyzer later, if this occurs a lot.

I like the suggestion of adding a debug log.
Should this be of severity _Warning_?

src/Sentry/SentryStructuredLogger.cs
Flash0ver: will fix in the follow-up
#4193
where we add to
- `Sentry.Extensions.Logging`
- `Sentry.AspNetCore`
- `Sentry.Maui`

src/Sentry/Protocol/SentryAttribute.cs
Flash0ver: Great catch!
I'll add more tests for more the numerics.

src/Sentry/Protocol/SentryAttribute.cs
Flash0ver: while adding more tests for this case, I discovered the possibility of Null-Reference-Exception ... fixed and added Log messages when passing unsupported types

src/Sentry/Internal/DefaultSentryStructuredLogger.cs
```diff
@@ -0,0 +1,78 @@
...
+        {
+            configureLog?.Invoke(log);
+        }
+        catch (Exception e)
+        {
+            _options.DiagnosticLogger?.LogError(e, "The configureLog callback threw an exception. The Log will be dropped.");
+            return;
+        }
+
+        log.SetAttributes(_options);
```
bruno-garcia: It's odd to pass _options to `SetAttributes`. Are we setting our SDK options as attrib to the logs?

src/Sentry/Internal/DefaultSentryStructuredLogger.cs
```diff
@@ -0,0 +1,78 @@
...
+        SentryLog log = new(timestamp, traceHeader.TraceId, level, message)
+        {
+            Template = template,
+            Parameters = ImmutableArray.Create(parameters),
+            ParentSpanId = traceHeader.SpanId,
+        };
+
+        try
+        {
+            configureLog?.Invoke(log);
```
bruno-garcia, line 47: I find it odd that we have this callback to be invoked directly when capturing a log.

What are the use cases that we'll call explicitly: `Log(a,b,c,d, (log) => {})` when `log` contains the fields I anyway just provided to the `Log()` siuch as `a,b,c,d`?)

The use case for a callback when the user isn't calling the `Log()` method directly makes a lot of sense to me, and we have that below with `beforeLog` but having both is odd.

src/Sentry/SentryLog.cs
```diff
@@ -0,0 +1,281 @@
...
+    /// </summary>
+    [Experimental(DiagnosticId.ExperimentalFeature)]
+    public void SetAttribute(string key, double value)
+    {
+        _attributes[key] = new SentryAttribute(value, "double");
+    }
+
+    internal void SetAttributes(SentryOptions options)
+    {
+        var environment = options.SettingLocator.GetEnvironment();
```
bruno-garcia, line 172: is the canonical location for the release in the SDK in this place now?
Gotta make sure it checks for `options.Release` at least (besides its checks for env var, attrib etc)

src/Sentry/SentryLog.cs
```diff
@@ -0,0 +1,281 @@
...
+        SetAttribute("sentry.environment", environment);
+
+        var release = options.SettingLocator.GetRelease();
+        if (release is not null)
+        {
+            SetAttribute("sentry.release", release);
+        }
+
+        SetAttribute("sentry.sdk.name", Constants.SdkName);
```
bruno-garcia: we modify the SDK name when using the SDK from a 'wrapping' SDK such as `Sentry.Maui` and `Sentry.AspNetCore`. So we'll want to have a place where we know where to pull the right SDK name from.

Now, it'll just report as `sentry.dotnet` even if the end user installed `Sentry.Maui` in their app

src/Sentry/SentryLog.cs
```diff
@@ -0,0 +1,281 @@
...
+    /// Set a key-value pair of <see cref="double"/> data attached to the log.
+    /// <para>This API is experimental and it may change in the future.</para>
+    /// </summary>
+    [Experimental(DiagnosticId.ExperimentalFeature)]
+    public void SetAttribute(string key, double value)
+    {
+        _attributes[key] = new SentryAttribute(value, "double");
+    }
+
+    internal void SetAttributes(SentryOptions options)
```
bruno-garcia: I mentioned at a callsite that I couldn't tell what was going on from the method name and the fact it took options. It's internal so not a big deal but perhaps `SetDefaultAttributes` or `SetBaseAttributes`

test/Sentry.Tests/SentryLogTests.cs
```diff
@@ -0,0 +1,424 @@
...
+                Assert.Equal(value, getValue(property.Value));
+            }, property =>
+            {
+                Assert.Equal("type", property.Name);
+                Assert.Equal(type, property.Value.GetString());
+            });
+    }
+}
+
+file static class JsonFormatterExtensions
```
bruno-garcia, line 422: `file` TIL

bruno-garcia (approved):
other than the failing tests, it LGTM

The notes I left can be addressed in follow up PRs, I'd just merge this asap as it's quite large already

src/Sentry/SentryLog.cs
Flash0ver: Cood call! I like `SetDefaultAttributes` as per specification/documentation.

src/Sentry/Internal/DefaultSentryStructuredLogger.cs
Flash0ver: renamed to `SetDefaultAttributes`

src/Sentry/SentryLog.cs
Flash0ver: Retrieving the SDK-Name and SDK-Version from `Scope` via the `Hub`.
Thanks @jamescrosswell for pairing on this.

src/Sentry/Protocol/SentryLog.cs
Flash0ver: Fixed by reading the SDK-Name and SDK-Version from the Scope via the now available Hub.

src/Sentry/SentryLog.cs
Flash0ver: It does
1. `SentryOptions.Environment`
1. if not set: env-var
1. if not set: default

PR #4164, opened 2025-05-04
fix: Redact Authorization header before sending events to Sentry

### LINKED ISSUE #4163: Always redact Authorization headers before sending to Sentry
Sentry Replay currently redacts the `Authorization` header to avoid private information leaks. Instead, SDK users see something like this in their Sentry portal:

Image

Ideally we'd also redact this in the SDK before it crosses process boundaries though.

The `FailedRequest` handler and `HttpContextExtensions` currently capture all headers when `SendDefaultPii` is true:
https://github.com/getsentry/sentry-dotnet/blob/f765f991da77ecffd8813b772ddba55acc27d473/src/Sentry/SentryHttpFailedRequestHandler.cs#L62-L73

ScopeExtensions excludes this rather than redacting it:
https://github.com/getsentry/sentry-dotnet/blob/f765f991da77ecffd8813b772ddba55acc27d473/src/Sentry.AspNetCore/ScopeExtensions.cs#L138-L151

jamescrosswell:
Resolves #4163:
- https://github.com/getsentry/sentry-dotnet/issues/4163


src/Sentry/Protocol/Response.cs
```diff
@@ -69,6 +69,11 @@ internal void AddHeaders(IEnumerable<KeyValuePair<string, IEnumerable<string>>>
     {
         foreach (var header in headers)
         {
+            // Always skip the Authorization header - we never want to capture this even if SendDefaultPii is true
+            if (header.Key.Equals("Authorization", StringComparison.OrdinalIgnoreCase))
+            {
+                continue;
```
jamescrosswell, line 75: Potentially rather than skipping, we could mask here...

src/Sentry/Internal/RedactedHeaders.cs
```diff
@@ -0,0 +1,51 @@
+using Sentry.Internal.Extensions;
+
+namespace Sentry.Internal;
+
+internal class RedactedHeaders : IDictionary<string, string>
+{
+    private readonly Dictionary<string, string> _inner = new(StringComparer.OrdinalIgnoreCase);
+
+    private static string Redact(string key, string value) =>
+        string.Equals(key, "Authorization", StringComparison.OrdinalIgnoreCase) ? "[Filtered]" : value;
```
jamescrosswell, line 10: We could be a bit more aggressive there. The [Python SDK redacts a bunch of headers](https://github.com/getsentry/sentry-python/blob/b16fa5ffbad39843ebd2e9bc4ea6e91c0c9aa192/sentry_sdk/integrations/_wsgi_common.py#L28-L35) (not just `Authorization`):

src/Sentry/Internal/RedactedHeaders.cs
```diff
@@ -0,0 +1,53 @@
+using Sentry.Internal.Extensions;
+
+namespace Sentry.Internal;
+
+internal class RedactedHeaders : IDictionary<string, string>
+{
+    private static readonly string[] SensitiveKeys = ["Authorization", "Proxy-Authorization"];
```
jamescrosswell, line 7: We could be a bit more aggressive there. The [Python SDK redacts a bunch of headers](https://github.com/getsentry/sentry-python/blob/b16fa5ffbad39843ebd2e9bc4ea6e91c0c9aa192/sentry_sdk/integrations/_wsgi_common.py#L39).

See [Potential suspects](https://github.com/dotnet/runtime/blob/main/src/libraries/System.Net.Http/src/System/Net/Http/Headers/KnownHeaders.cs).

aritchie (approved):

src/Sentry/HttpHeadersExtensions.cs
```diff
@@ -6,4 +9,14 @@ internal static string GetCookies(this HttpHeaders headers) =>
         headers.TryGetValues("Cookie", out var values)
             ? string.Join("; ", values)
             : string.Empty;
+
+    internal static RedactedHeaders? Redact(this Dictionary<string, string?>? headers)
```
bruno-garcia, line 13:
```suggestion
    internal static RedactedHeaders? ToRedactedHeaders(this Dictionary<string, string?>? headers)
```

src/Sentry/Protocol/Response.cs
```diff
@@ -144,7 +147,7 @@ public static Response FromJson(JsonElement json)
             BodySize = bodySize,
             Cookies = cookies,
             Data = data,
-            InternalHeaders = headers?.WhereNotNullValue().ToDict(),
+            InternalHeaders = headers.Redact(),
```
bruno-garcia, line 150: With the implicit operator, do we need to call the method here?

jamescrosswell: We do unfortunately. InternalHeaders is nullable and that can't be done with an implicit operator.

Issue #4127, opened 2025-04-21
Errors with sentry-native launching with Sentry.AspNetCore on Dockerized Native AOT builds on Linux


ionite34:
### Package

Sentry

### .NET Flavor

.NET

### .NET Version

9.0.201

### OS

Linux

### OS Version

alpine 3.21

### Development Environment

Other

### SDK Version

5.5.1

### Workload Versions

N/A

### UseSentry or SentrySdk.Init call

N/A

### Steps to Reproduce

1. Using any minimal api AOT example, such as:
```cs
var builder = WebApplication.CreateSlimBuilder(args);

builder.Services.AddSentry();

var app = builder.Build();

api.MapGet("test", () => "Test Ok");

await app.RunAsync();
```

2. Building with this Dockerfile:
```
FROM mcr.microsoft.com/dotnet/nightly/sdk:9.0.201-alpine3.21-aot AS build
WORKDIR /src

COPY --link . .
RUN dotnet publish -c Release -a linux-musl-x64 -o /app ./Project/Project.csproj
RUN rm /app/*.dbg /app/*.Development.json

FROM mcr.microsoft.com/dotnet/nightly/runtime-deps:9.0.3-alpine3.21-aot AS final
USER $APP_UID
EXPOSE 8080
EXPOSE 8081
ENV ASPNETCORE_URLS=http://+:8080

WORKDIR /app
COPY --link --from=build /app .
ENTRYPOINT ["./Project"]
```

### Expected Result

Should be able to launch normally

### Actual Result

Upon hitting endpoints we'll get a 500 error with:
```
Unhandled exception. System.DllNotFoundException: Unable to load shared library 'sentry-native' or one of its dependencies. In order to help diagnose loading problems, consider using a tool like strace. If you're using glibc, consider setting the LD_DEBUG environment variable: 

Error loading shared library sentry-native.so: No such file or directory
Error loading shared library libsentry-native.so: No such file or directory
Error loading shared library sentry-native: No such file or directory
Error loading shared library libsentry-native: No such file or directory

   at System.Runtime.InteropServices.NativeLibrary.LoadLibErrorTracker.Throw(String) + 0x46
   at Internal.Runtime.CompilerHelpers.InteropHelpers.FixupModuleCell(InteropHelpers.ModuleFixupCell*) + 0x139
[...]

PR #4182, opened 2025-05-12
WIP: fix: support musl on Linux

jpnurmi:
WIP

See also:
- https://github.com/getsentry/sentry-native/issues/1232
- https://github.com/getsentry/sentry-native/pull/1233
- https://github.com/getsentry/sentry-dotnet/issues/4127


.github/workflows/build.yml
```diff
@@ -85,6 +115,14 @@ jobs:
           key: sentry-native-Linux-${{ hashFiles('scripts/build-sentry-native.ps1') }}-${{ hashFiles('.git/modules/modules/sentry-native/HEAD') }}
           fail-on-cache-miss: true
 
+      - name: Download sentry-native (Linux musl)
+        if: ${{ (env.CI_PUBLISHING_BUILD == 'true') || (startsWith(matrix.runner.container, 'alpine')) }}
+        uses: actions/cache/restore@v4
+        with:
+          path: src/Sentry/Platforms/Native/sentry-native
+          key: sentry-native-alpine:3.21-${{ hashFiles('scripts/build-sentry-native.ps1') }}-${{ hashFiles('.git/modules/modules/sentry-native/HEAD') }}
```
jamescrosswell, line 123: Maybe use `${{ !matrix.runner.container }}` here instead of hard coding `alpine:3.21`?

src/Sentry/Platforms/Native/buildTransitive/Sentry.Native.targets
jamescrosswell, line 1: 🚀 ❤️

jamescrosswell:
Looks like an awesome start @jpnurmi - massive thank you 🚀 ❤️

.github/workflows/build.yml
jpnurmi: I resorted to temporarily hard-coding it because the macOS publishing job doesn't have `matrix.runner.container` defined. How about defining a helper "target" variable in the matrix so that one could use something prettier like 'Linux-musl' as a cache key? At the same time, it could help selecting a different `.slnf` because the Android SDK Build-Tools contain musl-incompatible prebuilt binaries... :slightly_frowning_face:

.github/workflows/build.yml
```diff
@@ -48,28 +62,46 @@ jobs:
...
+      - name: Install build dependencies for Alpine Linux
+        if: ${{ startsWith(matrix.runner.container, 'alpine') }}
+        run: |
+          apk update
+          apk add bash build-base cmake curl-dev git icu mono openjdk11 powershell sudo tar wget
+          git config --global --add safe.directory $GITHUB_WORKSPACE
+          # Grpc.Tools protoc is prebuilt with glibc (https://github.com/grpc/grpc/issues/24188#issuecomment-1403435551)
+          apk add grpc-plugins
+          echo "PROTOBUF_PROTOC=/usr/bin/protoc" >> $GITHUB_ENV
+          echo "GRPC_PROTOC_PLUGIN=/usr/bin/grpc_csharp_plugin" >> $GITHUB_ENV
```
jpnurmi, line 95: Using the edge branch is probably a bad idea, as it may suddenly change, but the [mono](https://pkgs.alpinelinux.org/packages?name=mono&branch=edge) is currently only available in the edge branch, so I'm experimenting with that for now. We could consider building a custom Docker image to hide all this boilerplate...

PR #4187, opened 2025-05-13
build on windows-11-arm

bruno-garcia:
~Since I tried my luck with:~
~* https://github.com/getsentry/sentry-native/issues/880~

~Blocked by:~
~* https://github.com/getsentry/sentry-native/issues/880~
Not blocked because sentry-native itself compiles on arm64 on Windows, the issue there is more about setting up the repo and integration tests.

A lot was figured out by @jpnurmi [when adding musl Linux](https://github.com/getsentry/sentry-dotnet/pull/4188/), so I just copied over

TODO:

- [ ] Test with `env.CI_PUBLISHING_BUILD` set to `true`
- [ ] Plugin the actual sentry-native arm64 build when available, for Native AOT support


.github/workflows/build.yml
```diff
@@ -56,7 +56,7 @@ jobs:
       matrix:
         # Pin ubuntu to ensure mono is installed
         # Pin macos to get the version of Xcode that we need: https://github.com/actions/runner-images/issues/10703
-        os: [ubuntu-22.04, windows-latest, macos-15]
+        os: [ubuntu-22.04, windows-latest, windows-11-arm, macos-15]
```
Flash0ver, line 59: via `./.github/actions/environment/action.yml`

_Eclipse Temurin_ Java versions `11` (LTS) and `17` (LTS) are not available for _Windows aarch64_.
Available are `21` (LTS) and `23`.

What would upgrading or rolling forward incur?

bruno-garcia: Not sure tbqh. This isn't critical so we can look at it later

bruno-garcia: Trying through: [`283f080` (#4187)](https://github.com/getsentry/sentry-dotnet/pull/4187/commits/283f080a9462f697bf2c8ad7980522d2d62c3405)

bruno-garcia:
```
/Users/runner/.dotnet/sdk/9.0.203/Sdks/NuGet.Build.Tasks.Pack/buildCrossTargeting/NuGet.Build.Tasks.Pack.targets(221,5): error : Could not find a part of the path '/Users/runner/work/sentry-dotnet/sentry-dotnet/src/Sentry/Platforms/Native/sentry-native/win-arm64'. [/Users/runner/work/sentry-dotnet/sentry-dotnet/src/Sentry/Sentry.csproj]
```

I think now the only thing left is building sentry-native for Windows arm64 which is being done on:
* https://github.com/getsentry/sentry-native/pull/1199

jpnurmi:
Does `actions/cache` with `enableCrossOsArchive` not work on `windows-11-arm`?

1. sentry-native build succeeds on `windows-11-arm` and it saves the cache.
2. The cache artifact is visible at https://github.com/getsentry/sentry-dotnet/actions/caches
3. .NET build for `windows-11-arm` manages to restore the cache.
4. .NET build for macOS (`CI_PUBLISHING_BUILD=true`) does not find the same cache entry.

jpnurmi:
> Does `actions/cache` with `enableCrossOsArchive` not work on `windows-11-arm`?

- https://github.com/actions/partner-runner-images/issues/99

.github/workflows/build.yml
```diff
@@ -11,26 +11,36 @@ on:
...
+        shell: pwsh
+        run: choco install zstandard
+
       - uses: actions/cache@v4
         id: cache
         with:
           path: src/Sentry/Platforms/Native/sentry-native
-          key: sentry-native-${{ runner.os }}-${{ hashFiles('scripts/build-sentry-native.ps1') }}-${{ hashFiles('.git/modules/modules/sentry-native/HEAD') }}
+          # On Windows, it'll be either: Windows or Windows-arm64:
+          key: sentry-native-${{ matrix.arch && format('{0}-{1}', runner.os, matrix.arch) || runner.os }}-${{ hashFiles('scripts/build-sentry-native.ps1') }}-${{ hashFiles('.git/modules/modules/sentry-native/HEAD') }}
```
jpnurmi, line 43: how about using `matrix.os`?

```yaml
key: sentry-native-${{ matrix.os }}-${{ hashFiles(...) }}-${{ hashFiles(...) }}
```

.github/workflows/build.yml
```diff
@@ -11,26 +11,36 @@ on:
 
 jobs:
   build-sentry-native:
-    name: sentry-native (${{ matrix.os }})
+    name: sentry-native (${{ matrix.os }}${{ matrix.arch && format('-{0}', matrix.arch) || '' }}
```
jpnurmi, line 14: perhaps this change can be reverted. the name becomes `sentry-native (windows-11-arm-arm64`

- the arch suffix is not necessary in this context
- missing closing parenthesis

.github/workflows/build.yml
```diff
@@ -94,36 +110,42 @@ jobs:
...
+          key: sentry-native-Windows-arm64-${{ hashFiles('scripts/build-sentry-native.ps1') }}-${{ hashFiles('.git/modules/modules/sentry-native/HEAD') }}
+          fail-on-cache-miss: true
+          enableCrossOsArchive: true
 
       - name: Build Native Dependencies
         uses: ./.github/actions/buildnative
 
       - name: Restore .NET Dependencies
-        run: dotnet restore Sentry-CI-Build-${{ runner.os }}.slnf --nologo
+        run: dotnet restore Sentry-CI-Build-${{ matrix.arch && format('{0}-{1}', runner.os, matrix.arch) || runner.os }}.slnf --nologo
```
jpnurmi, line 134: idea: maybe we could define solution filter names from the matrix to make it easier to understand?

```yaml
matrix:
  include:
    - os: ubuntu-22.04
      slnf: Sentry-CI-Build-Linux.slnf
    - os: macos-15
      slnf: Sentry-CI-Build-macOS.slnf
    - os: windows-latest
      slnf: Sentry-CI-Build-Windows.slnf
    - os: windows-11-arm
      slnf: Sentry-CI-Build-Windows-arm64.slnf

...

    - run: dotnet restore ${{ matrix.slnf }} ...
```

.github/workflows/build.yml
jpnurmi: Done, makes it also easier to merge with the musl changes:
- https://github.com/getsentry/sentry-dotnet/pull/4188

.github/workflows/build.yml
jpnurmi: Done

image

.github/workflows/build.yml
jpnurmi: Ditto, makes it easier to merge with the musl changes:
- https://github.com/getsentry/sentry-dotnet/pull/4188

.github/workflows/build.yml
```diff
@@ -205,15 +232,15 @@ jobs:
...
+      # We use macOS for the final publishing build so we get all the iOS/Mac Catalyst targets in the packages
       - name: Set Environment Variables
         run: echo "CI_PUBLISHING_BUILD=true" >> $GITHUB_ENV
 
       - name: Download sentry-native (macOS)
         uses: actions/cache/restore@v4
         with:
           path: src/Sentry/Platforms/Native/sentry-native
-          key: sentry-native-macOS-${{ hashFiles('scripts/build-sentry-native.ps1') }}-${{ hashFiles('.git/modules/modules/sentry-native/HEAD') }}
+          key: sentry-native-macos-15-${{ hashFiles('scripts/build-sentry-native.ps1') }}-${{ hashFiles('.git/modules/modules/sentry-native/HEAD') }}
```
bruno-garcia, line 243: @jpnurmi We'll prob forget to update the `15`, should we use the agent name as a variable?

jpnurmi: Good point. It would not be good if it accidentally used the old cache after bumping the version.

This part of the cache key is hard-coded when downloading the cache, because the publishing build has to fetch caches from all runners.

jamescrosswell:
Just to make sure I understand what's going on here before we merge:
- We've added another runner (Windows ARM64) to make sure that all of our unit tests etc pass when using versions of the SDK and integrations compiled for and running on Windows ARM64
- In terms of the actual NuGet packages (which all get built by the MacOS runner), the only thing that's changed is that we now ship a version of SentryNative that is compiled for Windows ARM64... which is the main thing that wasn't working on Windows ARM64 before (and is required for NativeAOT to work on Windows ARM64)?

Is that all accurate?

bruno-garcia:
> Just to make sure I understand what's going on here before we merge:
> 
> * We've added another runner (Windows ARM64) to make sure that all of our unit tests etc pass when using versions of the SDK and integrations compiled for and running on Windows ARM64
> * In terms of the actual NuGet packages (which all get built by the MacOS runner), the only thing that's changed is that we now ship a version of SentryNative that is compiled for Windows ARM64... which is the main thing that wasn't working on Windows ARM64 before (and is required for NativeAOT to work on Windows ARM64)?
> 
> Is that all accurate?

Right, we already had a Windows x64 build that the packaging build (on Mac) pulled from. So this would add a second one.

bruno-garcia:
Running a release off of this branch just to test it out: https://github.com/getsentry/sentry-dotnet/actions/runs/15174949147

cc @Flash0ver if we ship this (not planning on doing that, just want to test CI for a release build) it would take over the `.1` number

jamescrosswell (approved):

PR #4188, opened 2025-05-14
fix: support musl on Linux 

jpnurmi:
This PR adds native AOT support for `linux-musl-x64`:

```
Sentry.5.8.0.nupkg
└── sentry-native
    ├── linux-musl-x64
    │   └── libsentry-native.a # <== NEW
    ├── linux-x64
    │   └── libsentry-native.a
    ├── osx
    │   └── libsentry-native.a
    ├── win-arm64
    │   └── sentry-native.lib
    └── win-x64
        └── sentry-native.lib
```

TODO:
- [x] Build system
  - [x] Sentry.Native.targets
  - [x] .slnf without Android
- [x] CI
  - [x] build / sentry-native
  - [x] build / .NET
  - [x] Integration test
  - [x] Docker image: https://github.com/getsentry/sentry-dotnet/pull/4190

Notes:
- setup-dotnet is musl-compatible :+1:
- setup-java installs musl-incompatible binaries :-1: 
- setup-android installs musl-incompatible prebuilt (Android SDK Build-Tools) binaries :-1: 

Close: #4127


src/Sentry/Platforms/Native/Sentry.Native.targets
```diff
@@ -11,10 +11,13 @@
...
+    <NativeLibRelativePath-linux-musl-x64>linux-musl-x64</NativeLibRelativePath-linux-musl-x64>
+    <SentryNativeOutputDirectory-linux-musl-x64>$(SentryNativeOutputDirectory)$(NativeLibRelativePath-linux-musl-x64)\</SentryNativeOutputDirectory-linux-musl-x64>
     <NativeLibRelativePath-osx>osx</NativeLibRelativePath-osx>
     <SentryNativeOutputDirectory-osx>$(SentryNativeOutputDirectory)$(NativeLibRelativePath-osx)\</SentryNativeOutputDirectory-osx>
     <SentryNativeBuildOutputs Condition="$([MSBuild]::IsOsPlatform('Windows'))">$(SentryNativeOutputDirectory-win-x64)$(SentryNativeLibraryName).lib</SentryNativeBuildOutputs>
-    <SentryNativeBuildOutputs Condition="$([MSBuild]::IsOsPlatform('Linux'))">$(SentryNativeOutputDirectory-linux-x64)lib$(SentryNativeLibraryName).a</SentryNativeBuildOutputs>
+    <SentryNativeBuildOutputs Condition="'$(RuntimeIdentifier)' == 'linux-x64'">$(SentryNativeOutputDirectory-linux-x64)lib$(SentryNativeLibraryName).a</SentryNativeBuildOutputs>
+    <SentryNativeBuildOutputs Condition="'$(RuntimeIdentifier)' == 'linux-musl-x64'">$(SentryNativeOutputDirectory-linux-musl-x64)lib$(SentryNativeLibraryName).a</SentryNativeBuildOutputs>
```
jpnurmi, line 20: Is there any potential downside to checking the runtime identifier instead of the platform reported by msbuild? Are there any relevant cross-build scenarios we should consider?

bruno-garcia: hm, I'm not sure. I know that RID can be explicitly set, to make a build that's specific for a platform.
Or not set (it's optinal is what I mean) in which case the build is not platform specific. But I haven't looked at it in details in years. Maybe @Flash0ver has tips/pointers

bruno-garcia: > Publish the app for a specific runtime identifier using dotnet publish -r <RID>.

[Looks like a RID is required](https://learn.microsoft.com/en-us/dotnet/core/deploying/native-aot/?tabs=windows%2Cnet8) in order to build for Native AOT (which makes a lot of sense) so this shouldn't be an isssue

bruno-garcia:
> Is the approach with Sentry-CI-Build-Linux-musl.slnf sensible? I wish we could use NO_MOBILE=true or NO_ANDROID=true with the existing Sentry-CI-Build-Linux.slnf, but it pulls in several projects that fail to build without Android SDK:

I'd say so, and I copied your approach on https://github.com/getsentry/sentry-dotnet/pull/4187/ (which won't land anytime soon since it's blocked on `sentry-native` win-arm64 support but at least will be ready once that lands

jamescrosswell:
> Is the approach with `Sentry-CI-Build-Linux-musl.slnf` sensible? I wish we could use `NO_MOBILE=true` or `NO_ANDROID=true` with the existing `Sentry-CI-Build-Linux.slnf`, but it pulls in several projects that fail to build without Android SDK:

Yeah I think that makes sense. The names we give to these solutions are maybe what's confusing. Implicitly building on MUSL we're trying to provide better support for SDK users deploying ASP.NET Core apps using docker containers... so no android required.

However `Sentry-CI-Build-Linux.slnf` is just the base solution for contributors to the Sentry SDK who are running Linux and they could well be building Android apps... so we can't really rip it out of there.

src/Sentry/Platforms/Native/Sentry.Native.targets
jpnurmi: For some reason, checking for RID did not work as expected in this context :(

With RID checks:
```
$ dotnet build Sentry-CI-Build-Linux-musl.slnf -c Release
[...]
$ tree src/Sentry/Platforms/Native/
src/Sentry/Platforms/Native/
├── CFunctions.cs
├── NativeContextWriter.cs
├── NativeScopeObserver.cs
├── Sentry.Native.targets
├── SentryNative.cs
├── SentrySdk.cs
├── buildTransitive
│   └── Sentry.Native.targets
└── windows-config.cmake
```

Without RID checks:
```
$ dotnet build Sentry-CI-Build-Linux-musl.slnf -c Release
[...]
$ tree src/Sentry/Platforms/Native/
src/Sentry/Platforms/Native/
├── CFunctions.cs
├── NativeContextWriter.cs
├── NativeScopeObserver.cs
├── Sentry.Native.targets
├── SentryNative.cs
├── SentrySdk.cs
├── buildTransitive
│   └── Sentry.Native.targets
├── sentry-native
│   └── linux-musl-x64
│       └── libsentry-native.a # <==
└── windows-config.cmake

3 directories, 9 files
```

Flash0ver: I wonder if we're fine with the slightly disjoint behavior for local builds ... if I read it correctly:

For ordinary builds (`dotnet build` or build via IDE)
- on _Windows_ and _macOS_  we also build the native dependencies, although not necessarily required
- on _Linux_ we do not build the native dependencies during ordinary builds locally

However, when building for a specific platform, and in particular AOT, then the `$(RuntimeIdentifier)` is set and we build the native dependencies also on _Linux_.

If I understand the `Inputs="$(SentryNativeBuildInputs)" Outputs="$(SentryNativeBuildOutputs)"` of the `_BuildSentryNativeSDK` correctly.
Actually, I'm not entirely certain: if `$(SentryNativeBuildOutputs)` is empty ... which it is in the case of an "ordinary" build on _Linux_, when the `$(RuntimeIdentifier)` is set, will the `_BuildSentryNativeSDK` Target below then be invoked or not invoked?

jpnurmi: Aha, thanks! So it might just be my misunderstanding. ;) I compared what outputs were present after `dotnet build Sentry-CI-Build-Linux(-musl).slnf` on Ubuntu vs. Alpine Linux and tried to achieve the same.

On Ubuntu, if I run `dotnet workload restore` + `dotnet build Sentry-CI-Build-Linux.slnf` in a clean sentry-dotnet checkout of the main branch, it does seem to build `src/Sentry/Platforms/Native/sentry-native/linux-x64/libsentry-native.a`.

jamescrosswell:
I'm not sure if it's caused you any issues here but just FYI:
- https://github.com/getsentry/sentry-dotnet/issues/4203

bruno-garcia (approved):

PR #4193, opened 2025-05-16
feat(logs): Logs for `Sentry.Extensions.Logging` and integrations for `Sentry.AspNetCore` and `Sentry.Maui`

### LINKED ISSUE #4351: Experimental Sentry Logs not working
### Package

Sentry.AspNetCore

### .NET Flavor

.NET

### .NET Version

9.0.302

### OS

Windows

### OS Version

OS Name: Microsoft Windows 11 Pro
OS Version: 10.0.26100 N/A Build 26100

### Development Environment

Visual Studio v17.x

### SDK Version

5.12.0-alpha.0

Sentry 25.6.2

### Workload Versions
| Installed Workload Id  | Manifest Version | Installation Source |
| ------------- | ------------- | ------------- |
aspire | 8.2.2/8.0.100 | VS 17.14.36301.6

```json
"Sentry": {
  "SendDefaultPii": true,
  "MaxRequestBodySize": "Always",
  "MinimumBreadcrumbLevel": "Debug",
  "MinimumEventLevel": "Warning",
  "AttachStackTrace": true,
  "Debug": true,
  "DiagnosticLevel": "Error",
  "TracesSampleRate": 1.0
},
```

```csharp
{
    options.Dsn = "***";

#pragma warning disable SENTRY0001 // Type is for evaluation purposes only and is subject to change or removal in future updates. Suppress this diagnostic to proceed.
    options.Experimental.EnableLogs = true;
    options.ExperimentalLogging.MinimumLogLevel = LogLevel.Debug;
#pragma warning restore SENTRY0001 // Type is for evaluation purposes only and is subject to change or removal in future updates. Suppress this diagnostic to proceed.

    options.AddEntityFramework();
});

builder.Logging.AddSentry();
builder.Services.AddSentry();
```

### Steps to Reproduce

1. Just start the application

### Expected Result

Logs are committed to sentry

### Actual Result

```yaml
  Debug: Registering EF Core integration
  Error: Template string does not match the provided argument. The Log will be dropped.
System.FormatException: Input string was not in a correct format. Failure to parse near offset 136. Expected an ASCII digit.
   at System.Text.ValueStringBuilder.AppendFormatHelper(IFormatProvider provider, String format, ReadOnlySpan`1 args)
   at System.String.FormatHelper(IFormatProvider provider, String format, ReadOnlySpan`1 args)
   at System.String.Format(IFormatProvider provider, String format, Object[] args)
   at Sentry.Internal.DefaultSentryStructuredLogger.CaptureLog(SentryLogLevel level, String template, Object[] parameters, Action`1 configureLog)
```

```yaml
fail: Sentry.ISentryClient[0]
      Template string does not match the provided argument. The Log will be dropped.
      System.FormatException: Input string was not in a correct format. Failure to parse near offset 48. Expected an ASCII digit.
         at System.Text.ValueStringBuilder.AppendFormatHelper(IFormatProvider provider, String format, ReadOnlySpan`1 args)
         at System.String.FormatHelper(IFormatProvider provider, String format, ReadOnlySpan`1 args)
         at System.String.Format(IFormatProvider provider, String format, Object[] args)
         at Sentry.Internal.DefaultSentryStructuredLogger.CaptureLog(SentryLogLevel level, String template, Object[] parameters, Action`1 configureLog)
```

Flash0ver:
Initial implementation for `Sentry.Extensions.Logging`
Initial integrations for
- `Sentry.AspNetCore`
- `Sentry.Maui`

Logs: https://develop.sentry.dev/sdk/telemetry/logs/

fixes #4351

### Added
- Log implementation via `ILogger`
  - `Sentry.Extensions.Logging.SentryStructuredLogger`
- Log integrations via `ILoggerProvider` for
  - `Sentry.Extensions.Logging`
  - `Sentry.AspNetCore`
  - `Sentry.Maui`
- Option (and BindableOption)
  - `LogLevel MinimumLogLevel`
- _Benchmark_ for new ILogger
- A Polyfill method that is not provided by the `Polyfill` package
### Changed
- `SentryLog.Parameters`, used via `SentryOptions.SetBeforeSendLog` callback
  - ```diff
    -public ImmutableArray<object> Parameters { get; init; }
    +public ImmutableArray<KeyValuePair<string, object>> Parameters { get; init; }
     ```
  - where the `SentrySdk.Logger` APIs keep setting the index-based parameter names
  - and the new Log-Integration sets the named parameter names from `Microsoft.Extensions.Logging`
### Fixed
- Log `timestamp` now sent as `double` rather than `int` to include milliseconds

### Testing
Apart from unit testing, manually tested via Samples (Console, ASP.NET Core, and .NET MAUI),
via different means of logging:
```CSharp
// ILogger Extension Methods
logger.LogInformation(new EventId(1, "EventName"), "String={Name}, Number={Number}, Boolean={Boolean}", "Hello", 2025, true);

// LoggerMessage.Define
internal static readonly Action<ILogger, string, int, bool, Exception> Hello = LoggerMessage.Define<string, int, bool>(
  LogLevel.Information, new EventId(1, "EventName"), "String={Name}, Number={Number}, Boolean={Boolean}");

// Source-Generated Logging
[LoggerMessage(EventId = 1, Level = LogLevel.Information, Message = "String={Name}, Number={Number}, Boolean={Boolean}")]
internal static partial void Hello(this ILogger logger, string name, int number, bool boolean);
```

image


src/Sentry.AspNetCore/SentryWebHostBuilderExtensions.cs
```diff
@@ -93,10 +93,16 @@ public static IWebHostBuilder UseSentry(
...
                 .AddSingleton<IConfigureOptions<SentryAspNetCoreOptions>, SentryAspNetCoreOptionsSetup>();
             _ = logging.Services.AddSingleton<ILoggerProvider, SentryAspNetCoreLoggerProvider>();
+            _ = logging.Services.AddSingleton<ILoggerProvider, SentryAspNetCoreStructuredLoggerProvider>();
 
             _ = logging.AddFilter<SentryAspNetCoreLoggerProvider>(
                 "Microsoft.AspNetCore.Diagnostics.ExceptionHandlerMiddleware",
                 LogLevel.None);
+            _ = logging.AddFilter<SentryAspNetCoreStructuredLoggerProvider>(static (string? categoryName, LogLevel logLevel) =>
+            {
+                return categoryName is null
```
bruno-garcia, line 103: are these guaranteed to be all of the categories we add?
I wonder if there's some "Constant" or something with these, we could reflect on and write a test to make sure we filter all of them out

Flash0ver: I'm fairly certain ... not quite positive though. Those are the `ILogger<T>` usages that we're using in our packages.
I definitely have to add a test for that ... probably as an ASP.NET Core integration test ... I see we have the infrastructure already setup via the "TestUtils" project.

.github/actions/environment/action.yml
```diff
@@ -11,39 +11,46 @@ runs:
         echo "DOTNET_CLI_TELEMETRY_OPTOUT=1" >> $GITHUB_ENV
         echo "DOTNET_NOLOGO=1" >> $GITHUB_ENV
 
+    # Android SDK Tools are arm64-incompatible
```
jamescrosswell: Is this from another PR? It seems unrelated to logs...

Flash0ver: Indeed ... made a mistake when merging from the wrong source branch ... fixed.

CHANGELOG.md
```diff
@@ -1,5 +1,15 @@
 # Changelog
 
+## Unreleased
+
+### Features
+
+- Add experimental support for [Sentry Structured Logging](https://docs.sentry.io/product/explore/logs/) via `SentrySdk.Experimental.Logger` ([#4158](https://github.com/getsentry/sentry-dotnet/pull/4158))
+- Add experimental integrations of [Sentry Structured Logging](https://docs.sentry.io/product/explore/logs/) ([#4193](https://github.com/getsentry/sentry-dotnet/pull/4193))
```
jamescrosswell: If #4158 and #4193 are both going in the same release, we could probably compress it into a single changelog entry (referencing multiple PRs, as we often do for the dependency bumps)... maybe with some bullet/sub-list items for each of the integrations. It's a bit confusing otherwise having two change log entries announcing structured logging.

Flash0ver: merged

Flash0ver:
@Sentry review

src/Sentry.Extensions.Logging/SentryStructuredLogger.cs
```diff
@@ -0,0 +1,121 @@
...
+        string? template = null;
+        var parameters = ImmutableArray.CreateBuilder<KeyValuePair<string, object>>();
+        // see Microsoft.Extensions.Logging.FormattedLogValues
+        if (state is IReadOnlyList<KeyValuePair<string, object?>> formattedLogValues)
+        {
+            if (formattedLogValues.Count != 0)
+            {
+                parameters.Capacity = formattedLogValues.Count - 1;
+            }
+
```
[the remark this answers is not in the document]
Flash0ver: When `formattedLogValues.Count` is `0`, then the capacity is not set, which is guarded by the _If-Statement_ above.
Also, the setter of `System.Collections.Immutable.ImmutableArray<T>.Builder.Capacity` does not throw if `0` or less than `0` is set, but instead internally sets an empty array. This setter only throws an `ArgumentException` if the `Capacity` set is less than the current `Count`.
Also, the capacity actually does account for the fact that one item is the `{OriginalFormat}` entry, so the actual parameter count is `formattedLogValues.Count - 1`.

Anyway ... your feedback got me thinking:
Setting the `Capacity` on an `ImmutableArray<T>.Builder` is a performance optimization, to avoid unnecessary growth of the backing array.
But, I currently use `ImmutableArray<T>.Builder.ToImmutable()` to convert the builder into an `ImmutableArray<T>`, which always creates a copy of the internal array, since the builder instance may be reused. However, we do not reuse the builder, so we could actually use `DrainToImmutable()`, which will move the internal array to the new `ImmutableArray<T>` instance if the `Capacity` equals the `Count`, which is expected to be the case. Unlike `MoveToImmutable()`, which will throw an `InvalidOperationException` when the `Capacity` does not equal the `Count`, `DrainToImmutable()` will not throw in that case and actually creates a copy of the array with a length of `Count`, so we are safe.

Additionally, I will add a Benchmark to confirm that expectation.

src/Sentry.Extensions.Logging/SentryStructuredLogger.cs
```diff
@@ -0,0 +1,121 @@
...
+        Debug.Assert(level != default);
+
+        string message;
+        try
+        {
+            message = formatter.Invoke(state, exception);
+        }
+        catch (FormatException e)
+        {
+            _options.DiagnosticLogger?.LogError(e, "Template string does not match the provided argument. The Log will be dropped.");
```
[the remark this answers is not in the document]
Flash0ver: `formatter` is not null.

`Microsoft.Extensions.Logging` is annotated with C# 8 _Nullable Reference Types_.
The fourth parameter to this method, `Exception? exception`, is marked nullable and may be null.
The fifth parameter, `Func<TState, Exception?, string> formatter`, however, is not null.

This implemented method, `ILogger.Log<TState>`, is usually not called by user code directly,
but indirectly via library code, which does pass a non-null `formatter`.
If this was a method intended to be called by user code, since it's `public`, we would need to perform a null-check indeed.

src/Sentry/Internal/DefaultSentryStructuredLogger.cs
```diff
@@ -40,10 +40,21 @@ private protected override void CaptureLog(SentryLogLevel level, string template
             return;
         }
 
+        ImmutableArray<KeyValuePair<string, object>> @params = default;
+        if (parameters is not null && parameters.Length != 0)
+        {
+            var builder = ImmutableArray.CreateBuilder<KeyValuePair<string, object>>(parameters.Length);
+            for (var index = 0; index < parameters.Length; index++)
+            {
```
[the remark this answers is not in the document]
Flash0ver: `Microsoft.Extensions.Logging`-based formatting does support both index-based and named parameters, and even encourages the use of named parameters indeed.
The older `string.Format`-based formatting however only supports index-based parameters.

To be fair, we could change the implementation of `SentryStructuredLogger`, and `DefaultSentryStructuredLogger` in particular, to support named parameters as well. But index-based parameterization are expected to be supported at least.
We expect most users to use our `Microsoft.Extensions.Logging`-based integrations to send Logs to Sentry with.
If there is actual user demand to support named parameters for the `SentryStructuredLogger`-based implementation as well, we can add it in a non-breaking and backwards-compatible fashion in the future.

src/Sentry/SentryLog.cs
```diff
@@ -192,7 +202,11 @@ internal void WriteTo(Utf8JsonWriter writer, IDiagnosticLogger? logger)
     {
         writer.WriteStartObject();
 
-        writer.WriteNumber("timestamp", Timestamp.ToUnixTimeSeconds());
+#if NET9_0_OR_GREATER
+        writer.WriteNumber("timestamp", Timestamp.ToUnixTimeMilliseconds() / (double)TimeSpan.MillisecondsPerSecond);
+#else
+        writer.WriteNumber("timestamp", Timestamp.ToUnixTimeMilliseconds() / 1_000.0);
+#endif
```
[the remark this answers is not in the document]
Flash0ver: The precision is equivalent in both Target Frameworks. We have multi-targeted tests confirming the equivalent outcome.

src/Sentry.Extensions.Logging/SentryStructuredLogger.cs
```diff
@@ -0,0 +1,121 @@
...
+
+        log.SetDefaultAttributes(_options, _sdk);
+
+        if (_categoryName is not null)
+        {
+            log.SetAttribute("microsoft.extensions.logging.category_name", _categoryName);
+        }
+        if (eventId.Name is not null || eventId.Id != 0)
+        {
+            log.SetAttribute("microsoft.extensions.logging.event.id", eventId.Id);
```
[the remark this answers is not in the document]
Flash0ver: This is correct.
However, the _readonly struct_ `Microsoft.Extensions.Logging.EventId` (from the `Microsoft.Extensions.Logging.Abstractions` NuGet package) only has the following constructor:
```CSharp
public EventId(int id, string? name = null);
```
So you cannot set the `name` without also setting the `id`:
If `EventId.Id is 0` and `EventId.Name is null`, we cannot differentiate between
- creation via `default`
- creation via `new EventId()`
- creation via `new EventId(0)`
- creation via `new EventId(0, null)`
- implicit cast from an `int` via `EventId id = 0`

If `EventId.Name is not null`, we do know that the constructor was called, which means that the user must have also passed an `int id`, be it `default`, `0` or an integer other than `0`.
Therefore we derived the logic:
- when `EventId.Name is not null`, `EventId.Id` must have been passed as well and we consider both to be set by the user and write both as an attribute
- when `EventId.Name is null` and `EventId.Id is not 0`, an `id` has been set via either the constructor or an implicit conversion, so we only set the `Id` as an attribute, but not the `Name `
- when `EventId.Name is null` and `EventId.Id is 0`, we can't determine whether the struct instance is `default` or was explicitly created with `0`, and since `0` is the `default` value of an `int` we assume that no `Id` has been set and therefore we don't set either attribute

src/Sentry.Extensions.Logging/SentryLoggingOptions.cs
```diff
@@ -48,4 +50,39 @@ public class SentryLoggingOptions : SentryOptions
...
+    /// </remarks>
+    [Experimental(Infrastructure.DiagnosticId.ExperimentalFeature)]
+    public sealed class SentryLoggingExperimentalOptions
+    {
+        internal SentryLoggingExperimentalOptions()
+        {
+        }
+
+        /// <summary>
+        /// Gets or sets the minimum log level.
```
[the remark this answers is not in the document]
Flash0ver: Fair point.

The spec/developer docs, see https://develop.sentry.dev/sdk/telemetry/logs/, mention:

>### SDK Integrations
>
>SDKs should aim to have it so that console/logger integrations create logs as per the appropriate log level if `enableLogs`/`enable_logs` is set to true. Examples of this include JavaScript's `console` object and Pythons `logging` standard library.
>
>If SDK authors feel the need, they can also introduce additional options to beyond `enableLogs`/`enable_logs` to gate this functionality. For example an option to control log appenders added via external config like with `Log4j` in the Java SDK.

Per default, no logs are sent to Sentry, regardless of the `MinimumLogLevel`.
An explicit opt-in via `SentryOptions.Experimental.EnableLogs = true` (or through configuration and the `BindableSentryOptions`) enables both the Sentry-Logger API, as well as the `Sentry.Extensions.Logging`/`Sentry.AspNetCore`/`Sentry.Maui` integration.

@jamescrosswell @bruno-garcia 
What do you think? What should be the defaults? Currently it's:
```CSharp
SentryOptions.Experimental.EnableLogs = false;
SentryLoggingOptions.ExperimentalLogging.MinimumLogLevel = LogLevel.Trace;
```
which means that if `EnableLogs is true`, all `Microsoft.Extensions.Logging.LogLevel` are sent.

src/Sentry/Polyfilling/ImmutableCollectionsPolyfill.cs
```diff
@@ -0,0 +1,20 @@
+// ReSharper disable CheckNamespace
+namespace System.Collections.Immutable;
+
+internal static class ImmutableCollectionsPolyfill
+{
+#if !NET8_0_OR_GREATER
+    internal static ImmutableArray<T> DrainToImmutable<T>(this ImmutableArray<T>.Builder builder)
```
Flash0ver, line 7: note: this Polyfill is not contained in `Polyfill` (as of `8.7.3`)

test/Sentry.Testing/InMemorySentryStructuredLogger.cs
```diff
@@ -12,6 +12,12 @@ private protected override void CaptureLog(SentryLogLevel level, string template
         Entries.Add(LogEntry.Create(level, template, parameters));
     }
 
+    /// <inheritdoc />
+    protected internal override void CaptureLog(SentryLog log)
+    {
+        throw new NotSupportedException();
+    }
+
```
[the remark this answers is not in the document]
Flash0ver: Correct ... but this functionality is not used by any test yet ... tests are all passing ... only the other overload of `InMemorySentryStructuredLogger.CaptureLog` is currently in use.
I didn't want to just implement a method without any usage yet ... "consumer-driven development".
Once this overload will be used in testing too, the first consumer may implement it to their needs.

This doesn't mean that this overload is not tested in our production code.
The tests ... related to "Sentry Logs for ` Microsoft.Extensions.Logging`" ... are using `NSubstitute` as it was less boilerplate to implement and use in the tests.

src/Sentry/Internal/DefaultSentryStructuredLogger.cs
```diff
@@ -40,10 +40,21 @@ private protected override void CaptureLog(SentryLogLevel level, string template
             return;
         }
 
+        ImmutableArray<KeyValuePair<string, object>> @params = default;
+        if (parameters is not null && parameters.Length != 0)
```
jamescrosswell:
```suggestion
        if (parameters is { Length: > 0 })
```

src/Sentry.AspNetCore/SentryWebHostBuilderExtensions.cs
jamescrosswell: There's this one (although admittedly it's not used):
https://github.com/getsentry/sentry-dotnet/blob/c5db0917fe5adc3228662be0a2633d18de6227dd/src/Sentry.AspNetCore/RequestDecompression/RequestDecompressionMiddleware.cs#L33

btw: Why an integration test for this? It just needs to use reflection to iterate over a bunch of assemblies right? Should be possible in a unit test?

src/Sentry.Extensions.Logging/BindableSentryLoggingOptions.cs
```diff
@@ -9,11 +9,22 @@ internal class BindableSentryLoggingOptions : BindableSentryOptions
     public LogLevel? MinimumEventLevel { get; set; }
     public bool? InitializeSdk { get; set; }
 
+    [Experimental(Infrastructure.DiagnosticId.ExperimentalFeature)]
```
jamescrosswell: Not sure we need these (it's an `internal` class)... just more work to clean them up afterwards.

src/Sentry.Extensions.Logging/BindableSentryLoggingOptions.cs
```diff
@@ -9,11 +9,22 @@ internal class BindableSentryLoggingOptions : BindableSentryOptions
...
+    public BindableSentryLoggingExperimentalOptions ExperimentalLogging { get; set; } = new();
+
+    [Experimental(Infrastructure.DiagnosticId.ExperimentalFeature)]
```
jamescrosswell: Same again: any reason to mark an internal member Experimental?

src/Sentry.Extensions.Logging/LogLevelExtensions.cs
```diff
@@ -45,4 +45,19 @@ public static SentryLevel ToSentryLevel(this LogLevel level)
...
+        return logLevel switch
+        {
+            LogLevel.Trace => SentryLogLevel.Trace,
+            LogLevel.Debug => SentryLogLevel.Debug,
+            LogLevel.Information => SentryLogLevel.Info,
+            LogLevel.Warning => SentryLogLevel.Warning,
+            LogLevel.Error => SentryLogLevel.Error,
+            LogLevel.Critical => SentryLogLevel.Fatal,
+            LogLevel.None => default,
+            _ => default,
```
jamescrosswell, line 60: `default` for an enum value is always `0` right? And in this case, there is no enum value defined for `SentryLogLevel` with a value of `0`. Won't that potentially cause problems?

src/Sentry.Extensions.Logging/SentryLoggingOptions.cs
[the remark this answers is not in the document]
jamescrosswell: Hm, that could be quite verbose. `Information` might be a more typical default log level. 

MEL let's you [configure different log levels for different categories](https://learn.microsoft.com/en-us/aspnet/core/fundamentals/logging/?view=aspnetcore-9.0#configure-logging) as well... so eventually we might want to mimic/reuse that configuration.

src/Sentry.Extensions.Logging/SentryStructuredLogger.cs
```diff
@@ -0,0 +1,121 @@
...
+        _categoryName = categoryName;
+        _options = options;
+        _clock = clock;
+        _hub = hub;
+        _sdk = sdk;
+    }
+
+    public IDisposable? BeginScope<TState>(TState state) where TState : notnull
+    {
+        return NullDisposable.Instance;
```
jamescrosswell, line 26: What's the purpose of this method?

In our other logger, I think it's used as a convenient way to set a bunch of key/value pair properties on all logs within a particular scope:
https://github.com/getsentry/sentry-dotnet/blob/9ad2caa0c1e21296a90242c06e564d079a59020f/src/Sentry.Extensions.Logging/SentryLogger.cs#L27-L28

... in which case, do we want to do the same here?

src/Sentry.Extensions.Logging/SentryStructuredLogger.cs
```diff
@@ -0,0 +1,121 @@
...
+    }
+
+    public bool IsEnabled(LogLevel logLevel)
+    {
+        return _hub.IsEnabled
+            && _options.Experimental.EnableLogs
+            && logLevel != LogLevel.None
+            && logLevel >= _options.ExperimentalLogging.MinimumLogLevel;
```
jamescrosswell, line 34: What happens if `_options.ExperimentalLogging.MinimumLogLevel` is `LogLevel.None`? Should logs be disabled then too?

test/Sentry.Extensions.Logging.Tests/SentryStructuredLoggerTests.cs
```diff
@@ -0,0 +1,302 @@
...
+        log.TryGetAttribute("microsoft.extensions.logging.event.name", out object? _).Should().BeFalse();
+    }
+
+    [Theory]
+    [InlineData(true, true, LogLevel.Warning, LogLevel.Warning, true)]
+    [InlineData(false, true, LogLevel.Warning, LogLevel.Warning, false)]
+    [InlineData(true, false, LogLevel.Warning, LogLevel.Warning, false)]
+    [InlineData(true, true, LogLevel.Information, LogLevel.Warning, true)]
+    [InlineData(true, true, LogLevel.Error, LogLevel.Warning, false)]
+    public void IsEnabled_HubOptionsMinimumLogLevel_Returns(bool isHubEnabled, bool isLogsEnabled, LogLevel minimumLogLevel, LogLevel actualLogLevel, bool expectedIsEnabled)
```
jamescrosswell, line 253: I think we should add a test here to be explicit about what behaviour we expect when the `minimumLogLevel` is `LogLevel.None`.

From the [docs for LogLevel](https://learn.microsoft.com/en-us/dotnet/api/microsoft.extensions.logging.loglevel?view=net-9.0-pp&viewFallbackFrom=net-9.0), my expectation would be that logging would basically be disabled in that case.

src/Sentry.Extensions.Logging/BindableSentryLoggingOptions.cs
Flash0ver: no ... I was a bit overzealous ...
removed all `ExperimentalAttribute` from `internal` types and members

src/Sentry.Extensions.Logging/BindableSentryLoggingOptions.cs
Flash0ver: indeed ...
removed all `ExperimentalAttribute` from `internal` types and members

src/Sentry.Extensions.Logging/SentryStructuredLogger.cs
Flash0ver: Wouldn't we then, when both Loggers (existing `SentryLogger` and new `SentryStructuredLogger`) are enabled, push two scopes?

Or ... is this an indicator that we should move the logic of the new `SentryStructuredLogger` into the existing `SentryLogger` ... just as the existing `SentryLogger` is handling both _Events_ and _Breadcrumbs_.

Also considering
https://github.com/getsentry/sentry-dotnet/blob/b84faa879b5e383f9c07fe887b33763d78448c2c/src/Sentry.Extensions.Logging/SentryLoggingOptionsSetup.cs#L8-L12
may be another indicator that the new Sentry Logs should also be handled in the existing `SentryLogger` infrastructure ... what do you think?

jamescrosswell: > Wouldn't we then, when both Loggers (existing SentryLogger and new SentryStructuredLogger) are enabled, push two scopes?

That's a good point... and we wouldn't want to do that. So we need a way to de-dupe. 

> should move the logic of the new SentryStructuredLogger into the existing SentryLogger

I don't think the existing SentryLogger supports batching right? Could they (practically) be merged?

> internal class SentryLoggingOptionsSetup : IConfigureOptions<SentryLoggingOptions> 

Are you thinking the options for the structured logger would also be configured in a section like:
```json
  "Logging": {
      // your config values
  }
```

I guess it would be possible to have two separate subsections under there (as long as there's minimal/no overlap)... 🤔

Flash0ver: > I don't think the existing SentryLogger supports batching right? Could they (practically) be merged?

Yes. As the new (now about to be merged) `SentryStructuredLogger` just creates an instance of `SentryLog`, and then calls into `_hub.Logger.CaptureLog(log);`, which - if enabled - uses the `DefaultSentryStructuredLogger` which does the buffering and batching.
So the `SentryLogger` wouldn't know anything about buffering/batching, but just forwarding to the Structured-Logger that the SDK/Hub provides ... which I suppose would be consumed in a similar fashion by `sentry-unity` as well.

So move the new logic into the existing Logger is quite straight forward ...
but ...

> Are you thinking the options for the structured logger would also be configured in a section like:

We would only have a single `[ProviderAlias("Sentry")]` to have `LoggerFilterOptions` for.
But on the other hand, I was about to duplicate the `SentryLoggingOptions.Filters` (via `SentryLoggingOptionsExtensions.AddLogEntryFilter`) for the new `SentryStructuredLogger` as well.

So perhaps with our current infrastructure, and Configuration/Options, and `[ProviderAlias("")]` setup,
not adding a new `ILoggerProvider` and `ILogger` is the way to go for this minor release.

Discussed in DMs:
Refactor the existing `SentryLogger` to have 3 Log-Processors, one for Breadcrumbs, one for Events, and one for Structured-Logs.
Thanks @jamescrosswell

src/Sentry.Extensions.Logging/SentryLoggingOptions.cs
```diff
@@ -48,4 +50,39 @@ public class SentryLoggingOptions : SentryOptions
...
+        /// <para>This API is experimental and it may change in the future.</para>
+        /// </summary>
+        /// <remarks>
+        /// Logs with this level or higher will be stored as <see cref="SentryLog"/>.
+        /// </remarks>
+        /// <value>
+        /// The minimum log level.
+        /// </value>
+        public LogLevel MinimumLogLevel { get; set; } = LogLevel.Trace;
```
Flash0ver, line 86: suggestion: rename to `MinimumStructuredLogLevel`

also ensure XML docs mention `"structured"` as well

Flash0ver, line 86: consider: removing this option altogether

src/Sentry.Extensions.Logging/LogLevelExtensions.cs
Flash0ver: potentially constructed scenario, but possible
```CSharp
logger.Log((LogLevel)-1, "message");
logger.Log((LogLevel)7, "message");
```

What should happen?
Well .. these paths are unreachable, since we do an `ILogger.IsEnabled` check before.
Let's do a `Debug.Fail` in these cases.

src/Sentry.Extensions.Logging/SentryLoggingOptions.cs
Flash0ver: will remove in upcoming changeset

PR #4204, opened 2025-05-20
Revert W3C traceparent support to tracing client and utils (#4084)

### LINKED ISSUE #4185: Remove `traceparent` propagation (per default)
Support for W3C `traceparent` header is removed again from other SDKs, too.
It seems to do more harm than good: https://blog.thms.uk/2025/05/openai-sucks

See also getsentry/sentry-docs#13686
I don't think we added any .NET specific docs yet

Was originally added via #4084
Was initially published in [Sentry 5.6.0](https://www.nuget.org/packages/Sentry/5.6.0)

Alternatively, we could make this an explicit opt-in instead, where the default is disabled.

jamescrosswell:
This reverts commit 1ebdb8f3e8698991416f4179e0f9d4582652b3da.

Resolves #4185:
- https://github.com/getsentry/sentry-dotnet/issues/4185


bruno-garcia (approved):
it's really a shame we are reverting this. could be an opt in at least but not worth arguing over

PR #4212, opened 2025-05-22
UnsampledTransactions to reduce memory pressure

### LINKED ISSUE #3636: Sampling based `Transaction.NoOp` to easy memory pressure
## What

Came up in https://github.com/getsentry/sentry-dotnet/pull/3581#issuecomment-2378870314
It does not matter whether a transaction gets sampled or not, they still require the same amount of resources - i.e. memory.

## How

We could expand on this on the NoOp. Instead of a `NoOpTransaction` singleton we could pass around something like a `SampledTransaction` that are functionally NoOp but hold the data required to provide a client report - i.e. span count. I guess the same could work for spans? That way we'd get rid of all the data we know will get discarded in the end anyway.

## Context

We introduced the NoOp transaction with the SDK disabled here: https://github.com/getsentry/sentry-dotnet/pull/3581/files and we could expand on this.

When starting a `transaction` it either has its sampling decision already set or it goes through the `TracesSampler` (if provided)

https://github.com/getsentry/sentry-dotnet/blob/12c7d7ca046ab985b0267f43c5fd35f2caad108c/src/Sentry/Internal/Hub.cs#L149-L153

or it gets it's sampling state set randomly

https://github.com/getsentry/sentry-dotnet/blob/12c7d7ca046ab985b0267f43c5fd35f2caad108c/src/Sentry/Internal/Hub.cs#L162

So we know right from the beginning if a transaction ends up in a client report or not.

jamescrosswell:
Added a light weight `UnsampledTransaction` class to be used when transactions are not sampled. The aim is to reduce memory pressure when sampling less than 100% of transactions.

Resolves #3636:
- https://github.com/getsentry/sentry-dotnet/issues/3636

Replaces #3972:
- https://github.com/getsentry/sentry-dotnet/pull/3972


src/Sentry/Internal/UnsampledTransaction.cs
```diff
@@ -0,0 +1,103 @@
...
+/// <summary>
+/// We know already, when starting a transaction, whether it's going to be sampled or not. When it's not sampled, we can
+/// avoid lots of unecessary processing. The only thing we need to track is the number of spans that would have been
+/// created (the client reports detailing discarded events includes this detail).
+/// </summary>
+internal sealed class UnsampledTransaction : NoOpTransaction
+{
+    // Although it's a little bit wasteful to create separate individual class instances here when all we're going to
+    // report to sentry is the span count (in the client report), SDK users may refer to things like
+    // `ITransaction.Spans.Count`, so we create an actual collection
```
bruno-garcia, line 14: what's the use case for users to get count of spans from transactions?

the goal is to deprecate transactions in the long term. Spans are expected to stream out of the SDK
Anything done absed on span count is going to break so I wouldn't optimzie for that

src/Sentry/Internal/UnsampledTransaction.cs
```diff
@@ -0,0 +1,103 @@
+using Sentry.Extensibility;
+
+namespace Sentry.Internal;
+
+/// <summary>
+/// We know already, when starting a transaction, whether it's going to be sampled or not. When it's not sampled, we can
+/// avoid lots of unecessary processing. The only thing we need to track is the number of spans that would have been
+/// created (the client reports detailing discarded events includes this detail).
+/// </summary>
+internal sealed class UnsampledTransaction : NoOpTransaction
```
bruno-garcia, line 10: I was hoping we'd get away without having to create an instance for each transaction but I also dunno how we'd deal with trace propagation in that case. And since most of the allocation are on the spans anyway this prob going to shave off enough GC pressure

bruno-garcia (approved):
I couldn't spot a test that validates we're still sending the trace-id downstream from http requests for example. That's one concern I have, if we have an unsampled transaction, the trace id should still be used to link all data leaving the SDK (errors, user feedback, replay) as well as all outgoing HTTP request should have that trace id.

src/Sentry/Internal/UnsampledTransaction.cs
jamescrosswell: It's for the client report. At the moment the sampling is done at the transaction level... so if a transaction gets sampled out we're sampling out the transaction + it's tree of child spans. We report the number of spans we sampled out in the client report.

jamescrosswell:
> I couldn't spot a test that validates we're still sending the trace-id downstream from http requests for example. That's one concern I have, if we have an unsampled transaction, the trace id should still be used to link all data leaving the SDK

Good call. I don't actually think that is working at the moment as `GetSpan` currently returns `null` for an `UnsampledTransaction`, and it's used here:
https://github.com/getsentry/sentry-dotnet/blob/6faf0ccc33b1bc717a4cd44267b2bdbbad4724b0/src/Sentry/Internal/Hub.cs#L218-L223

All add some tests and then make sure they all pass before merging.

PR #4213, opened 2025-05-23
MauiAppBuilderExtensions Benchmarks - DO NOT MERGE

jamescrosswell:
Not to be merged... this is just to investigate the performance implications of various different techniques for initialising the Sentry.Maui integration. 

Third round of benchmarks, I realised much of what we wanted to measure was being hidden in the `[GlobalSetup]` for the Benchmark test. I moved all of this into the test itself (so that it would be properly measured) 

Results of the latest benchmarks:
| Method           | ResolveOptionsWithServiceProvider | Mean     | Error    | StdDev   | Median   | Gen0    | Gen1   | Allocated |
|----------------- |---------------------------------- |---------:|---------:|---------:|---------:|--------:|-------:|----------:|
| 'Build MAUI App' | Directly                          | 494.3 us | 19.62 us | 53.38 us | 476.0 us | 12.6953 | 2.9297 | 105.13 KB |                                                     
| 'Build MAUI App' | ServiceProvider                   | 488.1 us |  8.56 us | 12.55 us | 486.6 us | 15.6250 | 3.9063 | 129.52 KB |
| 'Build MAUI App' | InvokeConfigOptions               | 499.5 us |  9.93 us | 22.21 us | 501.9 us | 12.6953 | 3.9063 | 110.23 KB |

This time round, the winner actually appears to be the existing code (as surprising as that may be).

Overall, the difference between all 3 solutions is less than 12 millionths of a second (running on a Macbook with an M2 processor). Even assuming we're running on a device that is 1000 times slower, that amounts to approximately ~10 ms.

### DIFF
--- a/benchmarks/Sentry.Benchmarks/BenchmarkDotNet.Artifacts/results/MvvmBenchmarks-report-github.md
+++ b/benchmarks/Sentry.Benchmarks/BenchmarkDotNet.Artifacts/results/MvvmBenchmarks-report-github.md
@@ -1,0 +1,15 @@
+```
+
+BenchmarkDotNet v0.13.12, macOS 15.4.1 (24E263) [Darwin 24.4.0]
+Apple M2 Max, 1 CPU, 12 logical and 12 physical cores
+.NET SDK 9.0.203
+  [Host]     : .NET 9.0.4 (9.0.425.16305), Arm64 RyuJIT AdvSIMD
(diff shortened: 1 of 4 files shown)


jamescrosswell:
Permanently resolved in #4219:
- https://github.com/getsentry/sentry-dotnet/pull/4219

PR #4221, opened 2025-05-27
Fixes symbolication for net9.0-android applications in Release config

### LINKED ISSUE #4209: MAUI Android Line Numbers & File Names Don't Appear in .NET 9 (But They Do in .NET 8)
### Package

Sentry.Maui

### .NET Flavor

.NET

### .NET Version

9.0.203

### OS

Android

### OS Version

Android 13

### Development Environment

Visual Studio v17.x

### SDK Version

5.7.0

_No response_

### Workload Versions

| Installed Workload Id | Manifest Version | Installation Source |
| --- | --- | --- |
android | 35.0.61/9.0.100 | SDK 9.0.200, VS 17.13.35931.197
aspire | 8.2.2/8.0.100 | SDK 9.0.200, VS 17.13.35931.197
ios | 18.4.9288/9.0.100 | SDK 9.0.200, VS 17.13.35931.197
maccatalyst | 18.4.9288/9.0.100 | SDK 9.0.200, VS 17.13.35931.197
maui-windows | 9.0.51/9.0.100 | SDK 9.0.200, VS 17.13.35931.197
wasm-tools | 9.0.4/9.0.100 | SDK 9.0.200, VS 17.13.35931.197
wasm-tools-net8 | 9.0.4/9.0.100 | SDK 9.0.200 

options.Dsn = "MY-SENTRY-DSN";

### Steps to Reproduce

1. Create a new .NET MAUI App Project in Visual Studio. Be sure to select .NET 9.
2. Install `Sentry.Maui` version `5.7.0`.
3. Add the block of code for `UseSentry()` above to add the Dsn to startup options.
4. In the sample app, open up `MainPage.xaml.cs` and throw an exception in the `OnCounterClicked()` method, for example, just so that can be logged to Sentry.
5. In the sample app csproj file, add the following so that debug symbols are created and uploaded to Sentry automatically during the build. Replace SentryOrg, SentryProject, and SentryAuthToken accordingly:
```
<PropertyGroup Condition="'$(Configuration)' == 'Release'">
  <SentryOrg>MY-SENTRY-ORG-NAME</SentryOrg>
  <SentryProject>MY-SENTRY-PROJECT-NAME</SentryProject>
  <SentryAuthToken>MY-SENTRY-AUTH-TOKEN</SentryAuthToken>
  <SentryUploadSymbols>true</SentryUploadSymbols>
</PropertyGroup>
```

6. Build for Release, targeting Android.
7. Verify that debug symbols are uploaded to Sentry for your project in the Outputs section of Visual Studio.
8. Run the sample app without a debugger, targeting Android.
9. Click/tap on the `Click me` button to increment the counter.
10. Observe that the sample app should crash. Reopen the app to ensure that the crash details are sent up to Sentry.

### Expected Result

An unhandled exception should be sent to Sentry and when viewing the details of the associated event, the stack trace should include line numbers and file names.

### Actual Result

An unhandled exception is sent to Sentry, but when viewing the details of the associated event, the stack trace does not include line numbers and file names.

NOTE: If you repeat the above steps but use a project that targets .NET 8, line numbers and file names DO appear in the stack trace on Sentry.

jamescrosswell:
Resolves #4209:
- https://github.com/getsentry/sentry-dotnet/issues/4209#issuecomment-2911168352

### Testing

I ran into some challenges creating some automated tests for this. I think we should release this fix ASAP then and I've created #4227 to address the tests:
- https://github.com/getsentry/sentry-dotnet/pull/4227

## References
- [bundletool](https://developer.android.com/tools/bundletool#download) (also has a [GitHub Repo](https://github.com/google/bundletool))
- [dotnet/android Assembly Stores docs](https://github.com/dotnet/android/blob/main/Documentation/project-docs/AssemblyStores.md)
- [dotnet/Android Build Targets](https://github.com/dotnet/android/blob/288f107171a8ac0c9c62dc47af5632702598aaa5/src/Xamarin.Android.Build.Tasks/Xamarin.Android.Common.targets#L2968-L2972)


src/Sentry.Android.AssemblyReader/V2/MonoAndroidHelper.Basic.cs
```diff
@@ -89,4 +89,29 @@ public static string MakeZipArchivePath(string part1, ICollection<string>? pathP
...
+
+    /// <summary>
+    /// When an AAB file is deployed, the APK is split into multiple APKs so our modules can end up in a companion
+    /// architecture specific APK like split_config.arm64_v8a.apk. This method returns the path to that split_config APK
+    /// if it exists... otherwise we just return the original archive path.
+    /// </summary>
+    internal static string GetArchivePathForAbi(this string archivePath, string abi, DebugLogger? logger)
+    {
+        var basePath = Path.GetDirectoryName(archivePath) ?? string.Empty;
+        var abiPart = abi.Replace("-", "_");
```
bruno-garcia, line 113: I wonder if we can avoid this by having `supportedAbis` be provided in the right format

jamescrosswell: The `abi.Replace("-", "_")` you mean? Not really... most of the time it is used with a hyphen. When building the file names for the `split_config` APKs though it gets replaced with an underscore (I'm sure there's a reason... not sure what it is though).

jamescrosswell:
> Should we test this on Symbol Collector Android?

That might give a bit of reassurance yes. I've only tested it manually on the Sentry.Samples.Maui project so far. 

I'm still not sure whether the file we're after is always going to be called `split_config.{abiPart}.apk` for that matter - when I used the bundletool to try to engineer a test case, it called the file `base.{abiPart}.apk`... So having done a bit of headscratching (overnight), I think some more testing wouldn't hurt.

Another thing that should be tested is the scenario where someone builds with AssemblyStores disabled. I'm not sure who would do that or why (it just hurts performance... the only reason I can think to do it would be while debugging, to make sure the expected set of assemblies was bundled) but it's a valid configuration and Sentry should work with it.

Issue #4116, opened 2025-04-15
clang error building native AoT application with Sentry.AspNetCore on Linux


martincostello:
### Package

Sentry.AspNetCore

### .NET Flavor

.NET

### .NET Version

9.0.203

### OS

Linux

### OS Version

Ubuntu 24.04.2

### Development Environment

Other

### SDK Version

5.5.1

### Workload Versions

N/A

### UseSentry or SentrySdk.Init call

N/A

### Steps to Reproduce

1. Clone https://github.com/martincostello/website/pull/2504/commits/f36510dfe09d2cc1f3af69950163fe54c285576a on Ubuntu (or in a GitHub Actions runner)
2. Run `build.ps1` in the root of the repository

### Expected Result

The application builds successfully.

### Actual Result

```text
Generating native code
  /usr/bin/ld.bfd: cannot find -lcurl: No such file or directory
clang : error : linker command failed with exit code 1 (use -v to see invocation) [/home/runner/work/website/website/src/Website/Website.csproj]
[...]

PR #4222, opened 2025-05-27
feat: replace libcurl with .NET HttpClient for sentry-native

jpnurmi:
Install a custom function transport for sentry-native, and send native envelopes with the .NET HttpClient instead of relying on the default libcurl-based HTTP transport in sentry-native. This way, sentry-native can be built with `-DSENTRY_TRANSPORT=none` to eliminate the libcurl dependency, which is problematic for minimal/chiseled .NET Native AOT containers.

The behavior remains virtually identical compared to the current native transport with libcurl. The Native AOT integration test that [captures a native envelope on restart](https://github.com/getsentry/sentry-dotnet/blob/9beef00b325bd79c2cf43829f069142461aec113/integration-test/runtime.Tests.ps1#L163-L176) passes without any changes. The only difference is that native envelopes are sent with .NET HttpClient instead of libcurl. There's no native -> .NET envelope conversion nor hooks being called.

Close: #4116

### DIFF
--- a/src/Sentry/Platforms/Native/CFunctions.cs
+++ b/src/Sentry/Platforms/Native/CFunctions.cs
@@ -139,6 +139,14 @@ internal static class C
             }
         }
 
+        unsafe
+        {
+            var cTransport = sentry_transport_new(&nativeTransport);
+            sentry_transport_set_state(cTransport, GCHandle.ToIntPtr(GCHandle.Alloc(options)));
+            sentry_transport_set_free_func(cTransport, &nativeTransportFree);
+            sentry_options_set_transport(cOptions, cTransport);
+        }
+
         options.DiagnosticLogger?.LogDebug("Initializing sentry native");
         return 0 == sentry_init(cOptions);
     }
@@ -363,6 +371,61 @@ internal static class C
 
     [DllImport("sentry-native")]
     private static extern void sentry_options_set_auto_session_tracking(IntPtr options, int debug);
+
+    [DllImport("sentry-native")]
+    private static extern void sentry_options_set_transport(IntPtr options, IntPtr transport);
+
+    [DllImport("sentry-native")]
+    private static extern unsafe IntPtr sentry_transport_new(delegate* unmanaged/*[Cdecl]*/<IntPtr, IntPtr, void> sendFunc);
+
+    [DllImport("sentry-native")]
+    private static extern void sentry_transport_set_state(IntPtr transport, IntPtr state);
+
+    [DllImport("sentry-native")]
+    private static extern unsafe void sentry_transport_set_free_func(IntPtr transport, delegate* unmanaged/*[Cdecl]*/<IntPtr, void> freeFunc);
+
+    [DllImport("sentry-native")]
+    private static extern IntPtr sentry_envelope_serialize(IntPtr envelope, out UIntPtr sizeOut);
+
+    [DllImport("sentry-native")]
+    private static extern void sentry_envelope_free(IntPtr envelope);
+
+    [DllImport("sentry-native")]
+    private static extern void sentry_free(IntPtr ptr);
+
+    [UnmanagedCallersOnly]
+    private static void nativeTransport(IntPtr envelope, IntPtr state)
+    {
+        try
+        {
+            var options = GCHandle.FromIntPtr(state).Target as SentryOptions;
+            if (options is not null)
+            {
+                var data = sentry_envelope_serialize(envelope, out var size);
+                var content = new StringContent(Marshal.PtrToStringAnsi(data, (int)size));
+                sentry_free(data);
+
+                using var client = options.GetHttpClient();
+                using var request = options.CreateHttpRequest(content);
+                client.SendAsync(request).GetAwaiter().GetResult();
+            }
+        }
+        catch
+        {
+            // never allow an exception back to native code - it would crash the app
+        }
+        finally
+        {
+            sentry_envelope_free(envelope);
+        }
+    }
+
+    [UnmanagedCallersOnly]
+    private static void nativeTransportFree(IntPtr state)
+    {
+        var handle = GCHandle.FromIntPtr(state);
+        handle.Free();
+    }
 
     [DllImport("sentry-native")]
     private static extern unsafe void sentry_options_set_logger(IntPtr options, delegate* unmanaged/*[Cdecl]*/<int, IntPtr, IntPtr, IntPtr, void> logger, IntPtr userData);
--- a/src/Sentry/Platforms/Native/CFunctions.cs
+++ b/src/Sentry/Platforms/Native/CFunctions.cs
@@ -396,9 +396,9 @@ internal static class C
     [UnmanagedCallersOnly]
     private static void nativeTransport(IntPtr envelope, IntPtr state)
     {
+        var options = GCHandle.FromIntPtr(state).Target as SentryOptions;
         try
         {
-            var options = GCHandle.FromIntPtr(state).Target as SentryOptions;
             if (options is not null)
             {
                 var data = sentry_envelope_serialize(envelope, out var size);
@@ -409,9 +409,10 @@ internal static class C
                 client.SendAsync(request).GetAwaiter().GetResult();
             }
         }
-        catch
+        catch (Exception e)
         {
             // never allow an exception back to native code - it would crash the app
+            options?.DiagnosticLogger?.LogError(e, "Exception in native transport callback. The native envelope will not be sent.");
         }
         finally
         {
--- a/src/Sentry/Platforms/Native/CFunctions.cs
+++ b/src/Sentry/Platforms/Native/CFunctions.cs
@@ -391,7 +391,7 @@ internal static class C
     private static extern void sentry_envelope_free(IntPtr envelope);
 
     [DllImport("sentry-native")]
-    private static extern void sentry_free(IntPtr ptr);
+    internal static extern void sentry_free(IntPtr ptr);
 
     [UnmanagedCallersOnly]
     private static void nativeTransport(IntPtr envelope, IntPtr state)
@@ -402,8 +402,7 @@ internal static class C
             if (options is not null)
             {
                 var data = sentry_envelope_serialize(envelope, out var size);
-                var content = new StringContent(Marshal.PtrToStringAnsi(data, (int)size));
-                sentry_free(data);
+                using var content = new UnmanagedHttpContent(data, (int)size, options.DiagnosticLogger);
 
                 using var client = options.GetHttpClient();
                 using var request = options.CreateHttpRequest(content);
--- a/src/Sentry/Platforms/Native/CFunctions.cs
+++ b/src/Sentry/Platforms/Native/CFunctions.cs
@@ -376,13 +376,13 @@ internal static class C
     private static extern void sentry_options_set_transport(IntPtr options, IntPtr transport);
 
     [DllImport("sentry-native")]
-    private static extern unsafe IntPtr sentry_transport_new(delegate* unmanaged/*[Cdecl]*/<IntPtr, IntPtr, void> sendFunc);
+    private static extern unsafe IntPtr sentry_transport_new(delegate* unmanaged<IntPtr, IntPtr, void> sendFunc);
 
     [DllImport("sentry-native")]
     private static extern void sentry_transport_set_state(IntPtr transport, IntPtr state);
 
     [DllImport("sentry-native")]
-    private static extern unsafe void sentry_transport_set_free_func(IntPtr transport, delegate* unmanaged/*[Cdecl]*/<IntPtr, void> freeFunc);
+    private static extern unsafe void sentry_transport_set_free_func(IntPtr transport, delegate* unmanaged<IntPtr, void> freeFunc);
 
     [DllImport("sentry-native")]
     private static extern IntPtr sentry_envelope_serialize(IntPtr envelope, out UIntPtr sizeOut);
@@ -406,8 +406,17 @@ internal static class C
 
                 using var client = options.GetHttpClient();
                 using var request = options.CreateHttpRequest(content);
-                client.SendAsync(request).GetAwaiter().GetResult();
+#if NET5_0_OR_GREATER
+                var response = client.Send(request);
+#else
+                var response = client.SendAsync(request).GetAwaiter().GetResult();
+#endif
+                response.EnsureSuccessStatusCode();
             }
+        }
+        catch (HttpRequestException e)
+        {
+            options?.DiagnosticLogger?.LogError(e, "Failed to send native envelope.");
         }
         catch (Exception e)
         {
--- a/src/Sentry/Platforms/Native/UnmanagedHttpContent.cs
+++ b/src/Sentry/Platforms/Native/UnmanagedHttpContent.cs
@@ -4,7 +4,7 @@ namespace Sentry.Native;
 
 internal sealed class UnmanagedHttpContent : SerializableHttpContent
 {
-    private readonly IntPtr _content;
+    private IntPtr _content;
     private readonly int _length = 0;
     private readonly IDiagnosticLogger? _logger;
 
@@ -52,6 +52,7 @@ internal sealed class UnmanagedHttpContent : SerializableHttpContent
     protected override void Dispose(bool disposing)
     {
         C.sentry_free(_content);
+        _content = IntPtr.Zero;
         base.Dispose(disposing);
     }
 
--- a/src/Sentry/Platforms/Native/UnmanagedHttpContent.cs
+++ b/src/Sentry/Platforms/Native/UnmanagedHttpContent.cs
@@ -22,7 +22,7 @@ internal sealed class UnmanagedHttpContent : SerializableHttpContent
 
     protected override async Task SerializeToStreamAsync(Stream stream, TransportContext? context)
     {
-        ObjectDisposedException.ThrowIf(_content == IntPtr.Zero, this);
+        ThrowIfObjectDisposed();
         try
         {
             using var unmanagedStream = CreateStream();
@@ -37,7 +37,7 @@ internal sealed class UnmanagedHttpContent : SerializableHttpContent
 
     protected override void SerializeToStream(Stream stream, TransportContext? context, CancellationToken cancellationToken)
     {
-        ObjectDisposedException.ThrowIf(_content == IntPtr.Zero, this);
+        ThrowIfObjectDisposed();
         try
         {
             using var unmanagedStream = CreateStream();
@@ -66,5 +66,13 @@ internal sealed class UnmanagedHttpContent : SerializableHttpContent
     private unsafe UnmanagedMemoryStream CreateStream()
     {
         return new UnmanagedMemoryStream((byte*)_content.ToPointer(), _length);
+    }
+
+    private void ThrowIfObjectDisposed()
+    {
+        if (_content == IntPtr.Zero)
+        {
+            throw new ObjectDisposedException(GetType().FullName);
+        }
     }
 }
--- a/src/Sentry/Platforms/Native/UnmanagedHttpContent.cs
+++ b/src/Sentry/Platforms/Native/UnmanagedHttpContent.cs
@@ -22,6 +22,7 @@ internal sealed class UnmanagedHttpContent : SerializableHttpContent
 
     protected override async Task SerializeToStreamAsync(Stream stream, TransportContext? context)
     {
+        ObjectDisposedException.ThrowIf(_content == IntPtr.Zero, this);
         try
         {
             using var unmanagedStream = CreateStream();
@@ -36,6 +37,7 @@ internal sealed class UnmanagedHttpContent : SerializableHttpContent
 
     protected override void SerializeToStream(Stream stream, TransportContext? context, CancellationToken cancellationToken)
     {
+        ObjectDisposedException.ThrowIf(_content == IntPtr.Zero, this);
         try
         {
             using var unmanagedStream = CreateStream();
--- a/src/Sentry/Platforms/Native/UnmanagedHttpContent.cs
+++ b/src/Sentry/Platforms/Native/UnmanagedHttpContent.cs
@@ -15,6 +15,11 @@ internal sealed class UnmanagedHttpContent : SerializableHttpContent
         _logger = logger;
     }
 
+    ~UnmanagedHttpContent()
+    {
+        Dispose(false);
+    }
+
     protected override async Task SerializeToStreamAsync(Stream stream, TransportContext? context)
     {
         try
@@ -51,8 +56,8 @@ internal sealed class UnmanagedHttpContent : SerializableHttpContent
 
     protected override void Dispose(bool disposing)
     {
-        C.sentry_free(_content);
-        _content = IntPtr.Zero;
+        IntPtr content = Interlocked.Exchange(ref _content, IntPtr.Zero);
+        C.sentry_free(content);
         base.Dispose(disposing);
     }
 
--- a/src/Sentry/Platforms/Native/UnmanagedHttpContent.cs
+++ b/src/Sentry/Platforms/Native/UnmanagedHttpContent.cs
@@ -52,7 +52,7 @@ internal sealed class UnmanagedHttpContent : SerializableHttpContent
     protected override bool TryComputeLength(out long length)
     {
         length = _length;
-        return false;
+        return true;
     }
 
     protected override void Dispose(bool disposing)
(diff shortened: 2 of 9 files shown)


src/Sentry/Platforms/Native/CFunctions.cs
```diff
@@ -139,6 +139,14 @@ public static bool Init(SentryOptions options)
             }
         }
 
+        unsafe
+        {
+            var cTransport = sentry_transport_new(&nativeTransport);
```
Flash0ver, line 144: question: not sure if I get that right

My native is ... well, non-native 😉 ... I was wondering if I understand it correctly ... also not sure if my concern about changed addresses after a compacting GC are applicable:
1. we create a new native Transport instance via `sentry-native`, which uses the C# `nativeTransport` as a callback when envelopes are sent
2. we set the state of the native Transport instance, which is passed to both the `nativeTransport` and `nativeTransportFree` callbacks when invoked
3. we set a callback of the native Transport that is invoked via `sentry_transport_free`
   - when is `sentry_transport_free` called?
4. we set the Transport of the native `sentry_options_t` to the native Transport instance created before

Do I get the (somewhat) right?

Also, are there concerns about managed addresses that may change after a compacting GC?
To be fair, I haven't used `GCHandle`/`SafeHandle` before. And I have super limited know-how that I heard/read in the context of new generic handles (normal `GCHandle<T>`, `PinnedGCHandle<T>` and `WeakGCHandle<T>`) being added to .NET 10 (see https://github.com/dotnet/apireviews/blob/main/2025/01-07-quick-reviews/README.md).
Could we run into an issue where the address of the managed `SentryOptions options` changes during runtime after `Sentry.Native.C.Init(SentryOptions)` has been called?
Do we need to think about pinning our managed `SentryOptions` so that the `IntPtr` is always valid (i.e. via a pinned GC-Handle)?

src/Sentry/Platforms/Native/CFunctions.cs
```diff
@@ -364,6 +372,60 @@ internal struct sentry_value_t
...
+    private static extern IntPtr sentry_envelope_serialize(IntPtr envelope, out UIntPtr sizeOut);
+
+    [DllImport("sentry-native")]
+    private static extern void sentry_envelope_free(IntPtr envelope);
+
+    [DllImport("sentry-native")]
+    internal static extern void sentry_free(IntPtr ptr);
+
+    [UnmanagedCallersOnly]
+    private static void nativeTransport(IntPtr envelope, IntPtr state)
```
Flash0ver, line 397: question: Is the `envelope`'s address passed here a native unmanaged object?

src/Sentry/Platforms/Native/CFunctions.cs
```diff
@@ -364,6 +372,60 @@ internal struct sentry_value_t
...
+        {
+            var options = GCHandle.FromIntPtr(state).Target as SentryOptions;
+            if (options is not null)
+            {
+                var data = sentry_envelope_serialize(envelope, out var size);
+                using var content = new UnmanagedHttpContent(data, (int)size, options.DiagnosticLogger);
+
+                using var client = options.GetHttpClient();
+                using var request = options.CreateHttpRequest(content);
+                client.SendAsync(request).GetAwaiter().GetResult();
```
Flash0ver: note: `net5.0` added a synchronous `HttpClient.Send` method group

We could think about `#if` def to use the synchronous version for compatible target frameworks.

See https://source.dot.net/#q=System.Net.Http.HttpClient.Send

src/Sentry/Platforms/Native/CFunctions.cs
```diff
@@ -364,6 +372,60 @@ internal struct sentry_value_t
...
+            }
+        }
+        catch
+        {
+            // never allow an exception back to native code - it would crash the app
```
Flash0ver, line 423: question: Could it be helpful to add a diagnostic log message in this case?

src/Sentry/Platforms/Native/UnmanagedHttpContent.cs
```diff
@@ -0,0 +1,63 @@
+using Sentry.Extensibility;
+
+namespace Sentry.Native;
+
+internal class UnmanagedHttpContent : SerializableHttpContent
```
Flash0ver: nitpick: We could mark this type `sealed`

src/Sentry/Platforms/Native/UnmanagedHttpContent.cs
```diff
@@ -0,0 +1,63 @@
...
+    }
+
+    protected override Task SerializeToStreamAsync(Stream stream, TransportContext? context)
+    {
+        try
+        {
+            unsafe
+            {
+                using var unmanagedStream = new UnmanagedMemoryStream((byte*)_content.ToPointer(), _length);
+                return unmanagedStream.CopyToAsync(stream);
```
Flash0ver: attention: `CopyToAsync` is not `await`'d. Asynchronous exceptions are not caught.

Is this intended?
Should this method be `async`?

src/Sentry/Platforms/Native/UnmanagedHttpContent.cs
```diff
@@ -0,0 +1,63 @@
...
+        _logger = logger;
+    }
+
+    protected override Task SerializeToStreamAsync(Stream stream, TransportContext? context)
+    {
+        try
+        {
+            unsafe
+            {
+                using var unmanagedStream = new UnmanagedMemoryStream((byte*)_content.ToPointer(), _length);
```
Flash0ver: question: The address of `_content` points to unmanaged memory, right? So not concerns about pinning?

src/Sentry/Platforms/Native/UnmanagedHttpContent.cs
```diff
@@ -0,0 +1,63 @@
...
+    private readonly IDiagnosticLogger? _logger;
+
+    public UnmanagedHttpContent(IntPtr content, int length, IDiagnosticLogger? logger)
+    {
+        _content = content;
+        _length = length;
+        _logger = logger;
+    }
+
+    protected override Task SerializeToStreamAsync(Stream stream, TransportContext? context)
```
Flash0ver: question: overload

I believe there is also a virtual overload of `SerializeToStreamAsync` that takes a `CancellationToken cancellationToken`.
Would we like to override this overload as well, so that we also support asynchronous cancellation?
(And then have this method call into the added method passing `CancellationToken.None`)

src/Sentry/Platforms/Native/UnmanagedHttpContent.cs
```diff
@@ -0,0 +1,63 @@
...
+        {
+            _logger?.LogError(e, "Failed to serialize unmanaged content into the network stream");
+            throw;
+        }
+    }
+
+    protected override bool TryComputeLength(out long length)
+    {
+        length = _length;
+        return false;
```
Flash0ver: question: return `true`

I would have expected to return `true` here, to indicate that we were able to compute the length.
Or are there scenarios, where the `int length` passed in the constructor is not valid (like `0` or `-1` or similar)?

src/Sentry/Platforms/Native/UnmanagedHttpContent.cs
```diff
@@ -0,0 +1,63 @@
...
+    }
+
+    protected override void Dispose(bool disposing)
+    {
+        C.sentry_free(_content);
```
Flash0ver: question: Would a destructor/finalizer add value here?

I see that in the current usage we do use a `using` declaration, so everything is fine.
Also, it's an internal type so we have full control to use it correctly.

Nevertheless I'm wondering if adding a destructor/finalizer would make sense here ... just to be sure?

src/Sentry/Platforms/Native/CFunctions.cs
jpnurmi: Yes, the respective C-signature is:
```c
void (*send_func)(sentry_envelope_t *envelope, void *state)
```

And the ownership is transferred:
> `send_func`: This function will take ownership of an envelope, and is responsible for freeing it via `sentry_envelope_free`.

https://github.com/getsentry/sentry-native/blob/1fe91217f1ce839f46ef398bdf1f3bf58d99a0da/include/sentry.h#L673C1-L674C60

src/Sentry/Platforms/Native/UnmanagedHttpContent.cs
jpnurmi: It is intended because of `unsafe`:

> Cannot await in an unsafe context [CS4004](https://msdn.microsoft.com/query/roslyn.query?appId=roslyn&k=k(CS4004))

src/Sentry/Platforms/Native/UnmanagedHttpContent.cs
jpnurmi: Correct, this memory was allocated by the native `sentry_envelope_serialize` call and needs to be freed with `sentry_free`, currently done in `Dispose()`.

src/Sentry/Platforms/Native/UnmanagedHttpContent.cs
jpnurmi: Thanks, good catch! My bad while copying `EnvelopeHttpContent` :)

src/Sentry/Platforms/Native/UnmanagedHttpContent.cs
jpnurmi: What about something like this?

```csharp
UnmanagedMemoryStream unmanagedStream;
unsafe
{
    unmanagedStream = new UnmanagedMemoryStream((byte*)_content.ToPointer(), _length);
}
using (unmanagedStream)
{
    await unmanagedStream.CopyToAsync(stream).ConfigureAwait(false);
}
```

src/Sentry/Platforms/Native/CFunctions.cs
jpnurmi: Sounds about right! The native transport is owned by sentry-native and `sentry_transport_free` is automatically called when closing the SDK or replacing the transport.

I have to double-check `Normal` vs. `Pinned` GC handles. My assumption was that `GCHandle` would prevent the managed object from being garbage-collected until the handle is released, but it doesn't sound good if the object is still allowed to be moved around. :fearful:

jpnurmi: Ha, nice - thanks for catching it!

`Normal`
> This handle type represents an opaque handle, **meaning you cannot resolve the address of the pinned object through the handle**. You can use this type to track an object and prevent its collection by the garbage collector. This enumeration member is useful when an unmanaged client holds the only reference, which is undetectable from the garbage collector, to a managed object.

`Pinned`
> This handle type is similar to [Normal](https://learn.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.gchandletype?view=net-9.0#system-runtime-interopservices-gchandletype-normal), but **allows the address of the pinned object to be taken. This prevents the garbage collector from moving the object** and hence undermines the efficiency of the garbage collector. Use the [Free()](https://learn.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.gchandle.free?view=net-9.0#system-runtime-interopservices-gchandle-free) method to free the allocated handle as soon as possible.

jpnurmi:
```
  Unhandled Exception: System.ArgumentException: Object contains references. (Parameter 'value')
     at System.Runtime.InteropServices.GCHandle..ctor(Object, GCHandleType) + 0xaa
     at Sentry.Native.C.Init(SentryOptions) + 0x5ff
     at Sentry.SentrySdk.InitNativeSdk(SentryOptions) + 0x34
     at Sentry.SentrySdk.InitHub(SentryOptions) + 0x1f3
     at Sentry.SentrySdk.Init(Action`1) + 0x3a
     at Program.<Main>$(String[] args) + 0x49
     at console-app!<BaseAddress>+0x559c4c
```

src/Sentry/Platforms/Native/UnmanagedHttpContent.cs
jpnurmi: I was inclined to answer that using a finalizer sounds good to make it bullet-proof, but then I found this in the "Native interoperability best practices" docs:

> ### General guidance
> ❌ AVOID finalizers to manage lifetime of objects that encapsulate unmanaged resources. For more information, see [Implement a Dispose method](https://learn.microsoft.com/en-us/dotnet/standard/garbage-collection/implementing-dispose).

https://learn.microsoft.com/en-us/dotnet/standard/native-interop/best-practices#general-guidance

src/Sentry/Platforms/Native/CFunctions.cs
jpnurmi: The current implementation is the same as the second snippet here:

https://learn.microsoft.com/en-us/dotnet/standard/native-interop/best-practices#keeping-managed-objects-alive

jpnurmi: If I understand correctly, the GC is allowed to move the managed `SentryOptions` object. We're passing a pointer to the GC handle, not the managed object.

src/Sentry/Platforms/Native/CFunctions.cs
jpnurmi: I used the sync API at first for the sake of simplicity during prototyping, but there was a big red warning in the docs that scared me off :sweat_smile: 

> [!WARNING]
> Making HTTP requests is considered network I/O-bound work. A synchronous [HttpClient.Send](https://learn.microsoft.com/en-us/dotnet/api/system.net.http.httpclient.send) method exists, but the recommendation is to use the asynchronous APIs instead, unless you have good reason not to.

https://learn.microsoft.com/en-us/dotnet/fundamentals/networking/http/httpclient#make-an-http-request

src/Sentry/Platforms/Native/CFunctions.cs
Flash0ver: 💡 ... I dramatically misunderstood `GCHandle`
- I thought that `(From|To)IntPtr()` is utilizing the address of the object
- but `(From|To)IntPtr()` is utilizing the address of the `GCHandle` allocated
  - hence the `Free()` to clean the `GCHandle` up, as it's not tracked by the GC, so also no relocation concerns
- and therefore, only when `Pinned` can we get the actual Address of the managed `object`

Thanks for educating me!

src/Sentry/Platforms/Native/CFunctions.cs
jpnurmi: This piece of code executes on the next startup of a previously crashed run, so basically a normal application startup.

How would it be best to structure this code so that it doesn't unnecessarily slow down the startup but also doesn't risk leaking exceptions back to native code?

Technically, it could be a fire-and-forget async request because nothing relies on the request having been delivered, because we have already serialized the envelope and so it can be freed right away.

Flash0ver: Right ... when `async/await` is possible, then Task-based `..Async()` methods should be preferred and the returned "awaitable" (e.g. `Task<T>`) properly be `await`'d, so that the current Thread is not blocked and may do some other work until the operation has completed.

But here we are in a synchronous method, and `Task.GetAwaiter().GetResult()` is essentially doing the same thing: synchronously blocking the current Thread until the operation has completed (successfully, faulted or canceled).

Now I'm wondering if we should make the method `async void`, so that we can actually use `await`, and not synchronously block for completion. Usually `async void` is an Anti-Pattern. Usually you have `async Task` (or returning any "awaitable" type). `async void` has the problem of Exceptions being either uncatchable or unobserved (depending whether on a Background or Foreground Thread). One of the exceptions of the "Async-All-The-Way"-Rule are event handlers, as they are `void`-returning. However, I'm a bit uncertain about the combination with unmanaged callers:
- if the callback throws an exception asynchronously, the unmanaged caller would have no mechanism of catching / observing that exception
- the `async void` callback would complete quasi immediately (at the time of the first await of an "awaitable" type that has not yet completed), so the unmanaged caller has no mechanism of knowing when the asynchronous work has actually been completed
  - this may cause synchronization issues / concerns about thread safety
- also, there may be a slightly different behavior depending on the `SynchronizationContext` in use (or the lack thereof)
  - usually UI-based applications like MAUI, WPF, WinForms (and also classic ASP.NET) provide a `SynchronizationContext`

An alternative to `async/await` could be the use of `Task.ContinueWith()`, but we would end up with similar issues: the caller has no way of observing exceptions, nor observing actual completion of the asynchronous work.

On the other hand, synchronous blocking may be a contributing factor to thread starvation (of the `ThreadPool` in use).

My question:
In this case, is the native layer of `sentry-native` depending on synchronous completion and/or notification about exceptions?
Well ... we do a `try-catch-finally` around the entire code block of the callback, so perhaps exceptions are not a concern in this case ... but the asynchronous completion not being observable may be.

src/Sentry/Platforms/Native/UnmanagedHttpContent.cs
jpnurmi: Supposedly, it needs some `NET5_0_OR_GREATER` guards to retain compatibility with older targets. Do you want to do that in this PR, or is that something that could be done in `SerializableHttpContent` as a separate step?

src/Sentry/Platforms/Native/UnmanagedHttpContent.cs
Flash0ver: TIL ... `error CS4004`... right ... because of lifting local variables to fields of the compiler-generated StateMachine

Flash0ver: The suggestion does solve the issue, but it does look a bit "uncommon" to me.

Hmm ... I'm wondering if it would make sense to extract the `unsafe` part into a separate method:
- a new private instance factory method
  - e.g. `private unsafe UnmanagedMemoryStream CreateStream()`
- and then in this method we would no longer need to use any `unsafe`
- and could refactor the other `override` as well

But I guess at this point it's mostly style ... so I'm fine either way ... however you prefer.

jpnurmi: Nice, thanks!

src/Sentry/Platforms/Native/CFunctions.cs
jpnurmi: sentry-native doesn't have any requirements or expectations besides releasing the envelope once you're done. You are allowed to store it for later, for example.

`async void` + `await SendAsync()` does seem to work in the sense that it makes the request asynchronous, and makes the integration test flaky... It probably exits too early before the request has been fully sent.

src/Sentry/Platforms/Native/UnmanagedHttpContent.cs
Flash0ver: Our usage indeed uses deterministic cleanup via `Dispose`.
My comment was more about "completeness" of the "Dispose Pattern".

Technically, when the constructor fails, `Dispose` on the object is never called, but if the object has a Finalizer it's still added to the Finalize-Queue of the GC.

To be fair, the constructor just simply sets fields, so it cannot fail.
Also, the type is `internal` and we can ensure that we use it correctly, as in ensure that it's `Dispose`'d.

Long story short: we are safe cleaning up our unmanaged resoources.
We could - but don't have to - diligently implement the full "Dispose Pattern".

src/Sentry/Platforms/Native/UnmanagedHttpContent.cs
```diff
@@ -0,0 +1,62 @@
...
+
+    protected override bool TryComputeLength(out long length)
+    {
+        length = _length;
+        return true;
+    }
+
+    protected override void Dispose(bool disposing)
+    {
+        C.sentry_free(_content);
```
Flash0ver: now ... through our discussion in the other comment ... I realized that we are overriding an existing `Dispose(bool)` from the Dispose-Pattern.

Per guideline, it should be safe to call `Dispose` multiple times, where consecutive invocations of `Dispose` should either be No-Op or Side-Effect-less.

What would happen in our case if `Dispose` is called more than once on an instance?

Another guideline is that - via the "Dispose Pattern"
- when `bool disposing` is `true`, both managed and unmanaged resources should be cleaned up, because `Dispose` has been called deterministically either directly or via `using`
- when `bool disposing` is `false`, only unamanged resources should be cleaned up, because is has been called nondeterministically from a Finalizer

I believe this is why I was thinking about a Finalizer in the other comment.

To be fair, this is an `internal` type, and we have control over using and Disposing it correctly.
I am a bit uncertain how much time we should invest to make it technically adhere to the "Dispose Pattern" - or with our case - where the constructor cannot throw an exception and we have code in place that ensures that `Dispose` is always called exactly a single time - good enough is good enough.

Long story short:
- does it make sense to do the unmanaged cleanup only when `bool disposing` is `false`
- does it make sense to make subsequent calls of `Dispose` "safe"
  - or is there even a problem if `C.sentry_free(_content);` is called more than once?

jpnurmi: > or is there even a problem if C.sentry_free(_content); is called more than once?

thanks, good catch again. that would crash unless the pointer is cleared to prevent double-free.

src/Sentry/Platforms/Native/CFunctions.cs
Flash0ver: sum up of our chat:
- for now, we go with a synchronously blocking `client.SendAsync(request).GetAwaiter().GetResult();`, and maybe `client.Send();` for `net5.0` and higher
- since it's only called when a native crash happened before, and we send the crash report during the next startup
- if this proves to be a performance problem, we could follow-up with connecting the already existing `Begin`/`End` native callbacks to be fully asynchronous to no longer be blocking the main UI thread

src/Sentry/Platforms/Native/UnmanagedHttpContent.cs
Flash0ver: Oh ... right ... our `SerializableHttpContent`, that means this also applies to `EnvelopeHttpContent`.

Good call!
Yes. Let's improve that in a follow-up in a separate PR, because `EnvelopeHttpContent` currently does not support asynchronous cancellation either, for `NET5_0_OR_GREATER`.

src/Sentry/Platforms/Native/UnmanagedHttpContent.cs
Flash0ver: chat sum-up:
- we'll make `Dispose` to be invoked multiple times safe via either
  ```CSharp
   IntPtr content = Interlocked.Exchange(ref _content, IntPtr.Zero);
   if (content != IntPtr.Zero)
   {
     C.sentry_free(content);
   }
- or simpler ... if it's safe to `sentry_free` with `IntPtr.Zero`
   ```CSharp
   IntPtr content = Interlocked.Exchange(ref _content, IntPtr.Zero);
   C.sentry_free(content);
   ```

src/Sentry/Platforms/Native/UnmanagedHttpContent.cs
```diff
@@ -0,0 +1,68 @@
...
+    }
+
+    ~UnmanagedHttpContent()
+    {
+        Dispose(false);
+    }
+
+    protected override async Task SerializeToStreamAsync(Stream stream, TransportContext? context)
+    {
+        try
```
Flash0ver, line 26: Oh ... one more thing:
To be on the safe side, we should throw an `ObjectDisposedException` in the `protected override` methods, in case `Dispose` has been called ... in our case when `_content == IntPtr.Zero` ... because we would otherwise get some random memory, perhaps even a `AccessViolationException`.

But again ... the type is `internal` so we can ensure we're using it correctly.
But perhaps to pay it forward to the next developer reading/changing this portion of the code.

We could do this in a follow-up as well ... as our current usage is correct.
Sorry for noticing this so late.

Flash0ver: thanks for humoring me

src/Sentry/Platforms/Native/CFunctions.cs
```diff
@@ -364,6 +372,70 @@ internal struct sentry_value_t
...
+            {
+                var data = sentry_envelope_serialize(envelope, out var size);
+                using var content = new UnmanagedHttpContent(data, (int)size, options.DiagnosticLogger);
+
+                using var client = options.GetHttpClient();
+                using var request = options.CreateHttpRequest(content);
+#if NET5_0_OR_GREATER
+                var response = client.Send(request);
+#else
+                var response = client.SendAsync(request).GetAwaiter().GetResult();
```
bruno-garcia, line 412: @Flash0ver  might want to add a comment here that we only go through this native transport in crashes so we dorm care about blocking and that there's no risk of deadlock since no synchronization context

jpnurmi: Sorry if I was too eager to merge this. I wanted to get rid of the libcurl installation steps for #4247. :)

Issue #4228, opened 2025-05-28
Add overload to prevent delegate allocation for ConfigureScope


KnapSac:
### Problem Statement

I use the following code to configure a scope, capturing a value to be set as a tag on the scope. Unfortunately, this always results in a delegate allocation because of the capture.

```csharp
string tagValue = "value";
SentrySdk.ConfigureScope(scope => scope.SetTag("tag_name", tagValue));
```

### Solution Brainstorm

Perhaps there already is a better way to set a tag on the current scope, but otherwise I'd propose adding the following methods:
```diff
public static partial class SentrySdk
{
+    public static void ConfigureScope<TArg>(Action<Scope, TArg> configureScope, TArg arg);
+    public static Task ConfigureScopeAsync<TArg>(Func<Scope, TArg, Task> configureScope, TArg arg);
}
```

These methods could then be used like this:
```csharp
string tagValue = "value";
SentrySdk.ConfigureScope(
    static (scope, arg) => scope.SetTag("tag_name", arg),
    tagValue);
```

I'd be happy to provide the implementation if this is something you're interested in.

bruno-garcia:
We should start with adding `SentrySdk.SetTag`, surprised we actualy don't have that yet.

Also not against adding overloads that allow for 0 allocations of course, but I do think we can add some stuff to our public API like SetTag and other common things.

KnapSac:
That would be even better, as you wouldn't have to go through the hassle of using a callback.

Would you be open to PRs for adding both (i.e. `SentrySdk.SetTag` and these 0 allocation overloads)?

jamescrosswell:
`SentrySdk.SetTag` sounds like a great idea 🙌🏻  - would love a PR for that!

In principle, the overloads leveraging static lambdas are really nice as well. Having them there relies on people using them appropriately though. I think if we have methods like `SentrySdk.SetTag` then people are more likely to build code that is naturally performant (even if they don't know about or understand static lambda optimisation). I'm thinking if we don't have these overloads, that forces a conversation about other methods we could potentially add to `SentrySdk`. `SentrySdk.SetExtra` would be the other obvious one... but are there more?

On the other hand, having the overloads would provide ultimate flexibility. Maybe we do both then - maybe we add `SentrySdk.SetTag` and `SentrySdk.SetExtra` and also add the overloads.

KnapSac:
[...]

PR #4232, opened 2025-05-30
Add SentrySdk.SetTag #4228

KnapSac:
I also added `SentrySdk.UnsetTag`, because it feels closely related with `SetTag`.

Some questions I had while implementing this:
- `Hub.ConfigureScope` wraps the call to `ScopeManager.ConfigureScope` in a try/catch, would that be appropriate for these methods too? E.g. `PushScope` doesn't use a try/catch.
  https://github.com/getsentry/sentry-dotnet/blob/aebd6a2e711c31087ca77652c31c6a624e22075b/src/Sentry/Internal/Hub.cs#L90-L100
- There is an extension method on `IHasTags` which can set multiple tags at once. Does it make sense to add it to `SentrySdk` too?
  https://github.com/getsentry/sentry-dotnet/blob/aebd6a2e711c31087ca77652c31c6a624e22075b/src/Sentry/IHasTags.cs#L33

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Features
 
 - Reduced memory pressure when sampling less than 100% of traces/transactions ([#4212](https://github.com/getsentry/sentry-dotnet/pull/4212))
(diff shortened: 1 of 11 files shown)


bruno-garcia (approved):
Perfect! Thank you so much for adding this

PR #4233, opened 2025-05-30
poc: custom disk transport for sentry-native to eliminate curl

jpnurmi:
WIP: Another experiment to eliminate the libcurl dependency, which is problematic for minimal/chiseled .NET containers.

#skip-changelog for now


jpnurmi:
Dropped in favor of the less risky variant that doesn't deserialize native envelopes as .NET envelopes:
- https://github.com/getsentry/sentry-dotnet/pull/4222

Issue #4194, opened 2025-05-16
Mark mono/xamarin/dotnet system symbols as not inApp


bruno-garcia:
Example event: https://demo.sentry.io/issues/6610741528/?project=4509153190739968&query=is%3Aunresolved%20issue.priority%3A%5Bhigh%2C%20medium%5D&referrer=issue-stream&stream_index=0

![Image](https://github.com/user-attachments/assets/4a87dafb-2227-4fdd-9d0e-9fa462a50edf)

jamescrosswell:
Tricky, the JSON for those frames doesn't include any Module information:
```json
            {
              "function": "mono_jit_exec",
              "symbol": "mono_jit_exec",
              "package": "/Users/jamescrosswell/Library/Developer/CoreSimulator/Devices/CFB78342-B78D-48E0-87A2-3B3A1311A8EE/data/Containers/Bundle/Application/0B1D5D54-E77B-4582-9033-B45046B4C46C/Sentry.Samples.Maui.app/libmonosgen-2.0.dylib",
              "in_app": true,
              "data": {
                "orig_in_app": 0,
                "client_in_app": false,
                "symbolicator_status": "symbolicated"
              },
              "image_addr": "0x1158d8000",
              "instruction_addr": "0x11593ea3c",
              "symbol_addr": "0x11593e8d4"
            },
```

The currently implementation of `InAppExcludes` only matches against either `Module` or `Function` and trying to identify all of the function names to match against isn't very practical.

Ideally we could exclude based on the package name (e.g. `libmonosgen` bit in the example frame above).

bruno-garcia:
it's in package the json you shared

Flash0ver:
We would like to add a dedicated "Package-Matcher" next to the existing "Module-" and "Function-Matchers".
https://github.com/getsentry/sentry-dotnet/blob/9beef00b325bd79c2cf43829f069142461aec113/src/Sentry/SentryStackFrame.cs#L19-L20
See also https://github.com/getsentry/sentry-dotnet/pull/4236#discussion_r2122614108

PR #4236, opened 2025-06-01
fix: system frames marked as inApp

bruno-garcia:
While dogfooding Symbol Collector I came across a few issues. This patch attempts to fix one of them where clearly system frames were marked as InApp.

It seems that our logic in the SDK is not taking `Package` into account at all but we set it here:

https://github.com/getsentry/sentry-dotnet/blob/0e11682db1da08cf4052525e528328197bc26354/src/Sentry/Internal/DebugStackTrace.cs#L310-L311

And it seems in some conditions, `Module` isn't set at all. For example on my test events:

From a .NET 9 CLI tool:
```json
{
  "function": "void Interop.ThrowExceptionForIoErrno(ErrorInfo errorInfo, string path, bool isDirError)",
  "package": "System.Private.CoreLib, Version=9.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e",
  "filename": "Interop.IOErrors.cs",
  "abs_path": "/_/src/libraries/Common/src/Interop/Unix/Interop.IOErrors.cs",
  "lineno": 16,
  "pre_context": [
    "    private static void ThrowExceptionForIoErrno(ErrorInfo errorInfo, string? path, bool isDirError)",
    "    {",
    "        Debug.Assert(errorInfo.Error != Error.SUCCESS);",
    "        Debug.Assert(errorInfo.Error != Error.EINTR, \"EINTR errors should be handled by the native shim and never bubble up to managed code\");",
    ""
  ],
  "context_line": "        throw Interop.GetExceptionForIoErrno(errorInfo, path, isDirError);",
  "post_context": [
    "    }",
    "",
    "    internal static void CheckIo(Error error, string? path = null, bool isDirError = false)",
    "    {",
    "        if (error != Interop.Error.SUCCESS)"
  ],
  "in_app": true,
  "data": {
    "client_in_app": true,
    "symbolicator_status": "symbolicated"
  },
  "instruction_addr": "0x0",
  "addr_mode": "rel:0",
  "function_id": "0x2",
  "source_link": "https://raw.githubusercontent.com/dotnet/runtime/f57e6dc747158ab7ade4e62a75a6750d16b771e8/src/libraries/Common/src/Interop/Unix/Interop.IOErrors.cs"
}
```

Note `"in_app": true` when clearly this is not inApp (Package is `System.`)

And .NET 9 Android, but in this case seems like we didn't mark `Java.` as InApp

```json
{
  "function": "void JniStaticMethods.InvokeVoidMethod(string, JniArgumentValue*)",
  "package": "Java.Interop, Version=9.0.0.0, Culture=neutral, PublicKeyToken=84e04ff9cfb79065",
  "in_app": true,
  "data": {
    "client_in_app": true,
    "symbolicator_status": "missing_symbol"
  },
  "instruction_addr": "0x14"
},
```

At first I thought we were overriding the default list, hence the test and refactoring. Let me know if you rather I undo that.

### DIFF
--- a/src/Sentry/SentryOptions.cs
+++ b/src/Sentry/SentryOptions.cs
@@ -1292,48 +1292,7 @@ public class SentryOptions
         Native = new NativeOptions(this);
 #endif
 
-        InAppExclude = new() {
-                "System",
-                "Mono",
-                "Sentry",
-                "Microsoft",
-                "MS", // MS.Win32, MS.Internal, etc: Desktop apps
-                "ABI.Microsoft", // MAUI
-                "WinRT", // WinRT, UWP, WinUI
-                "UIKit", // iOS / MacCatalyst
-                "Newtonsoft.Json",
-                "FSharp",
-                "Serilog",
-                "Giraffe",
-                "NLog",
-                "Npgsql",
-                "RabbitMQ",
-                "Hangfire",
-                "IdentityServer4",
-                "AWSSDK",
-                "Polly",
-                "Swashbuckle",
-                "FluentValidation",
-                "Autofac",
-                "Stackexchange.Redis",
-                "Dapper",
-                "RestSharp",
-                "SkiaSharp",
-                "IdentityModel",
-                "SqlitePclRaw",
-                "Xamarin",
-                "Android", // Ex: Android.Runtime.JNINativeWrapper...
-                "Google",
-                "MongoDB",
-                "Remotion.Linq",
-                "AutoMapper",
-                "Nest",
-                "Owin",
-                "MediatR",
-                "ICSharpCode",
-                "Grpc",
-                "ServiceStack"
-        };
+        InAppExclude = GetDefaultInAppExclude();
 
 #if DEBUG
         InAppInclude = new()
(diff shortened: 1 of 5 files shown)


src/Sentry/SentryOptions.cs
```diff
@@ -1827,4 +1786,49 @@ internal void SetupLogging()
...
+        "MongoDB",
+        "Remotion.Linq",
+        "AutoMapper",
+        "Nest",
+        "Owin",
+        "MediatR",
+        "ICSharpCode",
+        "Grpc",
+        "ServiceStack",
+        "Java.Interop",
```
bruno-garcia, line 1832: I added this

test/Sentry.Tests/SentryOptionsTests.cs
```diff
@@ -334,6 +334,26 @@ public void AddInAppExclude_StoredInOptions()
...
+
+    [Fact]
+    public void InAppExcludes_IgnoresJavaFrames()
+    {
+        var sut = new SentryOptions();
+        const string expected = "test";
+        sut.AddInAppExclude(expected);
+        Assert.All(SentryOptions.GetDefaultInAppExclude(),
+            item => Assert.Contains(item, sut.InAppExclude!));
+    }
```
bruno-garcia: Ended up moving the test to another file
```suggestion
```

src/Sentry/SentryStackFrame.cs
```diff
@@ -201,6 +201,10 @@ public void ConfigureAppFrame(SentryOptions options)
         {
             ConfigureAppFrame(options, Module, LazyModuleMatcher.Value);
         }
+        else if (!string.IsNullOrEmpty(Package))
```
Flash0ver, line 204: question: re: getsentry/sentry-dotnet#4194

This is related to issue getsentry/sentry-dotnet#4194, but does not quite close it yet, as we need to follow-up with a proper "Package-Matcher", that also considers File-Paths ... do I get that right?

jamescrosswell: I think so yeah... in that other issue (which was an Android stack trace) the package had a slightly different format like:
```
"package": "/somepath/libmonosgen-2.0.dylib",
```

The `LazyModuleMatcher` is a prefix or pattern matcher.... so it wouldn't match against something simple like `libmonosgen`. We'd need to use a regex expression to get it to match against Android packages.

A simple solution would be to create a dedicated matcher for packages, like:
```csharp
 private static readonly Lazy<SubstringOrPatternMatcher> LazyPackageMatcher = new(() => new()); 
```

That would use the already existing [SubstringOrPatternMatcher](https://github.com/getsentry/sentry-dotnet/blob/708c2e974a924ba62ae379b526af97d58b1d010a/src/Sentry/SubstringOrPatternMatcher.cs) (maybe yields false positives as it's basically just looking for a substring anywhere in the path). Ideally we'd come up with something a bit more sophisticated then.

PR #4238, opened 2025-06-01
ref: replace dsn in sample

bruno-garcia:
#skip-changelog

### DIFF
--- a/samples/Sentry.Samples.NLog/Program.cs
+++ b/samples/Sentry.Samples.NLog/Program.cs
@@ -12,7 +12,7 @@ public static class Program
 
     // DSN used by the example: 'UsingCodeConfiguration'.
     // #### ADD YOUR DSN HERE:
-    private static readonly string DsnSample = "https://eb18e953812b41c3aeb042e666fd3b5c@o447951.ingest.sentry.io/5428537";
+    private static readonly string DsnSample = "https://eb18e953812b41c3aeb042e666fd3b5c@o447951.ingest.us.sentry.io/5428537";
 
     private static void Main()
     {
--- a/samples/Sentry.Samples.Android/MainActivity.cs
+++ b/samples/Sentry.Samples.Android/MainActivity.cs
@@ -9,7 +9,7 @@ public class MainActivity : Activity
     {
         SentrySdk.Init(options =>
         {
-            options.Dsn = "https://eb18e953812b41c3aeb042e666fd3b5c@o447951.ingest.sentry.io/5428537";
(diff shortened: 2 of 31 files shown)


samples/Sentry.Samples.NLog/Program.cs
```diff
@@ -12,7 +12,7 @@ public static class Program
 
     // DSN used by the example: 'UsingCodeConfiguration'.
     // #### ADD YOUR DSN HERE:
-    private static readonly string DsnSample = "https://eb18e953812b41c3aeb042e666fd3b5c@o447951.ingest.sentry.io/5428537";
+    private static readonly string DsnSample = "https://eb18e953812b41c3aeb042e666fd3b5c@o447951.ingest.us.sentry.io/5428537";
```
Flash0ver, line 15: question: we use the DSN both via `AddSentry`/`SentryNLogOptions`, and also `NLog.config`

Is this intentional to show both variants, or shall we consolidate into either one?

Flash0ver, line 15: low: my other question made me notice that we have 31 samples, but only changes within 30 samples here

`Sentry.Samples.AspNetCore.Serilog` uses a different DSN
https://github.com/getsentry/sentry-dotnet/blob/0e11682db1da08cf4052525e528328197bc26354/samples/Sentry.Samples.AspNetCore.Serilog/appsettings.json#L6

Is this intentional, or should we use the same DSN as with all the other samples there too?

jamescrosswell: I think it's intentional yeah:
https://github.com/getsentry/sentry-dotnet/blob/43e28adab2859d8defc0c67692e3e00b4b2060bc/samples/Sentry.Samples.NLog/Program.cs#L92-L100

jamescrosswell: Not intentional I don't think.

The only reason we have the DSN hard coded into the samples is to make it easier for us as the maintainers to fire up the samples and test things out. We could potentially remove all the DSNs and do that via an environment variable instead:
https://github.com/getsentry/sentry-dotnet/blob/1aad9ab9b432d9dc3118b77aa9d3002a2767a451/src/Sentry/Internal/SettingLocator.cs#L42-L43

That would prevent something like this from happening in future as well.

@bruno-garcia thoughts?

bruno-garcia: yeah that's a good idea, if it'll work that way in all app types I dunno

bruno-garcia:
Pleace recycle the DSN, we got millions of spans from some random app accidetnly and we don't want taht data into our test project and prob neither do them. it's just a waste of quota on our end

PR #4243, opened 2025-06-02
Rename MemoryInfo.AllocatedBytes to MemoryInfo.TotalAllocatedBytes

### LINKED ISSUE #4242: Memory Info allocated bytes name is unclear
### Problem Statement

Events contain an additional `MemoryInfo` context. This is useful, but especially the `Allocated bytes` name is a bit unclear. It's actually the total allocated bytes over the lifetime of the process, not the currently allocated bytes (which is represented by `Heap Size Bytes`).

https://github.com/getsentry/sentry-dotnet/blob/0e11682db1da08cf4052525e528328197bc26354/src/Sentry/Internal/MainSentryEventProcessor.cs#L166

If the process has been running for some time this becomes obvious, because we don't have machines with that much memory. But after startup this is not immediately obvious.
Image

### Solution Brainstorm

I'd suggest renaming `Allocated Bytes` to `Total Allocated Bytes` and maybe `Heap Size Bytes` to `Currently Allocated Bytes`.

Additionally, it would be nice if there was a way to provide some additional information about what these values represent. E.g. linking to the [documentation](https://learn.microsoft.com/en-us/dotnet/api/system.gcmemoryinfo?view=net-9.0). However, I don't know if the UI supports that.

KnapSac:
Fixes #4242.

### DIFF
--- a/test/Sentry.Tests/Internals/MainSentryEventProcessorTests.cs
+++ b/test/Sentry.Tests/Internals/MainSentryEventProcessorTests.cs
@@ -62,7 +62,6 @@ public class MainSentryEventProcessorTests
 
         _ = sut.Process(evt);
         var memory = (MemoryInfo)evt.Contexts[MainSentryEventProcessor.MemoryInfoKey];
-        Assert.NotEqual(0, memory.TotalAvailableMemoryBytes);
         Assert.NotEqual(0, memory.HighMemoryLoadThresholdBytes);
         Assert.NotEqual(0, memory.TotalAvailableMemoryBytes);
     }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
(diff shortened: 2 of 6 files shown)


test/Sentry.Tests/Internals/MainSentryEventProcessorTests.cs
```diff
@@ -62,7 +62,6 @@ public void EnsureMemoryInfoExists()
 
         _ = sut.Process(evt);
         var memory = (MemoryInfo)evt.Contexts[MainSentryEventProcessor.MemoryInfoKey];
-        Assert.NotEqual(0, memory.TotalAvailableMemoryBytes);
```
KnapSac, line 65: This assert was repeated 2 lines below, which is why I removed it.

jamescrosswell (approved):
Thanks heaps for raising and addressing this @KnapSac !

PR #4244, opened 2025-06-02
Add non allocating ConfigureScope overload

### LINKED ISSUE #4228: Add overload to prevent delegate allocation for ConfigureScope

KnapSac:
Fixes #4228.


src/Sentry/ISentryScopeManager.cs
```diff
@@ -10,18 +10,33 @@ namespace Sentry;
 public interface ISentryScopeManager
 {
     /// <summary>
-    /// Configures the current scope.
+    /// Configures the current scope through the callback.
```
KnapSac, line 13: I made the wording consistent with the methods on `SentrySdk`.

src/Sentry/ISentryScopeManager.cs
```diff
@@ -10,18 +10,33 @@ namespace Sentry;
...
     public void ConfigureScope(Action<Scope> configureScope);
 
     /// <summary>
-    /// Asynchronously configure the current scope.
+    /// Configures the current scope through the callback.
     /// </summary>
-    /// <param name="configureScope">The configure scope.</param>
+    /// <param name="configureScope">The configure scope callback.</param>
+    /// <param name="arg">The argument to pass to the configure scope callback.</param>
+    internal void ConfigureScope<TArg>(Action<Scope, TArg> configureScope, TArg arg);
```
KnapSac: @bruno-garcia I added the overloads here as `internal`, but that would mean that these have to be implemented explicitly on e.g. `DisabledHub`, otherwise they'll still show up in the public API. IMO that would be the best option, I don't really see a clear alternative.

Internal calls of `ConfigureScope` usually go through the `IHub` interface, but adding them there as `internal` members makes no difference, we'd still have to implement them explicitly.

Thoughts?

jamescrosswell: When we needed to add members to a public interface in the past, we created a separate internal interface. For example, there was an [IHubEx](https://github.com/getsentry/sentry-dotnet/blob/3.33.0/src/Sentry/Internal/IHubEx.cs) in v3.x. `Hub` implemented both `IHub` and `IHubEx`... which slightly simplified the logic for calling code (it could just check if the instance was an `IHubEx` rather than checking if it was in some long list of classes that had that method) and made it easier to refactor when we wanted to move all of those methods over to the public `IHub` interface (in the next major release).

KnapSac: Ideally there wouldn't be any type-checks at the call sites, because that gets quite messy because of the delegate you need to pass (you'd basically need to repeat it twice due to the difference in arguments).

I'll take a look tonight to see if I can figure something out.

src/Sentry/Extensibility/DisabledHub.cs
```diff
@@ -30,11 +30,23 @@ public void ConfigureScope(Action<Scope> configureScope)
     {
     }
 
+    /// <summary>
+    /// No-Op.
+    /// </summary>
+    public void ConfigureScope<TArg>(Action<Scope, TArg> configureScope, TArg arg)
```
Flash0ver, line 36: naming: `arg` vs `state`

I believe that, when it comes to cachable delegates / local-capture-free lambdas, the ecosystem usually refers to arguments passed to user-provided callbacks as "state": e.g.
- https://source.dot.net/#q=System.Threading.ThreadPool.QueueUserWorkItem
- https://source.dot.net/#q=System.Formats.Asn1.AsnWriter.Encode
- https://source.dot.net/#q=Microsoft.Extensions.Logging.IExternalScopeProvider.ForEachScope

Where "arg" is usually used when providing collection-allocation-avoiding overloads named `arg0`, `arg1`, ... `argN`: e.g.
- https://source.dot.net/#q=System.String.Concat
- https://source.dot.net/#q=System.String.Format

Well,
on the other hand, the Runtime uses "arg" (`arg{1-N}`) in the `Invoke` method on `delegate` as parameter names.
More specifically, in our case the `delegate void Action<in T1, in T2>(T1 arg1, T2 arg2)` uses:
```CSharp
public virtual extern void Invoke(T1 arg1, T2 arg2)
```

I believe both names would work.
But I do have a slight preference towards "state" since I have the feeling that "state" is the idiomatic name when it comes to passing arguments to a user-provided delegate in order to avoid capturing locals.

What do you think?

KnapSac: I went with `arg` because it's really intended to be an argument which you explicitly pass along instead of capturing implicitly. Whereas `state` to me indicates that this is a value which you mutate inside the closure, possibly because the closure is invoked multiple times.

See for example `ConcurrentDictionary.GetOrAdd`, which also uses `arg`.

src/Sentry/ISentryScopeManager.cs
KnapSac: I couldn't find a viable alternative to implementing these overloads explicitly. The best solution might even be making them public, but @bruno-garcia expressed concern about the complexity in https://github.com/getsentry/sentry-dotnet/issues/4228#issuecomment-2931534622.

src/Sentry/Extensibility/HubAdapter.cs
KnapSac, line 1: `GetTransaction` uses `ConfigureScope` to get the transaction from the current scope. It might make sense to expose this transaction (internally) directly. There are 2 other locations where the transaction is set using `ConfigureScope`.

https://github.com/getsentry/sentry-dotnet/blob/9beef00b325bd79c2cf43829f069142461aec113/src/Sentry/HubExtensions.cs#L248-L253

src/Sentry.AspNetCore/SentryMiddleware.cs
```diff
@@ -131,16 +131,16 @@ public async Task InvokeAsync(HttpContext context, RequestDelegate next)
                 // when the event fires.  Use `activeScope`, not `scope` or `hub`.
                 scope.OnEvaluating += (_, activeScope) =>
```
KnapSac, line 132: The outer `ConfigureScope` call no longer allocates a closure, but this `OnEvaluating` event handler still allocates unconditionally. One solution might be to store `KeyValuePair`s of the callback and its argument instead of using an event inside a `Stack`. Perhaps the `Stack` wouldn't even be necessary when the `Sentry.AspNetCore.Grpc` integration isn't loaded, as that seems to be the only other place (outside of tests) where this event handler is assigned.

src/Sentry/Extensibility/HubAdapter.cs
jamescrosswell: Internally this extension method isn't required, since the `CurrentScope` is available directly (without a callback):
https://github.com/getsentry/sentry-dotnet/blob/b3239bb80069607b4dc72d0c8e889414ab284939/src/Sentry/Internal/Hub.cs#L34

But that only works for the `Hub` (not for `IHub`, `HubAdapter` etc.).

src/Sentry/ISentryScopeManager.cs
jamescrosswell: Something about having an internal member on a public interface rubs me the wrong way. I wasn't actually aware it was possible until I saw this PR.

Would it be more appropriate to put this member on the [IInternalScopeManager](https://github.com/getsentry/sentry-dotnet/blob/72ad46933c0ce9ea4e845d62482b0d2dad294a66/src/Sentry/Internal/IInternalScopeManager.cs)? That's basically what that interface is there for.

src/Sentry/Extensibility/DisabledHub.cs
jamescrosswell: Either is fine for me.

src/Sentry.AspNetCore/SentryMiddleware.cs
jamescrosswell: We could store a list of evaluators on the scope. This would replace the `OnEvaluating` event. So something like:

```csharp
internal interface IScopeEvaluator {
    void EvaluateScope(Scope scope);
}

public class Scope : IEventLike
{
    internal List<IEvaluator> Evaluators { get; }

    // All the other Scope stuff
    // ...
}
```

And `SentryMiddleware` could implement `IScopeEvaluator` instead of defining an `OnEvaluating` callback.

src/Sentry/ISentryScopeManager.cs
KnapSac: Except `IHub` doesn't implement `IInternalScopeManager`, so this runs into the same issues as defining an additional internal `IHubEx` interface.

IMO it might be best to simply make these overloads public. Seeing as this is a relatively common pattern to avoid closure allocations across the ecosystem, and there not being an easy way to add them internally, I feel like the extra complexity (which only shows up when auto completing) is acceptable.

src/Sentry.AspNetCore/SentryMiddleware.cs
KnapSac: Would that work? Because `OnEvaluating` is invoked when an event is about to be send to Sentry, at which point the `SentryMiddleware` instance might no longer be valid. And you'd need to store the `HttpContext` (which is passed as a parameter to the middleware).

src/Sentry/Extensibility/HubAdapter.cs
KnapSac: Would something like this be a solution?
```csharp
internal static ITransactionTracer? GetTransaction(this IHub hub)
{
    switch (hub)
    {
        case Hub fullHub:
            return fullHub.ScopeManager.GetCurrent().Key.Transaction;
        case DisabledHub:
            return null;
        case HubAdapter hubAdapter:
        {
            ITransactionTracer ? transaction = null;
            hub.ConfigureScope(scope => transaction = scope.Transaction);
            return transaction;
        }
        default:
            throw new ArgumentOutOfRangeException(
                nameof( hub ),
                "Unsupported hub type");
    }
}
```

src/Sentry/ISentryScopeManager.cs
jamescrosswell: Fair enough - some stuff we really don't want to expose in the public APIs and we're willing to invest a bit of effort to hide those things. In this case, there's not really any danger in exposing the overloads. Maybe we just make sure there are good docs explaining it (or link to some good docs that explain it).

src/Sentry.AspNetCore/SentryMiddleware.cs
jamescrosswell: I don't think the idea would work anyway, since the implementation for `EvaluateScope` would also need a reference to the `HttpContext` (which is a local variable, not a class member).

So yeah, you could do something like your `KeyValuePair` idea. There starts to be a trade off between performance and maintainability at some point... if the code isn't too ugly we could do it. If the code looks kind of funky, maybe we run some load tests to understand how much less memory is used or how much faster the code runs before deciding to run with it.

src/Sentry/Extensibility/HubAdapter.cs
jamescrosswell: It would. I think it could be simplified to:
```csharp
internal static ITransactionTracer? GetTransaction(this IHub hub)
{
    // We can optimise when running a full Hub
    if (hub is Hub fullHub)
    {
        return fullHub.ScopeManager.GetCurrent().Key.Transaction;
    }

    // Default implementation
    ITransactionTracer ? transaction = null;
    hub.ConfigureScope(scope => transaction = scope.Transaction);
    return transaction;
}
```

The Default implementation would be a [NoOp for the `DisabledHub`](https://github.com/getsentry/sentry-dotnet/blob/a04c3a65ae959b54e137721a7460ef94661c18cd/src/Sentry/Extensibility/DisabledHub.cs#L26-L31) and would work fine for the `HubAdapter` (which is only ever used in testing anyway... doesn't matter if that's not optimised)... and for any other/future implementations of IHub - so should be resilient to things we can't yet foresee. 

btw: we don't generally want to throw exceptions from the Sentry SDK... so definitely don't throw an `ArgumentOutOfRangeException` exception from this method.

src/Sentry.AspNetCore/SentryMiddleware.cs
KnapSac: Definitely seems best to leave this to a separate PR.

Is there any infrastructure in place to measure the overhead of `SentryMiddleware`/`SentryTracingMiddleware` both in terms of time and memory? If not, that would  probably be a good thing to have. As these middleware implementations are invoked for every request if the SDK is enabled, it makes sense to optimize them as much as possible IMO.

test/Sentry.AspNetCore.Tests/SentryMiddlewareTests.cs
```diff
@@ -34,6 +34,18 @@ public Fixture()
             Hub.When(hub => hub.ConfigureScope(Arg.Any<Action<Scope>>()))
                 .Do(callback => callback.Arg<Action<Scope>>().Invoke(Scope));
 
+            Hub.When(hub => hub.ConfigureScope(
```
KnapSac: Any ideas what the proper way to do this would be? This is the only thing I could get to work, I tried using `Arg.AnyType` in the `Do` implementation but it wasn't able to find the argument.

src/Sentry.AspNetCore/SentryMiddleware.cs
jamescrosswell: Sounds good. I've created a separate issue to track this:
- https://github.com/getsentry/sentry-dotnet/issues/4253

test/Sentry.AspNetCore.Tests/SentryMiddlewareTests.cs
jamescrosswell: This is maybe as good as it gets...

test/Sentry.AspNetCore.Tests/SentryMiddlewareTests.cs
```diff
@@ -157,23 +156,16 @@ public async Task InvokeAsync_ScopePushed_BeforeConfiguringScope()
...
-                .First(c => c.ArgAt<Action<Scope>>(0)(scope))
-                .Then(_ =>
-                {
-                    Assert.True(scope.Locked);
-                    verified = true;
-                }));
+        var scopeLocked = false;
+        _fixture.Hub.When(h => h.PushAndLockScope()).Do(_ => scopeLocked = true);
+        _fixture.Hub.When(h => h.ConfigureScope(Arg.Any<Action<Scope>>()))
+            .Do(_ => Assert.True(scopeLocked));
```
KnapSac: I changed this test slightly because I couldn't get it to work with the new overload. It seems to me like this still functionally tests the same thing as before, although I'm not even sure the test implementation was correct.

The test used to first invoke the `ConfigureScope` callback and **then** assert that the scope is locked, or am I misunderstanding that?

jamescrosswell: I think the previous test was fine. It basically invoked the configurescope callback on the local `var scope` when this line executed (since the hub at this point is a substitute):
https://github.com/getsentry/sentry-dotnet/blob/34d849038e3e23868a12d56ea256c36b0a7f3a61/src/Sentry/HubExtensions.cs#L178

Then it checked to ensure that `scope.Locked == true` afterwards. 

I don't think the new test is validating anything since the test code itself is trying to set `scopeLocked` to true and then checking to see if it's true aftewards... it's not actually testing any code in the SDK. 

Note that [`hub.PushAndLockScope()`](https://github.com/getsentry/sentry-dotnet/blob/a46bce298a9aafb25ae847bc14d23fe498fd82e2/src/Sentry.AspNetCore/SentryMiddleware.cs#L95) is actually calling an extension method (so you can't mock this with NSubstitute). Ultimately what's being tested here is this line of code:
https://github.com/getsentry/sentry-dotnet/blob/6863c2e2f92208c61971dee68ada9bc60491c0a5/src/Sentry/HubExtensions.cs#L181

Did `HubSubstituteExtensions.SubstituteConfigureScope` not work here? It looks like you used this successfully everywhere else...

test/Sentry.Testing/HubSubstituteExtensions.cs
jamescrosswell, line 1: Fancy 😜

jamescrosswell (commented):
Generally looking really good - just need to to get that one test working and I think we're good to go.

test/Sentry.AspNetCore.Tests/SentryMiddlewareTests.cs
KnapSac: Yeah, you're right. Those substitute methods take some getting used to ;). I'll take a look and try to get the old test working.

KnapSac: I'm not sure why I can't remove the first substitute in this method, its `Then` callback doesn't seem to be called, but if I remove it the test no longer passes...

src/Sentry/Internal/Hub.cs
```diff
@@ -111,6 +123,18 @@ public async Task ConfigureScopeAsync(Func<Scope, Task> configureScope)
...
 
+    public async Task ConfigureScopeAsync<TArg>(Func<Scope, TArg, Task> configureScope, TArg arg)
+    {
+        try
+        {
+            await ScopeManager.ConfigureScopeAsync(configureScope, arg).ConfigureAwait(false);
+        }
+        catch (Exception e)
+        {
+            _options.LogError(e, "Failure to ConfigureScopeAsync");
```
bruno-garcia:
```suggestion
            _options.LogError(e, "Failure to ConfigureScopeAsync<TArg>");
```

src/Sentry/Internal/Hub.cs
```diff
@@ -99,6 +99,18 @@ public void ConfigureScope(Action<Scope> configureScope)
...
 
+    public void ConfigureScope<TArg>(Action<Scope, TArg> configureScope, TArg arg)
+    {
+        try
+        {
+            ScopeManager.ConfigureScope(configureScope, arg);
+        }
+        catch (Exception e)
+        {
+            _options.LogError(e, "Failure to ConfigureScope");
```
bruno-garcia:
```suggestion
            _options.LogError(e, "Failure to ConfigureScope<TArg>");
```

bruno-garcia (approved):
Just some minor suggestions but otherwise LGTM! Thanks!

test/Sentry.Tests/Protocol/SentryTransactionTests.cs
```diff
@@ -582,18 +582,18 @@ public void Finish_ResetsScopeAndSetsNewPropagationContext()
         var hub = Substitute.For<IHub>();
         var transaction = new TransactionTracer(hub, "test name", "test op");
 
-        Action<Scope> capturedAction = null;
-        hub.ConfigureScope(Arg.Do<Action<Scope>>(action => capturedAction = action));
+        Action<Scope, TransactionTracer> capturedAction = null;
```
Flash0ver, line 585: question: Would it make sense to keep the original test using `ConfigureScope(Action<Scope>)`, and add a new test using the new `ConfigureScope<TArg>(Action<Scope, TArg>, TArg)`?

test/Sentry.AspNetCore.Tests/SentryMiddlewareTests.cs
```diff
@@ -160,14 +159,22 @@ public async Task InvokeAsync_LocksScope_BeforeConfiguringScope()
...
         _fixture.Hub
-            .When(h => h.ConfigureScope(Arg.Any<Action<Scope>>()))
-            .Do(Callback
-                .First(c => c.ArgAt<Action<Scope>>(0)(scope))
-                .Then(_ =>
-                {
-                    Assert.True(scope.Locked);
-                    verified = true;
-                }));
+                .When(h => h.ConfigureScope(Arg.Any<Action<Scope>>()))
```
Flash0ver: nitpick: Would you mind restoring the previous 4-space-indentation, rather than the "aligned" indentation?

To be consistent in style within this method and document, and also to make the actual change more apparent.

src/Sentry/Internal/Hub.cs
```diff
@@ -272,7 +296,7 @@ public TransactionContext ContinueTrace(
         string? operation = null)
     {
         var propagationContext = SentryPropagationContext.CreateFromHeaders(_options.DiagnosticLogger, traceHeader, baggageHeader, _replaySession);
-        ConfigureScope(scope => scope.SetPropagationContext(propagationContext));
+        ConfigureScope(static (scope, propagationContext) => scope.SetPropagationContext(propagationContext), propagationContext);
```
Flash0ver, line 299: praise: I like the use of _static anonymous functions_ to highlight the non-capturing of locals/instances.

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Features
+
+- Added non allocating SentrySdk.ConfigureScope overload ([#4244](https://github.com/getsentry/sentry-dotnet/pull/4244))
```
Flash0ver: suggestion: rephrase the changelog entry slightly:

```suggestion
- Added non-allocating `ConfigureScope` and `ConfigureScopeAsync` overloads ([#4244](https://github.com/getsentry/sentry-dotnet/pull/4244))
```

- I believe "non-allocating" is a hyphenated compound
- Remove `SentrySdk` since we also add these overloads to Hub and ScopeManager
- Also mention the `Async` variant

Also ... I'm afraid we got another merge conflict with _main_.

src/Sentry/HubExtensions.cs
```diff
@@ -247,6 +244,11 @@ internal static ITransactionTracer StartTransaction(
 
     internal static ITransactionTracer? GetTransaction(this IHub hub)
     {
+        if (hub is Hub fullHub)
```
Flash0ver, line 247: praise: nice special-case (to reduce closure allocation) and consolidating in `StartSpan`

src/Sentry/HubExtensions.cs
```diff
@@ -56,14 +56,11 @@ public static ITransactionTracer StartTransaction(
     /// </summary>
     public static ISpan StartSpan(this IHub hub, string operation, string description)
     {
-        ITransactionTracer? currentTransaction = null;
-        hub.ConfigureScope(s => currentTransaction = s.Transaction);
-        return currentTransaction is { } transaction
+        return hub.GetTransaction() is { } transaction
```
Flash0ver, line 59: test: Should we have a test for `StartSpan` in `HubExtensionsTests.cs`

To be fair, these method was recently moved and made `public`, where we should have added a test for this, but missed it.
Now, that this method calls an `internal` method, I'm wondering if this is adding more "worth" of adding a test for this extension method.

@KnapSac, only if you feel like adding a test for this method.
To be fair, though, it should have been added in a preceding PR.

test/Sentry.Tests/Protocol/SentryTransactionTests.cs
KnapSac: I don't believe that would be possible, because we're substituting a `ConfigureScope` call here. As the old overload is no longer called, that test wouldn't pass anymore.

Or do you mean to add the old test and change it to check that the old overload isn't called?

test/Sentry.AspNetCore.Tests/SentryMiddlewareTests.cs
KnapSac: No problem.

I did run `dotnet format` locally (because for some reason my ReSharper doesn't format everything the same as `dotnet format`) and this is the result. Any ideas why `dotnet format` didn't format this consistently with the rest of this file?

test/Sentry.Tests/Protocol/SentryTransactionTests.cs
Flash0ver: oh ... right ... thanks for clarifying ... was a misunderstanding from my end

test/Sentry.AspNetCore.Tests/SentryMiddlewareTests.cs
Flash0ver: thanks for the follow-up

No .. I'm afraid I am uncertain about the inconsistencies of `dotnet format`.
Hmm .. perhaps something missing in our `.editorconfig`?
I'll keep an eye out for it in the future and will be using `dotnet format` more to detect and fix inconsistencies.

test/Sentry.AspNetCore.Tests/SentryMiddlewareTests.cs
jamescrosswell: I assume you're referring to this one:
https://github.com/getsentry/sentry-dotnet/blob/dc5c85318afa7abbf931060d9ac84cece3014d88/test/Sentry.AspNetCore.Tests/SentryMiddlewareTests.cs#L161-L169

That would ordinarily get called by `HubExtensions.LockScope` here:
https://github.com/getsentry/sentry-dotnet/blob/ee4ce7d955049064ff4eb3b130add7fd06dc56f6/src/Sentry/HubExtensions.cs#L175

Since the test uses a substitute for the Hub, that behaviour has to be imitated by the substitute.

It's definitely a bit tricky to walk down all the paths mentally and then setup the appropriate substitutes. 

In any event, this test looks all good now 👍🏻

PR #4245, opened 2025-06-03
fix: InvalidCastException in SentrySpanProcessor

### LINKED ISSUE #4235: InvalidCastException: Unable to cast object of type 'Sentry.Internal.UnsampledTransaction' to type 'Sentry.TransactionTracer'
### Package

~Sentry.Extensions.Logging~
Sentry.OpenTelemetry

### .NET Flavor

.NET Core

### .NET Version

9.0.5

### OS

Any (not platform specific)

### OS Version

_No response_

### Development Environment

Visual Studio v17.x

### SDK Version

5.9.0

_No response_

### Workload Versions

N/A

```
logging.AddSentry(options =>
{
    options.UseOpenTelemetry();
    options.DisableUnobservedTaskExceptionCapture();
});
```

### Steps to Reproduce

After upgrading the Sentry SDK from 5.8.0 to 5.9.0, this exception started appearing - seems like a regression in this release.

### Expected Result

No exception thrown.

### Actual Result

```
System.InvalidCastException: Unable to cast object of type 'Sentry.Internal.UnsampledTransaction' to type 'Sentry.TransactionTracer'.
  ?, in void SentrySpanProcessor.CreateRootSpan(Activity data)
  ?, in void SentrySpanProcessor.OnStart(Activity data)
  ?, in TracerProviderSdk(IServiceProvider serviceProvider, bool ownsServiceProvider)+(Activity activity) => { }
  ?, in Activity Activity.Start()
```

jamescrosswell:
Resolves #4235:
- https://github.com/getsentry/sentry-dotnet/issues/4235

From what I can tell, only the OpenTelemetry integration was affected by this.

I've added some extra unit tests to test the `OnStart` and `OnEnd` hooks for our `SpanProcessor` (the tests we had weren't testing `unsampled` scenarios) and updated the code to pass the new tests.

### DIFF
--- a/src/Sentry.OpenTelemetry/SentrySpanProcessor.cs
+++ b/src/Sentry.OpenTelemetry/SentrySpanProcessor.cs
@@ -139,9 +139,9 @@ public class SentrySpanProcessor : BaseProcessor<Activity>
             spanTracer.Origin = OpenTelemetryOrigin;
             spanTracer.StartTimestamp = data.StartTimeUtc;
             // Used to filter out spans that are not recorded when finishing a transaction.
-            spanTracer.IsFiltered = () => span.GetFused<Activity>() is { IsAllDataRequested: false, Recorded: false };
+            spanTracer.SetFused(data);
+            spanTracer.IsFiltered = () => spanTracer.GetFused<Activity>() is { IsAllDataRequested: false, Recorded: false };
         }
-        span.SetFused(data);
         _map[data.SpanId] = span;
     }
 
@@ -248,8 +248,8 @@ public class SentrySpanProcessor : BaseProcessor<Activity>
 
             // Spans set otel attributes in extras (passed to Sentry as "data" on the span).
             // Resource attributes do not need to be set, as they would be identical as those set on the transaction.
-            span.SetExtras(attributes);
-            span.SetExtra("otel.kind", data.Kind);
+            spanTracer.SetExtras(attributes);
+            spanTracer.SetExtra("otel.kind", data.Kind);
         }
 
         // In ASP.NET Core the middleware finishes up (and the scope gets popped) before the activity is ended.  So we
--- a/src/Sentry/Internal/UnsampledSpan.cs
+++ b/src/Sentry/Internal/UnsampledSpan.cs
@@ -1,6 +1,6 @@
 namespace Sentry.Internal;
 
-internal class UnsampledSpan(UnsampledTransaction transaction, SpanId? spanId = null) : NoOpSpan
+internal sealed class UnsampledSpan(UnsampledTransaction transaction, SpanId? spanId = null) : NoOpSpan
 {
     public override bool? IsSampled => false;
     public override SpanId SpanId { get; } = spanId ?? SpanId.Empty;
--- a/test/Sentry.OpenTelemetry.Tests/SentrySpanProcessorTests.cs
+++ b/test/Sentry.OpenTelemetry.Tests/SentrySpanProcessorTests.cs
@@ -426,7 +426,7 @@ public class SentrySpanProcessorTests : ActivitySourceTests
         span.Should().BeOfType<UnsampledSpan>();
 
         // There's nothing else to assert here, as long as calling OnEnd does not throw an exception,
-        // UnsampleSpan.Finish() is basically a no-op.
+        // UnsampledSpan.Finish() is basically a no-op.
     }
 
     [Fact]


CHANGELOG.md
```diff
@@ -6,6 +6,10 @@
 
 - Rename MemoryInfo.AllocatedBytes to MemoryInfo.TotalAllocatedBytes ([#4243](https://github.com/getsentry/sentry-dotnet/pull/4243))
 
+### Fixes
+
+- InvalidCastException in SentrySpanProcessor ([#4245](https://github.com/getsentry/sentry-dotnet/pull/4245))
```
Flash0ver: nitpick: shall we add some note about the containing/impacted package?

e.g.
```suggestion
- InvalidCastException in SentrySpanProcessor of `Sentry.OpenTelemetry` ([#4245](https://github.com/getsentry/sentry-dotnet/pull/4245))
```

src/Sentry.OpenTelemetry/SentrySpanProcessor.cs
```diff
@@ -236,10 +241,10 @@ public override void OnEnd(Activity data)
             // Transactions set otel attributes (and resource attributes) as context.
             transaction.Contexts["otel"] = GetOtelContext(attributes);
         }
-        else
+        else if (span is SpanTracer spanTracer)
         {
             // Use the end timestamp from the activity data.
-            ((SpanTracer)span).EndTimestamp = data.StartTimeUtc + data.Duration;
+            spanTracer.EndTimestamp = data.StartTimeUtc + data.Duration;
```
Flash0ver, line 247: nitpick: consistency

Should we also invoke `SetExtras` and `SetExtra` via the `SpanTracer spanTracer` variable rather than the `ISpan span` variable in the two lines below?

It does not make any runtime difference ... except that the JIT may have a higher chance to devirtualize the latter method call.

But my comment is more about style rather than perf.

src/Sentry.OpenTelemetry/SentrySpanProcessor.cs
```diff
@@ -133,12 +133,15 @@ private void CreateChildSpan(Activity data, ISpan parentSpan, SpanId? parentSpan
...
-        span.Origin = OpenTelemetryOrigin;
-        span.StartTimestamp = data.StartTimeUtc;
-        // Used to filter out spans that are not recorded when finishing a transaction.
+        var span = parentSpan.StartChild(context);
+        if (span is SpanTracer spanTracer)
+        {
+            spanTracer.Origin = OpenTelemetryOrigin;
+            spanTracer.StartTimestamp = data.StartTimeUtc;
+            // Used to filter out spans that are not recorded when finishing a transaction.
+            spanTracer.IsFiltered = () => span.GetFused<Activity>() is { IsAllDataRequested: false, Recorded: false };
```
Flash0ver: nitpick: `span.GetFused` vs `spanTracer.GetFused`

```suggestion
            spanTracer.IsFiltered = () => spanTracer.GetFused<Activity>() is { IsAllDataRequested: false, Recorded: false };
```

Total nitpick.
A question about style/explicitness, rather than functionality.

src/Sentry/ISpan.cs
```diff
@@ -70,15 +70,21 @@ public static ISpan StartChild(this ISpan span, string operation, string? descri
...
+            if (child is SpanTracer childTracer)
+            {
+                childTracer.Description = context.Description;
+                return childTracer;
+            }
         }
-
-        childSpan.Description = context.Description;
-        return childSpan;
+        if (transaction is UnsampledTransaction unsampledTransaction)
```
Flash0ver, line 83: question: case `SpanTracer`

Do we need a case for when `span.GetTransaction()` returns a `SpanTracer` as well?
Or is it fine to return a NoOp-Instance in that case? Or may that case even occur?

src/Sentry/Internal/UnsampledSpan.cs
```diff
@@ -0,0 +1,9 @@
+namespace Sentry.Internal;
+
+internal class UnsampledSpan(UnsampledTransaction transaction, SpanId? spanId = null) : NoOpSpan
```
Flash0ver: medium: `sealed`

Could we seal this class too? To be in sync with `UnsampledTransaction`.

test/Sentry.OpenTelemetry.Tests/SentrySpanProcessorTests.cs
```diff
@@ -338,6 +401,34 @@ public void OnEnd_FinishesSpan()
...
+        sut._map.TryGetValue(data.SpanId, out var span);
+
+        // Act
+        sut.OnEnd(data);
+
+        // Assert
+        span.Should().BeOfType<UnsampledSpan>();
+
+        // There's nothing else to assert here, as long as calling OnEnd does not throw an exception,
+        // UnsampleSpan.Finish() is basically a no-op.
```
Flash0ver: low: typo

```suggestion
        // UnsampledSpan.Finish() is basically a no-op.
```

src/Sentry/Internal/UnsampledTransaction.cs
```diff
@@ -94,10 +94,10 @@ public override ISpan StartChild(string operation)
         return span;
     }
 
-    private class UnsampledSpan(UnsampledTransaction transaction) : NoOpSpan
+    public ISpan StartChild(string operation, SpanId spanId)
     {
-        public override ISpan StartChild(string operation) => transaction.StartChild(operation);
-
-        public override bool? IsSampled => false;
+        var span = new UnsampledSpan(this, spanId);
```
Flash0ver, line 99: question: do existing casts to `SpanTracer` require changes as well?

I did a quick search for casts to `SpanTracer`, and found these two usages:

https://github.com/getsentry/sentry-dotnet/blob/9beef00b325bd79c2cf43829f069142461aec113/src/Sentry.DiagnosticSource/Internal/DiagnosticSource/EFConnectionDiagnosticSourceHelper.cs#L58-L64

https://github.com/getsentry/sentry-dotnet/blob/9beef00b325bd79c2cf43829f069142461aec113/src/Sentry/SentrySpan.cs#L116-L129

I'm just wondering if they also need some changes?
I don't think so, from what I can see ... but I just wanted to make sure.

CHANGELOG.md
```suggestion
- InvalidCastException in SentrySpanProcessor when using the Sentry.OpenTelemetry integration ([#4245](https://github.com/getsentry/sentry-dotnet/pull/4245))
```

src/Sentry/ISpan.cs
jamescrosswell: It would never occur no (you can check the implementation for `GetTransaction` to see why) and it doesn't make sense for `GetTransaction` to return a non-root span... the whole point is to find the root span (i.e. Transaction) for the `SpanTracer`.

src/Sentry/Internal/UnsampledTransaction.cs
jamescrosswell: No I don't think so. I checked those as well but they just add additional information in the case that a span is sampled. If the span is an UnsampledSpan there's nothing that needs to be done there.

PR #4247, opened 2025-06-03
WIP: CI: provide faster feedback for "no mobile" builds

jpnurmi:


.github/workflows/build.yml
```diff
@@ -192,7 +192,6 @@ jobs:
       # To save time and disk space, we only create and archive the NuGet packages when we're actually releasing.
```
jamescrosswell: I guess we should remove this comment too then?

.github/workflows/build.yml
```diff
@@ -205,23 +204,29 @@ jobs:
...
 
+      - name: Install build dependencies
+        if: runner.os == 'Linux' && !matrix.container
+        run: |
+          sudo apt update
+          sudo apt install libcurl4-openssl-dev
+
+      # Run integration tests as soon as possible for non-publishing builds
+      - name: Integration Test
+        if: env.CI_PUBLISHING_BUILD != 'true'
```
jamescrosswell: Why only for non-publishing builds? We can run them whenever we have NuGet packages right (which means for all builds)?

.github/workflows/build.yml
```diff
@@ -205,23 +204,29 @@ jobs:
...
-    name: Integration test (${{ matrix.container || matrix.os }})
+    name: Integration test (${{ matrix.os }})
     runs-on: ${{ matrix.os }}
-    container: ${{ matrix.container }}
 
     strategy:
       fail-fast: false
       matrix:
         include:
-          - os: ubuntu-22.04 # Pin ubuntu to ensure mono is installed
```
jamescrosswell, line 212: What's the motivation behind dropping integration tests on these runners?

.github/workflows/build.yml
jpnurmi: Good point! I started by conditionally moving the "no mobile" platforms first because they were easy targets. After some build system tweaks, the same CI flow works for all platforms :+1:

.github/workflows/build.yml
jpnurmi: All runners continue to run integration tests. These were just moved to the .NET build job to avoid waiting for macOS.

image

.github/workflows/build.yml
```diff
@@ -199,44 +196,24 @@ jobs:
...
-    steps:
-      - uses: actions/checkout@v4
+      - name: Sparse checkout
+        if: env.CI_PUBLISHING_BUILD == 'true'
+        uses: actions/checkout@v4
         with:
           # We only check out what is absolutely necessary to reduce a chance of local files impacting
           # integration tests, e.g. Directory.Build.props, nuget.config, ...
           sparse-checkout: |
-            Directory.Build.props
```
jpnurmi, line 226: The old "non-cone" mode supported file patterns, but the default cone mode requires a list of directories: https://git-scm.com/docs/git-sparse-checkout

> 'Directory.Build.props' is not a directory; to treat it as a directory anyway, rerun with --skip-checks

.github/workflows/build.yml
```diff
@@ -199,44 +196,24 @@ jobs:
...
-          - os: ubuntu-latest
-            container: ghcr.io/getsentry/sentry-dotnet-alpine:3.21
-          - os: macos-15 # Pin macos to get the version of Xcode that we need: https://github.com/actions/runner-images/issues/10703
-          - os: windows-latest
-          - os: windows-11-arm
-
-    steps:
-      - uses: actions/checkout@v4
+      - name: Sparse checkout
+        if: env.CI_PUBLISHING_BUILD == 'true'
```
jpnurmi, line 200: Not sure if this is specific to the publishing build or macOS, but here's the integration test error for reference. Other runners have no problems, this only occurs on macOS if we don't switch to a sparse checkout:

```
building maui-app
      Detected signing identity:
              
        Bundle Id: com.companyname.mauiapp
        App Id: com.companyname.mauiapp
      maui-app -> /maui-app/bin/Release/net8.0-ios17.0/iossimulator-arm64/maui-app.dll
      Optimizing assemblies for size may change the behavior of the app. Be sure to test after publishing. See: https://aka.ms/dotnet-illink
      Optimizing assemblies for size. This process might take a while.
      Tool /Users/runner/.dotnet/packs/Microsoft.NETCore.App.Runtime.AOT.osx-arm64.Cross.iossimulator-arm64/8.0.15/Sdk/../tools/mono-aot-cross execution finished (exit code = 137).
              
    /Users/runner/.dotnet/packs/Microsoft.iOS.Sdk.net8.0_17.0/17.0.8523/targets/Xamarin.Shared.Sdk.targets(1214,3): error : Failed to AOT compile aot-instances.dll, the AOT compiler exited with code 137. [/maui-app/maui-app.csproj::TargetFramework=net8.0-ios17.0]
      Tool /Users/runner/.dotnet/packs/Microsoft.NETCore.App.Runtime.AOT.osx-arm64.Cross.iossimulator-arm64/8.0.15/Sdk/../tools/mono-aot-cross execution finished (exit code = 1).
              
    /Users/runner/.dotnet/packs/Microsoft.iOS.Sdk.net8.0_17.0/17.0.8523/targets/Xamarin.Shared.Sdk.targets(1214,3): error : Failed to AOT compile Microsoft.iOS.dll, the AOT compiler exited with code 1. [/maui-app/maui-app.csproj::TargetFramework=net8.0-ios17.0]
    
    Build FAILED.
    
    /Users/runner/.dotnet/packs/Microsoft.iOS.Sdk.net8.0_17.0/17.0.8523/targets/Xamarin.Shared.Sdk.targets(1214,3): error : Failed to AOT compile aot-instances.dll, the AOT compiler exited with code 137. [/maui-app/maui-app.csproj::TargetFramework=net8.0-ios17.0]
    /Users/runner/.dotnet/packs/Microsoft.iOS.Sdk.net8.0_17.0/17.0.8523/targets/Xamarin.Shared.Sdk.targets(1214,3): error : Failed to AOT compile Microsoft.iOS.dll, the AOT compiler exited with code 1. [/maui-app/maui-app.csproj::TargetFramework=net8.0-ios17.0]
        0 Warning(s)
        2 Error(s)
```

src/Sentry/Platforms/Native/Sentry.Native.targets
```diff
@@ -34,42 +34,42 @@
...
-  <ItemGroup Condition="'$(CI_PUBLISHING_BUILD)' == 'true' or '$(RuntimeIdentifier)' == 'win-x64'">
+  <ItemGroup Condition="'$(CI_PUBLISHING_BUILD)' == 'true' or Exists('$(SentryNativeOutputDirectory-win-x64)$(SentryNativeLibraryName).lib')">
     <None Include="$(SentryNativeOutputDirectory-win-x64)$(SentryNativeLibraryName).lib">
       <Pack>true</Pack>
       <PackagePath>\sentry-native\$(NativeLibRelativePath-win-x64)</PackagePath>
     </None>
   </ItemGroup>
 
-  <ItemGroup Condition="'$(CI_PUBLISHING_BUILD)' == 'true' or '$(RuntimeIdentifier)' == 'win-arm64'">
+  <ItemGroup Condition="'$(CI_PUBLISHING_BUILD)' == 'true' or Exists('$(SentryNativeOutputDirectory-win-arm64)$(SentryNativeLibraryName).lib')">
```
vaind: In case a future change breaks the build in such a way that the artifacts are not produced, we'd not get a failure but it would be dealt with silently.

Why do we need this change or rather, is there another way we could achieve that while not relying on existence of files but instead check the environment/platform, etc?

vaind (changes requested):

src/Sentry/Platforms/Native/Sentry.Native.targets
jpnurmi: Originally, this was using `[MSBuild]::IsOsPlatform()` checks. It was no longer sufficient after we started adding multiple runtimes per OS (`linux-x64` vs. `linux-musl-x64` vs. `linux-arm64`, `win-x64` vs. `win-arm64`). I made a mistake using `$(RuntimeIdentifier)` to distinguish between the runtimes, but it only works for building, not packing. The problem went unnoticed in the CI because only `CI_PUBLISHING_BUILD` would pack. The problem surfaced now that each runner packs their own .nupkg.

The idea with the `Exists()` check was to intentionally relax the rules so that one could even build multi-platform packages locally if desired, because it picks whatever is available. If the files didn't exist, packing would fail on `CI_PUBLISHING_BUILD` and integration tests would fail on the rest of the runners. However, I've pushed a proposal to change it to `[System.Runtime.InteropServices.RuntimeInformation]::RuntimeIdentifier` to restore the original behavior. What do you think?

jamescrosswell:
> Before
> All platforms ran integration tests against a NuGet package built on macOS with sentry-native libraries included for all platforms.
> 
> After
> Windows and Linux run integration tests against a locally built NuGet package with a sentry-native library for that same platform only.

As I recall, it was done this way because the NuGet packages that get uploaded to nuget.org get built on the macOS runner... so the intention here was to make sure that we were running the tests against the exact same version of the code that we were shipping.

If we use separate builds of the NuGets (built using different processors and operating systems) then we're not necessarily testing the exact same thing that we're shipping to customers.

@bruno-garcia might have some input here.

bruno-garcia:
> As I recall, it was done this way because the NuGet packages that get uploaded to nuget.org get built on the macOS runner... so the intention here was to make sure that we were running the tests against the exact same version of the code that we were shipping.

> If we use separate builds of the NuGets (built using different processors and operating systems) then we're not necessarily testing the exact same thing that we're shipping to customers.

Right but does it matter? the #ifdef is about the compilation target, we're still running the code that will be generated for the right platform. Otherwise we're testing .NET's ability to cross compile, no?

src/Sentry/Platforms/Native/Sentry.Native.targets
jamescrosswell: > If the files didn't exist, packing would fail on CI_PUBLISHING_BUILD and integration tests would fail on the rest of the runners.

That seems sufficient right @vaind ? As long as it's not too hard to unravel why it fails - possibly we could log a message explaining what went wrong as well?

vaind: Yeah sounds good

vaind (approved):

PR #4248, opened 2025-06-04
Add MAUI AutomationId to generic element info

albyrock87:
- Add `AutomationId` to elements info: if the developer choose to set one it means it's an important info to discriminate a given element
- Avoid reading the same bindable properties twice (minor perf improvement)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 
 - Rename MemoryInfo.AllocatedBytes to MemoryInfo.TotalAllocatedBytes ([#4243](https://github.com/getsentry/sentry-dotnet/pull/4243))
 - Replace libcurl with .NET HttpClient for sentry-native ([#4222](https://github.com/getsentry/sentry-dotnet/pull/4222))
+- Add .NET MAUI `AutomationId` element information to breadcrumbs ([#4248](https://github.com/getsentry/sentry-dotnet/pull/4248)) 
 
 ### Fixes
 
--- a/src/Sentry.Maui/Internal/Extensions.cs
+++ b/src/Sentry.Maui/Internal/Extensions.cs
@@ -53,21 +53,26 @@ internal static class Extensions
         // The element ID seems to be mostly useless noise
         //data.Add(prefix + nameof(element.Id), element.Id.ToString());
 
-        if (element.StyleId != null)
+        if (element.AutomationId is { } automationId)
+        {
+            data.Add(prefix + nameof(Element.AutomationId), automationId);
+        }
+
+        if (element.StyleId is { } styleId)
         {
             // The StyleId correlates to the element's name if one is set in XAML
             // TODO: Is there a better way to get this?
-            data.Add(prefix + "Name", element.StyleId);
+            data.Add(prefix + "Name", styleId);
         }
 
-        if (options.IncludeTitleInBreadcrumbs && element is ITitledElement { Title: { } } titledElement)
+        if (options.IncludeTitleInBreadcrumbs && element is ITitledElement { Title: { } title })
         {
-            data.Add(prefix + nameof(titledElement.Title), titledElement.Title);
+            data.Add(prefix + nameof(ITitledElement.Title), title);
         }
 
-        if (options.IncludeTextInBreadcrumbs && element is IText { Text: { } } textElement)
+        if (options.IncludeTextInBreadcrumbs && element is IText { Text: { } text })
         {
-            data.Add(prefix + nameof(textElement.Text), textElement.Text);
+            data.Add(prefix + nameof(IText.Text), text);
         }
     }
 
--- a/test/Sentry.Maui.Tests/MauiEventsBinderTests.Element.cs
+++ b/test/Sentry.Maui.Tests/MauiEventsBinderTests.Element.cs
@@ -42,6 +42,7 @@ public partial class MauiEventsBinderTests
         _fixture.Binder.HandleElementEvents(parent);
 
         var child = new MockElement("child");
+        child.AutomationId = "child-automation-id";
 
         parent.RaiseEvent(eventName, new ElementEventArgs(child));
         Assert.Single(_fixture.Scope.Breadcrumbs); // Sanity check
--- a/test/Sentry.Maui.Tests/MauiEventsBinderTests.Element.cs
+++ b/test/Sentry.Maui.Tests/MauiEventsBinderTests.Element.cs
@@ -15,6 +15,7 @@ public partial class MauiEventsBinderTests
         _fixture.Binder.HandleElementEvents(parent);
 
         var child = new MockElement("child");
+        child.AutomationId = "child-automation-id";
 
         // Act
         parent.RaiseEvent(eventName, new ElementEventArgs(child));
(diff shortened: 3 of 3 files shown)


jamescrosswell:
Thanks for the PR @albyrock87 ! 

Generally looks good. If you mark it ready for review, once it compiles and passes CI, we can take another look at this and get it merged in.

albyrock87:
The PR is ready for review, I have no idea why it started as a draft.

jamescrosswell:
> The PR is ready for review, I have no idea why it started as a draft.

I put it back into draft, since it doesn't yet pass any of the CI checks. It looks like a compilation error - I'm surprised it builds for you locally.

albyrock87:
Something went definitely wrong with my commit :/
I've fixed the compilation error and rebased on top of `main`.

src/Sentry.Maui/Internal/Extensions.cs
```diff
@@ -53,21 +53,26 @@ public static void AddElementInfo(this IDictionary<string, string> data,
         // The element ID seems to be mostly useless noise
         //data.Add(prefix + nameof(element.Id), element.Id.ToString());
 
-        if (element.StyleId != null)
+        if (element.AutomationId is { } automationId)
```
Flash0ver, line 56: question: test

Is there a (reasonably) easy way to cover/assert this with a new unit test or extending an existing unit test?
(I'm not too familiar with MAUI yet ... I'm not too confident with impact analysis yet)

albyrock87: @Flash0ver thanks for pointing it out, I've added it to the existing `Element` binder tests.

test/Sentry.Maui.Tests/MauiEventsBinderTests.Element.cs
```diff
@@ -15,6 +15,7 @@ public void Element_ChildEvents_AddsBreadcrumb(string eventName)
         _fixture.Binder.HandleElementEvents(parent);
 
         var child = new MockElement("child");
+        child.AutomationId = "child-automation-id";
```
Flash0ver, line 18: please also add this line to the test-case below:
- `Element_UnbindChildEvents_DoesNotAddBreadcrumb`
   ```diff
    var child = new MockElement("child");
   +child.AutomationId = "child-automation-id";
   ```

to keep the `Arrange` parts of both related tests in sync

src/Sentry.Maui/Internal/Extensions.cs
Flash0ver: thanks:
nit-pick follow up to keep the test-cases in sync:

https://github.com/getsentry/sentry-dotnet/pull/4248/files#r2137338808

Flash0ver, line 56: question @jamescrosswell @albyrock87

Do we need an opt-in Option for this as well?

See
- `SentryMauiOptions.IncludeTextInBreadcrumbs`
- `SentryMauiOptions.IncludeTitleInBreadcrumbs`

Do I understand it right that we **do not** need an opt-in Option for the `AutomationId`,
because it won't include PII,
whereas `Text` and `Title` may contain PII and therefore need the explicit opt-in?

(it's been awhile since I used `AutomationId` in XAML-based App-Frameworks ... I'm not sure what the common usage patterns are)

albyrock87: That's my understanding: the `AutomationId` is a developer/QA-friendly identifier used mainly by QA automation software, so I highly doubt it would contain PII.

https://learn.microsoft.com/en-us/dotnet/api/microsoft.maui.controls.element.automationid?view=net-maui-9.0

jamescrosswell: Doesn't sound like a PII risk then 👍🏻

jamescrosswell (approved):
Thanks again for the PR @albyrock87 !

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 
 - Rename MemoryInfo.AllocatedBytes to MemoryInfo.TotalAllocatedBytes ([#4243](https://github.com/getsentry/sentry-dotnet/pull/4243))
 - Replace libcurl with .NET HttpClient for sentry-native ([#4222](https://github.com/getsentry/sentry-dotnet/pull/4222))
+- Add .NET MAUI `AutomationId` element information to breadcrumbs ([#4248](https://github.com/getsentry/sentry-dotnet/pull/4248))
```
KnapSac, line 9: @jamescrosswell it seems like this got merged in the wrong place (it got added to the release notes for `5.10`, instead of the `Unreleased` section). I noticed because I wanted to make sure my PR modified the changelog correctly.

It might be good to think about a way to better automate this process. Perhaps there is a way to have a bot automatically update the changelog, instead of having to do it manually. This way, the modifications always happen in the correct place, and not accidentally in already released sections.

KnapSac: Opened #4273 to fix this.

jamescrosswell: Thanks @KnapSac - much appreciated!

> It might be good to think about a way to better automate this process. Perhaps there is a way to have a bot automatically update the changelog, instead of having to do it manually. This way, the modifications always happen in the correct place, and not accidentally in already released sections.

Yeah it's a bit tricky. The PR that updates the changelog in the main branch happens when making a release. We'd need something that went through and automatically updates any other open PRs... I imaging that could be done - just hard to find the time to look at it (given all the other issues on the backlog).

albyrock87: I'm using `release-drafter` on my project and I'm very happy with it.
https://github.com/nalu-development/nalu/blob/main/.github/workflows/release-drafter.yml

I think it's much better than having a `CHANGELOG.md` file.
If we still want to have it, maybe some kind of automated process could pull the releases information from GitHub releases.

KnapSac: As a safe guard, maybe the GH action which checks for a changelog entry could also check that any modifications appear in the `Unreleased` section, and not in an already released section. It seems like that wouldn't be too hard to implement, but I'm not sure.

jamescrosswell: > I'm using release-drafter on my project and I'm very happy with it.

That might be an option if this was an isolated repo. The sentry-dotnet repo is one of over 650 in the getsentry org though so there's a standard release process used  across all of the SDKs... it does stuff like updating the latest release version to be referenced in docs and download links etc. 

> As a safe guard, maybe the GH action which checks for a changelog entry could also check that any modifications appear in the Unreleased section, and not in an already released section. It seems like that wouldn't be too hard to implement, but I'm not sure.

There is a GH action to check that already but some new commits have to be pushed to a PR for it to trigger.

The problem occurs when this sequence of events occurs:

PR1: push changes (changelog entries correctly in unreleased)
   -> All CI checks pass
PR2: push changes (changelog entries correctly in unreleased)
   -> All CI checks pass
PR1: approve and merge
Create New Release (includes PR1)
   -> Change log updated (Unreleased -> New Release)
   -> CI checks don't run again on open PRs (i.e. PR2)
PR2: approve and merge
   -> The change log entry is now unfortunately part of New Release

albyrock87: What about having an UNRELEASED.md file?

jamescrosswell: I'm not sure - I haven't spent time looking at it to see if that might work.

Releases happen here basically:
https://github.com/getsentry/sentry-dotnet/blob/aea1313c6434efdf7d0c80aad5a97e72fd5a95ef/.github/workflows/release.yml#L33-L34

That's a shared github action used by sentry-dotnet and the other repos, using Craft and tying into other things that are required to release stuff for Sentry.

PR #4254, opened 2025-06-05
fix: ExtraData not captured for Breadcrumbs in MauiEventsBinder

### LINKED ISSUE #4252: BreadcrumbEvent.ExtraData not captured in MauiEventsBinder
@jamescrosswell I'm sorry for this late review but I just realized there's a huge bug here.

What about `breadcrumb.ExtraData`? I think you completely missed it.

Also I don't find the `ExtraData` type very convenient.

I think that this could be more convenient:

```
public record BreadcrumbEvent(
    object? Sender,
    string EventName,
    params (string Key, string Value)[] ExtraData
);
```

_Originally posted by @albyrock87 in https://github.com/getsentry/sentry-dotnet/pull/3997#discussion_r2127055110_

jamescrosswell:
Resolves #4252:
- https://github.com/getsentry/sentry-dotnet/issues/4252


test/Sentry.Maui.Tests/ApiApprovalTests.Run.DotNet8_0.verified.txt
```diff
@@ -15,12 +15,12 @@ namespace Sentry.Maui
     {
         public BreadcrumbEvent(object? Sender, string EventName, [System.Runtime.CompilerServices.TupleElementNames(new string[] {
                 "Key",
-                "Value"})] params System.Collections.Generic.IEnumerable<System.ValueTuple<string, string>>[] ExtraData) { }
+                "Value"})] params System.ValueTuple<string, string>[] ExtraData) { }
```
jamescrosswell: Technically we're changing the contract here. I think this might be OK since this functionality has been in the wild for a really short time and there are probably only a handful of people meddling with their own binders.

If we wanted to be really strict about it, we could either stick with the old params declaration (which is pretty messy) or try to create some `BreadcrumbEvent2` that uses the new params type and could be converted implicitly/explicitly to `BreadcrumbEvent`... quite a lot of work for very little reward though IMO.

test/Sentry.Maui.Tests/ApiApprovalTests.Run.DotNet8_0.verified.txt
jamescrosswell, line 1: Technically a breaking change... I added this (obsolete) constructor for anyone who was creating a BreadcrumbEvent and passing ExtraData using the old params syntax:
https://github.com/getsentry/sentry-dotnet/blob/ddfcd8bfb49526005ba2ddb04154224c30ada5aa/src/Sentry.Maui/BreadcrumbEvent.cs#L43-L49

So I don't think folks would have to change their code when recompiling (except to get rid of the `Obsolete` warning). It's not binary compatible with the old `BreadcrumbEvent` but I think this is an OK compromise. This stuff has only been in the wild for a few weeks so I doubt anyone is doing anything too fancy with this yet. 

@albyrock87 what are your thoughts?

albyrock87: I would have done the same :)

Flash0ver: I believe it is also a source breaking change.

Previously we could have written:
```CSharp
var instance = new BreadcrumbEvent(Sender: null, EventName: "", new[] { ("", "") }, new[] { ("", "") })
{
  Sender = new object(),
};
var copy = instance with
{
  EventName = "Event-Name",
};
```

which now breaks for four reasons:
- the `.ctor` parameter names have changed from _PascalCase_ to (proper) _camelCase_
- the `[ParamArrayAttribute]` syntax for `IEnumerable` of "string-string-ValueTuples" is no longer possible
- object-initializer is not longer possible because `init`-accessor are gone
- the record's `with`er is no longer possible `init`-accessors are gone

This is why I personally dislike Primary-Constructors in general, but in particular on record-structs and record-classes ... they make it easy to introduce unobvious breaking changes and - in the case of records - produce unidiomatic C# code. (sorry for the rant ... I apparently had to vent 😉)

On the other hand, `IEnumerable<(string Key, string Value)>[]`,
a collection of arrays, that we internally represent as `IReadOnlyDictionary<string, string>?`,
looks a bit misleading ... unless there are MAUI idioms that I'm not familiar with yet ... please correct me if I'm wrong here. Also, since we need to flatten that enumerable-of-arrays, we cause extra heap allocations.

TL;DR: I think fixing this hard-to-use / non-idiomatic / perhaps even misleading API outweighs the impact of the breaking change.

albyrock87: Good catch, I haven't noticed that.
Can't we simply use uppercase names together with `params` keyword?
```
public BreadcrumbEvent( 
     object? Sender, 
     string EventName, 
     params IEnumerable<(string Key, string Value)>[] ExtraData)
        : this(Sender, EventName, ExtraData.SelectMany(e => e).ToArray()) 
```

test/Sentry.Maui.Tests/MauiEventsBinderTests.cs
```diff
@@ -45,4 +45,32 @@ public Fixture()
     private readonly Fixture _fixture = new();
 
     // Tests are in partial class files for better organization
+
+    [Fact]
```
Flash0ver, line 11: question(style): keep here or move to separate file

The other tests for this partial type are in other files.
Do we want to create a `MauiEventsBinderTests.Breadcrumb.cs` to keep the pattern?
Or is this not only overkill, but maybe also misleading, as the other tests are grouped in files by MAUI controls and events?

test/Sentry.Maui.Tests/ApiApprovalTests.Run.DotNet8_0.verified.txt
Flash0ver: Question 1: `record class` vs `ordinary class`:
Do we want to offer the `init`-accessors, or are they OK to be removed since they don't add too much value as we have constructors that set all required members.
If so, because without `init`-accessors the `with`-syntax of `records` can only really be used to create a copy, but not change any values actually, I wonder if it's better to change this type from a `record class` to an ordinary `class`, perhaps even `sealed class`.
Well, we would lose the generated `IEquatable<T>` implementation, `==`/`!=` operators and the automatically overridden `int GetHashCode()` from a `record` .. but one could argue that these are not too correct anyway, since Equality and GetHashCode uses the `string Key, string Value)[] ExtraData` as a reference, and does not do a `.SequenceEqual()` against the left and right ... so for "deep"-Equality, we would like to provide our own implementation and overrides anyway.

Question 2: general guidelines when using `record class` and `record struct`:
Now that I think a lot about `records` vs "ordinary" classes and structs, I now understand why the .NET BCL team basically banned public record-types.
And now I am wondering if we should also consider effectively banning `public` record-types, as they may come with some "implicitness" that we may not necessarily expect / becomes hard to change later.

Question 3: docs:
Would that be worth a note in the `CHANGELOG.md`? (e.g. an indented list item to the new entry)

src/Sentry.Maui/BreadcrumbEvent.cs
```diff
@@ -0,0 +1,50 @@
+namespace Sentry.Maui;
+
+/// <summary>
+/// Argument to the OnBreadcrumbCreateCallback
+/// </summary>
+public record BreadcrumbEvent
```
Flash0ver: Context: https://github.com/getsentry/sentry-dotnet/pull/4254/files#r2128517126
Question: `record class` vs `ordinary class`:

Do we want to offer the `init`-accessors, or are they OK to be removed since they don't add too much value as we have constructors that set all required members.

If so, because without `init`-accessors the `with`-syntax of `records` can only really be used to create a copy, but not change any values actually, I wonder if it's better to change this type from a `record class` to an ordinary `class`, perhaps even `sealed class`.

Well, we would lose the generated `IEquatable<T>` implementation, `==`/`!=` operators and the automatically overridden `int GetHashCode()` from a `record` .. but one could argue that these are not too correct anyway, since Equality and GetHashCode uses the `string Key, string Value)[] ExtraData` as a reference, and does not do a `.SequenceEqual()` against the left and right ... so for "deep"-Equality, we would like to provide our own implementation and overrides anyway.

Flash0ver: Which now leads me to a follow-up question:
The general guidelines of using `record class` and `record struct`:

Now that I think a lot about `records` vs "ordinary" classes and structs, I now understand why the .NET BCL team basically banned public record-types.

And now I am wondering if we should also consider effectively banning `public` record-types from our .NET SDKs, as they are easy to write but also easy to get wrong, and may come with some "implications" that we may not necessarily desire and/or become hard to change later.

test/Sentry.Maui.Tests/ApiApprovalTests.Run.DotNet8_0.verified.txt
Flash0ver: Hmmm ... well, for backwards-compat, yes.
And also, I'm now thinking this should be a general guideline for C#, that `record class` and `record struct` that use the primary-constructor syntax, should have all other constructors use _PascalCase_ parameter names as well for consistency.
(another TODO for my private Roslyn Analyzer project)

Another approach could be:
Since we are doing a breaking change here, one way or another, we could argue over biting the bullet and go even more breaking ... although that does sound kinda wrong when I think that out loud.

Which now leads me to a follow-up question for "records" vs "ordinary classes/structs" in the public API of our NuGet packages:
https://github.com/getsentry/sentry-dotnet/pull/4254/files#r2128592067

Flash0ver: And a question about docs:
Would that be worth a note in the `CHANGELOG.md`? (e.g. an indented list item to the new entry)

jamescrosswell: >  since we need to flatten that enumerable-of-arrays, we cause extra heap allocations.

Only if people use the old constructor though right (the one that's marked as obsolete)? I don't think this is an issue then. 

> Would that be worth a note in the CHANGELOG.md? (e.g. an indented list item to the new entry)

Yeah I think we just ignore backward compatibility for this particular change. The original params declaration was pretty funky (every param argument had to be passed in as it's own array/enumerable). It may even have been a typo.

As you suggest, we could just put a note in the changelog saying we broke something that was already broken.

test/Sentry.Maui.Tests/MauiEventsBinderTests.cs
jamescrosswell: What it's testing is this method from the `MauiEventsBinder`:
https://github.com/getsentry/sentry-dotnet/blob/94ee8f6faebae3849b293dace84bd3213daeea4c/src/Sentry.Maui/Internal/MauiEventsBinder.cs#L128

So I think it should be in the `MauiEventsBinderTests` class. That's how most of our tests are organised - they go in a `[ClassBeingTested]Tests.cs` file. 

I see why Allan organised a bunch of these in partial class files as it would be one very large file otherwise. I might refactor the `MauiButtonEventsBinder`, `MauiImageButtonEventsBinder` and `MauiGestureRecognizerEventsBinder` into different test classes though to follow the rule above.

src/Sentry.Maui/BreadcrumbEvent.cs
jamescrosswell: > Do we want to offer the init-accessors, or are they OK to be removed since they don't add too much value as we have constructors that set all required members.

I think the constructor we provide is easier (particularly to if setting extradata), so they don't seem necessary.  

> now understand why the .NET BCL team basically banned public record-types

I'm OK changing to a class. It would be a breaking change but we're considering making a breaking change here already, so if we're going to do it then we may as well do it properly. 

> Well, we would lose the generated IEquatable<T> implementation, ==/!= operators and the automatically overridden int GetHashCode() from a record 

We're not losing anything we actually need though.

test/Sentry.Maui.Tests/MauiEventsBinderTests.cs
Flash0ver: I like the change a lot ... it makes the tests more "SRP", especially considering
```diff
Binder = new MauiEventsBinder(
  Hub,
  options,
- [
-   new MauiButtonEventsBinder(),
-   new MauiImageButtonEventsBinder(),
-   new MauiGestureRecognizerEventsBinder()
- ]
+ elementEventBinders
); 
```

test/Sentry.Maui.Tests/ApiApprovalTests.Run.DotNet8_0.verified.txt
Flash0ver: I think this breaking change is acceptable, perhaps even preferable, as it's fixing the design of the broken `record` type (pun intended) `BreadcrumbEvent`.

With the learning to be careful when using `public record` types.
Do we want to add a note on the caveats of `record`-types in our published publish API surface area?
I'm happy to add a small section to our `CONTRIBUTING.md`.

Also, let's add a nested list item to the `CHANGELOG.md`.
Perhaps something like
```diff
 - ExtraData not captured for Breadcrumbs in MauiEventsBinder ([#4254](https://github.com/getsentry/sentry-dotnet/pull/4254))
+  - NOTE: Required breaking changes to the public API of `Sentry.Maui.BreadcrumbEvent`, while keeping an _Obsolete_ constructor for backward compatibility.
```
or similar?

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@
 
 - InvalidCastException in SentrySpanProcessor when using the Sentry.OpenTelemetry integration ([#4245](https://github.com/getsentry/sentry-dotnet/pull/4245))
 - Fix InApp Exclude for frames without Module by checking against frame's Package ([#4236](https://github.com/getsentry/sentry-dotnet/pull/4236))
+- ExtraData not captured for Breadcrumbs in MauiEventsBinder ([#4254](https://github.com/getsentry/sentry-dotnet/pull/4254))
```
```suggestion
- ExtraData not captured for Breadcrumbs in MauiEventsBinder ([#4254](https://github.com/getsentry/sentry-dotnet/pull/4254))
    - NOTE: Required breaking changes to the public API of `Sentry.Maui.BreadcrumbEvent`, while keeping an _Obsolete_ constructor for backward compatibility.
```

test/Sentry.Maui.Tests/ApiApprovalTests.Run.DotNet8_0.verified.txt
jamescrosswell: Done 👍🏻 

> I'm happy to add a small section to our CONTRIBUTING.md.

@Flash0ver that would be great. I think we do that in a separate PR though.

src/Sentry.Maui/BreadcrumbEvent.cs
```diff
@@ -0,0 +1,50 @@
...
+    /// <summary>
+    /// The event name (e.g. "Tapped", "Swiped", etc.)
+    /// </summary>
+    public string EventName { get; }
+
+    /// <summary>
+    /// Any extra data to be included in the breadcrumb. This would typically be event specific information (for example
+    /// it could include the X, Y coordinates of a tap event).
+    /// </summary>
+    public (string Key, string Value)[] ExtraData { get; }
```
bruno-garcia: why did we end up with a tuple array?
Is this preferred these days over IList or ICollection or something?
We don't want to allow one to grow it (no Add?)

src/Sentry.Maui/Internal/MauiEventsBinder.cs
```diff
@@ -132,7 +132,14 @@ internal void OnBreadcrumbCreateCallback(BreadcrumbEvent breadcrumb)
             breadcrumb.Sender,
             breadcrumb.EventName,
             UserType,
-            UserActionCategory
+            UserActionCategory,
+            extra =>
+            {
+                foreach (var (key, value) in breadcrumb.ExtraData)
+                {
+                    extra[key] = value;
```
bruno-garcia, line 140: should we keep a map instead then? it's odd we have an array but that'll be converted into a map later

src/Sentry.Maui/BreadcrumbEvent.cs
jamescrosswell: Originally `BreadcrumbEvent` was a record (not a class)... so it's not supposed to be mutable. All of the properties of this class are getter only can only be set via the constructor.

It's basically used to communicate the details to create a breadcrumb from the UI events (like tap or drag or whatever) to a callback like this:
https://github.com/getsentry/sentry-dotnet/blob/94ee8f6faebae3849b293dace84bd3213daeea4c/src/Sentry.Maui/Internal/MauiEventsBinder.cs#L128-L144

If we don't like the Tuple array, other options would be `KeyValuePair[]` or `Span<KeyValuePair>`... or just `Dictionary<string,string>`?

src/Sentry.Maui/Internal/MauiEventsBinder.cs
jamescrosswell: By Map you mean Dictionary? We could do that. Regardless, we're going to have to loop over the elements in the `BreadcrumbEvent.ExtraData` and copy them to the dictionary, unless we refactor this:
https://github.com/getsentry/sentry-dotnet/blob/d47d66f4f96be76c4a4716eee403b28ef551df80/src/Sentry.Maui/Internal/Extensions.cs#L14-L33

Potentially that method could take a `Dictionary<string,string>` and then the method could do:
```csharp
var data = new Dictionary<string, string>(addExtraData); 
```

Which may be more efficient.

src/Sentry.Maui/BreadcrumbEvent.cs
jamescrosswell: Ah, I see what you're saying... we don't need a specific concrete type here... we could leave it flexible (like `ICollection`).

I can do that 👍🏻

src/Sentry.Maui/Internal/MauiEventsBinder.cs
albyrock87: Then the most generic thing would be `IEnumerable<KeyValuePair<string, string>>? = null`.

jamescrosswell: Yeah, it's not quite as nice to use/call as with tuples (not sure why `KeyValuePair<TKey, TValue>` doesn't just provide an implicit conversion from `(TKey key, TValue value)`) but probably the best choice for internal storage. 

I'll add an overload that so that calling methods can pass tuples or kvps.

src/Sentry.Maui/BreadcrumbEvent.cs
jamescrosswell: OK, changed to:
https://github.com/getsentry/sentry-dotnet/blob/dbf9985f8e1afe1e636bb9673f54b8578723950b/src/Sentry.Maui/BreadcrumbEvent.cs#L22

I added a constructor overload so that people can pass tuples or KeyValuePairs but ultimately stored internally like this.

test/Sentry.Maui.Tests/ApiApprovalTests.Run.DotNet8_0.verified.txt
Flash0ver: Sounds good ... I'll take care of it.

src/Sentry.Maui/BreadcrumbEvent.cs
```diff
@@ -0,0 +1,71 @@
...
+    /// Creates a new BreadcrumbEvent
+    /// </summary>
+    public BreadcrumbEvent(
+        object? sender,
+        string eventName,
+        params IEnumerable<KeyValuePair<string, string>> extraData)
+    {
+        Sender = sender;
+        EventName = eventName;
+        ExtraData = extraData;
```
Flash0ver, line 42: question: Validate `null` keys?

I wonder if we need to validate against `null` keys in the `extraData` collection.

If the user has `#nullable enable`, they'll get a warning when constructing `BreadcrumbEvent` with `null` references.
But the user may have `#nullable disable warnings` and then not see a warning when setting a `null` key, or suppress it explicitly by passing `null!` or `default!`.

A `null` key would result in a `System.ArgumentNullException` when - in our code - we iterate through `breadcrumb.ExtraData` and set or overwrite the `Data` in the Breadcrumb via the `Dictionary`-callback
```CSharp
extra[key] = value;
```
potentially causing an Exception in our invocation of `OnBreadcrumbCreateCallback`.

If we would throw an `ArgumentException` in this `.ctor` instead already, the exception would occur in the user's code instead ... if I'm reading that right ... indicating that they are constructing our type incorrectly as early as possible.

TL;DL; to move the potential exception from the later Sentry-space, to the earlier User-space.

Alternatively,
when iterating through `breadcrumb.ExtraData` and accessing the Data-Dictionary via `extra[key] = value;` in `OnBreadcrumbCreateCallback`,
we could before accessing the dictionary check if the `key is null`, and then discard the Extra-Data-Item and instead Log a Warning via the Diagnostic Logger to still indicate to the user that a mistake was made and nullable annotations not respected.

src/Sentry.Maui/BreadcrumbEvent.cs
Flash0ver: Good catch. We intended this type to be immutable. But a `System.Array` is not immutable.

src/Sentry.Maui/BreadcrumbEvent.cs
Flash0ver: Another alternative is that we do this in a follow-up issue/changeset ... as I can imagine that we have more occurrences of `null`-keys from User-code propagating into SDK-code (or null references in general). Which would perhaps be a candidate for a `vNext-Major` release as we may introduce more `ArgumentNullExceptions` in `public` members that incur behavioral breaking changes.

I'm happy to open a corresponding issue if we think this is something to look into in a follow-up in large.

jamescrosswell: It sounds like a theoretical problem. SDK users would almost have to be deliberately trying to create an exception for this to happen.

Whilst we _could_ go through the codebase trying to make it impervious to almost deliberate misuse like this, IMO  it's hard to prioritise doing that vs fixing some bugs that users are actually running into (and have told us about). I'd argue we leave it as is. If one of our SDK users reports an `ArgumentNullException` when using a custom `BreadcrumbEvent`, we can surface for discussion again.

Flash0ver: Sounds great

src/Sentry.Maui/BreadcrumbEvent.cs
```diff
@@ -0,0 +1,71 @@
...
+    }
+
+    /// <summary>
+    /// Creates a new BreadcrumbEvent
+    /// </summary>
+    public BreadcrumbEvent(
+        object? sender,
+        string eventName,
+        params IEnumerable<(string key, string value)> extraData) : this(sender, eventName, extraData.Select(
```
Flash0ver, line 51: final comment: deferred execution

The `Select` assigned to `ExtraData` will have deferred execution.
Is this expected / intended?

On the other hand, since we accept `IEnumerable` in the `.ctor`s, the user could pass `IEnumerable`s with deferred execution anyway.

On second thought ... I believe this is great as it is.
Also considering we're enumerating through it only a single time, so materializing the `IEnumerable` beforehand would actually be slightly slower and allocate more memory.

Flash0ver (approved):

Issue #4255, opened 2025-06-05
.NET SDK version


KnapSac:
The current .NET SDK used is `9.0.203` as specified in `global.json`:
https://github.com/getsentry/sentry-dotnet/blob/723c84e6fe8a642be7ab2d40caf401bb6d38a8e3/global.json#L1-L8
Because `rollForward` is disabled, even `9.0.204` is not supported. Are there any plans to bump the SDK version? Or perhaps to allow rolling forward to the latest patch?

The reason I ask is that for some reason my system keeps automatically updating `9.0.203` to `9.0.204`, which means that I constantly have to uninstall `9.0.204` and install `9.0.203` (and then also restore the workloads) to be able to compile the code locally. This might very well be an issue with my setup (I have `9.0.300` installed globally), but I just thought I'd ask.

Flash0ver:
Thanks for the feedback.
RIght ... some IDEs do manage their own installation of the .NET SDK.

I believe there is a discussion emerging about
- reproducible / deterministic builds
  - both in _CI/CD_ and on local dev machines
- local dev experience / convenience
- .NET SDK update strategy
- .NET Workload version
  - in particular when it comes to **.NET MAUI**

I believe we had this discussion before, but I'm uncertain right now about the outcome.
And the _.NET MAUI_ argument may be tricky.

Let's also manifest the outcome in `CONTRIBUTING.md`.

jamescrosswell:
> Or perhaps to allow rolling forward to the latest patch

We've been progressively pinning everything to try to make CI/CD as deterministic as possible. If we allow roll forward, things that build and work one day may not build/work the next day, if Microsoft releases new versions of their workloads. This has occurred on multiple occasions in the past, so we basically decided to pin everywhere. 

In terms of why we landed on 9.0.203, from memory we needed that to address an issue in the runtime but otherwise we wanted to use the lowest common denominator (so people didn't have to be using the latest version of dotnet in order to work with the Sentry SDK). 

It's weird that you can't have multiple versions of dotnet installed on your machine. What's your development environment look like? How are you installing new versions of dotnet?

KnapSac:
> We've been progressively pinning everything to try to make CI/CD as deterministic as possible.

That makes sense.

> It's weird that you can't have multiple versions of dotnet installed on your machine. What's your development environment look like? How are you installing new versions of dotnet?

[...]

PR #4259, opened 2025-06-05
Bump .NET SDK to 9.0.204

KnapSac:
See #4255.

### DIFF
--- a/global.json
+++ b/global.json
@@ -1,7 +1,7 @@
 {
   "sdk": {
-    "version": "9.0.203",
-    "workloadVersion": "9.0.203",
+    "version": "9.0.204",
+    "workloadVersion": "9.0.204",
     "rollForward": "disable",
     "allowPrerelease": false
   }
--- a/.github/actions/environment/action.yml
+++ b/.github/actions/environment/action.yml
@@ -67,7 +67,7 @@ runs:
       with:
         dotnet-version: |
           8.0.x
-          9.0.203
+          9.0.204
 
     - name: Install .NET Workloads
       shell: bash


KnapSac:
I can't get this to work locally. Running `dotnet workload restore` after updating `global.json` attempts to install the workloads, but for some reason it fails and tries to rollback. Reinstalling `9.0.204` doesn't help. Removing the `workloadVersion` from `global.json` does allow restoring, but I'm not sure if that's ideal, as I presume it was added to pin to a specific workload version.

It does seem like the issues are mostly related to Maui and Android. I'm not familiar with Maui, so I don't really know what the best approach might be.

jamescrosswell:
> Removing the workloadVersion from global.json does allow restoring, but I'm not sure if that's ideal, as I presume it was added to pin to a specific workload version.

Exactly. Without that we're only pinning the core SDK but all of workloads still roll forward to latests. 

The workloads are the ones that usually break things - the iOS/Android workloads in particular. Often these require corresponding dependencies to be bumped (e.g. the Xcode version or a new Android SDK version). Sometimes there's other weird stuff going on.

I'll try to find some time to take a look. MAUI is a bit fiddly (understatement of the year).

jamescrosswell:
@KnapSac is this what you were seeing when trying to restore the workloads locally?
> Workload update failed: '"' is invalid after a single JSON value.

If so, then I ran into the same issue. 

I didn't have any issues when bumping to `9.0.300`, so maybe we just go all the way up to that. It looks like it includes some security patches, so not bad to have anyway.

CHANGELOG.md
```diff
@@ -1,5 +1,13 @@
 # Changelog
 
+## Unreleased
+
+### Dependencies
+
+- Bump the version of the .NET SDK that we use from 9.0.203 to 9.0.300 ([#4259](https://github.com/getsentry/sentry-dotnet/pull/4259))
+  - Note that this also required we bump various Java depdendencies (since version 9.0.300 of the Android workload requires newer versions of the these)
```
KnapSac:
```suggestion
  - Note that this also required we bump various Java dependencies (since version 9.0.300 of the Android workload requires newer versions of the these)
```

KnapSac:
Yup, that was at least one of the errors I was getting. Turns out that some process was messing up the `global.json` in the install state folder. Took me a while to find that.

Going to `9.0.300` is even better, the only downside is that people using Visual Studio to download and manage their SDKs will have to upgrade to at least 17.14, but that has been out for weeks now.

KnapSac:
The CI failure doesn't repro locally, but that's probably because I'm on Windows. It might be fixed by bumping the Verify version, but I'm not sure.

jamescrosswell:
> The CI failure doesn't repro locally, but that's probably because I'm on Windows. It might be fixed by bumping the Verify version, but I'm not sure.

I also can't reproduce it (running on a Mac), which is super annoying. It means that version of Verify can potentially work on macOS - it must be some other difference between the CI runner and my local dev environment 😞 I'll try bumping the version of Verify anyway, but not expecting miracles. 

@SimonCropp just in case, have you seen any issues with Argon in version 9.0.300 of the .NET SDK? Since bumping in this PR, we're seeing a [whole bunch of these](https://github.com/getsentry/sentry-dotnet/actions/runs/15548966645/attempts/1?pr=4259) in our Verify tests (seemingly only for net4.8 tests running on Mono in CI):
```
System.MissingMethodException : Method not found: void Argon.JsonWriter.WriteValue(System.ReadOnlySpan`1<char>)
```

Edit: After upgrading to the latest version of Verify we get a different error, which probably reflects the true cause of the error?

```
System.MissingMethodException : Method not found: System.Nullable`1<System.ValueTuple`2<string, string>> Sentry.Extensions.Logging.EventIdExtensions.ToTupleOrNull(Microsoft.Extensions.Logging.EventId)
  at (wrapper managed-to-native) System.Reflection.RuntimeMethodInfo.InternalInvoke(System.Reflection.RuntimeMethodInfo,object,object[],System.Exception&)
  at System.Reflection.RuntimeMethodInfo.Invoke (System.Object obj, System.Reflection.BindingFlags invokeAttr, System.Reflection.Binder binder, System.Object[] parameters, System.Globalization.CultureInfo culture) [0x0006a] in <d636f104d58046fd9b195699bcb1a744>:0 
```

jamescrosswell:
Aha... OK I can reproduce this on my local development machine. I had to:
- Upgrade mono from 6.12.0 to 6.14.1 (CI installs this version)
- Run `git clean -dfx` 
- Build and test the solution from the command line (verbatim what's run on GitHub Actions)... the tests pass when run from Rider for some reason... that may be related to Rider trying to use a different version of mono 😕

test/Sentry.AspNetCore.TestUtils/Sentry.AspNetCore.TestUtils.csproj
```diff
@@ -51,8 +51,8 @@
   <ItemGroup Condition="$(TargetFramework) == 'net9.0'">
     <PackageReference Include="Microsoft.AspNetCore.Mvc.Versioning" Version="5.1.0" />
     <PackageReference Include="Microsoft.AspNetCore.TestHost" Version="9.0.0" />
-    <PackageReference Include="Verify.AspNetCore" Version="4.0.0" />
-    <PackageReference Include="Verify.Http" Version="6.4.1" />
+    <PackageReference Include="Verify.AspNetCore" Version="4.2.0" />
+    <PackageReference Include="Verify.Http" Version="6.5.0" />
```
Flash0ver, line 57: question: Could/Should we update some `Verify.` versions for the above `<ItemGroup Condition="$(TargetFramework) == 'net8.0'">` as well?

Do I get that right?
- Both latest `Verify.` versions target `net9.0` only, so we need `net8.0` compatible versions above.

For above `Condition="$(TargetFramework) == 'net8.0'"`
may we update `Verify.AspNetCore` from `3.4.1` to `4.0.0`?
- `3.4.1` targets `net7.0`
- `4.0.0` targets `net8.0`
- `4.2.0` targets `net9.0`

For above `Condition="$(TargetFramework) == 'net8.0'"`
may we update `Verify.Http` from `4.3.2` to `6.4.1`?
- `4.3.2` targets `net6.0` and `net48`
- `6.4.1` targets `net8.0`
- `6.5.0` targets `net9.0`

global.json
```diff
@@ -1,7 +1,7 @@
 {
   "sdk": {
-    "version": "9.0.203",
-    "workloadVersion": "9.0.203",
+    "version": "9.0.300",
```
Flash0ver, line 3: questions
- Will we run into the same issue again once the next _patch version_ of _feature band_ `3` is released?
  - see #4255
- If so, is this an indicator that we should we keep bumping to the latest .NET SDK in the future?
  - Would this be automatable?
    - Perhaps partially (`global.json`, `.github/**/*.yml` and `CHANGELOG.md`)
    - `Xamarin.AndroidX.*` may be not as straightforward

SimonCropp:
@jamescrosswell nope not aware of any issue with Argon in version 9.0.300 of the .NET SDK. that is what the majority of my repos are running on. if u can supply a repro, i can take a look at it

global.json
jamescrosswell: > Will we run into the same issue again once the next patch version of feature band 3 is released?

We frequently run into issues when new versions of .NET are released. They are, thus far, each unique... some easier than others to fix. This is why we pin the version of .NET and the workloads in the repo (it means we can still make PRs and do other work while we're trying to work out how to upgrade to the latest set of "surprises"). 

> Would this be automatable?

We definitely don't want to turn on roll forward (automatic updates) in the global.json file. Before we were pinning the workload/sdk version in global.json, I was spending 1-2 days per week, consistently, fluffing around with this stuff. 

Potentially we could try to automatically create a PR that bumps to the latest version of .NET, like we do with other dependencies. That would give us a way to upgrade regularly but without breaking the main branch and halting all other work. It could be a massive time sink though. It's hard to say whether we're better off making smaller bumps once every 6-8 weeks or having long periods of focus (where we're working on the SDK instead) and then taking all the pain in one hit occasionally (like now). 

I'm inclined to stick to what we're doing for the time being - there's just so much other stuff on the backlog without adding what seems to be very meta work around automating something like this.

test/Sentry.AspNetCore.TestUtils/Sentry.AspNetCore.TestUtils.csproj
jamescrosswell: > Could/Should we update some Verify. versions for the above

The only reason I bumped Verify in this PR was to work out whether Verify was part of the problem. To some extent it was since the older version of Verify was hiding the real error but it's looking more like the latest version of .NET is compiling IL that the current version of Mono doesn't know how to handle. 

For this PR then, there's no reason to fiddle with net8.0 specific dependencies. The tests that are failing are all doing so on the net48 (mono) targets.

jamescrosswell:
> @jamescrosswell nope not aware of any issue with Argon in version 9.0.300 of the .NET SDK. that is what the majority of my repos are running on. if u can supply a repro, i can take a look at it

Thanks @SimonCropp . After bumping to the latests version of Verify, it doesn't look like the problem stems from Verify or Argon directly. Some of the reflection in Argon doesn't appear to be working on Mono but we have other tests that aren't using Verify which are failing as well.

It looks more like the latest version of .NET compiles IL that the current version of Mono isn't handling properly. This test, for example (which isn't using Verify) is failing:
https://github.com/getsentry/sentry-dotnet/blob/7513600e2f9a3da153b349ad623447de95f3ef74/test/Sentry.Tests/HttpStatusCodeRangeTests.cs#L51-L64

... with the error:
```
System.MissingMethodException : Method not found: Sentry.HttpStatusCodeRange Sentry.HttpStatusCodeRange.op_Implicit(System.ValueTuple`2<int, int>)
  at (wrapper managed-to-native) System.Reflection.RuntimeMethodInfo.InternalInvoke(System.Reflection.RuntimeMethodInfo,object,object[],System.Exception&)
  at System.Reflection.RuntimeMethodInfo.Invoke (System.Object obj, System.Reflection.BindingFlags invokeAttr, System.Reflection.Binder binder, System.Object[] parameters, System.Globalization.CultureInfo culture) [0x0006a] in <d636f104d58046fd9b195699bcb1a744>:0
```

Which I assume indicates mono is falling over on the implicit operator for that class:
https://github.com/getsentry/sentry-dotnet/blob/7513600e2f9a3da153b349ad623447de95f3ef74/src/Sentry/HttpStatusCodeRange.cs#L48-L52

I'll see if I can put together a minimal repro with that example and submit a bug report in the mono project.

jamescrosswell:
Replaced by #4272

PR #4264, opened 2025-06-07
fix: http integration uses correct parent span id

bruno-garcia:
We create a span for the HTTP operation, but we don't pass that span id as the parent to the outgoing request. 
Resulting in the incorrect parent-child relationship

Changing the implementation didn't break a single test though, which is quite concerning.

I added a test that failed, and with my changed passed. 
Verifying that we indeed use the correct span as parent int he outgoing HTTP request

Before:
image

After:
tested as 5.10.0-alpha.1 on symbol collector
image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Fixes
+
+- The HTTP instrumentation uses the span created for the outgoing request in the sentry-trace header, fixing the parent-child relationship between client and server ([#4264](https://github.com/getsentry/sentry-dotnet/pull/4264))
+
 ### Dependencies
 
 - Bump Native SDK from v0.8.5 to v0.9.0 ([#4260](https://github.com/getsentry/sentry-dotnet/pull/4260))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Changelog
 
-## Unreleased
+## 5.11.0-alpha.1
 
 ### Fixes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Changelog
 
-## 5.11.0-alpha.1
+## Unreleased
 
 ### Fixes
 
--- a/src/Sentry/SentryMessageHandler.cs
+++ b/src/Sentry/SentryMessageHandler.cs
@@ -86,7 +86,7 @@ public abstract class SentryMessageHandler : DelegatingHandler
         var span = ProcessRequest(request, method, url);
         try
         {
-            PropagateTraceHeaders(request, url);
+            PropagateTraceHeaders(request, url, span);
             var response = await base.SendAsync(request, cancellationToken).ConfigureAwait(false);
             HandleResponse(response, span, method, url);
             return response;
@@ -108,7 +108,7 @@ public abstract class SentryMessageHandler : DelegatingHandler
         var span = ProcessRequest(request, method, url);
         try
         {
-            PropagateTraceHeaders(request, url);
+            PropagateTraceHeaders(request, url, span);
             var response = base.Send(request, cancellationToken);
             HandleResponse(response, span, method, url);
             return response;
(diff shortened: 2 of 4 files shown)


CHANGELOG.md
```diff
@@ -1,6 +1,10 @@
 # Changelog
 
-## Unreleased
+## 5.11.0-alpha.1
```
```suggestion
## Unreleased
```

src/Sentry/SentryMessageHandler.cs
```diff
@@ -135,15 +135,17 @@ private void PropagateTraceHeaders(HttpRequestMessage request, string url)
...
     }
 
-    private void AddSentryTraceHeader(HttpRequestMessage request)
+    private void AddSentryTraceHeader(HttpRequestMessage request, ISpan? parentSpan)
     {
         // Set trace header if it hasn't already been set
-        if (!request.Headers.Contains(SentryTraceHeader.HttpHeaderName) && _hub.GetTraceHeader() is { } traceHeader)
+        if (!request.Headers.Contains(SentryTraceHeader.HttpHeaderName) &&
+            // Use the span created by this integration as parent, instead of its own parent
+            (parentSpan?.GetTraceHeader() ?? _hub.GetTraceHeader()) is { } traceHeader)
```
jamescrosswell, line 148: This shouldn't be necessary. The previous behaviour simply got the header from `_hub.GetTraceHeader()`, which should be reading it off the current span. 

This seems like a workaround for the fact that the span representing the HttpRequest hasn't been set correctly as the current span then.

Just looking at the code, I can't see why that would be. All of the machinery to make this happen seems to be in place.

Basically the HttpMessageHandler creates the span here:
https://github.com/getsentry/sentry-dotnet/blob/dc3c0dd39fe5f593e9d43ed0d6fbc7a7b42436ce/src/Sentry/SentryHttpMessageHandler.cs#L70-L73

And the new span gets pushed onto the stack to become the active span here:
https://github.com/getsentry/sentry-dotnet/blob/aebd6a2e711c31087ca77652c31c6a624e22075b/src/Sentry/TransactionTracer.cs#L311-L312

If you've got the code sample that results in the erroneous parent/child relationship, it'd be good to dig into this to try to work out what's going wrong with the active span tracking.

jamescrosswell (changes requested):
Something is off here... see comment below.

src/Sentry/SentryMessageHandler.cs
bruno-garcia: we don't want the http as current span. because otherwise any other span in parallel becomes a child of the http one. I saw this behavior before and it was broken 

the whole thing when using MAUI (or any global mode approach) is totally broken, with this I was able to set a span explicitly and have the parent child work properly for outgoing http regardless of what's explicitly set to the scope (see scope.Span)

bruno-garcia:
regardless, I think taking the header of the http span is the right approach, instead of hoping some magic functions finds the right span (which clearly doesn't in some cases). We know the newly created and internal span is the parent of what comes out of the server.

src/Sentry/SentryMessageHandler.cs
jamescrosswell: Just to summarise a chat that @bruno-garcia and I had...

The issue with not wanting the spans for outgoing HttpRequests to be mixed up with other spans (perhaps being performed in async Tasks) is a bit more nuanced.
 
Currently, the way the MessageHandler creates spans is this:
https://github.com/getsentry/sentry-dotnet/blob/dc3c0dd39fe5f593e9d43ed0d6fbc7a7b42436ce/src/Sentry/SentryHttpMessageHandler.cs#L68-L73

In the most common case then, if there's no `Scope.Transaction ` then no span gets created. If there is a `Scope.Transaction` then a span gets created as a child of the transaction... and implicitly it becomes the "last active span" for the transaction. 

The way Bruno managed to work around that in the symbol collector is by setting `Scope.Span` (not `Scope.Transaction`):
https://github.com/getsentry/symbol-collector/blob/8a0fdb2594c3028f63f180b9b56b1efea1cbce34/src/SymbolCollector.Core/Client.cs#L75

This appears to be completely undocumented. 

1. We should add documentation for this... it's a useful API that allows SDK users to force spans that get automatically instrumented by our integrations to be created on specific transactions without having to set those transactions on the Scope. 
2. It would be fantastic if we had a more intuitive solution to this. That would need some head scratching though (if it is even possible)

jamescrosswell (approved):

PR #4265, opened 2025-06-07
experiment: gcdump reading

bruno-garcia:
What if we read the gcdump, then augment the low memory event with the top offending types

Not sure how much memory would take to do this in-memory in a process that just triggered a memory dump for being out of memory
It probably needs to be done on a separate process. Or only on restart?

Or on the server, at ingestion time. Considering the mess it would be to get this into the SDK (I had to hack a lot to get it to run) it might be better doing this as a stand alone library, and load it during ingestion time in Sentry when processing `gcdump` files.

image

### DIFF
--- a/src/Directory.Build.props
+++ b/src/Directory.Build.props
@@ -4,7 +4,7 @@
 
   <PropertyGroup>
     <!--Generate xml docs for all projects under 'src'-->
-    <GenerateDocumentationFile>true</GenerateDocumentationFile>
+<!--    <GenerateDocumentationFile>true</GenerateDocumentationFile>-->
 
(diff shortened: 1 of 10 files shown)


jamescrosswell:
Looks like you can get at some quite cool functionality via the vendored in code 👍🏻  

If you wanted something quick and dirty, out of process, this might also work:
```
dotnet-gcdump report dump.gcdump | head -n 20
```

That avoids the complexity of vendoring in the code.

On the other hand, if we did vendor in that code above, we might avoid the dependency that we currently have on needing `dotnet-gcdump` installed... might make it easier to use.

PR #4270, opened 2025-06-10
meta: fix default labels of issue templates

Flash0ver:
@stephanie-anderson indicated that the default labels of our issue templates are incorrect

- Platform: use just `.NET` rather than `Platform: .NET`
- Type: "type" labels are _PascalCase_ rather than _camelCase_

#skip-changelog

### DIFF
--- a/.github/ISSUE_TEMPLATE/feature.yml
+++ b/.github/ISSUE_TEMPLATE/feature.yml
@@ -1,6 +1,6 @@
 name: 💡 Feature Request
 description: Tell us about a problem our SDK could solve but doesn't.
-labels: [".NET", "Improvement"]
+labels: [".NET", "Feature"]
 body:
   - type: textarea
     id: problem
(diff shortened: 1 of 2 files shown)


stephanie-anderson (approved):
Perfect, thank you!

.github/ISSUE_TEMPLATE/feature.yml
```diff
@@ -1,6 +1,6 @@
 name: 💡 Feature Request
 description: Tell us about a problem our SDK could solve but doesn't.
-labels: ["Platform: .NET", "enhancement"]
+labels: [".NET", "Feature"]
```
Flash0ver, line 3: note:
- we don't have an `Enhancement` label
- we do have an `Improvement` label

question: Should we use `Improvement` or `Feature` here?

follow-up question: Should we add another _issue template_ that uses `Improvement` as default label?

stephanie-anderson: You can add another one if you want, makes it easier to filter afterwards. But if you keep only one, then `Feature` would be the one

Flash0ver: Thanks for the quick feedback.
I'll leave it for now. If we get a lot of re-labels from `Feature` to `Improvement`, we'll that issue template.

PR #4272, opened 2025-06-11
Bump .NET SDK to 9.0.301

### LINKED ISSUE #4255: .NET SDK version

jamescrosswell:
See #4255

Resolve #4255
Replaces #4259 

> [!IMPORTANT]
> After bumping to SDK version 9.0.300 we're no longer able to test net462 compiled code on macOS or Linux using Mono. This PR includes code to conditionally target netFX in our unit test projects only when running the tests on Windows.
> For further information see:
> - https://github.com/dotnet/runtime/issues/116518


SimonCropp:
noting 9.0.300 has a cve and 9.0.301 is out today

bruno-garcia (approved):

Issue #4267, opened 2025-06-09
Document setting Scope.Span


jamescrosswell:
_Originally posted by @jamescrosswell in https://github.com/getsentry/sentry-dotnet/pull/4264#discussion_r2134899805_

The issue with not wanting the spans for outgoing HttpRequests to be mixed up with other spans (perhaps being performed in async Tasks) is a bit more nuanced.
 
Currently, the way the MessageHandler creates spans is this:
https://github.com/getsentry/sentry-dotnet/blob/dc3c0dd39fe5f593e9d43ed0d6fbc7a7b42436ce/src/Sentry/SentryHttpMessageHandler.cs#L68-L73

In the most common case then, if there's no `Scope.Transaction ` then no span gets created. If there is a `Scope.Transaction` then a span gets created as a child of the transaction... and implicitly it becomes the "last active span" for the transaction. 

The way Bruno managed to work around that in the symbol collector is by setting `Scope.Span` (not `Scope.Transaction`):
https://github.com/getsentry/symbol-collector/blob/8a0fdb2594c3028f63f180b9b56b1efea1cbce34/src/SymbolCollector.Core/Client.cs#L75

This appears to be completely undocumented. 

1. We should add documentation for this... it's a useful API that allows SDK users to force spans that get automatically instrumented by our integrations to be created on specific transactions without having to set those transactions on the Scope. 
2. It would be fantastic if we had a more intuitive solution to this. That would need some head scratching though (if it is even possible)

Issue #1837, opened 2022-08-03
Performance Monitoring for .NET MAUI


bruno-garcia:
### Problem Statement

- [ ] Performance monitoring for screen loading (https://github.com/getsentry/sentry-dotnet/issues/1074), 
	- [ ] Transaction annotations ([see spec](https://develop.sentry.dev/sdk/event-payloads/transaction/#transaction-annotations))
- [ ] [Mobile Vitals](https://docs.sentry.io/product/performance/mobile-vitals/), including app start time instrumentation
- [ ] DB queries as span (EF?)
- [ ] Automatic HTTP span and trace id propagation
- [ ] Transactions for UI events (see [this](https://github.com/getsentry/team-mobile/issues/4))
- [ ] Custom measurements API (Q3 goal)

bruno-garcia:
Kotlin Multiplatform version of it:

* https://github.com/getsentry/sentry-kotlin-multiplatform/pull/155/

de-preut:
Any updates on these tasks? We would really make use of the 'mobile vitals' in our app

bruno-garcia:
Hey @de-preut sorry for the late reply. I'd ask @bitsandfoxes and @jamescrosswell , when they are prioritizing it. I'd love to see it land too ;)

jamescrosswell:
> Having at least `mobile vitals` would indeed be nice.

I'll let you slot it into our priorities board somewhere @bitsandfoxes 👍🏻

jakircevic:
This would be huge for us. Thanks !

bruno-garcia:
came up again from a sales led org

jamescrosswell:
> came up again from a sales led org

I've bumped up the priority on this to "Must Have".

@aritchie probably a good one for you to pick up eventually. I think it probably starts with creating auto transactions for common UI events (like click). I think once we have those, we get tracing on DB requests and outbound HttpRequests for free from the other/separate integrations we have for those things already... but needs testing/verifying.

samykolahi:
Excellent news if this gets pushed forward.

bruno-garcia:
There's also a bunch of new features that we got in native Android and iOS as well as Flutter and React Native. Just MAUI is lagging behind.

This issue, and Session Replay are the highest priority features for MAUI now.

aritchie:
@bruno-garcia @jamescrosswell Thoughts/questions/etc on implementing the features listed. 

>  Including app start time instrumentation

[...]

PR #4274, opened 2025-06-11
Experimental SqlLite integration

jamescrosswell:
Relates to #1837:
- https://github.com/getsentry/sentry-dotnet/issues/1837

This is just mucking around for a couple of hours to see how such an integration might be pieced together. 

image

# Next Steps

If we want to proceed, there is some complexity around the `SqliteDataReader`. This only has an internal constructor, so we'd have to wrap it via a proxy class rather than inheritance....  which would doubtless involve a bunch of boilerplate grunt work.

We'd also have to make decisions about what the hang the spans off. Ideally these DB spans would be children of any active transaction (e.g. the transactions created by the Mvvm integration)... which means storing/retrieving those transaction on the Scope - and as @bruno-garcia recently discovered in the SymbolCollector project, that's fiddly business in GlobalMode (particularly for DB stuff that is likely to be running on a background thread). See https://github.com/getsentry/sentry-dotnet/issues/4267 for details.

### DIFF
--- a/samples/Sentry.Samples.Sqlite/Program.cs
+++ b/samples/Sentry.Samples.Sqlite/Program.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.IO;
 using Microsoft.Data.Sqlite;
 using Sentry.Sqlite;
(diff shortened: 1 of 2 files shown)


jamescrosswell:
Closing this as it was just an experiment. If we do want to go ahead with this, we can do it in a new PR.

PR #4275, opened 2025-06-12
chore: Added IsNet8OrGreater build prop for readablility

### LINKED ISSUE #4203: Conditionally enabling Native support on net8.0 or later is confusing (or maybe broken)
I think there might be a problem with how/when we enable the Native SDK. According to [the docs](https://learn.microsoft.com/en-us/visualstudio/msbuild/property-functions?view=vs-2022#msbuild-targetframework-and-targetplatform-functions) `IsTargetFrameworkCompatible` returns true if the "Candidate" (second argument) is compatible with the "Target" (first argument). 

These seem to be reversed here:
https://github.com/getsentry/sentry-dotnet/blob/084e3b4bf2c043b42ad6b11892a604e027b3d1a5/src/Sentry/Platforms/Native/buildTransitive/Sentry.Native.targets#L23

If they're not reversed, that `IsTargetFrameworkCompatible` function is just plain confusing (and thus error prone) and I think we should just avoid using it.

I had some issues with this elsewhere and added a more reliable (and understandable) way of doing minimum version checks:
https://github.com/getsentry/sentry-dotnet/blob/845de33a4467fb616b17fe0df053fe210b0b2ff7/Directory.Build.props#L112 

We could potentially add `TargetFrameworkIsNet8OrGreater` as well and use that to conditionally enable Native support instead.

jamescrosswell:
Resolves #4203:
- https://github.com/getsentry/sentry-dotnet/issues/4203

#skip-changelog

### DIFF
--- a/src/Sentry/Platforms/Native/buildTransitive/Sentry.Native.targets
+++ b/src/Sentry/Platforms/Native/buildTransitive/Sentry.Native.targets
@@ -20,7 +20,7 @@
 
   <PropertyGroup>
     <!-- net8.0 or greater -->
-    <FrameworkSupportsNative Condition="'$(IsNet8OrGreater)' == 'true' and ('$(OutputType)' == 'Exe' or '$(OutputType)' == 'WinExe')">true</FrameworkSupportsNative>
+    <FrameworkSupportsNative Condition="'$(_SentryIsNet8OrGreater)' == 'true' and ('$(OutputType)' == 'Exe' or '$(OutputType)' == 'WinExe')">true</FrameworkSupportsNative>
     <!-- Make it opt-out -->
     <FrameworkSupportsNative Condition="'$(SentryNative)' == 'false' or '$(SentryNative)' == 'disable'">false</FrameworkSupportsNative>
   </PropertyGroup>
--- a/src/Sentry/buildTransitive/Sentry.props
+++ b/src/Sentry/buildTransitive/Sentry.props
@@ -3,7 +3,7 @@
   <!-- Helpful properties used elsewhere -->
   <PropertyGroup Condition="'$(TargetFramework)' != ''">
     <_SentryTargetFrameworkVersion>$([MSBuild]::GetTargetFrameworkVersion($(TargetFramework)))</_SentryTargetFrameworkVersion>
-    <IsNet8OrGreater>$([MSBuild]::VersionGreaterThanOrEquals($(_SentryTargetFrameworkVersion), 8.0))</IsNet8OrGreater>
-    <IsNet9OrGreater>$([MSBuild]::VersionGreaterThanOrEquals($(_SentryTargetFrameworkVersion), 9.0))</IsNet9OrGreater>
+    <_SentryIsNet8OrGreater>$([MSBuild]::VersionGreaterThanOrEquals($(_SentryTargetFrameworkVersion), 8.0))</_SentryIsNet8OrGreater>
+    <_SentryIsNet9OrGreater>$([MSBuild]::VersionGreaterThanOrEquals($(_SentryTargetFrameworkVersion), 9.0))</_SentryIsNet9OrGreater>
   </PropertyGroup>
 </Project>
--- a/src/Sentry/Platforms/Native/Sentry.Native.targets
+++ b/src/Sentry/Platforms/Native/Sentry.Native.targets
@@ -76,7 +76,7 @@
     </None>
   </ItemGroup>
 
-  <Target Name="CleanNativeSDK" BeforeTargets="CoreClean" Condition="'$(IsNet8OrGreater)' == 'true'">
+  <Target Name="CleanNativeSDK" BeforeTargets="CoreClean" Condition="'$(_SentryIsNet8OrGreater)' == 'true'">
     <Message Text="Inside Custom Clean" Importance="high"/>
     <RemoveDir Directories="$(SentryNativeOutputDirectory)" />
     <RemoveDir Directories="$(SentryNativeSourceDirectory)build" />
@@ -86,7 +86,7 @@
        built already on each native platform and fetched for the final .net build. -->
   <Target Name="_BuildSentryNativeSDK"
     BeforeTargets="DispatchToInnerBuilds;BeforeBuild"
-    Condition="'$(IsNet8OrGreater)' == 'true' and '$(CI)' != 'true'"
+    Condition="'$(_SentryIsNet8OrGreater)' == 'true' and '$(CI)' != 'true'"
     Inputs="$(SentryNativeBuildInputs)"
     Outputs="$(SentryNativeBuildOutputs)">
     <!-- We want a "-Clean" because if the build script changes, previous cmake cache may contain invalid defines. -->
--- a/test/Directory.Build.props
+++ b/test/Directory.Build.props
@@ -14,7 +14,7 @@
 
     <!-- Visual tests need net9.0+ mobile and are flaky when using the headless runner on CI -->
     <EnableMauiDeviceTestVisualRunner>false</EnableMauiDeviceTestVisualRunner>
-    <EnableMauiDeviceTestVisualRunner Condition="'$(IsNet9OrGreater)' == 'true' AND '$(PlatformIsMobile)' == 'true' AND '$(ContinuousIntegrationBuild)' != 'true'">true</EnableMauiDeviceTestVisualRunner>
+    <EnableMauiDeviceTestVisualRunner Condition="'$(_SentryIsNet9OrGreater)' == 'true' AND '$(PlatformIsMobile)' == 'true' AND '$(ContinuousIntegrationBuild)' != 'true'">true</EnableMauiDeviceTestVisualRunner>
   </PropertyGroup>
   <!--
     Workaround for Verify issue with scrubbing when running in Rider on Windows.
--- a/test/Sentry.Maui.Device.TestApp/Sentry.Maui.Device.TestApp.csproj
+++ b/test/Sentry.Maui.Device.TestApp/Sentry.Maui.Device.TestApp.csproj
@@ -103,7 +103,7 @@
     <ProjectReference Include="..\Sentry.Tests\Sentry.Tests.csproj"/>
     <ProjectReference Include="..\Sentry.Extensions.Logging.Tests\Sentry.Extensions.Logging.Tests.csproj"/>
     <ProjectReference Include="..\Sentry.Maui.Tests\Sentry.Maui.Tests.csproj"/>
-    <ProjectReference Include="..\Sentry.Maui.CommunityToolkit.Mvvm.Tests\Sentry.Maui.CommunityToolkit.Mvvm.Tests.csproj" Condition="'$(IsNet9OrGreater)' == 'true'"/>
+    <ProjectReference Include="..\Sentry.Maui.CommunityToolkit.Mvvm.Tests\Sentry.Maui.CommunityToolkit.Mvvm.Tests.csproj" Condition="'$(_SentryIsNet9OrGreater)' == 'true'"/>
   </ItemGroup>
 
 </Project>


Directory.Build.props
```diff
@@ -114,6 +114,9 @@
   <!-- Helpful properties used elsewhere -->
   <PropertyGroup>
     <TargetFrameworkVersion>$([MSBuild]::GetTargetFrameworkVersion($(TargetFramework)))</TargetFrameworkVersion>
-    <TargetFrameworkIsNet9OrGreater Condition="$([MSBuild]::VersionGreaterThanOrEquals($(TargetFrameworkVersion), 9.0))">true</TargetFrameworkIsNet9OrGreater>
+    <IsNet8OrGreater>false</IsNet8OrGreater>
+    <IsNet8OrGreater Condition="$([MSBuild]::VersionGreaterThanOrEquals($(TargetFrameworkVersion), 8.0))">true</IsNet8OrGreater>
```
Flash0ver, line 118: style: combine

Could we combine each pair to a single
```XML
<IsNet8OrGreater>$([MSBuild]::VersionGreaterThanOrEquals($(TargetFrameworkVersion), 8.0))</IsNet8OrGreater>
```
?

Directory.Build.props
```diff
@@ -114,6 +114,9 @@
   <!-- Helpful properties used elsewhere -->
   <PropertyGroup>
     <TargetFrameworkVersion>$([MSBuild]::GetTargetFrameworkVersion($(TargetFramework)))</TargetFrameworkVersion>
```
Flash0ver: thought: I'm not sure if this will work as expected, when multi-targeting via `$(TargetFrameworks)` ... or even has ever worked as expected, for that matter.

I have the feeling that `$(TargetFramework)` is not yet available so early on for a `<PropertyGroup>` in `Directory.Build.props`.
I have the feeling that it only becomes available later in `<PropertyGroup>`s of `Directory.Build.targets`, as well as `<ItemGroup>`s and `<Target>`s.

Not entirely certain, thought, but I think we need to
- move this block to `Directory.Build.targets`
- ensure that the newly declared properties are only used in `ItemGroup`s and `Target`s
- and are neither used in other `PropertyGroup`s, nor in `<Project>` files that are packed (`Pack="true"`)

src/Sentry/Platforms/Native/buildTransitive/Sentry.Native.targets
```diff
@@ -20,7 +20,7 @@
 
   <PropertyGroup>
     <!-- net8.0 or greater -->
-    <FrameworkSupportsNative Condition="$([MSBuild]::IsTargetFrameworkCompatible('$(TargetFramework)', 'net8.0')) and ('$(OutputType)' == 'Exe' or '$(OutputType)' == 'WinExe')">true</FrameworkSupportsNative>
+    <FrameworkSupportsNative Condition="'$(IsNet8OrGreater)' == 'true' and ('$(OutputType)' == 'Exe' or '$(OutputType)' == 'WinExe')">true</FrameworkSupportsNative>
```
Flash0ver: thought: Will this work for published NuGet packages?

This `Sentry.Native.targets` packed into the NuGet-Package.
I don't think that our custom Property from `Directory.Build.props` are available at compile time for the consumer of the package.
Or is it?

Directory.Build.props
Flash0ver: style: combine

Could we combine each pair to a single
```diff
-<IsNet8OrGreater>false</IsNet8OrGreater>
-<IsNet8OrGreater Condition="$([MSBuild]::VersionGreaterThanOrEquals($(TargetFrameworkVersion), 8.0))">true</IsNet8OrGreater>
+<IsNet8OrGreater>$([MSBuild]::VersionGreaterThanOrEquals($(TargetFrameworkVersion), 8.0))</IsNet8OrGreater>
```
?

Directory.Build.props
```diff
@@ -114,6 +114,9 @@
...
-    <TargetFrameworkIsNet9OrGreater Condition="$([MSBuild]::VersionGreaterThanOrEquals($(TargetFrameworkVersion), 9.0))">true</TargetFrameworkIsNet9OrGreater>
+    <IsNet8OrGreater>false</IsNet8OrGreater>
+    <IsNet8OrGreater Condition="$([MSBuild]::VersionGreaterThanOrEquals($(TargetFrameworkVersion), 8.0))">true</IsNet8OrGreater>
+    <IsNet9OrGreater>false</IsNet9OrGreater>
```
Flash0ver: naming: `IsNet9OrGreater` vs `IsNet90OrGreater`

This may be just me, but I do have some trouble reading this property name.
I guess I would expect `90` instead of just `9`, to be in sync with the TFM `net9.0` and the constant `NET9_0`.

Well, putting both next to each other, with the upcoming `net10.0`
```
IsNet9OrGreater
IsNet90OrGreater
IsNet10OrGreater
IsNet100OrGreater
```
to me now `IsNet100OrGreater` reads weird and I can parse `IsNet10OrGreater` quicker.

I believe it's indeed best to leave it as proposed / as we had before: `IsNet9OrGreater`, because it's shorter, and we don't expect any minor TFMs from .NET anymore.

Directory.Build.props
```diff
@@ -114,6 +114,9 @@
...
+    <IsNet9OrGreater Condition="$([MSBuild]::VersionGreaterThanOrEquals($(TargetFrameworkVersion), 9.0))">true</IsNet9OrGreater>
```
Flash0ver: question/thought:
Should we use `$([MSBuild]::IsTargetFrameworkCompatible('$(TargetFramework)', 'net9.0'))` instead?

I believe this would be "technically more correct".
Because extracting the `TargetFrameworkVersion` is technically ambiguous with older versions, where
e.g. `2.0` could mean
- `netcoreapp2.0`
- `netstandard2.0`
- `net20`

Well, to be fair, .NET starts with `net5.0`, which is higher than all the other Frameworks, so we should be fine, practically speaking.

But `[MSBuild]::IsTargetFrameworkCompatible` also handles _OS-specific TFMs_ correctly.
I'm not sure if we could run into some unexpected corner cases with the current solution.

Directory.Build.props
jamescrosswell: Aha, thanks @Flash0ver - that might explain why Matt added this comment all the way back in 2022:
https://github.com/getsentry/sentry-dotnet/blob/31b6dcba8437c2396465327801d730d3e1d091b7/Directory.Build.props#L43-L50

src/Sentry/Platforms/Native/buildTransitive/Sentry.Native.targets
jamescrosswell: I think you might be right... it'll work for us when working with the solution locally. But there's nothing in the transitive build files that would set this property.

Directory.Build.props
jamescrosswell: The point of this PR was to replace `IsTargetFrameworkCompatible`. Have a look at the description for the issue it addresses: #4203

Directory.Build.props
jamescrosswell: I put together a little experiment to test this: https://github.com/jamescrosswell/msbuild-test-4275

- It looks like we should be safe to set these in the `Directory.Build.props` 
- We should avoid using `Directory.Build.targets` for these properties
- We probably want to guard setting these properties with a Condition like `'$(TargetFramework)' != ''` since it appears at certain stages during the build, this can be empty (and we get unexpected results)

src/Sentry/Platforms/Native/buildTransitive/Sentry.Native.targets
jamescrosswell: I refactored those helper properties into a Sentry.props file that we can include transitively (so that they're available both when building the SDK and when customers are building apps using the SDK).

Directory.Build.props
jamescrosswell: We can - that's simpler!

Directory.Build.props
Flash0ver: The experiment is not entirely fair ... from what I can glean:
- The `PropertyGroup` defined in `Directory.Build.targets` will set - to my knowledge - all the properties as expected
- But these properties, e.g. `IsNet9OrGreaterTarget` is used in a `PropertyGroup` of the `.csproj`, which is included before the `Directory.Build.targets` is included
  - hence the properties used in `DebugMessageTarget` of the `.csproj` are not set yet and always empty
  - however, they would be set as expected if used within either `<ItemGroup>` or `<Target>`, which are evaluated after all `<PropertyGroup>` have been evaluted

What I did so far in my smaller projects, when using `$(TargetFramework)`, `$(TargetFrameworkVersion)`, `$(TargetFrameworkIdentifier)`, ..., is to declare custom Properties in a `Directory.Build.targets` and then use these Properties only in either `<ItemGroup>` or `<Target>`, or in a `<PropertyGroup>` afterwards in the same `Directory.Build.targets` to set/overwrite Properties used by the compiler:

Example: https://github.com/Flash0ver/F0.Analyzers/blob/main/source/Directory.Build.targets

i.e. inlining
```diff
-<PropertyGroup>
-  <DebugMessageTarget>TargetFramework: $(TargetFramework), TargetFrameworkVersionTarget: $(TargetFrameworkVersionTarget), IsNet8OrGreaterTarget: $(IsNet8OrGreaterTarget), IsNet9OrGreaterTarget: $(IsNet9OrGreaterTarget)</DebugMessageTarget>
-</PropertyGroup>
-<Target Name="LogFrameworkPropertiesBeforeBuild" BeforeTargets="Build">
-  <Message Text="[$(MSBuildProjectName)] [BeforeBuild] $(DebugMessageTarget)" Importance="high" />
-</Target>
+<Target Name="LogFrameworkPropertiesBeforeBuild" BeforeTargets="Build">
+  <Message Text="[$(MSBuildProjectName)] [BeforeBuild] TargetFramework: $(TargetFramework), TargetFrameworkVersionTarget: $(TargetFrameworkVersionTarget), IsNet8OrGreaterTarget: $(IsNet8OrGreaterTarget), IsNet9OrGreaterTarget: $(IsNet9OrGreaterTarget)" Importance="high" />
+</Target>
```
should work as expected ... from what (I think) I know.

Flash0ver: ... oh wait ... I was wrong ... it's not entirely correct:
```
> dotnet build
  [TestConsole] [PrepareForBuild] TargetFramework: net8.0, TargetFrameworkVersionTarget: 8.0, IsNet8OrGreaterTarget: true, IsNet9OrGreaterTarget: false
  [TestConsole] [PrepareForBuild] TargetFramework: net9.0, TargetFrameworkVersionTarget: 9.0, IsNet8OrGreaterTarget: true, IsNet9OrGreaterTarget: true
  [TestLibrary] [PrepareForBuild] TargetFramework: net9.0, TargetFrameworkVersionTarget: 9.0, IsNet8OrGreaterTarget: true, IsNet9OrGreaterTarget: true
  [TestLibrary] [PrepareForBuild] TargetFramework: net8.0, TargetFrameworkVersionTarget: 8.0, IsNet8OrGreaterTarget: true, IsNet9OrGreaterTarget: false
  [TestLibrary] [BeforeCompile] TargetFramework: net9.0, TargetFrameworkVersionTarget: 9.0, IsNet8OrGreaterTarget: true, IsNet9OrGreaterTarget: true
  [TestLibrary] [AfterCompile] TargetFramework: net9.0, TargetFrameworkVersionTarget: 9.0, IsNet8OrGreaterTarget: true, IsNet9OrGreaterTarget: true
  [TestLibrary] [BeforeBuild] TargetFramework: net9.0, TargetFrameworkVersionTarget: 9.0, IsNet8OrGreaterTarget: true, IsNet9OrGreaterTarget: true
  [TestLibrary] [AfterBuild] TargetFramework: net9.0, TargetFrameworkVersionTarget: 9.0, IsNet8OrGreaterTarget: true, IsNet9OrGreaterTarget: true
  [TestLibrary] [BeforeCompile] TargetFramework: net8.0, TargetFrameworkVersionTarget: 8.0, IsNet8OrGreaterTarget: true, IsNet9OrGreaterTarget: false
  [TestLibrary] [AfterCompile] TargetFramework: net8.0, TargetFrameworkVersionTarget: 8.0, IsNet8OrGreaterTarget: true, IsNet9OrGreaterTarget: false
  [TestLibrary] [BeforeBuild] TargetFramework: net8.0, TargetFrameworkVersionTarget: 8.0, IsNet8OrGreaterTarget: true, IsNet9OrGreaterTarget: false
  [TestLibrary] [AfterBuild] TargetFramework: net8.0, TargetFrameworkVersionTarget: 8.0, IsNet8OrGreaterTarget: true, IsNet9OrGreaterTarget: false
  [TestLibrary] [BeforeBuild] TargetFramework: , TargetFrameworkVersionTarget: 0.0, IsNet8OrGreaterTarget: false, IsNet9OrGreaterTarget: false
  [TestLibrary] [AfterBuild] TargetFramework: , TargetFrameworkVersionTarget: 0.0, IsNet8OrGreaterTarget: false, IsNet9OrGreaterTarget: false
  [TestConsole] [BeforeCompile] TargetFramework: net9.0, TargetFrameworkVersionTarget: 9.0, IsNet8OrGreaterTarget: true, IsNet9OrGreaterTarget: true
  [TestConsole] [AfterCompile] TargetFramework: net9.0, TargetFrameworkVersionTarget: 9.0, IsNet8OrGreaterTarget: true, IsNet9OrGreaterTarget: true
  [TestConsole] [BeforeCompile] TargetFramework: net8.0, TargetFrameworkVersionTarget: 8.0, IsNet8OrGreaterTarget: true, IsNet9OrGreaterTarget: false
  [TestConsole] [AfterCompile] TargetFramework: net8.0, TargetFrameworkVersionTarget: 8.0, IsNet8OrGreaterTarget: true, IsNet9OrGreaterTarget: false
  [TestConsole] [BeforeBuild] TargetFramework: net9.0, TargetFrameworkVersionTarget: 9.0, IsNet8OrGreaterTarget: true, IsNet9OrGreaterTarget: true
  [TestConsole] [AfterBuild] TargetFramework: net9.0, TargetFrameworkVersionTarget: 9.0, IsNet8OrGreaterTarget: true, IsNet9OrGreaterTarget: true
  [TestConsole] [BeforeBuild] TargetFramework: net8.0, TargetFrameworkVersionTarget: 8.0, IsNet8OrGreaterTarget: true, IsNet9OrGreaterTarget: false
  [TestConsole] [AfterBuild] TargetFramework: net8.0, TargetFrameworkVersionTarget: 8.0, IsNet8OrGreaterTarget: true, IsNet9OrGreaterTarget: false
  [TestConsole] [BeforeBuild] TargetFramework: , TargetFrameworkVersionTarget: 0.0, IsNet8OrGreaterTarget: false, IsNet9OrGreaterTarget: false
  [TestConsole] [AfterBuild] TargetFramework: , TargetFrameworkVersionTarget: 0.0, IsNet8OrGreaterTarget: false, IsNet9OrGreaterTarget: false
```

There is one case where they are not set ... hmm ... today I learned.
Great demo you've prepared!

Hmm ... now I'm uncertain what the most stable way of doing this would be 🤔

Flash0ver: Hmm ... now I'm uncertain what the most stable way of doing this would be 🤔

Directory.Build.props
Flash0ver: Oh ... I believe I made myself misunderstood.
I meant to consolidate the use of `IsTargetFrameworkCompatible` within the `IsNet{MAJOR}OrGreater` properties only, where we review/ensure the correct usage once, and use our evaluated and shorter property everywhere else.

Because I think overwriting `$(TargetFrameworkVersion)` could be dangerous.
`$(TargetFrameworkVersion)` is a property set by MSBuild, and if some other build props/items/targets depend on it, we may cause unexpected behavior, because we effectively remove the `v` prefix of the `$(TargetFrameworkVersion)`:
e.g.
- .NET: from `v9.0` to `9.0`
- .NETFramework: from `v4.7.2` to `4.7.2`
- .NETCoreApp: from `v3.1` to `3.1`
- .NETStandard: from `v2.0` to `2.0`

But perhaps we can just prefix our variant of the property with `Sentry` ... so to rename it from `TargetFrameworkVersion` to `SentryTargetFrameworkVersion` ... so that we do not change that reserved/well-known property.

src/Sentry/buildTransitive/Sentry.props
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project ToolsVersion="4.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <!-- Helpful properties used elsewhere -->
+  <PropertyGroup Condition="'$(TargetFramework)' != ''">
+    <TargetFrameworkVersion>$([MSBuild]::GetTargetFrameworkVersion($(TargetFramework)))</TargetFrameworkVersion>
```
Flash0ver: _potentially dangerous ... I know we used the name `TargetFrameworkVersion` previously already ... I missed this in the preceding review ... but am noticing it now since we impact user builds_

`TargetFrameworkVersion` is a Common MSBuild project property.
See https://learn.microsoft.com/visualstudio/msbuild/common-msbuild-project-properties
User or other MSBuild code may depend on it.
And we essentially remove the `v` prefix from this property.

Perhaps we should prefix this property with `Sentry`.
Perhaps we should prefix all properties, that we inject into consuming builds, with `Sentry`.

This would make the `IsNet8OrGreater` slightly longer, but I think it's worth it to not accidentally override any potentially user-defined properties.

Directory.Build.props
jamescrosswell: Ew.... that doesn't sound good. 

I don't see `TargetFrameworkVersion` in the docs for SDK style projects but it looks like it exists/existed for old style projects. Easy enough to avoid - I've renamed to avoid conflicts.

Directory.Build.props
jamescrosswell: From what I can tell, it works when we put it in a props file... but there are situations in which it doesn't work when it's in a targets file. 

All of the integration tests are now passing, which implies this is working as we would want/expect.

Is there any reason to change how it's done currently then (in a props file)?

Flash0ver: No ... if it works, it works ... thanks for the discussion around it and the experiment project to unveil MSBuild behavior.

Flash0ver (approved):

PR #4276, opened 2025-06-12
fix: InvalidOperationException sending attachments on Android with LLVM enabled

### LINKED ISSUE #3101: Some attachments not send on MAUI
# Important Note

⚠ ⚠ ⚠ 👀 👀 👀   
**This was an issue with the .NET Runtime.** We were keeping it here as a tracking issue (see [comment](https://github.com/getsentry/sentry-dotnet/issues/3101#issuecomment-2594795474)).

It looks like there are finally [workarounds available](https://github.com/getsentry/sentry-dotnet/issues/3101#issuecomment-2943782262) for this however, so we'll get a fix out ASAP. Thanks @tranb3r for following up to let us know!

⚠ ⚠ ⚠ 👀 👀 👀 

### Package

Sentry.Maui

### .NET Flavor

.NET

### .NET Version

8.0.0

### OS

Android

### SDK Version

Sentry.Maui 3.41.3

_No response_

### Steps to Reproduce

Hello,

On my MAUI application, I send a "SentrySdk.CaptureMessage" where I configure in the scope a number of attachments which are ZIP files that contain logs (NLog) :

`scope.AddAttachment(attachment, AttachmentType.Default, "application/zip");`

However, on some phones and quite randomly, some of these files are not sent.
The files are not used and they are all stored in the same directory, and the zip is created in the same way.

I managed, with the Sentry DEBUG set to TRUE, to get the error :

```
Warning: Failed to serialize envelope item
System.InvalidOperationException: Arg_InvalidOperationException
  at System.Threading.Tasks.Sources.ManualResetValueTaskSourceCore`1[[System.Int64, System.Private.CoreLib, Version=8.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e]].GetStatus(Int16 )
  at Microsoft.Win32.SafeHandles.SafeFileHandle.ThreadPoolValueTaskSource.GetStatus(Int16 )
  at System.IO.Strategies.BufferedFileStreamStrategy.ReadAsync(Byte[] , Int32 , Int32 , CancellationToken )
  at System.IO.FileStream.ReadAsync(Byte[] , Int32 , Int32 , CancellationToken )
  at Sentry.Internal.PartialStream.ReadAsync(Byte[] buffer, Int32 offset, Int32 count, CancellationToken cancellationToken)
  at System.IO.Stream.<CopyToAsync>g__Core|27_0(Stream , Stream , Int32 , CancellationToken )
  at Sentry.Protocol.Envelopes.EnvelopeItem.BufferPayloadAsync(IDiagnosticLogger logger, CancellationToken cancellationToken)
  at Sentry.Protocol.Envelopes.EnvelopeItem.SerializeAsync(Stream stream, IDiagnosticLogger logger, CancellationToken cancellationToken)
  at Sentry.Protocol.Envelopes.Envelope.SerializeAsync(Stream stream, IDiagnosticLogger logger, ISystemClock clock, CancellationToken cancellationToken)
```

I've only managed to reproduce this problem on Android phones.

Can you help me?
Thanks

### Expected Result

All attachments send

### Actual Result

Some attachments not send

jamescrosswell:
Resolves #3101:
- https://github.com/getsentry/sentry-dotnet/issues/3101

Also see:
- CA2022: https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca2022
- CA1835: https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1835

### DIFF
--- a/src/Sentry/Internal/PartialStream.cs
+++ b/src/Sentry/Internal/PartialStream.cs
@@ -60,7 +60,8 @@ internal class PartialStream : Stream
             _innerStream.Position = innerPosition;
         }
 
-        var read = await _innerStream.ReadAsync(buffer, offset, actualCount, cancellationToken)
+        // See https://github.com/getsentry/sentry-dotnet/issues/3101
+        var read = await _innerStream.ReadAsync(buffer.AsMemory(offset, actualCount), cancellationToken)
             .ConfigureAwait(false);
 
         if (_length != null)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 ### Fixes
 
 - The HTTP instrumentation uses the span created for the outgoing request in the sentry-trace header, fixing the parent-child relationship between client and server ([#4264](https://github.com/getsentry/sentry-dotnet/pull/4264))
+- InvalidOperationException sending attachments on Android with LLVM enabled ([#4276](https://github.com/getsentry/sentry-dotnet/pull/4276))
 
 ### Dependencies
 


src/Sentry/Internal/PartialStream.cs
```diff
@@ -60,7 +60,8 @@ public override async Task<int> ReadAsync(byte[] buffer, int offset, int count,
             _innerStream.Position = innerPosition;
         }
 
-        var read = await _innerStream.ReadAsync(buffer, offset, actualCount, cancellationToken)
+        // See https://github.com/getsentry/sentry-dotnet/issues/3101
```
bruno-garcia, line 63: off, some .NET bug? how can we avoid this from happening again?

bruno-garcia (approved):

src/Sentry/Internal/PartialStream.cs
bruno-garcia: Don't we have treat-warning-as-error on?

Flash0ver: I believe we only have the `Default` ruleset enabled, depending on the .NET SDK version.
We could enable `All` rules, and also lock/pin to a specific .NET SDK ruleset for deterministic builds.

Opened new issue as follow-up for more code analysis rules and diagnostics: #4287

jamescrosswell: > Don't we have treat-warning-as-error on?

We do but the analyzers for this error only appear to trigger in the .NET 10 SDK (we weren't getting any errors before that).

PR #4283, opened 2025-06-16
Ensure HTTP Response Status Code is searchable when using OTel

### LINKED ISSUE #4094: Treat Activity.TagObjects "http.response.status_code" as a special value to import via OTel
### Problem Statement

 Inspired by the discussion here - 

Thanks to Coby from support for pointing me to this discussion.

> We attribute special meaning to some TagObjects based on the Key in the KVP. For example, we know the URL is important in ASP.NET Core apps:
> https://github.com/getsentry/sentry-dotnet/blob/9b29d1f7b6fe1f2638d344c5cbcb60281e16739c/src/Sentry.OpenTelemetry/SentrySpanProcessor.cs#L195
> 
> The key is really the only thing that provide some hint... but we can't use this for custom tags that you've defined. 
> 
> If you wanted to make some specific TagObjects searchable, you could potentially write some custom logic in the [SetBeforeSendTransaction](https://docs.sentry.io/platforms/dotnet/configuration/options/#set-before-send-transaction) callback to copy these across.

 _Originally posted by @jamescrosswell in [#4043](https://github.com/getsentry/sentry-dotnet/issues/4043#issuecomment-2737917001)_

This got me wondering if there could be the same special treatment for `http.response.status_code` ? 

This is the only OpenTelemetry attribute brought through by default with `.AddAspNetCoreInstrumentation()` that I cannot find in a more visible space on the traces page and would be a very welcome addition.

Image

Furthermore, as per the above discussion I did explore using `SetBeforeSendTransaction` and `SentrySdk` which worked as expected, however having the response code extraction built in by default might be a welcome feature for the community?

```
o.SetBeforeSendTransaction((transaction, hint) =>
{
    var act = Activity.Current;
    var statusCode = act?.TagObjects.FirstOrDefault(x => x.Key == "http.response.status_code");
    if (statusCode is not null && statusCode.HasValue)
        transaction.SetTag("http.response.status_code", statusCode.Value.ToString());

    return transaction;
});
```

```
        {
            scope.SetTag("ResponseStatusCode", context.Response.StatusCode.ToString());
        });
```

Image

### Solution Brainstorm

_No response_

jamescrosswell:
Resolves #4094:
- https://github.com/getsentry/sentry-dotnet/issues/4094

### DIFF
--- a/src/Sentry.OpenTelemetry/OpenTelemetryExtensions.cs
+++ b/src/Sentry.OpenTelemetry/OpenTelemetryExtensions.cs
@@ -7,18 +7,21 @@ internal static class OpenTelemetryExtensions
 {
     public static SpanId AsSentrySpanId(this ActivitySpanId id) => SpanId.Parse(id.ToHexString());
 
-    public static ActivitySpanId AsActivitySpanId(this SpanId id) => ActivitySpanId.CreateFromString(id.ToString().AsSpan());
+    public static ActivitySpanId AsActivitySpanId(this SpanId id) =>
+        ActivitySpanId.CreateFromString(id.ToString().AsSpan());
 
     public static SentryId AsSentryId(this ActivityTraceId id) => SentryId.Parse(id.ToHexString());
 
-    public static ActivityTraceId AsActivityTraceId(this SentryId id) => ActivityTraceId.CreateFromString(id.ToString().AsSpan());
+    public static ActivityTraceId AsActivityTraceId(this SentryId id) =>
+        ActivityTraceId.CreateFromString(id.ToString().AsSpan());
 
-    public static BaggageHeader AsBaggageHeader(this IEnumerable<KeyValuePair<string, string?>> baggage, bool useSentryPrefix = false) =>
+    public static BaggageHeader AsBaggageHeader(this IEnumerable<KeyValuePair<string, string?>> baggage,
+        bool useSentryPrefix = false) =>
         BaggageHeader.Create(
             baggage.Where(member => member.Value != null)
-                        .Select(kvp => (KeyValuePair<string, string>)kvp!),
+                .Select(kvp => (KeyValuePair<string, string>)kvp!),
             useSentryPrefix
-            );
+        );
 
     /// <summary>
     /// The names that OpenTelemetry gives to attributes, by convention, have changed over time so we often need to
@@ -40,6 +43,7 @@ internal static class OpenTelemetryExtensions
                 return value;
             }
         }
+
         return default;
     }
 
@@ -47,11 +51,21 @@ internal static class OpenTelemetryExtensions
         attributes.GetFirstMatchingAttribute<string>(
             OtelSemanticConventions.AttributeHttpRequestMethod,
             OtelSemanticConventions.AttributeHttpMethod // Fallback pre-1.5.0
-            );
+        );
 
     public static string? UrlFullAttribute(this IDictionary<string, object?> attributes) =>
         attributes.GetFirstMatchingAttribute<string>(
             OtelSemanticConventions.AttributeUrlFull,
             OtelSemanticConventions.AttributeHttpUrl // Fallback pre-1.5.0
-            );
+        );
+
+    public static short? HttpResponseStatusCodeAttribute(this IDictionary<string, object?> attributes)
+    {
+        var statusCode = attributes.GetFirstMatchingAttribute<int?>(
+            OtelSemanticConventions.AttributeHttpResponseStatusCode
+        );
+        return statusCode is >= short.MinValue and <= short.MaxValue
+            ? (short)statusCode.Value
+            : null;
+    }
 }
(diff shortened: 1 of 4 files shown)


src/Sentry.OpenTelemetry/SentrySpanProcessor.cs
```diff
@@ -230,10 +230,13 @@ public override void OnEnd(Activity data)
         span.Operation = operation;
         span.Description = description;
 
+        // Handle HTTP response status code specially
+        var statusCode = attributes.HttpResponseStatusCodeAttribute();
         if (span is TransactionTracer transaction)
         {
             transaction.Name = description;
             transaction.NameSource = source;
+            transaction.Contexts.Response.StatusCode = statusCode;
```
Flash0ver: question: unconditional overwrite

Is this behavior intended?
- when `attributes.HttpResponseStatusCodeAttribute()` returns `null`
  - e.g. when the `http.response.status_code` Attribute is not found
- then overwrite the `transaction.Contexts.Response.StatusCode` with `null`
  - even if `transaction.Contexts.Response.StatusCode` is not null

Or is this even a possible data-flow scenario?

src/Sentry.OpenTelemetry/SentrySpanProcessor.cs
```diff
@@ -250,6 +253,11 @@ public override void OnEnd(Activity data)
             // Resource attributes do not need to be set, as they would be identical as those set on the transaction.
             spanTracer.SetExtras(attributes);
             spanTracer.SetExtra("otel.kind", data.Kind);
+            if (statusCode is { } responseStatusCode)
+            {
+                // Set this as a tag so that it's searchable in Sentry
+                span.SetTag("HTTP Response Status Code", responseStatusCode.ToString());
```
Flash0ver: question: Are white spaces in Tags allowed?

See https://docs.sentry.io/platforms/dotnet/enriching-events/tags/

src/Sentry.OpenTelemetry/OpenTelemetryExtensions.cs
```diff
@@ -7,18 +7,21 @@ internal static class OpenTelemetryExtensions
...
 
-    public static ActivitySpanId AsActivitySpanId(this SpanId id) => ActivitySpanId.CreateFromString(id.ToString().AsSpan());
+    public static ActivitySpanId AsActivitySpanId(this SpanId id) =>
+        ActivitySpanId.CreateFromString(id.ToString().AsSpan());
 
     public static SentryId AsSentryId(this ActivityTraceId id) => SentryId.Parse(id.ToHexString());
 
-    public static ActivityTraceId AsActivityTraceId(this SentryId id) => ActivityTraceId.CreateFromString(id.ToString().AsSpan());
+    public static ActivityTraceId AsActivityTraceId(this SentryId id) =>
+        ActivityTraceId.CreateFromString(id.ToString().AsSpan());
```
Flash0ver, line 16: perf: intermediate string allocations

I know this is just a re-format of existing code.
But I noticed an extra string allocations just used as intermediate data for converting from `SpanId` to `ActivitySpanId`, and from `SentryId` to `ActivityTraceId`:
- allocating from `SpanId`/`SentryId` to `string` via `ToString()`
- converted allocation-free to a `ReadOnlySpan<char>`
- where `ActivitySpanId.CreateFromString` and `ActivityTraceId.CreateFromString` both internally create a new string that is passed to the `ActivitySpanId`/`ActivityTraceId` constructors

If we would - instead of converting to a heap-allocated `string` and calling `CreateFromString`,
would instead either
- format to a stack-allocated `Span<char>` and call `CreateFromString`
- format to a stack-allocated `ReadOnlySpan<byte>` and call `CreateFromUtf8String`
- format to a stack-allocated `ReadOnlySpan<byte>` and call `CreateFromBytes`

we should be able to get rid of half of the heap allocations there.

This is a smaller gain, where we could get, in the `SentryId` case 
- from `176 B` (GUID to String with Length 32 and 24 B Class overhead times two)
- to `88 B` (cutting above in half)

Do you think this is worth a future improvement?
If so, then I will create a new issue from this comment, labeled `Good First Issue` with a low priority.
Otherwise, we just resolve this comment.

src/Sentry.OpenTelemetry/SentrySpanProcessor.cs
jamescrosswell: Good catch - no they are not!

src/Sentry.OpenTelemetry/SentrySpanProcessor.cs
jamescrosswell: I think it's a very unlikely scenario... these transactions are auto-instrumented so pretty unlikely that the SDK user will be meddling with these.

I added a conditional and a unit test to make sure we only set (or overwrite) this if the OTEL attribute is present though, just for good measure.

src/Sentry.OpenTelemetry/OpenTelemetryExtensions.cs
jamescrosswell: Sure that's a good idea, especially since this method will be called quite frequently (it's going to be in a hot path for someone, somewhere).

jamescrosswell: Created #4289 to address this.

PR #4284, opened 2025-06-16
Added email validation to CaptureFeedback methods

### LINKED ISSUE #4246: Validate email in CaptureFeedback / CaptureUserFeedback APIs
> yeah we changed the order of the args then and didn't realize this is easy to mix up. A debug message makes sense. cc @jamescrosswell might be a quick win `(!contains(@)` just a best effort thing 

 _Originally posted by @bruno-garcia in [#2180](https://github.com/getsentry/sentry-unity/issues/2180#issuecomment-2923696656)_

jamescrosswell:
Resolves #4246
- https://github.com/getsentry/sentry-dotnet/issues/4246

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 
 - The HTTP instrumentation uses the span created for the outgoing request in the sentry-trace header, fixing the parent-child relationship between client and server ([#4264](https://github.com/getsentry/sentry-dotnet/pull/4264))
 - InvalidOperationException sending attachments on Android with LLVM enabled ([#4276](https://github.com/getsentry/sentry-dotnet/pull/4276))
+- When CaptureFeedback methods are called with invalid email addresses, a warning is now logged if Debug mode is enabled (explaining why the feedback has been dropped) ([#4284](https://github.com/getsentry/sentry-dotnet/pull/4284))
 
 ### Dependencies
 
--- a/src/Sentry/Internal/EmailValidator.cs
+++ b/src/Sentry/Internal/EmailValidator.cs
@@ -25,13 +25,5 @@ internal static partial class EmailValidator
     /// </summary>
     /// <param name="email">The email address to validate.</param>
     /// <returns>True if the email is valid, false otherwise.</returns>
-    public static bool IsValidEmail(string? email)
-    {
-        if (string.IsNullOrEmpty(email))
-        {
-            return true;
-        }
-
-        return Email.IsMatch(email);
-    }
+    public static bool IsValidEmail(string email) =>  Email.IsMatch(email);
 }
--- a/src/Sentry/Internal/EmailValidator.cs
+++ b/src/Sentry/Internal/EmailValidator.cs
@@ -11,7 +11,7 @@ internal static partial class EmailValidator
 
 #if NET9_0_OR_GREATER
     [GeneratedRegex(EmailPattern)]
-    private static partial Regex EmailRegex { get; }
+    private static partial Regex Email { get; }
 #elif NET8_0
     [GeneratedRegex(EmailPattern)]
     private static partial Regex EmailRegex();
@@ -32,10 +32,6 @@ internal static partial class EmailValidator
             return true;
         }
 
-#if NET9_0_OR_GREATER
-        return EmailRegex.IsMatch(email);
-#else
         return Email.IsMatch(email);
-#endif
     }
 }
--- a/src/Sentry/Internal/EmailValidator.cs
+++ b/src/Sentry/Internal/EmailValidator.cs
@@ -25,5 +25,5 @@ internal static partial class EmailValidator
     /// </summary>
     /// <param name="email">The email address to validate.</param>
     /// <returns>True if the email is valid, false otherwise.</returns>
-    public static bool IsValidEmail(string email) =>  Email.IsMatch(email);
+    public static bool IsValidEmail(string email) => Email.IsMatch(email);
 }
--- a/src/Sentry/Internal/EmailValidator.cs
+++ b/src/Sentry/Internal/EmailValidator.cs
@@ -1,0 +1,41 @@
+using System.Text.RegularExpressions;
+
+namespace Sentry.Internal;
+
+/// <summary>
+/// Helper class for email validation.
+/// </summary>
+internal static partial class EmailValidator
+{
+    private const string EmailPattern = @"^[^@\s]+@[^@\s]+\.[^@\s]+$";
+
+#if NET9_0_OR_GREATER
+    [GeneratedRegex(EmailPattern)]
+    private static partial Regex EmailRegex { get; }
+#elif NET8_0
+    [GeneratedRegex(EmailPattern)]
+    private static partial Regex EmailRegex();
+    private static readonly Regex Email = EmailRegex();
+#else
+    private static readonly Regex Email = new(EmailPattern, RegexOptions.Compiled);
+#endif
+
+    /// <summary>
+    /// Validates an email address.
+    /// </summary>
+    /// <param name="email">The email address to validate.</param>
+    /// <returns>True if the email is valid, false otherwise.</returns>
+    public static bool IsValidEmail(string? email)
+    {
+        if (string.IsNullOrEmpty(email))
+        {
+            return true;
+        }
+
+#if NET9_0_OR_GREATER
+        return EmailRegex.IsMatch(email);
+#else
+        return Email.IsMatch(email);
+#endif
+    }
+}
--- a/src/Sentry/Internal/Hub.cs
+++ b/src/Sentry/Internal/Hub.cs
@@ -553,12 +553,6 @@ internal class Hub : IHub, IDisposable
 
         try
         {
-            if (!string.IsNullOrWhiteSpace(feedback.ContactEmail) && !EmailValidator.IsValidEmail(feedback.ContactEmail))
-            {
-                _options.LogWarning("Feedback dropped due to invalid email format: '{0}'", feedback.ContactEmail);
-                return;
-            }
-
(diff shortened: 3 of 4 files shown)


src/Sentry/Internal/Hub.cs
```diff
@@ -553,6 +553,12 @@ public void CaptureFeedback(SentryFeedback feedback, Action<Scope> configureScop
 
         try
         {
+            if (!EmailValidator.IsValidEmail(feedback.ContactEmail))
+            {
+                _options.LogWarning("Feedback dropped due to invalid email format: '{0}'", feedback.ContactEmail);
+                return;
```
jamescrosswell: An alternative here might be to set the email to an empty string (since the email address isn't strictly required). That way SDK users wouldn't lose potentially valuable feedback if their users make a typo when entering their email address.

src/Sentry/Internal/Hub.cs
```diff
@@ -553,6 +553,12 @@ public void CaptureFeedback(SentryFeedback feedback, Action<Scope> configureScop
 
         try
         {
+            if (!EmailValidator.IsValidEmail(feedback.ContactEmail))
```
Flash0ver: question: `string.IsNullOrWhiteSpace`

Unlike the other two usages of `EmailValidator.IsValidEmail`, here we do not do a `string.IsNullOrWhiteSpace` before.
```diff
-if (!EmailValidator.IsValidEmail(feedback.ContactEmail))
+if (!string.IsNullOrWhiteSpace(feedback.ContactEmail) && !EmailValidator.IsValidEmail(feedback.ContactEmail))
```

- Should we do a `string.IsNullOrWhiteSpace` check here too, or is it consciously elided?
- Do we perhaps want to move the `string.IsNullOrWhiteSpace` check into the new `EmailValidator.IsValidEmail` method for consistent validation rules?

src/Sentry/Internal/EmailValidator.cs
```diff
@@ -0,0 +1,41 @@
...
+    /// <returns>True if the email is valid, false otherwise.</returns>
+    public static bool IsValidEmail(string? email)
+    {
+        if (string.IsNullOrEmpty(email))
+        {
+            return true;
+        }
+
+#if NET9_0_OR_GREATER
+        return EmailRegex.IsMatch(email);
```
Flash0ver: style: consistent member-name and member-kind of `Regex` across TFMs

**Member-Name**
Could we consistently name the `Regex` property and field, depending on the TFM, either `EmailRegex` or `Email`, so that we don't need the `#if` here and keep the conditional compilation part at a single location, where only the partial Method for `net8.0` uses a different name.

**Member-Kind**
Should we change the current `Email` _field_ when _.NET 8.0 or less_ to a _property_ as well, so that the `Regex` member-kind is always a _Property_, regardless of the TFM.

**Both**
Well ... to be fair ... the current style should not cause any issues as the members and their usage are all private.
But I have to admit I get slightly uneasy when member-names and member-kinds change across TFMs: Should we make any of these conditional members non-private in the future, we may run into _MissingMethodException/MissingMemberException_.
Example contrived and I'm also not entirely certain ... so this is more of a nitpick than it is a potential problem.

What are your thoughts ... am I overthinking this?

src/Sentry/Internal/Hub.cs
Flash0ver: right ... because the consumer of `Sentry` does not know whether the user at runtime has entered a valid EMail address ... and would need to perform validation on their end as well

Technically, this would be a behavioral change, as we would be more permissive than before.
On the other hand, we already permit the absence of an EMail address, so additionally permitting invalid EMail addresses should not be an issue ... but we would modify the invalid EMail address to `null` ... where I'm not sure if that is what SDK users will expect.
I guess with the current API, the consumer would need to add their own EMail-validation if they only want to allow correct EMail formats and present a warning to the runtime user to enter a valid EMail address.

For a "full" story, perhaps we'd need to add something to our API.
Some thoughts:
- returning a `bool` from `CaptureFeedback`
  - would be a breaking change
  - would be inconsistent with the other Capture-APIs
- we could add an overload that returns an `out bool valid`
  - feels inconsistent to the other Capture-APIs
- we could add a `BeforeSendFeedback` callback to the Sentry-Options
  - where we give the consumer a chance to intercept the (User)Feedback
  - but this couldn't be used to add some validation to a TextBox in the consumer's UI

I took a quick look at `sentry-java` and `sentry-javascript`, and from what I can glean, there is no EMail validation nor "correction"/modification either.

Long story short,
IMO surfacing Sentry's behavior early through the Debug-Message before sending is fine for now,
and if users request more flexibility with how Feedback is validated and being sent to Sentry, we could revisit this flow and introduce more changes.
But it's a weak, non-blocking preference from my side, to go with this proposed change for now as a small win.

jamescrosswell: Currently if someone enters an invalid email address in their production app they don't receive the feedback.

It feels like it would be better if they get the feedback but don't know the email address of the user who submitted it (not great but better). As long as we log this via Debug, it's probably OK. 

I don't think there's any benefit in changing the method signature to return a success/failure (we're going to silently scrub the email address if it's funky... logging is the best we can do really - we don't want to throw an exception). 

If they want to avoid these issues entirely, they just need to do some email validation on the form before submitting it...

src/Sentry/Internal/Hub.cs
jamescrosswell: True. I corrected in one commit... then reversed in the commit where the behaviour was changed from logging to scrubbing invalid emails (since this overload is just a wrapper around the other method where this happens now).

src/Sentry/Internal/EmailValidator.cs
jamescrosswell: Nah all good... this was my first attempt at using cursor to help put a PR together and it gave them separate names for some reason. I've given them the same name and removed the second set of (unnecessary) directives.

CHANGELOG.md
```diff
@@ -11,6 +11,7 @@
 
 - The HTTP instrumentation uses the span created for the outgoing request in the sentry-trace header, fixing the parent-child relationship between client and server ([#4264](https://github.com/getsentry/sentry-dotnet/pull/4264))
 - InvalidOperationException sending attachments on Android with LLVM enabled ([#4276](https://github.com/getsentry/sentry-dotnet/pull/4276))
+- When CaptureFeedback methods are called with invalid email addresses, a warning is now logged if Debug mode is enabled (explaining why the feedback has been dropped) ([#4284](https://github.com/getsentry/sentry-dotnet/pull/4284))
```
Flash0ver: docs: update CHANGELOG entry to reflect new behavior of removing/scrubbing the EMail address of (User)Feedback when invalid format

src/Sentry/Internal/Hub.cs
```diff
@@ -573,6 +573,12 @@ public void CaptureFeedback(SentryFeedback feedback, Scope? scope = null, Sentry
 
         try
         {
+            if (!string.IsNullOrWhiteSpace(feedback.ContactEmail) && !EmailValidator.IsValidEmail(feedback.ContactEmail))
+            {
+                _options.LogWarning("Feedback email scrubbed due to invalid email format: '{0}'", feedback.ContactEmail);
+                feedback.ContactEmail = null;
```
Flash0ver, line 579: question: does this warrant a mention in the docs
https://docs.sentry.io/platforms/dotnet/user-feedback/

Or is it implicitly documented enough via the Diagnostic-Log above?

src/Sentry/Internal/EmailValidator.cs
```diff
@@ -0,0 +1,37 @@
...
+#endif
+
+    /// <summary>
+    /// Validates an email address.
+    /// </summary>
+    /// <param name="email">The email address to validate.</param>
+    /// <returns>True if the email is valid, false otherwise.</returns>
+    public static bool IsValidEmail(string? email)
+    {
+        if (string.IsNullOrEmpty(email))
```
Flash0ver: follow-up to https://github.com/getsentry/sentry-dotnet/pull/4284#discussion_r2149780497

Before both usages of `EmailValidator.IsValidEmail`, we check for `string.IsNullOrWhiteSpace`.
And here, before the `Regex.IsMatch`, we check `string.IsNullOrEmpty`.

I'm wondering:
- should we consolidate the `string`-checks into the `IsValidEmail` method
- is a `IsNullOrEmpty` string a "valid" EMail address?
  - I guess it's a valid (User)Feedback EMail address, but not a valid address in general

Where I am trying to get at is:
- I would prefer renaming this method from `IsValidEmail` to `IsValidContactEmail` to clarify the intent.
- Could we move the `string`-checks from the usage in `Hub` into this method, to simplify the caller.

CHANGELOG.md
```suggestion
- When CaptureFeedback methods are called with invalid email addresses, the email address will be removed and, if Debug mode is enabled, a warning will be logged. This is done to avoid losing the Feedback altogether (Sentry would reject Feedback that has an invalid email address) ([#4284](https://github.com/getsentry/sentry-dotnet/pull/4284))
```

src/Sentry/Internal/EmailValidator.cs
jamescrosswell: An email validator seems like something fairly generic that we might use elsewhere... so I'll tighten it up (so that it doesn't accept null/empty strings) and then we can leave the additional `IsNullOrWhitespace` checks in place (there's only two of them, one of which will be removed in the next major version - so we're not really consolidating much there I don't think).

src/Sentry/Internal/Hub.cs
jamescrosswell: We could add something to the docs once this has been released. I'll add a new issue to track that.

Issue #4262, opened 2025-06-06
Skip reporting blocking calls from the SDK itself


bruno-garcia:
We have some blocking calls from the SDK which cant' be made async. These are public on the sandbox: https://sandbox.sentry.io/issues/6562584726/?project=4508160335675392&query=is%3Aunresolved&referrer=issue-stream&stream_index=9

They are just wasting quota and are not helpful. We need to build rules to skip reporting these:

See:
* https://github.com/sentry-demos/empower/issues/850

Example: ![Image](https://github.com/user-attachments/assets/68d5c4e1-1b92-439a-86d8-81fea26d9681)

jamescrosswell:
@bruno-garcia would this do what you want?

https://github.com/getsentry/sentry-dotnet/blob/d1e5efcdf3af763ad49f11cd2426cc14a315a901/samples/Sentry.Samples.AspNetCore.Mvc/Controllers/HomeController.cs#L45-L48

bruno-garcia:
Possibly, do we need to add this on the background worker?

jamescrosswell:
> Possibly, do we need to add this on the background worker?

I think so yes... it would require suppression to be added to any/all methods that are referencing the `_queue` (since most of the operations in that `_queue` are using a `lock` to ensure thread safety. 

We could instead/also refactor `ConcurrentQueueLite` to use a `ReaderWriterLockSlim` instead of a `lock`. I think that would avoid blocking calls on the read operations (like [Count](https://github.com/getsentry/sentry-dotnet/blob/adb6cfb5c84771128995e640ee4401d52067cd6a/src/Sentry/Internal/ConcurrentQueueLite.cs#L35-L44), [TryPeek](https://github.com/getsentry/sentry-dotnet/blob/adb6cfb5c84771128995e640ee4401d52067cd6a/src/Sentry/Internal/ConcurrentQueueLite.cs#L56-L68) and [ToArray](https://github.com/getsentry/sentry-dotnet/blob/adb6cfb5c84771128995e640ee4401d52067cd6a/src/Sentry/Internal/ConcurrentQueueLite.cs#L70-L76)). From memory, we settled on `lock` after some benchmarking - although this stuff is happening in a background thread so perhaps we'd rather it be thread pool friendly vs as fast as possible. 🤷🏻

jamescrosswell:
[...]

PR #4288, opened 2025-06-16
fix: remove blocking detection for background worker

bruno-garcia:
Example: https://demo.sentry.io/issues/6554714607/?project=4508968088698880&query=is%3Aunresolved&referrer=issue-stream&stream_index=5

image

Fixed:
* https://github.com/getsentry/sentry-dotnet/issues/4262

### DIFF
--- a/src/Sentry/Internal/GarbageCollectionMonitor.cs
+++ b/src/Sentry/Internal/GarbageCollectionMonitor.cs
@@ -1,5 +1,5 @@
-using Sentry.Extensibility;
 using Sentry.Ben.BlockingDetector;
+using Sentry.Extensibility;
 
 namespace Sentry.Internal;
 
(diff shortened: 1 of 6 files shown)


src/Sentry/Internal/GarbageCollectionMonitor.cs
```diff
@@ -10,8 +11,13 @@ internal sealed class GarbageCollectionMonitor
     private const int MaxGenerationThreshold = 10;
     private const int LargeObjectHeapThreshold = 10;
 
-    public static Task Start(Action onGarbageCollected, CancellationToken cancellationToken, IGCImplementation? gc = null) =>
-        Task.Run(() => MonitorGarbageCollection(onGarbageCollected, cancellationToken, gc), cancellationToken);
+    public static Task Start(Action onGarbageCollected, CancellationToken cancellationToken, IGCImplementation? gc = null)
+    {
+        using (new SuppressBlockingDetection())
```
bruno-garcia, line 16: @jamescrosswell this is the blocking detector itself right? we shouldn't add this then?

jamescrosswell: Yeah don't do that...

Issue #4278, opened 2025-06-12
Partial source context in Release with multi-project MAUI apps


NathanielJS1541:
### Package

Sentry.Maui

### .NET Flavor

MAUI

### .NET Version

9.0.301

### OS

Android

### OS Version

Android API 34/35.

### Development Environment

Visual Studio v17.x

### SDK Version

6.10.0

### Workload Versions

```
Installed Workload Id      Manifest Version       Installation Source
---------------------------------------------------------------------
android                    35.0.61/9.0.100        VS 17.14.36203.30
ios                        18.4.9288/9.0.100      VS 17.14.36203.30
maccatalyst                18.4.9288/9.0.100      VS 17.14.36203.30
maui-windows               9.0.51/9.0.100         VS 17.14.36203.30
wasm-tools                 9.0.6/9.0.100          VS 17.14.36203.30

Use `dotnet workload search` to find additional workloads to install.
```

### UseSentry or SentrySdk.Init call

```csharp
            var builder = MauiApp.CreateBuilder();
            builder
                .UseMauiApp<App>()
                .UseSentry(options => {
                    options.Dsn = "";

                    // Set SDK to debug mode, as this is a repro app.
                    options.Debug = true;

                    // Application release and distribution.
                    options.Release = $"maui-sentry-test@{AppInfo.VersionString}";
                    options.Distribution = AppInfo.BuildString;

                    // Set the environment based on build configuration.
#if DEBUG
                    options.Environment = "Development";
#else
                    options.Environment = "Production";
#endif
                })
```

### Steps to Reproduce

1. Clone my [repro project](https://github.com/NathanielJS1541/MauiSentryTest).
2. Follow the [README.md](https://github.com/NathanielJS1541/MauiSentryTest/blob/main/README.md) to configure the Sentry application and organisation slugs, and DSN.
3. Build the app in `Debug` mode, and deploy to a device (i.e. Android).
4. Launch the app and click the `CommonService Exception` button.
5. Check your Sentry issues for the report, observe the behaviour from the "Expected" section.
6. Build the app in `Release` mode, and deploy to a device.
7. Launch the app and click the `CommonService Exception` button.
8. Check your Sentry issues for the report, observe the behaviour from the "Actual Result" section.

This issue seems to be exhibited specigically when an exception occurs in a project separate to the main MAUI app, and the build is trimmed / optimised.

### Expected Result

The stack trace will show line numbers and source context for all managed source:

[...]

PR #4294, opened 2025-06-19
fix: Source context for class libraries when running on Android in Release mode

jamescrosswell:
Resolves: #4278, #4342
- https://github.com/getsentry/sentry-dotnet/issues/4278
- https://github.com/getsentry/sentry-dotnet/issues/4342

> [!NOTE]
> Introducing this change caused the Android Device Tests to start failing on CI only. After a lot of trial and error, I eventually narrowed it down to this test (which has therefore been disabled in CI on Android):
> https://github.com/getsentry/sentry-dotnet/blob/93462586cf4b985f67b1a3faecf3b9e6d9c7badb/test/Sentry.Maui.Tests/SentryMauiScreenshotTests.cs#L184-L189


src/Sentry.Android.AssemblyReader/V2/AndroidAssemblyStoreReaderV2.cs
```diff
@@ -119,10 +127,12 @@ private bool FindBestAssembly(string name, out ExplorerStoreItem? explorerAssemb
         {
             if (explorer.AssembliesByName?.TryGetValue(name, out var assembly) is true)
             {
+                _logger?.Invoke("Found best assembly {0} in APK AssemblyStore for target arch {1}", name, explorer.TargetArch);
                 explorerAssembly = new(explorer, assembly);
                 return true;
             }
         }
+        _logger?.Invoke("No best assembly for {0} in APK AssemblyStore", name);
```
bruno-garcia, line 135: looks like we have one logger but shouldn't this be a level Error log?

jamescrosswell: We'd have to refactor the AssemblyReader classes that we vendored in fomr MS in order to be able to do that (they use a different pattern for logging). 

Probably the easiest would be to replace [DebugLogger](https://github.com/getsentry/sentry-dotnet/blob/4ee96eeaf000d8f1d426a914622548a473616bd3/src/Sentry.Android.AssemblyReader/DebugLogger.cs#L8) with our own `IDiagnosticLogger`...  would involve quite a few changes, but then we've made so many changes to these already that the only way to merge in changes from new code is by hand. So a few more changes wouldn't hurt.

jamescrosswell:
> [!NOTE]
> This behaviour only manifests in CI (with the headless runners). Device tests work fine when run locally on a simulator. 

Something very unusual happening with the device tests. Running against API levels 27 and 33, 27 always fails. 33 fails frequently (but not always).

Since Google require folks target 34 and higher to be able to release anything on the play store now, I figured I'd bump the API levels we test against to 34 (oldest supported) and 36 (latest). The also both seem to fail. 

Looking at the logs, there don't appear to be any failed tests. 

In one run, I noticed the last line that gets logged from .NET SDK is:
```
07-14 05:34:09.084  4252  4294 I DOTNET  : 	[IGNORED] Sentry.Maui.Tests.SentryMauiScreenshotTests.CaptureException_BeforeCaptureScreenshot_DefaultAsync
```

Shortly after that I see some logs from sentry-native:
```
07-14 05:34:10.213  4252  5450 I sentry-native: entering signal handler
07-14 05:34:10.225  4252  4280 I sentry-native: entering signal handler
07-14 05:34:10.229  4252  5450 D sentry-native: captured backtrace from ucontext with 1 frames
07-14 05:34:10.232  4252  5450 D sentry-native: captured backtrace with 1 frames
07-14 05:34:10.232  4252  5450 D sentry-native: merging scope into event
07-14 05:34:10.232  4252  5450 D sentry-native: trying to read modules from /proc/self/maps
...
07-14 05:34:11.726  4252  5439 I sentry-native: entering signal handler
...
07-14 05:34:11.828  4252  5450 D sentry-native: read 475 modules from /proc/self/maps
07-14 05:34:11.828  4252  5450 D sentry-native: adding attachments to envelope
07-14 05:34:11.828  4252  5450 D sentry-native: sending envelope
07-14 05:34:11.829  4252  5450 I sentry-native: crash has been captured
``` 

So it looks like potentially there is a low level crash that sentry-native is picking up and sending as an envelope. To where though and from which test?

Even then, this was only on the Run against API level 34... level 36 doesn't show any native errors.

jamescrosswell:
The issues running android device tests in CI seem to manifest when running the `SentryMauiOptionsTests`, which is quite surprising since those tests are blindingly simple. 

About my only hypothesis is that, by default, SentryOptions creates a new `PollingNetworkStatusListener`:
https://github.com/getsentry/sentry-dotnet/blob/723c84e6fe8a642be7ab2d40caf401bb6d38a8e3/src/Sentry/SentryOptions.cs#L1326

How this is in any way related to the changes this PR makes to the Android Assembly Readers in .NET 9 escapes me... but I'll try substituting in a Fake reliable network status listener. In a MAUI app, a `MauiNetworkStatusListener` would actually be used instead of either of these... but these particular tests have nothing to do with network status listeners so a fake should be fine (and it's easier to wire up).

jamescrosswell:
In certain test runs, if I comment these lines out then the tests pass;
https://github.com/getsentry/sentry-dotnet/blob/849d704097f9f17631406e3e9ef5592bbc2f0b18/src/Sentry.Maui/SentryMauiOptions.cs#L26-L29

However later on, even after having refactored this to happen later (in an extension) I the device tests seemed to fail... presumably for other reasons. 

Perhaps then these lines are unrelated to the test failures. 

Regardless, the test failures/success seems to be flaky - the tests succeed periodically but fail most of the time, without any changes to the tests 😢

src/Sentry.Maui/SentryMauiOptions.cs
```diff
@@ -23,18 +23,14 @@ public SentryMauiOptions()
 #if !PLATFORM_NEUTRAL
         CacheDirectoryPath = Microsoft.Maui.Storage.FileSystem.CacheDirectory;
 #endif
-        AddDefaultEventBinder<MauiButtonEventsBinder>();
-        AddDefaultEventBinder<MauiImageButtonEventsBinder>();
-        AddDefaultEventBinder<MauiGestureRecognizerEventsBinder>();
-        AddDefaultEventBinder<MauiVisualElementEventsBinder>();
```
jamescrosswell, line 29: See https://github.com/getsentry/sentry-dotnet/pull/4294#issuecomment-3094833660

src/Sentry.Maui/SentryMauiOptions.cs
```diff
@@ -23,18 +23,14 @@ public SentryMauiOptions()
...
     }
 
-    internal List<IMauiElementEventBinderRegistration> DefaultEventBinders { get; } = [];
+    internal List<IMauiElementEventBinderRegistration> InjectedEventBinders { get; } = [];
```
jamescrosswell: Potentially we could call this property `EventBindersFromIntegrations`... at the moment, there's only one.

jamescrosswell:
OK after days of fluffing about disabling and enabling different tests in a protracted divide and conquer algorithm, it looks like this is the culprit (assuming there are no other flaky tests)... so I've disabled this in CI on Android:

https://github.com/getsentry/sentry-dotnet/blob/93462586cf4b985f67b1a3faecf3b9e6d9c7badb/test/Sentry.Maui.Tests/SentryMauiScreenshotTests.cs#L184-L189

src/Sentry.Android.AssemblyReader/V2/AndroidAssemblyStoreReaderV2.cs
jamescrosswell: @bruno-garcia one remaining question on this PR is whether we want to refactor all the vendored in code to use our  `IDiagnosticLogger` instead of the `DebugLogger` delegate that is used currently: 
https://github.com/getsentry/sentry-dotnet/blob/4ee96eeaf000d8f1d426a914622548a473616bd3/src/Sentry.Android.AssemblyReader/DebugLogger.cs#L8

The existing delegate doesn't have any notion of "log levels" (so doesn't support what you suggested above).

bruno-garcia: I wouldn't add that as a requirement to unblock this. But it's not unreasonable to add if it'll help us debug this better going forward

src/Sentry.Maui/Internal/MauiImageButtonEventsBinder.cs
```diff
@@ -21,6 +21,7 @@ public void Bind(VisualElement element, Action<BreadcrumbEvent> addBreadcrumb)
     /// <inheritdoc />
     public void UnBind(VisualElement element)
     {
+        addBreadcrumbCallback = null;
```
bruno-garcia: differently than the other field, which is _addBread...

(missing _)

bruno-garcia (approved):

test/Sentry.Maui.Device.TestApp/Sentry.Maui.Device.TestApp.csproj
```diff
@@ -100,6 +100,9 @@
     <ProjectReference Include="..\Sentry.Extensions.Logging.Tests\Sentry.Extensions.Logging.Tests.csproj"/>
     <ProjectReference Include="..\Sentry.Maui.Tests\Sentry.Maui.Tests.csproj"/>
     <ProjectReference Include="..\Sentry.Maui.CommunityToolkit.Mvvm.Tests\Sentry.Maui.CommunityToolkit.Mvvm.Tests.csproj" Condition="'$(_SentryIsNet9OrGreater)' == 'true'"/>
+    <ProjectReference Include="..\..\src\Sentry.SourceGenerators\Sentry.SourceGenerators.csproj"
```
Flash0ver, line 103: question: Is the `<ItemGroup>` with the `CompilerVisibleProperty` items here imported as well?
(via `src/Sentry/buildTransitive/Sentry.targets`)

I have the feeling, that our source generator here does run, but actually does not emit any sources.

src/Sentry.Android.AssemblyReader/V2/AndroidAssemblyStoreReaderV2.cs
jamescrosswell: k, I'll track that in a separate issue as I think we should get this fix out ASAP:
- https://github.com/getsentry/sentry-dotnet/issues/4382

test/Sentry.Maui.Device.TestApp/Sentry.Maui.Device.TestApp.csproj
jamescrosswell: It should be imported here for test projects:
https://github.com/getsentry/sentry-dotnet/blob/e59088663b6948a9cfc6e7eb576d628dd418a2e1/test/Directory.Build.targets#L34

bruno-garcia (approved):
lgtm

test/Sentry.Maui.Device.TestApp/Sentry.Maui.Device.TestApp.csproj
Flash0ver: oh .. right .. totally missed that

Issue #4299, opened 2025-06-20
test(ci): Visual Studio MSBuild on Windows


Flash0ver:
### Description

Recently, we run into issues with Sentry-package-consumer builds on _Windows_ via _Visual Studio's MSBuild_ and/or the .NET SDK bundles with _Visual Studio_: #4141, #4292

Let's improve and continuously test the development experience on Windows via Visual Studio for both building and consuming the Sentry .NET SDKs.

Option to catch related issues early:
Add a CI build step on Windows-only that utilizes Visual Studio
- [setup-msbuild Marketplace](https://github.com/marketplace/actions/setup-msbuild)
- [setup-msbuild Repository](https://github.com/microsoft/setup-msbuild)

Option to test related issues:
Add an integration test for Windows-only that utilizes Visual Studio

bruno-garcia:
Last resort, but if we need to we can use our self hosted runner where we have devkits. We'd need to have a private fork for that though so would need to be a pre-release step only or something, not a per PR check.

Flash0ver:
After #4300, we could still add more coverage for the user experience on Windows and Visual Studio via Integration Tests.

E.g.
- install Visual Studio, or the respective MSBuild tools, or the respective .NET SDK
  - matrix: supported (latest) versions of Visual Studio
    - [Visual Studio 2019 version 16.11)(https://learn.microsoft.com/visualstudio/releases/2019/servicing-vs2019) / .NET 5.0
    - Visual Studio 2022 / .NET 8.0 or .NET 9.0 (not sure if .NET 10 will remain fully supported on VS2022)
    - Visual Studio 2026 / .NET 10+
- create App
  - .NET Console App or UWP App or ...
  - Use Sentry
- add `Sentry` package
  - may pick `netstandard2.0` lib depending on the target
- build app
  - may catch incompatibilities in the `.props` and `.targets` that we ship
  - may catch incompatibilities in the Roslyn Compiler Extensions (e.g. Analyzers, Source-Generators, ...) that we ship
- run app
  - may catch other issues at runtime

PR #4297, opened 2025-06-19
WIP: WinUI & Native

jpnurmi:
> [!NOTE]
> Blocked by:
> - microsoft/microsoft-ui-xaml#10635

Experiment to see what it takes to capture native crashes for WinUI apps:
- #3244

#skip-changelog

Issue #4292, opened 2025-06-18
Maui iOS build error(s): The xcframework sentry.bindings.cocoa\5.10.0\lib\net8.0-ios17.0\Sentry.Bindings.Cocoa.resources.zip has an incorrect or unknown format and cannot be processed.


WeaverSystems:
### Package

Sentry

### .NET Flavor

.NET

### .NET Version

9.0

### OS

iOS

### Development Environment

Visual Studio v17.x

### SDK Version

5.10

### Workload Versions

-

### UseSentry or SentrySdk.Init call
```
                .UseSentry(options =>
                {
                    //The DSN is the only required setting
                    options.Dsn = **************;

#if DEBUG             
                    options.Debug = true;
                    options.TracesSampleRate = 1.0;
#else
                    options.TracesSampleRate = 0.2;
#endif

                    //All other Sentry options
                    options.IsGlobalModeEnabled = true;
                    options.IncludeTextInBreadcrumbs = true;
                    options.IncludeTitleInBreadcrumbs = true;
                    options.IncludeBackgroundingStateInBreadcrumbs = true;
                    options.AttachStacktrace = true;
                    options.CaptureFailedRequests = true;
                });
```
### Steps to Reproduce

1. Update Visual Studio to 17.14.6 (June 2025).
2. Try and build a Maui app with iOS and/or MacCatalyst platforms included.

### Expected Result

Everything should build with no errors.

### Actual Result

When trying to build our Maui project we're getting the following 4 errors:
- The xcframework C:\Users\******\.nuget\packages\sentry.bindings.cocoa\5.10.0\lib\net8.0-ios17.0\Sentry.Bindings.Cocoa.resources.zip has an incorrect or unknown format and cannot be processed.
- Can't process the zip file 'C:\Users\******\..nuget\packages\sentry.bindings.cocoa\5.10.0\lib\net8.0-maccatalyst17.0\Sentry.Bindings.Cocoa.resources.zip' on this platform: the file 'Sentry-Dynamic.xcframework/ios-arm64_arm64e_x86_64-maccatalyst/Sentry.framework/PrivateHeaders' is a symlink.
[...]

PR #4300, opened 2025-06-22
chore: fix Visual Studio build

jpnurmi:
[System.Runtime.InteropServices.RuntimeInformation::RuntimeIdentifier](https://learn.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.runtimeinformation.runtimeidentifier) is only available in .NET 5.0 and later. While `dotnet build` on the command line works, building Sentry from within Visual Studio throws such errors:

```
Sentry.Native.targets(37,14): error MSB4186: Invalid static method invocation syntax:
  "[System.Runtime.InteropServices.RuntimeInformation]::RuntimeIdentifier".
  Method 'System.Runtime.InteropServices.RuntimeInformation.RuntimeIdentifier' not found.
  Static method invocation should be of the form: $([FullTypeName]::Method()), e.g. $([System.IO.Path]::Combine(`a`, `b`)).
  Check that all parameters are defined, are of the correct type, and are specified in the right order.
```

TODO: Is `$(NETCoreSdkRuntimeIdentifier)` guaranteed to be set on all supported platforms and configurations, or do we need some manual fallback resolution based on the host OS and arch?

See also:
- #4299

#skip-changelog


Flash0ver:
Thanks @jpnurmi for looking into this ... I'm facing this issue on my Windows-machine right now.

jpnurmi:
@Flash0ver Did you experiment with MSBuild in some branch for the other build problem you investigated earlier? Was it something like this you had in mind?

Flash0ver:
@jpnurmi ... no ... IIRC ... this is something different ... but I may misunderstand something here
- this PR fixes the build of the `sentry-dotnet` source from Visual Studio and the non-.NET-SDK MSBuild
- #4299 refers to issues of building projects consuming a `Sentry`-Package via Visual Studio and the non-.NET-SDK MSBuild

I have just recently started experimenting with either scenario of building with Visual Studio / the non-.NET-SDK MSBuild.
- just 2 days ago to reproduce #4292
- and just now building `sentry-dotnet` from source facing
image

Locally only.
I created no branch so far.

I tried
```diff
-<ItemGroup Condition="'$(CI_PUBLISHING_BUILD)' == 'true' or '$([System.Runtime.InteropServices.RuntimeInformation]::RuntimeIdentifier)' == 'win-x64'">
+<ItemGroup Condition="'$(CI_PUBLISHING_BUILD)' == 'true' or '$(NETCoreSdkRuntimeIdentifier)' == 'win-x64'">
```
with the latest stable Visual Studio ... which seems to fix the build issue.
But I haven't checked yet, whether all _Native_ features still work as expected.

jpnurmi:
I got the same error while experimenting with WinUI (https://github.com/getsentry/sentry-dotnet/pull/4297) and tried adding Sentry as a project reference for prototyping purposes. Makes me wonder if this change should be marked as a fix instead because it impacts those who might be using Sentry as a project reference. Not sure how common it is to do that outside the Sentry project tree, though.

Flash0ver:
I don't think it's of type `fix:`, as I don't have the feeling this warrants a `CHANGELOG` entry.
Not sure if `build:` would be applicable here. But also not quite a `chore:`.
Hmm ... I'm uncertain ... I guess either of these is fine.

src/Sentry/Platforms/Native/Sentry.Native.targets
```diff
@@ -34,35 +34,35 @@
   </ItemGroup>
 
   <!-- Packaging the native library (NOTE: $(RuntimeIdentifier) is not set) -->
-  <ItemGroup Condition="'$(CI_PUBLISHING_BUILD)' == 'true' or '$([System.Runtime.InteropServices.RuntimeInformation]::RuntimeIdentifier)' == 'win-x64'">
+  <ItemGroup Condition="'$(CI_PUBLISHING_BUILD)' == 'true' or '$(NETCoreSdkRuntimeIdentifier)' == 'win-x64'">
```
Flash0ver, line 37: question: `NETCoreSdkRuntimeIdentifier`

I haven't seen `NETCoreSdkRuntimeIdentifier` so far.
Do I understand it correctly that the difference to `RuntimeIdentifier` is that it's not referring to the target platform(s) an application is being built for, but the platform of the currently running build process itself (_.NET SDK_ / _MSBuild_)?

jamescrosswell: It looks like the behaviour for `$([System.Runtime.InteropServices.RuntimeInformation]::RuntimeIdentifier)` [changed significantly in .NET 8](https://learn.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.runtimeinformation.runtimeidentifier?view=net-9.0#remarks). 

Interesting comment here:
- https://github.com/dotnet/sdk/pull/14093#issuecomment-708555022

Flash0ver: Cool ... I also didn't know `MSBuildRuntimeType` ... we could use it - in the future - to Multi-Target our MSBuild code between `Full`/`Core`/`Mono`.
(TIL where the term "Full Framework" is coming from)

Also, more info about the breaking change:
https://learn.microsoft.com/en-us/dotnet/core/compatibility/core-libraries/8.0/runtimeidentifier

But for this change:
`NETCoreSdkRuntimeIdentifier` looks good and Tests are green and I took a quick peek at the produced NuGet package ✅

Flash0ver (approved):

src/Sentry/Platforms/Native/Sentry.Native.targets
jamescrosswell: So we're good to merge?

The only thing that makes me slightly nervous is that there is very little documentation for `NETCoreSdkRuntimeIdentifier`. The only place I found it in the MS Docs was buried in these [additional notes](https://learn.microsoft.com/en-us/dotnet/core/native-interop/expose-components-to-com#additional-notes). It's not prefixed with an `_` character though, so presumably it's meant for public consumption.

jpnurmi: Ready to go if it's up to me. I'm just feeling shy to merge and wanted to let you chime in. :)

PR #4304, opened 2025-06-25
Custom middleware tracing

### LINKED ISSUE #3423: Tracing not registered automatically when using `WebApplication.CreateBuilder` with `UseRouting`
### Problem Statement

```csharp
// Program.cs

WebApplicationBuilder builder = WebApplication.CreateBuilder(args);
builder.WebHost.UseSentry();

WebApplication app = builder.Build();
app.Use(async (context, next) =>
{
    ISpan s1 = SentrySdk.GetSpan(); // null
    await next();
    ISpan s2 = SentrySdk.GetSpan(); // null
});
app.UseRouting();
app.MapControllers();

app.Run();
```

Tracing works fine inside my controller and all associated filters, but I would like to trace middleware operations in the same transaction as well, which does not seem possible with the current SDK implementation (`SentrySdk.GetSpan()` is `null` in the middleware).

### Solution Brainstorm

Maybe it should be (is?) possible to start the transaction earlier in the ASP.NET pipeline?

jamescrosswell:
Resolves #3423:
- https://github.com/getsentry/sentry-dotnet/issues/3423

### Example

image

#skip-changelog

### DIFF
--- a/samples/Sentry.Samples.AspNetCore.Basic/Program.cs
+++ b/samples/Sentry.Samples.AspNetCore.Basic/Program.cs
@@ -19,8 +19,7 @@ builder.WebHost.UseSentry(options =>
 
 var app = builder.Build();
 
-// An example ASP.NET Core middleware that throws an
-// exception when serving a request to path: /throw
+// An example ASP.NET Core endpoint that throws an exception when serving a request to path: /throw
 app.MapGet("/throw/{message?}", context =>
 {
     var exceptionMessage = context.GetRouteValue("message") as string;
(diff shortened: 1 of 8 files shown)


samples/Sentry.Samples.AspNetCore.Basic/Program.cs
```diff
@@ -48,4 +47,23 @@
...
+    {
+        var log = context.RequestServices.GetRequiredService<ILoggerFactory>()
+            .CreateLogger<Program>();
+
+        log.LogInformation("Just chilling for a bit...");
+        await Task.Delay(TimeSpan.FromMilliseconds(50)); // Simulate some work
+    }
+    finally
+    {
+        span.Finish();
```
bruno-garcia: updates the span status to a failure, and links the span and the exception
```suggestion
        span.Finish();
    }
    catch (Exception e)
    {
        span.Finish(e);
        throw;
```

Flash0ver (approved):

PR #4308, opened 2025-06-25
feat: add Sentry Logs (experimental)

Flash0ver:
Add support, API and integrations for Sentry Logs
- https://docs.sentry.io/product/explore/logs/
- https://develop.sentry.dev/sdk/telemetry/logs/

Integration branch / PR of
- https://github.com/getsentry/sentry-dotnet/pull/4158
- https://github.com/getsentry/sentry-dotnet/pull/4310
- https://github.com/getsentry/sentry-dotnet/pull/4193

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,8 +4,10 @@
 
 ### Features
 
-- Add experimental support for [Sentry Structured Logging](https://docs.sentry.io/product/explore/logs/) API ([#4158](https://github.com/getsentry/sentry-dotnet/pull/4158), [#4310](https://github.com/getsentry/sentry-dotnet/pull/4310))
-  - Add experimental integrations for `Sentry.Extensions.Logging`, `Sentry.AspNetCore` and `Sentry.Maui` ([#4193](https://github.com/getsentry/sentry-dotnet/pull/4193))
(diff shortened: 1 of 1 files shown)


src/Sentry.Extensions.Logging/SentryStructuredLogger.cs
```diff
@@ -0,0 +1,121 @@
...
+        _categoryName = categoryName;
+        _options = options;
+        _clock = clock;
+        _hub = hub;
+        _sdk = sdk;
+    }
+
+    public IDisposable? BeginScope<TState>(TState state) where TState : notnull
+    {
+        return NullDisposable.Instance;
```
jamescrosswell, line 26: What's our plan for this then? Do we just not support BeginScope for structured logging? Or are we relying on the scope having been pushed somewhere else?

jamescrosswell (approved):
Looks like we're 95% there (once it's passing CI). 

@Flash0ver and I discussed merging this logger with our other logger, so that we can elegantly handle things like `ILogger.BeginScope` but this is already pretty solid for an experimental feature so let's get it out there and get some feedback while we work on those other things. 

Nice work Stefan! ❤️ 🚀

Issue #4306, opened 2025-06-25
Improvements for logging transport


bruno-garcia:
### Description

Some ideas to improve logging transport:

If logging is enabled, and the `options.Transport` is not a custom/user defined one. And offline caching is not set.
We could have a second background worker just for logs. Both the main BW and the logs BW could share the transport (the HTTP one).

We should also consider increasing the default maxQueue from 30 to 100 or so.

jamescrosswell:
Will this be necessary if we're batching logs? If we only ever send one envelope every 500 ms then the queue size shouldn't be a bottleneck right? And the only reason we'd need another background worker would be to prevent other envelopes from blocking the delivery of logs.

jamescrosswell:
With the current implementation of batching, this no longer seems pressing… and may be completely unnecessary once we implement priority queues. Will mark as "Won't Do" for now... we can always circle back in the future if thinking changes.

Issue #3689, opened 2024-10-16
Leverage new Lock types


jamescrosswell:
C# 13 includes a [new Lock type](https://learn.microsoft.com/en-us/dotnet/csharp/whats-new/csharp-13#new-lock-object) intended to be used with the lock keyword.

Using Lock types as lock objects, the compiler can perform some optimisations that aren't possible when using standard object lock objects. It's a small thing, but we should probably replace the lock objects in our SDK to enable these optimisations.

Flash0ver:
[Polyfill v6.12.0](https://github.com/SimonCropp/Polyfill/milestone/101) adds the `System.Threading.Lock` type.
It satisfies the shape for the "C# 13 lock statement".
The implementation uses `System.Threading.Monitor` under the hood.

I am not 100 % certain that this is 100 % correct, because there is a difference in IL-code-gen between the two `lock`-variants:
```diff
-object obj = new object();
-bool lockTaken = false;
+Lock @lock = new Lock();
+Lock.Scope scope = @lock.EnterScope();
try
{
- Monitor.Enter(obj, ref lockTaken);
  Console.WriteLine("mutual-exclusion");
}
finally
{
- if (lockTaken)
- {
-   Monitor.Exit(obj2);
- }
+ scope.Dispose();
}
```

where - with the polyfilled version on `<= net8.0` - the implemented `Monitor.Enter` called via `EnterScope` would then be outside of the `try`-block.

Could this cause problems for our use cases?

Alternatively, we could `#if NET9_0_OR_GREATER` all usages of `System.Threading.Lock`, and use `object` instead, but that could bloat the source code quite a bit.
Alternatively, we could alias `using Lock = object;` globally (e.g. via C# or MSBuild) for `#if !NET9_0_OR_GREATER`.

Just some thoughts for the implementer.

jamescrosswell:
> Alternatively, we could #if NET9_0_OR_GREATER all usages of System.Threading.Lock, and use object instead, but that could bloat the source code quite a bit.

Yeah that's why we haven't done this already. That seemed like a pretty ugly solution (especially since we have no prospect of ever being able to drop net472 or netstandard). 

I think the Polyfill implementation of Lock is the best option. The IL will never be the same as the net9.0 version but that's OK as long as we don't take a step backwards, performance wise, over what we have now.

Flash0ver:
After a discussion
[...]

PR #4310, opened 2025-06-26
feat(logs): add Buffering and Batching

Flash0ver:
Add minimal support for Buffering / Batching for Sentry Logs.
Planned follow-up to avoid having high-frequency logging drop other, potentially more important, envelopes: #4306

#### Changes (Updated):
- `SentryStructuredLogger` is now `IDisposable`
  - and is `Dispose`d when `Hub` is `Dispose`d
  - also `Flush`es all logs when `Hub` is `Dispose`d
- the user-facing `SentryLog` is no longer `ISentryJsonSerializable`
- but the new _internal_ `StructuredLog` is `ISentryJsonSerializable`
  - this new _internal_ type groups one or many `SentryLog` instances for serialization
- the `DefaultSentryStructuredLogger` no longer directly captures and enqueues Log-Envelopes
  - but indirectly via the new `StructuredLogBatchProcessor` instead
- new type `StructuredLogBatchProcessor`
  - buffers Logs
    - double-buffer strategy
    - swap buffer when full or timeout exceeded
  - sends Logs (`Flush`) when
    - Count-Threshold is reached
    - Timeout is reached
  - utilizes a `lock`-free solution, based on a `CountdownEvent` that only signals/`Wait`s when `Flush`ing
- using a new helper type, `StructuredLogBatchBuffer`
  - a wrapper over an `Array`
  - that provides thread-safe `Add` and `Flush` operations
- using a new synchronization primitive, `ScopedCountdownLock`
  - based on `CountdownEvent`
  - has an initial count / offset of `1`
    - when the offset has not been signaled yet, cannot Set the event (used for tracking active `Enqueue/Add` operations
    - when the Lock is engaged (when a `Flush` is requested), can no longer `Enqueue/Add` and can `Wait` until existing `Enqueue/Add` operations have completed
    - after the `Flush` has completed, disengage the Lock to reset the count to the offset of `1` to allow `Enqueue/Add` operations again
- add _Benchmark_ for `StructuredLogBatchProcessor`

---
#skip-changelog


src/Sentry/Internal/BatchProcessor.cs
```diff
@@ -0,0 +1,99 @@
...
+/// <summary>
+/// The Sentry Batch Processor.
+/// This implementation is not complete yet.
+/// Also, the specification is still work in progress.
+/// </summary>
+/// <remarks>
+/// Sentry Specification: <see href="https://develop.sentry.dev/sdk/telemetry/spans/batch-processor/"/>.
+/// OpenTelemetry spec: <see href="https://github.com/open-telemetry/opentelemetry-collector/blob/main/processor/batchprocessor/README.md"/>.
+/// </remarks>
+internal sealed class BatchProcessor : IDisposable
```
jamescrosswell: We could think about giving this a more specific/descriptive name. `BatchProcessor` is very generic and we could have other batch processors (for metrics or other things) in the future. 

If we weren't worried about long names, I guess it would be a `StructuredLogBatchProcessor`... Or a more concise alternative might be `LogBatch` (so you'd do something like `_logBatch.Enqueue(log)`)?

Alternatively, it could be made generic (like the `BatchBuffer` class that it uses), but I'd recommend we do that if/when we need a batch processor for a second category of Sentry events (it's hard to make something properly generic until you have 2 or 3 different concrete implementations).

src/Sentry/Internal/BatchProcessor.cs
```diff
@@ -0,0 +1,99 @@
+using System.Timers;
+using Sentry.Protocol;
+using Sentry.Protocol.Envelopes;
+
+#if NET9_0_OR_GREATER
+using Lock = System.Threading.Lock;
+#else
+using Lock = object;
```
jamescrosswell: It looks like Simon added a Polyfill for this that we could use to avoid repeating code like this throughout the repo:
- https://github.com/SimonCropp/Polyfill/pull/230

src/Sentry/Internal/BatchProcessorTimer.cs
jamescrosswell, line 1: Seems like an unnecessary abstraction - ideally we could delete this file and just use the underlying system type like `System.Timers.Timer` directly.

If this needs to be mockable or something, it could be named something generic like `TimerAdapter`.

src/Sentry/Internal/BatchProcessor.cs
```diff
@@ -0,0 +1,99 @@
...
+/// </summary>
+/// <remarks>
+/// Sentry Specification: <see href="https://develop.sentry.dev/sdk/telemetry/spans/batch-processor/"/>.
+/// OpenTelemetry spec: <see href="https://github.com/open-telemetry/opentelemetry-collector/blob/main/processor/batchprocessor/README.md"/>.
+/// </remarks>
+internal sealed class BatchProcessor : IDisposable
+{
+    private readonly IHub _hub;
+    private readonly BatchProcessorTimer _timer;
+    private readonly BatchBuffer<SentryLog> _logs;
```
jamescrosswell: It looks like we have a `_lock ` and a `_logs` buffer here. 

Any reason not to use a `ConcurrentQueue` or a [ConcurrentQueueLite](https://github.com/getsentry/sentry-dotnet/blob/adb6cfb5c84771128995e640ee4401d52067cd6a/src/Sentry/Internal/ConcurrentQueueLite.cs)? It'd be good to have fewer classes to maintain (unless they're truly necessary).

src/Sentry/Internal/BatchProcessor.cs
```diff
@@ -0,0 +1,99 @@
...
+
+    private void Flush()
+    {
+        _lastFlush = DateTime.UtcNow;
+
+        var logs = _logs.ToArrayAndClear();
+        _ = _hub.CaptureEnvelope(Envelope.FromLog(new StructuredLog(logs)));
+    }
+
+    private void IntervalElapsed(object? sender, ElapsedEventArgs e)
```
jamescrosswell: Suggest renaming this to `OnIntervalElapsed` or `DoIntervalElapsed`.

src/Sentry/Internal/BatchProcessor.cs
```diff
@@ -0,0 +1,99 @@
...
+
+    internal void Enqueue(SentryLog log)
+    {
+        lock (_lock)
+        {
+            EnqueueCore(log);
+        }
+    }
+
+    private void EnqueueCore(SentryLog log)
```
jamescrosswell: This method is only called from the internal `Enqueue` method. I guess what it mainly does is avoid a bit of code indentation to make the `Enqueue` method easier to read... in which case it could be made an inline function of that method.

src/Sentry/Internal/BatchProcessor.cs
```diff
@@ -0,0 +1,99 @@
...
+    {
+        var empty = _logs.IsEmpty;
```
jamescrosswell: Suggest renaming this variable to `isFirstLog`... we're determining that by checking whether `_logs` was empty before adding the log but ultimately that's what we care about.

Later on, this then becomes a bit more readable/obvious:
```csharp
        if (isFirstLog && !_logs.IsFull)
        {
            _timer.Enabled = true;
        }
```

src/Sentry/Internal/BatchProcessor.cs
```diff
@@ -0,0 +1,99 @@
...
+        var added = _logs.TryAdd(log);
+        Debug.Assert(added, $"Since we currently have no lock-free scenario, it's unexpected to exceed the {nameof(BatchBuffer<SentryLog>)}'s capacity.");
+
+        if (empty && !_logs.IsFull)
```
jamescrosswell: What's the scenario in which `empty` and `_logs.IsFull` would both be true? Is that if we've configured a maximum capacity of 1?

src/Sentry/Internal/BatchProcessor.cs
```diff
@@ -0,0 +1,99 @@
...
+/// <remarks>
+/// Sentry Specification: <see href="https://develop.sentry.dev/sdk/telemetry/spans/batch-processor/"/>.
+/// OpenTelemetry spec: <see href="https://github.com/open-telemetry/opentelemetry-collector/blob/main/processor/batchprocessor/README.md"/>.
+/// </remarks>
+internal sealed class BatchProcessor : IDisposable
+{
+    private readonly IHub _hub;
+    private readonly BatchProcessorTimer _timer;
+    private readonly BatchBuffer<SentryLog> _logs;
+    private readonly Lock _lock;
```
jamescrosswell: If we do want to control concurrency in this class (rather than using one of the build in data structures supporting concurrency), consider using something like `ReaderWriterLockSlim`... this is mainly useful when you have lots of reads and a few writes. However if most of the operations are writes then it doesn't buy us much/anything.

src/Sentry/Internal/DefaultSentryStructuredLogger.cs
```diff
@@ -10,13 +9,33 @@ internal sealed class DefaultSentryStructuredLogger : SentryStructuredLogger
...
         Debug.Assert(options is { Experimental.EnableLogs: true });
 
         _hub = hub;
         _options = options;
         _clock = clock;
+
+        _batchProcessor = new BatchProcessor(hub, ClampBatchCount(options.Experimental.InternalBatchSize), ClampBatchInterval(options.Experimental.InternalBatchTimeout));
+    }
+
+    private static int ClampBatchCount(int batchCount)
```
jamescrosswell: Could this go in the setter for the `InternalBatchSize` on the options instead?

If an invalid value has been configured, should we either throw an exception (we'd only do that if it happened when initialising the options - we don't generally want the SDK to throw exceptions otherwise) or at the very least log a message letting people know they're not configuring things correctly?

src/Sentry/Internal/DefaultSentryStructuredLogger.cs
```diff
@@ -10,13 +9,33 @@ internal sealed class DefaultSentryStructuredLogger : SentryStructuredLogger
...
+    {
+        return batchCount <= 0
+            ? 1
+            : batchCount > 1_000_000
+                ? 1_000_000
+                : batchCount;
+    }
+
+    private static TimeSpan ClampBatchInterval(TimeSpan batchInterval)
```
jamescrosswell: Again, could this be moved to the setter?

src/Sentry/Protocol/StructuredLog.cs
jamescrosswell, line 1: Why do we have/need two classes to represent logs now? We have `StructuredLog` and `SentryLog`... why can't we just use one class?

src/Sentry/SentryOptions.cs
```diff
@@ -1897,5 +1897,22 @@ public void SetBeforeSendLog(Func<SentryLog, SentryLog?> beforeSendLog)
...
             _beforeSendLog = beforeSendLog;
         }
+
+        /// <summary>
+        /// This API will be removed in the future.
+        /// </summary>
+        /// <remarks>
+        /// Threshold of items in the buffer when sending all items, regardless of <see cref="InternalBatchTimeout"/>.
+        /// </remarks>
+        public int InternalBatchSize { get; set; } = 100;
```
jamescrosswell: If we'll be removing this in future, presumably we need to decide what an appropriate "final" or "permanent" value for this will be. How are we going to determine that (and so what is the value in exposing this publicly, temporarily)?

src/Sentry/Internal/BatchProcessor.cs
Flash0ver: Summary: changed to `object`/`Monitor`-based `lock` on all TFMs, continuing discussion in #3689

Remarks:
I'm personally not too happy with the slightly changed semantics of the Polyfill-Lock.
The reservations that I am feeling are towards the C# 13 `System.Threading.Lock`-based lowered code that gets emitted, while the implementation of the Polyfill-Lock is still falling back to using `System.Threading.Monitor`, where we would lose some fidelity of the `object`/`Monitor`-based `lock` on older TFMs.
Especially when it comes to our targets `netstandard2.0` and `net462`, because _.NET Framework_ allows `Thread.Abort`, which may lead to some unexpected behavior when conflating the `net9.0` semantics of `lock` with a `Monitor`-based implementation.
I would personally prefer to
- `NET9_0_OR_GREATER`: use `System.Threading.Lock`-based `lock`
- `!NET9_0_OR_GREATER`: use `object`/`Monitor`-based `lock`

Or instead use a Polyfill that also carries a `net9.0`-equivalent implementation of `System.Threading.Lock`, rather than forwarding to `System.Threading.Monitor`-based implementation details.

Anyway, to avoid partial usages of the new `System.Threading.Lock` in this single / some locations of our SDKs, I changed this PR to only use `object`/`Monitor`-based `lock` on all TFMs, and we continue the discussion in #3689.

src/Sentry/Internal/BatchProcessor.cs
Flash0ver: renamed to: `OnIntervalElapsed`

src/Sentry/Internal/BatchProcessor.cs
Flash0ver: exactly ... in that case the buffer-capacity is `1`, so we need to flush/enqueue every single log immediately

src/Sentry/Internal/BatchProcessor.cs
Flash0ver: I did indeed start with a more elaborate locking-strategy with `ReaderWriterLockSlim`.
And multi-targeted with ReaderWriterLock` on `NETFRAMEWORK` where `Thread.Abort` is allowed, since this code path may be invoked from user code.

Since we do expect more writes than reads, I then simplified it to just a "full-lock-all" approach for now, with the option to optimize later.

I just did a super quick benchmark: A "flushing" call of `EnqueueCore(SentryLog)` with a capacity of `100` takes between 2 and 3 milliseconds.

I'll do a bit of experimenting and benchmarking over the weekend with different locking strategies.

src/Sentry/Protocol/StructuredLog.cs
Flash0ver: - the _internal_ `StructuredLog` represents the `log` Envelope Item
- the _public_ `SentryLog`, user facing via `SetBeforeSendLog`, represents the Payload of one `log` Envelope Item
see https://develop.sentry.dev/sdk/telemetry/logs/#logs-protocol

Currently batching up multiple `SentryLog`s, which then get - when flushed - captured, enqueued and serialized via a `StructuredLog`.

Could we solve this differently?

src/Sentry/SentryOptions.cs
Flash0ver: Our docs/spec suggests, as a starting point, to use 100 items and 5 seconds: https://develop.sentry.dev/sdk/telemetry/logs/#buffering
But to use a more elaborate batching strategy later on, via the Batch Processor - currently work in progress - https://develop.sentry.dev/sdk/telemetry/spans/batch-processor/
The Batch Processor should also be used for span streaming.
The current guideline is: start with 5 seconds, with the option to go as high as 30 seconds, with an auto-flush of _1MiB_ in size.
For the _1MiB_, we could utilize `Utf8JsonWriter.BytesCommitted + Utf8JsonWriter.BytesPending`, but that would require a bit of rewriting / redesigning our Background-Worker / Transport layer.

We could also keep these options, but make them `internal`, to be able to choose different values depending on the SDK / Integration.

I was thinking about - during the `[Experimental]` phase - trying to explore the best initial values during dogfooding.

What do you think?

bruno-garcia:
@sentry review

src/Sentry/Protocol/StructuredLog.cs
jamescrosswell: Ah I see... SentryLog is actually a separate envelope type in the Sentry protocol... specifically used to serialise an array of logs. I wish we'd pluralised the name on this but anyway, looks like we're doing things by the book here.

src/Sentry/SentryOptions.cs
jamescrosswell: It depends where we expect to collect the feedback that will inform ultimately how these values are determined. 

If we're expecting SDK users to play around with different values and get back to us to let us know what the ideal values for these things are, then it makes sense for these options to be public. 

From the docs, it doesn't look like that's not how we'll decide the values to be used here... so I reckon we either bake in the "simple approach to buffer" ([described here](https://develop.sentry.dev/sdk/telemetry/logs/#buffering)) or implement the two triggers defined in the [specification](https://develop.sentry.dev/sdk/telemetry/spans/batch-processor/#specification) (5s or 1 MiB). The 1MiB trigger would require we calculate the size of each `SentryLog` and track the cumulative size of the logs in the `StructuredLog`.

Either way, I don't think we surface this up in the options to user. We could hold it in the state for the `BatchProcessor` instead.

src/Sentry/Internal/BatchProcessorTimer.cs
```diff
@@ -0,0 +1,61 @@
...
+    {
+        get => _timer.Enabled;
+        set => _timer.Enabled = value;
+    }
+
+    public override event EventHandler<ElapsedEventArgs>? Elapsed;
+
+    private void OnElapsed(object? sender, ElapsedEventArgs e)
+    {
+        Elapsed?.Invoke(sender, e);
```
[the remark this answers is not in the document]
jamescrosswell: Not necessary if we put a try/catch inside the Elapsed implementation (which the seer-by-sentry recommended above).

src/Sentry/Protocol/StructuredLog.cs
```diff
@@ -0,0 +1,42 @@
...
+/// Sentry Docs: <see href="https://docs.sentry.io/product/explore/logs/"/>.
+/// Sentry Developer Documentation: <see href="https://develop.sentry.dev/sdk/telemetry/logs/"/>.
+/// </remarks>
+internal sealed class StructuredLog : ISentryJsonSerializable
+{
+    private readonly SentryLog[] _items;
+
+    public StructuredLog(SentryLog log)
+    {
+        _items = [log];
```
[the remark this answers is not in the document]
jamescrosswell: This would never get called though right (because we'd only ever be calling this from the BatchProcessor, which would be passing an array)?

src/Sentry/Internal/BatchBuffer.cs
```diff
@@ -0,0 +1,80 @@
...
+    internal void Clear()
+    {
+        if (_count == 0)
+        {
+            return;
+        }
+
+        var count = _count;
+        _count = 0;
+        Array.Clear(_array, 0, count);
```
bruno-garcia: Might want to go with `Buffer.BlockCopy()`, not sure it's still relevant, but back in the day if this was critical path, there was a perf advantage. Since we cal it to `TooArrayAndClear` I imagine might be but worth profiling it before doing any perf improvement changes

src/Sentry/Internal/BatchProcessor.cs
```diff
@@ -0,0 +1,93 @@
...
+        : this(hub, batchCount, new TimersBatchProcessorTimer(batchInterval))
+    {
+    }
+
+    public BatchProcessor(IHub hub, int batchCount, BatchProcessorTimer timer)
+    {
+        _hub = hub;
+
+        _timer = timer;
+        _timer.Elapsed += OnIntervalElapsed;
```
bruno-garcia: are going to have more than one subscriber?
I'd avoid events. It's not great with asynchronous code (`async void`), if one subscribers throws the second might not get the message etc.

How about a Func or Action?

src/Sentry/Internal/BatchProcessor.cs
```diff
@@ -0,0 +1,93 @@
...
+        Debug.Assert(added, $"Since we currently have no lock-free scenario, it's unexpected to exceed the {nameof(BatchBuffer<SentryLog>)}'s capacity.");
+
+        if (isFirstLog && !_logs.IsFull)
+        {
+            _timer.Enabled = true;
+        }
+        else if (_logs.IsFull)
+        {
+            _timer.Enabled = false;
+            Flush();
```
bruno-garcia: We're doing way too much work in the critical path.

Think about what absolutely must happen inside the lock, and what we can move/copy around to have it happen async.

Once the reference that's mutated concurrently is swapped with a new one, new messages can start logging, and you can continue async doing capture stuff. Like allocating the strucutered log, envelope, calling Capture etc.

bruno-garcia: We might be better off using a couple of arrays and doing some reference swap.

src/Sentry/Internal/BatchProcessorTimer.cs
```diff
@@ -0,0 +1,61 @@
...
+{
+    protected BatchProcessorTimer()
+    {
+    }
+
+    public abstract bool Enabled { get; set; }
+
+    public abstract event EventHandler<ElapsedEventArgs> Elapsed;
+
+    public void Dispose()
```
bruno-garcia: this is better at the end of the file, so we start reading what really makes this exist isntead of ,NET boilerplate

src/Sentry/Internal/BatchProcessorTimer.cs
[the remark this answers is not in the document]
bruno-garcia: we should add it on the caller side though instead of hoping subscribers are doing the right thing

src/Sentry/Internal/BatchProcessorTimer.cs
```diff
@@ -0,0 +1,61 @@
...
+    {
+        _timer = new System.Timers.Timer(interval.TotalMilliseconds)
+        {
+            AutoReset = false,
+            Enabled = false,
+        };
+        _timer.Elapsed += OnElapsed;
+    }
+
+    public override bool Enabled
```
bruno-garcia: we're leaking the .NET API here and it's a pretty bad API. Using a property to set something with a side effect as a timer is not ideal.

Lets call this `Start()` and `Stop()` to be very clear what it does

src/Sentry/Internal/BatchProcessorTimer.cs
```diff
@@ -0,0 +1,61 @@
...
+    {
+        get => _timer.Enabled;
+        set => _timer.Enabled = value;
+    }
+
+    public override event EventHandler<ElapsedEventArgs>? Elapsed;
```
bruno-garcia: Timer uses an event, but we don't have to. Those were API decisions from 20 years ago,

src/Sentry/Internal/BatchProcessorTimer.cs
```diff
@@ -0,0 +1,61 @@
+using System.Timers;
+
+namespace Sentry.Internal;
+
+internal abstract class BatchProcessorTimer : IDisposable
```
bruno-garcia: This whole file for me seems like boilerplate.

It's just wrapping the .NET Timer while giving the same API?
Either make it better, (methods when side effect happens, skip the events, get Sentry diag logging, etc) or delete this altogether

src/Sentry/Protocol/StructuredLog.cs
```diff
@@ -0,0 +1,42 @@
...
+/// </summary>
+/// <remarks>
+/// Sentry Docs: <see href="https://docs.sentry.io/product/explore/logs/"/>.
+/// Sentry Developer Documentation: <see href="https://develop.sentry.dev/sdk/telemetry/logs/"/>.
+/// </remarks>
+internal sealed class StructuredLog : ISentryJsonSerializable
+{
+    private readonly SentryLog[] _items;
+
+    public StructuredLog(SentryLog log)
```
bruno-garcia: can we get rid of this? do we use this constructor? If we anyway allocate the array, just keep this simple and let the caller do that. It's an internal API. 

Less is more

bruno-garcia (commented):
please add a benchmark for this too, we definitely want to get an idea of how this behaves under load. In the tests, even if you're not committing that consider writing something that writes in a tight loop using a few threads (like 1 per CPU core you have) and check how that behaves too.

src/Sentry/Internal/BatchProcessorTimer.cs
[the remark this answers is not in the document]
jamescrosswell: SOLID.

test/Sentry.Tests/Internals/BatchProcessorTests.cs
```diff
@@ -0,0 +1,172 @@
...
+    {
+#if NET9_0_OR_GREATER
+        var e = new ElapsedEventArgs(signalTime);
+#else
+        var type = typeof(ElapsedEventArgs);
+        var ctor = BindingFlags.Instance | BindingFlags.NonPublic | BindingFlags.CreateInstance;
+        var instance = Activator.CreateInstance(type, ctor, null, [signalTime], null);
+        var e = (ElapsedEventArgs)instance!;
+#endif
+        Elapsed.Invoke(this, e);
```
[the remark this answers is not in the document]
Flash0ver: `System.Timers.ElapsedEventArgs` is `sealed`.
I'll chose a different approach for testing.

src/Sentry/Internal/BatchProcessorTimer.cs
Flash0ver: The intention was for Testing.

I simplified the solution by
- remove the `BatchProcessorTimer` abstraction
- instead make `OnIntervalElapsed`, called by the Timer-Callback, `internal` for testing

also
Switched from `System.Timers.Timer` to `System.Threading.Timer`, which has the functionality we need and seems to allocate a bit less.
`System.Threading.Timer` implements the `ITimer` interface, which could be used for abstraction/testing, but it's only available since `net8.0`. So making the `OnIntervalElapsed` method `internal` is the simplest way to enable testing without additional abstractions.

src/Sentry/Internal/BatchProcessorTimer.cs
Flash0ver: Abstraction deleted: see https://github.com/getsentry/sentry-dotnet/pull/4310#discussion_r2177748194

src/Sentry/Internal/BatchProcessorTimer.cs
Flash0ver: Enabled testing without this abstraction, see https://github.com/getsentry/sentry-dotnet/pull/4310#discussion_r2177748194.

But applying similar methods to the updated, abstraction-free implementation for clarity of intent.

src/Sentry/Internal/BatchProcessorTimer.cs
Flash0ver: Abstraction deleted: see https://github.com/getsentry/sentry-dotnet/pull/4310#discussion_r2177748194

src/Sentry/Internal/BatchProcessorTimer.cs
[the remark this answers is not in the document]
Flash0ver: Abstraction deleted: https://github.com/getsentry/sentry-dotnet/pull/4310#discussion_r2177748194

Making sure no exceptions may occur in the updated implementation, where I switched from `System.Timers.Timer` to `System.Threading.Timer`, and try/catch/Diagnostic-Log if there may.

src/Sentry/Internal/BatchProcessorTimer.cs
Flash0ver: Abstraction deleted: https://github.com/getsentry/sentry-dotnet/pull/4310#discussion_r2177748194

Using `System.Threading.Timer` (instead of `System.Timers.Timer`) directly.
For testing, made the callback method of the timer `internal`.

src/Sentry/Internal/BatchProcessorTimer.cs
```diff
@@ -0,0 +1,61 @@
...
+
+    public void Dispose()
+    {
+        Dispose(true);
+        GC.SuppressFinalize(this);
+    }
+
+    protected virtual void Dispose(bool disposing)
+    {
+    }
```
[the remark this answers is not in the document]
Flash0ver: Abstraction deleted: https://github.com/getsentry/sentry-dotnet/pull/4310#discussion_r2177748194

Also, the previous code was adhering to the .NET Dispose Pattern.
See https://learn.microsoft.com/dotnet/standard/garbage-collection/implementing-dispose

src/Sentry/Protocol/StructuredLog.cs
Flash0ver: Yes ... was only used from 2 tests.
And I do like that more as well ... it indicates to the caller it's not quite intended to be created with an individual Log item.

src/Sentry/Internal/BatchProcessor.cs
Flash0ver: I did start out with a lock-free solution for non-flushing logs.
And then realized a flushing log takes 2-3 ms.
But I was not considering the context of _Unity_ where this is unacceptable.
I'm going back to the (mostly) lock-free solution I had initially, and will build up on that to be mostly lock-free with something like a Ring-Buffer of Batch-Buffers.

Thanks for the feedback ... also thanks @bitsandfoxes.

src/Sentry/Internal/BatchProcessor.cs
Flash0ver: Deleted the Abstraction of Timer.
See https://github.com/getsentry/sentry-dotnet/pull/4310#discussion_r2177748194

Now using `System.Threading.Timer` directly, and made the one method that we need for timing-free tests `internal`.

src/Sentry/Internal/BatchBuffer.cs
Flash0ver: `Buffer.BlockCopy` only works with primitive types.
There is `Buffer.MemoryCopy`, but I think this one is also intended for `unmanaged` structs ... not sure about reference types.

src/Sentry/Protocol/StructuredLog.cs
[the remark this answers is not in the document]
Flash0ver: Indeed ... it was only used for testing.
I removed that `.ctor` and let the 2 test methods create the array explicitly.

src/Sentry/Internal/BatchProcessor.cs
Flash0ver: Agreed ... renamed to `StructuredLogBatchProcessor`

I initially intended it to be a general Batch-Processor. But you're right ... a single implementation is premature abstraction.
The implementation has turned to be very Log-specific.

Actually, prefixed most new types with `StructuredLog...`. They are all `internal` and we can extract more general abstractions later, when we can use them.

src/Sentry/Internal/BatchProcessor.cs
Flash0ver: Refactored the code significantly:
The BatchProcessor for Logs is now lock-free, except when a Buffer is being flushed (through either exceeding Capacity or Timeout) with a try-based approach, so if concurrently both the Timeout and the Capacity are exceeded, whichever thread comes first acquires the lock and blocks until all Add-Operations have completed (in case of a Timeout, so that we don't lose any logs) and the other thread trying to Flush will just skip over and returning immediately.

src/Sentry/Internal/BatchProcessor.cs
Flash0ver: Removed.
Reworked the Enqueue-Approach:
We now have two buffers.
If we can't add to the currently active buffer (because it's either full or currently being flushed after a Timeout), we try to add it to the other buffer (after swapping the active buffer, so that upcoming Enqueue-Operations try the new active buffer instead of the previous one being full and/or flushed). Should the second try to enqueue in the other buffer fail as well, we drop the event and record a discarded event for the Client Report.

src/Sentry/Internal/BatchProcessor.cs
Flash0ver: Code reworked ... the minimum capacity now is `2`.

src/Sentry/Internal/BatchProcessor.cs
Flash0ver: I ended up with a solution based on `CountdownEvent`, which does a `Wait()` when flushing the buffer (either when capacity is full or through the Timeout exceeded).
There is no more `lock`. And all non-Capacity-Exceeding Enqueue operations are lock/Wait-free.

src/Sentry/Internal/BatchProcessor.cs
Flash0ver: New strategy:
2 buffers (active and inactive)
- try to add to active buffer
  - when last item added, swap the active buffer
  - when trying to add to a full buffer, try the other buffer
  - when both buffers are full and/or flushed (timeout) drop the log and record a discard-event for Client-Report
- no more locks
- new strategy is based on `CountdownEvent`
  - the only `Wait` is when a Buffer is flushed when full or from Timeout
  - we `Wait` until all Add-Operations in progress are completing, so that we don't lose logs when the timeout is exceeding resulting in an auto-flush from a different thread

src/Sentry/Internal/BatchProcessor.cs
```diff
@@ -0,0 +1,99 @@
...
+    private void IntervalElapsed(object? sender, ElapsedEventArgs e)
+    {
+        _timer.Enabled = false;
+
+        lock (_lock)
+        {
+            if (!_logs.IsEmpty && e.SignalTime > _lastFlush)
+            {
+                Flush();
+            }
```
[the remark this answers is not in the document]
Flash0ver: `lock` removed.
New strategy based on `CountdownEvent`, with the only `Wait` when a Buffer is being flushed until all Add-Operations on that Buffer have completed (required when a Timeout triggers the Flush)

src/Sentry/SentryOptions.cs
```diff
@@ -1897,5 +1897,22 @@ public void SetBeforeSendLog(Func<SentryLog, SentryLog?> beforeSendLog)
...
+
+        /// <summary>
+        /// This API will be removed in the future.
+        /// </summary>
+        /// <remarks>
+        /// Time after which all items in the buffer are sent, regardless of <see cref="InternalBatchSize"/>.
+        /// <para>Must not exceed 30 seconds.</para>
+        /// </remarks>
+        public TimeSpan InternalBatchTimeout { get; set; } = TimeSpan.FromSeconds(5);
```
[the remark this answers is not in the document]
Flash0ver: Deleted both properties / options.

src/Sentry/Internal/DefaultSentryStructuredLogger.cs
```diff
@@ -10,13 +9,33 @@ internal sealed class DefaultSentryStructuredLogger : SentryStructuredLogger
...
+    {
+        return batchInterval.TotalMilliseconds is <= 0 or > int.MaxValue
+            ? TimeSpan.FromMilliseconds(int.MaxValue)
+            : batchInterval;
     }
```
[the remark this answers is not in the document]
Flash0ver: Method deleted.

src/Sentry/Internal/DefaultSentryStructuredLogger.cs
```diff
@@ -10,13 +9,33 @@ internal sealed class DefaultSentryStructuredLogger : SentryStructuredLogger
...
+        _batchProcessor = new BatchProcessor(hub, ClampBatchCount(options.Experimental.InternalBatchSize), ClampBatchInterval(options.Experimental.InternalBatchTimeout));
+    }
+
+    private static int ClampBatchCount(int batchCount)
+    {
+        return batchCount <= 0
+            ? 1
+            : batchCount > 1_000_000
+                ? 1_000_000
+                : batchCount;
```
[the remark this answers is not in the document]
Flash0ver: Method deleted.

src/Sentry/Internal/BatchProcessor.cs
```diff
@@ -0,0 +1,99 @@
...
+        }
```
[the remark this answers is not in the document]
Flash0ver: Reworked the code.
Also, the new `Flush` cannot throw. No user code is invoked and tests, as well as `Debug.Assert`s are guarding corner cases.

src/Sentry/Internal/BatchProcessor.cs
```diff
@@ -0,0 +1,99 @@
...
+        {
+            EnqueueCore(log);
+        }
+    }
+
+    private void EnqueueCore(SentryLog log)
+    {
+        var empty = _logs.IsEmpty;
+        var added = _logs.TryAdd(log);
+        Debug.Assert(added, $"Since we currently have no lock-free scenario, it's unexpected to exceed the {nameof(BatchBuffer<SentryLog>)}'s capacity.");
```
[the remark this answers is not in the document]
Flash0ver: Code reworked.
Also, the `Debug.Assert` is guarding an invariant / assumption at the time of writing the logic.
Should the strategy later change, the `Debug.Assert` should indicate to the developer that a previous assumption is no longer valid, and appropriate conditional logic must be considered.

src/Sentry/Internal/BatchBuffer.cs
```diff
@@ -0,0 +1,80 @@
...
+        var array = new T[_count];
+        Array.Copy(_array, array, _count);
+        return array;
+    }
+
+    internal void Clear()
+    {
+        if (_count == 0)
+        {
+            return;
```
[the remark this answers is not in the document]
Flash0ver: The branch not only clears the array, but also resets the `_count` to `0`.
The logic was inspired by `System.Collections.Generic.List<T>`.

src/Sentry/SentryOptions.cs
Flash0ver: Removed the property.
Introduced a `.ctor` overload for testing.
When created via the `Hub`, setting BatchSize to `100` (and a Timeout of `5 seconds`).
This value proved solid during performance testing.

src/Sentry/Internal/DefaultSentryStructuredLogger.cs
Flash0ver: Method removed ... no longer required because of the option being removed as well.

src/Sentry/Internal/DefaultSentryStructuredLogger.cs
Flash0ver: Method removed ... no longer required because of the option being removed as well.

src/Sentry/Threading/ScopedCountdownLock.cs
```diff
@@ -0,0 +1,149 @@
+namespace Sentry.Threading;
+
+/// <summary>
+/// A synchronization primitive that tracks the amount of <see cref="CounterScope"/>s held,
+/// and is signaled when the count reaches zero while a <see cref="LockScope"/> is held.
```
jamescrosswell, line 5: > is signaled when the count reaches zero while a `LockScope` is held

Is this accurate? Looking at the code it signals during `ExitCounterScope`:
https://github.com/getsentry/sentry-dotnet/blob/e8850db578a5f3a4ba9afb940fa934e365e89020/src/Sentry/Threading/ScopedCountdownLock.cs#L59

... which is only called when the lockObj is not null:
https://github.com/getsentry/sentry-dotnet/blob/e8850db578a5f3a4ba9afb940fa934e365e89020/src/Sentry/Threading/ScopedCountdownLock.cs#L110-L114

And that only happens if the CountdownEvent is **not zero**:
https://github.com/getsentry/sentry-dotnet/blob/e8850db578a5f3a4ba9afb940fa934e365e89020/src/Sentry/Threading/ScopedCountdownLock.cs#L49-L52

So it seems it's signalled when releasing a lock scope and when the count has not yet reached zero.

src/Sentry/Internal/StructuredLogBatchProcessor.cs
```diff
@@ -0,0 +1,136 @@
...
+
+    private bool TryEnqueue(StructuredLogBatchBuffer buffer, SentryLog log)
+    {
+        var status = buffer.Add(log);
+
+        if (status is BatchBufferAddStatus.AddedLast)
+        {
+            TrySwapActiveBuffer(buffer);
+            CaptureLogs(buffer);
+            return true;
```
jamescrosswell, line 91: Should this return true if the call to `TrySwapActiveBuffer` on line 89 returns false?

src/Sentry/Internal/StructuredLogBatchProcessor.cs
```diff
@@ -0,0 +1,136 @@
...
+        }
+
+        return status is BatchBufferAddStatus.AddedFirst or BatchBufferAddStatus.Added;
+    }
+
+    private bool TrySwapActiveBuffer(StructuredLogBatchBuffer currentActiveBuffer)
```
jamescrosswell: It doesn't look like the return value is every used. Maybe we remove the return value and rename it to `SwapActiveBuffer`.

src/Sentry/Internal/StructuredLogBatchProcessor.cs
```diff
@@ -0,0 +1,136 @@
...
+    private void CaptureLogs(StructuredLogBatchBuffer buffer)
+    {
+        SentryLog[]? logs = null;
+
+        using (var scope = buffer.TryEnterFlushScope())
+        {
+            if (scope.IsEntered)
+            {
+                logs = scope.Flush();
+            }
```
jamescrosswell, line 122: What happens if `!scope.IsEntered`? Does that need to be logged?

src/Sentry/Internal/StructuredLogBatchBuffer.cs
```diff
@@ -0,0 +1,304 @@
...
+        }
+
+        _additions = 0;
+        Array.Clear(_array, 0, length);
+    }
+
+    private void EnableTimer()
+    {
+        var updated = _timer.Change(_timeout, Timeout.InfiniteTimeSpan);
+        Debug.Assert(updated, "Timer was not successfully enabled.");
```
jamescrosswell: Should we be logging this using our `IDiagnosticLogger` instead?

It looks like the most likely scenario for this happening would be if this method was called when _timer had already been disposed.

src/Sentry/Threading/ScopedCountdownLock.cs
```diff
@@ -0,0 +1,149 @@
...
+    /// When successful, the lock <see cref="IsEngaged"/>, <see cref="Count"/> can reach <see langword="0"/> and the event be set/signaled.
+    /// Check via <see cref="LockScope.IsEntered"/> whether the Lock <see cref="IsEngaged"/>.
+    /// Use <see cref="LockScope.Wait"/> to block until every active <see cref="CounterScope"/> has exited before performing the locked operation.
+    /// After the locked operation has completed, disengage the Lock via <see cref="LockScope.Dispose"/>.
+    /// </summary>
+    internal LockScope TryEnterLockScope()
+    {
+        if (Interlocked.CompareExchange(ref _isEngaged, 1, 0) == 0)
+        {
+            _ = _event.Signal();
```
jamescrosswell: If I've understood correctly, this is decrementing the CountdownEvent towards zero...

The only codepath that calls `TryEnterLockScope` is triggered from here:
https://github.com/getsentry/sentry-dotnet/blob/79ce02e6988d8b13773d39dc7e2affb862f00318/src/Sentry/Internal/StructuredLogBatchProcessor.cs#L108-L114

So flushing the buffer just before capturing all the logs... it'd happen either due to reaching the timout or the buffer limit. 

Why do we signal the `CountdownEvent` here then? 

Is this because we're essentially using `_event` here to count the number of Add operations that are pending on the buffer and this is "off by one" because it's got an initial value of 1?
https://github.com/getsentry/sentry-dotnet/blob/e8850db578a5f3a4ba9afb940fa934e365e89020/src/Sentry/Threading/ScopedCountdownLock.cs#L20

And otherwise this wait would block indefinitely?
https://github.com/getsentry/sentry-dotnet/blob/f28cc6d7750e2b5fea81eaf78c1d0a23033ed2c1/src/Sentry/Internal/StructuredLogBatchBuffer.cs#L274

If so, it took me a **_long_** time to figure this out. Maybe worth adding a small comment here where we signal (explaining that the initial count is 1 to prevent this from firing until the LockScope is entered).

src/Sentry/Internal/StructuredLogBatchBuffer.cs
```diff
@@ -0,0 +1,304 @@
...
+    private int _additions;
+    private readonly ScopedCountdownLock _addLock;
+
+    private readonly ISystemClock _clock;
+    private readonly Timer _timer;
+    private readonly TimeSpan _timeout;
+
+    private readonly Action<StructuredLogBatchBuffer, DateTimeOffset> _timeoutExceededAction;
+
+    private DateTimeOffset _lastFlush = DateTimeOffset.MinValue;
```
jamescrosswell: Doesn't look like this variable is ever used.

src/Sentry/Internal/StructuredLogBatchBuffer.cs
```diff
@@ -0,0 +1,304 @@
...
+        private StructuredLogBatchBuffer? _lockObj;
+        private ScopedCountdownLock.LockScope _scope;
+
+        internal FlushScope(StructuredLogBatchBuffer lockObj, ScopedCountdownLock.LockScope scope)
+        {
+            _lockObj = lockObj;
+            _scope = scope;
+        }
+
+        internal bool IsEntered => _scope.IsEntered;
```
jamescrosswell, line 271: It seems like implicitly if there is a `_lockObj` then there should be a `_scope` and `_scope.IsEntered` should be `true`. Possibly this could be an assertion in the constructor above then (make sure `_scope.IsEntered == true`)? Since otherwise the _scope.Wait() here might wait indefinitely:
https://github.com/getsentry/sentry-dotnet/blob/f28cc6d7750e2b5fea81eaf78c1d0a23033ed2c1/src/Sentry/Internal/StructuredLogBatchBuffer.cs#L274

src/Sentry/Internal/StructuredLogBatchBuffer.cs
```diff
@@ -0,0 +1,304 @@
...
+/// </remarks>
+[DebuggerDisplay("Name = {Name}, Capacity = {Capacity}, Additions = {_additions}, AddCount = {AddCount}")]
+internal sealed class StructuredLogBatchBuffer : IDisposable
+{
+    private readonly SentryLog[] _array;
+    private int _additions;
+    private readonly ScopedCountdownLock _addLock;
+
+    private readonly ISystemClock _clock;
+    private readonly Timer _timer;
```
jamescrosswell, line 19: The `StructuredLogBatchProcessor` is already responsible for [handling the timeout](https://github.com/getsentry/sentry-dotnet/blob/79ce02e6988d8b13773d39dc7e2affb862f00318/src/Sentry/Internal/StructuredLogBatchProcessor.cs#L122-L129).

Hypothetically, the `_clock`, `_timer, and `_timeout` could all be moved to the `StructuredLogBatchProcessor`... which would then simplify this class. 

That could possibly make this class more reusable/lightweight if we needed a batch buffer but had no need to flush after a certain timeout. I'm not sure how likely that scenario is though so maybe we just leave as is for now. If that scenario ever pops up, it would be an easy refactor.

src/Sentry/SentryStructuredLogger.cs
```diff
@@ -100,4 +111,27 @@ public void LogFatal(string template, object[]? parameters = null, Action<Sentry
     {
         CaptureLog(SentryLogLevel.Fatal, template, parameters, configureLog);
     }
+
+    /// <summary>
+    /// When overridden in a derived <see langword="class"/>,
+    /// clears all buffers for this logger and causes any buffered logs to be sent by the underlying <see cref="ISentryClient"/>.
+    /// </summary>
+    protected internal virtual void Flush()
```
jamescrosswell: Any reason not to make this abstract?

Flash0ver: My initial thought process was: not every implementation actually supports flushing, thinking about the existing `DisabledSentryStructuredLogger` and `InMemorySentryStructuredLogger`. Only the `DefaultSentryStructuredLogger` does.

But ... on 2nd thought ... actually, this method is similar to the abstract `CaptureLog` method:
It's up to the implementation to explicitly decide what to do, since the `Flush` will be very depending on the `CaptureLog` implementation, so they should be treated equivalently.

Now I do prefer it to be `abstract`, too!
Changed to be `abstract` instead.

src/Sentry/Internal/StructuredLogBatchBuffer.cs
Flash0ver: At some point, I had the `StructuredLogBatchProcessor` having a single `Timer` that was used for both alternating/swapping Buffers.

However, there was a race condition in the swapping process,
where a thread that filled the buffer would disable the timer,
but after another thread that "overflowed" to the other buffer enabling the timer when adding the first element to it.

I then solved this race condition by basically checking with every add operation whether the timer is running, and starting it otherwise. I wanted to avoid this "ensure timer is enabled" by making it "thread-safe" where every Buffer has its own independent timer, which now only has to be enabled when the first item is added, and disabled if the buffer got full before the timeout has exceeded. No need to disable the buffer when the timeout has triggered, because there is only a "dueTime" no repetitive "period".

I also had a solution where I had a "base" `BatchBuffer`, that only had the Lock-Free Add,
and a derived `TimeoutBatchBuffer`, which added the Timer.
But I then had the thought of "avoid hasty abstractions" ... so I merged them into the `StructuredLogBatchBuffer`, because that is what its currently optimized for.

Likewise, I believe too that it shouldn't be too much work to split these two concerns later on again, should we find a usage of the "pure" timer-less thread-safe `BatchBuffer`.

Flash0ver: If you prefer, we could split it up already, into a
- lock-free and thread-safe `BatchBuffer`
  - either inheritable or `sealed`
- a derived or wrapping `TimeoutBatchBuffer` or `StructuredLogBatchBuffer`

Either in this PR ... or a follow-up.

src/Sentry/Internal/StructuredLogBatchBuffer.cs
Flash0ver: Correct.
Great observation!
Adding a `Debug.Assert` to the _.ctor_.

src/Sentry/Internal/StructuredLogBatchProcessor.cs
Flash0ver: made `void` and renamed: ~Try~SwapActiveBuffer

src/Sentry/Internal/StructuredLogBatchProcessor.cs
Flash0ver: In that case, `CaptureLogs` have been called during a race condition from both
- the thread that filled the buffer, disabled the timer and does an immediate Flush
- the Timer-Thread just before it has been disabled

In this case, first thread wins, and the other no-ops and returns immediately.

Not sure if logging this scenario would be actionable / informative.

src/Sentry/Internal/StructuredLogBatchBuffer.cs
Flash0ver: Oh .. I thought `Timer.Change` would return `false` when updated concurrently.
Oh .. it only returns false when already disposed.

This actually may occur.
Not sure if this is actionable though.
Perhaps we may just discard the return value 🤔?

Actually this now makes me think about "fixing" this scenario, when a Log message is written when the Hub and therefore the Logger are being disposed:

We could additionally add a `Close` method on the `SentryStructuredLogger`, that is called when the `Hub` is disposed, and marking the Logger to no longer allow any more Logs to be written,
after which we `Flush` both Buffers,
after which we `Dispose` the Logger and its Batch-Processor and both Buffers including the Timer,
rendering this scenario unexpected so that it indeed becomes an invariant.

What do you think?

src/Sentry/Internal/StructuredLogBatchBuffer.cs
Flash0ver: Right ... that's a relic / oversight when the Batch-Processor had a single Timer for both Batch-Buffers.

For thread-safety reasons, I moved the Timer into each Batch-Buffer
and guarding against a Double-Flush (explicitly when Buffer exceeded capacity and implicitly when Timeout got triggered) with a scope that can only be entered by a single thread,
so that the race-conditional other Flush will just no-op and return immediately.

Removed ... and also the `ISystemClock` as its no longer needed either.

bruno-garcia:
cursor review

src/Sentry/Threading/ScopedCountdownLock.cs
Flash0ver: Exactly!

I thought I documented the behavior and intent sufficiently.
Thanks for going through all of that!

I realized in the comments I only mention `"when the count reaches 0"`,
and after a LockScope was entered successfully `"the count may reach zero"`.
I haven't explicitly mentioned the "offset" with the initial count of `1`, which is only signaled when the LockScope has been successfully entered.

I will explicitly mention this in the structured comments,
and add a single-line-comment to both the "offset signal" and the "offset reset".

bruno-garcia:
cursor review

src/Sentry/Internal/StructuredLogBatchBuffer.cs
jamescrosswell: I guess if a log gets written when the Hub is being disposed, we want to discard the log and shutdown gracefully, with no errors/exceptions. 

If that's what happens already, no need to change I don't think. If that's not currently guaranteed then we should address it. I don't think this code in particular will create any errors during Hub shutdown.

jamescrosswell (approved):

src/Sentry/Internal/StructuredLogBatchBuffer.cs
Flash0ver: Adding test cases to ensure the behavior of:
When (or after) the Hub is being (or was already) disposed, discard new logs being written.

src/Sentry/Internal/StructuredLogBatchBuffer.cs
```diff
@@ -0,0 +1,297 @@
...
+    private void EnableTimer()
+    {
+        var updated = _timer.Change(_timeout, Timeout.InfiniteTimeSpan);
+        Debug.Assert(updated, "Timer was not successfully enabled.");
+    }
+
+    private void DisableTimer()
+    {
+        var updated = _timer.Change(Timeout.Infinite, Timeout.Infinite);
+        Debug.Assert(updated, "Timer was not successfully disabled.");
```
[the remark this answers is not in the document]
Flash0ver: Removed the `Debug.Assert` statements because they are no longer relevant:

The `StructuredLogBatchBuffer` now has a `_disposed` field. And `Add(SentryLog)` returns immediately if the instance has been disposed. So the timer can no longer be enabled/disabled after `Dispose`.

Also, added equivalent unit tests for all involved types to assert correct / intended behavior when the `Hub` is being disposed.

PR #4314, opened 2025-06-29
fix: Crontab validation when capturing checkins

### LINKED ISSUE #3719: .NET crontab syntax for every 2nd hour doesn't work
### Package

Sentry

### .NET Flavor

.NET

### .NET Version

8

### OS

Windows

### SDK Version

4.12.1

24.10.0

### Steps to Reproduce

When using the API to set interval we aren't allowed to use the syntax for every 2nd hour (`0 */2 * * *`). Alltough when we are entering this in the web-UI it seems to be allowed (and the tooltip displays the correct interpretation).
Image

### Expected Result

To be allowed to use the SetInterval method with a slash `/` sign to be able to configure the cron programmatically .
E.g. 
```
        private void ConfigureOptions(SentryMonitorOptions options)
        {
            options.Interval("0 */2 * * *"); // <-- this throws an error
            options.CheckInMargin = TimeSpan.FromMinutes(10);
            options.FailureIssueThreshold = 1;
        }
```

### Actual Result

Exception:
`System.ArgumentException: 'The provided crontab does not match the expected format of '* * * * *' translating to 'minute', 'hour', 'day of the month', 'month', and 'day of the week'.'`

I suspect that the regex needs some tweaking here.

jamescrosswell:
Resolves #3719:
- https://github.com/getsentry/sentry-dotnet/issues/3719

### DIFF
--- a/test/Sentry.Tests/SentryMonitorOptionsTests.cs
+++ b/test/Sentry.Tests/SentryMonitorOptionsTests.cs
@@ -1,0 +1,112 @@
+namespace Sentry.Tests;
+
+public class SentryMonitorOptionsTests
+{
+    /*
+        The time and date fields are:
+              field          allowed values
+              -----          --------------
+              minute         0-59
+              hour           0-23
+              day of month   1-31
+              month          1-12 (or names, see below)
+              day of week    0-7 (0 or 7 is Sunday, or use names)
+
+        Special characters are:
+            *	any value
+            ,	value list separator
+            -	range of values
+            /	step values
+    */
+    [Theory]
+    [InlineData("* * * * *")]
+    [InlineData("0 0 1 1 *")]
+    [InlineData("0 0 1 * 0")]
+    [InlineData("59 23 31 12 7")]
+    [InlineData("0 */2 * * *")]
+    [InlineData("0 8-10 * * *")]
+    [InlineData("0 6,8,9 * * *")]
+    // Step values (*/n)
+    [InlineData("*/15 * * * *")]        // Every 15 minutes
+    [InlineData("0 */6 * * *")]         // Every 6 hours
+    [InlineData("0 0 */2 * *")]         // Every 2 days
+    [InlineData("0 0 1 */3 *")]         // Every 3 months
+    // Complex ranges
+    [InlineData("1-15 * * * *")]        // Minutes 1 through 15
+    [InlineData("* 9-17 * * *")]        // Business hours
+    [InlineData("* * 1-15,16-31 * *")]  // Split day ranges
+    // Multiple comma-separated values
+    [InlineData("1,15,30,45 * * * *")]  // Specific minutes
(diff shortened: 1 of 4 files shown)


test/Sentry.Tests/SentryMonitorOptionsTests.cs
```diff
@@ -0,0 +1,112 @@
...
+    [InlineData("* * * * 8")]
+    // Invalid step values
+    [InlineData("*/0 * * * *")]         // Step value cannot be 0
+    [InlineData("*/100 * * * *")]       // Step value too large for minutes
+    [InlineData("* */25 * * *")]        // Step value too large for hours
+    [InlineData("* * */32 * *")]        // Step value too large for days
+    [InlineData("* * * */13 *")]        // Step value too large for months
+    [InlineData("* * * * */8")]         // Step value too large for weekdays
+    [InlineData("*/60 * * * *")]        // Step value equals max value for minutes
+    [InlineData("* */24 * * *")]        // Step value equals max value for hours
```
Flash0ver: question: compatibility / format

Are these values rejected by Sentry?
https://crontab.guru/ says that the strings of `// Invalid step values` (except the first one) are actually valid.
I am using cron tables super rarely ... but are there different subsets / "strictness-levels" of valid cron table formats?

test/Sentry.Tests/SentryMonitorOptionsTests.cs
```diff
@@ -0,0 +1,112 @@
...
+            *	any value
+            ,	value list separator
+            -	range of values
+            /	step values
+    */
+    [Theory]
+    [InlineData("* * * * *")]
+    [InlineData("0 0 1 1 *")]
+    [InlineData("0 0 1 * 0")]
+    [InlineData("59 23 31 12 7")]
```
Flash0ver, line 25: question: 0,7 = Sunday

Just for clarity:
Does Sentry support `day=7` to be equivalent to `day=0`?

Flash0ver:
~~question: merge target~~

~~Will this be merged to `main` after #4313 has been merged to `main`?~~
~~Or is this intended to be merged to #4313 and then both combined merged to `main`?~~

test/Sentry.Tests/SentryMonitorOptionsTests.cs
jamescrosswell: It seems to in the UI yes:

image

test/Sentry.Tests/SentryMonitorOptionsTests.cs
jamescrosswell: Yeah you're right... Sentry seems quite flexible regarding step values:
image

So it seems the step values can be anything (as long as they're greater than 0).

The ranges and list values have to be valid for the field type though.

I'll update the tests and the regex accordingly.

Flash0ver (approved):

PR #4315, opened 2025-06-30
fix: Sentry capturing compressed bodies when RequestDecompression middleware is enabled

### LINKED ISSUE #4312: Request.Body is not decompressed when Content-Encoding: gzip, even with ASP.NET Core’s UseRequestDecompression()
### Package

Sentry

### .NET Flavor

.NET Core

### .NET Version

9.0.6

### OS

Windows

### OS Version

11

### Development Environment

Visual Studio v17.x

### SDK Version

3.23.1

_No response_

### Workload Versions

--------------------------------------------------------------------
aspire                     8.2.2/8.0.100         VS 17.14.36212.18

When using Sentry SDK in an ASP.NET Core app with UseRequestDecompression(), HTTP requests with Content-Encoding: gzip are still logged with their raw compressed body in Sentry.

This occurs because UseRequestDecompression() only integrates with model binding and does not replace or decompress HttpContext.Request.Body, which Sentry reads directly.

### Steps to Reproduce

Configure ASP.NET Core pipeline with UseRequestDecompression() early.

Enable Sentry automatic request body logging.

Send a POST request with Content-Encoding: gzip and gzipped JSON body.

Trigger an exception.

Check Sentry issue – Request.Body is unreadable gzip binary.

### Expected Result

Sentry should log the decompressed request body when UseRequestDecompression() is enabled and the request has Content-Encoding: gzip.

### Actual Result

Image

jamescrosswell:
Resolves #4312:
- https://github.com/getsentry/sentry-dotnet/issues/4312

## Summary

Due to the [SentryStartupFilter](https://github.com/getsentry/sentry-dotnet/blob/263961124a49c5fa8290c8721294e7fd472e7027/src/Sentry.AspNetCore/SentryStartupFilter.cs#L9-L20), Sentry's middleware is the first middleware in the chain, so was being called before the `RequestDecompressionMiddleware`. Because of this, the `HttpContext.Request.Body` that Sentry was storing on the Scope (and therefore events) was not being uncompressed as you would expect when enabling this middleware.

## Solution

This PR alters the `SentryStartupFilter` so that it checks whether `IRequestDecompressionProvider` has been registered in the services collection (see [RequestDecompressionServiceExtensions](https://github.com/dotnet/aspnetcore/blob/9e69f19a5c7efa19577627a62c029ff387810d50/src/Middleware/RequestDecompression/src/RequestDecompressionServiceExtensions.cs#L19-L25)). If it is then the `RequestDecompressionMiddleware` is registered before Sentry's middleware. 

### Potential issues

Errors decompressing the request body in the `RequestDecompressionMiddleware` itself will not be captured by Sentry.

### TODO

- [ ] We could a wrapper around the app builder to prevent re-registrations of the `RequestDecompressionMiddleware` once we've already registered it (similar to [what we do for tracing](https://github.com/getsentry/sentry-dotnet/blob/bd62cf0421cc7a5d7b7804f8461fae00d47d43b8/src/Sentry.AspNetCore/SentryTracingStartupFilter.cs#L8-L15)).
- [ ] We could also add an option like `AutoRegisterRequestDecompressionMiddleware` to `SentryAspNetCoreOptions` to make it possible to disable this logic.

### DIFF
--- a/src/Sentry.AspNetCore/SentryStartupFilter.cs
+++ b/src/Sentry.AspNetCore/SentryStartupFilter.cs
@@ -1,5 +1,9 @@
 using Microsoft.AspNetCore.Builder;
 using Microsoft.AspNetCore.Hosting;
+using Microsoft.AspNetCore.RequestDecompression;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Options;
+using Sentry.Extensibility;
 
 namespace Sentry.AspNetCore;
 
@@ -11,10 +15,20 @@ public class SentryStartupFilter : IStartupFilter
     /// <summary>
     /// Adds Sentry to the pipeline.
     /// </summary>
-    public Action<IApplicationBuilder> Configure(Action<IApplicationBuilder> next) => e =>
+    public Action<IApplicationBuilder> Configure(Action<IApplicationBuilder> next) => app =>
     {
-        e.UseSentry();
+        app.UseSentry();
 
-        next(e);
+        // If we are capturing request bodies and the user has configured request body decompression, we need to
+        // ensure that the RequestDecompression middleware gets called before Sentry's middleware. The last middleware
+        // added is the first one to be executed.
+        var options = app.ApplicationServices.GetService<IOptions<SentryAspNetCoreOptions>>();
+        if (options?.Value is { } o&& o.MaxRequestBodySize != RequestSize.None
+            && app.ApplicationServices.GetService<IRequestDecompressionProvider>() is not null)
+        {
+            app.UseRequestDecompression();
+        }
+
+        next(app);
     };
 }
--- a/src/Sentry/Extensibility/BaseRequestPayloadExtractor.cs
+++ b/src/Sentry/Extensibility/BaseRequestPayloadExtractor.cs
@@ -18,24 +18,29 @@ public abstract class BaseRequestPayloadExtractor : IRequestPayloadExtractor
             return null;
         }
 
-        if (request.Body == null
-            || !request.Body.CanSeek
-            || !request.Body.CanRead
-            || !IsSupported(request))
+        if (request.Body is not { CanRead: true } || !IsSupported(request))
         {
             return null;
         }
 
-        var originalPosition = request.Body.Position;
+        // When RequestDecompression is enabled, the RequestDecompressionMiddleware will store a SizeLimitedStream
+        // in the request body after decompression. Seek operations throw an exception, but we can still read the stream
+        var originalPosition = request.Body.CanSeek ? request.Body.Position : 0;
         try
         {
-            request.Body.Position = 0;
+            if (request.Body.CanSeek)
+            {
+                request.Body.Position = 0;
+            }
 
             return DoExtractPayLoad(request);
         }
         finally
         {
-            request.Body.Position = originalPosition;
+            if (request.Body.CanSeek)
+            {
+                request.Body.Position = originalPosition;
+            }
         }
     }
 
--- a/samples/Directory.Build.targets
+++ b/samples/Directory.Build.targets
@@ -10,7 +10,7 @@
 
   <Target Name="GenerateSharedDsnConstant"
           BeforeTargets="BeforeCompile"
-          Condition=" '$(SENTRY_DSN)' != '' and '$(PlatformIsMobile)' == 'true'">
+          Condition="'$(SENTRY_DSN)' != ''">
 
     <Message Text="Generating shared EnvironmentVariables.g.cs with embedded DSN..." Importance="High" />
 
@@ -25,7 +25,7 @@ namespace Sentry.Samples%3B
 internal static class EnvironmentVariables
 {
     /// &lt;summary&gt;
-    /// To make things easier for the SDK maintainers we have a custom build target that writes the
+    /// To make things easier for the SDK maintainers, we have a custom build target that writes the
     /// SENTRY_DSN environment variable into an EnvironmentVariables class that is available for mobile
     /// targets. This allows us to share one DSN defined in the ENV across desktop and mobile samples.
     /// Generally, you won't want to do this in your own mobile projects though - you should set the DSN


src/Sentry.AspNetCore/RequestDecompression/RequestDecompressionMiddleware.cs
```diff
@@ -0,0 +1,107 @@
+// Adapted from: https://github.com/dotnet/aspnetcore/blob/c18e93a9a2e2949e1a9c880da16abf0837aa978f/src/Middleware/RequestDecompression/src/RequestDecompressionMiddleware.cs
+
+// // Licensed to the .NET Foundation under one or more agreements.
```
Flash0ver, line 3: nitpick: double `//`

src/Sentry.AspNetCore/RequestDecompression/RequestDecompressionMiddleware.cs
```diff
@@ -0,0 +1,107 @@
...
+            await decompressionStream.DisposeAsync().ConfigureAwait(false);
+        }
+    }
+
+    private void HandleException(Exception e)
+    {
+        const string description =
+            "An exception was captured and then re-thrown, when attempting to decompress the request body." +
+            "The web server likely returned a 5xx error code as a result of this exception.";
+        e.SetSentryMechanism("RequestDecompressionMiddleware", description, handled: false);
```
Flash0ver: nitpick: the type-string could be `nameof(RequestDecompressionMiddleware)`

src/Sentry.AspNetCore/SentryAspNetCoreOptions.cs
```diff
@@ -122,15 +122,15 @@ internal void SetEnvironment(IWebHostEnvironment hostingEnvironment)
 
             if (hostingEnvironment.IsProduction())
             {
-                Environment = Internal.Constants.ProductionEnvironmentSetting;
+                Environment = Sentry.Internal.Constants.ProductionEnvironmentSetting;
```
Flash0ver: question: I wonder where the ambiguity came from in this changeset 🤔.

test/Sentry.AspNetCore.Tests/RequestDecompressionMiddleware/RequestDecompressionMiddlewareTests.cs
```diff
@@ -0,0 +1,167 @@
...
+                        {
+                            using var reader = new StreamReader(context.Request.Body, Encoding.UTF8);
+                            var body = await reader.ReadToEndAsync();
+                            await context.Response.WriteAsync(body);
+                        });
+                    });
+                });
+        }
+
+        public void FakeDecompressionError()
```
Flash0ver: nitpick: Would you mind prefixing this method with `With` ... since I expected `FakeDecompressionError` to throw already.

test/Sentry.AspNetCore.Tests/RequestDecompressionMiddleware/RequestDecompressionMiddlewareTests.cs
```diff
@@ -0,0 +1,167 @@
...
+        }
+
+        public void Dispose()
+        {
+            _client.Dispose();
+            _server.Dispose();
+        }
+    }
+
+    private readonly Fixture _fixture = new();
```
Flash0ver, line 91: question: Should we `Dispose` of the `Fixture` in this Test-Class?

Since the `Fixture` is `IDisposable`, and we create a new instance per Test-Case,
shall we make this Test-Class also `IDisposable` and call `_fixture.Dispose` in it's `Dispose`?

src/Sentry/Extensibility/BaseRequestPayloadExtractor.cs
```diff
@@ -18,24 +18,29 @@ public abstract class BaseRequestPayloadExtractor : IRequestPayloadExtractor
...
-            || !IsSupported(request))
+        if (request.Body is not { CanRead: true } || !IsSupported(request))
         {
             return null;
         }
 
-        var originalPosition = request.Body.Position;
+        // When RequestDecompression is enabled, the RequestDecompressionMiddleware will store a SizeLimitedStream
+        // in the request body after decompression. Seek operations throw an exception, but we can still read the stream
+        var originalPosition = request.Body.CanSeek ? request.Body.Position : 0;
```
Flash0ver: question: Maybe split into 2 paths?

I wonder if it's cleaner to read when split into two paths.
```CSharp
if (!request.Body.CanSeek)
{
  return DoExtractPayLoad(request);
}

var originalPosition = request.Body.Position;
try
{
  request.Body.Position = 0;
  return DoExtractPayLoad(request);
}
finally
{
  request.Body.Position = originalPosition;
}
```

I guess it's a matter of taste ... however you prefer.

src/Sentry.AspNetCore/SentryStartupFilter.cs
```diff
@@ -11,10 +16,19 @@ public class SentryStartupFilter : IStartupFilter
...
+        if (options?.Value is { } o && o.MaxRequestBodySize != RequestSize.None
+            && app.ApplicationServices.GetService<IRequestDecompressionProvider>() is not null)
+        {
+            app.UseMiddleware<RequestDecompressionMiddleware>();
+        }
 
-        next(e);
+        app.UseSentry();
+
+        next(app);
```
Flash0ver, line 32: question: Double-decompression?

If the consuming App-Code has both `builder.Services.AddRequestDecompression()` and also `app.UseRequestDecompression()`, would the Stream now be decompressed twice in the pipline (via both Sentry's and AspNetCore's `RequestDecompressionMiddleware`).

Or is this prevented by
https://github.com/dotnet/aspnetcore/blob/8883b98afedee35e8360db3f24ce8cd2e17217ad/src/Middleware/RequestDecompression/src/DefaultRequestDecompressionProvider.cs#L52?

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@
 
 ### Fixes
 
+- When Sentry is configured to capture Request bodies in ASP.NET Core, the uncompressed content is not captured when RequestDecompression middleware is enabled. Previously the request bodies were being captured prior to decompression ([#4315](https://github.com/getsentry/sentry-dotnet/pull/4315))
```
Flash0ver: thought: I was a bit confused when reading the CHANGELOG entry.

After my first read, I thought that no Content is captured anymore when `RequestDecompressionMiddleware` is present.
But actually the Content is now (tried to be) decompressed before capturing. Yet still capturing an Exception when decompression fails (without the Body, though).

I'm wondering if this could be phrased a bit differently 🤔.
I couldn't really come up with a clearer description that is not too verbose, though.

src/Sentry.AspNetCore/SentryAspNetCoreOptions.cs
jamescrosswell: Ah, it no longer exists actually... I'll revert this.

test/Sentry.AspNetCore.Tests/RequestDecompressionMiddleware/RequestDecompressionMiddlewareTests.cs
jamescrosswell: It looks like that will work yes 👍🏻 
https://xunit.net/docs/shared-context#constructor

src/Sentry/Extensibility/BaseRequestPayloadExtractor.cs
jamescrosswell: Yeah that's heaps more readable.

src/Sentry.AspNetCore/SentryStartupFilter.cs
jamescrosswell: If the body is uncompressed (which it will be after we run our custom ) then `_provider.GetDecompressionStream` returns null (see [docs](https://learn.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.requestdecompression.irequestdecompressionprovider.getdecompressionstream?view=aspnetcore-9.0#returns)). 

So the second middleware [does nothing](https://github.com/dotnet/aspnetcore/blob/9e69f19a5c7efa19577627a62c029ff387810d50/src/Middleware/RequestDecompression/src/RequestDecompressionMiddleware.cs#L47-L51) basically.

Flash0ver (approved):

PR #4318, opened 2025-07-01
fix: EventProcessors not running for native crashes on iOS

### LINKED ISSUE #3621: dotnet code setting User.id not showing up in iOS events
### Package

Sentry

### .NET Flavor

.NET

### .NET Version

8.0.6 (and some on versions from 8.0.0 to 8.0.7, error still seen in all)

### OS

iOS

### SDK Version

4.4

_No response_

### Steps to Reproduce

Image attached describes code used to set User.Id and User.Username. This provides values for Android devices, but not for iOS devices. This code block comes directly after Sentry .init block. Localhost testing and logging show id does have value just before code block for iOS device tested.

Image

### Expected Result

Expected to see User.Id and User.Username values for both Android and iOS.

### Actual Result

No values for iOS devices. Customer believes this did work during initial testing, but no values seen in the last 90 days given code used (shown above).

jamescrosswell:
Resolves #3621:
- https://github.com/getsentry/sentry-dotnet/issues/3621

### DIFF
--- a/src/Sentry/Internal/SentryEventHelper.cs
+++ b/src/Sentry/Internal/SentryEventHelper.cs
@@ -2,9 +2,9 @@ using Sentry.Extensibility;
 
 namespace Sentry.Internal;
 
-internal class SentryEventHelper(SentryOptions options)
+internal static class SentryEventHelper
 {
-    public SentryEvent? ProcessEvent(SentryEvent? evt, IEnumerable<ISentryEventProcessor> processors, SentryHint? hint)
+    public static SentryEvent? ProcessEvent(SentryEvent? evt, IEnumerable<ISentryEventProcessor> processors, SentryHint? hint, SentryOptions options)
     {
         if (evt == null)
         {
@@ -12,7 +12,7 @@ internal class SentryEventHelper(SentryOptions options)
         }
 
         var processedEvent = evt;
-        var effectiveHint = hint ?? new SentryHint();
+        var effectiveHint = hint ?? new SentryHint(options);
 
         foreach (var processor in processors)
         {
@@ -30,7 +30,7 @@ internal class SentryEventHelper(SentryOptions options)
 #if NET6_0_OR_GREATER
     [UnconditionalSuppressMessage("Trimming", "IL2026: RequiresUnreferencedCode", Justification = AotHelper.AvoidAtRuntime)]
 #endif
-    public SentryEvent? DoBeforeSend(SentryEvent? @event, SentryHint hint)
+    public static SentryEvent? DoBeforeSend(SentryEvent? @event, SentryHint hint, SentryOptions options)
     {
         if (@event is null || options.BeforeSendInternal is null)
         {
--- a/src/Sentry/Internal/SentryEventHelper.cs
+++ b/src/Sentry/Internal/SentryEventHelper.cs
@@ -1,0 +1,76 @@
+using Sentry.Extensibility;
+
+namespace Sentry.Internal;
+
+internal class SentryEventHelper(SentryOptions options)
+{
+    public SentryEvent? ProcessEvent(SentryEvent? evt, IEnumerable<ISentryEventProcessor> processors, SentryHint? hint)
+    {
+        if (evt == null)
+        {
+            return evt;
+        }
+
+        var processedEvent = evt;
+        var effectiveHint = hint ?? new SentryHint();
+
+        foreach (var processor in processors)
+        {
+            processedEvent = processor.DoProcessEvent(processedEvent, effectiveHint);
+            if (processedEvent == null)
+            {
+                options.ClientReportRecorder.RecordDiscardedEvent(DiscardReason.EventProcessor, DataCategory.Error);
+                options.LogInfo("Event dropped by processor {0}", processor.GetType().Name);
+                break;
+            }
+        }
+        return processedEvent;
+    }
+
+#if NET6_0_OR_GREATER
+    [UnconditionalSuppressMessage("Trimming", "IL2026: RequiresUnreferencedCode", Justification = AotHelper.AvoidAtRuntime)]
+#endif
+    public SentryEvent? DoBeforeSend(SentryEvent? @event, SentryHint hint)
+    {
+        if (@event is null || options.BeforeSendInternal is null)
+        {
+            return @event;
+        }
+
+        options.LogDebug("Calling the BeforeSend callback");
+        try
+        {
+            @event = options.BeforeSendInternal?.Invoke(@event, hint);
+            if (@event == null) // Rejected event
+            {
+                options.ClientReportRecorder.RecordDiscardedEvent(DiscardReason.BeforeSend, DataCategory.Error);
+                options.LogInfo("Event dropped by BeforeSend callback.");
+            }
+        }
+        catch (Exception e)
+        {
+            if (!AotHelper.IsTrimmed)
+            {
+                // Attempt to demystify exceptions before adding them as breadcrumbs.
+                e.Demystify();
+            }
+
+            options.LogError(e, "The BeforeSend callback threw an exception. It will be added as breadcrumb and continue.");
+            var data = new Dictionary<string, string>
+            {
+                {"message", e.Message}
+            };
+            if (e.StackTrace is not null)
+            {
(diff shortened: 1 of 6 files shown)


src/Sentry/Platforms/Cocoa/SentrySdk.cs
```diff
@@ -224,32 +238,55 @@ private static CocoaSdk.SentryHttpStatusCodeRange[] GetFailedRequestStatusCodes(
...
             return evt!;
         }
         catch (Exception ex)
         {
-            options.LogError(ex, "Before Send Error");
+            options.LogError(ex, "Error running managed event processors for native event");
             return evt;
         }
+
+        IEnumerable<ISentryEventProcessor> GetEventProcessors()
```
Flash0ver: perf(nitpick): Consider making local function `static` and pass `IHub`

We are capturing the `IHub` local variable of the parent method.
To be fair, I believe that the generated closure type to capture the variable from the enclosing scope is a `struct`, yet the invocation of the local function without capture/closure should be ever so slightly faster.

src/Sentry/Platforms/Cocoa/SentrySdk.cs
```diff
@@ -224,32 +238,55 @@ private static CocoaSdk.SentryHttpStatusCodeRange[] GetFailedRequestStatusCodes(
...
+        {
+            if (hub is Hub fullHub)
+            {
+                return fullHub.ScopeManager.GetCurrent().Key.GetAllEventProcessors();
+            }
+            IEnumerable<ISentryEventProcessor>? eventProcessors = null;
+            hub.ConfigureScope(scope => eventProcessors = scope.GetAllEventProcessors());
```
Flash0ver, line 285: ~perf: use new `ConfigureScope` overload and pass `eventProcessors` rather than capturing it in closure~

We should be able to avoid the closure allocation of capturing the enclosing local variable that we assign to.

Something I could apply to logs (#4308) too.
And perhaps at some more locations, too.

---

**Update**
Oops ... I just realized that this doesn't work. It's copy _ByVal_, so it's a new reference.
We would need to pass it _ByRef_, but we can't via an `Action` ... so we would need to declare our own custom `delegate` type that takes the `TArg` _ByRef_ and then provide additional overloads to our `ConfigureScope` methods utilizing the custom `delegate` type.
The complexity is not worth it.
And to be fair ... in most cases we do have a `Hub` and we won't enter this code branch anyway.

src/Sentry/Platforms/Cocoa/SentrySdk.cs
```diff
@@ -224,32 +238,55 @@ private static CocoaSdk.SentryHttpStatusCodeRange[] GetFailedRequestStatusCodes(
...
+            }
 
             // we only support a subset of mutated data to be passed back to the native SDK at this time
-            result.CopyToCocoaSentryEvent(evt);
+            processedEvent.CopyToCocoaSentryEvent(evt);
 
-            // Note: Nullable result is allowed but delegate is generated incorrectly
+            // Note: A nullable result is allowed, but delegate is generated incorrectly
             // See https://github.com/xamarin/xamarin-macios/issues/15299#issuecomment-1201863294
             return evt!;
```
Flash0ver: nullable-suppressions: I believe we can remove this nullable suppression, as we return a nullable `CocoaSdk.SentryEvent?`, if I'm not missing something.

src/Sentry/Internal/SentryEventHelper.cs
```diff
@@ -0,0 +1,76 @@
+using Sentry.Extensibility;
+
+namespace Sentry.Internal;
+
+internal class SentryEventHelper(SentryOptions options)
```
Flash0ver: perf: Consider making `static`.

This type does not have any state / fields, and may be instantiated more than once.
Would it make sense to
- make the type `static`
- rename to plural `SentryEventHelpers`
- pass `SentryOptions` as first parameter to both methods

to avoid an instance allocation.

---

**Update**
I just noticed that it's also used in `SentryClient` as a field, and not just in `Cocoa/SentrySdk` where it is instantiated as a local variable.
Not sure how "worth" the refactoring is.
I'll refer to your preference.

src/Sentry/Internal/SentryEventHelper.cs
```diff
@@ -0,0 +1,76 @@
...
+{
+    public SentryEvent? ProcessEvent(SentryEvent? evt, IEnumerable<ISentryEventProcessor> processors, SentryHint? hint)
+    {
+        if (evt == null)
+        {
+            return evt;
+        }
+
+        var processedEvent = evt;
+        var effectiveHint = hint ?? new SentryHint();
```
Flash0ver: nitpick: Pass `options` to `SentryHint`.

I believe we have a `.ctor` overload of `SentryHint` that takes `SentryOptions`, rather than getting the options from the current Hub.

src/Sentry/Internal/SentryEventHelper.cs
```diff
@@ -0,0 +1,76 @@
...
+            return @event;
+        }
+
+        options.LogDebug("Calling the BeforeSend callback");
+        try
+        {
+            @event = options.BeforeSendInternal?.Invoke(@event, hint);
+            if (@event == null) // Rejected event
+            {
+                options.ClientReportRecorder.RecordDiscardedEvent(DiscardReason.BeforeSend, DataCategory.Error);
```
Flash0ver, line 46: question: Should we in general create a Client-Report when user-callbacks (`BeforeSend...`) request to drop an event?

I am asking because I may want to consider this for **Logs** too.

src/Sentry/Internal/SentryEventHelper.cs
```diff
@@ -0,0 +1,76 @@
+using Sentry.Extensibility;
+
+namespace Sentry.Internal;
+
+internal class SentryEventHelper(SentryOptions options)
+{
+    public SentryEvent? ProcessEvent(SentryEvent? evt, IEnumerable<ISentryEventProcessor> processors, SentryHint? hint)
```
Flash0ver: perf: Consider changing type of `processors` to `ImmutableArray<ISentryEventProcessor>`.

The caller creates an `ImmutableArray<ISentryEventProcessor> manualProcessors` via `ToImmutableArray`, rather than an `IEnumerable<ISentryEventProcessor>`.
Since `ImmutableArray<T>` is a `struct`, we have a boxing at the call site.

Alternatively, the caller could do just `ToArray` rather than `ToImmutableArray` to avoid the boxing allocation.
But `foreach`ing a concrete type rather than an `IEnumerable<T>` in this case would avoid the allocation or boxing of the underlying Enumerator-Type. So I would prefer changing this parameter to an `ImmutableArray<T>` (or alternatively `T[]`, but I do like the immutable collection here to communicate intent).

jamescrosswell: It's also called from SentryClient here:
https://github.com/getsentry/sentry-dotnet/blob/962346f2fe31f47b1615854643166b003c548448/src/Sentry/SentryClient.cs#L345

That one isn't an ImmutableArray... so I'll change the caller in the SentrySdk to be just `ToArray`.

src/Sentry/Platforms/Cocoa/SentrySdk.cs
jamescrosswell: I'll give it a go. It probably made sense back when Matt wrote it - the surrounding context must have changed a bit since then.

src/Sentry/Internal/SentryEventHelper.cs
jamescrosswell: I assume so... I simply refactored this code [from here](https://github.com/getsentry/sentry-dotnet/blob/6da328b407d284cf870d56a5cd0f8a1ba553b6d1/src/Sentry/SentryClient.cs#L365-L371)... Git blame says that was Matt's code (but then he might have been refactoring it from somewhere else). 

@bruno-garcia do you know?

bruno-garcia: yeah afaik we should report in those cases. it helps know how many events are filtered and for what reasons and helps with customers asking about why events are not making in

Flash0ver (approved):

PR #4322, opened 2025-07-03
ref: Support `Unity` as a platfom

bitsandfoxes:
This comes from https://github.com/getsentry/sentry-unity/pull/2227

The goal is to provide the Unity SDK a way to "internalize" the .NET SDK. This works because the Unity SDK builds the .NET SDK from source. The relevant changes happen in `SentrySdk.cs`.

We noticed the similarity between `Unity` and the platforms `Android`, `iOS`, and `Native`. All the Unity specific junk from the `Sentry.csproj` now goes into its own `.props`.

To lend the argument "Unity is its own platform" some credibility we dug out [some ms docs](https://learn.microsoft.com/en-us/dotnet/standard/net-standard?tabs=net-standard-1-0#net-standard-versions).

#skip-changelog

### DIFF
--- a/src/Sentry/Platforms/Unity/SentrySdk.Unity.cs
+++ b/src/Sentry/Platforms/Unity/SentrySdk.Unity.cs
@@ -1,0 +1,21 @@
+#if SENTRY_UNITY
+
+namespace Sentry;
+
+/// <summary>
+/// Internal Sentry SDK entrypoint.
+/// </summary>
+/// <remarks>
+/// This class is now internal. Use <c>Sentry.Unity.SentrySdk</c> instead.
+/// <para>
+/// To migrate your code:
+/// <list type="number">
+/// <item>Change <c>using Sentry;</c> to <c>using Sentry.Unity;</c></item>
+/// <item>Keep using the <c>SentrySdk</c> API itself - no changes needed to method calls</item>
+/// <item>Add <c>using Sentry;</c> if you need access to types like <c>SentryId</c>, <c>SentryLevel</c>, etc.</item>
+/// </list>
+/// </para>
+/// </remarks>
+internal static partial class SentrySdk;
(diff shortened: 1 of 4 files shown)


src/Sentry/Platforms/Unity/SentrySdk.Unity.cs
```diff
@@ -0,0 +1,21 @@
...
+/// This class is now internal. Use <c>Sentry.Unity.SentrySdk</c> instead.
+/// <para>
+/// To migrate your code:
+/// <list type="number">
+/// <item>Change <c>using Sentry;</c> to <c>using Sentry.Unity;</c></item>
+/// <item>Keep using the <c>SentrySdk</c> API itself - no changes needed to method calls</item>
+/// <item>Add <c>using Sentry;</c> if you need access to types like <c>SentryId</c>, <c>SentryLevel</c>, etc.</item>
+/// </list>
+/// </para>
+/// </remarks>
```
bitsandfoxes, line 18: This is only visible to Unity SDK users. The changes here are part of breaking changes to the Unity SDK and this aims to reduce friction when migrating.

src/Sentry/Sentry.csproj
```diff
@@ -7,29 +7,19 @@
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
   </PropertyGroup>
 
-  <PropertyGroup Condition="'$(SolutionName)' != 'Sentry.Unity'">
+  <PropertyGroup>
```
bitsandfoxes, line 10: We no longer need to conditionally set the TF, they get overwritten by the platform `.props` on line 4.

bruno-garcia (approved):
This is great work! Will simplify a lot working with Unity from this repo as well as the Unity one.

PR #4323, opened 2025-07-06
fix: Remove AppDomain.CurrentDomain.ProcessExit hook on shutdown

### LINKED ISSUE #3141: AppDomainProcessExitIntegration doesn't get unregistered when Sentry is disposed of/closed
### Package

Sentry

### .NET Flavor

.NET

### .NET Version

Any

### OS

Any (not platform specific)

### SDK Version

4.1.0

_No response_

### Steps to Reproduce

I'm not yet sure how to reproduce this in plain .net but I'm getting this in sentry-powershell tests when setting a custom diagnostic logger. The general idea of what's happening:

1. create a custom diagnostic logger instance that collects messages to a concurrent bag.
2. start sentry with this logger
3. close sentry
4. exit the program
5. we get a null-reference **when running in CI**

The full test file that reproduces this (relies on importing the Sentry powershell module):

```pwsh
Describe 'SentrySdk' {
    It 'init respects options' {
        class TestLogger:Sentry.Infrastructure.DiagnosticLogger
        {
            TestLogger([Sentry.SentryLevel]$level) : base($level) {}

            $entries = [System.Collections.Concurrent.ConcurrentQueue[string]]::new();

            [void]LogMessage([string] $message) { $this.entries.Enqueue($message); }
        }

        $logger = [TestLogger]::new([Sentry.SentryLevel]::Debug)

        $options = [Sentry.SentryOptions]::new()
        $options.Debug = $true
        $options.Dsn = 'https://key@127.0.0.1/1'
        $options.DiagnosticLogger = $logger
        [Sentry.SentrySdk]::init($options)
        [Sentry.SentrySdk]::IsEnabled | Should -Be $true

        $logger.entries.Count | Should -BeGreaterThan 0
        [Sentry.SentrySdk]::close()
    }
}
```

### Expected Result

Closing the SDK (or disposing the handle) should unregister the onExit handler so that it's not called anymore. Alternatively, the handler could check whether the SDK was closed already.

### Actual Result

https://github.com/getsentry/sentry-powershell/actions/runs/7857351729/job/21441148698?pr=9
```
Running tests.
[+] /home/runner/work/sentry-powershell/sentry-powershell/tests/init.tests.ps1 937ms (458ms|2[19](https://github.com/getsentry/sentry-powershell/actions/runs/7857351729/job/21441148698?pr=9#step:6:20)ms)
Tests completed in 951ms
Tests Passed: 4, Failed: 0, Skipped: 0 NotRun: 0
Processing code coverage result.

Unhandled exception. System.NullReferenceException: Object reference not set to an instance of an object.
   at System.Management.Automation.ScriptBlock.InvokeWithPipe(Boolean useLocalScope, ErrorHandlingBehavior errorHandlingBehavior, Object dollarUnder, Object input, Object scriptThis, Pipe outputPipe, InvocationInfo invocationInfo, Boolean propagateAllExceptionsToTop, List`1 variablesToDefine, Dictionary`2 functionsToDefine, Object[] args)
   at System.Management.Automation.ScriptBlock.InvokeAsMemberFunction(Object instance, Object[] args)
   at System.Management.Automation.Internal.ScriptBlockMemberMethodWrapper.InvokeHelper(Object instance, Object sessionStateInternal, Object[] args)
   at Sentry.Infrastructure.DiagnosticLogger.Log(SentryLevel logLevel, String message, Exception exception, Object[] args)
   at Sentry.Extensibility.DiagnosticLoggerExtensions.LogIfEnabled(IDiagnosticLogger logger, SentryLevel level, Exception exception, String message)
   at Sentry.Extensibility.DiagnosticLoggerExtensions.LogInfo(SentryOptions options, String message)
   at Sentry.Integrations.AppDomainProcessExitIntegration.HandleProcessExit(Object sender, EventArgs e)
   at Sentry.Internal.AppDomainAdapter.OnProcessExit(Object sender, EventArgs e)
Error: Process completed with exit code 134.
```

jamescrosswell:
Resolves #3141
- https://github.com/getsentry/sentry-dotnet/issues/3141

@vaind the repro involved adding the Sentry PowerShell package, which I assume wraps a specific version of the .NET SDK... so I wasn't sure how to go about testing whether this addressed the underlying error you were running into. Maybe building a new version of the powershell package or running a version of that package that is built using a source rather than package reference to sentry-dotnet?


src/Sentry/Internal/Hub.cs
```diff
@@ -84,6 +86,10 @@ internal Hub(
         {
             options.LogDebug("Registering integration: '{0}'.", integration.GetType().Name);
             integration.Register(this, options);
+            if (integration is ITidySdkIntegration tidyIntegration)
+            {
+                _integrationsToCleanup.Add(tidyIntegration);
```
Flash0ver: question: Could we instead "try to cast to the new interface and call the method" in `Dispose` as well?

If we iterate through all `options.Integrations` in `Dispose` again, and do the type-check there:
- no additional work in _Init_
- no additional allocation of a collection

But we would iterate through all `options.Integrations` yet again to find the ones that implement the new interface ... I guess the trade-off here is CPU vs Memory and optimizing Startup vs Shutdown.

Also, technically, user code could capture `SentryOptions` from an _Init_-callback and could `AddIntegration`/`RemoveIntegration` after the initialization occurred, so that we would iterate through potentially different Enumerables on _Startup_ and on _Shutdown_.

(I believe I kinda talked myself out of it already)

test/Sentry.Tests/HubTests.cs
```diff
@@ -1988,6 +1988,80 @@ public void CaptureUserFeedback_InvalidEmail_FeedbackDropped(string email)
...
+        _fixture.Options.AddIntegration(integration2);
+        _fixture.Options.AddIntegration(integration3);
+        var hub = _fixture.GetSut();
+
+        // Act
+        hub.Dispose();
+
+        // Assert
+        cleaned.Should().Contain(integration1);
+        cleaned.Should().Contain(integration2);
```
Flash0ver: question: Should we assert `integration3` as well?

We add 2 `ITidySdkIntegration` and 1 `ISdkIntegration`.
For completion, should we also assert that `integration3` is not contained.
Or instead perhaps assert that the `Count` of `cleaned` is exactly `2`?

Or is that redundant because of the type difference, where an instance of `ISdkIntegration` cannot be contained in a collection of `ITidySdkIntegration`.

src/Sentry/Integrations/ITidySdkIntegration.cs
```diff
@@ -0,0 +1,17 @@
+namespace Sentry.Integrations;
+
+/// <summary>
+/// Since we can't add methods to <see cref="ISdkIntegration"/> without it being  a breaking change, this interface
```
Flash0ver: note: I wish we could use default interface members ... but I guess we never can 😢

src/Sentry/Integrations/ITidySdkIntegration.cs
```diff
@@ -0,0 +1,17 @@
...
+/// allows us to add some cleanup logic as an internal interface. We can move this to <see cref="ISdkIntegration"/>
+/// in the next major release.
+/// </summary>
+internal interface ITidySdkIntegration : ISdkIntegration
```
Flash0ver: note: I'm personally not too fond of the name, since I haven't really encountered `Tidy` in a .NET API yet.
Also, I'm not too certain about `Cleanup`.

But let's postpone this discussion to when we actually make this feature publicly available.

An alternative could be similar to `IHostedService` and `Microsoft.Extensions.DependencyInjection`:
To call `Dispose` or `DisposeAsync` if the implementation of `ISdkIntegration` also implements `IDisposable` or `IAsyncDisposable`.

src/Sentry/Internal/Hub.cs
jamescrosswell: > no additional work in Init
> no additional allocation of a collection

Unfortunately not. `SentryOptions.Integrations` isn't actually a collection - it's a method that returns a bunch of integrations. If you call it multiple times, you get different instances of the integrations for each call. See [here](https://github.com/getsentry/sentry-dotnet/blob/723c84e6fe8a642be7ab2d40caf401bb6d38a8e3/src/Sentry/SentryOptions.cs#L160-L168).

It's an interesting design choice... unless we wanted to completely refactor all of that, we need to maintain some list of integrations in the Hub though (for when we want to call the `Cleanup()` method).

test/Sentry.Tests/HubTests.cs
jamescrosswell: > For completion, should we also assert that integration3 is not contained.

We could and I thought briefly about doing that. But it's technically impossible for integration3 to be in the collection, since it doesn't have any Cleanup method... so I don't think checking for that would do anything useful.

src/Sentry/Integrations/ITidySdkIntegration.cs
jamescrosswell: No I think they're .net9 plus only... so not much use to us until Microsoft retires .NET Framework in the year 2086.

src/Sentry/Internal/Hub.cs
Flash0ver: I see.
Yeah ... I guess for the refactoring we would need to manifest the `SentryOptions.Integrations` only once, and then reuse. And also, in order to ensure that passed `SentryOptions` instances are not changed after `SentrySdk.Init` either make a deep copy of the instance to disjoin it from the instance the user has a reference to, or "Freeze" the `SentryOptions` instance after `Init` so that it can no longer be modified, where _setters_ and Set-Methods become either _no-op_ or even throw.

Is this a change that we could imagine for a next major version?
(In short: make `SentryOptions` instance used in `Init` _read-only_ afterwards, or keep a deep copy internally)
Or at least something to consider and discuss further in a separate issue?

test/Sentry.Tests/HubTests.cs
Flash0ver: Right.
But would we later change the implementation, as the comment suggests, to move the `Cleanup` method to `ISdkIntegration`, then the `List<ITidySdkIntegration>` would also change to `List<ISdkIntegration>`, and then `integration3` could potentially be in the collection.

But I may be - well, I probably am - thinking way ahead of time, since when we change the implementation, the test needs to change as well, and we can then revise the tests accordingly.

src/Sentry/Integrations/ITidySdkIntegration.cs
Flash0ver: Follow-up question:
Is this a feature that we would/should add in an upcoming (potentially major) release?

Because then I can think of 3 options
- new interface for non-breaking addition
- new method in `ISdkIntegration`
- implicit support via `IDisposable` (unless we want to pass additional arguments to `Cleanup`)

If a new method, then my preference for the method name would be `Unregister` (similar to [CancellationTokenRegistration.Unregister](https://learn.microsoft.com/dotnet/api/system.threading.cancellationtokenregistration.unregister)).

Anyway, where I'm trying to get at:
Should we open an issue for that, to move the naming discussion there so that my nit-picking won't block this bugfix any further?

Also, I believe that `sentry-unity` could make use of this feature, too.

bitsandfoxes: Personally, I'm heavily in favor of extending the `ISdkIntegration` to something like
```
interface ISdkIntegration
{
    public void Register(IHub hub, SentryOptions options);
    public void Unregister();
}
```
that is getting called when the SDK closes down.

I've got a couple of integrations in the Unity SDK where I would benefit from having a place to detach the SDK from the engine again. I.e. [here](https://github.com/getsentry/sentry-unity/blob/5c1f1a36221a0b9d8185aaffc4e78c6ce0bb3d3d/src/Sentry.Unity/Integrations/UnityApplicationLoggingIntegration.cs#L34-L35)

src/Sentry/Internal/Hub.cs
bitsandfoxes: > Unfortunately not. SentryOptions.Integrations isn't actually a collection - it's a method that returns a bunch of integrations. If you call it multiple times, you get different instances of the integrations for each call. See [here](https://github.com/getsentry/sentry-dotnet/blob/723c84e6fe8a642be7ab2d40caf401bb6d38a8e3/src/Sentry/SentryOptions.cs#L160-L168).
It's an interesting design choice... unless we wanted to completely refactor all of that, we need to maintain some list of integrations in the Hub though (for when we want to call the Cleanup() method).

The fact that there is no place (not even the options) that holds a collection of actually registered integrations seems really odd to me.

src/Sentry/Integrations/ITidySdkIntegration.cs
jamescrosswell: Per @Flash0ver 's suggestion, I removed `ITidySdkIntegration` and simply used `IDisposable` instead... that way we don't have to wait for a new major release to roll this out (@bitsandfoxes you can start using this immediately in Unity).

test/Sentry.Tests/HubTests.cs
jamescrosswell: Even after removing the `ITidySdkIntegration`, I don't think there are any sensible checks we can make on the integration that doesn't implement `Cleanup`/`Dispose`. Let me know if you think there's anything additional that it would be useful to check in the new code.

src/Sentry/Internal/Hub.cs
jamescrosswell: I guess some of them don't really need to be held in memory... they're just functions that inject a dependency or wire up a hook.

I think with this PR, we can now retain a reference to any that we need to keep track of.

Flash0ver: Resolution:
Integrations added via `SentryOptions.DefaultIntegrations` are ephemeral and, more ore less, only containers to call `Register` on.
- Unless the type that implements `ISdkIntegration` additionally implements `IDisposable`.
- Then this integration is kept in memory to call `Dispose` on when the Host is shut down.

Integration added via `AddIntegration` are essentially rooted via the `SentryOptions` instance.
Perhaps we could un-root them from `SentryOptions` (by e.g. clearing the `List<ISdkIntegration> _integrations` that refers to and keeps them).
But this memory optimization is a smaller performance improvement for a follow-up changeset.

test/Sentry.Tests/HubTests.cs
Flash0ver: current tests LGTM ... thanks for the discussion and improvements

src/Sentry/Integrations/ITidySdkIntegration.cs
Flash0ver: thanks @jamescrosswell 
and thanks @bitsandfoxes for pointing out the usage scenarios in `sentry-unity`

Flash0ver (approved):

vaind:
thx for the fix @jamescrosswell  I'll verify once this gets to a released version (https://github.com/getsentry/sentry-powershell/pull/85)

vaind:
Seems to work fine - no more hangs in CI https://github.com/getsentry/sentry-powershell/pull/85

PR #4330, opened 2025-07-09
Default `SentryOptions.Debug = true` for Debug build configurations

### LINKED ISSUE #3212: Support Auto Debug Mode
This is the current code:
https://github.com/getsentry/sentry-dotnet/blob/6274944d00826b0442d038f675f4f46ac0fec182/src/Sentry/SentryOptions.cs#L620-L668

What this effectively does, AFAICT, is that we don't print any warnings when `Debug` is not set. That is pretty unexpected behavior, at least to me, especially considering what the docs say when you're setting up sentry:

```
    // When debug is enabled, the Sentry client will emit detailed debugging information to the console.
    // This might be helpful, or might interfere with the normal operation of your application.
    // We enable it here for demonstration purposes when first trying Sentry.
    // You shouldn't do this in your applications unless you're troubleshooting issues with Sentry.
    options.Debug = true;
```

Instead, I'd expect the DiagnosticLogger property to always be reachable. It already has a check which log levels should be printed.

jamescrosswell:
Resolves #3212:
- https://github.com/getsentry/sentry-dotnet/issues/3212


test/Sentry.Tests/Sentry.Tests.csproj
jamescrosswell, line 1: @Flash0ver any idea how I can get the source generator working in the `Sentry.Tests`?

I think we need some way to test this... my thinking was basically to wire up SentryTests the same way the Samples are wired up and then set expectations based on `#if DEBUG`, which is implicitly set by the build configuration as well.

src/Sentry/Internal/AotHelper.cs
```diff
@@ -13,16 +13,15 @@ static AotHelper()
...
     private static bool CheckIsTrimmed()
     {
-        if (TryGetBoolean("publishtrimmed", out var trimmed))
+        if (BuildProperties.TryGetBoolean("PublishTrimmed", out var trimmed))
         {
             return trimmed;
         }
 
-        if (TryGetBoolean("publishaot", out var aot))
+        if (BuildProperties.TryGetBoolean("PublishAot", out var aot))
```
jamescrosswell, line 24: I think this is a major bug that was introduced by the initial PR including the source generator...

test/Sentry.Tests/Sentry.Tests.csproj
jamescrosswell: Actually, from what I can tell it's not just in the unit tests... the source generators don't appear to be working anywhere (even in the main branch):
- https://github.com/getsentry/sentry-dotnet/issues/4331

src/Sentry/SentryOptions.cs
```diff
@@ -1324,6 +1325,18 @@ public SentryOptions()
         );
 
         NetworkStatusListener = new PollingNetworkStatusListener(this);
+
+#if NET8_0_OR_GREATER
```
Flash0ver, line 1329: suggestion: Remove the preprocessor directive.

User code could build a .NET Framework app or a `netcoreapp` or older `net` app, while using a .NET SDK - or another Toolchain invoking the Compiler - that is compatible with **Roslyn 4.3**, such as the `.NET SDK 6.0.400` or above.

In this case, I don't see a reason for "Auto Debug Mode" not to work.

The only thing, that we need to make very sure of, is that the source code that our Generator emits is compatible with `<LangVersion>7.3</LangVersion>`, which is the default C# language version used for _.NET Standard 2.0_ and _.NET Framework_ TFMs.
Or alternatively let the Generator only emit code if the consuming project supports the C# language version that our generated source code requires via [CSharpParseOptions.LanguageVersion](https://learn.microsoft.com/dotnet/api/microsoft.codeanalysis.csharp.csharpparseoptions.languageversion).

src/Sentry/CompilerServices/BuildProperties.cs
```diff
@@ -23,4 +23,27 @@ public static void Initialize(Dictionary<string, string> properties)
...
         Values ??= properties.AsReadOnly();
     }
+
+    /// <summary>
+    /// Tries to retrieve a boolean value from the build properties.
+    /// </summary>
+    /// <param name="key">The item key</param>
+    /// <param name="value">The value (if any)</param>
+    /// <returns>True if the key was found, false otherwise</returns>
+    public static bool TryGetBoolean(string key, out bool value)
```
Flash0ver, line 33: suggestion: make `internal`

I don't think this needs to be accessible from the user's assembly where the Generator is emitting source to.

---

On the same note ... I think that the only member here, that needs to be accessible from the user's assembly, is the method `Initialize(Dictionary<string, string>)`.

So I'm wondering if we should also make the `Values` property `internal`, since it should only be _Sentry_ code reading it. And it's only the `Initialize` method that needs to be accessible for the source-generated `[ModuleInitializer]` to work.

Technically a breaking change, but since our source generator wasn't emitting any source code up until now, this `IReadOnlyDictionary<string, string>?` was always `null` anyways, so perhaps **now** may just be the time to make the `Values` property `internal` too.

Perhaps
see also https://github.com/getsentry/sentry-dotnet/pull/4330/files#r2194786902
we could make the `Values` property all `private`, so that Sentry also only access it via `internal` methods.

src/Sentry/SentryOptions.cs
```diff
@@ -1324,6 +1325,18 @@ public SentryOptions()
...
+        // If source generators are available, we can detect the user's build configuration and use this
+        // to infer whether to enable debug mode by default.
+        // See: https://develop.sentry.dev/sdk/expected-features/#auto-debug-mode
+        string? buildConfig = null;
+        BuildProperties.Values?.TryGetValue("Configuration", out buildConfig);
+        if (buildConfig == "Debug")
```
Flash0ver, line 1335: question: Case-sensitivity?

I wonder if this string equality check should be case insensitive 🤔

```CSharp
buildConfig.Equals("Debug", StringComparison.OrdinalIgnoreCase)
```

Flash0ver, line 1335: nitpick(style): Could we make this a one-liner?

I find this code a bit verbose.
Perhaps we could make this a one-liner:
```CSharp
if (BuildProperties.TryGetString("Configuration", out var buildConfig) && buildConfig == "Debug")
```

by adding a new `internal` method to `BuildProperties`:
```CSharp
// public static class BuildProperties

internal static bool TryGetString(string key, [NotNullWhen(true)] out string? value)
{
  if (Values?.TryGetValue(key, out var foundValue) == true)
  {
    value = foundValue;
    return true;
  }

  value = null;
  return false;
}
```

and perhaps we could make the `BuildProperties.Values` property all `private`, so that we only have nicely consumable methods that work around the nullability of the `IReadOnlyDictionary<string, string>`.

test/Sentry.Tests/Sentry.Tests.csproj
```diff
@@ -44,4 +44,12 @@
   <ItemGroup Condition="'$(TargetFramework)' != 'net48'">
     <PackageReference Include="Microsoft.Extensions.Configuration.Binder" Version="8.0.1" />
   </ItemGroup>
+
+  <!-- These are needed to test the AutoDebug functionality in SentryOptions on net8.0 targets and greater -->
```
Flash0ver, line 48: note: Rephrase comment.

Depending on https://github.com/getsentry/sentry-dotnet/pull/4330/files#r2194449284
it may be independent of the target framework.

test/Sentry.Tests/SentryOptionsTests.cs
```diff
@@ -675,4 +675,17 @@ public void CachesInstallationId()
         installationId2.Should().Be(installationId1);
         logger.Received(0).Log(SentryLevel.Debug, "Resolved installation ID '{0}'.", null, Arg.Any<string>());
     }
+
+#if NET8_0_OR_GREATER
```
Flash0ver, line 679: note: Remove too.

Depending on https://github.com/getsentry/sentry-dotnet/pull/4330/files#r2194449284.

test/Sentry.Tests/Sentry.Tests.csproj
Flash0ver: Indeed ... @jamescrosswell and I figured out that the `[Generator]` is loaded correctly, and is actually executing as well in consuming projects, but we made a mistake in the implementation so that no sources are ever emitted.

src/Sentry/Internal/AotHelper.cs
Flash0ver: Great catch!
MSBuild is case-insensitive, and so should be the source that the `[Generator]` emits.
Will fix in #4331.

src/Sentry/SentryOptions.cs
jamescrosswell: I think it should, but we should configure the underlying dictionary to be case insensitive... that way we don't have to worry about it in method calls like this right?

src/Sentry/SentryOptions.cs
jamescrosswell: Looks like from https://github.com/getsentry/sentry-dotnet/pull/4333 we need to leave the `#if` clause in there but we can change it to `#if NET5_0_OR_GREATER`

samples/Directory.Build.targets
```diff
@@ -7,6 +7,7 @@
   <Import Project="$(MSBuildThisFileDirectory)..\src\Sentry.Bindings.Cocoa\buildTransitive\Sentry.Bindings.Cocoa.targets"
     Condition="'$(OutputType)' == 'Exe' And ('$(TargetPlatformIdentifier)' == 'ios' Or '$(TargetPlatformIdentifier)' == 'maccatalyst')" />
   <Import Project="$(MSBuildThisFileDirectory)..\src\Sentry\Platforms\Native\buildTransitive\Sentry.Native.targets" />
+  <Import Project="$(MSBuildThisFileDirectory)..\src\Sentry\buildTransitive\Sentry.SourceGenerators.targets" />
```
Flash0ver, line 10: note: We deleted `Sentry.SourceGenerators.targets` in #4333.

It was redundant in the NuGet package, as only `Sentry.props` and `Sentry.targets` are automatically picked up in consuming projects (via `PackageReference`), and `Sentry.SourceGenerators.targets` was never `Import`ed in any of these MSBuild files either.

src/Sentry/Internal/AotHelper.cs
Flash0ver: Fixed in #4333:

Changed the generated code to use case-insensitive lookup:
```diff
-BuildProperties.Initialize(new Dictionary<string, string>()
+BuildProperties.Initialize(new Dictionary<string, string>(StringComparer.OrdinalIgnoreCase)
```

src/Sentry/SentryOptions.cs
Flash0ver: Indeed ... I was not able to remove it entirely from the generated source.
The highest dependency is [ModuleInitializerAttribute](https://learn.microsoft.com/dotnet/api/system.runtime.compilerservices.moduleinitializerattribute), available since `net5.0`.

src/Sentry/SentryOptions.cs
Flash0ver: Via #4333, the generated lookup type is now case-insensitive:
```diff
-BuildProperties.Initialize(new Dictionary<string, string>()
+BuildProperties.Initialize(new Dictionary<string, string>(StringComparer.OrdinalIgnoreCase)
```
So we can lookup the `"Configuration"` Property case-insensitively.

But comparing the Value of the `"Configuration"`-Property here should also be done in a case-insensitive fashion. Or am I mixing something up here 🤔?

test/Sentry.Tests/Sentry.Tests.csproj
Flash0ver: Couldn't make it TFM-agnostic, but lowered it in #4333:
```diff
-#if NET8_0_OR_GREATER
+#if NET5_0_OR_GREATER
```
due to the generated code's dependency on [ModuleInitializerAttribute](https://learn.microsoft.com/dotnet/api/system.runtime.compilerservices.moduleinitializerattribute), added in `net5.0`.

test/Sentry.Tests/SentryOptionsTests.cs
Flash0ver: Couldn't make it TFM-agnostic, but lowered it in #4333:
```diff
-#if NET8_0_OR_GREATER
+#if NET5_0_OR_GREATER
```
due to the generated code's dependency on [ModuleInitializerAttribute](https://learn.microsoft.com/dotnet/api/system.runtime.compilerservices.moduleinitializerattribute), added in `net5.0`.

bitsandfoxes:
I'd just be careful putting this change in anything but a mayor release. There are plenty of people out there having debug builds in production. 
The SDK suddenly having opinions regarding when and where it logs sounds like a huge footgun to me and it even warns about the impact logging has on performance
https://github.com/getsentry/sentry-dotnet/blob/2c9c6d0e469960111a96d5ea8ca4c1f6ed9aa983/src/Sentry/SentryOptions.cs#L1770-L1773

bruno-garcia (changes requested):
We'll spam all our customers console when they are developing?
Don't think we should do this

bruno-garcia:
If we had `level: Error` we could consider this (what we do on Unity) but debug logs by default are there to help ppl understand why Sentry isn't working, and it's very verbose. The downside of that is that support and sales are trained to say `set debug:true` and in this case it'd be: `set diagnosticLevel to sentrylevel.debug`

jamescrosswell:
> We'll spam all our customers console when they are developing?
Don't think we should do this
> ...
> If we had `level: Error` we could consider this (what we do on Unity) but debug logs by default are there to help ppl understand why Sentry isn't working, and it's very verbose. The downside of that is that support and sales are trained to say `set debug:true` and in this case it'd be: `set diagnosticLevel to sentrylevel.debug`

So, basically, we shouldn't do this...

jamescrosswell:
Most of the changes from this PR were maintenance rather than functional... moved these out to https://github.com/getsentry/sentry-dotnet/pull/4344 instead. 

Closing this PR as it doesn't look like there's any elegant way to introduce auto-debug.

PR #4333, opened 2025-07-09
fix: BuildProperty-Generator and AOT compilation detection

### LINKED ISSUE #4331: Source Generators not working
### Description

As far as I can tell, the `Sentry.SourceGenerators` aren't working at all. 

### Steps to reproduce

1. `dotnet new console -f net9.0`
2. `dotnet add package Sentry --version 5.11.2`
3. Copy/paste the following into `program.cs`
4. `dotnet run -c Debug`

```csharp
using Sentry.CompilerServices;

SentrySdk.Init();

string? buildConfig = null;
BuildProperties.Values?.TryGetValue("Configuration", out buildConfig);
if (buildConfig is null)
{
    Console.WriteLine("Source generators not working...");
}
else if (buildConfig == "Debug")
{
    Console.WriteLine("You're in Debug Mode");
}
else
{
    Console.WriteLine("You ain't in Debug Mode anymore Toto!");
}
```

### Expected result

The application should output `You're in Debug Mode`.

### Actual result

The application outputs `Source generators not working...`

Flash0ver:
Resolves #4331:
- https://github.com/getsentry/sentry-dotnet/issues/4331

Proposing a bunch of changes to the Generated-code, considering the Generator so far never emitted any sources to the consuming compilation
- missing `<CompilerVisibleProperty Include="OutputType" />` in packed `Sentry.targets`, so the Generator never emitted any sources
  - replaced with `Compilation.Options.OutputKind`
  - this is more stable since it's evaluated Roslyn, rather than depending on an MSBuild Property
- change `new Dictionary<string, string>` to `new Dictionary<string, string>(StringComparer.OrdinalIgnoreCase)`
  - MSBuild is case-insensitive ... an so should our evaluation of Build-Properties be
  - see also https://github.com/getsentry/sentry-dotnet/issues/4331#issuecomment-3052519926
- change from `#if NET8_0_OR_GREATER` to `#if NET5_0_OR_GREATER`
  - [System.Runtime.CompilerServices.ModuleInitializerAttribute](https://learn.microsoft.com/dotnet/api/system.runtime.compilerservices.moduleinitializerattribute) is available since `net5.0`
  - the Source-Generator does not use any other types that are not available in `net5.0`
- added `#nullable enable`
  - `net5.0` has a default `LangVersion` of `9.0`
  - generated code does not consider the project's default of `<Nullable>enable</Nullable>`
  - so every generated document needs to set `#nullable enable` individually
- make generated type and member `internal`
  - a `[ModuleInitializer]` must be accessible from the containing module
  - so effectively `public` or `internal` in a non-`file`-local-type in a non-local-function
  - see [Module Initializers](https://learn.microsoft.com/dotnet/csharp/language-reference/proposals/csharp-9.0/module-initializers)
- replace `System.Runtime.CompilerServices.CompilerGeneratedAttribute` with `System.CodeDom.Compiler.GeneratedCodeAttribute`
  - the `CompilerGeneratedAttribute` attribute is reserved for Compiler usage only
    - e.g. for members that the Compiler generates for `record`-types
  - `GeneratedCodeAttribute` is intended for Tools and Roslyn-Generator authors
- renamed emitted Type-Name from `BuildVariableInitializer` to `BuildPropertyInitializer`
  - for consistency
- renamed HintName from `__BuildProperties.g.cs` to `Sentry.Generated.BuildPropertyInitializer.g.cs`
  - for consistency with the ecosystem
- fix inconsistent indentation
  - from mixed tabs and spaces
  - to spaces only

Additionally, proposing changes to the Generator itself and our MSBuild / Packaging
- deleted redundant `Sentry.SourceGenerators.targets`
  - only `Sentry.targets` is automatically picked up from Sentry's NuGet-Package
  - and `Sentry.SourceGenerators.targets` was never `Import`ed in any shipped Build-files
- made `BuildPropertySourceGenerator` Not-Inheritable

Tests
- slightly refactored existing tests
- added new tests
- added Verify-Scrubber that changes the Version of the generated `GeneratedCodeAttribute` to the string `"Version"` for test stability between releases

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,7 @@
 - Sentry now decompresses Request bodies in ASP.NET Core when RequestDecompression middleware is enabled ([#4315](https://github.com/getsentry/sentry-dotnet/pull/4315))
 - Custom ISentryEventProcessors are now run for native iOS events ([#4318](https://github.com/getsentry/sentry-dotnet/pull/4318))
 - Crontab validation when capturing checkins ([#4314](https://github.com/getsentry/sentry-dotnet/pull/4314))
-- Fixed an issue with the way Sentry detects build settings. This was causing Sentry to produce code that would could fail at runtime in AOT compiled applications. ([#4333](https://github.com/getsentry/sentry-dotnet/pull/4333))
+- Fixed an issue with the way Sentry detects build settings. This was causing Sentry to produce code that could fail at runtime in AOT compiled applications. ([#4333](https://github.com/getsentry/sentry-dotnet/pull/4333))
 - Native AOT: link to static `lzma` on Linux/MUSL ([#4326](https://github.com/getsentry/sentry-dotnet/pull/4326))
 - AppDomain.CurrentDomain.ProcessExit hook is now removed on shutdown ([#4323](https://github.com/getsentry/sentry-dotnet/pull/4323))
 


jamescrosswell:
A couple of the integration tests are failing since they're erroneously detecting the the app has been compiled AOT when this is not the case. 

I believe it's due to this:
https://github.com/getsentry/sentry-dotnet/blob/24ab5091021a299199189936860dc1979eecabc2/integration-test/runtime.Tests.ps1#L94-L105

The test itself is valid. The `PublishAot=true` build setting only has any effect when publishing (not when building)... which I think means we probably need to change this logic:
https://github.com/getsentry/sentry-dotnet/blob/0514bfe4b1e7ee0c3cb030bdd199ab974f66ca20/src/Sentry/Internal/AotHelper.cs#L20-L28

A couple of things would need to change in that logic then:
1. We'd also need to detect whether a publish operation is being performed. In MS Build that can be done by inspecting the `_IsPublishing` build property... 
2. `PublishTrimmed` has the same caveat plus another one
    - Trimming only occurs during a publish operation (not a build operation)
    - Trimming only occurs if `PublishSelfContained=true`

CHANGELOG.md
```diff
@@ -15,6 +15,7 @@
 - Sentry now decompresses Request bodies in ASP.NET Core when RequestDecompression middleware is enabled ([#4315](https://github.com/getsentry/sentry-dotnet/pull/4315))
 - Custom ISentryEventProcessors are now run for native iOS events ([#4318](https://github.com/getsentry/sentry-dotnet/pull/4318))
 - Crontab validation when capturing checkins ([#4314](https://github.com/getsentry/sentry-dotnet/pull/4314))
+- BuildProperty-Generator emits sources when consuming project is an Application ([#4333](https://github.com/getsentry/sentry-dotnet/pull/4333))
```
```suggestion
- Fixed an issue with the way Sentry detects build settings. This was causing Sentry to produce code that would could fail at runtime in AOT compiled applications. ([#4333](https://github.com/getsentry/sentry-dotnet/pull/4333))
```

src/Sentry/SentrySdk.cs
```diff
@@ -51,6 +51,7 @@ internal static IHub InitHub(SentryOptions options)
 
 #pragma warning disable CS0162 // Unreachable code detected
 #pragma warning disable 0162 // Unreachable code on old .NET frameworks
+        AotHelper.CheckIsTrimmed(options.DiagnosticLogger);
```
Flash0ver, line 57: question: Do we need this invocation?

Aren't we calling `CheckIsTrimmed` now twice?

jamescrosswell: This one was just to get logging of what's going on inside that method. As long as customers never have any issues with this particular part of the solution, we don't need that extra logging. I think it's worth leaving it in there though... partciularly since `_isPublishing` isn't really a public property (but there isn't anything else that we can use, that I'm aware of).

src/Sentry/buildTransitive/Sentry.targets
```diff
@@ -6,9 +6,11 @@
   </ItemGroup>
 
   <ItemGroup>
+    <CompilerVisibleProperty Include="_IsPublishing" />
```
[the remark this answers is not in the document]
Flash0ver: We deleted `Sentry.SourceGenerators.targets`, and it's no longer packed into the NuGet package.
Only `Sentry.targets` is automatically imported to the consuming Project via a `PackageReference`, and `Sentry.SourceGenerators.targets` was never `Import`ed from any "reachable" MSBuild.

src/Sentry/SentrySdk.cs
Flash0ver: 👌 ... let's follow-up on that in a different changeset as it's only a small perf-improvement to calling this method only once

Flash0ver:
Additional Test of the CI-produced NuGet package, consumed via a Local NuGet Feed:

```CSharp
if (Sentry.CompilerServices.BuildProperties.Values is null)
{
    Console.WriteLine("not initialized");
    return;
}

foreach (KeyValuePair<string, string> value in Sentry.CompilerServices.BuildProperties.Values)
{
    Console.WriteLine($"Key: {value.Key}, Value: {value.Value}");
}
```

> prints the available build variables, set via the `[ModuleInitializer]` of the generated code

image

jamescrosswell:
So we had a working PR back on commit https://github.com/getsentry/sentry-dotnet/pull/4333/commits/8fbade972cf23795ade86dca05a4c7686ae0f4ef... then the iOS device tests started failing.

Nothing in the new commits looks suspicious (essentially just tidying up around the edges). To test whether the new commits were potentially related, I forked this PR back on https://github.com/getsentry/sentry-dotnet/pull/4333/commits/8fbade972cf23795ade86dca05a4c7686ae0f4ef and pushed it up in a new branch as https://github.com/getsentry/sentry-dotnet/pull/4335... and indeed, the iOS device tests are failing there as well. 

Possibly the iOS device tests are flaky, and this is only apparent now that the source generators (and the AOT compilation detection) is now working?

jamescrosswell:
The logs in the broken CI runs are showing:
```diff
Failed to install the application
Cleaning up the failed installation from 'iPhone Xs (iOS 18.5) - created by XHarness'
Uninstalling the application 'io.sentry.dotnet.maui.device.testapp' from 'iPhone Xs (iOS 18.5) - created by XHarness'
Failed to uninstall the app bundle! Check logs for more details!
XHarness exit code: 78 (PACKAGE_INSTALLATION_FAILURE)
```

About the only reference I can find to this error is here:
- https://github.com/dotnet/maui/issues/20107#issuecomment-1922726082

CHANGELOG.md
```diff
@@ -15,6 +15,7 @@
 - Sentry now decompresses Request bodies in ASP.NET Core when RequestDecompression middleware is enabled ([#4315](https://github.com/getsentry/sentry-dotnet/pull/4315))
 - Custom ISentryEventProcessors are now run for native iOS events ([#4318](https://github.com/getsentry/sentry-dotnet/pull/4318))
 - Crontab validation when capturing checkins ([#4314](https://github.com/getsentry/sentry-dotnet/pull/4314))
+- Fixed an issue with the way Sentry detects build settings. This was causing Sentry to produce code that would could fail at runtime in AOT compiled applications. ([#4333](https://github.com/getsentry/sentry-dotnet/pull/4333))
```
```suggestion
- Fixed an issue with the way Sentry detects build settings. This was causing Sentry to produce code that could fail at runtime in AOT compiled applications. ([#4333](https://github.com/getsentry/sentry-dotnet/pull/4333))
```

jamescrosswell (approved):

Issue #4343, opened 2025-07-13
Don't load SentryNative on unsupported platforms


jamescrosswell:
> @devtekve Thanks a lot for the test case, very helpful! 👍 I should've guessed from "Mac" being mentioned that this was ARM64. 🙂 sentry-dotnet currently has native support for `linux-x64`, `linux-arm64`, and `linux-musl-x64`. The target you use i.e. `linux-musl-arm64` is missing, unfortunately. While sentry-dotnet should be fixed not to try to load the native library on an unsupported platform, you probably don't even need native for your deployment, maybe you could set `-p:SentryNative=false` to get past the loading issue? Feel free to open a request to add native support on `linux-musl-arm64` if it's relevant for your use case. 

 _Originally posted by @jpnurmi in [#4296](https://github.com/getsentry/sentry-dotnet/issues/4296#issuecomment-3066775157)_

jpnurmi:
Testing anything but `x86_64` with Alpine containers might be painful in GitHub Actions because even basic things such as `actions/checkout` doesn't work... 🙁 

<img width="2186" height="279" alt="Image" src="https://github.com/user-attachments/assets/133aec35-36a5-4ad8-b60a-b43a2fc06eb6" />

For some reason, `actions/runner` has a hard-coded architecture check for JS actions, so it doesn't even matter if we prepare the image with `nodejs` and `npm` pre-installed:
```csharp
            if (!Constants.Runner.PlatformArchitecture.Equals(Constants.Architecture.X64))
            {
                var os = Constants.Runner.Platform.ToString();
                var arch = Constants.Runner.PlatformArchitecture.ToString();
                var msg = $"JavaScript Actions in Alpine containers are only supported on x64 Linux runners. Detected {os} {arch}";
                throw new NotSupportedException(msg);
            }
```
https://github.com/actions/runner/blob/main/src/Runner.Worker/Handlers/StepHost.cs#L268-L274

jpnurmi:
- GitHub Actions container runners use pre-built Node from `/__e/node20` bind-mounted from the host.

  > The runner carries its own copy of node.js executable under `<runner_root>/externals/node20/`.
  >
  > All javascript base Actions will get executed by the built-in `node` at `<runner_root>/externals/node20/`.
  >
  >> Not the `node` from `$PATH`

  https://github.com/actions/runner/blob/main/docs/checks/nodejs.md

[...]

PR #4347, opened 2025-07-14
fix: Don't load SentryNative on unsupported platforms

jpnurmi:
TLDR; default `Sentry.Native.IsEnabled` to false for unsupported Native AOT platforms.

The existing `integration-test/aot.Tests.ps1` works great for testing unsupported platforms with a minor tweak to allow specifying the desired target `RuntimeIdentifier` in the environment. It only tests `dotnet publish` for a minimal Sentry console app, and that the compiled binary executes ok (without capturing any native exceptions). The original purpose was to test that `libsentry-native.a` doesn't have problematic dependencies (`libcurl.so`), but the same integration test works just as well for testing that sentry-dotnet doesn't try to load a non-existent `libsentry-native.so`.

Without the newly introduced changes that make `Sentry.Native.IsEnabled` default to false for unsupported Native AOT platforms, the test would produce such failures:

### `win-x86` (Windows 32-bit)
```
  Debug: This looks like a Native AOT application build.
Unhandled exception. System.DllNotFoundException: Unable to load DLL 'sentry-native' or one of its dependencies: The specified module could not be found.
   at System.Runtime.InteropServices.NativeLibrary.LoadLibErrorTracker.Throw(String) + 0x42
   at Internal.Runtime.CompilerHelpers.InteropHelpers.FixupModuleCell(InteropHelpers.ModuleFixupCell*) + 0xdf
   at Internal.Runtime.CompilerHelpers.InteropHelpers.ResolvePInvokeSlow(InteropHelpers.MethodFixupCell*) + 0x29
   at Internal.Runtime.CompilerHelpers.InteropHelpers.ResolvePInvoke(InteropHelpers.MethodFixupCell*) + 0xb
   at Sentry.Native.C.sentry_options_new() + 0x13
   at Sentry.Native.C.Init(SentryOptions) + 0x17
   at Sentry.SentrySdk.InitNativeSdk(SentryOptions) + 0x1e
   at Sentry.SentrySdk.InitHub(SentryOptions) + 0x179
   at Sentry.SentrySdk.Init(Action`1) + 0x2d
   at Program.<Main>$(String[] args) + 0x3e
```

### `linux-arm` (Linux ARM 32-bit)
```
  Debug: This looks like a Native AOT application build.
Unhandled exception. System.DllNotFoundException: Unable to load shared library 'sentry-native' or one of its dependencies. In order to help diagnose loading problems, consider using a tool like strace. If you're using glibc, consider setting the LD_DEBUG environment variable:
sentry-native.so: cannot open shared object file: No such file or directory
libsentry-native.so: cannot open shared object file: No such file or directory
sentry-native: cannot open shared object file: No such file or directory
libsentry-native: cannot open shared object file: No such file or directory

   at System.Runtime.InteropServices.NativeLibrary.LoadLibErrorTracker.Throw(String) + 0x5b
   at Internal.Runtime.CompilerHelpers.InteropHelpers.FixupModuleCell(InteropHelpers.ModuleFixupCell*) + 0x16f
   at Internal.Runtime.CompilerHelpers.InteropHelpers.ResolvePInvokeSlow(InteropHelpers.MethodFixupCell*) + 0x3d
   at Sentry.Native.C.sentry_options_new() + 0x1f
   at Sentry.Native.C.Init(SentryOptions) + 0x29
   at Sentry.SentrySdk.InitNativeSdk(SentryOptions) + 0x31
   at Sentry.SentrySdk.InitHub(SentryOptions) + 0x1e9
   at Sentry.SentrySdk.Init(Action`1) + 0x49
   at Program.<Main>$(String[] args) + 0x69
qemu: uncaught target signal 6 (Aborted) - core dumped
```

Note: These two platforms were chosen because both are straightforward to set up for testing. If either platform becomes supported in the future, the respective matrix item can be removed.

Resolves: #4343

### DIFF
--- a/integration-test/aot.Tests.ps1
+++ b/integration-test/aot.Tests.ps1
@@ -50,11 +50,22 @@ Console.WriteLine("Hello, Sentry!");
     }
 
     It 'Aot' {
-        dotnet publish -c Release | Write-Host
+        $rid = $env:RuntimeIdentifier
+        if ($env:RuntimeIdentifier)
+        {
+            dotnet publish -c Release -r $rid | Write-Host
+        }
+        else
+        {
+            dotnet publish -c Release | Write-Host
+        }
         $LASTEXITCODE | Should -Be 0
 
         $tfm = (Get-ChildItem -Path "bin/Release" -Directory | Select-Object -First 1).Name
-        $rid = (Get-ChildItem -Path "bin/Release/$tfm" -Directory | Select-Object -First 1).Name
+        if (-not $rid)
+        {
+            $rid = (Get-ChildItem -Path "bin/Release/$tfm" -Directory | Select-Object -First 1).Name
+        }
         & "bin/Release/$tfm/$rid/publish/hello-sentry" | Write-Host
         $LASTEXITCODE | Should -Be 0
     }
--- a/integration-test/aot.Tests.ps1
+++ b/integration-test/aot.Tests.ps1
@@ -51,7 +51,7 @@ Console.WriteLine("Hello, Sentry!");
 
     It 'Aot' {
         $rid = $env:RuntimeIdentifier
-        if ($env:RuntimeIdentifier)
+        if ($rid)
         {
             dotnet publish -c Release -r $rid | Write-Host
         }
--- a/src/Sentry/Platforms/Native/SentryNative.cs
+++ b/src/Sentry/Platforms/Native/SentryNative.cs
@@ -17,7 +17,7 @@ internal static class SentryNative
     // This way, `SentryNative.IsEnabled` should be treated as a compile-time constant for trimmed apps.
     [FeatureSwitchDefinition(SentryNativeIsEnabledSwitchName)]
 #endif
-    private static bool IsEnabled => !AppContext.TryGetSwitch(SentryNativeIsEnabledSwitchName, out var isEnabled) || isEnabled;
+    private static bool IsEnabled => AppContext.TryGetSwitch(SentryNativeIsEnabledSwitchName, out var isEnabled) && isEnabled;
 
     internal static bool IsAvailable => IsEnabled && IsAvailableCore;
 
--- a/src/Sentry/Platforms/Native/buildTransitive/Sentry.Native.targets
+++ b/src/Sentry/Platforms/Native/buildTransitive/Sentry.Native.targets
@@ -8,16 +8,6 @@
   Note:Target framework conditions should be kept synchronized with src/Sentry/buildTransitive/Sentry.Native.targets -->
 <Project>
 
-  <ItemGroup>
-    <!-- When user sets <SentryNative>false</SentryNative> or <SentryNative>disable</SentryNative> in their project -->
-    <!-- SentryNative.IsEnabled should result in compile-time constant for trimmed applications -->
-    <!-- Effectively disabling native library -->
-    <RuntimeHostConfigurationOption Include="Sentry.Native.IsEnabled"
-                                    Condition="'$(FrameworkSupportsNative)' == 'true'"
-                                    Value="true"
-                                    Trim="true" />
-  </ItemGroup>
-
   <!-- Helpful properties copied from Sentry.props -->
   <PropertyGroup Condition="'$(TargetFramework)' != ''">
     <_SentryTargetFrameworkVersion>$([MSBuild]::GetTargetFrameworkVersion($(TargetFramework)))</_SentryTargetFrameworkVersion>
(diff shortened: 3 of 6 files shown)


jamescrosswell (approved):
LGTM - thanks @jpnurmi !

integration-test/aot.Tests.ps1
```diff
@@ -50,11 +50,22 @@ Console.WriteLine("Hello, Sentry!");
     }
 
     It 'Aot' {
-        dotnet publish -c Release | Write-Host
+        $rid = $env:RuntimeIdentifier
+        if ($env:RuntimeIdentifier)
```
Flash0ver: nitpick: check `$rid` instead of `$env:RuntimeIdentifier`

I am wondering if it's more clear to use `$rid` for the check,
since below, we also use `if (-not $rid)`.

But I guess that's very much a style question.
My care-factor is neglectable 😉

src/Sentry/Platforms/Native/SentryNative.cs
```diff
@@ -17,7 +17,7 @@ internal static class SentryNative
     // This way, `SentryNative.IsEnabled` should be treated as a compile-time constant for trimmed apps.
     [FeatureSwitchDefinition(SentryNativeIsEnabledSwitchName)]
 #endif
-    private static bool IsEnabled => !AppContext.TryGetSwitch(SentryNativeIsEnabledSwitchName, out var isEnabled) || isEnabled;
+    private static bool IsEnabled => AppContext.TryGetSwitch(SentryNativeIsEnabledSwitchName, out var isEnabled) && isEnabled;
```
Flash0ver, line 20: praise:
If I understand it right, we're changing the behavior - in conjunction with the changes in `Sentry.Native.targets`
- from `true` when `Sentry.Native.IsEnabled` is _not_ set **or** set to `true`
- to `true` when `Sentry.Native.IsEnabled` _is_ set **and** set to `true`

I do find that semantics more clear than before ... it's more of a `false` by default, unless we detect it's supported, unless the consuming projects opts out via `SentryNative`.

src/Sentry/Platforms/Native/buildTransitive/Sentry.Native.targets
```diff
@@ -13,8 +13,8 @@
     <!-- SentryNative.IsEnabled should result in compile-time constant for trimmed applications -->
     <!-- Effectively disabling native library -->
     <RuntimeHostConfigurationOption Include="Sentry.Native.IsEnabled"
-                                    Condition="'$(SentryNative)' == 'false' or '$(SentryNative)' == 'disable'"
-                                    Value="false"
+                                    Condition="'$(FrameworkSupportsNative)' == 'true'"
```
Flash0ver: suggestion (non-blocking): move this `ItemGroup` just below the modified `PropertyGroup`

Since `ItemGroup`s do get evaluated after `PropertyGroup`s,
I wonder if it makes this project file more clear if we'd move this `ItemGroup` (that uses `FrameworkSupportsNative`), just below the `PropertyGroup`, that sets the `FrameworkSupportsNative` property.

So we also have visually a "set before use" flow.

(non-PR-blocking style suggestion)

Flash0ver (approved):
thanks @jpnurmi for the many improvements around CI and Native
I got 4 non-blocking comments

src/Sentry/Platforms/Native/buildTransitive/Sentry.Native.targets
```diff
@@ -26,8 +26,14 @@
...
+    <FrameworkSupportsNative Condition="'$(RuntimeIdentifier)' == 'win-x64' or '$(RuntimeIdentifier)' == 'win-arm64'">true</FrameworkSupportsNative>
+    <!-- Linux -->
+    <FrameworkSupportsNative Condition="'$(RuntimeIdentifier)' == 'linux-x64' or '$(RuntimeIdentifier)' == 'linux-arm64' or '$(RuntimeIdentifier)' == 'linux-musl-x64'">true</FrameworkSupportsNative>
+    <!-- macOS -->
+    <FrameworkSupportsNative Condition="'$(RuntimeIdentifier)' == 'osx-x64' or '$(RuntimeIdentifier)' == 'osx-arm64'">true</FrameworkSupportsNative>
     <!-- net8.0 or greater -->
-    <FrameworkSupportsNative Condition="'$(_SentryIsNet8OrGreater)' == 'true' and ('$(OutputType)' == 'Exe' or '$(OutputType)' == 'WinExe')">true</FrameworkSupportsNative>
+    <FrameworkSupportsNative Condition="'$(_SentryIsNet8OrGreater)' != 'true' or !('$(OutputType)' == 'Exe' or '$(OutputType)' == 'WinExe')">false</FrameworkSupportsNative>
     <!-- Make it opt-out -->
     <FrameworkSupportsNative Condition="'$(SentryNative)' == 'false' or '$(SentryNative)' == 'disable'">false</FrameworkSupportsNative>
```
Flash0ver, line 29: question: Could/should we also allow `<SentryNative>true</SentryNative>` and `<SentryNative>enable</SentryNative>`?

I wonder if explicitly setting `SentryNative` to `true` or `enable` in the consuming project
should set `FrameworkSupportsNative` to `true` unconditionally.

Well ... I guess I answered my own question already, since `FrameworkSupportsNative=true` may then be a _lie_ considering the unsupported OS/Runtimes.

So that thought of "completeness" could allow the user to fall into a problem.
We could reconsider adding this if there's a user ask, should we have problems detecting whether Native is supported in some scenarios, to allow the user an escape hatch of our detection in both directions.

What do you think?

jpnurmi: Thanks for the suggestion. Makes perfect sense. The heuristics above should provide a sane default, but in the end, an explicit value from the user should win.

integration-test/aot.Tests.ps1
jpnurmi: Oops, good catch. That's the way it was intended to be! :smile:

src/Sentry/Platforms/Native/SentryNative.cs
jpnurmi: I figured whitelisting is the way to go because we have a clear list of supported RIDs, and it's easy to add items if we add more platforms to the mix.

PR #4349, opened 2025-07-15
ci: specify RID & .slnf in the matrix

jpnurmi:
[.NET RID](https://learn.microsoft.com/en-us/dotnet/core/rid-catalog) offers a nice, clean, and concise description for the CI jobs; `linux-musl-x64` is easier to read and understand than `ghcr.io/getsentry/sentry-dotnet-alpine:3.21`. At the same time, it provides a future-proof and unique cache key for identifying sentry-native builds.

| Before | After |
|---|---|
|  | |

Also, now that we don't need to compose cache keys from the ambiguous target names ("Linux-arm64" vs. "Linux-musl"), we can pass the desired solution filter name as is, for easier readability.

Note: This might look like unnecessary noise at first, but this makes it easier to maintain the CI because one doesn't need to invent creative ways to conditionally compose cache keys and solution filter names out of os/arch/target when adding new platforms and builds... :wink: 

#skip-changelog

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -11,23 +11,29 @@ on:
 
 jobs:
   build-sentry-native:
-    name: sentry-native (${{ matrix.container.image || matrix.os }})
+    name: sentry-native (${{ matrix.rid }})
     runs-on: ${{ matrix.os }}
-    container: ${{ matrix.container.image }}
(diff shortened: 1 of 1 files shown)


jamescrosswell (approved):
Nice!

PR #4357, opened 2025-07-16
feat: Added `internal bool CaptureAttachment` to `Hub`

bitsandfoxes:
It's possible to send attachments to Sentry on their own and still have them associated with an event. See [dev docs](https://arc.net/l/quote/vtftdfsx).

This is something [the Unity SDK would make use of](https://github.com/getsentry/sentry-unity/pull/2240). Since screenshot capture is not possible at the time the SDK captures an event, screenshot capture itself is delayed and the attachment is then sent separately.

We keep this as `internal` on the hub as this should not be user-facing.

#skip-changelog

### DIFF
--- a/src/Sentry/IHub.cs
+++ b/src/Sentry/IHub.cs
@@ -124,12 +124,4 @@ public interface IHub : ISentryClient, ISentryScopeManager
     /// <param name="configureScope">Callback method to configure the scope.</param>
     /// <param name="hint">An optional hint providing high level context for the source of the event, including attachments</param>
     public void CaptureFeedback(SentryFeedback feedback, Action<Scope> configureScope, SentryHint? hint = null);
-
-    /// <summary>
-    /// Captures an attachment for an existing event.
-    /// </summary>
-    /// <param name="eventId">The ID of the event to add the attachment to.</param>
-    /// <param name="attachment">The attachment.</param>
-    /// <returns>true if the enveloped was queued, false otherwise.</returns>
-    public bool CaptureAttachment(SentryId eventId, SentryAttachment attachment);
 }
--- a/src/Sentry/IHub.cs
+++ b/src/Sentry/IHub.cs
@@ -124,4 +124,12 @@ public interface IHub : ISentryClient, ISentryScopeManager
     /// <param name="configureScope">Callback method to configure the scope.</param>
     /// <param name="hint">An optional hint providing high level context for the source of the event, including attachments</param>
     public void CaptureFeedback(SentryFeedback feedback, Action<Scope> configureScope, SentryHint? hint = null);
+
+    /// <summary>
+    /// Captures an attachment for an existing event.
+    /// </summary>
+    /// <param name="eventId">The ID of the event to add the attachment to.</param>
+    /// <param name="attachment">The attachment.</param>
+    /// <returns>true if the enveloped was queued, false otherwise.</returns>
+    public bool CaptureAttachment(SentryId eventId, SentryAttachment attachment);
 }
--- a/src/Sentry/Internal/Hub.cs
+++ b/src/Sentry/Internal/Hub.cs
@@ -757,6 +757,25 @@ internal class Hub : IHub, IDisposable
         return SentryId.Empty;
     }
 
+    internal bool CaptureAttachment(SentryId eventId, SentryAttachment attachment)
+    {
+        if (!IsEnabled || eventId == SentryId.Empty)
+        {
+            return false;
+        }
+
+        try
+        {
+            var envelope = Envelope.FromAttachment(eventId, attachment, _options.DiagnosticLogger);
+            return CaptureEnvelope(envelope);
+        }
+        catch (Exception e)
+        {
+            _options.LogError(e, "Failure to capture attachment");
+            return false;
+        }
+    }
+
     public async Task FlushAsync(TimeSpan timeout)
     {
         try
@@ -815,25 +834,6 @@ internal class Hub : IHub, IDisposable
             SentrySdk.CloseNativeSdk();
         }
 #endif
-    }
-
-    public bool CaptureAttachment(SentryId eventId, SentryAttachment attachment)
-    {
-        if (!IsEnabled)
-        {
-            return false;
-        }
-
-        try
-        {
-            var envelope = Envelope.FromAttachment(eventId, attachment, _options.DiagnosticLogger);
-            return CaptureEnvelope(envelope);
-        }
-        catch (Exception e)
-        {
-            _options.LogError(e, "Failure to capture attachment");
-            return false;
-        }
     }
 
     public SentryId LastEventId => CurrentScope.LastEventId;
--- a/src/Sentry/Internal/Hub.cs
+++ b/src/Sentry/Internal/Hub.cs
@@ -757,6 +757,8 @@ internal class Hub : IHub, IDisposable
         return SentryId.Empty;
     }
 
+    // Internal capture method that allows the Unity SDK to send attachments after an already captured event.
+    // The preferred way of adding attachments is either on the scope or directly on the event.
     internal bool CaptureAttachment(SentryId eventId, SentryAttachment attachment)
     {
         if (!IsEnabled || eventId == SentryId.Empty)
--- a/src/Sentry/Internal/Hub.cs
+++ b/src/Sentry/Internal/Hub.cs
@@ -817,5 +817,24 @@ internal class Hub : IHub, IDisposable
 #endif
     }
 
+    public bool CaptureAttachment(SentryId eventId, SentryAttachment attachment)
+    {
+        if (!IsEnabled)
+        {
+            return false;
+        }
+
+        try
+        {
+            var envelope = Envelope.FromAttachment(eventId, attachment, _options.DiagnosticLogger);
+            return CaptureEnvelope(envelope);
+        }
+        catch (Exception e)
+        {
+            _options.LogError(e, "Failure to capture attachment");
+            return false;
+        }
+    }
+
     public SentryId LastEventId => CurrentScope.LastEventId;
 }
--- a/src/Sentry/Internal/Hub.cs
+++ b/src/Sentry/Internal/Hub.cs
@@ -759,7 +759,8 @@ internal class Hub : IHub, IDisposable
     }
 
     // Internal capture method that allows the Unity SDK to send attachments after an already captured event.
-    // The preferred way of adding attachments is either on the scope or directly on the event.
+    // Kept internal as the preferred way of adding attachments is either on the scope or directly on the event.
+    // See https://develop.sentry.dev/sdk/data-model/envelope-items/#attachment
     internal bool CaptureAttachment(SentryId eventId, SentryAttachment attachment)
     {
         if (!IsEnabled || eventId == SentryId.Empty || attachment.IsNull())
--- a/src/Sentry/Internal/Hub.cs
+++ b/src/Sentry/Internal/Hub.cs
@@ -1,6 +1,7 @@
 using Sentry.Extensibility;
 using Sentry.Infrastructure;
 using Sentry.Integrations;
+using Sentry.Internal.Extensions;
 using Sentry.Protocol.Envelopes;
 using Sentry.Protocol.Metrics;
 
@@ -761,7 +762,7 @@ internal class Hub : IHub, IDisposable
     // The preferred way of adding attachments is either on the scope or directly on the event.
     internal bool CaptureAttachment(SentryId eventId, SentryAttachment attachment)
     {
-        if (!IsEnabled || eventId == SentryId.Empty)
+        if (!IsEnabled || eventId == SentryId.Empty || attachment.IsNull())
         {
             return false;
         }
--- a/src/Sentry/Platforms/Native/Sentry.Native.targets
+++ b/src/Sentry/Platforms/Native/Sentry.Native.targets
@@ -85,8 +85,9 @@
   <!-- Build the Sentry Native SDK (this only runs on local machines because in CI we expect the SDK to be
        built already on each native platform and fetched for the final .net build. -->
   <Target Name="_BuildSentryNativeSDK"
-    BeforeTargets="DispatchToInnerBuilds"
+    BeforeTargets="DispatchToInnerBuilds;BeforeBuild"
     Condition="'$(_SentryIsNet8OrGreater)' == 'true' and '$(CI)' != 'true'"
+    Inputs="$(SentryNativeBuildInputs)"
     Outputs="$(SentryNativeBuildOutputs)">
     <!-- We want a "-Clean" because if the build script changes, previous cmake cache may contain invalid defines. -->
     <Exec Command="pwsh $(SentryNativeBuildScript) -Clean" />
--- a/src/Sentry/Platforms/Native/Sentry.Native.targets
+++ b/src/Sentry/Platforms/Native/Sentry.Native.targets
@@ -1,10 +1,10 @@
 <Project>
 
   <PropertyGroup>
-    <SentryNativeSourceDirectory>..\..\modules\sentry-native\</SentryNativeSourceDirectory>
+    <SentryNativeSourceDirectory>$(MSBuildThisFileDirectory)..\..\..\..\modules\sentry-native\</SentryNativeSourceDirectory>
     <SentryNativeLibraryName>sentry-native</SentryNativeLibraryName>
-    <SentryNativeBuildScript>../../scripts/build-sentry-native.ps1</SentryNativeBuildScript>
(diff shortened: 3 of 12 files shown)


src/Sentry/IHub.cs
```diff
@@ -124,4 +124,12 @@ public TransactionContext ContinueTrace(
...
     /// <param name="hint">An optional hint providing high level context for the source of the event, including attachments</param>
     public void CaptureFeedback(SentryFeedback feedback, Action<Scope> configureScope, SentryHint? hint = null);
+
+    /// <summary>
+    /// Captures an attachment for an existing event.
+    /// </summary>
+    /// <param name="eventId">The ID of the event to add the attachment to.</param>
+    /// <param name="attachment">The attachment.</param>
+    /// <returns>true if the enveloped was queued, false otherwise.</returns>
+    public bool CaptureAttachment(SentryId eventId, SentryAttachment attachment);
```
bitsandfoxes: How do we deal with this? An addition to a public interface, but not added to `SentrySdk`. I don't even know what kind of changelog entry this should get. 😅 

Of course we could just make it public and then it's a feature?

jamescrosswell: We have tests that explicitly check `SentrySdk` implements everything from `IHub`... so if it has to go on IHub then it's public basically. 

The other option would be to have it on the Hub only (as an internal member) and in the calling code you do something like:
```csharp
var hub = GetSomeIHubImplementationFromSomewhere();
if (hub is Hub realHub)
{
    hub.CaptureAttachment(foo, bar);
}
```

Flash0ver: Since in `SentrySdk` we don't check the runtime type of the `CurrentHub`, but mostly just forward to the `CurrentHub`, I prefer the 2nd option to offer this method `internal` on `Hub`,
with an optional `internal` extension method on `HubExtensions` that does the `if (hub is Hub realHub)` check, returning `false` if it is not.

I'd like to keep this API `internal`, because it's a bit "dangerous": what if there was not or never will be an Event with the specified `SentryId` sent? ... at least for now, until there is an external user ask for it.

src/Sentry/Internal/Hub.cs
```diff
@@ -757,6 +757,25 @@ public SentryId CaptureCheckIn(
         return SentryId.Empty;
     }
 
+    internal bool CaptureAttachment(SentryId eventId, SentryAttachment attachment)
```
bruno-garcia, line 764: // Internal because e e give preference to attacking files directly to the event, using the scope (global or local to the capture function call).  ur we could consider making this public if more use cases come up.

src/Sentry/Platforms/Native/Sentry.Native.targets
```diff
@@ -1,10 +1,10 @@
 <Project>
 
   <PropertyGroup>
-    <SentryNativeSourceDirectory>..\..\modules\sentry-native\</SentryNativeSourceDirectory>
+    <SentryNativeSourceDirectory>$(MSBuildStartupDirectory)\modules\sentry-native\</SentryNativeSourceDirectory>
     <SentryNativeLibraryName>sentry-native</SentryNativeLibraryName>
-    <SentryNativeBuildScript>../../scripts/build-sentry-native.ps1</SentryNativeBuildScript>
-    <SentryNativeBuildInputs>../../.git/modules/modules/sentry-native/HEAD;$(MSBuildThisFileDirectory)Sentry.Native.targets;$(SentryNativeBuildScript)</SentryNativeBuildInputs>
+    <SentryNativeBuildScript>$(MSBuildStartupDirectory)/scripts/build-sentry-native.ps1</SentryNativeBuildScript>
```
bruno-garcia: how does this resolve if we call build form different directories
isn't it better to use $(project root)and then ../..?

bitsandfoxes: I mixed up on which branch I was. See https://github.com/getsentry/sentry-dotnet/pull/4366/files

src/Sentry/Internal/Hub.cs
Flash0ver: follow-up question:

Shall we add a link to the docs here as well?
https://develop.sentry.dev/sdk/data-model/envelope-items/#attachment

src/Sentry/Internal/Hub.cs
```diff
@@ -757,6 +757,27 @@ public SentryId CaptureCheckIn(
         return SentryId.Empty;
     }
 
+    // Internal capture method that allows the Unity SDK to send attachments after an already captured event.
+    // The preferred way of adding attachments is either on the scope or directly on the event.
+    internal bool CaptureAttachment(SentryId eventId, SentryAttachment attachment)
```
Flash0ver, line 764: question: modifying `internal` APIs

Due to `<InternalsVisibleTo Include="Sentry.Unity" />` we may change something in `sentry-dotnet` that will break `sentry-unity`.

I'm wondering how careful we should be in general about changing `internal` types and members,
since we may notice the break of `sentry-unity` only when we update to the latest `sentry-dotnet`.

Or, @bitsandfoxes, do you keep an eye out for potential issues we may cause with `internal` changes.
Or, do you not mind too much?

src/Sentry/Protocol/Envelopes/Envelope.cs
```diff
@@ -445,6 +445,24 @@ internal static Envelope FromClientReport(ClientReport clientReport)
...
+    /// <summary>
+    /// Creates an envelope that contains only an attachment for an existing event.
+    /// </summary>
+    internal static Envelope FromAttachment(SentryId eventId, SentryAttachment attachment, IDiagnosticLogger? logger = null)
+    {
+        var header = CreateHeader(eventId);
+        var items = new List<EnvelopeItem>();
+
+        // Safety check, in case the user forcefully added a null attachment.
+        if (attachment.IsNull())
```
Flash0ver: question: null-check required?

This is an `internal` API, and `sentry-unity` (at least during build-time) does use nullable reference types (or at least `#nullable enable warnings`), right? So we should get warnings/errors if possibly null from the caller side.

Well ... on second thought ... should we end up making this API `public` in the future, we could potentially forget adding that check.
And also, if this method is called by an API that is `public` in`sentry-unity`, and user code in a game now calls that API where they don't have `<Nullable>enable</Nullable>`, we actually may run into a null-reference here. Perhaps we actually *should* keep the null-check.

src/Sentry/Protocol/Envelopes/Envelope.cs
```diff
@@ -445,6 +445,24 @@ internal static Envelope FromClientReport(ClientReport clientReport)
...
+        {
+            logger?.LogWarning("Encountered a null attachment.  Skipping.");
```
Flash0ver: issue (non-blocking): shall we continue when `attachment is null`?

Should the `attachment` be `null`, we still continue with this method and invoke `AddEnvelopeItemFromAttachment`, which will throw a `NullReferenceException`, immediately catching it and trying to log an Error, which does throw again another `NullReferenceException` which is then caught by the other new method on `Hub`, eventually logging an Error.

In this case, the behavior is
- when `SentryAttachment` is `null`
- log: `logger?.LogWarning("Encountered a null attachment. Skipping.");`
- throw and catch `NullReferenceException` in `AddEnvelopeItemFromAttachment`
- fail: `logger?.LogError(exception, "Failed to add attachment: {0}.", attachment.FileName);`
  - throws another `NullReferenceException` because `attachment` is still `null`
  - but only if the `logger is not null` ... i.e. when `SentryOptions.Debug = true`
- catching the exception when called from `Hub.CaptureAttachment`
- log: `_options.LogError(e, "Failure to capture attachment");`

I personally find the behavior a bit confusing when we do pass `null`,
since the null-check via `if (attachment.IsNull())` only logs an additional warning,
but does not actually influence the control flow.

I can think of three options to maybe make this behavior more transparent:
1. either don't do `if (attachment.IsNull())` because, although logging a Warning, still a `NullReferenceException` is thrown out of this method
   - and, since it's an `internal` API, we could assume the caller has `<Nullable>enable</Nullable>`
2. return a `null` `Envelope` in case that `SentryAttachment` is `null`
   - would change the nullable annotation of the return type
3. move the _null-check_ from here to the proposed `Hub.CaptureAttachment`
   - where we can return `false`
   - and may or may not move the log message over there as well

I personally prefer a combination of option 1 and 3:
Move the null-check from here to the proposed `Hub.CaptureAttachment`
```diff
-if (!IsEnabled || eventId == SentryId.Empty)
+if (!IsEnabled || eventId == SentryId.Empty || attachment is null)
 {
   return false;
 }
..
```
This would also be consistent with this type,
because other `internal` methods in `Sentry.Protocol.Envelopes.Envelope` will throw `NullReferenceException`s when `null`s are passed.

src/Sentry/Protocol/Envelopes/Envelope.cs
```diff
@@ -445,6 +445,24 @@ internal static Envelope FromClientReport(ClientReport clientReport)
...
+        }
+
+        AddEnvelopeItemFromAttachment(items, attachment, logger);
```
Flash0ver: suggestion: use `EnvelopeItem.FromAttachment(SentryAttachment)` instead

Instead of using the method `AddEnvelopeItemFromAttachment` to create List and add just a single item,
we could use `EnvelopeItem.FromAttachment` instead, and pass the `EnvelopeItem` to the `Envelope` _.ctor_, like
```CSharp
return new Envelope(eventId, header, [item]);
```

Perhaps this simplifies the method even further.

test/Sentry.Tests/HubTests.cs
```diff
@@ -1730,6 +1730,54 @@ public void CaptureFeedback_ConfigureScope_ScopeApplied(bool enabled)
...
+        {
+            hub.Dispose();
+        }
+
+        _fixture.Client.CaptureEnvelope(Arg.Any<Envelope>()).Returns(true);
+
+        var eventId = SentryId.Create();
+        var attachment = new SentryAttachment(
+            AttachmentType.Default,
+            new StreamAttachmentContent(new MemoryStream(Encoding.UTF8.GetBytes("test content"))),
```
Flash0ver: thought: use `ByteAttachmentContent` or a new `Sentry.Testing.NullAttachmentContent` instead

I wonder if we could simplify the `//Arrange` code a bit (and make the test a bit faster by avoid a couple of heap allocations),
by using `new ByteAttachmentContent(Array.Empty<byte>())` instead of the `StreamAttachmentContent` (allocating both a new `byte`-Array and a `MemoryStream`.

Alternatively, we perhaps could introduce a new type for testing, in `Sentry.Testing`:
```CSharp
namespace Sentry.Testing;

public sealed class NullAttachmentContent : IAttachmentContent
{
    public static NullAttachmentContent Instance { get; } = new NullAttachmentContent();

    private NullAttachmentContent()
    {
    }

    public Stream GetStream() => Stream.Null;
}
```

which would make the `//Arrange` even simpler,
since we may get away with not requiring a non-Empty-Stream for this test case.

This new type could also be used in the `EnvelopeTests.FromAttachment_NullAttachment_DoesNotIncludeAttachment` test.

test/Sentry.Tests/HubTests.cs
```diff
@@ -1730,6 +1730,54 @@ public void CaptureFeedback_ConfigureScope_ScopeApplied(bool enabled)
...
+    public void CaptureAttachment_HubEnabled(bool enabled)
+    {
+        // Arrange
+        var hub = _fixture.GetSut();
+        if (!enabled)
+        {
+            hub.Dispose();
+        }
+
+        _fixture.Client.CaptureEnvelope(Arg.Any<Envelope>()).Returns(true);
```
Flash0ver, line 1745: thought: could/should we Assert that `CaptureEnvelope` actually was called in case the Hub is enabled, and was _not_ called in the case the Hub is disabled?

test/Sentry.Tests/Protocol/Envelopes/EnvelopeTests.cs
```diff
@@ -1090,4 +1090,64 @@ public async Task Serialization_RoundTrip_RecalculatesLengthHeader()
...
+            new StreamAttachmentContent(new MemoryStream(Encoding.UTF8.GetBytes("test content"))),
+            "test.txt",
+            "text/plain");
+
+        // Act
+        using var envelope = Envelope.FromAttachment(eventId, attachment);
+
+        // Assert
+        envelope.TryGetEventId().Should().Be(eventId);
+        envelope.Items.Should().HaveCount(1);
```
Flash0ver, line 1110: thought: could/should we also Assert the `Envelope.Header`s, similar to the `EnvelopeItem.Header`s?

test/Sentry.Tests/Protocol/Envelopes/EnvelopeTests.cs
```diff
@@ -1090,4 +1090,64 @@ public async Task Serialization_RoundTrip_RecalculatesLengthHeader()
...
     }
+
+    [Fact]
+    public void FromAttachment_ValidAttachment_CreatesEnvelope()
+    {
+        // Arrange
+        var eventId = SentryId.Create();
+        var attachment = new SentryAttachment(
+            AttachmentType.Default,
+            new StreamAttachmentContent(new MemoryStream(Encoding.UTF8.GetBytes("test content"))),
```
Flash0ver: suggestion: we could shorten this a little by using the `ByteAttachmentContent`, which creates a `new MemoryStream(_bytes)` in it's implementation

```suggestion
            new ByteAttachmentContent(Encoding.UTF8.GetBytes("test content")),
```

src/Sentry/Internal/Hub.cs
bitsandfoxes: I'm keeping an eye out and from the Unity POV I don't mind. Since the SDK is getting consumed as a submodule and built from source it's hard to imagine a really blocking situation.

src/Sentry/Protocol/Envelopes/Envelope.cs
bitsandfoxes: Ha, nice catch! I pulled the null-check into the hub as suggested.

src/Sentry/Internal/Hub.cs
Flash0ver: 👌
should we run into a problem of `sentry-dotnet` pulling the `internal` rug under `sentry-unity`'s feet all the time, we could think about adding a small Roslyn-Tool that lists all `internal` types and members used by `Sentry.Unity.csproj` for a quick index lookup to check in ahead of time when we're proposing an internally-breaking-change

test/Sentry.Tests/Protocol/Envelopes/EnvelopeTests.cs
bitsandfoxes: That is done implicitly through `envelope.TryGetEventId().Should().Be(eventId);`, right? What else should the test check on the `Envelope.Header` other than the `SentryId`?

src/Sentry/Protocol/Envelopes/Envelope.cs
bitsandfoxes: I completely missed the `EnvelopeItem.FromAttachment` method!

src/Sentry/Protocol/Envelopes/Envelope.cs
bitsandfoxes: I've moved the check out into the hub.

test/Sentry.Tests/Protocol/Envelopes/EnvelopeTests.cs
Flash0ver: Ah ... `Envelope._eventId` read from `Envelope.TryGetEventId()` is cached from the `event_id` Header.

test/Sentry.Testing/NullAttachmentContent.cs
```diff
@@ -0,0 +1,8 @@
+namespace Sentry.Testing;
+
+internal sealed class NullAttachmentContent : IAttachmentContent
```
Flash0ver, line 3: nitpick: we could make the `.ctor` `private`, so that we can only use the singleton `Instance`

test/Sentry.Tests/HubTests.cs
```diff
@@ -1730,6 +1730,67 @@ public void CaptureFeedback_ConfigureScope_ScopeApplied(bool enabled)
...
+
+        var eventId = SentryId.Create();
+        var attachment = new SentryAttachment(
+            AttachmentType.Default,
+            new ByteAttachmentContent("test content"u8.ToArray()),
```
Flash0ver, line 1750: question: could/should we add an `internal` _constructor_, that takes a `ReadOnlySpan<byte>`, so that we can omit the `.ToArray()` at the call sites?

```CSharp
internal ByteAttachmentContent(ReadOnlySpan<byte> bytes) => _bytes = bytes.ToArray();
```

Or is that overkill, because we only have 2 usages that would be shortened.

Flash0ver (approved):
thanks! ... I posted 2 more non-blocking comments

Issue #4355, opened 2025-07-16
fix(ci): MSBuild binlogs are not uploaded as artifact when build fails


Flash0ver:
### Description

While investigating #4091, we noticed that the `Upload build logs` step is `skipped` for the failing build.

See
https://github.com/getsentry/sentry-dotnet/blob/baa279b61f755f7f9110b413203d9497b3d07ae3/.github/workflows/build.yml#L170-L175

jpnurmi:
Would `if: ${{ always() }}` help?
```diff
  - name: Upload build logs 
-   if: ${{ steps.build.outcome != 'skipped' }} 
+   if: ${{ always() }}
    uses: actions/upload-artifact@v4 
    with: 
      name: ${{ matrix.target || runner.os }}-build-logs 
      path: build.binlog 
+     if-no-files-found: ignore # optional - a warning might be fine?
```

References:
- https://docs.github.com/en/actions/reference/evaluate-expressions-in-workflows-and-actions#always
- https://github.com/actions/upload-artifact?tab=readme-ov-file#usage
    ```yaml
        # The desired behavior if no files are found using the provided path.
        # Available Options:
        #   warn: Output a warning but do not fail the action
        #   error: Fail the action with an error message
        #   ignore: Do not output any warnings or errors, the action does not fail
        # Optional. Default is 'warn'
        if-no-files-found:
    ```

jamescrosswell:
> Would `if: ${{ always() }}` help?

The only scenario where it would be problematic is if the build itself hadn't yet run... so if there was an error restoring the .NET dependencies or something. That was the idea behind !skipped.

Possibly we could have two blocks:

```yaml
- name: Upload build logs (success)
  if: ${{ steps.build.outcome == 'success' }}
  uses: actions/upload-artifact@v4
  with:
    name: ${{ matrix.target || runner.os }}-build-logs
    path: build.binlog

- name: Upload build logs (failure)
  if: ${{ failure() && steps.build.outcome == 'failure' }}
  uses: actions/upload-artifact@v4
  with:
    name: ${{ matrix.target || runner.os }}-build-logs
    path: build.binlog
```

PR #4358, opened 2025-07-16
ci: always upload build logs

jpnurmi:
Example run: https://github.com/getsentry/sentry-dotnet/actions/runs/16322534643/job/46104110633

Resolves: #4355

#skip-changelog

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -182,7 +182,7 @@ jobs:
         run: dotnet build ${{ matrix.slnf }} -c Release --no-restore --nologo -v:minimal -flp:logfile=build.log -p:CopyLocalLockFileAssemblies=true -bl:build.binlog
 
       - name: Upload build logs
-        if: ${{ steps.build.outcome != 'skipped' }}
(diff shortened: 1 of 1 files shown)


jamescrosswell (commented):
Nice, thanks @jpnurmi 

I left a [comment on the issue](https://github.com/getsentry/sentry-dotnet/issues/4355#issuecomment-3081284324)... So basically I'm wondering what happens if the logs don't exist because the run fails before getting to the build step. Is that a problem or will it handle that gracefully?

jpnurmi:
This is how it behaves with the different [if-no-files-found](https://github.com/actions/upload-artifact?tab=readme-ov-file#inputs) values:

- `if-no-files-found: ignore`

   image

- `if-no-files-found: warn` (default)

   image

- `if-no-files-found: error`

   image

Either `ignore` or `warn` works for this use case. I think it's a matter of taste. `ignore` outputs a quiet message in the step, while `warn` produces an annotation in the job summary. Which do you like more? :)

jamescrosswell (approved):
Nice!

PR #4360, opened 2025-07-17
Improve the caching transport tests

jamescrosswell:
The caching transport tests can fail when running the device tests if the simulator looses network connectivity. This PR changes the test to use a `FakeReliableNetworkStatusListener ` implementation of `INetworkStatusListener` that ignores the real network status and just pretends that everything is OK, so that test reliability isn't impacted by network connectivity.

Extracted from https://github.com/getsentry/sentry-dotnet/pull/4294

#skip-changelog

### DIFF
--- a/test/Sentry.Testing/FakeReliableNetworkStatusListener.cs
+++ b/test/Sentry.Testing/FakeReliableNetworkStatusListener.cs
@@ -1,0 +1,20 @@
+namespace Sentry.Testing;
+
+/// <summary>
+/// To be used when we don't want the real network status to affect the reliability of our tests
+/// </summary>
+public class FakeReliableNetworkStatusListener : INetworkStatusListener
+{
+    public static readonly FakeReliableNetworkStatusListener Instance = new();
+
+    private FakeReliableNetworkStatusListener()
+    {
+    }
+
+    public bool Online => true;
+
+    public Task WaitForNetworkOnlineAsync(CancellationToken cancellationToken)
+    {
+        return Task.CompletedTask;
+    }
+}
--- a/test/Sentry.Tests/Internals/Http/CachingTransportTests.cs
+++ b/test/Sentry.Tests/Internals/Http/CachingTransportTests.cs
@@ -3,31 +3,38 @@ using Sentry.Internal.Http;
 
 namespace Sentry.Tests.Internals.Http;
 
-public class CachingTransportTests
+public class CachingTransportTests : IDisposable
 {
     private readonly TestOutputDiagnosticLogger _logger;
+    private readonly TempDirectory _cacheDirectory;
+    private readonly SentryOptions _options;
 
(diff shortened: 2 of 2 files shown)


test/Sentry.Testing/FakeReliableNetworkStatusListener.cs
```diff
@@ -0,0 +1,20 @@
+namespace Sentry.Testing;
+
+/// <summary>
+/// To be used when we don't want the real network status to affect the reliability of our tests
+/// </summary>
+public class FakeReliableNetworkStatusListener : INetworkStatusListener
```
Flash0ver, line 6: nitpick: could be `sealed`

test/Sentry.Tests/Internals/Http/CachingTransportTests.cs
```diff
@@ -222,46 +192,27 @@ public async Task EnvelopeReachesInnerTransport()
...
     {
         // Arrange
-        using var cacheDirectory = new TempDirectory();
-        var options = new SentryOptions
-        {
-            Dsn = ValidDsn,
-            DiagnosticLogger = _logger,
-            Debug = true,
-            CacheDirectoryPath = cacheDirectory.Path,
-            MaxCacheItems = 2
```
Flash0ver, line 232: suggestion: set `MaxCacheItems = 2`

This test no longer sets `MaxCacheItems = 2`, but uses the default of (currently) `30`.

Shall we add
```diff
+_options.MaxCacheItems = 2;
```
again, as the first line in the `//Arrange` of this test, to keep the previous behavior of the test?

I suppose this is only about performance to reduce the number of iterations in this test.

test/Sentry.Tests/Internals/Http/CachingTransportTests.cs
```diff
@@ -3,31 +3,38 @@
...
-    }
-
-    [Fact]
-    public async Task WithAttachment()
-    {
-        // Arrange
-        using var cacheDirectory = new TempDirectory();
-        var options = new SentryOptions
+        _cacheDirectory = new TempDirectory();
+        _options = new SentryOptions
```
Flash0ver, line 16: praise: nice de-duplication and cleanup

test/Sentry.Testing/FakeReliableNetworkStatusListener.cs
jamescrosswell: If it was in the public API, we could worry about that. Since it's just in our test assemblies (and we're the only ones who use those)... probably not worth fussing over.

test/Sentry.Tests/Internals/Http/CachingTransportTests.cs
jamescrosswell: Nice catch - thank you!

PR #4361, opened 2025-07-17
Bump to Google API levels 34 and 36 for device tests

jamescrosswell:
Extracted from https://github.com/getsentry/sentry-dotnet/pull/4294

#skip-changelog

### DIFF
--- a/.github/workflows/device-tests-android.yml
+++ b/.github/workflows/device-tests-android.yml
@@ -58,8 +58,9 @@ jobs:
       fail-fast: false
       matrix:
         tfm: [net9.0]
-        # We run against both an older and a newer API
-        api-level: [27, 33]
+        # Must be 34+ for new apps and app updates as of August 31, 2024.
+        # See https://apilevels.com/
+        api-level: [34, 36]
     env:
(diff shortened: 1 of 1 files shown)


Flash0ver (approved):

.github/workflows/device-tests-android.yml
```diff
@@ -58,8 +58,9 @@ jobs:
       fail-fast: false
       matrix:
         tfm: [net9.0]
-        # We run against both an older and a newer API
-        api-level: [27, 33]
+        # Must be 34+ for new apps and app updates as of August 31, 2024.
+        # See https://apilevels.com/
+        api-level: [34, 36]
```
Flash0ver, line 63: question: simplify or keep (line: 95)

Do we still need the condition on line: 95
```YAML
target: ${{ matrix.api-level >= 32 && 'google_apis' || 'default' }}
```

or is it nice to keep `matrix.api-level >= 32`,
although we're targeting `34` and `36`,
so that the absence of `default` is documented

(I do like it as is)

jamescrosswell: True... I'll leave the comment but remove the conditional

PR #4362, opened 2025-07-17
Added SimpleStackTraceFactory

### LINKED ISSUE #3439: Stack Trace missing for NativeAOT App
### Package

Sentry

### .NET Flavor

.NET Core

### .NET Version

8.0

### OS

Linux

### SDK Version

4.7.0

24.6.0

### Steps to Reproduce

I'm having an issue where the client is not exporting stack traces to the API for a NativeAOT compiled app. Works fine in a non-NativeAOT app.

### Expected Result

Event is sent with stack traces.

### Actual Result

Event is sent with no stack traces.

jamescrosswell:
Resolves #3439
- https://github.com/getsentry/sentry-dotnet/issues/3439

### DIFF
--- a/src/Sentry/Extensibility/SimpleStackTraceFactory.cs
+++ b/src/Sentry/Extensibility/SimpleStackTraceFactory.cs
@@ -1,0 +1,73 @@
+using System;
+using System.Collections.Generic;
+
+namespace Sentry.Extensibility;
+
+/// <summary>
+/// A rudimentary implementation of <see cref="ISentryStackTraceFactory"/> that simply parses the
+/// string representation of the stack trace from an exception. This lacks many of the features
+/// off the full <see cref="SentryStackTraceFactory"/>, however it may be useful in AOT compiled
+/// applications where the full factory is not returning a useful stack trace.
+/// <remarks>SimpleStackTraceFactory is currently EXPERIMENTAL.</remarks>
+/// </summary>
+public partial class SimpleStackTraceFactory : ISentryStackTraceFactory
+{
+    private readonly SentryOptions _options;
+    private const string StackTraceLinePattern = @"at (.+)\.(.+) \+";
+
+#if NET9_0_OR_GREATER
+    [GeneratedRegex(StackTraceLinePattern)]
+    private static partial Regex StackTraceLine { get; }
+#elif NET8_0
+     private static readonly Regex StackTraceLine = StackTraceLineRegex();
+
+     [GeneratedRegex(StackTraceLinePattern)]
+     private static partial Regex StackTraceLineRegex();
+#else
+    private static readonly Regex StackTraceLine = new (StackTraceLinePattern, RegexOptions.Compiled);
+#endif
+
+    /// <summary>
+    /// Creates a new instance of <see cref="SimpleStackTraceFactory"/>.
+    /// </summary>
+    /// <param name="options">The sentry options</param>
+    public SimpleStackTraceFactory(SentryOptions options)
+    {
+        _options = options;
+    }
+
+    /// <inheritdoc cref="ISentryStackTraceFactory.Create"/>
+    public SentryStackTrace? Create(Exception? exception = null)
+    {
+        _options.LogDebug("Source Stack Trace: {0}", exception?.StackTrace);
+
+        var trace = new SentryStackTrace();
+        var frames = new List<SentryStackFrame>();
+
+        var newlines = new[] { Environment.NewLine };
+        var lines = exception?.StackTrace?.Split(newlines, StringSplitOptions.RemoveEmptyEntries) ?? [];
+        foreach (var line in lines)
+        {
+            var match = StackTraceLine.Match(line);
+            if (match.Success)
+            {
+                frames.Add(new SentryStackFrame()
+                {
+                    Module = match.Groups[1].Value,
+                    Function = match.Groups[2].Value
+                });
+            }
+            else
+            {
+                _options.LogDebug("Regex match failed for: {0}", line);
+                frames.Add(new SentryStackFrame()
+                {
+                    Function = line
+                });
+            }
+        }
+
+        trace.Frames = frames;
+        return trace;
+    }
+}
--- a/src/Sentry/Extensibility/ISentryStackTraceFactory.cs
+++ b/src/Sentry/Extensibility/ISentryStackTraceFactory.cs
@@ -1,7 +1,7 @@
 namespace Sentry.Extensibility;
 
 /// <summary>
-/// Factory to <see cref="SentryStackTrace" /> from an <see cref="Exception" />.
+/// Factory to create a <see cref="SentryStackTrace" /> from an <see cref="Exception" />.
 /// </summary>
 public interface ISentryStackTraceFactory
 {
--- a/src/Sentry/SentryOptions.cs
+++ b/src/Sentry/SentryOptions.cs
@@ -1631,7 +1631,16 @@ public class SentryOptions
         => ExceptionProcessorsProviders.SelectMany(p => p());
 
     /// <summary>
+    /// <para>
     /// Use custom <see cref="ISentryStackTraceFactory" />.
+    /// </para>
+    /// <para>
+    /// By default, Sentry uses the <see cref="SentryStackTraceFactory"/> to create stack traces and this implementation
+    /// offers the most comprehensive functionality. However, full stack traces are not available in AOT compiled
+    /// applications. If you are compiling your applications AOT and the stack traces that you see in Sentry are not
+    /// informative enough, you could consider using the <see cref="SimpleStackTraceFactory" /> instead. This is not as
+    /// functional but is guaranteed to provide at least _something_ useful in AOT compiled applications.
+    /// </para>
     /// </summary>
     /// <param name="sentryStackTraceFactory">The stack trace factory.</param>
     public SentryOptions UseStackTraceFactory(ISentryStackTraceFactory sentryStackTraceFactory)
--- a/test/Sentry.Tests/ApiApprovalTests.Run.DotNet8_0.verified.txt
+++ b/test/Sentry.Tests/ApiApprovalTests.Run.DotNet8_0.verified.txt
@@ -1516,6 +1516,11 @@ namespace Sentry.Extensibility
         public SentryStackTraceFactory(Sentry.SentryOptions options) { }
         public Sentry.SentryStackTrace? Create(System.Exception? exception = null) { }
     }
+    public class SimpleStackTraceFactory : Sentry.Extensibility.ISentryStackTraceFactory
+    {
+        public SimpleStackTraceFactory(Sentry.SentryOptions options) { }
+        public Sentry.SentryStackTrace? Create(System.Exception? exception = null) { }
+    }
 }
 namespace Sentry.Http
 {
--- a/test/Sentry.Tests/ApiApprovalTests.Run.DotNet9_0.verified.txt
+++ b/test/Sentry.Tests/ApiApprovalTests.Run.DotNet9_0.verified.txt
@@ -1516,6 +1516,11 @@ namespace Sentry.Extensibility
         public SentryStackTraceFactory(Sentry.SentryOptions options) { }
         public Sentry.SentryStackTrace? Create(System.Exception? exception = null) { }
     }
+    public class SimpleStackTraceFactory : Sentry.Extensibility.ISentryStackTraceFactory
+    {
+        public SimpleStackTraceFactory(Sentry.SentryOptions options) { }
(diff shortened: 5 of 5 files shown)


src/Sentry/Extensibility/SimpleStackTraceFactory.cs
```diff
@@ -0,0 +1,99 @@
...
+namespace Sentry.Extensibility;
+
+/// <summary>
+/// A rudimentary implementation of <see cref="ISentryStackTraceFactory"/> that simply parses the
+/// string representation of the stack trace from an exception. This lacks many of the features
+/// off the full <see cref="SentryStackTraceFactory"/>, however it may be useful in AOT compiled
+/// applications where the full factory is not returning a useful stack trace.
+/// <remarks>SimpleStackTraceFactory is currently EXPERIMENTAL.</remarks>
+/// </summary>
+public partial class SimpleStackTraceFactory : ISentryStackTraceFactory
```
Flash0ver: suggestion: mark this type `[Experimental(DiagnosticId.ExperimentalFeature)]`

Flash0ver: thought: naming ... no need to consider if we mark the type with the `[ExperimentalAttribute]`

I personally don't like the name `SimpleStackTraceFactory`, since it's relative ... what if we come up with an even simpler implementation in the future?
Well ... to be fair ... we only had a single implementation so far ... yet still.

Maybe `RegexStackTraceFactory` 🤔

src/Sentry/Extensibility/SimpleStackTraceFactory.cs
```diff
@@ -0,0 +1,99 @@
...
+    [GeneratedRegex(FullStackTraceLinePattern)]
+    private static partial Regex FullStackTraceLine { get; }
+#elif NET8_0
+     private static readonly Regex FullStackTraceLine = FullStackTraceLineRegex();
+
+     [GeneratedRegex(FullStackTraceLinePattern)]
+     private static partial Regex FullStackTraceLineRegex();
+#else
+    private static readonly Regex FullStackTraceLine = new (FullStackTraceLinePattern, RegexOptions.Compiled);
+#endif
```
Flash0ver: thought: `RegexOptions.ExplicitCapture`

Since we use explicit capture groups in the pattern for the first Regex,
I wonder if it could help with clarity (and maybe performance) if we add the option `RegexOptions.ExplicitCapture`.

src/Sentry/Extensibility/SimpleStackTraceFactory.cs
```diff
@@ -0,0 +1,99 @@
...
+    /// <param name="options">The sentry options</param>
+    public SimpleStackTraceFactory(SentryOptions options)
+    {
+        _options = options;
+    }
+
+    /// <inheritdoc cref="ISentryStackTraceFactory.Create"/>
+    public SentryStackTrace? Create(Exception? exception = null)
+    {
+        _options.LogDebug("Source Stack Trace: {0}", exception?.StackTrace);
```
Flash0ver: question: Do we need to check for `SentryOptions.AttachStacktrace`?

Or does only our implicit and default `SentryStackTraceFactory` need to do the check, because when a `ISentryStackTraceFactory` is explicitly set via `SentryOptions.UseStackTraceFactory(ISentryStackTraceFactory)` an explicit opt-in is intended.

something like
https://github.com/getsentry/sentry-dotnet/blob/f98922d9a648576e5957c27ae325f938bbe77c50/src/Sentry/Extensibility/SentryStackTraceFactory.cs#L24-L28

src/Sentry/Extensibility/SimpleStackTraceFactory.cs
```diff
@@ -0,0 +1,99 @@
...
+
+        var trace = new SentryStackTrace();
+        var frames = new List<SentryStackFrame>();
+
+        var newlines = new[] { Environment.NewLine };
```
Flash0ver: perf: May be statically cached to avoid repetitive allocations.

Although ... do we even need the array?
Can't we just pass `Environment.NewLine` as the separator?

src/Sentry/Extensibility/SimpleStackTraceFactory.cs
```diff
@@ -0,0 +1,99 @@
...
+
+            _options.LogDebug("Stack frame match failed for: {0}", line);
+            frames.Add(new SentryStackFrame()
+            {
+                Function = line
+            });
+        }
+
+        trace.Frames = frames;
+        return trace;
```
Flash0ver: question: Return `null` when no frames?

Should we return `null` in case either the passed `Exception` or its `Exception.StackTrace` is null,
rather than returning a `SentryStackTrace` wihtout any `SentryStackFrame`?

test/Sentry.Tests/Internals/SimpleStackTraceFactoryTests.cs
```diff
@@ -0,0 +1,42 @@
...
+    [Fact]
+    public Task MethodGeneric()
+    {
+        // Arrange
+        const int i = 5;
+        var exception = Record.Exception(() => GenericMethodThatThrows(i));
+
+        var options = new SentryOptions
+        {
+            AttachStacktrace = true
```
Flash0ver: question: I believe `AttachStacktrace = true` currently has no effect ... should it have any?

test/Directory.Build.props
```diff
@@ -48,6 +48,7 @@
     <Using Include="NSubstitute.ReturnsExtensions" />
     <Using Include="Xunit" />
     <Using Include="Xunit.Abstractions" />
+    <Using Condition="'$(TargetPlatformIdentifier)'==''" Include="VerifyXunit" />
```
Flash0ver, line 51: question: Is this required?

I thought it was redundant, because `Verify.Xunit` already ships with `ImplicitUsings`-support:
https://github.com/VerifyTests/Verify/blob/2f64cd19406244a38c34d8e8b737d936dde5e563/src/Verify.Xunit/buildTransitive/Verify.Xunit.props#L3-L7

Or is there a problem that I don't see?

src/Sentry/Extensibility/SimpleStackTraceFactory.cs
jamescrosswell: It's an interesting question. The `AttachStacktrace` option is generally used when doing things like `SentrySdk.CaptureMessage`... if `SentryOptions.AttachStackTrace == true` then you end up with a stack trace attached to calls like that... otherwise (and by default) you only get a stack trace if there is an exception associated with the event. 

The `SentryStackTraceFactory` creates a stacktrace from an exception by doing `new StackTrace(exception, true)`.

I'm not sure it makes sense to manufacture a stack trace for this new stack trace factory though. In it's implementation it's parsing `exception.StackTrace` (which is a string). People would be using this factory specifically because `new StackTrace(exception, true)` isn't returning any useful information... 

So I think basically the new stack trace factory wouldn't support the `AttachStacktrace` option.

This would be much easier to decide if we could reproduce the problem that this stack trace factory is designed to solve (getting empty stack traces in AOT compiled applications). I haven't been able to do that yet so I'm not sure what would happen if we tried to parse `new StackTrace().ToString()` using this factory. Would `new StackTrace()` return anything useful (even though `new StackTrace(exception, true)` does not)?

Given that I can't test that functionality, I don't think I should be building it then...

test/Sentry.Tests/Internals/SimpleStackTraceFactoryTests.cs
jamescrosswell: See https://github.com/getsentry/sentry-dotnet/pull/4362#discussion_r2224295476

test/Directory.Build.props
jamescrosswell: I got compiler errors without this:

image

I'm not sure why... but adding this fixed it.

src/Sentry/Extensibility/SimpleStackTraceFactory.cs
jamescrosswell: I've made the underlying StackTraceFactory available only for net8.0 and greater... since it's designed for AOT scenarios, I think this makes sense. As such, I removed the regex logic for netstandard and netfx as redundant.

CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 ## Unreleased
 
+### Features
+
+- Sentry now includes an EXPERIMENTAL StringStackTraceFactory. As the name suggests, this factory isn't as feature rich as the full `SentryStackTraceFactory`. However, it may provide better results if you are compiling your application AOT and not getting useful stack traces from the full stack trace factory. ([#4362](https://github.com/getsentry/sentry-dotnet/pull/4362))
```
```suggestion
- Sentry now includes an EXPERIMENTAL StringStackTraceFactory. As the name suggests, this factory isn't as feature rich as the full `SentryStackTraceFactory`. However, it may provide better results if you are compiling your application AOT and not getting useful stack traces from the full stack trace factory. ([#4362](https://github.com/getsentry/sentry-dotnet/pull/4362))

```

src/Sentry/Extensibility/StringStackTraceFactory.cs
```diff
@@ -0,0 +1,106 @@
...
+/// <para>
+/// This factory is designed for AOT scenarios, so only available for net8.0+
+/// </para>
+/// </remarks>
+/// </summary>
+[Experimental(DiagnosticId.ExperimentalFeature)]
+public partial class StringStackTraceFactory : ISentryStackTraceFactory
+{
+    private readonly SentryOptions _options;
+    private const string FullStackTraceLinePattern = @"at (?<Module>[^\.]+)\.(?<Function>.*?) in (?<FileName>.*?):line (?<LineNo>\d+)";
```
bruno-garcia, line 25: Is this regex coming from another file in the repo or it's a new regex? curious how tested/vetted this is

jamescrosswell: It's a new regex and the short answer is "not very"... which is one reason this is marked as Experimental.

We do have [these unit tests](https://github.com/getsentry/sentry-dotnet/blob/45b3258c855495f451fa21bea9ddb461c6812f5b/test/Sentry.Tests/Internals/StringStackTraceFactoryTests.cs#L41-L64) but I'm not sure what this string is likely to look like in all the different real world scenarios.

This one (the fallback) is the regex that @filipnavara  is using [here](https://github.com/getsentry/sentry-dotnet/issues/3439#issuecomment-3077738277):
https://github.com/getsentry/sentry-dotnet/blob/45b3258c855495f451fa21bea9ddb461c6812f5b/src/Sentry/Extensibility/StringStackTraceFactory.cs#L26

I'm hoping that between the two of those we can extract something sensible.

CHANGELOG.md
```suggestion
- Sentry now includes an EXPERIMENTAL StringStackTraceFactory. This factory isn't as feature rich as the full `SentryStackTraceFactory`. However, it may provide better results if you are compiling your application AOT and not getting useful stack traces from the full stack trace factory. ([#4362](https://github.com/getsentry/sentry-dotnet/pull/4362))

```

src/Sentry/Extensibility/StringStackTraceFactory.cs
filipnavara: We have been using this regex in production since October 2024. It may not be fool-proof but it's somewhat battle-tested.

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Features
+
```
bruno-garcia, line 6: Is it on by default on AOT then?

How do I turn it on?

Would be nice to have a mention of that in the changelog too.

bruno-garcia (approved):
Some notes on changelog otherwise lgtm

src/Sentry/Extensibility/SimpleStackTraceFactory.cs
Flash0ver: Regex is just the implementation detail ... the Stack-Trace is created from the `Exception.StackTrace` string.

CHANGELOG.md
Flash0ver: Perhaps something we could also document in `Troubleshooting` when released?

jamescrosswell: Yeah when @Flash0ver and I discussed we were thinking we wouldn't promote this too widely, since it's quite experimental. However there are some additional docs on `SentryOptions.UseStackTraceFactory` (which is how you'd add this):
https://github.com/getsentry/sentry-dotnet/blob/e4120a976a1067c568fe6060cf3ad8158d6c44e6/src/Sentry/SentryOptions.cs#L1637-L1643

And we figured we'd add something to our [troubleshooting docs](https://docs.sentry.io/platforms/dotnet/troubleshooting/) as well... 

I'll add a blurb in the description of this PR as well.

Issue #4296, opened 2025-06-19
Sentry 5.11 causes DllNotFoundException in Native AoT application


martincostello:
### Package

Sentry.AspNetCore

### .NET Flavor

.NET

### .NET Version

9.0.301

### OS

Linux

### OS Version

Azure App Service

### Development Environment

Visual Studio v17.x

### SDK Version

5.11.0

### Workload Versions

?

### UseSentry or SentrySdk.Init call

[`builder.WebHost.UseSentry("{my-dsn}")`](https://github.com/martincostello/api/blob/4f28c79a5b5f0a55988c5f040ae5e00fa6b4385c/src/API/ApiBuilder.cs#L152)

### Steps to Reproduce

Run the application.

### Expected Result

Application works.

Appears that the default behaviour to disable sentry native has regressed as is no longer self-disabling.

### Actual Result

```text
2025-06-19T12:50:55.083948Z Unhandled exception. System.DllNotFoundException: Unable to load shared library 'sentry-native' or one of its dependencies. In order to help diagnose loading problems, consider using a tool like strace. If you're using glibc, consider setting the LD_DEBUG environment variable:
2025-06-19T12:50:55.0840016Z sentry-native.so: cannot open shared object file: No such file or directory
2025-06-19T12:50:55.0840052Z libsentry-native.so: cannot open shared object file: No such file or directory
2025-06-19T12:50:55.0840078Z sentry-native: cannot open shared object file: No such file or directory
2025-06-19T12:50:55.0840099Z libsentry-native: cannot open shared object file: No such file or directory
2025-06-19T12:50:55.0840127Z
2025-06-19T12:50:55.0840151Z    at System.Runtime.InteropServices.NativeLibrary.LoadLibErrorTracker.Throw(String) + 0x46
2025-06-19T12:50:55.0840223Z    at Internal.Runtime.CompilerHelpers.InteropHelpers.FixupModuleCell(InteropHelpers.ModuleFixupCell*) + 0x139
2025-06-19T12:50:55.0840258Z    at Internal.Runtime.CompilerHelpers.InteropHelpers.ResolvePInvokeSlow(InteropHelpers.MethodFixupCell*) + 0x35
2025-06-19T12:50:55.0840278Z    at Sentry.Native.C.sentry_options_new() + 0x1f
2025-06-19T12:50:55.0840299Z    at Sentry.Native.C.Init(SentryOptions) + 0x22
[...]

PR #4365, opened 2025-07-17
fix: Native AOT for `linux-musl-arm64`

jpnurmi:
Publishing a minimal Alpine .NET AOT container on a Mac with Apple silicon requires `linux-musl-arm64` as brought up in:
- https://github.com/getsentry/sentry-dotnet/issues/4296#issuecomment-3066775157

<details><summary>Problem: GitHub Actions hosted Node.js not available for MUSL ARM64</summary>

Node.js is required for JS actions, such as `actions/checkout`. For various reasons, GitHub Actions forces all containers to run a pre-compiled Node.js from the host, bind-mounted to the container in the `/__e` (aka. "externals") directory. The problem is that there's no suitable pre-compiled version for Linux MUSL ARM64.

A workaround is possible but it gets quite hacky. First of all, we need to bind-mount either `/__e/node20` or `/__e/node20/bin` to "escape" the built-in `/__e` bind-mount. Secondly, we need to make sure `/__e/node20/bin/node` points to a working node executable. In our case, we link it to the system installation i.e. `/usr/bin/node`. Finally, we need to alter `/etc/os-release` and pretend not to be Alpine (change `ID=alpine` to anything else) to avoid the following hard-coded Alpine container compatibility check:

> Error: JavaScript Actions in Alpine containers are only supported on x64 Linux runners. Detected ...

```csharp
// Check for Alpine container compatibility
if (!Constants.Runner.PlatformArchitecture.Equals(Constants.Architecture.X64))
{
    var os = Constants.Runner.Platform.ToString();
    var arch = Constants.Runner.PlatformArchitecture.ToString();
    var msg = $"JavaScript Actions in Alpine containers are only supported on x64 Linux runners. Detected {os} {arch}";
    throw new NotSupportedException(msg);
}
```
https://github.com/actions/runner/blob/main/src/Runner.Worker/Handlers/StepHost.cs

See also:
- https://github.com/actions/runner/issues/801
</details>

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -27,6 +27,12 @@ jobs:
             rid: linux-musl-x64
             container:
               image: ghcr.io/getsentry/sentry-dotnet-alpine:3.21
+          - os: ubuntu-24.04-arm
+            rid: linux-musl-arm64
+            container:
+              image: ghcr.io/jpnurmi/sentry-dotnet-alpine:3.21 # TODO: getsentry
+              volumes:
+                - /tmp/node20:/__e/node20
           - os: macos-15 # Pin macos to get the version of Xcode that we need: https://github.com/actions/runner-images/issues/10703
             rid: macos # universal (osx-arm64 + osx-x64)
           - os: windows-latest
@@ -35,6 +41,16 @@ jobs:
             rid: win-arm64
 
     steps:
+      ### TODO: clean up
+      - name: Prepare Node.js for non-x64 Alpine Linux
+        if: ${{ contains(matrix.container.image, 'alpine') && runner.arch != 'X64' }}
+        shell: bash
+        run: |
+          sudo mkdir -p /__e/node20/bin
+          sudo ln -sf /usr/bin/node /__e/node20/bin/node
+          sudo ln -sf /usr/bin/npm /__e/node20/bin/npm
+          sudo sed -i 's/ID=alpine/ID=unknown/' /usr/lib/os-release
+
       - name: Checkout
         uses: actions/checkout@v4
 
@@ -83,6 +99,14 @@ jobs:
               image: ghcr.io/getsentry/sentry-dotnet-alpine:3.21
               volumes:
                 - /var/run/docker.sock:/var/run/docker.sock
+          - os: ubuntu-24.04-arm
+            rid: linux-musl-arm64
+            slnf: Sentry-CI-Build-Linux-musl.slnf
+            container:
+              image: ghcr.io/jpnurmi/sentry-dotnet-alpine:3.21 # TODO: getsentry
+              volumes:
+                - /tmp/node20:/__e/node20
+                - /var/run/docker.sock:/var/run/docker.sock
           - os: macos-15 # Pin macos to get the version of Xcode that we need: https://github.com/actions/runner-images/issues/10703
             rid: macos # universal (osx-arm64 + osx-x64)
             slnf: Sentry-CI-Build-macOS.slnf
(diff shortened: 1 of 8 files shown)


scripts/build-sentry-native.ps1
```diff
@@ -37,11 +37,17 @@ try
     }
     elseif ($IsLinux)
     {
-        if ("Arm64".Equals([System.Runtime.InteropServices.RuntimeInformation]::OSArchitecture.ToString()))
+        $musl = (ldd --version 2>&1) -match 'musl'
+        $arm64 = ("Arm64".Equals([System.Runtime.InteropServices.RuntimeInformation]::OSArchitecture.ToString()))
```
Flash0ver: style-nitpick: to be in sync with `integration-test/runtime.Tests.ps1`

```suggestion
        $arm64 = "Arm64".Equals([System.Runtime.InteropServices.RuntimeInformation]::OSArchitecture.ToString())
```

.github/workflows/build.yml
```diff
@@ -83,6 +94,14 @@ jobs:
               image: ghcr.io/getsentry/sentry-dotnet-alpine:3.21
               volumes:
                 - /var/run/docker.sock:/var/run/docker.sock
+          - os: ubuntu-24.04-arm
+            rid: linux-musl-arm64
+            slnf: Sentry-CI-Build-Linux-musl.slnf
```
Flash0ver, line 99: question: reusing Solution Filter

Why do we need different filters between `linux-x64` and `linux-arm64`,
but `linux-musl-x64` and `linux-musl-arm64` may share the same?

src/Sentry/Platforms/Native/buildTransitive/Sentry.Native.targets
```diff
@@ -19,7 +19,7 @@
     <!-- Windows -->
     <FrameworkSupportsNative Condition="'$(RuntimeIdentifier)' == 'win-x64' or '$(RuntimeIdentifier)' == 'win-arm64'">true</FrameworkSupportsNative>
     <!-- Linux -->
-    <FrameworkSupportsNative Condition="'$(RuntimeIdentifier)' == 'linux-x64' or '$(RuntimeIdentifier)' == 'linux-arm64' or '$(RuntimeIdentifier)' == 'linux-musl-x64'">true</FrameworkSupportsNative>
+    <FrameworkSupportsNative Condition="'$(RuntimeIdentifier)' == 'linux-x64' or '$(RuntimeIdentifier)' == 'linux-arm64' or '$(RuntimeIdentifier)' == 'linux-musl-x64' or '$(RuntimeIdentifier)' == 'linux-musl-arm64'">true</FrameworkSupportsNative>
```
Flash0ver, line 22: question: Could/Should we add both `linux-musl-` targets to the `SingleFileTestApp.csproj` and `Sentry.Tests.Internals.ILSpy.SingleFileAppTests`?

Would this work / add any benefit?
I may totally misunderstand this though.

---

Wait ... actually ... it seems that the `SingleFileAppTests` are `[SKIP]`ped on all targets in CI.
This is a different issue unrelated to this changeset.

Flash0ver (approved):
🤘 ... 2 questions and 1 nitpick

.github/workflows/build.yml
jpnurmi: Yes, both are "Linux without mobile" because they don't have Android tooling available. We don't actually need the duplicates anymore now that the solution filter name is passed from the CI job matrix. What would be a good name? Something like `Sentry-CI-Build-Linux-NoMobile.slnf`?

Flash0ver: Oh ... I see ...
because `Sentry-CI-Build-Linux-arm64.slnf` and `Sentry-CI-Build-Linux-musl.slnf` are identical,
but different to `Sentry-CI-Build-Linux.slnf`.

Yes ... `..NoMobile..` sounds perfect,
as we already established a pattern with `SentryNoMobile.slnf` and `NO_MOBILE` MSBuild property.

jpnurmi: Sounds good, will follow up!

PR #4366, opened 2025-07-17
fix: Clean repo builds no longer fail when building native

bitsandfoxes:
When building the SDK I keep having the build break on me with.

```
Sentry net8.0 failed with 2 error(s) (3.2s)
  EXEC : CMake error : CMAKE_CXX_COMPILER not set, after EnableLanguage
  /Users/bitfox/Workspace/sentry-unity/src/sentry-dotnet/src/Sentry/Platforms/Native/Sentry.Native.targets(93,5): error MSB3073: The command "pwsh ../../scripts/build-sentry-native.ps1 -Clean" exited with code 1.
Sentry net9.0 failed with 1 error(s) (3.0s)
  /Users/bitfox/Workspace/sentry-unity/src/sentry-dotnet/src/Sentry/Platforms/Native/Sentry.Native.targets(93,5): error MSB3073: The command "pwsh ../../scripts/build-sentry-native.ps1 -Clean" exited with code 1.
```

Digging through it I run into this

```
Sentry net9.0 failed with 1 error(s) (6.1s)
      -- The C compiler identification is AppleClang 17.0.0.17000013
      -- ... a bunch of other flags ...
      -- Build files have been written to: /Users/bitfox/Workspace/sentry-unity/src/sentry-dotnet/modules/sentry-native/build
      [  0%] Building C object CMakeFiles/sentry.dir/vendor/mpack.c.o
      [100%] Linking C static library libsentry.a
      [100%] Built target sentry
      Copying modules/sentry-native/build/libsentry.a to src/Sentry/Platforms/Native/sentry-native/osx/libsentry-native.a
      Copy-Item: /Users/bitfox/Workspace/sentry-unity/src/sentry-dotnet/scripts/build-sentry-native.ps1:88
      Line |
        88 |      Copy-Item -Force -Path $srcFile -Destination $outFile
           |      ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
           | Cannot find path
           | '/Users/bitfox/Workspace/sentry-unity/src/sentry-dotnet/modules/sentry-native/build/libsentry.a' because it does not exist.
      /Users/bitfox/Workspace/sentry-unity/src/sentry-dotnet/src/Sentry/Platforms/Native/Sentry.Native.targets(93,5): error MSB3073: The command "pwsh /Users/bitfox/Workspace/sentry-unity/src/sentry-dotnet/scripts/build-sentry-native.ps1 -Clean" exited with code 1.
    Sentry net8.0 succeeded (6.7s) → src/Sentry/bin/Debug/net8.0/Sentry.dll
      -- The C compiler identification is AppleClang 17.0.0.17000013
      -- ... a bunch of other flags ...
      -- Build files have been written to: /Users/bitfox/Workspace/sentry-unity/src/sentry-dotnet/modules/sentry-native/build
      [ 23%] Building C object CMakeFiles/sentry.dir/vendor/mpack.c.o
      [ 96%] Building C object CMakeFiles/sentry.dir/src/transports/sentry_transport_none.c.o
      Linking C static library libsentry.a
      Built target sentry
      Copying modules/sentry-native/build/libsentry.a to src/Sentry/Platforms/Native/sentry-native/osx/libsentry-native.a
```

looks like building `Sentry` causes the build native target to run in parallel, corrupting the build output and failing.

So instead, we make use of [this article](https://learn.microsoft.com/visualstudio/msbuild/run-target-exactly-once) explaining how to run a target really only once.

While debugging, the resolving the path to the build script kept breaking. So instead of `../..` we can go with `$(MSBuildThisFileDirectory)_and then relative path from here_`.

#skip-changelog

### DIFF
--- a/src/Sentry/Platforms/Native/Sentry.Native.targets
+++ b/src/Sentry/Platforms/Native/Sentry.Native.targets
@@ -1,10 +1,10 @@
 <Project>
 
   <PropertyGroup>
-    <SentryNativeSourceDirectory>..\..\modules\sentry-native\</SentryNativeSourceDirectory>
+    <SentryNativeSourceDirectory>$(MSBuildStartupDirectory)\modules\sentry-native\</SentryNativeSourceDirectory>
     <SentryNativeLibraryName>sentry-native</SentryNativeLibraryName>
-    <SentryNativeBuildScript>../../scripts/build-sentry-native.ps1</SentryNativeBuildScript>
-    <SentryNativeBuildInputs>../../.git/modules/modules/sentry-native/HEAD;$(MSBuildThisFileDirectory)Sentry.Native.targets;$(SentryNativeBuildScript)</SentryNativeBuildInputs>
+    <SentryNativeBuildScript>$(MSBuildStartupDirectory)/scripts/build-sentry-native.ps1</SentryNativeBuildScript>
+    <SentryNativeBuildInputs>$(MSBuildStartupDirectory)/.git/modules/modules/sentry-native/HEAD;$(MSBuildThisFileDirectory)Sentry.Native.targets;$(SentryNativeBuildScript)</SentryNativeBuildInputs>
     <SentryNativeOutputDirectory>$(MSBuildThisFileDirectory)sentry-native\</SentryNativeOutputDirectory>
     <!-- List of runtime identifiers: https://github.com/dotnet/runtime/blob/main/src/libraries/Microsoft.NETCore.Platforms/src/runtime.json -->
     <NativeLibRelativePath-win-x64>win-x64</NativeLibRelativePath-win-x64>
--- a/src/Sentry/Platforms/Native/Sentry.Native.targets
+++ b/src/Sentry/Platforms/Native/Sentry.Native.targets
@@ -87,6 +87,7 @@
   <Target Name="_BuildSentryNativeSDK"
     BeforeTargets="DispatchToInnerBuilds"
     Condition="'$(_SentryIsNet8OrGreater)' == 'true' and '$(CI)' != 'true'"
+    Inputs="$(SentryNativeBuildInputs)"
     Outputs="$(SentryNativeBuildOutputs)">
     <!-- We want a "-Clean" because if the build script changes, previous cmake cache may contain invalid defines. -->
     <Exec Command="pwsh $(SentryNativeBuildScript) -Clean" />
--- a/src/Sentry/Platforms/Native/Sentry.Native.targets
+++ b/src/Sentry/Platforms/Native/Sentry.Native.targets
@@ -85,7 +85,7 @@
   <!-- Build the Sentry Native SDK (this only runs on local machines because in CI we expect the SDK to be
        built already on each native platform and fetched for the final .net build. -->
   <Target Name="_BuildSentryNativeSDK"
-    BeforeTargets="DispatchToInnerBuilds"
+    BeforeTargets="Build"
     Condition="'$(_SentryIsNet8OrGreater)' == 'true' and '$(CI)' != 'true'"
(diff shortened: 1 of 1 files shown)


src/Sentry/Platforms/Native/Sentry.Native.targets
```diff
@@ -85,11 +85,25 @@
...
     Inputs="$(SentryNativeBuildInputs)"
     Outputs="$(SentryNativeBuildOutputs)">
     <!-- We want a "-Clean" because if the build script changes, previous cmake cache may contain invalid defines. -->
     <Exec Command="pwsh $(SentryNativeBuildScript) -Clean" />
   </Target>
+
+  <!-- We want to build the sentry-native only once, BEFORE building TFM specific verions of the SDK -->
+  <!-- How to run a target exactly once: https://learn.microsoft.com/visualstudio/msbuild/run-target-exactly-once-->
+  <Target Name="_BuildSentryNativeSDKBeforeOuterBuild"
+    Condition="'$(_SentryIsNet8OrGreater)' == 'true' and '$(CI)' != 'true'"
```
bitsandfoxes, line 97: We have to keep the condition on the "outer" target as the TF gets removed on line `108`

src/Sentry/Platforms/Native/Sentry.Native.targets
```diff
@@ -85,11 +85,25 @@
   <!-- Build the Sentry Native SDK (this only runs on local machines because in CI we expect the SDK to be
        built already on each native platform and fetched for the final .net build. -->
   <Target Name="_BuildSentryNativeSDK"
-    BeforeTargets="DispatchToInnerBuilds;BeforeBuild"
```
bitsandfoxes, line 88: The "when" now gets resolved by this target being "clamped" by the two new ones.

src/Sentry/Platforms/Native/Sentry.Native.targets
```diff
@@ -1,10 +1,10 @@
 <Project>
 
   <PropertyGroup>
-    <SentryNativeSourceDirectory>..\..\modules\sentry-native\</SentryNativeSourceDirectory>
+    <SentryNativeSourceDirectory>$(MSBuildThisFileDirectory)..\..\..\..\modules\sentry-native\</SentryNativeSourceDirectory>
```
bruno-garcia, line 4: Does this work when building from the repo root as well as from, say `src/Sentry`?
if not, `$(ProjectRoot)/../../` might do the job

src/Sentry/Platforms/Native/Sentry.Native.targets
```diff
@@ -85,11 +85,25 @@
   <!-- Build the Sentry Native SDK (this only runs on local machines because in CI we expect the SDK to be
        built already on each native platform and fetched for the final .net build. -->
   <Target Name="_BuildSentryNativeSDK"
-    BeforeTargets="DispatchToInnerBuilds;BeforeBuild"
-    Condition="'$(_SentryIsNet8OrGreater)' == 'true' and '$(CI)' != 'true'"
```
bitsandfoxes, line 89: The condition needs to be added to the new one's as `'$(_SentryIsNet8OrGreater)' == 'true'` does not resolve due to TF being removed.

src/Sentry/Platforms/Native/Sentry.Native.targets
bitsandfoxes: It works because it is relative path from a known position of "this" file.
The `$(ProjectRoot)` seems to be a Powershell thing? And in the Unity SDK we've got a ps1 function that resolves that for us.

vaind (approved):
I'm flabbergasted how the target change helps at all... but if it does... 🤷

Issue #4368, opened 2025-07-18
Intermittent errors setting up environment for Win ARM64


jamescrosswell:
### Description

See https://github.com/getsentry/sentry-dotnet/actions/runs/16363439951/job/46235722773?pr=4294#step:5:37

```
zstandard not installed. An error occurred during installation:
 Error downloading 'zstandard.1.5.7.20250308' from 'https://community.chocolatey.org/api/v2/package/zstandard/1.5.7.20250308'.
```

jpnurmi:
Ideally, `zstd` would come pre-installed like on GitHub's Windows X64 images:
- https://github.com/actions/runner-images/issues/14081

It doesn't have `winget` either (potentially more stable than `choco`):
- https://github.com/actions/runner-images/issues/14083

PR #4369, opened 2025-07-18
ci: install zstd for win arm64 straight from github releases

jpnurmi:
Essentially does the same as what Chocolatey did under the hood <sup>[1]</sup>, but replaces Chocolatey API requests with one blazing-fast direct GitHub HTTP request.

| Before | After |
|---|---|
| image | image |

References:
1. https://community.chocolatey.org/packages/zstandard#files

Fixes: #4368

#skip-changelog

### DIFF
--- a/.github/actions/install-zstd/action.yml
+++ b/.github/actions/install-zstd/action.yml
@@ -1,0 +1,26 @@
+name: Install zstd on Windows ARM64
+description: |
+  zstd is needed for cross OS actions/cache but missing from windows-11-arm
+  (https://github.com/actions/partner-runner-images/issues/99). Once the issue
+  is resolved, this action can be removed.
+inputs:
+  version:
+    description: 'zstd version'
+    required: false
+    default: '1.5.7'
+
+runs:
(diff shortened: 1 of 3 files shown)


jamescrosswell (approved):
Thanks @jpnurmi !

.github/actions/install-zstd/action.yml
```diff
@@ -0,0 +1,26 @@
...
+    default: '1.5.7'
+
+runs:
+  using: composite
+  steps:
+    - name: Install zstd
+      if: runner.os == 'Windows' && runner.arch == 'ARM64'
+      shell: pwsh
+      env:
+        ZSTD_VERSION: ${{ inputs.version }}
```
bruno-garcia, line 19: is there a way to use sha pinning here?

jpnurmi: thanks for flagging it. we ~~can~~ should verify the checksum. would this do?
- https://github.com/getsentry/sentry-dotnet/pull/4396

PR #4374, opened 2025-07-21
fix: overwrite DSC `sample_rate` when sampling

### LINKED ISSUE #4392: fix: overwrite DSC `sample_rate` when sampling
### Overwrite Dynamic Sampling Context (DSC) `sample_rate` when sampling decision is made

See: https://linear.app/getsentry/project/update-the-dscs-sample-rate-e4e8398e6acc/overview
See also: https://github.com/getsentry/team-sdks/issues/117

> ### Description
> As required by Relay and the EAP, the sample_rate in the DSC must specify the actual sample rate of the trace. If an SDK overrides the sample rate of an incoming trace, for example by invoking the tracesSampler, the new rate must be set on the DSC.
> Note that usually the DSC is supposed to be immutable, which includes the sample rate, to ensure complete traces. Invoking the tracesSampler purposefully opts out of this and modifies the DSC.
>
> ### Implementation
> On an incoming trace, the DSC's `sample_rate` is updated if:
> - an explicit sampling decision is forced, e.g. `startTransaction(sampled: true)`
> - the `tracesSampler` is invoked
> - the `tracesSampleRate` is used
> - the SDK downsamples due to backpressure

#### Symptom
The Trace-View's extrapolation is based on the `"sample_rate"` of the _Dynamic Sampling Context_.
So it needs to reflect the actual `"sample_rate"` used, when the downstream SDK makes a sampling decision.

Flash0ver:
Fixes #4392
- https://github.com/getsentry/sentry-dotnet/issues/4392

---

### Overwrite Dynamic Sampling Context (DSC) `sample_rate` when sampling decision is made

Fixes: https://linear.app/getsentry/project/update-the-dscs-sample-rate-e4e8398e6acc/overview
See also: https://github.com/getsentry/team-sdks/issues/117

#### Symptom
The Trace-View's extrapolation is based on the `"sample_rate"` of the _Dynamic Sampling Context_.

#### Changes
When a DSC is available, then its immutability is ignored and its `sample_rate` overwritten
1. when `TracesSampler` is used and applied, then its resulting Sample-Rate is set
2. when `TracesSampleRate` is used and applied, that Sample-Rate is set
   - unless `ITransactionContext.IsSampled` is set (when an upstream sampling decision has been made), then the DSC `"sample_rate"` is kept unchanged

Also, refactored `test/Sentry.Tests/HubTests.cs`
- replaced `hub.ConfigureScope` with `hub.ScopeManager.ConfigureScope` for assertions
  - the former catches all exceptions, hiding assertion failures
  - the latter does not catch any exception, surfacing assertion failures
  - see also https://github.com/getsentry/sentry-dotnet/pull/4374#discussion_r2225736346
- made some tests more explicit / consistent

#### Unchanged
- no changes to making the sampling decision
- no changes to setting the `SampleRate` on neither the resulting `TransactionTracer` nor `UnsampledTransaction`
- possible follow-up: #4386

#### Note
The `sample_rate` overwritten in the DSC is now the same `SampleRate` set on the resulting `TransactionTracer` or `UnsampledTransaction` (`TransactionTracer.SampleRate` / `UnsampledTransaction.SampleRate`) with the following precedence
1. when `SentryOptions.TracesSampler` reaches a sampling decision (is not null and returns not null)
   - overwrite DSC
2. when `ITransactionContext.IsSampled` is set (an upstream sampling decision has been made)
   - keep DSC as is
3. when `SentryOptions.TracesSampleRate` reaches a sampling decision (is not null)
   - overwrite DSC
4. otherwise
   - keep DSC as is


src/Sentry/Internal/Hub.cs
```diff
@@ -196,6 +196,9 @@ internal ITransactionTracer StartTransaction(
         sampleRate ??= _options.TracesSampleRate ?? 0.0;
         isSampled ??= context.IsSampled ?? SampleRandHelper.IsSampled(sampleRand, sampleRate.Value);
 
+        // Ensure the actual sampleRate is set on the provided DSC (https://github.com/getsentry/team-sdks/issues/117)
+        dynamicSamplingContext = dynamicSamplingContext?.WithSampleRate(sampleRate.Value);
```
jamescrosswell: I think we only need to execute this logic if/when this code gets executed:
https://github.com/getsentry/sentry-dotnet/blob/9521bd145dca2ea11d6b8a2b5e9292085f279afe/src/Sentry/Internal/Hub.cs#L189

Since that's the only situation in which the sample rate gets overwritten (by the trace sampler). 

So it could probably be moved inside that if block.

test/Sentry.AspNetCore.Tests/SentryTracingMiddlewareTests.cs
```diff
@@ -316,7 +316,7 @@ public async Task Baggage_header_propagates_to_outbound_requests(bool shouldProp
                 "sentry-trace_id=75302ac48a024bde9a3b3734a82e36c8, " +
                 "sentry-public_key=d4d82fc1c2c4032a83f3a29aa3a3aff, " +
                 "sentry-sample_rand=0.1234, " +
-                "sentry-sample_rate=0.5";
+                "sentry-sample_rate=1";
```
jamescrosswell, line 319: I see this isn't this propagating the result of the incoming baggage header anymore, since the test uses a TraceSampler.

Maybe we need separate tests now:
1. Using a trace sampler in the options... then we expect the sentry-sample_rate to be overwrritten
2. Setting a tracesamplerate in the optinons... then we expect the value to be propagated

src/Sentry/Internal/Hub.cs
Flash0ver: Is otherwise `sample_rate` always the actual Sample-Rate used?

Including when a Baggage-Header is present in an Upstream/Downstream-SDK scenario?
i.e.
https://github.com/getsentry/sentry-dotnet/blob/6e50eaa87b3aa24a921824c4494566b01f0640f0/src/Sentry.AspNetCore/SentryMiddleware.cs#L118
https://github.com/getsentry/sentry-dotnet/blob/6e50eaa87b3aa24a921824c4494566b01f0640f0/src/Sentry.AspNetCore/SentryTracingMiddleware.cs#L66-L68

test/Sentry.AspNetCore.Tests/SentryTracingMiddlewareTests.cs
Flash0ver: Have I misunderstood the issue?
That, currently, `"sample_rate"` of the DSC remains unchanged after a sampling decision,
and instead should be overwritten with the actual `"sample_rate"` used?

e.g.
A downstream iOS application sends a Web-Request to an upstream ASP.NET Core application,
where currently the ASP.NET Core web-app is making its own sampling decision, but still sending the `"sample_rate"` of the iOS mobile-app,
resulting in Sentry's Traces View to extrapolate the graph wrongly.

src/Sentry/Internal/Hub.cs
jamescrosswell: Yeah there are three scenarios for sample rate basically:
1. There is no sample rate in the incoming header and there is no trace sampling function... in this case our SDK makes a sampling decision based on the `options.TracesSampleRate` configured and propagates that sample rate in any outgoing requests downstream
2. There is a sample rate in the incoming header and there is no trace sampling function... in this case we always respect the incoming sample decision and propagate the DSC verbatim (except perhaps the ReplayId... but this isn't really part of the DSC - it's included there just as a way to communicate this to Sentry Ingest)
3. There is an `options.TracesSampler` sampling function configured... in this case we just run the function and (in the PR apparently) then whatever rate it spits back is what gets written to the DSC (whether that DSC was created from an incoming request or not)

So unless there's a `TracesSampler` configured on the options, the SampleRate in the DSC will never be overwritten.

test/Sentry.AspNetCore.Tests/SentryTracingMiddlewareTests.cs
jamescrosswell: > That, currently, "sample_rate" of the DSC remains unchanged after a sampling decision

That's true... but unless the sampling decision was made by a TraceSampler function, that's the desired behaviour. 

So I think we need a few more test scenarios basically:
1. Retain the existing tests (that insure the incoming DSC is propagated verbatim, when no trace sampling function is configured. Options [like this](https://github.com/getsentry/sentry-dotnet/blob/3197ab6139fbe9f1369e179cf07d922aab323f0c/test/Sentry.AspNetCore.Tests/SentryTracingMiddlewareTests.cs#L466).
2. Add some new tests (or just permutations in the existing theories) to ensure the SampleRate is overwritten correctly in the DSC if there is a TraceSampler function. Options [like this](https://github.com/getsentry/sentry-dotnet/blob/3197ab6139fbe9f1369e179cf07d922aab323f0c/test/Sentry.AspNetCore.Tests/SentryTracingMiddlewareTests.cs#L523-L527).

At the moment, I think the tests use a mix of both of the above techniques to set the sample rate, without worrying too much about which one is being used... since up until now the behaviour was identical regardless of how the sampling decision was made.

src/Sentry/Internal/Hub.cs
```diff
@@ -196,6 +196,9 @@ internal ITransactionTracer StartTransaction(
         sampleRate ??= _options.TracesSampleRate ?? 0.0;
         isSampled ??= context.IsSampled ?? SampleRandHelper.IsSampled(sampleRand, sampleRate.Value);
 
+        // Ensure the actual sampleRate is set on the provided DSC (https://github.com/getsentry/team-sdks/issues/117)
```
Flash0ver: todo: when an explicit sampling decision is ~forced~ made upstream
- when `true`, set DSC `sample_rate` to `1.0`
- when `false`, set DSC `sample_rate` to `0.0`

We currently don't explicitly support forcing a sampling decision by the user when starting a new transaction:
See #4386.

test/Sentry.Tests/HubTests.cs
```diff
@@ -66,13 +66,13 @@ public void PushAndLockScope_DoesNotAffectOuterScope()
...
-        hub.ConfigureScope(s => Assert.False(s.Locked));
+        hub.ScopeManager.ConfigureScope(s => Assert.False(s.Locked));
         using (hub.PushAndLockScope())
         {
-            hub.ConfigureScope(s => Assert.True(s.Locked));
+            hub.ScopeManager.ConfigureScope(s => Assert.True(s.Locked));
         }
 
-        hub.ConfigureScope(s => Assert.False(s.Locked));
+        hub.ScopeManager.ConfigureScope(s => Assert.False(s.Locked));
```
Flash0ver: test: This test fails during CI (`Run Android API-34 Test (net9.0)` and `Run Android API-36 Test (net9.0)`)

Previously, assertions were using `hub.ConfigureScope(scope => ...));`.
But `Hub.ConfigureScope(Action<Scope>)` catches all exceptions:
https://github.com/getsentry/sentry-dotnet/blob/bc91dd3bf6cf844e239ac429892d95b9ce00a21b/src/Sentry/Internal/Hub.cs#L96-L106

Switching to instead use `Hub.ScopeManager.ConfigureScope(Action<Scope>)` no longer catches any exceptions in the callback:
https://github.com/getsentry/sentry-dotnet/blob/6e50eaa87b3aa24a921824c4494566b01f0640f0/src/Sentry/Internal/SentryScopeManager.cs#L35-L39

In short:
We were previously hiding a failed assertion that is now surfaced.

test/Sentry.AspNetCore.Tests/SentryTracingMiddlewareTests.cs
jamescrosswell: Spoke to @Flash0ver about this... there's another scenario which is that SDK users can force `IsSampled` via the `ITransactionContext` when calling `StartTransaction`... and once we implement the backpressure management, it might get overwritten by downsampling. So looks like we always have to run this logic.

test/Sentry.Tests/HubTests.cs
jamescrosswell: Since this is a P0 and it's unrelated to this PR, we should probably create a new issue for this and deal with it separately.

src/Sentry/Internal/Hub.cs
Flash0ver: When the `TracesSampler` is not applied,
and a sampling decision has been made upstream (distributed tracing) already,
then
- setting DSC `sample_rate` to `1.0` when upstream decision was `true`
- setting DSC `sample_rate` to `0.0` when upstream decision was `false`

src/Sentry/Internal/Hub.cs
Flash0ver: From the linked issue:

```TEXT
# Implementation
On an incoming trace, the DSC's `sample_rate` is updated if:
- an explicit sampling decision is forced, e.g. `startTransaction(sampled: true)`
- the `tracesSampler` is invoked
- the `tracesSampleRate` is used
- the SDK downsamples due to backpressure
```

- we currently don't explicitly support forced transactions ... see #4386
- we currently don't downsample due to backpressure

So I changed the precedence to:
1. when `TracesSampler` is used and the result is applied, set that Sample-Rate to the DSC
2. when `TracesSampleRate` is available, set that Sample-Rate to the DSC
3. otherwise, set `0.0`

This is also the value which is set in the returned `TransactionTracer.SampleRate` or `UnsampledTransaction.SampleRate` (which this PR leaves unchanged).

Is that correct - or have still not fully understood the expectation?

test/Sentry.Tests/HubTests.cs
Flash0ver: I erroneously created #4385.
The test failure is expected, because Android and iOS applications run in Global-Mode.

src/Sentry/Internal/Hub.cs
```diff
@@ -189,13 +189,23 @@ internal ITransactionTracer StartTransaction(
...
             }
         }
 
         // If the sampling decision isn't made by a trace sampler we check the trace header first (from the context) or
         // finally fallback to Random sampling if the decision has been made by no other means
-        sampleRate ??= _options.TracesSampleRate ?? 0.0;
-        isSampled ??= context.IsSampled ?? SampleRandHelper.IsSampled(sampleRand, sampleRate.Value);
+        if (isSampled == null)
+        {
+            Debug.Assert(sampleRate == null);
```
```suggestion
```
Not really necessary, given that it's directly below a check to see whether `isSampled == null`

test/Sentry.AspNetCore.Tests/SentryTracingMiddlewareTests.cs
jamescrosswell, line 319: Why is the `sentry-sample_rate` being overwritten in this scenario?

The [incoming](https://github.com/getsentry/sentry-dotnet/blob/23a45c20af9a9ad3630d67051c807fa77c1a602a/test/Sentry.AspNetCore.Tests/SentryTracingMiddlewareTests.cs#L300-L305) sample rate is `0.5` and there's no TracesSampler configured on the options:
https://github.com/getsentry/sentry-dotnet/blob/23a45c20af9a9ad3630d67051c807fa77c1a602a/test/Sentry.AspNetCore.Tests/SentryTracingMiddlewareTests.cs#L333-L337

I think we only want to overwrite the sentry-sample_rate in the DSC if:
- The `TracesSampler` overrides the sampling decision
- Sampling is forced by the SDK users when they create a transaction (we haven't yet implemented this in .NET)
- Downsampling occurs because of backpressure (again, we haven't yet implemented this in .NET)

Flash0ver: There is no `TracesSampler` indeed ... but the `TracesSampleRate` is set to `1.0`.

jamescrosswell: So are we saying we always overwrite the `sentry-sample_rate` now? The value from upstream will only ever be propagated if it happens to be the same as the value we have configured for the current process?

I'm OK with that as long as we're certain that's what we're supposed to be doing.

Flash0ver: Oh ... sorry ... we just realized that you don't have access to the issues linked above.
I should have copied the original text of the original issue there as well.
Opening issue in `getsentry/sentry-dotnet`.
See #4392

src/Sentry/Internal/Hub.cs
Flash0ver: I have reverted this code change, because I still was overwriting the `sample_rate` when a sampling decision has been made upstream which I wrongly interpreted as a "forced" sampling decision by the user, a scenario that we currently don't explicitly support yet: see #4386

test/Sentry.AspNetCore.Tests/SentryTracingMiddlewareTests.cs
jamescrosswell: OK, so it looks like the only scenario in which we don't overwrite the sample-rate now is if the .NET SDK is disabled... then we just propagate verbatim. Otherwise we overwrite.

src/Sentry/Internal/Hub.cs
Flash0ver: After a discussion, I have brought the code back ... but this time without the `Debug.Assert`.
See also https://github.com/getsentry/sentry-dotnet/pull/4374#discussion_r2235738488.

src/Sentry/DynamicSamplingContext.cs
```diff
@@ -98,6 +98,24 @@ private DynamicSamplingContext(SentryId traceId,
...
+                {
+                    var items = Items.ToDictionary(kvp => kvp.Key, kvp => kvp.Value);
+                    items["sample_rate"] = sampleRate.ToString(CultureInfo.InvariantCulture);
+                    return new DynamicSamplingContext(items);
+                }
+            }
+        }
+
+        return this;
+    }
```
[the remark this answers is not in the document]
Flash0ver: this is the intended behavior:
- only update the `sample_rate`
- only return a new instance if the desired `sample_rate` is different from the existing `sample_rate`
- a different name for this method could be `UpdateSampleRate`
  - but since `DynamicSamplingContext` is immutable, it will not mutate the instance itself, but instead create and return a new copy
  - as a performance optimization, when the existing `sample_rate` already equals the desired `sample_rate`, we can avoid creating a new copy and instead reuse the existing instance
  - therefore, I prefer `With..` over `Update..` to communicate the immutability of the instance
- this code will be revisited during #4381

src/Sentry/Internal/Hub.cs
Flash0ver: As discussed in another channel:
We currently don't support #4386 just yet, therefore the behavior, for now, should be:
1. when `SentryOptions.TracesSampler` reaches a sampling decision (is not null and returns not null)
   - overwrite DSC (if available)
2. when `ITransactionContext.IsSampled` is set (an upstream sampling decision has been made)
   - keep DSC as is
3. when `SentryOptions.TracesSampleRate` reaches a sampling decision (is not null)
   - overwrite DSC (if available)
4. otherwise
   - keep DSC as is

test/Sentry.AspNetCore.Tests/SentryTracingMiddlewareTests.cs
Flash0ver: update: see https://github.com/getsentry/sentry-dotnet/pull/4374#discussion_r2235738488

src/Sentry/Internal/Hub.cs
Flash0ver: This logic has been removed:
The final behavior of this changeset is:

When the TracesSampler is not applied,
and a sampling decision has been made upstream (distributed tracing) already,
then the DSC `sample_rate` **is not** updated/overwritten, but instead propagated as-is.

test/Sentry.Tests/HubTests.cs
Flash0ver: I have updated the test to reflect and assert this expectation.

test/Sentry.AspNetCore.Tests/SentryTracingMiddlewareTests.cs
Flash0ver: We ended up with this behavior:
https://github.com/getsentry/sentry-dotnet/pull/4374#discussion_r2235738488

src/Sentry/Internal/Hub.cs
```diff
@@ -188,14 +188,26 @@ internal ITransactionTracer StartTransaction(
...
         // If the sampling decision isn't made by a trace sampler we check the trace header first (from the context) or
         // finally fallback to Random sampling if the decision has been made by no other means
-        sampleRate ??= _options.TracesSampleRate ?? 0.0;
-        isSampled ??= context.IsSampled ?? SampleRandHelper.IsSampled(sampleRand, sampleRate.Value);
+        if (isSampled == null)
+        {
+            sampleRate = _options.TracesSampleRate ?? 0.0;
+            isSampled = context.IsSampled ?? SampleRandHelper.IsSampled(sampleRand, sampleRate.Value);
+
+            if (context.IsSampled is null && _options.TracesSampleRate is not null)
```
jamescrosswell, line 205: **TLDR;** I think we can remove the `if` block here.

#### Long version

Apologies, I think I got us confused on Slack:

- `context` is of type `ITransactionContext`... I think when an upstream span is sampled, `context.IsParentSampled` will be true (not `context.IsSampled`). 
- Whether or not the parent span was sampled is probably not relevant. It's possible that our process has a different sample-rate configured to the upstream node... as long as we use the same sample rand, we maximise (not guarantee) the chance of having a complete trace. If we have a sample-rand of 0.5 and the upstream node has a sample rate of 0.75 but we have a sample rate of 0.25 (and assuming I've understood the requirements) I think we need to sample out in this case and put our own sample-rate (0.25) on the header... so we just overwrite in this scenario. 
    - I'm not 100% sure whether that's how sampling is implemented at Sentry though... maybe we only support different sample rates on different nodes via the TracesSampler function. @cleptric do you know?
- If `context.IsSampled == true` I _**think**_ that means the SDK user has forced sampling to true (passing it explicitly via the call to `StartTransaction`). 
    - Note that we do sometimes set this to `false` ourselves internally (e.g. [here](https://github.com/getsentry/sentry-dotnet/blob/6e50eaa87b3aa24a921824c4494566b01f0640f0/src/Sentry/TransactionTracer.cs#L307) and [here](https://github.com/getsentry/sentry-dotnet/blob/6e50eaa87b3aa24a921824c4494566b01f0640f0/src/Sentry/Internal/UnsampledTransaction.cs#L39))
     - This is what I think I got us confused about. I suggested this field might be `true` when the upstream span was sampled in... with fresh eyes, I think I was mistaken and that's what the `IsParentSampled` field is for.

On the bright side, if all of the above is correct, it means it should be trivial to implement the logic to force sampling (per the other SDKs)... basically we can force a sample-rate of `0.0` or `1.0` if `context.IsSampled` is set.

Flash0ver: followed-up via #4386

PR #4375, opened 2025-07-22
fix: Sentry.Tunnel overwrites the X-Forwarded-For

### LINKED ISSUE #1309: `Sentry.Tunnel` doesn't include `X-Forwarded-To`

jamescrosswell:
Resolves #1309:
- https://github.com/getsentry/sentry-dotnet/issues/1309

## Summary

The `X-Forwarded-For` header is supposed to preserve a chain of IP addresses that the message was forwarded via ([see docs](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/X-Forwarded-For#syntax)).

So for example it might contain: 
```
X-Forwarded-For: 203.0.113.195, 2001:db8:85a3:8d3:1319:8a2e:370:7348
``` 

Which is a specific instance of:
```
X-Forwarded-For: <client>, <proxy>, …, <proxyN>
```

Previously, the `SentryTunnelMiddleware` was overwriting this value so that it only included the IP address from `DefaultHttpContext.Connection.RemoteIpAddress`... which was fine if the request didn't include any `X-Forwarded-For` but, if the request was proxied (e.g. by CloudFlare) then effectively by doing that we delete the upstream IP addresses so we no longer know where the message came from (or the User's real IP address). 

## Solution

This PR changes the behaviour of the `SentryTunnelMiddleware` so that the `RemoteIpAddress` is appended to the `X-Forwarded-For` so that the client IP address is retained in the left most address from the list in the `X-Forwarded-For` header.


test/Sentry.AspNetCore.Tests/Tunnel/IntegrationsTests.cs
```diff
@@ -86,6 +98,35 @@ public async Task TunnelMiddleware_CanForwardEnvelopeToWhiteListedHost()
...
     }
 
+    [Fact]
+    public async Task TunnelMiddleware_XForwardedFor_RetainsOriginIp()
+    {
+        // Arrange: Create a request with X-Forwarded-For header
+        var requestMessage = new HttpRequestMessage(new HttpMethod("POST"), "/tunnel")
+        {
+            Content = new StringContent(
+                @"{""sent_at"":""2021-01-01T00:00:00.000Z"",""sdk"":{""name"":""sentry.javascript.browser"",""version"":""6.8.0""},""dsn"":""https://dns@sentry.io/1""}
```
Flash0ver: suggestion: use Raw-String-Literals

With Raw String Literals, I would expect these JSON strings to become a little bit easier to read.

src/Sentry.AspNetCore/SentryTunnelMiddleware.cs
```diff
@@ -98,10 +98,14 @@ public async Task InvokeAsync(HttpContext context, RequestDelegate next)
                     Method = HttpMethod.Post,
                     Content = new StreamContent(memoryStream),
                 };
+                var existingForwardedFor = context.Request.Headers["X-Forwarded-For"];
                 var clientIp = context.Connection?.RemoteIpAddress?.ToString();
                 if (clientIp != null)
                 {
-                    sentryRequest.Headers.Add("X-Forwarded-For", context.Connection?.RemoteIpAddress?.ToString());
+                    var forwardedFor = string.IsNullOrEmpty(existingForwardedFor)
```
Flash0ver: question: Is the `"X-Forwarded-For"` always either none or a single string, possibly comma-separated ... or could this represent multiple values, actually?

When the `StringValues` represents more than 1 string, then the implicit conversion in `string.IsNullOrEmpty` would allocate a new string (from the internal string-array).
In this case, perhaps `StringValues.Count` could be more efficient to detect whether there is any value existing.

src/Sentry.AspNetCore/SentryTunnelMiddleware.cs
```diff
@@ -98,10 +98,14 @@ public async Task InvokeAsync(HttpContext context, RequestDelegate next)
...
+                        ? clientIp
+                        : $"{existingForwardedFor}, {clientIp}";
```
Flash0ver: question: Would `Add` to the `"X-Forwarded-For"` header work as well?
Does `X-Forwarded-For` need to be a single comma-separated string, or could this be multiple values?

Could we maybe avoid the extra string-allocation, and do something like
```CSharp
if (existingForwardedFor.Count == 1)
{
    sentryRequest.Headers.Add("X-Forwarded-For", existingForwardedFor.ToString());
}
else if (existingForwardedFor.Count > 1)
{
    sentryRequest.Headers.Add("X-Forwarded-For", existingForwardedFor.ToArray());
}
sentryRequest.Headers.Add("X-Forwarded-For", clientIp);
```

src/Sentry.AspNetCore/SentryTunnelMiddleware.cs
```diff
@@ -98,10 +98,14 @@ public async Task InvokeAsync(HttpContext context, RequestDelegate next)
                     Method = HttpMethod.Post,
                     Content = new StreamContent(memoryStream),
                 };
+                var existingForwardedFor = context.Request.Headers["X-Forwarded-For"];
```
Flash0ver: question: In case of `clientIp` being `null`, do we want to add existing `"X-Forwarded-For"` to the `sentryRequest` regardless, or only when there is a `RemoteIpAddress` set?

src/Sentry.AspNetCore/SentryTunnelMiddleware.cs
jamescrosswell: There's no Add method on the StringValues type (it's a struct)... but we can maybe leverage the `Count` to create some tiny optimisations.

src/Sentry.AspNetCore/SentryTunnelMiddleware.cs
```diff
@@ -122,6 +122,25 @@ public async Task InvokeAsync(HttpContext context, RequestDelegate next)
...
+        }
+
+        if (existingForwardedFor.Count == 0)
+        {
+            return clientIp;
+        }
+
+        return string.IsNullOrEmpty(existingForwardedFor)
+            ? clientIp
+            : $"{existingForwardedFor}, {clientIp}";
```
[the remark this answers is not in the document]
jamescrosswell: In context:
https://github.com/getsentry/sentry-dotnet/blob/b3312243826b8c9b16639436f5232b6a8d5f621f/src/Sentry.AspNetCore/SentryTunnelMiddleware.cs#L134-L141

So existingForwardedFor will never be empty when this codepath is executed.

PR #4396, opened 2025-07-26
ci: verify zstd checksum

jpnurmi:
Add missing checksum verification for:
- https://github.com/getsentry/sentry-dotnet/pull/4369#pullrequestreview-3058044922


jamescrosswell (approved):
LGTM

.github/actions/install-zstd/action.yml
```diff
@@ -8,6 +8,10 @@ inputs:
     description: 'zstd version'
     required: false
     default: '1.5.7'
+  checksum:
+    description: 'zstd checksum'
+    required: false
+    default: 'acb4e8111511749dc7a3ebedca9b04190e37a17afeb73f55d4425dbf0b90fad9'
```
bruno-garcia, line 14: Do we have an 'auto update' automation for this? We mainly manage them here: https://github.com/getsentry/github-workflows

bruno-garcia (approved):

.github/actions/install-zstd/action.yml
jpnurmi: Not at the moment, no. Ideally, we wouldn't have to maintain this at all if it was pre-installed on the `win-11-arm` runner :)
- https://github.com/actions/partner-runner-images/issues/99

I suppose the second-best option would be to stay in sync with the Windows x64 runner. Surprisingly enough, instead of pinning a specific version, GitHub fetches the latest version of Zstd at the time of building the Windows runner image: https://github.com/actions/runner-images/blob/main/images/windows/scripts/build/Install-Zstd.ps1.

jamescrosswell: It's definitely easier to "set things up" by just pulling the latest version.

I've tried to avoid floating versions for dependencies, since sometimes it results in our CI builds failing without our having made any changes. 

There's a relatively low chance of that happening with `zstd` but still, ideally we'd pin the version and configure auto-update so that when we do bump the version of this, it's done in a separate/isolated commit.

jpnurmi: GitHub release assets have checksums since early June:
https://github.blog/changelog/2025-06-03-releases-now-expose-digests-for-release-assets/

```sh
$ gh api repos/getsentry/sentry-dotnet/releases/tags/5.14.0 \
  --jq '.assets[] | select(.name == "Sentry.5.14.0.nupkg") | {name: .name, digest: .digest}'
{
  "digest": "sha256:362de4e2c771a40aabd82fa5683ab716fd9a7ad877ef25f969613a49e84e2bf9",
  "name": "Sentry.5.14.0.nupkg"
}
```

The latest zstd v1.5.7 was released in February and only has separate manually uploaded `.sha256` files for `zstd-1.5.7.tar.{gz|zst}` but not for the `.zip` files for Windows.

```sh
$ gh api repos/facebook/zstd/releases/tags/v1.5.7 \
  --jq '.assets[] | select(.name == "zstd-v1.5.7-win64.zip") | {name: .name, digest: .digest}'
{
  "digest": null,
  "name": "zstd-v1.5.7-win64.zip"
}
```

Starting with the next zstd release, it should be possible to query the checksum.

==> https://github.com/getsentry/sentry-dotnet/issues/4432

Issue #4091, opened 2025-04-08
Windows builds are failing intermittently on CI


jamescrosswell:
See [here](https://github.com/getsentry/sentry-dotnet/actions/runs/14322754417/job/40142695386), for an example. The error log shows:
```
Error: C:\Program Files\dotnet\packs\Microsoft.Android.Sdk.Windows\35.0.50\tools\Xamarin.Android.Common.targets(2038,3): error XA3007: Could not link native shared library: libxamarin-app.so [D:\a\sentry-dotnet\sentry-dotnet\test\AndroidTestApp\AndroidTestApp.csproj::TargetFramework=net9.0-android]
C:\Program Files\dotnet\packs\Microsoft.Android.Sdk.Windows\35.0.50\tools\Xamarin.Android.Common.targets(2038,3): error XA3007:  [D:\a\sentry-dotnet\sentry-dotnet\test\AndroidTestApp\AndroidTestApp.csproj::TargetFramework=net9.0-android]
Error: C:\Program Files\dotnet\packs\Microsoft.Android.Sdk.Windows\35.0.50\tools\Xamarin.Android.Common.targets(2038,3): error XA3007: stderr | ld: error: failed to write output 'obj\Release\net9.0-android\android-x64\app_shared_libraries\x86_64\libxamarin-app.so': permission denied [D:\a\sentry-dotnet\sentry-dotnet\test\AndroidTestApp\AndroidTestApp.csproj::TargetFramework=net9.0-android]
```

Which is [coming from here](https://github.com/dotnet/android/blob/1719a35b8a0348a4a8dd0061cfc4dd7fe6612a3c/src/Xamarin.Android.Build.Tasks/Xamarin.Android.Common.targets#L2038-L2044) in the `Xamarin.Android.Common.targets`.

@vaind commented in slack that this might be a concurrency issue:
> Could not link native shared library: libxamarin-app.so 
> And i've also got that on a clean build locally. I believe it's a concurrency issue, because running dotnet build again on the same project succeeded. (edited) 

I can also reproduce this reliably on a clean build locally.

... now all we have to do is work out why 🙃

vaind:
> Potentially we're running out of disk space on the Windows CI runners...
I've got about 600 GB free on the dev drive so that's definitely not the case for me locally.

jamescrosswell:
Yeah I can reproduce this locally _sometimes_ when running a build after `git clean -dfx`. 

I suspect it's related to this:
[...]

PR #4400, opened 2025-07-29
fix: failed to write libxamarin-app.so

jamescrosswell:
Another try at #4091

It might be related to [the restore](https://github.com/getsentry/sentry-dotnet/blob/25a2d578cd35766fcccf9e9fbab660d238640fdf/test/Sentry.Android.AssemblyReader.Tests/Sentry.Android.AssemblyReader.Tests.csproj#L19-L26) operation rather than the build.

This PR ensures `AndroidTestApp.csproj` is restored exactly once (even though it's built multiple times).

#skip-changelog

### DIFF
--- a/test/Sentry.Android.AssemblyReader.Tests/Sentry.Android.AssemblyReader.Tests.csproj
+++ b/test/Sentry.Android.AssemblyReader.Tests/Sentry.Android.AssemblyReader.Tests.csproj
@@ -28,7 +28,7 @@
           BeforeTargets="DispatchToInnerBuilds" />
 
   <!--  Build the Android test app in various configurations during the build of this test project. -->
-  <Target Name="BuildTestAPKs" BeforeTargets="BeforeBuild" DependsOnTargets="EnsureAndroidTestAppRestored" Condition="'$(TargetPlatformIdentifier)' != 'android'">
+  <Target Name="BuildTestAPKs" BeforeTargets="BeforeBuild" Condition="'$(TargetPlatformIdentifier)' != 'android'">
     <!-- Restore exactly once: https://learn.microsoft.com/en-gb/visualstudio/msbuild/run-target-exactly-once -->
     <MSBuild Projects="$(MSBuildProjectFullPath)" Targets="EnsureAndroidTestAppRestored" RemoveProperties="TargetFramework" />
     <MSBuild Projects="$(MSBuildProjectFile)" Targets="_InnerBuildTestAPKs" Properties="TargetFramework=net8.0-android" Condition="$(TargetFramework) == 'net8.0'" />
--- a/test/Sentry.Android.AssemblyReader.Tests/Sentry.Android.AssemblyReader.Tests.csproj
+++ b/test/Sentry.Android.AssemblyReader.Tests/Sentry.Android.AssemblyReader.Tests.csproj
@@ -23,8 +23,14 @@
     <Exec Command="dotnet restore ../AndroidTestApp/AndroidTestApp.csproj --nologo" />
   </Target>
 
+  <!-- Restore exactly once: https://learn.microsoft.com/en-gb/visualstudio/msbuild/run-target-exactly-once -->
+  <Target Name="BuildTestAPKsBeforeOuterBuild" DependsOnTargets="EnsureAndroidTestAppRestored"
+          BeforeTargets="DispatchToInnerBuilds" />
+
   <!--  Build the Android test app in various configurations during the build of this test project. -->
-  <Target Name="BuildTestAPKs" BeforeTargets="DispatchToInnerBuilds;BeforeBuild" DependsOnTargets="EnsureAndroidTestAppRestored" Condition="'$(TargetPlatformIdentifier)' != 'android'">
+  <Target Name="BuildTestAPKs" BeforeTargets="BeforeBuild" DependsOnTargets="EnsureAndroidTestAppRestored" Condition="'$(TargetPlatformIdentifier)' != 'android'">
+    <!-- Restore exactly once: https://learn.microsoft.com/en-gb/visualstudio/msbuild/run-target-exactly-once -->
+    <MSBuild Projects="$(MSBuildProjectFullPath)" Targets="EnsureAndroidTestAppRestored" RemoveProperties="TargetFramework" />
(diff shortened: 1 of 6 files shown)


test/Sentry.Android.AssemblyReader.Tests/Sentry.Android.AssemblyReader.Tests.csproj
```diff
@@ -17,12 +17,10 @@
...
   <!-- Target to ensure AndroidTestApp is restored -->
-  <Target Name="EnsureAndroidTestAppRestored" Condition="'$(TargetPlatformIdentifier)' != 'android' and '$(AndroidTestAppRestored)' != 'true'">
-    <!-- Only restore once. Spawn as a separate process to escape any "no-restore" arguments to the build -->
+  <Target Name="EnsureAndroidTestAppRestored"
+          Condition="!Exists('$(MSBuildProjectDirectory)\..\AndroidTestApp\obj\project.assets.json')">
+    <Message Importance="high" Text="Restoring AndroidTestApp via MSBuild task..." />
     <Exec Command="dotnet restore ../AndroidTestApp/AndroidTestApp.csproj --nologo" />
-    <PropertyGroup>
-      <AndroidTestAppRestored>true</AndroidTestAppRestored>
-    </PropertyGroup>
```
jpnurmi, line 25: Does it currently wait until `dotnet restore` finishes and then set the `AndroidTestAppRestored` property, risking multiple `dotnet restore` running in parallel? Would moving these lines above the `Exec` have any impact?

Flash0ver: Hmmm ... perhaps it would ... but we still have some race conditions, wouldn't we? 🤔

But maybe something similar like [Run a target exactly once](https://learn.microsoft.com/visualstudio/msbuild/run-target-exactly-once) is the MSBuild-way to go with.
This approach already helped with #4366.

jamescrosswell:
I tried:
```
<!-- Target to ensure AndroidTestApp is restored -->
<Target Name="EnsureAndroidTestAppRestored" BeforeTargets="CoreBuild;Restore" Condition="!Exists('$(MSBuildProjectDirectory)\..\AndroidTestApp\obj\project.assets.json')">
    <Message Importance="high" Text="Restoring AndroidTestApp via MSBuild task..." />
    <Exec Command="dotnet restore ../AndroidTestApp/AndroidTestApp.csproj --nologo" />
</Target>
```

However `EnsureAndroidTestAppRestored`:
- does not get run when running `dotnet restore Sentry-CI-Build-macOS.slnf -bl:build.binlog`
- does get run when running `dotnet restore Sentry.Android.AssemblyReader.Tests.csproj -bl:build.binlog`

So some annoying NuGet package resolution stuff kicks in an optimises it out of the picture.

test/Sentry.Android.AssemblyReader.Tests/Sentry.Android.AssemblyReader.Tests.csproj
jamescrosswell: It's a good idea - I'll try that (even if I do find the explanation of that technique very confusing).

test/Sentry.Android.AssemblyReader.Tests/Sentry.Android.AssemblyReader.Tests.csproj
```diff
@@ -17,16 +17,20 @@
...
   </Target>
 
+  <!-- Restore exactly once: https://learn.microsoft.com/en-gb/visualstudio/msbuild/run-target-exactly-once -->
+  <Target Name="BuildTestAPKsBeforeOuterBuild" DependsOnTargets="EnsureAndroidTestAppRestored"
+          BeforeTargets="DispatchToInnerBuilds" />
```
jamescrosswell:
```suggestion
  <Target Name="BuildTestAPKsBeforeOuterBuild" DependsOnTargets="EnsureAndroidTestAppRestored"
          Condition="'$(TargetPlatformIdentifier)' != 'android'"
          BeforeTargets="DispatchToInnerBuilds" />
```

test/Sentry.Android.AssemblyReader.Tests/Sentry.Android.AssemblyReader.Tests.csproj
```diff
@@ -17,16 +17,21 @@
...
+
   <!--  Build the Android test app in various configurations during the build of this test project. -->
-  <Target Name="BuildTestAPKs" BeforeTargets="DispatchToInnerBuilds;BeforeBuild" DependsOnTargets="EnsureAndroidTestAppRestored" Condition="'$(TargetPlatformIdentifier)' != 'android'">
+  <Target Name="BuildTestAPKs" BeforeTargets="BeforeBuild" Condition="'$(TargetPlatformIdentifier)' != 'android'">
+    <!-- Restore exactly once: https://learn.microsoft.com/en-gb/visualstudio/msbuild/run-target-exactly-once -->
+    <MSBuild Projects="$(MSBuildProjectFullPath)" Targets="EnsureAndroidTestAppRestored" RemoveProperties="TargetFramework" />
     <MSBuild Projects="$(MSBuildProjectFile)" Targets="_InnerBuildTestAPKs" Properties="TargetFramework=net8.0-android" Condition="$(TargetFramework) == 'net8.0'" />
     <MSBuild Projects="$(MSBuildProjectFile)" Targets="_InnerBuildTestAPKs" Properties="TargetFramework=net9.0-android" Condition="$(TargetFramework) == 'net9.0'" />
   </Target>
```
jpnurmi, line 37: Now that all three `MSBuild` tasks are siblings, won't `EnsureAndroidTestAppRestored` and `_InnerBuildTestAPKs` run in parallel?

Should we keep the `MSBuild` tasks for `EnsureAndroidTestAppRestored` and `_InnerBuildTestAPKs` in separate targets and declare a dependency? Or is it somehow taken care of by the new `BuildTestAPKsBeforeOuterBuild`? :exploding_head:

jamescrosswell: ChatGPT reckons they won't:
> No, MSBuild does not execute tasks in parallel by default, even if they are sibling <MSBuild /> tasks within a target.
>
> If you do want to run multiple MSBuild invocations in parallel, you need to:
> 1. Put the inputs into an ItemGroup.
> 2. Use batching.
> 3. Set BuildInParallel="true"

So I think we're safe...

PR #4402, opened 2025-08-01
fix(logs): flush on terminating

Flash0ver:
### Summary

Flush the Logs-Buffer of the `SentryStructuredLogger` when the app terminates ungracefully.

### Remarks

In #4310, we added a Buffer for Logs, and implemented a Batching & Flushing strategy.
We did test and cover the case to Flush the Logs when the `Hub` is disposed, which includes the case when the App closes gracefully, triggered through `ProcessExit`.

But we missed to test and cover the case when the consuming App crashes.

This changeset is adding to flush the Logs-Buffer just before flushing the `ISentryClient`, triggered through `UnhandledException`.
This new flush, however, is slightly different than the existing auto-flushes triggered when a Logs-Buffer reaches `100` items or when `5` seconds have passed after the first Log-Message buffered. This new flush does not `Wait` until all currently active `Add` operations of the Batch-Processor (when writing logs via `SentrySdk.Experimental.Logger.Log..` or via `Microsoft.Extensions.Logging.ILogger`) are completed, but instead force-flushes immediately, because this is not a graceful shutdown.

### Changes
- Pass a `TimeSpan timeout` to the `Flush` operations related to Sentry Logs.
  - The three existing `Flush` operations (`Hub` disposed on graceful shutdown, buffer reaches `100` items, `5` seconds have passed after the first log written) remain unchanged by passing `Timeout.InfiniteTimeSpan`, which is the default behavior of the underlying `System.Threading.CountdownEvent`.
  - This new `Flush` operation (App terminates ungracefully, see `event UnhandledException`) force-flushes immediately, regardless of whether there are currently still `Add` operations on the Batch-Processor / Log-Buffer in progress, by passing `TimeSpan.Zero` as `timeout`, basically rendering the underlying `System.Threading.CountdownEvent.Wait()` a no-op.
- Also added a clarifying comment to existing code.

---

#skip-changelog
(feature logs not yet shipped)


Flash0ver:
@Sentry review

PR #4410, opened 2025-08-05
chore: Remove dangling experimental metrics classes

### LINKED ISSUE #4404: Delete unused metrics options class
### Description

Some left overs from the metrics implementation:

https://github.com/search?q=repo%3Agetsentry%2Fsentry-dotnet+ExperimentalMetricsOptions&type=code

jamescrosswell:
Resolves #4404
- https://github.com/getsentry/sentry-dotnet/issues/4404

#skip-changelog

### DIFF
--- a/test/Sentry.Tests/ApiApprovalTests.Run.DotNet8_0.verified.txt
+++ b/test/Sentry.Tests/ApiApprovalTests.Run.DotNet8_0.verified.txt
@@ -187,6 +187,8 @@ namespace Sentry
     public interface IHub : Sentry.ISentryClient, Sentry.ISentryScopeManager
     {
         Sentry.SentryId LastEventId { get; }
+        [System.Diagnostics.CodeAnalysis.Experimental("SENTRY0001")]
+        Sentry.SentryStructuredLogger Logger { get; }
         void BindException(System.Exception exception, Sentry.ISpan span);
         Sentry.SentryId CaptureEvent(Sentry.SentryEvent evt, System.Action<Sentry.Scope> configureScope);
         Sentry.SentryId CaptureEvent(Sentry.SentryEvent evt, Sentry.SentryHint? hint, System.Action<Sentry.Scope> configureScope);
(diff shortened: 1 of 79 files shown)


jamescrosswell:
Need to update from main once this has been merged in:
- https://github.com/getsentry/sentry-dotnet/pull/4415

test/Sentry.Tests/ApiApprovalTests.Run.DotNet8_0.verified.txt
```diff
@@ -40,22 +40,6 @@ namespace Sentry
         [System.Runtime.Serialization.EnumMember(Value="critical")]
         Critical = 3,
     }
-    public static class BuiltInSystemDiagnosticsMeters
```
Flash0ver, line 43: question: Is this worth a CHANGELOG entry?

It's technically breaking ... but also not really breaking, because both types and their members are not used by any Sentry code any longer.

Or is it unlikely that user code is referring to any of the deleted types and members (also considering the "Experimental" prefix of the `ExperimentalMetricsOptions`)?

jamescrosswell: ### TLDR; 

No

### Verbose answer

We already announced that Metrics were being deprecated in the [5.0.0 release](https://github.com/getsentry/sentry-dotnet/blob/main/CHANGELOG.md#500)... this is just tidy up here. I can't think of any scenarios where SDK users would still be using these classes. They only ever existed to support the metrics functionality that we've already removed.

There's no way people would be using this one (it's very specific):
https://github.com/getsentry/sentry-dotnet/blob/61723f0b2315155e0a86efec9a3d8aaae3efd388/src/Sentry/ExperimentalMetricsOptions.cs#L5-L8

And I can't see people using this one either. They should be using the relevant Aspire or OpenTelemetry extensions to subscribe to these metrics (from memory, I reverse engineered these from an early Aspire release):
https://github.com/getsentry/sentry-dotnet/blob/61723f0b2315155e0a86efec9a3d8aaae3efd388/src/Sentry/BuiltInSystemDiagnosticsMeters.cs#L5-L8

Flash0ver (approved):

Issue #4421, opened 2025-08-09
Central Package Management (CPM)


bruno-garcia:
### Description

https://learn.microsoft.com/en-us/nuget/consume-packages/central-package-management

* Helps avoiding versions spread across the code base accidently mismatching
* Allows for `CentralPackageTransitivePinningEnabled` if we decide to go with that
* We can be explicit about versions overrwide with `VersionOverride`. which would deserve a comment

Flash0ver:
Could be a bit "dangerous" with our packable `/src/` projects, where we may unintentionally bump a (transitive) dependency.
Especially when we reference different versions of a NuGet-package depending on the _TFM_, as in `Sentry.Extensions.Logging`.

Well, we could use `VersionOverride` for all our `/src/`'s `PackageReference`s,
or we only enable _CPM_ for `/benchmarks/`, `samples` and `test`.

PR #4422, opened 2025-08-09
add central nuget version mgmt

bruno-garcia:
WIP

Resolves: https://github.com/getsentry/sentry-dotnet/issues/4421

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -46,6 +46,7 @@
     <PackageVersion Include="Microsoft.Extensions.Configuration" Version="8.0.0" />
     <PackageVersion Include="Microsoft.Extensions.Configuration.Json" Version="6.0.0" />
     <PackageVersion Include="Microsoft.Extensions.DependencyInjection" Version="6.0.1" />
+    <PackageVersion Include="Microsoft.Extensions.DependencyInjection.Abstractions" Version="9.0.0" />
     <PackageVersion Include="Microsoft.Extensions.Hosting" Version="8.0.0" />
     <PackageVersion Include="Microsoft.Extensions.Logging.Debug" Version="9.0.0" />
   </ItemGroup>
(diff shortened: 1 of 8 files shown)


src/Sentry.Extensions.Logging/Sentry.Extensions.Logging.csproj
```diff
@@ -16,20 +16,20 @@
...
   </ItemGroup>
 
   <ItemGroup Condition="'$(TargetFramework)' == 'net9.0'">
-    <PackageReference Include="Microsoft.Extensions.Configuration.Binder" Version="9.0.0" />
-    <PackageReference Include="Microsoft.Extensions.Logging.Configuration" Version="9.0.0" />
-    <PackageReference Include="Microsoft.Extensions.Http" Version="9.0.0" />
+    <PackageReference Include="Microsoft.Extensions.Configuration.Binder" />
+    <PackageReference Include="Microsoft.Extensions.Logging.Configuration" />
+    <PackageReference Include="Microsoft.Extensions.Http" />
   </ItemGroup>
```
Flash0ver, line 33: thought: CPM for non-Packable source only

Although I do like CPM, as it simplifies package updates, I have the feeling it's a bit "dangerous" for our `/src/..` packages, as we may accidentally update a dependency of our NuGet packages. We could also accidentally update transitive dependencies of our _packable_ production projects should we use `<CentralPackageTransitivePinningEnabled>true</CentralPackageTransitivePinningEnabled>` (not in this PR, but should we add it later we may miss something).
This makes - IMO - reviewing `Version` upgrades of `PackageReference` harder, as we would need to verify first whether a change to `Directory.Packages.props` also impacts any packable `/src/..` project, and if so, whether that change was intended.

I am now thinking about enabling CPM for non-production/packable-Projects only:
- explicitly enable for `/benchmarks/`, `/samples/`, and `/test/`
- explicitly disable for `/src/`

This would make `PackageReference`-related "Repository Maintenance" tasks easier and quicker,
while keeping our Production NuGet packages explicit and intentional.

Flash0ver:
Let's revive this PR.
- Let's give it a try if the TFM selection works in there (e.g. see `Sentry.Extensions.Logging`)
- if not ... we may not get too much benefit out of CPM.

PR #4428, opened 2025-08-11
fix(logs): race condition in high volume logging scenarios

Flash0ver:
### Summary

Fix a race condition (especially in high volume structured logging scenarios), causing an `InvalidOperationException` in the underlying `CountdownEvent`.

### Remarks

When testing/dogfooding [5.14.0](https://github.com/getsentry/sentry-dotnet/releases/tag/5.14.0) in a private project, I noticed that (particularly in high volume logging scenarios) we may run into a `InvalidOperationException` in the `CountdownEvent`:
- decrementing the event's count below zero
- incrementing the event's count when set/signaled

This change is fixing this thread-safety issue in the control-flows involved in Structured Logging.

### Changes
- add Benchmark `EnqueueAndFlush_Parallel` that consistently reproduced both error paths
- add more `Debug.Assert` invocations, to document assumptions, that would fail unit tests when violated
- fix by making sure the custom `ScopedCountdownLock` is reset before disengaging
  - with the other way around, we may end up signalling the underlying `CountdownEvent` when currently still set

### Additional Change
- now also no longer allowing a new `CounterScope` to enter when a `LockScope` is already active (i.e. when the Lock is engaged)
  - previously it was possible to enter a new `CounterScope` when a `LockScope` was active, but the counter has not reached `0` just yet ... which could have "dragged out" the actual flush as long as new logs are coming in and the event has not reached `0` yet, for a buffer that is not full yet (so when Flush is triggered through the `5` second timeout)
  - this ensures that when a `Flush` is forced, e.g. when the Application is terminating unexpectedly, that the Flush-operation can no longer "drag out" as long as new logs come in until the Event actually reaches `0` or the buffer does become full
    - see also #4425
  - without (notably) pessimizing performance, see updated Benchmark results


src/Sentry/Threading/ScopedCountdownLock.cs
```diff
@@ -84,13 +91,13 @@ internal LockScope TryEnterLockScope()
 
     private void ExitLockScope()
     {
-        if (Interlocked.CompareExchange(ref _isEngaged, 0, 1) == 1)
+        Debug.Assert(_event.IsSet);
+        _event.Reset(); // reset the signaled event to the initial count of 1, so that new 'CounterScope' instances can be entered again
```
Flash0ver, line 95: note: fix

fixed by performing the "Reset before CompareExchange" when existing in the reverse order than the "CompareExchange before Signal" when entering

src/Sentry/Threading/ScopedCountdownLock.cs
```diff
@@ -139,7 +146,7 @@ internal LockScope(ScopedCountdownLock lockObj)
         internal void Wait()
         {
             var lockObj = _lockObj;
-            lockObj?._event.Wait();
+            lockObj?._event.Wait(Timeout.Infinite, CancellationToken.None);
```
Flash0ver, line 149: note: this is not a behavioral change .. just making the behavior more apparent by "lifting" the method call of the parameterless `Wait` method

src/Sentry/Threading/ScopedCountdownLock.cs
```diff
@@ -49,6 +49,11 @@ internal ScopedCountdownLock()
     /// </remarks>
     internal CounterScope TryEnterCounterScope()
     {
+        if (IsEngaged)
```
Flash0ver, line 52: note: additional change

no longer allowing to `Add` new logs to a particular _Buffer_, while a `Flush` has been requested (in this case occurring during the `5` second timeout), which may have previously "dragged out" the Flush by still allowing new `Add` operations before the _Event_ is actually set by reaching a count of `0` (while the buffer is not full yet)

this also effects #4425, where - in the case of a terminating unhandled exception - no more `Add` operations are admitted and we only wait until all currently in-progress `Add` operations have concluded in order to do a "safe" `Flush`, which should complete quite quickly, not "unreasonably" dragging out the shutdown

Flash0ver:
@Sentry review

src/Sentry/Threading/ScopedCountdownLock.cs
```diff
@@ -84,13 +91,13 @@ internal LockScope TryEnterLockScope()
...
+
+        if (Interlocked.CompareExchange(ref _isEngaged, 0, 1) != 1)
         {
-            _event.Reset(); // reset the signaled event to the initial count of 1, so that new `CounterScope`s can be entered again
-            return;
+            Debug.Fail("The Lock should have not been disengaged without being engaged first.");
         }
-
-        Debug.Fail("The Lock should have not been disengaged without being engaged first.");
     }
```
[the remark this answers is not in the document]
Flash0ver: It was the original order that lead to a subtle race condition:
- when trying to enter the Lock, we first update `_isEngaged` atomically, before `Signal`ing the event
- when exiting the Lock (can only occur after the Lock has been entered successfully) ...
  - previously: first update `_isEngaged` atomically, before `Reset`ting the event
    - there was a brief window where new Locks can be entered again, but the Event has not been `Reset` yet, leading to an `InvalidOperationException` as we tried to `Signal` an event that has already been set
  - now: first `Reset`, before updating `_isEngaged` atomically
    - new Locks successfully entered are now ensured to have been `Reset` already

bruno-garcia:
bugbot review

PR #4445, opened 2025-08-14
Only detect custom session replay masks when necessary

### LINKED ISSUE #4439: MAUI Performance Issue
### Package

Sentry.Maui

### .NET Flavor

.NET

### .NET Version

9.0.203

### OS

iOS & Android

### OS Version

All? Tested on 17.7.9, 18.0.1 & 18.5

### Development Environment

Visual Studio v17.x

### SDK Version

5.7.0-beta.0

_No response_

### Workload Versions

```
[ios]
   Installation Source: SDK 9.0.200, VS 17.13.35931.197
   Manifest Version:    18.4.9288/9.0.100

[android]
   Installation Source: SDK 9.0.200, VS 17.13.35931.197
   Manifest Version:    35.0.61/9.0.100
```

```
		builder.UseSentry(options =>
		{
			options.AddEventProcessor(new SessionLogEventProcessor());

			options.TracesSampleRate = 0.01;
			options.ProfilesSampleRate = 0.01;

#if ANDROID
			options.Dsn = "***";

			// MT-117470
			options.Android.SuppressSegfaults = true;
#elif IOS
			options.Dsn = "***";
#elif WINDOWS
			options.Dsn = "***";

			var applicationVersion = AppInfo.VersionString;
			applicationVersion = applicationVersion[..applicationVersion.LastIndexOf('.')];
			options.Release = $"{AppInfo.PackageName}@{applicationVersion}+{AppInfo.BuildString}";
#endif
		});
#endif
```

### Steps to Reproduce

I'm working on trying to put together a reproduction project, but there is an issue when navigating back at a specific point in our app. The issue is significantly more pronounced on iOS

### Expected Result

In previous versions the previous page was completely responsive without delay

### Actual Result

As of `5.7.0-beta.0` the previous page becomes unresponsive for a few seconds

jamescrosswell:
Resolves #4439:
- https://github.com/getsentry/sentry-dotnet/issues/4439

### Summary

In applications with complex user interfaces consisting of hundreds of controls on any individual page, the overhead of adding and removing events hooks to all of the `VisualElement.Loaded` event can cause significant performance delays. We use this event hook to detect session replay mask settings for individual VisualElements (i.e. custom masks). 

This PR:
1. Adds an option to disable custom masks
2. Only adds hooks if these might be relevant... so if:
    - Session Replays are enabled
    - The replay options don't simply mask all text/images
    - Custom masks have not been disabled

### DIFF
--- a/src/Sentry.Maui/Internal/MauiCustomSessionReplayMaskBinder.cs
+++ b/src/Sentry.Maui/Internal/MauiCustomSessionReplayMaskBinder.cs
@@ -9,12 +9,12 @@ namespace Sentry.Maui.Internal;
 /// <summary>
 /// Masks or unmasks visual elements for session replay recordings
 /// </summary>
-internal class MauiVisualElementEventsBinder : IMauiElementEventBinder
+internal class MauiCustomSessionReplayMaskBinder : IMauiElementEventBinder
 {
     private readonly SentryMauiOptions _options;
     private readonly bool _isEnabled;
 
-    public MauiVisualElementEventsBinder(IOptions<SentryMauiOptions> options)
+    public MauiCustomSessionReplayMaskBinder(IOptions<SentryMauiOptions> options)
     {
         _options = options.Value;
 #if __ANDROID__
--- a/src/Sentry.Maui/SentryMauiAppBuilderExtensions.cs
+++ b/src/Sentry.Maui/SentryMauiAppBuilderExtensions.cs
@@ -67,7 +67,7 @@ public static class SentryMauiAppBuilderExtensions
         services.AddSingleton<IMauiElementEventBinder, MauiButtonEventsBinder>();
         services.AddSingleton<IMauiElementEventBinder, MauiImageButtonEventsBinder>();
         services.AddSingleton<IMauiElementEventBinder, MauiGestureRecognizerEventsBinder>();
-        services.AddSingleton<IMauiElementEventBinder, MauiVisualElementEventsBinder>();
+        services.AddSingleton<IMauiElementEventBinder, MauiCustomSessionReplayMaskBinder>();
 
         // Resolve the configured options and register any event binders that have been injected by integrations
         var options = new SentryMauiOptions();
--- a/src/Sentry/Platforms/Android/NativeOptions.cs
+++ b/src/Sentry/Platforms/Android/NativeOptions.cs
@@ -272,6 +272,30 @@ public partial class SentryOptions
             public double? SessionSampleRate { get; set; }
             public bool MaskAllImages { get; set; } = true;
             public bool MaskAllText { get; set; } = true;
+
+            /// <summary>
+            /// <para>
+            /// Apps with complex user interfaces, consisting of hundreds of visual controls on a single page, may experience
+            /// performance issues due to the overhead of detecting custom masking of visual elements for Session Replays.
+            /// </para>
+            /// <para>
+            /// In such cases you have a few different options:
+            /// <list type="bullet">
+            ///   <item>
+            ///     <description>Disable Session Replays entirely</description>
+            ///   </item>
+            ///   <item>
+            ///     <description>Mask all text and all images</description>
+            ///   </item>
+            ///   <item>
+            ///     <description>Disable custom session replay masks (so nothing gets masked)</description>
+            ///   </item>
+            /// </list>
+            /// Set this option to <c>true</c> to disable custom session replay masks.
+            /// </para>
+            /// </summary>
+            public bool DisableCustomSessionReplayMasks { get; set; } = false;
+
             internal HashSet<Type> MaskedControls { get; } = [];
             internal HashSet<Type> UnmaskedControls { get; } = [];
 
--- a/src/Sentry.Maui/Internal/MauiVisualElementEventsBinder.cs
+++ b/src/Sentry.Maui/Internal/MauiVisualElementEventsBinder.cs
@@ -12,22 +12,37 @@ namespace Sentry.Maui.Internal;
 internal class MauiVisualElementEventsBinder : IMauiElementEventBinder
 {
     private readonly SentryMauiOptions _options;
+    private readonly bool _isEnabled;
 
     public MauiVisualElementEventsBinder(IOptions<SentryMauiOptions> options)
     {
         _options = options.Value;
+#if __ANDROID__
+        var replayOptions = _options.Native.ExperimentalOptions.SessionReplay;
+        _isEnabled = (replayOptions.SessionSampleRate > 0.0 || replayOptions.OnErrorSampleRate > 0.0)
+            && replayOptions is not { MaskAllImages: true, MaskAllText: true }
+            && !replayOptions.DisableCustomSessionReplayMasks;
+#else
+        _isEnabled = false; // Session replay is only supported on Android for now
+#endif
     }
 
     /// <inheritdoc />
     public void Bind(VisualElement element, Action<BreadcrumbEvent> _)
     {
-        element.Loaded += OnElementLoaded;
+        if (_isEnabled)
+        {
+            element.Loaded += OnElementLoaded;
+        }
     }
 
     /// <inheritdoc />
     public void UnBind(VisualElement element)
     {
-        element.Loaded -= OnElementLoaded;
+        if (_isEnabled)
+        {
+            element.Loaded -= OnElementLoaded;
+        }
     }
 
     internal void OnElementLoaded(object? sender, EventArgs _)
--- a/test/Sentry.Maui.Tests/MauiCustomSessionReplayMaskBinderTests.cs
+++ b/test/Sentry.Maui.Tests/MauiCustomSessionReplayMaskBinderTests.cs
@@ -6,11 +6,11 @@ using View = Android.Views.View;
 
 namespace Sentry.Maui.Tests;
 
-public class MauiVisualElementEventsBinderTests
+public class MauiCustomSessionReplayMaskBinderTests
 {
     private class Fixture
     {
-        public MauiVisualElementEventsBinder Binder { get; }
+        public MauiCustomSessionReplayMaskBinder Binder { get; }
 
         public SentryMauiOptions Options { get; } = new();
 
@@ -21,7 +21,7 @@ public class MauiVisualElementEventsBinderTests
             logger.IsEnabled(Arg.Any<SentryLevel>()).Returns(true);
             Options.DiagnosticLogger = logger;
             var options = Microsoft.Extensions.Options.Options.Create(Options);
-            Binder = new MauiVisualElementEventsBinder(options);
+            Binder = new MauiCustomSessionReplayMaskBinder(options);
         }
     }
 


src/Sentry/Platforms/Android/NativeOptions.cs
```diff
@@ -272,6 +272,30 @@ public class NativeSentryReplayOptions
...
+            ///     <description>Disable Session Replays entirely</description>
+            ///   </item>
+            ///   <item>
+            ///     <description>Mask all text and all images</description>
+            ///   </item>
+            ///   <item>
+            ///     <description>Disable custom session replay masks (so nothing gets masked)</description>
+            ///   </item>
+            /// </list>
+            /// Set this option to <c>true</c> to disable custom session replay masks.
```
bruno-garcia: Why would someone have custom masking (which requires extra code) and then flip this flag?

jamescrosswell: They wouldn't... But we don't know if they've got custom masks without the binding this flag disables. 

At the moment, to use custom masks you just do something like:
```xaml
 <Image 
     Source="dotnet_bot.png" 
     sentry:SessionReplay.Mask="Unmask" 
     SemanticProperties.Description="Cute dot net bot waving hi to you!" 
     HeightRequest="200" 
     HorizontalOptions="Center" /> 
```

... and the `MauiSessionReplayMaskBinder` picks it up automatically. 

An alternative design would be to have custom masks disabled by default... so adding the `sentry:SessionReplay.Mask` (as above) wouldn't do anything, _unless_ you also enabled custom masks in the options.

src/Sentry.Maui/Internal/MauiCustomSessionReplayMaskBinder.cs
```diff
@@ -9,25 +9,40 @@ namespace Sentry.Maui.Internal;
...
+    private readonly bool _isEnabled;
 
-    public MauiVisualElementEventsBinder(IOptions<SentryMauiOptions> options)
+    public MauiCustomSessionReplayMaskBinder(IOptions<SentryMauiOptions> options)
     {
         _options = options.Value;
+#if __ANDROID__
+        var replayOptions = _options.Native.ExperimentalOptions.SessionReplay;
+        _isEnabled = (replayOptions.SessionSampleRate > 0.0 || replayOptions.OnErrorSampleRate > 0.0)
+            && replayOptions is not { MaskAllImages: true, MaskAllText: true }
```
jamescrosswell: Actually I'm not sure about this condition. I think we can...
- _either_ configure `{ MaskAllImages: true, MaskAllText: true }` and override this with the `sentry:SessionReplay.Mask="Unmask"` attribute 
- _or_ configure `{ MaskAllImages: false, MaskAllText: false }` and override this with the `sentry:SessionReplay.Mask="Mask"` attribute

src/Sentry/Platforms/Android/NativeOptions.cs
Flash0ver: We decided to make it an opt-in instead, to not pay the price for project that don't actively use custom masks. So for a new project / projects just starting to use custom masks, the feature must me activated alongside.

Existing applications already using custom masks now need to additionally also opt-in, but the feature is experimental so we're fine with the behavioral change.

jamescrosswell:
@Flash0ver it's lucky I took another look at this on the back of our chat yesterday. I had things completely backwards. See updated description in this PR and updated code.

It's maybe worth considering whether we want to offer the `MaskControlsOfType` and `UnmaskControlsOfType` options at all (given the performance risks/considerations). Although worth noting that these options are [present in the Java implementation](https://github.com/getsentry/sentry-java/blob/ee747ae50ffec79f1ea20d100ea98e8520ab4ed4/sentry/src/main/java/io/sentry/SentryReplayOptions.java#L237-L253)... so it would be nice to offer these for people that don't have crazy user interfaces. They do make things way more conventient:
https://github.com/getsentry/sentry-dotnet/blob/f19a466fa5ce9c20c0ec54b20bc9eb70976bedcd/samples/Sentry.Samples.Maui/MauiProgram.cs#L60

src/Sentry.Maui/SentryMauiAppBuilderExtensions.cs
```diff
@@ -67,11 +67,18 @@ public static MauiAppBuilder UseSentry(this MauiAppBuilder builder,
...
+        // injected by integrations
         var options = new SentryMauiOptions();
         configureOptions?.Invoke(options);
+#if __ANDROID__
+        var replayOptions = options.Native.ExperimentalOptions.SessionReplay;
+        if (replayOptions is { IsSessionReplayEnabled: true, IsTypeMaskingUsed: true })
+        {
+            services.AddSingleton<IMauiElementEventBinder, MauiCustomSessionReplayMaskBinder>();
+        }
+#endif
```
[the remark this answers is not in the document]
jamescrosswell: Not true.

src/Sentry.Maui/SentryMauiAppBuilderExtensions.cs
```diff
@@ -67,11 +67,18 @@ public static MauiAppBuilder UseSentry(this MauiAppBuilder builder,
...
-        services.AddSingleton<IMauiElementEventBinder, MauiVisualElementEventsBinder>();
 
-        // Resolve the configured options and register any event binders that have been injected by integrations
+        // Resolve the configured options and register any event binders that have been enabled via configuration or
+        // injected by integrations
         var options = new SentryMauiOptions();
         configureOptions?.Invoke(options);
+#if __ANDROID__
+        var replayOptions = options.Native.ExperimentalOptions.SessionReplay;
+        if (replayOptions is { IsSessionReplayEnabled: true, IsTypeMaskingUsed: true })
```
jamescrosswell: Potentially, if the backing properties for `IsSessionReplayEnabled` are set via .NET Configuration Bindings, `IsSessionReplayEnabled` could be false here (even though it's true in the options once the ServiceProvider has actually been built).

@Flash0ver thoughts? I could remove it from the pattern. The risk then would be that we might register the maui event binder even though session replay had been explicitly disabled.

I feel like the risk of a false negative for `IsSessionReplayEnabled` is relatively low... and outweighed by the benefits of disabling this event binder when session replay is disabled.

CHANGELOG.md
```diff
@@ -13,6 +13,7 @@
   - Remove `IDisposable` from `SentryStructuredLogger`. Disposal is intended through the owning `IHub` instance ([#4424](https://github.com/getsentry/sentry-dotnet/pull/4424))
   - Ensure all buffered logs are sent to Sentry when the application terminates unexpectedly ([#4425](https://github.com/getsentry/sentry-dotnet/pull/4425))
   - `InvalidOperationException` potentially thrown during a race condition, especially in concurrent high-volume logging scenarios ([#4428](https://github.com/getsentry/sentry-dotnet/pull/4428))
+- Only detects custom session replay masks when necessary to avoid performance issues in MAUI apps with complex UIs ([#4445](https://github.com/getsentry/sentry-dotnet/pull/4445))
```
```suggestion
- Only detect custom Session Replay masks when necessary to avoid performance issues in MAUI apps with complex UIs ([#4445](https://github.com/getsentry/sentry-dotnet/pull/4445))
```

also capitalized "Session Replay" since it's a product/feature name

src/Sentry.Maui/SentryMauiAppBuilderExtensions.cs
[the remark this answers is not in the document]
Flash0ver: Is that untrue because
- the `BindableProperty` does set the tags for native (`""sentry-mask""` and `"sentry-unmask"`) already
- where the `IMauiElementEventBinder` must now do the same for all configured `MaskedControls`/`UnmaskedControls`

?
(just clarifying if I'm understanding it)

src/Sentry.Maui/SentryMauiAppBuilderExtensions.cs
Flash0ver: Hmmm.

If I understand, then,
when masking Controls via `MaskControlsOfType<T>()` and setting `OnErrorSampleRate`/`SessionSampleRate` through Configuration, we would _not_ mask the specified Controls, which could lead to capturing PII unintentionally, which does sound potentially dangerous to me.

I suppose when it comes to potential PII, false masks are preferable to false unmasks.
I guess where I am trying to get at is that a performance optimization should not get in the way of functionality.

Could we perhaps, not remove, but instead move this check into the `Bind`/`UnBind` methods of the `IMauiElementEventBinder`?
Maybe cached as a `Lazy<bool>`, evaluated on first invocation?

So we would be still correct,
yet not register the `IMauiElementEventBinder` if neither `MaskedControls`nor `UnmaskedControls` contain any element.

(btw ... I do like this iteration without a new Option ... "the best API is no API" as the .NET team likes to say ^^)

Flash0ver: Hmmm.

If I understand, then,
when masking Controls via `MaskControlsOfType<T>()` and setting `OnErrorSampleRate`/`SessionSampleRate` through Configuration, we would _not_ mask the specified Controls, which could lead to capturing PII unintentionally, which does sound potentially dangerous to me.

I suppose when it comes to potential PII, false masks are preferable to false unmasks.
I guess where I am trying to get at is that a performance optimization should not get in the way of functionality.

Could we perhaps, not remove, but instead move this check into the `Bind`/`UnBind` methods of the `IMauiElementEventBinder`?
Maybe cached as a `Lazy<bool>`, evaluated on first invocation?

So we would be still correct,
yet not register the `IMauiElementEventBinder` if neither `MaskedControls`nor `UnmaskedControls` contain any element.

(btw ... I do like this iteration without a new Option ... "the best API is no API" as the .NET team likes to say ^^)
also, I like the feature-compat with Java, considering the `MaskedControls`/`UnmaskedControls`

jamescrosswell: > when masking Controls via MaskControlsOfType<T>() and setting OnErrorSampleRate/SessionSampleRate through Configuration, we would not mask the specified Controls, which could lead to capturing PII unintentionally, which does sound potentially dangerous to me.

Correct. 

> Could we perhaps, not remove, but instead move this check into the Bind/UnBind methods of the IMauiElementEventBinder?
Maybe cached as a Lazy<bool>, evaluated on first invocation?

We can't do any of those no. The performance impact was originally detected here:
https://github.com/getsentry/sentry-dotnet/blob/85c56e2d2e72042889162b120ab89e2099141372/src/Sentry.Maui/Internal/MauiVisualElementEventsBinder.cs#L28-L31

So simply enumerating all of the (many hundreds) of visual elements and removing an event hook (without any complex comparisons) incurs a significant performance penalty. If we do any extra processing in the Bind/Unbind methods, I think it would only make the problem worse. 

### Alternative solution

An alternative solution would be to remove the ability to use configuration bindings with the `NativeSentryReplayOptions` at all... so the only way to set these is via some kind of options callback. For example:

```csharp
options.EnableSessionReplay(replayOptions => {
    replayOptions.SessionSampleRate = 0.1;
    replayOptions.OnErrorSampleRate = 1.0;
    replayOptions.RedactAllText = true;
});
```

src/Sentry.Maui/SentryMauiAppBuilderExtensions.cs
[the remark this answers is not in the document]
jamescrosswell: The XML attribute-based session replay masking works just fine without this MAUI binder. It's taken care of by our custom bindable properties (which are always enabled):
https://github.com/getsentry/sentry-dotnet/blob/85c56e2d2e72042889162b120ab89e2099141372/src/Sentry.Maui/SessionReplay.cs#L22-L31

Under the hood we use these to set the native android tags that are used by the Java SDK in the `OnUnmaskedElementHandlerChanged` and `OnMaskedElementHandlerChanged` handlers.

src/Sentry.Maui/SentryMauiAppBuilderExtensions.cs
Flash0ver: chat:
Perhaps we could do a type-check / internal-marker-interface check,
to read from the Sentry-MAUI-Options (via `IOptions<T>` at the time of `Bind`.

If this does not work or yield the perf gain we want,
we'll go with the _Alternative solution_ from above.

jamescrosswell: The workaround we discussed works!

CHANGELOG.md
```diff
@@ -14,6 +14,7 @@
   - Ensure all buffered logs are sent to Sentry when the application terminates unexpectedly ([#4425](https://github.com/getsentry/sentry-dotnet/pull/4425))
   - `InvalidOperationException` potentially thrown during a race condition, especially in concurrent high-volume logging scenarios ([#4428](https://github.com/getsentry/sentry-dotnet/pull/4428))
 - Blocking calls are no longer treated as unhandled crashes ([#4458](https://github.com/getsentry/sentry-dotnet/pull/4458))
+- Only applies Session Replay masks to specific controls types when necessary, to avoid performance issues in MAUI apps with complex UIs ([#4445](https://github.com/getsentry/sentry-dotnet/pull/4445))
```
```suggestion
- Only apply Session Replay masks to specific control types when necessary to avoid performance issues in MAUI apps with complex UIs ([#4445](https://github.com/getsentry/sentry-dotnet/pull/4445))
```

PR #4452, opened 2025-08-19
Backpressure

### LINKED ISSUE #2410: Improved background worker
### Problem Statement

The internal [`BackgroundWorker`](https://github.com/getsentry/sentry-dotnet/blob/3a9222753316f49fc5114c8b0e8d1f298ad82c4b/src/Sentry/Internal/BackgroundWorker.cs) class manages back-pressure for the SDK.  When events, transactions, sessions, etc. are captured, the `SentryClient` adds them to a `ConcurrentQueue` within the background worker.  The background worker then has a separate task that is responsible for reading from the queue and sending events to the `ITransport` implementation (typically that's the `HttpTransport`).

This is generally a good design.  However, it suffers from the limitation that there's only one queue for all envelope types, non-prioritized, with a single limit set by `SentryOptions.MaxQueueItems` - defaulting to 30.  If the queue is full, envelopes are dropped when captured, never being added to the queue.  Therefore, it's quite possible (especially in a server environment) that a large flood of one type of envelope (such as transactions) can prevent another more important type of envelope (such as error events) from being sent to Sentry.

### Solution Brainstorm

We need to support concurrency on the producer side, so the built-in `PriortyQueue` class is out.  We need to be non-blocking on the consumer side, so `BlockingCollection` is also out.  There is no built-in `PriortyConcurrentQueue` class.  Thus, to implement prioritization, we will probably need to use _multiple_ `ConcurrentQueue` instances in the background worker - perhaps in a dictionary where the key is the envelope type.

We also should have more than one option for controlling the maximum queue size (this might be separate properties, or a dictionary).  Currently we just have `MaxQueueItems` - which is undocumented.

We should probably set higher default queue sizes for ASP.NET and ASP.NET Core.  Currently, `MaxQueueSize` defaults to 30.  We do show setting it higher in some of the sample apps, but with no explanation.

Whatever is decided in https://github.com/getsentry/team-webplatform-meta/issues/50 should be taken into consideration also.

jamescrosswell:
Resolves #2410
- https://github.com/getsentry/sentry-dotnet/issues/2410

---

> [!NOTE]
> <sup>[Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) is generating a summary for commit 415308df430dbf508a9ed5146b9816e3042f1d54. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


src/Sentry/Internal/BackpressureMonitor.cs
```diff
@@ -0,0 +1,158 @@
...
+
+    internal void IncrementDownsampleLevel()
+    {
+        var oldValue = _downsampleLevel;
+        if (oldValue < MaxDownsamples)
+        {
+            var newValue = oldValue + 1;
+            if (Interlocked.CompareExchange(ref _downsampleLevel, newValue, oldValue) == oldValue)
+            {
+                _logger?.LogDebug("System is under pressure, increasing downsample level to {DownsampleLevel}.", newValue);
```
bruno-garcia: if this is our diagnostic logger, it only support positional template since it uses string.Format
```suggestion
                _logger?.LogDebug("System is under pressure, increasing downsample level to {0}.", newValue);
```

src/Sentry/Internal/BackpressureMonitor.cs
```diff
@@ -0,0 +1,158 @@
...
+            var nowTicks = _clock.GetUtcNow().Ticks;
+            var recentOverflowCutoff = nowTicks - RecencyThresholdTicks;
+            var rateLimitCutoff = nowTicks - CheckIntervalTicks;
+            return LastQueueOverflowTicks < recentOverflowCutoff && LastRateLimitEventTicks < rateLimitCutoff;
+        }
+    }
+
+    public void Dispose()
+    {
+        _cts.Cancel();
```
bruno-garcia: Might wnt to wrap this whole method on a try catch:

https://learn.microsoft.com/en-us/dotnet/api/system.threading.cancellationtokensource.cancel?view=net-9.0#system-threading-cancellationtokensource-cancel

src/Sentry/SentryOptions.cs
```diff
@@ -94,6 +94,20 @@ public bool IsGlobalModeEnabled
     /// </summary>
     public bool EnableScopeSync { get; set; }
 
+    /// <summary>
+    /// Enables or disables automatic backpressure handling. When enabled, the SDK will monitor system health and
+    /// reduce the sampling rate of events and transactions when the system is under load.
+    /// </summary>
+    public bool EnableBackpressureHandling { get; set; } = false;
```
bruno-garcia, line 101: will it eventually be 'true' by default or just while we validate it? if this is disabled forever by default it wont be of much value

jamescrosswell: Is the benefit mainly for Sentry here?

I think what the Java SDK did was have this optional initially, until they got some feedback from SDK users with high throughput scenarios, then they turned it on by default when they were happy with it.

src/Sentry/Internal/Hub.cs
```diff
@@ -201,7 +205,7 @@ internal ITransactionTracer StartTransaction(
         // finally fallback to Random sampling if the decision has been made by no other means
         if (isSampled == null)
         {
-            sampleRate = _options.TracesSampleRate ?? 0.0;
+            sampleRate = (_options.TracesSampleRate ?? 0.0) * _options.BackpressureMonitor.GetDownsampleFactor();
```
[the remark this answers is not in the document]
jamescrosswell: I think this is probably what we want to happen. If we propagated the sample rate after downsampling, and downstream participants in the trace also implemented backpressure handling, after one or two hops sampling would grind to a halt.

For example in a trace from Node1 -> Node2 -> Node3:
- Initial Sample rate for all three participants is 1.0
- Node1 applies a downsampling factor of 1/1024 (effective rate is 1/1024)... propagates the effective rate
- Node2 applies a downsampling factor of 1/1024 (effective rate is 1/1,048,576)... propagates the effective rate 
- Node3 applies a downsampling factor of 1/1024... effective sample rate is now less than 1 in a billion

src/Sentry/SentryOptions.cs
Flash0ver: I'm happy with either .. enabling by default right away or later.

With `net10.0` and our 6.0 Release on the horizon, we could do an opt-in for now, and enable per default in 6.0.
The "pain point" there is that we need to update the docs then again.

src/Sentry/Internal/BackpressureMonitor.cs
```diff
@@ -0,0 +1,169 @@
...
+/// </para>
+/// <para>
+/// The heath checks used by the monitor are:
+/// </para>
+/// <list type="number">
+///     <item>if any events have been dropped due to queue being full in the last 2 seconds</item>
+///     <item>if any new rate limits have been applied since the last check</item>
+/// </list>
+/// This check is performed every 10 seconds. With each negative health check we halve tracesSampleRate up to 10 times, meaning the original tracesSampleRate is multiplied by 1, 1/2, 1/4, ... up to 1/1024 (~ 0.001%). Any positive health check resets to the original tracesSampleRate set in SentryOptions.
+/// </summary>
```
Flash0ver, line 19:
```suggestion
/// <summary>
/// <para>
/// Monitors system health and calculates a DownsampleFactor that can be applied to events and transactions when the
/// system is under load.
/// </para>
/// <para>
/// The health checks used by the monitor are:
/// </para>
/// <list type="number">
///     <item>if any events have been dropped due to queue being full in the last 2 seconds</item>
///     <item>if any new rate limits have been applied since the last check</item>
/// </list>
/// This check is performed every 10 seconds. With each negative health check we halve tracesSampleRate up to 10 times, meaning the original tracesSampleRate is multiplied by 1, 1/2, 1/4, ... up to 1/1024 (~ 0.001%). Any positive health check resets to the original tracesSampleRate set in SentryOptions.
/// </summary>
/// <seealso href="https://develop.sentry.dev/sdk/telemetry/traces/backpressure/">Backpressure Management</seealso>
```

src/Sentry/Internal/BackpressureMonitor.cs
```diff
@@ -0,0 +1,169 @@
+using Sentry.Extensibility;
+using Sentry.Infrastructure;
+
+namespace Sentry.Internal;
+
+/// <summary>
+/// <para>
+/// Monitors system health and calculates a DownsampleFactor that can be applied to events and transaction when the
```
Flash0ver:
```suggestion
/// Monitors system health and calculates a DownsampleFactor that can be applied to events and transactions when the
```

src/Sentry/Internal/BackpressureMonitor.cs
```diff
@@ -0,0 +1,169 @@
...
+/// system is under load.
+/// </para>
+/// <para>
+/// The heath checks used by the monitor are:
```
Flash0ver:
```suggestion
/// The health checks used by the monitor are:
```

src/Sentry/Internal/BackpressureMonitor.cs
```diff
@@ -0,0 +1,169 @@
...
+/// </para>
+/// <list type="number">
+///     <item>if any events have been dropped due to queue being full in the last 2 seconds</item>
+///     <item>if any new rate limits have been applied since the last check</item>
+/// </list>
+/// This check is performed every 10 seconds. With each negative health check we halve tracesSampleRate up to 10 times, meaning the original tracesSampleRate is multiplied by 1, 1/2, 1/4, ... up to 1/1024 (~ 0.001%). Any positive health check resets to the original tracesSampleRate set in SentryOptions.
+/// </summary>
+internal class BackpressureMonitor : IDisposable
```
Flash0ver, line 21:
```suggestion
/// </summary>
/// <seealso href="https://develop.sentry.dev/sdk/telemetry/traces/backpressure/">Backpressure Management</seealso>
internal class BackpressureMonitor : IDisposable
```

src/Sentry/Internal/BackpressureMonitor.cs
```diff
@@ -0,0 +1,169 @@
...
+
+    private async Task DoWorkAsync(CancellationToken cancellationToken)
+    {
+        try
+        {
+            while (!cancellationToken.IsCancellationRequested)
+            {
+                DoHealthCheck();
+
+                // Your periodic work here
```
Flash0ver: suggestion: remove comment

I don't think this comment adds any value ... as the periodic work is done above

src/Sentry/Internal/BackpressureMonitor.cs
```diff
@@ -0,0 +1,169 @@
...
+            }
+            finally
+            {
+                _cts.Dispose();
+            }
+        }
+        catch (Exception ex)
+        {
+            // We don't ever want this to throw
+            _logger?.LogDebug("Error in BackpressureMonitor.Dispose: {0}", ex);
```
Flash0ver: question: Level

I wonder if a higher level could be more appropriate, like _Warning_, since we don't expect this flow to occur.

src/Sentry/Internal/BackpressureMonitor.cs
```diff
@@ -0,0 +1,169 @@
...
+            var rateLimitCutoff = nowTicks - CheckIntervalTicks;
+            return LastQueueOverflowTicks < recentOverflowCutoff && LastRateLimitEventTicks < rateLimitCutoff;
+        }
+    }
+
+    public void Dispose()
+    {
+        try
+        {
+            try
```
Flash0ver: question: Is the double-try intentional?

I wonder if the double-try gives any benefit.
In the inner `finally`, we `Dispose` the CTS. But a `Dispose` should never throw.

I wonder if moving the outer `catch (Exception ex)` to after the inner `catch (AggregateException ex) when` would make the code a bit clearer.

test/Sentry.Tests/Internals/BackpressureMonitorTests.cs
```diff
@@ -0,0 +1,144 @@
...
+
+        // Act
+        monitor.RecordRateLimitHit(_fixture.Now);
+
+        // Assert
+        monitor.IsHealthy.Should().BeFalse();
+    }
+
+    [Fact]
+    public void DownsampleLevel_Increases_WhenUnhealthy()
```
Flash0ver: question: Max-Level

Is it worthy to have a test that verifies the Max-Downsampling-Level of 10?

src/Sentry/SentryOptions.cs
jamescrosswell: I've added an issue to make sure we circle back to this (by v6.0 at the latest):
- https://github.com/getsentry/sentry-dotnet/issues/4494

test/Sentry.Tests/SentryClientTests.cs
```diff
@@ -624,17 +624,29 @@ public void CaptureEvent_SamplingNull_DropsEvent()
...
+    [InlineData(0.75f, 1)]
+    [InlineData(0.25f, 3)]
+    [InlineData(0.50f, 3)]
+    [InlineData(0.75f, 3)]
+    public void CaptureEvent_WithSampleRate_AppropriateDistribution(float sampleRate, int downsampleLevel)
     {
         // Arrange
+        var now = DateTimeOffset.UtcNow;
+        var clock = new MockClock(now);
+        var backpressureMonitor = new BackpressureMonitor(null, clock, startImmediately: false);
```
Flash0ver:
```suggestion
        using var backpressureMonitor = new BackpressureMonitor(null, clock, startImmediately: false);
```

not really necessary, but to be consistent with the other tests

src/Sentry/SentryOptions.cs
```diff
@@ -94,6 +94,20 @@ public bool IsGlobalModeEnabled
...
+
+    /// <summary>
+    /// Stores a reference to the BackpressureMonitor for use by the rest of the SDK.
+    /// </summary>
+    /// <remarks>Note: this is necessary since the BackpressureMonitor is used by <see cref="HttpTransportBase"/>,
```
Flash0ver: question: taking the breaking change in 6.0?

Could this be something that we'd like to change with the next major Release?
Or is it fine as is, since we also retrieve the `ClientReportRecorder` in `HttpTransportBase` from the `_options`? Because similar to the `ClientReportRecorder` being configurable via `SentryOptions.SendClientReports`, the new `BackpressureMonitor` is configurable via `SentryOptions.EnableBackpressureHandling`.

In other words
- do we want to open an issue for Milestone 6.0 to move this to a .ctor-dependency
- or do we want to remove this comment as the design is consistent with our current architecture?

src/Sentry/Internal/BackpressureMonitor.cs
```diff
@@ -0,0 +1,166 @@
...
+    private long _lastQueueOverflow = DateTimeOffset.MinValue.Ticks;
+    private long _lastRateLimitEvent = DateTimeOffset.MinValue.Ticks;
+    private volatile int _downsampleLevel = 0;
+
+    private static readonly long RecencyThresholdTicks = TimeSpan.FromSeconds(RecentThresholdInSeconds).Ticks;
+    private static readonly long CheckIntervalTicks = TimeSpan.FromSeconds(CheckIntervalInSeconds).Ticks;
+
+    private readonly CancellationTokenSource _cts = new();
+
+    private Task WorkerTask { get; }
```
Flash0ver: suggestion (non-blocking): I believe this could be a `private readonly` field instead.

src/Sentry/Internal/BackpressureMonitor.cs
```diff
@@ -0,0 +1,166 @@
...
+    internal int DownsampleLevel => _downsampleLevel;
+    internal long LastQueueOverflowTicks => Interlocked.Read(ref _lastQueueOverflow);
+    internal long LastRateLimitEventTicks => Interlocked.Read(ref _lastRateLimitEvent);
+
+    public BackpressureMonitor(IDiagnosticLogger? logger, ISystemClock? clock = null, bool startImmediately = true)
```
Flash0ver:
```diff
-bool startImmediately = true
+bool enablePeriodicHealthCheck = true
```

To me `startImmediately` is a bit misleading, as there is no way to start the Periodic-Health-Check after construction of the Monitor.

I suggest to rename `startImmediately` to something like
- `enablePeriodicHealthCheck`
- `startPeriodicHealthCheck`
- `withPeriodicHealthCheck`

or similar.

src/Sentry/Internal/BackpressureMonitor.cs
```diff
@@ -0,0 +1,166 @@
...
+            var newValue = oldValue + 1;
+            if (Interlocked.CompareExchange(ref _downsampleLevel, newValue, oldValue) == oldValue)
+            {
+                _logger?.LogDebug("System is under pressure, increasing downsample level to {0}.", newValue);
+            }
+        }
+    }
+
+    /// <summary>
+    /// A multiplier that can be applied to  the SampleRate or TracesSampleRate to reduce the amount of data sent to
```
Flash0ver:
```suggestion
    /// A multiplier that can be applied to the SampleRate or TracesSampleRate to reduce the amount of data sent to
```

src/Sentry/Internal/BackpressureMonitor.cs
```diff
@@ -0,0 +1,166 @@
...
+        {
+            IncrementDownsampleLevel();
+        }
+    }
+
+    /// <summary>
+    /// Checks for any recent queue overflows or any rate limit events since the last check.
+    /// </summary>
+    /// <returns></returns>
+    internal bool IsHealthy
```
Flash0ver, line 134: thought: method rather than property

I'm wondering if this should be a method instead of a property.
As property, I would expect it to not be time-dependent.

But since it's internal, it may also be more of a stylistic choice.
I'm also fine with keeping it a Property.

src/Sentry/SentryClient.cs
```diff
@@ -375,16 +374,18 @@ private SentryId DoSendEvent(SentryEvent @event, SentryHint? hint, Scope? scope)
 
         if (_options.SampleRate != null)
         {
-            if (!_randomValuesFactory.NextBool(_options.SampleRate.Value))
+            var downsampleFactor = _options.BackpressureMonitor.GetDownsampleFactor();
+            var sampleRate = _options.SampleRate.Value * downsampleFactor;
+            if (!_randomValuesFactory.NextBool(sampleRate))
             {
                 _options.ClientReportRecorder.RecordDiscardedEvent(DiscardReason.SampleRate, DataCategory.Error);
```
Flash0ver: suggestion: Client Report and Discard Reason

Should we, when the `_options.BackpressureMonitor.DownsampleLevel != 0`, instead of the `DiscardReason.SampleRate` use the `DiscardReason.Backpressure`?

But there may be a race condition, when evaluating both the `DownsampleLevel` as well as the `DownsampleFactor`, should the Periodic-Health-Check run concurrently.

see https://develop.sentry.dev/sdk/telemetry/traces/backpressure/
> #### Client Report
> If possible, in `transaction.finish`, also record a client report with reason `backpressure` instead of `sample_rate` when the transaction is dropped so that we can track these backpressure outcome statistics.

Flash0ver: and ... are there more locations to consider the Discard Reason?
e.g. method `SentryClient.CaptureTransaction`

src/Sentry/SentryOptions.cs
jamescrosswell: Sure, we can add an issue to v6.0.0... probably nicer to inject it via the constructor.

jamescrosswell: Added:
- https://github.com/getsentry/sentry-dotnet/issues/4534

src/Sentry/Internal/BackpressureMonitor.cs
jamescrosswell: I've left this as a property (given the name, I think it would be expected that it may return true or false at different times).

src/Sentry/Internal/Hub.cs
```diff
@@ -845,6 +867,12 @@ public void Dispose()
         }
         //Don't dispose of ScopeManager since we want dangling transactions to still be able to access tags.
 
+        if (_options.BackpressureMonitor is { } backpressureMonitor)
+        {
+            _options.BackpressureMonitor = null;
+            backpressureMonitor.Dispose();
+        }
+
```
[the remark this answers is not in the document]
jamescrosswell: Hm, I guess we have to do reference counting and only dispose of it when all the references have been disposed or something (our own bespoke little GC implementation).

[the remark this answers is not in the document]
jamescrosswell: Refactored.

src/Sentry/Internal/BackpressureMonitor.cs
```diff
@@ -0,0 +1,166 @@
...
+            _workerTask.Wait();
+        }
+        catch (AggregateException ex) when (ex.InnerException is OperationCanceledException)
+        {
+            // Ignore cancellation
+        }
+        catch (Exception ex)
+        {
+            // Log rather than throw
+            _logger?.LogWarning("Error in BackpressureMonitor.Dispose: {0}", ex);
```
bruno-garcia: first arg should take Exception. The other overloads just call ToString on positional arguments
```suggestion
            _logger?.LogWarning(ex, "Error in BackpressureMonitor.Dispose");
```

src/Sentry/SentryClient.cs
```diff
@@ -375,16 +377,22 @@ private SentryId DoSendEvent(SentryEvent @event, SentryHint? hint, Scope? scope)
...
+            var downsampledRate = sampleRate * _backpressureMonitor.GetDownsampleFactor();
+            var sampleRand = _randomValuesFactory.NextDouble();
+            if (sampleRand >= downsampledRate)
             {
-                _options.ClientReportRecorder.RecordDiscardedEvent(DiscardReason.SampleRate, DataCategory.Error);
-                _options.LogDebug("Event sampled.");
+                // If sampling out is only a result of the downsampling then we specify the reason as backpressure
+                // management... otherwise the event would have been sampled out anyway, so it's just regular sampling.
+                var reason = sampleRand < sampleRate ? DiscardReason.Backpressure : DiscardReason.SampleRate;
+                _options.ClientReportRecorder.RecordDiscardedEvent(reason, DataCategory.Error);
```
bruno-garcia, line 388: Is `DataCategory.Error` by design?

Seems like it'll conflate meaning. Above we're dealing with Backpressure vs SampleRate

jamescrosswell: I haven't changed that in this PR. It comes from the main branch:
https://github.com/getsentry/sentry-dotnet/blob/58b5da05bfea6ab8615aa0a12a56275d665c76a2/src/Sentry/SentryClient.cs#L380

Seems like it always gets set to Transaction/Span in the CaptureTransaction method and otherwise it's always assumed to be Error:
https://github.com/getsentry/sentry-dotnet/blob/58b5da05bfea6ab8615aa0a12a56275d665c76a2/src/Sentry/SentryClient.cs#L168-L169

This PR doesn't change the data category (the kind of event being discarded)... it just changes the reason the event is being discarded. (sampling vs backpressure).

CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
+
+### Features
+
+- Added an option for Automatic backpressure handling. When enabled this automatically reduces the sample rate when the SDK detects events being dropped. ([#4452](https://github.com/getsentry/sentry-dotnet/pull/4452))
```
```suggestion
- Added `EnableBackpressureHandling` option for Automatic backpressure handling. When enabled this automatically reduces the sample rate when the SDK detects events being dropped. ([#4452](https://github.com/getsentry/sentry-dotnet/pull/4452))
```

As convenience for users to see it from the changelog already, without needing to consult the docs or the changeset.

src/Sentry/Http/HttpTransportBase.cs
```diff
@@ -42,6 +43,24 @@ protected HttpTransportBase(SentryOptions options,
...
+    /// </summary>
+    /// <param name="options">The Sentry options.</param>
+    /// <param name="backpressureMonitor">The Sentry options.</param>
+    /// <param name="getEnvironmentVariable">An optional method used to read environment variables.</param>
+    /// <param name="clock">An optional system clock - used for testing.</param>
+    internal HttpTransportBase(SentryOptions options, BackpressureMonitor? backpressureMonitor,
+        Func<string, string?>? getEnvironmentVariable = default,
+        ISystemClock? clock = default)
+    {
+        _options = options;
```
Flash0ver: suggestion (non-blocking): deduplicate

We could deduplicate a bit of code by having this .ctor calling into the other .ctor with `this`,
and only set the additional `_backpressureMonitor = backpressureMonitor;` in this new .ctor.

Optional ... we could also refactor this later ... there is a follow-up issue that we haven't closed yet which we could use for that.

src/Sentry/SentryOptions.cs
```diff
@@ -94,6 +94,12 @@ public bool IsGlobalModeEnabled
     /// </summary>
     public bool EnableScopeSync { get; set; }
 
+    /// <summary>
+    /// Enables or disables automatic backpressure handling. When enabled, the SDK will monitor system health and
+    /// reduce the sampling rate of events and transactions when the system is under load.
+    /// </summary>
+    public bool EnableBackpressureHandling { get; set; } = false;
```
Flash0ver, line 101: todo: add new option to docs: https://docs.sentry.io/platforms/dotnet/configuration/options/

jamescrosswell: https://github.com/getsentry/sentry-dotnet/issues/4563

Flash0ver:
✅ thanks

PR #4453, opened 2025-08-20
perf: Avoid shallow copies of DSC

### LINKED ISSUE #4381: perf: Avoid shallow copies of DSC when `sample_rate` or `replay_id` are overwritten
### Description

So far, the [Dynamic Sampling Context (DSC)](https://develop.sentry.dev/sdk/telemetry/traces/dynamic-sampling-context/) was immutable after the first propagation.

But in the meantime, immutability only partially applies, after
- #4133
- #4374

For both cases, we allocate a shallow copy of the `DynamicSamplingContext` instance, and the underlying `IReadOnlyDictionary`.
Since `DynamicSamplingContext` is an `internal` class, we can optimize the implementation details to avoid copy allocations, and instead mutate the internal `Dictionary` that we create in the _.ctor_.

Alternatively, we could rewrite or derive a `DynamicSamplingContext`, and/or introduce something similar like a `DynamicSamplingContextBuilder`. But since the `DynamicSamplingContext` is `internal` this may not be required.

jamescrosswell:
Resolves #4381
- https://github.com/getsentry/sentry-dotnet/issues/4381

#skip-changelog

> [!NOTE]
> A bunch of the tests had to be changed as they assumed the setting the sample rate or replay id on a DSC would return a copy of the DSC (not modify the original DSC). Our own code doesn't need to make such assumptions (or preserve the "original" DSC for future comparisons, as the tests are doing).

### DIFF
--- a/src/Sentry/DynamicSamplingContext.cs
+++ b/src/Sentry/DynamicSamplingContext.cs
@@ -105,7 +105,7 @@ internal class DynamicSamplingContext
         _items["sample_rate"] = sampleRate.ToString(CultureInfo.InvariantCulture);
     }
 
-    internal DynamicSamplingContext Clone() => new(new Dictionary<string, string>(Items));
+    internal DynamicSamplingContext Clone() => new(new Dictionary<string, string>(_items));
 
     public void SetReplayId(IReplaySession? replaySession)
     {
--- a/src/Sentry/DynamicSamplingContext.cs
+++ b/src/Sentry/DynamicSamplingContext.cs
@@ -9,16 +9,16 @@ namespace Sentry;
 /// <seealso href="https://develop.sentry.dev/sdk/performance/dynamic-sampling-context"/>
 internal class DynamicSamplingContext
 {
-    public IReadOnlyDictionary<string, string> Items { get; }
+    public Dictionary<string, string> Items { get; }
 
     public bool IsEmpty => Items.Count == 0;
 
-    private DynamicSamplingContext(IReadOnlyDictionary<string, string> items) => Items = items;
+    private DynamicSamplingContext(Dictionary<string, string> items) => Items = items;
 
     /// <summary>
     /// Gets an empty <see cref="DynamicSamplingContext"/> that can be used to "freeze" the DSC on a transaction.
     /// </summary>
-    public static readonly DynamicSamplingContext Empty = new(new Dictionary<string, string>().AsReadOnly());
+    public static DynamicSamplingContext Empty => new(new Dictionary<string, string>());
 
     private DynamicSamplingContext(SentryId traceId,
         string publicKey,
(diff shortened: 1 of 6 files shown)


src/Sentry/DynamicSamplingContext.cs
```diff
@@ -9,16 +9,16 @@ namespace Sentry;
 /// <seealso href="https://develop.sentry.dev/sdk/performance/dynamic-sampling-context"/>
 internal class DynamicSamplingContext
 {
-    public IReadOnlyDictionary<string, string> Items { get; }
+    public Dictionary<string, string> Items { get; }
```
Flash0ver: thought: keep the property as ``IReadOnlyDictionary`2``

I am wondering if we should keep this property unchanged,
and instead introduce an explicit `private` backing field, typed ``Dictionary`2``,
which we only mutate through the `Set..` methods provided.

This way we would control the mutations of the DSC a little bit more.

src/Sentry/DynamicSamplingContext.cs
```diff
@@ -9,16 +9,16 @@ namespace Sentry;
...
     public bool IsEmpty => Items.Count == 0;
 
-    private DynamicSamplingContext(IReadOnlyDictionary<string, string> items) => Items = items;
+    private DynamicSamplingContext(Dictionary<string, string> items) => Items = items;
 
     /// <summary>
     /// Gets an empty <see cref="DynamicSamplingContext"/> that can be used to "freeze" the DSC on a transaction.
     /// </summary>
-    public static readonly DynamicSamplingContext Empty = new(new Dictionary<string, string>().AsReadOnly());
+    public static DynamicSamplingContext Empty => new(new Dictionary<string, string>());
```
Flash0ver: thought: looks like a shared instance, but is a factory

This property looks similar to `SentryId.Empty` and `SpanId.Empty`, appearing to be a shared "empty" instance (I briefly thought we had a bug where we may end up mutating this shared instance).
But it is actually a Factory, creating a new instance.

I'd prefer if we'd change this factory property to a factory method:
e.g. `Empty()` or `CreateEmpty()`

My care factor is not too high though.

src/Sentry/DynamicSamplingContext.cs
```diff
@@ -98,34 +98,19 @@ private DynamicSamplingContext(SentryId traceId,
...
-                {
-                    var items = Items.ToDictionary(kvp => kvp.Key, kvp => kvp.Value);
-                    items["sample_rate"] = sampleRate.ToString(CultureInfo.InvariantCulture);
-                    return new DynamicSamplingContext(items);
-                }
-            }
-        }
-
-        return this;
+        Items["sample_rate"] = sampleRate.ToString(CultureInfo.InvariantCulture);
```
Flash0ver: question: change of behavior

The previous implementation was updating the Sample-Rate, only if it already existed.
The new implementation now does Set-or-Add the Sample-Rate.

Apparently, I didn't add a test for that behavior.
I wonder which behavior we want?
See also #4374, where initially introduced.

test/Sentry.Tests/HubTests.cs
```diff
@@ -759,12 +760,12 @@ public void StartTransaction_DynamicSamplingContextWithSampleRate_OverwritesSamp
             transactionTracer.SampleRate.Should().Be(expectedSampleRate);
             if (expectedDscOverwritten)
             {
-                transactionTracer.DynamicSamplingContext.Should().NotBeSameAs(dsc);
-                transactionTracer.DynamicSamplingContext.Should().BeEquivalentTo(dsc.ReplaceSampleRate(expectedSampleRate));
+                transactionTracer.DynamicSamplingContext.Should().NotBeSameAs(originalDsc);
+                transactionTracer.DynamicSamplingContext.Should().BeEquivalentTo(originalDsc.CloneWithSampleRate(expectedSampleRate));
```
Flash0ver, line 763: suggestion: remove/replace Reference-Equality-Checks

Since the DSC is no longer immutable,
I have the feeling that Reference-Equality-Checks could hide problems in cases where we reuse the same instance, but it's values have either changed or not changed unexpectedly.

I suggest replacing all `NotBeSameAs` and definitely `BeSameAs` assertions
with `BeEquivalentTo` assertions instead.

As a side effect, I believe we can remove some `var originalDsc = dsc.Clone();` copies
when doing equivalency checks, only.

What do you think?

src/Sentry/DynamicSamplingContext.cs
jamescrosswell: From the uses of this method (there are only 2), it looks like it's doing what it's supposed to.

Flash0ver (approved):

PR #4457, opened 2025-08-21
chore: Added DebugLoggerLevel to DebugLogger delegate

### LINKED ISSUE #4382: Refactor AssemblyReaders to use our  `IDiagnosticLogger`
The code we vendored in to create the Android Assembly Readers currently uses a `DebugLogger` delegate: 
https://github.com/getsentry/sentry-dotnet/blob/4ee96eeaf000d8f1d426a914622548a473616bd3/src/Sentry.Android.AssemblyReader/DebugLogger.cs#L8

The existing delegate doesn't have any notion of "log levels", so is a bit limited.

- [ ] We should refactor this code to use our  `IDiagnosticLogger` instead.

_Originally posted by @jamescrosswell in https://github.com/getsentry/sentry-dotnet/pull/4294#discussion_r2224947579_

jamescrosswell:
Resolves #4382
- https://github.com/getsentry/sentry-dotnet/issues/4382

#skip-changelog

### DIFF
--- a/src/Sentry.Android.AssemblyReader/DebugLogger.cs
+++ b/src/Sentry.Android.AssemblyReader/DebugLogger.cs
@@ -3,6 +3,7 @@ namespace Sentry.Android.AssemblyReader;
 /// <summary>
 /// Writes a log message for debugging.
 /// </summary>
+/// <param name="level">The debug log level.</param>
 /// <param name="message">The message string to write.</param>
 /// <param name="args">Arguments for the formatted message string.</param>
-public delegate void DebugLogger(string message, params object?[] args);
+public delegate void DebugLogger(DebugLoggerLevel level, string message, params object?[] args);
--- a/src/Sentry.Android.AssemblyReader/DebugLoggerLevel.cs
+++ b/src/Sentry.Android.AssemblyReader/DebugLoggerLevel.cs
@@ -1,0 +1,32 @@
+namespace Sentry.Android.AssemblyReader;
+
+/// <summary>
+/// Represents the level of debug logging.
+/// </summary>
+public enum DebugLoggerLevel : short
+{
+    /// <summary>
+    /// Debug level logging.
+    /// </summary>
+    Debug,
+
+    /// <summary>
+    /// Information level logging.
+    /// </summary>
+    Info,
+
+    /// <summary>
+    /// Warning level logging.
+    /// </summary>
+    Warning,
+
+    /// <summary>
+    /// Error level logging.
+    /// </summary>
+    Error,
+
+    /// <summary>
+    /// Fatal level logging.
+    /// </summary>
+    Fatal
+}
--- a/src/Sentry.Android.AssemblyReader/AndroidAssemblyReaderFactory.cs
+++ b/src/Sentry.Android.AssemblyReader/AndroidAssemblyReaderFactory.cs
@@ -17,29 +17,29 @@ public static class AndroidAssemblyReaderFactory
     /// <returns>The reader</returns>
     public static IAndroidAssemblyReader Open(string apkPath, IList<string> supportedAbis, DebugLogger? logger = null)
     {
-        logger?.Invoke("Opening APK: {0}", apkPath);
+        logger?.Invoke(DebugLoggerLevel.Debug, "Opening APK: {0}", apkPath);
 
 #if NET9_0
-        logger?.Invoke("Reading files using V2 APK layout.");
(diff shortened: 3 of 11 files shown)


src/Sentry.Android.AssemblyReader/DebugLogger.cs
```diff
@@ -3,6 +3,7 @@ namespace Sentry.Android.AssemblyReader;
 /// <summary>
 /// Writes a log message for debugging.
 /// </summary>
+/// <param name="level">The debug log level.</param>
 /// <param name="message">The message string to write.</param>
 /// <param name="args">Arguments for the formatted message string.</param>
-public delegate void DebugLogger(string message, params object?[] args);
+public delegate void DebugLogger(DebugLoggerLevel level, string message, params object?[] args);
```
Flash0ver, line 9: question: breaking change

Since we don't mention `Sentry.Android.AssemblyReader` in the README as a top-level package,
nor in our documentation as a platform guide,
we do not consider changes to the public surface area of this Assembly a breaking change,
since it's main purpose is to be consumed by the Android TFMs of `Sentry`,
right?

src/Sentry.Android.AssemblyReader/DebugLoggerLevel.cs
```diff
@@ -0,0 +1,32 @@
+namespace Sentry.Android.AssemblyReader;
+
+/// <summary>
+/// Represents the level of debug logging.
+/// </summary>
+public enum DebugLoggerLevel : short
```
Flash0ver, line 6: thought: naming

I find the name `DebugLogger` with the enum constant `DebugLoggerLevel.Debug` a bit odd.

Since both the `delegate` and the `enum` are mirroring `IDiagnosticLogger` and the `SentryLevel`,
I was wondering if we should rename
- `DebugLogger` -> `DiagnosticLogger`
- `DebugLoggerLevel` -> `DiagnosticLoggerLevel`

But then we already have a public type `DiagnosticLogger` in `Sentry`,
where I don't think we would have an identifier collisions, but definitely some confusion for the reader.

Another approach would be to make it shorter,
perhaps just `Logger` and `LoggerLevel`,
which again shouldn't collide with other names, but may still very much confuse.

As a third (and last) suggestion,
we could also lean into the fact, that this just represents a decoupling callback to the `Sentry`-declared `IDiagnosticLogger` and `DiagnosticLogger`,
so we could call the `delegate` -> `DiagnosticLoggerCallback`
and the `enum` -> `DiagnosticLoggerLevel`

I prefer option 3, but it also doesn't feel quite right, to be fair.
If neither of these options really work for you ... let's stick with `DebugLogger` and `DebugLoggerLevel`.

src/Sentry.Android.AssemblyReader/DebugLogger.cs
jamescrosswell: Hm, actually it's only used from here:
https://github.com/getsentry/sentry-dotnet/blob/eadf5f20b8014745d0fe42b0e707bf3f9eec9ffc/src/Sentry/SentryOptions.cs#L1196-L1206

It looks like Matt added that comment in https://github.com/getsentry/sentry-dotnet/pull/2127

@bruno-garcia given that sentry-xamarin is no longer being maintained, could we potentially move the AssemblyReader back into the main Sentry packages and make all this stuff internal again (not in this PR of course - probably for v6.0)?

src/Sentry.Android.AssemblyReader/DebugLoggerLevel.cs
jamescrosswell: Since we plan to get rid of it entirely (when we move the Android AssemblyReader functionality back into the Sentry assembly), let's leave it as is for now... no point in tidying up something we're going to nuke.

src/Sentry.Android.AssemblyReader/DebugLogger.cs
bruno-garcia: I'm in favor of that, just get rid of that package and make our lifes easier

bruno-garcia: to be fair, it should never have been created. It's very little code and only used in 2 package we own. We could have copy pasted things and manually keeped them in sync.

We knew Xamarin was schedule for retirement by then already.

Classic .NET overengineering

jamescrosswell: - https://github.com/getsentry/sentry-dotnet/issues/4466

Issue #4173, opened 2025-05-07
Support .NET 10


bruno-garcia:
We should have a branch with .NET 10 from preview 6 at least onwards. We can test in CI and raise any issues with Microsoft.

It probably makes sense to split this work into multiple PRs. The first one we're likely to be able to get passing CI is RC1 (since there's a blocking in the Preview 7 release relating to MS Build support).  

- [x] https://github.com/getsentry/sentry-dotnet/issues/4490
- [ ] https://github.com/getsentry/sentry-dotnet/pull/4461
- [ ] https://github.com/getsentry/sentry-dotnet/pull/4583
- [ ] RC 2
- [ ] GA

For reference:
* https://github.com/getsentry/sentry-dotnet/pull/3699

Flash0ver:
New/improved features we may want to take a closer look at:
- C# 14.0
  - see also https://github.com/dotnet/roslyn/blob/main/docs/Language%20Feature%20Status.md
- Runtime Async ("async2")
  - early plan is to preview as experimental feature in `net10` and GA for all platforms/runtimes in `net11`
  - Stack Traces closer to the source code, rather than the Compiler-generated state machine of "Compiler Async"
  - note: in general, some best practices around `async/await` may change, some perf-related workarounds may now become an anti-pattern
  - see also: https://github.com/dotnet/apireviews/tree/main/2025/04-22-quick-reviews
- Sampling & Buffering of Logs
  - see .NET 9.0 `IBufferedLogger` abstraction
    - https://github.com/dotnet/apireviews/tree/main/2024/07-16-quick-reviews
  - see .NET 10.0 `LoggingSampler` and `LogBuffer` and more
    - https://github.com/dotnet/apireviews/blob/main/2025/02-11-quick-reviews/README.md
    - https://github.com/dotnet/apireviews/blob/main/2025/02-18-quick-reviews/README.md
- Java Marshalling API
  - https://github.com/dotnet/apireviews/blob/main/2025/06-03-quick-reviews/README.md

PR #4461, opened 2025-08-21
Add support for net10.0-rc.1

jamescrosswell:
Initial commit for #4173
- https://github.com/getsentry/sentry-dotnet/issues/4173

### Done
- [x] SentryNoMobile.slnf builds
- [x] Sentry-CI-Build-macOS.slnf builds (including mobile targets)
- [x] Sentry-CI-Build-Windows.slnf builds on Windows
- [x] Unit Tests
- [x] Integration Tests
- [x] Trim Tests
- [x] MSBuild


integration-test/cli.Tests.ps1
```diff
@@ -155,12 +151,14 @@ Describe 'MAUI' -ForEach @(
             'libxamarin-app.so',
             'maui-app.pdb'
         )
+        $nonZeroNumberRegex = '[1-9][0-9]*';
         $result.ScriptOutput | Should -AnyElementMatch 'Uploaded a total of 1 new mapping files'
-        $result.ScriptOutput | Should -AnyElementMatch 'Found 25 debug information files \(1 with embedded sources\)'
+        $result.ScriptOutput | Should -AnyElementMatch "Found $nonZeroNumberRegex debug information files \($nonZeroNumberRegex with embedded sources\)"
```
bruno-garcia: Wouldn't we want the hard coded value instead?
It shouldn't change once we validate the value is correct, right? If instead of 25 we get 1, that'd prob be broken.

msbuild.binlog
bruno-garcia, line 1: delete this

samples/Sentry.Samples.MacCatalyst/Sentry.Samples.MacCatalyst.csproj
```diff
@@ -1,11 +1,11 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFramework>net8.0-maccatalyst17.0</TargetFramework>
+    <TargetFramework>net10.0-maccatalyst</TargetFramework>
```
bruno-garcia: we shouldn't pin to a specific version anymore?

samples/Sentry.Samples.Maui/Sentry.Samples.Maui.csproj
```diff
@@ -6,7 +6,7 @@
       On Mac, we'll also build for iOS and MacCatalyst.
       On Windows, we'll also build for Windows 10.
     -->
-    <TargetFrameworks>$(TargetFrameworks);net9.0-android35.0</TargetFrameworks>
+    <TargetFrameworks>$(TargetFrameworks);net9.0-android</TargetFrameworks>
```
bruno-garcia: shouldn't we set the version here either anymore? like android36 or something?

integration-test/cli.Tests.ps1
jamescrosswell: It did actually change between net9.0 and net10.0 (I forget which way - either 24 or 26). 

I noticed the `uploads symbols and sources for an iOS build`  test below used `$nonZeroNumberRegex` so copied the pattern here. 

We could go back to checking for a specific number of symbol files being uploaded. I'm not sure how many to expect though... what determines that?

samples/Sentry.Samples.MacCatalyst/Sentry.Samples.MacCatalyst.csproj
jamescrosswell: For the projects in the `src` folder (i.e. our NuGet packages) we definitely should. 

For the Samples projects it looks like we haven't been consistent in the past and, for the most part, we just targeted a .NET version (not a specific API level)... except for these couple of sample projects... so I took the API level off these to be consistent. 

The other option would be to revert these two changes and also add the API level on all of the other Sample projects,

So I think we either add it for all of them or none of them, basically.

Allan reckoned if you added it for MAUI projects they broke, but he wasn't specific about how or in what way, so I haven't confirmed that yet.

Directory.Build.props
```diff
@@ -20,11 +21,26 @@
...
+    <!-- For Xcode 26 support in .NET 10 -->
+    <NoWarn>$(NoWarn);XCODE_26_0_PREVIEW</NoWarn>
+
     <!-- https://learn.microsoft.com/en-us/dotnet/core/compatibility/sdk/8.0/dotnet-restore-audit#version-introduced -->
     <WarningsNotAsErrors>NU1902;NU1903</WarningsNotAsErrors>
+
+    <!-- Centralized TFM versions -->
+    <CurrentTfms>net8.0;net9.0;net10.0</CurrentTfms>
+    <LatestTfm>net10.0</LatestTfm>
+    <PreviousTfm>net9.0</PreviousTfm>
```
Flash0ver, line 32: thought:
I like this centralization. And this still goes in line with the change that STS versions are now supported for 6 more months: from 18 months to 24 months. See https://devblogs.microsoft.com/dotnet/dotnet-sts-releases-supported-for-24-months/.
There are now always 2 versions of .NET supported, every other year 3 versions.
The MAUI support policy remains unchanged though.

integration-test/cli.Tests.ps1
```diff
@@ -155,12 +151,14 @@ Describe 'MAUI' -ForEach @(
...
     }
 
     It "uploads symbols and sources for an iOS build" -Skip:(!$IsMacOS) {
         $result = RunDotnetWithSentryCLI 'build' 'maui-app' $True $True "$framework-ios$iosTpv"
+        Write-Host "UploadedDebugFiles: $($result.UploadedDebugFiles() | Out-String)"
```
Flash0ver, line 161: suggestion (non-blocking): We could add this line to the test above (`uploads symbols and sources for an Android build`) for consistency as well.

integration-test/common.ps1
```diff
@@ -58,9 +62,72 @@ BeforeAll {
...
+            'net10.0' { return '26' }    # aligns with ios26 / maccatalyst26
+            default { throw "Unsupported framework '$framework' for iOS target platform version." }
+        }
+    }
+
     function GetSentryPackageVersion()
     {
-        (Select-Xml -Path "$PSScriptRoot/../Directory.Build.props" -XPath "/Project/PropertyGroup/VersionPrefix").Node.InnerText
+        # Read version directly from Directory.Build.props
+        $propsFile = Join-Path $PSScriptRoot '..\Directory.Build.props'
```
Flash0ver, line 86:
```suggestion
        $propsFile = Join-Path (Get-Item -Path $PSScriptRoot).Parent 'Directory.Build.props'
```

Now looking at my suggestion again, I'm not sure which version is cleaner.
Whichever you prefer!

btw ... nice improvement!

integration-test/runtime.Tests.ps1
```diff
@@ -188,7 +188,7 @@ Describe 'Console app regression (missing System.Reflection.Metadata)' {
         dotnet remove ./net4-console/console-app.csproj package Sentry
     }
 
-    It 'Ensure System.Reflection.Metadata is not missing' {
+    It 'Ensure System.Reflection.Metadata is not missing' -Skip:(!$IsWindows) {
```
Flash0ver, line 191: question: Did this not fail on Windows before? What has changed? Only on .NET 10? Something we need to follow-up with a fix for in the SDK?

samples/Sentry.Samples.Android/Sentry.Samples.Android.csproj
```diff
@@ -1,6 +1,6 @@
 <Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
-    <TargetFramework>net8.0-android</TargetFramework>
+    <TargetFramework>net9.0-android</TargetFramework>
```
Flash0ver: question: Is it intended to be explicit in the samples, rather than using `PreviousAndroidTfm` and such?

samples/Sentry.Samples.AspNetCore.Basic/Sentry.Samples.AspNetCore.Basic.csproj
```diff
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk.Web">
 
   <PropertyGroup>
-    <TargetFramework>net8.0</TargetFramework>
+    <TargetFramework>net10.0</TargetFramework>
```
Flash0ver, line 4: question: Is it intended to use the current TFM for non-mobile samples, and the previous TFM for mobile-related samples? And I think this pattern is not entirely consistent with all samples.

samples/Sentry.Samples.Maui/Sentry.Samples.Maui.csproj
Flash0ver: also to be consistent with the conditional `TargetFrameworks` below

src/Sentry.AspNetCore/Sentry.AspNetCore.csproj
```diff
@@ -11,18 +11,15 @@
...
     <FrameworkReference Include="Microsoft.AspNetCore.App" />
   </ItemGroup>
 
-  <ItemGroup Condition="'$(TargetFramework)' == 'net8.0'">
-    <PackageReference Include="Microsoft.Extensions.Configuration.Binder" Version="8.0.0" />
-  </ItemGroup>
-  <ItemGroup Condition="'$(TargetFramework)' == 'net9.0'">
-    <PackageReference Include="Microsoft.Extensions.Configuration.Binder" Version="9.0.0" />
-  </ItemGroup>
-
```
Flash0ver, line 25: question: Removed because we depended on it transitively through `Sentry.Extensions.Logging` already?

src/Sentry.Bindings.Android/Sentry.Bindings.Android.csproj
```diff
@@ -43,16 +43,15 @@
...
        -->
-  <ItemGroup Condition="$(TargetFramework.StartsWith('net8'))">
-    <PackageReference Include="Xamarin.AndroidX.Lifecycle.Process" Version="2.6.1.3" />
-    <PackageReference Include="Xamarin.AndroidX.Lifecycle.Common.Java8" Version="2.6.1.3" />
-    <!-- MAUI 8 references this version indirectly via Xamarin.AndroidX.SwipeRefreshLayout (>= 1.1.0.14)   -->
-    <PackageReference Include="Xamarin.AndroidX.Core" Version="1.10.1.2" />
+  <ItemGroup Condition="$(TargetFramework.StartsWith('net10'))">
+    <PackageReference Include="Xamarin.AndroidX.Lifecycle.Process" Version="2.9.2.1" />
+    <PackageReference Include="Xamarin.AndroidX.Lifecycle.Common.Java8" Version="2.9.2.1" />
+    <!-- MAUI 9 references this version indirectly via Xamarin.AndroidX.SwipeRefreshLayout   -->
```
Flash0ver:
```suggestion
    <!-- MAUI 10 references this version indirectly via Xamarin.AndroidX.SwipeRefreshLayout   -->
```

src/Sentry.Bindings.Android/Sentry.Bindings.Android.csproj
```diff
@@ -43,16 +43,15 @@
...
+    <PackageReference Include="Xamarin.AndroidX.Core" Version="1.17.0" />
     <!-- We need 1.9.24 but there's no NuGet package for that version... the lowest version we can use is 2.0.0 -->
```
Flash0ver: question: Is this comment still valid (was from `net8.0`)? Or does it need to be updated?

src/Sentry.Bindings.Cocoa/SwiftApiDefinitions.cs
```diff
@@ -329,9 +329,6 @@ interface SentryRRWebEvent : SentrySerializable
     [Export ("initWithType:timestamp:data:")]
     [DesignatedInitializer]
     NativeHandle Constructor (SentryRRWebEventType type, NSDate timestamp, [NullAllowed] NSDictionary<NSString, NSObject> data);
-    // -(NSDictionary<NSString *,id> * _Nonnull)serialize __attribute__((warn_unused_result("")));
-    [Export ("serialize")]
-    new NSDictionary<NSString, NSObject> Serialize();
```
Flash0ver, line 334: question: Why is this no longer needed? We haven't updated the `sentry-cocoa` dependency here 🤔

src/Sentry.Google.Cloud.Functions/Sentry.Google.Cloud.Functions.csproj
```diff
@@ -19,6 +19,7 @@
   <ItemGroup Condition="'$(TargetFramework)' == 'net8.0'">
     <PackageReference Include="Microsoft.Extensions.Configuration.Binder" Version="8.0.0" />
   </ItemGroup>
+
   <ItemGroup Condition="'$(TargetFramework)' == 'net9.0'">
     <PackageReference Include="Microsoft.Extensions.Configuration.Binder" Version="9.0.0" />
   </ItemGroup>
```
Flash0ver: question: Do we need a `net10.0` entry too?

```XML
<ItemGroup Condition="'$(TargetFramework)' == 'net10.0'">
  <PackageReference Include="Microsoft.Extensions.Configuration.Binder" Version="10.0.0-rc.1.25451.107" />
</ItemGroup>
```

Or is this not needed? In which case, could we remove the `net8.0` and `net9.0` entries too?

src/Sentry.Maui/Internal/MauiEventsBinder.cs
```diff
@@ -298,9 +298,9 @@ internal void HandlePageEvents(Page page, bool bind = true)
         // https://github.com/dotnet/docs-maui/issues/583
         page.NavigatedTo -= OnPageOnNavigatedTo;
 
-        // Layout changed event
-        // https://docs.microsoft.com/dotnet/api/xamarin.forms.ilayout.layoutchanged
-        page.LayoutChanged -= OnPageOnLayoutChanged;
+        // Size changed event
+        // https://learn.microsoft.com/en-us/dotnet/api/microsoft.maui.controls.visualelement.sizechanged
```
Flash0ver:
```suggestion
        // https://learn.microsoft.com/dotnet/api/microsoft.maui.controls.visualelement.sizechanged
```

src/Sentry.Maui/Internal/MauiEventsBinder.cs
```diff
@@ -313,9 +313,9 @@ internal void HandlePageEvents(Page page, bool bind = true)
             // https://github.com/dotnet/docs-maui/issues/583
             page.NavigatedTo += OnPageOnNavigatedTo;
 
-            // Layout changed event
-            // https://docs.microsoft.com/dotnet/api/xamarin.forms.ilayout.layoutchanged
-            page.LayoutChanged += OnPageOnLayoutChanged;
+            // Size changed event
+            // https://learn.microsoft.com/en-us/dotnet/api/microsoft.maui.controls.visualelement.sizechanged
```
Flash0ver:
```suggestion
            // https://learn.microsoft.com/dotnet/api/microsoft.maui.controls.visualelement.sizechanged
```

src/Sentry/AttributeReader.cs
```diff
@@ -2,6 +2,25 @@ namespace Sentry;
 
 internal static class AttributeReader
 {
-    public static string? TryGetProjectDirectory(Assembly assembly) =>
-        assembly.GetCustomAttributes<AssemblyMetadataAttribute>().FirstOrDefault(x => x.Key == "Sentry.ProjectDirectory")?.Value;
+    public static string? TryGetProjectDirectory(Assembly assembly)
+    {
+        // Use metadata (CustomAttributeData) to avoid hard-referencing AssemblyMetadataAttribute,
+        // which may be trimmed. This safely returns null if the attribute instances were removed.
+        // The constructor is: `AssemblyMetadataAttribute(string key, string? value);`
```
Flash0ver, line 9: question: Not sure if I get that right. I cases where the `AssemblyMetadataAttribute` assembly attribute is trimmed, won't `GetCustomAttributes<T>` then not yield an element?
And/or the `AssemblyMetadataAttribute` type itself can't be trimmed, since we reference it type-safely, can it?
🤔

test/Sentry.AspNetCore.Tests/WebIntegrationTests.Versioning.DotNet10_0.verified.txt
```diff
@@ -3,7 +3,8 @@
   Payloads: [
```
Flash0ver, line 3: Nice removal/reuse of obsolete/wrong `...Versioning.DotNet.DotNet8_0....`

May these two files be removed too?
- `WebIntegrationTests.PreFlightIgnoresTransaction.DotNet.verified.txt`
- `WebIntegrationTests.Versioning.verified.txt`

test/Sentry.AspNetCore.Tests/WebIntegrationTests.Versioning.DotNet10_0.verified.txt
```diff
@@ -3,7 +3,8 @@
...
     {
       Source: {
-        Name: GET v1.1/Target,
+        Origin: auto.http.aspnetcore,
+        Name: GET /v1.1/Target,
```
Flash0ver, line 7: question: Are the differences to `net8.0` and `net9.0` expected? Or do we need a follow-up?

test/Sentry.Extensions.Logging.Tests/Sentry.Extensions.Logging.Tests.csproj
```diff
@@ -15,6 +15,10 @@
     <ProjectReference Include="..\Sentry.Testing\Sentry.Testing.csproj" />
   </ItemGroup>
 
+  <ItemGroup Condition="'$(TargetFramework)' == 'net10.0'">
+    <PackageReference Include="Microsoft.Extensions.DependencyInjection" Version="10.0.0-rc.1.25451.107" />
+  </ItemGroup>
```
Flash0ver: question: Why do we need this dependency for `net10.0`, but not for `net9.0` nor `net8.0`?

test/Sentry.AspNetCore.Tests/Sentry.AspNetCore.Tests.csproj
```diff
@@ -10,6 +10,10 @@
     <ProjectReference Include="..\Sentry.Testing\Sentry.Testing.csproj" />
   </ItemGroup>
 
+  <ItemGroup Condition="'$(TargetFramework)' == 'net10.0'">
+    <PackageReference Include="Microsoft.AspNetCore.TestHost" Version="10.0.0-rc.1.25451.107" />
+  </ItemGroup>
+
```
Flash0ver: question: Could we remove this dependency, as we receive it transitively through `Sentry.AspNetCore.TestUtils`?

test/Sentry.Google.Cloud.Functions.Tests/Sentry.Google.Cloud.Functions.Tests.csproj
```diff
@@ -1,12 +1,16 @@
...
 
   <ItemGroup>
     <ProjectReference Include="..\..\src\Sentry.Google.Cloud.Functions\Sentry.Google.Cloud.Functions.csproj" />
     <ProjectReference Include="..\Sentry.AspNetCore.TestUtils\Sentry.AspNetCore.TestUtils.csproj" />
   </ItemGroup>
 
+  <ItemGroup Condition="'$(TargetFramework)' == 'net10.0'">
+    <PackageReference Include="Microsoft.AspNetCore.TestHost" Version="10.0.0-rc.1.25451.107" />
+  </ItemGroup>
+
```
Flash0ver: question: Could we remove this dependency, as we receive it transitively through `Sentry.AspNetCore.TestUtils`?

test/Sentry.Log4Net.Tests/Sentry.Log4Net.Tests.csproj
```diff
@@ -1,8 +1,8 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <!-- We use net8.0 to test netstandard2.0 -->
-    <TargetFrameworks>net8.0</TargetFrameworks>
+    <!-- We use the oldest version of .NET that we have to test netstandard2.0 -->
+    <TargetFrameworks>$(PreviousTfm)</TargetFrameworks>
```
Flash0ver: question: Technically, the oldest supported Version is `net8.0`. Or do we prefer the `PreviousTfm` variable here for convenience with the next .NET upgrade?

test/Sentry.Maui.Tests/MauiEventsBinderTests.Page.cs
```diff
@@ -81,8 +88,11 @@ public void Page_NavigatedTo_AddsBreadcrumb()
         Assert.Equal(MauiEventsBinder.NavigationType, crumb.Type);
         Assert.Equal(MauiEventsBinder.NavigationCategory, crumb.Category);
         crumb.Data.Should().Contain($"{nameof(Page)}.Name", "page");
+#if !NET10_0
+        // TODO: Work out why these are missing in .NET 10
```
Flash0ver, line 92: question: Shall we create an issue for that, or is it already on your radar?

Flash0ver (commented):
oops ... sorry ... 22 comments 😟 ... I guess we can do some of these in a follow-up PR ... also lots of questions I am not certain about the change

samples/Sentry.Samples.Android/Sentry.Samples.Android.csproj
jamescrosswell: That was the idea yes. I figured leaving these explicit makes the samples easier to read for SDK users.

integration-test/common.ps1
jamescrosswell: To minimise the amount of work for this PR (which is already massive), I might skip this per the "If it ain't broke, don't fix it" maxim.

src/Sentry.AspNetCore/Sentry.AspNetCore.csproj
jamescrosswell: Honestly, I can't remember... but it no longer appears necessary (everything builds and the tests pass without any explicit reference to this).

It's possible I got some analyser warnings that told me to do this instead:
https://github.com/getsentry/sentry-dotnet/blob/0fe695b14b9ff4e310af6053a4b2dcebac06d8eb/src/Sentry.AspNetCore/Sentry.AspNetCore.csproj#L14-L16

integration-test/runtime.Tests.ps1
jamescrosswell: Skipping it for everything **_except_** Windows actually:
https://github.com/getsentry/sentry-dotnet/blob/f565a5db93f67854cf310cb27c535fe1de1cccdf/integration-test/runtime.Tests.ps1#L191-L193

The test appears to depend on net4 so Windows only. It might be possible to get it working with mono on other platforms but not sure that's worth the effort. If we do want to do this for some reason, I'd recommend we do it in a follow up PR.

samples/Sentry.Samples.AspNetCore.Basic/Sentry.Samples.AspNetCore.Basic.csproj
jamescrosswell: I think that's what I did initially when working with preview 7 yes, [due to this](https://learn.microsoft.com/en-us/dotnet/maui/whats-new/dotnet-10?view=net-maui-9.0#net-for-ios):
> Preview 7 release includes Xcode 26 Beta 4 support for targeting .NET 9. We will include builds for .NET 10 in a subsequent release. To use these new bindings, target net9.0-ios26 and/or net9.0-maccatalyst26 and include <NoWarn>$(NoWarn);XCODE_26_0_PREVIEW</NoWarn> in your project file.

Apparently that's [still the case in RC1](https://github.com/dotnet/macios/releases/tag/dotnet-10.0.1xx-rc.1-10727).

src/Sentry/AttributeReader.cs
jamescrosswell: Again, from memory, our previous code resulted in trim analyser warnings in .NET 10 saying the `AssemblyMetadataAttribute` would be stripped in trimmed applications (whether you reference the attributes in code or not isn't considered by the trimmer)... so I needed an alternate implementation that would work even for Trimmed/AOT compiled applications.

src/Sentry.Bindings.Cocoa/SwiftApiDefinitions.cs
jamescrosswell: It was never necessary... now it actively breaks the bindings. 

AFAICR the base class already implements Serialize and this type doesn't override or change it in any way.

test/Sentry.Maui.Tests/MauiEventsBinderTests.Page.cs
jamescrosswell: Good idea 👍🏻 
- https://github.com/getsentry/sentry-dotnet/issues/4578

src/Sentry.Bindings.Android/Sentry.Bindings.Android.csproj
jamescrosswell: It's still relevant yes - the dependency comes from the Sentry Java SDK. I've updated the comment to provide a bit more context. 
https://github.com/getsentry/sentry-dotnet/blob/d7a65af7b873b0c89fd8d25357f89717312830c0/src/Sentry.Bindings.Android/Sentry.Bindings.Android.csproj#L53-L55

test/Sentry.AspNetCore.Tests/WebIntegrationTests.Versioning.DotNet10_0.verified.txt
jamescrosswell: Hm, that does look suspicious. Looks like it was broken in .NET 9 already:
- https://github.com/getsentry/sentry-dotnet/issues/4579

test/Sentry.Extensions.Logging.Tests/Sentry.Extensions.Logging.Tests.csproj
jamescrosswell: Well spotted... looks like this is redundant.

test/Sentry.Log4Net.Tests/Sentry.Log4Net.Tests.csproj
jamescrosswell: I've changed to test against all current TFMs. 

btw: In the past, we did use the oldest TFMs to test netstandard... which we could still do. Basically if the TFM was the oldest TFM, we'd pass netstandard2.1 as the value for the TargetFramework property in the package reference to `Sentry.Log4Net`... which would force the netstandard version of our integration to be resolved when running unit tests against net8.0. The downside of this is that we would no longer be testing against `net8.0` so 🤷🏻

src/Sentry/buildTransitive/Sentry.targets
```diff
@@ -18,6 +18,7 @@
 
   <PropertyGroup>
     <SentryAttributesFile>Sentry.Attributes$(MSBuildProjectExtension.Replace('proj', ''))</SentryAttributesFile>
+    <SentryProGuardUUID Condition="'$(SentryProGuardUUID)' == ''">$([System.Guid]::NewGuid())</SentryProGuardUUID>
```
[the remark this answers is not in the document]
jamescrosswell: Not related to this PR.

src/Sentry.Bindings.Android/Sentry.Bindings.Android.csproj
Flash0ver: ahh ... now I get it ...
perhaps for clarity, could it help to add a brief note that `2.2.10` is the version available when we introduced `net10.0`

test/Sentry.Log4Net.Tests/Sentry.Log4Net.Tests.csproj
Flash0ver: I see.
For log4net we currently target `<TargetFrameworks>netstandard2.0;net462</TargetFrameworks>`
and test `net9.0` and `net48` (on Windows) .... so we actually do have full TFM converage.

Flash0ver: We are updating to `net10.0`, since we ran `net9.0` already before, so that we cover for potentially new issues.

integration-test/cli.Tests.ps1
jamescrosswell: I've reverted to hard coding the expectations... that way at least we'll detect when they change and can decide whether to be worried about it or not.

samples/Sentry.Samples.AspNetCore.Basic/Sentry.Samples.AspNetCore.Basic.csproj
jamescrosswell: Added a new issue to cover this off separately:
- https://github.com/getsentry/sentry-dotnet/issues/4584

samples/Sentry.Samples.MacCatalyst/Sentry.Samples.MacCatalyst.csproj
jamescrosswell: Reverted.

Flash0ver (approved):

PR #4462, opened 2025-08-21
feat: add `Serilog` integration

Flash0ver:
Add `Serilog` integration to Sentry Structured Logging.

### DIFF
--- a/src/Sentry.Serilog/SentrySink.Structured.cs
+++ b/src/Sentry.Serilog/SentrySink.Structured.cs
@@ -5,6 +5,8 @@ namespace Sentry.Serilog;
 
 internal sealed partial class SentrySink
 {
+    private static readonly SdkVersion Sdk = CreateSdkVersion();
+
     private void CaptureStructuredLog(IHub hub, LogEvent logEvent, string formatted, string? template)
     {
         var traceHeader = hub.GetTraceHeader() ?? SentryTraceHeader.Empty;
@@ -17,8 +19,7 @@ internal sealed partial class SentrySink
             ParentSpanId = traceHeader.SpanId,
         };
 
-        var scope = hub.GetScope();
-        log.SetDefaultAttributes(_options, scope?.Sdk ?? SdkVersion.Instance);
+        log.SetDefaultAttributes(_options, Sdk);
 
         foreach (var attribute in attributes)
         {
@@ -46,14 +47,14 @@ internal sealed partial class SentrySink
         {
             if (propertyTokens.Exists(prop => prop.PropertyName == property.Key))
             {
-                if (TryGetLogEventProperty(property, out var parameter))
+                foreach (var parameter in GetLogEventProperties(property))
                 {
                     @params.Add(parameter);
                 }
             }
             else
             {
-                if (TryGetLogEventProperty(property, out var attribute))
+                foreach (var attribute in GetLogEventProperties(property))
                 {
                     attributes.Add(new KeyValuePair<string, object>($"property.{attribute.Key}", attribute.Value));
                 }
@@ -63,14 +64,27 @@ internal sealed partial class SentrySink
         parameters = @params.DrainToImmutable();
         return;
 
-        static bool TryGetLogEventProperty(KeyValuePair<string, LogEventPropertyValue> property, out KeyValuePair<string, object> value)
+        static IEnumerable<KeyValuePair<string, object>> GetLogEventProperties(KeyValuePair<string, LogEventPropertyValue> property)
         {
             if (property.Value is ScalarValue scalarValue)
             {
                 if (scalarValue.Value is not null)
                 {
-                    value = new KeyValuePair<string, object>(property.Key, scalarValue.Value);
-                    return true;
+                    yield return new KeyValuePair<string, object>(property.Key, scalarValue.Value);
+                }
+            }
+            else if (property.Value is SequenceValue sequenceValue)
+            {
+                if (sequenceValue.Elements.Count != 0)
+                {
+                    yield return new KeyValuePair<string, object>(property.Key, sequenceValue.ToString());
+                }
+            }
+            else if (property.Value is DictionaryValue dictionaryValue)
+            {
+                if (dictionaryValue.Elements.Count != 0)
+                {
+                    yield return new KeyValuePair<string, object>(property.Key, dictionaryValue.ToString());
                 }
             }
             else if (property.Value is StructureValue structureValue)
@@ -79,30 +93,23 @@ internal sealed partial class SentrySink
                 {
                     if (LogEventProperty.IsValidName(prop.Name))
                     {
-                        if (prop.Value is ScalarValue scalarProperty)
-                        {
-                            if (scalarProperty.Value is not null)
-                            {
-                                value = new KeyValuePair<string, object>($"{property.Key}.{prop.Name}", scalarProperty.Value);
-                                return true;
-                            }
-                        }
-                        else
-                        {
-                            value = new KeyValuePair<string, object>($"{property.Key}.{prop.Name}", prop.Value);
-                            return true;
-                        }
+                        yield return new KeyValuePair<string, object>($"{property.Key}.{prop.Name}", prop.Value.ToString());
                     }
                 }
             }
             else if (!property.Value.IsNull())
             {
-                value = new KeyValuePair<string, object>(property.Key, property.Value);
-                return true;
+                yield return new KeyValuePair<string, object>(property.Key, property.Value);
             }
-
-            value = default;
-            return false;
         }
+    }
+
+    private static SdkVersion CreateSdkVersion()
+    {
+        return new SdkVersion
+        {
+            Name = SdkName,
+            Version = NameAndVersion.Version,
+        };
     }
 }
--- a/src/Sentry.Serilog/SentrySink.Structured.cs
+++ b/src/Sentry.Serilog/SentrySink.Structured.cs
@@ -1,0 +1,108 @@
+using Serilog.Parsing;
+using Sentry.Internal.Extensions;
+
+namespace Sentry.Serilog;
+
+internal sealed partial class SentrySink
+{
+    private void CaptureStructuredLog(IHub hub, LogEvent logEvent, string formatted, string? template)
+    {
+        var traceHeader = hub.GetTraceHeader() ?? SentryTraceHeader.Empty;
+        GetStructuredLoggingParametersAndAttributes(logEvent, out var parameters, out var attributes);
+
+        SentryLog log = new(logEvent.Timestamp, traceHeader.TraceId, logEvent.Level.ToSentryLogLevel(), formatted)
+        {
+            Template = template,
+            Parameters = parameters,
+            ParentSpanId = traceHeader.SpanId,
+        };
+
+        var scope = hub.GetScope();
+        log.SetDefaultAttributes(_options, scope?.Sdk ?? SdkVersion.Instance);
+
+        foreach (var attribute in attributes)
+        {
+            log.SetAttribute(attribute.Key, attribute.Value);
+        }
+
+        hub.Logger.CaptureLog(log);
+    }
+
+    private static void GetStructuredLoggingParametersAndAttributes(LogEvent logEvent, out ImmutableArray<KeyValuePair<string, object>> parameters, out List<KeyValuePair<string, object>> attributes)
+    {
+        var propertyTokens = new List<PropertyToken>();
+        foreach (var token in logEvent.MessageTemplate.Tokens)
+        {
+            if (token is PropertyToken property)
+            {
+                propertyTokens.Add(property);
+            }
+        }
+
+        var @params = ImmutableArray.CreateBuilder<KeyValuePair<string, object>>();
+        attributes = new List<KeyValuePair<string, object>>();
+
+        foreach (var property in logEvent.Properties)
+        {
+            if (propertyTokens.Exists(prop => prop.PropertyName == property.Key))
+            {
+                if (TryGetLogEventProperty(property, out var parameter))
+                {
+                    @params.Add(parameter);
+                }
+            }
+            else
+            {
+                if (TryGetLogEventProperty(property, out var attribute))
+                {
+                    attributes.Add(new KeyValuePair<string, object>($"property.{attribute.Key}", attribute.Value));
+                }
+            }
+        }
+
+        parameters = @params.DrainToImmutable();
+        return;
+
+        static bool TryGetLogEventProperty(KeyValuePair<string, LogEventPropertyValue> property, out KeyValuePair<string, object> value)
+        {
+            if (property.Value is ScalarValue scalarValue)
+            {
+                if (scalarValue.Value is not null)
+                {
+                    value = new KeyValuePair<string, object>(property.Key, scalarValue.Value);
+                    return true;
+                }
+            }
+            else if (property.Value is StructureValue structureValue)
+            {
+                foreach (var prop in structureValue.Properties)
+                {
+                    if (LogEventProperty.IsValidName(prop.Name))
+                    {
+                        if (prop.Value is ScalarValue scalarProperty)
+                        {
+                            if (scalarProperty.Value is not null)
+                            {
+                                value = new KeyValuePair<string, object>($"{property.Key}.{prop.Name}", scalarProperty.Value);
+                                return true;
+                            }
+                        }
+                        else
+                        {
+                            value = new KeyValuePair<string, object>($"{property.Key}.{prop.Name}", prop.Value);
+                            return true;
+                        }
+                    }
+                }
+            }
+            else if (!property.Value.IsNull())
+            {
+                value = new KeyValuePair<string, object>(property.Key, property.Value);
+                return true;
+            }
+
+            value = default;
+            return false;
+        }
+    }
+}
--- a/src/Sentry.Serilog/SentrySink.cs
+++ b/src/Sentry.Serilog/SentrySink.cs
@@ -5,7 +5,7 @@ namespace Sentry.Serilog;
 /// </summary>
 /// <inheritdoc cref="IDisposable" />
 /// <inheritdoc cref="ILogEventSink" />
-internal sealed class SentrySink : ILogEventSink, IDisposable
+internal sealed partial class SentrySink : ILogEventSink, IDisposable
 {
     private readonly IDisposable? _sdkDisposable;
     private readonly SentrySerilogOptions _options;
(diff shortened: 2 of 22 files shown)


test/Sentry.Serilog.Tests/SentrySerilogOptionsTests.cs
```diff
@@ -15,4 +15,11 @@ public void Ctor_MinimumEventLevel_Error()
         var sut = new SentrySerilogOptions();
         Assert.Equal(LogEventLevel.Error, sut.MinimumEventLevel);
     }
+
+    [Fact]
+    public void Ctor_EnableLogs_False()
```
Flash0ver: question: Keep or remove?

This test may be superfluous, as `EnableLogs` is declared in the "base" `SentryOptions` rather than the derived `SentrySerilogOptions`.
My original motivation to add this test was, because `EnableLogs` has direct impact within `Sentry.Serilog`.

src/Sentry.Serilog/SentrySinkExtensions.cs
```diff
@@ -86,6 +88,7 @@ public static LoggerConfiguration Sentry(
         LogEventLevel? minimumEventLevel = null,
         IFormatProvider? formatProvider = null,
         ITextFormatter? textFormatter = null,
+        bool? experimentalEnableLogs = null,
```
Flash0ver: question: Order

I am super uncertain about the order here ... I am going back and forth.
I believe I am now a bit more leaning towards adding at the end, because adding it in the middle is a breaking change when invoked with positional arguments.

src/Sentry.Serilog/SentrySinkExtensions.cs
```diff
@@ -170,15 +176,17 @@ public static LoggerConfiguration Sentry(
         LogEventLevel? minimumEventLevel = null,
         LogEventLevel? minimumBreadcrumbLevel = null,
         IFormatProvider? formatProvider = null,
-        ITextFormatter? textFormatter = null
+        ITextFormatter? textFormatter = null,
+        bool? experimentalEnableLogs = null
```
Flash0ver: question: Keep or remove?

I was going back and forth on that one, too.
But I am now slightly leaning towards removing it from here, as `EnableLogs` is declared in the "base" `SentryOptions` rather than the derived `SentrySerilogOptions`.
However, this option does have a direct impact on `Sentry.Serilog.SentrySink` in this assembly. That's why I put it there in the first place.

Flash0ver:
@Sentry review

src/Sentry.Serilog/SentrySinkExtensions.cs
```diff
@@ -354,7 +368,14 @@ public static LoggerConfiguration Sentry(
...
-        return loggerConfiguration.Sink(new SentrySink(options, sdkDisposable), minimumOverall);
+        if (options.Experimental.EnableLogs)
+        {
+            return loggerConfiguration.Sink(new SentrySink(options, sdkDisposable));
+        }
+        else
+        {
+            var minimumOverall = (LogEventLevel)Math.Min((int)options.MinimumBreadcrumbLevel, (int)options.MinimumEventLevel);
+            return loggerConfiguration.Sink(new SentrySink(options, sdkDisposable), minimumOverall);
+        }
```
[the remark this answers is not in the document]
Flash0ver: The `minimumOverall`, leading to a "Restricted-Sink", is an optimization, since when there is neither an Event nor a Breadcrumb produced, the `LogEvent` does not need to be `Emit`ted into the `SentrySink`.

However, unlike Events and Breadcrumbs, Structured Logs are sent (when enabled) without a Log-Level-Filter.
For Filtering of Structured Logs, a user callback set via `SetBeforeSendLog` is used.

So the actual bug would be:
- when Structured Logging is enabled via `EnableLogs`
- a LogEvent is emitted with a less Log-Level than the configured `MinimumBreadcrumbLevel` and `MinimumEventLevel`
- then the Structured Log wouldn't be captured either

Unless I'm misunderstanding something there.

Requesting more feedback.

src/Sentry.Serilog/SentrySink.Structured.cs
```diff
@@ -0,0 +1,115 @@
...
+            }
+            else if (property.Value is DictionaryValue dictionaryValue)
+            {
+                if (dictionaryValue.Elements.Count != 0)
+                {
+                    yield return new KeyValuePair<string, object>(property.Key, dictionaryValue.ToString());
+                }
+            }
+            else if (property.Value is StructureValue structureValue)
+            {
```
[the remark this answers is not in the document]
Flash0ver: Great suggestion.
This does entail a significant overhaul of the Structured Logging infrastructure.

I am leaning towards deferring to an evidence-based memory optimization,
when usage shows that `Serilog` is used quite a bit to send Structured Logs to Sentry.

src/Sentry.Serilog/SentrySink.Structured.cs
```diff
@@ -0,0 +1,115 @@
...
+    {
+        var propertyTokens = new List<PropertyToken>();
+        foreach (var token in logEvent.MessageTemplate.Tokens)
+        {
+            if (token is PropertyToken property)
+            {
+                propertyTokens.Add(property);
+            }
+        }
+
```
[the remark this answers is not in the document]
Flash0ver: Great suggestion! Will do!
However, the formatting is incorrect in this suggestion (incorrect indentation).
I'll change it manually in the morning.

src/Sentry.Serilog/SentrySinkExtensions.cs
```diff
@@ -354,7 +368,14 @@ public static LoggerConfiguration Sentry(
...
-        var minimumOverall = (LogEventLevel)Math.Min((int)options.MinimumBreadcrumbLevel, (int)options.MinimumEventLevel);
-        return loggerConfiguration.Sink(new SentrySink(options, sdkDisposable), minimumOverall);
+        if (options.Experimental.EnableLogs)
+        {
+            return loggerConfiguration.Sink(new SentrySink(options, sdkDisposable));
+        }
+        else
+        {
+            var minimumOverall = (LogEventLevel)Math.Min((int)options.MinimumBreadcrumbLevel, (int)options.MinimumEventLevel);
+            return loggerConfiguration.Sink(new SentrySink(options, sdkDisposable), minimumOverall);
```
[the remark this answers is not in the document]
Flash0ver: The solution would be to create a separate and independent Sink.
Which would need to do some processing of the `LogEvent` again, unless data is shared between both Sinks, which may lead to synchronization.

Not sure if this would actually increase performance 🤔

src/Sentry.Serilog/SentrySink.Structured.cs
```diff
@@ -0,0 +1,115 @@
...
+            Template = template,
+            Parameters = parameters,
+            ParentSpanId = traceHeader.SpanId,
+        };
+
+        log.SetDefaultAttributes(_options, Sdk);
+
+        foreach (var attribute in attributes)
+        {
+            log.SetAttribute(attribute.Key, attribute.Value);
```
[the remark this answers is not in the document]
Flash0ver: I don't see a path where `GetStructuredLoggingParametersAndAttributes` may throw.
The implementation of `hub.Logger.CaptureLog` actually does include error handling already.

src/Sentry.Serilog/SentrySink.cs
```diff
@@ -122,30 +122,33 @@ private void InnerEmit(LogEvent logEvent)
...
-
-        hub.AddBreadcrumb(
-            _clock,
-            string.IsNullOrWhiteSpace(formatted)
-                ? exception?.Message ?? ""
-                : formatted,
-            context,
-            data: data,
-            level: logEvent.Level.ToBreadcrumbLevel());
     }
```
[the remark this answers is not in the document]
Flash0ver: By design.
Structured Logs, when enabled, are not filter through a "minimum log level" (unlike both Events and Breadcrumbs).
Structured Logs are filtered later in the pipeline via a user callback configured through `SetBeforeSendLog`.

src/Sentry.Serilog/SentrySink.Structured.cs
```diff
@@ -0,0 +1,115 @@
...
+
+namespace Sentry.Serilog;
+
+internal sealed partial class SentrySink
+{
+    private static readonly SdkVersion Sdk = CreateSdkVersion();
+
+    private void CaptureStructuredLog(IHub hub, LogEvent logEvent, string formatted, string? template)
+    {
+        var traceHeader = hub.GetTraceHeader() ?? SentryTraceHeader.Empty;
```
jamescrosswell: Is the `ParentSpanId` required in the protocol for StructuredLogging? An empty SpanId probably isn't that useful otherwise eh?

src/Sentry.Serilog/SentrySink.Structured.cs
```diff
@@ -0,0 +1,115 @@
...
+                }
+            }
+            else if (!property.Value.IsNull())
+            {
+                yield return new KeyValuePair<string, object>(property.Key, property.Value);
+            }
+        }
+    }
+
+    private static SdkVersion CreateSdkVersion()
```
jamescrosswell: This method isn't doing much and is only used once (up on line 8). 

Could we remove the method and replace the call to it with it's implementation?

src/Sentry.Serilog/SentrySinkExtensions.cs
jamescrosswell: If we do add it, I'd say add it at the end. That way people's code won't break when they bump to the new version.

> [!NOTE]
>   Given the ridiculous number of parameters to this method, we should consider refactoring it into some kind of Builder in future versions.

src/Sentry.Serilog/SentrySinkExtensions.cs
jamescrosswell: My vote is to remove it. 

Unlike the overload above, this overload with _always_ be used in conjunction with some other call to initialise Sentry separately... so I think this one should be kept strictly to configuring the Serilog integration.

Otherwise, what if someone sets it to `true` in the call they make to initialise Sentry but `false` here? That just gets confusing and we have to invent rules to solve problems we didn't even have.

src/Sentry.Serilog/SentrySink.Structured.cs
Flash0ver: not quite, because `NameAndVersion` is also a static field in this type
```CSharp
private static readonly SdkVersion Sdk = new SdkVersion
{
  Name = SdkName,
  Version = NameAndVersion.Version, //Dereference of a possibly null reference
};
```
Although, Rider's/ReSharper's code analysis may be wrong here, potentially confused/caused by the `partial` type.

However, we are constructing and caching this object for an invocation of an `internal` method, anyway.
I'm now thinking about providing an overload to that internal method that takes SDK-Name and SDK-Version as two separate `string` parameters/arguments, and then just passing them directly.

src/Sentry.Serilog/SentrySink.Structured.cs
Flash0ver: No ... it's an optional "Default Attribute" ... the `trace_id` is _required_.

Oops ... I made this mistake in the Sentry-SDK and Microsoft.Extensions.Logging integrations, too.

I'll create a follow-up PR to fix this issue in all integrations.

test/Sentry.Serilog.Tests/SentrySerilogOptionsTests.cs
Flash0ver: removed

src/Sentry.Serilog/SentrySinkExtensions.cs
Flash0ver: Hmmm ... but now that I think about it again:
If this overload was used to configure the Serilog-Logger,
without an option to enable Structured-Logging (next to Events and Breadcrumbs),
then this overload just does not allow configuring our `SerilogSink` to also send Log Events as Structured Logs.

Or do I misunderstand something there?

src/Sentry.Serilog/SentrySink.Structured.cs
jamescrosswell: Hm, so this is what my LLM tells me about static initialisation order... 
> For a partial class, the compiler conceptually merges all parts into one declaration—but the order of parts is unspecified. That means if Sdk is in one file and NameAndVersion is in another (or even just separated strangely), the final textual order the compiler uses is not guaranteed. Relying on “the one I wrote first” can be fragile.

So the partial classes here may be confusing things. 

Possibly we could solve Rider's issues by moving the `Sdk` type into the same file as the other statics it depends on (that way we could control the order of initialisation):
```csharp
    private static readonly SdkVersion Sdk;
    internal static readonly SdkVersion NameAndVersion;

    static SentrySink()
    {
        NameAndVersion = typeof(SentrySink).Assembly.GetNameAndVersion();
        Sdk = new SdkVersion { Name = SdkName, Version = NameAndVersion.Version };
    }
```

Mike-E-angelo:
Thank you for your efforts here. 🙏  Is there a preview band we can try this out, perhaps?  I have two projects that would really benefit from having this in place.  Thank you for your continued efforts.

jacobsamo:
Looking forward to when this gets out, would really benefit my team

src/Sentry.Serilog/SentrySink.cs
```diff
@@ -67,6 +78,15 @@ public void Emit(LogEvent logEvent)
         }
     }
 
+    private bool IsEnabled(LogEvent logEvent)
+    {
+        var options = _hubAccessor()?.GetSentryOptions();
```
jamescrosswell:
```suggestion
        var options = _hubAccessor().GetSentryOptions();
```

_hubAccessor() is never null

This also looks unecessary:
https://github.com/getsentry/sentry-dotnet/blob/02defb5706520ee36c35897ca6abe0af6a3e005d/src/Sentry.Serilog/SentrySink.cs#L100-L101

src/Sentry.Serilog/SentrySink.cs
```diff
@@ -67,6 +78,15 @@ public void Emit(LogEvent logEvent)
...
+
+        return logEvent.Level >= _options.MinimumEventLevel
+            || logEvent.Level >= _options.MinimumBreadcrumbLevel
+            || options?.Experimental.EnableLogs is true;
```
jamescrosswell, line 87: I was a bit confused by this... so we don't filter any structured logs by Log Level - everything gets captured (all the way down to `SentryLogLevel.Trace`)?

Worth adding a code comment explaining why this is the case. I think we inherit the log level filter from MEL so we don't need to implement anything explicitly ourselves right?

src/Sentry.Serilog/SentrySink.cs
```diff
@@ -122,30 +142,34 @@ private void InnerEmit(LogEvent logEvent)
...
+                    ? exception?.Message ?? ""
+                    : formatted,
+                context,
+                data: data,
+                level: logEvent.Level.ToBreadcrumbLevel());
         }
 
-        Dictionary<string, string>? data = null;
-        if (exception != null && !string.IsNullOrWhiteSpace(formatted))
+        var options = hub.GetSentryOptions();
```
jamescrosswell, line 170: Maybe worth adding a note explaining why we fetch the options from the Hub here (rather than just using `_options` as we do above). I remember we had a conversation about this but I can't remember what the reason was anymore.

Flash0ver: good point ... done ... in cases where the `SentrySink` is added without a DSN (which does not initialize the SDK), then the `EnableLogs` option will be `default` ... so we read the options from the Hub that was actually initialized, which has the user-defined `EnableLogs` configured

src/Sentry.Serilog/SentrySink.cs
Flash0ver: That's how it's intended.
We currently have bug #4539 where the Configuration's Log-Level is not applied.
Will review and revise in the follow-up fix.

src/Sentry.Serilog/SentrySink.cs
Flash0ver: I haven't changed to logic about null-checking of the `IHub` ... just continued the existing pattern.
I'll review if this is a necessary null-check, or a redundant leftover.

Flash0ver: You are right ... in both cases (production: via `HubAdapter.Instance`; and also tests) the `Func<IHub>` never returns a null instance.
-> removed the redundant null-checks

src/Sentry.Serilog/SentrySink.cs
```diff
@@ -67,6 +78,15 @@ public void Emit(LogEvent logEvent)
...
     }
 
+    private bool IsEnabled(LogEvent logEvent)
+    {
+        var options = _hubAccessor().GetSentryOptions();
+
+        return logEvent.Level >= _options.MinimumEventLevel
+            || logEvent.Level >= _options.MinimumBreadcrumbLevel
+            || options?.Experimental.EnableLogs is true;
+    }
```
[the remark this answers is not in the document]
Flash0ver: Reviewed writes to and reads from the `Func<IHub> _hubAccessor` field:
- it's `private readonly`, so it can only be set from a .ctor
- both .ctors do set the field
- all .ctor invocations do pass a non-null-returning delegate

Also ... both the field and the `Func<IHub> hubAccessor` .ctor parameter are properly null-annotated,
so we would fail the build with a `nullable` warning as error if violated.

Mike-E-angelo:
Woohoo, thank you again for your amazing efforts out there, @Flash0ver, @jamescrosswell, and team!  This will really help out a lot with my projects. 🙏🙏🙏

Syed-RI:
release the kraken, I mean nuget package, please? Also idiot proof (I'm the idiot) docs please ❤️

Flash0ver:
We got 4 Structured-Logs-related fixes in the pipeline, small yet impactful changes, that I'd like to ship in one release. With 3 developers working on these, I'm confident with releasing early next week. Also, docs are ready to be published too (getsentry/sentry-docs#14693).

src/Sentry.Serilog/SentrySink.Structured.cs
Flash0ver: follow-up for both the SDK-Logger and the `Microsoft.Extensions.Logging` integration: #4565

Flash0ver:
[Release: 5.16.0](https://github.com/getsentry/sentry-dotnet/releases/tag/5.16.0)
Please let us know what you think: open an issue if you experience any problems, or [start a discussion to share your thoughts](https://github.com/getsentry/sentry-dotnet/discussions/4438)

Syed-RI:
@Flash0ver we we have to add `Sentry.Serilog` package and add that sink? I cant find it in the doc (told you I'm the idiot :P)

Issue #4506, opened 2025-09-09
Blazor WebAssembly Profiling Support


alexsohn1126:
### Problem Statement

We do not support Blazor WebAssembly profiling. We may be able to support Blazor profiling with WebAssembly performance profiling becoming available in .NET 10.

See:
* https://github.com/dotnet/runtime/issues/76316#issuecomment-2662463594
* https://learn.microsoft.com/en-us/aspnet/core/release-notes/aspnetcore-10.0?view=aspnetcore-9.0#blazor-webassembly-performance-profiling-and-diagnostic-counters

PR #4507, opened 2025-09-09
fix: Ensure profiling is disabled in Blazor WebAssembly

alexsohn1126:
THIS PULL REQUEST IS REPLACED BY #4512

# Problem
We do not support profiling in Blazor WebAssembly apps. However, it was possible to pass initialize a Blazor WASM app with profiling, which led to undefined behaviours. 

The issue above reported failing to publish Blazor WASM app when profiling is enabled. I was not able to replicate that, both in the current version of SDK, and also the version that the issue reporter was using (v4.9.0).

I did find that adding profiling in Blazor WASM led to errors in the browser:

image

# Solution

This pull request does not contain the implementation for enabling Blazor WASM profiling.

That will be handled in this issue:
* https://github.com/getsentry/sentry-dotnet/issues/4506

For now, we will check whether profiling is enabled. If so, we will print a message to the user saying:

Screenshot 2025-09-09 at 4 09 53 PM

And remove profiling integration from our SDK.

This removes the error mentioned above, and hopefully the build error as well.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,7 +16,7 @@
   - `InvalidOperationException` potentially thrown during a race condition, especially in concurrent high-volume logging scenarios ([#4428](https://github.com/getsentry/sentry-dotnet/pull/4428))
 - Blocking calls are no longer treated as unhandled crashes ([#4458](https://github.com/getsentry/sentry-dotnet/pull/4458))
 - Only apply Session Replay masks to specific control types when necessary to avoid performance issues in MAUI apps with complex UIs ([#4445](https://github.com/getsentry/sentry-dotnet/pull/4445))
-- Ensure profiling is disabled in Blazor WASM ([#4507](https://github.com/getsentry/sentry-dotnet/pull/4507))
+- Ensure profiling is disabled in Blazor WebAssembly ([#4507](https://github.com/getsentry/sentry-dotnet/pull/4507))
 
 ### Dependencies
 


src/Sentry.AspNetCore.Blazor.WebAssembly/WebAssemblyHostBuilderExtensions.cs
```diff
@@ -28,10 +30,27 @@ public static WebAssemblyHostBuilder UseSentry(this WebAssemblyHostBuilder build
...
+        {
+            return;
+        }
+
+        options.SetupLogging();
+        options.LogDebug("Detected Sentry profiling initialization in Blazor WebAssembly. " +
+                         "Sentry does not support Blazor WebAssembly profiling. Removing profiling integration. " +
+                         "Check https://github.com/getsentry/sentry-dotnet/issues/4506 for more information.");
+        // Ensure project doesn't have Profiling Integration
+        options.RemoveIntegration<ProfilingIntegration>();
```
jamescrosswell, line 52: Nice - thanks @alexsohn1126 !

We could also add a unit test to `SentryWebHostBuilderExtensionsTests` to make sure that the integration gets removed.

src/Sentry.AspNetCore.Blazor.WebAssembly/Sentry.AspNetCore.Blazor.WebAssembly.csproj
```diff
@@ -7,6 +7,7 @@
 
   <ItemGroup>
     <ProjectReference Include="..\Sentry.Extensions.Logging\Sentry.Extensions.Logging.csproj" />
+    <ProjectReference Include="..\Sentry.Profiling\Sentry.Profiling.csproj" />
```
Flash0ver, line 10: thought: new dependency

I'm a bit hesitant of adding this new dependency, as we might paint ourselves into a (potential) breaking change corner in the future:
Unless I see that wrong, we wouldn't implement getsentry/sentry-dotnet#4506 in `Sentry.Profiling`, but instead in `Sentry.AspNetCore.Blazor.WebAssembly`, or a new WASM-Profiling package, which would/could result into removing this dependency again ... so we could, strictly speaking, only change it in a major release.
Well ... now that I think about it again ... since `Sentry.Profiling` is not supported in `Sentry.AspNetCore.Blazor.WebAssembly` currently anyway, later removing `Sentry.Profiling` isn't actually really a breaking change.
So perhaps this is a non-issue.

The other thought that I had is that adding this dependency is increasing the size of the consuming Application, which especially for WASM apps is a concern where bundle sizes and download speeds to the Browser are worth a consideration ... because now every WASM app will depend on `Sentry.Profiling`, regardless of whether it actually tries to add the Profiling integration or not.
Which again isn't really a blocking issue either ... more of a "not-so-nice-to-have" thought.

Flash0ver: I am wondering if there is another way to signal and/or force-disable Profiling for WASM-apps 🤔.

- docs
  - we do mention that "Blazor WebAssembly" is not supported in our Profiling-Troubleshooting already
- MSBuild warning
  - I wonder if we could ship a custom Target in `Sentry.Profiling` that produces a build warning when `Sentry.Profiling` is used in a WASM app, and suggests to the user to remove the dependency again since it's currently not supported
- make `SentryOptionsProfilingExtensions.AddProfilingIntegration` and/or `ProfilingIntegration.Register` no-op and emitting the Diagnostic-Log if the `SentryOptions` type is `SentryBlazorOptions`
  - `Sentry.AspNetCore.Blazor.WebAssembly` has it's own `SentryBlazorOptions` type. We could instead move that check to `Sentry.Profiling`, to keep it "contained"
  - well to be fair ... this wouldn't be a Type-check due to the lacking dependency, but instead be a String-Type-Name check. So we should definitely have a unit test around that behavior. Not the most aesthetic solution either, but this way, we wouldn't "pessimize" every WASM project that doesn't even try to enable Profiling by now also including the dependency
- ...

Hmmm ... perhaps there is another, more elegant solution.

What do you think?

src/Sentry.AspNetCore.Blazor.WebAssembly/WebAssemblyHostBuilderExtensions.cs
```diff
@@ -28,10 +30,27 @@ public static WebAssemblyHostBuilder UseSentry(this WebAssemblyHostBuilder build
...
+
+    private static void RemoveBlazorProfilingIntegration(SentryBlazorOptions options)
+    {
+        if (!options.IsProfilingEnabled)
+        {
+            return;
+        }
+
+        options.SetupLogging();
+        options.LogDebug("Detected Sentry profiling initialization in Blazor WebAssembly. " +
```
Flash0ver, line 48: question: level

Since we are force-disabling an explicit opt-in from user code, I wonder if this may be of a higher Log-Level. Perhaps even _Warning_.
And it's only logged once per app lifetime, it wouldn't be too noisy, yet strongly suggesting to remove the `Sentry.Profiling` dependency from the consuming application, as it's currently not supported.

src/Sentry.AspNetCore.Blazor.WebAssembly/Sentry.AspNetCore.Blazor.WebAssembly.csproj
alexsohn1126: Great points Stefan, I'll go over some of the points that you made:

> I'm a bit hesitant of adding this new dependency, as we might paint ourselves into a (potential) breaking change corner in the future

Yes, I was also hesitant to make this dependency, since we really only use `Sentry.Profiling` to get the `ProfilingIntegration` type in line 52. It is also weird that we'd be adding profiling to a platform that we know that doesn't support profiling. I submitted this solution because I couldn't really think of another way to use `RemoveIntegration` function without importing the `ProfilingIntegration` type.

> MSBuild warning

I think the MSBuild warning is an interesting idea, we already do have custom Targets in `Sentry.Profiling` that errors out for IOS and Android platforms:

https://github.com/getsentry/sentry-dotnet/blob/1a6d7e7f1d478d77f8f7a18ac25baa7341c642ed/src/Sentry.Profiling/buildTransitive/Sentry.Profiling.targets#L6-L11

However, from a chat with @jamescrosswell, I learned that WASM doesn't really have a platform of it's own since it runs on the browser. So it will have to check for something entirely different, not platform.

I will try to explore more about custom Targets and how we can possibly use them in this context.

> make `SentryOptionsProfilingExtensions.AddProfilingIntegration` and/or `ProfilingIntegration.Register` no-op and emitting the Diagnostic-Log if the `SentryOptions` type is `SentryBlazorOptions`

Yes, this would definitely be a valid solution, and would keep the dependency away from the WASM project. I'll use this as a fallback in case I can't solve this problem with MSBuild solution that you mentioned. While I also do not particularly like the raw string check with types, I don't think there's a way around it, and it sounds like the lesser evil than creating another dependency.

I'll definitely add tests around that if I do end up going down this route!

src/Sentry.AspNetCore.Blazor.WebAssembly/WebAssemblyHostBuilderExtensions.cs
alexsohn1126: I was thinking about putting this message under `warning` level as well, so I will change the log to be a warning if I don't figure out how to make MSBuild solution work (mentioned above).

src/Sentry.AspNetCore.Blazor.WebAssembly/Sentry.AspNetCore.Blazor.WebAssembly.csproj
Flash0ver: hmmm ... re `MSBuild` ... your thoughts have brainstormed me into:
In the BCL, I noticed usages of `[SupportedOSPlatform("browser")]` and  `[UnsupportedOSPlatform("browser")]`, and I believe there is also a `BROWSER` preprocessor symbol.
That makes me wonder if there actually is a `TargetPlatformIdentifier == 'browser'` defined for WASM 🤔

checked grep.app ... the .NET Runtime has some usages
https://github.com/dotnet/runtime/blob/dcf86807937706d8d961a4aff0b98a8f55b563ba/src/libraries/System.Runtime.InteropServices.JavaScript/src/System.Runtime.InteropServices.JavaScript.csproj#L12-L15

If that works, it would be then a consistent solution with `android` indeed.

jamescrosswell: I just noticed this additional dependency when chatting with @Flash0ver last night... 

@alexsohn1126 I agree it's pretty weird to add a dependency on Profiling just to call a method that removes the profiling integration. This could potentially increase the size of our WASM integration (which would increase the size of the resulting WASM assemblies that our SDK users ship). 

I think we either find another way to detect whether someone has tried to enable profiling in WASM or we don't add this logic (we just leave it in the troubleshooting section of the docs instead).

jamescrosswell: Another option might be to find a way to do this differently (so that we don't need to have access to the `ProfilingIntegration` type in order to remove that integration type):
https://github.com/getsentry/sentry-dotnet/blob/eadf5f20b8014745d0fe42b0e707bf3f9eec9ffc/src/Sentry/SentryOptions.cs#L1342-L1346

We could, for example, add an internal `INamedSdkIntegration` and have the `ProfilingIntegration` implement this. Then we could do something like:
```
    internal void RemoveIntegrationByName(string name)
    {
        _integrations.RemoveAll(integration => integration is INamedSdkIntegration namedIntegration && namedIntegration.Name == name);
    }
```

alexsohn1126: From what I've tested so far, `TargetPlatformIdentifier` is not set when you in a Blazor WebAssembly application. If I try to print it out pre-build, it just prints a blank string, so I don't think that is possible.

I did find out we can use `UsingMicrosoftNETSdkBlazorWebAssembly` property:
https://github.com/dotnet/sdk/blob/main/src/BlazorWasmSdk/Sdk/Sdk.props#L14

I will make a change to the pull request whenever I can confirm that we can stop a user from building a wasm app with profiling. Though this would mean that they will be blocked from building the app which may not be the best experience. Then again, it makes sense to just not allow the user to use profiling in the first place if we don't support it 🤷‍♂️

If the method mentioned above doesn't work, what @jamescrosswell suggested looks pretty clean!

CHANGELOG.md
```diff
@@ -16,6 +16,7 @@
   - `InvalidOperationException` potentially thrown during a race condition, especially in concurrent high-volume logging scenarios ([#4428](https://github.com/getsentry/sentry-dotnet/pull/4428))
 - Blocking calls are no longer treated as unhandled crashes ([#4458](https://github.com/getsentry/sentry-dotnet/pull/4458))
 - Only apply Session Replay masks to specific control types when necessary to avoid performance issues in MAUI apps with complex UIs ([#4445](https://github.com/getsentry/sentry-dotnet/pull/4445))
+- Ensure profiling is disabled in Blazor WebAssembly ([#4507](https://github.com/getsentry/sentry-dotnet/pull/4507))
```
Flash0ver, line 19: note: needs to be updated after publishing release

- getsentry/publish#6257

Issue #4237, opened 2025-06-01
On .NET 9, Java exception on background thread double reports


bruno-garcia:
SDK version 5.9.0 (latest)

![Image](https://github.com/user-attachments/assets/8ff4cae8-ca33-4b10-8ade-899f11854bfd)

Originally, when we, added this method, it was only captured via the native Android SDK.
It seems at some point .NET started also capturing it.

But it's quite odd, because the one captured by .NET has no stack trace (so rather useless) 
And the screenshot seems to be sometime after the native one (not sure how many frames)

<details><summary>Managed details</summary>
<p>

![Image](https://github.com/user-attachments/assets/c08ed2db-f5b4-4486-bd79-9becfda16136)

```json
[...]

Issue #4174, opened 2025-05-07
Add e2e test


bruno-garcia:
We recently had symbolication related issues that would have been caught had we tested the SDK through Sentry itself.
Other repos have this, like React Native.

jamescrosswell:
@bruno-garcia is there a reference to the issue describing the symbolication difficulties? What kind of end to end tests are we talking about? Different to the integration tests?

vaind:
@jamescrosswell this came up in an offline meeting. The issue with current integration tests in the dotnet repo is that they only run locally, against a fake sentry server (a python script that prints out what it has received). This is fast and provides baseline checks what has been sent to the server by the CLI during build.
However, because the dotnet SDK that depends on server-side symbolication, it would be preferable to verify that it continues to work and doesn't break between .net versions, as it did in the past. This can be achieved by building a sample app, uploading it's debug files to sentry.io, running it, and checking sentry.io APIs to fetch the issue. There, we can verify the issue is symbolicated properly. I imagine this would be useful for native AOT, portable PDBs support, etc.
Similar tests have been done for other things than symbolication, for example replay. Maybe this would be useful to test profile capture, which also got broken by a past .net release.

jamescrosswell:
Cool, thanks @vaind - that sounds like a big step up.

Flash0ver:
Consideration:
Once we come to Android E2E-Testing, perhaps we can verify _ProGuard_ as well.
See also getsentry/sentry-dotnet#4647.

PR #4509, opened 2025-09-10
fix: de-duplicate Java.Lang.RuntimeException on Android

jpnurmi:
Java runtime exceptions are caught in two layers:
- Sentry Java SDK
- .NET Java Interop (JNI)

For example, the result of exactly one `CrashType.JavaThread` and one `CrashType.JavaBackgroundThread`:

| Before: 2+2 :x: | After: 1+1 :white_check_mark: |
|---|---|
| image |  Screenshot From 2025-09-10 12-46-30 |

Fixes: #4237

### DIFF
--- a/src/Sentry/Platforms/Android/SentrySdk.cs
+++ b/src/Sentry/Platforms/Android/SentrySdk.cs
@@ -168,6 +168,9 @@ public static partial class SentrySdk
             SentryAndroid.Init(AppContext, configuration);
         }
 
+        // Don't capture Java Runtime exceptions in the managed SDK, since we already capture them in the native SDK
+        options.AddExceptionFilterForType<Java.Lang.RuntimeException>();
+
         // Set options for the managed SDK that depend on the Android SDK. (The user will not be able to modify these.)
         options.AddEventProcessor(new AndroidEventProcessor(nativeOptions!));
         if (options.Android.LogCatIntegration != LogCatIntegrationType.None)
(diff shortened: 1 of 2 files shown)


src/Sentry/Platforms/Android/SentrySdk.cs
```diff
@@ -177,6 +177,8 @@ private static void InitSentryAndroidSdk(SentryOptions options)
         options.CrashedLastRun = () => JavaSdk.Sentry.IsCrashedLastRun()?.BooleanValue() is true;
         options.EnableScopeSync = true;
         options.ScopeObserver = new AndroidScopeObserver(options);
+        // Don't capture Java Runtime exceptions in the managed SDK, since we already capture them in the native SDK
+        options.AddExceptionFilterForType<Java.Lang.RuntimeException>();
```
Flash0ver, line 181: question: test

Is there a (somewhat) straightforward way to (unit) test the suppression of the double capture?

Perhaps via
```CSharp
#if ANDROID
    SentrySdk.CauseCrash(CrashType.Java);
#endif
```
or would this need to be an integration test?

jpnurmi: I couldn't find any suitable place for testing this. Native SDKs are not initialized in unit tests, and integration tests only run on desktop platforms. Once we have the infrastructure for e2e tests, though, then we can hopefully start testing these things:
- https://github.com/getsentry/sentry-dotnet/issues/4174

PR #4513, opened 2025-09-11
fix: Do not overwrite user ip if set manually in ASP.NET sdk

### LINKED ISSUE #3337: Sentry dotnet SDK is blowing away User.IpAddress if we set it explicitly and SendDefaultPii is true
### Package

Sentry

### .NET Flavor

.NET

### .NET Version

4.7.2

### OS

Windows

### SDK Version

4.4.0

_No response_

### Steps to Reproduce

1. Set SendDefaultPii to true during SentrySdk.Init call
2. Create new SentryEvent - and set the event's User.IpAddress to a specific value.
3. Call SentrySdk.CaptureEvent and pass in the event.

### Expected Result

The Sentry dashboard should show the User's IP as the one we set.

### Actual Result

It appears to always contain an internal/server IP instead. I believe this may be because of this code? https://github.com/getsentry/sentry-dotnet/blob/5bcb9cac8b54f283a4ce5cc16573bfdf2970cdf9/src/Sentry.AspNet/Internal/SystemWebRequestEventProcessor.cs#L75 

But I am honestly not certain.

If I am doing something wrong here, would love guidance. But we have our own logic that fires off before a Sentry event sends - and if it's inside a web requests - gets the current user context (who's logged in) and IP address data, then stamps it onto the SentryEvent.

alexsohn1126:
Fixes #3337

# Problem

We were overwriting Event's user IP address in ASP.NET if it was set manually.

# Solution

We only populate an event's user ip if it was not set before.

### DIFF
--- a/test/Sentry.AspNet.Tests/Internal/SystemWebRequestEventProcessorTests.cs
+++ b/test/Sentry.AspNet.Tests/Internal/SystemWebRequestEventProcessorTests.cs
@@ -79,6 +79,8 @@ public class SystemWebRequestEventProcessorTests :
         evt.User.IpAddress = userIp;
 
         Context = HttpContextBuilder.Build();
+        // Ensure user is not null
+        Context.User = new GenericPrincipal(new GenericIdentity("TestUser"), null);
         _fixture.SentryOptions.SendDefaultPii = true;
         var sut = _fixture.GetSut();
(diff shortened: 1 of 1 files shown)


alexsohn1126:
bugbot review

Issue #3369, opened 2024-05-14
Could not get pinvoke, or callbacks for method 'Sentry.Native.C::sentry_options_set_logger'


klemmchr:
### Package

Sentry

### .NET Flavor

.NET

### .NET Version

8.0.5

### OS

Any (not platform specific)

### SDK Version

4.6.2

### Steps to Reproduce

1. Compile a Blazor project with AoT enabled (Repro from #3263 should work)
2. Observe warning in build

### Expected Result

No warning

### Actual Result

During AoT the following warning is emitted

```
[WRN] /usr/share/dotnet/packs/Microsoft.NET.Runtime.WebAssembly.Sdk/8.0.5/Sdk/WasmApp.Native.targets(338,5): warning WASM0001: Could not get pinvoke, or callbacks for method 'Sentry.Native.C::sentry_options_set_logger' because 'Parsing function pointer types in signatures is not supported.' [/builds/acodia-care/acodia-care/src/Web/Web.Blazor/Web.Blazor.csproj]
```

The app itself works fine, I don't observe any errors in console or during build.

bitsandfoxes:
@vaind do you think we can disable the warning now that we're guarding and not initializing `sentry-native` on Blazor? It does sound like a potential foot gun down the road tho.

vaind:
I don't think we can do anything else than disable the warning. There's only a runtime check for us to use: https://github.com/dotnet/aspnetcore/issues/38667#issuecomment-981099588 thus we cannot get rid of the warning by not including the code at all.

jamescrosswell:
Closing as this issue seems to have been resolved now. Please feel free to reopen if you believe otherwise.

klemmchr:
Which version is supposed to fix this issue? We're running 5.14.1 but we still get a lot of WASM warnings.

```
13:11:23 [WRN] /root/.dotnet/packs/Microsoft.NET.Runtime.WebAssembly.Sdk/9.0.9/Sdk/WasmApp.Common.targets(793,5): warning WASM0001: Could not get pinvoke, or callbacks for method 'Sentry.Native.C::sentry_value_new_object' because 'System.NotSupportedException: Unsupported parameter type in method 'Sentry.Native.C.sentry_value_new_object' [/builds/acodia-care/acodia-care/src/Web/Web.Blazor/Web.Blazor.csproj]
[...]

PR #4519, opened 2025-09-12
fix: Stop warnings from showing in Blazor WASM projects

alexsohn1126:
fixes: #3369

# Problem

Warning appears when publishing Blazor WASM apps with AOT enabled

I thought I could pass in additional properties when referencing Sentry from Blazor WASM sdk like this:

```
    <ProjectReference Include="..\Sentry\Sentry.csproj">
      <AdditionalProperties>UsingSentryBlazorWebAssemblySdk=true</AdditionalProperties>
    </ProjectReference>
```

Then we can set a compile time processor directive referencing enabled when `UsingSentryBlazorWebAssemblySdk` is present, but I kept getting the warnings regardless of the things I've tried.

So I just suppressed the warning like mentioned in #3369.

### DIFF
--- a/src/Sentry/Platforms/Native/CFunctions.cs
+++ b/src/Sentry/Platforms/Native/CFunctions.cs
@@ -328,12 +328,18 @@ internal static class C
     // Blazor WASM interop generators.
     internal struct sentry_value_t
     {
-        internal ulong Bits { get; set; }
+        private ulong _bits;
+
+        internal ulong Bits
+        {
+            readonly get => _bits;
+            set => _bits = value;
+        }
 
         internal double Double
         {
-            readonly get => BitConverter.UInt64BitsToDouble(Bits);
-            set => Bits = BitConverter.DoubleToUInt64Bits(value);
+            readonly get => BitConverter.UInt64BitsToDouble(_bits);
+            set => _bits = BitConverter.DoubleToUInt64Bits(value);
         }
     }
 


src/Sentry/SentrySdk.cs
```diff
@@ -69,7 +69,7 @@ internal static IHub InitHub(SentryOptions options)
             InitSentryCocoaSdk(options);
 #elif ANDROID
             InitSentryAndroidSdk(options);
-#elif NET8_0_OR_GREATER
+#elif NET8_0_OR_GREATER && !_SENTRY_BLAZOR_WASM
```
alexsohn1126: Not sure if adding this is even necessary, we already have `SentryNative.IsAvailable` check there?

alexsohn1126: this is necessary, otherwise the warning appears

src/Sentry.AspNetCore.Blazor.WebAssembly/buildTransitive/Sentry.AspNetCore.Blazor.WebAssembly.targets
```diff
@@ -0,0 +1,8 @@
+<Project>
+
+  <!-- Suppress WASM0001 warning for Blazor WebAssembly projects using AOT compilation -->
+  <!-- This warning is expected when using Sentry with Blazor WebAssembly AOT due to reflection usage -->
+  <PropertyGroup Condition="'$(UsingMicrosoftNETSdkBlazorWebAssembly)' == 'true'">
+    <NoWarn>$(NoWarn);WASM0001</NoWarn>
```
jamescrosswell: I'm a little bit nervous about disabling this globally. In our case (if I've understood) the code in question never gets called at runtime, so we can safely ignore it... but there are going to be scenarios where that warning would alert SDK users to genuine concerns. 

I wonder if we could be a bit more specific, so that we only disable the warning for our package/code?

I'm not sure if these would work but we might be able to do:
```csharp
[UnsupportedOSPlatform("browser")]
internal static void InitNativeStuff()
{
    if (!OperatingSystem.IsBrowser())
    {
        // native/PInvoke etc.
    }
}
```

Or alternatively:
```csharp
#pragma warning disable WASM0001
internal static void InitNativeStuff() { /* ... */ }
#pragma warning restore WASM0001
```

Providing we can identify the specific code that is triggering the warnings...

Flash0ver: I have a similar concern: masking the issue too broadly.

I'm wondering if this is a more general concern, and not just Blazor WASM, when it comes to NativeAOT and `DllImportAttribute`.
Perhaps this could be fixed by using the `net7.0` Source-Generators via `LibraryImport`? But then I'm wondering why we haven't hit the problem with other app-models yet 🤔 ... also this would be a bigger issue with way more testing involved.

alexsohn1126: I think the code that is causing the warnings are all in here:

https://github.com/getsentry/sentry-dotnet/blob/7061fb92d58ecd138228e3689a97c72fdab3b6e4/src/Sentry/Platforms/Native/CFunctions.cs#L11-L20

I can try both of them and see what works. I don't think `UnsupportedOSPlatform("browser")` works if they don't have the TFM set to `net9.0-browser` according to [this docs](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1416#tfm-target-platforms). Blazor WASM applications [don't have to use the `browser` TFM](https://github.com/dotnet/designs/blob/main/accepted/2023/net8.0-browser-tfm.md).

So I'm leaning more towards the second option. I'll try that and report back.

Flash0ver: now I'm not sure if I'm jumping the gun with the source-generated P/Invoke ... since the scope of this is the entire SDK, I suppose this would be another issue

on second though, I'm now leaning towards a non-invasive quick fix for this issue, that we then follow-up in an overarching fashion with source-generated P/Invoke

alexsohn1126:
Weird observation I made between .NET 8 vs. 9.

If .NET 8 is used to make the Blazor WebAssembly project, it shows these warnings:

```
Sentry.Samples.AspNetCore.Blazor.Wasm succeeded with 3 warning(s) (56.6s)
    /usr/local/share/dotnet/packs/Microsoft.NET.Runtime.WebAssembly.Sdk/8.0.19/Sdk/WasmApp.Native.targets(338,5): warning WASM0001: Could not get pinvoke, or callbacks for method 'Sentry.Native.C::sentry_transport_new' because 'Parsing function pointer types in signatures is not supported.'

[same error in different functions that use `delegate *` in their parameters in Cfunctions.cs]
```

However, if the Blazor WebAssembly project is in .NET 9:

```
Sentry.Samples.AspNetCore.Blazor.Wasm succeeded with 16 warning(s) (60.3s)
    /usr/local/share/dotnet/packs/Microsoft.NET.Runtime.WebAssembly.Sdk/9.0.8/Sdk/WasmApp.Common.targets(793,5): warning WASM0001: 
      Could not get pinvoke, or callbacks for method 'Sentry.Native.C::sentry_value_new_object' because 'System.NotSupportedException: Unsupported parameter
       type in method 'Sentry.Native.C.sentry_value_new_object'
         at PInvokeCollector.<CollectPInvokes>g__CollectPInvokesForMethod|6_0(MethodInfo method, <>c__DisplayClass6_0&)
         at PInvokeCollector.CollectPInvokes(List`1 pinvokes, List`1 callbacks, HashSet`1 signatures, Type type)'

[same errors for many of the functions that use sentry_value_t type in Cfunctions.cs]
```

First, it's weird that the warnings disappeared for function pointer type warnings in .NET 9, and new warnings appeared out of nowhere.

Why is this the case? I'll dig in deeper, perhaps we can figure out a better solution to this.

alexsohn1126:
Had a chat with @Flash0ver about this issue, seems like there is more caveat to this than it first seems.

We have tried targeting `net8.0-browser`, and using `BROWSER` constant to block out some code from being compiled, but that did not work as long as the Blazor WebAssembly project's `TargetFramework` was `net8.0`. Even if the project's `TargetFramework` was `net8.0-browser`, it would cause build errors which we didn't have time to resolve. Also, it would be undesirable to tell the user "Hey, change the `TargetFramework`!", as they could be using their own target framework for whatever reason.

We also brainstormed the idea of possibly adding Roslyn analyzers to suppress these warnings conditionally. We haven't tried it, but it could be an option.

Using `#pragma warning disable` was not effective because the warnings in this issue occurs in Microsoft WASM sdk.

If there are any ideas, it would be great!

jpnurmi:
What if the native bindings were moved into a separate `Sentry.Bindings.Native` package? Would it help if it was only referenced on relevant platforms?

alexsohn1126:
> What if the native bindings were moved into a separate `Sentry.Bindings.Native` package? Would it help if it was only referenced on relevant platforms?

Ooh, this sounds promising. I'll look into this more!

jamescrosswell:
> Using `#pragma warning disable` was not effective because the warnings in this issue occurs in analyzers.

Another option might be the [SuppressMessageAttribute](https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.codeanalysis.suppressmessageattribute?view=net-9.0).

If worst comes to worst we can simply document this in the troubleshooting section of our docs and recommend users include the Sentry package like so, for WebAssembly apps:
```
<ItemGroup>
  <PackageReference Include="Sentry Version="1.0.0" NoWarn="WASM0001" />
</ItemGroup>
```

That way the warning would be suppressed for Sentry but not for any other packages.

Flash0ver:
I do like the separation into a new `Sentry.Bindings.Native`, but we would also need dedicated `net8.0-browser`/`net9.0-browser` targets, so that we can conditionally include/exclude the dependency.

I also like the `SuppressMessageAttribute` idea ... that would be the simplest solution ... but `SuppressMessageAttribute` will not be part of the output compilation / assembly, because it has a `ConditionalAttribute`.
However, the [UnconditionalSuppressMessageAttribute](https://learn.microsoft.com/dotnet/api/system.diagnostics.codeanalysis.unconditionalsuppressmessageattribute) is preserved in the output compilation / assembly.

Cool ... I knew you can give `Aliases` to a `PackageReference`, but I didn't know you can also do an isolated `NoWarn` 💡

alexsohn1126:
tldr; `UnconditionalSuppressMessageAttribute` and `SuppressMessageAttribute` do not work to suppress these WASM0001 warnings.

After trying out both [UnconditionalSuppressMessageAttribute](https://learn.microsoft.com/dotnet/api/system.diagnostics.codeanalysis.unconditionalsuppressmessageattribute) and [SuppressMessageAttribute](https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.codeanalysis.suppressmessageattribute?view=net-9.0), I found out these warnings cannot be suppressed by neither of them. 

`WASM0001` warnings are triggered by the task `ManagedToNativeGenerator`:

image

Looking at the source code for `ManagedToNativeGenerator` ([link](https://github.com/dotnet/runtime/blob/2f5ffdae88a7028db1ef10a700c14d940b7fe69d/src/tasks/WasmAppBuilder/ManagedToNativeGenerator.cs#L83)), we can see that it scans .dll files. 

Referring back to the .NET9.0 warnings, the warning is from `CollectPInvokes` function. In v8.0.16 of the file that contains the function, we can see where those .NET8.0 warnings originated from: 
- https://github.com/dotnet/runtime/blob/v8.0.16/src/tasks/WasmAppBuilder/PInvokeCollector.cs#L212

All previous .NET8.0 versions do not have this warning, and all .NET9.0 versions do not have this warning, so that explains why we were seeing the different warnings between .NET8 and .NET9.

And the new errors we encounter are probably from these lines of code that was added:
- ﻿https://github.com/dotnet/runtime/commit/ed9f475dde7f3cb0e05a7613e82d69bd1a42ce0e#diff-429d68fa67442212800d3b0c7ec9a048510188e463a890d2865fa04833e7a9c5R60-R66


All that is to say: 

WASM0001 is from build tasks from scanning assemblies (yes, i was wrong on saying it was from analyzers 😮‍💨), thus we cannot use `UnconditionalSuppressMessageAttribute` nor `SuppressMessageAttribute` to suppress these warnings. These suppression attributes only work with AOT/trimmer warnings, not WASM sdk triggered by MSBuild task like in this case...

We can either ask the users to mute the warnings just for Sentry like @jamescrosswell mentioned, or we can go with the refactoring - making `Sentry.Bindings.Native`.

alexsohn1126:
@Flash0ver 

> I do like the separation into a new Sentry.Bindings.Native, but we would also need dedicated net8.0-browser/net9.0-browser targets, so that we can conditionally include/exclude the dependency.

Can't we use the [`UsingMicrosoftNETSdkBlazorWebassembly`](https://github.com/dotnet/sdk/blob/main/src/BlazorWasmSdk/Sdk/Sdk.props#L14) property?

kg:
> 
> First, it's weird that the warnings disappeared for function pointer type warnings in .NET 9, and new warnings appeared out of nowhere.
> 
> Why is this the case? I'll dig in deeper, perhaps we can figure out a better solution to this.

It's been a while so my memory isn't perfect, but essentially the NET8 version of this analyzer and the associated code generator had a handful of bugs. We tried to fix all the bugs we could find while also making the analyzer more robust and consistent. That's why there are more warnings now, most likely - that stuff might have been silently broken before. One of the things we did is try to conform more closely to [the wasm C ABI](https://github.com/WebAssembly/tool-conventions/blob/main/BasicCABI.md).

From looking at your code one thing you could try is changing sentry_value_t so it's not a union - that may make the interop generator stuff be able to understand it correctly. Instead of using `FieldOffset(0)` you could use `Unsafe.As`, I think (preferably by wrapping it in helper property accessors).

jamescrosswell:
> From looking at your code one thing you could try is changing sentry_value_t so it's not a union - that may make the interop generator stuff be able to understand it correctly. Instead of using FieldOffset(0) you could use Unsafe.As, I think (preferably by wrapping it in helper property accessors).

This definitely sounds like a better solution, if we can make that work for us...

> or we can go with the refactoring - making Sentry.Bindings.Native

I can't quite wrap my head around how that would work... where/when would the decision to include Sentry.Bindings.Native get made? We have to delay this until the SDK user is building their solution and we know whether or not the target is WASM right?

Flash0ver:
> Can't we use the [UsingMicrosoftNETSdkBlazorWebassembly](https://github.com/dotnet/sdk/blob/main/src/BlazorWasmSdk/Sdk/Sdk.props#L14) property?

Not sure how to solve this with an MSBuild property. In a NuGet package, we can have `/lib/{TFM}/*.dll` to target several versions, where each version can have varying References/Versions to other Assemblies/Packages. This is resolved at build/pack time. We can't dynamically change a Dependency of an already built Assembly in consuming projects ... unless we load Assemblies dynamically (via Reflection / AssemblyLoadContext) which I would like to avoid since this may be another can of worms regarding Trimming and AOT.

alexsohn1126:
> From looking at your code one thing you could try is changing sentry_value_t so it's not a union - that may make the interop generator stuff be able to understand it correctly. Instead of using FieldOffset(0) you could use Unsafe.As, I think (preferably by wrapping it in helper property accessors).

Thank you so much for the detailed insight! I will try changing `sentry_value_t`, and see if we can come up with a fix.

> where/when would the decision to include Sentry.Bindings.Native get made? We have to delay this until the SDK user is building their solution and we know whether or not the target is WASM right?

I thought we could use the property that comes with the Microsoft Blazor WASM sdk.. Apparently not.

> We can't dynamically change a Dependency of an already built Assembly in consuming projects ... unless we load Assemblies dynamically (via Reflection / AssemblyLoadContext) which I would like to avoid since this may be another can of worms regarding Trimming and AOT.

Yea, I suppose that means we can't use that property to stop this warning using that blazor wasm sdk property.. I thought maybe we could do something like this:

```xml
<ItemGroup Condition="'$(UsingMicrosoftBlazorSdk)' != 'true''">
  <ProjectReference Include="..\Sentry.Bindings.Native.csproj" />
</ItemGroup>
```

but if that's resolved at build time, then this would be useless.

Flash0ver:
Yes, the `ProjectReference`s are resolved at `build`-time (also during `restore` and `pack`).

The `Sentry` package would have a dependency on `Sentry.Bindings.Native`, to not break existing scenarios.
So when `Sentry.AspNetCore.Blazor.WebAssembly` has a dependency on `Sentry`, it also transitively depends on `Sentry.Bindings.Native`.

Unless we can use the `net8.0-browser` and `net9.0-browser` TFMs, which - for `Sentry` - would not include the dependency to `Sentry.Bindings.Native`, and `Sentry.AspNetCore.Blazor.WebAssembly` would target these instead of `net9.0;net8.0`. But if I recall correctly, we gave that a brief try and it did not work as expected ... unless we have missed something in that quick test.

Another consideration could be a breaking change, so that we split into `Sentry` and `Sentry.Bindings.Native`, but `Sentry` does not depend on `Sentry.Bindings.Native`, so users would need to explicitly opt-in by also referencing the package `Sentry.Bindings.Native`. But I am a bit uncertain about the tradeoff regarding the impact.

Also, I am quite surprised that the `[UnconditionalSuppressMessage]` Attribute does not work for this case 🤔

alexsohn1126:
After changing `sentry_value_t` implementation to remove `FieldOffset` attribute, I was able to get rid of the warnings in .NET 9.0.

I do not see the value in trying to removing .NET 8.0 warnings as it was a WASM app builder limitation. WASM app builder was improved in .NET 9.0 to handle function pointers, and those warnings are now gone (even before this fix).

src/Sentry/Platforms/Native/CFunctions.cs
```diff
@@ -322,14 +322,19 @@ private static Dictionary<long, DebugImage> LoadDebugImagesOnce(IDiagnosticLogge
...
+    // Implemented with a single ulong backing field and BitConverter
+    // to reinterpret values, since explicit unions cause issues with
+    // Blazor WASM interop generators.
     internal struct sentry_value_t
     {
-        [FieldOffset(0)]
-        internal ulong _bits;
-        [FieldOffset(0)]
-        internal double _double;
+        internal ulong Bits { get; set; }
```
Flash0ver: thought: change to Property with Backing Field (for explicitness)

I'm wondering if, for interop types, we should avoid Auto-Properties,
and instead have a property with `private ulong _bits` as explicit backing field,
which the `Double` property should also be using in its `get`/`set`.

I can imagine this is mainly a choice of style.
@vaind @jpnurmi what do you think?

jpnurmi: I like the idea. Makes it easier to see the memory layout.

alexsohn1126: Fair point! I feel like as a reader, what you are suggesting makes it easier to understand too.

Changed in https://github.com/getsentry/sentry-dotnet/pull/4519/commits/78ea186165398d11ba060e5b0ec7aefb9fb0729f

Issue #4520, opened 2025-09-14
iOS: cannot suppress `SIGABRT` nor `EXC_BAD_ACCESS`


jpnurmi:
### Package

Sentry

### .NET Flavor

.NET

### .NET Version

9.0.304

### OS

iOS

### Development Environment

Rider 2024 (MacOS)

### SDK Version

main (3f12094)

### Self-Hosted Sentry Version

25.8.0

### Workload Versions

```
Using workload version 9.0.304, which was specified in the global.json file at /Users/jpnurmi/Projects/sentry/sentry-dotnet/global.json.

Installed Workload Id      Manifest Version       Installation Source
---------------------------------------------------------------------
android                    35.0.92/9.0.100        SDK 9.0.300
ios                        18.5.9215/9.0.100      SDK 9.0.300
maccatalyst                18.5.9215/9.0.100      SDK 9.0.300
macos                      15.5.9215/9.0.100      SDK 9.0.300
maui-android               9.0.82/9.0.100         SDK 9.0.300
maui-ios                   9.0.82/9.0.100         SDK 9.0.300
maui-maccatalyst           9.0.82/9.0.100         SDK 9.0.300
maui-tizen                 9.0.82/9.0.100         SDK 9.0.300
wasm-tools-net8            9.0.8/9.0.100          SDK 9.0.300
```

### UseSentry or SentrySdk.Init call

```diff
diff --git a/samples/Sentry.Samples.Maui/MauiProgram.cs b/samples/Sentry.Samples.Maui/MauiProgram.cs
index 1a8a6a30..69aab6b7 100644
--- a/samples/Sentry.Samples.Maui/MauiProgram.cs
+++ b/samples/Sentry.Samples.Maui/MauiProgram.cs
@@ -67,6 +67,9 @@ public static MauiApp CreateMauiApp()
                     // Return true to capture or false to prevent the capture
                     return true;
                 });
+
+                options.Native.SuppressSignalAborts = true;
+                options.Native.SuppressExcBadAccess = true;
             })

             .ConfigureFonts(fonts =>
```

### `SIGABRT`
1. Enable `options.Native.SuppressSignalAborts`
2. Run `Sentry.Samples.Maui`
3. Press _Throw Unhandled .NET Exception (Crash)_
4. Run `Sentry.Samples.Maui` again
5. Observe Sentry

### `EXC_BAD_ACCESS`
1. Enable `options.Native.SuppressExcBadAccess`
[...]

PR #4521, opened 2025-09-14
fix: `SentryOptions.Native.SuppressSignalAborts` and `SuppressExcBadAccess` on iOS

jpnurmi:
This PR fixes a regression in `SentryOptions.Native.SuppressExcBadAccess` and `SentryOptions.Native.SuppressSignalAborts`. Let `ProcessOnBeforeSend` filter out `SIGABRT` and `EXC_BAD_ACCESS` sent early on startup during the Cocoa SDK init, even before the .NET Hub instance has been created.

Fixes: #4520

### DIFF
--- a/src/Sentry/Platforms/Cocoa/SentrySdk.cs
+++ b/src/Sentry/Platforms/Cocoa/SentrySdk.cs
@@ -195,15 +195,15 @@ public static partial class SentrySdk
         => ProcessOnBeforeSend(options, evt, CurrentHub);
 
     /// <summary>
-    /// This overload allows us to inject an IHub for testing. During normal execution, the CurrentHub is used.
-    /// However, since this class is static, there's no easy alternative way to inject this when executing tests.
+    /// Apply suppression logic for redundant native `SIGABRT` and `EXC_BAD_ACCESS` crash events
+    /// that have already been captured as managed exceptions by the Sentry.NET SDK to avoid sending
+    /// duplicate events to Sentry - once managed and once native.
+    ///
+    /// The managed exception is what a .NET developer would expect, and it is sent by the Sentry.NET SDK
+    /// But we also get a native SIGABRT since it crashed the application, which is sent by the Sentry Cocoa SDK.
     /// </summary>
-    internal static CocoaSdk.SentryEvent? ProcessOnBeforeSend(SentryOptions options, CocoaSdk.SentryEvent evt, IHub hub)
+    private static bool SuppressNativeCrash(SentryOptions options, CocoaSdk.SentryEvent evt)
     {
-        // When we have an unhandled managed exception, we send that to Sentry twice - once managed and once native.
-        // The managed exception is what a .NET developer would expect, and it is sent by the Sentry.NET SDK
-        // But we also get a native SIGABRT since it crashed the application, which is sent by the Sentry Cocoa SDK.
-
         // There should only be one exception on the event in this case
         if ((options.Native.SuppressSignalAborts || options.Native.SuppressExcBadAccess) && evt.Exceptions?.Length == 1)
         {
@@ -219,7 +219,7 @@ public static partial class SentrySdk
                 // Don't send it
                 options.LogDebug("Discarded {0} error ({1}). Captured as  managed exception instead.", ex.Type,
                     ex.Value);
-                return null!;
+                return true;
             }
 
             // Similar workaround for NullReferenceExceptions. We don't have any easy way to know whether the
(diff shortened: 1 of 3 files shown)


src/Sentry/Platforms/Cocoa/SentrySdk.cs
```diff
@@ -200,11 +200,6 @@ private static CocoaSdk.SentryHttpStatusCodeRange[] GetFailedRequestStatusCodes(
     /// </summary>
     internal static CocoaSdk.SentryEvent? ProcessOnBeforeSend(SentryOptions options, CocoaSdk.SentryEvent evt, IHub hub)
     {
-        if (hub is DisabledHub)
```
jamescrosswell, line 203: When would this be a problem? If the SDK is disabled, Sentry shouldn't be capturing any events at all right (either in the .NET or the Cocoa SDK)?

jpnurmi: It's a problem on startup when the previous run crashed with a duplicate `SIGABRT` or `EXC_BAD_ACCESS` that needs to be suppressed. Even if the .NET SDK is not (yet) in a fully initialized state, `ProcessOnBeforeSend` needs to return `null` for the suppression mechanism to work:

```cs
static partial class SentrySdk
{
    internal static IHub InitHub(SentryOptions options)
    {
        // ...

        if (options.InitNativeSdks)
        {
            InitSentryCocoaSdk(options); // <== ProcessOnBeforeSend for SIGABRT/EXC_BAD_ACCESS from the previous run
        }

        // We init the hub after native SDK in case the native init needs to adapt some options.
        var hub = new Hub(options);
        // ...
        return hub;
    }
}
```

jamescrosswell: I see... subtle. So the Cocoa SDK is already sending events before the .NET SDK has even been initialised!

```
// We init the hub after native SDK in case the native init needs to adapt some options.
```

I wonder what that's all about. @vaind do you recall (git blames you for that line 😜)

src/Sentry/Platforms/Cocoa/SentrySdk.cs
```diff
@@ -239,6 +234,11 @@ private static CocoaSdk.SentryHttpStatusCodeRange[] GetFailedRequestStatusCodes(
             }
         }
 
+        if (hub is DisabledHub)
```
jamescrosswell, line 255: Maybe worth referencing or summarising https://github.com/getsentry/sentry-dotnet/pull/4521#discussion_r2347616896 in the code here to explain why this happens after the suppression code above.

As an aside, this is a bit of a gotcha for custom event processes that users implement (and which get run below). They may not to work for native exceptions that crash the process, since events for those will likely get fired off before the .NET SDK has been initialised, the next time the process runs.

jpnurmi: The check was added in #4318 but I wonder if it's needed at all? Tests pass without the check, but maybe there's some scenario that is not unit testable.

jamescrosswell: If the hub is disabled then I guess it saves us doing all this:
https://github.com/getsentry/sentry-dotnet/blob/67473c59a0b79e164441167d3af9589521da5027/src/Sentry/Platforms/Cocoa/SentrySdk.cs#L279-L288

That will return an empty array of processors ([ConfigureScope is a NoOp for the DisabledHub](https://github.com/getsentry/sentry-dotnet/blob/67473c59a0b79e164441167d3af9589521da5027/src/Sentry/Extensibility/DisabledHub.cs#L36-L38))... so it's an early exit really (marginally more efficient).

src/Sentry/Platforms/Cocoa/SentrySdk.cs
vaind: I don't recall specifically but what this is trying to get along, IMO, is that the `InitSentryCocoaSdk()` (in case of iOS) must be called before `new Hub(options);` because the [former changes `options`](https://github.com/getsentry/sentry-dotnet/blob/af70fb30d878335ea152c469ad6c82b45a4243d1/src/Sentry/Platforms/Cocoa/SentrySdk.cs#L11) which the latter uses.

jamescrosswell: Hm, thanks @vaind. The only managed options that get set by the Cocoa SDK are these:
https://github.com/getsentry/sentry-dotnet/blob/cc613c901d5cbac3a4856156128a26847456e6ec/src/Sentry/Platforms/Cocoa/SentrySdk.cs#L15-L17
https://github.com/getsentry/sentry-dotnet/blob/cc613c901d5cbac3a4856156128a26847456e6ec/src/Sentry/Platforms/Cocoa/SentrySdk.cs#L154-L158

I guess we'd want to detect the release and distribution before we initialise the .NET SDK Hub... and possibly wire up the `CrashedLastRun` handler (although I'm not sure why we delegate to the Cocoa SDK for this - Matt implemented that in https://github.com/getsentry/sentry-dotnet/pull/1849). 

Ultimately the BeforeX handlers for whatever SDK we initialise 2nd may not be executed for the first few events that get sent... unless we add some concurrency mechanism to delay sending events until both SDKs are initialised. Maybe in the future we could consider that, if this ever becomes a problem.

As an aside, we're overwriting `CrashedLastRun`, `EnableScopeSync` and `ScopeObserver` here, so any user configured values for these will get overwritten. If that's genuinely what we want, those options should probably be internal (not public). I'll create a separate issue for that.

jurganson:
@jamescrosswell Do we know when this will be released? 🙂

src/Sentry/Platforms/Cocoa/SentrySdk.cs
vaind: these being public may be related to Unity

jamescrosswell:
@jurganson I'll start a 5.15.1 release now. 

Additionally, it looks like this whole workaround might be unnecessary - I think @jpnurmi might have created a more reliable fix for this in the Sentry SDK for Cocoa:
- https://github.com/getsentry/sentry-cocoa/pull/6193

That one will take a little bit longer as it would need to be reviewed, merged and released in the Cocoa SDK, then we need to bump the version of the Cocoa SDK that we're using in .NET. I'm guessing all of that will take at least a week or two.

Issue #4482, opened 2025-08-29
iOS <EnableAssemblyILStripping>false</EnableAssemblyILStripping> not working: symbols missing


giuseppenovielli:
### Package

Sentry.Maui

### .NET Flavor

.NET

### .NET Version

9.0.200

### OS

iOS

### OS Version

18.4

### Development Environment

Visual Studio Code (MacOS)

### SDK Version

5.14.1

### Workload Versions

`maui-ios                            9.0.0/9.0.100               SDK 9.0.200  `

### UseSentry or SentrySdk.Init call

```
.UseSentry(options =>
            {
                // The DSN is the only required setting.
                options.Dsn = Constants.SentryConstants.DNS;

                // Set TracesSampleRate to 1.0 to capture 100% of transactions for tracing.
                // We recommend adjusting this value in production.
                options.TracesSampleRate = 1.0;
            })
```

### Steps to Reproduce

1. Create new .net maui application and sentry maui project
2. into .csproj paste the follow property group e compile it from sentry project

```
<PropertyGroup Condition="'$(Configuration)' == 'Release'">
		<Optimize>true</Optimize>
		
		<PublishTrimmed>true</PublishTrimmed>
		<TrimMode>partial</TrimMode>
		
		<DebugType>portable</DebugType>
		<DebugSymbols>true</DebugSymbols>

		<!-- Configure Sentry org and project -->
               <SentryOrg>example-org</SentryOrg>
               <SentryProject>example-project</SentryProject>

		<!-- Automatically creates a release when building your application. -->
		<SentryCreateRelease>true</SentryCreateRelease>

		<!-- Automatically associates commits with the release. -->
		<SentrySetCommits>true</SentrySetCommits>

		<!-- Optionally provide explicit flags to the set-commits command -->
		<SentrySetCommitOptions>--local</SentrySetCommitOptions>

		<!-- Sends symbols to Sentry, enabling symbolication of stack traces. -->
		<SentryUploadSymbols>true</SentryUploadSymbols>

		<!-- Sends sources to Sentry, enabling display of source context. -->
		<SentryUploadSources>true</SentryUploadSources>

		<!-- https://github.com/getsentry/sentry-dotnet/issues/4427 -->
		<EnableAssemblyILStripping>false</EnableAssemblyILStripping>

                 <!-- If you are targeting Android, sends proguard mapping file to Sentry. -->
[...]

PR #4503, opened 2025-09-09
fix: upload linked PDB for Mac Catalyst

jpnurmi:
### Debug files

Notice the linked .PDB is different:

```
$ dotnet build -c Release -f net9.0-maccatalyst -r maccatalyst-arm64 -v:detailed EmpowerPlant.csproj
[...]
$ find . -name "EmpowerPlant.pdb" \
  -exec sha256 {} \; \
  -exec sh -c 'sentry-cli debug-files check "$1" | grep "Debug ID:"' _ {} \;
SHA256 (./obj/Release/net9.0-maccatalyst/maccatalyst-arm64/EmpowerPlant.pdb) = 808cd311d68980163937ac7cbf3ff95feeae81548fce41a22d4400acaa246c86
    > Debug ID: 321a8892-9045-515a-afed-fee16c70ae58-fb0e5afa
SHA256 (./obj/Release/net9.0-maccatalyst/maccatalyst-arm64/linked/EmpowerPlant.pdb) = fe9da3d5f0e213335e0abc10118b0dccbee5b2fa072e3bb596c3dc39351d04cc
    > Debug ID: 0c9408e1-5ffe-b8a0-1ce1-17268aaab993-14c33dd9 # <=== !!!
SHA256 (./bin/Release/net9.0-maccatalyst/maccatalyst-arm64/EmpowerPlant.pdb) = 808cd311d68980163937ac7cbf3ff95feeae81548fce41a22d4400acaa246c86
    > Debug ID: 321a8892-9045-515a-afed-fee16c70ae58-fb0e5afa
```

### Before

```
$ dotnet build -c Release -f net9.0-maccatalyst -v:detailed
...
    Preparing upload to Sentry for project 'EmpowerPlant' (Release/net9.0-maccatalyst): collecting debug symbols from bin/Release/net9.0-maccatalyst/maccatalyst-arm64/
    > Found 63 debug information files (1 with embedded sources)
```

### After

```
$ dotnet build -c Release -f net9.0-maccatalyst -v:detailed
...
    Preparing upload to Sentry for project 'EmpowerPlant' (Release/net9.0-maccatalyst): collecting debug symbols from bin/Release/net9.0-maccatalyst/maccatalyst-arm64/ obj/Release/net9.0-maccatalyst/maccatalyst-arm64/linked/EmpowerPlant.pdb
    > Found 64 debug information files (2 with embedded sources)
```

Fixes: #4196

PR #4527, opened 2025-09-15
fix: upload linked PDBs for iOS

jpnurmi:
Upload `linked/*.pdb` (instead of just `linked/$(AssemblyName).pdb` added for Mac Catalyst in #4503) to make missing symbols available on iOS with `EnableAssemblyILStripping=false` as reported in:
- #4482

image

Close: #4482

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Fixes
 
 - Upload linked PDBs to fix non-IL-stripped symbolication for iOS ([#4527](https://github.com/getsentry/sentry-dotnet/pull/4527))
+- In MAUI Android apps, generate and inject UUID to APK and upload ProGuard mapping to Sentry with the UUID ([#4532](https://github.com/getsentry/sentry-dotnet/pull/4532))
 
 ## 5.15.1
 
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -267,7 +267,7 @@ jobs:
           path: src
 
       - name: Integration test
-        uses: getsentry/github-workflows/sentry-cli/integration-test/@v2
+        uses: getsentry/github-workflows/sentry-cli/integration-test/@a5e409bd5bad4c295201cdcfe862b17c50b29ab7 # v2.14.1
         with:
           path: integration-test
 
@@ -348,7 +348,7 @@ jobs:
           path: src
 
       - name: Test AOT
-        uses: getsentry/github-workflows/sentry-cli/integration-test/@v2
+        uses: getsentry/github-workflows/sentry-cli/integration-test/@a5e409bd5bad4c295201cdcfe862b17c50b29ab7 # v2.14.1
         env:
           RuntimeIdentifier: ${{ matrix.rid }}
         with:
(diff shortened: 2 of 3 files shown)


jpnurmi:
I'm having trouble <sup>[*]</sup> verifying the fix but the integration test shows that it uploads the symbol file reported missing in #4482. Is that a sufficient confirmation, or how should we proceed with this?

[*] I have a brand new iPhone Air running iOS 26, and an old iPhone 6S stuck on iOS 15. Neither of them seems good for verifying this fix because the former is too new i.e. not yet fully supported, and the latter might just be too old as even `xcrun xctrace list devices` shows it in offline status. 🙁

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@
 - Fail when building Blazor WASM with Profiling. We don't support profiling in Blazor WebAssembly projects. ([#4512](https://github.com/getsentry/sentry-dotnet/pull/4512))
 - Do not overwrite user IP if it is set manually in ASP.NET sdk ([#4513](https://github.com/getsentry/sentry-dotnet/pull/4513))
 - Fix `SentryOptions.Native.SuppressSignalAborts` and `SuppressExcBadAccess` on iOS ([#4521](https://github.com/getsentry/sentry-dotnet/pull/4521))
+- Upload linked PDBs to fix non-IL-stripped symbolication for iOS ([#4527](https://github.com/getsentry/sentry-dotnet/pull/4527))
```
Flash0ver: note: update after `5.15.1` release

jpnurmi: Good catch, thanks! :pray:

Flash0ver:
> I'm having trouble [*] verifying the fix but the integration test shows that it uploads the symbol file reported missing in #4482. Is that a sufficient confirmation, or how should we proceed with this?

We could do a pre-release. On the other hand, if an Integration Test is covering and asserting the scenario, it LGTM.

Could this cause any unintended regression? If it might be "dangerous" without further testing, I am leaning towards a pre-release. If it cannot break existing scenarios, but may not actually fix all affected scenarios, I am leaning towards merging.

What do you prefer?

jpnurmi:
I'm leaning towards merging. All we do is pass an extra wildcard argument to `sentry-cli debug-files upload` to instruct it to upload more debug files. The CLI tool ignores duplicates and existing files, so I doubt the change is in any way dangerous. The only potential harm is that it might waste quota by uploading unnecessary .pdb files.

Although, as @jamescrosswell brought up in [#4482](https://github.com/getsentry/sentry-dotnet/issues/4482#issuecomment-3291470312), it's a bit unclear why these files are not available on the Microsoft symbol server that the user had enabled:

> OK so it's not available on any of the symbol servers that Sentry checked. 
> 
> @jpnurmi it looks like this one was in a "linked" directory:
> 
> Image
> 
> Any chance that fix you rolled out for MacCatalyst is related here?
> 
> It is odd that we can't pull this from the Microsoft Symbol servers though.

jpnurmi:
> The only potential harm is that it might waste quota by uploading unnecessary .pdb files.

Actually, https://sentry.zendesk.com/hc/en-us/articles/25269209638555-How-much-does-it-cost-to-upload-Debug-Files

> Sentry does not charge for debug information files (DIF) uploads. There is no limit on how many files you can upload and files are retained for 90 days without being used.

Issue #100, opened 2018-09-10
SDK Information


HazAT:
All SDKs should send SDK information like this:

for `Sentry` it should look like this:

```json
{
  "name": "sentry.dotnet",
  "version": "1.0.0-rc2",
  "packages": [
    {
      "name": "nuget:Sentry",
      "version": "1.0.0-rc2"
    },
  ]
}
```

for `Sentry.AspNetCore` it should look like this:

```json
{
  "name": "sentry.dotnet.aspnetcore",
  "version": "1.0.0-rc2",
  "packages": [
    {
      "name": "nuget:Sentry",
      "version": "1.0.0-rc2"
    },
    {
      "name": "nuget:Sentry.AspNetCore",
      "version": "1.0.0-rc2"
    },
  ]
}
```

and so on ...

HazAT:
@bruno-garcia almost, it's like this https://github.com/getsentry/sentry-docs/pull/293/commits/02e4ff713ca7b379645d4cfd63ba017b82ad501c

Since I think that `Sentry.AspNetCore` and `Sentry.Extensions.Logging` are both "SDKs" on their own, they deserve to bin in the namespace like:

`sentry.dotnet.aspnetcore`
and
`sentry.dotnet.extensions-logging`

does this make sense?

That way we can boil it down to the individual SDK and still filter for everything `dotnet` related.
The packages can stay the way they are.

HazAT:
The identifiers should basically match the folders here:
https://github.com/getsentry/sentry-release-registry/tree/master/sdks

bruno-garcia:
Closed by #105

PR #4532, opened 2025-09-16
fix: Generate and inject uuid to apk and upload proguard with that uuid

### LINKED ISSUE #3872: Adding support for generating or passing UUID when uploading progurad mapping on .NET MAUI
### Problem Statement

Uploading a progurad file by `<SentryUploadAndroidProguardMapping>` but this file is not associated with any release

According https://docs.sentry.io/cli/dif/#uploading-files this association requires adding `io.sentry.proguard-uuid` meetadata to and using `sentry-cli upload-proguard  --uuid ...`

### Solution Brainstorm

Preferred solution: Generate UUID, inject it to `AndroidManifest.xml` and use during uploading
or: read this UUID from `AndroidManifest.xml` and use during uploading
or: add support for an environment variable

BTW: Shouldn't other switches `--app-id my.app.id`, `--version`.  `--version-code` be used as well?

alexsohn1126:
fixes #3872 

# Problem

We were not automatically injecting uuid to the MAUI Android apk, so developers would have to manually go in and edit `AndroidManifest.xml`.

# Solution

As proposed by @rafalka, we can inject the uuid to the Apk, and upload the proguard file with the same uuid.

## Before

Before the fix, running  in the MAUI project would upload the proguard file with the uuid, but it won't inject the uuid to the Android app.

## After

We should now see the proguard file in Sentry with some uuid:

image

If we open the resulting APK in Android studio to inspect it's `AndroidManifest.xml`, we can see that it correctly set the `meta-data` field:

image

## Testing method

To test this change using our `samples/Sentry.Samples.Maui`, we need to modify `src/Sentry/buildTransitive/Sentry.targets` file. If not using our sample MAUI project, ignore this step.

We have to remove the check whether this is a Sentry project or not. So delete the `and [...]` in line 72-73:

https://github.com/getsentry/sentry-dotnet/blob/7061fb92d58ecd138228e3689a97c72fdab3b6e4/src/Sentry/buildTransitive/Sentry.targets#L69-L73

Next, ensure `SentryUploadAndroidProguardMapping` is set to true in your MAUI project's .csproj file:

```
    <SentryUploadAndroidProguardMapping>true</SentryUploadAndroidProguardMapping>
```

In the same file, enable Proguard. This is what I used:

```
  <PropertyGroup Condition="'$(Configuration)' == 'Release'">
    <AndroidLinkTool>r8</AndroidLinkTool>
    <AndroidLinkMode>Full</AndroidLinkMode>
    <AndroidEnableProguard>true</AndroidEnableProguard>
  </PropertyGroup>
```

Run `dotnet publish -f net9.0-android35.0 -c Release` in the MAUI project root.

### DIFF
--- a/src/Sentry/buildTransitive/Sentry.targets
+++ b/src/Sentry/buildTransitive/Sentry.targets
@@ -18,6 +18,7 @@
 
   <PropertyGroup>
     <SentryAttributesFile>Sentry.Attributes$(MSBuildProjectExtension.Replace('proj', ''))</SentryAttributesFile>
+    <ProguardUuid>$([System.Guid]::NewGuid())</ProguardUuid>
   </PropertyGroup>
 
   <Target Name="_SentryEnsureAndroidEnableAssemblyCompressionDisabled"
@@ -125,11 +126,17 @@
       <SentrySetCommitReleaseOptions Condition="'$(SentryOrg)' != ''">$(SentrySetCommitReleaseOptions) --org $(SentryOrg)</SentrySetCommitReleaseOptions>
       <SentrySetCommitReleaseOptions Condition="'$(SentryProject)' != ''">$(SentrySetCommitReleaseOptions) --project $(SentryProject)</SentrySetCommitReleaseOptions>
 
+      <SentryProguardOptions Condition="'$(ProguardUuid)' != ''">$(SentryProguardOptions) --uuid &quot;$(ProguardUuid)&quot;</SentryProguardOptions>
+      <SentryProguardOptions Condition="'$(ApplicationId)' != ''">$(SentryProguardOptions) --app-id &quot;$(ApplicationId)&quot;</SentryProguardOptions>
+      <SentryProguardOptions Condition="'$(ApplicationDisplayVersion)' != ''">$(SentryProguardOptions) --version &quot;$(ApplicationDisplayVersion)&quot;</SentryProguardOptions>
+      <SentryProguardOptions Condition="'$(ApplicationVersion)' != ''">$(SentryProguardOptions) --version-code &quot;$(ApplicationVersion)&quot;</SentryProguardOptions>
+      <SentryProguardOptions Condition="'$(SentryProguardOptions.Trim())' != ''">$(SentryProguardOptions.trim())</SentryProguardOptions>
+
       <SentryCLIUploadOptions Condition="'$(SentryOrg)' != ''">$(SentryCLIUploadOptions) --org $(SentryOrg)</SentryCLIUploadOptions>
       <SentryCLIUploadOptions Condition="'$(SentryProject)' != ''">$(SentryCLIUploadOptions) --project $(SentryProject)</SentryCLIUploadOptions>
       <SentryCLIDebugFilesUploadCommand>$(SentryCLIBaseCommand) debug-files upload</SentryCLIDebugFilesUploadCommand>
       <SentryCLIDebugFilesUploadCommand Condition="'$(SentryCLIUploadOptions.Trim())' != ''">$(SentryCLIDebugFilesUploadCommand) $(SentryCLIUploadOptions.Trim())</SentryCLIDebugFilesUploadCommand>
-      <SentryCLIProGuardMappingUploadCommand>$(SentryCLIBaseCommand) upload-proguard</SentryCLIProGuardMappingUploadCommand>
+      <SentryCLIProGuardMappingUploadCommand>$(SentryCLIBaseCommand) upload-proguard $(SentryProguardOptions)</SentryCLIProGuardMappingUploadCommand>
       <SentryCLIProGuardMappingUploadCommand Condition="'$(SentryCLIUploadOptions.Trim())' != ''">$(SentryCLIProGuardMappingUploadCommand) $(SentryCLIUploadOptions.Trim())</SentryCLIProGuardMappingUploadCommand>
     </PropertyGroup>
 
@@ -265,6 +272,16 @@
       </Exec>
     <!-- } -->
     <Warning Condition="'$(_SentryCLIExitCode)' != '0'" Text="Sentry CLI could not upload debug files." />
+  </Target>
+
+  <Target Name="UpdateAndroidMetadata" BeforeTargets="GetAssemblyAttributes">
+
+    <ItemGroup>
+      <AssemblyAttribute Include="Android.App.MetaData">
+        <_Parameter1>io.sentry.proguard-uuid</_Parameter1>
+        <Value>$(ProguardUuid)</Value>
+      </AssemblyAttribute>
+    </ItemGroup>
   </Target>
 
   <!-- Upload Android Proguard mapping file to Sentry after the build. -->
--- a/src/Sentry/buildTransitive/Sentry.targets
+++ b/src/Sentry/buildTransitive/Sentry.targets
@@ -70,8 +70,7 @@
       <UseSentryCLI Condition="
         '$(UseSentryCLI)' == ''
         and ('$(SentryUploadSymbols)' == 'true' or '$(SentryUploadSources)' == 'true' or $(SentryUploadAndroidProguardMapping) == 'true' or $(SentryCreateRelease) == 'true' or $(SentrySetCommits) == 'true')
-        and '$(MSBuildProjectName)' != 'Sentry'
-        and !$(MSBuildProjectName.StartsWith('Sentry.'))">true</UseSentryCLI>
+        ">true</UseSentryCLI>
     </PropertyGroup>
   </Target>
 
@@ -274,7 +273,7 @@
     <Warning Condition="'$(_SentryCLIExitCode)' != '0'" Text="Sentry CLI could not upload debug files." />
   </Target>
 
-  <Target Name="UpdateAndroidMetadata" BeforeTargets="GetAssemblyAttributes">
+  <Target Name="UpdateAndroidMetadata" BeforeTargets="GetAssemblyAttributes" Condition="'$(SentryUploadAndroidProguardMapping)' == 'true'">
 
     <ItemGroup>
       <AssemblyAttribute Include="Android.App.MetaData">


alexsohn1126:
Looks like there is an endpoint for proguard artifact upload that is not implemented in sentry-cli integration tests server.

In other integration test servers, they have endpoints for `proguard-artifact-releases`:
- https://github.com/getsentry/sentry-maven-plugin/blob/c2403b7113af5351d732308b24b2016dba0a4714/test/integration-test-server.py#L45
- https://github.com/getsentry/sentry-android-gradle-plugin/blob/5e15a33821ffe42aaa86a2a9311738f01a943820/test/integration-test-server.py#L45

I'll make a PR over at `sentry-cli` tomorrow for this, or should I ask someone else?

Doesn't seem too hard to make it, what can possibly go wrong 🫨

alexsohn1126:
Actually, it's kind of weird that we are getting errors from uploading proguards now but we weren't before?

Maybe something's up with my implementation, but from what I can see, it doesn't touch the condition which the proguard file is sent so...

alexsohn1126:
Blocked by [getsentry/github-workflow #100](https://github.com/getsentry/github-workflows/pull/100)

src/Sentry/buildTransitive/Sentry.targets
```diff
@@ -125,11 +126,17 @@
       <SentrySetCommitReleaseOptions Condition="'$(SentryOrg)' != ''">$(SentrySetCommitReleaseOptions) --org $(SentryOrg)</SentrySetCommitReleaseOptions>
       <SentrySetCommitReleaseOptions Condition="'$(SentryProject)' != ''">$(SentrySetCommitReleaseOptions) --project $(SentryProject)</SentrySetCommitReleaseOptions>
 
+      <SentryProguardOptions Condition="'$(ProguardUuid)' != ''">$(SentryProguardOptions) --uuid &quot;$(ProguardUuid)&quot;</SentryProguardOptions>
+      <SentryProguardOptions Condition="'$(ApplicationId)' != ''">$(SentryProguardOptions) --app-id &quot;$(ApplicationId)&quot;</SentryProguardOptions>
+      <SentryProguardOptions Condition="'$(ApplicationDisplayVersion)' != ''">$(SentryProguardOptions) --version &quot;$(ApplicationDisplayVersion)&quot;</SentryProguardOptions>
+      <SentryProguardOptions Condition="'$(ApplicationVersion)' != ''">$(SentryProguardOptions) --version-code &quot;$(ApplicationVersion)&quot;</SentryProguardOptions>
+      <SentryProguardOptions Condition="'$(SentryProguardOptions.Trim())' != ''">$(SentryProguardOptions.trim())</SentryProguardOptions>
```
bruno-garcia: I'm surprised "trim" works, in C# it has to be Trim() since it's pascal case. some MSBuild thing?

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 
 - Fail when building Blazor WASM with Profiling. We don't support profiling in Blazor WebAssembly projects. ([#4512](https://github.com/getsentry/sentry-dotnet/pull/4512))
 - Do not overwrite user IP if it is set manually in ASP.NET sdk ([#4513](https://github.com/getsentry/sentry-dotnet/pull/4513))
+- In MAUI Android apps, generate and inject uuid to Apk and upload Proguard to Sentry with the uuid ([#4532](https://github.com/getsentry/sentry-dotnet/pull/4532))
```
bruno-garcia: the right casing is: ProGuard (not Proguard)

Please if it everywhere including mabuild props etc

src/Sentry/buildTransitive/Sentry.targets
```diff
@@ -125,11 +126,17 @@
       <SentrySetCommitReleaseOptions Condition="'$(SentryOrg)' != ''">$(SentrySetCommitReleaseOptions) --org $(SentryOrg)</SentrySetCommitReleaseOptions>
       <SentrySetCommitReleaseOptions Condition="'$(SentryProject)' != ''">$(SentrySetCommitReleaseOptions) --project $(SentryProject)</SentrySetCommitReleaseOptions>
 
+      <SentryProguardOptions Condition="'$(ProguardUuid)' != ''">$(SentryProguardOptions) --uuid &quot;$(ProguardUuid)&quot;</SentryProguardOptions>
```
bruno-garcia: do we need a main guard that checks to ensure all of the required props are included before we do this?

I wonder if only some of them are, we'll run the command with broken set of arguments and it'll fail. 

Should we let that happen? are some of these arguments optional?

I believe version is no longer neeeed btw, @szokeasaurusrex will know since he was proposing changes to this stuff.

src/Sentry/buildTransitive/Sentry.targets
```diff
@@ -267,6 +274,20 @@
...
 
+  <Target Name="UpdateAndroidMetadata" BeforeTargets="GetAssemblyAttributes"
+          Condition="
+            '$(TargetFramework)' != ''
+            and $(TargetFramework.Contains('-android'))
+            and '$(SentryUploadAndroidProguardMapping)' == 'true'">
+
+    <ItemGroup>
+      <AssemblyAttribute Include="Android.App.MetaData">
+        <_Parameter1>io.sentry.proguard-uuid</_Parameter1>
```
bruno-garcia, line 281: Does paramter1 mean we override an existing thing if it's there or? does it play well with the app developer or another library using the same strategy to add metadata?

src/Sentry/buildTransitive/Sentry.targets
alexsohn1126: Lol, weird. It is def a typo tho, will fix!

src/Sentry/buildTransitive/Sentry.targets
alexsohn1126: If an app developer tries to add another metadata to the APK using a similar strategy with a different name, it will get added without any trouble.

If an app developer uses the same name - `io.sentry.proguard-uuid` - (which they shouldn't) for the metadata, whichever gets processed later will be the one that gets put in the APK. 

Maybe we should mention not to overwrite the `io.sentry.proguard-uuid` metadata in the docs? I feel like you can infer it from the [docs](https://docs.sentry.io/cli/dif/#proguard-mapping-upload), but doesn't hurt to be more explicit.

src/Sentry/buildTransitive/Sentry.targets
alexsohn1126: Good point. I researched a bit more into the implementation of `sentry-cli`, and found that:
- `uuid` is not required, but I would assume this is available 100% of the time as long as `[System.Guid]::NewGuid()` works.. ([source](https://github.com/getsentry/sentry-cli/blob/master/src/commands/upload_proguard.rs?utm_source=chatgpt.com#L114-L128))
- `app-id` and `version` are dependent on each other. Meaning we either have none of them, or both. ([`version` source](https://github.com/getsentry/sentry-cli/blob/master/src/commands/upload_proguard.rs?utm_source=chatgpt.com#L37-L46), [`app-id` source](https://github.com/getsentry/sentry-cli/blob/master/src/commands/upload_proguard.rs?utm_source=chatgpt.com#L60-L69))
- `version-code` can only be set if both `app-id` and `version` is present. ([source](https://github.com/getsentry/sentry-cli/blob/master/src/commands/upload_proguard.rs?utm_source=chatgpt.com#L47-L58))

If version doesn't become needed in the future, we can remove it. It would also mean removing both `app-id` and `version-code` though.

Edit: I will remove `version`, `app-id`, and `version-code`.

CHANGELOG.md
```suggestion
- In MAUI Android apps, generate and inject UUID to APK and upload Proguard to Sentry with the UUID ([#4532](https://github.com/getsentry/sentry-dotnet/pull/4532))
```

I think UUID and APK are both usually spelled uppercase.

src/Sentry/buildTransitive/Sentry.targets
```diff
@@ -18,6 +18,7 @@
 
   <PropertyGroup>
     <SentryAttributesFile>Sentry.Attributes$(MSBuildProjectExtension.Replace('proj', ''))</SentryAttributesFile>
+    <ProguardUuid>$([System.Guid]::NewGuid())</ProguardUuid>
```
Flash0ver: naming:`Sentry`-prefix

I wonder if we should prefix `ProguardUuid` with `SentryProguardUuid` (consistently capitalized: `SentryProGuardUuid`) as well, to not accidentally collide with other properties.

Flash0ver: question: overwrite value

Is it intended to unconditionally overwrite potentially user-defined values for `ProguardUuid`?
Or do we want to keep the option for user-defined UUIDs, and we only set a new GUID if the property is not set / is empty?

If we want to allow user-defined values, and only set it ourselves if it is unset/empty, I guess we also want to mention the property in our docs: https://docs.sentry.io/platforms/dotnet/configuration/msbuild/

src/Sentry/buildTransitive/Sentry.targets
Flash0ver: question: naming

I'm not super certain, but would a `CLI` infix be more consistent?
```diff
-SentryProguardOptions
+SentryCLIProGuardOptions
```

FWICS, we omit the `CLI` infix when it's an option that we'd like users to consider and document in https://docs.sentry.io/platforms/dotnet/configuration/msbuild/.
And we include the `CLI` infix when it's a property that we use mainly for aggregation/combination of other (potentially user-defined / user-overwritten) properties.
(and prefix with `_` when it is an "internal" property that user-code should not use)

What do you think?

src/Sentry/buildTransitive/Sentry.targets
alexsohn1126: You're probably right, it'd be better to put `Sentry` prefix to avoid overlap!

alexsohn1126: My thought process was that we should generate a unique uuid is used for each build so there wouldn't be a case where a developer forgets to update their uuid if they have set it manually previously. 

It wouldn't be hard at all to allow users to put their own uuid (not overriding it), and I could see it being useful when someone wants to use a uuid from a whole another external source by providing the property at the cli call by doing something like: `dotnet build -p:SentryProGuardUuid=UUID`.

I'll add that and edit the docs once this change is merged!

src/Sentry/buildTransitive/Sentry.targets
alexsohn1126: I will add `SentryProGuardUuid` as an option that users would change, and users shouldn't really change `SentryProguardOptions` directly. So we can change `SentryProguardOptions` to `_SentryCLIProguardOptions`!

CHANGELOG.md
alexsohn1126: Seems like .NET uses [`Proguard` for their properties](https://learn.microsoft.com/en-us/dotnet/android/building-apps/build-properties#androidproguardmappingfile) so we can't change that, but will change everywhere else.

src/Sentry/buildTransitive/Sentry.targets
jamescrosswell: @bruno-garcia I think Alex has used [positional constructor arguments](https://learn.microsoft.com/en-us/visualstudio/msbuild/writecodefragment-task?view=vs-2022#parameters) here... so they shouldn't conflict with other properties in the MS Build file...

I guess folks shouldn't use those names for more "permanent" build properties 😉

jamescrosswell (approved):
Looks good - thanks @alexsohn1126 !

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 
 - Fail when building Blazor WASM with Profiling. We don't support profiling in Blazor WebAssembly projects. ([#4512](https://github.com/getsentry/sentry-dotnet/pull/4512))
 - Do not overwrite user IP if it is set manually in ASP.NET sdk ([#4513](https://github.com/getsentry/sentry-dotnet/pull/4513))
+- In MAUI Android apps, generate and inject UUID to APK and upload ProGuard mapping to Sentry with the UUID ([#4532](https://github.com/getsentry/sentry-dotnet/pull/4532))
```
Flash0ver: note: update (`5.15.1` has been released)

.github/workflows/build.yml
```diff
@@ -267,7 +267,7 @@ jobs:
           path: src
 
       - name: Integration test
-        uses: getsentry/github-workflows/sentry-cli/integration-test/@v2
+        uses: getsentry/github-workflows/sentry-cli/integration-test/@0ee6d58acd5ca949358c3f62263a3c592689553e
```
Flash0ver: suggestion: update to commit that is also tagged (latest release)
see: https://github.com/getsentry/github-workflows/releases/tag/2.14.1

```suggestion
        uses: getsentry/github-workflows/sentry-cli/integration-test/@a5e409bd5bad4c295201cdcfe862b17c50b29ab7 # v2.14.1
```

We started with an untagged version for a quick fix.

Optional.
- If you apply, please also apply the comment with the tag name. I believe our GitHub Actions updater is keeping the comment in sync, too ... see https://github.com/getsentry/sentry-dotnet/pull/4524
- If you apply, please also apply to the `- name: Test AOT` step below.

src/Sentry/buildTransitive/Sentry.targets
```diff
@@ -68,7 +69,7 @@
            We're explicitly skipping uploads for Sentry projects because they interfere with CLI integration test asserts. -->
       <UseSentryCLI Condition="
         '$(UseSentryCLI)' == ''
-        and ('$(SentryUploadSymbols)' == 'true' or '$(SentryUploadSources)' == 'true' or $(SentryUploadAndroidProguardMapping) == 'true' or $(SentryCreateRelease) == 'true' or $(SentrySetCommits) == 'true')
+        and ('$(SentryUploadSymbols)' == 'true' or '$(SentryUploadSources)' == 'true' or $(SentryUploadAndroidProGuardMapping) == 'true' or $(SentryCreateRelease) == 'true' or $(SentrySetCommits) == 'true')
```
Flash0ver, line 72: note(docs):

Afterwards, please also update the documentation with the new capitalization.
It's not a breaking change as MSBuild isn't case-sensitive.

https://docs.sentry.io/platforms/dotnet/configuration/msbuild/

.github/workflows/build.yml
alexsohn1126: Done!

CHANGELOG.md
alexsohn1126: Updated :)

Flash0ver (approved):
Left some more minor, non-blocking notes.

src/Sentry/buildTransitive/Sentry.targets
alexsohn1126: I'll make it a part of this pr:

https://github.com/getsentry/sentry-docs/pull/14995

PR #4533, opened 2025-09-17
Remove Headers and Modules from SentryCocoaFramework before packing

jamescrosswell:
Workaround for #4292:
- https://github.com/getsentry/sentry-dotnet/issues/4292#issuecomment-3083566046

> [!WARNING]
> We should publish a pre-release from this branch and validate this resolves the problem before merging into main... this is quite a major change.

### DIFF
--- a/src/Sentry.Bindings.Cocoa/ApiDefinitions.cs
+++ b/src/Sentry.Bindings.Cocoa/ApiDefinitions.cs
@@ -416,9 +416,9 @@ interface SentryEnvelopeItemHeader : SentrySerializable
     [Export ("initWithType:length:filenname:contentType:")]
     NativeHandle Constructor (string type, nuint length, string filename, string contentType);
 
-    // -(instancetype _Nonnull)initWithType:(NSString * _Nonnull)type length:(NSUInteger)length contentType:(NSString * _Nonnull)contentType itemCount:(NSNumber * _Nonnull)itemCount;
+    // -(instancetype _Nonnull)initWithType:(NSString * _Nonnull)type length:(NSUInteger)length contentType:(NSString * _Nullable)contentType itemCount:(NSNumber * _Nonnull)itemCount;
     [Export ("initWithType:length:contentType:itemCount:")]
-    NativeHandle Constructor (string type, nuint length, string contentType, NSNumber itemCount);
+    NativeHandle Constructor (string type, nuint length, [NullAllowed] string contentType, NSNumber itemCount);
 
     // @property (readonly, copy, nonatomic) NSString * _Nonnull type;
     [Export ("type")]
--- a/src/Sentry.Bindings.Cocoa/ApiDefinitions.cs
+++ b/src/Sentry.Bindings.Cocoa/ApiDefinitions.cs
@@ -225,7 +225,7 @@ interface SentryBreadcrumb : SentrySerializable
 
     // -(NSDictionary<NSString *,id> * _Nonnull)serialize;
     [Export ("serialize")]
-    new NSDictionary<NSString, NSObject> Serialize();
+    NSDictionary<NSString, NSObject> Serialize();
 
     // -(BOOL)isEqualToBreadcrumb:(SentryBreadcrumb * _Nonnull)breadcrumb;
     [Export ("isEqualToBreadcrumb:")]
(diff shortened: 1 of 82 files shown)


src/Sentry.Bindings.Cocoa/Sentry.Bindings.Cocoa.csproj
```diff
@@ -113,6 +113,15 @@
     <Exec Condition="'$(GITHUB_ACTIONS)' == 'true'" Command="pwsh ../../scripts/dirty-check.ps1 -PathToCheck $(MSBuildThisFileDirectory)" />
   </Target>
 
+  <!-- Sanitize the xcframework before NativeReference resolution and packing -->
+  <Target Name="SanitizeSentryCocoaFramework"
+          AfterTargets="_GenerateSentryCocoaBindings"
+          BeforeTargets="ResolveNativeReferences"
+          Condition="$([MSBuild]::IsOSPlatform('OSX')) and Exists('$(SentryCocoaFramework)')">
+    <Message Importance="High" Text="Sanitizing $(SentryCocoaFramework): removing dSYMs, Headers, Modules, PrivateHeaders (including symlinks)." />
+    <Exec Command="bash -c 'find &quot;$(SentryCocoaFramework)&quot; -depth \( -name dSYMs -o -name Headers -o -name Modules -o -name PrivateHeaders \) -exec echo Removing {} \; -exec rm -rf {} + 2&gt;/dev/null || true'" />
```
jpnurmi: I think this executes for multiple target frameworks in parallel. Might be the reason you had to ignore errors because the two commands race and conflict?

What if this was already done at the end of `_SetupCocoaSDK`, which is guarded to run only once? If we need to ignore errors regardless, we could get away without bash with `<Exec IgnoreExitCode="true" ... />`

jpnurmi: P.S. `rm` has verbose mode (`-v`) with quite similar output

jamescrosswell: I don't think the Android Device test failures had anything to do with these changes to the Cocoa Bindings project... they appear to be flaky on other unrelated branches as well:
- https://github.com/getsentry/sentry-dotnet/issues/4536

Given that, I think we can create a pre-release from this branch that some of our SDK users can test before we consider merging it into main.

jpnurmi: The same `find $(SentryCocoaFramework) ... -exec rm {}` command runs in parallel for `net8.0-ios17.0` and `net8.0-maccatalyst17.0` even if both operate on the same `modules/sentry-cocoa/Sentry-8.56.0.xcframework` directory.

Instead of ignoring errors caused by the race between those two commands running in the same directory, could we move the Exec-task to `_SetupCocoaSDK`, which is executed exactly once? That way, we could simplify the command to:
```suggestion
    <Exec Command="find &quot;$(SentryCocoaFramework)&quot; -depth \( -name dSYMs -o -name Headers -o -name Modules -o -name PrivateHeaders \) -exec rm -rfv {} +" />
```

jamescrosswell: Moved `_GenerateSentryCocoaBindings` and `SanitizeSentryCocoaFramework` up into the `_SetupSentrySdk` target (which executes only once).

src/Sentry.Bindings.Cocoa/ApiDefinitions.cs
```diff
@@ -416,9 +416,9 @@ interface SentryEnvelopeItemHeader : SentrySerializable
     [Export ("initWithType:length:filenname:contentType:")]
     NativeHandle Constructor (string type, nuint length, string filename, string contentType);
 
-    // -(instancetype _Nonnull)initWithType:(NSString * _Nonnull)type length:(NSUInteger)length contentType:(NSString * _Nonnull)contentType itemCount:(NSNumber * _Nonnull)itemCount;
+    // -(instancetype _Nonnull)initWithType:(NSString * _Nonnull)type length:(NSUInteger)length contentType:(NSString * _Nullable)contentType itemCount:(NSNumber * _Nonnull)itemCount;
     [Export ("initWithType:length:contentType:itemCount:")]
-    NativeHandle Constructor (string type, nuint length, string contentType, NSNumber itemCount);
+    NativeHandle Constructor (string type, nuint length, [NullAllowed] string contentType, NSNumber itemCount);
```
jamescrosswell, line 421: Why has this changed?

jpnurmi: I suppose we are missing dirty-check for `ApiDefinitions.cs`. It was changed to nullable in 8.56.1:
- getsentry/sentry-cocoa#6186

jamescrosswell:
Replaced by #4602
- #4602

PR #4544, opened 2025-09-19
Ensure template is not sent for Structured logs with no parameters

### LINKED ISSUE #4480: Make sure the .NET SDK only sets `sentry.message.template` on logs if `sentry.message.parameter` values are also defined
ref [https://github.com/getsentry/sentry-docs/pull/14759](https://github.com/getsentry/sentry-docs/pull/14759)

> If there are no `sentry.message.parameter.X` attributes included in the log, then the SDK MUST NOT attach a `sentry.message.template` attribute. This is important because it reduces duplicate content in the log and makes PII processing in Sentry much easier as only we don’t have to duplicate processing between the log message and the `sentry.message.template` attribute.

jamescrosswell:
Resolves #4480:
- https://github.com/getsentry/sentry-dotnet/issues/4480

---

> [!NOTE]
> Skip emitting `sentry.message.template` in structured logs without parameters; adds coverage and updates changelog.
> 
> - **Structured Logs**:
>   - Update `src/Sentry/SentryLog.cs` to only write `attributes.sentry.message.template` when `Template` is set and `Parameters` has length > 0.
> - **Tests**:
>   - Add `WriteTo_NoParameters_NoTemplate` theory in `test/Sentry.Tests/SentryLogTests.cs` to assert presence/absence of `sentry.message.template` based on parameters.
> - **Docs/Changelog**:
>   - Update `CHANGELOG.md` under Fixes to note templates aren’t sent for logs without parameters.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit faad68af1a8cf9d27b305be334b43e7237a29667. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/Sentry/SentryLog.cs
+++ b/src/Sentry/SentryLog.cs
@@ -226,7 +226,7 @@ public sealed class SentryLog
 
         // the SDK MUST NOT attach a sentry.message.template attribute if there are no parameters
         // https://develop.sentry.dev/sdk/telemetry/logs/#default-attributes
-        if (Template is not null && Parameters is { Length: > 0 })
+        if (Template is not null && !Parameters.IsDefaultOrEmpty)
         {
             SentryAttributeSerializer.WriteStringAttribute(writer, "sentry.message.template", Template);
         }


src/Sentry/Internal/DefaultSentryStructuredLogger.cs
```diff
@@ -53,7 +53,9 @@ private protected override void CaptureLog(SentryLogLevel level, string template
 
         SentryLog log = new(timestamp, traceHeader.TraceId, level, message)
         {
-            Template = template,
+            // the SDK MUST NOT attach a sentry.message.template attribute if there are no parameters
+            // https://develop.sentry.dev/sdk/telemetry/logs/#default-attributes
+            Template = parameters is { Length: > 0 } ? template : null,
```
Flash0ver: suggestion: move to Serialization or _.ctor_

Since this behavior should also be applied to integrations like `Microsoft.Extensions.Logging`, _ASP.NET Core_, MAUI, and _Serilog_,
perhaps moving it to `SentryLog.WriteTo(Utf8JsonWriter, IDiagnosticLogger)` or the _.ctor_ could consolidate it into a single location.

jamescrosswell: Good point 👍🏻

src/Sentry/SentryLog.cs
```diff
@@ -224,7 +224,9 @@ internal void WriteTo(Utf8JsonWriter writer, IDiagnosticLogger? logger)
         writer.WritePropertyName("attributes");
         writer.WriteStartObject();
 
-        if (Template is not null)
+        // the SDK MUST NOT attach a sentry.message.template attribute if there are no parameters
+        // https://develop.sentry.dev/sdk/telemetry/logs/#default-attributes
+        if (Template is not null && Parameters is { Length: > 0 })
```
Flash0ver: potential issue: Accessing `Length` could throw `NullReferenceException` when `default`

```suggestion
        if (Template is not null && !Parameters.IsDefaultOrEmpty)
```

---

Admittedly, this is a bit pedantic. We currently have no flow setting the `Parameters` to `default`, but we technically could make a (semi-)related change in the future and then run into an _NRE_ along the line.

```CSharp
ImmutableArray<KeyValuePair<string, object>> parameters = default;
_ = parameters.Length; // System.NullReferenceException: Object reference not set to an instance of an object.
```

The `ImmutableArray<T>.Length` property is accessing the underlying array directly, which is `null` when the `struct ImmutableArray<T>` is (explicitly/implicitly) `default`.
`ImmutableArray<T>.IsEmpty` has a similar behavior.

To be on the safe side, I suggest using `ImmutableArray<T>.IsDefaultOrEmpty` which ensures that this `struct` instance isn't `default` by checking whether the underlying array is `null` before accessing its `Length`.

**Context:**
I decided to use `ImmutableArray<T>` here, because we expose the `SentryLog.Parameters` publicly and to render this collection immutable through e.g. the user-defined `SetBeforeSendLog(Func<SentryLog, SentryLog?>)` delegate. It does have its usage quirks for the benefit of similar performance characteristics to using a mutable Array directly.

test/Sentry.Tests/SentryLogTests.cs
```diff
@@ -67,6 +67,31 @@ public void Protocol_Default_VerifyAttributes()
...
+            : [];
+        var log = new SentryLog(Timestamp, TraceId, SentryLogLevel.Debug, "message")
+        {
+            Template = "template",
+            Parameters = parameters,
+            ParentSpanId = ParentSpanId,
+        };
+
+        // Act
+        // var document = log.ToJsonString();
```
Flash0ver:
```suggestion
```

leftover?

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 ### Fixes
 
+- Templates are no longer sent with Structured that have no parameters ([#4544](https://github.com/getsentry/sentry-dotnet/pull/4544))
```
```suggestion
- Templates are no longer sent with Structured Logs that have no parameters ([#4544](https://github.com/getsentry/sentry-dotnet/pull/4544))
```

src/Sentry/SentryLog.cs
jamescrosswell: I think this us guarded against. If the condition was `Parameters is { }` then it would ensure Parameters isn't null... and so `Parameters is { Length: > 0 }` ensures that it's not null and has more than one element right? 

EDIT: Oh, I see... `default != null`

Flash0ver: Yeah ... ``ImmutableArray`1`` is a _value type_. But the underlying array is a _reference type_. And the `Length` property does not perform a `null` check on the underlying array, which actually is null should the ``ImmutableArray`1`` be initialized to `default`.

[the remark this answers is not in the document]
Flash0ver: Partially correct.
The `get_Length` property does indeed throw when `Parameters` is a default `ImmutableArray<T>`.
But it does not throw an `InvalidOperationException`, but a `NullReferenceException`!

See
https://github.com/dotnet/runtime/blob/e32756055c1427f7bbbc59c2f34f0190d30e6264/src/libraries/System.Collections.Immutable/src/System/Collections/Immutable/ImmutableArray_1.Minimal.cs#L164-L176

jamescrosswell: Not an ideal design but I guess it's there as a permanent foot gun now, so we just have to keep it in the back of our minds. Maybe eventually we'll get an analyser for this.

PR #4565, opened 2025-09-25
fix(logs): Structured Logs do not send ParentSpanId when no Span was active

Flash0ver:
This is a follow-up PR
from https://github.com/getsentry/sentry-dotnet/pull/4462#discussion_r2296973291
that is applying a change we made for the `Serilog` integration of  _Structured Logs_
to the SDK-Logger and the `Microsoft.Extensions.Logging` integration,
when reading values for the attributes `"trace_id"` and `"sentry.trace.parent_span_id"`.

Rather than using the `TraceHeader`,
instead use the currently active Span, or the Propagation Context as a fallback.

~During testing I noticed, that there doesn't seem to be a behavioral difference in Sentry, hence the `#skip-changelog`.~

---

> [!NOTE]
> Structured Logs now derive `trace_id` from the active span or scope and only include `sentry.trace.parent_span_id` when an active span exists.
> 
> - **Structured Logs behavior**:
>   - Read `trace_id`/`parent_span_id` via `SentryLog.GetTraceIdAndSpanId(...)` from the active `Span`; fall back to scope for `trace_id` only.
>   - Omit `sentry.trace.parent_span_id` when no active span.
> - **Implementations updated**:
>   - `Sentry.Extensions.Logging` logger and `DefaultSentryStructuredLogger` now use the new helper (replacing `GetTraceHeader`).
>   - `Sentry.Serilog` sink uses the helper; removes its local resolver.
>   - Adds helper to `SentryLog` core type.
> - **Tests**: Adjusted to validate new source of IDs and absence of `parent_span_id` without an active span.
> - **Changelog**: Note about not sending Parent-Span-IDs without an active span.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 2326d97741195fa1b3d660f3708ab36bf465d6d4. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/Sentry.Extensions.Logging/SentryStructuredLogger.cs
+++ b/src/Sentry.Extensions.Logging/SentryStructuredLogger.cs
@@ -42,7 +42,7 @@ internal sealed class SentryStructuredLogger : ILogger
         }
 
         var timestamp = _clock.GetUtcNow();
-        var traceHeader = _hub.GetTraceHeader() ?? SentryTraceHeader.Empty;
+        SentryLog.GetTraceIdAndSpanId(_hub, out var traceId, out var spanId);
 
         var level = logLevel.ToSentryLogLevel();
         Debug.Assert(level != default);
@@ -81,11 +81,11 @@ internal sealed class SentryStructuredLogger : ILogger
             }
         }
 
-        SentryLog log = new(timestamp, traceHeader.TraceId, level, message)
+        SentryLog log = new(timestamp, traceId, level, message)
         {
             Template = template,
             Parameters = parameters.DrainToImmutable(),
-            ParentSpanId = traceHeader.SpanId,
+            ParentSpanId = spanId,
         };
 
         log.SetDefaultAttributes(_options, _sdk);
--- a/src/Sentry.Serilog/SentrySink.Structured.cs
+++ b/src/Sentry.Serilog/SentrySink.Structured.cs
@@ -7,7 +7,7 @@ internal sealed partial class SentrySink
 {
     private static void CaptureStructuredLog(IHub hub, SentryOptions options, LogEvent logEvent, string formatted, string? template)
     {
-        GetTraceIdAndSpanId(hub, out var traceId, out var spanId);
+        SentryLog.GetTraceIdAndSpanId(hub, out var traceId, out var spanId);
         GetStructuredLoggingParametersAndAttributes(logEvent, out var parameters, out var attributes);
 
         SentryLog log = new(logEvent.Timestamp, traceId, logEvent.Level.ToSentryLogLevel(), formatted)
@@ -25,28 +25,6 @@ internal sealed partial class SentrySink
         }
 
         hub.Logger.CaptureLog(log);
-    }
-
-    private static void GetTraceIdAndSpanId(IHub hub, out SentryId traceId, out SpanId? spanId)
-    {
-        var span = hub.GetSpan();
-        if (span is not null)
-        {
-            traceId = span.TraceId;
-            spanId = span.SpanId;
-            return;
-        }
-
-        var scope = hub.GetScope();
-        if (scope is not null)
-        {
-            traceId = scope.PropagationContext.TraceId;
-            spanId = scope.PropagationContext.SpanId;
-            return;
-        }
-
-        traceId = SentryId.Empty;
-        spanId = null;
     }
 
     private static void GetStructuredLoggingParametersAndAttributes(LogEvent logEvent, out ImmutableArray<KeyValuePair<string, object>> parameters, out List<KeyValuePair<string, object>> attributes)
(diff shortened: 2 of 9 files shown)


Flash0ver:
@sentry review

src/Sentry.Extensions.Logging/SentryStructuredLogger.cs
```diff
@@ -42,7 +42,7 @@ public void Log<TState>(LogLevel logLevel, EventId eventId, TState state, Except
         }
 
         var timestamp = _clock.GetUtcNow();
-        var traceHeader = _hub.GetTraceHeader() ?? SentryTraceHeader.Empty;
+        SentryLog.GetTraceIdAndSpanId(_hub, out var traceId, out var spanId);
```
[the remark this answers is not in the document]
Flash0ver: Added full unit-test coverage, and checked manually end-to-end that the observed behavior has not changed ... at least for the use cases that I have tested.

test/Sentry.Extensions.Logging.Tests/SentryStructuredLoggerTests.cs
```diff
@@ -51,10 +51,12 @@ public Fixture()
...
-        public void WithTraceHeader(SentryId traceId, SpanId parentSpanId)
+        public void WithActiveSpan(SentryId traceId, SpanId parentSpanId)
         {
-            var traceHeader = new SentryTraceHeader(traceId, parentSpanId, null);
-            Hub.GetTraceHeader().Returns(traceHeader);
+            var span = Substitute.For<ISpan>();
+            span.TraceId.Returns(traceId);
+            span.SpanId.Returns(parentSpanId);
+            Hub.GetSpan().Returns(span);
         }
```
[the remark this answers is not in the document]
Flash0ver: This test is already a part of this changeset.

src/Sentry.Serilog/SentrySink.Structured.cs
```diff
@@ -7,7 +7,7 @@ internal sealed partial class SentrySink
 {
     private static void CaptureStructuredLog(IHub hub, SentryOptions options, LogEvent logEvent, string formatted, string? template)
     {
-        GetTraceIdAndSpanId(hub, out var traceId, out var spanId);
+        SentryLog.GetTraceIdAndSpanId(hub, out var traceId, out var spanId);
```
alexsohn1126, line 10: why do we set `ParentSpanId` to `spanId` in [line 17](https://github.com/getsentry/sentry-dotnet/pull/4565/files#diff-ca9350d91387377a1d3d9f98dda6fdd0b90170d4a6d546b9c8fa1a4eb8f8d1faR17)?

[I saw](https://develop.sentry.dev/sdk/data-model/event-payloads/span/#attributes) `ParentSpanId` is optional, seems like it always gets set here?

src/Sentry.Extensions.Logging/SentryStructuredLogger.cs
alexsohn1126, line 45: Unrelated question:

What is the purpose of `Func<TState, Exception?, string> formatter` parameter? It doesn't seem like we test it either?

Image

alexsohn1126 (commented):
A couple of *possibly dumb* questions 😛

src/Sentry.Serilog/SentrySink.Structured.cs
Flash0ver: These changes are related to the [Structured Logs](https://develop.sentry.dev/sdk/telemetry/logs/) protocol, not the `Span` protocol.

Per protocol, `sentry.trace.parent_span_id` _is_, or better to say _was_, optional.
I noticed that the Dev Docs have changed in the meantime
> `sentry.trace.parent_span_id`: The span id of the span that was active when the log was collected. This should not be set if there was no active span.

We should only look for an Active Span, and remove the ´PropagationContext´ fallback.

It's getting late ... will fix tomorrow.

src/Sentry.Extensions.Logging/SentryStructuredLogger.cs
Flash0ver: The `formatter` is a concept of [Microsoft.Extensions.Logging.ILogger](https://learn.microsoft.com/dotnet/api/microsoft.extensions.logging.ilogger),
where Microsoft's Abstractions are taking care of manifest the Log-Entry into a `string` message,
and we then enrich the _Structured Logs_ with additional attributes, like parameter names and parameter values.

We do _implicitly_ test the `formatter` for _Structured Logs_.
See `Sentry.Extensions.Logging.Tests.SentryStructuredLoggerTests`, where we call `Microsoft.Extensions.Logging` extension methods (e.g. [LoggerExtensions.Log](https://learn.microsoft.com/dotnet/api/microsoft.extensions.logging.loggerextensions.log)) which internally creates a `formatter`, that we then call and assert in the test.

The tests from your screenshot are from `Sentry.Extensions.Logging.Tests.SentryLoggerTests`, which are not testing _Structured Logs_, but instead _Breadcrumbs_ and _Events_, which we also create from `Microsoft.Extensions.Logging` logs, considering the related options (`SentryLoggingOptions.MinimumBreadcrumbLevel` and `SentryLoggingOptions.MinimumEventLevel`).

src/Sentry.Serilog/SentrySink.Structured.cs
Flash0ver: fixed: only set `"sentry.trace.parent_span_id"` when there was a Span active when the Log was recorded

alexsohn1126 (approved):
Looks good!

PR #4566, opened 2025-09-25
WIP fix: Ensure integration logs have sentry.origin set

### LINKED ISSUE #4560: Make sure logs generated from integrations have a `sentry.origin` attribute attached
### Description

When logs are generated via logging integrations, we should attach a sentry.origin attribute to them. Spec here: https://develop.sentry.dev/sdk/telemetry/logs/#sdk-integration-attributes.

alexsohn1126:
Fixes #4560

# Problem

According to our [SDK specification](https://develop.sentry.dev/sdk/telemetry/logs/#sdk-integration-attributes), we have to set `sentry.origin` whenever a log is sent from an integration.

# Solution

`sentry-dotnet` has a few integrations that send out logs. This PR affects the following integrations:
- `Sentry.Extensions.Logging`
- `Sentry.Serilog`

And affects other integrations that use `Sentry.Extensions.Logging` transitively.

# Comparison

Before, logs from `Sentry.Serilog` didn't have `sentry.origin`:
image

After, logs from `Sentry.Serilog` has `sentry.origin` populated:
Screenshot 2025-09-25 at 4 14 59 PM

### DIFF
--- a/src/Sentry.Extensions.Logging/SentryStructuredLogger.cs
+++ b/src/Sentry.Extensions.Logging/SentryStructuredLogger.cs
@@ -89,6 +89,7 @@ internal sealed class SentryStructuredLogger : ILogger
         };
 
         log.SetDefaultAttributes(_options, _sdk);
+        log.SetOrigin("auto");
 
         if (_categoryName is not null)
         {
--- a/src/Sentry.Extensions.Logging/SentryStructuredLogger.cs
+++ b/src/Sentry.Extensions.Logging/SentryStructuredLogger.cs
@@ -89,7 +89,7 @@ internal sealed class SentryStructuredLogger : ILogger
         };
 
         log.SetDefaultAttributes(_options, _sdk);
-        log.SetOrigin("auto");
+        log.SetOrigin("auto.logging.dotnet_extension");
 
         if (_categoryName is not null)
         {
--- a/src/Sentry.Serilog/SentrySink.Structured.cs
+++ b/src/Sentry.Serilog/SentrySink.Structured.cs
@@ -18,6 +18,7 @@ internal sealed partial class SentrySink
         };
 
         log.SetDefaultAttributes(options, Sdk);
+        log.SetOrigin("auto");
 
         foreach (var attribute in attributes)
         {
--- a/src/Sentry.Serilog/SentrySink.Structured.cs
+++ b/src/Sentry.Serilog/SentrySink.Structured.cs
@@ -18,7 +18,7 @@ internal sealed partial class SentrySink
         };
 
         log.SetDefaultAttributes(options, Sdk);
-        log.SetOrigin("auto");
+        log.SetOrigin("auto.logging.serilog");
 
         foreach (var attribute in attributes)
         {
--- a/src/Sentry/SentryLog.cs
+++ b/src/Sentry/SentryLog.cs
@@ -199,6 +199,11 @@ public sealed class SentryLog
         }
     }
 
+    internal void SetOrigin(string origin)
+    {
+        SetAttribute("sentry.origin", origin);
+    }
+
     internal void WriteTo(Utf8JsonWriter writer, IDiagnosticLogger? logger)
     {
         writer.WriteStartObject();
--- a/test/Sentry.AspNetCore.Tests/SentryAspNetCoreStructuredLoggerProviderTests.cs
+++ b/test/Sentry.AspNetCore.Tests/SentryAspNetCoreStructuredLoggerProviderTests.cs
@@ -93,7 +93,7 @@ public class SentryAspNetCoreStructuredLoggerProviderTests
         version.Should().Be(SentryMiddleware.NameAndVersion.Version);
 
         capturedLog.TryGetAttribute("sentry.origin", out object? origin).Should().BeTrue();
-        origin.Should().Be("auto");
+        origin.Should().Be("auto.logging.dotnet_extension");
     }
 
     [Fact]
--- a/test/Sentry.AspNetCore.Tests/SentryAspNetCoreStructuredLoggerProviderTests.cs
+++ b/test/Sentry.AspNetCore.Tests/SentryAspNetCoreStructuredLoggerProviderTests.cs
@@ -91,6 +91,9 @@ public class SentryAspNetCoreStructuredLoggerProviderTests
 
         capturedLog.TryGetAttribute("sentry.sdk.version", out object? version).Should().BeTrue();
         version.Should().Be(SentryMiddleware.NameAndVersion.Version);
+
+        capturedLog.TryGetAttribute("sentry.origin", out object? origin).Should().BeTrue();
+        origin.Should().Be("auto");
     }
 
     [Fact]
--- a/test/Sentry.Extensions.Logging.Tests/SentryStructuredLoggerProviderTests.cs
+++ b/test/Sentry.Extensions.Logging.Tests/SentryStructuredLoggerProviderTests.cs
@@ -93,7 +93,7 @@ public class SentryStructuredLoggerProviderTests
         version.Should().Be(SentryLoggerProvider.NameAndVersion.Version);
 
         capturedLog.TryGetAttribute("sentry.origin", out object? origin).Should().BeTrue();
-        origin.Should().Be("auto");
+        origin.Should().Be("auto.logging.dotnet_extension");
     }
 
     [Fact]
--- a/test/Sentry.Extensions.Logging.Tests/SentryStructuredLoggerProviderTests.cs
+++ b/test/Sentry.Extensions.Logging.Tests/SentryStructuredLoggerProviderTests.cs
@@ -91,6 +91,9 @@ public class SentryStructuredLoggerProviderTests
 
         capturedLog.TryGetAttribute("sentry.sdk.version", out object? version).Should().BeTrue();
         version.Should().Be(SentryLoggerProvider.NameAndVersion.Version);
+
+        capturedLog.TryGetAttribute("sentry.origin", out object? origin).Should().BeTrue();
+        origin.Should().Be("auto");
     }
 
     [Fact]
--- a/test/Sentry.Extensions.Logging.Tests/SentryStructuredLoggerTests.cs
+++ b/test/Sentry.Extensions.Logging.Tests/SentryStructuredLoggerTests.cs
@@ -109,7 +109,7 @@ public class SentryStructuredLoggerTests : IDisposable
         log.ParentSpanId.Should().Be(parentSpanId);
         log.AssertAttribute("sentry.environment", "my-environment");
         log.AssertAttribute("sentry.release", "my-release");
-        log.AssertAttribute("sentry.origin", "auto");
+        log.AssertAttribute("sentry.origin", "auto.logging.dotnet_extension");
         log.AssertAttribute("sentry.sdk.name", "SDK Name");
         log.AssertAttribute("sentry.sdk.version", "SDK Version");
         log.AssertAttribute("category.name", _fixture.CategoryName);
--- a/test/Sentry.Extensions.Logging.Tests/SentryStructuredLoggerTests.cs
+++ b/test/Sentry.Extensions.Logging.Tests/SentryStructuredLoggerTests.cs
@@ -109,6 +109,7 @@ public class SentryStructuredLoggerTests : IDisposable
         log.ParentSpanId.Should().Be(parentSpanId);
         log.AssertAttribute("sentry.environment", "my-environment");
         log.AssertAttribute("sentry.release", "my-release");
+        log.AssertAttribute("sentry.origin", "auto");
         log.AssertAttribute("sentry.sdk.name", "SDK Name");
         log.AssertAttribute("sentry.sdk.version", "SDK Version");
         log.AssertAttribute("category.name", _fixture.CategoryName);
--- a/test/Sentry.Serilog.Tests/SentrySinkTests.Structured.cs
+++ b/test/Sentry.Serilog.Tests/SentrySinkTests.Structured.cs
@@ -118,6 +118,8 @@ public partial class SentrySinkTests
         environment.Should().Be("test-environment");
         log.TryGetAttribute("sentry.release", out object? release).Should().BeTrue();
         release.Should().Be("test-release");
+        log.TryGetAttribute("sentry.origin", out object? origin).Should().BeTrue();
+        origin.Should().Be("auto");
         log.TryGetAttribute("sentry.sdk.name", out object? sdkName).Should().BeTrue();
         sdkName.Should().Be(SentrySink.SdkName);
         log.TryGetAttribute("sentry.sdk.version", out object? sdkVersion).Should().BeTrue();
--- a/test/Sentry.Serilog.Tests/SentrySinkTests.Structured.cs
+++ b/test/Sentry.Serilog.Tests/SentrySinkTests.Structured.cs
@@ -119,7 +119,7 @@ public partial class SentrySinkTests
         log.TryGetAttribute("sentry.release", out object? release).Should().BeTrue();
         release.Should().Be("test-release");
         log.TryGetAttribute("sentry.origin", out object? origin).Should().BeTrue();
-        origin.Should().Be("auto");
+        origin.Should().Be("auto.logging.serilog");
         log.TryGetAttribute("sentry.sdk.name", out object? sdkName).Should().BeTrue();
         sdkName.Should().Be(SentrySink.SdkName);
         log.TryGetAttribute("sentry.sdk.version", out object? sdkVersion).Should().BeTrue();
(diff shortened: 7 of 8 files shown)


alexsohn1126:
@sentry review

alexsohn1126:
bugbot review

alexsohn1126:
bugbot review

alexsohn1126:
@sentry review

CHANGELOG.md
```diff
@@ -11,6 +11,7 @@
 - Upload linked PDBs to fix non-IL-stripped symbolication for iOS ([#4527](https://github.com/getsentry/sentry-dotnet/pull/4527))
 - In MAUI Android apps, generate and inject UUID to APK and upload ProGuard mapping to Sentry with the UUID ([#4532](https://github.com/getsentry/sentry-dotnet/pull/4532))
 - Fixed WASM0001 warning when building Blazor WebAssembly projects ([#4519](https://github.com/getsentry/sentry-dotnet/pull/4519))
+- Populate `sentry.origin` property in logs ([#4566](https://github.com/getsentry/sentry-dotnet/pull/4566))
```
Flash0ver: suggestion: for consistency

```suggestion
- Populate `sentry.origin` default attribute in Structured Logs ([#4566](https://github.com/getsentry/sentry-dotnet/pull/4566))
```

- in the logs protocol it's an "attribute", not a property
- in the CHANGELOG we usually mention "logs" as "Structured Logs"

src/Sentry.Extensions.Logging/SentryStructuredLogger.cs
```diff
@@ -89,6 +89,7 @@ public void Log<TState>(LogLevel logLevel, EventId eventId, TState state, Except
         };
 
         log.SetDefaultAttributes(_options, _sdk);
+        log.SetOrigin("auto.logging.dotnet_extension");
```
Flash0ver: thought: not sure if that's the "right" origin value ... but it may be ... let me think out loud and please post your opinions and preferences

Docs:
- [Logs/SDK Behavior/Default Attributes](https://develop.sentry.dev/sdk/telemetry/logs/#sdk-integration-attributes)
- [Trace Origin](https://develop.sentry.dev/sdk/telemetry/traces/trace-origin/)

### Category
Other SDKs chose
- `auto.log. ...`
- `auto.logger. ...`
- `auto.logging. ...`

First, I was preferring `"log"`, since it's also the _envelope item key_ of the protocol, and short as well.
On second thought, `"logging"` fits the omnipresent `Microsoft.Extensions.Logging` name in the .NET ecosystem.
I personally don't favor `"logger"`, since it's a noun and doesn't sounds too well in my ear for a category name in that context.

### Integration Name / Integration Part
- not sure if we should include `dotnet`, since it's already contained in `sentry.sdk.name` attribute
- the SDK name of `Sentry.Extensions.Logging` is `"sentry.dotnet.extensions.logging"`, which we also set as the `sentry.sdk.name` attribute
  - `... .sentry.dotnet.extensions.logging` may be a bit lengthy
  - `... .extensions` is short but may be ambiguous
  - `... .microsoft` may also not be too unique
  - `... .microsoft.extensions` or `... .microsoft_extensions` perhaps 🤔
  - `... .extensions.logging` or `... .extensions_logging` maybe 🤔

Flash0ver: question: should we have a single origin for all `Microsoft.Extensions.Logging` providers, or a unique origin per `ILoggerProvider`

Currently, we would have a single origin for `ILogger<T>`-created Structured Logs for
- `Sentry.Extensions.Logging`
- `Sentry.AspNetCore`
- `Sentry.Maui`

One the one hand, we have different SDK names (attribute `"sentry.sdk.name"`), on the other hand, I'm uncertain if we should separate the `sentry.origin` too.

E.g.
- `auto.log.aspnetcore` via `SentryAspNetCoreStructuredLoggerProvider`
- `auto.logging.maui` via `SentryMauiLoggerProvider`

Flash0ver: Update: `sentry.sdk.name` and `sentry.origin` are combined for filtering ... so as long as the SDK-Name is unique (which it is), a shared `Microsoft.Extensions.Logging`-style origin could be just fine, also communicating the dependencies of both "derived" `Sentry.Maui` and `Sentry.AspNetCore` to the "base" `Sentry.Extensions.Logging` SDK.

alexsohn1126: Personally, I chose `"logging"` because I thought of `Sentry.Extensions.Logging` (similar to what you mentioned). I think I can get behind using `"log"`, since `Sentry.Serilog` is separate from `Sentry.Extensions.Logging`.

Could we do `... .mel`, short for `Microsoft.Extensions.Logging`? Maybe it's too short? 

`extensions_logging` sounds the best, especially if we change the category to `"log"`. If we keep `logging` -  `... .logging.extensions_logging` - it looks repetitive. 

I'm not sure if `extensions.logging` is allowed, since the 4th section of `sentry.origin` is supposed to be an `integration-part` which I think we don't have to put in our case. From the [trace origin docs](https://develop.sentry.dev/sdk/telemetry/traces/trace-origin/):

> `integration-part`
Optional. Is the part of the integration of the SDK that created the trace or span. This is only useful when one integration creates multiple traces or spans, and you'd like to differentiate. Therefore, most of the time, this will be omitted.

alexsohn1126: Yea, I feel like they should know what SDK they are using by looking at `sentry.sdk.name` and filter based on that instead of `sentry.origin`. I'll keep the current implementation then!

alexsohn1126: Actually, I may like `microsoft_extensions` more. just `extensions_logging` doesn't provide enough context... Like what is it the extension of?

I'll change it to `microsoft_extensions`.

CHANGELOG.md
alexsohn1126: Changed in https://github.com/getsentry/sentry-dotnet/pull/4566/commits/75fa353f79fae02ce6102d88112290422e5f4184

src/Sentry.Extensions.Logging/SentryStructuredLogger.cs
alexsohn1126: changed `logging` to `log`, and changed `dotnet_extension` to `microsoft_extensions` in https://github.com/getsentry/sentry-dotnet/pull/4566/commits/e866c0003ba4b71a8f7933d0830bd777043f76cf

Flash0ver (commented):
I'm not 100 % convinced of the origin names that we use right now, but also don't have a concrete suggestion in store ... if you don't mind, I'll let my mind wander a bit and will get back to you soon.
Also, feedback from other reviewers appreciated. @jamescrosswell what do you think?

jamescrosswell:
> I'm not 100 % convinced of the origin names that we use right now, but also don't have a concrete suggestion in store ... if you don't mind, I'll let my mind wander a bit and will get back to you soon.
> Also, feedback from other reviewers appreciated. @jamescrosswell what do you think?

From memory, we started setting the `sentry.origin` initially to support tracing... so the naming conventions for these were made with that in mind. It's even titled [Trace Origin](https://develop.sentry.dev/sdk/telemetry/traces/trace-origin/) in the dev docs.  

- The first/type part of the origin is obvious (either manual or auto).
- The second/category part is somewhat subjective (e.g. `db` or `ui`) and we try to match whatever the other SDKs are doing here... but some judgement required on our part since the specific databases etc. we're integrating with will often be .NET specific
- The third part is the integration name. Ideally this would match the name we used for the integration in our source... so for example in the `Sentry.EntityFramework` integration we used the origin `auto.db.entity_framework`. The challenge is that our NuGet packages are optimised to minimise dependencies, so where we can integrate from the Sentry.nupkg without any extra dependencies, we do... which I think is why we ended up with stuff like `auto.graphql` ([in the base Sentry NuGet](https://github.com/getsentry/sentry-dotnet/blob/3e70c3126515c87b1851b6a9a7afd99ffbb9aeb7/src/Sentry/SentryGraphQLHttpMessageHandler.cs#L16)). 
- I don't think we use the fourth part / integration-part anywhere in the .NET SDK...

src/Sentry.Extensions.Logging/SentryStructuredLogger.cs
```diff
@@ -89,6 +89,7 @@ public void Log<TState>(LogLevel logLevel, EventId eventId, TState state, Except
         };
 
         log.SetDefaultAttributes(_options, _sdk);
+        log.SetOrigin("auto.log.microsoft_extensions");
```
jamescrosswell:
```suggestion
        log.SetOrigin("auto.log.extensions_logging");
```

Possibly we could go with this, to mirror the way we've named the integration (`Sentry.Extensions.Logging`). It's weird, but at least it's consistently weird.

CHANGELOG.md
```diff
@@ -11,6 +11,7 @@
 - Upload linked PDBs to fix non-IL-stripped symbolication for iOS ([#4527](https://github.com/getsentry/sentry-dotnet/pull/4527))
 - In MAUI Android apps, generate and inject UUID to APK and upload ProGuard mapping to Sentry with the UUID ([#4532](https://github.com/getsentry/sentry-dotnet/pull/4532))
 - Fixed WASM0001 warning when building Blazor WebAssembly projects ([#4519](https://github.com/getsentry/sentry-dotnet/pull/4519))
+- Populate `sentry.origin` default attribute in _Structured Logs_ ([#4566](https://github.com/getsentry/sentry-dotnet/pull/4566))
```
```suggestion
- Structured Logs now have a `sentry.origin` attribute to so it's clearer where these come from ([#4566](https://github.com/getsentry/sentry-dotnet/pull/4566))
```

Ideally we write the change logs to highlight how the change will help our SDK users.

src/Sentry.Extensions.Logging/SentryStructuredLogger.cs
jamescrosswell: Are these always supposed to be `auto` for structured logs? If the SDK user does `ILogger.Log(...)` then shouldn't this be `manual`?

Flash0ver: ah ... I see
like SDK `sentry.dotnet.serilog` -> `auto.log.serilog`
SDK `sentry.dotnet.extensions.logging` -> `auto.log.extensions_logging`

alexsohn1126: This sounds like a good consistent naming! I'll go with this

alexsohn1126: Yep, good point. The [docs](https://develop.sentry.dev/sdk/telemetry/traces/trace-origin/) indeed say we should set the type to `manual` when the user manually sends a log... I'll fix it!

alexsohn1126: changed in https://github.com/getsentry/sentry-dotnet/pull/4566/commits/c291d8eeebac433618bf34afd8b857dc5a1b0ebf

src/Sentry.Extensions.Logging/SentryStructuredLogger.cs
```diff
@@ -89,6 +89,7 @@ public void Log<TState>(LogLevel logLevel, EventId eventId, TState state, Except
         };
 
         log.SetDefaultAttributes(_options, _sdk);
+        log.SetOrigin("auto.log.extensions_logging");
```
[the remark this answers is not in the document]
alexsohn1126: I think it is better to have it as `auto.log.extensions_logging`.

src/Sentry.Extensions.Logging/SentryStructuredLogger.cs
alexsohn1126: Wait, upon more thinking, this has more nuance than I thought. Wouldn't the SDK user pretty much always have to call `ILogger.Log(...)` to log something? Does that mean that `sentry.origin` should always be `manual.[...]`?

The documentation linked above states that:

>`traces`
Required. Can be either manual or auto. manual indicates that the user created the trace or span, auto indicates that the SDK or some integration created it.

It doesn't mention logs, but I assume if a user creates a log, then we put `manual`. So is it always `manual`, since logs cannot be created without some kind of user code calling `ILogger.Log(...)`?

@Flash0ver Do you have any input for this?

alexsohn1126: This is what the docs say for [SDK Integration Attributes](https://develop.sentry.dev/sdk/telemetry/logs/#sdk-integration-attributes):

> If a log is generated by an SDK integration, the SDK should also set the sentry.origin attribute, as per the [Trace Origin](https://develop.sentry.dev/sdk/telemetry/traces/trace-origin/) documentation. It is assumed that logs without a sentry.origin attribute are manually created by the user.
{
  "sentry.origin": "auto.db.graphql"
}

So maybe we differentiate when a user calls `ILogger.Log(...)` vs. `ILogger.LogInformation(...)`?

jamescrosswell: @alexsohn1126 sorry I might have sent you on a goose chase there. 

Speaking with @Flash0ver about this it might be that:
- If the logs are coming via one of our integrations (e.g. Sentry.Extensions.Logging or Sentry.Serilog) then you'd set this to `auto` (since we automatically converted a serilog log or an ILogger log to a Sentry Structured one in that case).
- However if an SDK user [used the Sentry API to create a log directly](https://github.com/getsentry/sentry-dotnet/blob/0a08754494b8fed2708f8b930419f1101134e0d0/samples/Sentry.Samples.Console.Basic/Program.cs#L97-L98) then that would be `manual`

@Flash0ver were you able to confirm this as the desired behaviour?

alexsohn1126: That makes things clear! Thanks 👍

Flash0ver: clarified with @AbhiPrasad 🥔 
- the SDK-Logger (via `SentrySdk.Logger.Log*`) never sends `sentry.origin` ... the absence of `sentry.origin` indicates the Log was captured through the Logger of the respective SDK (`sentry.sdk.name`)
- all integrations always send `auto. ..`, regardless of the Log-Message being created through the integration from user-code or Application-code/Library-code

alexsohn1126:
Whoops, forgot to mention that we will be consolidating the category for `sentry.origin` across other Sentry SDKs! Once we have decided on what category, I'll mark this PR as ready..

src/Sentry.Extensions.Logging/SentryStructuredLogger.cs
alexsohn1126: We have a new spec that clarifies this: https://develop.sentry.dev/sdk/telemetry/logs/#sdk-integration-attributes

alexsohn1126:
The Structured Logs spec was updated. See #4618 for more detail:
- #4618

CHANGELOG.md
```diff
@@ -11,6 +11,7 @@
 - Upload linked PDBs to fix non-IL-stripped symbolication for iOS ([#4527](https://github.com/getsentry/sentry-dotnet/pull/4527))
 - In MAUI Android apps, generate and inject UUID to APK and upload ProGuard mapping to Sentry with the UUID ([#4532](https://github.com/getsentry/sentry-dotnet/pull/4532))
 - Fixed WASM0001 warning when building Blazor WebAssembly projects ([#4519](https://github.com/getsentry/sentry-dotnet/pull/4519))
+- Structured Logs now have a `sentry.origin` attribute to so it's clearer where these come from ([#4566](https://github.com/getsentry/sentry-dotnet/pull/4566))
```
Flash0ver: issue: needs an update from `main` ... `5.16.0` has been released in the meantime

Flash0ver: thought: target branch `version6` rather than `main`?

We now merge new PRs towards the `version6` branch, which will be released in `6.0.0` (and pre-releases), unless it's a critical bug fix.

But ... on the other hand ... this is important for Sentry internally ... so I believe I am convincing myself to actually do merge this against `main` and release in `5.16.1` soon.

alexsohn1126: I wouldn't say this is urgent, at most this is a "nice to have", because it doesn't really affect the functionality whole lot. 

Not only that, it will also take time for the other SDK maintainers to follow the new spec for `sentry.origin` and update theirs.

Thoughts? @jamescrosswell

alexsohn1126: Fixed in https://github.com/getsentry/sentry-dotnet/pull/4566/commits/6608542614820f38b75b4645efa02d50a1d5d855

I'll keep this not resolved though, as we (probably) need to change it if we wanted to merge into `version6`

jamescrosswell: It may be quite a while before the majority of SDK users move to version 6.0.0 so probably worth leaving this in main and making a 5.16.1 release.

It'll get merged into version6 from there (along with a bunch of other stuff).

Issue #4554, opened 2025-09-23
Support `Sentry.SourceGenerators` via Visual Studio 2019 version 16.11


KrzysztofFryzlewicz:
### Environment

SaaS (https://sentry.io/)

### Steps to Reproduce

When building UWP project using Azure DevOps Pipelines we got this error:

Error CS8032: An instance of analyzer Sentry.SourceGenerators.BuildPropertySourceGenerator cannot be created from C:\Users\svcprd_tfbuild\.nuget\packages\sentry\5.15.1\analyzers\dotnet\cs\Sentry.SourceGenerators.dll : Could not load file or assembly 'Microsoft.CodeAnalysis, Version=4.3.0.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35' or one of its dependencies. The system cannot find the file specified..

Microsoft.CodeAnalysis Version=4.3.0.0 is not available in Nuget repository.

Sentry Version=5.15.1

EDIT:
TreatWarningAsErrors is set to true, so this error is acutally a warning:
CSC : warning CS8032: An instance of analyzer Sentry.SourceGenerators.BuildPropertySourceGenerator cannot be created from C:\Users\svcprd_tfbuild\.nuget\packages\sentry\5.14.1\analyzers\dotnet\cs\Sentry.SourceGenerators.dll : Could not load file or assembly 'Microsoft.CodeAnalysis, Version=4.3.0.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35' or one of its dependencies. The system cannot find the file specified..

### Expected Result

Build should finish without problems.

### Actual Result

Build fails with error described above

### Product Area

Unknown

Flash0ver:
Thanks @KrzysztofFryzlewicz for reporting this issue.

One aspect that catches my eye is that you mentioned using `v5.15.1` of the SDK,
but the error contains `\5.14.1\`.
Not sure if there is a restore issue, or just a message from a build before the version update.

Anyway, we need a bit more info about the environment to reproduce the error:
- which version of the .NET SDK / Visual Studio / MSBuild is used on your Azure Pipelines agent?

I have the feeling that this is less of an UWP problem, but more of an issue related to compatibility with MSBuild and the available Roslyn Compiler version.

KrzysztofFryzlewicz:
@Flash0ver yeah, issue exists in 5.14.1 and then I upgraded the package to see if this resolves anything, but no luck.
MSBuild version: 16.0
[...]

PR #4562, opened 2025-09-24
chore: Pin GitHub Actions to commit SHA

alexsohn1126:
Closes #4540

# Problem

According to [GitHub Actions security best practices](https://docs.github.com/en/actions/reference/security/secure-use#using-third-party-actions), we should **pin actions to a full-length commit SHA**.

#skip-changelog

---

> [!NOTE]
> Pins all CI actions and reusable workflows to specific commit SHAs across GitHub workflows and the buildnative composite action.
> 
> - **CI Workflows**:
>   - Pin commonly used actions to commit SHAs: `actions/checkout@v5`, `actions/cache@v4` (and `cache/restore`), `actions/upload-artifact@v4`, `actions/download-artifact@v5`, `codecov/codecov-action`, `microsoft/setup-msbuild`.
>   - Pin Docker-related actions: `docker/login-action`, `docker/setup-qemu-action`, `docker/setup-buildx-action`, `docker/build-push-action`.
>   - Pin analysis/testing actions: `github/codeql-action/{init,analyze}`, `gradle/actions/setup-gradle`, `reactivecircus/android-emulator-runner`, `styfle/cancel-workflow-action`.
>   - Pin reusable workflows and org actions: `getsentry/github-workflows` (danger, updater), `getsentry/action-prepare-release`, `getsentry/github-workflows/sentry-cli/integration-test`.
> - **Composite Action (`.github/actions/buildnative`)**:
>   - Replace `actions/cache@v3` with pinned `actions/cache@v4` SHAs for C and Android supplemental builds.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 4783fed3524ef3c4050dfb8cd4579e6185c1f0b9. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

PR #4569, opened 2025-09-26
test(ci): .NET 5.0 with MSBuild 16

jpnurmi:
This PR adds an integration test that builds and runs a minimal Sentry .NET app with different .NET SDK and MSBuild versions:
- .NET 5.0 (MSBuild 16, excluded on macOS)
- .NET 8.0 (MSBuild 17)
- .NET 9.0 (MSBuild 17)

.NET 5.0 installation and tests are excluded on macOS because .NET 5.0 does not support ARM64 on macOS. However, the test reproduces the warning reported in #4554 both on Windows and Linux:

image

Once #4554 has been fixed, we could consider `TreatWarningsAsErrors=true` to prevent new warnings from surfacing in the future.

Close: #4299

#skip-changelog

---

> [!NOTE]
> Adds MSBuild integration tests for .NET 5/8/9 and updates CI to install .NET 5 SDK and libssl1 on Linux, plus packs during MSBuild.
> 
> - **CI**
>   - Install legacy `libssl1.1` on Linux via `scripts/install-libssl1.sh` in `./github/actions/environment`.
>   - Install `.NET 5.0` SDK (excluded on macOS ARM64) alongside existing SDK setup.
>   - MSBuild job now runs `Restore,Build,Pack` and executes integration tests via `integration-test/msbuild.Tests.ps1`.
> - **Integration Tests**
>   - New `integration-test/msbuild.Tests.ps1` builds/runs a console app via MSBuild across `net5.0` (when supported), `net8.0`, and `net9.0`, validating Sentry capture.
>   - Add `integration-test/nuget5.config` for NuGet 5 (no packageSourceMapping) used with .NET 5.
>   - Update `integration-test/common.ps1` to use `dotnet restore /p:CheckEolTargetFramework=false`.
> - **Scripts**
>   - Add `scripts/install-libssl1.sh` to install `openssl1.1-compat` (Alpine) or `libssl1.1` (Ubuntu).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit dc2ddfde4b9aa289acdd65cece8fede0eae12db1. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/.github/actions/environment/action.yml
+++ b/.github/actions/environment/action.yml
@@ -99,7 +99,7 @@ runs:
     # .NET 5.0 does not support ARM64 on macOS
     - name: Install .NET 5.0 SDK
       if: ${{ runner.os != 'macOS' || runner.arch != 'ARM64' }}
-      uses: actions/setup-dotnet@v4
+      uses: actions/setup-dotnet@67a3573c9a986a3f9c594539f4ab511d57bb3ce9 # v4.3.1
       with:
         dotnet-version: 5.0.x
 


scripts/install-libssl1.sh
```diff
@@ -0,0 +1,22 @@
...
+    # https://github.com/actions/runner-images/blob/d43555be6577f2ac4e4f78bf683c520687891e1b/images/ubuntu/scripts/build/install-sqlpackage.sh#L11-L21
+    if [ "$(dpkg --print-architecture)" = "arm64" ]; then
+        echo "deb http://ports.ubuntu.com/ubuntu-ports focal-security main" | tee /etc/apt/sources.list.d/focal-security.list
+    else
+        echo "deb http://security.ubuntu.com/ubuntu focal-security main" | tee /etc/apt/sources.list.d/focal-security.list
+    fi
+    apt-get update
+    apt-get install -y --no-install-recommends libssl1.1
+    rm /etc/apt/sources.list.d/focal-security.list
+    apt-get update
```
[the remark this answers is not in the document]
jpnurmi: It is meant to be executed as `sudo ./scripts/install-libssl1.sh` to make the user give explicit permissions rather than triggering unclear root password prompts from within the script.

jpnurmi:
@Flash0ver Turns out we didn't need `microsoft/setup-msbuild` nor `choco install visualstudio2019buildtools` for this.

On the contrary, injecting MSBuild 16 into `PATH` would break the `dotnet` (9) command that we need to be able to execute for some preparational steps in the integration tests with `global.json` in the project root.

Fortunately, the same warning can be produced with `dotnet msbuild` with .NET 5.0 and MSBuild 16 installed on the side.

.github/actions/environment/action.yml
```diff
@@ -24,6 +24,12 @@ runs:
       shell: bash
       run: sudo chmod 666 /var/run/docker.sock
 
+    # Install old deprecated libssl1 for .NET 5.0 on Linux
```
Flash0ver, line 27: question: are we "pessimizing" non-related jobs?

We only need this and the installation of .NET 5.0 SDK for the `msbuild` job's integration tests.
Is it "worth" to do this for other jobs and workflows too?
Or is the added download and installation time insignificant and not really worth moving it into the `msbuild` job, potentially cluttering it.

integration-test/msbuild.Tests.ps1
```diff
@@ -0,0 +1,101 @@
...
+        $frameworks += @(
+            @{ framework = 'net8.0'; sdk = '8.0.400' },
+            @{ framework = 'net9.0'; sdk = '9.0.300' }
+        )
+    }
+
+    Context '(<framework>)' -ForEach $frameworks {
+        BeforeEach {
+            Write-Host "::group::Create msbuild-app"
+            DotnetNew 'console' 'msbuild-app' $framework
```
Flash0ver: suggestion: suppress  `NETSDK1138` for `net5.0` apps

The _implicit restore_ of `dotnet new`, as well as the explicit `dotnet restore`, are reporting a warning:
> The target framework 'net5.0' is out of support and will not receive security updates in the future. Please refer to https://aka.ms/dotnet-core-support for more information about the support policy.

See [NETSDK1138: The target framework is out of support](https://learn.microsoft.com/dotnet/core/tools/sdk-errors/netsdk1138)

Since we generally expect this for `msbuild.Tests.ps1`, and in particular for `net5.0` applications, I'm wondering if we may suppress this warning.

Ideas / Suggestions / Thoughts
- Since we do an explicit `dotnet restore` down below, we may want to suppress the _implicit restore_ of `dotnet new` via `--no-restore`
- For the explicit `dotnet restore` below, we could pass `/p:CheckEolTargetFramework=false`
  - for all TFMs, or only for `net5.0` ... I guess for all is just fine for this single test case
- For `dotnet msbuild -t:Build` below, I think we should also be able to add `/p:CheckEolTargetFramework=false`
- I don't think we need to change `dotnet msbuild -t:Build` ... I don't think this is reporting `NETSSDK1138`

Context
- [.NET SDK error list](https://learn.microsoft.com/en-us/dotnet/core/tools/sdk-errors/)
- [NETSDK1138](https://learn.microsoft.com/en-us/dotnet/core/tools/sdk-errors/netsdk1138)
- [Blog: Fixing build warning NETSSDK1138 when building projects with end-of-life .NET frameworks](https://andrewlock.net/fixing-build-warning-netsdk1138-when-building-end-of-life-projects-with-dotnet-5/)

Flash0ver (commented):
This is `Amazing.MaxValue`! It will give us more coverage and confidence when making changes to non-`/lib/` Assets, like `/build/(*.props|.targets)` and `/analyzers/*` (Compiler Extensions like Source Generators)
🥔

.github/actions/environment/action.yml
jpnurmi: Thanks, I double-checked with `strace -f pwsh integration-test/msbuild.Tests.ps1` that the old `libssl1` doesn't seem to be accessed at all during .NET 8.0 and 9.0 tests.

During .NET 5.0 tests, both `libssl1` and `libssl3` are being accessed as expected because some preparation operations execute through .NET 9's `dotnet` command:
```
[pid 537855] openat(AT_FDCWD, "/lib/x86_64-linux-gnu/libssl.so.3", O_RDONLY|O_CLOEXEC) = 79
...
[pid 537940] openat(AT_FDCWD, "/lib/x86_64-linux-gnu/libssl.so.1.1", O_RDONLY|O_CLOEXEC) = 61
```

However, during 8.0 and 9.0 tests, only `libssl3` is accessed as it should:
```
[pid 538354] openat(AT_FDCWD, "/lib/x86_64-linux-gnu/libssl.so.3", O_RDONLY|O_CLOEXEC) = 79
```

integration-test/common.ps1
```diff
@@ -175,7 +175,7 @@ BeforeAll {
     function DotnetNew([string] $type, [string] $name, [string] $framework)
     {
         Remove-Item -Path $name -Recurse -Force -ErrorAction SilentlyContinue
-        dotnet new $type --output $name --framework $framework | ForEach-Object { Write-Host $_ }
+        dotnet new $type --output $name --framework $framework --no-restore | ForEach-Object { Write-Host $_ }
```
Flash0ver: thought: This may break existing usages of `DotnetNew` that may depend on the _implicit restore_ 🤔

Could/should we make `--no-restore` and optional parameter,
where the default is still executing the _implicit restore_
and when explicitly set (to `true`), then we pass `--no-restore`.

jpnurmi: I didn't realize it was template-specific. `dotnet new console` has the `--no-restore` option but `dotnet new maui` doesn't :expressionless:

.github/actions/environment/action.yml
```diff
@@ -90,6 +96,13 @@ runs:
         global-json-file: global.json
         dotnet-version: 8.0.x
 
+    # .NET 5.0 does not support ARM64 on macOS
+    - name: Install .NET 5.0 SDK
+      if: ${{ runner.os != 'macOS' }}
+      uses: actions/setup-dotnet@v4
+      with:
+        dotnet-version: 5.0.x
```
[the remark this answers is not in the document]
jpnurmi: macos-15 is arm64

integration-test/common.ps1
Flash0ver: oh really?! ... I also thought this was an option on `dotnet new`, and not an individual option per _template_ 🤯

jpnurmi: I ended up calling `dotnet new console --no-restore` + `AddPackageReference Sentry` directly because we don't need all the other stuff in the `DotnetNew` helper.

.github/actions/environment/action.yml
```diff
@@ -90,6 +96,13 @@ runs:
         global-json-file: global.json
         dotnet-version: 8.0.x
 
+    # .NET 5.0 does not support ARM64 on macOS
+    - name: Install .NET 5.0 SDK
+      if: ${{ runner.os != 'macOS' || runner.arch != 'ARM64' }}
+      uses: actions/setup-dotnet@v4
```
Flash0ver:
```suggestion
      uses: actions/setup-dotnet@67a3573c9a986a3f9c594539f4ab511d57bb3ce9 # v4.3.1
```

Considering #4562, currently under review as well.

Flash0ver (approved):
Well done!
- no more `NETSDK1138` on .NET 5
- but `CS8032` reported on .NET 5 SDK, indeed utilizing MSBuild 16.11

image

Got one final nitpick related to another PR currently in review.

.github/actions/environment/action.yml
jpnurmi: Thanks, good catch! I wonder if we could have a simple CI check in the future to ensure that all actions are pinned to SHA1s. 🤔

Flash0ver: hmmm ... good idea ... I believe through CodeQL we only verify that non-trusted/third-party Actions are SHA-pinned

PR #4571, opened 2025-09-26
ci: use `global.json` for `actions/setup-dotnet`

jpnurmi:
No need to bump the SDK version in both `global.json` and `.github/actions/environment/action.yml` anymore...

#skip-changelog


jamescrosswell (approved):
Nice! 😎

Issue #3686, opened 2024-10-16
Use bools and enums with Interlocked.CompareExchange


jamescrosswell:
In various places in the solution we're using `int` instead of `bool` so that we can make use of the `Interlocked` class, which historically did not support `bool` and `enum`. For example:
https://github.com/getsentry/sentry-dotnet/blob/96a116c6076d358d6e0c06a1f012c1ed2c7a65d1/src/Sentry.Profiling/SampleProfilerSession.cs#L58-L68

With .net9 [Interlocked.CompareExchange supports more types](https://learn.microsoft.com/en-us/dotnet/core/whats-new/dotnet-9/libraries#interlockedcompareexchange-for-more-types). 

We should leverage the new type support to tidy up our code.

logiclrd:
How does this work with the fact that `Sentry.NET` explicitly targets .NET Standard 2.1?
```
  Sentry netstandard2.1 failed with 8 error(s) (10.6s)
    C:\code\sentry-dotnet\src\Sentry\TransactionTracer.cs(365,25): error CS0452: The type 'bool' must be a reference type in order to use it as parameter 'T' in the generic type or method 'Interlocked.Exchange<T>(ref T, T)'
    C:\code\sentry-dotnet\src\Sentry\Threading\ScopedCountdownLock.cs(21,22): error CS0029: Cannot implicitly convert type 'int' to 'bool'
    C:\code\sentry-dotnet\src\Sentry\Threading\ScopedCountdownLock.cs(34,27): error CS0019: Operator '==' cannot be applied to operands of type 'bool' and 'int'
    C:\code\sentry-dotnet\src\Sentry\Threading\ScopedCountdownLock.cs(40,32): error CS0019: Operator '==' cannot be applied to operands of type 'bool' and 'int'
    C:\code\sentry-dotnet\src\Sentry\Threading\ScopedCountdownLock.cs(82,25): error CS0452: The type 'bool' must be a reference type in order to use it as parameter 'T' in the generic type or method 'Interlocked.CompareExchange<T>(ref T, T, T)'
    C:\code\sentry-dotnet\src\Sentry\Threading\ScopedCountdownLock.cs(97,25): error CS0452: The type 'bool' must be a reference type in order to use it as parameter 'T' in the generic type or method 'Interlocked.CompareExchange<T>(ref T, T, T)'
...
```

Are you proposing removing targets that don't support the new overloads, or is this issue a non-starter?

[...]

PR #4585, opened 2025-10-01
Use new Interlocked.Exchange/CompareExchange overloads that support bool values

logiclrd:
This PR updates places that use `Interlocked.Exchange` and/or `Interlocked.CompareExchange` with `int` values that are really booleans because, historically, the type `bool` wasn't supported to use the new overloads in .NET 9.0+ that do support `bool`. As the code has targets older than .NET 9.0, these are all used conditionally, with fallback to `int` on older platforms. Constants named `TRUE` and `FALSE` are used, tracking the actual type of the backing field, so that the actual code doesn't need separate implementations.

Closes: #3686


src/Sentry.Profiling/SamplingTransactionProfilerFactory.cs
```diff
@@ -6,14 +6,21 @@ namespace Sentry.Profiling;
...
+    internal static bool _inProgress = FALSE;
+
+    const bool TRUE = true;
+    const bool FALSE = false;
+#else
+    internal static int _inProgress = FALSE;
+
+    const int TRUE = 1;
+    const int FALSE = 0;
+#endif
```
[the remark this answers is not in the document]
logiclrd: Well that's embarrassing. 😛

src/Sentry.Profiling/SampleProfilerSession.cs
```diff
@@ -56,18 +56,27 @@ private SampleProfilerSession(SentryStopwatch stopwatch, EventPipeSession sessio
...
+    private static int _throwOnNextStartupForTests = FALSE;
+
+    const int TRUE = 1;
+    const int FALSE = 0;
+#endif
 
     internal static bool ThrowOnNextStartupForTests
     {
-        get { return Interlocked.CompareExchange(ref _throwOnNextStartupForTests, 1, 1) == 1; }
+        get { return _throwOnNextStartupForTests != FALSE; }
```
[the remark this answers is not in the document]
logiclrd: I'm not entirely convinced there's a bug here, as this property is only used from outside of the object in the context of automated tests.

logiclrd:
If you accept this, please consider adding the `hacktoberfest-accepted` label to the PR (or the `hacktoberfest` topic to the repository as a whole). 😸

logiclrd:
Rebased to resolve merge conflict in `CHANGELOG.md`.

CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+- Use new `Interlocked.Exchange`/`CompareExchange` overloads that support `bool` values ([#4585](https://github.com/getsentry/sentry-dotnet/pull/4585))
```
Flash0ver: thought: I'm a bit uncertain about the `CHANGELOG.md` entry.

**Missing Heading** ... which heading to use?

Not sure if this entry fits either as `Features` or `Fixes`.
Usually, we document measurable performance improvements as `Fixes`.
But this change is less of a performance improvement ... although it may be. From what I can see is that the invocation of `Interlocked.(Compare)Exchange` is barely measurable. But we might be allocating less memory, especially in our instantiations of `TransactionTracer`, when changing from a 32-bit type to an 8-bit type we could allocate less ... depending on alignment.

@jamescrosswell, what do you think?

src/Sentry.EntityFramework/SentryDatabaseLogging.cs
```diff
@@ -5,14 +5,24 @@ namespace Sentry.EntityFramework;
...
+    private static bool _init;
+
+    const bool TRUE = true;
+    const bool FALSE = false;
+#else
+    private static int _init;
+
+    const int TRUE = 1;
+    const int FALSE = 0;
+#endif
```
Flash0ver: suggestion: consolidate and reuse

I have the same thought as @jamescrosswell:
The main intention of #3686 is to tidy up the code, and - as a nice side-effect - potentially (depending on alignment) allocate a bit less for supported TFMs. But duplicating the same pattern in multiple documents seems cluttering and error-prone to me.

I also suggest something similar:

```CSharp
// Sentry/Aliasing/InterlockedHelpers.cs
#if NET9_0_OR_GREATER
global using InterlockedBoolean = bool;
#else
global using InterlockedBoolean = int;
#endif

global using static Sentry.Aliasing.InterlockedHelpers;

namespace Sentry.Aliasing;

internal static class InterlockedHelpers
{
#if NET9_0_OR_GREATER
    internal const bool InterlockedFalse = false;
    internal const bool InterlockedTrue = true;
#else
    internal const int InterlockedFalse = 0;
    internal const int InterlockedTrue = 1;
#endif
}

// usage
private InterlockedBoolean _cancelIdleTimeout;
_ = Interlocked.Exchange(ref _cancelIdleTimeout, InterlockedFalse) == InterlockedTrue;
```

We could put this type into a namespace like `Aliases` or `Aliasing`,
which would then also be the namespace for getsentry/sentry-dotnet#3689.

Flash0ver: But the problem is that _Rider_ is now suggesting to use the _aliased_ type name for every variable declaration of type `int` and `bool` in the Project (and should we "share" it, then _Solution_-wide).

```CSharp
private bool _field;
public bool Property { get; set; }
public void Method(bool param)
{
  bool local = param;
}
```
> Use type alias 'InterlockedBoolean'

I would love to have a consolidation like this or what @jamescrosswell suggested,
but I'm a bit uncertain of how to streamline the editor experience to avoid accidental code fixers / refactorings.
I also need to check the IDE behavior of _Visual Studio_.

Flash0ver (commented):
Thanks @logiclrd for the PR!

Is there a way we can consolidate / centralize the pattern without deteriorating the Editor-experience?

logiclrd:
> Is there a way we can consolidate / centralize the pattern without deteriorating the Editor-experience?

What about wrapping the `Interlocked` functionality entirely?

```
#if NET9_0_OR_GREATER
using TBool = bool;
#else
using TBool = int;
#endif

internal struct InterlockedBoolean
{
  private TBool _value;

#if NET9_0_OR_GREATER
  private const TBool True = true;
  private const TBool False = false;
#else
  private const TBool True = 1;
  private const TBool False = 0;
#endif

  public InterlockedBoolean() {}

  [MethodImpl(MethodImplOptions.AggressiveInlining)]
  public InterlockedBoolean(bool value) { _value = value ? True : False; }

  [MethodImpl(MethodImplOptions.AggressiveInlining)]
  public static implicit operator bool(InterlockedBoolean _this) => (_this._value != False);
  [MethodImpl(MethodImplOptions.AggressiveInlining)]
  public static implicit operator InterlockedBoolean(bool _this) => new InterlockedBoolean(_this);

  [MethodImpl(MethodImplOptions.AggressiveInlining)]
  public bool Exchange(bool newValue)
  {
    TBool localNewValue = newValue ? True : False;

    TBool localReturnValue = Interlocked.Exchange(ref _value, localNewValue);

    return (localReturnValue != False);
  }

  [MethodImpl(MethodImplOptions.AggressiveInlining)]
  public bool CompareExchange(bool value, bool comparand)
  {
    TBool localValue = value ? True : False;
    TBool localComparand = comparand ? True : False;

    TBool localReturnValue = Interlocked.CompareExchange(ref _value, localValue, localComparand);

    return (localReturnValue != False);
  }
}
```

This one file would still be subject to Rider objecting to having two names for `bool`, but perhaps that could be shut off locally with a `#pragma`??

logiclrd:
I have proofed this out, I think it fits the bill :-) No repetition, concise usage sites, centralization of the version-specific logic.

```
    // We only allow a single profile so let's keep track of the current status.
    internal InterlockedBoolean _inProgress = false;
...
    public ITransactionProfiler? Start(ITransactionTracer _, CancellationToken cancellationToken)
    {
        // Start a profiler if one wasn't running yet.
        if (!_errorLogged && !_inProgress.Exchange(true))
        {
            if (!_sessionTask.IsCompleted)
            {
                _options.LogWarning("Cannot start a sampling profiler, the session hasn't started yet.");
                _inProgress = false;
                return null;
            }
...
                return new SamplingTransactionProfiler(_options, _sessionTask.Result, TIME_LIMIT_MS, cancellationToken)
                {
                    OnFinish = () => _inProgress = false
                };
```

And:

```
    private InterlockedBoolean _isEngaged;
...
    internal LockScope TryEnterLockScope()
    {
        if (_isEngaged.CompareExchange(true, false) == false)
        {
...
    private void ExitLockScope()
    {
        Debug.Assert(_event.IsSet);
        _event.Reset(); // reset the signaled event to the initial count of 1, so that new 'CounterScope' instances can be entered again

        if (_isEngaged.CompareExchange(false, true) != true)
        {
            Debug.Fail("The Lock should have not been disengaged without being engaged first.");
        }
    }

```

logiclrd:
@Flash0ver I installed Rider to see this suggestion ("Use type alias") for myself and see if I could find a way to disable it locally, but on a fresh installation, it doesn't seem to be giving that suggestion. Is there something I need to toggle to enable it, do you know??

logiclrd:
> @Flash0ver I installed Rider to see this suggestion ("Use type alias") for myself and see if I could find a way to disable it locally, but on a fresh installation, it doesn't seem to be giving that suggestion. Is there something I need to toggle to enable it, do you know??

Never mind, I figured it out :-) Looks like that suggestion only applies to `global using`s. As this proposed version doesn't introduce any `global using`s, that suggestion doesn't appear.

CHANGELOG.md
jamescrosswell: We probably don't even need a changelog entry for this tbh - it's not really a user facing change. 

@logiclrd you could remove the changelog entry and add `#skip-changelog` to the PR description.

src/Sentry.EntityFramework/SentryDatabaseLogging.cs
logiclrd: I was waiting to see if there was any specific feedback to my proposal, but I decided enough time had passed and I've pushed it up. Does the current iteration look better?

logiclrd:
Per the discussion earlier, I've reverted the change to `CHANGELOG.md`.

And, I've committed a change that switches this over to the proposal I most recently outlined.

Let me know if you'd like me to squash the commits down before merging (to reformulate the commit message), or if you'll just do it in the PR merge or whatnot. 🙂 

It also occurs to me that this is now a new testable unit.

test/Sentry.Tests/Internals/InterlockedBooleanTests.cs
```diff
@@ -0,0 +1,149 @@
...
+    [InlineData(true, false, true)]
+    [InlineData(true, true, false)]
+    [InlineData(true, true, true)]
+    public void InterlockedBoolean_CompareExchange_ReturnsExpected(bool initialState, bool comparand, bool newValue)
+    {
+        // Arrange
+        var sut = new InterlockedBoolean(initialState);
+
+        // Act
+        var result = sut.CompareExchange(comparand, newValue);
```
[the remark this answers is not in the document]
logiclrd: Oops. I fixed that in the other `.CompareExchange` test, but it isn't actually relevant in the context of this test so I overlooked it.

src/Sentry.EntityFramework/SentryDatabaseLogging.cs
jamescrosswell: > Does the current iteration look better?

Nice! I think this is easier to read and tidier than the original code (where we didn't take advantage of the new Interlocked types). So best of both worlds basically 🎉

logiclrd:
```
CSC : error CS8785: Generator 'JsonSourceGenerator' failed to generate source. It will not contribute to the output and compilation errors may occur as a result. Exception was of type 'InvalidOperationException' with message 'Unreachable'. [D:\a\sentry-dotnet\sentry-dotnet\test\Sentry.Tests\Sentry.Tests.csproj::TargetFramework=net48]
```

This is a weird build error and I'm not sure how to go about debugging it. In an earlier iteration, I had attributed the members of `InterlockedBoolean` with `[MethodImpl(MethodImplOptions.AggressiveInlining)]` and it triggered this build error. After some fiddling around, I discovered that when I removed the `[MethodImpl]` attributes, it went away for me. But now it's back. ☹️

src/Sentry.EntityFramework/SentryDatabaseLogging.cs
```diff
@@ -1,18 +1,20 @@
 namespace Sentry.EntityFramework;
 
+using Sentry.Internal;
```
Flash0ver, line 1: suggestion: move to top of file

We usually have _Using Directives_ before _Namespace Declarations_.

We may want to add [IDE0065](https://learn.microsoft.com/dotnet/fundamentals/code-analysis/style-rules/ide0065) to our `.editorconfig` to actively document that preference.

src/Sentry.EntityFramework/SentryDatabaseLogging.cs
```diff
@@ -1,18 +1,20 @@
...
+
 /// <summary>
 /// Sentry Database Logger
 /// </summary>
 internal static class SentryDatabaseLogging
 {
-    private static int Init;
+    private static InterlockedBoolean _init;
```
Flash0ver: suggestion: please keep as `Init`

See also
https://github.com/getsentry/sentry-dotnet/blob/e58a8f1067a25378f931e7cfd47f74368dc259c2/.editorconfig#L74-L82

src/Sentry/Ben.BlockingDetector/DetectBlockingSynchronizationContext.cs
```diff
@@ -10,8 +10,8 @@ internal sealed class DetectBlockingSynchronizationContext : SynchronizationCont
 
         internal int _isSuppressed;
 
-        internal void Suppress() => Interlocked.Exchange(ref _isSuppressed, _isSuppressed + 1);
-        internal void Restore() => Interlocked.Exchange(ref _isSuppressed, _isSuppressed - 1);
+        internal void Suppress() => Interlocked.Increment(ref _isSuppressed);
```
Flash0ver, line 13: thought: unexpected side effects?

Nice catch! ... this looks actually like a fix ... from non-thread-safe to being thread-safe.
I wonder if this may have unexpected (but probably desired) side effects 🤔

src/Sentry/Threading/ScopedCountdownLock.cs
```diff
@@ -1,5 +1,7 @@
 namespace Sentry.Threading;
 
+using Sentry.Internal;
```
Flash0ver, line 1: suggestion: move outside of the namespace, rather than inside the namespace

test/Sentry.Tests/Internals/InterlockedBooleanTests.cs
```diff
@@ -0,0 +1,149 @@
+namespace Sentry.Tests.Internals;
+
+#if NET9_0_OR_GREATER
+using TBool = bool;
+#else
+using TBool = int;
+#endif
```
Flash0ver, line 5: suggestion: for consistency with other files, please move the alias directives outside of the namespace

test/Sentry.Tests/Internals/InterlockedBooleanTests.cs
```diff
@@ -0,0 +1,149 @@
...
+#if NET9_0_OR_GREATER
+    private const TBool True = true;
+    private const TBool False = false;
+#else
+    private const TBool True = 1;
+    private const TBool False = 0;
+#endif
+
+    private TBool ToTBool(bool value) => value ? True : False;
+    private bool FromTBool(TBool value) => (value != False);
```
Flash0ver, line 20: thought: to me the `T` in both `ToTBool` and `FromTBool` is a bit hard to read.

I'm wondering if we'd invert both methods to be
```CSharp
private TBool FromBoolean(bool value) => ...
private bool ToBoolean(TBool value) => ...
```

could help readability 🤔

But I guess this may be just a matter of stylistic preference.

test/Sentry.Tests/Internals/InterlockedBooleanTests.cs
```diff
@@ -0,0 +1,149 @@
...
+
+    [Theory]
+    [InlineData(false)]
+    [InlineData(true)]
+    public void InterlockedBoolean_Constructor_ConstructsExpected(bool value)
+    {
+        // Arrange
+        var tboolValue = ToTBool(value);
```
Flash0ver: suggestion (non-blocking): rename `tbool...` to `expected`

Would you mind renaming
- `tboolValue` (2x)
- `tboolNewValue`
- `tboolExpected`

to `expected`?

To me `tbool` is a bit hard to read.
So that we don't name the local variable after the actual type, but the intended meaning of the value.

test/Sentry.Tests/Internals/InterlockedBooleanTests.cs
```diff
@@ -0,0 +1,149 @@
...
+    [InlineData(true, true)]
+    public void InterlockedBoolean_Exchange_SetsExpectedNewState(bool initialState, bool newValue)
+    {
+        // Arrange
+        var sut = new InterlockedBoolean(initialState);
+
+        var tboolNewValue = ToTBool(newValue);
+
+        // Act
+        var result = sut.Exchange(newValue);
```
Flash0ver: suggestion: unused

I believe `result` is unused ...
either remove the local variable, or explicitly discard it via `_ = sut.Exchange(newValue);`

src/Sentry/Internal/InterlockedBoolean.cs
```diff
@@ -0,0 +1,46 @@
...
+#else
+    private const TBool True = 1;
+    private const TBool False = 0;
+#endif
+
+    public InterlockedBoolean() { }
+
+    public InterlockedBoolean(bool value) { _value = value ? True : False; }
+
+    public static implicit operator bool(InterlockedBoolean? _this) => (_this != null) && (_this.Value._value != False);
```
Flash0ver: suggestion: do not pass `Nullable<InterlockedBoolean>`

From what I can see, we don't have any usages converting a `Nullable<InterlockedBoolean>` to a `System.Boolean`.
In order to avoid some implicit conversions there, please change this to be non-nullable, for best performance.

src/Sentry/Internal/InterlockedBoolean.cs
```diff
@@ -0,0 +1,46 @@
...
+    public static implicit operator InterlockedBoolean(bool _this) => new InterlockedBoolean(_this);
```
Flash0ver: suggestion: avoid `_` in parameter names

Parameter names leading with an `_` get flagged by Rider / R#.
- either rename the parameter to `value`
- or escape the keyword via `@this`

src/Sentry/Internal/InterlockedBoolean.cs
```diff
@@ -0,0 +1,46 @@
...
+
+#if NET9_0_OR_GREATER
+using TBool = bool;
+#else
+using TBool = int;
+#endif
+
+internal struct InterlockedBoolean
+{
+    internal volatile TBool _value;
```
Flash0ver: suggestion: make the field `private`

In order to avoid any wrong usages, please make the field `private`,
and instead offer an `internal` readonly-property for testing.

src/Sentry/Internal/InterlockedBoolean.cs
```diff
@@ -0,0 +1,46 @@
...
+    public static implicit operator InterlockedBoolean(bool _this) => new InterlockedBoolean(_this);
+
+    public bool Exchange(bool newValue)
+    {
+        TBool localNewValue = newValue ? True : False;
+
+        TBool localReturnValue = Interlocked.Exchange(ref _value, localNewValue);
+
+        return (localReturnValue != False);
+    }
```
Flash0ver, line 42: suggestion: potentially more efficient implementation ... candidate for a follow-up PR

I am wondering if we can make the implementation for `NET9_0_OR_GREATER` more efficient by removing `TBool localNewValue = newValue ? True : False;` which is only required for older TFMs.

Something like:
```CSharp
// ...
#if NET9_0_OR_GREATER
    internal volatile bool _value;
#else
    internal volatile int _value;
#endif
// ...
public bool Exchange(bool newValue)
{
#if NET9_0_OR_GREATER
    return Interlocked.Exchange(ref _value, newValue);
#else
    int localNewValue = newValue ? 1 : 0;
    int localReturnValue = Interlocked.Exchange(ref _value, localNewValue);
    return (localReturnValue != 0);
#endif
}
// ...
```

This way we wouldn't need any `TBool` alias anymore, which in my opinion becomes a bit more readable,
and the `NET9_0_OR_GREATER` implementation should be every so slightly more efficient (unless the JIT sees through it and optimized it).

However, I am quite happy to having this as a follow-up PR, if you'd like ... as the current implementation is functional and this change is about style and performance.

Flash0ver (commented):
Great implementation. And great test coverage!

I got a couple of nitpicks about style and naming,
and a candidate for a follow-up PR around performance and maintainability.

I'll take a look at `CS8785` ... this seems to be unrelated to this PR ... and on Windows only.
i.e. the usage of `MethodImplAttribute` should have no effect on this and you should be able to bring it back.

Flash0ver:
> This is a weird build error and I'm not sure how to go about debugging it. In an earlier iteration, I had attributed the members of `InterlockedBoolean` with `[MethodImpl(MethodImplOptions.AggressiveInlining)]` and it triggered this build error. After some fiddling around, I discovered that when I removed the `[MethodImpl]` attributes, it went away for me. But now it's back. ☹️

Warning (as error) `CS8785` in `JsonSourceGenerator` in `System.Text.Json (v6.0.10)` is an interesting one:

We're using an older version there, for .NET Standard and .NET Framework, triggered in `Sentry.Tests` via `net48`, which is not entirely forward compatible. Since it's based on an older version of Roslyn, it seems to have issues with [C# 12.0 allowing predefined types as alias](https://learn.microsoft.com/dotnet/csharp/language-reference/proposals/csharp-12.0/using-alias-types) in _Using Directives_ ... in our case `int` and `bool`.

It should work again, when you change the test `test/Sentry.Tests/Internals/InterlockedBooleanTests.cs` to use _Qualified Type Names_ instead:
```diff
 #if NET9_0_OR_GREATER
-using TBool = bool;
+using TBool = System.Boolean;
 #else
-using TBool = int;
+using TBool = System.Int32;
 #endif
```

So you should be able to bring the `MethodImplAttribute` back.

Flash0ver:
> If you accept this, please consider adding the `hacktoberfest-accepted` label to the PR (or the `hacktoberfest` topic to the repository as a whole). 😸

done ✅ (as topic)

src/Sentry/Internal/InterlockedBoolean.cs
logiclrd: I'm not able to predict exactly what it'll do, but I actually wouldn't be surprised if the JIT's code optimization just treated the same storage as `bool` and `int` contextually. I'm pretty sure in the ABI, a parameter on the call stack is always machine word-aligned, so on x86 the `bool` is internally an `Int32` and on x64 the `bool` is an `Int64`. The optimizer _might_ notice this and simply pass the memory address of the parameter on the stack in as the `ref int` for `Interlocked.Exchange`. I wouldn't depend on this, of course, but just to say that there might not really be a hugely noticeable speed difference.

Now I'm curious 🙂 I might do some poking just to see.

test/Sentry.Tests/Internals/InterlockedBooleanTests.cs
logiclrd: Mm, but the problem there is, "from boolean to what" and "to boolean from what"? Obviously, the code's not complex and the meaning surfaces, but it doesn't read naturally. In the inverse, the `value` being passed in is declared a `bool` on a nearby line, and the function name tells you what it's converting to.

It's such a minor thing, though, it's like arguing a technical point in a debate more than anything 😁

test/Sentry.Tests/Internals/InterlockedBooleanTests.cs
logiclrd: I like the convention of just always having the wording be `actual.Should().Be(expected)`. In the cases where the `expected` is a directly a parameter, do you think it would be better to rename the parameter to `expected`, or to copy the parameter into a local called `expected` during the Arrange phase?

logiclrd: For what it's worth, I like this:
```
    [InlineData(true, true)]
    public void InterlockedBoolean_Exchange_ReturnsExpected(bool initialState, bool newValue)
    {
        // Arrange
        var sut = new InterlockedBoolean(initialState);
        var expected = initialState;

        // Act
        var result = sut.Exchange(newValue);

        // Assert
        result.Should().Be(expected);
    }
```
...more than this:
```
    [InlineData(true, true)]
    public void InterlockedBoolean_Exchange_ReturnsExpected(bool expected, bool newValue)
    {
        // Arrange
        var sut = new InterlockedBoolean(expected);

        // Act
        var result = sut.Exchange(newValue);

        // Assert
        result.Should().Be(expected);
    }
```

logiclrd:
Okay, I think that got all of the points of feedback. I checked three times, but if I missed any, let me know.

src/Sentry.Profiling/SampleProfilerSession.cs
```diff
@@ -56,18 +56,17 @@ private SampleProfilerSession(SentryStopwatch stopwatch, EventPipeSession sessio
...
-        get { return Interlocked.CompareExchange(ref _throwOnNextStartupForTests, 1, 1) == 1; }
+        get { return _throwOnNextStartupForTests; }
         set
         {
             if (value)
-                Interlocked.CompareExchange(ref _throwOnNextStartupForTests, 1, 0);
+                _throwOnNextStartupForTests.CompareExchange(true, false);
             else
-                Interlocked.CompareExchange(ref _throwOnNextStartupForTests, 0, 1);
+                _throwOnNextStartupForTests.CompareExchange(false, true);
```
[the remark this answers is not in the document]
logiclrd: I think this warning is spurious because both branches call `.CompareExchange`. But, this could be simplified to a single call; the only reason it wasn't in the first place, I suspect, was that `bool` doesn't convert directly to `int`.

But at that point,

```
_throwOnNextStartupForTests.CompareExchange(value, !value);
```

How does that differ from just:

```
_throwOnNextStartupForTests = value;
```

If the `CompareExchange`:
- Only does the assignment if the value differs.
- Always sets the value to `value`.
- Has its return value ignored.

Then at that point it's literally just an assignment, right? 🙂

src/Sentry.Profiling/SampleProfilerSession.cs
```diff
@@ -56,19 +56,12 @@ private SampleProfilerSession(SentryStopwatch stopwatch, EventPipeSession sessio
...
-        get { return Interlocked.CompareExchange(ref _throwOnNextStartupForTests, 1, 1) == 1; }
-        set
-        {
-            if (value)
-                Interlocked.CompareExchange(ref _throwOnNextStartupForTests, 1, 0);
-            else
-                Interlocked.CompareExchange(ref _throwOnNextStartupForTests, 0, 1);
-        }
+        get { return _throwOnNextStartupForTests; }
+        set { _throwOnNextStartupForTests = value; }
```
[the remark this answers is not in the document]
jamescrosswell: If I'm following everything, line 63 here is ultimately just reading a volatile value (`InterlockedBoolean._value`), so that's OK. 

However on line 64, instead of using an Interlocked operation, we're now assigning a struct (via an implicit conversion operator) without any concurrency handling... so I think cursor might be right here. Just to be sure then, since these profiling tests have been a bit of a thorn in our side, can we do this?
```suggestion
        set { _throwOnNextStartupForTests.Exchange(value); }
```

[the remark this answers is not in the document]
logiclrd: A quick analysis: The structure is just a wrapper of a `bool`. The assignment there, under the hood, is effectively exactly the same as simply assigning to a local (`volatile`) `bool`. So, I don't _think_ there actually is any thread safety issue here.

I took a peek at the machine code generated for this to confirm it. Despite the hint to use aggressive inlining, it is:

```
00007FF7B5E69BAC  mov         eax,dword ptr [rbp+10h]  
00007FF7B5E69BAF  movzx       ecx,al  
00007FF7B5E69BB2  call        Sentry.Internal.InterlockedBoolean.op_Implicit(Boolean) (07FF7B609F960h)  
00007FF7B5E69BB8  mov         rcx, &_throwOnNextStartupForTests  (1C20023A908h)  
00007FF7B5E69BC2  mov         byte ptr [rcx],al
```

Line 1: Load the `bool` value passed into the `ThrowOnNextStartupForTests.set` accessor.
Line 2: Ignore all but the least significant byte and zero-extend that into ECX, which is the first parameter in the fastcall convention.
Line 3: Call the implicit conversion operator. Its return value (in EAX) will be a value of type `InterlockedBoolean`.
Line 4: Load the address of `_throwOnNextStartupForTests`. It's a constant in this context since it's `static`.
Line 5: The actual assignment: Write the `InterlockedBoolean` value, a single byte, into the memory assigned to the `_throwOnNextStartupForTests` static field.

The pattern would certainly be a valid cause for concern if the type were not fundamentally an atomic value under the hood, but it collapses down in this case.

But, this isn't a performance hot spot, so even if `.Exchange` cost a thousand times as much as a straight assignment (and it doesn't, I'm pretty sure? 🙂), it would be completely unnoticeable from that standpoint. There's not really a compelling argument to _not_ use `.Exchange`, even if it isn't strictly necessary.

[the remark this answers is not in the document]
jamescrosswell: I think this:
```csharp
_throwOnNextStartupForTests = value;
```
... is essentially calling [this](https://github.com/logiclrd/sentry-dotnet/blob/a446ec98dbb51fceb16c58b947f291a5092ca4ee/src/Sentry/Internal/InterlockedBoolean.cs#L32):
```csharp
public static implicit operator InterlockedBoolean(bool @this) => new InterlockedBoolean(@this);
```

So it takes a boolean value, passes that as the parameter to a constructor for `InterlockedBoolean`, then takes the result of that constructor and assigns it to `_throwOnNextStartupForTests` (which is a struct of type `InterlockedBoolean`). 

So I don't think it's assigning a volatile bool ... it's replacing the value of a struct with another struct.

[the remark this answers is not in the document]
logiclrd: Yep, but the value of the struct _is entirely a volatile bool_. Anything that boils down to accessing the `bool` value directly will follow `volatile` semantics, which also means that if something replaces the struct's value out from under it, it'll see it right away anyway because it's `volatile`. That's my understanding, in any case.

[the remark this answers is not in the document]
jamescrosswell: @Flash0ver you might know the answer here?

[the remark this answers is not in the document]
logiclrd: I'll throw in a commit to make it use `Interlocked.Exchange`. It's easy enough to remove later, as the case may be. :-)

src/Sentry.EntityFramework/SentryDatabaseLogging.cs
```diff
@@ -1,18 +1,20 @@
...
-    private static int Init;
+    private static InterlockedBoolean Init;
 
     internal static SentryCommandInterceptor? UseBreadcrumbs(
         IQueryLogger? queryLogger = null,
         bool initOnce = true,
         IDiagnosticLogger? diagnosticLogger = null)
     {
-        if (initOnce && Interlocked.Exchange(ref Init, 1) != 0)
+        if (initOnce && Init.Exchange(true))
```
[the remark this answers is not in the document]
logiclrd: ...huh??

I just made a quick test case as a sanity check. It enumerates the possible values for `initOnce` and `Init`, and then compares the results of the before & after expressions:

```
00 False False
01 False False
10 False False
11 True True
```

Column 1: `x` and `y` taking the place of `initOnce` and `Init`
Column 2: `y` as `int`: `x && Interlocked.Exchange(ref y, 1) != 0;`
Column 3: `y` as `bool`: `x && Interlocked.Exchange(ref y, true);`

And, the body of the `if` statement isn't where initialization happens. Rather, it enters the `if` in the situation where it _shouldn't_ repeat initialization. The condition will only evaluate to `true` when the caller asked for `initOnce` semantics _and_ a previous call already put `true` into `Init`.

I think the AI has the wrong end of the stick on this one. 🙂

Or am I missing something really obvious?? 🤔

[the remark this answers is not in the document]
logiclrd: I just got to wondering whether it knew something about the `InterlockedBoolean.Exchange` wrapper that I didn't. Automated testing of it notwithstanding, _maybe_ it didn't return the same value?? So I copied `InterlockedBoolean.cs` into my test app and added a column for `y` as `InterlockedBoolean`: `x && y.Exchange(true)`.

```
00 False False False
01 False False False
10 False False False
11 True True True
```

¯\\\_(ツ)\_/¯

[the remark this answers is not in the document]
Flash0ver: > The original condition returned true if already initialized, but the new Init.Exchange(true) returns the opposite boolean value needed for the condition.

I believe you're right.
`Init.Exchange(true)` returns the original value, which is `false` the first time, so the early-out is not entered,
but subsequent invocations are returning the now original value of `true` entering the "was already executed"-path.

Sounds correct to me, too, with the code as-is.

test/Sentry.Tests/Internals/InterlockedBooleanTests.cs
Flash0ver: sounds fair

src/Sentry.Azure.Functions.Worker/.editorconfig
```diff
@@ -2,4 +2,6 @@
 
 # Reason: Azure Functions worker doesn't set SynchronizationContext but Durable Functions does and required affinity.
 # (https://github.com/Azure/azure-functions-dotnet-worker/issues/1520)
-dotnet_diagnostic.CA2007.severity = none
\ No newline at end of file
+dotnet_diagnostic.CA2007.severity = none
+
+dotnet_style_namespace_declaration_with_usings_placement = prefer_outside_namespace
```
Flash0ver: suggestion: move to the "root" `.editorconfig`, as this style should apply to the entire codebase

logiclrd: Oh, right -- I didn't even notice at the time that there were multiple levels of `.editorconfig`. 😁

logiclrd: Huh, this is bizarre. I copy/pasted that option name from a warning message, but I don't remember where. I notice explanatory comments in the root `.editorconfig` file, and so I wanted to see if this setting applied to VB.NET `Import` declarations as well, for completeness in the comment (since the preceding comment explicitly calls out `Import` declarations). I did a web search for it _and I can't find any evidence that it actually exists_. Where on earth did I copy/paste it from?! I'm scratching my head.

In any case, the code style warning `IDE0065` is documented as using an `.editorconfig` option named `csharp_using_directive_placement`. I will switch it to that as part of moving it to the root `.editorconfig`.

src/Sentry/Internal/InterlockedBoolean.cs
logiclrd: Okay, I took a peek at what the JIT is doing.

In `Debug` mode, it's doing exactly what's written, no special efficiency.

Source code:

```
        TBool localNewValue = newValue ? True : False;
```

Compiled code:
```
// Load the parameter value as a 32-bit value, but then explicitly zero-extend the least significant byte
00007FFE80D61C39  mov         eax,dword ptr [rbp+68h]  
00007FFE80D61C3C  movzx       eax,al  
// EAX is now either 0 or 1 based on the bool value of newValue. Branch based on whether it is zero or not.
00007FFE80D61C3F  test        eax,eax  
00007FFE80D61C41  jnz         Sentry.Internal.InterlockedBoolean.Exchange(Boolean)+03Ah (07FFE80D61C4Ah)  
// - It is zero, write the value of False, which is also bool false, into temporary storage
00007FFE80D61C43  xor         eax,eax  
00007FFE80D61C45  mov         dword ptr [rbp+30h],eax  
00007FFE80D61C48  jmp         Sentry.Internal.InterlockedBoolean.Exchange(Boolean)+041h (07FFE80D61C51h)  
// - It is non-zero, write the value of True, which is bool true, into temporary storage
00007FFE80D61C4A  mov         dword ptr [rbp+30h],1  
// Now, just as when we read the parameter value, read the temporary storage as a 32-bit value, but then explicitly zero-extend the least significant byte (totally unnecessary 🙂)
00007FFE80D61C51  mov         eax,dword ptr [rbp+30h]  
00007FFE80D61C54  movzx       eax,al  
// Write the result to localNewValue
00007FFE80D61C57  mov         dword ptr [rbp+3Ch],eax  
```

So that's kind of disappointing but also not a huge surprise. 🙂 

Interestingly, the return value isn't quite so horrifically inefficient:

Source code, in which localReturnValue is a bool because I'm running in .NET 9:

```
        return (localReturnValue != False);
```

Compiled code:

```
// Load the value of localReturnValue
00007FFE80D61C79  mov         eax,dword ptr [rbp+38h]  
// Write it to temporary storage
00007FFE80D61C7C  mov         dword ptr [rbp+34h],eax  
00007FFE80D61C7F  nop  
// Load the temporary storage into EAX in preparation for RET (totally unnecessary 🙂 it's already there)
00007FFE80D61C80  mov         eax,dword ptr [rbp+34h]  
```

So, in the return code, it recognized that `boolValue != false` is just `boolValue`, but in the lead-in, it didn't recognize that `boolValue ? true : false` is just `boolValue`.

Anyway, that's the `Debug` profile, no optimization. Now for `Release`.

Source code:

```
        TBool localNewValue = newValue ? True : False;
```

Compiled code:

```
// With the fastcall calling convention, the this argument is in ECX and the parameter newValue is in EDX. Stash in temporary storage.
00007FFE6B3C7B53  mov         dword ptr [rbp+18h],edx 
// Similar to the Debug profile, load the value as a 32-bit word and then zero-extend. 
00007FFE6B3C7B56  mov         eax,dword ptr [rbp+18h]  
00007FFE6B3C7B59  movzx       eax,al  
// Test if it is zero or not.
00007FFE6B3C7B5C  test        eax,eax  
// No branching, ma! Set AL to 1 or 0 based on whether the bool value was true or false.
00007FFE6B3C7B5E  setnz       al  
// Zero-extend, just in case (this clears the top bytes of EAX), then store it in localNewValue
00007FFE6B3C7B61  movzx       eax,al  
00007FFE6B3C7B64  mov         dword ptr [rbp-4],eax  
```

As for the return, well, that gets interesting :-) The `Interlocked.Exchange` actually got treated as an intrinsic, and the result, which was computed into ECX, was then copied directly to EAX. The metadata didn't even attribute a byte range to the `return` line at all.

Source code:

```
        TBool localReturnValue = Interlocked.Exchange(ref _value, localNewValue);

        return (localReturnValue != False);
```

Compiled code:

```
// Load the memory address of _value
00007FFE6B3C7B67  mov         rax,qword ptr [rbp+10h]  
// This opcode is weird, possibly a bug in the JIT? It compares the byte at this memory address _with the least significant byte of the memory address_
00007FFE6B3C7B6B  cmp         byte ptr [rax],al  
// Load the memory address of _value again (unnecessary)
00007FFE6B3C7B6D  mov         rax,qword ptr [rbp+10h]  
// Load the value of localNewValue into ECX
00007FFE6B3C7B71  mov         ecx,dword ptr [rbp-4]  
// Interlocked.Exchange, a single instruction :-)
00007FFE6B3C7B74  xchg        cl,byte ptr [rax]  
// XCHG has swapped the byte in _value with the CL register. Now we zero-extend, turning it into a 32-bit representation of the bool value.
00007FFE6B3C7B76  movzx       ecx,cl  
// Store in EAX for return. All done.
00007FFE6B3C7B79  mov         eax,ecx  
```

So how does that compare against just a direct call to `Interlocked.Exchange`, if the target framework supports it?

Source code:

```
        return Interlocked.Exchange(ref _value, newValue);
```

Compiled code:

```
// Move the fastcall parameter from EDX into temporary storage
00007FFF5D766C78  mov         dword ptr [rbp+18h],edx  
// Load the memory address of _value
00007FFF5D766C7B  mov         rax,qword ptr [rbp+10h]  
// This opcode is weird, possibly a bug in the JIT? It compares the byte at this memory address _with the least significant byte of the memory address_
00007FFF5D766C7F  cmp         byte ptr [rax],al  
// Load the memory address of _value again (unnecessary)
00007FFF5D766C81  mov         rax,qword ptr [rbp+10h]  
// Load the value of localNewValue into ECX
00007FFF5D766C85  mov         ecx,dword ptr [rbp+18h]  
00007FFF5D766C88  movzx       ecx,cl  
// Interlocked.Exchange, a single instruction :-)
00007FFF5D766C8B  xchg        cl,byte ptr [rax]  
// XCHG has swapped the byte in _value with the CL register. Now we zero-extend, turning it into a 32-bit representation of the bool value.
00007FFF5D766C8D  movzx       ecx,cl  
// Store in EAX for return. All done.
00007FFF5D766C90  mov         eax,ecx  
```

So basically identical to the tail end of the generic implementation.

It's not perfectly-optimized, but it's pretty good.

What does this mean for performance?

It means the difference between being able to do `myInterlockedBoolean.Exchange(value)` 254 million times per second and only being able to do 251 million calls per second.

It appears that that extra lead-up is almost entirely inconsequential. I made a program to benchmark it, and very consistently, the "slow" path is only 1.2% slower than the "fast" path.

I'll polish up the benchmark code and throw it up into a repo somewhere, because maybe I made a silly mistake somewhere. 🙂

logiclrd: Even in `Debug` mode, the "slow" version is only about 11% slower than the "fast" version, ~158 million calls per second compared to ~140 million calls per second.

Here's the benchmark:

https://github.com/logiclrd/InterlockedExchangePerformanceTest

PR #4611, opened 2025-10-09
ScopeExtensions.Populate is now internal

### LINKED ISSUE #4352: ScopeExtensions.Populate should not be public
`ScopeExtensions.Populate` is currently public.

https://github.com/getsentry/sentry-dotnet/blob/4a764c929ed7a03557181fb95382174183b3ee4f/src/Sentry.AspNetCore/ScopeExtensions.cs#L25

We call this from the SentryMiddleware:
https://github.com/getsentry/sentry-dotnet/blob/6e50eaa87b3aa24a921824c4494566b01f0640f0/src/Sentry.AspNetCore/SentryMiddleware.cs#L258

However SDK users shouldn't need to call this so it should be internal.

Breaking changes so let's do it in the next major release.

jamescrosswell:
Resolves #4352:
- #4352

### DIFF
--- a/test/Sentry.AspNetCore.Tests/ApiApprovalTests.Run.DotNet10_0.verified.txt
+++ b/test/Sentry.AspNetCore.Tests/ApiApprovalTests.Run.DotNet10_0.verified.txt
@@ -39,7 +39,6 @@ namespace Sentry.AspNetCore
         public static string? TryGetHttpPath(this Sentry.TransactionSamplingContext samplingContext) { }
         public static string? TryGetHttpRoute(this Sentry.TransactionSamplingContext samplingContext) { }
     }
-    public static class ScopeExtensions { }
     [Microsoft.Extensions.Logging.ProviderAlias("Sentry")]
     public class SentryAspNetCoreLoggerProvider : Sentry.Extensions.Logging.SentryLoggerProvider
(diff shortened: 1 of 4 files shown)


test/Sentry.AspNetCore.Tests/ApiApprovalTests.Run.DotNet10_0.verified.txt
```diff
@@ -39,11 +39,7 @@ namespace Sentry.AspNetCore
         public static string? TryGetHttpPath(this Sentry.TransactionSamplingContext samplingContext) { }
         public static string? TryGetHttpRoute(this Sentry.TransactionSamplingContext samplingContext) { }
     }
-    public static class ScopeExtensions
-    {
-        public static void Populate(this Sentry.Scope scope, System.Diagnostics.Activity activity) { }
-        public static void Populate(this Sentry.Scope scope, Microsoft.AspNetCore.Http.HttpContext context, Sentry.AspNetCore.SentryAspNetCoreOptions options) { }
-    }
+    public static class ScopeExtensions { }
```
Flash0ver: suggestion: make the entire type `internal`, so that we don't have an "empty" type hanging off of the public surface area

PR #4613, opened 2025-10-09
CaptureFeedback now returns a SentryId

### LINKED ISSUE #4319: Add return value to `CaptureFeedback`
## Problem Statement

`CaptureFeedback` is a user facing (not just our users, but actually our users' user) method, where there is no way to validate successful capture right now.

## Context
https://github.com/getsentry/sentry-dotnet/blob/dc2b3261c8e475a99f8891bd4e26e4d65ba81e9f/src/Sentry/SentrySdk.cs#L536 
`CaptureFeedback` returns `void` right now. There are multiple conditions why the SDK might fail to send the feedback
https://github.com/getsentry/sentry-dotnet/blob/dc2b3261c8e475a99f8891bd4e26e4d65ba81e9f/src/Sentry/Internal/Hub.cs#L559-L564
https://github.com/getsentry/sentry-dotnet/blob/dc2b3261c8e475a99f8891bd4e26e4d65ba81e9f/src/Sentry/SentryClient.cs#L87-L91

The feedback could be dropped and without reading the logs, users will never know that it happened.
Capturing the feedback ends with a call to `CaptureEnvelope` which does return a `bool` but that gets swallowed right now.

https://github.com/getsentry/sentry-dotnet/blob/dc2b3261c8e475a99f8891bd4e26e4d65ba81e9f/src/Sentry/SentryClient.cs#L114-L116

## Proposal

Similar to how it is done in `CaptureEvent` calling `CaptureFeedback` should return the event ID.
https://github.com/getsentry/sentry-dotnet/blob/dc2b3261c8e475a99f8891bd4e26e4d65ba81e9f/src/Sentry/SentryClient.cs#L73-L81

This would allow
1. Users to validate whether the event/feedback was actually sent
2. Users to display the captured event ID to their users

jamescrosswell:
`SentrySdk.CaptureFeedback` now returns a `SentryId` that will be:
- either the id of the newly captured feedback (if the call succeeds)
- or `SentryId.Empty` and an error code (if the call fails)

An additional overload is also available with an extra `out CaptureFeedbackResult` that can be used to determine whether the call succeeded or what the specific reason for failure was otherwise.

Resolves #4319 
- #4319

### DIFF
--- a/test/Sentry.Tests/ApiApprovalTests.Run.Net4_8.verified.txt
+++ b/test/Sentry.Tests/ApiApprovalTests.Run.Net4_8.verified.txt
@@ -45,24 +45,12 @@ namespace Sentry
         public ByteAttachmentContent(byte[] bytes) { }
         public System.IO.Stream GetStream() { }
     }
-    public enum CaptureFeedbackErrorReason
+    public enum CaptureFeedbackResult
     {
-        UnknownError = 0,
-        DisabledHub = 1,
-        EmptyMessage = 2,
-    }
-    public sealed class CaptureFeedbackResult
-    {
-        public CaptureFeedbackResult(Sentry.CaptureFeedbackErrorReason errorReason) { }
-        public CaptureFeedbackResult(Sentry.SentryId eventId) { }
-        public Sentry.CaptureFeedbackErrorReason? ErrorReason { get; }
-        public Sentry.SentryId EventId { get; }
-        public bool Succeeded { get; }
-        public bool Equals(Sentry.CaptureFeedbackResult? other) { }
-        public override bool Equals(object? obj) { }
-        public override int GetHashCode() { }
-        public static bool operator !=(Sentry.CaptureFeedbackResult? left, Sentry.CaptureFeedbackResult? right) { }
-        public static bool operator ==(Sentry.CaptureFeedbackResult? left, Sentry.CaptureFeedbackResult? right) { }
+        Success = 0,
+        UnknownError = 1,
+        DisabledHub = 2,
+        EmptyMessage = 3,
     }
     public enum CheckInStatus
     {
@@ -146,6 +134,7 @@ namespace Sentry
         public static void AddBreadcrumb(this Sentry.IHub hub, string message, string? category = null, string? type = null, System.Collections.Generic.IDictionary<string, string>? data = null, Sentry.BreadcrumbLevel level = 0) { }
         public static void AddBreadcrumb(this Sentry.IHub hub, Sentry.Infrastructure.ISystemClock? clock, string message, string? category = null, string? type = null, System.Collections.Generic.IDictionary<string, string>? data = null, Sentry.BreadcrumbLevel level = 0) { }
         public static Sentry.SentryId CaptureException(this Sentry.IHub hub, System.Exception ex, System.Action<Sentry.Scope> configureScope) { }
+        public static Sentry.SentryId CaptureFeedback(this Sentry.IHub hub, Sentry.SentryFeedback feedback, System.Action<Sentry.Scope> configureScope, Sentry.SentryHint? hint = null) { }
         public static Sentry.SentryId CaptureMessage(this Sentry.IHub hub, string message, System.Action<Sentry.Scope> configureScope, Sentry.SentryLevel level = 1) { }
         public static void LockScope(this Sentry.IHub hub) { }
         public static System.IDisposable PushAndLockScope(this Sentry.IHub hub) { }
@@ -198,7 +187,7 @@ namespace Sentry
         void BindException(System.Exception exception, Sentry.ISpan span);
         Sentry.SentryId CaptureEvent(Sentry.SentryEvent evt, System.Action<Sentry.Scope> configureScope);
         Sentry.SentryId CaptureEvent(Sentry.SentryEvent evt, Sentry.SentryHint? hint, System.Action<Sentry.Scope> configureScope);
-        Sentry.CaptureFeedbackResult CaptureFeedback(Sentry.SentryFeedback feedback, System.Action<Sentry.Scope> configureScope, Sentry.SentryHint? hint = null);
+        Sentry.SentryId CaptureFeedback(Sentry.SentryFeedback feedback, out Sentry.CaptureFeedbackResult result, System.Action<Sentry.Scope> configureScope, Sentry.SentryHint? hint = null);
         Sentry.TransactionContext ContinueTrace(Sentry.SentryTraceHeader? traceHeader, Sentry.BaggageHeader? baggageHeader, string? name = null, string? operation = null);
         Sentry.TransactionContext ContinueTrace(string? traceHeader, string? baggageHeader, string? name = null, string? operation = null);
         void EndSession(Sentry.SessionEndStatus status = 0);
@@ -225,7 +214,7 @@ namespace Sentry
         Sentry.SentryId CaptureCheckIn(string monitorSlug, Sentry.CheckInStatus status, Sentry.SentryId? sentryId = default, System.TimeSpan? duration = default, Sentry.Scope? scope = null, System.Action<Sentry.SentryMonitorOptions>? configureMonitorOptions = null);
         bool CaptureEnvelope(Sentry.Protocol.Envelopes.Envelope envelope);
         Sentry.SentryId CaptureEvent(Sentry.SentryEvent evt, Sentry.Scope? scope = null, Sentry.SentryHint? hint = null);
-        Sentry.CaptureFeedbackResult CaptureFeedback(Sentry.SentryFeedback feedback, Sentry.Scope? scope = null, Sentry.SentryHint? hint = null);
+        Sentry.SentryId CaptureFeedback(Sentry.SentryFeedback feedback, out Sentry.CaptureFeedbackResult result, Sentry.Scope? scope = null, Sentry.SentryHint? hint = null);
         void CaptureSession(Sentry.SessionUpdate sessionUpdate);
         void CaptureTransaction(Sentry.SentryTransaction transaction);
         void CaptureTransaction(Sentry.SentryTransaction transaction, Sentry.Scope? scope, Sentry.SentryHint? hint);
@@ -453,7 +442,7 @@ namespace Sentry
         public Sentry.SentryId CaptureCheckIn(string monitorSlug, Sentry.CheckInStatus status, Sentry.SentryId? sentryId = default, System.TimeSpan? duration = default, Sentry.Scope? scope = null, System.Action<Sentry.SentryMonitorOptions>? configureMonitorOptions = null) { }
         public bool CaptureEnvelope(Sentry.Protocol.Envelopes.Envelope envelope) { }
         public Sentry.SentryId CaptureEvent(Sentry.SentryEvent? @event, Sentry.Scope? scope = null, Sentry.SentryHint? hint = null) { }
-        public Sentry.CaptureFeedbackResult CaptureFeedback(Sentry.SentryFeedback feedback, Sentry.Scope? scope = null, Sentry.SentryHint? hint = null) { }
+        public Sentry.SentryId CaptureFeedback(Sentry.SentryFeedback feedback, out Sentry.CaptureFeedbackResult result, Sentry.Scope? scope = null, Sentry.SentryHint? hint = null) { }
         public void CaptureSession(Sentry.SessionUpdate sessionUpdate) { }
         public void CaptureTransaction(Sentry.SentryTransaction transaction) { }
         public void CaptureTransaction(Sentry.SentryTransaction transaction, Sentry.Scope? scope, Sentry.SentryHint? hint) { }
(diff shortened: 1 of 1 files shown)


src/Sentry/ISentryClient.cs
```diff
@@ -34,7 +34,8 @@ public interface ISentryClient
     /// <param name="feedback">The feedback to send to Sentry.</param>
     /// <param name="scope">An optional scope to be applied to the event.</param>
     /// <param name="hint">An optional hint providing high level context for the source of the event</param>
-    public void CaptureFeedback(SentryFeedback feedback, Scope? scope = null, SentryHint? hint = null);
+    /// <returns>The SentryId of the captured feedback, if successful. SentryId.Empty if feedback capture fails.</returns>
+    public SentryId CaptureFeedback(SentryFeedback feedback, Scope? scope = null, SentryHint? hint = null);
```
jamescrosswell: Arguably this could be called `TryCaptureFeedback` now but I don't think that's consistent with the other SDKs (or the `CaptureEvent` method)... so for consistency I've left it as is.

test/Sentry.Tests/SentryClientTests.cs
```diff
@@ -912,7 +912,8 @@ public void CaptureFeedback_DisposedClient_DoesNotThrow()
         sut.Dispose();
 
         // Act / Assert
-        sut.CaptureFeedback(feedback);
+        var result = sut.CaptureFeedback(feedback);
+        result.Should().NotBe(SentryId.Empty);
```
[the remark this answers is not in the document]
jamescrosswell: Hm, I think I originally copied/tweaked that unit test from here:
https://github.com/getsentry/sentry-dotnet/blob/1a761774fd750f611b788b0b859319dae3db887e/test/Sentry.Tests/SentryClientTests.cs#L997-L1005

But it looks broken. It checks an exception is not thrown when the Client has been disposed of... but it does so using a Substitute for the BackgroundWorker. The Substitute does not throw... however the real BackgroundWorker does:
https://github.com/getsentry/sentry-dotnet/blob/9e2f5f223c8651891ff9760abb9fd16163f3c820/src/Sentry/Internal/BackgroundWorker.cs#L62-L67

I may as well fix that up in this PR (even though it's not caused by this PR).

test/Sentry.Tests/SentryClientTests.cs
```diff
@@ -1576,12 +1581,12 @@ public void Dispose_Worker_FlushCalled()
...
 
     [Fact]
-    public void Dispose_MultipleCalls_WorkerFlushedTwice()
+    public void Dispose_MultipleCalls_WorkerFlushedOnce()
     {
         var sut = _fixture.GetSut();
         sut.Dispose();
         sut.Dispose();
-        _fixture.BackgroundWorker?.Received(2).FlushAsync(_fixture.SentryOptions.ShutdownTimeout);
+        _fixture.BackgroundWorker?.Received(1).FlushAsync(_fixture.SentryOptions.ShutdownTimeout);
```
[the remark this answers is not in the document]
jamescrosswell: Not true. I changed the Dispose method implementation to ensure the Disposal logic is only ever called once.

src/Sentry/ISentryClient.cs
Flash0ver: I was also wondering about the other `SentryId`-returning `Capture*`-methods 🤔

Both other `SentryId`-returning `Capture*`-methods
- `SentryId CaptureEvent`
- `SentryId CaptureCheckIn`

do return `SentryId.Empty` when unsuccessful.

If I get that right, then `bool Try*`-methods should communicate the "main-exceptional" flow when returning `false`, not every reason for completing unsuccessfully.
E.g. `Dictionary<TKey, TValue>.TryAdd` only returns `false` when "An item with the same key has already been added.". But it still throws if `null` is passed as a key.

What would be the "main-error" that `TryCaptureFeedback` would communicate?
- Hub not enabled
- Message is null or empty
- Envelope not enqueued

`TryCaptureEvent` would also have multiple failure cases
- Event is null
- Event dropped via Exception-Filters
- Event dropped via Event-Processor
- Event dropped via `SetBeforeSend`
- Event dropped by sampling
- Event not enqueued

Hmm ... we could return something like a "CaptureEventStatus" `struct` or `enum`.
Similar to BCL methods returning [OperationStatus](https://learn.microsoft.com/dotnet/api/system.buffers.operationstatus) when it comes to buffers.

Or am I dramatically overthinking this?

Flash0ver: Also ... I guess we're a bit inconsistent with `void` vs `bool`, too:
- `void CaptureTransaction`
- `void CaptureSession`

Could (or should) these be `bool`-returning as well?

src/Sentry/SentryClient.cs
```diff
@@ -438,6 +439,12 @@ private SentryId DoSendEvent(SentryEvent @event, SentryHint? hint, Scope? scope)
     /// <inheritdoc cref="ISentryClient.CaptureEnvelope"/>
     public bool CaptureEnvelope(Envelope envelope)
     {
+        if (_isDisposed == 1)
+        {
+            _options.LogWarning("Enqueue envelope failed: disposed client");
```
bitsandfoxes:
```suggestion
            _options.LogWarning("Enqueue envelope failed. Client is disposed.");
```

src/Sentry/SentryClient.cs
```diff
@@ -456,6 +463,11 @@ public bool CaptureEnvelope(Envelope envelope)
     /// </summary>
     public void Dispose()
     {
+        if (Interlocked.Exchange(ref _isDisposed, 1) == 1)
+        {
+            return;
+        }
+
```
bitsandfoxes: That change seems unrelated to the CaptureFeedback, right?

src/Sentry/ISentryClient.cs
jamescrosswell: > Could (or should) these be bool-returning as well?

I think @bitsandfoxes wanted a SentryId on this, since it's not just SDK user facing but SDK user's user facing... so the SDK user could actually want to report success/failure of this method to the app user.

That's probably not the case for `CaptureTransaction` or `CaptureSession`. The SDK user is highly unlikely to do anything with any hypothetical return value for those functions. They're really just best effort. 

> What would be the "main-error" that TryCaptureFeedback would communicate?

This is probably the main one:
https://github.com/getsentry/sentry-dotnet/blob/0cda3df3ec9e03ebc72c40f21b6cb096f05487eb/src/Sentry/SentryClient.cs#L91-L95

`DisabledHub` is possible, if an empty string was used for the DSN when initialising the SDK. 

I think failure to enqueue the envelope would only happen if we'd disposed of the SentryClient (which would only happen when shutting down the app)... so a vanishingly small chance of that happening when the app is capturing user feedback (and no chance to do anything with that error code if/when it did happen).

bitsandfoxes: I'd argue to **not** name it `Try*` but instead that the static API is mostly intended as fire & forget. The return values are not intended to be used to control the flow or have the option to recover from a failure.

Consider the following
1. User calls `CaptureEvent`
2. Capture "fails" due to: SDK not initialized, dropped due to filtering, beforeSend, processor?
3. What do we expect the user to do now?

If they need the ID for some reason they should have it, but prepending the capture methods with a `Try*` implies some sort of recovery mechanism to me that does not exist.

bitsandfoxes:
This looks great from the SDK for Unity POV.

src/Sentry/SentryClient.cs
jamescrosswell: Yes, strictly speaking... but cursor picked it up in this PR, it seemed like a quick/easy fix, and I've already got about 10 PRs open:
- https://github.com/getsentry/sentry-dotnet/pull/4613/files/f58417ba8f35c4c911a7b0634688f5439e29102b#r2415890162

I can pull it out to a separate PR if we really want to.

src/Sentry/ISentryClient.cs
jamescrosswell: @bitsandfoxes agree re the method name. 

Is there any value in changing the result type from SentryId to `CaptureFeedbackResult` (to be able to distinguish failure reasons) or is this overkill do you think? See [this comment](https://github.com/getsentry/sentry-dotnet/pull/4613#discussion_r2419311842) above - I think there are only really two possible reasons for failure.

bitsandfoxes: It would be nice but I don't see this as a requirement. We've got that pattern all over the rest of the SDK. I think it's fine if we document it. I can do that together with the bump in the Unity SDK.

src/Sentry/SentryClient.cs
```diff
@@ -20,6 +20,7 @@ public class SentryClient : ISentryClient, IDisposable
     private readonly ISessionManager _sessionManager;
     private readonly RandomValuesFactory _randomValuesFactory;
     private readonly Enricher _enricher;
+    private volatile int _isDisposed = 0;
```
Flash0ver: question: Shall we use the `InterlockedBoolean` here too?

src/Sentry/ISentryClient.cs
jamescrosswell: Since we're making breaking changes here, we may as well add a most sophisticated result type.

test/Sentry.Tests/ApiApprovalTests.Run.DotNet10_0.verified.txt
```diff
@@ -191,7 +207,7 @@ namespace Sentry
         void BindException(System.Exception exception, Sentry.ISpan span);
         Sentry.SentryId CaptureEvent(Sentry.SentryEvent evt, System.Action<Sentry.Scope> configureScope);
         Sentry.SentryId CaptureEvent(Sentry.SentryEvent evt, Sentry.SentryHint? hint, System.Action<Sentry.Scope> configureScope);
```
Flash0ver, line 201: question: Is it worth considering having more returns of `SentryId` ... potentially in a follow-up issue/PR?

- `CaptureTransaction`
- `CaptureSession`

src/Sentry/CaptureFeedbackResult.cs
```diff
@@ -0,0 +1,95 @@
...
+        ErrorReason = errorReason;
+    }
+
+    /// <summary>
+    /// The Id of the captured feedback, if successful. <see cref="SentryId.Empty"/> if feedback capture fails.
+    /// </summary>
+    public SentryId EventId;
+
+    /// <inheritdoc cref="CaptureFeedbackErrorReason"/>
+    public CaptureFeedbackErrorReason? ErrorReason;
```
Flash0ver: thought: `Nullable<T>` vs non-nullable

I am a bit hesitant about a `Nullable<enum>`.
I'm thinking about having an additional `0` value in the enum, like `None`.
Or alternatively something like `Done` or `Succeeded` and calling the `enum` instead `CaptureFeedbackStatus`.
Similar to
- https://learn.microsoft.com/dotnet/api/system.buffers.operationstatus
- https://learn.microsoft.com/dotnet/api/system.threading.tasks.taskstatus

I do prefer the latter ... but personal preference ... I'm convincible 😉

src/Sentry/CaptureFeedbackResult.cs
```diff
@@ -0,0 +1,95 @@
...
+    public CaptureFeedbackResult(CaptureFeedbackErrorReason errorReason)
+    {
+        EventId = SentryId.Empty;
+        ErrorReason = errorReason;
+    }
+
+    /// <summary>
+    /// The Id of the captured feedback, if successful. <see cref="SentryId.Empty"/> if feedback capture fails.
+    /// </summary>
+    public SentryId EventId;
```
Flash0ver: issue: public fields

I'd prefer readonly properties here.

src/Sentry/CaptureFeedbackResult.cs
```diff
@@ -0,0 +1,95 @@
+namespace Sentry;
+
+/// <summary>
+/// The result type of the <see cref="ISentryClient.CaptureFeedback"/> method
+/// </summary>
+public struct CaptureFeedbackResult
```
Flash0ver: question: `readonly`

Can we make this `struct` readonly, as the result should never change?

Flash0ver: design: `default`

I'm wondering if this is a "valid" struct.

`default` (or `new()`) would be
- EventId == SentryId-Empty
- ErrorReason == null (or none)

Which sounds like an oxymoron.

---

An alternative design would be to not have this new struct, and have the `CaptureFeedback` methods:
- return: `CaptureFeedbackErrorReason` (or `CaptureFeedbackStatus`
- out-parameter: `out SentryId eventId`

But now that I write that down ... that's not too appealing either ... but perhaps this brainstorms something.

---

I may also totally be overthinking this, and we need to make sure to never return `default(CaptureFeedbackResult)`, and user code creating instances of `CaptureFeedbackResult` isn't really a useful scenario.

test/Sentry.Tests/ApiApprovalTests.Run.DotNet10_0.verified.txt
jamescrosswell: I don't think so. Those methods aren't user facing (they're SDK user facing). This method is an exception in that there will be some UI around it in the app and so some feedback is required for the actual user (not just the SDK user).  

Certainly nobody has ever asked for this...

src/Sentry/CaptureFeedbackResult.cs
jamescrosswell: I tried that but it means we have to do weird stuff like this in one of the constructors:
```
    public CaptureFeedbackResult(CaptureFeedbackResultStatus resultStatus)
    {
        if (resultStatus == CaptureFeedbackResultStatus.Success)
        {
            throw new ArgumentException("Use alternate constructor to indicate success.", nameof(resultStatus));
        }

        EventId = SentryId.Empty;
        ResultStatus = resultStatus;
    }
```

I think that's just as messy personally. I don't think the current design is any better or worse than the alternatives then.

> [!NOTE]
> This is one good use case for a Nominal Discriminating Union... but we don't have those yet.

src/Sentry/CaptureFeedbackResult.cs
jamescrosswell: Changed to a class, so that the constructors can't be sidestepped with `default`. That way we can guarantee member consistency.

src/Sentry/SentryClient.cs
jamescrosswell: N/A - removed.

src/Sentry/Internal/Hub.cs
```diff
@@ -842,7 +845,8 @@ public void Dispose()
         }
         //Don't dispose of ScopeManager since we want dangling transactions to still be able to access tags.
 
-        _backpressureMonitor?.Dispose();
+        // Don't dispose of _backpressureMonitor since we want the client to continue to process envelopes without
+        // throwing an ObjectDisposedException.
```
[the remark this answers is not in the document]
jamescrosswell: Intentional

src/Sentry/CaptureFeedbackResult.cs
jamescrosswell: Another possibility that would be much more disruptive to SDK users would be that we switch back to returning a `SentryId?` from the methods in `SentrySdk`, `IHub` etc. but add an `out CaptureFeedbackResult` to those same methods... then we add a `CaptureFeedbackResult.Succeeded` enum.

jamescrosswell: OK I've gone with an overload... so existing users won't be troubled. If people want the result code though, there's an overload that provides it as an out parameter.

PR #4619, opened 2025-10-10
Removed obsolete v6 APIs

### LINKED ISSUE #4608: Remove `Obsolete` APIs for `6.0.0`
### Description

During `5.*` we have marked APIs with the `ObsoleteAttribute`.
Let's review them for v6 and remove what's no longer needed.
May also need a review/update in docs.

jamescrosswell:
Resolves #4608 
- #4608 

## Not addressed in this PR

There are a couple of APIs left that are still marked as obsolete, which probably merit some discussion. I've created separate issues for these, so as not to block the work that has already been done in this PR. 

- https://github.com/getsentry/sentry-dotnet/issues/4640
- https://github.com/getsentry/sentry-dotnet/issues/4641
- https://github.com/getsentry/sentry-dotnet/issues/4642
- https://github.com/getsentry/sentry-dotnet/issues/4643


CHANGELOG.md
```diff
@@ -6,6 +6,9 @@
 
 - SentryOptions.IsEnvironmentUser now defaults to false on MAUI. The means the User.Name will no longer be set, by default, to the name of the device ([#4606](https://github.com/getsentry/sentry-dotnet/pull/4606))
 - Remove unnecessary files from SentryCocoaFramework before packing ([#4602](https://github.com/getsentry/sentry-dotnet/pull/4602))
+- Removed obsolete v6 APIs ([#4619](https://github.com/getsentry/sentry-dotnet/pull/4619))
```
Flash0ver: suggestion:

```suggestion
- Removed obsolete APIs ([#4619](https://github.com/getsentry/sentry-dotnet/pull/4619))
```

To me, the `v6` in the entry doesn't read well ... we haven't added any stable "v6" APIs yet.

Flash0ver: ### re: Extra
- `SentrySpan`
- `SentryTransaction`
- `TransactionTracer`

I do agree on the argument about "consolidating all the Extra, Data, Tags, and such into Attributes".

... we may have an additional problem:

Does this actually also include `Sentry.IHasExtra` and `Sentry.HasExtraExtensions`.
If it does, then we can't really remove the `Extra` indeed, because we haven't marked the `IHasExtra` interface nor its members, neither have we marked the `HasExtraExtensions` as obsolete. And also, we have missed the types `Sentry.SpanTracer` and `Sentry.Scope` as well:

```CSharp
hub.ConfigureScope(scope =>
{
  scope.SetExtra("Extra!", "Some extra information"); // does not warn
  _ = scope.Extra; // does not warn
});

ITransactionTracer transaction = SentrySdk.StartTransaction("", "");
transaction.SetExtra("", ""); // does not warn
transaction.SetExtras([]); // does not warn
_ = transaction.Extra; // does not warn

ISpan span = SentrySdk.StartSpan("", "");
span.SetExtra("", ""); // does not warn
span.SetExtras([]); // does not warn
_ = span.Extra; // does not warn

_ = (SentrySdk.StartTransaction("", "") as TransactionTracer)!.Extra; // CS0618
_ = (SentrySdk.StartSpan("", "") as SpanTracer)!.Extra; // does not warn
```

So we have been incomplete with marking the feature "Extra" as _Obsolete_.
Only if the concrete types `SentryTransaction`, or `SentrySpan` or `TransactionTracer` are used.
But if any interface or other concrete types are used, "Extra" is not marked as _Obsolete_.

Perhaps we should, for this release, make sure, that we mark "Extra" as _Obsolete_ consistently, including
- `IHasExtra` interface and members
- `HasExtraExtensions` type and members
- `Scope`
- `SpanTracer`
- ...

Because currently, we only produce on some usages of "Extra" an  _Obsolete_ warning, but not on all usages of "Extra", especially on the interfaces which are usually the types that we return.

Flash0ver: ### re: GetInternalSentryOptions

Agreed. `Obsolete` here is a misuse.
Although I do like something like "InternalApi" more than "Obsolete", this reminds me of [Pubternal](https://learn.microsoft.com/dotnet/core/compatibility/aspnet-core/5.0/localization-pubternal-apis-removed), which has been banned from the .NET libraries.

Can't we actually make this `internal`?
...
No, we cannot ... due to the usage in `Sentry.Hangfire` ... we removed the Strong-Name: #4099

I would love a solution to no longer have this extension method as "Pubternal", but properly "Internal".
But we'd probably want to do this in a separate PR.

In other words:
I vote for leaving this method `Obsolete` as-is, and think about a "proper" solution, preferably without the use of "Pubternal" through name/namespace/analyzer.

Flash0ver: ### re: CrashType

I do agree ... this is a bit of a misuse of `Obsolete` ... an Analyzer would be the better fit indeed.
A custom and specific diagnostic that indicates to the user that this method will crash.
This way we can also add a dedicated Help-Link, that the user may click in their IDE that could lead to a documentation page where we describe this and all our future _Analyzers_.

But I'd like to see this without an additional `Attribute`.
An Attribute like `InternalApi` reminds me of [Pubternal](https://learn.microsoft.com/dotnet/core/compatibility/aspnet-core/5.0/localization-pubternal-apis-removed), a pattern that has been "banned" by Microsoft for the .NET libraries.
Just a Diagnostic-Analyzer, that is enabled per default, reports a _Warning_ per default, and flags all usages of `Sentry.CrashType` and `SentrySdk.CauseCrash`.

Flash0ver: ### re: EnableWatchdogTerminationTracking

Hmmm .... I'm uncertain right now 🤔

jamescrosswell: We can't really mark IHasExtra as obsolete... too many classes implement it. I'd be very surprised if any of our SDK users are using that interface. 

I don't personally see any issue with removing this interface as part of a breaking change, at some point, without having previously marked it as obsolete.

We should definitely mark `HasExtraExtensions.SetExtras` as obsolete now though... that's pretty much free.

jamescrosswell: What are the possible alternative solutions?

We only have two options on the table at the moment:
- Leave it as is (abuse the `Obsolete` attribute)
  - The `Obsolete` attribute is misleading
  - **_Hopes_** people will read the comment before using it
- Create a more specific `InternalApi` atrribute and an analyzer
  - More specific attribute and error message
  - Guarantees people will see it before use, via the analyzer

Both are essentially implementations of Pubternal... The second solution seems superior in every way.

If we've got a 3rd solution that is even better, that would be awesome... but if not, we should do the best we can. I think the saying in product circles is, "Don't let perfect be the enemy of good".

jamescrosswell: > reports a Warning per default, and flags all usages of Sentry.CrashType and SentrySdk.CauseCrash

If not being used in one of our samples, I'm guessing (we don't really need/want the warning there).

This probably isn't a high priority and we could get away with making this change in a point release... I know the type is technically public but it's really just for demonstration purposes (and I think that's pretty obvious to our SDK users).

jamescrosswell:
@Flash0ver I think this PR turned out to be a bit too broad... so I created separate issues for all the unresolved discussions (see the PR description, linking to all of those). 

That should hopefully unblock the work that's already been done here, which doesn't have any dependencies on those other things so no reason to force them all into the same change.

CHANGELOG.md
jamescrosswell: I've created a separate issue to track this discussion and further work:
- https://github.com/getsentry/sentry-dotnet/issues/4640

jamescrosswell: Created a separate issue for this:
- https://github.com/getsentry/sentry-dotnet/issues/4641

jamescrosswell: Created a separate issue for this:
- https://github.com/getsentry/sentry-dotnet/issues/4642

jamescrosswell: Moved to a separate issue:
- https://github.com/getsentry/sentry-dotnet/issues/4643

test/Sentry.Maui.Tests/BreadcrumbEventTests.cs
```diff
@@ -14,10 +14,8 @@ public void BreadcrumbEvent_OldConstructor_EquivalentToNewConstructor()
         var eventName = "TestEvent";
 
         // Act
-        IEnumerable<(string Key, string Value)>[] extraData = [[("key1", "value1")], [("key2", "value2")]];
-#pragma warning disable CS0618 // Type or member is obsolete
+        IEnumerable<(string Key, string Value)> extraData = [("key1", "value1"), ("key2", "value2")];
```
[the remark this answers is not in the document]
jamescrosswell: Removed the test... no longer required.

src/Sentry/Protocol/Envelopes/EnvelopeItem.cs
```diff
@@ -417,18 +403,6 @@ private static async Task<ISerializable> DeserializePayloadAsync(
             return new JsonSerializable(sentryEvent);
         }
 
-        // User report
-        if (string.Equals(payloadType, TypeValueUserReport, StringComparison.OrdinalIgnoreCase))
```
Flash0ver, line 421: question: Feedback

After removing `UserFeedback.FromJson` I believe there is no `SentryFeedback.FromJson` ... is that needed?

Also, I haven't added Structured-Logs to this Deserialization-Flow ... is this something we need to follow-up with?

test/Sentry.Tests/Protocol/UserFeedbackTests.cs
```diff
@@ -1,36 +0,0 @@
-namespace Sentry.Tests.Protocol;
-
-public class UserFeedbackTests
```
Flash0ver, line 3: question: testing `SentryFeedback`

I believe we are lacking explicit/dedicated `SentryFeedbackTests.cs`.
Are they needed? Or do we cover the `SentryFeedback` serializable already through other tests?

test/Sentry.Tests/SentryClientExtensionsTests.cs
```diff
@@ -78,27 +78,6 @@ public void CaptureMessage_NullMessage_DoesNotCapturesEventWithMessage()
         Assert.Equal(default, id);
     }
 
-    [Fact]
-    public void CaptureUserFeedback_EnabledClient_CapturesUserFeedback()
```
Flash0ver, line 82: question: add tests for `CaptureFeedback`

We are already missing coverage for `CaptureFeedback`.
Could we, instead of deleting these tests, rewrite them to test `CaptureFeedback` instead?

Or should this be a separate PR, as it's adding coverage that is missing already.

test/Sentry.Tests/Protocol/UserFeedbackTests.cs
jamescrosswell: Not related to this PR, I don't think. I'll create a new issue to investigate that. 

Let's merge this PR and move on though.

PR #4620, opened 2025-10-10
chore: Log tweaks

bitsandfoxes:
I'm sorry, but the inconsistency has been bothering me when there are multiple logs in a row and they are only almost formatted the same. We've got the pattern of surrounding IDs with `'` almost everywhere. I just added it to where I found it missing.

```
Sentry: (Debug) Enqueuing envelope 72c00a9f348444918845d48a0a6049e5 
Sentry: (Info) Envelope queued up: '72c00a9f348444918845d48a0a6049e5' 
```

Same with instances of ending sentences without a `.`.

#skip-changelog

### DIFF
--- a/src/Sentry/GlobalSessionManager.cs
+++ b/src/Sentry/GlobalSessionManager.cs
@@ -118,7 +118,7 @@ internal class GlobalSessionManager : ISessionManager
                 try
                 {
                     var contents = _options.FileSystem.ReadAllTextFromFile(filePath);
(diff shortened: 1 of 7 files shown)


jamescrosswell:
@bitsandfoxes would it be possible to make this PR in the version6 branch rather than main? 

We're trying to minimise the number of changes we make in main at this point (to reduce double handling resulting from merges).

bitsandfoxes:
Sure thing!
Superseded by https://github.com/getsentry/sentry-dotnet/pull/4622

PR #4622, opened 2025-10-10
chore: Logging tweaks

bitsandfoxes:
I'm sorry, but the inconsistency has been bothering me when there are multiple logs in a row and they are only almost formatted the same. We've got the pattern of surrounding IDs with `''` almost everywhere. I just added it to where I found it missing.

```
Sentry: (Debug) Enqueuing envelope 72c00a9f348444918845d48a0a6049e5 
Sentry: (Info) Envelope queued up: '72c00a9f348444918845d48a0a6049e5' 
```

Same with instances of ending sentences without a `.`.

#skip-changelog


bitsandfoxes:
Ahh tests.. checking for exact log message matches.. lemme fix that.

PR #4627, opened 2025-10-13
QOL: Allow finishing spans and transactions with `using`

### LINKED ISSUE #4321: ITransactionTracer should implement IDisposable
### Problem Statement

Transaction scopes should be easier to manage in code by having the ITransactionTracer implement IDisposable.

### Solution Brainstorm

```CSharp
**using** var tran = SentrySdk.StartTransaction("Startup", "Init");
/* Do something */
/* Once the code goes out of scope, the transaction should "complete" automatically because of the "Dispose" method getting called explicitly, or implicitly. */
```

### Product Area

Unknown

jamescrosswell:
Resolves #4321 
- #4321


jamescrosswell:
@Flash0ver I'm trying to force an analyzer warning for the `CA2000` (so that I can build and test a way to suppress this). I've added this to the csproj file for `samples/Sentry.Samples.Console.Basic/Sentry.Samples.Console.Basic.csproj`:
```xml
<WarningsAsErrors>CA2000</WarningsAsErrors>
```

However I can't seem to get it to work:
image

The warning does appear for the undisposed use of `Foo` (which I added to test the warning was enabled)... but we're not seeing it for our newly disposable `ISpan` classes. `StartTransaction` returns an `ITransactionTracer` that descends from `ISpan`, which implements `IDisposable`.

Any ideas/thoughts?

Flash0ver:
Oh ... perhaps I totally misunderstood [CA2000](https://learn.microsoft.com/dotnet/fundamentals/code-analysis/quality-rules/ca2000) 😟

[DisposeObjectsBeforeLosingScope.cs](https://github.com/dotnet/sdk/blob/main/src/Microsoft.CodeAnalysis.NetAnalyzers/src/Microsoft.CodeAnalysis.NetAnalyzers/Microsoft.NetCore.Analyzers/Runtime/DisposeObjectsBeforeLosingScope.cs)

Hmmm ... it seems that the CA2000-DiagnosticAnalyzer does some Control-Flow-Analysis, that either tries to "see through" the implementation and does not report when it can't statically figure it out, or just does not report for some usage patterns.

```CSharp
var instance1 = new MyDisposable(); //CA2000
var instance2 = new MyFactory().CreateInstance(); //CA2000
var instance3 = MyFactory.CreateStatic(); //CA2000
var instance4 = MyFactory.Instance.CreateInstance(); // does not report CA2000

public sealed class MyDisposable : IDisposable
{
    public void Dispose()
    {
    }
}
public interface IMyFactory
{
    IDisposable CreateInstance();
}
public sealed class MyFactory : IMyFactory
{
    public static IMyFactory Instance { get; } = new MyFactory();
    public static IDisposable CreateStatic() => new MyDisposable();
    public IDisposable CreateInstance() => new MyDisposable();
}
```

Not sure if that analysis behavior is intended or not.
Perhaps the Analyzer cannot "see through" our rather complex control flow from `SentrySdk`, via the `IHub`, to some _extension-methods_.

---

Now, as a follow-up, I'm wondering:
We do have 2 `public` types that inherit `ISpan`, and with that `IDisposable`:
```CSharp
var transaction = new TransactionTracer(null!, null!); //CA2000
var span = new SpanTracer(null!, transaction, default, default, null!); //CA2000
span.Finish();
transaction.Finish();
```

Are there valid/intended usage scenarios to instantiate these types directly from user code, which would report `CA2000`, even when properly `Finish`ed?
Or ... perhaps we do not need a _DiagnosticSuppressor_ for `CA2000`, since it's not reported for our implementation ... which may change in the future 🤔

jamescrosswell:
> Are there valid/intended usage scenarios to instantiate these types directly from user code, which would report `CA2000`, 

I can't think of any scenario in which people would want to construct those manually... so I suspect not.

> Or ... perhaps we do not need a _DiagnosticSuppressor_ for `CA2000`, since it's not reported for our implementation ... 

That's possibly the case yes... and without being able to reproduce the warning that it's supposed to suppress, it's relatively difficult to test. I think this PR might be ready for review then...

src/Sentry/SpanTracer.cs
```diff
@@ -6,10 +6,11 @@ namespace Sentry;
 /// <summary>
 /// Transaction span tracer.
 /// </summary>
-public class SpanTracer : IBaseTracer, ISpan
+public sealed class SpanTracer : IBaseTracer, ISpan
```
Flash0ver, line 9: question: Is it worth a consideration to make `SpanTracer` internal, too? (in a potential follow-up)

So that users cannot run into `CA2000`
```CSharp
var span = new SpanTracer(null!, transaction, default, default, null!); //CA2000
span.Finish();
```

Or is it a potential user code scenario to try to cast?
```CSharp
if (SentrySdk.StartSpan("operation", "description") is SpanTracer span)
```

But this could have quite a large impact.

src/Sentry/SpanTracer.cs
```diff
@@ -190,5 +191,13 @@ internal set
             _origin = value;
         }
     }
-    private string? _origin;
+
+    /// <summary>
+    /// Automatically finishes the span at the end of a <c>using</c> block. This is a convenience method only. Disposing
+    /// is not required (and analyser warnings are suppressed).
```
Flash0ver: suggestion:
- shall we remove the note about `(and analyser warnings are suppressed)` ... as it seems we cannot reproduce `CA2000`, unless a `SpanTracer` is directly `new`ed.
- should we mention that
  - `Dispose` finishes the span as "OK" / Successful
  - is equivalent to `<see cref="Finish()"/>`
  - and to use any of these overloads to finish as error / unsuccessfully
    - `<see cref="Finish(SpanStatus)"/>`
    - `<see cref="Finish(Exception)"/>`
    - `<see cref="Finish(Exception, SpanStatus)"/>`

src/Sentry/TransactionTracer.cs
```diff
@@ -7,13 +7,14 @@ namespace Sentry;
 /// <summary>
 /// Transaction tracer.
 /// </summary>
-public class TransactionTracer : IBaseTracer, ITransactionTracer
+public sealed class TransactionTracer : IBaseTracer, ITransactionTracer
```
Flash0ver, line 10: question: Is it worth a consideration to make `TransactionTracer` internal, too? (in a potential follow-up)

So that users cannot run into `CA2000`
```CSharp
var transaction = new TransactionTracer(null!, null!); //CA2000
transaction.Finish();
```

Or is it a potential user code scenario to try to cast?
```CSharp
if (SentrySdk.StartTransaction("name", "operation") is TransactionTracer transaction)
```

But this could have quite a large impact.

src/Sentry/TransactionTracer.cs
```diff
@@ -431,4 +438,13 @@ private void ReleaseSpans()
 #endif
         _activeSpanTracker.Clear();
     }
+
+    /// <summary>
+    /// Automatically finishes the span at the end of a <c>using</c> block. This is a convenience method only. Disposing
+    /// is not required (and analyser warnings are suppressed).
```
Flash0ver: suggestion:
- shall we remove the note about `(and analyser warnings are suppressed)` ... as it seems we cannot reproduce `CA2000`, unless a `TransactionTracer` is directly `new`ed.
- should we mention that
  - `Dispose` finishes the transaction as "OK" / Successful
  - is equivalent to `<see cref="Finish()"/>`
  - and to use any of these overloads to finish as error / unsuccessfully
    - `<see cref="Finish(SpanStatus)"/>`
    - `<see cref="Finish(Exception)"/>`
    - `<see cref="Finish(Exception, SpanStatus)"/>`

test/Sentry.Tests/SpanTracerTests.cs
```diff
@@ -36,7 +36,38 @@ public async Task SetExtra_DataInserted_NoDataLoss()
 
             // Arrange
             // 4 tasks testing X amount should be the same amount as Extras.
-            Assert.Equal(4 * amount, transaction.Extra.Count);
+            Assert.Equal(4 * amount, spanTracer.Extra.Count);
         }
     }
+
+    [Fact]
+    public void Dispose_Finished_Finishes()
```
Flash0ver: question: Are part of the names here inverted from what each test case actually tests?

```diff
-public void Dispose_Finished_Finishes()
+public void Dispose_Unfinished_Finishes()

-public void Dispose_Unfinished_DoesNothing()
+public void Dispose_Finished_DoesNothing()
```

Or am I reading this wrongly?

src/Sentry/TransactionTracer.cs
```diff
@@ -361,6 +362,12 @@ public void Clear()
     /// <inheritdoc />
     public void Finish()
     {
+        // TODO: Replace with InterlockedBoolean once this has been merged into version6
```
Flash0ver: TODO (as a "resolvable" comment)

src/Sentry/TransactionTracer.cs
```diff
@@ -361,6 +362,12 @@ public void Clear()
...
+        if (Interlocked.Exchange(ref _hasFinished, 1) == 1)
```
Flash0ver: question: Shall we create a dedicated test for double-Disposal / double-Finishing?

Or is it actually sufficiently covered already by `TransactionTracerTests.Dispose_Finished_DoesNothing`?
I can imagine it is, as this new test drove this guard, right?

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 ### BREAKING CHANGES
 
 - SentryOptions.IsEnvironmentUser now defaults to false on MAUI. The means the User.Name will no longer be set, by default, to the name of the device ([#4606](https://github.com/getsentry/sentry-dotnet/pull/4606))
+- Spans and Transactions now implement `IDispose` so that they can be used with `using` statements that will automatically finish the span when it passes out of scope, to be consistent with `Activity` classes when using OpenTelemetry ([#4627](https://github.com/getsentry/sentry-dotnet/pull/4627))
```
Flash0ver: suggestion:
- `IDispose` -> `IDisposable`
- also mention using-declarations alongside using-statements
- specify "Successfully"


```suggestion
- Spans and Transactions now implement `IDisposable` so that they can be used with `using` statements/declarations that will automatically finish the span successfully when it passes out of scope, to be consistent with `Activity` classes when using OpenTelemetry ([#4627](https://github.com/getsentry/sentry-dotnet/pull/4627))
```

src/Sentry/SpanTracer.cs
Flash0ver, line 9: question: CHANGELOG

Is it worth a note in the CHANGELOG that both
- `SpanTracer`
- `TransactionTracer`

are no longer inheritable?

jamescrosswell: Yeah I'm not sure if people are trying to cast these.

I think for the time being we should leave it as is. Eventually the plan is to get rid of transactions and, at that point, there's probably less need to cast from ISpan so we could re-evaluate.

src/Sentry/SpanTracer.cs
jamescrosswell:
```suggestion
    /// is not required.
```

src/Sentry/TransactionTracer.cs
jamescrosswell: See comment re SpanTracer above.

test/Sentry.Tests/SpanTracerTests.cs
jamescrosswell: Per the [CONTRIBUTING.md](https://github.com/getsentry/sentry-dotnet/blob/main/CONTRIBUTING.md#naming-tests), the first part is the method `Dispose`.

The second part is the context (`Finished` or `Unfinished`... this could be more elaborate like `TransactionIsFinsished` and `TransactionIsUnfinished` but I figured that was clear from both the SUT and the implementation. 

The last part is the expectation (e.g. `DoesNothing`).

src/Sentry/TransactionTracer.cs
jamescrosswell: Yeah that's what the `Dispose_Unfinished_Finishes` and `Dispose_Finished_DoesNothing` tests are for so I think this is already covered.

src/Sentry/SpanTracer.cs
Flash0ver: Sounds good

test/Sentry.Tests/SpanTracerTests.cs
Flash0ver: Sorry ... I phrased it poorly.
What I meant was that `Dispose_Finished_Finishes` should actually be `Dispose_Unfinished_Finishes`,
and that `Dispose_Unfinished_DoesNothing` should be `Dispose_Finished_DoesNothing`.
It was just the name of the middle "context" portion what was flipped.

This is fixed now.

Flash0ver: But I believe that the names are still not quite correct:
- `Dispose_TransactionIsUnfinished_FinishesTransaction`
- `Dispose_TransactionIsFinsished_DoesNothing`

But both of these tests are in the `SpanTracerTests.cs`.
I guess it should say "SpanIs(Un)Finished" and "FinishesSpan".

But I am actually fine with the shorter version ... as it does make sense within the context of the respective test type:
- `Dispose_Unfinished_Finishes`
- `Dispose_Finished_DoesNothing`

Since we have the same test names then on `TransactionTracerTests.cs` as well.
Same test name, but in a different test class.

src/Sentry/SpanTracer.cs
```diff
@@ -190,5 +191,13 @@ internal set
             _origin = value;
         }
     }
-    private string? _origin;
+
+    /// <summary>
+    /// Automatically finishes the span at the end of a <c>using</c> block. This is a convenience method only. Disposing
+    /// is not required.
+    /// </summary>
```
Flash0ver: suggestion: shall we update this comment to be similar to the comment on `TransactionTracer.Dispose`?

```suggestion
    /// <summary>
    /// <para>
    /// Automatically finishes the span with a status of <see cref="SpanStatus.Ok" /> at the end of a
    /// <c>using</c> block, if it has not already been finished.
    /// </para>
    /// <para>
    /// This is the equivalent of calling <see cref="Finish()" /> when the transaction passes out of scope.
    /// </para>
    /// </summary>
    /// <remarks>This is a convenience method only. Disposing is not required.</remarks>
```

src/Sentry/TransactionTracer.cs
```diff
@@ -362,6 +363,11 @@ public void Clear()
     /// <inheritdoc />
     public void Finish()
     {
+        if (_hasFinished.Exchange(true))
+        {
+            return;
```
Flash0ver, line 368: question: Timer

~Should we, in the case of a Timer still being active, disable/dispose the timer too when the transactions is transitioning from unfinished to finished?~

Wait ... I read this wrong ... this is actually what is happening:
- the first `Finish` will Stop/Dispose the Timer, if active
- subsequent `Finish`es will early-out, but the Timer, if active, has been stopped/disposed previously

test/Sentry.Tests/SpanTracerTests.cs
```diff
@@ -36,7 +36,38 @@ public async Task SetExtra_DataInserted_NoDataLoss()
 
             // Arrange
             // 4 tasks testing X amount should be the same amount as Extras.
-            Assert.Equal(4 * amount, transaction.Extra.Count);
+            Assert.Equal(4 * amount, spanTracer.Extra.Count);
         }
     }
+
+    [Fact]
+    public void Dispose_IsUnfinished_FinishesTransaction()
```
Flash0ver: question: `.._FinishesTransaction`

Shouldn't this be `.._FinishesSpan`?

CHANGELOG.md
```diff
@@ -6,6 +6,8 @@
 
 - `BreadcrumbLevel.Critical` has been renamed to `BreadcrumbLevel.Fatal` for consistency with the other Sentry SDKs ([#4605](https://github.com/getsentry/sentry-dotnet/pull/4605))
 - SentryOptions.IsEnvironmentUser now defaults to false on MAUI. The means the User.Name will no longer be set, by default, to the name of the device ([#4606](https://github.com/getsentry/sentry-dotnet/pull/4606))
+- Spans and Transactions now implement `IDispose` so that they can be used with `using` statements/declarations that will automatically finish the span with a status of OK when it passes out of scope, if it has not already been finished, to be consistent with `Activity` classes when using OpenTelemetry ([#4627](https://github.com/getsentry/sentry-dotnet/pull/4627))
```
Flash0ver: suggestion: seems like a previously applied suggestion has been overridden/reverted somehow

```suggestion
- Spans and Transactions now implement `IDisposable` so that they can be used with `using` statements/declarations that will automatically finish the span with a status of OK when it passes out of scope, if it has not already been finished, to be consistent with `Activity` classes when using OpenTelemetry ([#4627](https://github.com/getsentry/sentry-dotnet/pull/4627))
```

src/Sentry/TransactionTracer.cs
jamescrosswell: The timer is only relevant if idle timeouts are enabled. That was added by Sean here:
- https://github.com/getsentry/sentry-dotnet/pull/2452

I think there were plans to maybe use transactions that timed out automatically to instrument MAUI, but this hasn't yet been done. So, in practice, that code path is never executed.

I'm not sure I understand though - do you think there's a potential problem here? In what scenario?

src/Sentry/TransactionTracer.cs
```diff
@@ -362,6 +363,11 @@ public void Clear()
...
+        }
+
```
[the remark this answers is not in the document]
jamescrosswell: We ignore Sentry requests... so this doesn't really matter.

Issue #4632, opened 2025-10-14
New Session status type `Unhandled`


bitsandfoxes:
### Description

Relevant docs PR https://github.com/getsentry/sentry-docs/pull/15086

The existing [`SessionEndStatus`](https://github.com/getsentry/sentry-dotnet/blob/c57c0f284b8a9c6564bd56727a67a05ad0d51ba8/src/Sentry/SessionEndStatus.cs) is to be extended by a new status `Unhandled`.
This splits the session update after error capture 
https://github.com/getsentry/sentry-dotnet/blob/c57c0f284b8a9c6564bd56727a67a05ad0d51ba8/src/Sentry/SentryClient.cs#L364-L376

in two pieces:
- Exceptions that cause the application to `crash`
- Exceptions that do not cause a crash - i.e. through the [`UnobservedTaskExceptionIntegration`](https://github.com/getsentry/sentry-dotnet/blob/c57c0f284b8a9c6564bd56727a67a05ad0d51ba8/src/Sentry/Integrations/UnobservedTaskExceptionIntegration.cs) or the Unity game engine

The `Unhandled` session end status is also terminal, meaning that the session updated as `Unhandled` is to be ended and not to receive further updates.

Issue #4186, opened 2025-05-13
(MAUI) It's hard to know if an event made an app crash


lucas-zimerman:
https://sentry.lkp.app.br/share/issue/6a79ea41e23348e9ad9680c4887ca800/

```
System.InvalidOperationException: PlatformView cannot be null here
  ?, in UIImageView ViewHandler<IImage, UIImageView>.get_PlatformView()
  ?, in UIImageView ImageHandler.Microsoft.Maui.Handlers.IImageHandler.get_PlatformView()
  ?, in void ImageImageSourcePartSetter.SetImageSource(UIImage platformImage)
  ?, in async Task<IImageSourceServiceResult<UIImage>> ImageSourcePartExtensions.UpdateSourceAsync(IImageSourcePart image, UIView destinationContext, IImageSourceServiceProvider services, Action<UIImage> setImage, float imageScale, CancellationToken cancellationToken)
  ?, in async Task ImageSourcePartLoader.UpdateImageSourceAsync()
  ?, in async Task ImageHandler.MapSourceAsync(IImageHandler handler, IImage image)
  ?, in async void TaskExtensions.FireAndForget(Task task, Action<Exception> errorCallback)
```

This issue is marked as unhandled, but the trace shows user used the app after the unhandled exception (all those stacktraces came from the Maui Assembly)

With that said, it's hard to know when an app actually crashed due to an event, that is helpful information to prioritize issues.

jamescrosswell:
@lucas-zimerman on iOS we can't use the `UnhandledException` callback so we use the `RuntimeMarshalManagedException` callback instead (see https://github.com/getsentry/sentry-dotnet/pull/3909).

One of the downsides of this is that the arguments to `RuntimeMarshalManagedException` don't indicate whether the exception is terminal or not. 

We're hoping the root cause gets addressed in .NET 10:
- https://github.com/getsentry/sentry-dotnet/issues/4180

jamescrosswell:
> would it be possible to mark the issues as unhandled, and after a period of time (lets say 5 seconds) if the code is still functional then mark it again as handled?

I don't think so no. In the case that the event does lead to an app crash, we want to try to capture the event as soon as possible. Since we don't know whether the exception is terminal or not, there's no way to work out whether we should send the event immediately or hold it in purgatory.

Issue #4180, opened 2025-05-12
See if we can replace with UnhandledException handler


jamescrosswell:
In order to reliably catch unhandled exceptions on iOS we listen to `Runtime.MarshalManagedException` rather than `AppDomain.UnhandledException`. Whilst this seems to work, it has the disadvantage that the args passed to `Runtime.MarshalManagedException` don't indicate whether the exception is terminal or not (we just assume it is). 

Apparently this should be fixed in .NET 10:
- https://github.com/dotnet/macios/issues/15252#issuecomment-2871164622

We'd need to test this out of course. If it works though, we could do away with the `Runtime.MarshalManagedException` callback.

jamescrosswell:
<p><a href="https://linear.app/getsentry/issue/NET-307">NET-307</a></p>

Issue #2877, opened 2023-11-21
Make SentryEvent.HasTerminalException() public


espenrl:
### Problem Statement

The scenario is regarding a MAUI app where common exceptions regarding network timeouts are of little interest to collect.  Unless they are terminal which is a scenario that effectively kills the app. Then even a network timeout exception is suddenly of high interest to collect.

### Solution Brainstorm

Make `SentryEvent.HasTerminalException()` public so the API can be used to make decisions about which exceptions to ignore.

bitsandfoxes:
Hey @espenrl! What stops you from filtering based on the exception in the `BeforeSend` callback?

espenrl:
Hi @bitsandfoxes What stops me is that there isn't anything indicating if the event is terminal or not. Terminal being a Sentry specific term - for MAUI it would be an unhandled exception that crashes the app.

I saw in the source code that there is an API to check if an event is terminal or not, though it's marked `internal`.

https://github.com/getsentry/sentry-dotnet/blob/65a7bba00a73db6c619aa5d4a44ca4da2092b738/src/Sentry/SentryEvent.cs#L184-L198

bruno-garcia:
I'd rename that to `IsUnhandled` then, with a big explanation that this is from Sentry's PoV, by default events are handled:true, and we explicitly set to false when we capture things in one of our hooks like UnhandledExceptionHandler, Middleware on ASPNET Core, etc

bitsandfoxes:
Adding it to the `hint` is a great idea. The thing is under-utilized as it is!
It would look like this:
```
options.SetBeforeSend((@event, hint) =>
{
    if (hint.Items.ContainsKey("HasTerminalException"))
    {
        // This event has a terminal exception
    }

    return @event;
});
```
The key is up for discussion tho. The only downside to this I can see right now is that it's basically undocumented in code.

bruno-garcia:
I'd hide all of that on an extension method to Hint (or a method on it) so the user doesn't need all that boilerplate:

options.SetBeforeSend((@event, hint) =>
{
    if (hint.IsUnhandled()))
    {
        // This event has an unhandled exception
    }

    return @event;
});

PR #4633, opened 2025-10-14
feat: Added `SessionEndStatus.Unhandled`

bitsandfoxes:
Relates to https://github.com/getsentry/sentry-dotnet/issues/4632

### Problem

The SDK should use the new `SessionEndStatus.Unhandled` when capturing an unhandled exception that does not cause the application to crash.

Unhandled exceptions typically lead to a terminal state of the application. The only exception (hehe) to this was the `UnobservedTaskExceptionIntegration`. Another instance where this is highly relevant is Unity: Unhandled exceptions get "swallowed" by the engine and the game just continues. 

### Context

Within the Sentry SDK for .NET the only non-terminal unhandled exception comes from the `UnobservedTaskExceptionIntegration` that also gets special cased in `SentryEvent.HasTerminalException()` where we compare against the mechanism key

https://github.com/getsentry/sentry-dotnet/blob/04c932b5537d3f236d6604a99f61b5e3a8d41b18/src/Sentry/SentryEvent.cs#L188-L191

This is limiting as it does not allow for other integrations i.e. coming from the Unity SDK to add to this check.

#### <del>Option 1: Add list of `NonTerminalMechanismKey` on the options</del>

One straight forward solution to this would be to have an `internal List<string>? NonTerminalMechanismKeys;` on
the options and have integrations add to the list during their registration. We'd need to pass the options into the `HasTerminalException` check but they are available everywhere we'd need them.

There are a couple of downsides to this:
1. Should the list be `internal` or `public`. If we keep it `internal` the Unity SDK can add to it but no user written integration can.
2. We're polluting the options with one more list of things.
3. It's just a bunch of string comparisons that make this work

#### <del>Option 2: Rework the Mechanism `bool? Handled` to an explicit `enum ExceptionHandledState`</del>

The `Terminal` state is only ever relevant in case of an `Unhandled Exception`. This makes the additional property somewhat redundant. An alternative solution would be to replace the `bool? Handled` with an explicit `ExceptionHandledState`. This would be quite the breaking change and we'd need to take special care during serialization.

```
public enum ExceptionHandledState
{
    /// <summary>Exception was never thrown (constructed and captured)</summary>
    NotThrown,

    /// <summary>Exception was caught by user code</summary>
    Handled,

    /// <summary>Unhandled exception that causes application termination</summary>
    UnhandledTerminal,

    /// <summary>Unhandled exception that doesn't cause termination (e.g., UnobservedTask, Unity)</summary>
    UnhandledNonTerminal
}

public ExceptionHandledState? HandledState { get; set; }
```

#### Option 3: Extend `Mechanism` to have a `bool? Terminal` property

The mechanism on the exception already has things like `Handled` and `Synthetic`. We can extend this by adding `bool? Terminal`. Integrations that capture unhandled non-terminal exceptions can set the `Terminal` state to `false` and we can check for this on the `SentryEvent`. We keep it on the top-level of `Mechanism` but we don't need to serialize it so it won't get sent to Sentry.

This sounds to me the most pragmatic and minimal invasive change. The only downside I can see right now is that the property is only ever going to be useful when `Handled` is `false`. But we can keep it nullable.

### DIFF
--- a/src/Sentry/Internal/Hub.cs
+++ b/src/Sentry/Internal/Hub.cs
@@ -585,7 +585,7 @@ internal class Hub : IHub, IDisposable
             scope.LastEventId = id;
             scope.SessionUpdate = null;
 
-            if (evt.GetExceptionType() is SentryEvent.ExceptionType.Unhandled
+            if (evt.GetExceptionType() is SentryEvent.ExceptionType.UnhandledTerminal
                 && scope.Transaction is { } transaction)
             {
                 // Event contains a terminal exception -> finish any current transaction as aborted
--- a/src/Sentry/Platforms/Android/LogCatAttachmentEventProcessor.cs
+++ b/src/Sentry/Platforms/Android/LogCatAttachmentEventProcessor.cs
@@ -55,7 +55,7 @@ internal class LogCatAttachmentEventProcessor : ISentryEventProcessorWithHint
             // Only send logcat logs if the event is unhandled if the integration is set to Unhandled
             if (_logCatIntegrationType == LogCatIntegrationType.Unhandled)
             {
-                if (exceptionType != SentryEvent.ExceptionType.Unhandled && exceptionType != SentryEvent.ExceptionType.UnhandledNonTerminal)
+                if (exceptionType != SentryEvent.ExceptionType.UnhandledTerminal && exceptionType != SentryEvent.ExceptionType.UnhandledNonTerminal)
                 {
                     return @event;
                 }
--- a/src/Sentry/SentryEvent.cs
+++ b/src/Sentry/SentryEvent.cs
@@ -182,7 +182,7 @@ public sealed class SentryEvent : IEventLike, ISentryJsonSerializable
     {
         None,
         Handled,
-        Unhandled,
+        UnhandledTerminal,
         UnhandledNonTerminal
     }
 
@@ -200,7 +200,7 @@ public sealed class SentryEvent : IEventLike, ISentryJsonSerializable
 
         if (HasUnhandledException())
         {
-            return ExceptionType.Unhandled;
+            return ExceptionType.UnhandledTerminal;
         }
 
         return ExceptionType.Handled;
--- a/src/Sentry/SentryClient.cs
+++ b/src/Sentry/SentryClient.cs
@@ -355,7 +355,7 @@ public class SentryClient : ISentryClient, IDisposable
                 scope.SessionUpdate = _sessionManager.EndSession(SessionEndStatus.Unhandled);
                 break;
 
-            case SentryEvent.ExceptionType.Unhandled:
+            case SentryEvent.ExceptionType.UnhandledTerminal:
                 _options.LogDebug("Ending session as 'Crashed', due to unhandled exception.");
                 scope.SessionUpdate = _sessionManager.EndSession(SessionEndStatus.Crashed);
                 break;
--- a/test/Sentry.Tests/Protocol/SentryEventTests.cs
+++ b/test/Sentry.Tests/Protocol/SentryEventTests.cs
@@ -204,7 +204,7 @@ public partial class SentryEventTests
         exception.SetSentryMechanism("AppDomain.UnhandledException", handled: false, terminal: true);
         var evt = new SentryEvent(exception);
 
-        Assert.Equal(SentryEvent.ExceptionType.Unhandled, evt.GetExceptionType());
+        Assert.Equal(SentryEvent.ExceptionType.UnhandledTerminal, evt.GetExceptionType());
     }
 
     [Fact]
@@ -215,7 +215,7 @@ public partial class SentryEventTests
             SentryExceptions = [new SentryException { Mechanism = new Mechanism { Handled = false } }]
         };
 
-        Assert.Equal(SentryEvent.ExceptionType.Unhandled, evt.GetExceptionType());
+        Assert.Equal(SentryEvent.ExceptionType.UnhandledTerminal, evt.GetExceptionType());
     }
 
     [Fact]


src/Sentry/Internal/MainExceptionProcessor.cs
```diff
@@ -187,6 +187,12 @@ private static Mechanism GetMechanism(Exception exception, int id, int? parentId
             mechanism.Handled = null;
         }
 
+        if (exception.Data[Mechanism.TerminalKey] is bool terminal)
+        {
+            mechanism.Terminal = terminal;
+            exception.Data.Remove(Mechanism.TerminalKey);
```
bruno-garcia: do we need to remove? curious why

src/Sentry/Protocol/Mechanism.cs
```diff
@@ -153,6 +181,7 @@ public static Mechanism FromJson(JsonElement json)
         var source = json.GetPropertyOrNull("source")?.GetString();
         var helpLink = json.GetPropertyOrNull("help_link")?.GetString();
         var handled = json.GetPropertyOrNull("handled")?.GetBoolean();
+        var terminal = json.GetPropertyOrNull("terminal")?.GetBoolean();
```
bruno-garcia: looks like wer'e changing the wire protocol here. Do we need this? afaik the Exception interface has some `Data` bag we could use instead.

In any case we need to check this is even accepted by Relay

bruno-garcia (commented):
while talking through this a bigger problem has been uncovered, with restarting sessions on unhandled

test/Sentry.AspNetCore.TestUtils/Sentry.AspNetCore.TestUtils.csproj
```diff
@@ -36,10 +36,8 @@
     <PackageReference Include="Microsoft.AspNetCore.Http" Version="2.1.22" />
     <PackageReference Include="Microsoft.AspNetCore.TestHost" Version="2.1.1" />
 
-    <!-- This is needed because the version that is brought in transitively also has a vulnerability warning -->
-    <PackageReference Include="Microsoft.AspNetCore.Server.Kestrel.Core" Version="2.1.25" />
-    <!-- https://github.com/advisories/GHSA-vmch-3w2x-vhgq -->
-    <PackageReference Include="Microsoft.AspNetCore.Server.Kestrel.Transport.Sockets" Version="2.1.40" />
+    <PackageReference Include="Microsoft.AspNetCore.Server.Kestrel.Core" Version="2.3.6" />
+    <PackageReference Include="Microsoft.AspNetCore.Server.Kestrel.Transport.Sockets" Version="2.3.0" />
```
Flash0ver: mismatch with this PR, where we don't bump `Microsoft.AspNetCore.Server.Kestrel.Transport.Sockets` which does not have a known vulnerability
- https://github.com/getsentry/sentry-dotnet/pull/4639

src/Sentry/Protocol/Mechanism.cs
Flash0ver: I only spotted the changes to the `SessionUpdate` protocol:
- https://github.com/getsentry/sentry-docs/pull/15086

Are there also changes to the [Mechanism](https://develop.sentry.dev/sdk/data-model/event-payloads/exception/#exception-mechanism) protocol?

src/Sentry/SentryEvent.cs
```diff
@@ -180,22 +180,39 @@ public IReadOnlyList<string> Fingerprint
...
-        // UNLESS it comes from the UnobservedTaskExceptionIntegration
+        if (Exception?.Data[Mechanism.HandledKey] is false)
+        {
+            return true;
+        }
+
+        return SentryExceptions?.Any(e => e.Mechanism is { Handled: false }) ?? false;
+    }
+
+    internal bool HasUnhandledNonTerminalException()
```
Flash0ver: question: order

Do I get that right that, for correct consumption, we need to call these methods in this order
1. `HasUnhandledNonTerminalException`
2. `HasUnhandledException`
3. `HasException`

Then there is also `HasTerminalException`.

It's a bit confusing to me ... I wonder if this could be designed a bit more clearly.
On the other hand ... this may just be me ... and the methods are all `internal`.

src/Sentry/SessionEndStatus.cs
```diff
@@ -13,10 +13,15 @@ public enum SessionEndStatus
     /// <summary>
     /// Session ended with an unhandled exception.
     /// </summary>
+    Unhandled,
```
Flash0ver, line 16: question: breaking change

We ship this in a major release, so this may be a non-concern.

But inserting an `enum` member is a breaking change for users that have code like:
```CSharp
_ = (SessionEndStatus)1;
```

Well ... I guess it's unlikely that a considerable amount of user code does this ... I just wanted to hear your thoughts.

src/Sentry/Protocol/Mechanism.cs
bruno-garcia: All server changes so far are about sessions. Nothing related to events

src/Sentry/SessionEndStatus.cs
bitsandfoxes: The only place where this can be used would be here
https://github.com/getsentry/sentry-dotnet/blob/2f46bf5e6228cb17d055b9b31101c1ba42fc0128/src/Sentry/SentrySdk.cs#L775
So a user would have to call `SentrySdk.EndSession((SessionEndStatus)1);` which does not seem very likely.

src/Sentry/SentryEvent.cs
bitsandfoxes: I see. I wrapped the thing with an `enum ExceptionType` and make the rest `private`. That makes it easier to reason about.

src/Sentry/Protocol/Mechanism.cs
bitsandfoxes: I've made some additional changes and the `Mechanism.TerminalKey` now stays on the client and won't get sent to Sentry.

bruno-garcia:
@bitsandfoxes CI is all broken

src/Sentry/Internal/Hub.cs
```diff
@@ -578,7 +578,7 @@ private SentryId CaptureEvent(SentryEvent evt, SentryHint? hint, Scope scope)
             {
                 // Event contains a terminal exception -> finish any current transaction as aborted
```
jamescrosswell, line 591: Originally I was going to suggest updating the comment here to:
```diff
- // Event contains a terminal exception -> finish any current transaction as aborted
+ // Event contains an unhandled exception -> finish any current transaction as aborted
```

But actually I see that `ExceptionType.Unhandled` is implicitly terminal:
https://github.com/getsentry/sentry-dotnet/blob/31416f9cf45d975184783b40e38b71ddb75d7d2d/src/Sentry/SentryEvent.cs#L196-L204

That wasn't obvious and is a bit confusing. Possibly we could change the name of that enum from `ExceptionType.Unhandled` to `ExceptionType.UnhandledTerminal` or just `ExceptionType.Terminal` to make that clearer.

bruno-garcia:
@sentry review

src/Sentry/Platforms/Android/LogCatAttachmentEventProcessor.cs
```diff
@@ -45,15 +45,17 @@ public SentryEvent Process(SentryEvent @event, SentryHint hint)
...
+            if (_logCatIntegrationType != LogCatIntegrationType.All && exceptionType == SentryEvent.ExceptionType.None)
             {
                 return @event;
             }
 
             // Only send logcat logs if the event is unhandled if the integration is set to Unhandled
             if (_logCatIntegrationType == LogCatIntegrationType.Unhandled)
             {
-                if (!@event.HasTerminalException())
+                if (exceptionType != SentryEvent.ExceptionType.Unhandled)
```
jamescrosswell:
```suggestion
                if (exceptionType != SentryEvent.ExceptionType.Unhandled && exceptionType != SentryEvent.ExceptionType.UnhandledNonTerminal)
```

`UnhandledNonTerminal` is still a form of unhandled right? So we want to attach logs for those?

bitsandfoxes: Good catch!

src/Sentry/Internal/Hub.cs
bitsandfoxes: Of course!

src/Sentry/SentryExceptionExtensions.cs
```diff
@@ -32,8 +32,9 @@ public static void AddSentryContext(this Exception ex, string name, IReadOnlyDic
     /// <param name="type">A required short string that identifies the mechanism.</param>
     /// <param name="description">An optional human-readable description of the mechanism.</param>
     /// <param name="handled">An optional flag indicating whether the exception was handled by the mechanism.</param>
+    /// <param name="terminal">An optional flag indicating whether the exception is considered terminal.</param>
     public static void SetSentryMechanism(this Exception ex, string type, string? description = null,
-        bool? handled = null)
+        bool? handled = null, bool? terminal = null)
```
jamescrosswell, line 37: The only thing I don't like about this is that it allows six possible combinations, some of which are inconsistent... `terminal` but `handled` doesn't make sense right? Also `terminal` + `handled == null` would be weird.

A nullable enum { Handled, Unhandled, Terminal } would constrain this to only consistent possibilities.

src/Sentry/SessionEndStatus.cs
```diff
@@ -13,10 +13,15 @@ public enum SessionEndStatus
...
+    /// <summary>
+    /// Session ended with a terminal unhandled exception.
+    /// </summary>
     Crashed,
 
     /// <summary>
     /// Session ended abnormally (e.g. device lost power).
     /// </summary>
-    Abnormal
+    Abnormal,
```
jamescrosswell, line 26: Can/Would we ever set this in practice? I see it's used in the tests...

src/Sentry/Protocol/Mechanism.cs
```diff
@@ -76,6 +84,15 @@ public string Type
...
     public bool? Handled { get; set; }
 
+    /// <summary>
+    /// Optional flag indicating whether the exception is terminal (will crash the application).
+    /// When false, indicates a non-terminal unhandled exception (e.g., unobserved task exception).
+    /// </summary>
+    /// <remarks>
+    /// This is an SDK-internal flag used for session tracking and is not serialized to Sentry servers.
+    /// </remarks>
+    public bool? Terminal { get; internal set; }
```
jamescrosswell, line 94: Any reason for this to be public? Could it also be scoped internal?

I guess it might be useful to SDK users eh?

jamescrosswell (commented):
In principle, no objections... just wonder if we can tidy up the API for the `SetSentryMechanism` extension method.

src/Sentry/Protocol/Mechanism.cs
bitsandfoxes: There is a number of issues where I think having access to whether the SDK thinks and event was terminal on top of whether the exception was unhandled comes in handy:
- https://github.com/getsentry/sentry-dotnet/issues/4186 (can a MAUI the integration set it?)
- https://github.com/getsentry/sentry-dotnet/issues/4180 Related to ^
- https://github.com/getsentry/sentry-dotnet/issues/2877 should resolve this and not have to make the methods on the `SentryEvent` public.

Something like this:

```
options.SetBeforeSend(@event =>
{
    var mechanism = @event.SentryExceptions?.First().Mechanism;
    if (mechanism != null)
    {
        WriteLine("The SDK considered this terminal: " + mechanism.Terminal);
    }

    return @event;
});
```

src/Sentry/SessionEndStatus.cs
bitsandfoxes: That's a good point and I think we're abusing this in the Unity SDK. It's a bit outside of the scope of this PR but I'll check when taking a stab at properly shutting down sessions during application closure.

src/Sentry/SentryExceptionExtensions.cs
bitsandfoxes: The protocol itself has `handled` and `terminal` and imo, the public API should mirror this instead of introducing an abstraction that needs to be transformed back and forth.

The combinations do have legitimate use cases:
- handled: `true`, terminal: `false`: Standard caught exception (try-catch)
- handled: `false`, terminal: `true`: Unhandled exception that crashes the app
- handled: `false`, terminal: `false`: Unhandled but non-terminal (e.g., UnobservedTaskException, unhandled or logged exceptions in Unity)
- handled: `true`, terminal: `true`: Could occur when an exception is caught but the app decides to terminate anyway, or when a shutdown is triggered during exception handling which is the only case the SDK does not set

The terminal flag is also SDK internal and not getting sent to Sentry and I'd argue most users will never set it directly, so I'd rather opt towards SDK needs than to theoretical type safety that a nullable enum might bring.

src/Sentry/Protocol/Mechanism.cs
jamescrosswell: Makes sense

PR #4653, opened 2025-10-17
feat: Cache `Unhandled` session instead of sending it immediately

bitsandfoxes:
Resolves https://github.com/getsentry/sentry-dotnet/issues/4632
Followup on https://github.com/getsentry/sentry-dotnet/pull/4633

### Problem

Sessions that end in `Unhandled` can later on still crash. Updating a session with an end status is a terminal operation, see https://github.com/getsentry/sentry-docs/pull/15086 and once updated, sessions must not receive further updates. We need to make sure the SDK finishes a session only when it is confident, that a session is actually finished.

### Proposal

Instead of updating (and ending) a session with `SessionEndStatus.Unhandled` right away we persist the state to disk. We can make use of the fact that there is a separation between [`PersistedSessionUpdate`](https://github.com/getsentry/sentry-dotnet/blob/4bc0e6c9b36b94616246e26d147657041ef46ce2/src/Sentry/PersistedSessionUpdate.cs) and [`SessionUpdate`](https://github.com/getsentry/sentry-dotnet/blob/4bc0e6c9b36b94616246e26d147657041ef46ce2/src/Sentry/SessionUpdate.cs).

- `PersistedSessionUpdate` is used locally only, the SDK does not send this to Sentry and we can add a flag `PendingUnhandled` to it
- `SessionUpdate` is the actual update the SDK sends to Sentry

The flow is as follows:

1. Unhandled non-terminal exception captured (i.e. via Unity loghandling)
2. The client triggers the session to be marked as `Unhandled` and the update is persisted to disk
3. If the app crashes we end the session as `Crashed`, overwriting the pending unhandled state
4. If the app exits normally, we check for the endstatus and overwrite `Exited` with `Unhandled`.

### TODO

Currently, the SDK relies on restarts of the application to send the `SessionEndStatus`. Ideally, we would also hook into some shutdown mechanism, or at the very least, closing the SDK should also end any active sessions.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,7 @@
 ### Features
 
 - The SDK now makes use of the new SessionEndStatus `Unhandled` when capturing an unhandled but non-terminal exception, i.e. through the UnobservedTaskExceptionIntegration ([#4633](https://github.com/getsentry/sentry-dotnet/pull/4633), [#4653](https://github.com/getsentry/sentry-dotnet/pull/4653))
+- The SDK now makes use of the new SessionEndStatus `Unhandled` when capturing an unhandled but non-terminal exception, i.e. through the UnobservedTaskExceptionIntegration ([#4633](https://github.com/getsentry/sentry-dotnet/pull/4633))
 
 ### Fixes
 
--- a/integration-test/common.ps1
+++ b/integration-test/common.ps1
@@ -209,6 +209,7 @@ BeforeAll {
       <SentryUploadAndroidProguardMapping>true</SentryUploadAndroidProguardMapping>
       <AndroidLinkTool Condition=`" '`$(AndroidLinkTool)' == '' `">r8</AndroidLinkTool>
       <AndroidDexTool Condition=`" '`$(AndroidDexTool)' == '' `">d8</AndroidDexTool>
+      <MtouchUseLlvm Condition=`"'`$(Configuration)' == 'Release'`">false</MtouchUseLlvm>
   </PropertyGroup>
 </Project>
 "@ | Out-File $name/Directory.Build.props
--- a/samples/Directory.Build.props
+++ b/samples/Directory.Build.props
@@ -7,7 +7,7 @@
   </PropertyGroup>
 
   <!-- Workaround for hang on compile issue.  See https://github.com/xamarin/xamarin-macios/issues/17825#issuecomment-1478568270. -->
-  <PropertyGroup Condition="'$(Configuration)' == 'Release' And $([MSBuild]::GetTargetPlatformIdentifier('$(TargetFramework)')) == 'ios'">
+  <PropertyGroup Condition="'$(Configuration)' == 'Release'">
     <MtouchUseLlvm>false</MtouchUseLlvm>
   </PropertyGroup>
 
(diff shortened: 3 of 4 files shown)


src/Sentry/GlobalSessionManager.cs
```diff
@@ -245,6 +249,13 @@ private void DeletePersistedSession()
 
     private SessionUpdate EndSession(SentrySession session, DateTimeOffset timestamp, SessionEndStatus status)
     {
+        // If we're ending as 'Exited' but he session has a pending 'Unhandled', end as 'Unhandled'
+        if (status == SessionEndStatus.Exited && session.IsMarkedAsPendingUnhandled)
+        {
+            status = SessionEndStatus.Unhandled;
+            _options.LogDebug("Session ended as 'Unhandled' due to pending status.");
```
jamescrosswell: Nit: just thinking of our future selves seeing this code in 12 month's time... maybe this is clearer to understand? Feel free to ignore if you disagree. I'm guessing this functionality will mostly be used by Unity users.

```suggestion
            _options.LogDebug("Session ended with pending 'Unhandled' (but not `Terminal`) exception.");
```

src/Sentry/SentrySession.cs
```diff
@@ -36,6 +36,13 @@ public class SentrySession : ISentrySession
     // Start at -1 so that the first increment puts it at 0
     private int _sequenceNumber = -1;
 
+    private int _isMarkedAsPendingUnhandled;
```
jamescrosswell:
```suggestion
    private InterlockedBoolean _isMarkedAsPendingUnhandled;
```

See: https://github.com/getsentry/sentry-dotnet/pull/4585

src/Sentry/SentrySession.cs
```diff
@@ -36,6 +36,13 @@ public class SentrySession : ISentrySession
...
+
+    /// <summary>
+    /// Gets whether this session has an unhandled exception that hasn't been finalized yet.
+    /// </summary>
+    internal bool IsMarkedAsPendingUnhandled => _isMarkedAsPendingUnhandled != 0;
```
jamescrosswell:
```suggestion
    internal bool IsMarkedAsPendingUnhandled => _isMarkedAsPendingUnhandled;
```

src/Sentry/SentrySession.cs
```diff
@@ -74,6 +81,12 @@ public SentrySession(string? distinctId, string release, string? environment)
     /// </summary>
     public void ReportError() => Interlocked.Increment(ref _errorCount);
 
+    /// <summary>
+    /// Marks the session as having an unhandled exception without ending it.
+    /// This allows the session to continue and potentially escalate to Crashed if the app crashes.
+    /// </summary>
+    internal void MarkUnhandledException() => Interlocked.Exchange(ref _isMarkedAsPendingUnhandled, 1);
```
jamescrosswell:
```suggestion
    internal void MarkUnhandledException() => _isMarkedAsPendingUnhandled.Exchange(true);
```

jamescrosswell (approved):
Looks good. I'd just use the new `InterlockedBoolean` shim that we have instead of an `int` for the thread safe `bool`.

PR #4657, opened 2025-10-20
feat(AI): `Microsoft.Extensions.AI` Agent Monitoring

### LINKED ISSUE #3762: Support for Microsoft.Extensions.AI
Announced last month:

https://devblogs.microsoft.com/dotnet/introducing-microsoft-extensions-ai-preview/

Sentry's LLM monitoring docs: https://docs.sentry.io/product/insights/ai/llm-monitoring/

Example API:

```csharp
app.Services.AddChatClient(builder => builder 
    // Add Sentry LLM monitoring
    .UseSentry()    
    .Use(new OpenAIClient(...)).AsChatClient(...)); 
```

alexsohn1126:
Resolves #3762
- #3762

# Purpose

This PR will add a new integration for `Sentry.Extensions.AI`.

It will allow users to instrument their LLM usage, including any tool calls and LLM API calls. `Microsoft.Extensions.AI.Abstractions` NuGet package is used in this SDK to use their interfaces and types.

# Usage

In order to use this SDK, the user will have to have a compatible SDK which can instantiate an `IChatClient`. In the samples, I have used [`Microsoft.Extensions.AI.OpenAI`](https://www.nuget.org/packages/Microsoft.Extensions.AI.OpenAI/9.10.1-preview.1.25521.4). Another popular LLM SDK is [`Anthropic.SDK`](https://www.nuget.org/packages/Anthropic.SDK).

There are a few AI-exclusive options the users can change.

- `IncludeAIRequestMessages` - boolean
  - Defaults to `true`
  - If `true`, we include the request message in the spans we send to Sentry.
- `IncludeAIResponseContent` - boolean
  - Defaults to `true`
  - If `true`, we include the response message in the spans we send to Sentry.
- `AgentName` - string
  - Defaults to `Agent`
  - Name of the agent which this `ChatClient` represents.
  
Once the user has set up and sends AI spans to Sentry, they will be able to see it in `Insights > AI` (from the sidebar).

https://github.com/user-attachments/assets/d3bc7b4e-53fa-487a-abae-7e622f6a075d

There are useful information about LLM usage in the webpage. Such as how many tokens the user has sent/received, what tool calls were made, how long they tool, etc.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,28 +14,6 @@
 - CaptureFeedback now returns a `SentryId` and a `CaptureFeedbackResult` out parameter that indicate whether feedback was captured successfully and what the reason for failure was otherwise ([#4613](https://github.com/getsentry/sentry-dotnet/pull/4613))
 - Deprecated `Sentry.Azure.Functions.Worker` as very few people were using it and the functionality can easily be replaced with OpenTelemetry. We've replaced our integration with a sample showing how to do this using our OpenTelemetry package instead. ([#4693](https://github.com/getsentry/sentry-dotnet/pull/4693))
 - UWP support has been dropped. Future efforts will likely focus on WinUI 3, in line with Microsoft's recommendations for building Windows UI apps. ([#4686](https://github.com/getsentry/sentry-dotnet/pull/4686))
-- The _Structured Logs_ APIs are now stable: removed `Experimental` from `SentryOptions` ([#4699](https://github.com/getsentry/sentry-dotnet/pull/4699))
-
-### Features
-
-- Implemented instance isolation so that multiple instances of the Sentry SDK can be instantiated inside the same process when using the Caching Transport ([#4498](https://github.com/getsentry/sentry-dotnet/pull/4498))
-- The SDK now makes use of the new SessionEndStatus `Unhandled` when capturing an unhandled but non-terminal exception, i.e. through the UnobservedTaskExceptionIntegration ([#4633](https://github.com/getsentry/sentry-dotnet/pull/4633), [#4653](https://github.com/getsentry/sentry-dotnet/pull/4653))
-- The SDK now provides a `IsSessionActive` to allow checking the session state ([#4662](https://github.com/getsentry/sentry-dotnet/pull/4662))
-- The SDK now makes use of the new SessionEndStatus `Unhandled` when capturing an unhandled but non-terminal exception, i.e. through the UnobservedTaskExceptionIntegration ([#4633](https://github.com/getsentry/sentry-dotnet/pull/4633))
-- Added a new SDK `Sentry.Extensions.AI` which allows LLM usage instrumentation via `Microsoft.Extensions.AI` ([#4657](https://github.com/getsentry/sentry-dotnet/pull/4657))
-- Added experimental support for Session Replay on iOS ([#4664](https://github.com/getsentry/sentry-dotnet/pull/4664))
-- Extended the App context by `app_memory` that can hold the amount of memory used by the application in bytes. ([#4707](https://github.com/getsentry/sentry-dotnet/pull/4707))
-
-### Fixes
-
-- The `Serilog` integration captures _Structured Logs_ (when enabled) independently of captured Events and added Breadcrumbs ([#4691](https://github.com/getsentry/sentry-dotnet/pull/4691))
-- Deliver system breadcrumbs in the main thread on Android ([#4671](https://github.com/getsentry/sentry-dotnet/pull/4671))
-- Memory leak when finishing an unsampled Transaction that has started unsampled Spans ([#4717](https://github.com/getsentry/sentry-dotnet/pull/4717))
-
-## 6.0.0-preview.2
-
-### BREAKING CHANGES
-
 - `BreadcrumbLevel.Critical` has been renamed to `BreadcrumbLevel.Fatal` for consistency with the other Sentry SDKs ([#4605](https://github.com/getsentry/sentry-dotnet/pull/4605))
 - SentryOptions.IsEnvironmentUser now defaults to false on MAUI. The means the User.Name will no longer be set, by default, to the name of the device ([#4606](https://github.com/getsentry/sentry-dotnet/pull/4606))
 - Remove unnecessary files from SentryCocoaFramework before packing ([#4602](https://github.com/getsentry/sentry-dotnet/pull/4602))
@@ -56,6 +34,7 @@
   - The SDK now makes use of the new SessionEndStatus `Unhandled` when capturing an unhandled but non-terminal exception, i.e. through the UnobservedTaskExceptionIntegration ([#4633](https://github.com/getsentry/sentry-dotnet/pull/4633), [#4653](https://github.com/getsentry/sentry-dotnet/pull/4653))
 - Implemented instance isolation so that multiple instances of the Sentry SDK can be instantiated inside the same process when using the Caching Transport ([#4498](https://github.com/getsentry/sentry-dotnet/pull/4498))
 - Extended the App context by `app_memory` that can hold the amount of memory used by the application in bytes. ([#4707](https://github.com/getsentry/sentry-dotnet/pull/4707))
+- Added a new SDK `Sentry.Extensions.AI` which allows LLM usage instrumentation via `Microsoft.Extensions.AI` ([#4657](https://github.com/getsentry/sentry-dotnet/pull/4657))
 
 ### Fixes
 


alexsohn1126:
@sentry review

src/Sentry.Extensions.AI/SentryChatClient.cs
```diff
@@ -0,0 +1,132 @@
...
+        {
+            return InstrumentStreamingResponseAsync(messages, options, outerSpan, initialSpan, cancellationToken);
+        }
+        catch (Exception ex)
+        {
+            initialSpan.Finish(ex);
+            outerSpan.Finish(ex);
+            _hub.CaptureException(ex);
+            throw;
+        }
```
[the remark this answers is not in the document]
jamescrosswell: True... the code returns a task, but doesn't await it so there won't be any exceptions.

src/Sentry.Extensions.AI/SentryChatClient.cs
```diff
@@ -0,0 +1,132 @@
...
+        outerSpan.Finish(SpanStatus.Ok);
+    }
+
+    /// <summary>
+    /// Starts a span or transaction based on whether there's an active transaction context.
+    /// </summary>
+    /// <param name="operation">The operation name</param>
+    /// <param name="description">The span/transaction description</param>
+    /// <returns>A child span of an existing transaction if available, else a new transaction</returns>
+    private ISpan StartSpanOrTransaction(string operation, string description)
```
jamescrosswell: I think the hub already contains a method to do this:
https://github.com/getsentry/sentry-dotnet/blob/eadf5f20b8014745d0fe42b0e707bf3f9eec9ffc/src/Sentry/HubExtensions.cs#L54-L62

alexsohn1126: Thank you!

src/Sentry.Extensions.AI/SentryChatClient.cs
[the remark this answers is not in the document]
alexsohn1126: https://github.com/getsentry/sentry-dotnet/blob/8502967ac091ce883ac951577003c3944432f0a4/src/Sentry.Extensions.AI/SentryChatClient.cs#L75-L110

I ended up getting the enumerator directly from `GetStreamingResponseAsync`. Then use `MoveNext()` so I can put everything other than `yield return` outside of the try-catch block.

src/Sentry.Extensions.AI/SentryChatClient.cs
```diff
@@ -0,0 +1,132 @@
...
+    {
+        var currentSpan = _hub.GetSpan();
+
+        if (currentSpan?.GetTransaction() != null)
+        {
+            return currentSpan.StartChild(operation, description);
+        }
+
+        var newTransaction = _hub.StartTransaction(description, operation);
```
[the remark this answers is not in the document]
alexsohn1126: I removed `StartSpanOrTransaction` in favor of `HubExtensions.StartSpan`:

https://github.com/getsentry/sentry-dotnet/blob/fe0968372da1fb4c3171f7af2e399707d52f4e8b/src/Sentry/HubExtensions.cs#L57

src/Sentry.Extensions.AI/SentryChatClient.cs
alexsohn1126: Fixed in https://github.com/getsentry/sentry-dotnet/pull/4657/commits/b08ec8ea8cb665995fe266dcd13177f5fac7d577

src/Sentry.Extensions.AI/SentryAISpanEnricher.cs
```diff
@@ -0,0 +1,158 @@
+using System.Text.Json;
+using Microsoft.Extensions.AI;
+
```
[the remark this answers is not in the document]
alexsohn1126: We should be fine without `using System` since we have a `GlobalUsings.cs`

alexsohn1126:
@sentry review

alexsohn1126:
@bugbot review

src/Sentry.Extensions.AI/SentryChatClient.cs
```diff
@@ -0,0 +1,142 @@
...
+            .GetStreamingResponseAsync(chatMessages, options, cancellationToken)
+            .GetAsyncEnumerator(cancellationToken);
+
+        while (true)
+        {
+            ChatResponseUpdate? current;
+
+            try
+            {
+                SentryAISpanEnricher.EnrichWithRequest(innerSpan, chatMessages, options, _sentryAIOptions);
```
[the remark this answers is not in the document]
alexsohn1126: Fixed in https://github.com/getsentry/sentry-dotnet/pull/4657/commits/a7bce63335fb9d52a9716a2123de366bf385c58e

src/Sentry.Extensions.AI/SentryInstrumentedFunction.cs
```diff
@@ -0,0 +1,47 @@
+using Microsoft.Extensions.AI;
+using Sentry.Extensibility;
+
+namespace Sentry.Extensions.AI;
+
+internal sealed class SentryInstrumentedFunction(AIFunction innerFunction, ChatOptions? aiOptions = null)
+    : DelegatingAIFunction(innerFunction)
+{
```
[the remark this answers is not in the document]
alexsohn1126: Fixed in https://github.com/getsentry/sentry-dotnet/pull/4657/commits/7f49d5471e36243a5bdadac5b3dca9954a145d1a

src/Sentry.Extensions.AI/SentryAISpanEnricher.cs
```diff
@@ -0,0 +1,158 @@
...
+    }
+
+    private static string FormatAvailableTools(IList<AITool> tools) =>
+        FormatAsJson(tools, tool => new { name = tool.Name, description = tool.Description });
+
+    private static string FormatRequestMessage(ChatMessage[] messages) =>
+        FormatAsJson(messages, message => new { role = message.Role, content = message.Text });
+
+    private static string FormatAsJson<T>(IEnumerable<T> items, Func<T, object> selector) =>
+        JsonSerializer.Serialize(items.Select(selector));
```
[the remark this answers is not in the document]
alexsohn1126: We have a `GlobalUsings.cs` file.

src/Sentry.Extensions.AI/SentryAISpanEnricher.cs
```diff
@@ -0,0 +1,158 @@
...
+}
```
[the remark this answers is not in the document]
alexsohn1126: We use `GlobalUsings.cs`

samples/Sentry.Samples.ME.AI.AspNetCore/Program.cs
```diff
@@ -0,0 +1,148 @@
...
+    options.SampleRate = 1;
+    options.TracesSampleRate = 1.0;
+    options.Experimental.EnableLogs = true;
+});
+
+var openAIClient = new OpenAI.Chat.ChatClient("gpt-4o-mini", Environment.GetEnvironmentVariable("OPENAI_API_KEY"))
+    .AsIChatClient()
+    .WithSentry(options =>
+    {
+        // In this case, we already initialized Sentry from ASP.NET WebHost creation, we don't need to initialize
```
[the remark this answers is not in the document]
alexsohn1126: Fixed in https://github.com/getsentry/sentry-dotnet/pull/4657/commits/a61f9c7dc44fe9ed32aed8701d478211e2c7e936

samples/Sentry.Samples.ME.AI.Console/Program.cs
```diff
@@ -0,0 +1,119 @@
...
+using Microsoft.Extensions.Logging;
+using Sentry.Extensions.AI;
+
+using var loggerFactory = LoggerFactory.Create(builder => builder.AddConsole());
+var logger = loggerFactory.CreateLogger<Program>();
+
+logger.LogInformation("Starting Microsoft.Extensions.AI sample with Sentry instrumentation");
+
+// Create OpenAI API client and wrap it with Sentry instrumentation
+var openAIClient = new OpenAI.Chat.ChatClient("gpt-4o-mini", Environment.GetEnvironmentVariable("OPENAI_API_KEY"))
```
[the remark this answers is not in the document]
alexsohn1126: Fixed in https://github.com/getsentry/sentry-dotnet/pull/4657/commits/a61f9c7dc44fe9ed32aed8701d478211e2c7e936

src/Sentry.Extensions.AI/SentryChatClient.cs
```diff
@@ -0,0 +1,142 @@
...
+            if (!ContainsFunctionCalls(response))
+            {
+                outerSpan.Finish(SpanStatus.Ok);
+                RootSpan = null;
+            }
+
+            return response;
+        }
+        catch (Exception ex)
+        {
```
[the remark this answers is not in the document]
alexsohn1126: Should be fixed in https://github.com/getsentry/sentry-dotnet/pull/4657/commits/136baf8102acba0eb550b9db46c3a833f5923640

alexsohn1126:
@bugbot review

alexsohn1126:
@sentry review

src/Sentry.Extensions.AI/SentryChatClient.cs
```diff
@@ -0,0 +1,141 @@
+using Microsoft.Extensions.AI;
+using Sentry.Extensibility;
+
+namespace Sentry.Extensions.AI;
+
+internal sealed class SentryChatClient : DelegatingChatClient
+{
+    private readonly HubAdapter _hub;
+    private readonly SentryAIOptions _sentryAIOptions;
+    internal static ISpan? RootSpan;
```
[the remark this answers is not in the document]
alexsohn1126: Hmm, I just tried wrapping RootSpan with `AsyncLocal`, and it resulted in something like this:
image

Compare it to the previous AI view we have which used normal `ISpan`:

image


A couple of problems:
- The tool spans are not marked as a child of the `invoke_agent` root span, because tool calls are invoked in a separate await call in `FunctionInvokingChatClient` (from `Microsoft.Extensions.AI` NOT `Abstractions`): [link](https://source.dot.net/#Microsoft.Extensions.AI/ChatCompletion/FunctionInvokingChatClient.cs,382).

- The outer spans (`invoke_agent` spans) are generated each time the LLM has to create a response after the tool call has finished.

I definitely do agree that 2 concurrent calls to `GetResponseAsync` could result in a wonky situation. I'll come up with a test to reproduce this issue and try to come up with a solution...

src/Sentry.Extensions.AI/SentryAISpanEnricher.cs
```diff
@@ -0,0 +1,158 @@
...
+    /// Enriches the span using the list of streamed in <see cref="ChatResponseUpdate"/>.
+    /// </summary>
+    /// <param name="span">span to enrich</param>
+    /// <param name="messages">a list of <see cref="ChatResponseUpdate"/></param>
+    /// <param name="aiOptions">AI-specific options</param>
+    public static void EnrichWithStreamingResponse(ISpan span, List<ChatResponseUpdate> messages, SentryAIOptions? aiOptions = null)
+    {
+        var inputTokenCount = 0L;
+        var outputTokenCount = 0L;
+        var finalText = new StringBuilder();
```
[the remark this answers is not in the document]
alexsohn1126: Dude, we have a `GlobalUsings.cs` which has `using System.Text.Json`, stop!!!

src/Sentry.Extensions.AI/SentryChatClient.cs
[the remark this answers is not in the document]
jamescrosswell: The transaction created by a call like this will be asynclocal:
https://github.com/getsentry/sentry-dotnet/blob/136baf8102acba0eb550b9db46c3a833f5923640/src/Sentry.Extensions.AI/SentryChatClient.cs#L126

See https://github.com/getsentry/sentry-dotnet/pull/3596

So I don't think there is any other option that won't break things.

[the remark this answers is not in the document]
alexsohn1126: This may also be an issue: https://github.com/dotnet/runtime/issues/47802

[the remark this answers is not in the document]
alexsohn1126: Should be solved in https://github.com/getsentry/sentry-dotnet/pull/4657/commits/a26501639fc28586d910d8aa44175e30825878a6

I "solved" this by following these steps:
- Added a concurrent dictionary that will map a `ChatMessage` instance to an `ISpan`. The span is a special span we call "agent span", which should be a parent to everything inside the initial `GetResponseAsync` call, including multiple chat spans, tool calls spans, etc..
  - This lives inside [`ChatOption.AdditionalProperties`](https://learn.microsoft.com/en-us/dotnet/api/microsoft.extensions.ai.chatoptions.additionalproperties?view=net-9.0-pp).

- For each `GetRequestAsync` call, we find whether we have the first message in a list of `ChatMessage` is already in the concurrent dictionary.
  - if so, we know this `GetRequestAsync` is a subsequent call from the original `ChatMessage`. Therefore we create an AI chat span as a child of the agent span
  - If not, this is our first time processing this list of `ChatMessage`. This is only true when the user code invokes `GetRequestAsync`, so we create a new agent span, and attach a new AI chat span to it.
  
- Once the inner `GetRequestAsync` is complete, we can then inspect the response and see if we have any tool calls
  - If there will be tool calls, then we inject the first `ChatMessage` from the request into the function arguments.

- When `SentryInstrumentedFunction` is initiated, it already has access to the same `ChatOptions` used in `SentryChatClient`. therefore we can get the concurrent dictionary.
  - We can get the agent span via the `ChatMessage` we injected into the function arguments. We make sure to remove this in order to keep the inner function calls working.
  - We then attach the tool call span to the agent span.

I do think this is a valid solution, but we can talk about it more if not.

alexsohn1126:
@bugbot review

src/Sentry.Extensions.AI/SentryChatClient.cs
```diff
@@ -0,0 +1,223 @@
...
+        }
+
+        // If we couldn't find the dictionary, we just initiate it now
+        agentSpanDict = new ConcurrentDictionary<ChatMessage, ISpan>();
+        if (options == null)
+        {
+            return agentSpanDict;
+        }
+
+        options.AdditionalProperties = new AdditionalPropertiesDictionary();
```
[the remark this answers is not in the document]
alexsohn1126: fixed in https://github.com/getsentry/sentry-dotnet/pull/4657/commits/717d191668f92f6a8f9907bcc8c5b74db6f04375

alexsohn1126:
@sentry review

alexsohn1126:
@bugbot review

src/Sentry.Extensions.AI/SentryChatClient.cs
```diff
@@ -0,0 +1,223 @@
...
+        _hub = HubAdapter.Instance;
+    }
+
+    /// <inheritdoc cref="IChatClient"/>
+    public override async Task<ChatResponse> GetResponseAsync(IEnumerable<ChatMessage> messages,
+        ChatOptions? options = null,
+        CancellationToken cancellationToken = new())
+    {
+        var chatMessages = messages as ChatMessage[] ?? messages.ToArray();
+        var keyMessage = chatMessages[0];
```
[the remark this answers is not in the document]
alexsohn1126: I thought this would be fine because literally [every](https://source.dot.net/#Microsoft.Extensions.AI.Abstractions/ChatCompletion/ChatClientExtensions.cs,a6136788f6488e7c) [single](https://source.dot.net/#Microsoft.Extensions.AI.Abstractions/ChatCompletion/ChatClientExtensions.cs,5110fa61ae95bd33) [over](https://source.dot.net/#Microsoft.Extensions.AI.Abstractions/ChatCompletion/DelegatingChatClient.cs,acfa90ee6b5d3aa6)[load](https://source.dot.net/#Microsoft.Extensions.AI.Abstractions/ChatCompletion/IChatClient.cs,5b5884d72f88f380) for `GetResponseAsync` either takes in a `ChatMessage` or creates one. Also it doesn't really make sense to send a `GetResponseAsync` without any messages.

But after thinking about it, I should actually add the check because who know what user code will be doing..

alexsohn1126:
@sentry review

alexsohn1126:
@bugbot review

src/Sentry.Extensions.AI/SentryAIActivityListener.cs
```diff
@@ -0,0 +1,39 @@
...
+        },
+    };
+
+    /// <summary>
+    /// Initializes Sentry's <see cref="ActivityListener"/> to tap into FunctionInvokingChatClient's Activity
+    /// </summary>
+    internal static void Init()
+    {
+        ActivitySource.AddActivityListener(FICCListener);
+    }
```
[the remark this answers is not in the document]
alexsohn1126: `AddIfNotExist` function in `ActivitySource.cs` doesn't allow multiple of the same activity listener to be registered: https://source.dot.net/#System.Diagnostics.DiagnosticSource/System/Diagnostics/ActivitySource.cs,5792c0e6425057ff

alexsohn1126:
@bugbot review

src/Sentry.Extensions.AI/SentryAIActivityListener.cs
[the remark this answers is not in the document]
alexsohn1126: `AddIfNotExist` function in `ActivitySource.cs` doesn't allow multiple of the same activity listener to be registered: https://source.dot.net/#System.Diagnostics.DiagnosticSource/System/Diagnostics/ActivitySource.cs,5792c0e6425057ff

src/Sentry.Extensions.AI/SentryChatClient.cs
```diff
@@ -0,0 +1,136 @@
...
+            catch (Exception ex)
+            {
+                innerSpan.Finish(ex);
+                _hub.CaptureException(ex);
+                throw;
+            }
+
+            yield return current;
+        }
+    }
```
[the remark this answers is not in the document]
alexsohn1126: Fixed this using `await using` on the enumerator, had to disable CA2007 because: https://github.com/dotnet/roslyn-analyzers/issues/5712

alexsohn1126:
@bugbot review

src/Sentry.Extensions.AI/SentryAIActivityListener.cs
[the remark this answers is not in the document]
alexsohn1126: `AddIfNotExist` function in `ActivitySource.cs` doesn't allow multiple of the same activity listener to be registered: https://source.dot.net/#System.Diagnostics.DiagnosticSource/System/Diagnostics/ActivitySource.cs,5792c0e6425057ff

alexsohn1126:
@sentry review

src/Sentry.Extensions.AI/SentryAIActivityListener.cs
```diff
@@ -0,0 +1,39 @@
...
+}
```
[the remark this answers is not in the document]
alexsohn1126: Like I said above, `Init` calls `AddIfNotExist` function in `ActivitySource.cs`, and it doesn't allow multiple of the same activity listener to be registered: https://source.dot.net/#System.Diagnostics.DiagnosticSource/System/Diagnostics/ActivitySource.cs,5792c0e6425057ff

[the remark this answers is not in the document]
alexsohn1126: See also:https://source.dot.net/#System.Diagnostics.DiagnosticSource/System/Diagnostics/ActivitySource.cs,348

src/Sentry.Extensions.AI/SentryChatClient.cs
```diff
@@ -0,0 +1,142 @@
...
+        try
+        {
+            configure?.Invoke(_sentryAIOptions);
+
+            if (_sentryAIOptions.InitializeSdk && !SentrySdk.IsEnabled)
+            {
+                SentrySdk.Init(_sentryAIOptions);
+            }
+        }
+        catch (Exception ex)
```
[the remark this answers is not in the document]
alexsohn1126: If we fail during SDK initialization, we should just throw the exception to the user to let them know it has failed.

src/Sentry.Extensions.AI/Sentry.Extensions.AI.csproj
```diff
@@ -0,0 +1,26 @@
...
+    <Using Include="Sentry.Extensibility" />
+    <Using Include="Sentry.Integrations" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="..\Sentry\Sentry.csproj" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Microsoft.Extensions.AI" Version="9.10.1" />
```
Flash0ver: question: Can we use `Microsoft.Extensions.AI.Abstractions` instead?

I checked the dependencies of the templates from `Microsoft.Extensions.AI.Templates`
- `aichatweb`
  - [Microsoft.Extensions.AI](https://www.nuget.org/packages/Microsoft.Extensions.AI)
- `mcpserver`
  - [ModelContextProtocol](https://www.nuget.org/packages/ModelContextProtocol/)
    - [ModelContextProtocol.Core](https://www.nuget.org/packages/ModelContextProtocol.Core/)
      - [Microsoft.Extensions.AI.Abstractions](https://www.nuget.org/packages/Microsoft.Extensions.AI.Abstractions/)

Now, I am uncertain if we actually could utilize this SDK package to also instrument .NET MCP-Servers ... or if we actually would need to create a `Sentry.ModelContextProtocol` package to provide a meaningful integration ... I need to test that out with an MCP-Server I still got on my disk ... unless you know the answer already 🤔.

src/Sentry.Extensions.AI/Sentry.Extensions.AI.csproj
```diff
@@ -0,0 +1,26 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFrameworks>net9.0;net8.0</TargetFrameworks>
```
Flash0ver: question: Do we want to support `netstandard2.0` too (since `Microsoft.Extensions.AI` does) to enable more scenarios, such as .NET Framework ... although I could imagine that usage might be neglectable.

src/Sentry.Extensions.AI/Extensions/SentryAIExtensions.cs
```diff
@@ -0,0 +1,51 @@
+using Microsoft.Extensions.AI;
+using Sentry.Extensibility;
+
+// ReSharper disable once CheckNamespace -- Discoverability
+namespace Sentry.Extensions.AI;
```
Flash0ver: suggestion: Move to `Microsoft.Extensions.AI` namespace

Both receivers, the [IChatClient Interface](https://learn.microsoft.com/dotnet/api/microsoft.extensions.ai.ichatclient) and the [ChatOptions Class](https://learn.microsoft.com/dotnet/api/microsoft.extensions.ai.chatoptions), are contained in the `Microsoft.Extensions.AI` namespace.

For discoverability, I suggest moving this type to the same namespace as well, so that the caller does not need to import an additional namespace for auto-completion.
I believe we do this for all of our `UseSentry` and `AddSentry` extension methods.

src/Sentry.Extensions.AI/Extensions/SentryAIExtensions.cs
```diff
@@ -0,0 +1,51 @@
...
+    /// </summary>
+    /// <remarks>
+    /// This method can be used either with an existing Sentry setup or as a standalone integration.
+    /// If Sentry is already initialized, it will use the existing configuration.
+    /// If not, it will initialize Sentry with the provided options.
+    /// </remarks>
+    /// <param name="client">The <see cref="IChatClient"/> to be instrumented</param>
+    /// <param name="configure">The <see cref="SentryAIOptions"/> configuration</param>
+    /// <returns>The instrumented <see cref="IChatClient"/></returns>
+    public static IChatClient WithSentry(this IChatClient client, Action<SentryAIOptions>? configure = null)
```
Flash0ver: question: naming `With`

I believe that for most of our "instrumentalizing" extension methods, we use `AddSentry` and `UseSentry`, especially within the context of Dependency-Injection and Host/App-Builders.

Is the verb `With` for Extension-Methods common in the context of AI in general,
and `IChatClient`/`ChatOptions` in particular?

Hmmm ... on 2nd thought ... since we're not acting on a Builder-type, and returning the same type that we receive ... perhaps `With` is indeed the proper verb ... reminds me of the Roslyn-APIs.

.generated.NoMobile.sln
```diff
@@ -209,6 +209,14 @@ Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Sentry.SourceGenerators.Tes
 EndProject
 Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Sentry.Maui.CommunityToolkit.Mvvm.Tests", "test\Sentry.Maui.CommunityToolkit.Mvvm.Tests\Sentry.Maui.CommunityToolkit.Mvvm.Tests.csproj", "{ADC91A84-6054-42EC-8241-0D717E4C7194}"
 EndProject
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Sentry.Samples.ME.AI.Console", "samples\Sentry.Samples.ME.AI.Console\Sentry.Samples.ME.AI.Console.csproj", "{CE591593-E51C-498E-BC0D-5083C8197544}"
+EndProject
```
Flash0ver, line 280: suggestion: automate generated/filtered solution changes via script

I believe we usually make manual changes to the `Sentry.sln` only, and then have all the other (Filtered) Solutions generated via `generate-solution-filters.ps1` and `generate-solution-filters-config.yaml`.

test/Sentry.Extensions.AI.Tests/Sentry.Extensions.AI.Tests.csproj
```diff
@@ -0,0 +1,17 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFrameworks>net9.0;net8.0</TargetFrameworks>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Microsoft.Extensions.AI.Abstractions" Version="9.10.1" />
```
Flash0ver, line 8: thought: inconsistent version bump

`Sentry.Extensions.AI` depends on `Microsoft.Extensions.AI`.
But here, we are bumping `Microsoft.Extensions.AI.Abstractions`, which is a transitive dependency of `Microsoft.Extensions.AI`.

For consistency, I suggest having the same package as in `Sentry.Extensions.AI.csproj`.
Or omitting it altogether if it's even the same version.
Related to https://github.com/getsentry/sentry-dotnet/pull/4657#discussion_r2491311311
Related to https://github.com/getsentry/sentry-dotnet/pull/4657#discussion_r2491468742

src/Sentry.Extensions.AI/Sentry.Extensions.AI.csproj
Flash0ver: question: Package Version

With the version, we usually try to go as low as feasible, to enable broadest support.

Is this version as restrictive as we need to go to enable meaningful support ... or could we lower the requirement to an older non-pre-release version to broaden support?

src/Sentry.Extensions.AI/Extensions/SentryAIExtensions.cs
```diff
@@ -0,0 +1,51 @@
...
+using Sentry.Extensibility;
+
+// ReSharper disable once CheckNamespace -- Discoverability
+namespace Sentry.Extensions.AI;
+
+/// <summary>
+/// Extensions to instrument Microsoft.Extensions.AI builders with Sentry
+/// </summary>
+[EditorBrowsable(EditorBrowsableState.Never)]
+public static class SentryAIExtensions
```
Flash0ver, line 10: thought: type name

Hmm ... I'm a bit uncertain about the type name.
We have extension methods receiving `ChatOptions` and `IChatClient`.
I guess we could split it into `ChatOptionsExtensions` and `ChatClientExtensions` types.
But on the other hand to group these in a single type and have a more general name is fine, too.

I guess I am realizing that I do prefer as-is ... but wanted to post my thoughts anyway if other's would like to chime in.

Flash0ver (commented):
Starting with "architectural" feedback and questions.

src/Sentry.Extensions.AI/Sentry.Extensions.AI.csproj
alexsohn1126: > Can we use Microsoft.Extensions.AI.Abstractions instead?

That is definitely my mistake... I meant to use the `Abstractions` package only!

>  I am uncertain if we actually could utilize this SDK package to also instrument .NET MCP-Servers

We do have a separate "MCP Instrumentation" for Python: [Sentry docs](https://docs.sentry.io/platforms/python/tracing/instrumentation/custom-instrumentation/mcp-module/). So it may be confusing if we were to put both MCP instrumentation and Agent instrumentation in the same place.

I can inspect what `ModelContextProtocol` does and how it's use to see if we can instrument that... If we can re-use the code in this PR to get the most of the things done, then that'd be the best... Whether we re-use the code from here or not, I think we need to create a separate `Sentry.ModelContextProtocol` package 😢

src/Sentry.Extensions.AI/Sentry.Extensions.AI.csproj
alexsohn1126: Sounds fair enough 👍

src/Sentry.Extensions.AI/Extensions/SentryAIExtensions.cs
alexsohn1126: I used `WithSentry` because I was wrapping the provided chat client with our chat client, so "With" made more sense at the time.

However, I think it's bad to have an outlier like this. I'll change it to `AddSentry`.

.generated.NoMobile.sln
alexsohn1126: I think I generated these using the PowerShell script. would that script automatically update the solution filters during the merge to main?

src/Sentry.Extensions.AI/Sentry.Extensions.AI.csproj
alexsohn1126: I see, I haven't considered that.

Upon further inspection, it seems like `9.7.0` is the earliest we can go without breaking current `SentryInstrumentedFunction.cs`.

I'll use `9.7.0` version of `Microsoft.Extensions.AI.Abstractions`!

src/Sentry.Extensions.AI/Extensions/SentryAIExtensions.cs
alexsohn1126: @jamescrosswell Any thoughts on this?

src/Sentry.Extensions.AI/SentryChatClient.cs
```diff
@@ -0,0 +1,163 @@
...
+        }
+
+        // If FunctionInvokingChatClient(FICC) wraps SentryChatClient, we should be able to get the agent span from the current activity
+        // The activity we attached the span to may be an ancestor of the current activity, we must search the parents for the span
+        var activeSpan = SentryAIUtil.GetActivitySpan();
+
+        // If we couldn't find the Activity, then FICC is not wrapping SentryChatClient. Return a new span from the hub
+        return activeSpan ?? _hub.StartSpan(SentryAIConstants.SpanAttributes.InvokeAgentOperation,
+            SentryAIConstants.SpanAttributes.InvokeAgentDescription);
+    }
```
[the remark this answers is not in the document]
alexsohn1126: Fixed in https://github.com/getsentry/sentry-dotnet/pull/4657/commits/d3421e06799eb1e5282dc659440bced9f4e4ba88

src/Sentry.Extensions.AI/Sentry.Extensions.AI.csproj
alexsohn1126: The PackageReference was fixed in https://github.com/getsentry/sentry-dotnet/pull/4657/commits/8d02f08e65895a1d75ab9113bb2350d4c2cc8dcf

src/Sentry.Extensions.AI/Sentry.Extensions.AI.csproj
alexsohn1126: Fixed in https://github.com/getsentry/sentry-dotnet/pull/4657/commits/adc3b9151ae5c15059fb785d334c085f515a5cb0

src/Sentry.Extensions.AI/Sentry.Extensions.AI.csproj
alexsohn1126: Updated in https://github.com/getsentry/sentry-dotnet/pull/4657/commits/1db751f54af44841ee64d62948e35a6c03ee195c

src/Sentry.Extensions.AI/Extensions/SentryAIExtensions.cs
alexsohn1126: Changed in https://github.com/getsentry/sentry-dotnet/pull/4657/commits/cc1e1a7f21300b4c89e84c516f66ff0c71af3063

src/Sentry.Extensions.AI/Extensions/SentryAIExtensions.cs
alexsohn1126: Changed in https://github.com/getsentry/sentry-dotnet/pull/4657/commits/0e0d2304c5c89022b8b2451247f3b5aa560cad9f

test/Sentry.Extensions.AI.Tests/Sentry.Extensions.AI.Tests.csproj
alexsohn1126: Fixed in https://github.com/getsentry/sentry-dotnet/pull/4657/commits/8d02f08e65895a1d75ab9113bb2350d4c2cc8dcf

samples/Sentry.Samples.ME.AI.AspNetCore/Program.cs
```diff
@@ -0,0 +1,193 @@
...
+if (openAiApiKey == null)
+{
+    throw new InvalidOperationException($"Environment variable for OpenAI API key '{varName}' is not set.");
+}
+
+var openAiClient = new OpenAI.Chat.ChatClient("gpt-4o-mini", openAiApiKey)
+    .AsIChatClient()
+    .AddSentry(options =>
+    {
+        // In this case, we already initialized Sentry from ASP.NET WebHost creation, we don't need to initialize
```
jamescrosswell: For the logging integrations, we have something like this (you can initialise Sentry via the extension methods we provide to configure Sentry logging). It has resulted in quite a lot of confusion... people don't know what parameters are required so they provide them all - including the DSN again - resulting in the SDK being reinitialised and losing the options that were configured for the first integration they configured).

It's a convenience really... it means if this is the only integration they're using, they don't need to also call `SentrySdk.Init`. It is a convenience that hands SDK users a foot gun though... so not sure if it's worth doing here.

How often do we think people will be using just our MS.AI integration, without also using one of the others (like logging, asp.net core, maui etc.). If it will be seldom, maybe we say you have to initialise the SDK separately and we don't have `SentryAIOptions` inherit from  `SentryOptions`. It will be slightly less convenient in those instances but could save a whole lot of confusion.

src/Sentry.Extensions.AI/Extensions/SentryAIExtensions.cs
jamescrosswell: We have a "generic" extension class [for the OTEL extensions](https://github.com/getsentry/sentry-dotnet/blob/54511b5c75f5f016e4c56940caf59cab1316ee2e/src/Sentry.OpenTelemetry/OpenTelemetryExtensions.cs)... admittedly I authored that class. But yeah, I'm OK lumping them together. If we did want to change this in the future, the change would be trivial and probably wouldn't require changes to user's code (only binary breaking really.. which we're sometimes OK with).

.generated.NoMobile.sln
jamescrosswell: It is run on CI and there is a dirty file check to make sure you did this locally and checked in your changes:
https://github.com/getsentry/sentry-dotnet/blob/10dbbae1d4c742a787caa3b4043304b472a2bbe6/.github/workflows/build.yml#L413-L415

Running that script would copy the `Sentry.sln` file over top of this one.

src/Sentry.Extensions.AI/Extensions/SentryAIExtensions.cs
```diff
@@ -0,0 +1,49 @@
...
+{
+    /// <summary>
+    /// Wrap tool calls specified in <see cref="ChatOptions"/> with Sentry agent instrumentation
+    /// </summary>
+    /// <param name="options">The <see cref="ChatOptions"/> that contains the <see cref="AIFunction"/> to instrument</param>
+    public static ChatOptions AddSentryToolInstrumentation(this ChatOptions options)
+    {
+        if (options.Tools is null || options.Tools.Count == 0)
```
jamescrosswell:
```suggestion
        if (options.Tools is not { Count: > 0 })
```

A bit more concise...

src/Sentry.Extensions.AI/SentryAIConstants.cs
```diff
@@ -0,0 +1,67 @@
+namespace Sentry.Extensions.AI;
+
+internal static class SentryAIConstants
+{
+    /// <summary>
```
jamescrosswell, line 5: Maybe worth adding a reference to the source code where we found these, in case they change in the future

src/Sentry.Extensions.AI/SentryAIActivityListener.cs
```diff
@@ -0,0 +1,39 @@
...
+    /// </summary>
+    private static readonly ActivityListener FICCListener = new()
+    {
+        ShouldListenTo = s => s.Name.StartsWith(SentryAIConstants.SentryActivitySourceName),
+        Sample = (ref ActivityCreationOptions<ActivityContext> options) =>
+            SentryAIConstants.FICCActivityNames.Contains(options.Name) ?
+                ActivitySamplingResult.AllDataAndRecorded : ActivitySamplingResult.None,
+        ActivityStarted = a =>
+        {
+            var agentSpan = HubAdapter.Instance.StartSpan(SentryAIConstants.SpanAttributes.InvokeAgentOperation, SentryAIConstants.SpanAttributes.InvokeAgentDescription);
```
jamescrosswell: The `HubAdapter` is basically a wrapper around the static `SentrySdk` class.

The idea is that your code here could accept an optional `IHub` argument to the `Init()` method... which would make it easy to mock this (otherwise static and hard to mock) class in unit tests. 

When no `IHub` is provided, your init method could use `HubAdapter.Instance` instead. 

tldr; this code probably shouldn't reference `HubAdapter.Instance` directly... it makes it hard to test.

src/Sentry.Extensions.AI/SentryAIActivitySource.cs
jamescrosswell, line 1: I don't think we need this... we're using the Diagnistic Source classes to listen in on the ME.AI telemetry, but we don't need to create any of our own (our own code generates Sentry spans directly).

src/Sentry.Extensions.AI/SentryAIConstants.cs
```diff
@@ -0,0 +1,67 @@
...
+    /// The list of strings which FunctionInvokingChatClient(FICC) uses to start the tool call <see cref="Activity"/>.
+    /// </summary>
+    internal static readonly string[] FICCActivityNames = ["orchestrate_tools", "FunctionInvokingChatClient.GetResponseAsync", "FunctionInvokingChatClient"];
+
+    /// <summary>
+    /// The string we use to identify our <see cref="ActivitySource"/>.
+    /// </summary>
+    internal const string SentryActivitySourceName = "Sentry.AgentMonitoring";
```
jamescrosswell: We should be able to get rid of this as well.

src/Sentry.Extensions.AI/SentryAISpanEnricher.cs
```diff
@@ -0,0 +1,216 @@
...
+        }
+
+        if (aiOptions?.AgentName is { } agentName)
+        {
+            span.SetData(SentryAIConstants.SpanAttributes.AgentName, agentName);
+        }
+
+        if (messages is { Length: > 0 }
+            && (aiOptions?.RecordInputs ?? true)
+            && !span.Data.TryGetValue(SentryAIConstants.SpanAttributes.RequestMessages, out _))
```
jamescrosswell: Since we discard the out param here, it's probably easier to use the [`ContainsKey` method](https://learn.microsoft.com/en-us/dotnet/api/system.collections.generic.ireadonlydictionary-2.containskey?view=netstandard-2.0) instead.

src/Sentry.Extensions.AI/SentryAISpanEnricher.cs
```diff
@@ -0,0 +1,216 @@
...
+            var content = message.Role == ChatRole.Tool ? FunctionCallToString(message.Contents) : message.Text;
+
+            return new
+            {
+                role = message.Role,
+                content
+            };
+        });
+
+        object FunctionCallToString(IList<AIContent> toolContents)
```
jamescrosswell:
```suggestion
        string FunctionCallToString(IList<AIContent> toolContents)
```

This should be a string right?

src/Sentry.Extensions.AI/SentryAISpanEnricher.cs
```diff
@@ -0,0 +1,216 @@
...
+
+                callList.Add(new
+                {
+                    call_id = functionResultContent.CallId,
+                    output = functionResultContent.Result,
+                    type = "function_call_output"
+                });
+            }
+
+            return callList;
```
jamescrosswell, line 186:
```suggestion
            return callList.ToString();
```

src/Sentry.Extensions.AI/SentryAISpanEnricher.cs
```diff
@@ -0,0 +1,216 @@
...
+                name = c.Name,
+                type = "function_call",
+                arguments = argumentsJson
+            };
+        });
+    }
+
+    private static string FormatAsJsonList<T>(IEnumerable<T> items, Func<T, object> selector)
+    {
+        return JsonSerializer.Serialize(items.Select(selector));
```
jamescrosswell: Personal preference... I don't think this method is doing enough to justify an extension method. I think it'd be simpler for the 3 call sites to just call `Select` and `JsonSerializer.Serialize` directly.

src/Sentry.Extensions.AI/SentryAIConstants.cs
```diff
@@ -0,0 +1,67 @@
...
+
+    /// <summary>
+    /// The string we use to retrieve the span from the <see cref="Activity"/> using a Fused property
+    /// </summary>
+    internal const string SentryActivitySpanAttributeName = "SentryCurrSpan";
```
jamescrosswell: Reading [the code](https://github.com/getsentry/sentry-dotnet/blob/9b72f4c545652c2a9a97ac1ebb3ab8443f6035e6/src/Sentry.Extensions.AI/SentryAIActivityListener.cs#L19-L29), it looks like this is used to store a reference to the span representing the FunctionInvokingChatClient... in which case I think we could give this a more specific name to make that clear.

```suggestion
    internal const string SentryFICCSpanAttributeName = "SentryFICCSpan";
```

src/Sentry.Extensions.AI/SentryAIUtil.cs
```diff
@@ -0,0 +1,22 @@
+using Sentry.Internal;
+
+namespace Sentry.Extensions.AI;
+
+internal static class SentryAIUtil
+{
+    internal static ISpan? GetActivitySpan()
```
jamescrosswell: We could rename this as well, I think... to make it clearer what it does.
```suggestion
    internal static ISpan? GetFICCSpan()
```

src/Sentry.Extensions.AI/SentryAIUtil.cs
```diff
@@ -0,0 +1,22 @@
+using Sentry.Internal;
+
+namespace Sentry.Extensions.AI;
+
+internal static class SentryAIUtil
```
jamescrosswell, line 5: Potentially this class could be renamed something like `FICCSpanTracker` and could encapsulate all the logic related to storing/retrieving the FICC span... that way none of the code outside this class would need to know implementation details like the `SentryAIConstants.SentryActivitySpanAttributeName` dictionary key (that could be kept private to this class) or that the Fused extensions are being used under the hood.

Not sure if you like that idea or not... maybe worth playing with to see if it results in tidier code.

src/Sentry.Extensions.AI/SentryChatClient.cs
```diff
@@ -0,0 +1,167 @@
...
+                responses.Add(current);
+            }
+            catch (Exception ex)
+            {
+                AfterResponseCleanup(chatSpan, agentSpan, ex);
+                throw;
+            }
+
+            yield return current;
+        }
```
jamescrosswell, line 109: Could maybe be made a bit more compact...

```suggestion
try
{
    while (await enumerator.MoveNextAsync())
    {
        var current = enumerator.Current;
        responses.Add(current);
        yield return current;
    }

    SentryAISpanEnricher.EnrichWithStreamingResponses(chatSpan, responses, _sentryAIOptions);
    SentryAISpanEnricher.EnrichWithStreamingResponses(agentSpan, responses, _sentryAIOptions);
    AfterResponseCleanup(chatSpan, agentSpan);
}
catch (Exception ex)
{
    AfterResponseCleanup(chatSpan, agentSpan, ex);
    throw;
}
```

src/Sentry.Extensions.AI/SentryInstrumentedFunction.cs
```diff
@@ -0,0 +1,57 @@
...
+                toolSpan.SetData(SentryAIConstants.SpanAttributes.ToolOutput, resultString);
+            }
+
+            toolSpan.Finish(SpanStatus.Ok);
+            return result;
+        }
+        catch (Exception ex)
+        {
+            toolSpan.Finish(SpanStatus.InternalError);
+            HubAdapter.Instance.CaptureException(ex);
```
jamescrosswell: Again, ideally we don't reference statics directly in our code. It would be better to inject an optional `IHub` via the constructor  (fall back to `HubAdapter.Instance` if none is provided) and reference that elsewhere in the code.

test/Sentry.Extensions.AI.Tests/SentryInstrumentedFunctionTests.cs
```diff
@@ -0,0 +1,214 @@
...
+
+namespace Sentry.Extensions.AI.Tests;
+
+public class SentryInstrumentedFunctionTests
+{
+    [Fact]
+    public async Task InvokeCoreAsync_WithValidFunction_ReturnsResult()
+    {
+        // Arrange
+        using var sentryDisposable = SentryHelpers.InitializeSdk();
```
jamescrosswell: If we can pass an `IHub` instance in via the constructor, it shouldn't be necessary to instantiate a real Hub to test this... same again for some of the other tests.

jamescrosswell (commented):
@alexsohn1126 overall looking pretty good 👍🏻 

I made lots of comments - mostly stuff around testability + some ideas about encapsulating some of the complexity related to the hack that we came up with that lets us capture the span for the FICC / agent.

And the main one: we definitely want this to target the `version6` branch.

samples/Sentry.Samples.ME.AI.AspNetCore/Program.cs
alexsohn1126: I was doing this because the logging also had the same feature.. But now that I think about it, people that will use MS.AI are probably using ASP.NET Core most likely. I'll remove the option to initialize SentrySdk, and let them know in the console sample that they need to initialize SentrySdk manually.

src/Sentry.Extensions.AI/SentryAIOptions.cs
```diff
@@ -0,0 +1,31 @@
+namespace Sentry.Extensions.AI;
+
+/// <summary>
+/// Sentry AI instrumentation options
+/// </summary>
+/// <inheritdoc />
+public class SentryAIOptions : SentryOptions
```
Flash0ver: suggestion: Experimental

To get it into the v6 release of the SDK, let's mark the public APIs as "Experimental".
We discussed different approaches:
- we could either wrap these properties in a "Experimental" property
- or ... when we no longer derive from `SentryOptions` and remove `InitializeSdk` ... we could also just prefix all three property names with "Experimental"

Additionally / optionally, we may want to put the `[Experimental]` attribute on the public methods of the public "Extensions"-Type.

test/Sentry.Extensions.AI.Tests/Sentry.Extensions.AI.Tests.csproj
```diff
@@ -0,0 +1,17 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFrameworks>net9.0;net8.0</TargetFrameworks>
```
Flash0ver: suggestion: "ApiApprovalTests"

In order to list and test the public API surface area, I suggest adding a `ApiApprovalTests.verify.cs`, similar to the other packages.

alexsohn1126:
Sorry for a lot of commit noises.. was trying to change the target to version6 :p

src/Sentry.Extensions.AI/SentryAIActivitySource.cs
alexsohn1126: I found that the source code for `FunctionInvokingChatClient` looks for the `ActivitySource` by using `innerClient.GetService<ActivitySource>`: [see here](https://source.dot.net/#Microsoft.Extensions.AI/ChatCompletion/FunctionInvokingChatClient.cs,90).

Without `_activitySource`, the activity will never be started, so I don't think we can just delete this...

src/Sentry.Extensions.AI/SentryAISpanEnricher.cs
alexsohn1126: Actually, it should be an object. That function name is very misleading.. I'll change that.

If we convert the list to a string, it shows up in the frontend like this:

image

So we have to let `JsonSerializer` to parse it correctly to make it show up like this:

image

src/Sentry.Extensions.AI/SentryAISpanEnricher.cs
alexsohn1126: See above, returning a string won't cause the JSON to render properly in the frontend

jamescrosswell:
> Sorry for a lot of commit noises.. was trying to change the target to version6 :p

In case it helps, I had to do the same for the isolated-cache branch... which I did by [merging version6 into that branch](https://github.com/getsentry/sentry-dotnet/pull/4498/commits/efaa83ed619e9f1792f794155efccdfda4a6233e) - was pretty painless.

src/Sentry.Extensions.AI/SentryAIUtil.cs
alexsohn1126: Could be worth the effort.. We don't use this class in a lot of places, only during `GetResponseAsync`, `GetStreamingResponseAsync` (in `SentryChatClient`), and `InvokeAsync` (in `SentryInstrumentedFunction`).

I think the code may be cleaner in `SentryAIActivityListener` after the refactoring. I'll try it out and see whether it is worth doing this minor refactoring.

alexsohn1126:
Yea, I tried rebasing and it ended up including every commit from version6 at one point 🤣... It's good now so I'll leave it alone, but that is good to know in the future!

src/Sentry.Extensions.AI/SentryChatClient.cs
alexsohn1126: That would indeed be cleaner... But `yield return` is not allowed inside a try-catch block: [link](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/statements/yield).

> `yield return` and `yield break` can not be used in [catch](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/statements/exception-handling-statements) and [finally](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/statements/exception-handling-statements) blocks, or in [try](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/statements/exception-handling-statements) blocks with a corresponding catch block. The yield return and yield break statements can be used in a try block with no catch blocks, only a finally block.

There seems to be some kind of work around making this work, but the PR is 2 years old 🙃 [link](https://github.com/dotnet/csharplang/discussions/765).

I do hate having `while(true)` in a production code, but this seems to be the only way of putting `MoveNextAsync` inside the try-catch block AFAIK...

src/Sentry.Extensions.AI/Extensions/SentryAIExtensions.cs
alexsohn1126: Fixed in https://github.com/getsentry/sentry-dotnet/pull/4657/commits/f2096bb3b3bad4a33248827ec628f880b995626c

src/Sentry.Extensions.AI/SentryAIConstants.cs
alexsohn1126: Updated in https://github.com/getsentry/sentry-dotnet/pull/4657/commits/88f0095c00e4d13dc429f3ed47d257d85071be97

src/Sentry.Extensions.AI/SentryAIActivityListener.cs
alexsohn1126: Fixed in https://github.com/getsentry/sentry-dotnet/pull/4657/commits/fd0ccbaada417317be80da7f61e4a24fcaa3bf5f

src/Sentry.Extensions.AI/SentryAISpanEnricher.cs
alexsohn1126: Fixed in https://github.com/getsentry/sentry-dotnet/pull/4657/commits/5e43b5bc677aa5e63401ea28aa2b22ed790e5148

src/Sentry.Extensions.AI/SentryAIActivityListener.cs
```diff
@@ -0,0 +1,43 @@
...
+    };
+
+    /// <summary>
+    /// Initializes Sentry's <see cref="ActivityListener"/> to tap into FunctionInvokingChatClient's Activity
+    /// </summary>
+    /// <param name="hub">Optional IHub instance to use. If not provided, HubAdapter.Instance will be used.</param>
+    internal static void Init(IHub? hub = null)
+    {
+        Hub = hub ?? HubAdapter.Instance;
+        ActivitySource.AddActivityListener(FICCListener);
```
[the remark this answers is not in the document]
alexsohn1126: False, `AddActivityListener` doesn't allow multiple listeners to be registered. We also have a test set up to ensure we don't send duplicate spans:

https://github.com/getsentry/sentry-dotnet/blob/274f7e549e187bf62d9589aa3a0e7db5683c3ea0/test/Sentry.Extensions.AI.Tests/SentryAIActivityListenerTests.cs#L90-L108

See also:
https://source.dot.net/#System.Diagnostics.DiagnosticSource/System/Diagnostics/ActivitySource.cs,348

src/Sentry.Extensions.AI/SentryAISpanEnricher.cs
alexsohn1126: Fixed: https://github.com/getsentry/sentry-dotnet/pull/4657/commits/faef263348561cd62dc9b16cf8a907ce2c55d979

src/Sentry.Extensions.AI/SentryAIConstants.cs
alexsohn1126: Fixed in https://github.com/getsentry/sentry-dotnet/pull/4657/commits/b4dc1687fd5a95a6334eca0f9b793ef671c7fe2e

src/Sentry.Extensions.AI/SentryAIUtil.cs
alexsohn1126: Fixed in https://github.com/getsentry/sentry-dotnet/pull/4657/commits/b4dc1687fd5a95a6334eca0f9b793ef671c7fe2e

src/Sentry.Extensions.AI/SentryInstrumentedFunction.cs
alexsohn1126: Fixed in https://github.com/getsentry/sentry-dotnet/pull/4657/commits/a2a89b313a9f74f8399d64ee6a157cef8f91f825

test/Sentry.Extensions.AI.Tests/SentryInstrumentedFunctionTests.cs
alexsohn1126: Fixed in https://github.com/getsentry/sentry-dotnet/pull/4657/commits/7d9365727232fae67382306333bef4336ea54996

samples/Sentry.Samples.ME.AI.AspNetCore/Program.cs
alexsohn1126: Fixed in https://github.com/getsentry/sentry-dotnet/pull/4657/commits/d13640c950d21a01d01b12a38816e19768079982

src/Sentry.Extensions.AI/SentryAIActivityListener.cs
```diff
@@ -0,0 +1,43 @@
...
+    }
```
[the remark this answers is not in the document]
alexsohn1126: False, `AddActivityListener` doesn't allow multiple listeners to be registered. We also have a test set up to ensure we don't send duplicate spans:

https://github.com/getsentry/sentry-dotnet/blob/274f7e549e187bf62d9589aa3a0e7db5683c3ea0/test/Sentry.Extensions.AI.Tests/SentryAIActivityListenerTests.cs#L90-L108

See also:
https://source.dot.net/#System.Diagnostics.DiagnosticSource/System/Diagnostics/ActivitySource.cs,348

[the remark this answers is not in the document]
alexsohn1126: dude are you serious i explained this like 3 times

src/Sentry.Extensions.AI/SentryAIOptions.cs
alexsohn1126: I marked `SentryAIOptions` both with the prefix `Experimental` and with the attribute, and the public methods with just the [experimental] attribute.

Updated in https://github.com/getsentry/sentry-dotnet/pull/4657/commits/45735c59a2f6694228a6154b466ea64b5ad0f2da

test/Sentry.Extensions.AI.Tests/Sentry.Extensions.AI.Tests.csproj
alexsohn1126: Updated in https://github.com/getsentry/sentry-dotnet/pull/4657/commits/cffccf937058ed5de0c1e65fa33e7d2ef559e108

I also changed the `TargetFrameworks` to use `CurrentTfms` like other tests 👍

src/Sentry.Extensions.AI/Extensions/SentryAIExtensions.cs
```diff
@@ -0,0 +1,59 @@
...
+    /// Wrap tool calls specified in <see cref="ChatOptions"/> with Sentry agent instrumentation
+    /// </summary>
+    /// <remarks>
+    /// This API is experimental, and it may change in the future.
+    /// </remarks>
+    /// <param name="options">The <see cref="ChatOptions"/> that contains the <see cref="AIFunction"/> to instrument</param>
+    [Experimental(DiagnosticId.ExperimentalFeature)]
+    public static ChatOptions AddSentryToolInstrumentation(this ChatOptions options)
+    {
+        if (options.Tools is { Count: 0 })
```
jamescrosswell:
```suggestion
        if (options.Tools is not { Count: > 0 })
```

The above basically reads:
> If `options.Tools` is not a non-null object with a `Count` greater than zero

The logic you have currently fails to match null values... it only matches a non-null object with a count of zero. That would technically be OK since the loop comparison below would compare `i` to `null` and the whole thing would evaluate as `false`... but:
- We can early exit here already (and avoid that comparison)
- It's not as obvious/explicit that will happen... anyone reading the code needs to do a bit of a double take to work that out

src/Sentry.Extensions.AI/SentryAIActivityListener.cs
```diff
@@ -0,0 +1,43 @@
...
+internal static class SentryAIActivityListener
+{
+    private static volatile IHub Hub = HubAdapter.Instance;
+
+    /// <summary>
+    /// Sentry's <see cref="ActivityListener"/> to tap into function invocation's Activity
+    /// </summary>
+    private static readonly ActivityListener FICCListener = new()
+    {
+        ShouldListenTo = source => source.Name.StartsWith(SentryAIConstants.SentryActivitySourceName),
```
jamescrosswell: I was expecting the instrumentation for the spans that ME AI adds to be done using this source name:
https://github.com/dotnet/extensions/blob/e7423719f55472f16e61150413fd107ea504c369/src/Libraries/Microsoft.Extensions.AI/OpenTelemetryConsts.cs#L11

... but even that we might not be able to rely on, since it appears people can override the source name for some reason:
https://github.com/dotnet/extensions/blob/f8f779a6ea004bb1f26649719ca77d63a9d9417c/src/Libraries/Microsoft.Extensions.AI/ChatCompletion/OpenTelemetryChatClientBuilderExtensions.cs#L29

Is that why you've overriden `SentryChatClient.GetService` here:
https://github.com/getsentry/sentry-dotnet/blob/d13640c950d21a01d01b12a38816e19768079982/src/Sentry.Extensions.AI/SentryChatClient.cs#L104-L107

It's an interesting solution... I'm wondering if there might be side effects, like if other people's code is also listening for those events and assuming they will have the original ActivitySource name. Given that it can be overriden, maybe they shouldn't be doing that anyway. 

@Flash0ver thoughts?

src/Sentry.Extensions.AI/SentryAIConstants.cs
```diff
@@ -0,0 +1,68 @@
+namespace Sentry.Extensions.AI;
+
+internal static class SentryAIConstants
+{
+    /// <summary>
+    /// The list of strings which FunctionInvokingChatClient(FICC) uses to start the tool call <see cref="Activity"/>.
+    /// These can be found in FunctionInvokingChatClient.cs in GetResponseAsync or GetStreamingResponseAsync function.
```
jamescrosswell: I was thinking more a permalink like:

```suggestion
    /// See:
    /// https://github.com/dotnet/extensions/blob/f8f779a6ea004bb1f26649719ca77d63a9d9417c/src/Libraries/Microsoft.Extensions.AI/ChatCompletion/FunctionInvokingChatClient.cs#L272
    /// 
```

I couldn't find anything using the names "FunctionInvokingChatClient.GetResponseAsync" or 
"FunctionInvokingChatClient"... I think you mentioned those came from older versions of the code? Would be good to add permalinks to those as well in that case.

src/Sentry.Extensions.AI/SentryAIActivityListener.cs
jamescrosswell, line 1: The point of passing `IHub` in as a dependency is to avoid statics (which are tricky to test). 

We should remove the `static` modifier from this class... can probably ditch the `Init` method in favour of a regular constructor as well.

If there was some need to only have one instance (strictly), we could implement a singleton pattern here ([like this](https://github.com/getsentry/sentry-dotnet/blob/90bf268ea0f3f1743525241ea1d742dd35c5588b/src/Sentry/Internal/NoOpSpan.cs#L10)).

I don't think that's necessary though...

src/Sentry.Extensions.AI/SentryAIActivityListener.cs
jamescrosswell: Sorry, that wasn't clear... see further [comment](https://github.com/getsentry/sentry-dotnet/pull/4657/files#r2512403111).

src/Sentry.Extensions.AI/SentryAIActivitySource.cs
jamescrosswell: Ah, I see... MS have made everything composable using wrappers... so unless [OpenTelemetry had been configured](https://github.com/dotnet/extensions/blob/f8f779a6ea004bb1f26649719ca77d63a9d9417c/src/Libraries/Microsoft.Extensions.AI/ChatCompletion/OpenTelemetryChatClientBuilderExtensions.cs#L26-L39) when building the chat client, no activity source gets added and there's no instrumentation.

Makes sense... and means we definitely need our own source. 👍🏻

src/Sentry.Extensions.AI/SentryAIActivityListener.cs
jamescrosswell: Ah... it starts to make more sense now I read this:
- https://github.com/getsentry/sentry-dotnet/pull/4657#discussion_r2512425211

So yeah, this looks like a nice solution.

src/Sentry.Extensions.AI/SentryChatClient.cs
jamescrosswell: Dang. @Flash0ver any ideas or do we leave as is?

samples/Sentry.Samples.ME.AI.AspNetCore/Program.cs
```diff
@@ -0,0 +1,192 @@
...
```
jamescrosswell:
```suggestion
```

N/A I think... now initialisation is no longer available.

src/Sentry.Extensions.AI/SentryAIConstants.cs
alexsohn1126: Yea, I was mentioning that from the official Microsoft Extensions AI codebase. A link would definitely be better..

samples/Sentry.Samples.ME.AI.AspNetCore/Program.cs
alexsohn1126: Fixed in https://github.com/getsentry/sentry-dotnet/pull/4657/commits/b99c53f255c2e60b1298dad22d40165798efd0a9

src/Sentry.Extensions.AI/SentryAIActivityListener.cs
alexsohn1126: I ended up using a singleton pattern because we only want one Listener registered at all times. Otherwise there may be multiple spans created for a single call to the LLMs.

Fixed in https://github.com/getsentry/sentry-dotnet/pull/4657/commits/9b440a336b5dd35987475ca584ba2269d8921e8e

src/Sentry.Extensions.AI/SentryAIConstants.cs
alexsohn1126: Fixed in https://github.com/getsentry/sentry-dotnet/pull/4657/commits/5745c0a4e6f655e0a6020956c07e0ceeab0ae45a

src/Sentry.Extensions.AI/SentryAIActivityListener.cs
```diff
@@ -0,0 +1,55 @@
...
+            },
+            ActivityStopped = activity =>
+            {
+                var agentSpan = activity.GetFused<ISpan>(SentryAIConstants.SentryFICCSpanAttributeName);
+                // Don't pass in OK status in case there was an exception
+                agentSpan?.Finish();
+            }
+        };
+        ActivitySource.AddActivityListener(Instance);
+    }
```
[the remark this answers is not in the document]
alexsohn1126: Resolved in https://github.com/getsentry/sentry-dotnet/pull/4657/commits/8fcebc901c0bc9d5663606b4ca47e539e565a908

src/Sentry.Extensions.AI/Extensions/SentryAIExtensions.cs
alexsohn1126: Fixed  in https://github.com/getsentry/sentry-dotnet/pull/4657/commits/ac9f7ceb1f22f3e0d325f7314f830c596cbb4018

alexsohn1126:
Seems like this is close to being finished.. (also bugbot didn't find anything so that's a plus? :) )

Please take a look @jamescrosswell @Flash0ver and let me know if anything's off..

src/Sentry.Extensions.AI/SentryAIActivityListener.cs
jamescrosswell: Probably easier to describe what I mean in a commit 😉 
- https://github.com/getsentry/sentry-dotnet/pull/4657/commits/fc64d02fcfe3bfbbd66a07e26847cb3080bc6238

Take a look at that and see if it makes sense... it's a bit tricky because the only way to add a listener is via `ActivitySource.AddActivityListener` (which is static), but possible to avoid all the other static references when running tests.

alexsohn1126: That looks good, using `Lazy` is something I haven't seen yet, that's interesting!

I changed the capitalization for AI in `SentryAIActivityListener` in this commit:
- https://github.com/getsentry/sentry-dotnet/pull/4657/commits/1bb5629df1fcaa33091c04a0a216beb7ef78791b

jamescrosswell (approved):
Nice work @alexsohn1126 - not a trivial integration so great to get this over the line!

PR #4662, opened 2025-10-21
feat: Added `SentrySdk.IsSessionActive`

bitsandfoxes:
Unity SDK integrations pause and resume sessions in response to lifecycle events without knowing if a session is active. This causes repeated warnings from GlobalSessionManager ("Attempted to pause/resume a session, but a session has not been started") with no way for integrations to prevent them.

https://github.com/getsentry/sentry-dotnet/blob/fe0968372da1fb4c3171f7af2e399707d52f4e8b/src/Sentry/GlobalSessionManager.cs#L298

This PR adds `IHub.IsSessionActive` to allow checking session state before pause/resume operations. This also gives users insights into whether there currently is a session active. I don't know how else they are supposed to deal any manual session handling anyway.


src/Sentry/Internal/Hub.cs
```diff
@@ -35,6 +35,8 @@ internal class Hub : IHub, IDisposable
 
     public bool IsEnabled => _isEnabled;
 
+    public bool IsSessionActive => _sessionManager.IsSessionActive;
```
jamescrosswell, line 38: @bitsandfoxes I'm wondering if what I'm building here might break this (and possibly other stuff):
https://github.com/getsentry/sentry-dotnet/pull/4664

Can Unity apps target iOS?

If so, I might need to build an `IosSessionManager` implementation for `ISessionManager`... which would be a wrapper around the CocoaSdk session manager.

Thoughts?

jamescrosswell: Although, I'm also guessing Session Replay isn't a thing in Unity... so maybe a non issue for this particular change.

bitsandfoxes: Unity targets iOS, but sessions are entirely kept and controlled on the C# side of things to keep as close to the user's code as possible. The .NET SDK already allows to pause and resume sessions, I don't see how allowing to check whether a session is active would break things.

PR #4559, opened 2025-09-24
iOS & Android integration tests

jpnurmi:
Helps verify that crash events are captured as expected after Cocoa and Java SDK updates (and eventually getsentry/sentry-cocoa#6193):
- `CrashType.Managed`
- `CrashType.Native`
- `NullReferenceException`

For example, see
- [51240026268](https://github.com/getsentry/sentry-dotnet/actions/runs/18010067972/job/51240026268?pr=4559) (iOS Device Tests -> Run Integration Tests)
- [51593510741](https://github.com/getsentry/sentry-dotnet/actions/runs/18129644787/job/51593510741?pr=4559) (Android Device Tests -> Run Integration Tests)

#skip-changelog

---

> [!NOTE]
> Adds MAUI-based Android and iOS integration tests executed on emulators/simulators in CI, with supporting test app, utilities, and workflow updates (including retries and result artifacts).
> 
> - **CI/Workflows**:
>   - Android: run new integration tests on emulator via `integration-test/android.Tests.ps1` with retry; checkout submodules; set up Java 17; fetch shared `github-workflows`; reuse emulator config via YAML anchors; upload both `test_output` and `integration-test/mobile-app/test_output`.
>   - iOS: add integration tests via `integration-test/ios.Tests.ps1` with retry; upload both result paths.
>   - Build: include `scripts` in sparse checkout for publishing builds.
> - **Tests/Infrastructure**:
>   - Add MAUI test app (`integration-test/net9-maui/**`) targeting `net9.0-android35.0` and `net9.0-ios18.0` to drive crash scenarios and verify envelopes.
>   - New Pester test suites: `integration-test/android.Tests.ps1`, `integration-test/ios.Tests.ps1`; shared helpers in `integration-test/pester.ps1`.
>   - Utilities: new `scripts/device-test-utils.ps1` (XHarness install, emulator/simulator helpers); `scripts/device-test.ps1` now uses these helpers and simplifies simulator selection.
>   - Refactor `integration-test/common.ps1` to source shared Pester extensions.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 8144e871a79e9d103d41d7a9546c2d2578e8f0de. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

PR #4666, opened 2025-10-22
test: run mobile integration tests in Release and Debug modes

jpnurmi:
This PR extends the recently added [mobile integration tests](https://github.com/getsentry/sentry-dotnet/pull/4559) by adding `Debug` configuration alongside the existing `Release` configuration, to get a better understanding of the impact of:

- https://github.com/getsentry/sentry-native/pull/1392
- https://github.com/getsentry/sentry-cocoa/pull/6193

Notes:
- Previously, the Android integration test rebuilt the app for every test function to be able to replace `{{SENTRY_DSN}}` in `MauiProgram.cs`. With an increasing number of test functions and variations, the integration test becomes slower and slower. Thus, I changed the integration test app to simply default to `http://127.0.0.1:8000`, which is the [default](https://github.com/getsentry/github-workflows/blob/0ae70c0a594b55a6d28db29b82dd4a658d61c8a5/sentry-cli/integration-test/action.psm1#L5) in the test server anyway. This way, we only need to build the Android app once with `Release` and `Debug` configurations, instead of 2x4=8 times.
- The test app was fixed to not auto-restart on Android. Unexpected/accidental auto-restarts caused prolonged "Waiting for app..." loops and extra envelopes.
- Client reports and automatic session tracking were disabled to ensure a predictable number of crash envelopes for test assertion purposes.

### DIFF
--- a/integration-test/net9-maui/MauiProgram.cs
+++ b/integration-test/net9-maui/MauiProgram.cs
@@ -12,7 +12,9 @@ public static class MauiProgram
             .UseMauiApp<App>()
             .UseSentry(options =>
             {
-                options.Dsn = System.Environment.GetEnvironmentVariable("SENTRY_DSN") ?? "http://key@127.0.0.1:8000/0";
+#if ANDROID
+                options.Dsn = "{{SENTRY_DSN}}";
+#endif
                 options.Debug = false;
                 options.DiagnosticLevel = SentryLevel.Error;
                 // predictable crash envelopes only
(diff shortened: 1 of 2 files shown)


integration-test/net9-maui/MauiProgram.cs
```diff
@@ -12,11 +12,12 @@ public static MauiApp CreateMauiApp()
             .UseMauiApp<App>()
             .UseSentry(options =>
             {
-#if ANDROID
-                options.Dsn = "{{SENTRY_DSN}}";
-#endif
+                options.Dsn = System.Environment.GetEnvironmentVariable("SENTRY_DSN") ?? "http://key@127.0.0.1:8000/0";
```
jamescrosswell: Is there a scenario where we would want to use a real DSN? That would send events to Sentry...

jamescrosswell (approved):
Had one small question but generally looks good!

integration-test/net9-maui/MauiProgram.cs
jpnurmi: Good point. There's a chance the upcoming e2e tests could reuse the same test app, so it makes sense to leave that door open: https://github.com/getsentry/sentry-dotnet/pull/4666/commits/541de2fc4f7d29db8d06374a7a50c86683901a5c

Issue #3954, opened 2025-02-10
Prevent Native/Cocoa SDKs from capturing errors that also get captured by .NET


jamescrosswell:
This would be the "long term solution" described in https://github.com/getsentry/sentry-dotnet/issues/3902 and the permanent resolution to the errors described in there:
- https://github.com/getsentry/sentry-dotnet/issues/3902

However it likely requires support from the Native team and we don't currently have capacity to work on this (see https://github.com/getsentry/sentry-dotnet/issues/3861#issuecomment-2586751817).

jurganson:
I'm facing an issue with duplicate logs in our Sentry implementation. While I've disabled native reports using options available in Sentry 5.3.0, my app still experiences significant native crashes that contain valuable information.

As a temporary workaround, I attempted to filter out unhelpful crashes by checking stack frames in `BeforeSend`, specifically targeting `signalhandler` mechanism crashes with no functions. However, during testing, I noticed that none of these native crashes went through my `BeforeSend` function.

Is this expected behavior for native crashes to bypass `BeforeSend`? I was testing on Sentry version 5.1.0, which may be relevant.

jamescrosswell:
> Is this expected behavior for native crashes to bypass `BeforeSend`? I was testing on Sentry version 5.1.0, which may be relevant.

Support for native BeforeSend was introduced for iOS ([#3958](https://github.com/getsentry/sentry-dotnet/pull/3958))  in [Version 5.3.0](https://github.com/getsentry/sentry-dotnet/blob/main/CHANGELOG.md#530).

This has been available on Android since well before version 5.0.0 but is guarded by a flag in the options:
https://github.com/getsentry/sentry-dotnet/blob/a621f42332c8adca362456544ea35069f50fe733/src/Sentry/Platforms/Android/NativeOptions.cs#L262

There's another option for Android that you can use to suppress SIGSEGV native errors (`options.Android.SuppressSegfaults`). Enabling that though will suppress native exceptions that are otherwise/already reported as a managed `NullRefrenceException`, however it _can also_ suppress SIGSEGV errors unrelated to managed code (in which case you wouldn't know about quite serious bugs that were crashing your app).

arek-corp:
Does the feature implemented in 5.5.0 for IScopeObserver help in this? 

PR:
#4026

jamescrosswell:
> Does the feature implemented in 5.5.0 for IScopeObserver help in this?

[...]

Issue #4660, opened 2025-10-21
CI: run Android integration tests on ARM64


jpnurmi:
### Description

There are minor differences in Mono and native crash handling between `x86_64` vs. `arm64`. Currently, we only run Android integration tests on `ubuntu-latest-4-cores`, which is `x86_64`. We should add `arm64` if/when possible.

Problems:
- `macos-xxx`: M1/M2 do not support nested virtualization :x:
- `ubuntu-xxx-arm`: Android Emulator not available on Linux ARM64 :x:

jamescrosswell:
@jpnurmi I take it this is a suggestion for some time in the future then (not something we can action immediately)?

jpnurmi:
Indeed, it might not be doable for the time being
- https://github.com/actions/runner-images/issues/9472

PR #4676, opened 2025-10-27
fix: prevent redundant native exceptions on Android/Mono

jpnurmi:
Experimental opt-in Android-only solution for:
- #3954

According to our newly introduced Android integration tests, `CHAIN_AT_START` works on `android-arm64` and `android-x64` in both `Release` and `Debug` configurations, but we'd like to validate the fix further in real-world conditions. To that end, we're making it opt-in initially so customers can try it on more devices, platforms, and configurations before considering it as the new default.

## Changelog Entry

- fix: prevent redundant native exceptions on Android/Mono
  - Note: opt in by setting `options.Native.ExperimentalOptions.SignalHandlerStrategy` to `Sentry.Android.SignalHandlerStrategy.ChainAtStart`

### DIFF
--- a/src/Sentry/Platforms/Android/NativeOptions.cs
+++ b/src/Sentry/Platforms/Android/NativeOptions.cs
@@ -273,10 +273,10 @@ public partial class SentryOptions
             /// The default value is <see cref="Android.SignalHandlerStrategy.Default"/>.
             /// </summary>
             /// <remarks>
-            /// .NET runtimes 10.0.0–10.0.3 (.NET SDKs 10.0.100–10.0.301) are not compatible with
+            /// .NET runtimes 10.0.0–10.0.3 (.NET SDKs 10.0.100–10.0.103) are not compatible with
             /// <see cref="Android.SignalHandlerStrategy.ChainAtStart"/>. On affected versions,
             /// the SDK automatically falls back to <see cref="Android.SignalHandlerStrategy.Default"/>.
-            /// The issue was resolved in .NET runtime 10.0.4 (.NET SDK 10.0.400). See
+            /// The issue was resolved in .NET runtime 10.0.4 (.NET SDK 10.0.200). See
             /// <see href="https://github.com/dotnet/runtime/pull/123346">dotnet/runtime#123346</see>.
             /// </remarks>
             public SignalHandlerStrategy SignalHandlerStrategy { get; set; } = SignalHandlerStrategy.Default;
--- a/src/Sentry/Platforms/Android/NativeOptions.cs
+++ b/src/Sentry/Platforms/Android/NativeOptions.cs
@@ -157,7 +157,7 @@ public partial class SentryOptions
         /// <summary>
         /// Gets or sets the strategy for how Sentry Native's signal handler interacts with the CLR/Mono
         /// signal handler.
-        /// The default value is <c>SignalHandlerStrategy.Default</c> (enabled).
+        /// The default value is <see cref="Android.SignalHandlerStrategy.Default"/>.
         /// </summary>
         /// <remarks>
         /// .NET runtimes 10.0.0–10.0.3 (.NET SDKs 10.0.100–10.0.301) are not compatible with
--- a/src/Sentry/Platforms/Android/NativeOptions.cs
+++ b/src/Sentry/Platforms/Android/NativeOptions.cs
@@ -161,8 +161,8 @@ public partial class SentryOptions
         /// </summary>
         /// <remarks>
         /// .NET runtimes 10.0.0–10.0.3 (.NET SDKs 10.0.100–10.0.301) are not compatible with
-        /// <see cref="Android.SignalHandlerStrategy.ChainAtStart"/>. On affected versions, the
-        /// SDK automatically falls back to <see cref="Android.SignalHandlerStrategy.Default"/>.
+        /// <see cref="Android.SignalHandlerStrategy.ChainAtStart"/>. Using it on affected
+        /// versions throws an <see cref="InvalidOperationException"/> during initialization.
         /// The issue was resolved in .NET runtime 10.0.4 (.NET SDK 10.0.400). See
         /// <see href="https://github.com/dotnet/runtime/pull/123346">dotnet/runtime#123346</see>.
         /// </remarks>
--- a/src/Sentry/Platforms/Android/NativeOptions.cs
+++ b/src/Sentry/Platforms/Android/NativeOptions.cs
@@ -154,20 +154,6 @@ public partial class SentryOptions
         /// </remarks>
         public bool EnableUserInteractionTracing { get; set; } = false;
 
-        /// <summary>
-        /// Gets or sets the strategy for how Sentry Native's signal handler interacts with the CLR/Mono
-        /// signal handler.
-        /// The default value is <see cref="Android.SignalHandlerStrategy.Default"/>.
-        /// </summary>
-        /// <remarks>
-        /// .NET runtimes 10.0.0–10.0.3 (.NET SDKs 10.0.100–10.0.301) are not compatible with
-        /// <see cref="Android.SignalHandlerStrategy.ChainAtStart"/>. Using it on affected
-        /// versions throws an <see cref="InvalidOperationException"/> during initialization.
-        /// The issue was resolved in .NET runtime 10.0.4 (.NET SDK 10.0.400). See
-        /// <see href="https://github.com/dotnet/runtime/pull/123346">dotnet/runtime#123346</see>.
-        /// </remarks>
-        public SignalHandlerStrategy SignalHandlerStrategy { get; set; } = SignalHandlerStrategy.Default;
-
         // ---------- From SentryOptions.java ----------
 
         /// <summary>
@@ -280,6 +266,20 @@ public partial class SentryOptions
         public class NativeExperimentalOptions
         {
             public NativeSentryReplayOptions SessionReplay { get; set; } = new();
+
+            /// <summary>
+            /// Gets or sets the strategy for how Sentry Native's signal handler interacts with the CLR/Mono
+            /// signal handler.
+            /// The default value is <see cref="Android.SignalHandlerStrategy.Default"/>.
+            /// </summary>
+            /// <remarks>
+            /// .NET runtimes 10.0.0–10.0.3 (.NET SDKs 10.0.100–10.0.301) are not compatible with
+            /// <see cref="Android.SignalHandlerStrategy.ChainAtStart"/>. Using it on affected
+            /// versions throws an <see cref="InvalidOperationException"/> during initialization.
+            /// The issue was resolved in .NET runtime 10.0.4 (.NET SDK 10.0.400). See
+            /// <see href="https://github.com/dotnet/runtime/pull/123346">dotnet/runtime#123346</see>.
+            /// </remarks>
+            public SignalHandlerStrategy SignalHandlerStrategy { get; set; } = SignalHandlerStrategy.Default;
         }
 
         public class NativeSentryReplayOptions
--- a/src/Sentry/Platforms/Android/NativeOptions.cs
+++ b/src/Sentry/Platforms/Android/NativeOptions.cs
@@ -1,3 +1,5 @@
+using Sentry.Android;
+
 // ReSharper disable once CheckNamespace
 namespace Sentry;
 
@@ -151,6 +153,19 @@ public partial class SentryOptions
         /// See https://docs.sentry.io/platforms/android/performance/instrumentation/automatic-instrumentation/#user-interaction-instrumentation
         /// </remarks>
         public bool EnableUserInteractionTracing { get; set; } = false;
+
+        /// <summary>
+        /// Gets or sets the strategy for how Sentry Native's signal handler interacts with the CLR/Mono
+        /// signal handler.
+        /// The default value is <c>SignalHandlerStrategy.Default</c> (enabled).
+        /// </summary>
+        /// <remarks>
+        /// When set to <see cref="Android.SignalHandlerStrategy.ChainAtStart"/>, .NET runtimes
+        /// 10.0.100–10.0.301 shipped in .NET SDKs 10.0.0–10.0.3 will crash. The issue was fixed in
+        /// <see href="https://github.com/dotnet/runtime/pull/123346">dotnet/runtime#123346</see>,
+        /// which shipped as .NET runtime 10.0.400 in .NET SDK 10.0.4.
+        /// </remarks>
+        public SignalHandlerStrategy SignalHandlerStrategy { get; set; } = SignalHandlerStrategy.Default;
 
         // ---------- From SentryOptions.java ----------
 
--- a/src/Sentry/Platforms/Android/NativeOptions.cs
+++ b/src/Sentry/Platforms/Android/NativeOptions.cs
@@ -274,8 +274,8 @@ public partial class SentryOptions
             /// </summary>
             /// <remarks>
             /// .NET runtimes 10.0.0–10.0.3 (.NET SDKs 10.0.100–10.0.301) are not compatible with
-            /// <see cref="Android.SignalHandlerStrategy.ChainAtStart"/>. Using it on affected
-            /// versions throws an <see cref="InvalidOperationException"/> during initialization.
+            /// <see cref="Android.SignalHandlerStrategy.ChainAtStart"/>. On affected versions,
+            /// the SDK automatically falls back to <see cref="Android.SignalHandlerStrategy.Default"/>.
             /// The issue was resolved in .NET runtime 10.0.4 (.NET SDK 10.0.400). See
             /// <see href="https://github.com/dotnet/runtime/pull/123346">dotnet/runtime#123346</see>.
             /// </remarks>
--- a/src/Sentry/Platforms/Android/NativeOptions.cs
+++ b/src/Sentry/Platforms/Android/NativeOptions.cs
@@ -160,10 +160,11 @@ public partial class SentryOptions
         /// The default value is <c>SignalHandlerStrategy.Default</c> (enabled).
         /// </summary>
         /// <remarks>
-        /// When set to <see cref="Android.SignalHandlerStrategy.ChainAtStart"/>, .NET runtimes
-        /// 10.0.100–10.0.301 shipped in .NET SDKs 10.0.0–10.0.3 will crash. The issue was fixed in
-        /// <see href="https://github.com/dotnet/runtime/pull/123346">dotnet/runtime#123346</see>,
-        /// which shipped as .NET runtime 10.0.400 in .NET SDK 10.0.4.
+        /// .NET runtimes 10.0.0–10.0.3 (.NET SDKs 10.0.100–10.0.301) are not compatible with
+        /// <see cref="Android.SignalHandlerStrategy.ChainAtStart"/>. On affected versions, the
+        /// SDK automatically falls back to <see cref="Android.SignalHandlerStrategy.Default"/>.
+        /// The issue was resolved in .NET runtime 10.0.4 (.NET SDK 10.0.400). See
+        /// <see href="https://github.com/dotnet/runtime/pull/123346">dotnet/runtime#123346</see>.
         /// </remarks>
         public SignalHandlerStrategy SignalHandlerStrategy { get; set; } = SignalHandlerStrategy.Default;
 
--- a/src/Sentry/Platforms/Android/SentrySdk.cs
+++ b/src/Sentry/Platforms/Android/SentrySdk.cs
@@ -64,7 +64,11 @@ public static partial class SentrySdk
             o.ServerName = options.ServerName;
             o.SessionTrackingIntervalMillis = (long)options.AutoSessionTrackingInterval.TotalMilliseconds;
             o.ShutdownTimeoutMillis = (long)options.ShutdownTimeout.TotalMilliseconds;
-            o.SetNativeHandlerStrategy(JavaSdk.Android.Core.NdkHandlerStrategy.SentryHandlerStrategyDefault);
+            o.SetNativeHandlerStrategy(options.Native.SignalHandlerStrategy switch
+            {
+                SignalHandlerStrategy.ChainAtStart => NdkHandlerStrategy.SentryHandlerStrategyChainAtStart,
+                _ => NdkHandlerStrategy.SentryHandlerStrategyDefault
+            });
 
             if (options.CacheDirectoryPath is { } cacheDirectoryPath)
             {
--- a/src/Sentry/Platforms/Android/SentrySdk.cs
+++ b/src/Sentry/Platforms/Android/SentrySdk.cs
@@ -64,7 +64,7 @@ public static partial class SentrySdk
             o.ServerName = options.ServerName;
             o.SessionTrackingIntervalMillis = (long)options.AutoSessionTrackingInterval.TotalMilliseconds;
             o.ShutdownTimeoutMillis = (long)options.ShutdownTimeout.TotalMilliseconds;
-            var signalHandlerStrategy = options.Native.SignalHandlerStrategy;
+            var signalHandlerStrategy = options.Native.ExperimentalOptions.SignalHandlerStrategy;
             if (signalHandlerStrategy == SignalHandlerStrategy.ChainAtStart
                 && System.Environment.Version is { Major: 10, Minor: 0, Build: < 4 })
             {
--- a/src/Sentry/Platforms/Android/SentrySdk.cs
+++ b/src/Sentry/Platforms/Android/SentrySdk.cs
@@ -64,7 +64,17 @@ public static partial class SentrySdk
             o.ServerName = options.ServerName;
             o.SessionTrackingIntervalMillis = (long)options.AutoSessionTrackingInterval.TotalMilliseconds;
             o.ShutdownTimeoutMillis = (long)options.ShutdownTimeout.TotalMilliseconds;
-            o.SetNativeHandlerStrategy(options.Native.SignalHandlerStrategy switch
+            var signalHandlerStrategy = options.Native.SignalHandlerStrategy;
+            if (signalHandlerStrategy == SignalHandlerStrategy.ChainAtStart
+                && System.Environment.Version is { Major: 10, Minor: 0, Build: < 4 })
+            {
+                options.LogWarning(
+                    "SignalHandlerStrategy.ChainAtStart is not compatible with .NET runtime {0}. " +
+                    "Falling back to SignalHandlerStrategy.Default. Update to .NET runtime 10.0.4 or later.",
+                    System.Environment.Version);
+                signalHandlerStrategy = SignalHandlerStrategy.Default;
+            }
+            o.SetNativeHandlerStrategy(signalHandlerStrategy switch
             {
                 SignalHandlerStrategy.ChainAtStart => NdkHandlerStrategy.SentryHandlerStrategyChainAtStart,
                 _ => NdkHandlerStrategy.SentryHandlerStrategyDefault
--- a/src/Sentry/Platforms/Android/SentrySdk.cs
+++ b/src/Sentry/Platforms/Android/SentrySdk.cs
@@ -68,11 +68,10 @@ public static partial class SentrySdk
             if (signalHandlerStrategy == SignalHandlerStrategy.ChainAtStart
                 && System.Environment.Version is { Major: 10, Minor: 0, Build: < 4 })
             {
-                options.LogWarning(
-                    "SignalHandlerStrategy.ChainAtStart is not compatible with .NET runtime {0}. " +
-                    "Falling back to SignalHandlerStrategy.Default. Update to .NET runtime 10.0.4 or later.",
-                    System.Environment.Version);
-                signalHandlerStrategy = SignalHandlerStrategy.Default;
+                var msg = $"SignalHandlerStrategy.ChainAtStart is not compatible with .NET runtime " +
+                    $"{System.Environment.Version}. Update to .NET runtime 10.0.4 or later.";
+                options.LogFatal(msg);
+                throw new InvalidOperationException(msg);
             }
             o.SetNativeHandlerStrategy(signalHandlerStrategy switch
             {
--- a/src/Sentry/Platforms/Android/SentrySdk.cs
+++ b/src/Sentry/Platforms/Android/SentrySdk.cs
@@ -68,10 +68,11 @@ public static partial class SentrySdk
             if (signalHandlerStrategy == SignalHandlerStrategy.ChainAtStart
                 && System.Environment.Version is { Major: 10, Minor: 0, Build: < 4 })
             {
-                var msg = $"SignalHandlerStrategy.ChainAtStart is not compatible with .NET runtime " +
-                    $"{System.Environment.Version}. Update to .NET runtime 10.0.4 or later.";
-                options.LogFatal(msg);
-                throw new InvalidOperationException(msg);
+                options.LogWarning(
+                    "SignalHandlerStrategy.ChainAtStart is not compatible with .NET runtime {0}. " +
+                    "Falling back to SignalHandlerStrategy.Default. Update to .NET runtime 10.0.4 or later.",
+                    System.Environment.Version);
+                signalHandlerStrategy = SignalHandlerStrategy.Default;
             }
             o.SetNativeHandlerStrategy(signalHandlerStrategy switch
             {
--- a/src/Sentry/Platforms/Android/SentrySdk.cs
+++ b/src/Sentry/Platforms/Android/SentrySdk.cs
@@ -64,6 +64,7 @@ public static partial class SentrySdk
             o.ServerName = options.ServerName;
             o.SessionTrackingIntervalMillis = (long)options.AutoSessionTrackingInterval.TotalMilliseconds;
             o.ShutdownTimeoutMillis = (long)options.ShutdownTimeout.TotalMilliseconds;
+
             var signalHandlerStrategy = options.Native.ExperimentalOptions.SignalHandlerStrategy;
             if (signalHandlerStrategy == SignalHandlerStrategy.ChainAtStart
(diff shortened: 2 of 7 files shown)


jpnurmi:
@jamescrosswell This is marked as a draft because it's still waiting for dependencies, but we could already kick off the review process if you have time. :)

Do you have a preferred approach for exposing this as an opt-in API? The current proposal mirrors the sentry-native enum.

```cs
#if ANDROID
    options.Native.SignalHandlerStrategy = Sentry.Android.SignalHandlerStrategy.ChainAtStart;
#endif
```

With only two values, a simple boolean flag could also work for the time being, but I'm unsure if potential Android Tombstone support could change things in the future.

jpnurmi:
Update: I'll get back to this after #4750 to make sure this works with .NET 10

jpnurmi:
While the signal handler strategy fixes the redundant `SIGSEGV` vs. `NullReferenceException` issue, the native exception test case breaks with .NET 10 on Android X86_64. 😢 All tests pass locally in an ARM64 Android emulator.

jpnurmi:
.NET 10 on ARM64 and .NET 9.0 on both ARM64 and X86_64 behave as desired when sentry-native calls the runtime's signal handler. It returns to sentry-native which then captures the crash.

.NET 10 on X86_64, however, crashes trying to lock a destroyed mutex? No crash captured.

## .NET 9.0 (Android ARM64 & X86_64) :+1: 

```
02-09 15:38:46.029  9141  9141 I DOTNET  :   Debug: Triggering a deliberate exception because SentrySdk.CauseCrash(CrashType.Native) was called
02-09 15:38:46.029  9141  9141 D app_process64:   Debug: Triggering a deliberate exception because SentrySdk.CauseCrash(CrashType.Native) was called
02-09 15:38:46.030  9141  9141 I sentry-native: entering signal handler
02-09 15:38:46.030  9141  9141 D sentry-native: defer to runtime signal handler at start
02-09 15:38:46.031  9141  9141 F libc    : Fatal signal 4 (SIGILL), code 0 (SI_USER) in tid 9141 (egrationtestapp), pid 9141 (egrationtestapp)
02-09 15:38:46.916  9176  9176 F DEBUG   : *** *** *** *** *** *** *** *** *** *** *** *** *** *** *** ***
02-09 15:38:46.916  9176  9176 F DEBUG   : Build fingerprint: 'google/sdk_gphone64_x86_64/emu64xa:16/BE2A.250530.026.F3/13894323:userdebug/dev-keys'
02-09 15:38:46.916  9176  9176 F DEBUG   : Revision: '0'
02-09 15:38:46.916  9176  9176 F DEBUG   : ABI: 'x86_64'
02-09 15:38:46.916  9176  9176 F DEBUG   : Timestamp: 2026-02-09 15:38:46.136533669+0100
02-09 15:38:46.916  9176  9176 F DEBUG   : Process uptime: 2s
02-09 15:38:46.916  9176  9176 F DEBUG   : Cmdline: io.sentry.dotnet.maui.device.integrationtestapp
02-09 15:38:46.916  9176  9176 F DEBUG   : pid: 9141, tid: 9141, name: egrationtestapp  >>> io.sentry.dotnet.maui.device.integrationtestapp <<<
02-09 15:38:46.916  9176  9176 F DEBUG   : uid: 10235
02-09 15:38:46.916  9176  9176 F DEBUG   : signal 4 (SIGILL), code 0 (SI_USER), fault addr --------
02-09 15:38:46.916  9176  9176 F DEBUG   :     rax 0000000000000000  rbx 00006fff13d88298  rcx 000070002be8ac70  rdx 0000000000000200
02-09 15:38:46.916  9176  9176 F DEBUG   :     r8  0000000000000002  r9  00007fff348293f0  r10 0000000000000005  r11 00006ffe94ecf9d0
02-09 15:38:46.916  9176  9176 F DEBUG   :     r12 000070008c9f9fb0  r13 000070002be8ac80  r14 00006ffe8a9f35f0  r15 000070011be91d80
02-09 15:38:46.916  9176  9176 F DEBUG   :     rdi 000070011be91d80  rsi 0000000000000006
02-09 15:38:46.916  9176  9176 F DEBUG   :     rbp 00006fff13d88298  rsp 00007fff348293c8  rip 00006ffe8a9f35f0
02-09 15:38:46.916  9176  9176 F DEBUG   : 2 total frames
02-09 15:38:46.916  9176  9176 F DEBUG   : backtrace:
02-09 15:38:46.916  9176  9176 F DEBUG   :       #00 pc 00000000000005f0  /data/app/~~Z3VBUbKXonuWeLlN4W2eXQ==/io.sentry.dotnet.maui.device.integrationtestapp-gqdBX9FklKvyqOkUvbN8Zw==/lib/x86_64/libsentrysupplemental.so (crash+0) (BuildId: c2d6e5ae745143957e682d1b4fd73eac1d41a339)
02-09 15:38:46.916  9176  9176 F DEBUG   :       #01 pc 00000000000f4cc8  <anonymous:4a5c6000>
02-09 15:38:46.931  9141  9141 D sentry-native: return from runtime signal handler, we handle the signal
02-09 15:38:46.937  9141  9141 D sentry-native: captured backtrace from ucontext with 2 frames
02-09 15:38:46.937  9141  9141 D sentry-native: captured backtrace with 2 frames
02-09 15:38:46.938  9141  9141 D sentry-native: merging global scope into event
02-09 15:38:46.938  9141  9141 D sentry-native: trying to read modules from /proc/self/maps
02-09 15:38:46.975  9141  9141 D sentry-native: read 443 modules from /proc/self/maps
02-09 15:38:46.975  9141  9141 D sentry-native: sending envelope
02-09 15:38:46.976  9141  9141 I sentry-native: crash has been captured
```

## NET 10.0 (Android ARM64) :+1: 

```
02-09 16:44:27.396  8471  8471 I DOTNET  :   Debug: Triggering a deliberate exception because SentrySdk.CauseCrash(CrashType.Native) was called
02-09 16:44:27.396  8471  8471 D app_process64:   Debug: Triggering a deliberate exception because SentrySdk.CauseCrash(CrashType.Native) was called
02-09 16:44:27.398  8471  8471 I sentry-native: entering signal handler
02-09 16:44:27.398  8471  8471 D sentry-native: defer to runtime signal handler at start
02-09 16:44:27.398  8471  8471 F libc    : Fatal signal 5 (SIGTRAP), code 1 (TRAP_BRKPT), fault addr 0x79af81d5a8 in tid 8471 (egrationtestapp), pid 8471 (egrationtestapp)
02-09 16:44:27.875  8508  8508 F DEBUG   : *** *** *** *** *** *** *** *** *** *** *** *** *** *** *** ***
02-09 16:44:27.875  8508  8508 F DEBUG   : Build fingerprint: 'google/sdk_gphone64_arm64/emu64a:16/BE2A.250530.026.D1/13818094:user/release-keys'
02-09 16:44:27.875  8508  8508 F DEBUG   : Revision: '0'
02-09 16:44:27.875  8508  8508 F DEBUG   : ABI: 'arm64'
02-09 16:44:27.875  8508  8508 F DEBUG   : Timestamp: 2026-02-09 16:44:27.434533683+0100
02-09 16:44:27.875  8508  8508 F DEBUG   : Process uptime: 2s
02-09 16:44:27.875  8508  8508 F DEBUG   : Cmdline: io.sentry.dotnet.maui.device.integrationtestapp
02-09 16:44:27.875  8508  8508 F DEBUG   : pid: 8471, tid: 8471, name: egrationtestapp  >>> io.sentry.dotnet.maui.device.integrationtestapp <<<
02-09 16:44:27.875  8508  8508 F DEBUG   : uid: 10339
02-09 16:44:27.875  8508  8508 F DEBUG   : tagged_addr_ctrl: 0000000000000001 (PR_TAGGED_ADDR_ENABLE)
02-09 16:44:27.875  8508  8508 F DEBUG   : pac_enabled_keys: 000000000000000f (PR_PAC_APIAKEY, PR_PAC_APIBKEY, PR_PAC_APDAKEY, PR_PAC_APDBKEY)
02-09 16:44:27.875  8508  8508 F DEBUG   : signal 5 (SIGTRAP), code 1 (TRAP_BRKPT), fault addr 0x00000079af81d5a8
02-09 16:44:27.875  8508  8508 F DEBUG   :     x0  00000079af81d5a8  x1  0000000000000006  x2  b400007bc1c84138  x3  b400007c51c1d888
02-09 16:44:27.875  8508  8508 F DEBUG   :     x4  0000007fdebc1238  x5  b400007c51c1d8e0  x6  6f732e6c61746e65  x7  b400007b91d58798
02-09 16:44:27.875  8508  8508 F DEBUG   :     x8  0000000000000200  x9  0000000000000002  x10 0000000000000001  x11 00000000b74f8fcb
02-09 16:44:27.875  8508  8508 F DEBUG   :     x12 0000000000019002  x13 b6f2fcf39918a7a7  x14 0000000000000008  x15 b400007b020651e8
02-09 16:44:27.875  8508  8508 F DEBUG   :     x16 0000000000000001  x17 0000007d7a904fd0  x18 0000007da9ef4000  x19 0000000000000001
02-09 16:44:27.875  8508  8508 F DEBUG   :     x20 0000007fdebc1400  x21 0000007ada32c298  x22 0000007ada32c298  x23 b400007b0276a2f8
02-09 16:44:27.875  8508  8508 F DEBUG   :     x24 b400007c91c07650  x25 0000007a49a5b590  x26 b400007bc1c84120  x27 0000000000000000
02-09 16:44:27.875  8508  8508 F DEBUG   :     x28 0000007fdebc1860  x29 0000007fdebc1220
02-09 16:44:27.875  8508  8508 F DEBUG   :     lr  0000007d99c724dc  sp  0000007fdebc1220  pc  00000079af81d5a8  pst 0000000080001000
02-09 16:44:27.875  8508  8508 F DEBUG   : 2 total frames
02-09 16:44:27.875  8508  8508 F DEBUG   : backtrace:
02-09 16:44:27.875  8508  8508 F DEBUG   :       #00 pc 00000000000005a8  /data/app/~~s1hgF6ZtWJQ9vnEajeatvA==/io.sentry.dotnet.maui.device.integrationtestapp-uNphE0qc7n7bR5e8MYGTxg==/lib/arm64/libsentrysupplemental.so (crash+0) (BuildId: f20ec64b4897db5a4517ffb4f61067017ceb3b87)
02-09 16:44:27.875  8508  8508 F DEBUG   :       #01 pc 000000000000b4d8  <anonymous:7d99c67000>
02-09 16:44:27.887  8471  8471 D sentry-native: return from runtime signal handler, we handle the signal
02-09 16:44:27.898  8471  8471 D sentry-native: captured backtrace from ucontext with 2 frames
02-09 16:44:27.898  8471  8471 D sentry-native: captured backtrace with 2 frames
02-09 16:44:27.898  8471  8471 D sentry-native: merging global scope into event
02-09 16:44:27.898  8471  8471 D sentry-native: trying to read modules from /proc/self/maps
02-09 16:44:27.944  8471  8471 D sentry-native: read 443 modules from /proc/self/maps
02-09 16:44:27.945  8471  8471 D sentry-native: sending envelope
02-09 16:44:27.945  8471  8471 I sentry-native: crash has been captured
```

## NET 10.0 (Android X86_64) :-1: 

```
02-09 15:41:14.428  9380  9380 I DOTNET  :   Debug: Triggering a deliberate exception because SentrySdk.CauseCrash(CrashType.Native) was called
02-09 15:41:14.428  9380  9380 D app_process64:   Debug: Triggering a deliberate exception because SentrySdk.CauseCrash(CrashType.Native) was called
02-09 15:41:14.429  9380  9380 I sentry-native: entering signal handler
02-09 15:41:14.429  9380  9380 D sentry-native: defer to runtime signal handler at start
02-09 15:41:14.429  9380  9411 F libc    : FORTIFY: pthread_mutex_lock called on a destroyed mutex (0x7001ba602c38)
02-09 15:41:14.429  9380  9410 F libc    : FORTIFY: pthread_mutex_lock called on a destroyed mutex (0x7001ba602c38)
02-09 15:41:14.458   711  2169 I ActivityManager: Process io.sentry.dotnet.maui.device.integrationtestapp (pid 9380) has died: fg  TOP
02-09 15:41:14.459   485   485 I Zygote  : Process 9380 exited due to signal 6 (Aborted)
```

jpnurmi:
Claude confidently claims the issue was fixed by:
- dotnet/runtime#123346

The fix was backported to `release/10.0`. We can give it a try later with the next .NET 10 release.

<details><summary>Investigation: .NET 10 x86_64 native crash test failure</summary>

Traced the root cause via `strace` on the Android emulator.

### What happens

When sentry-native defers to Mono's signal handler (`CHAIN_AT_START`), Mono's `mono_handle_native_crash` walks the managed stack. On .NET 10, this walk uses the `MONO_UNWIND_SIGNAL_SAFE` flag ([added in d34ef7e](https://github.com/dotnet/runtime/commit/d34ef7e2d3f41f85d35d23ec484f7af566fd0d2f)), which causes AOT method JIT info to be loaded in async-safe mode with `ji->async = TRUE`. The stack walk callback `print_stack_frame_signal_safe` was **not updated** to check `!ji->async` before calling `mono_jit_info_get_method()`, which hits:

- Assertion at jit-info.c:918, condition `!ji->async` not met

The assertion triggers `abort()`, and since `mono_handle_native_crash` already reset `SIGABRT` to `SIG_DFL`, the process is killed immediately — before control returns to sentry-native.

### Fix

Already merged upstream:
- **dotnet/runtime#123346** ([`f49873c`](https://github.com/dotnet/runtime/commit/f49873c56a0)) — adds `!ji->async` check to `print_stack_frame_signal_safe` and `mono_find_jit_info`
- Backported to `release/10.0` as **dotnet/runtime#123792** ([`cdc557a`](https://github.com/dotnet/runtime/commit/cdc557af23d)) on Feb 6, 2026
- Related issue: https://github.com/dotnet/runtime/issues/122797

</details>

jpnurmi:
SDK 10.0.103 ships with runtime 10.0.3, which was built 12 days before the fix was merged. The fix is in the `release/10.0` branch but hasn't shipped in a servicing release yet. It will likely be in runtime 10.0.4.

- Runtime pack: Microsoft.NETCore.App.Runtime.Mono.android-x64 10.0.3
- Built from: commit c2435c3e0f46de784341ac3ed62863ce77e117b4 in the dotnet VMR
- Package dates: January 25, 2026
- Fix merged: February 6, 2026 (commit cdc557af23d)

Confirmed with strace — the crash sequence is identical to before:
```
07:17:43.198063  at <unknown> <0xffffffff>
07:17:43.198083  at Sentry.SentrySdk:NativeCrash <0x000d8>
07:17:43.198099  at Sentry.SentrySdk:CauseCrash <0x00a0a>
07:17:43.198145  * Assertion at jit-info.c:918, condition `!ji->async' not met
07:17:43.198800  FORTIFY: pthread_mutex_lock called on a destroyed mutex
```

jpnurmi:
I confirmed with a locally built `libmonosgen-2.0.so` that dotnet/runtime#123346 fixes the issue.

I used the 10.0.3 baseline of `dotnet/runtime`, cherry-picked the fix, built (`./build.sh mono -os android -arch x64 -c Release`), and copied over `/usr/share/dotnet/packs/Microsoft.NETCore.App.Runtime.Mono.android-x64/10.0.3/runtimes/android-x64/native/libmonosgen-2.0.so`.

```
02-11 07:58:07.845  5437  5437 I DOTNET  :   Debug: Triggering a deliberate exception because SentrySdk.CauseCrash(CrashType.Native) was called
02-11 07:58:07.845  5437  5437 D app_process64:   Debug: Triggering a deliberate exception because SentrySdk.CauseCrash(CrashType.Native) was called
02-11 07:58:07.846  5437  5437 I sentry-native: entering signal handler
02-11 07:58:07.846  5437  5437 D sentry-native: defer to runtime signal handler at start
02-11 07:58:07.846  5437  5437 F libc    : Fatal signal 11 (SIGSEGV), code 1 (SEGV_MAPERR), fault addr 0x0 in tid 5437 (egrationtestapp), pid 5437 (egrationtestapp)
02-11 07:58:08.031  5475  5475 F DEBUG   : *** *** *** *** *** *** *** *** *** *** *** *** *** *** *** ***
02-11 07:58:08.031  5475  5475 F DEBUG   : Build fingerprint: 'google/sdk_gphone64_x86_64/emu64xa:16/BE2A.250530.026.F3/13894323:userdebug/dev-keys'
02-11 07:58:08.031  5475  5475 F DEBUG   : Revision: '0'
02-11 07:58:08.031  5475  5475 F DEBUG   : ABI: 'x86_64'
02-11 07:58:08.031  5475  5475 F DEBUG   : Timestamp: 2026-02-11 07:58:07.902802417+0100
02-11 07:58:08.031  5475  5475 F DEBUG   : Process uptime: 2s
02-11 07:58:08.031  5475  5475 F DEBUG   : Cmdline: io.sentry.dotnet.maui.device.integrationtestapp
02-11 07:58:08.031  5475  5475 F DEBUG   : pid: 5437, tid: 5437, name: egrationtestapp  >>> io.sentry.dotnet.maui.device.integrationtestapp <<<
02-11 07:58:08.031  5475  5475 F DEBUG   : uid: 10246
02-11 07:58:08.031  5475  5475 F DEBUG   : signal 11 (SIGSEGV), code 1 (SEGV_MAPERR), fault addr 0x0000000000000000
02-11 07:58:08.031  5475  5475 F DEBUG   : Cause: null pointer dereference
02-11 07:58:08.031  5475  5475 F DEBUG   :     rax 0000000000000000  rbx 000079c99f938298  rcx 000079ca67c9d010  rdx 0000000000000200
02-11 07:58:08.031  5475  5475 F DEBUG   :     r8  0000000000000002  r9  00007ffdbd520910  r10 0000000000000005  r11 000079c918eb6eb0
02-11 07:58:08.031  5475  5475 F DEBUG   :     r12 000079c9f8856a88  r13 000079ca67c9d020  r14 000079c909c8a5b0  r15 000079cba7cb20d0
02-11 07:58:08.031  5475  5475 F DEBUG   :     rdi 000079cba7cb20d0  rsi 0000000000000006
02-11 07:58:08.031  5475  5475 F DEBUG   :     rbp 000079c99f938298  rsp 00007ffdbd5208d8  rip 000079c909c8a5be
02-11 07:58:08.031  5475  5475 F DEBUG   : 2 total frames
02-11 07:58:08.031  5475  5475 F DEBUG   : backtrace:
02-11 07:58:08.031  5475  5475 F DEBUG   :       #00 pc 00000000000005be  /data/app/~~RnBcgBqVyUx2MEIc7xRDiA==/io.sentry.dotnet.maui.device.integrationtestapp-hriVX4xDjbd5AhYJnurtoA==/lib/x86_64/libsentrysupplemental.so (crash+14) (BuildId: 852925ef6da49aeb59a096644bd533ff2fce55dc)
02-11 07:58:08.031  5475  5475 F DEBUG   :       #01 pc 00000000000e47a8  <anonymous:405af000>
02-11 07:58:08.044  5437  5437 D sentry-native: return from runtime signal handler, we handle the signal
02-11 07:58:08.047   716   740 I ActivityManager: Showing crash dialog for package io.sentry.dotnet.maui.device.integrationtestapp u0
02-11 07:58:08.053  5437  5437 D sentry-native: captured backtrace from ucontext with 2 frames
02-11 07:58:08.053  5437  5437 D sentry-native: captured backtrace with 2 frames
02-11 07:58:08.053  5437  5437 D sentry-native: merging global scope into event
02-11 07:58:08.053  5437  5437 D sentry-native: trying to read modules from /proc/self/maps
02-11 07:58:08.077  5437  5437 D sentry-native: read 445 modules from /proc/self/maps
02-11 07:58:08.077  5437  5437 D sentry-native: sending envelope
02-11 07:58:08.078  5437  5437 I sentry-native: crash has been captured
```

jamescrosswell:
> I used the 10.0.3 baseline of `dotnet/runtime`, cherry-picked the fix, built (`./build.sh mono -os android -arch x64 -c Release`), and copied over `/usr/share/dotnet/packs/Microsoft.NETCore.App.Runtime.Mono.android-x64/10.0.3/runtimes/android-x64/native/libmonosgen-2.0.so`.

Nice! So once 10.0.30 is released we could put together a fix. Would our SDK users also need to be building with 10.0.3 or is it sufficient for us to be building the Sentry SDK for .NET with it?

jpnurmi:
Applying the fix when building Sentry alone is not sufficient. The fix must be present in the .NET runtime / Mono DLL that is packaged inside the application APK.

Unfortunately, the fix was not included in version 10.0.3, which was released a few days ago. We'll have to wait for 10.0.4, the next monthly servicing update.

jamescrosswell:
Looks like [10.0.4](https://github.com/dotnet/sdk/releases/tag/v10.0.200) is finally out and even 10.0.5, which we have a PR to bump to here:
- https://github.com/getsentry/sentry-dotnet/pull/5029

That PR needs some debugging - @jpnurmi would you have the bandwidth to look at it or should @Flash0ver or I pick it up?

jpnurmi:
Another issue has come up. 😓 While testing against 10.0.4, I discovered that the refactored inproc backend introduced in sentry-native [0.13](https://github.com/getsentry/sentry-dotnet/pull/4941) broke the `CHAIN_AT_START` signal chaining strategy. A fix is in progress:
- https://github.com/getsentry/sentry-native/pull/1572

To prevent similar regressions going forward, we're also adding a .NET Android integration test to the Native SDK:
- https://github.com/getsentry/sentry-native/pull/1574

jamescrosswell:
> Another issue has come up. 😓 While testing against 10.0.4, I discovered that the refactored inproc backend introduced in sentry-native [0.13](https://github.com/getsentry/sentry-dotnet/pull/4941) broke the `CHAIN_AT_START` signal chaining strategy.

Hm... what's your gut tell you about the robustness of the solution we're putting in place here? Is it something that's likely to work for long periods without us needing to pay attention to it or is it possibly fragile and going to be a time sink?

I mean, pretty much everything about MAUI seems fragile to some extent... is this going to be worse though?

jpnurmi:
As AI would put it, "You're absolutely right to question that!"

Chaining signal handlers as we do with the "chain-at-start" strategy is not an officially supported use case and could potentially break in any future version of .NET or Android.

Furthermore, there's an endless combination of versions and targets/devices/emulators/architectures to cover, and we have quite big gaps in testing as we currently can't even test on ARM64 (#4660). If we still want to release this, it's best to start with an opt-in to gather feedback.

jamescrosswell:
> As AI would put it, "You're absolutely right to question that!"

🤣 

> If we still want to release this, it's best to start with an opt-in to gather feedback.

That sounds like a good approach... maybe we leave it opt in for a decent amount of time to see if it can survive major version bumps in the various different moving parts.

jpnurmi:
~~Temporarily rebased on top of #5069.~~ Android jobs are green. Finally! 🎉

#5069 needs to be merged, and then this will be ready to go. Do the changes look otherwise good, btw?

src/Sentry/Platforms/Android/NativeOptions.cs
```diff
@@ -152,6 +154,19 @@ internal NativeOptions(SentryOptions options)
         /// </remarks>
         public bool EnableUserInteractionTracing { get; set; } = false;
 
+        /// <summary>
+        /// Gets or sets the strategy for how Sentry Native's signal handler interacts with the CLR/Mono
+        /// signal handler.
+        /// The default value is <c>SignalHandlerStrategy.Default</c> (enabled).
```
jamescrosswell: We could give a description here of what the two different strategies are/do and why someone might set this to something other than default.

src/Sentry/Platforms/Android/NativeOptions.cs
```diff
@@ -152,6 +154,19 @@ internal NativeOptions(SentryOptions options)
...
+        /// </summary>
+        /// <remarks>
+        /// When set to <see cref="Android.SignalHandlerStrategy.ChainAtStart"/>, .NET runtimes
+        /// 10.0.100–10.0.301 shipped in .NET SDKs 10.0.0–10.0.3 will crash. The issue was fixed in
+        /// <see href="https://github.com/dotnet/runtime/pull/123346">dotnet/runtime#123346</see>,
+        /// which shipped as .NET runtime 10.0.400 in .NET SDK 10.0.4.
+        /// </remarks>
+        public SignalHandlerStrategy SignalHandlerStrategy { get; set; } = SignalHandlerStrategy.Default;
```
jamescrosswell: Can we maybe provide a runtime check here on the setter here?

If an attempt is made to set this to `ChainAtStart` on an unsupported runtime version, we could fail early with a descriptive error message (rather than waiting for something cryptic and hard to debug to occur)? 

Or at the very least, a warning level debug so that when people get in touch they/we hopefully see that in the logs...

Also, not sure if we want that logic here or in `src/Sentry/Platforms/Android/SentrySdk.cs`... probably doesn't matter too much (unless @Flash0ver you have some preference).

jamescrosswell (commented):
Thanks @jpnurmi - the other dependencies are now merged so I think this is almost ready.

The changes themselves look fine functionally (pretty trivial - just wrapping an android SDK option). I made a few suggestions that might hopefully head off confusion for our SDK users though.

src/Sentry/Platforms/Android/NativeOptions.cs
jpnurmi: How do you like the idea of exposing the NDK's handler strategy as-is? I thought it would be the most straightforward, but we could just as well come up with some easier-to-understand boolean option that fits well alongside the existing `SuppressXxx` options... 🤔

src/Sentry/Platforms/Android/NativeOptions.cs
jpnurmi: Do you prefer to fail hard or log a warning and fall back to the default?

jpnurmi: I guess hard fail makes more sense because the app developer must have explicitly opted in, yet trying to build/run with an incompatible SDK/runtime.

src/Sentry/Platforms/Android/NativeOptions.cs
jpnurmi: Sorry for derailing a bit, but dumping this into `options.Native.ExperimentalOptions` could also be fine given that this is an imperfect (hopefully) short-term solution.

Edit: proposal pushed

src/Sentry/Platforms/Android/NativeOptions.cs
jamescrosswell: Yeah I think a hard fail is less likely to go unnoticed (and have us trying to debug domino issues further down the line).

src/Sentry/Platforms/Android/SentrySdk.cs
```diff
@@ -68,11 +68,10 @@ private static void InitSentryAndroidSdk(SentryOptions options)
...
             {
-                options.LogWarning(
-                    "SignalHandlerStrategy.ChainAtStart is not compatible with .NET runtime {0}. " +
-                    "Falling back to SignalHandlerStrategy.Default. Update to .NET runtime 10.0.4 or later.",
-                    System.Environment.Version);
-                signalHandlerStrategy = SignalHandlerStrategy.Default;
+                var msg = $"SignalHandlerStrategy.ChainAtStart is not compatible with .NET runtime " +
+                    $"{System.Environment.Version}. Update to .NET runtime 10.0.4 or later.";
+                options.LogFatal(msg);
+                throw new InvalidOperationException(msg);
```
jamescrosswell: Hm... now I see it I wonder if your old code isn't better (fall back to Default on the wrong runtime). 

In practice, this means the user's app would work on all runtimes but possibly they get duplicate exceptions reported to Sentry on runtimes that don't includ the fix. 

Whereas with the new code (that I suggested) the app would fail at startup on incompatible runtimes... which seems a bit severe, just to prevent duplicate exception reporting. 

Apologies @jpnurmi - could we revert to your original solution? My bad as I didn't actually see/notice the fallback logic on my first review.

src/Sentry/Platforms/Android/NativeOptions.cs
jamescrosswell: Sorry for flip flopping on you @jpnurmi - I just saw (finally) the fallback logic that you had in place, which I think is a better solution to the one I proposed.

src/Sentry/Platforms/Android/SentrySdk.cs
jpnurmi: No problem at all! Either way makes sense. The downside of the fallback is that it gets easily buried in the debug output.

jamescrosswell (approved):
LGTM assuming the iOS failure is just flaky tests.

Massive thank you for this @jpnurmi ❤️

src/Sentry/Platforms/Android/NativeOptions.cs
```diff
@@ -264,6 +266,20 @@ public void AddInAppInclude(string prefix)
...
         {
             public NativeSentryReplayOptions SessionReplay { get; set; } = new();
+
+            /// <summary>
+            /// Gets or sets the strategy for how Sentry Native's signal handler interacts with the CLR/Mono
+            /// signal handler.
+            /// The default value is <see cref="Android.SignalHandlerStrategy.Default"/>.
+            /// </summary>
+            /// <remarks>
+            /// .NET runtimes 10.0.0–10.0.3 (.NET SDKs 10.0.100–10.0.103) are not compatible with
```
Flash0ver, line 276: question: .NET 9.0

Should we mention net9.0 as well, which we still support for Android?

But we need to keep in mind the remove it for the next major, when we will drop support for `net9.0-android` (and `net9.0` in general).

Would also apply to the doc-comment in `src/Sentry/Platforms/Android/SignalHandlerStrategy.cs`.

src/Sentry/Platforms/Android/SentrySdk.cs
```diff
@@ -64,7 +64,21 @@ private static void InitSentryAndroidSdk(SentryOptions options)
             o.ServerName = options.ServerName;
             o.SessionTrackingIntervalMillis = (long)options.AutoSessionTrackingInterval.TotalMilliseconds;
             o.ShutdownTimeoutMillis = (long)options.ShutdownTimeout.TotalMilliseconds;
-            o.SetNativeHandlerStrategy(JavaSdk.Android.Core.NdkHandlerStrategy.SentryHandlerStrategyDefault);
+            var signalHandlerStrategy = options.Native.ExperimentalOptions.SignalHandlerStrategy;
```
Flash0ver, line 68: suggestion: new line

IMO, leading with a new break here would help to read this as a "block" configuring one Option.

```suggestion

            var signalHandlerStrategy = options.Native.ExperimentalOptions.SignalHandlerStrategy;
```

src/Sentry/Platforms/Android/SentrySdk.cs
```diff
@@ -64,7 +64,21 @@ private static void InitSentryAndroidSdk(SentryOptions options)
...
+            if (signalHandlerStrategy == SignalHandlerStrategy.ChainAtStart
+                && System.Environment.Version is { Major: 10, Minor: 0, Build: < 4 })
```
Flash0ver, line 70: potential issue: .NET 9?

We are not handling .NET 9 here,
so on any .NET 9 version, we would NOT enter this block.

Is this intended?

As a follow-up thought,
when upgrading to .NET 11, this block would also NOT be entered.

src/Sentry/Platforms/Android/SignalHandlerStrategy.cs
```diff
@@ -0,0 +1,31 @@
...
+    /// <see cref="Default"/>.
+    /// </summary>
+    /// <remarks>
+    /// .NET runtimes 10.0.0–10.0.3 (.NET SDKs 10.0.100–10.0.103) are not compatible with
+    /// this strategy. On affected versions, the SDK automatically falls back to
+    /// <see cref="Default"/>. The issue was resolved in .NET runtime 10.0.4
+    /// (.NET SDK 10.0.200). See
+    /// <see href="https://github.com/dotnet/runtime/pull/123346">dotnet/runtime#123346</see>.
+    /// </remarks>
+    ChainAtStart
```
Flash0ver, line 30: question: only two?

Are there only 2 signal handler strategies in general,
or are only 2 handler strategies currently exposed which might change in the future.

I was wondering if this could be a Boolean.

But, on second thought, considering that the underlying symbol we are using is a type (`NdkHandlerStrategy`), `enum` does sound like the idiomatic way for us to expose this.

Flash0ver (changes requested):
one blocking comment: are we not handling .NET 9 (and future .NET 11)? ... or is this intentional (I might have missed something)?

src/Sentry/Platforms/Android/SignalHandlerStrategy.cs
jpnurmi: The whole "chain-at-start" concept is somewhat problematic, but we're kinda stuck with this for now because Mono doesn't allow us to reorder the signal handlers until https://github.com/dotnet/runtime/pull/125835 lands. I doubt there will be more strategies in the future. It's more likely that the whole signal handler strategy gets deprecated as soon as we get a better alternative (the Mono PR).

I've also been flirting with the idea of making it a boolean. The enum was directly copied from the native SDK as an easy starting point. Naming is the hard part, as always. :D Maybe just `ChainSignalHandlerAtStart`?

src/Sentry/Platforms/Android/SentrySdk.cs
jpnurmi: Yes, .NET 9 works, so this is as intended. Only these specific early versions of .NET 10.0 are broken. Thanks for the sanity check! :)

src/Sentry/Platforms/Android/SignalHandlerStrategy.cs
Flash0ver: I see ... I like to idea of mimicking the Native SDK ... so that we don't swap "concepts".
Also - this is an _experimental_ option ... so we can review the shape when we make it stable ... if we ever make it stable.

src/Sentry/Platforms/Android/NativeOptions.cs
Flash0ver: see https://github.com/getsentry/sentry-dotnet/pull/4676#discussion_r3065000388

Flash0ver (approved):
✅ ... all my concerns were evaporated by a newline 😉

PR #4681, opened 2025-10-30
fix: Don't set current thread when one is already there

bitsandfoxes:
## The Problem

The Sentry SDK for Unity receives stringified stack traces directly from Unity's logging system for error events (see https://github.com/getsentry/sentry-unity/pull/2377). After parsing these stack traces, the Unity SDK sets them on the event as the current thread before passing the event to the .NET SDK for processing.

However, `MainSentryEventProcessor` is unconditionally adding a new "current thread" (with `Current = true`) whenever `AttachStacktrace` was enabled and no exception stack trace was present.

https://github.com/getsentry/sentry-dotnet/blob/396a7ec6dca321f7de99fa3d27833f0db0500057/src/Sentry/Internal/MainSentryEventProcessor.cs#L94-L101

This resulted in events having **two threads marked as current** - the one from Unity with the correct stack trace, and a duplicate one captured by the .NET SDK.

## Proposal

Just skip automatic current thread creation if the event already contains a thread with `Current = true`. This is not even related to the `AttachStackTrace` option as there only ever is **one** current thread anyway.

#skip-changelog

### DIFF
--- a/src/Sentry/Internal/MainSentryEventProcessor.cs
+++ b/src/Sentry/Internal/MainSentryEventProcessor.cs
@@ -84,8 +84,7 @@ internal class MainSentryEventProcessor : ISentryEventProcessor
         @event.Distribution ??= Distribution;
 
         // If there's no current thread on the event and there is no exception with a stack trace, get the current stack trace
-        var hasCurrentThread = @event.SentryThreads?.Any(t => t.Current == true) == true;
-        if (@event.Exception?.StackTrace is null && !hasCurrentThread)
+        if (@event.Exception?.StackTrace is null && @event.SentryThreads?.Any(t => t.Current == true) != true)
         {
             var stackTrace = @event.SentryExceptions?.FirstOrDefault()?.Stacktrace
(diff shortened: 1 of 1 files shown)


src/Sentry/Internal/MainSentryEventProcessor.cs
```diff
@@ -83,8 +83,9 @@ public SentryEvent Process(SentryEvent @event)
         @event.Release ??= Release;
         @event.Distribution ??= Distribution;
 
-        // if there's no exception with a stack trace, then get the current stack trace
-        if (@event.Exception?.StackTrace is null)
+        // If there's no current thread on the event and there is no exception with a stack trace, get the current stack trace
+        var hasCurrentThread = @event.SentryThreads?.Any(t => t.Current == true) == true;
+        if (@event.Exception?.StackTrace is null && !hasCurrentThread)
```
jamescrosswell:
```suggestion
        if (@event.Exception?.StackTrace is null && @event.SentryThreads?.Any(t => t.Current == true) != true)
```

Avoids a comparison if the stack trace is null.

Issue #4682, opened 2025-10-31
CI: iOS integration tests missing timeouts


jpnurmi:
### Description

The iOS integration tests should have reasonable step-level timeouts and retry instead of hanging until they exceed the maximum execution time of 6 hours:

> **ios-tests**
> The job has exceeded the maximum execution time of 6h0m0s

> **ios-tests**
> The operation was canceled.

https://github.com/getsentry/sentry-dotnet/actions/runs/18957342849/attempts/1

PR #4684, opened 2025-10-31
ci: add timeouts for iOS device tests

jpnurmi:
Simulators are complex - occasional hangs are bound to happen. These steps typically take somewhere around 15-20 minutes. The same 40-minute timeout as used in [`device-tests-android.yml`](https://github.com/getsentry/sentry-dotnet/blob/main/.github/workflows/device-tests-android.yml) seems generous enough per step.

Close: #4682


jamescrosswell (approved):
LGTM

PR #4691, opened 2025-11-04
fix: `Serilog` integration captures structured log when event with exception has been captured

### LINKED ISSUE #4654: Structured logs not being captured when exception attached.
https://github.com/getsentry/sentry-dotnet/blob/2f46bf5e6228cb17d055b9b31101c1ba42fc0128/src/Sentry.Serilog/SentrySink.cs#L140

The early return prevents the structured log from being sent for that Serilog event when an exception is present and the level is high enough to produce a Sentry Event.

Flash0ver:
Fixes #4654
- #4654

---

When the `Serilog` integration has captured an _Event_ with an _Exception_, then
- neither an additional _Breadcrumb_ has been added (expected)
- nor a _Structured Log_ has been captured (unexpected)

This fix keeps the prevention of adding an additional _Breadcrumb_ when an _Event_ with an _Exception_ has implicitly already added a _Breadcrumb_, but will still produce a _Structured Log_ message (if enabled).
When an _Event_ without an _Exception_ does not implicitly add a _Breadcrumb_ we still continue to explicitly add a _Breadcrumb_, while continuing to capture a _Structured Log_.

### DIFF
--- a/src/Sentry.Serilog/SentrySink.cs
+++ b/src/Sentry.Serilog/SentrySink.cs
@@ -105,7 +105,7 @@ internal sealed partial class SentrySink : ILogEventSink, IDisposable
         var exception = logEvent.Exception;
         var template = logEvent.MessageTemplate.Text;
         var formatted = FormatLogEvent(logEvent);
-        var addedBreadcrumb = false;
+        var addedBreadcrumbForException = false;
 
         if (logEvent.Level >= _options.MinimumEventLevel)
         {
@@ -138,11 +138,11 @@ internal sealed partial class SentrySink : ILogEventSink, IDisposable
             // Capturing exception events adds a breadcrumb automatically... we don't want to add another one
             if (exception != null)
             {
-                addedBreadcrumb = true;
+                addedBreadcrumbForException = true;
             }
         }
 
-        if (!addedBreadcrumb && logEvent.Level >= _options.MinimumBreadcrumbLevel)
+        if (!addedBreadcrumbForException && logEvent.Level >= _options.MinimumBreadcrumbLevel)
         {
             Dictionary<string, string>? data = null;
             if (exception != null && !string.IsNullOrWhiteSpace(formatted))


Flash0ver:
@sentry review

src/Sentry.Serilog/SentrySink.cs
```diff
@@ -105,6 +105,7 @@ private void InnerEmit(LogEvent logEvent)
         var exception = logEvent.Exception;
         var template = logEvent.MessageTemplate.Text;
         var formatted = FormatLogEvent(logEvent);
```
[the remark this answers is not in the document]
Flash0ver: renamed
```diff
-addedBreadcrumb
+addedBreadcrumbForException
```

to go in line with the `Hub.AddBreadcrumbForException` method, which is implicitly adding the _Breadcrumb_ when an _Exception_ is present.

CHANGELOG.md
```diff
@@ -14,6 +14,10 @@
 - The SDK now provides a `IsSessionActive` to allow checking the session state ([#4662](https://github.com/getsentry/sentry-dotnet/pull/4662))
 - The SDK now makes use of the new SessionEndStatus `Unhandled` when capturing an unhandled but non-terminal exception, i.e. through the UnobservedTaskExceptionIntegration ([#4633](https://github.com/getsentry/sentry-dotnet/pull/4633))
 
+### Fixes
+
+- The `Serilog` integration captures _Structured Logs_ (when enabled) independently of captured Events and added Breadcrumbs ([#4691](https://github.com/getsentry/sentry-dotnet/pull/4691))
```
Flash0ver, line 19: question: `6.0.0` vs `5.16.3`

Do we want to publish this fix in a `5.16.3` patch release,
or is it fine to include in `6.0.0` as it's just a week away.

jamescrosswell: I don't think we plan to GA structured logging in version 5 of the SDK right? So doesn't make sense to backport the fix for this.

PR #4694, opened 2025-11-05
feat!: pass SentryHint to IScopeObserver

jpnurmi:
> [!NOTE]
> BREAKING CHANGE

This would allow us to associate hints with breadcrumbs originating from the native layer so that we could sync certain native breadcrumbs to .NET without pushing them back.

For example, the following proposal currently disables the Android activity lifecycle breadcrumbs from the Java SDK. With the hints available, it could selectively sync Android `ui.lifecycle` breadcrumbs to .NET instead of reimplementing the logic with MAUI lifecycle callbacks:

- https://github.com/getsentry/sentry-dotnet/pull/4688

### DIFF
--- a/test/Sentry.Tests/ApiApprovalTests.Run.DotNet10_0.verified.txt
+++ b/test/Sentry.Tests/ApiApprovalTests.Run.DotNet10_0.verified.txt
@@ -215,7 +215,7 @@ namespace Sentry
     }
     public interface IScopeObserver
     {
-        void AddBreadcrumb(Sentry.Breadcrumb breadcrumb);
+        void AddBreadcrumb(Sentry.Breadcrumb breadcrumb, Sentry.SentryHint hint);
         void SetExtra(string key, object? value);
         void SetTag(string key, string value);
         void SetTrace(Sentry.SentryId traceId, Sentry.SpanId parentSpanId);
--- a/test/Sentry.Tests/ApiApprovalTests.Run.DotNet8_0.verified.txt
+++ b/test/Sentry.Tests/ApiApprovalTests.Run.DotNet8_0.verified.txt
@@ -215,7 +215,7 @@ namespace Sentry
     }
     public interface IScopeObserver
     {
-        void AddBreadcrumb(Sentry.Breadcrumb breadcrumb);
+        void AddBreadcrumb(Sentry.Breadcrumb breadcrumb, Sentry.SentryHint hint);
         void SetExtra(string key, object? value);
         void SetTag(string key, string value);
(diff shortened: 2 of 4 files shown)


src/Sentry/Internal/ScopeObserver.cs
```diff
@@ -17,15 +17,15 @@ public ScopeObserver(
         _options = options;
     }
 
-    public void AddBreadcrumb(Breadcrumb breadcrumb)
+    public void AddBreadcrumb(Breadcrumb breadcrumb, SentryHint hint)
```
Flash0ver, line 20: question: nullability

Do we always have a Hint ... or are Hint optional?
I'm wondering whether this is OK to be non-nullable ... or it could/should be nullable.

jpnurmi: We always have a hint at the [call site](https://github.com/getsentry/sentry-dotnet/blob/18ca31baeb018dd9635a875a7b8cbdab95c691b4/src/Sentry/Scope.cs#L351) because the other overload creates an empty hint:

https://github.com/getsentry/sentry-dotnet/blob/18ca31baeb018dd9635a875a7b8cbdab95c691b4/src/Sentry/Scope.cs#L313

jpnurmi:
I'm having second thoughts about this.

This was inspired by the [use of hints to keep track of hybrid SDKs](https://github.com/getsentry/sentry-java/blob/59fa06bfa4b4333a0690dcc71332abf7a593f60f/sentry/src/main/java/io/sentry/util/HintUtils.java#L34-L36) in the Java SDK, but I realized they are not passing hints to [IScopeObserver.addBreadcrumb](https://github.com/getsentry/sentry-java/blob/59fa06bfa4b4333a0690dcc71332abf7a593f60f/sentry/src/main/java/io/sentry/IScopeObserver.java#L19) either.

I had a prototype that associated Android Activity lifecycle breadcrumbs with a "native" hint in [BeforeBreadcrumbCallback](https://github.com/getsentry/sentry-dotnet/blob/18ca31baeb018dd9635a875a7b8cbdab95c691b4/src/Sentry/Platforms/Android/Callbacks/BeforeBreadcrumbCallback.cs#L14), synced them to .NET, and then checked the hint in [AndroidScopeObserver](https://github.com/getsentry/sentry-dotnet/blob/18ca31baeb018dd9635a875a7b8cbdab95c691b4/src/Sentry/Platforms/Android/AndroidScopeObserver.cs#L26) to avoid syncing duplicates back to Java. However, the same can be achieved one level up in the scope, and instead of passing hints to the observer, one could simply not call the observer for breadcrumbs with such a "native" hint.

PR #4697, opened 2025-11-06
perf(logs): avoid string allocation when no parameters are passed

### LINKED ISSUE #5118: Avoid string allocation when no parameters are passed to methods of Structured Logger
### Summary

Avoid a string allocation when there are no parameters that the template string should be formatted with.

### Remarks

Noticed during [UNITY-180](https://linear.app/getsentry/issue/UNITY-180/structured-logging-improvements).

Flash0ver:
Closes #5118.

Avoid a string allocation when there are no parameters that the template string should be formatted with.
Also, ensure that the `Parameters` of type ``ImmutableArray`1`` are never the `default`-struct, so that we can avoid `ImmutableArray.IsDefault` checks afterwards.

@bitsandfoxes, we talked about this optimization quite a while back. It's mainly for this code path:
- https://github.com/getsentry/sentry-unity/blob/e2f2db400d776b26108759688003b8a120094538/src/Sentry.Unity/Integrations/UnityApplicationLoggingIntegration.cs#L158


benchmarks/Sentry.Benchmarks/BenchmarkDotNet.Artifacts/results/Sentry.Benchmarks.SentryStructuredLoggerBenchmarks-report-github.md
```diff
@@ -0,0 +1,14 @@
...
+.NET SDK 10.0.201
+  [Host]     : .NET 9.0.8 (9.0.825.36511), Arm64 RyuJIT AdvSIMD
+  DefaultJob : .NET 9.0.8 (9.0.825.36511), Arm64 RyuJIT AdvSIMD
+
+
+```
+| Method               | Mean     | Error   | StdDev  | Gen0   | Gen1   | Allocated |
+|--------------------- |---------:|--------:|--------:|-------:|-------:|----------:|
+| LogWithoutParameters | 102.3 ns | 1.28 ns | 1.19 ns | 0.0640 | 0.0001 |     536 B |
+| LogWithParameters    | 248.5 ns | 4.86 ns | 5.40 ns | 0.1087 |      - |     912 B |
```
Flash0ver, line 14: note: perf comparison

Before this change:

| Method               | Mean     | Error   | StdDev  | Median   | Gen0   | Allocated |
|--------------------- |---------:|--------:|--------:|---------:|-------:|----------:|
| LogWithoutParameters | 122.2 ns | 2.31 ns | 5.87 ns | 120.5 ns | 0.0696 |     584 B |
| LogWithParameters    | 242.6 ns | 0.79 ns | 0.74 ns | 242.5 ns | 0.1087 |     912 B |

After this change:

| Method               | Mean     | Error   | StdDev  | Gen0   | Gen1   | Allocated |
|--------------------- |---------:|--------:|--------:|-------:|-------:|----------:|
| LogWithoutParameters | 102.3 ns | 1.28 ns | 1.19 ns | 0.0640 | 0.0001 |     536 B |
| LogWithParameters    | 248.5 ns | 4.86 ns | 5.40 ns | 0.1087 |      - |     912 B |

Result:
- A tiny pessimization in time (IMO neglectable) with params/args
  - one more jump and assignment
- A notable optimization (a bit of time and 48 bytes in space) without params/args
  - no invocation of `System.String.Format`

src/Sentry/Internal/DefaultSentryStructuredLogger.cs
```diff
@@ -30,26 +30,33 @@ private protected override void CaptureLog(SentryLogLevel level, string template
...
             {
                 builder.Add(new KeyValuePair<string, object>(index.ToString(), parameters[index]));
             }
             @params = builder.DrainToImmutable();
         }
+        else
+        {
+            message = template;
+            template = null!;
+            @params = ImmutableArray<KeyValuePair<string, object>>.Empty;
```
Flash0ver, line 58: question: Breaking Change?

This improvement is - technically speaking - also a behavioral (breaking?) change, when Logging a plain message without positional arguments, observable by user code through the `SetBeforeSendLog` callback:

```CSharp
options.SetBeforeSendLog(static (SentryLog log) =>
{
    _ = (log.Message, log.Template, log.Parameters);
    return log;
});
```

No changes to the resulting serialized payload, though - we have a check in place that does not serialize a `sentry.message.template` when there is no `sentry.message.parameter` ... but there are user-code observable changes to the `SentryLog.Template` and `SentryLog.Parameters` properties.

---

when Logging **without** positional Parameters: e.g. `SentrySdk.Logger.LogError("Without parameters");`
- Before
  - `log.Template == log.Message`
  - `log.Parameters.IsDefault == true`
- After
  - `log.Template == null`
  - `log.Parameters.IsDefault == false`
- Justification
  - `SentryLog.Template` was already declared nullable (`string?`)
    - no contractual change
    - and this now also mirrors what is actually serialized for the payload
  - the ``ImmutableArray`1`` of `Parameters` may no longer be `default`
    - when an ``ImmutableArray`1`` is `default`, invoking any instance member other than `IsDefault` or `IsDefaultOrEmpty` will throw a `NullReferenceException`, because the underlying _Array_ is `null`
    - this makes user-callbacks "safer"
    - but also the down-stream `sentry-unity` SDK

Another benefit of `ImmutableArray<KeyValuePair<string, object>> Parameters` no longer being `default` when logging **without** positional arguments is that we can omit `IsDefault` and `IsDefaultOrEmpty` checks in the `ISentryJsonSerializable.WriteTo` serialization logic, which makes it slightly faster, too.

---

when Logging **with** positional Parameters: e.g. `SentrySdk.Logger.LogError("With parameters {0} {1} {2}", 1, 2, 3);`
- no observable changes 

---

See also https://develop.sentry.dev/sdk/telemetry/logs/#default-attributes.

src/Sentry/SentryLog.cs
```diff
@@ -25,6 +25,8 @@ internal SentryLog(DateTimeOffset timestamp, SentryId traceId, SentryLogLevel le
         Message = message;
         // 7 is the number of built-in attributes, so we start with that.
         _attributes = new Dictionary<string, SentryAttribute>(7);
+        // ensure the ImmutableArray`1 is not default, so we can omit IsDefault checks before accessing other members
+        Parameters = ImmutableArray<KeyValuePair<string, object>>.Empty;
```
Flash0ver, line 29: note: this ensures that the ``ImmutableArray`1`` of `Parameters` is never `default`

In `sentry-dotnet`, there is currently no code path that would require this, as we always set the instance to at least "Empty".

But in the down-stream `sentry-unity`, there actually is a code path where it may be `default`.
- see https://github.com/getsentry/sentry-unity/blob/e2f2db400d776b26108759688003b8a120094538/src/Sentry.Unity/Integrations/UnityApplicationLoggingIntegration.cs#L158-L159
- this is also the main scenario this optimization is for
- as in `sentry-dotnet`, we assume most usages via `ILogger` and other Logging-Integrations

With this assignment, we ensure that it's never `default`, and both us and users can omit `IsDefault`/`IsDefaultOrEmpty` checks.

src/Sentry/SentryLog.cs
```diff
@@ -190,6 +200,8 @@ internal void SetOrigin(string origin)
 
     internal void WriteTo(Utf8JsonWriter writer, IDiagnosticLogger? logger)
     {
+        Debug.Assert(!Parameters.IsDefault);
```
Flash0ver, line 203: note: omit default checks

Since we ensured in the `internal` .ctor, as well as the `init-only` property, that the `ImmutableArray` is never `default`, and with that the underlying array never `null`, we can omit further `IsDefault` and `IsDefaultOrEmpty` checks.

test/Sentry.Tests/SentryMetricTests.cs
```diff
@@ -22,6 +22,20 @@ public SentryMetricTests(ITestOutputHelper output)
         _output = new TestOutputDiagnosticLogger(output);
     }
 
+    [Fact]
+    public void Create_Default_HasMinimalSpecification()
```
Flash0ver, line 26: note: I wanted to keep the tests for both Logs and Metrics in sync

src/Sentry/Internal/DefaultSentryStructuredLogger.cs
Flash0ver: and a follow-up question:

Shall we note this in the CHANGELOG?
E.g. in a separate category: `### Performance ⚡`

src/Sentry/Internal/DefaultSentryStructuredLogger.cs
```diff
@@ -30,26 +30,33 @@ private protected override void CaptureLog(SentryLogLevel level, string template
...
-            return;
-        }
+        ImmutableArray<KeyValuePair<string, object>> @params;
 
-        ImmutableArray<KeyValuePair<string, object>> @params = default;
         if (parameters is { Length: > 0 })
         {
+            try
+            {
+                message = string.Format(CultureInfo.InvariantCulture, template, parameters);
```
Flash0ver, line 39: question: acceptable change?

This is a behavioral change ... but not a breaking change as it becomes less restrictive:

#### Previously:
`System.String.Format` was always invoked,
leading `SentrySdk.Logger.LogInfo("Text {0} Text", []);` (with _composite format_, but without _arguments_)
to throwing a `System.FormatException`,
resulting in this log not being emitted.

#### With this change:
`System.String.Format` is only invoked when at least one _argument_ is passed,
leading `SentrySdk.Logger.LogInfo("Text {0} Text", []);` (with _composite format_, but without _arguments_)
to not throwing a `System.FormatException`,
resulting in this log actually being emitted.

#### TL;DR

Should e.g. `SentrySdk.Logger.LogInfo("Text {0} Text", []);`
emit a log or not emit a log?

jamescrosswell:
@Flash0ver converted to draft until the CI checks go green. Maybe needs updating from main?

test/Sentry.Tests/SentryLogTests.cs
```diff
@@ -22,6 +22,28 @@ public SentryLogTests(ITestOutputHelper output)
...
+        log.Parameters.Should().BeEmpty();
+        log.SpanId.Should().Be(null);
+
+#if DEBUG
+        var assignDefault = static () => new SentryLog(Timestamp, TraceId, (SentryLogLevel)24, "message")
+        {
+            Parameters = default,
+        };
+        assignDefault.Should().Throw<Exception>("Disallow default ImmutableArray");
+#endif
```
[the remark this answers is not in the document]
Flash0ver: Hence the `#if DEBUG`.

But we could argue that the change under test
- https://github.com/getsentry/sentry-dotnet/pull/4697/changes#diff-28d1620ef7e0107e355fa48fb9e9d9be34390e6e355d24e4e0d26d99a3745c8dR68

which ensures that we never set a `default` instance of `ImmutableArray`, which user code cannot do due to the `.ctor` of the containing type `SentryLog` being `internal` and the accessor `init`-only.

When the instance of `ImmutableArray` is never `default`, we can omit these two `default` checks:
- https://github.com/getsentry/sentry-dotnet/pull/4697/changes#diff-28d1620ef7e0107e355fa48fb9e9d9be34390e6e355d24e4e0d26d99a3745c8dL225
- https://github.com/getsentry/sentry-dotnet/pull/4697/changes#diff-28d1620ef7e0107e355fa48fb9e9d9be34390e6e355d24e4e0d26d99a3745c8dL230

I can imagine that this assertion is a bit overkill, and we can safely remove it, but still keeping the added `Debug.Assert` invocations.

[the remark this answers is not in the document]
Flash0ver: Removed the DEBUG-only assertion via https://github.com/getsentry/sentry-dotnet/pull/4697/changes/1dc4391f50b6b814c79cb709448b7cf62ace4570,
while keeping the `Debug.Assert` invocation to guard against internal contract violations.

src/Sentry/Internal/DefaultSentryStructuredLogger.cs
```diff
@@ -30,26 +30,33 @@ private protected override void CaptureLog(SentryLogLevel level, string template
...
             {
                 builder.Add(new KeyValuePair<string, object>(index.ToString(), parameters[index]));
             }
             @params = builder.DrainToImmutable();
         }
+        else
+        {
+            message = template;
+            template = null!; // SentryLog.Template is declared nullable (string?)
+            @params = ImmutableArray<KeyValuePair<string, object>>.Empty;
```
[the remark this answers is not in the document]
Flash0ver: Also discussed here: https://github.com/getsentry/sentry-dotnet/pull/4697#discussion_r3064471546

src/Sentry/Internal/DefaultSentryStructuredLogger.cs
Flash0ver: For comparison, behavior of `Microsoft.Extensions.Logging`:

Example: Console-Logger

```CSharp
// with args
logger.LogInformation("Parameters: none", 1, 2); // Parameters: none
logger.LogInformation("Parameters: {one}", 1, 2); // Parameters: 1
logger.LogInformation("Parameters: {one}, {two}", 1, 2); // Parameters: 1, 2
logger.LogInformation("Parameters: {one}, {two}, {three}", 1, 2); //throws ---> System.FormatException: Index (zero based) must be greater than or equal to zero and less than the size of the argument list.

// without args
logger.LogInformation("Parameters: none"); // Parameters: none
logger.LogInformation("Parameters: {one}"); // Parameters: {one}
logger.LogInformation("Parameters: {one}, {two}"); // Parameters: {one}, {two}
logger.LogInformation("Parameters: {one}, {two}, {three}"); // Parameters: {one}, {two}, {three}
```

jamescrosswell (approved):

PR #4723, opened 2025-11-11
bump to .NET 10.0.100 (GA)

Flash0ver:
#skip-changelog

> [!NOTE]
> I created a separate followup task to deal with additional integration tests: 
> - https://github.com/getsentry/sentry-dotnet/issues/4744

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -265,21 +265,22 @@ jobs:
           path: src
 
       - name: Integration test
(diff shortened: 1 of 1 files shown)


jamescrosswell:
> do we want to update MAUI/Mobile integration tests to .NET 10 now ... or later (when .NET 9 goes out of support)

Ideally we'd run all integration tests against all supported targets (so 9 and 10). The only reason we weren't doing this in the preview1/2 releases was because net10-ios26 didn't exist then. We should no though... this is an important sanity check that everything is working properly.

jamescrosswell (approved):

PR #4724, opened 2025-11-11
Improve SentryHttpFailedRequestHandler issue grouping

### LINKED ISSUE #3582: SentryHttpMessageHandler issue grouping is too eager
### Problem Statement

When using the SentryHttpMessageHandler, errors from all called URLs get lumped into one Issue. That's not great as the grouping is too eager; it's not very helpful to bundle HTTP errors from 25 different microservices and 15 external SaaS providers into a single Issue.

### Solution Brainstorm

I'd love if the grouping could use an auto-parameterized version of the called URL as a grouping key.

### Follow-up
- #4759

ericsampson:
Resolves #3582 

With the current SentryHttpFailedRequestHandler, the captured exception stack trace is always the same two frames; this leads to all events getting lumped into a single Issue, even though the application code path (and remote API) are totally different. As you can imagine, this is less than ideal :)

This PR addresses this issue by capturing a full stack trace on .NET 6+ using `ExceptionDispatchInfo.SetCurrentStackTrace`.

Tests have been expanded and improved. The test of the new functionality (`HandleResponse_StackTraceIncludesCallerContext`) was verified to fail using the existing code, and pass using the new code.

Very similar changes could likely be made to the `SentryGraphQLHttpFailedRequestHandler.cs` to give the same improvement, but I didn't want to bundle too many changes together into one PR.

I'd love to get this into 5.16.3 or 6.0.0 if possible, depending on the release timing :)

Thanks!


src/Sentry/HttpStatusCodeRangeExtensions.cs
```diff
@@ -0,0 +1,25 @@
...
+/// </summary>
+internal static class HttpStatusCodeRangeExtensions
+{
+    /// <summary>
+    /// Checks if any range in the collection contains the given status code.
+    /// </summary>
+    /// <param name="ranges">Collection of ranges to check.</param>
+    /// <param name="statusCode">Status code to check.</param>
+    /// <returns>True if any range contains the given status code.</returns>
+    internal static bool ContainsStatusCode(this IEnumerable<HttpStatusCodeRange> ranges, int statusCode)
```
ericsampson, line 14: Couldn't use the standard method name `Contains` because in that case the implicit conversion operators in HttpStatusCodeRange resulted in these extension methods not getting called.

test/Sentry.Tests/SentryHttpFailedRequestHandlerTests.cs
```diff
@@ -36,26 +36,119 @@ public void HandleResponse_Disabled_DontCapture()
...
-        var response = ForbiddenResponse(); // 403 is not in default range (500-599)
+        var response = new HttpResponseMessage((HttpStatusCode)statusCode);
         response.RequestMessage = new HttpRequestMessage();
 
         // Act
         sut?.HandleResponse(response);
 
         // Assert
-        _hub?.DidNotReceive().CaptureEvent(Arg.Any<SentryEvent>());
+        _hub?.DidNotReceiveWithAnyArgs().CaptureEvent(null!);
```
ericsampson, line 73: These tests weren't actually going to catch failures due to the use of `DidNotReceive().CaptureEvent(Arg.Any<SentryEvent>());`, which behaves in a subtle footgun manner for capturing methods that have optional parameters.

It's much safer to use `DidNotReceiveWithAnyArgs` and `ReceivedWithAnyArgs` in these scenarios, unless you're careful to specify an explicit value for every method parameter (whether Any or a specific value)

I made the same tweak throughout the rest of this test file.

src/Sentry/SentryHttpFailedRequestHandler.cs
```diff
@@ -15,67 +15,84 @@ internal SentryHttpFailedRequestHandler(IHub hub, SentryOptions options)
...
+            string.Format(
+                System.Globalization.CultureInfo.InvariantCulture,
+                "Response status code does not indicate success: {0}",
+                statusCode)
+            );
+
+#if NET6_0_OR_GREATER
+        // Add a full stack trace into the exception to improve Issue grouping,
+        // see https://github.com/getsentry/sentry-dotnet/issues/3582
+        ExceptionDispatchInfo.SetRemoteStackTrace(exception, Environment.StackTrace);
```
ericsampson: The meat of the PR.

jamescrosswell:
This looks awesome @ericsampson - we're a bit tight on time but will definitely do our best to get this reviewed and included in version 6... it's a slight change in behaviour so might be good to do this in the major release.

ericsampson:
Totally understood @jamescrosswell! Thanks. Would it be worthwhile to start just by approving all the GH workflows to run to make sure that they don't surface any issues.

1) Do you think we could get it into a 6.0.0 beta release? That way I can test it For Real in our prod API without having to vendor the repo into our codebase. If that sounds like a good plan, LMK if I should change this PR base branch and to what.

1b) Should I add a release note line?

2) Given the "slight change in behavior", do you think that I should add a way that customers can opt-out of this change in case that it turns out affecting them in a way that they don't like? I could potentially add it to `SentryExperimentalOptions`.

Best,
Eric

src/Sentry/SentryHttpFailedRequestHandler.cs
```diff
@@ -15,67 +15,84 @@ internal SentryHttpFailedRequestHandler(IHub hub, SentryOptions options)
...
+        catch (HttpRequestException ex)
         {
-            exception.SetSentryMechanism(MechanismType);
+            exception = ex;
+        }
+#endif
 
-            var @event = new SentryEvent(exception);
-            var hint = new SentryHint(HintTypes.HttpResponseMessage, response);
+        exception.SetSentryMechanism(MechanismType);
```
ericsampson, line 56: everything below here in the file is only appears changed due to indenting

src/Sentry/SentryHttpFailedRequestHandler.cs
```diff
@@ -15,67 +15,84 @@ internal SentryHttpFailedRequestHandler(IHub hub, SentryOptions options)
     protected internal override void DoEnsureSuccessfulResponse([NotNull] HttpRequestMessage request, [NotNull] HttpResponseMessage response)
```
Flash0ver, line 15: suggestion: update from `main`, as we've merged the big `version6` branch adding support for .NET 10 in the meantime

Flash0ver, line 15: suggestion: add a `CHANGELOG.md` entry briefly describing the change in behavior

src/Sentry/Internal/Extensions/HttpStatusExtensions.cs
```diff
@@ -1,28 +0,0 @@
-namespace Sentry.Internal.Extensions;
-
-#if !NET5_0_OR_GREATER
-internal static class HttpStatusExtensions
```
Flash0ver, line 4: suggestion: Also remove the related tests `Sentry.Tests.Internals.Extensions.HttpRequestExceptionMessageTests`

src/Sentry/SentryHttpFailedRequestHandler.cs
Flash0ver: question: Would `ExceptionDispatchInfo.SetCurrentStackTrace(Exception)` work, too?
Or what's the difference between these two variants?

test/Sentry.Tests/HttpStatusCodeRangeExtensionsTests.cs
```diff
@@ -0,0 +1,188 @@
+namespace Sentry.Tests;
+
+public class HttpStatusCodeRangeExtensionsTests
+{
+    [Fact]
+    public void Contains_EmptyList_ReturnsFalse()
```
Flash0ver: suggestion: rename the first portion of each test name in this file to match the Method-Under-Test:

```diff
-Contains_EmptyList_ReturnsFalse()
+ContainsStatusCode_EmptyList_ReturnsFalse()
```

test/Sentry.Tests/HttpStatusCodeRangeExtensionsTests.cs
```diff
@@ -0,0 +1,188 @@
...
+
+        // Assert
+        result.Should().BeTrue();
+    }
+
+    [Theory]
+    [InlineData(200)]
+    [InlineData(399)]
+    [InlineData(476)]
+    [InlineData(500)]
```
Flash0ver: suggestion(nitpick): would you mind adding explanatory comments to these four tests as well? ... just as with the test above ... for consistency


```suggestion
    [InlineData(200)] // Below first range
    [InlineData(399)] // Below first range
    [InlineData(476)] // Above second range
    [InlineData(500)] // Above second range
```

test/Sentry.Tests/HttpStatusCodeRangeExtensionsTests.cs
```diff
@@ -0,0 +1,188 @@
...
+        var result = ranges.ContainsStatusCode(statusCode);
+
+        // Assert
+        result.Should().BeTrue();
+    }
+
+    [Theory]
+    [InlineData(399)] // Just before start
+    [InlineData(500)] // Just after end
+    public void Contains_JustOutsideBoundaries_ReturnsFalse(int statusCode)
```
Flash0ver: question: Is this test redundant?

Aren't these cases already covered by `Contains_SingleRange_OutOfRange_ReturnsFalse`?

test/Sentry.Tests/HttpStatusCodeRangeExtensionsTests.cs
```diff
@@ -0,0 +1,188 @@
...
+        var result = ranges.ContainsStatusCode(HttpStatusCode.NotFound); // 404
+
+        // Assert
+        result.Should().BeTrue();
+    }
+
+    [Theory]
+    [InlineData(400)] // Start of range
+    [InlineData(499)] // End of range
+    public void Contains_RangeBoundaries_ReturnsTrue(int statusCode)
```
Flash0ver: question: Is this test redundant?

Aren't these cases already covered by `Contains_SingleRange_InRange_ReturnsTrue`?

test/Sentry.Tests/HttpStatusCodeRangeExtensionsTests.cs
```diff
@@ -0,0 +1,188 @@
...
+
+        // Act
+        var result = ranges.ContainsStatusCode(statusCode);
+
+        // Assert
+        result.Should().BeFalse();
+    }
+
+    [Fact]
+    public void Contains_HttpStatusCodeEnum_CallsIntOverload()
```
Flash0ver: suggestion: rename

I do like the test, but not the name: The name is describing implementation details that the test-case itself is not dependent on. Should we change the implementation detail, it's very likely that we will miss to rename this test-case.

```suggestion
    public void Contains_HttpStatusCodeEnum_InRange_ReturnsTrue()
```

To be consistent with the existing test case `Contains_SingleRange_InRange_ReturnsTrue`.

Flash0ver: suggestion: Add `false` test with `enum`.

This case is a "positive-enum" case ... let's also add a "negative-enum" case,
called
`public void Contains_HttpStatusCodeEnum_OutOfRange_ReturnsFalse()`

To be consistent with the existing test case: `Contains_SingleRange_OutOfRange_ReturnsFalse`.

test/Sentry.Tests/SentryHttpFailedRequestHandlerTests.cs
Flash0ver: Great catch!

test/Sentry.Tests/SentryHttpFailedRequestHandlerTests.cs
```diff
@@ -36,26 +36,119 @@ public void HandleResponse_Disabled_DontCapture()
...
+    [InlineData(404)] // Not Found
+    [InlineData(499)] // Edge of client error range
+    [InlineData(600)] // Beyond standard range
+    public void HandleResponse_EnabledButNotInRange_DontCapture(int statusCode)
     {
         // Arrange
         var sut = GetSut(new SentryOptions
         {
             CaptureFailedRequests = true
+            // default FailedRequestStatusCodes = (500,599)
```
Flash0ver: thought: I would remove this comment, because it is hard to track: Should we ever change the default range, we will very likely forget to update this comment.

On the other hand, it is quite unlikely that the default range will change.
So I'm wondering what your thoughts are on that.

Alternatively, we could rename the test to clarify that it's testing the default range:
```diff
-HandleResponse_EnabledButNotInRange_DontCapture
+HandleResponse_EnabledButNotInDefaultRange_DontCapture
```

test/Sentry.Tests/SentryHttpFailedRequestHandlerTests.cs
```diff
@@ -36,26 +36,119 @@ public void HandleResponse_Disabled_DontCapture()
...
+        // Assert
+        _hub.DidNotReceiveWithAnyArgs().CaptureEvent(null!);
+    }
+
+    [Theory]
+    [InlineData(200)] // OK
+    [InlineData(201)] // Created
+    [InlineData(204)] // No Content
+    [InlineData(299)] // Edge of success range
+    public void HandleResponse_RangeIncludesSuccess_SuccessNotCaptured(int statusCode)
```
Flash0ver: question: Is this the same behavior as it was before this change?

I'm asking whether we need to mention this in the `CHANGELOG`, too.
But if I get it right, this test is now asserting the previous behavior, that we had implicitly by using `EnsureSuccessStatusCode`, right?

test/Sentry.Tests/SentryHttpFailedRequestHandlerTests.cs
```diff
@@ -265,4 +347,62 @@ public void HandleResponse_Hint_Response()
             hint.Items[HintTypes.HttpResponseMessage].Should().Be(response);
         }
     }
+
+    [Fact]
+    public void HandleResponse_ExceptionHasStackTrace()
```
Flash0ver: suggestion: Add a "scenario-part" in the middle of the test name, to be compliant with out `CONTRIBUTING.md`.

Maybe something like

```suggestion
    public void HandleResponse_Capture_ExceptionHasStackTrace()
```

test/Sentry.Tests/SentryHttpFailedRequestHandlerTests.cs
```diff
@@ -265,4 +347,62 @@ public void HandleResponse_Hint_Response()
...
+        {
+            @event.Should().NotBeNull();
+            @event.Exception.Should().NotBeNull();
+            @event.Exception!.StackTrace.Should().NotBeNullOrWhiteSpace();
+        }
+    }
+
+#if NET6_0_OR_GREATER // This test is only valid on .NET 6+ where we can use SetRemoteStackTrace
+    [Fact]
+    public void HandleResponse_StackTraceIncludesCallerContext()
```
Flash0ver: suggestion: Add a "scenario-part" in the middle of the test name, to be compliant with out `CONTRIBUTING.md`.

Maybe something like

```suggestion
    public void HandleResponse_Capture_StackTraceIncludesCallerContext()
```

src/Sentry/SentryHttpFailedRequestHandler.cs
Flash0ver: TODO @Flash0ver approve workflows after update
(currently the build fails ... see `test/Sentry.Tests/Internals/Extensions/HttpRequestExceptionMessageTests.cs`) and https://github.com/getsentry/sentry-dotnet/pull/4724#discussion_r2533782372)

ericsampson:
@Flash0ver thanks for the review! Comments make sense, I'll work through them.

ericsampson:
@Flash0ver 
Given the "slight change in behavior", do you think that I should add a way that customers can opt-out of this change in case that it turns out affecting them in a way that they don't like? I could potentially add it to SentryExperimentalOptions.

src/Sentry/Internal/Extensions/HttpStatusExtensions.cs
ericsampson: Done. Missed seeing that due to the conditional compilation sorry!

src/Sentry/SentryHttpFailedRequestHandler.cs
ericsampson: I thought I had tested that during development because it's the logical choice and it wasn't doing what I needed, but I just switched back to it and the unit tests still pass and the captured stacktrace looks good. So I switched to this call because it's certainly the one that makes the most sense :) Cheers

src/Sentry/SentryHttpFailedRequestHandler.cs
ericsampson: Updated from main

test/Sentry.Tests/SentryHttpFailedRequestHandlerTests.cs
ericsampson: > this test is now asserting the previous behavior, that we had implicitly by using EnsureSuccessStatusCode, right?

That's exactly correct--I wanted the test to ensure that we maintain that contract as a result of this PR and going forward.
So no changelog entry should be needed.

test/Sentry.Tests/HttpStatusCodeRangeExtensionsTests.cs
ericsampson: yeah true. removed

test/Sentry.Tests/HttpStatusCodeRangeExtensionsTests.cs
ericsampson: agreed, removed.

test/Sentry.Tests/HttpStatusCodeRangeExtensionsTests.cs
ericsampson: resolved.

ericsampson: resolved.

test/Sentry.Tests/HttpStatusCodeRangeExtensionsTests.cs
ericsampson: Done yup.

test/Sentry.Tests/SentryHttpFailedRequestHandlerTests.cs
ericsampson: All fair feedback. I changed the test name and removed the comment. If the default range changes (though unlikely), the test will fail and then can be adjusted.

src/Sentry/SentryHttpFailedRequestHandler.cs
```diff
@@ -15,67 +15,84 @@ internal SentryHttpFailedRequestHandler(IHub hub, SentryOptions options)
...
+        var statusCode = (int)response.StatusCode;
+        // Match behavior of HttpResponseMessage.EnsureSuccessStatusCode
+        if (statusCode >= 200 && statusCode <= 299)
         {
-#if NET5_0_OR_GREATER
-            response.EnsureSuccessStatusCode();
+            return;
+        }
+
+        var exception = new HttpRequestException(
```
Flash0ver, line 33: > @Flash0ver Given the "slight change in behavior", do you think that I should add a way that customers can opt-out of this change in case that it turns out affecting them in a way that they don't like? I could potentially add it to SentryExperimentalOptions.

Hmmm .... I haven't thought about it ... but I believe that providing an opt-out is a great idea!
I'm personally uncertain if there should be an API for it ... it feels a bit odd that there is an "experimental" API for something that will be removed in the future, as usually we have "experimental" options for something that is added in a stable fashion in the future.
An alternative could be a runtime-switch, but the discoverability on that is pretty much through documentation only.

The advantage of an API opt-out via Options is that users will notice when we remove the opt-out.
Which is the drawback of a runtime-switch opt-out, as the opt-out just stops working once we remove it.

A second question is:
Once we remove the opt-out, we kinda need to keep it until v7 in a year, as it's a breaking change when we remove the opt-out.

We could keep the opt-out in mind, and add it, if users face problems with the new behavior,
and then think about whether we ship it as an API or a Runtime-Switch opt-out,
and how long we support that opt-out.

If we do provide an opt-out now, I personally prefer a runtime-switch over an Experimental-Option-Flag, and mention it in our documentation / upgrade-guides.

@ericsampson @jamescrosswell What do you think?

ericsampson: Maybe the KISS option is to do nothing now (especially since this would be going out as part of a Major release), and then if someone(s) actually report an issue then decide the right way to allow opt-out as it should be a quick turnaround to put out a patch version with that functionality.

src/Sentry/SentryHttpFailedRequestHandler.cs
ericsampson: Would you suggest this go in Features or Fixes?

ericsampson: Added it to Fixes, LMK if it looks ok.

src/Sentry/SentryHttpFailedRequestHandler.cs
jamescrosswell: Yeah this feels more like a fix than a change in behaviour to me... I'm OK to ship without an opt out and add one if we get anyone ask for it.

src/Sentry/SentryHttpFailedRequestHandler.cs
```diff
@@ -15,67 +15,84 @@ internal SentryHttpFailedRequestHandler(IHub hub, SentryOptions options)
...
+        if (!Options.FailedRequestStatusCodes.ContainsStatusCode(response.StatusCode))
         {
             return;
         }
 
         // Capture the event
-        try
+
+        var statusCode = (int)response.StatusCode;
+        // Match behavior of HttpResponseMessage.EnsureSuccessStatusCode
```
jamescrosswell, line 26:
```suggestion
        // Match behavior of HttpResponseMessage.EnsureSuccessStatusCode
        // See https://github.com/getsentry/sentry-dotnet/issues/2684
```

Good to preserve where this came from originally...

src/Sentry/HttpStatusCodeRangeExtensions.cs
jamescrosswell, line 1: I hesitated about whether or not we need the new extension method (which is pretty trivial and only used in one place in the code base)... it does make the calling code slightly easier to read though so OK to leave this in here.

ericsampson: Yeah for sure--the main reason I actually ended up adding it was that it allowed the tests to be better separated. Otherwise there were a bunch of tests of the `SentryHttpFailedRequestHandler` that were really just testing the `FailedRequestStatusCodes` handling in all variety of scenarios and became more verbose due to having to setup/run/capture hub events.

With a dedicated extension methods and corresponding tests, the ContainsStatusCode tests are very simple and comprehensive, and the SentryHttpFailedRequestHandler don't have to repeat them.

ericsampson:
@Flash0ver want to do a final re-review whenever you get a chance? Cheers

test/Sentry.Tests/SentryHttpFailedRequestHandlerTests.cs
```diff
@@ -250,10 +337,7 @@ public void HandleResponse_Hint_Response()
 
         // Act
         SentryHint hint = null;
-        ((ISentryClient)_hub).CaptureEvent(
-            Arg.Any<SentryEvent>(),
-            hint: Arg.Do<SentryHint>(h => hint = h)
-            );
+        _hub.CaptureEvent( Arg.Any<SentryEvent>(), hint: Arg.Do<SentryHint>(h => hint = h));
```
Flash0ver: suggestion: `dotnet format`

That's what's failing the workflow.

```suggestion
        _hub.CaptureEvent(Arg.Any<SentryEvent>(), hint: Arg.Do<SentryHint>(h => hint = h));
```

Flash0ver: question: push access

@jamescrosswell I wasn't able to push to the PR.
Am I missing a configuration / permission?
Or is this a setting from the fork's repository?

Context:
I wanted to push the missing `dotnet format` change:
```diff
-_hub.CaptureEvent( Arg.Any<SentryEvent>(), hint: Arg.Do<SentryHint>(h => hint = h));
+_hub.CaptureEvent(Arg.Any<SentryEvent>(), hint: Arg.Do<SentryHint>(h => hint = h));
```

src/Sentry/SentryHttpFailedRequestHandler.cs
Flash0ver: I wasn't all to sure either.
We mentioned grouping-related changes in both `Features` and `Fixes` sections previously.
Fixes sounds good here.

CHANGELOG.md
```diff
@@ -42,6 +42,7 @@
 - The `Serilog` integration captures _Structured Logs_ (when enabled) independently of captured Events and added Breadcrumbs ([#4691](https://github.com/getsentry/sentry-dotnet/pull/4691))
 - The SDK avoids redundant scope sync after transaction finish ([#4623](https://github.com/getsentry/sentry-dotnet/pull/4623))
 - sentry-native is now automatically disabled for WASM applications ([#4631](https://github.com/getsentry/sentry-dotnet/pull/4631))
+- Captured [Http Client Errors](https://docs.sentry.io/platforms/dotnet/guides/aspnet/configuration/http-client-errors/) on .NET 6+ now include a full stack trace in order to improve Issue grouping ([#4724](https://github.com/getsentry/sentry-dotnet/pull/4724))
```
Flash0ver: issue: I'm afraid we need one more merge from `main`, after the `6.0.0-rc.1` release, so this line needs to move up into the new `## Unreleased` section.

src/Sentry/Internal/Extensions/HttpStatusExtensions.cs
Flash0ver: Nothing to apologize ... that's what we got reviews for 😉

src/Sentry/SentryHttpFailedRequestHandler.cs
```diff
@@ -15,67 +15,85 @@ internal SentryHttpFailedRequestHandler(IHub hub, SentryOptions options)
...
+            string.Format(
+                System.Globalization.CultureInfo.InvariantCulture,
+                "Response status code does not indicate success: {0}",
+                statusCode)
+            );
+
+#if NET6_0_OR_GREATER
+        // Add a full stack trace into the exception to improve Issue grouping,
+        // see https://github.com/getsentry/sentry-dotnet/issues/3582
+        ExceptionDispatchInfo.SetCurrentStackTrace(exception);
```
Flash0ver: suggestion: totally a nitpick ... but [ExceptionDispatchInfo.SetCurrentStackTrace(Exception)](https://learn.microsoft.com/dotnet/api/system.runtime.exceptionservices.exceptiondispatchinfo.setcurrentstacktrace) was added in `net5.0`


```suggestion
#if NET5_0_OR_GREATER
        // Add a full stack trace into the exception to improve Issue grouping,
        // see https://github.com/getsentry/sentry-dotnet/issues/3582
        ExceptionDispatchInfo.SetCurrentStackTrace(exception);
```

Flash0ver:
@sentry review

src/Sentry/SentryHttpFailedRequestHandler.cs
ericsampson: You're right, it is SetRemoteStackTrace that was added in .NET 6. SetCurrentStackTrace was available in .NET 5, so I'll update all that!

CHANGELOG.md
ericsampson: sounds good

test/Sentry.Tests/SentryHttpFailedRequestHandlerTests.cs
ericsampson: I imagine it's because I forked the repo into my company's GH Org. If we do more PRs in the future we can make this easier :)

CHANGELOG.md
ericsampson: LMK if I did it correctly thanks.

test/Sentry.Tests/SentryHttpFailedRequestHandlerTests.cs
jamescrosswell: @ericsampson probably the easiest is if you run that format command locally and commit the changes. That way when this gets run by CI, there are on changes to commit and it doesn't fall over when executing this workflow.

Here's what you'd need to run:
https://github.com/getsentry/sentry-dotnet/blob/b3d9b2ffe733bd2515073a7edd8a6d1937055894/.github/workflows/format-code.yml#L37

CHANGELOG.md
```diff
@@ -42,6 +48,7 @@
 - The `Serilog` integration captures _Structured Logs_ (when enabled) independently of captured Events and added Breadcrumbs ([#4691](https://github.com/getsentry/sentry-dotnet/pull/4691))
 - The SDK avoids redundant scope sync after transaction finish ([#4623](https://github.com/getsentry/sentry-dotnet/pull/4623))
 - sentry-native is now automatically disabled for WASM applications ([#4631](https://github.com/getsentry/sentry-dotnet/pull/4631))
+- Captured [Http Client Errors](https://docs.sentry.io/platforms/dotnet/guides/aspnet/configuration/http-client-errors/) on .NET 5+ now include a full stack trace in order to improve Issue grouping ([#4724](https://github.com/getsentry/sentry-dotnet/pull/4724))
```
```suggestion
```

CHANGELOG.md
jamescrosswell: > LMK if I did it correctly thanks.

Almost... just need to [remove from the rc1 release](https://github.com/getsentry/sentry-dotnet/pull/4724/files#r2539878440) as well.

Flash0ver (approved):

Flash0ver:
@ericsampson, all ✅ from a review perspective.

There are merge conflicts, that I'm afraid we cannot view nor fix as we don't have push permissions.
I can imagine it's in the `CHANGELOG.md`. Perhaps we can think of a different way of adding entries to the `CHANGELOG.md`, as this is most commonly the merge conflict ... maybe we don't add it directly, but place it "somewhere" from where it's then pushed to the `CHANGELOG.md` in the correct place when (squash) merging 🤔 ... something for us to think about.

Anyway, sorry for that! We'll merge the PR after resolving.

ericsampson:
@Flash0ver @jamescrosswell thanks for your help getting this through!

Having outside contributors add their updates in a separate temporary `CHANGES.md` file to prevent merge conflicts might be a good thing.

jamescrosswell:
> @Flash0ver @jamescrosswell thanks for your help getting this through!
> 
> Having outside contributors add their updates in a separate temporary `CHANGES.md` file to prevent merge conflicts might be a good thing.

It's a problem that affects us when making changes from the same repository, to be honest. Fixing it likely requires changes to our release process (which is shared across Sentry) but yes, it's been a pain for long enough that we should do something about it.

jamescrosswell:
@ericsampson just merged. Huge thank you once again for the contribution!

PR #4747, opened 2025-11-18
fix: Sentry Tracing middleware crashes ASP.NET Core in .NET 10 in version 6.0.0-rc1 and earlier

### LINKED ISSUE #4735: Sentry Tracing crashes ASP.NET web apps on .NET 10
### Package

Sentry

### .NET Flavor

.NET

### .NET Version

10.0.0

### OS

Any (not platform specific)

### OS Version

_No response_

### Development Environment

Other

### SDK Version

5.14.1

_No response_

### Workload Versions

Doesn't matter, blank test project works to showcase

```cs
var builder = WebApplication.CreateBuilder(args);

builder.WebHost.UseSentry(x => ConfigureSentry(x, builder.Environment));

static void ConfigureSentry(SentryAspNetCoreOptions options, IWebHostEnvironment env)
{
    options.Dsn = "https://abc123@sentry.io.com/2";
    options.AutoRegisterTracing = true;
}
```

### Steps to Reproduce

1. Clone https://github.com/Mitch528/sentry-crash-repro
2. Run `dotnet run`
3. Open the home page
4. Observe that the app crashes

### Expected Result

The app shouldn't crash, the error page should be shown

### Actual Result

CLI process terminates. Disabling `options.AutoRegisterTracing = true;` removes the issue. Not having `await` before the crash also removes the issue so the async pipeline might also be related.

We also believe that https://github.com/dotnet/runtime/issues/121578 might be related to this issue.

jamescrosswell:
Fixes #4735:
- #4735

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Fixes
 
 - Sentry Tracing middleware crashed ASP.NET Core in .NET 10 in 6.0.0-rc.1 and earlier ([#4747](https://github.com/getsentry/sentry-dotnet/pull/4747))
+
 ### Dependencies
 
 - Bump Java SDK from v8.24.0 to v8.26.0 ([#4728](https://github.com/getsentry/sentry-dotnet/pull/4728))


Flash0ver:
@sentry review

src/Sentry.AspNetCore/SentryTracingMiddleware.cs
```diff
@@ -140,6 +140,9 @@ public async Task InvokeAsync(HttpContext context)
         catch (Exception e)
         {
             exception = e;
+            // Rethrow immediately so as not to disrupt the .net 10 pipeline behaviour
+            // See: https://github.com/getsentry/sentry-dotnet/issues/4735
+            ExceptionDispatchInfo.Capture(e).Throw();
```
Flash0ver: question: Change in StackTrace

Oh wow ... great catch.
I believe this changes the StackTrace of the Exception slightly.
Is this alright ... may this cause some changes in e.g. grouping?

Flash0ver: thought: Conditional

Do we want to make this conditional for `NET10_0` only?
But then we have a slightly different StackTrace depending on the runtime.

Do we want to remove this again with the next .NET SDK / .NET Runtime patch release?
But then applications not upgrading will start crashing again.

I guess I'm convincing myself that this change has come to stay.

Especially considering the "Best practices for exceptions"
[Don't raise exceptions in finally clauses](https://learn.microsoft.com/dotnet/standard/exceptions/best-practices-for-exceptions#dont-raise-exceptions-in-finally-clauses)

Unfortunately, [CA2219](https://learn.microsoft.com/dotnet/fundamentals/code-analysis/quality-rules/ca2219) only reports `throw`-clauses, but not the invocation of `ExceptionDispatchInfo.Throw` methods (both instance and static).

Flash0ver: question: `throw` statement vs `ExceptionDispatchInfo.Throw` invocation

Which one do we prefer?

I believe that in the case of a `throw` statement, this stack frame with _line 145_ would NOT be included.
And I think that with the current `ExceptionDispatchInfo.Throw` invocation, this stack frame with _line 145_ IS included.

I can imagine we want to keep the `ExceptionDispatchInfo.Throw` invocation to not change all to much about the StackTrace.

test/Sentry.AspNetCore.Tests/SentryTracingMiddlewareTests.cs
```diff
@@ -614,6 +614,60 @@ public async Task Transaction_binds_exception_thrown()
...
+            .UseSentry()
+            .ConfigureServices(services =>
+            {
+                services.RemoveAll(typeof(Func<IHub>));
+                services.AddSingleton<Func<IHub>>(() => hub);
+                services.AddRouting();
+            }).Configure(app =>
+            {
+                app.UseRouting();
+                app.UseSentryTracing();
```
Flash0ver: question: Explicitly redundant?

Since we call `.UseSentry()` in _line 631_ without configuring the options,
e.g. via `.UseSentry(o => o.AutoRegisterTracing = false)`,
the `SentryTracingMiddleware` is registered anyway,
regardless of _line 625_ (`AutoRegisterTracing = false`),
because the `Hub` is not aware of the derived `SentryAspNetCoreOptions`.

I personally find this a bit confusing,
because _line 625_ (`AutoRegisterTracing = false`) is essentially no-op,
`.UseSentry()` (_line 631_) registers `SentryTracingMiddleware` automatically,
and _line 640_ tries to register it again `app.UseSentryTracing();`.

To make the setup more clear/deliberate, we could:
1. remove _line 625_ `AutoRegisterTracing = false` from the `SentryAspNetCoreOptions` passed to the `Hub`
2. and instead call `.UseSentry()` in _line 631_ with this option
   - `.UseSentry(o => o.AutoRegisterTracing = false)`
3. so that _line 640_ (`app.UseSentryTracing();`) effectively does register the `SentryTracingMiddleware`

Is a bit of a nitpick, though ... but it could help clear things for future reads.

test/Sentry.AspNetCore.Tests/SentryTracingMiddlewareTests.cs
```diff
@@ -614,6 +614,60 @@ public async Task Transaction_binds_exception_thrown()
...
+        catch
+        {
+            // ignored
+        }
+
+        // Assert
+        // Make sure Kestrel is still alive by making another request
+        var response = await client.GetAsync("/");
+        response.StatusCode.Should().Be(System.Net.HttpStatusCode.OK);
+    }
```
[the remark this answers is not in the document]
Flash0ver: True.
When removing the re-throw entirely
https://github.com/getsentry/sentry-dotnet/blob/bd7b3d072c5e70a6127de6f3c08723030455dfe3/src/Sentry.AspNetCore/SentryTracingMiddleware.cs#L143-L145
no test in `SentryTracingMiddlewareTests` fails.

However, other tests are failing
- `Sentry.AspNetCore.Tests.IntegrationsTests.CaptureException_UseExceptionHandler_SetRouteDataFromInitialRequest`
- `Sentry.Google.Cloud.Functions.Tests.IntegrationTests.SentryIntegrationTest_CaptureUnhandledException`

Is it worth adding an assertion to `SentryTracingMiddlewareTests`, too?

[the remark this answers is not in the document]
jamescrosswell: It's not really the point of this fix. We could address it separately if we think we're lacking test coverage.

src/Sentry.AspNetCore/SentryTracingMiddleware.cs
jamescrosswell: I don't think users will notice any stack trace changes as a result of this as none of the frames concerned will be in app (so won't affect grouping).

test/Sentry.AspNetCore.Tests/SentryTracingMiddlewareTests.cs
jamescrosswell: I figured it's probably easier just to let auto tracing do it's thing... which means we can remove all of that code. It's not relevant/necessary for the test.

src/Sentry.AspNetCore/SentryTracingMiddleware.cs
jamescrosswell: Hm... I guess now we're doing it inside the `catch` block, we could just do a plain old `throw`. From what I can tell, when executed inside a catch block, `ExceptionDispatchInfo.Capture(e).Throw();` is functionally equivalent to `throw`... but `throw` is probably easier to read/understand, since folks (like me) don't bump into `ExceptionDispatchInfo` as often.

I imagine the reason @bruno-garcia originally did it with `ExceptionDispatchInfo.Capture(e).Throw()` is because it was being rethrown after all the code in the `finally` block was executed, so `throw` wasn't possible.

jamescrosswell: I think it simplifies our lives if we have one code path... so unless there's a really compelling reason why we want/need separate behaviour for .net 9 and earlier, let's leave it as is.

Flash0ver: I believe there IS a slight difference:
`ExceptionDispatchInfo.Capture(Exception)` (or capturing via the static `ExceptionDispatchInfo.Throw(Exception)` will include the current stack-frame / line while preserving the original StackTrace.
While just a re-throw via `throw;` does not change anything about the original StackTrace.

Flash0ver: Same as https://github.com/getsentry/sentry-dotnet/pull/4747#discussion_r2539707127 ... they're in-app anyway ... so let's not include the `ExceptionDispatchInfo` info since we are now within a `catch` block able to re-`throw;`.

tincann:
@Flash0ver Do you happen to know when this fix will be included in a release (or release candidate)? We’re currently blocked on our .NET 10 upgrade because of this crash, so any timing info would really help.

jamescrosswell:
> @Flash0ver Do you happen to know when this fix will be included in a release (or release candidate)? We’re currently blocked on our .NET 10 upgrade because of this crash, so any timing info would really help.

This was included in RC2, which just went out today:
- https://github.com/getsentry/sentry-dotnet/releases/tag/6.0.0-rc.2

PR #4748, opened 2025-11-18
ref: cleanup unused code after merge

Flash0ver:
I believe we were left with some now unused code after merges between
- #4498
- #4706

Removing code again from `SentryOptions`, that was removed in #4498, but accidentally restored in #4706.

### DIFF
--- a/src/Sentry/SentryOptions.cs
+++ b/src/Sentry/SentryOptions.cs
@@ -1854,31 +1854,6 @@ public class SentryOptions
         }
     }
 
-    internal string? TryGetDsnSpecificCacheDirectoryPath()
-    {
-        if (string.IsNullOrWhiteSpace(CacheDirectoryPath))
-        {
-            return null;
-        }
-
-        // DSN must be set to use caching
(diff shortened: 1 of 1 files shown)


Flash0ver:
@sentry review

src/Sentry/SentryOptions.cs
```diff
@@ -1854,31 +1854,6 @@ internal void LogDiagnosticWarning()
         }
     }
 
-    internal string? TryGetDsnSpecificCacheDirectoryPath()
```
Flash0ver, line 1857: question: no `#skip-changelog`

Why is the "Danger" workflow not failing, although I have not added `#skip-changelog` to the description, and pushed no changes to `CHANGELOG.md`?

jamescrosswell: curious

Flash0ver: ah ... because the Danger-Action does consider [Conventional Commits](https://www.conventionalcommits.org/) 💡

PR #4750, opened 2025-11-19
chore: Device/integration tests for mobile on .net10

### LINKED ISSUE #4744: MAUI/Mobile integration tests for .NET 10
### Description

MAUI/Mobile integration tests on .NET 10 as well
 integration-test/android.Tests.ps1
 integration-test/ios.Tests.ps1
 original PR: https://github.com/getsentry/sentry-dotnet/pull/4559

Splitting this off to a separate issue.

jamescrosswell:
Resolves #4744 

#skip-changelog

### DIFF
--- a/test/Sentry.Maui.Device.TestApp/Sentry.Maui.Device.TestApp.csproj
+++ b/test/Sentry.Maui.Device.TestApp/Sentry.Maui.Device.TestApp.csproj
@@ -36,7 +36,9 @@
 
     <SupportedOSPlatformVersion Condition="$([MSBuild]::GetTargetPlatformIdentifier('$(TargetFramework)')) == 'ios'">13.0</SupportedOSPlatformVersion>
     <SupportedOSPlatformVersion Condition="$([MSBuild]::GetTargetPlatformIdentifier('$(TargetFramework)')) == 'android'">21.0</SupportedOSPlatformVersion>
-
+    <!-- Pin target iOS version so that our tests don't break when new versions of Xcode are released.
+      'netX.0-ios' resolves the latest version of the iOS SDK otherwise. -->
+    <TargetPlatformVersion Condition="$([MSBuild]::GetTargetPlatformIdentifier('$(TargetFramework)')) == 'ios'">26.2</TargetPlatformVersion>
     <EmbedAssembliesIntoApk>true</EmbedAssembliesIntoApk>
 
     <!-- Avoid errors of Sentry.Testing not being a self-contained executable, while including it in a self-contained executable (this).
@@ -74,11 +76,6 @@
     <PackageReference Include="Microsoft.Maui.Controls" Version="$(MauiVersion)"/>
     <PackageReference Include="Microsoft.Maui.Core" Version="$(MauiVersion)"/>
     <PackageReference Include="Microsoft.Maui.Essentials" Version="$(MauiVersion)"/>
-  </ItemGroup>
-
-  <ItemGroup Condition="'$(TargetFramework)' == 'net9.0'">
-    <!-- https://github.com/advisories/GHSA-5f2m-466j-3848 -->
-    <PackageReference Include="System.Private.Uri" Version="4.3.2"/>
   </ItemGroup>
 
   <!-- Configure XUnit -->


jamescrosswell:
Tricky... locally everything passes for me:
image

However in CI we've got 2 tests failing in Debug and 2 tests failing in Release:
```
...
2025-11-20T03:50:51.4905577Z ##[error][-] MAUI app (net10.0, Release).Native crash (Release) 21.13s (21.12s|3ms)[0m
2025-11-20T03:50:51.4916597Z ##[group]Message[0m
2025-11-20T03:50:52.4056556Z Expected a collection with size 1, but got collection with size 25
2025-11-20T03:50:52.7693005Z at <ScriptBlock>, /home/runner/work/sentry-dotnet/sentry-dotnet/integration-test/android.Tests.ps1:170[0m
...
025-11-20T03:53:44.9710461Z ##[error][-] MAUI app (net10.0, Debug).Java crash (Debug) 10.9s (10.9s|1ms)[0m
2025-11-20T03:53:44.9723309Z ##[group]Message[0m,
2025-11-20T03:53:45.0186641Z Expected a collection with size 1, but got collection with size 2
2025-11-20T04:01:08.0804390Z at $result.Envelopes() | Should -HaveCount 1, /home/runner/work/sentry-dotnet/sentry-dotnet/integration-test/android.Tests.ps1:156[0m
...
2025-11-20T03:58:20.3428226Z ##[error][-] MAUI app (net10.0, Release).Native crash (Release) 22.23s (22.23s|3ms)[0m
2025-11-20T03:58:20.3429986Z ##[group]Message[0m
2025-11-20T03:58:21.0057104Z Expected a collection with size 1, but got collection with size 28
2025-11-20T03:50:52.7693005Z at <ScriptBlock>, /home/runner/work/sentry-dotnet/sentry-dotnet/integration-test/android.Tests.ps1:170[0m
...
2025-11-20T04:01:08.0369087Z ##[error][-] MAUI app (net10.0, Debug).Java crash (Debug) 9.43s (9.43s|1ms)[0m
2025-11-20T04:01:08.0387771Z ##[group]Message[0m
2025-11-20T04:01:08.0683949Z Expected a collection with size 1, but got collection with size 2
2025-11-20T04:01:08.0804390Z at $result.Envelopes() | Should -HaveCount 1, /home/runner/work/sentry-dotnet/sentry-dotnet/integration-test/android.Tests.ps1:156[0m
...
```

Looking at the 2 envelopes received for the Java crashes ([java_received.json](https://github.com/user-attachments/files/23665540/java_received.json)) the only difference is the timestamp:
image

... so it looks like we're getting duplicate envelopes being sent for some reason
(but only on CI of course) 😕

jamescrosswell:
@supervacuus wondering if you might have some insights. When bumping to .NET 10, our [Android integration tests fail](https://github.com/getsentry/sentry-dotnet/actions/runs/19687973666/job/56398099315?pr=4750) when creating a native crash...
> 2025-11-26T00:27:34.1540247Z Expected a collection with size 1, but got collection with size 16 

Basically we have a single crash, so we're expecting a single envelope... but instead we get loads of envelopes (from sentry.native). We don't see the same behaviour in .net 9 (those tests pass fine)... so maybe a change in behaviour in terms of how .net 10 wraps signals 🤷🏻 although, if so, not one that's documented in [the release notes](https://learn.microsoft.com/en-us/dotnet/core/whats-new/dotnet-10/overview). It's not always 16 envelopes - in previous runs it's been 26 or 28 etc. so basically just "way more than expected".

I can't reproduce this locally - it only happens to us in CI (which makes it pretty tricky to analyse). 

It looks like maybe there are [some tests that are already relevant in the native repo](https://github.com/getsentry/sentry-native/blob/96b3d348c311e806f870239ff26156f990146c9e/tests/test_dotnet_signals.py#L67), so I created a PR to bump these to .NET 10 to see if we see the same issue there... although those are run on Linux Alpine (not Android) so not necessarily expecting the same results:
- https://github.com/getsentry/sentry-native/pull/1457

supervacuus:
Hi @jamescrosswell 👋 

> Basically we have a single crash, so we're expecting a single envelope... but instead we get loads of envelopes (from sentry.native). We don't see the same behaviour in .net 9 (those tests pass fine)... so maybe a change in behaviour in terms of how .net 10 wraps signals 🤷🏻 although, if so, not one that's documented in [the release notes](https://learn.microsoft.com/en-us/dotnet/core/whats-new/dotnet-10/overview). It's not always 16 envelopes - in previous runs it's been 26 or 28 etc. so basically just "way more than expected".

From what I have seen over the last year in the .NET runtime's "documentation" on signal usage, I would be surprised if a change in behavior in that area made its way into the release notes.

I don't have an immediate idea and it is rather hard to follow for me, because i don't know what your test does and which of those outputs are relevant. 

However, intuitively, a variable number of envelopes rather than _one_ looks like a signal loop (typically caused by an `abort()`, because it resets the signal mask before it raises a `SIGABRT`), where some function after serializing the envelope causes a crash in the handler.

But it's really hard to say for me from what I know about your tests. It would be interesting to see _all_ envelopes rather than just their count.

> I can't reproduce this locally - it only happens to us in CI (which makes it pretty tricky to analyse).

But does it repro on CI stably? Would it be possible to `actions/upload-artifact` the entire envelope cache? The JSON printed in the logs only shows a stackrace with a `SIGILL` and a single frame from the `crash()` function `libsentrysupplemental.so`.

> It looks like maybe there are [some tests that are already relevant in the native repo](https://github.com/getsentry/sentry-native/blob/96b3d348c311e806f870239ff26156f990146c9e/tests/test_dotnet_signals.py#L67), so I created a PR to bump these to .NET 10 to see if we see the same issue there... although those are run on Linux Alpine (not Android) so not necessarily expecting the same results:
>
> https://github.com/getsentry/sentry-native/pull/1457

Yes, the .NET tests fail with that bump, but likely due to .NET 10 being an unsupported target (see [my response](https://github.com/getsentry/sentry-native/pull/1457#issuecomment-3580413028)).

I would also ping @jpnurmi on this topic.

jamescrosswell:
> I would also ping @jpnurmi on this topic.

Yes normally. He's unfortunately unavailable until next year 😢 

> I don't have an immediate idea and it is rather hard to follow for me, because i don't know what your test does and which of those outputs are relevant.

That's fair... JP actually wrote the tests. They're supposed to be high level simple integration tests in pester/powershell that compile a .net app and run it, passing in an argument that determines what the app should do. For example, in [the test that's failing here](https://github.com/getsentry/sentry-dotnet/blob/5b7fbc3382c26af9c7618c7227cf02dfe5e8e233/integration-test/android.Tests.ps1#L159-L171), the argument is `"Native"`, which just means the app should [throw a native exception](https://github.com/getsentry/sentry-dotnet/blob/19a8c1aba462f55e829c203b60adf15dbad11d4e/integration-test/net9-maui/App.xaml.cs#L69-L72). That will crash the app, so the test then runs the app (which crashes) then runs the app again (to ensure any crashes dumped to disk get handled) then checks to make sure an envelope containing the expected event was sent/received by the fake sentry server. 

> a variable number of envelopes rather than one looks like a signal loop (typically caused by an abort(), because it resets the signal mask before it raises a SIGABRT), where some function after serializing the envelope causes a crash in the handler.

Hm, that gives me some clues. After each test, the app gets shut down [with this](https://github.com/getsentry/sentry-dotnet/blob/b8d6c02c31359c710b6bfc004e360b42756d7b20/integration-test/net9-maui/App.xaml.cs#L100-L111). That should shut down Sentry before calling `KillProcess`, but I think I see an issue with the implementation for [cascading shutdown](https://github.com/getsentry/sentry-dotnet/blob/fed52443fe8ac0e074d7b0f0be9c815b4176ace0/src/Sentry/Internal/Hub.cs#L871-L881) to other SDKs here... in the case of Android and iOS there is no cascading so the Java SDK and it's embedded Native SDK don't get explicitly shutdown.

I'll dig into that and report back - thanks @supervacuus 🙏🏻

jamescrosswell:
So I tried cascading the SDK shutdown to the Java SDK:
https://github.com/getsentry/sentry-dotnet/blob/278c04c7de03431ca5ff076bee4fe9ccd33042bf/src/Sentry/Internal/Hub.cs#L874

Unfortunately the Java SDK also doesn't cascade this to the instance of the native SDK 😢 
https://github.com/getsentry/sentry-java/blob/fc5ccaf5844ae029d93c566b6fd1b4f5c2e4da83/sentry/src/main/java/io/sentry/Sentry.java#L715-L723

I guess it would be possible to rectify that in the Java SDK but that would require a series of new releases etc. I'll see if I can find an alternative workaround.

jamescrosswell:
@Flash0ver in the latest test runs we're seeing:
> Error: [-] MAUI app (net10.0, Release).Null reference exception (Release) 7.58s (7.57s|4ms)
  Expected string '"type":"SIGSEGV"' to match no elements in collection 

However when I run the same code in a sample application, I don't see the same behaviour, so I'm pretty sure this is as a result of the difficulty in finding a way to shut the app down in a way that does't create a signal loop and lots of other unrelated exceptions that get captured by the native client. 

One possible "proper fix" would be to [cascades the SDK shutdown logic](https://github.com/getsentry/sentry-dotnet/pull/4750#issuecomment-3587531291), but I don't want to delay the version 6 release in order to build that so I think we park this PR for the time being and circle back post v6 release (perhaps in the new year when JP is back on deck).

jpnurmi:
Hey, I have a fix coming up for duplicate Java events:
- https://github.com/getsentry/github-workflows/pull/137

The `SIGILL` loop is still a mystery. It doesn't seem to be related to `Process.KillProcess()` - the same [happens](https://github.com/getsentry/sentry-dotnet/actions/runs/20719184378/job/59478731315) even with `Process.KillProcess()` [removed](https://github.com/getsentry/sentry-dotnet/commit/e5e4ebccf8e2f1d4eac471d1af950ffac318bc25).

jpnurmi:
Good news. Both issues are in the test infrastructure, and fixes are on the way.

The `SIGILL` issue is luckily not a signal loop. Notice how the envelopes are sent at roughly 1s intervals. It is caused by the test app getting automatically relaunched because we trigger crashes a bit too early:
- https://github.com/getsentry/sentry-dotnet/pull/4833

jamescrosswell:
> Good news. Both issues are in the test infrastructure, and fixes are on the way.
> 
> The `SIGILL` issue is luckily not a signal loop. Notice how the envelopes are sent at roughly 1s intervals. It is caused by the test app getting automatically relaunched because we trigger crashes a bit too early:
> 
> * [test(integration): prevent auto-relaunch on Android #4833](https://github.com/getsentry/sentry-dotnet/pull/4833)

Awesome, thanks @jpnurmi - I'm back on deck now as well.

So once those two issues have been resolved, we could reopen this PR?

jpnurmi:
> So once those two issues have been resolved, we could reopen this PR?

Should be ready to go now. I merged the latest main and pushed some minor tweaks to make the Android device tests pass.

integration-test/android.Tests.ps1
```diff
@@ -12,11 +16,18 @@ BeforeDiscovery {
     $script:emulator = Get-AndroidEmulatorId
 }
 
-Describe 'MAUI app (<tfm>, <configuration>)' -ForEach @(
-    @{ tfm = "net9.0-android35.0"; configuration = "Release" }
-    @{ tfm = "net9.0-android35.0"; configuration = "Debug" }
-) -Skip:(-not $script:emulator) {
+$cases = @(
+    @{ Configuration = 'Release' }
+    @{ Configuration = 'Debug'   }
```
jpnurmi: huh, pwsh variables are case-insensitive?
```suggestion
    @{ configuration = 'Release' }
    @{ configuration = 'Debug'   }
```

integration-test/android.Tests.ps1
```diff
@@ -12,11 +16,18 @@ BeforeDiscovery {
...
+)
+Describe 'MAUI app (<dotnet_version>, <configuration>)' -ForEach $cases -Skip:(-not $script:emulator) {
     BeforeAll {
+        if ($dotnet_version -eq 'net9.0') {
+            $tfm = 'net9.0-android35.0'
+        } else {
+            $tfm = 'net10.0-android36.0'
+        }
```
jpnurmi: is `GetAndroidTpv` usable here?

jamescrosswell:
> what about ios integration tests?

That was going to be next yeah... wanted to get Android going first though, since it's typically easier.

integration-test/android.Tests.ps1
jamescrosswell: [Apparently](https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_variables?view=powershell-7.5#long-description) not, although my LLM reckons there are a couple of nuances:
image

test/Sentry.Maui.Device.TestApp/Sentry.Maui.Device.TestApp.csproj
```diff
@@ -32,9 +36,6 @@
 
     <SupportedOSPlatformVersion Condition="$([MSBuild]::GetTargetPlatformIdentifier('$(TargetFramework)')) == 'ios'">13.0</SupportedOSPlatformVersion>
     <SupportedOSPlatformVersion Condition="$([MSBuild]::GetTargetPlatformIdentifier('$(TargetFramework)')) == 'android'">21.0</SupportedOSPlatformVersion>
-    <!-- Pin target iOS version so that our tests don't break when new versions of Xcode are released.
-      'netX.0-ios' resolves the latest version of the iOS SDK otherwise. -->
-    <TargetPlatformVersion Condition="$([MSBuild]::GetTargetPlatformIdentifier('$(TargetFramework)')) == 'ios'">18.0</TargetPlatformVersion>
```
jamescrosswell, line 37: We could leave this in there (bump to 26.2 of course)... that would be consistent with what we've done for the `IntegrationTestApp`

test/Sentry.Maui.Device.TestApp/Sentry.Maui.Device.TestApp.csproj
```diff
@@ -75,6 +76,11 @@
     <PackageReference Include="Microsoft.Maui.Essentials" Version="$(MauiVersion)"/>
   </ItemGroup>
 
+  <ItemGroup Condition="'$(TargetFramework)' == 'net9.0'">
+    <!-- https://github.com/advisories/GHSA-5f2m-466j-3848 -->
+    <PackageReference Include="System.Private.Uri" Version="4.3.2"/>
+  </ItemGroup>
```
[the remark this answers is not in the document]
jamescrosswell: Seems this section is redundant now... when I tweaked the condition per the suggestion above I got:
> error NU1510: Warning As Error: PackageReference System.Private.Uri will not be pruned. Consider removing this package from your dependencies, as it is likely unnecessary.

I've removed it altogether.

integration-test/net9-maui/Sentry.Maui.Device.IntegrationTestApp.csproj
```diff
@@ -1,8 +1,11 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFrameworks>$(TargetFrameworks);net9.0-android35.0</TargetFrameworks>
-    <TargetFrameworks Condition="$([MSBuild]::IsOSPlatform('OSX'))">$(TargetFrameworks);net9.0-ios18.0</TargetFrameworks>
+    <TargetFrameworks>$(TargetFrameworks);net9.0-android35.0;net10.0-android36.0</TargetFrameworks>
+    <!-- Pin target iOS version so that our tests don't break when new versions of Xcode are released.
+         'netX.0-ios' resolves the latest version of the iOS SDK otherwise.
+    -->
+    <TargetFrameworks Condition="$([MSBuild]::IsOSPlatform('OSX'))">$(TargetFrameworks);net10.0-ios26.2</TargetFrameworks>
```
Flash0ver, line 8: follow-up: #4895

integration-test/android.Tests.ps1
```diff
@@ -12,11 +17,14 @@ BeforeDiscovery {
...
 
-Describe 'MAUI app (<tfm>, <configuration>)' -ForEach @(
-    @{ tfm = "net9.0-android35.0"; configuration = "Release" }
-    @{ tfm = "net9.0-android35.0"; configuration = "Debug" }
-) -Skip:(-not $script:emulator) {
+$cases = @(
+    @{ configuration = 'Release' }
+    @{ configuration = 'Debug'   }
+)
+Describe 'MAUI app (<dotnet_version>, <configuration>)' -ForEach $cases -Skip:(-not $script:emulator) {
```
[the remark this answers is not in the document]
Flash0ver: But the `dotnet_version` does not change per test case.
This is a false positive, right?

integration-test/ios.Tests.ps1
```diff
@@ -12,8 +12,11 @@ BeforeDiscovery {
 }
 
 Describe 'iOS app (<tfm>, <configuration>)' -ForEach @(
-    @{ tfm = "net9.0-ios18.0"; configuration = "Release" }
-    @{ tfm = "net9.0-ios18.0"; configuration = "Debug" }
+    # Note: we can't run against net10 and net9 becaus .NET 10 requires Xcode 26.2 and .NET 9 requires Xcode 26.0.
+    # The macOS GitHub Actions runners only have Xcode 26.1+ installed and no support for Xcode 26.2 is planned for
+    # net9.0-ios: https://github.com/dotnet/macios/issues/24199#issuecomment-3819021247
```
Flash0ver, line 17: follow-up: #4895

test/Sentry.Maui.Device.TestApp/Sentry.Maui.Device.TestApp.csproj
```diff
@@ -1,8 +1,12 @@
...
 
   <PropertyGroup>
-    <TargetFrameworks>$(TargetFrameworks);net9.0-android</TargetFrameworks>
-    <TargetFrameworks Condition="$([MSBuild]::IsOSPlatform('OSX'))">$(TargetFrameworks);net9.0-ios</TargetFrameworks>
+    <TargetFrameworks>$(TargetFrameworks);net9.0-android;net10.0-android</TargetFrameworks>
+<!-- Note: we can't run against net9-ios because .NET 10 requires Xcode 26.2 and .NET 9 requires Xcode 26.0.
+     The macOS GitHub Actions runners only have Xcode 26.1+ installed and no support for Xcode 26.2 is planned for
+     net9.0-ios: https://github.com/dotnet/macios/issues/24199#issuecomment-3819021247
+-->
+    <TargetFrameworks Condition="$([MSBuild]::IsOSPlatform('OSX'))">$(TargetFrameworks);net10.0-ios</TargetFrameworks>
```
Flash0ver, line 9: follow-up: #4895

test/Sentry.Maui.Device.TestApp/Sentry.Maui.Device.TestApp.csproj
Flash0ver: follow-up: #4895

Flash0ver (approved):
🤘

integration-test/android.Tests.ps1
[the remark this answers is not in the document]
jamescrosswell: Yeah I think it's working fine... just cursor isn't very good at powershell.

PR #4754, opened 2025-11-19
ref(AI): `Microsoft.Extensions.AI` Agent Monitoring

Flash0ver:
A follow-up to the initial addition of `Sentry.Extensions.AI`.
- #4657

Please see comments for questions and justifications.

### DIFF
--- a/test/Sentry.Extensions.AI.Tests/SentryChatClientTests.cs
+++ b/test/Sentry.Extensions.AI.Tests/SentryChatClientTests.cs
@@ -76,8 +76,8 @@ public class SentryChatClientTests : IDisposable
         var transaction = _fixture.Hub.StartTransaction("test-nonstreaming", "test");
         _fixture.Hub.ConfigureScope(scope => scope.Transaction = transaction);
         // Simulate FunctionInvokingChatClient's Activity
-        SentryAIActivityListener.CreateListener(_fixture.Hub);
-        var FICCActivity = _fixture.Source.StartActivity(SentryAIConstants.FICCActivityNames[0]);
+        var listener = SentryAIActivityListener.CreateListener(_fixture.Hub);
+        var activity = _fixture.Source.StartActivity(SentryAIConstants.FICCActivityNames[0]);
         var sentryChatClient = _fixture.GetSut();
         var chatClientOption = new ChatOptions
         {
@@ -91,7 +91,8 @@ public class SentryChatClientTests : IDisposable
         // Act
         var res = await Assert.ThrowsAsync<InvalidOperationException>(async () =>
             await sentryChatClient.GetResponseAsync([new ChatMessage(ChatRole.User, "hi")], chatClientOption));
-        FICCActivity?.Stop();
+        activity?.Stop();
+        listener.Dispose();
 
         // Assert
         Assert.Equal(expectedException.Message, res.Message);


test/Sentry.Extensions.AI.Tests/SentryAIActivityListenerTests.cs
```diff
@@ -1,5 +1,3 @@
-#nullable enable
-
 namespace Sentry.Extensions.AI.Tests;
 
 public class SentryAIActivityListenerTests
```
Flash0ver, line 3: @alexsohn1126 Is this code of `SentryAIActivityListener` correct?

https://github.com/getsentry/sentry-dotnet/blob/04e389010026e70c29db1b07300bc576adf6d8f9/src/Sentry.Extensions.AI/SentryAIActivityListener.cs#L36-L41

The comment mentions: `Don't pass in OK status in case there was an exception`
But we call `agentSpan?.Finish();`, which sets the SpanStatus to OK if not yet set.

Is there code that sets the SpanStatus to non-OK at another location?
Could you suggest a test that verifies the correct behavior about OK and non-OK?
(feel free to just push to this branch)

src/Sentry.Extensions.AI/Sentry.Extensions.AI.csproj
```diff
@@ -3,7 +3,7 @@
   <PropertyGroup>
     <TargetFrameworks>$(CurrentTfms);netstandard2.0</TargetFrameworks>
     <PackageTags>$(PackageTags);Microsoft.Extensions.AI;AI;LLM</PackageTags>
-    <Description>Microsoft.Extensions.AI integration for Sentry - captures AI Agent telemetry spans per Sentry AI Agents module.</Description>
+    <Description>Official Microsoft.Extensions.AI integration for Sentry - Open-source error tracking that helps developers monitor and fix crashes in real time.</Description>
```
Flash0ver, line 6: note: to be in sync with other packages ... e.g. `Sentry.Extensions.Logging`
https://github.com/getsentry/sentry-dotnet/blob/04e389010026e70c29db1b07300bc576adf6d8f9/src/Sentry.Extensions.Logging/Sentry.Extensions.Logging.csproj#L6

src/Sentry.Extensions.AI/SentryChatClient.cs
```diff
@@ -5,19 +5,20 @@ namespace Sentry.Extensions.AI;
 internal sealed class SentryChatClient : DelegatingChatClient
 {
     private readonly ActivitySource _activitySource;
-    private readonly HubAdapter _hub = HubAdapter.Instance;
+    private readonly IHub _hub;
```
Flash0ver, line 8: note: make Hub injectable, so that the Tests don't need to rely on the static `SentrySdk`/`HubAdapter`, in order to have the Tests isolated ... so that they could both run in parallel and are independent of the run-order of other tests ... while still using `HubAdapter.Instance` when in production

test/Sentry.Extensions.AI.Tests/Sentry.Extensions.AI.Tests.csproj
```diff
@@ -2,6 +2,11 @@
 
   <PropertyGroup>
     <TargetFrameworks>$(CurrentTfms)</TargetFrameworks>
+    <TargetFrameworks Condition="$([MSBuild]::IsOSPlatform('Windows'))">$(TargetFrameworks);net48</TargetFrameworks>
```
Flash0ver, line 5: note: adding .NET Framework 4.8, so that we also test the .NET Standard 2.0 TFM

test/Sentry.Extensions.AI.Tests/Sentry.Extensions.AI.Tests.csproj
```diff
@@ -2,6 +2,11 @@
...
+  </PropertyGroup>
+
+  <PropertyGroup>
+    <Nullable>enable</Nullable>
```
Flash0ver, line 9: note: enable `Nullable` on a project level, so that we don't have to enable it individually per `.cs` document

test/Sentry.Extensions.AI.Tests/SentryChatClientTests.cs
```diff
@@ -151,25 +151,28 @@ public async Task CompleteStreamingAsync_HandlesErrors_AndFinishesSpanWithExcept
...
             .Returns(CreateFailingStreamingUpdatesAsync(expectedException));
         var sentryChatClient = _fixture.GetSut();
+        var results = new List<ChatResponseUpdate>();
 
         // Act
         var actualException = await Assert.ThrowsAsync<InvalidOperationException>(async () =>
         {
             await foreach (var update in sentryChatClient.GetStreamingResponseAsync([new ChatMessage(ChatRole.User, "hi")]))
             {
-                // Should not reach here due to exception
```
Flash0ver, line 167: note: it actually does reach here ... but only once ... moving to the 2nd item in the sequence will throw ... so that this Task will not complete successfully

test/Sentry.Extensions.AI.Tests/SentryInstrumentedFunctionTests.cs
```diff
@@ -31,14 +30,8 @@ public async Task InvokeCoreAsync_WithValidFunction_ReturnsResult()
         // Assert
         // AIFunctionFactory returns JsonElement, so we need to check the actual content
         Assert.NotNull(result);
-        if (result is JsonElement jsonElement)
-        {
-            Assert.Equal("test result", jsonElement.GetString());
-        }
-        else
-        {
-            Assert.Equal("test result", result);
```
Flash0ver, line 40: note: removing the "defensive" paths of the test ... to be more strict and have less nesting

test/Sentry.Extensions.AI.Tests/SentryAIExtensionsTests.cs
```diff
@@ -1,4 +1,3 @@
-#nullable enable
 using Microsoft.Extensions.AI;
 
 namespace Sentry.Extensions.AI.Tests;
```
Flash0ver, line 3: @alexsohn1126 Is this code in the _Console-Sample_ required?

https://github.com/getsentry/sentry-dotnet/blob/04e389010026e70c29db1b07300bc576adf6d8f9/samples/Sentry.Samples.ME.AI.Console/Program.cs#L105-L106

The Sentry-SDK should ensure that all envelopes from the Background-Worker's queue are flushed/sent before the process exits/terminates.

I wonder if we are in need of a Flush in our SDK / Integration codebase 🤔

samples/Sentry.Samples.ME.AI.Console/Program.cs
```diff
@@ -96,7 +96,7 @@
     "Please help me with the following tasks: 1) Find Alice's age, 2) Get weather in New York, and 3) Calculate a complex result for number 15. Please use the appropriate tools for each task.",
     options);
 
-logger.LogInformation("Response: {ResponseText}", response.Messages?.FirstOrDefault()?.Text ?? "No response");
+logger.LogInformation("Response: {ResponseText}", response.Messages.FirstOrDefault()?.Text ?? "No response");
```
Flash0ver, line 99: note: the null-conditional access on `Messages` should not be necessary:

```CSharp
// class Microsoft.Extensions.AI.ChatResponse
[AllowNull]
public IList<ChatMessage> Messages { get; set; }
```

Where the "ReadState" is _NotNull_ (from declared type),
and the "WriteState" is _Nullable_ (from `[AllowNull]`).

test/Sentry.Extensions.AI.Tests/SentryAIActivityListenerTests.cs
alexsohn1126: There is a case where we would finish the agent span with an exception. 

https://github.com/getsentry/sentry-dotnet/blob/04e389010026e70c29db1b07300bc576adf6d8f9/src/Sentry.Extensions.AI/SentryChatClient.cs#L116-L125

`ActivityStopped` runs regardless whether the activity was stopped due to a normal reason or an exception. Therefore we don't want to override the span's status with an `Ok` status when it was set to `InternalError` status from `AfterResponseCleanup`.

This behaviour should be captured under this test:
https://github.com/getsentry/sentry-dotnet/blob/04e389010026e70c29db1b07300bc576adf6d8f9/test/Sentry.Extensions.AI.Tests/SentryChatClientTests.cs#L71

We check whether the agent span has the correct status here:
https://github.com/getsentry/sentry-dotnet/blob/04e389010026e70c29db1b07300bc576adf6d8f9/test/Sentry.Extensions.AI.Tests/SentryChatClientTests.cs#L101

test/Sentry.Extensions.AI.Tests/SentryAIExtensionsTests.cs
alexsohn1126: Yea, I'm not quite sure how that got there either 😓 We can delete that

test/Sentry.Extensions.AI.Tests/SentryAIActivityListenerTests.cs
Flash0ver: got it ... thanks
Please add a test verifying this behavior.

alexsohn1126: Test created: https://github.com/getsentry/sentry-dotnet/pull/4754/commits/09c0af0f8c8b8439ae41741ceb6d1b4c9fde2a73

src/Sentry.Extensions.AI/Sentry.Extensions.AI.csproj
bruno-garcia: Why keep in sync? This package is about monitoring AI so the other description was more accurate.

The current text I wrote in 2018 when Sentry was just error monitoring. "Aligning" this text doesn't add any value. If anything we could have better description for each package.

That said, I wouldn't bother changing just surprised this ended up being changed in this PR

PR #4760, opened 2025-11-21
Move setting /NODEFAULTLIB:MSVCRT LinkerArg to windows-specific targets in Sentry.Native.targets

akoeplinger:
Before the flag was added even if you built for an android RID on windows, causing an error.


jamescrosswell (commented):
Awesome - thank you @akoeplinger ❤️ 

The only thing missing is a change log... which typically describes the impact for SDK users. If there is no impact, it can be skipped but in this case I think it would resolve some build issues for folks building on Windows, so worth adding.

akoeplinger:
@jamescrosswell done

Flash0ver:
@sentry review

jamescrosswell (approved):
Fantastic - thank you @akoeplinger!

PR #4762, opened 2025-11-22
SentryGraphQLHttpFailedRequestHandler improved issue grouping.

### LINKED ISSUE #4759: Fix issue grouping on GraphQL requests
### Description

We should apply the same fix to GraphQL that was applied in:
- #4724

ericsampson:
Fixes #4759
- #4759

---

Applies the same improvements for GraphQL as in this PR for plain HTTP: https://github.com/getsentry/sentry-dotnet/pull/4724

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
+
+### Fixes
+
+- Captured [GraphQL Client Errors](https://docs.sentry.io/platforms/dotnet/configuration/graphql-client-errors/) on .NET 5+ now include a full stack trace in order to improve Issue grouping ([#4762](https://github.com/getsentry/sentry-dotnet/pull/4762))
+
 ## 6.0.0-rc.2
 
 ### Features
@@ -9,7 +15,6 @@
 ### Fixes
 
 - Captured [Http Client Errors](https://docs.sentry.io/platforms/dotnet/guides/aspnet/configuration/http-client-errors/) on .NET 5+ now include a full stack trace in order to improve Issue grouping ([#4724](https://github.com/getsentry/sentry-dotnet/pull/4724))
-- Captured [GraphQL Client Errors](https://docs.sentry.io/platforms/dotnet/configuration/graphql-client-errors/) on .NET 5+ now include a full stack trace in order to improve Issue grouping ([#4762](https://github.com/getsentry/sentry-dotnet/pull/4762))
 - Sentry Tracing middleware crashed ASP.NET Core in .NET 10 in 6.0.0-rc.1 and earlier ([#4747](https://github.com/getsentry/sentry-dotnet/pull/4747))
 - Avoid appending `/NODEFAULTLIB:MSVCRT` to NativeAOT linker arguments on Windows when targetting non-Windows platforms (Android, Browser) ([#4760](https://github.com/getsentry/sentry-dotnet/pull/4760))
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Fixes
 
 - Captured [Http Client Errors](https://docs.sentry.io/platforms/dotnet/guides/aspnet/configuration/http-client-errors/) on .NET 5+ now include a full stack trace in order to improve Issue grouping ([#4724](https://github.com/getsentry/sentry-dotnet/pull/4724))
+- Captured [GraphQL Client Errors](https://docs.sentry.io/platforms/dotnet/configuration/graphql-client-errors/) on .NET 5+ now include a full stack trace in order to improve Issue grouping ([#4762](https://github.com/getsentry/sentry-dotnet/pull/4762))
 - Sentry Tracing middleware crashed ASP.NET Core in .NET 10 in 6.0.0-rc.1 and earlier ([#4747](https://github.com/getsentry/sentry-dotnet/pull/4747))
 
 ### Dependencies
--- a/src/Sentry/SentryGraphQLHttpFailedRequestHandler.cs
+++ b/src/Sentry/SentryGraphQLHttpFailedRequestHandler.cs
@@ -36,13 +36,13 @@ internal class SentryGraphQLHttpFailedRequestHandler : SentryFailedRequestHandle
                     var exception = new GraphQLHttpRequestException(errorMessage);
 
 #if NET5_0_OR_GREATER
-                    // Add a full stack trace into the exception to improve Issue grouping,
-                    // see https://github.com/getsentry/sentry-dotnet/issues/3582
-                    ExceptionDispatchInfo.SetCurrentStackTrace(exception);
+    // Add a full stack trace into the exception to improve Issue grouping,
+    // see https://github.com/getsentry/sentry-dotnet/issues/3582
+    ExceptionDispatchInfo.Throw(ExceptionDispatchInfo.SetCurrentStackTrace(exception));
+#else
+                    // Where SetCurrentStackTrace is not available, just throw the Exception
+                    throw exception;
 #endif
-                    // Wrap in a new exception to preserve the stack trace when thrown and caught.
-                    // The inner exception will have the full stack trace for better issue grouping.
-                    throw new GraphQLHttpRequestException(errorMessage, exception);
                 }
             }
             // No GraphQL errors, but we still might have an HTTP error status
--- a/src/Sentry/SentryGraphQLHttpFailedRequestHandler.cs
+++ b/src/Sentry/SentryGraphQLHttpFailedRequestHandler.cs
@@ -36,9 +36,9 @@ internal class SentryGraphQLHttpFailedRequestHandler : SentryFailedRequestHandle
                     var exception = new GraphQLHttpRequestException(errorMessage);
 
 #if NET5_0_OR_GREATER
-    // Add a full stack trace into the exception to improve Issue grouping,
-    // see https://github.com/getsentry/sentry-dotnet/issues/3582
-    ExceptionDispatchInfo.Throw(ExceptionDispatchInfo.SetCurrentStackTrace(exception));
+                    // Add a full stack trace into the exception to improve Issue grouping,
+                    // see https://github.com/getsentry/sentry-dotnet/issues/3582
+                    ExceptionDispatchInfo.Throw(ExceptionDispatchInfo.SetCurrentStackTrace(exception));
 #else
                     // Where SetCurrentStackTrace is not available, just throw the Exception
                     throw exception;
--- a/src/Sentry/SentryGraphQLHttpFailedRequestHandler.cs
+++ b/src/Sentry/SentryGraphQLHttpFailedRequestHandler.cs
@@ -1,3 +1,4 @@
+using System.Runtime.ExceptionServices;
 using Sentry.Internal;
 using Sentry.Internal.Extensions;
 using Sentry.Protocol;
@@ -24,7 +25,7 @@ internal class SentryGraphQLHttpFailedRequestHandler : SentryFailedRequestHandle
         JsonElement? json = null;
         try
         {
-            json = GraphQLContentExtractor.ExtractResponseContentAsync(response, _options).Result;
+            json = GraphQLContentExtractor.ExtractResponseContentAsync(response, _options).GetAwaiter().GetResult();
             if (json is { } jsonElement)
             {
                 if (jsonElement.TryGetProperty("errors", out var errorsElement))
@@ -32,7 +33,16 @@ internal class SentryGraphQLHttpFailedRequestHandler : SentryFailedRequestHandle
                     // We just show the first error... maybe there's a better way to do this when multiple errors exist.
                     // We should check what the Java code is doing.
                     var errorMessage = errorsElement[0].GetProperty("message").GetString() ?? "GraphQL Error";
-                    throw new GraphQLHttpRequestException(errorMessage);
+                    var exception = new GraphQLHttpRequestException(errorMessage);
+
+#if NET5_0_OR_GREATER
+                    // Add a full stack trace into the exception to improve Issue grouping,
+                    // see https://github.com/getsentry/sentry-dotnet/issues/3582
+                    ExceptionDispatchInfo.SetCurrentStackTrace(exception);
+#endif
+                    // Wrap in a new exception to preserve the stack trace when thrown and caught.
+                    // The inner exception will have the full stack trace for better issue grouping.
+                    throw new GraphQLHttpRequestException(errorMessage, exception);
                 }
             }
             // No GraphQL errors, but we still might have an HTTP error status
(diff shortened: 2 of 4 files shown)


src/Sentry/SentryGraphQLHttpFailedRequestHandler.cs
```diff
@@ -24,15 +25,24 @@ protected internal override void DoEnsureSuccessfulResponse([NotNull] HttpReques
...
+                    var exception = new GraphQLHttpRequestException(errorMessage);
+
+#if NET5_0_OR_GREATER
+                    // Add a full stack trace into the exception to improve Issue grouping,
+                    // see https://github.com/getsentry/sentry-dotnet/issues/3582
+                    ExceptionDispatchInfo.SetCurrentStackTrace(exception);
+#endif
+                    // Wrap in a new exception to preserve the stack trace when thrown and caught.
+                    // The inner exception will have the full stack trace for better issue grouping.
+                    throw new GraphQLHttpRequestException(errorMessage, exception);
```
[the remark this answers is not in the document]
ericsampson: I think that's fine.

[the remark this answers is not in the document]
jamescrosswell: There's something weird about wrapping an exception that we just created (on line 36) in another identical exception. 

Could we do this the same way as the other handler:
https://github.com/getsentry/sentry-dotnet/blob/ab2d7c45a396e8575423fc4299a9519f9136f907/src/Sentry/SentryHttpFailedRequestHandler.cs#L40-L54

src/Sentry/SentryGraphQLHttpFailedRequestHandler.cs
```diff
@@ -24,15 +25,24 @@ protected internal override void DoEnsureSuccessfulResponse([NotNull] HttpReques
         JsonElement? json = null;
         try
         {
-            json = GraphQLContentExtractor.ExtractResponseContentAsync(response, _options).Result;
+            json = GraphQLContentExtractor.ExtractResponseContentAsync(response, _options).GetAwaiter().GetResult();
```
Flash0ver, line 28: praise: An additional improvement ... changing from a potential `System.AggregateException` via `Task<TResult>.Result`, to the actual `Exception` that transitioned the `Task` to the `Faulted` status. Well done!

src/Sentry/SentryGraphQLHttpFailedRequestHandler.cs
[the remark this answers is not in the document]
Flash0ver: Would something like this work?

```CSharp
#if NET5_0_OR_GREATER
    // Add a full stack trace into the exception to improve Issue grouping,
    // see https://github.com/getsentry/sentry-dotnet/issues/3582
    ExceptionDispatchInfo.Throw(ExceptionDispatchInfo.SetCurrentStackTrace(exception));
#else
    // Where SetRemoteStackTrace is not available, just throw the Exception
    throw exception;
#endif
```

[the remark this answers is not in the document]
ericsampson: I think so, I'll give it a shot.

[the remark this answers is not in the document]
ericsampson: it worked fine and is cleaner thanks, updated the PR.

test/Sentry.Tests/SentryGraphQlTestHelpers.cs
```diff
@@ -46,17 +46,20 @@ public static StringContent ResponesContent(string responseText) => JsonContent(
...
     /// e.g.
-    /// "[{"message":"Query does not contain operation \u0027getAllNotes\u0027.","extensions":{"code":"INVALID_OPERATION","codes":["INVALID_OPERATION"]}}]"
+    /// "{"errors": [{"message":"Query does not contain operation \u0027getAllNotes\u0027.","extensions":{"code":"INVALID_OPERATION","codes":["INVALID_OPERATION"]}}]}"
     /// </summary>
     public static StringContent ErrorContent(string errorMessage, string errorCode) => JsonContent(
-        new dynamic[]
+        new
         {
-            new
+            errors = new dynamic[]
```
ericsampson, line 54: FYI one or two of the original tests weren't testing what they though they were because this `ErrorContent` helper wasn't generating a valid GraphQL response schema due to the missing top-level `errors` key, so that the handler wasn't getting into this code path as it should have been: `if (jsonElement.TryGetProperty("errors", out var errorsElement))`

src/Sentry/SentryGraphQLHttpFailedRequestHandler.cs
```diff
@@ -24,15 +25,24 @@ protected internal override void DoEnsureSuccessfulResponse([NotNull] HttpReques
...
                     // We should check what the Java code is doing.
                     var errorMessage = errorsElement[0].GetProperty("message").GetString() ?? "GraphQL Error";
-                    throw new GraphQLHttpRequestException(errorMessage);
+                    var exception = new GraphQLHttpRequestException(errorMessage);
+
+#if NET5_0_OR_GREATER
+    // Add a full stack trace into the exception to improve Issue grouping,
+    // see https://github.com/getsentry/sentry-dotnet/issues/3582
+    ExceptionDispatchInfo.Throw(ExceptionDispatchInfo.SetCurrentStackTrace(exception));
+#else
```
Flash0ver, line 42: suggestion: align format

```suggestion
#if NET5_0_OR_GREATER
                    // Add a full stack trace into the exception to improve Issue grouping,
                    // see https://github.com/getsentry/sentry-dotnet/issues/3582
                    ExceptionDispatchInfo.Throw(ExceptionDispatchInfo.SetCurrentStackTrace(exception));
#else
```

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@
 ### Fixes
 
 - Captured [Http Client Errors](https://docs.sentry.io/platforms/dotnet/guides/aspnet/configuration/http-client-errors/) on .NET 5+ now include a full stack trace in order to improve Issue grouping ([#4724](https://github.com/getsentry/sentry-dotnet/pull/4724))
+- Captured [GraphQL Client Errors](https://docs.sentry.io/platforms/dotnet/configuration/graphql-client-errors/) on .NET 5+ now include a full stack trace in order to improve Issue grouping ([#4762](https://github.com/getsentry/sentry-dotnet/pull/4762))
```
Flash0ver: note: I'm afraid ... as usual (😟) ... please update from `main` and move into the "new" `Unreleased > Fixes` section, now after [6.0.0-rc.2](https://github.com/getsentry/sentry-dotnet/releases/tag/6.0.0-rc.2) release

test/Sentry.Tests/SentryGraphQlTestHelpers.cs
Flash0ver: praise: great catch

Flash0ver (commented):
again: thank you so much for the improvement ... and well done on the testing part!

some small changes needed
- align code indentation, according to our `dotnet format` rules and workflow
- merge from `main` to update the `CHANGELOG`

other than that it's ✅ from my end

Flash0ver:
@sentry review

src/Sentry/SentryGraphQLHttpFailedRequestHandler.cs
```diff
@@ -24,15 +25,24 @@ protected internal override void DoEnsureSuccessfulResponse([NotNull] HttpReques
...
+                    // Where SetCurrentStackTrace is not available, just throw the Exception
+                    throw exception;
+#endif
```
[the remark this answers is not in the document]
Flash0ver: hmmm ... correct ... but

in `SentryHttpFailedRequestHandler` we essentially only read the Status-Code

here, in `SentryGraphQLHttpFailedRequestHandler`, we also try to extract JSON content from the `HttpResponseMessage` ... I'm a bit uncertain whether this will always succeed ... should the JSON Content from the Response be malformed then `GraphQLContentExtractor.ExtractResponseContentAsync` would throw a `JsonReaderException`

I guess this would be another issue/changeset to refactor this approach,
as _this_ PR is not changing anything about the throw-behavior of _this_ method,
but instead replaced the Stack-Trace of the Exception with a richer Stack-Trace for more detailed grouping in Sentry.
Technically, I do agree that this would be nice to have, to make the `NET5_0_OR_GREATER` path "throw-free".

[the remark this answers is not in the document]
ericsampson: For context I wanted to keep the changes targeted and minimal vs the current strategy (at least for a first PR), I figured it's safer to have the outer try as in the original to capture json extraction failures--that seems like "not an edge case".

src/Sentry/SentryGraphQLHttpFailedRequestHandler.cs
ericsampson: @Flash0ver sorry for the formatting issues, I'm having some trouble getting the formatter to run locally.

CHANGELOG.md
ericsampson: ah blah, I didn't realize there was an rc2 cut. OK :)

src/Sentry/SentryGraphQLHttpFailedRequestHandler.cs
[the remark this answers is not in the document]
jamescrosswell: Seems unrelated to this PR yeah... I'm not sure I see a problem. If the JSON is malformed, that would be caught and converted to a sentry event [here](https://github.com/getsentry/sentry-dotnet/blob/b74ff2995da36e1cfb6e4d6b1d4d7d6ed05194b2/src/Sentry/SentryGraphQLHttpFailedRequestHandler.cs#L55), which I don't think is unexpected behaviour. Is there a better alternative?

src/Sentry/SentryGraphQLHttpFailedRequestHandler.cs
```diff
@@ -24,15 +25,24 @@ protected internal override void DoEnsureSuccessfulResponse([NotNull] HttpReques
...
                     // We just show the first error... maybe there's a better way to do this when multiple errors exist.
                     // We should check what the Java code is doing.
                     var errorMessage = errorsElement[0].GetProperty("message").GetString() ?? "GraphQL Error";
-                    throw new GraphQLHttpRequestException(errorMessage);
+                    var exception = new GraphQLHttpRequestException(errorMessage);
+
+#if NET5_0_OR_GREATER
+                    // Add a full stack trace into the exception to improve Issue grouping,
+                    // see https://github.com/getsentry/sentry-dotnet/issues/3582
+                    ExceptionDispatchInfo.Throw(ExceptionDispatchInfo.SetCurrentStackTrace(exception));
```
jamescrosswell, line 41: So I think I'm getting confused now. 

How is this:
```
var exception = new GraphQLHttpRequestException(errorMessage);
ExceptionDispatchInfo.Throw(ExceptionDispatchInfo.SetCurrentStackTrace(exception));
```

... different from this?
```
var exception = new GraphQLHttpRequestException(errorMessage);
throw exception;
```

I get that the behaviour for `ExceptionDispatchInfo.Throw` is different to `throw` when we're talking about exceptions that have been thrown elsewhere in the code and have a stack trace to preserve... in this case we're creating the exception and throwing it on the very next line though... so what is the value of `ExceptionDispatchInfo.Throw` vs a plain old `throw` in this scenario? Aren't the stack traces going to be almost identical?

ericsampson: It is different, I know it's not obvious--try it via the unit test and debugging, if you're interested. The one test should fail if you just throw

Flash0ver: indeed

via `Sentry.Samples.GraphQL.Server` and `Sentry.Samples.GraphQL.Client.Http`

```diff
    at Sentry.SentryGraphQLHttpFailedRequestHandler.DoEnsureSuccessfulResponse(HttpRequestMessage request, HttpResponseMessage response) in /../sentry-dotnet/src/Sentry/SentryGraphQLHttpFailedRequestHandler.cs:line 40
+   at Sentry.SentryFailedRequestHandler.HandleResponse(HttpResponseMessage response) in /../sentry-dotnet/src/Sentry/SentryFailedRequestHandler.cs:line 47
+   at Sentry.SentryGraphQLHttpMessageHandler.HandleResponse(HttpResponseMessage response, ISpan span, String method, String url) in /../sentry-dotnet/src/Sentry/SentryGraphQLHttpMessageHandler.cs:line 95
+   at Sentry.SentryMessageHandler.SendAsync(HttpRequestMessage request, CancellationToken cancellationToken) in /../sentry-dotnet/src/Sentry/SentryMessageHandler.cs:line 91
+   at System.Runtime.CompilerServices.AsyncTaskMethodBuilder`1.AsyncStateMachineBox`1.ExecutionContextCallback(Object s)
+   at System.Threading.ExecutionContext.RunInternal(ExecutionContext executionContext, ContextCallback callback, Object state)
+   at ...
+   at System.Threading.ThreadPoolWorkQueue.Dispatch()
+   at System.Threading.PortableThreadPool.WorkerThread.WorkerThreadStart()
+   at System.Threading.Thread.StartCallback()
+--- End of stack trace from previous location ---
+   at Sentry.SentryGraphQLHttpFailedRequestHandler.DoEnsureSuccessfulResponse(HttpRequestMessage request, HttpResponseMessage response) in /../sentry-dotnet/src/Sentry/SentryGraphQLHttpFailedRequestHandler.cs:line 40
```

Flash0ver: however ....

there isn't really a difference in user frames ...
so in _Sentry_, I don't really see a difference in Grouping / Fingerprinting

🤔

Flash0ver: @ericsampson, could you give [6.0.0-rc.2](https://github.com/getsentry/sentry-dotnet/releases/tag/6.0.0-rc.2) a try to confirm that #4724 indeed improves your scenario?

ericsampson: @Flash0ver sent you and James an email just now with some info, it doesn't look like it's working how I expected/wanted 😭

jamescrosswell: That's OK - this problem space turns us all on our heads at some point. 

Thanks heaps for looking into this @ericsampson and should we close this PR then?

EnesEfeTokta (commented):
I think it looks okay. Good jobb. 💡

src/Sentry/SentryGraphQLHttpFailedRequestHandler.cs
ericsampson: I'd be interested in at least figuring out why it's turning out this way, for the HTTP Client case at least. It feels to me like maybe the Sentry server-side fingerprinting or something is transforming things in a way that is unexpected to me...

Flash0ver:
We'll probably not move forward with the functional change, but will cherry-pick the improvements to the tests, as well as changing from `.Result` to `.GetAwaiter().GetResult()`.

ericsampson:
Sounds good. If I ever have some free time at work, I can investigate why it wasn't working as I'd expect re the Sentry side processing, and can revive if desired. Cheers

PR #4700, opened 2025-11-06
fix(logs): minimum Log-Level for Structured Logs

Flash0ver:
Fixes #4539
Fixes #4734
- #4539
- #4734

---

### Summary
Previously, the Logging-Configuration was inconsistent between SDKs (**M.E.L**, **ASP.NET Core**, **MAUI**, **Google-Cloud-Functions**):
- **M.E.L** correctly ignores Logging-Configuration for _Breadcrumbs_ and _Events_
- **ASP.NET Core**, **MAUI** and **Google-Cloud-Functions** were incorrectly considering Logging-Configuration for _Breadcrumbs_ and _Events_
- all (**M.E.L**, **ASP.NET Core** and **MAUI**) were incorrectly ignoring Logging-Configuration for _Structured Logs_
- **Google-Cloud-Functions** did not support _Structured Logs_

This changeset is normalizing the applied Logging-Filter-Rules across all integrations:
Logging-Configuration (i.e. `Logging:LogLevel:Default` and `Logging:Sentry:LogLevel:Default` via e.g. `appsettings.json`) is now respected for _Structured Logs_, but now ignored for _Breadcrumbs_ and _Events_, since we have existing `MinimumBreadcrumbLevel` and `MinimumEventLevel`, as well as `AddLogEntryFilter` extension methods for filtering _Breadcrumbs_ and _Events_.

### New
- `Sentry.Google.Cloud.Functions` now supports _Structured Logs_
  - when enabled via `EnableLogs = true`

### Changed
- remove `SentryLoggingOptions.ExperimentalLogging.MinimumLogLevel` for Structured-Logs
  - instead, allow configuration of Structured-Logs via the `"Sentry"` logging provider name
- the minimum Log-Level Configuration (`appsettings.json`) now only applies to _Structured Logging_
- the minimum Log-Level Configuration (`appsettings.json`) no longer applies to _Breadcrumbs_
  - instead, configure via the already existing `SentryLoggingOptions.MinimumBreadcrumbLevel` and `SentryLoggingOptionsExtensions.AddLogEntryFilter`
- the minimum Log-Level Configuration (`appsettings.json`) no longer applies to _Events_
  - instead, configure via the already existing `SentryLoggingOptions.MinimumEventLevel` and `SentryLoggingOptionsExtensions.AddLogEntryFilter`
- made all types that implement `ILoggerProvider` and `IConfigureOptions` `internal`
  - in order to ensure that `LoggerFilterRule`s are correctly set
[...]

PR #4768, opened 2025-11-25
test: remove obsolete verified file

Flash0ver:
During #4700, I noticed a leftover `*.verified.txt` file.

`Sentry.AspNetCore.Tests` is targeting `net8.0;net9.0;net10.0`.
It does not target `net48`.


Flash0ver:
@sentry review

PR #4774, opened 2025-11-27
meta: update Bug Report issue template

Flash0ver:
Updating the .NET and IDE versions of the Bug Report issue template.

Please see my questions as PR comments

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -85,9 +85,9 @@ body:
         - Visual Studio Code (Windows)
         - Visual Studio Code (macOS)
         - Visual Studio Code (Linux)
-        - Rider 2025 (Windows)
-        - Rider 2025 (macOS)
-        - Rider 2025 (Linux)
+        - Rider 2025.x (Windows)
+        - Rider 2025.x (macOS)
+        - Rider 2025.x (Linux)
         - Other
     validations:
       required: true


.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -81,11 +81,13 @@ body:
       label: Development Environment
       description: What development environment are you using
       options:
-        - Visual Studio v17.x
+        - Visual Studio v18.x
```
Flash0ver, line 84: question: do we want to keep `v17` ... or do we "support" the latest version only?

Also, technically speaking, `v16.11` is still supported until April 2029.

.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -81,11 +81,13 @@ body:
...
         - Visual Studio Code (Windows)
-        - Visual Studio Code (MacOS)
-        - Rider 2024 (Windows)
-        - Rider 2024 (MacOS)
+        - Visual Studio Code (macOS)
+        - Visual Studio Code (Linux)
```
Flash0ver, line 87: question: Do we want to add `Linux` ... or stick with `Windows` and `macOS`

.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -81,11 +81,13 @@ body:
...
+        - Rider 2025 (Windows)
+        - Rider 2025 (macOS)
+        - Rider 2025 (Linux)
```
Flash0ver: question: Rider version

Should we be more precise by stating the latest version `2025.3`?
But then we have to update this 3 times per year.
Or is updating it yearly fine?

Should we say `2025.x` instead?

.github/ISSUE_TEMPLATE/bug_report.yml
jamescrosswell: I think in the template we give people the option to describe their situation as accurately as possible. Whether we support or not is something we can handle in our replies.

.github/ISSUE_TEMPLATE/bug_report.yml
jamescrosswell: Sure, I think it's helpful to know if someone is running Linux (especially for build/tooling related issues)

.github/ISSUE_TEMPLATE/bug_report.yml
jamescrosswell: I like the 2025.x idea... I'm not sure how often the specific version of Rider/VS is relevant and we can just ask in those cases.

PR #4814, opened 2025-12-16
fixed: Thread safety in AssemblyStoreReader when sending events on Android

### LINKED ISSUE #4799: Running several  SentrySdk.CaptureMessage in parallel freezes the MAUI Android app
### Package

Sentry

### .NET Flavor

.NET

### .NET Version

9.0.203

### OS

Android

### OS Version

Android 15

### Development Environment

Visual Studio v18.x

### SDK Version

5.16.2

_No response_

### Workload Versions

android                    35.0.39/9.0.100        SDK 9.0.200, VS 17.13.35931.197
aspire                     8.2.2/8.0.100          SDK 9.0.200, VS 17.13.35931.197
ios                        18.2.9173/9.0.100      SDK 9.0.200, VS 17.13.35931.197
maccatalyst                18.2.9173/9.0.100      SDK 9.0.200, VS 17.13.35931.197
maui-windows               9.0.14/9.0.100         SDK 9.0.200, VS 17.13.35931.197

`.UseSentry("")`

### Steps to Reproduce

1. Run SentrySdk.CaptureMessage("Hello") in parallel
2.  Use this somewhere in Main thread, like on a button press
```
 Parallel.Invoke(
     () => SentrySdk.CaptureMessage("Hello"),
     () => SentrySdk.CaptureMessage("Hello"),
     () => SentrySdk.CaptureMessage("Hello"),
     () => SentrySdk.CaptureMessage("Hello"),
     () => SentrySdk.CaptureMessage("Hello")
	);
```

### Expected Result

5 Messages will be reported to Sentry

### Actual Result

App freezes and stops responding. This seems like a regression as at least the Sentry.Maui 5.4.0 version works fine with the same code. Please reach out if a sample is needed

jamescrosswell:
Fixes #4799:
- #4799

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@
 
 ### Fixes
 
-- Fixed threadsafety issue on Android when multiple events are captured concurrently ([#4814](https://github.com/getsentry/sentry-dotnet/pull/4814))
+- Fixed thread-safety issue on Android when multiple events are captured concurrently ([#4814](https://github.com/getsentry/sentry-dotnet/pull/4814))
 
 ### Dependencies
 
--- a/src/Sentry.Android.AssemblyReader/V2/AssemblyStoreReader.cs
+++ b/src/Sentry.Android.AssemblyReader/V2/AssemblyStoreReader.cs
@@ -11,7 +11,7 @@ internal abstract class AssemblyStoreReader
 
     private static readonly UTF8Encoding ReaderEncoding = new UTF8Encoding(false);
 
-    internal object StreamLock { get; } = new();
+    internal Lock StreamLock { get; } = new();
     protected Stream StoreStream { get; }
 
     public abstract string Description { get; }
--- a/test/Sentry.Android.AssemblyReader.Tests/StoreReaderTests.cs
+++ b/test/Sentry.Android.AssemblyReader.Tests/StoreReaderTests.cs
@@ -9,13 +9,14 @@ public class StoreReaderTests
     {
         // Arrange
         var buffer = new byte[1024 * 1024];
-        var memoryStream = new MemoryStream(buffer);
+        using var memoryStream = new MemoryStream(buffer);
         var storeReader = new StoreReader(memoryStream, "testStore", null);
 
         // Act
-        Parallel.For(0, 10, _ => storeReader.IsSupported());
+        var result = Parallel.For(0, 10, _ => storeReader.IsSupported());
 
-        // No Assert - test passes if no exceptions are thrown
+        // Test passes if no exceptions are thrown, but we can also assert completion
+        result.IsCompleted.Should().BeTrue();
     }
 
     [Fact]
@@ -23,7 +24,7 @@ public class StoreReaderTests
     {
         // Arrange
         var buffer = new byte[1024 * 1024];
-        var memoryStream = new MemoryStream(buffer);
+        using var memoryStream = new MemoryStream(buffer);
         var storeReader = new StoreReader(memoryStream, "testStore", null);
         var entry = new StoreReader.StoreItemV2(
             AndroidTargetArch.Arm64,
@@ -43,8 +44,9 @@ public class StoreReaderTests
             ignore: false);
 
         // Act
-        Parallel.For(0, 10, _ => storeReader.ReadEntryImageData(entry));
+        var result = Parallel.For(0, 10, _ => storeReader.ReadEntryImageData(entry));
 
-        // No Assert - test passes if no exceptions are thrown
+        // Test passes if no exceptions are thrown, but we can also assert completion
+        result.IsCompleted.Should().BeTrue();
     }
 }


src/Sentry.Android.AssemblyReader/V2/AssemblyStoreReader.cs
```diff
@@ -11,6 +11,7 @@ internal abstract class AssemblyStoreReader
 
     private static readonly UTF8Encoding ReaderEncoding = new UTF8Encoding(false);
 
+    internal object StreamLock { get; } = new();
```
Flash0ver: question: Could / Should we use `System.Threading.Lock` instead?

CHANGELOG.md
```diff
@@ -6,6 +6,10 @@
 
 - Extended `SentryThread` by `Main` to allow indication whether the thread is considered the current main thread ([#4807](https://github.com/getsentry/sentry-dotnet/pull/4807))
 
+### Fixes
+
+- Fixed threadsafety issue on Android when multiple events are captured concurrently ([#4814](https://github.com/getsentry/sentry-dotnet/pull/4814))
```
Flash0ver: suggestion: I believe this should be hyphenated.

```suggestion
- Fixed thread-safety issue on Android when multiple events are captured concurrently ([#4814](https://github.com/getsentry/sentry-dotnet/pull/4814))
```

test/Sentry.Android.AssemblyReader.Tests/StoreReaderTests.cs
```diff
@@ -0,0 +1,50 @@
...
+namespace Sentry.Android.AssemblyReader.Tests;
+
+public class StoreReaderTests
+{
+    [Fact]
+    public void IsSupported_Concurrent_IsThreadSafe()
+    {
+        // Arrange
+        var buffer = new byte[1024 * 1024];
+        var memoryStream = new MemoryStream(buffer);
```
Flash0ver: thought (non-blocking): Dispose

We could make this a `using` expression .. but totally optional, I believe, as it's an array-based Memory-Stream ... unless I'm mistaken.

```suggestion
        using var memoryStream = new MemoryStream(buffer);
```

test/Sentry.Android.AssemblyReader.Tests/StoreReaderTests.cs
```diff
@@ -0,0 +1,50 @@
...
+
+        // No Assert - test passes if no exceptions are thrown
+    }
+
+    [Fact]
+    public void ReadEntryImageData_Concurrent_IsThreadSafe()
+    {
+        // Arrange
+        var buffer = new byte[1024 * 1024];
+        var memoryStream = new MemoryStream(buffer);
```
Flash0ver: thought (non-blocking): Dispose

We could make this a using expression .. but totally optional, I believe, as it's an array-based Memory-Stream ... unless I'm mistaken.

```suggestion
        using var memoryStream = new MemoryStream(buffer);
```

test/Sentry.Android.AssemblyReader.Tests/StoreReaderTests.cs
```diff
@@ -0,0 +1,50 @@
...
+    [Fact]
+    public void IsSupported_Concurrent_IsThreadSafe()
+    {
+        // Arrange
+        var buffer = new byte[1024 * 1024];
+        var memoryStream = new MemoryStream(buffer);
+        var storeReader = new StoreReader(memoryStream, "testStore", null);
+
+        // Act
+        Parallel.For(0, 10, _ => storeReader.IsSupported());
```
Flash0ver: question: Assertion

Should we, instead of no assertion, maybe assert the `ParallelLoopResult` returned by `Parallel.For`?

```CSharp
Assert.True(result.IsCompleted);
```

If we'd like to do that, please also apply to the test method below: `ReadEntryImageData_Concurrent_IsThreadSafe`.

src/Sentry.Android.AssemblyReader/V2/AssemblyStoreReader.cs
```diff
@@ -44,34 +45,44 @@ protected AssemblyStoreReader(Stream store, string path, DebugLogger? logger)
...
+                    var nread = StoreStream.Read(buffer, 0, (int)Math.Min(bufferSize, remainingToRead));
+                    stream.Write(buffer, 0, nread);
+                    remainingToRead -= (long)nread;
+                }
+                stream.Flush();
+                stream.Seek(0, SeekOrigin.Begin);
+            }
+            finally
+            {
+                Utils.BytePool.Return(buffer);
```
Flash0ver, line 82: praise(perf): fixed ArrayPool exhaustion 🤘🥔

src/Sentry.Android.AssemblyReader/V2/AssemblyStoreReader.cs
```diff
@@ -44,34 +45,44 @@ protected AssemblyStoreReader(Stream store, string path, DebugLogger? logger)
...
-        {
-            throw new NotImplementedException();
-        }
-
-        const long BufferSize = 65535;
-        byte[] buffer = Utils.BytePool.Rent((int)BufferSize);
-        long remainingToRead = entry.DataSize;
-
-        while (remainingToRead > 0)
+        lock (StreamLock)
```
Flash0ver, line 54: thought: lock on `Stream`

First I thought: why not lock on the actual `StoreStream` object, rather than creating an extra object to lock on?
But now realizing that we are passing in the `Stream` via the constructor, I now have the feeling it's safer to have a separate lock-object owned by `this` instance.

src/Sentry.Android.AssemblyReader/V2/AssemblyStoreReader.cs
```diff
@@ -44,34 +45,44 @@ protected AssemblyStoreReader(Stream store, string path, DebugLogger? logger)
 
     protected BinaryReader CreateReader() => new BinaryReader(StoreStream, ReaderEncoding, leaveOpen: true);
 
-    protected abstract bool IsSupported();
+    protected internal abstract bool IsSupported();
```
Flash0ver, line 48: thought (non-blocking): alternatively derive a "Spy" for testing

Instead of making the members `protected internal`, so that they can be accessed by the tests as well,
we could instead derive a new type in the Tests, maybe as "file-scoped type", which then in turn makes these `protected` members accessibly by offering a `public` or `internal` member that calls the `protected` member.

But it's just a thought I had ... considering that these types are `internal` and Android-only, it's also fine as is. I just wanted to write down my thought, maybe you like this solution more.

jamescrosswell: I like the idea of a Spy as it makes it more explicit why some members have internal visibility... but I think we should get this fix out ASAP so let's consider a spy for future members that we need to expose for testing.

PR #4818, opened 2025-12-17
chore: Enable automatic release management

jamescrosswell:
We frequently run into merge coflicts on the `CHANGELOG.md` file, so we're trying [automatic changelog management](https://github.com/getsentry/craft?tab=readme-ov-file#changelog-policies).
- [Docs](https://getsentry.github.io/craft/configuration/#auto-mode)

I copied the initial `release.yml` file [from the Python repo](https://github.com/getsentry/sentry-python/blob/86d326cb0148a28c2ec3d5038431ab99110902dc/.github/release.yml) and then tweaked this to match how we've historically done change logs in this repo.

For this to work, we also need to consider the automatic changelog generation that happens when [Updater](https://github.com/getsentry/github-workflows/blob/main/updater/README.md) runs to update dependencies. Unfortunately it seems the only option there is to either enable or disable changelog generation for new PRs. I've disabled it then. 

> [!TIP]
> Perhaps later we can circle back to use the [original script](https://github.com/getsentry/github-workflows/blob/main/updater/scripts/get-changelog.ps1) that was used to generate changelog entries somehow in conjunction with the automatic changelog management. 

#skip-changelog

### DIFF
--- a/.generated.NoMobile.sln
+++ b/.generated.NoMobile.sln
@@ -222,6 +222,9 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "scripts", "scripts", "{A8A9
 		scripts\generate-cocoa-bindings.ps1 = scripts\generate-cocoa-bindings.ps1
 		scripts\generate-solution-filters.ps1 = scripts\generate-solution-filters.ps1
 		scripts\generate-solution-filters-config.yaml = scripts\generate-solution-filters-config.yaml
+		scripts\device-test-utils.ps1 = scripts\device-test-utils.ps1
+		scripts\install-libssl1.sh = scripts\install-libssl1.sh
+		scripts\patch-cocoa-bindings.cs = scripts\patch-cocoa-bindings.cs
 	EndProjectSection
 EndProject
 Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = ".github", ".github", "{EC6ADE8A-E557-4848-8F03-519039830B5F}"
@@ -241,6 +244,7 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "workflows", "workflows", "{
 		.github\workflows\vulnerabilities.yml = .github\workflows\vulnerabilities.yml
 		.github\workflows\device-tests-ios.yml = .github\workflows\device-tests-ios.yml
 		.github\workflows\device-tests-android.yml = .github\workflows\device-tests-android.yml
+		.github\workflows\changelog-preview.yml = .github\workflows\changelog-preview.yml
 	EndProjectSection
 EndProject
 Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "actions", "actions", "{5D50D425-244F-4B79-B9F5-21D26DD52DC1}"
--- a/.github/workflows/alpine.yml
+++ b/.github/workflows/alpine.yml
@@ -25,7 +25,7 @@ jobs:
     steps:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
 
-      - uses: docker/login-action@5e57cd118135c172c3672efd75eb46360885c0ef # v3.6.0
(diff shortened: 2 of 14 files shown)


.github/release.yml
```diff
@@ -0,0 +1,33 @@
+changelog:
```
Flash0ver, line 1: question: keep current style

Currently we have (sometimes) `BREAKING CHANGES`, and usually `Features`, `Fixes`, `Dependencies`.
Shall we keep that style, and suppress non-user facing changes?

.github/release.yml
jamescrosswell, line 1: We should update this to match how we currently do changelogs: 
- Features, Fixes, Dependencies

In order for that to work, we need to configure dependabot to not generage changelogs.

.generated.NoMobile.sln
```diff
@@ -217,7 +217,6 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "scripts", "scripts", "{A8A9
 		scripts\build-sentry-cocoa.sh = scripts\build-sentry-cocoa.sh
 		scripts\update-project-xml.ps1 = scripts\update-project-xml.ps1
 		scripts\build-sentry-native.ps1 = scripts\build-sentry-native.ps1
-		scripts\ios-simulator-utils.ps1 = scripts\ios-simulator-utils.ps1
```
jamescrosswell: I have no idea where this is coming from. The file doesn't appear to exist in the repo so seems OK to remove it from the solution file (which is generated in any case).

.github/release.yml
```diff
@@ -0,0 +1,23 @@
...
+  policy: auto
+  exclude:
+    labels:
+      - skip-changelog
+    authors:
+      - dependabot
+  categories:
```
Flash0ver, line 8: question: `BREAKING CHANGES`

Do we want to add a "BREAKING CHANGES" section?
See pattern: https://craft.sentry.dev/configuration/#default-configuration

(which I believe would need to go first)

.github/release.yml
```diff
@@ -0,0 +1,23 @@
...
+    - title: Features ✨
```
Flash0ver, line 14: question: emojis

Do we want to adapt the emojis as well?
We haven't used emojis so far ... just text for the header names.

.github/release.yml
```diff
@@ -0,0 +1,23 @@
...
+      labels:
+        - "*"
+      commit_patterns:
+        - "^feat\\b"
```
Flash0ver: question: `commit_patterns`

The `commit_patterns` differ from:
https://craft.sentry.dev/configuration/#default-configuration

Not sure if they're fine .. or we need to update them.

.github/release.yml
```diff
@@ -0,0 +1,23 @@
...
+    - title: Fixes 🐛
+      labels:
+        - "*"
+      commit_patterns:
+        - "^(fix|bugfix)\\b"
+    - title: Dependencies ⬆️
```
Flash0ver, line 24: question: `semver`

Do we also want to adapt the `semver` policy.
Or not yet?

I believe this is requires for ... or maybe by this change (`"auto"`):
https://github.com/getsentry/sentry-dotnet/pull/4835/changes#diff-87db21a973eed4fef5f32b267aa60fcee5cbdf03c67fafdc2a9b553bb0b15f34

.github/release.yml
jamescrosswell: I don't think anyone will mind the addition of emojis. These mostly match what's being used in python (except the dependencies one - they don't have that section in Python so I had to make that one up).

.github/release.yml
jamescrosswell: Done

.github/release.yml
jamescrosswell: I tightened up our commit patterns so they require a semicolon after the prefixes. I've excluded the `!` character since anything with a `!` should be matched by the new "Breaking changes" section.

.github/release.yml
jamescrosswell: I'd rather do this incrementally... for the first go at this, let's continue to set the version numbers manually. Later, once we're comfortable everything is working with the change log generation, we can investigate automatic versioning.

PR #4320, opened 2025-07-02
App Hang Tracking for iOS is now disabled by default

jamescrosswell:
Resolves #4150:
- https://github.com/getsentry/sentry-dotnet/issues/4150#issuecomment-3025873457

Issue #4150, opened 2025-04-28
(MAUI) False App Hang on iOS


lucas-zimerman:
### Package

Sentry.Maui

### .NET Flavor

Mono

### .NET Version

9.0.4

### OS

iOS

### OS Version

15.8.4

### Development Environment

Visual Studio v17.x

### SDK Version

8.46.0

### Self-Hosted Sentry Version

24.8.0

### Workload Versions

Related to https://github.com/getsentry/team-mobile/issues/186

Calls to clipboard are being marked as app hang on the native code

### UseSentry or SentrySdk.Init call

```
        builder.UseSentry((option) =>
        {
            option.SendDefaultPii = true;
            option.ReportAssembliesMode = ReportAssembliesMode.None;
            option.Dsn = EnvironmentConfig.SentryDsn;
            option.AutoSessionTracking = false;
            option.TracesSampleRate = 1.0;
            option.AddEventProcessor(new SentryEventProcessor());
            option.AddExceptionProcessor(new SentryExceptionProcessor());
            option.AddExceptionFilterForType<InvalidSessionException>();
            option.Environment = EnvironmentConfig.AppEnvironment.ToString();
            option.AttachScreenshot = true;
            option.SetBeforeScreenshotCapture(BeforeScreenshotCapture);
            if (EnvironmentConfig.AppEnvironment == EAppEnvironment.Prod)
            {
                option.MaxBreadcrumbs = 50;
            }
        });

```

### Steps to Reproduce

On an entry field, try to use the clipboard for more than 2 seconds

### Expected Result

Should not show app hang event

### Actual Result

https://sentry.lkp.app.br/share/issue/b445df88db814198a034bcb6eb6bbaa7/

```
  CoreFoundation      0x180cb3994  <redacted>
  CoreFoundation      0x180c8b530  <redacted>
  CoreFoundation      0x180c8b530  <redacted>
  CoreFoundation      0x180c4cd20  <redacted>
  CoreFoundation      0x180c56e48  <redacted>
  Foundation          0x1823404e0  <redacted>
  UIKitCore           0x183590f7c  <redacted>
  UIKitCore           0x1831de5ac  <redacted>
  libobjc.A           0x1984bb7d0  <redacted>
  libobjc.A           0x1984b8858  <redacted>
  libobjc.A           0x1984c17d4  <redacted>
  UIKitCore           0x1834800f8  <redacted>
  libobjc.A           0x1984bb7d0  <redacted>
[...]

PR #4830, opened 2026-01-07
Update bindings for Cocoa SDK v9

jpnurmi:
> [!NOTE]
> - `PrivatesHeader.h` and `PrivateSentrySDKOnly.h` were moved from `PrivateHeaders/` to `Headers/`
> - `SentryOptions.enableAppHangTrackingV2` was removed: https://github.com/getsentry/sentry-cocoa/pull/6334
> - `SentryOptions.inAppExcludes` was removed: https://github.com/getsentry/sentry-cocoa/pull/6646
> - `PrivateSentrySDKOnly.getDebugImages` was removed: https://github.com/getsentry/sentry-cocoa/pull/6454

#skip-changelog (merges to #4781)

### DIFF
--- a/src/Sentry/Platforms/Cocoa/SentrySdk.cs
+++ b/src/Sentry/Platforms/Cocoa/SentrySdk.cs
@@ -115,7 +115,9 @@ public static partial class SentrySdk
         nativeOptions.AppHangTimeoutInterval = options.Native.AppHangTimeoutInterval.TotalSeconds;
         nativeOptions.IdleTimeout = options.Native.IdleTimeout.TotalSeconds;
         nativeOptions.Dist = options.Distribution;
-        nativeOptions.EnableAppHangTracking = options.Native.EnableAppHangTracking;
+#pragma warning disable CS0618 // Type or member is obsolete
+        nativeOptions.EnableAppHangTracking = options.Native.EnableAppHangTracking || options.Native.EnableAppHangTrackingV2;
+#pragma warning restore CS0618 // Type or member is obsolete
         nativeOptions.EnableAutoBreadcrumbTracking = options.Native.EnableAutoBreadcrumbTracking;
         nativeOptions.EnableAutoPerformanceTracing = options.Native.EnableAutoPerformanceTracing;
         nativeOptions.EnableCoreDataTracing = options.Native.EnableCoreDataTracing;


jamescrosswell (approved):
Nice work - thanks @jpnurmi !

modules/sentry-cocoa.properties
Flash0ver, line 1: question:
- close #4781

This replaces the above auto-generated PR, right?

scripts/patch-cocoa-bindings.cs
```diff
@@ -72,6 +72,10 @@
     .WithAttribute("SentryBeforeBreadcrumbCallback", "return: NullAllowed")
     .WithAttribute("SentryBeforeSendEventCallback", "return: NullAllowed")
     .WithAttribute("SentryTracesSamplerCallback", "return: NullAllowed")
+    // Fix nullable return attributes
+    .RemoveAttribute("PrivateSentrySDKOnly", "Capture*", "NullAllowed")
```
Flash0ver: question:

Should we be explicitly removing the `NullAllowed` attribute for `CaptureScreenshots` and `CaptureViewHierarchy`, since we are adding `[return: NullAllowed]` on both of these methods below? Or are there more?

I am wondering if we could accidentally miss some new `Capture*` methods in the future.

src/Sentry/Platforms/Cocoa/SentryOptions.cs
Flash0ver: suggestion: change phrasing

```diff
-/// IMPORTANT: This feature is experimental and may have bugs.
+/// IMPORTANT: App Hangs V2 is now the default.
```

(didn't do a GitHub suggestion, as the result looked incorrect)

Flash0ver, line 76: suggestion: add clarification

Add the updated text from: https://docs.sentry.io/platforms/apple/configuration/app-hangs/#app-hangs-v2

```diff
 /// As of version 8.39.0-beta.1 of the sentry-cocoa SDK, you can enable AppHangsV2, which is available on iOS and tvOS.
+/// Starting with version 9.0.0, App Hangs V2 is enabled by default.
 /// The main difference is that AppHangsV2 differentiates between fully-blocking and non-fully-blocking
```

(I didn't do a GitHub suggestion, as the result looked a bit weird in formatting)

src/Sentry/Platforms/Cocoa/SentryOptions.cs
```diff
@@ -83,6 +83,7 @@ internal NativeOptions(SentryOptions options)
         /// <remarks>
         /// See https://docs.sentry.io/platforms/apple/configuration/app-hangs/#app-hangs-v2
         /// </remarks>
+        [Obsolete("Use EnableAppHangTracking")]
         public bool EnableAppHangTrackingV2 { get; set; } = false;
```
Flash0ver, line 88: question: forward to `EnableAppHangTracking`

Since has no effect anymore,
shall we "forward" it to `EnableAppHangTracking`?
Like
```CSharp
public bool EnableAppHangTrackingV2 { get => EnableAppHangTracking; set => EnableAppHangTracking = value; }
```

Or do we indeed want to make this option essentially no-op, since this option is now `Obsolete`.
In this case, I'd make the message more clear:
```CSharp
[Obsolete("This option has no effect. Use EnableAppHangTracking instead.")]
//or maybe
[Obsolete("App Hangs V2 is now the default. Use EnableAppHangTracking instead.")]
```
or something similar.

I'm a bit uncertain whether we should
- make `EnableAppHangTrackingV2` no-op
- make `EnableAppHangTrackingV2` an alias to `EnableAppHangTracking`

Considering that we are releasing the Cocoa SDK major version in a minor .NET SDK version,
what is the "least breaking" option?

Also considering all combinations/permutations, that the user might have:
```CSharp
options.EnableAppHangTracking = true|false:
options.EnableAppHangTrackingV2 = true|false;
```

I'm quite uncertain, I'm afraid.

src/Sentry/Platforms/Cocoa/SentryOptions.cs
Flash0ver, line 69: question: default to `true`?

See https://docs.sentry.io/platforms/apple/configuration/app-hangs/:
> Starting with version 8.0.0, this feature has been enabled by default.

But also considering, that we are releasing this changeset as a minor version,
perhaps we do not want to change the default yet,
but instead open an issue to align the default with our `v7` end of the year.

Or ... to we want to keep this default to `false` in general?

src/Sentry/Platforms/Cocoa/BindableSentryOptions.cs
```diff
@@ -35,7 +35,9 @@ public void ApplyTo(SentryOptions.NativeOptions options)
             options.AppHangTimeoutInterval = AppHangTimeoutInterval ?? options.AppHangTimeoutInterval;
             options.IdleTimeout = IdleTimeout ?? options.IdleTimeout;
             options.EnableAppHangTracking = EnableAppHangTracking ?? options.EnableAppHangTracking;
+#pragma warning disable CS0618 // Type or member is obsolete
             options.EnableAppHangTrackingV2 = EnableAppHangTrackingV2 ?? options.EnableAppHangTrackingV2;
```
Flash0ver, line 39: question: obsolete

Should we mark this option (`BindableSentryOptions.EnableAppHangTrackingV2`) also as obsolete?
Or is that not necessary as this type is `internal`, and we won't "forget" to remove it for our next major.

src/Sentry/Platforms/Cocoa/SentryOptions.cs
```diff
@@ -83,6 +83,7 @@ internal NativeOptions(SentryOptions options)
         /// <remarks>
         /// See https://docs.sentry.io/platforms/apple/configuration/app-hangs/#app-hangs-v2
         /// </remarks>
+        [Obsolete("Use EnableAppHangTracking")]
```
Flash0ver: question: v7?

Do we plan to remove this option for our next major - v7 - in November 2026?
Shall we create an issue already, and mark it for "Major v7" in Linear?

src/Sentry/Platforms/Cocoa/SentryOptions.cs
jpnurmi: Looks like it was deliberately disabled in Sentry.NET by default:

- https://github.com/getsentry/sentry-dotnet/pull/4320
- https://github.com/getsentry/sentry-dotnet/issues/4150

modules/sentry-cocoa.properties
jpnurmi: This does only the noisy binding update part and merges into #4781. There's still more to do, such as bumping the minimum iOS version from 13.0 to 15.0.

scripts/patch-cocoa-bindings.cs
jpnurmi: Sounds good, fixed in https://github.com/getsentry/sentry-dotnet/pull/4830/commits/70594e82122603a75757c1861948f39798ed05ba

src/Sentry/Platforms/Cocoa/SentryOptions.cs
jpnurmi: Thanks, fixed in https://github.com/getsentry/sentry-dotnet/pull/4830/commits/3884b507e3488821fab957d3495e98f4f33036f6

jpnurmi: Added in https://github.com/getsentry/sentry-dotnet/pull/4830/commits/b57e205d047e3eb24d615164d80b4c3d954755cb :+1:

src/Sentry/Platforms/Cocoa/SentryOptions.cs
jpnurmi: Thanks, the forwarding proposal feels the "least breaking" option: https://github.com/getsentry/sentry-dotnet/pull/4830/commits/66618f978ffd14a43dac03a8da3530d1fc41afec

> Also considering all combinations/permutations, that the user might have:
> ```
> options.EnableAppHangTracking = true|false:
> options.EnableAppHangTrackingV2 = true|false;
> ```

Fortunately, accessing `EnableAppHangTrackingV2` now throws warnings, so any problematic combo would not go unnoticed unless they ignore warnings, right?

src/Sentry/Platforms/Cocoa/BindableSentryOptions.cs
jpnurmi: I followed the example of `BindableSentryOptions.NativeOptions.EnableWatchdogTerminationTracking` in the same file, but I don't have strong feelings either way. :)

src/Sentry/Platforms/Cocoa/SentryOptions.cs
jpnurmi: I ended up reverting the forwarding and going back to no-op after the feedback from Cursor

- https://github.com/getsentry/sentry-dotnet/pull/4830#discussion_r2726757666
- https://github.com/getsentry/sentry-dotnet/pull/4830#discussion_r2726846994

src/Sentry/Platforms/Cocoa/SentryOptions.cs
Flash0ver: I see ... now I prefer not to change the default yet in a minor version ... but open an issue for `v7` to do so in the next major.

src/Sentry/Platforms/Cocoa/BindableSentryOptions.cs
Flash0ver: Right ... let's keep that.
Local consistency for the win!

modules/sentry-cocoa.properties
Flash0ver: Right ... sorry ... I overlooked the merge target branch.

src/Sentry/Platforms/Cocoa/SentryOptions.cs
jpnurmi: Next attempt, proposed by @Flash0ver: https://github.com/getsentry/sentry-dotnet/pull/4830/commits/a9f36105b0455dd15116e77ca753285d389ce92a

src/Sentry/Platforms/Cocoa/SentryOptions.cs
Flash0ver: opened https://github.com/getsentry/sentry-dotnet/issues/4877

src/Sentry/Platforms/Cocoa/SentryOptions.cs
```diff
@@ -248,10 +249,9 @@ internal NativeOptions(SentryOptions options)
         /// <remarks>
         /// https://docs.sentry.io/platforms/apple/configuration/options/#in-app-exclude
         /// </remarks>
+        [Obsolete("This option had no effect and was removed.")]
         public void AddInAppExclude(string prefix)
```
Flash0ver, line 253: todo: remove obsolete option
for the next major version
see https://github.com/getsentry/sentry-dotnet/issues/4877

src/Sentry/Platforms/Cocoa/SentryOptions.cs
Flash0ver: yes

todo: remove obsolete option
for the next major version
see https://github.com/getsentry/sentry-dotnet/issues/4877

PR #4784, opened 2025-12-02
feat(logs): apply default attributes from Scope

Flash0ver:
fixes #4493
closes #4733

- #4493
- #4733

See docs: https://develop.sentry.dev/sdk/telemetry/logs/#default-attributes

---

### Changes
- Adds missing "default attributes" as per Sentry Developer Documentation.
- Applying the values - if available - from the current `Scope`.
  - Changed the `internal` APIs for that

However, it seems that we're missing a lot of the **Contexts** on the `Scope`,
where we currently mostly _apply_ them to `Event`s.

### TODO
- Enrich the `Scope` with the data needed
  - or set the data directly for the time being
- add `CHANGELOG.md` entry

Issue #4749, opened 2025-11-18
Metrics Docs


dingsdax:
* create user docs (see [Python](https://docs.sentry.io/platforms/python/metrics/), [Javascript](https://docs.sentry.io/platforms/javascript/metrics/))
* clarify whether to add quickstart example for your SDK
* when merged, change SDK from [upcoming](https://docs.sentry.io/product/explore/metrics/getting-started/#upcoming-sdks) to [supported](https://docs.sentry.io/product/explore/metrics/getting-started/#supported-sdks)

Metrics Feature-Issue:
- #4741

Flash0ver:
Initial docs PR:
- getsentry/sentry-docs#16054

Flash0ver:
Metrics: Getting Started page:
- https://github.com/getsentry/sentry-docs/pull/16262

Flash0ver:
Published Docs: https://docs.sentry.io/platforms/dotnet/metrics/

Also, added Getting-Started docs and Onboarding code snippets to Sentry
- https://github.com/getsentry/sentry/pull/106551
- https://github.com/getsentry/sentry/pull/108034

Issue #4741, opened 2025-11-16
Sentry Metrics for .NET


dingsdax:
[https://develop.sentry.dev/sdk/telemetry/metrics/](https://develop.sentry.dev/sdk/telemetry/metrics/)

Add support for [Sentry Metrics ](https://docs.sentry.io/product/explore/metrics/getting-started/)to the .NET SDK!

This will enable users to record [counters, gauges, and distributions](https://develop.sentry.dev/sdk/telemetry/metrics/#metric-types).

1. Define the [Metrics Protocol](https://develop.sentry.dev/sdk/telemetry/metrics/#metrics-protocol) and `trace_metric` envelope type in the SDK.
2. Add the Public API (SDK options and methods) as per the [docs](https://develop.sentry.dev/sdk/telemetry/metrics/#public-api).
3. Make sure the the SDK follow the documented [behavior](https://develop.sentry.dev/sdk/telemetry/metrics/#sdk-behavior).
4. Attach [default attributes](https://develop.sentry.dev/sdk/telemetry/metrics/#default-attributes) and optional [user attributes](https://develop.sentry.dev/sdk/telemetry/metrics/#user-attributes) to the SDK as per docs.
5. Follow requirements for rate limiting, buffering, batching & flushing and build in measures to prevent excessive memory usage.

For general feedback & questions about Sentry Metrics use this [GH discussion](https://github.com/getsentry/sentry/discussions/102275); for .NET SDK specific feedback & questions use the comment functionality on this GH issue.

intrepid-developer:
Do you have a planned release for this yet? This would be really good for us 😄 

And, will the .Net version include all the .Net metrics that we see in things like the Aspire Dashboard?

E.g. these:

<img width="547" height="1135" alt="Image" src="https://github.com/user-attachments/assets/8eb18c97-f48b-4414-ae3c-bb3bc4cad5ca" />

Flash0ver:
Metrics are currently in beta, and we are continuously widening the support for our SDKs.

And our .NET SDK is indeed on the Roadmap: we plan to ship Metrics as an `Experimental` API soon after we publish `v6.0.0` as stable ... depending a bit on the upcoming holiday season.
We're starting with an `Experimental` API since the feature's still in _Beta_, and also for us to get feedback on the API shape. In particular on the "Attributes"-related parts, as "Attributes" will be quite pervasive in future versions of the SDK.

[...]

Issue #4733, opened 2025-11-13
Add device attributes to Logs in the MAUI SDK


dalnoki:
### Problem Statement

The Sentry MAUI SDK currently [does not include ](https://docs.sentry.io/platforms/dotnet/guides/maui/logs/#default-attributes) device-specific attributes (such as device.model, device.platform, device.manufacturer, and device.brand) in structured logs, even though this information is collected and sent with error events and issues.

munkii:
As the customer that triggered support rasing this ticket I thought I'd had some background. As it currently stands the logs and more importantly the lack of supplementary data mean they really only work in the scope of a single developer with one maybe two devices.

As soon as QA or production gets involved we need to know what devices are triggering which logs and that is impossible at the moment.

munkii:
As a workaround I haved changed SetBeforeSendLog to read the User details out of Scope and then apply them to Log with SetAttribute

 ```
options.Experimental.SetBeforeSendLog(log =>
{
#if DEBUG
	SentryLogLevel minimumLevel = SentryLogLevel.Debug;
#else
	SentryLogLevel minimumLevel = SentryLogLevel.Info;
#endif

	log.SetAttribute("DeviceName", DeviceInfo.Name);
	log.SetAttribute("Manufacturer", DeviceInfo.Manufacturer);
	log.SetAttribute("OSVersion", DeviceInfo.VersionString);
	log.SetAttribute("DeviceModel", DeviceInfo.Model);

	SentryUser sentryUser = null; ;


	SentrySdk.ConfigureScope(scope =>
	{
		if (scope.HasUser())
		{
			sentryUser = scope.User;
		}

	});

	if (sentryUser != null) {
		log.SetAttribute("UserId", sentryUser.Id);
		log.SetAttribute("UserName", sentryUser.Username);
	}

	if (log.Level < minimumLevel)
	{
		return null;
	}
	
	return log;
});
```

PR #4834, opened 2026-01-09
feat(metrics): Trace-connected Metrics

Flash0ver:
Implementation of Trace-connected Metrics
See https://develop.sentry.dev/sdk/telemetry/metrics/

This changeset contains the initial implementation of Trace-connected Metrics for Sentry.

### Changes
- made Batch-Buffer generic, so that it can be used for both Structured Logging and Trace-connected Metrics
- added new types (`SentryTraceMetrics` and derived), which follow the same pattern and structure from Structured Logs
- the SDK-APIs are generic, which are inspired by the [System.Diagnostics.Metrics](https://learn.microsoft.com/dotnet/core/diagnostics/metrics-instrumentation) APIs of the BCL / `System.Diagnostics.DiagnosticSource` package
- naming of the public APIs differs a bit from the SDK spec in order to be idiomatic .NET
- currently, made APIs experimental

Please see the discussion comments below to discuss these design decisions.

### Related changesets
- Tests: #4839
- Samples: #4841
- Analyzers: #4840
- Documentation
  - Platform Docs: getsentry/sentry-docs#16054
  - Getting Started page: getsentry/sentry-docs#16262
- Sentry
  - for all .NET platforms, except Xamarin
  - enable Metrics: getsentry/sentry#106551
  - Onboarding pages: getsentry/sentry#106552

### Related issues and discussions
- #4741
- #4749
- #4838

### Follow-ups
- More default Attributes: #4784

### Enable public beta
- Sentry Docs: `dotnet` platform
- Sentry Docs: Supported Platforms overview
- Sentry Onboarding
- then, enable in Sentry "Platform Categories"

### DIFF
--- a/test/Sentry.Tests/SentryMetricEmitterTests.Types.cs
+++ b/test/Sentry.Tests/SentryMetricEmitterTests.Types.cs
@@ -289,15 +289,15 @@ public partial class SentryMetricEmitterTests
     }
 
     [Fact]
-    public void Type_EmitMethods_StringUnitParameterIsObsoleteForForwardCompatibility()
+    public void Type_EmitMethods_StringUnitParameterOverloadsAreObsoleteForForwardCompatibility()
     {
         var type = typeof(SentryMetricEmitter);
 
         type.Methods()
-            .Where(static method => method.IsPublic && method.ReturnType == typeof(void) && method.IsGenericMethod && method.Name.StartsWith("Emit"))
-            .Should().NotBeEmpty().And.AllSatisfy(static method =>
+            .Where(method => method.IsPublic && method.ReturnType == typeof(void) && method.IsGenericMethod && method.Name.StartsWith("Emit"))
+            .Should().NotBeEmpty().And.AllSatisfy(method =>
             {
-                var unitParameter = method.GetParameters().SingleOrDefault(static parameter => parameter.Name == "unit");
+                var unitParameter = method.GetParameters().SingleOrDefault(parameter => parameter.Name == "unit");
 
                 if (unitParameter is null || unitParameter.ParameterType == typeof(MeasurementUnit))
                 {


src/Sentry/SentryMetric.cs
```diff
@@ -0,0 +1,357 @@
...
+
+    /// <summary>
+    /// Set a key-value pair of data attached to the metric.
+    /// </summary>
+    public void SetAttribute(string key, object value)
+    {
+        _attributes ??= new Dictionary<string, object>();
+
+        _attributes[key] = new SentryAttribute(value);
+    }
```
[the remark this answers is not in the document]
Flash0ver: fixed in https://github.com/getsentry/sentry-dotnet/pull/4834/changes/e65db5f13c9b976fedf3b4b726e20fb5659742a5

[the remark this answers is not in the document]
Flash0ver: fixed in https://github.com/getsentry/sentry-dotnet/pull/4834/changes/e65db5f13c9b976fedf3b4b726e20fb5659742a5

src/Sentry/SentryMetricEmitter.Counter.cs
```diff
@@ -0,0 +1,57 @@
...
+public abstract partial class SentryTraceMetrics
+{
+    /// <summary>
+    /// Increment a counter.
+    /// </summary>
+    /// <param name="name">The name of the metric.</param>
+    /// <param name="value">The value of the metric.</param>
+    /// <typeparam name="T">The numeric type of the metric.</typeparam>
+    /// <remarks>Supported numeric value types for <typeparamref name="T"/> are <see langword="byte"/>, <see langword="short"/>, <see langword="int"/>, <see langword="long"/>, <see langword="float"/>, and <see langword="double"/>.</remarks>
+    public void EmitCounter<T>(string name, T value) where T : struct
```
Flash0ver, line 12: question: discuss identifiers of public method groups

See spec in dev docs [Metrics > Metrics Module](https://develop.sentry.dev/sdk/telemetry/metrics/#metrics-module)

```diff
-Sentry.metrics.count(..)
+SentrySdk.Metrics.EmitCounter(..);
-Sentry.metrics.gauge(..)
+SentrySdk.Metrics.EmitGauge(..);
-Sentry.metrics.distribution(..)
+SentrySdk.Metrics.EmitDistribution(..);
```

- `count` is both a verb and a noun
- `gauge` is both a verb and a noun
- `distribution` is a noun only

So my thought was to - in particular for `Distribution` - use a verb as a prefix to form a verb phrase.
And then - for consistency - use a verb as a prefix for all three method groups.

The respective [.NET Metrics](https://learn.microsoft.com/dotnet/core/diagnostics/metrics-instrumentation) (and [Open Telemetry](https://opentelemetry.io/docs/languages/dotnet/metrics/)) are

```CSharp
using System.Diagnostics.Metrics;

static Meter meter = new Meter(name: "my-meter");

static Counter<int> counter = meter.CreateCounter<int>(name: "my-counter");
static Gauge<int> gauge = meter.CreateGauge<int>(name: "my-gauge");
static Histogram<int> histogram = meter.CreateHistogram<int>(name: "my-histogram");

counter.Add(delta: 1);
gauge.Record(value: 2);
histogram.Record(value: 3);
```

The first draft started with
```CSharp
SentrySdk.Metrics.AddCounter
SentrySdk.Metrics.RecordGauge
SentrySdk.Metrics.RecordDistribution
```
although with the verbs consistent with .NET/Open-Telemtry, these verbs do not fit, as they may suggest client-side aggregation, rather than sending the metric as-provided to Sentry with query-time aggregation at Sentry.

Then changing to the verb `Emit`. This verb is also used in the dev docs. The verb `Emit` is also used in [Logs in Sentry for Go](https://docs.sentry.io/platforms/go/logs/).

So the current naming scheme is, almost, `Emit{Type}`.
However, currently, instead of `EmitCount`, I decided to use `EmitCounter` instead, to be closer to the .NET BLC / [System.Diagnostics.DiagnosticSource](https://www.nuget.org/packages/System.Diagnostics.DiagnosticSource/) NuGet package.

An argument can be made to instead be consistent with the spec ... or at least more consistent with the spec. So here are some alternatives up for discussion.

```CSharp
// "type": "counter"
SentrySdk.Metrics.Count() // as in spec: "Sentry.metrics.count()"
SentrySdk.Metrics.EmitCount() // as verb phrase
SentrySdk.Metrics.EmitCounter() // Counter instead of Count, for consistency with the actual type name

// "type": "gauge"
SentrySdk.Metrics.Gauge() // as in spec: "Sentry.metrics.gauge()"
SentrySdk.Metrics.EmitGauge() // as verb phrase

// "type": "distribution"
SentrySdk.Metrics.Distribution() // as in spec: "Sentry.metrics.distribution()"
SentrySdk.Metrics.EmitDistribution() // as verb phrase
```

What do you prefer?

UPDATE: oh no ... I just realized we had old metrics before ... should we restore (or adapt parts of) that API shape?

src/Sentry/SentryMetricEmitter.Gauge.cs
```diff
@@ -0,0 +1,73 @@
...
+public abstract partial class SentryTraceMetrics
+{
+    /// <summary>
+    /// Set a gauge value.
+    /// </summary>
+    /// <param name="name">The name of the metric.</param>
+    /// <param name="value">The value of the metric.</param>
+    /// <typeparam name="T">The numeric type of the metric.</typeparam>
+    /// <remarks>Supported numeric value types for <typeparamref name="T"/> are <see langword="byte"/>, <see langword="short"/>, <see langword="int"/>, <see langword="long"/>, <see langword="float"/>, and <see langword="double"/>.</remarks>
+    public void EmitGauge<T>(string name, T value) where T : struct
```
Flash0ver, line 12: question: Overloads and Parameters of Method Groups

The SDK spec [Metrics > Metrics Module](https://develop.sentry.dev/sdk/telemetry/metrics/#metrics-module) suggests these parameters for the three methods `Sentry.metrics.count`, `Sentry.metrics.gauge`, and `Sentry.metrics.distribution`:
- `name`, String, required
- `value`, Number, required
- `options`, Object, optional
  - `unit`, String, optional, only for distribution and gauge, should not be an `enum`
  - `attributes`, Object, optional
  - `scope`, Scope, optional

#### `name`
I decided to have several overloads per method group, rather than a single method per type. This is also what the [Meter type / Instrument types](https://learn.microsoft.com/dotnet/core/diagnostics/metrics-instrumentation) of the `net6.0`+ BCL APIs do.

#### `value`
With .NET being typed, I decided to make the APIs generic, as this is also what the .NET BCL / Open Telemetry / [System.Diagnostics.DiagnosticSource](https://www.nuget.org/packages/System.Diagnostics.DiagnosticSource/) NuGet package does.
This supports the numeric types `byte`, `short`, `int`, `long`, `float`, `double`. But don't support `decimal` (.NET Instruments do support it) as it's 128-bit, and Sentry only supports 64-bit signed integral types and 64-bit floating-point numbers.

#### `options`
Instead of an Options-Bag, I went with a method group and respective overloads, as again, they are similar to the `System.Diagnostics.Metrics` APIs.
However, if we'd like, we could do a struct-based Options-bag, per type (`struct CounterOptions`, `struct GaugeOptions`, and `struct DistributionOptions`).

#### `unit`
The `unit` overload only exists in the `Gauge` and `Distribution` method groups.
Typed to `string`, rather than `Sentry.MeasurementUnit`, as the SDK spec suggests to:
> SDKs must not restrict what can be sent in (e.g. by using an enum) but should offer constants or similar that help customers send in [units we support](https://develop.sentry.dev/sdk/telemetry/attributes/#units).

We could, however, offer an overload that takes `Sentry.MeasurementUnit` instead of a `string`.
Or we could add this overload later on.
The motivation of having a string rather than an `enum` is that Sentry may start supporting more Units at some time, and SDK users should be able to pass them once supported without upgrading the SDK.
Currently, a "custom" unit is not supported, and Relay will either discard or set the Unit to `None` if an unsupported unit is passed by the user --- the metric itself will not be discarded in that case.

#### `attributes`
Offer both an `IEnumerable<KeyValuePair<string, object>>`, and a `ReadOnlySpan<KeyValuePair<string, object>>` overload, again, very similar to "tags" of `System.Diagnostics.Metrics`.

#### `scope`
Just a nullable `Scope`.
Currently not available, but we could now - or later - add an overload for passing a `Scope` without a `unit`:
`..(string name, T value, Scope? scope)`

---

What do you think?

UPDATE: oh no ... I just realized we had old metrics before ... should we restore (or adapt parts of) that API shape?

src/Sentry/SentryOptions.cs
```diff
@@ -1905,4 +1907,55 @@ internal static List<StringOrRegex> GetDefaultInAppExclude() =>
...
+        /// When the delegate throws an <see cref="Exception"/> during invocation, the metric will not be captured.
+        /// </summary>
+        /// <typeparam name="T">The numeric type of the metric.</typeparam>
+        /// <remarks>
+        /// It can be used to modify the metric object before being sent to Sentry.
+        /// To prevent the metric from being sent to Sentry, return <see langword="null"/>.
+        /// Supported numeric value types for <typeparamref name="T"/> are <see langword="byte"/>, <see langword="short"/>, <see langword="int"/>, <see langword="long"/>, <see langword="float"/>, and <see langword="double"/>.
+        /// </remarks>
+        /// <seealso href="https://develop.sentry.dev/sdk/telemetry/metrics/"/>
+        public void SetBeforeSendMetric<T>(Func<SentryMetric<T>, SentryMetric<T>?> beforeSendMetric) where T : struct
```
Flash0ver: question: Design

With the Metric type being generic, supporting `byte`, `short`, `int`, `long`, `float`, and `double`, this API is also quite similar to the `MeterListener` of `System.Diagnostics.Metrics`.
Where you can set one callback per metric type via `SetMeasurementEventCallback`.

See [Collect metrics](https://learn.microsoft.com/dotnet/core/diagnostics/metrics-collection).

What do you think?

UPDATE: oh no ... I just realized we had old metrics before ... should we restore (or adapt parts of) that API shape?

src/Sentry/SentryOptions.cs
```diff
@@ -1905,4 +1907,55 @@ internal static List<StringOrRegex> GetDefaultInAppExclude() =>
...
         "Java.Interop",
     ];
+
+    /// <summary>
+    /// Sentry features that are currently in an experimental state.
+    /// </summary>
+    /// <remarks>
+    /// Experimental features are subject to binary, source and behavioral breaking changes in future updates.
+    /// </remarks>
+    public ExperimentalSentryOptions Experimental { get; }
```
Flash0ver: question: `Experimental` vs Stable.

Do we want to publish it as experimental first ... or stable directly?

CHANGELOG.md
```diff
@@ -1,10 +1,29 @@
 # Changelog
 
-## Unreleased
+## 6.1.0-alpha.2
+
+### BREAKING CHANGES
+
+- Rename [Trace-connected Metrics](https://docs.sentry.io/product/explore/metrics/) APIs to avoid implying aggregation ([#4834](https://github.com/getsentry/sentry-dotnet/pull/4834))
```
jamescrosswell: We probably want to tidy up all of these change log entries when we include metrics in a non-preview release... For people who haven't been using the pre-releases we could just say:
```
- Add _experimental_ support for [Sentry trace-connected Metrics](https://docs.sentry.io/product/explore/metrics/) ([#4834](https://github.com/getsentry/sentry-dotnet/pull/4834))
```

Directory.Build.props
```diff
@@ -1,7 +1,7 @@
 <Project>
 
   <PropertyGroup>
-    <VersionPrefix>6.0.0</VersionPrefix>
+    <VersionPrefix>6.1.0-alpha.2</VersionPrefix>
```
jamescrosswell: Revert before merging into the `main` branch... to whatever the version in the main branch is. This gets set by the release process automatically.

```suggestion
    <VersionPrefix>6.0.0</VersionPrefix>
```

Directory.Build.props
```diff
@@ -13,6 +13,7 @@
 
     <!-- Ignore our own diagnostic ids - these are meant to be external warnings only -->
     <NoWarn>$(NoWarn);SENTRY0001</NoWarn>
+    <NoWarn>$(NoWarn);SENTRYTRACECONNECTEDMETRICS</NoWarn> <!--https://github.com/getsentry/sentry-dotnet/discussions/4838-->
```
jamescrosswell, line 16: Do we need a specific warning for trace connected metrics? In the past we've simply had `SENTRY0001` to ensure SDK users deliberately opt into experimental features (like this one) and don't use these by accident in production.

test/Sentry.Testing/InMemorySentryTraceMetrics.cs
```diff
@@ -0,0 +1,120 @@
+#nullable enable
+
+namespace Sentry.Testing;
+
+public sealed class InMemorySentryTraceMetrics : SentryTraceMetrics
```
jamescrosswell: Where does this get used?

src/Sentry/SentryMetricEmitter.Counter.cs
jamescrosswell: I don't think we have to worry about the API shape of the previous metrics product we had. I think the naming as you have it is fine.

src/Sentry/SentryMetricEmitter.Gauge.cs
jamescrosswell: > Instead of an Options-Bag, I went with a method group and respective overloads, as again, they are similar to the System.Diagnostics.Metrics APIs.

Using an Options type would mean less overloads to maintain... and if the options type was sealed then we could add members to it later. It's probably easier and less ceremony for SDK users to use various overloads with the options they care about. 

It really depends how many dimensions the options type exposes... eventually (like with the `SentryOptions`) the only practical way to set these is via some kind of callback/delegate.

I think it's fine as is for now.

jamescrosswell: > Currently not available, but we could now - or later - add an overload for passing a Scope without a unit:
..(string name, T value, Scope? scope)

What was the rational for omitting this overload currently?

src/Sentry/SentryOptions.cs
jamescrosswell: There's quite a lot of code so we could publish it as experimental for one point release... try to get some feedback from SDK customers using it in anger.  There may be strong opinions some of the design decisions (e.g. method groups vs. options types/callbacks) so good to have the flexibility to revisit some of those decisions for a wee while if need be.

Otherwise it's more of a commercial question (is Sentry sure we're going to GA this metrics product?). I assume we captured enough learnings from the first metrics product to be pretty confident of the design and cost structure this time around. @dingsdax ?

src/Sentry/SentryOptions.cs
jamescrosswell: It's a pain since the SDK user may want a callback like:
```
  if (metric.attributes.contains(a => a.name == "foo"))
  {
    return null;
  }
```

But they'd need to define that function once for each numeric type our metrics support and if we add support for new numeric types their code might break. It doesn't feel like a great API design.

An alternative option (not saying it's better, but just to consider) would be to have the parameter to the callback be a flattening of the various generic types... e.g.
```
SentryMetricArgType
{
    SentryMetricType type;
    string name;
    object value;
    string? unit;
    IEnumerable<KeyValuePair<string, object>> attributes;
    Scope? scope;
}
```

That might be a bit messy: impedance mismatch, mappings, type safety issues etc. but it might be better than the alternative.

A more complex variant would be to pass something like a `SentryMetricTypeAccessor` that mitigated some of the type safety issues by exposing properties of the metric that the SDK user might want to read via getter methods and exposed properties the SDK user might want to modify (if any) via setter methods... so they could do something like:
```
options.Metrics.SetBeforeSend(accessor => {
  if (accessor.GetMetricValue() is double && accessor.Attributes.Contains("foo"))
  {
   return null;
  }
});
```

Probably have to play around with different options to see how the implementation and SDK usage feels in practice. I could see what we have in this PR at the moment being pretty onerous for SDK users though, so I think we need to find a better option.

What solution have the other strongly typed SDKs (like Java) opted for here?

test/Sentry.Testing/InMemorySentryTraceMetrics.cs
Flash0ver: In the test.
See #4839.
I split this up poorly.

Before the final review/approval, we should indeed merge #4839 and #4841 (both targeting _this_ branch) into this branch before,
which I split up in the hopes of making the initial review easier.

Directory.Build.props
Flash0ver: We don't necessarily need a specific `DiagnosticId`.
But with a specific `DiagnosticId`, we may provide a `UrlFormat`, currently leading to the related GitHub discussion.

```CSharp
[Experimental("SENTRYTRACECONNECTEDMETRICS", UrlFormat = "https://github.com/getsentry/sentry-dotnet/discussions/4838")]
```

Alternatively,
we could create a small page in the docs, called "Experiments" or "Experimental Features",
that contain more information for each individual experimental features currently shipped in the _latest_ version.

src/Sentry/SentryOptions.cs
Flash0ver: We'll release it as _Experimental_ feature initially, but going stable by March.

Directory.Build.props
jamescrosswell: OK, let's leave it as is then... bit more work to maintain for us and a hoop to jump through for SDK users but it is nice to have a link to more contextual help.

test/Sentry.Testing/InMemorySentryTraceMetrics.cs
jamescrosswell: Let's merge them now... the code is tightly coupled anyway - I think it's harder to review having them in separate PRs.

src/Sentry/SentryOptions.cs
Flash0ver: re: Attributes
I'd like to "fix" them during a separate follow-up PR, if you don't mind,
to get a dedicated design review on this API / these APIs,
as it will impact not only Metrics, but Logs and Scopes (for now) as well.
Opened #4882.

For now, I went with similar `SetAttribute`/`TryGetAttribute` methods as we have for Logs.
We haven't "burned" the public identifier `Attributes` on Logs yet, so we can then adapt the same Collection-based property for Logs as well, after shipping and testing it during the experimental phase for Metrics.

Flash0ver: re: generic `SetBeforeSendMetric` Options-API
You are right ... if we'd add a new Value-Type (Relay/Sentry may support 128-bit integral/floating-point numbers later on), this would be a breaking change indeed.

Other typed SDKs just have a `double`.
`sentry-cocoa`, having Metrics in Swift, uses `UInt` for `count` and `Double` for `distribution`+`gauge`.

I started out with a `double` as well, but ran into a loss-of-precision problem during testing when converting from `float`:
```CSharp
Metrics.EmitCounter<float>("", 1.1f); // { "value": 1.1 }
Metrics.EmitCounter<double>("", 1.1d); // { "value": 1.1 }
Metrics.EmitCounter<double>("", 1.1f); // { "value": 1.100000023841858 }
```

Well ... we could argue
- the potential imprecision when floats should be used are neglectable
- Relay/Sentry probably won't support > 64-bit numbers anytime soon ... maybe ever

So maybe just a `double` is perfectly fine ... similar to the "old" Metrics 🤔

src/Sentry/SentryMetricEmitter.Gauge.cs
Flash0ver: `Scope` currently does not have `Attributes` yet: opened #4882.
Also the `Scope` type isn't too prominent yet:
- `SentrySdk.ConfigureScope` doesn't return a scope .. but the user could "capture" it out of the delegate
- `SentrySdk.PushScope` returns an `IDisposable` (instance of internal `ScopeSnapshot`)

I was thinking of adding it when we've added `Attributes` to `Scope` in the follow-up.
But we could add and approve it right away to get feedback on the full set of overloads during the experimental phase.

src/Sentry/SentryOptions.cs
```diff
@@ -1905,4 +1907,55 @@ internal static List<StringOrRegex> GetDefaultInAppExclude() =>
...
+        }
+
+        internal SentryTraceMetricsCallbacks? BeforeSendMetricInternal => _beforeSendMetric;
+
+        /// <summary>
+        /// When set to <see langword="false"/>, the SDK does not generate and send metrics to Sentry via <see cref="SentrySdk"/>.
+        /// Defaults to <see langword="true"/>.
+        /// </summary>
+        /// <seealso href="https://develop.sentry.dev/sdk/telemetry/metrics/"/>
+        public bool EnableMetrics { get; set; } = true;
```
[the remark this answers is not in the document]
jamescrosswell: That's a fair comment... enabling by default will result in build warnings or failures for folks upgrading to the new version. If it's experimental, we should disable by default.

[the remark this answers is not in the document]
Flash0ver: It was discussed whether the default for Metrics be `true` or `false`.
The decision was made to be `true` per default. In particular because we don't have a default integration (currently) and the user needs to actively use the Metrics-APIs in order to use Sentry Metrics.

Moreover the boolean flag has no influence on any experimental build warnings, as the SDK-APIs need to be used actively. I'd like to keep the default here to `true`.

https://develop.sentry.dev/sdk/telemetry/metrics/#initialization-options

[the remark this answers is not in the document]
jamescrosswell: I think for GA that makes sense. But assume we include this in a 6.1 release later this week... anyone upgrading from  v6.0.0 to 6.1.0 would start seeing warnings about experimental features and, if they've got warnings as errors (like we do) then their builds would break. As an SDK user, that's not what I'd expect from a point release - we don't want minor upgrades to break people's builds.

So maybe we disable by default while it's experimental and then enable by default once we GA in the 6.2.0 release?

src/Sentry/SentryMetric.Factory.cs
```diff
@@ -0,0 +1,93 @@
...
+        var scope = hub.GetScope();
+        if (scope is not null)
+        {
+            traceId = scope.PropagationContext.TraceId;
+            return;
+        }
+
+        Debug.Assert(hub is not Hub, "In case of a 'full' Hub, there is always a Scope. Otherwise (disabled) there is no Scope, but this branch should be unreachable.");
+        traceId = SentryId.Empty;
+    }
```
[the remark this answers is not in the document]
jamescrosswell: That's a fair call. We could refactor to [HubExtensions](https://github.com/getsentry/sentry-dotnet/blob/fed52443fe8ac0e074d7b0f0be9c815b4176ace0/src/Sentry/HubExtensions.cs#L261-L271).

src/Sentry/Internal/Hub.cs
```diff
@@ -82,6 +81,7 @@ internal Hub(
         }
 
         Logger = SentryStructuredLogger.Create(this, options, _clock);
+        Metrics = SentryTraceMetrics.Create(this, options, _clock);
```
jamescrosswell: I was a little confused by the name of this initially... I wondered what the difference between `SentryMetric` and `SentryTraceMetrics` was - since the both appeared to be dealing with trace metrics. Then I realised one of them was plural... so I guess I'd expect that to be some kind of collection (based purely on the name). 

Actualli the difference is that one of them is a data structure and the other one provides functionality. I wonder if we could do as we've done with logs and use a noun for the data structure and a nounified verb for the bit that does stuff. 

Possibly `SentryMetricEmitter`, `SentryMetricRecorder` or, since it's effectively a wrapper for the batch processor, `SentryMetricProcessor` would make the purpose of the class easier to infer from the name?

src/Sentry/Protocol/StructuredLog.cs
```diff
@@ -34,4 +35,9 @@ public void WriteTo(Utf8JsonWriter writer, IDiagnosticLogger? logger)
         writer.WriteEndArray();
         writer.WriteEndObject();
     }
+
+    internal static void Capture(IHub hub, SentryLog[] logs)
+    {
+        _ = hub.CaptureEnvelope(Envelope.FromLog(new StructuredLog(logs)));
+    }
```
jamescrosswell: Why is this necessary in this PR? I would have expected this change in the logging PR?

src/Sentry/SentryLog.cs
```diff
@@ -189,6 +189,7 @@ internal void SetOrigin(string origin)
         SetAttribute("sentry.origin", origin);
     }
 
+    /// <inheritdoc cref="ISentryJsonSerializable.WriteTo(Utf8JsonWriter, IDiagnosticLogger)" />
```
jamescrosswell: Not strictly related to this PR? We can leave it in but ideally we'd try to minimise noise on big PRs like this... there's already enough other stuff going on.

src/Sentry/SentryMetric.cs
```diff
@@ -263,11 +271,13 @@ internal void SetAttributes(ReadOnlySpan<KeyValuePair<string, object>> attribute
         }
     }
 
-    internal void Apply(Scope? scope)
+    private void Apply(Scope? scope)
```
jamescrosswell: Probably rename this `SetScopeAttributes` to make it clear to calling code what it will be doing... unless it's doing something beyond that. Scope.Apply is used elsewhere to do all kinds of magical things.

src/Sentry/SentryMetric.Factory.cs
jamescrosswell, line 1: We could possibly merge this into the `SentryMetric.cs` file... I'm not sure about you but I'm getting RSI from having to scroll through all the files in the Internal folder with my mouse. Maybe the answer is to break them  down further into subfolders or something.

src/Sentry/SentryUnits.cs
jamescrosswell, line 1: This seems to overlap heavily with [MeasurementUnit](https://github.com/getsentry/sentry-dotnet/blob/0fde2e34300a961fd62fcfd77614ec51b11275e3/src/Sentry/MeasurementUnit.cs)

Is there any way we can reuse/tweak the existing classes and avoid reinventing the wheel?

test/Sentry.Tests/Protocol/TraceMetricTests.cs
```diff
@@ -0,0 +1,58 @@
...
+    {
+        var metric = new TraceMetric([CreateMetric()]);
+
+        var items = metric.Items;
+
+        Assert.Equal(1, items.Length);
+    }
+
+    [Fact]
+    public void WriteTo_Empty_AsJson()
```
jamescrosswell, line 45: Potentially a good candidate for a Verify test...

jamescrosswell (commented):
Nice! Looking heaps better @Flash0ver - way easier for SDK users to use this (particuliary the options callbacks). 

I left a few comments but once we've worked through those, I think we'll be able to merge 👍🏻

src/Sentry/SentryOptions.cs
```diff
@@ -1905,4 +1907,53 @@ internal static List<StringOrRegex> GetDefaultInAppExclude() =>
...
+    /// </remarks>
+    public class ExperimentalSentryOptions
+    {
+        private readonly SentryOptions _options;
+        private Func<SentryMetric, SentryMetric?>? _beforeSendMetric;
+
+        internal ExperimentalSentryOptions(SentryOptions options)
+        {
+            _options = options;
+        }
```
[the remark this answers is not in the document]
Flash0ver: Redundant indeed ... but not anticipating future needs ... instead forgot to clean up no longer required needs.
From: https://github.com/getsentry/sentry-dotnet/pull/4834/changes/4a080e74243a78a0ca6b5c77080ea9e8910de0ef

Removed via https://github.com/getsentry/sentry-dotnet/pull/4834/changes/afb011bf931b3e500c9aa1e73e117f6d6c367987

test/Sentry.Tests/Protocol/TraceMetricTests.cs
Flash0ver: This test is in the same shape as it's equivalent for Structured-Logs.
If it's fine for you, I'd like to fix both tests in an immediate follow-up.

src/Sentry/SentryUnits.cs
Flash0ver: Fair point!

Context:
https://develop.sentry.dev/sdk/telemetry/metrics/#method-signatures
> `unit` **String, optional**: The unit of measurement (only used for `distribution` and `gauge`). SDKs must not restrict what can be sent in (e.g. by using an `enum`) but should offer constants or similar that help customers send in [units we support](https://develop.sentry.dev/sdk/telemetry/attributes/#units).

I can think of three options:
1. we offer a `MeasurementUnit` overload for the three methods groups, alongside the `string`-based `unit`
   - I asked internally, and this would be alright
   - in this case, we should also mark the `string`-based overload as `deprecated` with a note that this is for future compatibility when Relay/Sentry allows custom units
2. we make `MeasurementUnit` implicitly convertible to `string`
   - doesn't work, as we would need to implicitly convert an `enum` to a string, which C# currently can't do
3. an extension-method on each `MeasurementUnit-enum`
   - ```CSharp
     public readonly partial struct MeasurementUnit
     {
       public enum Duration { Nanosecond, Microsecond, Millisecond, Second, ...}
     }
     public static class MeasurementUnitConversions
     {
       extension(MeasurementUnit.Duration duration)
       {
          public string ToUnitString() // naming is hard
          {
             MeasurementUnit unit = duration;
             return unit.ToString();
          }
       }
     }

     // caller
     SentrySdk.Metrics.EmitDistribution("name", 1.2, MeasurementUnit.Duration.Second.ToUnitString());
     // not ideal ... as just `ToString` would not be recommended
     SentrySdk.Metrics.EmitDistribution("name", 1.2, MeasurementUnit.Duration.Second.ToString());
     ```

And I got a follow-up question:
When an invalid / out-of-range enum-value is used (e.g. `(MeasurementUnit.Duration)(-1)`) should we
1. use `"unit": "none"` instead
2. set the unit to `null` and with that omit it from the sent Envelope-Item

I prefer the overload and using `null` for an invalid enum.
I do find it a bit weird to use `[Obsolete("For future compatibility, when Sentry/Relay support custom units.")]` ... but it is my current preference over the options I could think of as it is the most descriptive as to why we have a `string` overload in the first place.

In all cases, we would remove the new static/const string-container type again.

What do you think?

src/Sentry/SentryMetric.Factory.cs
Flash0ver: Can do!

Alternatively, but unfortunately, Rider does not support Nesting Files based on Wildcards: https://www.jetbrains.com/help/rider/File_Nesting_Dialog.html

What we could do is apply manual nesting to all projects:
```xml
  <ItemGroup>
    <Compile Update="SentryMetric.Factory.cs;SentryMetric.Generic.cs">
      <DependentUpon>SentryMetric.cs</DependentUpon>
    </Compile>
    <Compile Update="SentryTraceMetrics.Counter.cs;SentryTraceMetrics.Distribution.cs;SentryTraceMetrics.Gauge.cs">
      <DependentUpon>SentryTraceMetrics.cs</DependentUpon>
    </Compile>
  </ItemGroup>
```

image
image

If you like this file-nesting approach to solve this issue for the new and also existing files as well,
I'd like to do this in a separate follow-up PR.

What do you think?

src/Sentry/SentryMetric.cs
Flash0ver: Good point.
I am linking this comment to the mentioned issue (#4882) ... and actually removed this method from this PR.
Because there is another follow-up to also apply Release, Environment, and Contexts like User-ID from the Scope.
We currently don't carry all the necessary information on the Scope for each SDK ... but again, a follow-up for later.

Removed via https://github.com/getsentry/sentry-dotnet/pull/4834/changes/b0786216f5e5c09bb0d1c9af1921ae0bc0c00322

src/Sentry/SentryLog.cs
Flash0ver: It's unrelated indeed.
I removed it from this PR.
Removed via https://github.com/getsentry/sentry-dotnet/pull/4834/changes/b0786216f5e5c09bb0d1c9af1921ae0bc0c00322

src/Sentry/Protocol/StructuredLog.cs
Flash0ver: Somewhat necessary:
In this PR, I changed the `BatchProcessor` from being non-generic to generic, so that it can be used for both Structured-Logging and Trace-connected Metrics.
Apart from the type that is batched, we also create an Envelope with all batched Envelope-Items differently.
So I decided to make the "capture-to-envelope" action a delegate for the `BatchProcessor<T>`,
which is then passing this method via:
https://github.com/getsentry/sentry-dotnet/pull/4834/changes#diff-8cc2476d3d8956d3bf1bf9ff4716d90be9765269631258122bca0d731acfc3a5

As an alternative,
we could make the now generic `BatchProcessor<T>` abstract,
and then derive both a
- `StructuredLogBatchProcessor`
- `TraceMetricBatchProcessor`

from it, and move this `Capture` method into the respectively derived concrete (non-generic) type.

What do you think?

src/Sentry/Internal/Hub.cs
Flash0ver: My original motivation was to have the naming scheme derive from what we used for logs:

| public Before-Send-Type | SDK-API from spec | public Sentry-SDK Type | internal Protocol-Envelope-Type (multiple items) |
| --- | --- | --- | --- |
| SentryLog | `Sentry.logger.` | SentryStructuredLogger | StructuredLog |
| SentryMetric | `Sentry.metrics.` | SentryTraceMetrics | TraceMetric |

My motivation to go here with a name that ends in `...Metrics`, because the SDK-API spec suggests `Sentry.metrics.X`,
in order to be equivalent with the `SentryStructuredLogger` (ending in `...Logger`) where the SDK-API spec suggests `Sentry.logger.X`.

On the other hand ... there is indeed an argument to be made that a plural name, such as `Metrics`, is not a great .NET type name ... but more the name of a specific collection.

Also: https://learn.microsoft.com/dotnet/standard/design-guidelines/names-of-classes-structs-and-interfaces
> ✔️ DO name classes and structs with nouns or noun phrases, using PascalCasing.

`Metrics` ... to be fair ... is either a collection-name ... or a namespace name.

I'm convinced with renaming.
Considering that the `SentryStructuredLogger` has all public methods starting with `Log`,
I prefer `SentryMetricEmitter`, because all it's public methods start with `Emit`.

Flash0ver: todo: rename public Metrics type to `SentryMetricEmitter`
todo: likewise, rename all derived types
- `DefaultSentryTraceMetrics` -> `DefaultSentryMetricEmitter`
- `DisabledSentryTraceMetrics` -> `DisabledSentryMetricEmitter`
- `InMemorySentryTraceMetrics` -> `InMemorySentryMetricEmitter`

src/Sentry/SentryMetric.Factory.cs
jamescrosswell: Yeah the file nesting would help a lot... keep things more compact in the IDE and also make it really visually obvious which files are essentially just different parts of the same class.

I'm not a big fan of partial classes in general, except where they're used with source generators... but I appreciate that's just my opinion so OK to be flexible on this if we can find a better way to organise all the extra files they create.

src/Sentry/Protocol/StructuredLog.cs
jamescrosswell: Ah I see. I think I would prefer making `BatchProcessor<T>` abstract. It means the 'delegate' is implied by the generic type parameter, which simplifies the BatchProcessor constructor definition and simplifies instantiation of the BatchProcessor for calling code... unless there's some weird edge case where we'd want to pass in another delegate but it doesn't look like we're doing that anywhere (even in tests).

src/Sentry/SentryUnits.cs
jamescrosswell: Hm, I see. I think whatever we do, we don't want extra complexity for SDK users - so I think that eliminates any public extension method. 

We could go with an overload (string or enum) and, in the implementation for the enum variant, use an internal version of the extension method. That way SDK users get the apparent simplicity of enums but under the hood we conform to Sentry's protocol.

And yeah, if someone gives an invalid enum value (edge case I think) "none" seems fine - presumably we log something as well.

src/Sentry/SentryMetric.cs
```diff
@@ -0,0 +1,376 @@
...
+        if (attributes.TryGetNonEnumeratedCount(out var count))
+        {
+            _ = _attributes.EnsureCapacity(_attributes.Count + count);
+        }
+#endif
+
+        foreach (var attribute in attributes)
+        {
+            _attributes[attribute.Key] = new SentryAttribute(attribute.Value);
+        }
```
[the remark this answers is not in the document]
Flash0ver: In case that the value is `null`, we log a diagnostic warning later in the flow (when serializing to the Sentry Protocol), because the user could still provide a non-null value in the `SetBeforeSendMetric` callback.
see: `src/Sentry/Protocol/SentryAttribute.cs`

src/Sentry/SentryMetricEmitter.Gauge.cs
Flash0ver: Added via https://github.com/getsentry/sentry-dotnet/pull/4834/changes/7faa68261870245cef1067853666c942cd9b57f3

src/Sentry/Internal/BatchProcessor.cs
```diff
@@ -61,21 +62,21 @@ internal void Enqueue(SentryLog log)
...
         var activeBuffer = _activeBuffer;
 
-        if (!TryEnqueue(activeBuffer, log))
+        if (!TryEnqueue(activeBuffer, item))
         {
             activeBuffer = ReferenceEquals(activeBuffer, _buffer1) ? _buffer2 : _buffer1;
-            if (!TryEnqueue(activeBuffer, log))
+            if (!TryEnqueue(activeBuffer, item))
             {
                 _clientReportRecorder.RecordDiscardedEvent(DiscardReason.Backpressure, DataCategory.Default, 1);
```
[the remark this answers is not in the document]
Flash0ver: But it's `"{0}"` twice, so both filled by `arg0` by the underlying `System.String.Format()`

src/Sentry/SentryOptions.cs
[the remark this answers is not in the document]
Flash0ver: Not sure if I understand correctly.
The `EnableMetrics` flag essentially instantiates the "default"-implementation when `true`, and an instance of the "no-op"-implementation when `false`.

The experimental errors only show when any of the Experimental APIs are actively used.
So just by upgrading, regardless of `EnableMetrics` being `true` or `false`,
no experimental errors will show up, unless the user starts invoking the new methods actively.

Additionally, I added less `[Experimental]` Attributes this time around - compared to Logs:
I only added the `[Experimental]` Attributes on `IHub.Metrics` property (and the properties of the implementing classes respectively: `HubAdapter.Metrics` and `DisabledHub.Metrics`).
I omitted the `Experimental`-Attributes on `SentrySdk.Experimental`, `SentrySdk.Experimental.Metrics`, as well as `SentryOptions.Experimental` and `SentryOptions.Experimental.EnableMetrics` and `SentryOptions.Experimental.SetBeforeSendMetric`.

In other words:
- no `[ExperimentalAttribute]` on APIs where we have an `Experimental` property / nested type in-between
- `[ExperimentalAttribute]` are only on Hub-APIs, where we don't have an `Experimental` property in-between

```CSharp
// no errors
options.Experimental.EnableMetrics = false;
options.Experimental.SetBeforeSendMetric(static metric => metric);

SentrySdk.Experimental.Metrics.EmitCounter("name", 1);

var emitter = SentrySdk.Experimental.Metrics;
emitter.EmitCounter("name", 1.1);

// error: [SENTRYTRACECONNECTEDMETRICS] 'Sentry.Extensibility.HubAdapter.Metrics' is for evaluation purposes only and is subject to change or removal in future updates. Suppress this diagnostic to proceed. (https://github.com/getsentry/sentry-dotnet/discussions/4838)
HubAdapter.Instance.Metrics.EmitDistribution("name", 1);

// error: [SENTRYTRACECONNECTEDMETRICS] 'Sentry.Extensibility.HubAdapter.Metrics' is for evaluation purposes only and is subject to change or removal in future updates. Suppress this diagnostic to proceed. (https://github.com/getsentry/sentry-dotnet/discussions/4838)
var emitter = HubAdapter.Instance.Metrics;

IHub hub = HubAdapter.Instance;
_ = hub.Metrics; // error: [SENTRYTRACECONNECTEDMETRICS] 'Sentry.IHub.Metrics' is for evaluation purposes only and is subject to change or removal in future updates. Suppress this diagnostic to proceed. (https://github.com/getsentry/sentry-dotnet/discussions/4838)
```

Also ... semi-related ... even without Warnings-As-Errors, I believe, `[Experimental]`-based diagnostics are error-level (not warnings) and must be actively suppressed.

src/Sentry/SentryMetric.Factory.cs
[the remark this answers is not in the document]
Flash0ver: Too much copy-paste-rename from Logs 😉.
De-duplicated via https://github.com/getsentry/sentry-dotnet/pull/4834/changes/a9e4e9abffc34a81d6fe94ff20579807a0fb4ae4

src/Sentry/SentryMetric.cs
```diff
@@ -0,0 +1,376 @@
...
+
+        if (sdk.Name is { } name)
+        {
+            SetAttribute("sentry.sdk.name", name);
+        }
+        if (sdk.Version is { } version)
+        {
+            SetAttribute("sentry.sdk.version", version);
+        }
+    }
```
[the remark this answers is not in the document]
Flash0ver: This will change with #4733.

src/Sentry/SentryMetric.cs
```diff
@@ -0,0 +1,376 @@
...
+
+        foreach (var attribute in attributes)
+        {
+            _attributes[attribute.Key] = new SentryAttribute(attribute.Value);
+        }
+    }
+
+    internal void Apply(Scope? scope)
+    {
+    }
```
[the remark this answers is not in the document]
Flash0ver: Removed via https://github.com/getsentry/sentry-dotnet/pull/4834/changes/b0786216f5e5c09bb0d1c9af1921ae0bc0c00322

src/Sentry/SentryOptions.cs
```diff
@@ -1905,4 +1907,53 @@ internal static List<StringOrRegex> GetDefaultInAppExclude() =>
...
+    /// Experimental features are subject to binary, source and behavioral breaking changes in future updates.
+    /// </remarks>
+    public class ExperimentalSentryOptions
+    {
+        private readonly SentryOptions _options;
+        private Func<SentryMetric, SentryMetric?>? _beforeSendMetric;
+
+        internal ExperimentalSentryOptions(SentryOptions options)
+        {
+            _options = options;
```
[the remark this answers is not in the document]
Flash0ver: Removed via https://github.com/getsentry/sentry-dotnet/pull/4834/changes/afb011bf931b3e500c9aa1e73e117f6d6c367987

src/Sentry/Internal/Hub.cs
Flash0ver: Renamed via https://github.com/getsentry/sentry-dotnet/pull/4834/changes/be699bfbce9432956784ae880ad68862a22c6b78

src/Sentry/Protocol/StructuredLog.cs
Flash0ver: The delegate was the same for all cases indeed.

via https://github.com/getsentry/sentry-dotnet/pull/4834/changes/98e991947222b58a0c18014d8605620a765403ae
- made now generic `BatchProcessor` abstract
- derived a constructed `SentryLogBatchProcessor`
- derived a constructed `SentryMetricBatchProcessor`
- moved the code of the extra method into the derived types accordingly

src/Sentry/SentryOptions.cs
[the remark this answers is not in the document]
jamescrosswell: If it won't impact people until/unless they use the APIs then all good 👍🏻

src/Sentry/SentryMetric.cs
```diff
@@ -0,0 +1,317 @@
...
+
+    internal void SetAttribute(string key, char value)
+    {
+        _attributes[key] = new SentryAttribute(value.ToString(), "string");
+    }
+
+    internal void SetAttribute(string key, int value)
+    {
+        _attributes[key] = new SentryAttribute(value, "integer");
+    }
```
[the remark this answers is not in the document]
Flash0ver: Currently symmetrical to Logs.
Will be followed-up during #4882.

src/Sentry/SentryMetric.Factory.cs
```diff
@@ -0,0 +1,65 @@
...
+            Unit = unit,
+        };
+
+        scope ??= hub.GetScope();
+        metric.SetDefaultAttributes(options, scope?.Sdk ?? SdkVersion.Instance);
+
+        metric.SetAttributes(attributes);
+
+        return metric;
+    }
```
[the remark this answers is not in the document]
Flash0ver: De-duplicated via https://github.com/getsentry/sentry-dotnet/pull/4834/changes/e56b0a46b6c780bf9d134cae9da2f3bf5bca36e3

src/Sentry/SentryUnits.cs
Flash0ver: replaced new `SentryUnits` with existing `MeasurementUnit`: https://github.com/getsentry/sentry-dotnet/pull/4834/changes/0dc98384ed5a55b67d778d9023246c2f6a6c6ad6

Also updated docs: https://github.com/getsentry/sentry-docs/pull/16054/changes/8279c739f2c6ceddcdcda1159f24882ec49acb23

src/Sentry/SentryMetric.Factory.cs
Flash0ver: Follow-up: getsentry/sentry-dotnet#4894

test/Sentry.Tests/SentryMetricEmitterTests.Types.cs
```diff
@@ -287,8 +287,72 @@ public void Emit_Name_Empty_DoesNotCaptureEnvelope(SentryMetricType type, string
...
         entry.Args.Should().BeEquivalentTo<object>([arg0, arg1]);
     }
+
+    [Fact]
+    public void Type_EmitMethods_StringUnitParameterIsObsoleteForForwardCompatibility()
+    {
+        var type = typeof(SentryMetricEmitter);
+
+        type.Methods()
+            .Where(static method => method.IsPublic && method.ReturnType == typeof(void) && method.IsGenericMethod && method.Name.StartsWith("Emit"))
```
jamescrosswell: Is there any benefit to using static in the tests btw? Presumably this test will be compiled and run once for every test run... so this doesn't save us any allocations right? It's certainly not a hot path.

src/Sentry/SentryOptions.cs
Flash0ver: the public `SentryMetric` type, as well as the `SetBeforeSendMetric` method, are now non-generic, but there is a generic `TryGetValue` method on the instance to extract the value

via https://github.com/getsentry/sentry-dotnet/pull/4834/changes/4a080e74243a78a0ca6b5c77080ea9e8910de0ef

test/Sentry.Tests/SentryMetricEmitterTests.Types.cs
Flash0ver: In this case, no: there is no conflicting identifier that might get captured.
As a matter of fact, since no symbols are captured, the compiler behaves the same as without the `static` keyword here.
And even if ... it is indeed "just" a test in this case.

It's just a force of habit for me.

To be fair, the `static` not having an effect here, I have indeed now removed it, to make the code a bit shorter and the test easier to read.

Via https://github.com/getsentry/sentry-dotnet/pull/4834/changes/0798441d813c5c11c7ba104511657531741aaa58
(but I made the method name a bit longer to be slightly more descriptive 😉)

jamescrosswell:
@Flash0ver is there anything outstanding on this PR now or are we good to go?

Flash0ver:
> @Flash0ver is there anything outstanding on this PR now or are we good to go?

@jamescrosswell, not from my perspective, considering the follow-ups #4882 and #4894.

jamescrosswell (approved):
Nice! I think we're good to go 😃

PR #4840, opened 2026-01-15
feat(metrics): Trace-connected Metrics (Analyzers)

Flash0ver:
based on (but to be merged to `main` individually)
- ~#4834~
- #5023

### Summary

My little weekend-project:
Add a Diagnostic-Analyzer to our Compiler-Extensions, that warn users when a metric would not be emitted due to an unsupported numeric value type -or- likewise, a `SentryMetric.TryGetValue<>` invocation with an unsupported numeric value type is detected.

### Changelog Entry

Report a new _Diagnostic_ (`SENTRY1001`) when a Metrics-API is invoked with an unsupported numeric type

### Remarks

For the numeric type of a Metric, we currently allow 64-bit sized integral (signed) and floating-point numbers.
That means that e.g. `ulong`, `System.Int128`, or `decimal` are currently not supported by Sentry.
The compile-time constraint of `Sentry.SentryMetric<T>` only constrains to non-nullable value types.
Alternatively, we could have implemented respectively types overloads for the method groups:
```CSharp
public void EmitCounter(string name, long value);
public void EmitCounter(string name, float value);
public void EmitCounter(string name, double value);
public void EmitCounter(string name, long value, Scope? scope)
public void EmitCounter(string name, float value, Scope? scope)
public void EmitCounter(string name, double value, Scope? scope)
public void EmitCounter ..

public void EmitGauge ..
public void EmitDistribution ..
```
To avoid this explosion of overloads per method group,
and be similar to the implementation of [System.Diagnostics.Metrics](https://learn.microsoft.com/dotnet/core/diagnostics/metrics-instrumentation),
we are not compile-time constraining unsupported types,
but are instead run-time constraining unsupported types (no-op and Debug-Diagnostic-Logging).

To still warn users unfamiliar with the [System.Diagnostics.Metrics.Meter](https://learn.microsoft.com/dotnet/api/system.diagnostics.metrics.meter)-based APIs,
I built an Analyzer over some weekends to guide new users.

### Examples

```CSharp
options.SetBeforeSendMetric(static SentryMetric? (SentryMetric metric) =>
{
    if (metric.TryGetValue<int>(out int int32) && int32 < 0)
    {
        return null;
    }

    if (metric.TryGetValue<uint>(out uint uint32) && uint32 == 0) //SENTRY1001
    {
        return null;
    }

    return metric;
});

SentrySdk.Metrics.EmitCounter("my.counter", 1);
SentrySdk.Metrics.EmitCounter("my.counter", 1m); //SENTRY1001
SentrySdk.Metrics.EmitCounter("my.counter", StringComparison.CurrentCultureIgnoreCase); //SENTRY1001
SentrySdk.Metrics.EmitCounter<ulong>("my.counter", default); //SENTRY1001
```


src/Sentry.Compiler.Extensions/AnalyzerReleases.Unshipped.md
```diff
@@ -0,0 +1,8 @@
+﻿; Unshipped analyzer release
+; https://github.com/dotnet/roslyn-analyzers/blob/main/src/Microsoft.CodeAnalysis.Analyzers/ReleaseTrackingAnalyzers.Help.md
+
+### New Rules
+
+Rule ID | Category | Severity | Notes
+--------|----------|----------|-------
+SENTRY1001 | Support | Warning | TraceConnectedMetricsAnalyzer
```
Flash0ver: question: ID

I'm a a bit certain about the ID.
Also, I'm not sure if it should be two separate IDs, one for the `Emit*` methods, and one for `SentryMetric.TryGetValue<TValue>(out TValue value)` method.

src/Sentry.Compiler.Extensions/Sentry.Compiler.Extensions.csproj
```diff
@@ -15,11 +15,36 @@
...
-    <PackageReference Include="Microsoft.CodeAnalysis.CSharp" Version="4.3.1" PrivateAssets="all"/>
+    <PackageReference Include="Microsoft.CodeAnalysis.CSharp" Version="4.3.1" PrivateAssets="all" />
   </ItemGroup>
 
+  <!--
+    We use Simon Cropp's Polyfill source-only package to access APIs in lower targets.
+    https://github.com/SimonCropp/Polyfill
+  -->
+  <ItemGroup>
+    <PackageReference Include="Polyfill" Version="9.8.1" PrivateAssets="all" />
```
Flash0ver: note: see also #4879

src/Sentry.Compiler.Extensions/Analyzers/TraceConnectedMetricsAnalyzer.cs
```diff
@@ -0,0 +1,113 @@
...
+
+    private static readonly DiagnosticDescriptor Rule = new(
+        id: DiagnosticIds.Sentry1001,
+        title: Title,
+        messageFormat: MessageFormat,
+        category: DiagnosticCategories.Support,
+        defaultSeverity: DiagnosticSeverity.Warning,
+        isEnabledByDefault: true,
+        description: Description,
+        helpLinkUri: null
```
Flash0ver, line 25: question: docs

Should we document the Analyzer?
If so, we could provide a link to the documentation page here.

src/Sentry.Compiler.Extensions/Analyzers/TraceConnectedMetricsAnalyzer.cs
```diff
@@ -0,0 +1,113 @@
...
+/// </summary>
+[DiagnosticAnalyzer(LanguageNames.CSharp)]
+public sealed class TraceConnectedMetricsAnalyzer : DiagnosticAnalyzer
+{
+    private static readonly string Title = "Unsupported numeric type of Metric";
+    private static readonly string MessageFormat = "{0} is unsupported type for Sentry Metrics. The only supported types are byte, short, int, long, float, and double.";
+    private static readonly string Description = "Integers should be a 64-bit signed integer, while doubles should be a 64-bit floating point number.";
+
+    private static readonly DiagnosticDescriptor Rule = new(
+        id: DiagnosticIds.Sentry1001,
```
Flash0ver, line 18: question: CHANGELOG

Should we mention this Analyzer, and Analyzer in general, in the CHANGELOG?
If so, in the `Features` category, or a separate category?

src/Sentry.Compiler.Extensions/Analyzers/TraceConnectedMetricsAnalyzer.cs
```diff
@@ -0,0 +1,113 @@
...
+using Microsoft.CodeAnalysis.Diagnostics;
+using Microsoft.CodeAnalysis.Operations;
+
+namespace Sentry.Compiler.Extensions.Analyzers;
+
+/// <summary>
+/// Guide consumers to use the public API of <see href="https://develop.sentry.dev/sdk/telemetry/metrics/">Sentry Trace-connected Metrics</see> correctly.
+/// </summary>
+[DiagnosticAnalyzer(LanguageNames.CSharp)]
+public sealed class TraceConnectedMetricsAnalyzer : DiagnosticAnalyzer
```
Flash0ver, line 11: question: do we want to do this in general?

This started off as a weekend-project for me.
Do we want to provide this Analyzer in the first place?

src/Sentry.Compiler.Extensions/Analyzers/TraceConnectedMetricsAnalyzer.cs
jamescrosswell: I think the features section... they're basically there to improve the developer experience for SDK users right?

src/Sentry.Compiler.Extensions/Analyzers/TraceConnectedMetricsAnalyzer.cs
jamescrosswell: > To avoid this explosion of overloads per method group,
and be similar to the implementation of [System.Diagnostics.Metrics](https://learn.microsoft.com/dotnet/core/diagnostics/metrics-instrumentation),
we are not compile-time constraining unsupported types,
but are instead run-time constraining unsupported types (no-op and Debug-Diagnostic-Logging).

Maybe not... we're trying to create a compile time constraint here (using analysers) because we don't want to create a compile time constraint using the typing system???

It might be a better experience for SDK users and less effort for us to maintain, simply to add some overloads that convert the other numeric types automatically to supported types.

src/Sentry.Compiler.Extensions/DiagnosticCategories.cs
```diff
@@ -0,0 +1,6 @@
+namespace Sentry.Compiler.Extensions;
+
+internal static class DiagnosticCategories
+{
+    internal const string Sentry = nameof(Sentry);
```
Flash0ver, line 5: question: Category

I'm a bit uncertain about the Category.

The Category can be used to e.g. configure all Diagnostic of a category, rather than or in addition to configuring specific diagnostics per ID,
e.g. via an `.globalconfig` file
```
is_global = true

# Configure this new rule only
dotnet_diagnostic.SENTRY1001.severity = none

# Configure all rules within the "Sentry" Category
dotnet_analyzer_diagnostic.category-Sentry.severity = none
```

The ideas / variants I am having are
- a single Category for all of Sentry's Diagnostic
- have a Category per Feature set
  - e.g.
    - Traces
    - Logs
    - Metrics
- re-use the same Categories that the Analyzers of the .NET SDK include
  - see https://learn.microsoft.com/dotnet/fundamentals/code-analysis/categories

src/Sentry.Compiler.Extensions/Analyzers/TraceConnectedMetricsAnalyzer.cs
Flash0ver: Will do after the merge / after releasing it via #4962.

src/Sentry.Compiler.Extensions/Analyzers/TraceConnectedMetricsAnalyzer.cs
Flash0ver: added CHANGELOG entry via https://github.com/getsentry/sentry-dotnet/pull/4840/changes/6c5ca237fe3921a32cbebf4681733a822eb9a3b2

src/Sentry.Compiler.Extensions/Analyzers/TraceConnectedMetricsAnalyzer.cs
Flash0ver: We drafted an alternative API shape, with explicit overloads.

But the devil is in the "API explosion":
```diff
-public void EmitCounter<T>(string name, T value) where T : struct
+public void EmitCounter(string name, int value)
+public void EmitCounter(string name, long value)
+// for all EmitCounter overloads
+// for EmitGauge and EmitDistribution method groups
```

And a follow-up uncertainty is about the Before-Send-Callback:
```CSharp
options.SetBeforeSendMetric(static metric =>
{
    if (metric.TryGetValue(out int integer)) // if not emitted as `Int32`, do we want to cast to `Int32`?
    if (metric.TryGetInt32(out int integer)) // similar "explosion" of methods

    return metric;
});
```

So we decided to keep the API shape as is and actually do go with the Analyzer.
Should we have made a mistake, we can change both the API and the Analyzer for the next major version.

src/Sentry.Compiler.Extensions/Sentry.Compiler.Extensions.csproj
```diff
@@ -15,11 +15,25 @@
...
-    <PackageReference Include="Microsoft.CodeAnalysis.CSharp" Version="4.3.1" PrivateAssets="all"/>
+    <PackageReference Include="Microsoft.CodeAnalysis.CSharp" Version="4.3.1" PrivateAssets="all" />
+  </ItemGroup>
+
+  <!--
+    We use Simon Cropp's Polyfill source-only package to access APIs in lower targets.
+    https://github.com/SimonCropp/Polyfill
+  -->
+  <ItemGroup>
+    <PackageReference Include="Polyfill" Version="1.33.2" PrivateAssets="all" />
```
Flash0ver, line 26: note: to be in sync with #5007

src/Sentry.Compiler.Extensions/DiagnosticCategories.cs
Flash0ver: I am currently leaning towards a `Sentry` category (as currently proposed),
as I don't believe we are going to have a significant number of Diagnostics and/or Diagnostics in different Categories anytime soon,
so that users have an easy opt-out to all of our Analyzers if there are any issues.

(to be exemplified / documented via #4962)

jamescrosswell: I'm OK with that. Let's give it a go and see if we want to tweak later.

src/Sentry.Compiler.Extensions/AnalyzerReleases.Unshipped.md
Flash0ver: going with a single ID ... as both flagged invocations are essentially no-op

PR #4857, opened 2026-01-21
chore: upgrade to C# 14.0

Flash0ver:
Upgrade to C# 14.0

See [Language Feature Status](https://github.com/dotnet/roslyn/blob/main/docs/Language%20Feature%20Status.md).

Also a prerequisite to upgrade to the latest version of `Polyfill`, after SimonCropp/Polyfill#454.

### DIFF
--- a/src/Sentry.Android.AssemblyReader/ELFSharp/UImage/UImage.cs
+++ b/src/Sentry.Android.AssemblyReader/ELFSharp/UImage/UImage.cs
@@ -35,6 +35,7 @@ namespace ELFSharp.UImage
             Type = (ImageType)reader.ReadByte();
             Compression = (CompressionType)reader.ReadByte();
             var nameAsBytes = reader.ReadBytes(32);
+            // See: https://learn.microsoft.com/dotnet/csharp/whats-new/breaking-changes/compiler%20breaking%20changes%20-%20dotnet%2010#enumerablereverse
             Name = Encoding.UTF8.GetString(Enumerable.Reverse(nameAsBytes).SkipWhile(x => x == 0).Reverse().ToArray());
 
             if (multiFileImage)


src/Sentry.Android.AssemblyReader/ELFSharp/UImage/UImage.cs
```diff
@@ -35,7 +35,7 @@ internal UImage(Stream stream, bool multiFileImage, bool ownsStream)
             Type = (ImageType)reader.ReadByte();
             Compression = (CompressionType)reader.ReadByte();
             var nameAsBytes = reader.ReadBytes(32);
-            Name = Encoding.UTF8.GetString(nameAsBytes.Reverse().SkipWhile(x => x == 0).Reverse().ToArray());
+            Name = Encoding.UTF8.GetString(Enumerable.Reverse(nameAsBytes).SkipWhile(x => x == 0).Reverse().ToArray());
```
Flash0ver, line 39: note: other method bound with [[Proposal]: First-Class Span Types](https://github.com/dotnet/csharplang/issues/7905)

```CSharp
byte[] bytes = new byte[0];
_ = bytes.Reverse();
```
now prefers `public static void Reverse<T>(this Span<T> span)` in `System.MemoryExtensions` (if available),
rather than `public static IEnumerable<TSource> Reverse<TSource>(this TSource[] source)` in `System.Linq.Enumerable`.
Invoking the LINQ extension method explicitly instead.

See https://learn.microsoft.com/dotnet/csharp/whats-new/breaking-changes/compiler%20breaking%20changes%20-%20dotnet%2010#enumerablereverse

test/Sentry.AspNetCore.Grpc.Tests/GrpcRequestHandler.cs
```diff
@@ -9,12 +9,10 @@ public class GrpcRequestHandler<TRequest, TResponse>
 {
     public MethodDescriptor Method { get; set; }
 
-    private Func<TRequest, ServerCallContext, Task<TResponse>> _handler;
-
     public Func<TRequest, ServerCallContext, Task<TResponse>> Handler
     {
-        get => _handler ?? ((_, _) => Task.FromResult<TResponse>(null));
-        set => _handler = value;
+        get => field ?? ((_, _) => Task.FromResult<TResponse>(null));
```
Flash0ver: note: only applied to Test projects

I'm not sure yet whether I like the "`field` keyword in properties" for all cases yet.

jamescrosswell: I think if we can get rid of some more boiler plate it's all good (for fields that are only ever referenced from the property getter/setter).

I wouldn't bother spending time changing all the existing code... but we can use this for new code and refactor other stuff as/when we're changing it.

Flash0ver: I reverted this change, so that we only update to C# 14 here, and fix the breaking change.

src/Sentry.Android.AssemblyReader/ELFSharp/UImage/UImage.cs
jamescrosswell, line 39:
```suggestion
            // See: https://learn.microsoft.com/dotnet/csharp/whats-new/breaking-changes/compiler%20breaking%20changes%20-%20dotnet%2010#enumerablereverse
            Name = Encoding.UTF8.GetString(Enumerable.Reverse(nameAsBytes).SkipWhile(x => x == 0).Reverse().ToArray());
```

That's quite obscure... let's put a comment in there to explain the change.

Flash0ver: Good point ... thanks!

Issue #4192, opened 2025-05-16
Android Session Replay - Add network calls


bruno-garcia:
[example replay from demo org](https://demo.sentry.io/explore/replays/bb99effbfa654b62a856bdd88639b19f/?fov=0%2C1490000&project=4509153190739968&query=&referrer=%2Fexplore%2Freplays%2F&statsPeriod=7d&t_main=network&yAxis=count%28%29)

Similar to React Native and Flutter, the web requests added to Replay are done at the outer layer.
Unless somehow HttpClient requrests get captured through our native SDKs but doesn't seem to be the case:

The network calls are not added:
![Image](https://github.com/user-attachments/assets/bf0a5300-6982-48f1-b663-19a8022fd140)

But looking at the trace, we can see requests were made:
![Image](https://github.com/user-attachments/assets/97e982c0-93ab-4560-90d8-878f7d12e3e2)

jamescrosswell:
The Network tab of the replay details doesn't actually show Transactions or Spans... It shows something similar called a [performanceSpan](https://develop.sentry.dev/sdk/data-model/event-payloads/replay-recording/#performancespan) (which is a kind of custom replay event). 
- [RN implementation](https://github.com/getsentry/sentry-react-native/blob/a06e222554643898a91fcafdbbada4dcb6a23b3a/packages/core/android/src/main/java/io/sentry/react/RNSentryReplayBreadcrumbConverter.java#L128-L168) for reference
- If relying on the native breadcrumb converters, we need [these fields for Android](https://github.com/getsentry/sentry-java/blob/94bff8dc0a952ad8c1b6815a9eda5005e41b92c7/sentry-android-replay/src/main/java/io/sentry/android/replay/DefaultReplayBreadcrumbConverter.kt#L97-L104)
- Maybe [slightly different fields for iOS](https://github.com/getsentry/sentry-cocoa/blob/2b4e787e55558e1475eda8f98b02c19a0d511741/Sources/Swift/Integrations/SessionReplay/SentrySRDefaultBreadcrumbConverter.swift#L70-L86)

However these would be captured in the same way that network spans are currently captured... 

Could they be captured independently if performance isn't enabled?

PR #4860, opened 2026-01-22
feat: Network calls for Session Replay on Android

jamescrosswell:
Resolves https://github.com/getsentry/sentry-dotnet/issues/4192:
- https://github.com/getsentry/sentry-dotnet/issues/4192

This PR ensures [performanceSpans](https://develop.sentry.dev/sdk/data-model/event-payloads/replay-recording/#performancespan) are sent with replay events so that outbound network requests in Maui for Android are displayed correctly in the `Replay -> Network` tab. 

image

### Implementation notes

The performanceSpan itself doesn't include a replyId, so it has to be sent along with the replay event - i.e. it gets sent by the mobile/Android SDK. The way android does this is to inspect the stream of breadcrumbs and convert these into RRWeb custom events. Typicallly this is done via the [DefaultBreadcrumbConverter](https://github.com/getsentry/sentry-java/blob/96eeafa39947db536529ee3c26216f969e6eb599/sentry-android-replay/src/main/java/io/sentry/android/replay/DefaultReplayBreadcrumbConverter.kt#L97-L104) but it's also possible to swap that out with a custom converter (which is [what our ReactNative SDK does](https://github.com/getsentry/sentry-react-native/blob/a06e222554643898a91fcafdbbada4dcb6a23b3a/packages/core/android/src/main/java/io/sentry/react/RNSentryReplayBreadcrumbConverter.java#L128-L168)).

This PR is an attempt to utilise the `DefaultReplayBreadcrumbConverter` by ensuring that our breadcrumbs contain all [the data properties it's expecting](https://github.com/getsentry/sentry-java/blob/94bff8dc0a952ad8c1b6815a9eda5005e41b92c7/sentry-android-replay/src/main/java/io/sentry/android/replay/DefaultReplayBreadcrumbConverter.kt#L97-L104). 

### Additional comments

If we  eventually try to implement this for iOS, [iOS looks for slightly different field names](https://github.com/getsentry/sentry-cocoa/blob/2b4e787e55558e1475eda8f98b02c19a0d511741/Sources/Swift/Integrations/SessionReplay/SentrySRDefaultBreadcrumbConverter.swift#L70-L86)

### DIFF
--- a/src/Sentry/Platforms/Android/DotnetReplayBreadcrumbConverter.cs
+++ b/src/Sentry/Platforms/Android/DotnetReplayBreadcrumbConverter.cs
@@ -10,9 +10,11 @@ internal class DotnetReplayBreadcrumbConverter(Sentry.JavaSdk.SentryOptions opti
 
     public override global::IO.Sentry.Rrweb.RRWebEvent? Convert(Sentry.JavaSdk.Breadcrumb breadcrumb)
     {
-        // The Java converter expects httpStartTimestamp/httpEndTimestamp to be Double or Long.
-        // .NET breadcrumb data is always stored as strings. We convert these to numeric here so that the base.Convert()
-        // method doesn't throw an exception.
+        // The Java SDK automatically converts breadcrumbs for outgoing http requests into performance spans
+        // that show in the Network tab of session replays... however, it expects certain data to be stored in a
+        // specific format in the breadcrumb.data. It needs values for httpStartTimestamp and httpEndTimestamp
+        // stored as Double or Long representations of timestamps (milliseconds since epoch).
+        // .NET breadcrumb data is always stored as strings, so we have to convert these to numeric values here.
         try
         {
             if (breadcrumb is { Category: HttpCategory, Data: { } data })
@@ -32,13 +34,11 @@ internal class DotnetReplayBreadcrumbConverter(Sentry.JavaSdk.SentryOptions opti
 
     private static void NormalizeTimestampField(IDictionary<string, Java.Lang.Object> data, string key)
     {
-        data.TryGetValue(key, out var value);
-        if (value is null or Java.Lang.Long or Java.Lang.Double or Java.Lang.Integer or Java.Lang.Float)
+        if (!data.TryGetValue(key, out var value) || value is Java.Lang.Number)
         {
             return;
         }
 
-        // Note: `data.Get` returns `Java.Lang.Object`, not a .NET `string`.
         var str = (value as Java.Lang.String)?.ToString() ?? value.ToString();
         if (string.IsNullOrWhiteSpace(str))
         {
--- a/src/Sentry/SentryHttpMessageHandler.cs
+++ b/src/Sentry/SentryHttpMessageHandler.cs
@@ -91,6 +91,7 @@ public class SentryHttpMessageHandler : SentryMessageHandler
             {"method", method},
             {"status_code", ((int) response.StatusCode).ToString()}
         };
+#if ANDROID
         if (span is not null)
         {
             // Ensure the breadcrumb can be converted to RRWeb so that it shows up in the network tab in Session Replay.
@@ -98,6 +99,7 @@ public class SentryHttpMessageHandler : SentryMessageHandler
             breadcrumbData[HttpStartTimestampKey] = span.StartTimestamp.ToUnixTimeMilliseconds().ToString("F0", CultureInfo.InvariantCulture);
             breadcrumbData[HttpEndTimestampKey] = DateTimeOffset.UtcNow.ToUnixTimeMilliseconds().ToString("F0", CultureInfo.InvariantCulture);
         }
+#endif
         _hub.AddBreadcrumb(string.Empty, "http", "http", breadcrumbData);
 
         // Create events for failed requests
--- a/test/Sentry.Tests/Platforms/Android/DotnetReplayBreadcrumbConverterTests.cs
+++ b/test/Sentry.Tests/Platforms/Android/DotnetReplayBreadcrumbConverterTests.cs
@@ -29,7 +29,8 @@ public class DotnetReplayBreadcrumbConverterTests
         rrwebEvent.Should().BeOfType<IO.Sentry.Rrweb.RRWebSpanEvent>();
         var rrWebSpanEvent = rrwebEvent as IO.Sentry.Rrweb.RRWebSpanEvent;
         Assert.NotNull(rrWebSpanEvent);
-        // Note the converter divides by 1000 to get ms
+        // Note the converter divides by 1000 to get ms.
+        // See https://github.com/getsentry/sentry-java/blob/94bff8dc0a952ad8c1b6815a9eda5005e41b92c7/sentry-android-replay/src/main/java/io/sentry/android/replay/DefaultReplayBreadcrumbConverter.kt#L215-L228
         rrWebSpanEvent.StartTimestamp.Should().Be(1625079600L);
         rrWebSpanEvent.EndTimestamp.Should().Be(1625079660L);
     }
--- a/test/Sentry.Tests/SentryHttpMessageHandlerTests.cs
+++ b/test/Sentry.Tests/SentryHttpMessageHandlerTests.cs
@@ -611,8 +611,30 @@ public class SentryHttpMessageHandlerTests
         Assert.True(breadcrumbGenerated.Data.ContainsKey(statusKey));
         Assert.Equal(expectedBreadcrumbData[statusKey], breadcrumbGenerated.Data[statusKey]);
     }
+
+    [Fact]
+    public void Send_Executed_FailedRequestsCaptured()
+    {
+        // Arrange
+        var hub = Substitute.For<IHub>();
+        var failedRequestHandler = Substitute.For<ISentryFailedRequestHandler>();
+        var options = new SentryOptions();
+        var url = "https://localhost/";
+
+        using var innerHandler = new FakeHttpMessageHandler();
+        using var sentryHandler = new SentryHttpMessageHandler(hub, options, innerHandler, failedRequestHandler);
+        using var client = new HttpClient(sentryHandler);
+
+        // Act
+        client.Get(url);
+
+        // Assert
+        failedRequestHandler.Received(1).HandleResponse(Arg.Any<HttpResponseMessage>());
+    }
+
 #endif
 
+#if ANDROID
     [Fact]
     public void HandleResponse_SpanExists_AddsReplayBreadcrumbData()
     {
@@ -681,4 +703,5 @@ public class SentryHttpMessageHandlerTests
         breadcrumb.Data!.Should().NotContainKey(SentryHttpMessageHandler.HttpStartTimestampKey);
         breadcrumb.Data.Should().NotContainKey(SentryHttpMessageHandler.HttpEndTimestampKey);
     }
+#endif
 }


src/Sentry.Bindings.Android/Transforms/Metadata.xml
```diff
@@ -162,7 +163,4 @@
 
   <!-- Remove UpdateStatus to avoid CS0108 for conflicting _members fields in nested subclasses -->
   <remove-node path="/api/package[@name='io.sentry']/class[@name='UpdateStatus']" />
-
-  <!-- error CS0050: Inconsistent accessibility: return type 'INetworkBody' is less accessible than method 'NetworkBody.FromXxx' -->
-  <remove-node path="/api/package[@name='io.sentry.util.network']/interface[@name='NetworkBody']" />
```
jamescrosswell: This is not related to this PR... figured I'd fix a warning while I was messing with this file anyway though. Could be done via a separate PR if we really wanted to be strict about things.

jamescrosswell: See:
- https://github.com/getsentry/sentry-dotnet/pull/4874

src/Sentry.Bindings.Android/Sentry.Bindings.Android.csproj
jamescrosswell, line 1: I've moved most of this out to a separate PR as well, since it's just a means to an end to diagnose the problem (not actually part of the change this PR is meant to effect):
- https://github.com/getsentry/sentry-dotnet/pull/4873

samples/Sentry.Samples.Maui/MainPage.xaml.cs
```diff
@@ -1,3 +1,4 @@
+using System.Globalization;
```
Flash0ver: question: required?

Is this `using declaration` required?
I can't spot what has changed.

jamescrosswell: Nah that made it in there by accident...

src/Sentry/Platforms/Android/DotnetReplayBreadcrumbConverter.cs
```diff
@@ -0,0 +1,60 @@
...
+            if (breadcrumb is { Category: HttpCategory, Data: { } data })
+            {
+                NormalizeTimestampField(data, SentryHttpMessageHandler.HttpStartTimestampKey);
+                NormalizeTimestampField(data, SentryHttpMessageHandler.HttpEndTimestampKey);
+            }
+        }
+        catch
+        {
+            // Best-effort: never fail conversion because of parsing issues... we may be parsing breadcrumbs that don't
+            // originate from the .NET SDK.
```
Flash0ver, line 29: question: Diagnostic Log

Something to report as a Diagnostic Log?

src/Sentry/Platforms/Android/DotnetReplayBreadcrumbConverter.cs
```diff
@@ -0,0 +1,60 @@
...
+        }
+
+        return base.Convert(breadcrumb);
+    }
+
+    private static void NormalizeTimestampField(IDictionary<string, Java.Lang.Object> data, string key)
+    {
+        data.TryGetValue(key, out var value);
```
Flash0ver: suggestion: use or discard explicitly

First I was surprised that we don't use the return value of this Try-Get.
But then I realized that we check the value to be non-null anyway.

If we intentionally discard the return value, I suggest explicitly discarding it for future reads:
- `_ = data.TryGetValue(key, out var value);`

If we did not intend to discard the return value, I suggest using it:
- `if (!data.TryGetValue(key, out var value) || value is null or Java.Lang.Long or ...`

src/Sentry/Platforms/Android/DotnetReplayBreadcrumbConverter.cs
```diff
@@ -0,0 +1,60 @@
...
+internal class DotnetReplayBreadcrumbConverter(Sentry.JavaSdk.SentryOptions options)
+    : DefaultReplayBreadcrumbConverter(options), IReplayBreadcrumbConverter
+{
+    private const string HttpCategory = "http";
+
+    public override global::IO.Sentry.Rrweb.RRWebEvent? Convert(Sentry.JavaSdk.Breadcrumb breadcrumb)
+    {
+        // The Java converter expects httpStartTimestamp/httpEndTimestamp to be Double or Long.
+        // .NET breadcrumb data is always stored as strings. We convert these to numeric here so that the base.Convert()
+        // method doesn't throw an exception.
```
Flash0ver: question: String vs Number

Is it a bug on our end that we send both of these Timestamps as Strings rather than Numbers?
Or is this a difference between v1 and v2 Spans?

I guess - either way - something to consider for Span-first/-streaming, is it?

src/Sentry/SentryHttpMessageHandler.cs
```diff
@@ -89,6 +91,13 @@ protected internal override void HandleResponse(HttpResponseMessage response, IS
             {"method", method},
             {"status_code", ((int) response.StatusCode).ToString()}
         };
+        if (span is not null)
```
Flash0ver, line 95: question: also for non-Session-Relay

Is this also valid for use cases other than Session Replay,
or would this be overhead in space and time with not much benefit?

I'm wondering if we need to check whether Session Replay is enabled.
Or do conditional compilation when `Android`.

If this should be done in all cases, we could consider setting a capacity when initializing the `breadcrumbData` Dictionary, because I think the default capacity is `3`, so we would always (or at least in most cases) have some Array-Growing going on.

src/Sentry/SentryHttpMessageHandler.cs
```diff
@@ -89,6 +91,13 @@ protected internal override void HandleResponse(HttpResponseMessage response, IS
...
+        {
+            // Ensure the breadcrumb can be converted to RRWeb so that it shows up in the network tab in Session Replay.
+            // See https://github.com/getsentry/sentry-java/blob/94bff8dc0a952ad8c1b6815a9eda5005e41b92c7/sentry-android-replay/src/main/java/io/sentry/android/replay/DefaultReplayBreadcrumbConverter.kt#L195-L199
+            breadcrumbData[HttpStartTimestampKey] = span.StartTimestamp.ToUnixTimeMilliseconds().ToString("F0", CultureInfo.InvariantCulture);
+            breadcrumbData[HttpEndTimestampKey] = DateTimeOffset.UtcNow.ToUnixTimeMilliseconds().ToString("F0", CultureInfo.InvariantCulture);
```
Flash0ver, line 100: could you point me to the docs? I didn't find them 🙁
This is not related to the `performanceSpan`, is it?
(if it is, then I misunderstand something here)

I'm wondering if the format is correct.
E.g., for _Logs_ and _Metrics_, the Timestamps are in seconds since the Unix epoch.

test/Sentry.Tests/Platforms/Android/DotnetReplayBreadcrumbConverterTests.cs
```diff
@@ -0,0 +1,37 @@
...
+        // Act
+        var rrwebEvent = converter.Convert(breadcrumb);
+
+        // Assert
+        rrwebEvent.Should().BeOfType<IO.Sentry.Rrweb.RRWebSpanEvent>();
+        var rrWebSpanEvent = rrwebEvent as IO.Sentry.Rrweb.RRWebSpanEvent;
+        Assert.NotNull(rrWebSpanEvent);
+        // Note the converter divides by 1000 to get ms
+        rrWebSpanEvent.StartTimestamp.Should().Be(1625079600L);
+        rrWebSpanEvent.EndTimestamp.Should().Be(1625079660L);
```
Flash0ver, line 35: question: `performanceSpan`

Is this related to the `performanceSpan`?
The `performanceSpan` - so I believe - is in seconds ... but this is in Milliseconds.
No wait ... this is in Seconds ... but why is it then a `long` rather than a `double`.
OK ... I believe there is something I misunderstand.

UPDATE:
Oh .... `StartTimestamp` and `EndTimestamp` actually are `double`s ... and not `long`s.
I guess the `L` suffix threw me off.
Is it worth changing the `L` suffix to `.0` or `d` to indicate that it's indeed a `double`?

```diff
-1625079600L
+1625079600.0
+1625079600d
```

test/Sentry.Tests/SentryHttpMessageHandlerTests.cs
```diff
@@ -611,25 +611,74 @@ public void Send_Executed_BreadcrumbCreated()
         Assert.True(breadcrumbGenerated.Data.ContainsKey(statusKey));
         Assert.Equal(expectedBreadcrumbData[statusKey], breadcrumbGenerated.Data[statusKey]);
     }
+#endif
 
     [Fact]
-    public void Send_Executed_FailedRequestsCaptured()
```
Flash0ver, line 616: question: existing test removed

Is this test no longer valid?
It seems odd to me to remove that test.
What am I missing?

src/Sentry/Platforms/Android/DotnetReplayBreadcrumbConverter.cs
jamescrosswell: If it fails, it would be because we're handling a breadcrumb created by an SDK user.
1. I think this would be an edge case... they'd have to create a pretty specific breadcrumb for it to get tangled up in here
2. I don't think they'd be expecting the converter to run (or even aware that the converter exists).

So no, I don't think we need to log anything here.

src/Sentry/Platforms/Android/DotnetReplayBreadcrumbConverter.cs
jamescrosswell: It's nothing to do with spans... it's this:
https://github.com/getsentry/sentry-dotnet/blob/c32dfbbc0246141d944fd95f5c9409326936db22/src/Sentry/Breadcrumb.cs#L53

So the .NET SDK stores breadcrumb data as a dict<string,string>... there's no way to store numeric values in there.

The Java SDK monitors those breadcrumbs and converts them to [performance spans](https://develop.sentry.dev/sdk/data-model/event-payloads/replay-recording/#performancespan) (different from tracing spans), which show up in the Network tab of session replay sessions. 

It's a bit confusing - I think they should have named them `ReplayEvents` or something rather than `PerformanceSpans`. Internally, they're even converted to `RRWebSpanEvents` so event would have been a better name... coulda, shoulda though - we can't change it now.

src/Sentry/SentryHttpMessageHandler.cs
jamescrosswell: I had actually done something like that in a follow up PR for iOS:
- https://github.com/getsentry/sentry-dotnet/pull/4891/changes#diff-69e7774461ffb8785e75a86321b28a9c9ff2e64ec20b4dc4cadc269bceb59ed0

I thought of doing it here... it would have made unit testing a bit more painful because then the unit tests for this functionality need to be run as device tests on Android. We could bring forward the `#if ANDROID` stuff in the code/tests from that other PR and include it here though.

src/Sentry/SentryHttpMessageHandler.cs
jamescrosswell: It's kind of related to the performanceSpan yes... we're only storing it as a string temporarily at this point. It gets converted to a long here:
https://github.com/getsentry/sentry-dotnet/blob/d63141806ccb5f80a632fc52e7a6dd45efbfacde/src/Sentry/Platforms/Android/DotnetReplayBreadcrumbConverter.cs#L20-L21

The only reason we're adding this data at all, is so that it will be picked up by the Android SDK and converted to a performanceSpan.

So we could store it in whatever string format we want... there are no "docs".

test/Sentry.Tests/Platforms/Android/DotnetReplayBreadcrumbConverterTests.cs
jamescrosswell: It's specific to the Java implementation:
https://github.com/getsentry/sentry-java/blob/94bff8dc0a952ad8c1b6815a9eda5005e41b92c7/sentry-android-replay/src/main/java/io/sentry/android/replay/DefaultReplayBreadcrumbConverter.kt#L215-L228

It gets divided by 1000 on the Java side so we have to send it in ms.

test/Sentry.Tests/SentryHttpMessageHandlerTests.cs
jamescrosswell: Ouch... that was just a mistake. Will revert.

src/Sentry/Platforms/Android/DotnetReplayBreadcrumbConverter.cs
Flash0ver: got it ... thanks

test/Sentry.Tests/Platforms/Android/DotnetReplayBreadcrumbConverterTests.cs
Flash0ver: oh ... it actually is in `ms` ... then `L` is indeed more specific

src/Sentry/SentryHttpMessageHandler.cs
Flash0ver: ah ... I see ... thanks for clarifying it in the `//` comment

PR #4868, opened 2026-01-23
ref: revert accessibility change in `AssemblyStoreReader`

Flash0ver:
Follow-up to
- https://github.com/getsentry/sentry-dotnet/pull/4814/changes#r2707861556

I wanted to give [The `TestAccessor` Pattern](https://github.com/dotnet/roslyn/blob/main/docs/TestAccessor%20Pattern.md) from _Roslyn_ a try.
See comments.

But changed to deriving a file-local type and use it in the test, as originally suggested.

We eventually decided against this change.
Making one member `internal` is just less code to maintain than derived types and/or Test-Accessors.

### DIFF
--- a/.github/workflows/alpine.yml
+++ b/.github/workflows/alpine.yml
@@ -23,7 +23,7 @@ jobs:
       packages: write
 
     steps:
-      - uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # v6.0.1
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
 
       - uses: docker/login-action@5e57cd118135c172c3672efd75eb46360885c0ef # v3.6.0
         with:
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -48,7 +48,7 @@ jobs:
           curl -sSL https://raw.githubusercontent.com/${{ github.repository }}/${{ github.sha }}/.github/alpine/setup-node.sh | sudo bash /dev/stdin
 
       - name: Checkout
-        uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # v6.0.1
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
 
       - run: git submodule update --init modules/sentry-native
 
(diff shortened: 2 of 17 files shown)


src/Sentry.Android.AssemblyReader/V2/StoreReader.TestAccessor.cs
```diff
@@ -0,0 +1,19 @@
...
+    {
+        private readonly StoreReader _instance;
+
+        internal TestAccessor(StoreReader instance)
+        {
+            _instance = instance;
+        }
+
+        internal bool IsSupported() => _instance.IsSupported();
+    }
```
Flash0ver: question: What do you think about Roslyn's [`TestAccessor` Pattern](https://github.com/dotnet/roslyn/blob/main/docs/TestAccessor%20Pattern.md) in this case?

Flash0ver:
@sentry review

src/Sentry.Android.AssemblyReader/V2/StoreReader.TestAccessor.cs
jamescrosswell: Seems like there are pros and cons.

Pros:
- I like the fact that it's explicit about what is being exposed and why. 
- It would allow us to leave some stuff private... although I can't remember a time when we got into trouble over the distinction between private and internal so I think in practice this is a negligible benefit

Cons:
- It is way more verbose than just changing the visibility of stuff to internal

I think we could do it for new code if we really wanted to. It's really a styling choice. For existing code, it's a whole lot of work for practically no benefit at all...

Flash0ver: Thanks for your feedback.
I removed the `TestAccessor`, and replaced it with a derived file-local type, as originally suggested in https://github.com/getsentry/sentry-dotnet/pull/4814/changes#r2707861556.

jamescrosswell: Wouldn't that also sometimes require changing member visibility just for testing (e.g. private -> protected)? And we couldn't do this for sealed classes either.

What is the key problem we're trying to solve here?
1. The risk of one of the maintainers using an internal member when they shouldn't?
2. The lack of clarity around why some members are internal?
3. Something else?

I don't think 1 is a serious concern... especially given our review process. 2 is usually handled by the compiler... if you try to make those internal members private, the tests won't compile anymore... 3 then?

It would be nice/pretty if C# had an attribute you could decorate members with to say they're public only to test assemblies... but short of that, I don't think there are any workarounds that don't come with both pros and cons. I think the approach we've been taking so far (using `InternalsVisibleTo`) is probably the best available alternative... it requires very little work, doesn't affect the source types at all for SDK users and has very little risk. It's not 'pretty' is about my only complaint with it but I think we have to be pragmatic.

Flash0ver: Fair point!
You convinced me.

Let's keep it as is.
Thanks for your feedback.

PR #4879, opened 2026-01-28
chore: update Polyfill from 1.32.0 to 9.8.1

Flash0ver:
### Description
Update Polyfill from `1.32.0` to `9.8.1`.

### Related
- https://github.com/SimonCropp/Polyfill/issues/454
- https://github.com/getsentry/sentry-dotnet/pull/4857

### Additional Changes
- remove now redundant custom Polyfills
- enable Polyfill-feature `FeatureHttp` for _.NET Standard 2.0_ and _.NET Standard 2.1_
- ensure Polyfill-feature `FeatureMemory` is not used when `System.Memory` is in an incompatible version
  - for Roslyn Components with Compiler Extensions

### New Options
The new version of `Polyfill` also comes with "StringInterpolation" polyfills.
- Pro: Enabling the "StringInterpolation" polyfill will have the compiler consider the `DefaultInterpolatedStringHandler` for string Interpolation (`$"{}"`) which is generally more efficient compared to `String.Format` or `String.Concat` (when not all elements are `string`s)
- Con: The assembly increases in size ... see [Assembly size impact](https://github.com/SimonCropp/Polyfill?tab=readme-ov-file#assembly-size-impact)

### Follow-up
Enable and use "ArgumentExceptions" polyfills:
- #4982

### DIFF
--- a/src/Sentry/Sentry.csproj
+++ b/src/Sentry/Sentry.csproj
@@ -91,6 +91,12 @@
     <PackageReference Include="System.Text.Json" Version="8.0.5" />
   </ItemGroup>
 
+  <!-- Due to Polyfill, System.Memory needs to be a top-level dependency, otherwise app start may fail with:
+       System.IO.FileLoadException: Could not load file or assembly 'System.Memory' or one of its dependencies. The located assembly's manifest definition does not match the assembly reference. -->
+  <ItemGroup Condition="'$(TargetFramework)' == 'netstandard2.0' Or '$(TargetFramework)' == 'net462'">
+    <PackageReference Include="System.Memory" Version="4.6.3" />
+  </ItemGroup>
+
   <!--
     Include Sentry's custom targets file in the nuget package.
     This file contains targets that are invoked during the end-user's build.


Flash0ver:
Potential fixes for making the namespace `Polyfills` globally available for projects that reference `src/Sentry.csproj` and has it's internals made been visible to:
- add `<ItemGroup><Using Include="Polyfills" /></ItemGroup>` to each of these projects
- some MSBuild additions and restructuring
- via an repo-internal Source-Generator utilizing [GivesAccessTo](https://learn.microsoft.com/dotnet/api/microsoft.codeanalysis.iassemblysymbol.givesaccessto)

_Update_:
I went with a simpler conditional `<Using Include="Polyfills" ...`.

src/Sentry/Internal/Extensions/CollectionsExtensions.cs
```diff
@@ -69,18 +69,14 @@ public static IEnumerable<KeyValuePair<TKey, TValue>> WhereNotNullValue<TKey, TV
...
+
     public static IReadOnlyList<T> AsReadOnly<T>(this IList<T> list) =>
         list as IReadOnlyList<T> ?? new ReadOnlyCollection<T>(list);
 
-#if !NET7_0_OR_GREATER
-    public static IReadOnlyDictionary<TKey, TValue> AsReadOnly<TKey, TValue>(this IDictionary<TKey, TValue> dictionary)
-        where TKey : notnull =>
-        new ReadOnlyDictionary<TKey, TValue>(dictionary);
-#endif
-
```
Flash0ver, line 83: note: now included via `Polyfill`

src/Sentry/Sentry.csproj
```diff
@@ -65,8 +65,14 @@
     https://github.com/SimonCropp/Polyfill
   -->
   <ItemGroup>
-    <PackageReference Include="Polyfill" Version="1.32.0" PrivateAssets="all" />
+    <PackageReference Include="Polyfill" Version="9.8.1" PrivateAssets="all" />
   </ItemGroup>
+  <PropertyGroup>
+    <PolyStringInterpolation>true</PolyStringInterpolation>
```
Flash0ver, line 71: note: `PolyStringInterpolation` emits a `DefaultInterpolatedStringHandler`, which the compiler considers for String Interpolation when not all holes are strings. This is more efficient than `String.Format`. But makes our `netstandard2.1`, `netstandard2.0` and `net462` targets, and with that the entire `Sentry` NuGet package, a bit larger:

See [Assembly size impact](https://github.com/SimonCropp/Polyfill?tab=readme-ov-file#assembly-size-impact)

src/Directory.Build.props
```diff
@@ -60,4 +60,10 @@
     <PackageReference Include="SIL.ReleaseTasks" Version="2.5.0" PrivateAssets="All" Condition="'$(Configuration)' != 'Debug'" />
   </ItemGroup>
 
+  <!-- "Sentry" uses "Polyfill" and gives access to it's internals via "InternalsVisibleTo", but some Polyfills are in a separate namespace -->
+  <ItemGroup Condition="'$(MSBuildProjectName)' != 'Sentry' And $(MSBuildProjectName.StartsWith('Sentry.'))">
+    <!-- the latest TargetFramework requires no Polyfills; "Sentry.Compiler.Extensions" has its own Polyfills; "Android" and "Cocoa" projects are referenced by "Sentry" -->
+    <Using Include="Polyfills" Condition="'$(TargetFramework)' != $(LatestTfm) And $(MSBuildProjectName) != 'Sentry.Compiler.Extensions' And $(MSBuildProjectName) != 'Sentry.Android.AssemblyReader' And $(MSBuildProjectName) != 'Sentry.Bindings.Android' And $(MSBuildProjectName) != 'Sentry.Bindings.Cocoa'" />
+  </ItemGroup>
+
```
Flash0ver, line 68: note: this (and see also `test/Directory.Build.props`) is required

because some of the Polyfills have been moved into the `Polyfills` namespace.

This needs to be conditional,
because `Sentry`, to which `Polyfill` is emitting the Polyfills to as `internal` types,
does not give access of it's internals to every project in the `src` (and `test`) directory.

Alternatively,
we could add the global using (`<Using Include="Polyfills" />`) to every project that needs it,
but I found that to be quite a lot of additions, and new projects don't have them then per default,
so I suggest here to do it in a more central place.

src/Sentry.Analyzers/Sentry.Analyzers.csproj
```diff
@@ -24,8 +24,19 @@
       https://github.com/SimonCropp/Polyfill
     -->
     <ItemGroup>
-      <PackageReference Include="Polyfill" Version="1.32.0" PrivateAssets="all" />
+      <PackageReference Include="Polyfill" Version="9.8.1" PrivateAssets="all" />
```
Flash0ver, line 27: issue: Test failure

- Project: `Sentry.Tests`
- Test: `Sentry.Tests.SentrySdkCrashTests.CauseCrashInSeparateProcess`
- TFM: `net48`

```
Unhandled Exception: System.IO.FileLoadException: Could not load file or assembly 'System.Memory, Version=4.0.1.2, Culture=neutral, PublicKeyToken=cc7b13ffcd2ddd51' or one of its dependencies. The located assembly's manifest definition does not match the assembly reference. (Exception from HRESULT: 0x80131040)
   at Sentry.SentrySdk.CauseCrash(CrashType crashType)
   at Sentry.Testing.CrashableApp.Program.Main(String[] args) in D:\GitHub\sentry-dotnet\test\Sentry.Testing.CrashableApp\Program.cs:line 9
```

src/Sentry/Sentry.csproj
```diff
@@ -86,6 +91,12 @@
     <PackageReference Include="System.Text.Json" Version="8.0.5" />
   </ItemGroup>
 
+  <!-- Due to Polyfill, System.Memory needs to be a top-level dependency, otherwise app start may fail with:
+       System.IO.FileLoadException: Could not load file or assembly 'System.Memory' or one of its dependencies. The located assembly's manifest definition does not match the assembly reference. -->
+  <ItemGroup Condition="'$(TargetFramework)' == 'netstandard2.0' Or '$(TargetFramework)' == 'net462'">
+    <PackageReference Include="System.Memory" Version="4.6.3" />
+  </ItemGroup>
+
```
Flash0ver, line 99: thought: This could be an issue!

related to https://github.com/getsentry/sentry-dotnet/pull/4879/changes#r2895293577
(`Sentry.Tests.SentrySdkCrashTests.CauseCrashInSeparateProcess(CrashType)` is a great test !!!)

I wonder if this may be considered a breaking change,
and if we should hold off on this upgrade for our next Major (Sentry v7 supporting .NET 11).

But this would mean to similarly hold off on follow-up changes like
- #4985

(although, actually, we could simply write our own `ANE.ThrowIfNull(object?)` internal static extension method for now, and later replace it with the Polyfill-variant)

jamescrosswell: Yeah I'd say we avoid adding dependencies to the core package... especially when it's just for polyfills which we don't need for any functional reason (they're just to make the code slightly more attractive to read really). 

What's the most recent version of Polyfill that doesn't require we change any dependencies for SDK users? We could try bumping to that instead maybe?

Flash0ver: Great idea!
Unfortunately, the `Polyfill` changelog is quite ... unavailable.
Will "binary search" on a _Windows_ machine.

Flash0ver: No changelog / release notes == no fun ☹️.

`1.34.0` includes this - for our scenario - breaking change.

Closing this PR.
Instead opened PR #5007 to update to the latest non-breaking version (`1.33.2`).
Opened issue #5006 to be considered for our next Major release.

Flash0ver:
Superseded by #5007 without incurring https://github.com/getsentry/sentry-dotnet/pull/4879#discussion_r2903984833.
Followed-up by #5006 for our next Major release.

PR #4883, opened 2026-01-30
fix: Apply `Enricher` to User Feedback

bitsandfoxes:
Resolves https://github.com/getsentry/sentry-unity/issues/2510

Without the enricher, User Feedback is lacking `environment` and `release`.

Screenshot 2026-01-30 at 14 32 02

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@
 
 ### Fixes
 
-- User Feedback now correctly contains `Environment` and `Release` ([#4883](https://github.com/getsentry/sentry-dotnet/pull/4883))
+- User Feedback now contains additional Context and Tags, like `Environment` and `Release` ([#4883](https://github.com/getsentry/sentry-dotnet/pull/4883))
 - Allow Sentry failures from the Sentry CLI when SENTRY_ALLOW_FAILURE is set ([#4852](https://github.com/getsentry/sentry-dotnet/pull/4852))
 - The SDK now logs a specific error message when envelopes are rejected due to size limits (HTTP 413) ([#4863](https://github.com/getsentry/sentry-dotnet/pull/4863))
 - Fixed thread-safety issue on Android when multiple events are captured concurrently ([#4814](https://github.com/getsentry/sentry-dotnet/pull/4814))


test/Sentry.Tests/SentryClientTests.cs
```diff
@@ -995,6 +995,27 @@ public void CaptureFeedback_WithHint_HasHintAttachment()
...
+        Envelope envelope = null;
+        var sut = _fixture.GetSut();
+        sut.Worker.EnqueueEnvelope(Arg.Do<Envelope>(e => envelope = e));
+        var feedback = new SentryFeedback("Test feedback");
+
+        // Act
+        sut.CaptureFeedback(feedback);
+
+        // Assert
+        var item = envelope.Items.First(x => x.TryGetType() == EnvelopeItem.TypeValueFeedback);
```
Flash0ver, line 1014: suggestion: assert result

Would you mind also asserting the resulting ID not to be `SentryId.Empty`, for consistency with the other `CaptureFeedback` tests.


```suggestion
        // Act
        var result = sut.CaptureFeedback(feedback);

        // Assert
        result.Should().NotBe(SentryId.Empty);
        var item = envelope.Items.First(x => x.TryGetType() == EnvelopeItem.TypeValueFeedback);
```

CHANGELOG.md
```diff
@@ -1271,7 +1272,6 @@ If you have compilation errors you can find the affected types or overloads miss
 - The `WithScope` and `WithScopeAsync` methods have been removed. We have discovered that these methods didn't work correctly in certain desktop contexts, especially when using a global scope. ([#2717](https://github.com/getsentry/sentry-dotnet/pull/2717))
 
   Replace your usage of `WithScope` with overloads of `Capture*` methods:
-
   - `SentrySdk.CaptureEvent(SentryEvent @event, Action<Scope> scopeCallback)`
```
Flash0ver, line 1275: praise: thanks for the cleanup 🥔

src/Sentry/SentryClient.cs
```diff
@@ -108,6 +108,7 @@ public SentryId CaptureFeedback(SentryFeedback feedback, out CaptureFeedbackResu
         // Evaluate and copy before invoking the callback
         scope.Evaluate();
         scope.Apply(evt);
+        _enricher.Apply(evt);
```
Flash0ver, line 111: question: more than `environment` and `release`

More than `environment` and `release` are applied ... is this intended?

Flash0ver, line 111: question: precedence

Is the order of first applying the Scope, and then the Event-Enricher correct?
Or should instead e.g. Tags on the Scope overwrite Default-Tags from the Enricher?

On the other hand, this question would then similarly apply to e.g. Transactions, too, wouldn't it?
And changing the order the could be considered a behavioral breaking change ... or maybe a fix?

On the other other hand ... we could just let "local consistency" win and do it as you suggested.
What are your thoughts?

jamescrosswell: Possibly related - some time ago we made this change:
- https://github.com/getsentry/sentry-dotnet/pull/2496

No mention of the enricher was made for that particular initiative, but ideally the events would be enriched before any user defined filters get applied, so that the  enriched values are available for inspection in before send callbacks etc. As long as we're doing that, should be all good.

jamescrosswell: It looks like we already do this here as well...
https://github.com/getsentry/sentry-dotnet/blob/fed52443fe8ac0e074d7b0f0be9c815b4176ace0/src/Sentry/SentryClient.cs#L177-L180

Maybe we refactor it into a tiny little method (which could conceivably be reused in CaptureCheckIn, CaptureSession or some other future CaptureFoo method)? That way we'd ensure consistency across envelope items.

jamescrosswell:
btw: this will need to be updated from `main` once we've approved/merged https://github.com/getsentry/sentry-dotnet/pull/4884:
- https://github.com/getsentry/sentry-dotnet/pull/4884

src/Sentry/SentryClient.cs
Flash0ver: verified: the event is enriched **before** the `SetBeforeSend` delegate is invoked

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@
 
 ### Fixes
 
+- User Feedback now correctly contains `Environment` and `Release` ([#4883](https://github.com/getsentry/sentry-dotnet/pull/4883))
```
Flash0ver: suggestion: rephrase more vaguely / "inclusive"

After a re-review, I believe all is correct.
Let's perhaps rephrase the CHANGELOG entry, as this fix also sets other Tags like `"runtime"`, and other Contexts like `"runtime"`.

```suggestion
- User Feedback now contains additional Context and Tags, like `Environment` and `Release` ([#4883](https://github.com/getsentry/sentry-dotnet/pull/4883))
```

src/Sentry/SentryClient.cs
Flash0ver: Also, the `Enricher` mostly performs _null-coalescing assignment_ (`??=`). The order should be fine.

Flash0ver (approved):

PR #4884, opened 2026-02-01
fix: Permissions for changelog preview workflow

jamescrosswell:
Not sure how this worked before merging, but it appears to require a permission that it's missing. 

See https://github.com/getsentry/sentry-dotnet/actions/runs/21570267671:
> The workflow is not valid. .github/workflows/changelog-preview.yml (Line: 16, Col: 3): Error calling workflow 'getsentry/craft/.github/workflows/changelog-preview.yml@v2'. The workflow is requesting 'statuses: write', but is only allowed 'statuses: none'.

#skip-changelog

### DIFF
--- a/Sentry-CI-Build-Windows-arm64.slnf
+++ b/Sentry-CI-Build-Windows-arm64.slnf
@@ -3,6 +3,8 @@
     "path": "Sentry.sln",
     "projects": [
       "benchmarks\\Sentry.Benchmarks\\Sentry.Benchmarks.csproj",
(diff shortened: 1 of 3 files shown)


.github/workflows/changelog-preview.yml
```diff
@@ -11,6 +11,7 @@ on:
 permissions:
   contents: write
   pull-requests: write
+  statuses: write
```
Flash0ver, line 14: note: see https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#permissions

PR #4894, opened 2026-02-04
style: file nesting

Flash0ver:
Nest C# files that are all parts of a `partial` type.

Unfortunately, Rider does not support Nesting Files based on Wildcards: https://www.jetbrains.com/help/rider/File_Nesting_Dialog.html
So we need to be quite explicit.

After: getsentry/sentry-dotnet#4834

See also: https://github.com/getsentry/sentry-dotnet/pull/4834#discussion_r2756076282

image


jamescrosswell (approved):
QOL - thanks @Flash0ver

PR #4907, opened 2026-02-10
feat(blazor): Add navigation breadcrumbs for Blazor WebAssembly

### LINKED ISSUE #4906: Blazor WebAssembly SDK: Next-Gen Feature Expansion
## Context

Issue #2021 tracked initial Blazor WebAssembly support and resulted in the `Sentry.AspNetCore.Blazor.WebAssembly` package. That package currently provides a `UseSentry` entry point on `WebAssemblyHostBuilder` that wires up `Sentry.Extensions.Logging` with WASM-safe defaults (no compression, fast startup detection, global mode). `SentryBlazorOptions` is an empty class extending `SentryLoggingOptions`.

This issue tracks the next steps: expanding the package into a fully-featured Blazor WebAssembly SDK. PR #2569 (Hackweek spike) explored many of these ideas. The dotnet/runtime#91049 (`build_id` in WASM binaries) is milestoned for .NET 11 and unblocks native WASM symbolication.

### Phase 1 — Quick Wins (Pure C#, No JS Dependency)

- [ ] **Navigation Breadcrumbs** — Subscribe to `NavigationManager.LocationChanged` to create breadcrumbs with `from`/`to` URLs. Keep `scope.Request.Url` updated with current route.

- [ ] **Browser/OS Context via User-Agent** — Read `navigator.userAgent` via JS interop at startup. Set on scope so Sentry extracts browser, OS, and device info.

- [ ] **ErrorBoundary Integration** — Provide a `SentryErrorBoundary` component (extending `ErrorBoundaryBase`) that calls `SentrySdk.CaptureException()` with Blazor-specific context (component name, route).

- [ ] **Scope Defaults** — Set `Runtime` context to "Blazor WebAssembly" + .NET version, proper `sdk.name`, and populate `Environment` from `IWebAssemblyHostEnvironment.Environment`.

### Phase 2 — Core Error & Tracing Infrastructure

- [ ] **Unhandled Exception Capture via JS** — Register `window.addEventListener('error', ...)` and `window.addEventListener('unhandledrejection', ...)` to capture errors that escape the .NET runtime (async failures, JS interop errors).

- [ ] **WASM Stack Trace Enhancement** — Custom `ISentryStackTraceFactory` that parses WASM frames (`wasm-function[N]:0xADDR`) into structured `SentryStackFrame` objects with `Platform = "native"`, `InstructionAddress`, and `AddressMode` for server-side symbolication. PR #2569 had `WasmStackTraceFactory` with tests.

- [ ] **HttpClient Instrumentation + Trace Propagation** — This is the most impactful tracing feature. Blazor WASM apps are inherently client-server; connecting the client trace to the backend trace into one coherent view is critical.

  The challenge: if both the JS SDK (via `fetch` instrumentation) and C# SDK (via `SentryHttpMessageHandler`) instrument outgoing HTTP, we get duplicate spans and conflicting trace headers. The C# layer should own HTTP instrumentation since `HttpClient` is where developers make calls. Options:
  - C# `DelegatingHandler` injects `sentry-trace`/`baggage` headers, disable `fetch` instrumentation in JS SDK's `BrowserTracing`
  - Hybrid: C# generates trace ID and attaches headers, JS SDK configured to continue that trace rather than start a new one

- [ ] **Bundle JavaScript SDK as Static Asset** — Download Sentry JS SDK at package build time (MSBuild `DownloadFile`). Serve via `_content/` static web assets. Auto-inject or provide a `<SentryScript>` component. Enables Session Replay, Browser Performance, and native error capture.

### Phase 3 — Rich Observability

- [ ] **C#-to-JS Scope Synchronization** — `IScopeObserver` implementation forwarding `SetTag`, `SetExtra`, `SetUser`, `AddBreadcrumb` to JS SDK via `IJSRuntime`. PR #2569 had `JavaScriptScopeObserver`.

- [ ] **Session Replay** — Works "pretty well" per Hackweek testing. Add `EnableReplay` / sample rate options to `SentryBlazorOptions`, pass to JS SDK init.

- [ ] **Browser Performance Monitoring** — Enable `BrowserTracing` in JS SDK for page load timing, web vitals (LCP, FID, CLS). Initial version: just enable the integration. Advanced: bridge C# spans into JS trace.

- [ ] **Blazor Component Lifecycle Spans** — Instrument `OnInitializedAsync`, `OnParametersSetAsync`, `OnAfterRenderAsync` as spans. In .NET 10+, `Microsoft.AspNetCore.Components` emits OpenTelemetry activities (used in Blazor Server sample's `BlazorSentryIntegration`). Investigate if those activity sources are available in WASM — if so, this could be lower effort.

- [ ] **User Feedback Widget** — JS SDK feedback widget + `<SentryUserFeedback>` Razor component passing `SentrySdk.LastEventId`.

### Phase 4 — Infrastructure

- [ ] **Automated Debug Symbol Upload for WASM** — Ensure Sentry CLI MSBuild integration works for Blazor WASM publish output (PDBs for managed code, DWARF for AOT). Test both interpreted and AOT modes.

- [ ] **WASM `build_id` / Debug Image Support** — Blocked on dotnet/runtime#91049 (.NET 11). Once available, register debug images in events for native symbolication. Interim: opt-in JS-side `WebAssembly.instantiateStreaming` patching to extract `build_id` from custom sections (requires `<BlazorCacheBootResources>false</BlazorCacheBootResources>`).

- [ ] **Offline Event Caching (PWA)** — Blazor WASM PWAs work offline. Implement envelope cache via `IndexedDB` (JS interop) to store events and flush when connectivity returns. Core SDK's `CachingTransport` uses filesystem APIs unavailable in WASM.

### Phase 5 — Advanced / Future

- [ ] **Lazy-Loaded Assembly Awareness** — Track which assemblies are loaded via `LazyAssemblyLoader` at error time, add as event context.
- [ ] **Profiling** — Browser JS profiling API or .NET EventPipe in WASM (both experimental).

## Key Dependencies & Risks

| Risk | Detail |
|------|--------|
| **`dotnet/runtime#91049`** | `build_id` in WASM milestoned for .NET 11. Full native symbolication blocked until then. |
| **JS SDK version pinning** | Bundling creates maintenance burden — must track upstream releases. |
| **Safari/WebKit** | `@sentry/wasm` frame parsing doesn't work on WebKit (sentry-javascript#7223). Degrade gracefully. |
| **AOT vs. Interpreted** | Stack trace behavior differs significantly. Must test both modes. |
| **C#/JS trace deduplication** | Core challenge for HttpClient instrumentation — one layer must own HTTP spans to avoid duplicates and conflicting trace headers. |

## References

- #2021 — Original feature request (closed, resulted in basic package)
- #2569 — Hackweek spike PR with prototype implementations
- dotnet/runtime#91049 — WASM `build_id` section (.NET 11)
- sentry-javascript#8259 — `@sentry/wasm` blob URL limitations

bruno-garcia:
## Summary

- Automatically create `navigation` breadcrumbs when users navigate between pages in a Blazor WebAssembly app
- Breadcrumbs use `type: "navigation"`, `category: "navigation"`, and `data: { from, to }` with relative paths — matching the JS SDK behavior so the Sentry UI renders the navigation icon and from/to display correctly
- Keeps `scope.Request.Url` updated with the current route

Closes #4906

## Implementation

- **`BlazorWasmOptionsSetup`** (`IConfigureOptions<SentryBlazorOptions>`) injects `NavigationManager`, subscribes to `LocationChanged`, and creates breadcrumbs on each navigation
- Registered in DI via `WebAssemblyHostBuilderExtensions.UseSentry()`
- Uses `NavigationManager.ToBaseRelativePath()` to convert absolute URLs to relative paths (prepending `/` to match JS SDK format)

## Test plan

- [x] 7 unit tests covering:
  - Initial `scope.Request.Url` is set correctly (root and with path)
  - Breadcrumb has correct `type` and `category`
  - `from`/`to` values are relative paths
  - `scope.Request.Url` is updated on navigation
  - Multiple navigations track `from` correctly (previous → current)
  - Navigation from a non-root initial path tracks correct `from`
- [x] All tests pass on net8.0, net9.0, net10.0

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Features
+
+- Add navigation breadcrumbs for Blazor WebAssembly ([#4907](https://github.com/getsentry/sentry-dotnet/pull/4907))
+
 ### Dependencies
 
 - Bumped Xamarin.Kotlin.StdLib.Jdk8 to 2.2.20 ([#4876](https://github.com/getsentry/sentry-dotnet/pull/4876))
@@ -11,7 +15,6 @@
 
 ### Features
 
-- Add navigation breadcrumbs for Blazor WebAssembly ([#4907](https://github.com/getsentry/sentry-dotnet/pull/4907))
 - Add _experimental_ support for [Sentry trace-connected Metrics](https://docs.sentry.io/product/explore/metrics/) ([#4834](https://github.com/getsentry/sentry-dotnet/pull/4834))
 - Extended `SentryThread` by `Main` to allow indication whether the thread is considered the current main thread ([#4807](https://github.com/getsentry/sentry-dotnet/pull/4807))
 


test/Sentry.AspNetCore.Blazor.WebAssembly.Tests/BlazorWasmOptionsSetupTests.cs
```diff
@@ -0,0 +1,163 @@
...
+    public void Navigation_CreatesBreadcrumbWithCorrectTypeAndCategory()
+    {
+        // Arrange
+        _sut.Configure(new SentryBlazorOptions());
+
+        // Act
+        _navigationManager.NavigateTo("/dashboard");
+
+        // Assert
+        SentrySdk.ConfigureScope(scope =>
```
bruno-garcia: Should be testing against `Hub` instances to avoid the static state

src/Sentry.AspNetCore.Blazor.WebAssembly/Internal/BlazorWasmOptionsSetup.cs
```diff
@@ -0,0 +1,56 @@
...
+            scope.Request.Url = ToRelativePath(previousUrl);
+        });
+
+        _navigationManager.LocationChanged += (_, args) =>
+        {
+            var from = ToRelativePath(previousUrl);
+            var to = ToRelativePath(args.Location);
+
+            SentrySdk.AddBreadcrumb(
+                message: "",
```
bruno-garcia: should we really be sending `message: ""`? can/should we skip the property altogether or have an actual message there instead?

test/Sentry.AspNetCore.Blazor.WebAssembly.Tests/FakeNavigationManager.cs
```diff
@@ -0,0 +1,26 @@
+using Microsoft.AspNetCore.Components;
+using Microsoft.AspNetCore.Components.Routing;
```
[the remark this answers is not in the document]
bruno-garcia: Fixed — removed the unused `using Microsoft.AspNetCore.Components.Routing;`.

test/Sentry.AspNetCore.Blazor.WebAssembly.Tests/FakeNavigationManager.cs
```diff
@@ -0,0 +1,26 @@
...
+    public FakeNavigationManager(string baseUri = "https://localhost/", string initialUri = "https://localhost/")
+    {
+        Initialize(baseUri, initialUri);
+    }
+
+    public void NavigateTo(string uri)
+    {
+        var absoluteUri = ToAbsoluteUri(uri).ToString();
+        Uri = absoluteUri;
+        NotifyLocationChanged(isInterceptedLink: false);
```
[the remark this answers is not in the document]
bruno-garcia: Fixed — removed the public `NavigateTo(string)` method. Tests now call the base `NavigateTo` which routes through `NavigateToCore`.

src/Sentry.AspNetCore.Blazor.WebAssembly/Internal/BlazorWasmOptionsSetup.cs
```diff
@@ -0,0 +1,63 @@
...
+
+    internal BlazorWasmOptionsSetup(NavigationManager navigationManager, IHub hub)
+    {
+        _navigationManager = navigationManager;
+        _hub = hub;
+    }
+
+    public void Configure(SentryBlazorOptions options)
+    {
+        var previousUrl = _navigationManager.Uri;
```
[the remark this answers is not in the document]
bruno-garcia: Fixed — added `ArgumentNullException.ThrowIfNull(options);` for consistency with other `IConfigureOptions` implementations.

src/Sentry.AspNetCore.Blazor.WebAssembly/Internal/BlazorWasmOptionsSetup.cs
```diff
@@ -0,0 +1,63 @@
...
+                    }));
+
+            _hub.ConfigureScope(scope =>
+            {
+                scope.Request.Url = to;
+            });
+
+            previousUrl = args.Location;
+        };
+    }
```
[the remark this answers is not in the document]
bruno-garcia: Fixed — added a `_initialized` guard to prevent duplicate subscriptions if `Configure` is called more than once.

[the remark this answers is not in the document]
bruno-garcia: Intentional design. Using `IConfigureOptions` as an initialization hook is how we wire up side-effects that depend on DI-resolved services (`NavigationManager`). The options parameter isn't needed today but keeps the door open for Blazor-specific configuration. This is the same approach used in MAUI (`SentryMauiOptionsSetup`).

src/Sentry.AspNetCore.Blazor.WebAssembly/Internal/BlazorWasmOptionsSetup.cs
```diff
@@ -0,0 +1,63 @@
...
+
+    public void Configure(SentryBlazorOptions options)
+    {
+        var previousUrl = _navigationManager.Uri;
+
+        // Set the initial scope request URL
+        _hub.ConfigureScope(scope =>
+        {
+            scope.Request.Url = ToRelativePath(previousUrl);
+        });
```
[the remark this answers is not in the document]
bruno-garcia: Good observation. The initial `ConfigureScope` call is best-effort — it depends on whether options are resolved after SDK init. The `LocationChanged` handler (the main feature) always works correctly since navigations happen well after initialization. Leaving as-is for now; the worst case is the initial URL is missed on the scope, but it gets set correctly on the first navigation.

bruno-garcia:
## Review Comments Summary

| # | Author | Comment | Outcome |
|---|--------|---------|---------|
| 1 | @bruno-garcia | Tests should use Hub instances, not static state | **Fixed** (previous commit) — added internal `IHub` constructor, tests use NSubstitute mock |
| 2 | @bruno-garcia | Should not send `message: ""` in breadcrumb | **Fixed** (previous commit) — uses internal `Breadcrumb` constructor with `message: null` |
| 3 | Copilot | Unused `using Microsoft.AspNetCore.Components.Routing;` | **Fixed** — removed |
| 4 | Copilot | `FakeNavigationManager.NavigateTo` hides base member (CS0108) | **Fixed** — removed the public method; tests now call the inherited `NavigateTo` which routes through `NavigateToCore` |
| 5 | Copilot | Missing `ArgumentNullException.ThrowIfNull(options)` | **Fixed** — added for consistency with other `IConfigureOptions` implementations |
| 6 | cursor[bot] | Duplicate event handlers on repeated `Configure` calls | **Fixed** — added `_initialized` boolean guard |
| 7 | cursor[bot] | `IConfigureOptions` misused for side-effects | **No change** — intentional design; same pattern as `SentryMauiOptionsSetup`. `IConfigureOptions` is the standard hook for initialization requiring DI services |
| 8 | cursor[bot] | Initial scope URL may not be set due to init ordering | **No change** — best-effort by design. The `LocationChanged` handler (core feature) works correctly; initial URL is set on first navigation regardless |

All fixes are in commit 11ada43f.

Sentry.slnx
jamescrosswell, line 1: Ahhh... so much nicer 😜

CHANGELOG.md
jamescrosswell, line 1: FYI: We don't strictly require a changelog entry anymore - it should get generated automatically from the commit description since #4896

src/Sentry.AspNetCore.Blazor.WebAssembly/Internal/BlazorWasmOptionsSetup.cs
```diff
@@ -0,0 +1,72 @@
...
+                        { "from", from },
+                        { "to", to }
+                    }));
+
+            _hub.ConfigureScope(scope =>
+            {
+                scope.Request.Url = to;
+            });
+
+            previousUrl = args.Location;
```
jamescrosswell, line 63: Thank god for closures... makes it nice and simple.

src/Sentry.AspNetCore.Blazor.WebAssembly/Internal/BlazorWasmOptionsSetup.cs
```diff
@@ -0,0 +1,72 @@
...
+
+    public void Configure(SentryBlazorOptions options)
+    {
+        ArgumentNullException.ThrowIfNull(options);
+
+        if (_initialized)
+        {
+            return;
+        }
+        _initialized = true;
```
jamescrosswell, line 33: Is there a potential race condition here? Does this need to be thread safe? 

I think `IOptions<T>` gets evaluated lazily the first time it's requested... so if multiple requests hit the server concurrently after it starts, you could get two threads checking it `_initialized` is true, seeing that they're not and then setting up duplicate event handlers (resulting in duplicate breadcrumbs).

Better to use `Interlocked.Exchange`?

But if this is all compiling down to WASM, maybe there are no threads? There are none in javascript - I assume WASM is the same? @Flash0ver this sounds like your area of expertise...

edit: Turns out .NET does have [experimental support for threading in WASM](https://devblogs.microsoft.com/dotnet/asp-net-core-updates-in-dotnet-7-rc-2/#webassembly-multithreading-(experimental))... maybe we use Interlocked here just to be on the safe side then (even though it might not be strictly required)? I think for navigation events in particular, it's extremely unlikely to cause problems, so not pushing super hard for this.

jamescrosswell (commented):
The PR looks good. I had a question about thread safety on the `IConfigureOptions`... maybe this is a non-concern in WASM apps though.

src/Sentry.AspNetCore.Blazor.WebAssembly/Internal/BlazorWasmOptionsSetup.cs
bruno-garcia: This runs single threaded, since it runs on the main thread when the app starts up

CHANGELOG.md
bruno-garcia: Do I need to remove it? it's already there

src/Sentry.AspNetCore.Blazor.WebAssembly/Internal/BlazorWasmOptionsSetup.cs
```diff
@@ -0,0 +1,72 @@
...
+        {
+            scope.Request.Url = ToRelativePath(previousUrl);
+        });
+
+        _navigationManager.LocationChanged += (_, args) =>
+        {
+            var from = ToRelativePath(previousUrl);
+            var to = ToRelativePath(args.Location);
+
+            _hub.AddBreadcrumb(
```
Flash0ver, line 48: note: @Flash0ver 
- update getsentry/sentry-docs#16330 after merge

src/Sentry.AspNetCore.Blazor.WebAssembly/Internal/BlazorWasmOptionsSetup.cs
```diff
@@ -0,0 +1,72 @@
...
+                    data: new Dictionary<string, string>
+                    {
+                        { "from", from },
+                        { "to", to }
+                    }));
+
+            _hub.ConfigureScope(scope =>
+            {
+                scope.Request.Url = to;
+            });
```
Flash0ver, line 61: suggestion: we could avoid the closure here, as `to` is already a string-ified relative path, by passing it as "state"

src/Sentry.AspNetCore.Blazor.WebAssembly/WebAssemblyHostBuilderExtensions.cs
```diff
@@ -30,6 +33,8 @@ public static WebAssemblyHostBuilder UseSentry(this WebAssemblyHostBuilder build
             blazorOptions.IsGlobalModeEnabled = true;
         });
 
+        builder.Services.AddSingleton<IConfigureOptions<SentryBlazorOptions>, BlazorWasmOptionsSetup>();
```
Flash0ver, line 36: issue: with that change, `Sentry.Samples.AspNetCore.Blazor.Wasm` doesn't fully initialize any longer and just hangs

src/Sentry.AspNetCore.Blazor.WebAssembly/Internal/BlazorWasmOptionsSetup.cs
Flash0ver: I'm a bit uncertain if this is (or may be) an issue,
but considering that the change using our `InterlockedBoolean` is both cheap in runtime and maintenance (would result in less code, actually),
I vote for a thread-safe guard here using our internal `InterlockedBoolean`.

src/Sentry.AspNetCore.Blazor.WebAssembly/WebAssemblyHostBuilderExtensions.cs
Flash0ver: however, the Unit Tests and Playwright Tests (#4908) do pass

CHANGELOG.md
jamescrosswell: Nah you can leave it in... it would be generated automatically for the PR (from the title) when making a release otherwise. Just FYI that you don't need to add this anymore.

PR #4899, opened 2026-02-09
feat: Add support to send OTEL traces via OTLP

jamescrosswell:
Resolves: #4859
- #4859

## Problem

Sentry's .NET SDK has an existing OpenTelemetry integration (`Sentry.OpenTelemetry`) that works by acting as a *bridge* — it intercepts OTel spans via a custom SpanProcessor and converts them into Sentry's proprietary transaction/span format before sending them to Sentry's envelope endpoint. This works, but it means Sentry is translating OTel data into its own model, which adds complexity and loses fidelity. Meanwhile, Sentry now has a native OTLP (OpenTelemetry Protocol) endpoint that can ingest standard OTel traces directly — no translation needed.

## Solution

This PR introduces a new NuGet package, `Sentry.OpenTelemetry.Exporter`, that lets users send traces to Sentry via the standard OTLP protocol instead of Sentry's proprietary format. The key pieces are:

### New package with OTLP exporter wiring
`TracerProviderBuilderExtensions.cs` provides an `AddSentryOtlp(dsn)` extension method that configures the standard `OpenTelemetry.Exporter.OpenTelemetryProtocol` exporter to point at Sentry's OTLP traces endpoint (derived from the DSN). Users call `.AddSentryOtlp(dsn)` on their `TracerProviderBuilder` instead of the old `.AddSentry()`.

### Propagation context bridging
`OtelPropagationContext.cs` implements `IPropagationContext` by reading trace/span IDs from the current `Activity` (OTel's span representation) rather than from Sentry's internal state. `SentryPropagator.cs` ensures that Sentry's `sentry-trace` and `baggage` headers are still injected into outbound requests alongside the standard W3C `traceparent`, so that errors, replays, and other Sentry events can be linked to the correct OTLP trace.

### Disabling redundant Sentry tracing

A new `DisableSentryTracing` option provides a clean way to turn off Sentry's built-in span creation when OTel is handling it. To preserve backward compatibility, this is set to false when using the old Sentry.OpenTelemetry integration. It's disabled by default (i.e. set to `true`) when using the newer `Sentry.OpenTelemetry.Exporter` package via the `UseOtlp()` extension method on `SentryOptions` (disabling Sentry's automatic span creation from `SentryHttpMessageHandler`, `DiagnosticSource` listener, EF interception etc.) to avoid duplicate spans. When using OTLP you must instrument traces using OTEL packages rather than Sentry packages. 

### Samples and dependency updates
[...]

PR #4910, opened 2026-02-10
ci(danger): disable Danger check

Flash0ver:
Now with #4818, which should automatically generate `CHANGELOG` entries on release (via the PR title or a section in the PR description, we want to disable the changelog-validation of our [danger](https://github.com/getsentry/github-workflows/blob/main/danger/README.md) action / workflow.

For example, #4899 still asks us to `Please consider adding a changelog entry for the next release.`.
Also, we had to still manually update the CHANGELOG: see getsentry/publish#7429

For that, we need to update the [Danger Composite Action](https://github.com/getsentry/github-workflows/tree/main/danger).

This PR temporarily disabled the Danger workflow.

### DIFF
--- a/.github/workflows/danger.yml
+++ b/.github/workflows/danger.yml
@@ -11,6 +11,7 @@ permissions:
 
 jobs:
   danger:
+    if: false
     runs-on: ubuntu-latest
     steps:
       - uses: getsentry/github-workflows/danger@747c4c2906d373f5cd809abe94a7fd732a2421a7 # 3.2.1


jamescrosswell:
It doesn't look like there's any way to configure the [danger action](https://github.com/getsentry/github-workflows/blob/main/danger/action.yml) to skip the changelog check.

They [don't appear to be using this workflow](https://github.com/getsentry/sentry-python/tree/master/.github/workflows) at all in the Python repository. I think we need to remove it.

The only question then is if we can recreate some of the checks that were actually valuable from that workflow. 3 of [these 4 checks](https://github.com/getsentry/github-workflows/blame/fb91df38f94d5953e45f764307f69fec275342c7/danger/dangerfile.js#L230-L235) are probably still useful.

We could copy [this action](https://github.com/getsentry/github-workflows/blob/fb91df38f94d5953e45f764307f69fec275342c7/danger/action.yml) into our own repo so that we could use our own `dangerfile.js`... which would also be 99% copied from [here](https://github.com/getsentry/github-workflows/blame/main/danger/dangerfile.js#L230-L235). That seems kinda daft though. Ideally we would tweak the original action so that it accepted another parameter (`skip-changelog-check` or something like that) which it could pass to the danger check [here](https://github.com/getsentry/github-workflows/blob/fb91df38f94d5953e45f764307f69fec275342c7/danger/action.yml#L92). I'm not sure if that's possible... would need to investigate the docs for [danger](https://danger.systems/js/guides/getting_started#creating-a-dangerfile).

jamescrosswell (approved):

PR #4911, opened 2026-02-10
chore: Use SLNX format for solution files

jamescrosswell:
Resolves #4040:
- #4040

#skip-changelog

PR #4918, opened 2026-02-12
ci: move from dotnet-format tool to .NET SDK command

Flash0ver:
### Summary
- Remove `dotnet-format` global tool as it is now included in the .NET SDK
- Additionally, upload a `--report`, similar to build logs and test results
- Merge all three `--exclude` Options into a single space-separated list

### Remarks

[dotnet-format](https://www.nuget.org/packages/dotnet-format) has been marked _deprecated_ on NuGet.org.
It is included as a CLI command [since the the .NET 6.0 SDK](https://github.com/dotnet/format/issues/1268).

[dotnet format](https://learn.microsoft.com/dotnet/core/tools/dotnet-format) supports a `--report` option, that this changeset is uploading as a non-zipped artifact (see #4966).

Unfortunately, this does not fix the automatic removal of an Assertion (see #4911).
`--exclude ./test/Sentry.Tests/AttributeReaderTests.cs` is still required ... this may be a bug in a Diagnostic-Analyzer if `xUnit.net` and it's associated Code-Fix-Provider.

This PR also merges all three `--exclude` options in to a single space-separated list ... see [dotnet format](https://learn.microsoft.com/dotnet/core/tools/dotnet-format).

### DIFF
--- a/.github/workflows/format-code.yml
+++ b/.github/workflows/format-code.yml
@@ -39,9 +39,9 @@ jobs:
 
       - name: Upload Format Report
         if: ${{ always() }}
-        uses: actions/upload-artifact@b7c566a772e6b6bfb58ed0dc250532a479d7789f # v6.0.0
+        uses: actions/upload-artifact@bbbca2ddaa5d8feaa63e36b76fdaad77386f024f # v7.0.0
         with:
-          name: dotnet-format-report
+          archive: false
           path: ./dotnet-format-report/
           if-no-files-found: ignore
 


.github/workflows/format-code.yml
```diff
@@ -28,17 +28,28 @@ jobs:
...
+        run: dotnet format Sentry.slnx --no-restore --report ./dotnet-format-report --exclude ./modules ./**/*OptionsSetup.cs ./test/Sentry.Tests/AttributeReaderTests.cs
+
+      - name: Upload Format Report
+        if: ${{ always() }}
+        uses: actions/upload-artifact@bbbca2ddaa5d8feaa63e36b76fdaad77386f024f # v7.0.0
+        with:
+          archive: false
+          path: ./dotnet-format-report/
+          if-no-files-found: ignore
+
```
Flash0ver, line 47: note: see e.g. https://github.com/getsentry/sentry-dotnet/actions/runs/22679547094?pr=4918

I do like the unzipped upload for single files, so that in this case I can view the `.json` file directly in the browser.

.github/workflows/format-code.yml
```diff
@@ -28,17 +28,28 @@ jobs:
...
+      - name: Exclude Format Report from Git
+        run: |
+          if [ -d ./dotnet-format-report ]; then
+            echo '*' > ./dotnet-format-report/.gitignore
+          fi
```
Flash0ver, line 53: note: so that the `Commit Formatted Code` step does not consider the generated report as a change

jamescrosswell (approved):
lgtm

.github/workflows/format-code.yml
```diff
@@ -28,17 +28,28 @@ jobs:
...
-        run: dotnet format Sentry.slnx --no-restore --exclude ./modules --exclude ./**/*OptionsSetup.cs --exclude ./test/Sentry.Tests/AttributeReaderTests.cs
+        run: dotnet format Sentry.slnx --no-restore --report ./dotnet-format-report --exclude ./modules ./**/*OptionsSetup.cs ./test/Sentry.Tests/AttributeReaderTests.cs
+
+      - name: Upload Format Report
+        if: ${{ always() }}
+        uses: actions/upload-artifact@bbbca2ddaa5d8feaa63e36b76fdaad77386f024f # v7.0.0
+        with:
+          archive: false
+          path: ./dotnet-format-report/
+          if-no-files-found: ignore
```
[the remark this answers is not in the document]
Flash0ver: my intention was that CI should fail if anything goes unexpected ... starting out more aggressive ... we can always loosen strictness later

PR #4919, opened 2026-02-12
test(android): Use volatile to produce SIGSEGV in native crash test

jpnurmi:
The null-pointer dereference was undefined behavior that the compiler optimized into a trap instruction, producing `SIGILL` (x86_64) or `SIGTRAP` (arm64) instead of a real crash. Adding volatile forces an actual memory access, resulting in a consistent `SIGSEGV` across architectures.

Minor CI tweaks were needed because the supplemental lib hadn’t changed since `win|linux-arm64` builds were added.

### DIFF
--- a/.github/actions/buildnative/action.yml
+++ b/.github/actions/buildnative/action.yml
@@ -8,7 +8,7 @@ runs:
     - name: Set Environment Variables
       shell: bash
       run: |
-        if [[ "$RUNNER_OS" == "macOS" ]]; then
+        if [[ "$RUNNER_ARCH" == "ARM64" ]]; then
           echo "JAVA_HOME_11=$JAVA_HOME_11_ARM64" >> $GITHUB_ENV
         else
           echo "JAVA_HOME_11=$JAVA_HOME_11_X64" >> $GITHUB_ENV


.github/actions/buildnative/action.yml
```diff
@@ -8,7 +8,7 @@ runs:
     - name: Set Environment Variables
       shell: bash
       run: |
-        if [[ "$RUNNER_OS" == "macOS" ]]; then
+        if [[ "$RUNNER_ARCH" == "ARM64" ]]; then
```
Flash0ver, line 11: question: impact

May this have some effects on any non-testing artifacts that we ship via our NuGet packages?

.github/workflows/build.yml
```diff
@@ -197,7 +197,7 @@ jobs:
           enableCrossOsArchive: true
 
       - name: Build Native Dependencies
-        if: ${{ !matrix.container }}
+        if: ${{ !matrix.container && !env.NO_MOBILE }}
```
Flash0ver, line 200: question: non-mobile sentry-native

We also got sentry-native for non-mobile NativeAOT builds ... is there something we need to consider?

lib/sentrysupplemental/bin/arm64-v8a/libsentrysupplemental.so
Flash0ver, line 1: question: committed

Are these binaries intended to be checked in?

(
message for future @Flash0ver as reference
- `.a` - linked statically
- `.so` - linked dynamically during runtime

thanks @jpnurmi for explaining this to me the other day 🥔
)

jpnurmi: They are manually [built](https://github.com/getsentry/sentry-dotnet/blob/main/lib/sentrysupplemental/build.sh) and checked in, probably because they rarely change.

.github/workflows/build.yml
jpnurmi: The naming is a bit confusing (predates NativeAOT?), but the [buildnative](https://github.com/getsentry/sentry-dotnet/blob/main/.github/actions/buildnative/action.yml) action only builds native C and Java crash helpers for Android. It doesn't actually have anything to do with sentry-native or NativeAOT.

.github/actions/buildnative/action.yml
jpnurmi: This shouldn't affect the "publishing" macOS build (M1, ARM64).

PR #4936, opened 2026-02-19
chore: Refactor SentryAttributes into a separate class

jamescrosswell:
Before closing #4912 I figured we could extract some of the refactorings that might be useful in other PRs dealing with Attributes. Basically there's some functionality that is only accessible in the `SentryMetrics` class at the moment, but which will be needed by any other classes that also have Attributes.

#skip-changelog


test/Sentry.Testing/SentryAttributesExtensions.cs
```diff
@@ -0,0 +1,16 @@
...
+    {
+        attributes.TryGetAttribute<T>(key, out var value).Should().BeTrue();
+        value.Should().Be(expected);
+    }
+
+    internal static void ShouldNotContain<T>(this SentryAttributes attributes, string key, T expected)
+    {
+        attributes.TryGetAttribute<T>(key, out var value).Should().BeFalse();
+        value.Should().Be(default(T));
+    }
```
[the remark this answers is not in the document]
jamescrosswell: Fair, I've removed the parameter.

src/Sentry/SentryMetric.cs
```diff
@@ -114,6 +112,8 @@ private protected SentryMetric(DateTimeOffset timestamp, SentryId traceId, Sentr
     /// </remarks>
     public string? Unit { get; init; }
 
+    internal SentryAttributes Attributes { get; }
+
```
Flash0ver, line 116: suggestion: Logs

Would you mind applying this change to Logs, too, so that we're consistent.

src/Sentry/SentryLog.cs
```diff
@@ -115,78 +115,27 @@ internal SentryLog(DateTimeOffset timestamp, SentryId traceId, SentryLogLevel le
...
-        {
-            value = (string)attribute.Value;
-            return true;
-        }
-
-        value = null;
-        return false;
-    }
+    internal bool TryGetAttribute(string key, [NotNullWhen(true)] out string? value) =>
+        _attributes.TryGetAttribute(key, out value);
```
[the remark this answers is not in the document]
jamescrosswell: @Flash0ver what do you think about this? Do you want to restore the previous code/behaviour or OK with this?

jamescrosswell:
@Flash0ver is there any reason not to approve/merge this PR? The changes are internal only (so we can easily refactor again later if we need) and all the tests pass. Can we not just get this out of the way so we can move on to other stuff? It's been waiting for a review for over 2 months now.

src/Sentry/Protocol/SentryAttributes.cs
```diff
@@ -0,0 +1,175 @@
...
+    ///     <description><c>ToString</c> as <c>"type": "string"</c></description>
+    ///   </item>
+    ///   <item>
+    ///     <term><see langword="null"/></term>
+    ///     <description>ignored</description>
+    ///   </item>
+    /// </list>
+    /// </remarks>
+    /// <seealso href="https://develop.sentry.dev/sdk/telemetry/metrics/"/>
+    public bool TryGetAttribute<TAttribute>(string key, [MaybeNullWhen(false)] out TAttribute value)
```
Flash0ver, line 65: suggestion: remove Metrics-specific comment from shared Attributes-implementation

The usages reachable from `public` methods on both `SentryLog` and `SentryMetric` do yield that comment already.

src/Sentry/Protocol/SentryAttributes.cs
```diff
@@ -0,0 +1,175 @@
...
+        }
+
+        value = default;
+        return false;
+    }
+
+    /// <summary>
+    /// Set a key-value pair of data attached to the metric.
+    /// </summary>
+    public void SetAttribute<TAttribute>(string key, TAttribute value) where TAttribute : notnull
```
Flash0ver, line 80: suggestion: remove Metrics-specific comment from shared Attributes-implementation

The usages reachable from `public` methods on both `SentryLog` and `SentryMetric` do yield that comment already.

src/Sentry/Protocol/SentryAttributes.cs
```diff
@@ -0,0 +1,175 @@
...
+    {
+        writer.WriteStartObject();
+
+        foreach (var attribute in this)
+        {
+            SentryAttributeSerializer.WriteAttribute(writer, attribute.Key, attribute.Value, logger);
+        }
+
+        writer.WriteEndObject();
+    }
```
Flash0ver, line 169: suggestion: remove the `WriteTo(Utf8JsonWriter, IDiagnosticLogger?)` method

And remove the Attribute
```diff
-internal class SentryAttributes : Dictionary<string, SentryAttribute>, ISentryJsonSerializable
+internal class SentryAttributes : Dictionary<string, SentryAttribute>
```

Because the implementation is actually not shared between `SentryLog` and `SentryMetric`.
Where `SentryMetric` is actually calling this method,
but `SentryLog` is not and cannot, as there is other data that is encoded in the `attributes` JSON-property of the payload.

So I do find this abstraction, that is only called from a single location a bit confusing.

Non-blocking care factor, though, to be honest.

src/Sentry/SentryLog.cs
```diff
@@ -14,7 +14,7 @@ namespace Sentry;
 [DebuggerDisplay(@"SentryLog \{ Level = {Level}, Message = '{Message}' \}")]
 public sealed class SentryLog
 {
-    private readonly Dictionary<string, SentryAttribute> _attributes;
+    private readonly SentryAttributes _attributes;
```
Flash0ver: suggestion: change to property ...

... and locate this `internal` property after the `SpanId` property.

So that the implementation is in sync with `SentryMetric`.
Alternative: change property of `SentryMetric` to field as well.

Justification: I'd like to keep both `SentryLog` and `SentryMetric` implementation in sync, to make upcoming changes (like Global- and Scope-Attributes) simpler.

This way, I believe we could also remove some `internal` methods on this type (`SentryLog`):
- `internal bool TryGetAttribute(string key, [NotNullWhen(true)] out string? value)`
- `internal void SetAttribute(string key, string value)`
- `internal void SetAttribute(string key, char value)`
- `internal void SetAttribute(string key, int value)`

As these methods are used internally only, and/or for testing only, so we can "delegate" to the actual implementation in the shared `SentryAttributes` now.

This would also require an update to the tests (i.e. redirect to the `Attributes` property and use the `ShouldContain` extension method) ... but does then have both the `SentryMetric` and `SentryLog` implementation/usage in sync.

src/Sentry/SentryLog.cs
```diff
@@ -125,78 +125,27 @@ public ImmutableArray<KeyValuePair<string, object>> Parameters
...
-        {
-            SetAttribute("sentry.sdk.version", version);
-        }
-    }
-
-    internal void SetOrigin(string origin)
-    {
-        SetAttribute("sentry.origin", origin);
-    }
+    internal void SetOrigin(string origin) => SetAttribute("sentry.origin", origin);
```
Flash0ver: suggestion: call `string-overload` on attribute

Only if / after https://github.com/getsentry/sentry-dotnet/pull/4936/changes#r3217737068
So that we still call the internal `string`-overload on `SentryAttributes` for best performance.

```suggestion
    internal void SetOrigin(string origin) => Attributes.SetAttribute("sentry.origin", origin);
```

src/Sentry/SentryMetric.cs
```diff
@@ -301,14 +212,7 @@ internal void WriteTo(Utf8JsonWriter writer, IDiagnosticLogger? logger)
...
         writer.WritePropertyName("attributes");
-        writer.WriteStartObject();
-
-        foreach (var attribute in _attributes)
-        {
-            SentryAttributeSerializer.WriteAttribute(writer, attribute.Key, attribute.Value, logger);
-        }
-
-        writer.WriteEndObject();
+        Attributes.WriteTo(writer, logger);
```
Flash0ver, line 215: suggestion: revert

see also https://github.com/getsentry/sentry-dotnet/pull/4936/changes#r3217641003

As this abstraction of `SentryAttributes` is only used in `SentryMetric`, but not in `SentryLog`, I personally find this abstraction a bit "hasty".

test/Sentry.Tests/SentryMetricTests.cs
```diff
@@ -66,18 +66,12 @@ public void Protocol_Default_VerifyAttributes()
...
-        metric.TryGetAttribute<string>("sentry.sdk.version", out var version).Should().BeTrue();
-        version.Should().Be(sdk.Version);
-        metric.TryGetAttribute<object>("not-found", out var notFound).Should().BeFalse();
-        notFound.Should().BeNull();
+        metric.Attributes.ShouldContain("attribute", "value");
+        metric.Attributes.ShouldContain("sentry.environment", options.Environment);
+        metric.Attributes.ShouldContain("sentry.release", options.Release);
+        metric.Attributes.ShouldContain("sentry.sdk.name", sdk.Name);
+        metric.Attributes.ShouldContain("sentry.sdk.version", sdk.Version);
+        metric.Attributes.ShouldNotContain<string>("not-found");
```
Flash0ver: suggestion: change to `object`

To not change the previous semantics of the test.

```suggestion
        metric.Attributes.ShouldNotContain<object>("not-found");
```

test/Sentry.Tests/SentryMetricTests.cs
```diff
@@ -374,7 +368,7 @@ public void WriteTo_Attributes_AsJson()
 #else
         metric.SetAttribute("object", new KeyValuePair<string, string>("key", "value"));
 #endif
-        metric.SetAttribute("null", null!);
+        metric.Attributes.SetAttribute("null", null!);
```
Flash0ver: suggestion: call the `public` API, rather than the `internal` API

So that we have all "Arrange" methods go through the public API that the user could call, too.

```suggestion
        metric.SetAttribute<object>("null", null!);
```

Flash0ver (commented):
No reason ... I was just slow ... getting all the reviews done now.

I love the refactoring ❤️.
This removes duplicate code, and gives even more control to testing.

I got a couple of suggestions to make the implementation and usage aligned across `SentryMetric` and `SentryLog`.

src/Sentry/SentryLog.cs
```diff
@@ -125,78 +125,27 @@ public ImmutableArray<KeyValuePair<string, object>> Parameters
...
-    public void SetAttribute(string key, object value)
-    {
-        _attributes[key] = new SentryAttribute(value);
-    }
-
-    internal void SetAttribute(string key, string value)
-    {
-        _attributes[key] = new SentryAttribute(value, "string");
-    }
+    public void SetAttribute(string key, object value) => _attributes.SetAttribute(key, value);
```
Flash0ver: thought: behavioral change

This is changing behavior when setting a `null`-value via `SentryLog.SetAttribute("key", null!)` (e.g. via `SetBeforeSendLog`):

- previous
  - add to Attributes of Logs, but drop later during serialization emitting a Diagnostic-Log
  - https://github.com/getsentry/sentry-dotnet/blob/eadf5f20b8014745d0fe42b0e707bf3f9eec9ffc/src/Sentry/Protocol/SentryAttribute.cs#L42-L46
- with this change
  - does not add the the Attributes of Logs, without any Diagnostic-Log
  - https://github.com/getsentry/sentry-dotnet/blob/0e0505d9687563dc8029d8bd80fec68e2fa67172/src/Sentry/Protocol/SentryAttributes.cs#L80-L88

However ... I just noticed that the mistake was on my end, and your changes are actually consolidating the behavior to be the same in `SentryLog` and `SentryMetrics` ... which I like a lot!

How should the behavior be:
- both: do not even add `null`-Value-Attributes to the `SentryAttributes` collection
- both: add them, but drop them when serializing and emitting a diagnostic log
- keep: `SentryLog` drops during serialization, `SentryMetric` does not event add to collection

What is sent to Sentry would be the same, in both cases.

src/Sentry/SentryMetric.cs
jamescrosswell: It was originally used in these places as well:
https://github.com/getsentry/sentry-dotnet/blob/28f0b7527bcfca1d7bb522b7de196933ac637242/src/Sentry/Protocol/Spans/SentryLink.cs#L27
https://github.com/getsentry/sentry-dotnet/blob/28f0b7527bcfca1d7bb522b7de196933ac637242/src/Sentry/Protocol/Spans/SpanV2.cs#L112

So it seemed like a logical refactoring... This whole refactoring would be premature if we hadn't already tried to use it in #4912.

We can revert this if you really want, but I'm pretty sure we're just going to bring it back in again as soon as we build the next thing that has attributes. I think it just makes the code a bit more readable - one descriptively named method call instead of 9 lines of code (whose purpose you have to reverse engineer).

src/Sentry/Protocol/SentryAttributes.cs
jamescrosswell: I think this is the same conversation as below:
https://github.com/getsentry/sentry-dotnet/pull/4936#discussion_r3222996123

src/Sentry/SentryLog.cs
jamescrosswell: Applied in [c48cbd8](https://github.com/getsentry/sentry-dotnet/pull/4936/commits/c48cbd85dd848a6beb19554f425f6c3258513de9)

src/Sentry/SentryLog.cs
jamescrosswell: Done in the same commit.

src/Sentry/SentryLog.cs
jamescrosswell: Done in c48cbd85.

src/Sentry/Protocol/SentryAttributes.cs
jamescrosswell: Done in 2ed634f1.

src/Sentry/Protocol/SentryAttributes.cs
jamescrosswell: Done in 2ed634f1.

src/Sentry/SentryLog.cs
jamescrosswell: Done in c48cbd85.

test/Sentry.Tests/SentryMetricTests.cs
jamescrosswell: Done in bd885f8f.

test/Sentry.Tests/SentryMetricTests.cs
jamescrosswell: Done in bd885f8f.

src/Sentry/SentryLog.cs
jamescrosswell: I think it's good for them to be consistent. 

To retain the diagnostic logs, I've removed this:
https://github.com/getsentry/sentry-dotnet/blob/2ed634f16c747da6b31f3517b00bdbe67f895946/src/Sentry/Protocol/SentryAttributes.cs#L82-L85

Done in [7d0a860](https://github.com/getsentry/sentry-dotnet/pull/4936/commits/7d0a86025aff7ab9e156d0efb350bd11f74b0c06)

src/Sentry/SentryLog.cs
```diff
@@ -125,78 +125,18 @@ public ImmutableArray<KeyValuePair<string, object>> Parameters
...
-    public void SetAttribute(string key, object value)
-    {
-        _attributes[key] = new SentryAttribute(value);
-    }
-
-    internal void SetAttribute(string key, string value)
-    {
-        _attributes[key] = new SentryAttribute(value, "string");
-    }
+    public void SetAttribute(string key, object value) => Attributes.SetAttribute(key, value);
```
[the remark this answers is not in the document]
jamescrosswell: Duplicate of this conversation:
https://github.com/getsentry/sentry-dotnet/pull/4936#discussion_r3218513719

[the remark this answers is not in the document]
jamescrosswell: Fixed in [7d0a860](https://github.com/getsentry/sentry-dotnet/pull/4936/commits/7d0a86025aff7ab9e156d0efb350bd11f74b0c06)

src/Sentry/Protocol/SentryAttributes.cs
```diff
@@ -0,0 +1,170 @@
...
+
+    public SentryAttributes(int capacity) : base(capacity, StringComparer.Ordinal)
+    {
+    }
+
+    /// <summary>
+    /// Gets the attribute value associated with the specified key.
+    /// </summary>
+    /// <remarks>
+    /// Returns <see langword="true"/> if this contains an attribute with the specified key which is of type <typeparamref name="TAttribute"/> and it's value is not <see langword="null"/>.
```
Flash0ver: suggestion: consistent wording across XML-comments

Earlier, I actually meant to delete the entire XML-comment altogether.
But I guess it's fine to keep.
A small suggestion to keep the wording consistent:

```suggestion
    /// Returns <see langword="true"/> if this attribute collection contains an attribute with the specified key which is of type <typeparamref name="TAttribute"/> and it's value is not <see langword="null"/>.
```

src/Sentry/Protocol/SentryAttributes.cs
```diff
@@ -0,0 +1,170 @@
...
+    ///     <term>Collections</term>
+    ///     <description><c>ToString</c> as <c>"type": "string"</c></description>
+    ///   </item>
+    ///   <item>
+    ///     <term><see langword="null"/></term>
+    ///     <description>ignored</description>
+    ///   </item>
+    /// </list>
+    /// </remarks>
+    /// <seealso href="https://develop.sentry.dev/sdk/telemetry/logs/"/>
```
Flash0ver: suggestion: avoid specifics about Logs/Metrics

Earlier, I intended to suggest to delete the entire XML-comment, since this comment already exists on the `public`ly accessible members ... but it's actually fine to keep ... just a small suggestion to link to the actual Attribute-docs:

```suggestion
    /// <seealso href="https://develop.sentry.dev/sdk/foundations/state-management/scopes/attributes/"/>
```

test/Sentry.Tests/SentryLogTests.cs
```diff
@@ -68,18 +68,14 @@ public void Protocol_Default_VerifyAttributes()
         log.SpanId.Should().Be(SpanId);
 
         // should only show up in sdk integrations
-        log.TryGetAttribute("sentry.origin", out object origin).Should().BeFalse();
+        log.Attributes.ShouldNotContain<string>("sentry.origin");
 
         log.TryGetAttribute("attribute", out object attribute).Should().BeTrue();
         attribute.Should().Be("value");
```
Flash0ver: suggestion: consistent assertions

```suggestion
        log.Attributes.ShouldContain<string>("attribute", "value");
```

test/Sentry.Tests/SentryLogTests.cs
```diff
@@ -68,18 +68,14 @@ public void Protocol_Default_VerifyAttributes()
...
-        log.TryGetAttribute("sentry.sdk.name", out string name).Should().BeTrue();
-        name.Should().Be(sdk.Name);
-        log.TryGetAttribute("sentry.sdk.version", out string version).Should().BeTrue();
-        version.Should().Be(sdk.Version);
+        log.Attributes.ShouldContain<string>("sentry.environment", options.Environment!);
+        log.Attributes.ShouldContain<string>("sentry.release", options.Release!);
+        log.Attributes.ShouldContain<string>("sentry.sdk.name", sdk.Name!);
+        log.Attributes.ShouldContain<string>("sentry.sdk.version", sdk.Version!);
         log.TryGetAttribute("not-found", out object notFound).Should().BeFalse();
         notFound.Should().BeNull();
```
Flash0ver: suggestion: consistent assertions

```suggestion
        log.Attributes.ShouldNotContain<object>("not-found");
```

test/Sentry.Tests/SentryLogTests.cs
```diff
@@ -68,18 +68,14 @@ public void Protocol_Default_VerifyAttributes()
...
-        log.TryGetAttribute("sentry.release", out string release).Should().BeTrue();
-        release.Should().Be(options.Release);
-        log.TryGetAttribute("sentry.sdk.name", out string name).Should().BeTrue();
-        name.Should().Be(sdk.Name);
-        log.TryGetAttribute("sentry.sdk.version", out string version).Should().BeTrue();
-        version.Should().Be(sdk.Version);
+        log.Attributes.ShouldContain<string>("sentry.environment", options.Environment!);
+        log.Attributes.ShouldContain<string>("sentry.release", options.Release!);
+        log.Attributes.ShouldContain<string>("sentry.sdk.name", sdk.Name!);
+        log.Attributes.ShouldContain<string>("sentry.sdk.version", sdk.Version!);
```
Flash0ver: suggestion: remove `!`

The null-forgiving is not necessary here.

```suggestion
        log.Attributes.ShouldContain<string>("sentry.environment", options.Environment);
        log.Attributes.ShouldContain<string>("sentry.release", options.Release);
        log.Attributes.ShouldContain<string>("sentry.sdk.name", sdk.Name);
        log.Attributes.ShouldContain<string>("sentry.sdk.version", sdk.Version);
```

Flash0ver (approved):
✅ with some final set of nitpicks about comments and tests/assertions

src/Sentry/SentryMetric.cs
Flash0ver: Ah I see ... I haven't considered the origin this refactoring stems from.
And now I changed my mind ... let's keep the new `SentryAttributes` implementing `ISentryJsonSerializable` as proposed ✅.

src/Sentry/SentryLog.cs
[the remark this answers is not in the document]
Flash0ver: No ... don't restore ... I think it's OK as proposed.
The now dropped `internal` `string`-based `TryGetAttribute`-overload of `SentryLog` was - I believe - only used for testing anyway, and was a bit of a hasty/premature "we might need this in the future" fallacy from my end.
What I should have done instead is to introduce testing helpers, like the `test/Sentry.Testing/SentryAttributesExtensions.cs` you are adding in this PR ... which I do like a lot!

src/Sentry/Protocol/SentryAttributes.cs
Flash0ver: Indeed ... I wasn't considering the origin of this code consolidation ... I did now change my mind and would like to keep it as proposed.

src/Sentry/SentryLog.cs
Flash0ver: Cool ... so we - on top - also got a fix / alignment between Logs and Metrics ... love it!

src/Sentry/Protocol/SentryAttributes.cs
```diff
@@ -0,0 +1,170 @@
...
+        value = default;
+        return false;
+    }
+
+    /// <summary>
+    /// Set a key-value pair of data attached to the attribute collection.
+    /// </summary>
+    public void SetAttribute<TAttribute>(string key, TAttribute value) where TAttribute : notnull
+    {
+        this[key] = new SentryAttribute(value!);
```
Flash0ver: suggestion: remove null-suppression

I believe it's not necessary here.

```suggestion
        this[key] = new SentryAttribute(value);
```

src/Sentry/Protocol/SentryAttributes.cs
```diff
@@ -0,0 +1,170 @@
...
+            value = attributeValue;
+            return true;
+        }
+
+        value = default;
+        return false;
+    }
+
+    /// <summary>
+    /// Set a key-value pair of data attached to the attribute collection.
```
Flash0ver: suggestion: rephrase

This is the comment on `SentryLog`/`SentryMetric`, to which Attributes are attached ... but this collection type does represent the Attributes, so I suggest rephrasing it a bit to something like:

```suggestion
    /// Set a key-value pair in this attribute collection.
```

src/Sentry/Protocol/SentryAttributes.cs
jamescrosswell: Applied in 7699aaa5.

src/Sentry/Protocol/SentryAttributes.cs
jamescrosswell: Applied in 7699aaa5.

test/Sentry.Tests/SentryLogTests.cs
jamescrosswell: Applied in 7699aaa5.

test/Sentry.Tests/SentryLogTests.cs
jamescrosswell: Applied in 7699aaa5.

test/Sentry.Tests/SentryLogTests.cs
jamescrosswell: Applied in 7699aaa5.

src/Sentry/Protocol/SentryAttributes.cs
jamescrosswell: Applied in 7699aaa5.

src/Sentry/Protocol/SentryAttributes.cs
jamescrosswell: Applied in 7699aaa5.

PR #4942, opened 2026-02-23
fix: CaptureFeedback now applies event processors

### LINKED ISSUE #4808: CaptureFeedback, missing environment
### Package

Sentry

### .NET Flavor

.NET

### .NET Version

9.0

### OS

Android

### OS Version

15

### Development Environment

Visual Studio v18.x

### SDK Version

6.0.0

_No response_

### Workload Versions

Workload version: 10.0.101

Installed Workload Id      Manifest Version         Installation Source
-----------------------------------------------------------------------------------------------------
android                    36.1.2/10.0.100          SDK 10.0.100, VS 18.1.11304.174, VS 17.14.36804.6
ios                        26.1.10502/10.0.100      SDK 10.0.100, VS 18.1.11304.174, VS 17.14.36804.6
maccatalyst                26.1.10502/10.0.100      SDK 10.0.100, VS 18.1.11304.174, VS 17.14.36804.6
maui-android               10.0.1/10.0.100          SDK 10.0.100
maui-windows               10.0.1/10.0.100          SDK 10.0.100, VS 18.1.11304.174, VS 17.14.36804.6

```C#
        var builder = MauiApp.CreateBuilder();
        builder
            .SetupSentry()
            .UseMauiApp<App>()
```
...
```C#
        builder.UseSentry((option) =>
        {
            option.Native.ExperimentalOptions.SessionReplay.OnErrorSampleRate = 0.1;
            option.Native.ExperimentalOptions.SessionReplay.SessionSampleRate = 0.1;
            option.Native.ExperimentalOptions.SessionReplay.MaskAllText = false;
            option.Native.ExperimentalOptions.SessionReplay.MaskAllImages = false;
            option.ReportAssembliesMode = ReportAssembliesMode.None;
            option.Dsn = EnvironmentConfig.SentryDsn;
...
```

### Steps to Reproduce

initialize sentry, set "Prod" in environments.
capture any error, you will see it has Prod as environment in sentry.io
call
`SentrySdk.CaptureFeedback(new SentryFeedback("test", "test@test.com", "test", associatedEventId: SentrySdk.LastEventId));`

### Expected Result

Sentry.io with Prod on environment

### Actual Result

Image

jamescrosswell:
fixes #4808:
- #4808


src/Sentry/SentryClient.cs
```diff
@@ -117,12 +117,19 @@ public SentryId CaptureFeedback(SentryFeedback feedback, out CaptureFeedbackResu
             evt.Level = scope.Level;
         }
 
+        if (SentryEventHelper.ProcessEvent(evt, scope.GetAllEventProcessors(), hint, _options) is not { } processedEvent)
+        {
+            _options.LogWarning("Feedback dropped by event processor");
+            result = CaptureFeedbackResult.UnknownError;
+            return SentryId.Empty;  // Dropped by an event processor
+        }
```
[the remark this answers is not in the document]
jamescrosswell: Good catch - have fixed in [b2508fa](https://github.com/getsentry/sentry-dotnet/pull/4942/commits/b2508facb8d540f3609075ccbf4d389483cb6987)

src/Sentry/SentryClient.cs
```diff
@@ -117,12 +117,20 @@ public SentryId CaptureFeedback(SentryFeedback feedback, out CaptureFeedbackResu
             evt.Level = scope.Level;
         }
 
+        if (SentryEventHelper.ProcessEvent(evt, scope.GetAllEventProcessors(), hint, _options, DataCategory.Feedback)
+            is not { } processedEvent)
+        {
+            _options.LogWarning("Feedback dropped by event processor");
```
Flash0ver: question: log twice

If I'm reading it correctly, then `SentryEventHelper.ProcessEvent` already logs `SentryLevel.Info` with message `"Event dropped by processor {0}""` in case of an event discarded.

https://github.com/getsentry/sentry-dotnet/blob/b2508facb8d540f3609075ccbf4d389483cb6987/src/Sentry/Internal/SentryEventHelper.cs#L18-L27

Here, we log a `SentryLevel.Warning`, with a more specific `Feedback dropped by event processor"` message.

I'm a bit uncertain about logging twice,
and also with two different levels.

Should we remove this log message here?

src/Sentry/SentryClient.cs
```diff
@@ -117,12 +117,20 @@ public SentryId CaptureFeedback(SentryFeedback feedback, out CaptureFeedbackResu
...
+            result = CaptureFeedbackResult.UnknownError;
```
Flash0ver: suggestion: add `CaptureFeedbackResult.DroppedByEventProcessor` -or- a doc comment

`CaptureFeedbackResult.UnknownError` is a bit unspecific.
We could add a `DroppedByEventProcessor` value.

But - actually - now on second thought ... the user of the Sentry-monitored app cannot really react to a potential `DroppedByEventProcessor` and change the feedback so that it is captured ... well ... we also got `DisabledHub` to which the end user cannot really react either.

So perhaps - for consistency - we could add something like `CaptureFeedbackResult.DroppedByEventProcessor`.

### Alternative
Alternatively, we could document the behavior in the XML comment,
and should there later be an explicit user ask to add this constant value to the `CaptureFeedbackResult` enum, we can always add it back later ... but adding it later could kind of be considered a breaking change ... but on the other hand this very change is changing the behavior too and may now drop more feedback than previously.

https://github.com/getsentry/sentry-dotnet/blob/fed52443fe8ac0e074d7b0f0be9c815b4176ace0/src/Sentry/CaptureFeedbackErrorReason.cs#L25-L34

Long story short:
I suggest to either add an enum value,
or (my current preference) just add this behavior to the XML documentation comment by adding a new `item`.

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 ### Fixes
 
 - The SDK now logs a `Warning` instead of an `Error` when being ratelimited ([#4927](https://github.com/getsentry/sentry-dotnet/pull/4927))
+- Common tags such as `Environment` and `Release` are now correctly applied to CaptureFeedback events ([#4942](https://github.com/getsentry/sentry-dotnet/pull/4942))
```
Flash0ver: suggestion: Also mention "Event Processor" ... perhaps in a separate entry

Because users may also register Event Processors, that are now applied, too

```CSharp
options.AddEventProcessor(new MySentryEventProcessor());
options.AddEventProcessorProvider(() => [new MySentryEventProcessor()]);
```

test/Sentry.Tests/SentryClientTests.cs
```diff
@@ -976,6 +976,57 @@ public void CaptureFeedback_WithScope_ScopeCopiedToEvent()
         Assert.Equal(scope.Breadcrumbs, @event.Breadcrumbs);
     }
 
+    [Fact]
+    public void CaptureFeedback_EventProcessorApplied()
```
Flash0ver: suggestion: keep naming consistent


```suggestion
    public void CaptureFeedback_WithEventProcessor_EventProcessorApplied()
```

src/Sentry/SentryClient.cs
jamescrosswell:
```suggestion
```

jamescrosswell: Nice catch!

src/Sentry/SentryClient.cs
jamescrosswell: I'll add another enum.

CHANGELOG.md
```suggestion
- Common tags such as `Environment` and `Release` and custom event processors are all now correctly applied to CaptureFeedback events ([#4942](https://github.com/getsentry/sentry-dotnet/pull/4942))
```

test/Sentry.Tests/SentryClientTests.cs
```diff
@@ -976,6 +976,57 @@ public void CaptureFeedback_WithScope_ScopeCopiedToEvent()
...
+        var sut = _fixture.GetSut();
+
+        //Act
+        var result = sut.CaptureFeedback(feedback);
+
+        //Assert
+        result.Should().Be(SentryId.Empty);
+        var expectedReason = DiscardReason.EventProcessor;
+        _fixture.ClientReportRecorder.Received(1).RecordDiscardedEvent(expectedReason, DataCategory.Feedback);
+    }
```
Flash0ver, line 1029: suggestion: also assert the new `CaptureFeedbackResult.DroppedByEventProcessor`

```suggestion
        //Act
        var id = sut.CaptureFeedback(feedback, out var result);

        //Assert
        result.Should().Be(CaptureFeedbackResult.DroppedByEventProcessor);
        id.Should().Be(SentryId.Empty);
        var expectedReason = DiscardReason.EventProcessor;
        _fixture.ClientReportRecorder.Received(1).RecordDiscardedEvent(expectedReason, DataCategory.Feedback);
    }
```

Naming of local fields should be consistent with similar tests.

Flash0ver (commented):
One final suggestion.
Also - we now got a merge conflict ... sorry it took so long.

Issue #4939, opened 2026-02-20
When Application Includes a Refit Assembly, Many System.IO.IOExceptions Are Thrown


Mike-E-angelo:
### Package

Sentry

### .NET Flavor

.NET

### .NET Version

10.0.103

### OS

Windows

### OS Version

Windows 11

### Development Environment

Visual Studio v18.x

### SDK Version

6.1.0

### Workload Versions

Workload version: 10.0.200-manifests.796b3495

Installed Workload Id      Manifest Version         Installation Source
---------------------------------------------------------------------------------------
android                    36.1.30/10.0.100         VS 18.3.11512.155, VS 18.4.11513.90
ios                        26.2.10197/10.0.100      VS 18.3.11512.155, VS 18.4.11513.90
maccatalyst                26.2.10197/10.0.100      VS 18.3.11512.155, VS 18.4.11513.90
maui-windows               10.0.20/10.0.100         VS 18.3.11512.155, VS 18.4.11513.90

### UseSentry or SentrySdk.Init call

"Serilog": {
  "WriteTo": [
    {
      "Name": "Sentry",
      "Args": {
        "enableLogs": true,
        "environment": "Development",
        "minimumEventLevel": "Debug",
        "minimumBreadcrumbLevel": "Debug"
      }
    }
  ]
},

### Steps to Reproduce

When using Sentry 6.1.0 with a Refit assembly, System.IO.IOExceptions are thrown due to attempting to load the assembly, but it is in memory:

<img width="1349" height="464" alt="Image" src="https://github.com/user-attachments/assets/6912dc1d-3e62-46ad-9cc5-8b9e85738e76" />

### Expected Result

No exceptions are thrown

### Actual Result

Many exceptions are thrown and caught, leaving messages in the debug window:

<img width="646" height="364" alt="Image" src="https://github.com/user-attachments/assets/9ad5521a-3c42-4850-aa62-1ba3a1c08bdc" />

jamescrosswell:
@Mike-E-angelo I'm not familiar with Refit. Are you able to provide a small reproduction of the issue (e.g. as a github repository) so that we can investigate?

Mike-E-angelo:
Hi @jamescrosswell pardon my confusion here but this was not Refit after all.  The error said `RefEmit` which is something completely different. 🤦  The issue has to do with dynamic-created/compiled delegates and I have a reproduction for you below:

```
public sealed class Program
{
[...]

PR #4954, opened 2026-02-24
perf: DebugStackTrace no longer creates a stack trace for dynamic methods

jamescrosswell:
resolves: #4939

#skip-changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,10 +2,6 @@
 
 ## Unreleased
 
-### Features
-
-- DebugStackTrace no longer creates a stack trace for dynamic methods, which should result in slightly better performance when capturing errors ([#4954](https://github.com/getsentry/sentry-dotnet/pull/4954))
-
 ### Fixes
 
 - The SDK now logs a `Warning` instead of an `Error` when being ratelimited ([#4927](https://github.com/getsentry/sentry-dotnet/pull/4927))


src/Sentry/Internal/DebugStackTrace.cs
```diff
@@ -532,14 +532,18 @@ private static void DemangleLambdaReturnType(SentryStackFrame frame)
                 return reader.Invoke(assemblyName);
             }
 
-            var assembly = options.FileSystem.OpenFileForReading(assemblyName);
-            return new PEReader(assembly);
+            if (options.FileSystem.FileExists(assemblyName))
```
Flash0ver, line 535: question: I guess it's not too easy - or better to say "economic" - to add a unit test for this, right?

CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 ## Unreleased
 
+### Features
```
Flash0ver: thought: not sure if `Features` is the right section

I have the feeling that we list Performance-related changes usually as `Fixes`,
as it's not really a new feature that the user can observe ... well ... in this case actually we can (as described in the issue)
(although: there is the saying "Performance as a Feature" 😉)

I am a bit uncertain on this.

CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
...
+- DebugStackTrace no longer creates a stack trace for dynamic methods, which should result in slightly better performance when capturing errors ([#4954](https://github.com/getsentry/sentry-dotnet/pull/4954))
+
```
Flash0ver: suggestion: insert new line between section and first list item for consistency


```suggestion
### Features

- DebugStackTrace no longer creates a stack trace for dynamic methods, which should result in slightly better performance when capturing errors ([#4954](https://github.com/getsentry/sentry-dotnet/pull/4954))

```

src/Sentry/Internal/DebugStackTrace.cs
jamescrosswell: It could possibly be done. This isn't really a functional change though (it doesn't change the behaviour of our SDK)... it's more of a performance change. We're avoiding using exceptions for control flow here and the need to create stack traces for those exceptions. 

Given that it's not a functional change, I think a unit test would be overkill... And we could create a benchmark but, again, kind of pointless since we know it's going to be quicker than the previous/alternative code.

CHANGELOG.md
jamescrosswell: I've removed the changelog entry entirely... it'd probably not really a customer facing change (even though it was triggered by a customer conversation).

PR #4960, opened 2026-02-26
fix: libmonosgen and libxamarin frames currently showing as inapp

### LINKED ISSUE #4194: Mark mono/xamarin/dotnet system symbols as not inApp

jamescrosswell:
fixes #4194 
- #4194

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@
 ### Fixes
 
 - The SDK now logs a `Warning` instead of an `Error` when being ratelimited ([#4927](https://github.com/getsentry/sentry-dotnet/pull/4927))
-- libmonosgen and libxamarin frames no longer show as inapp ([#4960](https://github.com/getsentry/sentry-dotnet/pull/4960))
+- `libmonosgen` and `libxamarin` frames no longer show as in-app ([#4960](https://github.com/getsentry/sentry-dotnet/pull/4960))
 
 ## 6.2.0-alpha.0
 
--- a/src/Sentry/Internal/InAppExcludeRegexes.cs
+++ b/src/Sentry/Internal/InAppExcludeRegexes.cs
@@ -21,14 +21,14 @@ internal static partial class InAppExcludeRegexes
 #elif NET8_0
     [GeneratedRegex(LibMonoSgenPattern)]
     private static partial Regex LibMonoSgenRegex();
-    internal static readonly Regex LibMonoSgen = LibMonoSgenRegex();
+    internal static Regex LibMonoSgen { get; } = LibMonoSgenRegex();
 
     [GeneratedRegex(LibXamarinPattern)]
     private static partial Regex LibXamarinRegex();
-    internal static readonly Regex LibXamarin = LibXamarinRegex();
+    internal static Regex LibXamarin { get; } = LibXamarinRegex();
 #else
-    internal static readonly Regex LibMonoSgen = new(LibMonoSgenPattern, RegexOptions.Compiled);
-    internal static readonly Regex LibXamarin = new(LibXamarinPattern, RegexOptions.Compiled);
+    internal static Regex LibMonoSgen { get; } = new(LibMonoSgenPattern, RegexOptions.Compiled);
+    internal static Regex LibXamarin { get; } = new(LibXamarinPattern, RegexOptions.Compiled);
 #endif
 }
 


test/Sentry.Tests/Protocol/Exceptions/SentryStackFrameTests.cs
```diff
@@ -239,6 +239,46 @@ public void ConfigureAppFrame_WithDefaultOptions_NotBuiltInIgnoredMarkedAsInApp(
...
+        {
+            Function = "async void MainActivity.OnCreate(Bundle savedInstanceState)+(?) =\\u003E { }",
+            Package = package
+        };
+
+        // Act
+        sut.ConfigureAppFrame(options);
+
+        // Assert
+        Assert.Equal(sut.InApp, expectedInApp);
```
jamescrosswell:
```suggestion
        Assert.Equal(expectedInApp, sut.InApp);
```

test/Sentry.Tests/Protocol/Exceptions/SentryStackFrameTests.cs
```diff
@@ -239,6 +239,46 @@ public void ConfigureAppFrame_WithDefaultOptions_NotBuiltInIgnoredMarkedAsInApp(
...
```
jamescrosswell:
```suggestion
        Assert.Equal(expectedInApp, sut.InApp);
```

src/Sentry/Internal/InAppExcludeRegexes.cs
```diff
@@ -0,0 +1,34 @@
...
+    internal static partial Regex LibXamarin { get; }
+#elif NET8_0
+    [GeneratedRegex(LibMonoSgenPattern)]
+    private static partial Regex LibMonoSgenRegex();
+    internal static readonly Regex LibMonoSgen = LibMonoSgenRegex();
+
+    [GeneratedRegex(LibXamarinPattern)]
+    private static partial Regex LibXamarinRegex();
+    internal static readonly Regex LibXamarin = LibXamarinRegex();
+#else
```
Flash0ver, line 29: suggestion: change from _fields_ to _properties_

I'm a bit uncertain whether this could be an actual issue in some corner-case scenarios: the fields are `internal`, but should we invoke the fields in assemblies that the Internals are made Visible To, and the user have an app that is binding against a _< .NET 9_ Sentry SDK that is having the fields, but at runtime loading a _>= .NET 9_ Sentry SDK here these are now properties, the consuming application could run time `MemberAccessException` or `MissingMemberException` ... theoretically.

Not sure if this is a practical problem ... but my mind would be at ease to have the same API shape across TFMs.

```suggestion
#elif NET8_0
    [GeneratedRegex(LibMonoSgenPattern)]
    private static partial Regex LibMonoSgenRegex();
    internal static Regex LibMonoSgen { get; } = LibMonoSgenRegex();

    [GeneratedRegex(LibXamarinPattern)]
    private static partial Regex LibXamarinRegex();
    internal static Regex LibXamarin { get; } = LibXamarinRegex();
#else
```

src/Sentry/Internal/InAppExcludeRegexes.cs
```diff
@@ -0,0 +1,34 @@
...
+    internal static readonly Regex LibMonoSgen = new(LibMonoSgenPattern, RegexOptions.Compiled);
+    internal static readonly Regex LibXamarin = new(LibXamarinPattern, RegexOptions.Compiled);
+#endif
+}
```
Flash0ver: suggestion: use _properties_ over _fields_ to match the `net9.0`/`net10.0` TFM

again ... could be just an issue in theory, rather than actuality, but my mind would be at ease if the `internal` members would have the same shape across all TFMs due to our usage of `[InternalsVisibleToAttribute]`

```suggestion
    internal static Regex LibMonoSgen { get; } = new(LibMonoSgenPattern, RegexOptions.Compiled);
    internal static Regex LibXamarin { get; } = new(LibXamarinPattern, RegexOptions.Compiled);
#endif
}
```

src/Sentry/Internal/InAppExcludeRegexes.cs
```diff
@@ -0,0 +1,34 @@
...
+    // ([^/\\]*libxamarin[^/\\]*) — last segment contains libxamarin (no slashes before/after)
+    // $ — end of string
+    private const string LibXamarinPattern = @"(^|[/\\])([^/\\]*libxamarin[^/\\]*)$";
+
+#if NET9_0_OR_GREATER
+    [GeneratedRegex(LibMonoSgenPattern)]
+    internal static partial Regex LibMonoSgen { get; }
+
+    [GeneratedRegex(LibXamarinPattern)]
+    internal static partial Regex LibXamarin { get; }
```
Flash0ver, line 20: question: use `RegexOptions.ExplicitCapture` and/or `RegexOptions.CultureInvariant`

#### re `RegexOptions.ExplicitCapture`
When the Regex-Pattern includes parentheses (`(` and `)`), these are also "Capture Groups" per default. But not sure if we actually consume these automatically captures Groups ... where we may do some work that is not necessarily needed ... in other words: impacting performance slightly.

We have three `IStringOrRegexMatcher`
- `DelimitedPrefixOrPatternMatcher`
  - calls `Regex.Matches`
- `PrefixOrPatternMatcher`
  - calls `Regex.IsMatch`
- `SubstringOrPatternMatcher`
  - calls `Regex.IsMatch`

On the other hand ... although we are not actively using Capture Groups right now ... we might in the future and then the In-App detection might not work as expected.

I believe I'm actually talking myself out of this change now ... considering it's only about performance, not correctness.

#### re `RegexOptions.CultureInvariant`
Not sure if adding this option would actually be about correctness in our case ... perhaps also only performance.

---

I guess keeping the Regex-Options as they are (i.e. `RegexOptions.Compiled` for non-source-generated _Regex_) seems to be fine.

What do you think?

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 ### Fixes
 
 - The SDK now logs a `Warning` instead of an `Error` when being ratelimited ([#4927](https://github.com/getsentry/sentry-dotnet/pull/4927))
+- libmonosgen and libxamarin frames no longer show as inapp ([#4960](https://github.com/getsentry/sentry-dotnet/pull/4960))
```
Flash0ver: suggestion: a bit of markdown formatting to make the entry a bit easier to read

```suggestion
- `libmonosgen` and `libxamarin` frames no longer show as in-app ([#4960](https://github.com/getsentry/sentry-dotnet/pull/4960))
```

Also, I believe that in both the product and the documentation we use `in-app` over `inapp`.

PR #4977, opened 2026-03-03
feat: Add API to retrieve persisted breadcrumbs on Android

### LINKED ISSUE #5243: feat: Add API to retrieve persisted breadcrumbs on Android
### Summary

Add a user-accessible API to retrieve "persisted breadcrumbs" from the Android-SDK (`sentry-java`).

### Remarks

User Scenario: ANR
Does not use provided auto ANRs (environment always defaulted to prod, tracking app closures as ANR).
Set up custom solution, that is lacking the breadcrumb data from the previously crashed session.

Note: The Persisted Breadcrumb data, that the Android SDK stores, is not pure JSON, so a custom reader/parser is used.

Note: Retrieving the Persisted Breadcrumbs must be done before the default Integrations of the Android SDK override the previously Persisted Breadcrumbs. But actually during initialization, where the Sentry-Options are already available with the Path to the Cache-Directory.

#### Solution: Read during Init
- see getsentry/sentry-dotnet#4977
- At the end of the Java/Android-Options configuration callback, where we transfer .NET-SDK-Options to the Java/Android-SDK-Options, read the ReadPersisted Breadcrumbs into a static field that can be read by the user through a public member on `SentrySdk` on the Android Platform only.

Pro: We control the right timing of the read, have all internal types available to properly parse the data into a .NET type.
Con: we are pessimizing the Init/Startup of over .NET Android and .NET MAUI on Android app for this (currently) singular use case.

#### Solution: New static "Read" method with pay-for-play complexity
- see https://github.com/getsentry/sentry-dotnet/pull/4977/changes#r3273624965

Pro: separate method that can be invoked on an on-demand basis; could be marked `unsafe` considering [C# 15's Unsafe Evolution](https://github.com/dotnet/roslyn/blob/main/docs/Language%20Feature%20Status.md).
Con: required types are `internal`, and are not really meant to be constructed/owned by user code.

#### Solution: provide .NET integration
- `sentry-dotnet` implements a `io.sentry.Integration`

Pro: not added per default
Con: uncommon pattern of user code constructing the Integration in order to own a member via which the written Persisted Breadcrumbs can be read; the `sentry-dotnet` bindings would need to make two types public
- `io.sentry.SentryOptions` via `Sentry.JavaSdk.SentryOptions`
- `io.sentry.IScopes` via `Sentry.JavaSdk.IScopes`

#### Solution: provide a more generic way of Before/Pre-Init and After/Post-Init integration
- `sentry-java` exposes a generic integration layer

Pro: Persisted Breadcrumbs would just be one use case to use this interop layer
Con: Significant work required in `sentry-java` for a - currently - single use case

#### Solution: user reads file directly
- user reads the Persisted Breadcrumb data from Sentry as is

Pro: Sentry does not expose interop types/members for a (currently) singular scenario.
Con: brittle solution, as the user would need to duplicate/imitate internals (such as the format for the Persisted Breadcrumbs file as well as building the path to it) that might change in the future.

#### Solution: make some Android Bindings public to facilitate user code reading the file directly
- `sentry-dotnet` changes the accessibility of `io.sentry.cache.PersistingScopeObserver` from `internal` to `public`
- `sentry-dotnet` starts create bindings for `io.sentry.cache.tape.QueueFile` (currently remofed/excluded

Pro: exposing the already available building blocks to users.
Con: exposing the Android Bindings to .NET consumers introduces a new layer/vector of potential breaking changes.

romtsn:
closes #5243

## Summary
- Adds `SentrySdk.GetPersistedBreadcrumbs()` public API that returns breadcrumbs from the previous app session on Android
- Reads from the native Android SDK's `PersistingScopeObserver` during init and caches the result
- No internal binding types are exposed — customers get back .NET `Breadcrumb` objects

### Usage
```csharp
#if __ANDROID__
var breadcrumbs = SentrySdk.GetPersistedBreadcrumbs();
foreach (var breadcrumb in breadcrumbs)
{
    Console.WriteLine($"[{breadcrumb.Timestamp}] {breadcrumb.Category}: {breadcrumb.Message}");
}
#endif
```

## Test plan
- [ ] Verify breadcrumbs are read correctly on Android after a previous session with breadcrumbs
- [ ] Verify empty list is returned when no persisted breadcrumbs exist
- [ ] Verify the API is accessible from a MAUI sample app

### DIFF
--- a/src/Sentry/Platforms/Android/SentrySdk.cs
+++ b/src/Sentry/Platforms/Android/SentrySdk.cs
@@ -302,28 +302,35 @@ public static partial class SentrySdk
 
     private static List<Breadcrumb>? ReadPersistedBreadcrumbs(JavaSdk.SentryOptions nativeOptions)
     {
-        var observer = nativeOptions.FindPersistingScopeObserver();
-        if (observer is null)
+        try
         {
-            return null;
-        }
-
-        var result = observer.Read(nativeOptions, "breadcrumbs.json",
-            Java.Lang.Class.FromType(typeof(Java.Util.IList)));
-
-        if (result is not Java.Util.IList javaList)
-        {
-            return null;
-        }
-
-        var breadcrumbs = new List<Breadcrumb>();
-        for (var i = 0; i < javaList.Size(); i++)
-        {
-            if (javaList.Get(i)?.JavaCast<JavaSdk.Breadcrumb>() is { } javaBreadcrumb)
+            var observer = nativeOptions.FindPersistingScopeObserver();
+            if (observer is null)
             {
-                breadcrumbs.Add(javaBreadcrumb.ToBreadcrumb());
+                return null;
             }
+
+            var result = observer.Read(nativeOptions, "breadcrumbs.json",
+                Java.Lang.Class.FromType(typeof(Java.Util.IList)));
+
+            if (result is not Java.Util.IList javaList)
+            {
+                return null;
+            }
+
+            var breadcrumbs = new List<Breadcrumb>();
+            for (var i = 0; i < javaList.Size(); i++)
+            {
+                if (javaList.Get(i)?.JavaCast<JavaSdk.Breadcrumb>() is { } javaBreadcrumb)
+                {
+                    breadcrumbs.Add(javaBreadcrumb.ToBreadcrumb());
+                }
+            }
+            return breadcrumbs;
         }
-        return breadcrumbs;
+        catch (Java.Lang.Exception)
+        {
+            return null;
+        }
     }
 }
--- a/src/Sentry/Platforms/Android/SentrySdk.cs
+++ b/src/Sentry/Platforms/Android/SentrySdk.cs
@@ -173,6 +173,11 @@ public static partial class SentrySdk
             o.ConnectionStatusProvider =
                 new AndroidConnectionStatusProvider(AppContext, o, buildInfoProvider, timeProvider, mainHandler).JavaCast<IConnectionStatusProvider>();
             o.AddIntegration(new SystemEventsBreadcrumbsIntegration(AppContext, mainHandler).JavaCast<JavaSdk.IIntegration>());
+
+            // Read persisted breadcrumbs now, inside the options callback, before the SDK starts.
+            // This avoids a race condition with breadcrumb-producing integrations that start writing
+            // to the QueueFile as soon as SentryAndroid.Init() completes.
+            _persistedBreadcrumbs = ReadPersistedBreadcrumbs(o);
         });
 
         // Now initialize the Android SDK (with a logger only if we're debugging)
@@ -185,9 +190,6 @@ public static partial class SentrySdk
         {
             SentryAndroid.Init(AppContext, configuration);
         }
-
-        // Read persisted breadcrumbs from the previous session before they get overwritten
-        _persistedBreadcrumbs = ReadPersistedBreadcrumbs(nativeOptions!);
 
         // Set options for the managed SDK that depend on the Android SDK. (The user will not be able to modify these.)
         options.AddEventProcessor(new AndroidEventProcessor(nativeOptions!));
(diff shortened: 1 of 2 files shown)


src/Sentry/Platforms/Android/SentrySdk.cs
```diff
@@ -286,4 +293,42 @@ private static void AndroidEnvironment_UnhandledExceptionRaiser(object? _, Raise
...
+    {
+        if (nativeOptions.CacheDirPath is null)
+        {
+            return null;
+        }
+
+        // Create a temporary observer to read breadcrumbs from the previous session's QueueFile.
+        // This must be called before SentryAndroid.Init() completes, to avoid a race condition
+        // with breadcrumb-producing integrations that write to the same QueueFile concurrently.
+        using var observer = new JavaSdk.Cache.PersistingScopeObserver(nativeOptions);
```
Flash0ver, line 315: question: could the user code just read the file directly and/or find and read the `"breadcrumbs.json"` themselves?

We're a bit hesitant of adding an API for a single use case, that is also either pessimizing all other scenarios, or a bit of a "dangerous" API (order of invocations).

romtsn: nope, not easily -- the file has a special format and `QueueFile` is [excluded](https://github.com/getsentry/sentry-dotnet/blob/a4a8ded702be42a023b23f2319d53d2dc3c2228e/src/Sentry.Bindings.Android/Transforms/Metadata.xml#L167) from the bindings, plus, some necessary classes are package-private.  Though, theoretically, exposing the QueueFile should be enough -- the rest can be achieved with a bit more code on the user's side

jamescrosswell: Maybe a middle ground, we could expose the `ReadPersistedBreadcrumbs` method publicly so users can call this themselves (before initialising the Android SDK) and dispose of the breadcrumbs once they're done with them. 

We wouldn't then call this automatically and store the results for the lifetime of the application in `_persistedBreadcrumbs` (even users that don't care about this functionality).

That way the functionality is there for folks that need to use it, but everyone else doesn't pay a feature bloat tax.

romtsn: yep, that's what I suggested to @Flash0ver internally as well. If y'all are fine with that, we could go that way 👍

Flash0ver:
Closing for now ... needs more discussion ... see getsentry/sentry-dotnet#5243.

src/Sentry/Platforms/Android/SentrySdk.cs
Flash0ver: drafted ...

```CSharp

namespace Sentry.InteropServices;

public static partial class SentryMarshal
{
    /// <summary>
    /// Retrieves breadcrumbs that were persisted to disk by the Android SDK's scope observer.
    /// These are breadcrumbs from the previous app session that were saved before the app terminated.
    /// </summary>
    /// <remarks>
    /// Read persisted breadcrumbs before the SDK starts.
    /// This avoids a race condition with breadcrumb-producing integrations that start writing
    /// to the QueueFile as soon as <see cref="O:SentryAndroid.Init"/> completes.
    /// </remarks>
    /// <returns>A list of persisted breadcrumbs, or an empty list if unavailable.</returns>
    [SupportedOSPlatform("android")]
    public static IReadOnlyList<Breadcrumb> GetPersistedBreadcrumbs(SentryAndroidOptions options)
    {
#if ANDROID
        return ReadPersistedBreadcrumbs(options) ?? [];
#else
        throw new PlatformNotSupportedException();
#endif
    }

    private static List<Breadcrumb>? ReadPersistedBreadcrumbs(JavaSdk.SentryOptions nativeOptions)
    {
        if (nativeOptions.CacheDirPath is null)
        {
            return null;
        }

        // Create a temporary observer to read breadcrumbs from the previous session's QueueFile.
        // This must be called before SentryAndroid.Init() completes, to avoid a race condition
        // with breadcrumb-producing integrations that write to the same QueueFile concurrently.
        using var observer = new JavaSdk.Cache.PersistingScopeObserver(nativeOptions);
        var result = observer.Read(nativeOptions, JavaSdk.Cache.PersistingScopeObserver.BreadcrumbsFilename,
            Java.Lang.Class.FromType(typeof(Java.Util.IList)));

        if (result is not Java.Util.IList javaList)
        {
            return null;
        }

        var breadcrumbs = new List<Breadcrumb>();
        for (var i = 0; i < javaList.Size(); i++)
        {
            if (javaList.Get(i)?.JavaCast<JavaSdk.Breadcrumb>() is { } javaBreadcrumb)
            {
                breadcrumbs.Add(javaBreadcrumb.ToBreadcrumb());
            }
        }
        return breadcrumbs;
    }
}
```

...
with two problems:
- bindings are `internal`, including
  - `Sentry.JavaSdk.Android.Core.SentryAndroidOptions`
  - `Sentry.JavaSdk.SentryOptions`
- the static method cannot be called before Init, as the Options are not yet available
  - the options are not created/owned by user code, but only modified by a user-provided callback

PR #4983, opened 2026-03-04
chore: Leverage new Lock types

### LINKED ISSUE #3689: Leverage new Lock types

jamescrosswell:
Resolves #3689
- #3689

#skip-changelog

### DIFF
--- a/GlobalUsings.cs
+++ b/GlobalUsings.cs
@@ -1,7 +1,8 @@
 // These are used in all projects except samples.
 
 // Use the new System.Threading.Lock type on .NET 9+, which enables compiler optimisations.
-// On earlier targets, alias it to object so the same lock pattern works everywhere.
+// On earlier targets, alias it to object so the same lock pattern works everywhere,
+// but don't use the variant of "Polyfill", as it may cause issues on Runtimes that support "Thread.Abort" (e.g. .NET Framework)
 #if !NET9_0_OR_GREATER
 global using Lock = object;
 #endif


src/Sentry.Log4Net/SentryAppender.cs
```diff
@@ -123,7 +123,7 @@ protected override void Append(LoggingEvent loggingEvent)
 
         evt.SetExtras(GetLoggingEventProperties(loggingEvent));
 
-        if (SendIdentity && !string.IsNullOrEmpty(loggingEvent.Identity))
+        if (SendIdentity && !string.IsNullOrEmpty(loggingEvent.Idntity))
```
jamescrosswell:
```suggestion
        if (SendIdentity && !string.IsNullOrEmpty(loggingEvent.Identity))
```

GlobalUsings.cs
Flash0ver, line 1: question: Also use for non-dedicated lock instances?

I wonder if it's "worth" using a separate `Lock` object for these four occurrences we have,
where we don't lock on a dedicated Lock-object yet,
but on the reference-type-instance in-use.

https://github.com/getsentry/sentry-dotnet/blob/adb6cfb5c84771128995e640ee4401d52067cd6a/src/Sentry/Internal/ConcurrentQueueLite.cs#L11

https://github.com/getsentry/sentry-dotnet/blob/217c11cea611a5f26e4a1abfbb7df37f34695722/src/Sentry/Platforms/Native/SentrySdk.cs#L9

https://github.com/getsentry/sentry-dotnet/blob/34b55980cd9f84d3922f80ff70c87b76d38ae01d/src/Sentry.Profiling/SamplingTransactionProfiler.cs#L15

https://github.com/getsentry/sentry-dotnet/blob/bc5c060f9ab0c371dc12ce596a9f5e0a023dcf2e/test/Sentry.Testing/RecordingTransport.cs#L7

Hmm ... if even ... maybe the `ConcurrentQueueLite` may have a noticeable performance benefit on `net9.0+`, which is worth the trade off of allocating an additional object.

But I believe - as ever so often - I'm actually now talking myself out of it.

GlobalUsings.cs
```diff
@@ -1,5 +1,11 @@
 // These are used in all projects except samples.
 
+// Use the new System.Threading.Lock type on .NET 9+, which enables compiler optimisations.
+// On earlier targets, alias it to object so the same lock pattern works everywhere.
+#if !NET9_0_OR_GREATER
```
Flash0ver: suggestion: Explain why we don't use `Polyfill`'s variant

```suggestion
// On earlier targets, alias it to object so the same lock pattern works everywhere,
// but don't use the variant of "Polyfill", as it may cause issues on Runtimes that support "Thread.Abort" (e.g. .NET Framework)
#if !NET9_0_OR_GREATER
```

src/Sentry.Maui/Internal/MauiDeviceData.cs
```diff
@@ -6,10 +6,8 @@ namespace Sentry.Maui.Internal;
 
 internal static class MauiDeviceData
 {
-#if NET9_0_OR_GREATER && ANDROID
+#if ANDROID
```
Flash0ver, line 9: praise: Nice cleanup

GlobalUsings.cs
jamescrosswell: Yeah I saw those... but I think we're better to use the existing reference object when we can (it's a less exclusive lock). The only alternative would be to create some Lock type with the same lifetime/scope as the reference types being locked... which seems a bit pointless.

PR #4988, opened 2026-03-05
chore: Set up dotagents with agent skills

Flash0ver:
closes #4970
- #4970

Add dotagents configuration (agents.toml, agents.lock) and install agent skills for Claude, Cursor, and VS Code. Includes skills for committing, branching, code review, PR writing, and security review from getsentry/skills.

Also update .gitignore to exclude the generated .agents/.gitignore file.

PR #4974, opened 2026-03-03
feat: Add AGENTS.md

jamescrosswell:
Closes #4971

## Summary

Adds `AGENTS.md` at the repo root (with a `CLAUDE.md` symlink) per the [repository-docs standard](https://develop.sentry.dev/sdk/getting-started/standards/repository-docs/#agentsmd) and the [agents-md skill](https://github.com/getsentry/skills/tree/main/plugins/sentry-skills/skills/agents-md). Also added a symlink for copilot as that's what I'm using mostly in Rider.

# Testing

Tested by asking (in copilot in my case, but using the Claude Sonnet 4.6 model): `Implement https://github.com/getsentry/sentry-dotnet/issues/4105 and create a pull request for this.` Claude implemented this in:
- #4975  

#skip-changelog

Issue #4970, opened 2026-03-02
Adopt `.agents`


Flash0ver:
### Description
See https://github.com/getsentry/dotagents

PR #4993, opened 2026-03-09
chore: Set up dotagents with agent skills for GitHub Copilot

Flash0ver:
Follow-up to
- #4970
- #4988

---

This enables the skills for GitHub Copilot,
which are managed via https://github.com/getsentry/dotagents locally for this repository.

Similar to the GitHub Copilot symlink for the `AGENTS.md`:
- https://github.com/getsentry/sentry-dotnet/pull/4974/changes#diff-227c2c26cb2ee0ce0f46a320fc48fbcbdf21801a57f59161b1d0861e8aad55f5

Related GitHub Copilot documentation
- https://docs.github.com/copilot/concepts/agents/about-agent-skills

### DIFF
--- a/.github/skills
+++ b/.github/skills
@@ -1,0 +1,1 @@
+.agents/skills


.github/skills
```diff
@@ -0,0 +1 @@
+.agents/skills
```
Flash0ver, line 1: todo: test

Does this actually work for GitHub Copilot (https://docs.github.com/copilot/concepts/agents/about-agent-skills#about-agent-skills).

Or is it even required in the first place and it works already even without this **symlink**?

jamescrosswell: Reading through the docs, at least, it looks like copilot will also pick up any symlinks configured for Claude... but no harm in adding these as well I guess.

Flash0ver:
closing
as GitHub Copilot does indeed pick up the `.claude/skills` symlink

.github/skills
jamescrosswell: After a bit of digging and testing, it does't work in Rider. It seems it depends on the extension being used to access copilot and the Rider extension doesn't look for skills directories yet. There might be a way to configure an MCP server that leverages the skills in a particular directory, but that kind of defeats the purpose of skills.

It may work in VS Code - that seems to look for the claude skills out of the box. Will have to see how good the R# extension for VS Code is. Otherwise maybe use one IDE for working with the LLM and another for crafting code by hand.

PR #5000, opened 2026-03-11
chore: Added cron job to check for upstream changes to vendored in code

### LINKED ISSUE #4995: Create watch on upstream repositories that we've vendored in
### Description

See #4994. Since these aren't proper submodules, we don't currently receive any notification when code we've vendored in changes. We need to create some kind of poll/monitor that will notify us of new pull requests affecting any code we've vendored in.

The dotnet/android repository is a particular thorn in our side, since it's still very much under active development... so we could start with a workflow that notifies us or relevant changes to this repository and, if that works OK, maybe extend the job so that we could give it a configurable list of paths within repositories that we what to monitor.

We don't want to be notified of all/any change in the [dotnet/android repository](https://github.com/dotnet/android)... just changes to any of the files in the `/tools/assembly-store-reader-mk2` folder and it's subfolders (since that's the code we vendored in).

Possibly/probably there are already github workflows out there that do this already so we should check for existing solutions to this problem before reinventing the wheel.

jamescrosswell:
Resolves #4995
- #4995

# Notes for testing

All the github workflow does is run the script on a cron schedule, passing it some parameters.  The script can be tested locally by running:

```
export GH_TOKEN="$(gh auth token)"
export GH_REPO="getsentry/sentry-dotnet"
./scripts/watch-upstream.sh dotnet/android tools/assembly-store-reader-mk2 src/Sentry.Android.AssemblyReader/
```

... that assumes you're logged into the github cli locally.

> [!IMPORTANT]
> The script will create one (and only one) issue for every upstream change it detects affecting the configured source tree. For testing, it created this issue:
> - https://github.com/getsentry/sentry-dotnet/issues/5001
> ... so it's not going to do anything again until we see a new commit affecting that source tree. 

#skip-changelog

### DIFF
--- a/scripts/watch-upstream.sh
+++ b/scripts/watch-upstream.sh
@@ -30,7 +30,7 @@ echo "Checking upstream: ${UPSTREAM_REPO}/${UPSTREAM_PATH}"
 # Resolve the upstream repo's default branch so URLs are correct even for repos
 # that use a branch name other than 'main'.
 DEFAULT_BRANCH=$(gh api "repos/${UPSTREAM_REPO}" --jq '.default_branch')
-UPSTREAM_URL="https://github.com/${UPSTREAM_REPO}/tree/${DEFAULT_BRANCH}/${UPSTREAM_PATH}"
+UPSTREAM_URL="${GITHUB_SERVER_URL}/${UPSTREAM_REPO}/tree/${DEFAULT_BRANCH}/${UPSTREAM_PATH}"
 
 # Get the latest commit SHA affecting the tracked path.
 LATEST_SHA=$(gh api "repos/${UPSTREAM_REPO}/commits?path=${UPSTREAM_PATH}&per_page=1" \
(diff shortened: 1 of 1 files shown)


scripts/watch-upstream.sh
```diff
@@ -0,0 +1,100 @@
...
+
+# Avoid creating duplicate issues: skip if any issue (open or closed) already
+# tracks this exact upstream commit SHA. The SHA in the title makes it unique.
+ISSUE_LABEL="upstream-watch"
+EXISTING_ISSUE=$(gh issue list \
+  --label "$ISSUE_LABEL" \
+  --state all \
+  --search "\"${UPSTREAM_REPO} ${UPSTREAM_PATH} @ ${LATEST_SHORT}\"" \
+  --json number,title \
+  --jq '.[0].number // empty')
```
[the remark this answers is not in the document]
jamescrosswell: This logic seems to work fine when I test from the command line. Let's address this if/when it ever becomes an issue.

scripts/watch-upstream.sh
```diff
@@ -0,0 +1,100 @@
...
+# Ensure the label exists (idempotent).
+gh label create "$ISSUE_LABEL" \
+  --description "Upstream vendored code has changed — review required" \
+  --color "E4E669" 2>/dev/null || true
+
+COMMIT_URL="https://github.com/${UPSTREAM_REPO}/commit/${LATEST_SHA}"
+HISTORY_URL="https://github.com/${UPSTREAM_REPO}/commits/${DEFAULT_BRANCH}/${UPSTREAM_PATH}"
+
+if [ -n "${GITHUB_RUN_ID:-}" ] && [ -n "${GITHUB_REPOSITORY:-}" ]; then
+  FOOTER="> _Automatically opened by the [Watch Upstream Changes](${GITHUB_SERVER_URL}/${GITHUB_REPOSITORY}/actions/runs/${GITHUB_RUN_ID}) workflow._"
```
Flash0ver: question: usage of `GITHUB_SERVER_URL`

This is the only occurrence of `GITHUB_SERVER_URL`.
Should this be used instead of `https://github.com` within this script?
I believe there are **3** occurrences of `https://github.com` that could be replaced.

scripts/watch-upstream.sh
```diff
@@ -0,0 +1,100 @@
...
+
+# Ensure the label exists (idempotent).
+gh label create "$ISSUE_LABEL" \
+  --description "Upstream vendored code has changed — review required" \
+  --color "E4E669" 2>/dev/null || true
+
+COMMIT_URL="https://github.com/${UPSTREAM_REPO}/commit/${LATEST_SHA}"
+HISTORY_URL="https://github.com/${UPSTREAM_REPO}/commits/${DEFAULT_BRANCH}/${UPSTREAM_PATH}"
+
+if [ -n "${GITHUB_RUN_ID:-}" ] && [ -n "${GITHUB_REPOSITORY:-}" ]; then
```
Flash0ver: question: `GH_REPO` vs `GITHUB_REPOSITORY`

I'm afraid I'm a bit confused about `GITHUB_REPOSITORY`, where I expected `GH_REPO`.
Are they synonymous ... or what is the difference?

scripts/watch-upstream.sh
jamescrosswell: Good call - I'll replace the hard coded instances for consistency.

scripts/watch-upstream.sh
jamescrosswell: Replaced `GITHUB_REPOSITORY` with `GH_REPO`

PR #5003, opened 2026-03-11
chore: Add error monitoring solution question to bug report template

markushi:
## Description

Adds a new dropdown and optional input field to the .NET bug report issue template asking users if they are using any other error monitoring solution alongside Sentry, and if so, which one.

## Motivation and Context

Other error monitoring SDKs can interfere with Sentry's error reporting in .NET applications. Knowing upfront whether a user has another solution installed helps triage bugs faster.

## How did you test it?

Reviewed the YAML template manually.

#skip-changelog

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -93,6 +93,25 @@ body:
     validations:
       required: true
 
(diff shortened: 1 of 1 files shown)


Flash0ver:
`Bug Report` issue template after merge:

image

PR #5014, opened 2026-03-11
chore: Apply recommended fixes for updated dotagents

jamescrosswell:
When installing dotagents locally in a fresh clone of sentry-dotnet:
```zsh
% npx @sentry/dotagents install 
Warning: agents.lock should be in .gitignore. Run 'dotagents doctor --fix' to fix.
Installed 13 skill(s): dotagents, agents-md, claude-settings-audit, code-review, code-simplifier, commit, create-branch, create-pr, find-bugs, gha-security-review, iterate-pr, pr-writer, security-review
% npx @sentry/dotagents doctor --fix
  ✓ agents.toml contains 'gitignore' which is no longer used in v1. Gitignore is always managed. Remove it.
  ✓ Missing from .gitignore: agents.lock. These files should not be committed.
  ✓ Generated files checked into git: agents.lock. Remove them with 'git rm --cached'.
  ✓ .agents/.gitignore exists.
  ✓ Skills directory exists.
  ✓ All 13 declared skill(s) installed.
  ✓ All symlinks intact.

Fixed 3 issue(s).
```

#skip-changelog

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -225,4 +225,4 @@ And yes, Sentry has a [Skill](https://github.com/getsentry/skills) to maintain t
 ### .agents
 
 We use [dotagents](https://github.com/getsentry/dotagents) as a package manager for agent skills and more.
-See [agents.toml](./agents.toml) and [agents.lock](./agents.lock) for our current configuration.
+See [agents.toml](./agents.toml) for our current configuration.


.gitignore
```diff
@@ -37,3 +37,5 @@ modules/sentry-cocoa
 
 # .agents/.gitignore is generated by dotagents — don't commit it.
 .agents/.gitignore
+# Auto-generated by dotagents — do not commit these files.
+agents.lock
```
jamescrosswell, line 40: @Flash0ver I thought the whole purpose of the lock file was so that we'd keep our versions in sync... maybe the guidance/warning I'm getting from dotagents when installing these is erroneous? If so, it'd be good if we could suppress it somehow.

Flash0ver: The warning and this changeset is as expected.

Our guidance on how repositories should adopt `dotagents` has changed:
Pinning is no longer supported, see getsentry/dotagents#56.

The lock-file no longer contains the version, only the source repository for each Agent Skill.

Issue #4529, opened 2025-09-16
Should CrashedLastRun, `EnableScopeSync` and `ScopeObserver` be internal?


jamescrosswell:
We're overwriting `CrashedLastRun`, `EnableScopeSync` and `ScopeObserver` here, so any user configured values for these will get overwritten. If that's genuinely what we want, those options should probably be internal (not public). I'll create a separate issue for that.

_Originally posted by @jamescrosswell in https://github.com/getsentry/sentry-dotnet/pull/4521#discussion_r2350765015_

Flash0ver:
We are thinking about removing `SentryOptions.EnableScopeSync` from the public API surface area,
and instead have it implicitly enabled by user code setting a `SentryOptions.ScopeObserver`,
or our Native/Android/iOS implementations setting a respective platform `IScopeObserver`,
or an aggregate of both a User and Sentry `IScopeObserver`.

Flash0ver:
Discussed again: We keep all options, but will add wrappers/combiners/aggregators for both `CrashedLastRun` and `ScopeObserver`, so that user-provided instances/delegates will not overwrite the ones that we set.

jamescrosswell:
@Flash0ver  after looking at this again, I think we misunderstood things the first time around. 

## Scope observers

It's not obvious, but the existing `CocoaScopeObserver` and `AndroidScopeObserver` implementations do preserve/chain any user defined scope observers here:
https://github.com/getsentry/sentry-dotnet/blob/438ae83fd12beee596d64be3b5876eabb81c94e0/src/Sentry/Platforms/Cocoa/CocoaScopeObserver.cs#L16-L18
https://github.com/getsentry/sentry-dotnet/blob/438ae83fd12beee596d64be3b5876eabb81c94e0/src/Sentry/Platforms/Android/AndroidScopeObserver.cs#L16-L18

[...]

PR #5038, opened 2026-03-18
feat: Scope sync attachments

bitsandfoxes:
Relates to https://github.com/getsentry/sentry-native/pull/1584 and https://github.com/getsentry/sentry-java/pull/5211

Allows syncing attachments to native layers. This will require a followup to implement the actual Android/iOS scope sync but unblocks Unity SDK.

#skip-changelog

### DIFF
--- a/src/Sentry/Platforms/Android/AndroidScopeObserver.cs
+++ b/src/Sentry/Platforms/Android/AndroidScopeObserver.cs
@@ -107,13 +107,25 @@ internal sealed class AndroidScopeObserver : IScopeObserver
 
     public void AddAttachment(SentryAttachment attachment)
     {
-        // TODO: Missing corresponding functionality on the Android SDK
-        _innerObserver?.AddAttachment(attachment);
+        try
+        {
+            // TODO: Missing corresponding functionality on the Android SDK
+        }
+        finally
+        {
+            _innerObserver?.AddAttachment(attachment);
+        }
     }
 
     public void ClearAttachments()
     {
-        // TODO: Missing corresponding functionality on the Android SDK
-        _innerObserver?.ClearAttachments();
+        try
+        {
+            // TODO: Missing corresponding functionality on the Android SDK
+        }
+        finally
+        {
+            _innerObserver?.ClearAttachments();
+        }
     }
 }
--- a/src/Sentry/Platforms/Cocoa/CocoaScopeObserver.cs
+++ b/src/Sentry/Platforms/Cocoa/CocoaScopeObserver.cs
@@ -115,13 +115,25 @@ internal sealed class CocoaScopeObserver : IScopeObserver
 
     public void AddAttachment(SentryAttachment attachment)
     {
-        // TODO: Missing corresponding functionality on the Cocoa SDK
-        _innerObserver?.AddAttachment(attachment);
+        try
+        {
+            // TODO: Missing corresponding functionality on the Cocoa SDK
+        }
+        finally
+        {
+            _innerObserver?.AddAttachment(attachment);
+        }
     }
 
     public void ClearAttachments()
     {
-        // TODO: Missing corresponding functionality on the Cocoa SDK
-        _innerObserver?.ClearAttachments();
+        try
+        {
+            // TODO: Missing corresponding functionality on the Cocoa SDK
+        }
+        finally
+        {
+            _innerObserver?.ClearAttachments();
+        }
     }
 }


src/Sentry/Scope.cs
```diff
@@ -535,7 +546,8 @@ public void Apply(Scope other)
 
         foreach (var attachment in Attachments)
         {
-            other.AddAttachment(attachment);
+            // Set the attachment directly to avoid triggering a scope sync
+            other._attachments.Add(attachment);
```
bitsandfoxes, line 550: The pushing and popping the scope does not get propagated to the native layer. So syncing the same attachments again doesn't make sense.

src/Sentry/Platforms/Native/NativeScopeObserver.cs
jamescrosswell, line 1: This was actually just added in 0.13.3 of the Native SDK:
- https://github.com/getsentry/sentry-dotnet/pull/5045#issue-4105528155

Added #5052 as a follow up:
- #5052

src/Sentry/Scope.cs
jamescrosswell: Android and iOS would both be using Global mode anyway, so no pushing/popping of scopes.

I am wondering what this means for native AOT applications that are using the Native SDK though. For ASP.NET Core apps, for example, we wouldn't want attachments from one async local scope context to end up being synced to some 'global' scope in the native SDK (and an attachment from one request ending up being captured in a native event for a completely unrelated request). 

To be clear, I'm not sure if this is possible - just a concern that popped up when you made this remark.

I've added a note to #5052 to investigate this before implementing for Native.

jamescrosswell:
@bitsandfoxes this looks fine. The only thing I'm wondering is whether we're likely to implement this for Android, iOS and Native... for native, I can see some potential issues. If we don't end up implementing for one/more of them, I guess we can just add some comments to the relevant `Impl` methods explaining why. It's a bit weird having abstract members that aren't implemented in some descendants  (potentially a base implementation that is a no-op is preferable)... but they're all internal anyway, so we can always refactor once we've investigated the other platforms.

@Flash0ver does that sound good to you?

src/Sentry/Scope.cs
bitsandfoxes: > an attachment from one request ending up being captured in a native event for a completely unrelated request)

A possible solution to this is to limit the attachment sync to global scope mode. That would also be a simple mental model: "global scope persists across all layers".

jamescrosswell: > A possible solution to this is to limit the attachment sync to global scope mode.

That's a good idea... I think probably the only safe way to do it 👍🏻

test/Sentry.Tests/AttachmentTests.cs
```diff
@@ -1,7 +1,38 @@
 namespace Sentry.Tests;
 
+public class ByteAttachmentContentTests
+{
+    [Fact]
+    public void Bytes_ReturnsConstructorValue()
+    {
+        var data = new byte[] { 1, 2, 3 };
+        var content = new ByteAttachmentContent(data);
+        Assert.Same(data, content.Bytes);
```
Flash0ver, line 10: question: Does Unity need to access the `Bytes`?

I'm wondering if this test is worth it,
and if it is consequently worth it to make the `Bytes` internal.

If we actually only access it via `GetStream` - as the test below,
I'd believe I prefer to also only test that surface area,
as the test below does confirm the expected behavior,
and this test is "only" testing implementation detail whether the backing field of a property is set correctly,
which I think is a bit of noise.

But - if Unity does need to access the `Bytes` property,
than I actually do like this test,
as Unity is consuming this property directly,
and this test is confirming that path.

If not, though, I prefer to make the field private again and not have it accessible via a property.

test/Sentry.Tests/AttachmentTests.cs
```diff
@@ -1,7 +1,38 @@
...
+    }
+}
+
 public class FileAttachmentContentTests
 {
+    [Fact]
+    public void FilePath_ReturnsConstructorValue()
+    {
+        var attachment = new FileAttachmentContent("/some/path/file.txt");
+        Assert.Equal("/some/path/file.txt", attachment.FilePath);
```
Flash0ver, line 33: question: does `FilePath` need to be accessible for Unity?

I'm wondering if we made `FilePath` only internal for this test.
This test is "only" testing whether the constructor sets a backing field correctly.

If Unity does need access to it, then I do like the test as we have a consumer directly accessing it.

If Unity does not need to access it,
then I'd prefer to make the field private again, and remove the accessor,
and instead rewrite this test to actually check whether the created file - similar to the test below - does return the expected content via the actual `GetStream` interface.

So that our tests don't depend on too much implementation details,
but actually on the expected code paths ... if that makes sense.

Flash0ver:
> @bitsandfoxes this looks fine. The only thing I'm wondering is whether we're likely to implement this for Android, iOS and Native... for native, I can see some potential issues. If we don't end up implementing for one/more of them, I guess we can just add some comments to the relevant `Impl` methods explaining why. It's a bit weird having abstract members that aren't implemented in some descendants (potentially a base implementation that is a no-op is preferable)... but they're all internal anyway, so we can always refactor once we've investigated the other platforms.
> 
> @Flash0ver does that sound good to you?

Yes.
I guess we could instead make the "no-op" implementations `virtual` rather than `abstract`.
But currently we are locally consistent which also has some value.

If we later don't actually have an implementation for _native_,
I believe we still could make the `...Impl` methods `virtual` instead of `abstract` (to not have an `override` implementation that's just _no-op_) without a **breaking change** ... I think ... the other way around (from `virtual` to `abstract`) would be a **breaking change**, though.

bitsandfoxes:
> question: does FilePath need to be accessible for Unity?

Yes, both `Byte` and `FilePath` were made internal to be accessible from inside the Unity SDK, hence the tests.

src/Sentry/Platforms/Android/AndroidScopeObserver.cs
```diff
@@ -104,4 +104,28 @@ public void SetTrace(SentryId traceId, SpanId parentSpanId)
     {
         // TODO: This requires sentry-java 8.4.0
```
Flash0ver, line 105: question: Bug?

Is it a bug that we don't forward to the Inner-Observer's `SetTrace`?

src/Sentry/Platforms/Android/AndroidScopeObserver.cs
```diff
@@ -104,4 +104,28 @@ public void SetTrace(SentryId traceId, SpanId parentSpanId)
...
     }
+
+    public void AddAttachment(SentryAttachment attachment)
+    {
+        try
```
Flash0ver, line 110: question: try-finally

Do we need the try-finally-blocks?
Or could we just forward to the Inner-Observer's method?

Or did we add it to be sure we don't forget later?
I'm a bit uncertain about this now 🤔.

test/Sentry.Tests/ApiApprovalTests.Run.DotNet10_0.verified.txt
```diff
@@ -217,7 +217,9 @@ namespace Sentry
     }
     public interface IScopeObserver
     {
+        void AddAttachment(Sentry.SentryAttachment attachment);
         void AddBreadcrumb(Sentry.Breadcrumb breadcrumb);
+        void ClearAttachments();
         void SetExtra(string key, object? value);
```
Flash0ver, line 223: question: Shall we document the breaking changes?

test/Sentry.Tests/AttachmentTests.cs
Flash0ver: Needs to be accessible for Unity-SDK, which has access to this internals.
https://github.com/getsentry/sentry-dotnet/pull/5038#issuecomment-4117738870

test/Sentry.Tests/AttachmentTests.cs
Flash0ver: Yes ... the Unity-SDK needs access to this ... to which the internals are made accessible to ... https://github.com/getsentry/sentry-dotnet/pull/5038#issuecomment-4117738870

test/Sentry.Tests/ApiApprovalTests.Run.DotNet10_0.verified.txt
bitsandfoxes: Based on https://github.com/getsentry/sentry-dotnet/issues/4529 the `ScopeObserver` should be `internal` in the first place. So I don't think I'd advocate for new API that we know is going to get removed.

src/Sentry/Platforms/Android/AndroidScopeObserver.cs
bitsandfoxes: It should get forwarded. I'm not sure there is an inner observer to forward this in the first place. Also tracked by https://github.com/getsentry/sentry-dotnet/issues/4074

src/Sentry/Platforms/Android/AndroidScopeObserver.cs
bitsandfoxes: This was added to not forget doing it later and follows the pattern established in the methods above.

PR #5039, opened 2026-03-18
feat: User.Id can now be overriden (set to null) in Global mode

### LINKED ISSUE #4172: (MAUI) user.id is always set and can't remove it with SendDefaultPii or null
Linear:
[NET-492](https://linear.app/getsentry/issue/NET-492/maui-userid-is-always-set-and-cant-remove-it-with-senddefaultpii-or)

### Package

Sentry

### .NET Flavor

.NET

### .NET Version

9.0.4

### OS

Android

### OS Version

_No response_

### Development Environment

Visual Studio v17.x

### SDK Version

5.6.0

24.8.0

### Workload Versions

A break change that I noticed from the Xamarin SDK to MAUI is that User.id is always set.

`SendDefaultPii` does nothing when set to false, it keeps sending this identifier, and there doesn't seems to be an easy way remove.

Even with 
```C#
            SentrySdk.CaptureMessage("teste user id", scope =>
            {
                scope.User.Id = null;
            });
```
it still sends an event with an user.id

Image

Image

```
        builder.UseSentry((option) =>
        {
            option.ReportAssembliesMode = ReportAssembliesMode.None;
            option.Dsn = EnvironmentConfig.SentryDsn;
            option.AutoSessionTracking = false;
            option.TracesSampleRate = 1.0;
            option.AddEventProcessor(new SentryEventProcessor());
            option.AddExceptionProcessor(new SentryExceptionProcessor());
            option.AddExceptionFilterForType<InvalidSessionException>();
            option.Environment = EnvironmentConfig.AppEnvironment.ToString();
            option.AttachScreenshot = true;
            option.SendDefaultPii = false;
            option.SetBeforeScreenshotCapture(BeforeScreenshotCapture);
            if (EnvironmentConfig.AppEnvironment == EAppEnvironment.Prod)
            {
                option.MaxBreadcrumbs = 50;
            }
#if DEBUG
            option.Debug = true;
#endif
        });
```

### Steps to Reproduce

- Don't set user.id
And one of the following options:
- 1) Set user.id as null on configure scope or inside the capture message scope.
- 2) Have SendDefaultPii as false.
- 3) Using 
 ```C#
{
    scope.User = new SentryUser();
});
```
One of the above options should unset the user.id from any event.

### Expected Result

user.id not set with installationID

### Actual Result

events with user.id not set by the developer

### PR DESCRIPTION
**Pull request by jamescrosswell**
Resolves #4172
- #4172

### CODE REVIEW

File: src/Sentry/Internal/Enricher.cs [hunk 1]
```diff
@@ -92,7 +92,12 @@ public void Apply(IEventLike eventLike)
 
             eventLike.User.IpAddress ??= DefaultIpAddress;
         }
-        eventLike.User.Id ??= _options.InstallationId;
+        // Set by the GlobalRootScopeIntegration In global mode so that it can be overridden by the user.
+        // In non-global mode (e.g. ASP.NET Core) the enricher sets it here as a fallback.
```
jamescrosswell, line 95: @Flash0ver arguably we could remove this code entirely... not sure if we actually want to set the same user id for all ASP.NET requests (where no user was logged in). 

@bruno-garcia any thoughts from your side?

test/Sentry.Tests/SentryOptionsTests.Integrations_default_ones_are_properly_registered.DotNet10_0.DotNet.verified.txt
```diff
@@ -1,13 +1,4 @@
 ﻿[
-  {
-    Message: Initializing Hub for Dsn: '{0}'.,
-    Args: [
-      https://d4d82fc1c2c4032a83f3a29aa3a3aff@fake-sentry.io:65535/2147483647
-    ]
-  },
-  {
-    Message: Starting BackpressureMonitor.
```
jamescrosswell, line 9: These were never really part of the test anyway (which is to ensure all the default integrations get registered)

src/Sentry/Integrations/GlobalRootScopeIntegration.cs
jamescrosswell, line 1: We could potentially set more stuff here. There are lots of scope properties that never change and which we currently set on the scope every time we capture an event. These are all candidates, I think:
https://github.com/getsentry/sentry-dotnet/blob/f50b36095ca58d50837602fba76372311684a779/src/Sentry/Scope.cs#L498-L517


More of a performance improvement but might result in a subtle behavioural change for some users (perhaps code that checks/expects things to not be set in certain circumstances and sets them conditionally on this basis ). To be safe, perhaps we delay a change like that until the next major release.

test/Sentry.Tests/SentryClientTests.cs
```diff
@@ -302,10 +301,17 @@ public void CaptureEvent_EventAndScope_CopyScopeIntoEvent()
...
-    [Fact]
+    [SkippableFact]
     public void CaptureEvent_UserIsNull_SetsFallbackUserId()
     {
+#if NET5_0_OR_GREATER
+        // In global mode the userid gets set at app startup via the GlobalRootScopeIntegration, rather than by an
+        // enricher during capture... so this functionality in SentryClient only works when IsGlobalModeEnabled is false
+        Skip.If(System.OperatingSystem.IsAndroid() || System.OperatingSystem.IsIOS(),
+            "On mobile, User.Id is set by GlobalRootScopeIntegration at startup, not the enricher.");
+#endif
```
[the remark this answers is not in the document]
jamescrosswell: Explained in the comment... the tests assert a behaviour we don't expect from the enricher on Mobile.

src/Sentry/Integrations/GlobalRootScopeIntegration.cs
```diff
@@ -0,0 +1,14 @@
...
+{
+    public void Register(IHub hub, SentryOptions options)
+    {
+        if (!options.IsGlobalModeEnabled)
+        {
+            return;
+        }
+
+        hub.ConfigureScope(scope => scope.User.Id ??= options.InstallationId);
+    }
```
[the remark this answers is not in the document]
jamescrosswell: It's not PII data and this is consistent with the  behaviour before this PR.

src/Sentry/Internal/Enricher.cs
```diff
@@ -92,7 +91,12 @@ public void Apply(IEventLike eventLike)
...
+        if (!_options.IsGlobalModeEnabled)
+        {
+            eventLike.User.Id ??= _options.InstallationId;
+        }
```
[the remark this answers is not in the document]
jamescrosswell: I don't think there's a scenario where someone uses a SentryClient independent of the Hub... so this isn't a real concern.

src/Sentry/SentryOptions.cs
```diff
@@ -1362,6 +1367,7 @@ public SentryOptions()
 #if NET8_0_OR_GREATER
                                | DefaultIntegrations.SystemDiagnosticsMetricsIntegration
 #endif
+                               | DefaultIntegrations.GlobalRootScopeIntegration
                                ;
```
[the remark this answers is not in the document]
jamescrosswell: Arguably it's not an 'integration'... it's just some behaviour we want on all global root scopes (there's no need for an opt-out). We could move it out of the integrations list and bake the logic into the `Hub` constructor instead.

Otherwise I think it's OK to simply omit the disable integration method for this one. @Flash0ver thoughts?

lucas-zimerman:
Thank you for the PR!

test/Sentry.Tests/Integrations/GlobalRootScopeIntegrationTests.cs
```diff
@@ -0,0 +1,145 @@
...
+        };
+
+        var hub = Substitute.For<IHub>();
+        var integration = new GlobalRootScopeIntegration();
+
+        // Act
+        integration.Register(hub, options);
+
+        // Assert
+        hub.Received(1).ConfigureScope(Arg.Any<Action<Scope>>());
```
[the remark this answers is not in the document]
jamescrosswell: See [219f31e](https://github.com/getsentry/sentry-dotnet/pull/5039/commits/219f31ee4754e3c8ea0dabcac01e069cfd97d58d)

test/Sentry.Tests/SentryClientTests.cs
```diff
@@ -1294,10 +1300,17 @@ public void CaptureTransaction_ScopeContainsAttachments_GetAppliedToHint()
         hint.Attachments.Should().Contain(attachments);
     }
 
-    [Fact]
+    [SkippableFact]
     public void CaptureTransaction_UserIsNull_SetsFallbackUserId()
     {
+#if NET5_0_OR_GREATER
+        Skip.If(System.OperatingSystem.IsAndroid() || System.OperatingSystem.IsIOS(),
+            "On mobile, User.Id is set by GlobalRootScopeIntegration at startup, not the enricher.");
```
Flash0ver: suggestion: use `nameof`

For "go to declaration"-ability.

```suggestion
            $"On mobile, User.Id is set by {nameof(GlobalRootScopeIntegration)} at startup, not the enricher.");
```

test/Sentry.Tests/SentryClientTests.cs
```diff
@@ -302,10 +301,17 @@ public void CaptureEvent_EventAndScope_CopyScopeIntoEvent()
...
 
-    [Fact]
+    [SkippableFact]
     public void CaptureEvent_UserIsNull_SetsFallbackUserId()
     {
+#if NET5_0_OR_GREATER
+        // In global mode the userid gets set at app startup via the GlobalRootScopeIntegration, rather than by an
+        // enricher during capture... so this functionality in SentryClient only works when IsGlobalModeEnabled is false
+        Skip.If(System.OperatingSystem.IsAndroid() || System.OperatingSystem.IsIOS(),
+            "On mobile, User.Id is set by GlobalRootScopeIntegration at startup, not the enricher.");
```
Flash0ver: suggestion: use `nameof`

For easier navigation in a code editor.

```suggestion
            $"On mobile, User.Id is set by {nameof(GlobalRootScopeIntegration)} at startup, not the enricher.");
```

test/Sentry.Tests/SentryClientTests.cs
```diff
@@ -302,10 +301,17 @@ public void CaptureEvent_EventAndScope_CopyScopeIntoEvent()
         Assert.Equal(scope.Breadcrumbs, @event.Breadcrumbs);
     }
 
-    [Fact]
+    [SkippableFact]
     public void CaptureEvent_UserIsNull_SetsFallbackUserId()
     {
+#if NET5_0_OR_GREATER
+        // In global mode the userid gets set at app startup via the GlobalRootScopeIntegration, rather than by an
+        // enricher during capture... so this functionality in SentryClient only works when IsGlobalModeEnabled is false
```
Flash0ver: suggestion (nit pick): move comment

Move comment between `// Arrange` and `_fixture.SentryOptions.IsGlobalModeEnabled = false;`,
so that this test is in sync with `CaptureTransaction_UserIsNull_SetsFallbackUserId`.

src/Sentry/Integrations/GlobalRootScopeIntegration.cs
Flash0ver: Agreed ... let's open an issue for the .NET 11 / `v7` project.

src/Sentry/Integrations/GlobalRootScopeIntegration.cs
```diff
@@ -0,0 +1,14 @@
...
+internal class GlobalRootScopeIntegration : ISdkIntegration
+{
+    public void Register(IHub hub, SentryOptions options)
+    {
+        if (!options.IsGlobalModeEnabled)
+        {
+            return;
+        }
+
+        hub.ConfigureScope(scope => scope.User.Id ??= options.InstallationId);
```
Flash0ver, line 12: question: compliant with new PII rules?

See https://develop.sentry.dev/sdk/foundations/client/data-collection/.
I haven't checked the spec yet ... currently in the draft phase ... need to go though it.

I just wanted to make sure we don't paint ourselves into a corner when - later this year - we implement the new PII rules via a new `dataCollection`-Options.

src/Sentry/Internal/Enricher.cs
Flash0ver: Could this be a behavior breaking change?
Would this be something for .NET 11 / `v7`?

src/Sentry/PlatformAbstractions/FrameworkInfo.cs
```diff
@@ -36,6 +36,7 @@ public static partial class FrameworkInfo
             {528372, "4.8"},
             {528449, "4.8"},
             {533320, "4.8.1"},
-            {533325, "4.8.1"}
+            {533325, "4.8.1"},
+            {533509, "4.8.1"}
```
Flash0ver, line 40: question: how did you notice this?

CI just started to fail when a new servicing update is applied to the runners?

src/Sentry/SentryOptions.cs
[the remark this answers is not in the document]
Flash0ver: We also don't offer a `DisableAutoSessionTrackingIntegration` opt-out for `DefaultIntegrations.AutoSessionTrackingIntegration`.
Not sure if by design or an oversight.

But for all other `DefaultIntegrations`, there is a `Disable*` opt-out.

Perhaps a question that helps to decide is:
- Should this emit a diagnostic log?
https://github.com/getsentry/sentry-dotnet/blob/f50b36095ca58d50837602fba76372311684a779/src/Sentry/Internal/Hub.cs#L100-L108

On the other hand, we don't really have any "logic" in `Hub.ctor` other than:
https://github.com/getsentry/sentry-dotnet/blob/f50b36095ca58d50837602fba76372311684a779/src/Sentry/Internal/Hub.cs#L77-L81

... I am not answering the question ...

If we are happy that this emits a Debug-Log,
and if the Debug-Log does not necessarily indicate to the user that this is behavior that can be opted-out of,
but instead just gives more insights into the system and debuggability that Global-Mode is enabled and the SDK applies some default behavior,
then I am slightly leaning towards the `ISdkIntegration`.

If otherwise this could cause more confusion to the user than informational benefit, I guess then the location to invoke the `ConfigureScope` is after:
https://github.com/getsentry/sentry-dotnet/blob/f50b36095ca58d50837602fba76372311684a779/src/Sentry/Internal/Hub.cs#L77-L81

Although I am currently ever so slightly preferring the `ISdkIntegration`-based approach, I am very overrulable here, considering my feedback/thoughts.

test/Sentry.Tests/Integrations/GlobalRootScopeIntegrationTests.cs
```diff
@@ -0,0 +1,115 @@
...
+        };
+
+        var hub = Substitute.For<IHub>();
+        var integration = new GlobalRootScopeIntegration();
+
+        // Act
+        integration.Register(hub, options);
+
+        // Assert
+        hub.DidNotReceive().ConfigureScope(Arg.Any<Action<Scope>>());
```
Flash0ver, line 27: suggestion: test consistency

Here, we check whether an invocation occurred or not.
But the following two tests check `scope.User.Id`.

Could we here do a similar setup with the Scope and assert that,
so that we are consistent and indeed test correlated behavior.

In other words ... could we
- also create a `Scope`
- call `hub.SubstituteConfigureScope(scope);`
- and then assert that the `scope.User.Id` is null

And also - to be consistent the other way around to - add to the other tests an assertion that the invocation did occur ... to have consistent assertions:
- `hub.Received(1).ConfigureScope(Arg.Any<Action<Scope>>());`

src/Sentry/PlatformAbstractions/FrameworkInfo.cs
jamescrosswell: I think I ran the unit tests on my Windows box which is running the latest version of Windows and service packs, which broke the tests. Presumably those are now hitting the CI runners.

src/Sentry/SentryOptions.cs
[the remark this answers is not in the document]
jamescrosswell: Let's just leave it as an integration then... I'm not seeing any compelling reason to change it. The debug logs may be helpful to know what's going on and don't affect production apps in any way.

src/Sentry/Integrations/GlobalRootScopeIntegration.cs
jamescrosswell: https://github.com/getsentry/sentry-dotnet/issues/5197

src/Sentry/Internal/Enricher.cs
jamescrosswell: Nobody is asking for the change, so let's just leave it.

src/Sentry/Integrations/GlobalRootScopeIntegration.cs
jamescrosswell: Without the installation id, some functionality like Session Health simply doesn't work. So I can't see how we can avoid setting this.

This PR also doesn't introduce the collection of any new information - it just shifts when we're collecting it. So if we no longer want to generate installation ids and send these with events etc. that's fine - but I don't think related to this PR (that would be a separate issue and separate PR in the backlog).

Arguably this PR improves the situation since previously users couldn't opt out of the installation id being set/collected. Now, since we set it early, they can override it by nulling it out if they want.

Flash0ver, line 12: We could use the non-Closure-allocating overload:

```diff
-hub.ConfigureScope(scope => scope.User.Id ??= options.InstallationId);
+hub.ConfigureScope(static (scope, options) => scope.User.Id ??= options.InstallationId, options);
```

But let's do this in a follow-up, as this would also need a change to the tests, and it's not really worth going through the entire cycle again.

Opened #5219.

Flash0ver (approved):

PR #5040, opened 2026-03-19
chore: Demonstrate screen capture in mobile samples

### LINKED ISSUE #4117: docs: Set AttachScreenshot=true in the install wizard(s)
We should have `attachScreenshot=true` in our install wizards, the docs and the in-product/new project wizard.

Probably a comment above it:
```csharp
// If your app doesn't have sensitive data, you can get screenshots on error events automatically
```

jamescrosswell:
Resolves #4117
- #4117

### DIFF
--- a/samples/Sentry.Samples.Android/MainActivity.cs
+++ b/samples/Sentry.Samples.Android/MainActivity.cs
@@ -31,6 +31,12 @@ public class MainActivity : Activity
             // Enable Native Android SDK ANR detection
             options.Native.AnrEnabled = true;
 
+            // If your app doesn't have sensitive data, you can attach screenshots automatically when a Java/native
+            // error is captured.
+            // Note: this only applies to Java-based errors; .NET managed exceptions are not currently supported.
+            // https://docs.sentry.io/platforms/android/enriching-events/screenshots/
+            options.Native.AttachScreenshot = true;
+
             // Currently experimental support is only available on Android
             options.Native.ExperimentalOptions.SessionReplay.OnErrorSampleRate = 1.0;
             options.Native.ExperimentalOptions.SessionReplay.SessionSampleRate = 1.0;
--- a/samples/Sentry.Samples.Android/MainActivity.cs
+++ b/samples/Sentry.Samples.Android/MainActivity.cs
@@ -33,8 +33,7 @@ public class MainActivity : Activity
 
             // If your app doesn't have sensitive data, you can attach screenshots automatically when a Java/native
             // error is captured.
-            // Note: this only applies to Java-based errors; .NET managed exceptions are not currently supported.
-            // https://docs.sentry.io/platforms/android/enriching-events/screenshots/
+            // https://docs.sentry.io/platforms/android/configuration/options/#attachScreenshot
             options.Native.AttachScreenshot = true;
 
             // Currently experimental support is only available on Android
--- a/samples/Sentry.Samples.Ios/AppDelegate.cs
+++ b/samples/Sentry.Samples.Ios/AppDelegate.cs
@@ -36,6 +36,10 @@ public class AppDelegate : UIApplicationDelegate
             // Enable Native iOS SDK App Hangs detection
             options.Native.EnableAppHangTracking = true;
 
+            // If your app doesn't have sensitive data, you can get screenshots on error events automatically
+            // https://docs.sentry.io/platforms/apple/guides/ios/configuration/options/#attach-screenshot
+            options.Native.AttachScreenshot = true;
+
             options.CacheDirectoryPath = Path.GetTempPath();
 
             options.SetBeforeSend(evt =>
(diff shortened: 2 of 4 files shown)


samples/Sentry.Samples.Ios/AppDelegate.cs
```diff
@@ -36,6 +36,10 @@ public override bool FinishedLaunching(UIApplication application, NSDictionary?
             // Enable Native iOS SDK App Hangs detection
             options.Native.EnableAppHangTracking = true;
 
+            // If your app doesn't have sensitive data, you can get screenshots on error events automatically
+            // https://docs.sentry.io/platforms/apple/guides/ios/configuration/options/#attach-screenshot
```
Flash0ver: issue: anchor not available

```suggestion
            // https://docs.sentry.io/platforms/apple/guides/ios/configuration/options/#attachScreenshot
```

samples/Sentry.Samples.MacCatalyst/AppDelegate.cs
```diff
@@ -23,6 +23,10 @@ public override bool FinishedLaunching(UIApplication application, NSDictionary?
             options.Dsn = EnvironmentVariables.Dsn;
 #endif
             options.Debug = true;
+
+            // If your app doesn't have sensitive data, you can get screenshots on error events automatically
+            // https://docs.sentry.io/platforms/apple/guides/mac-catalyst/configuration/options/#attach-screenshot
```
Flash0ver: issue: Page Not Found

samples/Sentry.Samples.Maui/MauiProgram.cs
```diff
@@ -30,7 +30,7 @@ public static MauiApp CreateMauiApp()
                 // If you want to see everything we can capture from MAUI, you may wish to use a larger value.
                 options.MaxBreadcrumbs = 1000;
 
-                // Be aware that screenshots may contain PII
+                // If your app doesn't have sensitive data, you can get screenshots on error events automatically
```
Flash0ver, line 33: suggestion: Add link for consistency with the other comments

```suggestion
                // If your app doesn't have sensitive data, you can get screenshots on error events automatically
                // https://docs.sentry.io/platforms/dotnet/guides/maui/configuration/options/#AttachScreenshot
```

samples/Sentry.Samples.Android/MainActivity.cs
```diff
@@ -31,6 +31,12 @@ protected override void OnCreate(Bundle? savedInstanceState)
             // Enable Native Android SDK ANR detection
             options.Native.AnrEnabled = true;
 
+            // If your app doesn't have sensitive data, you can attach screenshots automatically when a Java/native
+            // error is captured.
+            // Note: this only applies to Java-based errors; .NET managed exceptions are not currently supported.
+            // https://docs.sentry.io/platforms/android/enriching-events/screenshots/
```
Flash0ver: question: consistency of links

This comments links to feature documentation,
where the other links in the other comments link to the Configuration / Option.

Should we make all links consistently either point to the feature documentation of the respective platform, or the Configuration/Option of the respective platform?

samples/Sentry.Samples.Android/MainActivity.cs
```diff
@@ -31,6 +31,12 @@ protected override void OnCreate(Bundle? savedInstanceState)
             // Enable Native Android SDK ANR detection
             options.Native.AnrEnabled = true;
 
+            // If your app doesn't have sensitive data, you can attach screenshots automatically when a Java/native
+            // error is captured.
+            // Note: this only applies to Java-based errors; .NET managed exceptions are not currently supported.
```
Flash0ver: question: Note for iOS?

Does this Note also apply to both the iOS and Mac Catalyst samples?

Flash0ver (commented):
Got a few questions / notes about consistency and link correctness.

samples/Sentry.Samples.Android/MainActivity.cs
jamescrosswell: Tbh, I can't remember where I copied this from (presumably some documentation for the Android SDK but I can no longer find it). 

Let's just remove this line as it's potentially confusing. 

To answer your question though, I think yes this would apply to both Android and iOS... it'd to do with the order in which the signal handlers fire and .NET will try to insert it's own signal handler in the queue first so that if an exception occurs in what was, originally, managed code, it gets handled by the .NET exception handler (e.g. a `try..catch` block). The native SDK that is capturing screenshots (per this setting) will only get activated when capturing native exceptions (i.e. if .NET's signal handler didn't already fire).

@jpnurmi does it sound like I've got that right? You've been going deep down the rabbit hole on this recently...

samples/Sentry.Samples.Android/MainActivity.cs
jamescrosswell: Soulds good... those docs link to the feature docs for further information anyway.

samples/Sentry.Samples.Android/MainActivity.cs
jpnurmi: Yeah, with today's signal-handling order, both sentry-native and sentry-cocoa step on the .NET runtime's toes and accidentally handle certain .NET exceptions as fatal.

jpnurmi (approved):
LGTM 👍

PR #5053, opened 2026-03-23
chore: Link more root files in solution

Flash0ver:
### Summary

Add more links to files from the repository root to the Solution file(s).

### Remarks

Not sure if it's worth adding all of the suggested files.
As a minimum, I feel that the recently added "AI-files" are useful:
- `./AGENTS.md`
- `./agents.toml`
- `./warden.toml`

### DIFF
--- a/.generated.NoMobile.slnx
+++ b/.generated.NoMobile.slnx
@@ -57,13 +57,22 @@
   <Folder Name="/root/">
     <File Path=".codecov.yml" />
     <File Path=".gitignore" />
+    <File Path="AGENTS.md" />
+    <File Path="agents.toml" />
+    <File Path="before.Sentry.sln.targets" />
+    <File Path="build.cmd" />
+    <File Path="build.sh" />
     <File Path="CHANGELOG.md" />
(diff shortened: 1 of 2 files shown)


.generated.NoMobile.slnx
```diff
@@ -57,13 +57,22 @@
   <Folder Name="/root/">
     <File Path=".codecov.yml" />
     <File Path=".gitignore" />
+    <File Path="AGENTS.md" />
+    <File Path="agents.toml" />
+    <File Path="before.Sentry.sln.targets" />
```
Flash0ver, line 62: question: Worth adding all of these?

Not sure if we should indeed add all of these files ... or only the most prevalent files, that we are also likely to edit frequently.

Since I am using "Find in Files" quite a lot in Rider, which only includes Solution-included files, and not the entire repo's "File System", I thought about adding them all.

jamescrosswell: Worth adding them for convenience I think.

jamescrosswell (approved):

PR #5057, opened 2026-03-23
chore: Added code intelligence plugins

jamescrosswell:
See:
https://code.claude.com/docs/en/discover-plugins#code-intelligence

#skip-changelog

### DIFF
--- a/.claude/settings.json
+++ b/.claude/settings.json
@@ -33,7 +33,6 @@
   },
   "enabledPlugins": {
     "csharp-lsp@claude-plugins-official": true,
-    "typescript-lsp@claude-plugins-official": true,
     "clangd-lsp@claude-plugins-official": true,
     "jdtls-lsp@claude-plugins-official": true,
     "swift-lsp@claude-plugins-official": true


.claude/settings.json
```diff
@@ -30,5 +30,12 @@
       "Bash(mv:*)"
     ],
     "deny": []
+  },
+  "enabledPlugins": {
+    "csharp-lsp@claude-plugins-official": true,
+    "typescript-lsp@claude-plugins-official": true,
```
Flash0ver: question: Do we have _TypeScript_ in this repo? 🤔

I guess we got some _JavaScript_ in our Blazor samples.

Oh .. wait .. am I embarrassing myself because the TS LSP - of course - also supports JS? 😓

AGENTS.md
```diff
@@ -158,3 +158,16 @@ AI commits MUST include:
...
 ```
+
+### Code Intelligence
+
+Prefer LSP over Grep/Read for code navigation — it's faster, precise, and avoids reading entire files:
+
+- `workspaceSymbol` to find where something is defined
+- `findReferences` to see all usages across the codebase
+- `goToDefinition` / `goToImplementation` to jump to source
+- `hover` for type info without reading the file
```
Flash0ver: question: Member info

How does Member info work?
Or does this `hover` instruction already include member infos like available and invoked Methods, Properties and nested types?

I'm a bit uncertain as to how to "test" this.
Or do you have a scenario where you already see the impact?

.claude/settings.json
jamescrosswell: > question: Do we have _TypeScript_ in this repo? 🤔

Maybe not... 
image

I added it because it's so ubiquitous but maybe we leave it out in our case. There's very limited use of even javascript that I can find in the Blazor samples:
https://github.com/getsentry/sentry-dotnet/blob/e75d537c14ded617c99eb33f19cf4ce69ca4de36/samples/Sentry.Samples.AspNetCore.Blazor.Server/Pages/_Layout.cshtml#L31

I guess if we give the model less (unnecessary) context, it will behave better.

AGENTS.md
jamescrosswell: Hm, that instruction doesn't make much sense to me. Let's remove it.

jamescrosswell:
```suggestion
```

.claude/settings.json
```diff
@@ -30,5 +30,12 @@
...
+    "clangd-lsp@claude-plugins-official": true,
+    "jdtls-lsp@claude-plugins-official": true,
+    "swift-lsp@claude-plugins-official": true
```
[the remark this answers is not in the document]
jamescrosswell: That's not quite true... we have the Cocoa, Android and Native modules as git submodules so worth keeping those I think.

PR #5077, opened 2026-03-27
fix: `CaptureFeedback` now supports multiple attachments correctly

bitsandfoxes:
Follow up on: https://github.com/getsentry/sentry-dotnet/pull/3981

I could not find any source, neither in the dev docs nor on Sentry, for the "only one attachment per feeedback" restriction. The envelop created from a feedback can hold as many attachments as any other and will get processed by Sentry accordingly.

image

### DIFF
--- a/src/Sentry/Protocol/Envelopes/Envelope.cs
+++ b/src/Sentry/Protocol/Envelopes/Envelope.cs
@@ -315,6 +315,13 @@ public sealed class Envelope : ISerializable, IDisposable
         {
             foreach (var attachment in attachments)
             {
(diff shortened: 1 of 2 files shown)


bruno-garcia:
It was just copied from the JS SDK which does have this restriction, but that's just a browser/JS thing, not a protocol thing so we should remove.
It had come up before, I thought ti was already chaged. Thanks for doing this

jamescrosswell:
Yeah I think when I first implemented it the protocol wouldn't accept more than one Attachment. We did discuss that and the server was changed - we just never circled back to address the SDK. 

Thanks @bitsandfoxes !

PR #5084, opened 2026-03-30
fix: Workaround FileNotFoundException on Android when recovering sessions

### LINKED ISSUE #5083: Possible regression: Failed to persist session on the file system
> I think this might be back. I've had to disable auto tracking to avoid a crash. I thought a try catch would work, but doesn't:

```
System.IO.FileNotFoundException thrown
  at Interop.ThrowExceptionForIoErrno(Interop.ErrorInfo errorInfo, string path, bool isDirError) in System.Private.CoreLib.dll
  at Interop.CheckIo(Interop.Error error, string path, bool isDirError) in System.Private.CoreLib.dll
  at Microsoft.Win32.SafeHandles.SafeFileHandle.Open(string path, Interop.Sys.OpenFlags flags, int mode, bool failForSymlink, out bool wasSymlink, Func<Interop.ErrorInfo, Interop.Sys.OpenFlags, string, Exception> createOpenException) in System.Private.CoreLib.dll
  at Microsoft.Win32.SafeHandles.SafeFileHandle.Open(string fullPath, FileMode mode, FileAccess access, FileShare share, FileOptions options, long preallocationSize, UnixFileMode openPermissions, out long fileLength, out UnixFileMode filePermissions, bool failForSymlink, out bool wasSymlink, Func<Interop.ErrorInfo, Interop.Sys.OpenFlags, string, Exception> createOpenException) in System.Private.CoreLib.dll
  at Microsoft.Win32.SafeHandles.SafeFileHandle.Open(string fullPath, FileMode mode, FileAccess access, FileShare share, FileOptions options, long preallocationSize, UnixFileMode? unixCreateMode, Func<Interop.ErrorInfo, Interop.Sys.OpenFlags, string, Exception> createOpenException) in System.Private.CoreLib.dll
  at System.IO.Strategies.OSFileStreamStrategy..ctor(string path, FileMode mode, FileAccess access, FileShare share, FileOptions options, long preallocationSize, UnixFileMode? unixCreateMode) in System.Private.CoreLib.dll
  at System.IO.Strategies.UnixFileStreamStrategy..ctor(string path, FileMode mode, FileAccess access, FileShare share, FileOptions options, long preallocationSize, UnixFileMode? unixCreateMode) in System.Private.CoreLib.dll
  at System.IO.Strategies.FileStreamHelpers.ChooseStrategyCore(string path, FileMode mode, FileAccess access, FileShare share, FileOptions options, long preallocationSize, UnixFileMode? unixCreateMode) in System.Private.CoreLib.dll
  at System.IO.Strategies.FileStreamHelpers.ChooseStrategy(FileStream fileStream, string path, FileMode mode, FileAccess access, FileShare share, int bufferSize, FileOptions options, long preallocationSize, UnixFileMode? unixCreateMode) in System.Private.CoreLib.dll
  at System.IO.FileStream..ctor(string path, FileMode mode, FileAccess access, FileShare share, int bufferSize, FileOptions options, long preallocationSize) in System.Private.CoreLib.dll
  at System.IO.FileStream..ctor(string path, FileMode mode, FileAccess access, FileShare share, int bufferSize, FileOptions options) in System.Private.CoreLib.dll
  at System.IO.FileStream..ctor(string path, FileMode mode, FileAccess access, FileShare share, int bufferSize, bool useAsync) in System.Private.CoreLib.dll
  at System.IO.FileStream..ctor(string path, FileMode mode, FileAccess access, FileShare share) in System.Private.CoreLib.dll
  at System.IO.File.OpenRead(string path) in System.Private.CoreLib.dll
  at Sentry.Internal.FileSystemBase.OpenFileForReading(string path) in Sentry.dll
  at Sentry.Internal.Json.Load<PersistedSessionUpdate>(IFileSystem fileSystem, string filePath, Func<JsonElement, PersistedSessionUpdate> factory) in Sentry.dll
  at Sentry.GlobalSessionManager.<.ctor>b__11_0(string filePath) in Sentry.dll
  at Sentry.GlobalSessionManager.TryRecoverPersistedSession() in Sentry.dll
  at Sentry.Internal.Hub.StartSession() in Sentry.dll
  at Sentry.Integrations.AutoSessionTrackingIntegration.Register(IHub hub, SentryOptions options) in Sentry.dll
  at Sentry.Internal.Hub..ctor(SentryOptions options, ISentryClient client, ISessionManager sessionManager, ISystemClock clock, IInternalScopeManager scopeManager, RandomValuesFactory randomValuesFactory, IReplaySession replaySession, ISampleRandHelper sampleRandHelper, BackpressureMonitor backpressureMonitor) in Sentry.dll
  at Sentry.SentrySdk.InitHub(SentryOptions options) in Sentry.dll
  at Sentry.SentrySdk.Init(SentryOptions options) in Sentry.dll
  at Sentry.Maui.Internal.SentryMauiInitializer.Initialize(IServiceProvider services) in Sentry.Maui.dll
  at Microsoft.Maui.MauiContextExtensions.InitializeAppServices(MauiApp mauiApp) in //src/Core/src/MauiContextExtensions.cs:line 83
  at Microsoft.Maui.Hosting.MauiAppBuilder.Build() in //src/Core/src/Hosting/MauiAppBuilder.cs:line 198
  at CodeLogicMaster.MauiProgram.CreateMauiApp() in D:\AppsInStores\Maui\CodeLogicMaster\CodeLogicMaster\MauiProgram.cs:line 131
  at CodeLogicMaster.Platforms.Android.MainApplication.CreateMauiApp() in D:\AppsInStores\Maui\CodeLogicMaster\CodeLogicMaster\Platforms\Android\MainApplication.cs:line 14
  at Microsoft.Maui.MauiApplication.OnCreate() in /_/src/Core/src/Platform/Android/MauiApplication.cs:line 36
  at Android.App.Application.n_OnCreate(nint jnienv, nint native__this) in Mono.Android.dll
```  

 _Originally posted by @BillyMartin1964 in [#3653](https://github.com/getsentry/sentry-dotnet/issues/3653#issuecomment-4150476557)_

jamescrosswell:
Fixes #5083 
- #5083

### DIFF
--- a/src/Sentry/GlobalSessionManager.cs
+++ b/src/Sentry/GlobalSessionManager.cs
@@ -151,6 +151,12 @@ internal class GlobalSessionManager : ISessionManager
         }
 
         var filePath = Path.Combine(_persistenceDirectoryPath, PersistedSessionFileName);
+        if (!_options.FileSystem.FileExists(filePath))
+        {
+            _options.LogDebug("A persisted session file was not found at '{0}'.", filePath);
+            return null;
+        }
+
         try
         {
             var recoveredUpdate = _persistedSessionProvider(filePath);
@@ -194,12 +200,6 @@ internal class GlobalSessionManager : ISessionManager
                 recoveredUpdate.PendingUnhandled);
 
             return sessionUpdate;
-        }
-        catch (Exception ex) when (ex is FileNotFoundException or DirectoryNotFoundException)
-        {
-            // Not a notable error
-            _options.LogDebug("A persisted session does not exist ({0}) at {1}.", ex.GetType().Name, filePath);
-            return null;
         }
         catch (Exception ex)
         {
--- a/test/Sentry.Tests/GlobalSessionManagerTests.cs
+++ b/test/Sentry.Tests/GlobalSessionManagerTests.cs
@@ -264,7 +264,7 @@ public class GlobalSessionManagerTests
     public void TryRecoverPersistedSession_NoSessionFile_LogDebug()
     {
         // Arrange
-        _fixture.PersistedSessionProvider = _ => throw new FileNotFoundException();
+        _fixture.PersistedSessionProvider = _ => throw new UnreachableException("Unexpected attempt to read a file that does not exist.");
         var sut = _fixture.GetSut();
 
         // Act
--- a/test/Sentry.Tests/GlobalSessionManagerTests.cs
+++ b/test/Sentry.Tests/GlobalSessionManagerTests.cs
@@ -261,45 +261,28 @@ public class GlobalSessionManagerTests
     }
 
     [Fact]
-    public void TryRecoverPersistedSession_FileNotFoundException_LogDebug()
+    public void TryRecoverPersistedSession_NoSessionFile_LogDebug()
     {
         // Arrange
+        _fixture.PersistedSessionProvider = _ => throw new FileNotFoundException();
         var sut = _fixture.GetSut();
-        sut = new GlobalSessionManager(
-            _fixture.Options,
-            persistedSessionProvider: _ => throw new FileNotFoundException());
 
         // Act
         sut.TryRecoverPersistedSession();
 
         // Assert
-        _fixture.Logger.Entries.Should().Contain(e => e.Level == SentryLevel.Debug);
-    }
-
-    [Fact]
-    public void TryRecoverPersistedSession_DirectoryNotFoundException_LogDebug()
-    {
-        // Arrange
-        var sut = _fixture.GetSut();
-        sut = new GlobalSessionManager(
-            _fixture.Options,
-            persistedSessionProvider: _ => throw new DirectoryNotFoundException());
-
-        // Act
-        sut.TryRecoverPersistedSession();
-
-        // Assert
-        _fixture.Logger.Entries.Should().Contain(e => e.Level == SentryLevel.Debug);
+        _fixture.Logger.Entries.Should().Contain(e =>
+            e.Level == SentryLevel.Debug
+            && e.Message.Contains("A persisted session file was not found"));
     }
 
     [Fact]
     public void TryRecoverPersistedSession_EndOfStreamException_LogError()
     {
         // Arrange
+        _fixture.PersistedSessionProvider = _ => throw new EndOfStreamException();
         var sut = _fixture.GetSut();
-        sut = new GlobalSessionManager(
-            _fixture.Options,
-            persistedSessionProvider: _ => throw new EndOfStreamException());
+        sut.StartSession();
 
         // Act
         sut.TryRecoverPersistedSession();
@@ -383,14 +366,14 @@ public class GlobalSessionManagerTests
     public void TryRecoverPersistedSession_CrashDelegateReturnsTrueWithPauseTimestamp_EndsAsCrashed()
     {
         // Arrange
-        _fixture.Options.CrashedLastRun = () => true;
-        // Session was paused before persisted:
-        var pausedTimestamp = DateTimeOffset.Now;
+        var pausedTimestamp = DateTimeOffset.Now; // Session was paused before persisted:
         _fixture.PersistedSessionProvider = _ => new PersistedSessionUpdate(
             AnySessionUpdate(),
             pausedTimestamp);
+        _fixture.Options.CrashedLastRun = () => true;
 
         var sut = _fixture.GetSut();
+        sut.StartSession();
 
         // Act
         var persistedSessionUpdate = sut.TryRecoverPersistedSession();
@@ -412,6 +395,7 @@ public class GlobalSessionManagerTests
             pausedTimestamp);
 
         var sut = _fixture.GetSut();
+        sut.StartSession();
 
         // Act
         var persistedSessionUpdate = sut.TryRecoverPersistedSession();
--- a/test/Sentry.Tests/GlobalSessionManagerTests.cs
+++ b/test/Sentry.Tests/GlobalSessionManagerTests.cs
@@ -366,11 +366,12 @@ public class GlobalSessionManagerTests
     public void TryRecoverPersistedSession_CrashDelegateReturnsTrueWithPauseTimestamp_EndsAsCrashed()
     {
         // Arrange
-        var pausedTimestamp = DateTimeOffset.Now; // Session was paused before persisted:
+        _fixture.Options.CrashedLastRun = () => true;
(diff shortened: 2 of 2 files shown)


test/Sentry.Tests/GlobalSessionManagerTests.cs
```diff
@@ -261,45 +261,28 @@ public void TryRecoverPersistedSession_SessionNotStarted_ReturnsNull()
...
-            persistedSessionProvider: _ => throw new DirectoryNotFoundException());
-
-        // Act
-        sut.TryRecoverPersistedSession();
-
-        // Assert
-        _fixture.Logger.Entries.Should().Contain(e => e.Level == SentryLevel.Debug);
+        _fixture.Logger.Entries.Should().Contain(e =>
+            e.Level == SentryLevel.Debug
+            && e.Message.Contains("A persisted session file was not found"));
```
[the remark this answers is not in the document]
jamescrosswell: That's true... it's in there really just to make sure `_fixture.PersistedSessionProvider` never gets called. 

I think this is fine.

test/Sentry.Tests/GlobalSessionManagerTests.cs
```diff
@@ -261,45 +261,28 @@ public void TryRecoverPersistedSession_SessionNotStarted_ReturnsNull()
     }
 
     [Fact]
-    public void TryRecoverPersistedSession_FileNotFoundException_LogDebug()
+    public void TryRecoverPersistedSession_NoSessionFile_LogDebug()
     {
         // Arrange
+        _fixture.PersistedSessionProvider = _ => throw new FileNotFoundException();
```
jamescrosswell:
```suggestion
        _fixture.PersistedSessionProvider = _ => throw new FileNotFoundException(); // This should never get called - just to make sure though, we throw an/any exception here
```

jamescrosswell: @Flash0ver we could add this comment... could potentially remove the code altogether but I think it provides some assurance the `_fixture.PersistedSessionProvider` never gets called and that the correct code path is taken.

src/Sentry/GlobalSessionManager.cs
```diff
@@ -151,6 +151,12 @@ private void DeletePersistedSession()
         }
 
         var filePath = Path.Combine(_persistenceDirectoryPath, PersistedSessionFileName);
+        if (!_options.FileSystem.FileExists(filePath))
+        {
+            _options.LogDebug("A persisted session file was not found at '{0}'.", filePath);
+            return null;
+        }
```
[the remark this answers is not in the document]
jamescrosswell: Overkill I think. Worst case, we log something at error level rather than debug... but these are all just debug logs anyway (ignored in production). If someone has Debug mode enabled, they probably want to know about weird race conditions like this, which we don't really anticipate happening in normal execution.

test/Sentry.Tests/GlobalSessionManagerTests.cs
Flash0ver: We could also, instead of throwing `FileNotFoundException`, throw the `net7.0`-added `System.Diagnostics.UnreachableException`, that we also polyfill, which is the now-canonical way of indicating that a certain code path is not expected to be executed ... and potentially add additional context to the `UnreachableException`-`Message`:

```suggestion
        _fixture.PersistedSessionProvider = _ => throw new UnreachableException("Unexpected attempt to read a file that does not exist.");
```

Because this exception should never be caught and we cannot accidentally "mask"/"hide" unexpected behavior.

src/Sentry/GlobalSessionManager.cs
```diff
@@ -195,12 +201,6 @@ private void DeletePersistedSession()
 
             return sessionUpdate;
         }
-        catch (Exception ex) when (ex is FileNotFoundException or DirectoryNotFoundException)
-        {
-            // Not a notable error
-            _options.LogDebug("A persisted session does not exist ({0}) at {1}.", ex.GetType().Name, filePath);
-            return null;
-        }
```
Flash0ver, line 203: thought: I wonder if it's "worth" changing the behavior for all targets for an issue on Android only.

Also considering, that with .NET 11, the default runtime for .NET Android apps is _CoreCLR_ ([.NET 11 Preview 1 Release Notes](https://github.com/dotnet/core/blob/main/release-notes/11.0/preview/preview1/dotnetmaui.md#coreclr-by-default)) ... but actually could still be switched back to _Mono_ ... assuming that this is a Mono-Runtime issue combined with AOT.

I am thinking about keeping this code, and do the above `FileExists` check preprocessor-compiled on Android only.

Because technically - see https://github.com/getsentry/sentry-dotnet/pull/5084/changes#r3007303264 - there could be a discrepancy between File-Exists-check and File-Read-operation.

But yes ... to be fair ... I may be totally overthinking this 😉.

test/Sentry.Tests/GlobalSessionManagerTests.cs
```diff
@@ -391,6 +374,7 @@ public void TryRecoverPersistedSession_CrashDelegateReturnsTrueWithPauseTimestam
             pausedTimestamp);
 
         var sut = _fixture.GetSut();
+        sut.StartSession();
```
Flash0ver, line 377: question: Why do we need to start sessions now, but didn't before?

src/Sentry/GlobalSessionManager.cs
jamescrosswell: I still stand by my reply to that comment... it seems overkill to use exceptions for logic handling (with all the expense of creating stack traces that comes with) when:
- we can simply check for the existence of the file directly
- the worst case scenarion (in the unlikely event of a race conditoin) is that the log level is elevated from Debug to Error (and we're talking about debug logs, that would only ever show if someone had debug logging enabled). 

Definitely not worth making an extra commit on the PR for... just slows us down and gains us nothing.

test/Sentry.Tests/GlobalSessionManagerTests.cs
jamescrosswell: Because of the early exit if no session file exists. We probably should have been doing this from the start, but there was a silent try..catch block in there hiding the behaviour.

test/Sentry.Tests/GlobalSessionManagerTests.cs
jamescrosswell: Nice - that's a better choice 👍🏻

jpnurmi (approved):

Issue #5095, opened 2026-03-31
iOS device tests timeout when installing the test app


jamescrosswell:
### Description

This happens muitiple times a day (e.g. [here](https://github.com/getsentry/sentry-dotnet/actions/runs/23781536438), [here](https://github.com/getsentry/sentry-dotnet/actions/runs/23774837669/job/69274275191) and [here](https://github.com/getsentry/sentry-dotnet/actions/runs/23771697921/job/69264441771)). 

The message is always the same:
> [09:11:58.2496530] Xamarin.Hosting: Installing on iPhone 17 (9D89631E-DA6A-4E6B-8CE7-CFBE4BCC8391) by executing 'xcrun simctl install 9D89631E-DA6A-4E6B-8CE7-CFBE4BCC8391 /Users/runner/work/sentry-dotnet/sentry-dotnet/test/Sentry.Maui.Device.TestApp/bin/Release/net10.0-ios/iossimulator-arm64/Sentry.Maui.Device.TestApp.app'
[09:18:11.1935550] Killing process 22259 as it was cancelled

So timing out when running `xcrun simctl install`. 

# Possible solutions

## iOS version

[This user](https://github.com/actions/runner-images/issues/7971#issuecomment-1661932074) indicated they only had the problem on iOS 17.0 and resolved it by forcing the simulator to use an older version of iOS.

## Timeout command wrapper or active user session

Gemini has this to say about it, so possible worth exploring:
- **CI/CD specific issues**: In continuous integration environments (like CircleCI or GitHub Actions), simulators might hang if there is no active graphical user session on the machine.
  - **Solution**: Workarounds often involve using a timeout command wrapper for the boot command or ensuring a user session is active. 
    - https://discuss.circleci.com/t/xcode-26-rc/54066/18

# Previous failed attempts to fix this
- #5063

PR #5105, opened 2026-04-01
ci: disable host crash reporter for iOS simulator tests

jpnurmi:
Just to see if it might also help to disable the irrelevant and CPU-hungry crash reporter service on the host, also responsible for popping up those annoying crash dialogs on the screen...

Related to: #5095


jpnurmi:
Seems hard to make it fail after https://github.com/getsentry/sentry-dotnet/pull/5096 so this is probably not needed, yay!

PR #5126, opened 2026-04-09
fix: prevent redundant native exceptions on iOS

jpnurmi:
Fixes duplicate native exceptions (#3954) on iOS by reordering signal handlers to let .NET/Mono handle managed exceptions first, and then chain real native exceptions to Sentry.

**Before**:
```
           ┌──────────────┐     ┌───────────┐     ┌────────┐
Signal────>│ Sentry Cocoa │────>│ .NET/Mono │────>│ System │
           └──────────────┘     └───────────┘     └────────┘
```

**After**:
```
           ┌───────────┐     ┌──────────────┐     ┌────────┐
Signal────>│ .NET/Mono │────>│ Sentry Cocoa │────>│ System │
           └───────────┘     └──────────────┘     └────────┘
```

PR #4956, opened 2026-02-25
fix: MSB4006 circular dependency involving target "SetupCocoaSDK"

jamescrosswell:
Resolves #4955:
- #4955

## Explanation

Me and Claude spent a bit of time looking at this. I think we now have a solution that is OK.

### What the original code did

The recursive MSBuild call with `RemoveProperties="TargetFramework" `meant that during an inner build, instead of running `_SetupCocoaSDK` directly, it spawned a separate MSBuild process on the same project with no TargetFramework set. That separate process ran `_SetupCocoaSDK` once in isolation. So across both inner TFM builds, the actual work happened in one subprocess — not in the inner builds themselves.

### What the new code does

`_SetupCocoaSDK` now runs once per inner build (once for `net9.0-ios18.0`, once for `net9.0-maccatalyst18.0`). 

However, every sub-target of `_SetupCocoaSDK` is idempotent and has checks to make sure we're not double handling:
- `_DownloadCocoaSDK` — has Condition="... And !Exists('$(SentryCocoaFramework)')", so ,skips if already downloaded
- `_BuildCocoaSDK` — has Inputs/Outputs incremental build guards, skips if up-to-date
- `_GenerateSentryCocoaBindings` — has Inputs/Outputs, skips if up-to-date
- `SanitizeSentryCocoaFramework` — has "!Exists('$(SentryCocoaFramework).sanitized.stamp"

#skip-changelog

PR #5131, opened 2026-04-13
build: switch to sentry-cocoa submodule for managed builds

jpnurmi:
Allows building sentry-cocoa with `SENTRY_CRASH_MANAGED_RUNTIME` for eliminating duplicate native exceptions on iOS:
- getsentry/sentry-dotnet#5126

Note: The old `.properties` -based release download is left intact to make it easy to switch back in the future if we ever get an official managed build variant...

#skip-changelog

### DIFF
--- a/scripts/build-sentry-cocoa.sh
+++ b/scripts/build-sentry-cocoa.sh
@@ -5,19 +5,23 @@ pushd "$(dirname "$0")" >/dev/null
 cd ../modules/sentry-cocoa
 
 mkdir -p Carthage
-PID_FILE="$PWD/Carthage/.build.pid"
-trap 'if [[ "$(cat "$PID_FILE" 2>/dev/null)" == "$$" ]]; then rm -f "$PID_FILE"; fi' EXIT
+LOCK_DIR="$PWD/Carthage/.build.lock"
+trap 'if [[ "$(cat "$LOCK_DIR/pid" 2>/dev/null)" == "$$" ]]; then rm -rf "$LOCK_DIR"; fi' EXIT
 
 # Serialize concurrent invocations; parallel xcodebuilds race on DerivedData.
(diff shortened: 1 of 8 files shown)


scripts/build-sentry-cocoa.sh
```diff
@@ -4,7 +4,32 @@ set -euo pipefail
...
+        echo "Previous build did not complete (pid $build_pid); cleaning up and retrying" >&2
+        # Atomically claim the stale dir via rename
+        if mv "$LOCK_DIR" "$LOCK_DIR.stale.$$" 2>/dev/null; then
+            rm -rf "$LOCK_DIR.stale.$$"
+        fi
+        continue
+    fi
+    sleep 2
+done
+echo $$ > "$LOCK_DIR/pid"
```
jpnurmi: Avoiding parallel per-TFM calls from MSBuild is painful (#4956), so this attempts to solve parallelization conflicts in the build script instead...

scripts/build-sentry-cocoa.sh
jamescrosswell, line 1: Fancy Mutex 😜

jamescrosswell (approved):
I guess there's no running from it. 

I did a first pass on all the MS Build stuff and didn't see any issues. Main thing is that the builds still work and CI is green. I saw you put a pretty robust MUTEX in there to prevent multi-targeting build issues, so let's give it a go.

scripts/build-sentry-cocoa.sh
jpnurmi: Heh, it started out much simpler. The AI hardening process made it bulletproof, which is great, but also a bit more... verbose. 🙃

jpnurmi:
@jamescrosswell @Flash0ver Heads up, this will need a manual `rm -rf modules/sentry-cocoa` when pulled into existing clones and worktrees. ✌️

PR #5133, opened 2026-04-14
fix: ConcurrentBag memory leak (shows whhen profiling is enabled)

### LINKED ISSUE #4721: Replace applicable `ConcurrentBag` usages with custom lightweight thread-safe collection
### Description

In #3431 (fixed by #3432) and #4709 (fixed by #4717),
we noticed that _Transaction_ and _Span_ types utilizing [System.Collections.Concurrent.ConcurrentBag<T>](https://learn.microsoft.com/dotnet/api/system.collections.concurrent.concurrentbag-1) may cause **memory leaks** in our usages.
Most `System.Collections.Concurrent` collections have additional semantics and features, that we don't really need in most use cases (like `IProducerConsumerCollection<T>`, as mentioned in https://github.com/getsentry/sentry-dotnet/pull/3432#issuecomment-2175807165 and https://github.com/getsentry/sentry-dotnet/pull/3432#discussion_r1645214211.
In most cases, we just need a thread-safe code path of adding new items to a collection, and later iterating through that collection, in a concurrent fashion.

Some potential ideas
1. use `BlockingCollection<T>` instead
   - alleviates the memory leak for our _Transaction_ and _Span_ types
   - but has semantics of "Completion" that we don't need in most cases
2. build a simple custom concurrent Collection, backed by a `List<T>`, that uses a blocking `lock` for synchronization
   - quick implementation
   - synchronous blocking is not quite ideal
     - especially when trying to add while already being concurrently enumerated
4. build a complex custom concurrent Collection, that is backed by an `Array` and behaves similar to the double-in-size growing algorithm of a `List<T>` when adding new items, based on `Interlocked.Increment` to determine the index to add to
   - non-blocking when non-growing (i.e. adding when the `Count` is less than the `Capacity`)
   - blocking when growing (i.e. adding when the `Count` equals the `Capacity`)

### LINKED ISSUE #5113: Sentry profiler seems to be leaking memory
### Package

Sentry.Profiling

### .NET Flavor

.NET

### .NET Version

10.0.104

### OS

Any (not platform specific)

### OS Version

Any

### Development Environment

Rider 2025.x (Windows)

No

_No response_

### SDK Version

6.3.0

6.1.0

### Workload Versions

N/A

//...
x.AddProfilingIntegration();
//...

### Steps to Reproduce

1. Enable Sentry Profiler Integration via AddProfilerIntegration()
2. Deploy to production for real scale workload
3. Watch memory climb until OOM issues

### Expected Result

It shouldn't :P

### Actual Result

We seem to have found memory leaks in the Sentry Profiler integration added via `AddProfilingIntegration()` due to relying on `Microsoft.Diagnostics.Tracing.Etlx`. The cause seems to be from undisposed instances of `TraceLog`, with several gigabytes of call stacks accumulating over the period of roughly an hour for us, eventually causing process memory starvation and crashes.

jamescrosswell:
fixes: #5113
fixes: #4721

Fixes memory leaks caused by `ConcurrentBag` and `ConcurrentQueue`.

## Why ConcurrentBag leaks
ConcurrentBag is backed by a ThreadLocal<ThreadLocalList> — internally it keeps a per-thread linked list. Every thread that ever touches the bag allocates a ThreadLocalList node that stays attached to the bag for the bag's lifetime, even after that thread has died. In a high-throughput app (profiler + ASP.NET Core thread pool), each bag instance accumulates stranded per-thread slots, and each SdkVersion instance carries that weight.

## Why the profiler amplifies it
Every SentryEvent, SentryTransaction, Scope, and TransactionTracer allocates its own SdkVersion → two fresh ConcurrentBag instances. AddPackage gets called on most of these from multiple callsites (SentryMiddleware.cs:252, Enricher.cs:72, Scope.cs:527). The profiler drives the event/transaction throughput way up, so both the count of bags and the per-bag thread-local overhead explode.

There's also one hotter target: NoOpTransaction.cs:14 returns SdkVersion.Instance (the singleton) as its Sdk. Any code path that ends up calling AddPackage on a NoOpTransaction's Sdk writes into that singleton bag forever.

## The fix pattern
Same as https://github.com/getsentry/sentry-dotnet/pull/3432: replace `ConcurrentBag` with a simpler lock-backed list (as suggested in https://github.com/getsentry/sentry-dotnet/issues/4721).

# Testing

Memory does keep growing in the process but this seems to be just because I have loads of memory so the GC doesn't bother kicking in. If I force a GC then memory is released again:
image

As far as I can tell then, there are no remaining memory leaks in the [minimal repro](https://github.com/getsentry/sentry-dotnet/issues/5113#issuecomment-4225368488) that @kanadaj provided.

### DIFF
--- a/src/Sentry/Internal/ConcurrentBagLite.cs
+++ b/src/Sentry/Internal/ConcurrentBagLite.cs
@@ -6,7 +6,7 @@ namespace Sentry.Internal;
 /// We're using this to avoid the same class of memory leak that <see cref="ConcurrentQueueLite{T}"/>
 /// was introduced to avoid. See https://github.com/getsentry/sentry-dotnet/issues/5113
 /// </summary>
-internal class ConcurrentBagLite<T> : IEnumerable<T>
+internal class ConcurrentBagLite<T> : IReadOnlyCollection<T>
 {
     private readonly List<T> _items;
 
--- a/src/Sentry/Internal/ConcurrentBagLite.cs
+++ b/src/Sentry/Internal/ConcurrentBagLite.cs
@@ -1,0 +1,73 @@
+namespace Sentry.Internal;
+
+/// <summary>
+/// A minimal replacement for <see cref="ConcurrentBag{T}"/>.
+///
+/// We're using this to avoid the same class of memory leak that <see cref="ConcurrentQueueLite{T}"/>
+/// was introduced to avoid. See https://github.com/getsentry/sentry-dotnet/issues/5113
+/// </summary>
+internal class ConcurrentBagLite<T> : IEnumerable<T>
+{
+    private readonly List<T> _items;
+
+    public ConcurrentBagLite()
+    {
+        _items = new List<T>();
+    }
+
+    public ConcurrentBagLite(IEnumerable<T> collection)
+    {
+        _items = new List<T>(collection);
+    }
+
+    public void Add(T item)
+    {
+        lock (_items)
+        {
+            _items.Add(item);
+        }
+    }
+
+    public int Count
+    {
+        get
+        {
+            lock (_items)
+            {
+                return _items.Count;
+            }
+        }
+    }
+
+    public bool IsEmpty => Count == 0;
+
+    public void Clear()
+    {
+        lock (_items)
+        {
+            _items.Clear();
+        }
+    }
+
+    public T[] ToArray()
+    {
+        lock (_items)
+        {
+            return _items.ToArray();
+        }
+    }
+
+    public IEnumerator<T> GetEnumerator()
+    {
+        // Return a snapshot to avoid holding the lock during iteration
+        // and to prevent InvalidOperationException if the collection is modified.
+        T[] snapshot;
+        lock (_items)
+        {
+            snapshot = _items.ToArray();
+        }
+        return ((IEnumerable<T>)snapshot).GetEnumerator();
+    }
+
+    IEnumerator IEnumerable.GetEnumerator() => GetEnumerator();
+}
--- a/test/Sentry.DiagnosticSource.Tests/SentrySqlListenerTests.cs
+++ b/test/Sentry.DiagnosticSource.Tests/SentrySqlListenerTests.cs
@@ -120,11 +120,10 @@ public class SentrySqlListenerTests
                 }));
 
         // Assert
-        var spans = fixture.Spans.Where(s => s.Operation != "abc");
+        var spans = fixture.Spans.Where(s => s.Operation != "abc").ToList();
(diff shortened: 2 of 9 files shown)


test/Sentry.DiagnosticSource.Tests/SentrySqlListenerTests.cs
jamescrosswell, line 1: The SQL listener test had a bug that the LIFO mask would hide. The test did `OrderByDescending(x => x.StartTimestamp).First()` to get the "newest" span, but this fails whenever two spans share the same
timestamp — which is common on Windows x64 due to ~15ms clock resolution. Making `ConcurrentBagLite` LIFO would make the test pass again accidentally, because the newer span would happen to enumerate first and
survive the stable sort... so changing/fixing the test instead.

src/Sentry/Internal/ConcurrentBagLite.cs
```diff
@@ -0,0 +1,73 @@
...
+    }
+
+    public ConcurrentBagLite(IEnumerable<T> collection)
+    {
+        _items = new List<T>(collection);
+    }
+
+    public void Add(T item)
+    {
+        lock (_items)
```
vaind, line 25: Worth checking whether we have a hotspot where `add(IEnumearble<T> items)` would be useful.

vaind (approved):

src/Sentry/Internal/ConcurrentBagLite.cs
jamescrosswell: Good call... we have these:
https://github.com/getsentry/sentry-dotnet/blob/a16638f9b2c1774686d0a15389a0941b53665926/src/Sentry/Scope.cs#L530-L534
https://github.com/getsentry/sentry-dotnet/blob/a16638f9b2c1774686d0a15389a0941b53665926/src/Sentry/Scope.cs#L555-L568
https://github.com/getsentry/sentry-dotnet/blob/a16638f9b2c1774686d0a15389a0941b53665926/src/Sentry/Scope.cs#L687-L690
https://github.com/getsentry/sentry-dotnet/blob/a16638f9b2c1774686d0a15389a0941b53665926/src/Sentry/Scope.cs#L713-L716
https://github.com/getsentry/sentry-dotnet/blob/a16638f9b2c1774686d0a15389a0941b53665926/src/Sentry/Scope.cs#L739-L742

I don't think I'd describe any of those as a hot path though... with the exception of the first one (attachments) it's all one off stuff that happens at init for a very constrained number of items - and even attachments is likely to be only a handful of items.

I think I'll leave it for now then, in the interests of expediting the fix to the memory leak.

PR #5134, opened 2026-04-16
chore: fix missing skill

jamescrosswell:
Seeing the following error:

```
./dev.cs aiup
[dev] Installing/updating AI agent files
==> npx @sentry/dotagents install: npx @sentry/dotagents install
Failed to resolve skill "create-pr": Skill "create-pr" not found in getsentry/skills. Tried conventional directories. Use the 'path' field to specify the location explicitly.
```

Seems to be because this skill is missing so removing it from `agents.toml`

### DIFF
--- a/agents.toml
+++ b/agents.toml
@@ -34,10 +34,6 @@ name = "create-branch"
 source = "getsentry/skills"
 
 [[skills]]
-name = "create-pr"
-source = "getsentry/skills"
-
-[[skills]]
 name = "find-bugs"
 source = "getsentry/skills"
 


jpnurmi (approved):
The `create-pr` alias was removed in:
- https://github.com/getsentry/skills/pull/110

It was an alias for `pr-writer` that is already listed:
https://github.com/getsentry/sentry-dotnet/blob/cdd5e8584e03709cd68c78d3480c5712f42ac54b/agents.toml#L48-L50

PR #5146, opened 2026-04-21
chore: move Benchmarks for Sentry.Extensions.Logging into separate project

Flash0ver:
### Summary

Move the Benchmark for `Sentry.Extensions.Logging` into a new `Sentry.Extensions.Logging.Benchmarks`project.

### Remarks

In #4308, I broke the `SentryCore.slnf` filtered solution, by adding a dependency to `Sentry.Extensions.Logging` in the `Sentry.Benchmarks` project.
See https://github.com/getsentry/sentry-dotnet/commit/eadf5f20b8014745d0fe42b0e707bf3f9eec9ffc#diff-a68c64a5b74a24eb0d8d7d04aeb893acd2358768b2e0efe3a747d79199a0bea1R17.

Fixing this problem (without pulling in a bunch of projects into `SentryCore.slnf`) by separating the Benchmarks for `Microsoft.Extensions.Logging` into a new separate Benchmarking project.

This structure also now mimics `/tests/`, where we have one Test-Project per Project-Under-Test.


benchmarks/Directory.Build.props
```diff
@@ -2,7 +2,6 @@
 
   <Import Project="..\Directory.Build.props" />
 
-  <!-- Never use Sentry CLI for benchmark projects. -->
```
Flash0ver, line 5: note: remove redundant comment

`<UseSentryCLI>false</UseSentryCLI>` was removed via https://github.com/getsentry/sentry-dotnet/commit/91b22a9a1bdc790a1bb1b9934b176db8d517244b#diff-f431a929d80ac5e052346ee209d3600c6f7385325f3811b74dbba9bf66c21054L7, but the comment remained.

benchmarks/Sentry.Benchmarks/Sentry.Benchmarks.csproj
```diff
@@ -3,7 +3,6 @@
   <PropertyGroup>
     <OutputType>Exe</OutputType>
     <TargetFramework>$(PreviousTfm)</TargetFramework>
-    <IsPackable>false</IsPackable>
```
Flash0ver, line 6: note: removing, as already declared in "parent" `Directory.Build.props`

benchmarks/Sentry.Extensions.Logging.Benchmarks/Program.cs
```diff
@@ -0,0 +1,23 @@
+using BenchmarkDotNet.Configs;
+using BenchmarkDotNet.Diagnosers;
+using BenchmarkDotNet.Exporters;
+using BenchmarkDotNet.Running;
+
+namespace Sentry.Extensions.Logging.Benchmarks;
+
+internal class Program
```
Flash0ver, line 8: thought: code duplication

This file pretty much is a duplicate from the existing `benchmarks/Sentry.Benchmarks/Program.cs`, except for the namespace ... which we could argue is fine to keep in the default namespace of the project when using Top-Level-Statements instead.

Should we have more Benchmark-Projects in the future, we could have a single `.cs` file, that we `<Compile Include="..." />` across all projects.
Or we do that right away already.

jamescrosswell (approved):

PR #5159, opened 2026-04-23
feat: Add PreferTransactionNameProvider option

### LINKED ISSUE #5123: TransactionNameProvider is never called for routed requests
## Summary

`TransactionNameProvider` is only invoked when the transaction name is empty (i.e. routing failed entirely). When ASP.NET Core routing resolves a route template successfully, the provider is silently skipped — making it impossible to partially materialize route parameters in the transaction name without writing a custom middleware.

## Current behavior

In `SentryTracingMiddleware`, the `TransactionNameProvider` delegate is guarded by:

```csharp
if (transaction.Name == string.Empty)
{
    transaction.Name = options.TransactionNameProvider?.Invoke(ctx) ?? ...;
}
```

Because routing resolves route templates like `GET /api/v4/objects/{objectTypeName}/{objectId}/{operation}`, `transaction.Name` is already populated before the provider is consulted. The delegate is never called.

## Gap

There is no SDK-level hook that runs unconditionally after routing resolves and allows the transaction name to be rewritten. The full matrix of available hooks:

| Hook | Status |
|---|---|
| `TransactionNameProvider` | Only fires when routing fails (name is empty) — unusable for routed requests |
| `SetBeforeSendTransaction` | `transaction.Name` is `private set`; `NameSource` has no setter — read-only |
| `ITransactionProcessor` / `AddTransactionProcessor` | Same constraint — name is not writable |

The practical workaround today is to inject a custom ASP.NET middleware that runs after routing and rewrites the transaction name via `SentrySdk.GetSpan()` or similar. This is undiscoverable and fragile.

A related issue (#3106) was closed with the same custom-middleware workaround.

## Options

**Option A — Unconditional `TransactionNameProvider` invocation**
Remove (or make optional) the `if (transaction.Name == string.Empty)` guard. Let the provider run on every routed request. A null return falls back to the SDK default. This is the smallest possible change and preserves backward compatibility for users who don't configure a provider.

**Option B — Expose a mutable `ITransactionContext` in `SetBeforeSendTransaction`**
Add a writable `Name` and `NameSource` to the type passed into `SetBeforeSendTransaction` / transaction processors. This is a broader API surface change but gives maximum flexibility.

**Option C — Add a dedicated `TransactionProcessor` hook with mutation support**
Introduce a post-routing pipeline hook (e.g. `OnTransactionNameResolved`) that receives the resolved route data and can return a replacement name.

## Recommendation

Option A is the lowest-risk path: make `TransactionNameProvider` fire unconditionally and treat a null return as "keep the SDK default." This directly fixes the documented (but broken) use case with no breaking API change.

Action taken on behalf of a Sentry support escalation.

jamescrosswell:
Resolves #5123
- #5123


AGENTS.md
```diff
@@ -127,7 +127,8 @@ Public API diffs are stored as Verify snapshot files. After any public API chang
 ## Changelog
 
 - If the change is not user-facing, add `#skip-changelog` to the PR description
-- Otherwise generated automatically from [Commit message conventions](https://develop.sentry.dev/engineering-practices/commit-messages/)
+- Otherwise generated automatically from [Commit message conventions](https://develop.sentry.dev/engineering-practices/commit-messages/).
+  - Do **NOT** add changelogs manually.
```
jamescrosswell, line 131: Not specifically related to this PR but I ran into it on this run. Can extract to a separate PR if we really want.

src/Sentry.AspNetCore/SentryAspNetCoreOptions.cs
```diff
@@ -63,6 +63,20 @@ public class SentryAspNetCoreOptions : SentryLoggingOptions
...
+    /// </summary>
+    /// <remarks>
+    /// When <c>false</c> (default), the provider is only invoked as a fallback for unresolved routes —
+    /// matching the legacy behaviour. When <c>true</c>, the provider is invoked for every request
+    /// after routing; a non-empty return value sets the transaction name. If the provider returns
+    /// <c>null</c> or empty for a routed request, the name falls back to the URL path (the route
+    /// template is <b>not</b> retained); callers that want to preserve the route template should
+    /// return it themselves from the provider.
+    /// </remarks>
+    public bool AlwaysCallTransactionNameProvider { get; set; }
```
Flash0ver: question: temporary until next major?

I'm wondering if I'm understanding it right:
Is this a temporary option for the current major-cycle to not introduce behavior-breaking changes, and we will wither then
- flip with the next major (enabled by default but possible opt-out) ... and then remove in the major after that
- or remove in the next major directly

This thought is now "escalating" in a general question:
How to deal with temporary opt-ins that we make the default in the next major, and potentially even remove then in the next major?

I think what .NET does is not doing this via .NET APIs (since the approach of the .NET Framework is that "APIs are forever" ... and instead introduce MSBuild properties (like Runtime Async's `<Features>runtime-async=on</Features>`),
or Runtime Feature Switches,
or Environment Variables.

On the other hand, unlike the .NET Framework, we do have a single entry point via Init(Options)/AddSentry(Options)/`UseSentry(Options)`, which is the canonical way of configuring our SDK ... but then I'm thinking if we should "mark" such "temporary opt-ins that might become the default in an upcoming major" in a particular way ... like putting them in an `Experimental` namespace/type ... but that also doesn't quite fit as it's not `Experimental`, but _Temporary_ ... for the current major-cycle ... hmmm ...

I see that I am a bit uncertain here.
Where are your thoughts?

jamescrosswell: A temporary opt in, moving to a temporary opt out is a fairly straight forward way to let SDK customers easily do what they want... unless there are some compelling/damning arguments against it. What is the concern? Like what would be the worst case scenario?

About the most serious consequence that I can imagine is that some SDK customers might have to set a new option (either in csharp or in an appsettings.json file) when they bump to v7...

src/Sentry.AspNetCore/SentryTracingMiddleware.cs
```diff
@@ -178,13 +178,15 @@ public async Task InvokeAsync(HttpContext context)
...
                 {
                     var method = context.Request.Method.ToUpperInvariant();
 
                     // If we've set a TransactionNameProvider, use that here
-                    var customTransactionName = context.TryGetCustomTransactionName();
-                    if (!string.IsNullOrEmpty(customTransactionName))
+                    if (customName.Value is { } customTransactionName)
                     {
                         transaction.Name = $"{method} {customTransactionName}";
                         tracer.NameSource = TransactionNameSource.Custom;
```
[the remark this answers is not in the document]
jamescrosswell: This appears to be referencing stale code... this is the new code:
https://github.com/getsentry/sentry-dotnet/blob/b484f6d977cca4c9555f64db3b43af9bc35fee0e/src/Sentry.AspNetCore/SentryTracingMiddleware.cs#L180-L202

AGENTS.md
Flash0ver: IMO, **small** updates to automation do make sense to bundle in one changeset so that later - when using git annotate - we have the context for which the change was indicated

src/Sentry.AspNetCore/SentryTracingMiddleware.cs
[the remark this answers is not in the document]
jamescrosswell: This is something that can be controlled via the implementation of the `customTransactionName` function that the SDK user supplies (they can check if a value exists and return `null` if they don't want to override this). I think it's more flexible if we leave it like this - otherwise they wouldn't be able to use the customTransactionName to override values. 

@Flash0ver  thoughts?

src/Sentry.AspNetCore/SentryAspNetCoreOptions.cs
Flash0ver: Not really a concrete concern ... more of a general question.

We actually can do source-breaking changes with major releases, where .NET libraries rarely do (and then usually caused by language changes). And the advantage of declaring these on `SentryOptions` is discoverability.

Thanks.

src/Sentry.AspNetCore/SentryAspNetCoreOptions.cs
```diff
@@ -63,6 +63,19 @@ public class SentryAspNetCoreOptions : SentryLoggingOptions
...
+    /// when routing fails to resolve a template.
+    /// </summary>
+    /// <remarks>
+    /// When <c>false</c> (default), the provider is only invoked as a fallback for unresolved routes —
+    /// matching the legacy behaviour. When <c>true</c>, the provider is invoked for every request
+    /// after routing; a non-empty return value overrides the route-derived name. If the provider
+    /// returns <c>null</c> or <see cref="string.Empty"/>, the routed name is preserved (or the URL-path
+    /// fallback applies when no route was resolved).
+    /// </remarks>
+    public bool AlwaysCallTransactionNameProvider { get; set; }
```
Flash0ver: thought: rename

I'm not sure I like the name:

To me, `AlwaysCall..` reads more of an implementation detail, in particular the `..Call..` part, rather than the semantic effect.

I'll try some alternatives, let me know what you think:
- `OverrideTransactionName` / `OverwriteTransactionName`
  - doesn't imply the use of the user-provided `TransactionNameProvider`
- `OverrideTransactionNameWithTransactionNameProvider`
  - too long
- `PreferTransactionNameProvider`
  - implies the `TransactionNameProvider`, and also precedence
- `ForceTransactionNameProvider`
  - might be too strong, because when the user-delegate returns `null`, it's not overwritten

I kinda like `PreferTransactionNameProvider`.
What do you think?
Or should we stick with `AlwaysCallTransactionNameProvider`?

src/Sentry.AspNetCore/SentryTracingMiddleware.cs
```diff
@@ -178,13 +178,15 @@ public async Task InvokeAsync(HttpContext context)
                 var status = SpanStatusConverter.FromHttpStatusCode(context.Response.StatusCode);
 
                 // If no Name was found for Transaction, then we don't have the route.
-                if (transaction.Name == string.Empty)
+                // Also, run this block if the caller has opted into always calling the TransactionNameProvider.
+                var customName = new Lazy<string?>(context.TryGetCustomTransactionName);
+                var forceCustomName = _options.AlwaysCallTransactionNameProvider && customName.Value is not null;
+                if (transaction.Name.Length == 0 || forceCustomName)
```
Flash0ver: thought: I find the `Lazy` here a bit hard to read

And it's also a bit of an overkill, since there's no thread contention.
If I get it right, it's used as a cache to run the user-callback only once.

I wonder if we can rewrite it a bit without the `Lazy<>` 🤔.
If it would be harder to read without the `Lazy<>`, then let's use the hint that it is not accessed concurrently:

```CSharp
new Lazy<string?>(context.TryGetCustomTransactionName, LazyThreadSafetyMode.None)
```

Flash0ver: question: string-check ... potential NRE

The "TransactionName"-check changed:
```diff
-transaction.Name == string.Empty
+transaction.Name.Length == 0
```

From a defensive perspective:
Are we sure that `transaction.Name` here is never `null`?

src/Sentry.AspNetCore/SentryTracingMiddleware.cs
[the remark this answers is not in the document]
Flash0ver: Hmm ... how could the user control this?

With the `TransactionNameProvider` receiving `HttpContext` as in and returning `string?` as out, how could the user delegate perform this check without the actual `ITransactionTracer`?

src/Sentry.AspNetCore/SentryTracingMiddleware.cs
jamescrosswell: Yeah it's basically to avoid resolving it twice (lines 183 and lines 189) and also avoid resolving it at all  if the transaction.Name is not empty and PreferTransactionNameProvider is false. 

I didn't find any more elegant ways to do this (the alternatives were all basically me rolling out my own version of Lazy). I've added `LazyThreadSafetyMode.None`.

jamescrosswell: I'll revert.

jamescrosswell: [efa2d17](https://github.com/getsentry/sentry-dotnet/pull/5159/commits/efa2d17f0c6bc7a51506b1aca6d4228822e68b57)

jamescrosswell: [efa2d17](https://github.com/getsentry/sentry-dotnet/pull/5159/commits/efa2d17f0c6bc7a51506b1aca6d4228822e68b57)

src/Sentry.AspNetCore/SentryTracingMiddleware.cs
```diff
@@ -178,13 +178,15 @@ public async Task InvokeAsync(HttpContext context)
...
                 {
                     var method = context.Request.Method.ToUpperInvariant();
 
                     // If we've set a TransactionNameProvider, use that here
-                    var customTransactionName = context.TryGetCustomTransactionName();
-                    if (!string.IsNullOrEmpty(customTransactionName))
+                    if (customName.Value is { } customTransactionName)
                     {
                         transaction.Name = $"{method} {customTransactionName}";
                         tracer.NameSource = TransactionNameSource.Custom;
```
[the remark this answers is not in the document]
jamescrosswell: This is intentional - the whole purpose of this PR is to address the gap highlighted in the original issue:
> There is no SDK-level hook that runs unconditionally after routing resolves and allows the transaction name to be rewritten.

Anyone setting this option will have to check the existing transaction name (by reading the transaction from the scope) if they want to use that in the logic defined for returning a custom transaction name.

Flash0ver (approved):

src/Sentry.AspNetCore/SentryAspNetCoreOptions.cs
```diff
@@ -63,6 +63,19 @@ public class SentryAspNetCoreOptions : SentryLoggingOptions
...
+    /// when routing fails to resolve a template.
+    /// </summary>
+    /// <remarks>
+    /// When <c>false</c> (default), the provider is only invoked as a fallback for unresolved routes —
+    /// matching the legacy behaviour. When <c>true</c>, the provider is invoked for every request
+    /// after routing; a non-empty return value overrides the route-derived name. If the provider
+    /// returns <c>null</c> or <see cref="string.Empty"/>, the routed name is preserved (or the URL-path
+    /// fallback applies when no route was resolved).
+    /// </remarks>
+    public bool PreferTransactionNameProvider { get; set; }
```
jamescrosswell, line 77: Added follow up issue to complete documentation:
- https://github.com/getsentry/sentry-dotnet/issues/5349

PR #5162, opened 2026-04-23
fix(unity): update `SentryEvent` to have `IsCaptured` to allow dropping screenshots of filtered events

JoshuaMoelans:
As suggested in https://github.com/getsentry/sentry-unity/pull/2642 ,  we add a flag so we can read from it during the [CaptureScreenshotCoroutine](https://github.com/getsentry/sentry-unity/blob/4.2.0/src/Sentry.Unity/ScreenshotEventProcessor.cs#L36). We gate it behind `#if SENTRY_UNITY` to not expose it for the other .NET SDK consumers.

---

As described in the 'Future Ideas' section of the unity PR, at some point we might want to provide sentry-unity with a wrapper `ISentryClient` which can overwrite methods, to avoid needing changes in .NET for unity specific stuff like we need to do in this PR.


src/Sentry/SentryClient.cs
```diff
@@ -411,7 +411,14 @@ private SentryId DoSendEvent(SentryEvent @event, SentryHint? hint, Scope? scope)
 
         var attachments = hint.Attachments.ToList();
         var envelope = Envelope.FromEvent(processedEvent, _options.DiagnosticLogger, attachments, scope.SessionUpdate);
-        return CaptureEnvelope(envelope) ? processedEvent.EventId : SentryId.Empty;
+        if (CaptureEnvelope(envelope))
+        {
+#if SENTRY_UNITY
+            @event.WasCaptured = true; // See SentryEvent.unity.cs for more details.
```
Flash0ver: suggestion: fix capitalization

```suggestion
            @event.WasCaptured = true; // See SentryEvent.Unity.cs for more details.
```

Flash0ver: thought: alternative

As an alternative, we could use our "Fused-Data" infrastructure, based on [ConditionalWeakTable<,>](https://learn.microsoft.com/dotnet/api/system.runtime.compilerservices.conditionalweaktable-2).

e.g.
```CSharp
/* here */
@event.SetFused(new EnvelopeCaptureStatus(true));
```

```CSharp
/* in sentry-dotnet's ScreenshotEventProcessor */
if (a.GetFused<EnvelopeCaptureStatus>() is { IsCaptured: true })
{
    // Data found in CWT ... process and capture screenshot
}
else
{
    // no Data found in CWT ... stop processing the screenshot
}
```

```CSharp
/* new file in src/Sentry/Platforms/Unity/ */
#if SENTRY_UNITY

namespace Sentry.Internal;

internal sealed class EnvelopeCaptureStatus
{
    public EnvelopeCaptureStatus(bool isCaptured)
    {
        IsCaptured = isCaptured;
    }

    public bool IsCaptured { get; }
}

#endif
```

But to be fair ... now after manifesting this thought,
the "physical" property is just as good,
since it's `internal` and Unity-only anyway,
and also more navigable within an IDE.

I'd still like to hear feedback, though 😉

src/Sentry/Platforms/Unity/SentryEvent.Unity.cs
```diff
@@ -0,0 +1,16 @@
+#if SENTRY_UNITY
+
+namespace Sentry;
+
+public sealed partial class SentryEvent
+{
+
+    /// <summary>
```
Flash0ver, line 7: suggestion(style): no new line after opening brace

```suggestion
public sealed partial class SentryEvent
{
    /// <summary>
```

src/Sentry/Platforms/Unity/SentryEvent.Unity.cs
```diff
@@ -0,0 +1,16 @@
...
+    /// Indicates whether this event was actually captured and sent to Sentry.
+    /// Used by the Unity SDK's async screenshot capture to avoid sending orphaned attachments.
+    /// Defaults to false (safe: if unset, attachments are skipped rather than orphaned).
+    /// </summary>
+    internal bool WasCaptured { get; set; }
```
Flash0ver: suggestion: rename to `IsCaptured`?

```diff
-WasCaptured
+IsCaptured
```

Prefixing boolean names with either `Is` or `Has` is common.

In this case, when reading "WasCaptured", it could refer to - since it's past tense - the time the Event was created at.
Where - at least to me - `IsCaptured` represents the current state at query time, which also can change over the lifetime of the Event from `false` to `true`.

But since it's internal, and also Unity-only, this is definitely non-blocking.

src/Sentry/SentryClient.cs
JoshuaMoelans: Hmm yeah that is an option, but I'm a bit more drawn towards the (maybe less elegant-) simpler "physical" property solution for now. Since it doesn't introduce any potential future breakage of a public API, it'll be easy to replace it with something better at relatively low maintenance cost in the meantime.

src/Sentry/SentryClient.cs
```diff
@@ -411,7 +411,14 @@ private SentryId DoSendEvent(SentryEvent @event, SentryHint? hint, Scope? scope)
 
         var attachments = hint.Attachments.ToList();
         var envelope = Envelope.FromEvent(processedEvent, _options.DiagnosticLogger, attachments, scope.SessionUpdate);
-        return CaptureEnvelope(envelope) ? processedEvent.EventId : SentryId.Empty;
+        if (CaptureEnvelope(envelope))
+        {
+#if SENTRY_UNITY
+            @event.IsCaptured = true; // See SentryEvent.Unity.cs for more details.
+#endif
```
Flash0ver, line 418: @jamescrosswell is this "pattern" alright?

(adding a member via `src/Sentry/Platforms/Unity/SentryEvent.Unity.cs` partial type,
and using it here conditionally)

src/Sentry/SentryClient.cs
jamescrosswell: Since it's a class we own and can control, we may as well go with the internal property... comes with type checking and easier refactoring etc. 

Fused is good if we need to bolt a property onto a class from another code base where adding properties isn't an option.

src/Sentry/SentryClient.cs
jamescrosswell: It's nice and simple. The only alternative I can think of is that Unity could register a `UnityEventProcessor` that did this... I'm not sure that's easier to read/maintain though unless Unity starts to do quite complex special processing of these events.

jamescrosswell (approved):
LGTM

Flash0ver (approved):

src/Sentry/SentryClient.cs
Flash0ver: I'm not quite positive that this would work reliably, as there might be
- other Event-Processors dropping the Event
- a user's Before-Send dropping the Event
- Sampling of the Event
- the Backpressure-Monitor influencing the Sampling decision
- or the Background-Worker's queue being full

`sentry-unity` might provide it's own `ISentryClient` though, which is
- wrapping the default `sentry-unity` `SentryClient`
- wrapping a potential user client (`SentrySdk.BindClient(ISentryClient)`)
- attaching "fused" data to `SentryEvent`

So that it's a 100 % `sentry-unity`-owned solution.

See also getsentry/sentry-unity#2642: "Future Ideas"

PR #5172, opened 2026-04-29
feat(logs): add `log4net` integration

### LINKED ISSUE #4731: Add Sentry Logs support to Log4Net
### Problem Statement

The Log4Net adapter is the one common logging provider that hasn't yet gotten support for Sentry Logs.
I don't believe that Log4Net has support for message-templates though AFAIK, so that doesn't have to be part of the ACs.

### Solution Brainstorm

_No response_

Flash0ver:
closes #4731

### Changes
Add `log4net` integration to Sentry Structured Logging.

### Docs
- https://github.com/getsentry/sentry-docs/pull/18739

### See also
- #5216

## Implementation Notes

Sentry's log4net support is a **log4net appender** (`SentryAppender : AppenderSkeleton`). Every log call flows through `Append(LoggingEvent)`.

Previously each event became either a `SentryEvent` (error) or a breadcrumb. Now, when `options.EnableLogs` is on, `Append` *also* calls `CaptureStructuredLog` to emit a Sentry structured log.

### log4net-specific stuff

**1. Level mapping.** log4net doesn't have a clean enum — it has ~18 named `Level` objects (All, Finest, Verbose, Finer, Trace, Fine, Debug, Info, Notice, Warn, Error, Severe, Critical, Alert, Fatal, Emergency, …) *plus* user-defined custom levels with arbitrary integer values. [`ToSentryLogLevel`](src/Sentry.Log4Net/LevelMapping.cs:33) collapses all that into Sentry's 6 levels using `>=` threshold comparisons on the numeric value, so custom levels land in the nearest bucket and `Level.Off` drops the log (returns `null`).

**2. Message: rendered only.** log4net pre-renders the final string into `LoggingEvent.RenderedMessage`, which becomes the log `Message`. 

**3. No template, no parameters.** log4net can't give us the original template of parameters:

```csharp
const string? template = null;   // cannot get format-string from log4net.Util.SystemStringFormat
var parameters = ImmutableArray<...>.Empty; // cannot get arguments either
```

log4net's message object is typically a `SystemStringFormat` that keeps the composite format string (`"{0} {1}"`) and the args array as **private fields with no public accessor**. This is a limitation of log4net rather than Sentry's integration.

**4. Properties → `property.*` attributes, with log4net-specific filtering.** log4net merges event-level properties, `ThreadContext.Properties`, `GlobalContext.Properties`, and its own metadata into one bag via `GetProperties()`. The loop ([Structured.cs:20-29](src/Sentry.Log4Net/SentryAppender.Structured.cs:20)) maps each `DictionaryEntry` to a `property.<key>` attribute, skipping:
   - empty keys,
   - keys starting with `log4net:` (log4net's internal entries — HostName, Identity, etc.),
   - GUID-shaped keys (log4net's transient/internal correlation entries).

   This `property.<key>` convention matches **Serilog**, not MEL. (MEL doesn't have an ambient property bag — it emits parameters plus `category.name`/`event.id`/`event.name`.)


test/Sentry.Log4Net.Tests/SentryAppenderTests.Structured.cs
```diff
@@ -0,0 +1,203 @@
...
+        LoggingEvent loggingEvent = new(data);
+        sut.DoAppend(loggingEvent);
+
+        var log = capturer.Logs.Should().ContainSingle().Which;
+        log.Level.Should().Be(SentryLogLevel.Info);
+        log.Message.Should().Be("Test Message");
+
+        //TODO: assert Count/Length of Attributes
+        //requires: https://github.com/getsentry/sentry-dotnet/pull/4936
+        //should not contain "log4net:.." properties
```
Flash0ver: todo: add missing assertions

test/Sentry.Log4Net.Tests/SentryAppenderTests.Structured.cs
```diff
@@ -0,0 +1,236 @@
...
+        log.TryGetAttribute("sentry.environment", out object? environment).Should().BeTrue();
+        environment.Should().Be("test-environment");
+        log.TryGetAttribute("sentry.release", out object? release).Should().BeTrue();
+        release.Should().Be("test-release");
+        log.TryGetAttribute("sentry.origin", out object? origin).Should().BeTrue();
+        origin.Should().Be("auto.log.log4net");
+        log.TryGetAttribute("sentry.sdk.name", out object? sdkName).Should().BeTrue();
+        sdkName.Should().Be(SentryAppender.SdkName);
+        log.TryGetAttribute("sentry.sdk.version", out object? sdkVersion).Should().BeTrue();
+        sdkVersion.Should().Be(SentryAppender.NameAndVersion.Version);
```
Flash0ver: suggestion: reference project `Sentry.Testing` and simplify assertions with extensions added in #4936

jamescrosswell:
@Flash0ver I put this back in draft state until CI passes. Looks like there is some warden feedback worth checking as well.

src/Sentry.Log4Net/SentryAppender.Structured.cs
```diff
@@ -0,0 +1,33 @@
...
+    private static void CaptureStructuredLog(IHub hub, SentryOptions options, LoggingEvent loggingEvent)
+    {
+        var level = loggingEvent.ToSentryLogLevel();
+        if (level.HasValue)
+        {
+            DateTimeOffset timestamp = new(loggingEvent.TimeStampUtc);
+            const string? template = null; // cannot get format-string from `log4net.Util.SystemStringFormat` via `LoggingEvent.MessageObject`
+            var parameters = ImmutableArray<KeyValuePair<string, object>>.Empty; // cannot get arguments from `log4net.Util.SystemStringFormat` via `LoggingEvent.MessageObject`
+
+            var log = SentryLog.Create(hub, timestamp, level.Value, loggingEvent.RenderedMessage, template, parameters);
```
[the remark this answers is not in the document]
jamescrosswell: [441b655](https://github.com/getsentry/sentry-dotnet/pull/5172/commits/441b65511b5e58e2bf40f0b38b27dd54d54c97cf)

test/Sentry.Log4Net.Tests/SentryAppenderTests.Structured.cs
```diff
@@ -0,0 +1,236 @@
...
+            var span = Substitute.For<ISpan>();
+            span.TraceId.Returns(SentryId.Create());
+            span.SpanId.Returns(SpanId.Create());
+            _fixture.Hub.GetSpan().Returns(span);
+        }
+        else
+        {
+            _fixture.Hub.GetSpan().Returns((ISpan?)null);
+        }
+
```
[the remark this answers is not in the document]
jamescrosswell: Good catch — the four `ThreadContext.Properties` entries are thread-static and were leaking into `DoAppend_StructuredLogging_Properties` / `_DefaultProperties` on the same thread. They're now cleared in the test `Dispose()`, so subsequent tests start clean. Fixed in 441b6551.

test/Sentry.Log4Net.Tests/SentryAppenderTests.Structured.cs
jamescrosswell: Done in 0a7f3885 — switched these to the `ShouldContain` extensions from #4936, and added `Sentry.Log4Net.Tests` to `Sentry.Testing`'s `InternalsVisibleTo` so it can reach them. Kept `BeEquivalentTo` for the collection property, since the extension compares by value equality (`Should().Be()`) which fails for arrays. Thanks for the pointer!

samples/Sentry.Samples.Log4Net/Program.cs
```diff
@@ -10,9 +10,11 @@ internal class Program
 
     private static void Main()
     {
+#if NETFRAMEWORK
```
jamescrosswell: We only target net481 in this csproj, so I think we can remove the conditional compilation here.

src/Sentry.Log4Net/LevelMapping.cs
```diff
@@ -29,4 +29,20 @@ internal static class LevelMapping
...
+            var level when level == Level.Off => null,
+            var level when level >= Level.Fatal => SentryLogLevel.Fatal,
+            var level when level >= Level.Error => SentryLogLevel.Error,
+            var level when level >= Level.Warn => SentryLogLevel.Warning,
+            var level when level >= Level.Info => SentryLogLevel.Info,
+            var level when level >= Level.Debug => SentryLogLevel.Debug,
+            var level when level >= Level.Trace => SentryLogLevel.Trace,
+            var level when level >= Level.All => SentryLogLevel.Trace,
+            _ => null,
+        };
```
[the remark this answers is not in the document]
jamescrosswell: This looks like a false positive — the level ordering in the description is inverted. In log4net 2.0.15 the numeric `Level.Value`s are:

```
Emergency / Log4Net_Debug  120000
Fatal                      110000   <- the ">= Fatal" threshold
Alert                      100000
Critical                    90000
Severe                      80000
Error                       70000   <- the ">= Error" threshold
```

`Alert` (100000), `Critical` (90000) and `Severe` (80000) are all *below* `Fatal` (110000), so `level >= Level.Fatal` is false for them and they fall through to the `>= Level.Error` branch → `SentryLogLevel.Error`. That matches `ToSentryLevel`/`ToBreadcrumbLevel`, and the existing theory data asserts exactly this:

```csharp
{ Level.Alert,    SentryLogLevel.Error },
{ Level.Critical, SentryLogLevel.Error },
{ Level.Severe,   SentryLogLevel.Error },
```

Only `Emergency` and `Fatal` (both ≥ 110000) map to Fatal. So there is no Alert/Critical/Severe → Fatal misclassification here.

samples/Sentry.Samples.Log4Net/Program.cs
jamescrosswell: Done in 56abb519 — removed the `#if NETFRAMEWORK` guard and kept the `SetPrincipalPolicy` call unconditionally, since the sample only targets `net481`.

src/Sentry/SentryLog.Factory.cs
jamescrosswell, line 1: We don't need a separate file for one method. Let's move this back into SentryLog.cs and remove the special treatment for this file in the csproj file as well.

jamescrosswell: Done in b919a280 — moved `Create` back into `SentryLog.cs`, dropped the `partial` modifier (no longer needed), and removed the `SentryLog.Factory.cs` `DependentUpon` entry from `Sentry.csproj`.

src/Sentry.Log4Net/SentryAppender.Structured.cs
```diff
@@ -0,0 +1,40 @@
+namespace Sentry.Log4Net;
+
+public partial class SentryAppender
+{
+    private static void CaptureStructuredLog(IHub hub, SentryOptions options, LoggingEvent loggingEvent)
+    {
+        var level = loggingEvent.ToSentryLogLevel();
+        if (level.HasValue)
```
jamescrosswell: We can avoid nesting here by inverting and returning here, and also consolidate the conditional with the assignment.
```csharp
        if (loggingEvent.ToSentryLogLevel() is not {} level)
        {
                return;
        }
```

jamescrosswell: Done in 1f3468cc — inverted to a pattern-matched guard clause with an early return, which drops the nesting and folds the null check into the `level` assignment. `level.Value` becomes just `level` throughout.

src/Sentry.Log4Net/SentryAppender.Structured.cs
```diff
@@ -0,0 +1,41 @@
...
+        foreach (var property in loggingEvent.GetProperties())
+        {
+            if (property is DictionaryEntry { Key: string key, Value: { } value })
+            {
+                if (key.Length != 0 && !key.StartsWith("log4net:", StringComparison.OrdinalIgnoreCase) && !Guid.TryParse(key, out _))
+                {
+                    log.SetAttribute($"property.{key}", value);
+                }
+            }
+        }
```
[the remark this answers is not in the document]
jamescrosswell: Added the guard in c5d512d3. `GetProperties()` is non-null in current log4net (verified — it always builds a `PropertiesDictionary` via `CreateCompositeProperties()`/`Flatten()`), but the sibling `GetLoggingEventProperties` guards against null too, so this keeps the two consistent and avoids relying on log4net internals.

src/Sentry.Log4Net/SentryAppender.cs
```diff
@@ -84,6 +90,12 @@ protected override void Append(LoggingEvent loggingEvent)
             }
         }
 
+        var options = _hub.GetSentryOptions();
+        if (options is { EnableLogs: true })
+        {
+            CaptureStructuredLog(_hub, options, loggingEvent);
+        }
```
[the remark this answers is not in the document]
jamescrosswell: Valid — fixed in 2b315340. `CaptureStructuredLog` now receives `Environment` and `SendIdentity` and applies them (overriding the scope/options defaults), mirroring the `SentryEvent` path. Since both are per-appender opt-in settings, `SendIdentity` follows the event path and isn't additionally gated on `SendDefaultPii`. Added test coverage for both.

src/Sentry.Log4Net/SentryAppender.cs
```diff
@@ -88,17 +94,30 @@ protected override void Append(LoggingEvent loggingEvent)
...
         if (MinimumEventLevel is not null && loggingEvent.Level < MinimumEventLevel)
         {
-            var message = !string.IsNullOrWhiteSpace(loggingEvent.RenderedMessage) ? loggingEvent.RenderedMessage : string.Empty;
-            var category = loggingEvent.LoggerName;
-            var level = loggingEvent.ToBreadcrumbLevel();
-            IDictionary<string, string> data = GetLoggingEventProperties(loggingEvent)
-                .Where(kvp => kvp.Value != null)
-                .ToDictionary(kvp => kvp.Key, kvp => kvp.Value!.ToString() ?? "");
+            AddBreadcrumbFromLoggingEvent(loggingEvent);
+            return;
```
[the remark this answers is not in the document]
jamescrosswell: Good catch — real regression from the Append refactor. Fixed in 2b315340 by moving the structured-log capture ahead of the breadcrumb/event branching, so events below `MinimumEventLevel` produce structured logs again. The existing `DoAppend_StructuredLoggingWithoutException_LeavesBreadcrumb` test covers this and is green again.

src/Sentry.Log4Net/SentryAppender.cs
```diff
@@ -84,21 +90,40 @@ protected override void Append(LoggingEvent loggingEvent)
             }
         }
 
+        // Structured logs are captured for every event, independent of the breadcrumb/event branching below.
```
jamescrosswell:
```suggestion
```

jamescrosswell (approved):

PR #5176, opened 2026-05-01
feat(logs): add `NLog` integration

### LINKED ISSUE #5167: Add Sentry Logs support to NLog
### Description

Similar to `Serilog`, `SentryOptions.EnableLogs` - when `true` - should also emit [Sentry Structured Logs](<https://sentry.io/product/logs/>).

See also

* [NET-524](https://linear.app/getsentry/issue/NET-524/add-sentry-logs-support-to-log4net)

Flash0ver:
closes #5167

### Changes
Add `NLog` integration to Sentry Structured Logging.

### Docs
- https://github.com/getsentry/sentry-docs/pull/18738

### See also
- #5216

## Implementation notes

`SentryTarget` is an NLog *target*. Every log NLog routes to it lands in `InnerWrite` ([SentryTarget.cs:331](src/Sentry.NLog/SentryTarget.cs:331)), which now fans out to **two independent pipelines**:

1. **Classic path** (pre-existing): `logEvent` → Sentry **event** (if `Level >= MinimumEventLevel`) and/or **breadcrumb**.
2. **Structured-logs path** (this PR): `logEvent` → Sentry **structured log**, gated on `EnableLogs`.

The two are additive and independently gated — enabling logs doesn't change event/breadcrumb behavior.

### The dispatch gate

While the classic path continues to be gated by `MinimumEventLevel`, a new `EnableLogs` property has been added to determine whether structured logs are created from NLog logs:

```csharp
var sentryOptions = hub.GetSentryOptions();
if (sentryOptions?.EnableLogs is true)
    CaptureStructuredLog(hub, sentryOptions, logEvent);
```

### NLog-specifics

`CaptureStructuredLog` ([SentryTarget.Structured.cs](src/Sentry.NLog/SentryTarget.Structured.cs)) translates NLog's `LogEventInfo` into a `SentryLog`. 

The NLog-specific concerns:

| NLog concept | Sentry structured-log field | Handling |
|---|---|---|
| `logEvent.Level` (`NLog.LogLevel`) | `level` | `ToSentryLogLevel()` maps Trace/Debug/Info/Warn/Error/Fatal; **`Off` → null → drop** ([LogLevelExtensions.cs:33](src/Sentry.NLog/LogLevelExtensions.cs:33)). |
| `logEvent.FormattedMessage` (rendered) | `body` (`message`) | The message with arguments already substituted. |
| `logEvent.Message` (raw template) | `template` | e.g. `"User {UserId} logged in"`. |
| `logEvent.MessageTemplateParameters` | `parameters` (→ `sentry.message.parameter.*`) | e.g. `UserId `. Note that NLog also allows unnamed parammeters to be passed like `{}`. These have an empty name so `GetParameters` falls back to the positional index as the name, to avoid collisions. |
| `logEvent.Properties` | `property.*` attributes | NLog's structured/context properties. `GetStructuredLoggingParametersAndAttributes` filters out null/blank keys **and any key already used as a message parameter** (`!parameterNames.Contains(key)`) so a value isn't emitted twice. |
| `logEvent.LoggerName` | `category.name` attribute | The source logger identifier. |
| `logEvent.TimeStamp` (`DateTime`) | `timestamp` | Wrapped into a `DateTimeOffset`. |

Everything else — `Environment`, `Release`, SDK name/version, `server.address`, `user.*` — comes from `SetDefaultAttributes(options, scope, Sdk)`, shared with all integrations, plus `SetOrigin("auto.log.nlog")`.


test/Sentry.NLog.Tests/SentryTargetTests.Structured.cs
```diff
@@ -0,0 +1,175 @@
...
+        log.TryGetAttribute("sentry.environment", out object? environment).Should().BeTrue();
+        environment.Should().Be("test-environment");
+        log.TryGetAttribute("sentry.release", out object? release).Should().BeTrue();
+        release.Should().Be("test-release");
+        log.TryGetAttribute("sentry.origin", out object? origin).Should().BeTrue();
+        origin.Should().Be("auto.log.nlog");
+        log.TryGetAttribute("sentry.sdk.name", out object? sdkName).Should().BeTrue();
+        sdkName.Should().Be(Constants.SdkName);
+        log.TryGetAttribute("sentry.sdk.version", out object? sdkVersion).Should().BeTrue();
+        sdkVersion.Should().Be(SentryTarget.NameAndVersion.Version);
```
Flash0ver: suggestion: reference project `Sentry.Testing` and simplify assertions with extensions added in #4936

src/Sentry.NLog/SentryTarget.cs
```diff
@@ -331,6 +346,11 @@ private void InnerWrite(LogEventInfo logEvent)
         var shouldOnlyLogExceptions = exception == null && IgnoreEventsWithNoException;
         var shouldIncludeProperties = ContextProperties?.Count > 0 || ShouldIncludeProperties(logEvent);
 
+        if (Options.EnableLogs)
+        {
+            CaptureStructuredLog(hub, Options, logEvent);
+        }
```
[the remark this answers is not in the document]
jamescrosswell: This checks out - I think the same reasoning applies here:
https://github.com/getsentry/sentry-dotnet/blob/d3d2f718575c03c5a2fcbadb8d99d91bf8c754ea/src/Sentry.Serilog/SentrySink.cs#L168-L175

NB: More evidence to support https://github.com/getsentry/sentry-dotnet/issues/5245

src/Sentry.NLog/SentryTarget.Structured.cs
```diff
@@ -0,0 +1,72 @@
...
+    {
+        var parameters = logEvent.MessageTemplateParameters;
+
+        if (parameters.Count == 0)
+        {
+            parameterNames = new HashSet<string>();
+            return ImmutableArray<KeyValuePair<string, object>>.Empty;
+        }
+
+#if NETSTANDARD2_1_OR_GREATER || NET472_OR_GREATER || NETCOREAPP2_0_OR_GREATER
```
jamescrosswell: Could potentially simplify this. I think the only scenario where we don't take this codepath is if the TFM is `netstandard2.0` so we could potentially just check the TFM is not netstandard2.0?

src/Sentry/SentryLog.Factory.cs
jamescrosswell, line 1: Not a big fan of splitting a single method out into a separate file (and the extra overhead that adds to the csproj file). My preference would be to simply add it to SentryLog.cs... the method name itself lets you know it's a factory method.

jamescrosswell (commented):
Overall looks pretty good. No blocking comments from me - mostly stylistic.

The main one that needs to be dealt with is the one that the Sentry bot called out since that would be a bug for certain initialisation paths.

src/Sentry/SentryLog.Factory.cs
jamescrosswell: [ac37cef](https://github.com/getsentry/sentry-dotnet/pull/5176/commits/ac37cefa5c51bbff79848e9c6bab8b1b80d299a2)

test/Sentry.NLog.Tests/SentryTargetTests.Structured.cs
jamescrosswell: [7e572c9](https://github.com/getsentry/sentry-dotnet/pull/5176/commits/7e572c9690fe42b99273c504261db9fcc570dc34)

src/Sentry.NLog/SentryTarget.Structured.cs
```diff
@@ -0,0 +1,77 @@
...
+        parameterNames = new HashSet<string>();
+#endif
+
+        var @params = ImmutableArray.CreateBuilder<KeyValuePair<string, object>>(parameters.Count);
+
+        foreach (var parameter in parameters)
+        {
+            parameterNames.Add(parameter.Name);
+            @params.Add(new KeyValuePair<string, object>(parameter.Name, parameter.Value));
+        }
```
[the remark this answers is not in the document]
jamescrosswell: [5f3c30c](https://github.com/getsentry/sentry-dotnet/pull/5176/commits/5f3c30c20591450478f7ca3253c09a06ca8bfc8f)

src/Sentry.NLog/SentryTarget.Structured.cs
```diff
@@ -0,0 +1,87 @@
+namespace Sentry.NLog;
+
+public sealed partial class SentryTarget
+{
+    private static void CaptureStructuredLog(IHub hub, SentryOptions options, LogEventInfo logEvent)
+    {
+        var level = logEvent.Level.ToSentryLogLevel();
+        if (level.HasValue)
+        {
+            DateTimeOffset timestamp = new(logEvent.TimeStamp);
```
jamescrosswell: It's a nitpick but we could invert the logic here to reduce nesting... we could also combine the assignment and the branch into the same statement to make this more concise:
```csharp
        if (logEvent.Level.ToSentryLogLevel() is not { } level)
        {
            return;
        }
```

src/Sentry.NLog/SentryTarget.cs
```diff
@@ -331,6 +346,15 @@ private void InnerWrite(LogEventInfo logEvent)
...
         var shouldIncludeProperties = ContextProperties?.Count > 0 || ShouldIncludeProperties(logEvent);
 
+        // Read the options from the Hub, rather than the Target's NLog-Options, because 'EnableLogs' is declared in the base 'SentryOptions', rather than the derived 'SentryNLogOptions'.
+        // In cases where Sentry's NLog-Target is added without a DSN (i.e., without initializing the SDK) and the SDK is initialized differently (e.g., through ASP.NET Core),
+        // then the 'EnableLogs' option of this Target's NLog-Options is default, but the Hub's Sentry-Options have the actual user-defined value configured.
+        var sentryOptions = hub.GetSentryOptions();
+        if (sentryOptions?.EnableLogs is true)
+        {
+            CaptureStructuredLog(hub, sentryOptions, logEvent);
+        }
```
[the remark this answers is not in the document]
jamescrosswell: [9775861](https://github.com/getsentry/sentry-dotnet/pull/5176/commits/9775861e6e343cd5d3cdc0661a844e0d0b788a42)

jamescrosswell (approved):

PR #5177, opened 2026-05-01
feat: Add public API for checking unhandled exceptions

### LINKED ISSUE #2877: Make SentryEvent.HasTerminalException() public

jamescrosswell:
## Summary

Adds two new public extension methods to `SentryEvent`:
- `IsFromUnhandledException()`: Checks if the event is from any unhandled exception
- `IsFromTerminalException()`: Checks if the event is from a terminal (crash-causing) exception

## Motivation

Fixes #2877

Users need to filter events in callbacks like `BeforeSend` based on whether exceptions are unhandled/terminal. This is particularly useful in MAUI apps where you might want to filter out common exceptions (like network timeouts) but always capture them if they're terminal.

## Implementation

Following @bruno-garcia's feedback, this uses extension methods rather than making internal methods public. The methods check both `Exception.Data` flags and `SentryExceptions` mechanism properties.

Key differences:
- **IsFromUnhandledException**: Returns true for any unhandled exception (terminal or non-terminal)
- **IsFromTerminalException**: Returns true only for unhandled exceptions that are terminal (caused app crash)

Both are well-documented with XML docs and usage examples.

## Exception.Data vs  SentryException.Mechanism

These representations the same information at different pipeline stages.
- `Exception.Data[Mechanism.HandledKey]` is the input form. Unhandled-exception hooks set this in the raw System.Exception's Data dictionary.
- `SentryException.Mechanism` is what gets serialized and sent to Sentry.
- [MainExceptionProcessor:171](https://github.com/getsentry/sentry-dotnet/blob/ae09873b2181a6d7503065f195e01861ae3937d6/src/Sentry/Internal/MainExceptionProcessor.cs#L171-L176) converts the prior to the later

That's why the API checks both sources - depending on when someone calls it, the truth lives in a different place - and why the tests need to cover both variants.

They will never, in practice, be divergent/inconsistent so that's not a scenario we need to worry about.

### Changelog Entry

Add public API for checking unhandled exceptions (`SentryEventExtensions.IsFromUnhandledException`) and terminal exceptions (`SentryEventExtensions.IsFromTerminalException`)


src/Sentry/SentryEventExtensions.cs
```diff
@@ -0,0 +1,110 @@
...
+    {
+        // Check if the original exception was marked as unhandled
+        if (@event.Exception?.Data[Mechanism.HandledKey] is false)
+        {
+            return true;
+        }
+
+        // Check if any of the Sentry exceptions have an unhandled mechanism
+        return @event.SentryExceptions?.Any(e => e.Mechanism is { Handled: false }) ?? false;
+    }
```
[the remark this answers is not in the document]
jamescrosswell: I think we can remove the private methods entirely and have existing call sites call the extension methods instead.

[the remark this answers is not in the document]
jamescrosswell: Fixed in [8fc8368](https://github.com/getsentry/sentry-dotnet/pull/5177/commits/8fc8368ea4ef3d882b966e1dffd21d2e80253897)

src/Sentry/SentryEventExtensions.cs
```diff
@@ -0,0 +1,110 @@
...
+    public static bool IsFromUnhandledException(this SentryEvent @event)
+    {
+        // Check if the original exception was marked as unhandled
+        if (@event.Exception?.Data[Mechanism.HandledKey] is false)
+        {
+            return true;
+        }
+
+        // Check if any of the Sentry exceptions have an unhandled mechanism
+        return @event.SentryExceptions?.Any(e => e.Mechanism is { Handled: false }) ?? false;
```
Flash0ver: thought: move back to `SentryEvent.cs`

I prefer to keep the `private bool HasUnhandledException()` method in the `SentryEvent`, where it was, but make it `internal` instead, and then have this Extension Method just call the `internal` API.
This way, thinking of cohesion, the logic would remain alongside the other `HasException` and `HasUnhandledNonTerminalException` methods.

src/Sentry/SentryEventExtensions.cs
```diff
@@ -0,0 +1,110 @@
+using Sentry.Protocol;
+
+namespace Sentry;
+
+/// <summary>
+/// Extension methods for <see cref="SentryEvent"/>.
+/// </summary>
+public static class SentryEventExtensions
```
Flash0ver, line 6: question: documentation

I wonder how to document - in `docs.sentry.io` - additional Types and Members apart from `SentrySdk` and `SentryOptions` (and derived) 🤔

We don't really have a "hook" where to put e.g. methods and extension methods on `SentryEvent`.

Do we need something like the `.NET API catalog`?
- like https://learn.microsoft.com/dotnet/api/
- and https://apisof.net/

But - to be fair - this would be a bit of a "project of its own" ... Hackweek-potential 😉.

test/Sentry.Tests/SentryEventExtensionsTests.cs
```diff
@@ -0,0 +1,287 @@
...
+                }
+            }
+        };
+
+        // Act
+        var result = sentryEvent.IsFromTerminalException();
+
+        // Assert
+        Assert.True(result);
+    }
```
Flash0ver, line 232: test: `new Mechanism { Handled = false }`

I think there is a test case missing, where we set `Handled = false` as Mechanism, which then returns `false`. As the Mechanism-equivalent to the test case `IsFromTerminalException_HandledExceptionInExceptionProperty_ReturnsFalse`

src/Sentry/SentryEventExtensions.cs
```diff
@@ -0,0 +1,110 @@
...
+            return true;
+        }
+
+        // Check if any Sentry exceptions are unhandled and terminal
+        // (handled: false and terminal: not explicitly false)
+        return @event.SentryExceptions?.Any(e =>
+            e.Mechanism is { Handled: false } &&
+            e.Mechanism.Terminal != false
+        ) ?? false;
+    }
```
Flash0ver: question: should we move this to an `internal` method in `SentryEvent` ...

... which is then called from this public method?
So that we have all the `HasException`, `HasUnhandledNonTerminalException`, and so on in one location?

src/Sentry/SentryEventExtensions.cs
jamescrosswell: Hm, it could maybe go somewhere near the before send filter:
https://docs.sentry.io/platforms/dotnet/configuration/filtering/#using-before-send

That's a likely place where it would be used. 

Alternatively, we could put it in the Capturing Errors section:
https://docs.sentry.io/platforms/dotnet/usage/

Added a follow up issues to be done once this has been merged:
- https://github.com/getsentry/sentry-dotnet/issues/5348

src/Sentry/SentryEventExtensions.cs
jamescrosswell: Done in 3c19aec8. Moved the mechanism-inspection logic back into `SentryEvent` as `internal HasUnhandledException()` / `HasUnhandledTerminalException()`, alongside the existing `HasException`/`HasUnhandledNonTerminalException`. The public extension methods now just delegate to these, and `GetExceptionType()` calls the internal method directly instead of the extension.

src/Sentry/SentryEventExtensions.cs
jamescrosswell: Done in 3c19aec8. Moved the mechanism-inspection logic back into `SentryEvent` as `internal HasUnhandledException()` / `HasUnhandledTerminalException()`, alongside the existing `HasException`/`HasUnhandledNonTerminalException`. The public extension methods now just delegate to these, and `GetExceptionType()` calls the internal method directly instead of the extension.

test/Sentry.Tests/SentryEventExtensionsTests.cs
jamescrosswell: Added `IsFromTerminalException_HandledExceptionInSentryExceptions_ReturnsFalse` in 198a4c68.

One note: I used `new Mechanism { Handled = true }` rather than `Handled = false`. A handled mechanism is what returns `false` and mirrors `IsFromTerminalException_HandledExceptionInExceptionProperty_ReturnsFalse` (which sets `Handled = true` on the exception). `Handled = false` with `Terminal` unset returns `true` and is already covered by `IsFromTerminalException_UnhandledTerminalExceptionInSentryExceptions_ReturnsTrue`. Let me know if you meant something different.

Flash0ver: This is actually what I meant ... I made a typo in my comment. Thanks!

Flash0ver (approved):

PR #5182, opened 2026-05-04
feat: Add AddToTransactions to SentryAttachment

### LINKED ISSUE #750: Add addToTransaction to Attachment
The attachments should offer a flag `addToTransaction`, that specific if the attachment is added to transactions or not. The default should be `false`.

Related to https://github.com/getsentry/sentry-java/issues/1185

### LINKED ISSUE #5397: Collapse Envelope.FromTransaction overloads into a single method with optional parameters
### Background

Follow-up from [#5182 (review)](https://github.com/getsentry/sentry-dotnet/pull/5182#discussion_r3595300784).

`Envelope.FromTransaction` is currently split into two overloads:

```csharp
public static Envelope FromTransaction(SentryTransaction transaction) =>
    FromTransaction(transaction, null, null);

public static Envelope FromTransaction(
    SentryTransaction transaction,
    IDiagnosticLogger? logger,
    IReadOnlyCollection<SentryAttachment>? attachments)
```

The sibling factory methods `FromEvent` and `FromFeedback` instead expose a single method with optional parameters defaulting to `null`:

```csharp
public static Envelope FromFeedback(
    SentryEvent @event,
    IDiagnosticLogger? logger = null,
    IReadOnlyCollection<SentryAttachment>? attachments = null,
    SessionUpdate? sessionUpdate = null)
```

As [@bitsandfoxes noted](https://github.com/getsentry/sentry-dotnet/pull/5182#discussion_r3595300784), collapsing `FromTransaction` into a single method with optional parameters would bring it in line with the rest.

### Proposal

Replace the two `FromTransaction` overloads with a single method:

```csharp
public static Envelope FromTransaction(
    SentryTransaction transaction,
    IDiagnosticLogger? logger = null,
    IReadOnlyCollection<SentryAttachment>? attachments = null)
```

jamescrosswell:
## Summary

* Adds `AddToTransactions` boolean property to `SentryAttachment` (defaults to `false`)
* New constructor overload: `SentryAttachment(type, content, fileName, contentType, addToTransactions)`
* `Envelope.FromTransaction` now accepts optional `attachments` parameter and includes those with `AddToTransactions = true`
* `SentryClient.CaptureTransaction` filters attachments to only pass `AddToTransactions = true` ones to the envelope

Implements getsentry/sentry-dotnet#750. See also the [Java reference implementation](<https://github.com/getsentry/sentry-java/pull/1191>) and the [SDK spec](<https://develop.sentry.dev/sdk/telemetry/attachments/>).

> [!NOTE]
> We didn't add this for Maui Screenshots since those get added automatically/internally by the Sentry SDK and **_only_** for Events and Feedback (never for transactions - which makes sense - we don't want to capture a screenshot every time a transaction is finished). 

## Docs

* [https://github.com/getsentry/sentry-docs/pull/18741](<https://github.com/getsentry/sentry-docs/pull/18741>)

### Changelog Entry

Attachments can now be sent with transactions by setting `AddToTransactions` on `SentryAttachment` [#5182](<https://github.com/getsentry/sentry-dotnet/pull/5182>)


src/Sentry/Protocol/Envelopes/Envelope.cs
```diff
@@ -337,7 +337,16 @@ public static Envelope FromFeedback(
...
-    public static Envelope FromTransaction(SentryTransaction transaction)
+    public static Envelope FromTransaction(SentryTransaction transaction) =>
+        FromTransaction(transaction, null, null);
+
+    /// <summary>
+    /// Creates an envelope that contains a single transaction and optional attachments.
+    /// </summary>
+    public static Envelope FromTransaction(
+        SentryTransaction transaction,
+        IDiagnosticLogger? logger,
```
Flash0ver: note: parameter order

Since the `IDiagnosticLogger` is the "least functional" parameter,
and the `attachments` having a "bigger impact" on the behavior,
I suggest swapping these two parameters in this public API.

However, the existing `FromEvent` and `FromFeedback` public methods also have the `IDiagnosticLogger` as second parameter ... so we are consistent ✅

src/Sentry/Protocol/Envelopes/Envelope.cs
```diff
@@ -358,6 +367,20 @@ public static Envelope FromTransaction(SentryTransaction transaction)
...
 
+        if (attachments is not null)
+        {
+            foreach (var attachment in attachments)
+            {
+                if (attachment.IsNull())
+                {
+                    logger?.LogWarning("Encountered a null attachment.  Skipping.");
+                    continue;
+                }
```
Flash0ver, line 372: nitpick: coverage

Should we also have a test - or a test path - that covers and asserts the skipping behavior (and optionally the log message), when there is/are attachment/attachments in the collection that are `null`?

src/Sentry/Protocol/Envelopes/Envelope.cs
```diff
@@ -337,7 +337,16 @@ public static Envelope FromFeedback(
...
     /// Creates an envelope that contains a single transaction.
     /// </summary>
-    public static Envelope FromTransaction(SentryTransaction transaction)
+    public static Envelope FromTransaction(SentryTransaction transaction) =>
+        FromTransaction(transaction, null, null);
+
+    /// <summary>
+    /// Creates an envelope that contains a single transaction and optional attachments.
+    /// </summary>
+    public static Envelope FromTransaction(
```
Flash0ver, line 340: question: `public`

Just to make sure ... is this method intended to be `public`?

src/Sentry/SentryAttachment.cs
```diff
@@ -73,10 +79,24 @@ public SentryAttachment(
...
         string fileName,
         string? contentType)
+        : this(type, content, fileName, contentType, false)
+    {
+    }
+
+    /// <summary>
+    /// Initializes an instance of <see cref="SentryAttachment"/>.
+    /// </summary>
+    public SentryAttachment(
```
Flash0ver, line 89: question: derived `ViewHierarchyAttachment`

Should we offer this flag in a similarly overloaded _.ctor_ for derived `SentryAttachment` types as well?

test/Sentry.Tests/SentryClientTests.cs
```diff
@@ -1294,6 +1294,54 @@ public void CaptureTransaction_ScopeContainsAttachments_GetAppliedToHint()
...
+        {
+            IsSampled = true,
+            EndTimestamp = DateTimeOffset.Now
+        };
+        var attachment = new SentryAttachment(
+            AttachmentType.Default,
+            new StreamAttachmentContent(new MemoryStream(new byte[] { 1 })),
+            "include.txt",
+            null,
+            addToTransactions: true);
```
Flash0ver: suggestion: add overload to `AttachmentHelper.FakeAttachment`

To make this test a bit shorter, and consistent with the `false`-test below,
we could add an overload or optional parameter to `AttachmentHelper.FakeAttachment` so that we can pass the `bool addToTransactions` parameter there too.

src/Sentry/SentryAttachment.cs
```diff
@@ -73,10 +79,24 @@ public SentryAttachment(
...
+        AttachmentType type,
+        IAttachmentContent content,
+        string fileName,
+        string? contentType,
+        bool addToTransactions)
```
Flash0ver, line 94: question: add parameter to `AddAttachment` methods

Should we add this new parameter to the `AddAttachment` methods on `Scope` and `SentryHint` too?
Although:
- on `Scope`, we have the overload `AddAttachment(SentryAttachment attachment)`, via which the user can create an instance of `SentryAttachment` with the new _.ctor_-overload
- on `SentryHint`, we don't have an overload that takes a "constructed" `SentryAttachment`

So I think that `Scope` is fine.
But we might want to add an `AddAttachment(SentryAttachment attachment)` overload to `SentryHint`.

But I have the feeling this is currently not required ... we could always add these overloads later, should users ask for it, or we have a first-party use-case for it.

I was just wondering if these are required for a "complete user journey".

src/Sentry/SentryClient.cs
```diff
@@ -215,7 +215,8 @@ public void CaptureTransaction(SentryTransaction transaction, Scope? scope, Sent
             processedTransaction.Redact();
         }
 
-        CaptureEnvelope(Envelope.FromTransaction(processedTransaction));
+        var attachments = hint.Attachments.Where(a => a.AddToTransactions).ToList();
```
Flash0ver: note: `Scope.Attachments`

I was wondering why we only read the `Attachments` from the `SentryHint` but not from the `Scope`. But then I realized that we have logic where we apply the Attachments from the Scope onto the Hint.

And this is actually also covered by the test case, where we pass the Attachments via the Scope, and pass a `null`-Hint ✅

src/Sentry/Protocol/Envelopes/Envelope.cs
jamescrosswell: If we do this, let's do it consistently for all three methods. Breaking change so I've created a follow up issue for the next major:
- #5379

src/Sentry/SentryAttachment.cs
jamescrosswell: Yeah good call - also needed on `ScreenshotAttachment`.

See [a8edc5d](https://github.com/getsentry/sentry-dotnet/pull/5182/commits/a8edc5d48e9138161fe53761eab7627ea9bfc2e3)

src/Sentry/SentryAttachment.cs
jamescrosswell: `SentryHint` does already have the `WithAttachments` method:
https://github.com/getsentry/sentry-dotnet/blob/f8e84bb327d2a7088f02b8b02414530e7cabb594/src/Sentry/SentryHint.cs#L104-L109

So I think we're covered... [See here](https://github.com/getsentry/sentry-dotnet/pull/5182/changes#r3567675067) for the explanation as to the different APIs.

src/Sentry/SentryClient.cs
jamescrosswell: Yeah, the Hint is ephemeral - it's basically just used to form a convenient parameter to the BeforeSend callback.

I think that's why we have the fluent `SentryHint.WithAttachments` API vs `Scope.AddAttachment` (scope is long lived so you'd likely retain a reference to it).

src/Sentry/Protocol/Envelopes/Envelope.cs
jamescrosswell: We don't seem very consistent here actually. 

### Public

https://github.com/getsentry/sentry-dotnet/blob/416ec0b4020581b17078c2af624bd5d8661d7747/src/Sentry/Protocol/Envelopes/Envelope.cs#L340-L341
https://github.com/getsentry/sentry-dotnet/blob/416ec0b4020581b17078c2af624bd5d8661d7747/src/Sentry/Protocol/Envelopes/Envelope.cs#L434
https://github.com/getsentry/sentry-dotnet/blob/416ec0b4020581b17078c2af624bd5d8661d7747/src/Sentry/Protocol/Envelopes/Envelope.cs#L419

### Internal

https://github.com/getsentry/sentry-dotnet/blob/416ec0b4020581b17078c2af624bd5d8661d7747/src/Sentry/Protocol/Envelopes/Envelope.cs#L403
https://github.com/getsentry/sentry-dotnet/blob/416ec0b4020581b17078c2af624bd5d8661d7747/src/Sentry/Protocol/Envelopes/Envelope.cs#L446

If there is a rule, it might be that where we have a public method to capture something on the SDK (e.g. Event or Transaction) we seem to have made the `Envelope.From*` method public as well... 

I'm not sure SDK users are really supposed to form their own envelopes anyway, so maybe in a future major release, we could make all of these internal. Not sure if that would break Unity/GoDot though @bitsandfoxes ?

src/Sentry/Protocol/Envelopes/Envelope.cs
jamescrosswell: [1d0635b](https://github.com/getsentry/sentry-dotnet/pull/5182/commits/1d0635b108a900f08dc130c9aef6aad94fde0e20)

test/Sentry.Tests/SentryClientTests.cs
jamescrosswell: [748a7e5](https://github.com/getsentry/sentry-dotnet/pull/5182/commits/748a7e5350244ad6f078f9c10ba907d29d3d5efe)

src/Sentry/SentryClient.cs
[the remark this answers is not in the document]
jamescrosswell: [56aa34e](https://github.com/getsentry/sentry-dotnet/pull/5182/commits/56aa34ef07cadd543590a2866d138a30735fd4ec)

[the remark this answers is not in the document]
jamescrosswell: [56aa34e](https://github.com/getsentry/sentry-dotnet/pull/5182/commits/56aa34ef07cadd543590a2866d138a30735fd4ec)

src/Sentry/SentryClient.cs
```diff
@@ -223,7 +223,10 @@ public void CaptureTransaction(SentryTransaction transaction, Scope? scope, Sent
             processedTransaction.Redact();
         }
 
-        CaptureEnvelope(Envelope.FromTransaction(processedTransaction));
+        // Keep null entries so the null-attachment guard in Envelope.FromTransaction handles them
+        // (consistent with the event/feedback capture paths); dereferencing them here would throw.
+        var attachments = hint.Attachments.Where(a => a is null || a.AddToTransactions).ToList();
```
dingsdax, line 228: can hint be null?

dingsdax (commented):
I guess this should get some changelog/docs updates too, implementation looks good, one comment about hint being possibly null

src/Sentry/SentryClient.cs
jamescrosswell: No we ensure that here:
https://github.com/getsentry/sentry-dotnet/blob/5e9a7c44b99da2eff65eb824bca7b72350da2326/src/Sentry/SentryClient.cs#L202-L203

Earlier (in the Hub) they can be null but not by the time we capture the Envelope.

jamescrosswell:
> I guess this should get some changelog/docs updates too, implementation looks good, one comment about hint being possibly null

Added an explicit `### Changelog Entry`  to the PR description (get's used by craft when publishing) and also linked to some draft docs (to be completed/reviewed separately).

See  my reply re the question about Hint nullability.

src/Sentry/Protocol/Envelopes/Envelope.cs
bitsandfoxes: I think we're fine there. Unity and Godot don't create their own envelopes but route them through the SDK.

And looking at the sibling methods i.e. 
```
    public static Envelope FromFeedback(
        SentryEvent @event,
        IDiagnosticLogger? logger = null,
        IReadOnlyCollection<SentryAttachment>? attachments = null,
        SessionUpdate? sessionUpdate = null)
```
they are `public` with the optional parameters defaulting to `null`. 
It's binary breaking to do that now but if we were to collapse the `FromTransaction` into one as well it'd perfectly aligned with the rest.

src/Sentry.Maui/Internal/ScreenshotAttachment.cs
```diff
@@ -17,8 +17,9 @@ private ScreenshotAttachment(
         AttachmentType type,
         IAttachmentContent content,
         string fileName,
-        string? contentType)
-        : base(type, content, fileName, contentType)
+        string? contentType,
+        bool addToTransactions = false)
```
bitsandfoxes: If we default to `false` here, should we make that configurable in the `public ctor`? I'm not terribly in-tune with "would i want a screenshot on a transaction on MAUI".

bitsandfoxes (approved):
LGTM!

dingsdax (approved):

src/Sentry.Maui/Internal/ScreenshotAttachment.cs
jamescrosswell: Ah, interesting. This entire constructor was never available... we don't use it for anything in particular internally - I think it should be public in case people want to add screenshots that aren't JPEGs.

Edit: Ah actually I see `ScreenshotAttachment` itself is internal... since we're not using this extra overload, we may as well just get rid of it on `ScreenshotAttachment`.

jamescrosswell: [c70ef2a](https://github.com/getsentry/sentry-dotnet/pull/5182/commits/c70ef2a4f70e263f1bb223152c7648012a2b7f9c)

PR #4717, opened 2025-11-10
fix: memory leak when finishing unsampled Transactions/Spans

Flash0ver:
Fixes #4709
- #4709

Similar to #3432,
but for the newer types `UnsampledTransaction` and `UnsampledSpan`.

---
With the reproduction code from the related issue (see #4709):

**Before**
image

**After**
image

---
Also verified against `Sentry.Samples.AspNetCore.Basic`:
- Memory-Profile keeps stable
  - with and without `options.AutoRegisterTracing = false;`
  - with `options.TracesSampleRate = 1.0;` and `options.TracesSampleRate = 0.0;`
- The `SentryTracingMiddleware` is already finishing the Transactions, hence now also releasing the tracked Spans
  - https://github.com/getsentry/sentry-dotnet/blob/2a3bf15f2661cc27ed20892071d4d065482dee6f/src/Sentry.AspNetCore/SentryTracingMiddleware.cs#L144-L149

PR #5185, opened 2026-05-04
fix(otel): skip SentrySpanProcessor when performance monitoring is disabled

### LINKED ISSUE #5184: Sentry.OpenTelemetry: memory leak when AddSentry() is used with TracesSampleRate = 0
### Package

Sentry.OpenTelemetry

### .NET Flavor

.NET

### .NET Version

10.0

### OS

Linux

### OS Version

*No response*

### Development Environment

Rider 2025.x (Linux)

No

*No response*

### SDK Version

5.16.3

*No response*

### Workload Versions

n/a

```csharp
services.AddSentry().AddSentryOptions(o => {
    o.Dsn = "...";
    o.UseOpenTelemetry();
    o.TracesSampleRate = 0.0;
});

services.AddOpenTelemetry().WithTracing(b => b.AddSentry());
```

### Steps to Reproduce

1. Configure Sentry with `UseOpenTelemetry()` and `TracesSampleRate = 0.0`.
2. Register the Sentry span processor via `services.AddOpenTelemetry().WithTracing(b => b.AddSentry())`.
3. Run an [ASP.NET](<http://ASP.NET>) Core service under steady HTTP/gRPC load.
4. Observe managed heap growth over time.

### Expected Result

When `TracesSampleRate = 0`, no performance data is collected or sent. The Sentry span processor should have no effect on memory — ideally it should not be registered at all, since there is no value in creating `UnsampledTransaction` objects that will never be sampled or sent.

### Actual Result

With `TracesSampleRate = 0` and `AddSentry()` wired into the OTel `TracerProvider`, every `Activity.Stop()` causes the `SentrySpanProcessor` to create an `UnsampledTransaction` with child `UnsampledSpan` objects. These accumulate in `ThreadLocal<ConcurrentBag<ISpan>>` and are not released, producing a continuous managed memory leak under any meaningful request load.

Memory dump excerpt showing the leak:

```
Sentry.Internal.UnsampledTransaction         ~400,000 instances
Sentry.Internal.UnsampledSpan                ~400,000 instances
System.Threading.ThreadLocal<ConcurrentBag<ISpan>>+LinkedSlotVolatile[]  large allocations
```

**Relationship to** getsentry/sentry-dotnet#4709 **/ PR** [#4717](<https://github.com/getsentry/sentry-dotnet/issues/4717>)

PR [#4717](<https://github.com/getsentry/sentry-dotnet/issues/4717>) (released in 5.16.3) fixed `UnsampledTransaction.Finish()` not releasing child spans, which addressed the leak in the `SentryTracingMiddleware` path. However, when using the OTel `SentrySpanProcessor`, the leak persists because the processor still creates `UnsampledTransaction` objects for every `Activity` regardless of the sample rate. Users must currently guard against this manually:

```csharp
// Workaround: only wire up Sentry into OTel when tracing is actually enabled
if (tracesSampleRate > 0) {
    o.UseOpenTelemetry();
    o.TracesSampleRate = tracesSampleRate;
    services.AddOpenTelemetry().WithTracing(b => b.AddSentry());
}
```

**Suggested fix**

`AddSentry()` on the OTel `TracerProvider` (or `UseOpenTelemetry()`) should check whether tracing is actually enabled (`TracesSampleRate > 0`) and skip registering the `SentrySpanProcessor` when it is not. This makes the safe behavior the default and prevents the footgun entirely.

HrMathematiker:
## Summary

Fixes #5184

When `TracesSampleRate = 0` and no `TracesSampler` is configured, `AddSentry()` on the OTel `TracerProvider` previously still registered `SentrySpanProcessor`. For every `Activity.Stop()` the processor created an `UnsampledTransaction` with child `UnsampledSpan` objects stored in `ThreadLocal<ConcurrentBag<ISpan>>`. These accumulated and were never released, causing a sustained managed memory leak under load.

This is related to the fix in #4717 (which addressed `UnsampledTransaction.Finish()` not releasing child spans in the `SentryTracingMiddleware` path), but the OTel `SentrySpanProcessor` path still requires users to manually guard against registering the processor when tracing is disabled.

## Changes

- `ImplementationFactory` in `TracerProviderBuilderExtensions` now checks `IsPerformanceMonitoringEnabled` before creating a `SentrySpanProcessor`. When tracing is off it returns `DisabledSpanProcessor` instead, and logs an `Info` message explaining why.
- Existing tests updated: `GetOptions()` in the fixture now sets `TracesSampleRate = 1.0` (the existing tests exercise the tracing-enabled path, so this is the correct baseline).
- New test: `ImplementationFactory_WithTracingDisabled_ReturnsDisabledSpanProcessor` verifies the fix directly.

## Before / After

**Before** — users had to guard manually:
```csharp
if (tracesSampleRate > 0) {
    services.AddOpenTelemetry().WithTracing(b => b.AddSentry());
}
```

**After** — `AddSentry()` is safe to call unconditionally; it no-ops when tracing is disabled:
```csharp
services.AddOpenTelemetry().WithTracing(b => b.AddSentry());
```

### DIFF
--- a/test/Sentry.OpenTelemetry.Tests/TracerProviderBuilderExtensionsTests.cs
+++ b/test/Sentry.OpenTelemetry.Tests/TracerProviderBuilderExtensionsTests.cs
@@ -15,7 +15,15 @@ public class TracerProviderBuilderExtensionsTests
         public SentryOptions GetOptions(string dsn = "https://123@o456.ingest.sentry.io/789") => new()
         {
             Instrumenter = Instrumenter.OpenTelemetry,
-            Dsn = dsn
+            Dsn = dsn,
+            TracesSampleRate = 1.0
+        };
+
+        public SentryOptions GetOptionsWithTracingDisabled(string dsn = "https://123@o456.ingest.sentry.io/789") => new()
+        {
+            Instrumenter = Instrumenter.OpenTelemetry,
+            Dsn = dsn,
+            TracesSampleRate = 0.0
         };
 
         public IServiceProvider GetServiceProvider() => ServiceProvider;
@@ -90,5 +98,24 @@ public class TracerProviderBuilderExtensionsTests
 
         // Assert
         result.Should().BeOfType<DisabledSpanProcessor>(); // FluentAssertions
+    }
+
+    [Fact]
+    public void ImplementationFactory_WithTracingDisabled_ReturnsDisabledSpanProcessor()
+    {
+        // Arrange
+        var fixture = new Fixture();
+        SentryClientExtensions.SentryOptionsForTestingOnly = fixture.GetOptionsWithTracingDisabled();
+        fixture.Hub.IsEnabled.Returns(true);
+        var services = fixture.GetServiceProvider();
+
+        // Act
+        var result = TracerProviderBuilderExtensions.ImplementationFactory(services);
+
+        // Assert
+        // When TracesSampleRate = 0 and no TracesSampler is configured, registering SentrySpanProcessor
+        // would create UnsampledTransaction + UnsampledSpan objects per Activity that accumulate in
+        // ThreadLocal storage and are never released, causing a managed memory leak.
+        result.Should().BeOfType<DisabledSpanProcessor>();
     }
 }
--- a/src/Sentry.OpenTelemetry/TracerProviderBuilderExtensions.cs
+++ b/src/Sentry.OpenTelemetry/TracerProviderBuilderExtensions.cs
@@ -48,11 +48,22 @@ public static class TracerProviderBuilderExtensions
         var hub = services.GetService<IHub>() ?? SentrySdk.CurrentHub;
         if (hub.IsEnabled)
         {
+            var options = hub.GetSentryOptions();
+            if (options is { IsPerformanceMonitoringEnabled: false })
+            {
+                var logger = services.GetService<IDiagnosticLogger>();
+                logger?.LogInfo(
+                    "Sentry performance monitoring is disabled (TracesSampleRate is 0 and no TracesSampler " +
+                    "is configured). Skipping SentrySpanProcessor registration to avoid memory overhead " +
+                    "from unsampled transactions.");
+                return DisabledSpanProcessor.Instance;
+            }
+
             return new SentrySpanProcessor(hub, enrichers);
         }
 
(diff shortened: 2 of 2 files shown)


test/Sentry.OpenTelemetry.Tests/TracerProviderBuilderExtensionsTests.cs
```diff
@@ -15,7 +15,15 @@ public Fixture()
         public SentryOptions GetOptions(string dsn = "https://123@o456.ingest.sentry.io/789") => new()
         {
             Instrumenter = Instrumenter.OpenTelemetry,
-            Dsn = dsn
+            Dsn = dsn,
+            TracesSampleRate = 1.0
+        };
+
+        public SentryOptions GetOptionsWithTracingDisabled(string dsn = "https://123@o456.ingest.sentry.io/789") => new()
```
jamescrosswell: Probably no reason to add a helper method for this. It's only used in one test so the property could be set inline in the one test that uses this. That would avoid adverse side effects (now or in the future) on any other tests that use the existing `GetOptions` method (which then also doesn't need to be changed).

jamescrosswell:
@HrMathematiker `TracesSampleRate` and `TracesSampler` aren't directly relevant to OTEL tracing... they're options used to control the sampling decision for Sentry's trace instrumentation. So setting a tracesSampleRate of 0.0 for Sentry traces doesn't necessarily mean you don't want to capture any OTEL traces... those OTEL trace sampling decisions are controlled via the OpenTelemetry configuration settings. 

I think we recently (in the last 12 hours) discovered a potential memory leak when using OTEL with Sentry and when OTEL spans get filtered out... so this PR may resolve the memory issues you're seeing:
- https://github.com/getsentry/sentry-dotnet/pull/5186

Ultimately all of this code is eventually going to be replaced by the `Sentry.OpenTelemetry.Exporter`... we're retaining the SpanProcessor in the v6 code line since removing it would be a breaking change but once v7 comes out the only way to use OTEL will be via the OLTP exporter and it will no longer be possible to mix Sentry and OTEL trace instrumentation - you'd either use just Sentry tracing or just OTEL tracing (not both) - at which point there is no span processor:
- https://github.com/getsentry/sentry-dotnet/pull/4899

I'll close this PR in any case since those sentry options aren't related to OTEL traces.

PR #5191, opened 2026-05-05
ci: enable dependabot updates for .github/actions

### LINKED ISSUE #5188: github/actions not being auto-updated by Dependabot
question: why where these not auto-updated by Dependabot?

Because they are **not** in `.github/workflows/*.yml`?
Are none of our `.github/actions/*/*.yml` auto-updated?
Do we need to do a manual pass on these from time to time?

Or can we add (in a follow-up) an additional Dependabot rule for the `github-actions` ecosystem and this `directory` (or these directories)?

_Originally posted by @Flash0ver in https://github.com/getsentry/sentry-dotnet/pull/5180#discussion_r3183201059_

jpnurmi:
Solution found in:
- https://github.com/dependabot/dependabot-core/issues/6345

And tested here:
- https://github.com/jpnurmi/dba/blob/main/.github/dependabot.yml
- https://github.com/jpnurmi/dba/pulls?q=sort%3Aupdated-desc+is%3Apr+is%3Aopen

Close: #5188

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -19,6 +19,8 @@ updates:
   open-pull-requests-limit: 2
 
 - package-ecosystem: "github-actions"
-  directory: "/"
+  directories:
+    - "/"
+    - "/.github/actions/*"
   schedule:
     interval: weekly


.github/dependabot.yml
```diff
@@ -19,6 +19,8 @@ updates:
   open-pull-requests-limit: 2
 
 - package-ecosystem: "github-actions"
-  directory: "/"
+  directories:
```
Flash0ver, line 22: note: Today I Learned: https://docs.github.com/en/code-security/reference/supply-chain-security/dependabot-options-reference#directories-or-directory--

Flash0ver (approved):

PR #5206, opened 2026-05-07
ci: bump craft minVersion to 2.26.3 to enable auto versioning

### LINKED ISSUE #5205: Bump minVersion in .craft.yml to enable 'auto' versioning
## Problem

Running `craft prepare auto` fails with:

```
[error] Auto-versioning requires minVersion >= 2.14.0 in .craft.yml.
Please update your configuration or specify the version explicitly.
```

The repo's `.craft.yml` currently has `minVersion: 1.1.1`, which is below the `2.14.0` threshold that craft enforces before allowing `auto` versioning.

See the failed release run: https://github.com/getsentry/sentry-dotnet/actions/runs/25367472828/job/74381977290

## Background

The `auto` version strategy was introduced in craft 2.14.0. It analyzes git history using conventional commits to automatically calculate the correct semver bump (patch/minor/major). The repo already has `changelogPolicy: auto` which works hand-in-hand with this — but the `minVersion` guard is blocking it.

## Fix

Update `.craft.yml`:

```yaml
# Before
minVersion: 1.1.1

# After
minVersion: 2.26.3
```

Bumping to the current latest (2.26.3) rather than just the minimum (2.14.0) so we also pick up any improvements and fixes released since then.

jamescrosswell:
Fixes #5205

Bumps `minVersion` in `.craft.yml` from `1.1.1` to `2.26.3`. The `auto` versioning strategy (used via `craft prepare auto`) was introduced in craft 2.14.0 and requires `minVersion >= 2.14.0` to be declared in `.craft.yml` as a safety guard. Setting it to the current latest (2.26.3) rather than just the minimum.

#skip-changelog

### DIFF
--- a/.craft.yml
+++ b/.craft.yml
@@ -1,4 +1,4 @@
-minVersion: 1.1.1
+minVersion: 2.26.3
 changelogPolicy: auto
 targets:
   - name: nuget


Flash0ver (approved):
not sure if there was a breaking change that we can observe from our current `.craft.yml` ... we need to observe that during the next release

Issue #4882, opened 2026-01-29
feat: add Attributes APIs


Flash0ver:
### Description

Attributes are searchable **Name**-**Value**-Pairs with a **Type** and an _optional_ **Unit**
- where currently supported **Types** for the **Values** are
  - `string` and `string[]`
  - `boolean` and `boolean[]`
  - `integer` and `integer[]` (64-bit signed integral number)
  - `double` and `double[]` (64-bit floating-point number)
- where the _optional_ **Units** must be supported by Relay
  - may be removed or set to `none` otherwise

Precedence of Attributes on **Logs** and **Metrics** - higher priorities overwrite lower priorities:
1. Default Attributes
1. Scope Attributes
1. SDK-APIs
   - Attributes passed via `SentrySdk.Logger.Log{SentryLogLevel}` methods
   - Attributes passed via `SentrySdk.Experimental.Metrics.Emit{SentryMetricType}` methods
1. Option-APIs
   - Attributes set via `SentryOptions.SetBeforeSendLog(Func<SentryLog, SentryLog?>)`
   - Attributes set via `SentryOptions.SetBeforeSendMetric<T>(Func<SentryMetric<T>, SentryMetric<T>?>)`

Add `Attributes` APIs for
- Logs
- Metrics
- Scopes

See related docs
- Logs: https://develop.sentry.dev/sdk/telemetry/logs/
- Metrics: https://develop.sentry.dev/sdk/telemetry/metrics/
- Scopes: https://develop.sentry.dev/sdk/telemetry/scopes/
- Attributes: https://develop.sentry.dev/sdk/telemetry/attributes/

In the longer term, this (may) act as a replacement / unification of _Tags_ and _Extras_.

Flash0ver:
Naming hint: see https://github.com/getsentry/sentry-dotnet/pull/4834#discussion_r2752209195

Flash0ver:
While at it, we may want to remove the `Enricher`, and instead push the information onto the "Root-Scope" at init-time, which is then applied to Events and such. The `Scope` might have various internal (extension) methods ("base" ApplyTo, "ApplyToFeedback", ...).
Another layer of complexity we could simplify is the `IEventLike`-hierarchy of interfaces.
Thanks @bitsandfoxes for the input ... I'll come back to that once we "Scope-ify" our infrastructure.

jamescrosswell:
[...]

PR #5214, opened 2026-05-08
fix: sync default tags to native layer

### LINKED ISSUE #2668: `DefaultTags` from `SentryUnityOptions` are not propagated to native crash events
### Problem Statement

Tags configured via `options.DefaultTags` on `SentryUnityOptions` are correctly attached to managed (C#) error events, but they are missing from native crash events (Android NDK, iOS, etc.). Users who rely on `DefaultTags` to slice, filter, group, and alert on their data in Sentry lose that ability the moment a crash is captured by the native layer instead of the managed layer, which is exactly when that context is most needed.

Given a configuration like:

```csharp
private static void AddDefaultTags(SentryUnityOptions options)
{
    options.DefaultTags.Add(ApplicationVersionStringKey, "1.1.1");
    // ...additional default tags
}
```

* Managed C# events (handled exceptions, `SentrySdk.CaptureException`, log integrations, etc.) arrive in Sentry with the configured default tags.
* Native crash events (e.g. SIGSEGV / SIGABRT captured by `sentry-native` on Android or iOS) arrive without those tags.

### Why the Current Behavior Is Insufficient

`DefaultTags` is the primary configuration surface for users to attach build-level metadata (app version, build number, deployment ring, store channel, region, A/B variant, etc.) to every event. The implicit and documented contract is that default tags show up on every event.

Native crashes are typically the highest-priority events users want to triage. Missing tags forces them to either set the same tags a second time via `SentrySdk.ConfigureScope` (which duplicates configuration and is easy to forget), or lose the ability to filter, group, or alert on native crashes by those dimensions entirely.

The root cause appears to be a scope sync gap. `DefaultTags` is applied at the managed event-processing layer but is not pushed down to the native scope before a native crash is captured. Native crashes are written by `sentry-native` from a signal handler and can only see whatever scope was synced ahead of time. The existing scope sync path (`IScopeObserver` in `sentry-java`, equivalent on Cocoa) handles tags set via `SentrySdk.ConfigureScope`, but tags coming from `options.DefaultTags` do not appear to be funneled through that same path.

Related: [[getsentry/sentry-dotnet#3855 (comment)](https://github.com/getsentry/sentry-dotnet/issues/3855#issuecomment-3043498549)] reports the same root cause family on the .NET MAUI Android side.

### Proposed Solutions or Approaches

The user-facing requirement is that `DefaultTags` appear on every event the SDK sends, including native crashes. One direction worth considering is propagating `DefaultTags` through the existing managed-to-native scope sync at init time, so they ride the same path as tags set via `ConfigureScope`. Open to whatever approach the team thinks is right.

### Platforms Affected

Android, iOS, and any other platform where a native crash backend is active.

bitsandfoxes:
Came up through and resolves https://github.com/getsentry/sentry-unity/issues/2668

### The Problem

When a user sets `DefaultTags` on the options, these tags do not make it to native events from Android or iOS. This is not only an issue for Unity but affects C# Android and iOS app as well.
The comment seems to be out of date and prior to the addition of the extensive native support the SDK has now.
```
// NOTE: Tags in options.DefaultTags should not be passed down, because we already call SetTag on each
//       one when sending events, which is relayed through the scope observer.
```

### Context

When a user sets the `DefaultTags` on the options, in the regular processing pipeline, these get applied by the `Enricher` on each individual event.
https://github.com/getsentry/sentry-dotnet/blob/af4ebc3d09c0969366310d4e6584d523a7fb143b/src/Sentry/Internal/Enricher.cs#L104

I don't know what the grand design here was, and why tags are not getting applied to the scope instead. Maybe this comes from one of the other integrations, or behaviour outside of having a `GlobalScope`? If not, we could move it out of the `Enricher` in general. Happy to follow up on this if that's the case.

### Proposal

During initialization, and if scope sync is enabled, and if there is a `ScopeObserver`, we set those tags, making them available on the native layer. The regular flow inside the C# processing stays preserved. 

Sample events taken from the Unity SDK with `options.DefaultTags.Add("my", "tag");`:
- [managed](https://sentry-sdks.sentry.io/issues/6770060527/events/bbd3051c173a46dcb99654be08327b73/)
- [native](https://sentry-sdks.sentry.io/issues/7467622769/events/19e96c3e7006475ee955312b905d060c/)

### DIFF
--- a/src/Sentry/SentrySdk.cs
+++ b/src/Sentry/SentrySdk.cs
@@ -93,17 +93,14 @@ static partial class SentrySdk
 
         // Default tags are applied per-event by the Enricher to events going through the .NET pipeline,
         // but native crashes are captured and uploaded by the native SDK without going through that pipeline.
-        // Push them to the scope so the scope observer relays them to the native layer and they're attached
-        // to native crash reports.
-        if (options is { EnableScopeSync: true, ScopeObserver: not null } && options.DefaultTags.Count > 0)
+        // Forward them to the scope observer so the native layer attaches them to crash reports.
+        // Bypassing the .NET scope keeps scope.Tags identical between native and non-native apps.
+        if (options is { EnableScopeSync: true, ScopeObserver: { } observer } && options.DefaultTags.Count > 0)
         {
-            hub.ConfigureScope(static (scope, opts) =>
+            foreach (var tag in options.DefaultTags)
             {
-                foreach (var tag in opts.DefaultTags)
-                {
-                    scope.SetTag(tag.Key, tag.Value);
-                }
-            }, options);
+                observer.SetTag(tag.Key, tag.Value);
+            }
         }
 
         // Platform specific check for profiler misconfiguration.
--- a/src/Sentry/SentrySdk.cs
+++ b/src/Sentry/SentrySdk.cs
@@ -91,6 +91,21 @@ static partial class SentrySdk
         }
         options.PostInitCallbacks.Clear();
 
+        // Default tags are applied per-event by the Enricher to events going through the .NET pipeline,
+        // but native crashes are captured and uploaded by the native SDK without going through that pipeline.
+        // Push them to the scope so the scope observer relays them to the native layer and they're attached
+        // to native crash reports.
+        if (options is { EnableScopeSync: true, ScopeObserver: not null } && options.DefaultTags.Count > 0)
+        {
+            hub.ConfigureScope(static (scope, opts) =>
+            {
+                foreach (var tag in opts.DefaultTags)
+                {
+                    scope.SetTag(tag.Key, tag.Value);
+                }
+            }, options);
+        }
+
         // Platform specific check for profiler misconfiguration.
 #if __IOS__
         // No user-facing warning necessary - the integration is part of InitSentryCocoaSdk().
--- a/test/Sentry.Tests/SentrySdkTests.cs
+++ b/test/Sentry.Tests/SentrySdkTests.cs
@@ -1005,6 +1005,7 @@ public class SentrySdkTests : IDisposable
             InitNativeSdks = false,
         };
         options.DefaultTags["env"] = "production";
+        options.DefaultTags["region"] = "us-east-1";
 
         // Act
         SentrySdk.InitHub(options);
--- a/test/Sentry.Tests/SentrySdkTests.cs
+++ b/test/Sentry.Tests/SentrySdkTests.cs
@@ -968,6 +968,52 @@ public class SentrySdkTests : IDisposable
 #endif
 
     [Fact]
+    public void InitHub_DefaultTagsWithScopeSync_RelayedToScopeObserver()
+    {
+        // Arrange
+        var observer = Substitute.For<IScopeObserver>();
+        var options = new SentryOptions
+        {
+            Dsn = ValidDsn,
+            ScopeObserver = observer,
+            EnableScopeSync = true,
+            BackgroundWorker = Substitute.For<IBackgroundWorker>(),
+            InitNativeSdks = false,
+        };
+        options.DefaultTags["env"] = "production";
+        options.DefaultTags["region"] = "us-east-1";
+
+        // Act
+        SentrySdk.InitHub(options);
+
+        // Assert
+        observer.Received(1).SetTag("env", "production");
+        observer.Received(1).SetTag("region", "us-east-1");
+    }
+
+    [Fact]
+    public void InitHub_DefaultTagsWithoutScopeSync_NotRelayedToScopeObserver()
+    {
+        // Arrange
+        var observer = Substitute.For<IScopeObserver>();
+        var options = new SentryOptions
+        {
+            Dsn = ValidDsn,
+            ScopeObserver = observer,
+            EnableScopeSync = false,
+            BackgroundWorker = Substitute.For<IBackgroundWorker>(),
+            InitNativeSdks = false,
+        };
+        options.DefaultTags["env"] = "production";
+
+        // Act
+        SentrySdk.InitHub(options);
+
+        // Assert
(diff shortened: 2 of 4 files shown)


src/Sentry/SentrySdk.cs
```diff
@@ -91,6 +91,18 @@ internal static IHub InitHub(SentryOptions options)
         }
         options.PostInitCallbacks.Clear();
 
+        // Default tags are applied per-event by the Enricher to events going through the .NET pipeline,
+        // but native crashes are captured and uploaded by the native SDK without going through that pipeline.
+        // Forward them to the scope observer so the native layer attaches them to crash reports.
+        // Bypassing the .NET scope keeps scope.Tags identical between native and non-native apps.
+        if (options is { EnableScopeSync: true, ScopeObserver: { } observer } && options.DefaultTags.Count > 0)
```
jamescrosswell, line 98: It looks like this would work in Global mode (where there is only one scope). 

I'm wondering what would happen for an Asp.Net Core application that's compiled AOT. I believe we enable Sentry Native for AOT compiled applications - but I imagine Sentry Native doesn't have any notion of an 'AsyncLocal' scope stack so we effectively end up with Global scopes (and no scope observer in any case) for native events in Asp.Net Core apps compiled AOT. So that would (probably) work as expected as well (i.e. anything you put in the Default tags would end up on all native events in Asp.Net Core AOT apps).

It might be worth testing that scenario just to check (or getting an LLM to build a little test app to check that - can't really add it to our test suite as we have no easy way to test AOT yet - so it would just be a sanity check).

jamescrosswell, line 98: Can `DefaultTags.Count > 0` be merged into the pattern match to simplify this?

jamescrosswell (commented):
Generally LGTM... two comments, would be good to check the ASP.NET Core behaviour before merging, just to be sure we're not breaking anything there (seems unlikely but that could be famous last words).

Honestly I'm not sure why Default Tags. Changing that would be breaking so have added a separate issue to be considered in the next major release:
- https://github.com/getsentry/sentry-dotnet/issues/5225

src/Sentry/SentrySdk.cs
```diff
@@ -91,6 +91,18 @@ internal static IHub InitHub(SentryOptions options)
...
+        {
+            foreach (var tag in options.DefaultTags)
+            {
+                observer.SetTag(tag.Key, tag.Value);
```
Flash0ver, line 102: question: re
> I don't know what the grand design here was, and why tags are not getting applied to the scope instead. Maybe this comes from one of the other integrations, or behaviour outside of having a `GlobalScope`? If not, we could move it out of the `Enricher` in general. Happy to follow up on this if that's the case.

Should we follow-up on that refactoring now,
or keep it in mind when doing "Global Attributes".
I can imagine, though, that a good opportunity to do the refactoring is when we're looking into removing the `Enricher` altogether.

See also #4882.

test/Sentry.Tests/SentrySdkTests.cs
```diff
@@ -967,6 +967,52 @@ public void InitHub_GlobalModeOff_NoWarningOrErrorLogged()
...
+        var observer = Substitute.For<IScopeObserver>();
+        var options = new SentryOptions
+        {
+            Dsn = ValidDsn,
+            ScopeObserver = observer,
+            EnableScopeSync = false,
+            BackgroundWorker = Substitute.For<IBackgroundWorker>(),
+            InitNativeSdks = false,
+        };
+        options.DefaultTags["env"] = "production";
```
Flash0ver, line 1007: suggestion: keep tests in sync

This is totally a nit-pick, though ... my care factor is quite minimal 😉

```suggestion
        options.DefaultTags["env"] = "production";
        options.DefaultTags["region"] = "us-east-1";
```

src/Sentry/SentrySdk.cs
Flash0ver: follow-up question:

Do we now apply "Default Tags" to Events _twice_, in scenarios where we have a Scope-Observer/Scope-Sync (Unity, Android, iOS, Native)?

So should we actually move `ApplyDefaultTags` out of the `Enricher` in an immediate follow-up?

I guess it's not really a problem, double-setting/overwriting Tags in the underlying `Dictionary<,>` via the Scope-Observer and the Enricher ... but it's not all too great either.

jamescrosswell: > Do we now apply "Default Tags" to Events _twice_, in scenarios where we have a Scope-Observer/Scope-Sync (Unity, Android, iOS, Native)?

I think they only get applied by the .NET SDK for managed events (by the enricher)... and they would only get applied by the native SDKs for native events right?

jamescrosswell: >> I don't know what the grand design here was, and why tags are not getting applied to the scope instead.

It may be because we don't currently provide any simple mechanism to initialise the 'Root' scope (when not in Global mode). 

> Should we follow-up on that refactoring now,

In terms of setting DefaultTags via some other mechanism than the enricher, I can't see any urgent need/benefit to doing that. 

If we could replace DefaultTags with an easy way to initialise the RootScope, that would probably be more powerful/flexible since it would allow SDK users to set default all kinds of things (not just tags). Removing DefaultTags sounds like a breaking change... so I don't think we want to do that immediately. Maybe something that we could do for the next major (if .net 11 isn't too painful to upgrade to).

Flash0ver: Indeed something for the next major ... for which we also plan Span-first/SpanV2 that has Attributes which inherit Attributes from the Scope.

jamescrosswell: - https://github.com/getsentry/sentry-dotnet/issues/5225

jamescrosswell (approved):

PR #5216, opened 2026-05-11
fix: populate Default attributes on structured logs from current scope

### LINKED ISSUE #5209: Scoped attributes are not being sent w/ logs
### Package

Sentry

### .NET Flavor

.NET

### .NET Version

10.0.203

### OS

Any (not platform specific)

### OS Version

26.4.1 (25E253)

### Development Environment

Visual Studio Code (macOS)

No

_No response_

### SDK Version

6.5.0

_No response_

### Workload Versions

Workload version: 10.0.200

| Installed Workload Id | Manifest Version    | Installation Source |
|-----------------------|---------------------|---------------------|
| ios                   | 26.2.10217/10.0.100 | SDK 10.0.200        |
| maui                  | 10.0.20/10.0.100    | SDK 10.0.200        |

```csharp
.UseSentry(options =>
{
    options.Dsn = "...";
    options.EnableLogs = true;
#if DEBUG
    options.Environment = "Development";
    options.Debug = true;
#else
    options.Environment = "Production";
    options.Debug = false;
#endif
    options.AutoSessionTracking = true;
    options.IsGlobalModeEnabled = true;
    options.AttachScreenshot = true;
    options.IncludeTextInBreadcrumbs = true;
    options.IncludeTitleInBreadcrumbs = true;
    options.IncludeBackgroundingStateInBreadcrumbs = true;
    options.SetBeforeBreadcrumb(FilterNoisyBreadcrumbs);
})
```

### Steps to Reproduce

1. Configure scope:
```
{
    scope.User = new SentryUser
    {
        Id = id,
        Username = username,
        Other = other,
    };
});
```

2. Create log:
```
SentrySdk.Logger.LogInfo(
    log =>
    {
        log.SetAttribute("...", ...);
        log.SetAttribute("...", ...);
        log.SetAttribute("...", ...);
    },
    "Recording started"
);
``` 

### Expected Result

Log contains user attributes automatically as mentioned [here in the MAUI Sentry Set Up Logs guide](https://docs.sentry.io/platforms/dotnet/guides/maui/logs/#user-attributes).

### Actual Result

No user attributes appear in the log

jamescrosswell:
Fixes #5209

## Problem

When a user is set on the current scope via `SentrySdk.ConfigureScope`, the user attributes (`user.id`, `user.name`, `user.email`) were not being included on structured logs, despite being [documented as automatic](https://docs.sentry.io/platforms/dotnet/logs/#user-attributes).

The root cause was in `DefaultSentryStructuredLogger.CaptureLog` — the scope was fetched but only used to retrieve the SDK version. The scope's user was never consulted.

Although the initial report only called out the missing user attributes, we should ensure all of the expected [Default Attributes](https://docs.sentry.io/platforms/dotnet/logs/#default-attributes) are set correctly:
- Core Attributes
- User Attributes
- Server Attributes
- Integration Attributes

Then Message Template Attributes are [already handled correctly](https://github.com/getsentry/sentry-dotnet/blob/8ba1b3e4b92763973a65a2ad0dad892c9516cfff/src/Sentry/SentryLog.cs#L251-L256). 

## Audit

All three log emission paths set the default attributes per the [docs](https://docs.sentry.io/platforms/dotnet/logs/#default-attributes):

| Attribute | `DefaultSentryStructuredLogger` (direct API) | `Sentry.Extensions.Logging` | `Sentry.Serilog` |
|---|---|---|---|
| `sentry.environment` | ✅ `SetDefaultAttributes` | ✅ | ✅ |
| `sentry.release` | ✅ `SetDefaultAttributes` (when non-null) | ✅ | ✅ |
| `sentry.sdk.name` / `sentry.sdk.version` | ✅ from `scope.Sdk` | ✅ from `_sdk` override | ✅ from `Sdk` override |
| `server.address` | ✅ `SetDefaultAttributes` | ✅ | ✅ |
| `user.id` / `user.name` / `user.email` | ✅ `SetDefaultAttributes` | ✅ | ✅ |
| `sentry.origin` | ❌ not set (intentional) | ✅ `auto.log.extensions_logging` | ✅ `auto.log.serilog` |
| `sentry.message.template` / `sentry.message.parameter.X` | ✅ written in `SentryLog.WriteTo` | ✅ | ✅ |

- **`sentry.environment`** — set unconditionally in `SentryLog.SetDefaultAttributes` via `options.SettingLocator.GetEnvironment()`. That call always returns a value (defaults to `"production"` or `"debug"` when nothing is configured), so this attribute is always present.
- **`sentry.release`** — set in `SentryLog.SetDefaultAttributes` when `options.SettingLocator.GetRelease()` returns non-null (falls back to the `SENTRY_RELEASE` env var, then `ApplicationVersionLocator`).
- **`sentry.origin`** — *not* set by `SetDefaultAttributes`. Each integration calls `log.SetOrigin(...)` itself after `SetDefaultAttributes`:
  - `Sentry.Extensions.Logging/SentryStructuredLogger.cs` → `"auto.log.extensions_logging"`
  - `Sentry.Serilog/SentrySink.Structured.cs` → `"auto.log.serilog"`

  Direct `SentrySdk.Logger.LogInfo(...)` calls (via `DefaultSentryStructuredLogger`) do not set `sentry.origin`, which matches the docs: *"added when generated by an SDK integration"*.

> [!NOTE]
> `Sentry.NLog` and `Sentry.Log4Net` don't have structured-log paths yet.

### DIFF
--- a/src/Sentry/Internal/DefaultSentryStructuredLogger.cs
+++ b/src/Sentry/Internal/DefaultSentryStructuredLogger.cs
@@ -75,7 +75,8 @@ internal sealed class DefaultSentryStructuredLogger : SentryStructuredLogger, ID
             return;
         }
 
-        log.SetDefaultAttributes(_options, _hub.GetScope());
+        var scope = _hub.GetScope();
+        log.SetDefaultAttributes(_options, scope, scope?.Sdk);
 
         CaptureLog(log);
     }
--- a/src/Sentry/Internal/DefaultSentryStructuredLogger.cs
+++ b/src/Sentry/Internal/DefaultSentryStructuredLogger.cs
@@ -75,8 +75,7 @@ internal sealed class DefaultSentryStructuredLogger : SentryStructuredLogger, ID
             return;
         }
 
-        var scope = _hub.GetScope();
-        log.SetDefaultAttributes(_options, scope?.Sdk ?? SdkVersion.Instance, scope);
+        log.SetDefaultAttributes(_options, _hub.GetScope());
 
         CaptureLog(log);
     }
--- a/src/Sentry/Internal/DefaultSentryStructuredLogger.cs
+++ b/src/Sentry/Internal/DefaultSentryStructuredLogger.cs
@@ -76,7 +76,7 @@ internal sealed class DefaultSentryStructuredLogger : SentryStructuredLogger, ID
         }
 
         var scope = _hub.GetScope();
-        log.SetDefaultAttributes(_options, scope, scope?.Sdk);
+        log.SetDefaultAttributes(_options, scope);
 
         CaptureLog(log);
     }
--- a/src/Sentry/Internal/DefaultSentryStructuredLogger.cs
+++ b/src/Sentry/Internal/DefaultSentryStructuredLogger.cs
@@ -76,7 +76,7 @@ internal sealed class DefaultSentryStructuredLogger : SentryStructuredLogger, ID
         }
 
         var scope = _hub.GetScope();
-        log.SetDefaultAttributes(_options, scope?.Sdk ?? SdkVersion.Instance);
+        log.SetDefaultAttributes(_options, scope?.Sdk ?? SdkVersion.Instance, scope);
 
         CaptureLog(log);
     }
--- a/src/Sentry/SentryLog.cs
+++ b/src/Sentry/SentryLog.cs
@@ -172,7 +172,7 @@ public sealed class SentryLog
         _attributes[key] = new SentryAttribute(value, "integer");
     }
 
-    internal void SetDefaultAttributes(SentryOptions options, SdkVersion sdk)
+    internal void SetDefaultAttributes(SentryOptions options, SdkVersion sdk, Scope? scope = null)
     {
         var environment = options.SettingLocator.GetEnvironment();
         SetAttribute("sentry.environment", environment);
@@ -190,6 +190,22 @@ public sealed class SentryLog
         if (sdk.Version is { } version)
         {
             SetAttribute("sentry.sdk.version", version);
+        }
+
+        if (scope?.User is { } user)
+        {
+            if (user.Id is { } userId)
+            {
+                SetAttribute("user.id", userId);
+            }
+            if (user.Username is { } username)
+            {
+                SetAttribute("user.name", username);
+            }
+            if (user.Email is { } email)
+            {
+                SetAttribute("user.email", email);
+            }
         }
     }
 
--- a/src/Sentry/SentryLog.cs
+++ b/src/Sentry/SentryLog.cs
@@ -184,12 +184,12 @@ public sealed class SentryLog
             SetAttribute("sentry.release", release);
         }
 
-        sdk ??= scope?.Sdk;
-        if (sdk?.Name is { } name)
+        sdk ??= scope?.Sdk ?? SdkVersion.Instance;
+        if (sdk.Name is { } name)
         {
             SetAttribute("sentry.sdk.name", name);
         }
-        if (sdk?.Version is { } version)
+        if (sdk.Version is { } version)
         {
             SetAttribute("sentry.sdk.version", version);
         }
--- a/src/Sentry/SentryLog.cs
+++ b/src/Sentry/SentryLog.cs
@@ -172,7 +172,7 @@ public sealed class SentryLog
         _attributes[key] = new SentryAttribute(value, "integer");
     }
 
-    internal void SetDefaultAttributes(SentryOptions options, Scope? scope)
+    internal void SetDefaultAttributes(SentryOptions options, Scope? scope, SdkVersion? sdk)
     {
         // Core Attributes
         var environment = options.SettingLocator.GetEnvironment();
@@ -184,12 +184,11 @@ public sealed class SentryLog
             SetAttribute("sentry.release", release);
         }
 
-        var sdk = scope?.Sdk ?? SdkVersion.Instance;
-        if (sdk.Name is { } name)
+        if (sdk?.Name is { } name)
         {
             SetAttribute("sentry.sdk.name", name);
         }
-        if (sdk.Version is { } version)
+        if (sdk?.Version is { } version)
         {
             SetAttribute("sentry.sdk.version", version);
         }
--- a/src/Sentry/SentryLog.cs
+++ b/src/Sentry/SentryLog.cs
@@ -172,8 +172,9 @@ public sealed class SentryLog
         _attributes[key] = new SentryAttribute(value, "integer");
     }
 
-    internal void SetDefaultAttributes(SentryOptions options, SdkVersion sdk, Scope? scope = null)
+    internal void SetDefaultAttributes(SentryOptions options, Scope? scope)
     {
+        // Core Attributes
         var environment = options.SettingLocator.GetEnvironment();
         SetAttribute("sentry.environment", environment);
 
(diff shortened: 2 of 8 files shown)


src/Sentry/SentryLog.cs
```diff
@@ -172,7 +172,7 @@ internal void SetAttribute(string key, int value)
         _attributes[key] = new SentryAttribute(value, "integer");
     }
 
-    internal void SetDefaultAttributes(SentryOptions options, SdkVersion sdk)
+    internal void SetDefaultAttributes(SentryOptions options, SdkVersion sdk, Scope? scope = null)
```
jamescrosswell: The sdk information is on the scope already, so passing both sdk and scope parameters is redundant... we can infer the sdk information from the scope.

src/Sentry/SentryLog.cs
```diff
@@ -191,6 +191,22 @@ internal void SetDefaultAttributes(SentryOptions options, SdkVersion sdk)
         {
             SetAttribute("sentry.sdk.version", version);
         }
+
+        if (scope?.User is { } user)
```
jamescrosswell, line 153: Although the issue that was initially reported only highlighted the missing user attributes, we should ensure all of the expected [Default Attributes](https://docs.sentry.io/platforms/dotnet/logs/#default-attributes) are set correctly:
- Core Attributes
- User Attributes
- Server Attributes
- Integration Attributes

Then Message Template Attributes are [already handled correctly](https://github.com/getsentry/sentry-dotnet/blob/8ba1b3e4b92763973a65a2ad0dad892c9516cfff/src/Sentry/SentryLog.cs#L251-L256).

src/Sentry/SentryLog.cs
jamescrosswell: Done — `SetDefaultAttributes` now takes `(SentryOptions, Scope?)` and derives the SDK from `scope?.Sdk ?? SdkVersion.Instance`. Updated Serilog and Extensions.Logging integrations to mirror their SDK identity onto the scope before logging (so the right SDK name still flows through).

src/Sentry/SentryLog.cs
jamescrosswell: Added server attributes — `server.address` is now populated from `options.ServerName` when set, falling back to `Environment.MachineName` when `SendDefaultPii` is enabled (mirroring `MainSentryEventProcessor`). Origin and message-template attributes are already handled elsewhere, and core/user/server are now all covered.

src/Sentry/Internal/DefaultSentryStructuredLogger.cs
```diff
@@ -76,7 +76,7 @@ private protected override void CaptureLog(SentryLogLevel level, string template
         }
 
         var scope = _hub.GetScope();
-        log.SetDefaultAttributes(_options, scope?.Sdk ?? SdkVersion.Instance);
+        log.SetDefaultAttributes(_options, scope);
```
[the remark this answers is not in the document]
jamescrosswell: Fixed in [1a5d694](https://github.com/getsentry/sentry-dotnet/pull/5216/commits/1a5d694e854206160cb81fd8b19e272ea9b52467)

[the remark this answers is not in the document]
Flash0ver: This is not a new problem, but existing behavior, that we now add more default attributes that potentially overwrite user-defined attributes. Previously, we would overwrite
- `sentry.environment`
- `sentry.release`
- `sentry.sdk.name`
- `sentry.sdk.version`

We could open a follow-up to change this behavior.

This is technically a change in behavior, so we might want to do this in the next major (`v7.0.0`).
But we probably do want to do this, because with Metrics we did get it right where user-attributes might overwrite default-attributes.

Flash0ver (approved):
I merged from main and resolved merge conflicts.
@jamescrosswell please review if I applied the changes correctly:
- `src/Sentry/SentryLog.cs`: all into existing `Attributes.SetDefaultAttributes(options, sdk);`
- tests: use `ShouldContain<T>` and `ShouldNotContain<T>` extending `SentryAttributes`

Both of these changes come from #4936.

Also, there are two Sentry-Bot comments ... I think one is fine, and the other one might be a follow-up.

src/Sentry/SentryLog.cs
```diff
@@ -133,9 +133,40 @@ public bool TryGetAttribute(string key, [NotNullWhen(true)] out object? value) =
...
+            }
+            if (user.Username is { } username)
+            {
+                SetAttribute("user.name", username);
+            }
+            if (user.Email is { } email)
+            {
+                SetAttribute("user.email", email);
+            }
+        }
```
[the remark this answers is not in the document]
Flash0ver: User code might set the User explicitly.
And I think that implicit setting of User through any of our integrations, is already guarded by `SendDefaultPii`.

Do I get that right?

src/Sentry/SentryLog.cs
```diff
@@ -133,9 +133,40 @@ public bool TryGetAttribute(string key, [NotNullWhen(true)] out object? value) =
     /// </summary>
     public void SetAttribute(string key, object value) => Attributes.SetAttribute(key, value);
 
-    internal void SetDefaultAttributes(SentryOptions options, SdkVersion sdk) =>
+    internal void SetDefaultAttributes(SentryOptions options, Scope? scope, SdkVersion? sdk = null)
+    {
+        // Core Attributes
+        sdk ??= scope?.Sdk ?? SdkVersion.Instance;
```
Flash0ver, line 139: note: empty `SdkVersion.Name` and `SdkVersion.Version` for Console-Apps

I just noticed that both `Name` and `Version` of `SdkVersion` are empty when running `Sentry.Samples.Console.Basic`, resulting in neither `"sentry.sdk.name"` nor `"sentry.sdk.version"` being set as Attributes. This is probably then also the same issue for Metrics.

This is - I believe - not a new issue, but an existing one that I just noticed ... so something to follow-up on.

src/Sentry/SentryLog.cs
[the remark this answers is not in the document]
jamescrosswell: As far as I can tell, yes. When `SendDefaultPii` is `false` we never put PII on the scope (e.g. [here](https://github.com/getsentry/sentry-dotnet/blob/ac5b47b43930045f49d1671901474a96f9392100/src/Sentry.AspNetCore/ScopeExtensions.cs#L40-L52) and [here](https://github.com/getsentry/sentry-dotnet/blob/6e50eaa87b3aa24a921824c4494566b01f0640f0/src/Sentry.OpenTelemetry/AspNetCoreEnricher.cs#L11-L20)). 

So the above would only ever be a problem if the SDK user themselves had put something sensitive on the scope. We could try do a `SendDefaultPii` check here as well but the user might be confused about why the information they explicitly set on the scope isn't being set. 

I think this is a user code issue rather than an SDK issue.

src/Sentry/Internal/DefaultSentryStructuredLogger.cs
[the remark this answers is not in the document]
jamescrosswell: We should definitely address it. I'm not sure how frequently we'll bump into it but certainly users would expect any values they but in the `configureLog` callback to trump whatever is on the scope (or any 'default' values). 

It's not specifically related to this PR though so yes, let's create a separate issue and then decide whether we want to address it sooner (as a bug fix) or later (as a breaking change). 

### Follow up issue
- https://github.com/getsentry/sentry-dotnet/issues/5351

src/Sentry/SentryLog.cs
jamescrosswell: Added:
- https://github.com/getsentry/sentry-dotnet/issues/5352

PR #5217, opened 2026-05-11
feat: Propagate W3C + Sentry headers in Sentry.OpenTelemetry.Exporter

### LINKED ISSUE #5210: Sentry.OpenTelemetry.Exporter should accept and propagate both Sentry and W3C headers by default
### Problem Statement

The new Sentry.OpenTelemetry.Exporter is awesome, but IMO it's an adoption footgun that out of the box it only accepts/propagates Sentry headers.

I strongly feel that the OOTB default is that it should accept and propagate both Sentry and W3C headers, transparently.

Thanks!

### Solution Brainstorm

`new CompositeTextMapPropagator(new TextMapPropagator[]
    {
        new TraceContextPropagator(),  // W3C traceparent + tracestate
        new SentryPropagator(),        // sentry-trace + baggage (also inherits BaggagePropagator)`

jamescrosswell:
Fixes #5210

## Summary

- The default `TextMapPropagator` used by `AddSentryOtlpExporter` (and `UseOtlp`) was previously `SentryPropagator`, meaning only `sentry-trace` and `baggage` headers were propagated out of the box.
- This PR changes the default to a `CompositeTextMapPropagator` containing both `TraceContextPropagator` (W3C `traceparent`/`tracestate`) and `SentryPropagator` (`sentry-trace` + `baggage`).
- Users who pass a custom `defaultTextMapPropagator` are unaffected — the new default only applies when no propagator is supplied.

### DIFF
--- a/src/Sentry.OpenTelemetry.Exporter/SentryTracerProviderBuilderExtensions.cs
+++ b/src/Sentry.OpenTelemetry.Exporter/SentryTracerProviderBuilderExtensions.cs
@@ -30,12 +30,14 @@ public static class SentryTracerProviderBuilderExtensions
     /// <param name="defaultTextMapPropagator">
     ///     <para>The default TextMapPropagator to be used by OpenTelemetry.</para>
     ///     <para>
-    ///         If this parameter is not supplied, the <see cref="SentryPropagator"/> will be used, which propagates the
-    ///         baggage header as well as Sentry trace headers.
+    ///         If this parameter is not supplied, a <see cref="CompositeTextMapPropagator"/> containing both
+    ///         <see cref="TraceContextPropagator"/> (W3C <c>traceparent</c>/<c>tracestate</c>) and
+    ///         <see cref="SentryPropagator"/> (<c>sentry-trace</c> and <c>baggage</c>) will be used.
+    ///         This allows Sentry to interoperate with services that use W3C trace context headers.
     ///     </para>
     ///     <para>
-    ///         The <see cref="SentryPropagator"/> is required for Sentry's OpenTelemetry integration to work, but you
-    ///         could wrap this in a <see cref="CompositeTextMapPropagator"/> if you needed other propagators as well.
+    ///         The <see cref="SentryPropagator"/> is required for Sentry's OpenTelemetry integration to work. Supply
+    ///         a custom propagator only if you need to replace the defaults entirely.
     ///     </para>
     /// </param>
     /// <returns>The supplied <see cref="TracerProviderBuilder"/> for chaining.</returns>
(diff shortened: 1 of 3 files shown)


ericsampson:
Thank you sir 🫡

src/Sentry.OpenTelemetry.Exporter/SentryTracerProviderBuilderExtensions.cs
```diff
@@ -47,7 +49,11 @@ public static TracerProviderBuilder AddSentryOtlpExporter(this TracerProviderBui
             throw new ArgumentException(MissingDsnWarning, nameof(dsnString));
         }
 
-        defaultTextMapPropagator ??= new SentryPropagator();
+        defaultTextMapPropagator ??= new CompositeTextMapPropagator(new TextMapPropagator[]
```
Flash0ver, line 52: question: `propagateTraceparent` option

Should this option be guarded by the `SentryOptions.PropagateTraceparent`?
Is this a preceding oversight? Or should 

Note: This option is currently only used in the base `SentryMessageHandler` (derived: `SentryHttpMessageHandler` and `SentryGraphQLHttpMessageHandler`)

Flash0ver: follow-up question: docs

In the docs we mention
> The following SDKs support the `propagateTraceparent` option:
see https://docs.sentry.io/concepts/otlp/sentry-with-otel/

Should we list `.NET` there too?

jamescrosswell: > Should this option be guarded by the `SentryOptions.PropagateTraceparent`? Is this a preceding oversight? Or should
> 
> Note: This option is currently only used in the base `SentryMessageHandler` (derived: `SentryHttpMessageHandler` and `SentryGraphQLHttpMessageHandler`)

Hm, interesting question. [According to the docs](https://develop.sentry.dev/sdk/telemetry/traces/otlp/#integration-spec):

> The integration MUST NOT set up an automatic propagator. Cross-service trace propagation is handled by the propagateTraceparent setting or by user-configured OTel propagators. How to set up an OTel propagator is a documentation concern.

So even [the previous behaviour we had](https://github.com/getsentry/sentry-dotnet/blob/42f6ca61e9ba53c2374cc3758c7ccdf04e949cf7/src/Sentry.OpenTelemetry.Exporter/SentryTracerProviderBuilderExtensions.cs#L55) doesn't seem consistent with the docs. When using OTLP, presumably users can and should configure the propagators via OTEL (not Sentry):
- SDK users should call `Sdk.SetDefaultTextMapPropagator` to configure the OTEL propagation settings
-  We should stop calling that in our code:
    https://github.com/getsentry/sentry-dotnet/blob/42f6ca61e9ba53c2374cc3758c7ccdf04e949cf7/src/Sentry.OpenTelemetry.Exporter/SentryTracerProviderBuilderExtensions.cs#L56
- We should make the above clear in the docs for the OTLP integration

@ericsampson this is a bit different to what you were expecting... but it's also how all of our other OTLP integrations behave so I don't think we have much wiggle room on this one.

jamescrosswell: Added:
- https://github.com/getsentry/sentry-dotnet/issues/5299
- https://github.com/getsentry/sentry-dotnet/issues/5300

jamescrosswell:
We shouldn't actually be implementing this. See:
- https://github.com/getsentry/sentry-dotnet/issues/5299

PR #5218, opened 2026-05-11
perf: avoid MeasurementUnit enum boxing allocations

### LINKED ISSUE #4844: perf: allocation-free MeasurementUnit
### Description

Usage of `Sentry.MeasurementUnit` currently incurs two avoidable heap allocations.

```CSharp
MeasurementUnit unit = MeasurementUnit.Duration.Millisecond; // implicit conversion
return unit.ToString(); // overridden `ToString`
```

- implicit conversion from `enum MeasurementUnit.Duration` to `struct MeasurementUnit`
  - invokes the .ctor boxing `MeasurementUnit.Duration` to `System.Enum`
- overridden `ToString` on `MeasurementUnit`
  - invokes `ToString` on the previously boxed `enum`
  - and additionally converting it `ToLowerInvariant`

However

```CSharp
MeasurementUnit unit = MeasurementUnit.Custom("millisecond");
return unit.ToString();
```

is heap-allocation-free, because it uses the other field already of type `string`,
and calling `ToLowerInvariant` on the passed argument, which is no-op if already lower case based on the invariant culture.

jamescrosswell:
## Summary
Closes https://github.com/getsentry/sentry-dotnet/issues/4844 by removing avoidable allocations in `MeasurementUnit`:

- Replace boxed `Enum` storage with compact internal representation (`UnitKind` + `int value`)
- Keep custom units as string-backed values
- Replace `Enum.ToString().ToLowerInvariant()` with precomputed lowercase name lookups for built-in units
- Preserve existing behavior for parsing, equality, hashing, and string serialization

This removes both enum boxing and per-call lowercase string allocation on the enum-backed path.

## Testing
- Ran: `dotnet test SentryNoMobile.slnf`
  - Result: all relevant suites passed for this change
  - Known baseline failures: `test/Sentry.AspNet.Tests` (net48) has 2 failing tests due to `MissingMethodException` in `SentryAspNetOptionsExtensionsTests`, reproducible on `main` with no local changes.

- Reproduced baseline on main:
  - `dotnet test test/Sentry.AspNet.Tests/Sentry.AspNet.Tests.csproj -f net48`
  - Same 2 failures occur on `main`.

#skip-changelog

### DIFF
--- a/src/Sentry/MeasurementUnit.cs
+++ b/src/Sentry/MeasurementUnit.cs
@@ -19,22 +19,41 @@ public readonly partial struct MeasurementUnit : IEquatable<MeasurementUnit>
         Custom = 4
     }
 
-    private static readonly string[] DurationNames = CreateLowercaseNames<Duration>();
+    private static readonly string[] DurationNames =
+    [
+        "nanosecond",
+        "microsecond",
+        "millisecond",
+        "second",
+        "minute",
+        "hour",
+        "day",
+        "week"
+    ];
 
-    private static readonly string[] InformationNames = CreateLowercaseNames<Information>();
+    private static readonly string[] InformationNames =
+    [
+        "bit",
+        "byte",
+        "kilobyte",
+        "kibibyte",
+        "megabyte",
+        "mebibyte",
+        "gigabyte",
+        "gibibyte",
+        "terabyte",
+        "tebibyte",
+        "petabyte",
+        "pebibyte",
+        "exabyte",
+        "exbibyte"
+    ];
 
-    private static readonly string[] FractionNames = CreateLowercaseNames<Fraction>();
-
-    private static string[] CreateLowercaseNames<T>() where T : struct, Enum
-    {
-        var names = Enum.GetNames<T>();
-        for (var i = 0; i < names.Length; i++)
-        {
-            names[i] = names[i].ToLowerInvariant();
-        }
-
-        return names;
-    }
+    private static readonly string[] FractionNames =
+    [
+        "ratio",
+        "percent"
+    ];
 
     private MeasurementUnit(Duration unit)
     {
--- a/src/Sentry/MeasurementUnit.cs
+++ b/src/Sentry/MeasurementUnit.cs
@@ -21,9 +21,9 @@ public readonly partial struct MeasurementUnit : IEquatable<MeasurementUnit>
 
     private static readonly string[] DurationNames = Enum.GetNames<Duration>();
 
-    private static readonly string[] InformationNames =   Enum.GetNames<Information>();
+    private static readonly string[] InformationNames = Enum.GetNames<Information>();
 
-    private static readonly string[] FractionNames =  Enum.GetNames<Fraction>();
+    private static readonly string[] FractionNames = Enum.GetNames<Fraction>();
 
     private MeasurementUnit(Duration unit)
     {
--- a/src/Sentry/MeasurementUnit.cs
+++ b/src/Sentry/MeasurementUnit.cs
@@ -19,41 +19,11 @@ public readonly partial struct MeasurementUnit : IEquatable<MeasurementUnit>
         Custom = 4
     }
 
-    private static readonly string[] DurationNames =
-    [
-        "nanosecond",
-        "microsecond",
-        "millisecond",
-        "second",
-        "minute",
-        "hour",
-        "day",
-        "week"
-    ];
+    private static readonly string[] DurationNames = Enum.GetNames<Duration>();
 
-    private static readonly string[] InformationNames =
-    [
-        "bit",
-        "byte",
-        "kilobyte",
-        "kibibyte",
-        "megabyte",
-        "mebibyte",
-        "gigabyte",
-        "gibibyte",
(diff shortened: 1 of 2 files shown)


src/Sentry/MeasurementUnit.cs
Flash0ver, line 1: thought:

This is a similar implementation a coding agent on my end suggested when I gave it a try when looking at it with a friend of mine just last weekend 😉.

I'm not too certain if I'm happy with this approach, though:
- the size of the struct is increasing
- we embed more data into the assembly (via the `string`-arrays)

On the other hand, this (or a similar) implementation is necessary, in order to not change the `IEquatable`-semantics.

We were then experimenting with the struct only having a `string` field,
but that would change the Equality-semantics, where
- `MeasurementUnit.Duration.Millisecond == MeasurementUnit.Custom("millisecond")`
  - old/this: would not be equal, because constructed differently
  - only with `string` field: would now become equal, being a behavioral change

Flash0ver: wait ... actually ... I believe I am wrong concerning the size ... at least partially wrong

- previous
  - `System.Enum` + `System.String`
  - on 64-bit: 8-byte-reference + 8-byte reference = 16B
  - on 32-bit: 4-byte-reference + 4-byte reference = 8B
- this change
  - `UnitKind` enum (byte) + `System.Int32` + `System.String`
  - on 64-bit: 1-byte enum + 4-byte int + 3-byte padding, 8-byte reference = 16B
  - on 32-bit: 1-byte enum + 3-byte padding + 4-byte int + 4-byte reference + padding = 12B

So we are only increasing the size of the `struct` by 4B on 32-bit systems/processes.

Yeah ... that's neglectable considering we no longer box to `System.Enum`.

Flash0ver: about the Equals-Semantics

If we would only keep a `string` field, assigning the `string` via a `switch` on constructions, we would simplify the code, reduce the struct size to 4B/8B (32-/64-bit systems); but we would change the Equals, as it would no longer matter whether the `MeasurementUnit` was constructed via `MeasurementUnit.Custom` or the respective implicit-ennum-conversion.

This being a behavioral change, we could simplify `MeasurementUnit` in the next Major ... or should we keep the Equals-Behavior as is, where there is a difference how the `MeasurementUnit` was constructed.

What do you think?

src/Sentry/MeasurementUnit.cs
```diff
@@ -6,18 +6,80 @@ namespace Sentry;
...
     {
-        _unit = unit;
+        None = 0,
+        Duration = 1,
+        Information = 2,
+        Fraction = 3,
+        Custom = 4
+    }
+
+    private static readonly string[] DurationNames =
```
Flash0ver, line 22: thought: `switch` instead of Array-Lookups

My AI assistant also suggested a similar solution.

Instead of embedding these arrays into the assembly and then indexing into these arrays,
we could do a `switch` in all of the three (`Duration`, `Fraction`, `Information`) implicit conversions and assign the name/string directly, while still keeping the new `UnitKind` enum so that we don't change Equality-semantics on how the `struct` is created (`MeasurementUnit.Custom` or via any of the three implicit "from-enum"-conversions).

src/Sentry/MeasurementUnit.cs
jamescrosswell: To err on the side of caution (since this API is probably quite widely used already), we could delay the behavioural change (and saving the extra 4B) until the next major release. The only reason to release this sooner is that it would avoid potential merge conflicts and double handling (so efficiency). 

@bitsandfoxes what do you think? Delay or just go ahead and release?

fwiw I think the minor change in behaviour is very unlikely to bite us... but that's just my take.

src/Sentry/MeasurementUnit.cs
jamescrosswell: Thanks @Flash0ver. I didn't much like the duplication in the old code at all. 

We could do a switch statement, but I think it would suffer from the same duplication and whilst it would reduce heap allocations (likley G2 for statics), we grow the size of the codebase by the same amount we save in heap allocations. I think the end result is having to load more data into L1 cache every time the method executes, which might actually be less performant. 

I changed the LLMs code so that the names are generated dynamically in the static constructor (negligible one time hit at startup and allocates a couple of tiny arrays to G2 heap that won't trouble the GC). That makes the code easier to maintain whilst still avoiding the boxing, I believe.

jamescrosswell: Dang... `Enum.GetNames` isn't trim safe. 

Seems we have to go with either hard coded arrays or a source generator. Hard coded is probably easier - will add some unit tests to ensure these don't break if we add any enum values in the future. 

If we did want to use a source generator, this is actually the example that [Andrew Lock uses in his blog about them](https://andrewlock.net/netescapades-enumgenerators-a-source-generator-for-enum-performance/), so he's already built and tested some we could use OOTB.

jamescrosswell: Took the array + test approach

bitsandfoxes (approved):
LGTM!

PR #5230, opened 2026-05-14
build: enable CA2007 enforcement via nested src/.editorconfig

### LINKED ISSUE #2308: Investigate CA2007 checking
In #1277, we switched from `ConfigureAwaitChecker` to built-in analysis rules for CA2007 checking (requiring `ConfigureAwait(false)` in library code).

It was working previously, but for some unknown reason it is now only working in CI builds.  In local development, omitting `ConfigureAwait(false)` is not caught in either Rider or with `dotnet build`.  Investigate why.

jamescrosswell:
## Summary

Move `ConfigureAwait(false)` enforcement (CA2007) from the legacy `src/CodeAnalysis.ruleset` to a nested `src/.editorconfig`.

The ruleset file no longer reliably enables CA2007 with current .NET SDKs (CA2007 is **disabled by default** in modern .NET analyzers, and `<CodeAnalysisRuleSet>` only overrides severity for rules that are already active). As a result, the rule only fired in CI Release builds and was silently ignored locally — both in Rider and with `dotnet build` — the regression reported in #2308.

Using `dotnet_diagnostic.CA2007.severity = error` inside a nested editorconfig under `/src` has two advantages over the previous setup:

- It correctly activates analyzer rules that are disabled by default, so violations fail locally and in CI
- It scopes the rule to library code only. Tests, samples, benchmarks and integration-test projects are unaffected, where calling `ConfigureAwait(false)` is unnecessary noise.

## Verification

Tested locally on macOS:

- Clean build of `src/Sentry/Sentry.csproj` in **Debug** → 0 errors.
- Temporarily replaced `await sentryTask.ConfigureAwait(false);` with `await sentryTask;` in `SpotlightHttpTransport.cs` → build fails with:
  ```
  error CA2007: Consider calling ConfigureAwait on the awaited task
  ```
  ...across all target frameworks (`net8.0`, `net9.0`, `net10.0`). Reverted before commit.
- Built `test/Sentry.Tests/Sentry.Tests.csproj` with its existing `await client.GetAsync(...);` calls (no `ConfigureAwait`) → 0 CA2007 errors, confirming the rule is correctly scoped to `/src` only.

Closes #2308

#skip-changelog

### DIFF
--- a/src/.editorconfig
+++ b/src/.editorconfig
@@ -1,0 +1,9 @@
+# Settings that apply only to library code under /src
+# (so tests, samples, benchmarks, and integration-tests are not affected).
+
+[*.cs]
+
+# CA2007: Consider calling ConfigureAwait on the awaited task.
+# Library code must always ConfigureAwait(false) to avoid deadlocks in
+# callers that have a SynchronizationContext.
+dotnet_diagnostic.CA2007.severity = error
--- a/src/CodeAnalysis.ruleset
+++ b/src/CodeAnalysis.ruleset
@@ -1,5 +1,0 @@
-﻿<RuleSet Name="Microsoft.Analyzers.ManagedCodeAnalysis" Description="Microsoft.Analyzers.ManagedCodeAnalysis" ToolsVersion="14.0">
-  <Rules AnalyzerId="Microsoft.Analyzers.ManagedCodeAnalysis" RuleNamespace="Microsoft.Rules.Managed">
-      <Rule Id="CA2007" Action="Error" />          <!-- Consider calling ConfigureAwait on the awaited task -->
-  </Rules>
-</RuleSet>
--- a/src/Directory.Build.props
+++ b/src/Directory.Build.props
@@ -32,7 +32,6 @@
     <IncludeSourceRevisionInInformationalVersion>false</IncludeSourceRevisionInInformationalVersion>
 
     <Nullable>annotations</Nullable>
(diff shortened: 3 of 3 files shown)


src/.editorconfig
```diff
@@ -0,0 +1,9 @@
+# Settings that apply only to library code under /src
+# (so tests, samples, benchmarks, and integration-tests are not affected).
```
Flash0ver, line 2: question: `.editorconfig` vs `.globalconfig`

I'm wondering, which file is better here:
- `.editorconfig`
- `.globalconfig`

To me, `.editorconfig` is more for style choices,
and `.globalconfig` is more for code quality rules.

The difference on a "content level" is that (unlike `.editorconfig`), `.globalconfig` does not have a language section (e.g. `[*.cs]`).

See also https://learn.microsoft.com/dotnet/fundamentals/code-analysis/configuration-files#global-analyzerconfig

But ... to be fair ... this might be just total bikeshedding from my end, and is more of a "stylistic" choice rather than a "functional" choice.

I just wanted to raise awareness that there is also a `.globalconfig` option we could use.

Flash0ver (approved):
One optional comment.

src/.editorconfig
jamescrosswell: Let's change if it ever causes us issues further down the road. For the time being, I think this does what we want and we can move on to other stuff.

PR #5231, opened 2026-05-15
build: download ACKNOWLEDGEMENTS.md for sentry-cli when packing

### LINKED ISSUE #4588: Pack acknowledgements for third-party libraries of Sentry CLI
### Description

Most permissive OSS licenses require "ATTRIBUTION", or "NOTICE", or "ACKNOWLEDGEMENTS", or similar, when (parts of) the software are re-packaged/re-distributed.

Most of our dependencies are through NuGet, so we don't re-package/re-distribute these.

We do include the [sentry-cli](https://github.com/getsentry/sentry-cli) in our `Sentry` NuGet package,
which does directly ship (parts of) 3rd party software, and these do need to be mentioned.
Since [Sentry](https://www.nuget.org/packages/Sentry) does ship `./tools/sentry-cli-*` executables,
we need to include this new file as well.

See also
- getsentry/sentry-cli#2823
- getsentry/sentry-cli#2821

jamescrosswell:
## Summary

The `Sentry` NuGet package bundles `sentry-cli` executables under `tools/`. Those binaries statically compile dozens of third-party Rust crates whose OSS licenses (Apache 2.0, MIT, etc.) require attribution when the software is redistributed. This PR adds the infrastructure to download and pack the `ACKNOWLEDGEMENTS.md` that sentry-cli will publish alongside its binaries.

### What changed

- Inside `DownloadSentryCLI`, added a `DownloadFile` step for the platform-independent `ACKNOWLEDGEMENTS.md` from `https://downloads.sentry-cdn.com/sentry-cli/<version>/ACKNOWLEDGEMENTS.md`.
- The existing `<None Include="$(SentryCLIDirectory)**" Pack="true" PackagePath="tools\">` wildcard already covers this file, so no extra pack entry is needed.
- `ContinueOnError="WarnAndContinue"` is used while [getsentry/sentry-cli#2823](https://github.com/getsentry/sentry-cli/issues/2823) is still in progress (the file is not yet published in sentry-cli releases). Once that ships:
  - Remove `ContinueOnError`.
  - Add the file hash to the release registry spec and extend `update-cli.ps1` to update it.

### Testing

- Local build emits a single MSBuild warning (`MSB3923: Failed to download`) for the missing file — not an error — confirming graceful degradation.
- Once sentry-cli#2823 ships, a version-bump via `update-cli.ps1 set-version <new>` + publish build will pull the file and pack it automatically.

Closes #4588

#skip-changelog

### DIFF
--- a/src/Sentry/Sentry.csproj
+++ b/src/Sentry/Sentry.csproj
@@ -116,13 +116,27 @@
       <SentryCLIDownload Condition="'$(CI_PUBLISHING_BUILD)' == 'true' Or ($([MSBuild]::IsOSPlatform('Windows')) And $(_OSArchitecture) != 'X86')" Include="sentry-cli-Windows-x86_64.exe" FileHash="572b4a6c04504302a46e2e1c196ec31cda266b574b1444605426aa8e41d46b8c" />
     </ItemGroup>
 
-    <!-- Download the files -->
+    <!-- Download the platform-specific CLI binaries. -->
     <DownloadFile SourceUrl="https://downloads.sentry-cdn.com/sentry-cli/$(SentryCLIVersion)/%(SentryCLIDownload.Identity)" DestinationFolder="$(SentryCLIDirectory)" Condition="!Exists('$(SentryCLIDirectory)%(Identity)')" Retries="3">
       <Output TaskParameter="DownloadedFile" ItemName="SentryCLIDownloadedFile" />
     </DownloadFile>
 
     <!-- Build will fail if any downloaded files don't match the expected hash. -->
     <VerifyFileHash File="$(SentryCLIDirectory)%(SentryCLIDownload.Identity)" Hash="%(FileHash)" />
+
+    <!--
+      Download the platform-independent ACKNOWLEDGEMENTS.md that lists third-party OSS
+      attributions compiled into the sentry-cli binary. Required for license compliance
+      when redistributing the CLI executables in the Sentry NuGet package.
+      ContinueOnError is temporary while getsentry/sentry-cli#2823 is in progress —
+      remove it once sentry-cli starts publishing this file in releases.
+    -->
+    <DownloadFile
+      SourceUrl="https://downloads.sentry-cdn.com/sentry-cli/$(SentryCLIVersion)/ACKNOWLEDGEMENTS.md"
+      DestinationFolder="$(SentryCLIDirectory)"
+      Condition="!Exists('$(SentryCLIDirectory)ACKNOWLEDGEMENTS.md')"
+      ContinueOnError="WarnAndContinue"
+      Retries="3" />
 
     <!-- Set executable permissions for local usage. -->
     <Exec Command="chmod +x $(SentryCLIDirectory)*" Condition="!$([MSBuild]::IsOSPlatform('Windows'))" />


jamescrosswell:
@Flash0ver is this what you had in mind?

src/Sentry/Sentry.csproj
```diff
@@ -116,14 +116,28 @@
...
+      attributions compiled into the sentry-cli binary. Required for license compliance
+      when redistributing the CLI executables in the Sentry NuGet package.
+      ContinueOnError is temporary while getsentry/sentry-cli#2823 is in progress —
+      remove it once sentry-cli starts publishing this file in releases.
+    -->
+    <DownloadFile
+      SourceUrl="https://downloads.sentry-cdn.com/sentry-cli/$(SentryCLIVersion)/ACKNOWLEDGEMENTS.md"
+      DestinationFolder="$(SentryCLIDirectory)"
+      Condition="!Exists('$(SentryCLIDirectory)ACKNOWLEDGEMENTS.md')"
+      ContinueOnError="WarnAndContinue"
```
Flash0ver, line 138: question: should we do this now?

Since getsentry/sentry-cli#2823 is not complete yet, this will always fail.
Should we wait until getsentry/sentry-cli#2823 is completed?

src/Sentry/Sentry.csproj
```diff
@@ -116,14 +116,28 @@
...
     <VerifyFileHash File="$(SentryCLIDirectory)%(SentryCLIDownload.Identity)" Hash="%(FileHash)" />
 
+    <!--
+      Download the platform-independent ACKNOWLEDGEMENTS.md that lists third-party OSS
+      attributions compiled into the sentry-cli binary. Required for license compliance
+      when redistributing the CLI executables in the Sentry NuGet package.
+      ContinueOnError is temporary while getsentry/sentry-cli#2823 is in progress —
+      remove it once sentry-cli starts publishing this file in releases.
+    -->
+    <DownloadFile
```
Flash0ver, line 134: question: multi-targeting

Is this task executed multiple times, do to multiple `TargetFrameworks`?
If so - can we easily ensure to only run once?

src/Sentry/Sentry.csproj
```diff
@@ -116,14 +116,28 @@
...
+      SourceUrl="https://downloads.sentry-cdn.com/sentry-cli/$(SentryCLIVersion)/ACKNOWLEDGEMENTS.md"
```
Flash0ver, line 135: question: file hash

Could / Should we verify this file's hash, too?
Well ... a `.md` file isn't executed ... but there might be a legal compromise.

UPDATE: oh wait ... we can only do this, once the file is actually shipped

src/Sentry/Sentry.csproj
```diff
@@ -116,14 +116,28 @@
...
+      DestinationFolder="$(SentryCLIDirectory)"
+      Condition="!Exists('$(SentryCLIDirectory)ACKNOWLEDGEMENTS.md')"
+      ContinueOnError="WarnAndContinue"
+      Retries="3" />
+
     <!-- Set executable permissions for local usage. -->
     <Exec Command="chmod +x $(SentryCLIDirectory)*" Condition="!$([MSBuild]::IsOSPlatform('Windows'))" />
   </Target>
```
Flash0ver, line 143: question: make `.md` file executable

Already existing code ... but would we make the `.md` file with the existing/following `Exec` task "execute" too? Could this be an issue?

Flash0ver (commented):
Not sure if we should do this before getsentry/sentry-cli#2823 is completed and shipped.

src/Sentry/Sentry.csproj
jamescrosswell: How would that work? What would we be checking the file hash against (and why)? What would the process be if/when the `ACKNOWLEDGEMENTS.md` file in thes sentry-cli changed?

src/Sentry/Sentry.csproj
jamescrosswell: Yeah I've marked the issue this closes as blocked by 2823.

src/Sentry/Sentry.csproj
Flash0ver: right ... no File-Hash is published for the `ACKNOWLEDGEMENTS.md`, that also the dependencies-updater could keep in sync ... only for the `sentry-cli` binaries per platform

PR #5236, opened 2026-05-15
fix: exclude SentrySqlListener from .NET Framework builds

### LINKED ISSUE #3173: SqlListener should not be compiled/available in NetFX
See the [thread on Issue #3115](https://github.com/getsentry/sentry-dotnet/issues/3115#issuecomment-1958492025_) for context. 

Essentially our [SqlListener won't work in NetFx](https://github.com/dotnet/SqlClient/issues/1529#issuecomment-1063166442), so it's a bit misleading to include it in the Net4x packages. 

We could add some conditional compilation blocks to strip it out for that target, along with some comments explaining why.

jamescrosswell:
## Summary

SqlClient's DiagnosticSource-based events are not delivered on .NET Framework (see [dotnet/SqlClient#1529](https://github.com/dotnet/SqlClient/issues/1529#issuecomment-1063166442)), so including `SentrySqlListener` in the `net462` package is misleading — it compiles and registers but silently does nothing.

- Wrapped `SentrySqlListener` entirely in `#if !NETFRAMEWORK`
- Guarded the SQL listener field, lazy initializer, and `SqlClientDiagnosticListener` subscription in `SentryDiagnosticSubscriber` with the same guard
- Excluded `SentrySqlListenerExtensions` and `SentrySqlListenerTests` from .NET Framework test builds

All existing tests pass unchanged on `net9.0`.

#skip-changelog

Closes #3173

### DIFF
--- a/src/Sentry.DiagnosticSource/Internal/DiagnosticSource/SentryDiagnosticSubscriber.cs
+++ b/src/Sentry.DiagnosticSource/Internal/DiagnosticSource/SentryDiagnosticSubscriber.cs
@@ -13,7 +13,9 @@ internal class SentryDiagnosticSubscriber : IObserver<DiagnosticListener>
     // Thus, we will create the instances lazily.
 
     private readonly Lazy<SentryEFCoreListener> _efListener;
+#if !NETFRAMEWORK
     private readonly Lazy<SentrySqlListener> _sqlListener;
+#endif
 
     public SentryDiagnosticSubscriber(IHub hub, SentryOptions options)
     {
@@ -23,11 +25,13 @@ internal class SentryDiagnosticSubscriber : IObserver<DiagnosticListener>
             return new SentryEFCoreListener(hub, options);
         });
 
+#if !NETFRAMEWORK
         _sqlListener = new Lazy<SentrySqlListener>(() =>
         {
             options.Log(SentryLevel.Debug, "Registering SQL Client integration.");
             return new SentrySqlListener(hub, options);
         });
+#endif
     }
 
     public void OnCompleted() { }
@@ -44,13 +48,16 @@ internal class SentryDiagnosticSubscriber : IObserver<DiagnosticListener>
                     break;
                 }
 
+#if !NETFRAMEWORK
             case "SqlClientDiagnosticListener":
                 {
                     listener.Subscribe(_sqlListener.Value);
                     break;
                 }
+#endif
         }
 
+#if !NETFRAMEWORK
         // By default, the EF listener will duplicate spans already given by the SQL Client listener.
         // Thus, we should disable those parts of the EF listener when they are both registered.
         if (_efListener.IsValueCreated && _sqlListener.IsValueCreated)
@@ -59,5 +66,6 @@ internal class SentryDiagnosticSubscriber : IObserver<DiagnosticListener>
             efListener.DisableConnectionSpan();
             efListener.DisableQuerySpan();
         }
+#endif
     }
 }
--- a/src/Sentry.DiagnosticSource/Internal/DiagnosticSource/SentrySqlListener.cs
+++ b/src/Sentry.DiagnosticSource/Internal/DiagnosticSource/SentrySqlListener.cs
@@ -1,3 +1,7 @@
+// SqlClient's DiagnosticSource integration is not supported on .NET Framework.
+// See: https://github.com/dotnet/SqlClient/issues/1529#issuecomment-1063166442
+#if !NETFRAMEWORK
+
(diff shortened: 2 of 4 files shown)


src/Sentry.DiagnosticSource/Internal/DiagnosticSource/SentrySqlListener.cs
```diff
@@ -1,3 +1,7 @@
+// SqlClient's DiagnosticSource integration is not supported on .NET Framework.
+// See: https://github.com/dotnet/SqlClient/issues/1529#issuecomment-1063166442
+#if !NETFRAMEWORK
+
```
Flash0ver, line 4: question: docs

Should we mention this limitation of SqlClient on .NET Framework in both our `README.md` and the docs? (perhaps also linking to the related issue in `dotnet/SqlClient`)

- README: `src/Sentry.DiagnosticSource/README.md`
- docs: https://docs.sentry.io/platforms/dotnet/performance/instrumentation/automatic-instrumentation/#diagnosticsource-integration

src/Sentry.DiagnosticSource/Internal/DiagnosticSource/SentryDiagnosticSubscriber.cs
```diff
@@ -13,7 +13,9 @@ internal class SentryDiagnosticSubscriber : IObserver<DiagnosticListener>
     // Thus, we will create the instances lazily.
 
     private readonly Lazy<SentryEFCoreListener> _efListener;
+#if !NETFRAMEWORK
     private readonly Lazy<SentrySqlListener> _sqlListener;
+#endif
```
Flash0ver, line 18: thought: indirect consumption

User applications that consume Sentry indirectly, via a `netstandard2.0` library that is referenced from a .NET Framework application. In this case there would be a `netfx` `SqlClient`, but still the `netstandard2.0` asset of Sentry is consumed that includes the `SentrySqlListener`. But this is a bit of a corner-case, not really a problem (as the `SentrySqlListener` wouldn't work either way). Also, this would be easily "fixable" in user code by multi-targeting that Sentry-consuming library to `net462+`.

Flash0ver (approved):
- 1 non-blocking potential follow-up
- 1 non-actionable thought

src/Sentry.DiagnosticSource/Internal/DiagnosticSource/SentryDiagnosticSubscriber.cs
jamescrosswell: Definitely a corner case. I can't think of anything we can do about this (certainly this PR doesn't make that situation worse so if we do think that's a problem, we can address it in a separate follow up issue).

src/Sentry.DiagnosticSource/Internal/DiagnosticSource/SentrySqlListener.cs
jamescrosswell: We should yes... although when I think how those docs would read it starts to create a bit of an existential crisis for this package. 

Basically this package now provides EFCore instrumentation for NetFX (anyone on .NET Core would get that directly from the Sentry core library). It also provides either EFCore or SqlClient instrumentation for netstandard targets - but in what scenarios would those be used?
- If someone is building a NetFX app, they should use the NetFX target and they only get the EFCore diagnostics source functionality. 
- Folks building .NET Core. apps should be getting the DiagnosticSource integration [directly via the Sentry package](https://github.com/getsentry/sentry-dotnet/blob/1ecbc971159a65a1723288e3b7bfcdf49d624e0e/src/Sentry/Sentry.csproj#L53-L61)
- The only others I can think of are. Unity (unlikely to need this anyway), Xamarin (which we don't support anymore) and Mono (basically just used as a runtime sometimes when debugging iOS apps or on Android - both of which imply recent versions of .NET Core)

Maybe we should look refactoring this package in v7 to:
- Drop the netstandard targets
- Make it clear in the description that it's only needed if people want to instrument EntityFrameworkCore on NetFX (DiagnosticsSource integrations for EFCore and SqlClient come OOTB for .NET Core TFMs in the Sentry package)
- Possibly consider renaming it `Sentry.DiagnosticSource.EntityFrameworkCore` or `Sentry.EntityFrameworkCore` or something so that the name more accurately represents it's reason for being

jamescrosswell: There is no specific readme file for this package, but I've opened a PR to make it clear in the Sentry docs:
-  https://github.com/getsentry/sentry-docs/pull/18559

PR #5240, opened 2026-05-17
ref: replace concrete TargetFramework string comparisons with TargetFrameworkIdentifier/Version properties

### LINKED ISSUE #2657: Remove references to specific target frameworks from csproj files
See https://x.com/Tyrrrz/status/1706262274909065656?s=20

💡 Your [#dotnet](https://twitter.com/hashtag/dotnet?src=hashtag_click) project file contains conditional item or property groups that depend on specific target frameworks?

❌ Don't check for concrete values of $(TargetFramework). This doesn't scale well if your project has several target frameworks.

✅ Do this instead:
image
image

jamescrosswell:
## Summary

Replace hardcoded `$(TargetFramework)` string comparisons in `.csproj` files with the more scalable `$(TargetFrameworkIdentifier)` + `$(TargetFrameworkVersion)` property checks, as recommended by [MSBuild best practices](https://x.com/Tyrrrz/status/1706262274909065656).

- `'$(TargetFramework)' == 'net8.0'` → `'$(TargetFrameworkIdentifier)' == '.NETCoreApp' And '$(TargetFrameworkVersion)' == 'v8.0'`
- `'$(TargetFramework)' == 'net48'` → `'$(TargetFrameworkIdentifier)' == '.NETFramework' And '$(TargetFrameworkVersion)' == 'v4.8'`
- `'$(TargetFramework)' == 'netstandard2.0'` → `'$(TargetFrameworkIdentifier)' == '.NETStandard' And '$(TargetFrameworkVersion)' == 'v2.0'`
- `$(TargetFramework.Contains('-ios'))` → `'$(TargetPlatformIdentifier)' == 'ios'`

14 files updated across `src/`, `test/`, and `samples/`.

#skip-changelog

Closes #2657

### DIFF
--- a/samples/Sentry.Samples.Console.Basic/Sentry.Samples.Console.Basic.csproj
+++ b/samples/Sentry.Samples.Console.Basic/Sentry.Samples.Console.Basic.csproj
@@ -56,7 +56,7 @@
     <ProjectReference Include="..\..\src\Sentry\Sentry.csproj" />
   </ItemGroup>
 
-  <ItemGroup Condition="'$(TargetFramework)' == 'net462'">
+  <ItemGroup Condition="'$(TargetFrameworkIdentifier)' == '.NETFramework' And '$(TargetFrameworkVersion)' == 'v4.6.2'">
     <PackageReference Include="System.Net.Http" Version="4.3.4" />
   </ItemGroup>
 
--- a/samples/Sentry.Samples.Console.Basic/Sentry.Samples.Console.Basic.csproj
+++ b/samples/Sentry.Samples.Console.Basic/Sentry.Samples.Console.Basic.csproj
@@ -56,7 +56,7 @@
     <ProjectReference Include="..\..\src\Sentry\Sentry.csproj" />
   </ItemGroup>
 
-  <ItemGroup Condition="'$(TargetFrameworkIdentifier)' == '.NETFramework' And '$(TargetFrameworkVersion)' == 'v4.6.2'">
+  <ItemGroup Condition="'$(TargetFrameworkIdentifier)' == '.NETFramework'">
     <PackageReference Include="System.Net.Http" Version="4.3.4" />
   </ItemGroup>
 
--- a/samples/Sentry.Samples.Console.Basic/Sentry.Samples.Console.Basic.csproj
+++ b/samples/Sentry.Samples.Console.Basic/Sentry.Samples.Console.Basic.csproj
@@ -6,7 +6,7 @@
     <Nullable>enable</Nullable>
     <TargetFrameworks>net9.0;net10.0;net462</TargetFrameworks>
   </PropertyGroup>
-  <PropertyGroup Condition="!$(TargetFramework.StartsWith('net462'))">
+  <PropertyGroup Condition="'$(TargetFrameworkIdentifier)' != '.NETFramework'">
     <PublishAot>true</PublishAot>
     <PublishTrimmed>true</PublishTrimmed>
   </PropertyGroup>
@@ -56,7 +56,7 @@
     <ProjectReference Include="..\..\src\Sentry\Sentry.csproj" />
   </ItemGroup>
 
-  <ItemGroup Condition="'$(TargetFrameworkIdentifier)' == '.NETFramework' And '$(TargetFrameworkVersion)' == 'v4.6.2'">
+  <ItemGroup Condition="'$(TargetFrameworkIdentifier)' == '.NETFramework'">
     <PackageReference Include="System.Net.Http" Version="4.3.4" />
   </ItemGroup>
 
--- a/src/Sentry.EntityFramework/Sentry.EntityFramework.csproj
+++ b/src/Sentry.EntityFramework/Sentry.EntityFramework.csproj
@@ -21,11 +21,11 @@
     <Using Include="Sentry.Integrations" />
   </ItemGroup>
 
-  <ItemGroup Condition="'$(TargetFramework)' == 'net462'">
+  <ItemGroup Condition="'$(TargetFrameworkIdentifier)' == '.NETFramework' And '$(TargetFrameworkVersion)' == 'v4.6.2'">
     <PackageReference Include="EntityFramework" Version="6.2.0" />
   </ItemGroup>
 
-  <ItemGroup Condition="'$(TargetFramework)' == 'netstandard2.1'">
+  <ItemGroup Condition="'$(TargetFrameworkIdentifier)' == '.NETStandard' And '$(TargetFrameworkVersion)' == 'v2.1'">
     <PackageReference Include="EntityFramework" Version="6.5.1" />
   </ItemGroup>
 
--- a/src/Sentry.EntityFramework/Sentry.EntityFramework.csproj
+++ b/src/Sentry.EntityFramework/Sentry.EntityFramework.csproj
@@ -21,7 +21,7 @@
     <Using Include="Sentry.Integrations" />
   </ItemGroup>
 
-  <ItemGroup Condition="'$(TargetFrameworkIdentifier)' == '.NETFramework' And '$(TargetFrameworkVersion)' == 'v4.6.2'">
+  <ItemGroup Condition="'$(TargetFrameworkIdentifier)' == '.NETFramework'">
     <PackageReference Include="EntityFramework" Version="6.2.0" />
   </ItemGroup>
 
--- a/src/Sentry.EntityFramework/Sentry.EntityFramework.csproj
+++ b/src/Sentry.EntityFramework/Sentry.EntityFramework.csproj
@@ -21,11 +21,11 @@
     <Using Include="Sentry.Integrations" />
   </ItemGroup>
 
-  <ItemGroup Condition="'$(TargetFrameworkIdentifier)' == '.NETFramework' And '$(TargetFrameworkVersion)' == 'v4.6.2'">
+  <ItemGroup Condition="'$(TargetFrameworkIdentifier)' == '.NETFramework'">
     <PackageReference Include="EntityFramework" Version="6.2.0" />
   </ItemGroup>
 
-  <ItemGroup Condition="'$(TargetFrameworkIdentifier)' == '.NETStandard' And '$(TargetFrameworkVersion)' == 'v2.1'">
+  <ItemGroup Condition="'$(TargetFrameworkIdentifier)' == '.NETStandard'">
     <PackageReference Include="EntityFramework" Version="6.5.1" />
   </ItemGroup>
 
--- a/src/Sentry.Extensions.AI/Sentry.Extensions.AI.csproj
+++ b/src/Sentry.Extensions.AI/Sentry.Extensions.AI.csproj
@@ -19,7 +19,7 @@
     <PackageReference Include="Microsoft.Extensions.AI.Abstractions" Version="9.7.0" />
   </ItemGroup>
 
-  <ItemGroup Condition="'$(TargetFramework)' == 'netstandard2.0'">
+  <ItemGroup Condition="'$(TargetFrameworkIdentifier)' == '.NETStandard' And '$(TargetFrameworkVersion)' == 'v2.0'">
     <PackageReference Include="System.Diagnostics.DiagnosticSource" Version="9.0.0" />
   </ItemGroup>
 
--- a/src/Sentry.Extensions.AI/Sentry.Extensions.AI.csproj
+++ b/src/Sentry.Extensions.AI/Sentry.Extensions.AI.csproj
@@ -19,7 +19,7 @@
     <PackageReference Include="Microsoft.Extensions.AI.Abstractions" Version="9.7.0" />
   </ItemGroup>
 
-  <ItemGroup Condition="'$(TargetFrameworkIdentifier)' == '.NETStandard' And '$(TargetFrameworkVersion)' == 'v2.0'">
+  <ItemGroup Condition="'$(TargetFrameworkIdentifier)' == '.NETStandard'">
     <PackageReference Include="System.Diagnostics.DiagnosticSource" Version="9.0.0" />
   </ItemGroup>
 
--- a/src/Sentry.Extensions.Logging/Sentry.Extensions.Logging.csproj
+++ b/src/Sentry.Extensions.Logging/Sentry.Extensions.Logging.csproj
@@ -15,24 +15,24 @@
     <ProjectReference Include="..\..\src\Sentry\Sentry.csproj" />
   </ItemGroup>
 
-  <ItemGroup Condition="'$(TargetFramework)' == 'netstandard2.0'">
+  <ItemGroup Condition="'$(TargetFrameworkIdentifier)' == '.NETStandard' And '$(TargetFrameworkVersion)' == 'v2.0'">
     <PackageReference Include="Microsoft.Extensions.Logging.Configuration" Version="2.1.0" />
     <PackageReference Include="Microsoft.Extensions.Http" Version="2.1.0" />
   </ItemGroup>
 
-  <ItemGroup Condition="'$(TargetFramework)' == 'net8.0'">
+  <ItemGroup Condition="'$(TargetFrameworkIdentifier)' == '.NETCoreApp' And '$(TargetFrameworkVersion)' == 'v8.0'">
     <PackageReference Include="Microsoft.Extensions.Configuration.Binder" Version="8.0.0" />
     <PackageReference Include="Microsoft.Extensions.Logging.Configuration" Version="8.0.0" />
     <PackageReference Include="Microsoft.Extensions.Http" Version="8.0.0" />
   </ItemGroup>
 
-  <ItemGroup Condition="'$(TargetFramework)' == 'net9.0'">
+  <ItemGroup Condition="'$(TargetFrameworkIdentifier)' == '.NETCoreApp' And '$(TargetFrameworkVersion)' == 'v9.0'">
     <PackageReference Include="Microsoft.Extensions.Configuration.Binder" Version="9.0.0" />
     <PackageReference Include="Microsoft.Extensions.Logging.Configuration" Version="9.0.0" />
     <PackageReference Include="Microsoft.Extensions.Http" Version="9.0.0" />
   </ItemGroup>
 
-  <ItemGroup Condition="'$(TargetFramework)' == 'net10.0'">
+  <ItemGroup Condition="'$(TargetFrameworkIdentifier)' == '.NETCoreApp' And '$(TargetFrameworkVersion)' == 'v10.0'">
     <PackageReference Include="Microsoft.Extensions.Configuration.Binder" Version="10.0.0" />
     <PackageReference Include="Microsoft.Extensions.Logging.Configuration" Version="10.0.0" />
     <PackageReference Include="Microsoft.Extensions.Http" Version="10.0.0" />
--- a/src/Sentry.Extensions.Logging/Sentry.Extensions.Logging.csproj
+++ b/src/Sentry.Extensions.Logging/Sentry.Extensions.Logging.csproj
@@ -15,7 +15,7 @@
     <ProjectReference Include="..\..\src\Sentry\Sentry.csproj" />
   </ItemGroup>
 
-  <ItemGroup Condition="'$(TargetFrameworkIdentifier)' == '.NETStandard' And '$(TargetFrameworkVersion)' == 'v2.0'">
+  <ItemGroup Condition="'$(TargetFrameworkIdentifier)' == '.NETStandard'">
     <PackageReference Include="Microsoft.Extensions.Logging.Configuration" Version="2.1.0" />
     <PackageReference Include="Microsoft.Extensions.Http" Version="2.1.0" />
   </ItemGroup>
--- a/src/Sentry.Serilog/Sentry.Serilog.csproj
+++ b/src/Sentry.Serilog/Sentry.Serilog.csproj
@@ -27,11 +27,11 @@
     <Using Include="Serilog.Formatting" />
   </ItemGroup>
 
-  <ItemGroup Condition="$(TargetFramework) == 'net462' or $(TargetFramework) == 'netstandard2.0'">
+  <ItemGroup Condition="'$(TargetFrameworkIdentifier)' == '.NETFramework' Or ('$(TargetFrameworkIdentifier)' == '.NETStandard' And '$(TargetFrameworkVersion)' == 'v2.0')">
     <PackageReference Include="Serilog" Version="2.7.1" />
   </ItemGroup>
 
-  <ItemGroup Condition="$(TargetFramework) != 'net462' and $(TargetFramework) != 'netstandard2.0'">
+  <ItemGroup Condition="'$(TargetFrameworkIdentifier)' != '.NETFramework' And ('$(TargetFrameworkIdentifier)' != '.NETStandard' Or '$(TargetFrameworkVersion)' != 'v2.0')">
     <PackageReference Include="Serilog" Version="2.10.0" />
   </ItemGroup>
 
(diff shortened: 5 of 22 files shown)


samples/Sentry.Samples.Console.Basic/Sentry.Samples.Console.Basic.csproj
```diff
@@ -56,7 +56,7 @@
     <ProjectReference Include="..\..\src\Sentry\Sentry.csproj" />
   </ItemGroup>
 
-  <ItemGroup Condition="'$(TargetFramework)' == 'net462'">
+  <ItemGroup Condition="'$(TargetFrameworkIdentifier)' == '.NETFramework' And '$(TargetFrameworkVersion)' == 'v4.6.2'">
```
jamescrosswell:
```suggestion
  <ItemGroup Condition="'$(TargetFrameworkIdentifier)' == '.NETFramework'">
```

We don't really care about the version here... just that it's  netfx.

src/Sentry.EntityFramework/Sentry.EntityFramework.csproj
```diff
@@ -21,11 +21,11 @@
     <Using Include="Sentry.Integrations" />
   </ItemGroup>
 
-  <ItemGroup Condition="'$(TargetFramework)' == 'net462'">
+  <ItemGroup Condition="'$(TargetFrameworkIdentifier)' == '.NETFramework' And '$(TargetFrameworkVersion)' == 'v4.6.2'">
```
jamescrosswell:
```suggestion
  <ItemGroup Condition="'$(TargetFrameworkIdentifier)' == '.NETFramework'">
```

This is specific enough.

src/Sentry.EntityFramework/Sentry.EntityFramework.csproj
```diff
@@ -21,11 +21,11 @@
...
     <PackageReference Include="EntityFramework" Version="6.2.0" />
   </ItemGroup>
 
-  <ItemGroup Condition="'$(TargetFramework)' == 'netstandard2.1'">
+  <ItemGroup Condition="'$(TargetFrameworkIdentifier)' == '.NETStandard' And '$(TargetFrameworkVersion)' == 'v2.1'">
```
jamescrosswell:
```suggestion
  <ItemGroup Condition="'$(TargetFrameworkIdentifier)' == '.NETStandard'">
```

This is specific enough - we only target one version of netstandard.

src/Sentry.Extensions.AI/Sentry.Extensions.AI.csproj
```diff
@@ -19,7 +19,7 @@
     <PackageReference Include="Microsoft.Extensions.AI.Abstractions" Version="9.7.0" />
   </ItemGroup>
 
-  <ItemGroup Condition="'$(TargetFramework)' == 'netstandard2.0'">
+  <ItemGroup Condition="'$(TargetFrameworkIdentifier)' == '.NETStandard' And '$(TargetFrameworkVersion)' == 'v2.0'">
```
jamescrosswell:
```suggestion
  <ItemGroup Condition="'$(TargetFrameworkIdentifier)' == '.NETStandard'">
```

Specific enough.

src/Sentry.Extensions.Logging/Sentry.Extensions.Logging.csproj
```diff
@@ -15,24 +15,24 @@
     <ProjectReference Include="..\..\src\Sentry\Sentry.csproj" />
   </ItemGroup>
 
-  <ItemGroup Condition="'$(TargetFramework)' == 'netstandard2.0'">
+  <ItemGroup Condition="'$(TargetFrameworkIdentifier)' == '.NETStandard' And '$(TargetFrameworkVersion)' == 'v2.0'">
```
jamescrosswell:
```suggestion
  <ItemGroup Condition="'$(TargetFrameworkIdentifier)' == '.NETStandard'">
```

Specific enough.

test/Sentry.AspNetCore.TestUtils/Sentry.AspNetCore.TestUtils.csproj
```diff
@@ -11,7 +11,7 @@
     <ProjectReference Include="..\Sentry.Testing\Sentry.Testing.csproj" />
   </ItemGroup>
 
-  <ItemGroup Condition="$(TargetFramework) != 'net48'">
+  <ItemGroup Condition="'$(TargetFrameworkIdentifier)' != '.NETFramework' Or '$(TargetFrameworkVersion)' != 'v4.8'">
```
jamescrosswell:
```suggestion
  <ItemGroup Condition="'$(TargetFrameworkIdentifier)' != '.NETFramework'">
```

test/Sentry.AspNetCore.TestUtils/Sentry.AspNetCore.TestUtils.csproj
```diff
@@ -29,7 +29,7 @@
   an updated version of the ASP.NET Core runtime on their hosting server.
   See https://github.com/dotnet/aspnetcore/issues/15423
 -->
-  <ItemGroup Condition="$(TargetFramework) == 'net48'">
+  <ItemGroup Condition="'$(TargetFrameworkIdentifier)' == '.NETFramework' And '$(TargetFrameworkVersion)' == 'v4.8'">
```
jamescrosswell:
```suggestion
  <ItemGroup Condition="'$(TargetFrameworkIdentifier)' == '.NETFramework'">
```

test/Sentry.DiagnosticSource.IntegrationTests/Sentry.DiagnosticSource.IntegrationTests.csproj
```diff
@@ -34,7 +34,7 @@
   </ItemGroup>
 
   <!-- Test EF Core 3.1 on .NET Framework -->
-  <ItemGroup Condition="'$(TargetFramework)' == 'net48'">
+  <ItemGroup Condition="'$(TargetFrameworkIdentifier)' == '.NETFramework' And '$(TargetFrameworkVersion)' == 'v4.8'">
```
jamescrosswell:
```suggestion
  <ItemGroup Condition="'$(TargetFrameworkIdentifier)' == '.NETFramework'">
```

test/Sentry.DiagnosticSource.Tests/Sentry.DiagnosticSource.Tests.csproj
```diff
@@ -7,27 +7,27 @@
...
+  <ItemGroup Condition="'$(TargetFrameworkIdentifier)' == '.NETCoreApp' And '$(TargetFrameworkVersion)' == 'v8.0'">
     <PackageReference Include="Microsoft.EntityFrameworkCore" Version="8.0.0" />
     <PackageReference Include="Microsoft.EntityFrameworkCore.Sqlite" Version="8.0.0" />
     <!-- https://github.com/advisories/GHSA-qj66-m88j-hmgj -->
     <PackageReference Include="Microsoft.Extensions.Caching.Memory" Version="8.0.1" />
   </ItemGroup>
 
   <!-- Test .NET Framework -->
-  <ItemGroup Condition="'$(TargetFramework)' == 'net48'">
+  <ItemGroup Condition="'$(TargetFrameworkIdentifier)' == '.NETFramework' And '$(TargetFrameworkVersion)' == 'v4.8'">
```
jamescrosswell:
```suggestion
  <ItemGroup Condition="'$(TargetFrameworkIdentifier)' == '.NETFramework'">
```

test/Sentry.Extensions.Logging.Tests/Sentry.Extensions.Logging.Tests.csproj
```diff
@@ -15,7 +15,7 @@
     <ProjectReference Include="..\Sentry.Testing\Sentry.Testing.csproj" />
   </ItemGroup>
 
-  <ItemGroup Condition="'$(TargetFramework)' == 'net48'">
+  <ItemGroup Condition="'$(TargetFrameworkIdentifier)' == '.NETFramework' And '$(TargetFrameworkVersion)' == 'v4.8'">
```
jamescrosswell:
```suggestion
  <ItemGroup Condition="'$(TargetFrameworkIdentifier)' == '.NETFramework'">
```

src/Sentry.Serilog/Sentry.Serilog.csproj
```diff
@@ -27,11 +27,11 @@
     <Using Include="Serilog.Formatting" />
   </ItemGroup>
 
-  <ItemGroup Condition="$(TargetFramework) == 'net462' or $(TargetFramework) == 'netstandard2.0'">
+  <ItemGroup Condition="'$(TargetFrameworkIdentifier)' == '.NETFramework' Or ('$(TargetFrameworkIdentifier)' == '.NETStandard' And '$(TargetFrameworkVersion)' == 'v2.0')">
     <PackageReference Include="Serilog" Version="2.7.1" />
   </ItemGroup>
 
-  <ItemGroup Condition="$(TargetFramework) != 'net462' and $(TargetFramework) != 'netstandard2.0'">
+  <ItemGroup Condition="'$(TargetFrameworkIdentifier)' != '.NETFramework' And ('$(TargetFrameworkIdentifier)' != '.NETStandard' Or '$(TargetFrameworkVersion)' != 'v2.0')">
```
Flash0ver: question: logical change

Shouldn't this be
```diff
-'$(TargetFrameworkIdentifier)' != '.NETFramework' And ('$(TargetFrameworkIdentifier)' != '.NETStandard' Or '$(TargetFrameworkVersion)' != 'v2.0')
+'$(TargetFrameworkIdentifier)' != '.NETFramework' And !('$(TargetFrameworkIdentifier)' == '.NETStandard' And '$(TargetFrameworkVersion)' == 'v2.0')
```
to be logically equivalent to what it was before?

If I'm reading this right:
- before: if not `.NET Framework` AND not `.NET Standard 2.0`
- after: if not `.NET Framework` AND (either not `.NET Standard` or not `2.0`)

src/Sentry/Sentry.csproj
```diff
@@ -2,7 +2,7 @@
 
   <PropertyGroup>
     <Description>Official SDK for Sentry - Open-source error tracking that helps developers monitor and fix crashes in real time.</Description>
-    <NoWarn Condition="'$(TargetFramework)' == 'netstandard2.0'">$(NoWarn);RS0017</NoWarn>
+    <NoWarn Condition="'$(TargetFrameworkIdentifier)' == '.NETStandard' And '$(TargetFrameworkVersion)' == 'v2.0'">$(NoWarn);RS0017</NoWarn>
```
Flash0ver: thought: readability / consistency

Although I find the previous version more readable ... I do prefer consistency ... so all good.

test/Sentry.Maui.Device.TestApp/Sentry.Maui.Device.TestApp.csproj
```diff
@@ -51,7 +51,7 @@
     <!-- This allows us to use NSubstitute on iOS when targeting net8.0-ios or greater
         https://learn.microsoft.com/en-us/dotnet/maui/macios/interpreter?view=net-maui-8.0
     -->
-    <UseInterpreter Condition="$(TargetFramework.Contains('-ios')) and '$(Configuration)' == 'Release'">true</UseInterpreter>
+    <UseInterpreter Condition="'$(TargetPlatformIdentifier)' == 'ios' And '$(Configuration)' == 'Release'">true</UseInterpreter>
```
Flash0ver, line 54: question: potential confusion

I got a bit confused by `TargetPlatformIdentifier`:
- we use it here already
- but we actually declare it in the `PropertyGroup` below (line 63)

I then realized, that `TargetPlatformIdentifier` is already declared by MSBuild, and we just overwrite it below in line 63.

So the usage here seems correct.
But I'm wondering if we should prefix the `TargetPlatformIdentifier` that we declare below in line 63 with an underscore to `_TargetPlatformIdentifier`?

Or is this overwrite intentional?
Perhaps to fix some corner cases?

src/Sentry/Sentry.csproj
Flash0ver: issue: does this work?

I have the feeling there is an issue: Are `TargetFrameworkIdentifier` and `TargetFrameworkVersion` already available here? Was `TargetFramework` even available before? Did this ever work?

`RS0017` is from [Microsoft.CodeAnalysis.PublicApiAnalyzers](https://github.com/dotnet/roslyn/blob/main/src/RoslynAnalyzers/PublicApiAnalyzers/Microsoft.CodeAnalysis.PublicApiAnalyzers.md).
But we currently don't use `Microsoft.CodeAnalysis.PublicApiAnalyzers` ... we verify our APIs via our `ApiApprovalTests` using `PublicApiGenerator` and `Verify.Xunit`.

So I wonder if we could just remove this line, since I don't think it works, and we don't report `RS0017` anyway.

src/Sentry/Sentry.csproj
Flash0ver, line 1: I believe there are some more occurrences that we could replace:

- https://github.com/getsentry/sentry-dotnet/blob/844cc745d695d2ef164b765bd4ca85b03971a9fe/Directory.Build.props#L109
- https://github.com/getsentry/sentry-dotnet/blob/6ee1a4aacd0e14fd3e52e22ba1fb607235c37d11/samples/Sentry.Samples.Console.Basic/Sentry.Samples.Console.Basic.csproj#L9
- https://github.com/getsentry/sentry-dotnet/blob/4080ac6b894ff263da6e01805bc91df1b3f84f2e/src/Directory.Build.targets#L6
- https://github.com/getsentry/sentry-dotnet/blob/6d1bbf3c9face7ea742ff6a2f9ec6b75ca3c5bf6/src/Sentry.Bindings.Android/Sentry.Bindings.Android.csproj#L78
- https://github.com/getsentry/sentry-dotnet/blob/6d1bbf3c9face7ea742ff6a2f9ec6b75ca3c5bf6/src/Sentry.Bindings.Android/Sentry.Bindings.Android.csproj#L93
- https://github.com/getsentry/sentry-dotnet/blob/545ac315983b60328b7f37b21924abc9709b4ba8/src/Sentry/buildTransitive/Sentry.targets#L169-L173
- https://github.com/getsentry/sentry-dotnet/blob/7f5e8def60893b53ed41aefdd0c8a1b5b25f17a3/src/Sentry/Sentry.csproj#L45
- https://github.com/getsentry/sentry-dotnet/blob/7f5e8def60893b53ed41aefdd0c8a1b5b25f17a3/src/Sentry/Sentry.csproj#L54
- https://github.com/getsentry/sentry-dotnet/blob/7f5e8def60893b53ed41aefdd0c8a1b5b25f17a3/src/Sentry/Sentry.csproj#L57
- https://github.com/getsentry/sentry-dotnet/blob/7f5e8def60893b53ed41aefdd0c8a1b5b25f17a3/src/Sentry/Sentry.csproj#L75
- https://github.com/getsentry/sentry-dotnet/blob/7f5e8def60893b53ed41aefdd0c8a1b5b25f17a3/src/Sentry/Sentry.csproj#L80
- https://github.com/getsentry/sentry-dotnet/blob/7f5e8def60893b53ed41aefdd0c8a1b5b25f17a3/src/Sentry/Sentry.csproj#L85
- https://github.com/getsentry/sentry-dotnet/blob/1de31a900472218bcbd4744020d460470b2f13bd/test/Directory.Build.props#L77
- https://github.com/getsentry/sentry-dotnet/blob/65363ca3f18c0c57dfc062a0e0833dc9eeb8efab/test/Sentry.Tests/Sentry.Tests.csproj#L50
- https://github.com/getsentry/sentry-dotnet/blob/14ec84fb8524e7022109b590d691c1f0c59c1c6f/test/Sentry.Testing/Sentry.Testing.csproj#L25
- https://github.com/getsentry/sentry-dotnet/blob/65363ca3f18c0c57dfc062a0e0833dc9eeb8efab/test/Sentry.Serilog.Tests/Sentry.Serilog.Tests.csproj#L23

Should we convert these too?
And if - should we do it in a follow-up?

Flash0ver (commented):
Found some more potential occurrences we could replace ... and a potential problem.

test/Sentry.Maui.Device.TestApp/Sentry.Maui.Device.TestApp.csproj
Flash0ver: update: if I'm reading the history right ... this overwrite is actually intentional due to some ordering issues that we ran into ... do I get that right?

src/Sentry.Serilog/Sentry.Serilog.csproj
jamescrosswell: So the two expressions are logically equivalent but yours is easier to read... will implement that instead.

src/Sentry/Sentry.csproj
jamescrosswell: > Did this ever work?

- Yes, originally (2019, PR #234) it suppressed RS0017 from PublicApiAnalyzers.
- nothing has emitted RS0017 since PR #1313 removed the analyzer in Nov 2021

So yeah, we can remove. Nice catch!

test/Sentry.Maui.Device.TestApp/Sentry.Maui.Device.TestApp.csproj
jamescrosswell: A bit of digital archaeology required to track this down but yes, intentional:

- PR #2002 — Matt introduced the props-and-targets dual-definition along with an explanatory comment.
- PR #2241 — Matt added the local recompute in this device-test csproj, so the RuntimeIdentifier conditions resolve correctly (workaround for xamarin-macios#17841).
- github.com/jamescrosswell/msbuild-test-4275 — I put together a minimal repro repo to pin down the exact ordering behaviour.

I have no recollection of any of this - this is just from GIT.

src/Sentry/Sentry.csproj
jamescrosswell: Good call — I've converted all of these in this PR rather than a follow-up. Two of them needed more than a find-and-replace, so flagging the reasoning:

**1. The `<PropertyGroup>` conditions needed `TargetFrameworkIdentifier` to be available early.**

`$(TargetFrameworkIdentifier)` / `$(TargetFrameworkVersion)` are only set *late* by the SDK (in its own targets), so they're empty during early `<PropertyGroup>` evaluation — they're only reliably populated in time for `<ItemGroup>` conditions. Converting the property-group conditions naively made them silently mis-evaluate: it actually turned `PublishAot` **on for net462** in the Console.Basic sample (`NETSDK1207`). Same trap would hit `PlatformIsLegacy` in `Directory.Build.props`, `Nullable` in `src/Directory.Build.targets`, and the `HAS_DIAGNOSTIC_INTEGRATION` group in `Sentry.csproj`.

Fix: pre-compute both in `Directory.Build.props`, right next to the existing `TargetPlatformIdentifier` computation, exactly as the SDK does (including the `v` prefix on the version):

```xml
<TargetFrameworkIdentifier>$([MSBuild]::GetTargetFrameworkIdentifier('$(TargetFramework)'))</TargetFrameworkIdentifier>
<TargetFrameworkVersion>v$([MSBuild]::GetTargetFrameworkVersion('$(TargetFramework)'))</TargetFrameworkVersion>
```

With that, the property-based syntax works consistently in both `<PropertyGroup>` and `<ItemGroup>` conditions — no `.StartsWith` exceptions left anywhere.

**2. The native-AOT block in `buildTransitive/Sentry.targets` → the existing `_SentryIsNet8OrGreater` helper.**

Rather than convert `!net4 && !netstandard2 && !net6 && !net7` literally, I switched it to `'$(_SentryIsNet8OrGreater)' == 'true'` — already defined in `Sentry.props` and used by the sync-twin `Sentry.Native.targets`. It's clearer, keeps the two files genuinely in sync, and is *more* correct: the old list let `net5` through, whereas `_SentryIsNet8OrGreater` excludes net5/6/7 properly. (We can't use `PlatformIsLegacy` here — it's defined in our repo `Directory.Build.props`, which isn't shipped in the package.)

I also dropped the incidental .NET Framework version pins (the specific version was never required), while keeping the netstandard `2.0`-vs-`2.1` checks where the version genuinely matters (Serilog package versions, nullability).

Build-verified across net462 / netstandard2.0 / netstandard2.1 / net9 / net10. The AOT `buildTransitive` change is consumer-side so it isn't exercised by our own build — the Native-AOT integration tests should cover it.

src/Sentry.Extensions.Logging/Sentry.Extensions.Logging.csproj
```diff
@@ -15,24 +15,24 @@
     <ProjectReference Include="..\..\src\Sentry\Sentry.csproj" />
   </ItemGroup>
 
-  <ItemGroup Condition="'$(TargetFramework)' == 'netstandard2.0'">
+  <ItemGroup Condition="'$(TargetFrameworkIdentifier)' == '.NETStandard'">
     <PackageReference Include="Microsoft.Extensions.Logging.Configuration" Version="2.1.0" />
     <PackageReference Include="Microsoft.Extensions.Http" Version="2.1.0" />
   </ItemGroup>
```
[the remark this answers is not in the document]
jamescrosswell: Never going to happen. .NET Standard is basically legacy at this point.

src/Sentry.Serilog/Sentry.Serilog.csproj
Flash0ver: oh ... right ... "not standard or not v2" is only `false` for .NET Standard 2.0 ... that's a little bit embarrassing 😔

test/Sentry.Maui.Device.TestApp/Sentry.Maui.Device.TestApp.csproj
Flash0ver: thanks

Flash0ver (approved):
What a changeset ... contender for the most inclusions of `".NET"` 😉

I got one final question:

src/Sentry/Sentry.csproj
Flash0ver: Thanks for the elaboration 🤘

test/Sentry.Maui.Device.TestApp/Sentry.Maui.Device.TestApp.csproj
Flash0ver, line 54: question: order for `TargetPlatformIdentifier`

I asked this indirectly already ... but now more explicitly:

We ensure that `TargetPlatformIdentifier` is set in the PropertyGroup below in line 63.

Can it be that `TargetPlatformIdentifier` is not yet available here?
But then, don't we already set it in `/Directory.Build.props`?
- https://github.com/getsentry/sentry-dotnet/blob/06abd70e656d7b051443b7178746c4a2d9210893/Directory.Build.props#L64-L73

Where we ensure that the triplet of `TargetPlatformIdentifier`, `TargetFrameworkIdentifier`, and `TargetFrameworkVersion` are set and available.

So could we actually remove - from this `.csproj` - line 63, where we set `TargetPlatformIdentifier` potentially "again"?

On the other hand - even if - it does not hurt to keep it, to be fair.

jamescrosswell: Correct on both counts.

`TargetPlatformIdentifier` **is** available here at line 54: this project imports the root `Directory.Build.props` (via `test/Sentry.Maui.Device.TestApp/Directory.Build.props`), which computes the identifier/version triplet early — before the project body — so the `UseInterpreter` condition evaluates correctly. And that's exactly what makes the line-63 assignment redundant: it recomputed the same value root had already set. I've removed it (keeping `OSArchitecture` and the `RuntimeIdentifier` conditions, which still use `TargetPlatformIdentifier` — now sourced from root).

Flash0ver: ty

PR #5241, opened 2026-05-18
feat: add TagList overload to Metrics-Emit APIs as Attributes parameter

### LINKED ISSUE #5259: feat: add TagList overload to Metrics-Emit APIs as Attributes parameter
### Summary

Add overloads to `Sentry.SentrySdk.Metrics.Emit*` methods with a by-ref `System.Diagnostics.TagList` parameter for **Sentry Attributes**.

### Remarks

Added in .NET 6.0, and via the out-of-band NuGet package [System.Diagnostics.DiagnosticSource](https://www.nuget.org/packages/System.Diagnostics.DiagnosticSource/), the [System.Diagnostics.TagList](https://learn.microsoft.com/dotnet/api/system.diagnostics.taglist) is a collection of `KeyValuePair<string, object?>` that stores up to 8 (eight) tags within this `struct`, and only allocates contiguous regions of memory on the heap when representing more than 8 (eight) tags (starting with an array of 16 elements, growing again by eight when exceeded).
The `TagList` is used widely across the `System.Diagnostics.Metrics` APIs.

We can provide an overload to the `EmitCounter`, `EmitGauge` and `EmitDistribution` method groups so that callers can make use of this type, too, by representing [Sentry Attributes](https://develop.sentry.dev/sdk/foundations/state-management/scopes/attributes/).

A common scenario is when collecting/exporting `Meter`-based Metrics via the [System.Diagnostics.Metrics.MeterListener](https://learn.microsoft.com/dotnet/api/system.diagnostics.metrics.meterlistener), that we are currently prototyping.

We should pass this `struct` `ByRef` rather than `ByVal`, as it's a larger struct. Preferred in a `readonly` fashion, where `in` might be the best fit (because an _rvalue_ also makes sense semantically since we don't necessarily require a location). But make sure to only invoke `readonly` members of that `struct`, as it's not a `readonly struct`.

Also, consume via a `for` loop rather than a `foreach` loop, because the implicit `GetEnumerator` implementation does not return the `struct-based` `TagList.Enumerator`, but just the `IEnumerator<KeyValuePair<string, object?>>` interface, which is allocating a box for the `struct` on the heap on runtimes that cannot de-virtualize the call via dynamic PGO.
See blog [Performance Improvements in .NET 10](https://devblogs.microsoft.com/dotnet/performance-improvements-in-net-10/).

### Potential follow-up
Propose to `dotnet/runtime` to expose the `internal` `Tags` property, returning `ReadOnlySpan<KeyValuePair<string, object?>>`, via an unsafe Marshal type, similar to `System.Runtime.InteropServices.AsSpan<>(System.Collections.Generic.List<T>?)`, highlighting our usage example where we could forward to our already existing internal `KeyValuePair<string, object>` overloads, rather than providing a whole new code path for `in TagList`.

Flash0ver:
fixes #5259

### Summary

Add an overload to all Metrics method groups (`EmitCounter`, `EmitGauge`, `EmitDistribution`) that accept _Attributes_, alongside the existing `ReadOnlySpan<>` and `IEnumerable<>` overloads.

### Remarks

TODOs
- [ ] Move "EditorBrowsable Never" Attributes to separate PR
- [ ] Verify that we don't run into problems when adding conditionally-compiled non-public `abstract` methods to a type that is non-constructible by user code, e.g. when having a `netstandard2.0` library with a `PackageReference` to `Sentry` that is referenced by a `net8.0` application that also has a `PackageReference` to `Sentry`
- [ ] Explain why `for` over `foreach` (only .NET 10+ can devirtualize to `TagList.Enumerator`

### DIFF
--- a/benchmarks/Sentry.Benchmarks/BenchmarkDotNet.Artifacts/results/Sentry.Benchmarks.SentryMetricEmitterBenchmarks-report-github.md
+++ b/benchmarks/Sentry.Benchmarks/BenchmarkDotNet.Artifacts/results/Sentry.Benchmarks.SentryMetricEmitterBenchmarks-report-github.md
@@ -1,0 +1,16 @@
+```
+
+BenchmarkDotNet v0.13.12, macOS 26.4.1 (25E253) [Darwin 25.4.0]
+Apple M3 Pro, 1 CPU, 12 logical and 12 physical cores
+.NET SDK 10.0.203
+  [Host]     : .NET 9.0.8 (9.0.825.36511), Arm64 RyuJIT AdvSIMD
+  DefaultJob : .NET 9.0.8 (9.0.825.36511), Arm64 RyuJIT AdvSIMD
+
+
+```
+| Method                        | Mean      | Error    | StdDev   | Gen0   | Gen1   | Allocated |
+|------------------------------ |----------:|---------:|---------:|-------:|-------:|----------:|
+| EmitWithoutAttributes         |  99.82 ns | 1.894 ns | 1.945 ns | 0.0640 | 0.0001 |     536 B |
+| EmitWithAttributes_Enumerable | 148.19 ns | 0.544 ns | 0.454 ns | 0.0851 |      - |     712 B |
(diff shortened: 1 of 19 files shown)


jamescrosswell:
On hold... we'll pick this up later when we have the bandwidth.

PR #5293, opened 2026-06-16
fix: Unsampled Transactions no longer propagate empty trace headers

Flash0ver:
This is a follow-up to
- #4302

#4302 applied the fix to `Sentry.AspNetCore` only.

### Changes
This changeset is
- porting the fix to `Sentry.AspNet` as well
- adding an additional Test-Case to `Sentry.OpenTelemetry`/`Sentry.OpenTelemetry.Exporter` for existing related behavior that is intentionally left unchanged

### DIFF
--- a/test/Sentry.OpenTelemetry.Exporter.Tests/SentryPropagatorTests.cs
+++ b/test/Sentry.OpenTelemetry.Exporter.Tests/SentryPropagatorTests.cs
@@ -169,4 +169,30 @@ public class SentryPropagatorTests
             outContext.Baggage.GetBaggage("sentry-sample_rate").Should().Be("1.0");
         }
     }
+
+    [Fact]
+    public void Extract_WithEmptyTraceId_From_SentryTraceHeader()
+    {
+        // Arrange
+        var carrier = new Dictionary<string, string>()
+        {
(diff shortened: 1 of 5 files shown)


test/Sentry.OpenTelemetry.Exporter.Tests/SentryPropagatorTests.cs
```diff
@@ -169,4 +169,30 @@ public void Extract_PropagationContext_From_Carrier()
...
     }
+
+    [Fact]
+    public void Extract_WithEmptyTraceId_From_SentryTraceHeader()
+    {
+        // Arrange
+        var carrier = new Dictionary<string, string>()
+        {
+            { "sentry-trace", "00000000000000000000000000000000-1000000000000000-1" }
+        };
```
Flash0ver, line 180: question: is this a valid test case?

This test fails on Windows on `net48`.
Is this a valid test case to add (no changes in related production code)?

jamescrosswell: Would need to do some digging to confirm, but my gut tells me no.... I don't think we propagate empty trace ids (I seem to recall these get filtered or handled specially when parsing the incoming trace header).

PR #5294, opened 2026-06-17
fix: don't attach `sentry-trace` header with empty Trace ID

Flash0ver:
### Summary

Don't attach a new `sentry-trace` header when the "trace-id" is empty (all `0`).

### Remarks

A user is having an issue in a distributed setup instrumented with multiple Sentry-SDKs where a tail-SDK is emitting Metrics with an empty Trace ID (`SentryId.Empty` / `00000000000000000000000000000000`), propagated via the `sentry-trace` header from a head-SDK.

### Update

This less-of-a-fix and more-of-a-workaround/hack might have become obsolete since we found the root cause:
- getsentry/sentry-electron#1389


Flash0ver:
`sentry-dotnet` was not the "root" SDK emitting an empty `sentry-trace` header, but just propagating it.

PR #5297, opened 2026-06-18
docs: add XML comments for boolean return of ILogEntryFilter

### LINKED ISSUE #5296: docs: Indicate which `bool` return value of `AddLogEntryFilter` actually excludes a message
### Description

See in title; right now the XML docs don't actually tell you that "true" excludes the value. We do have this indicated for IExceptionFilter but not for AddLogEntryFilter's lambdas

Flash0ver:
resolves #5296

### Summary

Add missing documentation comments for the `bool` return value of `Sentry.Extensions.Logging` Filter-Option.

### Remarks

Also fixes a typo.

### Changelog Entry

docs: add XML comments for boolean return of `Sentry.Extensions.Logging` Filters

### DIFF
--- a/src/Sentry.Extensions.Logging/DelegateLogEntryFilter.cs
+++ b/src/Sentry.Extensions.Logging/DelegateLogEntryFilter.cs
@@ -13,7 +13,7 @@ public class DelegateLogEntryFilter : ILogEntryFilter
     /// <summary>
     /// Creates a new instance of <see cref="DelegateLogEntryFilter"/>
     /// </summary>
-    /// <param name="filter"></param>
+    /// <param name="filter">The filter <see langword="delegate"/>.</param>
     public DelegateLogEntryFilter(Func<string, LogLevel, EventId, Exception?, bool> filter)
     {
         ArgumentNullException.ThrowIfNull(filter);
--- a/src/Sentry.Extensions.Logging/SentryLoggingOptionsExtensions.cs
+++ b/src/Sentry.Extensions.Logging/SentryLoggingOptionsExtensions.cs
@@ -9,26 +9,28 @@ namespace Sentry.Extensions.Logging;
 public static class SentryLoggingOptionsExtensions
 {
     /// <summary>
-    /// Add an log event filter
+    /// Add a log event filter.
     /// </summary>
     /// <remarks>
     /// Filters are called before sending an event.
     /// This allows the filter to decide whether the log message should not be recorded at all.
+    /// The log entry is neither captured as <see cref="SentryEvent"/> nor added as <see cref="Breadcrumb"/> when any filter returns <see langword="true"/>.
     /// </remarks>
     /// <param name="options">The <see cref="SentryLoggingOptions"/> to hold the filter.</param>
-    /// <param name="filter">The filter.</param>
+    /// <param name="filter">The <see cref="ILogEntryFilter"/> filter.</param>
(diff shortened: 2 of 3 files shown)


src/Sentry.Extensions.Logging/SentryLoggingOptionsExtensions.cs
```diff
@@ -9,26 +9,28 @@ namespace Sentry.Extensions.Logging;
...
+    /// Add a log event filter.
     /// </summary>
     /// <remarks>
     /// Filters are called before sending an event.
     /// This allows the filter to decide whether the log message should not be recorded at all.
+    /// The log entry is neither captured as <see cref="SentryEvent"/> nor added as <see cref="Breadcrumb"/> when any filter returns <see langword="true"/>.
     /// </remarks>
     /// <param name="options">The <see cref="SentryLoggingOptions"/> to hold the filter.</param>
-    /// <param name="filter">The filter.</param>
+    /// <param name="filter">The filter <see langword="delegate"/>. Return <see langword="true"/> if the log entry should be filtered.</param>
```
Flash0ver: suggestion: https://github.com/getsentry/sentry-dotnet/issues/5296#issuecomment-4742250008

```suggestion
    /// <param name="filter">The filter <see langword="delegate"/>. Return <see langword="true"/> if the log entry should be filtered out.</param>
```

Flash0ver (commented):
question: do we want to mention XML-doc only changes in the `CHANGELOG.md`?

I could imagine that we don't.

src/Sentry.Extensions.Logging/DelegateLogEntryFilter.cs
```diff
@@ -13,7 +13,7 @@ public class DelegateLogEntryFilter : ILogEntryFilter
     /// <summary>
     /// Creates a new instance of <see cref="DelegateLogEntryFilter"/>
     /// </summary>
-    /// <param name="filter"></param>
+    /// <param name="filter">The filter <see langword="delegate"/>.</param>
```
Flash0ver, line 16: question: do we want to mention XML-doc only changes in the CHANGELOG.md?

If not - which I can imagine - we should remove the `### Changelog Entry` from the PR description, and also add the `skip-changelog` label.

Flash0ver: this is more of a general question, I realize:

the documentation of Craft mentions:
https://craft.sentry.dev/configuration/#changelog-policies

image

But we didn't do this before ... so I'm leaning towards not doing it now either.

src/Sentry.Extensions.Logging/SentryLoggingOptionsExtensions.cs
```diff
@@ -9,26 +9,28 @@ namespace Sentry.Extensions.Logging;
 public static class SentryLoggingOptionsExtensions
 {
     /// <summary>
-    /// Add an log event filter
+    /// Add a log event filter.
     /// </summary>
     /// <remarks>
     /// Filters are called before sending an event.
     /// This allows the filter to decide whether the log message should not be recorded at all.
+    /// The log entry is neither captured as <see cref="SentryEvent"/> nor added as <see cref="Breadcrumb"/> when any filter returns <see langword="true"/>.
```
jamescrosswell:
```suggestion
    /// The log entry is neither captured as a <see cref="SentryEvent"/> nor added as a <see cref="Breadcrumb"/> when any filter returns <see langword="true"/>.
```

jamescrosswell (approved):
LGTM

Issue #3481, opened 2024-07-15
Support Spotlight without sending to Sentry


jessetan:
### Problem Statement

One of the first lines on the [Spotlight about page](https://spotlightjs.com/about/) is
> One of the great benefits of this approach is you get the power of Sentry’s telemetry without needing to utilize Sentry’s production monitoring.

The [blogpost about the dotnet v4 SDK](https://blog.sentry.io/sentry-dotnet-sdk-4-for-dotnet-8/#spotlight---sentry-for-development) also says
> It’s a tool to help you see all of Sentry’s telemetry data directly in your development environment, without the need to send anything to Sentry itself.

However this does not seem to be possible, since the [SDK will not initialise Sentry without a valid DSN](https://github.com/getsentry/sentry-dotnet/blob/b07c089d9d454a0208ee0849580e787503ad2794/src/Sentry/SentrySdk.cs#L41) and wil not send anything to Spotlight.

### Solution Brainstorm

Requested: when `EnableSpotlight = true` is passed to `SentrySdk.Init`, Sentry is always initialized even if there is no DSN.
Optional: when Spotlight is enabled, ignore transaction sample rate setting and send everything to Sentry, while still using the configured rate to send to Sentry.

I did not test if the requested behaviour is present in other SDKs, except for the in-browser JS SDK. That one requires a separate `Spotlight.init()` call outside of `Sentry.init` and does work without a DSN. For the other SDKs, Spotlight is turned on inside `Sentry.init` so might also ignore Spotlight if there is no DSN.

bitsandfoxes:
Thanks for the feedback! I'll have to see what the other SDKs are up to.
> It’s a tool to help you see all of Sentry’s telemetry data directly in your development environment, without the need to send anything to Sentry itself.

I can see the being read in a way that you would not expect the events to have to go to Sentry.

bitsandfoxes:
From what it looks like we should update the SDK to check for `Is there a DSN || EnableSpotlight` before initializing.

vaind:
Does spotlight currently work for native events (crashes)? If so, that would also need to be taken into account.

bitsandfoxes:
AFAIK, it currently does not, no. Oh no, that adds another layer..

xt0rted:
It took me a bit to set Spotlight up in our .net projects due to the different DSN behavior compared to the JS sdk I set up first. It was also a huge surprise today to realize everything our .net code has logged to Spotlight is now also in Sentry 😕

[...]

PR #5302, opened 2026-06-23
feat: Spotlight spec compliance - DSN-less mode, pipeline bypass, env var support

mattico:
A follow-up to #5025, implements the remaining parts of #3481. Endeavors to implement the remaining behaviors mentioned in the Spotlight integration spec: functioning in DSN-less mode when Spotlight is enabled, having Spotlight envelopes bypass the processing pipeline, and the specification for env/config behavior.

As [mentioned previously](https://github.com/getsentry/sentry-dotnet/issues/3481#issuecomment-4040094663) the Spotlight integration spec for the config is a bit at odds with the existing SentryOptions interface - this could be cleaned up at the next semver break but for now it maintains API compatibility at the cost of slight mismatch between interface and semantics.

I tried to make the pipeline changes as un-invasive as possible but a fair amount of refactoring was required because the integration point for Spotlight moved much earlier in order to implement the processing-bypass behavior.

There is some tricky stuff related to envelope serialization that I tried to test thoroughly but could take a careful review.

---

I probably won't have time to iterate on this PR for several weeks but I didn't want this code to languish on my hard drive any longer.

### DIFF
--- a/src/Sentry/BindableSentryOptions.cs
+++ b/src/Sentry/BindableSentryOptions.cs
@@ -111,8 +111,16 @@ internal partial class BindableSentryOptions
         options.UseAsyncFileIO = UseAsyncFileIO ?? options.UseAsyncFileIO;
         options.DisableSentryHttpMessageHandler = DisableSentryHttpMessageHandler ?? options.DisableSentryHttpMessageHandler;
         options.JsonPreserveReferences = JsonPreserveReferences ?? options.JsonPreserveReferences;
-        options.EnableSpotlight = EnableSpotlight ?? options.EnableSpotlight;
-        options.SpotlightUrl = SpotlightUrl ?? options.SpotlightUrl;
+        // Deliberately not the usual `?? options.X` pattern: only assign when bound, to avoid marking
+        // these as explicitly set (which drives env-var precedence in SettingLocator.ResolveSpotlight).
+        if (EnableSpotlight.HasValue)
+        {
+            options.EnableSpotlight = EnableSpotlight.Value;
+        }
+        if (SpotlightUrl is not null)
+        {
+            options.SpotlightUrl = SpotlightUrl;
+        }
 
 #if ANDROID
         Android.ApplyTo(options.Android);
--- a/src/Sentry/Internal/Hub.cs
+++ b/src/Sentry/Internal/Hub.cs
@@ -52,9 +52,9 @@ internal class Hub : IHub, IDisposable
         ISampleRandHelper? sampleRandHelper = null,
         BackpressureMonitor? backpressureMonitor = null)
     {
-        if (string.IsNullOrWhiteSpace(options.Dsn))
+        if (string.IsNullOrWhiteSpace(options.Dsn) && !options.EnableSpotlight)
         {
-            const string msg = "Attempt to instantiate a Hub without a DSN.";
+            const string msg = "Attempt to instantiate a Hub without a DSN or Spotlight enabled.";
             options.LogFatal(msg);
             throw new InvalidOperationException(msg);
         }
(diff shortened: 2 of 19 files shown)


src/Sentry/Internal/Hub.cs
```diff
@@ -330,7 +347,10 @@ public BaggageHeader GetBaggage()
         }
 
         var propagationContext = CurrentScope.PropagationContext;
-        return propagationContext.GetOrCreateDynamicSamplingContext(_options, _replaySession).ToBaggageHeader();
+        // GetOrCreateDynamicSamplingContext returns null when there is no DSN (e.g. Spotlight-only mode);
```
jamescrosswell, line 350: This implies trace propagation will be broken. That may be unavoidable but worth noting that there will be capability degradation in the SDK when using spotlight only - Sentry's distributed tracing won't work.

src/Sentry/Internal/Hub.cs
```diff
@@ -248,6 +248,23 @@ internal ITransactionTracer StartTransaction(
 
         if (isSampled is false)
         {
+            if (_options.EnableSpotlight)
+            {
+                // When Spotlight is enabled, use a full TransactionTracer even for sampled-out transactions
```
jamescrosswell, line 253: Is this per some documentation on how this should be implemented or an arbitraty design decision? It definitely adds complexity since whether a full transaction tracer is present is now determined by an extra variable (that we may or may not have test cases for in other unrelated logic in the solution). 

This additional complexity also seems like throw away code since the plan is to phase out transactions entirely and replace them with a streaming 'span only' performance tracing solution.

I'd be reluctant to pour time into trying to get this right (and fix any follow up bugs we don't catch in the initial review) if this is the case.

src/Sentry/Internal/SdkComposer.cs
```diff
@@ -26,30 +26,43 @@ private ITransport CreateTransport()
...
-                transport = CachingTransport.Create(transport, _options);
+                _options.LogDebug("No cache directory path specified. Skipping caching transport creation.");
             }
         }
         else
         {
-            _options.LogDebug("No cache directory path specified. Skipping caching transport creation.");
+            // No DSN — use a no-op transport (e.g. Spotlight-only mode).
+            _options.LogDebug("No DSN configured. Using no-op transport for Sentry.");
+            transport = NoOpTransport.Instance;
```
jamescrosswell, line 61: This seems like the guts of the solution - it's an interesting solution to the problem if we can work around some of the other side effects of not having a DSN (like distributed tracing not working). 

Possibly an alternative solution would be to have some 'reserved' DSN that our regular transport basically treated as the equivalent of `dev/null`... that way we would have a valid DSN that could be used for things like the creation of a `DynamicSamplingContext`.  I'm not sure if that's a better (or even good) idea - but maybe worth considering.

src/Sentry/BindableSentryOptions.cs
```diff
@@ -111,8 +111,16 @@ public void ApplyTo(SentryOptions options)
         options.UseAsyncFileIO = UseAsyncFileIO ?? options.UseAsyncFileIO;
         options.DisableSentryHttpMessageHandler = DisableSentryHttpMessageHandler ?? options.DisableSentryHttpMessageHandler;
         options.JsonPreserveReferences = JsonPreserveReferences ?? options.JsonPreserveReferences;
-        options.EnableSpotlight = EnableSpotlight ?? options.EnableSpotlight;
-        options.SpotlightUrl = SpotlightUrl ?? options.SpotlightUrl;
+        // Deliberately not the usual `?? options.X` pattern: only assign when bound, to avoid marking
```
jamescrosswell, line 114: Won't this be inconsistent with the way other settings get bound/resolved? What's the reason for this?

jamescrosswell (commented):
Thanks for the PR @mattico - it's a big one with lot's of potential gotchas and unintended consequences so I think it will need careful discussion, review and probably a few attempts to get right. 

I left some comments. The main one is the question I raised about the high level approach:
- Should we try to tweak the SDK do that it can be initialised without a DSN when Spotlight is enabled? This has lots of domino effects (some of which I'm worried won't be caught by our existing test suite because we never anticipated this possibility). 
- One alternative might be to have a dedicated 'Spotlight only' DSN that our regular transport recognised and simply discarded envelopes for... this might limit the change that needs to happen elsewhere in the codebase (where we'd previously been able to assume the presence of a DSN if the SDK was enabled). 

I think the AI bots made some useful observations as well... some of them only relevant if we decide to do this by allowing the SDK to be initialised without a DSN but others would need addressing regardless.

PR #5310, opened 2026-06-24
fix: OpenTelemetry transactions for failed requests keep their route name and otel context

### LINKED ISSUE #5091: `http.server` spans are incorrectly generated for non-2xx status codes when using OTel integration
### Package

Sentry.OpenTelemetry

### .NET Flavor

.NET Core

### .NET Version

10.0.104

### OS

Linux

### OS Version

Fedora 43 Workstation

### Development Environment

Rider 2025.x (Linux)

No

_No response_

### SDK Version

6.2.0

25.11.0

### Workload Versions

not applicable

```csharp
using OpenTelemetry.Trace;
using Sentry.OpenTelemetry;

WebApplicationBuilder builder = WebApplication.CreateBuilder(args);

builder.Services.AddControllers();

builder.WebHost.UseSentry(c =>
{
    c.DisableSentryHttpMessageHandler = true;
    c.UseOpenTelemetry();
});

builder.Services.AddOpenTelemetry()
    .WithTracing(t =>
    {
        t.AddAspNetCoreInstrumentation();
        t.AddSentry();
    });

WebApplication app = builder.Build();

app.MapControllers();

app.Run();
```

### Steps to Reproduce

1. Create 2 endpoinds:
    - one returns 200, call it `POST /GoodEndpoint`
    - second throws an exception, call it `POST /BadEndpoint`
2. Call them and examine traces in Sentry.

### Expected Result

Endpoints that throws an exception should generate `http.server` transaction with `POST /BadEndpoint`  name.
Failure rate for `POST /BadEndpoint` is 100%.

### Actual Result

Endpoints that throws an exception generate `http.server` transaction with `Microsoft.AspNetCore.Hosting.HttpRequestIn`  name.
Failure rate is `POST /BadEndpoint` is 0%, as all failures are recorded under `Microsoft.AspNetCore.Hosting.HttpRequestIn`.

This really prevents from analyzing what endpoint throws and what doesnt when looking at list of 100 endpoints. All of them have 0% failure rate, as all failures are collected under single `Microsoft.AspNetCore.Hosting.HttpRequestIn`.

jamescrosswell:
Closes #5091

## Summary

Fixes the issue where, with the OpenTelemetry integration, transactions for **failed** requests were reported under the raw activity name (e.g. `Microsoft.AspNetCore.Hosting.HttpRequestIn`) instead of the route (e.g. `GET /Test/BadEndpoint`).

## Problem

With OTel instrumentation, `SentrySpanProcessor` creates the transaction on Activity **start** and only applies the final name, operation and `otel`/`response` contexts on Activity **end** (parsed from `http.route` etc., which aren't available at start).

When a request threw an unhandled exception, `Hub.CaptureEvent` finished the scope's transaction as `Aborted` immediately after capturing the exception event — *before* the Activity ended — synchronously capturing/sending it. By the time `SentrySpanProcessor.OnEnd` ran, the transaction was already finished, so all the OTel enrichment (route-based name, `http.server` operation, `otel`/`response` contexts) was discarded.

This early-abort is correct for the native Sentry.AspNetCore tracing path (where the name is set up-front), but wrong when OTel owns the transaction lifecycle.

## Fix

Skip the early `Aborted` finish for OTel-instrumented transactions. The `SentrySpanProcessor` owns their lifecycle and finishes them with the correct name/operation/status when the Activity ends.

## Verification

Reproduced with the customer's [minimal repro](https://github.com/tvardero/sentry-bug-repro) on .NET 10.

`POST/GET BadEndpoint` (throws → 500):

| | Before | After |
|---|---|---|
| name | `Microsoft.AspNetCore.Hosting.HttpRequestIn` | `GET Test/BadEndpoint` |
| source | Custom | Route |
| operation | `Microsoft.AspNetCore.Hosting.HttpRequestIn` | `http.server` |
| status | Aborted | UnknownError (registers as a failure) |
| `otel`/`response` context | missing | present |

Added a regression test: `CaptureEvent_TerminalUnhandledException_DoesNotAbortOpenTelemetryTransaction`.

### DIFF
--- a/src/Sentry/Internal/Hub.cs
+++ b/src/Sentry/Internal/Hub.cs
@@ -667,8 +667,22 @@ internal class Hub : IHub, IDisposable
             {
                 // Event contains a terminal exception -> finish any current transaction as aborted
                 // Do this *after* the event was captured, so that the event is still linked to the transaction.
-                _options.LogDebug("Ending transaction as Aborted, due to unhandled exception.");
-                transaction.Finish(SpanStatus.Aborted);
+                // Exception: OpenTelemetry-instrumented transactions are owned and finished by the
+                // SentrySpanProcessor when the underlying Activity ends. That's also where the transaction
+                // name, operation and otel/response contexts get populated (from the http.route etc.
+                // attributes, which aren't available yet at this point). Finishing it early here would
+                // capture it before that enrichment, sending it with the raw activity name (e.g.
+                // "Microsoft.AspNetCore.Hosting.HttpRequestIn") and no otel context. See issue #5091.
+                if (transaction is IBaseTracer { IsOtelInstrumenter: true })
+                {
+                    _options.LogDebug(
+                        "Not ending OpenTelemetry transaction as Aborted; it is finished by the SentrySpanProcessor.");
+                }
+                else
+                {
+                    _options.LogDebug("Ending transaction as Aborted, due to unhandled exception.");
+                    transaction.Finish(SpanStatus.Aborted);
+                }
             }
 
             return id;
--- a/src/Sentry/Internal/Hub.cs
+++ b/src/Sentry/Internal/Hub.cs
@@ -667,12 +667,7 @@ internal class Hub : IHub, IDisposable
             {
                 // Event contains a terminal exception -> finish any current transaction as aborted
                 // Do this *after* the event was captured, so that the event is still linked to the transaction.
-                // Exception: OpenTelemetry-instrumented transactions are owned and finished by the
-                // SentrySpanProcessor when the underlying Activity ends. That's also where the transaction
-                // name, operation and otel/response contexts get populated (from the http.route etc.
-                // attributes, which aren't available yet at this point). Finishing it early here would
-                // capture it before that enrichment, sending it with the raw activity name (e.g.
-                // "Microsoft.AspNetCore.Hosting.HttpRequestIn") and no otel context. See issue #5091.
+                // Skip for OpenTelemetry transactions - these get handled by the SpanProcessor instead. See https://github.com/getsentry/sentry-dotnet/pull/5310
                 if (transaction is IBaseTracer { IsOtelInstrumenter: true })
                 {
                     _options.LogDebug(
--- a/test/Sentry.Tests/HubTests.cs
+++ b/test/Sentry.Tests/HubTests.cs
@@ -546,6 +546,36 @@ public partial class HubTests : IDisposable
     }
 
     [Fact]
+    public void CaptureEvent_TerminalUnhandledException_DoesNotAbortOpenTelemetryTransaction()
+    {
+        // OpenTelemetry-instrumented transactions are owned and finished by the SentrySpanProcessor
+        // when the underlying Activity ends (which is also where the transaction name, operation and
+        // otel/response contexts get populated). Finishing them early here would capture them with the
+        // raw activity name and no otel context. See https://github.com/getsentry/sentry-dotnet/issues/5091
+
+        // Arrange
+        _fixture.Options.TracesSampleRate = 1.0;
+        _fixture.Options.Instrumenter = Instrumenter.OpenTelemetry;
+        var hub = _fixture.GetSut();
+
+        var transactionContext = new TransactionContext("test", "operation")
+        {
+            Instrumenter = Instrumenter.OpenTelemetry
+        };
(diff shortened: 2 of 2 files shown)


src/Sentry/Internal/Hub.cs
```diff
@@ -667,8 +667,22 @@ private SentryId CaptureEvent(SentryEvent evt, SentryHint? hint, Scope scope)
...
                 // Event contains a terminal exception -> finish any current transaction as aborted
                 // Do this *after* the event was captured, so that the event is still linked to the transaction.
-                _options.LogDebug("Ending transaction as Aborted, due to unhandled exception.");
-                transaction.Finish(SpanStatus.Aborted);
+                // Exception: OpenTelemetry-instrumented transactions are owned and finished by the
+                // SentrySpanProcessor when the underlying Activity ends. That's also where the transaction
+                // name, operation and otel/response contexts get populated (from the http.route etc.
+                // attributes, which aren't available yet at this point). Finishing it early here would
+                // capture it before that enrichment, sending it with the raw activity name (e.g.
+                // "Microsoft.AspNetCore.Hosting.HttpRequestIn") and no otel context. See issue #5091.
```
```suggestion
                // Skip for OpenTelemetry transactions - these get handled by the SpanProcessor instead. See https://github.com/getsentry/sentry-dotnet/pull/5310
```

src/Sentry/Internal/Hub.cs
```diff
@@ -667,8 +667,17 @@ private SentryId CaptureEvent(SentryEvent evt, SentryHint? hint, Scope scope)
...
-                _options.LogDebug("Ending transaction as Aborted, due to unhandled exception.");
-                transaction.Finish(SpanStatus.Aborted);
+                // Skip for OpenTelemetry transactions - these get handled by the SpanProcessor instead. See https://github.com/getsentry/sentry-dotnet/pull/5310
+                if (transaction is IBaseTracer { IsOtelInstrumenter: true })
+                {
+                    _options.LogDebug(
+                        "Not ending OpenTelemetry transaction as Aborted; it is finished by the SentrySpanProcessor.");
+                }
+                else
+                {
```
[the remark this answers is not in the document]
jamescrosswell: OnEnd is already the only path that finishes OTel transactions — every successful request relies on it, not just failed ones, so this doesn't add a new dependency. The cases where it wouldn't fire are (1) the Activity being sampled out, where we deliberately don't send a transaction, and (2) a process-terminating crash, where the flush is unreliable regardless and the only thing we'd have sent is the malformed transaction this PR removes. 

For the ASP.NET Core path in the issue, the host always stops the activity, so OnEnd runs. A timeout fallback would just resurrect the early-finish bug.

src/Sentry/Internal/Hub.cs
```diff
@@ -667,8 +667,17 @@ private SentryId CaptureEvent(SentryEvent evt, SentryHint? hint, Scope scope)
             {
                 // Event contains a terminal exception -> finish any current transaction as aborted
                 // Do this *after* the event was captured, so that the event is still linked to the transaction.
-                _options.LogDebug("Ending transaction as Aborted, due to unhandled exception.");
-                transaction.Finish(SpanStatus.Aborted);
+                // Skip for OpenTelemetry transactions - these get handled by the SpanProcessor instead. See https://github.com/getsentry/sentry-dotnet/pull/5310
+                if (transaction is IBaseTracer { IsOtelInstrumenter: true })
```
Flash0ver, line 671: question: SpanProcessor vs OTLP Exporter

Does this only apply to `Sentry.OpenTelemetry`, and not to `Sentry.OpenTelemetry.Exporter`?
Doesn't the OTLP Exporter also set `SentryOptions.Instrumenter` to `Instrumenter.OpenTelemetry`?

Or am I confusing something here?

Flash0ver (commented):
I might have a misunderstanding:

src/Sentry/Internal/Hub.cs
Flash0ver: oh ... wait ... I do misunderstand

`Sentry.OpenTelemetry.Exporter` doesn't start/finish Transactions/Spans, but "only" extracts/injects Sentry's Trace- and Baggage-Headers in order to propagate context ... due to also setting `SentryOptions.DisableSentryTracing = true`, and the `NoOpTransaction` class does not implement the `IBaseTracer` interface.

Flash0ver (approved):

PR #5317, opened 2026-06-25
feat: Add SentryOptions.AutoSetScopeTransactions to auto-store transactions on the scope

### LINKED ISSUE #2399: Consider setting the transaction on the current scope automatically
### Problem Statement

When manually instrumenting a transaction, it's advised to add it to the current scope:

```csharp
```

This is in the docs: https://docs.sentry.io/platforms/dotnet/performance/instrumentation/custom-instrumentation/

And samples:

https://github.com/getsentry/sentry-dotnet/blob/a674815a4ed0fe807b0bec4205c4ea48c085a988/samples/Sentry.Samples.Console.Basic/Program.cs#L38-L40

Adding the transaction to the scope is required for methods like `SentrySDK.GetSpan()` to work - which is useful for then calling methods like `StartChild()` to instrument a specific portion of the app.  If the transaction is not set on the scope, then `GetSpan()` returns `null`.

It's very easy to forget to do this.  It would be better if it happened automatically when starting a transaction, instead of having to call a separate API.

### Solution Brainstorm

Consider when calling `StartTransaction`, if there's an active scope that doesn't already have a transaction set, that we could set it on the scope at that time.

Likewise, when a transaction is finished, if it is the one on the scope then we can clear it from the scope.

We should be careful that any existing code that is setting it manually doesn't break, by only setting it when null, and only clearing it if we were the one to set it.

### LINKED ISSUE #5313: Decide whether auto instrumented transactions in MAUI should set Scope.Transaction
### Description

See:
https://github.com/getsentry/sentry-dotnet/blob/34d849038e3e23868a12d56ea256c36b0a7f3a61/src/Sentry.Maui.CommunityToolkit.Mvvm/MauiCommunityToolkitMvvmEventsBinder.cs#L145-L153

The code comment contradicts the code implementation. Either we change/remove the comment or we change the code.

If the comment is correct then the code itself needs to be changed (seems likely but need to verify).

jamescrosswell:
Closes #2399
Closes #5313 

## What

This is a QOL improvement that adds an opt-in `SentryOptions.AutoSetScopeTransactions` option (defaults to `false`). When enabled, `StartTransaction` automatically stores the started transaction on the current scope, so APIs like `SentrySdk.GetSpan()` / `StartChild()` work without the user having to manually call `SentrySdk.ConfigureScope(scope => scope.Transaction = transaction)`.

Since `SentryOptions.AutoSetScopeTransactions` defaults to `false` this is opt-in and non-breaking. We could potentially change the default in future major releases though.

## Per-call override (internal)

`StartSpan`/`StartTransaction` gain an internal `bool? autoSetScopeTransaction` override that wins over the global option, so integrations can control this independently of the user's setting. The MAUI CommunityToolkit binder uses it to ensure we don't store the transaction on the (global, `AsyncLocal`-backed) scope even when a user enables the `AutoSetScopeTransactions` option globally.

> [!NOTE]
> This changes the MAUI CommunityToolkit binder to finally match its own comment - see #5313

### DIFF
--- a/src/Sentry/HubExtensions.cs
+++ b/src/Sentry/HubExtensions.cs
@@ -54,11 +54,36 @@ public static class HubExtensions
     /// <summary>
     /// Starts a span or transaction if a transaction is not already active on the scope
     /// </summary>
-    public static ISpan StartSpan(this IHub hub, string operation, string description)
+    public static ISpan StartSpan(this IHub hub, string operation, string description) =>
+        hub.StartSpan(operation, description, null);
+
+    /// <summary>
+    /// Starts a span or transaction if a transaction is not already active on the scope.
+    /// <paramref name="autoSetScopeTransaction"/> overrides <see cref="SentryOptions.AutoSetScopeTransactions"/> for a
+    /// newly started transaction (it has no effect when a child span is started off an existing transaction). This lets
+    /// integrations control whether a transaction is stored on the scope independently of the user's global setting -
+    /// e.g. the MAUI CommunityToolkit binder opts out because the scope is global and AsyncLocal-backed.
+    /// </summary>
+    internal static ISpan StartSpan(this IHub hub, string operation, string description, bool? autoSetScopeTransaction)
     {
-        return hub.GetTransaction() is { } transaction
-            ? transaction.StartChild(operation, description)
-            : hub.StartTransaction(operation, description); // this is actually in the wrong order but changing it may break other things
+        if (hub.GetTransaction() is { } transaction)
+        {
+            return transaction.StartChild(operation, description);
+        }
+
+        // No active transaction - start a new one. Note the operation/description argument order below is intentionally
+        // preserved from the original (arguably wrong) behaviour of StartSpan; changing it may break other things.
+        if (hub is Hub fullHub)
+        {
+            return fullHub.StartTransaction(
+                new TransactionContext(operation, description),
+                new Dictionary<string, object?>(),
+                null,
+                autoSetScopeTransaction);
+        }
+
+        // Fallback for non-Hub IHub implementations, which can't honour the override.
+        return hub.StartTransaction(operation, description);
     }
 
     /// <summary>
--- a/src/Sentry/HubExtensions.cs
+++ b/src/Sentry/HubExtensions.cs
@@ -83,6 +83,7 @@ public static class HubExtensions
         }
 
         // Fallback for non-Hub IHub implementations, which can't honour the override.
+        // In practice this means the DisabledHub or HubAdapter (only used for testing).
         return hub.StartTransaction(operation, description);
     }
 
--- a/src/Sentry/Internal/Hub.cs
+++ b/src/Sentry/Internal/Hub.cs
@@ -294,10 +294,6 @@ internal class Hub : IHub, IDisposable
             return;
         }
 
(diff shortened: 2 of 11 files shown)


src/Sentry/Internal/Hub.cs
```diff
@@ -280,9 +282,25 @@ internal ITransactionTracer StartTransaction(
...
+        // The per-call override (used by some integrations) takes precedence over the global option.
+        if (!(autoSetScopeTransaction ?? _options.AutoSetScopeTransactions))
+        {
+            return;
+        }
+
+        // Only set the transaction on the scope if there isn't already one set (manually by the user or by another
+        // integration). We never overwrite an existing transaction. There's no need to track whether we were the one
+        // to set it for clearing purposes: TransactionTracer.Finish calls Scope.ResetTransaction, which only clears
+        // the scope's transaction if it still references this exact instance.
```
```suggestion
```

src/Sentry/HubExtensions.cs
```diff
@@ -54,11 +54,36 @@ public static ITransactionTracer StartTransaction(
...
+        if (hub is Hub fullHub)
+        {
+            return fullHub.StartTransaction(
+                new TransactionContext(operation, description),
+                new Dictionary<string, object?>(),
+                null,
+                autoSetScopeTransaction);
+        }
+
+        // Fallback for non-Hub IHub implementations, which can't honour the override.
```
jamescrosswell, line 85:
```suggestion
        // Fallback for non-Hub IHub implementations, which can't honour the override.
        // In practice this means the DisabledHub or HubAdapter (only used for testing).
```

src/Sentry/SentryOptions.cs
```diff
@@ -103,6 +103,24 @@ public bool IsGlobalModeEnabled
...
+    /// <summary>
+    /// When enabled, starting a transaction (for example via <see cref="SentrySdk.StartTransaction(string, string)"/>)
+    /// automatically stores it on the current scope, provided the scope does not already have a transaction set. This
+    /// means APIs such as <see cref="SentrySdk.GetSpan"/> work without having to manually call
+    /// <c>SentrySdk.ConfigureScope(scope =&gt; scope.Transaction = transaction)</c>.
+    /// </summary>
+    /// <remarks>
+    /// Defaults to <c>false</c>. An automatically stored transaction is only ever cleared from the scope by the SDK if
+    /// the scope still references that exact transaction when it finishes, so a transaction set manually (or by another
+    /// integration) is never overwritten or cleared.
```
jamescrosswell:
```suggestion
    /// Defaults to <c>false</c>
```

bitsandfoxes:
I'm not entirely convinced of this option. This requires such a detailed mental model of the way the SDK handles Spans and Transactions that I don't think this would actually help an actual user. 
And on top there is the move to be span first that'd make it obsolete. And I know we've been talking about making transactions obsolete for over a year but now it's happening for real. 😅

jamescrosswell:
> And I know we've been talking about making transactions obsolete for over a year but now it's happening for real.

Probably more like 2-3 years but yes, it does feel like we're getting closer. 

I guess we close as won't do for now then and rely on the move to span first to resolve this. We may still require some notion of `Scope.ActiveSpan` in that world though as  we still have the problem of 'which span should be the parent of this node'... semi-solved in OTEL (`Activity.Current`) but not in the Sentry tracing APIs without the use of scope.

PR #5319, opened 2026-06-26
fix(otlp): don't configure an automatic OTel propagator

### LINKED ISSUE #5299: Disable automatic propagator for OTLP
[According to the docs](https://develop.sentry.dev/sdk/telemetry/traces/otlp/#integration-spec):

> The integration MUST NOT set up an automatic propagator. Cross-service trace propagation is handled by the propagateTraceparent setting or by user-configured OTel propagators. How to set up an OTel propagator is a documentation concern.

So [our current behaviour](https://github.com/getsentry/sentry-dotnet/blob/42f6ca61e9ba53c2374cc3758c7ccdf04e949cf7/src/Sentry.OpenTelemetry.Exporter/SentryTracerProviderBuilderExtensions.cs#L55) is inconsistent with the docs.

When using OTLP, presumably users can and should configure the propagators via OTEL (not Sentry):
- SDK users should call `Sdk.SetDefaultTextMapPropagator` to configure the OTEL propagation settings
-  We should stop calling that in our code:
    https://github.com/getsentry/sentry-dotnet/blob/42f6ca61e9ba53c2374cc3758c7ccdf04e949cf7/src/Sentry.OpenTelemetry.Exporter/SentryTracerProviderBuilderExtensions.cs#L56
- We should make the above clear in the docs for the OTLP integration

@ericsampson this is a bit different to what you were expecting... but it's also how all of our other OTLP integrations behave so I don't think we have much wiggle room on this one.

_Originally posted by @jamescrosswell in https://github.com/getsentry/sentry-dotnet/pull/5217#discussion_r3456008124_

### Documentation
Make sure that the docs are consistent when publishing this change:
- see https://github.com/getsentry/sentry-docs/pull/18441#discussion_r3463279950

jamescrosswell:
Closes #5299

## Summary

[The OTLP integration spec](https://develop.sentry.dev/sdk/telemetry/traces/otlp/#integration-spec) states:

> The integration MUST NOT set up an automatic propagator. Cross-service trace propagation is handled by the `propagateTraceparent` setting or by user-configured OTel propagators.

Our OTLP exporter was inconsistent with this: `AddSentryOtlpExporter` (and `UseOtlp`) called `Sdk.SetDefaultTextMapPropagator(new SentryPropagator())`, automatically configuring a global propagator. This brings us in line with the spec and with our other OTLP integrations.

## Changes

- Remove the `Sdk.SetDefaultTextMapPropagator(...)` call from `AddSentryOtlpExporter`.
- Remove the now-purposeless `defaultTextMapPropagator` parameter from both `AddSentryOtlpExporter` and `SentryOptionsExtensions.UseOtlp`.
- Update XML docs to point users at `PropagateTraceparent` / their own OTel propagator config.
- Drop the obsolete custom-propagator test; add a regression test asserting `AddSentryOtlpExporter` leaves `Propagators.DefaultTextMapPropagator` untouched.

> ⚠️ **Breaking API change.** The `defaultTextMapPropagator` parameter shipped in 6.5.0/6.6.0 and is removed here (source + binary breaking for anyone who passed it). This PR therefore targets `version7` rather than `main`. Note the classic (non-OTLP) `Sentry.OpenTelemetry` package is intentionally left unchanged — it still requires `SentryPropagator`.

Users who want cross-service propagation should now either set `SentryOptions.PropagateTraceparent` or configure OpenTelemetry propagators themselves (e.g. `Sdk.SetDefaultTextMapPropagator(...)`).

## Follow-up

- Docs need a matching update so the OTLP integration page explains propagator configuration — see getsentry/sentry-docs#18441 (discussion [r3463279950](https://github.com/getsentry/sentry-docs/pull/18441#discussion_r3463279950)).

### DIFF
--- a/src/Sentry.OpenTelemetry.Exporter/SentryOptionsExtensions.cs
+++ b/src/Sentry.OpenTelemetry.Exporter/SentryOptionsExtensions.cs
@@ -1,4 +1,3 @@
-using OpenTelemetry.Context.Propagation;
 using OpenTelemetry.Trace;
 using Sentry.OpenTelemetry.Exporter;
 
@@ -21,24 +20,19 @@ public static class SentryOptionsExtensions
     /// <param name="tracerProviderBuilder"><see cref="TracerProviderBuilder"/></param>
     /// <param name="collectorUrl">A custom endpoint to export OLTP trace information to. If no url is provided, the
     /// endpoint will be inferred automatically from the DSN.</param>
-    /// <param name="defaultTextMapPropagator">
-    ///     <para>The default TextMapPropagator to be used by OpenTelemetry.</para>
-    ///     <para>
-    ///         If this parameter is not supplied, the <see cref="OpenTelemetry.SentryPropagator"/> will be used, which propagates the
-    ///         baggage header as well as Sentry trace headers.
-    ///     </para>
-    ///     <para>
-    ///         The <see cref="OpenTelemetry.SentryPropagator"/> is required for Sentry's OpenTelemetry integration to work, but you
-    ///         could wrap this in a <see cref="CompositeTextMapPropagator"/> if you needed other propagators as well.
-    ///     </para>
-    /// </param>
-    public static void UseOtlp(this SentryOptions options, TracerProviderBuilder tracerProviderBuilder, Uri? collectorUrl = null, TextMapPropagator? defaultTextMapPropagator = null)
+    /// <remarks>
+    /// In line with the OTLP integration specification, this method does not configure an OpenTelemetry propagator.
+    /// Cross-service trace propagation should be enabled either via the <see cref="SentryOptions.PropagateTraceparent"/>
+    /// option or by configuring OpenTelemetry propagators yourself (e.g. by calling
+    /// <c>Sdk.SetDefaultTextMapPropagator</c>).
+    /// </remarks>
+    public static void UseOtlp(this SentryOptions options, TracerProviderBuilder tracerProviderBuilder, Uri? collectorUrl = null)
     {
         if (string.IsNullOrWhiteSpace(options.Dsn))
         {
             throw new ArgumentException("Sentry DSN must be set before calling `SentryOptions.UseOtlp`", nameof(options.Dsn));
         }
-        tracerProviderBuilder.AddSentryOtlpExporter(options.Dsn, collectorUrl, defaultTextMapPropagator);
+        tracerProviderBuilder.AddSentryOtlpExporter(options.Dsn, collectorUrl);
         options.UseOtlp();
     }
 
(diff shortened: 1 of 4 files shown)


src/Sentry.OpenTelemetry.Exporter/SentryOptionsExtensions.cs
```diff
@@ -21,24 +20,19 @@ public static class SentryOptionsExtensions
...
-    ///         could wrap this in a <see cref="CompositeTextMapPropagator"/> if you needed other propagators as well.
-    ///     </para>
-    /// </param>
-    public static void UseOtlp(this SentryOptions options, TracerProviderBuilder tracerProviderBuilder, Uri? collectorUrl = null, TextMapPropagator? defaultTextMapPropagator = null)
+    /// <remarks>
+    /// In line with the OTLP integration specification, this method does not configure an OpenTelemetry propagator.
+    /// Cross-service trace propagation should be enabled either via the <see cref="SentryOptions.PropagateTraceparent"/>
+    /// option or by configuring OpenTelemetry propagators yourself (e.g. by calling
+    /// <c>Sdk.SetDefaultTextMapPropagator</c>).
+    /// </remarks>
```
jamescrosswell, line 27:
```suggestion
    /// <remarks>
    /// This method does not configure an OpenTelemetry propagator.
    /// Cross-service trace propagation should be enabled via the OpenTelemetry SDK (e.g. by calling
    /// <c>Sdk.SetDefaultTextMapPropagator</c>).
    /// </remarks>
```

src/Sentry.OpenTelemetry.Exporter/SentryTracerProviderBuilderExtensions.cs
```diff
@@ -27,20 +25,15 @@ public static class SentryTracerProviderBuilderExtensions
...
-    ///         could wrap this in a <see cref="CompositeTextMapPropagator"/> if you needed other propagators as well.
-    ///     </para>
-    /// </param>
     /// <returns>The supplied <see cref="TracerProviderBuilder"/> for chaining.</returns>
+    /// <remarks>
+    /// In line with the OTLP integration specification, this method does not configure an OpenTelemetry propagator.
+    /// Cross-service trace propagation should be enabled either via the <see cref="SentryOptions.PropagateTraceparent"/>
+    /// option or by configuring OpenTelemetry propagators yourself (e.g. by calling
+    /// <c>Sdk.SetDefaultTextMapPropagator</c>).
+    /// </remarks>
```
jamescrosswell, line 33:
```suggestion
    /// <remarks>
    /// This method does not configure an OpenTelemetry propagator.
    /// Cross-service trace propagation should be enabled via the OpenTelemetry SDK (e.g. by calling
    /// <c>Sdk.SetDefaultTextMapPropagator</c>).
    /// </remarks>
```

dingsdax:
@jamescrosswell this should land in the next major, we'll do the same change in Ruby & Python in their next majors

adinauer (approved):
LGTM, I assume you'll update docs once v7 is released.

PR #5330, opened 2026-06-28
fix: BackpressureMonitor.Dispose() no longer deadlocks on single-threaded targets

### LINKED ISSUE #5237: BackpressureMonitor.Dispose() deadlocks on single-threaded targets (Unity WebGL)
### Package

Sentry

### .NET Flavor

IL2CPP

### .NET Version

2.1 (Unity 2021.3.x)

### OS

Browser

### OS Version

Chrome 148.0.7778.97 (Official Build) (arm64) MacOS 26.4.1

### Development Environment

Rider 2025.x (macOS)

No

_No response_

### SDK Version

4.3.1

_No response_

### Workload Versions

Unity 2021.3.24f1
emscripten (whatever ships with Unity 2021.3.x WebGL — 2.0.x range)
Build settings: IL2CPP, WebGL exception support ExplicitlyThrownExceptionsOnly, Brotli compression
react-unity-webgl v10 on the JS side (drives unityInstance.Quit())

Auto-initialized via Sentry Unity's SentryOptions.asset ScriptableObject (the default Sentry Unity SDK init path — no manual SentrySdk.Init() call in user code). Options Configuration is wired to a SentryOptionsConfiguration : Sentry.Unity.SentryOptionsConfiguration subclass that sets options.EnableBackpressureHandling = false; inside #if UNITY_WEBGL && !UNITY_EDITOR — that one line is the workaround.

### Steps to Reproduce

1. Create a Unity 2021.3.x project with Sentry Unity 4.x installed (default settings — EnableBackpressureHandling = true is the .NET 6.0 default).
2. Build for WebGL target.
3. Embed the build in a web page via react-unity-webgl (or any wrapper that calls unityInstance.Quit() on page navigation).
4. Load the page; let Unity fully initialize.
5. Trigger unityInstance.Quit() — e.g. press the browser Back button on a page that has a popstate listener wired to call Quit(), or close the tab, or navigate away programmatically.
6. Observe: the returned Quit() Promise never resolves; the Unity main loop never returns.

### Expected Result

unityInstance.Quit() resolves; the page navigates cleanly.

### Actual Result

The page wedges indefinitely. The tab appears frozen (no UI updates, the URL may have already updated in the address bar). The user must force-reload to recover. No exception is thrown, no Sentry event is emitted, no console error appears.

### Root cause (analysis included for triage convenience, AI generated):

BackpressureMonitor.Dispose() in [Internal/BackpressureMonitor.cs](https://github.com/getsentry/sentry-dotnet/blob/main/src/Sentry/Internal/BackpressureMonitor.cs) deadlocks on single-threaded platforms:

```
public void Dispose()
{
    try
    {
        _cts.Cancel();
        _workerTask.Wait();   // ← blocks main thread on WebGL
    }
    ...
}
```

The monitor's worker is started via Task.Run(() => DoWorkAsync(_cts.Token)) and loops on await Task.Delay(10s, ct).

On iOS/Android/Windows this works because Task.Run schedules on a real thread-pool worker. The main thread calls _cts.Cancel(), the worker thread observes cancellation from inside Task.Delay, throws OperationCanceledException, and the task completes — Wait() returns.

On Unity WebGL (single-threaded emscripten) Task.Run schedules on the same main loop as Unity. Task.Delay's continuation is also scheduled on that main loop. When Quit() triggers Dispose(), _workerTask.Wait() synchronously blocks the main thread waiting for a continuation that can only run on the main thread. Classic deadlock.

Confirmed by single-line ablation: setting options.EnableBackpressureHandling = false on WebGL avoids constructing the monitor at all and eliminates the wedge. Bisected version-wise: 3.2.4 (pre-BackpressureMonitor) ✅, 4.0.0 ❌, 4.3.1 ❌, 4.3.1 + workaround ✅.

jamescrosswell:
Fixes #5237

## Problem

On single-threaded runtimes (Unity WebGL / browser-wasm) there is no separate thread pool — `Task.Run` and `Task.Delay` continuations all run on the one main thread. `BackpressureMonitor.Dispose()` did:

```csharp
_cts.Cancel();
_workerTask.Wait();   // blocks the only thread
```

`_workerTask.Wait()` synchronously blocks the calling thread waiting for the worker's post-cancellation continuation, which on a single-threaded runtime can only be scheduled on that same blocked thread (e.g. calling `unityInstance.Quit()` in Unity WebGL).

`ConfigureAwait(false)` does not help here. On a desktop host the cancellation continuation escapes to a real thread-pool thread. On single-threaded targets the "thread pool" *is* the main thread, so the continuation has nowhere to run except the thread that is blocked in `Wait()`.

## Fix

`Dispose()` now only requests cancellation and returns — it does not block on the worker. The worker observes the token, exits its `Task.Delay` loop and unwinds on its own; it produces no result anyone needs to await.

Because the monitor's public methods (`DownsampleFactor` / `RecordQueueOverflow` / `RecordRateLimitHit`) never touch `_cts`, it is now safe to dispose the monitor even while the client keeps draining envelopes. This lets `Hub.Dispose()` dispose the monitor again rather than leaking the worker for the process lifetime.

## ⚠️ `Hub.Dispose()`

The call to `_backpressureMonitor?.Dispose()` was removed in #4613 with the comment:
> Don't dispose … without throwing an `ObjectDisposedException`

I've restored it because, with a non-blocking `Dispose()` that only cancels/disposes the CTS, none of the monitor's post-disposal call sites touch the disposed token source — so the original `ObjectDisposedException` should not recur. 

> [!WARNING]
> **TODO**: Confirm this matches the scenario #4613 was guarding against. We can always reinstate the fix from #4613 if we're concerned about this.

## Affected versions

The live deadlock path (`Hub.Dispose()` → `_backpressureMonitor?.Dispose()`) exists in 4.x/5.x (and the Sentry Unity 4.x that reported this). It was *incidentally* removed on `main`/6.0.0+ by #4613, so 6.x no longer reaches the deadlock - but at the cost of leaking the worker. `Dispose()` itself remained latently broken. This PR fixes the root cause so the call is safe again. Users on ≤5.x  operating in a single threaded environment should set `EnableBackpressureHandling = false` as a workaround instead.

### DIFF
--- a/src/Sentry/Internal/BackpressureMonitor.cs
+++ b/src/Sentry/Internal/BackpressureMonitor.cs
@@ -40,7 +40,8 @@ internal class BackpressureMonitor : IDisposable
     internal long LastQueueOverflowTicks => Interlocked.Read(ref _lastQueueOverflow);
     internal long LastRateLimitEventTicks => Interlocked.Read(ref _lastRateLimitEvent);
 
-    public BackpressureMonitor(IDiagnosticLogger? logger, ISystemClock? clock = null, bool enablePeriodicHealthCheck = true)
+    public BackpressureMonitor(IDiagnosticLogger? logger, ISystemClock? clock = null, bool enablePeriodicHealthCheck = true,
+        TaskScheduler? scheduler = null)
     {
         _logger = logger;
         _clock = clock ?? SystemClock.Clock;
@@ -48,7 +49,12 @@ internal class BackpressureMonitor : IDisposable
         if (enablePeriodicHealthCheck)
         {
             _logger?.LogDebug("Starting BackpressureMonitor.");
-            _workerTask = Task.Run(() => DoWorkAsync(_cts.Token));
+            // Default to the thread pool. The scheduler is only injected by tests, to model single-threaded
+            // runtimes (e.g. Unity WebGL) where the worker and its continuations run on the same thread.
+            _workerTask = scheduler is null
+                ? Task.Run(() => DoWorkAsync(_cts.Token))
+                : Task.Factory.StartNew(() => DoWorkAsync(_cts.Token), _cts.Token, TaskCreationOptions.None, scheduler)
+                    .Unwrap();
         }
         else
         {
@@ -146,12 +152,14 @@ internal class BackpressureMonitor : IDisposable
     {
         try
         {
+            // Request cancellation but do NOT block on _workerTask here. On single-threaded runtimes
+            // (e.g. Unity WebGL / browser-wasm) the worker's cancellation continuation can only be scheduled
+            // on the calling thread, so a synchronous _workerTask.Wait() would block the only thread that
+            // could complete it - a deadlock. The worker observes the token, exits its Task.Delay loop and
+            // unwinds on its own; it produces no result we need to await, and the methods callers may still
+            // invoke after disposal (GetDownsampleFactor / RecordQueueOverflow / RecordRateLimitHit) don't
+            // touch the cancellation token source. See https://github.com/getsentry/sentry-dotnet/issues/5237
             _cts.Cancel();
-            _workerTask.Wait();
-        }
-        catch (AggregateException ex) when (ex.InnerException is OperationCanceledException)
-        {
-            // Ignore cancellation
         }
         catch (Exception ex)
         {
--- a/src/Sentry/Internal/BackpressureMonitor.cs
+++ b/src/Sentry/Internal/BackpressureMonitor.cs
@@ -153,12 +153,9 @@ internal class BackpressureMonitor : IDisposable
         try
         {
             // Request cancellation but do NOT block on _workerTask here. On single-threaded runtimes
-            // (e.g. Unity WebGL / browser-wasm) the worker's cancellation continuation can only be scheduled
-            // on the calling thread, so a synchronous _workerTask.Wait() would block the only thread that
-            // could complete it - a deadlock. The worker observes the token, exits its Task.Delay loop and
-            // unwinds on its own; it produces no result we need to await, and the methods callers may still
-            // invoke after disposal (GetDownsampleFactor / RecordQueueOverflow / RecordRateLimitHit) don't
-            // touch the cancellation token source. See https://github.com/getsentry/sentry-dotnet/issues/5237
+            // (e.g. Unity WebGL / browser-wasm) _workerTask.Wait() would cause a deadlock.
+            // The worker observes the token and unwinds on its own.
+            // See https://github.com/getsentry/sentry-dotnet/issues/5237
             _cts.Cancel();
         }
         catch (Exception ex)
--- a/src/Sentry/Internal/BackpressureMonitor.cs
+++ b/src/Sentry/Internal/BackpressureMonitor.cs
@@ -166,10 +166,9 @@ internal class BackpressureMonitor : IDisposable
         }
         finally
         {
-            // Dispose the CancellationTokenSource only once the worker has stopped using the token, but
-            // without blocking the calling thread. Disposing it inline would race the worker: if it ran while
-            // the worker was registering its Task.Delay continuation, the worker could observe an
-            // ObjectDisposedException - which it doesn't catch - and fault with an unobserved task exception.
+            // Dispose the CTS once the worker has stopped using the token, but
+            // without blocking the calling thread. Disposing it inline would race the worker 
+            //  and could lead to an ObjectDisposedException
             _workerTask.ContinueWith(
                 static (_, state) => ((CancellationTokenSource)state!).Dispose(),
                 _cts,
--- a/src/Sentry/Internal/BackpressureMonitor.cs
+++ b/src/Sentry/Internal/BackpressureMonitor.cs
@@ -36,6 +36,7 @@ internal class BackpressureMonitor : IDisposable
     private readonly CancellationTokenSource _cts = new();
 
     private readonly Task _workerTask;
+    internal Task WorkerTask => _workerTask;
     internal int DownsampleLevel => _downsampleLevel;
     internal long LastQueueOverflowTicks => Interlocked.Read(ref _lastQueueOverflow);
     internal long LastRateLimitEventTicks => Interlocked.Read(ref _lastRateLimitEvent);
@@ -168,7 +169,16 @@ internal class BackpressureMonitor : IDisposable
         }
         finally
         {
-            _cts.Dispose();
+            // Dispose the CancellationTokenSource only once the worker has stopped using the token, but
+            // without blocking the calling thread. Disposing it inline would race the worker: if it ran while
+            // the worker was registering its Task.Delay continuation, the worker could observe an
+            // ObjectDisposedException - which it doesn't catch - and fault with an unobserved task exception.
+            _workerTask.ContinueWith(
+                static (_, state) => ((CancellationTokenSource)state!).Dispose(),
+                _cts,
+                CancellationToken.None,
+                TaskContinuationOptions.ExecuteSynchronously,
+                TaskScheduler.Default);
         }
     }
 }
--- a/src/Sentry/Internal/BackpressureMonitor.cs
+++ b/src/Sentry/Internal/BackpressureMonitor.cs
@@ -51,12 +51,13 @@ internal class BackpressureMonitor : IDisposable
         if (enablePeriodicHealthCheck)
         {
             _logger?.LogDebug("Starting BackpressureMonitor.");
-            // Default to the thread pool. The scheduler is only injected by tests, to model single-threaded
-            // runtimes (e.g. Unity WebGL) where the worker and its continuations run on the same thread.
-            _workerTask = scheduler is null
-                ? Task.Run(() => DoWorkAsync(_cts.Token))
-                : Task.Factory.StartNew(() => DoWorkAsync(_cts.Token), _cts.Token, TaskCreationOptions.None, scheduler)
-                    .Unwrap();
+            // Equivalent to Task.Run(() => DoWorkAsync(_cts.Token)): same creation options (DenyChildAttach) and,
+            // by default, the same scheduler (the thread pool). Tests inject a scheduler to model single-threaded
+            // runtimes (e.g. Unity WebGL) where the worker and its continuations must run on the same thread.
+            scheduler ??= TaskScheduler.Default;
+            _workerTask = Task.Factory
+                .StartNew(() => DoWorkAsync(_cts.Token), CancellationToken.None, TaskCreationOptions.DenyChildAttach, scheduler)
+                .Unwrap();
         }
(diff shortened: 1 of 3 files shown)


src/Sentry/Internal/BackpressureMonitor.cs
```diff
@@ -146,12 +152,14 @@ public void Dispose()
     {
         try
         {
+            // Request cancellation but do NOT block on _workerTask here. On single-threaded runtimes
+            // (e.g. Unity WebGL / browser-wasm) the worker's cancellation continuation can only be scheduled
+            // on the calling thread, so a synchronous _workerTask.Wait() would block the only thread that
+            // could complete it - a deadlock. The worker observes the token, exits its Task.Delay loop and
+            // unwinds on its own; it produces no result we need to await, and the methods callers may still
+            // invoke after disposal (GetDownsampleFactor / RecordQueueOverflow / RecordRateLimitHit) don't
+            // touch the cancellation token source. See https://github.com/getsentry/sentry-dotnet/issues/5237
```
jamescrosswell:
```suggestion
            // Request cancellation but do NOT block on _workerTask here. On single-threaded runtimes
            // (e.g. Unity WebGL / browser-wasm) _workerTask.Wait() would cause a deadlock.
            // The worker observes the token and unwinds on its own.
            // See https://github.com/getsentry/sentry-dotnet/issues/5237
```

src/Sentry/Internal/Hub.cs
```diff
@@ -943,8 +943,10 @@ public void Dispose()
         }
         //Don't dispose of ScopeManager since we want dangling transactions to still be able to access tags.
 
-        // Don't dispose of _backpressureMonitor since we want the client to continue to process envelopes without
-        // throwing an ObjectDisposedException.
+        // Stop the backpressure monitor's worker so it doesn't outlive the hub. Its Dispose only cancels the
+        // worker (it does not block, and its public methods don't touch the cancellation token source), so the
+        // client can still safely read the downsample factor while it drains any remaining envelopes.
```
```suggestion
        // Stop the backpressure monitor's worker so it doesn't outlive the hub...  this cancels the 
        // worker but clients can still read the downsample factor while draining.
```

src/Sentry/Internal/BackpressureMonitor.cs
```diff
@@ -146,12 +152,11 @@ public void Dispose()
...
+            // The worker observes the token and unwinds on its own.
+            // See https://github.com/getsentry/sentry-dotnet/issues/5237
             _cts.Cancel();
-            _workerTask.Wait();
-        }
-        catch (AggregateException ex) when (ex.InnerException is OperationCanceledException)
-        {
-            // Ignore cancellation
         }
         catch (Exception ex)
```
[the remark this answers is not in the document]
jamescrosswell: Good catch — this is valid, and it's a regression from this PR. The old `_workerTask.Wait()` did double duty: besides being the source of the deadlock, it also guaranteed the worker had stopped touching the token before the `finally` disposed `_cts`. Removing the `Wait()` opened a window where, on a multi-threaded target, the worker could be inside `Task.Delay(…, token)` registering its continuation exactly as the disposing thread disposes the CTS — and on .NET Framework `Register` on a disposed source throws `ObjectDisposedException`, which the worker doesn't catch → unobserved task exception. (No race on wasm itself, but the SDK ships net462/netstandard, so it's real.)

Fixed in af942a3f by disposing the CTS from a **non-blocking continuation** on the worker task instead of inline, so it's only disposed once the worker has fully unwound — preserving the deadlock fix while closing the race:

```csharp
_cts.Cancel();
...
_workerTask.ContinueWith(
    static (_, state) => ((CancellationTokenSource)state!).Dispose(),
    _cts, CancellationToken.None,
    TaskContinuationOptions.ExecuteSynchronously, TaskScheduler.Default);
```

Added `Dispose_WorkerRunsToCompletionWithoutFaulting` which awaits the worker after `Dispose()` and asserts `RanToCompletion` (it would surface the fault if the CTS were pulled out from under it).

src/Sentry/Internal/BackpressureMonitor.cs
```diff
@@ -160,7 +166,16 @@ public void Dispose()
         }
         finally
         {
-            _cts.Dispose();
+            // Dispose the CancellationTokenSource only once the worker has stopped using the token, but
+            // without blocking the calling thread. Disposing it inline would race the worker: if it ran while
+            // the worker was registering its Task.Delay continuation, the worker could observe an
+            // ObjectDisposedException - which it doesn't catch - and fault with an unobserved task exception.
```
jamescrosswell:
```suggestion
            // Dispose the CTS once the worker has stopped using the token, but
            // without blocking the calling thread. Disposing it inline would race the worker 
            //  and could lead to an ObjectDisposedException
```

src/Sentry/Internal/BackpressureMonitor.cs
```diff
@@ -146,12 +153,11 @@ public void Dispose()
...
+            // See https://github.com/getsentry/sentry-dotnet/issues/5237
             _cts.Cancel();
-            _workerTask.Wait();
-        }
-        catch (AggregateException ex) when (ex.InnerException is OperationCanceledException)
-        {
-            // Ignore cancellation
         }
         catch (Exception ex)
         {
```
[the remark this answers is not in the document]
jamescrosswell: Valid (LOW), and it wasn't actually fixed by the commit the bot auto-linked (`a56be9e` was just a comment tweak that happened to touch these lines). Addressed properly in 756255b2 with a thread-safe idempotency guard:

```csharp
public void Dispose()
{
    if (Interlocked.Exchange(ref _disposed, 1) != 0)
    {
        return;
    }
    ...
}
```

Now only the first caller runs the disposal logic — no spurious `ObjectDisposedException` warning or redundant continuation on repeated calls. Added `Dispose_CalledMultipleTimes_IsIdempotent` to cover it.

src/Sentry/Internal/BackpressureMonitor.cs
```diff
@@ -36,19 +36,27 @@ internal class BackpressureMonitor : IDisposable
...
 
         if (enablePeriodicHealthCheck)
         {
             _logger?.LogDebug("Starting BackpressureMonitor.");
-            _workerTask = Task.Run(() => DoWorkAsync(_cts.Token));
+            // Default to the thread pool. The scheduler is only injected by tests, to model single-threaded
+            // runtimes (e.g. Unity WebGL) where the worker and its continuations run on the same thread.
+            _workerTask = scheduler is null
+                ? Task.Run(() => DoWorkAsync(_cts.Token))
+                : Task.Factory.StartNew(() => DoWorkAsync(_cts.Token), _cts.Token, TaskCreationOptions.None, scheduler)
```
Flash0ver: suggestion: from `TaskCreationOptions.None` to `TaskCreationOptions.DenyChildAttach`

See `Task.Run`:
https://github.com/dotnet/dotnet/blob/cda556645c4a4e9a40d606893e53b6bbd714221f/src/runtime/src/libraries/System.Private.CoreLib/src/System/Threading/Tasks/Task.cs#L5688-L5691

jamescrosswell:
```suggestion
            scheduler ??= TaskScheduler.Default;
            _workerTask = Task.Factory.StartNew(() => DoWorkAsync(_cts.Token), _cts.Token, TaskCreationOptions.DenyChildAttach, scheduler)
```

This might be easier to read - it makes it clear that the same code path is taken for testing and production - the only thing that changes is the scheduler. The arguments provided to `Task.Factory.StartNew` here are desiged to match exactly what would happen if we ran the old/previous code: `Task.Run(() => DoWorkAsync(_cts.Token))`, with the one obvious exception of swapping out the TaskScheduler with one that simulates a single threaded environment for testing purposes.

src/Sentry/Internal/BackpressureMonitor.cs
```diff
@@ -160,7 +174,15 @@ public void Dispose()
...
-            _cts.Dispose();
+            // Dispose the CTS once the worker has stopped using the token, but
+            // without blocking the calling thread. Disposing it inline would race the worker 
+            //  and could lead to an ObjectDisposedException
+            _workerTask.ContinueWith(
+                static (_, state) => ((CancellationTokenSource)state!).Dispose(),
+                _cts,
+                CancellationToken.None,
+                TaskContinuationOptions.ExecuteSynchronously,
+                TaskScheduler.Default);
```
jamescrosswell, line 186: Should we be passing `TaskScheduler.Default` here or `TaskScheduler.Current`? In production it doesn't really matter, since those two things will always be the same. They are only ever different if/when we explicitly pass in a non default scheduler at task creation time... for us, that only happens when running unit tests.  So if we used `TaskScheduler.Current`, I think we'd end up using the Scheduler injected for unit testing  purposes here as well (not just when creating the Task above). It probably doesn't really matter since I don't think we have any concurrency problems with this code (it's just disposing of the cts synchronously once the task completes) but for the sake of consistency, maybe it's better to use `TaskScheduler.Current` here?

src/Sentry/Internal/BackpressureMonitor.cs
jamescrosswell: Done in 35ece534 — collapsed to a single `Task.Factory.StartNew` path with `scheduler ??= TaskScheduler.Default` and `DenyChildAttach`, so prod and tests differ only in the scheduler. 👍

One small deviation from the exact snippet: I kept the outer creation token as `CancellationToken.None` rather than `_cts.Token`. `Task.Run(() => DoWorkAsync(_cts.Token))` passes **no** token to the task factory — the `_cts.Token` only flows into `DoWorkAsync` — so `None` is the exact match. Passing `_cts.Token` to `StartNew` would let the worker be cancelled *before it starts*, which turns `_workerTask` into a `Canceled` task and made `Dispose_WorkerRunsToCompletionWithoutFaulting` flaky (it asserts `RanToCompletion`). With `None` it's stable across repeated runs.

jamescrosswell: Applied in 35ece534 — switched to `TaskCreationOptions.DenyChildAttach` to match `Task.Run`. Thanks for the pointer to the runtime source.

src/Sentry/Internal/BackpressureMonitor.cs
jamescrosswell: **`TaskScheduler.Current` wouldn't pick up the injected scheduler anyway.** `Current` is evaluated at the point `ContinueWith` is called — i.e. *inside `Dispose()`*, on whatever thread calls it. In the tests `Dispose()` is invoked from the test thread (or a plain `Task.Run`), not from within a task running on the injected scheduler, so `Current` resolves to `Default` there too. The injected scheduler only governs the worker task's own execution, not this continuation. So `Current` vs `Default` makes no observable difference in our tests.

bitsandfoxes (approved):
Thanks for the fix!

PR #5333, opened 2026-07-01
feat: add IHub.RecordTransaction to record completed transactions and spans

### LINKED ISSUE #2683: Ability to overwrite start/end times for transactions and spans
### Problem Statement

We have some systems that are not able to run a sentry SDK directly due to limitations on 3rd party binaries.
For these we are planning to manually log a simple span tree and forward the data to a proxy. The proxy would then use a Sentry SDK to upload it all to Sentry.

In order for this to work we'd need to be able to manually set start/end times via the Sentry SDK as they won't be matching the times on the proxy, but rather those manually recorded on the instrumented servers.

### Solution Brainstorm

We've noticed that the Javascript SDK has the ability to change these fields, see here:
https://github.com/getsentry/sentry-javascript/blob/5c2546f6ee2b56321614ba6f78811163fe783db4/packages/core/src/tracing/span.ts
https://github.com/getsentry/sentry-javascript/blob/5c2546f6ee2b56321614ba6f78811163fe783db4/packages/core/src/tracing/transaction.ts

whereas those same members are private in the .NET equivalent:
https://github.com/getsentry/sentry-dotnet/blob/d59d08cbc54669a0e98684516d2ce54839b6b55a/src/Sentry/Transaction.cs

Is this something you would be open to adding so our proxy does not have to use javascript?
Or, if the use case is better handled some other way that we've missed, please let us know.

jamescrosswell:
## Overview

Draft to spitball the shape of an API for **recording transactions/spans that completed elsewhere** — e.g. work measured on another machine or process and replayed through a proxy, which is the use case in #2683.

Closes #2683

## Approach

The obvious option is public start/end setters on `ISpan`. We discussed a few reasons not to:

- The live-tracing timestamps are derived from a monotonic stopwatch; a bare `StartTimestamp` setter that doesn't reset the stopwatch baseline lets you produce a span that starts in the past but ends "now" (inflated duration). The start/end pair has an invariant a lone setter can't enforce.
- It widens/pollutes `ISpan` (the most-implemented interface) with mutation that only makes sense for replayed data.

Instead this adds a **builder-style recorder**:

```csharp
hub.RecordTransaction("checkout", "http.server", start, duration,
    traceId: originTraceId, spanId: originRootSpanId, configure: tx =>
    {
        tx.Release = "1.2.3";
        tx.RecordSpan("db.query", childStart, childDuration, spanId: originChildId, configure: c =>
        {
            c.Description = "SELECT ...";
            c.RecordSpan("db.connect", gcStart, gcDuration); // arbitrary nesting
        });
    });
```

### Why this shape

- **Timing is up-front** (`start` + `duration`), so a recorded span can't be half-specified and duration can't be negative — the original footgun is structurally unreachable rather than merely diagnosed.
- **Nesting** lives on the recorder itself (`ISpanRecorder.RecordSpan`), so a whole tree can be built with plain recursion. Parent is structural; only `spanId` is overridable per node.
- **Trace stitching**: origin `traceId` / `spanId` / `parentSpanId` can be preserved.
- **Captured once** when the callback returns — no stopwatch, idle timer, or sampling roll. It's not a live transaction, so those don't apply.
- **No live-scope bleed**: captured against a clean scope so the current process's breadcrumbs/user/tags/contexts don't attach to work that happened elsewhere.

### Public surface

Intentionally minimal — `RecordTransaction` is an **extension method on `IHub`**, so `IHub` / `ISpan` / `ITransaction` are all unchanged (no breaking interface changes needed to ship). New public types: `ISpanRecorder`, `ITransactionRecorder`. Internally it reuses the existing `TransactionTracer`/`SpanTracer` → `SentryTransaction` conversion path, so there are no changes to serialization.

## Open questions for reviewers

- **`duration` vs `end`**: went with `start` + `duration` so the value can't express a reversed pair. Alternative is a required `end` (matches the wire protocol and the origin data, which usually carries start+end). Easy to offer both.
- **Discoverability**: currently only on `IHub`. Worth mirroring on the `SentrySdk` static facade?
- **Scope isolation**: draft uses a fresh `Scope(options)`. Do we want any current-scope data (release/environment/dist) to flow through, or keep it fully isolated and rely on the `Release`/`Environment` setters?
- **`BeforeSendTransaction`** still runs (last-chance scrubbing); only the sampling roll is skipped. Confirm that's the desired split.

### Changelog Entry

Added `SentrySdk.RecordTransaction` to record already-completed transactions and spans (e.g. replayed through a proxy) [#5333](https://github.com/getsentry/sentry-dotnet/pull/5333)


src/Sentry/HubExtensions.Record.cs
```diff
@@ -0,0 +1,78 @@
...
+
+        var transaction = new SentryTransaction(tracer);
+
+        // A recorded transaction represents work that happened elsewhere, so we don't want the current
+        // (live) scope's breadcrumbs/user/tags/contexts leaking onto it. Capture against a clean scope when
+        // we have options to construct one; otherwise fall back (e.g. a disabled hub, where this is a no-op).
+        var options = hub.GetSentryOptions();
+        if (options is not null)
+        {
+            hub.CaptureTransaction(transaction, new Scope(options), null);
```
jamescrosswell: The SDK user is potentially going to want to be able to set scope values as well (to match values captured during the original trace). We might need to expose this new/fresh scope via the `ITransactionRecorder` so that it can be configured via the callback.

src/Sentry/HubExtensions.Record.cs
jamescrosswell, line 1: There's no need for this to be in a separate file... let's just put it in `HubExtensions.cs`

src/Sentry/HubExtensions.Record.cs
```diff
@@ -0,0 +1,78 @@
...
+    /// <param name="duration">How long the transaction ran. Must not be negative.</param>
+    /// <param name="traceId">Optional trace id to preserve from the originating system; generated when omitted.</param>
+    /// <param name="spanId">Optional root span id to preserve; generated when omitted.</param>
+    /// <param name="parentSpanId">Optional parent span id, when continuing a trace from another service.</param>
+    /// <param name="configure">
+    /// Optional callback to set metadata and record the span tree via
+    /// <see cref="ISpanRecorder.RecordSpan(string, DateTimeOffset, TimeSpan, SpanId?, Action{ISpanRecorder}?)"/>.
+    /// </param>
+    /// <returns>The event id of the captured transaction.</returns>
+    public static SentryId RecordTransaction(
```
jamescrosswell: We will need to add this to SentrySdk.

SDK users aren't going to be able to access the `IHub` intance in every application type.

src/Sentry/HubExtensions.Record.cs
jamescrosswell: Done — merged into `HubExtensions.cs` and dropped the separate file (`HubExtensions` no longer needs to be `partial`).

src/Sentry/HubExtensions.Record.cs
jamescrosswell: Added `SentrySdk.RecordTransaction` as a facade over `CurrentHub.RecordTransaction`, mirroring the `StartTransaction` overloads.

src/Sentry/HubExtensions.Record.cs
jamescrosswell: Exposed the fresh scope via `ITransactionRecorder.ConfigureScope(Action<Scope>)` — the caller can now set user/tags/contexts/breadcrumbs captured alongside the original trace, still without inheriting the current live scope. Added a test (`RecordTransaction_ConfigureScope_AppliesToCaptureScope`).

src/Sentry/ISpanRecorder.cs
```diff
@@ -0,0 +1,87 @@
...
+        SpanId? spanId = null,
+        Action<ISpanRecorder>? configure = null);
+}
+
+/// <summary>
+/// Builds a transaction (and its span tree) that has already completed elsewhere. Obtained inside the
+/// <c>configure</c> callback of <see cref="HubExtensions.RecordTransaction"/>. When the callback returns, the
+/// whole tree is materialized and captured once — no live tracing, stopwatch, or sampling roll is involved.
+/// </summary>
+public interface ITransactionRecorder : ISpanRecorder
```
jamescrosswell, line 63: This could go in it's own file (we generally have one file per type in this repo, unless there is some specific reason to hold things together). 

On the other hand, we've got both of the concrete implementations in the same `SpanRecorder.cs` file... so 🤷🏻 I think the main argument for that is that they both share `SpanRecorderInternals` - arguably that could be refactored:
- `SpanRecorderInternals.RecordChild` could be moved to be a static method of either `SpanRecorder` or `TransactionRecorder`
- Alternatively `SpanRecorderInternals` could be renamed `SpanRecorderBase` and made abstract rather than static... it's essentially just responsible for storing the `TransactionTracer _owner` and setting this on any spans created by the recorder callback. That  could be achieved via inheritence.

src/Sentry/Internal/SpanRecorder.cs
```diff
@@ -0,0 +1,143 @@
+namespace Sentry.Internal;
+
+/// <summary>
+/// Shared logic for materializing a recorded child span onto an existing <see cref="TransactionTracer"/>.
+/// </summary>
+internal static class SpanRecorderInternals
```
jamescrosswell:
```suggestion
file static class SpanRecorderInternals
```

Only needs to be accessible from within this file.

src/Sentry/ISpanRecorder.cs
jamescrosswell: Went with the inheritance option. `SpanRecorderInternals` is gone — replaced by an abstract `SpanRecorderBase` that holds the owning `TransactionTracer` plus the node as an `ISpan`, so the `Description`/`Status`/`SetTag`/`SetData`/`SpanId` proxying is written once and `SpanRecorder` collapses to a one-line subclass. `TransactionRecorder` just adds the transaction-only members (`TraceId`/`Release`/`Environment`/`ConfigureScope`). No measurable perf cost — `RecordSpan` was already interface-dispatched, so nothing new there; the only delta is the metadata setters going through `ISpan`, which is dwarfed by StartChild + serialization + send. Public API unchanged.

src/Sentry/Internal/SpanRecorder.cs
jamescrosswell: Moot now — `SpanRecorderInternals` was removed in the inheritance refactor, so there's no longer a helper class to scope to `file`.

src/Sentry/Internal/SpanRecorder.cs
```diff
@@ -0,0 +1,114 @@
...
+        var recorder = new SpanRecorder(Owner, span);
+        configure?.Invoke(recorder);
+        return recorder;
+    }
+}
+
+/// <summary>
+/// <see cref="ISpanRecorder"/> backed by a <see cref="SpanTracer"/> whose timing has been set explicitly.
+/// </summary>
+internal sealed class SpanRecorder : SpanRecorderBase
```
jamescrosswell, line 69: **Note:** Arguably we don't need `SpanRecorderBase` (`TransactionRecorder` could descend from `SpanRecorder and we just fold all the stuff from the Base class into here). That would give us one less class.

Probably the main reason to do this is so that we can add the `sealed` keyword to `SpanRecorder`. Since the cost of doing this is minimal, I've left the base class in for the time being (unless anyone has really strong opinions about this).

dingsdax (approved):

PR #5335, opened 2026-07-02
Added Sentry.Samples.OpenTelemetry.MongoDB

jamescrosswell:
Demonstrates how to instrument MongoDB using OTLP. Spans captured this way are automatically scrubbed and available in Sentry's Queries dashboard.

image

I'm not sure if we want to push this in our core repo - technically it's not demonstrating anything Sentry has built (except the OTLP integration - but there are tens of thousands of different ways that can be used). On the other hand, MongoDB is extremely popular and wiring all of this up might not be entirely obvious to SDK users. 

Other options (not mutually exclusive):
- Put it in https://github.com/sentry-demos 
- Write up a short blog post (good opportunity to talk about the new OTLP integration and couple it with a valuable use case)

> [!NOTE]
> It would be possible to make Mongo DB instrumentation available even to SDK customers who are not using our OTLP integration... basically we'd have to build an integration that includes a custom ActivityListener that listens to the MongoDB events (similar to what we did to support Microsoft.Extensions.AI.Abstractions) and then translate those into Sentry spans. I'm not sure we want to do that though - it's another thing to maintain. I guess there would have to be clear reasons why customers would want MongoDB instumentation but not be able to or not want to use our OTLP integration. Currently we don't have anyone in that situation talking to us so it may be an imaginary problem to solve.

### DIFF
--- a/samples/Sentry.Samples.OpenTelemetry.MongoDB/Program.cs
+++ b/samples/Sentry.Samples.OpenTelemetry.MongoDB/Program.cs
@@ -1,0 +1,92 @@
+/*
+ * This sample demonstrates how MongoDB queries can be captured as Sentry spans using the
+ * MongoDB.Driver's built-in OpenTelemetry instrumentation (available since driver version 3.7.0).
+ *
+ * The sample requires a MongoDB server. See the README.md alongside this sample for instructions
+ * on how to start one using Docker.
+ */
+
+using System.Diagnostics;
+using MongoDB.Bson;
+using MongoDB.Driver;
+using MongoDB.Driver.Core.Configuration;
+using OpenTelemetry;
+using OpenTelemetry.Trace;
+using Sentry.OpenTelemetry;
+using Sentry.OpenTelemetry.Exporter;
+
+#if SENTRY_DSN_DEFINED_IN_ENV
+var dsn = Environment.GetEnvironmentVariable("SENTRY_DSN")
+          ?? throw new InvalidOperationException("SENTRY_DSN environment variable is not set");
+#else
+// A DSN is required. You can set here in code, or you can set it in the SENTRY_DSN environment variable.
+// See https://docs.sentry.io/product/sentry-basics/dsn-explainer/
+var dsn = SamplesShared.Dsn;
+#endif
+
+var activitySource = new ActivitySource("Sentry.Samples.OpenTelemetry.MongoDB");
+
+SentrySdk.Init(options =>
+{
+    options.Dsn = dsn;
+    options.Debug = true;
+    options.TracesSampleRate = 1.0;
+    options.UseOtlp(); // <-- Configure Sentry to use OpenTelemetry trace information
+});
+
+using var tracerProvider = Sdk.CreateTracerProviderBuilder()
+    .AddSource(activitySource.Name)
+    .AddSource(MongoTelemetry.ActivitySourceName) // <-- Subscribe to the MongoDB driver's built-in instrumentation
+    .AddSentryOtlpExporter(dsn) // <-- Configure OpenTelemetry to send traces to Sentry over OTLP
+    .Build();
+
+// MongoDB.Driver 3.7.0+ creates OpenTelemetry activities for every command out of the box.
+// TracingOptions is only needed to tweak that behaviour - here we opt in to capturing the query
+// text on each span (off by default), which shows up in Sentry as the `db.query.text` span data.
+var mongoUri = Environment.GetEnvironmentVariable("MONGODB_URI") ?? "mongodb://localhost:27017";
+var clientSettings = MongoClientSettings.FromConnectionString(mongoUri);
+clientSettings.ServerSelectionTimeout = TimeSpan.FromSeconds(3); // <-- Fail fast if MongoDB isn't running
+clientSettings.TracingOptions = new TracingOptions
+{
+    QueryTextMaxLength = 4096
+};
+var mongoClient = new MongoClient(clientSettings);
+
+var db = mongoClient.GetDatabase("sentry_mongo_sample");
+var fruit = db.GetCollection<BsonDocument>("fruit");
+
+try
+{
+    // Everything within this activity gets sent to Sentry as a single transaction, containing one
+    // span for each MongoDB command.
+    using (activitySource.StartActivity("Fruit Salad"))
+    {
+        await fruit.InsertManyAsync([
+            new BsonDocument { { "name", "Apple" }, { "color", "Red" } },
+            new BsonDocument { { "name", "Banana" }, { "color", "Yellow" } },
+            new BsonDocument { { "name", "Cherry" }, { "color", "Red" } },
+            new BsonDocument { { "name", "Kiwi" }, { "color", "Green" } }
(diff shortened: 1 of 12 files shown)


samples/Sentry.Samples.OpenTelemetry.MongoDB/Program.cs
```diff
@@ -0,0 +1,92 @@
...
+
+using var tracerProvider = Sdk.CreateTracerProviderBuilder()
+    .AddSource(activitySource.Name)
+    .AddSource(MongoTelemetry.ActivitySourceName) // <-- Subscribe to the MongoDB driver's built-in instrumentation
+    .AddSentryOtlpExporter(dsn) // <-- Configure OpenTelemetry to send traces to Sentry over OTLP
+    .Build();
+
+// MongoDB.Driver 3.7.0+ creates OpenTelemetry activities for every command out of the box.
+// TracingOptions is only needed to tweak that behaviour - here we opt in to capturing the query
+// text on each span (off by default), which shows up in Sentry as the `db.query.text` span data.
```
jamescrosswell, line 46:
```suggestion
// text on each span (off by default), which shows up in Sentry as the `db.query.text` span data.
// Sentry automatically scrubs query parameter values to protect PII.
```

samples/Sentry.Samples.OpenTelemetry.MongoDB/README.md
```diff
@@ -0,0 +1,71 @@
...
+`TracingOptions` on the client settings, so each span includes the (truncated) command as
+`db.query.text`:
+
+```csharp
+clientSettings.TracingOptions = new TracingOptions
+{
+    QueryTextMaxLength = 4096
+};
+```
+
```
jamescrosswell, line 28:
```suggestion

Note that Sentry automatically scrubs query parameter values to protect PII.
```

samples/Sentry.Samples.OpenTelemetry.MongoDB/Program.cs
```diff
@@ -0,0 +1,93 @@
...
+        await mongoClient.DropDatabaseAsync(db.DatabaseNamespace.DatabaseName);
+    }
+}
+catch (TimeoutException)
+{
+    Console.WriteLine($"""
+        Could not connect to MongoDB at {mongoUri}
+        Is MongoDB running? See the README.md for this sample for instructions on starting MongoDB using Docker.
+        """);
+}
```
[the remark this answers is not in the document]
jamescrosswell: This is a false positive — `catch (TimeoutException)` is correct here.

When the server can't be selected within `ServerSelectionTimeout`, the driver propagates a plain `System.TimeoutException`, **not** a `MongoException`. The `MongoConnectionException` mentioned above is real, but it's only embedded as the `HeartbeatException` inside the cluster-state diagnostics of the timeout message — it's the *inner reason* a server couldn't be selected, not the type that reaches the caller.

Source: [`Cluster.HandleServerSelectionException`](https://github.com/mongodb/mongo-csharp-driver/blob/v3.9.0/src/MongoDB.Driver/Core/Clusters/Cluster.cs#L296-L302) (driver v3.9.0) wraps and throws a `new TimeoutException(...)` — that's the message with the embedded `MongoConnectionException` that this finding was pattern-matching on.

Verified empirically with driver 3.9.0 and this sample's exact settings (`ServerSelectionTimeout = 3s`), catching `Exception` and printing the runtime type:

| Scenario | Thrown type | `is TimeoutException` | `is MongoException` |
|---|---|---|---|
| Connection refused (closed port) | `System.TimeoutException` | ✅ | ❌ |
| Unreachable host | `System.TimeoutException` | ✅ | ❌ |

Broadening the catch would actually be a regression: the intent is to show the "Is MongoDB running?" hint *only* for the connection-timeout case. A genuine `MongoException` (auth failure, a real query bug, etc.) should surface as itself rather than be masked behind a misleading "is MongoDB running?" message.

samples/Sentry.Samples.OpenTelemetry.MongoDB/Program.cs
```diff
@@ -0,0 +1,93 @@
...
+using var tracerProvider = Sdk.CreateTracerProviderBuilder()
+    .AddSource(activitySource.Name)
+    .AddSource(MongoTelemetry.ActivitySourceName) // <-- Subscribe to the MongoDB driver's built-in instrumentation
+    .AddSentryOtlpExporter(dsn) // <-- Configure OpenTelemetry to send traces to Sentry over OTLP
+    .Build();
+
+// MongoDB.Driver 3.7.0+ creates OpenTelemetry activities for every command out of the box.
+// TracingOptions is only needed to tweak that behaviour - here we opt in to capturing the query
+// text on each span (off by default), which shows up in Sentry as the `db.query.text` span data.
+// Sentry automatically scrubs query parameter values to protect PII.
```
dingsdax: Does it when going through OTLP endpoint? I am not sure about that 🤔?

jamescrosswell: Hm, so it turns out it kind of does but not really. It does for the purposes of grouping queries in the dashboard but ultimately the raw queries are retained for the Trace/Span samples... so from a PII perspective it doesn't. 

I've corrected in [ce26817](https://github.com/getsentry/sentry-dotnet/pull/5335/commits/ce268174cb6a9df4a2fd1a15ea200cc3d0cc64c6) - also added a custom OTEL SpanProcessor to demonstrate how SDK users can implement scrubbing client side.

The blog post will need updating as well - once this PR is merged, I can link to the sample in the repo from the blog post.

dingsdax: please have a final look at the .NET content from getsentry/sentry-docs#18441 I'm quite confident it's OK, but still 🙏, will need to check the last Symfony error and then it's good to go 🚀

samples/Sentry.Samples.OpenTelemetry.MongoDB/RedactSensitiveMongoData.cs
```diff
@@ -0,0 +1,38 @@
...
+    // Matches "contributor": "<value>" in MongoDB's extended JSON (spacing may vary).
+    [GeneratedRegex("(\"contributor\"\\s*:\\s*)\"[^\"]*\"")]
+    private static partial Regex SensitiveField();
+
+    public override void OnEnd(Activity activity)
+    {
+        if (activity.GetTagItem("db.query.text") is string queryText)
+        {
+            activity.SetTag("db.query.text", SensitiveField().Replace(queryText, "$1\"[Filtered]\""));
+        }
```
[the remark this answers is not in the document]
jamescrosswell: Not true - the redacted spans are hitting Sentry just fine. See [this trace](https://sentry-sdks.sentry.io/explore/traces/trace/d9467070fc6b5722c500b2f306a69776/?mode=samples&node=span-bc4ab0c770503d47&pageEnd&pageStart&project=5428537&source=traces&statsPeriod=24h&table=trace&timestamp=1783401665.611), for example:
image

dingsdax (approved):

PR #5337, opened 2026-07-03
fix(otel): only set Scope.Transaction in SentrySpanProcessor when null

### LINKED ISSUE #5314: OpenTelemetry SpanProcessor should only set Scope.Transaction when it is null
### Background

Split out from getsentry/sentry-dotnet#2399.

The OpenTelemetry `SentrySpanProcessor` currently sets `Scope.Transaction` **unconditionally** when it starts a new (root) transaction:

[https://github.com/getsentry/sentry-dotnet/blob/main/src/Sentry.OpenTelemetry/SentrySpanProcessor.cs#L175](<https://github.com/getsentry/sentry-dotnet/blob/main/src/Sentry.OpenTelemetry/SentrySpanProcessor.cs#L175>)

This is guarded by the parent-span logic above it (a new transaction is only started when the activity has no parent span we already track):

[https://github.com/getsentry/sentry-dotnet/blob/14c1d8c251b76f525206d8f344ab18838ba39fd1/src/Sentry.OpenTelemetry/SentrySpanProcessor.cs#L96-L114](<https://github.com/getsentry/sentry-dotnet/blob/14c1d8c251b76f525206d8f344ab18838ba39fd1/src/Sentry.OpenTelemetry/SentrySpanProcessor.cs#L96-L114>)

So in practice this only overwrites an existing `Scope.Transaction` in the edge case where there is already an OTEL transaction on the scope and the current span doesn't have that transaction as its parent.

### Proposal

In that edge case, the processor should **only set** `Scope.Transaction` **when it is currently** `null`, rather than overwriting whatever is already there. There's no perfect answer here (it depends on what the user intends, which we don't have enough context to know), but *not* overwriting an existing transaction is the more intuitive default.

### Notes

* This is intentionally **not** bundled into the getsentry/sentry-dotnet#2399 PR (which adds the opt-in `SentryOptions.AutoSetScopeTransactions`) so it can be evaluated independently.
* Related global-mode discussion: getsentry/sentry-dotnet#5313.

tsushanth:
Fixes #5314.

## Problem

`SentrySpanProcessor.OnStart` calls `ConfigureScope` unconditionally:

```csharp
_hub.ConfigureScope(static (scope, transaction) => scope.Transaction = transaction, transaction);
```

When two root OTel activities overlap and neither is a child of the other, this silently overwrites whatever transaction was already on the scope with the new one. As the issue notes, the intuitive default is to _not_ overwrite an existing transaction.

## Fix

Add a null-check guard inside the `ConfigureScope` lambda:

```csharp
_hub.ConfigureScope(static (scope, transaction) =>
{
    if (scope.Transaction is null)
    {
        scope.Transaction = transaction;
    }
}, transaction);
```

`Scope.Transaction` is only set when it is currently `null`; an already-present transaction is left unchanged.

## Test

Added `OnStart_WithExistingTransactionOnScope_DoesNotOverwriteExistingTransaction` to `SentrySpanProcessorTests`: sets a transaction on the scope, starts a new root activity, calls `OnStart`, and asserts the scope still holds the original transaction.

This is intentionally **not** bundled with #2399 (the `AutoSetScopeTransactions` opt-in), per the issue's notes.

### DIFF
--- a/src/Sentry.OpenTelemetry/SentrySpanProcessor.cs
+++ b/src/Sentry.OpenTelemetry/SentrySpanProcessor.cs
@@ -172,7 +172,15 @@ public class SentrySpanProcessor : BaseProcessor<Activity>
             tracer.Contexts.Trace.Origin = OpenTelemetryOrigin;
             tracer.StartTimestamp = data.StartTimeUtc;
         }
-        _hub.ConfigureScope(static (scope, transaction) => scope.Transaction = transaction, transaction);
+        // Only set Scope.Transaction when null to avoid clobbering an existing transaction
+        // when two root OTel activities overlap. See #5314.
+        _hub.ConfigureScope(static (scope, transaction) =>
+        {
+            if (scope.Transaction is null)
+            {
+                scope.Transaction = transaction;
+            }
+        }, transaction);
         transaction.SetFused(data);
         _map[data.SpanId] = transaction;
     }
--- a/src/Sentry.OpenTelemetry/SentrySpanProcessor.cs
+++ b/src/Sentry.OpenTelemetry/SentrySpanProcessor.cs
@@ -172,14 +172,9 @@ public class SentrySpanProcessor : BaseProcessor<Activity>
             tracer.Contexts.Trace.Origin = OpenTelemetryOrigin;
             tracer.StartTimestamp = data.StartTimeUtc;
         }
-        // Only set Scope.Transaction when null to avoid clobbering an existing transaction
-        // when two root OTel activities overlap. See #5314.
         _hub.ConfigureScope(static (scope, transaction) =>
         {
-            if (scope.Transaction is null)
-            {
-                scope.Transaction = transaction;
-            }
+            scope.Transaction ??= transaction;
         }, transaction);
         transaction.SetFused(data);
         _map[data.SpanId] = transaction;
--- a/test/Sentry.OpenTelemetry.Tests/SentrySpanProcessorTests.cs
+++ b/test/Sentry.OpenTelemetry.Tests/SentrySpanProcessorTests.cs
@@ -1012,4 +1012,29 @@ public class SentrySpanProcessorTests : ActivitySourceTests
         description.Should().Be("POST https://example.com/foo");
         source.Should().Be(TransactionNameSource.Custom);
     }
+
+    [Fact]
+    public void OnStart_WithExistingTransactionOnScope_DoesNotOverwriteExistingTransaction()
+    {
+        // Arrange
+        _fixture.Options.Instrumenter = Instrumenter.OpenTelemetry;
+        _fixture.Options.TracesSampleRate = 1.0;
+        _fixture.ScopeManager = Substitute.For<IInternalScopeManager>();
+        var scope = new Scope();
+        var clientScope = new KeyValuePair<Scope, ISentryClient>(scope, _fixture.Client);
+        _fixture.ScopeManager.GetCurrent().Returns(clientScope);
+        var sut = _fixture.GetSut();
+
+        using var firstActivity = Tracer.StartActivity("Existing");
+        sut.OnStart(firstActivity!);
+        var existingTransaction = scope.Transaction;
+
+        // Act: second root activity starts while Existing is still on the scope
+        using var secondActivity = Tracer.StartActivity("NewRootActivity");
+        sut.OnStart(secondActivity!);
+
+        // Assert
+        scope.Transaction.Should().BeSameAs(existingTransaction,
+            "CreateRootSpan should not overwrite an already-set Scope.Transaction");
+    }
 }
--- a/test/Sentry.OpenTelemetry.Tests/SentrySpanProcessorTests.cs
+++ b/test/Sentry.OpenTelemetry.Tests/SentrySpanProcessorTests.cs
@@ -1016,21 +1016,31 @@ public class SentrySpanProcessorTests : ActivitySourceTests
     [Fact]
     public void OnStart_WithExistingTransactionOnScope_DoesNotOverwriteExistingTransaction()
     {
+        // Arrange
         _fixture.Options.Instrumenter = Instrumenter.OpenTelemetry;
-        _fixture.Options.TracesSampleRate = 1.0;
         var sut = _fixture.GetSut();
 
-        using var firstActivity = Tracer.StartActivity("Existing");
-        sut.OnStart(firstActivity!);
-        Assert.True(sut._map.TryGetValue(firstActivity.SpanId, out var firstTransaction));
+        var parentContext = new TransactionContext("Program", "Main")
+        {
+            Instrumenter = Instrumenter.OpenTelemetry,
+        };
+        var parent = _fixture.Hub.StartTransaction(parentContext);
+        _fixture.Hub.ConfigureScope(scope => scope.Transaction = parent);
 
-        using var secondActivity = Tracer.StartActivity("NewRootActivity");
-        sut.OnStart(secondActivity!);
+        using var data = Tracer.StartActivity("TestActivity");
+
+        // Act
+        sut.OnStart(data!);
+
+        // Assert
+        data.ParentSpanId.Should().Be(default(ActivitySpanId), $"{nameof(data)} should be a new root Activity, not a child Activity");
+        Assert.True(sut._map.TryGetValue(data.SpanId, out var span));
+        span.Should().BeOfType<TransactionTracer>($"{nameof(data)} should be a new root Transaction, not a child Span");
 
         object scopeTransaction = null;
         _fixture.Hub.ConfigureScope(scope => scopeTransaction = scope.Transaction);
 
-        scopeTransaction.Should().BeSameAs(firstTransaction,
+        scopeTransaction.Should().BeSameAs(parent,
             "CreateRootSpan should not overwrite an already-set Scope.Transaction");
     }
 }
--- a/test/Sentry.OpenTelemetry.Tests/SentrySpanProcessorTests.cs
+++ b/test/Sentry.OpenTelemetry.Tests/SentrySpanProcessorTests.cs
@@ -1016,25 +1016,21 @@ public class SentrySpanProcessorTests : ActivitySourceTests
     [Fact]
     public void OnStart_WithExistingTransactionOnScope_DoesNotOverwriteExistingTransaction()
     {
-        // Arrange
         _fixture.Options.Instrumenter = Instrumenter.OpenTelemetry;
         _fixture.Options.TracesSampleRate = 1.0;
-        _fixture.ScopeManager = Substitute.For<IInternalScopeManager>();
-        var scope = new Scope();
-        var clientScope = new KeyValuePair<Scope, ISentryClient>(scope, _fixture.Client);
-        _fixture.ScopeManager.GetCurrent().Returns(clientScope);
         var sut = _fixture.GetSut();
 
         using var firstActivity = Tracer.StartActivity("Existing");
         sut.OnStart(firstActivity!);
-        var existingTransaction = scope.Transaction;
+        Assert.True(sut._map.TryGetValue(firstActivity.SpanId, out var firstTransaction));
 
-        // Act: second root activity starts while Existing is still on the scope
         using var secondActivity = Tracer.StartActivity("NewRootActivity");
         sut.OnStart(secondActivity!);
 
-        // Assert
-        scope.Transaction.Should().BeSameAs(existingTransaction,
+        object scopeTransaction = null;
+        _fixture.Hub.ConfigureScope(scope => scopeTransaction = scope.Transaction);
+
+        scopeTransaction.Should().BeSameAs(firstTransaction,
             "CreateRootSpan should not overwrite an already-set Scope.Transaction");
     }
 }


test/Sentry.OpenTelemetry.Tests/SentrySpanProcessorTests.cs
```diff
@@ -1012,4 +1012,26 @@ public void ParseOtelSpanDescription_HttpClient()
...
+
+        var existing = _fixture.Hub.StartTransaction("Existing", "op.existing");
+        _fixture.Hub.ConfigureScope(scope => scope.Transaction = existing);
+
+        // Start a new root activity (no parent → would normally set scope.Transaction)
+        using var data = Tracer.StartActivity("NewRootActivity");
+        sut.OnStart(data!);
+
+        // Assert - the pre-existing transaction must still be on the scope
+        ITransaction? scopeTransaction = null;
```
Flash0ver: issue: test was neither run nor built

- there is no `ITransaction` type
- this project does not have _Nullable Reference Types_ enabled

test/Sentry.OpenTelemetry.Tests/SentrySpanProcessorTests.cs
```diff
@@ -1012,4 +1012,26 @@ public void ParseOtelSpanDescription_HttpClient()
         description.Should().Be("POST https://example.com/foo");
         source.Should().Be(TransactionNameSource.Custom);
     }
+
+    [Fact]
+    public void OnStart_WithExistingTransactionOnScope_DoesNotOverwriteExistingTransaction()
```
Flash0ver, line 1017: issue: test is ineffective

Currently, this test does not actually cover the modification, and does not cover the modified method `SentrySpanProcessor.CreateRootSpan`.

test/Sentry.OpenTelemetry.Tests/SentrySpanProcessorTests.cs
```diff
@@ -1012,4 +1012,26 @@ public void ParseOtelSpanDescription_HttpClient()
...
+    {
+        // Arrange - simulate a root OTel activity starting while another transaction is
+        // already on the scope (e.g. two overlapping root activities). The processor should
+        // not overwrite the existing transaction (see GH-5314).
```
Flash0ver: thought: comments don't add additional value

IMO, the comments in this test apart from `Arrange/Assert/Assert` don't really add any additional value:
- Assert: the existing transaction is called `"Existing"`
- Act: the New Root Activity that is started is called `"NewRootActivity"`
  - also: actually, `Act` is missing for the Act portion of the test
- Assert: the `because` string parameter to the assertion describes

To me, the combination of Test-Name + Test-Logic + String-Parameters already describe what this test is covering and asserting.

src/Sentry.OpenTelemetry/SentrySpanProcessor.cs
```diff
@@ -172,7 +172,17 @@ private void CreateRootSpan(Activity data)
             tracer.Contexts.Trace.Origin = OpenTelemetryOrigin;
             tracer.StartTimestamp = data.StartTimeUtc;
         }
-        _hub.ConfigureScope(static (scope, transaction) => scope.Transaction = transaction, transaction);
+        // Only set Scope.Transaction when it is currently null. Overwriting an existing
+        // transaction (which can happen when two root OTel activities overlap and neither is
+        // a child of the other) silently drops the previous transaction's scope context.
+        // See https://github.com/getsentry/sentry-dotnet/issues/5314
```
Flash0ver: suggestion: shorten comment

Although this comment does add value ... I find it a bit lengthy.
Perhaps we can shorten it a bit: for example: `Only set Scope.Transaction when it is currently null` is just describing the C# syntax in use.

Flash0ver: question: do we want to add links to issues more aggressively? @jamescrosswell

Previously, we only added links to the respective GitHub issue, when the code isn't too obvious what it's trying to achieve in a rather complex flow ... or when the solution in place should be reviewed/revisited at some point in time when it is a workaround.

But it seems that Coding Agents like to add both a summary from the GitHub issue, as well as a link to the GitHub issue, and additionally describe the code semantics ahead.

In this case, I would prefer to either have a one-liner, or two/three words + link to the GitHub issue ... but both - to me - is a bit lengthy. To be fair, I might exaggerate "the problem" here a bit. What do you think?

Flash0ver (changes requested):

Flash0ver:
Thank you very much, @tsushanth, for your contribution to the Sentry .NET SDK!
I left some comments about the test-case, and commented on the comments 😉

test/Sentry.OpenTelemetry.Tests/SentrySpanProcessorTests.cs
Flash0ver: @jamescrosswell I wonder if Mutation Testing could cover such issues early ... e.g. if we only run mutation tests on the changes per PR, and only via the test cases that cover the `/src/` changes of the PR.

src/Sentry.OpenTelemetry/SentrySpanProcessor.cs
jamescrosswell: I think we could probably remove the comment entirely in this case... it's pretty apparent from the code what's going on.

jamescrosswell: We get this for free from git blame (will identify the specific commit - relatively trivial to find the PR from there with a squash and merge as we're doing). When the code is particularly unintuitive and a full explanation in code comments isn't practical, it's worth linking to a PR  description that provides a full explanation (make it clear we  expect people will have to read the PR description to fully 'get it' in those cases).

For  this piece of code, I  don't think it's worth it though. We just don't want to overwrite previous/manually set values on the scope, which is obvious from the code.

src/Sentry.OpenTelemetry/SentrySpanProcessor.cs
```diff
@@ -172,7 +172,17 @@ private void CreateRootSpan(Activity data)
...
+        _hub.ConfigureScope(static (scope, transaction) =>
+        {
+            if (scope.Transaction is null)
+            {
+                scope.Transaction = transaction;
+            }
```
jamescrosswell:
```suggestion
            scope.Transaction ??= transaction;
```

src/Sentry.OpenTelemetry/SentrySpanProcessor.cs
```diff
@@ -172,7 +172,17 @@ private void CreateRootSpan(Activity data)
...
+        }, transaction);
```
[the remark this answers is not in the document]
jamescrosswell: This isn't a problem...  scope.Transaction isn't supposed to (nor can it) track every active root activity. What scope.Transaction really/ultimately determines is the parent span for any automated instrumentation (e.g. DB spans that Sentry's integrations create automatically). SDK users can use whatever (alternate) mechanism they like to track what they consider to be the 'current' activity in a particular context (e.g. the context of their background worker or a particular thread or whatever).

test/Sentry.OpenTelemetry.Tests/SentrySpanProcessorTests.cs
jamescrosswell: That might be nice. It has to save us more effort than we spend maintaining CI to automate it. In the case of mutation tests that seems likely. Let's add an issue for prioritisation.
- https://github.com/getsentry/sentry-dotnet/issues/5343

test/Sentry.OpenTelemetry.Tests/SentrySpanProcessorTests.cs
jamescrosswell: Plus the test itself is called `OnStart_WithExistingTransactionOnScope_DoesNotOverwriteExistingTransaction`...

tsushanth:
Thanks for the detailed review @Flash0ver! Updated:

- **Source comment**: shortened to a two-liner with issue link
- **Test**: rewritten to use `ScopeManager`/`Scope` directly (same pattern as `OnStart_SampledWithoutParentSpanId_StartsNewTransaction`), so it actually exercises `CreateRootSpan`. First activity sets scope.Transaction via the processor; second root activity is the act; assertion checks scope.Transaction is still the first
- **Types**: removed `ITransaction` (doesn't exist here); scope.Transaction is tested via `BeSameAs` on the captured reference
- **Inline comments**: removed the narrative comments, kept only Arrange/Act/Assert markers

test/Sentry.OpenTelemetry.Tests/SentrySpanProcessorTests.cs
tsushanth: Fixed — rewrote the test to use the real `Hub` instead of a mock `IInternalScopeManager`. The mock was eating `ConfigureScope` calls so the guard never ran. The test now captures the first transaction from `sut._map` and verifies `scope.Transaction` still points to it after the second `OnStart`.

test/Sentry.OpenTelemetry.Tests/SentrySpanProcessorTests.cs
tsushanth: Addressed in the same rewrite — the test now goes through the real `Hub`'s `ConfigureScope`, so the null guard in `CreateRootSpan` actually runs.

test/Sentry.OpenTelemetry.Tests/SentrySpanProcessorTests.cs
tsushanth: Removed the inline comments from the test.

src/Sentry.OpenTelemetry/SentrySpanProcessor.cs
tsushanth: Removed.

tsushanth: Agreed, removed.

src/Sentry.OpenTelemetry/SentrySpanProcessor.cs
tsushanth: Applied.

jamescrosswell: Thanks @tsushanth - did you forget to push your changes to remote maybe? I don't see them here...

tsushanth:
Sorry about the delay — was pushing to the wrong local branch name. Commits are now on the correct branch: `??=` operator in the source, and the test rewritten to use the real Hub (not a mock `IInternalScopeManager`) so `ConfigureScope` actually executes the null guard.

Flash0ver:
thanks ... I'll take another look in the morning (CEST)

test/Sentry.OpenTelemetry.Tests/SentrySpanProcessorTests.cs
```diff
@@ -1012,4 +1012,25 @@ public void ParseOtelSpanDescription_HttpClient()
...
+        Assert.True(sut._map.TryGetValue(firstActivity.SpanId, out var firstTransaction));
+
+        using var secondActivity = Tracer.StartActivity("NewRootActivity");
+        sut.OnStart(secondActivity!);
+
+        object scopeTransaction = null;
+        _fixture.Hub.ConfigureScope(scope => scopeTransaction = scope.Transaction);
+
+        scopeTransaction.Should().BeSameAs(firstTransaction,
+            "CreateRootSpan should not overwrite an already-set Scope.Transaction");
```
Flash0ver, line 1044: issue: does not cover the "Scope-Transaction already set" branch

I'm afraid I'm a stickler with the test effectiveness:
Although we do now invoke `CreateRootSpan`, and do cover `ConfigureScope`, the branch where `Scope.Transaction` is not overwritten because it's already set is actually not covered.

```suggestion
        // Arrange
        _fixture.Options.Instrumenter = Instrumenter.OpenTelemetry;
        var sut = _fixture.GetSut();

        var parentContext = new TransactionContext("Program", "Main")
        {
            Instrumenter = Instrumenter.OpenTelemetry,
        };
        var parent = _fixture.Hub.StartTransaction(parentContext);
        _fixture.Hub.ConfigureScope(scope => scope.Transaction = parent);

        using var data = Tracer.StartActivity("TestActivity");

        // Act
        sut.OnStart(data!);

        // Assert
        data.ParentSpanId.Should().Be(default(ActivitySpanId), $"{nameof(data)} should be a new root Activity, not a child Activity");
        Assert.True(sut._map.TryGetValue(data.SpanId, out var span));
        span.Should().BeOfType<TransactionTracer>($"{nameof(data)} should be a new root Transaction, not a child Span");

        object scopeTransaction = null;
        _fixture.Hub.ConfigureScope(scope => scopeTransaction = scope.Transaction);

        scopeTransaction.Should().BeSameAs(parent,
            "CreateRootSpan should not overwrite an already-set Scope.Transaction");
```

I also suggested to bring back the `Arrange-Act-Assert` comments, and group all assertions at the end, to be consistent with most of our test cases.
And also suggesting to make the test a bit more consistent with the existing tests in this document.

Flash0ver (changes requested):
One final suggestion ... to make the test actually cover the branch where `scope.Transaction` is already set AND `_hub.ConfigureScope` is invoked.
(I took the liberty to apply the change and re-run the checks/workflows)

test/Sentry.OpenTelemetry.Tests/SentrySpanProcessorTests.cs
Flash0ver: applied ... now, the test does actually fail without the `/src/` code changes ... this makes a slightly stringer case for #5343 too

Flash0ver (approved):

jamescrosswell (approved):

jamescrosswell:
Thank you very much for the PR and your patience working through the review process with us @tsushanth ❤️

PR #5356, opened 2026-07-08
perf: avoid heap allocation via local `System.Lazy` instance

Flash0ver:
### Summary

Avoid a heap allocation via a local [System.Lazy](https://learn.microsoft.com/dotnet/api/system.lazy-1) instance.

### Remarks

In #5159, we introduced a local `System.Lazy` instance in favor of readability and maintainability. This change is introducing a non-thread-safe `struct`-based Lazy variant to have your cake and eat it too.

#### Is this new type worth it
I'd argue: yes ... to keep readability of the usage of `PreferTransactionNameProvider` high, yet avoid `40 B` of heap allocations per traced request, which can add up in busy web services

#### Naming (is hard)
I initially went with `LazyLite`, to align with `ConcurrentBagLite` and `ConcurrentQueueLite`. But arguably, both `ConcurrentBagLite` and `ConcurrentQueueLite` are replacements for `ConcurrentBag` and `ConcurrentQueue` respectively, where `LazyLite` does not replace `Lazy`, but should only be used for local variables when there is no thread contention at all.

### Changes
- move existing `BatchProcessorBenchmarks` to a new `Internal` namespace
  - the list of documents in the root of `Sentry.Benchmarks` is getting long
  - to be consistent with how `/test/` is also mimicking the respective `/src/` namespaces
- a fix to the command in `benchmarks/Sentry.Benchmarks/README.md`
- added `LazyLite` + Tests + Benchmarks
- replaced the local `Lazy` usage with `LazyLite`
  - added via #5159

---

#skip-changelog


src/Sentry/Internal/LazyLite.cs
```diff
@@ -0,0 +1,38 @@
...
+
+    public T Value => _factory is null ? _value! : CreateValue();
+
+    private T CreateValue()
+    {
+        var factory = _factory;
+        if (factory is not null)
+        {
+            _factory = null;
+            _value = factory();
```
Flash0ver, line 33: question: exception

If the factory-callback throws, we propagate the Exception, but only once ... I think `System.Lazy` captures it and re-throws it every time accessing `Value`. On the other hand, we don't really need this functionality right now.

PR #5361, opened 2026-07-09
feat: Add BeforeSendFeedback callback to inspect, modify, or drop user feedback before it's sent

### LINKED ISSUE #4092: Add `BeforeSendFeedback` callback to SentryOptions
EventProcessors should only be used for exceptions. We should add a `BeforeSendFeedback` callback to the `SentryOptions` that SDK users can wire up to do pre-processing on V2 Feedback events.

> 1. You should create a `beforeSendFeedback`. `beforeSend` is meant for errors only, 

 _Originally posted by @aliu39 in [#88232](https://github.com/getsentry/sentry/issues/88232#issuecomment-2779575605)_

vladbrincoveanu:
Refers to #4092

#### Summary

Adds a `BeforeSendFeedback` callback to `SentryOptions`, mirroring the existing `BeforeSend` pattern but scoped to user feedback events. It lets applications inspect, modify, or drop user feedback before it is sent to Sentry.

#### Changes

- Two `SetBeforeSendFeedback` overloads on `SentryOptions` (with and without `SentryHint`), consistent with `SetBeforeSend` / `SetBeforeSendTransaction`.
- `SentryEventHelper.DoBeforeSendFeedback`, invoked in `SentryClient.CaptureFeedback` **after** event processors run.
- New `CaptureFeedbackResult.DroppedByBeforeSendFeedback` result value
- Dropped feedback is recorded via `ClientReportRecorder` under `DiscardReason.BeforeSend` / `DataCategory.Feedback`.  Originally, was specified the use of DataCategory USER_REPORT_V2.
https://github.com/getsentry/sentry/issues/88232#issuecomment-2779575605
 
#### Behavior notes for reviewers

Assumption: Feedback is a common place to scrub PII, so a callback that cannot complete should not leak partially-scrubbed or unintended feedback. 
Implementation: Both the null-return and throw paths record a client-report discard. This is documented on the public API and covered by tests.

#### Tests

`SentryClientTests` covers: callback invocation + mutation passthrough, hint received from caller, null and throw drops feedback.

#### Unrelated changes
Added a small unit test for CaptureFeedbackResult.UnknownError.

### DIFF
--- a/src/Sentry/SentryOptions.cs
+++ b/src/Sentry/SentryOptions.cs
@@ -548,8 +548,15 @@ public class SentryOptions
     internal Func<SentryEvent, SentryHint, SentryEvent?>? BeforeSendFeedbackInternal => _beforeSendFeedback;
 
     /// <summary>
-    /// Configures a callback to invoke before sending user feedback to Sentry
+    /// Configures a callback function to be invoked before sending user feedback to Sentry
     /// </summary>
+    /// <remarks>
+    /// The event returned by this callback will be sent to Sentry. This allows the
+    /// application a chance to inspect and/or modify the user feedback before it's sent.
+    /// If the feedback should not be sent at all, return null from the callback. Unlike
+    /// <see cref="SetBeforeSend(Func{SentryEvent, SentryHint, SentryEvent?})"/>, an exception
+    /// thrown from this callback also drops the feedback (fail closed) rather than sending it.
+    /// </remarks>
     /// <param name="beforeSendFeedback">The callback</param>
     public void SetBeforeSendFeedback(Func<SentryEvent, SentryHint, SentryEvent?> beforeSendFeedback)
     {
@@ -557,8 +564,15 @@ public class SentryOptions
     }
 
     /// <summary>
-    /// Configures a callback to invoke before sending user feedback to Sentry
+    /// Configures a callback function to be invoked before sending user feedback to Sentry
     /// </summary>
+    /// <remarks>
+    /// The event returned by this callback will be sent to Sentry. This allows the
+    /// application a chance to inspect and/or modify the user feedback before it's sent.
+    /// If the feedback should not be sent at all, return null from the callback. Unlike
+    /// <see cref="SetBeforeSend(Func{SentryEvent, SentryEvent?})"/>, an exception thrown
+    /// from this callback also drops the feedback (fail closed) rather than sending it.
+    /// </remarks>
     /// <param name="beforeSendFeedback">The callback</param>
     public void SetBeforeSendFeedback(Func<SentryEvent, SentryEvent?> beforeSendFeedback)
     {
--- a/src/Sentry/SentryOptions.cs
+++ b/src/Sentry/SentryOptions.cs
@@ -543,6 +543,28 @@ public class SentryOptions
         _beforeSendTransaction = (transaction, _) => beforeSendTransaction(transaction);
     }
 
+    private Func<SentryEvent, SentryHint, SentryEvent?>? _beforeSendFeedback;
+
+    internal Func<SentryEvent, SentryHint, SentryEvent?>? BeforeSendFeedbackInternal => _beforeSendFeedback;
+
+    /// <summary>
+    /// Configures a callback to invoke before sending user feedback to Sentry
+    /// </summary>
+    /// <param name="beforeSendFeedback">The callback</param>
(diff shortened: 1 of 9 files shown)


src/Sentry/SentryOptions.cs
```diff
@@ -543,6 +543,30 @@ public void SetBeforeSendTransaction(Func<SentryTransaction, SentryTransaction?>
...
+    {
+        _beforeSendFeedback = beforeSendFeedback;
+    }
+
+    /// <summary>
+    /// Configures a callback to invoke before sending user feedback to Sentry.
+    /// Return null or throw to drop the feedback.
+    /// </summary>
+    /// <param name="beforeSendFeedback">The callback</param>
+    public void SetBeforeSendFeedback(Func<SentryEvent, SentryEvent?> beforeSendFeedback)
```
jamescrosswell, line 565: More of a question than a comment, but I'm not sure if we need this overload. Originally we added an overload on the BeforeSend event when we wanted to introduce a variant that accepted a SentryHint parameter without that being a breaking change... we never made a conscious decision that we wanted two overloads of the callback though. 

If we only had a single version of the callback that accepted both a SentryEvent and a SentryHint argument, users could easily ignore the second argument. 

On the other hand, if we do that now for this callback, `SetBeforeSendFeedback` is then the odd one out (inconsistent)... so probably not something we want to do in this PR but something we could consider for all of the BeforeSend* callbacks as a breaking change in the next major:
- Do we want to simplify the BeforeSend* callbacks so that these are not overloaded?

@Flash0ver @vladbrincoveanu what are your thoughts?

test/Sentry.Tests/SentryClientTests.cs
```diff
@@ -1035,6 +1035,116 @@ public void CaptureFeedback_EventDropped_SendsClientReport()
...
+
+        //Assert
+        result.Should().Be(CaptureFeedbackResult.DroppedByBeforeSendFeedback);
+        id.Should().Be(SentryId.Empty);
+        _ = sut.Worker.DidNotReceive().EnqueueEnvelope(Arg.Any<Envelope>());
+        _fixture.ClientReportRecorder.Received(1).RecordDiscardedEvent(DiscardReason.BeforeSend, DataCategory.Feedback);
+    }
+
+    [Fact]
+    public void CaptureFeedback_EnqueueFails_ReturnsUnknownError()
```
jamescrosswell, line 1133: Seems unrelated to this PR but harmless...

jamescrosswell (approved):
Thanks for the PR @vladbrincoveanu ❤️ 

This looks all good to me. Nothing blocking from my side - I did have one question for you that might help guide the direction of these APIs in the future though.

Flash0ver:
> Refers to #4092

"Refers" is not a verb that links a PR to an issue.

GitHub links PRs to an issue by using any of these keywords:

```
close
closes
closed
fix
fixes
fixed
resolve
resolves
resolved
```

See https://docs.github.com/en/issues/tracking-your-work-with-issues/using-issues/linking-a-pull-request-to-an-issue

src/Sentry/SentryOptions.cs
vladbrincoveanu: These overloads are just adapters. They just use the SentryHint version and their docs hide half the feature.
I agree, we can batch mark them as [Obsolete] for a next minor and remove them in the future major, so users gets some warnings first. This could be another good starter ticket, for cleanups.

3 concerns:
- the overload without SentryHint is the one used in docs, we need to updated the entries too.
- by marking them as obsolete first in a minor vers, the consumers either get a warning or their build can actually break if they have _TreatWarningsAsErrors_ set up. Is this acceptable?
- _SetBeforeSendLog_ and _SetBeforeSendMetric_ do not have 2 overloads and we need to decide if it makes sense to add a SentryHint for logging and metrics.

1 suggestion:
- another possible doc update: new data for Before callbacks will go into SentryHint, not into a new delegate parameter, so we enforce it. Ofc, if we ever need to add a cancellationtoken or async, we will have a new overload.

jamescrosswell: > we can batch mark them as [Obsolete] for a next minor and remove them in the future major

Probably more effort that it's worth... The overloads don't add too much noise.

> SetBeforeSendLog and SetBeforeSendMetric do not have 2 overloads and we need to decide if it makes sense to add a SentryHint for logging and metrics.

The main reason for the Hint parameter was to make it possible to pass in attachments... which I don't think make sense for metrics/logs (neither support attachments).

PR #5365, opened 2026-07-09
feat: sync `environment` to the native layers

### LINKED ISSUE #5385: Sync `Environment` via `IScopeObserver` in Hybrid SDK scenarios
Tracking issue for :
- #5365
- #5376

## Problem 

Per #5376: the SDK has drifted into keeping the `Scope` largely unpopulated and relying on the `Enricher` to lazily backfill values like `Environment`/`Release`, plus a *second* pipeline (`SetDefaultAttributes`) living outside both mechanisms... meaning a user-set `Environment` on the scope could simply be ignored. 

1. **At Init:** `SentryOptions.Environment` isn't being pushed into the native SDK, so events originating natively (e.g. a native crash) could carry the wrong/default environment while managed events carried the user's value — inconsistent tagging on the same app.
2. **After Init:** The scope lets a user overwrite `Environment` at runtime, but that change isn't propagated. 

> [!NOTE]
>  This issue relates to `sentry-native` and `sentry-cocoa` only. `sentry-java` is addressed separately via:
> - https://github.com/getsentry/sentry-dotnet/issues/5387

bitsandfoxes:
Closes #5385

The scope exposes `Environment` to be overwritten after initialization by the user. This needs to be propagated to the respective native layers. API available on
- `sentry-native` ✅ 
- `sentry-cocoa` ✅ 
- `sentry-java` needs to be followed up on.

We need to make sure that whatever it is that the SDK will end up setting on an event is also available on the native layer. The `Enricher` lazily guarantees that the context is getting populated, so clearing the Scope needs to restore the native scope to the fallback.

Initially, I thought we needed to have the `Scope` fetch and set the `Environment` from the options on the backing field in the constructor. The idea was that we'd avoid any duplicate scope sync. But this would only be an issue on non-global-mobe platforms, and those don't have scope sync in the first place.
So we can "get away" with simply having `Environment` be a smarter property.

### Changelog Entry

The `Environment` set on the `Scope` now gets synchronized to the native layers (`sentry-cocoa` and `sentry-native`)


src/Sentry/IScopeObserver.cs
```diff
@@ -35,6 +35,11 @@ public interface IScopeObserver
     /// </summary>
     public void SetTrace(SentryId traceId, SpanId parentSpanId);
 
+    /// <summary>
+    /// Sets the environment.
+    /// </summary>
+    public void SetEnvironment(string? environment);
```
Flash0ver, line 41: thought: `interface` vs `abstract class`

Making the ScopeObserver an `interface`, rather than an `abstract class`, is a design decision that is haunting us now: we continuously add new (non-default) members to it in minor versions, which is breaking user code when implementing their own ScopeObserver ... arguably the ScopeObserver it a more niche use case .. but source breaking nonetheless.

I wonder if an `abstract class` would have alleviated this problem, with which we could add new `virtual` members with an empty implementation, that user code might then `override`, but does not have to.

Changing this `interface` into an `abstract class` now would be even more unexpectedly breaking. But something to consider for every `public interface` that we expose in the future.

src/Sentry/Scope.cs
```diff
@@ -144,8 +144,25 @@ public SentryUser User
     /// <inheritdoc />
     public string? Distribution { get; set; }
 
+    private string? _environment;
```
Flash0ver: nitpick: could use `field` keyword instead of explicit backing field

If we don't use the `_environment` backing field anywhere outside of the `Environment` property accessors, with C# 14.0, we could remove that explicit backing field, and use "`field` backed properties" instead.

See https://learn.microsoft.com/dotnet/csharp/whats-new/csharp-14#the-field-keyword

src/Sentry/IScopeObserver.cs
jamescrosswell: We could make the change in the next major version... I think it's a good suggestion. We can mock/substitute abstract classes just as well as interfaces.

bitsandfoxes: I understand the "technically source breaking" argument but we have this discussion every time I make changes to the interface. The thing should go being `internal` in the next major and we'd be done with it.

For what it's worth we're not really breaking user code. This is a hypothetical. The `ScopeObserver` is public because when we added it we had no other mechanism to make it available to `sentry-unity` and not because it's some load-bearing functionality we're expecting users to use in the first place.

Flash0ver: Can / should we make `IScopeObserver` `internal`,
since we're exposing this feature to user code:
https://github.com/getsentry/sentry-dotnet/blob/ebef6ff26030703420a20931174fad325064b621/src/Sentry/SentryOptions.cs#L86-L95

Users might today already provide their own Scope-Observer.

The `AndroidScopeObserver` and `CocoaScopeObserver` do wrap any Options/User-provided `IScopeObserver`.
But the `NativeScopeObserver` does not, and just overwrites any user-provided `IScopeObserver`.

So we do have a bit of an incomplete story here.

bitsandfoxes: The wrapping is happening because you guys asked me to respect any potentially added ScopeObserver since it is public. Fair enough. But that should not be a precedence of us having to support this or not change it. We have a very limited set of hooks that we want to support on the SDK, right? Where users are supposed to filter, mutate or drop data/events, i.e. in `Init` and all those `Before_X` callbacks, the `X_Samplers`. The `ScopeObserver` is an outlier built out of necessity. 
Like what problem is this supposed to solve for the user? How is the user supposed to use this? They want to receive a callback when they are setting something (but not everything) on the scope?

We can continue weighing the pros and cons of dropping it in https://github.com/getsentry/sentry-dotnet/issues/4529

Flash0ver: sounds good ... I actually didn't quite intend to spawn off a heavy discussion on this ... just wanted to note, that for the next `public interface` we design, we should consider making it an `abstract class` if it's going to "evolve"

jamescrosswell: There are other examples of APIs that we make public, purely to make them accessible to Unity/Godot. e.g. `SdkVersion.Name` and `SdkVersion.Version` both have public setters for this reason.

The easiest would be to document this in code comments (hey this thing is public for Unity, but not intended to be used by SDK users - we can't promise we won't break it).

I don't know if there are any easy alternative workarounds. I don't think we can easily add the Unity SDK to the InternalsVisibleTo list right (from memory, Unity packages aren't and can't be signed - but my memory is terrible)?

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 ### Features ✨
 
 - feat: Add exponential backoff and log deduplication to Spotlight transport by @mattico in [#5025](https://github.com/getsentry/sentry-dotnet/pull/5025)
+- feat: The `Environment` set on the `Scope` now gets synchronized to the native layers (`sentry-cocoa` and `sentry-native`) by @bitsandfoxes [#5365](https://github.com/getsentry/sentry-dotnet/pull/5365)
```
Flash0ver: issue: don't modify `CHANGELOG.md` manually

Please add a "friendly" Changelog Entry via a `### Changelog Entry` section in the PR's description.

We just merged updated guidance via #5369
See also: https://craft.sentry.dev/configuration/#custom-changelog-entries-from-pr-descriptions

src/Sentry/SentrySdk.cs
```diff
@@ -63,6 +63,9 @@ internal static IHub InitHub(SentryOptions options)
 #pragma warning restore 0162
 #pragma warning restore CS0162 // Unreachable code detected
 
+        // This happens before the native SDKs get initialized
+        options.Environment = options.SettingLocator.GetEnvironment();
```
Flash0ver, line 67: note: not only for _Native_, but also for _Android_ and _Apple_ platforms during **Init**

This now sets the `Environment` on the _Native_ Options:
https://github.com/getsentry/sentry-dotnet/blob/cc9de07cc1f9169f48cc2e507c0ec3468cb4d8bb/src/Sentry/Platforms/Native/CFunctions.cs#L95-L99

It also sets it for _Apple_:
https://github.com/getsentry/sentry-dotnet/blob/35f0027fb84236e633d8942fb78f2f75e3e47e75/src/Sentry/Platforms/Cocoa/SentrySdk.cs#L40-L43

and _Android_ platforms:
https://github.com/getsentry/sentry-dotnet/blob/11068e07c883649830613f61a5ef5de89106d90f/src/Sentry/Platforms/Android/SentrySdk.cs#L54

See also: https://github.com/getsentry/sentry-dotnet/pull/5376#discussion_r3570485608

Flash0ver: question: Could this cause some behavioral breaking side-effects?

src/Sentry/Scope.cs
```diff
@@ -406,7 +436,7 @@ public void Clear()
         User = new();
         Release = default;
         Distribution = default;
-        Environment = default;
+        Environment = Options.Environment; // Restore to keep in sync with native
```
Flash0ver: question: native only?

For all Hybrid-SDK scenarios, isn't this keeping it in sync for all downstream SDKs? (or upstream, I always mix it up ... I guess depending from a point of view: SDK Release/Publishing -vs- Runtime Instrumentation and Data-Flow)
Something worth clarifying in this comment?

src/Sentry/Scope.cs
```diff
@@ -288,6 +315,9 @@ internal Scope(SentryOptions? options, SentryPropagationContext? propagationCont
     {
         Options = options ?? new SentryOptions();
         PropagationContext = new SentryPropagationContext(propagationContext);
+
+        // Skip scope sync with backing field. The native SDKs already received the environment set on the options.
+        _environment = Options.Environment;
```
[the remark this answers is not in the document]
Flash0ver: issue: might be unintended behavior

Similar to: https://github.com/getsentry/sentry-dotnet/pull/5376#discussion_r3570418617

```CSharp
SentrySdk.Init(options =>
{
    options.Environment = "options-environment";
});

SentrySdk.ConfigureScope(static scope =>
{
    scope.Environment = "scope-environment";
});
using var scope = SentrySdk.PushScope();
SentrySdk.ConfigureScope(static scope =>
{
    Console.WriteLine(scope.Environment); //options-environment
});
```

src/Sentry/Scope.cs
```diff
@@ -144,8 +144,35 @@ public SentryUser User
     /// <inheritdoc />
     public string? Distribution { get; set; }
 
+    private string? _environment;
+
     /// <inheritdoc />
-    public string? Environment { get; set; }
+    public string? Environment
```
Flash0ver, line 148: question: Nullability

Can `Environment` then ever be `null`?
I wonder if we might change from `string?` to `string`, together with an `[AllowNull]` attribute, as the setter might set `null`, but the getter never returns _null_.

But this is - should we consider it - something for a follow-up ... and potentially even for the next Major Release.

src/Sentry/Scope.cs
```diff
@@ -145,7 +145,32 @@ public SentryUser User
     public string? Distribution { get; set; }
 
     /// <inheritdoc />
-    public string? Environment { get; set; }
+    public string? Environment
+    {
+        get;
+        set
```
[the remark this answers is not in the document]
bitsandfoxes: This works as intended and avoids any erroneous scope syncs.

src/Sentry/Scope.cs
```diff
@@ -145,7 +145,32 @@ public SentryUser User
...
+            }
+            else
+            {
+                field = value;
+            }
+
+            if (Options is { EnableScopeSync: true, ScopeObserver: { } observer })
+            {
+                observer.SetEnvironment(field);
+            }
```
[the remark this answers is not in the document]
bitsandfoxes: The scope sync is only active on platforms that specifically have `GlobalMode` enabled.

src/Sentry/Scope.cs
```diff
@@ -145,7 +145,32 @@ public SentryUser User
...
+                field = Options.Environment;
+            }
+            else
+            {
+                field = value;
+            }
+
+            if (Options is { EnableScopeSync: true, ScopeObserver: { } observer })
+            {
+                observer.SetEnvironment(field);
```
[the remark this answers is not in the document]
bitsandfoxes: This works as intended. The `Environment` on the options used the same mechanism to retrieve the environment before even initializing.

src/Sentry/Scope.cs
bitsandfoxes: Outdated comment. Restored original state.

src/Sentry/SentrySdk.cs
bitsandfoxes: As far as I can tell this is a bug that is now fixed. 
Previously, each SDK would resolve the `environment` itself with no guarantee that those would match. Now, the .NET SDK acts as proper head-SDK, providing the native layer with the correct user configuration.

src/Sentry/Scope.cs
bitsandfoxes: Based on the `Enricher` and the way the `environment` gets resolved it can never be `null`, no.
Sounds like something for the next mayor.

src/Sentry/SentrySdk.cs
bitsandfoxes: > note: not only for Native, but also for Android and Apple platforms during Init

With `native` we refer to the SDK that is "native" to the current platform we're running on. To differentiate between the layers.

src/Sentry/Scope.cs
jamescrosswell: - #5388

jamescrosswell (approved):
LGTM - thanks @bitsandfoxes !

PR #5369, opened 2026-07-10
fix(changelog): remove stale Unreleased entry and guard against manual entries

jamescrosswell:
## Why the 6.7.0 changelog broke

The [6.7.0 release](https://github.com/getsentry/publish/issues/8841) shipped with a **single** changelog entry despite dozens of merged PRs.

Root cause: the repo generates its changelog automatically via craft (`changelogPolicy: auto` in `.craft.yml`). Craft only regenerates the `## Unreleased` section **when it is empty** ([prepare.ts](https://github.com/getsentry/craft/blob/master/src/commands/prepare.ts#L568) — `if (!changeset.body)` guards the git-based generation, and the body is [`.trim()`-ed](https://github.com/getsentry/craft/blob/master/src/utils/changelog.ts#L464) first). A stale, hand-written entry (#5025) had been left under `## Unreleased`, so craft saw non-empty content, skipped generation entirely, and just renamed `## Unreleased` → `## 6.7.0`. Everything else was silently dropped.

That stale line got there because `CONTRIBUTING.md` still instructed contributors to hand-add `## Unreleased` entries — contradicting the `auto` policy (and `AGENTS.md`/`CLAUDE.md`, which already say *do not add changelogs manually*).

## How changelog entries are meant to be produced

Under `changelogPolicy: auto`, nobody edits `CHANGELOG.md`. craft builds each entry one of two ways:

1. **From the commit message / PR title** — the default. A `feat:` PR becomes a Feature, a `fix:` PR a Fix, etc.
2. **From a `### Changelog Entry` section in the PR description** — if present, craft uses that text verbatim instead of the title, which lets a PR provide a richer or multi-line entry (or several entries). See [craft docs](https://craft.sentry.dev/configuration/#custom-changelog-entries-from-pr-descriptions).

Either way, the `## Unreleased` section must stay empty so craft regenerates it at release time.

## Changes

- **`CHANGELOG.md`** — remove the stale manual `## Unreleased` entry so craft regenerates 6.7.0 from git history (the #5025 feature is still captured — it's a `feat:` commit in range).
- **`.github/workflows/changelog-guard.yml` + `scripts/verify-changelog.sh`** — a `Changelog Guard` check that fails any PR leaving content under `## Unreleased`. It mirrors craft's own emptiness test: **any** non-whitespace content fails (a bullet, or even a bare `### Features` sub-heading), while an empty `## Unreleased` heading passes. Reports the offending `CHANGELOG.md` line numbers. Runnable locally: `./scripts/verify-changelog.sh`.
- **`CONTRIBUTING.md`** — replace the stale "add a `## Unreleased` heading" instructions with the automatic-changelog policy, documenting both entry mechanisms above.
- **`AGENTS.md`** — add a Getting Started note to check out submodules first (`./dev.cs cleanslate`); without them, tooling like `generate-solution-filters.ps1` silently drops projects from the `*.slnf` files.
- **`Sentry.slnx` / `.generated.NoMobile.slnx`** — register the new workflow and script in the solution.

## Follow-up (not in this PR)

The 6.7.0 release itself needs to be re-cut after this lands so craft regenerates the full changelog. Consider making `Changelog Guard` a required status check.

### DIFF
--- a/.github/workflows/changelog-guard.yml
+++ b/.github/workflows/changelog-guard.yml
@@ -1,0 +1,19 @@
+name: Changelog Guard
+
+on:
+  pull_request:
+    branches:
+      - main
+      - release/**
+
+permissions:
+  contents: read
+
+jobs:
+  no-manual-unreleased-entries:
+    name: No manual "## Unreleased" entries
+    runs-on: ubuntu-24.04
+    steps:
+      - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
+      - name: Verify CHANGELOG.md has no manual "## Unreleased" entries
+        run: ./scripts/verify-changelog.sh
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -113,6 +113,8 @@ To help users see value in updating the SDK, we maintain a changelog file with e
 
 **Do not edit `CHANGELOG.md` manually.** The changelog is generated automatically at release time by [craft](https://github.com/getsentry/craft) (`changelogPolicy: auto` in `.craft.yml`) from the pull requests merged since the previous release. Entries are categorized from your [commit message / PR title](https://develop.sentry.dev/engineering-practices/commit-messages/) (e.g. a `feat:` PR becomes a Feature, a `fix:` PR becomes a Fix), so there's nothing to add by hand.
 
+If the PR title doesn't capture the change well — you want more detail, or a single PR should produce several entries — add a `### Changelog Entry` section to the PR description. craft uses the text under that heading verbatim instead of the PR title. See [Custom changelog entries from PR descriptions](https://craft.sentry.dev/configuration/#custom-changelog-entries-from-pr-descriptions).
+
 If the change isn't user-facing and you'd rather it not appear in the changelog at all, add the `skip-changelog` label to the PR or write `#skip-changelog` in the PR description.
 
 ## Naming tests
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -113,7 +113,6 @@ To help users see value in updating the SDK, we maintain a changelog file with e
 
 **Do not edit `CHANGELOG.md` manually.** The changelog is generated automatically at release time by [craft](https://github.com/getsentry/craft) (`changelogPolicy: auto` in `.craft.yml`) from the pull requests merged since the previous release. Entries are categorized from your [commit message / PR title](https://develop.sentry.dev/engineering-practices/commit-messages/) (e.g. a `feat:` PR becomes a Feature, a `fix:` PR becomes a Fix), so there's nothing to add by hand.
 
-
 If the change isn't user-facing and you'd rather it not appear in the changelog at all, add the `skip-changelog` label to the PR or write `#skip-changelog` in the PR description.
 
 ## Naming tests
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -113,7 +113,6 @@ To help users see value in updating the SDK, we maintain a changelog file with e
 
 **Do not edit `CHANGELOG.md` manually.** The changelog is generated automatically at release time by [craft](https://github.com/getsentry/craft) (`changelogPolicy: auto` in `.craft.yml`) from the pull requests merged since the previous release. Entries are categorized from your [commit message / PR title](https://develop.sentry.dev/engineering-practices/commit-messages/) (e.g. a `feat:` PR becomes a Feature, a `fix:` PR becomes a Fix), so there's nothing to add by hand.
 
-Adding an entry under `## Unreleased` yourself is actively harmful: craft only auto-generates the section when it's empty, so a single leftover manual entry suppresses generation and drops every other change from the release notes. A `Changelog Guard` GitHub Action fails any PR that adds a manual entry under `## Unreleased`; you can run the same check locally with `./scripts/verify-changelog.sh`.
 
 If the change isn't user-facing and you'd rather it not appear in the changelog at all, add the `skip-changelog` label to the PR or write `#skip-changelog` in the PR description.
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -109,28 +109,13 @@ To do that, run the build locally (i.e: `./build.sh` or `build.cmd`) and commit 
 ## Changelog
 
 We'd love for users to update the SDK everytime and as soon as we make a new release. But in reality most users rarely update the SDK.
-To help users see value in updating the SDK, we maintain a changelog file with entries split between two headings:
+To help users see value in updating the SDK, we maintain a changelog file with entries split between headings such as `### Features` and `### Fixes`.
 
-1. `### Features`
-2. `### Fixes`
+**Do not edit `CHANGELOG.md` manually.** The changelog is generated automatically at release time by [craft](https://github.com/getsentry/craft) (`changelogPolicy: auto` in `.craft.yml`) from the pull requests merged since the previous release. Entries are categorized from your [commit message / PR title](https://develop.sentry.dev/engineering-practices/commit-messages/) (e.g. a `feat:` PR becomes a Feature, a `fix:` PR becomes a Fix), so there's nothing to add by hand.
 
-We add the heading in the first PR that's adding either a feature or fixes in the current release.
-After a release, the [changelog file will contain only the last release entries](https://github.com/getsentry/sentry-dotnet/blob/main/CHANGELOG.md).
+Adding an entry under `## Unreleased` yourself is actively harmful: craft only auto-generates the section when it's empty, so a single leftover manual entry suppresses generation and drops every other change from the release notes. A `Changelog Guard` GitHub Action fails any PR that adds a manual entry under `## Unreleased`; you can run the same check locally with `./scripts/verify-changelog.sh`.
 
-When you open a PR in such cases, you need to add a heading 2 named `## Unreleased`, which is replaced during release with the version number chosen.
-Below that, you'll add heading 3 mentioned above. For example, if you're adding a feature "Attach screenshots when capturing errors on WPF", right after a release, you'd add to the changelog:
-
-```
-## Unreleased
-
-### Features
-
-- Attach screenshots when capturing errors on WPF (#PR number)
-```
-
-There's a GitHub action check to verify if an entry was added.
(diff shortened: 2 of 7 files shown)


CONTRIBUTING.md
```diff
@@ -109,28 +109,13 @@ To do that, run the build locally (i.e: `./build.sh` or `build.cmd`) and commit
...
 
-1. `### Features`
-2. `### Fixes`
+**Do not edit `CHANGELOG.md` manually.** The changelog is generated automatically at release time by [craft](https://github.com/getsentry/craft) (`changelogPolicy: auto` in `.craft.yml`) from the pull requests merged since the previous release. Entries are categorized from your [commit message / PR title](https://develop.sentry.dev/engineering-practices/commit-messages/) (e.g. a `feat:` PR becomes a Feature, a `fix:` PR becomes a Fix), so there's nothing to add by hand.
 
-We add the heading in the first PR that's adding either a feature or fixes in the current release.
-After a release, the [changelog file will contain only the last release entries](https://github.com/getsentry/sentry-dotnet/blob/main/CHANGELOG.md).
+Adding an entry under `## Unreleased` yourself is actively harmful: craft only auto-generates the section when it's empty, so a single leftover manual entry suppresses generation and drops every other change from the release notes. A `Changelog Guard` GitHub Action fails any PR that adds a manual entry under `## Unreleased`; you can run the same check locally with `./scripts/verify-changelog.sh`.
```
jamescrosswell:
```suggestion
```

CONTRIBUTING.md
```diff
@@ -109,28 +109,12 @@ To do that, run the build locally (i.e: `./build.sh` or `build.cmd`) and commit
 ## Changelog
 
 We'd love for users to update the SDK everytime and as soon as we make a new release. But in reality most users rarely update the SDK.
-To help users see value in updating the SDK, we maintain a changelog file with entries split between two headings:
+To help users see value in updating the SDK, we maintain a changelog file with entries split between headings such as `### Features` and `### Fixes`.
 
-1. `### Features`
-2. `### Fixes`
+**Do not edit `CHANGELOG.md` manually.** The changelog is generated automatically at release time by [craft](https://github.com/getsentry/craft) (`changelogPolicy: auto` in `.craft.yml`) from the pull requests merged since the previous release. Entries are categorized from your [commit message / PR title](https://develop.sentry.dev/engineering-practices/commit-messages/) (e.g. a `feat:` PR becomes a Feature, a `fix:` PR becomes a Fix), so there's nothing to add by hand.
```
Flash0ver, line 114: suggestion: mention & link `### Changelog Entry`

I'd like to also mention the manual/explicit entry via `### Changelog Entry`, and also link to the docs: https://craft.sentry.dev/configuration/#custom-changelog-entries-from-pr-descriptions,
because with more explicit and descriptive Changelog Entries, that contain more helpful details to adapt new features which does not fit into a commit message / PR title.

This way we have fixed our original problem of manual `CHANGELOG.md` entries, where we had merge conflicts all the time, while not sacrificing "fidelity"/"quality" of the CHANGELOG.

scripts/verify-changelog.sh
```diff
@@ -0,0 +1,56 @@
...
+if printf '%s\n' "$unreleased_body" | grep -Eq "$entry_re"; then
+  echo "::error file=$CHANGELOG::Manual changelog entries found under '## Unreleased'."
+  echo ""
+  echo "This repository generates its changelog automatically at release time"
+  echo "(changelogPolicy: auto in .craft.yml). A manual entry under '## Unreleased'"
+  echo "suppresses that generation and causes other changes to be dropped from the"
+  echo "release notes."
+  echo ""
+  echo "Offending line(s):"
+  printf '%s\n' "$unreleased_body" | grep -En "$entry_re" || true
```
Flash0ver: issue (non-blocking): line number incorrect

I tested:

CHANGELOG:
```
# Changelog

## Unreleased

### Features ✨

- feat: Description

### Fixes 🐛

- fix: Description

## 6.6.0
```

shows
```
Offending line(s):
4:- feat: Description
8:- fix: Description
```

The line number seems off.
I expected line 7 and line 11.

scripts/verify-changelog.sh
```diff
@@ -0,0 +1,56 @@
...
+  echo ""
+  echo "Please remove them. Your change is added to the changelog automatically,"
+  echo "based on the PR title / commit message. If it is not user-facing, add the"
```
Flash0ver: question: `### Changelog Entry`

Should we mention `### Changelog Entry` too?
https://craft.sentry.dev/configuration/#custom-changelog-entries-from-pr-descriptions

Something like
```
... or a `### Changelog Entry` section in the PR description ...
```

.github/workflows/changelog-guard.yml
```diff
@@ -0,0 +1,19 @@
+name: Changelog Guard
```
Flash0ver, line 1: suggestion: add to solution

We might add this to `Sentry.slnx` and `.generated.NoMobile.slnx`.

scripts/verify-changelog.sh
```diff
@@ -0,0 +1,56 @@
+#!/usr/bin/env bash
+#
+# Verifies that CHANGELOG.md contains no manually-added entries under an
+# "## Unreleased" heading.
+#
+# This repository generates its changelog automatically at release time via
+# craft (`changelogPolicy: auto` in .craft.yml), from the pull requests merged
+# since the previous release. craft only regenerates the section when it is
```
Flash0ver: nitpick: capitalization

```suggestion
# since the previous release. Craft only regenerates the section when it is
```

scripts/verify-changelog.sh
```diff
@@ -0,0 +1,56 @@
...
+# empty, so a single leftover manual entry under "## Unreleased" suppresses the
+# auto-generation and silently drops every other change from the release notes.
+# (This is what broke the 6.7.0 release, which shipped with a single entry.)
```
Flash0ver: question: keep note?

```
# (This is what broke the 6.7.0 release, which shipped with a single entry.)
```

Not sure if this note is helpful ... perhaps instead link to this PR for more context?

.github/workflows/changelog-guard.yml
```diff
@@ -0,0 +1,19 @@
...
+  contents: read
+
+jobs:
+  no-manual-unreleased-entries:
+    name: No manual "## Unreleased" entries
+    runs-on: ubuntu-24.04
+    steps:
+      - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
+      - name: Verify CHANGELOG.md has no manual "## Unreleased" entries
+        run: ./scripts/verify-changelog.sh
```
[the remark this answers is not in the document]
jamescrosswell: Not a real issue. `scripts/verify-changelog.sh` is committed with mode `100755` (`git ls-files --stage` shows the executable bit set), so `./scripts/verify-changelog.sh` runs fine. The `Changelog Guard` job has also already run green on this PR, which it could not have done if the script lacked execute permission.

scripts/verify-changelog.sh
```diff
@@ -0,0 +1,56 @@
...
+  }
+  in_section { print }
+' "$CHANGELOG")"
+
+# A manual changelog entry is a bullet line ("- ..." or "* ..."). This
+# deliberately ignores an empty "## Unreleased" heading and blank/sub-heading
+# lines, so only real entries fail the check.
+entry_re='^[[:space:]]*[-*][[:space:]]+[^[:space:]]'
+
+if printf '%s\n' "$unreleased_body" | grep -Eq "$entry_re"; then
```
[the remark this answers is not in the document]
jamescrosswell: Good catch — this is a real gap. craft trims the section body and skips generation on any non-empty `changeset.body` ([changelog.ts#L464](https://github.com/getsentry/craft/blob/master/src/utils/changelog.ts#L464) + [prepare.ts#L568](https://github.com/getsentry/craft/blob/master/src/commands/prepare.ts#L568)), so a heading-only leftover would indeed slip past a bullet-only check. Fixed in c5771de5: the guard now fails on any non-whitespace content under `## Unreleased` (a bare/empty heading still passes), matching craft's behaviour exactly.

CONTRIBUTING.md
jamescrosswell: Good call. Added [739de6c](https://github.com/getsentry/sentry-dotnet/pull/5369/commits/739de6c2658d9343338b64cbd4fb342443ff9aa1)

scripts/verify-changelog.sh
jamescrosswell: Done in `739de6c2` — the guard's failure message now also mentions the `### Changelog Entry` PR-description option alongside the PR title / commit message.

scripts/verify-changelog.sh
jamescrosswell:
```suggestion
```

.github/workflows/changelog-guard.yml
jamescrosswell: Done in `cb6a8072` — added `.github/workflows/changelog-guard.yml` and `scripts/verify-changelog.sh` to `Sentry.slnx`, and regenerated `.generated.NoMobile.slnx` via `scripts/generate-solution-filters.ps1` so the two stay in sync.

scripts/verify-changelog.sh
```diff
@@ -0,0 +1,57 @@
...
+  echo "::error file=$CHANGELOG::The '## Unreleased' section is not empty."
+  echo ""
+  echo "This repository generates its changelog automatically at release time"
+  echo "(changelogPolicy: auto in .craft.yml). craft only regenerates the"
+  echo "'## Unreleased' section when it is empty, so ANY leftover content there"
+  echo "(entries or even a bare sub-heading) suppresses generation and causes the"
+  echo "rest of the release notes to be dropped."
+  echo ""
+  echo "Offending line(s):"
+  printf '%s\n' "$unreleased_body" | grep -En '[^[:space:]]' || true
```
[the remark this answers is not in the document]
jamescrosswell: Fixed in `b9cdf5f5`. The check now finds the offending lines in a single awk pass and emits real file line numbers (`NR`) instead of grepping the extracted body, so the reported locations point at the right lines in `CHANGELOG.md`.

scripts/verify-changelog.sh
jamescrosswell: Fixed in [b9cdf5f](https://github.com/getsentry/sentry-dotnet/pull/5369/commits/b9cdf5f54e34e9b7c1ca62d3c6334b16f226ecd3)

Flash0ver (approved):
🤘

Issue #5197, opened 2026-05-06
Move setup code to GlobalRootScopeIntegration (v7)


jamescrosswell:
> We could potentially set more stuff here. There are lots of scope properties that never change and which we currently set on the scope every time we capture an event. These are all candidates, I think:
> https://github.com/getsentry/sentry-dotnet/blob/f50b36095ca58d50837602fba76372311684a779/src/Sentry/Scope.cs#L498-L517
> 
> 
> More of a performance improvement but might result in a subtle behavioural change for some users (perhaps code that checks/expects things to not be set in certain circumstances and sets them conditionally on this basis ). To be safe, perhaps we delay a change like that until the next major release. 

 _Originally posted by @jamescrosswell in [#5039](https://github.com/getsentry/sentry-dotnet/pull/5039/changes#r2957195128)_

Some of these things can currently be set via the SentryOptions:
https://github.com/getsentry/sentry-dotnet/blob/55b2d5dcc829c078d22637a5f7ffc0efb13cc196/src/Sentry/SentrySdk.cs#L67-L68

But that in itself is confusing since now it's hard to tell which one should win (Scope or Options). 

See:
- https://github.com/getsentry/sentry-dotnet/pull/5376/changes#r3559383552

And it seems we're always copying them from the options into the Scope anyway 😕.

It might be easier if we removed those things from the options and just configured them in a root scope (or allowed the user to configure them in a root scope)... so there was only ever one place these were configured (on the scope).

PR #5376, opened 2026-07-10
fix: fetch default attributes from `Scope` first

bitsandfoxes:
Follows #5365 

The issue I see is that 
- We keep the `Scope` basically unpopulated
- We let users overwrite things on the `Scope`, i.e. `Environment`, or `Release`
- To make sure certain stuff is there we let the `Enricher` backfill (because they were never set on the scope) https://github.com/getsentry/sentry-dotnet/blob/ebef6ff26030703420a20931174fad325064b621/src/Sentry/Internal/Enricher.cs#L81
- And now we have a secondary processing pipeline with the `SetDefaultAttributes` that lives outside of either mechanism. 

Personally, I am heavily in favour of having the scope as the source of truth. Especially, since the scope is used to sync data between the C# and the native layer.
And with this change whatever is set on the scope is at least getting applied where previously, any user set `Environment` on the Scope would just get ignored.

## Behaviour before this PR

Two separate inconsistent pipelines:

**1. Events / traces (`IEventLike`) — scope wins, Options is only a fallback**

- `Scope.Environment` / `Scope.Release` are plain settable properties, default `null` ([Scope.cs:142](sentry-dotnet/src/Sentry/Scope.cs#L142), [148](sentry-dotnet/src/Sentry/Scope.cs#L148)). The constructor does **not** seed them from Options ([Scope.cs:287](sentry-dotnet/src/Sentry/Scope.cs#L287)).
- `Scope.Apply` copies scope→event only if unset: `other.Environment ??= Environment; other.Release ??= Release;` ([Scope.cs:491-493](sentry-dotnet/src/Sentry/Scope.cs#L491)).
- Then the `Enricher` backfills from Options/env-var/default, again only if still null: `eventLike.Environment ??= _options.SettingLocator.GetEnvironment();` ([Enricher.cs:75,81](sentry-dotnet/src/Sentry/Internal/Enricher.cs#L75)).

So if a user sets `scope.Environment = "Staging"`, **that wins** and Options never overrides it. Options is purely the fallback.

**2. Logs / metrics (`SetDefaultAttributes`) — scope is ignored, Options wins**

- `SetDefaultAttributes` reads straight from `SettingLocator`, never looking at the scope: `options.SettingLocator.GetEnvironment()` / `GetRelease()` ([SentryAttributes.cs:100-109](sentry-dotnet/src/Sentry/Protocol/SentryAttributes.cs#L100)).
- `SettingLocator.GetEnvironment/GetRelease` resolve `Options.Environment ?? SENTRY_ENVIRONMENT ?? default` ([SettingLocator.cs:80,105](sentry-dotnet/src/Sentry/Internal/SettingLocator.cs#L80)) — and, notably, cache the resolved value back onto Options (`_options.Environment = environment`).

So for logs, a user-set `scope.Environment` **is silently ignored** — which is exactly bitsandfoxes' complaint ("any user set Environment on the Scope would just get ignored"). That statement is true *for the logs pipeline*, not for events.

## What sentry-python and sentry-java do

Both take the **opposite** stance from "scope as source of truth":

- **sentry-python:** `Scope` has no `environment`/`release` fields. They live in client `options`. On events, `_prepare_event` fills them only if absent: `if event.get(key) is None and self.options[key] is not None: event[key] = ...` for `release`/`environment`. Logs (`set_global_attributes`) read directly from `options`. There is **no per-scope override** for these fields.
- **sentry-java:** `Scope` has no `getEnvironment/setEnvironment/getRelease/setRelease`. `MainEventProcessor` does `if (event.getRelease() == null) event.setRelease(options.getRelease());` and likewise for environment. Again, options-level only, no scope override.

In both, `environment`/`release` are effectively program-lifetime invariants set at init on options.

## Recommendation ???

If we want consistency *and* to avoid the whole clone/override bug class, the java/python model is the cleaner solution??? @bitsandfoxes what do you think of this?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Features ✨
 
 - feat: Add exponential backoff and log deduplication to Spotlight transport by @mattico in [#5025](https://github.com/getsentry/sentry-dotnet/pull/5025)
+- feat: The `Environment` set on the `Scope` now gets synchronized to the native layers (`sentry-cocoa` and `sentry-native`) by @bitsandfoxes [#5365](https://github.com/getsentry/sentry-dotnet/pull/5365)
 
 ## 6.6.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@
 - feat: Add exponential backoff and log deduplication to Spotlight transport by @mattico in [#5025](https://github.com/getsentry/sentry-dotnet/pull/5025)
 - feat: The `Environment` set on the `Scope` now gets synchronized to the native layers (`sentry-cocoa` and `sentry-native`) by @bitsandfoxes [#5365](https://github.com/getsentry/sentry-dotnet/pull/5365)
 
+### Fixes
+
+- fix: When setting `Environment` and `Release` on the `Scope` these are  now also getting applied on emitted Metrics and Logs by bitsandfoxes [#5376](https://github.com/getsentry/sentry-dotnet/pull/5376)
+
 ## 6.6.0
 
 ### Features ✨
--- a/src/Sentry/Platforms/Android/AndroidScopeObserver.cs
+++ b/src/Sentry/Platforms/Android/AndroidScopeObserver.cs
@@ -112,6 +112,18 @@ internal sealed class AndroidScopeObserver : IScopeObserver
         }
     }
 
+    public void SetEnvironment(string? environment)
+    {
+        try
+        {
+            // TODO: Missing corresponding scope-level functionality on the Android SDK
+        }
+        finally
+        {
+            _innerObserver?.SetEnvironment(environment);
+        }
+    }
+
     public void AddAttachment(SentryAttachment attachment)
     {
         try
--- a/src/Sentry/Scope.cs
+++ b/src/Sentry/Scope.cs
@@ -316,6 +316,7 @@ public class Scope : IEventLike
         Options = options ?? new SentryOptions();
         PropagationContext = new SentryPropagationContext(propagationContext);
 
+        // Skip scope sync with backing field. The native SDKs already received the environment set on the options.
         _environment = Options.Environment;
     }
 
--- a/src/Sentry/Scope.cs
+++ b/src/Sentry/Scope.cs
@@ -144,8 +144,25 @@ public class Scope : IEventLike
     /// <inheritdoc />
     public string? Distribution { get; set; }
 
+    private string? _environment;
+
     /// <inheritdoc />
-    public string? Environment { get; set; }
+    public string? Environment
+    {
+        get => _environment;
+        set
+        {
+            if (_environment != value)
+            {
+                _environment = value;
+                if (Options.EnableScopeSync &&
+                    Options.ScopeObserver is { } observer)
+                {
+                    observer.SetEnvironment(value);
+                }
+            }
+        }
+    }
 
     // TransactionName is kept for legacy purposes because
     // SentryEvent still makes use of it.
--- a/src/Sentry/Scope.cs
+++ b/src/Sentry/Scope.cs
@@ -434,7 +434,7 @@ public class Scope : IEventLike
         Request = new();
         Contexts.Clear();
         User = new();
-        Release = default;
+        Release = Options.Release;
         Distribution = default;
         Environment = Options.Environment; // Restore to keep in sync with native
         TransactionName = default;
--- a/src/Sentry/Scope.cs
+++ b/src/Sentry/Scope.cs
@@ -152,14 +152,24 @@ public class Scope : IEventLike
         get => _environment;
         set
         {
-            if (_environment != value)
+            if (_environment == value)
+            {
+                return;
+            }
+
+            if (value is null)
+            {
+                Options.LogWarning("Environment cannot be null. Reverting to default value from the options.");
+                _environment = Options.Environment;
+            }
+            else
             {
                 _environment = value;
-                if (Options.EnableScopeSync &&
-                    Options.ScopeObserver is { } observer)
-                {
-                    observer.SetEnvironment(value);
-                }
+            }
+
+            if (Options is { EnableScopeSync: true, ScopeObserver: { } observer })
+            {
+                observer.SetEnvironment(_environment);
             }
         }
     }
@@ -305,6 +315,8 @@ public class Scope : IEventLike
     {
         Options = options ?? new SentryOptions();
         PropagationContext = new SentryPropagationContext(propagationContext);
+
+        _environment = Options.Environment;
     }
 
     // For testing. Should explicitly require SentryOptions.
@@ -423,7 +435,7 @@ public class Scope : IEventLike
         User = new();
         Release = default;
         Distribution = default;
-        Environment = default;
+        Environment = Options.Environment; // Restore to keep in sync with native
         TransactionName = default;
         Transaction = default;
         Fingerprint = Array.Empty<string>();
--- a/src/Sentry/Scope.cs
+++ b/src/Sentry/Scope.cs
@@ -317,6 +317,7 @@ public class Scope : IEventLike
         PropagationContext = new SentryPropagationContext(propagationContext);
 
         _environment = Options.Environment;
+        Release = Options.Release;
     }
 
     // For testing. Should explicitly require SentryOptions.
--- a/src/Sentry/SentrySdk.cs
+++ b/src/Sentry/SentrySdk.cs
@@ -65,6 +65,7 @@ static partial class SentrySdk
 
         // This happens before the native SDKs get initialized
         options.Environment = options.SettingLocator.GetEnvironment();
+        options.Release = options.SettingLocator.GetRelease();
 
         // Initialize native platform SDKs here
         if (options.InitNativeSdks)
--- a/src/Sentry/SentrySdk.cs
+++ b/src/Sentry/SentrySdk.cs
@@ -63,6 +63,9 @@ static partial class SentrySdk
 #pragma warning restore 0162
 #pragma warning restore CS0162 // Unreachable code detected
 
+        // This happens before the native SDKs get initialized
+        options.Environment = options.SettingLocator.GetEnvironment();
+
         // Initialize native platform SDKs here
         if (options.InitNativeSdks)
         {
--- a/test/Sentry.Tests/SentryLogTests.cs
+++ b/test/Sentry.Tests/SentryLogTests.cs
@@ -77,6 +77,42 @@ public class SentryLogTests
     }
 
     [Fact]
+    public void SetDefaultAttributes_ScopeOverridesOptions_UsesScopeValues()
+    {
+        var options = new SentryOptions
+        {
+            Environment = "options-environment",
+            Release = "options-release",
+        };
+        var scope = new Scope(options)
+        {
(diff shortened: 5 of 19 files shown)


test/Sentry.Tests/SentryMetricTests.cs
```diff
@@ -69,11 +72,47 @@ public void Protocol_Default_VerifyAttributes()
...
+        metric.Attributes.ShouldContain("sentry.release", "scope-release");
+    }
+
+    [Fact]
+    public void SetDefaultAttributes_NullScope_FallsBackToOptionsSettings()
+    {
+        var options = new SentryOptions
+        {
+            Release = "options-release",
+        };
```
Flash0ver, line 108: question: `Environment`

Should we set the `Environment` here too, as with the test above, as we also assert it below?

```suggestion
        var options = new SentryOptions
        {
            Environment = "options-environment",
            Release = "options-release",
        };
```

test/Sentry.Tests/SentryLogTests.cs
```diff
@@ -76,6 +76,42 @@ public void Protocol_Default_VerifyAttributes()
...
+        log.Attributes.ShouldContain<string>("sentry.release", "scope-release");
+    }
+
+    [Fact]
+    public void SetDefaultAttributes_NullScope_FallsBackToOptionsSettings()
+    {
+        var options = new SentryOptions
+        {
+            Release = "options-release",
+        };
```
Flash0ver, line 107: question: `Environment`

Should we set the `Environment` here too, as with the test above, as we alos assert it below?

```suggestion
        var options = new SentryOptions
        {
            Environment = "options-environment",
            Release = "options-release",
        };
```

test/Sentry.Tests/SentryMetricTests.cs
```diff
@@ -84,7 +123,7 @@ public void WriteTo_Envelope_MinimalSerializedSentryMetric()
         };
 
         var metric = new SentryMetric<int>(Timestamp, TraceId, SentryMetricType.Counter, "sentry_tests.sentry_metric_tests.counter", 1);
-        metric.Attributes.SetDefaultAttributes(options, new SdkVersion());
+        metric.Attributes.SetDefaultAttributes(options, new Scope());
```
Flash0ver: suggestion: pass `options`

To be consistent with the test in `SentryLogTests.cs`

```suggestion
        metric.Attributes.SetDefaultAttributes(options, new Scope(options));
```

test/Sentry.Tests/SentryMetricTests.cs
```diff
@@ -162,7 +201,7 @@ public void WriteTo_EnvelopeItem_MaximalSerializedSentryMetric()
         metric.SetAttribute("boolean-attribute", true);
         metric.SetAttribute("integer-attribute", 3);
         metric.SetAttribute("double-attribute", 4.4);
-        metric.Attributes.SetDefaultAttributes(options, new SdkVersion { Name = "Sentry.Test.SDK", Version = "1.2.3-test+Sentry" });
+        metric.Attributes.SetDefaultAttributes(options, new Scope { Sdk = { Name = "Sentry.Test.SDK", Version = "1.2.3-test+Sentry" } });
```
Flash0ver: suggestion: pass `options`

To be consistent with the equivalent test of `SentryLogTests.cs`

```suggestion
        metric.Attributes.SetDefaultAttributes(options, new Scope(options) { Sdk = { Name = "Sentry.Test.SDK", Version = "1.2.3-test+Sentry" } });
```

src/Sentry/Scope.cs
```diff
@@ -317,6 +317,7 @@ internal Scope(SentryOptions? options, SentryPropagationContext? propagationCont
         PropagationContext = new SentryPropagationContext(propagationContext);
 
         _environment = Options.Environment;
+        Release = Options.Release;
```
Flash0ver, line 321: question: `Release` not equivalent to `Environment`

The property setter of `Environment` performs a scope-sync.
Should the setter of `Release` likewise perform a scope-sync?
And if so, do we need a similar pattern as with `Environment`, so that we set to the backing field here directly rather than via the property setter?

src/Sentry/SentrySdk.cs
```diff
@@ -65,6 +65,7 @@ internal static IHub InitHub(SentryOptions options)
 
         // This happens before the native SDKs get initialized
         options.Environment = options.SettingLocator.GetEnvironment();
+        options.Release = options.SettingLocator.GetRelease();
```
jamescrosswell, line 68: 👍🏻

src/Sentry/Scope.cs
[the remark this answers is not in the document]
jamescrosswell: Not an immediate solution (would require a major version bump) but maybe reinforces this idea:
- https://github.com/getsentry/sentry-dotnet/issues/5197

In the interim, it seems like Options should only be used to populate the root scope. Any scope that gets cloned after that should probably just apply itself (not the options) to the new clone. 

Kind of opinionated, but can anyone see a reason to do it otherwise?

bitsandfoxes: Yes, but neither Cocoa nor Android accept the `Release` to sync as of right now. That'll require a followup.

CHANGELOG.md
```diff
@@ -7,6 +7,10 @@
 - feat: Add exponential backoff and log deduplication to Spotlight transport by @mattico in [#5025](https://github.com/getsentry/sentry-dotnet/pull/5025)
 - feat: The `Environment` set on the `Scope` now gets synchronized to the native layers (`sentry-cocoa` and `sentry-native`) by @bitsandfoxes [#5365](https://github.com/getsentry/sentry-dotnet/pull/5365)
 
+### Fixes
+
+- fix: When setting `Environment` and `Release` on the `Scope` these are  now also getting applied on emitted Metrics and Logs by bitsandfoxes [#5376](https://github.com/getsentry/sentry-dotnet/pull/5376)
+
```
Flash0ver, line 13: issue: don't edit `CHANGELOG.md` manually

Sorry for the confusion I've caused ... I meant to add a Changelog Entry _indirectly_ via `### Changelog Entry` in the PR description (https://craft.sentry.dev/configuration/#custom-changelog-entries-from-pr-descriptions).

This way we can have "great changelog entries" while avoiding merge conflicts when editing `CHANGELOG.md` manually.

We just merged new guidance: #5369

src/Sentry/Scope.cs
```diff
@@ -318,6 +318,7 @@ internal Scope(SentryOptions? options, SentryPropagationContext? propagationCont
 
         // Skip scope sync with backing field. The native SDKs already received the environment set on the options.
         _environment = Options.Environment;
+        Release = Options.Release;
```
Flash0ver, line 321: question: `Distribution`

Do we need similar treatment for `Distribution` ... to apply it from the `SentryOptions` onto the (root) `Scope`?

src/Sentry/Scope.cs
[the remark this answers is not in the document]
Flash0ver: Then ... I believe ... we have some unintended behavior!

By setting `Release = Options.Release;` in the `Scope`'s .ctor, when pushing a `Scope`:
1. `SentryScopeManager.PushScope` calls `Scope.Clone` (when not in Global-Mode and when not Locked)
1. `Scope.Clone` instantiates a new `Scope` by passing `Options` and `PropagationContext`
   - `Release` gets set on the new `Scope` instance from the `Options`
1. `Apply` the current `Scope` onto the new `Scope`
   - which only sets `Release` when `null`
   - https://github.com/getsentry/sentry-dotnet/blob/8efd196d25e5a5f4df78cb526e4cead8d6fd6d09/src/Sentry/Scope.cs#L522-L526

So `Release` is not only applied to the root Scope, and then copied over from the root Scope to pushed Scopes,
but actually always applied from the `Options`, so that the current `Scope`'s `Release` is overwritten.

```CSharp
SentrySdk.Init(options =>
{
    options.Release = "options-release";
});

SentrySdk.ConfigureScope(static scope =>
{
    scope.Release = "scope-release";
});
var scope = SentrySdk.PushScope();
SentrySdk.ConfigureScope(static scope =>
{
    Console.WriteLine(scope.Release); //options-release
});
```

Perhaps we need a facility, where we have a `CreateRootScope` factory method, that applies `Release` and other `SentryOptions`, and when later pushing a new `Scope`, we call existing .ctor with existing behavior that does not apply `Release` from the `Options`, but from the current `Scope` instead.

src/Sentry/SentrySdk.cs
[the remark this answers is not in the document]
Flash0ver: confirmed:

This does now apply `SentryOptions.Release` for _Native_:
https://github.com/getsentry/sentry-dotnet/blob/cc9de07cc1f9169f48cc2e507c0ec3468cb4d8bb/src/Sentry/Platforms/Native/CFunctions.cs#L89-L93

But, it changes existing behavior for _Android_
https://github.com/getsentry/sentry-dotnet/blob/11068e07c883649830613f61a5ef5de89106d90f/src/Sentry/Platforms/Android/SentrySdk.cs#L29

and also _iOS_ / _Mac Catalyst_
https://github.com/getsentry/sentry-dotnet/blob/35f0027fb84236e633d8942fb78f2f75e3e47e75/src/Sentry/Platforms/Cocoa/SentrySdk.cs#L16

Do we intend this change in behavior for _Android_/_Apple_?

src/Sentry/Scope.cs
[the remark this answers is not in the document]
jamescrosswell: TBH it's a bit weird for stuff like `Environment` and `Release` to be on the scope at all. Would these ever change during the program's execution? If the SDK user sets these to `Staging` and `7.6.5` then they would expect those values to be attached to all events/traces/etc (in the .NET, Apple, Android and Native SDKs). 

We seem to be using the scope as a vehicle to apply these properties to things that are `IEventLike`. Is there a  real world scenario where the user sets one of these things in the options and then sets it to something else on  the scope though? Or where they set them on the root scope and then want to push a new scope that has a different value for these?

> [!NOTE]
> `sentry-python` and `sentry-java` don't put environment/release on the scope at all — they're options-level settings

So maybe we go the opposite direction? When `Environment` and `Release` are set on the Options, that **_always_** overrides whatever is on the scope (so these things become invariants)?

And in the next major, we see if we can remove these from the Scope entirely. They should probably never have been added there.

@bitsandfoxes thoughts 🤔

src/Sentry/Platforms/Android/AndroidScopeObserver.cs
jamescrosswell, line 119: This might point to a more fundamental difference between how python/java are doing things vs .net/cocoa. 

I suspect the java/python approach makes more sense for these properties (these properties only exist on the options for those SDKs - not on the scope).

bitsandfoxes:
Turning that into a draft for now, I'll pick it up in a bit. This will also be taking the ideas from https://github.com/getsentry/sentry-dotnet/pull/4784

jamescrosswell:
> Turning that into a draft for now, I'll pick it up in a bit. This will also be taking the ideas from #4784

@bitsandfoxes I think we need to resolve what we want to do before putting a PR together. See [my comment](https://github.com/getsentry/sentry-dotnet/pull/5376#discussion_r3584155738) above.

I'll maybe post a link to that convo in the SDK channel on slack to see what other SDKs are doing.

PR #5377, opened 2026-07-10
feat: Add IgnoreTransactions option to filter transactions by name

### LINKED ISSUE #2306: Add ignoreTransactions and ignoreErrors
### Background:
To further simplfiy ignoring certain events from being sent to sentry, it is also suggested to provide `ignoreTransactions` and `ignoreErrors`

More info here develop docs updated: https://develop.sentry.dev/sdk/features/#event-sampling

- [ ] ignoreErrors
- [ ] ignoreTransactions
- [ ] SDK docs

Adham-Kiwan:
## Summary

Adds a first-class `SentryOptions.IgnoreTransactions` option: a list of substrings or regular expressions. A transaction whose name matches any entry is dropped before it is sent, mirroring [`ignoreTransactions`](https://docs.sentry.io/platforms/javascript/configuration/filtering/#using-ignore-transactions) in the JavaScript and Python SDKs.

Previously this required writing a `BeforeSendTransaction` callback (as noted by the maintainers in the issue). The new option gives parity with the other SDKs and the same shape as the existing `TagFilters` / `TracePropagationTargets` options.

Fixes #2306

(The `ignoreErrors` half of that issue is intentionally out of scope — it was marked "won't do" there, since exception filters already cover it.)

### Changelog Entry

Add `IgnoreTransactions` option to filter out transactions by name, matching substrings or regular expressions against the transaction name ([#5377](https://github.com/getsentry/sentry-dotnet/pull/5377))

## Implementation

- `SentryOptions.IgnoreTransactions` — `IList<StringOrRegex>`, defaulting to empty (opt-in), matching the `TagFilters` pattern.
- `SentryClient.CaptureTransaction` — drops a transaction whose `Name` matches, using the existing `MatchesSubstringOrRegex` helper, right after the required-field validation and before `BeforeSendTransaction`. Discards are recorded against the client report recorder (`DiscardReason.EventProcessor`)
- `BindableSentryOptions` — added the `List<string>?` surrogate + `ApplyTo` mapping for AOT-safe configuration binding, as required by the `BindableProperties_MatchOptionsProperties` guard.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,6 @@
 
 ### Features ✨
 
-- Add `IgnoreTransactions` option to filter out transactions by name, matching substrings or regular expressions against the transaction name ([#5377](https://github.com/getsentry/sentry-dotnet/pull/5377))
 - feat: Add exponential backoff and log deduplication to Spotlight transport by @mattico in [#5025](https://github.com/getsentry/sentry-dotnet/pull/5025)
 
 ## 6.6.0
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Features ✨
 
+- Add `IgnoreTransactions` option to filter out transactions by name, matching substrings or regular expressions against the transaction name ([#5377](https://github.com/getsentry/sentry-dotnet/pull/5377))
 - feat: Add exponential backoff and log deduplication to Spotlight transport by @mattico in [#5025](https://github.com/getsentry/sentry-dotnet/pull/5025)
 
 ## 6.6.0
--- a/src/Sentry/SentryClient.cs
+++ b/src/Sentry/SentryClient.cs
@@ -163,18 +163,6 @@ public class SentryClient : ISentryClient, IDisposable
             return;
         }
 
-        if (_options.IgnoreTransactions.MatchesSubstringOrRegex(transaction.Name))
-        {
-            // IgnoreTransactions is a built-in filter, so discards are recorded under
-            // EventProcessor (matching the exception filter path and the JS inbound filters),
-            // not BeforeSend, which is reserved for the user's BeforeSendTransaction callback.
-            var ignoredSpanCount = transaction.Spans.Count + 1; // 1 for each span + 1 for the transaction itself
-            _options.ClientReportRecorder.RecordDiscardedEvent(DiscardReason.EventProcessor, DataCategory.Transaction);
-            _options.ClientReportRecorder.RecordDiscardedEvent(DiscardReason.EventProcessor, DataCategory.Span, ignoredSpanCount);
-            _options.LogInfo("Transaction dropped by IgnoreTransactions option.");
-            return;
-        }
-
         // Unfinished transaction can only happen if the user calls this method instead of
         // transaction.Finish().
         // We still send these transactions over, but warn the user not to do it.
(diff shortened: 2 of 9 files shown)


hiSandog:
Placing this filter before `BeforeSendTransaction` means user callbacks will not observe ignored transactions, which seems consistent with an early built-in filter. Is that ordering intended to match the JavaScript and Python SDK behavior, and should it be called out in the option documentation? A test asserting that the callback is not invoked for a matched transaction would pin the contract and prevent an accidental ordering change later.

Adham-Kiwan:
Yes — the ordering is intentional and matches the JavaScript and Python SDKs, where `ignoreTransactions` is handled by the built-in inbound filter that runs ahead of the user's `before_send_transaction` hook.

I've updated the `IgnoreTransactions` doc comment to state this explicitly, and added `CaptureTransaction_MatchesIgnoreTransactions_BeforeSendTransactionNotInvoked`, which asserts the callback is not invoked for a matched transaction, to pin the contract against accidental reordering later. Thanks for the suggestion.

Adham-Kiwan:
Good catch — fixed in 1499ffc.

`IgnoreTransactions` now records its discards under `DiscardReason.EventProcessor` instead of `DiscardReason.BeforeSend`, matching the exception-filter path (`DoSendEvent`) and the JS inbound filters. `BeforeSend` is reserved for the user's `BeforeSendTransaction` callback, which doesn't run for an ignored transaction, so attributing the drop there would have misled drop-rate debugging. Updated the test assertion to match.

Adham-Kiwan:
Confirmed and fixed in 1a525c7 — this was a real (telemetry-only) issue.

The `IgnoreTransactions` check now runs **after** the `IsSampled is false` sampling gate. So a transaction that was sampled out (e.g. via trace propagation) is still recorded under `SampleRate` even if its name also matches an ignore pattern, rather than being misattributed to `EventProcessor`. Sampling is the earlier, primary drop reason, and `IgnoreTransactions` only meaningfully applies to transactions that would otherwise be sent — so this is also slightly more efficient (no filter check on already-sampled-out transactions). The check reuses the span count already computed for the sampling branch.

Added `CaptureTransaction_SampledOutAndMatchesIgnoreTransactions_RecordedAsSampleRate` to pin the ordering; I verified it fails if the filter is moved back ahead of the sampling check.

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 ### Features ✨
 
+- Add `IgnoreTransactions` option to filter out transactions by name, matching substrings or regular expressions against the transaction name ([#5377](https://github.com/getsentry/sentry-dotnet/pull/5377))
```
```suggestion
```

We need to remove this. See:
- https://github.com/getsentry/sentry-dotnet/pull/5369

The changelog entry can either be generated automatically from the title or added automatically based on a section in the PR description. The later is probably preferable for fixes and features like this - I've already added it for this PR.

src/Sentry/SentryClient.cs
```diff
@@ -185,6 +185,19 @@ public void CaptureTransaction(SentryTransaction transaction, Scope? scope, Sent
...
+        // Applied after the sampling check so that a transaction which is sampled out is
+        // still attributed to sampling, not to this filter. IgnoreTransactions is a built-in
+        // filter, so its discards are recorded under EventProcessor (matching the exception
+        // filter path and the JS inbound filters), not BeforeSend, which is reserved for the
+        // user's BeforeSendTransaction callback.
+        if (_options.IgnoreTransactions.MatchesSubstringOrRegex(transaction.Name))
+        {
+            _options.ClientReportRecorder.RecordDiscardedEvent(DiscardReason.EventProcessor, DataCategory.Transaction);
+            _options.ClientReportRecorder.RecordDiscardedEvent(DiscardReason.EventProcessor, DataCategory.Span, spanCount);
+            _options.LogInfo("Transaction dropped by IgnoreTransactions option.");
```
jamescrosswell, line 197: Arguably this could be `LogDebug` but we can tweak that in a follow up if anyone is really unhappy with it.

jamescrosswell (approved):
Awesome - thanks for the PR @Adham-Kiwan ! Very much appreciated 🙏🏻

PR #5381, opened 2026-07-13
fix: Create a single TaskBlockingListener per process for CaptureBlockingCalls

### LINKED ISSUE #5378: CaptureBlockingCalls leaks one TaskBlockingListener (EventListener) per request — every await in the process degrades linearly with total requests served
### Package

Sentry.AspNetCore

### .NET Flavor

.NET

### .NET Version

10.0

### OS

Linux

### OS Version

_No response_

### Development Environment

Visual Studio v18.x

No

_No response_

### SDK Version

6.6.0

_No response_

### Workload Versions

Workload version: 10.0.300-manifests.2c758974

{
    options.Dsn = "***";  // from configuration
    options.Environment = builder.Environment.EnvironmentName;
    options.Release = "***";  // from configuration

    options.MinimumEventLevel = LogLevel.Error;

    options.StackTraceMode = StackTraceMode.Enhanced;
    options.AttachStacktrace = true;

    options.TracesSampleRate = 0.2;

    // This is the option that triggers the leak. It was our default; we have
    // since flipped it to false as the workaround.
    options.CaptureBlockingCalls = true;

    options.SendDefaultPii = true;
    options.MaxRequestBodySize = RequestSize.Medium;
    options.SetBeforeSend((SentryEvent evt, SentryHint _) =>
    {
        evt.ServerName = null;
        return evt;
    });
});

### Steps to Reproduce

1. ASP.NET Core app with `UseSentry()` and `options.CaptureBlockingCalls = true`.
2. Let it serve requests — even just a Docker healthcheck hitting a trivial endpoint every 10s is enough.
3. Watch per-request latency, background CPU, and managed heap grow with uptime. Restarting the process resets everything.

### Expected Result

`CaptureBlockingCalls` should allocate its `TaskBlockingListener` once per process, with constant overhead.

### Actual Result

One `TaskBlockingListener` is created **per HTTP request** and never disposed, so the runtime's global `EventListener` list grows unboundedly and every `await` in the process pays a dispatch loop proportional to the total number of requests served since startup.

**Why it happens:**

- `SentryMiddleware` implements `IMiddleware` and is registered with `AddTransient<SentryMiddleware>()`, so DI constructs a new instance **for every request**.
- With `CaptureBlockingCalls = true`, the middleware **constructor** runs:

  ```csharp
  if (_options.CaptureBlockingCalls)
  {
      _monitor = new BlockingMonitor(_getHub, _options);
      _detectBlockingSyncCtx = new DetectBlockingSynchronizationContext(_monitor);
      _listener = new TaskBlockingListener(_monitor);
  }
  ```

- `TaskBlockingListener` derives from `System.Diagnostics.Tracing.EventListener`. Its base constructor registers it in the runtime's global listener chain, and it calls `EnableEvents(TplEventSource, Verbose, keywords: 3)`. Nothing ever calls `Dispose()` on it (the middleware is transient and holds it in a field).
- Net effect: the listener list grows by one per request, `TplEventSource` stays enabled forever, and `EventSource.DispatchToAllListeners` walks the whole chain on **every task-wait event of every await in the process** — not just inside requests.

**Observed impact (production, two services, effectively idle — only a 10s healthcheck):**

- p95 latency of a trivial health endpoint grew from ~5 ms to ~150 ms over 5 days (~45k requests → ~45k leaked listeners), growing linearly with request count. The two services degraded at exactly the same rate because they share the same healthcheck-driven request rate.
- `dotnet-trace` (thread-time profile) on both processes shows **96–97% of all CPU time inside `EventSource.DispatchToAllListeners`**, reached from every await (Kestrel middleware pipeline, HttpClient, timers, background workers).
- Steadily growing background CPU (one service reached ~26% of a core while idle) and steadily growing managed heap (leaked listeners are GC roots).
- Restarting the container resets latency/CPU/heap, after which the degradation immediately begins accumulating again.

**Suggested fix:** create the `TaskBlockingListener` (and `BlockingMonitor`) once per process — e.g. resolve them as singletons from DI or lazily initialize a static instance — rather than in the constructor of a transient middleware; or dispose the listener when the middleware instance is released. The per-request `DetectBlockingSynchronizationContext` swap can stay as is.

jamescrosswell:
## Summary

Fixes a per-request resource leak in `CaptureBlockingCalls` (ASP.NET Core), reported in #5378.

`SentryMiddleware` implements `IMiddleware` and is registered with `AddTransient<SentryMiddleware>()`, so DI constructs a **new instance per request**. With `CaptureBlockingCalls = true`, its constructor created a new `TaskBlockingListener` (a `System.Diagnostics.Tracing.EventListener`) on every request and never disposed it:

```csharp
_monitor = new BlockingMonitor(_getHub, _options);
_detectBlockingSyncCtx = new DetectBlockingSynchronizationContext(_monitor);
_listener = new TaskBlockingListener(_monitor); // registers itself globally, never disposed
```

Every `EventListener` registers itself in the runtime's **process-global** listener chain and enables `TplEventSource` (Verbose). So the chain grew by one per request and `EventSource.DispatchToAllListeners` walked the whole (ever-growing) chain on **every task-wait event of every `await` in the process** — not just inside requests.

The reporter observed, on effectively-idle services (only a 10s healthcheck): p95 of a trivial endpoint growing ~5 ms → ~150 ms over 5 days (~45k requests → ~45k leaked listeners), 96–97% of CPU inside `EventSource.DispatchToAllListeners`, and steadily growing background CPU and managed heap — all reset by a restart. Full credit to @gerardfontmora for the diagnosis.

## Fix

- Register `BlockingMonitor` and `TaskBlockingListener` as **singletons**, and have the transient middleware **resolve** (rather than construct) them, so exactly one `TaskBlockingListener` exists per process with constant overhead. They are only resolved when `CaptureBlockingCalls` is enabled, so disabling the feature has no cost.
- The per-request `DetectBlockingSynchronizationContext` swap in `InvokeAsync` is unchanged — it carries per-request suppression state and is correctly created per request.
- Drive-by: `TaskBlockingListener.DefaultState` was an expression-bodied property (`=> new()`) that allocated a fresh `Lazy<>` on every access instead of returning a single shared instance; changed to a static auto-property.

## Regression test

`Constructor_CaptureBlockingCalls_SharesSingleListenerAcrossInstances` builds two middleware instances (simulating two requests) and asserts they share one `Listener`/`Monitor`. Verified it fails against the previous per-request construction and passes with this change.

Fixes #5378

### DIFF
--- a/src/Sentry.AspNetCore/SentryMiddleware.cs
+++ b/src/Sentry.AspNetCore/SentryMiddleware.cs
@@ -35,11 +35,11 @@ internal class SentryMiddleware : IMiddleware
     private static readonly string ProtocolPackageName = "nuget:" + NameAndVersion.Name;
 
     // Ben.BlockingDetector
-    private readonly BlockingMonitor? _monitor;
+    private readonly IBlockingMonitor? _monitor;
     private readonly TaskBlockingListener? _listener;
 
     // Internal for testing
-    internal BlockingMonitor? Monitor => _monitor;
+    internal IBlockingMonitor? Monitor => _monitor;
     internal TaskBlockingListener? Listener => _listener;
 
     /// <summary>
@@ -81,7 +81,7 @@ internal class SentryMiddleware : IMiddleware
         if (_options.CaptureBlockingCalls)
         {
             // Resolve shared singletons to keep overhead constant - See #5378.
-            _monitor = serviceProvider.GetRequiredService<BlockingMonitor>();
+            _monitor = serviceProvider.GetRequiredService<IBlockingMonitor>();
             _listener = serviceProvider.GetRequiredService<TaskBlockingListener>();
         }
     }
--- a/src/Sentry.AspNetCore/SentryMiddleware.cs
+++ b/src/Sentry.AspNetCore/SentryMiddleware.cs
@@ -1,5 +1,6 @@
 using Microsoft.AspNetCore.Diagnostics;
 using Microsoft.AspNetCore.Http;
+using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Options;
 using Sentry.AspNetCore.Extensions;
@@ -35,8 +36,12 @@ internal class SentryMiddleware : IMiddleware
 
     // Ben.BlockingDetector
     private readonly BlockingMonitor? _monitor;
-    private readonly DetectBlockingSynchronizationContext? _detectBlockingSyncCtx;
     private readonly TaskBlockingListener? _listener;
+
+    // Exposed for tests, to verify that a single monitor/listener is shared across
+    // (transient) middleware instances rather than being allocated per request.
+    internal BlockingMonitor? Monitor => _monitor;
+    internal TaskBlockingListener? Listener => _listener;
 
     /// <summary>
     /// Initializes a new instance of the <see cref="SentryMiddleware"/> class.
@@ -48,6 +53,7 @@ internal class SentryMiddleware : IMiddleware
     /// <param name="eventExceptionProcessors">Custom Event Exception Processors</param>
     /// <param name="eventProcessors">Custom Event Processors</param>
     /// <param name="transactionProcessors">Custom Transaction Processors</param>
+    /// <param name="serviceProvider">The service provider, used to resolve shared blocking-call detection services.</param>
     /// <exception cref="ArgumentNullException">
     /// next
     /// or
@@ -60,7 +66,8 @@ internal class SentryMiddleware : IMiddleware
         ILogger<SentryMiddleware> logger,
         IEnumerable<ISentryEventExceptionProcessor> eventExceptionProcessors,
         IEnumerable<ISentryEventProcessor> eventProcessors,
-        IEnumerable<ISentryTransactionProcessor> transactionProcessors)
+        IEnumerable<ISentryTransactionProcessor> transactionProcessors,
+        IServiceProvider serviceProvider)
     {
         ArgumentNullException.ThrowIfNull(getHub);
 
@@ -74,9 +81,15 @@ internal class SentryMiddleware : IMiddleware
 
         if (_options.CaptureBlockingCalls)
         {
-            _monitor = new BlockingMonitor(_getHub, _options);
-            _detectBlockingSyncCtx = new DetectBlockingSynchronizationContext(_monitor);
-            _listener = new TaskBlockingListener(_monitor);
+            // The monitor and listener are registered as process-wide singletons (see
+            // SentryWebHostBuilderExtensions). This middleware is transient (one instance per
+            // request), so constructing them here would create - and never dispose - a new
+            // TaskBlockingListener (an EventListener) on every request. Each leaked listener stays
+            // rooted in the runtime's global listener chain, keeps TplEventSource enabled, and makes
+            // EventSource.DispatchToAllListeners walk an ever-growing chain on every await in the
+            // process. Resolving the shared singletons instead keeps the overhead constant. See #5378.
+            _monitor = serviceProvider.GetRequiredService<BlockingMonitor>();
+            _listener = serviceProvider.GetRequiredService<TaskBlockingListener>();
         }
     }
 
@@ -150,7 +163,12 @@ internal class SentryMiddleware : IMiddleware
                 if (_options.CaptureBlockingCalls && _monitor is not null)
                 {
                     var syncCtx = SynchronizationContext.Current;
-                    SynchronizationContext.SetSynchronizationContext(syncCtx == null ? _detectBlockingSyncCtx : new DetectBlockingSynchronizationContext(_monitor, syncCtx));
+                    // The synchronization context wraps per-request suppression state, so it must be
+                    // created per request - unlike the monitor/listener, which are shared singletons.
+                    var detectingSyncCtx = syncCtx is null
+                        ? new DetectBlockingSynchronizationContext(_monitor)
+                        : new DetectBlockingSynchronizationContext(_monitor, syncCtx);
+                    SynchronizationContext.SetSynchronizationContext(detectingSyncCtx);
                     try
                     {
                         // For detection to work we need ConfigureAwait=true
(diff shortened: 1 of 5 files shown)


src/Sentry.AspNetCore/SentryMiddleware.cs
```diff
@@ -35,9 +36,13 @@ internal static readonly SdkVersion NameAndVersion
 
     // Ben.BlockingDetector
     private readonly BlockingMonitor? _monitor;
-    private readonly DetectBlockingSynchronizationContext? _detectBlockingSyncCtx;
     private readonly TaskBlockingListener? _listener;
 
+    // Exposed for tests, to verify that a single monitor/listener is shared across
+    // (transient) middleware instances rather than being allocated per request.
```
jamescrosswell:
```suggestion
    // Internal for testing
```

src/Sentry.AspNetCore/SentryMiddleware.cs
```diff
@@ -48,6 +53,7 @@ internal static readonly SdkVersion NameAndVersion
     /// <param name="eventExceptionProcessors">Custom Event Exception Processors</param>
     /// <param name="eventProcessors">Custom Event Processors</param>
     /// <param name="transactionProcessors">Custom Transaction Processors</param>
+    /// <param name="serviceProvider">The service provider, used to resolve shared blocking-call detection services.</param>
```
jamescrosswell:
```suggestion
    /// <param name="serviceProvider">The service provider, used to resolve dependencies.</param>
```

src/Sentry.AspNetCore/SentryMiddleware.cs
```diff
@@ -74,9 +81,15 @@ public SentryMiddleware(
...
-            _monitor = new BlockingMonitor(_getHub, _options);
-            _detectBlockingSyncCtx = new DetectBlockingSynchronizationContext(_monitor);
-            _listener = new TaskBlockingListener(_monitor);
+            // The monitor and listener are registered as process-wide singletons (see
+            // SentryWebHostBuilderExtensions). This middleware is transient (one instance per
+            // request), so constructing them here would create - and never dispose - a new
+            // TaskBlockingListener (an EventListener) on every request. Each leaked listener stays
+            // rooted in the runtime's global listener chain, keeps TplEventSource enabled, and makes
+            // EventSource.DispatchToAllListeners walk an ever-growing chain on every await in the
+            // process. Resolving the shared singletons instead keeps the overhead constant. See #5378.
```
jamescrosswell:
```suggestion
            // Resolve shared singletons to keep overhead constant - See #5378.
```

src/Sentry.AspNetCore/SentryWebHostBuilderExtensions.cs
```diff
@@ -112,6 +114,15 @@ public static IWebHostBuilder UseSentry(
         _ = builder.ConfigureServices(c => _ =
             c.AddTransient<IStartupFilter, SentryStartupFilter>()
              .AddTransient<IStartupFilter, SentryTracingStartupFilter>()
+             // Blocking-call detection uses an EventListener that registers itself in the runtime's
+             // process-global listener chain, so exactly one must exist per process. Register the
+             // monitor and listener as singletons; the transient SentryMiddleware resolves (rather
+             // than constructs) them, avoiding a per-request listener leak. See issue #5378.
```
jamescrosswell:
```suggestion
```

src/Sentry.AspNetCore/SentryWebHostBuilderExtensions.cs
```diff
@@ -112,6 +114,11 @@ public static IWebHostBuilder UseSentry(
         _ = builder.ConfigureServices(c => _ =
             c.AddTransient<IStartupFilter, SentryStartupFilter>()
              .AddTransient<IStartupFilter, SentryTracingStartupFilter>()
+             .AddSingleton<BlockingMonitor>(p => new BlockingMonitor(
+                 p.GetRequiredService<Func<IHub>>(),
+                 p.GetRequiredService<IOptions<SentryAspNetCoreOptions>>().Value))
+             .AddSingleton<TaskBlockingListener>(p => new TaskBlockingListener(
+                 p.GetRequiredService<BlockingMonitor>()))
```
jamescrosswell: Can this code be simplified? We appear to be constructing the singleton instances here simply by resolving the dependencies manually. Won't DI do that for us if we do this?

```suggestion
             .AddSingleton<BlockingMonitor>()  
             .AddSingleton<TaskBlockingListener>()
```

jamescrosswell: Done in [c58db3e](https://github.com/getsentry/sentry-dotnet/pull/5381/commits/c58db3e558ab17fd8c3e967e6c8de2d1b5414a26) although`TaskBlockingListener`'s constructor takes an `IBlockingMonitor` so had to add an `IBlockingMonitor` → `BlockingMonitor` as well.

src/Sentry.AspNetCore/SentryWebHostBuilderExtensions.cs
```diff
@@ -112,6 +113,11 @@ public static IWebHostBuilder UseSentry(
         _ = builder.ConfigureServices(c => _ =
             c.AddTransient<IStartupFilter, SentryStartupFilter>()
              .AddTransient<IStartupFilter, SentryTracingStartupFilter>()
+             // Single listener/monitor per process (the listener is a global EventListener) - See #5378.
+             // The IBlockingMonitor forward keeps the listener on the same monitor instance.
+             .AddSingleton<BlockingMonitor>()
+             .AddSingleton<IBlockingMonitor>(p => p.GetRequiredService<BlockingMonitor>())
```
jamescrosswell: Can this not also be simplified to:
```suggestion
             .AddSingleton<IBlockingMonitor, BlockingMonitor>()
```

jamescrosswell: [6724a01](https://github.com/getsentry/sentry-dotnet/pull/5381/commits/6724a01a11185846cd724679ef4f4d4b7ace898b)

bruno-garcia (approved):
LGTM, I'd add some integration tests too as a follow up

jamescrosswell:
> LGTM, I'd add some integration tests too as a follow up

Follow up issue:
- https://github.com/getsentry/sentry-dotnet/issues/5394

Issue #5350, opened 2026-07-06
Different behavior between Sentry Open Telemetry and Sentry .NET SDK


yasmoradi:
### Description

I'm enabling Open Telemetry by following code:

```cs
services.AddLogging(loggingBuilder =>
{
    loggingBuilder.ConfigureLoggers(configuration);

    loggingBuilder.AddOpenTelemetry(options =>
    {
        options.IncludeScopes = true;
        options.IncludeFormattedMessage = true;
        configuration.Bind("Logging:OpenTelemetry", options);
    });
});

var openTelemetry = services.AddOpenTelemetry()
    .WithMetrics()
    .WithTracing()
    .ConfigureResource(resource =>
    {
        resource.AddAttributes([new("service.name", Application.ProductName!)]);
    });

var useOtlpExporter = string.IsNullOrWhiteSpace(configuration["OTEL_EXPORTER_OTLP_ENDPOINT"]) is false
    || string.IsNullOrEmpty(configuration["OTEL_EXPORTER_OTLP_LOGS_ENDPOINT"]) is false;

if (useOtlpExporter)
{
    openTelemetry.UseOtlpExporter();
}
```

It shows logs in `https://xxx.sentry.io/explore/logs/` and shows traces in `https://xxx.sentry.io/explore/traces/`.

My question is, why it doesn't show errors in `https://xxx.sentry.io/issues/` (Which basically is a home page)
Before switching to Open Telemetry, I could see the in home page and this is so confusing for people who expects to see something in the home page.

The other question that I've is as I can see, there's a metrics page `https://xxx.sentry.io/explore/metrics/`

But it seems `OTEL_EXPORTER_OTLP_METRICS_ENDPOINT` is not supported at the moment as it's not documented in `https://xxx.sentry.io/settings/projects/dotnet/keys/`

jamescrosswell:
> It shows logs in `https://xxx.sentry.io/explore/logs/` and shows traces in `https://xxx.sentry.io/explore/traces/`.
> 
> My question is, why it doesn't show errors in `https://xxx.sentry.io/issues/` 

OpenTelemetry itself is a three legged stool that covers: Logs, Metrics and Traces. 

There isn't any first class support for Errors in OpenTelemetry.
[...]

Issue #4859, opened 2026-01-21
Investigate OTLP integration


jamescrosswell:
# Description

- [Developer docs](https://develop.sentry.dev/sdk/telemetry/traces/otlp/)
- Python reference implementation
  - [Documentation](https://docs.sentry.io/platforms/python/integrations/otlp/)
  - [Pull request](https://github.com/getsentry/sentry-python/pull/4877)

> [!NOTE]
> The [Exception Interceptor](https://develop.sentry.dev/sdk/telemetry/traces/otlp/#exception-interceptor) is something we already have in the POTEL implementation for .NET but [have problems with](https://github.com/open-telemetry/opentelemetry-dotnet/issues/2439#issuecomment-1577314568). This [will be deprecated in the OTEL APIs](https://github.com/open-telemetry/opentelemetry-python/issues/4561). The recommendation is to use [the OTEL logging APIs for this instead](https://opentelemetry.io/docs/specs/semconv/exceptions/exceptions-logs/).

## Initial head scratching

Currently the meat of our OTEL integration is implemented in the `SentrySpanProcessor`. Basically it takes OTEL spans and converts these into Sentry spans, stored on the Sentry scope. We have a couple of different scope implementations: global and AsyncLocal... since `Activity.Current` (Microsoft's implementation of OTEL) is AsyncLocal, this works best in backend applications. In front end apps using a global scope, often spans from background services can end up interleaved with front end operations.

From what I can gather from the docs and the Python implementation, moving to OTLP would flip this on it's head and we'd instead send Sentry instrumented spans to an OTEL exporter which would pipe both the Sentry spans and any spans from the user's code through to a DSN specific OLTP endpoint on the Sentry server/ingest.

### Problems so solve

The `SentrySpanProcessor` currently ensures OTEL and Sentry spans are parented to one another correctly:
https://github.com/getsentry/sentry-dotnet/blob/54511b5c75f5f016e4c56940caf59cab1316ee2e/src/Sentry.OpenTelemetry/SentrySpanProcessor.cs#L96-L114
https://github.com/getsentry/sentry-dotnet/blob/54511b5c75f5f016e4c56940caf59cab1316ee2e/src/Sentry.OpenTelemetry/SentrySpanProcessor.cs#L175

It also does a bit of parsing to synthesise errors and enrich the spans:
[...]

PR #5382, opened 2026-07-13
spike: Activity-based tracing — SentrySpanProcessor as a raw ActivityListener in core (experiment, do not merge)

jamescrosswell:
## ⚠️ Experiment — not intended to be merged

This is part (a) of a two-part spike exploring whether `System.Diagnostics.Activity` can become the single source of truth for tracing in the .NET SDK (context: #3238, #3288, #5350, #4859, and the internal discussion about the Sentry-vs-OTel instrumentation ecosystems bifurcating). It exists to inform that discussion with working code; it will be closed, not merged.

## Question this spike answers

> Can the Activity→Sentry conversion logic in `Sentry.OpenTelemetry`'s `SentrySpanProcessor` live in the **core Sentry package**, driven by a raw `System.Diagnostics.ActivityListener`, with **zero dependency on the OpenTelemetry SDK** — on **all TFMs** — and pass the existing `SentrySpanProcessorTests` suite unchanged?

**Answer: yes.** The port was near-verbatim. Exactly two things needed to change, both predicted in advance:

1. **No `BaseProcessor<Activity>` base class** — `ActivityListener.ActivityStarted`/`ActivityStopped` are drop-in replacements for `OnStart`/`OnEnd`.
2. **OTel `Resource` detection** (`ParentProvider.GetResource()`) has no `System.Diagnostics` equivalent — replaced with an injectable `resourceAttributeResolver` (defaults to empty).

Everything else — the shadow span map, transaction-vs-child parent inference (including mixed Sentry/OTel parenting), DSC-from-baggage, DSN self-request filtering, scope restoration across async, pruning/leak guards, error generation from exception events, op/description/status mapping — is pure `System.Diagnostics` code and ported as-is.

## Packaging model: all TFMs, no new core dependency

The sources live in `src/Sentry.DiagnosticSource/Internal/Tracing/` and ship through the **same split the EF/SqlClient DiagnosticSource integration already uses**:

- **Modern TFMs**: compiled directly into the core `Sentry` package via the existing `<Compile Include>` in `Sentry.csproj` — `ActivityListener` is part of the shared framework there, so **core takes no new dependency**.
- **Legacy TFMs** (`netstandard2.0/2.1`, `net462`): the code ships in the `Sentry.DiagnosticSource` package, which carries the `System.Diagnostics.DiagnosticSource` package reference. Legacy apps opt in by adding that package — exactly as they do today for EF/SQL instrumentation.

📌 **Version finding**: `ActivityListener` needs DiagnosticSource **5.0** and `ActivityStatusCode` needs **6.0**, but the `netstandard2.0`/`net462` assets of the 6.0 package lag the modern API surface (no `Activity.HasRemoteParent`), so **8.0.1 is the effective minimum** (bumped from 4.5.0).

A `HAS_ACTIVITY_LISTENER` constant (defined in both csproj files) gates compilation; only two small source concessions were needed for the legacy TFMs (hex-string id conversion branches, no `KeyValuePair` deconstruction).

## What's here

| File | What it is |
|---|---|
| `src/Sentry.DiagnosticSource/Internal/Tracing/SentryActivityProcessor.cs` | Port of `SentrySpanProcessor` (kept diffable against the original) |
| `src/Sentry.DiagnosticSource/Internal/Tracing/SentryActivityListener.cs` | Replaces the OTel `TracerProvider` registration glue with a raw `ActivityListener` |
| `src/Sentry.DiagnosticSource/Internal/Tracing/ActivityIdExtensions.cs`, `ActivityAttributeExtensions.cs`, `ISentryActivityEnricher.cs` | Ports of the small helpers the processor needs (distinct namespace to avoid `InternalsVisibleTo` collisions with the copies in the OTel packages) |
| `test/Sentry.DiagnosticSource.Tests/Tracing/SentryActivityProcessorTests.cs` | All **33** tests from `SentrySpanProcessorTests` ported 1:1 — bodies kept diffable — passing against the core port |
| `test/Sentry.DiagnosticSource.Tests/Tracing/SentryActivityListenerTests.cs` | **4** new end-to-end tests: root activity → captured transaction, child activity → correctly-parented span, unsampled → dropped, unmatched source → `StartActivity` returns `null` (the zero-cost `HasListeners` property). No OTel SDK anywhere in the pipeline. |

The tests live in `Sentry.DiagnosticSource.Tests` deliberately: modern TFMs exercise the compiled-into-core path, while `net48` (Windows CI) exercises the standalone-package path — one suite covers both packagings.

## Test evidence (local, macOS)

- Original suite: `SentrySpanProcessorTests` 33/33 ✅
- Ported suite against the core processor: 33/33 ✅ (+ 4/4 new listener e2e tests; `Sentry.DiagnosticSource.Tests` 119/119 on net10.0)
- Full `Sentry.Tests` (net10.0): 2402 passed, 0 failed ✅
- `src/Sentry` builds clean on all TFMs ✅; `Sentry.DiagnosticSource` builds clean on `netstandard2.0/2.1`/`net462` with DiagnosticSource 8.0.1 ✅
- OTel packages build unchanged ✅
- `net48` test execution happens in Windows CI (not runnable locally on macOS)

## What this spike deliberately does NOT address

- **Sampling at creation time** — the listener's `Sample` callback returns `AllDataAndRecorded` for parity with today's POTEL behaviour (record everything, Sentry samples at conversion). Moving Sentry's sampling into the `Sample` callback is future work and needs answers for the `TracesSampler` `customSamplingContext` gap and multi-listener composition (the runtime takes the most permissive result).
- **The `Instrumenter.OpenTelemetry` constructor guard** is retained as-is for test parity; in a real implementation it inverts (Activity-based becomes the default, not the opt-in).
- Idle/deadline transactions (MAUI, #5138) — that's part (b) of the spike.
- Migrating any integration to emit Activities, propagation, or public API shape.

#skip-changelog


src/Sentry.DiagnosticSource/Internal/Tracing/ActivityIdExtensions.cs
```diff
@@ -0,0 +1,45 @@
+#if HAS_ACTIVITY_LISTENER
```
jamescrosswell: Do we need this now everything is in `Sentry.DiagnosticSource`? We'll always have the ActivityListener available right?

PR #5393, opened 2026-07-15
fix(OpenTelemetry): fuse Activity weakly so never-ended spans are pruned

### LINKED ISSUE #5392: Memory leak: never-ended Activities accumulate in SentrySpanProcessor._map (sibling of #3197)
### Package

Sentry.OpenTelemetry

### .NET Flavor / Version

.NET 10.0.7 (also reproduced on .NET 8 and 9)

### OS

Linux (production). Reproduces on any OS.

### SDK Version

6.6.0

### Summary

`SentrySpanProcessor` leaks spans whose Activity never receives `OnEnd`, for example aborted HTTP/2 streams or abandoned gRPC server streams. Such spans accumulate in `SentrySpanProcessor._map` and are never removed, so a long running service grows in memory until it runs out.

This is a sibling of #3197 / #3198. #3198 fixed the filtered span case, and its description mentions holding weak references to Activities so `PruneFilteredSpans` could also evict collected ones. But the reference is stored via `SetFused`, which is backed by a `ConditionalWeakTable` and holds its values strongly (only keys are weak). So `_map -> span -> Activity` pins the Activity alive, the `activity is null` branch in `PruneFilteredSpans` can never run, and recorded spans that never end stay in the map forever.

### UseSentry / SentrySdk.Init call

```csharp
builder.WebHost.UseSentry(o =>
{
    o.Dsn = "<redacted>";
    o.UseOpenTelemetry();
    o.DisableSentryHttpMessageHandler = true;
    o.TracesSampler = ctx => 0.1; // sample 10% of root traces
});

builder.Services.AddOpenTelemetry().WithTracing(t => t
    .AddAspNetCoreInstrumentation()
    .AddSentry());
```

### Steps to Reproduce

Minimal repro: https://gist.github.com/Ermabo/41e0301e0afabfe4d28cff799d4e79b6

1. `dotnet run -c Release -- --close true` (normal spans): memory stays flat.
2. `dotnet run -c Release -- --close false` (Activities started but never stopped, i.e. an aborted request): memory grows without bound, about 1.9 GB over 300k iterations.

### Expected Result

Spans whose Activity never ends are eventually evicted from `_map`, and memory stays bounded.

### Actual Result

They are retained forever. In our production gRPC service we observed about 91k `Sentry.Internal.UnsampledTransaction` retained with `_isFinished == false`, `gcroot` terminating at `SentrySpanProcessor._map`. Managed gen2 grew from about 190 MB to 2.4 GB over roughly 6 days before an out of memory crash, repeating roughly every 4 days.

A fix PR is on the way. cc @jamescrosswell (author of #3198).

Ermabo:
Fixes #5392. Follow-up to #3198 (from #3166 / #3197).

## Problem

`SentrySpanProcessor._map` accumulates spans that never receive `OnEnd` (aborted HTTP/2 streams, abandoned gRPC streams, other never-completed requests), so a long running service grows in memory until it runs out.

## Root cause

`OnStart` fuses the Activity onto the span with `SetFused(data)`. `SetFused` is backed by a `ConditionalWeakTable`, which holds values strongly (only keys are weak), so `_map -> span -> Activity` pins the Activity. The `activity is null` branch in `PruneFilteredSpans` (added in #3198 to evict collected activities) can therefore never run, and recorded spans that never end stay forever. #3198's description mentions weak references, but `SetFused` stores them strongly.

## Fix

Fuse the Activity via `WeakReference<Activity>` so `_map` no longer pins it; the existing prune null-check then evicts orphans once the Activity is collected. `IsFiltered` still reads the sampling flags lazily from the Activity (works while it is alive).

## Evidence

Repro: https://gist.github.com/Ermabo/41e0301e0afabfe4d28cff799d4e79b6 (never-ended spans grow the heap ~1.9 GB over 300k iterations; normal spans stay flat). In production we saw ~91k `UnsampledTransaction` retained (`_isFinished == false`, `gcroot` terminating at `_map`); gen2 grew ~190 MB to 2.4 GB over ~6 days before an OOM.

## Tests

`OnStart_FusesActivityWeakly` plus the existing `PruneFilteredSpans_GarbageCollectedActivity_Pruned`; all 44 `SentrySpanProcessorTests` pass.

### Changelog Entry

fix: `SentrySpanProcessor` no longer leaks spans whose Activity never ends (e.g. aborted requests); the Activity is now held via a `WeakReference` so orphaned spans are pruned once it is garbage-collected.

## Note

One edge remains from making the reference weak: `IsFiltered` is read at parent-transaction finish, so if a filtered (sampled-out) child's Activity is collected before the parent serializes, the check returns false and that span is included rather than dropped. This is rare, is minor telemetry noise rather than a correctness/stability issue, and is strictly better than the OOM this PR fixes.

An eager snapshot at `OnStart` was tried and reverted (see commit history): `OnStart` runs before OTel instrumentation filters (e.g. HttpClient `Filter`) clear the sampling flags, so it captured filtered spans as unfiltered, a worse trade. The only read point that is both post-filter and GC-safe is `OnEnd`; capturing there would resolve the edge fully but adds complexity and since this package it deprecated I don't think it warrants the refactor.

### DIFF
--- a/src/Sentry.OpenTelemetry/SentrySpanProcessor.cs
+++ b/src/Sentry.OpenTelemetry/SentrySpanProcessor.cs
@@ -136,7 +136,7 @@ public class SentrySpanProcessor : BaseProcessor<Activity>
         var span = parentSpan.StartChild(context);
         // Fuse the Activity via a WeakReference so _map does not pin it, letting PruneFilteredSpans evict
         // never-ended spans. #3198 intended weak refs here but SetFused stores the value strongly.
-        span.SetFused("Activity", new WeakReference<Activity>(data));
+        SetFusedActivity(span, data);
         if (span is SpanTracer spanTracer)
         {
             spanTracer.Origin = OpenTelemetryOrigin;
@@ -179,7 +179,7 @@ public class SentrySpanProcessor : BaseProcessor<Activity>
             scope.Transaction ??= transaction;
         }, transaction);
         // Fuse weakly so _map does not pin the Activity.
-        transaction.SetFused("Activity", new WeakReference<Activity>(data));
+        SetFusedActivity(transaction, data);
         _map[data.SpanId] = transaction;
     }
 
@@ -317,9 +317,15 @@ public class SentrySpanProcessor : BaseProcessor<Activity>
         }
     }
 
+    private const string ActivityPropertyName = "Activity";
+
+    // Fuses the Activity onto the span via a WeakReference so _map does not pin it.
+    private static void SetFusedActivity(ISpan span, Activity data) =>
+        span.SetFused(ActivityPropertyName, new WeakReference<Activity>(data));
+
     // Returns the fused Activity, or null once it has been garbage-collected.
     private static Activity? GetFusedActivity(ISpan span) =>
-        span.GetFused<WeakReference<Activity>>("Activity") is { } weakRef && weakRef.TryGetTarget(out var activity)
+        span.GetFused<WeakReference<Activity>>(ActivityPropertyName) is { } weakRef && weakRef.TryGetTarget(out var activity)
             ? activity
             : null;
 
--- a/src/Sentry.OpenTelemetry/SentrySpanProcessor.cs
+++ b/src/Sentry.OpenTelemetry/SentrySpanProcessor.cs
@@ -141,8 +141,11 @@ public class SentrySpanProcessor : BaseProcessor<Activity>
         {
             spanTracer.Origin = OpenTelemetryOrigin;
             spanTracer.StartTimestamp = data.StartTimeUtc;
-            // Used to filter out spans that are not recorded when finishing a transaction.
-            spanTracer.IsFiltered = () => GetFusedActivity(spanTracer) is { IsAllDataRequested: false, Recorded: false };
+            // Capture the sampling decision now; it is fixed at span creation. Reading it lazily from the
+            // fused Activity would fail once the Activity has been garbage-collected (weak ref returns null),
+            // causing a filtered span to be included in the transaction instead of dropped.
+            var isFiltered = data is { IsAllDataRequested: false, Recorded: false };
+            spanTracer.IsFiltered = () => isFiltered;
         }
         _map[data.SpanId] = span;
     }
--- a/src/Sentry.OpenTelemetry/SentrySpanProcessor.cs
+++ b/src/Sentry.OpenTelemetry/SentrySpanProcessor.cs
@@ -134,13 +134,15 @@ public class SentrySpanProcessor : BaseProcessor<Activity>
         };
 
         var span = parentSpan.StartChild(context);
-        // Used to filter out spans that are not recorded when finishing a transaction
-        span.SetFused(data);
+        // Fuse the Activity via a WeakReference so _map does not pin it, letting PruneFilteredSpans evict
+        // never-ended spans. #3198 intended weak refs here but SetFused stores the value strongly.
+        span.SetFused("Activity", new WeakReference<Activity>(data));
         if (span is SpanTracer spanTracer)
         {
             spanTracer.Origin = OpenTelemetryOrigin;
             spanTracer.StartTimestamp = data.StartTimeUtc;
-            spanTracer.IsFiltered = () => spanTracer.GetFused<Activity>() is { IsAllDataRequested: false, Recorded: false };
+            // Used to filter out spans that are not recorded when finishing a transaction.
+            spanTracer.IsFiltered = () => GetFusedActivity(spanTracer) is { IsAllDataRequested: false, Recorded: false };
         }
         _map[data.SpanId] = span;
     }
@@ -176,7 +178,8 @@ public class SentrySpanProcessor : BaseProcessor<Activity>
         {
             scope.Transaction ??= transaction;
         }, transaction);
-        transaction.SetFused(data);
+        // Fuse weakly so _map does not pin the Activity.
+        transaction.SetFused("Activity", new WeakReference<Activity>(data));
         _map[data.SpanId] = transaction;
     }
 
@@ -304,15 +307,21 @@ public class SentrySpanProcessor : BaseProcessor<Activity>
         foreach (var mappedItem in _map)
         {
             var (spanId, span) = mappedItem;
-            var activity = span.GetFused<Activity>();
-            // Also prune when the activity has been GC'd (weak ref returns null): the activity is gone, so it
-            // can never call OnEnd, and the span will never be removed otherwise — causing a memory leak.
+            var activity = GetFusedActivity(span);
+            // Prune when the activity has been GC'd (weak ref returns null): it can no longer call OnEnd, so
+            // the span would otherwise stay in _map forever.
             if (activity is null or { Recorded: false, IsAllDataRequested: false })
             {
                 _map.TryRemove(spanId, out _);
             }
         }
     }
+
+    // Returns the fused Activity, or null once it has been garbage-collected.
+    private static Activity? GetFusedActivity(ISpan span) =>
+        span.GetFused<WeakReference<Activity>>("Activity") is { } weakRef && weakRef.TryGetTarget(out var activity)
+            ? activity
+            : null;
 
     private bool NeedsPruning()
     {
--- a/src/Sentry.OpenTelemetry/SentrySpanProcessor.cs
+++ b/src/Sentry.OpenTelemetry/SentrySpanProcessor.cs
@@ -141,11 +141,8 @@ public class SentrySpanProcessor : BaseProcessor<Activity>
         {
             spanTracer.Origin = OpenTelemetryOrigin;
             spanTracer.StartTimestamp = data.StartTimeUtc;
-            // Capture the sampling decision now; it is fixed at span creation. Reading it lazily from the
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
+
+### Fixes 🐛
+
+- fix: `SentrySpanProcessor` no longer leaks never-ended spans. [#3198](https://github.com/getsentry/sentry-dotnet/pull/3198) intended to hold Activities via weak references so `PruneFilteredSpans` could evict orphaned (never-`OnEnd`ed) spans, but `SetFused` stored them strongly, pinning each Activity alive via the internal map so it was never collected. Aborted HTTP/2 streams and other requests that never call `OnEnd` then built up in the map until the process ran out of memory. The Activity is now fused via a `WeakReference` so orphaned spans are pruned once it is garbage-collected. by @Ermabo in [#5393](https://github.com/getsentry/sentry-dotnet/pull/5393)
+
```
```suggestion
```

See:
- https://github.com/getsentry/sentry-dotnet/pull/5369

Ermabo:
Cheers @jamescrosswell! Removed the manual CHANGELOG edit (missed the "do not edit manually" note in CONTRIBUTING) and added a `### Changelog Entry` section to the PR description for craft; it's force-pushed.

Ermabo:
Separate from this fix:

The underlying design here (and in #3197) is that _map treats the span as the owner and infers the Activity's lifecycle indirectly. A more robust long-term design might invert that and store the span on the Activity via Activity.SetCustomProperty so its lifetime follows the Activity's. This shrinks _map down to just parent lookups and removes this prune/GC-timing class entirely. 

It's not free: remote/ActivityContext parents still need a small lookup. It fixes the memory orphan problem but not never-finished spans, and it likely has cross-SDK implications; so happy to open a discussion to explore rather than scope-creep this PR, if there is interest 😊

jamescrosswell:
> The underlying design here (and in #3197) is that _map treats the span as the owner and infers the Activity's lifecycle indirectly. A more robust long-term design might invert that and store the span on the Activity via Activity.SetCustomProperty so its lifetime follows the Activity's.

Thanks @Ermabo - given that this integration is [marked as deprecated](https://docs.sentry.io/platforms/dotnet/tracing/instrumentation/opentelemetry/), probably not worth burning too many cycles on it. 

Longer term, the [OTLP integration](https://docs.sentry.io/platforms/dotnet/tracing/instrumentation/opentelemetry-otlp/) will be the primary method of integrating with OpenTelemetry.

Still possibly interesting as part of [this work](https://github.com/getsentry/sentry-dotnet/pull/5382) - but that's a much bigger project (currently on hold).

src/Sentry.OpenTelemetry/SentrySpanProcessor.cs
```diff
@@ -134,13 +134,15 @@ private void CreateChildSpan(Activity data, ISpan parentSpan, SpanId? parentSpan
         };
 
         var span = parentSpan.StartChild(context);
-        // Used to filter out spans that are not recorded when finishing a transaction
-        span.SetFused(data);
+        // Fuse the Activity via a WeakReference so _map does not pin it, letting PruneFilteredSpans evict
+        // never-ended spans. #3198 intended weak refs here but SetFused stores the value strongly.
+        span.SetFused("Activity", new WeakReference<Activity>(data));
```
jamescrosswell: Purely stylistic, but while this is functionally correct, it kind of bleeds (and relies on) some of the implementation details of the Fused extension methods.

Since we already have a `GetFusedActivity` method defined, an alternative would be to create a symmetrical `SetFusedActivity(ISpan span, Activity data)` method and have both of them pass a `const string ActivityPropertyName = "Activity"` as the `propertyName` argument.

That way (and whilst this isn't likely to happen) if the implementation of the GetFused and SetFused extension methods ever changed the 'default' property name they used when none was explicitly provided, this code would continue to function correctly. 

If you wanted to take this even further, it would be possible to refactor these two methods as generic extension methods in `src/Sentry/Internal/ObjectExtensions.cs`... say `SetWeakFused` and `GetWeakFused` or something like that... that might be a little nicer actually - you could dispense with passing the `propertyName` argument entirely (and have it inferred from the generic typename, same as the existing non-weak GetFused/SetFused extension methods. 

I don't think this is a blocking change though. Let me know if you'd like to go down this path. If not, I can approve and merge as is.

jamescrosswell (commented):
@Ermabo huge thank you for the PR ❤️ 

I made one non-blocking stylistic suggestion. Let me know whether you'd like to implement that or if you'd prefer I approve and merge as is.

src/Sentry.OpenTelemetry/SentrySpanProcessor.cs
Ermabo: Thanks for the feedback! I like the idea of a `SetFusedActivity `to keep things symmetrical and prevent drift - made the change and pushed the commit.

The generic design could be neat if weakly-fused objects come up elsewhere, but as you said this package is deprecated, so probably not worth prioritizing right now.

src/Sentry.OpenTelemetry/SentrySpanProcessor.cs
```diff
@@ -134,13 +134,18 @@ private void CreateChildSpan(Activity data, ISpan parentSpan, SpanId? parentSpan
...
         if (span is SpanTracer spanTracer)
         {
             spanTracer.Origin = OpenTelemetryOrigin;
             spanTracer.StartTimestamp = data.StartTimeUtc;
-            spanTracer.IsFiltered = () => spanTracer.GetFused<Activity>() is { IsAllDataRequested: false, Recorded: false };
+            // Capture the sampling decision now; it is fixed at span creation. Reading it lazily from the
+            // fused Activity would fail once the Activity has been garbage-collected (weak ref returns null),
+            // causing a filtered span to be included in the transaction instead of dropped.
+            var isFiltered = data is { IsAllDataRequested: false, Recorded: false };
+            spanTracer.IsFiltered = () => isFiltered;
```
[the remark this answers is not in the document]
Ermabo: Fixed in 0ec4953, filter decision is now captured eagerly at span creation, so it no longer depends on the Activity being alive.

Ermabo:
@jamescrosswell happy to contribute! 

Both addressed and pushed:
- SetFusedActivity + shared const (your suggestion): 7eea926
- Eager filter-decision capture (Cursor Bugbot finding): 0ec4953

Updated the PR description to match. Ready for another look whenever 🙂

EDIT: the eager filter-decision capture (0ec4953) was reverted in 22a607, OnStart runs before OTel instrumentation filters clear the flags, so it captured filtered spans as unfiltered. Back to the lazy read; see follow-up comment below.

src/Sentry.OpenTelemetry/SentrySpanProcessor.cs
```diff
@@ -134,13 +134,15 @@ private void CreateChildSpan(Activity data, ISpan parentSpan, SpanId? parentSpan
...
+        // Fuse the Activity via a WeakReference so _map does not pin it, letting PruneFilteredSpans evict
+        // never-ended spans. #3198 intended weak refs here but SetFused stores the value strongly.
+        SetFusedActivity(span, data);
         if (span is SpanTracer spanTracer)
         {
             spanTracer.Origin = OpenTelemetryOrigin;
             spanTracer.StartTimestamp = data.StartTimeUtc;
-            spanTracer.IsFiltered = () => spanTracer.GetFused<Activity>() is { IsAllDataRequested: false, Recorded: false };
+            // Used to filter out spans that are not recorded when finishing a transaction.
+            spanTracer.IsFiltered = () => GetFusedActivity(spanTracer) is { IsAllDataRequested: false, Recorded: false };
```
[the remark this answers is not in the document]
Ermabo: This is a design tradeoff mentioned in the PR description. The GC timing condition may cause sampled-out spans to be sent but it's arguably better than to drop them, and it's a small chance this happens.

Better design would be to defer the capture to OnEnd but that is more code for an already deprecated package.

Ermabo:
Reverted the eager filter capture (22a607b), Bugbot was right, OnStart runs before OTel instrumentation filters, so it snapshotted too early. Back to the lazy read 😅

PR is now just the leak fix + SetFusedActivity cleanup. Ready for another look @jamescrosswell 😊

jamescrosswell (approved):
Awesome - thanks @Ermabo ❤️

Ermabo:
Cheers again for reviewing this and merging it @jamescrosswell! Do you have any ETA when this patch will be released?

Currently our high traffic service goes OOM every ~4 days which requires manual restart; would be great to get a new version asap

jamescrosswell:
@Ermabo hopefully this week. I wanted to get [this](https://github.com/getsentry/sentry-dotnet/pull/5412#event-28351704707) included in the same release (another important fix)... so just waiting for a colleague to review/approve - then we can ship 6.8.0.

jamescrosswell:
@Ermabo 6.8.0 is now available:
- https://github.com/getsentry/sentry-dotnet/releases/tag/6.8.0

Thanks again for your help with this.

Ermabo:
> @Ermabo 6.8.0 is now available:
> 
>     * https://github.com/getsentry/sentry-dotnet/releases/tag/6.8.0
> 
> 
> Thanks again for your help with this.

Cheers @jamescrosswell! Thank you for your help. I'm happy to have contributed 😊

PR #5409, opened 2026-07-20
ref: migrate Cocoa bindings from PrivateSentrySDKOnly to SentryObjCSDK.internal

### LINKED ISSUE #5331: Migrate from `PrivateSentrySDKOnly` to `SentryObjCSDK.internal`
[`sentry-cocoa` 9.19.0](<https://github.com/getsentry/sentry-cocoa/releases/tag/9.19.0>) introduces a new structured hybrid SDK API via `SentrySDK.internal` (Swift) and `SentryObjCSDK.internal` (ObjC). The old `PrivateSentrySDKOnly` class is deprecated and will be removed in the next major version.

**Implementation PR:** getsentry/sentry-cocoa#8097<br>**Design doc:** [SENTRY-INTERNAL-API.md](<https://github.com/getsentry/sentry-cocoa/blob/main/develop-docs/SENTRY-INTERNAL-API.md>)

## Benefits

* **Structured, discoverable API.** Methods are grouped by feature area instead of one flat class.
* **Deprecation-clean builds.** `PrivateSentrySDKOnly` has deprecation annotations — migrating eliminates warnings.
* **SPM / xcframework compatible.** Works with binary distribution — no exposed private headers.

## Current Usage

The .NET SDK uses Xamarin/MAUI bindings to `PrivateSentrySDKOnly` via `ApiDefinitions.cs`. The binding interface exposes the full `PrivateSentrySDKOnly` API surface and a patching script (`patch-cocoa-bindings.cs`) modifies the generated bindings.

The migration involves:

1. Generating new bindings against the `SentryObjCInternalApi` headers instead of `PrivateSentrySDKOnly.h`
2. Updating the C# wrapper code that calls the bindings
3. Updating `patch-cocoa-bindings.cs` for the new API shape

## Key Changes

The new ObjC API uses **instance methods** on `[SentryObjCSDK internal]` instead of **static methods** on `PrivateSentrySDKOnly`.

### SDK Metadata

| Before (binding) | After (binding) |
| -- | -- |
| `[Static] SetSdkName(string, string)` | `[SentryObjCSDK internal].sdk` → `SetName:version:` |
| `[Static] string SdkName { get; }` | `[SentryObjCSDK internal].sdk.name` |
| `[Static] string SdkVersionString { get; }` | `[SentryObjCSDK internal].sdk.versionString` |
| `[Static] NSDictionary ExtraContext { get; }` | `[SentryObjCSDK internal].sdk.extraContext` |
| `[Static] string InstallationID { get; }` | `[SentryObjCSDK internal].sdk.installationID` |

### Profiling

| Before (binding) | After (binding) |
| -- | -- |
| `[Static] StartProfilerForTrace(SentryId)` | `[[[SentryObjCSDK internal] profiling] startForTrace:]` |
| `[Static] CollectProfileBetween(ulong, ulong, SentryId)` | `[[[SentryObjCSDK internal] profiling] collectBetween:and:forTrace:]` |
| `[Static] DiscardProfilerForTrace(SentryId)` | `[[[SentryObjCSDK internal] profiling] discardForTrace:]` |

### Performance & Frames

| Before (binding) | After (binding) |
| -- | -- |
| `[Static] bool FramesTrackingMeasurementHybridSDKMode { get; set; }` | `[SentryObjCSDK internal].performance.framesTrackingHybridSDKMode` |
| `[Static] bool IsFramesTrackingRunning { get; }` | `[SentryObjCSDK internal].performance.isFramesTrackingRunning` |
| `[Static] SentryScreenFrames CurrentScreenFrames { get; }` | `[SentryObjCSDK internal].performance.currentScreenFrames` |

### Screenshot & View Hierarchy

| Before (binding) | After (binding) |
| -- | -- |
| `[Static] NSData[] CaptureScreenshots()` | `[[[SentryObjCSDK internal] screenshot] capture]` |
| `[Static] NSData CaptureViewHierarchy()` | `[[[SentryObjCSDK internal] viewHierarchy] capture]` |

### Replay

| Before (binding) | After (binding) |
| -- | -- |
| `[Static] ConfigureSessionReplayWith(converter, provider)` | `[[[SentryObjCSDK internal] replay] configureBreadcrumbConverter:screenshotProvider:]` |
| `[Static] CaptureReplay()` (void) | `[[[SentryObjCSDK internal] replay] capture]` → returns `BOOL` |
| `[Static] string ReplayId { get; }` | `[SentryObjCSDK internal].replay.replayId` |

### Breadcrumbs & Users

| Before (binding) | After (binding) |
| -- | -- |
| `[Static] UserWithDictionary(NSDictionary)` | `[[[SentryObjCSDK internal] user] fromDictionary:]` |
| `[Static] BreadcrumbWithDictionary(NSDictionary)` | `[[[SentryObjCSDK internal] breadcrumbs] fromDictionary:]` |

## Affected Files (as of latest `main`)

* `src/Sentry.Bindings.Cocoa/ApiDefinitions.cs` — `PrivateSentrySDKOnly` binding interface
* `scripts/patch-cocoa-bindings.cs` — binding patching script referencing `PrivateSentrySDKOnly` methods

## Checklist

- [ ] Generate new Xamarin/MAUI bindings against `SentryObjCInternalApi` headers
- [ ] Replace `PrivateSentrySDKOnly` interface in `ApiDefinitions.cs` with `SentryObjCInternalApi` + sub-object interfaces
- [ ] Update `patch-cocoa-bindings.cs` for new API shape
- [ ] Update C# wrapper code to use new binding methods
- [ ] Handle the static→instance method change (old API was all `+` class methods, new API uses instance methods on `[SentryObjCSDK internal]`)
- [ ] Bump minimum `sentry-cocoa` dependency to `9.19.0`+
- [ ] Run full test suite to verify no regressions

jamescrosswell:
Migrates the iOS / Mac Catalyst hybrid-SDK bindings off the deprecated `PrivateSentrySDKOnly` (static methods, slated for removal in the next sentry-cocoa major) to the structured `SentryObjCSDK.internal` API introduced in sentry-cocoa 9.19. `main` already pins sentry-cocoa 9.22.0, so no dependency bump is required.

Key incremental/separate commits, in case this helps for review purposes:

- **Phase 1** — regenerate bindings against the `SentryObjC` headers (build-sentry-cocoa.sh, generate-cocoa-bindings.ps1, patch-cocoa-bindings.cs, ApiDefinitions.cs). `PrivateSentrySDKOnly` kept in parallel - [e532a2f](https://github.com/getsentry/sentry-dotnet/pull/5409/commits/e532a2ff0ee9920ae3a7c7d871eb9658c202ce6f)
- **Phase 2** — migrate the 5 call sites and the `SentryCocoaHybridSdk` alias to the new instance-based API [cada275](https://github.com/getsentry/sentry-dotnet/pull/5409/commits/cada27595020cf7f70138bd27b70255aecf1698c)
- **Phase 3** — remove `PrivateSentrySDKOnly` from the binding, patch/generate scripts, and props [55ff0b8](https://github.com/getsentry/sentry-dotnet/pull/5409/commits/55ff0b8580666c952cb0d36f28d983b065d3b042)

### Note: We build from source now

- The .NET-consumed surface is only 5 call sites, via the `SentryCocoaHybridSdk` MSBuild alias (`SetSdkName`, `SetTrace`, `StartProfilerForTrace`, `CollectProfileBetween`, `IgnoreNextSignal`). Every one has a direct equivalent within `[SentryObjCSDK internal]`.
- ⚠️ **The published `SentryObjC-Dynamic.xcframework` is not usable as-is.** It is fully self-contained (~4.4 MB/slice): it *defines* `SentryHubInternal`, `SentryClientInternal`, `SentryObjCSDK` and the SentrySwift classes, and links no external `Sentry.framework`. Bundling it alongside the `Sentry.framework` we already ship would put **two full copies of the SDK** in the app, with split runtime state — the managed bindings would talk to one instance and the new `internal` calls to another.
- sentry-cocoa exposes a buildable **`SentryObjC` Xcode scheme**; it builds cleanly and ships the `SentryObjCSDK` / `SentryObjCInternalApi` public headers.
- ✅ **Building `SentryObjC` + `SentryObjCCompat` from source produces thin two frameworks** (136 KB + 1.2 MB) that only *define* the new additive `SentryObjC*` wrapper types and *import* the real SDK classes (`_SentryScope`, `_SentryBreadcrumb`, `_SentryUser`, …) as undefined symbols resolved at load time from the existing `Sentry.framework` via `@rpath`. No duplication; the new API shares the same SDK instance as the current bindings.

### Issues

Closes #5331


scripts/generate-cocoa-bindings.ps1
```diff
@@ -164,12 +164,21 @@ else
...
+# the deprecated PrivateSentrySDKOnly. We only bind the entry point (SentryObjCSDK), the internal
+# API surface, and the two id types the .NET wrappers pass; patch-cocoa-bindings.cs trims everything
+# else these headers pull in via forward declarations.
 Write-Output 'Generating bindings with Objective Sharpie.'
 sharpie bind -sdk $iPhoneSdkVersion `
     -scope "$CocoaSdkPath" `
     "$HeadersPath/Sentry.h" `
     "$HeadersPath/Sentry-Swift.h" `
-    "$HeadersPath/PrivateSentrySDKOnly.h" `
+    "$HeadersPath/SentryObjCSDK.h" `
```
jamescrosswell, line 164:
```suggestion
    # SentryObjC.h* exposes the structured hybrid API (SentryObjCSDK.internal)
    "$HeadersPath/SentryObjCSDK.h" `
```

scripts/generate-cocoa-bindings.ps1
```diff
@@ -164,12 +164,21 @@ else
 }
 
 # Generate bindings
+# The SentryObjC* headers expose the structured hybrid API (SentryObjCSDK.internal) that replaces
+# the deprecated PrivateSentrySDKOnly. We only bind the entry point (SentryObjCSDK), the internal
+# API surface, and the two id types the .NET wrappers pass; patch-cocoa-bindings.cs trims everything
+# else these headers pull in via forward declarations.
```
jamescrosswell:
```suggestion
```

scripts/patch-cocoa-bindings.cs
```diff
@@ -116,17 +106,23 @@
...
-    // SentryAppStartMeasurement is not whitelisted
-    .RemoveProperty("PrivateSentrySDKOnly", "*AppStartMeasurement*")
     // Minimize SentryDependencyContainer
     .RemoveMethod("SentryDependencyContainer", "*")
     .KeepProperties("SentryDependencyContainer", "SharedInstance", "DebugImageProvider")
     // SentryUserFeedbackConfiguration is not whitelisted
     .RemoveProperty("SentryOptions", "ConfigureUserFeedback")
     .RemoveProperty("SentryOptions", "UserFeedbackConfiguration")
+    // SentryObjCSDK.internal is the entry point to the new hybrid API (replacing PrivateSentrySDKOnly).
+    // We only bind the `internal` accessor; every other SentryObjCSDK member references SentryObjC*
```
jamescrosswell:
```suggestion
    // SentryObjCSDK.internal is the entry point to the API for hybrid SDKs
    // We only bind the `internal` accessor; every other SentryObjCSDK member references SentryObjC*
```

scripts/patch-cocoa-bindings.cs
```diff
@@ -116,17 +106,23 @@
...
+    // types we don't whitelist.
+    .KeepProperties("SentryObjCSDK", "Internal")
+    .RemoveMethod("SentryObjCSDK", "*")
+    // On the internal API the .NET wrappers only use the SDK-metadata and profiling sub-objects plus
+    // setTrace / ignoreNextSignal. The remaining accessors return sub-API/options types that aren't bound.
```
jamescrosswell:
```suggestion
```

src/Sentry/Platforms/Cocoa/CocoaProfiler.cs
```diff
@@ -36,14 +36,18 @@ public void Finish()
...
+        if (collected is null)
         {
             _options.LogWarning("Trace {0} collected profile payload is null", _traceId);
             return null;
         }
         _options.LogDebug("Trace {0} profile payload collected", _traceId);
 
+        // The SentryObjCSDK.internal profiling API returns an immutable NSDictionary (the old
+        // PrivateSentrySDKOnly API returned a mutable one), so copy it before mutating below.
+        var payload = (NSMutableDictionary)collected.MutableCopy();
```
jamescrosswell:
```suggestion
        // The SentryObjCSDK.internal profiling API returns an immutable NSDictionary, so copy it before mutating below.
        var payload = (NSMutableDictionary)collected.MutableCopy();
```

src/Sentry.Bindings.Cocoa/Sentry.Bindings.Cocoa.csproj
```diff
@@ -54,6 +59,9 @@
     <ObjcBindingApiDefinition Include="PrivateApiDefinitions.cs" />
     <ObjcBindingCoreSource Include="StructsAndEnums.cs" />
     <NativeReference Include="$(SentryCocoaFramework)" Kind="Framework" />
+    <!-- SentryObjCCompat before SentryObjC (dependency order); both link Sentry.framework above. -->
```
jamescrosswell: Why specifically does the order matter? What would break if we reversed the order (and why)?

scripts/build-sentry-cocoa.sh
```diff
@@ -83,9 +115,35 @@ xcodebuild -create-xcframework \
...
+        -framework "./Carthage/output-objc-maccatalyst.xcarchive/Products/Library/Frameworks/$fw.framework" \
+        -output "./Carthage/Build-maccatalyst/$fw.xcframework"
+done
+echo "::endgroup::"
+
+# Xcode embeds each framework's dynamic dependencies under <Framework>.framework/Frameworks/ (e.g.
+# SentryObjC.framework/Frameworks/SentryObjCCompat.framework/Frameworks/Sentry.framework). We bundle
+# Sentry, SentryObjCCompat and SentryObjC as separate NativeReferences - each embedded into the
+# consuming app - so those nested copies are redundant, and their deep paths blow past NuGet's path
+# length limit (NU5123). Strip them; the frameworks resolve each other via @rpath at the app level.
```
jamescrosswell:
```suggestion
# length limit (NU5123). To fix that, we strip them. The frameworks resolve each other via @rpath at the app level instead/anyway.
```

src/Sentry.Bindings.Cocoa/Sentry.Bindings.Cocoa.csproj
```diff
@@ -54,6 +59,11 @@
     <ObjcBindingApiDefinition Include="PrivateApiDefinitions.cs" />
     <ObjcBindingCoreSource Include="StructsAndEnums.cs" />
     <NativeReference Include="$(SentryCocoaFramework)" Kind="Framework" />
+    <!-- The thin hybrid-API frameworks: SentryObjC depends on SentryObjCCompat, which depends on
+         Sentry.framework above. They're dynamic frameworks that resolve each other via @rpath at
+         load time, so the order here is just readability (dependencies first), not a requirement. -->
```
jamescrosswell:
```suggestion
```

src/Sentry.Bindings.Cocoa/Sentry.Bindings.Cocoa.csproj
```diff
@@ -24,6 +24,11 @@
   <!-- Override values for local Cocoa SDK builds -->
   <PropertyGroup Condition="Exists('$(SentryCocoaCache).git')">
     <SentryCocoaFramework>$(SentryCocoaCache)Carthage\Build-$(TargetPlatformIdentifier)\Sentry.xcframework</SentryCocoaFramework>
+    <!-- Structured hybrid API (SentryObjCSDK.internal). These two thin frameworks dynamically link
+         Sentry.framework above; build-sentry-cocoa.sh builds them from source. They are only produced
+         on the build-from-source path (there is no non-duplicating released artifact - see #5331). -->
```
jamescrosswell:
```suggestion
    <!-- Structured hybrid API (SentryObjCSDK.internal). These two thin frameworks dynamically link
         Sentry.framework above; build-sentry-cocoa.sh builds them from source. →
```

src/Sentry/Platforms/Cocoa/CocoaProfiler.cs
```diff
@@ -36,14 +36,17 @@ public void Finish()
...
+        var collected = SentryCocoaHybridSdk.Internal.Profiling.CollectBetweenStartTime(_startTimeNs, _endTimeNs, _cocoaTraceId);
+        if (collected is null)
         {
             _options.LogWarning("Trace {0} collected profile payload is null", _traceId);
             return null;
         }
         _options.LogDebug("Trace {0} profile payload collected", _traceId);
 
+        // The SentryObjCSDK.internal profiling API returns an immutable NSDictionary, so copy it before mutating below.
+        var payload = (NSMutableDictionary)collected.MutableCopy();
```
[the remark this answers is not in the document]
jamescrosswell: False positive:

1. **This exact line already runs on a real iOS runtime in CI and passes.** The device test `Profiler_RunningUnderFullClient_SendsProfileData` exercises `Collect()` end-to-end (`CollectBetweenStartTime` → `MutableCopy()` → this cast → the mutations below). It has passed on the last two green `ios-tests` runs — and it's the same test that caught the genuine runtime bug this line fixes (`NotSupportedException` when mutating the immutable payload), so we know it surfaces runtime exceptions at precisely this spot. An `InvalidCastException` here could not pass that test.
2. **The sibling line two lines below has shipped this exact pattern for years.** `payload["transaction"]?.MutableCopy() as NSMutableDictionary` predates this PR; if `MutableCopy()`'s managed wrapper couldn't be treated as `NSMutableDictionary`, that `as`-cast would have yielded `null` and profiling would always have bailed with "payload doesn't have transaction information". It doesn't.
3. **Mechanically, the cast is what the runtime bridge produces.** The binding's *static* return type of `MutableCopy()` is `NSObject`, but the *runtime wrapper* is created by `Runtime.GetNSObject`, which maps the native class (a private `NSMutableDictionary` subclass like `__NSDictionaryM`) to its nearest registered managed ancestor — `NSMutableDictionary`. The static return type being `NSObject` is not a "known issue" that makes the cast fail; it's just why a cast is needed at all.

src/Sentry.Bindings.Cocoa/Sentry.Bindings.Cocoa.csproj
```diff
@@ -54,6 +58,8 @@
     <ObjcBindingApiDefinition Include="PrivateApiDefinitions.cs" />
     <ObjcBindingCoreSource Include="StructsAndEnums.cs" />
     <NativeReference Include="$(SentryCocoaFramework)" Kind="Framework" />
+    <NativeReference Include="$(SentryCocoaObjCCompatFramework)" Kind="Framework" Condition="'$(SentryCocoaObjCCompatFramework)' != ''" />
+    <NativeReference Include="$(SentryCocoaObjCFramework)" Kind="Framework" Condition="'$(SentryCocoaObjCFramework)' != ''" />
```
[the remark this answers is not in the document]
jamescrosswell: We should probably remove the prebuilt path entirely... since it no longer works.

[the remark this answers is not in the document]
jamescrosswell: Fixed in 5025efe3 by removing the pre-built path entirely (per the comment above). The download path could only ever supply `Sentry-Dynamic.xcframework` — there is no non-duplicating released artifact for the `SentryObjC`/`SentryObjCCompat` frameworks (the published `SentryObjC-Dynamic.xcframework` is self-contained and would embed a second copy of the SDK) — so if activated it would produce exactly the runtime failure described. It was already dormant (`modules/sentry-cocoa.properties` doesn't exist and CI checks out the submodule). `_DownloadCocoaSDK`, the released-builds property group and `SanitizeSentryCocoaFramework` are gone; the build now fails with a clear error if the submodule is missing.

src/Sentry.Bindings.Cocoa/buildTransitive/Sentry.Bindings.Cocoa.targets
```diff
@@ -45,7 +45,9 @@
   -->
   <Target Name="_SentryRemoveSimulatorResourcesForHotRestartBuilds" AfterTargets="_CopyLocalBindingResources" Condition="'$(IsHotRestartBuild)' == 'true'">
     <ItemGroup>
-      <ReferenceCopyLocalPaths Remove="@(ReferenceCopyLocalPaths)" Condition="$([System.String]::new('%(Identity)').Contains('Sentry.xcframework\ios-arm64_x86_64-simulator'))" />
+      <!-- Only the frameworks bundled with this package - enumerated explicitly because this file is
+           imported transitively into consuming apps, where ReferenceCopyLocalPaths spans all packages. -->
```
jamescrosswell:
```suggestion
      <!-- Only strip from the frameworks bundled with this package - this file is
           imported transitively into consuming apps -->
```

src/Sentry.Bindings.Cocoa/Sentry.Bindings.Cocoa.csproj
```diff
@@ -113,10 +74,12 @@
...
 
-  <!-- Choose between download and build -->
   <Target Name="_SetupCocoaSDK"
-          DependsOnTargets="_DownloadCocoaSDK;_BuildCocoaSDK;_GenerateSentryCocoaBindings;SanitizeSentryCocoaFramework"
-          Condition="$([MSBuild]::IsOSPlatform('OSX'))" />
+          DependsOnTargets="_BuildCocoaSDK;_GenerateSentryCocoaBindings"
+          Condition="$([MSBuild]::IsOSPlatform('OSX'))">
+    <Error Condition="!Exists('$(SentryCocoaCache).git')"
+           Text="The modules/sentry-cocoa submodule is missing. Run 'git submodule update --init --recursive' - the Cocoa SDK is built from source (see #5331)." />
+  </Target>
```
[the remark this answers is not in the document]
jamescrosswell: False positive in practice, but the fragility it points at was real — addressed in 1e8b02a0.

**Why the error was actually reachable:** with no `.git`, `SentryCocoaBindingInputs` is never defined, so MSBuild skips `_GenerateSentryCocoaBindings` under its empty-`Inputs` rule (and `_BuildCocoaSDK` is skipped by its own condition), leaving the `Error` in `_SetupCocoaSDK` to fire. Verified empirically by simulating a missing submodule locally — the build fails with the intended message:

> `error : The modules/sentry-cocoa submodule is missing. Run 'git submodule update --init --recursive' - the Cocoa SDK is built from source (see #5331).`

**Hardening anyway:** reachability shouldn't hinge on a non-obvious skip rule, so the check now lives in a dedicated `_EnsureCocoaSubmodule` target that runs first in `_SetupCocoaSDK`'s `DependsOnTargets` chain. Both paths re-verified: missing submodule → friendly error; normal build → succeeds.

src/Sentry.Bindings.Cocoa/StructsAndEnums.cs
```diff
@@ -175,3 +175,18 @@ internal enum SentryTransactionNameSource : long
...
+    Custom
+}
+
+[Native]
+internal enum SentryObjCLastRunStatus : long
+{
+    Unknown = 0,
+    DidNotCrash,
+    DidCrash
+}
```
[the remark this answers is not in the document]
jamescrosswell: Valid finding — fixed in 247a9f0c. Sharpie emits these because the SentryObjC headers reference them, but the members that used them are trimmed by `patch-cocoa-bindings.cs`, leaving dead enums. Added a `RemoveEnum` filter and dropped all `SentryObjC*` enums (the hybrid-API surface we bind uses none), then regenerated — `StructsAndEnums.cs` loses exactly the two dead enums and the binding builds clean.

scripts/build-sentry-cocoa.sh
```diff
@@ -67,6 +67,38 @@ xcodebuild -create-xcframework \
     -output ./Carthage/Build-ios/Sentry.xcframework
 echo "::endgroup::"
 
+# The SentryObjC scheme adds the structured hybrid API (SentryObjCSDK.internal), which the .NET
```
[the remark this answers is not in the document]
jamescrosswell: Valid for local developer caches — fixed in the commit above. One correction to the scenario: CI never caches the Carthage output (the only native build caches are for sentry-native), so CI always builds the full recipe; the stale-stamp case only bites a local checkout that built cocoa with the pre-PR script at the same submodule SHA. The stamp now includes the build script's own SHA-256 alongside the submodule SHA (mirroring the sentry-native cache key convention), so any recipe change invalidates cached output. Verified both paths locally: old-format stamp → mismatch → rebuild; current-recipe stamp → instant early-exit.

philprime (approved):
LGTM from a code-level review perspective.

**Regarding the duplicate SDKs:**

The published `SentryObjC` target is designed as a full replacement for the `Sentry` target. Therefore, bundling both published frameworks would result in duplicate SDK framework slices. Given that the .NET SDK still depends on `Sentry`, building `SentryObjC` and `SentryObjCCompat` from source as thin frameworks seems like a reasonable interim solution for this pull request.

As a follow-up, I propose migrating the remaining bindings from `Sentry` to `SentryObjC`. Most types should have one-to-one replacements, and this would allow the SDK to ship only `SentryObjC`. We plan to remove Objective-C compatibility from the `Sentry` target in the next major release later this year, so completing that migration will be important.

jamescrosswell:
> As a follow-up, I propose migrating the remaining bindings from `Sentry` to `SentryObjC`. Most types should have one-to-one replacements, and this would allow the SDK to ship only `SentryObjC`. We plan to remove Objective-C compatibility from the `Sentry` target in the next major release later this year, so completing that migration will be important.

Ah, great - thanks @philprime !

I've created https://github.com/getsentry/sentry-dotnet/issues/5444 as a follow up for the remaining work (which will need to be done before the next major for you guys):
- https://github.com/getsentry/sentry-dotnet/issues/5444

src/Sentry.Bindings.Cocoa/buildTransitive/Sentry.Bindings.Cocoa.targets
```diff
@@ -45,7 +45,9 @@
   -->
   <Target Name="_SentryRemoveSimulatorResourcesForHotRestartBuilds" AfterTargets="_CopyLocalBindingResources" Condition="'$(IsHotRestartBuild)' == 'true'">
     <ItemGroup>
-      <ReferenceCopyLocalPaths Remove="@(ReferenceCopyLocalPaths)" Condition="$([System.String]::new('%(Identity)').Contains('Sentry.xcframework\ios-arm64_x86_64-simulator'))" />
+      <!-- Only strip from the frameworks bundled with this package - this file is
+           imported transitively into consuming apps -->
+      <ReferenceCopyLocalPaths Remove="@(ReferenceCopyLocalPaths)" Condition="$([System.String]::new('%(Identity)').Contains('Sentry.xcframework\ios-arm64_x86_64-simulator')) Or $([System.String]::new('%(Identity)').Contains('SentryObjC.xcframework\ios-arm64_x86_64-simulator')) Or $([System.String]::new('%(Identity)').Contains('SentryObjCCompat.xcframework\ios-arm64_x86_64-simulator'))" />
```
jpnurmi: maybe worth something like `Sentry(ObjC|ObjCCompat)?\.xcframework\\ios-arm64_x86_64-simulator` pattern match?

CONTRIBUTING.md
```diff
@@ -194,27 +194,23 @@ should be updated from the main branch and the `modules/make-internal.sh` script
...
-```
-
-To switch back to the pre-built SDK, delete the `modules/sentry-cocoa` directory
-and let the next build download the pre-built SDK again:
-
-```sh
-$ rm -rf modules/sentry-cocoa
-$ dotnet build ... # downloads pre-built Cocoa SDK into modules/sentry-cocoa
+$ cd modules/sentry-cocoa
+$ git fetch origin && git checkout <tag-or-sha>
```
jpnurmi: Side-note: arbitrary checkouts worked with a standalone clone, but I recall the automatic submodule update in `before.Sentry.sln.targets` fights back when trying to do this with submodules:

https://github.com/getsentry/sentry-dotnet/blob/65c0c59fb888c0a2b7fa8b4833ed53968ad3d44f/before.Sentry.sln.targets#L1-L6

jpnurmi (approved):
I couldn't find anything blocking either. Just a couple of nitpicks. 😅

src/Sentry.Bindings.Cocoa/buildTransitive/Sentry.Bindings.Cocoa.targets
jamescrosswell: Done in 6e575395 — collapsed to `Sentry(ObjCCompat|ObjC)?\.xcframework[\/]ios-arm64_x86_64-simulator`. Used a `[\/]` separator instead of a literal `\` so it's testable off-Windows (on macOS MSBuild normalizes item-spec backslashes to `/`); functionally identical on the Hot Restart / Windows path. Verified in MSBuild that it strips exactly the three simulator slices and leaves device slices + other packages' xcframeworks untouched.

CONTRIBUTING.md
jamescrosswell: Good catch — confirmed: with a submodule, checking out a ref and building reverts it, because the solution build's `RestoreSubmodule` runs `git submodule update`, which restores the commit recorded in the superproject index. Staging the gitlink (`git add modules/sentry-cocoa`) after checkout makes `git submodule update` a no-op, so the ref sticks (verified both ways). Updated the CONTRIBUTING instructions in the same commit to include the staging step and explain why.

src/Sentry.Bindings.Cocoa/buildTransitive/Sentry.Bindings.Cocoa.targets
jamescrosswell: [6e57539](https://github.com/getsentry/sentry-dotnet/pull/5409/commits/6e5753956fb03c626e9677566cddc1bbe09bbcd7)

PR #5412, opened 2026-07-20
fix: match rate limits by data category not envelope item type

### LINKED ISSUE #4535: Review and update Rate Limiting data categories

jamescrosswell:
## Summary

Rate limits in the `X-Sentry-Rate-Limits` header are keyed by **data category** (`error`, `monitor`, `log_item`, `metric_bucket`, …) in Relay/Rust.

However since 2020 our SDK matched them against the **envelope item type** (`event`, `check_in`, `statsd`, `log`…). Those two only line up by coincidence for some types, so limits silently didn't match for the rest — the SDK ignored the backoff and kept sending.

- `error` was broken from day one: Relay has always sent `error`, we compared it to the item type `event`. (Not a Relay change — we were wrong from our first rate-limit commit.)
- `monitor` (check-ins) and `log_item` (logs) broke the moment we added those features: the category already existed in Relay, we just never mapped the item type to it.
- `metric_bucket` was patched in 2024 with a one-off special case rather than a real fix.

Additionally, since we'd keyed client reports the default category, if Default ever got rate limited, client reports would erroneously be dropped. 

Ultimately we had **two parallel notions** of an item's category — a correct `EnvelopeItem.DataCategory` map (used for client-report accounting) and a separate raw-item-type string comparison in the rate-limiter. They were never reconciled, so every new item type whose type-string ≠ category-name reintroduced the bug.

## Fix

The fix deletes the item-type comparison and matches on `item.DataCategory`. Any new item type now only needs its `DataCategory` mapping (which it needs anyway for client reports), and rate limiting follows automatically, so **_this class of bug can't silently reappear_**.

### Changelog Entry

The SDK was incorrectly ignoring server rate limits for errors, check-ins, and logs

### Issues

Closes #4535

### DIFF
--- a/src/Sentry/Internal/Http/RateLimitCategory.cs
+++ b/src/Sentry/Internal/Http/RateLimitCategory.cs
@@ -17,20 +17,10 @@ internal class RateLimitCategory : IEquatable<RateLimitCategory>
             return true;
         }
 
-        var type = item.TryGetType();
-        if (string.IsNullOrWhiteSpace(type))
-        {
-            return false;
-        }
-
-        return type switch
-        {
-            EnvelopeItem.TypeValueMetric =>
-                // Metrics are a bit unique - the envelope item type is `statsd` but the category is `metric_bucket`
-                string.Equals(Name, "metric_bucket", StringComparison.OrdinalIgnoreCase),
-            // For most reporting categories, the envelope item type matches the client report category
-            _ => string.Equals(Name, type, StringComparison.OrdinalIgnoreCase)
-        };
+        // Rate limits are keyed by data category (e.g. "error", "monitor", "metric_bucket"), which is not
+        // always the same as the envelope item type (e.g. "event", "check_in", "statsd"). Compare against the
+        // item's data category so limits for those categories are honoured rather than silently ignored.
+        return string.Equals(Name, item.DataCategory.ToString(), StringComparison.OrdinalIgnoreCase);
     }
 
     public bool Equals(RateLimitCategory? other)
--- a/src/Sentry/Internal/Http/RateLimitCategory.cs
+++ b/src/Sentry/Internal/Http/RateLimitCategory.cs
@@ -10,18 +10,9 @@ internal class RateLimitCategory : IEquatable<RateLimitCategory>
 
     public RateLimitCategory(string name) => Name = name;
 
-    public bool Matches(EnvelopeItem item)
-    {
-        if (IsMatchAll)
-        {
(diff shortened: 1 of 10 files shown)


src/Sentry/Internal/Http/RateLimitCategory.cs
```diff
@@ -17,20 +17,10 @@ public bool Matches(EnvelopeItem item)
...
-        {
-            EnvelopeItem.TypeValueMetric =>
-                // Metrics are a bit unique - the envelope item type is `statsd` but the category is `metric_bucket`
-                string.Equals(Name, "metric_bucket", StringComparison.OrdinalIgnoreCase),
-            // For most reporting categories, the envelope item type matches the client report category
-            _ => string.Equals(Name, type, StringComparison.OrdinalIgnoreCase)
-        };
+        // Rate limits are keyed by data category (e.g. "error", "monitor", "metric_bucket"), which is not
+        // always the same as the envelope item type (e.g. "event", "check_in", "statsd"). Compare against the
+        // item's data category so limits for those categories are honoured rather than silently ignored.
```
jamescrosswell: Keep this concise...
```suggestion
        // Rate limits are keyed by data category (e.g. "error", "monitor", "metric_bucket")
```

src/Sentry/Internal/Http/RateLimitCategory.cs
jamescrosswell: This could be simplified to a lambda that simply returns `IsMatchAll || string.Equals(...`

src/Sentry/Protocol/Envelopes/EnvelopeItem.cs
```diff
@@ -51,9 +51,22 @@ public sealed class EnvelopeItem : ISerializable, IDisposable
...
+
+        // The "check_in" item type corresponds to the "monitor" data category
+        TypeValueCheckIn => DataCategory.Monitor,
+
+        // The "statsd" item type corresponds to the "metric_bucket" data category
+        TypeValueMetric => DataCategory.MetricBucket,
+
+        // Client reports are SDK telemetry that Relay never rate-limits, so map them to a data category
+        // ("internal") that never appears in a rate-limit header. This keeps them from being dropped by a
+        // limit for the "default" category (which is what the fallback below would otherwise assign).
```
jamescrosswell:
```suggestion
        // Client reports are SDK telemetry that Relay never rate-limits - mapped to internal
```

Litarnus (approved):

Issue #5444, opened 2026-07-24
Migrate remaining Cocoa bindings from Sentry.framework to SentryObjC (drop Sentry.xcframework)


jamescrosswell:
## Background

The .NET SDK wraps the native sentry-cocoa SDK via the `Sentry.Bindings.Cocoa` project, which generates C# bindings from the framework's Objective-C headers. Historically it bound the classic `Sentry*` ObjC classes that live in `Sentry.framework`, plus a deprecated grab-bag of hybrid-only hooks (`PrivateSentrySDKOnly`).

getsentry/sentry-dotnet#5331 (PR getsentry/sentry-dotnet#5409) migrated the **hybrid/private** surface to sentry-cocoa's new structured `SentryObjCSDK.internal` API. But that was only the private hooks. Everything else — options, SDK lifecycle, the whole event model, scope, enums, debug images — still binds the classic `Sentry*` ObjC types, so we still ship `Sentry.framework` alongside the new `SentryObjC`/`SentryObjCCompat` frameworks.

The sentry-cocoa team is removing the Objective-C surface from the `Sentry` target in **v10 (\~Q3/Q4 2026)**. Once that lands, the classic `Sentry*` ObjC classes we bind today will be gone, so this migration is needed to support cocoa v10.

### Rebinding

Migrate the remaining bound `Sentry*` ObjC surface to `SentryObjC*`. This is a mechanical rebind — the `SentryObjC*` types already exist with matching shape (options + callbacks, SDK entry point, the event model, enums, and debug images via `[SentryObjCSDK internal].debug`).

### Scope → event enrichment (currently a private-API hack)

`CocoaEventProcessor` enriches .NET-captured events with the native SDK's contexts (device, OS, app…). It does this by creating a throwaway native event and calling `scope.applyToEvent(event, maxBreadcrumb: 0)` — a **private** cocoa API bound via `PrivateApiDefinitions.cs`. (This is the last remaining instance of the private-API pattern getsentry/sentry-dotnet#5331 set out to eliminate; the code has carried a "this is a private API, find a better way" TODO since cocoa 8.0.)

`SentryObjCScope` exposes `serialize` but no `applyToEvent`.

**To be decided:** Align with the sentry-cocoa team on what shape the new API should have so that we can continue to enrich .NET captured events with Cocoa SDK context (there might be an easier way to do this).

#### No exported `SentrySerializable` protocol

[...]

PR #5450, opened 2026-07-27
ref: migrate Cocoa bindings to the SentryObjC surface

jamescrosswell:
## Summary

Migrates `Sentry.Bindings.Cocoa` from the classic `Sentry*` Objective-C surface (in `Sentry.framework`) to the new `SentryObjC*` surface, and — now that sentry-cocoa **9.26.0** ships the structured hybrid serialization APIs we needed — **removes all remaining use of the classic APIs**. This branch merges `main` (which bumped cocoa to 9.26.0).

Addresses the first "Done when" of #5444: **no binding references classic `Sentry*` ObjC types.** (See _Remaining_ below for the second bullet.)

## Notes for reviewers

It looks like this is a massive change - roughly **+1500 / -2900** but:
- **+1,316 / -2,592** is in the `ApiDefinitions.cs` and `StructsAndEnums.cs` files generated by csharpie
- Of the remaining **+184 / -308**, most of it is just renaming types from `CocoaSdk.Sentry*` to `CocoaSdk.SentryObjC*`

The only bits that really deserve any attention then are:

#### scripts/*
  - generate-cocoa-bindings.ps1 now feeds Objective Sharpie the SentryObjC.h umbrella instead of the classic Sentry.h
  - patch-cocoa-bindings.cs retargets the keep-list/rules to the SentryObjC* surface

#### Sentry.Bindings.Cocoa.csproj:120

Added `Exists(…SupportDelegates.g.cs)` so the "make generated delegates internal" workaround only runs when bgen actually emits that file... prevents a potential build failure

#### CocoaExtensions.cs:29-35

Removed the `obj is ISentrySerializable → serialize()` check since everything reaching this method is already a plain wire-format NSDictionary now (from serializeEvent/serializedContexts)

#### SerializableNSObject.cs:28-33

Same again

#### CocoaEventProcessor.cs
Use of the new structured SentryCocoaHybridSdk.Internal.Scope.SerializedContexts API instead of creating a throwaway event to infer these

 #### PrivateApi.cs

Goes away, since everything we need is now available via the new internal APIs. 

## What changed

**Binding (`Sentry.Bindings.Cocoa`)**
- `generate-cocoa-bindings.ps1` binds the `SentryObjC.h` umbrella only (no classic `Sentry.h` / `Sentry-Swift.h`).
- `patch-cocoa-bindings.cs` keeps the `SentryObjC*` surface the .NET SDK consumes, the public SDK entry points (`startWithOptions` / `configureScope` / `crash` / `crashedLastRun`), and the hybrid `internal` accessor — now including `internal.scope` and `internal.serializer`.
- **`PrivateApiDefinitions.cs` deleted.** The classic `Sentry.framework` holdouts it used to hand-bind (`SentrySDK` / `SentryScope` / `SentryEvent` / `SentrySerializable`) are gone.

**Adopting the new hybrid APIs (cocoa 9.26.0)**
- **Event → managed conversion** (`ToSentryEvent`, used by native before-send / event processors) now serializes via `SentryObjCSDK.internal.serializer.serializeEvent(event)` and deserializes into a managed `SentryEvent` — the full-fidelity wire-format round-trip, no property mapping.
- **Scope enrichment** (`CocoaEventProcessor`) now reads `SentryObjCSDK.internal.scope.serializedContexts()` instead of creating a throwaway native event and calling the private `applyToEvent`. The native `trace` context is dropped so it can't overwrite the managed event's own trace linkage.
- Removed the now-dead `SentrySerializable` type-tests in `CocoaExtensions`/`SerializableNSObject` and the `SentryCocoaLegacySdk` alias.

## Remaining (second "Done when")

`Sentry.xcframework` is **still shipped**, because the from-source `SentryObjC` / `SentryObjCCompat` thin frameworks dynamically link `Sentry.framework` at runtime (they don't embed the SDK). No C# binding references it any longer — it's now purely a runtime link dependency. Physically dropping `Sentry.xcframework` means switching to the self-contained prebuilt `SentryObjC-Dynamic.xcframework` (keeping a from-source toggle for debugging) — tracked as a separate packaging follow-up in #5492.

Related: #5444, #5492


bitsandfoxes:
bless your kind heart for tackling this

src/Sentry.Bindings.Cocoa/Sentry.Bindings.Cocoa.csproj
```diff
@@ -116,8 +116,12 @@
     <Exec Condition="'$(GITHUB_ACTIONS)' == 'true'" Command="pwsh ../../scripts/dirty-check.ps1 -PathToCheck $(MSBuildThisFileDirectory)" />
   </Target>
 
-  <!-- Workaround for https://github.com/xamarin/xamarin-macios/issues/15299 -->
-  <Target Name="_SetGeneratedSupportDelegatesInternal" BeforeTargets="CoreCompile" Condition="$([MSBuild]::IsOSPlatform('OSX'))"
+  <!-- Workaround for https://github.com/xamarin/xamarin-macios/issues/15299.
+       Only runs when bgen actually emits SupportDelegates.g.cs (i.e. when the binding declares
+       named delegate types). The SentryObjC surface uses only inline Func/Action callbacks, whose
+       trampolines bgen emits as internal delegates in Trampolines.g.cs, so no patching is needed. -->
```
jamescrosswell:
```suggestion
  <!-- Workaround for https://github.com/xamarin/xamarin-macios/issues/15299. -->
```

src/Sentry/Platforms/Cocoa/Extensions/CocoaExtensions.cs
```diff
@@ -302,18 +302,63 @@ public static void CopyToCocoaSentryEvent(this SentryEvent managed, CocoaSdk.Sen
...
 
-    public static SentryEvent? ToSentryEvent(this CocoaSdk.SentryObjCEvent sentryEvent)
+    public static SentryEvent? ToSentryEvent(this CocoaSdk.SentryObjCEvent sentryEvent, IDiagnosticLogger? logger = null)
     {
-        using var stream = sentryEvent.ToJsonStream();
-        if (stream == null)
-            return null;
+        // Unlike the classic SentryEvent, SentryObjCEvent exposes no serialize() to round-trip through
+        // JSON, so we reconstruct the managed event from its properties. We map the same subset that
+        // CopyToCocoaSentryEvent writes back, plus the exception derived from the native NSError.
```
jamescrosswell: I guess there's some symmetry to this (matching CopyToCocoaSentryEvent) so we could go down that route. It would be nicer/easier if we could rely on serialisation to convert events from Cocoa to .NET and back again (we have to maintain the serialisation code to be able to send this stuff to Sentry anyway - so we get .NET <-> Cocoa conversion for free if we do this).

scripts/generate-cocoa-bindings.ps1
```diff
@@ -155,18 +156,16 @@ else
 }
 
 # Generate bindings
-# The SentryObjC*.h headers expose the structured hybrid API (SentryObjCSDK.internal)
+# We bind the SentryObjC umbrella header only. It exposes the full public surface
+# (options, SDK entry point, event model, scope, enums) plus the structured hybrid API
+# (SentryObjCSDK.internal, incl. debug images) via SentryObjC* classes. The classic
+# Sentry.framework ObjC surface is no longer bound here; the few remaining classic
+# holdouts (scope.applyToEvent enrichment + the SentrySerializable protocol) are declared
+# by hand in PrivateApiDefinitions.cs. See getsentry/sentry-dotnet#5444.
```
jpnurmi: the comment looks slightly outdated becase `PrivateApiDefinitions.cs` no longer exists

scripts/generate-cocoa-bindings.ps1
jpnurmi: this can be removed because `PrivatesHeader.h` is no longer used

scripts/generate-cocoa-bindings.ps1
```diff
@@ -114,6 +114,7 @@ foreach ($file in $filesToPatch)
     if (Test-Path $file)
     {
         $content = Get-Content -Path $file -Raw
+        $content = $content -replace '<SentryObjC/([^>]+)>', '"$1"'
         $content = $content -replace '<Sentry/([^>]+)>', '"$1"'
         $content = $content -replace '#\s*import SENTRY_HEADER\(([^)]+)\)', '#import "$1.h"'
```
jpnurmi: these can probably be removed. the `<Sentry/*.h>` headers are no longer used, and the problematic `SENTRY_HEADER` macro is not used in `SentryObjC` headers

EDIT: unless we're able to drop the whole "filesToPatch" loop - see below

scripts/generate-cocoa-bindings.ps1
```diff
@@ -155,18 +156,16 @@ else
 }
```
jpnurmi: this can be removed because `Sentry-Swift.h` is no longer used

scripts/generate-cocoa-bindings.ps1
jpnurmi:
```suggestion
## Imports in the various header files are provided in the "new" style of:
#     `#import <SentryObjC/SomeHeader.h>`
# ...instead of:
#     `#import "SomeHeader.h"`
```

EDIT: unless we're able to drop the whole "filesToPatch" loop - see below

jpnurmi: a quick local test suggests that the whole "filesToPatch" loop can probably be dropped. all `SentryObjC` headers have appropriate `__has_include` guards and don't use any problematic `SENTRY_HEADER` macros 🎉

src/Sentry/Platforms/Cocoa/Sentry.Cocoa.props
```diff
@@ -6,10 +6,9 @@
 
   <ItemGroup>
     <Using Include="Foundation" />
-    <Using Include="Sentry.CocoaSdk.SentryOptions" Alias="SentryCocoaSdkOptions" />
-    <Using Include="Sentry.CocoaSdk.SentrySDK" Alias="SentryCocoaSdk" />
+    <Using Include="Sentry.CocoaSdk.SentryObjCOptions" Alias="SentryCocoaSdkOptions" />
+    <Using Include="Sentry.CocoaSdk.SentryObjCSDK" Alias="SentryCocoaSdk" />
     <Using Include="Sentry.CocoaSdk.SentryObjCSDK" Alias="SentryCocoaHybridSdk" />
-    <Using Include="Sentry.CocoaSdk.SentryDependencyContainer" Alias="SentryCocoaDependencyContainer" />
```
jpnurmi, line 12: Yay, feels good to get rid of this old hack!

jpnurmi (commented):
I tried quickly removing the whole `filesToPatch` loop, and both `PrivateHeader.h` and `Sentry-Swift.h` modifications from `generate-cocoa-bindings.ps1‎`. A clean build passed fine without any of those, so I believe we might be able to clean up the old workarounds quite a bit. Could you confirm?

jpnurmi (approved):

PR #5455, opened 2026-07-28
ci: inline Alpine node setup to fix flaky linux-musl CI

### LINKED ISSUE #5408: CI: flaky '.NET (linux-musl-arm64)' — 'Initialize Alpine Linux' curl|bash fails on transient raw.githubusercontent fetch
## Summary

The **`.NET (linux-musl-arm64)`** job in the `build` workflow (`.github/workflows/build.yml`) fails **intermittently** at the **"Initialize Alpine Linux"** step. It is flaky — it fails on some `main` pushes and passes on others with no relevant code change:

| build run | commit | `.NET (linux-musl-arm64)` |
|-----------|--------|---------------------------|
| 29707134747 | 8ae5d529 | ❌ failure |
| 29706770989 | 62658120 | ✅ success |
| 29701980739 | 253719e7 | ❌ failure |
| 29701978671 | 4c3ca8cb | ✅ success |
| 29701976936 | a212c214 | ✅ success |

(`sentry-native (linux-musl-arm64)` passes every time — only the `.NET` leg is affected.)

This is **unrelated** to the Xcode 26.6 breakage in #5406 and unrelated to the recent dependency bumps — it's a pre-existing fragility in the Alpine setup step.

## What fails

The "Initialize Alpine Linux" step runs:

```
curl -sSL https://raw.githubusercontent.com/getsentry/sentry-dotnet/<sha>/.github/alpine/setup-node.sh | sudo bash /dev/stdin
```

On a failing run the log shows a ~31-second gap between the `curl` starting and the error, then:

```
/dev/stdin: line 2: syntax error near unexpected token `newline'
##[error]Process completed with exit code 2.
```

`setup-node.sh` itself is fine (a small POSIX `sh` script that works around *"JavaScript Actions in Alpine containers are only supported on x64 Linux runners"*). The `line 2: syntax error near unexpected token 'newline'` is bash choking on **non-script content** — i.e. `curl` returned something other than the script (an error/rate-limit page or a truncated/empty body from `raw.githubusercontent.com`) and it got piped straight into `bash`.

## Why it's fragile

The step pipes a network fetch directly into a shell with no safety net:

- **No `curl --fail`** — on an HTTP 4xx/5xx (including GitHub raw rate-limiting), `curl -sSL` prints the error body to stdout instead of failing, and that body is executed by `bash`.
- **No retry** — a single transient hiccup fails the whole job.
- **No integrity check** — a truncated/partial response is executed as-is.
- **Unnecessary network dependency** — the script already exists in the repo at `.github/alpine/setup-node.sh`; the job is building the checked-out repo, so it doesn't need to re-download the file from `raw.githubusercontent.com` at all.

## Suggested fix

Prefer running the checked-out copy instead of curl-piping from the network, e.g.:

```
sudo sh .github/alpine/setup-node.sh
```

If fetching over the network is required for ordering reasons, at minimum harden it: `curl --fail --retry 3 --retry-all-errors -sSL <url> -o /tmp/setup-node.sh` and run the saved file (optionally verifying a checksum) rather than piping an unchecked stream into `bash`.

## References

- Example failing run: https://github.com/getsentry/sentry-dotnet/actions/runs/29707134747/job/88246002629
- Example passing run (same job, adjacent commit): https://github.com/getsentry/sentry-dotnet/actions/runs/29706770989

jamescrosswell:
## Summary

The `linux-musl-arm64` "Initialize Alpine Linux" step has been failing on almost every push for the last few days (#5408). It piped a network-fetched script straight into a shell:

```
curl -sSL https://raw.githubusercontent.com/.../.github/alpine/setup-node.sh | sudo bash /dev/stdin
```

With no `--fail`/retry and a raw pipe, a rate-limited or truncated fetch got executed as-is, failing with `line 2: syntax error near unexpected token 'newline'`. It's non-deterministic — the same script passes and fails on adjacent commits.

## Detailed explanation

GitHub's Actions runner has a hard limitation: [JavaScript actions (like `actions/checkout`) inside Alpine containers are only supported on x64 Linux runners](https://github.com/actions/runner/blob/8a9b96806d12343f7d123c669e29c629138023dd/src/Runner.Worker/Handlers/StepHost.cs#L283-L290). On the `linux-musl-arm64` job (Alpine on ARM), the runner normally refuses to run any JS action.

Our `build.yml` contained  a workaround that runs **before** `Checkout`, because checkout *is* a JS action — it needs the workaround in place first. 

```yaml
run: |
  curl -sSL https://raw.githubusercontent.com/${{ github.repository }}/${{ github.sha }}/.github/alpine/setup-node.sh | sudo bash /dev/stdin
```

This **downloaded** a script (`setup-node.sh`) from GitHub's raw CDN and piped it straight into `sudo bash`. That script contained:

```sh
if [ "$(uname -m)" != "x86_64" ]; then
    for node in /__e/node*; do
        mkdir -p $node/bin
        ln -s /usr/bin/node $node/bin/node
        ln -s /usr/bin/npm  $node/bin/npm
    done
    sed -i 's/ID=alpine/ID=unknown/' /usr/lib/os-release
fi
```

The downloading of the script is the bit that frequently fails - the script itself runs fine. 

## Solution

The solution is to move the logic that the downloaded script would have run directly into our `build.yml` file... to avoid the flaky download. So the build workflow does exactly the same thing as before, but without a flaky network dependency. 

### Issues

Fixes #5408

#skip-changelog


.github/workflows/build.yml
```diff
@@ -44,8 +44,21 @@ jobs:
     steps:
       - name: Initialize Alpine Linux
         if: ${{ contains(matrix.container.image, 'alpine') }}
+        # A workaround for "JavaScript Actions in Alpine containers are only supported on x64 Linux runners."
+        # https://github.com/actions/runner/blob/8a9b96806d12343f7d123c669e29c629138023dd/src/Runner.Worker/Handlers/StepHost.cs#L283-L290
+        # Inlined rather than fetched over the network: this step runs before checkout (checkout is itself a
+        # JavaScript action that needs node symlinked first), so it can't be read from the working tree, and
+        # piping a curl download straight into a shell made CI flaky when the fetch was rate-limited or
+        # truncated. See https://github.com/getsentry/sentry-dotnet/issues/5408.
```
jamescrosswell:
```suggestion
        # Enable JavaScript Actions in Alpine containers on Arm64 runners
        # See https://github.com/getsentry/sentry-dotnet/pull/5455
```

.github/workflows/build.yml
```diff
@@ -108,8 +121,21 @@ jobs:
...
```
jamescrosswell:
```suggestion
        # Enable JavaScript Actions in Alpine containers on Arm64 runners
        # See https://github.com/getsentry/sentry-dotnet/pull/5455
```

jpnurmi (approved):

